Issue #9518, opened 2021-04-29
Custom shaders for models and 3D Tiles


lilleyse:
We're in the process of refactoring the glTF / model system and one of the end goals in the next few months is to add support for custom shaders similar to what you'd find in other engines (see Three.js [ShaderMaterial](https://threejs.org/docs/#api/en/materials/ShaderMaterial) and Bablyon.js [ShaderMaterial](https://doc.babylonjs.com/advanced_topics/introToShaders/shaderMaterial)). This will give developers full control over visualization of their models and 3D Tiles.

For background CesiumJS already has various levels of support for custom shaders:

* [Fabric](https://github.com/CesiumGS/cesium/wiki/Fabric#source) for primitives
* [Post processing stages](https://cesium.com/docs/cesiumjs-ref-doc/PostProcessStage.html?classFilter=postprocess)
* glTF 2.0 models with the [KHR_techniques_webgl](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_techniques_webgl) extension
* [Declarative Styling](https://github.com/CesiumGS/3d-tiles/tree/master/specification/Styling) - not strictly custom shaders, but custom styling of features based on their metadata

With 3D Tiles Next around the corner we have new methods of storing metadata including per-texel metadata that are ready to be unlocked.

### Approaches

Two possible approaches for supporting custom shaders are described below:

1. Add a custom shader callback to the fragment shader that has well defined inputs: vertex attributes, custom uniforms (including textures); and well defined outputs: color, show. Attributes that are unused can be automatically disabled (for example if normals are unused they'll be disabled in the `VertexArray`). We could also add a callback to the vertex shader where outputs might be position and point size, and maybe some abstraction for varyings. This is roughly similar to [Fabric](https://github.com/CesiumGS/cesium/wiki/Fabric#source) and [Post processing stages](https://cesium.com/docs/cesiumjs-ref-doc/PostProcessStage.html?classFilter=postprocess).
2. Give full control over model building - creating textures, buffers, shaders, uniforms, etc. This offers the most control but is probably overkill and would likely require exposing the private `Renderer` API. This is similar to [KHR_techniques_webgl](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_techniques_webgl).

For now I'm leaning towards option 1. A rough example might be

```glsl
// Color fragments based on a per-vertex classification

// a_classification and a_normal are available as vertex attributes
[...]

PR #9722, opened 2021-08-12
Adds boundingSphere and debugShowBoundingVolume to ModelExperimental

sanjeetsuhag:
This is a pretty simple PR in terms of features being added, but it'll pave the path for how we want to handle dynamic updates (cases where we need to modify or reset draw commands).

Additionally, implementing this revealed a flaw in the way we were handling the `readyPromise`. The model is not considered ready in `Model.js` until the first `update()`, so I fixed that here.

[...]

PR #9725, opened 2021-08-12
Add custom shaders to `ModelExperimental`

ptrgags:
WIP experimental implementation of custom shaders for `ModelExperimental` Opening the PR early for feedback and so I can take notes on things that need to be changed.

Description last updated: 2021-08-16: 13:45 EDT

## Overview

This is the core functionality of #9518, though there will be additional PRs in the future to add the metadata capabilities.

Custom shaders allows creating custom vertex and fragment shader code that can be applied anywhere a Model (for now, `ModelExperimental`) is used.

Note that this PR only implements the basic implementation. A future PR will add metadata to custom shaders, though this requires more architecture elsewhere in `ModelExperimental`-related classes.

There are two ways to try this feature right now:
* Load a tileset that uses .glb tiles via `3DTILES_content_gltf`. This requires `Cesium.ExperimentalFeatures.enableModelExperimental `. B3DM and other tile types are not supported yet, we've only wired up `ModelExperimental` to `Gltf3DTileContent` so far.
* Load a GLB or GLTF file via `ModelExperimental.fromGltf()`

## Sandcastle Examples

EDIT 2021-08-25: Due to recent changes the following examples won't work without a small tweak to the vertex shader, the new signature should be:

```glsl
void vertexMain(VertexInput vsInput, inout vec3 position) {
}
```

* [Simple example of tinting the baseColor texture](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/model-experimental-custom-shaders/Apps/Sandcastle/index.html#c=fVZrT+M4FP0rFp9SqaQtA+wOLy1TKFPUUqYEGEaVGDdxG28dOxM7lLKa/77HzqPloZFQiq/v69x7fO0u0zxP/L6SfsRmNBfmNAyZ1oFaMEmOyWSLrS7j6UXIR/yyf/vS71zxvu7L8V7Y7e/3F+n3u+7lZx9Kv6KLhVVaPCR38ejsjg9X7dXDzrflIBgnVy+nZnQ/3LvikN1fxsOz087wvpf8uOjzQfcy/QFnw2C+NwwWup+IOHLrh9XoLNyFbDX63vZ/nF1/jdJkl/Ue8ufwZi53754H+zvXsWJ3s71g/uvsurezuv3M9r9Ntg4nciK7Bbjz55RlPGHSUNFj1OQZ0z6TdCrYUEVMbO4Dssly5syfaEaeOFuyDFLJlqT0d+dk3mQrdOuugiGXLJtsNWo7bWhmAjiFKbzPVJZQGTJfqqW31gpzbVRyE9PobYzuxo7330QSkktuvegD4paEtFrk/JkmqWBEzYhmxnA5J7R0WujkjwY51CaEPFGRY12n16w24C0YnY0OCI0iovM0VZkhCEgMe7YVq+NXBmaVwlGZ7m2xGUDm9waj06D0+9v9Fl9E6OZZhjITlRqupHb+dYE+QSeIJ7g2iD79l4WGGEXCmMo5IxFLmYwsPCVJrtksFxIkbZDS73B01u89PA5Pg/Nx/3RAvJLLDbJNvIQatBet3T4pa1PFhEDweWzr1iAmZoTLNDekNlhyIciUkTRj26lKc4GNCFITV1UBhAz+rO1cBD3iTalmXSVU1iTS9lwgySDmmuCvSLwoKpeuW7UbyGiaipUVpkqb7TRT9hzaNZvNUA5dYh2fXw9Ou+drsID4B1gXDMxE3iBGjcvlrEOIw9hHeqzMda2xEd3LGI02c21YLHrB05RFfg1urlTkoDnTKM/gpbLwbeq2wQcf0dueQf9NByvC2NzMUtV4nBdNKPh4/WVMqIzI7dWgH5TqVlaeNr22CakIbe8s5UpFZ1RwDgcH2YvHXkbnrnPE9jCyTKvK4Ud8NkP3XLxaSEUa0yZhTyxbmdg1SkDHlmOOeqI4NliVhZs0Nf7BptR32VSQx3YLDVlaHnCbtHMpXbaC6JRijNRlf39WPDuyCjWSOF8gobI0jplILQHhSytMphA4dcOGBQT0qmhHgKYdkJ+FOPxU7g0x4bw792/fHZIn7X6bhVKVaqOaNBlD52UtP3RjwH5+VjhrrrmRZNu1wEEvCt3SKbMty1ynqdZ5go4Yh0Jg1NLM1V2SZYzPK3LYotSUmKGndq6/R6Z4VO86bL1yUaCbVei4VFiGL8mjK/+wSrrKvrExjRO6YDrmM0OWdjeherHm0tebL0CHuYtboDiAU8AubF0JYz3FFWAjjS++BAr63lv+NQ4L/ZlQ1DhvPTj6Aj+wpFPtwYf/jIHQ9vf39yttZCZA3KFN55h0/DYUXIJW8srjhpo2LPXafqf5Kkxxc5VO7Qxzk8+aYbysmBBquT3PGJMbuFTG51xSUZyt43eH6nBD17pEuaDEn71XhgXPkNFeE+j+tp9Ow373KpjvDuuHXooIzRpqY4OZE/m7AIjL2V3xvg4xPX07j2/TCP59XI7n4J0ZoCh2sHqzXIaW4F5JhM0b3cdoKS9GPBaKq3iy1STeuwcBOlLfyA3SIp12u21T+b1+KhguGPwB1evc8HDBCXvCkEVunk1h8xXhfj6dBYWxt+bqNaLZZ4Pt30gycq8yEZGAZRkOA7lHK0kPKaIZ9gfc6Pg7jswYlWpa8jbP1hMNL8gx0yrPgMnS+1QjYD/y9nZ3d/7abTTryCmOsx1BNvLre6sSYkzjDRYalfnk3E5XB764jdHYNQHtDNQUk6xwUGxsduDg1cp1GvOuLOofylgW2yqWai9KJYFab2w1t460WQl2Ur2H/uGJezOhKJ7vtwzDywyU0a1pHi6Y8UO8V0qqHrU2TY8i/kR4dPzBe5KEAtXCDsa2uOEvoM/JUQv670xxfu0lPUK5BF1ZtbhzMiiEvu8ftbD82NIoJaY0e+P5fw) (Updated 2021-08-16 13:40) -- this shows how the `MODIFY_MATERIAL` mode can be used to update the textures 
image
* [PBR Shading the Ocean](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/model-experimental-custom-shaders/Apps/Sandcastle/index.html#c=fVdhc+I4Ev0rqnw5c8sYwpBcbSYzdRkIWVIQsoyTTKaoygpbYC2y5bVkCLma/35Pkm1ISE1VCpDU3eru1/3U6THFi8QfytSP2IIWQl+EIVMqkCuWks9kdsS21/H8KuQTfj28exke3/ChGqbTk7A3PB2usu/3vevffQj9E12tjNDqMbmPJ/17Pt62t4+dPzejYJrcvFzoycP45IZj7+E6HvcvjscPg+TH1ZCPetfZDxgbB8uTcbBSw0TEkV0/bif9sIu97eR72//Rv/0jypIuGzwWz+G3Zdq9fx6ddm5jye4XJ8Hyn/7toLO9+52d/jk7+jRLZ2nPBXf5nLGcJyzVVAwY1UXOlM9SOhdsLCMm9s8Rss4LZtXXNCdrzjYsx27KNqS0d2/3vNlRaNc9CUWesnx21Kj1lKa5DmAUqrC+kHlC05D5qdx4O6mwUFom32Iavb2jt3fi/W+WElKk3FhRZ8QuCWm1yOUzTTLBiFwQxbTm6ZLQ0qiTKZ40fKhVCFlTUWBdu9esDmAtmPQnZ4RGEVFFlslcE1xINHs2GavvrxT0NoOh0t07dxhgzx+MJhdBafen/XafuKFX5DnSTGSmuUyVta9c9AmQIJ7gSuP2+d8s1ERLEsY0XTISsYylkQlPpqRQbFGIFEXaKM2OJ/3h4PFpfBFcTocXI/KBeAnVwBR4fvhSJqS6CBuCL2OTrAbRMSM8zQpNaoUNF4LMGcly9iGTWSFwEGFXx1Uq4HcOe0Z3KYIB8eZUsZ4UMm+S1AAtVIMEMVcEf85bl0meWohqM9ijWSa2ZjOTSn/Icmmaz6zZYoEcqDLC6eXt6KJ3+SrEX4R1xVCO8BvVUMdlfVYhtsPYh3us9HUnsXe7lzMa7fvaMLGoFc8yFvl1cEspIxuaVY2KHFYqDd+4blA9e6+mTeP5b3CrqsT4pjeyjsdaUYSiCG+/TglNI3J3MxoGpbjZK1tM7XRCKkKDnamzUtAquUJDt8B78TTI6dIiRwyGkSmvKh1+xBcLoGfvqzepyGLaJGzN8q2OLVACMiYdS+QTyTGXVV5YeqnjH+3v+nC7CnhqDgDHxlQBNy5bg6n1VRCVUTBHnfTD9vAMSzkxklhbKEFpijhmIjPlB1tKgoxCROnaBgEAKQdGAMjOyF9uO/xYno1Bat69/Tm0LbJW9rvphCpXGxW55Ay4p/X+J9v55uOvKs660iwLGbBW6G2X5pbKmAEstzhTpYoEeGgbhQC70txmPSWbGB+vSsMkpS6IBRA1VH4YmeRRfWpjG5QLF92iio6nEsvwJXmy6R9XTlfeN/YIOKErpmK+0GRjThOqVrtK+uPbV0QHqgXxu/abI2yna1MYqzlY39w0vfoaSMh7b6uv8cnJL4Sk2lobwNBX2IEmnSsPNvxn0EHbPz09fS1tXRoblz6TY78NIaVZ5rX94+YrS5VWHdXIlDySCq5GDhEMRfeRXBbLuIKrzkat9GA2DrUimq9s3CRheF8FD3dl4P0NRigfART/Jqb6X8oUkOJ4mxu16RsJKtM4JTyhqHeX4boGcGcqNdkyc47n0GCKc3TAwZWqtmlpXkgJ5/ByGs9Ty+iLwrDXHkgCyXAU8fmAGz7tydlsV4JmB4lunzRJ9dmt0lwbsQk175ixzJ89YGQEPzZ3yB3o1F3iVKp7uo3mngO/MlDh5/Tr4Jq1y4032nXGqNY5nxe6atzg8ntvMpn2n9quZQuQO/gWxCu2Zopw7Wtz6lrL37OgMGWFuk5fhxRruHQo6IN9elLm0VP7bZWO0XuO1TBKcepo04FYvsJNAhvoRPhTHtjIXvfImh3jZrx6Xgc98m/bjhnHj65dFmvbXiduYcept122Zp33LURSG4Q6DtmPNrfFugFz3XfM1RA573Htbw6kttF2jv6G+jiBpr30AGbHtrP0p8MNM6adVH0VYh7wzYRxl0VQ8YHOJbhUj8ACZlTwFkUamgR6JbntD6Y+HstyvsPM63yeHTWJdzDXWoYpB8sGaZHjdrttXPm5m3g1Fwz2TIu88g3zN16NNQCHb55xYX8Ytl8f+4FT9nb8e4vbzPRrsJ2kjDzIXEQkYHkOgicPDD0+gIsoRfMFtI79jqUPPP5yXrZ5ke/eaPwjNGVKFjliMpR9oXDhMPJOut3Of9Bl9c0Znqiqql5PYtUmBg9b5DL3yaWZF2zwjnjQgbUlbd51RVHHzoA72Efg7NXKIg1yLJP6izSWyTaCpdiLlEkgdwdHzaNzpbeCfanG+v+CQs3oj6R4vt/SDIyKklGteRGumPZDjN1lwZ639lXPI74mPPr8zr9FJBTIFk4wiohv/AXl8+W8BfkDVXSUGTsnSJegWyMWH38ZuU3f989bWL6vqaUUc5q/sfx/) (Updated 2021-08-16 13:40) -- this demonstrates how you can set the PBR properties properties like `material.specular` or `material.roughness`.
image
* [Custom glTF attributes example](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/model-experimental-custom-shaders/Apps/Sandcastle/index.html#c=fVZtTxs5EP4rFp82d7AJFHoqB+hyCeGCEkLpAqWKRJ1dJ+vDa29tb0I48d/vsfeFAFUltMHjeZ9nZtxjhhdZOFQyTNicFsJ245gZE6kHJskxmW6x9Xk6O4v5hJ8Pr5+Guxd8aIby6iDuDT8OH/KvN73zTyGYfiRnD47p4S67SSf9Gz5ed9Z3e59Xo+gqu3jq2snt+OCCg3Z7no773d3x7SD7djbko955/g3KxtHiYBw9mGEm0sSf79aTfrwP2nrytRN+61/+k+TZPhvcFY/xl4Xcv3kcfdy7TBW7mR9Eix/9y8He+voT+/h5uvXnVE5lrwzu9DFnmmdMWioGjNpCMxMySWeCjVXCxOY9Qra6YF58STVZcrZiGlTJVqTSd+NpwXQr9ueegiCXTE+3Wo2csVTbCEohCu1zpTMqYxZKtQpeuOLCWJV9SWny1kZv4yb4byoJKSR3Wswh8UdC2m1y+kizXDCi5sQwa7lcEFopLXmKewsfGhFCllQUODfubdcX0BZN+pNDQpOEmCLPlbYEBolljy5jjf1awK5zKKrcvS4vI9DCwWjSjSq9z/63/MJCr9AaaSYqt1xJ4/WbMvoMlSCB4MbC+uxfFltiFYlTKheMJCxnMnHhKUkKw+aFkABpq1I7nvSHg7v7cTc6vRp2R2SHBBm1qCnquXNSJaQ2BILgi9Qlq0VsygiXeWFJI7DiQpAZI7lmO7nKC4GLBFSb1qmA3xr6nOxCRAMSzKhhPSWU3ibSFVqYFolSbgj+Sm/LTHLpS9SoAY3muVg7Yq6M3cm1cs3nzmw+Rw5MFeHV6eWo2zt9FeIvwjpjgCP8BhqauLzPJgY5TkO4xypfXzg2rAea0WTT15aLxTzwPGdJ2AS3UCrxoXnRpNDQUkuEznVX1cOfYdo1XvimbjVKnG92pZp4vBZDKEB4+fcVoTIh1xejYVSxO1rVYuZFJqYidrVzOKsYvVAJNHQLvBf3A00XvnLE1TBx8KrTESZ8Pkf1vL2GSEWe0m3ClkyvbeoLJcDj0rFAPpEcZ6z2wo+XJv7RJjWE23XAV+4C5Vg5FHDnslcova+CmJxicjRJf98egZtSJRvJvC5AUDkQp0zkDn7QZRSGUYwoy7ZBAKhUWYwIJTsk30ty/KG6G2OoBTf+36FvkaXxv9slU+1qa2MejekD831ROl7IxLePM++ovz+ShGv47iN8FV6pYi4UtWRFl25sdsJd8hsBHIMPYQf/1QbDR6C/HGxukjo5zQA5+ZK9372LQSfsbBP/cSpL5mf3+V5nvsG+n4sOPgigKnzb5MxBSHvkUWOKDAixPq8C855qjwNJVik+r8DqytRAdA6MueXyPteKJ82tz/agOpT5ntf55lLhGD9l9z5j49rp2vvNEgwJjWOe+GUm1mBxaEfpS7DaFPnVLOMyMSTzi2OuuCjjVrLRUshYZRkyaXU5B2Kqk3JgFagk5ljM7RprhMcPTBvXNx5fmANJUY2tCkwzqteoZhVMSK3VfFZYDAh3E0OX5vF9rBQMVOVsuq2pwLFX8/Zaq2KRukXg0bL39rpu4OMaDfsODeVnEwxT+VyuZCxlv9pDA7dY6EbydZ5AXYileIpS2xH2k5utwRwZclALqtxvbvIQ06VaiHgklEidbm2T4N1DAEBuNnGLtMlup9Nxrjy/PBEsFwz6XBCvfMODBVhfIo3wLXAubL4e/M+HflQKBy/wuIQ1V3XXjhPJyK3SIiER0xr4I7cMu28AFwF09wOw7IZ7fjZiWqpZ1aaFfhlqeDleMaMKjZjciukaGBwmwcH+/oc/Oq3txnKODqoHwevVVRMxqfH2iq3SITl1A9YHXy5k1LHRZN0gNBRwKxWUF5sVOHx18pXG0KuS+os0Vsl2jBXbk1JZpF4utra3joxdC3ZSv4P+4pl/KyEpQRi2LcOLDJAx7VmB3rBhjHdKhcyj9qboUcKXhCfHP3lHklggW7jB7BZf+BPgc3LUBv87UQxM158TpEvQtWNLd09GJTEMw6M2jj+XtEoJNNUbzf8D) (Updated 2021-08-16 13:40) -- in this case I made a model with a custom attribute `_BARYCENTRIC_COORDS`. This is accessible via `fsInput.attributes.barycentric_coords`.
image
* [Custom varyings](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/model-experimental-custom-shaders/Apps/Sandcastle/index.html#c=fVf7c9o4EP5XNPnJ9Ighr941TTuXI5CSgZKmTtJ0mEmFLbAP23IlmUdu+r/ft/IDSNrOJAav9r3frpaO0FGeuH2ZuoGY8jw2574vtPbkXKTsHRvvifVVOLn0o1F01b996h98jPq6n96c+J3+6/48+3LXuXrjgul7cDknpvlDcheOLu6i4bq9fjj8tBx4N8nHp3Mzuh+efIxAu78KhxfnB8P7XvL1sh8NOlfZVygberOToTfX/SQOA/v+sB5d+MegrUdf2u7Xi+sPQZYci95DvvI/z9Lju9Xg9eF1KMXd9MSbfb+47h2ub9+I15/Ge2/H6TjtFMF1V5lQUSJSw+Oe4CZXQrsi5ZNYDGUg4u1zhGxULqz4giu2iMRSKFBTsWSlvjtLc8Z7vn3vSAhGqVDjvUYtpw1XxoNSiEL7VKqEp75wU7l0Nlx+ro1MPoc8eG6js3Xi/DdOGcvTiLToU2ZfGWu1WHfFkywWTE6ZFsZE6YzxUmnBkz8a+FCLMLbgcY732r1mdQBt3uhidMp4EDCdZ5lUhsEgM2JFGavtVwJmnUFR6e5tceiB5vYGo3Ov4Pph1RdPxLuGgxv/F49BBD+QlZ6SiSezWttdwbqrrdACNzu5UqgVk5mJZKqtk7pIYYJyMieGWoQw+Vf4hhnJ/JCnM8ECkYk0oBzJlOVaTPM4BdIbpdrh6KLfe3gcnnvdm/75gO0zJ+EGwAAo9t+XWa0MgRBHs5Ay3mAmFCxKs9ywWmAZxTGbCJYpsZ/JLI9xEIBqwiqf8BtRW9lZ7PWYM+FadGQsVZOlhJZYN5gXRprhr/C2KEeU2jrXakDjWRZTwlgmtdnPlKQOpncxnSIHuozwpns9OO90d0L8TViXApiG34BUHZf1Wfsg+6EL90Tp64Zjy7qjBA+2fW1QLHoeZZkI3Dq4mZSBDc2KBrmClkrCJdepqqc/awzqXvdZ3SqUkG9mKet4rBbNOJB8/c8N42nAbj8O+l7JTrSyT/VGxuexT7UjnJWMVqgAGloO3sePPcVntnKMahgQvKp0uEE0naJ61l5N5HEW8iYTC6HWJrSFisFD6Zghn0gOGau8sDOqjn+wTXXhdhXwDR2gHEtCQUQuW4Wp9TVmOuMYP3XSX7aHQ6OuYGOJ1QUISgJxKOKM4AddWmKi+YiyaBsEgEoVxfBQslP2rSD7R+XZEJPRubNf+7ZFFtp+NgumytXG1lCjyiFlmI4YrErGmoUIqgRqNUVsIVGAQgqcaPlpLLlhnz90B4PHD93+5QcPM/XQbb/9xbzBaUVxyBun7babzD62lTSaGzdLVbWvCZ8L28MaSYoxIwPb6pQqov6xggWFPNtq7JSiUFG4vOQLuifa7gF7xdA6zpHbxrfKqrtCpxaTvHJACbRHuqn0H+xZAKSyYP5Bj28VSuo+tRcBQR0BlCBt6UwQ3JVNLtc6T4BmYzEQ44LjymI2ZcsQj53GIkjV7TRFP9Bt+hIXMgrqU4uMXvlSYGNaYSNKJV79p+TRZmxYOV15X8PlWf5OkIi2e3hSpvHQpnE3cxZ2E6AIAqU9lxujokluMG/oxIdDKvIffSlVoN9uW0qiFOQkkyndQO/o3aF/EnNXTavYXTfKL0+NHWGfIPKOEuAbZ1fTK8LpLjeN2kdyxa4GfKIdim/fatnlTLieg0UnUppQG5GB87DEwZaWXSFC/nMh8B/R40WnVKL1BKuRQilY1d3TaFY4PCD7f9HjCETy8IUOJfNZSDdwqaTkP/gFezVJdywelWEi4I3tdqmmsE2hbHfCOP1RLGBYwewi52rkR7h0d95mAcy5WIG6wLkZQJQuQWeap7aHnRJ423ubiylUrj9YCQuwjfeazHmx9qF49d7VYC120G7bmv/YLIQmigX0Ichd37CeotEXACh8c8iF7V3RfhxdeIWwsxml17BGyyHNolEq2L1UccA8oRSaj90THntwEV1OH0DFgXtoLzFca3JSzqhcbW4f/E64EVrmCjERtM41DPYD5+T4+OhPlL+2nGF8VFNwd8eoiDSyVe4bqVzWpZvQBl9sTqhzrcnQjaV5IkoF5cTfqsDpzputNG6nMqm/SWOZbGIs2Z4kwX1zsNfcO9NmHYv31db7d5TYzRhJcVy3ZQT2b0BGtya5PxfG9bFQlsg9a22LngXRgkXBu5/8amB+jGzhBJds/Dl6Anzen7XA/0IUnUsL1Qjpivma2MKD94OC6LruWQuvP5c0UsaYSc80/w8) (Updated 2021-08-16 13:45) -- this example shows how to define additional varyings, set them in the custom vertex shader and use them in the custom fragment shader.
image

## `CustomShader` Documentation

>Update 2021-08-23: The latest documentation for custom shaders is now in a [README file] on this branch. It is temporarily in `Source/Scene/ModelExperimental/CustomShaderGuide/` for now while this feature is experimental. 

CustomShader is a bit involved, so here's how it gets used. This may become the start of a user guide for custom shader, so I'll keep updating this as I update the PR.

Constructor:

```js
var customShader = new Cesium.CustomShader({
    // Any custom uniforms the user wants to add to the shader.
    // these can be changed at runtime via customShader.setUniform()
    uniforms: {
        u_time: {
            value: 0, // initial value
            type: UniformType.FLOAT
        },
    }
    // custom varyings that will appear in the custom vertex and fragment shader text.
    // the user is responsible for setting these in the vertex shader and using them
    // in the fragment shader.
    varyings: {
       v_customTexCoords: VaryingType.VEC2
    }, // optional
    // configure where in the fragment shader's materials/lighting pipeline the 
    // custom shader goes. More on this below. 
    mode: Cesium.CustomShaderMode.MODIFY_MATERIAL,
    // either PBR (physically-based rendering) or UNLIT depending on the desired
    // results.
    lightingModel: Cesium.LightingModel.PBR,
    // Custom vertex shader. Right now this is a function from model space -> model space.
    // VertexInput will be documented below
    // NOTE: which coordinate system is used here may change (e.g. could become model -> view space
    vertexShaderText: `
        vec3 vertexMain(VertexInput vsInput, vec3 position) {
            // code goes here. e.g. for a no-op:
            return position;
        }
    `,
    // Custom fragment shader. 
    // FragmentInput will be documented below
    // Regardless of the mode, this always takes in a material and modifies it in place.
    fragmentShaderText: `
        void fragmentMain(FragmentInput fsInput, inout czm_modelMaterial material) {
            // code goes here. e.g. to set the diffuse color to solid red:
            material.diffuse = vec3(1.0, 0.0, 0.0);
        }
    `,
});
```

### Custom Shader Modes

EDIT 2021-08-16: I removed `BEFORE_MATERIAL` and `AFTER_LIGHTING` to simplify the interface.

NOTE: the list of modes is subject to change. We might decide to make further simplifications.

The custom fragment shader is configurable so it can go before/after materials or lighting. here's a summary of what
modes are available.

| Mode | Fragment shader pipeline | Description |
|---------| -----------------------------------|----------------|
| `MODIFY_MATERIAL` (default) | material -> custom shader -> lighting | The custom shader modifies the results of the material stage |
| `REPLACE_MATERIAL` | custom shader -> lighting | Don't run the material stage at all, but procedurally generate it in the custom shader |

In the above, "material" does preprocessing of glTF textures, resulting in a `czm_modelMaterial`. This is mostly relevant for PBR, but even for UNLIT, the base color texture is handled.

### `VertexInput` struct

An automatically-generated GLSL struct that contains glTF attributes. In the future we may add automatic fields. in a future PR, metadata will be added.

Right now an example struct looks like this:

```glsl
// this struct represents the raw attribute values.
struct Attributes {
    // required semantics
    vec3 position; // model space position. always present as glTF POSITION is required
    
    // optional semantics (added depending on what's in the glTF)
    vec3 normal; // glTF NORMAL
    vec3 tangent;
    vec3 texCoord_0;
    // etc.
    
    // custom attribues
    vec3 custom_attribute; // corresponds to glTF attribute _CUSTOM_ATTRIBUTE
};

struct VertexInput {
    // raw attribute values
    Attributes attributes;
    // in the future we may add another struct for derived attributes (e.g. positionEC)
};
```

### `FragmentInput` struct

This struct is similar to `VertexInput`, but there are a few more variables for positions in various coordinate spaces.

```glsl
// this struct represents the raw glTF attributes. The varyings required to make this work are handled
// automatically. 
struct Attributes {
    // required semantics
    vec3 position; // Raw model space position. always present as glTF POSITION is required
    
    // optional semantics (added depending on what's in the glTF)
    vec3 normal; // glTF NORMAL
    vec3 tangent;
    vec3 texCoord_0;
    // etc.
    
    // custom attribues
    vec3 custom_attribute; // corresponds to glTF attribute _CUSTOM_ATTRIBUTE
};

struct FragmentInput {
    // raw glTF attribute values interpolated (but not normalized) from varyings.
    Attributes attributes;
    vec3 positionMC; // model space
    vec3 positionWC; // World coords (ECEF). Low precision.
    vec3 positionEC; // Eye coordinates
};
```

### `czm_modelMaterial` struct

This one is a built-in, see the [documentation comment](https://github.com/CesiumGS/cesium/blob/model-experimental-custom-shaders/Source/Shaders/Builtin/Structs/modelMaterial.glsl). This is similar to `czm_material` from the old Fabric system, but slightly different fields as this one supports PBR lighting.

This struct serves as the basic input/output of the fragment shader pipeline stages. For example:

* the material stage produces a material
* the lighting stage takes in a material, computes lighting, and stores the result into `material.diffuse`
* Custom shaders (regardless of where in the pipeline it is) takes in a material (even if it's a material with default values) and modifies this.

## To Do:

- [x] Document usage more thoroughly in this PR description
- [x] Add development sandcastle
- [x] Where should documentation go?
- [x] Add unit tests - the basic regiment of tests are implemented, I'm trying to keep them updated in each commit.
- [x] uniform and varying types should be enums, not raw strings
- [x] `uniform sampler2D`s should be added. This requires some resource management, as the custom shader will need to fetch images and create `Texture` objects
- [x] Discussion: if we allow texture samplers should we also allow cubemap samplers? though this should go beyond the scope of this PR. (see https://github.com/CesiumGS/cesium/issues/9741)
- [x] Consider adding a built-in uniform for implicit tile coordinates (when `3DTILES_implicit_tiling` is used) EDIT: We're holding off on this, I opened #9736 to track this feature.
- [x] Discuss whether the custom vertex shader should be in model space, view space, or some other space. Right now it's a callback that goes from model -> model space.
- [x] Implement various reference frames (model, cartesian, cartographic, and view space) EDIT: cartographic will have to wait. Opened https://github.com/CesiumGS/cesium/issues/9735 for this.
- [x] Discuss the shader modes more with @lilleyse and @sanjeetsuhag about the shader modes. We're considering changing the list of them. For example, instead of `AFTER_LIGHTING`, we could just have an optional `afterLighting` GLSL callback. This way you could both modify the material _and_ post-process the lighting EDIT: We decided to skip this for now
- [x] handle heterogenous data: what if a tileset has multiple types of models, some with normals, some without? This should output a warning and a sensible default should be used.
- [x] Update unit tests. Recent changes added new classes and significantly changed the behavior of `CustomShaderStage`


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
This morning I had some good discussion with @lilleyse about the design of custom shaders. I updated the corresponding issue #9722 to keep this PR thread focused on actionable items.

The first thing I plan to address is providing the fragment shader with more reference frames for position (model, cartesian, cartographic and view space)

ptrgags:
Just pushed some changes:
* add `fsInput.positionMC` (model space position), `fsInput.positionWC` (32-bit precision ECEF), and `fsInput.positionEC` (eye space position). Cartographic space is not yet implemented, and may go beyond the scope of this first PR.
* added `UniformType` and `VaryingType`. So for example, when specifying uniforms, use `Cesium.UniformType.FLOAT` instead of `float`
* removed `CustomShaderMode.BEFORE_MATERIAL` and `CustomShaderMode.AFTER_LIGHTING` to simplify the lighting modes.

I'll update the description next and the sandcastles once the deployment finishes building.

ptrgags (commented):
Came across some questions while updating the implementation

Source/Scene/Cesium3DTileset.js
```diff
@@ -951,6 +951,16 @@ function Cesium3DTileset(options) {
    */
   this.metadata = undefined;
 
+  /**
+   * A custom shader to apply to the tileset. Only used for contents that use
+   * {@link ModelExperimental}
+   *
+   * @type {CustomShader}
+   * @private
```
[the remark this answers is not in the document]
sanjeetsuhag: I think it needs to be kept private while ModelExperimental is private, just like all our other stuff. Any reason to deviate from that?

Source/Scene/ModelExperimental/UniformType.js
```diff
@@ -0,0 +1,21 @@
+var UniformType = {
```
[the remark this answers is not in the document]
sanjeetsuhag: Stuff like this should live outside of `ModelExperimental` - someone working in another part of CesiumJS could utilize this.

Source/Scene/Cesium3DTileset.js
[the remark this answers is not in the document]
ptrgags: None, just wanted to make sure we were on the same page.

Source/Scene/ModelExperimental/UniformType.js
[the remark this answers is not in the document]
ptrgags: Right, though where should this live? in `Renderer` or elsewhere? @ebogo1 or @lilleyse any thoughts?

Source/Scene/ModelExperimental/UniformType.js
```diff
@@ -0,0 +1,21 @@
...
+  INT_VEC3: "ivec3",
+  INT_VEC4: "ivec4",
+  BOOL: "bool",
+  BOOL_VEC2: "bvec2",
+  BOOL_VEC3: "bvec3",
+  BOOL_VEC4: "bvec4",
+  MAT2: "mat2",
+  MAT3: "mat2",
+  MAT4: "mat4",
+  SAMPLER_2D: "sampler2D",
```
ptrgags, line 17: Textures are useful to support. 

The implementation should be similar to what `Material` does with textures. (hm... I wonder if we can factor out a helper function like `Texture.fromUri() -> Promise<Texture>` or if that's not easy to do given that you need access to the Context...)

The one thing though is managing the resources. @sanjeetsuhag and I talked about this, we think the model should own the CustomShader and therefore should call `customShader.destroy()` from `ModelExperimental.destroy()`. This is similar to how the scene owns its primitives.

Source/Scene/ModelExperimental/CustomShaderStage.js
```diff
@@ -0,0 +1,171 @@
...
+
+  // Fields for the Attribute struct. for example:
+  // "    vec3 position;"
+  var attributeField = "    " + glslType + " " + variableName + ";";
+
+  // Initializing attribute structs are just a matter of copying the
+  // attribute or varying: E.g.:
+  // "    vsInput.attributes.position = a_position;" -- vertex shader
+  // "    fsInput.attributes.position = v_position;" -- fragment shader
+  var vertexInitialization =
```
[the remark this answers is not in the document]
ptrgags: Oh yikes... default values aren't quite as feasible as I thought... how to determine what _type_ the attributes are? To know all the possible attributes, you'd need to know all the attributes across every primitive of every model of the tileset.

One thing you could do is disable the custom shader for primitives that are not compatible (e.g. their set of attributes doesn't cover the requirements of the shader). What do you think @sanjeetsuhag @lilleyse?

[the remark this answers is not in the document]
ptrgags: wait, actually you could infer the type for built-in things like `normal`, `texCoord_x`, it's only the custom attributes that would be completely unknown. And that could be handled by skipping the primitive with a warning.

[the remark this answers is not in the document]
ptrgags: https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#meshes -- well the inferencing would _nearly_ work, colors can technically be VEC3 _or_ VEC4. I think this would be enough of a corner case that we can just pick one

Source/Scene/ModelExperimental/CustomShader.js
```diff
@@ -0,0 +1,298 @@
...
+      "Uniform " +
+        uniformName +
+        " must be declared in the CustomShader constructor."
+    );
+  }
+  //>>includeEnd('debug');
+  var uniform = this.uniforms[uniformName];
+  if (uniform.type === UniformType.SAMPLER_2D) {
+    // Textures are fetched asynchronously and updated in update();
+    fetchTexture2D(this, uniformName, uniform.value);
```
ptrgags, line 209: @lilleyse I can load a texture the first time, but even though `fetchTexture2D()` is correctly updating the uniformMap, I don't see a visual change in the texture. Is there another step needed to mark the texture as dirty or something? 

[local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=hVcNc9s2Ev0rGM3dDHVRqI/YuavjZE4nWak8UuQ6sl2n6jgUCYmoQIAFQEtyJ//9HgCSlu20nbEtAVgsdt++/fCAalZk4ViKMKGrqOCmH8dU67ncUEHek0WD7s/T5ceYzdj5+Oph3P3ExnosLo/jwfjteJP/fD04/yGE0O/Jx40V2txm1+lseM2m+87+tvfTdjK/zD499M3sZnr8iWHv5jydDvvd6c0o+/JxzCaD8/wLlE3n6+PpfKPHGU8Tt77dz4bxEfb2s5874ZfhxY9Jnh3R0W2xiz+vxdH1bvK2d5FKer06nq9/H16MevurH+jbnxaNdwuxEAPv3Nkup4plVJiIj2hkCkV1SEW05HQqE8oPz+GyUQV11+8jRe4Z3VKFXUG3pNR37faCRSN264HERSaoWjSa9T1tImXmUIqr0L6SKotETEMht8GjVFxoI7PPaZQ8f2NwcBL8sRCEFIJZLfqEuCUh7TY520VZzimRK6KpMUysSVQq9TLFnYEN9RVC7iNeYF2b16oOoG0+G85OSJQkRBd5LpUheJAYurOI1e9XF8w+h6LS3Ct/OMdeOJrM+vNS77dWZUccGX1gx5/e/tyfXkzOLu96w9YzmxeN1Jj8pN3OeRTTDTOGijCWWbvb+4/9XTTKN+2HfxhODQqlEFkic8Ok0M4l7QHPEHwScKYNHF7+RmNDjCRxGok1JQnNqUgsolKQQtNVwQXyolmqnc6G49Ht3bQ/P7sc9yfkNQmyyIBGoNDrD2UMqoewwdk6tfFpEpNSwkReGFJf2DLOyZKSXNHXucwLjoMEuyat0IfdCvrs3TWfj0iwjDQdSC5ViwjLLa6bZJ4yTfDjrfXBY8KxolaDvSjP+d5u5lKb17mSNt/tmq5WwECXHl6eXUz6g7MnLv6FWx8pMgB2g4C1X85mHWM7TkOYR0tbHyUOXg8UjZJDW5vWF71heU6TsHZuLWXiXHNXk0JBS3UjtKbbqJ58L41srofP4laxxNpmtrL2x2nRJALvL/53SSKRkKtPk/G8FLd7ZVbrxztxxGMbO8uzUtBd8kRDgsJ6fjdS0dpFjtgYJpZeFRxhwlYrRM+9V29GPE+jFqH3VO1N6gLFIWPhWANPgGMfq6xwFa32f3K4G8LsyuFLe4BwbC0LmDXZKRTOVk50jhR7BP1legS2MHoxkjldoKC0JE4pzy39oEtL1L8YXvq0gQOIlA/GHCE7IV/9dvymPJuijgbX7uvYpci9dp8tL1SZ2jwogdNoQ11eeMMLkbj0sc/b3Vc7kjAF252HT9zzKlZcRoZso3tbqTthl/yLgI7Bm7CDb9WD4Q7s97XUFm97T1FQTjyi98qZGHTCTou4P1alF3YF6WuFfM19V9YsfeBAGfi2zqmlkHLMi7QuMjDEOFw5WkykHA8E2ab484SsNkw1RVfgmO1nL7GWLKlPHdqjcuHxXlV4MyGxjB+yO4fYtDK6sv4wBGMSxTFLXP/ke4hYtiP0nqwmBb6KZkwkmmSuV60k495vKWothUAtz4CkUb4OxJFKfMEqEEnUsZiZPToXizdUaZs3jl+oA0lRlq2STMtI7RHN0pkwMkaxZWFQIOxJDF2KxXexlHigDGedbXUE3tcBPWo+F1KyWKe2HTjO9N7Vbx8R2+fsEOFLUm8Y+NbXckaFu/0LXVXOv3dXQ7VeHpBmIb75aQHzgps6Qg3zaWhL91WeQEeIfn0GSpgJ+pitwcEKSFpKBmWMDoeMEFWo7LaYXzyjF40WCV7MKCB8PSQ0SZt0O52ONeXbwYzDHqj19Zfe8dsWOe72WpDqHf36zh8j4HRnAcIaz44FHAbnX9iH2LvaRo5RI2MJmrRIjiCjk1hsqGvfwMbKVjoD/+UV6TbJP70hIadibVIHXmUdJN3ZL078V3eGF7TkNORyDQj0Fu3QdXnQ3c8UfozzSgplJ8KvdvDQ35k8/vGHVf+t+vz67u/wtiG2eEOvAxOwlbh6zAzjFFcs+Z6EG+Mpysw9vES4A/vK4azoPt4M5/5y8JiZFwigTThbCWeCkhupeELmVCmkPrmhGDtGMA01xn4gT7thz7UlNCq5LCskbK37Cf5PuKRaFgo0sd29r21kk+D46OjNvzvNVv1yjuJV1eCnU0O1iTBg0o6NVCE5c/G3zvtZCPlQazK2B+kIme4V+INDkE+erFzyoN+UoP4FjCXYVrAUe5Aym8vHg0arcarNntMP1Sz6X5a5yRigBGHYNhTzN7JQt5cFypIJY4yIZYaftg+vnibsnrDk/Xf+ayAxB1o4Qdvkn0GkRePDaRvyL66iV9nSOANcPNpbsbT7YeI3wzA8bWP5/ZtGSo4K9Ezz/wE) -- every 5 seconds the texture should change.

Source/Scene/ModelExperimental/CustomShader.js
```diff
@@ -0,0 +1,298 @@
...
+  };
+}
+
+function createUniformFunction(customShader, uniformName) {
+  return function () {
+    return customShader.uniforms[uniformName].value;
+  };
+}
+
+function fetchTexture2D(customShader, uniformName, textureValue) {
```
ptrgags, line 126: This texture fetching code is a simplification of the rather convoluted code from `Material.js`.

@lilleyse I feel like we should have a more general image loader (GltfImageLoader is too gltf-specific) that can handle loading from a URI string, Resource, and other things (videos, KTX2 like the old Material, array buffers). What do you think? Though this is beyond the scope of this PR

Source/Scene/Cesium3DTileset.js
```diff
@@ -2663,6 +2673,11 @@ function update(tileset, frameState, passStatistics, passOptions) {
     return false;
   }
 
+  var customShader = tileset.customShader;
```
ptrgags, line 2676: @lilleyse another thing is I don't like that the CustomShader now has GPU resources that A. need to be loaded in `update()` and destroyed in destroy(). and B. it means that either the Tileset or the Model is now responsible for cleanup depending on if this is 3D tiles or not... is there a better way we can do this with less mess?

Source/Scene/ModelExperimental/CustomShader.js
ptrgags: https://github.com/CesiumGS/cesium/issues/9741 -- opened this issue for discussion

Source/Scene/ModelExperimental/UniformType.js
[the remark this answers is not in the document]
ptrgags: We talked about this with @lilleyse. Eventually these could even go in `Core` or somewhere else. for now though, it's okay to keep it in `ModelExperimental/`

Source/Scene/ModelExperimental/CustomShader.js
ptrgags: We talked about this on Friday. For now, I'm going to at least isolate this code into a `TextureManager` class. The caller is responsible for calling destroy() on the CustomShader (because it's hard to tell the intent if a custom shader is used for multiple tilesets)

Source/Scene/Cesium3DTileset.js
ptrgags: This is not necessary with the `TextureManager`, the implementation will only call update() once per frame. So the Model can always be responsible for updating the `CustomShader`

ptrgags:
Today I added a couple significant improvements:

* Custom shaders now support texture uniforms via `TextureUniform` as a value. a `TextureManager` handles the resources.
* If a primitive does not have all the attributes needed to satisfy the custom shader text, it will gracefully disable the custom shader.

I need to do another pass of updating the unit tests, these changes both added new files and changed the behavior of `CustomShaderStage`.

ptrgags:
@sanjeetsuhag I still have a few more unit tests to add and a development sandcastle, but otherwise this PR is ready for a review pass today.

Specs/Scene/ModelExperimental/ModelExperimentalPrimitiveSpec.js
```diff
@@ -40,6 +54,7 @@ describe("Scene/ModelExperimental/ModelExperimentalPrimitive", function () {
   it("configures the pipeline stages", function () {
```
ptrgags, line 54: Oh looks like I didn't add any tests for configuring the pipeline with custom shaders. I should do that.

Source/Scene/ModelExperimental/CustomShaderStage.js
```diff
@@ -0,0 +1,519 @@
...
+  return {
+    attributeField: generateStructField(glslType, attributeName),
+    value: value,
+  };
+}
+
+function generateVertexShaderLines(customShader, namedAttributes) {
+  var categories = partitionAttributes(
+    namedAttributes,
+    customShader._usedVariablesVertex.attributeSet
```
lilleyse, line 230: Should `_usedVariablesVertex` be a non-underscore property?

Source/Scene/ModelExperimental/CustomShaderStage.js
```diff
@@ -0,0 +1,519 @@
...
+  );
+}
+
+function generatePositionBuiltins(customShader) {
+  var fragmentInputFields = [];
+  var initializationLines = [];
+  var usedVariables = customShader._usedVariablesFragment.positionSet;
+
+  // Model space position is the same position as in the glTF accessor.
+  if ("positionMC" in usedVariables) {
```
lilleyse, line 310: I think this is the first time I've seen `in` instead of `hasOwnProperty`. In this case there should be no difference because the object doesn't have inherited properties.

Source/Scene/ModelExperimental/TextureManager.js
```diff
@@ -0,0 +1,170 @@
+import defined from "../../Core/defined.js";
```
lilleyse, line 1: Out of curiosity have you tried loading a KTX2 texture into a custom shader? See some examples in https://sandcastle.cesium.com/?src=Materials.html

Source/Scene/ModelExperimental/TextureManager.js
```diff
@@ -0,0 +1,170 @@
...
+  if (defined(typedArray)) {
+    texture = new Texture({
+      context: context,
+      source: {
+        arrayBufferView: typedArray,
+        width: textureUniform.width,
+        height: textureUniform.height,
+      },
+      sampler: sampler,
+    });
```
lilleyse, line 90: I think this should pass in `flipY: false`, assuming the bottom-left of the texture is the origin. The checkboard example in the custom shader sandcastle makes it a little hard to verify correctness.

And in case I missed it, the documentation should state the pixel order for typed array input, e.g. row major from bottom row to top row.

Specs/Scene/ModelExperimental/CustomShaderStageSpec.js
```diff
@@ -0,0 +1,804 @@
...
+  CustomShader,
+  CustomShaderStage,
+  LightingModel,
+  ModelAlphaOptions,
+  ModelLightingOptions,
+  ShaderBuilder,
+  UniformType,
+  VaryingType,
+  _shadersCustomShaderStageVS,
+  _shadersCustomShaderStageFS,
```
lilleyse, line 12: Huh this is the first place I've seen a spec include a shader... but I guess it works because I see it's handled in the gulpfile.

Specs/Scene/ModelExperimental/TextureManagerSpec.js
```diff
@@ -0,0 +1,127 @@
...
+  Resource,
+  TextureManager,
+  TextureUniform,
+  when,
+} from "../../../Source/Cesium.js";
+import createScene from "../../createScene.js";
+import pollToPromise from "../../pollToPromise.js";
+
+describe(
+  "Scene/ModelExperimental/TextureManager",
```
lilleyse, line 12: There should be a check that the previous texture is destroyed

lilleyse (commented):
A masterpiece @ptrgags! Can't wait to see what we build with this.

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -56,6 +57,7 @@ export default function ModelExperimental(options) {
 
   this._ready = false;
   this._readyPromise = when.defer();
+  this._customShader = options.customShader;
```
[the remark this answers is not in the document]
ptrgags: The application. Where's the best place to document this between the README and the `CustomShader` class?

Source/Scene/ModelExperimental/CustomShader.js
```diff
@@ -0,0 +1,256 @@
...
+ *   },
+ *   vertexShaderText: `
+ *   void vertexMain(VertexInput vsInput, inout vec3 position) {
+ *     v_selectedColor = mix(vsInput.attributes.color_0, vsInput.attributes.color_1, u_colorIndex);
+ *     position += 0.1 * vsInput.attributes.normal;
+ *   }
+ *   `,
+ *   fragmentShaderText: `
+ *   void fragmentMain(FragmentInput fsInput, inout czm_modelMaterial material) {
+ *     material.normal = texture2D(u_normalMap, fsInput.attributes.normal);
```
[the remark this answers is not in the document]
ptrgags: :open_mouth: yep, silly me. I should really run these examples in a Sandcastle to check them...

Source/Scene/ModelExperimental/CustomShader.js
```diff
@@ -0,0 +1,256 @@
...
+export default function CustomShader(options) {
+  options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+
+  this.mode = defaultValue(options.mode, CustomShaderMode.MODIFY_MATERIAL);
+  this.lightingModel = options.lightingModel;
+  this.uniforms = defaultValue(options.uniforms, defaultValue.EMPTY_OBJECT);
+  this.varyings = defaultValue(options.varyings, defaultValue.EMPTY_OBJECT);
+  this.vertexShaderText = options.vertexShaderText;
+  this.fragmentShaderText = options.fragmentShaderText;
+  this.isTranslucent = defaultValue(options.isTranslucent, false);
```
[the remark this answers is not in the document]
ptrgags: as of right now, you can't change these settings after the shader is created. Only the uniforms can be updated via `setUniform`

I can document them and list them as `@private` for now

Source/Scene/ModelExperimental/CustomShaderStage.js
```diff
@@ -0,0 +1,519 @@
...
+  // Fields for the Attribute struct. for example:
+  // "    vec3 position;"
+  return generateStructField(glslType, name);
+}
+
+// GLSL types of standard attribute types when uniquely defined
+var attributeTypeLUT = {
+  position: "vec3",
+  normal: "vec3",
+  tangent: "vec4",
```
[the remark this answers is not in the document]
ptrgags: Yes, it would appear as a vec4. I haven't worked with tangents in shaders before, what would you expect to see here? it seems that glTF allows for different orientations via the w channel, is that something typically needed even for other types of 3D models?

Okay you're the second one to bring this up. The thing though is I don't have a VertexAttributeSemantic, I have a string from the shader here. So I'd still need a LUT from "normal" -> "NORMAL" -> "vec3" here... what do we want to do?

Source/Scene/ModelExperimental/CustomShaderStage.js
ptrgags: Yeah I was wondering about if this was good style or not. I can switch to hasOwnProperty for consistency.

Source/Scene/ModelExperimental/CustomShaderStage.js
ptrgags: hm yeah I can change it to be a regular property that's marked `@private`

Source/Scene/ModelExperimental/TextureManager.js
ptrgags: ah I was wondering about this. Is this only for the typed array version of creating a Texture, or both?

Source/Scene/ModelExperimental/TextureManager.js
ptrgags: This is not currently supported to keep the scope of this PR focused on the essentials.

Though how important is supporting KTX2 here? can it wait until #9741?

Source/Scene/ModelExperimental/CustomShaderStage.js
```diff
@@ -0,0 +1,519 @@
...
+
+    var variableName;
+    if (defined(semantic)) {
+      variableName = VertexAttributeSemantic.getVariableName(
+        semantic,
+        setIndex
+      );
+    } else {
+      // Handle user defined vertex attributes. They must begin with an underscore
+      // For example, "_TEMPERATURE" will be converted to "a_temperature".
```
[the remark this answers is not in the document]
ptrgags: oh yeah you're right, the end result is used for both, with the `a_` prepended when needed.

Source/Scene/ModelExperimental/CustomShaderStage.js
```diff
@@ -0,0 +1,457 @@
...
+  tangent: "vec4(1.0, 0.0, 0.0, 1.0)",
+  texCoord: "vec2(0.0)",
+  joints: "vec4(0.0)",
+  weights: "vec4(0.0)",
+};
+
+function inferAttributeDefaults(attributeName) {
+  // remove trailing set indices. E.g. "texCoord_0" -> "texCoord"
+  var trimmed = attributeName.replace(/_[0-9]+$/, "");
+  var glslType = attributeTypeLUT[trimmed];
```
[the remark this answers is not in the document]
ptrgags: hm, maybe we have conflicting intents here, the `Attributes` struct so far has been the raw values from the glTF accessors, so far no preprocessing like this was done. Should it be this way, or should some things have more preprocessing like you're saying?

The position variables like `positionWC` right now are scoped to  the `FragmentInput` struct. 

@lilleyse should `Attributes` do more to preprocess things (which would require some special-case logic in CustomShaderStage) or do we want a separate version of tangents/colors that are more like the "derived" stuff?

Source/Scene/ModelExperimental/CustomShaderGuide/README.md
```diff
@@ -0,0 +1,270 @@
+# `CustomShader` Documentation
```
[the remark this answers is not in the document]
ptrgags: in the initial example I show how isTranslucent and set `material.alpha`

Source/Scene/ModelExperimental/TextureManager.js
lilleyse: Just for the typed array version. PNG and JPEG are top-left origin so they would use the default `flipY: true`

Source/Scene/ModelExperimental/CustomShaderStage.js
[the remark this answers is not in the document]
lilleyse: Let's discuss a bit offline.

Source/Scene/ModelExperimental/CustomShader.js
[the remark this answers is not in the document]
lilleyse: Sounds good

Source/Scene/ModelExperimental/ModelExperimental.js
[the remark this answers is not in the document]
lilleyse: Document in both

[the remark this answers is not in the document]
ptrgags: We talked about this on a call, the best place to document it in the code is the constructor for `CustomShader`

Source/Scene/ModelExperimental/CustomShaderStage.js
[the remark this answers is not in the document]
ptrgags: We decided that this is a larger issue beyond what we have time for this week. I'll write up an issue.

Source/Scene/ModelExperimental/TextureManager.js
ptrgags: We talked about this, we want to make this flipY change and also add one more sandcastle example with a gradient to check the flip-y

ptrgags: Just made this change and added a sandcastle demo. It's a little hard to tell from the model which way is up, but it does appear right-side-up in Spector:

image

Source/Scene/ModelExperimental/CustomShaderStage.js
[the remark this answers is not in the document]
ptrgags: Opened https://github.com/CesiumGS/cesium/issues/9771 which refers to this thread.

Source/Scene/ModelExperimental/CustomShaderStage.js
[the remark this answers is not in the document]
ptrgags: Opened https://github.com/CesiumGS/cesium/issues/9771 for now.

ptrgags:
@lilleyse updated!

Apps/Sandcastle/gallery/development/Custom Shaders ModelExperimental.html
```diff
@@ -201,6 +201,52 @@
...
+            for (var j = 0; j < size; j++) {
+              var index = i * size + j;
+              // red increases in the +x direction (to the right)
+              typedArray[3 * index + 0] = j;
+              // Green increases in the +y direction (from bottom to top)
+              typedArray[3 * index + 1] = i;
+              // blue is 0 so it is omitted.
+            }
+          }
+          console.log(typedArray);
```
lilleyse, line 223: Remove stray `console.log`

ptrgags: oops my bad

Specs/Scene/ModelExperimental/CustomShaderSpec.js
```diff
@@ -220,6 +220,17 @@ describe("Scene/ModelExperimental/CustomShader", function () {
...
       });
 
+      var shaders = [];
+      afterEach(function () {
+        for (var i = 0; i < shaders.length; i++) {
+          var shader = shaders[i];
+          if (!shader.isDestroyed()) {
+            shader.destroy();
+          }
+          shaders.length = 0;
```
lilleyse, line 230: Should `shaders.length = 0;` be outside the loop? Same comment for `textureManagers` below.

ptrgags:
@lilleyse Travis passed!

Issue #9710, opened 2021-08-04
glTFLoader spec issue when releasing CesiumJS 1.84


srothst1:
The following specs failed when I ran them on my Linux machine yesterday afternoon during the CesiumJS 1.84 release. @ebogo1 was not able to reproduce these spec failures. Let me know if you would like more details - I can't think of much more to add here. @ebogo1 feel free to chime in with any more information that you have collected.

[final.pdf](https://github.com/CesiumGS/cesium/files/6927819/final.pdf)

ebogo1:
One thing to note is that @srothst1 was able to reproduce these failures on the 1.83 release too, so this was not a recent regression.

lilleyse:
The relevant failures are:

```
Scene/GltfLoader resolves before textures are loaded when incrementallyLoadTextures is true
Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. in http://localhost:8081/ThirdParty/jasmine-2.2.0/jasmine.js (line 1803)
attemptAsync/timeoutId<@http://localhost:8081/ThirdParty/jasmine-2.2.0/jasmine.js:1803:23

Scene/GltfLoader destroys glTF loader
Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. in http://localhost:8081/ThirdParty/jasmine-2.2.0/jasmine.js (line 1803)
attemptAsync/timeoutId<@http://localhost:8081/ThirdParty/jasmine-2.2.0/jasmine.js:1803:23
```

lilleyse:
From https://github.com/CesiumGS/cesium/pull/9712#issuecomment-897176570

> @lilleyse hm that is the same type of error that we're seeing. Our unit tests were modeled after the GltfLoader one so whatever the fix there is, it's probably applicable here too... We should discuss this tomorrow


> @lilleyse I see that `waitForLoaderProcess` only waits for the `promise` and not the `texturesLoadedPromise`. Do you think that's playing a part here?
> 
> https://github.com/CesiumGS/cesium/blob/ffcc025a9ed8c9f54e7d7ce6cb09bf89bcdac273/Specs/waitForLoaderProcess.js#L3-L14
>

PR #9727, opened 2021-08-12
Fix glTF loader test timeouts

lilleyse:
Fixes https://github.com/CesiumGS/cesium/issues/9710

Certain glTF loader tests were timing out because the `JobScheduler` was was getting clogged after repeatedly loading GPU resources and not getting reset. This was a bug in the testing code only.

Normally scene is responsible for resetting the job scheduler every frame but since we're not calling `scene.renderForSpecs` we need to reset budgets manually. This is only required for loaders that use the job scheduler like `GltfVertexBufferLoader`, `GltfIndexBufferLoader`, and `GltfTextureLoader`

I can pretty frequently reproduce the timeouts in Firefox on Linux with cache disabled (this seems to indirectly effect the speed at which the JobScheduler fills up), but it may vary by device.

### DIFF
--- a/Specs/Scene/GltfDracoLoaderSpec.js
+++ b/Specs/Scene/GltfDracoLoaderSpec.js
@@ -9,6 +9,7 @@ import {
   when,
 } from "../../Source/Cesium.js";
 import createScene from "../createScene.js";
+import loaderProcess from "../loaderProcess.js";
 import pollToPromise from "../pollToPromise.js";
 import waitForLoaderProcess from "../waitForLoaderProcess.js";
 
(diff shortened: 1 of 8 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags (commented):
Code looks sensible to me, I'm going to try running this a few different ways to confirm that I don't get any timeouts

ptrgags:
@lilleyse This looks good and should solve all the timeout issues @sanjeetsuhag and I were having this week. Thanks!

PR #9729, opened 2021-08-13
Add ability to specify Bing Maps geocoder culture

mramato:
1. Add optional `culture` property to BingMapsGeocoderService constructor and use it to set culture when making a request. This causes the results to be displayed in the specified culture and language.

2. Update Resource.js to use CesiumMath's RNG so we can have a random seed with consistent results

3. Slightly improve BingMapsGeocoderService tests to support the new functionality and double check some existing code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 # Change Log
 
-### 1.85 - 2021-08-10
+### 1.85 - 2021-09-01
+
+##### Additions :tada:
+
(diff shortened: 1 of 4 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

mramato:
Found several issues (#9730, #9731) while figuring out why CI was failing. Ended up tweaking the tests to compensate but those issues are likely to bite us again if we don't address them at some point.

ebogo1:
Thanks @mramato! I confirmed this works in Sandcaslte and that specs pass. The code changes look good to me. I pushed a fix to the link in the new docs but otherwise this is good to go.

PR #9706, opened 2021-08-01
Start moving third party libraries to use npm

ebogo1:
This is a chunk of progress towards https://github.com/CesiumGS/cesium/issues/9473. 

Comparing the generated .zip files with the one in main after https://github.com/CesiumGS/cesium/pull/9705 gets merged:
- 64.1 MB in this branch vs 63.6 MB in main
- In Build/Cesium/:
  - `Cesium.js` is 3.9 MB in this branch vs 3.7 MB in main (uncompressed)
  - ThirdParty/Workers/ now only has basis and draco files (we removed deflate/inflate.js and ktx-parse-modern.js)
- In Source/ThirdParty/:
  - This folder is 3.6 MB up from 3.1 MB in main (uncompressed)
  - The new files generated from npm modules seem correct; they share the same folder as the third party files currently in main

Comparing the .tgz with main:
- 22.0 MB here vs 21.4 MB in main
- `decodeGoogleEarthEnterprisePacket` is about 1.5x as big in this branch; probably because we're importing more than we need from `pako`..

Issue #9473, opened 2021-04-07
Replace third party libraries with npm modules


ebogo1:
This issue will serve as a task list for replacing third party modules with npm modules. Here is the task list (incomplete for now; I'll be updating this as we go).

- [x] Create staging branch for npm changes - https://github.com/CesiumGS/cesium/tree/npm-third-party-staging
- [x] Merge first pass PR into staging (https://github.com/CesiumGS/cesium/pull/9466/)
- [x] Move to npm; or other - see https://github.com/CesiumGS/engineering/issues/69#issue-834201104)
    - [x] `sprintf` (https://github.com/CesiumGS/cesium/pull/9476)
    - [x] bitmap-sdf.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [ ] ~google-earth-dbroot-parser.js~ leaving in ThirdParty, at least for now
    - [x] graphmesplitter.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [x] jsep.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [ ] ~knockout-3.5.1.js, knockout-es5.js, knockout.js~ (need longer term plan to get rid of knockout)
    - [x] LercDecode.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [x] measureText.js (https://github.com/CesiumGS/cesium/pull/9765)
    - [x] mersenne-twister.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [x] NoSleep.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [x] pako_inflate.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [x] protobuf-minimal.js (https://github.com/CesiumGS/cesium/pull/9747)
    - [x] purify.js (https://github.com/CesiumGS/cesium/pull/9688)
    - [x] Uri.js (https://github.com/CesiumGS/cesium/pull/9740)
    - [x] zip.js (https://github.com/CesiumGS/cesium/pull/9696)
- [ ] Other files in Source/ThirdParty:
    - [x] GltfPipeline (https://github.com/CesiumGS/cesium/pull/9699)
        - [x] Follow-up to update instructions in https://github.com/CesiumGS/gltf-pipeline README.md
    - [ ] Basis: Workers/basis_transcoder.js & basis_transcoder.wasm
[...]

PR #9740, opened 2021-08-20
Replace Uri.js with urijs npm module

ebogo1:
This should not be merged until https://github.com/CesiumGS/cesium/pull/9706 is in `main`.

Part of https://github.com/CesiumGS/cesium/issues/9473. This code touches a lot of important files and there were a few things that came up; see my review comments.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Core/Resource.js
```diff
@@ -653,9 +653,15 @@ Resource.prototype.getDerivedResource = function (options) {
     parseQuery(uri, resource, true, preserveQueryParameters);
 
     // Remove the fragment as it's not sent with a request
-    uri.fragment = undefined;
+    uri.fragment("");
 
-    resource._url = uri.resolve(new Uri(getAbsoluteUri(this._url))).toString();
+    if (uri.is("urn")) {
```
ebogo1, line 658: `URI.absoluteTo()` throws an exception when trying to concatenate URNs, whereas Uri.js returned the unchanged URN. Is this the correct behavior for `Resource`? There are a couple other places where this came up.

Source/Core/getAbsoluteUri.js
```diff
@@ -38,8 +38,13 @@ getAbsoluteUri._implementation = function (relative, base, documentObject) {
...
-  var baseUri = new Uri(base);
-  var relativeUri = new Uri(relative);
-  return relativeUri.resolve(baseUri).toString();
+  var relativeUri = new URI(relative);
+  if (relativeUri.is("urn")) {
+    return relativeUri.toString();
+  }
+  var url = relativeUri.absoluteTo(base).toString();
+  // URI.absoluteTo() escapes the placeholders. Undo that.
+  url = url.replace(/%7B/g, "{").replace(/%7D/g, "}");
```
ebogo1, line 47: This code was copied from `Resource.prototype.getUrlComponent()`; it's needed because the `urijs` implementation of `absoluteTo` normalizes URLs (this is also consistent with the `URL(relative, base)` constructor in JS. Is this logic safe here?

Specs/Core/buildModuleUrlSpec.js
```diff
@@ -1,13 +1,15 @@
...
+import { urijs as URI } from "../../Source/Cesium.js";
 
 describe("Core/buildModuleUrl", function () {
   it("produces an absolute URL for a module", function () {
     var url = buildModuleUrl("Workers/transferTypedArrayTest.js");
 
     expect(url).toMatch(/Workers\/transferTypedArrayTest.js$/);
-    expect(new Uri(url).isAbsolute()).toBe(true);
+    var uri = new URI(url);
+    expect(uri.scheme().length).toBeGreaterThan(0);
```
ebogo1, line 11: `urijs` does not have an equivalent `isAbsolute()` function so I checked the scheme and fragment manually per the old Uri.js implementation. This change happens again in `TaskProcessor`.

Specs/DataSources/KmlDataSourceSpec.js
```diff
@@ -1344,7 +1344,7 @@ describe("DataSources/KmlDataSource", function () {
       var billboard = entities[0].billboard;
       expect(billboard.scale.getValue()).toEqual(2.0);
       expect(billboard.rotation.getValue()).toEqual(CesiumMath.toRadians(-4.0));
-      expect(billboard.image.getValue().url).toEqual("http://test.invalid");
+      expect(billboard.image.getValue().url).toEqual("http://test.invalid/");
```
ebogo1, line 1347: These three specs failed because of the trailing `/` character. I updated the specs to pass, but is it still correct if the image URLs have trailing slashes? If not I can look into where the slash is being added.

ebogo1 (commented):
For reference, the [urijs docs](http://medialize.github.io/URI.js/docs.html).

Source/Scene/CreditDisplay.js
```diff
@@ -521,8 +521,8 @@ function getDefaultCredit() {
       logo.indexOf("https://") !== 0 &&
       logo.indexOf("data:") !== 0
     ) {
-      var logoUrl = new Uri(logo);
-      logo = logoUrl.getPath();
+      var logoUrl = new URI(logo);
```
ebogo1, line 524: This code is never reached by specs..

mramato:
Straight from their [README](https://github.com/medialize/URI.js)

> IMPORTANT: You may not need URI.js anymore! Modern browsers provide the URL and URLSearchParams interfaces.

What missing features are forcing us to use urljs (or any third party for that matter) instead of just using URL and URLSearchParams?

mramato:
@ebogo1 you please merge in `npm-third-party-staging` (or main if that gets merged before you get to this)

ebogo1:
> What missing features are forcing us to use urljs

@mramato URL expects a valid web URL and throws a type error for things like filepaths - [URL docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).

mramato:
> @mramato URL expects a valid web URL and throws a type error for things like filepaths - [URL docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL).

I'm not sure what you mean. `const url = new URL("file://users/mamato/test.gltf");` works fine.

ebogo1:
@mramato `TaskProcessor` is one place that comes to mind; something like `new URL("Workers/worker.js")` will throw.

mramato:
> `TaskProcessor` is one place that comes to mind; something like `new URL("Workers/worker.js")` will throw. M

That doesn't sound quite right.

> Maybe less important but URL also does not have a `normalize` method..

MDN specifically mentioned you can just use the `href` property to normalize the URL.

All this being said, if this is a stop gap and we plan on taking a serious look into using stock URL post release, then I won't hold things up.  It could also affect things like Node.js support. But in general it sounds like we assumed too much or gave up to early in trying to use the standard.

Specs/Core/ResourceSpec.js
```diff
@@ -6,7 +6,7 @@ import { RequestErrorEvent } from "../../Source/Cesium.js";
 import { RequestScheduler } from "../../Source/Cesium.js";
 import { Resource } from "../../Source/Cesium.js";
 import createCanvas from "../createCanvas.js";
-import { Uri } from "../../Source/Cesium.js";
+import { urijs as URI } from "../../Source/Cesium.js";
```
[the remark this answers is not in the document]
mramato: `Uri` makes more sense since it will significantly reduce the overall code diff.

Source/Core/Resource.js
mramato: Where exactly are the `urn` urls coming from?

Source/Core/getAbsoluteUri.js
mramato: It almost feels like this fix is in the wrong place, like the calling code is the thing that should be handling `{` or perhaps a specialized function like `getAbsoluteTemplateStringUri.js`  Feels weird to just have something as generic as `getAbsoluteUri` embed code like this.

Source/Core/Resource.js
ebogo1: It's possible in .gltf files, i.e. the `images` in https://github.com/CesiumGS/cesium/blob/main/Specs/Data/Models/GltfLoader/BoxTextured/glTF-Embedded/BoxTextured.gltf, used in -https://github.com/CesiumGS/cesium/blob/1837859300e61ae2ea1bc0e3832923f87e752fed/Specs/Scene/GltfLoaderSpec.js#L48.

Not sure how common this is..

Source/Core/getAbsoluteUri.js
ebogo1: I moved the string replace to the spec in question. Is there no risk of of this mattering at runtime because browsers treat encoded symbols the same as unencoded ones?

Source/Core/Resource.js
lilleyse: I thought data uris were different than urn?

According to https://stackoverflow.com/a/28865747 urn's start with `urn:`

Source/Core/getAbsoluteUri.js
mramato: I didn't realize the specs were the only place this happens.. did you verify any sandcastles using template uris for imager work?

Source/Core/Resource.js
mramato: Yeah, there's definitely confusion here.  almost everything is an urn.. so I'm not sure what differentiates from data uri or other in `is`

ebogo1: Ah that's my mistake.. I misunderstood the terminology. But, now I realize there's a cleaner fix - checking whether the url's `scheme === "data"` should do the trick and is more readable I think.

mramato: Yeah, that's a check we do in a bunch of places (we even have an `isDataUri` helper function for it)

Source/Core/getAbsoluteUri.js
ebogo1: I verified that imagery Sandcastles work as expected, though there are bunch of `Failed to obtain image tile X ... Y: ... Level: ...` in the console.. this happens in the same way in `main`, doesn't seem like we have an open issue for it?

As far as I can tell only the spec calls `getAbsoluteUri` with a templated url - by the time the imagery urls make their way to `fetchImage` the template values are already string replaced correctly.

Source/Core/Resource.js
ebogo1: On second thought, I don't think just checking for data uris is enough - `ModelOutlineLoaderSpec` catches cases in Resource.getDerivedResource where the scheme is `"blob"`. The third party Uri.js in `main` ignores the base URL's properties (including scheme) if the relative URL has them set: 

https://github.com/CesiumGS/cesium/blob/1837859300e61ae2ea1bc0e3832923f87e752fed/Source/ThirdParty/Uri.js#L177-L182

Sandcastles and specs pass if everywhere that (incorrectly) checks `Uri.is("urn")` is changed to `Uri.scheme() === ""`. This happens in `RequestScheduler.getServerKey`, `Resource.getDerivedResource`, and `getAbsoluteUri`. Is checking just for data schemes as opposed to checking for any scheme before calling `Uri.absoluteTo` preferable anywhere? Checking for any scheme at all is more faithful to the behavior in `main`.

mramato: So what is the behavior for these urls in `main`?  Do we just do nothing and return the original url?  If so, then we can do the same thing here for now, but it could ultimately be masking a bug. Like should we be trying to absolutize or otherwise manipulate a opaque URL at all?

mramato: And wouldn't `is("relative")` be what we would want to check instead?

mramato: It's basically the same check under the hood, but much clearer that we are only doing certain things for relative paths

ebogo1: > Do we just do nothing and return the original url?

Yes, this is what the current third party Uri.js does.

> And wouldn't is("relative") be what we would want to check instead?

Not quite, since doing so falsely flags urls like `"test.png"`, which causes spec failures. In main, `new Uri("test.png").resolve(new Uri("http://localhost:8080/Specs/SpecRunner.html"))` turns into `"http://localhost:8080/Specs/test.png"`. This is consistent with this npm module's `absoluteTo`.

ebogo1:
>  Checking for any scheme at all is more faithful to the behavior in `main`.

Pushed a change in ca6a116 to replace `is("urn")` with `scheme() !== ""`.

PR #9745, opened 2021-08-22
Scene: change options to required

puxiao:
Even though i know :

1. Generally, a scene is not created directly.

2. in the Scene.js：

   ```
   function Scene(options) {
     ...
   }
   ```

<br>

*Cesium.d.ts*

```
export class Scene {
    constructor(options?: {
        canvas: HTMLCanvasElement;
        ...
    )
}
```

<br>

But if you really do

```
var scene = new Cesium.Scene();
```

Throws:
DeveloperError : options and options.canvas are required.

<br>

Emm...

I think that is not a good type description.

So:

```diff
- * @param {Object} [options] Object with the following properties:
+ * @param {Object} options Object with the following properties:
```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -294,3 +294,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Ethan Wong](https://github.com/GetToSet)
 - [Calogero Mauceri](https://github.com/kalosma)
 - [Ren Jianqiang](https://github.com/renjianqiang)
+- [Yang Puxiao](https://github.com/puxiao)
--- a/Source/Scene/Scene.js
+++ b/Source/Scene/Scene.js
@@ -124,7 +124,7 @@ var requestRenderAfterFrame = function (scene) {
  * @alias Scene
  * @constructor
  *
- * @param {Object} [options] Object with the following properties:
+ * @param {Object} options Object with the following properties:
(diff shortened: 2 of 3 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ebogo1:
@puxiao Thanks for the pull request. I think this is a fine change, especially since `options.canvas` was already required per the JS docs. It looks like the Cesium build fails because we now have a broken Scene constructor call in `Specs/Typescript/index.ts` -
https://github.com/CesiumGS/cesium/blob/61876fb494c0783a6f688937b02504ec2032904d/Specs/TypeScript/index.ts#L157.

I wonder if the best fix is to do something like:
```
let canvasElement = document.createElement("canvas");
dataSource = new KmlDataSource({
  canvas: canvasElement,
  camera: new Camera(new Scene({ canvas: canvasElement })),
});
```
@mramato does that seem fine here?

mramato:
@ebogo1 yes, if that gets it to bass then you're golden. This was just an oversight when we originally wrote that smokescreen (probably because the TS definition was wrong to begin with!)  Thanks @puxiao!

puxiao:
What should I do next ?
Go to modify `Specs/TypeScript/index.ts` ?

ebogo1:
@puxiao Yes please; I believe the fix I suggested will fix the build issues. Thanks.

ebogo1:
Thanks @puxiao - this looks good to me!

PR #9747, opened 2021-08-22
Use protobufjs from npm

mramato:
Locked to version 6.7.0 since that is the version we are already using and updating it requires additional work because of the use of `util.lazyResolve` in the generated google-earth-dbroot-parser.js code.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

mramato:
Unlike the other libraries, we had to include the existing compiled prototobuf dist files we were already using, rather than rely on npm package resolution rules.. i.e. we do:

```js
import * as protobuf from 'protobufjs/dist/minimal/protobuf.js';
```

This is because there are circular dependencies in the npm version which, when built, lead to undefined variables/functions that cause exceptions on load.

Once I got it building, tests were still failing.  Debugging the `GoogleEarthEnterpriseMetadata` with all exceptions turned on pointed to `google-earth-dbroot-parser.js` trying to use `util.lazyResolve` from protobuf.  It turns out that the latest version, 6.11.0, does not have this function so I locked it to the known working version we are using in main.

ThirdParty/npm/protobufjs.js
```diff
@@ -0,0 +1,2 @@
+import * as protobuf from 'protobufjs/dist/minimal/protobuf.js';
```
ebogo1, line 1: Is there a reason `import * as protouf from...` instead of `import protobuf...` here? Both seem to work with specs in my testing.

ebogo1 (commented):
Thanks @mramato. This looks good to me, just one question I had about the protobuf import.

I'll wait to merge this until `npm-third-party-staging` is in main.

ThirdParty/npm/protobufjs.js
mramato: The `import *` means give me everything where as `import protobuf` means "give me the default export only". In this case I'm pretty sure they are equivalent and it doesn't really matter.

mramato: Also, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import is a great reference for imports in general.

ebogo1:
Thanks @mramato. As a bonus, this shaved 0.2 MB off Cesium.js.map :tada:

PR #9765, opened 2021-08-29
Rewrite measureText in `writeTextToCanvas`

ebogo1:
Part of https://github.com/CesiumGS/cesium/issues/9473. 

This PR rewrites the code from Source/ThirdParty/measureText as a helper function in `writeTextToCanvas`. The point of the helper function is to extend `CanvasRenderingContext2D.measureText()` with `minx` and `height` properties (and truncate bounding box ascent/descent while we're there). 

The code is very inspired by the original third party function, but I believe stock ascent/descent values are good for what we're doing and I also think the loop to find `minx` was wrong in the third party version. We lose support for `writeTextToCanvas` on Firefox for Android though:

image

@lilleyse Let me know what you think - if it's too close to the original I can rework it some more.

Tested the Labels Sandcastle on:
- Linux Firefox, Chrome
- macOS Chrome, Safari


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/measure-text/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ebogo1:
Noticing a bunch of 
`Unknown mime type for cesium/measure-text/Build/Documentation/...` in the CI logs; should `measureText()` be marked @private to avoid this?

ebogo1:
There are definitely discrepancies between different browsers (noticeable with the "Offset label by distance" option in the [Labels Sandcastle](https://sandcastle.cesium.com/?src=Labels.html) -

Linux Firefox on this branch:
image

Linux Firefox on `main`
image


Linux Chrome on this branch: 
image

Linux Chrome on `main`:
image

lilleyse:
Can this be simplified further by using `actualBoundingBoxLeft` and `actualBoundingBoxRight`? Ideally we wouldn't need to draw the text to a canvas at all to determine its bounding box.

Here's another relevant StackOverflow issue: https://stackoverflow.com/questions/60347194/how-to-fit-text-to-a-precise-width-on-html-canvas. Assuming we still need a fallback for browsers that don't support `actualBoundingBox****`, including Firefox for Mobile and browsers that haven't been updated since early 2020, how do those solutions compare to the one here?

Maybe it's my lack of Canvas2D experience showing, but the overall approach taken in `writeTextToCanvas` feels pretty awkward. Is it possible for a canvas to "auto-expand" when content is written to it instead of trying to determine the perfect width/height upfront? The code was written 9 years ago. Maybe things have changed?

ebogo1:
> Can this be simplified further by using actualBoundingBoxLeft and actualBoundingBoxRight? Ideally we wouldn't need to draw the text to a canvas at all to determine its bounding box.

Good call; seems to work fine with `minx = actualBoundingBoxLeft | 0`.

> how do those solutions compare to the one here?

It's written differently but the idea is the same; iterating one pixel at a time in the direction of that's being checked (i.e. up/down/left/right).

It looks like the two options are a tradeoff: either we have a cleaner implementation but have inconsistent results in different browsers (and loose some browser support), or have the messier "brute force" implementation but get consistent browser results. If we use the brute force method as a fallback for unsupported browsers, there's also the question of whether the inconsistent results are acceptable; to me they look a bit bad in the above screenshots. I think if there's a fallback is in the code at all we may as well use it all the time for consistency.

ebogo1:
>  I think if there's a fallback in the code at all we may as well use it all the time for consistency.

Talked with @lilleyse offline and we agreed. Since the browser implementations are inconsistent we'll use the pixel counting method for now (opened https://github.com/CesiumGS/cesium/issues/9767).

Source/Core/writeTextToCanvas.js
```diff
@@ -39,8 +39,27 @@ function measureText(context2D, textString, font, stroke, fill) {
     // Context image data has width * height * 4 elements, because
     // each pixel's R, G, B and A are consecutive values in the array.
     var pixelData = ctx.getImageData(0, 0, width, height).data;
+    var length = pixelData.length;
     var width4 = width * 4;
     var i, j;
+
+    var ascent, descent;
+    // Find the number of rows (from the top) until the first non-white pixel
```
ebogo1, line 47: Had to expand these because eslint doesn't like empty code blocks in loops..

ebogo1:
looks like CI failed due to the https://github.com/CesiumGS/cesium/issues/9769 timeouts..

Source/Core/writeTextToCanvas.js
```diff
@@ -1,9 +1,95 @@
-import measureText from "../ThirdParty/measureText.js";
 import Color from "./Color.js";
 import defaultValue from "./defaultValue.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 
+function measureText(context2D, textString, font, stroke, fill) {
+  var metrics = context2D.measureText(textString);
```
lilleyse, line 7: Let's avoid settings our own properties on the `TextMetrics` object like the previous implementation did. Instead create a separate object with the properties `width`, `height`, `ascent`, `descent`, `minx` and return that.

Source/Core/writeTextToCanvas.js
```diff
@@ -1,9 +1,95 @@
...
+      for (j = 0; j < height; ++j) {
+        if (
+          pixelData[i * 4 + j * width4] !== 255 ||
+          pixelData[i * 4 + j * width4 + 1] !== 255 ||
+          pixelData[i * 4 + j * width4 + 2] !== 255 ||
+          pixelData[i * 4 + j * width4 + 3] !== 255
+        ) {
+          minx = i;
+          break;
+        }
```
lilleyse, line 75: Factor out the index into its own variable since it's used multiple times

```suggestion
        var pixelIndex = i * 4 + j * width4;
        if (
          pixelData[pixelIndex] !== 255 ||
          pixelData[pixelIndex + 1] !== 255 ||
          pixelData[pixelIndex + 2] !== 255 ||
          pixelData[pixelIndex + 3] !== 255
        ) {
          minx = i;
          break;
        }
```

Source/Core/writeTextToCanvas.js
```diff
@@ -1,9 +1,95 @@
...
+    // Find the number of rows (from the top) until the first non-white pixel
+    for (i = 0; i < length; ++i) {
+      if (pixelData[i] !== 255) {
+        ascent = (i / width4) | 0;
+        break;
+      }
+    }
+
+    // Find the number of rows (from the bottom) until the first non-white pixel
+    for (i = length - 1; i > 0; --i) {
```
lilleyse, line 56: Should it be `>=`?

```suggestion
    for (i = length - 1; i >= 0; --i) {
```

Source/Core/writeTextToCanvas.js
```diff
@@ -72,32 +157,28 @@ function writeTextToCanvas(text, options) {
   context2D.lineWidth = strokeWidth;
   context2D[imageSmoothingEnabledName] = false;
 
-  // textBaseline needs to be set before the measureText call. It won't work otherwise.
-  // It's magic.
-  context2D.textBaseline = defaultValue(options.textBaseline, "bottom");
```
lilleyse, line 77: Is this no longer relevant?

Source/Core/writeTextToCanvas.js
```diff
@@ -72,32 +157,28 @@ function writeTextToCanvas(text, options) {
...
-
   // in order for measureText to calculate style, the canvas has to be
   // (temporarily) added to the DOM.
   canvas.style.visibility = "hidden";
   document.body.appendChild(canvas);
 
-  var dimensions = measureText(context2D, text, stroke, fill);
+  var dimensions = measureText(context2D, text, font, stroke, fill);
   canvas.dimensions = dimensions;
```
lilleyse, line 166: Can `canvas.dimensions = dimensions;` be removed? `HtmlCanvasElement` doesn't have a `dimensions` property and it wouldn't make sense for it to be set to a `TextMetrics` object anyways.

lilleyse: Never mind, I see that this property is used by `LabelCollection`. Could you add a comment about that? Otherwise others might get confused.

lilleyse:
Labels, Map Pins, and Clustering sandcastles look identical to main. The glyph spacing still leaves something to be desired, especially on Firefox, but that's out of scope for this PR.

Source/Core/writeTextToCanvas.js
ebogo1: I'm not sure why it was - the `measureText` function does not use the context's baseline anywhere. Maybe 9 years ago it worked differently? I didn't notice any changes when I removed this line.

ebogo1:
@lilleyse Should be good for final review. I pushed a small commit to the gulpfile now that there's no third party measureText file.

PR #9776, opened 2021-08-31
Zoom out the camera to fix a log depth bug in ModelExperimental tests

ptrgags:
This branch makes a small tweak to one test to ensure it runs even in environments where log depth is disabled. We have internal code where this becomes a problem. 

To test this, you need to change `Scene.js` line 214 so that `this._logDepthBuffer = false`, not `context.fragmentDepth`. This will cause the test to fail because the camera zooms in so far it begins to clip:

### DIFF
--- a/Specs/Scene/ModelExperimental/ModelExperimentalSpec.js
+++ b/Specs/Scene/ModelExperimental/ModelExperimentalSpec.js
@@ -6,6 +6,7 @@ import {
   ModelExperimental,
   Cartesian3,
   defined,
+  HeadingPitchRange,
   when,
   ShaderProgram,
 } from "../../../Source/Cesium.js";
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-model-experimental-pick-log-depth/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag (approved):
This change looks good. I'm sure it will help when we add more picking tests. Thanks @ptrgags!

lilleyse:
For more background, this isn't really a log depth bug, rather when log depth is enabled it uses a 0.1 meter near plane and when disabled it uses a 1 meter near plane. The test was zooming close enough to the model that it started to clip through the near plane and not pick the model when log depth was disabled.

PR #9779, opened 2021-09-01
Fix regressions in Billboard properties

ptrgags:
Fixes some regressions that were causing the [Clamp to Terrain Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Clamp%20to%20Terrain.html&label=All) to break when you select "Draw Billboard" 

There weren't any specs for Billboard, though hm... I should check BillboardCollectionSpec too.

### DIFF
--- a/Specs/Scene/BillboardCollectionSpec.js
+++ b/Specs/Scene/BillboardCollectionSpec.js
@@ -241,6 +241,193 @@ describe(
       expect(b.disableDepthTestDistance).toEqual(10.0);
     });
 
+    it("required properties throw for undefined", function () {
+      var b = billboards.add();
+      b.show = false;
+      b.position = new Cartesian3(1.0, 2.0, 3.0);
+      b.pixelOffset = new Cartesian2(1.0, 2.0);
+      b.eyeOffset = new Cartesian3(1.0, 2.0, 3.0);
+      b.horizontalOrigin = HorizontalOrigin.LEFT;
+      b.verticalOrigin = VerticalOrigin.BOTTOM;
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-billboard-depth-test-dist/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/BillboardCollectionSpec.js
```diff
@@ -241,6 +241,170 @@ describe(
...
+        b.distanceDisplayCondition = 10;
+      }).toThrowDeveloperError();
+      expect(function () {
+        b.disableDepthTestDistance = "far";
+      }).toThrowDeveloperError();
+    });
+
+    it("image property setter creates image with GUID for non-uris", function () {
+      var b = billboards.add();
+      b.image = 42;
```
ptrgags, line 402: @lilleyse this setter is weird, if the value is defined and isn't a String, Resource, Image, etc, it creates a new image with a GUID... how is this supposed to be used?

ptrgags:
@lilleyse updated!

Specs/Scene/BillboardCollectionSpec.js
lilleyse: I'm not completely sure, but my best guess is it has something to do with `LabelCollection`

lilleyse:
Looks good, will merge when CI passes

PR #9794, opened 2021-09-07
Fix `Viewer.bottomContainer` type in JSDoc/TypeScript

puxiao:
*Viewer.js*

```
  // Bottom container
  var bottomContainer = document.createElement("div");
  bottomContainer.className = "cesium-viewer-bottom";

...

  /**
   * Gets the DOM element for the area at the bottom of the window containing the
   * {@link CreditDisplay} and potentially other things.
   * @memberof Viewer.prototype
   * @type {Element}
   * @readonly
   */
  bottomContainer: {
    get: function () {
      return this._bottomContainer;
    },
  },
```

<br>

If use TypeScript: 

```
viewer.bottomContainer.style.backgroundColor = 'blue';
```

<br>

:exclamation: ​PROBLEMS: 

```
Property 'style' does not exist on type 'Element'.
```

<br>

```diff
- viewer.bottomContainer.style.backgroundColor = 'blue';
+ (viewer.bottomContainer as HTMLElement).style.backgroundColor = 'blue';
```

<br>

<br>

**In fact, I think all `Element` should be changed to `HTMLElement` in Cesium.js.**

<br>

### DIFF
--- a/Source/Widgets/Viewer/Viewer.js
+++ b/Source/Widgets/Viewer/Viewer.js
@@ -947,7 +947,7 @@ Object.defineProperties(Viewer.prototype, {
    * Gets the DOM element for the area at the bottom of the window containing the
    * {@link CreditDisplay} and potentially other things.
    * @memberof Viewer.prototype
-   * @type {Element}
+   * @type {HTMLDivElement}
    * @readonly
    */
   bottomContainer: {


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/patch-2/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Widgets/Viewer/Viewer.js
```diff
@@ -947,7 +947,7 @@ Object.defineProperties(Viewer.prototype, {
    * Gets the DOM element for the area at the bottom of the window containing the
    * {@link CreditDisplay} and potentially other things.
    * @memberof Viewer.prototype
-   * @type {Element}
+   * @type {HTMLDivElement}
```
ebogo1, line 950: Is there reason to use `HTMLDivElement` instead of `HTMLElement`? `HTMLElement` is more generic and has the style property as well.

puxiao: From the actual code, it is indeed an HTMLDivElement.

bottomContainer



But if you want to define it as HtmlElement, it doesn't seem to be a problem.

You decide.

puxiao: I see that viewer.infoBox.container is currently also an Element. 
If the user does not define it, it will be created as HTMLDivElement internally.

At least, I think all Element should be replaced with HTMLElement.

cesium-concierge:
Thanks again for your contribution @puxiao! 

ggetz:
@ebogo1 Would you be able to finish up the review here?

ebogo1:
@puxiao Thanks for the pull request!
> In fact, I think all Element should be changed to HTMLElement in Cesium.js.

I'm not sure if this is applicable to all Element types but there are definitely other parts of JSDoc that have the same error; for example `InspectorShared` has a couple functions that return type `Element` even though the object is created with `document.createElement("div")` - 
https://github.com/CesiumGS/cesium/blob/9a7088e9891b989b99afc9da49cdbceaff3b5bc5/Source/Widgets/InspectorShared.js#L10-L17

If you're up for it, doing another pass for places where the change to a stricter JSDoc type makes sense would be great. I think if we make this change for `Viewer.bottomContainer` we should also do it everywhere else where it's correct. No worries if you're not - we can revisit this on our own later!

puxiao:
I hope one day, cesium.js will have its own independent TypeScript types, just like [three-ts-types](https://github.com/three-types/three-ts-types) for "three.js".

PR #9797, opened 2021-09-07
Reverse axis order in WMS 1.3.0 for certain EPSGs

tantricllama:
As described in the Axis Ordering section in the link below, some EPSGs define a reversed axis order in WMS 1.3.0. Currently in Cesium, this requires specifying the GetMap URL with a bbox override.
https://docs.geoserver.org/latest/en/user/services/wms/basics.html

The fix is based on the one specified in the community forum link below, with some additional handling for exceptions. The list of exceptions is likely not complete, but can easily be updated.
https://community.cesium.com/t/crs-84-not-equals-epsg-4326/7884/2


cesium-concierge:
Thanks for the pull request @tantricllama!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/wms-1.3.0-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/WebMapServiceImageryProvider.js
```diff
@@ -239,6 +239,37 @@ function WebMapServiceImageryProvider(options) {
...
     );
+
+    // The axis order in previous versions of the WMS specifications was to always use easting (x or lon ) and northing (y or
+    // lat). WMS 1.3.0 specifies that, depending on the particular CRS, the x axis may or may not be oriented West-to-East,
+    // and the y axis may or may not be oriented South-to-North. The WMS portrayal operation shall account for axis order.
+    // This affects some of the EPSG codes that were commonly used such as ESPG:4326. The current implementation
+    // makes sure that coordinates passed to the server (as part of the GetMap BBOX parameter) as well as those advertised
+    // in the capabilities document reflect the inverse axe orders for EPSG codes between 4000 and 5000.
+    //  - Taken from Section 9.1.3 of https://download.osgeo.org/mapserver/docs/MapServer-56.pdf
+    var parts = parameters.crs.split(":");
```
initialed85, line 250: @tantricllama Is it all possible for an incorrectly defined imagery provider (with `crs` parameter that isn't a sane EPSG code) to make it to here?

If so then it's possible that the `parts[0]` and `parts[1]` to be `undefined` however in my isolated testing I think that's fine.

If `parts[0]` is `undefined` then we don't end up inside the if statement.

If the `crs` parameter was just "EPSG" then `parts[1]` would be `undefined` and we would make it into the if statement, however `Number(undefined)` returns `NaN` which shouldn't break anything.

So I guess it's okay?

Source/Scene/WebMapServiceImageryProvider.js
```diff
@@ -239,6 +239,37 @@ function WebMapServiceImageryProvider(options) {
...
+    if (parts[0] === "EPSG") {
+      var includesReverseAxis = [
```
initialed85, line 252: @tantricllama I guess these lookups could grow over time, it might be better housekeeping to keep them at the top of the file (rather than redefine them per invocation of `WebMapServiceImageryProvider`).

smills2929:
Hi Cesium guys ( @lilleyse I saw you liked a comment on this ). What do you need from @tantricllama to get this merged?

smills2929:
@IanLilleyT @hpinkos Do you need anything else here to merge this? I'm nudging again because we are keen to avoid having to work from a fork of CesiumJS if at all possible.

Source/Scene/WebMapServiceImageryProvider.js
```diff
@@ -8,6 +8,29 @@ import GetFeatureInfoFormat from "./GetFeatureInfoFormat.js";
...
+
+/**
+ * EPSG codes known to not include reverse axis orders, and are within 4000-5000.
+ *
+ * @type {number[]}
+ */
+var excludesReverseAxis = [
+  4471, // Mayotte
+  4559, // French Antilles
+];
```
lilleyse, line 32: Out of curiosity how did you come up with this list and do you have a sense of how exhaustive it is? It should be possible to get a full list with a proj script. Doesn't necessarily need to be a blocker for the PR though since it's a good incremental improvement.

lilleyse (commented):
Thanks @tantricllama. I went through the WMS docs you posted and the change seems good. Just one question about the lists of EPSG codes that's included.

lilleyse:
I updated `CHANGES.md` in https://github.com/CesiumGS/cesium/pull/9797/commits/94877af4adbee24792865efba4d46dc6cb8ab72b

I'm still curious to hear about https://github.com/CesiumGS/cesium/pull/9797#discussion_r765352979 but this should be good to merge.

Source/Scene/WebMapServiceImageryProvider.js
tantricllama: https://github.com/mapserver/mapserver/issues/3582
I got the current list from this mapserver issue and used epsg.io to confirm each of them.
I've never used proj4 before, if you can provide me with some more info on how I could get the full list then I will give it a shot.

PR #9819, opened 2021-09-18
Added a `features` getter to `Cesium3DTileContent` that gets the features in the tile.

puxiao:
**Issues:** null

<br>

**It is convenient for us to traverse each feature.**

```diff
-  var featuresLength = content.featuresLength;
-  for (var i = 0; i < featuresLength; ++i) {
-      var feature = content.getFeature(i);
-      //...
-  }


+  content.features.forEach((feature) => {
+      //...
+  })
```

<br>

*Cesium.d.ts* after modification

```
export class Cesium3DTileContent {
    /**
     * Gets the features in the tile.
     */
    readonly features: Cesium3DTileFeature[];
}
```

### DIFF
--- a/Source/Scene/Batched3DModel3DTileContent.js
+++ b/Source/Scene/Batched3DModel3DTileContent.js
@@ -71,6 +71,16 @@ function Batched3DModel3DTileContent(
 Batched3DModel3DTileContent._deprecationWarning = deprecationWarning;
 
 Object.defineProperties(Batched3DModel3DTileContent.prototype, {
+  features: {
+    get: function () {
+      if (defined(this._batchTable)) {
+        createFeatures(this);
+        return this._features;
+      }
+      return [];
+    },
+  },
+
   featuresLength: {
     get: function () {
       return this._batchTable.featuresLength;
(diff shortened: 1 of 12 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/patch-4/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ebogo1:
@ptrgags @sanjeetsuhag Are either of you available to look at this PR? Thanks!

lilleyse:
Hi @puxiao, thanks for the PR but the current interface is intentional and this new interface could pose some issues:

* We want to prevent against unintended usage like adding or removing features from the array
* [EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) adds support for multiple feature ID sets and a `setId` parameter would not work with a plain getter
* In the future we might not always store features in arrays, and the overhead of packing them into arrays may be prohibitively expensive

PR #9828, opened 2021-09-24
Switch from uglify to terser

### LINKED ISSUE #9129: npm reports security vulnerability in rollup-plugin-uglify -> serialize-javascript
```
                       === npm audit security report ===


                                 Manual Review
             Some vulnerabilities require your attention to resolve

          Visit https://go.npm.me/audit-guide for additional guidance


  High            Remote Code Execution

  Package         serialize-javascript

  Patched in      >=3.1.0

  Dependency of   rollup-plugin-uglify [dev]

  Path            rollup-plugin-uglify > serialize-javascript

  More info       https://npmjs.com/advisories/1548

found 1 high severity vulnerability in 917 scanned packages
```

Because `rollup-plugin-uglify` is only used at build time to serialize the rollup plugin's configuration, it is not actually a relevant security vulnerability for CesiumJS. However, it _is_ a scary-looking message at npm install time, either from a release ZIP or from github (installing CesiumJS from npm is not affected because we won't get any `devDependencies` that way). We should fix it.

https://github.com/TrySound/rollup-plugin-uglify/pull/85 will fix it, but we need to wait for the `rollup-plugin-uglify` maintainers to merge that PR and release a new version. npm doesn't have an easy way to override a dependency of a dependency ([unlike yarn](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/)). Other options:

* Fork `rollup-plugin-uglify` and merge https://github.com/TrySound/rollup-plugin-uglify/pull/85 into our fork.
* Switch to [rollup-plugin-terser])(https://www.npmjs.com/package/rollup-plugin-terser), which is about 10x more popular according to npm stats.

mramato:
terser replaced uglify as the standard minifier most projects a long time ago and Cesium has been meaning to make the switch for a while.  According to CI, terser shaved about 90 seconds off of CI.

File sizes are nearly identical.  terser is actually slightly larger, but only because we are now properly including the `@license` comments in minified code (Which is something we used to do but lost during the great ES6 switchover) so we are glad to have this back.

Also fixes #9129


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/terser/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

gulpfile.cjs
```diff
@@ -1376,7 +1380,7 @@ function combineJavaScript(options) {
     promises.push(streamToPromise(stream));
 
     const everythingElse = ["Source/**", "!**/*.js", "!**/*.glsl"];
-    if (optimizer === "uglify2") {
+    if (minify) {
```
ebogo1, line 1383: I noticed while testing that this function leaves behind the empty `combineOutput` folder inside Build.

ebogo1 (commented):
Thanks @mramato - I tested this and it all seems to be good by me. Just in case I missed anything, I tested:

- That the `minify` task still generates a minified Cesium.js file (which is the same size as the one generated in `main`)
- That the output of `makeZipFile` is built correctly (Sandcastle, Documentation, Viewer, Hello World)
- Just in case, that regular unminified Build still works as expected

ebogo1:
Any reason not to update [Documentation/Contributors/BuildGuide/README.md](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/BuildGuide/README.md#build-scripts) with a link to [terser](https://github.com/terser/terser) instead of uglify?

> `minify` - Runs `combine`, plus minifies Cesium.js using UglifyJS2 for a smaller deployable file.

mramato:
> Any reason not to update [Documentation/Contributors/BuildGuide/README.md](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/BuildGuide/README.md#build-scripts) with a link to [terser](https://github.com/terser/terser) instead of uglify?

I just removed the extra info completely.  No need to list what minifier we are using in that part of the doc

gulpfile.cjs
mramato: Yeah, that's been the case forever. I wouldn't worry about it as part of this change since as you mentioned, it happens in main as well.

ebogo1:
Looks good, thanks @mramato (the CI failure looks to be a fluke in an unrelated spec that I haven't seen before, but I'll keep an eye out in case it keeps coming up).

Issue #9787, opened 2021-09-02
Revisit Draco3d downloads from gstatic


ebogo1:
This is a follow-up to https://github.com/CesiumGS/cesium/pull/9778, which just got merged in time for the 1.85 release. Instead of manually downloading .wasm and wrapper files from the [Draco repo](https://github.com/google/draco/), CesiumJS now fetches them from gstatic during install time with a `prepare` script in package.json. There are several loose ends from these changes:

- **Using version 1.3.5 instead of the latest (1.4.1 at the time of writing):** As of 1.3.6, it seems the Draco decoder does not work with our `decodeDraco` worker. I suspect this is due to our usage of `SkipAttributeTransform`: https://github.com/CesiumGS/cesium/blob/a02065697660f2ee1960548a5630578a8a11e3d0/Source/WorkersES6/decodeDraco.js#L278-L284 When these lines are commented out, newer Draco versions work fine but the attributes are returned as Float32 arrays instead of Uint16 arrays. The action item here is to do some digging into the `SkipAttributeTransform` source code and try to make an example demonstrating a bug if there is one (and to open an issue in the Draco3d repo). 
- **Using the gstatic files instead of the npm package:** I tried two things with the npm module:
  - **Point the Draco `TaskProcessor` directly to the .wasm and wrapper files in the npm package dist:** This didn't work due to a runtime error, possibly related to the first bullet point but I'm not sure.
  - **Compile the .wasm module from the npm package using the decoder API:** Similarly to the other libraries we pull from npm, I imported the Decoder from the Draco module. Then, following the [example](https://github.com/google/draco/blob/master/javascript/npm/draco3d/draco_minimal_decoder_example.js) in the Draco repo,  modified the `decodeDraco(event)` function in our Draco worker to load the module using the `createDecoderModule ` API. I think the changes in the worker were correct, but I was unable to test because I ran into build errors due to a dependency on `fs` and `path` inside the Decoder from npm when building the files in Source/ThirdParty/.
- **Updating the gulp "prepare" task:** With the inclusion of a `"prepare"` script in package.json, we needed to remove the script from the package.json that gets bundled into our zip file. Since we rushed to get the Draco download changes into 1.85, we might've missed a cleaner way to do this; ideally, we wouldn't need to write out a temporary `package.noprepare.json` file during the `"makeZipFile"` task and delete it immediately after.

lilleyse:
[...]

PR #9841, opened 2021-09-28
Grab draco3d from npm instead of gstatic

ebogo1:
Band-aid fix for https://github.com/CesiumGS/cesium/issues/9787.

This PR is a temporary workaround for streamlining all our third party libraries to come from npm. Unfortunately, as mentioned in the above issue, the draco3d npm package seems to have a bug with quantization so this PR introduces a regression by disabling quantization in our `decodeDraco` worker. 

The examples on the draco3d repo instantiate the .wasm module using the Decoder API, but doing so with our current npm build system doesn't play nicely with rollup because the Decoder module requires Node built-in libraries (`fs` and `path`) which are missing at build time. Because of all this, the PR changes are:
1. Add the `draco3d` npm package to package.json
2. In the `postinstall` script, copy the wrapper (now named `draco_decoder_nodejs.js` instead of `draco_wasm_wrapper.js`) and .wasm file from node_modules/ to their original locations in Source/ThirdParty/.
3. Comment out the lines dealing with quantization for meshes in the `decodeDraco` worker. A few specs failed because of this change so for now they are `xit`'d out.

This also updates Draco to 1.4.1.

### DIFF
--- a/Source/WorkersES6/decodeDraco.js
+++ b/Source/WorkersES6/decodeDraco.js
@@ -272,30 +272,6 @@ function decodePointCloud(parameters) {
   return result;
 }
 
-function findQuantizationTransform(array, numComponents) {
-  var min = [];
-  var max = [];
-  var i, j;
-  for (i = 0; i < numComponents; ++i) {
-    min.push(array[i]);
-    max.push(array[i]);
-  }
-  for (i = 0; i < array.length; i += numComponents) {
-    for (j = 0; j < numComponents; ++j) {
-      min[j] = Math.min(array[i + j], min[j]);
-      max[j] = Math.max(array[i + j], max[j]);
-    }
-  }
-  var range = max[0] - min[0];
-  for (i = 1; i < numComponents; ++i) {
-    range = Math.max(max[i] - min[i], range);
-  }
-  return {
-    min: min,
-    range: range,
-  };
-}
-
 function decodePrimitive(parameters) {
   var dracoDecoder = new draco.Decoder();
 
@@ -305,12 +281,6 @@ function decodePrimitive(parameters) {
     for (var i = 0; i < attributesToSkip.length; ++i) {
       dracoDecoder.SkipAttributeTransform(draco[attributesToSkip[i]]);
     }
-  }
-
-  // SkipAttributeTransform appears to work on "NORMAL" but not on other attributes.
-  // See https://github.com/google/draco/issues/742.
-  if (parameters.dequantizeInShader) {
-    dracoDecoder.SkipAttributeTransform(draco["NORMAL"]);
   }
 
   var bufferView = parameters.bufferView;
@@ -334,9 +304,6 @@ function decodePrimitive(parameters) {
 
   var attributeData = {};
 
-  // Skip "NORMAL" because we called SkipAttributeTransform(draco["NORMAL"])
-  var attributesToQuantize = ["POSITION", "COLOR", "TEX_COORD"];
-
   var compressedAttributes = parameters.compressedAttributes;
   for (var attributeName in compressedAttributes) {
     if (compressedAttributes.hasOwnProperty(attributeName)) {
@@ -350,35 +317,6 @@ function decodePrimitive(parameters) {
         dracoDecoder,
         dracoAttribute
       );
-      if (
-        parameters.dequantizeInShader &&
-        attributesToQuantize.includes(attributeName)
-      ) {
-        var data = attributeData[attributeName].data;
-        var minMax = findQuantizationTransform(
-          attributeData[attributeName].array,
-          data.componentsPerAttribute
-        );
-        var range = minMax.range;
-        var length = attributeData[attributeName].array.length;
-        var quantizedArray = new Uint16Array(length);
-        for (var i = 0; i < length; i += data.componentsPerAttribute) {
-          for (var j = 0; j < data.componentsPerAttribute; ++j) {
-            var value = attributeData[attributeName].array[i + j];
-            quantizedArray[i + j] = ((value - minMax.min[j]) / range) * 65535;
-          }
-        }
-        attributeData[attributeName].array = quantizedArray;
-        attributeData[attributeName].data.byteStride /= 2;
-        attributeData[attributeName].data.componentDatatype =
-          ComponentDatatype.UNSIGNED_SHORT;
-        attributeData[attributeName].data.quantization = {
-          octEncoded: false,
-          quantizationBits: 16,
-          minValues: minMax.min,
-          range: minMax.range,
-        };
-      }
     }
   }
 
--- a/Source/WorkersES6/decodeDraco.js
+++ b/Source/WorkersES6/decodeDraco.js
@@ -299,6 +299,16 @@ function findQuantizationTransform(array, numComponents) {
 function decodePrimitive(parameters) {
   var dracoDecoder = new draco.Decoder();
 
+  // Skip all parameter types except generic
+  var attributesToSkip = ["POSITION", "NORMAL", "COLOR", "TEX_COORD"];
+  if (parameters.dequantizeInShader) {
+    for (var i = 0; i < attributesToSkip.length; ++i) {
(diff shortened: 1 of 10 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/no-quantization/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/WorkersES6/decodeDraco.js
```diff
@@ -276,12 +276,12 @@ function decodePrimitive(parameters) {
...
-    for (var i = 0; i < attributesToSkip.length; ++i) {
-      dracoDecoder.SkipAttributeTransform(draco[attributesToSkip[i]]);
-    }
-  }
+  // var attributesToSkip = ["POSITION", "NORMAL", "COLOR", "TEX_COORD"];
+  // if (parameters.dequantizeInShader) {
+  //   for (var i = 0; i < attributesToSkip.length; ++i) {
+  //     dracoDecoder.SkipAttributeTransform(draco[attributesToSkip[i]]);
+  //   }
+  // }
```
lilleyse, line 284: To help keep the code clean let's delete this code and failing tests in a single isolated commit, then in a new issue mention that the commit should be reverted once the draco issue is fixed upstream and we update the npm package.

Source/WorkersES6/decodeDraco.js
```diff
@@ -276,12 +276,12 @@ function decodePrimitive(parameters) {
   var dracoDecoder = new draco.Decoder();
 
   // Skip all parameter types except generic
-  var attributesToSkip = ["POSITION", "NORMAL", "COLOR", "TEX_COORD"];
-  if (parameters.dequantizeInShader) {
-    for (var i = 0; i < attributesToSkip.length; ++i) {
-      dracoDecoder.SkipAttributeTransform(draco[attributesToSkip[i]]);
-    }
-  }
+  // var attributesToSkip = ["POSITION", "NORMAL", "COLOR", "TEX_COORD"];
```
ebogo1, line 279: @lilleyse Is this ok to merge as is? I remember you suggested commenting out the line in Model.js instead? I think it was -
https://github.com/CesiumGS/cesium/blob/e8f8fb970823b46318362fc5972e42e3832f647d/Source/Scene/Model.js#L638

ebogo1:
Looks like CI/specs are failing with a few more that I missed, I think those should be `xit`'d too: 
```
1) loads Duck
     Scene/GltfLoader
2) adds a dequantization function
     Scene/ModelExperimental/DequantizationPipelineStage
3) adds dequantization uniforms
     Scene/ModelExperimental/DequantizationPipelineStage
4) prepares Draco model for dequantization stage
     Scene/ModelExperimental/GeometryPipelineStage
```

Source/WorkersES6/decodeDraco.js
lilleyse: Yeah this is ok to merge as is.

lilleyse:
I just want to emphasize that this PR is an unfortunate but necessary tradeoff that we hope is just temporary.

Geometry memory usage for 3D Tiles will **double** because we can no longer dequantize data on the GPU due to a bug in newer versions of draco https://github.com/google/draco/issues/742. We can no longer use our older working version of Draco due to a requirement to move all third party libraries to npm, which is further complicated by the fact that older versions of draco on npm use asm.js instead of WebAssemly and aren't compatible with our web worker system.

Source/WorkersES6/decodeDraco.js
lilleyse: Actually, can keep this code and only enable `"NORMAL"`? Since it is working for normals we may as well take what we can get.

lilleyse:
@ebogo1 just had an idea - even though draco will return the full decoded data now, we can still quantize the data ourselves in the worker so that we don't have to resort to doubling the memory usage. We'll just need to convert the `Float32Array` back to a `Uint16Array` or `Uint8Array` manually.

Source/WorkersES6/decodeDraco.js
```diff
@@ -272,16 +272,36 @@ function decodePointCloud(parameters) {
...
+  }
+  for (i = 0; i < array.length; i += numComponents) {
+    for (j = 0; j < numComponents; ++j) {
+      min[j] = Math.min(array[i + j], min[j]);
+      max[j] = Math.max(array[i + j], max[j]);
+    }
+  }
+  var range = max[0] - min[0];
+  for (i = 1; i < numComponents; ++i) {
+    range = Math.max(max[i] - min[i], range);
```
ebogo1, line 291: This is not what I expected - the [draco source code](https://github.com/google/draco/blob/85a614fa6639d155f47ec68a776258a0e28b918f/src/draco/attributes/attribute_quantization_transform.h#L96-L97) say `range` is the "max delta over all components." I first thought this would mean the magnitude of `max - min`, but doing it this way matches the range returned by the decoder in main.

ptrgags (commented):
@ebogo1 looks pretty good, though I'm noticing one problem with the NYC dataset, picking is now selecting way too many buildings:

Maybe check if the `_BATCH_ID` property is going down the correct path?

image

The milk truck model worked fine though!

Source/WorkersES6/decodeDraco.js
ptrgags: Oh, I didn't see this, maybe ignore my comment in the PR review :sweat_smile:

Source/WorkersES6/decodeDraco.js
```diff
@@ -317,6 +337,32 @@ function decodePrimitive(parameters) {
         dracoDecoder,
         dracoAttribute
       );
+      if (parameters.dequantizeInShader && attributeName !== "NORMAL") {
```
ptrgags, line 340: @ebogo1 correction, from the [B3DM Spec](https://github.com/CesiumGS/3d-tiles/tree/main/specification/TileFormats/Batched3DModel#binary-gltf) it's called `_BATCHID` (all one word). It's probably due to the fact that this is not `NORMAL`, but it's not one of the built-ins like `POSITION` either.

I wonder if it's going down the wrong path due to this if statement.

ebogo1: Good catch - with the code in `main` it looks like we wouldn't expect to use Draco's quantization transform on anything other than the properties in `var attributesToSkip = ["POSITION", "NORMAL", "COLOR", "TEX_COORD"];`.

Source/WorkersES6/decodeDraco.js
```diff
@@ -272,16 +272,36 @@ function decodePointCloud(parameters) {
...
+    min.push(array[i]);
+    max.push(array[i]);
+  }
+  for (i = 0; i < array.length; i += numComponents) {
+    for (j = 0; j < numComponents; ++j) {
+      min[j] = Math.min(array[i + j], min[j]);
+      max[j] = Math.max(array[i + j], max[j]);
+    }
+  }
+  var range = max[0] - min[0];
```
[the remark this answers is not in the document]
ebogo1: Based on inspecting the code in `main` I think this is supposed to be a number. That's also in line with the Draco [source code](https://github.com/google/draco/blob/85a614fa6639d155f47ec68a776258a0e28b918f/src/draco/attributes/attribute_quantization_transform.h#L96-L97):

```cpp
int32_t quantization_bits_;

// Minimal dequantized value for each component of the attribute.
std::vector<float> min_values_;

// Bounds of the dequantized attribute (max delta over all components).
float range_;
```

ptrgags:
@ebogo1 code looks good (aside from the tests that we want to temporarily remove). NYC dataset is picking correctly again!

let me know when you remove the broken tests and when you've documented what commits need to be reverted after draco is fixed, then I'll merge.

ptrgags:
@ebogo1 changes look good, thanks!

ebogo1:
Thanks @ptrgags - I opened https://github.com/CesiumGS/cesium/issues/9847 to follow up.

PR #9842, opened 2021-09-29
Add support for `TILE_GEOMETRIC_ERROR` semantic

lilleyse:
Implements the 3D Tiles Next spec change in https://github.com/CesiumGS/3d-tiles/pull/487 which allows you to specify geometric error for each tile in an implicit subtree. This gives more flexibility that the default behavior which is to half the geometric error at each level.

The tests files were generated from this commit in 3d-tiles-samples-generator: https://github.com/CesiumGS/3d-tiles-validator/commit/9a67d2112f7c9ae50c205ad8a3e9be11e946ba49


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/implicit-tile-geometric-error/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
Code looks good, thanks @lilleyse!

PR #9845, opened 2021-09-30
Adding urn:ogc:def:crs:EPSG::4283 in crsNames

muntakim1:
Horizontal component of 3D system. Here are more detailed information about this crsName https://epsg.io/4283

### DIFF
--- a/Source/DataSources/GeoJsonDataSource.js
+++ b/Source/DataSources/GeoJsonDataSource.js
@@ -35,6 +35,7 @@ var crsNames = {
   "urn:ogc:def:crs:OGC:1.3:CRS84": defaultCrsFunction,
   "EPSG:4326": defaultCrsFunction,
   "urn:ogc:def:crs:EPSG::4326": defaultCrsFunction,
+  "urn:ogc:def:crs:EPSG::4283":defaultCrsFunction,
 };
 
 var crsLinkHrefs = {};


cesium-concierge:
Thank you so much for the pull request @muntakim1! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/muntakim1/cesium/blob/main/CONTRIBUTORS.md) file!
* :x: Missing CLA.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
I don't think this change is correct. `defaultCrsFunction` transforms WGS84 geographic coordinates to cartesian coordinates and would not work for GDA94. A more complex function would be required for that, possibly using proj4js. You can add your own coordinate transform like this:

```
GeoJsonDataSource.crsNames["urn:ogc:def:crs:EPSG::4283"] = function(coordindates) {
    ...add custom code here...
    return cartesian;
}
```

lilleyse:
Ok so maybe they are close enough to be considered equivalent? From http://www.geoproject.com.au/gda.faq.html

> **What is the difference between GDA94 and WGS84?**
> WGS84 is the datum used by the GPS system, and also for many small scale maps (e.g. maritime charts).
> 
>
> When GDA94 was defined, it was a requirement that coordinates of points in these 2 systems should be within a metre of each other. This was achieved, and is generally still the case in 2000. The spheroids used for WGS84 and GDA84 are also almost identical, and both systems are geocentric.
> 
> Thus for most mapping, exploration and GIS uses, WGS84 and GDA94 coordinates will be the same. If you use a GPS receiver (alone, or with a differential service) and the GDA datum is not a built-in option, you can select WGS84.
> 
> For precise surveys, however, the difference between WGS84 and GDA94 may be significant, and changes slowly over time. There are two reasons:
> 
> WGS84 is recalculated on a world-wide basis every few years for GPS control
> Tectonic movements of the Australian plate affect WGS coordinates for Australian stations, but do not affect the GDA coordinates (see answer below "How will continental drift affect GDA94?").
>

Issue #9855, opened 2021-10-07
Broken example sandcastle for 'CZML Model Articulations'


fc:
Your sandcastle example:
https://sandcastle.cesium.com/?src=CZML%20Model%20Articulations.html&label=CZML

The error says:
```
RuntimeError: Failed to load model: https://assets.agi.com/models/launchvehicle.glb
undefined
Error
    at new RuntimeError (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:6891:13)
    at https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:140774:34
    at Promise.then (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:15824:35)
    at https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:15904:15
    at processQueue (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:16254:6)
    at _resolve (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:15940:6)
    at promiseReject (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:15973:13)
    at Promise.then (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:15824:35)
    at https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:15905:9
    at processQueue (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:16254:6)
```

When I load the GLB URL in the browser directly it seems to load fine.. definitely not obvious what the issue is.

fc:
It still doesn't work for me -- guess it will get deployed later on?

hpinkos:
@fc the fix will be deployed to sandcastle.cesium.com with the next release.  In the meantime, if you'd like to test it out, you can open the demo and replace `"https://assets.agi.com/models/launchvehicle.glb"` with `"https://cesium.com/public/SandcastleSampleData/launchvehicle.glb"`

PR #9856, opened 2021-10-07
Fix articulated model CZML Sandcastle example

hpinkos:
Fixes https://github.com/CesiumGS/cesium/issues/9855

It looks like there `assets.agi.com` no longer has CORS enabled so https://sandcastle.cesium.com/?src=CZML%20Model%20Articulations.html&label=CZML can't load the sample model.  I put a copy on cesium.com/public/SandcastleSampleData instead.

### DIFF
--- a/Apps/Sandcastle/gallery/CZML Model Articulations.html
+++ b/Apps/Sandcastle/gallery/CZML Model Articulations.html
@@ -53,7 +53,8 @@
               cartographicDegrees: [-77, 37, 10000],
             },
             model: {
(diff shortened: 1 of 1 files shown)


cesium-concierge:
Thanks for the pull request @hpinkos!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-articulated-model/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #9857, opened 2021-10-08
Fix binding parameter error

lovefengruoqing:


lilleyse:
I pushed this fix in https://github.com/CesiumGS/cesium/pull/9877.

PR #9858, opened 2021-10-08
Switch to Node 16 for CI

mramato:
Node 16 will be LTS on `2021-10-26` so we need to ensure it's passing and works.

We might want to consider testing against multiple Node versions down the line, but using the latest LTS has worked well so far (though apparently we skipped 14 completely here)

I expect this branch to fail until fixes we know are needed for Node 16 get merged.

CC @ebogo1


lilleyse:
CI is passing now that https://github.com/CesiumGS/cesium/pull/9861 was merged. Thanks @mramato.

Issue #9785, opened 2021-09-02
Cloud noise texture is too big for some devices


lilleyse:
The cloud texture is a 3D texture with dimensions 128x128x128 that's emulated by a 2D texture with dimensions 16384 x 128. This is exceeding the maximum texture size on certain devices (see `Max Texture Size` on https://webglreport.com/).

The easiest solution is to get closer to a square aspect ratio. A 128x128x128 texture can be packed as 1536 x 1408 texture (or something similar, haven't worked out the ideal configuration).

A workaround in the meantime is to reduce the noise texture size. This can be done by adding the following line after creating a `CloudCollection`:

```
cloudCollection._noiseTextureLength = 64;
```

Just note that this does alter the appearance of the clouds compared to 128, the default.

PR #9885, opened 2021-10-21
Reduce cloud texture dimensions

ebogo1:
Fixes https://github.com/CesiumGS/cesium/issues/9785.

The context for this fix is that the 3D noise texture is represented an array of "slices," each of which is a 2D square texture. In `main`, slices are outputted from `CloudNoiseFS` as a single texture with dimensions `sliceWidth * sliceWidth` x `sliceWidth` (width x height). With a default sliceWidth of `128`, the width of the noise texture is 16384 pixels (128 * 128), which exceeds the max WebGL texture size of some machines and browsers. 

The fix in this PR uses the same 3D noise texture as before, but writes the 2D slices to four rows instead of one (i.e., with a slice width of 8, the first four slices would be on row 1 and the second four slices would be on row 2). With a slice width of 128, the dimensions of the noise texture become 4096 x 512 (the width is 128 * 128 / 4, the height is 128 * 4). According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#understand_system_limits), 4096 is supported by "most systems." Maybe there should still be a check to disable clouds if the system doesn't support it? I am not sure how common this is.

Changes: - 
- `CloudCollection` now creates a texture with dimensions `length * length / 4` x `length * 4` instead of `length * length` x `length`.
- The `u_noiseTextureLength` uniform was renamed to `u_textureSliceWidth` to avoid confusion between the individual slices and the one flattened 3D texture.
- `CloudNoiseVS` and `CloudNoiseFS` now write the 3D noise texture to multiple rows instead of one row.
- `CloudCollectionFS` has new sampling logic to transform 3D coordinates to UV space.. I'll add review comments about some ideas here.
- A stray `console.log` was removed from `CumulusCloud`.

I recommend viewing the diff without whitespace changes.

### DIFF
--- a/Source/Shaders/CloudCollectionFS.glsl
+++ b/Source/Shaders/CloudCollectionFS.glsl
@@ -42,7 +42,7 @@ vec2 voxelToUV(vec3 voxelIndex) {
 // Interpolate a voxel with its neighbor (along the positive X-axis)
 vec4 lerpSamplesX(vec3 voxelIndex, float x) {
     vec2 uv0 = voxelToUV(voxelIndex);
-    vec2 uv1 = uv0 + vec2(inverseNoiseTextureDimensions.x, 0.0);
+    vec2 uv1 = voxelToUV(voxelIndex + vec3(1.0, 0.0, 0.0));
     vec4 sample0 = texture2D(u_noiseTexture, uv0);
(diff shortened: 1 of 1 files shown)


lilleyse:
Good catch there at the end, generally mods/floors are fast so it's not too big a concern.

Issue #9847, opened 2021-10-01
Re-enable Draco quantization


ebogo1:
After https://github.com/CesiumGS/cesium/pull/9841 gets merged, we will no longer quantize any attributes in the Draco worker. Some immediate tech debt from this PR: 

- This is a temporary workaround while we switch to the `draco3d` npm module, but it should be doable to at least get some quantization benefits by packing compressed attributes into 16 bits. In the PR, c9ac980 has some progress towards getting the math to work in the shader but has inconsistent results with different models (i.e. with `Specs/Data/Models/DracoCompression/BoxVertexColorsDracoRGBA.gltf` Draco returns a null pointer even on the `"NORMAL"` attribute, which is unexpected). 
- Some specs related to quantization were changed or deleted. These two commits should be reverted once quantization is added back: 1a3b697, 97c186c

Since this is a sizable regression, we should fix as much as we can in main ASAP and consider a follow-up 1.86.1 release.

related: #9787

ebogo1:
Update: most of the specs are now back in `main` after https://github.com/CesiumGS/cesium/pull/9904 got merged. The specs for PointCloud3DTileContentSpec.js are the only ones left, and should be added in once the Draco issue is fixed.

ebogo1:
After https://github.com/CesiumGS/cesium/pull/9908/ is merged, all the specs will have been added back in. Once the Draco bug is fixed in the next release and we update the Draco version in CesiumJS, this issue can be closed.

PR #9904, opened 2021-10-29
Use `GetAttribute` to enable Draco quantization when possible

ebogo1:
This is a temporary workaround for https://github.com/CesiumGS/cesium/issues/9847. See https://github.com/google/draco/issues/742#issuecomment-954934088 for more info. Since `SkipAttributeTransform` breaks lookups with `GetAttributeByUniqueId`, this PR uses the alternative approach with `GetAttributeId` and `GetAttribute` to preserve quantization when possible. 

The downside is that this might not work when there are multiple sets of an attribute, i.e. `TEXCOORD_0` and `TEXCOORD_1`. Because of this, we only call `SkipAttributeTransform` with the new `GetAttribute` approach for "single" attributes. Otherwise, we continue without quantization and use the `GetAttributeByUniqueId` call as before.

I'm working on adding back some of the specs that were removed previously, but the core change is all in decodeDraco.js.

### DIFF
--- a/Source/WorkersES6/decodeDraco.js
+++ b/Source/WorkersES6/decodeDraco.js
@@ -275,6 +275,24 @@ function decodePointCloud(parameters) {
 function decodePrimitive(parameters) {
   var dracoDecoder = new draco.Decoder();
 
+  // Skip all parameter types except generic
+  // Note: As a temporary work-around until GetAttributeByUniqueId() works after
+  // calling SkipAttributeTransform(), we will not skip attributes with multiple
+  // sets of data in the glTF.
+  var attributesToSkip = ["POSITION", "NORMAL"];
+  var compressedAttributes = parameters.compressedAttributes;
+  if (!defined(compressedAttributes["COLOR_1"])) {
+    attributesToSkip.push("COLOR");
+  }
+  if (!defined(compressedAttributes["TEXCOORD_1"])) {
+    attributesToSkip.push("TEX_COORD");
+  }
+  if (parameters.dequantizeInShader) {
+    for (var i = 0; i < attributesToSkip.length; ++i) {
+      dracoDecoder.SkipAttributeTransform(draco[attributesToSkip[i]]);
+    }
+  }
+
   var bufferView = parameters.bufferView;
   var buffer = new draco.DecoderBuffer();
   buffer.Init(parameters.array, bufferView.byteLength);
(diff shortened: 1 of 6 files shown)


cesium-concierge:
    ```
    ```
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/draco-workaround/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/WorkersES6/decodeDraco.js
```diff
@@ -275,6 +275,24 @@ function decodePointCloud(parameters) {
...
+  // calling SkipAttributeTransform(), we will not skip attributes with multiple
+  // sets of data in the glTF.
+  var attributesToSkip = ["POSITION", "NORMAL"];
+  var compressedAttributes = parameters.compressedAttributes;
+  if (!defined(compressedAttributes["COLOR_1"])) {
+    attributesToSkip.push("COLOR");
+  }
+  if (!defined(compressedAttributes["TEXCOORD_1"])) {
+    attributesToSkip.push("TEX_COORD");
+  }
```
lilleyse, line 289: glTF does allow for `TEXCOORD_2`, `TEXCOORD_3`, etc so some of this glTF attribute name -> draco name mapping could be organized in a more general way. Though to be fair, `TEXCOORD_2` and higher are only used in `ModelExperimental` and if we're expecting a bug fix in draco soon then maybe we don't have to worry about it here.

ebogo1: My thinking was that if there were more than one TEXCOORD, they numbering would go in order, meaning that at least `TEXCOORD_1` would exist - in which case, we would not call `SkipAttributeTransform()`. Is it possible to just have one TEXCOORD and call it `TEXCOORD_2`? I'm not sure I'm following your comment.

lilleyse: Oh that's true, never mind then. From the glTF spec:

> All indices for indexed attribute semantics MUST start with 0 and be consecutive positive integers

lilleyse:
Thanks for the quick turnaround @ebogo1. For 99% of cases we can use quantization again!

lilleyse:
Forgot to mention, I tested all the 3D models and 3D Tiles related sandcastles and ran the unit tests locally.

ebogo1:
Thanks @lilleyse. I tested [microcosm-draco.gltf.zip](https://github.com/CesiumGS/cesium/files/7445349/microcosm-draco.gltf.zip) (converted from https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Models/GltfLoader/Microcosm/glTF with gltf-pipeline), which has multiple TEXCOORD attributes and all seems to work as expected.

PR #9908, opened 2021-11-01
Add Draco quantization workaround for point clouds

ebogo1:
This fixes a regression from 1.86 after updating Draco, where `SkipAttributeTransform` broke future attribute lookups with `GetAttributeByUniqueId`. To reproduce, see this [Sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20Formats.html&label=All) and select the "PointCloudDraco" tileset. 

This change is analogous to https://github.com/CesiumGS/cesium/pull/9904/, but since `decodePointCloud` always tries to call `SkipAttributeTransform` on `"POSITION"` and `"NORMAL"`, those are the two attributes that use `GetAttributeId` and `GetAttribute` instead of `GetAttributeByUniqueId`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,13 +5,14 @@
 ##### Additions :tada:
 
 - Added `ScreenOverlay` support to `KMLDataSource`. [#9864](https://github.com/CesiumGS/cesium/pull/9864)
-- Fixed crashes caused by the cloud noise texture exceeding WebGL's maximum supported texture size. [#9885](https://github.com/CesiumGS/cesium/pull/9885)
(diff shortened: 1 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/draco-point-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
@ebogo1 looks good. At one point you mentioned that there was a point cloud draco test that had been removed in https://github.com/CesiumGS/cesium/pull/9841 that was still failing. Is that still the case, or can we add it back now?

ebogo1:
@lilleyse Good catch - I added the tests back in 6c55ad1.

PR #9917, opened 2021-11-06
Fix error handling for ModelExperimental

lilleyse:
Model loading errors were not getting propagated correctly in `ModelExperimental`. I added two unit tests but this should have been unit tested when `ModelExperimental` was first added.

`loadAndZoomToModelExperimental.js` was also not handling rejected promises correctly. I modeled it after `waitForLoaderProcess.js`. It's a little verbose but it should work for catching any errors.

### DIFF
--- a/Source/Scene/ModelExperimental/ModelExperimental.js
+++ b/Source/Scene/ModelExperimental/ModelExperimental.js
@@ -202,17 +202,17 @@ function initialize(model) {
       });
       model._resourcesLoaded = true;
     })
-    .otherwise(function () {
-      ModelExperimentalUtility.getFailedLoadFunction(this, "model", resource);
-    });
+    .otherwise(
+      ModelExperimentalUtility.getFailedLoadFunction(model, "model", resource)
+    );
 
   loader.texturesLoadedPromise
     .then(function () {
       model._texturesLoaded = true;
     })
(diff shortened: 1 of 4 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-error-handling/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -202,17 +202,17 @@ function initialize(model) {
       });
       model._resourcesLoaded = true;
     })
-    .otherwise(function () {
-      ModelExperimentalUtility.getFailedLoadFunction(this, "model", resource);
-    });
+    .otherwise(
+      ModelExperimentalUtility.getFailedLoadFunction(model, "model", resource)
+    );
```
lilleyse, line 207: This is a combination of two bugs:

* `getFailedLoadFunction` is creating the function that otherwise calls so it shouldn't be wrapped in a function
* `this` should be `model`

sanjeetsuhag (approved):
Changes look good. Thanks @lilleyse!

PR #9942, opened 2021-11-28
Add option to load glTF vertex and index buffers as typed arrays

lilleyse:
Adds a new option to `GltfLoader` called `loadAsTypedArray` which loads vertex and index buffers as typed arrays instead of `Buffer` objects that contain WebGL resources. This is useful for code that uses `GltfLoader` but has separate methods of uploading the data to the GPU.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/load-buffer-as-typed-array/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/GltfLoader.js
```diff
@@ -69,6 +69,7 @@ var GltfLoaderState = {
  * @param {Boolean} [options.incrementallyLoadTextures=true] Determine if textures may continue to stream in after the glTF is loaded.
  * @param {Axis} [options.upAxis=Axis.Y] The up-axis of the glTF model.
  * @param {Axis} [options.forwardAxis=Axis.Z] The forward-axis of the glTF model.
+ * @param {Boolean} [options.loadAsTypedArray=false] Load all attributes as typed arrays instead of GPU buffers.
```
sanjeetsuhag, line 72: > Load all **attributes** as typed arrays instead of GPU buffers.

The indices are also loaded as typed arrays. Does that need to be explicitly mentioned here? Not sure if the `attributes` umbrella implicitly covers those too.

sanjeetsuhag (commented):
@lilleyse The code and specs here look good, I just had one suggestion regarding the names because I am not sure if the increased verbosity is necessary here.

lilleyse:
@sanjeetsuhag thanks for the naming suggestions, I like them better than what I had. This PR is ready for another review.

sanjeetsuhag (approved):
Changes look good. Thanks @lilleyse!

Issue #9814, opened 2021-09-16
`GltfLoader` isn't correctly detecting quantized vertex colors


ptrgags:
in https://github.com/CesiumGS/cesium/pull/9806#issuecomment-920945046, @sanjeetsuhag noticed that a model with Draco + vertex colors was not working correctly with `ModelExperimental`. Data is the same as in #6561

I tracked it down, it's the `GltfLoader` code that's causing issues, the quantization is not detected correctly. 

Opening this issue since it's getting beyond the scope of #9806 though should be addressed soon.

ptrgags:
There's at least two problems here:

1. the models above have a batch ID attribute which is not correctly handled (https://github.com/CesiumGS/cesium/issues/9956). This is really a separate bug but it prevents me from loading that particular dataset. I might try using `gltfpack` on the point cloud model from #9955 instead since that doesn't have any batch IDs so I can address this separately from #9956
2. Quantized vertex colors need some special handling:
  * The quantized attribute can be uploaded as a vec3, but should be declared in the shader as a `vec4` so the w component is 1 
   * The offset uniform should be promoted to a Cartesian4 with a w component of 0. 
   * The scaling uniform should be promoted to a Cartesian4 with a w component of 1
   * Both of those uniforms should be declared as a vec4 in the shader

PR #9955, opened 2021-12-02
`ModelExperimental` - fix handling of RGB vertex colors

### LINKED ISSUE #9948: Cannot render a glTF point cloud with 3-component colors
When trying to render a glTF that contains a "point cloud" with colors that are declared as a `VEC3` with `GL_FLOAT` components, it generates an error message:

> Fragment shader compile log: ERROR: 0:142: 'assign' : cannot convert from 'highp 3-component vector of float' to 'highp 4-component vector of float'

The title and the quick summary here already contain many guesses. For example, I have to admit that I did not try out whether it can render a glTF with `VEC4` colors, but ... based on the error message, I think that the `VEC3` is the reason for the crash: The fragment shader code probably contains a hard-wired `vec4` for the colors, and attempting to feed it with a `vec3` does not work. If this is the case, the fragment shader code "template" has to fetch the information of whether it should use `vec3` or `vec4` from the glTF accessor.

The attached file contains

- The GLB file
- A tileset that refers to that GLB file (with `3DTILES_content_gltf`)
- The `ViewerCode.txt` with the sandcastle code where I tested this locally (reduced to the bare minimum to reproduce)
- The `ErrorMessage.txt` with the full stack trace

[Cesium testColorVec3 2021-11-30.zip](https://github.com/CesiumGS/cesium/files/7628192/Cesium.testColorVec3.2021-11-30.zip)

ptrgags:
Fixes #9948

I was incorrectly handling the `COLOR_n` attributes (I was checking for `color` instead of `color_n` for the shader variable). I found the cleanest fix was to do the type promotion from `vec3 -> vec4` 

Note that this PR does not address quantized vertex colors (see #9814), I plan to do that in a separate PR since it's more involved


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-vec3-color/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag (approved):
Thanks @ptrgags! The changes here look good. The shader generated correct, as shown below:

| Vertex Shader | Fragment Shader |
| - | - |
| image | image |

PR #9960, opened 2021-12-03
Add a `pointSize` field to custom vertex shaders

### LINKED ISSUE #9851: Custom Shaders: let the shader control `gl_PointSize`
It would be useful for custom shaders to be able to control the point size when rendering point clouds. This is available in styling language, so custom shaders should support the same.

I'm thinking: 

* add a `float vsInput.pointSize` to the `VertexInput` struct in the custom vertex shader
* in `ModelExperimentalVS.glsl` should be used to set `gl_PointSize` if custom shaders is used (which we consider mutually exclusive with styling)
* I'm not sure how this will interact with the point cloud attenuation... maybe the struct value would be initialized with the attenuation point size and then the custom shader could modify this value if so desired?

CC @lilleyse @sanjeetsuhag

ptrgags:
Fixes #9851

This PR adds the following:

* The `vertexMain` callback signature is changed to take a struct that contains both the `positionMC` and a `pointSize` field
* The `pointSize` field is used to customize `gl_PointSize` in the shader.

I updated [this sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Custom%20Shaders%20Models.html&label=3D%20Tiles%20Next) with an example.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/custom-shaders-point-size/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
Really nice! Now anyone can write their own point cloud attenuation function.

Issue #8481, opened 2019-12-19
Globe picking is very slow


IanLilleyT:
The current approach for globe picking is intersect the ray against all rendered tiles (using their bounding spheres) and sort the intersected tiles by closeness to ray. Then, for the closest tile do ray-triangle intersection tests against all of its triangles and return the closest. If there was none, test the next tile. And repeat.

This could be optimized some more. `ScreenSpaceCameraController` does one or two globe picks per frame and that alone can slow down Cesium severely with detailed terrain providers like `ArcGIS`.

Some ideas on how to improve this:

- Custom intersection test for heightmap providers since we know their triangles are on a grid
- Per-tile octree for non-heightmap providers (too slow to generate?)
- Remove `TerrainEncoding` quantization for faster position decode from vertex buffer (only affects nearby tiles)
- Cache recently used triangles in a form that makes future picks faster, maybe by pre-calculating some of the math from `IntersectionTests.rayTriangleParametric`.
- Unroll picking math

[...]

PR #9961, opened 2021-12-04
Fast picking with an octree 🎉

DanielLeone:
Hi 😊 

(At this stage, I'd call this PR a working proof of concept)
(The code isn't quite production ready, but it's fully functional, and I think the design is open for further discussion)

(Oh, and all this code wasn't written by me, the original code and idea was from Ian :) )

# What is this?

* An in memory octree data structure for terrain mesh's to speed up Globe picking calls by an order of magnitude
* The main use case was ArcGIS Tiled Elevation Terrain, which is currently quite slow (in main branch, we iterate through every triangle in a tile, every frame, and ArcGIS has ~100k triangles  👀)
    * However, since I ended up going with an octree structure, this is also used for Cesium World Terrain, and
      we get a nice speed up there too.

# Pics

I created a new Sandcastle page to help debug this, you can go check it out if you're interested!

http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FArcGIS%20Tiled%20Elevation%20Terrain%20Debug.html&label=Development

Click anywhere on the globe to draw a Ray, and then use the options on the left to show/hide the triangles/octree layers for the ray you've clicked.

Drawing the Octree at Layer 2 (well, 3rd layer, currently the deepest layer). Also drawing all triangles by their respective octree node.
img

You can turn on showing only the tested triangles as well. So currently in main branch, this would be every triangle. compared to this branch, only a subset is tested (orange triangles were tested)

img_1

In terms of speed-up for ArcGIS terrain, I'm getting a consistent order or magnitude

- ~2.89ms (left/green) Octree
- ~30ms (right/pink) Original

img_2

The same Sandcastle page works for Cesium World Terrain as well (first dropdown option)

img_3

and the speed-up is similar for CWT:

- ~0.29ms (left/green) Octree
- ~5ms (right/pink) Original

img_4

You can also disable the default pick strategy to feeeel the speed-up :)

img_5

How I was testing/making optimizations:

* I added two node.js scripts to `package.json` for debugging purposes
* `run_terrain_trace` runs a puppeteer script pointing at a pre-built sandcastle which just waits for all terrain to load
    * it records all the chrome debug logs as well
* `parse_traces` goes through the trace and log files recorded by `run_terrain_trace` and prints out some statistics, and
  relevant logs if you call it with `--logs`.
  
For comparison purposes, each ArcGIS terrain tile was taking about ~75ms to process (on the worker thread), and the new octree (and triangle packing) functions were contributing about 27ms of that.

If you run the two node scripts in that order, you should get a print out with similar relative results:

| section of code | mean duration |
| ----------- | ----------- |
| computeVertices | 74ms |
| making packed triangles | 14ms |
| creating actual octree | 13ms |

There's plenty of room for improvement in here as well (and a couple of Chrome deopts in the logs as well we could probably avoid). Again, not production ready code yet. Working proof of concept.

# Problemos

* The recent feature added to support on-the-fly updates to terrain exaggeration, basically makes this PR useless :) Damn.
    * This PR was relying on terrain exaggeration being known beforehand, so the mesh/octree could be created using that
      value.
    * In this PR, if terrain exaggeration is set to anything other than 1, the octree is not used (and that does work
      fine).
        * I guess we'd be playing a game of heuristics now, if we went with both approaches.
        * Is the user more likely to change terrain exaggeration (no point in generating the octree's)
        * or is the user unlikely to change exaggeration (do some optimizations)

### What do you want? 

If anyone has a chance to have a look, let me know what you think. 
My main concern is the conflict with the on-the-fly terrain exaggeration which this doesn't work for.
But at least for my use case at work, we don't use terrain exaggeration, so this would be a very welcome addition for us.

Worst case, the two algorithms I ended up with I think are nice, I'm sure they'll have some utility some day for quick octree generation and searching, even if this isn't the spot or language for them right now.

Here's a nice comparison turning the default pick strategy off at run time, and bringing us well within Chrome's frame budget while using ArcGIS terrain :) 

Screen Shot 2021-12-04 at 10 42 32 pm

Thanks for reading!


cesium-concierge:
Thank you so much for the pull request @DanielLeone! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/DanielLeone/cesium/blob/fastPickingWithOctree/CONTRIBUTORS.md) file!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fastPickingWithOctree/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
This looks great @DanielLeone! And a long time in the making too. We occasionally get high res terrain that suffers from horrible picking performance so I'm glad to see support for more than just ArcGIS terrain.

It would be great to get this into an upcoming release, maybe Feb 1st. @IanLilleyT would you have a chance to review?

IanLilleyT:
I started trying this out and the performance is really good. I didn't go too deep with code review because of the big question you brought up of what to do about dynamic terrain exaggeration. Sorry for merging the dynamic terrain exaggeration PR first :stuck_out_tongue:

As you pointed out, dynamic terrain exaggeration makes this a lot more difficult because the octree "bakes" the triangles into the cells of the axis aligned box, and when the axis aligned box is exaggerated it's no longer a box shape, but curved. So now all the triangles that belong to an axis aligned cell in the unexaggerated version map to a curved cell in the exaggerated version. Maybe it would work to raycast against curved boxes, but I feel like that would be a significant performance hit from all the extra math. A lot of axis aligned optimizations would go away when exaggeration is turned on.

Then I started wondering what would happen if the exaggerated curved box is converted back to an exaggerated axis aligned box. If we could find all the curved cells that overlap the axis aligned cell and loop over their triangles it should be possible to intersect the right triangle. The downside is there needs to be some sort of overlap check between axis aligned cells and curved cells. This could be made faster by converting the curved cells to oriented bounding boxes but it would mean even more false positives. Plus, rectangle to oriented bounding box is not the cheapest operation. Chances are this conversion would have to happen whenever the exaggeration changes. It would be nice if there were a shortcut like: the axis aligned cell overlaps at most the nearest 8 curved cells. Then the octree traversal would have to check the 8 curved cells. Still cheaper than checking all cells at least. I'm not sure if such a shortcut exists.

[Here's a sandcastle showing the axis aligned and curved cells](https://sandcastle.cesium.com/#c=1Rppc9u49a9w/GWphKFFHbYkK546rtPsTLzOOOnuJnWnA4mwxJoiNTxs0Tv+730AiIuEbNLStlN/sAjgXXgXHo57lFj3AX7AifXeivCDdY7TIF+5v9I+++ZgTtvncZShIMLJzYFj/XETWdY6if+N51kQR1+C+R1OJlaW5NixbqKnzslNdBPdl5QvYx+HQJxiPeA0m1jvvF7X7TqkAyPaMeTtNM6z5cQS41GckPaAt1dB9AkHiyUgDbvkj/ejDe/3uuoA3qDFAieISErGOJ91GGQpMGKtZfxwlQQ4yrD/Id7gtJwNHzvbBOlZGCwi8/A1XgD1GlKSxA+y84lqpVTvXRTP72Cqbpag+Z0t9ERURxSXxXE4Q8QmfjzPVyCYu8DZRYjJ54fiZx8sU8LcHBCkKl20XofFhyDyg2iRSvoOp0xwbuPEsgm3CK2wFUTSXB1mreDWkqjuEqVXD9GXJF7jJCtsgtQpAS2ryh+kvZqlOLlHsxCr/Cmam+azdJ4EM2znax9lmIhjWU+gJeo61PncdI4j7M4BIUFuijPik3bJzwdHCiJqVWuiOu45SjL4QlHf9kbHQ2/UcwfD/vGoP+oPHevdYOCNh0eDgdsdD8beeNyDvl7PG42842N30B+PR92j8XHHYVxi6hMGLp8wIpr9EmTz5XUchjb41bjf6/bH/dFocAS0BwOg7LmD4/Gwe9Tt9o/G/ePhCDqPXG80HvSHntc7HnaHA68jQobpwmaN2zyi4WXxXqZqTTfrJFgFWXCPUzfBq/gen4EoDB0gwbIk3sCJfslXM4hmaUvSz3ROoEgQmqBIv4SioWkCowMSjoasCY4OKFyVyDRyV8YllkgAJhQxqMDzxGCE54PKLGliME6TjiiQ1aQBSAp0dVTDqyaUGmoVQMMu800NqezXOdEsVKdPuykkj3Rmtymzswhsmo3wak2WBwJwwrq5kSlw2ccdhIDzeOaiJLBWoGgRYhgvY+iad7m3Sbz6K14kGKc28UyHkXKoYzpWxW+EkS90B5Ke8UbzrborVBGFixgQuXak0jTV4DAM1mkc+KWGRWguwniGXTF8IlHWcViEsJaeQ95gK6i+9H6pjdudEl1k7A2gdE/gZ/q+9FhovH0rRFNACwZaqKCFBsrECqkcwk40RVhvLXtjHZZ4HdCPMh74wvKCBso0GswjgEixhciS6r1CZd2VTiLTOfWSa0i7KEptENYh3BxpdUfaolMl6LUhuMUbttCvW9NFvm8rymXVgFIcSNQ0yFjh8I911wEx/6mNUw2TmkXrXcFikAQonPAZXZYddD7fijW2VXCrBsb9i3ozgXd0+D/0pmXN4zBO9EWW9JA1z7HoP499DTsVUk9qUx18kiosYcqfbcsb0Wld1aJs0OO0zIPCxRMc+ThhvbZwPs13hNWd+jw9MU/41+GSv0j0QnMeY/JpyKw2QW1hMU7zKw02m8WcY71y0to/48wb8NlBDyb+paNoebmyWGoZOp7NZPRLxeW0NAZwlga43CJ25Ex4j1Sc6OFzYR288BJ8K+vMPmXQtFeTRx2tysasB0wrpgN5n3VHEeO6GxgJNfbxClG+ARC17y4uTdMWr5p5ZfcR9ltgCuArViS99PscR+fLPLqz3isI5nVP4qDseRxlmRNITEwTni2XoHeWVtISAmJhT8siAH6nYmVP1SpAgpZFAPwqoIW5YEgfS+BHFfiRAjNQfXWlIL+xjUatfADh3uiKPTFgXyAztg3oby2v04TG17L6rFcfMOs3uqFM+L+UFa2hegECmhTPULlW6lzF72VsC205itiOVIOjSCODVuNxqWyBZNH7lhjsTc23TEJeKnsilYANFNSJmsnIL5EBPmrpS1eEUxHaqcjAA/XEMlUEWzLClvTVIuQ5CPqEwlul4IQyKQk2fXKAAckqX0U2UHaXAHQGKwskK2fLaUNH7lAoTa85Ta8pzV5zmr1naQqq5lJ7lYdZQA6Qiq9zFKLEZloCqm6j+XstqHqNqfZaUO29TFVzkvM4iegZaJ3BHPYGmGp3Ttbt5AVRXyKX5jNy6JfZCqBjcQUrnTuT8/ZLrlcjp69n1PWC9Pd2HiVCtIEHaHy+t/OxV/P50c7rmvLReLXR22t01kZf3ivpN9VTYx0ZrEEPyhSUMgnaim8rgJVMmYpMqQHSlK57sKOT2o26V6H+fa/UexXqP4zUmbv9TytHvQyJb29hSg1dXoaIY9GCEDYML3hmldP3lpyIkYpXcfrRkhMx2GMLToZy7ooyfmbt0nJ5Ax4vUSWHMQqMw83pqJh7IPh93wR/1Aia1EkPDBQ1GTbsOnFjxGkFss0pv7ba1Svs1nvfVx2CVI8fqgcg2vEHm7acsFKPn2ydlW2s2tG+y/F9l+J7K8NlGYC6nzE5I4KiFXZfhgSCFlGQ5T6GAkCdWWMsT5W/MVZP3yzArnRljLcIBlAYPGJanzSwSBNKzWzRhFKzzZAfrHCUljdrZnhqJofp3WGKlLKw7QHgyr2CZuNZvKEXcMxnTJXMoIw8pml3UwYaU5ds9rQm4yRHGW6h4xY6bqHjFjruo477qOM+6ri8yV9WmL9g9y2zhKKQqzwjR/p/w/EKZ0khDfmhNsauCYWNxEsA0TNRvinbJ9W6ktnPUQp5bV45ouFs+Cinvyj7JwZxeSKUVp1UrMxVl0Fjlmfk6Ygohcp7FfVotCrEGcejk2enpyxdloSfKhN95vqkvLhVLhr5uLytWlT4pxOD3sR5M1qvMUpIl3Y79AUnHJRKfCbA1Gux2xBltUsxtjJ8zVCGJ/pFFBHgN3YhBppdkqt+m+7wtSnDkhSs8tVZGKAU+585jnb1JL6fhBI73De3nz63X3MFwmv9zXR+yLGkJgWbieG+4B5D6tp8JJkwm7xsH/fXi+tvF7//6+PV9eXZN0EFb8BMPvb5C6v6TciSjwh1cM3+/wdAzY778X8ozcFFKxFQc0mSLMEpD5yDaZoVIT5lw38JVus4yaw8CW3XPSQvLiCccHo4y+d3OHPnKa1Dp4ccaeoH91bgvzc86ANJUJrCyG0ehl9hvbw5OJ0eAryGFsb0xdMVeFSICgKy9E4/s07XdaeH0KxjiTdqpDcjb8FOp9ks9otyHtMsOZUqmWb+KTkanx7Ch949DaJ1nllZscZAFpS5ADGJI0LrnTfqkgbaQIN9pxlekwZ8+ihD72ZBRMS5R2EOtmJvTGjj7/Rx1cT6idL/iU5dMIfPZKug5LD+zxaUvX/ZUVB6w9BW0rEUdPyynOVznR0FpXcef66g9CnRzoJeBpHF0lxLaRXrs0eqUmAolbbIrKw2u8qNNv9NueXSuGuoKdfJbSVn4jHhyaviJpLrz352DT96DtZSbE9R+MvRV16B7SroMn4oX7xin5R9+GWp50s8vwPQimC0myxutccqbeWBei61EHvbsUeZqi9G2oqVsBdGe5GlfK3UWjP0MeJ+tEFJGQWAH7pcwy9dvUkPrO7/AQ)

Bottom = not exaggerated
Top = exaggerated
Magenta = axis aligned box
Red = curved cells
White = oriented box cells

Note how the two green dots are at the same longitude and latitude, but different heights. This puts the exaggerated green cell in a different axis aligned cell. But by checking all the curved cells that overlap the axis aligned cell (which is pretty much every curved cell in the picture), it should recover the right triangle.

Screenshot from 2022-01-18 16-52-26-dots

IanLilleyT:
Right now I'm thinking the best chance of this working is by converting the exaggerated node's box to a region and finding out which cells from the exaggerated region overlap it. This would return the cells that have triangles that need to be checked. This overlap checks boils down to a region <-> region check, so it should be fast. The main slow part will be `BoxToRegion` and `RegionToBox` (`BoxToRegion` doesn't exist in CesiumJS yet), but it might be fast if tiles are small (things get weird when regions wrap over halfway around the globe). All of this would be done when exaggeration changes and the results would be saved in the nodes. During traversal, each node has a list of overlapping region nodes which contain the triangles that need to be checked.

Pseudocode:

When exaggeration changes:
```
fullBoxOrig = min/max positions
fullRegionOrig = BoxToRegion(fullBoxOrig)
fullRegionExag = Exaggerate(fullRegionOrig)
fullBoxExag = RegionToBox(fullRegionExag)

for level in octree
    for node in level
        node.box = subdivide(fullBoxExag, node.coords)
        node.region = BoxToRegion(node.box)
        node.overlappingNodes[] = getOverlappingNodes(node.region, fullRegionExag, node.coords)
```

Raycast:
```
nodeExag = intersect(ray, fullRegionExag)
for nodeOverlap in node.overlappingNodes
    for triangle in nodeOverlap.triangles
        checkIntersection(exaggerate(triangle))
```
 
___
Overall I'm trying to avoid the "check every triangle" slow path when exaggeration is on. If there's no good outcome we could probably merge that. But it will be pretty jarring for the user so I can't guarantee it would be merged.

DanielLeone:
Hey @IanLilleyT thanks for having a look at this. Apologies for the late reply, GitHub notification settings are hard 😞

Yeah I'm happy to slowly put more time into this to see if we can get something working for terrain exaggeration too. Most of this goes straight over my head though! 😛 😢  

I need more time to think about the ideas you've put down, thanks for the sandcastle as well!

DanielLeone:
Hey guys, evidently, I unfortunately haven't been able to prioritize working on fixing the remaining issues in this PR to get it working for exaggerated terrain.

Ideas beyond what Ian was already suggesting above:
* extend the Cesium API enough to let me implement this in "userland"; the API changes probably aren't too big.
* move the octree generation onto the GPU as a lazily built thing, so as terrain exaggeration changes, rebuild the octree each time, and use it on the next frame, or something.
  * I don't think this is infeasible, the octree is a fixed size structure, easily serializable into a byte array
  * at 3 layers deep, it'd mean that at most each triangle could be in 64 nodes of the tree, so only 64 bits per triangle would be required to store the octree in a flat buffer).
  * We'd lose some efficiency by building the octree in parallel, but without testing, it might be fine since it'd be in parallel off the main thread anyway. I don't think we've got a "compute shader" per triangle though - I haven't looked too much - just an idea.
* make the feature opt-in only, so you don't unexpectedly gain/lose the benefit if you're using terrain exaggeration.

If none of these are feasible - I'll probably close the PR in a couple of months, it was fun working on it :)

cesium-concierge:
Thanks again for your contribution @DanielLeone! 

malaretv:
Is addressing this issue on the roadmap for CesiumJS team? (Is there a public roadmap available?)

It is a fairly serious performance kneecap when using non-cesium-ion terrain providers in combination with `clampToGround` entities.

I am not skilled enough in the graphics space to push @DanielLeone 's octtree pull request forward -- but if it would be acceptable to the Cesium team to modify the branch to make his approach opt-in, I would be willing to try my hand at putting that together

calogeromauceri:
The changes in [#9961](https://github.com/CesiumGS/cesium/pull/9961) look great! This would bring a significant performance boost for our high-resolution terrain.

We're not using dynamic terrain exaggeration, so making these changes opt-in sounds like a good approach. I'd be happy to help with making them configurable and ensuring they work with the latest Cesium code. Let me know how I can contribute!

ggetz:
Discussed potential paths forward for this PR in https://github.com/CesiumGS/cesium/issues/8481#issuecomment-2667038250. Would love to hear input!

mzschwartz5:
@DanielLeone I pulled your branch down locally and I'm getting it up to date with main. I wanted to try out the sandcastle you created and mention in the PR description, but it doesn't look like it was committed (or maybe it's been deleted since?). 

If you still have a copy of that, could you add it please?

malaretv:
> @DanielLeone I pulled your branch down locally and I'm getting it up to date with main. I wanted to try out the sandcastle you created and mention in the PR description, but it doesn't look like it was committed (or maybe it's been deleted since?).
> 
> If you still have a copy of that, could you add it please?

Hi @mzschwartz5!  it was I who removed the tracing hooks and debug sandcastle in one of my commits to prepare this branch for merging into the project, it's all here:
https://github.com/CesiumGS/cesium/pull/9961/commits/6c5c2134f0ca76e01bb0af669439b9a9b4b9e98e

mzschwartz5:
I've been reworking the octree implementation in a major way because:
1. Building the octree, and even picking from it (with ArcGIS terrain) is still pretty slow. For ArcGIS, the workers can take upwards of 50ms per tile to build a 3-layer deep octree (64 leaf nodes). And for all that, picking can still takes upwards of ~20ms (e.g greater than an entire frame's budget).
2. I haven't measured, but I suspect the added memory consumption of the octree is relatively high. It's basically keeping a duplicate of all the triangle indices of all rendered tiles. (Technically more than a duplicate, since triangles  can be (and frequently _are_ in) multiple tree nodes simultaneously).

What if, instead of building the octree on tile load, we build it as needed: when picking happens. The idea is: on pick, schedule a worker to build the octree and, in the meantime, use the fallback strategy of iterative triangle testing (the current strategy). Then, on subsequent picks, once the octree is built, picking runs faster. Thus, octrees are only built for tiles that are picked against, not for every tile every time it loads in. An added benefit of this approach is that we can dirty octrees whenever vertical exaggeration changes, and rebuild them next pick, enabling octree picking to work with dynamic terrain exaggeration (without having to rebuild _everything_, all the time).

This is the approach I had been taking, _however_, I ran into a roadblock. In order for a worker to build the octree, we need to give it access to the tile's vertices. We have three options here:
1. Transfer the vertices to the worker: zero-copy, which is ideal, but the main thread cannot use the vertices for the fallback triangle testing approach in the meantime. 
2. Copy the vertices to the worker: effective, but slow and would defeat some of the performance gains here.
3. Use a shared buffer array: allows zero-copy without any transfer of ownership. Unfortunately, it requires cross origin isolation, which is not enabled on most web apps.

In light of this, here is the path forward that I'm considering, and it doesn't even involve worker threads at all! There are two key insights: the first is that picking tends to happen repeatedly against the same tiles (e.g. the camera picks against the globe every frame. Another example: several entities getting clamped to ground in the same region). The second insight is that the current picking strategy is identical to the octree picking strategy in the case where the octree only has one level: a root node. 

So what if we _start_ with just a one-level octree, and each time we pick from it, we use the results of the triangle intersection tests to _incrementally_ build up the octree with each pick:

1. Tiles load in
2. A pick is performed against Tile X.
3. Tile X's octree picker scans triangles in level 0 (aka all triangles in Tile X), and puts them into level 1 child nodes. It also returns the picked triangle.
4. Another pick is performed against Tile X.
5. Tile X's octree picker finds all intersected leaf nodes in level 1, scans their triangles (until an intersection is found in a leaf node), and puts those triangles into level 2 child nodes... etc.

Pros of this method compared to original octree implementation:
- Lower memory consumption (only build octree for picked tiles, and only build up subnodes of picked octree nodes)
- Less time spent building octrees for the same reasons 
- Faster picking due to deeper trees.
- Maintain ability to support dynamic terrain exaggeration.

Cons:
- Warm-up time: there may still be some lag on ArcGIS terrain while moving the camera around as it takes a few picks for the octree to build up. In the original octree implementation, octrees were always being built in the background. That was more work done total, but could be done while otherwise idling.

And honestly, I'm not sure yet if this warm-up time cost is worth it... there are a lot of pros, but if the user pans around the globe, they may not really feel those benefits.

mzschwartz5:
> here is the path forward that I'm considering, and it doesn't even involve worker threads at all

Alternatively... there may be a happy medium solution where, say, we _do_ build the trees on tile load in worker threads - but _only_ to one level of depth. That helps warm-start the iterative tree building process without costing too much worker time.

However, this approach wouldn't work so easily with vertical exaggeration... when exaggeration changes, trees would have to be rebuilt with no warm start.

mzschwartz5:
Noticed (and documented) [this relevant issue](https://github.com/CesiumGS/cesium/issues/13003) while testing on ArcGIS terrain. It doesn't affect the overall approach, but it does affect the precision and (maybe?) the speed of picking.

Beilinson:
Has anyone considered using a quad tree instead of an oct tree?

It should have the benefit of not needing to be rebuilt with vertical exaggeration changes, being faster to generate, and should still bring the same performance benefits in most use cases because most picking use cases (camera/billboard clamping) are for straight down rays, which in the oct tree case ends up working as a quad tree anyways

Terrain providers are quad tree, this PR is essentially creating lower LOD versions of higher Z level terrain meshes by splitting the tile without refining the 4 new tiles by adding any additional triangles

mzschwartz5:
It's funny that you mentioned that - I'm literally in the middle of a refactor to use a quadtree. Took me a little while to realize that was the obvious choice... I guess because I inherited this PR literally titled "...octree". 

However, a lot of the work I've done is still needed - it's true that it doesn't need to be rebuilt with exaggeration changes, but I believe it still needs to be rebuilt on scene mode changing; so the shift to build the tree incrementally is still useful there (as well as for the other reasons listed (memory consumption, worker performance, etc.)

mzschwartz5:
I haven't fully thought it through, but it's even possible vertical exaggeration still triggers rebuilds of a quad tree. To determine the intersection of a ray and a quad tile, you would use the quad tile's bounding volume. Those bounding volumes would have to change on exaggeration changes.

But it may be possible to at least keep the same triangles in each tree node... I'm not sure yet if that's just an approximation, or if it's exact. (I think the former... but maybe accurate enough?)

edit- actually, even if - theoretically - the triangles are in the same tree nodes, I believe Cesium rebuilds the vertex / index buffers on exaggeration changes. So either way we have to rebuild the tree using these new verts to be safe.

packages/engine/Source/Core/TerrainMesh.js
```diff
@@ -231,25 +237,89 @@ function computeTransform(mesh, result) {
...
+  const centerX = southwest.x + dx * 0.5;
+  const centerY = southwest.y + dy * 0.5;
+  const centerZ = exaggeratedMinHeight + dz * 0.5;
+
+  Cartesian3.fromElements(dx, dy, dz, scratchScale2D);
+  Cartesian3.fromElements(centerX, centerY, centerZ, scratchCenter2D);
+
+  Matrix4.fromTranslation(scratchCenter2D, result);
+  Matrix4.setScale(result, scratchScale2D, result);
+  Matrix4.multiply(Transforms.swizzleMatrix, result, result);
```
javagl: Oh... I'm occasionally looking into things like this (those GitHub notifications are like ... oh, look, a squirrel 🐿️ ... where was I...?). I don't know what this swizzle is, where it comes from, and why it has to be done, but ...I think the _same_ could be done by directly filling the "scratch" variables above with the `z, x, y` order, right?

mzschwartz5: I'm sure there's a way to fill in the scratch variables such that it just "works" without the swizzle, but it's not simply changing the order to `z, x, y` (I thought so too, but it didn't work. Then I took a look at `Transforms.js` and saw the use of the swizzle matrix in `basisTo2D` and `ellipsoidTo2DModelMatrix` and thought "that looks handy")

Luckily, the transform is updated lazily, not frequently, so I'm not super concerned about the extra swizzle.

mzschwartz5: I'll also add: I tried using `basisTo2D` on the 3D transform, but it wasn't working because the 3D transform included scale, and the `basisTo2D` function doesn't really work with scale; it just carries through whatever is in the upper-left 3x3 matrix to the end result instead of adjusting or projecting it in any way.

javagl: I've read the documentation of the `swizzleMatrix`, and of all functions that _use_ it (😏 ) and hoped that it could be possible to derive what is happening there (and why), but ... given that this was just a short distraction, I accepted my level of understanding, namely "The `swizzleMatrix` is a matrix that swizzles". 

When you suggest that just changing the order didn't work, then I'm (more) curious again. I'm reasonably sure that it indeed _should_ only swizzle...

https://sandcastle.cesium.com/index.html#c=jZHNboMwEIRfxfKJSsgE6K0EVcq1VQ+9+uLANrWy2Mg2SeHpC2XTQn/UyIeV95vZHcvSVNb4wPxZDwPCowpOv7EtM3BmO/C6a8Tcu42kYWwjNvGypv/c/6rX6tb6mztpJAVurkx62ZRRzT8np8RT4inxlHhGPCOeEc+I58Rz4jnx/FtSB77D8HvSWbduimaU6xb7aPUpMT05poFfKyyCQHuIJJ+J5BNbkqVjPjzmhQ89QjmFvtdNa11gncNIiCRA06IK4JN9Vx0hiMr7yVwkF0tR6xPT9Vby6iP7zpqgtAEnOatQeT+Slw7xWQ8geVkko35lQ6tqbQ5PJ3Co+knympYPc1MIUSTj9acrWIt75RYT3wE

But I think that I mixed up the operation - the swizzle matrix is using the constructor (row-major), and not the actual order (column-major), so the matrix is swizzling _rows_, and not columns.

(This is not sooo much about performance. It is primarily about understanding what's going on. But **iff** this "swizzling" is required, and it's not clear how often the functions are called that are doing it, then it may be worthwhile to at least _consider_ to have some `Matrix4.swizzleRows(m, 2, 0, 1, 3, r)` function that does not do n<sup>3</sup> multiplications, but 0 (zero))

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
mzschwartz5, line 1: I added a new argument to the `TerrainMesh` constructor - the terrain tile's rectangle. Several changed files in this PR are simply a result of passing in this argument at existing call sites.

packages/engine/Source/Core/TerrainMesh.js
mzschwartz5, line 1: Changes in this file:
- Addition of a `TerrainPicker` (new class in this PR)
- Picking has been pushed down from `GlobeSurfaceTile` to `TerrainMesh` to automatically defer picking to the right mesh (the "fill" mesh or the actual mesh)
- Computation of transforms (2D/CV and 3D) for the `TerrainPicker`
- Responds to changes in vertical exaggeration / scene mode

packages/engine/Source/Core/Transforms.js
```diff
@@ -1149,6 +1149,7 @@ const swizzleMatrix = new Matrix4(
   0.0,
   1.0,
 );
+Transforms.swizzleMatrix = swizzleMatrix;
```
mzschwartz5: This wasn't exported before. I'm using it in `TerrainMesh` now, so I exported it.

packages/engine/Source/Scene/Globe.js
```diff
@@ -943,7 +943,9 @@ Globe.prototype.getHeight = function (cartographic) {
 
   const intersection = tile.data.pick(
     ray,
-    undefined,
+    // Globe height is the same at a given cartographic regardless of the scene mode,
+    // but the ray is constructed via a surface normal (which assumes 3D), so pick in 3D mode.
+    SceneMode.SCENE3D,
```
mzschwartz5, line 948: This was implicitly using 3D before (via passing in `undefined`). 

`globe.getHeight` still needs some work though. It will pick in 3D mode even when the user is in 2D/CV mode which _works_ (in that it gets the right height), but it bypasses the mechanism for detecting scene mode changes and dirtying/updating the `TerrainMesh`'s transform.

packages/engine/Source/Scene/GlobeSurfaceTile.js
mzschwartz5, line 1: Changes in this file:
- A lot of the picking logic (brute-force intersection tests with triangles) has been moved to `TerrainPicker` and used in the quadtree-building process.
- Added a function similar to `updateExaggeration`, called `updateSceneMode`, to push changes to scene mode to the underlying TerrainMesh

packages/engine/Specs/Core/TerrainPickerSpec.js
mzschwartz5, line 1: Several of these tests were moved from the `GlobeSurfaceTile` spec, as they were really testing the picking interface, which is now part of `TerrainPicker`. I added several tests as well.

I deliberated over whether there should be tests specific to the quadtree building process. Technically, those are implementation details of `TerrainPicker`, so those wouldn't be tested by unit tests per typical testing philosophy. But... they do feel like important details to test. The proper solution may be to split out the quadtree interface, itself, into a separate, reusable module - and create a spec for that.

DanielLeone:
Nice work @mzschwartz5 the incremental building is cool 👍

packages/engine/Source/Core/Transforms.js
lilleyse: Some tweaks to be like others constants (like [`Matrix4.IDENTITY`](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/Matrix4.js#L2965-L2990))
```js
/**
 * An immutable matrix that swaps x, y, z for 2D.
 *
 * @type {Matrix4}
 * @constant
 * @private
 */
Transforms.SWIZZLE_MATRIX = Object.freeze(
  new Matrix4(
    0.0,
    0.0,
    1.0,
    0.0,
    1.0,
    0.0,
    0.0,
    0.0,
    0.0,
    1.0,
    0.0,
    0.0,
    0.0,
    0.0,
    0.0,
    1.0,
  ),
);
```

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+  if (!defined(vertices)) {
+    throw new Error("vertices is required.");
+  }
+  if (!defined(indices)) {
+    throw new Error("indices is required.");
+  }
+  if (!defined(encoding)) {
+    throw new Error("encoding is required.");
+  }
+  //>>includeEnd('debug');
```
lilleyse, line 34: `Check.defined` would make this more concise.

```js
  Check.defined("vertices", vertices);
  Check.defined("indices", indices);
  Check.defined("encoding", encoding);
  //>>includeEnd('debug');
```

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+  // SharedArrayBuffers could be used, but most environments do not support them.
+  Matrix4.inverse(tileTransform, terrainPicker._inverseTransform);
+
+  terrainPicker._needsRebuild = false;
+  const intersectingTriangles = new Uint32Array(
+    terrainPicker._indices.length / 3,
+  );
+  terrainPicker._rootNode.intersectingTriangles = Uint32Array.from(
+    intersectingTriangles.keys(),
+  );
```
lilleyse: An old school for loop would be more efficient here (in particular, less allocations)

```js
const triangleCount = terrainPicker._indices.length / 3;
const intersectingTriangles = new Uint32Array(triangleCount);

for (let i = 0; i < triangleCount; ++i) {
  intersectingTriangles[i] = i;
}

terrainPicker._rootNode.intersectingTriangles = intersectingTriangles;
```

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+  /**
+   * The terrain mesh's vertex buffer.
+   * @type {Float32Array}
+   */
+  this._vertices = vertices;
+  /**
+   * The terrain mesh's index buffer.
+   * @type {Uint32Array}
+   */
+  this._indices = indices;
```
lilleyse, line 45: Should it be `@type {Uint8Array|Uint16Array|Uint32Array}` to match `indices` in `TerrainMesh`?

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+  /**
+   * The axis-aligned bounding box of this node (in the tree's local space).
+   * @type {AxisAlignedBoundingBox}
+   */
+  this.aabb = createAABBForNode(this.x, this.y, this.level);
+  /**
+   * The indices of the triangles that intersect this node.
+   * @type {Uint32Array}
+   */
+  this.intersectingTriangles = [];
```
lilleyse: Should this be `this.intersectingTriangles = new Uint32Array(0)`?

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+/**
+ * Adds a child node to this node.
+ *
+ * @param {number} childIdx The index of the child to add (0-3).
+ * @memberof TerrainPickerNode
+ */
+TerrainPickerNode.prototype.addChild = function (childIdx) {
+  //>>includeStart('debug', pragmas.debug);
+  if (childIdx < 0 || childIdx > 3) {
+    throw new Error("TerrainPickerNode child index must be between 0 and 3.");
```
lilleyse: Would be better to use `DeveloperError`

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+    projection,
+  );
+};
+
+/**
+ * Resets the terrain picker's quadtree structure to just the root node. Done whenever the underlying terrain mesh changes.
+ * @param terrainPicker The terrain picker to reset.
+ * @private
+ */
+function reset(terrainPicker, tileTransform, mode, projection) {
```
lilleyse: `mode` and `projection` don't seem to be used

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+
+  return AxisAlignedBoundingBox.fromCorners(aabbMin, aabbMax);
+}
+
+/**
+ * Packs triangle vertex positions and index into provided buffers, for the worker to process.
+ * (The worker does tests to organize triangles into child nodes of the quadtree.)
+ * @param {Float32Array} trianglePositionsBuffer The buffer to pack triangle vertex positions into.
+ * @param {Uint32Array} triangleIndicesBuffer The buffer to pack triangle indices into.
+ * @param {Cartesian3[]} trianglePosition The triangle's vertex positions.
```
lilleyse: Rename to `trianglePositions`?

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+ * @param {TerrainPicker} terrainPicker The terrain picker.
+ * @param {IntersectingNode[]} intersections The nodes that intersect the ray, along with the intersection intervals along said ray.
+ * @param {Ray} ray The ray to test.
+ * @param {boolean} cullBackFaces Whether to cull back faces.
+ * @param {SceneMode} mode The scene mode (2D/3D/Columbus View).
+ * @param {MapProjection} projection The map projection.
+ * @returns The closest point in world space, or undefined if no intersection.
+ * @private
+ */
+function getClosestPointInNode(
```
lilleyse: Should the function name and description be tweaked? Since it's checking multiple node intersections, not just a single node.

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -0,0 +1,610 @@
...
+  node,
+  triangleIndices,
+  trianglePositions,
+) {
+  node.buildingChildren = true;
+
+  // Prepare data to be sent to a worker
+  const matrixArray = new Array(16);
+  Matrix4.pack(inverseTransform, matrixArray, 0);
+  const inverseTransformPacked = new Float64Array(matrixArray);
```
lilleyse: I think this can be simplified to:

```suggestion
  // Prepare data to be sent to a worker
  const inverseTransformPacked = new Float64Array(16);
  Matrix4.pack(inverseTransform, inverseTransformPacked, 0);
```

packages/engine/Source/Core/Transforms.js
lilleyse: Maybe even rename to `SWIZZLE_3D_TO_2D_MATRIX` so the purpose is clearer.

packages/engine/Source/Core/TerrainPicker.js
mzschwartz5: Yes, I think so- good point.

packages/engine/Source/Core/TerrainPicker.js
mzschwartz5: Yes, good catch

packages/engine/Source/Core/TerrainPicker.js
mzschwartz5: Good point! They were at one time, when I was recalculating transforms in the picker itself. Then I moved that logic into the TerrainMesh class.

packages/engine/Source/Core/TerrainPicker.js
mzschwartz5: Yeah that's very fair. Is `findClosestPointInClosestNode` too verbose? I'm fine with it, personally

lilleyse: I'm fine with the more verbose name

mzschwartz5:
@lilleyse Definitely a good idea to test clamping. I think [the clamping entities sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html?id=clamp-entities-to-ground) is a good test that it's all still working. I'll add it to the testing steps.

Otherwise, I addressed all your points and I'm sending this back for review!

packages/engine/Source/Core/TerrainPicker.js
lilleyse: @mzschwartz5 I saw you renamed `trianglePositionBuffer` and `triangleIndicesBuffer` 

...but actually I was commenting on renaming `trianglePosition` to `trianglePositions`.

lilleyse:
@mzschwartz5 I just had one comment here: https://github.com/CesiumGS/cesium/pull/9961#discussion_r2519660789

Otherwise looks good.

packages/engine/Source/Core/TerrainPicker.js
mzschwartz5: Oh 🤦🏻 my bad, I'll fix that

mzschwartz5:
@lilleyse Fixed that small variable rename issue, back to you

Beilinson:
Drive by comment before merging (very exciting), the `TerrainPicker` as I understand it can be used in-place for the same performance benefits on 3D tiles for example? Possibly providing a significant improvement for #11814 (either in tandem with or instead of the suggested solutions)

mzschwartz5:
@Beilinson Just read through that issue (and a little deeper into some of the linked issues). I think you're right; sounds like this would be tremendously helpful there, as well. So long as 3D tiles have the same sort of data available to construct a `TerrainPicker` instance (which should maybe be renamed if it starts being used for 3D tiles), it shouldn't be that hard to do, either.

It also automatically addresses some of the concerns @javagl wrote about vertical exaggeration.

PR #9963, opened 2021-12-04
Small fixes to match the latest 3D Tiles Next specs

lilleyse:
Some of the code was out of date with the spec changes made over the last couple of months.

* Added `DESCRIPTION` semantic: https://github.com/CesiumGS/3d-tiles/pull/540
* Added `schema.id`: https://github.com/CesiumGS/3d-tiles/pull/539
* Removed `name` and `description` from group and tileset metadata: https://github.com/CesiumGS/3d-tiles/pull/524
* Horizon occlusion point semantic is now a `VEC3`
* Fixed types that should have been `componentType`

### DIFF
--- a/Specs/Scene/GltfFeatureMetadataLoaderSpec.js
+++ b/Specs/Scene/GltfFeatureMetadataLoaderSpec.js
@@ -166,7 +166,7 @@ describe(
       }).toThrowDeveloperError();
     });
 
-    it("throws if neither extension nor extension is defined", function () {
+    it("throws if neither extension nor extensionLegacy is defined", function () {
       expect(function () {
         return new GltfFeatureMetadataLoader({
           gltf: gltf,
--- a/Specs/Scene/GltfFeatureMetadataLoaderSpec.js
+++ b/Specs/Scene/GltfFeatureMetadataLoaderSpec.js
@@ -65,7 +65,7 @@ describe(
         ortho: {
           properties: {
             vegetation: {
-              type: "UINT8",
+              componentType: "UINT8",
               normalized: true,
             },
           },
--- a/Specs/Scene/MetadataClassPropertySpec.js
+++ b/Specs/Scene/MetadataClassPropertySpec.js
@@ -103,7 +103,7 @@ describe("Scene/MetadataClassProperty", function () {
       property: {
         name: "Population",
         description: "Population (thousands)",
-        type: "INT32",
+        componentType: "INT32",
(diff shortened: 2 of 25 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/small-3d-tiles-next-fixes/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/parseFeatureMetadataLegacySpec.js
```diff
@@ -16,17 +16,17 @@ describe(
         building: {
           properties: {
             name: {
-              type: "STRING",
+              componentType: "STRING",
```
ptrgags, line 19: I don't think this change is correct, this file is for parsing `EXT_feature_metadata`, not `EXT_mesh_features`. it's one of very few places that should use `type`

Specs/Scene/MetadataClassPropertySpec.js
```diff
@@ -103,7 +103,7 @@ describe("Scene/MetadataClassProperty", function () {
       property: {
         name: "Population",
         description: "Population (thousands)",
-        type: "INT32",
+        componentType: "INT32",
```
ptrgags, line 106: This test is for `EXT_feature_metadata`, it should be `type`. The other change in this file is fine.

Specs/Scene/GltfFeatureMetadataLoaderSpec.js
```diff
@@ -65,7 +65,7 @@ describe(
         ortho: {
           properties: {
             vegetation: {
-              type: "UINT8",
+              componentType: "UINT8",
               normalized: true,
             },
           },
```
ptrgags, line 71: Just noticed something in this file (not in the diff): 

one of the tests should be renamed to be "throws if neither extension nor [extensionLegacy] is defined"

ptrgags (commented):
It seems I definitely could have been more thorough when I did the type renaming :sweat_smile: . The backwards compatibility for `EXT_feature_metadata` means most of these wouldn't show up as unit test failures.

There were a couple spots where the older `EXT_feature_metadata` schema was intentional. It would be good to put those back, and add some comments explaining the few spots that are intentionally different.

ptrgags:
looks good, thanks @lilleyse!

Issue #6763, opened 2018-07-02
debugShowGlobeDepth and debugShowPickDepth are broken


likangning93:
Using `CesiumInspector`
* Show pick depth doesn't seem to do anything
* Show globe depth draws "clear" instead of the globe depth

These options have been temporarily removed from the `CesiumInspector` UI until they are fixed #7014

mramato:
By dating back to, do you mean these happen in 1.46 as well?

hpinkos:
Globe depth and pick depth broke in 1.45.  Probably related to log depth changes.

The frustum bug is a known issue: #6187

hpinkos:
Hid these options from the UI in #7014

hpinkos:
Related #5965

ebogo1:
The code for debug globe and pick depth was removed in https://github.com/CesiumGS/cesium/pull/9965. Those commits can be reverted if this is fixed.

ggetz:
Closing this. If we get further requests or have another reason to put this options back, we may re-open.

PR #9965, opened 2021-12-07
Remove broken debug depth options in Scene.js

ebogo1:
This PR removes some code that's been broken for a while: https://github.com/CesiumGS/cesium/issues/6763. If we ever fix this code these commits can be reverted, but for now it's helpful to simplify Scene.js, GlobeDepth.js, and PickDepth.js so the Framebuffer code can be refactored.

- Removes `debugGlobeDepthViewportCommand` in `GlobeDepth`
- Removes `debugPickDepthViewportCommand` in `PickDepth`
- Removes `debugShowGlobeDepth` and `debugShowPickDepth` (which was only set through the inspector) options from `Scene`
- Removes `debugGlobeDepths` array in `View`
- Removes globe depth and pick depth toggles in the `CesiumInspector` widget and view model
- Removes a test testing debug globe depth in `SceneSpec`

### DIFF
--- a/Source/Scene/GlobeDepth.js
+++ b/Source/Scene/GlobeDepth.js
@@ -49,7 +49,6 @@ function GlobeDepth() {
   this._useScissorTest = false;
   this._scissorRectangle = undefined;
 
-  this._useLogDepth = undefined;
   this._useHdr = undefined;
   this._clearGlobeDepth = undefined;
 }
--- a/Source/Scene/GlobeDepth.js
+++ b/Source/Scene/GlobeDepth.js
@@ -8,7 +8,6 @@ import Framebuffer from "../Renderer/Framebuffer.js";
 import PixelDatatype from "../Renderer/PixelDatatype.js";
 import RenderState from "../Renderer/RenderState.js";
 import Sampler from "../Renderer/Sampler.js";
-import ShaderSource from "../Renderer/ShaderSource.js";
 import Texture from "../Renderer/Texture.js";
 import PassThrough from "../Shaders/PostProcessStages/PassThrough.js";
 import PassThroughDepth from "../Shaders/PostProcessStages/PassThroughDepth.js";
--- a/Source/Scene/GlobeDepth.js
+++ b/Source/Scene/GlobeDepth.js
@@ -53,8 +53,6 @@ function GlobeDepth() {
   this._useLogDepth = undefined;
   this._useHdr = undefined;
   this._clearGlobeDepth = undefined;
-
-  this._debugGlobeDepthViewportCommand = undefined;
 }
 
 Object.defineProperties(GlobeDepth.prototype, {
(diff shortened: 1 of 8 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/remove-debug-depths/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/GlobeDepth.js
```diff
@@ -70,47 +67,6 @@ Object.defineProperties(GlobeDepth.prototype, {
...
-        uniformMap: {
-          u_depthTexture: function () {
-            return globeDepth._globeDepthTexture;
-          },
-        },
-        owner: globeDepth,
-      }
-    );
-
-    globeDepth._useLogDepth = useLogDepth;
```
lilleyse, line 108: `this._useLogDepth` can be removed

Source/Scene/PickDepth.js
```diff
@@ -19,49 +18,6 @@ function PickDepth() {
...
-        uniformMap: {
-          u_texture: function () {
-            return pickDepth._depthTexture;
-          },
-        },
-        owner: pickDepth,
-      }
-    );
-
-    pickDepth._useLogDepth = useLogDepth;
```
lilleyse, line 61: Same comment here

Source/Scene/Scene.js
```diff
@@ -471,18 +470,6 @@ function Scene(options) {
...
-   * This property is for debugging only; it is not for production use.
-   * <p>
-   * Displays depth information for the indicated frustum.
-   * </p>
-   *
-   * @type Boolean
-   *
-   * @default false
-   */
-  this.debugShowGlobeDepth = false;
```
lilleyse, line 484: Since this property is part of the public API removing it counts as a breaking change and should be mentioned in `CHANGES.md`. I don't think we need a deprecation period since it is for debugging only and has been broken for a while now.

Source/Widgets/CesiumInspector/CesiumInspectorViewModel.js
```diff
@@ -156,20 +156,6 @@ function CesiumInspectorViewModel(scene, performanceContainer) {
...
-   */
-  this.globeDepth = false;
-
-  /**
-   * Gets or sets the show pick depth state.  This property is observable.
-   * @type {Boolean}
-   * @default false
-   */
-  this.pickDepth = false;
-
```
lilleyse, line 172: Same comment about mentioning both of these in the breaking changes section.

ebogo1:
@lilleyse I made the changes above. Should the CHANGES.md entry have any more info? The original issue says the debug depths may have been broken since 1.45, but usually any extra info in the Breaking section is for alternative APIs to use after deprecation.

lilleyse:
That should be fine. Thanks @ebogo1.

Issue #9900, opened 2021-10-28
Implement MSAA with WebGL 2


ebogo1:
CesiumJS has noticeable aliasing artifacts, especially with thin geometry (see https://github.com/CesiumGS/cesium/issues/8228). With multisampled render targets in WebGL 2, MSAA is a good step towards fixing them. 

To implement MSAA in CesiumJS, we need to decide which render targets will need to get multisampled to avoid artifacts. One idea was to start with the color framebuffer in [GlobeDepth.js](https://github.com/CesiumGS/cesium/blob/main/Source/Scene/GlobeDepth.js). There are also questions about how this would interact with translucency passes, and whether there might need to be additional logic for compute commands to perform MSAA if needed.

I'm opening this issue to have a space to discuss the MSAA roadmap and implementation details for CesiumJS.

ebogo1:
## SMAA Screenshot dump

no AA, `viewer.useBrowserRecommendedResolution = false` 
![liberty-noaa](https://user-images.githubusercontent.com/31491650/140558400-b43c8b6d-0eec-444d-8d1c-6c42fc07bb74.png)
SMAA, `viewer.useBrowserRecommendedResolution = false` 
![liberty-smaa](https://user-images.githubusercontent.com/31491650/140558428-758cdf73-926d-4f33-91e3-6a09d403cf32.png)
no AA, `viewer.useBrowserRecommendedResolution = true`
![noaa](https://user-images.githubusercontent.com/31491650/140559731-9e69302b-a8ea-4fe0-b667-d1c288c84f51.png)
SMAA, `viewer.useBrowserRecommendedResolution = true`
![smaa](https://user-images.githubusercontent.com/31491650/140559748-124fd85c-5bbd-4c8b-92a7-3bd14de5920f.png)
FXAA, `viewer.useBrowserRecommendedResolution = true`
![fxaa](https://user-images.githubusercontent.com/31491650/140559768-d214e5f3-1697-4da1-b697-ae1d9bd58b11.png)

no AA, `viewer.useBrowserRecommendedResolution = true`
![lines-noaa](https://user-images.githubusercontent.com/31491650/140561448-4a11e41c-181c-4fdd-b0f0-10ea70309074.png)
SMAA, `viewer.useBrowserRecommendedResolution = true`
![lines-smaa](https://user-images.githubusercontent.com/31491650/140561482-f79604f1-4b9a-47cc-9ecd-94bb7ec45459.png)
[...]

PR #9844, opened 2021-09-29
Fixed rendering issues related to environmentState.renderTranslucentDepthForPick.

ugnelis:
Solves #9824.

I have tested only cases related to transparent entities and point clouds with shading.

It is possible that this should also have to be fixed for the Cesium VR https://github.com/CesiumGS/cesium/blob/main/Source/Scene/Scene.js#L2925-L2927

PR #9719, opened 2021-08-11
Fix eye dome lighting crashes

lilleyse:
[...]

PR #9399, opened 2021-03-02
Add translucent tile classification

ebogo1:
- Adds `TranslucentTileClassification` and specs for handling translucent tile classification
- Adds `DrawCommand. depthForTranslucentClassification ` and logic in `Cesium3DTile` for using it
- Adds handling of translucent classification in `Scene` and `View`

PR #9966, opened 2021-12-07
Add `FramebufferManager` class

ebogo1:
This PR adds the `FramebufferManager` class, which wraps Framebuffer resources (textures, depth stencil textures, renderbuffers, and the framebuffer itself) in a way that can be used by multiple classes that share similar code in constructing their own framebuffers. In [`wrapper-msaa`](https://github.com/CesiumGS/cesium/compare/framebuffer-manager...wrapper-msaa?expand=1), `FramebufferManager` also supports the new `MultisampleFramebuffer` class, which will eventually make it easier to add MSAA to framebuffers that need it. To make the changes easier to see, this also removes the `primitiveColorFramebuffer` from GlobeDepth.js.

I think this should be opened as a draft PR first, because there are a few things I'm still unsure about: 

- ~Some framebuffers are constructed with multiple texture attachments, and `FramebufferManager` currently expects a single attachment for color and depth stencil textures/renderbuffers. It feels clunky to have a `textureOptions` argument to keep track of constructor options in `FramebufferManager.update`, maybe if this is needed then it could go in the constructor instead?~ (resolved by adding constructor options)
- ~Some framebuffers depend on attachments that belong to another framebuffer. I think it's simplest to avoid the `FramebufferManager` class for this case and to construct a `Framebuffer` manually instead.~ (resolved by allowing shared attachments to be added to a `FramebufferManager` manually)
- Should the goal be for every class to use the `FramebufferManager` constructor instead of directly creating `Framebuffer` objects? If not, should the `FramebufferManager` class be tailored to framebuffers that will need MSAA support? 
- Which framebuffers need MSAA? Some candidates are the ones in `PickDepth`, `PickDepthFramebuffer`, `ShadowMap`, `GlobeTranslucencyFramebuffer`, `OIT`, `PointCloudEyeDomeLighting`, and `TranslucentTileClassification`. 
 
The MSAA issue is here - https://github.com/CesiumGS/cesium/issues/9900

----
To-do: 
- [x] JS docs
- [x] Write unit tests

### DIFF
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -17,10 +17,13 @@ import PixelFormat from "../Core/PixelFormat.js";
  * @param {Boolean} [options.color=true] Whether the FramebufferManager will use color attachments.
  * @param {Boolean} [options.depth=false] Whether the FramebufferManager will use depth attachments.
  * @param {Boolean} [options.depthStencil=false] Whether the FramebufferManager will use depth-stencil attachments.
- * @param {Boolean} [options.supportsDepthTexture=false] Whether the FramebufferManager will create a depth texture when the extensions are supported.
+ * @param {Boolean} [options.supportsDepthTexture=false] Whether the FramebufferManager will create a depth texture when the extension is supported.
  * @param {Boolean} [options.createColorAttachments=true] Whether the FramebufferManager will construct its own color attachments.
- * @param {Boolean} [options.createDepthAttachments=true] Whether the FramebufferManager will construct its own depth attachments.*
+ * @param {Boolean} [options.createDepthAttachments=true] Whether the FramebufferManager will construct its own depth attachments.
+ * @param {PixelDatatype} [options.pixelDatatype=undefined] The default pixel datatype to use when creating color attachments.
+ * @param {PixelFormat} [options.pixelFormat=undefined] The default pixel format to use when creating color attachments.
  *
+ * @exception {DeveloperError} Must enable at least one type of framebuffer attachment.
  * @exception {DeveloperError} Cannot have both a depth and depth-stencil attachment.
  *
  * @private
@@ -43,7 +46,7 @@ function FramebufferManager(options) {
   //>>includeStart('debug', pragmas.debug);
   if (!this._color && !this._depth && !this._depthStencil) {
     throw new DeveloperError(
-      "must enable at least one type of framebuffer attachment."
+      "Must enable at least one type of framebuffer attachment."
     );
   }
   if (this._depth && this._depthStencil) {
@@ -62,10 +65,11 @@ function FramebufferManager(options) {
     true
   );
 
+  this._pixelDatatype = options.pixelDatatype;
+  this._pixelFormat = options.pixelFormat;
+
   this._width = undefined;
   this._height = undefined;
-  this._pixelDatatype = undefined;
-  this._pixelFormat = undefined;
 
   this._framebuffer = undefined;
   this._colorTextures = undefined;
@@ -127,14 +131,16 @@ FramebufferManager.prototype.update = function (
   //>>includeEnd('debug');
   pixelDatatype = defaultValue(
     pixelDatatype,
-    this._color ? PixelDatatype.UNSIGNED_BYTE : undefined
+    this._color
+      ? defaultValue(this._pixelDatatype, PixelDatatype.UNSIGNED_BYTE)
+      : undefined
   );
   pixelFormat = defaultValue(
     pixelFormat,
-    this._color ? PixelFormat.RGBA : undefined
+    this._color ? defaultValue(this._pixelFormat, PixelFormat.RGBA) : undefined
   );
 
-  if (this.isDirty(width, height, pixelDatatype)) {
+  if (this.isDirty(width, height, pixelDatatype, pixelFormat)) {
     this.destroy();
     this._width = width;
     this._height = height;
@@ -236,8 +242,8 @@ FramebufferManager.prototype.setColorTexture = function (texture, index) {
     );
   }
   //>>includeEnd('debug');
+  this._attachmentsDirty = texture !== this._colorTextures[index];
   this._colorTextures[index] = texture;
-  this._attachmentsDirty = true;
 };
 
 FramebufferManager.prototype.getDepthRenderbuffer = function () {
@@ -252,8 +258,8 @@ FramebufferManager.prototype.setDepthRenderbuffer = function (renderbuffer) {
     );
   }
   //>>includeEnd('debug');
+  this._attachmentsDirty = renderbuffer !== this._depthRenderbuffer;
   this._depthRenderbuffer = renderbuffer;
-  this._attachmentsDirty = true;
 };
 
 FramebufferManager.prototype.getDepthTexture = function () {
@@ -268,8 +274,8 @@ FramebufferManager.prototype.setDepthTexture = function (texture) {
     );
   }
   //>>includeEnd('debug');
+  this._attachmentsDirty = texture !== this._depthTexture;
   this._depthTexture = texture;
-  this._attachmentsDirty = true;
 };
 
 FramebufferManager.prototype.getDepthStencilRenderbuffer = function () {
@@ -286,8 +292,8 @@ FramebufferManager.prototype.setDepthStencilRenderbuffer = function (
     );
   }
   //>>includeEnd('debug');
+  this._attachmentsDirty = renderbuffer !== this._depthStencilRenderbuffer;
   this._depthStencilRenderbuffer = renderbuffer;
-  this._attachmentsDirty = true;
 };
 
 FramebufferManager.prototype.getDepthStencilTexture = function () {
@@ -302,8 +308,8 @@ FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
     );
   }
   //>>includeEnd('debug');
+  this._attachmentsDirty = texture !== this._depthStencilTexture;
   this._depthStencilTexture = texture;
-  this._attachmentsDirty = true;
 };
 
 FramebufferManager.prototype.clear = function (
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -119,7 +119,7 @@ FramebufferManager.prototype.update = function (
   depthTexture = defaultValue(depthTexture, false);
   hdr = defaultValue(hdr, false);
 
-  if (this.isDirty(width, height, hdr)) {
+  if (this._texturesSet || this.isDirty(width, height, hdr)) {
     if (!this._texturesSet) {
       this.destroyResources();
     }
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -22,20 +22,26 @@ function FramebufferManager(options) {
     options.colorAttachmentsLength,
     1
   );
+
+  this._depth = defaultValue(options.depth, false);
+  this._depthStencil = defaultValue(options.depthStencil, false);
+  //>>includeStart('debug', pragmas.debug);
+  if (this._depth && this._depthStencil) {
+    throw new DeveloperError(
+      "Cannot have both a depth and depth-stencil attachment."
+    );
+  }
+  //>>includeEnd('debug');
+
   this._createColorAttachments = defaultValue(
     options.createColorAttachments,
     true
   );
   this._createDepthAttachments = defaultValue(
-    options.createDepthStencilAttachments,
+    options.createDepthAttachments,
     true
   );
-  this._createDepthStencilAttachments = defaultValue(
-    options.createDepthStencilAttachments,
-    true
-  );
-  this._depth = defaultValue(options.depth, false);
-  this._depthStencil = defaultValue(options.depthStencil, false);
+
   this._useHdr = false;
 
   this._framebuffer = undefined;
@@ -43,6 +49,7 @@ function FramebufferManager(options) {
   this._colorTextures = [];
   this._depthStencilRenderbuffer = undefined;
   this._depthStencilTexture = undefined;
+  this._depthRenderbuffer = undefined;
   this._depthTexture = undefined;
 }
 
@@ -92,22 +99,21 @@ FramebufferManager.prototype.update = function (
   context,
   width,
   height,
-  depthStencilTexture,
+  depthTexture,
   hdr
 ) {
   //>>includeStart('debug', pragmas.debug);
   if (
     (!defined(width) || !defined(height)) &&
     (this._createColorAttachments ||
-      (this._depthStencil && this._createDepthStencilAttachments) ||
-      (this._depth && this._createDepthAttachments))
+      ((this._depth || this._depthStencil) && this._createDepthAttachments))
   ) {
     throw new DeveloperError(
       "width and height must be provided if color or depth attachments are created"
     );
   }
   //>>includeEnd('debug');
-  depthStencilTexture = defaultValue(depthStencilTexture, false);
+  depthTexture = defaultValue(depthTexture, false);
   hdr = defaultValue(hdr, false);
 
   if (this.isDirty(width, height, hdr)) {
@@ -136,8 +142,8 @@ FramebufferManager.prototype.update = function (
     }
 
     // Create depth stencil texture or renderbuffer
-    if (this._depthStencil && this._createDepthStencilAttachments) {
-      if (depthStencilTexture) {
+    if (this._depthStencil && this._createDepthAttachments) {
+      if (depthTexture) {
         this._depthStencilTexture = new Texture({
           context: context,
           width: width,
@@ -158,20 +164,30 @@ FramebufferManager.prototype.update = function (
 
     // Create depth texture
     if (this._depth && this._createDepthAttachments) {
-      this._depthTexture = new Texture({
-        context: context,
-        width: width,
-        height: height,
-        pixelFormat: PixelFormat.DEPTH_COMPONENT,
-        pixelDatatype: PixelDatatype.UNSIGNED_INT,
-        sampler: Sampler.NEAREST,
-      });
+      if (depthTexture) {
+        this._depthTexture = new Texture({
+          context: context,
+          width: width,
+          height: height,
+          pixelFormat: PixelFormat.DEPTH_COMPONENT,
+          pixelDatatype: PixelDatatype.UNSIGNED_INT,
+          sampler: Sampler.NEAREST,
+        });
+      } else {
+        this._depthRenderbuffer = new Renderbuffer({
+          context: context,
+          width: width,
+          height: height,
+          format: RenderbufferFormat.DEPTH_COMPONENT16,
+        });
+      }
     }
 
     this._framebuffer = new Framebuffer({
       context: context,
       colorTextures: this._colorTextures,
       depthTexture: this._depthTexture,
+      depthRenderbuffer: this._depthRenderbuffer,
       depthStencilTexture: this._depthStencilTexture,
       depthStencilRenderbuffer: this._depthStencilRenderbuffer,
       destroyAttachments: false,
@@ -198,9 +214,9 @@ FramebufferManager.prototype.setColorTexture = function (texture, index) {
 
 FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
   //>>includeStart('debug', pragmas.debug);
-  if (this._createDepthStencilAttachments) {
+  if (this._createDepthAttachments) {
     throw new DeveloperError(
-      "If setDepthStencilTexture is called, createDepthStencilAttachments must be false."
+      "If setDepthStencilTexture is called, createDepthAttachments must be false."
     );
   }
   //>>includeEnd('debug');
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -205,6 +205,14 @@ FramebufferManager.prototype.destroyResources = function () {
   }
   this._colorTextures = [];
 
+  if (
+    defined(this._depthStencilTexture) &&
+    !this._depthStencilTexture.isDestroyed()
+  ) {
+    this._depthStencilTexture.destroy();
+  }
+  this._depthStencilTexture = undefined;
+
   if (defined(this._framebuffer)) {
     this._framebuffer.destroy();
     this._framebuffer = undefined;
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -23,6 +23,7 @@ function FramebufferManager(options) {
     1
   );
 
+  this._color = defaultValue(options.color, true);
   this._depth = defaultValue(options.depth, false);
   this._depthStencil = defaultValue(options.depthStencil, false);
   //>>includeStart('debug', pragmas.debug);
@@ -52,7 +53,7 @@ function FramebufferManager(options) {
   this._depthRenderbuffer = undefined;
   this._depthTexture = undefined;
 
-  this._texturesSet = false;
+  this._attachmentsSet = false;
 }
 
 Object.defineProperties(FramebufferManager.prototype, {
@@ -108,25 +109,30 @@ FramebufferManager.prototype.update = function (
   //>>includeStart('debug', pragmas.debug);
   if (
     (!defined(width) || !defined(height)) &&
-    (this._createColorAttachments ||
+    ((this._color && this._createColorAttachments) ||
       ((this._depth || this._depthStencil) && this._createDepthAttachments))
   ) {
     throw new DeveloperError(
-      "width and height must be provided if color or depth attachments are created"
+      "width and height must be provided if color or depth attachments are created."
+    );
+  }
+  if (!this._color && !this._depth && !this._depthStencil) {
+    throw new DeveloperError(
+      "must enable at least one type of framebuffer attachment."
     );
   }
   //>>includeEnd('debug');
   depthTexture = defaultValue(depthTexture, false);
   hdr = defaultValue(hdr, false);
 
-  if (this._texturesSet || this.isDirty(width, height, hdr)) {
-    if (!this._texturesSet) {
+  if (this._attachmentsSet || this.isDirty(width, height, hdr)) {
+    if (!this._attachmentsSet) {
       this.destroyResources();
     }
     this._useHdr = hdr;
 
     // Create color texture
-    if (this._createColorAttachments) {
+    if (this._color && this._createColorAttachments) {
       var pixelDatatype = hdr
         ? context.halfFloatingPointTexture
           ? PixelDatatype.HALF_FLOAT
@@ -197,7 +203,7 @@ FramebufferManager.prototype.update = function (
       depthStencilRenderbuffer: this._depthStencilRenderbuffer,
       destroyAttachments: false,
     });
-    this._texturesSet = false;
+    this._attachmentsSet = false;
   }
 };
 
@@ -216,7 +222,21 @@ FramebufferManager.prototype.setColorTexture = function (texture, index) {
   //>>includeEnd('debug');
   index = defaultValue(index, 0);
   this._colorTextures[index] = texture;
-  this._texturesSet = true;
+  this._attachmentsSet = true;
+};
+
+FramebufferManager.prototype.setDepthStencilRenderbuffer = function (
+  renderbuffer
+) {
+  //>>includeStart('debug', pragmas.debug);
+  if (this._createDepthAttachments) {
+    throw new DeveloperError(
+      "If setDepthStencilRenderbuffer is called, createDepthAttachments must be false."
+    );
+  }
+  //>>includeEnd('debug');
+  this._depthStencilRenderbuffer = renderbuffer;
+  this._attachmentsSet = true;
 };
 
 FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
@@ -228,7 +248,7 @@ FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
   }
   //>>includeEnd('debug');
   this._depthStencilTexture = texture;
-  this._texturesSet = true;
+  this._attachmentsSet = true;
 };
 
 FramebufferManager.prototype.clear = function (
@@ -248,10 +268,13 @@ FramebufferManager.prototype.clear = function (
   passState.framebuffer = framebuffer;
 };
 
+FramebufferManager.prototype.destroyFramebuffer = function () {
+  this._framebuffer = this._framebuffer && this._framebuffer.destroy();
+};
+
 FramebufferManager.prototype.destroyResources = function () {
-  var i;
   var length = this._colorTextures.length;
-  for (i = 0; i < length; ++i) {
+  for (var i = 0; i < length; ++i) {
     var texture = this._colorTextures[i];
     if (defined(texture) && !texture.isDestroyed()) {
       this._colorTextures[i].destroy();
@@ -272,9 +295,6 @@ FramebufferManager.prototype.destroyResources = function () {
     this._depthTexture = undefined;
   }
 
-  if (defined(this._framebuffer)) {
-    this._framebuffer.destroy();
-    this._framebuffer = undefined;
-  }
+  this.destroyFramebuffer();
 };
 export default FramebufferManager;
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -18,6 +18,7 @@ import PixelFormat from "../Core/PixelFormat.js";
  * @param {Boolean} [options.color=true] Whether the FramebufferManager will use color attachments.
  * @param {Boolean} [options.depth=false] Whether the FramebufferManager will use depth attachments.
  * @param {Boolean} [options.depthStencil=false] Whether the FramebufferManager will use depth-stencil attachments.
+ * @param {Boolean} [options.supportsDepthTexture=false] Whether the FramebufferManager will create a depth texture when the extensions are supported.
  * @param {Boolean} [options.createColorAttachments=true] Whether the FramebufferManager will construct its own color attachments.
  * @param {Boolean} [options.createDepthAttachments=true] Whether the FramebufferManager will construct its own depth attachments.*
  *
@@ -36,6 +37,10 @@ function FramebufferManager(options) {
   this._color = defaultValue(options.color, true);
   this._depth = defaultValue(options.depth, false);
   this._depthStencil = defaultValue(options.depthStencil, false);
+  this._supportsDepthTexture = defaultValue(
+    options.supportsDepthTexture,
+    false
+  );
   //>>includeStart('debug', pragmas.debug);
   if (!this._color && !this._depth && !this._depthStencil) {
     throw new DeveloperError(
@@ -103,7 +108,6 @@ FramebufferManager.prototype.update = function (
   context,
   width,
   height,
-  depthTexture,
   pixelDatatype,
   pixelFormat
 ) {
@@ -112,7 +116,6 @@ FramebufferManager.prototype.update = function (
     throw new DeveloperError("width and height must be defined.");
   }
   //>>includeEnd('debug');
-  depthTexture = defaultValue(depthTexture, false);
   pixelDatatype = defaultValue(
     pixelDatatype,
     this._color ? PixelDatatype.UNSIGNED_BYTE : undefined
@@ -141,7 +144,7 @@ FramebufferManager.prototype.update = function (
 
     // Create depth stencil texture or renderbuffer
     if (this._depthStencil && this._createDepthAttachments) {
-      if (depthTexture) {
+      if (this._supportsDepthTexture && context.depthTexture) {
         this._depthStencilTexture = new Texture({
           context: context,
           width: width,
@@ -162,7 +165,7 @@ FramebufferManager.prototype.update = function (
 
     // Create depth texture
     if (this._depth && this._createDepthAttachments) {
-      if (depthTexture) {
+      if (this._supportsDepthTexture && context.depthTexture) {
         this._depthTexture = new Texture({
           context: context,
           width: width,
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -52,7 +52,7 @@ function FramebufferManager(options) {
   this._depthRenderbuffer = undefined;
   this._depthTexture = undefined;
 
-  this._dirty = false;
+  this._texturesSet = false;
 }
 
 Object.defineProperties(FramebufferManager.prototype, {
@@ -92,15 +92,10 @@ FramebufferManager.prototype.isDirty = function (width, height, hdr) {
       texture.height !== height
     ) {
       texturesDirty = true;
+      break;
     }
   }
-  return (
-    this._dirty ||
-    this._createColorAttachments ||
-    length === 0 ||
-    texturesDirty ||
-    this._useHdr !== hdr
-  );
+  return length === 0 || texturesDirty || this._useHdr !== hdr;
 };
 
 FramebufferManager.prototype.update = function (
@@ -125,7 +120,9 @@ FramebufferManager.prototype.update = function (
   hdr = defaultValue(hdr, false);
 
   if (this.isDirty(width, height, hdr)) {
-    // this.destroyResources();
+    if (!this._texturesSet) {
+      this.destroyResources();
+    }
     this._useHdr = hdr;
 
     // Create color texture
@@ -200,7 +197,7 @@ FramebufferManager.prototype.update = function (
       depthStencilRenderbuffer: this._depthStencilRenderbuffer,
       destroyAttachments: false,
     });
-    this._dirty = false;
+    this._texturesSet = false;
   }
 };
 
@@ -219,7 +216,7 @@ FramebufferManager.prototype.setColorTexture = function (texture, index) {
   //>>includeEnd('debug');
   index = defaultValue(index, 0);
   this._colorTextures[index] = texture;
-  this._dirty = true;
+  this._texturesSet = true;
 };
 
 FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
@@ -231,7 +228,7 @@ FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
   }
   //>>includeEnd('debug');
   this._depthStencilTexture = texture;
-  this._dirty = true;
+  this._texturesSet = true;
 };
 
 FramebufferManager.prototype.clear = function (
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -104,7 +104,8 @@ FramebufferManager.prototype.update = function (
   width,
   height,
   depthTexture,
-  pixelDatatype
+  pixelDatatype,
+  pixelFormat
 ) {
   //>>includeStart('debug', pragmas.debug);
   if (!defined(width) || !defined(height)) {
@@ -116,6 +117,7 @@ FramebufferManager.prototype.update = function (
     pixelDatatype,
     this._color ? PixelDatatype.UNSIGNED_BYTE : undefined
   );
+  pixelFormat = defaultValue(pixelFormat, PixelFormat.RGBA);
 
   if (this.isDirty(width, height, pixelDatatype)) {
     this.destroyResources();
@@ -130,7 +132,7 @@ FramebufferManager.prototype.update = function (
           context: context,
           width: width,
           height: height,
-          pixelFormat: PixelFormat.RGBA,
+          pixelFormat: pixelFormat,
           pixelDatatype: pixelDatatype,
           sampler: Sampler.NEAREST,
         });
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -13,6 +13,16 @@ import PixelFormat from "../Core/PixelFormat.js";
 /**
  * Creates a wrapper object around a framebuffer and its resources.
  *
+ * @param {Object} options Object with the following properties:
+ * @param {Number} [options.colorAttachmentsLength=1] The number of color attachments this FramebufferManager will create.
+ * @param {Boolean} [options.color=true] Whether the FramebufferManager will use color attachments.
+ * @param {Boolean} [options.depth=false] Whether the FramebufferManager will use depth attachments.
+ * @param {Boolean} [options.depthStencil=false] Whether the FramebufferManager will use depth-stencil attachments.
+ * @param {Boolean} [options.createColorAttachments=true] Whether the FramebufferManager will construct its own color attachments.
+ * @param {Boolean} [options.createDepthAttachments=true] Whether the FramebufferManager will construct its own depth attachments.*
+ *
+ * @exception {DeveloperError} Cannot have both a depth and depth-stencil attachment.
+ *
  * @private
  * @constructor
  */
@@ -60,16 +70,6 @@ Object.defineProperties(FramebufferManager.prototype, {
   framebuffer: {
     get: function () {
       return this._framebuffer;
-    },
-  },
-  depthStencilTexture: {
-    get: function () {
-      return this._depthStencilTexture;
-    },
-  },
-  depthStencilRenderbuffer: {
-    get: function () {
-      return this._depthStencilRenderbuffer;
     },
   },
   status: {
@@ -225,6 +225,10 @@ FramebufferManager.prototype.setColorTexture = function (texture, index) {
   this._attachmentsSet = true;
 };
 
+FramebufferManager.prototype.getDepthStencilRenderbuffer = function () {
+  return this._depthStencilRenderbuffer;
+};
+
 FramebufferManager.prototype.setDepthStencilRenderbuffer = function (
   renderbuffer
 ) {
@@ -237,6 +241,10 @@ FramebufferManager.prototype.setDepthStencilRenderbuffer = function (
   //>>includeEnd('debug');
   this._depthStencilRenderbuffer = renderbuffer;
   this._attachmentsSet = true;
+};
+
+FramebufferManager.prototype.getDepthStencilTexture = function () {
+  return this._depthStencilTexture;
 };
 
 FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -1,0 +1,126 @@
+import Framebuffer from "./Framebuffer.js";
+import PixelDatatype from "./PixelDatatype.js";
+import Renderbuffer from "./Renderbuffer.js";
+import RenderbufferFormat from "./RenderbufferFormat.js";
+import Sampler from "./Sampler.js";
+import Texture from "./Texture.js";
+import Color from "../Core/Color.js";
+import defaultValue from "../Core/defaultValue.js";
+import defined from "../Core/defined.js";
+import PixelFormat from "../Core/PixelFormat.js";
+
+/**
+ * Creates a wrapper object around a framebuffer and its resources.
+ *
+ * @private
+ * @constructor
+ */
+function FramebufferManager(options) {
+  options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+  this._depthStencil = defaultValue(options.depthStencil, false);
+  this._useHdr = false;
+
+  this._framebuffer = undefined;
+  this._colorRenderbuffer = undefined;
+  this._colorTexture = undefined;
+  this._depthStencilRenderbuffer = undefined;
+  this._depthStencilTexture = undefined;
+}
+
+FramebufferManager.prototype.isDirty = function (width, height, hdr) {
+  hdr = defaultValue(hdr, false);
+  var colorTexture = this._colorTexture;
+  return (
+    !defined(colorTexture) ||
+    colorTexture.width !== width ||
+    colorTexture.height !== height ||
+    this._useHdr !== hdr
+  );
+};
+
+FramebufferManager.prototype.update = function (
+  context,
+  width,
+  height,
+  depthTexture,
+  hdr
+) {
+  depthTexture = defaultValue(depthTexture, false);
+  hdr = defaultValue(hdr, false);
+
+  if (this.isDirty(width, height, hdr)) {
+    this.destroyResources();
+    this._useHdr = hdr;
+
+    // Create color texture
+    var pixelDatatype = hdr
+      ? context.halfFloatingPointTexture
+        ? PixelDatatype.HALF_FLOAT
+        : PixelDatatype.FLOAT
+      : PixelDatatype.UNSIGNED_BYTE;
+    this._colorTexture = new Texture({
+      context: context,
+      width: width,
+      height: height,
+      pixelFormat: PixelFormat.RGBA,
+      pixelDatatype: pixelDatatype,
+      sampler: Sampler.NEAREST,
+    });
+
+    // Create depth stencil texture or renderbuffer
+    if (this._depthStencil) {
+      if (depthTexture) {
+        this._depthStencilTexture = new Texture({
+          context: context,
+          width: width,
+          height: height,
+          pixelFormat: PixelFormat.DEPTH_STENCIL,
+          pixelDatatype: PixelDatatype.UNSIGNED_INT_24_8,
+          sampler: Sampler.NEAREST,
+        });
+      } else {
+        this._depthStencilRenderbuffer = new Renderbuffer({
+          context: context,
+          width: width,
+          height: height,
+          format: RenderbufferFormat.DEPTH_STENCIL,
+        });
+      }
+    }
+
+    this._framebuffer = new Framebuffer({
+      context: context,
+      colorTextures: [this._colorTexture],
+      depthStencilTexture: this._depthStencilTexture,
+      depthStencilRenderbuffer: this._depthStencilRenderbuffer,
+      destroyAttachments: false,
+    });
+  }
+};
+
+FramebufferManager.prototype.getFramebuffer = function () {
+  return this._framebuffer;
+};
+
+FramebufferManager.prototype.clear = function (
+  context,
+  passState,
+  clearColor,
+  clearCommand
+) {
+  var framebuffer = passState.framebuffer;
+
+  passState.framebuffer = this._framebuffer;
+  Color.clone(clearColor, clearCommand.color);
+  clearCommand.execute(context, passState);
+
+  passState.framebuffer = framebuffer;
+};
+
+FramebufferManager.prototype.destroyResources = function () {
+  if (defined(this._framebuffer)) {
+    this._framebuffer.destroy();
+    this._framebuffer = undefined;
+  }
+};
+export default FramebufferManager;
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -51,6 +51,8 @@ function FramebufferManager(options) {
   this._depthStencilTexture = undefined;
   this._depthRenderbuffer = undefined;
   this._depthTexture = undefined;
+
+  this._dirty = false;
 }
 
 Object.defineProperties(FramebufferManager.prototype, {
@@ -93,6 +95,7 @@ FramebufferManager.prototype.isDirty = function (width, height, hdr) {
     }
   }
   return (
+    this._dirty ||
     this._createColorAttachments ||
     length === 0 ||
     texturesDirty ||
@@ -122,7 +125,7 @@ FramebufferManager.prototype.update = function (
   hdr = defaultValue(hdr, false);
 
   if (this.isDirty(width, height, hdr)) {
-    this.destroyResources();
+    // this.destroyResources();
     this._useHdr = hdr;
 
     // Create color texture
@@ -197,6 +200,7 @@ FramebufferManager.prototype.update = function (
       depthStencilRenderbuffer: this._depthStencilRenderbuffer,
       destroyAttachments: false,
     });
+    this._dirty = false;
   }
 };
 
@@ -215,6 +219,7 @@ FramebufferManager.prototype.setColorTexture = function (texture, index) {
   //>>includeEnd('debug');
   index = defaultValue(index, 0);
   this._colorTextures[index] = texture;
+  this._dirty = true;
 };
 
 FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
@@ -226,6 +231,7 @@ FramebufferManager.prototype.setDepthStencilTexture = function (texture) {
   }
   //>>includeEnd('debug');
   this._depthStencilTexture = texture;
+  this._dirty = true;
 };
 
 FramebufferManager.prototype.clear = function (
@@ -261,13 +267,13 @@ FramebufferManager.prototype.destroyResources = function () {
     !this._depthStencilTexture.isDestroyed()
   ) {
     this._depthStencilTexture.destroy();
+    this._depthStencilTexture = undefined;
   }
-  this._depthStencilTexture = undefined;
 
   if (defined(this._depthTexture) && !this._depthTexture.isDestroyed()) {
     this._depthTexture.destroy();
+    this._depthTexture = undefined;
   }
-  this._depthTexture = undefined;
 
   if (defined(this._framebuffer)) {
     this._framebuffer.destroy();
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -64,6 +64,11 @@ Object.defineProperties(FramebufferManager.prototype, {
       return this._depthStencilTexture;
     },
   },
+  depthStencilRenderbuffer: {
+    get: function () {
+      return this._depthStencilRenderbuffer;
+    },
+  },
   status: {
     get: function () {
       return this._framebuffer.status;
--- a/Source/Renderer/FramebufferManager.js
+++ b/Source/Renderer/FramebufferManager.js
@@ -66,6 +66,7 @@ function FramebufferManager(options) {
   this._width = undefined;
   this._height = undefined;
   this._pixelDatatype = undefined;
+  this._pixelFormat = undefined;
 
   this._framebuffer = undefined;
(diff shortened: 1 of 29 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/framebuffer-manager/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,126 @@
...
+      context: context,
+      width: width,
+      height: height,
+      pixelFormat: PixelFormat.RGBA,
+      pixelDatatype: pixelDatatype,
+      sampler: Sampler.NEAREST,
+    });
+
+    // Create depth stencil texture or renderbuffer
+    if (this._depthStencil) {
```
ebogo1, line 71: With MSAA support, I think this will turn into: 

```js
if (this._depthStencil) {
  if (depthTexture) {
    // create a new depth stencil Texture
    if (numSamples > 1) {
      // create a new multisampled depth stencil Renderbuffer
    }
  } else {
    // create a new depth stencil Renderbuffer
  }
}
```

This seems to work in the [`wrapper-msaa`](https://github.com/CesiumGS/cesium/compare/framebuffer-manager...wrapper-msaa?expand=1) branch but might need to be refactored into something more general.

Source/Scene/PickDepth.js
```diff
@@ -1,115 +1,36 @@
...
   var width = depthTexture.width;
   var height = depthTexture.height;
 
-  var texture = pickDepth._depthTexture;
-  var textureChanged =
-    !defined(texture) || texture.width !== width || texture.height !== height;
-  if (!defined(pickDepth._framebuffer) || textureChanged) {
-    createFramebuffers(pickDepth, context, width, height);
+  var textureChanged = pickDepth._framebuffer.isDirty(width, height);
+  if (!defined(pickDepth.framebuffer) || textureChanged) {
```
lilleyse, line 32: Is the `!defined(pickDepth.framebuffer)` part necessary?

Source/Scene/PickDepth.js
```diff
@@ -186,9 +99,6 @@ PickDepth.prototype.isDestroyed = function () {
 };
 
 PickDepth.prototype.destroy = function () {
-  destroyTextures(this);
-  destroyFramebuffers(this);
-
```
lilleyse, line 191: The framebuffer manager needs to be destroyed

Source/Scene/GlobeDepth.js
```diff
@@ -2,14 +2,10 @@ import BoundingRectangle from "../Core/BoundingRectangle.js";
 import Color from "../Core/Color.js";
```
lilleyse, line 2: `SceneFramebuffer` and `PickDepth` can't be trimmed down much further but there's still a lot of room for improvement in `GlobeDepth`. I'm mainly thinking about the render state and scissor code which is in four different files and somewhat complex and mistake prone. I don't have a clear idea of how this might be integrated in `FramebufferManager`, if at all, but keep it in mind as you update the rest of the code.

Source/Scene/GlobeDepth.js
```diff
@@ -21,26 +17,20 @@ import StencilOperation from "./StencilOperation.js";
...
 function GlobeDepth() {
-  this._globeColorTexture = undefined;
-  this._primitiveColorTexture = undefined;
-  this._depthStencilTexture = undefined;
-  this._globeDepthTexture = undefined;
-  this._tempGlobeDepthTexture = undefined;
   this._tempCopyDepthTexture = undefined;
 
-  this._globeColorFramebuffer = undefined;
-  this._primitiveColorFramebuffer = undefined;
```
lilleyse, line 32: While you're at it could you remove the code in `Scene.js` related to `separatePrimitiveFramebuffer`?

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,126 @@
...
+  var colorTexture = this._colorTexture;
+  return (
+    !defined(colorTexture) ||
+    colorTexture.width !== width ||
+    colorTexture.height !== height ||
+    this._useHdr !== hdr
+  );
+};
+
+FramebufferManager.prototype.update = function (
```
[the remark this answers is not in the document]
lilleyse: For now, I'd say the current approach is manageable. `depthTexture` probably doesn't need to be passed in because it can be gotten from `context.depthTexture` and I don't think any of the calling code needs to override that.

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,126 @@
...
+) {
+  depthTexture = defaultValue(depthTexture, false);
+  hdr = defaultValue(hdr, false);
+
+  if (this.isDirty(width, height, hdr)) {
+    this.destroyResources();
+    this._useHdr = hdr;
+
+    // Create color texture
+    var pixelDatatype = hdr
```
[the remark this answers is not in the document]
lilleyse: Double check this, but I _think_ in WebGL all color attachments most have the same `internalFormat`/`format`/`type` so they would either all be floating point or all unsigned byte.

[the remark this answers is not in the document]
lilleyse: > - Some framebuffers are constructed with multiple texture attachments, and `FramebufferManager` currently expects a single attachment for color and depth stencil textures/renderbuffers. It feels clunky to have a `textureOptions` argument to keep track of constructor options in `FramebufferManager.update`, maybe if this is needed then it could go in the constructor instead?

I think my comment partly addresses this. But see where the OIT implementation takes you. That uses multiple attachments.

lilleyse:
> - Some framebuffers depend on attachments that belong to another framebuffer. One example is `GlobeDepth._updateDepthFramebuffer` - I think it's simplest to avoid the `FramebufferManager` class for this case and to construct a `Framebuffer` manually instead.

While unfortunate I agree. Generally we want a `FramebufferManager` to own its resources. If this comes up again in some other area of the code we could reconsider.

lilleyse:
> - Should the goal be for every class to use the `FramebufferManager` constructor instead of directly creating `Framebuffer` objects? If not, should the `FramebufferManager` class be tailored to framebuffers that will need MSAA support? 

Yes, the `FramebufferManager` should be used wherever possible.

lilleyse:
> - Which framebuffers need MSAA? Some candidates are the ones in `PickDepth`, `PickDepthFramebuffer`, `ShadowMap`, `GlobeTranslucencyFramebuffer`, `OIT`, `PointCloudEyeDomeLighting`, and `TranslucentTileClassification`. 

I don't have a concrete answer right now, but I think it will become clear as you start to update the other classes. General rule of thumb though is that any framebuffer that gets geometry rendered to it will need MSAA. This excludes framebuffers only used in full screen passes.

Source/Renderer/FramebufferManager.js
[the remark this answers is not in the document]
ebogo1: A couple places use the `context.depthTexture` check like `SceneFramebuffer` and `GlobeTranslucenyFramebuffer`, but others don't.. I'll revisit this when I add the FramebufferManager to more classes.

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,126 @@
...
+ * @private
+ * @constructor
+ */
+function FramebufferManager(options) {
+  options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+  this._depthStencil = defaultValue(options.depthStencil, false);
+  this._useHdr = false;
+
+  this._framebuffer = undefined;
+  this._colorRenderbuffer = undefined;
```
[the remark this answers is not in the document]
ebogo1: yes - should it be removed until MSAA is added?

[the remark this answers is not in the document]
lilleyse: It's ok to keep

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,277 @@
...
+  var length = this._colorTextures.length;
+  for (i = 0; i < length; ++i) {
+    var texture = this._colorTextures[i];
+    if (defined(texture) && !texture.isDestroyed()) {
+      this._colorTextures[i].destroy();
+    }
+  }
+  this._colorTextures = [];
+
+  if (
```
ebogo1, line 259: Now that it's possible for `FramebufferManager` to store shared attachments, should it be responsible for destroying them if they belong to another class? One example that came up was `TranslucentTileClassification` having several framebuffers all using the the same `translucentDepthStencilTexture` attachment. The logic in a few other places would need to be revisited but maybe something like this would make more sense:

```js
if (!this._createXAttachments && defined(this._XTexture)) {
  this._XTexture.destroy();
}
this._XTexture = undefined;
```

For color texture attachments, all the attachments would get skipped in `destroyResources` if `createColorAttachments` was set to false.

Source/Scene/GlobeTranslucencyFramebuffer.js
```diff
@@ -39,110 +30,70 @@ function GlobeTranslucencyFramebuffer() {
...
-    sampler: Sampler.NEAREST,
-  });
-
-  globeTranslucency._packedDepthFramebuffer = new Framebuffer({
-    context: context,
-    colorTextures: [globeTranslucency._packedDepthTexture],
-    destroyAttachments: false,
-  });
-}
+// var pixelDatatype = hdr
```
ebogo1, line 68: This needs to be removed, just commenting it out for now in case I move it back (see https://github.com/CesiumGS/cesium/pull/9966/files#r764464629).

Source/Scene/GlobeTranslucencyFramebuffer.js
```diff
@@ -39,110 +30,70 @@ function GlobeTranslucencyFramebuffer() {
 Object.defineProperties(GlobeTranslucencyFramebuffer.prototype, {
```
ebogo1, line 30: A few of these properties are only here for easy access in `GlobeTranslucencyFramebufferSpec`. Should those be removed? 

Related question - should `FramebufferManager` expose all its attachments as object property getters even if it isn't needed, just so the design is symmetrical for all attachment types?

ebogo1:
@lilleyse ShadowMap is an interesting case, since the class doesn't own any of the Framebuffers or attachments it creates. This could change, but then ShadowMap would store more resources than it does now which is kind of the opposite of what I've been using `FramebufferManager` for. Any thoughts?

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Creates a wrapper object around a framebuffer and its resources.
+ *
+ * @private
+ * @constructor
+ */
+function FramebufferManager(options) {
+  options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+  this._depthStencil = defaultValue(options.depthStencil, false);
```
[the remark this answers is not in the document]
ebogo1: This happens in `ShadowMap` in the `createFramebufferDepth` function. I added the `color` boolean for symmetry even though I'm not sure about switching ShadowMap to `FramebufferManager` yet.

Source/Renderer/FramebufferManager.js
ebogo1: talked offline: owners of a `FramebufferManager` with shared attachments are responsible for deleting the shared attachments themselves. I added a `destroyFramebuffer` function to `FramebufferManager` to help in b10c8ec.

Source/Renderer/FramebufferManager.js
```diff
@@ -272,9 +295,6 @@ FramebufferManager.prototype.destroyResources = function () {
     this._depthTexture = undefined;
   }
 
-  if (defined(this._framebuffer)) {
-    this._framebuffer.destroy();
-    this._framebuffer = undefined;
-  }
+  this.destroyFramebuffer();
```
ebogo1, line 298: I may revisit this - I think it's useful to separate the code to destroy the framebuffer and attachmenets into two functions, and it might be cleaner if `destroyResources` becomes `destroyAttachments` and doesn't always call `destroyFramebuffer` at the end.

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,308 @@
...
+    if (
+      !defined(texture) ||
+      texture.width !== width ||
+      texture.height !== height
+    ) {
+      texturesDirty = true;
+      break;
+    }
+  }
+  return length === 0 || texturesDirty || this._useHdr !== hdr;
```
ebogo1, line 99: This function does not consider whether `setAttachments` is true or false. I was unsure about this but I think it makes sense because of the logic in `update`.. 

Also, now that it's possible to not create color attachments, should this check depth attachments too? What about renderbuffers?

Source/Renderer/FramebufferManager.js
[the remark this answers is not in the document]
ebogo1: There are also some classes that can make use of this option to force a depth-stencil renderbuffer instead of texture (i.e. `PickFramebuffer._fb`).

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,308 @@
...
+    throw new DeveloperError(
+      "must enable at least one type of framebuffer attachment."
+    );
+  }
+  //>>includeEnd('debug');
+  depthTexture = defaultValue(depthTexture, false);
+  hdr = defaultValue(hdr, false);
+
+  if (this._attachmentsSet || this.isDirty(width, height, hdr)) {
+    if (!this._attachmentsSet) {
```
ebogo1, line 129: This check makes me unsure if `attachmentsSet` should be part of the class at all. If `createXAttachments` being set to false for any type of attachment means that the manager can expect that attachment to have been set before `update`, I think we can get rid of this member variable.

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,308 @@
...
+};
+
+FramebufferManager.prototype.getColorTexture = function (index) {
+  index = defaultValue(index, 0);
+  return this._colorTextures[index];
+};
+
+FramebufferManager.prototype.setColorTexture = function (texture, index) {
+  //>>includeStart('debug', pragmas.debug);
+  if (this._createColorAttachments) {
```
ebogo1, line 217: Is it safe to allow this function to work on any index? The `Framebuffer` equivalent is much stricter.

Source/Scene/GlobeDepth.js
```diff
@@ -56,194 +50,25 @@ function GlobeDepth() {
...
 function destroyFramebuffers(globeDepth) {
-  globeDepth._globeColorFramebuffer =
-    globeDepth._globeColorFramebuffer &&
-    !globeDepth._globeColorFramebuffer.isDestroyed() &&
-    globeDepth._globeColorFramebuffer.destroy();
-  globeDepth._copyDepthFramebuffer =
-    globeDepth._copyDepthFramebuffer &&
-    !globeDepth._copyDepthFramebuffer.isDestroyed() &&
-    globeDepth._copyDepthFramebuffer.destroy();
+  globeDepth._colorFramebuffer.destroyResources();
```
ebogo1, line 59: If `FramebufferManager.destroyResources` will no longer call `destroyFramebuffer`, this code will need to be updated.

Source/Scene/GlobeDepth.js
```diff
@@ -316,15 +141,16 @@ function updateCopyCommands(globeDepth, context, width, height, passState) {
       {
         uniformMap: {
           u_depthTexture: function () {
-            return globeDepth._depthStencilTexture;
+            return globeDepth._colorFramebuffer._depthStencilTexture;
```
ebogo1, line 144:
```suggestion
            return globeDepth._colorFramebuffer.getDepthStencilTexture();
```

Source/Scene/GlobeDepth.js
```diff
@@ -429,7 +226,13 @@ GlobeDepth.prototype.update = function (
   var width = viewport.width;
   var height = viewport.height;
 
-  updateFramebuffers(this, context, width, height, hdr, clearGlobeDepth);
+  var textureChanged = this._colorFramebuffer.isDirty(width, height, hdr);
+  if (textureChanged) {
```
ebogo1, line 230: I think this `if (textureChanged)` check can be removed, along with the `destroyResources()` calls, unless it's possible for only one of the two framebuffers to be dirty during an update. This is because neither framebuffer uses shared attachments, which means the built-in `isDirty` check in `FramebufferManager.update` should do the same thing.

We may have discussed this offline already, but I wanted to double check.

Source/Scene/GlobeDepth.js
```diff
@@ -450,23 +253,28 @@ GlobeDepth.prototype.executeUpdateDepth = function (
   clearGlobeDepth
 ) {
   var depthTextureToCopy = passState.framebuffer.depthStencilTexture;
-  if (clearGlobeDepth || depthTextureToCopy !== this._depthStencilTexture) {
+  if (
+    clearGlobeDepth ||
+    depthTextureToCopy !== this._colorFramebuffer._depthStencilTexture
```
ebogo1, line 258:
```suggestion
    depthTextureToCopy !== this._colorFramebuffer.getDepthStencilTexture()
```

Source/Scene/GlobeTranslucencyFramebuffer.js
```diff
@@ -39,109 +30,72 @@ function GlobeTranslucencyFramebuffer() {
...
+      return this._framebuffer.getColorTexture();
     },
   },
   classificationFramebuffer: {
     get: function () {
-      return this._framebuffer;
+      return this._framebuffer.framebuffer;
+    },
+  },
+  packedDepthFramebuffer: {
```
ebogo1, line 41: A bunch of these getters were added for easy access in specs. It was a tradeoff of making this file longer or making the specs file too reliant on knowing how to access `GlobeTranslucencyFramebuffer` properties. I think I've been inconsistent with choosing which getters to add to which classes..

Source/Scene/GlobeTranslucencyFramebuffer.js
```diff
@@ -39,109 +30,72 @@ function GlobeTranslucencyFramebuffer() {
...
 
 function updateResources(globeTranslucency, context, width, height, hdr) {
-  var colorTexture = globeTranslucency._colorTexture;
-  var textureChanged =
-    !defined(colorTexture) ||
-    colorTexture.width !== width ||
-    colorTexture.height !== height ||
-    hdr !== globeTranslucency._useHdr;
-  if (textureChanged) {
+  if (globeTranslucency._framebuffer.isDirty(width, height, hdr)) {
```
ebogo1, line 83: Another case where the `isDirty` check could be removed if it's impossible for just one of the framebuffers to be dirty during an update, because neither framebuffer uses shared attachments.

Source/Scene/OIT.js
```diff
@@ -86,20 +98,10 @@ function destroyTextures(oit) {
...
-    oit._alphaFBO && !oit._alphaFBO.isDestroyed() && oit._alphaFBO.destroy();
-  oit._adjustTranslucentFBO =
-    oit._adjustTranslucentFBO &&
-    !oit._adjustTranslucentFBO.isDestroyed() &&
-    oit._adjustTranslucentFBO.destroy();
-  oit._adjustAlphaFBO =
-    oit._adjustAlphaFBO &&
-    !oit._adjustAlphaFBO.isDestroyed() &&
-    oit._adjustAlphaFBO.destroy();
+  oit._translucentFBO.destroyResources();
```
ebogo1, line 101: Technically this is a slight change from the original code because it now destroys the attachments and the framebuffers.

Source/Scene/PickDepthFramebuffer.js
```diff
@@ -59,16 +50,12 @@ PickDepthFramebuffer.prototype.update = function (
   var width = viewport.width;
   var height = viewport.height;
 
-  if (
-    !defined(this._framebuffer) ||
-    width !== this._depthStencilTexture.width ||
-    height !== this._depthStencilTexture.height
-  ) {
+  if (this._framebuffer.isDirty(context, width, height)) {
```
ebogo1, line 53: Again, I think this `isDirty` check could be removed.

Source/Scene/SceneFramebuffer.js
```diff
@@ -33,138 +24,56 @@ function SceneFramebuffer() {
...
-      format: RenderbufferFormat.DEPTH_STENCIL,
-    });
-    this._depthStencilIdRenderbuffer = new Renderbuffer({
-      context: context,
-      width: width,
-      height: height,
-      format: RenderbufferFormat.DEPTH_STENCIL,
-    });
+  var depthTexture = context.depthTexture;
+  if (this._colorFramebuffer.isDirty(width, height, hdr)) {
```
ebogo1, line 48: Another `isDirty` check that can be removed if either all or none of the framebuffers will be dirty during an update.

Source/Scene/TranslucentTileClassification.js
```diff
@@ -149,39 +112,27 @@ function updateTextures(transpClass, context, width, height) {
...
   destroyFramebuffers(transpClass);
+  transpClass._drawClassificationFBO.setDepthStencilTexture(
+    transpClass._translucentDepthStencilTexture
+  );
+  transpClass._drawClassificationFBO.update(
+    context,
+    width,
+    height,
+    false,
+    false
```
ebogo1, line 127:
```suggestion
```

Source/Scene/TranslucentTileClassification.js
```diff
@@ -149,39 +112,27 @@ function updateTextures(transpClass, context, width, height) {
...
-  transpClass._accumulationFBO = new Framebuffer({
-    context: context,
-    colorTextures: [transpClass._accumulationTexture],
-    depthStencilTexture: transpClass._translucentDepthStencilTexture,
-    destroyAttachments: false,
-  });
+  transpClass._accumulationFBO.setDepthStencilTexture(
+    transpClass._translucentDepthStencilTexture
+  );
+  transpClass._accumulationFBO.update(context, width, height, false, false);
```
ebogo1, line 133:
```suggestion
  transpClass._accumulationFBO.update(context, width, height);
```

Source/Scene/TranslucentTileClassification.js
```diff
@@ -149,39 +112,27 @@ function updateTextures(transpClass, context, width, height) {
...
 
-  transpClass._packFBO = new Framebuffer({
-    context: context,
-    colorTextures: [transpClass._packedTranslucentDepth],
-    destroyAttachments: false,
-  });
+  transpClass._packFBO.update(context, width, height, false, false);
```
ebogo1, line 135:
```suggestion
  transpClass._packFBO.update(context, width, height);
```

Source/Scene/TranslucentTileClassification.js
```diff
@@ -199,18 +150,9 @@ function updateResources(
...
   var width = transpClass._opaqueDepthStencilTexture.width;
   var height = transpClass._opaqueDepthStencilTexture.height;
-
-  var colorTexture = transpClass._colorTexture;
-  var textureChanged =
-    !defined(colorTexture) ||
-    colorTexture.width !== width ||
-    colorTexture.height !== height;
-  if (textureChanged) {
+  if (transpClass._drawClassificationFBO.isDirty(width, height)) {
```
ebogo1, line 153: I think this check can also be removed because `drawClassificationFBO`'s color texture is not shared. Is it possible for only some of the framebuffers to be dirty here?

ebogo1 (commented):
@lilleyse As of e42e8f8 this PR is almost ready to be undrafted. 

Classes that are still constructing framebuffers manually (trimmed down after [7917da9](https://github.com/CesiumGS/cesium/pull/9966/commits/7917da94f4ea22bdc38b0bc4c451051b48a27500)) :
- `ComputeEngine` constructs a framebuffer for use in commands that are executed and then destroys the framebuffer right after. I'm not sure this is a good use case for `FramebufferManager`.
- `ShadowMap` constructs framebuffers with different sets of attachments based on variables that I believe can change at runtime (i.e. `isPointLight`). Also, between the cubemap color textures and regular color texture, I think the value of `FramebufferManager._createColorAttachments` would also need to be changed at runtime. It is doable to change this code to use `FramebufferManager` but I'm still not sure if it's a "good" use of the class unless the API is changed to give more runtime control over attachment management.
- `TextureAtlas` is another case of creating a framebuffer and immediately destroying (and unbinding) it just to copy a texture over. This could probably use `FramebufferManager` but after a first pass that doesn't seem like an improvement over the current code.

Still to-do: 

- [ ] More docs where needed
- [ ] `FramebufferManager` specs
- [ ] Validate edge cases in browser

Source/Renderer/FramebufferManager.js
ebogo1: Resolved by having `FramebufferManager` store its own width and height; `isDirty` checks those instead.

Source/Renderer/FramebufferManager.js
ebogo1: Resolved by removing the `attachmentsSet` variable.

Source/Scene/GlobeTranslucencyFramebuffer.js
ebogo1: Resolved by adding comments above getters that are only used for specs.

Source/Scene/TranslucentTileClassification.js
ebogo1: As of 7917da9 it looks like this check is still needed, I'll have to dig into this a bit more.

Source/Scene/InvertClassification.js
```diff
@@ -239,21 +249,17 @@ InvertClassification.prototype.update = function (context) {
...
     }
 
-    this._fbo = new Framebuffer({
-      context: context,
-      colorTextures: [this._texture],
-      depthStencilTexture: depthStencilTexture,
-      depthStencilRenderbuffer: depthStencilRenderbuffer,
-      destroyAttachments: false,
-    });
+    this._fbo.setColorTexture(this._texture);
```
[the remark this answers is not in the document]
ebogo1: Resolved by using a default sampler.. I need to double check this is fine.

Source/Scene/PickDepth.js
ebogo1: The logic would be different without it unless `isDirty` also checked that `framebuffer` was defined, but I don't think that's correct if the attachments are being set manually. It could be made to work either way, I'm not sure I have a preference.

Specs/Renderer/FramebufferManagerSpec.js
```diff
@@ -0,0 +1,369 @@
...
+      } else {
+        var renderbuffer = fbm.getDepthStencilRenderbuffer();
+        expect(renderbuffer).toBeDefined();
+        expect(renderbuffer.width).toEqual(1);
+        expect(renderbuffer.height).toEqual(1);
+        expect(renderbuffer.format).toEqual(RenderbufferFormat.DEPTH_STENCIL);
+      }
+    });
+
+    it("creates renderbuffer depth attachments if supportsDepthTexture is false", function () {
```
ebogo1, line 266: This behavior may cause errors to be avoided where previously they would be thrown. i.e., if a class used to always create depth textures without checking `context.depthTexture`, an error would be thrown when the extension was unsupported. Instead, this now creates a renderbuffer if `supportsDepthTexture` is disabled or if the extension is unsupported. 

I noticed this in `PickDepthFramebuffer` and `PointCloudEyeDomeLighting`, but there might be a couple similar places.

Source/Scene/InvertClassification.js
```diff
@@ -225,9 +193,13 @@ InvertClassification.prototype.update = function (context) {
...
-  if (!defined(this._fbo) || textureChanged || previousFramebufferChanged) {
-    this._fbo = this._fbo && this._fbo.destroy();
-    this._fboClassified = this._fboClassified && this._fboClassified.destroy();
+  if (
+    !defined(this._fbo.framebuffer) ||
+    textureChanged ||
+    previousFramebufferChanged
+  ) {
+    this._fbo.destroyResources();
+    this._fboClassified.destroyFramebuffer();
```
lilleyse, line 202: There might be a case where the resources aren't recreated if the framebuffer manager update function determines that it's not dirty. Maybe FramebufferManager should consider itself dirty if the framebuffer is undefined.

Also I think `this._fboClassified.destroyFramebuffer();` could just be `this._fboClassified.destroyResources();` since I wouldn't expect this code to be called frequently enough that `fboClassified` would need to hold onto the same color texture.

Source/Scene/InvertClassification.js
```diff
@@ -324,12 +290,12 @@ InvertClassification.prototype.clear = function (context, passState) {
   var framebuffer = passState.framebuffer;
 
   if (defined(this._previousFramebuffer)) {
-    passState.framebuffer = this._fbo;
+    passState.framebuffer = this._fbo.framebuffer;
     this._clearColorCommand.execute(context, passState);
```
lilleyse, line 294: Could you call the FramebufferManager clear function instead, here and below?

Source/Scene/InvertClassification.js
```diff
@@ -363,8 +329,7 @@ InvertClassification.prototype.isDestroyed = function () {
 };
 
 InvertClassification.prototype.destroy = function () {
-  this._fbo = this._fbo && this._fbo.destroy();
-  this._texture = this._texture && this._texture.destroy();
+  this._fbo.destroyResources();
```
lilleyse, line 332: Does `fboClassified` need to be destroyed too?

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Creates a wrapper object around a framebuffer and its resources.
+ *
+ * @param {Object} options Object with the following properties:
+ * @param {Number} [options.colorAttachmentsLength=1] The number of color attachments this FramebufferManager will create.
+ * @param {Boolean} [options.color=true] Whether the FramebufferManager will use color attachments.
+ * @param {Boolean} [options.depth=false] Whether the FramebufferManager will use depth attachments.
+ * @param {Boolean} [options.depthStencil=false] Whether the FramebufferManager will use depth-stencil attachments.
+ * @param {Boolean} [options.supportsDepthTexture=false] Whether the FramebufferManager will create a depth texture when the extensions are supported.
```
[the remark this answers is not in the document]
lilleyse: After reading `InvertClassification`, I guess there are cases where you could have a depth texture or a depth renderbuffer and the framebuffer manager wouldn't know in advance.

lilleyse (commented):
I still have a little ways to go but here's everything up to `InvertClassification`

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,332 @@
...
+  },
+});
+
+FramebufferManager.prototype.isDirty = function (width, height, pixelDatatype) {
+  var dimensionChanged = this._width !== width || this._height !== height;
+  var attachmentsNeedUpdate =
+    (this._color && !defined(this._colorTextures[0])) ||
+    (this._color && !this._createColorAttachments) ||
+    (this._depth && !this._createDepthAttachments) ||
+    (this._depthStencil && !this._createDepthAttachments);
```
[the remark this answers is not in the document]
ebogo1: Isn't that the correct behavior? Classes that set their own textures expect the FramebufferManager to create a new framebuffer during `update`.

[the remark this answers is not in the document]
ebogo1: Also, `_color`, `_depth`, and `_depthStencil` can be set to false through the constructor, so I'm not sure I fully follow.

Source/Scene/InvertClassification.js
ebogo1: The classification texture isn't getting destroyed in `main` but I don't see anything weird if I add the extra `destroy` call..

Source/Scene/OIT.js
lilleyse: Seems like a good fix to me

Source/Scene/PickDepth.js
lilleyse: Possibly relevant discussion https://github.com/CesiumGS/cesium/pull/9966#discussion_r771014464:
> Maybe FramebufferManager should consider itself dirty if the framebuffer is undefined

Then the only code needed here would be `pickDepth._framebuffer.update(context, width, height);`

Source/Scene/PickDepthFramebuffer.js
```diff
@@ -59,16 +51,10 @@ PickDepthFramebuffer.prototype.update = function (
...
-  if (
-    !defined(this._framebuffer) ||
-    width !== this._depthStencilTexture.width ||
-    height !== this._depthStencilTexture.height
-  ) {
-    destroyResources(this);
-    createResources(this, context);
-  }
+  destroyResources(this);
+  createResources(this, context);
```
lilleyse, line 55: This will destroy/create the resources every frame.

`framebuffer.update` ought to be sufficient

Source/Renderer/FramebufferManager.js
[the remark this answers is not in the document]
lilleyse: Offline we discussed that a new framebuffer should only be created if an attachment changes

Source/Scene/PointCloudEyeDomeLighting.js
```diff
@@ -146,21 +107,12 @@ function createCommands(processor, context) {
...
-  if (!defined(colorGBuffer) || resized) {
-    destroyFramebuffer(processor);
-    createFramebuffer(processor, context);
-    createCommands(processor, context);
-    nowDirty = true;
-  }
+  processor._framebuffer.update(context, width, height);
+  createCommands(processor, context);
+  nowDirty = true;
   return nowDirty;
```
lilleyse, line 116: `nowDirty` will be true every frame, which will cause a bunch of extra work in `update`.

Looking at this closer, `nowDirty` and `dirty` could be removed entirely since `derivedCommand.framebuffer !== this.framebuffer` should cover the bases now. Previously this was not the case.

One way to ensure that eye dome lighting still works is to run through the closed issues in https://github.com/CesiumGS/cesium/pull/9719 and make sure this PR hasn't introduced any regressions.

lilleyse:
@ebogo1 that should cover the rest of the files

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,332 @@
...
+
+  return dimensionChanged || attachmentsNeedUpdate || datatypeChanged;
+};
+
+FramebufferManager.prototype.update = function (
+  context,
+  width,
+  height,
+  pixelDatatype,
+  pixelFormat
```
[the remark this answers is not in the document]
ebogo1: It's only used in `PostProcessStageTextureCache`, where I believe a framebuffer's pixel datatype and format can never change at runtime (otherwise a new framebuffer would be created in `getFramebuffer()`...). I've added both to constructor options, which should simplify that code a bit.

Source/Scene/BrdfLutGenerator.js
```diff
@@ -39,35 +39,14 @@ function createCommand(generator, context) {
...
-  if (!defined(this._colorTexture)) {
+  if (!defined(this.colorTexture)) {
     var context = frameState.context;
 
-    createFramebuffer(this, context);
+    this._framebuffer.update(context, 256, 256);
     createCommand(this, context);
     this._drawCommand.execute(context);
-    this._framebuffer = this._framebuffer && this._framebuffer.destroy();
+    this._framebuffer.destroyFramebuffer();
```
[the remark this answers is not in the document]
ebogo1: I think it needs to be a member variable for `UniformState.update` to work - 
https://github.com/CesiumGS/cesium/blob/4c29dd20873c1bbbb09387bf562f45f24ecc964b/Source/Renderer/UniformState.js#L1265-L1269

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,332 @@
...
+) {
+  var framebuffer = passState.framebuffer;
+  var clearCommandColor = clearCommand.color;
+
+  passState.framebuffer = this._framebuffer;
+  clearCommand.color = Color.clone(clearColor, clearCommand.color);
+  clearCommand.execute(context, passState);
+
+  clearCommand.color = clearCommandColor;
+  passState.framebuffer = framebuffer;
```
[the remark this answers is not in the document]
ebogo1: Is it ever necessary to temporarily set the pass state's framebuffer instead of a command's framebuffer? I'm wondering if the `passState` argument can be removed, since `Context` has a default pass state to use if one isn't given.

Source/Scene/PointCloudEyeDomeLighting.js
ebogo1: Checked the local Sandcastles in the PR and didn't run into any issues after removing `nowDirty` and the return statement in `createResources`. Is this possible because FramebufferManager doesn't recreate the framebuffer when `nowDirty` used to be false?

Source/Renderer/FramebufferManager.js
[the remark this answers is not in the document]
lilleyse: > Is it ever necessary to temporarily set the pass state's framebuffer instead of a command's framebuffer?

I don't think so. But the pass state argument should still passed in because its other properties might be important.

Source/Scene/BrdfLutGenerator.js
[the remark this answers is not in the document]
lilleyse: Well the color texture could still be a member variable. The framebuffer would just exist temporarily to create the color texture.

Source/Scene/PointCloudEyeDomeLighting.js
lilleyse: > Is this possible because FramebufferManager doesn't recreate the framebuffer when `nowDirty` used to be false?

Yeah. The opposite is true too, FramebufferManager will recreate the framebuffer when `nowDirty` used to be true.

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+    throw new DeveloperError(
+      "createDepthAttachments must be false if setDepthStencilTexture is called."
+    );
+  }
+  //>>includeEnd('debug');
+  this._depthStencilTexture = texture;
+  this._attachmentsDirty = true;
+};
+
+FramebufferManager.prototype.clear = function (
```
ebogo1, line 309: @lilleyse I thought it didn't seem right to have `clear` take in a clear color argument, and from the comments above always temporarily use the clear command's framebuffer during the `execute` call.

Source/Scene/PostProcessStageTextureCache.js
```diff
@@ -245,7 +244,10 @@ function getFramebuffer(cache, stageName, dependencies) {
     pixelDatatype: pixelDatatype,
     clearColor: clearColor,
     stages: [stageName],
-    buffer: undefined,
+    buffer: new FramebufferManager({
+      pixelFormat: pixelFormat,
+      pixelDatatype: pixelDatatype,
```
lilleyse, line 249: It doesn't looks like a `pixelFormat` and `pixelDatatype` are in the FramebufferManager constructor. What are your latest thoughts here - should they be part of the constructor or should they be passed through to `update`?

 CC https://github.com/CesiumGS/cesium/pull/9966#discussion_r770961733

Specs/Renderer/FramebufferManagerSpec.js
```diff
@@ -0,0 +1,446 @@
...
+        context: context,
+        width: 1,
+        height: 1,
+      });
+      fbm.setDepthTexture(texture);
+      expect(fbm.getDepthTexture()).toBeDefined();
+      expect(fbm.getDepthTexture()).toEqual(texture);
+      texture.destroy();
+
+      var renderbuffer = new Renderbuffer({
```
lilleyse, line 140: `renderBuffer` should be destroyed at the end of the function

Specs/Renderer/FramebufferManagerSpec.js
```diff
@@ -0,0 +1,446 @@
...
+        context: context,
+        width: 1,
+        height: 1,
+      });
+      fbm.setDepthStencilTexture(texture);
+      expect(fbm.getDepthStencilTexture()).toBeDefined();
+      expect(fbm.getDepthStencilTexture()).toEqual(texture);
+      texture.destroy();
+
+      var renderbuffer = new Renderbuffer({
```
lilleyse, line 174: `renderBuffer` should be destroyed at the end of the function

Specs/Renderer/FramebufferManagerSpec.js
```diff
@@ -0,0 +1,446 @@
...
+      expect(fbm.getDepthTexture()).toBeUndefined();
+      expect(fbm.getDepthRenderbuffer()).toBeDefined();
+    });
+
+    it("creates renderbuffer depth attachments if supportsDepthTexture is true but context.depthTexture is false", function () {
+      fbm = new FramebufferManager({
+        depth: true,
+        supportsDepthTexture: true,
+      });
+      context._depthTexture = false;
```
lilleyse, line 296: `_depthTexture` should be set back to its original state so as not to interfere with subsequent tests. See PickingSpec for an example https://github.com/CesiumGS/cesium/blob/4c29dd20873c1bbbb09387bf562f45f24ecc964b/Specs/Scene/PickingSpec.js#L1292-L1308

Specs/Renderer/FramebufferManagerSpec.js
```diff
@@ -0,0 +1,446 @@
...
+
+    it("does not destroy resources if texture dimensions haven't changed", function () {
+      fbm = new FramebufferManager();
+      spyOn(FramebufferManager.prototype, "destroy").and.callThrough();
+      fbm.update(context, 1, 1);
+      fbm.update(context, 1, 1);
+      expect(FramebufferManager.prototype.destroy.calls.count()).toEqual(1);
+    });
+
+    it("destroys resources after texture dimensions change", function () {
```
lilleyse, line 370: Worth adding a few more related tests:

* `destroys resources after pixel datatype changes`
* `destroys resources after pixel format changes`
* `destroys resources after new attachment is set`

Source/Scene/BrdfLutGenerator.js
```diff
@@ -39,35 +36,28 @@ function createCommand(generator, context) {
...
-    colorTextures: [colorTexture],
-    destroyAttachments: false,
-  });
-
-  generator._framebuffer = framebuffer;
-}
-
 BrdfLutGenerator.prototype.update = function (frameState) {
-  if (!defined(this._colorTexture)) {
+  if (!defined(this.colorTexture)) {
```
lilleyse, line 40: Nitpick - since the getter is now simply returning `_colorTexture` there's no need to use the getter for internal code, the private member variable can be accessed directly.

```suggestion
  if (!defined(this._colorTexture)) {
```

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * Creates a wrapper object around a framebuffer and its resources.
+ *
+ * @param {Object} options Object with the following properties:
+ * @param {Number} [options.colorAttachmentsLength=1] The number of color attachments this FramebufferManager will create.
+ * @param {Boolean} [options.color=true] Whether the FramebufferManager will use color attachments.
+ * @param {Boolean} [options.depth=false] Whether the FramebufferManager will use depth attachments.
+ * @param {Boolean} [options.depthStencil=false] Whether the FramebufferManager will use depth-stencil attachments.
+ * @param {Boolean} [options.supportsDepthTexture=false] Whether the FramebufferManager will create a depth texture when the extensions are supported.
```
lilleyse, line 20:
```suggestion
 * @param {Boolean} [options.supportsDepthTexture=false] Whether the FramebufferManager will create a depth texture when the extension is supported.
```

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+ * @param {Boolean} [options.createColorAttachments=true] Whether the FramebufferManager will construct its own color attachments.
+ * @param {Boolean} [options.createDepthAttachments=true] Whether the FramebufferManager will construct its own depth attachments.*
```
lilleyse, line 22:
```suggestion
 * @param {Boolean} [options.createDepthAttachments=true] Whether the FramebufferManager will construct its own depth attachments.
```

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * @exception {DeveloperError} Cannot have both a depth and depth-stencil attachment.
```
lilleyse, line 24:
```suggestion
 * @exception {DeveloperError} Must enable at least one type of framebuffer attachment.
 * @exception {DeveloperError} Cannot have both a depth and depth-stencil attachment.
```

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+  this._depth = defaultValue(options.depth, false);
+  this._depthStencil = defaultValue(options.depthStencil, false);
+  this._supportsDepthTexture = defaultValue(
+    options.supportsDepthTexture,
+    false
+  );
+  //>>includeStart('debug', pragmas.debug);
+  if (!this._color && !this._depth && !this._depthStencil) {
+    throw new DeveloperError(
+      "must enable at least one type of framebuffer attachment."
```
lilleyse, line 46:
```suggestion
      "Must enable at least one type of framebuffer attachment."
```

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+  pixelDatatype = defaultValue(
+    pixelDatatype,
+    this._color ? PixelDatatype.UNSIGNED_BYTE : undefined
+  );
+  pixelFormat = defaultValue(
+    pixelFormat,
+    this._color ? PixelFormat.RGBA : undefined
+  );
+
+  if (this.isDirty(width, height, pixelDatatype)) {
```
lilleyse, line 137:
```suggestion
  if (this.isDirty(width, height, pixelDatatype, pixelFormat)) {
```

Source/Renderer/FramebufferManager.js
```diff
@@ -0,0 +1,354 @@
...
+    );
+  }
+  if (index >= this._colorAttachmentsLength) {
+    throw new DeveloperError(
+      "index must be smaller than total number of color attachments."
+    );
+  }
+  //>>includeEnd('debug');
+  this._colorTextures[index] = texture;
+  this._attachmentsDirty = true;
```
lilleyse, line 240: Maybe only set `attachmentsDirty` if the texture changed. That way if something sets the color texture to the same value every frame the framebuffer wouldn't be considered dirty. I don't think this is happening anywhere now but if it did happen in the future the code would stay fast.

```suggestion
  this._attachmentsDirty = texture !== this._colorTextures[index];
  this._colorTextures[index] = texture;
```

Source/Renderer/FramebufferManager.js
lilleyse: Good with me

Source/Scene/PostProcessStageTextureCache.js
ebogo1: oops, I forgot to set them from constructor options. Even though this is maybe the only place that could benefit from both options, I think it doesn't hurt to have them since it simplifies the `update` arguments later. I'm for adding them as constructor options :+1:

Source/Renderer/FramebufferManager.js
ebogo1: Sounds good, I like that change. Any reason to only check this for color textures and not all the attachments?

lilleyse: Ah I just commented on this one, but you could do the same thing for all the attachments

ebogo1: Cool, I made that change in 26bf8ad

lilleyse:
This is a pretty far reaching change so I went through a bunch of sandcastles to see if I could trigger any unexpected behavior. I also wanted to make sure that framebuffers were only recreated as needed, such as when the screen is resized or if HDR is toggled. I can't promise I've tested every corner case but things seem to be working well so far. Here's what I tested.

* Tested that `AutoExposure` works by setting `this._autoExposureEnabled = true` in `PostProcessStageCollection` and loading the [High Dynamic Range](http://localhost:8080/Apps/Sandcastle/index.html?src=High%20Dynamic%20Range.html) sandcastle. It has the same behavior as main, which is good, but I don't think the exposure is actually working properly in either branch.
* `BrdfLutGenerator` - tested with the [Image-Based Lighting](http://localhost:8080/Apps/Sandcastle/index.html?src=Image-Based%20Lighting.html) sandcastle
* `GlobeDepth` - tested several things
  * Made sure `czm_globeDepthTexture` was working correctly. It's used by billboards that are clamped to ground as a coarse grained depth test (see [BillboardClampToGround](http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FBillboardClampToGround.html)). It is also used by `GlobeTranslucency` when `requiresManualDepthTest` is true which can be triggered by [sandcastle1](http://localhost:8080/Apps/Sandcastle/index.html#c=xVpbc9u2Ev4rnLyIniowSPACxnHOsS3nWFMn9thKM506k6FJSEJDkTq82FZT//cuCF4lSlacqOGDeMHuYrH7LbBY6M6NlTvO7lmsHCohu1dOWMKzGfot/6b2vPz1JApTl4cs7vWVrzehAlfK4hg+XcbRHfdZ/Kpk9GLmpuxjFAf+SJKoe33JwsNxdBw9vFLGbpCwm/Bx7+AmvAnvQIXEYyEDDaQqKH89kE2TILoVTfk3lL/lbPI9gf5YeD13PXbizljsClXjKAhACgvd24CdwAtPeBQOWMq8FB5AWK4BiMnFoTR2wyTIQKC3KLh8IErjjFUapjxgCUvbVpI3MhjJRvWrksVBZYphFF6xJMpij6FxHM2OEqAZ+qqBqWXtKfnw5SjmMZ/xlN+xBLm+rxZ91eaZR8FikiteGIiFKdAX5IVHSqJXpYvENeVgk9ibLuBzQ/FLSXtWtqrlgNw4hSc3JLnGV64Pz8kR+HGh/vFSQ0RzDGI5um0TzcC63lcwshxKsaWbBDtY3PuKJCSOaToU69jQNbsidDTTprat6aZRExJN17FjOsQqCW3ohuiGA6aqBBrEpLpuYMMEC1Z0tuZYpmlajkM/7ZVQE9cMcBhzN1AqhwAUohhdnQ7QPU+nR8F86qoYmU0mL3CThI+55wqkjBZz1mBfaUPHF6MzyfzYwrPwRcCbkF7nsZyq5bJ5lHDRQdLoue2WAZsA6ku31JziemmbyMJgcUKwgYmFtb5iYIQNDeuU6jZ2DHDRKo/pOBbVTd3QoL3kweARDWyLqa118gj3aiahOsW45LHBUTY4nZh2J49lGw4BMgP8WfJQzQKNdSqU7mYiGETajmZrRDIRh1oAOOiKGnonj2kBhixqOI5NSx4AKcAFYLpmQIajEyB3HOyUPNQGNp0alknX8Ji24wCCLbvqhxKIEaxBBNjU6mbSDUoA3LphlgOybQA2WE3XnW4PgSoaWMHBhFgFD/QJYajBbY1XiWliMB3wgJUkj0mpbUEgEpOYNcunZhjccz+dAv5odzwtzSQCwxdZKm7vCiJYF+YsThfq17ZSngjB5Yj8/fT8/OLjkvqRFPix0ETvbj7pknd8fnTya03+uBTgs/ko+l8cZaEPjGKS74rg2yBj1/Mpi5+MYRmt62K1mELVYgrTACqO05zAKKWaTqCNiqlTxxgRCvEARIRqgKBSeQbL2DyJuN+eLWKQz9suqTVQdfC3EFr+bjM/Hp9/OF2aILsMFDP/x9vHfMo+P906sHp0WaO9blzM5QR+qPxRpUsPKYjqFWvIQjmO0mmv6A9IWQC5CRCMszDPUdS99pKQL9eovHesUodPrFIHbWkidlD18Hx5wgqNnHBpkEUG+G+Nc3R6dXU0fP8jh9oWuXm0ZKDkieC/NdyT0+vhh3efyeDzaHh++iNH3SlZDP5TBfgyPtZD/QqS6HI5eLZJqjA8fCqJ2+SiEdyz+Afr00Non8/cCUv2pWqfz6NJ9Fnq9+d80uu0W3Ovsd52b2PYxigXc/f/GVOOXe9L8by13h0bm7GQ+RY2SrnNoFMN4YONHLfQcTdDp5VXlR6Gd7D1gu3Uz9Ubf6veo0pa+rM1N5/WvLLyzpCCd4KUJb13Axa8E7Asqb4rvOCd4KWh7Q4RY+4AMSua7woz5g4ws6L87lBjfg9q5Fp17Ya+5yZpwEQ2P4qi4NaN37EwUztTXbH8rmdZShY2E3eskHWeHUSTAZvDjrAqzzVLhgia3RjSghn3crLjbDzOS5slX+fYJpOAHWdpCsbuwRqu+IKy16+Y+rUzvCnzvjC/8klDn6KpsOQ3q9XcTLhwS+p9lXyHZvmAEpYepWnMb7OUqb3UveWhzx5A4R7u7R0o+/uw32E+pDtppMyzVBlHXpYAtpR0ygrhlawo9AIOYDxcBVxBkXOrQrvHXEMf3meAWWG70zt4OOdJCqOM1d4XtvCje8j3K2GKykpxfAwvCEiUw0PInrye8vffSv7hJPJZ/tGym2CXphBF3aYpxPtBiyZJhSl7LYtLOmEpUcmGzehN2Ku5BMcvwCKefZakPMyxvHaj2FN+KTRB5X4WPcA3YfKOpsX6pr9EE+wz16oTxRxM2qHOGYO9bDi55Kk3vYqCoKnUVLatdiuIV76K8vhTarDQz6encRTP6i0wpM8xfzDQcHD6fjQc/d7ND0hufPUgfqMgjwEVKPbavothsnNDiL+l4n/pQC+aAYKZcOJVSarudUuvlZyiNCqKpGrVA7oHT+89kzeJsnT6XGbmPr/jMIrrjhs1h41gD7cFtqaLwitFoiYvyqm61ldeGjaxbNNA1KCGpWNiiLoh1RxLQ0Bs2I6OHVLWSrbHK0aGg0WBFfpybE1ziqMC0yQY3qiJ89os6StVIWZLFJZG6VxTyol9GCYcpplZlOVnWY05ql5ht7DqtpZ9KSrnBFmGaelgTMN0+oppEVsjFqK6rRsWNuy+ojvU1mwbwdgt4jg6bdagtretjjTdJuAZQjVR8aVgWUhKLOFC6hg2qGKCM0X9DGOd2ESzLKyblBqahkmzz21NLgurB9vZ/j2rT9B2bfjvh/TOYf09dhaTZp6fjMAWRxNAcpIWZbY6J5I533qyNS1PJkcy1QFLp4orOctzYPDqGqEbkqf1CrZzqecPp2E0GMzF7Z8wryaN81/ZNIUhB8sn4Nf1yXKe5ZxJIrXEqMiNhPSCWcB1GMJadSTHWg16Ft0xkS7VWM/TmgKv1cH2HJKwyyJLqHiqvKFcPEQOVejnszEPma9WovZaGwVBWQ+51VQqIY44Tqu6+cYqfvMKITzFruYY2Ou6e6+/SlnX/islO6jWlO6b1/PL+M3rmw88mtdj+9NjYzlvtjWX9t6lSKxrI8gUrPRWueOC8SesOfBNIrKwcHuvUe8s/SkgxAqPt3C1Hk8rfUktm5JK0uUekqoLP+ZBcLl9PwNBn8ta7S0PKdhBrwnCvDp9fvp29PnkfHjya/u4bkssb8ZvbetusOUoK3++8/QpvyRqVx+3PapbPo3aNH+PGltrpfijC8zcYtXYME1v/JNMPVNv0f/Z4Eos/2Ly3dCfRNWUT6aDBTiLe1eQCrNv7esDhEo8kce842jy9DAbfzBCWc0sT5irzpX/dJw1K43I3E67t9trBMpvsneJ9AfYg/vl4iA/yhdIfOcQWVH8jj9wiMgX/Revk3QRsDeyl//y2Rw2GeKvSypC+ymbzQPAW7J/m0FPKfKSPChf75dMr31+p3D/8ObF0r/Dbl7Ic1BoGWdBcM3/Yjcv3rzeB/oWWxDlOdTFHYsDdyFIptqbc/kRIfR6H15XuVKZUzYk/gM) in https://github.com/CesiumGS/cesium/pull/8726 when `Front Translucent Back Opaque` is selected and `Depth test against terrain` is unchecked. Also used by ground polylines and textured ground primitives such as in [Classification Types](http://localhost:8080/Apps/Sandcastle/index.html?src=Classification%20Types.html). This last example also tests that `executeUpdateDepth` is working correctly. Picking the polygons and polylines works as well.
  * Checked that post processing effects that use depth textures work: [Depth of Field](http://localhost:8080/Apps/Sandcastle/index.html?src=Fog%20Post%20Process.html&label=Post%20Processing), [Fog Post Process](http://localhost:8080/Apps/Sandcastle/index.html?src=Fog%20Post%20Process.html&label=Post%20Processing)
* Tested invert classification: [Classification](http://localhost:8080/Apps/Sandcastle/index.html?src=Classification.html&label=All)
* Tested viewsheds with invert classification in cesium-analytics: [Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rRnbbuO28ld0cl7k1lF8TWI7WdSN3a4PcqvtpthtioUi0TZRShQoyol3kX/vUCQl6mJvTtCXRBzOfYYzQ3rrMmuL0TNi1qUVomfrCsU4CZyHFGY/Hnnp+oqG3MUhYo9HTevbY2hZHDEGkHtGt9hHbKgJPYZcjv6gjPhLiWI3mo/ha2P0GG5BWOyhEIEsKdRJl7CV/nfWhD4hJ964Pn2OAUnxXKSAG+ojZ3o7/vl6OgEKyY5jgmLEAVdyiBgOMMdbFDuu79tCU8Mq+a87WUoqO7XEshJGMv1nNJyjmCbMQ86K0WAcA+LMt/tn7Y4wxLJeG49hI1OAIY+74TohLlugMKbCj9IH8/KOFAdqcsNfVy7j8OWG3VTcBK0ZQrF9fNZ3+oOzfrvXtHotp9U9O+2dN61z+JRaMNfHSTy0Oi0ApZANAlC4HlrHAwmIMPc2+ZJRAmZq7JePLlmNwzVBwKOfgnYGqCuxPOLGMV5hz+WYhstdhHLFK1vO1XQx+/3mS3fyZTm7nuqoSz/5NEDXNFxjnvgi/sftTsdpn5+f908H3cHgtHt+NjIwgalC7J067UG/f9bqtwa9816r3TPwbnCIgyT4iPB6I5Kg0xqcOsCzgOO+FHG6wmUFxf7NyMmUKpjbNGDKMANUsEGAKwGuRFgDVIT1cn+EVQrsqqD/L8LL6Xw+nt2mlOIExxvkP+AYPxF0RQk1nCNW8gCRaOMqrxQ278fX04fZ3fV0OZ9OlD9aTj/3gBZw53kEXOYfllAA344fPjUFt0aehask9IQhe4NMI7EdN2SBE7mhHH4v3DwH5xZr5MfSbtHIG5dvHE7ngOOGsWbuKJaN5luQ0/i+DVXEvpH6bqS1D6BgEkBn+CWvpUsGVCvKgkyVTP8l/QW/IP8X5gZI2aK5y9RXipTpilIZ5Wm+5CKlCj1njbj87CruhoKKteFdjVoyijIMupQk/JZAQFkIsDQh5koFycLWGsk0kGyIIHDJImEr14MSASvsluL7K8O+3oE+FyHGd7bUpJ7c8UTyFZmk+WiLM2m1zT+H+SBC0qwGXpDEh1AJdOUrmoSisn2zXoZWu9O0dvCvZb0a9qZV7p3G1tDutzQ3slWwtJZJjZl1eHtt7BRsrGvFqsPrYpqdlnStks4slIcPWY6pD+XuzbS7Cm19RId74EbHqFDUABV6vKHP1wV+4IWVS2JkIEwJwVFMsf8RTtdXGu7HnOSCDvHJUThLTIwHVdKLtPWdRDuubrdEWOoQZcrCtiKt63tZtavsGSbMRCWMUdpKqi5YbhhN1pvMEwUPoJJ/8gAWmtdyPr5d3I/n09tlOm+qFGeIJyzUgzMUQZhz1ZirsjyiMZZq1dZto3QOzYXarhyfYRWkBmDQ6FUotXBD33NjTpBQY0kpeXLZDQoT+0+BqNTi6IUPrcej7sRKB2+4QyiFwhgREAFu1L3ZbmgqS1v6ldJgSe2KLrKugDpSqWaNRHUDeZ/AfDCsSvorH2y/pmc6rWAtNXXuNMisCPezL3cP0/mX5R93Ck310XldwzyAPwthgAlQWg3rS44qusbI48KIWTvzGMOObpOfy527W2iqn21pcbXjfzpM+Mne7SGs0gUJ4TgiOzUoZBKaxfXnN40NbzasFBBJvU+zXIdMQP5ljhoalk5eZM9sZCpiIJbsL06CxgXk83R+V5yT4BDWDn0IjustZXzze1Qz71UOmaNripjeHlySID1o/y8hIHoCWeWE9NluZA4/OH5qe0sBBm1LoTWcsC/IvX9hNiyP1uVB1RxRa9PMzLOK84oKGSvlpnIF+PGycsh1qRX2eRBwumZutMHewmMuzN+lEcxAsPWLi6fTZC+JTKKMYKNvypXbaQHBrEPtrPbJvQlmskNme9ViNMnqq65CmWCzrEEzDiJb3zxShB8rUn4o67Tvfm3CzXcBM3nxyrK1KpeXVdysY1RtPZbGQpOAZh+jWlamOgdYZZzyE2XGP/eSGfT3v0RsTPfUJFqe5nlfzIrDHmU4zbLLrjLWU5Px8pcNSIYeZupqJV7r8klnkZqT1HsgjfkchT60dZhQpltIjWscc9hids0MsL9PjgoIZu6OijORvIkEdIsK0kZ756WfE85FpT9a0rUYG8ZShHhprWgo8kl52kcruBb5do2sRmZPzWZmSs0exDEJFWMzjb8dosjcr9PVePUTI3F2F0tn5Vo/CLszNyyAxnqgJAlSH+QsavxR4P+ffLWnHlchNfcj8Zb8Pkbm7ajKxTg379TjuxwOKVB+BRP9XN2PxIM3SQT0Lr03QE5pQPFMFcbrHOlt0/VbvLitaARm5IJGmtdBT7yNiRrm33eVyVmru4Xkdsin7Van1/guVqfVO/8+Vq81OG3k1xBZ6nC4RYwXH23BcHH/HB3Cudr3lCO0aDnqx4HqBzQo+UPIUfPoIuY7gj7o8PyEgwhmTPHDiu04JxwFkXjUiE+eEu9vxB0vjhtZLP/LpbMtHEYJ/5PDdfvy8QjGvzWc/7/y/LGsZ+zzzdA6a0UvGfVrPRuTTJgMtpJjl+A1XH0D7PsEjXKECCIOM9cxp9HQ6hjM860nCtUpKO4q0RcnpvEXPt5a2L+s+eVMvjnAziohZIG/gnkfLk4Av0JKaDoE3oHmxN0JtE37w7UEOo5zcQLLekrlBIOzxPgH)
* `OIT` has three different modes based on which WebGL extensions are supported. I tested these by modifying `Context.js`.
  * `colorBufferFloat: true, depthTexture: true, drawBuffers: false`
  * `colorBufferFloat: true, depthTexture: true, drawBuffers: true`
  * `colorBufferFloat: false, depthTexture: false, drawBuffers: false`
* `PickDepth` - checked that `scene.pickPosition` works, e.g.  [Drawing on Terrain](http://localhost:8080/Apps/Sandcastle/index.html?src=Drawing%20on%20Terrain.html) with log depth enabled and disabled
* `PickDepthFramebuffer` - checked that `scene.pickTranslucentDepth` and `scene.pickPosition` work - cherry-picked the code from https://github.com/CesiumGS/cesium/pull/9844 and ran the sandcastle from [here](https://github.com/CesiumGS/cesium/pull/9844#issuecomment-992503373)
* `PickFramebuffer` - checked that `scene.pick` works
* `EyeDomeLighting` - checked [Montreal Point Cloud](http://localhost:8080/Apps/Sandcastle/index.html?src=Montreal%20Point%20Cloud.html). Previously @ebogo1 ran through the linked issues in https://github.com/CesiumGS/cesium/pull/9719
* `PostProcessStageTextureCache` - checked most of the post processing sandcastles
* `SceneFramebuffer` - spot checked a few sandcastle examples with `context.depthTexture` returning false
* `SunPostProcess` - angled the camera so the sun intersects the horizon
* `TranslucentTileClassification` - tested the sandcastle from [here](https://github.com/CesiumGS/cesium/pull/9399#issuecomment-790849113)

@ebogo1 as a final test could you go through _all_ the sandcastles as if you were doing a release? I suggest doing something similar to what I did - print a message whenever a new framebuffer is created in FramebufferBuffer and make sure it gets printed only when it makes sense like when the screen is resized or HDR changes. Also check if there's any behavior I didn't test for.

ebogo1:
Going through all the Sandcastles, nothing looks out of place. Most framebuffers get created at runtime and when the canvas size changes. A couple other things I checked:
- No duplicate framebuffer or resource creation as far as I can tell; tested by checking when the body of `FramebufferManager.update` gets reached
- InvertClassification with translucency creates the same framebuffers as `main` when translucency is enabled/disabled or changed
- OIT and GlobeDepth recreate their framebuffers when HDR is toggled on/off - [HDR Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=High%20Dynamic%20Range.html&label=All)
- Tested the [3D Tiles Next Photogrammetry Classification Sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20Next%20Photogrammetry%20Classification.html) to make sure picking works, especially with the translucent windows option when a different pickable surface is visible behind a pickable window (also made sure this worked after resizing the canvas, just in case)
- Confirmed that the same number of framebuffers get created on this branch vs `main` in the [Multiple Synced Views Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Multiple%20Synced%20Views.html&label=All), which is a special case that creates the `PickFramebuffer` multiple times

lilleyse:
@ebogo1 cool, should be good to merge after the January 3rd release.

ebogo1:
@lilleyse I added a CHANGES.md entry for the 1.90 release - if the entry looks good this should be good to merge once CI passes.

lilleyse:
> edit - do we want an entry for these changes? It's a big change but most of the new API is in a private class..

No need to add to CHANGES.md since it's all private and invisible to the user. I'll remove the commit and then merge.

Issue #9972, opened 2021-12-09
ModelExperimental fails to load Draco compressed model


sanjeetsuhag:
[...]

PR #9974, opened 2021-12-09
Fix loading of Draco compressed model

lilleyse:
Fixes https://github.com/CesiumGS/cesium/issues/9972

The componentType of the indices accessor was 5125 (UNSIGNED_INT) but the typed array returned by `decodeDraco` was `Uint16Array`, causing a mismatch between the buffer's index datatype and the buffer's contents that resulted in a WebGL error. The fix was to use the index datatype of the decoded data instead of the accessor's component type.

Why does this model use 5125 for it's componentType? I'm not sure. The model only has 600 or so vertices. But technically it's valid so we should support this case.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag (approved):
The model renders correctly now!

image

Thanks for the quick fix, @lilleyse!

Issue #9730, opened 2021-08-14
PointCloud.js should not call CesiumMath.setRandomNumberSeed


mramato:
`CesiumMath.setRandomNumberSeed` is meant to be called by user (and test) code only because it sets a global RNG seed. It looks like `PointCloud.js `call it in `getRandomValues`.  Why?

It should be removed.  If a repeatable RNG is needed, it should create its own instance of `MersenneTwister`.

Stumbled on this while addressing random CI failures in https://github.com/CesiumGS/cesium/pull/9729

PR #9968, opened 2021-12-08
Adds support for using I3DMs with ModelExperimental

sanjeetsuhag:
### Overview

This PR enables loading of [Instanced 3D Models]() with `ModelExperimental`. It adds the following classes:
- `I3dmParser.js` - Used by `Cesium3DTileContent` classes for parsing the `.i3dm` payload.
- `I3dmLoader.js` - Used by `ModelExperimental`, used to convert `.i3dm` to `ModelComponents`

### Discussion

The transcoding of I3DM payload to `EXT_mesh_gpu_instancing` is handled as follows:
- If the `RTC_CENTER` is not present, a `BoundingSphere` is constructed from the `POSITION`/`POSITION_QUANTIZED` attributes (post-dequantization) and the `TRANSLATION` attribute always stores positions relative to the center. This mirrors the behavior of `ModelInstanceCollection` and it is required for samples such as [`3d-tiles-samples\1.0\TilesetWithTreeBillboards`](https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.0/TilesetWithTreeBillboards) which store positions without `RTC_CENTER`, causing jittering if loaded directly.
- `EXT_mesh_gpu_instancing` can only be applied to `nodes` with `primitives`, so the `I3dmLoader` iterates through all `ModelComponents.Nodes` and adds the `ModelComponents.Instances` if there are `ModelComponents.Primitives` present.

### Testing

[...]

Issue #9836, opened 2021-09-27
Transcode 3D Tiles 1.0 formats to `ModelExperimental`


ptrgags:
Once `ModelExperimental` is further along, we want to have the other tile formats (B3DM, I3DM, PNTS, etc) transcode to a ModelExperimental.

Some notes:

* for B3DM, we have a temporary solution in place from https://github.com/CesiumGS/cesium/pull/9830 where the B3DM content checks for the `ExperimentalFeatures.enableModelExperimental` flag. However, it would be better to transcode at parse time to a `Gltf3DTileContent`

ptrgags:
`ModelExperimental3DTileContent` is now used for both loading glTF and b3dm tiles. However, i3dm, pnts and cmpt still need to be handled.

ptrgags:
See #9968 for i3dm and #9978 for pnts.

`Composite3DTileContent` uses the `Cesium3DTileFactory` so it should "just work" but it would still be good to double check this once the other two PRs are finished.

Issue #9944, opened 2021-11-29
ModelExperimental: GPU Styling


ptrgags:
Though `CustomShaders` is the new way to do GPU styling, we still are supporting the existing styling language. some parts of that must be handled on the GPU (e.g. per-point styling).

Some notes from @lilleyse excerpted from [This branch](https://github.com/CesiumGS/cesium/blob/4e1a92c817ee37af619983f631f6d7f7ce1ca9d9/Source/Scene/CustomShader.js#L926-L972) describe when to use which type of styling:

> Styles may be evaluated on the CPU or GPU depending on the style and the
> types of properties involved. On the CPU styles are evaluated per-feature
> and the color/show results are stored in a batch texture, which is later
> applied on the GPU, either in the vertex shader or fragment shader.
> For GPU styling the style is converted to a shader and executed in the
> vertex shader or fragment shader directly. CPU styling is preferred.
>
> In some cases a style may require both CPU styling and GPU styling, in which
> case the style can't be applied and an error is thrown.
>
> Situations where CPU styling is required:
>   * Style uses language features not supported in GLSL like strings or regex
>   * Style uses properties that are not GPU compatible like strings, variable-size arrays, or fixed sized arrays with more than 4 components
>   * Style uses properties in `JsonMetadataTable` or `BatchTableHierarchy`
>   * Style uses custom evaluate functions, see `Cesium3DTileStyle#color`
>
> Situations where GPU styling is required:
>   * Style uses per-point properties. Large point clouds are generally impractical to style on the CPU.
>   * Style uses per-vertex properties. Per-vertex properties need to be interpolated before being styled.
>   * Style uses feature texture properties
>   * Style uses vertex attributes like position, color, etc
>   * Style uses uniforms
>   * Style references features in different feature tables
>   * Point size style is used
>
> Situations where the style must be applied in the fragment shader:
>   * Primitive uses a feature ID texture
>   * Style uses feature texture properties
>   * Style uses per-vertex properties
>   * Style uses interpolated vertex attributes
>   * Style is evaluated on the CPU and vertex texture fetch is not supported
>
> Situations where the style must be applied in the vertex shader:
>   * Point size style is used
>
> Since style variables aren't scoped there can be name collisions. They are
> resolved in the following order of precedence:
>
>   1. Input semantics (e.g. `POSITION`)
>   2. Attributes (e.g. `_TEMPERATURE`)
>   3. Uniforms
[...]

PR #9978, opened 2021-12-14
.pnts loader for ModelExperimental

ptrgags:
Fixes https://github.com/CesiumGS/cesium/issues/9730

Similar to #9968, this PR adds support for loading point clouds with ModelExperimental, each part of https://github.com/CesiumGS/cesium/issues/9836

This one was more involved, as pnts does not contain a glTF. Instead, the `PntsLoader` generates a `ModelComponents` object (our new runtime data structure for representing a 3D model) with a `POINTS` primitive.

This PR Does _NOT_ cover the following. These will be future PRs:
* Eye dome lighting
* Attenuation

This PR is most of the way there, but the following things need to be updated: 

- [x] Fix existing unit tests for `PointCloud3DTIleContent`
- [x] Fix existing unit tests for `TimeDynamicPointCloud`
- [x] Handle oct-encoded normals in PntsLoader
- [x] decode RGB565 colors on the CPU in PntsLoader
- [x] Add more unit tests for `PntsLoader`
- [x] Add unit tests for `PntsParser`
- [x] I just came across this issue: https://github.com/CesiumGS/cesium/issues/9730, this impacts this PR too so I should look into it.
- [x] Investigate bounding spheres that aren't in the right spot


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/pnts-loader/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelExperimental/PntsLoader.js
```diff
@@ -0,0 +1,481 @@
...
+  }
+
+  return attribute;
+}
+
+var randomValues;
+function getRandomValues(samplesLength) {
+  // Use same random values across all runs
+  if (!defined(randomValues)) {
+    CesiumMath.setRandomNumberSeed(0);
```
ptrgags, line 347: See https://github.com/CesiumGS/cesium/issues/9730

ptrgags:
Most of the way done, though still need to figure out why the time dynamic point cloud tests are failing (it renders, but then after the `camera.moveRight(10)` call it still renders... that shouldn't be happening.)

Oh right, I remember at one point turning on the bounding sphere draw commands and one model was showing up incorrectly. I'll add that to the checklist above.

Source/Scene/ModelExperimental/PntsLoader.js
```diff
@@ -0,0 +1,523 @@
...
+  var components = new Components();
+  components.scene = scene;
+  components.nodes = [node];
+  components.featureMetadata = makeFeatureMetadata(parsedContent);
+
+  loader._components = components;
+}
+
+PntsLoader.prototype.unload = function () {
+  this._components = undefined;
```
ptrgags, line 522: Oh yikes, I need to check to see what needs to be done to unload WebGL resources, e.g. the `Buffer`s

Source/Scene/PointCloud.js
```diff
@@ -1,4 +1,3 @@
-import arraySlice from "../Core/arraySlice.js";
 import BoundingSphere from "../Core/BoundingSphere.js";
```
ptrgags, line 1: I'm still trying to figure out why the `TimeDynamicPointCloud` unit tests are failing. Taking down some notes here. 

I checked the camera setup and frustum, nothing noticeably different between the unit test and [this sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rVVbb9s2FP4rhF8qFTJFyte4TrDA2dAOHRI0xh4WBQMjMTERijQoym065L/vkLpm84ICCyBIJD+e79yPNrwUVYF//rbnRhRcWSZ/4cxWhpeYK3Yn+W8653KIo1N0z2TJP6QqVQdm0EHwr9zAseJf0aYm/N2fBeko8/uNBkmhuElHEforVQiVO13J/FyJglm+QtZUPErVc9ix5nBeAumNu52OEkKXY7IY05Mtna3ockXIH0D2nyCevQbTV8HXZZNXwVb2tnOkMmLgB8YxPNes2Et+wSyL64BNLrZC8jK+0kLZjdRVPlhuIfIXT4oVIosJ3itbdhb8Xzr6tnTJ29JN3pZu2tPV6bmvVGaFVq7Y2IZJeceyxwBEuTkwGSGhcv4trAvWcOgKVa+RS+rKZ/bG37n1Fj4D6XObdguKPzVMGy0lrzWdth2yPYrje6OLT6VezgkFF/m5MewpcOSNYtFj5apukqhGhj6sXuxq41LVN9e+i8/Lth1Eqw9h4FW3UQGtx13zWjKpnfZ6JGC/8+elfZLQ5gNVw9xdOzRoPPS2XYvvcJ2SxrdMS21W6J3/wlQxPE9H4bvasbAdHI3WMuOK4z3MK2HFAQYZy/Og97iPQmmZsX1Gfq2kYMpFdpiFwMf4htx6DV5K739QyL+x5OrB7tAY0dtGd22nC6OEoYi/a11sdeDNiTx/b6MPIagbRhRA/8VewuUMLvh1h2SVMTCtj2JOQQfofS/j3l+YeuC9e5vuDH++vLz683p7efWydRrjIeZQEGFbpgi9SEbGCm4Yllo/ntvmbjSsho+c5UI9XAmb7by6gGASoTHM4kWEZpiELiSO1zdYo3RYTuAhrJiaBK0FNKEzQhO8XCbL5WK+mCyjFhpPF5P5hE4wmVF6cjI7mXbQlCzpnMxwktBkTqZJfR7WGYnfd9lrTHBwX1ue5Qfcgt8TeAXrsO7K97FjL7lPmK5sEITo9OxoDAt94F/Ew84G1MlH0CSEhB9G0Wjtm+ysdeQnUew11HdlZABj03IYm64e47sqe+QWZ2XZxnQdD0XXuTggkZ8e+XlDObKyBOS+ktL1aDo6W8dw/1+iUnvPLw/cSPbkru3o2ef6EGO8jmF7XNJqDVPL/IP5bw), I haven't yet seen a difference aside from the different aspect ratio (since the unit test renders to a `1x1` canvas.

The [particular unit test](http://localhost:8080/Specs/SpecRunner.html?spec=Scene%2FTimeDynamicPointCloud%20renders%20in%203D) does the following:
* renders the model, expects a red pixel (passes)
* adjusts the timeline and renders again, expects a red pixel (passes) 
* Moves the camera 10m to the right, expects nothing to render (fails)

I do know that it's something from `PointCloud.js` that causes the problem, because if I do 

```
git checkout main -- Source/Scene/PointCloud.js
```
then the test runs fine. 

Also I've tried increasing the moveRight amount, if I increase it to `30.0` the test passes.

I guess I'll read through `TimeDynamicPointCloud` closer as well as look at this diff in more depth since I'm not yet seeing the problem.

@lilleyse does anything jump out at you at what I might have done wrong in this file?

ptrgags: Update: I figured out the problem. I wasn't passing in the typed array correctly when computing the approximate bounding volume. Despite most of the properties needed being undefined, somehow it computed _a_ bounding volume with an infinite radius.

I'm also not sure why this changes the unit test. My guess is that the camera zoom was incorrect due to the bad bounding sphere... but it was zooming to a specific Cartesian3 + heading-pitch-roll. Not sure if that makes use of bounding spheres.

ptrgags:
@lilleyse the unit tests should now be working correctly now. This is ready for review

Source/Scene/ModelExperimental/PntsLoader.js
```diff
@@ -0,0 +1,532 @@
...
+    attribute.constant = Cartesian4.fromColor(attributeInfo.constantColor);
+  } else {
+    var buffer = Buffer.createVertexBuffer({
+      typedArray: typedArray,
+      context: context,
+      usage: BufferUsage.STATIC_DRAW,
+    });
+    loader._buffers.push(buffer);
+    attribute.buffer = buffer;
+    attribute.typedArray = typedArray;
```
lilleyse, line 349: Is there any need to hold onto the typed array?

Source/Scene/ModelExperimental/PntsLoader.js
```diff
@@ -0,0 +1,532 @@
...
+  var byteOffset = defaultValue(options.byteOffset, 0);
+
+  //>>includeStart('debug', pragmas.debug);
+  Check.typeOf.object("options.arrayBuffer", arrayBuffer);
+  //>>includeEnd('debug');
+
+  this._arrayBuffer = arrayBuffer;
+  this._byteOffset = byteOffset;
+
+  this._parsedContent = undefined;
```
lilleyse, line 50: `this._parsedContent` should be set back to undefined once it is no longer needed to help the garbage collector free some large typed arrays.

lilleyse (commented):
Looks pretty solid so far. I'll do a closer review tomorrow including unit tests and actual sandcastle testing.

Source/Scene/PointCloud.js
```diff
@@ -196,366 +195,85 @@ Object.defineProperties(PointCloud.prototype, {
...
+
+// Use MersenneTwister directly to avoid interfering with CesiumMath.nextRandomNumber()
+// See https://github.com/CesiumGS/cesium/issues/9730
+var randomNumberGenerator = new MersenneTwister(0);
 var randomValues;
 
 function getRandomValues(samplesLength) {
   // Use same random values across all runs
   if (!defined(randomValues)) {
-    CesiumMath.setRandomNumberSeed(0);
```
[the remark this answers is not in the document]
ptrgags: Are you saying the `MersenneTwister` should be function scoped, or just the constructor call to make sure it's set back to 0?

ptrgags:
@lilleyse aside from that one clarification about the MersenneTwister, this is ready for review again

Source/Scene/PointCloud.js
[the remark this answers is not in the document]
lilleyse: Never mind about my comment. The code was fine because all tiles share the same randomValues array.

ptrgags:
@lilleyse pulled in `main` and updated to use `tileset.enableModelExperimental` so it's easier for you to review

lilleyse:
For testing purposes I put together a sandcastle that loads all the `Specs/Data` point clouds with `enableModelExperimental: false` on the left and `enableModelExperimental: true` on the right. None of the side by sides look exactly right and some crash. The only ones I would expect to look different are point clouds with normals since the older point cloud shader uses lambert shading and the new point cloud shader uses PBR. Other things that should be tested are - (1) does CPU styling work, e.g. for `PointCloudBatched`? (2) Do custom shaders work for accessing vertex attributes and per-point metadata? (perhaps add a new sandcastle for point cloud custom shaders)

[Local sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rVZbc+I2FP4rGmZnMFlXpmWz0wmQdpPs5aG57JKmDyEzEfYB1MoSI8kkbIb/3iPZBhtM2s50JjH2uV++c6RYSWPJksMTaDIkEp7IORiepfTO04J27D/PlbSMS9DtTn8sY69lYpCASrk29Z/ILNmSpWCQfT+WhIxbN4pLey5Ulnz7fDZuhbvUc6fE8EMJpRv4Z8zGc0gaOBeaxeoQ/R/Ubpi2nIkG9pU6FMmV0ikT5jDnOrYfsQhJo9uvGSbJv7/Oe9UC1u/4/XEz40MD+Y/Po5/fNdG5nd+A9oQbrRaApQBzQPBWM2mmmN8B/u+Su3gLO6sr1/xxaywfPCIEIFzm6ulMZTLhcnanRJY67EyxXA40TsByAQbstUjqhCt48kammYwtV5LEGpiF25wbOJyFBJ7RL08BYSo65MWFmMOwMIK+PEDpAqW45UswlCVJUEF8/tO7KA17I4RkWpyQR0qj/G+0gNhEF8yyqKYQbesRvXlxQa2jwjf90yj5GObmEphks9FeLU4a6lNogGQTAZdYXfGxkuVJLWcnuu644cTe4H905J5lAItNbKM5cw4osxZkxnw9h8TqDLwqnxJfUDIcDv8VWIpab10ZuxJQXyXVOo0cuyyt6xHO2Al68i9BhxyRNy8WUrTPbKZhXcDNJecDXBNAxByOczPv/1dgRS+9o7YBj8DuI/kllwjaGpJ2h5yUnxORQbvjkF8N2j2OorI3GjA1WQaGEusavGc4Awv35jMs8ih0ypzg2aJPD/2cglgHgdEhtTQUbEpA9rGvIVVL+CBEkIdYqRQOIJapacj8tO7J43wekHew2opvS+vmEsWTFSIp5cZ1pejI0xwkZRjV/TYYWhUNK15rjId9R1U2tWg5aCoNIUumCdbH+iOw4ndSTONoMQcNNBfp1/U0jlNmXtXLRXb0cPi2WW+Wq6HAjMVjBBfq4lZ94s+QfNJYyyB33tn1rmw5woWpS2Y1f35HEUT5ay9AV2EV9SW9s2tNTaf5pqyOCB6Q+MZkLyhSPSI/0W5IusVj14h1uYjGqHo0zYTlC7E6W90hVnFeygTCwnl4wHct1goCUrcVc+v7NZhi62+34QSV0GpR5/eX7wqlVbDtYy2WL+D35g3H9fKNyRkEvgA/dOlxSDaV6XUrga7LVz/+68Nzbko4+ibkJHdxevD6iAsSuOOQI63bx59BfreiAuTMzvvk7Vu+wXOhTheZmQf1TWLu+UOnspCKlVJolOGNmExiBKEAdz7eKiUmTF8ihoJqvJ1+k+xsJuAssxYdjlvujCPlKJClP9Fwm+dLJNxuKVzX8V+QFBk03hIKkX7lSHNzduAodUf9HrGq64DzH3Wxl62wNfBnyGnZ4F95usBRdTeEAG8G7thCdOFVYJJhuJbGxpQQGERV1UHCl4QnQzz26rfrcYvEghmDnGkmxAjvguPW6SBC+T1VoTwir5egBVs5sfmPp7/lRErpIMLPZk2bN3XH8t8)

ptrgags:
To make a list of specific things to investigate from that Sandcastle
- [ ] `PointCloudConstantColor` - something's wrong with the vertex attribute. I'll need to check that I set the constant value correctly
- [ ] `PointCloudBatched` throws an error about a missing `a_featureId_0` -- I wonder if it has to do with the difference in names: BATCH_ID vs FEATURE_ID
- [ ] Make sure models are rendering as unlit when there are no normals
- [ ] Compare the shaders and see if there's any way I can get the colors to match
- [ ] Check if sRGB handling is done correctly
- [ ] Try this sandcastle again with styling
- [ ] Try this sandcastle again with custom shaders

Source/Scene/ModelExperimental/PntsLoader.js
```diff
@@ -0,0 +1,532 @@
...
+    featureIdAttribute.propertyTableId = 0;
+    featureIdAttribute.setIndex = 0;
+    primitive.featureIdAttributes = [featureIdAttribute];
+  }
+
+  var node = new Node();
+  node.primitives = [primitive];
+
+  if (defined(parsedContent.rtcCenter)) {
+    node.matrix = Matrix4.fromTranslation(parsedContent.rtcCenter);
```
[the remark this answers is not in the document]
lilleyse: @ptrgags I ended up opening a separate PR for that change: https://github.com/CesiumGS/cesium/pull/9988. Once that PR is merged you can do `components.transform = Matrix4.fromTranslation(parsedContent.rtcCenter);`

lilleyse:
>  Check if sRGB handling is done correctly

It seems like pnts colors are expected to be sRGB and glTF colors are expected to be linear, which explains the discrepancy here. The model shader is doing the right thing but we'll need a `#define HAS_COLOR_SRGB` to accommodate pnts.

ptrgags:
I talked with @lilleyse about the sRGB handling a bit so now I have a better understanding of what the goal here is. To summarize:

* `.pnts` files are assumed to store sRGB colors. So in the material stage, _always_ convert from sRGB to linear space when computing `material.diffuse`.
* when `scene.highDynamicRange` is true, no tonemapping or sRGB conversion should be applied since the frame buffer will store linear float values.
* when `scene.highDynamicRange` is false, tonemapping is applied (only for PBR) to reduce the range (still linear), followed by linear->sRGB conversion at the end.

Other notes:

* We should rename some of the builtin shader functions to be lowercase, e.g. `HSBToRGB -> hsbToRgb` to be consistent with our style guide.
* Adding `LinearToSrgb` and `srgbToLinear` functions would also make sense in the same PR. These functions should have no `#ifdefs`, but the calling code should.

ptrgags:
Fixed the sRGB handling, but I'm noticing a few aditional things after playing with an [updated Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rVdtTyM3EP4rVnRSNlzqzTVQVRBoL3B3rXS8lFD6gaDi7E4S97z2au0lBJT/3rG9m2ySDUelCsTuzovnmfEzYxMpqQ155DCDjBwTCTNyCprnCb11sqAZuc9TJQ3jErJm62goI+elI5CATt6buk9Uuied8sn0bC5ZwqNrJifWzmS51ZfuqAON4ruhJGTYuFJcmlOh8vj6S3/YaG9KT60Tww8lVFaj7zMTTSGu0ZxlLFK75N9xu2KZ4UzUqC/ULiQXKkuY0Ls1l5H5hEWIa8P+kWOS/Pl13asrYP0OfjqoV3ysEf/1ZfDzfp2cm+kVZE5wlakUsBSgdxjeZEzqMea3Q/+n5BZvsc78wm7+sDGU944RApBOUzXrq1zGXE5ulcgTy5kxlsuSxhoYLkCDuRTxuuACZhVaRbk2KhlMWbzJ6NOKJnixIAWS1GC4c0QmDku7r1Upvepfu4TCkGuXpMiR4ObQ0dlpHjEhePLr3sATqh6cWPG40J1j6wS37vV3mebYcNo924RLhZ/Rc/J3YqN5m8vceCP/0iIOLFkKaGpLO0AiYIL7tHNk1Qv758EBGmdskiDGekil1oH6XHx4WOMdsM6ZgQzbgCTFyxJSGJYiGvPxONduIEDUDT7QTpt0ij+tCsShXLTcftmfcS4jw5UkUQa40I3f0cDOhjaBJ+QKt/CWEf0eFxuPofywSdGKG/4ImrI4Dqp77h7ds3LhAnae4XY/UBr630EKkQ7PmGHhmkO44nD47sWCWoRFbPqPVtKXm5AYRvlksMXfwxpOFx4g2UiAY9inSpaHazm7mrV87TwFCTI4A2KmgG9zBII7RZgjeVp0FsEmJFLN2s4KZ1v0baSeyIwLUUS1ioS6Bff4mLhik+Pj4zc1/3Lnyzo4HBuNVqnhwKrLstv9w5l5iJHcS9Aie+Tdi4EE12cmz2BRjA+buOcMAZwAZCfO5fz+v4AV++wCNTU4dnYeyC/eImhmEDdb5LD8HIkcmi07yaqg98JyzzLAtGQJCrWLNdpPcJ6l9s1lV+RQ+JT5YP9iPNcSXoI9AAKRobRcKFimT7Z7IoNEPcJHIQIPr1IlHKZYorrmc5N3yx5n7Q57Ow5X5quy2n5F83iOLEq4mw/FbsymIClDVHcrMLRq2q5EXVPcbweqqikSXAZ1pcERyDJi57c7HCpxR0WXDtIpZEC9ydG6X8ZinutX/bzJhh/IfJX18qDUFJg2eCXAwzG9UZ/5E8Q4jhMIfPDWZnRlmOPMcikcyxl/2qdIIv/aDTBUu8r4Ut7aXE2Nx36CVtsDLzv4xmQ3KFLdIz9uj/HVIsadh7WoujTJheGpmPfnt8hV7JUygXYRvL0j9hrWCgPKowhX367BGLf+ZgUnqEBbQ+3vqs8KrVWw2sc1LL8hl3BLrziOFnd5DVwBfujQgzZZVqbbqQBdlK/ujFvs7nNd0tFtghfZS/C987ezO7BXG46yzhE+ev6eTAXIiZkekffv+ZLPhTtNcz0N1ieJvuP3rcqpW4yUwqOEN2AyjpCEAuy5eaOUGLHsHDkUVPH603rTdjIR0M+NwYDDhj37SNkKeM2wJx1Ocj9E2qsp5Y4jiIsMam98hYmDXumzHUesvQJsCau+ljj/0XfxhoQ/GoNlcvT6fp6VNNLlwWUvZxiVstVK9dnbDN7uhuAb7UbPHX4nJTt/5UmKc8ZeewK87tjzFlsD7zejHN0MjbQu+dsLq669mD8SHh/jeb3+b+CwQSLBtEbNOBfC3kWHjZNeiPZbrkI5yJd4FxZsbs2mH06+eiGltBfiZ72n8YzcWPlf)

- [ ] Translucency is not enabled for a constant color with alpha
- [ ] If the model has RGBA colors the model is not translucent
- [ ] `DracoBatched` is not rendering
- [ ] `NoColor` - before it defaults to dark gray, here I'm defaulting to white. I should change it to grey
- [ ] oct encoded normals don't seem to be used in the PBR shader, I see the default
- [ ] quantized models seem to be jittering (is something using world coordinates by mistake?)

ptrgags:
For `DracoBatched`, haven't fully investigated why it wasn't rendering but a warning in the console gives a hint: it says "vertex fetch requires 1000, but attribs only supply 500" - sounds like a mismatch of data types when uploading.

ptrgags:
Still investigating the crash when I try to apply a style. So far what I see happening:

* draw commands are built on the first frame
* the style is applied
* draw commands are rebuilt
* but somehow the attribute `vertexBuffer` was already destroyed so an error is thrown

ptrgags:
I figured out the crash - I overlooked that the vertex attribute buffers need to be marked as "not destroyable" so they don't get double destroyed when the style changes.

Now I'm noticing:
- [ ] looks like I have a bit more work to make styleable properties work correctly, even the `PointCloud3DTileContent` version isn't rendering correctly (should be red and blue). Probably a line or two missing in `PntsParser`.
image

ptrgags:
Turns out it wasn't that, there was a syntax error in the style that was being swallowed

[Updated Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rVdtb9s2EP4rhFHAcqtR7twMg2Nna9KXFcjbajf7EAcLLZ1srhRpiJQTJ/B/35GSbMmWsw0YkNjSvfCeOz53pEMltSFLDg+QkiGR8EDOQPMsoTdO5rVD93qmpGFcQtruHE9k6Lx0CBLQKfem7hWVpVqyBDSqbyeSkEnrWnFpzoTKoq+fTyctf1d6Zp0Yviih0gb9KTPhHKIGzYeUheqQ/B/crllqOBMN6kt1CMmlShMm9GHNVWg+YhGixrC/Z5gkf3pZ9+IKWL+jn46aFe8bxH98Hv38rknOzfwaUie4TtUCsBSgDxiOUyZ1jPkd0H+T3OIt1lld2s2ftCbyzjFCANJlrh5OVSYjLmc3SmSJ5U6M5bKksQaGC9BgrkRUF1zCQ4VWYaaNSkZzFu0y9qyi8Z4tSMFnc4PhLhCZ6Jd251Up/XZ5/mXsUopTNktAmnyFMTyaPrm3iqXi0UZ7gW3gfSpevshFZkis3bdPuFT4Gj4lfyZ27QtmIEV2kaR46BAHi2wENOJxnGnXRRD2vLe065Nu8WE7jZC1/UAY644rQpzJ0HAlSZgCLjLOS+TZZvMJPGLxuQW2iZUXragkhnFdShdoxQ1fgqYsirxqEd1X70O5cAE4S7F+95QG+d9oAaEOPjDDgppDsCVF8OrZgloHRWz6l1by3s+Xi2CazUZ7hOg3kKTwAMmmAtyWfaxk2a/lXNgGQZUk/RplXEk7eWntf7DBt9hAt5YYnzJjQGbMlXtITJqBc+MxcfUmw+HwXzXUZtvLUNqsBOyQt1LGkVWXlbdbiHOoj5Hcg9chr8mrZwMJrs9MlsK6aEmbWE4ZAthVh3FuZuL/BazYaheorcERtNsmv+QWXjuFqN0h/fJ1KjJod+x0qIIudyQFTEqWkFC3rvF+hhNiYZ9cbkUGhU+ZDbYuRnM9kUuwCUAgLpSWC3mb5Ml+U6SQqCW8F8LLwVVqhOMJC9TUfW6W7dnj9Dpgbwm1Nd8W1TYsmkcr5FDC3XAo9uJhDpIyRHW7BUOrpn4lak1xtx+oqqYGV/aaSoPTj6UE62PcuK3EnRZtOlrMIQWamxzX/VJspEy/6Jeb7Phh222z3hw9mgLTBg9ZPG4WY/WJP0KEkzgBLw/e2Y2uTNm8xVI4kVP++I4iifLHnoeh/CrfS3lndzUVx/kIrTYHXh/wicmeV6T6mvy4P8O3ixibi2hE1aNJJgxfiNXp6ga5ip1SJuAXwf0DsWtYKwwoTyFcfb8GMW79eAvHq0Croc5vd08KrZW33ccalt/ATcxrjoPlK5Mz8FwBfujSI59sKtPrVoCuy0fX+OvDfa5LOrpNyEX2Wnnn/JEXxLOXBY6y7jF+DfKbJxUgZ2Z+TN684Rs+F+50kem5V58k+pbfdSqjqBgphUcJb8RkFCIJBdiDc6yUmLL0AjnkVfHmR/Wu7Wwm4DQzBgNOWvbwI2Ur4A3DHnU4x/Mh4m+nFA7q8DtERQaNd6jCxEGv9NmBM9beAfaEVV9LnP/oi3vZ8lsDd3qclBv8K08W2Kr26uDhlcEeWMguvCNMM4RraKh1SYFBUHUdRHxJeDTEA6/+22PSIqFgWqMmzoQY4U150joZBGi/5yqUY+TVElLBVtZs/vbkPBdSSgcBvjZ7mnxTd1b+Gw)

PointCloudBatched now renders correctly (the difference in color is due to PBR vs Lambert shading)
image

Per-point properties will not work, but that falls under #9944. 

@lilleyse at this point this PR is ready for review again!

lilleyse:
Thanks @ptrgags. I tested HDR on/off and spot checked some other sandcastles. glTF point clouds work fine as well.

PR #9981, opened 2021-12-16
Use availability to determine when entity paths should be shown

srcejon:
As per https://community.cesium.com/t/entity-path-not-disappearing-when-entity-is-not-available/8551 and #1740, it seems entity paths are always shown even when other parts of the entity are no longer displayed due to its availability value.

This patch makes it so that the path is only displayed when the entity is available.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 - Fixed `scene.pickTranslucentDepth` for translucent point clouds with eye dome lighting. [#9991](https://github.com/CesiumGS/cesium/pull/9991)
 - Added a setter for `tileset.pointCloudShading` that throws if set to `undefined` to clarify that this is disallowed. [#9998](https://github.com/CesiumGS/cesium/pull/9998)
 - Fixes handling .b3dm `_BATCHID` accessors in `ModelExperimental` [#10008](https://github.com/CesiumGS/cesium/pull/10008)
+- Fixed path entity being drawn when data is unavailable [#1704](https://github.com/CesiumGS/cesium/pull/1704)
 
 ### 1.89 - 2022-01-03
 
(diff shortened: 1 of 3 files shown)


cesium-concierge:
Thank you so much for the pull request @srcejon! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/srcejon/cesium/blob/entity_path_use_availability/CONTRIBUTORS.md) file!
* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/entity_path_use_availability/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

srcejon:
I've just signed the CLA.

Daniel-Brosnan-Blazquez:
Dears,

We have our use cases broken since 1.90 version of Cesiumjs due to this change as we were using path availability to limit the display of segments (as the cartesian positions in the czml go beyond the stop time of the segments) and using the show parameter of path to display the segments all the time no matter where the timeline was set. So we were able to display something like this:
image

Now, since version 1.90, segments are only allowed to be shown during the availability period as shown here:
image

So, there is no possibility now to show all paths all the time.

I think that the czml options were already allowing to limit when the entity paths should be displayed through the "show" parameter of path. It does not seem to be the correct way to limit this by changing the interpretation of availability, it should be re-directed to change the "show" parameter.

Because, if this change is maintained, could you please tell us how we could obtain the same results?

Thanks in advance for any feedback.
Regards.

srcejon:
> It does not seem to be the correct way to limit this by changing the interpretation of availability

But that is the way it seems to work for all of the other parts of an entity, 

> it should be re-directed to change the "show" parameter.

How are you setting the value of show when there's no supposedly no data available?

Daniel-Brosnan-Blazquez:
> How are you setting the value of show when there's no supposedly no data available?

"show/interval" in this use case (showing several segments all the time) is set to cover all the period of the timeline and "availability" only covers every specific segment. So that the combination makes available a list of positions during the corresponding "availability" period but shows this list of positions during the corresponding "show/interval" period. And with this combination you can also configure the input to show the positions only during the "availability" period.

Daniel-Brosnan-Blazquez:
So, to put it in a different perspective, now there seems to be no difference between "availability" and "show/interval".

Daniel-Brosnan-Blazquez:
Dears,

Could you please tell if these last comments are going to be considered?

Thanks in advance for any feedback.
Regards.

PR #9984, opened 2021-12-19
Processor leak

H91011:
Hi,

There is a memory leak when we add big number of lines with big number of vertex especially added with ` clampToGround: true`
the reason of the problem is taskprocessors are does not destroy properly so it cause memory leak, you can see result in picture in commit.


cesium-concierge:
Thank you so much for the pull request @H91011! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/H91011/cesium/blob/processor-leak/CONTRIBUTORS.md) file!
* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/processor-leak/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

H91011:
I signed CLA

srothst1:
Hi @H91011. Thank you for the pull request. We really appreciate that you are taking the time to contribute to CesiumJS. While we will need to review the core changes that your code makes, there are some high-level issues that also need to be addressed. Your commits are causing the Travi CI Build to fail. I recommend investigating the edits that caused this issue.

Source/Scene/Primitive.js
```diff
@@ -1203,7 +1204,14 @@ var numberOfCreationWorkers = Math.max(
   1
 );
 var createGeometryTaskProcessors;
-var combineGeometryTaskProcessor = new TaskProcessor("combineGeometry");
+function removeGeometryTaskProcessors() {
+  if (createGeometryTaskProcessors) {
+    for (var i = 0; i < createGeometryTaskProcessors.length; i++) {
+      createGeometryTaskProcessors[i].destroy();
```
srothst1, line 1210: Are you sure that this is a safe way to destroy the `createGeometryTaskProcessors` object?

Source/Core/TaskProcessor.js
```diff
@@ -3,7 +3,7 @@ import when from "../ThirdParty/when.js";
 import buildModuleUrl from "./buildModuleUrl.js";
 import defaultValue from "./defaultValue.js";
 import defined from "./defined.js";
-import destroyObject from "./destroyObject.js";
+// import destroyObject from "./destroyObject.js";
```
srothst1, line 6: If this import is not being used, it should be deleted not commented out. What is the reason behind not using `destroyObject`?

Source/Scene/Primitive.js
```diff
@@ -1,3 +1,4 @@
+// import { createTaskProcessorWorker } from "../Cesium.js";
```
srothst1, line 1: If this import is not being used, it should be deleted not commented out.

srothst1 (commented):
@H91011 to be candid, I do not have the expertise to fully review and merge this pull request. However, I do have experience contributing to CesiumJS and an understanding of the best practices when contributing. This review is mostly styling suggestions and high-level questions to improve my understanding.

Source/Core/TaskProcessor.js
H91011: because `destroyObject` does not destroy properly 
you can llok at the `destroyObject` method its only delete functions, actually I did not understand why only remove functions.
it can be an array or an object with big amount of data right ?
I dont know why they do this way.  
So I remove it and I remove everything in object

Source/Scene/Primitive.js
H91011: I call this function after workers complete so if worker complete and then values taken we dont need worker anymore.

Source/Core/TaskProcessor.js
```diff
@@ -352,7 +351,6 @@ TaskProcessor.prototype.destroy = function () {
   if (defined(this._worker)) {
     this._worker.terminate();
   }
-  return destroyObject(this);
```
ggetz, line 355: Thanks for the PR @H91011! This change is a fairly sweeping one, which in turn is causing a lot of the unit tests to fail. Would you mind explaining in more detail why the task processor is not getting properly destroyed? We'll want to make sure this solution is generic to all uses of the `TaskProcessor`, not just its use in `Primitive.js`.

H91011: Yes I realize that because of unit tests. I try to some changes but I can not pass.
 you can take a look this topic I share results: https://community.cesium.com/t/memory-usage-did-not-reduce-after-remove-datasource/15203/4
```
For instance, if there’s an object in a global variable, and that object has a property referencing another object, that object is considered reachable. And those that it references are also reachable. 
```
you can read here : https://javascript.info/garbage-collection

Your task processor object properties is still reachable after your progress so they can not release from memory. I found this way for release memory. The code is a bit crowded here so I can found this after search one week. 

I think there is a misunderstanding about cesium destroy mechanism, it says destroy but it actually not destroy complete. I think if you say destroy it must be destroy real, not only remove functions. 

There is another leaks for destroying objects, after destroy function some gl objects are still stay reachable so it cause memory leak also(I realize in points object)

ggetz: Based on your [example](https://github.com/CesiumGS/cesium/pull/9984/files#diff-a2fe521025654f48a319ee31aead9312fbf0ef4537cf188b488701d6868430bb), I'm not sure this is actually a memory leak. The worker associated with task processor is intentionally not destroyed after creating the geometry because it must remain around to do other work. This is why the unit tests are failing in this branch: They are trying to update the primitive, but can't since the resources have been destroyed. 

If you're interested in getting better performance while your app is idle, I would just [request render mode](https://cesium.com/blog/2018/01/24/cesium-scene-rendering-performance/). If you're still seeing problems, I would ask that you take this to the [Cesium Forum](https://community.cesium.com/) where there are members of the community and developers from the team who can help troubleshoot.

I'm going to close this PR for now.

H91011: then why its not release memory ?  in my topic at forum 
I share result after add 1k line (each line has 1k vetex) it still hold 4GB memory is this normal ? I dont think this amount of objects hold 4GB memory. And when you delete it it still hold 4 GB memory. if you want to explain  I wait here  https://meet.google.com/han-wszx-erd

ggetz: See these explanations on the forum of [why immediately destroying a primitive does not immediately show up in the profiler](https://community.cesium.com/t/why-gpu-memory-no-release/16866) and [why we do not destroy web workers agressively](https://community.cesium.com/t/webworkers-memory-issues/11683).

H91011: ok nevermind, I work on project like this about 3 years and our application does not have leak like this. It is a leak. If you dont want its ok.

PR #9988, opened 2022-01-03
Move upAxis, forwardAxis, and transform into ModelComponents

lilleyse:
Previous upAxis and forwardAxis were attached to the Scene object and transform was attached to the loader. Neither of these seemed like the right place. They are now part of the Components object.

### DIFF
--- a/Source/Scene/ModelComponents.js
+++ b/Source/Scene/ModelComponents.js
@@ -694,6 +694,9 @@ function Components() {
 
   /**
    * A world-space transform to apply to the primitives.
+   *
+   * @type {Matrix4}
+   * @private
    */
   this.transform = Matrix4.clone(Matrix4.IDENTITY);
 }
--- a/Source/Scene/ModelComponents.js
+++ b/Source/Scene/ModelComponents.js
@@ -1,6 +1,7 @@
 import Cartesian3 from "../Core/Cartesian3.js";
 import Cartesian4 from "../Core/Cartesian4.js";
 import Matrix3 from "../Core/Matrix3.js";
+import Matrix4 from "../Core/Matrix4.js";
 import AlphaMode from "./AlphaMode.js";
 
 /**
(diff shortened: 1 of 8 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-components-transform/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelComponents.js
```diff
@@ -690,6 +675,27 @@ function Components() {
...
+  /**
+   * The model's forward axis.
+   *
+   * @type {Axis}
+   * @private
+   */
+  this.forwardAxis = undefined;
+
+  /**
+   * A world-space transform to apply to the primitives.
```
sanjeetsuhag, line 696:
```suggestion
   * A world-space transform to apply to the primitives.
   *
   * @type {Matrix4}
   * @private
```

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -181,7 +181,7 @@ function initialize(model) {
     .then(function (loader) {
       Matrix4.multiply(
         model._modelMatrix,
-        loader.transform,
+        loader.components.transform,
```
sanjeetsuhag, line 184: Why does the `loader` need to persist again? Is it because it may still be loading textures etc. after the model has initialized? When implementing transcoding for I3DM, it'll need to pass this `components.transform` as a uniform, and it's not ideal to have to extract that from the loader (or maybe the scene graph since it also keeps the `components` as a property, however, it's not passed into any pipeline stage yet). Either way, I'd just like to double check that the loader does indeed need to persist and that we can't just hand off the components to the model.

lilleyse: The loader needs to persist for incremental texture loading and so that the loader can be destroyed when the model is destroyed, for cache reference counting purposes. But it's also fine if loader hands off components to model since the loader shouldn't be accessed besides those two cases.

Source/Scene/ModelComponents.js
lilleyse: Updated

lilleyse:
We were just waiting for the release to be complete before merging this. Merging now.

Issue #9298, opened 2020-12-23
CesiumWidget no released from memory when destroyed?


hpinkos:
From https://community.cesium.com/t/releasing-cesiumwidget-object/11831

Hello

I’m working on an application where Cesium viewer is destroyed and created again later in some scenarios. When searching for memory issues I have found that CesiumWidget object is still retaining after the viewer is destroyed via viewer.destroy().

It is reproduced in the simple [sandcastle example](https://sandcastle.cesium.com/#c=hZFBa4QwEIX/SvCigkT3XFdK91AKhR4sPXnJmtltaEwkmWSxZf97o6tl2RWa25t5b+ZL4pkhXsAJDNkSBSeyAytcRz+mWtJE7aR3WiETCkwTpQ+NapQPOaEQjGdyKtRM8ZZZlEAZ5+9ayz0zTw5RqyR+1nFGDk61KIJMyU+jSDjLgLDaAr7MKrk3kpmRcrBo9JCMENeNVfj4hj1eUueMbIqiGOU5/Z++Rt2v8rcSmPnDXm4zj42yqLQ4SKgW1EfR9dogcUYmlOYIXS8Zgs33rv0CpK21C2GZX0dLLjwRfLvyG4GBWRs6BydlLb6hiaoyD/67qNSMC3V882AkG0bb56Z6vRQppWUe5HoSL89xM/kX)

Click Go, wait for several seconds, click Stop, make a heap snapshot. There are multiple CesiumWidget objects are retaining. (Checked via Chrome Dev Tools):

wuming2019:
So this problem can't be solved ？

ggetz:
Hi @wuming2019, we haven't gotten the chance to work on this issue directly. If you are able to look into the issue and provide more details, or even propose a fix, that would be helpful.

ItayBa:
Any updates?

Issue #5991, opened 2017-11-20
Handle lost webgl contexts more gracefully


ggetz:
When a webgl context is lost, an error is thrown that is not really relevant.

![image](https://user-images.githubusercontent.com/4439461/33042682-b489482a-ce10-11e7-8331-0db3c41b789e.png)

Brought up on the forum here: https://groups.google.com/forum/#!topic/cesium-dev/406L37WfjWs

It looks like the lost context can be specifically checked: https://www.khronos.org/webgl/wiki/HandlingContextLost

pjcozzi:
Cesium is unlikely to fully restore the state as I suspect potentially doing it in the application is the right level of abstraction, e.g., an app can easily reload a CZML file compared to Cesium tracking and recreating all the renderer/primitive/entity state.

However, we may expose this event if it is widely requested.

Most engines do not implement this AFAIK.

emackey:
SketchFab shows a friendly dialog box saying something along the lines of "Sorry, but it appears your browser has had a problem running WebGL.  Please reload this page to restore functionality." and there's a RELOAD button to click.  I realize this is an app-level concern but maybe we could expose some kind of helper for this.

thw0rted:
I recognize this isn't absolutely universal but I suspect most consuming "apps" use a Viewer (or a directly-managed CesiumWidget) as a cornerstone of the page being viewed.  Since the app is likely to be more or less useless with a crashed WebGL, you could show the default error handler with a message along the lines of Ed's suggestions ("Something has gone wrong with your browser's 3D graphics; reloading this page will likely fix the problem.", etc).

That said, if it's a  Viewer, I would naively expect it to be possible to destroy the underlying Widget and re-populate it by interrogating `viewer.dataSources`.  That would be far from guaranteed to get back to previous state, but it might be a better user experience than "please click the refresh button".

tinco:
Hi! We're running into this problem as well. I'd like to catch the webgl context lost event as well. I don't need Cesium to gracefully recover of anything, just something that we get control back and I can reboot cesium would already help a lot.

Also, what kind of thing causes webgl contexts to be lost? It almost never happens to me on my MacBook, nor on my colleagues laptop, but our business development guy that does product demo's frequently has them (as you can imagine this is awkward for us) on his surface book. I tried reducing the GPU memory limit, to 256, and that didn't work. We're working with complex models with high resolution textures.

thw0rted:
[...]

PR #10011, opened 2022-01-15
fixes crash issue when create Viewer too many times

renjianqiang:
I find that viewer will crash when i created too many viewers (the number is 16 in chrome) even i had destroy the previous viewers. Look at these code:
``` js
var viewerOldest = new Cesium.Viewer("cesiumContainer");

// then i create some other viewers, and i destroyed them when i didn't need it.
// the code can be simplified to:
for ( var  i =0; i<16; ++i){
   var viewer = new Viewer("cesiumContainer");

   // do something here...

   viewer.destroy();
}
```
 Then, the earliest created viewer will crash. I find the chrome console output the warning:

> WARNING: Too many active WebGL contexts. Oldest context will be lost.

After search some keyword, i think this crash is caused by that webgl context is not released when viewer destroyed. I make this  PR to fix this problem, please let me know if i have mistakes.

### DIFF
--- a/Source/Widgets/CesiumWidget/CesiumWidget.js
+++ b/Source/Widgets/CesiumWidget/CesiumWidget.js
@@ -9,6 +9,7 @@ import Ellipsoid from "../../Core/Ellipsoid.js";
 import FeatureDetection from "../../Core/FeatureDetection.js";
 import formatError from "../../Core/formatError.js";
 import requestAnimationFrame from "../../Core/requestAnimationFrame.js";
+import RuntimeError from "../../Core/RuntimeError.js";
 import ScreenSpaceEventHandler from "../../Core/ScreenSpaceEventHandler.js";
 import createWorldImagery from "../../Scene/createWorldImagery.js";
 import Globe from "../../Scene/Globe.js";
@@ -373,7 +374,16 @@ function CesiumWidget(container, options) {
         that.showErrorPanel(title, undefined, error);
       }
     };
+
     scene.renderError.addEventListener(this._onRenderError);
+    canvas.addEventListener("webglcontextlost", function () {
+      that._onRenderError(
+        scene,
+        new RuntimeError(
+          "The browser lost webgl context, please refresh the page to restore functionality!"
+        )
+      );
+    });
   } catch (error) {
     if (showRenderLoopErrors) {
       var title = "Error constructing CesiumWidget.";


cesium-concierge:
Thanks for the pull request @renjianqiang!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/contextLostCrash/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @renjianqiang! It sounds like you're running into an issue we're tracking where [Cesium should better handle lost WebGL contexts](https://github.com/CesiumGS/cesium/issues/5991).

If I understand correctly, its up to the browser to handle the contexts and this change is only simulating the context lost event. See https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_lose_context/loseContext
> The WEBGL_lose_context.loseContext() method is part of the WebGL API and allows you to simulate losing the context of a WebGLRenderingContext context.

The fix would be to handle the event and destroy object gracefully, perhaps restoring the context if we do have the mechanism to do so.

renjianqiang:
@ggetz, yes, you are right, i just simulate the context lost event to completely free the webgl context when destroying the `Context` object, otherwise too many destroyed viewers will cause the current rendering viewer which previously created crashed. It is unreasonable because these viewers have been destroyed and shouldn't have effect anymore.

And for restoring context problem, i agree with @pjcozzi that it's unlikely or unnecessarily to fully restore the render state. If we want caution users more elegantly after webgl context lost, which by accident or there are too many active webgl contexts simultaneous, we can listen for `webglcontextlost` event  and give some tips.

ggetz:
@renjianqiang Regarding the process of destroying the viewer, I think the actual fix would need to address the root cause of viewer holding onto extra memory, (see https://github.com/CesiumGS/cesium/issues/9298) which would be more involved.

For the WebGL context, I think we would want to remove the `loseContext` call and instead properly handle the `'webglcontextlost'` event and warn the user, as you suggest. Would you be willing to work on that change here?

renjianqiang:
I‘m pleasure to add some handler for&nbsp;'webglcontextlost' event.
Up to now i think the&nbsp;loseContext&nbsp;call is&nbsp;necessary, maybe the&nbsp;viewer holding onto extra memory is the root cause, and i will try to dig deep into the #9298 in the next few days.&nbsp;

---Original---
From: "Gabby ***@***.***&gt;
Date: Thu, Jan 20, 2022 23:22 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [CesiumGS/cesium] fixes crash issue when create Viewer too manytimes (PR #10011)




 
@renjianqiang Regarding the process of destroying the viewer, I think the actual fix would need to address the root cause of viewer holding onto extra memory, (see #9298) which would be more involved.
 
For the WebGL context, I think we would want to remove the loseContext call and instead properly handle the 'webglcontextlost' event and warn the user, as you suggest. Would you be willing to work on that change here?
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

renjianqiang:
@ggetz, after a few days of trying to figure out the issue #9298, i make sure you are right that the extra memory is the root cause. The root cause is [detached window memory leaks](https://web.dev/detached-window-memory-leaks/).

In short, the cesium widget and other object keep a reference to a DOM element which prevents the element from being garbage collected since it can still be accessed, i.e., canvas is not destroyed even the viewer has been destroyed. So the webgl context is not released by chrome. 

I try to fix this problem by set the cesium object's properties to undefined when object is destroyed and it worked, the cesium objects are disappeared in heap snapshot as expected,  the webgl context has lost after viewer destroyed. 

But the question is that i don't know if the problem has been fully solved, and there are must more files require modify, and after that i find the some test case also need be modified otherwise failed. I think i should create another PR and spend more time on it.

ggetz:
@renjianqiang Yes, I think it's best to address #9298 separately. I would also suggest summarizing your fix there before opening a PR so we can discuss the right approach and test cases. Thanks!

renjianqiang:
Ok, my fix to this issue is to modify the `destroyObject.js` :
```js
// function destroyObject
// ...
for (var key in object) {
  if (typeof object[key] === "function") {
    object[key] = throwOnDestroyed;
  } else if (typeof object[key] === "object") {
    Object.defineProperty(object, key, { value: undefined }); // add this line to unset references
  }
}
```
And then i do some modifications to other files:
```js
// Source\Scene\Cesium3DTileset.js
// Cesium3DTileset.prototype.destroy 
// line 2894
while (stack.length > 0) {
  var tile = stack.pop();
  // tile.destroy();

  var children = tile.children;
  var length = children.length;
  for (var i = 0; i < length; ++i) {
    stack.push(children[i]);
  }
  tile.destroy(); // move to here because of tile.children is undefined after tile is destroyed
}

```

```js
// Source\Widgets\Viewer\Viewer.js
// Viewer.prototype.destroy
// line 1755
if (defined(this._selectionIndicator)) {
  this._element.removeChild(this._selectionIndicator.container);
  this._selectionIndicator = this._selectionIndicator.destroy();
}

// destroy inspector if it has attached to viewer
if(defined(this.cesiumInspector)){
  this.cesiumInspector.destroy();
}

if(defined(this.cesium3DTilesInspector)){
  this.cesium3DTilesInspector.destroy();
}

```

And the follow is partial modifications for test case:

```js
// Specs\Scene\Cesium3DTilesetSpec.js
// "destroys before external tileset JSON file finishes loading"
// line 2468
expect(statistics.numberOfPendingRequests).toEqual(1);
var rootContentReadyPromise = root.contentReadyPromise; // add line to record promise before set to undefined
scene.primitives.remove(tileset);

// return root.contentReadyPromise
return rootContentReadyPromise
// ...
```

```js
// Source\Scene\Primitive.js
// function setReady(primitive, frameState, state, error) {
// line 2555
frameState.afterRender.push(function () {
 // add handler if primitive is destroyed during current render frame
  if (primitive.isDestroyed()) {
    return;
  }
// ...
```
Please review my modifications, and give me some advice, Thanks @ggetz !

renjianqiang:
Now, the cesium object will gc successfully after destroyed:
image

But it looks like there are still some object is unreleased:
image

renjianqiang:
I alse find there are maybe some problems in `Source\Widgets\ClockViewModel.js` and `Source\Widgets\Animation\AnimationViewModel.js`.

```js
// Source\Widgets\ClockViewModel.js
// line 40
this.startTime = knockout.observable(clock.startTime);
this.startTime.equalityComparer = JulianDate.equals;
// this.startTime.subscribe(function (value) {
//   clock.startTime = value;
//   this.synchronize();
// }, this);

// maybe we should record the subscription and dispose in destroy function
this._subscriptions.push(
    this.startTime.subscribe(function (value) {
      clock.startTime = value;
      this.synchronize();
    }, this)
  );

// ClockViewModel.prototype.destroy 
// line 189
var subscriptions = this._subscriptions;
for (var i = 0, len = subscriptions.length; i < len; i++) {
  subscriptions[i].dispose();
}
```

```js
// Source\Widgets\Animation\AnimationViewModel.js
// maybe we should add destroy function for AnimationViewModel and call it in Animation.prototype.destroy function
AnimationViewModel.prototype.destroy = function () {
  var that = this;
  this._definedProperties.forEach(function (property) {
    knockout.getObservable(that, property).dispose();
  });
  
  destroyObject(this);
};

```

I'm not sure if we need to make this change, and i'd like to have your suggestions!

cesium-concierge:
Thanks again for your contribution @renjianqiang! 

cesium-concierge:
Thanks again for your contribution @renjianqiang! 

cesium-concierge:
Thanks again for your contribution @renjianqiang! 

ItayBa:
Did anyone solved it?

cesium-concierge:
Thanks again for your contribution @renjianqiang! 

anthonylowther21:
Can this PR either be rejected or accepted?

We have this exact issue. When the map is loaded and unloaded too many times, the first map screen turns white, and we get the rror that too many webgl contexts exist

cesium-concierge:
Thanks again for your contribution @renjianqiang! 

ggetz:
I'm closing this issue due to inactivity. If you believe this is still an issue, please feel free to re-open and we'll give this a fresh review. Thanks!

anthonylowther21:
@ggetz, this is still an issue on our projects. Did this PR get rejected, or did this get resolved in a separate PR?

danrockstheplanet:
@ggetz, I'm also interested in knowing the answer to @anthonylowther21's question. Has this been resolved? If so, can you tell us when?

Issue #9884, opened 2021-10-21
`Model` EXT_mesh_features: handle feature IDs without a property table


ptrgags:
in #9872 I noticed that supporting feature IDs without property tables (a new detail in `EXT_mesh_features` is a little more involved than expected:

* missing property tables should be created as empty tables
* need to create a list of ModelFeatureTable or content feature tables across the whole model
* Need to make sure the new generated indices are stored on the attributes

#9872 was getting large, and also I want to make sure I have recent changes from #9873 so just leaving a note for this now. #9880 is a higher priority right now.

ptrgags:
One subtlety I'm noticing: for picking, the model has `featureIdAttributeIndex` -- does that correspond to `FEATURE_ID_n` or is that supposed to be the index within the `featureIds` array (which can also include implicit feature ID attributes and feature ID textures)? That's something that we should sort out

ptrgags:
Some more specific notes about what needs to be updated (part of a larger set of notes I'll post in another issue)

## parseFeatureMetadata

- If `extension.featureIds.length > extension.propertyTables.length`, create empty property tables to make up the difference

## GltfLoader

- In `loadPrimitiveMetadata`, check if `propertyTablesArray[i]` exists before pushing.
- Similarly for `loadInstanceMetadata`

ptrgags:
opened #10007 for the larger details about handling multiple sets of feature IDs. This issue should be addressed first.

ptrgags:
Today I hit a major snag: with the current picking system, it's unclear how to handle the batch texture since the total range of values is unknown (no `propertyTable.count`), potentially large and sparse (think unique ids into an external database, not necessarily contiguous).

Talking with @lilleyse, for right now we'll skip picking (though I do still plan to add the value to custom shaders). See also https://github.com/CesiumGS/cesium/issues/9852#issuecomment-1013508890

lilleyse:
Requested on the forum: https://community.cesium.com/t/gltf-feature-extensions-support-in-cesium-ion/22911

lilleyse:
Brought up on the forum: https://community.cesium.com/t/enquiry-on-use-of-extension-ext-mesh-features-without-ext-structural-metadata/44925

PR #10018, opened 2022-01-19
Handle multiple feature ID sets in `ModelExperimental`

### LINKED ISSUE #9894: Adding feature IDs to custom shaders
It may be useful to have access to the feature ID value in the custom shaders. Based on offline discussion with @ptrgags, this is the design we came up with:

```glsl
struct FeatureIds {
  float attributes[3];
  float textures[2];
}
```

This example shows the struct that would be generated for a primitive with 3 feature ID vertex attributes and 2 feature ID textures. Inside the custom shader, the values could be accessed by something like `fsInput.featureIds.attributes[0]` or `fsInput.featureIds.textures[1]`.

A couple of open questions:

1. Is there a better name for `textures`? Currently, this seems to denote that you're accessing a texture, instead of a value extracted from a texture.
2. Does there need to be another member of the struct that denotes the "active" feature ID i.e. the one that is used for picking/styling?

CC: @lilleyse

### LINKED ISSUE #10007: Handle multiple sets of feature Ids in `ModelExperimental`
## Overview

Right now, ModelExperimental only uses feature IDs for the following:
* Using one selected feature ID for picking/styling. This doesn't always work, #9884 needs to be addressed. 
* feature ID attributes (only the explicit `FEATURE_ID_n` ones) show up in custom shaders as `fsInput.attributes.featureId_n`

However, we also want to support the following:
* consistent method for referencing feature IDs. Each primitive has a list of feature ID sets (some from attributes, some from an implicit range, some from a texture). in the API "feature ID n" or "instance feature ID n" should refer to the `n`-th entry in these lists of feature IDs
* multiple feature IDs in custom shaders. This should be seamless, the user shouldn't need to worry about texture reads
* Longer term, having the feature IDs accessible will be an important prerequisite for #9572 and #9944

Here are some initial notes of things that'll need to be updated. There's likely more details to consider as I implement this. 

#9884 should be addressed first, then this issue.

## CustomShader

- Add parsing logic for `[fv]sInput.featureIds.(featureId|instanceId)_(\d+)`
- `featureId_n` corresponds to a position within a primitive's featureIds array
- `instanceId_n` corresponds to a position within the instance's featureIds array 

## FeatureIdPipelineStage

- Maybe split into a `ActiveFeaturePipelineStage` and `FeatureIdPipelineStage`
- Move this stage earlier in the pipeline
- Change this stage to process _all_ feature IDs
- Generate implicit attributes and use a prefix in the shader
- Add a `FeatureIds` dynamic struct
- Attributes: `featureIds.(feature|instance)Id_n = a_featureId_m`
- Implicit attributes: `featureIds.(feature|instance)Id_n = a_implicitFeatureId_m`
- Textures (FS only): `featureIds.(feature|instance)Id_n = floor(texture2D(FEATURE_ID_TEXTURE, FEATURE_ID_TEXCOORD).FEATURE_ID_CHANNEL * 255.0 + 0.5);`

## CustomShaderPipelineStage

- Add `FeatureIds` to `vsInput` and `fsInput`
- Warn if a texture feature ID is used in the vertex shader

## ModelExperimental

- Replace `featureIdAttributeIndex, featureIdTextureIndex` with `featureIdIndex` and `instanceFeatureIdIndex`

## ModelExperimentalFS

- Add a note that `Feature feature;` is global-scoped because of its use in picking (maybe also wrap in a `#ifdef` clause?)

ptrgags:
Fixes #10007 
Fixes #9894

This also partially addresses #9884, though picking is not currently supported for feature IDs without a property table, and the current picking system would need larger changes to address this (see https://github.com/CesiumGS/cesium/issues/9884#issuecomment-1013511841)

Summary of changes:

* In `ModelComponents`, combined `featureIdAttributes` and `featureIdTextures` into one `featureIds` array. This is more consistent with the current `EXT_mesh_features` spec.
* in `ModelExperimental`, replaced `featureIdAttributeIndex` and `featureIdTextureIndex` with `featureIdIndex` and `instanceFeatureIdIndex`, as the latter pair better match the glTF structure
* `FeatureIdPipelineStage` now processes all the feature ID attributes instead of only the selected one. It populates a `FeatureIds` struct that can be passed into both feature ID selection (next bullet) and custom shaders. 
* `SelectedFeatureIdStage` handles the shader details for taking the selected feature ID set and populating a struct used by styling and picking. This is significantly easier now with the `FeatureIds` struct, it's simply `selectedFeature.id = featureIds.SELECTED_FEATURE_ID` (using a define to substitute in the right variable).
* adds feature IDs to Custom Shaders via `fsInput.featureIds.featureId_N` or `fsInput.featureIds.instanceFeatureId_N`
* updated the [3D Tiles Next Photogrammetry Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Next%20Photogrammetry%20Classification.html&label=3D%20Tiles%20Next) to use the new `fsInput.featureIds.featureId_N` syntax to avoid exposing private API.

To Do:

- [x] Update custom shader guide
- [x] Update documentation for feature ID stages
- [x] Fix broken unit tests
- [x] Add new unit tests
- [x] add `featureIdIndex` and `instanceFeatureIdIndex` to Cesium3DTileset for easier configuration
- [x] Update `CHANGES.md`
- [x] Check if instanced `FEATURE_ID_n` is added to the `ProcessedAttributes` correctly
- [x] fix race condition about `featureIdIndex` being set before the scene graph is constructed.
- [x] investigate styling, sometimes higher LODs aren't displaying the right color
- [x] Update other existing sandcastles
- [x] Investigate ModelColorPipelineStage DeveloperError
- [x] Investigate draco tile failures in Custom Shaders 3D Tiles sandcastle -> opened https://github.com/CesiumGS/cesium/pull/10031


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/multi-feature-ids/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

javagl:
Just a short note: I think that http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Next%20S2%20Globe.html&label=3D%20Tiles%20Next is still using the (now no longer working, internal) `FEATURE_ID_TEXTURE` approach for the custom shader.

ptrgags:
@lilleyse this PR is now ready for review. I do still want to add a sandcastle to show multiple feature Id sets once you finish updating the Ferry Building tileset and upload it to ion

Source/Scene/Cesium3DTileset.js
```diff
@@ -963,6 +963,28 @@ function Cesium3DTileset(options) {
...
+   */
+  this.featureIdIndex = defaultValue(options.featureIdIndex, 0);
+
+  /**
+   * The index into the list of instance feature IDs used for picking and
+   * styling. If both per-primitive and per-instance feature IDs are present,
+   * the instance feature IDs take priority.
+   *
+   * @type {Number}
+   * @readonly
```
ptrgags, line 983: This shouldn't be readonly

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -582,32 +578,53 @@ Object.defineProperties(ModelExperimental.prototype, {
...
-   * The index of the feature ID attribute to use for picking features per-instance or per-primitive.
+   * The index into the list of primitive feature IDs used for picking and
+   * styling. For EXT_feature_metadata, feature ID attributes are listed before
+   * feature ID textures. If both per-primitive and per-instance feature IDs are
+   * present, the instance feature IDs take priority.
    *
    * @memberof ModelExperimental.prototype
    *
    * @type {Number}
+   * @readonly
```
ptrgags, line 589: Also not readonly

ptrgags:
@lilleyse I updated the [3D Tiles Next Photogrammetry Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Next%20Photogrammetry%20Classification.html&label=3D%20Tiles%20Next) with new styling and custom shader examples

Source/Scene/ModelExperimental/PickingPipelineStage.js
```diff
@@ -109,23 +108,19 @@ function processPickTexture(renderResources, primitive, instances) {
   var model = renderResources.model;
   var featureTableId;
   var featureIdAttribute;
-  var featureIdAttributeIndex = model.featureIdAttributeIndex;
+  var featureIdIndex = model.featureIdIndex;
+  var instanceFeatureIdIndex = model.featureIdIndex;
```
lilleyse, line 112:
```suggestion
  var instanceFeatureIdIndex = model.instanceFeatureIdIndex;
```

Copy-paste mistake?

lilleyse (commented):
Seems to be in pretty good shape overall. Here's my first round of feedback, I plan to test some examples next.

lilleyse:
@javagl just to confirm, have you tested all the samples in https://github.com/CesiumGS/3d-tiles-samples/pull/40 on this branch?

Here's the testing strategy I was going to go with. Are there any that can already be checked off?

* [ ] Test picking feature without corresponding property table (expect it not to pick anything until [#98834](https://github.com/CesiumGS/cesium/issues/9884#issuecomment-1013511841) is addressed)
* [ ] Test setting the active feature id and then picking and styling
* [ ] Test setting the active instance feature id and then picking and styling
* [ ] Test picking pnts with BATCH_ID
* [ ] Test picking pnts with per point properties (expect picking to work now, but maybe not in the future [#10025](https://github.com/CesiumGS/cesium/issues/10025))
* [ ] Test picking i3dm
* [ ] Test picking b3dm
* [ ] Test picking glTF with feature ID attributes
* [ ] Test picking glTF with feature ID textures
* [ ] Test picking glTF with implicit feature IDs
* [ ] Test picking glTF with instance feature ids
* [ ] Test picking glTF with implicit instance feature ids
* [ ] Test custom shader for any of the above

ptrgags:
Addressed the concrete feedback. Next step is investigating a bug @lilleyse pointed out to me where switching between a style and a custom shader leaves the tileset in a bad state:

Peek 2022-01-25 15-43

ptrgags:
@lilleyse this is ready for review again.

ptrgags:
merging in the let/const changes didn't seem to work properly... I'm going to reset my branch and try again

lilleyse:
The new sandcastle option `Multiple Feature ID Sets` might be a little too busy with the color scheme. Maybe it's better to go with something simpler, like tinting roof windows yellow and facade windows purple.

lilleyse:
Aside from that, the recent changes look good. @javagl once you're satisfied with your testing we can get this merged.

ptrgags:
@lilleyse I simplified the sandcastle like you suggested
image

javagl:
I did some tests, _roughly_ based on the bullet points above, but with some disclaimers (summarized below)

* [X] Test picking feature without corresponding property table (expect it not to pick anything until [#9884](https://github.com/CesiumGS/cesium/issues/9884#issuecomment-1013511841) is addressed)
  - This seems to work. based on the `ExplicitFeatureIds` and `ImplicitFeatureIds` samples from https://github.com/CesiumGS/3d-tiles-samples/pull/40, which both define feature IDs without associated properties.

* [X] Test setting the active feature id and then picking and styling
  - This seems to work, based on the (preliminary) `ComplexExample` mentioned in https://github.com/CesiumGS/3d-tiles-samples/pull/40#issuecomment-1021390611 

* [ ] Test setting the active instance feature id and then picking and styling

* [X] Test picking pnts with BATCH_ID
* [X] Test picking pnts with per point properties (expect picking to work now, but maybe not in the future [#10025](https://github.com/CesiumGS/cesium/issues/10025))
  - I *think* these are working as expected. When loading the `cesium/Apps/SampleData/Cesium3DTiles/PointCloud/PointCloudBatched` in my test sandcastle, it shows the "feature ID" (aka Batch ID) and some per-point properties in the Tooltip, if one manages to move the mouse _directly_ over that single pixel :D
 
* [X] Test picking i3dm
* [X] Test picking b3dm
  - These also seem to work, based on the `Instanced/InstancedWithBatchTable` and `Batched/BatchedWithBatchTable` sample data...

* [X] Test picking glTF with feature ID attributes
* [X] Test picking glTF with feature ID textures
* [X] Test picking glTF with implicit feature IDs
  - These seem to work based on the `ExplicitFeatureIdsAndSimpleProperty`, `ImplicitFeatureIdsAndSimpleProperty` and `SimpleFeatureIdTexture` samples from https://github.com/CesiumGS/3d-tiles-samples/pull/40

* [ ] Test picking glTF with instance feature ids
* [ ] Test picking glTF with implicit instance feature ids
* [ ] Test custom shader for any of the above

---

Disclaimers:

- In general, I wasn't sure about the interplay of metadata and styling. For example, there is this [line with `color: "color(${color})"`](https://github.com/CesiumGS/cesium/blob/2a61e4f8f2694d757c6066b6a8fc7feabe53b018/Apps/Sandcastle/gallery/3D%20Tiles%20Next%20Photogrammetry%20Classification.html#L88) in one of the samples, and I don't know what `color` is. (No not that one, the other one. I mean the innermost one...)
- Setting a style with a `pointSize: 10.0` does not seem to have any effect:

        pointSize: 10.0
      });

  Is this a known issue?

- Everything related to *instance* IDs: I'll have to look into that, in terms of example data and usage.
- Custom shaders: Also would need a bit more time here to try out "things"...

---

General: 

I've been testing most of this based on the (sketchy) "Test Sandcastle" that may, at some point, be a "template" for the example sandcastles that will go into the READMEs for the samples. 

There are several issues with that:

Right now, accessing the metadata generically is ... clumsy: It has to do many `instanceof/undefined` checks, and a *lot* of "implicit, specific knowledge" is carved into this code. It might be that there are some abstractions in the latest state of this PR. If some of that code could be simplified, that would be great. 

There is an experimental function for "Setting the active feature ID", but ... I have no idea how to know _the actual number of available feature IDs_ in the first place. More generally: I assume that the concept of having an "active feature ID" is a tribute to having to transport this information through the rendering process into "the picking world". I could question the now 54 (fifty-four!) properties in the `options` for a tileset, and ask who is going to create a "mental model" of what they mean and what values they should have. But even if it is OK to just add more properties there, this is part of a "global state" (global in terms of the tileset), and I see this as a (possibly severe) limitation. What if someone wants to do _styling_ based on _one_ feature ID, and show a _tooltip_ based on _another_ feature ID? (Related to that: The ambiguities between "feature ID" and "instance feature ID" are ... not so nice. Conveying the difference and their meaning in context, in some sort of documentation/tutorial, may be difficult...)

It might be that these issues are already resolved by more recent changes. So to put it that way: If somebody could improve (as in: _simplify_ and _generalize_) this test code at https://github.com/javagl/3d-tiles-samples/blob/add-ext-mesh-features-samples/EXT_mesh_features/CommonSandcastleCode.txt , that would be great...

ptrgags:
@javagl a couple things to point out:

* Styling with property names should work for meshes, but not for point clouds, that falls under #9944 
* As far as picking goes, indeed the metadata is cumbersome to access, but this PR is only concerned with
* pointSize in `Cesium3DTileStyle`... :grimacing: I don't think that was implemented. I'll open an issue.

lilleyse:
I finished up testing. All looks good.

> * [x] Test setting the active instance feature id and then picking and styling
> * [x] Test custom shader for any of the above
> * [x] Test picking glTF with instance feature ids ([Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=vVVhb9s2EP0rhD/JQ0JJsWMnaRosTZPUW4IVtbENm4aAkc4WEYoUSMpxWuS/70hKiucm6LcZhiSSd7x3996RuZLGkjWHR9DkPZHwSC7A8Kaiv/u5KBvkfnyhpGVcgs4Gw3eZzGTwoblQ+QPNG61B2gWvAHdpd/ilEZzJj8wCXWpVzYw6miRplElCssFBcpDup+l+crw4SE4OpiejKU3SSTIZj6eHo/HR5Gh8PJr+lQ0yGeLFMblVBQiTydyDrvwIw/3tdvzmHoRY2NgT3D5AIB+YEErJbLAXltERBORosmxkbrmSJBp2voSERR8m6uYcWErj8J/XkJsYU2JxABNfC7u8UawAHX9QmxkCYzKHIl6JxVV8rzb7vJu6u/xzcbcEZhsNdxVYVuA2dIX+PTz3Sw+7b5e3ez+7Fz7+8XUIyZf1Dl2fgBVcrj5zm5dflAj427VbZktq1Re0YNJE6VEy9AGT8AwVDvsu+QaKK80qWGi0XSpdvRDaTxmKrDMR7NRV73MNKBBmlW5Jlkrbss0uGzyCsS98Cmgp9KOejm0KGi32SAl8VdqWpACyVoZ74x7ZBdMWv5gceal9hJUGMC2H++nBiCbT8XiSHreFHo9pcpiMpsmknQhR3HdAR9qeoCbHnGiteYUh12Cohkqt4RwrHPjxOSCSt+xZUbQ4Ojacw+WmBrTBnmHCQ3Y6inohOlWcEJd/N+PDIJGab05e4aPcoX+blS0pd5Xb0hsqaWvU7nwpBK+N4gX943p+NN4yeEUgvV5bs+fhdh09cKwaK54+Y57cALUlyOilAb1F34XduYIRNEOdqYdz22fwHf63OoDJ1XberzfC+HC49yOb/fQ/RiGbe9VIF2tel6CBarRtDPmJvLRu+AgKeXavZ1eNOZNFzowV4GSxUErcM30Lsgk1MMOtDvdC2tFV36YlbiR2T+x5jqKX85rlcLlGZX0KRlGQZM7kmvkIrTM1YGeybuy556EnJHL6dspsKQkpd8fYVTjAZsVMFrBBAInP0XVzzfMHKHCq7QEc9nvRjrlQEr4kUYu6gCVeK0UUvIe9DlyiCgsl1Kpbe9cehD/ElHrL/w3P895bFCyeaqA3l1eLu4ub2cWvzmOwNzg19knAWSeWn3lV40Hp2j3CO8ZCVQu8NE1832AYS3NjuqvgNN52PS34mvDi/St3NMkFMwZXlo0Qc/4VssHZaYz237kK5dvmtzVowZ6cWZme3YRJSulpjMPXPW3Q787O/wI))
> * [x] Test picking glTF with implicit instance feature ids (Same sandcastle as above)

PR #10022, opened 2022-01-20
Add a feature ID getter for picking

ptrgags:
This PR adds two getters:

* `Cesium3DTileFeature.featureId` (before this, it was only accessible through the private `pickedFeature._batchId`)
* `ModelFeature.featureId` (formally the private `pickedFeature._featureId`)

@lilleyse two questions on this one:

1. Right now I marked these as experimental in case we have better ideas of how to expose feature ID values to the public API. Do we want this experimental flag or is this a good addition to the public API?
1. I also realized it would be easy to expose `featuresLength` as well to get the total count of features in the `Cesium3DTileBatchTable`/`ModelFeatureTable`. Do we want this too?

[Updated photogrammetry classification sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Next%20Photogrammetry%20Classification.html&label=3D%20Tiles%20Next)

### DIFF
--- a/Apps/Sandcastle/gallery/3D Tiles Next Photogrammetry Classification.html
+++ b/Apps/Sandcastle/gallery/3D Tiles Next Photogrammetry Classification.html
@@ -256,7 +256,7 @@
                 "Component: " +
                 pickedObject.getProperty("component") +
                 "\nFeature ID: " +
-                pickedObject._batchId;
+                pickedObject.featureId;
               nameOverlay.textContent = message;
             } else {
               nameOverlay.style.display = "none";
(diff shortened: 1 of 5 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/feature-id-getter/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
> 1. I also realized it would be easy to expose `featuresLength` as well to get the total count of features in the `Cesium3DTileBatchTable`/`ModelFeatureTable`. Do we want this too?

My only hestitation is that `featuresLength` would not be well defined for feature IDs without an associated property table.

(That is, once we actually get picking working in that case. CC https://github.com/CesiumGS/cesium/issues/9884#issuecomment-1013511841h)

I'm good with merging this PR before we make a decision on `featuresLength`.

Issue #9718, opened 2021-08-10
Plan to switch to ES6 syntax


ebogo1:
Now that IE support has been dropped in 1.84, we should make a plan to update our code. This issue can be a place to brainstorm ideas for these changes.

There are some things that need to happen before we can start making changes in CesiumJS:

- [ ] Update coding guide
- [ ] Update eslint config

Other things to think about:

- The `when` library is not completely compatible with native promises, use of `async` and `await` might need to be delayed until we get rid of `when`.
- Some ES6 features might be a performance downgrade from ES5
- Can we make the switch from `var` to `let` and `const` now? Some of this effort already started with the last gltf-pipeline PR: https://github.com/CesiumGS/gltf-pipeline/pull/600.

ptrgags:
I'm curious about the switch to `let` and `const`, is this going to be a change for new code moving forward, or are you translating all the old code to use `let` and `const`? (are there tools for this? sounds like a big task otherwise)

I'm also curious to see when we can use `for ... of` loops for objects instead of the rather verbose:

```
for (var key in obj) {
    if (obj.hasOwnProperty(key)) {
    }
}
```

And other ES6 features (e.g. spread syntax where right now we use shallow `combine()`... there's a lot of helpful things to consider here (of course, weighing this against any performance differences)

onsummer:
I'm glad to see the official intention to update the syntax of the source code.

Using `let` and `const` avoids the problem of variable promotion in JavaScript. 

`const` maintains the immutability of variables:

```js
import { Cartesian3 } from 'cesium'

const coords = new Cartesian3(100, 200, 0)
coords = 0 // Uncaught TypeError: Assignment to constant variable.
```

 If you use `var` to make the code run normally, there is generally no need to change it to `let` and `const`.

For the `for ... of` mentioned by @ptrgags , I have seen in some evaluation reports that the performance of iterative access to the object's `key` and `value` is acceptable, but the performance of iterative access to the array may not be as good as that of `for ( let i = 0; i <someLength; i++) {/* some code */ }` and the `Array.prototype.forEach` method.

Modularization technology has been changed from `requirejs` to `esmodule` in version 1.63, which has done a good job.

The above are some of my simple views, I hope it will be useful to the official.

I have used the `vitejs` packaging tool, and have used `CesiumJS` in the `reactjs` and `vuejs` frameworks. Adding some guidance documents combined with these front-end frameworks maybe a good idea.

---

[...]

PR #10026, opened 2022-01-21
Change "var" to "let"/"const" ES6 syntax

ggetz:
Part of https://github.com/CesiumGS/cesium/issues/9718

Now that IE support has been dropped, there's nothing holding us back from using ES6 features. This PR focuses on moving to `let` and `const` syntax so that it can be used across the board going forward.

The primary change here was to enable `no-var` and `prefer-const` error checking in eslint, and fixing those errors. I also updated the Coding Guide to reflect the technicality of scoping, and adding a bullet about why `const` is preferred when possible.

Note that by turning on the above two rules, I had to disable `es/no-block-scoped-variables` for `Source` and `Spec` modules as Cesium makes use of may non-function scoped variables such as scratch variables. I believe this is safe to disable as these variables are all module-scoped. I also had to bump `ecmaVersion` for `Apps` to 2015/ES6 in order to support `let`/`const` usage in Sandcastle and other demos.

For ease of review, the primary changes were made in these two commits:
* [Coding Guide updates](https://github.com/CesiumGS/cesium/commit/7711d2c50db16b2f3f87d88ff7eabe69e9b82e81)
* [eslint config changes](https://github.com/CesiumGS/cesium/commit/3de446615cc0506818646e3df3334f0fdb426c53)

And all of the corresponding fixes were made in [this commit](https://github.com/CesiumGS/cesium/commit/8143df4436b79260e9861f63c6ac134d7910a818).


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/let-const/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
> Will there be a way to easily update open branches? I recall when Prettier was first added, there were tags before and after the change

The commits here are split up such that https://github.com/CesiumGS/cesium/pull/10026/commits/3de446615cc0506818646e3df3334f0fdb426c53 contains the configuration changes but no fixes, and https://github.com/CesiumGS/cesium/pull/10026/commits/8143df4436b79260e9861f63c6ac134d7910a818 contains all the fixes. We could tag these commits and be able to follow a similar process as https://community.cesium.com/t/cesiumjs-source-code-is-now-formatted-with-prettier/9362, substituting `npm run eslint -- --fix` for `npm run prettier`. The main difference is that we can't fully automate the fix step-- eslint is able to detect and fix most occurrences of `var`, but not all of them. So an author may need to do some manually fixes.

mramato:
Doesn't have to be this PR, but the `disable prefer-const` in `Particle System Weather.html` is only needed there because the code itself is not well laid out and the example should be cleaned up so that it flows better. I would just write up a clean-up issue and it's probably a "good first issue" too.

mramato:
Where we actually need it, we should use `eslint-disable-next-line prefer-const` to disable the next problematic line, rather than disabling eslint completely and needing to use `eslint-enable` on a separate line.

mramato:
While we probably want to take a hard look at Cesium's eslint rules in our new post-IE11 world, I think that is outside the scope of this PR and those few comments are the only ones I had.

I haven't actually run this yet but will do so once `main` is merged in and then I'm happy to merge assuming no runtime problems pop up.

ggetz:
> Where we actually need it, we should use eslint-disable-next-line prefer-const to disable the next problematic line, rather than disabling eslint completely and needing to use eslint-enable on a separate line.

I believe the only places I've used it are in Sandcastle examples. There only there for the exact reason you mention above with the code needing cleanup to have a better flow. I added them to be scoped to the entire example so they don't show up in the Sandcastle code itself. But if we're OK with that, happy to change it to what you suggest here.

mramato:
There was a spot in `Camera.js` I noticed (just search for prefer-const), the Sandcastle can probably stay for now since we need to clean it up anyway.

ggetz:
@mramato `main` is merged in and this is ready for another look.

mramato:
It's the beginning of a whole new era.

PR #10045, opened 2022-01-27
Particle System Sandcastle ES6 Update

### LINKED ISSUE #10038: Particle System Weather Sandcastle code needs cleanup
The "Particle System Weather" is structured in such a way that we need to disable the `prefer-const` rule to allow it to work as is. Really, the code needs to be restructured in such a way that disabling the rule is not necessary.

(See https://github.com/CesiumGS/cesium/pull/10026#issuecomment-1022406834)

srothst1:
Fixes #10038

The goal of this pull request is to re-work the Particle System Weather sandcastle demo such that it is functional when the `prefer-const` ES6 rule is enabled.

My initial commits have made the following changes:
- Removed the line disabling one of our global ES6 rules: `/* eslint-disable prefer-const */`
- Updated where `snowGravityScratch` and `rainGravityScratch` are declared.
- Removed the `snowSystem` and `rainSystem` variables.
- In the original implementation, both the `snowSystem` and the `rainSystem` objects were added to the scene as primitives. Their visibility was toggled using the `show`. I suspect that this may have caused performance issues. I added a commit that ensures that only the necessary primitives are kept in the scene.

I am now able to run the Particle System Weather demo locally without the line:
`/* eslint-disable prefer-const */`

It has been mentioned various times that the [current code](https://sandcastle.cesium.com/?src=Particle%20System%20Weather.html) is not well structured. As seen above, I made various edits. The biggest thing that stood out to me was how primitives were handled. However, if there is anything specific that still needs to be reworked, I would be happy to take a look at it.

### DIFF
--- a/Apps/Sandcastle/gallery/Particle System Weather.html
+++ b/Apps/Sandcastle/gallery/Particle System Weather.html
@@ -34,7 +34,6 @@
     <script id="cesium_sandcastle_script">
       function startup(Cesium) {
         "use strict";
-        /* eslint-disable prefer-const */
         //Sandcastle_Begin
         const viewer = new Cesium.Viewer("cesiumContainer", {
           shouldAnimate: true,
--- a/Apps/Sandcastle/gallery/Particle System Weather.html
+++ b/Apps/Sandcastle/gallery/Particle System Weather.html
@@ -68,10 +68,8 @@
           snowParticleSize * 2.0,
           snowParticleSize * 2.0
         );
-        let snowSystem;
-
-        let snowGravityScratch = new Cesium.Cartesian3();
         const snowUpdate = function (particle, dt) {
+          let snowGravityScratch = new Cesium.Cartesian3();
           snowGravityScratch = Cesium.Cartesian3.normalize(
             particle.position,
             snowGravityScratch
@@ -86,7 +84,6 @@
             snowGravityScratch,
             particle.velocity
           );
-
           const distance = Cesium.Cartesian3.distance(
             scene.camera.position,
             particle.position
@@ -99,7 +96,7 @@
           }
         };
 
-        snowSystem = new Cesium.ParticleSystem({
+        const snowSystem = new Cesium.ParticleSystem({
           modelMatrix: new Cesium.Matrix4.fromTranslation(
             scene.camera.position
           ),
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/particle-system-es6/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Apps/Sandcastle/gallery/Particle System Weather.html
```diff
@@ -95,42 +91,20 @@
             particle.endColor.alpha = 0.0;
           } else {
             particle.endColor.alpha =
-              snowSystem.endColor.alpha / (distance / snowRadius + 0.1);
+              Cesium.Color.WHITE.withAlpha(1.0).alpha /
```
ggetz, line 94: This will always evaluate to `1.0`, so let's simplify:
```suggestion
              1.0 /
```

Apps/Sandcastle/gallery/Particle System Weather.html
```diff
@@ -68,10 +67,8 @@
           snowParticleSize * 2.0,
           snowParticleSize * 2.0
         );
-        let snowSystem;
-
-        let snowGravityScratch = new Cesium.Cartesian3();
         const snowUpdate = function (particle, dt) {
+          let snowGravityScratch = new Cesium.Cartesian3();
```
ggetz, line 71: I like the intention of scoping the variable to the block where it's needed, but this was scoped a level up on purpose. See [the section on scratch variables in the Coding Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#result-parameters-and-scratch-variables).

ggetz (commented):
Thanks @srothst1! I have a few comments.

Apps/Sandcastle/gallery/Particle System Weather.html
srothst1: Nice catch 👍

srothst1:
Hi @ggetz! Thank you very much for the feedback. I just added various commits and comments that address your suggestions.

ggetz:
Awesome, thanks @srothst1 !

PR #10055, opened 2022-01-31
Replace `Gltf3DTileContent` with `ModelExperimental3DTileContent`

### LINKED ISSUE #10049: Replace Gltf3DTileContent with ModelExperimental3DTileContent
`Gltf3DTileContent` was a temporary holdover to implement the [`3DTILES_content_gltf`](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_content_gltf) extension. This extension simply allows glTF and binary glTF (.glb) models inside a tileset (normally meshes must be in .b3dm or .i3dm format).

Since then, we've created `ModelExperimental` (which will someday replace `Model`), and `ModelExperimental3DTileContent` to integrate this with 3D Tiles. While `ModelExperimental` is not yet at feature parity with the old `Model` (see #9520), it's far enough along that we can always enable it for glTF/glb contents. This way, new features such as custom shaders and GPU instancing are available right away without needing to set `ExperimentalFeatures.enableModelExperimental = true`.

This task involves:

- [x] In `Source/Scene/Cesium3DTileContentFactory.js`, update the `gltf` and `glb` to always use `ModelExperimental3DTileContent`. i.e. the if check can be removed.
- [x] Remove `Source/Scene/Gltf3DTileContent.js`
- [x] Update unit tests
  - [x] Add the "renders glb" unit test to `Specs/Scene/ModelExperimental/ModelExperimental3DTileContent.js`, it looks like this one was missing.
  - [x] Finally, remove `Specs/Scene/Gltf3DTileContentSpec.js`. There's a few other unit tests in this file that don't apply to `ModelExperimental3DTileContent`, but I'll link to them from their respective issues.
  - [ ] Check if any other specs need updating. If anything, it would probably be in `Specs/Scene/Cesium3DTilesetSpec.js`.
- [x] Update Sandcastles as necessary
  - [x] The 3D Tiles Next S2 Globe is a GLB tileset, so you can remove the line `Cesium.ExperimentalFeatures.enableModelExperimental = true;` from the top of the example code.
  - [x] Same thing with 3D Tiles Next Photogrammetry Classification
  - [x] Same thing with 3D Tiles Next CDB Yemen

j9liu:
Fixes #10049.

In this PR, all instances of `Gltf3DTileContent` have been replaced by instances of `ModelExperimental3DTileContent`. The corresponding specs were removed. Additionally, missing tests were added for the `glb` file format, and the 3D Tiles Next sandcastles were updated to no longer require the `enableModelExperimental` flag.

Implementing these changes causes three unit tests to fail:

image

Two of these failures are expected and will be addressed in a later issue. However, the color style test should be working, and @ptrgags mentioned he would look into this tomorrow.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/replace-gltf3dtilecontent/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu:
@ptrgags - Following up on your suggestion about the color style test, specifying the time doesn't fix the test failure.

image

If you want to replicate my test, I put this at the beginning of `expectColorStyle` and changed all instances of `expect(scene)` in the function to `expect(renderOptions)`.
```
      const renderOptions = {
        scene: scene,
        time: new JulianDate(2457522.154792),
      };
```

ptrgags:
@j9liu okay, I'll take a look at this later today

Specs/Scene/Cesium3DTilesetSpec.js
```diff
@@ -1961,11 +1962,16 @@ describe(
       return checkDebugColorizeTiles(pointCloudUrl);
     });
 
-    it("debugColorizeTiles for glTF", function () {
+    xit("debugColorizeTiles for glTF", function () {
```
ptrgags, line 1965: You could put a comment here `// see https://github.com/CesiumGS/cesium/issues/10061` to link to the corresponding issue.

Specs/Scene/Cesium3DTilesetSpec.js
```diff
@@ -2577,7 +2583,7 @@ describe(
       return testBackFaceCulling(withoutBatchTableUrl, setViewOptions);
     });
 
-    it("renders glTF tileset when back face culling is disabled", function () {
+    xit("renders glTF tileset when back face culling is disabled", function () {
```
ptrgags, line 2586: Similarly here, though the link is `https://github.com/CesiumGS/cesium/issues/10056`

Source/Scene/Cesium3DTileContentFactory.js
```diff
@@ -1,7 +1,6 @@
 import Batched3DModel3DTileContent from "./Batched3DModel3DTileContent.js";
```
ptrgags, line 1: We should mention that glTF contents now default to use ModelExperimental in `CHANGES.md`

@lilleyse which category of the change log should it show up under? Since Gltf3DTileContent is experimental code, it's not subject to the deprecation policy, but it does feel closer to a "breaking change" than an "addition"

ptrgags: Oh also when you update the changes file, add a new section for `1.91 - 2022-03-01`, since this PR will be merged after today's CesiumJS release.

ptrgags (commented):
@j9liu this code is looking solid and the sandcastles are working fine. 

Just a couple small things aside from that one test that's breaking (which I'll investigate more later today)

j9liu:
@ptrgags -- thanks for the feedback! Just implemented your changes, so all that's left is figuring out what's wrong with that last test.

ptrgags:
@j9liu I made a [simpler sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVDNTgIxEH6VpheWhLQYb7IQE4heTDxAPPVSugNWZ9tNO7sEjO/udosJKL20M/P9TY13kVhn4QCBzZmDA1tCtG0t3oZeobgZ6qV3pK2DoPh4ppxyZmCSRYhAPTVriGjAgWiCrS3ZDqLQVVVcyObrfrXJxOJLOcbagA9McbluwES50qTlFU4+I+1SAnAk97h5kmdf8RG9U1y573FOdU5x8r7e+OKMyqNfSqQjwvWul2brNM6xjEcfUrDhUYy22MJowqZiOs6eMz7h5aC3SPh0Hm3d+EBppUIISVA3qKlfYduaz97dxJjiJGgpL6llZTtmq/mND2cGdYz9ZNciru0JFF+Ussf/o6LXlXX71w4C6mOCvd8tXnJTCFHKvrzNJO9xq8Mf5R8) to see what's happening with the style. And it's not handling the alpha correctly.

I'm going to investigate more today, There's a chance I could have introduced a regression in #10018.

ptrgags:
@j9liu I confirmed that the model color stage is where the problem is. Though the fix shouldn't happen in the shader, it should be the render state that's updated. I looked back at `Model.js` and when the model color is invisible (alpha === 0) it updates the render state to skip rendering color and depth (because that would just be wasted GPU cycles)

In `Scene/ModelExperimental/ModelColorPipelineStage.js`, in the code that checks the alpha, you'll want to add these lines:

```
const renderStateOptions = renderResources.renderStateOptions;
  if (color.alpha === 0.0) {
    // When the model is invisible disable color and depth writes but still write into the stencil buffer
    renderStateOptions.colorMask = {
      red: false,
      green: false,
      blue: false,
      alpha: false,
    };
    renderStateOptions.depthMask = false;
  } 
```

However, `renderStateOptions` right now only exists in `PrimitiveRenderResources`, so there's a few other small changes needed here:

- [ ] `ModelRenderResources` - initialize `renderStateOptions` to an empty object
- [ ] `NodeRenderResources` - initialize `renderStateOptions` to clone from `modelRenderResources.renderStateOptions` (see other examples in the same file)
- [ ] `PrimitiveRenderResources` - instead of cloning, use `combine()` to combine the additional properties with `nodeRenderResources.renderStateOptions`

And update unit tests as necessary.

j9liu:
@ptrgags - Thanks for the guidance. I added your changes and the test passes now. Let me know if there are any concerns with the code I added.

Source/Scene/ModelExperimental/ModelColorPipelineStage.js
```diff
@@ -51,7 +51,17 @@ ModelColorPipelineStage.process = function (
 
   // Pass the model's color as a uniform. Set the pass type to translucent, if needed.
   const color = model.color;
-  if (color.alpha > 0.0 && color.alpha < 1.0) {
+  const renderStateOptions = renderResources.renderStateOptions;
+  if (color.alpha === 0.0) {
```
ptrgags, line 55: Looks good! please also add a unit test that it sets the render state as needed.

Specs/Scene/ModelExperimental/ModelRenderResourcesSpec.js
```diff
@@ -20,6 +20,7 @@ describe("Scene/ModelExperimental/ModelRenderResources", function () {
 
     expect(modelResources.model).toBe(mockModel);
     expect(modelResources.shaderBuilder).toBeDefined();
+    expect(modelResources.renderStateOptions).toBeDefined();
```
ptrgags, line 23: To be a little more specific:
```suggestion
    expect(modelResources.renderStateOptions).toEqual({});
```

Specs/Scene/ModelExperimental/NodeRenderResourcesSpec.js
```diff
@@ -49,16 +50,29 @@ describe("Scene/ModelExperimental/NodeRenderResources", function () {
     expect(nodeResources.runtimeNode).toBe(runtimeNode);
     expect(nodeResources.modelMatrix).toBe(runtimeNode.transform);
     expect(nodeResources.attributes).toEqual([]);
+    expect(nodeResources.renderStateOptions).toBeDefined();
```
ptrgags, line 53:
```suggestion
    expect(nodeResources.renderStateOptions).toEqual({});
```

ptrgags (commented):
@j9liu this is looking very close, just a few small tweaks before I merge it

j9liu:
@ptrgags - just pushed!

ptrgags:
@j9liu changes look good, I'm merging!

PR #10066, opened 2022-02-02
Allow ES6 template literals for new code?

ptrgags:
This PR is a little experiment related to #9718. I just enabled template literals in the `eslint` config so we can use them in `ModelExperimental`.

In this `ModelExperimental`/3D Tiles Next effort, we generate a lot of shader code. Template literals make this a lot more concise and readable compared to concatenating strings with `+`

This also allows us to replace the idiom for custom shader shader lines:

```
vertexShaderText: [
  "line 1",
  "line 2",
  "line 3",
].join("\n");
```

With the simpler multi-line template literal:
```
vertexShaderText: `
  line 1
  line 2
  line 3
  `
```

(technically speaking, this may add more whitespace on each line, but for user-defined GLSL shaders that doesn't matter much)

The point here is it's very easy to relax the eslint rules for new changes without having to convert all the old code at once.

### DIFF
--- a/Apps/Sandcastle/gallery/3D Tiles Next Photogrammetry Classification.html
+++ b/Apps/Sandcastle/gallery/3D Tiles Next Photogrammetry Classification.html
@@ -101,72 +101,72 @@
         const translucentWindowsShader = new Cesium.CustomShader({
           lightingModel: Cesium.LightingModel.UNLIT,
           isTranslucent: true,
-          fragmentShaderText: [
-            "const float WINDOW = 0.0;",
-            "const float SKYLIGHT = 4.0;",
-            "const float TOTAL_FEATURES = 12.0;",
-            "",
-            "void fragmentMain(FragmentInput fsInput, inout czm_modelMaterial material) {",
-            "  float featureId = fsInput.featureIds.featureId_0;",
-            "",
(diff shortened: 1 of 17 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/template-literals/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
@ptrgags and I discussed and decided that if we want to enable specific features, to keep things consistant we should make the change throughout the code base rather than just allowing in new code. Especially since its possible to mostly automate the change with `eslint --fix`.

ptrgags:
@ggetz and I talked about this offline. Since there are further ES6 related changes happening soon, if we want these template literals, it would be best to make the changes everywhere to avoid fragmenting the code base. Though I think that's possible here because:

* https://eslint.org/docs/rules/prefer-template -- this eslint option can fix the boilerplate of replacing `+` concatenation with template literals
* The only other changes I wanted to make in ModelExperimental were in a few sandcastles (removing the `[].join("\n")`, those could be done separately.

I'm going to close this PR for now and open a new one later once I get the `prefer-template` fix working

PR #10067, opened 2022-02-02
Use ES6 template literals instead of string concatenation

ptrgags:
For the 3D Tiles Next/`ModelExperimental` changes I've been working on, it would be very helpful to have ES6 string template literals for dynamic shader code generation. This was easy to do by enabling the eslint `prefer-template` rule and running `npm run eslint -- --fix`

In the discussion from #10066, we decided that if we're to make this change at all, it's best to do it to the entire codebase to avoid fragmentation, so I made this PR, using https://github.com/CesiumGS/cesium/pull/10026 as a guide.

There are 3 commits:

* [Update eslint config](https://github.com/CesiumGS/cesium/commit/16b3ad0b0d006d02c50f7c770402f88ce5af4577)
* [Small update to the Coding Guide](https://github.com/CesiumGS/cesium/commit/12aedb3ac997e6401444b209b8ff6075725d21ca)
* [use eslint to fix the code](https://github.com/CesiumGS/cesium/commit/2fa70153136da99eaf3ea81c0476d930b8532c7a)

@ggetz could you review?

### DIFF
--- a/Apps/CesiumViewer/CesiumViewer.js
+++ b/Apps/CesiumViewer/CesiumViewer.js
@@ -83,7 +83,7 @@ function main() {
   }
 
   const showLoadError = function (name, error) {
-    const title = "An error occurred while loading the file: " + name;
+    const title = `An error occurred while loading the file: ${name}`;
     const message =
       "An error occurred while loading the file, which may indicate that it is invalid.  A detailed error report is below:";
     viewer.cesiumWidget.showErrorPanel(title, message, error);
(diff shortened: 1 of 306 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/enable-template-literals/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @ptrgags! Looks good. Could you write up a post similar to [the instructions for merging let and const](https://community.cesium.com/t/cesiumjs-source-code-now-uses-let-and-const-syntax/17054)? It could even be in the same thread (and I can change the title) to keep ES6 syntax changes in the same place, but I'll leave it up to you.

ptrgags:
@ggetz ah good to hear we reached the same conclusion about putting the instructions in the same forum thread to keep everything in one place and in the correct order.

I [posted a comment with instructions](https://community.cesium.com/t/cesiumjs-source-code-now-uses-let-and-const-syntax/17054/2?u=ptrgags) for this change.

PR #10072, opened 2022-02-03
Workaround ambiant const enum errors as per TS docs

### LINKED ISSUE #10069: Ambiant const enums are being generated in TypeScript type definitions
Since version 1.90.0 of  cesium, some ambient enums TypeScript type definitions are being generated as const enums, which triggers an error in TypeScript projects that use the isolatedModules flag (typically all CRA projects).

The explantation of why an error is being triggered can be found here [Isolated Modules](https://www.typescriptlang.org/tsconfig#isolatedModules)

Typically, the following error is being thrown, which prevents the project from being transpiled:

```
TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.
                  labelEntity.label.style = new ConstantProperty(LabelStyle.FILL_AND_OUTLINE);
                                                               ^^^^^^^^^^
```

@ggetz It seems that this has been caused by PR #10026

ggetz:
Fixes #10069

This provides a workaround for the ambiant `const enum` errors that arise in TypeScript projects that use the `isolatedModules` flag.

[As per TypeScript documentation](https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls), this adds the `preserveConstEnums` to our `tsconfig.json`, then strips out the `const` from the `.d.ts` file [the same way as TypeScript themselves do it](https://github.com/microsoft/TypeScript/blob/1a981d1df1810c868a66b3828497f049a944951c/Gulpfile.js#L144).

I also added `isolatedModules` to the spec `tsconfig.json`, so this can be tested by running `build-ts`.

@mramato Could you please review?

### DIFF
--- a/Tools/jsdoc/tsconfig.json
+++ b/Tools/jsdoc/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
       "noEmit": true,
-      "types": []
+      "types": [],
+      "preserveConstEnums": true
   },
   "files": [
     "../../Source/Cesium.d.ts"
--- a/gulpfile.cjs
+++ b/gulpfile.cjs
@@ -1599,7 +1599,9 @@ function createTypeScriptDefinitions() {
       /= "WebGLConstants\.(.+)"/gm,
       // eslint-disable-next-line no-unused-vars
       (match, p1) => `= WebGLConstants.${p1}`
-    );
+    )
+    // Strip const enums which can cause errors - https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls
+    .replace(/^(\s*)(export )?const enum (\S+) {(\s*)$/gm, "$1$2enum $3 {$4");
 
   // Wrap the source to actually be inside of a declared cesium module
   // and add any workaround and private utility types.
--- a/Specs/TypeScript/tsconfig.json
+++ b/Specs/TypeScript/tsconfig.json
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "noEmit": true,
     "types": [],
-    "strict": true
+    "strict": true,
+    "isolatedModules": true
   },
   "include": [
     "index.ts",


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/const-enum-ts/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

gulpfile.cjs
```diff
@@ -1599,7 +1599,9 @@ function createTypeScriptDefinitions() {
       /= "WebGLConstants\.(.+)"/gm,
       // eslint-disable-next-line no-unused-vars
       (match, p1) => `= WebGLConstants.${p1}`
-    );
+    )
+    // Strip const enums which can cause errors - https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls
+    .replace(/^(\s*)(export )?const enum (\S+) {(\s*)$/gm, "$1$2enum $3 {$4");
```
mramato, line 1604: This looks good to me. We probably could have gotten away with an even simpler `const enum` `enum` call, but the good news is that if this ever breaks the build-ts step will fail so it's fine either way.

mramato:
Thanks @ggetz!

Tools/jsdoc/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
       "noEmit": true,
-      "types": []
+      "types": [],
+      "preserveConstEnums": true
```
mramato, line 5: @ggetz  I just realized, shouldn't this be false? And if we need to process manually anyway, does it matter?  Should it just be deleted?  Feel free to just put a follow-up commit in main.

ggetz: @mramato The documentation made it sounds like I should enabled the flag, but its not explicit.

> [preserveConstEnums](https://www.typescriptlang.org/tsconfig#preserveConstEnums) emits the same JavaScript for const enums as plain enums. You can then safely strip the const modifier from .d.ts files [in a build step](https://github.com/microsoft/TypeScript/blob/1a981d1df1810c868a66b3828497f049a944951c/Gulpfile.js#L144).

And the default value is different depending on whether `isolatedModules` is enabled or not.

> Default:
true if [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules),
false otherwise.

As long as we strip the const enums,`build-ts` passes either way, or if the flag is omitted entirely. So I assume we should omit it?

mramato: I honestly don't know and it probably doesn't matter, but if setting this to true vs false ends up producing the same exact output for us, then it's probably better to remove it to avoid confusion in the future

ggetz: Removed in https://github.com/CesiumGS/cesium/pull/10076

PR #10075, opened 2022-02-04
Issue 10057: Validate date parameters

### LINKED ISSUE #10057: `GregorianDate` constructor does not properly handle values
I found this issue while trying to convert a `GregorianDate` into a `JulianDate`. The [`GregorianDate` constructor](https://github.com/CesiumGS/cesium/blob/48c8abaca6257b568edc8c992ede5f2c487cdd38/Source/Core/GregorianDate.js) simply takes the parameters and puts them in `GregorianDate`'s respective fields, without actually validating them. For instance, the `hour` value should be in the range [0, 23], but nothing in the constructor actually enforces this.

Additionally, the [`GregorianDate` documentation](https://cesium.com/learn/cesiumjs/ref-doc/GregorianDate.html) says that all parameters can be optional. This seems strange -- what is the expected outcome of saying `var date = new GregorianDate()` ? It may make sense for some of these fields to be optional, like `second` and `millisecond`, but there is nothing in the constructor that provides a default value. This becomes problematic if I'm writing code along these lines:

```
var gregorianDate = new Cesium.GregorianDate(2022, 2, 1, 12, 0, 0);
var julianDate = new Cesium.JulianDate.fromGregorianDate(gregorianDate);
```

`JulianDate` expects the `GregorianDate` to have valid values. However, I didn't specify the `millisecond` value of the `GregorianDate`, so that is undefined. Thus, when `JulianDate` tries to perform this conversion, it results in a `NaN`.

SubhajitS:
Fixes #10057 

Following are the use cases covered:
- Validate inclusive range for year (0001 to 9999)
- Validate inclusive range for month (1 to 12)
- Validate inclusive range for month (1 to 28/29/30/31 depending on month & year)
- Validate inclusive range for hour (0 to 23)
- Validate inclusive range for minute (0 to 59)
- Validate inclusive range for second (1 to 59/60 depending on isLeapSecond)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug where updating `ModelExperimental`'s model matrix would not update its bounding sphere. [#10078](https://github.com/CesiumGS/cesium/pull/10078)
+- Fixed a bug where `GregorianDate` constructor would not validate the input parameters for valid date. Also default date is set as 0001-01-01 00:00:00:0000. [#10075](https://github.com/CesiumGS/cesium/pull/10075)
 
 ### 1.90 - 2022-02-01
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug where updating `ModelExperimental`'s model matrix would not update its bounding sphere. [#10078](https://github.com/CesiumGS/cesium/pull/10078)
-- Fixed a bug where `GregorianDate` constructor would not validate the input parameters for valid date. Also default date is set as 0001-01-01 00:00:00:0000. [#10075](https://github.com/CesiumGS/cesium/pull/10075)
+- Fixed a bug where `GregorianDate` constructor would not validate the input parameters for valid date. [#10075](https://github.com/CesiumGS/cesium/pull/10075)
 
 ### 1.90 - 2022-02-01
 
--- a/Source/Core/GregorianDate.js
+++ b/Source/Core/GregorianDate.js
@@ -1,3 +1,7 @@
+import defaultValue from "./defaultValue.js";
+import DeveloperError from "./DeveloperError.js";
+import isLeapYear from "./isLeapYear.js";
+
 /**
  * Represents a Gregorian date in a more precise format than the JavaScript Date object.
  * In addition to submillisecond precision, this object can also represent leap seconds.
@@ -25,21 +29,21 @@ function GregorianDate(
   millisecond,
   isLeapSecond
 ) {
-  const minYear = 1;
-  const minMonth = 1;
-  const minDay = 1;
-  const minHour = 0;
-  const minMinute = 0;
-  const minSecond = 0;
-  const minMillisecond = 0;
+  const minimumYear = 1;
+  const minimumMonth = 1;
+  const minimumDay = 1;
+  const minimumHour = 0;
+  const minimumMinute = 0;
+  const minimumSecond = 0;
+  const minimumMillisecond = 0;
 
-  if (!year) year = minYear;
-  if (!month) month = minMonth;
-  if (!day) day = minDay;
-  if (!hour) hour = minHour;
-  if (!minute) minute = minMinute;
-  if (!second) second = minSecond;
-  if (!millisecond) millisecond = minMillisecond;
+  year = defaultValue(year, minimumYear);
+  month = defaultValue(month, minimumMonth);
+  day = defaultValue(day, minimumDay);
+  hour = defaultValue(hour, minimumHour);
+  minute = defaultValue(minute, minimumMinute);
+  second = defaultValue(second, minimumSecond);
+  millisecond = defaultValue(millisecond, minimumMillisecond);
   validateRange();
   validateDate();
 
@@ -85,13 +89,13 @@ function GregorianDate(
   this.isLeapSecond = isLeapSecond;
 
   function validateRange() {
-    const maxYear = 9999;
-    const maxMonth = 12;
-    const maxDay = 31;
-    const maxHour = 23;
-    const maxMinute = 59;
-    const maxSecond = 59;
-    const excludedMaxMilisecond = 1000;
+    const maximumYear = 9999;
+    const maximumMonth = 12;
+    const maximumDay = 31;
+    const maximumHour = 23;
+    const maximumMinute = 59;
+    const maximumSecond = 59;
+    const excludedMaximumMilisecond = 1000;
 
     if (
       isNaN(year) ||
@@ -102,44 +106,48 @@ function GregorianDate(
       isNaN(second) ||
       isNaN(millisecond)
     )
-      throw "Invalid value passed in parameter";
+      throw new DeveloperError("Invalid value passed in parameter");
 
-    if (year < minYear || year > maxYear)
-      throw "Year parameter represent an invalid date";
-    if (month < minMonth || month > maxMonth)
-      throw "Month parameter represent an invalid date";
-    if (day < minDay || day > maxDay)
-      throw "Day parameter represent an invalid date";
-    if (hour < minHour || hour > maxHour)
-      throw "Hour parameter represent an invalid date";
-    if (minute < minMinute || minute > maxMinute)
-      throw "Combination of Minute and IsLeapSecond parameters represent an invalid date";
+    if (year < minimumYear || year > maximumYear)
+      throw new DeveloperError("Year parameter represent an invalid date");
+    if (month < minimumMonth || month > maximumMonth)
+      throw new DeveloperError("Month parameter represent an invalid date");
+    if (day < minimumDay || day > maximumDay)
+      throw new DeveloperError("Day parameter represent an invalid date");
+    if (hour < minimumHour || hour > maximumHour)
+      throw new DeveloperError("Hour parameter represent an invalid date");
+    if (minute < minimumMinute || minute > maximumMinute)
+      throw new DeveloperError(
+        "Combination of Minute and IsLeapSecond parameters represent an invalid date"
+      );
     if (
-      second < minSecond ||
-      (second > maxSecond && !isLeapSecond) ||
-      (second > maxSecond + 1 && isLeapSecond)
+      second < minimumSecond ||
+      (second > maximumSecond && !isLeapSecond) ||
+      (second > maximumSecond + 1 && isLeapSecond)
     )
-      throw "Second parameter represent an invalid date";
-    if (millisecond < minMillisecond || millisecond >= excludedMaxMilisecond)
-      throw "Millisecond parameter represent an invalid date";
+      throw new DeveloperError("Second parameter represent an invalid date");
+    if (
+      millisecond < minimumMillisecond ||
+      millisecond >= excludedMaximumMilisecond
+    )
+      throw new DeveloperError(
+        "Millisecond parameter represent an invalid date"
+      );
   }
 
   // Javascript date object supports only dates greater than 1901. Thus validating with custom logic
   function validateDate() {
-    const minNumberOfDaysInMonth = 28;
-    const monthsWith31Days = [1, 3, 5, 7, 8, 10, 12];
+    const minimumDaysInMonth = 28;
+    const daysInYear = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
 
     if (
       month === 2 &&
-      ((year % 4 === 0 && day > minNumberOfDaysInMonth + 1) ||
-        (year % 4 !== 0 && day > minNumberOfDaysInMonth))
+      ((isLeapYear(year) && day > minimumDaysInMonth + 1) ||
+        (!isLeapYear(year) && day > minimumDaysInMonth))
     )
-      throw "Year, Month and Day represents invalid date";
-    else if (
-      monthsWith31Days.indexOf(month) === -1 &&
-      day > minNumberOfDaysInMonth + 2
-    )
-      throw "Month and Day represents invalid date";
+      throw new DeveloperError("Year, Month and Day represents invalid date");
+    else if (month !== 2 && day > daysInYear[month - 1])
+      throw new DeveloperError("Month and Day represents invalid date");
   }
 }
 export default GregorianDate;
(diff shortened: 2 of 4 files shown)


cesium-concierge:
Thank you so much for the pull request @SubhajitS! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/SubhajitS/cesium/blob/issue-10057/CONTRIBUTORS.md) file!
* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/issue-10057/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

SubhajitS:
- [x]  CLA is submitted
- [x] Added name in CONTRIBUTORS.md

ggetz (commented):
Thanks @SubhajitS for your first PR! Welcome!

I recommend you read through the [Cesium Coding Guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md), especially the section on default parameters and throwing exceptions. I left a few comments here where that applies. Keeping our coding style uniform helps us keep everything consistent between contributors and make things less prone to bugs.

I'd also ask that you write [unit tests](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/TestingGuide) to confirm your validation code.

Source/Core/GregorianDate.js
```diff
@@ -69,22 +70,57 @@ function GregorianDate(
...
       throw "Second parameter represent an invalid date";
-    if (millisecond < 0 || millisecond >= 1000)
+    if (millisecond < minMilisecond || millisecond >= excludedMaxMilisecond)
       throw "Millisecond parameter represent an invalid date";
   }
+
+  // Javascript date object supports only dates greater than 1901. Thus validating with custom logic
+  function validateDate() {
+    const minNumberOfDaysInMonth = 28;
+    const monthsWith31Days = [1, 3, 5, 7, 8, 10, 12];
```
j9liu, line 130: This may be nitpicky, but can we handle the amount of days in each month [the same way that `JulianDate.js` handles them](https://github.com/CesiumGS/cesium/blob/6bbf84b7d0bc734ff3c7678830fb19899675950d/Source/Core/JulianDate.js#L12)? You would still have to check for the Feburary leap year case, but I think this should simplify the rest of the code.

j9liu (commented):
@ggetz covered all of the important parts. I also added a few suggestions that would make this code more consistent with Cesium's. Can you also log this change in `CHANGES.md`?

SubhajitS:
@ggetz @j9liu  thanks a lot for your feedback. I will incorporate these and let you folks know.

Source/Core/GregorianDate.js
```diff
@@ -25,6 +25,24 @@ function GregorianDate(
...
+  const minMillisecond = 0;
+
+  if (!year) year = minYear;
+  if (!month) month = minMonth;
+  if (!day) day = minDay;
+  if (!hour) hour = minHour;
+  if (!minute) minute = minMinute;
+  if (!second) second = minSecond;
+  if (!millisecond) millisecond = minMillisecond;
+  validateRange();
```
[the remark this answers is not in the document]
SubhajitS: Hi @ggetz I have doubt about the pragmas. If we remove these validations during release then how do we handle the runtime error that may occur?

Source/Core/GregorianDate.js
```diff
@@ -65,5 +83,63 @@ function GregorianDate(
...
+    if (hour < minHour || hour > maxHour)
+      throw "Hour parameter represent an invalid date";
+    if (minute < minMinute || minute > maxMinute)
+      throw "Combination of Minute and IsLeapSecond parameters represent an invalid date";
+    if (
+      second < minSecond ||
+      (second > maxSecond && !isLeapSecond) ||
+      (second > maxSecond + 1 && isLeapSecond)
+    )
+      throw "Second parameter represent an invalid date";
```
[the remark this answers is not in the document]
SubhajitS: Removed throws from all but one places. Instead of custom range check used the Check.js.

Source/Core/GregorianDate.js
```diff
@@ -25,6 +25,24 @@ function GregorianDate(
...
 ) {
+  const minYear = 1;
+  const minMonth = 1;
+  const minDay = 1;
+  const minHour = 0;
+  const minMinute = 0;
+  const minSecond = 0;
+  const minMillisecond = 0;
+
+  if (!year) year = minYear;
```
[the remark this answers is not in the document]
SubhajitS: Used defaultValue function to check and assign default values.

Source/Core/GregorianDate.js
```diff
@@ -65,5 +83,63 @@ function GregorianDate(
...
+    const excludedMaxMilisecond = 1000;
+
+    if (
+      isNaN(year) ||
+      isNaN(month) ||
+      isNaN(day) ||
+      isNaN(hour) ||
+      isNaN(minute) ||
+      isNaN(second) ||
+      isNaN(millisecond)
```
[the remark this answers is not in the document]
SubhajitS: This check is replaced with Check.typeOf

Source/Core/GregorianDate.js
```diff
@@ -69,22 +70,57 @@ function GregorianDate(
   this.isLeapSecond = isLeapSecond;
 
   function validateRange() {
-    if (year < 1 || year > 9999)
+    const maxYear = 9999;
```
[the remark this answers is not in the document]
SubhajitS: Changed according to the naming conventions. Thanks for pointing out.

Source/Core/GregorianDate.js
SubhajitS: Refactored this method to make it simpler and more readable.

SubhajitS:
Hi @ggetz and @j9liu I have incorporated the comments. Please review once. 

Also please let me know whether we need to change the behaviour of the parameterless constructor as suggested by @javagl. As I am missing most of the context in terms of usage of the function thus I do not have much input in this discussion.

Source/Core/GregorianDate.js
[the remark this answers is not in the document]
javagl: An aside:  Code like
```
if (something) { ... }
```
hardly ever makes sense (unless, maybe, you already know that `something` is a boolean value, and it is *not* "undefined"). 

Cesium offers functions like `Cesium.defined` and `Cesium.defaultValue`. But these are not ~"some specific coding practice in Cesium". Functions like these are **required** - in every environment - if the code should remain readable and still make sense. 

Consider the following snippet:
```
var yearA = 123;
var yearB = undefined;
var yearC = null;
var yearD = 0;
var yearE = "";

if (yearA) {
  console.log('True for ' + yearA);
} else{
  console.log('False for ' + yearA);
}

if (yearB) {
  console.log('True for ' + yearB);
} else{
  console.log('False for ' + yearB);
}

if (yearC) {
  console.log('True for ' + yearC);
} else{
  console.log('False for ' + yearC);
}

if (yearD) {
  console.log('True for ' + yearD);
} else{
  console.log('False for ' + yearD);
}

if (yearE) {
  console.log('True for ' + yearE);
} else{
  console.log('False for ' + yearE);
}
```

The output is 
```
True for 123
False for undefined
False for null
False for 0
False for 
```

Note that even for the value `0` and for the empty string `""`, the `if`-statement will still behave "as if the value was `false` (or `undefined`)". Not taking this into account will inevietably lead to bugs.

j9liu:
@SubhajitS - the `GregorianDate` tests weren't actually being run because they weren't added to `SpecList.js`. Can you import them there? (it should be between the `GoogleEarthEnterpriseTerrainProviderSpec.js` and `GroundPolylineGeometrySpec.js` imports to keep to alphabetical order)

Specs/Core/GregorianDateSpec.js
```diff
@@ -0,0 +1,300 @@
+import { GregorianDate } from "../../Source/Cesium.js";
+
+describe("Core/GregorianDate", function () {
+  describe("Positive scenarios", function () {
```
j9liu, line 4: A lot of our tests don't use nested `describes` unless they get very long, but it's fine to leave this in. I would just be specific -- instead of "positive scenarios", say "valid parameters." Instead of "negative scenarios," "invalid parameters."

Specs/Core/GregorianDateSpec.js
```diff
@@ -0,0 +1,300 @@
...
+    it("Constructs any valid date", function () {
+      const validDate = new GregorianDate(2022, 2, 4, 23, 54, 0, 999.9, false);
+      expect(validDate.year).toEqual(2022);
+      expect(validDate.month).toEqual(2);
+      expect(validDate.day).toEqual(4);
+      expect(validDate.hour).toEqual(23);
+      expect(validDate.minute).toEqual(54);
+      expect(validDate.second).toEqual(0);
+      expect(validDate.millisecond).toEqual(999.9);
+      expect(validDate.isLeapSecond).toBeFalsy();
```
j9liu, line 14: We want to ensure that `isLeapSecond` is a `boolean` like the documentation is described. However, `toBeFalsy()` doesn't enforce the boolean type: it will coerce non-boolean values into a `false` value (e.g. `null` or 0). Use `toBe(false)` or `toBe(true)` instead.

You can read more about it [here](https://ralzohairi.medium.com/the-difference-between-tobe-true-tobetrue-and-tobetruthy-in-the-jasmine-framework-2dd652ff13a2) if you want more detail about the differences.

Specs/Core/GregorianDateSpec.js
```diff
@@ -0,0 +1,300 @@
...
+      expect(minimumDate.second).toEqual(0);
+      expect(minimumDate.millisecond).toEqual(0);
+      expect(minimumDate.isLeapSecond).toBeFalsy();
+    });
+
+    it("Constructs valid dates for edge cases of days", function () {
+      expect(function () {
+        return new GregorianDate(2022, 1, 31);
+      }).not.toThrowDeveloperError();
+      expect(function () {
```
j9liu, line 45: Even though you test 2/28/2022 in the other tests, I think you should include it here for consistency.

Specs/Core/GregorianDateSpec.js
```diff
@@ -0,0 +1,300 @@
...
+      }).toThrowDeveloperError();
+      expect(function () {
+        return new GregorianDate(2022, 0, 4, 23, 54, 0, 999.9, false);
+      }).toThrowDeveloperError();
+      expect(function () {
+        return new GregorianDate(2022, 13, 4, 23, 54, 0, 999.9, false);
+      }).toThrowDeveloperError();
+    });
+
+    it("Should throw error if invalid day is passed", function () {
```
j9liu, line 186: For easier readability can we split this up into three tests?
- "Should throw error if invalid day is passed," which handles invalid values (-10, 0, 32)
- "Should throw error if day is out of range for month," which handles cases where months with 30 days are given an input of 31, and where February is given an input of 30
- "Should throw error if leap day is given for non-leap year", which handles the special February 29 case

Organizing them this way will help us remember the intentions of each test for future reference

j9liu (commented):
@SubhajitS - looks better now, thanks for the changes! I left some remarks for the specs.

Also remember to log this in `CHANGES.md`. You can follow the template for the previous releases; just add it under a new "Fixes" section under 1.91. Let me know when you're ready for another look!

j9liu:
> @SubhajitS - the `GregorianDate` tests weren't actually being run because they weren't added to `SpecList.js`. Can you import them there? (it should be between the `GoogleEarthEnterpriseTerrainProviderSpec.js` and `GroundPolylineGeometrySpec.js` imports to keep to alphabetical order)

Hold on, you shouldn't actually have to do this because `SpecList.js` isn't included in the repo. I'm going to ask some peers and see why it wasn't being added to the list of tests.

SubhajitS:
Hi @j9liu, I have incorporated these changes. Also marked the PR as ready. Please review the same.

CHANGES.md
```diff
@@ -13,6 +13,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug where updating `ModelExperimental`'s model matrix would not update its bounding sphere. [#10078](https://github.com/CesiumGS/cesium/pull/10078)
+- Fixed a bug where `GregorianDate` constructor would not validate the input parameters for valid date. Also default date is set as 0001-01-01 00:00:00:0000. [#10075](https://github.com/CesiumGS/cesium/pull/10075)
```
j9liu, line 16: I would delete the "Also default date is set as 0001-01-01 00:00:00:0000" part of the comment, it's not necessary here.

j9liu (commented):
Things are looking good! Just one small change in `CHANGES.md`.

@ggetz do you want to take one final look before we merge this PR?

Source/Core/GregorianDate.js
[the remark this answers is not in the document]
ggetz: The extra validation step can effect performance, especially if its called every frame. In Cesium, we typically opt to put validation behind the debug pragmas rather than always error check. Errors that occur at runtime can be thrown using `RuntimeError` rather than the `DeveloperError` which is thrown as part of `Check`, but we do not usually consider parameter validation a runtime error.

Source/Core/GregorianDate.js
```diff
@@ -65,5 +91,60 @@ function GregorianDate(
...
+      millisecond,
+      excludedMaximumMilisecond
+    );
+  }
+
+  // Javascript date object supports only dates greater than 1901. Thus validating with custom logic
+  function validateDate() {
+    const daysInYear = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+
+    if (month === 2 && isLeapYear(year)) daysInYear[month - 1] += 1;
```
ggetz, line 144:
```suggestion
    if (month === 2 && isLeapYear(year)) { 
        daysInYear[month - 1] += 1; 
    }
```

The cesium style guide says to always use curly brackets.

Source/Core/GregorianDate.js
```diff
@@ -65,5 +91,60 @@ function GregorianDate(
...
+    Check.typeOf.number.lessThan(
+      "Millisecond",
+      millisecond,
+      excludedMaximumMilisecond
+    );
+  }
+
+  // Javascript date object supports only dates greater than 1901. Thus validating with custom logic
+  function validateDate() {
+    const daysInYear = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
```
ggetz, line 142: This reallocates a new array every time we create a new date. Can this be scoped to the module instead?

ggetz:
Hi @SubhajitS. Since this PR has been inactive for a bit, I'm going to close this to keep things tidy. This was really close to finished, so if you get the chance to update this, please re-open! Thanks!

PR #9997, opened 2022-01-07
Fixes ModelExperimental modelMatrix update bug

sanjeetsuhag:
This PR fixes a bug introduced in https://github.com/CesiumGS/cesium/pull/9989 that did not allow in-place modification of a `ModelExperimental`'s `modelMatrix`. The associated test for this functionality has also been modified to test for changing of the model matrices by applying a transformation in-place.

[...]

PR #10078, opened 2022-02-04
Make `ModelExperimental`'s bounding sphere update with model matrix

### LINKED ISSUE #10074: ModelExperimental bounding sphere not updated when model matrix changes
`ModelExperimental`'s bounding sphere is not getting updated when the model matrix changes.

[Sandcastle repro](https://sandcastle.cesium.com/#c=jVVdb9s2FP0rhJ/kwaElW7Edzw3WpG3aYsGC2uiATnugpWuLCEUaJOUkG/Lfd0VKshI7WA1DEsnD+3XOJVMljSV7Dg+gyTsi4YFcg+FlQb+7uSDppW58raRlXIJOev1fE5lIv4emQqX3NC21BmlXvAC0Ulv4WgrO5AdmgW60Kr4YNZuEUZBIQpLeKBxFZ1F0Fl6sRuF8NJ2PpzSMJuEkjqfn43g2mcUX4+mPpJdI7284JLcqA2ESmbqgCzdCd39VFv+tHoRYeLRzNO9DIFdMCKVk0hv4ZdwIAlKEbEqZWq4kCfrNXkL8onMTNHNVsJQO8b9kxU4A5sOG3vfQe6mdvBzRrVi3bqtfdN58V/lU72e3Wj3/dhn6tPLdKyI+A8u43N5xm+bflPCR1Wu3zObUqm+IYNIE0SzsO6Ohf/raebsb/gjZJ80KWGnEbpQuDlS1U4Yin0x4nPrU7rkBpJ5ZpWv6pNI2r/NLeg9g7IEpATU5btQWulvcUosByYFvc1uX3we5U4Y7cBvZNdMWv5gcOxF9gK0GMDU7Z9FoTMNpHE+ii7rUcUzD83A8DSf1hPdSffvoSK12alLMie40L9DlHgzVUKg9vMcKe4ZcDhjJW3iWZXUcDRvVho+PO0AMdgMTLuQbYTdBK7Etjuakyr+ZyWBdbpe5erhSpayY/q5EWcCcWF1Ci3LBIN2aP85PsJa/EkmXu46Um/p2dIl664xqyx+F4DujeEb/vFnO4g7ghIxaXdew53632i5wrC3Lnu6wGtwAtTnI4NCADtF2YXOuoAfNUI3q/r1tMziK/60+YXLbzft0u8Tn/cH/Yc6iFyCfzbpmarnLQQPViC0N+YUcWtx/eB09V6/nqhpLJrOUGSugEs9KKbFm+hZk6WtgfM1Ooq5Ka5XE4/gWNYqN1x5gwYsGshUpgr3sIS+b2KlxdQAEneId2iyIwgGp/n0ffWVXYShCbYOO9U6L0I42j50WpbB8J56CI+ygG+3g2JYr28+U5E5zaUlDCjGOlVYrbxWrSeokpdjrFnQbQW/QWxj7JOCyYfg3XuzwDKw6OcC7wQLeDXjTmeG6TO/B0tSY5pxfDLtbFxnfE569O3GxklQwY3BlUwqx5P8gz5eLIeKPtgrltP7HHrRgTxUsjy5/95OU0sUQh6d3Wl+7V5b/Aw)

* Click "Print bounding sphere position"
* Click "Move"
* Click "Print bounding sphere position" again

The console prints identical positions.

This only affects the main bounding sphere; the draw commands' bounding spheres seem to be updated properly.

Second problem is that `buildDrawCommands` continually pushes new bounding spheres to `this._boundingSpheres` instead of resetting the array, which will cause the final bounding sphere to be incorrect. I don't think the member variable is necessary anymore. The `boundingSpheres` array can be local to the function.

j9liu:
Fixes #10074.

This PR fixes a bug where `ModelExperimental`'s bounding sphere wouldn't be affected by changes in the model matrix. I added a unit test for it and observed that it worked in @lilleyse's [sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=jVVdb9s2FP0rhJ/kwaElW7Edzw3WpG3aYsGC2uiATnugpWuLCEUaJOUkG/Lfd0VKshI7WA1DEsnD+3XOJVMljSV7Dg+gyTsi4YFcg+FlQb+7uSDppW58raRlXIJOev1fE5lIv4emQqX3NC21BmlXvAC0Ulv4WgrO5AdmgW60Kr4YNZuEUZBIQpLeKBxFZ1F0Fl6sRuF8NJ2PpzSMJuEkjqfn43g2mcUX4+mPpJdI7284JLcqA2ESmbqgCzdCd39VFv+tHoRYeLRzNO9DIFdMCKVk0hv4ZdwIAlKEbEqZWq4kCfrNXkL8onMTNHNVsJQO8b9kxU4A5sOG3vfQe6mdvBzRrVi3bqtfdN58V/lU72e3Wj3/dhn6tPLdKyI+A8u43N5xm+bflPCR1Wu3zObUqm+IYNIE0SzsO6Ohf/raebsb/gjZJ80KWGnEbpQuDlS1U4Yin0x4nPrU7rkBpJ5ZpWv6pNI2r/NLeg9g7IEpATU5btQWulvcUosByYFvc1uX3we5U4Y7cBvZNdMWv5gcOxF9gK0GMDU7Z9FoTMNpHE+ii7rUcUzD83A8DSf1hPdSffvoSK12alLMie40L9DlHgzVUKg9vMcKe4ZcDhjJW3iWZXUcDRvVho+PO0AMdgMTLuQbYTdBK7Etjuakyr+ZyWBdbpe5erhSpayY/q5EWcCcWF1Ci3LBIN2aP85PsJa/EkmXu46Um/p2dIl664xqyx+F4DujeEb/vFnO4g7ghIxaXdew53632i5wrC3Lnu6wGtwAtTnI4NCADtF2YXOuoAfNUI3q/r1tMziK/60+YXLbzft0u8Tn/cH/Yc6iFyCfzbpmarnLQQPViC0N+YUcWtx/eB09V6/nqhpLJrOUGSugEs9KKbFm+hZk6WtgfM1Ooq5Ka5XE4/gWNYqN1x5gwYsGshUpgr3sIS+b2KlxdQAEneId2iyIwgGp/n0ffWVXYShCbYOO9U6L0I42j50WpbB8J56CI+ygG+3g2JYr28+U5E5zaUlDCjGOlVYrbxWrSeokpdjrFnQbQW/QWxj7JOCyYfg3XuzwDKw6OcC7wQLeDXjTmeG6TO/B0tSY5pxfDLtbFxnfE569O3GxklQwY3BlUwqx5P8gz5eLIeKPtgrltP7HHrRgTxUsjy5/95OU0sUQh6d3Wl+7V5b/Aw).

### DIFF
--- a/Source/Scene/ModelExperimental/ModelExperimental.js
+++ b/Source/Scene/ModelExperimental/ModelExperimental.js
@@ -134,6 +134,7 @@ export default function ModelExperimental(options) {
 
   // Keeps track of resources that need to be destroyed when the Model is destroyed.
   this._resources = [];
+  this._boundingSphere = undefined;
 
   const pointCloudShading = new PointCloudShading(options.pointCloudShading);
   this._attenuation = pointCloudShading.attenuation;
@@ -533,10 +534,7 @@ Object.defineProperties(ModelExperimental.prototype, {
       }
       //>>includeEnd('debug');
 
-      return BoundingSphere.transform(
-        this._sceneGraph.boundingSphere,
-        this.modelMatrix
-      );
+      return this._boundingSphere;
     },
   },
 
@@ -745,6 +743,12 @@ ModelExperimental.prototype.update = function (frameState) {
 
   if (!Matrix4.equals(this.modelMatrix, this._modelMatrix)) {
     this._sceneGraph.updateModelMatrix(this);
+    this._modelMatrix = Matrix4.clone(this.modelMatrix);
+    BoundingSphere.transform(
+      this._sceneGraph.boundingSphere,
+      this.modelMatrix,
+      this._boundingSphere
+    );
   }
 
   if (this._backFaceCullingDirty) {
--- a/Source/Scene/ModelExperimental/ModelExperimental.js
+++ b/Source/Scene/ModelExperimental/ModelExperimental.js
@@ -134,7 +134,7 @@ export default function ModelExperimental(options) {
 
   // Keeps track of resources that need to be destroyed when the Model is destroyed.
   this._resources = [];
-  this._boundingSphere = undefined;
+  this._boundingSphere = new BoundingSphere();
 
   const pointCloudShading = new PointCloudShading(options.pointCloudShading);
   this._attenuation = pointCloudShading.attenuation;
@@ -741,6 +741,8 @@ ModelExperimental.prototype.update = function (frameState) {
     this._debugShowBoundingVolumeDirty = false;
   }
 
+  // This is done without a dirty flag so that the model matrix can be update in-place
+  // without needing to use a setter.
   if (!Matrix4.equals(this.modelMatrix, this._modelMatrix)) {
     this._sceneGraph.updateModelMatrix(this);
     this._modelMatrix = Matrix4.clone(this.modelMatrix);
--- a/Source/Scene/ModelExperimental/ModelExperimental.js
+++ b/Source/Scene/ModelExperimental/ModelExperimental.js
@@ -1,3 +1,4 @@
+import BoundingSphere from "../../Core/BoundingSphere.js";
 import Check from "../../Core/Check.js";
 import ColorBlendMode from "../ColorBlendMode.js";
 import defined from "../../Core/defined.js";
@@ -133,8 +134,6 @@ export default function ModelExperimental(options) {
 
   // Keeps track of resources that need to be destroyed when the Model is destroyed.
   this._resources = [];
-
-  this._boundingSphere = undefined;
 
   const pointCloudShading = new PointCloudShading(options.pointCloudShading);
   this._attenuation = pointCloudShading.attenuation;
@@ -534,7 +533,10 @@ Object.defineProperties(ModelExperimental.prototype, {
       }
       //>>includeEnd('debug');
 
-      return this._sceneGraph.boundingSphere;
+      return BoundingSphere.transform(
+        this._sceneGraph.boundingSphere,
+        this.modelMatrix
+      );
     },
   },
 
--- a/Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
+++ b/Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
@@ -105,16 +105,6 @@ export default function ModelExperimentalSceneGraph(options) {
   this._drawCommands = [];
 
   /**
-   * The array of bounding spheres of all the primitives in the scene graph.
-   *
-   * @type {BoundingSphere[]}
-   * @readonly
-   *
-   * @private
-   */
-  this._boundingSpheres = [];
-
-  /**
    * Pipeline stages to apply to this model. This
    * is an array of classes, each with a static method called
    * <code>process()</code>
@@ -293,6 +283,7 @@ ModelExperimentalSceneGraph.prototype.buildDrawCommands = function (
     modelPipelineStage.process(modelRenderResources, model, frameState);
   }
 
+  const boundingSpheres = [];
   for (i = 0; i < this._runtimeNodes.length; i++) {
     const runtimeNode = this._runtimeNodes[i];
     runtimeNode.configurePipeline();
@@ -337,7 +328,8 @@ ModelExperimentalSceneGraph.prototype.buildDrawCommands = function (
       runtimePrimitive.boundingSphere = BoundingSphere.clone(
         primitiveRenderResources.boundingSphere
       );
-      this._boundingSpheres.push(primitiveRenderResources.boundingSphere);
+
+      boundingSpheres.push(primitiveRenderResources.boundingSphere);
 
       const drawCommands = buildDrawCommands(
         primitiveRenderResources,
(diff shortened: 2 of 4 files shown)


Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -534,7 +533,10 @@ Object.defineProperties(ModelExperimental.prototype, {
       }
       //>>includeEnd('debug');
 
-      return this._sceneGraph.boundingSphere;
+      return BoundingSphere.transform(
```
ptrgags, line 536: Hm, I don't think this is the right place to apply the `transform()`, as this is allocating every time you call the `model.boundingSphere` getter. 

It would probably be better to have `model._boundingSphere` that represents the final bounding sphere, and whenever the model matrix is updated, that bounding sphere should be updated in place (using the `result` parameter).

ptrgags (commented):
@j9liu this PR does work as advertised. Though the bounding volume update code is spread out in several places, so we might want to review it holistically to see what if anything needs to be updated/simplified The model matrix and bounding sphere get updated in a few places:

* in `buildDrawCommands.js` I see that a bounding volume is computed from the render resources' copy, in-place. Does that make sense to do still? Note that sometimes `buildDrawCommands()` gets called a second time if draw commands need to be re-built (such as making a big change like swapping out a custom shader or style). Though since the whole pipeline gets re-run, this might be okay.
* I see that the bounding sphere gets updated in `ModelMatrixUpdateStage` -- we might need something similar for the final model bounding volume (like in the review comment below)

We can discuss this more on a call if needed since this is a little more nuanced than it first appears.

Source/Scene/ModelExperimental/ModelExperimental.js
j9liu: Yeah that makes sense, I think I was trying to incorporate Sean's observation that `model._boundingSphere` wasn't being used and could be deleted. But it makes more sense to actually use that to hold the final bounding sphere, and whenever model matrix is updated, to update that in turn.

j9liu:
@ptrgags - just wanted to clarify your `buildDrawCommands.js` comment. I see that the `boundingSphere` is transformed by the model matrix in place. Is the concern that if `buildDrawCommands` needs to get run again, the `boundingSphere` will get transformed by the model matrix twice?

ptrgags:
@j9liu yes, that's what I meant, though now that I think about it, even when `rebuildDrawCommands()` gets run, the entire pipeline gets run, so the bounding sphere I believe would be created from scratch so I think it's okay.

j9liu:
@ptrgags - got it. Is there anything else I should look at to wrap-up this PR?

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -743,6 +743,12 @@ ModelExperimental.prototype.update = function (frameState) {
 
   if (!Matrix4.equals(this.modelMatrix, this._modelMatrix)) {
     this._sceneGraph.updateModelMatrix(this);
+    this._modelMatrix = Matrix4.clone(this.modelMatrix);
```
ptrgags, line 746: wait a sec... not sure why we had both `this.modelMatrix` and `this._modelMatrix`... is the underscore one used for anything except for checking if the matrix is dirty? I think this should be turned into a proper getter/setter and just use a dirty flag for consistency.

Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
```diff
@@ -347,8 +339,11 @@ ModelExperimentalSceneGraph.prototype.buildDrawCommands = function (
       runtimePrimitive.drawCommands = drawCommands;
     }
   }
-  this._boundingSphere = BoundingSphere.fromBoundingSpheres(
-    this._boundingSpheres
+
+  this._boundingSphere = BoundingSphere.fromBoundingSpheres(boundingSpheres);
+  this._model._boundingSphere = BoundingSphere.transform(
+    this._boundingSphere,
+    this._model.modelMatrix
```
ptrgags, line 346: here you can pass in `this._model._boundingSphere` as the third parameter so it gets updated in place if it was already created instead of allocating a new one.

ptrgags (commented):
@j9liu closer, though I noticed that the original code's `_modelMatrix` vs `modelMatrix` is a bit confusing, must have been missed in a past review. I left some notes of how to neaten that up.

Source/Scene/ModelExperimental/ModelExperimental.js
lilleyse: The current design is intentional so that the model matrix can be updated in place without needing to use the setter. See https://github.com/CesiumGS/cesium/pull/9997.

lilleyse: Maybe a comment is needed since it's the only property that deviates from the dirty flag approach

ptrgags:
@j9liu looks good now, I'll merge once CI passes!

PR #10086, opened 2022-02-09
Handle JSON-based subtrees in implicit tiling

### LINKED ISSUE #10084: Add JSON-based subtrees for implicit tiling
An upcoming spec change is to allow a JSON-based subtree format for implicit tiling in addition to the current binary format.

This will involve:

- [ ] in `Cesium3DTileContentType`, add an enum constant for the JSON format, e.g. `IMPLICIT_SUBTREE_JSON: "subtreeJson"`. (Note: leave the existing `IMPLICIT_SUBTREE` as-is since it must store the magic number of `subt`).
- [ ] in `preprocess3DTileContent.js` (which detects which type of content a tile is), towards the end where it handles JSON files, add an if clause to check for `json.tileAvailability` (really anyJSON field can distinguish a subtree from a glTF JSON or an external tileset.json would work)
- [ ] in `Cesium3DTileContentFactory.js`, add a new clause for `subtreeJson` (it must match the value of `IMPLICIT_SUBTREE_JSON`). the callback should look like the `gltf` one where the last parameter is `json`. This should return an `Implicit3DTileContent` created from JSON
- [ ] in `Implicit3DTileContent` and `ImplicitSubtree`, we now need to be able to construct this either from JSON or from binary. (`ModelExperimental3DTileContent` and `ModelExperimental` might be useful for inspiration, since those also have to handle JSON and binary)

CC @j9liu

j9liu:
Resolves #10084.

This PR adds the ability to load subtrees through JSONs in addition to binary subtree files. I updated the specs for the affected classes, but they are relatively simple because we don't have sample data that uses JSON subtrees. Let me know if I should add any more.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/Implicit3DTileContent.js
```diff
@@ -164,22 +166,23 @@ Object.defineProperties(Implicit3DTileContent.prototype, {
  * up a promise chain to expand the immediate subtree.
  *
  * @param {Implicit3DTileContent} content The implicit content
+ * @param {Object} json The JSON containing the subtree. If parsing a binary subtree file, this will be undefined.
```
ptrgags, line 169: also update the documentation in this file to mark `json` and `arrayBuffer` as mutually exclusive. Though the check only need to go in the constructor.

ptrgags (commented):
@j9liu looks pretty solid, my feedback is mostly about code style and documentation tweaks

j9liu:
@ptrgags - thank you for the review! I just pushed some changes, let me know if there's anything else that I need to correct.

PR #10083, opened 2022-02-08
Rename `maximumLevel` to `availableLevels`

j9liu:
This PR renames the `maximumLevel` variable in `ImplicitTileset.js` to `availableLevels` and updates the unit tests accordingly.

@lilleyse @ptrgags - could one of you take a look?

PR #10094, opened 2022-02-11
Replace `maximumLevel` with `availableLevels` in implicit tiling specs

j9liu:
In #10083, the `maximumLevel` of implicit subtrees was to be replaced with `availableLevels`. I forgot to update the data in the specs to be consistent with this change -- this PR should have taken care of that.

@ptrgags


ptrgags (approved):
@j9liu looks good, sorry I missed these in the last review. I searched through the codebase and confirmed that these are all the places that need to be updated. I'll merge once CI passes.

PR #10098, opened 2022-02-12
Make DrawCommand use a bitmask instead of a bunch of booleans.

kring:
`DrawCommand` had a bunch of boolean fields. This PR turns them into a bitmask instead, which will make commands quite a bit smaller. I doubt you'll be able to measure much of a difference, but this ought to be faster too, despite a few extra bit manipulation instructions.

For a weekend project, I was going to add another flag to DrawCommand. Doing this first makes it easier to justify, because adding a new flag won't increase the size of the DrawCommand. So I thought I'd do it as a separate PR.

As far as I can tell based on a bunch of searching, no one was using any of the old, private bool fields other than via the property getters/setters, so the impact of this change is very localized. I don't think it needs to be mentioned in CHANGES.md.

### DIFF
--- a/Source/Renderer/DrawCommand.js
+++ b/Source/Renderer/DrawCommand.js
@@ -2,6 +2,17 @@ import defaultValue from "../Core/defaultValue.js";
 import defined from "../Core/defined.js";
 import PrimitiveType from "../Core/PrimitiveType.js";
 
+const Flags = {
+  CULL: 1,
+  OCCLUDE: 2,
+  EXECUTE_IN_CLOSEST_FRUSTUM: 4,
+  DEBUG_SHOW_BOUNDING_VOLUME: 8,
+  CAST_SHADOWS: 16,
+  RECEIVE_SHADOWS: 32,
+  PICK_ONLY: 64,
+  DEPTH_FOR_TRANSLUCENT_CLASSIFICATION: 128,
+};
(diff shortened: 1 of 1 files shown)


ebogo1 (commented):
These changes look good to me, though I'm not sure if increasing powers of 2 are easier to keep track of than an extra boolean field. Is it correct that the DrawCommand becomes smaller because a boolean is a byte instead of a bit per flag? 

Maybe @lilleyse to give a final :+1: or :-1: on this - I think it's a neat change but maybe someone else should give final approval before merging.

kring:
> Is it correct that the DrawCommand becomes smaller because a boolean is a byte instead of a bit per flag?

It's all a bit of optimizer magic, but I think even saying each boolean field is one byte is optimistic, particularly because the boolean fields are interleaved with other fields. We can use the Chrome Memory tab to take a look. Take a snapshot, search for DrawCommand, and look at the "Shallow Size" of each DrawCommand instance. That's the size of the instance itself, without counting any other objects it might reference. In main (sandcastle.cesium.com) each DrawCommand instance is 120 bytes. In this branch, each instance is 92 bytes. So it looks like each of the 8 booleans was occupying 4 bytes each.

lilleyse:
:+1:  seems like a fine change to me

PR #10102, opened 2022-02-14
Require class definition in metadata constructors

j9liu:
This PR makes a `MetadataClass` definition a requirement in `MetadataEntity` and consequently, all other metadata-related classes. It deletes the unit tests that expected functionality without a class and updates the others so they don't fail.

### DIFF
--- a/Source/Scene/JsonMetadataTable.js
+++ b/Source/Scene/JsonMetadataTable.js
@@ -33,7 +33,7 @@ export default function JsonMetadataTable(options) {
  * @private
  */
 JsonMetadataTable.prototype.hasProperty = function (propertyId) {
-  return MetadataEntity.hasProperty(propertyId, this._properties);
+  return MetadataEntity.hasProperty(propertyId, this._properties, {});
 };
 
 /**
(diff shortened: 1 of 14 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/MetadataTableSpec.js
```diff
@@ -64,7 +64,7 @@ describe("Scene/MetadataTable", function () {
     );
   });
 
-  it("constructor throws without count", function () {
+  it("constructor throws without parameters", function () {
```
ptrgags, line 67: I think we do want to keep this as "throws without count", just pass in `{}` for class

ptrgags (commented):
@j9liu This looks most of the way there, just a few changes needed.

Source/Scene/JsonMetadataTable.js
```diff
@@ -15,6 +15,9 @@ import MetadataEntity from "./MetadataEntity.js";
  * @constructor
  * @private
  */
+
+const emptyClass = {};
```
ptrgags, line 19: Maybe put a comment here along the lines of "JsonMetadataTable is an older type of metadata table that does not have a class definition"

ptrgags:
Looks good, thanks @j9liu!

Issue #4164, opened 2016-08-01
Fit texture coordinates to rectangle/trapezoid geometry


lilleyse:
It would be useful to support stretching a texture along a rectangle/trapezoid geometry, or perhaps even support explicit texture coordinates. The effect would look like the second image:

<img width="640" alt="screen_shot_2016_07_25_at_11_22_39_pm" src="https://cloud.githubusercontent.com/assets/915398/17307382/d2500752-5802-11e6-9a2b-f247a2dda6a1.png">
<img width="640" alt="cesium_good" src="https://cloud.githubusercontent.com/assets/915398/17307383/d393e3e0-5802-11e6-89ef-b9662f95200e.png">

Martin Paredes created a demo for testing this:

https://cesiumjs.org/Cesium/Apps/Sandcastle/index.html?src=Hello%20World.html&label=Showcases&gist=079b7bbc6c2c16bdaf23c57fe0a7f2c4

hpinkos:
We also need this to support KML latLonQuads

pjcozzi:
We have done this in some of our other projects; see http://blogs.agi.com/insight3d/index.php/2009/11/20/trapezoidal-texture-projection-with-opengl/

hpinkos:
I took the very first steps of this in https://github.com/AnalyticalGraphicsInc/cesium/tree/trapezoid but this requires `GroundPrimitive` to support other appearances than just `PerInstanceColorAppearance`

pjcozzi:
Specifically, this requires texturing ground primitive.  @bagnell has most of the work done in this branch: https://github.com/AnalyticalGraphicsInc/cesium/compare/ground-primitive-tex-coords

hpinkos:
This came up again on the forum: https://groups.google.com/d/msg/cesium-dev/x3sYLgIr8Mc/wH2tdn_hCwAJ

lilleyse:
Came up on the forum again: https://groups.google.com/forum/?hl=en#!topic/cesium-dev/zYikVZcfRMo

lilleyse:
Talked about this offline with @likangning93. This will require specific changes to `GroundPrimitive`.

Some early thoughts:

[...]

PR #10109, opened 2022-02-16
Direct assignment of texture coordinates in PolygonGraphics specification

jeffechua:
Added ability to specify a fill image in properties for GeoJsonDataSource, and added a direct textureCoordinates property to PolygonGraphics, PolygonGeometry and CoplanarPolygonGeometry which overrides the current default stRotation-based texture coordinate computation behaviour. The plumbing is set up so that "texture-coordinates" can be provided in a GeoJsonDataSource and it flows through to be assigned to the geometry st in computeAttributes.

This is a feature that's really useful for me, since trying to coax stRotation into doing what I want is not necessarily the most scalable solution even if the feature is very convenient for some cases.


cesium-concierge:
Thank you so much for the pull request @jeffechua! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/jeffechua/cesium/blob/custom-texture-coordinates/CONTRIBUTORS.md) file!
* :x: Missing CLA.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jeffechua:
Fixed CONTRIBUTORS.md and sent in CLA.

ggetz:
Hi @jeffechua! Can you tell us a little more about the use of this property in geojson? Is it a standard feature, or is this specific to your workflow?

ggetz:
@jeffechua Thanks for the context! The "stroke" and "fill" properties are part of the [simplestyle spec](https://github.com/mapbox/simplestyle-spec) which is supported in other implementations, such as Mapbox. Since this new property isn't more broadly supported, its probably not the best fit for the core Cesium library.

As for the texture coordinate support in Polygon graphics, can you describe why you need to adjust the texture coordinates with more granularity than rotation? Depending on your use case, there are alternatives.

jeffechua:
@ggetz Perfectly understandable about the simplestyle spec. My use case is that I have models which need to be formatted in GeoJSON for compatibility purposes with other software which interacts with the data, which come in textures mapped by texture coordinates. It is my understanding that Cesium currently supports KML/glTF models with texture coordinates? It would technically be possible to do the conversion, but it would result in a lot of additional complexity.

More in general, stRotation does not seem to easily support atlases, and also it's unclear from the documentation how surfaces not parallel to the ground behave—I'm sure it's deterministic and can be finagled into behaving as desired for any specific single textured polygon, but it's very much not intuitive, and developing an automatic way to convert 3D geometries textured in texture coordinates to an stRotation and ImageMaterialProperty specification seems decidedly nontrivial.

To be honest, it is a bit odd to me that Cesium currently does not support the standard, most common and most straightforward way of texture mapping—providing texture coordinates in the vertex attributes—and instead forces users to use a system which is very useful for a lot of use cases but which I would have expected to be a convenience feature strapped onto the normal method. I would be happy to hear or read about the logic behind this design decision if there is something I'm not understanding about the problem.

ggetz:
@jeffechua Sorry for the delayed response--

As far as I'm aware, typically users do not have a need to provide custom texture mapping through a GeoJSON file. So if I understand correctly, you are formatting a model as a set of polygons in GeoJSON? 

We try to ere on the side of providing the best default we can rather than giving users an overwhelming amount of options. With glTF, which is typically how you'd get a model into CesiumJS, the UVs are embedded in the file and there's no need to change it at runtime. Though there has been some cases of developers wanting to use custom UVs - https://community.cesium.com/t/manually-set-uv-texture-coordinates-on-model/10264/2.

jeffechua:
Yes, that is correct. I have also seen that post; thanks for digging it up. I agree that having a simple default and avoiding option bloat is a good design philosophy, but as simple a use case as rendering a single right-angled triangle with a texture correctly is an exercise in trigonometry and coordinate conversion with stRotation, not to mention any irregular triangle—furthermore, it is downright impossible to map a texture to a polygon with any stretching, distortion or cropping. Meanwhile, these are absolutely trivial with texture coordinates—instead of laborious image editing and math, one simply provides three pairs of numbers.

It is not so much a matter of changing UVs at runtime as _mapping the textures in the first place_, which is a headache with the current set up for anything that isn't mapping the whole of a rectangular texture onto an exactly rectangular geometry. One must agree that it is somewhat ridiculous to have to do a bunch of calculations to determine what a set of texture coordinates is in stRotation terms, only for the system to take that rotation and do a bunch of calculations to transform it back into texture coordinates. In my view, the ability to set texture coordinates are a critical feature for texture rendering, if perhaps slightly less so in the broader rendering space, and is at least on par with stRotation in terms of how worthwhile it is to have.

With respect to option bloat, texture coordinates/UVs are such a standard concept that they should be fairly self-explanatory, and if someone doesn't know what they are, the name is sufficiently descriptive that someone not looking for it will be immediately able to discard it. Essentially, while it does put one more item on the property list, I don't see it adding significant additional mental overhead to users of PolygonGraphics. The usefulness-to-added complexity ratio here is beyond favourable, in my opinion.

lilleyse:
Hi @jeffechua. This reminds me of an issue I opened a while back: https://github.com/CesiumGS/cesium/issues/4164

> It would be useful to support stretching a texture along a rectangle/trapezoid geometry, or perhaps even support explicit texture coordinates.

I think this would be a useful addition to the API but there are some details to work out.

* Does it work if the polygon has holes? E.g. if a non-trivial `polygonHierarchy` is supplied.
* Does it work for ground primitives? Explicit texture coordinates would need to be remapped to ENU coordinates in  [Geometry._textureCoordinateRotationPoints](https://github.com/CesiumGS/cesium/blob/999ee4e6b82438996c30955cc47103bcfce4a35f/Source/Core/Geometry.js#L229-L253).
* Does it work for extruded polygons? I think so based on a quick look at the code.

There are other cases I may not have considered. I would suggest at least splitting this into two PRs: one for the geometry changes and another for the GeoJSON changes. I haven't really though much about the GeoJSON part but it would be good to see other examples of textured GeoJSON in the wild to see if there's a common texture coordinate definition.

jeffechua:
Hi, thank you for your comment. In order:

- It does work if the polygon has holes, including multiple holes, and the texture coordinates simply map like the vertices are serialised. However, it seems not to work with _nested_ holes. The behaviour looks to me as if nested holes get decomposed into different geometries and the texture coordinates get copied, not partitioned—it's not clear to me where this is happening as I am not super familiar with the code base, and it's possible I'm completely wrong. If my guess is right, this should be fixable; even if not, this is a rare use case.
- It does not work for ground primitives. Right now, texture coordinates do nothing on a ground primitive, which is better than it breaking, at least.
- It works for extruded polygons in a sense. An extruded polygon will consume more texture coordinates; the top face of the extrusion (the actual positions passed in by the user) will map to the first texture coordinates provided. However, the order of the other vertices is non-obvious—probably determinable from the code which generates them, but not the logic is not exposed to the end user.

I have added a Sandcastle demo, development/Polygon Textures, to illustrate the above behaviour.

It also seems that non-extruded 2D polygons with a large enough area to experience significant curvature are altered to follow the curvature of the Earth (?), creating many extra points and breaking the texture coordinate input. This might be theoretically fixable (as in, make whatever function is doing this also dynamically generate the extra texture coordinates), but it also might be simpler to leave as a not-supported edge case.

I agree about splitting the GeoJSON part and geometry part.

Can I ask if there is a way to know whether, if these details are resolved, I can reasonably expect that the geometry texture coordinates pull request be merged? I asked because @ggetz was a bit skeptical about it, and some indication of a merge being probable would be appreciated before I spend time on fixes not relevant to my own use case.

ggetz:
Thanks @lilleyse and @jeffechua!

@jeffechua I'm still a bit skeptical about the geojson portion of this, as the use does seem pretty niche. However, based on the above conversation we'd be happy to move forward with a PR for the geometry texture coordinates portion. The nested holes and ground primitive limitations are not deal breakers, though we may want to experiment a bit to see if its possible.

jeffechua:
I have removed the GeoJsonDataSource changes.

I have also fixed the nested holes behaviour to behave correctly now. In addition, I have clarified in the documentation and through the Sandcastle demo development/Polygon Texture Coordinates the exact scope of support of the texture coordinates feature.

- Extrusions are not supported.
- Ground primitives are not supported.
- Non-perPositionHeight polygons large enough to be subdivided over the Earth's curvature are not supported.

These all fail gracefully, i.e. the texture coordinates are ignored, as seen in the Sandcastle example. The textureCoordinates members in relevant classes are now documented as:

_Texture coordinates as a PolygonHierarchy of Cartesian2 points. Has no effect for extrusions, ground primitives, and sufficiently large polygons without per-position heights._

In my personal opinion, the three failing cases are sufficiently distinct features from the main use case—sort of orthogonal to the texture coordinates feature, if that makes sense, and relatively unlikely to be required in combination, though I may be corrected on this—and work on them could be a separate feature request/pull request. This is in contrast with the nested holes issue, which was a genuine bug (and which is now fixed). So, I am inclined to suggest that this is an appropriate stopping point in terms of feature scope? Of course someone may want to scrutinise the code and documentation for quality and style.

Since the title of this pull request is still accurate, and I'm not really too invested in the GeoJSON portion, do we still want to have a separate pull request or is here fine?

Source/Core/CoplanarPolygonGeometry.js
```diff
@@ -128,18 +129,27 @@ function createGeometryFromPolygon(
...
-        position,
-        scratchPosition
-      );
-      const st = projectPointTo2D(p, stScratch);
-      Cartesian2.subtract(st, stOrigin, st);
+      if (
+        hardcodedTextureCoordinates &&
+        hardcodedTextureCoordinates.positions.length === length
+      ) {
+        textureCoordinates[i * 2] = hardcodedTextureCoordinates.positions[i].x;
```
sanjeetsuhag, line 136:
```suggestion
        textureCoordinates[stIndex++] = hardcodedTextureCoordinates.positions[i].x;
```

Would this also just work? It'd be better if we could just use the variable in both cases.

Source/Core/CoplanarPolygonGeometry.js
```diff
@@ -334,7 +359,17 @@ CoplanarPolygonGeometry.pack = function (value, array, startingIndex) {
...
-  array[startingIndex] = value.packedLength;
+  if (value._textureCoordinates) {
+    startingIndex = PolygonGeometryLibrary.packPolygonHierarchy(
+      value._textureCoordinates,
+      array,
+      startingIndex,
+      Cartesian2
+    );
+  } else {
+    array[startingIndex++] = -1.0;
```
sanjeetsuhag, line 370: I understand what this is doing, but I'm not sure if it would be the recommended approach. @lilleyse Do you have any thoughts here for setting a flag like this for packing/unpacking an optional value?

Source/Core/CoplanarPolygonGeometry.js
```diff
@@ -468,12 +521,24 @@ CoplanarPolygonGeometry.createGeometry = function (polygonGeometry) {
...
     false
   );
   const hierarchy = results.hierarchy;
   const polygons = results.polygons;
 
+  const textureCoordinatePolygons = textureCoordinates
+    ? PolygonGeometryLibrary.polygonsFromHierarchy(
+        textureCoordinates,
+        true,
+        function (identity) {
```
sanjeetsuhag, line 535: This function should be declared as a variable, then passed in.

Source/Core/PolygonGeometry.js
```diff
@@ -1102,6 +1155,17 @@ PolygonGeometry.createGeometry = function (polygonGeometry) {
   const hierarchy = results.hierarchy;
   const polygons = results.polygons;
 
+  const textureCoordinatePolygons = textureCoordinates
+    ? PolygonGeometryLibrary.polygonsFromHierarchy(
+        textureCoordinates,
+        true,
+        function (identity) {
```
sanjeetsuhag, line 1162: Same comment applies here, as above.

sanjeetsuhag (commented):
Thanks for the contribution @jeffechua! The Sandcastle works great - this is an awesome feature to add to CesiumJS. I've left a few comments on the code style, documentation and some areas that need clarification.

Source/Core/CoplanarPolygonGeometry.js
jeffechua: Agreed, will commit in batch.

Source/Core/CoplanarPolygonGeometry.js
```diff
@@ -468,12 +521,24 @@ CoplanarPolygonGeometry.createGeometry = function (polygonGeometry) {
 
   const results = PolygonGeometryLibrary.polygonsFromHierarchy(
     polygonHierarchy,
+    defined(textureCoordinates),
```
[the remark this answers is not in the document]
jeffechua: Fixed.

Source/Core/CoplanarPolygonGeometry.js
jeffechua: Fixed

jeffechua: Fixed.

Source/Core/PolygonGeometry.js
jeffechua: Fixed.

lilleyse:
> Non-perPositionHeight polygons large enough to be subdivided over the Earth's curvature are not supported

What's the level of effort for generating texture coordinates for the subdivided geometry? A size limitation would feel very arbitrary from an end user perspective.

I'd also be curious to know the level of effort for extruded polygons. I'm not deeply familiar with the geometry code but when I looked at it earlier it seemed like the attributes would be copied for the extruded face, and I wonder if it can be made to work without too many changes.

jeffechua:
For both cases it feels like the kind of thing that would be an hour or two for someone who is deeply familiar with the code, but quite many times more for someone who is not, since the changes would have to be woven into the (branching!) pipelines performing subdivision and extrusion. I definitely expect it to be possible.

It would, however, be very inelegant and clunky, since you would essentially be passing both texture coordinate PolygonHierarchy and position PolygonHierarchy into everything, checking every other line whether or not the texture coordinates are defined, and then doing texture coordinate copies/interpolations, probably increasing the length of those functions by 50-100%?

It would be more elegant to define a new coordinate class by extending Cartesian3 with texture coordinates (5 numbers) and overloading/implementing the arithmetic to correctly interpolate and project, which would slot nicely into the current pipeline without adding too much ugly plumbing and repetition. However, this would be even more work and something I don't have the Cesium ion familiarity to do.

---

I think that the size limitation is a little arbitrary taken at face value, but also vaguely delimits two meaningfully distinct use cases? Essentially, the texture coordinates are good for texturing buildings or other city structures, for applications "on the ground", but not for shapes spanning significant fractions of the Earth's surface, "visible from space". Also, the problem does not exist if heights are provided, which solves most use cases which aren't "colour in this octant of the Earth".

Similarly, using an extrusion is sort of saying "I don't want to think too hard about my thing; can Cesium handle the heavy lifting for me if I give some minimal specification?" and is more on theme with using stRotation than explicitly providing texture coordinates. (Of course, it being more on theme to use a particular combination of features has never stopped anyone from requiring a different, unconventional combination because of convoluted reasons.)

Perhaps my perspective is better articulated by saying that the competency I was aiming to implement was "directly render a textured 3D model through PolygonGeometry/CoplanarPolygonGeometry", as a more convenient alternative to "convert to glTF". This is useful for piping in data which isn't pre-formatted in nice model files, or for mutating textured 3D data at runtime, and it is basically trivial to think of an number of use cases which are marginally convenienced by this (animating adding floors to a building by simultaneously modifying texture coordinates and positions, procedurally generated cities, a scrolling billboard...). Texturing extrusions and "flat" areas on the Earth's curved surface with high specificity I saw as existent but distinct use cases from this, and I don't have off the top of my head any specific applications (...overlaying a choropleth map on a country? It would probably be easier to stRotation that, though...).

---

If it were up to me, I would keep the scope here for now and register "texture coordinate support for extrusions and subdivided polygons" as a "type - enhancement" issue, copy this discussion, and see if anyone with a relevant use case finds the lack inconvenient enough to intervene, but I am of course biased. :)

sanjeetsuhag:
Thanks for the additional context, @jeffechua. I see that this feature brings good value to Cesium for the use case of small polygons, (buildings, models, etc), however, from a usability perspective, the fact that a user cannot use it beyond an arbitrary size limit is something of a blocker at the moment. We currently do not have the bandwidth to implement the fix in this PR, however, we're more than happy to review any new progress on that front.

sanjeetsuhag:
I looked into why larger polygons were breaking with the current implementations. First, I investigated a small sample, and scaled it up till it broke.

| [Polygon A](http://localhost:8080/Apps/Sandcastle/index.html#c=lVVhb9owEP0rJz7BRJ3QqR+W0WoVbadKm1a13TRpVJVJjuDNiSPboWUT/33nJA4psLZDCOJ37853785OEMBEI7cIdoGwFPiAmk3zWOXGNks4hhwfYIJGlBn7VmH9aS+u1hOVWy5y1NPe4L33MzHmSG5NuGpJxo2ZS2c+rLAggPNHnhUEjSI40/wBOJiMSwkWH22pMYFCyVWqcqCvSzOVaoZgSj3nMVK2Ei1RjLCC4lPgJtUJ15aeeP6WzbXKzjDViOZUa77q/5jmAAchOzyCN3VCQ4eEbLSF7HIOdklh9/c1/Gf3neZ3g0oaV1ejwUQpnYicGuUK/ON82oojqKqBbpva2g/74RBGg+ELjPAFBjt6kTN6DaPJ5I5+15t5cd2+kNxe0ornMT4duY+oMrR65a39qvy0QSNPu6qHxLOrnl95jWqfJ6K1j03OS6REHy+UzriNagh87M8kvBZcnhYFcu2SaKGbsiiUtuz2/Pvt1+vzM9aN04TebWK0B6vIayfQuu5/dThOkwSQx4v2EAgvklVQaJFREUs0dAyqc8Y6EE+SvovZ0fLKW70g6Za4m2mC3b7UhrumKt6KEXW32NXK7wWQNba9DhsawJzPtIgj6EAk46qgraa9y4ynOO0Nu7YyF3OS3Gy5AAhHdl6MBdWzCep97z+pVN1PlFSa/SzSrXjrzmrzvB4Mt0tp+h9t6P8/M21QSzwjS2qkjUgDabAxNRuvB9Pczwb4+zXmGWrO5nJ1q7yGCRrrpormO3r+PqxO/xCOQvqw0N9Bvk6lBeXSxGmVXSCnmU2jTqkLZtU1oZQ/3RbhRqdC2HjxL+bBuydcrSSNBkG+6uY/KXWTQlirUEnQG/bGxq4knnj/DyJzukKpZZ/abZFeK+5kBbMy/oWWxcY4V0cdB13XcSKWIJLjPa81iCU3hizzUsob8ZsG72QcEH/HVapKlS90AUi+crTF6ORTDTLGxgEt93tapeSM663IfwE) | [Polygon B](http://localhost:8080/Apps/Sandcastle/index.html#c=lVVtT9swEP4rp35qp+KkTHxYVtBQgQlp0xCwadKKkJtcU29OHNlOoZv633dO4jS0HbCqSuy7596eOztBABON3CLYBcJS4ANqNs1jlRvbbOEYcnyACRpRZuxbJetPe3G1n6jccpGjnvYG772diTFHMmvcVVtSbtRcOvUhG1XSIIDzR54VJBxFcKb5A3AwGZcSLD7aUmMChZKrVOVAf5doKtUMwZR6zmOkfCVaghhhBUUg102yE64trXj+ls21ys4w1YjmVGu+6v+Y5gAHITs8gjd1SkMnCdloS7KLOdgFhd3na/DPxp3md4OKGldXw8FEKZ2InFrlCvzjbNqKI6iqgW6j2toP++EQRoPhC4jwBQQ7ehEzeg2iyeSOnuvNxLhuX0huL2nH8xifDt1HVBlavfLaflV+2kgjD7uqh8Sjq55feY5qmyektcsm5yVSoo8XSmfcRrUIvO/PRLwWXJ4WBXLtkmhFN2VRKG3Z7fn326/X52es66dxvdvEaI+sAq8dQeu6/9XhOE0SQB4v2kMgPElWQaFFRkUs0dAxqE4a64h4kvSdzw6XV17rCUm3yN1ME+z2pVbcNVXxloyoG2KXKx8LIGt0ew02MIA5n2kRR9AREY2rgkJNe5cZT3HaG3Z1ZS7mRLnZMgEQDuysGAuqtQnquPefVKruJ0oqzX4W6Za/dWe3Wa8Hw+1Smv5HG/j/z0zr1BLOyJIaaSPiQBpsVE3g9WCa+9kAf8PGPEPN2VyubpXnMEFj3VTRfEfP34fV6R/CUUg/Fvo7yNeptKBcGj8tswvkNLNp1Cl1way6JinlT7dFuOGpEDZe/At58O4JVitJo0EiX3XzTkrdpBDWLFQU9Ia9sbEriSfe/oPIHK9Qatmndlukz4o7WcGsjH+hZbExztRBx0HXdJyIJYjkeM+HDWLJjSHNvJTyRvymwTsZB4TfMZWqYuULXQCSrxxsMTr5VAsZY+OAtvstrVJyxvWW578) |
| - | - |
| image | image |

I investigated the draw calls using Spector and found that there was an additional position being added in the case of Polygon B.

| Polygon A | Polygon B |
| - | - |
| polygon-texcoord-1 | polygon-texcoord-2 |

Here's what the vertices look like on the globe for the processed polygons:

| Polygon A | Polygon B |
| - | - |
| polygon-points-1 | polygon-points-2 |

This addition of a vertex happens in:
https://github.com/CesiumGS/cesium/blob/ba337eadd1dbea068df2f7b9b96a254c87b40e51/Source/Core/PolygonPipeline.js#L85-L102

The core condition that determines whether subdivision happens is:
https://github.com/CesiumGS/cesium/blob/ba337eadd1dbea068df2f7b9b96a254c87b40e51/Source/Core/PolygonPipeline.js#L189-L191

So, I think the right fix here needs to be that this function needs to be provided with the hard coded texture coordinates, which would be copied/re-arranged based on the subdivision. I did dig too deep into the actual subdivision strategy for the triangles here, but I think it should be possible to extend it to the texture coordinates too. @lilleyse Am I on the right track here?

lilleyse:
> So, I think the right fix here needs to be that this function needs to be provided with the hard coded texture coordinates, which would be copied/re-arranged based on the subdivision. I did dig too deep into the actual subdivision strategy for the triangles here, but I think it should be possible to extend it to the texture coordinates too. @lilleyse Am I on the right track here?

@sanjeetsuhag yes I think so. The texture coordinates would need to be subdivided alongside the other attributes.

sanjeetsuhag:
@lilleyse And before, committing to this fix, do we think this is a worthwhile addition given the context:

> Perhaps my perspective is better articulated by saying that the competency I was aiming to implement was "directly render a textured 3D model through PolygonGeometry/CoplanarPolygonGeometry", as a more convenient alternative to "convert to glTF". This is useful for piping in data which isn't pre-formatted in nice model files, or for mutating textured 3D data at runtime, and it is basically trivial to think of an number of use cases which are marginally convenienced by this (animating adding floors to a building by simultaneously modifying texture coordinates and positions, procedurally generated cities, a scrolling billboard...).

To me, this seems like it's a use case for opening up the `ModelComponents` API, enabling users to generate meshes on the fly by specifying custom vertices/indices/texture coordinates. The proposed fix, I think, would only take a couple of days if we limited the scope to non-extruded, non-ground primitives, but I wanted to make sure that we're actually solving the right problem here.

lilleyse:
Also @sanjeetsuhag, have you looked into extruded polygons at all? It might be easier to support than we think.

sanjeetsuhag:
So, it seems to have worked pretty well!

image

@lilleyse I have not yet looked into extrusions at all, but I'll handle that once I add the specs for this. We may want to consider putting that into a separate PR, but I'll certainly take a look at what's needed to make it work.

sanjeetsuhag:
@lilleyse Turns out supporting extruded polygons wasn't much trouble!

image

Also, just to clarify, are we talking about supporting custom texture coordinates for the generated "walls" too? Because at the moment, the texture coordinates are only applied to the top and bottom faces.

sanjeetsuhag:
Added support for texture "bleeding" on the wall geometry. Here is a [Sandacastle](http://localhost:8080/Apps/Sandcastle/index.html#c=tVjrc9o4EP9XPHyJmSOKDcG80vZSwrXMJIEBco85Mh3FFuBWtjyyTJu74X+/lS2/MCHJNMcHx1r9drXvlWMzPxTa1iXfCdfeaT75rg1J6EYe+j2m6cuaHa+HzBfY9Qlf1uqDpW/HfAELXeHCK7AqtiHmAt6w30IrzrwrsuaEhJec40f976Wvaaem0UXnDfna6qiXEi2ldJCRosyU1kbWHiqnxALupXIV9f5wxeYzcdcb8SJFE2j4tL5tA35PaN4q7lVsKO3uWfPE3t6ZJQsF+SEiToaMccf1MRgkTZ0HLqVg6b+SL3NCX4vt0YpBzpzQ1I2GZtYbzyAM1Gw/A0Ltl8E6r8Apxe7huTtq/f9j9IssfgZlvgxTsXTpz7Hv2DgUlCDsOAvG6APmHyMhmA/VOcSB9MGy1tBWkW9Lo/V64oTESaFNfAJ+SWocxctBcRvS3g83TJZGKkHTY5iSo2nJM2Gwsb/F0tExBCXLQRGxcim5xZ48dFnL5S9rJRR1/W+AcJgdecQXCHAQwRElcqWf4JO6gksgcth3nzLsSCVT8b9oJyjw1yep2FgTJNhHyh70zBkPsMoMyeRtOFmBrLvZtTp48vCV2ALWCcNgD29T1/6mp+RdSTdOPLYl2WbiF8hAMSO+Ay5P9kdbsOvaDQXscj13S8K2i58VVgj4Mb5d0guKoYWAeIRjFBIhO7geG+6QUMgCkaHtH0y+ln5qtjqG2TNRp9tpds12t9tsaKdto3feNC0EK+O8Y1ithtbqAsGwULPXbLYMAFpJzjLugqYHDvlMMJTneuoKezNjlOptBKJMy2xaVrvTkwV+aiDL7HTNVrdttltWy+pYDQ2O6LaAalm9XrNr9FRtgGcWHPvhinEPDlKH3GDB3R/naHw1ul2MF39lzqEEBgF3PegHW6JKKsvztXQqRH/K6OMakoUFcddQ+ZJXSUD4VLWUbI4oLKrsFaoLc3vxGJACWlGKFSi79SJpZQVgkTxItSlFOrMqTFMwo9QHZf1deGDfJuUR/4kwjwj+OFa7uqqStaJnzlXuSfHxwExtDvWstgpdt+qwDwfHcT+nNlIxYDOPHOIkkMOyDODMp7D8bQmk8o/fICmw6OfVm6cH4S6ml0FAMJfGZqR5FASMC7QY/bm4m42uUFFSJr46a/rlyH04Ooz7B3Yz2Son+ulLtlGx/IAz0o4UM6mc1/KM3+v8xaSB5qJXxtE03c/Dut5Lk3yoyl+aWpnS97ldmbNL/aAaC73Qnj21e5ClCNS0FX7grt3XSkQIVezLZW3s4bUcjOXdyHdl5wgrbGCKZJCcCJ3F7+FZcv6Xa7ZmX4aMMo6+BuuKzF1pXVzt6o2qaSrj+kWm1+dpQbCQ/ZBGEGDIkBWmYR6NTIFdXf6Np0Z+g1LNJr02ZYmotLlM1ujTaHI1mo+HsagDSZkfWSwJRS/f2dQRk+zgOJVUKGSNSP9fpnqkB2ceBx5CYVD3C3eVwnTf67L5bX/fkkGe3IcnQDrpk7JqHFNy9vnu5uObaRhLe7169wUfb+KwHHdxGjpIptnd1ejqtfofmohxuN/As6lu09HsdDqZjxfjye1b6AdjhfycZ1UTP+7aLP1vJ9p8Or6+fq3uezeCt3JrptdbKPUzvnzim+aG+JFe7g9SwtPYUp4fh5YDJ7G1Ru0iFI+UvE+t+NX1ZIPVIk51mACCeAGVc/rsIbK/EfhCCTOLLs6KrBeOu9Vc592B/5VoNsVhCDuriNK5+w/Mo/cXZ4CvsMqvG7glT+D2QfGjhG3M99cJESF0cQbLw5wisXNP8n8) to test.

screenshot

sanjeetsuhag:
@lilleyse This PR is ready for you to review. Here's a couple of Sandcastles to test:

| [Development Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FPolygon%20Texture%20Coordinates.html&label=Development) | [Custom Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=tVjrc9o4EP9XPHyJmSOKDcG80vZSwrXMJIEBco85Mh3FFuBWtjyyTJu74X+/lS2/MCHJNMcHx1r9drXvlWMzPxTa1iXfCdfeaT75rg1J6EYe+j2m6cuaHa+HzBfY9Qlf1uqDpW/HfAELXeHCK7AqtiHmAt6w30IrzrwrsuaEhJec40f976Wvaaem0UXnDfna6qiXEi2ldJCRosyU1kbWHiqnxALupXIV9f5wxeYzcdcb8SJFE2j4tL5tA35PaN4q7lVsKO3uWfPE3t6ZJQsF+SEiToaMccf1MRgkTZ0HLqVg6b+SL3NCX4vt0YpBzpzQ1I2GZtYbzyAM1Gw/A0Ltl8E6r8Apxe7huTtq/f9j9IssfgZlvgxTsXTpz7Hv2DgUlCDsOAvG6APmHyMhmA/VOcSB9MGy1tBWkW9Lo/V64oTESaFNfAJ+SWocxctBcRvS3g83TJZGKkHTY5iSo2nJM2Gwsb/F0tExBCXLQRGxcim5xZ48dFnL5S9rJRR1/W+AcJgdecQXCHAQwRElcqWf4JO6gksgcth3nzLsSCVT8b9oJyjw1yep2FgTJNhHyh70zBkPsMoMyeRtOFmBrLvZtTp48vCV2ALWCcNgD29T1/6mp+RdSTdOPLYl2WbiF8hAMSO+Ay5P9kdbsOvaDQXscj13S8K2i58VVgj4Mb5d0guKoYWAeIRjFBIhO7geG+6QUMgCkaHtH0y+ln5qtjqG2TNRp9tpds12t9tsaKdto3feNC0EK+O8Y1ithtbqAsGwULPXbLYMAFpJzjLugqYHDvlMMJTneuoKezNjlOptBKJMy2xaVrvTkwV+aiDL7HTNVrdttltWy+pYDQ2O6LaAalm9XrNr9FRtgGcWHPvhinEPDlKH3GDB3R/naHw1ul2MF39lzqEEBgF3PegHW6JKKsvztXQqRH/K6OMakoUFcddQ+ZJXSUD4VLWUbI4oLKrsFaoLc3vxGJACWlGKFSi79SJpZQVgkTxItSlFOrMqTFMwo9QHZf1deGDfJuUR/4kwjwj+OFa7uqqStaJnzlXuSfHxwExtDvWstgpdt+qwDwfHcT+nNlIxYDOPHOIkkMOyDODMp7D8bQmk8o/fICmw6OfVm6cH4S6ml0FAMJfGZqR5FASMC7QY/bm4m42uUFFSJr46a/rlyH04Ooz7B3Yz2Son+ulLtlGx/IAz0o4UM6mc1/KM3+v8xaSB5qJXxtE03c/Dut5Lk3yoyl+aWpnS97ldmbNL/aAaC73Qnj21e5ClCNS0FX7grt3XSkQIVezLZW3s4bUcjOXdyHdl5wgrbGCKZJCcCJ3F7+FZcv6Xa7ZmX4aMMo6+BuuKzF1pXVzt6o2qaSrj+kWm1+dpQbCQ/ZBGEGDIkBWmYR6NTIFdXf6Np0Z+g1LNJr02ZYmotLlM1ujTaHI1mo+HsagDSZkfWSwJRS/f2dQRk+zgOJVUKGSNSP9fpnqkB2ceBx5CYVD3C3eVwnTf67L5bX/fkkGe3IcnQDrpk7JqHFNy9vnu5uObaRhLe7169wUfb+KwHHdxGjpIptnd1ejqtfofmohxuN/As6lu09HsdDqZjxfjye1b6AdjhfycZ1UTP+7aLP1vJ9p8Or6+fq3uezeCt3JrptdbKPUzvnzim+aG+JFe7g9SwtPYUp4fh5YDJ7G1Ru0iFI+UvE+t+NX1ZIPVIk51mACCeAGVc/rsIbK/EfhCCTOLLs6KrBeOu9Vc592B/5VoNsVhCDuriNK5+w/Mo/cXZ4CvsMqvG7glT+D2QfGjhG3M99cJESF0cQbLw5wisXNP8n8) |
| - | - |
| Screen Shot 2022-05-06 at 9 13 24 AM | Screen Shot 2022-05-06 at 9 13 50 AM |

lilleyse:
@sanjeetsuhag sorry for the delay. I'm not super familiar with this area of the code but I'll give it a pass sometime this week.

Source/Core/PolygonGeometryLibrary.js
```diff
@@ -191,6 +257,57 @@ PolygonGeometryLibrary.subdivideLine = function (p0, p1, minDistance, result) {
...
+  ellipsoid,
+  p0,
+  p1,
+  minDistance,
+  result
+) {
+  // Compute the surface distance.
+  const c0 = ellipsoid.cartesianToCartographic(p0, scratchCartographic0);
+  const c1 = ellipsoid.cartesianToCartographic(p1, scratchCartographic1);
+  const rhumb = new EllipsoidRhumbLine(c0, c1, ellipsoid);
```
lilleyse, line 285: This could be a scratch variable instead. Use `setEndPoints` to reset the start/end.

Source/Core/PolygonGeometry.js
```diff
@@ -696,6 +713,7 @@ function createGeometryFromPositionsExtruded(
  * @param {Boolean} [options.closeTop=true] When false, leaves off the top of an extruded polygon open.
  * @param {Boolean} [options.closeBottom=true] When false, leaves off the bottom of an extruded polygon open.
  * @param {ArcType} [options.arcType=ArcType.GEODESIC] The type of line the polygon edges must follow. Valid options are {@link ArcType.GEODESIC} and {@link ArcType.RHUMB}.
+ * @param {PolygonHierarchy} [options.textureCoordinates] Texture coordinates as a {@link PolygonHierarchy} of {@link Cartesian2} points. Has no effect for extrusions, ground primitives, and sufficiently large polygons without per-position heights.
```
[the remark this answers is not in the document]
lilleyse: There are the few other places that will need this change

Apps/Sandcastle/gallery/development/Polygon Texture Coordinates.html
```diff
@@ -0,0 +1,443 @@
+<!DOCTYPE html>
```
lilleyse, line 1: Could you add some of the polygons from this sandcastle to "Geometry and Appearances" sandcastle? That way there a non-development Sandcastle using this feature that we can look at when doing releases and there's example usage in the entity API.


"Development/Geometry and Appearances" could be updated too just so they stay in sync with each other

Source/DataSources/PolygonGeometryUpdater.js
```diff
@@ -406,6 +406,10 @@ PolygonGeometryUpdater.prototype._setStaticOptions = function (
     Iso8601.MINIMUM_VALUE,
     ArcType.GEODESIC
   );
+  options.textureCoordinates = Property.getValueOrUndefined(
+    polygon.textureCoordinates,
+    Iso8601.MINIMUM_VALUE
+  );
```
lilleyse, line 412: This file doesn't seem complete, e.g. `textureCoordinates` isn't declared in the constructor and there's other places where it might be missing.

Source/DataSources/PolygonGraphics.js
```diff
@@ -100,6 +101,7 @@ function PolygonGraphics(options) {
   this._classificationTypeSubscription = undefined;
   this._zIndex = undefined;
   this._zIndexSubscription = undefined;
+  this._textureCoordinates = undefined;
```
lilleyse, line 104: `_textureCoordinatesSubscription` is missing

Specs/DataSources/PolygonGeometryUpdaterSpec.js
```diff
@@ -272,12 +277,14 @@ describe(
 
     it("Creates coplanar polygon", function () {
```
lilleyse, line 278: I think the test below "dynamic updater sets properties" should be updated as well

Specs/DataSources/PolygonGraphicsSpec.js
```diff
@@ -291,6 +303,7 @@ describe("DataSources/PolygonGraphics", function () {
     );
     testDefinitionChanged(property, "arcType", ArcType.GEODESIC, ArcType.RHUMB);
     testDefinitionChanged(property, "zIndex", 54, 3);
+    testDefinitionChanged(property, "textureCoordinates", [], []);
```
lilleyse, line 306: Would be better to use two different values in this test.

Source/Core/PolygonGeometry.js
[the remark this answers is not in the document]
sanjeetsuhag: Updated documentation in `PolygonGeometry` and `PolygonGraphics`.

Source/Core/PolygonGeometryLibrary.js
sanjeetsuhag: Updated to use scratch variable.

Apps/Sandcastle/gallery/development/Polygon Texture Coordinates.html
sanjeetsuhag: Add an extruded textured polygon and a textured polygon with holes to both Sandcastles. Also, updated the thumbnails for both the Sandcastles.

Source/DataSources/PolygonGeometryUpdater.js
sanjeetsuhag: Thanks for catching this, I completely missed what's going on in this file - I've done a pass over the file and handled `textureCoordinates` like all the other properties.

Specs/DataSources/PolygonGraphicsSpec.js
sanjeetsuhag: Updated the test case to use different values for `textureCoordinates` and `hierarchy`.

sanjeetsuhag:
@lilleyse Addressed your feedback and merged in latest changes from `main`.

lilleyse:
The recent changes look good. Thanks @jeffechua and @sanjeetsuhag

Issue #9875, opened 2021-10-14
Revert early returns on image bitmap specs in Firefox and Safari


ebogo1:
After https://github.com/CesiumGS/cesium/pull/9874 gets merged, the `"correctly ignores custom color profile when ImageBitmapOptions are supported"` and `"correctly ignores gamma color profile when ImageBitmapOptions are supported"` Resource.js specs should stop returning early once Firefox and/or Safari add support for the `colorSpaceConversion` option for [`createImageBitmap()`](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap).

lilleyse:
> All good for the Firefox side. We can live with the occasional color space artifact until the `colorSpaceConversion` option is supported, as we did for several years with Chrome. It's worth the tradeoff of actually being able to use `ImageBitmap` on Firefox now.

This assumption ended up biting us for feature id textures. This issue can be closed once https://github.com/CesiumGS/cesium/pull/10111 is merged.

PR #9624, opened 2021-06-17
Add option to ignore colorspace information in texture images

j9liu:
Fixes #9390.

The `UNPACK_COLORSPACE_CONVERSION_WEBGL` flag was not set to `NONE` as specified in the glTF [documentation](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#texture-data).

When the Sandcastle from the issue is copy pasted into a local session, it should now look like this. The spheres across each row are identical.

image

cc @ebogo1 @shehzan10

PR #10111, opened 2022-02-17
Don't use ImageBitmap unless colorSpaceConversion option is supported

lilleyse:
We were seeing some artifacts in 3D Tiles Next demos that use feature id textures on Safari desktop and mobile. Interestingly the S2 demo, which also uses feature id textures, didn't have these artifacts. After more investigation the only difference was that the first two demos had images with embedded color profiles, and the S2 demo did not.

| | |
--|--
Screen Shot 2022-02-16 at 2 04 42 PM|Screen Shot 2022-02-16 at 1 20 17 PM

The glTF spec requires that embedded color profiles in images are ignored since it's the material definition that determines whether images are considered sRGB or linear. The same rule extends to feature ID textures. There are two ways of ensuring that:

* if `ImageBitmap` is **not** supported then `gl.pixelStorei(gl.UNPACK_COLORSPACE_CONVERSION_WEBGL, gl.NONE)` should be called before the `HTMLImageElement` is passed to `gl.texImage2D`
* otherwise if `ImageBitmap` is supported `colorSpaceConversion: "none"` should be passed to `createImageBitmap`, since `UNPACK_COLORSPACE_CONVERSION_WEBGL` has no effect for `ImageBitmap`

Most of this was already done in https://github.com/CesiumGS/cesium/pull/9624. The problem is that Firefox and Safari support `ImageBitmap` but they don't support the `colorSpaceConversion` option so they would always apply some sort of color correction, which really breaks feature id textures. We don't want gamma correction applied to non-color data :smile: 

So this PR disables `ImageBitmap` if the `colorSpaceConversion` option isn't supported. Unfortunately there's no way for the browser to tell us this, made worse by the fact that the `createImageBitmap` promise resolves even when a non-default `colorSpaceConversion` option is passed in, so the approach here is for `Resource.supportsImageBitmapOptions` to create an ImageBitmap from a PNG with an embedded color profile twice, once with `colorSpaceConversion: "none"` and again with `colorSpaceConversion: "default"`, and check that the results are different.

Downside is that we can no longer use `ImageBitmap` in Firefox and Safari. But at least feature ID textures are working properly now.

Fixes https://github.com/CesiumGS/cesium/issues/9875


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/require-color-space-conversion-for-image-bitmap-support/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu:
Went through the same tests and it works the same, so I'll merge now.

PR #10116, opened 2022-02-17
Metadata type updates from the latest `EXT_structural_metadata` spec

ptrgags:
This PR partially updates metadata types to use the new schema from [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/tree/ext-mesh-features-revision/extensions/2.0/Vendor/EXT_structural_metadata/schema)

It's not everything yet, this was intentional since the changes are already quite large.

This PR addresses the following changes:

* `STRING`, `BOOLEAN` and `ENUM` are now values for `type` not `componentType`
* instead of `ARRAY` we now have a boolean `hasFixedCount` + `count`. (though this is still under discussion...)
* `SINGLE -> SCALAR`
* the schema changes allow for arrays of vectors

this PR does NOT address the following:
* the change from `bufferView -> values`, @j9liu and I think this is better done after a branch they are working on
* Offset/scale. The logic could get a bit tricky about the default handling, I'd like to do that in a separate PR
* `EXT_mesh_features` is now split into `EXT_mesh_features` + `EXT_structural_metadata`. I want to do this in a separate PR to avoid bloating this one.
* functionality of required/optional/noData. This didn't really exist before, so that should be considered as its own PR (something tells me there's lots of details to consider there)

To Do:

- [x] Clarify how arrays of matrices and vectors should be unpacked. This might require changes to `MatrixN`
- [x] Remove the stubs I had for offset/scale, I'll get to those in a future PR.
- [x] Consider unit tests to add to cover new functionality
- [x] Self review to make sure I didn't miss anything. This is a very wide-reaching change, it's very likely I overlooked something.
- [x] Update array types after clarifying details in the schema


Source/Scene/MetadataClassProperty.js
```diff
@@ -336,33 +586,39 @@ MetadataClassProperty.prototype.unnormalize = function (value) {
   return normalize(this, value, MetadataComponentType.unnormalize);
 };
 
+MetadataClassProperty.prototype.scaleRange = function (value) {
+  // TODO
+  return value;
+};
+
+MetadataClassProperty.prototype.unscaleRange = function (value) {
```
ptrgags, line 594: Oh remove these two for now, this will be handled in a future PR

Source/Scene/MetadataClassProperty.js
```diff
@@ -377,22 +633,18 @@ MetadataClassProperty.prototype.unpackVectorAndMatrixTypes = function (value) {
...
-      return value;
+  const MathType = MetadataType.getMathType(this._type);
+  const isArray = this._isArray;
+
+  if (!defined(MathType)) {
+    return value;
   }
+
+  if (isArray) {
+    return MathType.packArray(value, []);
```
ptrgags, line 644: Unfortunately, `MatrixN` doesn't have a `packArray()` function. I'll need to add that. I'll make a separate PR for this today. This will tightly pack the values (unlike packing things for the GPU which requires padding between elements).

Source/Scene/MetadataClassProperty.js
```diff
@@ -377,22 +633,18 @@ MetadataClassProperty.prototype.unpackVectorAndMatrixTypes = function (value) {
  * @private
  */
 MetadataClassProperty.prototype.packVectorAndMatrixTypes = function (value) {
```
ptrgags, line 635: Add some unit tests for this, I don't think MetadadataClassProperty tests this function (it gets tested indirectly through MetadataTableProperty)

Specs/Scene/MetadataTablePropertySpec.js
```diff
@@ -484,42 +495,46 @@ describe("Scene/MetadataTableProperty", function () {
   it("get returns fixed size arrays", function () {
```
ptrgags, line 495: Add tests for getting/setting an array of vectors/matrices (once the array packing is available)

ptrgags (commented):
Did a pass through the code to see what still needs updating

Source/Scene/MetadataClassProperty.js
```diff
@@ -310,6 +363,203 @@ Object.defineProperties(MetadataClassProperty.prototype, {
   },
 });
 
+function isLegacy(property) {
```
ptrgags, line 366: Also add some tests that this legacy detection works.

Source/Scene/MetadataTable.js
```diff
@@ -328,7 +327,7 @@ function getDefault(classDefinition, propertyId) {
   const classProperty = classProperties[propertyId];
```
ptrgags, line 327: unit test this function, looks like it was missing in the code coverage

Source/Scene/MetadataTable.js
```diff
@@ -328,7 +327,7 @@ function getDefault(classDefinition, propertyId) {
...
   if (defined(classProperty) && defined(classProperty.default)) {
     let value = classProperty.default;
-    if (classProperty.type === MetadataType.ARRAY) {
+    if (classProperty.isArray) {
       value = value.slice(); // clone
```
ptrgags, line 331: this also requires deep cloning

ptrgags (commented):
A couple more items after checking code coverage.

Source/Scene/MetadataClassProperty.js
ptrgags: Opened https://github.com/CesiumGS/cesium/pull/10118 for this

Source/Scene/MetadataTableProperty.js
```diff
@@ -273,37 +294,42 @@ function get(property, index) {
...
+  const isArray = classProperty.isArray;
+  const type = classProperty.type;
+  const isVectorOrMatrix =
+    MetadataType.isVectorType(type) || MetadataType.isMatrixType(type);
 
   if (defined(property._unpackedValues)) {
     const value = property._unpackedValues[index];
-    if (classProperty.type === MetadataType.ARRAY) {
+    if (isArray) {
+      // TODO: if this is an array of vectors, maybe we need to deep clone?
```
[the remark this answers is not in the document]
ptrgags: ah yes `clone(x, deep)` would work fine. sometimes I forget that `typeof Array === 'object'` :sweat_smile:

Source/Scene/MetadataClassProperty.js
ptrgags: correction, there were unit tests, just not for arrays of vectors. I'll add some.

Specs/Scene/TilesetMetadataSpec.js
```diff
@@ -292,12 +299,13 @@ describe("Scene/TilesetMetadata", function () {
       class: {
         properties: {
           height: {
+            type: "SCALAR",
             componentType: "FLOAT32",
-            optional: true,
+            required: false,
```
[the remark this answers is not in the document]
ptrgags: Well, right now these are not used, only `default`. I think I'll hold off until a future PR, because the scope for required/noData/etc. is a little ill-defined at present and I don't want to bloat this PR too much.

ptrgags:
@lilleyse this PR is now ready for review

Specs/Data/Models/GltfLoader/BoxInstanced/glTF/box-instanced.gltf
```diff
@@ -16,19 +16,21 @@
           "box": {
```
ptrgags, line 16: I just realized I didn't update the models in Apps/SampleData, I should check that (and double-check the sandcastles while I'm at it)

Source/Scene/MetadataTableProperty.js
```diff
@@ -316,10 +347,14 @@ function set(property, index, value) {
   }
 
   const classProperty = property._classProperty;
+  const isArray = classProperty.isArray;
+  const type = classProperty.type;
+  const isVectorOrMatrix =
+    MetadataType.isVectorType(type) || MetadataType.isMatrixType(type);
```
lilleyse, line 353: Should there be a helper for this, called `MetadataType.hasComponents`? If we add other built-in types like quaternions, colors, etc this helper may come in handy.

Specs/Scene/TilesetMetadataSpec.js
```diff
@@ -417,7 +427,7 @@ describe("Scene/TilesetMetadata", function () {
           position: {
             type: "VEC3",
             componentType: "FLOAT32",
-            optional: true,
+            required: false,
```
lilleyse, line 430: I see some others occurrences of `optional: true` in `MetadataTableSpec.js` and `TilesetMetadataSpec.js`. Can those be updated now or are they waiting for the future required/no-data PR?

lilleyse (commented):
Looks really solid @ptrgags. Just a few comments from me.

lilleyse:
The CDB demo will need to be updated - note the `undefined` values for string and boolean properties - it's calling `property.componentType` but that no longer exists for certain types. For this demo the simplest thing to do is `defaultValue(property.componentType, property.type)`.

image

Source/Scene/MetadataClassProperty.js
```diff
@@ -549,41 +780,61 @@ function checkValue(classProperty, value) {
...
-    length = value.length;
-    for (i = 0; i < length; ++i) {
-      value[i] = normalizeFunction(value[i], valueType);
-    }
-  } else if (
+  if (
+    isArray ||
     MetadataType.isVectorType(type) ||
     MetadataType.isMatrixType(type)
   ) {
```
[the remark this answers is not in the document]
ptrgags: Yes, see the spec `Scene/MetadataClassProperty normalize array values`, I added a vector and a matrix property (including a variable length one).

The normalize function operates on the flattened array values so no changes were needed here

Source/Scene/MetadataTableProperty.js
ptrgags: Actually, I recently updated `MetadataType.getComponentCount()` to always return a value even for `STRING` and `ENUM`. checking for `MetadataType.getComponentCount(type) > 1` can be used for this, and will scale better to these cases. 

I'll sweep through and see what parts of the code would benefit from this

Specs/Scene/TilesetMetadataSpec.js
ptrgags: ah yeah I had just missed some spots

Source/Scene/MetadataTableProperty.js
ptrgags: just updated this. Note that there were a couple places in `MetadataClassProperty` (`isLegacy` and `parseType`) where I intentionally left the `isVector/isMatrix` calls in place to A. be more explicit, and B. because for legacy schemas `type` isn't always a valid `MetadataType` yet

Source/Scene/MetadataClassProperty.js
```diff
@@ -466,55 +722,30 @@ function validateVectorOrMatrix(value, type, componentType) {
...
     if (javascriptType !== "string" || !defined(enumType.valuesByName[value])) {
       return `value ${value} is not a valid enum name for ${enumType.id}`;
     }
     return;
   }
+}
 
-  const valueType = classProperty._valueType;
-  const normalized = classProperty._normalized;
+function validateScalar(value, valueType, normalized) {
```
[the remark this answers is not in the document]
ptrgags: I'll check through carefully, I know the `isFinite()` checks would catch `NaN`s, but that's not in every if branch.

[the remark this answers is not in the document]
ptrgags: Reorganized the logic of `validateScalar` to make sure each branch checks `isFinite()` at the appropriate time, and also added a unit test for this.

ptrgags:
@lilleyse updated!

PR #10118, opened 2022-02-18
Add `packArray()` for Matrix types

ptrgags:
To make the `MatrixN` API more consistent with `CartesianN`, I added `packArray()` and `unpackArray()`. I also updated the unit tests for matrices to use `createPackableSpecs` and `createPackableArraySpecs()` for consistency.

This change is needed for #10116, since `EXT_structural_metadata` supports arrays of matrices.

@lilleyse could you review?


Source/Core/Matrix2.js
```diff
@@ -86,6 +87,67 @@ Matrix2.unpack = function (array, startingIndex, result) {
...
+  //>>includeEnd('debug');
+
+  const length = array.length;
+  const resultLength = length * 4;
+  if (!defined(result)) {
+    result = new Array(resultLength);
+  } else if (!Array.isArray(result) && result.length !== resultLength) {
+    throw new DeveloperError(
+      "If result is a typed array, it must have exactly array.length * 4 elements"
+    );
```
lilleyse, line 110: This should be inside a debug pragma block. Looks like the `Cartesian` classes need to be updated as well.

Source/Core/Matrix4.js
```diff
@@ -167,6 +168,67 @@ Matrix4.unpack = function (array, startingIndex, result) {
...
+  Check.defined("array", array);
+  //>>includeEnd('debug');
+
+  const length = array.length;
+  const resultLength = length * 16;
+  if (!defined(result)) {
+    result = new Array(resultLength);
+  } else if (!Array.isArray(result) && result.length !== resultLength) {
+    throw new DeveloperError(
+      "If result is a typed array, it must have exactly array.length * 4 elements"
```
lilleyse, line 190:
```suggestion
      "If result is a typed array, it must have exactly array.length * 16 elements"
```

Specs/Core/Matrix3Spec.js
```diff
@@ -1505,4 +1465,54 @@ describe("Core/Matrix3", function () {
...
+      1,
+      2,
+      2,
+      2,
+      3,
+      3,
+      3,
+    ],
+    9
+  );
```
lilleyse, line 1517: These might be justifiable uses for `// prettier-ignore`

ptrgags:
@lilleyse updated!

PR #10120, opened 2022-02-21
Update normalization equations for 3D Tiles metadata

lilleyse:
The 3D Metadata Specification now has equations for normalization, defined [here](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/specification/Metadata#normalized-values). These are the same as the equations defined by glTF for animations ([link](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#animations)). The main change is that the `min` value and the `min + 1` value both map to `-1.0` 

* E.g. for `INT8` — `-128` and `-127` both map to `-1.0`

### DIFF
--- a/Source/Scene/MetadataClassProperty.js
+++ b/Source/Scene/MetadataClassProperty.js
@@ -426,7 +426,7 @@ function parseType(property, enums) {
     arrayLength = property.count;
     isVariableLengthArray = !defined(property.count);
   } else {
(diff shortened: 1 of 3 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

IanLilleyT (changes requested):

IanLilleyT (approved):

PR #10123, opened 2022-02-22
Add implicit content metadata to 3DTiles

### LINKED ISSUE #10104: Handle `tileMetadata` and `contentMetadata` in implicit tiling
The latest schema for [implicit tiling subtrees](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions-implicit/extensions/3DTILES_implicit_tiling/schema/subtree) handles metadata a bit differently. Now there are dedicated `tileMetadata` and `contentMetadata` fields in the subtree JSON file.

Older schema (for backwards compatibility): (see the [spec](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_metadata#implicit-tile-properties))

- tile metadata is stored in `subtreeJson.extensions.3DTILES_metadata`
- content metadata didn't exist

Newer schema: [implicit tiling subtree schema](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions-implicit/extensions/3DTILES_implicit_tiling/schema/subtree)

- tile metadata is now stored in `subtreeJson.tileMetadata`
- content metadata now exists as an array in `subtreeJson.contentMetadata`. This will contain 1 element in most cases, but multiple if `3DTILES_multiple_contents` is used
- the schema for the property table has a couple fields renamed (like `bufferView -> values`) so look at the schema carefully.

This involves:

- [x] in `ImplicitSubtree.js`, there was a variable called `tileMetadataExtension`. `tilePropertyTableJson` might be a better name. 
- [x] in `ImplicitSubtree.js`, locate the property table JSONs based on either the new schema or the old schema as described above
- [x] in the same file, update markActiveMetadataBufferViews()` (or make that one a legacy function and create a new one) to handle both the new schema and the old schema
- [x] in the same file, update the parsing logic (see `parseMetadataTable()` to create not just the `tileMetadataTable`, but also an array of `contentMetadataTables` (also of type `MetadataTable` which will become the owner of the typed arrays)
- [x] Add a `contentMetadataTables` getter. 
- [ ] Create a `ImplicitContentMetadata` class, modeled off `ImplicitTileMetadata`. Though two key differences: 1. it should take an array of metadata tables. 2. in addition to the `entityId`, you also want a `contentIndex` (to select the correct content metadata in the case of multiple contents). Note: `ImplicitTileMetadata` is like a weak pointer into `implicitSubtree.tileMetadataTable` for a single tile.
- [x] For each of the `Cesium3DTile` subclasses (there are many of them in `Scene`, we'll need to add a `this.metadata` property of type `ImplicitTileMetadata`.)
- [ ] Create the `ImplicitContentMetadata` and attach it to `content.metadata`... not 100% where's the best place to put this code, more on this below.
- [x] Update `Cesium3DTileFeature.getPropertyInherited`, check for `content.metadata` and if so do the usual `getPropertyBySemantic/getProperty` checks.

### Attaching the metadata to the content.

The one tricky part is figuring out the best way to attach the metadata to the content. 

This can't be done in `Implicit3DTileContent` in exactly the same way we do for tile metadata, because this transcoder creates tiles, but the contents are not loaded until later. 

Some possibilities to consider:

* in `Implicit3DTileContent`, you could set a flag like `childTile.hasImplicitContentMetadata`, and then in `Cesium3DTile.makeContent()`, check for that flag and after creating the content, set `content.metadata = new ImplicitContentMetadata`. There might be some caveats here (e.g. `Multiple3DTileContent` needs to be handled differently)
* OR in each `Cesium3DTileContent` (there's many subclasses), in the constructor check for `tile.implicitSubtree.contentMetadataTables[0]` and create the object as needed. I'm not too keen on this approach since it involves quite a bit of changes to many files, but it's a possibility in a pinch.

EDIT: I think the first case will be the better approach, considering we'll also have to implement #10105 soon too.

CC @j9liu

j9liu:
Closes #10104.

This PR presents the following changes:
- Create new class `ImplicitMetadataTableView` to handle both implicit tile and content metadata (thus removing the need for `ImplicitTileMetadata`.
- Update `ImplicitSubtree` to handle content metadata
- Add `metadata` field to the `Cesium3DTileContent` interface, as well as all subclasses
- Update `Cesium3DTileFeature.getPropertyInherited` to access the content metadata

TODO:
- [ ] Create unit tests to test handling of multiple contents in`Multiple3DTileContentSpec`
- [ ] Update the implicit tile metadata test in `Cesium3DTileSpec`
- [ ] Add a content metadata one to the same file

@ptrgags - if you could take a look at what I have so far, that would be great!


Specs/Scene/Multiple3DTileContentSpec.js
```diff
@@ -365,6 +365,28 @@ describe(
...
+          }
+        );
+      });
+
+      it("assigning content metadata throws", function () {
+        return Cesium3DTilesTester.loadTileset(scene, multipleContentsUrl).then(
+          function (tileset) {
+            expect(function () {
+              const content = tileset.root.content;
+              content.groupMetadata = groupMetadata;
```
ptrgags: This should be setting `content.metadata`,

ptrgags (commented):
@j9liu I did a first pass through the code, but not the unit tests next (I'll look at those next)

It's looking pretty solid overall, though there's a few nuances to point out, + a few small changes to update the metadata types.

ptrgags (commented):
@j9liu this is getting pretty close! just a few minor things to refactor in the unit tests.

I also checked through the sandcastles to make sure the implicit examples (CDB Yemen and the S2 globe) are working properly

ptrgags:
Looks good to me! Thanks @j9liu!

Issue #7249, opened 2018-11-08
Erratic test failure in master


mramato:
The last master build failed with the below error.  If I had to guss, this test is using an external server.  We should probably bring whatever data its using in-house to make sure CI doesn't randomly fail.

CC @likangning93 @OmarShehata 

```
  Core/GoogleEarthEnterpriseTerrainData
    upsample
      ✗ works for all four children of a simple quad
	Error: Timeout - Async callback was not invoked within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 5096)
	<Jasmine>
5.002 secs: Core/GoogleEarthEnterpriseTerrainData upsample works for all four children of a simple quad
3.236 secs: Core/GoogleEarthEnterpriseTerrainData createMesh creates specified vertices plus skirt vertices
1.667 secs: Core/GoogleEarthEnterpriseMetadata resolves readyPromise
1.187 secs: Core/VideoSynchronizer Syncs time when not looping
1.151 secs: Core/ApproximateTerrainHeights initializes
1.132 secs: Scene/GroundPrimitive does not pick when allowPicking is false
1.109 secs: Scene/GroundPrimitive larger scene renders batched instances
1.087 secs: Scene/Vector3DTileContent renders polygons
1.069 secs: DataSources/GeometryVisualizer Creates and removes static color open geometry
1.053 secs: Scene/GroundPrimitive picking in 2D
Firefox 63.0.0 (Linux 0.0.0): Executed 10347 of 10355 (1 FAILED) (skipped 8) (2 mins 22.893 secs / 2 mins 13.272 secs)
TOTAL: 1 FAILED, 10346 SUCCESS
```

OmarShehata:
I just had a build fail on Travis I believe for a similar reason as above:

```
    ✗ does not change commands for pick calls
	Error: Timeout - Async callback was not invoked within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 5451)
	<Jasmine>
  Scene/Tileset3DTileContent
    ✗ resolves readyPromise
	Error: Timeout - Async callback was not invoked within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 5451)
	<Jasmine>
[...]

PR #10133, opened 2022-02-23
Remove `xdescribe` from `Vector3DTileContentSpec`

ebogo1:
This `xdescribe` was from [this comment](https://github.com/CesiumGS/cesium/issues/7249#issuecomment-548177981) in #7249. I ran Travis three times on this branch and all three builds passed; there will be a fourth run for the second commit. If that passes too this should be safe to merge to `main`. 

I don't want to use the "fixes" keyword since #7249 has more than this test failure but that issue can probably be closed after this is merged.

@ggetz Could you please review/merge once CI passes? Thanks!

### DIFF
--- a/Specs/Scene/Vector3DTileContentSpec.js
+++ b/Specs/Scene/Vector3DTileContentSpec.js
@@ -23,7 +23,7 @@ import Cesium3DTilesTester from "../Cesium3DTilesTester.js";
 import createScene from "../createScene.js";
 
 // See https://github.com/CesiumGS/cesium/issues/7249#issuecomment-546347729
-xdescribe(
+describe(
   "Scene/Vector3DTileContent",
   function () {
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/vector-specs-test/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

mramato:
@ebogo1 did you run these tests locally? Because they still fail for me every time. travis does not run with a real WebGL context, so I'm curious if they pass for anyone else.

ebogo1:
@mramato I definitely thought I did but I can't get them to pass so maybe not. I'll close the PR since there's more digging to do here.

PR #10136, opened 2022-02-23
Fixed leftover `var` 

IanLilleyT:
The `var` -> `const`/`let` tool didn't catch places that had:
- `eslint-disable-next-line`
- `eslint-disable-line`

This PR fixes this and uses `eslint-disable-next-line` instead of `eslint-disable-line`.

### DIFF
--- a/Apps/HelloWorld.html
+++ b/Apps/HelloWorld.html
@@ -28,7 +28,7 @@
   <body>
     <div id="cesiumContainer"></div>
     <script>
-      var viewer = new Cesium.Viewer("cesiumContainer");
+      const viewer = new Cesium.Viewer("cesiumContainer");
     </script>
   </body>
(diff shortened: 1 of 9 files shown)


cesium-concierge:
Thanks for the pull request @IanLilleyT!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/var-const-let-leftover/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Looks good @IanLilleyT, thanks!

I did notice that the entirety of `buildModuleUrl` is ignored by eslint, which is why we missed it in the first place. But I'd want to confirm why we do that before we make a change, so no need to hold up this PR.

Issue #10125, opened 2022-02-22
camera.getPickRay throws exception if scene is not visible


mramato:
Calling `getPickRay` throws an exception if the scene is not rendered:

* Paste the following code into Sandcastle:
```js
const viewer = new Cesium.Viewer("cesiumContainer");

Sandcastle.addToolbarButton('Pick', function(){
    const ray = viewer.scene.camera.getPickRay(new Cesium.Cartesian2(0,0));
});

Sandcastle.addToolbarButton('Hide', function(){
  viewer.scene.canvas.style.display = 'none';
});
````
* Hit `Pick` (no problems)
* Hit `Hide` (globe disappears)
* Hit `Pick` again, exception inside of `getPickRay`

I'm pretty sure all of our other pick operations return `undefined` if the pick fails, but it's been a while. It doesn't make sense for the user to have to perform the additional checking here, since it's all internal to Cesium.

ephjos:
Related https://github.com/CesiumGS/smart-construction/issues/5560

mramato:
Looks like this was fixed in the above PR, but not closed?

ggetz:
Yes, fixed in https://github.com/CesiumGS/cesium/pull/10139.

PR #10139, opened 2022-02-24
camera.getPickRay exception when Scene is not rendered

srothst1:
The goal of this pull request is to address the issue with `camera.getPickRay` outlined in #10125

`camera.getPickRay` is currently throwing an exception that is not consistent with the rest of our `Camera` spec when `Scene` is not rendered. Based on the documentation (see [pickEllipsoid](https://cesium.com/learn/cesiumjs/ref-doc/Camera.html?classFilter=came#pickEllipsoid), etc.) and various tests that I ran, pick operations are supposed to return `undefined` if the pick fails.

Currently, after the viewer is updated as follows:
```
viewer.scene.canvas.style.display = 'none';
```
I receive this error:
```
Uncaught DeveloperError: normalized result is not a number
Error
    at new DeveloperError (http://localhost:8080/Source/Core/DeveloperError.js:39:11)
    at Function.Cartesian3.normalize (http://localhost:8080/Source/Core/Cartesian3.js:422:11)
    at getPickRayPerspective (http://localhost:8080/Source/Scene/Camera.js:2922:14)
    at Camera.getPickRay (http://localhost:8080/Source/Scene/Camera.js:2994:12)
    at <anonymous>:7:37
    at HTMLButtonElement.button.onclick (http://localhost:8080/Apps/Sandcastle/Sandcastle-header.js:62:9) (on line 422 of http://localhost:8080/Source/Core/Cartesian3.js)
```

I am using [this sandcastle demo](http://localhost:8080/Apps/Sandcastle/index.html#c=nVE9b8IwEP0rVpYECTm0awOqytKhUqtSdcpy2AdYOGdkX4Jo1f/eS7IgYKjqyfd870PPJlBi1Tk8YlRzRXhUS0yubfTngBV1ZoZ5GYjBEcY6mzzUVNMKyBpI7FGDtR8h+DXEp5Y5UJG/ObPPp2rTkmEnwOS7JqXMYBbhJE6jpU4GCbWBBiPoLXJPfIdTcRZkCZHlBnRfzKazSe8+SgWx9mFbiGAP/vwh17OzeJ3rIgt1kHTikyhYlw5+yJtTIMz/aLPaheM/bdY+SHWjTzbNqmFh0dP78+iaQ4is2ugLrUvGRnjSTrluzR5Zm5TGfpSqynNqZV2nnJ3f+E5lPKQkL5vW+5X7wjpbVKXsX1F9AOto+9phlLT92u5u8TKCWuuqlPE2k8d+LpR/AQ) for my testing.

A few lingering TODO items:

- [X] Update `CHANGES.md`
- [x] Update the `Camera` spec for `getPickRay`
- [x] Fix specs that I broke

### DIFF
--- a/Source/Scene/Camera.js
+++ b/Source/Scene/Camera.js
@@ -2974,6 +2974,8 @@ function getPickRayOrthographic(camera, windowPosition, result) {
  * @returns {Ray} Returns the {@link Cartesian3} position and direction of the ray.
  */
 Camera.prototype.getPickRay = function (windowPosition, result) {
+  //TODO: remove this
+  console.log("At the top of getPickRay");
   //>>includeStart('debug', pragmas.debug);
   if (!defined(windowPosition)) {
     throw new DeveloperError("windowPosition is required.");
@@ -2983,9 +2985,14 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
   if (!defined(result)) {
     result = new Ray();
   }
-
+  //TODO: check if the display is 'none'
+  const scene = this._scene;
   const frustum = this.frustum;
-  if (
+  console.log("this._scene.canvas.style['display']");
+  console.log(scene.canvas.style["display"]);
+  if (scene.canvas.style["display"] === "none") {
+    return undefined;
+  } else if (
     defined(frustum.aspectRatio) &&
     defined(frustum.fov) &&
     defined(frustum.near)
--- a/Source/Scene/Camera.js
+++ b/Source/Scene/Camera.js
@@ -2971,8 +2971,7 @@ function getPickRayOrthographic(camera, windowPosition, result) {
  *
  * @param {Cartesian2} windowPosition The x and y coordinates of a pixel.
  * @param {Ray} [result] The object onto which to store the result.
- * @returns {Ray} Returns the {@link Cartesian3} position and direction of the ray.
- * If the Scene is not rendered or the pickRay cannot be determined, returns undefined.
+ * @returns {Ray|undefined} Returns the {@link Cartesian3} position and direction of the ray, or undefined if the pick ray cannot be determined.
  */
 Camera.prototype.getPickRay = function (windowPosition, result) {
   //>>includeStart('debug', pragmas.debug);
--- a/Source/Scene/Camera.js
+++ b/Source/Scene/Camera.js
@@ -2985,9 +2985,9 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
     result = new Ray();
   }
 
-  const scene = this._scene;
+  const canvas = this._scene.canvas;
   const frustum = this.frustum;
-  if (defined(scene.canvas.style) && scene.canvas.style["display"] === "none") {
+  if (canvas.clientWidth <= 0 || canvas.clientHeight <= 0) {
     return undefined;
   } else if (
     defined(frustum.aspectRatio) &&
@@ -2995,9 +2995,9 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
     defined(frustum.near)
   ) {
     return getPickRayPerspective(this, windowPosition, result);
+  } else {
+    return getPickRayOrthographic(this, windowPosition, result);
   }
-
-  return getPickRayOrthographic(this, windowPosition, result);
 };
 
 const scratchToCenter = new Cartesian3();
--- a/Source/Scene/Camera.js
+++ b/Source/Scene/Camera.js
@@ -2974,7 +2974,6 @@ function getPickRayOrthographic(camera, windowPosition, result) {
  * @returns {Ray} Returns the {@link Cartesian3} position and direction of the ray.
  */
 Camera.prototype.getPickRay = function (windowPosition, result) {
-  //TODO: remove this
   //>>includeStart('debug', pragmas.debug);
   if (!defined(windowPosition)) {
     throw new DeveloperError("windowPosition is required.");
@@ -2984,7 +2983,7 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
   if (!defined(result)) {
     result = new Ray();
   }
-  //TODO: check if the display is 'none'
+
   const scene = this._scene;
   const frustum = this.frustum;
   if (scene.canvas.style["display"] === "none") {
--- a/Source/Scene/Camera.js
+++ b/Source/Scene/Camera.js
@@ -2972,6 +2972,7 @@ function getPickRayOrthographic(camera, windowPosition, result) {
  * @param {Cartesian2} windowPosition The x and y coordinates of a pixel.
  * @param {Ray} [result] The object onto which to store the result.
  * @returns {Ray} Returns the {@link Cartesian3} position and direction of the ray.
+ * If the Scene is not rendered or the pickRay cannot be determined, returns undefined.
  */
 Camera.prototype.getPickRay = function (windowPosition, result) {
   //>>includeStart('debug', pragmas.debug);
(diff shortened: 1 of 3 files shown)


cesium-concierge:
    ```
    ```
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/camera-PickRay-exception/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

srothst1:
The following updates have been made to `Camera.prototype.getPickRay` in `Camera.js`:
- Added logic that checks if `scene.canvas.style["display"] === "none"`
- If scene.canvas.style["display"] === "none"`, return `undefined` given that the scene is not visible. 
- This seems more practical than calling `getPickRayPerspective` or `getPickRayOrthographic`, which yield runtime errors.

Source/Scene/Camera.js
```diff
@@ -2984,8 +2984,11 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
     result = new Ray();
   }
 
+  const scene = this._scene;
   const frustum = this.frustum;
-  if (
+  if (defined(scene.canvas.style) && scene.canvas.style["display"] === "none") {
```
ggetz, line 2989: Rather than putting a band-aid on the specific case in which this error arises, can we look a little deeper at what's causing the error and be a bit more wholistic on what we can catch up front?

I believe what's happening is that in both `getPickRayX` functions, we are `canvas.clientWidth` and `canvas.clientHeight` are returning `0` since the canvas is hidden, which is causing divide-by-zero issues. Checking if the height and width are greater than `0` will catch more than just if the `display: none` rule is applied.

Source/Scene/Camera.js
```diff
@@ -2984,8 +2984,11 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
...
+    return undefined;
```
ggetz, line 2990: The jsDoc comment for this function will need to be modified to include that `undefined` will be returned if the pickRay cannot be determined.

ggetz (commented):
Thanks @srothst1 ! Before you get too deep in the specs, I have a suggestion about the approach here.

Source/Scene/Camera.js
srothst1: Nice catch! I want to make sure that we keep our documentation consistent. I just added a commit that should address this issue.

Source/Scene/Camera.js
srothst1: Hi Gabby. Thank you very much for the feedback. I just added a commit that should address this shortcoming of my original approach. Specifically, the commit does the following:
- Modifies the logic in `getPickRay` to return `undefined` if either the canvas width or the canvas height is less than or equal to `0`.
- Adds `getPickRayOrthographic` to a newly created `else` statement. This should help clean up the code a little bit.

From my testing, these edits pass CLI and still work correctly with the sandcastle demo test that I shared above.

Source/Scene/Camera.js
```diff
@@ -2972,6 +2972,7 @@ function getPickRayOrthographic(camera, windowPosition, result) {
  * @param {Cartesian2} windowPosition The x and y coordinates of a pixel.
  * @param {Ray} [result] The object onto which to store the result.
  * @returns {Ray} Returns the {@link Cartesian3} position and direction of the ray.
+ * If the Scene is not rendered or the pickRay cannot be determined, returns undefined.
```
ggetz, line 2975: We can actually do this directly in the jsDoc and don't need to add it a sperate bullet:

```suggestion
 * @returns {Ray|undefined} Returns the {@link Cartesian3} position and direction of the ray, or undefined if the pick ray cannot be determined.
```

It's important to update the return type since we generate TypeScript definitions from our jsDoc comments.

Source/Scene/Camera.js
```diff
@@ -2984,16 +2985,19 @@ Camera.prototype.getPickRay = function (windowPosition, result) {
...
-  if (
+  if (canvas.clientWidth <= 0 || canvas.clientHeight <= 0) {
+    return undefined;
+  } else if (
     defined(frustum.aspectRatio) &&
     defined(frustum.fov) &&
     defined(frustum.near)
   ) {
     return getPickRayPerspective(this, windowPosition, result);
+  } else {
```
ggetz, line 2998: We favor returning from a function at the top-most level we can, and enforce this using the [`no-else-return`](https://eslint.org/docs/rules/no-else-return) eslint rule. That means that this should be written to omit the `else` as, since we returned in the other two conditions, it's not necessary. You can also factor out the `else if` statement, so that the pattern becomes

````js
if (condition) {
   return undefined;
}

if (secondCondition) {
   return getPickRayPerspective(this, windowPosition, result);
}

return getPickRayOrthographic(this, windowPosition, result);
````

Comment (srothst1) [reply Source/Scene/Camera.js]: Hi @ggetz. Thank you for the styling feedback. I just pushed two commits that should address this matter.

**Comment by srothst1** [comment]
Hi @ggetz. I am wondering if you have any feedback on the Travis CI build failures. When I run our specs using Jasmine on my local build, I seem to be getting various errors. These errors seem to occur even if I revert my branch back to [918c253](https://github.com/CesiumGS/cesium/pull/10139/commits/918c253b5d9677ede9d84703ffe3be9de99c02b7), which had no issues when it was committed.

File: Specs/Scene/CameraSpec.js [hunk 5]
```diff
@@ -3148,6 +3148,17 @@ describe("Scene/Camera", function () {
...
 
+  it("get pick ray returns undefined if the Scene is not fully rendered", function () {
+    const windowCoord = new Cartesian2(
+      scene.canvas.clientWidth / 2,
+      scene.canvas.clientHeight
+    );
+
+    scene.canvas.clientWidth = 0;
+    const ray = camera.getPickRay(windowCoord);
+    expect(ray).toEqual(undefined);
```
ggetz, line 3159:
```suggestion
    expect(ray).toBeUndefined();
```

This is knit-picky, but there's a specific [matcher](https://jasmine.github.io/api/2.7/matchers.html) for this case, you don't always need to use `toEqual`.

ggetz (commented):
Awesome, looking great @srothst1 ! I just have one small comment.

ggetz:
> I suspect that the error was related to our https://github.com/CesiumGS/cesium/pull/10139#discussion_r814799692.

Yes, eslint was reporting an error and failing CI. You can check for eslint errors locally before you commit by running `npm run eslint.

srothst1:
Hi @ggetz. Please let me know if there is anything else that I need to take care of before this can be merged.

ggetz:
Looks good, thanks @srothst1 !

mramato:
This is a breaking change and should probably be listed as such. `getPickRay` can now return `undefined`, which was not the case before.

srothst1:
@ggetz should I open up another PR that fixes `CHANGES.md`?

ggetz:
@mramato Even if those cases would have thrown a non-obvious exception before?

@srothst1 If so, can you please open a PR to update `CHANGES.md` so that your bullet is under "Breaking changes"?

srothst1:
Hi @ggetz. Yes, I will open up another PR 👍

mramato:
> Even if those cases would have thrown a non-obviously related exception before?

Without TypeScript, I would agree with you, but with TypeScript, yes. The below TypeScript compiles fine with `1.90` but will fail with `main` after this PR.

```ts
let result: Ray;
const camera = new Camera(new Scene({ canvas: canvasElement }));
result = camera.getPickRay(new Cartesian2(0, 0));
```

With the error

> Specs/TypeScript/index.ts:396:1 - error TS2322: Type 'Ray | undefined' is not assignable to type 'Ray'.

But more philosophically, changing a return type should always be documented as a breaking change because any code calling `getPickRay` must take into account that undefined is possible, and any code written against the previous documentation will not necessarily have done that.

PR #10138, opened 2022-02-23
Show glTF copyrights under the credits display

j9liu:
Closes #10127.

This PR causes the copyright in glTF files to be parsed and shown in the credits display. For example, a glTF with the following `asset` definition...
```
{
    "asset": {
        "version": "2.0",
        "generator": "collada2gltf@f356b99aef8868f74877c7ca545f2cd206b9d3b7",
        "copyright": "First Source, Second Source, Third Source"
    }
}
```
... will have its copyright appear in under "Data Attribution" as shown.

image

This change was done for both `Model` and `ModelExperimental`.

[...]

PR #10141, opened 2022-02-24
Sort credits by frequency of occurences

### LINKED ISSUE #10128: Sort credits based on the number of occurrences
In the terrain/imagery code (and soon 3D Tiles https://github.com/CesiumGS/cesium/issues/10127) we aggregate credits for tiles that are in view. However we don't sort these in any particular order. Instead we should be keeping a count of the number of occurrences of each credit and then sort the list so that more frequent credits appear sooner.

j9liu:
Closes #10128.

This PR makes `CreditsDisplay` sort its credits by their number of occurrences, with more frequently occurring credits appearing first. I added a tileset with duplicate copyrights (a modified version of the one added in #10138) to test this, which when put in a [Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=xVRRb9owEP4rVp6CxBwYo6UtRWtB2x6QVgHtXvJinCPx6tiRfQHRaf99dkImmnYPq5Cal/h8d99999k+rpVFshWwA0OuiYIdmYIVZU4fqr0wDnhlT7VCJhSYOOhcxYpXeZaDApdW59PKdM7GjUKCBXyOW/8Gs1XtDH/FipDSyEsSB9GyAG6jGUMWPYuLvkrceAqg8IfAbAEFMIRkqou9EWmGNkrl6kt0qEh/Wq3ioOuhE1iX6TLTu1tdqkSo9EHLModLgqaEWP3uVIQr6rQwIhcotmApS5LwgNY5asmFIZi5VqnAMvG9f+jTQf/ibHTVCmHYRPTo2cVodP6pFeE8jSTMoFsxNaAbo/MFS9zahq1a3RbyMa2saJ3eN2C+1zuBPFswlULotfBfr9usHPPh+V+r/3FYLz1uo6NxMPs7x0lYoJiBCjel4ii0ImGHVEdXK8dZDoZRqfXjDR6Ydz0tj3bQeMlUwplFCV7dldZyzcxtiaiVu2Y3UpLqrN25kabKqYvcF4UTfg4brGq9sVQrKNdb8JBhf9j7h/u+aJz/QXPhL/aJeVaYpyM617t30XOmd+oNRN9D0RdUg24wtriXMGme3meRF9qgn4IhpRFCXkg33Wy0Lvmje4Xc2hqckHF0nDpOxJaI5PqVIU24ZNY6z6aUcimeXMeTceTiX6RKXY2K71swku19WNafzOtNSuk4cubrmViL3EL+Aw) results in the following:

image

### DIFF
--- a/Source/Scene/CreditDisplay.js
+++ b/Source/Scene/CreditDisplay.js
@@ -66,8 +66,16 @@ function createCreditElement(element, elementWrapperTagName) {
 function displayCredits(container, credits, delimiter, elementWrapperTagName) {
   const childNodes = container.childNodes;
   let domIndex = -1;
+
+  // Sort the credits such that more frequent credits appear first
+  const creditsSorted = credits.slice();
+  creditsSorted.sort(function (credit1, credit2) {
+    return credit2[1] - credit1[1];
+  });
+
   for (let creditIndex = 0; creditIndex < credits.length; ++creditIndex) {
-    const credit = credits[creditIndex];
+    const creditInfo = creditsSorted[creditIndex];
+    const credit = creditInfo[0];
     if (defined(credit)) {
       domIndex = creditIndex;
       if (defined(delimiter)) {
@@ -341,10 +349,15 @@ function CreditDisplay(container, delimiter, viewport) {
   this._cesiumCredit = cesiumCredit;
   this._previousCesiumCredit = undefined;
   this._currentCesiumCredit = cesiumCredit;
+
+  // Each AssociativeArray contains both the credit and the number of times
+  // it has been added to the display. This is used to sort the credits by
+  // frequency of appearance when they are later displayed.
   this._currentFrameCredits = {
     screenCredits: new AssociativeArray(),
     lightboxCredits: new AssociativeArray(),
   };
+
   this._defaultCredit = undefined;
 
   this.viewport = viewport;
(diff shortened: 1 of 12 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/sort-credits-by-frequency/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/CreditDisplay.js
```diff
@@ -66,8 +66,16 @@ function createCreditElement(element, elementWrapperTagName) {
...
+  // Sort the credits such that more frequent credits appear first
+  const creditsSorted = credits.slice();
+  creditsSorted.sort(function (credit1, credit2) {
+    return credit2[1] - credit1[1];
+  });
+
   for (let creditIndex = 0; creditIndex < credits.length; ++creditIndex) {
-    const credit = credits[creditIndex];
+    const creditInfo = creditsSorted[creditIndex];
+    const credit = creditInfo[0];
```
lilleyse, line 78: I like the tuple concept but it can be easy to lose track of which array element does what. Maybe a struct would be better? Something like

```js
function CreditDisplayElement(credit) {
    this.credit = credit;
    this.count = 0;
}
```

Source/Scene/CreditDisplay.js
```diff
@@ -376,10 +399,18 @@ CreditDisplay.prototype.addCredit = function (credit) {
...
   if (!credit.showOnScreen) {
-    this._currentFrameCredits.lightboxCredits.set(credit.id, credit);
+    credits = this._currentFrameCredits.lightboxCredits;
+  } else {
+    credits = this._currentFrameCredits.screenCredits;
+  }
+
+  if (credits.contains(credit.id)) {
+    const creditCount = credits.get(credit.id).count;
+    credits.set(credit.id, new CreditDisplayElement(credit, creditCount + 1));
```
lilleyse, line 411: Instead of creating a new CreditDisplayElement, could the `count` of the existing CreditDisplayElement be incremented?

Source/Scene/CreditDisplay.js
```diff
@@ -376,10 +399,18 @@ CreditDisplay.prototype.addCredit = function (credit) {
...
   } else {
-    this._currentFrameCredits.screenCredits.set(credit.id, credit);
+    credits.set(credit.id, new CreditDisplayElement(credit));
   }
```
lilleyse, line 414: I should have noticed this sooner. Both the tuple approach and struct approach will allocate a new object for each unique credit that's added. We should avoid that.

One fix is to have two separate `AssociateArray`, one for the credits and other for the counts, as discussed offline. Another is to recycle `CreditDisplayElement` objects using a `ManagedArray`.

lilleyse:
Thanks @j9liu!

Issue #9578, opened 2021-06-01
computeIcrfToFixedMatrix test suite has intermittent failures


Samulus:
The following tests have intermittently failed on me while running all tests using Firefox 88.0.1 on Ubuntu Linux (20.04.2 LTS):

```
Core/Transforms computeIcrfToFixedMatrix works with data from STK Components
Core/Transforms computeIcrfToFixedMatrix works with hard-coded data
Core/Transforms computeIcrfToFixedMatrix works over day boundary
Core/Transforms computeIcrfToFixedMatrix works over day boundary backwards
Core/Transforms computeIcrfToFixedMatrix works with position rotation
Core/Transforms computeIcrfToFixedMatrix works without EOP data loaded
Core/Transforms computeIcrfToFixedMatrix throws a RuntimeError when asked to compute with invalid EOP data
Core/Transforms computeIcrfToFixedMatrix throws a RuntimeError when asked to compute with a missing EOP data file
```

sanjeetsuhag:
Facing this issue right now with the 1.88.0 release. They always pass when only this suite is run, however, when all the unit tests are run, it causes intermittent failures.

Issue #3967, opened 2016-05-26
Upgrade when.js to latest


mmacaula:
When.js is currently at v1.7.1, latest available (as of now) is 3.7.7.  Looking at the [changes](https://github.com/cujojs/when/blob/master/CHANGES.md) betwen latest and then, there are plenty of 'massive' performance improvements.  Might be a good idea to upgrade to take advantage.  There are two 'breaking' versions between there so would require looking through all promise code to bring into a+ compliance.  

As an aside, you may want to consider putting when.js as a dependency in package.json (though there are probably good reasons for just committing it).

hpinkos:
Thanks @mmacaula!  We've been discussing changing our promise library recently.  I think we're going to look into replacing `when` with `bluebird`.  @mramato can elaborate, but I think there's a feature in the version of when that we're using that makes it a little more work to upgrade, but we'd like to take care of this soon.

mmacaula:
Sounds great!  Yeah I think i've seen that 'feature' around the code, I suspect it's the passing of `onFulfilled` handlers to things like `when()`. Bluebird definitely seems to be the fastest out there now, but when.js is definitely not far behind (or far enough to matter maybe).  

Feel free to close this issue if it's easier to manage.

shunter:
Specifically the feature in the old version of when.js that we rely on (for better or worse) is that calling .then on resolved promises executes immediately.  Promises A+ requires that handlers **must** be invoked on the next turn of the loop, which is an unfortunate choice from a performance perspective.  More detail here: http://thanpol.as/javascript/promises-a-performance-hits-you-should-be-aware-of/

In practical terms in Cesium, upgrading without addressing this change in behavior would add several frames of extra latency on every terrain and imagery tile load.

shunter:
It looks like Bluebird has features that would help with dealing with this: http://bluebirdjs.com/docs/api/synchronous-inspection.html

mramato:
@shunter I was just looking at the synchronous inspection stuff yesterday.  I may try and do this at the code sprint next week.

pjcozzi:
@mramato would we upgrade to Bluebird or use ES6 promises?  Could we wait that long for ES6 to be widely used?

mramato:
Native ES6 Promises are fairly limited and it still makes a lot of sense for us to use Bluebird in Cesium. (For example, there is no `finally` function, no map, each, etc..)

thw0rted:
[...]

PR #8525, opened 2020-01-12
Replace when with native Promises

mramato:
Me and @kring were discussing this last week and I had a branch that had it mostly done, so I figured I'd start fresh and get a PR open.

1. Replace all use of `when` with native promises.
2. Add `Core/defer.js` to replace `when.defer`.
3. Remove `when.js`

- [ ] Fix failing unit tests
- [ ] Polyfill web workers to support IE11
- [ ] Update CHANGES.md to state IE11 support requires a Promise polyfill (and recommend one).
- [ ] Decide if we should deprecate and put back `when.js`, but it's basically impossible for this not to be a breaking change, so why bother.
- [ ] Any performance implications we need to worry about?
- [x] Do full release testing to be sure of no surprises.

PR #10149, opened 2022-02-28
Replace when.js with native promises

ggetz:
Replaces https://github.com/CesiumGS/cesium/pull/8525
Fixes https://github.com/CesiumGS/cesium/issues/3967
Fixes https://github.com/CesiumGS/cesium/issues/9578

This PR removes `when.js` in favor of native promises. Besides removing the files related to when itself, this generally follows the migration guide as laid out in [the migration guide](https://community.cesium.com/t/cesiumjs-is-switching-from-when-js-to-native-promises-which-will-be-a-breaking-change-in-1-92/17213), swapping out `when.defer` with a new `defer` function, switching `otherwise` for `catch`, and `always` for `finally`. 

There were a few areas in `Source` where a function was assumed to be executed synchronously when a function was resolved. Native Promises however, by spec, will resolve at the end of a frame. There were also some strange handling of promise rejection in imagery providers that I cleaned up a bit in order to make them testable.

  * In `LabelCollection` and `EntityCluster`, order of execution adjustment where made.
  * Added `TextureAtlas.addImageSync` to support the order of execution fix in `LabelCollection`
  * `ArcGisMapServerImageryProvider.readyPromise` will not reject if there is a failure unless the request cannot be retried.
  * `SingleTileImageryProvider.readyPromise` will not reject if there is a failure unless the request cannot be retried.

The majority of the changes lie in the Specs, where unresolved promises weren’t being awaited before finishing executions, and where resolved promises are assumed to be synchronous all over the place and needed a good amount of fixes. Another issue which came up  was calling `Promise.reject` in the body of a spec can cause node to halt execution when running via the command line.

Finally, I did need to tweak `eslint` in order to turn off the rule which disallows promises.

~Opening this a bit early. Still to do:~
- [x] Merge master
- [x] Fix remaining spec failures, travis-only failures
- [x] Open issues for excluded specs, and for refactoring out use of `defer` in general
- [x] Add documentation for defer
- [x] Update `CHANGES.md` - Add note about `ArcGisImageServerProvider` rejecting ready promise if there is a failure and retry is `false`.
- [x] Check `build-ts` warnings


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/when-to-native/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

chris-cooper:
It will be great to see this at last!  Did you consider a major version update due to the API change?  https://semver.org/

ggetz:
Thanks @chris-cooper, historically cesiumjs has chosen not to use semver due to frequent breaking changes especially early in development. However, we are presently considering moving to semver to make updates like this easier. We're in the process of confirming this decision since it will likely have impacts among government users. So we're thinking about it, but I can't promise a version increment for this next release.

ggetz:
I know @kring mentioned that `defer` is not the preferred way to go, and to use the `Promise` constructor instead as its better practice for handling errors. I think for the sake of limiting the scope of this PR, we leave it in. However, I'm assuming we should at least add a warning to the `defer` docs explaining that's its preferable to use `Promise` directly instead?

mramato:
> I'm assuming we should at least add a warning to the `defer` docs explaining that's its preferable to use `Promise` directly instead?

This is private, right?  So I assume you mean the internal documentation?

I agree with @kring that all of Cesium's `defer` usage would be better served via `new Promise`. I'm also okay with making that a second pass. It will be a lot easier to do than the effort this took, either way.  We should avoid adding any new `defer` calls in future code.

ggetz:
@mramato `when.defer` was part of the public API, so I assumed we should make `defer` public as well, at least for this initial change. If you recommend making a clean break, I can update the migration guide and suggest any usage of `when.defer` should instead be replaced with a `new Promise`.

mramato:
@ggetz I would definitely recommend a clean break. Anyone that was using `defer` is going to have to update their code regardless.

ggetz:
The docs, `CHANGES.md`, and the migration guide on the forum have been updated accordingly. 

I'm still fixes a handful of specs, but this is almost complete and should be ready for review. @sanjeetsuhag Could you please do a pass on this?

sanjeetsuhag:
@ggetz I went through all the Sandcastles and it seems like the only problematic one is [CZML Custom Properties](http://localhost:8080/Apps/Sandcastle/index.html?src=CZML%20Custom%20Properties.html&label=All).

ggetz:
Thanks @sanjeetsuhag, fixed! That test just needed to await loading the CZML. 

The tests are also in order now. I had to exclude a handful, mostly 3D tiles related, specs as there was some questionable error handling that led to errors being thrown that cannot be sensibly caught. I'm opening an issue for these excluded tests and will link it back into the specs.

Specs/Scene/ModelSpec.js
```diff
@@ -2824,20 +2824,25 @@ describe(
...
           expect(scene).toRenderAndCall(function (rgba) {
             expect(rgba).not.toEqual([0, 0, 0, 255]);
             loadedColor = rgba;
           });
 
           model = m;
+
+          // Render at least once to initialize
+          scene.renderForSpecs();
+          return Promise.resolve();
```
mramato, line 2837: When already inside a `then` block, there is no need to return `Promise.resolve`, you don't need the return statement at all, the result will just be a Promise that resolves to udnefined.

ggetz:
I wasn't able to duplicate locally, but I believed I've handled the intermittent failure that was happening in CI. @sanjeetsuhag do you have any additional feedback?

sanjeetsuhag:
@ggetz I looked at the latest commits, and I have no comments on the code changes. I can only speak to the excluded `ModelExperimental` specs and that https://github.com/CesiumGS/cesium/issues/10178 should be a high priority fix before the release: lots of important tests in there.

ggetz:
Thanks @sanjeetsuhag. I do agree that its high priority, and labeled it accordingly. There's also only two tests in `ModelExperimentalSpec` which have been excluded because of the error handling, not the whole suite.

Apps/Sandcastle/CesiumSandcastle.js
```diff
@@ -826,7 +824,7 @@ require({
             const html = defined(htmlFile) ? htmlFile.content : defaultHtml; // Use the default html for old gists
             applyLoadedDemo(code, html);
           })
-          .otherwise(function (error) {
+          .catch(function (error) {
```
mramato, line 827: I doubt this functionality is ever used (unless we have some die hard Cesium folks still hanging on to Sandcastle gists) but since dojoscript is going to return a dojo when, `otherwise` is probably the right thing to have here. The other option is to wrap the `dojoscript.get` call in a `Promise.resolve` so it turns a thenable into a real promis.

Source/.eslintrc.json
```diff
@@ -14,6 +14,7 @@
         "es/no-modules": "off",
         "es/no-typed-arrays": "off",
         "es/no-block-scoped-variables": "off",
-        "es/no-template-literals": "off"
+        "es/no-template-literals": "off",
+        "es/no-promise": "off"
```
mramato, line 18: This can easily be a follow-up PR, but given all of the changes in the past year, we should revisit Cesium's default eslintrc configuration and the shared eslint-cesium plugin we also maintain in this repository. A lof of these settings are out of date ad unneeded (like the es/restrict-to-es5).  It's not a ton of work, but does affect JS code cesium-wide, so I would just write an issue and we can easily bang it out as soon as we get a chance.

mramato:
Have we ran any performance benchmarks against main to ensure that we haven't introduced a perf regression in any of the most common areas? (Like CWT + 3DTiles loading) or simple.czml animating on an empty globe.

Not sure if we have any "canned" perf tests ready to go, perhaps @lilleyse might have some ideas about what we have done for 3D Tiles testing.

Source/DataSources/EntityCluster.js
```diff
@@ -659,7 +659,10 @@ function createGetEntity(
 
     entityIndices[entityIndexProperty] = index;
 
-    this._clusterDirty = true;
+    const that = this;
+    Promise.resolve(entityItem.ready).then(function () {
```
mramato, line 663: Is this code intentional? EntityCluster as a whole is confusingly implemented, but I tried to figure out what entityItem was and it seems to be different depending on the item (like a billboard or a label)  and while sometimes `entityItem.ready` will be something that exists, it's never a promise, only a boolean. So why is the Promise.resolve here?

Source/Scene/Globe.js
```diff
@@ -937,7 +936,9 @@ Globe.prototype.beginFrame = function (frameState) {
     const oceanNormalMapUrl = oceanNormalMapResource.url;
     if (defined(oceanNormalMapUrl)) {
       const that = this;
-      when(oceanNormalMapResource.fetchImage(), function (image) {
```
mramato, line 940: Doesn't `fetchImage` always return a Promise?  Why do we need the resolve here?

Source/Scene/LabelCollection.js
```diff
@@ -908,13 +909,21 @@ LabelCollection.prototype.update = function (frameState) {
...
   }
 
+  const that = this;
   if (!defined(this._backgroundTextureAtlas)) {
     this._backgroundTextureAtlas = new TextureAtlas({
       context: context,
       initialSize: whitePixelSize,
     });
     backgroundBillboardCollection.textureAtlas = this._backgroundTextureAtlas;
-    addWhitePixelCanvas(this._backgroundTextureAtlas, this);
```
mramato, line 917: This looks like a bugfix in addition to Promise update? Is that true? If so, what bug does it fix and should we include that in CHANGES? (even if we have no issue number for it).

I noticed a few places where there are little changes like this that appear to be bug fixes of some kind. If we just figured out the code is wrong but don't know how/if a bug manifested from it, that's fine too, nothing to report. But I just wanted to point it out in case.

Source/Scene/TextureAtlas.js
```diff
@@ -353,6 +352,49 @@ function addImage(textureAtlas, image, index) {
   textureAtlas._guid = createGuid();
 }
 
+function getIndex(atlas, image) {
```
mramato, line 355: As a general rule, stuff like this should be called out in the PR description.  I assume by looking at it we had to add some branch new code in order to provide a path for synchronous operations that would otherwise be async in our new "native" Promise world, but in a PR this big; it would be easy for a reviewer to miss.  Also, if someone needs to use revision history to track where/why this code was added, it won't be obvious to them by reading the PR description (which it should be).

Even just a single bullet like "Added TextureAtlas.addImageSync to xxxx" would be extremely valuable to future devs.

This is _especially_ true given the terseness of many of the commit messages in this PR (or perhaps we are squashing) but either way, a paper trail is important.

Source/Scene/Vector3DTilePolygons.js
```diff
@@ -304,7 +304,7 @@ function createPrimitive(polygons) {
       return;
     }
 
-    when(verticesPromise, function (result) {
+    Promise.resolve(verticesPromise).then(function (result) {
```
mramato, line 307: Isn't this always a promise at this point? is the `resolve` wrapper needed?

Source/Scene/Vector3DTilePolylines.js
```diff
@@ -228,7 +228,7 @@ function createVertexArray(polylines, context) {
       return;
     }
 
-    when(verticesPromise)
+    Promise.resolve(verticesPromise)
```
mramato, line 231: SAme comment as Vector3DTilePolygons.js, is this needed?

Source/Widgets/Viewer/Viewer.js
```diff
@@ -179,8 +179,7 @@ function pickImageryLayerFeature(viewer, windowPosition) {
     description: "Loading feature information...",
   });
 
-  when(
-    imageryLayerFeaturePromise,
+  Promise.resolve(imageryLayerFeaturePromise).then(
```
mramato, line 182: Another spot that I think is always a Promise.

mramato:
Two big things that jumped out at me:

* Holy smokes do we abuse defer!
* Moving to async/await (Which can happen organically/gradually) would be a huge improvement for readability and verbosity for our code base.

mramato:
Just noticed this: 

Screenshot_20220310_192919

Is that because of test fixes or something else? (This is more out of curiosity than anything else).

mramato:
Okay @ggetz I'll admit I kind of lost steam when I got to the specs, so I didn't review them thoroughly by any means. But this PR looks great and those minor comments were all I had. This is a major stepping stone for CesiumJS and an awesome way to start a true modernization effort.

Source/.eslintrc.json
ggetz: Definitely. We already planned on addressing eslint configuration in the near future, but I'll write up an issue to be clear about it.

Source/Scene/LabelCollection.js
ggetz: These were to address cases where the code assumed a promise was executing synchronously. When transitioning to native promises, even in cases where the promise immediately resolved, it wouldn't do so until the end of the frame. So the timing of the rest of the function is thrown off. I don't think these fix any existing bugs, just keep behavior working as before.

Source/Scene/TextureAtlas.js
ggetz: I planned on squashing commits with a more useful commit message. But understood, I'll call out these changes in the PR explicitly.

ggetz:
> Have we ran any performance benchmarks against main to ensure that we haven't introduced a perf regression in any of the most common areas? (Like CWT + 3DTiles loading) or simple.czml animating on an empty globe.

I've run several 3D Tiles and CWT examples including various photogrammetry, point clouds, and OSM buildings. I'm seeing similar loading times using the Chome profiler on main and this branch. Likewise with `simple.czml` I'm seeing comparable framerates. @lilleyse If we have more exact tests for performance here, please let me know and I'd be happy to compare.

lilleyse:
I usually just record how long it takes for tiles to load for a given view. Sometimes I'll also set up a flight path. Here are two of my go-to sandcastles:

Globe|3D Tiles
--|--
terrain[Sandcastle](http://localhost:8080/Build/Apps/Sandcastle/index.html#c=hVRhT+Q2EP0r1n65oC7eOIkdm9tDbemeikQ5xK5aVYpUmWQWrHptZDtL6Yn/XifZLCzQuy9RPDPP782zx7U1PqCtggdw6BMy8IDOwKt2g3/vY8mHul+eWROkMuA+TNHXyiAUwLkYuHJ2q5qIPRlxtQMZ4A/rdLMaapKjyjwdfaxMZeqeztdgILINtLhffhyTtdyAkzHbh/Gw7MFDoIGb9nZ5Zx8+u5jyV+CWEJFNRATXwkDTg7CH0HWR9IIb8EEZGZQ1UeyLRs+kC/FPmjw5zlhGKWeY54SVWUl4PkXHRZGXghY4LQpGWFYWU5SXRZalDJdFnmeEC3o07TisUxB9esvxK8hGmdsrFeq7a6t1QnHJGC04K6goeCrKyJPiVJBSkIxyWpKoI49MDGecUCFyUTIisnzHBKZZOWn82rrNs/e/yeDUPwU+/2VxuTpf/bn3fSlNU0sfNGDZNCtr9Y10P7chWJNUk2WIFlSTKVq3pu7EJ0foK0Idz87JtX5c2cHHAye/YWTOSMoxTUnKOpdoZ2QuKOMp5rHBgmWMdEbyQpAcZ0KwknGSlUN/B15+z0ouUsKKlDOa8TwXvZVZWnJBKClJHtloubMyS6mIdQVlBc9Gqpdm/q+Xu9qmdTtNOU5HvPRR0eedefsdFgdhfHF+ufjp+q/LL5eLHa62m3sNAU4OjB9yXbafle5sVmrTzcs9uE6iNDVgYx+S7mxf1jrY2C0sttG3C+VDnBa3n6N768N1bDTOW7wBBzXJe+wIqTVKBuyttjeAg9LgL6xsoDmoG9k7F7+n87l69DGWj8Dj52a/ARim3UfcfosZImkap+cNysb7ru1tvOLdpieomqAfXm90KO4dCw/kP42/T2O0j/Sr+JlMJ3MfHjWcjnU/qs29dQG1TicYzwLEI4/vo5/dtPXfEHDt9wrms5fQeaO2SDWfqsmrFzg2UWvpfcysW62X6l+oJqfzWax/A9W2H5YvW3BaPnZld+T0YghijOezuHwfGYY34tXO/wE)|3dtiles[Sandcastle](http://localhost:8080/Build/Apps/Sandcastle/index.html#c=hVJda9swFP0rwi+VWSe72yijc8NG14dCYGMJe/KLIl0nYtK9RpLtpaP/fbKdlCYdzC/mSufrHlsRhsh6AwN4dssQBnYHwXRO/JzO+IWaxjvCKA2Cv8g/1TjjRVCAILaWNiDCjoYk0EgbICFqVJNyNBYCxHRzwmm9cSaaHoKQWvMXtvPr/df1TOR/amSs8/bmCHgg/AGBOq9ANJ7cl5BgD5p/KD9eX+eXNT7l+RTgYC2c/G1c51bKA+CqlQruvadx23einJCHaI9Ebk38wJtFbMoeovRxbRycyHqQev89BTABRNwB8qZDFQ0hz9mU+pmXrAaDmgbRgm/IO4kpPNLAR5On07gGUzHSTvsvSWrQY0X3PWBcmhBTe/6V09w1oP6/2xGsOy9HiYQ+8t6ebHoOXEGadUj4Z2rBrsqynEqc0WRBWNryOhtFblidsTfnAoeVs8usCnFvYTGSx+ezcS35OH5sLkQRwbVWRgjFplO/UjEqhHkBxqriJbXSpmdG39bZ2b+a7JWVIaSbprN2ZR6hzhZVkfCvqDZVbXD7rQdv5X6E7a4Wy/lQCFEVafw3MxLZjfRnyn8B)

PR #10150, opened 2022-02-28
Add explicit content metadata

### LINKED ISSUE #10105: Handle content metadata for explicit tiling
Another spec change (schema TBD, @lilleyse when ready could you link it here?) is to allow metadata within a content (likely `content.metadata`). This needs to be handled in the implementation.

This will be somewhat similar to how `TileMetadata` works. However, since there are a large number of `Xxx3DTileContent`, it may be better to have `Cesium3DTile` create the object and set `content.metadata = new ContentMetadata`.

This involves:

- [x] Create a `ContentMetadata` class. This will be very similar to `TileMetadata`. 
- [x] in `Cesium3DTile.makeContent()`, parse the content metadata extension (see where `content.groupMetadata` is initialized) and set `content.metadata`. This way, you don't have to add logic to every single type of content
- [x] Update `Cesium3DTileFeature.getPropertyInherited()` to look for `content.metadata`. This should go between the batch table and the tile metadata.

Some of these steps will be similar to #10104, as they both will populate the `content.metadata` field with either a `ContentMetadata` or an `ImplicitContentMetadata`. The two are mutually exclusive by the spec, though in the implementation for tile metadata I think I did this by not checking and just making the implicit one overwrite the explicit one.

j9liu:
Closes #10105.

This PR includes the following changes:

- Add `ContentMetadata` and specs (similar to `TileMetadata`)
- Update `Cesium3DTile.makeContent()` and `createInnerContents()` in `Multiple3DTileContent` to parse content metadata
  - **Question**: should the way that content metadata is parsed be put in its own function, similar to `findGroupMetadata`?
- Update `Cesium3DTileFeature.getPropertyInherited()`
- Add sample data and tests to `Cesium3DTilesetSpec`

cc @ptrgags for review


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags (commented):
@j9liu looks pretty good, my only feedback is to do that refactor you were asking about

j9liu:
@ptrgags - uploaded some changes. I had to nest the `findContentMetadata` function in an `if` statement because it was causing problems for tilesets with implicit content. This is because of the change I introduced to `Implicit3DTileContent`, which prevents the metadata field from being assigned.

ptrgags:
@j9liu good catch on the `if` block. This looks solid, I'm merging!

PR #10153, opened 2022-03-01
Fold tileset-level `3DTILES_metadata` extension into 3D Tiles 1.1

j9liu:
This PR incorporates the tileset-level metadata extension into 3D Tiles 1.1, with the following changes

- Update `Cesium3DTilesetMetadata` to accept either the 1.1 tileset JSON or the original extension object.
  - **Question:** Maybe this can be handled more cleanly, because the main difference is that 1.1 uses the `metadata` field instead of `tileset`?
- Change parsing of metadata extension in `Cesium3DTileset`
- Create 1.0 and 1.1 versions of relevant sample data

cc @ptrgags for review


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Data/Cesium3DTiles/Metadata/AllMetadataTypes/1.1/tileset.json
```diff
@@ -0,0 +1,336 @@
+{
+  "asset": {
+    "version": "1.0",
```
ptrgags, line 3: @j9liu just to confirm, this was intentional since we'll be making the change to allow `asset.version: "1.1"` in a future PR?

CHANGES.md
```diff
@@ -19,6 +19,9 @@
   - `3DTILES_metadata`:
     - Added support for explicit content metadata. [#10150](https://github.com/CesiumGS/cesium/pull/10150)
 
+- Updated extensions to match the latest 3D Tiles 1.1 schema. See the [latest schema PR](https://github.com/CesiumGS/3d-tiles/pull/634):
+  - Folded the tileset-level metadata extension into the tileset schema. [#10153](https://github.com/CesiumGS/cesium/pull/10153)
```
ptrgags, line 23: In the change log, I would be a little more explicit about the change, "folding" might not be obvious. Maybe

"tileset metadata is now stored in `tilesetJson.metadata` rather than the `3DTILES_metadata` extension"

Source/Scene/Cesium3DTileset.js
```diff
@@ -1016,6 +1016,10 @@ function Cesium3DTileset(options) {
     .then(function (tilesetJson) {
       // This needs to be called before loadTileset() so tile metadata
       // can be initialized synchronously in the Cesium3DTile constructor
+      const metadataJson = tilesetJson.metadata;
+      if (!defined(metadataJson)) {
+        return processMetadataExtensionLegacy(that, tilesetJson);
+      }
       return processMetadataExtension(that, tilesetJson);
```
ptrgags, line 1023: This is super nitpicky, but I would invert the logic of this if statement:

```
if (defined(metadataJson)) {
  // process metadata extension
} else {
  // process legacy extension
}
```

Just for consistency with metadata parsing elsewhere (parse new stuff formats, fall back to legacy) and readability (no negative)

ptrgags (commented):
@j9liu started reviewing, though it sounds like you had further changes, so I'll continue after you push them.

Specs/Data/Cesium3DTiles/Metadata/AllMetadataTypes/1.1/tileset.json
j9liu: yeah, I figured it would be better to change all of these in one sweep during that PR

Source/Scene/Cesium3DTilesetMetadata.js
```diff
@@ -15,55 +16,74 @@ import TilesetMetadata from "./TilesetMetadata.js";
  * </p>
  *
  * @param {Object} options Object with the following properties:
- * @param {String} options.extension The extension JSON object.
+ * @param {Object} [options.tilesetJson] The tileset JSON object. If this is undefined, then options.extension must be defined.
+ * @param {Object} [options.extension] The JSON object, used for backwards compatibility with the 3DTILES_metadata extension.
```
ptrgags, line 20: Can these parameters be merged if at this level it's essentially the same schema? maybe call it `metadataJson`, then you either pass `metadataJson: tilesetJson` for 1.1 or `metadataJson: tilesetJson.extensions.3DTILES_metadata`

Source/Scene/Cesium3DTileset.js
```diff
@@ -2050,6 +2049,44 @@ function processMetadataExtension(tileset, tilesetJson) {
...
+/**
+ * If tileset metadata is present, initialize the {@link Cesium3DTilesetMetadata}
+ * instance. This is asynchronous since metadata schemas may be external.
+ *
+ * @param {Cesium3DTileset} tileset The tileset
+ * @param {Object} tilesetJson The tileset JSON
+ * @return {Promise<Object>} A promise that resolves to tilesetJson for chaining.
+ * @private
+ */
+function processMetadataExtension(tileset, tilesetJson) {
```
ptrgags, line 2061: After looking at the change I suggest in `Cesium3DTilesetMetadata`, see if anything else can be simplified here between the two functions. It looks like the steps are close to identical, it's mainly the input that differs

Specs/Scene/Cesium3DTileFeatureSpec.js
```diff
@@ -85,9 +85,9 @@ describe(
       });
     });
 
-    describe("3DTILES_metadata", function () {
+    describe("metadata", function () {
```
ptrgags, line 88: Hm, this test structure feels like it's nested a little too much.

image


Maybe the organization should be like this:

* `3D Tiles 1.1 metadata` 
* `3DTILES_metadata` (as a sibling, not a child)
* `3DTILES_implicit_tiling metadata` (explicitly add "metadata" to make it clear)
  * subtree metadata
  * implicit content metadata

ptrgags: Essentially just close the `metadata` describe() block after the two tests

ptrgags (commented):
@j9liu just a few additional comments

CHANGES.md
```diff
@@ -5,6 +5,7 @@
 ##### Additions :tada:
 
 - Updated 3D Tiles Next extensions to match the latest specification. See the [latest specification PR](https://github.com/CesiumGS/3d-tiles/pull/630):
+
```
ptrgags, line 8: I would omit these newlines, sometimes that can confuse markdown formatters sometimes.

Source/Scene/Cesium3DTilesetMetadata.js
```diff
@@ -15,55 +15,62 @@ import TilesetMetadata from "./TilesetMetadata.js";
  * </p>
  *
  * @param {Object} options Object with the following properties:
- * @param {String} options.extension The extension JSON object.
+ * @param {Object} options.tilesetJson The tileset JSON object.
```
ptrgags, line 18: I would stick with the `metadataJson` instead. While a tileset JSON may be a valid object with metadata, if you use the `3DTILES_metadata` extension, it's not true.

Also please document this parameter more to explain the two cases that are valid, thanks!

ptrgags (commented):
@j9liu this looks close, just a couple more things.

ptrgags (approved):
@j9liu looks good, just had one tiny documentation tweak. I'll commit that and merge!

PR #10162, opened 2022-03-04
Handle arrays of vectors correctly for JSON metadata

ptrgags:
Back in #10116 I had added support for arrays of vectors, but I had incorrectly implemented it for JSON properties. Arrays of vectors are represented as nested JSON arrays:

```
properties:
  arrayOfVectors: [
    [1, 2, 3],
    [4, 5, 6]
  ]
```

Unlike property tables where everything is stored as one big flat array:

```
properties:
  arrayOfVectors: [1, 2, 3, 4, 5, 6]
```

This required some modifications to `MetadataClassProperty` and `MetadataEntity`. 

One nice side-effect is I made a more general "transformInPlace()" function that will be reusable when I get to transforming values with offset/scale.

@lilleyse could your review?

### DIFF
--- a/Source/Scene/MetadataClassProperty.js
+++ b/Source/Scene/MetadataClassProperty.js
@@ -567,16 +567,10 @@ MetadataClassProperty.prototype.normalize = function (value) {
   }
 
   const valueType = this._valueType;
-  const valueTransform = function (x) {
+  const normalizeFunction = function (x) {
     return MetadataComponentType.normalize(x, valueType);
   };
-
-  if (Array.isArray(value)) {
-    transformValuesInPlace(value, valueTransform);
-    return value;
-  }
-
-  return valueTransform(value);
+  return transformInPlace(value, normalizeFunction);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/MetadataClassProperty.js
```diff
@@ -562,7 +562,15 @@ function parseType(property, enums) {
...
-  return normalize(this, value, MetadataComponentType.normalize);
+  if (!this._normalized) {
+    return value;
+  }
+
+  const valueType = this._valueType;
+  const normalizeFunction = function (x) {
+    return MetadataComponentType.normalize(x, valueType);
+  };
+  return transformInPlace(value, normalizeFunction);
```
lilleyse, line 573: Is it possible to avoid creating a closure like this? The "function pointer" approach from before might be more performant.

lilleyse (commented):
Looks good, just one comment about performance

Source/Scene/MetadataClassProperty.js
lilleyse: Or... not worry about it right now since it the alternative feels complicated

ptrgags: Options we considered:
1. precompute the closure in MetadataClassProperty  to avoid allocation. though this doesn't change the nested function calls.
2. a lookup table by component type. This would create a lot of boilerplate code and still be closures
3. use `callback.call(undefined, value, ...args)` but this requires ES6
4. use `callback.apply(undefined, [x, otherArgs])` but this causes more allocation...

Issue #10134, opened 2022-02-23
Update and consolidate jasmine usage


ggetz:
We are currently using two separate versions of jasmine-- One older version in `ThirdParty/jasmine-2.2.0` which is being used for running tests in the browser, and a newer version of `jasmine-core` via the `karma-jasmine` module which is being used for running tests on the command line and in CI.

- Is there anything preventing us from updating to the newest version of `jasmine-core`?
- Remove the existing copy of jasmine in `ThirdParty` and set up standalone jasmine consistant with the version of `jasmine-core` in `package.json`
- Audit our custom matchers,`customizejasmine`, and `spec-main` files for functionality redundant with an up-to-date copy of jasmine (namely that after we completely remove when.js and replace with promises, we won't need to do anything special to account for returned promises).

ggetz:
One thing to consider is whether we can drop support for standalone jasmine entirely. I was under the impression that some developers use this for debugging unit tests, but we should discern if that is still the case. We should also take a look at what similar projects do.

ebogo1:
From informal discussions offline, it looks like dropping the standalone SpecRunner files could be the way to go since we don't need the browser debugger for most of our use cases.

I mostly got the [4.0.1 standalone release](https://github.com/jasmine/jasmine/releases/tag/v4.0.1) working with minimal work. The main thing that I couldn't get right was filtering specs that are focused with `fit` or `fdescribe` - the spec runner showed all the skipped tests and took a while to get through them before displaying the tests I intended to run. Steps to reproduce:

- Download 4.0.1 release zip from the link above.
- Copy `jasmine-standalone-4.0.1/lib/jasmine-4.0.1/` folder into `ThirdParty/`.
- Point all the URLs in SpecRunner.html to the new 4.0.1 folder.
- Modify boot1.js to work with Cesium.js similar to the current spec-main.js file. I used the built-in `jasmine.QueryString` instead of `Cesium.queryToObject` for simplicity but I believe either way should work. Once we switch away from `Cesium.when` this step will get simpler too. 

Modified boot1.js file for future reference in case it's helpful: [spec-main.js.txt](https://github.com/CesiumGS/cesium/files/8157296/spec-main.js.txt)

ggetz:
[...]

PR #10170, opened 2022-03-07
Add new options for command-line testing

ebogo1:
This is part of https://github.com/CesiumGS/cesium/issues/10134 - I think removing the standalone Jasmine SpecRunner files should be a follow-up PR so that this one focuses on the new command-line options for `npm run test`:

- `--debug`: overrides the `singleRun = true` flag we set in `karma.conf.cjs`, which allows users to use the "Debug" button in the Karma browser to use the browser debugger for unit tests
- `--includeName {Folder}/{FileName}`: does the same thing as clicking on a test suite in SpecRunner.html. This checks for string equality but could be changed to something else like `contains` if that makes more sense. I don't believe we have access to individual test names inside a suite, so the `fit` syntax is still needed to run specific tests alone.
- `--excludeName {Folder}/{FileName}`: same as `--includeName` but for excluding specs. This wasn't possible in SpecRunner.html but it mirrors the logic we have for including categories on the command line.

I also added a couple lines to the TestingGuide README. I recommend looking at the diff without whitespace changes.


Specs/customizeJasmine.js
```diff
@@ -26,6 +28,12 @@ function customizeJasmine(
       return;
     }
 
+    if (includedName && name !== includedName) {
```
ggetz, line 31: Could we include partial matches? That way we can include all tests in a category. For example @ptrgags said he uses the web browser to run all `ModelExperimental` tests, which span multiple files.

Documentation/Contributors/TestingGuide/README.md
```diff
@@ -197,6 +205,22 @@ It is also possible for Karma to run all tests against each browser installed on
...
 
+Alternatively, a single test suite can be run from the command line with the `includeName` option and a test name argument that matches the name in the `describe` function of the spec file. For example, the `Cartesian2` tests can be run with:
+
+`npm run test -- --includeName Core/Cartesian2`
+
+Similarly, test suites can be excluded with the `--excludeName` flag.
+
+#### Using Browser Debugging Tools
+
+If it is helpful to step through a unit test in a browser debugger, the `singleRun` Karma option can be disabled like so:
```
ggetz, line 216:
```suggestion
If it is helpful to step through a unit test in a browser debugger, run the tests with the `debug` flag.
```

I don't think readers need or want to know the underlying karma confirmation we're using, so let's just omit it for simplicity's sake.

ggetz (commented):
Thanks @ebogo1! I just have a few comments.

Specs/customizeJasmine.js
ptrgags: So just for some more information on this, the SpecRunner query seems to do a case-sensitive substring search.

Though I think even a regex search would be easy to add with `new RegExp(query)` -- might be nice? then you could run custom queries like `--includeName 'Cesium3DTileset|ModelExperimental'`. Not necessary but might be nice to have

ebogo1:
@ggetz I changed the string comparison to check for matching substrings in [b06c640](https://github.com/CesiumGS/cesium/pull/10170/commits/b06c6406eaf7c4f4d6824ebc099f3193c9a7e309) - should be good for another look. Per Peter's [comment](https://github.com/CesiumGS/cesium/pull/10170#discussion_r821742167) - I can add a regex check instead but I'm not sure how common that use case is. Yesterday we decided it might be overkill so I didn't make the change yet.

ggetz:
> I can add a regex check instead but I'm not sure how common that use case is. Yesterday we decided it might be overkill so I didn't make the change yet.

I agree, we were looking to get to parody with the existing browser runner so this should be sufficient.

ggetz:
Thanks @ebogo1! All the new options are working as expected by themselves and alongside other options.

PR #10172, opened 2022-03-07
Handle required/noData/default metadata

ptrgags:
This PR adds support for the following details about metadata properties in 3D Tiles 1.1 and prior extensions:

1. `required` properties (and `optional` in EXT_feature_metadata) are now validated
2. `noData` values return `undefined`. 
3. `default` values are returned when the value is undefined.

@lilleyse could you review?


Source/Scene/MetadataClassProperty.js
```diff
@@ -679,6 +737,15 @@ MetadataClassProperty.prototype.packVectorAndMatrixTypes = function (
...
  */
 MetadataClassProperty.prototype.validate = function (value) {
+  if (!defined(value) && defined(this._default)) {
+    // no value, but we have a default to use.
+    return undefined;
+  }
+
+  if (this._required && !defined(value)) {
+    return `required property must have a value`;
+  }
```
lilleyse, line 747: Should these two statements be swapped? If the property is required it should error out even if it has a default value.

This should also be enforced at the spec level. `default` and `noData` should only be allowed if `required` is false. I'll add that clarification.

lilleyse (commented):
Looks good!

PR #10174, opened 2022-03-08
Fold `3DTILES_multiple_contents` into tilesets for 3D Tiles 1.1

j9liu:
This PR incorporates the multiple contents extension into the tilesets, such that content is available in `tile.contents` without requiring the extension. 

- Added checks for `tile.contents` and `tile.content`, as well as `extension.contents` and `extension.content` (in accordance with [this change](https://github.com/CesiumGS/3d-tiles/issues/496))
- Updated and reorganized multiple content specs
- Updated sample data in `MultipleContents`, `Implicit`, and `Metadata` tileset folders

cc @ptrgags for review


Source/Scene/Cesium3DTile.js
```diff
@@ -56,7 +56,17 @@ import Pass from "../Renderer/Pass.js";
...
   this._tileset = tileset;
   this._header = header;
-  const contentHeader = header.content;
+
+  const hasContentsArray = defined(header.contents);
+  const hasMultipleContents =
+    (hasContentsArray && header.contents.length > 1) ||
+    hasExtension(header, "3DTILES_multiple_contents");
+
+  // In an older schema, content is stored in tile.content instead of tile.contents
```
ptrgags, line 65:
```suggestion
  // In the 1.0 schema, content is stored in tile.content instead of tile.contents
```

Source/Scene/Cesium3DTileset.js
```diff
@@ -1995,14 +1995,15 @@ function makeTile(tileset, baseResource, tileHeader, parentTile) {
...
-      content: {
-        uri: contentUri,
-      },
-    };
 
     const deepCopy = true;
-    const tileJson = combine(contentJson, tileHeader, deepCopy);
+    const tileJson = clone(tileHeader, deepCopy);
+    // Replace contents with the subtree
+    tileJson.contents = [
```
ptrgags, line 2002: what if `tileJson.content` (singular) exists? should that key be deleted?

Specs/Scene/Cesium3DTilesetSpec.js
```diff
@@ -5597,16 +5561,342 @@ describe(
...
-          function (tileset) {
+        return Cesium3DTilesTester.loadTileset(scene, multipleContentsUrl)
+          .then(function (tileset) {
             const statistics = tileset._statistics;
             expect(statistics.visited).toEqual(1);
             expect(statistics.numberOfCommands).toEqual(totalCommands);
-          }
-        );
+          })
+          .otherwise(console.error);
```
ptrgags, line 5570: Please remove this debugging line

ptrgags (commented):
@j9liu just a few points of feedback

Source/Scene/Cesium3DTileset.js
```diff
@@ -1995,14 +1995,19 @@ function makeTile(tileset, baseResource, tileHeader, parentTile) {
...
         uri: contentUri,
       },
-    };
+    ];
 
-    const deepCopy = true;
-    const tileJson = combine(contentJson, tileHeader, deepCopy);
+    if (defined(tileJson.content)) {
+      delete tileJson.content;
+    }
```
ptrgags, line 2010: `delete` always works even if the key is not defined, the if clause can be removed.

Source/Scene/ImplicitTileset.js
```diff
@@ -246,10 +251,12 @@ function makeTileHeaderTemplate(tileJson) {
...
   // content is handled separately, so remove content-related properties
-  delete template.content;
-
   if (hasExtension(template, "3DTILES_multiple_contents")) {
     delete template.extensions["3DTILES_multiple_contents"];
+  } else if (defined(template.contents)) {
+    delete template.contents;
+  } else if (defined(template.content)) {
+    delete template.content;
   }
```
ptrgags, line 260: :thinking: this section can also be simplified I think.

Though the one thing to be careful of is make sure `template.extensions` is defined first, `delete template.foo` will work even if foo is undefined, but `delete template.foo.bar` will fail if foo is undefined.

ptrgags (commented):
@j9liu just noticed one small thing about the `delete` statements so a couple lines here can be simplified a bit.

j9liu:
@ptrgags updated!

ptrgags:
@j9liu Looks good, just waiting on CI

ptrgags:
Thanks @j9liu!

PR #10183, opened 2022-03-09
Fix `pnts` crash when using `ModelExperimental`

lilleyse:
https://github.com/CesiumGS/cesium/pull/10138 introduced a small regression where `pnts` tiles would crash when using `ModelExperimental`. This is because the code that adds credits expects the `asset` property to exist, which it does for anything going through `GltfLoader`, but not other things that create `ModelComponents` directly like `pnts`.

https://github.com/CesiumGS/cesium/blob/2d5fa80f61b47a5266478e57cee25353096c5000/Source/Scene/ModelExperimental/ModelExperimental.js#L815-L823

The quick fix is to always just initialize the `asset` property with a default constructed `Asset`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### 1.92 - 2022-04-01
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu:
@lilleyse looks good to me! I tested locally on the "3D Tiles Point Cloud Classification" sandcastle and the fix prevents it from crashing. Merging now.

PR #10186, opened 2022-03-10
A proposed implementation for giving control on fog rendering. 

xtassin:
In some circumstances, we may want to enjoy the benefits of fog based tile loading optimization without the visual rendering of fog (ie. custom fog shader)


cesium-concierge:
Thanks for the pull request @xtassin!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fogRenderingControl/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @xtassin! 

@lilleyse Could you take a quick look and verify this is something that makes sense for us to support through the API before moving forward? Or is this an option that should live on the tileset or terrain itself?

lilleyse:
I think it's fine since `enabled` is already a global property of `Fog` and `renderable` feels like similar scope.

Specs/Scene/GlobeSurfaceTileProviderSpec.js
```diff
@@ -519,6 +519,35 @@ describe(
...
+        scene.camera.lookUp(1.2); // Horizon-view
+
+        return updateUntilDone(scene.globe).then(function () {
+          expect(scene).notToRender([0, 0, 0, 255]);
+
+          scene.fog.enabled = true;
+          scene.fog.renderable = false;
+          scene.fog.density = 1.0;
+          scene.fog.screenSpaceErrorFactor = 0.0;
+
```
ggetz, line 544: Thanks for adding the test! To make this a bit more robust, I would check to make sure that what renders when `renderable` is `false` is also not the color which renders when fog is enabled with `renderable` set to `true`. 


```suggestion
          scene.fog.enabled = true;
          scene.fog.density = 1.0;
          scene.fog.screenSpaceErrorFactor = 0.0;

          let result;
          expect(scene).toRenderAndCall(function (rgba) {
            result = rgba;
            expect(rgba).not.toEqual([0, 0, 0, 255]);
          });
          
          scene.fog.renderable = false;
          expect(scene).notToRender(result);
```

ggetz:
@xtassin Could you please merge in the upstream changes from `main`? I'm having trouble checking on your branch to resolve the changes myself.

Specs/Scene/GlobeSurfaceTileProviderSpec.js
```diff
@@ -519,6 +519,46 @@ describe(
...
+        return updateUntilDone(scene.globe).then(function () {
+          expect(scene).notToRender([0, 0, 0, 255]);
+
+          scene.fog.enabled = true;
+          scene.fog.renderable = false;
+          scene.fog.density = 1.0;
+          scene.fog.screenSpaceErrorFactor = 0.0;
+          scene.fog.enabled = true;
+          scene.fog.density = 1.0;
```
ggetz, line 545:
```suggestion
          scene.fog.density = 0.001;
```

Sorry, something went a bit weird with the suggestion. I think it was my fault.

xtassin:
@ggetz Original PR is too old and I need to create a new Pull request in order to merge upstream: do you want me to do that?

ggetz:
There shouldn't be a need to open a brand new pull request. You should be able to merge changes from the main branch of cesium into this one [using these instructions](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodeReviewGuide#merge), but let me know if you run into any issues. 

As a part of that process, you'll resolve the conflict in `CHANGES.md` by accepting both sets of changes. 

After merging, you should also be able to run `npm run prettier` to format all the code and commit the result, which should resolve the error which is showing up in the logs after running the checks on this branch.

ggetz:
Awesome, thanks for the updates here @xtassin! Looking forward to getting your other pull requests in as well.

PR #10187, opened 2022-03-10
A proposed implementation for passing imagery layers data to Fragment Shaders

xtassin:
A proposed implementation for passing imagery layers raster data to a globe material vertex shader for further processing. A current limitation is the way the layers are being iterated in GlobeFS.glsl before setting values to the czm_materialInput->layerColor array. This induces long shader compilation times and requires to set a fixed number of processed layers (set as the GLSL constant layerColorNumber)

### DIFF
--- a/Source/Shaders/Builtin/Structs/materialInput.glsl
+++ b/Source/Shaders/Builtin/Structs/materialInput.glsl
@@ -25,4 +25,5 @@ struct czm_materialInput
     float height;
     float slope;
     float aspect;
+    vec4 layerColor[czm_layerColorNumber];
 };
--- a/Source/Shaders/GlobeFS.glsl
+++ b/Source/Shaders/GlobeFS.glsl
@@ -1,6 +1,7 @@
 uniform vec4 u_initialColor;
 
 #if TEXTURE_UNITS > 0
+uniform int u_layerIndex[TEXTURE_UNITS];
 uniform sampler2D u_dayTextures[TEXTURE_UNITS];
 uniform vec4 u_dayTextureTranslationAndScale[TEXTURE_UNITS];
 uniform bool u_dayTextureUseWebMercatorT[TEXTURE_UNITS];
@@ -160,8 +161,13 @@ bool inTranslucencyRectangle()
 }
 #endif
 
+#ifdef APPLY_MATERIAL
+czm_materialInput materialInput;
+#endif
+
 vec4 sampleAndBlend(
     vec4 previousColor,
+    int layerIndex,
     sampler2D textureToSample,
     vec2 tileTextureCoordinates,
     vec4 textureCoordinateRectangle,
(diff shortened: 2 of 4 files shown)


Source/Shaders/GlobeFS.glsl
```diff
@@ -191,17 +197,27 @@ vec4 sampleAndBlend(
...
 
+    #ifdef APPLY_MATERIAL
+    if (textureAlpha > 0.0) {
+        // itterate over the layerColor array to check for matching layerIndex and send color value to materialInput
+        //TODO: Find a different way to set pass layerColor values. Here, only the first two layers are handled as this loop leads to performance problems (shaders compilation time)
+        for (int i = 0; i < czm_layerColorNumber; i++) {
+            if (i == layerIndex) {materialInput.layerColor[i] = value; break;}
+        }
+    }
+    #endif
```
lilleyse, line 215: It's unfortunate that `czm_layerColorNumber` would be capped to the first two layers due to slow compilation.

I can't think of a good way to avoid this in WebGL 1 unless `layerIndex` were somehow made to be a constant expression, and that seems tricky since `GlobeSurfaceShaderSet.prototype.getShaderProgram` doesn't have access to the imagery layers.

Since WebGL2 allows dynamically indexed arrays this would simply be `materialInput.layerColor[layerIndex] = value`. Another reason to switch to WebGL 2 soon.

Source/Shaders/Builtin/Structs/materialInput.glsl
```diff
@@ -25,4 +25,5 @@ struct czm_materialInput
     float height;
     float slope;
     float aspect;
+    vec4 layerColor[czm_layerColorNumber];
```
lilleyse, line 28: Just confirming - to check that a layer exists would the material have code like `materialInput.layerColor[index] != vec4(0.0)`?

`layerColor` should be added to the documentation above.

lilleyse, line 28: Should there be a `layerColorNumber.glsl` file? I had to create one locally to test this. Also a sandcastle demo would be helpful. I modified the Globe Materials sandcastle.

Source/Shaders/GlobeFS.glsl
xtassin: Thanks for the feedback @lilleyse. This pull request is just me thinking out loud about this feature and how best to implement it. In my use case, only one layer is made available to the shader. To work around the limitation of a hard coded number of layers, could this number be set by the user in the form of a "define"?

Source/Shaders/Builtin/Structs/materialInput.glsl
xtassin: This can be a way to test it, indeed. If ever the number of layers passed to the material could be set by the user, it would be more obvious which layer colors are made available to the material.

xtassin: Sorry, I forgot to check in layerColorNumber.glsl
I would like to find the best way to implement this and make sure it just makes sense at all before writing any demo. This is why your input is very valuable here.

Source/Shaders/GlobeFS.glsl
lilleyse: > To work around the limitation of a hard coded number of layers, could this number be set by the user in the form of a "define"?

Possibly. I considered that as well. The only problem is that `materialInput.glsl` is used by other systems that don't have access to the define — though they could mock the define and set it to 1.

I'm not sure about the exact requirements of your application, but would it be acceptable for the material to be a single color after all the imagery layers have been resolved rather than an array of layer colors? That would simplify things quite a bit.

xtassin: My application needs the color of one specific layer, so using the resulting mix of all layers will not be acceptable. In any case, this pull request aims at finding a general solution that can be useful to Cesium users at large. I will investigate a "define" based approach to the problem and post results in the PR.

cesium-concierge:
Thanks again for your contribution @xtassin! 

xtassin:
I haven't given up on this feature and am still in the process of evaluating performance from a live implementation. Handling the number of layers dynamically is still very much an issue that need to be addressed in a acceptable manner.

cesium-concierge:
Thanks again for your contribution @xtassin! 

cesium-concierge:
Thanks again for your contribution @xtassin! 

cesium-concierge:
Thanks again for your contribution @xtassin! 

cesium-concierge:
Thanks again for your contribution @xtassin! 

PR #10190, opened 2022-03-10
Add unit tests for unnormalizing metadata

ptrgags:
Fixing an omission in the original metadata code, we test `normalize` but not `unnormalize`.

I refactored the unit tests to make use of round-trip values for simplicity. Along the way, I realized that `normalize` and `unnormalize` are intentionally not inverses (they're _almost_ inverses except for the minimum value for signed integers, see [this Desmos graph](https://www.desmos.com/calculator/vgx2m46ons). I left a note in the descriptions to warn about that.

@j9liu could you review?


j9liu (commented):
@ptrgags this looks good to me, I'm gonna go ahead and merge. thanks!

PR #4858, opened 2017-01-12
Imagery Split Rendering

jasonbeverage:
Hi Cesium Team!

I've got another fun feature that I'd love your feedback on.

This PR adds support for rendering ImageryLayers with a splitter.  I've put a demo here:
http://cesiumsplitscreen.netlify.com/apps/sandcastle/?src=Imagery%20Layers%20Split.html&label=Showcases

The primary use case for this is to compare imagery layers in the same area that were collected at different times.

The big changes to the Cesium codebase are:
*  A new "split" property on ImageryLayer that controls where the image should render.  -1.0 means only show to the left of the splitter, 1.0 means show to the right of the splitter and 0.0 means render as usual.
*  Changes to GlobeSurfaceTileProvider.js, GlobeSurfaceShaderSet.js, GlobeFS.glsl to support splitting.
*  A new uniform in AutomaticUniforms.js called czm_splitPosition that is a float between 0.0 and 1.0 representing the relative splitter position in the viewport.  This is driven by the splitPosition property in FrameState.js.

The sandcastle example I've linked to uses a very simple HTML/JS/CSS slider implementation to control the splitter position.  I thought it was important not to get too heavy handed in providing a splitter implementation and allow users to control it and style it using whatever they wanted to in their application.

Thanks!

Jason

Issue #9520, opened 2021-04-30
3D Tiles Next CesiumJS Roadmap


ptrgags:
**3D Tiles Next** adds several new, experimental capabilities to 3D Tiles. Currently, they take the form of extensions to 3D Tiles 1.0 and glTF 2.0. An overview of these specifications can be found in the [`3d-tiles`](https://github.com/CesiumGS/3d-tiles/tree/main/next) repository. See also the [list of resources](https://github.com/CesiumGS/3d-tiles/blob/main/next/RESOURCES.md)

This issue tracks the changes to CesiumJS and the implementation status for each.

## 3D Tiles Next Extensions - Initial Implementation

These features are now available as of CesiumJS 1.87.1

- [x] [`3DTILES_implicit_tiling`](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0) - **experimental** implementation  in #9517
    - [x] Implement an `Implicit3DTileContent`, which transcodes implicit tiles to `Cesium3DTile` objects when the content loads. This naïve transcoder is similar in function `Tileset3DTileContent` (external tilesets).
- [x] [`3DTILES_multiple_contents`](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_multiple_contents/0.0.0) - **experimental** implementation in #9517 - However, we found some caveats to the specification, in particular that there is no way to use contents that require mixed `ADD/REPLACE` refinement. This specification may be revised in the future
    - [x] Implement `Multiple3DTileContent`
    - [x] Add support for multiple contents in implicit tiling.
- [x] [`3DTILES_metadata`](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_metadata/1.0.0) - **experimental** implementation in #9517
    - [x] Create classes for representing metadata concepts
    - [x] Add tileset, tile, and group metadata to 3D Tiles 1.0 classes
    - [x] Add tile and group metadata support to implicit tiling
    - [x] Add group metadata support to multiple contents
    - [x] Add `Cesium3DTileFeature.getPropertyInherited()` for CPU styling. Now the styling language will use tileset, tile, and group metadata
- [x] [`3DTILES_content_gltf`](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_content_gltf) - **experimental** implementation in #9517
[...]

PR #10193, opened 2022-03-13
Add the ability to show a Cesium3DTileset on only the left or right side of the splitter, for visual comparison

kring:
Imagery Layers have a nice feature, added in #4858, where you can display them only one side or the other of a screen-space splitter, making it really easy to do visual comparisons of two imagery layers.  This was hugely popular in TerriaJS, and probably in lots of other applications as well. And not long after that PR was merged, people started asking for the ability to do the same with other types of things.

Just five short years later, this PR adds support for splitting 3D Tiles. And it adds a bit of infrastructure that should make it pretty easy to add the feature to other primitives, too.

Here's a link to the new Sandcastle demo that uses the feature to compare Melbourne Photogrammetry with Cesium OSM Buildings:
http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/split-primitives/Apps/Sandcastle/index.html?src=3D%20Tiles%20Compare.html

And here's another (not included as a default Sandcastle app) that compares "New York City 3D Buildings" with Cesium OSM Buildings (zoom in toward Manhattan):
[Long Sandcastle Link](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/split-primitives/Apps/Sandcastle/index.html#c=tVVtT9swEP4rtzKkdAOniKFppaDx0g0koIgW+FIJgnOhFo4d2U5LQfz32XFSUt6FtFZq6vNzz909d3aoFNrAmOEEFWyAwAnsoGZ5Ss4KWzBs0GK9I4WJmEA1bDTXh2IoaOHJMTHWzxMQTVEgyRRLmWFj1CSK42AooM7rH6u7A8ZRownu3T5Arni7guxLcYJa5ooiSZRMt7QF7sfBz7XVH6vNJefw0BwKn4fLgOiMM7PLFFLDpLAJlUz9OTs56P4Z1JJX7Hr0XvYlEVUYGezpdDtnPGbiWgdNF7+g+Gj4k/2/ez5+GfFOynQgA1eCLyYMoT8VFMwIIZOaFXQyKdaas9g2acLMyK8d+QxV1VSiNiCWNE9RGHKNpsvR/d2eWhGHDQ/xfZwrvWA8rsJuOJ09lsgksS04cM0OK1sWKctZUpeIcxabUU3hUSRi/nSy+lZMFP0sotgdW989Dwo8b9VVA6kc4xZ1rbD+ScQ1FltJLrzMbj9wPy6BJhSDxBIIvjw6llYAhSZXYr0YHUcCUI4A8sgBe0X6Nk7FR1DElRTkdv3R4wWRAILnOn1/wt2E0GM/IN8MpM2UIykP2eXX+5VWi7Tg23wSD4uXhc9bvZz3sPBChbI9xB0vkeVmqxA2mCkclPrNdcKo3DXiYem1fg6mGRZH7WK3d37k+vkf4xzvH+3sXfQHWycDPzmvxHLUbzMd9k773YvD3ln3jZTfp/EJVTSfrL2a9o+IfHr8aYk/GMaX1D3adYEaS41OMZabfpwBfrM0k8q4KzwgJDSYZnbwUYdXOb1BQ6jWXgoPXyivqPvKALNLrA3RlZY8N7j+uOmmvw1rrcWazcisDa3stma6iujNtZK5iJep5FK1YSFedd8aZuLOl+Wacxyhu8PbYM9WPcTdMhMx3rbhl/3M7A/P6miPrKJz1dBcaRcfJ8vKqnqHdW/36IR1ATsxGwOLN1540wLlkdZ2J8k571umYWOm+qNbdZ9vdkJrq1jr/2dQLiP38urZjHk0dS6jlc0DbySEdEK7fNnTSMmvoqdR/gE)


cesium-concierge:
Thanks for the pull request @kring!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Awesome @kring ! Does this include any tile culling optimizations like what exists for clipping planes?

kring:
> Does this include any tile culling optimizations like what exists for clipping planes?

No, it doesn't currently. It's probably not too hard to do, as long as off-center frustum culling works correctly, which I think it does. It _might_ make for a worse user experience, though. The way this is typically used is the user drags the split point back and forth across the screen. If tiles at the split point initially pop in at lower detail, or if there's any frame rate drop (e.g. on slower systems) while dragging the split point, that won't be so good. That might be just an excuse, but I'd probably suggest treating it as a separate feature/PR in any case.

Source/Scene/Model.js
```diff
@@ -537,6 +540,18 @@ function Model(options) {
    */
   this.showOutline = defaultValue(options.showOutline, true);
 
+  /**
+   * The {@link SplitDirection} to apply to this model.
+   *
+   * @type {SplitDirection}
+   * @default {@link SplitDirection.NONE}
+   */
+  this.splitDirection = defaultValue(
```
ggetz, line 549: Should this property also be added to `ModelExperimental` at this point? CC @lilleyse?

lilleyse: Sure, that would be helpful. Otherwise we can add it to https://github.com/CesiumGS/cesium/issues/9520

PR #10199, opened 2022-03-15
Apply offset/scale to property tables for CPU metadata classes

ptrgags:
This PR handles the `offset/scale` properties that transform float values as follows:

```
transform(value) = scale * value + offset
```

This change was made for property tables and the CPU-side metadata classes (e.g. MetadataClassProperty, MetadataTableProperty, MetadataEntity). The details about textures and property mappings are not handled since that will only become relevant once we add metadata to custom shaders.

@lilleyse could you review?


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/GltfLoaderSpec.js
```diff
@@ -1204,6 +1204,18 @@ describe(
           const expected = expectedEmployeeCounts[i];
           expect(propertyTable.getProperty(i, "employeeCount")).toBe(expected);
         }
+
+        // the offset/scale from the property table should be used, so the
+        // temperatures should be between the range 18-24°C rather than
+        // the class property range of 0-100°C
```
lilleyse, line 1210: Nice idea for the unit test here!

lilleyse:
Great way to finish the 3D Tiles 1.1 PRs. Thanks @ptrgags!

PR #10124, opened 2022-02-22
Affine transformation helpers functions for `Matrix` classes

IanLilleyT:
This PR fills in some gaps in `Matrix2`, `Matrix3`, and `Matrix4` for affine transformations. This includes:

- `setScale` - for `Matrix2`, `Matrix3`
  - exists in `Matrix4` already
- `setUniformScale` - for `Matrix2`, `Matrix3`, `Matrix4`
- `multiplyByUniformScale` - for `Matrix2`, `Matrix3`
  - exists in `Matrix4` already
- `setRotation` - for `Matrix2`, `Matrix3`, `Matrix4`
- `getRotation` - for `Matrix2`, `Matrix4`
  - exists in `Matrix3` already
- `fromRotation` - for `Matrix4`
  - `Matrix2` and `Matrix3` already have their own versions of this

Also did some cleanup in the specs so that all tests are grouped by:
- Capability tests
- Missing parameter tests
- Missing result parameter tests

Misc:
- Inlined most of the math for better performance
- Coverage is 100%

PR #10209, opened 2022-03-18
Matrix4 setScale bugfix

IanLilleyT:
:facepalm: 

I made a mistake in https://github.com/CesiumGS/cesium/pull/10124 and the unit test was not good enough to catch it. Fortunately this bug never made it into a release.


cesium-concierge:
Thanks for the pull request @IanLilleyT!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/matrix4ScaleFix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ebogo1 (approved):
Good timing with the fix, thanks @IanLilleyT

PR #10211, opened 2022-03-21
Update processPbrMaterials.js

pasu:
it should be specularColor (F0) instead of F, I think.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -315,3 +315,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Justin Peter](https://github.com/themagicnacho)
 - [Gu Miao](https://github.com/Gu-Miao)
 - [Shen WeiQun](https://github.com/ShenWeiQun)
+- [Peter Lu](https://github.com/pasu)
--- a/Source/Scene/processPbrMaterials.js
+++ b/Source/Scene/processPbrMaterials.js
@@ -1012,13 +1012,13 @@ function generateTechnique(
     fragmentShader +=
       "    vec3 specularIBL = czm_sampleOctahedralProjection(czm_specularEnvironmentMaps, czm_specularEnvironmentMapSize, cubeDir,  roughness * czm_specularEnvironmentMapsMaximumLOD, czm_specularEnvironmentMapsMaximumLOD);\n";
     fragmentShader += "#endif \n";
-    fragmentShader += "    specularIBL *= F * brdfLut.x + brdfLut.y;\n";
+    fragmentShader += "    specularIBL *= specularColor * brdfLut.x + brdfLut.y;\n";
     fragmentShader += "#else \n";
     fragmentShader += "    vec3 specularIBL = vec3(0.0); \n";
     fragmentShader += "#endif \n";
 
     fragmentShader +=
-      "    color += diffuseIrradiance * diffuseColor + specularColor * specularIBL;\n";
+      "    color += diffuseIrradiance * diffuseColor + specularIBL;\n";
 
     fragmentShader += "#endif \n";
   } else {


cesium-concierge:
Thank you so much for the pull request @pasu! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/pasu/cesium/blob/main/CONTRIBUTORS.md) file!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu:
Hi @pasu ! Thank you for the contribution. I should be able to review this later this week and check for correctness.

pasu:
Thank you for your quick response and review .

j9liu:
Hi @pasu,

Thank you for your patience. When I tested this out, I compared the results of your changes with our current implementation of image-based lighting, using the [Image-Based Lighting sandcastle](https://sandcastle.cesium.com/?src=Image-Based%20Lighting.html). Here's what a side-by-side comparison looks like:

sidebyside

Your changes make the pawns a lot brighter, especially the gold and black ones, but I don't think that's physically correct. For reference, the image-based lighting in Cesium is implemented similarly to the code below, which is taken from [the LearnOpenGL guide](https://learnopengl.com/PBR/IBL/Specular-IBL) on image-based lighting:
```
vec3 F = FresnelSchlickRoughness(max(dot(N, V), 0.0), F0, roughness);

vec3 kS = F;
vec3 kD = 1.0 - kS;
kD *= 1.0 - metallic;	  
  
vec3 irradiance = texture(irradianceMap, N).rgb;
vec3 diffuse    = irradiance * albedo;
  
const float MAX_REFLECTION_LOD = 4.0;
vec3 prefilteredColor = textureLod(prefilterMap, R,  roughness * MAX_REFLECTION_LOD).rgb;   
vec2 envBRDF  = texture(brdfLUT, vec2(max(dot(N, V), 0.0), roughness)).rg;
vec3 specular = prefilteredColor * (F * envBRDF.x + envBRDF.y);
  
vec3 ambient = (kD * diffuse + specular) * ao; 
```
So the `F` term is actually important because it factors in the material's roughness and other properties. When we remove it, the lighting becomes too bright and physically inaccurate.

I hope this explanation makes sense -- feel free to ask if anything is unclear. I'm going to close this pull request now.

pasu:
Hi, @j9liu 

Theoretically, [the LearnOpenGL guide](https://learnopengl.com/PBR/IBL/Specular-IBL) shows it should be F_0
image

I also checked the glTF's IBL demo here:
[        vec3 specular = specularLight * (pbrInputs.specularColor * brdf.x + brdf.y);](https://github.com/erich666/glTF-WebGL-PBR/blob/master/shaders/pbr-frag.glsl#L159)

and Threejs's related fragment BRDF_Specular_GGX_Environment, which is based on  [EnvBRDFApprox](https://www.unrealengine.com/en-US/blog/physically-based-shading-on-mobile)

both use specularColor  (F_0), so, I still think the code from LearnOpenGL is not accurate.

j9liu:
Hi @pasu ,

The image you posted is from a different part of the webpage I sent. That part is relevant to calculating the BRDF texture, `brdfLut`. What I was discussing is under the "Completing the IBL Reflectance" section, which is also what you were suggesting to change in Cesium's code.

I'd have to look more closely at the two links you posted in order to understand their logic. Regardless, following your suggestions in Cesium leads to an overly brightened model that is inaccurate to its PBR qualities. For reference, I put the same model in two other rendering engines. This is how it looks in [Babylon.js Sandbox](https://sandbox.babylonjs.com/):
image

And here are the pawns in [PlayCanvas](https://playcanvas.com/viewer):
image

In both engines, the black and blue pawns are not intensely brightened by the image-based lighting. Compare this to what the pawns look like with your changes:
image

What we have currently is more accurate. Different rendering engines may interpret and implement rendering equations differently, and small changes in the code can lead to drastically different results. Therefore, it's not good to implement this change *just because* that is how other engines are doing it.

Let me know if I can help with anything else.

pasu:
Hi, @j9liu 

I added SRGBtoLINEAR3 for the gamma correction, and now, it doesn't look as bright.


[`    fragmentShader += "    specularIBL *= SRGBtoLINEAR3(specularColor * brdfLut.x + brdfLut.y);\n";`](https://github.com/pasu/cesium/blob/main/Source/Scene/processPbrMaterials.js#L1015)

j9liu:
Hi @pasu,

After searching through other implementations of image-based lighting, I did find that a couple of them used `specularColor` in their computation like you suggest. However, others use the Fresnel coefficient in their calculations, such as [Google Filament](https://google.github.io/filament/Filament.html#lighting/imagebasedlights/distantlightprobes) and [glTF-Sample-Viewer](https://github.com/KhronosGroup/glTF-Sample-Viewer/blob/master/source/Renderer/shaders/ibl.glsl#L19). In particular, we are more inclined to use glTF-Sample-Viewer as reference, and according to it, the Fresnel coefficient does factor into the final computation.

Wrapping the specular color in `SRGBtoLINEAR3` doesn't make sense here, either. Unless a variable is being read from an `SRGB` texture, it should already be in linear space. So it isn't accurate to apply it to the `specularColor` here. It is confusing why it appears in the procedural skybox code, so thank you for pointing that out. We'll have to look into that later, but it's not a big priority right now.

Let me know if anything doesn't make sense.

pasu:
Hi, @j9liu 

Thank you for your sharing, I got this trick.

The result is that either is reasonable based on this [article](https://bruop.github.io/ibl/):
F_0 is for single-scattering while F_r is for multiple-scattering, F_r is called 'Roughness dependent fresnel'

Thank you for this meaningful discussion.

PR #10217, opened 2022-03-21
Update 3D Tiles Interior Sandcastle

srothst1:
The goal of this pull request is to update the current version of our [3D Tiles Interior sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20Interior.html) with some new functinality. 

In [this private support thread](https://community.cesium.com/t/3d-tiles-interior/17562), I added a bit of extra functionality to our 3D Tiles Interior demo. This functionality makes it easier to move the camera in and out of the interior of the building.

image

I just wanted to give folks the opportunity to add this update to the live sandcastle 😀 📸


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/update-3d-tiles-interior/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @srothst1.  Could you provide more details about what you're trying to demonstrate with the new views? Is it to show that you can easily navigate from an exterior to an interior? 

If that's the case, I think that concept is a good showcase of Cesium, but I don't think it's entirely clear from this example as the outside of the model doesn't appear to be intended for a view from the exterior. Would it be possible to get a tileset with a more deliberate exterior and interior to better demonstrate this concept?

srothst1:
Hi @ggetz. Yes, the point is to showcase that you can easily navigate from an exterior space to an interior space. Sometimes CesiumJS camera controls can be a bit fidgety in small spaces. This sandcastle presents a simple workaround that met the needs of various community members. Sadly, I don't know of a tileset that has a better exterior to interior transition than this one.

ggetz:
I see, thanks! I understand that this may be useful as an example for the forum, but I'm not entirely sure it's working as a showcase. Let me know if you want to find a tileset that may be better suited to a showcase like this, otherwise I think we'll opt to include this at the moment.

PR #10219, opened 2022-03-22
Remove unneeded Jasmine matchers

ebogo1:
This has the following changes:
- `toBeGreaterThanOrEqualTo` switched to built-in `toBeGreaterThanOrEqual` 
- `toBeLessThanOrEqualTo` switched to built-in `toBeLessThanOrEqual`
- `toBeInstanceOf` uses built-in matcher
- `toThrowRuntimeError` and `toThrowSyntaxError` switched to built-in `toThrowError`
- a few minor whitespace fixes

Other candidates for removal are `toEqualEpsilon`, which could use the built-in `toBeCloseTo` if we accounted for our custom matcher handling arrays, and `toThrowDeveloperError` using `toThrowError`. The `toThrowDeveloperError` change is tricky because the behavior depends on whether specs are ran on a release build - the best fix is probably to override the `toThrowError` matcher to support debug vs release builds but I ran into some problems on my first attempt at this. I can open a separate issue for this so we can revisit in the future or leave a comment in https://github.com/CesiumGS/cesium/issues/10134.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/remove-matchers/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #8623, opened 2020-02-20
Sandcastle: Update Codemirror + JSHint to support ES6+ syntax

ptrgags:
when using Sandcastle locally for debugging, I've found it quite annoying that the linter complains about my use of ES6 syntax which I find helpful for prototyping quickly.

I know that there are lots of things that could be better in Sandcastle, but this one was easy to fix in the short term. I updated CodeMirror to 5.52.0 and JSHint to 2.11.0 to support ES6 syntax and beyond in Sandcastle. 

As far as I could tell from searching the code, this shouldn't impact anything beyond Sandcastle, but please correct me if there are any other side effects I'm not aware of.

I set `esversion: 10` in the Sandcastle `.jshintrc`. Maybe that's a bit presumptuous of a setting, but it allows the most flexibility for the user, and it will probably make this tweak effective for a longer period of time without further maintenance.

[...]

PR #10220, opened 2022-03-22
Update jsHint to esversion 11 for Sandcastle

lilleyse:
ES11 features can now be used in Sandcastle without annoying warnings.

* Consolidated `Apps/.jshintrc` and `Apps/Sandcastle/.jshintrc` since jsHint is only used in sandcastle
* Updated jsHint from version `2.11.0` to `2.13.4`
* Updated `esversion` from 10 to 11 so that we can use newer features without warnings showing up in Sandcastle, e.g. optional chaning:

```js
function getMetadata(picked) {
  return picked?.content?.tile?.metadata;
};
```

It might be possible to remove `.jshintrc` and work with `jsHintOptions.js` directly and remove it from the build step, but I didn't quite know what I was getting into there.

This is essentially a follow up to https://github.com/CesiumGS/cesium/pull/8623 but I didn't update CodeMirror.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/jshint-esversion-11/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Apps/Sandcastle/.jshintrc
```diff
@@ -1,12 +1,69 @@
...
+  "supernew": false,
+  "validthis": false,
+  "browser": true,
+  "browserify": false,
+  "couch": false,
+  "devel": true,
+  "dojo": false,
+  "jasmine": false,
+  "jquery": false,
+  "mocha": true,
```
ggetz, line 58: I know this was `true` in the previous configuration, but should the [`mocha` option](https://jshint.com/docs/options/#mocha) be `false`? @mramato Do you know why this option may have been enabled before?

ggetz:
Thanks @lilleyse! This is working great. I just had one comment which doesn't need to hold up this PR.

Apps/Sandcastle/.jshintrc
mramato: Most of these options probably aren't needed at all and can just be removed entirely. I vaguely remember that there was a reason for us to set them all explicitly at some point, but that was so early in development that there isn't even GitHub history for it :open_mouth: 

Some of these probably don't even exist anymore/  All the options are on https://jshint.com/docs/options/, though it looks like it doesn't include default values.

One approach would be delete everything in this file except esversion and then just put stuff back if we noticed it putting up warnings we don't want.  Since this is _just_ used for Sandcastle, it may also be better to play a little looser than what we do in the actual Cesium source.

PR #10224, opened 2022-03-23
CPU Metadata API consistency

ptrgags:
This PR addresses the inconsistency in the API across the different metadata granularities. See https://github.com/CesiumGS/3d-tiles-samples/pull/43#discussion_r832694763

This PR makes these changes:

* In `Cesium3DTileset`, I moved the `Cesium3DTilesetMetadata` object to a separate `metadataExtension` field (intended for internal use). This contains the schema, groups, tileset metadata. 
* Added `tileset.metadata` and `tileset.schema` getters since those are the two most common things the user will need to fetch from the tileset. I could add more like groups, but for now I'd rather keep the API surface area small.
* Added a small wrapper object, `Cesium3DContentGroup` so you can access metadata as `content.group.metadata` rather than the asymmetric `content.groupMetadata`. It may feel a bit like overkill, but this way is better documented and future-proof.

To Do:

- [x] Update CHANGES.md
- [x] Double check sandcastles


ptrgags:
@lilleyse could you review when you get a chance?

lilleyse:
Looks good!

PR #10233, opened 2022-03-24
Only queue billboards for update if they are dirty.

andrewjhill9:
BillboardCollection.prototype._updateBillboard function queues billboards for update if their dirty flag is FALSE. This leads to iteration over the entire BillboardCollection whenever a new billboard is added. This is a huge performance hit when the number of billboards starts to grow into the tens of thousands within a single collection and billboards are added frequently to the collection (multiple times a second). 
If this change is incorrect for some reason, then I would recommend adding a comment to this IF check as it is extremely unintuitive why we would queue billboards for update if they are not dirty.

### DIFF
--- a/Source/Scene/BillboardCollection.js
+++ b/Source/Scene/BillboardCollection.js
@@ -566,7 +566,7 @@ BillboardCollection.prototype._updateBillboard = function (
   billboard,
   propertyChanged
 ) {
-  if (!billboard._dirty) {
+  if (billboard._dirty) {
     this._billboardsToUpdate[this._billboardsToUpdateIndex++] = billboard;
   }
 


ggetz:
Hi @andrewjhill9, this change seems to be breaking some billboard behavior. For example [clamping to ground](https://sandcastle.cesium.com/#c=1VdbU+M2FP4rap7sEhQDhQQSmLJZuuxMuExIdx+aDqPYx46miuSR5ZB0J/+9km9RnEDDtLPTvgCWvnP7Pp0j4QueKDSn8AISXSIOL6gPCU1n+Eu25owbfvbdF1wRykGOG030bcwRUiClXnmUYk4DkBeloS+BKPgqJAtGOcRxm2O+crtjngfCiQ8ccMTEBHAAsZqOIFHXkYYmqrDRySiZgrbxsxQzE71oe6g29U8FC6W3c8/Ft94f83wlgEkaPU3Fyy+SzCB5BPkEGhWsw5S+gDEaJ4IGtWBFutV2FXxCGZsIIoO+YAx8RQWvm8aSzqiic0gwCQLHsGcx/WHbgZMxjPJyLjacNc3Oyh1zN0u61UKJXvVJxQXo2uVA8IiqNDCUHR7hk87psXf608nJ8dl5+/i8WwMTVWI9fNZut8+Pz047XqfTOTo7saiJJA2+0kBNNfCOqCkOmRDSyf6UhAdi5rjoR3Tkedhzu7bVLdBoqt5lJjUThEcMbgkLn+ifeXae553aYm2e2WFpk1FcZ+Jw22fTwpUk/A2scnewn7sdsFK7UEjkMFBoaWrr6l89i60uOjhYunmvVchFjlwUyEwNA1wUQIRyYpilf0FOxjcDGTs5ECHAL7rvmutPINbnArWQs5b8EB25+ZZJ3oq0PjxvBEpEqqZWJC6k9b0sQxXn5LVYsUho0V+W6n0ilYgkiafUd6rCm1VmOdnGy45WzRryW5lIGeBiPQawb7kfCRNMhyXcKbFuVQadkUh367iBcSv7O2mFxKeMqiWOaKgnZ4mcZnUOIQQJ3Idqdt5uruP+4Pru8Xn08Pxp+PDr/cfCfpWzstJD1VRmzkVVWVbrk+4qX2vJwJQ3EoJNiPyQKqUny7hxHQRrvBnnYcrzseUUp4iGyPmhSCmAUI/9wKks3OqoVUtaj3+V2tIGvSpzrRmbta5rmmlixknp6U2R8gDPAxGJZzEHycgSxzyy9Ep8wrSRh9vvlnB4M7geff5y85aK7h6qDWGmc9tDuHfoZmklM/cWurutccoLp/skfgc8dX4zwCKquZEN70MwjalLUULPMKGdVkTrJgeT0MV2ZZl0e1W3kTWuyYQu1yC0v2Td0mpVaJffws0d5fUZmcX/3dpqE+Udhd0LDt+1mtdKuH+4v3k9799d632gb4NP+olRPh2Kt8PbjfaZmxdsAus79B+0WvHMskaXzmVzAGYjL59+1Xgrq1uzU47P2rVoGeH1nX9wWVZe72JsXaH7z2F7305g39n1Ef73lB5+d0obzUYvUUsGV6XNz3QW63cTSiVz9O2lYBbrUaovsEnq/wEK+0lSVtFr2aa9gM4RDS53/DOHfEaSRO+EKWPZ47Rx1Wtp/JYpEySgPHrIL0gDmx5dDfJFjHGvpT93W6r8ONQ8/wU) doesn't seem to be working in your branch.

@lilleyse Do you happen to know why this check is as it is?

lilleyse:
Hm that does seem unintuitive.

I'm not super familiar with this code but it looks like this check prevents the billboard from being added to `_billboardsToUpdate` multiple times within the same frame if multiple properties are changed. The code is correct, just confusing.

>  This is a huge performance hit when the number of billboards starts to grow into the tens of thousands within a single collection and billboards are added frequently to the collection (multiple times a second).

I'm not sure of your exact use case, but could you pre-allocate the billboard collection and recycle billboards for better performance?

andrewjhill9:
Thank you for the quick responses. Your reason makes sense lilleyse. My use case is that I am creating billboards for live data updates, and the billboards collections get fairly large.

Also, after thinking about what you said and digging a bit more, I believe I have misdiagnosed the issue. I think it's not so much the iteration that is causing slow performance, but the array assignment itself on Line 570. I have noticed that when arrays is Javascript get rather large, increasing the size of the array utilizing `array[newIndex] = newValue` tends to be a slow operation. I wonder if there are more performant ways to keep track of this data.  I will investigate this further and respond with my findings.

ggetz:
@andrewjhill9 In the interest of keeping the list of pull requests tidy, I'm going to close this PR. But if you have any additional updates, feel free to re-open. Thanks!

Issue #10200, opened 2022-03-15
Unit test command line usage is more limited than the old SpecRunner


ptrgags:
I understand why SpecRunner was removed (to avoid maintaining it in `ThirdParty` separate from `npm`, but the command line testing workflow is more limiting.

One thing I do quite often in SpecRunner is to filter for tests. That way when I'm working on code, I can select a single test or many tests using `?spec=<search query>` or the "Select a unit test to run". This is a lot easier to work with than `fit/fdescribe`.

I've been trying out the new `--includeName` and it's not finding tests from nested describe blocks as well as the `?spec=<search query>` in the `SpecRunner.html` does.

(on branch `extension-revisions`) I wanted to run the tests from the describe block `Scene/MetadataClassProperty > "normalize and unnormalize"` but when I run `--includeName 'normalize and unnormalize'` (being very careful to use strong quotes in bash to avoid escaping the spaces) or even just `--includeName normalize`, 0 tests execute so I think this test filter doesn't look at nested describe blocks. Meanwhile in the SpecRunner, `?spec=normalize and unnormalize` works fine and gives me the 10 specs I want to run.

...I suppose I could always do `fdescribe()`, though I've gotten used to the search query workflow that doesn't involve fiddling with the code (and accidentally committing it in the process)

Talking about this with @ebogo1, one of the limitations is that `--includeName` only has access to the top-level tests, not the inner describe blocks. That's unfortunate.

@j9liu you had a few more limitations of using `npm run test`, could you add the details in a comment here?

IanLilleyT:
GUI is useful because I never remember the name of the unit I'm debugging. It's usually some randomly constructed sentence describing the behavior. (Check out all the unit test descriptions -- there's no consistency, and I'm not sure there should be either). The html make it easy to find what I'm looking for.

Just my initial thoughts. I haven't used the new system much but I will be in the coming days

ggetz:
I appreciate you both writing up your thoughts. As a more meta comment, it would be great of we could move discussions like this to the forum, to engage more of the community. We have [a thread already for discussion around the testing changes](https://community.cesium.com/t/changes-to-unit-test-debugging-in-cesiumjs/17537).

Besides the fact that before we were clicking a link to a static page and now we're running a command to start a server, everything we were able to do with the standalone runner, we should be able to do through karma. We're still using jasmine just like before, just changing the form factor of how it's run.

[...]

PR #10238, opened 2022-03-25
Add `karma-jasmine-html-reporter` 

ebogo1:
This should address some of the concerns in https://github.com/CesiumGS/cesium/issues/10200. The `kjhtml` reporter adds a clickable list of tests after they finish running. Since the Karma Chrome window auto-closes, this is mostly useful when running tests with `npm run test -- --debug`. 

One thing I wasn't sure about was how this interacts with headless Chrome on CI - I ran CI a couple times and the build time seemed reasonable but it's possible it's more correct to disable the html spec reporter on CI runs.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/html-reporter/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ebogo1:
@ggetz Sounds good - I added a `--html` flag that will push the html reporter to the karma config if provided. Should be good for another look.

Specs/customizeJasmine.js
```diff
@@ -126,6 +126,7 @@ function customizeJasmine(
     window.webglStub = true;
   }
 
+  env.hideDisabled(true);
   //env.catchExceptions(true);
```
ggetz, line 130: I know this is unrelated, but's let's go ahead and remove this line.

ggetz (commented):
@ebogo1 

* You mentioned there's an option to disable the dots, right? Can we disable them by default? This seem a bit silly since we have so many tests:

image

* Also is it possible to show the entire spec list rather than just the failing specs? I believe that's a big appeal of the html reporter.

* Does it make sense to use `--html` ever _without_ the `--debug` option? It doesn't seem like there is much benefit. If so, we should probably update the documentation to just say `npm run test -- --html --debug` by default.

* We are hiding a lot of these deprecation warnings when running from the command line due to suppressing much of the output. Not for this PR, we'll probably want to clean these up soon, especially if we want to upgrade to the latest jasmine.

image

mramato:
> Since the Karma Chrome window auto-closes, 

This is because we tell it to auto-close, we also tell it to run the tests as soon as it opens.  All of these things are configurable.

ebogo1:
> This is because we tell it to auto-close, we also tell it to run the tests as soon as it opens. All of these things are configurable.

@mramato Currently the `--debug` flag added in https://github.com/CesiumGS/cesium/pull/10170 disables auto-close with `singleRun`. That's why using the HTML reporter without the debug option might not make much sense (per @ggetz's [comment](https://github.com/CesiumGS/cesium/pull/10238#pullrequestreview-923216003)). 

In [db28d2e](https://github.com/CesiumGS/cesium/pull/10238/commits/db28d2e1b8e8f5aa88d0bbc0e0399191b8aaaa12) I added a `"ChromeDebug"` launcher to our Karma config to open the `/debug.html` page on launch (requested by @IanLilleyT [here](https://github.com/CesiumGS/cesium/issues/10200#issuecomment-1081993732)). The `--debug` flag will always enable the HTML reporter and use the new launcher.

Other loose ends - 

> You mentioned there's an option to disable the dots, right? Can we disable them by default? 

I have not found a clean way to do this since the dots come from a CSS file in the node module folder..  

> Also is it possible to show the entire spec list rather than just the failing specs? I believe that's a big appeal of the html reporter.

This took me too long to figure out but the "Spec List" text at the bottom is a button that lets you toggle between the full list and failures:

image

From Ian's comment in the related issue: 

> Have Karma watch files automatically instead of doing npm run build whenever there are code changes. That way I can keep the Karma window open and refresh the page whenever I make changes.

Karma has `autoWatch` and `restartOnFileChange` flags ([docs](http://karma-runner.github.io/6.3/config/configuration-file.html)) but both of these cause the every enabled test suite to re-run. I haven't been able to get code changes to show up in the Karma window by refreshing it yet and am not sure if it's possible.

mramato:
@ebogo1 I don't have all the answers off the top of my head, but I'm confident there is a good solution that works for all of our needs here. What I think is missing is a clear definition of what our goals are for unit testing overall. What do we want the test and debugging experience to be like?

I feel like we are just trying to glob onto the existing jasmine/karma configuration and what we really need to do is take a step back, make sure we understand how all of the tools interact with each other, and then re-implement the karma/jasmine configuration from scratch, deferring to the "old" configuration as a reference.  We should be able to clearly explain why every option is there and what its effect is. 

For example, in the extreme case, `jasmine-browser-runner` did not exist when we added karma. It's how jasmine recommends you run browser-based tests.  Have we evaluated it at all as a replacement for karma?

ebogo1:
@mramato (and as a note to anyone following this PR) - `jasmine-browser-runner` depends on Jasmine 4.0. Upgrading was something we planned on doing anyway so I'm now testing a different branch to resolve the deprecation warnings we see on 3.99 (almost all of them are to do with updating our custom equality matcher syntax). If updating to 4.0 is easy enough, I'll open a PR for the update and try out `jasmine-browser-runner` in earnest. If that works out then I'll close this PR and focus on the other one.

> What I think is missing is a clear definition of what our goals are for unit testing overall. What do we want the test and debugging experience to be like?

There's some urgency to this since the current setup has some extra developer friction compared to the old SpecRunner (https://github.com/CesiumGS/cesium/issues/10200), but our main goals are to have: 
- A quick way to filter specs by name and suite (with similar filtering as the previous web runner had, ideally without having to write/maintain our own comparison code)
- An intuitive process for stepping through unit tests in a debugger (we plan on maintaining a VSCode task to do so, but we’d also like to support a similar workflow via the browser as before as an alternative)
- An easy way to browse all tests visually

ggetz:
Before we removed the old spec runner, [we made sure to ask the community about the implications](https://community.cesium.com/t/changes-to-unit-test-debugging-in-cesiumjs/17537/8) including the change to debug workflow and filtering. After it was removed, it looks like this came down to being too drastic a change in developer workflow preference. If we're not able to use the web runner to address the above points, we should be able to restore the standalone jasmine runner (and we should be able to use a more up-to-date version).

PR #10208, opened 2022-03-18
Fixed private tag for defer.js

IanLilleyT:
This fixes a warning when doing `npm run generateDocumentation`:

```
WARNING: The @private tag does not permit a value; the value will be ignored. File: defer.js, line: 24
WARNING: The @private tag does not permit a value; the value will be ignored. File: defer.js, line: 29
```
See https://app.travis-ci.com/github/CesiumGS/cesium/builds/248010115 for an example

PR #10261, opened 2022-04-01
Added option for generating private documentation

IanLilleyT:
How to use:
```
npm run generateDocumentation -- --private
```

This will generate documentation for all `@private` elements. It's intended for local development purposes only.

This PR also adds the `--pedantic` flag which treats warnings as errors. See https://jsdoc.app/about-commandline.html. This would have caught https://github.com/CesiumGS/cesium/pull/10208. Are there any downsides to adding this flag?


IanLilleyT:
Yep, just for verification purposes. Trying to make sure I'm being careful

ggetz:
Cool, works for me. Thanks @IanLilleyT !

Issue #10258, opened 2022-04-01
spec failure: renders terrain with lambertDiffuseMultiplier


jiangheng90:
`renders terrain with lambertDiffuseMultiplier` will fail when I run test in afternoon(China)
I think it need to set a sunlight direction

ptrgags:
Hi @jiangheng90, does the error message look like the one in https://github.com/CesiumGS/cesium/issues/10168 (failing to load a texture) or does this give a different error?

Issue #10168, opened 2022-03-07
Globe `lambertDiffuseMultiplier` test failing/timing out


ptrgags:
In the latest `main`, one unit test is either failing or timing out all of a sudden.

![image](https://user-images.githubusercontent.com/8422414/157065137-38fe36f5-7e8f-4b78-b3d9-e4d24f4127d7.png)

error in the console (thanks @ebogo1):
```
An error occurred in "SingleTileImageryProvider": Failed to load image Data/Images/Red16x16.png
```

[Unit test link](localhost:8080/Specs/SpecRunner.html?spec=Scene%2FGlobe%20renders%20terrain%20with%20lambertDiffuseMultiplier)

Browser: 

Chrome 98.0.4758.80
Firefox 97.0.2

Operating System: Ubuntu 18.04

lilleyse:
Fixed in https://github.com/CesiumGS/cesium/pull/10263#issuecomment-1086262447

PR #10263, opened 2022-04-01
Adds lighting to lambertDiffuseMultiplier spec

sanjeetsuhag:
This fixes https://github.com/CesiumGS/cesium/issues/10258 and https://github.com/CesiumGS/cesium/issues/10168

### DIFF
--- a/Specs/Scene/GlobeSpec.js
+++ b/Specs/Scene/GlobeSpec.js
@@ -7,6 +7,7 @@ import { Color } from "../../Source/Cesium.js";
 import { Cartesian3 } from "../../Source/Cesium.js";
 import { HeadingPitchRoll } from "../../Source/Cesium.js";
 import { NearFarScalar } from "../../Source/Cesium.js";
+import { JulianDate } from "../../Source/Cesium.js";
 import createScene from "../createScene.js";
 import pollToPromise from "../pollToPromise.js";
 
@@ -315,6 +316,10 @@ describe(
     });
 
     it("renders terrain with lambertDiffuseMultiplier", function () {
+      const renderOptions = {
+        scene: scene,
+        time: new JulianDate(2557522.0),
+      };
       globe.enableLighting = true;
       globe.dynamicAtmosphereLighting = true;
       globe.dynamicAtmosphereLightingFromSun = true;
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/lambert-test-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
This doesn't seem like the right fix. The test should work even if atmosphere is disabled.

lilleyse:
Could you try setting the time of day as suggested in https://github.com/CesiumGS/cesium/issues/10258? Depending on the light direction the lamber diffuse multiplier will have no effect.

```
float diffuseIntensity = clamp(czm_getLambertDiffuse(czm_lightDirectionEC, normalize(v_normalEC)) * u_lambertDiffuseMultiplier + 0.3, 0.0, 1.0);
```

sanjeetsuhag:
@lilleyse Thanks for the suggestion, committed the fix.

Specs/Scene/GlobeSpec.js
```diff
@@ -315,7 +316,13 @@ describe(
     });
 
     it("renders terrain with lambertDiffuseMultiplier", function () {
+      const renderOptions = {
+        scene: scene,
+        time: new JulianDate(2557522.0),
+      };
       globe.enableLighting = true;
+      globe.dynamicAtmosphereLighting = true;
+      globe.dynamicAtmosphereLightingFromSun = true;
```
lilleyse, line 325: Can these be removed now?

Specs/Scene/GlobeSpec.js
```diff
@@ -315,7 +316,13 @@ describe(
     });
 
     it("renders terrain with lambertDiffuseMultiplier", function () {
+      const renderOptions = {
+        scene: scene,
+        time: new JulianDate(2557522.0),
+      };
```
lilleyse, line 322: Could you also do this in the test above? It's written very similarly.

sanjeetsuhag:
Committed the fixes.

So, from the syntax of these, what I'm guessing is that you only use `toRenderAndCall` when you want to read back the RGBA value. Otherwise, just use `expect().toRender` or `expect().notToRender`.

PR #10271, opened 2022-04-04
Support morph targets in `ModelExperimental`

j9liu:
This PR adds support for morph targets in `ModelExperimental`. Since animations are not yet supported, the main model used for testing is `MorphPrimitivesTest`. It should show up in this [sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=dVRNc9owEP0rGp9Mh8imQJICYZohgR6SKROY9uKLsBesiSx5JJmPdvLfu5aAEEouHmv19u3ue2unShpL1hw2oMkdkbAhIzC8KugvFwuTIHXnkZKWcQk6CZrkbyIJMbmqRHYvecEs9IjVFTQT+dboJzKRqeMtleGWK4nMe9YR0xbfmGzTpVbFA6w0gAlrvqvW1zaNbzqd69a3Zh3odGjcjds38XUia1bPmZdnjf4AlnG5mnKb5i9KCEe2v3tmNqdWvSCCSRO2buOGo47985R3ybeQjTUrYK4Ru1S6eG/7GDJUqJQJj1PjY84EUBpmlXbVk0AqbXOUyp82YGwS+GqHepUWyJ8E0ayE1EQPzLLoWWUgTDQRdvmkWAYaI7rMp5oXqOMazByJopWYjy9d0BXmJcFJjaLmwyreX2pS7JKWxyTKsuyDWjX8cVsCIgDtFs6jupvQOU5IXaFXt970Z1cARdZ827ugVX5mzalioWcgxyVpHgLo8PF9z/koBC+N4hn9PZnddo7XF0zzV97mt8ZBc9co1djPboozcQPU5iDDZSVTt6GhQzTIftC9Yikya4aeq9d7+3nHn20jk6v3OQ9b92GwswW96nQbJyDf9UJVsuaclTlooBqxlSFfSPcAbOxn7vuZ+0EzGBi7EzA8IL7zosSNrJ0LKY0sFKXAr9ZEiyp9BUtTY3w2IYPoNHWQ8TXh2d2F3wBJBTMGb5aVEDP+B5JgOIgQ/1+qUE6Tn2vQgu1qWN4aPvkgpXQQ4fFyplVKLJg+Y/4H) with the same 3D offsets that it has in the [glTF-Sample-Viewer](https://github.khronos.org/glTF-Sample-Viewer-Release/).

**TODO**: the unit test I added to `ModelExperimental` doesn't reliably pass. It passes when `fit`-ed, but fails when `npm run test -- --includeName ModelExperimental` is run, but passes for the regular `npm run test`. I'll look into this later.

@ptrgags - can you review?


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-experimental-morph-targets/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelExperimental/MorphTargetsPipelineStage.js
```diff
@@ -0,0 +1,255 @@
...
+
+  shaderBuilder.addDefine(
+    "HAS_MORPH_TARGETS",
+    undefined,
+    ShaderDestination.VERTEX
+  );
+
+  addGetMorphedAttributeFunctionDeclarations(shaderBuilder);
+
+  for (let i = 0; i < primitive.morphTargets.length; i++) {
```
ptrgags, line 56: Hm, I see why this is a bit tricky to refactor, the morph targets are stored by index, then by semantic. However the shader code cares about the semantic first, then the index. Is there a reasonable way to reorganize this?

Specs/Scene/ModelExperimental/MorphTargetsPipelineStageSpec.js
```diff
@@ -0,0 +1,238 @@
...
+
+      ShaderBuilderTester.expectHasVertexUniforms(shaderBuilder, [
+        "uniform float u_morphWeights[2];",
+      ]);
+
+      const uniformMap = renderResources.uniformMap;
+      expect(uniformMap.u_morphWeights()).toBe(primitive.morphWeights);
+    });
+  });
+});
```
ptrgags, line 238: Oh, since this uses GltfLoader (which may create GPU resources along the way), we should mark this as a WebGL test.
```suggestion
}, "WebGL");
```

We really should do a sweep through the pipeline stages at some point to see what test suites might be missing this. Not everyone needs it, but some do.

ptrgags (commented):
@j9liu Overall the approach looks good, though I see some places where this could be refactored

ptrgags:
>Since we plan to bring ModelExperimental to production, should this PR close that issue?

Yeah, Once this one is merged, we could close that one and say this was implemented in the new `ModelExperimental`, which will supersede `Model` later this year.

ptrgags:
Also regarding the test failures, I haven't noticed any issues when I run this locally, but I'll run it again just to make sure it's not a fluke.

Specs/Scene/ModelExperimental/ModelExperimentalSpec.js
```diff
@@ -45,6 +45,8 @@ describe(
     const boxBackFaceCullingUrl =
       "./Data/Models/Box-Back-Face-Culling/Box-Back-Face-Culling.gltf";
     const boxBackFaceCullingOffset = new HeadingPitchRange(Math.PI / 2, 0, 2.0);
+    const morphPrimitivesTestUrl =
+      "./Data/Models/GltfLoader/MorphPrimitivesTest/glTF/MorphPrimitivesTest.gltf";
```
ptrgags, line 49: I see we only really use the glTF version of these assets. I would suggest two things:

1. Add a test for the one quantized model, that _should_ work since `KHR_mesh_quantization` is applied well before we get to the morph targets stage, but a unit test will let us prove it
2. The other models (like the binary ones) that are unused can be removed from this PR (please squash your commits to keep unused binary files from the history)

j9liu: For now the quantized model won't work because it's animated, but I'll leave it in the directory for future testing.

j9liu:
@ptrgags updated!

Source/Scene/ModelExperimental/MorphTargetsPipelineStage.js
```diff
@@ -0,0 +1,240 @@
...
+  const uniformMap = {
+    u_morphWeights: function () {
+      return primitive.morphWeights;
+    },
+  };
+
+  renderResources.uniformMap = combine(uniformMap, renderResources.uniformMap);
+};
+
+const scratchAttributeInfo = {};
```
ptrgags, line 103: I would explicitly initialize this with the two fields explicitly set to undefined. More readable and _might_ make it easier for a JS engine to optimize it as a fixed-size structure.

Source/Scene/ModelExperimental/MorphTargetsPipelineStage.js
```diff
@@ -0,0 +1,240 @@
...
+) {
+  const shaderBuilder = renderResources.shaderBuilder;
+
+  addMorphTargetAttributeToRenderResources(
+    renderResources,
+    attribute,
+    attributeIndex
+  );
+
+  getMorphTargetAttributeInfo(attribute, scratchAttributeInfo);
```
ptrgags, line 119: This function should return the attribute info, so it should be

`const attributeInfo = getMorphTargetAttributeInfo(...)` and then use the value in the function call below

ptrgags (commented):
@j9liu almost there, noticed a couple small tweaks needed.

ptrgags:
@j9liu changes look good, I'm merging!

Issue #10246, opened 2022-03-30
3D Tiles inspector debug labels not working


lilleyse:
The 3D Tiles inspector debug labels aren't working in main. `git bisect` points to https://github.com/CesiumGS/cesium/commit/31c154a299927af69c9a66415b2e451d96d10dbb (Remove when.js and replace with native promises). Marking as `next release` since it's a regression.

[Local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVHLbsIwEPwVK6cgIadVK1RBQK3oBamoUkt7ysXYC1h1vJHXCY+q/944CYIWfPKuZ2Zn1hIteVZp2IJjY2Zhy6ZAusz5Z9OLs0g29RStF9qCy6I++84sY2uDSxiylTAEmf3pjTKb2VaJw86DVXEn1Tbb4u55oQ3QzFIB0qOb6522gSsbJ/rYD+PnqMDUrjpReVWgUW+QjYFWxgcI+BOXJFjghdO59roC4kKp+CztH3Pg4yZh6czwCJihfQPC0kngK4f5E9WwmYrvbx4Gg17I3y4gs5cR+MlOdxtdhdEGtx8uJPauhPN9HhDzBcYdO0yK+lFKfm9gEpyG86jzAp0PrmPOEw95YYQHSpal/ALPJVEgBmianFNTpSum1fjKVzNpBFH9siqNedcHyKJJmtT4C6pBobRdv1bgjNgH2OZ28tI2OedpUpfXmR7RLIX7p/wL)

Before|After
--|--
![before](https://user-images.githubusercontent.com/915398/160889616-ed4c026f-9c16-401a-9329-16e809bb2817.png)|![after](https://user-images.githubusercontent.com/915398/160889614-c4014e57-f343-4483-a665-c1b9726a6ee2.jpg)

sanjeetsuhag:
From a quick look, I found that the draw commands for the billboards are being created, however, they aren't actually being rendered.

| Working (`1.91`) | Not Working (`main`) |
| - | - |
| ![Screen Shot 2022-04-01 at 9 34 32 AM](https://user-images.githubusercontent.com/5172619/161274276-ea34dbc2-1556-4e39-b09a-fa1b1617efc2.png) | ![Screen Shot 2022-04-01 at 9 34 35 AM](https://user-images.githubusercontent.com/5172619/161274297-77e270b5-8821-4714-a431-7763a0198381.png) |

PR #10275, opened 2022-04-05
Fix label rendering for tileset debug labels

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/10246

The issue was that labels and billboards, due to the fact that native promise result at the end of the frame instead of synchronously like when promises, were no longer created synchronously. Tileset debug labels are all removed during `update` and then replaced with new labels, leaving no time for the labels to actually render.

This PR allows billboards which do not need asynchronous loading to proceed synchronously through the normal update flow, removing the need to wait on any promise.

@sanjeetsuhag Can you please review?


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/Billboard.js
```diff
@@ -1142,7 +1142,37 @@ Billboard.prototype._loadImage = function () {
...
 
+  const that = this;
+  function completeImageLoad(index) {
+    if (
+      that._imageId !== imageId ||
+      that._image !== image ||
+      !BoundingRectangle.equals(that._imageSubRegion, imageSubRegion)
+    ) {
+      // another load occurred before this one finished, ignore the index
+      return;
```
sanjeetsuhag, line 1153: I am trying to understand why we need to exhaustively check all three conditions here: is it because they can change independently?

Also, just double checking: we need this check because this function is not module scoped and can be triggered by an image?

Source/Scene/TextureAtlas.js
```diff
@@ -74,6 +74,7 @@ function TextureAtlas(options) {
   this._textureCoordinates = [];
   this._guid = createGuid();
   this._idHash = {};
+  this._indexHash = {};
```
sanjeetsuhag, line 77: It's a little unfortunate that we need to have two objects keeping track here: `id` -> `indexHash[id]` but I suppose we just need the `TextureAtlas` operates with the `index`. Is that correct?

sanjeetsuhag (commented):
I can confirm that this does seem to fix the problem with the debug labels.

| [`main`](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/main/Apps/Sandcastle/index.html#c=bVJdb9swDPwrQp4cIJVSpMWALA3WpQ8LsK5FG2wY4BdHYmxismRItNN02H+f5I8g2fJkk7o7kkcKwT7XqBWanKmMMiZt7Qj8gdkd+/ZzxZ4qMA/xobKOMj1nBVE1F2K/319zc5A8t43wSCCURSKBBgkzwga8mKmrbS/OqyyH1EhrPLEGYQ+O3TEDe7YCj3XJv7e5JB3JNl5ZQxkacOlown6nhjEC50Lm2dkGFbj5QJQOMoIf1mm16SDJeJKaP+OPqUlNV4p7CQZ4ru0WuIKKig14us8D2FPPCu2Qq+GUBW8ERiV9oS7ZBbOHDWrwa+MrkGTdI76hiRW7AXHIx6kerQId1HtReVGgVW+RbQOdDEUI0LlRZw0AJa05tdNHQ9bWvIAPa5TAd86W9z7A1ir5cDu7mR2tOTOmclhiuzOeKZX0dTsH+4AHm9UhuF+iB04FmGRXG0loDUvG3Yp6zXdry41NYoYNM0y66GSOL0EvHMYzkixeMpNDT2BsyqeT4f9qym+PwdDK1tYmcl+rAhxwF4RqzwS74dMOOo6f2H476mgyWng6aFgOQp+wjOccbUs4FwRlpcMRebGt5a9QQHofiRG6EKfUhcKGobq7cKZM6sz78LKrtX7Fd0hHy4UI+P+o2raTPzXgdHaIsOJ6+bVLcs4XIoSXmWSt3mbuH+W/) | [`labels-fix`](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/labels-fix/Apps/Sandcastle/index.html#c=bVJdb9swDPwrQp4cIJVSpMWALA3WpQ8LsK5FG2wY4BdHYmxismRItNN02H+f5I8g2fJkk7o7kkcKwT7XqBWanKmMMiZt7Qj8gdkd+/ZzxZ4qMA/xobKOMj1nBVE1F2K/319zc5A8t43wSCCURSKBBgkzwga8mKmrbS/OqyyH1EhrPLEGYQ+O3TEDe7YCj3XJv7e5JB3JNl5ZQxkacOlown6nhjEC50Lm2dkGFbj5QJQOMoIf1mm16SDJeJKaP+OPqUlNV4p7CQZ4ru0WuIKKig14us8D2FPPCu2Qq+GUBW8ERiV9oS7ZBbOHDWrwa+MrkGTdI76hiRW7AXHIx6kerQId1HtReVGgVW+RbQOdDEUI0LlRZw0AJa05tdNHQ9bWvIAPa5TAd86W9z7A1ir5cDu7mR2tOTOmclhiuzOeKZX0dTsH+4AHm9UhuF+iB04FmGRXG0loDUvG3Yp6zXdry41NYoYNM0y66GSOL0EvHMYzkixeMpNDT2BsyqeT4f9qym+PwdDK1tYmcl+rAhxwF4RqzwS74dMOOo6f2H476mgyWng6aFgOQp+wjOccbUs4FwRlpcMRebGt5a9QQHofiRG6EKfUhcKGobq7cKZM6sz78LKrtX7Fd0hHy4UI+P+o2raTPzXgdHaIsOJ6+bVLcs4XIoSXmWSt3mbuH+W/) |
| - | - |
| main | labels-fix |

I just have a couple comments, or clarifications, regarding the changes. Additionally, since this bug was not caught in a test. should one be added?

Source/Scene/TextureAtlas.js
ggetz: > but I suppose we just need the TextureAtlas operates with the index

I'm not sure exactly what you mean. But in the case where we are asynchronously adding an image, we do need to be able to return the `Promise`, not just the resulting `index`.

Source/Scene/Billboard.js
ggetz: Honestly I'm not entirely sure. As far as I can see, `imageId` and `image` cannot change independently. But this just moves the logic removed below to a function so we can finish early if there is no promise to resolve, so I prefer not to modify the existing logic.

ggetz:
Thanks @sanjeetsuhag!

There are some existing `show tile debug labels ...` tests in `Cesium3dTilesetSpec` that test if the commands have been added. However it's difficult to test whether they actually render the labels themselves without access some very internal aspects of labels and billboards. For example `expect(scene).toRender(...)` isn't able to catch whether the labels are rendering.

sanjeetsuhag:
> However it's difficult to test whether they actually render the labels themselves without access some very internal aspects of labels and billboards.

Ah, I see. Maybe that's something we should look into in the future, but this fix is good for now. I've tested all the relevant Sancastles for Billboards and Labels and they seem to be working correctly. Thanks @ggetz!

PR #10276, opened 2022-04-05
Update to Jasmine 4.0.1

ebogo1:
Most deprecation warnings we have are from `customEqualityTesters` no longer being passed as an argument. Instead, they're now added to the `MatchersUtil` object before tests run (see [docs](https://jasmine.github.io/tutorials/upgrading_to_Jasmine_4.0#matchers-cet) for more info). Removing the argument from our matchers and `equals.js` should work for everything as is except for `toEqualEpsilon`, which has some changes:

- Uses its own `equalityTester` function instead of `equals.js` in case it passes or fails; this mimics Jasmine's behavior of trying custom matchers before the built-in `equals` check.
- First the tester function converts `a` and `b` to mimic the behavior in `equals.js`. 
- Adds a loop over `a`'s object keys if the arguments are not arrays and don't have `equalsEpsilon` member functions; this loop recursively calls the equality tester on keys that both `a` and `b` have.
- If none of the early returns are hit, checks for exact equality using `equals.js`

There is one last deprecation warning because of an empty describe in [Cesium3DTilesInspectorSpec](https://github.com/CesiumGS/cesium-analytics/blob/4c5a2824c29c29d36ff6510b42f33c0b0fb4437d/Specs/Widgets/Cesium3DTilesInspector/Cesium3DTilesInspectorSpec.js#L57) -

```
'DEPRECATION: describe with no children (describe() or it()) is deprecated and will be removed in a future version of Jasmine. Please either remove the describe or add children to it.
    at <Jasmine>
    at window.describe (http://localhost:9876/base/Specs/customizeJasmine.js:36:5)
    at Proxy.<anonymous> (http://localhost:9876/base/Specs/Widgets/Cesium3DTilesInspector/Cesium3DTilesInspectorSpec.js?e3399e5268c227836e31512241a0d2df36611c45:57:5)
    at <Jasmine>
Note: This message will be shown only once. Set the verboseDeprecations config property to true to see every occurrence.'
```

I wasn't sure right away how we wanted to handle this describe block - @ggetz any thoughts on this one?

### DIFF
--- a/Specs/Core/BingMapsGeocoderServiceSpec.js
+++ b/Specs/Core/BingMapsGeocoderServiceSpec.js
@@ -76,7 +76,7 @@ describe("Core/BingMapsGeocoderService", function () {
     });
   });
 
-  it("returns no geocoder results if Bing has no results", function (done) {
+  it("returns no geocoder results if Bing has no results", function () {
     const query = "some query";
     const data = {
       resourceSets: [],
@@ -89,13 +89,12 @@ describe("Core/BingMapsGeocoderService", function () {
       deferred.resolve(data);
     };
     const service = new BingMapsGeocoderService({ key: "" });
-    service.geocode(query).then(function (results) {
+    return service.geocode(query).then(function (results) {
       expect(results.length).toEqual(0);
-      done();
     });
   });
 
-  it("returns no geocoder results if Bing has results but no resources", function (done) {
+  it("returns no geocoder results if Bing has results but no resources", function () {
     const query = "some query";
     const data = {
       resourceSets: [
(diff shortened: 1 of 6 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/jasmine-4.0/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
> There is one last deprecation warning because of an empty describe in [Cesium3DTilesInspectorSpec](https://github.com/CesiumGS/cesium-analytics/blob/4c5a2824c29c29d36ff6510b42f33c0b0fb4437d/Specs/Widgets/Cesium3DTilesInspector/Cesium3DTilesInspectorSpec.js#L57) -

From the file history, it looks like the tests just got removed over time and no one cleaned up the describe block. I'm going to go ahead and remove it.

PR #10279, opened 2022-04-07
Removes when.js file from Source/ThirdParty

sanjeetsuhag:
In https://github.com/CesiumGS/cesium/pull/10186, the `when.js` was accidentally added back to `Source/ThirdParty`. This PR removes that file.


ggetz:
Thanks @sanjeetsuhag! My oversight, apologies.

PR #10282, opened 2022-04-08
Support skinning in `ModelExperimental`

j9liu:
This PR adds support for glTF 2.0 skins in `ModelExperimental`. Although `ModelExperimental` currently does not support animations, the skinned mesh should at least appear in its default, unanimated position, without any incorrect transformations. For example, Cesium Man should look the same in the [ModelExperimental 3D Models sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Model%20Experimental%203D%20Models.html&label=3D%20Tiles%20Next).

Some notes:
- Although skins are assigned at the `node` level, the shader code is based on the number of joint and weight attributes there are in the primitive. I checked through the specs to see if this number was required to be consistent across primitives, but it doesn't seem to disallow two primitives with the same skin to have different numbers of `JOINTS_n` attributes. Thus, it makes more sense for `SkinningPipelineStage` to be processed at the primitive level.
- I'm not sure what kind of unit test to add to `ModelExperimental` for this -- the model will render regardless of its skin. Let me know if there are any ideas.
- I anticipate that `updateJointMatrices` in `ModelExperimentalSkin` and `ModelExperimentalNode` will need to be called for animations, but I haven't figured out where / when, so it's not included in the `ModelExperimental` update code for the time being.

@ptrgags - could you review?

### DIFF
--- a/Source/Scene/GltfLoader.js
+++ b/Source/Scene/GltfLoader.js
@@ -1440,7 +1440,9 @@ function loadSkin(loader, gltf, gltfSkin, nodes) {
         skin.inverseBindMatrices = inverseBindMatrices;
       });
     }
-  } else {
+  }
+
+  if (!defined(skin.inverseBindMatrices)) {
     skin.inverseBindMatrices = arrayFill(
       new Array(jointsLength),
       Matrix4.IDENTITY
--- a/Source/Scene/GltfLoader.js
+++ b/Source/Scene/GltfLoader.js
@@ -1417,11 +1417,13 @@ function loadSkin(loader, gltf, gltfSkin, nodes) {
   }
   skin.joints = joints;
 
+  let useDefaultBindMatrices = true;
   const inverseBindMatricesAccessorId = gltfSkin.inverseBindMatrices;
   if (defined(inverseBindMatricesAccessorId)) {
     const accessor = gltf.accessors[inverseBindMatricesAccessorId];
     const bufferViewId = accessor.bufferView;
     if (defined(bufferViewId)) {
+      useDefaultBindMatrices = false;
       const bufferViewLoader = loadBufferView(loader, gltf, bufferViewId);
       bufferViewLoader.promise.then(function (bufferViewLoader) {
         if (loader.isDestroyed()) {
@@ -1442,7 +1444,7 @@ function loadSkin(loader, gltf, gltfSkin, nodes) {
     }
   }
 
-  if (!defined(skin.inverseBindMatrices)) {
+  if (useDefaultBindMatrices) {
     skin.inverseBindMatrices = arrayFill(
       new Array(jointsLength),
       Matrix4.IDENTITY
--- a/Source/Scene/GltfLoader.js
+++ b/Source/Scene/GltfLoader.js
@@ -1333,52 +1333,10 @@ function loadInstanceFeaturesLegacy(
   }
 }
 
-function loadSkin(loader, gltf, gltfSkin, nodes) {
-  const skin = new Skin();
-
-  const jointIds = gltfSkin.joints;
-  const jointsLength = jointIds.length;
-  const joints = new Array(jointsLength);
-  for (let i = 0; i < jointsLength; ++i) {
-    joints[i] = nodes[jointIds[i]];
-  }
-  skin.joints = joints;
-
-  const inverseBindMatricesAccessorId = gltfSkin.inverseBindMatrices;
-  if (defined(inverseBindMatricesAccessorId)) {
-    const accessor = gltf.accessors[inverseBindMatricesAccessorId];
-    const bufferViewId = accessor.bufferView;
-    if (defined(bufferViewId)) {
-      const bufferViewLoader = loadBufferView(loader, gltf, bufferViewId);
-      bufferViewLoader.promise.then(function (bufferViewLoader) {
-        if (loader.isDestroyed()) {
-          return;
-        }
-        const bufferViewTypedArray = bufferViewLoader.typedArray;
-        const packedTypedArray = getPackedTypedArray(
-          gltf,
-          accessor,
-          bufferViewTypedArray
-        );
-        const inverseBindMatrices = new Array(jointsLength);
-        for (let i = 0; i < jointsLength; ++i) {
-          inverseBindMatrices[i] = Matrix4.unpack(packedTypedArray, i * 16);
-        }
-        skin.inverseBindMatrices = inverseBindMatrices;
-      });
-    }
-  } else {
-    skin.inverseBindMatrices = arrayFill(
-      new Array(jointsLength),
-      Matrix4.IDENTITY
-    );
-  }
-
-  return skin;
-}
-
 function loadNode(loader, gltf, gltfNode, supportedImageFormats, frameState) {
   const node = new Node();
+
+  node.name = gltfNode.name;
 
   node.matrix = fromArray(Matrix4, gltfNode.matrix);
   node.translation = fromArray(Cartesian3, gltfNode.translation);
@@ -1424,13 +1382,15 @@ function loadNodes(loader, gltf, supportedImageFormats, frameState) {
   const nodesLength = gltf.nodes.length;
   const nodes = new Array(nodesLength);
   for (i = 0; i < nodesLength; ++i) {
-    nodes[i] = loadNode(
+    const node = loadNode(
       loader,
       gltf,
       gltf.nodes[i],
       supportedImageFormats,
       frameState
     );
+    node.index = i;
+    nodes[i] = node;
   }
 
   for (i = 0; i < nodesLength; ++i) {
@@ -1443,14 +1403,78 @@ function loadNodes(loader, gltf, supportedImageFormats, frameState) {
     }
   }
 
+  return nodes;
+}
+
+function loadSkin(loader, gltf, gltfSkin, nodes) {
+  const skin = new Skin();
+
+  const jointIds = gltfSkin.joints;
+  const jointsLength = jointIds.length;
+  const joints = new Array(jointsLength);
+  for (let i = 0; i < jointsLength; ++i) {
+    joints[i] = nodes[jointIds[i]];
+  }
+  skin.joints = joints;
+
+  const inverseBindMatricesAccessorId = gltfSkin.inverseBindMatrices;
+  if (defined(inverseBindMatricesAccessorId)) {
+    const accessor = gltf.accessors[inverseBindMatricesAccessorId];
+    const bufferViewId = accessor.bufferView;
+    if (defined(bufferViewId)) {
+      const bufferViewLoader = loadBufferView(loader, gltf, bufferViewId);
+      bufferViewLoader.promise.then(function (bufferViewLoader) {
+        if (loader.isDestroyed()) {
+          return;
+        }
+        const bufferViewTypedArray = bufferViewLoader.typedArray;
+        const packedTypedArray = getPackedTypedArray(
+          gltf,
+          accessor,
+          bufferViewTypedArray
+        );
+        const inverseBindMatrices = new Array(jointsLength);
+        for (let i = 0; i < jointsLength; ++i) {
+          inverseBindMatrices[i] = Matrix4.unpack(packedTypedArray, i * 16);
+        }
+        skin.inverseBindMatrices = inverseBindMatrices;
+      });
+    }
+  } else {
+    skin.inverseBindMatrices = arrayFill(
+      new Array(jointsLength),
+      Matrix4.IDENTITY
+    );
+  }
+
+  return skin;
+}
+
+function loadSkins(loader, gltf, nodes) {
+  let i;
+
+  const gltfSkins = gltf.skins;
+  if (!defined(gltfSkins)) {
+    return [];
+  }
+
+  const skinsLength = gltf.skins.length;
+  const skins = new Array(skinsLength);
+  for (i = 0; i < skinsLength; ++i) {
+    const skin = loadSkin(loader, gltf, gltf.skins[i], nodes);
+    skin.index = i;
+    skins[i] = skin;
+  }
+
+  const nodesLength = nodes.length;
   for (i = 0; i < nodesLength; ++i) {
     const skinId = gltf.nodes[i].skin;
     if (defined(skinId)) {
-      nodes[i].skin = loadSkin(loader, gltf, gltf.skins[skinId], nodes);
+      nodes[i].skin = skins[skinId];
     }
   }
 
-  return nodes;
+  return skins;
 }
 
 function loadStructuralMetadata(
@@ -1518,6 +1542,7 @@ function parse(loader, gltf, supportedImageFormats, frameState) {
   }
 
   const nodes = loadNodes(loader, gltf, supportedImageFormats, frameState);
+  const skins = loadSkins(loader, gltf, nodes);
   const scene = loadScene(gltf, nodes);
 
   const components = new Components();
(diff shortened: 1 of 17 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-experimental-skinning/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
```diff
@@ -173,6 +194,59 @@ Object.defineProperties(ModelExperimentalSceneGraph.prototype, {
 function initialize(sceneGraph) {
   const components = sceneGraph._modelComponents;
   const scene = components.scene;
+
+  computeModelMatrix(sceneGraph);
+
+  const nodes = components.nodes;
+  const nodesLength = nodes.length;
+  sceneGraph._runtimeNodes = new Array(nodesLength);
```
ptrgags, line 202: If I'm reading this code correctly, this array has room for all the nodes in the glTF, but the way the scene graph is traversed, only the ones used in the main `scene` are populated?

Probably okay, but it would be good to put a comment explaining this.

Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
```diff
@@ -450,6 +494,43 @@ ModelExperimentalSceneGraph.prototype.updateModelMatrix = function () {
...
+      node.transformToRoot,
+      scratchObjectSpace
+    );
+
+    const computedJointMatrices = node.computedJointMatrices;
+    const joints = node.joints;
+    const inverseBindMatrices = node.inverseBindMatrices;
+
+    for (let m = 0; m < inverseBindMatrices.length; ++m) {
+      // [joint-matrix] = [node-to-root^-1][joint-to-root][inverse-bind]
```
ptrgags, line 516: minor formatting tweak

```suggestion
      // jointMatrix = nodeToRoot^(-1) * jointToRoot * inverseBind
```

Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
```diff
@@ -450,6 +494,43 @@ ModelExperimentalSceneGraph.prototype.updateModelMatrix = function () {
...
+const scratchObjectSpace = new Matrix4();
+
+ModelExperimentalSceneGraph.prototype.updateJointMatrices = function () {
+  const model = this._model;
+  const skinnedNodes = model._skinnedNodes;
+
+  for (let i = 0; i < skinnedNodes.length; i++) {
+    const node = skinnedNodes[i];
+
+    const objectSpace = Matrix4.inverseTransformation(
```
ptrgags, line 506: If I'm reading this correctly, `objectSpace` corresponds to `nodeToRoot^(-1)` in the matrix equation below? we should rename one of them so they match. 

... or maybe rename this matrix `rootToNode`? simpler than `inverseNodeToRoot`

Source/Shaders/ModelExperimental/SkinningStageVS.glsl
```diff
@@ -0,0 +1,18 @@
...
+{
+    mat4 skinningMatrix = getSkinningMatrix();
+    mat3 skinningMatrixMat3 = mat3(skinningMatrix);
+
+    vec4 positionMC = vec4(attributes.positionMC, 1.0);
+    attributes.positionMC = vec3(skinningMatrix * positionMC);
+
+    #ifdef HAS_NORMALS
+    vec3 normalMC = attributes.normalMC;
+    attributes.normalMC = skinningMatrixMat3 * normalMC;
```
ptrgags, line 11: Just to check my understanding, this works because the normal matrix is applied later in the Geometry stage, right?

Specs/Scene/ModelExperimental/SkinningPipelineStageSpec.js
```diff
@@ -0,0 +1,183 @@
...
+        ShaderBuilderTester.expectHasVertexFunction(
+          shaderBuilder,
+          SkinningPipelineStage.FUNCTION_ID_GET_SKINNING_MATRIX,
+          SkinningPipelineStage.FUNCTION_SIGNATURE_GET_SKINNING_MATRIX,
+          [
+            "    mat4 skinnedMatrix = mat4(0);",
+            "    skinnedMatrix += a_weights_0.x * u_jointMatrices[int(a_joints_0.x)];",
+            "    skinnedMatrix += a_weights_0.y * u_jointMatrices[int(a_joints_0.y)];",
+            "    skinnedMatrix += a_weights_0.z * u_jointMatrices[int(a_joints_0.z)];",
+            "    skinnedMatrix += a_weights_0.w * u_jointMatrices[int(a_joints_0.w)];",
```
ptrgags, line 104: Just to make sure I read the spec correctly, even though there's two joints, this still works because `a_weights_0.z`, `a_weights_0.w` would both be 0 in the glTF, correct?

Source/Scene/ModelExperimental/SkinningPipelineStage.js
```diff
@@ -0,0 +1,120 @@
...
+      attribute.semantic === VertexAttributeSemantic.JOINTS ||
+      attribute.semantic === VertexAttributeSemantic.WEIGHTS;
+
+    if (!isJointsOrWeights) {
+      continue;
+    }
+
+    if (attribute.setIndex > setIndex) {
+      setIndex = attribute.setIndex;
+    }
```
ptrgags, line 79: This could be simplified with `Math.max()`

Specs/Scene/ModelExperimental/SkinningPipelineStageSpec.js
```diff
@@ -0,0 +1,183 @@
...
+
+        const runtimeNode = renderResources.runtimeNode;
+        const uniformMap = renderResources.uniformMap;
+        expect(uniformMap.u_jointMatrices()).toBe(
+          runtimeNode.computedJointMatrices
+        );
+      });
+    });
+
+    it("processes skin with many joints", function () {
```
ptrgags, line 124: Hm... anywhere we can get a small glTF with more than 4 joints? Would be good to have a test of that, unless it's a hassle to find such a model.

Specs/Scene/ModelExperimental/SkinningPipelineStageSpec.js
```diff
@@ -0,0 +1,183 @@
...
+        attributes: [],
+        shaderBuilder: new ShaderBuilder(),
+        attributeIndex: 1,
+        model: {
+          type: ModelExperimentalType.TILE_GLTF,
+        },
+        runtimeNode: {
+          computedJointMatrices: mockJointMatrices,
+        },
+      };
```
ptrgags, line 81: If we add a third test for a model with >4 joints, consider wrapping this setup code in a function. Our mocking code tends to get messy. See `mockRenderResources` in `FeatureIdPipelineStage`

ptrgags (commented):
@j9liu not quite done reviewing the code yet, but I did a first pass of feedback

Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
j9liu: whoops, that was an artifact of copying `Model.js` code. thanks for the catch

Source/Scene/ModelExperimental/ModelExperimentalSceneGraph.js
j9liu: Yes -- I can see how this would be unideal for glTFs with multiple scenes, but I'm not sure how to refactor this in a way that doesn't involve a map of indices the glTF to indices in the runtime array (unless that is preferred?)

Source/Shaders/ModelExperimental/SkinningStageVS.glsl
j9liu: Yes

Specs/Scene/ModelExperimental/SkinningPipelineStageSpec.js
j9liu: Yes. The [Skins](https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_020_Skins.md) section of the glTF Tutorial repo might help illustrate this since it breaks down an actual example.

Source/Scene/GltfLoader.js
```diff
@@ -1443,14 +1403,78 @@ function loadNodes(loader, gltf, supportedImageFormats, frameState) {
...
+  for (let i = 0; i < jointsLength; ++i) {
+    joints[i] = nodes[jointIds[i]];
+  }
+  skin.joints = joints;
+
+  const inverseBindMatricesAccessorId = gltfSkin.inverseBindMatrices;
+  if (defined(inverseBindMatricesAccessorId)) {
+    const accessor = gltf.accessors[inverseBindMatricesAccessorId];
+    const bufferViewId = accessor.bufferView;
+    if (defined(bufferViewId)) {
```
ptrgags, line 1424: Looking at this code (and I see this is from before this PR), what should happen if this inner if statement is false? According to [the spec](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#_accessor_bufferview) it looks like an accessor is initialized to zeros unless it's a sparse accessor. 

I know sparse accessors are beyond the scope of this PR, but should this be handled like the else clause below?

Source/Scene/ModelExperimental/ModelExperimentalNode.js
```diff
@@ -246,3 +280,42 @@ ModelExperimentalNode.prototype.configurePipeline = function () {
 
   updateStages.push(ModelMatrixUpdateStage);
 };
+
+/**
+ * Updates the joint matrices for this node, where each matrix is computed as
+ * [computedJointMatrix] = [nodeWorldTransform]^-1 * [skinjointMatrix].
```
ptrgags, line 286:
```suggestion
 * [computedJointMatrix] = [nodeWorldTransform]^-1 * [skinJointMatrix].
```

Source/Scene/ModelExperimental/ModelExperimentalSkin.js
```diff
@@ -0,0 +1,176 @@
...
+
+  return result;
+}
+
+/**
+ * Updates the joint matrices for the skin.
+ *
+ * @private
+ */
+ModelExperimentalSkin.prototype.updateJointMarices = function () {
```
ptrgags, line 168:
```suggestion
ModelExperimentalSkin.prototype.updateJointMatrices = function () {
```

Source/Scene/ModelExperimental/ModelExperimentalSkin.js
```diff
@@ -0,0 +1,176 @@
...
+   */
+  joints: {
+    get: function () {
+      return this._joints;
+    },
+  },
+
+  /**
+   * The joint matrices for the skin, where each joint matrix is computed as
+   * [jointMatrix] = [jointWorldTransform] * [bindMatrix].
```
ptrgags, line 103: wait, `bindMatrix` is really `inverseBindMatrix` based on the code below? 

I would use `inverseBindMatrix` everywhere to be unambiguous

Source/Scene/ModelExperimental/SkinningPipelineStage.js
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * This pipeline stage processes the joint matrices of a skinned primitive, adding
+ * the relevant functions and uniforms to the shaders. The joint and weight attributes
+ * themselves are processed in the geometry pipeline stage.
+ *
+ * Processes a primitive. This stage modifies the following parts of the render resources:
+ * <ul>
+ *  <li> adds the uniform declaration for the joint matrices in the vertex shader
+ *  <li> adds the function to compute the skinning matrix in the vertex shader
```
ptrgags, line 29:
```suggestion
 *  <li> adds the uniform declaration for the joint matrices in the vertex shader</li>
 *  <li> adds the function to compute the skinning matrix in the vertex shader</li>
```

Source/Scene/ModelExperimental/SkinningPipelineStage.js
```diff
@@ -0,0 +1,118 @@
...
+ * </ul>
+ *
+ * @param {PrimitiveRenderResources} renderResources The render resources for this primitive.
+ * @param {ModelComponents.Primitive} primitive The primitive.
+ * @private
+ */
+
```
ptrgags, line 36: remove this blank line

Source/Scene/ModelExperimental/SkinningPipelineStage.js
```diff
@@ -0,0 +1,118 @@
...
+    "mat4",
+    `u_jointMatrices[${jointMatrices.length}]`,
+    ShaderDestination.VERTEX
+  );
+
+  shaderBuilder.addVertexLines([SkinningStageVS]);
+
+  const uniformMap = {
+    u_jointMatrices: function () {
+      return jointMatrices;
```
ptrgags, line 56: I doubt this would change at runtime, but nevertheless it would be good practice to use `runtimeNode.jointMatrices` to be more future-proof. This way the latest value is always used, not a constant.

Source/Scene/ModelExperimental/SkinningPipelineStage.js
```diff
@@ -0,0 +1,118 @@
...
+  );
+
+  let setIndex;
+  let componentIndex;
+  const componentStrings = ["x", "y", "z", "w"];
+  const maximumSetIndex = getMaximumAttributeSetIndex(primitive);
+  for (setIndex = 0; setIndex <= maximumSetIndex; setIndex++) {
+    for (componentIndex = 0; componentIndex <= 3; componentIndex++) {
+      const component = componentStrings[componentIndex];
+      const line = `skinnedMatrix += a_weights_${setIndex}.${component} * u_jointMatrices[int(a_joints_${setIndex}.${component})];`;
```
ptrgags, line 103: For lines like this with a lot of variable substitution, a comment with a concrete example can be helpful:

```
// Example: skinnedMatrix += a_weights_0.x * u_jointMatrices[int(a_joints_0.x)];
```

ptrgags (commented):
@j9liu left some more comments

Specs/Scene/ModelExperimental/SkinningPipelineStageSpec.js
ptrgags: https://github.com/cx20/gltf-test -- just as one last thing to check, is there anything in here besides the official sample models?

if we can't find anything, I wouldn't worry about this.

Source/Scene/GltfLoader.js
j9liu: Yeah, let's handle it like the else clause for now.

j9liu:
@ptrgags updated! Let me know if anything else needs tweaking

j9liu:
@ptrgags realized I handled something wrong so I just pushed a change. For now I have a `useDefaultBindMatrices` variable to catch if there is no accessor.

This behavior might change later -- `inverseBindMatrices` is marked optional in the spec, and some implementations generate their own bind matrices if this accessor doesn't exist. I'm wondering if Cesium should at some point do the same.

Specs/Scene/ModelExperimental/SkinningPipelineStageSpec.js
j9liu: No, I don't see any animated models that aren't already included in the samples model repo I linked above.

ptrgags:
@j9liu looks good, I ran the tests again and checked the existing Sandcastles to ensure existing examples still load and render properly.

PR #10288, opened 2022-04-12
Fix 3dtiles single content failed with no error message

jiangheng90:
I found in sometime ```Cesium3Dtile.singleContentFailed``` called with no error message and throw new error. so I add error message in case ```handleTileFailure``` in ```Cesium3DTileset``` can call with error message.
```javascript
function handleTileFailure(tileset, tile) {
  return function (error) {
    const url = tile._contentResource.url;
    if (tileset.tileFailed.numberOfListeners > 0) {
      tileset.tileFailed.raiseEvent({
        url: url,
        message: message,
      });
    } else {
      console.log(`A 3D tile failed to load: ${url}`);
      console.log(`Error: ${message}`);
    }
  };
}
```
so the ```message``` in ```handleTileFailure``` will not throw ```error``` undefined


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-3dtiles-singleContentFailed-with-no-error-message/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #10289, opened 2022-04-12
fix Cesium3DTile singleContentFailed have no error

jiangheng90:
First time pull request, I have many useless commit, so I close the previouse pr.
So back to the point. This problem is found by a modify in 3dtiles traversal. 
when ```Cesium3DTile.requestSingleContent``` called, the code will go into ```  if (tile.isDestroyed()) {```
```javascript
// Cesium3DTile.js
promise
    .then(function (arrayBuffer) {
      if (tile.isDestroyed()) {
        // Tile is unloaded before the content finishes loading
        singleContentFailed(tile, tileset);
        return;
      }

      const content = makeContent(tile, arrayBuffer);

      if (expired) {
        tile.expireDate = undefined;
      }

      tile._content = content;
      tile._contentState = Cesium3DTileContentState.PROCESSING;
      tile._contentReadyToProcessPromise.resolve(content);
      --tileset.statistics.numberOfPendingRequests;

      return content.readyPromise.then(function (content) {
        if (tile.isDestroyed()) {
          // Tile is unloaded before the content finishes processing
          singleContentFailed(tile, tileset,);
          return;
        }
      ........
```
the third params is empty, so ```handleTileFailure``` in ```Cesium3DTilset```
```javascript
function handleTileFailure(tileset, tile) {
  return function (error) {
    const url = tile._contentResource.url;
    if (tileset.tileFailed.numberOfListeners > 0) {
      tileset.tileFailed.raiseEvent({
        url: url,
        message: message,
      });
    } else {
      console.log(`A 3D tile failed to load: ${url}`);
      console.log(`Error: ${message}`);
    }
  };
}
```

the ```error``` will be undefined and throw error, so I just add params in ``` singleContentFailed```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -315,3 +315,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Justin Peter](https://github.com/themagicnacho)
 - [Gu Miao](https://github.com/Gu-Miao)
 - [Shen WeiQun](https://github.com/ShenWeiQun)
+- [Jiang Heng](https://github.com/jiangheng90)
--- a/Source/Scene/Cesium3DTile.js
+++ b/Source/Scene/Cesium3DTile.js
@@ -1197,7 +1197,11 @@ function requestSingleContent(tile) {
     .then(function (arrayBuffer) {
       if (tile.isDestroyed()) {
         // Tile is unloaded before the content finishes loading
-        singleContentFailed(tile, tileset);
+        singleContentFailed(
+          tile,
+          tileset,
+          "Tile is unloaded before the content finishes loading"
+        );
         return;
       }
 
@@ -1215,7 +1219,11 @@ function requestSingleContent(tile) {
       return content.readyPromise.then(function (content) {
         if (tile.isDestroyed()) {
           // Tile is unloaded before the content finishes processing
-          singleContentFailed(tile, tileset);
+          singleContentFailed(
+            tile,
+            tileset,
+            "Tile is unloaded before the content finishes processing"
+          );
           return;
         }
         updateExpireDate(tile);


TJPovey:
Could we also add a catch block to ```tile.contentReadyToProcessPromise``` in **Cesium3DTileset.js**? The catch block is required as error aren't being caught once it is rejected.

It's closely related to your PR which is why I think it makes sense to include it here. See my recent discussion on the forums:
https://community.cesium.com/t/upgrading-to-cesium-1-92-handletilefailure-error-undefined-tile-is-unloaded-before-the-content-finishes-loading/18138/2

I'm not sure what should be within the catch block, we don't want to raise the tile failed event here since it is raised when ```contentReadyPromise``` fails, where both the promises are rejected at the same point:

```
function singleContentFailed(tile, tileset, error) {
  if (tile._contentState === Cesium3DTileContentState.PROCESSING) {
    --tileset.statistics.numberOfTilesProcessing;
  } else {
    --tileset.statistics.numberOfPendingRequests;
  }
  tile._contentState = Cesium3DTileContentState.FAILED;
  tile._contentReadyPromise.reject(error);
  tile._contentReadyToProcessPromise.reject(error);
}
...
function multipleContentFailed(tile, tileset, error) {
  // note: The Multiple3DTileContent handles decrementing the number of pending
  // requests if the state is LOADING.
  if (tile._contentState === Cesium3DTileContentState.PROCESSING) {
    --tileset.statistics.numberOfTilesProcessing;
  }

  tile._contentState = Cesium3DTileContentState.FAILED;
  tile._contentReadyPromise.reject(error);
  tile._contentReadyToProcessPromise.reject(error);
}
```

Maybe just an empty function call in the then block will be enough, since the error is being handled within ``` tile._contentReadyPromise```:

```
//Cesium3DTileset.js
    tile.contentReadyToProcessPromise
      .then(addToProcessingQueue(tileset, tile), function() {});
```

jiangheng90:
@TJPovey Hi, I think my pr may not be your problem you meet.This pr just fix the error undefine. But your problem is your code cause     ‘’Tile is unloaded before the content finishes processing’’.

PR #10290, opened 2022-04-12
Parse animations in `GltfLoader`

j9liu:
This PR modifies `ModelComponents` and `GltfLoader` such that animation data can be loaded into `ModelComponents `from the glTF.

**Question**: I have a `loadAccessor` function in `GltfLoader` that basically acts as a wrapper for the `loadBufferView` code. I made this because:
- loading the accessor for the `inverseBindMatrices` in `loadSkin()` was basically the same as loading the `input` and `output` accessors for `loadAnimationSampler()`
- the key difference is what to do with the `packedTypedArray` in the promise, and the default behavior if the accessor is invalid

I'm not proud of the way I wrote the `loadAccessor` function, so any feedback is appreciated. I wanted to circumvent writing a whole  `GltfAccessorLoader` class (which would also require modifying `ResourceCache` and `ResourceCacheKey`), while being able to write something reusable.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/gltfloader-animation-parsing/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
The structure of the PR looks good but I agree that `loadAccessor` doesn't feel quite right.

According to the glTF spec when `accessor.bufferView` is undefined it must be initialized with zeros. So I think it's fine if `loadAccessor` returns a promise to a typed array filled with zeros if the buffer view is undefined. Then the callbacks won't be needed.

To simplify the code further `loadAccessor` could return an array of the unpacked type (e.g. `Number`, `Cartesian3`, `Matrix4`) instead of a typed array. That way the calling code doesn't have to do the unpacking, instead it can pass the type to `loadAccessor` to let it do the unpacking.

Source/Scene/GltfLoader.js
```diff
@@ -1504,6 +1578,116 @@ function loadStructuralMetadata(
...
+      animationSampler.interpolation = InterpolationType.STEP;
+      break;
+    case "CUBICSPLINE":
+      animationSampler.interpolation = InterpolationType.CUBICSPLINE;
+      break;
+    case "LINEAR":
+    default:
+      animationSampler.interpolation = InterpolationType.LINEAR;
+      break;
+  }
```
ptrgags, line 1599: since the enum names are an exact match, you can use:

```
animationSampler.interpolation = defaultVaue(
  InterpolationType[gltfInterpolation], 
  InterpolationType.LINEAR
);
```

I've seen other examples in the codebase like this, such as these lines from `MetadataEnum`:

```
const valueType = defaultValue(
  MetadataComponentType[enumDefinition.valueType],
  MetadataComponentType.UINT16
);  
```

Source/Scene/GltfLoader.js
```diff
@@ -1504,6 +1578,116 @@ function loadStructuralMetadata(
...
+
+  const nodeIndex = gltfTarget.node;
+  // If the node isn't defined, the animation channel should be ignored.
+  // It's easiest to signal this by returning undefined.
+  if (!defined(nodeIndex)) {
+    return undefined;
+  }
+
+  animationTarget.node = nodes[nodeIndex];
+  switch (gltfTarget.path) {
```
ptrgags, line 1618: similarly here for the path enum

ptrgags (commented):
@j9liu this is heading in the right direction. I pointed out some things that could be simplified a bit.

Source/Scene/ModelComponents.js
```diff
@@ -810,6 +810,156 @@ function Scene() {
...
+   *
+   * @type {ModelComponents.InterpolationType}
+   * @private
+   */
+  this.interpolation = undefined;
+
+  /**
+   * The keyframe data of the animation.
+   *
+   * @type {Number[]|Cartesian3[]|Quaternion[]}
```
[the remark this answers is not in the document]
j9liu: Morph weights are different from the joint weights attributes (`WEIGHTS_n`), which are the `vec4` attributes you're thinking of. The morph weights depend on how many morph targets the model has, so they can't be cleanly grouped into a `Cartesian` and have to be left as plain numbers.

PR #8903, opened 2020-06-03
More typescript fixes

thw0rted:
## Make `Event` generic and add a callback signature in a few places

There are a lot more places this could be used but it should be non-breaking and just allows for future refinement.

## `ConstructorOptions` for `Resource`

I can't remember why this one got taken out?

## `exportKml` overload actually works

The catch is that I had to exclude it from `prettier` until they address https://github.com/prettier/prettier/issues/7724.  The overloads do show up separately in the docs but I think it looks OK that way.

## Another stab at making PropertyBag actually work

I pinged https://github.com/jsdoc/jsdoc/issues/1285 for an update.  All the SO questions, etc, that I can find about how to describe something like PropertyBag (a class with its own methods and properties, and also an index signature for "other" properties) say the only option is an intersection type, which JSDoc refuses to support on the basis that it's not part of Closure Compiler.  (The questions actually all say "this is an antipattern, don't do this in the first place" but it's probably too late to change?)

I think the hack I came up with will work for now.  Basically, the class is called `PropertyBag` so that it gets the right name during `generateDocumentation`, but for Typescript purposes it's renamed to `PropertyBagBase` with a regex replace, then a top level type alias is declared using union types.

The one thing I wasn't able to do is figure out how to document in JSDoc that `PropertyBag` has an index signature.  Is there a more Javascript-y name for "index signature"?  I think I would have called it the "index operator" back before I learned TS, but I can't find anything online using either name.  This isn't a new problem -- the docs have never actually given an example of using e.g. `ent.properties.someProperty` -- but since I was in there anyway I was looking to improve it.  If you can find the right tag to use, by all means let's add it.

Issue #10265, opened 2022-04-03
When selecting 'Show tile coordinates' imagery provider crashes


MuhammedKalkan:
Happens to be on the  sandcastle 1.92

```
An error occurred while rendering.  Rendering has stopped.
TypeError: imagePromise.then is not a function
TypeError: imagePromise.then is not a function
    at doRequest (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:298518:10)
    at ImageryLayer._requestImagery (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:298526:5)
    at Imagery.processStateMachine (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:297392:25)
    at TileImagery.processStateMachine (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:297468:20)
    at GlobeSurfaceTile.processImagery (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:296642:47)
    at Function.GlobeSurfaceTile.processStateMachine (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:296575:46)
    at GlobeSurfaceTileProvider.loadTile (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:301855:22)
    at processSinglePriorityLoadQueue (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:306894:20)
    at processTileLoadQueue (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:306853:22)
    at QuadtreePrimitive.endFrame (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:305966:5)
```

ggetz:
Hi @MuhammedKalkan. Is this one of the default Sandcastle examples, or your own? Can you provide a link or code sample to reproduce?

MuhammedKalkan:
Hey Gabby,

Default samples behave the same way. Just add inspector mixin and try to show tile coordinates. Here is a link for you

[...]

PR #9783, opened 2021-09-01
 Fix buildModuleUrl.setBaseUrl and .getCesiumBaseUrl in JSDoc/TypeScript

puxiao:
**Issue** [8922](https://github.com/CesiumGS/cesium/issues/8922) [8924](https://github.com/CesiumGS/cesium/issues/8924)

**Description**

If we use TypeScript, we will encounter the following problems.

```
import { buildModuleUrl } from 'cesium'

buildModuleUrl.setBaseUrl('./static/Cesium/')
buildModuleUrl.getCesiumBaseUrl()
```

PROBLEMS:

```
Property 'setBaseUrl' does not exist on type '(relativeUrl: string) => string'.
Property 'getCesiumBaseUrl' does not exist on type '(relativeUrl: string) => string'.
```

<br>

**Why?**

*Source/Core/buildModuleUrl.js*

```
/**
 * Given a relative URL under the Cesium base URL, returns an absolute URL.
 * @function
 *
 * ...
 */
function buildModuleUrl(relativeUrl) {
  ...
}



/**
 * Sets the base URL for resolving modules.
 * @param {String} value The new base URL.
 */
buildModuleUrl.setBaseUrl = function (value) {
  ...
};

/**
 * Gets the base URL for resolving modules.
 */
buildModuleUrl.getCesiumBaseUrl = getCesiumBaseUrl;
```

We can see that:

1. `buildModuleUrl` is a function
2. `buildModuleUrl` also is a namespace

When we execute the command `build-ts`, this is very confusing for `JSDoc`.

*Cesium.d.ts*

```
/**
 * Given a relative URL under the Cesium base URL, returns an absolute URL.
 * @example
 * ...
 */
export function buildModuleUrl(relativeUrl: string): string;
```

We can see that `buildModuleUrl.setBaseUrl` and `buildModuleUrl.getCesiumBaseUrl` is not exported.

<br>

**Wrong solution**

I have tried but failed.

```diff
+ /** @namespace buildModuleUrl */

 /**
  * @function
  */
 function buildModuleUrl(relativeUrl) {
   ...
 }


 /**
+ * @memberof buildModuleUrl
  * ...
  */
 buildModuleUrl.setBaseUrl = function (value) {
   ...
 };

 /**
+ * @memberof buildModuleUrl
  * ...
  */
 buildModuleUrl.getCesiumBaseUrl = getCesiumBaseUrl;
```

`yarn run build-ts`

*Cesium.d.ts*

```
export namespace buildModuleUrl {
[...]

Issue #8922, opened 2020-06-05
buildModuleUrl not exported in TypeScript


jeanpul:
Hello,

From Cesium 1.7

import { buildModuleUrl } from 'cesium'; 
leads to Module '"cesium"' has no exported member 'buildModuleUrl'

to fix it I've changed Cesium.d.ts :

// in the declare module "cesium" part

export function buildModuleUrl(value: string);

export namespace buildModuleUrl {    

    function setBaseUrl(value: string): void;

}

// after the declare module "cesium" part

declare module "cesium/Source/Core/buildModuleUrl" { import { buildModuleUrl } from 'cesium'; export default buildModuleUrl; }
 
thx

mramato:
Thanks for bringing this up @jeanpul!

Oddly enough, `buildModuleUrl` was never part of the public API and has always been a private function. This is now obvious because TypeScript only includes the officially supported API.

That being said, the fact that it is not part of the official API is indeed a bug.  We even use it in Sandcastle examples.  I'll open a PR to add JSDoc for it and this fix will go out early next week for a 1.70.1 patch release that we want to put out specifically to improve these new TS definitions.

Thanks again.

jeanpul:
Hello,

I've checked the 1.70.1 release and in the Cesium.d.ts still have a problem with buildModuleUrl :
The generated signature is 
export function buildModuleUrl(relativeUrl: string): void;
but should be
export function buildModuleUrl(relativeUrl: string): string;
in order to returns string

Thx

mramato:
Thanks @jeanpul I pushed 638c6ebffe86a17637acc8dc5c223f404131f993 directly to master with the fix and it will go out with our next release.

ggetz:
Re-opening this as there issues with the current docs for generating TS definitions for `setBaseUrl` and `getCesiumBaseUrl`. See https://github.com/CesiumGS/cesium/pull/9783 and https://github.com/CesiumGS/cesium/pull/10292#discussion_r849323954

PR #10292, opened 2022-04-13
TypeScript fixes again

thw0rted:
This is an updated branch for #8903.  I hate to lose the conversation from there but submitting these changes as a single commit allowed me to significantly reduce the number of files in the patchset -- a lot of them were only related to `when.js` Promises, and that's no longer relevant.  Hopefully this makes it easier to review and merge.

### DIFF
--- a/Source/Core/HeightmapTerrainData.js
+++ b/Source/Core/HeightmapTerrainData.js
@@ -500,9 +500,8 @@ HeightmapTerrainData.prototype.interpolateHeight = function (
  * @param {Number} descendantX The X coordinate within the tiling scheme of the descendant tile for which we are upsampling.
  * @param {Number} descendantY The Y coordinate within the tiling scheme of the descendant tile for which we are upsampling.
  * @param {Number} descendantLevel The level within the tiling scheme of the descendant tile for which we are upsampling.
- * @returns {Promise.<HeightmapTerrainData>|undefined} A promise for upsampled heightmap terrain data for the descendant tile,
- *          or undefined if too many asynchronous upsample operations are in progress and the request has been
- *          deferred.
+ * @returns {HeightmapTerrainData|undefined} The upsampled heightmap terrain data for the descendant tile,
+ *          or undefined if the mesh is unavailable.  Note that this provider always returns synchronously.
  */
 HeightmapTerrainData.prototype.upsample = function (
   tilingScheme,
--- a/Source/Core/ScreenSpaceEventHandler.js
+++ b/Source/Core/ScreenSpaceEventHandler.js
@@ -934,12 +934,13 @@ function ScreenSpaceEventHandler(element) {
   registerListeners(this);
 }
 
+// TODO: document callback signature for each type of event
 /**
  * Set a function to be executed on an input event.
  *
  * @param {Function} action Function to be executed when the input event occurs.
- * @param {Number} type The ScreenSpaceEventType of input event.
- * @param {Number} [modifier] A KeyboardEventModifier key that is held when a <code>type</code>
+ * @param {ScreenSpaceEventType} type The ScreenSpaceEventType of input event.
+ * @param {KeyboardEventModifier} [modifier] A KeyboardEventModifier key that is held when a <code>type</code>
  * event occurs.
  *
  * @see ScreenSpaceEventHandler#getInputAction
@@ -966,8 +967,8 @@ ScreenSpaceEventHandler.prototype.setInputAction = function (
 /**
  * Returns the function to be executed on an input event.
  *
- * @param {Number} type The ScreenSpaceEventType of input event.
- * @param {Number} [modifier] A KeyboardEventModifier key that is held when a <code>type</code>
+ * @param {ScreenSpaceEventType} type The ScreenSpaceEventType of input event.
+ * @param {KeyboardEventModifier} [modifier] A KeyboardEventModifier key that is held when a <code>type</code>
  * event occurs.
  *
  * @returns {Function} The function to be executed on an input event.
@@ -989,8 +990,8 @@ ScreenSpaceEventHandler.prototype.getInputAction = function (type, modifier) {
 /**
  * Removes the function to be executed on an input event.
  *
- * @param {Number} type The ScreenSpaceEventType of input event.
- * @param {Number} [modifier] A KeyboardEventModifier key that is held when a <code>type</code>
+ * @param {ScreenSpaceEventType} type The ScreenSpaceEventType of input event.
+ * @param {KeyboardEventModifier} [modifier] A KeyboardEventModifier key that is held when a <code>type</code>
  * event occurs.
  *
  * @see ScreenSpaceEventHandler#getInputAction
--- a/Source/Core/buildModuleUrl.js
+++ b/Source/Core/buildModuleUrl.js
@@ -5,6 +5,7 @@ import Resource from "./Resource.js";
 
 /*global CESIUM_BASE_URL,define,require*/
 
+// TODO: all these should have JSDoc tags, or mark as private
 const cesiumScriptRegex = /((?:.*\/)|^)Cesium\.js(?:\?|\#|$)/;
 function getBaseUrlFromCesiumScript() {
   const scripts = document.getElementsByTagName("script");
--- a/Source/DataSources/CallbackProperty.js
+++ b/Source/DataSources/CallbackProperty.js
@@ -104,8 +104,8 @@ CallbackProperty.prototype.equals = function (other) {
  * A function that returns the value of the property.
  * @callback CallbackProperty.Callback
  *
- * @param {JulianDate} [time] The time for which to retrieve the value.
- * @param {Object} [result] The object to store the value into, if omitted, a new instance is created and returned.
- * @returns {Object} The modified result parameter or a new instance if the result parameter was not supplied or is unsupported.
+ * @param {JulianDate} time The time for which to retrieve the value.
+ * @param {Object} [result] The object to store the value into. If omitted, the function must create and return a new instance.
+ * @returns {Object} The modified result parameter, or a new instance if the result parameter was not supplied or is unsupported.
  */
 export default CallbackProperty;
--- a/Source/DataSources/CompositePositionProperty.js
+++ b/Source/DataSources/CompositePositionProperty.js
@@ -87,7 +87,7 @@ Object.defineProperties(CompositePositionProperty.prototype, {
  *
  * @param {JulianDate} time The time for which to retrieve the value.
  * @param {Object} [result] The object to store the value into, if omitted, a new instance is created and returned.
- * @returns {Object} The modified result parameter or a new instance if the result parameter was not supplied.
+ * @returns {Cartesian3 | undefined} The modified result parameter or a new instance if the result parameter was not supplied.
  */
 CompositePositionProperty.prototype.getValue = function (time, result) {
   return this.getValueInReferenceFrame(time, ReferenceFrame.FIXED, result);
@@ -99,7 +99,7 @@ CompositePositionProperty.prototype.getValue = function (time, result) {
  * @param {JulianDate} time The time for which to retrieve the value.
  * @param {ReferenceFrame} referenceFrame The desired referenceFrame of the result.
  * @param {Cartesian3} [result] The object to store the value into, if omitted, a new instance is created and returned.
- * @returns {Cartesian3} The modified result parameter or a new instance if the result parameter was not supplied.
+ * @returns {Cartesian3 | undefined} The modified result parameter or a new instance if the result parameter was not supplied.
  */
 CompositePositionProperty.prototype.getValueInReferenceFrame = function (
   time,
--- a/Source/DataSources/TimeIntervalCollectionPositionProperty.js
+++ b/Source/DataSources/TimeIntervalCollectionPositionProperty.js
@@ -57,6 +57,7 @@ Object.defineProperties(TimeIntervalCollectionPositionProperty.prototype, {
    * Gets the interval collection.
    * @memberof TimeIntervalCollectionPositionProperty.prototype
    * @type {TimeIntervalCollection}
+   * @readonly
    */
   intervals: {
     get: function () {
@@ -81,7 +82,7 @@ Object.defineProperties(TimeIntervalCollectionPositionProperty.prototype, {
  *
  * @param {JulianDate} time The time for which to retrieve the value.
  * @param {Object} [result] The object to store the value into, if omitted, a new instance is created and returned.
- * @returns {Object} The modified result parameter or a new instance if the result parameter was not supplied.
+ * @returns {Cartesian3 | undefined} The modified result parameter or a new instance if the result parameter was not supplied.
  */
 TimeIntervalCollectionPositionProperty.prototype.getValue = function (
   time,
@@ -96,7 +97,7 @@ TimeIntervalCollectionPositionProperty.prototype.getValue = function (
  * @param {JulianDate} time The time for which to retrieve the value.
  * @param {ReferenceFrame} referenceFrame The desired referenceFrame of the result.
  * @param {Cartesian3} [result] The object to store the value into, if omitted, a new instance is created and returned.
- * @returns {Cartesian3} The modified result parameter or a new instance if the result parameter was not supplied.
+ * @returns {Cartesian3 | undefined} The modified result parameter or a new instance if the result parameter was not supplied.
  */
 TimeIntervalCollectionPositionProperty.prototype.getValueInReferenceFrame = function (
   time,
--- a/Source/Scene/ImageryProvider.js
+++ b/Source/Scene/ImageryProvider.js
@@ -296,7 +296,7 @@ ImageryProvider.prototype.getTileCredits = function (x, y, level) {
  * @param {Number} y The tile Y coordinate.
  * @param {Number} level The tile level.
  * @param {Request} [request] The request object. Intended for internal use only.
- * @returns {Promise.<HTMLImageElement|HTMLCanvasElement>|undefined} A promise for the image that will resolve when the image is available, or
+ * @returns {HTMLImageElement|HTMLCanvasElement|ImageBitmap|Promise.<HTMLImageElement|HTMLCanvasElement|ImageBitmap>|undefined} A promise for the image that will resolve when the image is available, or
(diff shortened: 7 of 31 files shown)


cesium-concierge:
Thanks for the pull request @thw0rted!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/more-typescript-redo/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

thw0rted:
I just noticed that #10265 was called out in the changelog -- turns out that two of the "fixes" I made to the return type info should have been correct and it was the code that was wrong.  It might be worth looking at the function bodies for the other return-type changes I have in this PR to double check that there aren't other similar errors waiting to bite somebody.

Source/Core/HeightmapTerrainData.js
```diff
@@ -500,9 +500,8 @@ HeightmapTerrainData.prototype.interpolateHeight = function (
  * @param {Number} descendantX The X coordinate within the tiling scheme of the descendant tile for which we are upsampling.
  * @param {Number} descendantY The Y coordinate within the tiling scheme of the descendant tile for which we are upsampling.
  * @param {Number} descendantLevel The level within the tiling scheme of the descendant tile for which we are upsampling.
- * @returns {Promise.<HeightmapTerrainData>|undefined} A promise for upsampled heightmap terrain data for the descendant tile,
- *          or undefined if too many asynchronous upsample operations are in progress and the request has been
- *          deferred.
+ * @returns {HeightmapTerrainData|undefined} The upsampled heightmap terrain data for the descendant tile,
```
thw0rted, line 503: Likely another Promise bug like #10265

Source/Core/buildModuleUrl.js
```diff
@@ -5,6 +5,7 @@ import Resource from "./Resource.js";
 
 /*global CESIUM_BASE_URL,define,require*/
 
+// TODO: all these should have JSDoc tags, or mark as private
```
thw0rted, line 8: I think this TODO was intended to be attached to the two property-functions (?) attached to the exported `buildModuleUrl`.  Basically, you can `import buildModuleUrl from "./buildModuleUrl.js"`, then call `buildModuleUrl.setBaseUrl(...)` or `buildModuleUrl.getCesiumBaseUrl()`, but the JSDoc for two functions is not recognized and doesn't show up in either the generated "globals" documentation, nor in the TS definitions.

My personal recommendation would to refactor the module to stop using `export default` and produce three named exports instead, but that's a significant breaking change that the team would have to discuss.

Source/Scene/ImageryProvider.js
```diff
@@ -296,7 +296,7 @@ ImageryProvider.prototype.getTileCredits = function (x, y, level) {
  * @param {Number} y The tile Y coordinate.
  * @param {Number} level The tile level.
  * @param {Request} [request] The request object. Intended for internal use only.
- * @returns {Promise.<HTMLImageElement|HTMLCanvasElement>|undefined} A promise for the image that will resolve when the image is available, or
+ * @returns {HTMLImageElement|HTMLCanvasElement|ImageBitmap|Promise.<HTMLImageElement|HTMLCanvasElement|ImageBitmap>|undefined} A promise for the image that will resolve when the image is available, or
```
thw0rted, line 299: Per #10265 , is this an undesired change?  Or would it be better if the caller had to expect to handle both immediate value-returns as well as a returned Promise?  (If y'all switched over to async/await this would be a lot simpler...)

thw0rted: Also the issue would have been much more visible if you used actual ES6 classes with inheritance rather than having "ImageryProviders" that don't actually inherit from ImageryProvider, not sure if that's still on the table....

Source/Scene/ImageryProvider.js
```diff
@@ -338,6 +338,10 @@ ImageryProvider.prototype.pickFeatures = function (
 
 const ktx2Regex = /\.ktx2$/i;
 
+// TODO: This can definitely return a (Promise for) ImageBitmap, and also
```
thw0rted, line 341: I guess this is really a continuation of my question above.  Basically the documented return type is at least incomplete, and might actually be wrong as well.  If all ImageryProviders inherited from ImageryProvider, this would be caught automatically.

Source/Scene/SingleTileImageryProvider.js
```diff
@@ -454,10 +454,7 @@ SingleTileImageryProvider.prototype.getTileCredits = function (x, y, level) {
  * @param {Number} y The tile Y coordinate.
  * @param {Number} level The tile level.
  * @param {Request} [request] The request object. Intended for internal use only.
- * @returns {Promise.<HTMLImageElement|HTMLCanvasElement>|undefined} A promise for the image that will resolve when the image is available, or
- *          undefined if there are too many active requests to the server, and the request
- *          should be retried later.  The resolved image may be either an
- *          Image or a Canvas DOM object.
+ * @returns {HTMLImageElement|ImageBitmap} The resolved image
```
thw0rted, line 457: Probably the error from #10265 again

Source/Core/buildModuleUrl.js
ggetz: I'd appreciate if you can open a new issue for this. There's already some discussion around whether this function is intended for public use.

Source/DataSources/CallbackProperty.js
```diff
@@ -104,8 +104,8 @@ CallbackProperty.prototype.equals = function (other) {
  * A function that returns the value of the property.
  * @callback CallbackProperty.Callback
  *
- * @param {JulianDate} [time] The time for which to retrieve the value.
- * @param {Object} [result] The object to store the value into, if omitted, a new instance is created and returned.
- * @returns {Object} The modified result parameter or a new instance if the result parameter was not supplied or is unsupported.
+ * @param {JulianDate} time The time for which to retrieve the value.
```
ggetz, line 107: I think the `time` parameter _is_ optional.

Source/DataSources/CompositePositionProperty.js
```diff
@@ -87,7 +87,7 @@ Object.defineProperties(CompositePositionProperty.prototype, {
  *
  * @param {JulianDate} time The time for which to retrieve the value.
  * @param {Object} [result] The object to store the value into, if omitted, a new instance is created and returned.
```
ggetz, line 89: This should also be `Cartesian3`, right?

Source/Core/ScreenSpaceEventHandler.js
```diff
@@ -934,12 +934,13 @@ function ScreenSpaceEventHandler(element) {
   registerListeners(this);
 }
 
+// TODO: document callback signature for each type of event
```
ggetz, line 937: Do you plan on tackling this TODO?

Source/DataSources/TimeIntervalCollectionPositionProperty.js
```diff
@@ -57,6 +57,7 @@ Object.defineProperties(TimeIntervalCollectionPositionProperty.prototype, {
    * Gets the interval collection.
    * @memberof TimeIntervalCollectionPositionProperty.prototype
    * @type {TimeIntervalCollection}
+   * @readonly
```
ggetz, line 60: `@readonly` applies to `intervals` as well, correct?

Source/Scene/ImageryProvider.js
ggetz: I think we want to have each `ImageryProvider` implementation return a `Promise` or `undefined`. Yes, I agree it's difficult to enforce with the current state of things, but full migration to ES6 classes is a bit out of scope for this PR. 😅

ggetz (commented):
Thanks @thw0rted! I have a few comments. It appears that most of the typescript-specific question have been discussed back in https://github.com/CesiumGS/cesium/pull/8903, so mine comments are mostly about the documentation itself.

Source/Core/buildModuleUrl.js
thw0rted: Hrm.  I found #8922 and #9783 so it looks like your team was already looking at / working on this.  Maybe you should re-open one or both of those until the discussion is finished?

Source/DataSources/CallbackProperty.js
thw0rted: This is a kind of a mess.

Since this is a callback-function signature, what we're actually saying here is whether or not the function is guaranteed to be invoked with a time value, or whether the function must handle the case where `time` is undefined.  The callback is invoked by passing through the `time` argument when `getValue` is called.

[Here](https://github.com/thw0rted/cesium/blob/more-typescript-redo/Source/DataSources/CallbackProperty.js#L52) is the `time` argument for `CallbackProperty#getValue` -- it looks like this was erroneously copy-pasted from `ConstantProperty`, since obviously a ConstantProperty does not vary with time but a CallbackProperty (probably?) could.  *If* the `time` argument for CallbackProperty should in fact be required, then the callback signature can be marked as non-optional, and callbacks can safely assume `time` is not undefined.  My inclination is to resolve this by marking the `time` argument to `getValue` as required as well.  That would be in line with the [argument in the base Property interface](https://github.com/thw0rted/cesium/blob/more-typescript-redo/Source/DataSources/Property.js#L55), which is required.

Source/Scene/ImageryProvider.js
thw0rted: I'm trying to address this by exporting a typedef to represent all the types an ImageryProvider might return.  So, providers that punt to `ImageryProvider.loadImage` can just resolve as the typedef, but providers that always return one known type can be more specific in their return annotation.

Source/DataSources/CompositePositionProperty.js
thw0rted: Good catch!

Source/DataSources/TimeIntervalCollectionPositionProperty.js
thw0rted: You might have been looking at the next field, `referenceFrame` -- yes, that should also be `@readonly`, should be fixed in the changes I'm about to push.

Source/Core/ScreenSpaceEventHandler.js
thw0rted: It's a little complicated.  I'm making typedefs and a callback to describe the event types that can be passed to a callback, but the right way to solve it involves a number of overloads for `setInputAction`, at least one per type of callback (of which there are five).  JSDoc overloads don't work very well (see previous PR with `exportKML`), and cause issues with `prettier`.

For now, I'm going to push out a version with a type-union callback parameter.  I'll look at doing overloads but will probably hold off unless/until there's a better way to implement them.

thw0rted: OK I got this 100% working, I'm pretty sure, but it is going to take a little explaining.  Overloads in JSDoc are a bit hairy; there are pros and cons to my approach.

On the pro side, Typescript users can now write `viewer.screenSpaceEventHandler.setInputAction(evt => doStuff(Cartographic.fromCartesian(evt.position)), ScreenSpaceEventType.MOUSE_DOWN)` and the `position` property will be correctly typed.  This is pretty great!  It's the same way that TS lets you write e.g. `input.addEventListener("input", evt => doStuff(evt.inputType))` and get strong typing where `evt` is automatically an `InputEvent`.

On the con side, there's a bunch of extra typedefs and `@callback`s that shouldn't really need to be named.  It takes up extra space in the source file (though it's all comments so it should minify out for production) and adds some entries to the docs of questionable utility.

I did update the "container" template in the docs so that any function declaration with a `@variation` tag greater than 1 is not printed in the docs.  This avoids a problem described in the previous version of this PR where the overridden function would get one entire doc entry for each override -- `@variation 1` is the one that gets printed.  (I just found `Documentation\Contributors\DocumentationGuide`, will add a quick description of this behavior there!)

Source/Core/buildModuleUrl.js
ggetz: Re-opened https://github.com/CesiumGS/cesium/issues/8922.

Source/DataSources/CallbackProperty.js
ggetz: OK, I think I understand your point, thanks! I agree the parameter for the `getValue` function should be required.

Source/Core/ScreenSpaceEventHandler.js
ggetz: I have to say I'm a bit skeptical about this approach. @kring, would you be able to weigh in here?

thw0rted: If we need to back this out, let me know.  I really dislike the syntax we're forced to use to implement overloading via (tsd-)JSDoc; in TypeScript it would be one line per overload instead of 10+, and it's *much* easier to read.

That said, the resulting overloaded types really are quite helpful, especially when registering event handlers via a one-liner lambda function.  The alternative is that callbacks always type as a plain `Function` (`(...args: any[]) => void`) which, when the TS compiler is in "strict" mode, means that every callback has to explicitly type the arguments.  Worse, as far as I can tell, they aren't actually written out explicitly anywhere in (the current version of) the docs.

ggetz: Thanks @thw0rted, let's back it out so we can focus on getting the other improvements in this PR merged.

Feel free to open an issue or a separate PR for the overloaded types, and we can address them separately. :+1:

thw0rted: Please take a look and let me know if it's what you had in mind.  (I also caught one more little error, I just keep this branch open and correct stuff when it bites me.)

Source/Core/buildModuleUrl.js
ggetz, line 8: To get this merged, there shouldn't be any `TODO` comments in the code. Any action item should be in an issue so we can track them. Or let me know if there was discussion around any of these that wasn't resolved.

ggetz:
Thanks @thw0rted! After the last commit, the `build-ts` task is failing in CI. Could you please take a look?

Source/Core/buildModuleUrl.js
thw0rted: I took out the comment, should be addressed by https://github.com/CesiumGS/cesium/issues/8922 which you already re-opened.

Source/Scene/ImageryProvider.js
```diff
@@ -338,17 +347,19 @@ ImageryProvider.prototype.pickFeatures = function (
...
```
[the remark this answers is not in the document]
thw0rted: No sweat, I missed that too.  I was about to remove it, but should ask first, do you want the return type for the method (`loadImage`) to include `CompressedTextureBuffer`?  It looks like that's what you get if you pass it a `ktx2` file.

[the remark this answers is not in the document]
ggetz: Yes, please add `CompressedTextureBuffer`. Thanks!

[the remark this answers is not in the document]
thw0rted: Fixed.  I also resolved a couple very minor warnings from `tsd-jsdoc` that I noticed while testing this change, mostly self explanatory.  The `@private` ModelExperimental property is tagged private in a sibling class so I assumed it was an oversight.  The change from `String` to `string` is an ongoing effort to stop claiming to require/return boxed primitives when that's not actually true. (It's never actually true...)

ggetz:
Thanks @thw0rted! Latest changes look good. Can you back out the `Documentation/` changes since they're no longer needed? Then this is ready to be merged. :+1:

Source/Core/buildModuleUrl.js
```diff
@@ -142,6 +142,9 @@ buildModuleUrl.setBaseUrl = function (value) {
 
 /**
  * Gets the base URL for resolving modules.
+ *
+ * @function
+ * @returns {string} the configured base URL
```
ggetz, line 147: We do use capital `S` `String` throughout the codebase. Is there a reason that's changed in this file?

thw0rted: This is one of those questions where I convinced myself of the answer years ago but forgot most of the justification in the meantime.  [Typescript is very opinionated about it](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), but doesn't really offer much justification in their docs.  There's some good discussion on [this SO question](https://stackoverflow.com/questions/6793228/does-a-matter-whether-a-value-is-primitive-or-boxed).  (If you poke around on SO, you'll also find questions about "why do boxed types even exist?" with quotes from the language designers blaming management and generally regretting the decision.)

The short version is: lowercase-s `string` is a primitive and title-case `String` is an object ("box").  The most visible difference called out in that SO answer is that `typeof new String()` is `object`, not `string`. Primitives are boxed automatically if needed, but if only a primitive will do, you don't get free un-boxing.  I think the idea is that you probably almost never return `new String` in your code, but you might do a `typeof` check on a function argument at runtime, so the safest thing is to always use the primitive type in docs.

thw0rted:
Re the `Documentation` changes (and the corresponding JSDoc template update), does this mean that the team isn't willing to allow for any function overloading?  I think I had this conversation with Matt on the previous PR, several years ago, but at the time I hadn't figured out a way to avoid having multiple entries in the actual docs -- which is of course pretty confusing.

5+ overloads in one file looks bad, but the previous case was just one, I think so that exportKML gave the correct return type based on the options with which it was called.  That would be a lot less verbose/confusing with the new system.  I guess in the worst case I can always dig up the old commit if we need it again, but it'd be nice to leave it in place as an option.  Your call?

ggetz:
I would say hold off on adding the guide and the tooling until we have a verified use case in the code. It may be the path we end up going down eventually, and we'll have the record of it in this PR. But until we have a use case in the code itself, maintaining the documentation and tolling is questionable.

Source/Core/buildModuleUrl.js
thw0rted: I guess we could back it out?  I don't see the harm in making piecemeal changes, at least if we're touching adjacent docs (as I was in this case).  Not a big deal one way or another.

ggetz: Yes, thanks. I think we prefer the consistency. This is also how we've approach code style changes to the code base: All at once, ideally automating the change where possible.

ggetz:
Awesome, thanks @thw0rted! Thanks for your patience throughout the process.

PR #10295, opened 2022-04-14
ThirdParty.json

ggetz:
Adds `ThirdParty.json` as a convince for tracking third party dependencies. It's generated as part of the `build` step based on info we manually enter in `ThirdParty.extra.json` and then by pulling data from each module's `package.json` where available. `ThirdParty.extra.json` contains a list of packages which are shipped with the cesiumJS release, which more or less matches the `Third-Party Code` section of `LICENSE.md`.
  * Notably, I chose _not_ to include libraries included in `LICENSE.md` which the implementations were only referenced in comments. Let me know if leaving those out was an incorrect assumption.
  * Where a library is not sourced from an npm module, I chose the license and version available at the time when the library was last updated in our codebase.
  * I updated `LICENSE.md`, removing out-of-date entries and adding one which I believe was missed.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/third-party/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ThirdParty.json
```diff
@@ -0,0 +1,196 @@
+[
+  {
+    "name": "@zip.js/zip.js",
+    "license": [
```
shehzan10, line 4: @ggetz Can we make these inline?

shehzan10 (commented):
@ggetz Assuming you have done a thorough cleanup of the thirdparty licenses, these changes look good to me. Just one minor comment about the license arrays being in-line.

Also thanks for adding the gulp task. That makes it a lot easier for future use.

Should we document the use of when/how `Thirdparty.extra.json` should be used?

ThirdParty.json
ggetz: We're using [`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) to output formatted JSON. I'm not aware of an option that would keep arrays inline while still leaving the rest of the file formatted.

We could set up a plugin to format JSON like we do with JS files, but I think that's a bit out-of-scope for this PR.

ggetz:
Thanks @shehzan10. Updated the documentation to include steps for updating `package.extra.json`.

@sanjeetsuhag Could you do a final review and merge when ready?

gulpfile.cjs
```diff
@@ -1724,6 +1728,140 @@ function createSpecList() {
...
 
+/**
+ * Reads `ThirdParty.extra.json` file
+ * @param path {string} Path to `ThirdParty.extra.json`
+ * @param discoveredDependencies {Array<string>} List of previously discovered modules
+ * @returns {Promise<Array<Object>>} A promise to an array of objects with 'name`, `license`, and `url` strings
+ */
+function getLicenseDataFromThirdPartyExtra(path, discoveredDependencies) {
+  if (!fs.existsSync(path)) {
+    return Promise.resolve([]);
```
sanjeetsuhag, line 1739: Shouldn't we fail here and put out a message since the `ThirdParty.extra.json` should always exist?

sanjeetsuhag (commented):
@ggetz I just ran `npm i;  npm run build;` on two different machines. For the `@zip.js/zip.js` dependency, a different version was installed, causing a change in the `ThirdParty.json`. This is unfortunate because simply running `npm run build` is now causing a change in the repo that needs to be handled. Additionally, is it intentional that we are not shipping this file with the release ZIP?

ggetz:
> Additionally, is it intentional that we are not shipping this file with the release ZIP?

@shehzan10 Is `ThirdParty.json` something we want to ship? I doesn't seem like something that would be useful to users beyond `LICENSE.md`

ggetz:
> I just ran npm i; npm run build; on two different machines. For the @zip.js/zip.js dependency, a different version was installed, causing a change in the ThirdParty.json. This is unfortunate because simply running npm run build is now causing a change in the repo that needs to be handled. 

The point is for the script to increment versions numbers to match the shipped version. Since this seems to generate too much noise, I'll break this into its own node script.

shehzan10:
> @shehzan10 Is `ThirdParty.json` something we want to ship? I doesn't seem like something that would be useful to users beyond `LICENSE.md`

@ggetz Right, since we're shipping License.md, no need to ship ThirdParty.json too.

ggetz:
Thanks @shehzan10.

@sanjeetsuhag This is updated.

sanjeetsuhag (approved):
Changes look good! Thanks @ggetz!

Issue #10191, opened 2022-03-10
Race condition in billboard clamping


mramato:
I found this while testing out native promises, but confirmed it happens in master as well.

1. Open Billboard Sandcastle: https://sandcastle.cesium.com/?src=Billboards.html
2. Select "Add marker billboards" It will look like this: ![Screenshot_20220310_184309](https://user-images.githubusercontent.com/1495196/157774113-f750f7d4-6fb4-4393-a013-06806ba0c998.png)
3. Focus should be on the combo box, so press arrow down to switch to "Disable depth test when clamped to ground.
4. As soon as the billboard is visualized and while the terrain is still going, press arrow up to go back to "Add marker billboards" it will now look like this: 
![Screenshot_20220310_184417](https://user-images.githubusercontent.com/1495196/157774222-adddcbd9-767c-46e3-96d0-b5e738cc373f.png)

The billboard will remain in the wrong location for other examples as well. Something thinks this billboard is supposed to be clamped to ground when it's not.

P.S. You can use the mouse to switch between the demos as well, but the arrow key makes it easy

yurichen17:
Hi there, I would like to try and fix this bug! Could I be assigned to this?

mramato:
Hi, @yurichen17, thanks for offering to help! We don't use of the "Assignee" field in GitHub. For bugs like this one there is no need to ask for permission and you can work on the issue and open a pull request for review when ready. If you have any questions, you can comment on the issue or on the PR itself if it's far enough along to open.

If you work on something larger in the future, like a new feature or something that requires a lot of code changes, then it's customary to comment on the issue or start a thread on our forum beforehand. While we love contributions, larger features or code changes usually need some coordination among the maintainers.

There's also a lot of standard documentation in https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors

Thanks, again!

yurichen17:
Hi @mramato , thank you so much for your quick response and the resources! I'm new to contributing so I appreciate the help in getting started. I'll be sure to go over the documentation and open a pull request for this issue. Thank you again!

yurichen17:
[...]

PR #10306, opened 2022-04-19
Billboard clamping

yurichen17:
Fixes issue [#10191 ](https://github.com/CesiumGS/cesium/issues/10191)

This is my first pull request so I am very open to feedback! It was a challenge but I hope that I helped fix this bug.

**Context:** 
Inside QuadtreePrimitive.js, there is updateHeight function that is called inside Billboards.js that takes in a position and the function updateFunction. The issue that arises is that updateHeight creates an object that stores the updateFunction and an old position, which gets recalled during updates and therefore changes the position of the billboard to the wrong place, and appear to "clamp." To combat this, inside object.removeFunc, I set the object callback to undefined. Because of this change, I have to add an "if" check on line 1509 in QuadtreePrimitive.js to only run the callback function if it is defined. With the combined changes, the callback function is actually changed to "undefined" rather than keeping it's original value that contained old positioning of the billboards. 

I added a test but I struggled with making a test that imitated the structure from the issue. I would love to get some help on that front!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 - Fixed debug label rendering in `Cesium3dTilesInspector`. [#10246](https://github.com/CesiumGS/cesium/issues/10246)
 - Fixed a crash that occurred in `ModelExperimental` when loading a Draco-compressed model with tangents. [#10294](https://github.com/CesiumGS/cesium/pull/10294)
 - Fixed an incorrect model matrix computation for `i3dm` tilesets that are loaded using `ModelExperimental`. [#10302](https://github.com/CesiumGS/cesium/pull/10302)
+- Fixed clamping issues with billboards after using mode "Disable depth test when clamped to ground." [#10191](https://github.com/CesiumGS/cesium/issues/10191)
 
 ### 1.92 - 2022-04-01
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@
 - Fixed debug label rendering in `Cesium3dTilesInspector`. [#10246](https://github.com/CesiumGS/cesium/issues/10246)
 - Fixed a crash that occurred in `ModelExperimental` when loading a Draco-compressed model with tangents. [#10294](https://github.com/CesiumGS/cesium/pull/10294)
 - Fixed an incorrect model matrix computation for `i3dm` tilesets that are loaded using `ModelExperimental`. [#10302](https://github.com/CesiumGS/cesium/pull/10302)
-- Fixed clamping issues with billboards after using mode "Disable depth test when clamped to ground." [#10191](https://github.com/CesiumGS/cesium/issues/10191)
+- Fixed race condition during billboard clamping when the height reference changes. [#10191](https://github.com/CesiumGS/cesium/issues/10191)
 
 ### 1.92 - 2022-04-01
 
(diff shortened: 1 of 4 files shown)


cesium-concierge:
Thanks for the pull request @yurichen17!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/BillboardCollectionSpec.js
```diff
@@ -2491,6 +2491,21 @@ describe(
         expect(b._clampedPosition).toBeUndefined();
       });
 
+      it("disableDepthTest after another function", function () {
+        const b = billboardsWithHeight.add({
+          heightReference: HeightReference.CLAMP_TO_GROUND,
+          position: Cartesian3.fromDegrees(-122, 46.0),
+          disableDepthTestDistance: Number.POSITIVE_INFINITY,
+        });
+        expect(scene.globe.callback).toBeDefined();
```
ggetz, line 2500: Out of context, I would have suggested keeping the test limited to the billboard collection itself to avoid coupling this test to behavior with another class. However, I see this is the same way we are testing `"callback updates the position"`, so I think its fine to keep consistant.

A better way in both cases might be to call `renderForSpecs` after setting the `heightReference` and confirming that `._clampedPosition` has or hasn't changed based on the expected behavior, but there may be reason why `"callback updates the position"` wasn't written that way to begin with that I'm not aware of.

ggetz:
Thanks for the PR @yurichen17! 😄 

I can confirm that this fixes the behavior described in the issue.

Please add a line to [`CHANGES.md`](https://github.com/CesiumGS/cesium/blob/main/CHANGES.md) under "Fixes" describing the fix.

Specs/Scene/BillboardCollectionSpec.js
yurichen17: That makes sense; I've updated the test changes, though only in the test I made and not in "callback updates the position."

yurichen17:
this is ready

CHANGES.md
```diff
@@ -15,6 +15,7 @@
 - Fixed debug label rendering in `Cesium3dTilesInspector`. [#10246](https://github.com/CesiumGS/cesium/issues/10246)
 - Fixed a crash that occurred in `ModelExperimental` when loading a Draco-compressed model with tangents. [#10294](https://github.com/CesiumGS/cesium/pull/10294)
 - Fixed an incorrect model matrix computation for `i3dm` tilesets that are loaded using `ModelExperimental`. [#10302](https://github.com/CesiumGS/cesium/pull/10302)
+- Fixed clamping issues with billboards after using mode "Disable depth test when clamped to ground." [#10191](https://github.com/CesiumGS/cesium/issues/10191)
```
ggetz, line 18:
```suggestion
- Fixed race condition during billboard clamping when the height reference changes. [#10191](https://github.com/CesiumGS/cesium/issues/10191)
```

ggetz: Just a slight tweak to the message since the issue applies more broadly than just that sandcastle example.

yurichen17: Sounds good! Thank you!

ggetz:
Awesome, thanks @yurichen17 !

Issue #3911, opened 2016-05-11
What tasks should be supported from within the release zip?


hpinkos:
Reported on the forum: https://groups.google.com/forum/?hl=en#!topic/cesium-dev/eaY-tlnabK0

Running `npm run build` from the release fails because `.jshintrc` is missing from the root, apps, and apps/sandcastle directories

mramato:
You aren't supposed to be able to run `npm run build` from the release zip.  The release zips are only meant for using Cesium, not building it.

hpinkos:
Then why do we include the gulpfile?

mramato:
It was probably just an oversight from when we switched to gulp, however, having the gulpfile should let you run the tests from the command line (which is something we currently support in the released zips).  Though I haven't tried it as we are still using the old Jasmine testing for release testing.

hpinkos:
Discussion from #6025 (deleting gulp file from release)

I'm not sure this really fixes the issue because there will still be gulp-based tasks listed in `package.json`, so it's not going to stop our users from attempting to run build tasks against the release zip and submitting github issues or forum posts because they think it's broken.  We could modify the package.json during zip creation so that it only include support commands.

But the real question we need to answer is, "What else are do we support doing with the release zip?":

1. Should you be able to rebuild Cesium (and Apps and documentation)?
    - sure, if it's easy enough
2. Should you be able to run all tests?
    - yes
3. Should you be able to run jsHint/eslint and other build stuff?
    - sure, if it's easy enough

Depending on the answers to these questions (and maybe others I glanced over) we can modify the zip further to better suit the needs.

ggetz:
We removed the `gulpfile` in https://github.com/CesiumGS/cesium/pull/10281, but after seeing #6025, I think that was the wrong move.

The reason we removed it was that it was no longer possible to run _any_ gulp tasks from within the release zip. However, this was due to https://github.com/CesiumGS/cesium/pull/8659 and can be fixed by including `.gulp.json`. 

Once that file is restored, it's possible to run a subset of the tasks from within the release zip:

🔴  `build` and `build-watch` - fails with `Error: ENOENT: no such file or directory, open 'Apps/Sandcastle/.jshintrc'`
⚠️  `build-ts` - Can run if `Tools/jsdoc/**` is included in the releaseZip (~85KB uncompressed)
🔴  `buildApps` - fails to an implicit dependency on the `build` task
✅  `clean`
✅  `cloc`
[...]

PR #10281, opened 2022-04-08
Removes Specs, gulpfile.cjs and web.config file from makeZipFile

sanjeetsuhag:
Fixes https://github.com/CesiumGS/cesium/issues/10270

| | `main` | `remove-specs-from-zip` |
| - | :-: | :-: |
| Compressed Size (MB) | 66.2 | 52.3 |
| Uncompressed Size (MB) | 192.3 | 159 |

I noticed that we don't really do much with the `gulpfile.cjs` and the `web.config` files, so this PR removes them too. Not sure if that's the right thing to do, but none of the commands through gulp from inside the zip file anyway.

PR #8659, opened 2020-03-04
Fix Cesium in Node.js 12+.

kring:
Node.js 12+ requires that ES modules are loaded with `import` rather than `require`, and it throws an exception if you get this wrong. An ES module is any module that is named `.mjs` or whose closest package.json specifies `"type": "module"` (as Cesium's does).

The [Node.js module docs](https://nodejs.org/api/esm.html#esm_writing_dual_packages_while_avoiding_or_minimizing_hazards) explain a couple of approaches for making a single package that works when imported as either an ES module or as a CommonJS module. But the first requires that the library be natively CommonJS (which Cesium is not), and the second requires two copies of the library (one CommonJS modules, one ES modules) and a dodgy approach to allow the two to share state 🙄. There's no scenario where the library is natively ES modules and can be loaded in CommonJS apps using `esm`, and short of terrible hackery it appears that this is no longer possible to achieve in Node 12+.

So, this PR does some terrible hackery, as suggested here:
https://github.com/standard-things/esm/issues/855#issuecomment-566825957
(By the way, the commentor above implemented the check and exception that is causing us problems, as well as large parts of Node's module system in general I believe, so presumably knows what he's talking about! https://github.com/nodejs/node/pull/29492)

This PR also adds a `package.json` with `"type": "commonjs"` to the Tools directory so that the .js files there are treated as CommonJS. This is needed for several of our build tools to work in Node 12+. It also renames gulpfile.js to gulpfile.cjs so that file isn't interpreted as an ES module, and it adds an explicit parameter to all the `npm run scripts` telling gulp to use it. gulp isn't (yet?) able to look for a gulpfile.cjs itself.

Other approaches we could take, not implemented in this PR:
* Remove `"type": "module"`. If we do that, we don't need any of the other changes in this PR and everything will work great in Node 12+. BUT users won't be able to import Cesium as "real" ES modules using `--experimental-modules` anymore. They'll always have to use `esm`.
* Ship separate CommonJS and ES builds of Cesium in the npm package. Name the files `.mjs` or `.cjs` as appropriate (or put a separate package.json in each version's subdirectory with a different `"type"`). Ideally break all of Cesium's shared state into separate CommonJS modules as described in  Approach 2 in the node.js docs above. That way if one thing imports the CommonJS version and another thing imports the ES version, will "only" have two versions of all the code rather than everything breaking cause there's two versions of the global data.

Fixes #8452

PR #6025, opened 2017-12-01
Remove gulpfile from release

hpinkos:
Fixes #3911

The release zips are only meant for using Cesium, not building it, so there's no need to include the `gulpfile`.

PR #10311, opened 2022-04-21
Fix gulpfile tasks in release zip, but limit packaged scripts

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/3911

We removed the `gulpfile` in https://github.com/CesiumGS/cesium/pull/10281, but after seeing #6025, I think that was the wrong move.

The reason we removed it was that it was no longer possible to run _any_ gulp tasks from within the release zip. However, this was due to https://github.com/CesiumGS/cesium/pull/8659 is fixed here by including `.gulp.json`. 

Additionally, it's only possible and helpful a subset of scripts in the release zip. So this PR limits those to the following list:

* `start`
* `start-public`
* `clean`
* `cloc`
* `coverage`
* `eslint`
* `build-specs`
* `test-*`
* `prettier-check`

This list is arrived at by omitting build and watch tasks, as well as the `deploy-*` tasks. Including `generateDocumentation` and `build-ts` is arguable since they can be run to verify the validity of the docs. But we already included the generated output, and the size of the documentation build tools are not as negligable as eslint or prettier. 

In order to run those, we include include `.eslintrc.json`,`.eslintignore`, related eslint config files in the release zip (~7KB uncompressed); `.prettierignore` (<1KB uncompressed); and `Specs/**` which was previously, but I believe erroneously, excluded in https://github.com/CesiumGS/cesium/pull/10281.

Finally, after https://github.com/CesiumGS/cesium/pull/10293 adds back the browser spec runner and associated files, we can re-add the link to the spec runner removed from `index.release.html` in https://github.com/CesiumGS/cesium/pull/10184.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/packaged-scripts/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

gulpfile.cjs
```diff
@@ -451,6 +451,29 @@ gulp.task(
...
+    delete packageJson.scripts.buildApps;
+    delete packageJson.scripts.combine;
+    delete packageJson.scripts.combineRelease;
+    delete packageJson.scripts.generateDocumentation;
+    delete packageJson.scripts["generateDocumentation-watch"];
+    delete packageJson.scripts.makeZipFile;
+    delete packageJson.scripts.minify;
+    delete packageJson.scripts.minifyRelease;
+    delete packageJson.scripts.release;
+    delete packageJson.scripts.combineRelease;
```
ebogo1, line 469: duplicate of line 462?

ebogo1 (commented):
Looks good, just a minor comment!

gulpfile.cjs
ggetz: Yes, good catch!

ggetz:
Thanks @ebogo1! Updated.

ebogo1:
Hm, this is a pretty niche use case but I noticed that if we run `npm run clean` followed by `npm run build-specs`, the latter fails because `clean` removes the SpecList.js file. This happens in the non-release version as well, but if there's no way to rebuild SpecList.js from the release zip (it's part of the `"build"` task) maybe it's worth reconsidering how we include the `build-specs` task. 

I think it's minor enough to not do anything about it in this PR but thought I'd mention it.

ggetz:
Good point. Maybe we should omit the `clean` script entirely then, since its possible to clean files which can't be re-created from within the release zip.

ebogo1:
Should the `cloc` script be removed too then?

ggetz:
Yeah, I think that follows. I also don't see an equivalent command in packages from comparable projects.

ggetz:
Updated once again.

ebogo1:
Cool, thanks @ggetz!

PR #10317, opened 2022-04-22
Ensure arrays exist in `ModelComponents`

ptrgags:
I noticed a regression today, while loading a `.pnts` tileset with `ModelExperimental`, a line that was accessing `node.skins.length` was failing because `PntsLoader` doesn't set `node.skins` and it defaults to `undefined` in `ModelComponents`. 

I did a sweep through `ModelComponents` and related code to ensure array properties default to an empty array, so it's always safe to do `component.property.length` or loop over the array. This simplifies the code in a few places, one less `define()` check.

@j9liu could you review when you get a chance?

### DIFF
--- a/Source/Scene/ModelExperimental/B3dmLoader.js
+++ b/Source/Scene/ModelExperimental/B3dmLoader.js
@@ -304,32 +304,23 @@ function createStructuralMetadata(loader, components) {
 
 // Recursive function to add the feature ID attribute to all primitives that have a feature ID vertex attribute.
 function processNode(node) {
-  if (!defined(node.children) && !defined(node.primitives)) {
-    return;
+  for (let i = 0; i < node.children.length; i++) {
+    processNode(node.children[i]);
   }
 
-  let i;
-  if (defined(node.children)) {
-    for (i = 0; i < node.children.length; i++) {
-      processNode(node.children[i]);
-    }
-  }
-
-  if (defined(node.primitives)) {
-    for (i = 0; i < node.primitives.length; i++) {
-      const primitive = node.primitives[i];
-      const featureIdVertexAttribute = ModelExperimentalUtility.getAttributeBySemantic(
-        primitive,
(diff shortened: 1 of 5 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-component-array-defaults/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelExperimental/B3dmLoader.js
```diff
@@ -304,32 +304,23 @@ function createStructuralMetadata(loader, components) {
 
 // Recursive function to add the feature ID attribute to all primitives that have a feature ID vertex attribute.
 function processNode(node) {
-  if (!defined(node.children) && !defined(node.primitives)) {
-    return;
+  for (let i = 0; i < node.children.length; i++) {
```
j9liu, line 307: Can we extract the length variable here while we're at it? e.g. `const childrenLength = node.children.length`. just for consistency with the other changes in this PR

Source/Scene/ModelExperimental/ModelMatrixUpdateStage.js
```diff
@@ -68,20 +67,17 @@ function updateRuntimeNode(runtimeNode, sceneGraph, transformToRoot) {
     }
   }
 
-  if (defined(runtimeNode.children)) {
-    for (i = 0; i < runtimeNode.children.length; i++) {
-      const childRuntimeNode =
-        sceneGraph._runtimeNodes[runtimeNode.children[i]];
+  for (i = 0; i < runtimeNode.children.length; i++) {
```
j9liu, line 70: same thing here, store the length in a `const` variable

ptrgags:
@j9liu updated!

j9liu:
@ptrgags - looks good, I'll merge once Travis passes.

PR #10325, opened 2022-04-26
Disable text selection in timeline and credits

jdfwarrior:
This has bugged me for a long time. Pressing Ctrl + A to select all text selects the text in the timeline and the credits area. Maybe you actually want the credits area? But it seems like the timeline shouldn't be selectable because, once you get it selected, you have to click very specific areas to deselect and it's annoying.

Text in the animation controller can still be selected right now. I may try to figure out how to resolve that as well.

No changes to the changelog necessary (in my opinion) since it's something some people may have never noticed in the first place and doesn't modify the API at all.

### DIFF
--- a/Source/Widgets/CesiumWidget/CesiumWidget.css
+++ b/Source/Widgets/CesiumWidget/CesiumWidget.css
@@ -18,7 +18,6 @@
   font-size: 10px;
   text-shadow: 0px 0px 2px #000000;
   padding-right: 5px;
-  user-select: none;
 }
 
 .cesium-widget-credits a,
--- a/Source/Widgets/CesiumWidget/CesiumWidget.css
+++ b/Source/Widgets/CesiumWidget/CesiumWidget.css
@@ -18,6 +18,7 @@
   font-size: 10px;
   text-shadow: 0px 0px 2px #000000;
   padding-right: 5px;
+  user-select: none;
 }
 
 .cesium-widget-credits a,
--- a/Source/Widgets/Timeline/Timeline.css
+++ b/Source/Widgets/Timeline/Timeline.css
@@ -4,6 +4,9 @@
   bottom: 0;
   overflow: hidden;
   border: solid 1px #888;
+  -moz-user-select: none;
+  -webkit-user-select: none;
+  -ms-user-select: none;
   user-select: none;
 }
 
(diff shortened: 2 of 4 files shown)


cesium-concierge:
Thank you so much for the pull request @jdfwarrior! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/jdfwarrior/cesium/blob/disable-text-selection/CONTRIBUTORS.md) file!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/disable-text-selection/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Widgets/CesiumWidget/CesiumWidget.css
```diff
@@ -18,6 +18,7 @@
   font-size: 10px;
   text-shadow: 0px 0px 2px #000000;
   padding-right: 5px;
+  user-select: none;
```
ggetz, line 21: I think its arguable that the credits shouldn't be selectable. It's [possible for users to add additional credits that show onscreen](https://cesium.com/learn/cesiumjs/ref-doc/Credit.html).

Source/Widgets/Timeline/Timeline.css
```diff
@@ -4,6 +4,7 @@
   bottom: 0;
   overflow: hidden;
   border: solid 1px #888;
+  user-select: none;
```
ggetz, line 7: It looks like anywhere else in the css where we apply this rule, we apply all the browser-specific rules as well. Let's do so here for consistency. 

```
  -moz-user-select: none;
  -webkit-user-select: none;
  -ms-user-select: none;
  user-select: none;
```

ggetz (commented):
Thanks @jdfwarrior! I have a few comments. 

I would also ask that you add a line to `CHANGES.md`. Although the change will most likely not be noticed by most, it will keep a record of the change in case it affects someone.

Source/Widgets/CesiumWidget/CesiumWidget.css
jdfwarrior: Fair enough. Yeah I wasn't 100% sure that this was needed for the credits.

Source/Widgets/Timeline/Timeline.css
jdfwarrior: Added that in. Thanks @ggetz

ggetz:
Thanks @jdfwarrior, and congrats on your first cesium PR!

Issue #9827, opened 2021-09-24
Some materials not working in iOS/Safari 15


alexrobin:
Some materials don't render correctly in the latest release of iOS 15 that now uses Metal backend of WebGL (Webkit). I noticed in particular translucent materials are not working on some geometries and dashed polylines are not working but there might be others.

Sandcastle example:
https://sandcastle.cesium.com/gallery/Partial%20Ellipsoids.html
https://sandcastle.cesium.com/gallery/Polygon.html
https://sandcastle.cesium.com/gallery/Plane.html
https://sandcastle.cesium.com/gallery/Polyline%20Dash.html
https://sandcastle.cesium.com/gallery/Wall.html
https://sandcastle.cesium.com/gallery/Spheres%20and%20Ellipsoids.html

Browser:
Safari 15

Operating System:
iPadOS 15

mramato:
I can reproduce this on my own iPad Pro (10.5 inch MQDT2LL/A), but oddly enough not my iPhone Xs.

So there's a chance this is iPadOS only and not on desktop or iPhone.  Can anyone else confirm?

mramato:
Also, thanks for the bug reports, @alexrobin!

sanjeetsuhag:
I can confirm that all these Sandcastles work fine on my iPhone XR running iOS 15.0

ebogo1:
+1 for Sandcastles running fine on my macOS laptop (11.6, Safari 15)

alexrobin:
Yes, I forgot to mention this but our application that experienced these issues also seems to work OK on the latest iPad Pro 11" but not on iPad Pro 10.5".

alexrobin:
FYI, setting `orderIndependentTranslucency` to false in the viewer options seems to be a workaround for this issue.

alexrobin:
Did you make any progress on this?
Let me know if you think it's a problem on Apple's side. We're in touch with them on this.
Thanks

ebogo1:
@alexrobin No progress at this time, but we're hoping to look into it more when we get a chance. Please keep us posted if you find anything new!

hayderGo:
I can reproduce this on my iPhone 12 Pro Max (IOS 15.4)

mramato:
[...]

PR #10337, opened 2022-04-29
Temporarily disable OIT by default for iOS and iPad

ggetz:
Due to increased reports of https://github.com/CesiumGS/cesium/issues/9827, this disables `orderIndependentTranslucency` by default on iOS and iPad as a temporary workaround.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ### 1.93 - 2022-05-02
 
+##### Breaking Changes :mega:
+
+- Temporarily disable `Scene.orderIndependentTranslucency` by default on iPad and iOS due to a WebGL regression, see [#9827](https://github.com/CesiumGS/cesium/issues/9827). The old default will be restored once the issue has been resolved.
+
 ##### Additions :tada:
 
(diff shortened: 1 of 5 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ebogo1 (approved):
Changes look good. I assume `navigator.platform` works as expected but I don't have any iOS devices to test on. Have we checked that this branch fixes issues on the newer devices?

ggetz:
Thanks @ebogo1.

> Have we checked that this branch fixes issues on the newer devices?

Do you mean other than the issues which were brought up in https://github.com/CesiumGS/cesium/issues/9827?

ebogo1:
I meant more to check that `navigator.platform` was identifying iOS devices as the `iPad` `iPhone` `iPod` strings. I couldn't find an official list of its possible values. Happy to merge if we're confident that those are correct, I just wanted to double check.

ggetz:
I was able to verify this is working on my iPhone, logging the value of `Scene.orderIndependentTranslucency`:

image

I tried my iPad, which is a iPad Pro. Unfortunately I believe this type of iPad always presents the desktop version of a website, so the value of `navigator.platform` is `MacIntel`, same as my desktop. However, I believe it should work for [smaller and older models of iPad](https://stackoverflow.com/questions/57776001/how-to-detect-ipad-pro-as-ipad-using-javascript).

ebogo1:
Cool, thanks @ggetz!

Issue #10342, opened 2022-05-03
Label background is broken in 1.93 (unwanted gradient)


UniquePanda:
Hi all,
We've just updated from 1.92 to 1.93 and found out that our labels (Cesium entities) now have some kind of gradient added to their background. As I wasn't able to find any related changelogs or code changes, I guess this is a bug?

You can see it in action in the attached sandcastle and in this screenshot:  
![image](https://user-images.githubusercontent.com/3686982/166413213-d8f245d0-2ffc-41a9-8d69-1d2fa39e22b1.png)

For comparison here's a screenshot of a label from our App with a previous Cesium version:  
![image](https://user-images.githubusercontent.com/3686982/166413483-3a4d39e1-d4ee-4575-8dce-f50c7e6b17f2.png)

Sandcastle example:  
https://sandcastle.cesium.com/#c=bZBRT8IwFIX/ys2eRjK7IIJhDqJME40kPmj0pS/ddoGGriVtx0TDf7fdUIzSh6b33POdm95CSWNhy7FBDROQ2ECGhtcVeW21kAZFW2dKWsYlahr0rqiksmMISsstR0NYWYafVAJslHGKksl3Usa0dS8mB2ShVXWLS41owrPLIemPLvqj0WUEgzEZD8/96UU+RLAcRQJtIIDFd5sADeZehYbbFcxYsV5qVcuSBlHnWnAhMiWUPk72FZnNb7LHg8esVHNEE7C6xkMr/5FPhbzdP7zctc69u/d+B0EUpMbuBE67AIBrXm2UtlBrERISW6w2grm/x3ldrNGSwhgPemsa/0bTkm+Bl5MT24ZCMGNcZ1EL8cw/kAbTNHb+f6hQrORy+bRFLdjO21b96bwTCSFp7MrTpFVK5Ez/Sf4C

Browser:  
Tested on Chrome 100.0.4896.127

Operating System:
Windows 10

anne-gropler:
I confirm this happens on Firefox, too. (Windows 10)

GatorScott:
Looks good though!

ggetz:
Thanks for the report @UniquePanda! I can replicate the issue. This is a bug, not an intentional change

PR #10343, opened 2022-05-03
Add label background image asynchronously

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/10342

[Here's a local sandcastle that easily shows the issue.](http://localhost:8080/Apps/Sandcastle/index.html#c=bZBRT8IwFIX/ys2eRjK7IIJhDqJME40kPmj0pS/ddoGGriVtx0TDf7fdUIzSh6b33POdm95CSWNhy7FBDROQ2ECGhtcVeW21kAZFW2dKWsYlahr0rqiksmMISsstR0NYWYafVAJslHGKksl3Usa0dS8mB2ShVXWLS41owrPLIemPLvqj0WUEgzEZD8/96UU+RLAcRQJtIIDFd5sADeZehYbbFcxYsV5qVcuSBlHnWnAhMiWUPk72FZnNb7LHg8esVHNEE7C6xkMr/5FPhbzdP7zctc69u/d+B0EUpMbuBE67AIBrXm2UtlBrERISW6w2grm/x3ldrNGSwhgPemsa/0bTkm+Bl5MT24ZCMGNcZ1EL8cw/kAbTNHb+f6hQrORy+bRFLdjO21b96bwTCSFp7MrTpFVK5Ez/Sf4C)

This arose during https://github.com/CesiumGS/cesium/pull/10275. I've confirmed that this fixes the above issue while still maintaining the 3d tileset inspector  label behavior that was restored as part of that pull request. 

I also removed `_whitePixelIndex` as it was assigned but never read anywhere as far as I could tell.

However, I'm having trouble understanding _why_ this fix works. The canvas `canvas` passed to the `addImage` function does not require loading and therefore does not require awaiting a promise in order to add it to the texture atlas.

@lilleyse Would you be able to give this a look?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed the inaccurate computation of bounding spheres for `ModelExperimental`. [#10339](https://github.com/CesiumGS/cesium/pull/10339/)
+- Fixed [#10342](https://github.com/CesiumGS/cesium/issues/10342)
 
 ### 1.93 - 2022-05-02
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 ##### Fixes :wrench:
 
 - Fixed the inaccurate computation of bounding spheres for `ModelExperimental`. [#10339](https://github.com/CesiumGS/cesium/pull/10339/)
-- Fixed [#10342](https://github.com/CesiumGS/cesium/issues/10342)
+- Fixed label background rendering. [#10342](https://github.com/CesiumGS/cesium/issues/10342)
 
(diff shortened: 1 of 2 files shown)


lilleyse:
It seems like canvas drawing operations have a synchronous API but are actually queued up and drawn when the browser hits an animation refresh. I assume by wrapping it in a promise the promise gets resolved after that refresh.

I would've expected `gl.texImage2D` to wait until the canvas is ready. So I'm not really sure what's going on either without diving deeper into this.

ggetz:
Thanks @lilleyse. I'm not sure what may be happening with `gl.texImage2D`, but that at least explains what's happening with the update flow here. Added a comment to explain.

PR #10351, opened 2022-05-04
Fix crash for zero-area `region` bounding volumes

lilleyse:
One tileset I was looking at had a `region` bounding volume where west and east were the same value. This caused a divide by zero in the `TileBoundingRegion` math.

This isn't disallowed in the spec and it's a pretty straightforward fix. If the east-to-west vector has a magnitude of zero the west normal is used instead. This is fine to do because when a region is infinitesimally small the west vector and east-to-west vector have the same direction.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Issue #10222, opened 2022-03-23
Continuously printing error messages when zooming to a tileset that cannot be loaded


javagl:
<sup>Tested with Cesium 1.91, with the current [state of `main`](https://github.com/CesiumGS/cesium/tree/3520c520abd447c7d0b7f13f4d910d154bb20b02), on Google Chrome, Version 99.0.4844.82 - although the latter is likely not relevant here</sup>

The following sandcastle tries to load a tileset that does not exist, and then zoom to it:
```
const viewer = new Cesium.Viewer("cesiumContainer");
const tileset = viewer.scene.primitives.add(
  new Cesium.Cesium3DTileset({
    url: "http://example.com/tileset.json"
  })
);
viewer.zoomTo(tileset);
```
It causes the following error to be printed continuously, flooding the console:
```
Uncaught (in promise) RequestErrorEvent {... }
  Promise.then (async)		
  updateZoomTarget	@	Viewer.js:2194
  Viewer._postRender	@	Viewer.js:2176
  Event.raiseEvent	@	Event.js:138
  Scene.render	@	Scene.js:3897
  CesiumWidget.render	@	CesiumWidget.js:790
  render	@	CesiumWidget.js:42
  requestAnimationFrame (async)		
  requestAnimationFramePolyFill	@	requestAnimationFrame.js:65
  render	@	CesiumWidget.js:43
  ...
```
This happens regardless of what caused the loading error - a malformed JSON or any other invalid element in the tileset causes the same behavior, with a different error message.

---

My gut feeling is that somewhere around [`Viewer#updateZoomTarget`](https://github.com/CesiumGS/cesium/blob/3520c520abd447c7d0b7f13f4d910d154bb20b02/Source/Widgets/Viewer/Viewer.js#L2194), there should be something like
```
if (target.failedToLoad()) { return nopeNotGonnaTryThatAgain(); }
else { return target.readyPromise().then(...); }
```
or maybe the Error that is thrown from trying to load the tileset should cause a call to `cancelZoom` or `clearZoom` or so in the chained promise that is created there, but these are just guesses until now.

ggetz:
Thanks @javagl, there's a few areas where we're in the process of cleaning up error handling after the transition from when.js to native promises-- https://github.com/CesiumGS/cesium/issues/10178. I believe this should be addressed as part of that effort.

javagl:
This still seems to happen in 1.95 - OK to reopen this?

ggetz:
[...]

Issue #10178, opened 2022-03-09
Factor out "defer" usage for better error handling


ggetz:
Cesium uses `defer` throughout the codebase as a remnant from when we were using `when.js`. However, the recommended best practice is to use `new Promise` constructor directly because it makes it much harder to let an exception escape. Functions passed to the construct should only reject the promise, never throw.

https://github.com/CesiumGS/cesium/pull/10149 excluded several specs in `Cesium3DTilesetSpec`, `GltfLoaderSpec`, `TimeDynamicPointCloudSpec`, and `ModelExperimentalSpec` which can be re-enabled once errors are better handled as a result of this change.

ggetz:
@lilleyse @ptrgags Is there a particular reason why the loaders, like [`BufferLoader.js`](https://github.com/CesiumGS/cesium/blame/main/Source/Scene/BufferLoader.js) have a `promise` rather than having the `load` function return a promise? As far as I'm aware, the later is a much more common pattern and would make removing `defer` much more streamlined.

lilleyse:
Separate promises + a new state could be ok. Would this move some of the burden to the caller to load and then process the loader?

`Model.js` seems like it will have a similar problem and I'm curious how we'll approach that.

ggetz:
I'm not entirely sure yet, but I don't think so. So far it seems it can be handled internally and we just need to move around where the caller is actually handling the promises.

I agree `Model.js` will have similar problems. What do you think the timeline is for removing it? Maybe we can leave defer there until its replaced by `ModelExperimental`. It's not causing issues in the specs like the loaders are currently.

lilleyse:
> I'm not entirely sure yet, but I don't think so. So far it seems it can be handled internally and we just need to move around where the caller is actually handling the promises.

Ok that doesn't sound too bad. I was imagining that a larger restructuring would be needed.

> I agree `Model.js` will have similar problems. What do you think the timeline is for removing it? Maybe we can leave defer there until its replaced by `ModelExperimental`. It's not causing issues in the specs like the loaders are currently.

We're aiming to remove it in June. Holding off until `ModelExperimental` replaces it sounds good.

mramato:
FYI, we ran into issues in unit testing a Cesium-based app related to this behavior and had to insert `model.readyPromise.catch(() => {});` into some spots in the tests in order to ensure it doesn't break the entire testing suite when it resolves (simply awaiting doesn't work).

[...]

PR #10361, opened 2022-05-09
Remove defer from 3D Tiles, glTF, and point cloud loading

ggetz:
Part of https://github.com/CesiumGS/cesium/issues/10178
Fixes https://github.com/CesiumGS/cesium/issues/10363

Removed use of `defer` from 3D Tiles, glTF, and point cloud loading.

I don't think this is the be-all-end-all improvement, since in a few places we are still not entirely encapsulating the entire promise resolve/reject behavior within a `new Promise` callback, and probably won't be able to do so without a good amount more refactoring, and probably deprecating `readyPromises` throughout.

But, I think this should be at least a step in the right direction as this does remove `defer` from plus this does now handle several errors which were not being handled properly before. One particular area this cleaned up was handling 3D Tiles promises when destroyed. In doing so, this allow us to re-introduce all of the specs that had to be excluded as part of https://github.com/CesiumGS/cesium/pull/10149.

TODO:
- [x] Merge main
- [x] Confirm performance of loading was not affected
- [x] Clear up https://github.com/CesiumGS/cesium/compare/defer-b-gone?expand=1#diff-70f07d262fd26759318ee0562b7a1a45915ea04e2d091f654a41018ecb504a92R113
- [x] Update `CHANGES.md`


Source/Scene/BufferLoader.js
```diff
@@ -98,32 +97,39 @@ Object.defineProperties(BufferLoader.prototype, {
...
 
 function loadExternalBuffer(bufferLoader) {
   const resource = bufferLoader._resource;
   bufferLoader._state = ResourceLoaderState.LOADING;
-  BufferLoader._fetchArrayBuffer(resource)
+  const promise = BufferLoader._fetchArrayBuffer(resource);
+
+  // TODO: I don't think this is OK-- The request is cancelled, no?
+  if (!defined(promise)) {
+    return Promise.resolve(bufferLoader);
```
ggetz, line 114: This was one area I wasn't sure about, and could possible be an existing bug. It's possible for `BufferLoader._fetchArrayBuffer` to return `undefined` if the request scheduler is full, meaning the request needs to be retried later. But the existing behavior appeared to swallow the error.

lilleyse: `resource.throttle` should be false here so the promise should always be defined. Only the tile's main content (.glb or .gltf) is throttled currently. I edited one of the test tilesets to have an external bin file and confirmed that this is the case.

lilleyse: No, that issue is about improving throttling behavior for the things that are being throttled.

lilleyse:
Just to confirm, does this also fix https://github.com/CesiumGS/cesium/issues/10222?

ggetz:
> Just to confirm, does this also fix https://github.com/CesiumGS/cesium/issues/10222?

@lilleyse No. That will need changes in `Viewer.js`, and I'm aiming to keep this PR as small as possible.

ggetz:
@lilleyse This is ready for review. Could you take a look, or recommend a reviewer familiar with glTF and 3D Tile loading?

lilleyse (commented):
Thanks @ggetz this looks pretty good. It took me a little while to understand the `process` functions refactor but I see why it's done that way.

I wish we had avoided `defer` in the first place and saved you all this trouble :sweat_smile:

ptrgags (commented):
@ggetz I took a look at `Cesium3DTile` and `ModelExperimental`-related code. The approach looks good, just a couple comments from me.

Source/Scene/Cesium3DTile.js
```diff
@@ -1189,69 +1190,76 @@ function requestSingleContent(tile) {
...
   const tileset = tile._tileset;
   tile._contentState = Cesium3DTileContentState.LOADING;
-  tile._contentReadyToProcessPromise = defer();
-  tile._contentReadyPromise = defer();
   ++tileset.statistics.numberOfPendingRequests;
+  const contentReadyToProcessPromise = promise.then(function (arrayBuffer) {
+    if (tile.isDestroyed()) {
+      // Tile is unloaded before the content finishes loading
+      singleContentFailed(tile, tileset);
+      return;
```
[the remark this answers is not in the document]
ggetz: I actually ended up matching `multipleContentFailed` to `singleContentFailed` to be more explicit about what is being returned in turn of `undefined` vs a rejection.

Source/Scene/GltfLoader.js
```diff
@@ -231,40 +235,130 @@ GltfLoader.prototype.load = function () {
   this._textureState = GltfLoaderState.LOADING;
 
   const that = this;
-  gltfJsonLoader.promise
+  let textureProcessPromise;
+  const processPromise = new Promise(function (resolve, reject) {
+    textureProcessPromise = new Promise(function (
```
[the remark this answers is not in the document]
ggetz: I don't think this is possible. The `process` function needs access to the `resolve` and `reject` callbacks for both the master readyPromise and texture readyPromise. If we move the callbacks to member functions, then we just go back to the problematic `defer` pattern.

ggetz:
Thanks @lilleyse and @ptrgags! Updated!

lilleyse:
Latest changes look good. Thanks @ggetz.

PR #10377, opened 2022-05-13
Tweak Atmosphere Sandcastle example

mramato:
The Atmosphere Sandcastle disabled OIT for no reason. As we've learned in the past, this can lead to users copy and pasting it thinking they need it.


sanjeetsuhag (approved):
Looks good, @mramato. I tested the Sandcastle and everything is working as intended. I'll merge once CI passes.

PR #10384, opened 2022-05-16
Add 2D / CV support to tilesets using ModelExperimental 

### LINKED ISSUE #9934: `ModelExperimental` - Support rendering in 2D and Columbus View modes
Making a note to keep track of this, ModelExperimental does not yet make draw commands for 2D and Columbus View modes.

Without this, the model ends up out of view or renders incorrectly:

image

j9liu:
This PR adds 2D / CV projection to `ModelExperimental` with the `projectTo2D` option. This option is always set to true by tilesets, and it projects the positions more accurately than the old `Model` class. Here's a preview of the results:

| `Model` | `ModelExperimental` |
| -------- | ----------------------- |
| image | image |

**TODO:**
- [x] Add an option to the `ModelExperimental` constructor to enable accurate projections. This will default to false for models loaded in with `fromGltf`, but should always be true for models loaded in from a tileset.
   - [x] Decide on a name for this -> `enableProjection2D`? `projectTo2D`?
   - [x] This should control whether or not the vertex arrays should be loaded in.
- [ ] ~Add old `Model` projection code to `ModelExperimental` for models that aren't associated with a tileset. This allows models to show up in 2D / CV mode without breaking from options like `minimumPixelSize`~ -- will be done in a separate PR
- [x] Figure out lighting difference with `ModelExperimental` (only happens in 2D/CV)
- [ ] ~Fix instancing errors -- current method doesn't support instancing~ -- save for another PR
- [x] Unit tests

This is related to #9934, but the issue shouldn't be closed until `ModelExperimental` gets the old `Model` projection code for individual models (will be a separate PR).


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-experimental-2d-cv-support/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
> Fix instancing errors -- current method doesn't support instancing

One idea - maybe the instanced transforms get projected but the geometry doesn't? That way it's like a hybrid of the 3D Tiles approach and the Model approach.

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -446,4 +593,182 @@ function handleBitangents(shaderBuilder, attributes) {
...
+  const buffer = Buffer.createVertexBuffer({
+    context: frameState.context,
+    typedArray: projectedPositions,
+    usage: BufferUsage.STATIC_DRAW,
+  });
+  buffer.vertexArrayDestroyable = false;
+
+  // Store the reference point for repeated future use.
+  attribute.buffer2D = buffer;
+  attribute.referencePoint2D = referencePoint;
```
ptrgags, line 771: Hm, the typical pattern we've been using is to add generated resources to `model._resources`, and whenever the pipeline is rebuilt, those resources are `.destroy()-ed` to start over fresh. Makes it easier to manage the memory, but perhaps is more wasteful if things can be cached. We might want to discuss this more

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -446,4 +593,182 @@ function handleBitangents(shaderBuilder, attributes) {
   );
 }
 
+const scratchProjectedPosition = new Cartesian3();
+const scratchCartographic = new Cartographic();
+
+function projectPositionTo2D(position, modelMatrix, projection, result) {
```
ptrgags, line 599: Hm, this file is getting rather large. Though I do see it's a bit tricky to separate  the 2D details from the basic geometry details.

Though can any of these helper functions be moved to their own utility module for map projections?

ptrgags (commented):
@j9liu I did a light first pass review, pointed out some questions I have before I review it further.

I think one big question is when we consider the shader as a whole, what parts use the 3D position and what parts use the 2D position? If they all use one or the other, could `a_positionMC` store either and the only difference is the matrices?

Source/Shaders/ModelExperimental/GeometryStageVS.glsl
```diff
@@ -4,7 +4,14 @@ void geometryStage(inout ProcessedAttributes attributes, mat4 modelView, mat3 no
     vec3 positionMC = attributes.positionMC;
     v_positionMC = positionMC;
     v_positionEC = (modelView * vec4(positionMC, 1.0)).xyz;
+
+    #ifdef USE_2D_POSITIONS
```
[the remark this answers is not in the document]
j9liu: The lighting is currently incorrect so I think this is a mistake on my end. I'll have to see if using it for v_positionEC will change things.

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -334,14 +458,19 @@ function addAttributeDeclaration(shaderBuilder, attributeInfo) {
...
-  if (semantic === VertexAttributeSemantic.POSITION) {
+  const isPosition = semantic === VertexAttributeSemantic.POSITION;
+  if (isPosition) {
     shaderBuilder.setPositionAttribute(glslType, attributeName);
   } else {
     shaderBuilder.addAttribute(glslType, attributeName);
   }
+
+  if (isPosition && use2D) {
+    shaderBuilder.addAttribute("vec3", "a_position2D");
```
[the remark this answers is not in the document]
j9liu: I don't think this is an accurate name because "model coordinates" is supposed to be the positions with respect to the model. But the 2D coordinates are with respect to the world.

If you want I can name this `a_positionWC2D` ?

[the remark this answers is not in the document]
j9liu: Is it accurate to say they're in the model coordinate system? I know they're not in world space because they're with respect to the reference point, but how technical are we being here?

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -121,22 +141,46 @@ GeometryPipelineStage.process = function (renderResources, primitive) {
...
     );
+    const isPositionAttribute =
+      attribute.semantic === VertexAttributeSemantic.POSITION;
 
     let index;
     if (attributeLocationCount > 1) {
       index = renderResources.attributeIndex;
       renderResources.attributeIndex += attributeLocationCount;
-    } else if (attribute.semantic === VertexAttributeSemantic.POSITION) {
+    } else if (isPositionAttribute && !use2D) {
```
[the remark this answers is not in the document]
j9liu: No. In 2D mode location 0 is still the position attribute. see `addAttributeToRenderResources`

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -1343,6 +1358,21 @@ function getScale(model, frameState) {
...
+      const cartographic = projection.ellipsoid.cartesianToCartographic(
+        scratchPosition,
+        scratchCartographic
+      );
+      projection.project(cartographic, scratchPosition);
+      Cartesian3.fromElements(
+        scratchPosition.z,
+        scratchPosition.x,
+        scratchPosition.y,
+        scratchPosition
```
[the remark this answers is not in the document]
j9liu: This is how Columbus View coordinates are calculated in all other classes related to 2D / CV projection.

Source/Scene/ModelExperimental/GeometryPipelineStage.js
j9liu: I think I can repurpose `SceneTransforms.computeActualWgs84Position` here; it's basically the same thing, though I would have to force the `frameState.mode` to be `SceneMode.COLUMBUS_VIEW` so that the function doesn't drop the x-coordinate.

j9liu: I also wonder if it's possible to create the 2D buffer somewhere else, then just reference it in the `GeometryPipelineStage`. I'm not sure where that would go, though.

Source/Scene/ModelExperimental/ModelExperimental.js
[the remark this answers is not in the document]
ptrgags: On further inspection of the animation when you switch the view mode, the camera always looks at `(lon, lat) = (0, 0)`, which means the camera indeed is on the x-axis. So the map is laid out in the yz-plane.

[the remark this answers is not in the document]
j9liu: I personally am unfamiliar with the 2D / CV mode, but based on the comment below I'll be removing this function anyway.

Source/Scene/ModelExperimental/GeometryPipelineStage.js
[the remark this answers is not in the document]
ptrgags: I guess not, it's the 2D equiavelent of model space... okay, keep it as-is.

Source/Shaders/ModelExperimental/GeometryStageVS.glsl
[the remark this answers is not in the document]
ptrgags: Definitely curious to see the difference here since I'm not sure which looks better: as it stands it's like you're lighting in 3D space and projecting the results. The other way would be lighting the 2D scene after projecting.

The other thing to consider is how custom shaders should react to 2D mode

Source/Scene/ModelExperimental/ModelExperimental.js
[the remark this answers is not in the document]
ptrgags: Ah here's a way to visualize what I mean, see [this Sandcastle](https://sandcastle.cesium.com/#c=zRmLbhs38ldYoQVWiLy201yRk5zgFEVxgzoP2OkVRRVc6V1KIkIt97hcW3Kgf+o33JfdkFxyuS9LcpC2gGFzyXnPcB50xJNMohtKbolAz1BCbtGEZDRfhf/We8GsF+nvCU8kpgkRs15/NEsijSf5S7IQhGSAWqC9wXIZun2AnCXzPIkk5QlKscCMERYwLKnMYzJAEWdcDNBC4CRnWFC56aPPswQhwyDBKwK0Z733BSos0SNk8UcKUhCZi6TQISQJHFGShTiOA00JaSpD/XtgNlLONgyUGaICQm1lVAmZDa0mEywkrHDyfTgXfFVoNBYCb4LfLBpCR6dPTwblp1PNg/jnDoAdx7vQdwjwse+WtzSWyyF67DawiD5sUuJ0Hpvv8PLHn9+8cFArLImgmA0Ld9l9z2tD/6MA2Oq/WxUu20ocrIBaDIYNGE8WBwbCmwK3CARL4O8QCaU2Xb5rAdh17gfP38+TDF8TNuFcxDQB0lkQgak4EEiXNKo40Fq1zBQTDxQyhjNylcaoJGEjG0i4BFMBDi1EH+i9omsSB08qBKx976FgQTpIrAHVqhKu22E2PsymHebOh7nrEFcZF+B+B7WG6NvPVrvt//6YzRKQdAh7Vl6zCT9rBbrewmqjVhu1ulOru+3vOh3vc0+saEO3KiJBy+RdFknWIJvedcGULfntCxx9WgieJ/EQSZGXET2HOjKEy3v6JF2jFU94luKIzHo7cwb+RM7h6gdJvrom4t38Jb2hmbqmLncovtWYK4qGKk8K3xaR7JJEucjoDQkM1/KSuUWTjdnXvNTS95RNaJaRTVJNRmWuLld7snJMwjTPloFLooqQge33CwcbqUDXC0htSihniRCOIiwddmZoW9BwzsUUR8vA2T1Q2bFvHa4+QmVnoKn+jJy3vLiyxFp92OGDFU0u/MK1wuvKd0xSuRwY3bRVCpEYkcA2y5kEiX776PuEximnidoPPOpQNDzafXSMHhe1QxExpnV9iqVRty+do0BLhJ6jE2ccwzfjuVwSkVjT3xt4TcXNXsHWbhheR+jU7bjYsNHhCiUXh7Gvsaobfm/2zgsVA9h4M6eDqngGdVsJHgPYGjodt0pZsFI4lQqVjS8PHlc4HlWot4ePu5kPCZ9bksma+zqzSYvuXxo+BB/GviV6GvIcFj6+AerR40u3O3iUe1WSGkOBS3mM2fvCpXPMMuKGmFTQFZS4G62w9gX5b44lF8NyYIEMa7sWJXr44uLnqWnGFLYz0bBsbbsxMsJIJEmspXGldOVIQM0kc9Awti1qIUTzpCjG1W1TQXFF56/FhfHbS5Jxlqs7qqrzsKzTj2sGeD++mJ5fTqdvB8b+xhhLuoB00EnkHzUiL8eXP727HL89n3pUttUpM09jaEUngmfZElNhGjzdWcI97GpPK05RBdMFRVg58nIEroeVh1M9G9lLX+hSGDKoUA6tB/ouI9QbNEFWHG5fB9roACQbBwch6UDYgVFVfF/halh7SlfD8sTb+m2QzlAX+88OTRcTD/uoQm7UZHTIjNHGysMPahQfqaaxj75D3/9wUnA+Pkavk0hAYiQICqs/uwF7RFepAFMhHEHqxtEG3S5Jgu44X5EYUU92FZDi3EN+hk7Ck5NTT0IBKM+q1/Fy+nLUSGnO6V7fWVSMXaYf2Fpgc09drLIbbo9nVbJsBr6fpfPBg3na8ez+GbgSIdEGJ3UbTn4dv22Jg+pjXGVW9t4gCiUbwTOo7Vc7upNSpY772nRdByml0U67dVzvprO69HgIl073WCbWNR0m8MeccTOdd+h0GJaW7V4Uncngkk8YjT6pG50t6Vzqqx6pCnekS9wsKVJkBqmAJFdqpJ7eQL78EScxU/tEvk7SXI51ifSmuxXPMzveORoEhrwU+JnT0D4B+OlKYJUiCowIg/o4XBD5HrAu8eYexAXj16RENcz0pgfkKjYA6jMjDjAdVBD9hvqbWnF1NMp6arrEZoXwL2zH+5S6c9UXKkNbE9u741AOdR3NVc1X+rnuYvrqw38mF68nPxndtPN1gGiXO7vgNclsAytwTGkp95QxmmacxuEv51dPn4T62HW7eE2zK3qnXPCY/DBykdN8Baq9lTZeSr13UvcIWktZ5tE00BKE6wGCpvikfMO8FxiKnRW1wDNoHwv02gPoPs+f5eNnvYi5llY/ZfwJFgGFSqPsskgJXDXKV7OIadP/ApuUmu5hk1azfDWbqPmtapJZcgXZNYKRlBFljw98sWDkRS4lZNhZ70pd2mKSnPUG5hUUlZk3WpLoE4mL5OTNDgWOLQsF3OgArnomdbW1jbnOfYawew3xRKgMtV8gCMyHR8LNdtCc0lhJoye3A8RpjJmHPFOWYtt35d3SA2ihgZlPW5RQMEYPvepwa0WNtll3lybduhTa6FK2rz9c+2Fmlb190fqK4gvzgIm4YybuGiCdRR7Wrj20YXtAy6Y9on3TG/TOMrlh5Ll15r9gIONColywIAyPJVmlTDWlx9c52FKGUZbZoffs2Ec9i+kNovGzln/Ho4jhLIOTec6YyoGz3vOzY4BvoDIO+TJZvLshguGNAluePr8wm2EYnh3DZzum5JxdY1Gj/H8) and switch to 2D/CV while watching what happens to the x-axis (red), it ends up in the center. 

image

image

Source/Shaders/ModelExperimental/GeometryStageVS.glsl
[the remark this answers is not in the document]
j9liu: Maybe other users could weigh in, but personally I'm thinking... say I have a custom shader that shades models based on height. If I go into 2D mode, wouldn't I want my custom shader to be the same?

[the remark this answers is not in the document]
ptrgags: Indeed. If the original model coordinates was an ENU frame, then `positionMC.z` is the height. But then based on how the projection works, that height value is passed through unaltered, but it winds up in `position2D.x`, which would probably be surprising. So yeah, perhaps using the 3D coordinates in custom shader is indeed best here.

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -1,14 +1,23 @@
-import defined from "../../Core/defined.js";
-import ComponentDatatype from "../../Core/ComponentDatatype.js";
-import PrimitiveType from "../../Core/PrimitiveType.js";
 import AttributeType from "../AttributeType.js";
-import VertexAttributeSemantic from "../VertexAttributeSemantic.js";
+import BoundingSphere from "../../Core/BoundingSphere.js";
```
ptrgags, line 2: @j9liu and I discussed this offline, we think it's possible to move these changes to its own pipeline stage that just generates the single attribute (and related uniforms/defines). This way we avoid bloating the geometry stage which is already one of the more complex stages.

j9liu:
@ptrgags I think this is ready for another look!

Source/Scene/ResourceCache.js
```diff
@@ -536,8 +539,8 @@ ResourceCache.loadVertexBuffer = function (options) {
  * @param {Resource} options.baseResource The {@link Resource} that paths in the glTF JSON are relative to.
  * @param {Object} [options.draco] The Draco extension object.
  * @param {Boolean} [options.asynchronous=true] Determines if WebGL resource creation will be spread out over several frames or block until all WebGL resources are created.
- * @param {Boolean} [options.loadAsTypedArray=false] Load index buffer as a typed array instead of a GPU index buffer.
- * @param {Boolean} [options.loadForWireframe=false] Load index buffer as a typed array in order to generate wireframes in WebGL1. This will be ignored if using WebGL2.
+ * @param {Boolean} [options.loadAsTypedArray=false] Load index buffer as a typed array instead of a GPU index buffer. This is mutually exclusive with loadForWireframe.
+ * @param {Boolean} [options.loadForWireframe=false] Load index buffer as a typed array in order to generate wireframes in WebGL1. This is mutually exclusive with loadAsTypedArray.
```
ptrgags, line 543: Given my other comment, this documentation probably also needs to change.

ptrgags, line 543: Given my other comment, this documentation probably also needs to change.

Source/Scene/GltfVertexBufferLoader.js
```diff
@@ -37,6 +37,7 @@ import ComponentDatatype from "../Core/ComponentDatatype.js";
  * @param {Boolean} [options.asynchronous=true] Determines if WebGL resource creation will be spread out over several frames or block until all WebGL resources are created.
  * @param {Boolean} [options.dequantize=false] Determines whether or not the vertex buffer will be dequantized on the CPU.
  * @param {Boolean} [options.loadAsTypedArray=false] Load vertex buffer as a typed array instead of a GPU vertex buffer.
+ * @param {Boolean} [options.loadFor2D=false] Load vertex buffer as both a buffer and typed array, the latter of which will be projected to 2D.
```
ptrgags, line 40: Similarly to the indices, does it make sense to rephrase these lower-level options as `loadTypedArray` and `loadBuffer`? If so, such options would not be mutually exclusive.

ptrgags, line 40: Similarly to the indices, does it make sense to rephrase these lower-level options as `loadTypedArray` and `loadBuffer`? If so, such options would not be mutually exclusive.

Source/Scene/ModelExperimental/SceneMode2DPipelineStage.js
```diff
@@ -0,0 +1,284 @@
...
+      computedModelMatrix,
+      boundingSphere2D,
+      frameState
+    );
+
+    runtimePrimitive.positionBuffer2D = buffer2D;
+    model._modelResources.push(buffer2D);
+
+    // Unload the typed array
+    positionAttribute.typedArray = undefined;
```
ptrgags, line 94: Have we tested this with multiple models that share the same geometry?

ptrgags, line 94: Have we tested this with multiple models that share the same geometry?

Source/Scene/ModelExperimental/SceneMode2DPipelineStage.js
```diff
@@ -0,0 +1,284 @@
...
+    boundingSphere2D.center,
+    new Matrix4()
+  );
+  const modelView = new Matrix4();
+
+  const camera = frameState.camera;
+  const uniformMap = {
+    u_modelView2D: function () {
+      return Matrix4.multiplyTransformation(
+        camera.viewMatrix,
```
ptrgags, line 115: Should this be using `context.uniformState.view` instead? or was this intentional?

ptrgags, line 115: Should this be using `context.uniformState.view` instead? or was this intentional?

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -230,19 +268,44 @@ function addAttributeToRenderResources(
     renderResources.featureIdVertexAttributeSetIndex = setIndex + 1;
   }
 
+  // The position attribute should always be in the first index.
+  const isPositionAttribute = semantic === VertexAttributeSemantic.POSITION;
+  const index = isPositionAttribute ? 0 : attributeIndex;
```
ptrgags, line 273: I'm a little confused, from code earlier in this file, the 2D position has a semantic of POSITION right? is that returning 0 in this case?

Source/Scene/ModelExperimental/GeometryPipelineStage.js
```diff
@@ -230,19 +268,44 @@ function addAttributeToRenderResources(
...
+
+  if (!isPositionAttribute || !use2D) {
+    return;
+  }
+
+  // Add an additional attribute for the projected positions in 2D / CV.
+  const buffer2D = renderResources.runtimePrimitive.positionBuffer2D;
+  const positionAttribute2D = {
+    index: attributeIndex,
+    value: buffer2D ? undefined : attribute.constant,
```
ptrgags, line 298: Is there ever a case where buffer2D won't be defined here?

ptrgags (commented):
@j9liu did a pass through the code. Overall the architecture looks a lot cleaner now that the 2D logic is in its own pipeline stage!

I had some ideas about better managing buffers/typed arrays in the loaders, and a few other things here and there.

I still need to try some things out in Sandcastle, I'll do that next.

Source/Scene/GltfIndexBufferLoader.js
```diff
@@ -338,12 +337,7 @@ GltfIndexBufferLoader.prototype.process = function (frameState) {
     return;
   }
 
-  // WebGL1 has no way to retrieve the contents of buffers that are
-  // on the GPU. Therefore, the index buffer must be stored in CPU memory
-  // to generate wireframes for models.
-  const useWebgl2 = frameState.context.webgl2;
-  const loadTypedArrayForWireframe = !useWebgl2 && this._loadForWireframe;
-  if (this._loadAsTypedArray || loadTypedArrayForWireframe) {
+  if (this._loadAsTypedArray || this._loadForWireframe) {
```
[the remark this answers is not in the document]
j9liu: Yeah that might be better. I can rework this for `GltfIndexBufferLoader` and `GltfVertexBufferLoader`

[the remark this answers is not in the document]
j9liu: the WebGL2 check is handled in `GltfIndexBufferLoader`.

Source/Scene/ModelExperimental/SceneMode2DPipelineStage.js
j9liu: I have, and there weren't any problems that came up.

Source/Scene/ModelExperimental/GeometryPipelineStage.js
j9liu: No, I guess it doesn't make sense for it to be

Source/Scene/ModelExperimental/GeometryPipelineStage.js
j9liu: 2D position will always use `attributeIndex`. This just allows us to set the 3D position with an index of 0, and the 2D position as the next attribute index

ptrgags:
@j9liu I tried the Sandcastle links you gave me, noticed a couple things:

In [the model one](http://localhost:8080/Apps/Sandcastle/index.html#c=tVb7T+M4EP5XrP7Unoqb9AndLjooXY4T1SJaWOkup5VJ3MaHY0e2U+BW/O83sZM05bG3x4oKBT/G833z8IxDKbRBG0bvqEIfkaB3aEo1yxJ8bdeaQSO086kUhjBBVdBoo2+BQEjHMuPRkWAJMXSMjMpo266HVNC5jGCtULUoV/D08/nV/Phq8fX6bPYlEI+tD4EIhIPHIZfhLQ4zpagwS5ZQIFRo+D3jjIgTAMIrJZMzLfeHnt/M4YJG1+v6e76/5x0su964Oxr3Rtjzh96w3x8Nev394X7/oDf6I2gEwuF1Oihnw3UgQmt/YmcA92eu0VqHkKH3ZgzqrwRnBh3Le7Dc7cAZymkIu6tMhIZJgZqt8hg4wG5aBHAfxh34W5Ak5RQMIB0H1gGFVnM1wGtuVrl3vZxlrujR4rnvE1LOL+iYcC6leBOzYi334GscHUoBsjsDtjcVbP7zB+X4B+gvbpkQNELTmCgSGptV72jBnJTsv8LwGXXvfzA/Ab4STWWSKqo1mGC5vBd9i7YFK81h/HapMrgsRcq81RQp6Ps63kLUxz+XNqdKZiJC1zRmIX836g6lANmd/VTqnEGpISKEnHlrMfkB8qC6wtmZPM+V77n+L1soXXWM0yet4TdKIibWF8yE8aXkjlixNycmxkZeggQRuunvey2r1HNfV4Kd3hW7p9EnRRK6VCC7kirZVvxqSWNoC4Q7OfmpOnMKPUURI1XRBYRUJi7sCxp3VJttwa88W/dmpngbxZStY1P429FKpWZWuOIyJcrAiIie7T4ndK0o1UU49vxuD3ujfn/oHxTO7fexN/B6I29YLDiUfOz4oKLjYtsrcapYApAbqrGiidzQI/Cpi0mtPwGf106RKCrYlFHID8zuUwoy0EwJt8RPIQOaVWaB/eP8U2WERYH4KXY/fiEM8ZOo14NRS83SfbVEgwSqzQrNM85ZqiWL8JfTxX6/JvBCXlSJWomlSv4NoVzK7ol7exQZ3Kp72VoEPiXRwwXYzzTFJqaiub1pVqK6buVDBKAVwSv+sJTH+e0HsxdpTNXWTqf5Zmez4vZta0uUQY4C0ri4AO4nVytN4ca/dqeIWNdd+trl6g9a7f+W2vOfiL3EHSuQzjT6BQ22oq1y+OgGpVtLFdC5If3cCxBstHmYp25lPzwU0vFOTlbC57CFL2cXs6NluwydzXj77zEHWhARhUQbKLugeCklvyFqTkXmoqYdnUa7MdHmgdPDEvVXlqRQDPLcbkKNNBRqJLwcdecGWiY1ONS6LHiTTv3oJGIbxKKPL7x5UciJ1rCzyjhfsH+g+RxOOiD/7CiXNpCfN1Rx8pCLxf7huVvEGE86MH35pHEWPtH8Lw) the drone propellers and ground vehicle's wheels aren't rotated correctly. Is this expected? (I know some details will be in a future PR)
image
image

Also I notice the animation doesn't seem to happen in 2D/CV, is that expected behavior?

j9liu:
@ptrgags found the error, it was because nodes that used the same mesh were modifying the same typed array, instead of calling `slice` on it before making changes.

Here's a [sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=zVdtT+M4EP4rVj+1p+ImbWhLt4sOCstxAi2ihZX2clqZxCU+HDuynQK34r/fxE7SF8reLiukrarUL+N5nnnJjBtJoQ1aMHpPFXqPBL1HE6pZnuJru9YMG5GdT6QwhAmqwkYbfQ0FQjqROY8PBEuJoSNkVE7bdj2igp7LGNZKVdNqBU8+nl2dH15Nv1yfHn8KxVPrXShC4eBxxGV0h6NcKSrMjKUUCJUa/sw5I+IIgPBcyfRUy2Hf85sFXNjoel1/x/d3vL1Z1xt1B6PeAHt+3+sHwWC3Fwz7w2CvN/gcNkLh8DodVLDhOhSRtT+1M4D7q9BorUPI0AczAvVXgjODDuUDWO524AzlNILdeS4iw6RAzVZ1DBxgNy0CuA/jDnynJM04BQNIx4F1QKHVXA/wLTfzwrtewbJQ9GTx3HODlPMLOiScSylexaxcKzz4EkeHUoKsz4DtTQ1bfPzdavwd9Kd3TAgao0lCFImMzao3tOCcVOy/wPAZde8HmB8BX4kmMs0U1RpMsFzeir5FW4JV5jB+N1M5vCxlyrzWFCno2zreQqyOfy5tTpTMRYyuacIi/mbUHUoJsj77qdQ5hVJDRAQ589pi8h3kQXWNszZ5nivfcv3ftlC66phkG63hD0piJm4vmImSS8kdsXLvnJgEG3kJEkTopj/0Wlap556uBDu9c/ZA4w+KpHSmQHYuVbqs+PWSxtAWCHdy8kN95gR6iiJGqrILCKlMUtoXNu6pNsuCX3t21Zu54m2UUHabmNLfjlYmNbPCNZcJUQZGRPRs9zmit4pSXYZjx+/2sDcIgr6/Vzo3CLC36/UGXr9ccCjF2Hn6OVj3R9GGa1jBVqDit+yttinjTLEU4BZUY0VTuaAHEDxHaaURApWXTpE4LolU4S4OHD9kFGSgaxNuOZ9AqjXrFAZHj4pHnXoWBRJFsYfRlngnG+m1GvWVd6By3UpGQ6auzErNx5yzTEsW408n02GwIrAlAes3ohbLlPwHcmYmu0fuklO+Kq31cMKd4iCOIcHAkbGz8Fvu/xUd2f01PGmNgvwk8eMFuIBpik1CRXNZHq1EXSOr2yOAK4Ln/HEmD4uSDZZPs4SqpalO883aZs3u69KaOIfCAkijsmq5j5zPNYUy/VIhJOJ21asvVcRgt9X+f6kdf0NsG3esQDrX6De0uxRtVcMnN6gKQaUCrluQge7aDjbaVCzKQG0/3O6y0Vpa1sJnsIUvjy+OD2btKng2bvbnqQCaEhFHRBvolaB4JiW/IeqcitxFTTs6jXZjrM0jp/sV6u8szaCCF+ndhMZmKDQ2uO7rzg3cc6jBkdZVlxp3Vo+OY7ZALH6/5Y8KijjRGnbmOedT9i/cGPbHHZB/dpRLG8iPC6o4eSzEEn//zC1ijMcdmG4/aZyFG5r/Aw) for testing two projected instances of the same model.

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -67,6 +69,7 @@ import SplitDirection from "../SplitDirection.js";
  * @param {ShadowMode} [options.shadows=ShadowMode.ENABLED] Determines whether the model casts or receives shadows from light sources.
  * @param {Boolean} [options.showCreditsOnScreen=false] Whether to display the credits of this model on screen.
  * @param {SplitDirection} [options.splitDirection=SplitDirection.NONE] The {@link SplitDirection} split to apply to this model.
+ * @param {Boolean} [options.projectTo2D=false] Whether to accurately project the model's positions in 2D mode. This disables minimumPixelSize and prevents future modification to its model matrix. This cannot be set after the model has loaded.
```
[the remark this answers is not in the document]
j9liu: This will be true when I do the PR for individual models, but it's not true now. Until then, though, I'll add a note that the model won't show up without this flag enabled.

Source/Scene/GltfIndexBufferLoader.js
```diff
@@ -338,12 +342,7 @@ GltfIndexBufferLoader.prototype.process = function (frameState) {
     return;
   }
 
-  // WebGL1 has no way to retrieve the contents of buffers that are
-  // on the GPU. Therefore, the index buffer must be stored in CPU memory
-  // to generate wireframes for models.
-  const useWebgl2 = frameState.context.webgl2;
-  const loadTypedArrayForWireframe = !useWebgl2 && this._loadForWireframe;
-  if (this._loadAsTypedArray || loadTypedArrayForWireframe) {
+  if (!this._loadBuffer) {
```
ptrgags, line 345: We discussed this offline, we decided to reorganize this function a little bit to avoid the duplicate code and early return. See https://github.com/CesiumGS/cesium/pull/10397/files#diff-d58e7af8b44d5b26b11ae14d06508206d2b7a07722ac16bb3c1546fafdef8da8 for how I did it in my PR

Source/Scene/ModelExperimental/SceneMode2DPipelineStage.js
```diff
@@ -0,0 +1,284 @@
...
+      positionAttribute,
+      computedModelMatrix,
+      boundingSphere2D,
+      frameState
+    );
+
+    runtimePrimitive.positionBuffer2D = buffer2D;
+    model._modelResources.push(buffer2D);
+
+    // Unload the typed array
```
ptrgags, line 93: Could we reword this to explain that when the typed array is shared, this doesn't cause issues because positionAttribute.typedArray is just a pointer to the array in the vertex buffer loader? This wasn't obvious.

Source/Scene/ModelExperimental/SceneMode2DPipelineStage.js
```diff
@@ -0,0 +1,284 @@
...
+  for (let i = startIndex; i < length; i += 3) {
+    const initialPosition = Cartesian3.fromArray(result, i, scratchPosition);
+
+    const transformedPosition = Matrix4.multiplyByPoint(
+      modelMatrix,
+      initialPosition,
+      initialPosition
+    );
+
+    const projectedPosition = SceneTransforms.computeActualWgs84Position(
```
ptrgags, line 230: I was trying the [San Francisco Ferry Building Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Next%20Photogrammetry%20Classification.html&label=3D%20Tiles%20Next) and this line seems to hit a divide by zero error when I zoom in to the tileset.

image

ptrgags (commented):
@j9liu looks pretty close, though I noticed a bug when 

Also I tried the [S2 Base Globe](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Next%20S2%20Globe.html&label=3D%20Tiles%20Next) in 2D/CV mode, it's mostly working (aside from the IDL wrapping, I know that's a future PR.)

The one thing I'm uncertain of is whether the S2 bounding volumes are handled correctly in 2D. Not sure how bounding volume intersections are handled in projected coords. Maybe take a quick look and see if this is correct behavior or something we need to open an issue for?

image

ptrgags:
This morning @j9liu and I looked at the code for bounding volumes in 2D - it looks like Cesium3DTile always uses a bounding sphere around the projected positions for 2D, so that probably explains why it doesn't react the way I was expecting. I don't think there's any action needed.

ptrgags:
@j9liu I can confirm that you fixed the `NaN` issue.

I notice that for OSM buildings, NYC is working fine, but Tokyo or Sydney don't show up. Though I think that's beyond the scope of this PR, probably has more to do with tile bounding volumes than `ModelExperimental`-specific stuff. Could you open an issue?

I'm merging this now, I know everything else will be handled in future PRs

PR #10395, opened 2022-05-23
Issue #10394: Compute boundingSphere correctly for models created from glTF

### LINKED ISSUE #10394: Model boundingSphere uses wrong forwardAxis and is transformed incorrectly
When the _ModelUtility.computeBoundingSphere_ function is called for a new model, some transformations are applied to the bounding sphere in ModelUtility.js:

https://github.com/CesiumGS/cesium/blob/46ed67c0bf601b52f3b1acdb7bb13d5decc5a181/Source/Scene/ModelUtility.js#L270-L277

However, I don't think the above code ever gets executed in the current version of Cesium (v.1.93). This is because the if statement is calling model._forwardAxis rather than model.forwardAxis, which would take the glTF version 2.0 into account.

Following the breadcrumbs back to where _forwardAxis is set, the property is set as part of undocumented options that are not listed as possible parameters in the CesiumJS documentation:

https://github.com/CesiumGS/cesium/blob/46ed67c0bf601b52f3b1acdb7bb13d5decc5a181/Source/Scene/Model.js#L600

This results in _forwardAxis always being undefined. Calling it directly when boundingSphere is computed does nothing.

You can see that the other part of the code that uses forwardAxis (linked below) calls the property correctly. These parts of the code even have the exact same comment:

https://github.com/CesiumGS/cesium/blob/46ed67c0bf601b52f3b1acdb7bb13d5decc5a181/Source/Scene/Model.js#L5537-L5543

this.forwardAxis is defined in Model.js in a way that indeed acknowledges the glTF 2.0 specification:

https://github.com/CesiumGS/cesium/blob/46ed67c0bf601b52f3b1acdb7bb13d5decc5a181/Source/Scene/Model.js#L1072-L1079

I have been able to verify on my own that something odd happens with the bounding sphere of a model that is created from a glTF. I suspect that removing the underscore on line 270 of ModelUtility.js would fix this issue.

VsKatshuma:
This is a pull request aiming to fix #10394. When glTF version 2.0 support was added for models, most likely the only part of the code that was tested was the modelMatrix that determines where the models are drawn, but the logic for bounding spheres was broken.

In current version of Cesium, models created from glTFs that are model-to-world transformed have their bounding spheres in a different place than where the models actually are.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@
 
 ##### Fixes :wrench:
 
+- Fixed the inaccurate computation of bounding spheres for models not centered at (0,0,0) in their local space. Models were rendered correctly previously, but their bounding spheres ended up in the wrong place. [#10394](https://github.com/CesiumGS/cesium/issues/10394)
 - Fixed the inaccurate computation of bounding spheres for `ModelExperimental`. [#10339](https://github.com/CesiumGS/cesium/pull/10339/)
 - Fixed race condition which can occur when updating `Cesium3DTileStyle` before its `readyPromise` has resolved. [#10345](https://github.com/CesiumGS/cesium/issues/10345)
 - Fixed label background rendering. [#10342](https://github.com/CesiumGS/cesium/issues/10342)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,7 +16,7 @@
 
 ##### Fixes :wrench:
 
-- Fixed the inaccurate computation of bounding spheres for models not centered at (0,0,0) in their local space. Models were rendered correctly previously, but their bounding spheres ended up in the wrong place. [#10394](https://github.com/CesiumGS/cesium/issues/10394)
+- Fixed the inaccurate computation of bounding spheres for models not centered at (0,0,0) in their local space. [#10395](https://github.com/CesiumGS/cesium/pull/10395)
 - Fixed the inaccurate computation of bounding spheres for `ModelExperimental`. [#10339](https://github.com/CesiumGS/cesium/pull/10339/)
 - Fixed race condition which can occur when updating `Cesium3DTileStyle` before its `readyPromise` has resolved. [#10345](https://github.com/CesiumGS/cesium/issues/10345)
 - Fixed label background rendering. [#10342](https://github.com/CesiumGS/cesium/issues/10342)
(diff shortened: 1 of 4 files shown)


cesium-concierge:
Thanks for the pull request @VsKatshuma!
* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-gltf-bounding-sphere/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
@lilleyse Would you be able to confirm this? See https://github.com/CesiumGS/cesium/issues/10394 for the rationale.

j9liu:
Hi @VsKatshuma,

Thanks for the PR! Can you sign an individual [Community License Agreement](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) before I review this PR?

VsKatshuma:
My GitHub user should be under a newly submitted Corporate CLA. Do I also need to sign an individual agreement despite this?

I had not commented on this yet since there seems to be one test that's failing for me. I'm currently looking into it.

Regardless, I'm a bit concerned. Suppose I'm right and computing the bounding sphere has always skipped one transformation. But what if Cesium's users have understood this as the intended functionality, and have constructed their solutions accordingly. Then "fixing" this issue in Cesium has the potential to break other apps.

Then again, the functionality should ideally be correct at the source. I thought about applying the missing transformation in our project, but thought it would be more proper to correct the issue in Cesium instead.

j9liu:
Hi @VsKatshuma,

My mistake, I didn't realize you put it under the corporate CLA. You're good to go.

I think what you have is correct. It's similar to the model matrix update code in `update`, and the transformations that `computedModelMatrix` undergoes should be similar to the ones for the bounding sphere. https://github.com/CesiumGS/cesium/blob/46ed67c0bf601b52f3b1acdb7bb13d5decc5a181/Source/Scene/Model.js#L5524-L5544

If we log this change in `CHANGES.md` as a bugfix, then I think people can adapt their code accordingly if the change breaks their programs.

j9liu:
@VsKatshuma The `renders with model matrix` test failure has been noted for a while, so no worries. I wonder if it's possible to add a unit test for this change in `ModelSpec`, though?

Yes, please feel free to add the line to CHANGES.md yourself!

j9liu (commented):
@VsKatshuma just one suggestion for the changelog! Everything else looks fine. I went through several sandcastles using `Model` to make sure that they behaved the same and I didn't see anything unusual.

j9liu:
@VsKatshuma Merging now; thanks for the contribution!

Issue #10373, opened 2022-05-12
Revisit CesiumJS build tooling


ggetz:
Main motivations behind taking a more modern approach to the build process:
 * Bundling modules by default to improve loading performance in the browser. It's likely that this will workaround https://github.com/CesiumGS/cesium/issues/9416 which is caused by the sheer number of individual modules when we run a Cesium app.
 * Avoid manually copying over third-party dependencies
  * Preprocessing will allow us to use [ES6+ language features](https://github.com/CesiumGS/cesium/issues/9718) more freely without sacrificing browser support

Factoring in [community feedback and our own research](https://community.cesium.com/t/cesiumjs-build-tooling-revamp/17757), we plan on using [esbuild](https://esbuild.github.io/) to build, due to
- Speed, which is crucial for decreasing developer friction when re-building after changes have been made
- Relatively minimal configuration when compared with tools like webpack, which we've run into some fragility with in the past (for instance https://github.com/CesiumGS/cesium/issues/7669), and [vite](https://vitejs.dev/) (which uses esbuild under the hood). We won't be able to make much use out of the extra features these tools offer, such as hot module reloading.
- TypeScript support [if we chose to go that direction](https://github.com/CesiumGS/cesium/issues/4434)
- Our experience using esbuild internally in other projects

As part of this, we'll want to:

* [bundle](https://esuild.github.io/api/#bundle) Cesium.js 
    * Remove `buildThirdParty` from the `build` task, and instead include npm modules directly, as we do in `ThirdParty/npm`
    * Can we further simplify the glsl to js build step?
* Build the Workers as separate entry points 
    * Remove requirejs from cesiumWorkerBootstrapper and use [`importScripts`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts) directly (which I think is one of the last actionable items in https://github.com/CesiumGS/cesium/issues/9473)
* Add a `watch` task to rebuild when code changes are made (make use of caching to [build incrementally](https://esbuild.github.io/api/#incremental)?). 
    * I'm thinking we'd also want to break down the build talk into ints components and only re-build workers, shaders, or source depending on what changes.
* replace rollup+terser to bundle for release and [minify](https://esbuild.github.io/api/#minify), including copyright header as a banner
* Strip code between debug pragmas... We may need to write our own plugin (similar to [strip-pragma-loader](https://github.com/CesiumGS/strip-pragma-loader) for webpack)
[...]

Issue #9906, opened 2021-11-01
Regression: sampleTerrain no longer works in NodeJS


bkuster:
# Issue
`sampleTerrainMostDetailed` no longer works. Error received: `require is not defined`.

# Reasoning
`require` is no longer defined within an ESM module.

# Reproducable
```javascript
import { CesiumTerrainProvider, Cartographic, sampleTerrainMostDetailed } from "cesium";

const provider = new CesiumTerrainProvider({ url: 'https://berlin.virtualcitymap.de/datasource-data/5d2c8614-d474-442d-8975-c2b4a4c0ae94/' });
sampleTerrainMostDetailed(provider, [Cartographic.fromDegrees(0, 0)])
  .then((updated) => {
    console.log(updated);
  })
  .otherwise((err) => {
    console.error(err);
  });
```

# Fix/Workaround
Use deferred import statements, as can be seen in this commit: https://github.com/virtualcitySYSTEMS/cesium/commit/25f59d6fb7635d3f0547b19419fa47294a02ea38

The question is, if cesium can support this for all its platforms? If so, I'm happy to open a PR.

hpinkos:
Thanks for reporting this @bkuster!  @ebogo1 we should look at this ASAP since it's a regression

ebogo1:
@bkuster I am able to reproduce this issue in CesiumJS 1.73 on Node 14 and 16. Which versions of CesiumJS and Node did you notice this regression in?

bkuster:
@ebogo1 I just stumbled across this issue today using Node 12 and cesium 1.79 (but I would assume the regression is older) while updating my module to ESM. I could also reproduce it in Node 14 & 16 (using 1.79). To verify, I then created a new package with the above code and used the latest cesium from npm & Node 14.

ebogo1:
I don't believe there's a workaround without revisiting how we package CesiumJS for Node - this issue will happen with any file that happens to call `loadWithHttpRequest` in Resource.js.

xtassin:
I am still getting a "is not defined" exception when trying to call Cesium.sampleTerrainMostDetailed from a NodeJS Application with a locally built Cesium 1.97.0

I use require to declare Cesium in my application such as:
const Cesium = require("./cesium");

Most other code methods seem to be working fine.

Issue #9388, opened 2021-02-23
Keep third-party code license under minification?


pjcozzi:
Third-party code use to always have the license in a comment at the top of the file that was not removed during minification.

For example, see [Tween.js](https://github.com/CesiumGS/cesium/blob/master/Source/ThirdParty/Tween.js)

>/**
>@license
>tween.js - https://github.com/sole/tween.js
>Copyright (c) 2010-2012 Tween.js authors.

Looks like that is no longer always the case, e.g., see the Draco dependencies in [ThirdParty/Workers](https://github.com/CesiumGS/cesium/tree/master/Source/ThirdParty/Workers).

Draco is, however, mentioned in [LICENSE.md](https://github.com/CesiumGS/cesium/blob/master/LICENSE.md#draco).

What is the latest policy?  Why?  And is Cesium consistently following it?

kring:
I don't think minified builds include license comments at all anymore, since I broke it in #8659. But even if they did, the Draco licenses wouldn't be included because they aren't marked with `@license`, so we wouldn't be able to tell them apart from other comments that we definitely do want removed in minified builds.

I think it's reasonably common practice to consider a minified build to be a compiled form of the code, which requires that the license information be _with_ it (i.e. in LICENSE.md), but not _in_ it. If you disagree I can work on bringing back the license comments, at the cost of a bigger Cesium.js. I didn't intentionally remove them before, it just happened as a result of a tooling change, but it may not be trivial to bring them back.

kring:
Actually, if I'm not mistaken, my change in #8659 only removed the Cesium copyright header from the top, not the third-party licensing info. If we ever had the third-party info in minified builds, I'm not sure where it went to.

Issue #8245, opened 2019-10-03
Generate Source Maps


mramato:
Now that we have switched over to roll-up with #8224, it should be straightforward to generate sourcemaps for our code.  The only tricky part may be our use of pragmas which remove code from the "built" version.  But roll-up hasn't let me down yet so there's a good chance it will "just work"

kring:
#8659 adds sourcemaps for the unminified build. I tried adding them for the minified too, but it didn't go well. The debugger ended up in totally wrong locations.

Removing uglify fixed it, but I'm not sure why uglify would be the problem because sourcemaps are supposed to work with rollup+uglify.

To try it, add this to the end of `rollup-plugin-strip-pragma`'s `transform` function:

```
            if (options.sourcemap !== false) {
                result.map = magicString.generateMap();
            }
```

And change the two places in `gulpfile.cjs` with `"sourcemap": debug` to `"sourcemap": true`.

fredj:
I haven't tried but the latest version of magic-string fixes a bug related to the source map:
https://github.com/Rich-Harris/magic-string/blob/master/CHANGELOG.md

PR #10399, opened 2022-05-24
Streamline build processes with esbuild

### LINKED ISSUE #10400: Custom Shader Guide included in Build folder
After running `npm run makeZipFile` the Custom Shader Guide gets added to the Build folder.

The quickest fix would be to remove `Source/Scene/ModelExperimental/CustomShaderGuide/README.md` since it just points to the new location at `Documentation/CustomShaderGuide/README.md`. Otherwise we'd have to modify the gulpfile so that it doesn't copy this file.

Screen Shot 2022-05-25 at 10 38 03 AM

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/10373
Fixes #10400
Fixes https://github.com/CesiumGS/cesium/issues/6464
Fixes https://github.com/CesiumGS/cesium/issues/9906
Fixes https://github.com/CesiumGS/cesium/issues/8245
Fixes https://github.com/CesiumGS/cesium/issues/6993
Fixes https://github.com/CesiumGS/cesium/issues/9388
Should fix https://github.com/CesiumGS/cesium/issues/7669, verification needed
Workaround for https://github.com/CesiumGS/cesium/issues/9416

### How to build

For a single build, the workflow is the same as before:

```
npm install
npm run build
npm start
```

For an ongoing build task which incrementally rebuilds based on file changes, run:

```
npm install
npm run-build watch
```

and as a sperate process run:

```
npm start
```

For this to be truly a "just in time" build such that the build is run on `npm start`, we would want to run the equivalent of `build-watch` in `server.js`. In my mind at least, that would require a bit of a refactor of `gulpfile.cjs` so that the common build functions (a majority of the file) is either factor out into a separate file, or exported. Given we want to follow up this PR with  revisions to the build scripts such as [naming](https://github.com/CesiumGS/cesium/issues/10313), I thinks it best to defer that change to a subsequent PR.

`build` now encompasses the functionality of both `combine` and `build-specs`, so they have been removed.

Ideally, for both simplicity and performance, we we use esbuild for workers as well, but we see the end result is significantly increased in size (1.1MB -> 10.1MB minified). I believe this is because there's no code splitting going on. [However, code splitting is not supported for IIFE with esbuild](https://github.com/evanw/esbuild/issues/16), and [Firefox still does not support ESM modules in workers](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687). Potentially, we could output ESM with code splitting, and then post-process with rollup. This PR leaves building workers as is. When the Firefox support is there, it may make sense to holistically address our worker strategy.

### Performance Stats
| script | this branch | esbuild for workers | `main` |
| --- | --- | --- | --- |
| `build` (previously `combine`+ `build-specs`) | ~7s |  ~4s |  ~25s + ~11s = ~36s |
| `build-watch` rebuild | 300ms - 800ms | 300ms - 800ms | ~6.5s (and stuck in an infinite re-build loop) |
| `build --minify --pragmas` (`minifyRelease`) | ~7s | ~4.5s | ~40s  |
| `buildApps` | ~2s | ~2s  | ~30s |

### Size Stats

| | this branch | `main` |
| --- | --- | --- |
| minified | 3.6 MB | 4.3 MB |
| unminified | 8.4MB | 12.5 MB |
| source map | 17.2 MB | 22 MB | 

### Highlights

* Bundles and minifies Cesium.jswith esbuild, resulting in faster build and build-watch times.
    * CI times have gone from 15 min -> 12 min. The bottleneck here isn't build/bundle time so much as tests and building documentation. 
* Removes manual ThirdParty build/copy step; Instead use esbuild and resolve from `node_modules` directly
* There are no more AMD/UMD style modules; Instead the build output is IIFE for the browser, generic ESM modules, or CJS for node.
    * Other projects, like three.js, ship separate files for separate targets
    * CJS needs to be built differently in order to run (we use the "node"  target esbuild), CC https://github.com/CesiumGS/cesium/issues/9906
* Due mostly to removing jsdoc comments, the unminified distribution is much smaller. We also get a slight benefit to the minified version as well. 
* Make sure `@legal` comments _are not_ removed during bundling or minification.
* Cut down on redundant build tasks; Instead we have fewer tasks with command line options (which shouldn't be needed as part of normal development). Also use a consistant kabab naming convention.
* Refine granularity of `build-watch` task, only rebuilding certain aspects depending on the file changed
* Remove `convertToModules`. This task was put in place for the ES6 module transition a few years ago, and we don't have a need to maintain it going forward.
* Test no longer use the `Source` code directly. Instead, they use the unminified IIFE build with source maps by default. Despite this, debugging still works as before since the debugger will handle the mapping back to the original source code. Therefor, there's no longer any need to have a `build` option or query anymore. The `release` option instead loads the minified IIFE build with pragmas removed instead, pretty much the same as before.
* Update build guide. Notably, minimum recommended Node version is now documented as 14. Adds `engine` field to `package.json` to warn about this.
* One big change in terms of ingesting Cesium is that, while you can still load the raw `Source` modules, we no longer build the workers to the `Source/Workers` directory. If a user want to do this, they should see the `CESIUM_BASE_URL` to one of the built directories. I'm not 100% sure this is the right decision.
* For coverage, we now instrument up-front using the Istanbul instrumenter directly, as opposed to using a karma plugin. This gives us more control, and allows us to build the code _after_ instrumentation rather than instrumenting the bundled output (so we don't need a karma plugin anymore).

Known TODOS:
- [x] Release packaging
	- Update `makeZipFile` task to reflect new files
	- Are the output files all named appropriately?
	- Should CSS files have a copyright header?
- [x] Smooth out build task granularity, and update build guide
- [x] Some workers are not loading correctly
- [x] Specs are not rebuilding on watch
- [x] Update testing guide
- [x] Include GLSL build in `build-watch` task
- [x] Can we add a quick spec to make sure Cesium is working in .cjs/node?
- [x] Build warnings when running on travis? This worker code shouldn't be entered.
    ````
      > node_modules/@zip.js/zip.js/lib/core/codecs/codec-pool-worker.js:110:51: warning: "import.meta" is not available in the configured target environment ("es2015") and will be empty
  		return new Worker(new URL(workerData.scripts[0], import.meta.url), options);
  
  > node_modules/@zip.js/zip.js/lib/core/codecs/codec-pool-worker.js:110:51: warning: "import.meta" is not available in the configured target environment ("es2015") and will be empty
  		return new Worker(new URL(workerData.scripts[0], import.meta.url), options);
   ````
- [x] Build Sandcastle
- [x] Update remove RequireJS from credited third party libraries
- [x] coverage shouldn't include third party modules
- [x] Make sure third party licensing is not removed during minification (https://github.com/CesiumGS/cesium/issues/9388)
- [x] Do an optimization pass; esbuild is fast, but biggest bottlenecks will be our JS plugins
- [x] The workers have significantly increased in size (1.1MB -> 10.1MB minified). I believe this is because there's no code splitting going on. [However, code splitting is not supported for IIFE with esbuild](https://github.com/evanw/esbuild/issues/16), and [Firefox still does not support ESM modules in workers](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687). Potentially, we could output ESM with code splitting, and then post-process with rollup. Or we can rollback to rollup for web workers entirely for the time being until the Firefox support is there, at which time it may make sense to holistically address our worker strategy.
- [x] Update `CHANGES.md`


ggetz:
@sanjeetsuhag I'm still making a few tweaks based on the final TODO's, but most of this PR should be static. Could you give this a try and let me know if you have any general feedback?

Documentation/Contributors/BuildGuide/README.md
```diff
@@ -120,39 +131,41 @@ npm run [target-name]
...
 - **Utility scripts** -- code coverage, static code analysis, and other utilities
-  - `coverage` - Runs coverage and opens the default browser with the results.
-  - `eslint` - Runs [ESLint](http://eslint.org/), a static code analysis tool, on the entire source tree.
-  - `eslint-watch` - A never-ending task that watches your file system for changes to Cesium and runs ESLint on any changed source files.
-  - `clean` - Removes all generated build artifacts.
+  - `clean` - Removes all generated build artifacts
   - `cloc` - Runs [CLOC](https://github.com/AlDanial/cloc) to count the lines of code on the Source and Specs directories. This requires [Perl](http://www.perl.org/) to execute.
+  - `coverage` - Runs coverage and opens the default browser with the results
+  - `eslint` - Runs [ESLint](http://eslint.org/), a static code analysis tool, on the entire source tree
+  - `eslint-watch` - A never-ending task that watches your file system for changes to Cesium and runs ESLint on any changed source files
```
sanjeetsuhag, line 150: This isn't a valid command in this branch. In VSCode, we use the extension, but I suppose something like this would be helpful for people who prefer coding in other tools.

sanjeetsuhag (commented):
@ggetz It's amazing how much faster all the scripts are working. Can't wait to see how much of an impact this will have on CI. The debugging workflow for workers is much easier now too! Otherwise, I just had a few comments.

Documentation/Contributors/BuildGuide/README.md
```diff
@@ -56,7 +56,7 @@ _NOTE: If you change branches, you might have to rebuild._
 
 Prerequisites:
 
-- Install [Node.js](http://nodejs.org/) on your system. Building Cesium requires Node 6.x or newer.
+- Install [Node.js](http://nodejs.org/) on your system. Building Cesium requires Node 14.x or newer.
```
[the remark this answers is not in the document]
ggetz: CC https://github.com/CesiumGS/cesium/issues/6993

gulpfile.cjs
```diff
@@ -116,111 +81,289 @@ const filesToClean = [
...
-    rimraf.sync(file);
+/**
+ * Bundles the workers and outputs to the Source/Workers directory
+ * @param {Object} options
+ * @param {boolean} [options.minify=false] true if the worker output should be minified
+ * @param {boolean} [options.sourcemap=false] true if a sourcemap should be generated
+ * @param {String} options.path true if the worker output should be minified
+ * @returns {Promise.<Object>} esbuild result
+ */
+async function createWorkers(options) {
```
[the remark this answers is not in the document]
ggetz: This is happening via the sourcemap. The existing sourcemaps generated before this PR also include maps for third party code, but it was copied over to the `ThirdParty` directory as a separate step beforehand. Now we are bundling code straight from `node_modules`. Other projects show something similar, for example, here's BablyonJS:

image

ggetz:
Thanks @sanjeetsuhag! Your feedback should all be addressed.

mramato:
There's no explicit mention of how we tested this. For example:

* Did we compare the unminified output of the build process in this branch to be compared directly with `main` to look for any obvious issues? I know esbuild has different rules for how it brings in npm modules.
* Did we compare overall bundle sizes to make sure minification was still working as expected? 
* What about performance?

I haven't thought about it much, but I'm sure there are more checks we'll want to do and post the results of.

> * Removes rollup

* It looks like all of the npm rollup dependencies are still there

mramato:
| this branch | main
-- | -- | --
minified | 3.6 MB | 4.3 MB
unminified | 8.4MB | 12.5 MB
source map | 17.2 MB | 22 MB

These numbers are so good that they are concerning. Do we understand how/why the unminified version so much smaller compared to `main`? What exactly changed?  These numbers alone would warrant a major improvement in CHANGES and a blog post, so we need to be sure to highlight them.

mramato:
I'm also confused as to the new developer workflow and I think that warrants a summary in the PR.  My assumption is that all I would need to do is

```
npm install
npm start
```

And everything would "just work" since I assume esbuild is now "just-in-time" (this is how other repositories without a build step we have work).

I tried instead running

```
npm run build-watch
```

But immediately ran into the error I posted a few comments up.

package.json
```diff
@@ -116,28 +111,22 @@
   },
   "scripts": {
     "prepare": "gulp prepare && husky install",
-    "convertToModules": "gulp convertToModules",
     "start": "node server.cjs",
-    "startPublic": "node server.cjs --public",
+    "start-public": "node server.cjs --public",
```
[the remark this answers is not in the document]
ggetz: Gotcha. The breakdown of some of these build task no longer made sense after the changes, so I went ahead and did a pass on everything. I'll see what makes sense to break off into a separate PR.

ggetz:
> But immediately ran into the error I posted a few comments up.

@mramato What was the error you were running into? Not sure I see the comment.

mramato:
@ggetz sorry, not sure what happened to my previous comment: 

```
$ npm run make-zip

> cesium@1.94.3 make-zip
> gulp make-zip

[13:28:08] Using gulpfile ~/Git/CesiumGS/cesium/gulpfile.cjs
[13:28:08] Starting 'make-zip'...
[13:28:08] Starting 'release'...
[13:28:08] Starting '<anonymous>'...
▲ [WARNING] Top-level "this" will be replaced with undefined since this file is an ECMAScript module [this-is-undefined-in-esm]

    Source/ThirdParty/Workers/draco_decoder_nodejs.js:2:510:
      2 │ ...[n];if(l&&l.Math==Math)return l}throw Error("Cannot find global object");};$jscomp.global=$jscomp.getGlobal(this);
        │                                                                                                                ~~~~
        ╵                                                                                                                undefined

  This file is considered to be an ECMAScript module because the enclosing "package.json" file sets
  the type of this file to "module":

    package.json:42:10:
      42 │   "type": "module",
         ╵           ~~~~~~~~

✘ [ERROR] Transforming async generator functions to the configured target environment ("es2015") is not supported yet

    node_modules/@zip.js/zip.js/lib/core/zip-reader.js:90:1:
      90 │   async* getEntriesGenerator(options = {}) {
         ╵   ~~~~~

[13:28:09] '<anonymous>' errored after 1.16 s
[13:28:09] Error: Build failed with 1 error:
node_modules/@zip.js/zip.js/lib/core/zip-reader.js:90:1: ERROR: Transforming async generator functions to the configured target environment ("es2015") is not supported yet
    at failureErrorWithLog (/home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:1605:15)
    at /home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:1251:28
    at runOnEndCallbacks (/home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:1164:65)
    at buildResponseToResult (/home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:1249:7)
    at /home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:1358:14
    at /home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:666:9
    at handleIncomingPacket (/home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:763:9)
    at Socket.readFromStdout (/home/mamato/Git/CesiumGS/cesium/node_modules/esbuild/lib/main.js:632:7)
    at Socket.emit (node:events:527:28)
    at Socket.emit (node:domain:537:15)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at Pipe.onStreamRead (node:internal/stream_base_commons:190:23)
    at Pipe.callbackTrampoline (node:internal/async_hooks:130:17)
[13:28:09] 'release' errored after 1.16 s
[13:28:09] 'make-zip' errored after 1.16 s
```

ggetz:
> These numbers are so good that they are concerning. Do we understand how/why the unminified version so much smaller compared to main? What exactly changed? 

A big part of the unminified bundle sizes decreasing is that in this version comments and whitespace are removed. For instance, `DeveloperError` becomes:

````js
  // Source/Core/DeveloperError.js
  function DeveloperError(message) {
    this.name = "DeveloperError";
    this.message = message;
    let stack;
    try {
      throw new Error();
    } catch (e2) {
      stack = e2.stack;
    }
    this.stack = stack;
  }
  if (defined_default(Object.create)) {
    DeveloperError.prototype = Object.create(Error.prototype);
    DeveloperError.prototype.constructor = DeveloperError;
  }
  DeveloperError.prototype.toString = function() {
    let str = `${this.name}: ${this.message}`;
    if (defined_default(this.stack)) {
      str += `
${this.stack.toString()}`;
    }
    return str;
  };
  DeveloperError.throwInstantiationError = function() {
    throw new DeveloperError("This function defines an interface and should not be called directly.");
  };
  var DeveloperError_default = DeveloperError;
````

Comments shouldn't be needed in the bundle since we do include the sourcemap.

Additionally, in both minified and unminified versions, it appears additional whitespace in the shaders has been removed.

Overall, [esbuild minification appears to be very comparable to terser](https://github.com/privatenumber/minification-benchmarks/tree/cd3e5acb8d38da5f86426d44ac95974812559683#readme), and the difference we are seeing here appear to be due to whitespace and comments.

**Comment by ggetz** [comment]
```
✘ [ERROR] Transforming async generator functions to the configured target environment ("es2015") is not supported yet

    node_modules/@zip.js/zip.js/lib/core/zip-reader.js:90:1:
      90 │   async* getEntriesGenerator(options = {}) {
         ╵   ~~~~~
```

The above error [arose because of a recent zipjs update](https://github.com/gildas-lormeau/zip.js/issues/334), which uses async generator syntax. For now, we can opt to lock the package version to the previous one, but we can also consider upgrading the target output to ES2018, which has been supported by our supported browsers for a few years now. I'm leaning towards the later option.

**Comment by mramato** [comment]
> ES2018

Cesium should probably be on ES2020 at this point. That's what the eslint work @sanjeetsuhag is doing is moving too anyway.

**Comment by ggetz** [comment]
@sanjeetsuhag Could you please give the latest updates a round of review? Assuming this is in a good place, it would help to get this merged and move https://github.com/CesiumGS/cesium/pull/10538 along as well.

File: Documentation/Contributors/BuildGuide/README.md [hunk 5]
```diff
@@ -119,40 +134,43 @@ npm run [target-name]
...
+    - `--minify` - [Minifies](<http://en.wikipedia.org/wiki/Minification_(programming)>) the output for optimized loading. Specifying this option will output to `Build/Cesium`.
+    - `--removePragmas` - Optimizes the output by removing debugging code that validates function input and throws `DeveloperError`s. The removed sections are marked with `//>>includeStart('debug', pragmas.debug);` blocks in the code.
+    - `--node` - Bundles an `index.cjs` module targeted for use in NodeJS
+  - `build-watch` - A never-ending task that watches your file system for changes to Cesium and builds the source code as needed. All `build` options are also available for this task.
+  - `minify` - Bundles bundles the source modules, plus [minifies](<http://en.wikipedia.org/wiki/Minification_(programming)>), to produce an all-in-one files in the `Build/Cesium` directory.
+  - `combineRelease` - Bundles plus removes debugging code that validates function input and throws DeveloperErrors. The removed sections are marked with `//>>includeStart('debug', pragmas.debug);` blocks in the code.
```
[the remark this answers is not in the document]
sanjeetsuhag: Thinking about this a little more, I don't know of a better term here. This is basically a Release build without minification, but `release` is already used up. So maybe its fine this way.

sanjeetsuhag (commented):
@ggetz This is very close. I just had some minor comments. One big thing in running the code - it seems like the `build` step is slow.

On this branch, I ran `git clean -xdf; npm install; npm run build` and the `build` step took 5.84s. I switched to `main` and ran the same command and the `build` step took 5.04s.

Documentation/Contributors/BuildGuide/README.md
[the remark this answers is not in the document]
ggetz: Not really, but it's outside of the scope of this PR to add and remove the other build tasks. I'm simply updating what's already here with the state as of this PR.

package.json
```diff
@@ -154,6 +151,9 @@
     "prettier": "prettier --write \"**/*\"",
     "prettier-check": "prettier --check \"**/*\""
   },
+  "engines": {
+    "node": ">=14.0.0"
```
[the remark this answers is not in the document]
ggetz: Updates this in https://github.com/CesiumGS/cesium/pull/10538/files, but I will bring it back here as well.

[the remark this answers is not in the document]
ggetz: This is the minimum version which will work, not necessarily the recommended version- https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

ggetz:
Thanks @sanjeetsuhag! This is updated.

sanjeetsuhag (approved):
Everything seems to be in good shape. This is an awesome improvement! Thanks @ggetz! We should also update the Release Guide since it references outdated steps like `build-specs`.

CHANGES.md
```diff
@@ -5,6 +5,10 @@
 ##### Breaking Changes :mega:
 
 - Tilesets and entities now use `ModelExperimental` by default. This can still be changed by setting `ExperimentalFeatures.enableModelExperimental = false`. [#10530](https://github.com/CesiumGS/cesium/pull/10530)
+- Built `Cesium.js` is no longer AMD format. This may or may not be a breaking change depending on how you use Cesium in your app. See our [blog post](TODO) for the full details. [#10399](https://github.com/CesiumGS/cesium/pull/10399)
+  - If you were ingesting individual ESM-style modules from the combined file `Build/Cesium/Cesium.js` or `Build/CesiumUnminified/Cesium.js`, instead use `Build/Cesium/index.js` or `Build/CesiumUnminified/index.js` respectively.
+  - `CESIUM_BASE_URL` should be set to either `Build/Cesium` or `Build/CesiumUnminfied`.
+  - Built `Cesium.js` has gone from `12.5MB` to `8.4MB` unminified and from `4.3MB` to `3.6MB` minified. `Cesium.js.map` has gone from `22MB` to `17.2MB`.
```
mramato, line 11: This should probably be absolutely top of the list with :mega:, it's not a breaking change and it's a huge deal.

CHANGES.md
```diff
@@ -5,6 +5,10 @@
 ##### Breaking Changes :mega:
 
 - Tilesets and entities now use `ModelExperimental` by default. This can still be changed by setting `ExperimentalFeatures.enableModelExperimental = false`. [#10530](https://github.com/CesiumGS/cesium/pull/10530)
+- Built `Cesium.js` is no longer AMD format. This may or may not be a breaking change depending on how you use Cesium in your app. See our [blog post](TODO) for the full details. [#10399](https://github.com/CesiumGS/cesium/pull/10399)
+  - If you were ingesting individual ESM-style modules from the combined file `Build/Cesium/Cesium.js` or `Build/CesiumUnminified/Cesium.js`, instead use `Build/Cesium/index.js` or `Build/CesiumUnminified/index.js` respectively.
```
mramato, line 9: Doesn't Cesium now 100% require a bundler to use modules? (And a bundler that allows for node resolution rules at that). Or are we still embedding third party code?

ggetz: We're embedding the third-party code in these builds. In the blog post I mention that production apps that use their own bundler should be using modules directly `Source`, which does not have third party code bundled. Which is what we've previously recommended as well if I understand correctly.

mramato: We have previously recommended it, but since we embedded ThirdParty code in source, you didn't _need_ a bundler to use them. Now you do.

Another problem I just noticed is that `package.json` lists _everything_ as a dev dependency, which may not be correct because that means if I `npm install Cesium` and then try to use `Source` it's going to pull in modules that don't exist because they are not a `dependency` or `peerDependency`.

ggetz: Gotcha, I see the distinction. For the point about the bundler, I will update both `CHANGES.md` and the blog post with the requirement.

For the latter, I will `package.json` such that modules in `Source/ThirdParty` are now marked as direct dependencies. Sound correct?

thw0rted: Hey all, speaking of `package.json`, would this be a good time to address https://github.com/CesiumGS/cesium/issues/9212 as well?  It's really important to get the exports right, I had a workaround but had to get rid of it to be able to use [another important dependency](https://github.com/angular/components/issues/25305) at all...

ggetz: Hi @thw0rted, thanks for pointing this out! We have some additional thoughts on related improvements beyond just this PR, including considering https://github.com/CesiumGS/cesium/issues/9212. We'll add our thoughts or any updates to that issue.

ashley-mort:
Does this or will this provide any mechanism to preload the workers up front?  Loading ~50 web worker files on our network kills us by delaying user polygon drawing.

thw0rted:
I was going to say that your application might be able to improve the situation using `link rel="preload"` tags, but [at least in Chrome, maybe not](https://bugs.chromium.org/p/chromium/issues/detail?id=946510).  Might help in other browsers though?  I have not benchmarked, but would be interested in learning what provides the best user experience..

ashley-mort:
The [inlining static assets in a single distributable](https://github.com/CesiumGS/cesium/issues/10619) would be great.  I have tried out the c137.js stuff and it a huge improvement for us regarding the time it takes to display user created polygons.

PR #10417, opened 2022-06-01
Fixes the extension check required for OIT

sanjeetsuhag:
See https://github.com/CesiumGS/cesium/issues/9827 for more discussion.

Based on the information in this [WebKit thread](https://bugs.webkit.org/show_bug.cgi?id=240919), this PR adds a check for [`EXT_float_blend` ](https://www.khronos.org/registry/webgl/extensions/EXT_float_blend/) when enabling OIT. It also removes the restriction on OIT for iOS devices added as a workaround in https://github.com/CesiumGS/cesium/pull/10337.

| Before | After |
| - | - |
| photo_2022-06-01_09-58-00 | photo_2022-06-01_09-57-55 |

*Tested on an iPhone XR running iOS 15.5 using WebGL via Metal.*

### DIFF
--- a/Source/Scene/Scene.js
+++ b/Source/Scene/Scene.js
@@ -15,7 +15,6 @@ import destroyObject from "../Core/destroyObject.js";
 import DeveloperError from "../Core/DeveloperError.js";
 import EllipsoidGeometry from "../Core/EllipsoidGeometry.js";
 import Event from "../Core/Event.js";
-import FeatureDetection from "../Core/FeatureDetection.js";
 import GeographicProjection from "../Core/GeographicProjection.js";
 import GeometryInstance from "../Core/GeometryInstance.js";
 import GeometryPipeline from "../Core/GeometryPipeline.js";


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/ext-color-buffer-float-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

CHANGES.md
```diff
@@ -18,6 +18,7 @@
 
 ##### Fixes :wrench:
 
+- Fixed the rendering issues related to order-independent translucency on iOS devices. [#10417](https://github.com/CesiumGS/cesium/pull/10417)
```
sanjeetsuhag, line 21: Not sure if this is how we should phrase it or if it needs to go into breaking changes (since the workaround initially went in breaking changes)

ggetz:
@sanjeetsuhag I think you have one eslint error to clean up.

@lilleyse Could you confirm the fix?

PR #10424, opened 2022-06-02
Fix REPLACE styling for glTF without pbrMetallicRoughness object

lilleyse:
We noticed that `Cesium3DTileColorBlendMode.REPLACE` wasn't working for certain tilesets. Specifically, if a glTF material omitted the `pbrMetallicRoughness` object then the style would not get applied.

Why....? The code uses a special `_3DTILESDIFFUSE` semantic to figure out which attribute or uniform should be replaced with the feature color. Usually it attaches itself to the vertex color attribute, base color texture, or base color factor, but in this case none of those could be found.

The fix was to add a uniform for the default `baseColorFactor` value. The semantic gets added to that and everything works again.

As a side note, this bug doesn't seem to affect `ModelExperimental` which handles things very differently.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### 1.95 - 2022-07-01
+
+##### Fixes :wrench:
+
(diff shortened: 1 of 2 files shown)


ptrgags:
Looks good, thanks @lilleyse!

PR #10427, opened 2022-06-03
Fix vector tile crash by making sure billboardCollection is defined

ggetz:
Fixes a regression from https://github.com/CesiumGS/cesium/pull/10361 which can cause vector tile points to crash under certain circumstances.

I'd like to verify this works with an example, but in the meantime, I think this should fix the issue. The exception itself is in `Vector3DTilePoints.pointsLength` where it's trying to access an undefined `this._billboardCollection`.  I don't think there is a reason to wait to define the collection until after the internal promise resolves.


ggetz:
@lilleyse I believe this is working, could you please take a look?

lilleyse:
At some point we should get the `Vector3DTileContentSpec` tests running again, but this looks like a good fix and cleans up some of the logic.

PR #10428, opened 2022-06-03
Adds eslint and markdownlint to pre-commit hook

sanjeetsuhag:
⚠️ Depends on https://github.com/CesiumGS/cesium/pull/10411 being marged into `main` first. Once that is done, the destination branch can be changed to `main`.

In order to reduce cases where ESLint errors are caught too late in CI process, this PR adds `eslint` and `markdownlint` to the pre-commit hook.

If the `eslint` cache is not built, this means that the first commit will take a while to run, however, subsequent commits will be fast.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,5 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx pretty-quick --staged
+npm run eslint
+npm run markdownlint


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
@sanjeetsuhag I couldn't get markdownlint to fail on an invalid commit. Did you have any luck?

Also, eslint takes a few seconds to run on the whole codebase. Plus it runs on files which may not have have been staged for the commit. [`lint-staged`](https://github.com/okonet/lint-staged) is a package you can use to configure the precommit hooks to only run on the staged files.

sanjeetsuhag:
@ggetz I was able to get it to work. Just to double check, did you run `npm install` in case you switched branches?

Screen Shot 2022-06-03 at 2 15 28 PM
 
I'll take a look at `lint-staged` and update the PR.

package.json
```diff
@@ -151,5 +151,9 @@
     "deploy-set-version": "gulp deploy-set-version",
     "prettier": "prettier --write \"**/*\"",
     "prettier-check": "prettier --check \"**/*\""
+  },
+  "lint-staged": {
+    "*.{js,cjs,html}": ["pretty-quick", "eslint --cache --quiet"],
```
ggetz, line 156: Maybe we should run eslint before bothering to run prettier. If there's an error, we'll want to bail out early, right?

package.json
```diff
@@ -151,5 +151,14 @@
     "deploy-set-version": "gulp deploy-set-version",
     "prettier": "prettier --write \"**/*\"",
     "prettier-check": "prettier --check \"**/*\""
+  },
+  "lint-staged": {
+    "*.{js,cjs,html}": [
+      "eslint --cache --quiet",
+      "pretty-quick"
```
ggetz, line 158: @sanjeetsuhag I notices this is taking much longer than before. I think the `pretty-quick` package is no longer needed since we added `lint-staged`. I think this line can become `prettier --write` instead.

PR #10456, opened 2022-06-14
Only build-ts once in travis

ggetz:
It looks like we're currently running `build-ts` twice in CI. Once by itself and once as part of `makeZipFile`. It's not a dependency for any task that comes after it, so I don't see why we shouldn't remove the first run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ script:
   - npm --silent run prettier-check
 
   - npm --silent run build
-  - npm --silent run build-ts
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/build-ts-once/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #10433, opened 2022-06-06
Add support for projecting instanced models to 2D in `ModelExperimental`

j9liu:
This PR adds support for projecting instanced models and `.i3dm` tilesets to 2D / CV when they are loaded with `ModelExperimental`. Most of this code is done in `InstancingPipelineStage`, though let me know if it seems better to move this to `SceneMode2DPipelineStage` instead.

Here's a side-by-side comparison with `Model`: with `ModelExperimental` the trees load into the correct places 🌳 

| Model | ModelExperimental |
| ------ | ---------------------- |
| i3dm | i3dmexperimental |

Sandcastles:
[...]

PR #10471, opened 2022-06-21
Remove unused parameter in `InstancingPipelineStage`

j9liu:
In #10433, I unknowingly left an unused parameter in a function call, after refactoring the function to not need it anymore. This PR removes it.

@ptrgags can you take a look?

### DIFF
--- a/Source/Scene/ModelExperimental/InstancingPipelineStage.js
+++ b/Source/Scene/ModelExperimental/InstancingPipelineStage.js
@@ -660,8 +660,7 @@ function processTransformAttributes(
     transforms = getInstanceTransformsAsMatrices(
       instances,
       count,
-      renderResources,
-      use2D
+      renderResources
     );
 
     const transformsTypedArray = transformsToTypedArray(transforms);


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/remove-unused-parameter/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
@j9liu looks good, merging!

javagl:
Sorry, that comment in the other PR by @gpu was by me (I had stumbled over this while doing something completely unrelated). And maybe I should have mentioned that there are more instances of this problem in this file, namely useless '...Scratch' arguments to constructors...

- https://github.com/CesiumGS/cesium/blob/faedcb13be97a564f582cf63aad956b6e1bf0601/Source/Scene/ModelExperimental/InstancingPipelineStage.js#L542 
- https://github.com/CesiumGS/cesium/blob/faedcb13be97a564f582cf63aad956b6e1bf0601/Source/Scene/ModelExperimental/InstancingPipelineStage.js#L561
- https://github.com/CesiumGS/cesium/blob/faedcb13be97a564f582cf63aad956b6e1bf0601/Source/Scene/ModelExperimental/InstancingPipelineStage.js#L568

... and more generally, I noticed this by browsing a page that reports 36 (!) instances of this problem across 21 (!) files: https://lgtm.com/projects/g/CesiumGS/cesium/alerts/?mode=tree&ruleFocus=4860081 . Maybe this should be addressed systematically, I just addressed that in an issue.

javagl:
The cases where a "Superfluous trailing argument" is reported are cases where the argument is *really* superfluous (i.e. where the function _really_ does not have that parameter, not even as the optional 'scratch' parameter).

---

<sup>I tried to sort some of my thoughts here, and it turned into a 'wall of text'. Feel free to ignore the following. If this should be summarized somewhere else than in a random PR comment, drop me a note...</sup>

The usage and implementation of the 'scratch' parameter seems to cause some confusion. There may be 4 "kinds" of functions, and we can count from 0 to 3 in binary:

1. The `result` parameter is optional. The return value may be `undefined`
2. The `result` parameter is optional. The return value is never `undefined`
3. The `result` parameter is required. The return value may be `undefined`
4. The `result` parameter is required. The return value is never `undefined`

Case 2 is the classical use of 'scratch' parameters. The point is to move the responsibility of the "memory management" to the caller, giving the option to avoid unnecessary allocations and GC, and that's a reasonable (see 'Details about 2.' below). 

Cases 1 and 3 raise the question _under which condition_ the return value is `undefined`. Returning `undefined` from such a function seems to be some sort of 'error indicator', but it can be difficult to strictly define the actual behavior of such a function (see 'Details about 3.' below)

For cases 2. and 4., it would be nice if it was possible to make the assumption that if the `result` parameter was given, then the return value is *identical* to the `result` parameter. This, however, would make the return value unnecessary: These would exactly be the cases where the given `result` parameter can be _guaranteed_ to be modified by the function call.

An aside: It _can_ make sense to require the `result = compute(..., result)` convention nevertheless, because it leaves the freedom of _changing the implementation later_. Right now, a function could _always_ return the given result. In the future, it might be changed to _always_ create a _new_ result. Code that uses the `result = compute(..., result)` convention would still work afterwards, while code that just used `compute(..., result)` might be broken by such a change. 

(Follow-up questions may thus be: How strict does Cesium want to be in terms of guarantees about the behavior and implementation of such functions? How to 'educate' clients to use this pattern? How to _enforce_ that pattern to be used in the core codebase?)

---

Details about 2.:

If a function like `Cartesian3.add` always returned a new object, then a loop like
```
for (let i=0; i<1000; i++) {
    result = Cartesian3.add(result, array[i]);
}
``` 
would allocate 1000 useless objects, whereas
```
for (let i=0; i<1000; i++) {
    Cartesian3.add(result, array[i], result);
}
``` 
would not allocate any object at all. But still, the user may choose to use the benefit of automatic allocations. This benefit is _convenience_ on the one hand, and _scoping_ on the other hand (it's not always clear what the best scope of 'scratch' variables is. Having a long list of `const cartesianForLine42OfFunctionX = new Cartesian3()` can be cumbersome, and should be weighted with the performance benefit that one has <strike>to expect</strike> measured with a profiler run).

----

Details about 3.:

A function with a (required or optional) `result` parameter may return `undefined` to indicate some sort of error. And this can be a justification for the statement in the Coding Guide, which is

> Because result parameters aren't always required or returned, don't strictly rely on the result parameter you passed in to be modified. For example:
> ```
> Cartesian3.add(v0, v1, result);
> Cartesian3.add(result, v2, result);
> ```
> is better written as
> ```
> result = Cartesian3.add(v0, v1, result);
> result = Cartesian3.add(result, v2, result);
> ```

This might be elaborated or extended with an example that makes sense: In the given example, the returned value will **always** be the given result parameter (as of [the current implementation](https://github.com/CesiumGS/cesium/blob/646274019c9ae68a75d7e69d6324ea9a2737e15e/Source/Core/Cartesian3.js#L523) ), so that example does not bring the point across.

But it raises one question regardless: 
```
let result = undefined;
result = Something.computeSomething(..., result);
if (!defined(result)) {
    // Is it 'undefined' because of an error, or because 
    // I passed 'undefined' as the 'result' in the first place?
}
```

PR #10485, opened 2022-06-27
Removes eslint-config-cesium, enables ECMA2020 usage

sanjeetsuhag:
This PR removes the `eslint-config-cesium` source from this repository and instead uses the package that is independently hosted in [ts own repository](https://github.com/CesiumGS/eslint-config-cesium).

This PR is dependent on the merging of https://github.com/CesiumGS/eslint-config-cesium/pull/1 and publishing of the `9.0.0` package.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/eslint-test/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Documentation/Contributors/CodingGuide/README.md
```diff
@@ -111,7 +111,7 @@ A few more naming conventions are introduced below along with their design patte
...
+For syntax and style guidelines, we use the ESLint recommended settings (the list of rules can be found [here](http://eslint.org/docs/rules/)) as a base and extend it with additional rules via a shared config Node module, [eslint-config-cesium](https://www.npmjs.com/package/eslint-config-cesium). This package is maintained as a part of the Cesium repository and is also used throughout the Cesium ecosystem. For a list of which rules are enabled, look in [index.js](https://github.com/CesiumGS/cesium/blob/main/Tools/eslint-config-cesium/index.js), [browser.js](https://github.com/CesiumGS/eslint-config-cesium/blob/main/index.jshttps://github.com/CesiumGS/eslint-config-cesium/blob/main/browser.js), and [node.js](https://github.com/CesiumGS/eslint-config-cesium/blob/main/node.js).
```
ggetz, line 114: Looks like the `browser.js` link has a typo in the url.

Source/.eslintrc.json
```diff
@@ -1,14 +1,14 @@
...
+        "cesium/browser"
     ],
     "plugins": [
         "html",
         "es"
     ],
     "rules": {
-        "no-unused-vars": ["error", {"vars": "all", "args": "none"}]
+        "no-unused-vars": ["error", {"vars": "all", "args": "none"}],
+        "no-loss-of-precision": "off"
```
ggetz, line 12: Why is this set as a global rule rather than inline?

Specs/.eslintrc.json
```diff
@@ -4,7 +4,8 @@
         "jasmine": true
     },
     "rules": {
-        "no-self-assign": "off"
+        "no-self-assign": "off",
+        "no-duplicate-imports":"off"
```
ggetz, line 8: I know that a lot of our specs have duplicate imports due to the fact that we migrated from es6 modules and replaced the requirejs-style imports via script. However, I think this is something we'd like to clean up. 

That being said, it's probably best to do that in a follow up PR and a single commit for merging sake.

(I don't think there's an automatic `--fix` option for this rule unfortunately, but I think there's plugin you could use locally to make the change easier.)

ggetz (commented):
Thanks @sanjeetsuhag !

Documentation/Contributors/CodingGuide/README.md
sanjeetsuhag: Fixed.

Specs/.eslintrc.json
sanjeetsuhag: Ok, I'll look into that next.

sanjeetsuhag:
@ggetz Updated.

server.cjs
```diff
@@ -76,7 +76,7 @@
   });
 
   function checkGzipAndNext(req, res, next) {
-    const reqUrl = url.parse(req.url, true);
+    const reqUrl = url.URL(req.url, true);
```
ggetz, line 79: This is causing some internal server errors when running Sandcastles such as [3D Tiles Adjust Height](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Adjust%20Height.html).

You'll need to add the `new` keyword and also the second parameter is not a boolean, [it's a string which specifies the base URL](https://nodejs.org/api/url.html#class-url).

Also, for the to be a bit cleaner, you can use a destructor to only import the parts of `url` that we need

````js
const { format, URL } = require("url");
````

File: server.cjs [hunk 5]
```diff
@@ -169,7 +172,7 @@
 
     request.get(
       {
-        url: url.format(remoteUrl),
+        url: remoteUrl.toString(),
```
sanjeetsuhag, line 175: @ggetz Since the `url.format` is marked as legacy, I think just using `toString()` is fine here, given how the URL is built previously. Not sure if there are cases where this may fail though.

ggetz: :+1: There's a [non-legacy `format`](https://nodejs.org/api/url.html#urlformaturl-options), but looks like that without any options it will produce the same results as `toString`.

ggetz:
Awesome, thanks @sanjeetsuhag!

PR #10488, opened 2022-06-28
Remove polyfills for Array methods

jjhembd:
The polyfill methods `Core/arrayFill` and `Core/arraySlice` are no longer needed, since the JavaScript methods [Array.fill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) and [Array.slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) are now implemented in all supported browsers.

This PR removes the methods and related specs, and replaces all references in `Source` and `Spec` with the JavaScript methods.


lilleyse:
@jjhembd could you also update https://github.com/CesiumGS/gltf-pipeline to use `Array.fill` instead of `arrayFill`?

Afterwards I can publish a new version of gltf-pipeline so that it doesn't break when CesiumJS 1.95 is released.

sanjeetsuhag (approved):
Everything looks good! Thanks @jjhembd!

PR #10490, opened 2022-06-28
Add `asynchronous` option to `ModelExperimental.fromGltf`

j9liu:
This PR adds `options.asynchronous` to `ModelExperimental.fromGltf` to match the public API of `Model`. This option was always true by default, but it allows the user to set it to false if wanted.

There are differences between the two implementations, especially because `ModelExperimental`'s architecture is more synchronous compared to `Model`. The main difference is that loading the buffers and textures through `GltfLoader` will be affected by the `asynchronous` flag.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-experimental-asynchronous-loading/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -2182,6 +2184,7 @@ ModelExperimental.fromB3dm = function (options) {
     arrayBuffer: options.arrayBuffer,
     byteOffset: options.byteOffset,
     releaseGltfJson: options.releaseGltfJson,
+    asynchronous: options.asynchronous,
```
ptrgags, line 2187: Can this go in `makeModelOptions()`? is anything else common enough to move to that function? I think I've been forgetting to check.

ptrgags (commented):
@j9liu looks good, just had one comment about refactoring options in the `fromXxx` constructors

Source/Scene/ModelExperimental/ModelExperimental.js
j9liu: @ptrgags I don't think it can because the `asynchronous` field needs to be passed to the loaders (`B3dmLoader, I3dmLoader`). It doesn't actually get passed to the `ModelExperimental` constructor.

ptrgags: :facepalm: you're right, my bad.

ptrgags:
Opened https://github.com/CesiumGS/cesium/issues/10493 for the test failure, that's not relevant to this PR, I can reproduce it on `main`

PR #10495, opened 2022-06-29
Fix TextureManager test failure on Linux

### LINKED ISSUE #10493: Test failure with resizing textures in `TextureManager`
On the latest `main` using Chrome or Firefox on Linux, I'm seeing the following 2 test failures:

image.

Might be related to #10482 which was recently merged.

@j9liu was not able to reproduce this on Windows, so it might be an OS difference.

CC @jjhembd

jjhembd:
Resolves #10493.

Removes an extra `/` in the URL to one of the testing datasets, which made the in-browser SpecRunner fail with an HTTP 404 on Linux.


ptrgags:
Looks good, thanks @jjhembd!

PR #10496, opened 2022-06-29
Removes defer from imagery and terrain provider classes

sanjeetsuhag:
This PR removes usage of `defer.js` from the `*ImageryProvider.js`* files.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/remove-defer-imagery/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz (commented):
Thanks @sanjeetsuhag!

While this does get rid of one more use of `defer` directly, it's not great that this does not truly remove [the anti-pattern](https://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it). I think we'd want to have `requestMetadata` create and return a new promise rather than handling existing resolve and reject callbacks. 

As discussed offline, `TileProviderError.handleError` was not created with callbacks in mind, not promises. It should not call a retry function at all, and should likely be deprecated.

However, we could get around this for now by not providing `retryFunction` to the calls to `handleError` (and so passing in the new reolve and reject parameters shouldn't be needed). Instead, we could call `handleError` and check the `error.retry` result afterwards. Then, since promises are compossible, when `retry` is true, we can return another call to `requestMetadata`. 

For example, in `Source/Scene/ArcGisMapServerImageryProvider.js`, that would look like:
````js
function metadataSuccess(e) {
    ...
   metadataError = TileProviderError.handleError(metadataError,
              that,
              that._errorEvent,
              extentMessage);
   if (metadataError.retry) {
        return requestMetaData();
   }
   return Promise.reject(new RuntimeError(extentMessage));
}

function metadataFailure(e) {
    ...
    return Promise.reject(new RuntimeError(message));
}

function requestMetadata() {
    ...
    return resource.fetchJsonp().then(metadataSuccess).catch(metadataFailure);
}

this._readyPromise = requestMetadata();
````

Does that make sense?

If we don't chose to go that route, we should have an issue written up for how to address the error handling more holistically for ImageryProviders.

**Comment by sanjeetsuhag** [comment]
@ggetz Thanks for the explanation, it definitely helped clear things up! I believe the latest changes reflect the required architecture. Additionally, it also adds the changes to the `*TerrainProvider.js` classes. I have a couple of test related bugs that I am working through, but the core code is ready for a review pass.

**Comment by ggetz** [comment]
Awesome, thanks @sanjeetsuhag! Good idea to keep this PR on the smaller size and just scoped to just the `Providers`. I'll give the non-test-related code a look.

File: Source/Core/VRTheWorldTerrainProvider.js [hunk 1]
```diff
@@ -86,8 +85,17 @@ function VRTheWorldTerrainProvider(options) {
     if (srs === "EPSG:4326") {
       that._tilingScheme = new GeographicTilingScheme({ ellipsoid: ellipsoid });
     } else {
-      metadataFailure(`SRS ${srs} is not supported.`);
```
ggetz, line 89: I think this path was not previously retried for a reason-- This is not a failed request like the others. This is a case of the result not being supported, which would not change upon retrying.

Source/Scene/GoogleEarthEnterpriseImageryProvider.js
```diff
@@ -229,7 +229,7 @@ function GoogleEarthEnterpriseImageryProvider(options) {
         const e = new RuntimeError(
           `The server ${metadata.url} doesn't have imagery`
         );
-        metadataError = TileProviderError.handleError(
+        metadataError = TileProviderError.reportError(
```
ggetz, line 232: So this was previously not passing the correct paramaters to the old `handleError` correctly, right? There should have been an additional `undefined` for the retryCallback?

Source/Scene/TileMapServiceImageryProvider.js
```diff
@@ -367,6 +356,11 @@ TileMapServiceImageryProvider.prototype._metadataSuccess = function (xml) {
 };
 
 TileMapServiceImageryProvider.prototype._metadataFailure = function (error) {
+  // Only reject if the error is from parsing
+  if (error instanceof RuntimeError) {
```
ggetz, line 360: Is it possible to move the call to `_matadataFailure` to a different part of the promise chain so that this check isn't needed?

Specs/Core/CesiumTerrainProviderSpec.js
```diff
@@ -265,19 +264,21 @@ describe("Core/CesiumTerrainProvider", function () {
...
+          provider.getLevelMaximumGeometricError(1) * 2.0,
+          CesiumMath.EPSILON10
+        );
+        expect(provider.getLevelMaximumGeometricError(1)).toEqualEpsilon(
+          provider.getLevelMaximumGeometricError(2) * 2.0,
+          CesiumMath.EPSILON10
+        );
+      })
+      .catch((e) => {
+        console.log(e);
```
ggetz, line 280: If you're seeing an error here, it may be because the `readyPromise` in the previous spec isn't being waited on, and throwing something during spec cleanup.

ggetz (commented):
Looking good @sanjeetsuhag, I just have a few comments.

Source/Scene/TileMapServiceImageryProvider.js
sanjeetsuhag: I think I can see how this pattern is problematic given that we're having to differentiate for type of error, but semantically, I think this is doing the right thing - where we want to explicitly resolve in the case on an error with `fetchXML`, but not an error within `metadataSuccess`. Given what the requirement is here (i.e. to resolve even if there is a failure), I think this is a case where promise chaining is working against us (where we have to depend on the result of `fetchXML` but any error from it or `metadataSuccess` are both caught in the same block). I saw the `instanceof` pattern used here in https://javascript.info/promise-error-handling, so that's how I initially checked it. I changed it up slightly so that the error type checking is happening in the catch block, where I feel it is better suited than inside the metadataFailure function.

Specs/Core/CesiumTerrainProviderSpec.js
sanjeetsuhag: Yeah, nearly all my testing woes were due to issues like this. Thanks for the tip!

Source/Core/VRTheWorldTerrainProvider.js
sanjeetsuhag: Good catch. Fixed!

Source/Scene/TileMapServiceImageryProvider.js
ggetz: > I saw the instanceof pattern used here in https://javascript.info/promise-error-handling, so that's how I initially checked it. 

I appreciate the citation! I think the major different here is that, in linked example, the error type is very specific to the type of error. In our case, I'm concerned that since `Runtime Error` is very generic, it will be easy for this to become prone to bugs when the code changes in the future-- Any new code that throws a `RuntimeError` will go into this branch rather than being handled by `metadataFailure`. Perhaps we can throw a more specific error, or check the error message.

...Or, this becomes a lot cleaner with async/await syntax, which is now available in our toolkit after removing the eslint es5 restrictions:

````js
async function processMetadata() {
   const result = await this._xmlResource.fetchXML();
   try {
       return this._metadataSuccess(result);
    } catch (e) {
       return this._metadataFailure;
    }
}

````

In the above example, `processMetadata` rejects with any error from `fetchXML` when in happens, and otherwise returns the result of `metadataSuccess` or `metadataFailure` accorsingly.

**Comment by sanjeetsuhag** [comment]
@ggetz This is ready for review.

File: Specs/Core/CesiumTerrainProviderSpec.js [hunk 5]
```diff
@@ -254,8 +253,12 @@ describe("Core/CesiumTerrainProvider", function () {
     const provider = new CesiumTerrainProvider({
       url: "made/up/url",
     });
-    expect(provider.errorEvent).toBeDefined();
-    expect(provider.errorEvent).toBe(provider.errorEvent);
+    return pollToPromise(function () {
+      return provider.ready;
```
ggetz, line 257: No need to change this, but I just wanted to mention that polling for `ready` should be equivalent to awaiting the `readyPromise`.

Specs/Scene/TileMapServiceImageryProviderSpec.js
```diff
@@ -400,11 +402,11 @@ describe("Scene/TileMapServiceImageryProvider", function () {
     const provider = new TileMapServiceImageryProvider({
       url: "http://server.invalid?query=1",
     });
+    provider.readyPromise.catch(() => {});
```
ggetz, line 405: Here and throughout- We should never blindly catch an error in the specs, as it could (and probably is in many cases) covering up an underlying issue. If the test is expected to throw an error

1) There should be an expect statement ensuring the type or message of the error thrown is the intended one
2) There should be a `fail()` stipulation to catch in case the promise resolves instead of rejecting
3) The test name should be specific enough to mention that throwing an error is expected, ie. "throws if format is unsupported" or "rejects on invalid url"

ggetz (commented):
Thanks @sanjeetsuhag! This is down to making sure the tests are in good shape.

Specs/Scene/TileMapServiceImageryProviderSpec.js
sanjeetsuhag: Added these where appropriate.

sanjeetsuhag:
@ggetz The tests have been updated.

Specs/Core/CesiumTerrainProviderSpec.js
```diff
@@ -489,16 +494,17 @@ describe("Core/CesiumTerrainProvider", function () {
...
+      expect(event.message).toContain(
         "does not specify any tile URL templates"
       );
+      provider.errorEvent.removeEventListener(errorMatcher);
+    };
+
+    provider.errorEvent.addEventListener(errorMatcher);
+
+    return provider.readyPromise.then(fail).catch((e) => {
+      expect(e.message).toContain("An error occurred while accessing");
```
ggetz, line 507: This is starting to get nit-picky, but to cover all our bases in these tests, we should have an expectation to confirm that `errorMatcher` was called after the `readyPromise` is rejected.

Specs/Scene/SceneSpec.js
```diff
@@ -1519,16 +1519,22 @@ describe(
...
       });
 
-      expect(scene.terrainProvider).toBe(globe.terrainProvider);
-
-      scene.globe = undefined;
-      expect(function () {
-        scene.terrainProvider = new CesiumTerrainProvider({
-          url: "//newTerrain/tiles",
+      return scene.terrainProvider.readyPromise
+        .catch(() => {
```
ggetz, line 1523: I'm not sure I follow why all these `catch` blocks are needed. What error is being thrown? Is it expected, or can we tweak something about the tests such that the `readyPromises` resolve?

ggetz (commented):
Thanks @sanjeetsuhag! The tests are in a lot better shape. This is very close.

Specs/Scene/SceneSpec.js
sanjeetsuhag: I should have been explicit in the `catch` here too. This is expected to fail since we're providing invalid URLs to the `CesiumTerrainProvider`. I've updated the spec to return valid data by overriding `Resource` load functions (and resetting them after) as is done in the `CesiumTerrainProviderSpec`.

Specs/Core/CesiumTerrainProviderSpec.js
```diff
@@ -404,9 +404,8 @@ describe("Core/CesiumTerrainProvider", function () {
       provider.errorEvent.removeEventListener(errorMatcher);
     };
 
-    provider.errorEvent.addEventListener(errorMatcher);
-
     return provider.readyPromise.then(fail).catch((e) => {
+      provider.errorEvent.addEventListener(errorMatcher);
```
ggetz, line 408: I just realized my wording in the last comment was ambiguous-- Sorry!

Adding the event listener itself can remain where it was before. I meant that in the catch block, we need an `expect` statement to ensure that the listener was actually called. Otherwise this test could return a false positive.

Something like:

````js
const errorMatcher = function (event) {
      ...
      listenerCalled = true;
    };

    provider.errorEvent.addEventListener(errorMatcher);
    
    return provider.readyPromise.then(fail).catch((e) => {
        ...
        expect(listenerCalled).toBeTrue();
    });
    ````

ggetz:
Looks great, thanks @sanjeetsuhag!

PR #10097, opened 2022-02-11
Fix shadows for ModelExperimental models without normals

lilleyse:
ModelExperimental models without normals crash when shadows are enabled. 

image

`ShadowMapShader` checks for the existence of normals by doing a keyword search over the shader text for words like `v_normalEC` or `v_normal`. When it finds a match it applies some additional shadow techniques that use the normal.

This check is pretty fragile and it produces false positives for ModelExperimental. That's because ModelExperimental  shaders always have something like

```
#ifdef HAS_NORMALS
v_normalEC = normal
#endif
```

So the shadow code thinks these models have normals, but some may not, and that leads to a shader compilation error when it tries to use the nonexistent varying.

The fix here is to first check for `#ifdef HAS_NORMALS`, then check for `#define HAS_NORMALS`.

The proper long term fix would be to include shadow code as a model shader stage instead of as a derived command.

Sandcastles for testing:

[Sandcastle 1](http://localhost:8080/Apps/Sandcastle/index.html#c=fVJRb9MwEP4rVp5SKTibeEE0q9BaBA9DTGu1J7+49i0xOHYUn1s6xH/n7KRTgYkocny+77v7fF+UdwHZwcARRnbDHBzZGoKJPX/MZ6UoVI7X3qE0DkZRLJbCqcwLChwQbeLzHL4k42gpJQrOa3q3sh8sbCTKemrwdrMzFkJ9743DtfVRX2wfPt3WmNKA/FvwThRU9lx4TlDx3JAPo+kNmgMELrUuL+7wRyfA8qdwLOl6n5YqBeDk3sIXr8F+/DEAVQK6JwFwjCDcr8UiNz5rGV4UbjupjWu5RAQXJRrvSFBiLf8HhxNsfA93pu2QYqI8SRsgN5mn+Ox9v/PlXKO69OQz5Cr3BlX3IF0L5RW/qtib67S+o/0sdyudVjKghTSRnW9bC7cR0TvyM0nxxyCKaupNn+hUukCpOlDfQS9YHtTZ1QlPUuf0Ms1lWVRFE/BkYZWw6flg+sGP2fiSHEcgxyWSxftINOQqhCQvQZv6ktpoc2BG37zyrzFlZQiUeYrWbs0ziGLV1IT/h2p9ns3XA4xWnhKsu17dTYec86am8HUmem/3cvyr8m8)
[...]

PR #10501, opened 2022-06-30
Fix shadow artifacts in `ModelExperimental`

j9liu:
This PR removes the shadow artifacts that appeared whenever shadows were toggled for `ModelExperimental`.

| `ModelExperimental` | `Model` |
| ---------------------- | --------- |
| image |image |

(The shadows are in a different place because these were taken at different times of day.)

The bug resulted from a fix in #10097. The shader had to be checked for `#define HAS_NORMALS` in order to confirm that normals did exist for the model, and thus had a varying in the fragment shader. However, the actual shader text doesn't contain defines; these are stored in a separate `defines` array under `ShaderSource`. So this was incorrectly assuming that the model didn't provide its own normals, and the shadow shader would use a dummy value `vec3(1.0)` as the normal instead.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/model-experimental-shadows-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
Looks good, thanks @j9liu!

image

PR #10523, opened 2022-07-08
Removes defer

sanjeetsuhag:
Relies on https://github.com/CesiumGS/cesium/pull/10496 getting merged first.

This PR does additional cleanup on removing defer from the not-networking related Core/Scene classes

### DIFF
--- a/Source/Core/Iau2006XysData.js
+++ b/Source/Core/Iau2006XysData.js
@@ -1,6 +1,5 @@
 import buildModuleUrl from "./buildModuleUrl.js";
 import defaultValue from "./defaultValue.js";
-import defer from "./defer.js";
 import defined from "./defined.js";
 import Iau2006XysSample from "./Iau2006XysSample.js";
 import JulianDate from "./JulianDate.js";
@@ -246,10 +245,6 @@ function requestXysChunk(xysData, chunkIndex) {
     return xysData._chunkDownloadsInProgress[chunkIndex];
   }
 
-  const deferred = defer();
-
-  xysData._chunkDownloadsInProgress[chunkIndex] = deferred;
-
   let chunkUrl;
   const xysFileUrlTemplate = xysData._xysFileUrlTemplate;
   if (defined(xysFileUrlTemplate)) {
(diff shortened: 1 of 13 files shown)


Source/Core/Iau2006XysData.js
```diff
@@ -275,9 +270,10 @@ function requestXysChunk(xysData, chunkIndex) {
       samples[startIndex + i] = newSamples[i];
     }
 
-    deferred.resolve();
+    return Promise.resolve();
```
ggetz, line 273: Since you're already in a `then` block, there's no need to explicitly return a resolved promise-- Just the resolved value. Since in this case that's `undefined`, you don't need to return at all.

Source/Widgets/Viewer/Viewer.js
```diff
@@ -2074,8 +2073,12 @@ function zoomToOrFly(that, zoomTarget, options, isFlight) {
   //We can't actually perform the zoom until all visualization is ready and
   //bounding spheres have been computed.  Therefore we create and return
   //a deferred which will be resolved as part of the post-render step in the
-  //frame that actually performs the zoom
-  const zoomPromise = defer();
+  //frame that actually performs the zoom.
+  const zoomPromise = new Promise((resolve) => {
```
ggetz, line 2077: There is [an issue related to error handling for zoomTo](https://github.com/CesiumGS/cesium/issues/10222) that we're tracking. If it's a quick fix here, it would be great to fix it. But let me know if that's actually out-of-scope for this PR.

ggetz (commented):
Thanks @sanjeetsuhag, this is looking good!

Source/Widgets/Viewer/Viewer.js
sanjeetsuhag: Fixed it by catching the error in the `tileset.readyPromise` and calling `cancelZoom` in the `catch` block. Added a spec to make sure that `zoomTo` returns in the case of a failed tileset load - previously, it would never return.

ggetz:
Thanks for the updates @sanjeetsuhag! Please merge in `main` when you get the chance.

sanjeetsuhag:
@ggetz Merged in `main`, but the CI test failure here is due to a change in `main`, not this branch. Seems like it was seen in `main` here: https://github.com/CesiumGS/cesium/commit/921a246fc05e5782d36d2645afbd46964483ab86

ggetz:
Got it. Thanks @sanjeetsuhag!

PR #10530, opened 2022-07-08
Set `ExperimentalFeatures.enableModelExperimental` to true

ptrgags:
in preparation for #10346, we wanted to enable ModelExperimental by default early in the month so there's time to catch any bugs we may have missed.

That said, there are a few known issues (see #10346) so in a few sandcastles we explicitly set the flag to false:

* BIM model that needs to be updated to not use techniques:
  * Apps/Sandcastle/gallery/3D Tiles BIM.html
  * Apps/Sandcastle/gallery/3D Tiles Clipping Planes.html
  * Apps/Sandcastle/gallery/Ambient Occlusion.html
  * Apps/Sandcastle/gallery/MSAA.html
  * Apps/Sandcastle/gallery/Polylines on 3D Tiles.html
* getNode needs to be implemented:
  * Apps/Sandcastle/gallery/CZML Model - Node Transformations.html
  * Apps/Sandcastle/gallery/Time Dynamic Wheels.html
* Point cloud styling is not yet supported
  * Apps/Sandcastle/gallery/3D Tiles Point Cloud Styling.html
  * Apps/Sandcastle/gallery/Montreal Point Cloud.html
* Classification model needs to be updated to use GltfLoader
  * Apps/Sandcastle/gallery/3D Tiles Photogrammetry * Classification.html

All other 3D Tiles sandcastles should work fine.

@j9liu could you review?

PR #10532, opened 2022-07-11
Add `ANGLE_instanced_arrays` to the WebGL stubs for unit testing

ptrgags:
This PR adds stubs for instancing in WebGL 1 so instancing-related unit tests can run in Travis CI.

While working on #10530, @j9liu and I noticed that 30+ tests in `Cesium3DTileset` will break if `ModelExperimental` is used because the implementation of instanced models is GPU only (no `ModelInstanceCollection` which has a CPU fallback). We'd rather not skip those tests completely in Travis, and stubbing the extension is an easy way around it.

For the record, [`ANGLE_instanced_arrays` is available in all the major browsers](https://caniuse.com/mdn-api_angle_instanced_arrays) and is only needed in WebGL 1, as these features are built-in in WebGL 2.

@j9liu could you review?

### DIFF
--- a/Specs/Renderer/VertexArraySpec.js
+++ b/Specs/Renderer/VertexArraySpec.js
@@ -867,7 +867,7 @@ describe(
       }).toThrowDeveloperError();
     });
 
-    it("throws when instanceDivisor is greater than zero instancing is disabled", function () {
+    it("throws when instanceDivisor is greater than zero and instancing is disabled", function () {
       // disable extension
       const instancedArrays = context._instancedArrays;
       context._instancedArrays = undefined;
--- a/Specs/Renderer/VertexArraySpec.js
+++ b/Specs/Renderer/VertexArraySpec.js
@@ -867,35 +867,38 @@ describe(
       }).toThrowDeveloperError();
     });
 
-    it("throws when instanceDivisor is greater than zero and the instanced arrays extension is not supported.", function () {
-      if (!context.instancedArrays) {
-        const buffer = Buffer.createVertexBuffer({
+    it("throws when instanceDivisor is greater than zero instancing is disabled", function () {
+      // disable extension
+      const instancedArrays = context._instancedArrays;
+      context._instancedArrays = undefined;
+
+      const buffer = Buffer.createVertexBuffer({
+        context: context,
+        sizeInBytes: 3,
+        usage: BufferUsage.STATIC_DRAW,
+      });
+
+      const attributes = [
+        {
+          index: 0,
+          vertexBuffer: buffer,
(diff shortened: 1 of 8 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/stub-instancing/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Renderer/VertexArraySpec.js
```diff
@@ -867,35 +867,38 @@ describe(
       }).toThrowDeveloperError();
     });
 
-    it("throws when instanceDivisor is greater than zero and the instanced arrays extension is not supported.", function () {
-      if (!context.instancedArrays) {
-        const buffer = Buffer.createVertexBuffer({
+    it("throws when instanceDivisor is greater than zero instancing is disabled", function () {
```
j9liu, line 870:
```suggestion
    it("throws when instanceDivisor is greater than zero and instancing is disabled", function () {
```

Specs/Scene/ModelExperimental/ModelExperimental3DTileContentSpec.js
```diff
@@ -132,10 +128,6 @@ describe("Scene/ModelExperimental/ModelExperimental3DTileContent", function () {
   });
 
   it("renders i3dm content", function () {
```
j9liu, line 130: I know Travis hasn't passed yet, but something tells me this won't work, since the draw call is a no-op.

j9liu (commented):
@ptrgags ~I'm pretty sure one of the tests will fail because it's a rendering test for i3dm...~ other than that, just one typo

Also, there's a lot of lines in the Travis output that look like this:
```
Unknown mime type for cesium/stub-instancing/Build/Documentation/global.html#ArcType
```

Are these problematic / is there a way to prevent these?

Specs/Scene/ModelExperimental/ModelExperimental3DTileContentSpec.js
j9liu: Never mind? I guess it works

ptrgags: I'll admit I'm not entirely sure how it works, but in `addDefaultMatchers.js` it does account for the webgl stub somehow so the test can still run.

https://github.com/CesiumGS/cesium/blob/9fcc11dfa71d6e64c49aac65202f8a5c296a66f9/Specs/addDefaultMatchers.js#L239-L257

PR #10541, opened 2022-07-12
`loadAttributesAsTypedArray` now creates packed typed arrays

lilleyse:
Previously there were three different ways to load attributes in GltfLoader: as a WebGL buffer, as a typed array counterpart of the WebGL buffer, and as a packed typed array (de-interleaved)

```
  /**
   * A typed array containing tightly-packed attribute values, as they appear
   * in the model file.
   *
   * @type {Uint8Array|Int8Array|Uint16Array|Int16Array|Uint32Array|Int32Array|Float32Array}
   * @private
   */
  this.packedTypedArray = undefined;
```

```
  /**
   * A typed array containing a CPU copy of the vertex buffer for further
   * processing. Since GPU buffers may have padding, attribute values are
   * accessed using byteOffset and byteStride.
   *
   * @type {Uint8Array}
   * @private
   */
  this.typedArray = undefined;
```

The second version ended up not being very useful and was only complicating the code. Now when attributes are loaded as typed arrays they are always packed. This should work fine for voxels and the new ClassificationModel, which are the only two systems that use the `loadAttributesAsTypedArray` option.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/load-as-typed-array-packed/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/Scene/GltfLoader.js
```diff
@@ -698,24 +698,20 @@ function loadAccessor(loader, gltf, accessorId, useQuaternion) {
...
+        accessor,
+        bufferViewTypedArray
+      );
 
-        useQuaternion = defaultValue(useQuaternion, false);
-        loadAccessorValues(accessor, packedTypedArray, values, useQuaternion);
-      })
-      .catch(function () {
-        loadDefaultAccessorValues(accessor, values);
-      });
```
lilleyse, line 718: It seemed unsafe to catch and ignore errors from the buffer view loader so I removed it. The unit tests still pass so maybe this wasn't actually needed?

Source/Scene/GltfLoader.js
```diff
@@ -883,6 +858,33 @@ function loadAttribute(
...
+      if (loadBuffer) {
+        attribute.buffer = vertexBufferLoader.buffer;
+      }
+
+      if (loadTypedArray) {
+        attribute.typedArray = ComponentDatatype.createArrayBufferView(
+          vertexBufferLoader.quantization.componentDatatype,
+          vertexBufferLoader.typedArray.buffer
+        );
+      }
```
lilleyse, line 871: This refactor exposed a bug related to packed typed arrays of Draco content. Basically `getPackedTypedArray` is not quantization aware and it fails with an out of bounds error.

Here Draco uses its own method to create the typed array.

Source/Scene/GltfLoader.js
```diff
@@ -979,8 +986,23 @@ function loadInstancedAttribute(
...
+  // - translations are required for 2D
+  // - GPU instancing is not supported.
+  let loadTypedArray =
+    loader._loadAttributesAsTypedArray ||
+    ((hasRotation || !hasTranslationMinMax) && isTransformAttribute) ||
+    modelSemantic === InstanceAttributeSemantic.FEATURE_ID ||
+    !frameState.context.instancedArrays;
+
+  const loadBuffer = !loadTypedArray;
+  loadTypedArray = loadTypedArray || loadFor2D;
```
lilleyse, line 1005: I tried to put all the load-as-typed-array logic in one place. Previously it was spread across this function and the calling code.

Source/Scene/GltfVertexBufferLoader.js
```diff
@@ -384,7 +384,11 @@ function loadFromDraco(vertexBufferLoader) {
     }
 
     // Now wait for process() to run to finish loading
-    vertexBufferLoader._typedArray = typedArray;
+    vertexBufferLoader._typedArray = new Uint8Array(
+      typedArray.buffer,
+      typedArray.byteOffset,
+      typedArray.byteLength
+    );
```
lilleyse, line 391: Before this fix `_typedArray` was not actually a `Uint8Array`.

This and https://github.com/CesiumGS/cesium/pull/10541/files#r919520142 were causing Draco attributes to fail when packed.

Source/Scene/GltfLoader.js
ptrgags, line 871: Could we factor out a function to contain this logic? this section of code is getting rather nested.

Source/Scene/GltfLoader.js
```diff
@@ -883,6 +858,33 @@ function loadAttribute(
...
+        // The accessor's byteOffset and byteStride should be ignored since values
+        // are tightly packed in a typed array
+        const bufferViewTypedArray = vertexBufferLoader.typedArray;
+        attribute.typedArray = getPackedTypedArray(
+          gltf,
+          accessor,
+          bufferViewTypedArray
+        );
+        attribute.byteOffset = 0;
+        attribute.byteStride = undefined;
```
ptrgags, line 886: similarly here, factor out a function

ptrgags (commented):
@lilleyse changes are working nicely! I only had a small suggestion about refactoring a function to limit nested ifs

lilleyse:
@ptrgags updated

ptrgags:
This looks good now, thanks @lilleyse!

PR #10545, opened 2022-07-13
Removes defer usage from TaskProcessor

sanjeetsuhag:
This PR removes usage of `defer.js` from TaskProcessor (and related classes and specs).

### DIFF
--- a/Source/Core/TaskProcessor.js
+++ b/Source/Core/TaskProcessor.js
@@ -62,7 +62,7 @@ function canTransferArrayBuffer() {
 
 const taskCompletedEvent = new Event();
 
-function completeTask(processor, data) {
+function completeTask(processor, data, resolve, reject) {
   --processor._activeTasks;
 
   const id = data.id;
@@ -70,9 +70,6 @@ function completeTask(processor, data) {
     // This is not one of ours.
     return;
   }
-
-  const deferreds = processor._deferreds;
-  const deferred = deferreds[id];
 
   if (defined(data.error)) {
     let error = data.error;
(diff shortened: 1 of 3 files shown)


Source/Core/TaskProcessor.js
```diff
@@ -11,58 +10,9 @@ import isCrossOriginUrl from "./isCrossOriginUrl.js";
...
-      return TaskProcessor._canTransferArrayBuffer;
-    }
-
-    const deferred = defer();
-
-    worker.onmessage = function (event) {
-      const array = event.data.array;
-
-      // some versions of Firefox silently fail to transfer typed arrays.
-      // https://bugzilla.mozilla.org/show_bug.cgi?id=841904
```
sanjeetsuhag, line 47: This bug has been marked as resolved.

Source/Core/TaskProcessor.js
```diff
@@ -300,37 +240,27 @@ TaskProcessor.prototype.initWebAssemblyModule = function (webAssemblyOptions) {
...
-      }
+    let transferableObjects;
+    const binary = wasmConfig.wasmBinary;
+    if (defined(binary)) {
+      transferableObjects = [binary];
+    }
 
+    const compilePromise = new Promise((resolve) => {
       worker.onmessage = function (event) {
-        worker.onmessage = function (event) {
```
sanjeetsuhag, line 319: Since `scheduleTask` will set the `onmessage`, I don't think this is needed here.

sanjeetsuhag:
Given that this PR doesn't actually address the defer pattern, we're closing this PR. We plan on fixing this when we revisit `Resource.js`.

PR #10548, opened 2022-07-13
Include the original stack trace in the `ModelExperimental` error handler

ptrgags:
When an uncaught exception happens in `ModelExperimental`, the stack trace was getting swallowed since the error is turned into a `new RuntimeError()`. Now, I update the runtime error's `stack` property to include the original stack since this is more helpful for debugging.

This will help make investigating some `ModelExperimental` bugs easier to see when running in Travis.

To test, I did the following:

1. in `GltfLoader.js`, inside the block `textureProcessPromise = new Promise(...)` (but outside of `that._process()`!!) throw any exception.
2. Load a `ModelExperimental` and check the console for the uncaught exception 
3. Or run a single `ModelExperimental` test in the `SpecRunner`. It should time out, and you'll see the exception

@j9liu could you review?

### DIFF
--- a/Source/Scene/ModelExperimental/ModelExperimentalUtility.js
+++ b/Source/Scene/ModelExperimental/ModelExperimentalUtility.js
@@ -33,7 +33,15 @@ ModelExperimentalUtility.getFailedLoadFunction = function (model, type, path) {
     if (defined(error)) {
       message += `\n${error.message}`;
     }
-    return Promise.reject(new RuntimeError(message));
+
+    const runtimeError = new RuntimeError(message);
+    if (defined(error)) {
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/better-error-handler/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu:
@ptrgags I was actually able to confirm that this works without needing to simulate an exception -- one occurred locally for me of its own accord 😅 
image

Since the Travis (and local) failure occurred before this PR, I'm going to merge this. Thanks!

Issue #10533, opened 2022-07-11
Alpha mask + translucent feature not working 


lilleyse:
This is a `ModelExperimental` bug

Background: When a feature in a primitive become translucent we generate an opaque command and a translucent command. `CPUStylingPipelineStage` is responsible for rendering only opaque features in the opaque command and only translucent features in the translucent command.

The problem is that `renderResources.alphaOptions.alphaMode` is set to `AlphaMode.BLEND` and `#define ALPHA_MODE_BLEND` is set for both commands.

In `ModelExperimentalFS.glsl` this means the opaque command actually goes down the `ALPHA_MODE_BLEND` code path and doesn't discard alpha-mask pixels. Which leads to the artifacts in the gif below.

```glsl
vec4 handleAlpha(vec3 color, float alpha)
{
    #ifdef ALPHA_MODE_MASK
    if (alpha < u_alphaCutoff) {
        discard;
    }
    return vec4(color, 1.0);
    #elif defined(ALPHA_MODE_BLEND)
    return vec4(color, alpha);
    #else // OPAQUE
    return vec4(color, 1.0);
    #endif
}
```

![Peek 2022-07-11 14-47](https://user-images.githubusercontent.com/915398/178336250-47a44d63-0fca-43db-8bea-5513c617e11e.gif)

[...]

PR #10549, opened 2022-07-13
Fix alpha mask + translucent features

lilleyse:
Fixes https://github.com/CesiumGS/cesium/issues/10533


Source/Scene/ModelExperimental/AlphaPipelineStage.js
```diff
@@ -23,16 +23,15 @@ AlphaPipelineStage.process = function (renderResources, primitive, frameState) {
 
   const renderStateOptions = renderResources.renderStateOptions;
   if (alphaOptions.pass === Pass.TRANSLUCENT) {
+    renderStateOptions.cull.enabled = false;
+    renderStateOptions.depthMask = false;
     renderStateOptions.blending = BlendingState.ALPHA_BLEND;
-  } else {
-    renderStateOptions.blending = BlendingState.DISABLED;
   }
```
lilleyse, line 29: I removed the code that unsets the blending state because `BlendingState.DISABLED` is the default.

Source/Scene/ModelExperimental/MaterialPipelineStage.js
```diff
@@ -104,19 +104,13 @@ MaterialPipelineStage.process = function (
   // Configure back-face culling
   const model = renderResources.model;
   const cull = model.backFaceCulling && !material.doubleSided;
-  const translucent = defined(model.color) && model.color.alpha < 1.0;
-  renderResources.renderStateOptions.cull = {
-    enabled: cull && !translucent,
-  };
+  renderResources.renderStateOptions.cull.enabled = cull;
```
lilleyse, line 107: The `translucent` check seems unsafe because there are multiple ways a models could become translucent. Better to set `cull` to true in the `AlphaPipelineStage` after all the stages have been true.

Source/Scene/ModelExperimental/ModelColorPipelineStage.js
```diff
@@ -52,21 +51,9 @@ ModelColorPipelineStage.process = function (
...
-    // When the model is invisible, only disable the color and depth writes
-    // if the model does not have a silhouette. Otherwise, the silhouette
-    // will not render.
-    renderStateOptions.colorMask = {
-      red: false,
-      green: false,
-      blue: false,
-      alpha: false,
-    };
-    renderStateOptions.depthMask = false;
```
lilleyse, line 66: I removed this code because setting `depthMask` to false already happens in the `AlphaPipelineStage`, and setting `colorMask` to false is more like an optimization that's not strictly necessary.

Source/Scene/ModelExperimental/ModelExperimentalDrawCommand.js
```diff
@@ -639,7 +639,6 @@ function deriveTranslucentCommand(command) {
   derivedCommand.pass = Pass.TRANSLUCENT;
   const rs = clone(command.renderState, true);
   rs.cull.enabled = false;
-  rs.depthTest.enabled = true;
```
lilleyse, line 642: Since `depthTest` is true by default I removed it here.

Source/Scene/ModelExperimental/ModelRenderResources.js
```diff
@@ -67,5 +71,14 @@ export default function ModelRenderResources(model) {
...
+  this.renderStateOptions = clone(
+    RenderState.fromCache({
+      depthTest: {
+        enabled: true,
+        func: DepthFunction.LESS_OR_EQUAL,
+      },
+      blending: BlendingState.DISABLED,
+    }),
+    true
+  );
```
lilleyse, line 83: Moved the default render state from `PrimitiveRenderResources` to `ModelRenderResources` so that it starts at the top and propagates down.

j9liu (commented):
@lilleyse just one clarifying question  + some small suggestions.

I tested some sandcastles and noticed this difference in the "Custom Shaders Models" sandcastle:

| Before | After |
| ------- | ------|
| image | image |

It looks like the model still had backface-culling enabled even though the model was transparent, so this seems like an improvement.

However, I found that "3D Tiles Next Photogrammetry Classification" is buggy now: toggling the "Translucent Windows" setting results in this:

transparency bug

The windows themselves are translucent, but it looks like a lot of the model doesn't want to render? I wonder if it's a level-of-detail bug or something, because the model loads when you zoom very close to it, then disappears again when you zoom out. Or maybe it has to do with backface culling?

lilleyse:
As for the Ferry Building... I think the problem is that translucent commands now set `depthMask: false` which disables the z-buffer and causes whatever triangles were rendered last to show up on top, and it's very noticeable because most of the surface is actually opaque and OIT is disabled.

One workaround is to use a style instead, since styles will generate separate opaque and translucent commands. I wonder if we need a similar concept for custom shaders...

Source/Shaders/ModelExperimental/ModelExperimentalFS.glsl
```diff
@@ -23,12 +23,9 @@ vec4 handleAlpha(vec3 color, float alpha)
...
         discard;
     }
-    return vec4(color, 1.0);
-    #elif defined(ALPHA_MODE_BLEND)
-    return vec4(color, alpha);
-    #else // OPAQUE
-    return vec4(color, 1.0);
     #endif
+
+    return vec4(color, alpha);
```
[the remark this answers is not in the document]
lilleyse: More like, blending is only enabled if the pass is `TRANSLUCENT`. So if for whatever reason the alpha is < 1.0 in the opaque pass it will get written to the framebuffer that way but it won't blend.

j9liu:
@ptrgags can you review?

Specs/Scene/ModelExperimental/MaterialPipelineStageSpec.js
```diff
@@ -142,7 +142,7 @@ describe(
           uniformMap: uniformMap,
           lightingOptions: new ModelLightingOptions(),
           alphaOptions: new ModelAlphaOptions(),
-          renderStateOptions: {},
+          renderStateOptions: RenderState.getState(RenderState.fromCache()),
```
ptrgags, line 145: Could we add a `mockRenderResources()` function in this file?

Specs/Scene/ModelExperimental/ModelColorPipelineStageSpec.js
```diff
@@ -103,7 +102,7 @@ describe("Scene/ModelExperimental/ModelColorPipelineStage", function () {
       uniformMap: {},
       model: mockModel,
       alphaOptions: defaultAlphaOptions,
-      renderStateOptions: {},
+      renderStateOptions: RenderState.getState(RenderState.fromCache()),
```
ptrgags, line 105: Similarly here, could we create a `mockRenderResources()` function for this file?

ptrgags (commented):
@j9liu Code changes look good to me, and I can confirm that the Sandcastles are working! Just a couple of comments about refactoring the test code.

image

image

j9liu:
@ptrgags updated. I noticed that the `MaterialPipelineStage` was adding the `USE_WIREFRAME` define to the shader, which was leftover from when `WireframePipelineStage` didn't exist. I moved it to that stage, let me know if the [wireframe sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=tVdtb9s2EP4rhPehzmDLsuMkrpcVy1uzDg0a1GkLbBoGSjpbhClSICnHWZH/viMpyXLsbl2L5JUvd7zn7h7e0YkU2pAVg3tQ5Gci4J5cgGZlHnx0a92ok7j5hRSGMgEq6vTI50gQojNZ8vRMsJwamBKjSuhF4vHgp0hEwp8YJFwmyyAplQJh7lgOaKM6/7eSMyouUTeYK5m/0XJyHA679uSoMwpHw/5w2A9f3o3C6ehkengShMPj8Hg8Pjk6HE+OJ+OXhye/R51IeHuDAbmRKXAdicS5lLsZmvvDnugAE2JgbaZ4vIdAzinnUgr0yG+jInBIUGReisQwKUj3oNZFh92mM9Ot1yzYIBjgz4zmBQf0hw687YG3UhnZngULHjdm7dfwqB5bf+z/R7fr/z6BP1syISAlFxlVNDEuJ8/owQ2t0f+Fwx3o4f9Afol4JbmQeaFAa3TBYXku+M7axljtDuPLO1UiLxfczL/ZlQ+CM0PO5fqb0H8RMx7oTm4GNUoSflV8pYDnZYMz0R5/H5evlSxFSj5CxhL+bNC9lcrI9uy7+PwGSw0VCRL5W4nwFeDx6MbO1mSXwP8W+j9doUyagu8AYIF0uFKIy8UnpmCuaA5XgsYcUsRPuXZVvaWbFU86xa9AUyYWt8wk2XvJvVPV3g01WWDke5SgQneHk/DAAQr9X1++/blztob0tTV/p1B2LlW+6RbNkg6wpVDu5eTrRucasDdRI1XVQYRUJqtiE3XuQZtNs+BQ9QecNRlqZ6VUvEcyYIvMVHnzEAupmRNucF1QZXBExaHrYpewUAC6Smt/ODoMwpPx+Hj4skrSeByER+HhSXhcLXgrduyxkaoZBzpBj4JCsRxNrkAHCnK5gjOMr89tXuXvS/I0TSscdS6swtW6AJTBdky5g3yNHOo23LSMmhLrf73izGAaFVtP92Qje5L8dk5a7K4j1+Iq8qg1q06+4pwVWrI0+HQ9m4xbAnvo0XB9A5bhc6TMb1GWz9jf+CwZjiabXbq2uzPkD+6MQvxq9sBR/nLrGjSPGi+SPtlsLlGw9/ZU9++gnVsXTMwkTR9uMfZMQ2AyEN1NnXASTbGon1F4qqLIfLk8M01Ud2L6pTtJxaKdi/1Xc3x00Psvmf5wS8h7E9tiirZmRQYKAoWypSY/kk0p8oM6BrUevlyQp/79iF44wlp2N1xEd4vpFnkb4be4Fby/ur06u+vVcXaXwv17tIZmVKQJ1QYrPB58JyWPqboBUfoQ6+bdiHRKlgSlSYKJMUAwIyRmzidNkGduwWbCA6/rlfFH4g1MZVLaGxUswFxxsMPzhzd4+zqVTNSx1ipPlgLfxLI0gbGGuw2L9onQouAP5xWWjWivNr5Hx0bBAnkXa1Ary8S2XtTZz9WOS1Kgy1gnisXQYuSK8hIaRuZ7CG+LkBXaZMB+d3qdU20eOLyqE/oLywssy7a8dLHTGcBOhwHXgxhfY2CCROu6bRHyQx3fzxvCxRiwhWveU6IWMe2ORz1S/4bBZKONtwOzjlGbknGxbi3HUqWg+p6l25uPO6aZKErTBrACZRiWjz7lbCGmWG/SlMOu1b6x1B1tWa63YmmMzLd3d027wgpb3s/xI1j/3vWLKfrB06fqp4N2vE9TtiIs/XnPJziScKo17sxL7upk1Hl1OkD5HVUuXR15h45z+mDFsuGrt34xCILTAU73azbUb9J/6sNpHgqwoDJIlrF9NXnyWDT2tYHzFB8/fXsDazH7FNnL257X/VCk7jPoC2fhRcvmLJP35L7WqsO0wfsP) doesn't work for you.

ptrgags:
Updates look good, thanks @lilleyse and @j9liu!

Issue #10551, opened 2022-07-13
"renders implicit tileset with multiple contents" fails for `ModelExperimental`


j9liu:
In `Cesium3DTileset`, the "renders implicit tileset with multiple contents" test and its legacy counterpart fail when using `ModelExperimental`. Tis is because the tileset is considered loaded / ready even though all of the contents have not finished loading -- that is, some of the models have `ready = false` when the unit test executes.

ptrgags:
#10540 disabled all the tests that use those same tilesets while we look into it.

j9liu:
#10557 -- same issue with tilesets using implicit multiple contents, with metadata

ptrgags:
Update: @j9liu, @lilleyse and I have been looking into this a bit this morning and at least found a few leads:

* The failure seems easier to reproduce in Firefox than in Chrome.
* when the test fails, at least one of the `Multiple3DTileContents` is unable to immediately schedule requests due to throttling by `RequestScheduler` (limit is currently 6, and there are 9 contents in the tileset). 
* in the unit tests, `waitForTilesLoaded()` waits until `tileset.tilesLoaded` is true. This happens at least once between when the implicit tileset transcodes the tiles into `Cesium3DTile` objects and when the tileset load promise for the test resolves. The tileset uses the statistics object to determine when the tiles are loaded; it doesn't directly wait on any ready promise.
* When the tileset load promise in the test resolves, not all of the contents are ready. This might have to do with the timing of transcoding implicit tiles) and waiting for those tiles' contents to be ready.
* @lilleyse and I have a hunch that it has to do with the removal of `defer()` promises in #10361. I'm currently checking if I can reproduce the problem before vs after that branch was merged. I'll post the results of this testing soon.

The TL;DR is the tileset is incorrectly detecting when the tiles are loaded, and it might have to do with how `Multiple3DTileContent` manages the promises when requests are throttled.

ptrgags:
Results from testing the `defer()` changes. I enabled ModelExperimental for `Cesium3DTileset` specs and put a breakpoint where the failure would happen. (all of this in Firefox)

Commit before the merge:
- I ran the tests 10 times (given that this error is sometimes hard to reproduce). I was not able to reproduce the error
Commit after the merge:
- On the first try, I was able to reproduce the error

So that seems to confirm that the removal of `defer()` was problematic. 

[...]

PR #10552, opened 2022-07-13
Disable `ModelExperimental` for failing `Cesium3DTileset` tests

j9liu:
#10530 introduced a test failure in Travis that will be hard to properly fix. I opened an issue so we can come back to it later #10551. Until then, we'll disable `ModelExperimental` for those unit tests.

### DIFF
--- a/Specs/Scene/Cesium3DTilesetSpec.js
+++ b/Specs/Scene/Cesium3DTilesetSpec.js
@@ -5842,7 +5842,11 @@ describe(
       it("renders implicit tileset with multiple contents", function () {
         return Cesium3DTilesTester.loadTileset(
           scene,
-          implicitMultipleContentsUrl
+          implicitMultipleContentsUrl,
+          {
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/travis-failure-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/Cesium3DTilesetSpec.js
```diff
@@ -6168,7 +6173,11 @@ describe(
       it("renders implicit tileset with multiple contents (legacy)", function () {
```
ptrgags, line 6173: looks like the test
`renders implicit tileset with multiple contents (legacy with 'content')` also has the same issue, let's disable that one too.

ptrgags:
Looks good, thanks @j9liu!

PR #10556, opened 2022-07-15
Fix flaky zoom test

ggetz:
This test has been failing intermittently in CI.

I think the root of the flakiness is that the viewer's `zoomTo` function starts handling the zoom target's `readyPromise` at the end of the frame. If the `readyPromise` fails too quickly, such as with an invalid URI, the `catch` block has not yet been set up.

The `zoomTo` function would still behave as expected, but the yet-to-be-handled promise rejection fails the test. The user should see no functional difference in the browser.

As a solution, we can await the failure, assert its the expected one, then continue with the `zoomTo`.

### DIFF
--- a/Specs/Widgets/Viewer/ViewerSpec.js
+++ b/Specs/Widgets/Viewer/ViewerSpec.js
@@ -1207,7 +1207,7 @@ describe(
 
       return tileset.readyPromise
         .catch(function (e) {
-          expect(e.message).toEqual("Request has failed. Status Code: 404");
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-zoom-test/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #10570, opened 2022-07-19
Fix multiple contents race condition

ptrgags:
Fixes https://github.com/CesiumGS/cesium/issues/10551

In multiple contents, request throttling was creating a corner case where the contents were not yet created, but the `readyPromise` looped over an empty list of contents and resolved too soon.

The fix was to make the readyPromise return `undefined` and have `Cesium3DTile` content short-circuit the promise chain. The tile automatically re-requests the content as long as the tile is still in the `LOADING` state.

I also fixed a couple details about rendering statistics for Multiple contents/implicit tiling along the way:

* `tileset.statistics.numberOfTilesLoaded` wasn't being updated for implicit tiling like `Cesium3DTileset.loadTileset()` does
* when a request was canceled, multiple contents wasn't incrementing the number of attempted requests.

(note that for multiple contents, statistics are a bit ill-defined. if the property has "requests" in the name, it's counting the number of requests (one per inner content) but if the property has "tile" in the name, it's counting the number of tiles (1 regardless of how many inner contents there are)

@j9liu could you review when you get a chance?

### DIFF
--- a/Source/Scene/Cesium3DTile.js
+++ b/Source/Scene/Cesium3DTile.js
@@ -1105,8 +1105,12 @@ function requestMultipleContents(tile) {
   tile._contentState = Cesium3DTileContentState.LOADING;
   const contentReadyToProcessPromise = multipleContents.contentsFetchedPromise.then(
     function () {
-      if (tile._contentState !== Cesium3DTileContentState.LOADING) {
-        // tile was canceled, short circuit.
+      if (
+        tile._contentState !== Cesium3DTileContentState.LOADING ||
+        !defined(multipleContents.readyPromise)
+      ) {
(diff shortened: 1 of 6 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-multi-content-promise/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu (commented):
@ptrgags looks good to me, just one suggestion

ptrgags:
@j9liu updated!

j9liu:
@ptrgags merging now, thanks!

Issue #10346, opened 2022-05-03
`ModelExperimental` -> `Model` Roadmap


j9liu:
One of the major items on the 3D Tiles Next Roadmap (#9520) is bringing `ModelExperimental` up to feature parity with `Model`, so that it can entirely replace it. In other words, we aim to remove the current `Model` class and any related files, and rename `ModelExperimental` to `Model` without any hiccups.

This issue goes into more detail than the overall roadmap and will be updated as necessary.

**Remaining `ModelExperimental` Features**
- [x] Basic 2D / CV rendering: #9934
- [x] Clamp to ground #10448 
- [x] Support `AGI_articulations` #10479 
- [x] Support silhouettes #10457 
- [x] Support `distanceDisplayCondition` #10481 
- [x] Add `options.id` for picking #10491 
- [x] Add `options.credit` #10489 
- [x] Add `options.asynchronous` #10490 
- [x] Rendering statistics #9886
- [x] Support outline extension for OSM buildings: #10338
- [x] Integration with entities and CZML systems: #10285
- [ ] Use `GltfLoader` in `ClassificationModel`

**Important bugfixes**
- [ ] Normal mapping rendering bug: #10296
- [x] Custom shaders and instancing bug: #10113
- [ ] 3D Tiles BIM model crashes: [comment](https://github.com/CesiumGS/cesium/issues/10379#issuecomment-1127913756) in #10379 
- [x] OSM buildings render differently #10338 

**Lower-priority bugfixes**
- [ ] Quantized animation bug: #10321 
- [ ] Morph targets unit test failure:  #10354 
- [ ] Artifacts in 2D / CV when rendering over IDL: #10401 (fix in the distant future)
- [ ] Some tilesets not showing up with `projectTo2D` in 2D / CV #10402 (more related to bounding volumes in 2D than `ModelExperimental` itself)

**Testing**
- [x] Ensure all models from [glTF-Sample-Models](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0) load correctly with `ModelExperimental`
- [ ] Ensure that tilesets load correctly with `ExperimentalFeatures.enableModelExperimental` set to true
- [x] Ensure that entities and CZML work with `ModelExperimental`
- [ ] Go through the "category - gltf" label for any important lingering issues
- [ ] Go through Sandcastles and make sure they work with ModelExperimental. See https://github.com/CesiumGS/cesium/issues/10379

**Files to remove**
- [ ] `Model`
[...]

PR #10594, opened 2022-07-25
Remove glTF 1.0 techniques from Photogrammetry Classification tileset

ptrgags:
The [3D Tiles Photogrammetry Classification](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Photogrammetry%20Classification.html) used a b3dm asset that uses glTF 1.0-style techniques (even though it was a glTF 2.0 model otherwise). 

I manually converted the glTF asset to avoid using techniques (since those are deprecated). The Sandcastle linked above should work correctly (using Model). 

Note that if you set `enableModelExperimentalFeatures = true`, the classification will not work because `ClassificationModel` does not yet work for ModelExperimental (it's an item on the roadmap: https://github.com/CesiumGS/cesium/issues/10346), but it doesn't crash.

@j9liu could you review?


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/classification-no-techniques/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

j9liu:
@ptrgags looks good -- I confirmed that `Model` behaves just the same, and it doesn't cause a crash anymore for `ModelExperimental`. Thanks!

Issue #6506, opened 2018-04-27
Generate normals automatically for glTF 2.0 models


lilleyse:
From the [glTF spec](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0)

> Implementation note: When normals are not specified, client implementations should calculate flat normals.

This would replace the stopgap solution in https://github.com/AnalyticalGraphicsInc/cesium/pull/6501.

emackey:
Note: Make sure to *not* calculate normals if the `unlit` extension is in use (#6977)

lilleyse:
Related issue: https://github.com/CesiumGS/cesium/issues/10465

ptrgags:
The implementation for this will be quite similar to what I did for outlines in #10506. This can be done by:

* Create a class (perhaps `FlatNormalGenerator`) that takes typed arrays for indices/attributes and generates the flat normal attribute. Vertices will need to be duplicated in most cases since the normal is different for each adjacent face. See `PrimitiveOutlineGenerator` for an example of how I did the indices/attribute processing
* In `PrimitiveLoadPlan`:
  * Add a `needsFlatNormals` flag (similar to the `needsOutlines` flag)
  * update `postProcess()` to generate flat normals (using `FlatNormalGenerator`) if needed.
  * I think normals should be processed first because it's more likely to duplicate vertices. The more the vertices are duplicated, the less work the outline stage needs to do to generate outline coordinates.
* Update `GltfLoader` to check for primitives without normals. if so, update details in the `loadPlan` so the attributes load as `attribute.packedTypedArray` and then are post-processed.

j9liu:
This came up on the forum: https://community.cesium.com/t/lightning-or-normals-problem-with-3d-tiles/19985

lilleyse:
Came up again here: https://community.cesium.com/t/how-to-create-detail-texture/24165/4

javagl:
Came up again in https://community.cesium.com/t/3d-model-does-not-display-shades/30269/11

PR #10600, opened 2022-07-26
Ensure spec parity between `Model` and `ModelExperimental`

j9liu:
This PR reorganizes `ModelExperimentalSpec`, and adds a handful of new specs to `ModelExperimental` and related classes. These are specs from `ModelSpec` that didn't have existing `ModelExperimental` counterparts.

There are two `xit`-ed unit tests. We need to return to these later and fix them:
-  "disables culling"
- "renders model with emissive texture"

For comparison, here's how the latter test looks:
| `Model` | `ModelExperimental` |
| -------- | ----------------------- |
| image | image |

### DIFF
--- a/Source/Scene/ModelExperimental/ModelExperimental.js
+++ b/Source/Scene/ModelExperimental/ModelExperimental.js
@@ -373,6 +373,8 @@ export default function ModelExperimental(options) {
 
   this._sceneGraph = undefined;
   this._nodesByName = {}; // Stores the nodes by their names in the glTF.
+
+  this._ignoreCommands = defaultValue(options.ignoreCommands, false);
 }
 
 function createModelFeatureTables(model, structuralMetadata) {
@@ -2012,7 +2014,7 @@ function submitDrawCommands(model, frameState) {
     displayConditionPassed &&
     (!invisible || silhouette);
 
-  if (showModel) {
+  if (showModel && !model._ignoreCommands) {
     addCreditsToCreditDisplay(model, frameState);
     const drawCommands = model._sceneGraph.getDrawCommands(frameState);
     frameState.commandList.push.apply(frameState.commandList, drawCommands);
--- a/Source/Scene/ModelExperimental/ModelExperimental.js
+++ b/Source/Scene/ModelExperimental/ModelExperimental.js
@@ -66,7 +66,7 @@ import SplitDirection from "../SplitDirection.js";
  * @param {HeightReference} [options.heightReference=HeightReference.NONE] Determines how the model is drawn relative to terrain.
  * @param {Scene} [options.scene] Must be passed in for models that use the height reference property.
  * @param {DistanceDisplayCondition} [options.distanceDisplayCondition] The condition specifying at what distance from the camera that this model will be displayed.
- * @param {Color} [options.color] A color that blends with the model's rendered color.
+ * @param {Color} [options.color=Color.WHITE] A color that blends with the model's rendered color.
  * @param {ColorBlendMode} [options.colorBlendMode=ColorBlendMode.HIGHLIGHT] Defines how the color blends with the model.
  * @param {Number} [options.colorBlendAmount=0.5] Value used to determine the color strength when the <code>colorBlendMode</code> is <code>MIX</code>. A value of 0.0 results in the model's rendered color while a value of 1.0 results in a solid color, with any value in-between resulting in a mix of the two.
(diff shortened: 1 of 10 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/ModelExperimental/ModelExperimentalSpec.js
```diff
@@ -485,300 +407,378 @@ describe(
...
+            expect(rgba[3]).toEqual(result[3]);
+          });
         });
       });
     });
 
-    it("displays all types of credits", function () {
-      const resource = Resource.createIfNeeded(boxWithCreditsUrl);
+    // This test works for Model but does not work for ModelExperimental.
+    xit("renders model with emissive texture", function () {
```
ptrgags, line 612: I see what's happening here, this box has no normals but has an emissive texture.

Without normals, our implementation sets the lighting to `UNLIT` mode (which doesn't support emissive), but in this particular case, it really isn't an unlit, it's a PBR material without 

Not sure what to do about this in the short term, ideally we'd have #6506 so it would have normals and then it would go down the PBR path correctly...

ptrgags (commented):
@j9liu specs are running fine though when I compare the spec lists looks like there's quite a few tests the old `Model` had that aren't in `ModelExperimental`. Maybe that was intentional (e.g. some render tests might be better handled in other files) but wasn't sure.

j9liu:
Two things to note from this PR:
- `GltfJsonLoader` automatically upgrades glTF 1.0 to glTF 2.0. We should remove that officially when we deprecate glTF 1.0 models.
- Trying to include a check for the `glTF` magic broke a lot of unit tests, so I'm not doing anything with that for now.

j9liu:
@ptrgags updated!

Specs/Scene/ModelExperimental/ModelExperimentalSpec.js
```diff
@@ -1134,1164 +995,2219 @@ describe(
...
+            result = rgba;
+          });
+
+          const ibl = model.imageBasedLighting;
+          ibl.sphericalHarmonicCoefficients = undefined;
+          expect(renderOptions).toRenderAndCall(function (rgba) {
+            expect(rgba).not.toEqual(result);
+          });
+
+          scene.highDynamicRange = false;
```
ptrgags, line 3037: This should go in an `afterEach` (or alternatively, set it to the default in `beforeEach`), this way if this test throws an exception it doesn't impact other tests that use the same scene

Source/Scene/ModelExperimental/ModelExperimental.js
```diff
@@ -195,8 +195,8 @@ export default function ModelExperimental(options) {
   this._id = options.id;
   this._idDirty = false;
 
-  const color = options.color;
-  this._color = defined(color) ? Color.clone(color) : undefined;
+  const color = defaultValue(options.color, Color.WHITE);
```
ptrgags, line 198: I'm guessing this is why the `3D Tiles Next Photogrammetry` example looks a bit washed-out now?

image

Is this what it's supposed to look like, and if so, I wonder if we should tweak the settings in the Sandcastle to look a bit more vibrant

ptrgags (commented):
@j9liu almost there, just a couple comments!

Source/Scene/ModelExperimental/ModelExperimental.js
lilleyse: Does `color` need to default to white? That will cause blending to happen when the mode is mix or replace. And the `ModelColorPipelineStage` will always be added.

It's ok if `Model` and `ModelExperimental` deviate a little bit here. I assume `Model` treats white as a special case where it disables blending, but it's a little nicer if `color = undefined` means no blending.

j9liu: Ah, I just uploaded a fix that disables the `ModelColorStage` if CPU styling is being used.

The bug was happening because `Model`'s default color was `Color.WHITE`, and the shader enabled the `MIX` blend mode so the styling colors would look more blended with the model. But this was making it also blend the color for `ModelExperimental`, so it looked paler than usual.

Should I revert to a default `undefined` color for `ModelExperimental`?

lilleyse: > Should I revert to a default `undefined` color for `ModelExperimental`?

Yeah, I think that's the best thing to do

j9liu:
@ptrgags updated!

ptrgags:
Updates look good, thanks @j9liu!

Issue #10517, opened 2022-07-06
viewer.zoomTo rectangle failed


zouyaoji:
Sandcastle example:

https://sandcastle.cesium.com/#c=bVJNbxoxEP0ro73EKxFv0ooeEkBVSG+VWpUoh3aryNjDYsXYyB6DSMV/rxfY7C7Bpxm/efP1psxiQAjktaQyuy8tnF5RzIRVUgQy+PKAlbYtJp0NBBuNW/QwBotbmGLQccWfD3+szOTBnzpLQlv0ZZbfn/OVIBFc9BL7OaYxkFs9JnR2QNkVYaCrvOUfK3P1HhK4UIq1CXNOS7RMBRhP4F9LBFCBoyVN+sTpgQAeJQlbGbyDM6Tu2nmlrUjd3PX6FcbMhXz96d0aPe0Yyz+UbbJT9Lbh/Wpq8YVP82LlEQO7/nLDhwP4/GkA18PaGObnifYDWAgT8AzYd919B+zaPVm/WdUi7T9faKvDEtV3J9LAFetq1wlL+3tyLs3uHyKRs0n2386tnlyZpQ6jlaSdhbSN3ipq8V1iG1exk5BvBxa7IOtLx/lz8zfvTdW7Vlg4DwrnsQJnza6Fttoqt+Xv13o0skE2CrQzOGkiv+rV2nmC6A3jvCBcrU2tdTGP8hWJyxCaiqOiSx0pvQGtxheOHqQRISRkEY2Z6Tcss8moSPEfqOa46h8b9Ebs6rDl7eS0f875qEjuZSYdJTjL/B8

As shown in the above code, the positioning of the rectangle loaded with CallbackProperty fails, and it has been in the pending state.

zouyaoji:
@ggetz 
Is there any follow-up?

ggetz:
@zouyaoji I'm planning to take a look this week and will let you know what I find.

ggetz:
Hi @zouyaoji, this is actually a duplicate of https://github.com/CesiumGS/cesium/issues/5164. I'll close this so that conversation can continue in that original issue. This happens since the callback property makes the geometry dynamic instead of static.

zouyaoji:
@ggetz 

But in some 1.8x versions, there is no such bug.

zouyaoji:
There is no such problem in version 1.89

zouyaoji:
@ggetz 
[...]

PR #10607, opened 2022-07-28
Fix bug where dynamic geometry was not marked as ready

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/10517

The above bug was being caused because the bounding sphere was never being computed due to the dynamic geometry never being marked as ready. 

This arose after https://github.com/CesiumGS/cesium/pull/10149/files. Previously, when.js would resolve promises synchronously in the same frame. Native promises on the other hand will resolve after a frame. In the case of dynamic geometry, a primitive will have been re-created, so `ready` always appeared to be `false` even if the old promise resolved.

Here, instead of waiting on the promise, we check the `ready` promise synchronously after each frame.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag (approved):
Can confirm that this fixes the issue shown in the Sandcastle in https://github.com/CesiumGS/cesium/issues/10517. Thanks @ggetz!

Issue #10379, opened 2022-05-13
`ModelExperimental` Sandcastle Testing


ptrgags:
## ModelExperimental Sandcastles

As part of https://github.com/CesiumGS/cesium/issues/10346, I did a sweep through the old 3D Tiles and Model Sandcastles to see what happens when I set `enableModelExperimental = true` or use `ModelExperimental.fromGltf` instead of `Model.fromGltf`.

EDITED: 07/08/2022

### 3D Tiles Sandcastles

Working Sandcastles:

* 3D Tiles Adjust Height
* 3D Tiles Batch Table Hierarchy
* 3D Tiles Compare
* 3D Tiles Feature Picking
* 3D Tiles Feature Styling
* 3D Tiles Formats
* 3D Tiles Inspector
* 3D Tiles Interior
* 3D Tiles Photogrammetry
* 3D Tiles Point Cloud Classification
* 3D Tiles Point Cloud Shading
* 3D Tiles Point Cloud
* 3D Tiles Terrain Classification
* Clamp to 3D Tiles
* Sample Height from 3D Tiles
* Development/3D Tiles Split

Broken Sandcastles:

* 3D Tiles BIM
  * Shader compilation error. Expected, this is glTF 1.0
* ~3D Tiles Batch Table Hierarchy~ -- fixed in #10525 
  * "Color all doors" doesn't work, as `batchTable.isExactClass` is not a function
  * "Color all features derived from door" doesn't work, as `batchTable.isClass` is not a function
* 3D Tiles Clipping Planes
  * Fails due to the glTF 1.0 BIM model
* ~3D Tiles Compare~ fixed in #10506 
  * OSM Buildings have incorrect shading (see #10338)
* ~3D Tiles Feature Styling~ fixed in #10514 
  * OSM Buildings have incorrect shading
  * Color by Distance to Selected Location crashes shader -- fixed by #10514 
* ~3D Tiles Formats~ fixed in #10524 
  * Wireframe mode will need `enableDebugWireframe: true`
  * Composite tileset is repeated for some reason
* ~3D Tiles Inspector~ fixed in #10526 
  * Wireframe mode will need `enableDebugWireframe: true`
  * Statistics are not available (not yet implemented)
* ~3D Tiles Interior~ fixed in #10524 
  * Some parts of the model (most notably the floor!) are not loading
* 3D Tiles Photogrammetry Classification
  * The classification isn't showing up
* 3D Tiles Point Cloud Styling
[...]

PR #10615, opened 2022-08-01
Convert the BIM tileset to avoid using techniques

ptrgags:
Since we deprecated the `KHR_techniques_webgl` extension, we wanted to update the BIM tileset used in several Sandcastle examples so they don't break when we remove the old Model (See #10379 and #10346).

The tileset was non-trivial to convert, and visually there are some differences including:
* The materials now use PBR, so the lighting looks a bit different
* Some dashed polylines in the original model now render as solid, as the custom GLSL code to render them doesn't translate directly to GLSL.

This PR updates the relevant sandcastles to use the new asset ID. It also adjusts the timeline and other minor visual tweaks.

Also note that I only updated thumbnails where the difference in lighting would be noticeable. 

Note that if you set `ExperimentalFeatures.enableModelExperimental = true`, all 5 of the affected sandcastles will render without crashing.

@j9liu could you review?


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/bim-no-techniques/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
Could you resize the new image to 225x150?

image

Apps/Sandcastle/gallery/Polylines on 3D Tiles.html
```diff
@@ -39,12 +39,56 @@
...
         const powerplant = scene.primitives.add(
           new Cesium.Cesium3DTileset({
-            url: Cesium.IonResource.fromAssetId(8564),
+            url: Cesium.IonResource.fromAssetId(1240402),
             show: false,
           })
         );
+
+        // Hide duplicate floor geometry to avoid
+        // z-fighting artifacts.
```
lilleyse, line 54: Glad to see this artifact finally addressed!

ptrgags:
@lilleyse I updated the images and squashed the commits!

PR #10624, opened 2022-08-02
Improve control over CustomShader translucency settings

ptrgags:
Originally, CustomShader had a `isTranslucent` flag that worked like this:

* `isTranslucent: true` - override the material settings and always render in the translucent pass
* `isTranslucent: false (default)` - override the material settings and always render in an opaque pass

However, this is sometimes confusing and may be limiting in cases where primitives have some opaque and some translucent materials.

Instead, I added `translucencyMode` with 3 states:

* ~NO_CHANGE~ `INHERIT` - Inherit settings from the primitive's material (i.e. if the shader is applied to an opaque primitive, the alpha channel will be ignored, but if applied to a translucent primitive, the alpha channel will be used)
* `TRANSLUCENT` - override the material and render as translucent
* `OPAQUE` - override the material and render as opaque.

@lilleyse could you review?

### DIFF
--- a/Apps/Sandcastle/gallery/Custom Shaders Models.html
+++ b/Apps/Sandcastle/gallery/Custom Shaders Models.html
@@ -171,8 +171,9 @@
               value: checkerboardTexture,
             },
           },
-          // necessary when setting material.alpha
-          isTranslucent: true,
+          // This model normally renders opaque so material.alpha would be ignored.
+          // this setting forces the shader to render in the translucent pass.
+          translucencyMode: Cesium.CustomShaderTranslucencyMode.TRANSLUCENT,
           fragmentShaderText: `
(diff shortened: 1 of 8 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/custom-shader-translucency-mode/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
This is pretty minor, but I noticed a comment that still uses the name `isTranslucent`

    // always check the current value, because custom shaders may
    // change the value with the isTranslucent flag
    return renderResources.alphaOptions.pass === Pass.TRANSLUCENT;

lilleyse:
The actual code looks good and I'm pretty much fine with the naming.

ptrgags:
@lilleyse updated!

Issue #10256, opened 2022-03-31
Make it easier to debug custom shader compile errors


ptrgags:
Right now, the way I debug compilation errors for a custom shader is putting a breakpoint in `ShaderProgram.createAndLinkProgram()` on one of the lines that throw `DeveloperError`, and then inspect the variables `vsSource` and `fsSource` which contain the complete shader text (custom shader + ModelExperimental shader code + CesiumJS builtin-stuff). But this is cumbersome and might be confusing to explain.

The tricky thing is the shader code is built in stages:

* `CustomShader` only has the code the user added
* `ModelExperimental` has a `ShaderBuilder` object when creating draw commands. This contains the complete shader, of which the custom shader is a small portion. This is usually the code I want to look at when debugging since it has more context about attributes and structs in the shader.
* the draw command has the `ShaderProgram` which has the full shader (ModelExperimental code + built in renderer code)

It would be great to make some of these more accessible for debugging. Some ideas:

1. Perhaps the easiest, instead of ShaderProgram throwing a `DeveloperError`, maybe it could throw a new `ShaderCompilationError` that includes the vertex and fragment text.
2. Or just `console.error` with the shader text
3. the `ShaderBuilder` could certainly have a method like `getFullFragmentShaderText()` that combines the text like it does when building the `ShaderProgram`, but returns the string. The only tricky part is how would the user access the shader builder, as this is pretty deep in the private API.
4. a similar method could be used for the `ShaderProgram`, but again it would be cumbersome to access and would be exposing more private API. 

I'm starting to like the sound of 1) then the user could catch the error and handle it as desired.

IanLilleyT:
> Perhaps the easiest, instead of ShaderProgram throwing a DeveloperError, maybe it could throw a new ShaderCompilationError that includes the vertex and fragment text.

I like this. And if the event has no listeners, it could throw a `DeveloperError`, like the pseudocode:

```
try {
    shaderProgram.compile();
    compilationEvent.raiseEvent(true);
} catch (error) {
    shaderProgram.destroy();
    if (compilationEvent.numberOfListeners === 0) {
        throw new DeveloperError("Error generating shader");
    } else {
        compilationEvent.raiseEvent(false);
    }
}
```

jjhembd:
> 2. Or just console.error with the shader text

This is what https://github.com/CesiumGS/cesium/pull/10629 does. 

[...]

PR #10629, opened 2022-08-03
Fix debugShaders output when shaders fail to compile

jjhembd:
In `ShaderProgram`, the `createAndLinkProgram` function was deleting the shaders before calling [WEBGL_debug_shaders.getTranslatedShaderSource()](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_debug_shaders/getTranslatedShaderSource). As a result, the problematic shader code was never output to the console as intended.

This PR reorders the `createAndLinkProgram` function to complete all logging and debug messages _before_ calling [gl.deleteShader](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/deleteShader).

This addresses the original problem in https://github.com/CesiumGS/cesium/issues/10256, although that issue mentions some ideas for further improvement, which are not implemented here.

### DIFF
--- a/Source/Renderer/ShaderProgram.js
+++ b/Source/Renderer/ShaderProgram.js
@@ -245,12 +245,12 @@ function createAndLinkProgram(gl, shader) {
   if (!gl.getShaderParameter(fragmentShader, gl.COMPILE_STATUS)) {
     log = gl.getShaderInfoLog(fragmentShader);
     console.error(`${consolePrefix}Fragment shader compile log: ${log}`);
-    logTranslatedSource(fragmentShader, "fragment");
+    console.error(`${consolePrefix} Fragment shader source:\n${fsSource}`);
     errorMessage = `Fragment shader failed to compile.  Compile log: ${log}`;
   } else if (!gl.getShaderParameter(vertexShader, gl.COMPILE_STATUS)) {
     log = gl.getShaderInfoLog(vertexShader);
     console.error(`${consolePrefix}Vertex shader compile log: ${log}`);
-    logTranslatedSource(vertexShader, "vertex");
+    console.error(`${consolePrefix} Vertex shader source:\n${vsSource}`);
     errorMessage = `Vertex shader failed to compile.  Compile log: ${log}`;
   } else {
     log = gl.getProgramInfoLog(program);
--- a/Source/Renderer/ShaderProgram.js
+++ b/Source/Renderer/ShaderProgram.js
@@ -197,9 +197,6 @@ function createAndLinkProgram(gl, shader) {
   gl.attachShader(program, vertexShader);
   gl.attachShader(program, fragmentShader);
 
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/shader-errors/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag:
@jjhembd If I introduce a syntax error in the shader, am I expecting to see the full shader source in the console log? I just tried introducing a syntax error on `main` and on this branch, and the output is identical.

jjhembd:
Thanks for checking @sanjeetsuhag!
I just confirmed your test. A syntax error will make an individual shader fail to compile, in which case [debugShaders.getTranslatedShaderSource](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_debug_shaders/getTranslatedShaderSource) returns nothing.
The difference will be when the vertex and fragment shaders both compile, but `gl.linkProgram` fails. Then `debugShaders.getTranslatedShaderSource` will return the translated source from the previous compile stages.
I'm not sure yet how to make a simple example of a linker error.

jjhembd:
I just pushed another edit: 
For the case where the vertex or fragment shader fails to compile, it now logs the raw GLSL instead of attempting to get the (non-existent) translated source.
This should be more useful for finding simple syntax errors in Custom Shaders.

sanjeetsuhag:
Courtesy of @lilleyse, I was able to trigger a link error by adding a `varying`s of different types in a shader (I used the `SkyAtmosphereVS.glsl` and `SkyAtmosphereFS.glsl`) and reading the value in the fragment shader. And the code output was in HLSL.

jjhembd:
Thanks! I tried that and the output makes sense. On my machine (Chrome on Windows) it prints both GLSL and HLSL. Both are output from [getTranslatedShaderSource](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_debug_shaders/getTranslatedShaderSource):
```
[Cesium WebGL] Shader program link log: Types of varying 'v_test' differ between VERTEX and FRAGMENT shaders.
FRAGMENT varying v_test does not match any VERTEX varying

[Cesium WebGL] Translated vertex shaderSource:
// VERTEX SHADER BEGIN

// GLSL BEGIN

#define LOG_DEPTH
#define OES_texture_float_linear

/* ... */

void main() 
{ 
    czm_log_depth_main(); 
    czm_vertexLogDepth(); 
} 


// GLSL END


// INITIAL HLSL BEGIN

#pragma warning( disable: 3556 3571 )
struct _czm_ray
{
    float3 _origin;
    float3 _direction;
};
```

sanjeetsuhag (approved):
Looks good. Thanks @jjhembd!

Issue #10555, opened 2022-07-15
camera.flyTo goes underground 


angrycat9000:
`camera.flyTo` ends up underground in a location where the globe does not have a height.

The expected result is that that camera flys to 100 meters above the ground at the given location and the vehicle is visible on the surface of the ground.

[Sandcastle example](https://sandcastle.cesium.com/#c=hVNdb9owFP0rVrSHIDFnDKoNGtAmkLqHrlQt614stca+BGuOjWyHilX89zk2MMKYeIl8zz33w8cnTCvr0FrAKxg0RApe0RisqEr8FLCUJCzEY60cFQoMSdrojSiEHBjjkXuj14KDGewLmQHq4Kc2ks8iJW21idq2rokiioWBUqtCuIrDBAoDYP3oTq+PrzpX3e7H6wOLuhPS+24P9z/3PvX7R81W2gontPL53Qpjapw/UdXFC6PLXX1ab306uR3A5qCAfSDqeGNQnrLxI6JWOMQCLKacp0GP/RqDwyn0KTUHOYiSIVQZMUAkwTh7pOVKwoQ6mgWKzW6MrhR/gqVgEpoRLuTcCx97LEEUS/cACzCgGByE/9bE8fj26/f759n0+eZh+uNuEqq3py/BvFS6MHTl5zQF3KOXJUQXNZTg9n6JW/4V0jJQ9f303H/BxXR6vFZrZwjtdZC6SF8anYbv3hrx9iWObLRntARD8UJuZjq+FgfrhKLxwS7b5j/GOX/tcHGEDq5P2klu3UbCKGYR+iLKlTbO20Gm3gsOvBf8T2OzecV+gcPM2rqwpubZcWnOxRoJPjzzWyImqbU+s6ikfBS/gSSjPPP8f0qlplyoYroGI+mmpi07o9sIYozzzIfnK53Wck7NSec/)

Browser: Chrome 103.0.5060.114

Operating System: MacOS

mramato:
Regarding:

```
let terrainHeight = viewer.scene.globe.getHeight(cartographic);
console.log(`terrainHeight=${terrainHeight}`);
```

It's correct behavior that `terrainHeight` is undefined here because `globe.getHeight` only works with terrain tiles that are actually loaded and only suitable for querying data as-is on-screen.  You need to use the async `sampleTerrainMostDetailed` to get the true height of terrain at a location, which will definitely return a height.

[...]

PR #10631, opened 2022-08-03
Fixes Viewer.flyTo for models with height references

sanjeetsuhag:
Fixes https://github.com/CesiumGS/cesium/issues/10555. More details provided in https://github.com/CesiumGS/cesium/issues/10555#issuecomment-1204045063.

The primary change made here is to use `sampleTerrainMostDetailed` to derive the result in `ModelVisualizer.getBoundingSphere`, instead of the `ModelExperimental._clampedModelMatrix`, which may initially be computed from a lower LOD tile causing the camera to fly below the ground.

Use [this Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rVTvb9owEP1XLL40SMxpR6uNH0XrgLWVWqiAsQ/LVLnOAdYcG9kOHav432fHJJSMbv1QCaT4/O7du3eXUCm0QSsGj6DQORLwiLqgWZrgaRYLogrNzl0pDGECVFSpoadIIGRAKRu5U3LFYlDNPJEqIAa+ScXjiYcE1VokNtVWJCJBs4JcijkzaQw9mCsAbUufnDbw2clZvf6+VaCIKYHe1U9x4+Pph0ajAC2AzRfGERwfHz8rsZSaGSaFvdkK6xJl7BMRdTxTMtmyBq6Xsp5aFtwvn8V8tUj4ZjgYBMKi1raKNxFnZwYakzgOMqNyJc3iKaNKZAy86b1EKFWsiaIKxqH9jUmy5NAjhoQZSoeXSqYinsKCUQ77JzznD3YonsbrG8EMFAgKxVCu9uN4MBz0XcZmN5lIzFJBM8s0mHQZlLiqXqr7l1tVkMgVBN4Kx+Zh/7fmRXPK9ryZQQcsKgU8bpPhN3kzm91iyaVT6bbxu7vbCjTwyzh9A4m812hHmNe2WcCBWlzhdFDddehd/8fAvJhCW+1A+S63ziAjkXfgTUp3by5u7+4nw/vL0fDroPcKFSNw784K3ljIqH9zMbme9l/UEokf2SqPiYgp0cbO3q7aREr+QNQtiDTYDs/lHQR9To2RIjj6wtdoIo9qTtx5J1/gGV9PZL7mr+K4eCTrguUp56EkAUUc3ZVMIKi23KJVapW2NmsOndyMTyxZSmXs6vPALr0Bu/T206rDh5T+BIOp1rkB7fB5ajtmK8Ti8wMfb0Q50drezFLOx+y33c5OO7T4v1K5JDET8+EKFCdrB1ucdG58EGPcDu3xcKbxPpSY/wA) to see the results below.

| | ` NONE ` | `CLAMP` | `RELATIVE` |
| - | - | - | - |
| main | image | image | image | 
| PR | image | image | image |

The scenario used in the modified spec is visualized in [this Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=zVVtb9s4DP4rQj4lgKtIst6cZcV2adEbMKxD23VfAgyuzSTCFCuQ5e56h/33k+20eVlXbLgvhwSwSJF8SD6kXbiqDujewDfw6DWq4BuaQW2aNb7tdMP5oOjkmatCbirw80GC/plXCAXwPmo+endvSvCTR8fCQx7gs/O2vOlNhqNkXn0fvZpX7a/oIIt8DT6PkD027uVo0h/wxtUmGFcdJjXLfYinvEqHjKRpljEslZKSMaYTdEIF55kiGjMuJRGKJYiSjOs0k1gKpjknmRztQErjoXgR5YRgSqTUjEnGqRREa5UggjWN8IylqSCCMCoiOMFCSaqU4IprxTTlfcV9vR7qxoY/XFOVplpeb1bgYQZVOG77PjYXJBOpasE011LLBDEpuMwEliSlhCmdReBUphFU4FgelzoTogPuYeGvTawQyt8E5porlmKlOKEyy3QLzAnRCmeSZJoq2gNTlcZkaEw0PhXdA/Z5aZo6QnAstsQftmJHfUwmUg01zsty2I3WI/mTF9qWtIZgrdnUzpSTfiZRh2smPymsTyo5eIyS3nFhrJ2gRW5r2GpcE2yc+AkKvjnSzZx1u4nvJHxxdX7+YWtWW1PAxwjclVFPEGWPNyEvvj5z8323IofM/VqfXuL5f9epq/Oz/96nvkt1ARXsWtSJe020rlqa0JRwBksP0M4j5RkWVMTtfQpi83BkdJJynMUdyLK9YHuvpB8ahhferbf+wzbPY+Qu+SOgTkcJIfPqgPmW54df4/3x1IVauxLsE7+Nj+zOBxiP4/86X28snOUhH3dW9fjCt6NyCytTWDiU8NLexdd8H2YFZrkKV7CI81QV8ETln4f6SOr7tzfvbs+/3Fx+ubi6/PTh7Hiut+Us7MONG/aLPXo1SAbTOjxYOO3hEHpj1hvnQ8zfDmPmAWLm8YNSj++a4isEXNR1G681nY73XaeluUemfP3MJwsVNq/reLNorL02f8N8cDodR/sfXK3L2/25vAdv84fWbEVP3/dKjPF0HMXnPYNz9i73R5H/BQ). I did not want to use exact values (in case the terrain data is later updated), hence only the inequality check.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/camera-model-fly-to-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
@sanjeetsuhag Would it make sense to have @angrycat9000 review this PR for behavior?

j9liu:
@sanjeetsuhag just a drive-by comment, I noticed the "this sandcastle" link doesn't direct to a sandcastle, it just links to this PR :')

sanjeetsuhag:
Thanks @j9liu. Seems like I forgot to copy the link. Updated now.

Source/DataSources/ModelVisualizer.js
```diff
@@ -370,8 +376,68 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
...
+            clampedBoundingSphere.center = scratchPosition;
+
+            this._modelHash[
+              entity.id
+            ].clampedBoundingSphere = BoundingSphere.clone(
+              clampedBoundingSphere
+            );
+          }
+        );
+        this._modelHash[entity.id].awaitingSampleTerrain = true;
```
angrycat9000, line 431: - Expected this get set to `false` when the promise is resolved, but didn't see that.
- From a readabilty perspective it would be better to set this before the then clause.  That puts key change all at the top of the if.  It also keeps the code in the same reading order as it would be executed.  Now I'm reading through the `then` where I have the terrain and get to this line and have to rewind back in time to before the `then` ran.  This is also why `async/await` tends to read better.

Specs/DataSources/ModelVisualizerSpec.js
```diff
@@ -372,36 +373,39 @@ describe(
...
-        return state !== BoundingSphereState.PENDING;
-      }).then(function () {
-        expect(state).toBe(BoundingSphereState.DONE);
-        expect(modelPrimitive._clampedModelMatrix).toBeDefined();
-        const expected = BoundingSphere.clone(
-          modelPrimitive.boundingSphere,
-          new BoundingSphere()
+      scene.globe.terrainProvider = createWorldTerrain();
+
+      return scene.globe.terrainProvider.readyPromise.then(() => {
```
angrycat9000, line 378: Why do we need to await this in the test?   What happens if this is still pending in production when `getBoundingSphere` is called?

angrycat9000 (commented):
Thanks for fixing this @sanjeetsuhag .  I verified that the sandcastle now works.

Had a few comments about the code.   Overall looks good. 

Would be good if someone else familiar with the Cesium base could give it a review as well @ggetz

Specs/DataSources/ModelVisualizerSpec.js
sanjeetsuhag: Since `sampleTerrainMostDetailed` already awaits the terrain provider's `readyPromise`, I suppose this is not needed in the spec.

Source/DataSources/ModelVisualizer.js
```diff
@@ -370,8 +376,68 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
...
+      clampedBoundingSphere = new BoundingSphere();
+
+      // Since this function is called per-frame, we set a flag when sampleTerrainMostDetailed
+      // is called and check for it to avoid calling it again.
+      const awaitingSampleTerrain = this._modelHash[entity.id]
+        .awaitingSampleTerrain;
+      if (!awaitingSampleTerrain) {
+        Cartographic.clone(cartoPosition, scratchCartographic);
+        sampleTerrainMostDetailed(terrainProvider, [scratchCartographic]).then(
+          () => {
```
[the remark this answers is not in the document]
sanjeetsuhag: > Is there a standard for that in the Cesium codebase?
We haven't started using `async`/`await` in the main codebase just yet. I think it's only used in our scripts.

Source/DataSources/ModelVisualizer.js
sanjeetsuhag: > Expected this get set to false when the promise is resolved, but didn't see that.

Good catch. Fixed.

> From a readabilty perspective it would be better to set this before the then clause. 

I agree, updated.

Specs/DataSources/ModelVisualizerSpec.js
```diff
@@ -372,15 +373,17 @@ describe(
     });
 
     it("Computes bounding sphere with height reference", function () {
+      scene.globe.terrainProvider = createWorldTerrain();
+
       const time = JulianDate.now();
       const testObject = entityCollection.getOrCreateEntity("test");
       const model = new ModelGraphics({
-        heightReference: HeightReference.CLAMP_TO_GROUND,
+        heightReference: HeightReference.RELATIVE_TO_GROUND,
```
ggetz, line 381: Since there's potential for different behavior between the two, should there be a test for both `CLAMP_TO_GROUND` and `RELATIVE_TO_GROUND`?

Specs/DataSources/ModelVisualizerSpec.js
```diff
@@ -396,12 +399,11 @@ describe(
         return state !== BoundingSphereState.PENDING;
       }).then(function () {
         expect(state).toBe(BoundingSphereState.DONE);
-        expect(modelPrimitive._clampedModelMatrix).toBeDefined();
         const expected = BoundingSphere.clone(
           modelPrimitive.boundingSphere,
           new BoundingSphere()
         );
-        expect(result).toEqual(expected);
+        expect(result).not.toEqual(expected);
```
ggetz, line 406: Can we be more specific on what's expected here?

Source/DataSources/ModelVisualizer.js
```diff
@@ -370,8 +376,70 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
...
+    // for the camera.
+    let clampedBoundingSphere = this._modelHash[entity.id]
+      .clampedBoundingSphere;
+
+    if (!defined(clampedBoundingSphere)) {
+      clampedBoundingSphere = new BoundingSphere();
+
+      // Since this function is called per-frame, we set a flag when sampleTerrainMostDetailed
+      // is called and check for it to avoid calling it again.
+      const awaitingSampleTerrain = this._modelHash[entity.id]
```
ggetz, line 405: Do we initiate the properties of `_modelHash[id]` anywhere? Or do we dynamically add properties to it elsewhere in the code?

I ask because [adding dynamic properties can have an impact on performance](https://dev.to/fromaline/why-does-dynamically-adding-properties-is-slow-in-javascript-4hm8).

Source/DataSources/ModelVisualizer.js
```diff
@@ -370,8 +376,70 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
...
+              scratchPosition
+            );
+
+            // Update the bounding sphere with the updated position.
+            BoundingSphere.clone(model.boundingSphere, clampedBoundingSphere);
+            clampedBoundingSphere.center = scratchPosition;
+
+            this._modelHash[
+              entity.id
+            ].clampedBoundingSphere = BoundingSphere.clone(
```
ggetz, line 429: Since `clampedBoundingSphere` is created, modified, and accessed only in the scope of this function, I don't think this `clone` is necessary.

ggetz:
Thanks @sanjeetsuhag! I have a few comments, and I think there may be some side effects of the spec changes which are causing other specs to fail. You may need to destroy the terrain provider after the spec is completed.

Specs/DataSources/ModelVisualizerSpec.js
sanjeetsuhag: Added a more detailed check that uses the value returned from `sampleTerrainMostDetailed`.

Source/DataSources/ModelVisualizer.js
sanjeetsuhag: Added these values to when `modelData` is initialized.

Source/DataSources/ModelVisualizer.js
sanjeetsuhag: I think this one is necessary. Removing it break the flyTo function and it flies under the ground.

Specs/DataSources/ModelVisualizerSpec.js
sanjeetsuhag: Added specs for each.

sanjeetsuhag:
Seems like tests on Chrome are failing in CI. Looking into it now.

sanjeetsuhag:
Testing this locally, I am unable to reproduce the CI error at all.

sanjeetsuhag:
@ggetz Seems like CI passed here. Not sure how those first failures happened.

ggetz:
Thanks @sanjeetsuhag. I'm re-running the build just to ensure reliability going forward. Assuming that passes, this should be ready to merge.

ggetz:
Please merge in `main` when you get the chance.

sanjeetsuhag:
`sampleTerrainMostDetailed` is buggy on CI. I think I'll just fake the call in the spec.

sanjeetsuhag:
@ggetz Updated.

ggetz:
Thanks @sanjeetsuhag! It's definitely less than ideal for the specs that `sampleTerrainMostDetailed` is non-deterministic. The spy is an OK strategy for circumventing this, but it does make the tests a bit more brittle in the ling run. The [`sampleTerrainMostDetailedSpec` instead accepts values within a certain range](https://github.com/CesiumGS/cesium/blob/main/Specs/Core/sampleTerrainMostDetailedSpec.js#L25). Is it possible to do the same here?

sanjeetsuhag:
@ggetz Updated - incorporated a margin of error and used the actual `sampleTerrainMostDetailed` instead of a mock. I still exposed it as a property of `ModelVisualizer` because it's useful to create a spy and check that it's only being called once from `ModelVisualizer.getBoundingSphere`.

ggetz:
Thanks @sanjeetsuhag!

Source/DataSources/ModelVisualizer.js
```diff
@@ -364,8 +375,68 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
...
+
+          // Update the bounding sphere with the updated position.
+          BoundingSphere.clone(model.boundingSphere, clampedBoundingSphere);
+          clampedBoundingSphere.center = scratchPosition;
+
+          this._modelHash[
+            entity.id
+          ].clampedBoundingSphere = BoundingSphere.clone(clampedBoundingSphere);
+        });
+      }
```
mramato, line 431: Shouldn't there be a catch with error handling here? Otherwise it is just stuck in awaiting forever, isn't it?

PR #10623, opened 2022-08-02
Make `ModelExperimental` handle classification

j9liu:
This PR adds support for classification models in `ModelExperimental`. This was accomplished without needing to create an equivalent of `ClassificationModel`. Instead, a `ClassificationModelDrawCommand` is created to handle the special classification commands.

classification

[...]

PR #10635, opened 2022-08-04
Remove old b3dm/i3dm/pnts content classes

ptrgags:
As part of #10346, this PR removes `Batched3DModel3DTileContent`, `Instanced3DModel3DTileContent` and `PointCloud3DTileContent`. `Model3DTileContent` will be used instead.

A couple of notes:

* I also removed specs for b3dm classification, as #10623 will handle this
* I disabled the tests for `Cesium3DTileBatchTable` until we can consolidate that with `ModelFeatureTable`, since the tests used `b3dm` files which no longer use the old style of batch table (only `.vctr` and `.geom` do)

@j9liu could you review?


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
@j9liu I checked the Sandcastles and they're all good except for the ones with the BIM model (since that change is only on `main`). I'm going to merge `main -> replace-model` then `replace-model -> remove-1.0-contents` so we can better test those Sandcastles.

Specs/Cesium3DTilesTester.js
```diff
@@ -193,6 +197,7 @@ Cesium3DTilesTester.rejectsReadyPromiseOnError = function (
       fail("should not resolve");
     })
     .catch(function (error) {
+      console.log(error);
```
j9liu, line 200: Was this added for debugging purposes, or is this an intentional change?

ptrgags:
@j9liu okay, the Sandcastles with the BIM tileset should work correctly now!

Source/Scene/ModelExperimental/Model3DTileContent.js
```diff
@@ -384,7 +393,9 @@ Model3DTileContent.fromGeoJson = function (tileset, tile, resource, geoJson) {
     content,
     additionalOptions
   );
-  content._model = ModelExperimental.fromGeoJson(modelOptions);
+  const model = ModelExperimental.fromGeoJson(modelOptions);
+  content._model = model;
+  content._readyPromise = model.readyPromise;
   return content;
```
j9liu, line 399:
```suggestion
  content._readyPromise = model.readyPromise;
  
  return content;
```
one more nitpick for consistency with the other `fromXXX` functions

j9liu (changes requested):
@ptrgags looks good for the most part, just a few comments!

I'll run through the Sandcastles locally as well to make sure no new bugs came up

Specs/Cesium3DTilesTester.js
ptrgags: oops forgot to remove it :sweat_smile:

ptrgags:
@j9liu oops forgot to mention the PR is ready for review again!

j9liu:
I found a regression but I don't think it has to hold up this PR. In the 3D Tiles Formats sandcastle, `PointCloudDraco` isn't properly responding to the "Debug Colorize" option in the inspector:
draco point

For reference, this is what it does on [sandcastle.com](https://sandcastle.cesium.com/?src=3D%20Tiles%20Formats.html&label=3D%20Tiles) right now:

draco point main

Everything else looks good. Thanks @ptrgags !

PR #10643, opened 2022-08-05
Remove ability to toggle `classificationType` at runtime

j9liu:
Due to a picking bug with the current implementation of classification with `Model` (issue to be opened soon), the way that classification is done in `Model` needs to be somewhat refactored. This will involve loading some attributes as typed arrays, which means toggling `classificationType` at runtime is no longer possible for `Model`. This PR removes that ability.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
Thanks @j9liu!

PR #10688, opened 2022-08-15
Make it easier to debug render tests

ptrgags:
Given that render tests render to 1x1 px canvases, it can be a pain to debug because it's hard to tell what the camera is seeing. Over the weekend I had an epiphany: What if you could temporarily make the canvas size bigger while debugging? That would be a lot easier than trying to replicate the test setup in Sandcastle.

This PR does just that. it adds the parameters `debugWidth` and `debugHeight` to `createScene()`. Example usage:

1. Use `fit()` on the render test to be debugged so only that test runs
2. Find the `createScene()` call in `beforeAll()`/`beforeEach()`, and set a larger canvas size:

```js
scene = createScene({
  debugWidth: 300,
  debugHeight: 300
});
```
3. Put a breakpoint in the test around when the first render call is
4. Step through the test using the debugger. After every call to `scene.renderForSpecs()`, check the browser window, you'll see a better picture of what was rendered
image

@sanjeetsuhag could you review? Curious if you have any thoughts of other improvements for test debugging

### DIFF
--- a/Documentation/Contributors/TestingGuide/README.md
+++ b/Documentation/Contributors/TestingGuide/README.md
@@ -36,6 +36,7 @@ All new code should have 100% code coverage and should pass all tests. Always ru
     - [Testing Exceptions](#testing-exceptions)
     - [Before and After Tests and Suites](#before-and-after-tests-and-suites)
     - [Rendering Tests](#rendering-tests)
+      - [Debugging Rendering Tests](#debugging-rendering-tests)
     - [GLSL](#glsl)
     - [Spies](#spies)
     - [Test Data and Services](#test-data-and-services)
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/specs-debug-canvas-size/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag:
@ptrgags This is working well so far, could be very useful in catching tricky testing bugs or just setting up better tests in general. I think the main thing missing in the PR is maybe a section in the [Testing Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/TestingGuide) that talks about using this.

ptrgags:
@sanjeetsuhag good idea, I updated the Testing Guide

sanjeetsuhag (approved):
Looks good @ptrgags! Thanks!

Documentation/Contributors/TestingGuide/README.md
```diff
@@ -502,6 +503,46 @@ it("can declare automatic uniforms", function () {
...
+1. Use `fit()` to focus on the test that is failing.
+2. Use the `debugWidth` and `debugHeight` options for `createScene()` to make the canvas larger.
+3. Put a breakpoint in the code around where the first rendering code happens, such as a call of `scene.renderForSpecs()`.
+4. Step through the test. After each render, check the browser window to see the frame that was just rendered.
+
+```js
+beforeAll(function () {
+  scene = createScene({
+    // Create a 300x300 px canvas so we can see what the camera sees.
+    // This is intended for temporary debugging, do not commit these debug options.
```
ggetz, line 524: Would it be possible to send these options down with a command line argument instead of in the code? That way it wouldn't get accidentally committed.

Issue #10684, opened 2022-08-12
Cesium no longer works under Node.js when using ESM modules


mramato:
1. Create a file called `test.mjs` with contents `import { Cartographic } from "./Source/Cesium.js";`
2. Run node `test.mjs`
3. It fails with the below callstack

```
$ node test.mjs 
file:///home/mamato/Git/CesiumGS/cesium/node_modules/@zip.js/zip.js/lib/core/streams/crc32-stream.js:33
class Crc32Stream extends TransformStream {
                          ^

ReferenceError: TransformStream is not defined
    at file:///home/mamato/Git/CesiumGS/cesium/node_modules/@zip.js/zip.js/lib/core/streams/crc32-stream.js:33:27
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

I did a little digging and it appears `zip.js` module released a new version in the past month that caused this. Pinning to an older version fixes the issue (need to experiment to find out the exact version). May be worth opening an issue with them to see what they have to say.

Once we address this, it would be good to add the test.mjs and run it as part of CI to avoid breaking ESM support in the future.

CC @ggetz

ggetz:
Thanks for the report @mramato! I opened a fix for this in https://github.com/CesiumGS/cesium/pull/10694/. Looks like [the zip.js maintainer addressed this and suggests](https://github.com/gildas-lormeau/zip.js/issues/340#issuecomment-1177329994) either locking to `2.4.x` or adding a polyfill for `TransformStream`. I've locked it in the above PR.

mramato:
Thanks @ggetz!

Should we write up another issue to undo the pinned npm version once the zip.js maintainer has a better fix on their end?

ggetz:
Opened https://github.com/CesiumGS/cesium/issues/10712

Issue #9212, opened 2020-10-22
Can not import widgets.css file not exported from package.json


3DGISKing:
I am working on https://github.com/CesiumGS/cesium-webpack-example.
This is very simple web pack configured web app.

It works well with webpack 4(original version).
But after I update web pack version, it shows following error.

Module not found: Error: Package path ./Build/Cesium/Widgets/widgets.css is not exported from package E:\MyResearch\CesiumJsGIS\node_modules\cesium (see exports fiel
d in E:\MyResearch\CesiumJsGIS\node_modules\cesium\package.json)

In a short I could not import css file from node_module.

hope you export widgets.css in package.json.

bampakoa:
Hey @3DGISKing I think it would make sense to file an issue to the cesium-webpack-example repository instead 😉

3DGISKing:
No. this happens because webpack5 updates module resolve logic.

bampakoa:
@3DGISKing I am not sure if it is indeed related to this repo, (it may have to do with the webpack configuration) but I think that reporting it there would be a good start for filing the issue 😃 . However, either way, looks valid to me 👍

3DGISKing:
please have a look at this.

https://github.com/webpack/webpack/issues/11787

bampakoa:
@3DGISKing i do not know if it is ok to export CSS files through `package.json`, the maintainers of Cesium can tell that, but you have always the option of the `copy-webpack-plugin`

TJKoury:
Try [C137.js](https://www.npmjs.com/package/c137.js), webpack example [here](https://github.com/DigitalArsenal/cesium-webpack-example/tree/c137.js)

dzungpng:
Closing this issue as we have not received a response from the author in a while. Please re-open if there is additional discussion to be made.

thw0rted:
Please re-open this issue.

I've never had to set up the `exports` field in a `package.json` before but apparently it's a pretty big deal and can break a lot of things, especially now that Webpack 5 fully respects it.  I'm going by the [webpack docs](https://webpack.js.org/guides/package-exports/) which refer to the [node docs](https://nodejs.org/api/packages.html#packages_package_entry_points) here.  (cc @kring because [he created the current version](https://github.com/CesiumGS/cesium/commit/47fa26074b8a98efb78af8ad6c7225d329672a58#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R34) )

[...]

PR #10694, opened 2022-08-16
Fix for NodeJS with ESM modules

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/10684

NodeJS was failing to run ESM modules mainly because of a recent update to `zip.js` which requires `TransformStream` to be defined. Looks like [the zip.js maintainer addressed this and suggests](https://github.com/gildas-lormeau/zip.js/issues/340#issuecomment-1177329994) either locking to `2.4.x` or adding a polyfill for `TransformStream`. Here, I've locked the package version.

To prevent related issues from going unnoticed in the future, I added a `.mjs` version of `test.cjs`, which is now run as part of CI.

In doing so, I also caught two other issues. One was a faulty import, and the other was the use of `require` in `Resource.js`. If I understand correctly, this is code that should only ever be executed in a NodeJS environment, and can be replaced with a dynamic import , satisfying both the ESM and CJS use cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,5 +30,6 @@ script:
   - node -e "const Cesium = require('./');"
   - NODE_ENV=development node Specs/test.cjs
   - NODE_ENV=production node Specs/test.cjs
+  - node Specs/test.mjs
 
   - npm --silent run cloc
--- a/Source/Core/Resource.js
+++ b/Source/Core/Resource.js
@@ -2059,59 +2059,65 @@ function loadWithHttpRequest(
   overrideMimeType
 ) {
   // Note: only the 'json' and 'text' responseTypes transforms the loaded buffer
-  /* eslint-disable no-undef */
-  const URL = require("url").parse(url);
-  const http = URL.protocol === "https:" ? require("https") : require("http");
-  const zlib = require("zlib");
-  /* eslint-enable no-undef */
+  let URL;
+  let zlib;
+  Promise.all([import("url"), import("zlib")])
+    .then(([urlImport, zlibImport]) => {
+      URL = urlImport.parse(url);
+      zlib = zlibImport;
 
-  const options = {
-    protocol: URL.protocol,
-    hostname: URL.hostname,
-    port: URL.port,
-    path: URL.path,
-    query: URL.query,
-    method: method,
-    headers: headers,
-  };
+      return URL.protocol === "https:" ? import("https") : import("http");
+    })
+    .then((http) => {
+      const options = {
+        protocol: URL.protocol,
+        hostname: URL.hostname,
+        port: URL.port,
+        path: URL.path,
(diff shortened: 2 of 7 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

.travis.yml
```diff
@@ -30,5 +30,6 @@ script:
   - node -e "const Cesium = require('./');"
   - NODE_ENV=development node Specs/test.cjs
   - NODE_ENV=production node Specs/test.cjs
+  - node Specs/test.mjs
```
sanjeetsuhag, line 33: Do we need to test this in both `production` and `development` mode too, like we do with `test.cjs`? Looking at `test.cjs`, it is not obvious to me what the point of setting the environment to `production` is.

Source/Core/Resource.js
```diff
@@ -2059,59 +2059,65 @@ function loadWithHttpRequest(
...
-    .on("response", function (res) {
-      if (res.statusCode < 200 || res.statusCode >= 300) {
-        deferred.reject(
-          new RequestErrorEvent(res.statusCode, res, res.headers)
-        );
-        return;
-      }
+  let URL;
+  let zlib;
+  Promise.all([import("url"), import("zlib")])
```
sanjeetsuhag, line 2064: Just to make sure - these imports are cached, right?

sanjeetsuhag (commented):
This looks good to me, @ggetz. I just have a couple of things I wanted to clarify.

.travis.yml
ggetz: Good question! So `test.cjs` imports from the root `cesium` module. In out `package.json`, we've defined [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) such that requiring or importing `cesium` which grab a different file depending on the environment. In CJS with a require, we redirect to `index.cjs`, which in turn requires either the minified release or unminified release depending on whether or not the environment is production. (We've recently had [some discussion whether or not `index.cjs` and the runtime operations are needed or not](https://github.com/CesiumGS/cesium/issues/9212#issuecomment-1218449631).)

MJS on the other hand can handle ESM directly, so imports from `Source/Cesium.js` file. We could (and probably should) have `test.mjs` import from the root `cesium`, and it should be functionally equivalent to `Source/Cesium.js`, and so would also verifies our `package.json` configuration as well.

Source/Core/Resource.js
ggetz: This code should only ever be executed in environments where `XMLHttpRequest` is undefined, which is not in the browser but in NodeJS. In that case, the are dynamically imported from the `node_modules` directory.

sanjeetsuhag (approved):
Looks good. Thanks @ggetz!

PR #10695, opened 2022-08-16
Control unit test canvas size from the CLI/SpecRunner URL

ptrgags:
This is a revision to #10688 based on this feedback: https://github.com/CesiumGS/cesium/pull/10688#discussion_r946832437

This PR adds the `debugCanvasWidth` and `debugCanvasHeight` parameters to `npm run test` as well as SpecRunner's URL query parameters. For example:

```
npm run test -- --debugCanvasWidth 300 --includeName Model
```

or in the browser:

```
http://localhost:8080/Specs/SpecRunner.html?spec=Model&debugCanvasWidth=300
```

@ggetz could you review?

PR #10696, opened 2022-08-16
Debug ShadowMap tests

ptrgags:
This PR re-enables a few tests that were `xit-ed` out in `ShadowMap` and fixes them so they run again. The problems were:

1. Due to rendering differences The rendering color wasn't an exact match in some cases
2. For the point light test, it wasn't waiting for the ready promise, so the boxes weren't showing up. I changed it to use a single box and just reposition it in the scene as needed.

@j9liu this is ready for review, though wait for #10695 to be merged before merging

### DIFF
--- a/Specs/Scene/ShadowMapSpec.js
+++ b/Specs/Scene/ShadowMapSpec.js
@@ -771,7 +771,7 @@ describe(
         const red = rgba[0];
         const unshadowedRed = unshadowedColor[0];
         const shadowedRed = shadowedColor[0];
-        expect(shadowedRed).toBeLessThan(red);
+        expect(red).toBeGreaterThan(shadowedRed);
         expect(red).toBeLessThan(unshadowedRed);
       }, endTime);
 
@@ -835,8 +835,6 @@ describe(
       scene.light = new DirectionalLight({
         direction: lightDirectionAngle,
       });
-
-      // Change the light direction so that the shadows are no longer pointing straight down
       renderAndCall(function (rgba) {
         expect(rgba).not.toEqual(backgroundColor);
         expect(rgba).not.toEqual(shadowedColor);
@@ -847,7 +845,7 @@ describe(
         const red = rgba[0];
         const unshadowedRed = unshadowedColor[0];
         const shadowedRed = shadowedColor[0];
-        expect(shadowedRed).toBeLessThan(red);
+        expect(red).toBeGreaterThan(shadowedRed);
         expect(red).toBeLessThan(unshadowedRed);
       });
 
--- a/Specs/Scene/ShadowMapSpec.js
+++ b/Specs/Scene/ShadowMapSpec.js
@@ -74,6 +74,8 @@ describe(
     let primitiveFloor;
     let primitiveFloorRTC;
 
+    let originalLight;
+
     beforeAll(function () {
       scene = createScene();
       scene.frameState.scene3DOnly = true;
@@ -306,6 +308,10 @@ describe(
       scene.destroyForSpecs();
     });
 
+    beforeEach(function () {
+      originalLight = scene.light;
+    });
+
     afterEach(function () {
       const length = scene.primitives.length;
       for (let i = 0; i < length; ++i) {
@@ -314,6 +320,7 @@ describe(
 
       scene.globe = undefined;
       scene.shadowMap = scene.shadowMap && scene.shadowMap.destroy();
+      scene.light = originalLight;
     });
 
     function createCascadedShadowMap() {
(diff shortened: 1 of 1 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Specs/Scene/ShadowMapSpec.js
```diff
@@ -825,7 +865,9 @@ describe(
     });
 
     // TODO: Need to debug this test
```
j9liu, line 867: Delete this comment

Specs/Scene/ShadowMapSpec.js
```diff
@@ -797,7 +827,17 @@ describe(
       scene.light = new DirectionalLight({
         direction: lightDirectionAngle,
       });
-      renderAndExpect(unshadowedColor);
+
+      // Change the time so that the shadows are no longer pointing straight down
+      renderAndCall(function (rgba) {
+        // Due to shading differences, the result might not exactly the same, but
```
j9liu, line 833: What does "shading differences" refer to here -- differences between new `Model` and old `Model`?

Specs/Scene/ShadowMapSpec.js
```diff
@@ -797,7 +827,17 @@ describe(
...
+        // it should be close
+        const [r, g, b, a] = colorAbsoluteDifference(rgba, unshadowedColor);
+        expect(r).toBeLessThan(48);
+        expect(g).toBeLessThan(4);
```
j9liu, line 837: The epsilons seem arbitrary to me since they differ so much for each test... is there no way to make them more consistent? Why do we consider a color difference of 48 or even 64 "close" ?

Specs/Scene/ShadowMapSpec.js
```diff
@@ -797,7 +827,17 @@ describe(
       scene.light = new DirectionalLight({
         direction: lightDirectionAngle,
       });
-      renderAndExpect(unshadowedColor);
+
+      // Change the time so that the shadows are no longer pointing straight down
```
j9liu, line 831: Delete this comment since the time isn't being changed

j9liu (commented):
@ptrgags I left a review -- I feel iffy about the approach because the difference in the expected colors seems wildly different. Maybe we can change something about the test setup so the expectations aren't as arbitrary.

ptrgags:
@j9liu I updated those tests to be a bit more robust: 

- I changed the camera setup so it's pointing straight down above the floor to ensure it sees the shadow. 
- I noticed that one test didn't pass `startTime` to the render call, which explains why I was getting different results depending on when I was running it yesterday.
- I changed how the colors are checked to rely less on arbitrary thresholds.

Specs/Scene/ShadowMapSpec.js
```diff
@@ -818,25 +818,37 @@ describe(
...
+        expect(shadowedRed).toBeLessThan(unshadowedRed);
       });
 
       // Change the light so that the shadows are no longer pointing straight down
       scene.light = new DirectionalLight({
         direction: lightDirectionAngle,
       });
 
-      // Change the time so that the shadows are no longer pointing straight down
+      // Change the light direction so that the shadows are no longer pointing straight down
```
j9liu, line 839: this is now a duplicate of the comment above :')

Specs/Scene/ShadowMapSpec.js
```diff
@@ -788,16 +818,38 @@ describe(
...
+        expect(rgba).not.toEqual(backgroundColor);
+        expect(rgba).not.toEqual(shadowedColor);
+        expect(rgba).not.toEqual(unshadowedColor);
+
+        // After changing the light direction, the floor will appear
+        // a bit darker, but not as dark as when it was shadowed by the box
+        const red = rgba[0];
+        const unshadowedRed = unshadowedColor[0];
+        const shadowedRed = shadowedColor[0];
+        expect(shadowedRed).toBeLessThan(red);
```
j9liu, line 850: same suggestion as above

```suggestion
        expect(red).toBeGreaterThan(shadowedRed);
```

Specs/Scene/ShadowMapSpec.js
```diff
@@ -732,23 +743,42 @@ describe(
...
+        expect(rgba).not.toEqual(backgroundColor);
+        expect(rgba).not.toEqual(shadowedColor);
+        expect(rgba).not.toEqual(unshadowedColor);
+
+        // After changing the sunlight direction, the floor will appear
+        // a bit darker, but not as dark as when it was shadowed by the box
+        const red = rgba[0];
+        const unshadowedRed = unshadowedColor[0];
+        const shadowedRed = shadowedColor[0];
+        expect(shadowedRed).toBeLessThan(red);
```
j9liu, line 774: this may seem nitpicky, but I think this reads better:

```suggestion
        expect(red).toBeGreaterThan(shadowedRed);
```

j9liu (commented):
@ptrgags This looks a lot better :) just a couple tweaks

ptrgags:
@j9liu updated!

j9liu:
@ptrgags looks good, thanks!

PR #10703, opened 2022-08-18
Add support for point cloud back face culling

### LINKED ISSUE #10632: Support point cloud back face culling in `ModelExperimental`
the old `PointCloud` supported back face culling when `pointCloudShading.backFaceCulling = true` and the point cloud had per-point normals. `ModelExperimental` should be updated to include this feature. this involves:

1. Adding some shader code to the fragment shader to discard points that are back facing (using (`normalEC`)
2. Since the `PointCloudStylingStage` handles the `PointCloudShading` object, it should check the `pointCloudShading.backFaceCulling` flag and set a define
3. if this setting needs to be toggleable at runtime, `ModelExperimental` would need to check for changes and rebuild the shader if so

ptrgags:
Fixes #10632 

This PR adds support for `pointCloudShading.backFaceCulling` which culls back-facing points for point clouds that have normals.

[Sandcastle example](http://localhost:8080/Apps/Sandcastle/index.html#c=bZFfa8IwFMW/SuhTBUkYe5tVhpU9DTZQ9pSXmFw1M80tzW1Fx777krYbbrN9aP6cc/idW40+EOssnKBhc+bhxEoItq34W3+Wy0z3+xI9KeuhkdlkJn16de8l6yAARfOQwoMGD7xubGXJdhC4Mia/Ch4+96vNYMw/pGesbdwDk5lY16CDWClS4pdOvKL1VDpszdVyqUgfwIiRgb8H9DKbSv85GSBHpAtitcF8lA1Xa+WNVoEcJMAN7vcOli0R+th5qfSRPSkNrGyds34fU3/6HZTBU5iyXes1WfQsTw0iiD6CkX7C+kbfTPUP7ToaYxTfxvCUPUbHyY3eWQKfZdOsCHR2sEgx6Xm0VY0NpSHlnAuCqnaK4lC2bbQR1yGkTklaiGtrYWzHrJnf+IlMOxVCvNlFirW9gMwWhYj6f1aHPfZLB41T5yQ73C2eh0POeSHi9raTEN1WNX+SvwA)
2022-08-18_PointsBackFaceCulling

### DIFF
--- a/Source/Scene/Model/PointCloudStylingPipelineStage.js
+++ b/Source/Scene/Model/PointCloudStylingPipelineStage.js
@@ -117,7 +117,7 @@ PointCloudStylingPipelineStage.process = function (
   }
 
   const pointCloudShading = model.pointCloudShading;
-  if (model._attenuation) {
+  if (pointCloudShading.attenuation) {
     shaderBuilder.addDefine(
       "HAS_POINT_CLOUD_ATTENUATION",
       undefined,
--- a/Source/Scene/Model/PointCloudStylingPipelineStage.js
+++ b/Source/Scene/Model/PointCloudStylingPipelineStage.js
@@ -125,6 +125,14 @@ PointCloudStylingPipelineStage.process = function (
     );
   }
 
+  if (pointCloudShading.backFaceCulling) {
+    shaderBuilder.addDefine(
+      "HAS_POINT_CLOUD_BACK_FACE_CULLING",
+      undefined,
+      ShaderDestination.VERTEX
+    );
+  }
+
   let content;
   let is3DTiles;
   let usesAddRefinement;
--- a/Specs/Scene/Model/MaterialPipelineStageSpec.js
+++ b/Specs/Scene/Model/MaterialPipelineStageSpec.js
@@ -138,62 +138,10 @@ describe(
       });
     });
 
-    it("adds material uniforms", function () {
+    it("adds material and metallic roughness uniforms", function () {
       return loadGltf(boomBox).then(function (gltfLoader) {
         const components = gltfLoader.components;
         const primitive = components.nodes[0].primitives[0];
-        const renderResources = mockRenderResources();
-        const shaderBuilder = renderResources.shaderBuilder;
-        const uniformMap = renderResources.uniformMap;
-
-        MaterialPipelineStage.process(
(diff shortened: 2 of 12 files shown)


ptrgags:
@j9liu this PR is now ready for review!

Source/Scene/Model/PointCloudStylingPipelineStage.js
```diff
@@ -125,6 +125,14 @@ PointCloudStylingPipelineStage.process = function (
     );
   }
 
+  if (pointCloudShading.backFaceCulling) {
```
j9liu, line 128: The `if` statement above this is `if(model._attenuation)`, which also comes from `pointCloudShading`. Can we make the `if` statements match each other so it's clear that their properties come from the same source? Doesn't matter which one you change, just be consistent

Specs/Scene/Model/Model3DTileContentSpec.js
```diff
@@ -1014,55 +1015,34 @@ describe(
         });
       });
 
-      // This will be added in a separate PR
-      xit("Supports back face culling when there are per-point normals", function () {
+      it("Supports back face culling when there are per-point normals", function () {
+        // Since this test relies on picking, it will not work properly with webglStub
+        if (webglStub) {
+          return;
+        }
```
j9liu, line 1022: sorry for nitpick, can we have a line break here :')

Specs/Scene/Model/MaterialPipelineStageSpec.js
```diff
@@ -19,6 +19,7 @@ import {
   ShaderBuilder,
 } from "../../../Source/Cesium.js";
 import createScene from "../../createScene.js";
+import ShaderBuilderTester from "../../ShaderBuilderTester.js";
```
j9liu, line 22: `"adds material uniforms"` and `"adds metallic roughness uniforms"` test basically the same thing: the only thing that's different are the `expectedUniforms`, which are subsets of the larger uniform map. let's combine them so they're not redundant

j9liu (commented):
@ptrgags just a few comments, otherwise this looks good

ptrgags:
@j9liu updated!

j9liu:
@ptrgags looks good, thanks!

PR #10763, opened 2022-09-06
Fix specs when log depth is false

ptrgags:
In some internal code outside this repo, we turn off the logarithmic depth buffer. But when the specs are run in such an environment, some specs fail due to models clipped by the near plane.

The fix is usually to move the camera back just a little bit so the spec works whether the depth buffer is on or off.

@ggetz could you review?

### DIFF
--- a/Specs/Scene/Model/ModelSpec.js
+++ b/Specs/Scene/Model/ModelSpec.js
@@ -641,12 +641,19 @@ describe(
     });
 
     it("renders model with the KHR_materials_pbrSpecularGlossiness extension", function () {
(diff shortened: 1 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-log-depth-failures/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
All looks good. Thanks for brining this in @ptrgags!

PR #10770, opened 2022-09-08
Fix scene crash caused by not checking if gltf.nodes exist before accessing it. 

### LINKED ISSUE #10769: Scene crash when GLTF has no `nodes` attribute
When loading Cesium3DTileset tile that has empty nodes list (empty in this case means that no `nodes` attribute exist in corresponding GLTFs), scene crashes with error:

image

This error occurs because `gltf.nodes.length` is accessed without checking if it exists, which leads to scene crash when it does not exist.

As attribute  `nodes`  is [not required](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_gltf_nodes) according to GLTF 2.0. specification, the GTLF models without this attribute considered valid and should not lead to scene crash. 

The fix is quite simple, it is required to check if `gltf.nodes` exist in the `loadNodes()` method of the `GltfLoader()` :

```Javascript
if (!defined(gltf.nodes)) {
    return [];
 }
```

https://github.com/CesiumGS/cesium/blob/ba2835b96d46274b4c87a1e094eb8e521ac814ac/Source/Scene/GltfLoader.js#L2133

ilyaly:
Fixes #10769.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,10 +6,6 @@
 
 - Added support for the `WEB3D_quantized_attributes` extension found in some glTF 1.0 models.
 
-##### Fixes :wrench:
-
-- Fixed scene crash when loading Cesium3DTilest with tiles containing no `gltf.nodes` attribute. [#10769](https://github.com/CesiumGS/cesium/issues/10769)
-
 ### 1.97 - 2022-09-01
 
 #### Major Announcements :loudspeaker:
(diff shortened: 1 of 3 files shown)


ilyaly (commented):
Removing change note from CHANGES.md

ilyaly:
Ready for review!

lilleyse:
Thanks @ilyaly

PR #10778, opened 2022-09-12
Make private constructors more obvious in the JSDoc output

ptrgags:
This PR does the following:

1. The original intent is to document a breaking change where the `Model` constructor becomes private
  * The old Model _usually_ used `.fromGltf()` everywhere (including all the Sandcastle examples and the code for model entities), but only now did I notice the line in the original documentation that said the constructor can be used directly in some corner cases.
  * The new Model _must_ use `.fromGltf()`/`.fromB3dm()`/etc. since it requires a resource loader (`GltfLoader`/`B3dmLoader`/etc.) but these are all private classes.
  * There wasn't an obvious way to document this with the existing JSDoc tags, so we wanted to change the template to make the internal constructor and correct usage more obvious.
2. Defines a `@privateParam` tag that makes it possible to document parameters in the code, but hide them from the JSDoc output.
3. In the CSS, I defined a `div.notice` class that creates a yellow background for making a line of text stand out. This is used to make the correct usage more obvious
4. For the `@internalConstructor` tag, I added an explicit

The end result looks something like this:

| before | after |
|--------|-------|
| 2022-09-12_ModelConstructorOld | 2022-09-12_ModelConstructorNew |

There were quite a few other cases in the code that used `@internalConstructor`, so I gave them similar treatment.

To Do:

- [x] Update CHANGES.md


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/jsdoc-private-constructor/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
This PR is now ready for review. @ggetz could you review when you get a chance?

CHANGES.md
```diff
@@ -2,9 +2,14 @@
...
 
+#### Breaking Changes :mega:
+
+- `new Model()` is an internal constructor and must not be used directly. Use `Model.fromGltf()` instead. [#10778](https://github.com/CesiumGS/cesium/pull/10778)
+
 ##### Additions :tada:
 
-- Added support for the `WEB3D_quantized_attributes` extension found in some glTF 1.0 models.
+- Added support for the `WEB3D_quantized_attributes` extension found in some glTF 1.0 models. [#10758](https://github.com/CesiumGS/cesium/pull/10758)
+- Update JSDoc template to mark internal constructors more clearly. [#10778](https://github.com/CesiumGS/cesium/pull/10778)
```
ggetz, line 12: I don't think you'll need this bullet, as the public API did not change, just how they are formatted in the docs.

ggetz (commented):
To clarify, the documentation formatting changes in this PR are due to the fact that using `@internalConstructor`, while valid JSDoc, does not actually hide the constructor or otherwise indicate that it is not part of the public API. Correct?

How does `internalConstructor` get used in generating the TypeScript definitions? Is it properly marked it as private in the output?

ptrgags:
@ggetz The new definition for `@internalConstructor` does the same steps as the `@class` tag. It calls this [onTagged callback](https://github.com/jsdoc/jsdoc/blob/main/packages/jsdoc/lib/jsdoc/tag/dictionary/definitions.js#L319-L343), so it has `kind = class`. The only two differences are:
 * The tag name is different
 * `@internalConstructor` sets a flag `isInternalConstructor = true` so the template knows when to add the extra formatting.

So my understanding is that anything that looks for `kind === class` (our JSDocTemplate) will find the constructor, but anything that's expecting `@class` or `@constructor` will not find it since it's not explicitly a synonym, and thus it'll get skipped.
 
As for how does the `ts-doc` react, I see that now the parameters are not added to the constructor anymore. Is this what we want?

image
image

ggetz:
> As for how does the ts-doc react, I see that now the parameters are not added to the constructor anymore (probably has more to do with @privateParam than the @internalConstructor) Is this what we want?

This may be a matter for a separate PR, but I would expect we want to mark the constructor as private in the TypeScript definitions.

ptrgags:
@ggetz updated!

ggetz:
Thanks @ptrgags! I verified the doc output is all as expected.

PR #10780, opened 2022-09-12
Slack notifications

ggetz:
Our team spoke offline and determined it’s too easy for a CI failure in `main` to go unnoticed if the merged branch CI was passing before the merge. So we’ve added a [Travis Slack Notification](https://docs.travis-ci.com/user/notifications/)  to notify us here when it fails a build in `main`. We don’t expect it to be too noisy given `main` _should_ always be passing

This uses the default notification frequency, which is always notify on failure and notify on success when the value changes back.

Here's the result which I had sent to a testing channel:
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ node_js:
 addons:
   chrome: stable
   firefox: latest
+notifications:
+  slack: cesium:$SLACK_TOKEN
 script:
   - npm --silent run deploy-set-version -- --buildVersion $TRAVIS_BRANCH.$TRAVIS_BUILD_NUMBER
   - npm --silent run deploy-status -- --status pending --message 'Waiting for build'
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/slack-notifications/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

.travis.yml
```diff
@@ -4,6 +4,12 @@ node_js:
 addons:
   chrome: stable
   firefox: latest
+notifications:
+  slack:
+    secure: JKzk2sJSbZ9h2PUVWj6KtOAdFbEEnOtv/VZy05pJ2H41xRgUHiGdtMW/vMSeq6XX3IJN8eW2zd0cJTgkFn0ioAlYvID8zRhcvkFHg60QXquoqtp5y65dxjtVz79hefxSo7FO1NhMZBQWE9Tg6R7XkoyTMth62+T9vqOgu2Hms6M=
+    if: branch = main
+    on_success: never # default: change
```
sanjeetsuhag, line 11: > ... notify on success when the value changes back.

With `on_success` set to `never`, will this happen?

ggetz: Apologies, non-email integration appear to have a different default from emails and I was testing something out. This has been updated to be inline with the PR description.

sanjeetsuhag (approved):
Based on the documentation, I think the changes look good. Thanks @ggetz, this will be helpful in catching failures early.

PR #10785, opened 2022-09-13
Use more concise syntax for PipelineStage constants

jjhembd:
As suggested by @ptrgags in https://github.com/CesiumGS/cesium/pull/10683#discussion_r968563051, this PR updates the declaration of PipelineStage constants to use a more concise syntax. 

The new syntax is consistent with what we already have in MetadataPipelineStage.

### DIFF
--- a/Source/Scene/Model/AlphaPipelineStage.js
+++ b/Source/Scene/Model/AlphaPipelineStage.js
@@ -11,8 +11,9 @@ import Pass from "../../Renderer/Pass.js";
  *
  * @private
  */
-const AlphaPipelineStage = {};
-AlphaPipelineStage.name = "AlphaPipelineStage"; // Helps with debugging
+const AlphaPipelineStage = {
+  name: "AlphaPipelineStage", // Helps with debugging
+};
 
 AlphaPipelineStage.process = function (renderResources, primitive, frameState) {
   const alphaOptions = renderResources.alphaOptions;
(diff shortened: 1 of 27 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/pipeline-constants/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag:
@jjhembd or @ptrgags: What about functions that are exposed for testing by being attached to the main export? For example,

https://github.com/CesiumGS/cesium/blob/6711352fac656b7a30c2403b9980383ed6498d17/Source/Scene/Model/InstancingPipelineStage.js#L1038-L1040

Do we want to attach these at the top too, for discoverability? I think it's also ok to leave them as is, just wanted to make sure we're covering everything.

jjhembd:
> What about functions that are exposed for testing by being attached to the main export?

Thanks @sanjeetsuhag, I think it makes sense to use similar syntax for these. I just updated `InstancingPipelineStage`, along with `CustomShaderPipelineStage`, `MaterialPipelineStage`, and `PrimitiveStatisticsPipelineStage`

sanjeetsuhag:
@jjhembd The changes here look good, but there's several other `Source/Scene/Model/*PipelineStage.js` files that need to be updated to match this syntax.

sanjeetsuhag (approved):
Changes look good! I ran the `Scene` spec suite, just to make sure, and ran into no issues there. Thanks @jjhembd!

Issue #10807, opened 2022-09-22
when i3dm tile without rotation property,the scale will be incorrectory(V1.97)


MingFanZhu:
In my application,there is an i3dm tile that has only "POSITION" and "SCALE_NON_UNIFORM" property.
but as [Source/Scene/Model/I3dmLoader.js#L598](https://github.com/CesiumGS/cesium/blob/main/Source/Scene/Model/I3dmLoader.js#L598) write,`scaleTypedArray` will be replaced by `translationTypedArray`,then it will looks uncorrectory,like that:
![image](https://user-images.githubusercontent.com/44891554/191691036-23a2f2f4-2edf-4ece-b74a-00dedc08d82b.png)
If change `translationTypedArray` to `scaleTypedArray`,it looks correctory:
![image](https://user-images.githubusercontent.com/44891554/191691663-24446ff3-959b-4d2d-bf87-3e34e65647ce.png)
So,whether should i modified like this?

PR #10808, opened 2022-09-22
Fix instance scales

lilleyse:
Fixes https://github.com/CesiumGS/cesium/issues/10807

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 ##### Fixes :wrench:
 
 - Fixed a bug where instanced models without normals would not render. [#10765](https://github.com/CesiumGS/cesium/pull/10765)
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-scale/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag (approved):
Looks good to me. Both `InstancedScale` and `InstancedScaleNonUniform` sample tilesets also have `rotation` so I can't test this directly, but looking at the example in the issue, I think this should suffice. Thanks @lilleyse!

PR #10813, opened 2022-09-23
Filter unused runtime nodes

ptrgags:
This PR fixes a corner case where unused glTF nodes would cause `Model` to crash at runtime.

`ModelSceneGraph` keeps an array of runtime nodes that is the same length as the list of nodes in the glTF. However, unused nodes are `undefined` entries in this array. The loops in `configurePipeline` and `update` weren't filtering out this corner case.

This bug was first reported in [this forum post](https://community.cesium.com/t/3d-tiles-crashes-with-cesium-1-97/20464), see the thread for an example tileset.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/filter-unused-nodes/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
This PR is now ready, @sanjeetsuhag could you review?

sanjeetsuhag (approved):
Looks good @ptrgags. I tested it by adding empty nodes to the `BoxInstanced.gltf` file and it fixed the crash. From the screenshot posted on the community forum, seems like it was fixed there too.

PR #10822, opened 2022-09-26
Voxels uniform fix

jjhembd:
Some Firefox users reported crashes when loading more than one `VoxelPrimitive`.

`VoxelPrimitive.prototype.update` was throwing an error if a uniform from a new primitive was already defined. This PR converts the error to a `oneTimeWarning`.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
As a temporary workaround this fine, but the DeveloperError is likely there for a reason and it would be good to know why it gets triggered, and how we can avoid that from happening.

lilleyse:
Merging. @jjhembd could you open an issue for https://github.com/CesiumGS/cesium/pull/10822#issuecomment-1258621669

PR #10823, opened 2022-09-26
Fix attribute name conflict in Cesium3DTilesVoxelProvider

jjhembd:
`Cesium3DTilesVoxelProvider.prototype.requestData` merges information from two sources:
- attribute buffers from GltfLoader
- schema information from the tileset, stored in the provider constructor

Much of this information is in array format, and arrays from the two sources may be ordered differently.

This PR compares the attribute name to ensure the merged information is correctly aligned.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Issue #10825, opened 2022-09-27
3d tiles can not display in two viewer at the same time


eclosion:
I create two viewer, they add the same 3d tiles, but the models only display in one of them

![image](https://user-images.githubusercontent.com/13535652/192514941-45724394-8a71-4fbb-bec1-9c7f2d2f42db.png)

Cesium: 1.97

Sandcastle example:
[...]

PR #10828, opened 2022-09-27
Handle model GPU resources across different WebGL contexts

lilleyse:
Fixes https://github.com/CesiumGS/cesium/issues/10825

Adds `context.id` to the cache key for GPU resources like vertex buffers, index buffers, and textures. This prevents the GPU resource from being shared by multiple contexts which is the root cause of https://github.com/CesiumGS/cesium/issues/10825.

Selection_062


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Issue #10636, opened 2022-08-04
Publish smaller packages


ggetz:
CesiumJS is an extensive library with a lot of functionality that, depending on the app or use case, goes unused. Installing the `cesium` npm package also syncs third party libraries which may go unused. All features are tied to the monthly Cesium versioning, whether or not anything has changed.

Many users should continue to use the "traditional" CesiumJS releases, but there are also many use cases that would benefit from ingesting smaller, more specific packages. 

See [this community forum thread](https://community.cesium.com/t/is-there-interest-in-a-compact-cesium-math-library/18366) for some seeds of this discussion.

- Maintain the traditional combined release as before. There may be some internal refactoring, but from an end user standpoint, the `cesium` npm package and release would remain as-is
- Maintain all code from within the same repo. We don't want to introduce additional development burden with additional repos or submodules
	- [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) are built into npm for use cases such as this. It allows dependency management to work similar to before with all modules installed to one top-level `node_modules` directory
	- All packages would continue to share the same tooling
- Publish individual packages along architectural boundaries. In theory, there should be nothing stopping us from breaking up and publishing code as-is with some additional configuration, however due to some interconnections in the architecture, this may require some refactoring so that we are not "cheating" by using private APIs.
	- Each new package would be able to be published individually
	- Each new package would only require relevant dependencies-- This means dependencies would only be synced if they are strictly needed, rather than the approach we take now where all dependencies are installed regardless
	- Each new package would follow semver, while traditional CesiumJS release could remain with its monthly versioning strategy
- Most major JS projects implement a variation of this (React, Babel, BabylonJS, Angular…)
- While not needed for an initial implementation, eventually, this could allow us take advantage of tooling for dependency graphs and more, allowing us to do things like run only the unit tests relevant to a change
	- [lerna](https://github.com/lerna/lerna) is a tool [recommended by npm](https://blog.npmjs.org/post/186494959890/monorepos-and-npm.html) for managing monorepos, also [nx](https://github.com/nrwl/nx) for more advanced use cases
	
A few use cases:
- Widgets
    - `CesiumViewer` and other Widgets work well for simple use cases that want to get something really basic up quickly. But a more sophisticated app does not need them, and they can actively be a hindrance as they bring in many static assets.
[...]

PR #6331, opened 2018-03-14
HTML Credits

hpinkos:
Other web mapping platforms like Leaflet and OpenLayers define attributions as HTML strings.  This changes our `Credit` to do the same.  HTML allows for much more flexibility than only exposing options for setting the text, image source and URL link.  For example, will allow users to create credits like 
"My source is from [some website](http://invalid.link) with [some license](http://invalid.link)", something that is frequently needed for proper attribution, but not currently supported with our `Credit`.

Summary of changes:
- Modified `Credit` to accept an html string instead of an `option` object
- Added [xxs.js](https://github.com/leizongmin/js-xss) library for sanitizing the HTML before creating a DOM element from it
- Since `Credit` no longer differentiates between 'text' credits and 'image' credits, modified `CreditDisplay` to add all credits to the same credit container
- To keep the Cesium credit first, created a separate `CreditDisplay._cesiumCreditContainer` to add the Cesium credit

PR #10845, opened 2022-10-07
Changes Viewer to CesiumWidget in KmlTour

sanjeetsuhag:
This PR deprecates the `Viewer` parameter in favor of a `CesiumWidget` in the`KmlTour.prototype.play`function. Since the `CesiumWidget` will be a part of the `engine` package, we don't want to have to depend on anything from the `widgets` package (which `CesiumViewer` is intended to be a part of).


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/kml-tour-viewer/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

Source/DataSources/KmlTour.js
```diff
@@ -81,14 +82,42 @@ KmlTour.prototype.addPlaylistEntry = function (entry) {
...
  *
  * @param {Viewer} viewer viewer widget.
  * @param {Object} [cameraOptions] these options will be merged with {@link Camera#flyTo}
  * options for FlyTo playlist entries.
+ *
+ * @deprecated
  */
 KmlTour.prototype.play = function (viewer, cameraOptions) {
+  deprecationWarning(
+    "KmlTour.prototype.play",
```
ggetz, line 117: What's the motivation to deprecate this whole function? It's possible to deprecate just first parameter instead. [For example](https://github.com/CesiumGS/cesium/pull/6331/files#diff-d3495d81898f88370dc3635be83a0f6d87b7acf98c676debde79c59b852e907dR41).

Source/DataSources/KmlTour.js
```diff
@@ -81,14 +82,42 @@ KmlTour.prototype.addPlaylistEntry = function (entry) {
...
 /**
  * Play this tour.
  *
  * @param {Viewer} viewer viewer widget.
  * @param {Object} [cameraOptions] these options will be merged with {@link Camera#flyTo}
  * options for FlyTo playlist entries.
+ *
+ * @deprecated
  */
 KmlTour.prototype.play = function (viewer, cameraOptions) {
```
ggetz, line 115: The Sandcastle `KML Tours.html` will also need to be updated to use the new API.

Source/DataSources/KmlTour.js
sanjeetsuhag: If we wanted to do just a parameter deprecation, we'd have to do an `instanceof` check against `Viewer` to print the `deprecationWarning`, right? Then we'd have to import `Viewer`, `which will cause issues in the workspaces PR.

ggetz: Good point. `instanceof` is generally what we'd use since it's a more cut and dry check, but in this case I agree we want to avoid importing the class. We could check for the presence of a `Viewer` property that is not present on an instance `CesiumWidget`, say `.cesiumWidget`?

sanjeetsuhag: Yeah, that's a good idea! To ensure that we don't break TypeScript, I changed the types to `{Viewer|CesiumWidget}`. I'll create an issue to change that in 1.100. Then, in the workspaces PR, I can just add the following to make TS compile:

```typescript
/**
 * @typedef {Viewer} Viewer
 * @property {Scene} scene The scene in the widget.
 */
```

Source/DataSources/KmlTour.js
sanjeetsuhag: Updated.

sanjeetsuhag:
@ggetz Updated.

ggetz:
@mramato Would you be able to take a pass on this? This came up because the package separation in https://github.com/CesiumGS/cesium/issues/10636.

ggetz:
Thanks @sanjeetsuhag!

PR #10866, opened 2022-10-19
I3S: enhancement to feature data loading

Tamrat-B:
- Added a new function, getFieldsForFeature on I3SNode
- Removed _loadFeatureData call in _createContentURL when the I3S version is 1.7 or higher (obsolete)
- Fixed an issue loading some newer I3S datasets where the feature data uri was different than “feature/0”


cesium-concierge:
Thanks for the pull request @Tamrat-B!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/i3s_support/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #10873, opened 2022-10-26
Add `willReadFrequently` option when getContext("2d")

onsummer:
Resolved a warning on the last few versions of the browser:

image

### DIFF
--- a/Apps/HelloWorld.html
+++ b/Apps/HelloWorld.html
@@ -11,7 +11,7 @@
       content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"
     />
     <title>Hello World!</title>
-    <script src="../Build/Cesium.js"></script>
+    <script src="../Build/CesiumUnminified/Cesium.js"></script>
     <style>
       @import url(../Build/CesiumUnminified/Widgets/widgets.css);
       html,
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -527,6 +527,10 @@ async function pruneScriptsForZip(packageJsonPath) {
   delete scripts["deploy-set-version"];
   delete scripts["website-release"];
 
+  // Set server tasks to use production flag
+  scripts["start"] = "node server.js --production";
+  scripts["start-public"] = "node server.js --public --production";
+
   // Write to a temporary package.json file.
   const noPreparePackageJson = join(
     dirname(packageJsonPath),
@@ -571,6 +575,7 @@ export const makeZip = gulp.series(release, async function () {
       "Build/Cesium/**",
       "Build/CesiumUnminified/**",
       "Build/Documentation/**",
+      "Build/Specs/**",
       "Build/package.json",
       "packages/engine/Build/**",
       "packages/widgets/Build/**",
--- a/server.js
+++ b/server.js
@@ -28,6 +28,10 @@ const argv = yargs(process.argv)
       description:
         'A comma separated list of hosts that will bypass the specified upstream_proxy, e.g. "lanhost1,lanhost2"',
     },
+    production: {
+      type: "boolean",
+      description: "If true, skip build step and serve existing built files.",
+    },
   })
   .help().argv;
 
@@ -144,7 +148,15 @@ const serveResult = (result, fileName, res, next) => {
 
 (async function () {
   const gzipHeader = Buffer.from("1F8B08", "hex");
-  let { esmResult, iifeResult, specResult } = await generateDevelopmentBuild();
+  let esmResult, iifeResult, specResult;
+  const production = argv.production;
+
+  if (!production) {
+    const bundles = await generateDevelopmentBuild();
+    esmResult = bundles.esmResult;
+    iifeResult = bundles.iifeResult;
+    specResult = bundles.specResult;
+  }
 
   // eventually this mime type configuration will need to change
   // https://github.com/visionmedia/send/commit/d2cb54658ce65948b0ed6e5fb5de69d022bef941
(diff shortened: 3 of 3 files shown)


cesium-concierge:
Thanks for the pull request @onsummer!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Hi @onsummer, thanks for the PR!

We had some previous discussion around this warning on the forum in these two threads:
* https://community.cesium.com/t/canvas2d-multiple-readback-operations-using-getimagedata-are-faster-with-the-willreadfrequently-attribute-set-to-true/20703
* https://community.cesium.com/t/how-to-eliminate-cesium-warning-canvas2d-multiple-readback-operations-using-getimagedata-are-faster-with-the-willreadfrequently-attribute-set-to-true/20727

Basically, we should document performance with and without this change to ensure that it's the right one. Would you be able to post some performance tests results in this PR?

@lilleyse Do you have any additional thoughts?

onsummer:
@ggetz Busy for working recently and do not notice those discussion, I do not know how to estimate the performance with and without this option, can u please give me some tips? Thanks.

onsummer:
@ggetz Got a blog from cesium.com - [Entity Api Performance](https://cesium.com/blog/2018/06/21/entity-api-performance/), Can I test the performance of a large number of LabelEntity using the method described in this blog?

ggetz:
@onsummer Yes, that should work! Getting the average FPS will also be helpful:

```
```

onsummer:
> @onsummer Yes, that should work! Getting the average FPS will also be helpful:
> 
> ```
> viewer.scene.debugShowFramesPerSecond = true;
> ```

OKay, I will test later

onsummer:
@ggetz I use that code and reduce label's count to 1500:

```js

const randomCoordinateJitter = (degree, margin) => {
  return degree + margin * (Math.random() - 0.5) / 0.5;
};
const nCats = 1500;
for (let i = 0; i <= nCats; i++) {
  viewer.entities.add({
    position: Cesium.Cartesian3.fromDegrees(
      randomCoordinateJitter(-77, 0.5),
      randomCoordinateJitter(38.88, 0.5),
    ),
    label: {
      text: `CUTE #${i}`,
    },
    billboard: {
      height: 24,
      width: 36,
      image: 'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRhFpaxDXCS5O9hx90F3ufJI2VnC_wW0lPnrr6BIb18P4V5JXxBCg',
    },
    ellipse: {
      semiMajorAxis: 100,
      semiMinorAxis: 80,
    }
  });
}
```

Regardless of whether `willReadFrequently` option is used, when scene rendering stabilizes, I observe that frame rate is roughly around 39~41 fps on my machine. (R7 4800U with 16GB RAM, not using discrete gpu)

Here are the testing profile file:

[Profile-WithOrWithoutFrequentlyOption.zip](https://github.com/CesiumGS/cesium/files/9872367/Profile-WithFrequentlyOption.zip)

with option

without option

Combining the above statistical data, I thought about it, `willReadFrequently` may be helpful for those 2d canvas read pixel operations frequently, but the 2 performance tests seem that maybe there is no (or almost no) 2d canvas read-pixels operation? I'm not sure if my assumption is correct.

ctr1key:
Adding some helpful information to consider should it help this effort.
There is a [link](https://github.com/whatwg/html/issues/5614#issuecomment-866944763) to a document written by the Chrome team about the history of needing willReadFrequently and its performance impact.
From what I gather:

1.  It should not be used on canvases where you never expect to read data from the canvas with getImageData()/toDataUrl()/toBlob(). Using {willReadFrequently: true} will disable hardware acceleration for that canvas.

2. Without using willReadFrequently, it will use GPU rendering for draw commands before the first call to getImageData() and CPU rendering for subsequent draw commands. 
The impact to performance of getImageData() when using {willReadFrequently: true} vs {willReadFrequently: undefined} is only seen during the first call on a canvas.
After the first call to getImageData(), the performance of getImageData() is the same whether you use {willReadFrequently: true} or {willReadFrequently: undefined}.
Using {willReadFrequently: false} will incurred the performance hit of getImageData() for every call.
3. The performance gain of using getImageData() without hardware acceleration {willReadFrequently: true} is roughly equivalent to performance loss of doing approximately 30000 draw commands with CPU rendering instead of GPU rendering.
Using {willReadFrequently: false} will prevent the fallback to CPU rendering in a case of heaving drawing and infrequent calls to getImageData() can improve performance.
5. It's not expected to see a difference in FPS if you are making less than 4 getImageData() calls in a single frame when the canvas is hardware accelerated.  At one call per frame getImageData(), it may be worse to not use {willReadFrequently: false} as you give up other benefits of hardware acceleration.
6. You will not see a performance difference regarding the use of willReadFrequently if hardware acceleration is already disabled globally for all canvases.

ctr1key:
As for you example of 1500 static label, you may not see much performance difference because I believe only 16 total canvases are used for your entities.
1 canvas is used per character in a label. 6 canvases for 'CUTE #' and 10 canvases for the digits. I don't believe there is a canvas for the billboard.
I believe labels call getImageData() twice so 32 calls (16 slow, then 16 quick after it falls back to CPU rendering) with {willReadFrequently: undefined} vs 32 quick calls with {willReadFrequently: true} is not much difference.
It will likely be briefly noticeable if you have some way to dynamically change traits of all labels like font family, style, or weight. A change to any of those will force a new canvas and calls to getImageData().

Aside from canvases for labels and getimagepixels, I'm not sure if the other places you used {willReadFrequently: true} will even use getImageData(). It will have an adverse effect by disabling hardware acceleration with no benefit if they don't need to use getImageData().

PR #10883, opened 2022-11-01
Differentiate workspace APIs in ref docs

ggetz:
As a follow up from https://github.com/CesiumGS/cesium/pull/10824, this updates the doc generation process such that the list of classes is broken up based on the workspaces.

image

image

image

If we think it's warranted, I can also add a tag to to individual item pages. But it might be enough to list the path.

### DIFF
--- a/Tools/jsdoc/cesium_template/publish.js
+++ b/Tools/jsdoc/cesium_template/publish.js
@@ -209,7 +209,7 @@ function attachModuleSymbols(doclets, modules) {
  * @return {string} The HTML for the navigation sidebar.
  */
 function buildNav(members) {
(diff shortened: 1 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/workspaces-docs/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag (approved):
Looks great @ggetz! I think eventually we will want to include a more visible indicator of which package a class belongs to when you open its page, but that can probably wait till when we have more packages. I think this is a good start to organizing our docs.

Issue #10801, opened 2022-09-19
Wrong argument type definition in Matrix multiplyByScale functions


javagl:
The type of the `scale` parameter is annotated as `Number` in https://github.com/CesiumGS/cesium/blob/fb8158eba1ece2406556fad53fa58403b659f76f/Source/Core/Matrix3.js#L1196 and https://github.com/CesiumGS/cesium/blob/fb8158eba1ece2406556fad53fa58403b659f76f/Source/Core/Matrix2.js#L785 , although it should be `Cartesian3` and `Cartesian2`, respectively. 

This may raise the question of whether it should be `Cartesian4` (instead of `Cartesian3`) at https://github.com/CesiumGS/cesium/blob/fb8158eba1ece2406556fad53fa58403b659f76f/Source/Core/Matrix4.js#L2097 ...

javagl:
I could create a PR for that. But **iff** the type is changed from `Cartesian3` to `Cartesian4` for the `Matrix4`, then this would be a breaking change. (I think it _should_ be, for consistency, but am not sure about the trade-offs here...)

lilleyse:
`Cartesian3` is fine for `Matrix4` since scale is only expected to apply to the top-left 3x3 portion of the matrix.

jjhembd:
A little more context for non-graphics people who come across this thread:

We are using `Matrix4` to transform [homogeneous coordinates](https://en.wikipedia.org/wiki/Homogeneous_coordinates). The scaling matrix in this case is usually:

$$ {\left\lbrack \matrix{s_x & 0 & 0 & 0 \cr 0 & s_y & 0 & 0 \cr 0 & 0 & s_z & 0 \cr 0 & 0 & 0 & 1} \right\rbrack}$$

See [Yan-Bin Jia's course notes](https://faculty.sites.iastate.edu/jia/files/inline-files/homogeneous-transform.pdf), or [this interactive guide](https://wordsandbuttons.online/interactive_guide_to_homogeneous_coordinates.html).

javagl:
The same type of error - just added here to keep track of it: https://github.com/CesiumGS/cesium/blob/fb8158eba1ece2406556fad53fa58403b659f76f/Source/Core/S2Cell.js#L267 

It returns a `string`, not a `BigInt`.

Issue #10455, opened 2022-06-13
Update or remove tsd-jsdoc


ggetz:
As part of the build process, [CesiumJS generates TypeScript definitions based off our jsdoc comments](https://github.com/CesiumGS/cesium/pull/8878).

While we're generally pretty meticulous about having accurate documentation, there are a few instances of invalid jsdoc that have slipped through. The `build-ts` scripts, which converts our jsdoc to TypeScript definitions, has recently been spitting out a ton of warnings indicating probable issues with our meta tags.

There is a helpful list of common pitfalls the link above, including:

> * Use proper `@enum` notation for enums.
> * Use proper `@namespace` notation for static classes.
> * Use proper `@function` notation for standalone functions.
> * Fix Promise markup to actually include the type in all cases, i.e. `Promise` => `Promise<void>` or `Promise<Cartesian3[]>`.

This list likely covers most of the warnings. We should fix theses to ensure we have accurate documentation and accurate TYpeScript definitions.

jjhembd:
I captured the warnings in a file to evaluate. There are 1234 warnings, and 1230 of them are "Failed to find parent of doclet...". I think this is a [known issue with the tsd-jsdoc plugin](https://github.com/englercj/tsd-jsdoc/issues/133).

The plugin repo seems to be dead. I think we might be better off switching to typescript itself to generate the type definitions. Typescript [added this functionality](https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html) shortly after the tsd-jsdoc plugin was developed.

My first quick test of the typescript compiler is throwing a few errors about "requires using private name from module". This may be worth another look when we have time.

sanjeetsuhag:
[...]

PR #10899, opened 2022-11-12
Minor argument type fixes in JSDoc

javagl:
Fixes https://github.com/CesiumGS/cesium/issues/10801 (which I had kept open in a tab for ~2 months now...)

I assume that there are (potentially many) similar issues like that, with different levels of severity

- types that are plainly wrong
- types that are not found
- types that are capitalized wrong (e.g. `{Number}` instead of `{number}` - whatever is "right" there, it's certainly not _consistent_ right now - see https://jsdoc.app/tags-type.html )

In order to find them, one should use a dedicated tool (maybe one of the tools that have once been used in the first experiments for generating TypeScript bindings?). 

But these are the ones that I noticed in the issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### 1.100 - 2022-12-01
+
+##### Fixes :wrench:
+
+- Fixed the JSDoc and TypeScript definitions of arguments in `Matrix2.multiplyByScalar`, `Matrix3.multiplyByScalar`, and several functions in the `S2Cell` class. [#10899](https://github.com/CesiumGS/cesium/pull/10899)
+
 ### 1.99 - 2022-11-01
 
 #### Major Announcements :loudspeaker:
--- a/packages/engine/Source/Core/Matrix2.js
+++ b/packages/engine/Source/Core/Matrix2.js
@@ -782,7 +782,7 @@ Matrix2.multiplyByScalar = function (matrix, scalar, result) {
  * Computes the product of a matrix times a (non-uniform) scale, as if the scale were a scale matrix.
  *
  * @param {Matrix2} matrix The matrix on the left-hand side.
- * @param {Number} scale The non-uniform scale on the right-hand side.
+ * @param {Cartesian2} scale The non-uniform scale on the right-hand side.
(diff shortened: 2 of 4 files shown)


cesium-concierge:
Thank you so much for the pull request @javagl! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/javagl/cesium/blob/argument-type-fixes-2022-11-12/CONTRIBUTORS.md) file!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/argument-type-fixes-2022-11-12/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

javagl:
The `CONTRIBUTORS.md` update can be skipped for such a fix. 

Regarding the `CHANGES.md`... I'm not sure: Strictly speaking, this **is** a breaking change, namely for the generated TypeScript bindings (and that's where I noticed that issue in the first place...)

ggetz:
Thanks for the fix, @javagl!

> In order to find them, one should use a dedicated tool (maybe one of the tools that have once been used in the first experiments for generating TypeScript bindings?).

The tool we current use does issue some warnings, but unfortunately [due to an issue with the tool, the output is very noisy](https://github.com/CesiumGS/cesium/issues/10455). Happy to hear if you have alternative suggestions!

> Regarding the CHANGES.md... I'm not sure: Strictly speaking, this is a breaking change, namely for the generated TypeScript bindings (and that's where I noticed that issue in the first place...)

Yes, there should be an item for this in `CHANGES.md` for this exact reason. However, we have put previous items like this under "Fixes" rather than "Breaking Changes".

javagl:
Similar to https://github.com/CesiumGS/cesium/pull/10912#issuecomment-1321704234 , I wasn't sure about the form for the `CHANGES.md` update. I guess the approach is to just add a section with the new version (and a somewhat "arbitrary" date - this has to be checked during the release anyhow). I have updated `CHANGES.md` accordingly. The responsibility for resolving merge conflicts here will likely be determined by the merge order, but consider this PR as "low priority" compared to others.

ggetz:
Thanks @javagl!

Issue #10897, opened 2022-11-11
command error on npm run build-watch


jiangheng90:
for ```npm run build-watch``` command. both my computer in home and company got error below
```
[21:53:07] Finished 'build' after 18 s
[21:53:07] Starting '<anonymous>'...
[21:53:07] '<anonymous>' errored after 527 ms
[21:53:07] TypeError: (intermediate value) is not iterable
    at file:///E:/cesium/gulpfile.js:162:44
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
[21:53:07] 'buildWatch' errored after 18 s
```
for error throw in gulpfile, it comes from
```javascript
let [esmResult, iifeResult, cjsResult] = await bundleCesiumJs({
  minify: minify,
  path: outputDirectory,
  removePragmas: removePragmas,
  sourcemap: sourcemap,
  incremental: true,
});
```
the deconstruction throw it. I see the return of ```bundleCesiumJs``` is a object.

I have tried to fix this one with:
```javascript
const bundleResult = await bundleCesiumJs({
  minify: minify,
  path: outputDirectory,
  removePragmas: removePragmas,
  sourcemap: sourcemap,
  incremental: true,
});
let esmResult = bundleResult.esmBundle;
let iifeResult= bundleResult.iifeBundle;
let cjsResult= bundleResult.nodeBundle;
...
```

then ```bundleWorkers``` comes another ```options.input``` undefined cause ```globby``` throw error
```
[22:14:17] Finished 'build' after 23 s
[22:14:17] Starting '<anonymous>'...
[22:14:17] '<anonymous>' errored after 786 ms
[22:14:17] TypeError: Patterns must be a string or an array of strings
    at assertPatternsInput (file:///E:/cesium/node_modules/globby/index.js:15:9)
    at toPatternsArray (file:///E:/cesium/node_modules/globby/index.js:21:2)
    at file:///E:/cesium/node_modules/globby/index.js:55:66        
    at bundleWorkers (file:///E:/cesium/build.js:420:25)
    at file:///E:/cesium/gulpfile.js:177:9
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
[22:14:17] 'buildWatch' errored after 23 s
```

then I stop on this error, recently I wiil keep track on this

ggetz:
Hi there, thanks for the report. What version of node are you using?

For example:

```sh
[...]

PR #10901, opened 2022-11-14
Fixes incremental builds and watch scripts

sanjeetsuhag:
Fixes https://github.com/CesiumGS/cesium/issues/10897

- Fixes an error in the `build-watch` command
- Fixes the incremental building of bundles/workers/shaders
- Reduces the noise in build output to only include the JS bundles that are generated

### DIFF
--- a/build.js
+++ b/build.js
@@ -119,7 +119,6 @@ export const defaultESBuildOptions = () => {
     bundle: true,
     color: true,
     legalComments: `inline`,
-    logLevel: `info`,
     logLimit: 0,
     target: `es2020`,
   };
@@ -177,7 +176,7 @@ export async function bundleCesiumJs(options) {
     js: combinedCopyrightHeader,
   };
   // print errors immediately, and collect warnings so we can filter out known ones
-  buildConfig.logLevel = "error";
+  buildConfig.logLevel = "info";
 
   const bundles = {};
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -35,13 +35,12 @@ import {
   buildCesium,
   buildEngine,
   buildWidgets,
-  bundleCesiumJs,
   bundleWorkers,
   glslToJavaScript,
   createSpecList,
-  bundleCombinedSpecs,
   createJsHintOptions,
   defaultESBuildOptions,
+  bundleCombinedWorkers,
 } from "./build.js";
 
 // Determines the scope of the workspace packages. If the scope is set to cesium, the workspaces should be @cesium/engine.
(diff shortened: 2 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/watched-files-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @sanjeetsuhag, this does fix the error (confirmed in NodeJS 14, 16, 18).

Perhaps out of the scope of this fix, but since running `npm start` will override most of the output, is keeping `build-watch` around (and maintaining the code going forward) worth it?

For instance @jiangheng90, why do you use `npm run build-watch` (which rebuilds on file change) instead of the output from `npm start` (which rebuilds when the relevant files are requested in the browser)?

sanjeetsuhag:
> Perhaps out of the scope of this fix, but since running npm start will override most of the output, is keeping build-watch around (and maintaining the code going forward) worth it?

I was wondering that too. Would it be useful in a NodeJS context?

package.json
```diff
@@ -150,7 +150,7 @@
     "prettier-check": "prettier --check \"**/*\""
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
```
ggetz, line 153: [Double checking this now](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines), I believe NodeJS is required for _development_ on this package (running the build and other dev commands with the `--workspaces` flag) and does belong in the build guide, but not for apps which depend on this package. I think we can safely restore this to `14` for those installing the package on npm.

sanjeetsuhag: Updated.

jiangheng90:
> Thanks @sanjeetsuhag, this does fix the error (confirmed in NodeJS 14, 16, 18).
> 
> Perhaps out of the scope of this fix, but since running `npm start` will override most of the output, is keeping `build-watch` around (and maintaining the code going forward) worth it?
> 
> For instance @jiangheng90, why do you use `npm run build-watch` (which rebuilds on file change) instead of the output from `npm start` (which rebuilds when the relevant files are requested in the browser)?

@ggetz I use `npm run build-watch` is just coincident, At that time I found workspace is diffrent, I can not make src code changes work at sandcastle and just find `npm run build-watch` can rebuild and use it

jiangheng90:
> Thanks @sanjeetsuhag, this does fix the error (confirmed in NodeJS 14, 16, 18).
> 
> Perhaps out of the scope of this fix, but since running `npm start` will override most of the output, is keeping `build-watch` around (and maintaining the code going forward) worth it?
> 
> For instance @jiangheng90, why do you use `npm run build-watch` (which rebuilds on file change) instead of the output from `npm start` (which rebuilds when the relevant files are requested in the browser)?

@ggetz I use `npm run build-watch` is just coincident, At that time I found workspace is diffrent, I can not make src code changes work at sandcastle and just find `npm run build-watch` can rebuild and use it

Issue #10611, opened 2022-08-01
Unable to run `npm start` inside release ZIP file


sanjeetsuhag:
Due to the following missing files in the release ZIP, the `npm start` command fails to run:
- `build.cjs`
- `Apps/Sandcastle/.jshintrc`

I could be wrong but I think we should not have to run any build step and simply serve up the pre-built files.

ggetz:
We can consider a completely static server on `start` for the release zip, but we'll need to configure a sperate task (or sperate options) for release vs local development.

PR #10904, opened 2022-11-15
Skip build/rebuild when serving in release zip

ggetz:
- Fixes https://github.com/CesiumGS/cesium/issues/10611 by including a "production" flag in `server.js` which skips the build/rebuild process. The flag is included in the modified `package.json` which we create for the release zip as part of `make-zip`.
 - I noticed the `Spec` files were no longer packaged in the release zip. Although they should be omitted from the nom packages,  they should be included in the release zip so that test can be run in both the browser and on the command line. See discussion in https://github.com/CesiumGS/cesium/pull/10311.
 
 (Most of the changes here are whitespace changes. You can ignore with the PR code display settings.)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -575,6 +575,7 @@ export const makeZip = gulp.series(release, async function () {
       "Build/Cesium/**",
       "Build/CesiumUnminified/**",
       "Build/Documentation/**",
+      "Build/Specs/**",
       "Build/package.json",
       "packages/engine/Build/**",
       "packages/widgets/Build/**",
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -527,6 +527,10 @@ async function pruneScriptsForZip(packageJsonPath) {
   delete scripts["deploy-set-version"];
   delete scripts["website-release"];
 
+  // Set server tasks to use production flag
+  scripts["start"] = "node server.js --production";
+  scripts["start-public"] = "node server.js --public --production";
+
   // Write to a temporary package.json file.
   const noPreparePackageJson = join(
     dirname(packageJsonPath),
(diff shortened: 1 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/start-static/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

gulpfile.js
```diff
@@ -571,6 +575,7 @@ export const makeZip = gulp.series(release, async function () {
       "Build/Cesium/**",
       "Build/CesiumUnminified/**",
       "Build/Documentation/**",
+      "Build/Specs/**",
```
sanjeetsuhag, line 578: We also need to include the `Specs` for the packages. That aside, the motivation for including the Spec files (instead of the bundles `SpecList.js` that allows running of the tests from the ZIP file) is a little unclear to me.

ggetz: This is needed to run the test commands from within the packaged release zip, which was the previous behavior and was discussed in https://github.com/CesiumGS/cesium/issues/3911. I

I was able to execute the top-level tests, both in browser after running `npm start` and on the command line after running `npm test` with this inclusion. 

This just includes the built specs ready for execution, right?

sanjeetsuhag: Yes, you're right, sorry for the confusion. What I meant to ask is whether or not the root level `Specs` folder should be included given that we're not including the `Specs` from the workspaces.

ggetz: Since the complete spec list is used as the entry point to create the bundle, the top-level specs have all the code needed to run all the specs together, correct?

PR #10912, opened 2022-11-18
Fix point primitive scaling by distance

### LINKED ISSUE #10908: PointPrimitive size is growing too much and too fast when using NearFarScalar for scaleByDistance
Hi all,

I am having some issues using a `NearFarScalar` for the `scaleByDistance` property of a PointPtimitive.
If I correctly understand the documentation and also the "Points" sandcastle example (https://sandcastle.cesium.com/?src=Points.html), it should work like the following:
1. The point should have its `pixelSize` multiplied by `NearFarScalar.nearValue` when the camera is at a distance of `NearFarScalar.near` or nearer to the primitive.
2. The point should have its `pixelSize` multiplied by `NearFarScalar.farValue` when the camera is at a distance of `NearFarScalar.far` or further away from the primitive.
3. The point's size should be an interpolated value between the values of 1. and 2. if the camera is in a distance between `NearFarScalar.near` and `NearFarScalar.far`.

In the sandcastle example below, I'd therefore expect the point to be 1 pixel in diameter when the camera is at a distance of 2e5 meters or greater and to be 200 pixel in diameter when it is at a distance of 1 meter or nearer, because the point's `pixelSize` is set to 1 and the `NearFarScalar` is set to `near: 1, nearValue: 200, far: 2e5, farValue: 1`.

What is actually happening (when you zoom onto the point) is that the point is getting way bigger than 200 pixels when the camera is still much further away than 1 meter. I think, it stops growing when the camera is still some kilometers above the point.
I measured the on screen size with the screen ruler of the Microsoft PowerToys (and measuring in the screenshot in an editing program yields the same result):
point-too-large

Note: Browser zoom is at 100% and OS zoom is also at 100% for the screen the picture was taken on.

Sandcastle example:
https://sandcastle.cesium.com/#c=bVDRTsIwFP2Vmz1tyew2kQxhECPEJ6MPJD715dJdsLFrSduBaPh3yyYJUR5603PuOSc9FUY7DztJe7IwBU17mJOTbcPeOi7mkejw3GiPUpPlUTLhmuvew0h76SU5hnUdf3MNsDUuMEaPz0lztD7cUA/Y2ppmQRtL5OKbcsiG92VZpnCXs3wwGg2StA+Q2o+hCwtIfpJayi8aQ5H2lBOo6PGwkM6jFmFx8ewXQvuEdhkkaOMihds8D4OGKRR9/DHM46lDlEY8mpyo8wGonD8omvUA4EE2W2M9tFbFjGWemq3CUCZbteKDPBPOJZNfZ3ZprWq5A1lPr3wfCIXOhc26VV0xHs2qLOj/WZXBWurN646swsNJ9l7MnnuSMVZlAV53emPUCu2f5B8

Browser:
Google Chrome 107.0.5304.107

Operating System:
Windows 11 10.0.22621 Build 22621

UniquePanda:
Fixes #10908 

The problem was, that the size of every point primitive was increased by 3 pixels in `PointPrimitiveCollectionVS.glsl` (for anti-aliasing purposes) before any other modification of the its size was done.  
This is now moved further done in the shader behind the multiplication with the "near far scalar" which is used for "scale by distance".

Previously the additional 3 pixels were also multiplied by the "near far scalar" which resulted in bigger point sizes than expected.
I am not 100% sure if this really was a bug or intentional. If this was intentional, I am happy to change this PR to only add a note about that to the documentation, if that's okay.

~~I also just realized that this change might be considered a breaking change, because it will affect all applications that currently render points which are scaled with a "near far scalar".~~ *(It's not breaking, see comments below)*

Here is a [local sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVDBbsIwDP0Vq6cidWkZQ2VQ0DTQTtN2QNqpF5MaZi1NUJLC2MS/L201CW3kEPk9v/diRxrtPByYjmRhDpqOsCTHTS3eOi4uI9nhpdEeWZMto8Gs1KXuPYK0Z8/kBFZV/F1qgL1xgTF6+pu0ROtDhXokttbUK9pZIhff5GMxvs/zPIG7TGSjyWQ0SPoA1n4KXVhA/ElqzV80hWGW9JyTqOjxtGLnUcvQuZj7hdA+oV0HCdp4mIWTtE64pXEo+ifO4T4PZlESFc6fFC36XIAHrvfGemisioVIPdV7hWH6dNPID/JCOteu30qL9NJaVHwAruZX/gukQudCZ9uobpMyWhRp0P+zKoMV693rgazCUyt7Hy6ee1IIUaQBXnd6Y9QG7Z/kHw) that demonstrates the issue.  
In the current main branch, the point has a size of about 130 pixels when zoomed in (measured with Microsoft PowerToys on screen ruler). Which is the initial size of 10 pixels + 3 pixels multiplied by 10 (near value of near far scalar).
With this PR the point size is about 100 pixels, which is the inital size of 10 pixels multiplied by 10.

Current main:
image

With this PR:
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### 1.100 - 2022-12-01
+
+##### Fixes :wrench:
+
+- Fixed a bug where the scale of a `PointPrimitive` was incorrect when `scaleByDistance` was set to a `NearFarScalar` [#10912](https://github.com/CesiumGS/cesium/pull/10912)
+
 ### 1.99 - 2022-11-01
 
 #### Major Announcements :loudspeaker:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,7 +24,6 @@
     - `Widgets`(CSS files only)
   - The ability to import modules and TypeScript definitions from individual files has been removed. Any imports should originate from the `cesium` module (`import { Cartesian3 } from "cesium";`) or the combined `Cesium.js` file (`import { Cartesian3 } from "Source/Cesium.js";`);
 
-
 ### 1.99 - 2022-11-01
 
 #### Major Announcements :loudspeaker:
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -338,3 +338,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Gabriel Aldous](https://github.com/Sn00pyW00dst0ck)
 - [金俊](https://github.com/jinjun1994)
 - [Oussama Bonnor](https://github.com/oussamabonnor1)
+- [Marcel Wendler](https://github.com/UniquePanda)
--- a/packages/engine/Source/Shaders/PointPrimitiveCollectionVS.glsl
+++ b/packages/engine/Source/Shaders/PointPrimitiveCollectionVS.glsl
@@ -28,8 +28,6 @@ void main()
     float outlinePercent = outlineWidthBothSides / totalSize;
     // Scale in response to browser-zoom.
     totalSize *= czm_pixelRatio;
-    // Add padding for anti-aliasing on both sides.
-    totalSize += 3.0;
 
     float temp = compressedAttribute1.x * SHIFT_RIGHT8;
     float show = floor(temp);
@@ -111,6 +109,9 @@ void main()
 #ifdef EYE_DISTANCE_SCALING
     totalSize *= czm_nearFarScalar(scaleByDistance, lengthSq);
 #endif
+    // Add padding for anti-aliasing on both sides.
+    totalSize += 3.0;
+
     // Clamp to max point size.
     totalSize = min(totalSize, u_maxTotalPointSize);
     // If size is too small, push vertex behind near plane for clipping.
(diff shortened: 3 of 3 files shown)


cesium-concierge:
Thanks for the pull request @UniquePanda!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/10908-fix-point-primitive-scaling-by-distance/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the PR @UniquePanda!

@sanjeetsuhag Would you be able to review this PR? The current behavior sounds unintentional, but there may be something I'm missing related to anti-aliasing.

> I also just realized that this change might be considered a breaking change, because it will affect all applications that currently render points which are scaled with a "near far scalar".

While this will alter behavior slightly, given the change shouldn't cause depending applications to fail without a change, I think we can safely list this under "Fixes" in `CHANGES.md`. Please add an item for this fix, thanks!

UniquePanda:
> Please add an item for this fix, thanks!

Done.  
As there was no current paragraph for the upcoming release, I added one. Please have a look if that's correct.

Issue #10917, opened 2022-11-21
Superfluous characters in "HTML body & CSS" tab of all Sandcastles


javagl:
The contents of the "HTML body & CSS" tab of sandcastles starts with an unnecessary `";`:

![Cesium SandcastleHtml](https://user-images.githubusercontent.com/5597569/203085148-493371f6-c72f-4795-a3c8-c3dd1cb4e574.png)

Note that this does **only** happen in the deployed version of the Sandcastles, at https://sandcastle.cesium.com/index.html . It does not happen when running CesiumJS locally.

I started investigating this a bit. I'm reasonably sure that this is caused by the messed up semicolons/string quote characters in [this line](https://github.com/CesiumGS/cesium/blob/ee8bc4b6ffed3c13e79476b7aa87f1d1f90906a7/gulpfile.js#L2067), which causes an output of

    <script>window.CESIUM_BASE_URL = "../CesiumUnminified/";</script>";

where it should only be

    <script>window.CESIUM_BASE_URL = "../CesiumUnminified/"</script>

as it can be seen in the source view:

![Cesium Sandcastle HTML Characters](https://user-images.githubusercontent.com/5597569/203113625-30b8d382-05cd-468e-8bd4-cdb3ac46723f.png)

The combination of my lack of knowledge about the build- and deployment process as a whole and the fact that it only appears in the deployed version makes it impossible (for me, in a reasonable time frame) to confirm this, but ... I'd say that it's very likely that this is the reason. (Even if it was not the reason for the issue, this should still be fixed...)

PR #10918, opened 2022-11-21
Remove unnecessary characters in sandcastle HTML

javagl:
This is supposed to address https://github.com/CesiumGS/cesium/issues/10917 , but I can not easily test whether this _really_ fixes it.

If somebody can 

- glance at it and say "Yep, that's it" with reasonable certainty (or ... just try it out)
- knows (or can make sure) that this change does _not_ affect anything _except_ for the deployed sandcastle

then this might be an easy fix. Otherwise, I'd leave it to others to investigate the issue further.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -338,3 +338,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Gabriel Aldous](https://github.com/Sn00pyW00dst0ck)
 - [金俊](https://github.com/jinjun1994)
 - [Oussama Bonnor](https://github.com/oussamabonnor1)
+- [Marco Hutter](https://github.com/javagl)
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2087,7 +2087,7 @@ function buildSandcastle() {
         gulpReplace(
           '    <script type="module" src="../load-cesium-es6.js"></script>',
           '    <script src="../../../Build/CesiumUnminified/Cesium.js"></script>\n' +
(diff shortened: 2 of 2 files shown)


cesium-concierge:
Thank you so much for the pull request @javagl! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/javagl/cesium/blob/fix-sandcastle-html-characters/CONTRIBUTORS.md) file!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-sandcastle-html-characters/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
I believe this is an isolated fix. Thanks @javagl! 

I think at this point, you can go ahead and add yourself to `CONTRIBUTORS.md`. 😄

javagl:
@ggetz I (finally) updated CONTRIBUTORS.md - even though they are only small contributions for now.

Should I try to test this further in any way? If I should test it locally, I'd pragmatically set `isProduction=true` near https://github.com/CesiumGS/cesium/blob/ee8bc4b6ffed3c13e79476b7aa87f1d1f90906a7/gulpfile.js#L2060 and see whether the change here makes the expected difference, but I wonder whether the change here might have unexpected side effects that I don't have on the radar. (That `if (isProduction)` scares me... :D )

ggetz:
Thanks @javagl!

ggetz:
The `if (isProduction) ` path builds the version of sand castle which is deployed to the website. Since it also has the same issue, I applied the same fix. It should also apply for the standalone version, which is less obvious because the source is hidden.

Issue #9952, opened 2021-12-01
Fix tests that rely on connection to Cesium ion


sanjeetsuhag:
The specs excluded in [this commit](https://github.com/CesiumGS/cesium/commit/b85d8a449284c9d6f6bcc9fc4226e47cc0594630) on the `cesium.com` branch (which change the Ion default access token) should be fixed to not have to rely on connection to Cesium ion.

TheMagicNacho:
@sanjeetsuhag , I'm noticing that the test runner won't load at all on this branch, regardless of network connection. This issue is limited to this branch though.
I'm going through looking for a root cause now.

ggetz:
These tends to account for most of the intermittent failures in CI these days, especially those which rely on terrain. 

For instance:

```
      ✗ height reference accounts for change in terrain provider
	Expected true to be false.
	    at <Jasmine>
	    at UserContext.<anonymous> (packages/engine/Specs/Scene/Model/ModelSpec.js:2477:36 <- Build/Specs/SpecList.js:270203:36)
```

PR #10933, opened 2022-12-02
Updates coverage CI stage to only run on appropriate repo/branches

sanjeetsuhag:
These changes are needed for the following reasons:

1. Test failures are guaranteed on the `cesium.com` branch due to some tests (like `sampleTerrain`) requiring connection to Cesium ion, and the access token on the `cesium.com` branch is limited to usage from the `cesium.com` domain. Similar to the exclusion of the test step in `test-release.sh`, a check has been added to ensure that tests are not run on the `cesium.com` branch.
2. The uploading of the coverage is done through the `aws s3 sync` command, and this is only setup to work from the `CesiumGS/cesium` repo, so there's a check added for that step too.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/ci-exclusions/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

travis/coverage.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ev
+if [ $TRAVIS_BRANCH != "cesium.com" ]; then
```
sanjeetsuhag, line 3: A potentially cleaner way of accomplishing this could be to use this condition to omit the stage entirely, like so:

```
    - stage:
      name: "Coverage"
      node_js: "18"
      if: repo != CesiumGS/cesium OR branch != cesium.com
      script:
        - ./travis/coverage.sh
```

However, I'm not sure if this omission of the stage would be confusing. We already have a lot of branching in the CI scripts and it might not be the best ideas to introduce the conditions in two different ways. I'm looking for feedback for this.

ggetz: I _think_ we still want to run the coverage step itself in non "CesiumGS/cesium" repos, as it's the only non-release version of the test suite that's run. I can see an argument for how that may not be catching anything but niche test failures, but I think it would be useful to provide an external contributor the test failure sooner in the process to avoid extra cycles of review.

sanjeetsuhag: Ok, that sounds good to me.

ggetz:
Thanks @sanjeetsuhag! Can you add a note in https://github.com/CesiumGS/cesium/issues/9952 to remove the condition once tests are refactored?

PR #10934, opened 2022-12-02
Updates postversion gulp task to update all dependent workspaces

sanjeetsuhag:
This PR updates the `postversion` gulp task that is run after the `npm version` command used in the release process. The main change here is that when you run `npm version` for workspace `X`, it updates the `package.json` for all workspaces that depend on workspace `X`.

Since running any gulp task requires having run `npm install`, the release guide has been updated to explicitly call for an `npm install` before running `npm version`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -545,20 +545,36 @@ async function pruneScriptsForZip(packageJsonPath) {
   return gulp.src(noPreparePackageJson).pipe(gulpRename(packageJsonPath));
 }
 
-export const postversion = function () {
+export const postversion = async function () {
   const workspace = argv.workspace;
   if (!workspace) {
     return;
   }
-
-  // After the workspace version is bumped by `npm version`,
-  // update the root `package.json` file to depend on the new version
   const directory = workspace.replaceAll(`@${scope}/`, ``);
   const workspacePackageJson = require(`./packages/${directory}/package.json`);
   const version = workspacePackageJson.version;
 
-  packageJson.dependencies[workspace] = version;
-  return writeFile("package.json", JSON.stringify(packageJson, undefined, 2));
+  // Iterate through all package JSONs that may depend on the updated package and
+  // update the version of the updated workspace.
+  const packageJsons = await globby([
+    "./package.json",
+    "./packages/*/package.json",
+  ]);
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/release-guide-update-order/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

gulpfile.js
```diff
@@ -545,20 +545,36 @@ async function pruneScriptsForZip(packageJsonPath) {
...
+  ]);
+  const promises = packageJsons.map(async (packageJsonPath) => {
+    // Ensure that we don't check the updated workspace itself.
+    if (basename(dirname(packageJsonPath)) === directory) {
+      return Promise.resolve();
+    }
+    // Ensure that we only update workspaces where the dependency to the updated workspace already exists.
+    const packageJson = require(packageJsonPath);
+    if (!(workspace in packageJson.dependencies)) {
+      return Promise.resolve();
```
ggetz, line 571: Same as above.

gulpfile.js
```diff
@@ -545,20 +545,36 @@ async function pruneScriptsForZip(packageJsonPath) {
...
+    "./packages/*/package.json",
+  ]);
+  const promises = packageJsons.map(async (packageJsonPath) => {
+    // Ensure that we don't check the updated workspace itself.
+    if (basename(dirname(packageJsonPath)) === directory) {
+      return Promise.resolve();
+    }
+    // Ensure that we only update workspaces where the dependency to the updated workspace already exists.
+    const packageJson = require(packageJsonPath);
+    if (!(workspace in packageJson.dependencies)) {
```
ggetz, line 570: As dependencies are ultimately pretty arbitrary key/value pairs, I think this may be an [undesirable use of `in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#inherited_properties) and we should opt for `Object.hasOwn` instead.

sanjeetsuhag:
@ggetz Updated.

ggetz:
Thanks @sanjeetsuhag!

Issue #10349, opened 2022-05-04
Update `IndexDatatype.createTypedArray` to create Uint8Arrays


j9liu:
Currently `IndexDatatype.createTypedArray` only outputs `Uint16Array` and `Uint32Array`. This caused a bug in #10344 because it returned a `Uint16Array` for uint8 indices, which was too big for the `getBufferData` function. The fix itself is easy but it causes a lot of unit tests to break, which will take more time to fix.

Once this function is changed to return `Uint8Array`, the manual check in `buildDrawCommands` can be removed to accommodate this.

malitherl:
Hello, may I be assigned to work on this issue?

j9liu:
Hi @malitherl,

Thanks for your interest in contributing! For smaller issues (usually labeled "good first issue"), there's no need to ask for permission. You can work on the issue and open a pull request for review when ready.

If you have any questions, you can comment on the issue or on the PR itself if it's far enough along to open. There's also a lot of standard documentation in [the Contributes folder](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors) if you need guidance.

Harjot4698:
hie @malitherl,
If you are seeing this message , please reply

Sugarlust:
I've made the necessary changes but I need some help with the test case as I'm new to this.

rravula77:
Hello, my team and I would like to work on the issue

ggetz:
Great @rravula77! Let us know if you have any questions or would like to discuss the approach.

Charlyjcmu:
I am part of rahul's team. We are a little confused on how to change the test cases. Many of the errors are due to the expected render not matching the actual render. Should we be copying the actual render to the expected render or should we be calculating if the output is correct? 

We don't really understand what is a render or a scene, so we don't understand how to calculate the correct render. Where can we learn more about this?

ggetz:
Hi there, a "render" is the process by which a single frame or image is created., while the "scene" is a top-level object which generally defines what is being rendered. To understand how it's used in unit testing, 
check out the [Rendering Tests section of our Testing Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/TestingGuide#rendering-tests) to learn more. 

Would you also mind posting the test errors so we can help troubleshoot the specifics?

Charlyjcmu:
So most of the failed render test cases are failing due the pixel not matching, We are a little confused on how to correct for this error. 

[...]

PR #10894, opened 2022-11-08
Enables WebGL 2 by default

sanjeetsuhag:
Fixes https://github.com/CesiumGS/cesium/issues/10887, https://github.com/CesiumGS/cesium/issues/10064

### Overview

This PR enables WebGL 2 by default. The main changes here are the upgrade of shaders to GLSL 300 from GLSL 100. Instead of using `modernizeShader` to upgrade the GLSL version at runtime, from now, when a WebGL 1 context is detected, shaders will be downgraded using `demodernizeShader`.

### To-Do

- [x] Upgrade GLSL shader files
- [x] Upgrade JS shader code
- [x] Implement demodernize shaders
- [x] Fix tests
- [x] Test Sandcastles
- [x] Update documentation/PR description

### Testing

To test all Sandcastles with a WebGL 1 context, use this [git patch](https://github.com/CesiumGS/cesium/files/10191276/sandcastle_webgl1.patch.txt) and use this command:

```bash
git apply --ignore-space-change --ignore-whitespace sandcastle_webgl1.patch
```

### Known Issues

 - [Scene/PostProcessStage/can use a texture uniform](http://localhost:8080/Specs/SpecRunner.html?category=none&spec=Scene%2FPostProcessStage%20can%20use%20a%20texture%20uniform) spec failure
 - ~[3D Tile Classification Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20Terrain%20Classification.html&label=Development) is broken~

PR #10955, opened 2022-12-14
Adding Support for Uint8arrays in createTypedArrays

Charlyjcmu:
We have added support for Uint8arrays when using IndexDatatype.createTypedArrays in IndexDatatype.js for issue #10349 . We have added a numberOfVertices check inside of the function and created a new constant called TWO_FIFTY_SIX_BYTES inside Math.js

We have corrected a few test cases inside Core/Geometry/Pipeline, Core/IndexDatatype, Scene/PrimitivePipeline, Scene/Model/GeoJsonLoader.

We are a little confused on how to \ correct the render test cases, because there is an internal GL_INVALID_OPERATION: Insufficient buffer size error. Most of the errors are just expecting to render some pixel but not rendering it.

We would like some help on how to get started correcting the following test cases:
- [ ] Scene/Geometry3DTileContent
- [ ] Scene/Vector3DTileContent
- [ ] Scene/Vector3DTileGeometry
- [ ] Scene/Vector3DTilePolygons
- [ ] Scene/Vector3DTilePolylines


cesium-concierge:
Thanks for the pull request @Charlyjcmu!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
@j9liu As you opened the original issue, would you mind providing some guidance on the tests here? Particularly around the error:

> `DeveloperError: IndexDatatype.UNSIGNED_INT requires OES_element_index_uint, which is not supported on this system.  Check context.elementIndexUint.`

If this requires a particular extension, should there be a fall back in place for systems that its not supported on?

packages/engine/Source/Core/Math.js
```diff
@@ -183,6 +183,13 @@ CesiumMath.SOLAR_RADIUS = 6.955e8;
  */
 CesiumMath.LUNAR_RADIUS = 1737400.0;
 
+/**
+ * 256
+ * @type {Number}
+ * @constant
+ */
+CesiumMath.TWO_FIFTY_SIX_BYTES = 256;
```
j9liu, line 191: Keep in mind that `IndexDatatype.UNSIGNED_BYTE` corresponds to just one byte, and one byte = 256 bits.
```suggestion
CesiumMath.TWO_FIFTY_SIX_BITS = 256;
```

j9liu:
@ggetz I'll take a look at this PR but I am confused why the `UNSIGNED_INT` would be causing errors now, since this PR is just adding support for typed arrays with uint8 or `UNSIGNED_BYTE`. Was there no fallback in the first place for using `UNSIGNED_INT` ?

It also seems like this functionality is provided by WebGL2, and [only in WebGL1 is the extension needed](https://developer.mozilla.org/en-US/docs/Web/API/OES_element_index_uint). So what would the correct procedure be for checking `UNSIGNED_INT` support when #10894 is merged?

packages/engine/Source/Core/IndexDatatype.js
```diff
@@ -132,8 +132,11 @@ IndexDatatype.createTypedArray = function (
   if (numberOfVertices >= CesiumMath.SIXTY_FOUR_KILOBYTES) {
     return new Uint32Array(indicesLengthOrArray);
   }
+  if (numberOfVertices >= CesiumMath.TWO_FIFTY_SIX_BYTES) {
```
j9liu, line 135: I didn't note this when I wrote the issue, but this change should probably apply to `createTypedArrayFromArrayBuffer` as well (the function below this one)

packages/engine/Source/Core/IndexDatatype.js
```diff
@@ -109,12 +109,12 @@ IndexDatatype.validate = function (indexDatatype) {
 };
 
 /**
- * Creates a typed array that will store indices, using either <code><Uint16Array</code>
- * or <code>Uint32Array</code> depending on the number of vertices.
+ * Creates a typed array that will store indices, using either <code><Uint8Array</code>
+ * or <code><Uint16Array</code> or <code>Uint32Array</code> depending on the number of vertices.
```
j9liu, line 113: Minor readability tweak:
```suggestion
 * Creates a typed array that will store indices, using either <code><Uint8Array</code>,
 * <code><Uint16Array</code>, or <code>Uint32Array</code> depending on the number of vertices.
```

packages/engine/Source/Core/IndexDatatype.js
```diff
@@ -109,12 +109,12 @@ IndexDatatype.validate = function (indexDatatype) {
...
  *
  * @param {Number} numberOfVertices Number of vertices that the indices will reference.
  * @param {Number|Array} indicesLengthOrArray Passed through to the typed array constructor.
- * @returns {Uint16Array|Uint32Array} A <code>Uint16Array</code> or <code>Uint32Array</code> constructed with <code>indicesLengthOrArray</code>.
+ * @returns {Uint8Array|Uint16Array|Uint32Array} A <code>Uint8Array</code> or <code>Uint16Array</code> or <code>Uint32Array</code> constructed with <code>indicesLengthOrArray</code>.
```
j9liu, line 117:
```suggestion
 * @returns {Uint8Array|Uint16Array|Uint32Array} A <code>Uint8Array</code>, <code>Uint16Array</code>, or <code>Uint32Array</code> constructed with <code>indicesLengthOrArray</code>.
```

ggetz:
Thanks @j9liu!

> Was there no fallback in the first place for using UNSIGNED_INT ?

`Context.elementIndexUint` has a check in place already, but bases on the test result, my guess is that value is not being properly accounted for.

> It also seems like this functionality is provided by WebGL2, and [only in WebGL1 is the extension needed](https://developer.mozilla.org/en-US/docs/Web/API/OES_element_index_uint). So what would the correct procedure be for checking UNSIGNED_INT support when https://github.com/CesiumGS/cesium/pull/10894 is merged? 

It will still be possible to opt into WebGL1, in which case we should still be checking for extension support. The above function should already account for that.

j9liu (commented):
The problem with the unit tests comes from rendering problems with the geometry itself. The changes have caused holes in some of the geometry:

| main | this branch |
| ---- | -------------|
| image | image |

There might be code in the geometry classes that assumes the size of the indices being output from `createTypedArrays`. I'll have to do a sweep through the code for mentions of `IndexDatatype.createTypedArrays` and see where the problem areas are.

j9liu:
I'll have to return to this later but from snooping around I found a couple things:

- There's this part of the code that has a `Uint16` vs `Uint32` bit check that we missed
https://github.com/CesiumGS/cesium/blob/27abb992eb5a434f1c9391f5e926643473ebc985/packages/engine/Source/Renderer/VertexArray.js#L678-L697
- This is what happens in the 3D Tiles Terrain Classification sandcastle (which uses `Vector3DTileContent`). This also happens in the ArcticDEM sandcastle.
image
- Something I noticed about `EllipseGeometry` while trying to debug it: it looks like both `Uint16` and `Uint8` arrays get made at some point in the progress. So if these differently-sized indices get packed together, they're going to lead to bad results. We'd need to look closer into that code though to figure out what incorrect assumptions are being made.

j9liu:
Hi @Charlyjcmu,

Were you and your team going to continue work on this? Even though it has become more difficult than expected, I'm willing to investigate the areas that are producing these bugs and provide guidance on what needs changing.

Otherwise, this PR will be closed for inactivity and I will update the issue to describe the full scope of this problem.

Charlyjcmu:
Yes, our team will keep working on it. We took a break for a while, but we will keep working on it for a while longer.

PR #10958, opened 2022-12-15
Fix file watcher for workers

ggetz:
Previously, when running `npm start` and modifying worker files, `server.js` would 1) incorrectly build the worker files, and 2) not continue watching after an error. This passes the correct options through to `bundleWorkers` and catches and logs any errors while continuing to watch files for more updates.


sanjeetsuhag (approved):
Changes look good. Thanks @ggetz!

PR #10959, opened 2022-12-15
Update autolinker and jsep dependencies

ggetz:
This updates the remaining outdated direct dependencies– jsep and autolinker. For both, there were slight tweaks to the APIs which were updated here.

Lerc is still an outstanding direct dependency update. The new version now requires using wasm. It looks like, despite the fact that the wasm file is built to support multiple ways of loading the module, [the lerc API only allows one way to load wasm files](https://github.com/Esri/lerc/blob/master/OtherLanguages/js/src/Lerc.ts#L153) and that way does not fit within our worker architecture, throwing errors when loading. I'll writeup a separate issue to cover it.

### DIFF
--- a/packages/engine/Source/DataSources/GpxDataSource.js
+++ b/packages/engine/Source/DataSources/GpxDataSource.js
@@ -41,11 +41,9 @@ const autolinker = new Autolinker({
   stripPrefix: false,
   email: false,
   replaceFn: function (linker, match) {
-    if (!match.protocolUrlMatch) {
-      // Prevent matching of non-explicit urls.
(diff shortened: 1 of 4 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/update-packages/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

sanjeetsuhag:
Not exactly familiar with these dependencies and how they work, but the changes look fine to me on a surface level and the tests and Sandcastle for related files are working as expected.

Thanks @ggetz!

Issue #10970, opened 2022-12-20
Intermittent CI failure


ggetz:
Multiple tests occasionally fail in CI with the following error:

```
RuntimeError: An error occurred while retrieving the EOP data from the URL Data/EarthOrientationParameters/EOP-2011-July.json.
```

<img width="1038" alt="image" src="https://user-images.githubusercontent.com/4439461/208704646-44ef5495-7ad1-47c2-8b7b-93137ad2fb1d.png">

ggetz:
This seems to be due to an asynchronous antipattern in the `EarthOrientationParameters` constructor, as described in https://github.com/CesiumGS/cesium/issues/10909.

Issue #10909, opened 2022-11-17
Race condition in Cesium3DTileset / Reconsider `readyPromise` pattern


mramato:
* Paste this into a new file called `aTrivialSpec.js` into `packages/engine/Specs/Scene/`.

```js
import { Cesium3DTileset, DeveloperError } from "../../index.js";
import createScene from "../../../../Specs/createScene.js";

fdescribe("trivialSpec", function () {
  let scene;
  beforeAll(function () {
    scene = createScene();
  });

  afterAll(function () {
    scene.destroyForSpecs();
  });

  it("a trivial test", function () {
    const tileset = new Cesium3DTileset({ url: "test.invalid" });
    expect(function () {
      throw new DeveloperError();
    }).toThrowDeveloperError();
    tileset.destroy();
  });
});
```

* Run the unit tests
* Most of the time it will fail/hang with the below error
```
$ npm -- run test --browsers ChromeHeadless

> cesium@1.99.0 test
> gulp test --browsers ChromeHeadless

[20:29:15] Using gulpfile ~/Git/CesiumGS/cesium/gulpfile.js
[20:29:15] Starting 'test'...

  trivialSpec
    ✓ a trivial test
Chrome Headless 107.0.5304.110 (Linux x86_64) ERROR
  An error was thrown in afterAll
  Unhandled promise rejection: Request has failed. Status Code: 404 thrown
Chrome Headless 107.0.5304.110 (Linux x86_64) ERROR
  An error was thrown in afterAll
  Unhandled promise rejection: Request has failed. Status Code: 404 thrown
0.068 secs: trivialSpec a trivial test
0 secs: Renderer/Buffer fails to destroy: WebGL 2
0 secs: Renderer/Buffer fails to copy a large array view: WebGL 2
0 secs: Renderer/Buffer fails to provide an array view: WebGL 2
0 secs: Renderer/Buffer destroys: WebGL 2
0 secs: Renderer/Buffer copyFromBuffer with index buffers: WebGL 2
0 secs: Renderer/Buffer copyFromBuffer with vertex buffers: WebGL 2
0 secs: Renderer/Buffer copyFromBuffer throws when readBuffer is the same buffer and copy range overlaps: WebGL 2
0 secs: Renderer/Buffer copyFromBuffer throws with one index buffer and the other is not an index buffer: WebGL 2
0 secs: Renderer/Buffer copyFromBuffer throws with invalid sizeInBytes: WebGL 2

[...]

PR #10971, opened 2022-12-20
Remove async paths from EOP constructor

ggetz:
Should fix https://github.com/CesiumGS/cesium/issues/10970

* Moves async functionality in the EOP constructor to `fromUrl` (this is a private API so no deprecation or `CHANGES.md` update necessary)
* Fixup error handling so there should only be one error thrown
* Fixup specs to reflect new error handling

I believe this is the first PR to use async/await. I figured it's time given we're about to submit more async changes as part of https://github.com/CesiumGS/cesium/issues/10909, and this should help flatten a code for better readability.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/no-async-eop-constructor/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

packages/engine/Source/Core/EarthOrientationParameters.js
```diff
@@ -19,10 +20,6 @@ import TimeStandard from "./TimeStandard.js";
  * @constructor
```
sanjeetsuhag, line 20: Should we add a warning here that this should typically not be constructed directly, like we do with `Model.js`?

https://github.com/CesiumGS/cesium/blob/b9106c0a915be1b59d1f35878ebbbe6f97505f43/packages/engine/Source/Scene/Model/Model.js#L42

sanjeetsuhag (commented):
This is a great improvement for readability and feels like cleaner code overall! Love that we don't have to have smaller state/promise variables hanging around.

packages/engine/Source/Core/EarthOrientationParameters.js
ggetz: There are cases where the constructor may be used. I've add a disclaimer similar to `IonResource` - 

> This object is normally not instantiated directly, use {@link EarthOrientationParameters.fromUrl}.

ggetz:
Thanks @sanjeetsuhag! Updated!

sanjeetsuhag:
Thanks @ggetz!

Issue #7895, opened 2019-06-03
Edge Test Failure - Scene/Cesium3DTileset tile expires


hpinkos:
Scene/Cesium3DTileset tile expires
Expected -0.21600000000034924 to equal epsilon 0, 0.1.

http://localhost:8080/Specs/SpecRunner.html?spec=Scene%2FCesium3DTileset%20tile%20expires

sanjeetsuhag:
This came up again in the 1.98 release.

jjhembd:
I get this error on my local machine. Can we simply increase the tolerance?

The problem code is:
```javascript
          // Check that expireDuration and expireDate are correctly set
          const expireDate = JulianDate.addSeconds(
            JulianDate.now(),
            5.0,
            new JulianDate()
          );
          expect(
            JulianDate.secondsDifference(tile.expireDate, expireDate)
          ).toEqualEpsilon(0.0, CesiumMath.EPSILON1);
```

The errors I have seen are all smaller than 1. Replacing `CesiumMath.EPSILON1` with `1.0` should be enough.

PR #10972, opened 2022-12-20
Spec fixes 2022-12-20

javagl:
I noticed various failures when running the tests locally (on a _German_ Windows...). I haven't yet figured out the reasons (even less solutions) for all of them, but will update this PR as I move on:

1. Errors due to assumptions about the locale that is used - for example:
```
1) show tile debug labels with memory usage
     Scene/Cesium3DTileset
     Expected 'Texture Memory: 0
Geometry Memory: 0,007' to equal 'Texture Memory: 0
Geometry Memory: 0.007'.
Error: Expected 'Texture Memory: 0
Geometry Memory: 0,007' to equal 'Texture Memory: 0
Geometry Memory: 0.007'.
    at <Jasmine>
    at packages/engine/Specs/Scene/Cesium3DTilesetSpec.js:2294:58 <- Build/Specs/SpecList.js:139998:58
```
This was caused by comparing a string that was created with `toLocaleString` (resulting in `0,007` on a German system) to a fixed string (namely `0.007`). The current solution is still a bit brittle, because ... the formatting for the labels does _more_ than just use a different locale (i.e. a string like `0.000123` would still have unexpected results). But I think that it could be OK...

2. Tile expiration tests. This is the same as https://github.com/CesiumGS/cesium/issues/7895 - I'll see whether I find a solution for that one

3. Some tests _seem_ to pass, but they actually fail, and that can only be seen when running the test manually. For example, the test http://localhost:8080/Specs/SpecRunner.html?category=none&spec=Scene%2FGlobe%20renders%20terrain%20with%20vertexShadowDarkness seems to work within `npm run test`, but when running it via the link above, it prints "Unhandled promise rejection: SyntaxError: Unexpected token '�', "�������="... is not valid JSON". Something's wrong with the structure of the test, I guess...

### DIFF
--- a/packages/engine/Specs/Scene/Cesium3DTilesetSpec.js
+++ b/packages/engine/Specs/Scene/Cesium3DTilesetSpec.js
@@ -2289,7 +2289,9 @@ describe(
         expect(tileset._tileDebugLabels).toBeDefined();
         expect(tileset._tileDebugLabels.length).toEqual(1);
 
-        const expected = "Texture Memory: 0\n" + "Geometry Memory: 0.007";
+        const expected =
+          "Texture Memory: 0\n" +
+          `Geometry Memory: ${(0.007).toLocaleString()}`;
 
         expect(tileset._tileDebugLabels._labels[0].text).toEqual(expected);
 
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/spec-fixes-2022-12-20/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @javagl! I think just getting in the fix for tests which differ by locale will be an improvement. Is this ready for review and merge?

javagl:
I intended to collect "all" the fixes for these failures in one PR. But ... the reasons for some of them are subtle, so that may have been too ambitous anyhow, and ... it's often "better" to make smaller, "single-issue-PRs". So I'll just mark that as RFR. (Assuming that such a hyper-special test fix is not supposed to be mentioned in CHANGES.md)

Issue #7471, opened 2019-01-07
Incorrect decoding of draco model with high quantization bit number


OmarShehata:
This was reported [by Chris Wang on the forum](https://groups.google.com/d/msg/cesium-dev/x_GU7yw0lTw/TJyW75ZeDAAJ). 

The glTF file attached [in this post](https://groups.google.com/d/msg/cesium-dev/x_GU7yw0lTw/3kDUYfV4DAAJ) was encoded with the following Draco parameters:

```
sOpt.pos_quantization_bits = 20;
sOpt.tex_coords_quantization_bits = 15;
```

And renders incorrectly in Cesium. Here's what it looks like in Bablyon:

![babylon](https://user-images.githubusercontent.com/1711126/50790876-252ef080-128e-11e9-8c67-426e21c315ff.jpg)

And in CesiumJS:

![cesium](https://user-images.githubusercontent.com/1711126/50790863-1cd6b580-128e-11e9-84a4-a04e8c455079.jpg)

@lilleyse believes this might be because the quantization is hardcoded in a few places in our Draco decoding.

lilleyse:
> @lilleyse believes this might be because the quantization is hardcoded in a few places in our Draco decoding.

Specifically I wouldn't be surprised if UNSIGNED_SHORT is hardcoded somewhere.

wangzhongliang:
@lilleyse Hi Sean, will this be fixed soon? if not, can you tell me where is possibly hardcoded, so that I can fix myself.

wangzhongliang:
When I set dequantizeInShader to false, the model is not distorted, and displays well. 
Hope it will help.

lilleyse:
This also came up in https://github.com/AnalyticalGraphicsInc/gltf-pipeline/issues/511

lilleyse:
Opened an issue in the Draco repo about whether it's possible to query the quantization bits without decoding to quantized form: https://github.com/google/draco/issues/573

@sanjeetsuhag and I talked offline - if there's no way to do it in the API, we'll need to dequantize in our own code.

Ben-Mack:
@lilleyse Has this bug been fixed on the latest Cesium?

lilleyse:
@Ben-Mack, no, still an open issue

lilleyse:
Came up again in https://github.com/CesiumGS/cesium/issues/10977

lilleyse:
Fixed in https://github.com/CesiumGS/cesium/pull/10979

Issue #10977, opened 2022-12-27
draco will decode wrong when quantizationBits is above 16


jiangheng90:
In draco worker ```decodeDraco.js```
```javascript
function decodeQuantizedDracoTypedArray(
  dracoGeometry,
  dracoDecoder,
  dracoAttribute,
  quantization,
  vertexArrayLength
) {
  let vertexArray;
  let attributeData;
  if (quantization.quantizationBits <= 8) {
    attributeData = new draco.DracoUInt8Array();
    vertexArray = new Uint8Array(vertexArrayLength);
    dracoDecoder.GetAttributeUInt8ForAllPoints(
      dracoGeometry,
      dracoAttribute,
      attributeData
    );
  } else {
    attributeData = new draco.DracoUInt16Array();
    vertexArray = new Uint16Array(vertexArrayLength);
    dracoDecoder.GetAttributeUInt16ForAllPoints(
      dracoGeometry,
      dracoAttribute,
      attributeData
    );
  }

  for (let i = 0; i < vertexArrayLength; ++i) {
    vertexArray[i] = attributeData.GetValue(i);
  }

  draco.destroy(attributeData);
  return vertexArray;
}
```
when single data stride above 16 
```
vertexArray[i] = attributeData.GetValue(i);
```
this can not store data correctly

lilleyse:
Duplicate of https://github.com/CesiumGS/cesium/issues/7471

PR #10979, opened 2022-12-27
support quantizationBits above 16 for draco decode

jiangheng90:
fix quantizationBits above 16 draco decode went wrong in  #10977


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
Thanks @jiangheng90! I just want to point out one possible side effect which is that uint32 values will lose some precision when stored in a `Float32Array` since some integers above 2^24 can't be represented exactly in 32-bit floating point, but since the compression is lossy already it's not a big concern. There isn't really a good alternative since [`vertexAttribPointer`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/vertexAttribPointer) doesn't support uint32 values.

I was also concerned that the code would have some assumption that quantized values need to be stored in an integer typed array, but after stepping through the code that doesn't seem to be the case, which is good. Maybe the old Model code had this assumption.

Fixes https://github.com/CesiumGS/cesium/issues/7471. I confirmed that the original model in the issue is working now: 

image

PR #11027, opened 2023-01-09
Build specs as part of test command

ggetz:
This change makes running tests on the command line a bit simpler for convenience's sake. Previously, you would need to re-run (or rely on `npm start` to be running in the background) after making a change to a spec to see the updated code reflected in `npm run test`. That could be particularly annoying if you made a quick change, such as focusing a test with `fit`, as you need to remember to rebuild.

The `test` command now incorporate that build spec step, which should only take about an extra second or so.

I also clean up the function names a bit here for clarity, and remove an unused function.

### DIFF
--- a/build.js
+++ b/build.js
@@ -288,6 +288,10 @@ const workspaceSpecFiles = {
   widgets: ["packages/widgets/Specs/**/*Spec.js"],
 };
 
+/**
+ * Creates a single entry point file, Specs/SpecList.js, which imports all individual spec files.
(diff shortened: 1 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/build-test/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

PR #11040, opened 2023-01-17
Fix loading of label background

### LINKED ISSUE #10461: Label background is broken when use inputAction (unwanted gradient)
Hi all,
Can I get help with this problem, Is this a bug?
As far as I know, it still happens 1.9 ~ later

You can see it in action in the attached sandcastle and in this screenshot:
bug_label_background

Sandcastle example:
https://sandcastle.cesium.com/#c=fZTdbuIwEIVfxZurILGm3b1LKdpumgq0EVSQbW8iVcYZwMLYke1A6Yp3Xzt/BUp7g+yZOd8Zjx2oFNqgLYMdKHSLBOxQCJoVG/xUxvzUo+U+lMIQJkClXucmFamgpZKJvDCnwhlVAGKWEwrRFoQZEpFxS6pMMCViS7RjlFqswYzc4o4aJoW/KES1AKftoH+pQKjyUmRvnWqMpiDAwjagCF6CeWR0PSX7SoZzqZmjOBuE2ML/VjeXwcIeIvMtq1PDEVJgCiXK0oP7ebdsOOe+Sy7ngHPr6Ujdk+Snnm1TXxrXKHsKWwwakyzz7USJgZjMgUcuvj9i3VTCQ/ez+Sf7HHAcPSQvYTwK/1S314wZfYGuu6yaPL7hurA+RFMftKtuleAOGjRnRcjAqwlQ6iXRLEm9bhPWZs8haNhlJzMXwg+jOH65G9+/TP4m8WgctQpZGG4nGkouVSssd/h5OEo+FD6zzKwCdN3G54Sul0oWIrvEmEb3782t5O53Wx0gowpok1tQhlHCJ4otmWgpTydhnEweW8VKKvbmvqRzzfAsgcNonETTVpizV+CTxcJ+LcHxVYREGbsi4od/1UXff151akn5ng7usg9e1+uXQx40tF9sk0tlUKG4j3HPwCbn9hXo3rygazCYal29K4T6vWNpP2NbxLLbC/8KiHKitc0sCs5n7A1Sb9Dv2foPUi5JxsRyYsfHyd6Vra4HcRXEGPd7dntZaaTkc6LOyP8B

Browser:
Tested on Chrome 102.0.5005.115

Operating System:
Windows 10

UniquePanda:
Fixes #10461

This adds a new boolean parameter `updateImageIndex` to `TextureAtlas.prototype.addSubRegion()` which is `false` by default. If it is true, the `_indexHash` value for the given image id is updated to point to the index of the new sub region. Otherwise the index will still point to the region of the whole image (same behavior as currently).

This is needed because billboards create such a subregion, but wouldn't currently use this subregion in later calls to `_loadImage` which results in the problems described in the issue.

I am not really sure if my solution is good, because I am no expert on this topic. There are some open questions:
1. Does this has any side effects, that I overlooked?
2. Should the index be updated always (without the need for the `updateImageIndex` parameter)?
3. Should there be a way to still access the previous region? Maybe two arrays are needed, a `_indexHash` and a `_subRegionIndexHash`?
4. `addSubRegion` isn't used anywhere else (only by `_loadImage()` of `Billboard.js`). Should this be refactored in general somehow?

I didn't add/update any tests yet, because of those questions.

Longer description of what I think is the problem of the current code that is fixed by this PR:
An `imageSubRegion` for the billboard backgrounds is added to the texture atlas on the first load of a billboard.
Later calls will not add anything to the texture atlas, but will request the correct image by relying on the texture atlas' `_indexHash`.  
In these later calls the index is retrieved from the texture atlas and then directly passed to `completeImageLoad()`.
This function then uses this index to get the correct texture coordinates from the texture atlas. The problem is, that the image sub region is not taken into account anymore at this point.
The initial call, that added the sub region, is getting the index by calling `TextureAtlas.addSubRegion()` which returns a promise that resolves to the index. This is the index of the created subregion.
However, later calls will receive the index of the whole image, because the `_indexHash` will still point to the index of the image and not its sub region. That's why the background looks different then.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 - Fixed a bug decoding glTF Draco attributes with quantization bits above 16. [#7471](https://github.com/CesiumGS/cesium/issues/7471)
 - Fixed an edge case in `viewer.flyTo` when flying to a imagery layer with certain terrain providers. [#10937](https://github.com/CesiumGS/cesium/issues/10937)
 - Fixed a crash in terrain sampling if any points have an indefined position due to being outside the rectangle. [#10931](https://github.com/CesiumGS/cesium/pull/10931)
+- Fixed label background rendering. [#11040](https://github.com/CesiumGS/cesium/issues/11040)
 
 ### 1.101 - 2023-01-02
 
--- a/packages/engine/Source/Scene/TextureAtlas.js
+++ b/packages/engine/Source/Scene/TextureAtlas.js
@@ -498,17 +498,19 @@ TextureAtlas.prototype.addSubRegion = function (id, subRegion) {
     }
     const atlasWidth = that._texture.width;
     const atlasHeight = that._texture.height;
-    const numImages = that.numberOfImages;
 
     const baseRegion = that._textureCoordinates[index];
     const x = baseRegion.x + subRegion.x / atlasWidth;
     const y = baseRegion.y + subRegion.y / atlasHeight;
     const w = subRegion.width / atlasWidth;
     const h = subRegion.height / atlasHeight;
-    that._textureCoordinates.push(new BoundingRectangle(x, y, w, h));
+    const newIndex =
+      that._textureCoordinates.push(new BoundingRectangle(x, y, w, h)) - 1;
+    that._indexHash[id] = newIndex;
+
     that._guid = createGuid();
 
-    return numImages;
+    return newIndex;
   });
 };
 
--- a/packages/engine/Specs/Scene/TextureAtlasSpec.js
+++ b/packages/engine/Specs/Scene/TextureAtlasSpec.js
@@ -804,6 +804,8 @@ describe(
         expect(coordinates[index4].y).toEqual(0.5 / atlasHeight);
         expect(coordinates[index4].width).toEqual(0.5 / atlasWidth);
         expect(coordinates[index4].height).toEqual(0.5 / atlasHeight);
+
+        expect(atlas._indexHash[greenGuid]).toEqual(3);
       });
     });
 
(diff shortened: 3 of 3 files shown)


cesium-concierge:
Thanks for the pull request @UniquePanda!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/10461-fix-label-background/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

UniquePanda:
> I think this was an oversight in #10275

Ah yes, makes sense. I was wondering, why the problem didn't occur before and had forgotten about that PR :)

I removed the parameter, the index hash is now updated always, when a sub region is added.
I also added one line to the tests to test that behavior. However, I wasn't able to get that particulat test to fail, even if I randomly changed numbers in it. So for me the tests passes locally, but it does so always, regardless if the assertions are correct or not 😅  
So you can either keep the line in and trust it (if the CI passes) or someone else should try if the test fails with changed numbers 😅

ggetz:
Another good catch @UniquePanda! That test was not properly calling the final expectations of that test. I re-wrote the problematic specs a bit so that the expectation are actually being executed.  I also made a tweak to your new expectation so that we avoid calling private members of the class where possible because accessing private members makes tests more tightly coupled to the source code, and therefore more fragile.

If you could add a note to `CHANGES.md` mentioning the issue being fixed, and assuming my tweaks pass CI, this should be good to merge!

UniquePanda:
> That test was not properly calling the final expectations of that test

Haha, I have to admit that confused me a bit. I actually started to change the test like you did, but then thought I might just don't understand what it was doing  Thanks for fixing. 😄 

I updated the `CHANGES.md`.

ggetz:
There's some antiquated promise behavior in some tests, so it's good to get them cleaned up. 😄 Thanks again @UniquePanda, happy to get this fix in!

PR #8182, opened 2019-09-19
Scale geometric error with tileset

hpinkos:
Fixes #5330
Fixes #7398

Clean up from https://github.com/AnalyticalGraphicsInc/cesium/pull/7411

@lilleyse please review

Issue #5330, opened 2017-05-16
3D Tiles scaling does not scale geometric error


mramato:
When setting a scale transform on the root tile in order to scale a 3D Tileset, bad things happen.  The tileset may draw, but at the wrong resolution and it may also disappear completely when it should take up the whole screen.  @likangning93 thinks the problem is that Cesium is not taking the scale transform into account when computing the final client geometric error, leading to incorrect tile selection.

lilleyse:
If we decide to address this in Cesium, we should also update the spec:

From `tile.schema.json`
> A floating-point 4x4 affine transformation matrix, stored in column-major order, that transforms the tile's content, i.e., its features and content.boundingVolume, and boundingVolume and viewerRequestVolume from the tile's local coordinate system to the parent tile's coordinate system, or tileset's coordinate system in the case of the root tile.  **transform does not apply to geometricError** nor does it apply any volume property when the volume is a region, which is defined in EPSG:4326 coordinates.

pjcozzi:
I probably wrote that part of the spec, but it is not going to work.  Two options:

* Explicitly disallow scale as part of the spec.  It will make the client a bit easier to implement if they actually follow the spec.  Cesium could still expose its own scale/transform.
* Allow scale in the spec - including the GE - and fix Cesium, which I do not think will be hard unless I am missing something.

I would go with 2.

PR #11047, opened 2023-01-24
Apply root tile scale to tileset geometric error

lilleyse:
This is related to the problem that was brought up in https://github.com/CesiumGS/cesium/issues/5330. As of https://github.com/CesiumGS/cesium/pull/8182 we take a scale into account when computing a tile's geometric error (see [`updateGeometricErrorScale`](https://github.com/CesiumGS/cesium/blob/f78063329b4bb1b845b27f793c42decbf41ec383/packages/engine/Source/Scene/Cesium3DTile.js#L1790-L1794)) but we don't take scale into account when computing the top-level tileset geometric error.

This can lead to situations where the tileset's geometric error is less than the root tile's geometric error and traversal [exits early](https://github.com/CesiumGS/cesium/blob/f78063329b4bb1b845b27f793c42decbf41ec383/packages/engine/Source/Scene/Cesium3DTilesetTraversal.js#L60-L66).

The fix is to scale the tileset's geometric error by the root tile's scale.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-tileset-geometric-error/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the fix, @lilleyse! I tested this out on a tileset where the traversal was exiting early and the issue is resolved.

jo-chemla:
Cross-posting from the cesium [community forum](https://community.cesium.com/t/3d-model-disappearing-in-location-editor-when-camera-is-far-away/20864/18): 

I have the same issue, where setting a `maximumScreenSpaceError of 0.1` makes it so that my tileset appears at lower zoom levels - and it disappears when this max error is set to simply 1.0. I’m using Cesium version 1.106. Has this fix been merged to the current cesium build? Thanks a lot for getting back!

Looks like the fix indeed has been [merged](https://github.com/CesiumGS/cesium/commit/e912d6c2e84ad5fedc1aba8ee480aebdc6fd4cd1) into 1.102, but I’m still seeing this issue. If you’d like a link to the dataset, I can share it privately.

ggetz:
Hi @jo-chemla, you can send private data to support@cesium.com. Please include a link to the original post so we know what the context is. Thanks!

PR #11056, opened 2023-01-30
Fix incorrectly-passing spec for `AttributeCompression.decodeRGB565`

j9liu:
The value 2881 was _supposed_ to correspond to an unnormalized RGB value of (1, 1, 1), but it actually represented (1, 26, 1). This was never caught by the unit test because it used an incorrectly-sized `for` loop to test the result.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-rgb565-spec/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
Nice catch!

PR #11059, opened 2023-01-31
Deprecate ready promises

ggetz:
Will fix https://github.com/CesiumGS/cesium/issues/10909 by deprecating `readyPromise` pattern throughout the API.

### Why

The `readyPromise` pattern has been a problematic part of the CesiumJS API that does not match modern promise best practices. 
 * The background asynchronous nature can cause race conditions, especially if they do not properly handle destroyed objects. This has been coming up often in tests, leading to  "random" CI failures as mentioned in #10909. 
 * The existing API can [cause confusion around how to handle errors](https://community.cesium.com/t/cesium3dtileset-not-loading-tileset-json-from-any-url-and-silently-failing-with-bad-urls/16287). 
 * When `readyPromises` are unhandled, in many modern systems, a failed promise can lead to a critical error.

### How

The plan is to deprecate `readyPromise` properties throughout the API starting with the next release (1.103). As this affects high-touch areas of the API like imagery providers, terrain providers, and 3D Tiles, we'll remove the old code paths after three releases. 

The new implementation moves asynchronous operations out of synchronous constructors and into async factory functions such as `fromUrl`, which perform operations like loading necessary metadata up front, returning a promise that resolves to the created 3D tileset or provider. The created object can be assumed fully ready, and [all properties can be accessed immediately](https://community.cesium.com/t/urltemplateimageryprovider-broken-after-upgrade-from-1-91-to-1-92/22245/5) (such as `Cesium3DTileset.boundingSphere`). 3D tilesets or providers will still make requests for new tiles as changes to the camera and scene occur and is handled internally, the success or failure of which are exposed via events.

A few examples of this are as follows:

##### Before
```js
const tileset = scene.primitives.add(
  new Cesium.Cesium3DTileset({
    url: Cesium.IonResource.fromAssetId(1240402),
  })
);

tileset.readyPromise
  .then(function (tileset) {
    viewer.zoomTo(
      tileset,
      new Cesium.HeadingPitchRange(
        0.5,
        -0.2,
        tileset.boundingSphere.radius * 4.0
      )
    );
  })
  .catch(function (error) {
    console.log(error);
  });
```

##### After
```js
try {
  const tileset = await Cesium3DTileset.fromUrl(Cesium.IonResource.fromAssetId(1240402));
  scene.primitives.add(tileset);
  viewer.zoomTo(
      tileset,
      new Cesium.HeadingPitchRange(
        0.5,
        -0.2,
        tileset.boundingSphere.radius * 4.0
      )
    );
[...]

PR #11060, opened 2023-01-31
Use top level await in Sandcastle

ggetz:
This  updates Sandcastle code to use top-level `await` syntax by making `window.startup` and async function. That function is now awaited and wrapped with a `catch`, and any errors are logged in the console.

As a result of this, this PR also removes any blocks with the `(async () => { ... })();` pattern.

This DOES NOT account for a final pass on how we are handling async code in Sandcastle examples before we publish changes discussed in https://github.com/CesiumGS/cesium/pull/11059. This is an update that mainly simplifies syntax some existing examples.

### DIFF
--- a/Apps/Sandcastle/Sandcastle-helpers.js
+++ b/Apps/Sandcastle/Sandcastle-helpers.js
@@ -4,7 +4,7 @@
   window.embedInSandcastleTemplate = function (code, addExtraLine) {
     return (
       `${
-        "window.startup = function (Cesium) {\n" +
+        "window.startup = async function (Cesium) {\n" +
         "    'use strict';\n" +
         "//Sandcastle_Begin\n"
       }${addExtraLine ? "\n" : ""}${code}//Sandcastle_End\n` +
@@ -12,7 +12,9 @@
       `};\n` +
       `if (typeof Cesium !== 'undefined') {\n` +
       `    window.startupCalled = true;\n` +
-      `    window.startup(Cesium);\n` +
+      `     window.startup(Cesium).catch((error) => {\n` +
+      `      console.error(error);\n` +
+      `    });\n` +
       `}\n`
     );
   };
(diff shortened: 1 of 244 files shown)


jjhembd:
Thanks @ggetz, this is a big cleanup for the Sandcastles that await something, and a minimal change for Sandcastles without async code. I confirmed that links to local Sandcastles developed before this PR will still work.

One question about the pattern at the end of the gallery Sandcastles: Some of them end with this pattern, like `AmbientOcclusion.html`:
```javascript
        //Sandcastle_End
        Sandcastle.finishedLoading();
      };
      if (typeof Cesium !== "undefined") {
        window.startupCalled = true;
        try {
          window.startup(Cesium);
        } catch (error) {
          console.error(error);
        }
      }
```

Others end like this (`3D Tiles Next S2 Globe.html`):
```javascript
      if (typeof Cesium !== "undefined") {
        window.startupCalled = true;
        window.startup(Cesium).catch((error) => {
          "use strict";
          console.error(error);
        });
        Sandcastle.finishedLoading();
      }
```

I think this is neater than the `try...catch` block, but:
- I don't think the "use strict" is needed?
- Do we always want the `Sandcastle.finishedLoading()` inside the if block?

A few I3S-related Sandcastles don't have any `.finishedLoading` call.

ggetz:
Thanks for pointing that out. These should use the `startup().catch` pattern. The try/catch pattern would need another higher level `async` function and an addition `await` keyword. The mis-match is due to and incomplete find/replace which  I've updated. I've also update the I3S Sandcastles.

> I don't think the "use strict" is needed?
eslint is calling for it, I believe it's there for the same reason as the `"use strict";` as the first line of the `startup` function.

> Do we always want the Sandcastle.finishedLoading() inside the if block?
I suppose it doesn't have to be. I was thinking if we do have a long-running `await` in the Sandcastle code, we probably might want to show the loading overlay until the `startup` function is complete rather than just called. You can see the viewer once created even if additional async code is running. However, it should be consistant in all the Sandcastles either way.

jjhembd:
Looks good, merging now!

Issue #11061, opened 2023-02-01
implement smooth zoom with mouse wheel


jiangheng90:
I found inertiaZoom in `ScreenSpaceCameraController` is not working.
Then I found this issue is still open in #4537 
so I implement it. by define aggregator
this pull request is in #11062

ggetz:
Thanks @jiangheng90! We're taking a look at your PR now. In the meantime, I'm going to close this issue since it's a duplicate of https://github.com/CesiumGS/cesium/issues/4537.

Issue #4537, opened 2016-10-27
Zooming with the mouse wheel is jumpy


tfili:
When you use the right mouse button to zoom it is very smooth, but when you use the wheel it is very jumpy. For the sake of consistency we should have the wheel move smoothly as well.

emackey:
@bagnell could this be done by having the wheel impart zoom-inertia, rather than simply change the position?

bagnell:
Yes. The wheel used to have inertia. @hpinkos tracked it all the way back to b24 when it was removed.

pjcozzi:
CC #2968

AliFlux:
Any idea if this feature is coming back? Google Earth zoom is pretty smooth, and there's another opensource lib called OpenGlobus with same same inertial zoom:
http://openglobus.org/examples/emptyTerrain/emptyTerrain.html

OmarShehata:
@AliFlux if you'd like to propose an implementation of this in a PR that could help generate interest/see if others would like to see this change too.

PR #11062, opened 2023-02-01
implement inertiaZoom in aggregator

jiangheng90:
implement of smooth zoom with mouse wheel in #11061 and #4537


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/smooth-zoom-with-mouse-wheel/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Hi @jiangheng90, thanks for the PR! Do you still have interest in contributing this?

jiangheng90:
> Hi @jiangheng90, thanks for the PR! Do you still have interest in contributing this?

Yes, But After I finish test, it has performance problem on pickGlobe. so I'm not sure is this the reason inertiaZoom has not been support for so many years.

jiangheng90:
> Hi @jiangheng90, thanks for the PR! Do you still have interest in contributing this?

@ggetz

ggetz:
> Yes, But After I finish test, I found it has performance problem on pickGlobe. So I'm not sure is this the reason inertiaZoom has not been support for so many years.

Ah, good point. https://github.com/CesiumGS/cesium/issues/8481, right? Can you explain how you tested performance and what your results were? I wonder if we can make any optimizations to skip successive picks.

jiangheng90:
> > Yes, But After I finish test, I found it has performance problem on pickGlobe. So I'm not sure is this the reason inertiaZoom has not been support for so many years.
> 
> Ah, good point. #8481, right? Can you explain how you tested performance and what your results were? I wonder if we can make any optimizations to skip successive picks.

@ggetz 
I‘m not sure my explain will be accurate.
First of all, my test environment:
i9-10900k  RTX-3070 at 3160 * 2160 resolution
i7-4700k    Quadro K2000 at 1920 * 1080 resolution

test in emply globe, with default ion Imagery, no terrain.
at 1080p environment the fps will drop down to 30 with frequently zoom in and zoom out
at 2160p environment it will be fine at 60fps

test with Aerometrex San Francisco 3dtiles(almost cached), cesium terrain, with default ion Imagery
at 2160p environment it will be drop to 38-52 with frequently zoom in and zoom out

after that I have make a performance trace on zoom in and zomm out
企业微信截图_20230206091512
we can see execute time of ```readPixel``` in ```zoom3D``` is very hgh

jiangheng90:
@ggetz after I do test again this morning to describle the issue. I found the ```pickGlobe``` in ```zoom3D``` may reduce call by use pick result last frame

ggetz:
Thanks @jiangheng90! I'll do a  review on this PR tomorrow.

jiangheng90:
> Thanks @jiangheng90! I'll do a review on this PR tomorrow.

👌

ggetz:
Thanks @jiangheng90! I'm happy with the performance numbers. The zoom behavior on trackpad and touch screen also appear to be consistant with the old behavior, which is good.

My one concern while trying this out was that it seemed to zoom very fast now that intertia is in place with the scroll wheel. What do you think about halfing the delta value we get from the mouse wheel event? https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/ScreenSpaceEventHandler.js#L424-L438

jiangheng90:
> Thanks @jiangheng90! I'm happy with the performance numbers. The zoom behavior on trackpad and touch screen also appear to be consistant with the old behavior, which is good.
> 
> My one concern while trying this out was that it seemed to zoom very fast now that intertia is in place with the scroll wheel. What do you think about halfing the delta value we get from the mouse wheel event? https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/ScreenSpaceEventHandler.js#L424-L438

@ggetz  I have already halfing ```delta``` value by calculate the ```arcLength```

packages/engine/Source/Scene/CameraEventAggregator.js
```diff
@@ -141,21 +144,29 @@ function listenToWheel(aggregator, modifier) {
...
-      const arcLength = 15.0 * CesiumMath.toRadians(delta);
-      if (!update[key]) {
-        movement.endPosition.y = movement.endPosition.y + arcLength;
-      } else {
-        Cartesian2.clone(Cartesian2.ZERO, movement.startPosition);
-        movement.endPosition.x = 0.0;
-        movement.endPosition.y = arcLength;
-        update[key] = false;
-      }
+      const arcLength = 7.5 * CesiumMath.toRadians(delta);
```
ggetz, line 163: > I have already halfing delta value by calculate the arcLength

@jiangheng90 I see. I think we should further reduce this since the zoom is currently fast and its easy to loose control. 

If you could, please also remove the comment that says `TODO: magic numbers`. We should not have left the `TODO` comment in the code to begin with.

jiangheng90: @ggetz already removed TODO comment. So. I see may TODO comment in cesium. But I'm not sure in which case this will be removed

jiangheng90:
@ggetz already remove ```TODO``` comment.

jiangheng90:
@ggetz this pr going to continue or paused?

ggetz:
Sorry for the delay @jiangheng90! I didn't realize my comment on zoom speed had been addressed. Thank you!

We would like to get this merged. However, one thing I notice here is that the camera gets "jumpy" and is no longer smooth once it approaches the ellipsoid surface. You can reproduce this by zooming straight into the ground when terrain is disabled.

Since the building dataset is technically below the ellipsoid surface as well, I find this easy to reproduce in [the San Fransisco example](https://sandcastle.cesium.com/?src=Aerometrex%20San%20Francisco.html) after clicking "Pier 39" and zooming into the building.

I think we may need to adjust the behavior here when close to the surface. Perhaps we fall back to the previous behavior in that case.

jiangheng90:
@ggetz I think this is caused by disable pickGlobe when camera approaches height 0. the distance will across 0 and previouse result for calculate distance  maybe incorrect. so in a very low height. I reopen ```pickGlobe ``` in a height above -50 and below 50

jiangheng90:
And I think test failure is from the server error

ggetz:
Thanks @jiangheng90! The behavior here is looking fantastic. 

I've also restarted CI, and the tests have passed.

Thank you again for your patience here. This is a great fix to get in.

Issue #9529, opened 2021-05-03
SingleTileImageryProvider fetches images when show is false


jnewmoyer:
Using the SingleTielImageryProvider with a layer, we are seeing an issue where Cesium pulls the image resource even when show is set to false on the layer. It does not display the image, but the request can be seen to the resource in the Network tab. 

The example below can be used to observe this. 

Sandcastle example: https://sandcastle.cesium.com/#c=bVJdb9owFP0rVp6CRB3KxOggRWNdpUXKtKpU60sk5CSXYNWxI9sE0ar/ff4gLJS+xLHvOb7n+NyWSNRS2INEt4jDHt2Borsa/3VnYRYUbn8nuCaUg8yCIXrLOEK0JhXIw4MULS1BzjpiIYFoeBaSlYmHhA6PkNIHBidcIngftLJFvLx/TJbp+jl5+rVOlz/u09XQct8HbsmJgpQcQD7Q4sW23BCmwJTeB/OMZ7w1XpitK+PFm8KqAA74KNaR1dwjc0aKl99E5gwM3PMwKTtBnbHQdu49jNH9EfGGiFKgk3KGvtxcj43cjHtFvR6YsGZLTKcRnlzUckmrreagrPIxHp3sNMce5+GsKK8YPFEGF1L8U+8km6EswDhyzlXkietUVGItWpDGLm54ZdL0BAmFJvbSUz6P3QneSFH/hEoCqPBqOsGjIRrf2O/V16nbfMPTyeAY1HkQ57KTXgph58yMk9qK/cxl6fj/owjd72AeDIPYTc/Cq0XoO60bIbU1GhqXGuqGmbFTUb4zo6FxoZS9ykLjqE+NS9oiWt5+MtioYCZHU9nsGFvRV8iCRRwZ/AWVCVKaCP74h7Sw7fUi9YcY4zgy28+ZWgiWE/nh5n8

Browser: Chrome

Operating System: Windows

This seems like a bug to me. In my opinion the resource should not be fetched until the layer is actually turned on. This is the way other providers work, including the WMS provider. 

In our use case, we may have many user added images on the map, but only a few selected at a time, so we do not want the overhead of many unnecessary requests going on in the background. 

Thanks.

ebogo1:
Thanks for opening the issue @jnewmoyer. 

@mramato is this a bug? I vaguely remember discussing this a while ago (or something similar) and don't remember what the decision was at the time.

ggetz:
[...]

PR #11075, opened 2023-02-07
Deprecate ImageryProvider ready promises

ggetz:
Part of #10909

Deprecates `readyPromise` and `ready` properties of `ImageryProvider` and derived classes.

* `ImageryProvider` properties and functions no longer throw an exception since there's no need to wait for `ready`
* Where applicable, `url` (or the relevant resource-related options) have been deprecated from constructors. In their place, there are new `fromUrl` (or equivalent functions), which are asynchronous functions that resolve to a new instance of the provider after successfully requesting the required resources. This function call throws any relevant errors directly rather than logging it to the console or raising an error event.
* I have not yet touched the Sandcastle examples or Widgets yet, as given discussion in https://github.com/CesiumGS/cesium/pull/11033 around the top level API, I'm thinking it would be most productive to address in a separate PR.
* `ImageryLayer.getViewableRectangle` can become synchronous, so this was deprecated in favor of `ImageryLayer.getImageryRectangle.
* Fixes https://github.com/CesiumGS/cesium/issues/9529 by allowing lazy-loading for `SingleTileImageryProvider` if `tileWidth` and `tileHeight` are provided to the constructor.

TODO:
- [x] Test failure - `RuntimeError: An error occurred while accessing http://test.invalid/flatfile?q2-0-q.1: TypeError: Worker.postMessage: Element of argument 2 is not an object.`
- [x] Merge https://github.com/CesiumGS/cesium/pull/11078, create Imagery Provider equivalent, and update Sandcastle


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

CHANGES.md
```diff
@@ -11,12 +11,29 @@
 
 ##### Additions :tada:
 
-- Added `ArcGISTiledElevationTerrainProvider.fromUrl`, `CesiumTerrainProvider.fromUrl`, `GoogleEarthEnterpriseMetadata.fromUrl`, `GoogleEarthEnterpriseTerrainProvider.fromMetadata`, `VRTheWorldTerrainProvider.fromUrl`, and `createWorldTerrainAsync` for better async flow and error handling.
+- Added `ArcGisMapServerImageryProvider.fromUrl`, `ArcGISTiledElevationTerrainProvider.fromUrl`, `BingMapsImageryProvider.fromUrl`, `CesiumTerrainProvider.fromUrl`, `GoogleEarthEnterpriseMetadata.fromUrl`, `GoogleEarthEnterpriseImageryProvider.fromMetadata`, `GoogleEarthEnterpriseTerrainProvider.fromMetadata`, `IonImageryProvider.fromAssetId`, `SingleTileImageryProvider.fromUrl`, `TileMapServiceImageryProvider.fromUrl`, `VRTheWorldTerrainProvider.fromUrl`, and `createWorldTerrainAsync` for better async flow and error handling.
```
ggetz, line 14: Please note that I kept the deprecation strings as 1.102/1.104. This matches the existing ready deprecation messages that are already in place, so it will be easier to find/replace them as a whole before https://github.com/CesiumGS/cesium/pull/11059 is merged, most likely for 1.103.

packages/engine/Source/DataSources/ModelVisualizer.js
```diff
@@ -386,7 +386,8 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
   if (hasHeightReference) {
     // We cannot query the availability of the terrain provider till its ready, so the
     // bounding sphere state will remain pending till the terrain provider is ready.
-    if (!terrainProvider.ready) {
+    // ready is deprecated. This is here for backwards compatibility
```
ggetz, line 389: There's a few places notes "ready is deprecated. This is here for backwards compatibility" which can be removed when the rest of the deprecated API is. This will be noted in the follow up deprecation issue.

packages/engine/Source/Scene/ImageryLayer.js
```diff
@@ -496,12 +497,37 @@ const terrainRectangleScratch = new Rectangle();
...
+  // readyPromise has been deprecated. This is here for backward compatibility and can be removed with readyPromise.
+  await imageryProvider._readyPromise;
+  return Rectangle.intersection(imageryProvider.rectangle, rectangle);
+};
+
+/**
+ * Computes the intersection of this layer's rectangle with the imagery provider's availability rectangle,
+ * producing the overall bounds of imagery that can be produced by this layer.
+ *
+ * @returns {Rectangle} A promise to a rectangle which defines the overall bounds of imagery that can be produced by this layer.
```
jjhembd, line 517: I think this is a synchronous method now

packages/engine/Source/Scene/WebMapServiceImageryProvider.js
```diff
@@ -340,6 +341,9 @@ function WebMapServiceImageryProvider(options) {
     ),
     enablePickFeatures: options.enablePickFeatures,
   });
+
+  this._ready = true;
+  this._readyPromise = Promise.resolve(true);
```
jjhembd, line 346: I may have missed something, but can you explain what these private properties are for? I don't see them used internally, and the public `.ready` and `.readyPromise` properties point to the relevant properties of UrlTemplateImageryProvider

ggetz: Sure!

So since the constructor method and ready properties are being deprecated, they still need to work before they are entirely removed. The `ready` and `readyPromise` properties are part of the public API, but they are also used internally for rendering imagery and terrain (ie https://github.com/CesiumGS/cesium/pull/11075#discussion_r1098820299 for terrain providers, https://github.com/CesiumGS/cesium/pull/11075/files#diff-da7b68dacf6bee3b99070639c851312908f93636cd7d4a0372ba03dfc254f6b5R244 for imagery providers). In order to avoid printing the depreciation warning from within internal code that a user may not have touched, we're now using the private versions of these properties.

packages/engine/Source/Scene/SingleTileImageryProvider.js
```diff
@@ -159,42 +152,41 @@ function SingleTileImageryProvider(options) {
...
 
-  function success(image) {
-    that._image = image;
-    that._tileWidth = image.width;
-    that._tileHeight = image.height;
-    that._ready = true;
-    TileProviderError.reportSuccess(that._errorEvent);
-    return Promise.resolve(true);
-  }
+  const resource = Resource.createIfNeeded(options.url);
```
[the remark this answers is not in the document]
ggetz: `SingleTileImageryProvider` is a bit of a special case. As discussed in https://github.com/CesiumGS/cesium/issues/9529, there's not a strict need to request the image up front, unless `tileHeight` and `tileWidth` are not provided. As part of the changes for this deprecation, I left a path to use the synchronous constructor if `options.url`, `options.tileHeight`, and `options.tileWidth` are all provided. The deprecation message here is confusing though, I'll update the wording.

I'll also leave a comment to explain this as its non-obvious.

packages/engine/Source/Scene/WebMapServiceImageryProvider.js
jjhembd: Thanks, that makes sense! I hadn't seen the usage in `GlobeSurfaceTileProvider`

packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js
```diff
@@ -61,14 +62,220 @@ import ImageryProvider from "./ImageryProvider.js";
...
+}
+
+/**
+ * Complete ArcGisMapServerImageryProvider creation based on builder values.
+ *
+ * @private
+ *
+ * @param {ArcGisMapServerImageryProvider} provider
+ */
+ImageryProviderBuilder.prototype.build = function (provider) {
```
jjhembd, line 102: If we add this code to `metadataSuccess`, the extra class might not be needed

ggetz:
In order to simplify the imagery API for users, especially on initial configuration with the CesiumWidget and Viewer, the following changes have been made:
- Allow imagery layers to be created from a provider promise. The imagery layer can be immediately used and its properties modified, or destroyed. The asynchronous imagery provider operation are exposed via a `readyEvent` and an `errorEvent`. By default errors are logged to the console, but the behavior can be overridden by subscribing to `errorEvent`.
- To accommodate this change `defaultAlpha`, `defaultGamma`, and related `default*` properties have been deprecated from imagery providers. The only providers which were using one of these properties (`defaultGamma`) were `BingMapsImageryProvider` (which just set the default anyway), and` GoogleEarthEnterpriseMapsProvider`. These properties were never covered by unit tests to begin with, and were not used in any Sandcastle examples. Removing these properties greatly simplifies imageryLayer creation and reduces the overall API.
- `CesiumWidget` and `Viewer` have had the `imageryProvider` constructor option deprecated in favor of a new `baseLayer` property, which prioritized the non async path for initialization.
- Sandcastle examples and documentation examples have been updated to reflect this change.

For example:
```js
          baseLayer: Cesium.ImageryLayer.fromProviderAsync(
            Cesium.TileMapServiceImageryProvider.fromUrl(
              Cesium.buildModuleUrl("Assets/Textures/NaturalEarthII")
            )
          ),
          baseLayerPicker: false,
        });
```

while the async provider creation is still available:
```js
try {
  const imageryProvider = await Cesium.TileMapServiceImageryProvider.fromUrl(
    Cesium.buildModuleUrl("Assets/Textures/NaturalEarthII")
   );
   const imageryLayer = new ImageryLayer(imageryProvider);
   scene.imageryLayers.add(imageryLayer);
} catch (error) {
  // Handle error
}
```

packages/engine/Source/Scene/ImageryLayer.js
```diff
@@ -37,98 +41,130 @@ import SplitDirection from "./SplitDirection.js";
...
+ * @typedef {Object} ImageryLayer.ConstructorOptions
  *
- * @alias ImageryLayer
- * @constructor
+ * Initialization options for the ImageryLayer constructor.
  *
- * @param {ImageryProvider} imageryProvider The imagery provider to use.
- * @param {Object} [options] Object with the following properties:
- * @param {Rectangle} [options.rectangle=imageryProvider.rectangle] The rectangle of the layer.  This rectangle
+ * @property {Rectangle} [options.rectangle=imageryProvider.rectangle] The rectangle of the layer.  This rectangle
```
jjhembd, line 48: I don't think we need the `options.` prefix anymore

jjhembd:
Thanks @ggetz, I finally had a chance to go through your latest changes. I really like how much cleaner it is, after you deprecated all the duplicated styling properties (`.defaultAlpha` etc.) from the providers. These make more sense as properties of `ImageryLayer` anyway.

I think there are just 3 minor things left:

- [ ] HTTP 401 error in 2 terrain specs
- [ ] Constructor types in `ImageryLayer`
- [ ] Deprecation message version numbers

ggetz:
Thank @jjhembd! Updated!

jjhembd:
Looks good, thanks @ggetz! Merging now

PR #11078, opened 2023-02-08
Async event handler for terrain providers

ggetz:
[We are deprecating `readyPromises` throughout the API and making a move towards an API which makes better use of `async/await` type patterns](https://github.com/CesiumGS/cesium/pull/11059).

However, based on discussion in https://github.com/CesiumGS/cesium/pull/11033, it would be convenient for most users to have a simpler API, while still preserving the .fromXYZ-style factory functions for those users who may prefer using async/await and simplifying code further down the stack.

In this approach, `Terrain` is a new object that handles the async error handling and exposes an error event and a ready event. One plus here is that this is more in line with event-driven UIs then `readyPromise` was. `Viewer` and `CesiumWidget` now take `terrain` options, which is then handled by `Scene`.  The result is behavior much like what exists now in `main`, and prevents some common race conditions that could arise if use async/await.
* Passing config values to a `Viewer` or `CesiumWidget` constructors can begin rendering the globe immediately while waiting on a terrain provider promise in the background
* By default, errors are caught and logged to the console, but the user can subscribe to `errorEvent` to override this behavior
 * The existing terrain immediately disappears when setting the new `Terrain` object, so that it's obvious that the update process has started
 * Handles "competing" values, like when setting a terrain provider and then immediately setting another, or a `terrainProvider` directly
 * Handles `Viewer`, `CesiumWidget`, or `Scene` being destroyed while a terrain update in in progress

One downside of this approach is that (without TypeScript) it's not possible to confirm that a promise resolves to the expected type during object creation. However, there are many other areas of the code where this is the case, and hopefully won't be a concern if/when we move to TypeScript.

TODO:
- [x] Sweep other Sandcastle examples and update the API where applicable

### DIFF
--- a/packages/engine/Source/Scene/Terrain.js
+++ b/packages/engine/Source/Scene/Terrain.js
@@ -37,15 +37,16 @@ async function handlePromise(instance, promise) {
  * @example
  * // Create
  * const viewer = new Cesium.Viewer("cesiumContainer", {
- *   terrainProvider: new Cesium.Terrain(Cesium.CesiumTerrainProvider.fromUrl("https://myTestTerrain.com"));
+ *   terrain: new Cesium.Terrain(Cesium.CesiumTerrainProvider.fromUrl("https://myTestTerrain.com"));
  * });
  *
  * @example
  * // Handle loading events
  * const terrain = new Cesium.Terrain(Cesium.CesiumTerrainProvider.fromUrl("https://myTestTerrain.com"));
  *
+ * scene.setTerrain(terrain);
+ *
  * terrain.readyEvent.addEventListener(provider => {
- *   scene.terrainProvider = provider;
  *   scene.globe.enableLighting = true;
  *
  *   terrain.provider.errorEvent.addEventListener(error => {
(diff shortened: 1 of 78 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

packages/engine/Source/Scene/Terrain.js
```diff
@@ -0,0 +1,182 @@
+import Check from "../Core/Check.js";
+import Event from "../Core/Event.js";
+import createWorldTerrainAsync from "../Core/createWorldTerrainAsync.js";
+
+function handleError(errorEvent, error) {
```
jjhembd, line 5: Can we move the helper functions down, so the [constructor is at the top of the file](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#put-the-constructor-function-at-the-top-of-the-file)?

jjhembd:
Thanks @ggetz, this looks good! I just mentioned one nitpick about the ordering of a file.

ggetz:
Thanks @jjhembd! I've responded to your feedback, and I've also gone ahead and swept through Sandcastle examples dealing with terrain. A few made sense to use `await` inside of a `try/catch`, but otherwise, especially where terrain is not the point of the example, I've changed to this new `Terrain` handler.

jjhembd:
Looks good to me @ggetz! Merging now

PR #11098, opened 2023-02-22
Add support for Access Token when creating an ArcGISMapServerImageryProvider

Tamrat-B:
Added support for ArcGIS Access Token when creating an ArcGisMapServerImageryProvider. Updated samples as well.


cesium-concierge:
Thanks for the pull request @Tamrat-B!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/arcgis_secure_updates/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js
```diff
@@ -185,6 +194,72 @@ function ArcGisMapServerImageryProvider(options) {
...
 
+  let accessToken;
+  switch (options.basemap_id) {
+    case "satellite":
+      {
+        accessToken = defaultValue(
+          options.token,
+          ArcGISMapServiceWorldImagery.defaultAccessToken
+        );
+        const server = Resource.createIfNeeded(
```
ggetz, line 205: I don't believe the value of `server` is actually being used. 

It's also not clear that the intention is that setting `options.basemapId` will override the `options.url` parameter. 

Perhaps instead a factory function would make determining the URL from a particular type a bit more straightforward:
```
ArcGisMapServerImageryProvider.fromBasemapType = function (basemapType, options) {
    let server;
    switch (basemapType) {
       // ...
    }

   return new ArcGisMapServerImageryProvider({
       ...options,
       url: server,
   });
}
```

This is similar to the `IonResource.fromAssetId` factory functions.

The usage then becomes:

```
  imageryProvider: Cesium.ArcGisMapServerImageryProvider.fromBasemapType({
     ArcGisBaseMapType.SATELLITE,
    //token: "<token>",
  }),
});
```

packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js
```diff
@@ -185,6 +194,72 @@ function ArcGisMapServerImageryProvider(options) {
   const resource = Resource.createIfNeeded(options.url);
   resource.appendForwardSlash();
 
+  let accessToken;
+  switch (options.basemap_id) {
+    case "satellite":
```
ggetz, line 199: We use enums to manage a defined set of types like this in order to reduce the chance of error (see `ArcType` for an example):

```
/**
 * Types of ArcGIS Basemap layers supported by default.
 *
 * @enum {number}
 * @see ArcGisMapServerImageryProvider
 */
const ArcGisBaseMapType = {
   SATELLITE: 1,
   OCEANS: 2,
   HILLSIDE: 3
}
```

ggetz (changes requested):

packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js
```diff
@@ -357,10 +429,26 @@ function ArcGisMapServerImageryProvider(options) {
       }
 
       that._useTiles = true;
+    } else {
+      const message = `The service does not have any Pre Cached Tiles or the required tileInfo resources and is not supported.`;
```
ggetz, line 433: This is technically a breaking change and will need to be noted as such in `CHANGES.md`

ggetz:
Thanks @Tamrat-B! Great to get the most up-to-date servers in!

A few items are needed before we can merge this PR:
- We'll want to update the [release guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/ReleaseGuide) with instructions on how to rotate the token. 
- There are a few failing unit tests. We'll need those to be passing.
- `CHANGES.md` will need to be updated with with a summary of the changes.

ggetz:
@gowin20 Hi George, nice to meet you! We understand your urgency around having this PR merged. Is it still possible for you to address my comments above? If not, let me know where you'd want some additional help. For example, I'm happy to updates the release guide or help with unit tests.

ggetz:
Hi @gowin20 and @Tamrat-B, just wanted to bump my comment above. To get this change into the upcoming release on March 1, this would need to be merged by tomorrow. Thanks!

ggetz:
@Tamrat-B Great, thank you! 

I'll update the release guide based on my understanding of the token rotation process. 

For unit tests, let me know what issues you may run into and I'll be happy to provide guidance.

packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js
```diff
@@ -212,8 +220,10 @@ function ArcGisMapServerImageryProvider(options) {
   if (typeof credit === "string") {
     credit = new Credit(credit);
   }
-  this._credit = credit;
-
+  if (defined(options.warning)) {
```
ggetz, line 223: With this change, `this._credit` is no longer assigned unless `options.warning` is defined.This breaks the existing behavior– Now when not using `options.warning`, the provided `options.credit` will not show up at all.

I would suggest removing the `warning` parameter option from the constructor here, and instead in `fromBasemapType`, pass the `credit` option when creating the provider.

```
let credit;
if (defined(defaultTokenCredit)) {
    credit = Credit.clone(defaultTokenCredit);
}
return new ArcGisMapServerImageryProvider({
    ...options,
    url: server,
    token: accessToken,
    credit: credit
  });
```

Additionally, there should be no need to set `this._useTiles` in this case. It defaults to `true`, but you can also set `options.usePreCachedTilesIfAvailable` to `true` if you'd like to be explicit.

```
return new ArcGisMapServerImageryProvider({
    ...options,
    url: server,
    token: accessToken,
    credit: credit,
    usePreCachedTilesIfAvailable: true // ArcGIS Base Map Service Layers only support Tiled views
  });
```

packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js
```diff
@@ -398,6 +422,87 @@ function ArcGisMapServerImageryProvider(options) {
...
+        if (defined(defaultTokenCredit)) {
+          warningCredit = Credit.clone(defaultTokenCredit);
+        }
+      }
+      break;
+    default:
+  }
+
+  return new ArcGisMapServerImageryProvider({
+    options,
```
ggetz, line 499: I think you intended to use the spread operator in order to pass the provided options other than the overridden ones below.

```suggestion
    ...options,
```

ggetz (commented):
Thanks for the updates @Tamrat-B and @gowin20. I noticed a couple issues. Would yu be able to take a look?

ggetz:
@Tamrat-B and @gowin20 We will shortly begin the release process for 1.103. This still needs a bit of work, so I'd expect this to go into next month's release.

Tamrat-B:
@ggetz thanks for the review; will address the asks and also add the unit tests; I have few additional changes as well. ok on the schedule.

ggetz:
Hi @Tamrat-B, we just merged a larger-scale PR that affects some of the same areas of the codebase as this one. Another merge from `main` will be required, but I'm happy to help resolve some of those conflicts if needed.

Tamrat-B:
@ggetz updated pr based on reveiw and also updated tests to pass (thanks for the help earlier today!). Can you take a look and see if you can resolve the conflicts based on your latest changes... I sent you invite for write access to the repo. thx!

ggetz:
Thanks @Tamrat-B! I merged in the latest changes. Can you confirm everything is working as intended?

I'm noticing a few 404's for a specific tile when any ArcGIS basemap is set. Is this expected behavior?

Tamrat-B:
@ggetz Thanks! looks good. we just have to swap out the eval keys with yours when they are ready. But other than that looks good. (the 404s are happening when zooming beyond zoom level (request for tiles that don't exist)..the server could be configured to return empty/blank tiles for those & client app needs to handle it. This was not implemented so for now it's expected behavior.

ggetz (approved):

ggetz:
Thanks @Tamrat-B! I've updated this with our own generated token and added instructions for updating it each release. Looking forward to this going out on May 1st!

Issue #11107, opened 2023-02-23
Camera will go througth 3dtiles in zoom


jiangheng90:
The issuse is from #11062 , this morning I found camera will go throght buildings when camera is above 50m
I think this may comes from
```javascript
  const inertiaMovementApproachingGround = Math.abs(height) < 50;

  const needPickGlobe = inertiaMovement
    ? inertiaMovementApproachingGround
    : height < controller._minimumPickingTerrainHeight;
  if (needPickGlobe) {
    intersection = pickGlobe(controller, windowPosition, zoomCVIntersection);
  }
```
this condition will still fail when camera go through 3dtiles in a relative high altitude. so I think use the distance in previouse maybe more reliable.
fixed in pr #11108

ggetz:
Fixed in https://github.com/CesiumGS/cesium/pull/11108

PR #11108, opened 2023-02-23
fix camera go througth building in zoom

jiangheng90:
from issue #11107  fix camera go througth building in zoom by changing approaching condition

### DIFF
--- a/packages/engine/Source/Scene/ScreenSpaceCameraController.js
+++ b/packages/engine/Source/Scene/ScreenSpaceCameraController.js
@@ -224,6 +224,12 @@ function ScreenSpaceCameraController(scene) {
   this.minimumPickingTerrainHeight = 150000.0;
   this._minimumPickingTerrainHeight = this.minimumPickingTerrainHeight;
   /**
+   * The minimum distance the camera must be before testing for collision with terrain when zoom with inertia.
+   * @type {number}
+   * @default 4000.0
+   */
+  this.minimumPickingTerrainDistanceWithInertia = 4000.0;
+  /**
    * The minimum height the camera must be before testing for collision with terrain.
    * @type {number}
    * @default 15000.0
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/improvement-of-smooth-zoom/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Hi @jiangheng90, I believe zoom has always gone through 3D Tiles buildings. While we have checks in place for preventing the camera from going underground, there is currently no equivalent for 3D Tiles.

jiangheng90:
> Hi @jiangheng90, I believe zoom has always gone through 3D Tiles buildings. While we have checks in place for preventing the camera from going underground, there is currently no equivalent for 3D Tiles.

@ggetz emmm... maybe my description of this issuse is not exactly.
Not only 3dtiles, camera will go througth as well as it fly to high mountains such as everest.
Because of a inaccurate distance prediction in high mountains. the camera zoom behavior maybe weird. 
this behavior can be reproduced in sandcastle terrain expamle neat the mountain everest.
I'm not sure this is a correct behavior. so I update this part to solve it.
And I adjust the collision distance.Also I have test this in Terrain Exaggeration. it have being fine now.

packages/engine/Source/Scene/ScreenSpaceCameraController.js
```diff
@@ -2189,10 +2191,10 @@ function zoom3D(controller, startPosition, movement) {
     zoom3DCartographic
   ).height;
 
-  const inertiaMovementApproachingGround = Math.abs(height) < 50;
+  const approachingCollision = Math.abs(preIntersectionDistance) < 4000;
```
ggetz, line 2194: Since this is an arbitrary value it's probably best to expose this value as a property, like what is done for `controller.minimumPickingTerrainHeight`. I would call it something like `minimumPickingTerrainHeightWithInertia`.

ggetz:
Thanks @jiangheng90 for clarifying!

I just have the one comment.

jiangheng90:
> Thanks @jiangheng90 for clarifying!
> 
> I just have the one comment.

@ggetz already done and default value is 4000.0

Issue #10510, opened 2022-07-05
Big performance drop from v1.93 onward


cmcleese:
Sandcastle example:
[CodeSandbox Cesium v1.92](https://codesandbox.io/s/test-quasar-vue3-cesium-v1-92-0-e786s4?file=/src/pages/Index.vue)
[CodeSandbox Cesium v.193](https://codesandbox.io/s/test-quasar-vue3-cesium-v1-93-0-40p6zp?file=/src/pages/Index.vue)

Browser: Chrome 103.0.5060.66

Operating System: Win10

**Important:** Open codeSanbox link and select the `open in new window` button in the top of the preview window bar.  
Performance will not be noticeable unless open in new window.

Both cesium home locations in the sandboxes are set to the same location and camera view.  
All version from v1.93 and up seem to have the same performance drop.  Tested with 1.93, 1.94, 1.95 with same issue.
What used to be average 60fps is now 30fps.  

![image](https://user-images.githubusercontent.com/2061466/177287353-1675aed7-9bea-4607-8b74-9f2ed4c14615.png)
![image](https://user-images.githubusercontent.com/2061466/177287657-7d7948dc-c58c-4461-a1a4-dc4fdd7e198a.png)

cmcleese:
My initial observations is pointing to with the changes/additions in v1.93 of Improved rendering of ground and sky atmosphere. https://github.com/CesiumGS/cesium/pull/10063

cmcleese:
https://webglreport.com/ results:

![image](https://user-images.githubusercontent.com/2061466/177395349-3a4c01b8-61f0-4063-a3c8-b6bdaadded2b.png)
![image](https://user-images.githubusercontent.com/2061466/177395554-7c49b471-acfc-42ae-bf81-4cb2226834a2.png)

sanjeetsuhag:
I just tested on a device with a similar GPU and there is a performance hit: 54 FPS to 46 FPS average on a 1080p globe

![image](https://user-images.githubusercontent.com/5172619/177826854-b1871545-e45c-4870-85a0-b581781aa2be.png)

sanjeetsuhag:
There is more computation involved in the new atmosphere shader since it's a more physically accurate version. One option may be to expose the ray sampling numbers in the public API, allowing users to reduce the number, if needed, to help performance.

cmcleese:
Is there any push to improve the performance impact due to this new implementation of atmosphere?

glathoud:
[...]

PR #11109, opened 2023-02-23
Performance improvement

glathoud:
Context: https://github.com/CesiumGS/cesium/issues/10510#issuecomment-1438267847

### DIFF
--- a/packages/engine/Source/Shaders/AtmosphereCommon.glsl
+++ b/packages/engine/Source/Shaders/AtmosphereCommon.glsl
@@ -58,11 +58,9 @@ void computeScattering(
 
     // To deal with smaller values of PRIMARY_STEPS (e.g. 4)
     // we implement a split strategy: sky or horizon.
-    //
     // For performance reasons, instead of a if/else branch
     // a soft choice is implemented through a weight 0.0 <= w_stop_gt_lprl <= 1.0
     float x = 1e-7 * primaryRayAtmosphereIntersect.stop / length(primaryRayLength);
-    // w_stop_gt_lprl: similar to (1+tanh)/2
     // Value close to 0.0: close to the horizon
     // Value close to 1.0: above in the sky
     float w_stop_gt_lprl = 0.5 * (1.0 + approximateTanh(x));
@@ -73,27 +71,21 @@ void computeScattering(
     // The ray should end at the exit from the atmosphere or at the distance to the vertex, whichever is smaller.
     primaryRayAtmosphereIntersect.stop = min(primaryRayAtmosphereIntersect.stop, length(primaryRayLength));
 
-
-    // Distinguish inside or outside atmosphere (outer space)
-    // Reasons: 
-    // (1) from outer space we have to use the original implementation to get an unmodified/acceptable rendering
-    // (2) within atmosphere we need a speedup
+    // For the number of ray steps, distinguish inside or outside atmosphere (outer space)
+    // (1) from outer space we have to use more ray steps to get a realistic rendering
+    // (2) within atmosphere we need fewer steps for faster rendering
     float x_o_a = start_0 - ATMOSPHERE_THICKNESS; // ATMOSPHERE_THICKNESS used as an ad-hoc constant, no precise meaning here, only the order of magnitude matters
     float w_inside_atmosphere = 1.0 - 0.5 * (1.0 + approximateTanh(x_o_a));
-    
-    // (1) Outside of the atmosphere: original (16.0,4.0) values for good rendering
-    // (2) Inside atmosphere: smaller (4.0,2.0) values for a speedup
     int PRIMARY_STEPS = int(16.0 - w_inside_atmosphere * 12.0); // Number of times the ray from the camera to the world position (primary ray) is sampled.
-    int LIGHT_STEPS   = int(4.0 - w_inside_atmosphere * 2.0); // Number of times the light is sampled from the light source's intersection with the atmosphere to a sample position on the primary ray.
+    int LIGHT_STEPS = int(4.0 - w_inside_atmosphere * 2.0); // Number of times the light is sampled from the light source's intersection with the atmosphere to a sample position on the primary ray.
 
     // Setup for sampling positions along the ray - starting from the intersection with the outer ring of the atmosphere.
     float rayPositionLength = primaryRayAtmosphereIntersect.start;
-    // (1) Outside of the atmosphere: constant rayStepLength
-    // (2) Inside atmosphere: variable rayStepLength to compensate the originally rough rendering of the smaller (4.0,2.0) values
-    //     Implementation: sky vs horizon: constant step in one case, increasing step in the other case
-    float rayStepLengthIncrease = w_inside_atmosphere * ((1.0 - w_stop_gt_lprl)*(primaryRayAtmosphereIntersect.stop - primaryRayAtmosphereIntersect.start) / (float(PRIMARY_STEPS*(PRIMARY_STEPS+1))/2.0));
-    float rayStepLength         = max(1.0-w_inside_atmosphere, w_stop_gt_lprl) * (primaryRayAtmosphereIntersect.stop - primaryRayAtmosphereIntersect.start) / max(7.0*w_inside_atmosphere,float(PRIMARY_STEPS));
-
+    // (1) Outside the atmosphere: constant rayStepLength
+    // (2) Inside atmosphere: variable rayStepLength to compensate the rough rendering of the smaller number of ray steps
+    float totalRayLength = primaryRayAtmosphereIntersect.stop - rayPositionLength;
+    float rayStepLengthIncrease = w_inside_atmosphere * ((1.0 - w_stop_gt_lprl) * totalRayLength / (float(PRIMARY_STEPS * (PRIMARY_STEPS + 1)) / 2.0));
+    float rayStepLength = max(1.0 - w_inside_atmosphere, w_stop_gt_lprl) * totalRayLength / max(7.0 * w_inside_atmosphere, float(PRIMARY_STEPS));
 
     vec3 rayleighAccumulation = vec3(0.0);
     vec3 mieAccumulation = vec3(0.0);
@@ -104,7 +96,7 @@ void computeScattering(
     for (int i = 0; i < PRIMARY_STEPS; i++) {
         // Calculate sample position along viewpoint ray.
         vec3 samplePosition = primaryRay.origin + primaryRay.direction * (rayPositionLength + rayStepLength);
-        
+
         // Calculate height of sample position above ellipsoid.
         float sampleHeight = length(samplePosition) - atmosphereInnerRadius;
 
@@ -115,7 +107,7 @@ void computeScattering(
         // Generate ray from the sample position segment to the light source, up to the outer ring of the atmosphere.
         czm_ray lightRay = czm_ray(samplePosition, lightDirection);
         czm_raySegment lightRayAtmosphereIntersect = czm_raySphereIntersectionInterval(lightRay, origin, atmosphereOuterRadius);
-        
+
         float lightStepLength = lightRayAtmosphereIntersect.stop / float(LIGHT_STEPS);
         float lightPositionLength = 0.0;
 
@@ -145,9 +137,7 @@ void computeScattering(
         mieAccumulation += sampleDensity.y * attenuation;
 
         // Increment distance on primary ray.
-        
-        // --- Modified: increasing step length
-        rayPositionLength += (rayStepLength+=rayStepLengthIncrease);
+        rayPositionLength += (rayStepLength += rayStepLengthIncrease);
     }
 
     // Compute the scattering amount.
--- a/packages/engine/Source/Shaders/AtmosphereCommon.glsl
+++ b/packages/engine/Source/Shaders/AtmosphereCommon.glsl
@@ -93,7 +93,13 @@ void computeScattering(
     vec2 heightScale = vec2(u_atmosphereRayleighScaleHeight, u_atmosphereMieScaleHeight);
 
     // Sample positions on the primary ray.
-    for (int i = 0; i < PRIMARY_STEPS; i++) {
+    for (int i = 0; i < 999999999; ++i) {
+
+        // The loop should be: for (int i = 0; i < PRIMARY_STEPS; ++i) {...} but WebGL1 cannot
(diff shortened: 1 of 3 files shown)


cesium-concierge:
Thank you so much for the pull request @glathoud! I noticed this is your first pull request and I wanted to say welcome to the Cesium community!

    * Please add yourself to the [contributors](https://github.com/glathoud/cesium_1.102/blob/glathoud-patch-1-atmosphere-performance/CONTRIBUTORS.md) file!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/glathoud-patch-1-atmosphere-performance/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the PR @glathoud! We'll give this a review shortly!

ggetz:
Hi @glathoud, I pulled down this branch the test out the changes you made, but I'm seeing an error compiling the shader on load:
` [Cesium WebGL] Fragment shader compile log: ERROR: 0:77: 'rayStepLength' : redefinition`

Would you mind taking a look?

glathoud:
> Hi @glathoud, I pulled down this branch the test out the changes you made, but I'm seeing an error compiling the shader on load: ` [Cesium WebGL] Fragment shader compile log: ERROR: 0:77: 'rayStepLength' : redefinition`
> 
> Would you mind taking a look?

@ggetz Sure, just committed a fix.

ggetz:
Thanks @glathoud!

Unfortunately, I'm seeing a significant downgrade in the quality of the ground atmosphere (what the ground looks like from space), even after restoring the constants to 

```
const int PRIMARY_STEPS = 16; 
const int LIGHT_STEPS = 4;
```

| `main` | branch |
| --- | --- |
| image | image  |

The sky atmosphere appears comparable though, which is good.

Any chance you would be able to address the ground atmosphere appearance? The performance improvements are valuable, but we'd be hesitant to merge them with a downgrade in visual quality.

glathoud:
> The sky atmosphere appears comparable though, which is good.
> 
> Any chance you would be able to address the ground atmosphere appearance? The performance improvements are valuable, but we'd be hesitant to merge them with a downgrade in visual quality.

@ggetz Sure, I'll give it a try. By decreasing priority:
 * fix the ground atmosphere appearance for (4,2) while keeping good performance.
 * maybe also ensure that (16,4) gives the same appearance as with the original implementation, or very very close.

glathoud:
@ggetz Just commited. Now the planet should look good from outer space, basically using the original implementation in that case (incl. constants 16.0,4.0), while still activating the performance improvement when close to the ground.

Feel free to test!

glathoud:
This is what I get locally:

original implementation: 

localhost_4443_src_cesium-1 102_Apps_Sandcastle_

modified implementation:

localhost_4443_src_cesium-1 102_Apps_Sandcastle_ (1)

and to explain the choice of (16,4) when viewing the planet from in outer space, here is what the original implementation with lower values (4,2) would give:

localhost_4443_src_cesium-1 102_Apps_Sandcastle_ (2)

ggetz:
Awesome, thanks @glathoud! This is looking great, and I just check ed that all the existing atmosphere parameters continue to work as expected. 

@lilleyse Would you be able to do a pass on the shader code for a sanity check?

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -50,14 +48,46 @@ void computeScattering(
...
+    // we implement a split strategy: sky or horizon.
+    //
+    // For performance reasons, instead of a if/else branch
+    // a soft choice is implemented through a weight 0.0 <= w_stop_gt_lprl <= 1.0
+    float lprl = length(primaryRayLength);
+    float x = 1e-7 * primaryRayAtmosphereIntersect.stop / lprl;
+    // w_stop_gt_lprl: similar to (1+tanh)/2
+    // Value close to 0.0: close to the horizon
+    // Value close to 1.0: above in the sky
+    float w_stop_gt_lprl = max(0.0,min(1.0,(1.0 + x * ( 27.0 + x * x ) / ( 27.0 + 9.0 * x * x ))/2.0));
```
jjhembd, line 61: Can the `tanh` approximation be pulled out into a separate function? Something like:
```glsl
/**
 * Rational approximation to tanh(x)
*/
float approximateTanh(float x) {
    return x * (27.0 + x * x) / (27.0 + 9.0 * x * x);
}
``` 

I think it would make some of these long lines a little easier to read.

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -50,14 +48,46 @@ void computeScattering(
...
+    
+    // (1) Outside of the atmosphere: original (16.0,4.0) values for good rendering
+    // (2) Inside atmosphere: smaller (4.0,2.0) values for a speedup
+    int PRIMARY_STEPS = int(16.0 - w_inside_atmosphere * 12.0); // Number of times the ray from the camera to the world position (primary ray) is sampled.
+    int LIGHT_STEPS   = int(4.0 - w_inside_atmosphere * 2.0); // Number of times the light is sampled from the light source's intersection with the atmosphere to a sample position on the primary ray.
 
     // Setup for sampling positions along the ray - starting from the intersection with the outer ring of the atmosphere.
-    float rayStepLength = (primaryRayAtmosphereIntersect.stop - primaryRayAtmosphereIntersect.start) / float(PRIMARY_STEPS);
     float rayPositionLength = primaryRayAtmosphereIntersect.start;
+    // (1) Outside of the atmosphere: original implementation
```
jjhembd, line 84: Once this PR is merged, someone reading the code will not know what the "original implementation" was. These comments could simply state what is different in each case. For example:

```glsl
// (1) Outside the atmosphere: constant rayStepLength
```

jjhembd (commented):
Hi @glathoud, thanks for this PR! The performance improvement is very nice to see.

I left a few comments for the GLSL style.

packages/engine/Source/Shaders/AtmosphereCommon.glsl
glathoud: Yes. I carefuly tested the performance implications, tried various options, and ended up with the one in https://github.com/CesiumGS/cesium/pull/11109/commits/2290ec24291d9d977492cf2fedbf24e68d81120d  : 
```c
/**
 * Rational approximation to tanh(x)
*/
float approximateTanh(float x) {
    float x2 = x * x;
    return max(-1.0, min(+1.0, x * (27.0 + x2) / (27.0 + 9.0 * x2)));
}
```

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -50,14 +48,46 @@ void computeScattering(
...
-    primaryRayAtmosphereIntersect.stop = min(primaryRayAtmosphereIntersect.stop, length(primaryRayLength));
+    primaryRayAtmosphereIntersect.stop = min(primaryRayAtmosphereIntersect.stop, lprl);
+
+
+    // Distinguish inside or outside atmosphere (outer space)
+    // Reasons: 
+    // (1) from outer space we have to use the original implementation to get an unmodified/acceptable rendering
+    // (2) within atmosphere we need a speedup
+    float x_o_a = start_0 - ATMOSPHERE_THICKNESS; // ATMOSPHERE_THICKNESS used as an ad-hoc constant, no precise meaning here, only the order of magnitude matters
+    float w_inside_atmosphere = 1.0 - max(0.0,min(1.0,(1.0 + x_o_a * ( 27.0 + x_o_a * x_o_a ) / ( 27.0 + 9.0 * x_o_a * x_o_a ))/2.0)); // similar to 1.0 - (1+tanh)/2
```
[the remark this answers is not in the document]
glathoud: Done in https://github.com/CesiumGS/cesium/commit/2290ec24291d9d977492cf2fedbf24e68d81120d , hope this is ok now.

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -50,14 +48,46 @@ void computeScattering(
         return;
     }
 
+    // To deal with smaller values of PRIMARY_STEPS (e.g. 4)
+    // we implement a split strategy: sky or horizon.
+    //
+    // For performance reasons, instead of a if/else branch
+    // a soft choice is implemented through a weight 0.0 <= w_stop_gt_lprl <= 1.0
+    float lprl = length(primaryRayLength);
```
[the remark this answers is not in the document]
glathoud: Ok I removed it in https://github.com/CesiumGS/cesium/commit/2290ec24291d9d977492cf2fedbf24e68d81120d and put `lengh(primaryRayLength)` in the two places where it is used. No performance loss.

jjhembd:
Hi @glathoud, there was still some non-standard whitespace, and a few remaining references to the "original" implementation. I just pushed a quick edit to fix these. Please check my changes to confirm I didn't change anything about your algorithm.

One thing we will need to resolve: the new algorithm will not work for a user running in a WebGL1 context. Most devices now support WebGL2, which is why WebGL2 is now [enabled by default](https://github.com/CesiumGS/cesium/pull/10894). But we are still supporting WebGL1, especially for users wanting to run old Custom Shader code, which may still be in WebGL1.

You can test in a WebGL1 context by adding the following lines to the Viewer constructor.
```javascript
  contextOptions: { requestWebgl1: true },
});
```

A WebGL1 context gives me the following error with your code:
image

This is because the loop conditions `PRIMARY_STEPS` and `LIGHT_STEPS` are no longer constant values, fixed at compile time.

We have worked around this limitation in the past by using a large fixed value for the loop condition, and then breaking out of the loop based on a non-constant condition. See for example [lines 70-73 in IntersectionUtils.glsl](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Shaders/Voxels/IntersectionUtils.glsl#:~:text=for%20(int%20i,%3E%3D%20n)%20%7B%20break%3B%20%7D).

Could you try this trick in AtmosphereCommon.glsl?

glathoud:
Hello @jjhembd , thanks for the commit. I checked, it looks fine. Still tested as well for safety, result fine as well.

I also just commited WebGL1 support, as you described.

ggetz:
@glathoud Could you please update `CHANGES.md` with a summary of the changes here?

@jjhembd Is there any additional feedback, or is this PR good to go?

glathoud:
> @glathoud Could you please update `CHANGES.md` with a summary of the changes here?

@ggetz Done, feel free to adjust/format.

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -50,25 +56,53 @@ void computeScattering(
...
+    float rayStepLength = max(1.0 - w_inside_atmosphere, w_stop_gt_lprl) * totalRayLength / max(7.0 * w_inside_atmosphere, float(PRIMARY_STEPS));
 
     vec3 rayleighAccumulation = vec3(0.0);
     vec3 mieAccumulation = vec3(0.0);
     vec2 opticalDepth = vec2(0.0);
     vec2 heightScale = vec2(u_atmosphereRayleighScaleHeight, u_atmosphereMieScaleHeight);
 
     // Sample positions on the primary ray.
-    for (int i = 0; i < PRIMARY_STEPS; i++) {
+    for (int i = 0; i < 999999999; ++i) {
```
jjhembd, line 96: This works on my machine, but I believe some WebGL1 compilers will [unroll the loop](https://en.wikipedia.org/wiki/Loop_unrolling). Can we use a smarter limit here? I think we should never need more than 16 steps.

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -50,25 +56,53 @@ void computeScattering(
...
+
+        // The loop should be: for (int i = 0; i < PRIMARY_STEPS; ++i) {...} but WebGL1 cannot
+        // loop with non-constant condition, so it has to break early instead
+        if (i >= PRIMARY_STEPS)
+            break;
```
jjhembd, line 101: We follow the default rule for one-line if blocks. See https://eslint.org/docs/latest/rules/curly

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -79,14 +113,19 @@ void computeScattering(
...
-        
+
         float lightStepLength = lightRayAtmosphereIntersect.stop / float(LIGHT_STEPS);
         float lightPositionLength = 0.0;
 
         vec2 lightOpticalDepth = vec2(0.0);
 
         // Sample positions along the light ray, to accumulate incidence of light on the latest sample segment.
-        for (int j = 0; j < LIGHT_STEPS; j++) {
+        for (int j = 0; j < 999999999; ++j) {
```
jjhembd, line 123: I think the limit here could be 4

packages/engine/Source/Shaders/AtmosphereCommon.glsl
```diff
@@ -79,14 +113,19 @@ void computeScattering(
...
+
+            // The loop should be: for (int j = 0; i < LIGHT_STEPS; ++j) {...} but WebGL1 cannot
+            // loop with non-constant condition, so it has to break early instead
+            if (j >= LIGHT_STEPS)
+                break;
```
jjhembd, line 128: This statement should also have braces around it

jjhembd (commented):
Hi @glathoud, sorry for the delayed review. I think this is very close--I just pointed out 2 more things, and then it should be ready to merge!

glathoud:
@jjhembd Done with the syntax adjustments, feel free to have a look at it.

glathoud:
FYI, I still checked the speed on my hardest use case so far. Recap:
 * Before the improvement: 33.30 fps, 37.53 fps after a few optims
 * After: 51.50 fps

ggetz:
Awesome, thanks @glathoud! Great to see that we can improve the performance without a noticeable hit to the visual quality, and thanks for being responsive throughout the review process!

ggetz:
(The CI failure here is unrelated to this PR).

glathoud:
> Awesome, thanks @glathoud! Great to see that we can improve the performance without a noticeable hit to the visual quality, and thanks for being responsive throughout the review process!

You're welcome. It was a pleasure to work with your team.

Issue #11130, opened 2023-03-01
CDN version for Cesium `1.103.0` unavailable


turansky:
Cesium example - https://cesium.com/learn/cesiumjs-learn/cesiumjs-quickstart/#import-from-cdn

### Test case
1. Open CDN url from example - https://cesium.com/downloads/cesiumjs/releases/1.103/Build/Cesium/Cesium.js

### Expected
1. `Cesium.js` content

### Actual
1. Page 404

<img width="1464" alt="Screenshot 2023-03-01 at 11 03 54 PM" src="https://user-images.githubusercontent.com/9942723/222264623-a7b43975-8ddc-4ab4-8f99-3d8c0a85d5ad.png">

ggetz:
Thanks for the report @turansky, we are looking into this now!

ggetz:
This should now be fixed on the website. A fix for the publish step is in https://github.com/CesiumGS/cesium/pull/11131. Thanks again!

PR #11131, opened 2023-03-01
Fix release publish step

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11130

This is now fixed on the website itself, but deploying our Cesium releases was broken after the AWS SDK updates.

The error code are no longer what we expected before. Instead I followed [parsing the errors as recommended in the docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html#troubleshooting), though notably the documentation is not correct for the particular error we need to handle– the property is `$metadata` instead of `$$metadata`

We didn't catch this the first time around because the file already existed, and therefor didn't throw an error.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1088,29 +1088,32 @@ async function deployCesiumRelease(bucketName, s3Client, errors) {
       `Cesium version ${release.tag} up to date. Skipping release deployment.`
     );
   } catch (error) {
-    // The current version is not uploaded
-    if (error.code === "NotFound") {
-      console.log("Updating cesium version...");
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/release-deployment/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Makes sense, thanks @ggetz!

PR #11158, opened 2023-03-15
Fix description of OrientedBoundingBox cube.

kring:
The doc for the halfAxes parameter to `OrientedBoundingBox` previously said:

> The three orthogonal half-axes of the bounding box. Equivalently, the transformation matrix, to rotate and scale a 1x1x1 cube centered at the origin.

But this isn't true. It's a 2x2x2 cube because the coordinates of each corner are (±1, ±1, ±1), making the length of each side 2. I fixed the doc and added a test that demonstrates this.


ggetz:
Makes sense. Thanks @kring!

PR #11173, opened 2023-03-20
Deprecate Model and Cesium3DTileset readyPromise

ggetz:
Part of https://github.com/CesiumGS/cesium/issues/10909 - To be merged into the staging branch

Deprecates `readyPromise` pattern through the Model and Cesium3DTileset APIs and related functions. Added async factory functions for object creation, and events for tracking render-loop time errors or other events.

The tileset API now looks like this:
````js
  try {
    const tileset = await Cesium.Cesium3DTileset.fromUrl(url);
  } catch (error) {
    console.log(`Failed to load tileset: ${error}`);
  }
````
 Notably, `fromUrl` loads the tileset JSON and creates the tileset tree of tile metadata, but does not actually load any tiles until added to the scene and rendered.
 
The model API now looks like this:
```js
    try {
        url: "../../SampleData/models/CesiumMan/Cesium_Man.glb",
      });
        // model is ready for rendering
      });
    } catch (error) {
      console.log(`Failed to load model. ${error}`);
    }
 ```

This includes several under-the-hood changes, most notably to the loaders responsible for model and related resource loading. At a high level, this involves:
 - Moving resource fetches to async factory functions– which can be called up front and verified before continuing with any processing
 - All other frameState-dependent gl resource created to synchronous calls via `update` or related function (usually `process` in loader terms)
     - This means errors are thrown in synchronously during the render loop, where the appropriate object can catch and report the errors as events. By default, any error events are logged to the console.
     - This means that several internal promises that span across rendered frames to be removed, such as the loader `promise` properties `Cesium3DTile.contentReadyPromises` and `Ceisum3DTile.contentReadyToProcessPromise`
     
 ----

One unfortunate API issue here is that Cesium3DTileContent is public, but each of its implementers are private. While this means we should technically be able to remove content readyPromise properties without warning, they are likely being used despite being private API. I think the best thing to do here is to deprecate the promises anyway. Then, and this may be out of scope of the ready promise changes, we should probably check the approach of private/public API around Cesium3DTileContent

This PR also 
- Explained API updates in `CHANGES.md`. I added a top-level "Major announcement" bullet to summarize changes.
- Updates several guides to match new APIs– Most notably the testing guide was updated to explain the now preferred method of testing promises and async functions.
- Updates Sandcastles to match new APIs
- Updates unit tests. Due to some flakey tests that came to a head after the API updates, there were further changes needed beyond the API changes. In just about every case, any background (or foreground!) asynchronous behavior was not being properly awaited. This includes cases where the tests were never actually running all expectations, leading to false positives.

TODOs:
- [x] Add test for fromIonAssetId
- [x] Clipping planes modelMatrix issue in Sandcastle
- [x] CDB Terrain tileset issue in Sandcastle
- [x] Fix CI test failure
- [x] Deprecate `Cesium3DTileContent`-inherted `readyPromise`
- [x] Address Cesium3DTileContent public/private breakdown in separate issue -> Opened https://github.com/CesiumGS/cesium/issues/11191
- [x] 3D Tiles Terrain Classification Sandcastle issue

### CODE REVIEW

File: Apps/Sandcastle/gallery/MSAA.html [hunk 1]
```diff
@@ -79,6 +79,26 @@
           viewer.scene.camera.lookAt(target, offset);
         }
 
+        let currentAssetId;
+        async function createTileset(assetId) {
```
jjhembd, line 83: Do we also need to write an async version of the `createModel` function above? (line 48, 67... sorry but https://github.com/github/roadmap/issues/347)

jjhembd (commented):
Thanks @ggetz, I did a first pass through the READMEs and the Sandcastles, to make sure I understand the public API changes. It makes sense overall. I left 2 minor comments.

I'll look at the main code changes in another pass.

Apps/Sandcastle/gallery/MSAA.html
ggetz: Nope, model entities handle the async model primitive creation through modelVisualizer, and the error is logged. This is similar for other visualizers which already exist.

jjhembd (commented):
@ggetz, I have gone through most of the specs and about half the code. I left a few minor comments

packages/engine/Specs/Scene/Model/ModelSpec.js
```diff
@@ -701,57 +804,52 @@ describe(
...
-          spyOn(decoder, "scheduleTask").and.callFake(function () {
-            return Promise.reject({ message: "Custom error" });
-          });
 
-          const model = scene.primitives.add(
-            Model.fromGltf({
-              url: dracoCesiumManUrl,
-            })
-          );
+      const decoder = DracoLoader._getDecoderTaskProcessor();
```
jjhembd, line 820: It looks like `DracoLoader._getDecoderTaskProcessor()` should be converted to an async function. (Maybe not for this PR)

jjhembd (commented):
Hi @ggetz, I finished going through the rest of the specs and code. It is a very nice cleanup overall. Some of the bigger diffs that I was avoiding till the end actually made the most sense (like `ResourceCache`). I left a few mostly minor comments.

Overall, there is one recurring pattern that I think reduces readability:
```javascript
this._promise = (async () => {
  try {
    const asyncResult = await someAsyncFunction();
    // ... do something with asyncResult
  } catch {
    // ... handle errors
  }
})();
```

This pattern appears in 5 of the `Gltf*Loader`s. 

I like the `await` syntax in most casese. But when we have to wrap it in an IIFE, it becomes a lot harder to read. I think these would be better as:

```javascript
this._promise = someAsyncFunction()
  .then((asyncResult => {
    // ... do something with asyncResult
  })
  .catch((error) => {
    // ... handle errors
  });
```

Especially when we are assigning to something named "promise", the old `.then` syntax seems like a good fit.

packages/engine/Source/Scene/GltfIndexBufferLoader.js
```diff
@@ -397,7 +324,63 @@ GltfIndexBufferLoader.prototype.process = function (frameState) {
...
+        this._typedArray = typedArray;
+        // The index datatype may be a smaller datatype after draco decode
+        this._indexDatatype = ComponentDatatype.fromTypedArray(typedArray);
+      }
+    } catch (error) {
+      handleError(this, error);
+    }
+  }
+
+  if (!defined(typedArray)) {
```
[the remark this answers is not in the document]
ggetz: Yes, that should replace the outer-scoped `typedArray`. I made these the same scope for clarity.

packages/engine/Specs/Scene/Model/ModelSpec.js
ggetz: I agree- Though these PRs are large, and this shouldn't be a significant API change so it doesn't need to go out timed with this batch of deprecations.

ggetz:
Thanks for the review @jjhembd! I believe I've responded to all your feedback and addressed all TODO's for this PR.

jjhembd:
Looks good @ggetz! Merging now

Issue #10905, opened 2022-11-15
Inconsistent info box in 3D Tiles Feature Picking Sandcastle


javagl:
The "3D Tiles Feature Picking" sandcastle has a large `if`-statement for different highlighting modes depending on `isSilhouetteSupported`. In both cases, an "info box" with information about the picked feature should be displayed. The contents of this box is different in both cases: 

- https://github.com/CesiumGS/cesium/blob/0ad93508edfa19b84fb7e1413450c92e054d8af1/Apps/Sandcastle/gallery/3D%20Tiles%20Feature%20Picking.html#L174
- https://github.com/CesiumGS/cesium/blob/0ad93508edfa19b84fb7e1413450c92e054d8af1/Apps/Sandcastle/gallery/3D%20Tiles%20Feature%20Picking.html#L273

This should probably be moved into a `setFeatureInfoboxDescription` function that shows the same information for both cases.

---

(Subjectively: Yes, silhouettes look nice. But maybe the sandcastle could be made more concise by focussing less on the silhouette rendering, and more on the feature picking, because that's what is supposed to be ... _highlighted_ there)

PR #11180, opened 2023-03-21
Cleanups for 3D Tiles Feature Picking sandcastle

javagl:
Fixes https://github.com/CesiumGS/cesium/issues/10905 

Further issues that are fixed here:

- At some places, `pickedFeature.getProperty("name")` was called, which is always `undefined` for the features here
- Specifically: In the non-silhouette-case, the overlay displayed this property when hovering a building - causing an empty overlay (small black square) to be displayed!
- At some places, `pickedFeature.getProperty("longitude")` was called, which must be `pickedFeature.getProperty("Longitude")` (capitalized) - otherwise, the info box showed a bunch of `undefined`s.
- At some places, the comments used `"it's"` (contraction) where they should have used `"its"` (possessive noun)

---

Note: A "pragmatic" way to check whether the non-silhouette case works is to change
`if (Cesium.PostProcessStageLibrary.isSilhouetteSupported(viewer.scene)) {`
to 
`if (false && Cesium.PostProcessStageLibrary.isSilhouetteSupported(viewer.scene)) {` 
when running it locally...

### DIFF
--- a/Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
+++ b/Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
@@ -33,6 +33,7 @@
       window.startup = function (Cesium) {
         "use strict";
         //Sandcastle_Begin
+
         // A simple demo of 3D Tiles feature picking with hover and select behavior
         // Building data courtesy of NYC OpenData portal: http://www1.nyc.gov/site/doitt/initiatives/3d-building.page
         const viewer = new Cesium.Viewer("cesiumContainer", {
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/feature-picking-sandcastle-fixes/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the updates here @javagl! Overall, much cleaner and thanks for catching those other issues as well.

A bit out of scope for this PR, but I do generally agree with your statement in https://github.com/CesiumGS/cesium/issues/10905 that it would be more straightforward and educational to focus on either picking or silhouettes. We do have hopes of soon taking a pass over all the Sandcastle examples with an eye towards creating a learning path that easier to follow.

PR #11183, opened 2023-03-24
Fix example in CustomShaderGuide

jjhembd:
Fixes the problems in the CustomShaderGuide, as reported [on the forum](https://community.cesium.com/t/customshader-upgrading-cesium/22652).

### DIFF
--- a/Documentation/CustomShaderGuide/README.md
+++ b/Documentation/CustomShaderGuide/README.md
@@ -143,12 +143,12 @@ The user is responsible for assigning a value to this varying in
 const customShader = new Cesium.CustomShader({
   // Varying is declared here
   varyings: {
-    v_selectedColor: VaryingType.VEC3,
+    v_selectedColor: Cesium.VaryingType.VEC4,
   },
   // User assigns the varying in the vertex shader
   vertexShaderText: `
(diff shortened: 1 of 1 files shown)


javagl:
Coincidentally, I had some TODO on my list for a few months now, and just wanted to do this, but ... noticed that `custom-shader-docs` branch: This might be a good place to sneak in an update for the links that say "documentation comment" at https://github.com/CesiumGS/cesium/blob/4a74622a0972f2694e3134715473e88d7b7a4cb9/Documentation/CustomShaderGuide/README.md#czm_modelvertexoutput-struct and the next section, which lead to a 404 right now due to the engines/widgets directory structure.

The first link should point to https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Shaders/Builtin/Structs/modelVertexOutput.glsl
The second link should point to https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Shaders/Builtin/Structs/modelMaterial.glsl

(And it would be nice if they could be a permalinks, but .... hmmm)

If this is merged before this change, I'll create a new PR a bit later.

jjhembd:
> This might be a good place to sneak in an update for the links

Thanks for catching that @javagl! I just pushed a fix for those 2 links

ggetz:
Thanks @jjhembd and @javagl!

PR #11184, opened 2023-03-24
GroundPolyline BoundingSphere fix?

malaretv:
We work a lot with cesium on terrain datasets with fairly extreme values compared to Earth. For example, lunar terrain is roughly between -11000 and 10000 meters.

We've been trying to get GroundPolyline to work reliably on these terrain providers but even after generating a custom `approximateTerrainHeights.json` we were still seeing issues with the polylines disappearing. Upon investigation it was because the bounding spheres weren't stretching to encompass the actual geometry on the terrain.

Within the GroundPolylineGeometry source `sumHeights` is used only in one place below
https://github.com/CesiumGS/cesium/blob/4244bc02def1c931808dbeb7bd5c748218c31046/packages/engine/Source/Core/GroundPolylineGeometry.js#L1590-L1591
 It seems to me maybe the purpose of `sumHeights` is really more to get the average height distance from the ellipsoid across segments, and then using that value to adjust the bounding sphere radius accordingly.
 
Ensuring we're getting the absolute value of minimum and maximum heights when calculating sumHeights seems to get us better behavior on terrain datasets with large negative min elevation values such as the moon.


cesium-concierge:
Thanks for the pull request @malaretv!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/groundpolyline-boundingspheres/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks again @malaretv. Would you mind:

1. Adding a short inline comment to explain why `Math.abs` value is being used? It's a bit unintuitive without the explanation
2. Add a unit tests to ensure this is working with terrain values causing the issue
3. Update `CHANGES.md` with a summary of the fix?

malaretv:
will do. thanks @ggetz for taking a look!

ggetz:
Hi @malaretv, what's the status on this PR?

malaretv:
Thanks for pinging me on this @ggetz

Today had free time to wrap up the remaining items. Let me know if anything else remains.

ggetz:
All looks good, thanks @malaretv!

Issue #11185, opened 2023-03-24
CZMLDataSource with custom ellipsoids


malaretv:
CZMLDataSource has a number of spots where Ellipsoid.WGS84 is assumed and it makes the datasource unusable for non-earth maps.

One example is here where Ellipsoid.WGS84 is explicitly called
https://github.com/CesiumGS/cesium/blob/4244bc02def1c931808dbeb7bd5c748218c31046/packages/engine/Source/DataSources/CzmlDataSource.js#L275-L286

And also like here where the absence of an ellipsoid parameter to `Cartesian3.fromDegreesArrayHeights()` causes WGS84 to be assumed internally.
https://github.com/CesiumGS/cesium/blob/4244bc02def1c931808dbeb7bd5c748218c31046/packages/engine/Source/DataSources/CzmlDataSource.js#L1828-L1830

I am happy to provide a fix for this but was wondering what the team thinks is the best path to add this option?

I could add an `ellipsoid` to be passed in with the LoadOptions with each `new CZMLDataSource()` call. e.g.
```js
let datasource = new CZMLDataSource(url, {ellipsoid: myCustomEllipsoid})
```
This seems nicest in terms of API but implementation-wise would mean passing the `ellipsoid` value through all the individual nested `process*()`  calls.

The other option is the default ellipsoid could exist as a static property on CZMLDataSource. All internal places an ellipsoid is required would use this static property. The user code to then use a different ellipsoid could be something like this:
```js
CZMLDataSource.DEFAULT_ELLIPSOID = myCustomEllipsoid;
let datasource = new CZMLDataSource(url)
```
Much simpler to implement but not great if multiple ellipsoid values are desired within an application -- I imagine that case is pretty rare?

thoughts?

Issue #3543, opened 2016-02-10
Make Entity API use custom ellipsoid


hpinkos:
It looks like the Entity API only uses the default Ellipsoid.WGS84.  As a result, you can't really use it if you need a custom ellipsoid.  In this example, the polygon is hovering above the globe because the ellipsoid is smaller than earth:

``` javascript
var radius = 5e6;
var ellipsoid = new Cesium.Ellipsoid(radius, radius, radius);
var projection = new Cesium.GeographicProjection(ellipsoid);

var viewer = new Cesium.Viewer('cesiumContainer', {
    mapProjection: projection,
    baseLayerPicker: false,
    imageryProvider: new Cesium.BingMapsImageryProvider({
        url : '//dev.virtualearth.net',
        mapStyle : Cesium.BingMapsStyle.AERIAL,
        ellipsoid: ellipsoid
    })
});

var redPolygon = viewer.entities.add({
    name : 'Red polygon on surface',
    polygon : {
        hierarchy : Cesium.Cartesian3.fromDegreesArray([-115.0, 37.0,
                                                        -115.0, 32.0,
                                                        -107.0, 33.0,
                                                        -102.0, 31.0,
                                                        -102.0, 35.0]),
        material : Cesium.Color.RED
    }
});
```

mramato:
The (simply by hacky) workaround for this is to actually modify `WGS84` at the start of your app so that it has the custom size, then just use the default everywhere.

It might actually be a decent real solution to have `Ellipsoid.default` in addition to `Ellipsoid.WGS84` and just have the former equal the latter by default.  Then we use `Ellipsoid.default` everywhere we currently use `Ellipsoid.WGS84`.

This seems way more elegant of a solution when you only have a single custom ellipsoid in an application.

mramato:
The actual problem here is that all of the `XXXGeometryUpdater` objects use the default ellipsoid value rather than getting it from the scene.  Should be a straightforward fix, but I don't have the bandwidth to do it myself right now.

ggetz:
Also brought up for CZML in https://github.com/CesiumGS/cesium/issues/11185.

PR #11190, opened 2023-03-28
Be consistent about using Ellipsoid.WGS84 in CzmlDataSource

malaretv:
The current recommended path by maintainers to support non-wgs84 ellipsoids is to modify the Ellipsoid.WGS84 (see https://github.com/CesiumGS/cesium/issues/11185#issuecomment-1486882571 and https://github.com/CesiumGS/cesium/issues/3543#issuecomment-182451689)
For CzmlDataSource this approach wasn't working because of the existence of Cartesian3 functions that default to other hardcoded WGS84 values if no ellipsoid is provided.

This change makes sure `Ellipsoid.WS84` is explicitly passed to those function calls in CzmlDataSource.


cesium-concierge:
Thanks for the pull request @malaretv!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/czml-wgs84/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

malaretv:
Added tests that run successfully locally.
Not sure about the Travis CI errors
Screenshot 2023-03-31 at 11 50 17 AM

They seem unrelated to the tests

ggetz:
> Not sure about the Travis CI errors

We've seen this in this repo as well as others. The `spin` failure is from travis CI itself, not our CI scripts. The failure shouldn't affect the build state.

The failure in CI was due to a separate configuration change on our part, which has since been fixed. I've restarted the build.

ggetz:
Thanks for the PR @malaretv!

While this does certainly work _around_ the issue when setting `Ellipsoid.WGS84` to another value, the fix itself stops a bit short of remedying the situation such that its obvious via the API how this works with custom ellipsoids.

Consider this comment from #3543:

> It might actually be a decent real solution to have Ellipsoid.default in addition to Ellipsoid.WGS84 and just have the former equal the latter by default. Then we use Ellipsoid.default everywhere we currently use Ellipsoid.WGS84.

Would you have any interest in adding `Ellipsoid.default` such that it can be used throughout the API in this manner?

malaretv:
> Thanks for the PR @malaretv!
> 
> While this does certainly work _around_ the issue when setting `Ellipsoid.WGS84` to another value, the fix itself stops a bit short of remedying the situation such that its obvious via the API how this works with custom ellipsoids.
> 
> Consider this comment from #3543:
> 
> > It might actually be a decent real solution to have Ellipsoid.default in addition to Ellipsoid.WGS84 and just have the former equal the latter by default. Then we use Ellipsoid.default everywhere we currently use Ellipsoid.WGS84.
> 
> Would you have any interest in adding `Ellipsoid.default` such that it can be used throughout the API in this manner?

Certainly happy to at least take a look later this week and see the scale of work involved in adding `Ellipsoid.default`! This would definitely be a very helpful change for our usage of Cesium. Two big questions I have off the top of my head with a `Ellipsoid.default` solution are:

1) Will it work with Workers? Would need to go through the worker code and make sure there is nothing that could fall back to using Ellipsoid.default -- since Workers bundle their own subsets of Cesium any overriding of `Ellipsoid.default` would not translate to that environment.

2) There are other locations where there may be hardcoded WGS84 defaults. One certain case is in `Cartesian3` below:
https://github.com/CesiumGS/cesium/blob/9295450e64c3077d96ad579012068ea05f97842c/packages/engine/Source/Core/Cartesian3.js#L895-L899

 
FWIW, this pull request would still be required even with an `Ellipsoid.default` implementation -- I would just need to replace `Ellipsoid.WGS84` with `Ellipsoid.default`. The key thing about this pull request is being explicit about the Ellipsoid used when calling `Carteisan.fromRadians()` such that it doesn't fall back to the hardcoded values in `Cartesian3` for `wgs84RadiiSquared`

ggetz:
> Certainly happy to at least take a look later this week and see the scale of work involved in adding Ellipsoid.default! This would definitely be a very helpful change for our usage of Cesium.

Understood! We would appreciate the contribution if you have the bandwidth! This would go a long way to improve usability and function for all those using CesiumJS for non-Earth cases.

> Will it work with Workers? Would need to go through the worker code and make sure there is nothing that could fall back to using Ellipsoid.default -- since Workers bundle their own subsets of Cesium any overriding of Ellipsoid.default would not translate to that environment.

Good point. One solution would be to pass in the ellipsoid as a parameter to the workers. For example, `createVerticesFromQuantizedTerrainMesh` is already structured this way.

> There are other locations where there may be hardcoded WGS84 defaults. One certain case is in Cartesian3 below

Agreed. We could build in pre-computed optimizations like this into the `Ellipsoid` object itself and compute on construction.

> FWIW, this pull request would still be required even with an Ellipsoid.default implementation

Yep!

cesium-concierge:
Thanks again for your contribution @malaretv! 

malaretv:
@ggetz I've merged in `main`, still get test errors seemingly unrelated to my changes. The errors seem to be 1) the `spin` issue that showed up before and 2) errors related to some CesiumTerrain requests timing out

As for the larger fix, indeed it was more than I could handle. I got a bit stuck trying to figure out how to best handle the hardcoded wgs84 defaults in Cartesian3.

ggetz:
Thanks @malaretv, understood, no worries.

I restarted your build to confirm the issues were transient. Looks good now.

The last step here would be to update `CHANGES.md` with a summary of the fix.

ggetz:
Thanks @malaretv!

PR #11100, opened 2023-02-22
Update esbuild

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11063 by updating to the newest version of esbuild. The notable difference here is that [the `rebuild` API](https://esbuild.github.io/api/#rebuild) has changed. Additionally, the initial build no longer happens when making an incremental build. This is fine, as when running the dev server we continue to build and copy all non-JS files up front and can defer the JS bundling until a specific path is requested.

This also:
- modifies the `test` task so that it can be run in succession without the `build` task
- tweaks `server.js` to only build the necessary components from the workspaces (`index.js`) before building CesiumJS as a whole to save a bit of time

To test:
1. Start with a clean state and fresh install
```
git clean -dxf
npm install
```
2. Confirm the dev builds and server run, make sure Sandcastle and the other apps are working as expected
```
npm start
```
3. Make sure individual buils are working:
```
 npm run build
npm run build -- --workspace engine
npm run build -- --workspace widgets
```
4. Confirm the test commands are working
```
 npm run test
 ```
5. Create the release zip
```
 npm run make-zip
 ```
6. Make sure the release zip tasks can run, by extracting the contents into a new directory and runing
```
npm install
npm start
npm test
```

PR #11192, opened 2023-03-29
Fix built viewer after esbuild update

ggetz:
I believe this broke after https://github.com/CesiumGS/cesium/pull/11100. Instead of using `inject` to insert a bit of code to define the `window.CESIUM_BASE_URL` variable, this simplifies things a bit by using the `define` option directly.

To test, run `build-apps` and `npm start -- --production` and navigate to http://localhost:8080/Build/Apps/CesiumViewer/index.html. In main, it will be broken, but it should be working after this fix.

### DIFF
--- a/Apps/CesiumViewer/CesiumViewer.js
+++ b/Apps/CesiumViewer/CesiumViewer.js
@@ -1,6 +1,7 @@
-if (window.CESIUM_BASE_URL === undefined) {
-  window.CESIUM_BASE_URL = "../../Build/CesiumUnminified/";
-}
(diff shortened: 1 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-built-viewer/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Thanks @ggetz! This fixes the built viewer on my machine.

PR #11200, opened 2023-03-30
Cesium ion Headers

ggetz:
For consistency across our products, CesiumJS now includes these headers in  HTTP requests to Cesium ion:

```
X-Cesium-Client: CesiumJS
X-Cesium-Client-Version: 1.103
```

Note that to send the version string, we do set a `CESIUM_VERSION` global based on the version field of `package.json`, which should work in both NodeJS and the browser. For consistency's sake, this uses the top-level `cesium` version string for all packages: `cesium`, `@cesium/engine`, and `@cesium/widgets`.


jjhembd (approved):
Looks good, thanks @ggetz!

PR #11201, opened 2023-03-31
Don't create contexts in the global scope in specs

ggetz:
Creating a context in the global scope has some background side effects, namely requesting some data files, which could cause side effects even in standalone tests.

* Remove any code which created context in the global state
* Create helper to test code in WebGL1/WebGL2, which was the offending reason the context was being created
* Updated testing guide to call out these cases, as well as how to use the helper function


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-context-specs/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
I tried setting up a failing test. The failure is much cleaner in this branch, without any of the strange side-effect failures in unrelated tests.

Thanks @ggetz!

Issue #11197, opened 2023-03-29
Repeated URI parsing slows 3D Tiles performance


jjhembd:
In large tilesets, especially where many tiles are pointing to external tilesets, URI parsing can add a significant overhead to tile loading. Profiling on one of these tilesets shows more than 20% of JavaScript time being used by `URI.parse` and related functions.

There are two things we could do to improve:
- Store a parsed URI in the `Resource` class, so that frequently-used prototype methods like `.clone`, `.getUrlComponent` and `.getDerivedResource` don't have to re-parse it from a string every time.
- Add a `.serverKey` property to the Request constructor. This would reduce calls to `RequestScheduler.getServerKey` , which parses a URI from a string. Note: this property is already assumed in `Cesium3DTile`, where `requestSingleContent` sends a `serverKey` to the constructor&mdash;but the constructor doesn't use it.

Related: https://github.com/CesiumGS/cesium/issues/11168. Ideally, any changes should be written in a way that will be easy to convert to native [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

Issue #11168, opened 2023-03-17
Replace urijs dependency with native URL and URLSearchParams


jjhembd:
The [urijs repository](https://github.com/medialize/URI.js) is no longer active. The README recommends switching to [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

## Challenges
We have [discussed this before](https://github.com/CesiumGS/cesium/pull/9740#issuecomment-903307995), and noticed one complication for the switch:

> URL expects a valid web URL and throws a type error for things like filepaths

The error should usually be avoidable by supplying a base path when constructing a URL from a relative path:

```javascript
const url = new URL(relativePath, location);
```

`location` should be a suitable base URL in most cases, since it is even [defined in Workers](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/location).) But some situations like [parsing KMLs](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/DataSources/KmlDataSource.js#L644) may require more thought

## Performance Implications
Unfortunately, switching to the native API may not provide any performance benefit. If anything, the native API is slightly slower. See this [rough benchmark](https://observablehq.com/@jjhembd/benchmarking-urijs-vs-native-url).

`URI` can already be a significant performance hit for some datasets, so a PR to replace it will need to be thoroughly performance tested.

jjhembd:
One place where we are currently leaning on relative URIs is in the `ImplicitTileset` constructor. The subtree and content URI templates are constructed without any base path:
```javascript
  this.subtreeUriTemplate = new Resource({ url: implicitTiling.subtrees.uri });
  // ...
    const contentResource = new Resource({ url: contentHeader.uri });
```

These URIs are often specified in the tileset.json file as relative URIs, and [according to the spec](https://github.com/CesiumGS/3d-tiles/tree/main/specification#uris),

> When the URI is relative, its base is always relative to the referring tileset JSON file.

Within the `ImplicitTileset` constructor, we have the URL of the "referring tileset JSON file" via the `baseResource` parameter, so we could simply do this:
```javascript
  this.subtreeUriTemplate = baseResource.getDerivedResource({ url: implicitTiling.subtrees.uri });
  // ...
    const contentResource = baseResource.getDerivedResource({ url: contentHeader.uri });
```

[...]

PR #11202, opened 2023-03-31
Reduce URI parsing

jjhembd:
Resolves https://github.com/CesiumGS/cesium/issues/11197.

I put `console.count` calls inside `URI.parse`, and tested loading a detailed photogrammetry tileset. The number of calls to URI.parse in `main` was 1,134,036. With this PR, that number drops to 310,761.

There were several places where URIs were being parsed unnecessarily:
- The Request constructor was not reading `options.serverKey` (even though `Cesium3DTile.prototype.requestContent` was sending a key with every request). This meant that the serverKey had to be re-generated in `RequestScheduler` every time, even for re-tries of a previous request for the same tile.
- `RequestScheduler.getServerKey` re-constructed the URI if the supplied URL was relative (instead of re-using the existing URI object)
- `Resource.prototype.clone` re-parsed the URL for query parameters. However, the URL stored in a `Resource` already has the query parameters stripped off and stored in an object
- `Resource.prototype.getUrlComponent` re-parsed the URL to add the query string back, rather than adding it via a simple template literal

This PR includes 2 public API changes: 
1. Added a `parseUrl` flag to the `Resource` constructor options object. This can be set to false if the supplied URL already has the query string and fragment removed (as when cloning an existing `Resource`).
2. Added a a `serverKey` property to the Request constructor options object


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/reduce-uri-parsing/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

packages/engine/Source/Core/Resource.js
```diff
@@ -227,6 +49,7 @@ function combineQueryParameters(q1, q2, preserveQueryParameters) {
  * @property {Resource.RetryCallback} [retryCallback] The Function to call when a request for this resource fails. If it returns true, the request will be retried.
  * @property {number} [retryAttempts=0] The number of times the retryCallback should be called before giving up.
  * @property {Request} [request] A Request object that will be used. Intended for internal use only.
+ * @property {boolean} [parseUrl=true] If true, parse the url for query parameters; otherwise store the url without change
```
ggetz, line 52: Could we add unit tests to ensure each value of `parseUrl` behaves as expected?

packages/engine/Source/Core/Request.js
```diff
@@ -105,7 +106,7 @@ function Request(options) {
    *
    * @private
    */
-  this.serverKey = undefined;
+  this.serverKey = options.serverKey;
```
ggetz, line 109: Let's also add a unit test to ensure this value is being set as well.

packages/engine/Source/Core/Resource.js
```diff
@@ -568,32 +533,55 @@ Resource.prototype.getUrlComponent = function (query, proxy) {
     return this._url;
   }
 
-  const uri = new Uri(this._url);
+  let url = query
```
ggetz, line 536: This may be bikeshedding a bit, but this ternary may be a bit obtuse.
```
let url = this._url;
if (query) {
    url = `${url}${stringifyQuery(this.queryParameters)}`
}
```
seems more readable (and therefore more maintainable) to me.

packages/engine/Source/Core/Resource.js
```diff
@@ -765,7 +745,6 @@ Resource.prototype.clone = function (result) {
   result.proxy = this.proxy;
   result.retryCallback = this.retryCallback;
   result.retryAttempts = this.retryAttempts;
-  result._retryCount = 0;
```
ggetz, line 768: I think this is still needed, right?

packages/engine/Source/Core/Resource.js
```diff
@@ -753,8 +725,16 @@ Resource.prototype.retryOnError = function (error) {
...
+    return new Resource({
       url: this._url,
+      queryParameters: this.queryParameters,
+      templateValues: this.templateValues,
+      headers: this.headers,
+      proxy: this.proxy,
+      retryCallback: this.retryCallback,
+      retryAttempts: this.retryAttempts,
+      request: this.request.clone(),
+      parseUrl: false,
```
ggetz, line 737: `getDerivedResource` is also a heavily used function. Is it possible to use the same optimization in some cases there?

ggetz (commented):
Thanks @jjhembd! Exciting we can get such an improvement from a relatively small adjustment.

I have a few comments on the code itself, then I'll test.

packages/engine/Source/Core/Resource.js
jjhembd: Yes, you're right. I was confusing it with the case with undefined `result`, where the `Resource` constructor already sets this value. But with a supplied `result`, we do need to reset it.

packages/engine/Source/Core/Request.js
```diff
@@ -166,7 +167,7 @@ Request.prototype.clone = function (result) {
   result.serverKey = this.serverKey;
 
   // These get defaulted because the cloned request hasn't been issued
-  result.state = this.RequestState.UNISSUED;
```
jjhembd, line 169: This was a clear bug uncovered by the new tests. Apparently we were never calling the `Request` constructor with any options.

packages/engine/Source/Core/Resource.js
```diff
@@ -319,13 +142,30 @@ function Resource(options) {
...
+    this._url = options.url;
+  }
+}
 
-  // Remove the fragment as it's not sent with a request
-  uri.fragment("");
+/**
+ * Clones a value if it is defined, otherwise returns the default value
+ *
+ * @param {*} [val] The value to clone.
```
ggetz, line 156: `{Resource}` for all of these?

Also, we try not to use abbreviations (ie use `value` instead of `val`).

packages/engine/Source/Core/Resource.js
```diff
@@ -555,6 +388,138 @@ Resource.prototype.toString = function () {
...
+ * Parses a query string and returns the object equivalent.
+ *
+ * @param {string} queryString The query string
+ * @returns {object}
+ *
+ * @private
+ */
+function parseQueryString(queryString) {
+  if (queryString.length === 0) {
+    return {};
```
ggetz, line 430: The previous iteration of this function did not create new objects. Could we pass in a result parameter?

ggetz:
Thanks @jjhembd! I ran some tests on imagery and terrain, and found very little variance between this branch and `main`– Both with cache enabled and cache disabled. While the benefits here are generally isolated to 3D Tiles, it's still good to know it didn't negatively affect performance in other areas of the code.

packages/engine/Source/Core/Resource.js
jjhembd: This would require a change to `Core/queryToObject`, which currently creates a new result object. I could do that here, but I was thinking of dropping that function entirely when we resolve https://github.com/CesiumGS/cesium/issues/11168.

packages/engine/Source/Core/Resource.js
jjhembd: This helper is mostly used to process options objects, without defined types. I cleaned up the naming, and also deleted the check for `val.clone`, since none of the values passed to the function had such a method.

jjhembd:
Thanks again @ggetz! I cleaned up one of those helper functions you flagged. Sorry for the confusing diff: these were old functions, just rearranged in https://github.com/CesiumGS/cesium/pull/11202/commits/37ca5ea44882206ace971b9b916e81eeafffa0f0 so I could [put the constructor at the top of the file](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#put-the-constructor-function-at-the-top-of-the-file).

I think the other helper could be cleaned up later when we resolve https://github.com/CesiumGS/cesium/issues/11168.

Issue #6991, opened 2018-08-31
Why does BaseLayerPickerViewModel set depthTestAgainstTerrain ?


hpinkos:
https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Widgets/BaseLayerPicker/BaseLayerPickerViewModel.js#L247

This recently led to some confusion in behavior in https://github.com/AnalyticalGraphicsInc/cesium/issues/6990

I don't think BaseLayerPicker should be setting this option.  This seems like a decision the developer should make, not the widget.

We also set `depthTestAgainstTerrain` in `CesiumInpector`.  I think I did this initially because one of the features needed it, but we should see if we can remove it from there now too.

Thoughts?  @mramato @emackey

mramato:
It has always been like this (since the feature was added over 4 years ago #1607 In general, people don't expect data to show through the terrain and not having it turn on automatically leads to a lot of "why is my data wavy/floaty" questions on the forum.

But ultimately it's a moot point because when we enable terrain by default, depthTestAgainstTerrain will also default to true.  At that point, I would be okay with removing it from the BaseLayerPicker.  I wouldn't do anything before that time though.

I'm okay with leaving this open until that change happens just so we don't forget about it.

OmarShehata:
I was just confused by this today. It makes sense why adding terrain changes the depthTestAgainstTerrain default to true, but removing baseLayerPicker sets it back to false **even when there is terrain in the scene**.

[...]

PR #11207, opened 2023-04-03
Set depthTestAgainstTerrain when using the viewer with baseLayerPicker enabled

ggetz:
Due to https://github.com/CesiumGS/cesium/issues/6991, after changes in https://github.com/CesiumGS/cesium/pull/11059, it's required to set this value up front when the `terrain` viewer option is provided and `baseLayerPicker` is not set to `false` in order to not have any breaking behavior. This is because we are now bypassing the baseLayerPIcker entirely when setting the terrain provider on viewer creation.


jjhembd (approved):
Looks good, thanks @ggetz!

PR #11210, opened 2023-04-04
Ensure offCenterFrustum's values are updated

ggetz:
There was a bug where calling the following functions in immediate succession would result in an `DeveloperError: normalized result is not a number`.

```
camera.switchToOrthographicFrustum();
viewer.flyTo(tileset);
```

As illustrated in [this forum post](https://community.cesium.com/t/developererror-normalized-result-is-not-a-number-camera-frustum-change/23160).

This was because the values of `_offCenterFrustum` were not up to date. To remedy this, I made a getter for external use which calls `update`, and updated the external references to the property, ensuring they are called once to avoid running the update function multiple times (though it shouldn't be too much overhead regardless, which is why other getters do the same).

I also re-ordered the specs such that they should now fail if offCeneterFrustum values are not up to date.


ggetz:
@jjhembd Conflicts here have been resolved and this is ready for review!

jjhembd (approved):
Thanks @ggetz, the changes look good, and it fixes the problem on my machine.

PR #11211, opened 2023-04-04
Use Matrix4.multiplyTransformation to update tile transforms

jjhembd:
`Cesium3DTile.prototype.updateVisibility` is called for every tile in the scene, every pass of every frame. One of the most expensive operations within this call is the `Matrix4.multiply` in `updateTransform`.

This PR replaces `Matrix4.multiply` with the more efficient `Matrix4.multiplyTransformation`.

### DIFF
--- a/packages/engine/Source/Scene/Cesium3DTile.js
+++ b/packages/engine/Source/Scene/Cesium3DTile.js
@@ -1796,7 +1796,7 @@ Cesium3DTile.prototype.createBoundingVolume = function (
  */
 Cesium3DTile.prototype.updateTransform = function (parentTransform) {
   parentTransform = defaultValue(parentTransform, Matrix4.IDENTITY);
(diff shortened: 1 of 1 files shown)


ggetz:
Thanks @jjhembd. Between this PR and https://github.com/CesiumGS/cesium/pull/11211, it would be great to add a bullet to `CHANGES.md` with a summary of the performance improvements seen.

ggetz:
Looks good otherwise @jjhembd! Let me know if you plan on updating `CHANGES.md` here, otherwise I'm happy to merge.

jjhembd:
Thanks @ggetz, I just added a note to `CHANGES.md` in https://github.com/CesiumGS/cesium/pull/11202, so I think this one is good to go.

ggetz:
Thanks @jjhembd!

PR #11226, opened 2023-04-12
10957 zoom to entities without globe fix

### LINKED ISSUE #10957: Can not zoom to entities when globe is not displayed
The following (working!) sandcastle creates an entity and zooms to it, using `viewer.zoomTo(viewer.entities)`:

https://sandcastle.cesium.com/index.html#c=bZFRS8MwFIX/yqVPFWrSuuqw64ayPQo+THwKSNbe1WCajCTd2GT/3TTdROYogXtu7/lyeltpZR1sBe7QwBQU7mCOVnQteQ+9mEVV0HOtHBcKDYsS+GYKgNJG6hUWsObSYsLU8WbCFFNVQG60FU5o5aEn4Jwb5yuuRmRtdLvAxiDauEfdZncjko7z/CF7TPpGnpP0Ph2N04cgszRNmerxAxyVE27v0UNwErRAS3hdxyHc+fritwqgVtcoiyE/QGdEASwihC55u5G44I7TMGLpEPpZmFP14UvSyBWLeu/x/LmnAAet2zcdX8QZ9tE/URKxaNI7zwegtG4vcTYIgCfRbrRxPpSMfSKHPhH3C6OrrvpCRyobeMFJ/1rLWmxB1NMrfwoqya31b9adlEtxQBbNSurn/1ml5rVQzesWjeT7fuwzm70MTUJISb287nRayxU3F+Qf

However, when UN-commenting the line that sets
`globe: false,`
for the viewer, it creates an error message:
```
TypeError: Cannot read properties of undefined (reading 'ellipsoid')
    at ModelVisualizer.getBoundingSphere (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:143767:29)
    at DataSourceDisplay.getBoundingSphere (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:146162:32)
    at updateZoomTarget (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:217126:47)
...
```

This is caused by [this line in `ModelVisualizer.js`](https://github.com/CesiumGS/cesium/blob/828d6799916196a6f9c79dd8e5b001d58305e093/packages/engine/Source/DataSources/ModelVisualizer.js#L382) that accesses the globe and its ellipsoid:
```
  const globe = scene.globe;
  const ellipsoid = globe.ellipsoid;
```

Note: A fragment like `const ellipsoid = globe.ellipsoid;` appears in several other `*Visualizer` classes. **Some** of them do check whether the `globe` is valid. Others may not. This has to be investigated. (Another questions is why obtaining the bounding sphere of a model requires the `globe` and its `ellipsoid`, but that might be beyond the scope of this issue)

fullstacc:
Fixes #10957 .
- Added 2 lines which check for the presence of a globe & terrain provider and adds a substitute ellipsoid if globe doesn't exist.
- Based on the description for the issue, I looked at the other visualizers and didn't see this needed in any of the others

Screenshot 2023-04-12 070841

The behavior noted from the issue is fixed

### DIFF
--- a/packages/engine/Source/DataSources/ModelVisualizer.js
+++ b/packages/engine/Source/DataSources/ModelVisualizer.js
@@ -18,7 +18,6 @@ import BoundingSphereState from "./BoundingSphereState.js";
 import Property from "./Property.js";
 import sampleTerrainMostDetailed from "../Core/sampleTerrainMostDetailed.js";
 import Cartographic from "../Core/Cartographic.js";
-import Ellipsoid from "../Core/Ellipsoid.js";
 
 const defaultScale = 1.0;
 const defaultMinimumPixelSize = 0.0;
@@ -446,13 +445,13 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
 
   const scene = this._scene;
   const globe = scene.globe;
-  // Check if the globe is undefined and use a substitute ellipsoid if it is not
-  const ellipsoid = globe ? globe.ellipsoid : Ellipsoid.WGS84;
+
   // cannot access a terrainprovider if there is not one; formally set to undefined
   const terrainProvider = globe ? globe.terrainProvider : undefined;
 
   const hasHeightReference = model.heightReference !== HeightReference.NONE;
-  if (hasHeightReference) {
+  if (globe !== undefined && hasHeightReference) {
+    const ellipsoid = globe.ellipsoid;
     // We cannot query the availability of the terrain provider till its ready, so the
     // bounding sphere state will remain pending till the terrain provider is ready.
     // ready is deprecated. This is here for backwards compatibility
--- a/packages/engine/Source/DataSources/ModelVisualizer.js
+++ b/packages/engine/Source/DataSources/ModelVisualizer.js
@@ -18,6 +18,7 @@ import BoundingSphereState from "./BoundingSphereState.js";
 import Property from "./Property.js";
 import sampleTerrainMostDetailed from "../Core/sampleTerrainMostDetailed.js";
 import Cartographic from "../Core/Cartographic.js";
+import Ellipsoid from "../Core/Ellipsoid.js";
 
 const defaultScale = 1.0;
 const defaultMinimumPixelSize = 0.0;
@@ -63,6 +64,7 @@ function ModelVisualizer(scene, entityCollection) {
   this._entityCollection = entityCollection;
   this._modelHash = {};
   this._entitiesToVisualize = new AssociativeArray();
+
   this._onCollectionChanged(entityCollection, entityCollection.values, [], []);
 }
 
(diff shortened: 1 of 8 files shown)


cesium-concierge:
Thanks for the pull request @fullstacc!
* :x: Missing CLA.
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/10957-zoom-to-entities-without-globe-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

fullstacc:
CLA submitted

packages/engine/Source/DataSources/ModelVisualizer.js
```diff
@@ -444,8 +446,10 @@ ModelVisualizer.prototype.getBoundingSphere = function (entity, result) {
 
   const scene = this._scene;
   const globe = scene.globe;
-  const ellipsoid = globe.ellipsoid;
-  const terrainProvider = globe.terrainProvider;
+  // Check if the globe is undefined and use a substitute ellipsoid if it is not
+  const ellipsoid = globe ? globe.ellipsoid : Ellipsoid.WGS84;
```
ggetz, line 450: Instead of assuming the ellipsoid value, I think it would actually be better to return early and skip some of these unneeded calculations below when `globe` is undefined.

I believe when the globe is undefined, we can simply  skip the `hasHeightReference` block entirely as there is no ellipsoid to offset the position against, and we can just use the position as-is.

ggetz (commented):
Thanks @fullstacc! This will be a helpful fix for sure!

After we determine the best approach, it would be appreciated if you could add:
1. A summary of the fix to `CHANGES.md`
2. A unit test to make sure the issue doesn't reoccur in the future

Let us know if you need any guidance on those.

packages/engine/Source/DataSources/ModelVisualizer.js
fullstacc: thanks for the feedback! I added a condition to `hasHeightReference` that skips that block if the globe is undefined. I also removed the ellipsoid being instantiated as `Ellipsoid.WGS84` if one isn't present, and moved the ellipsoid being defined to within the `hasHeightReference` block as you would need a globe for this anyway

fullstacc:
hi @ggetz ! I'm revisiting this after repeated tries last week. It looks like the CI build fails due to an issue importing the Viewer in the `createViewer` file. I imported and leveraged `createViewer` to test my `modelvisualizer` changes, but I didn't modify the `createViewer` file.

Should I not be using this `createViewer` for my tests?

ggetz:
Hi @fullstacc, `createViewer` depends on widget code, which is a higher level of abstraction than the visualizer code. So really, any test calling `viewer.zoomTo` belong in `ViewerSpec`.

To be more explicit about the architecture, and to keep this from happening in the future, I moved `createViewer` to `packages/widgets/Specs`.

Additionally, `zoomTo` returns a promise which needs to be awaited, otherwise side effects will occur.

Finally, we use the `defined` function to check for both `undefined` and `null`.

I pushed a fix with the above changes.

PR #11247, opened 2023-04-21
Use alpha from rendered scene in AmbientOcclusionModulate

### LINKED ISSUE #10106: Ambient Occlusion doesn't work！
Sandcastle example:Ambient Occlusion

Browser:Chrome

Operating System: macos

image
image
when i check ambient occlusion,full screen change to black and can not look anything!

jjhembd:
Partially addresses #10106. `AmbientOcclusionModulate.glsl` was not setting an output alpha, which broke ambient occlusion on some hardware.

Thanks @shotamatsuda for tracking down the problem!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@
 - Fixed a crash when Cesium3DTileStyle's scaleByDistance, translucencyByDistance or distanceDisplayCondition set to StyleExpression
   which returns `undefined`. [#11228](https://github.com/CesiumGS/cesium/pull/11228)
 - Fixed handling of `out_FragColor` layout declarations when translating shaders to WebGL1. [#11230](https://github.com/CesiumGS/cesium/pull/11230)
+- Fixed a problem with Ambient Occlusion that affected some MacOS hardware. [#10106](https://github.com/CesiumGS/cesium/issues/10106)
 
 #### @cesium/widgets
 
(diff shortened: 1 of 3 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the PR @jjhembd! Unfortunately this does not seem to fix the artifacts on my system (MacOS). Any ideas?

| `main` | `ambient-occlusion-fix` |
| --- | --- |
| image | image |

shotamatsuda:
@ggetz I guess that is out of the scope of the original issue. If you mean "artifacts" as the blocky noise, the problem is that the`randomTexture` coordinate is not scaled. Since the size of `randomTexture` is 256 x 256, and `randomVal` should be random per fragment pixel, the correct coordinates are:

```glsl
- float randomVal = texture(randomTexture, v_textureCoordinates).x;
+ float randomVal = texture(randomTexture, v_textureCoordinates / pixelSize / 255.0).x;
```

1.104:
image

Correct coordinates:
image

`Blur Step Size` is intentionally set to zero for clarity.

However, I think this random value has little contribution to stochatic sampling. i.e. It is just adding some noise.

jjhembd:
@shotamatsuda thanks for the tip on the texture coordinates! Since it's a one-line fix, I went ahead and included it here.

ggetz:
Thanks @jjhembd and @shotamatsuda! This is looking better!

PR #11296, opened 2023-05-18
Terrain availability race condition

ggetz:
`requestTileGeometry` returns an error when `layerToUse` for that particular tile is `undefined`. Doing so marks the entire tile as failed, which means it does not ever get re-requested.

This is fine when we know all the tile availability up-front. In that case, the error is valid.

However, when cut-out terrain is used, it's possible to not know the availability for different layers at the time of request. The existing logic skipped over any layer with unknown availability for that tile, and if there were no other valid layers, then this tile was marked as failed.

This PR adjust things so that we don't continue with the request unless:
1. We find a layer where the tile that is definitely available
2. All of the availability data is ready (in some cases, this could still technically result in the error)

We can't do an early return in `requestTileGeometry` when availability is unknown or the tiles get re-requested too quickly and block other valid requests, ultimately leading to stalling. `checkLayer` caches results to prevent repeat requests, so its not problematic to call in quick succession.

This problem only really becomes apparent when there are a lot of layers with competing requests to get availability data.

@mramato Can you take a look?


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/terrain-availability/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

mramato:
@ggetz This makes sense and seems like a good fix to me. Terrain that used to trigger an error now seems to be error free.

Would it be possible to add a unit test for this?

ggetz:
@mramato Yes, unit test added.

mramato:
Thanks, @ggetz! Confirmed the new unit test fails in main and passes in this branch.

PR #11299, opened 2023-05-22
Update GeometryPipeline.js

TJKoury:
This should be a DeveloperError; certain complex geometries fail to normalize for edge-cases, should not stop rendering.

### DIFF
--- a/packages/engine/Source/Core/GeometryPipeline.js
+++ b/packages/engine/Source/Core/GeometryPipeline.js
@@ -2127,7 +2127,11 @@ function generateBarycentricInterpolateFunction(
     CartesianType.add(value, v2, value);
 
     if (normalize) {
-      CartesianType.normalize(value, value);
+      try {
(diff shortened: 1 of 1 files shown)


cesium-concierge:
Thanks for the pull request @TJKoury!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/patch-4/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @TJKoury! I agree that failure to normalize should generally be developer error as that's consistant with existing behavior. 

However, I'm curious why this is throwing a different type of error. The `normalize` function should already be throwing a developer error for cases where NaN arises. Do you happen to have a code example which reproduces the relevant error?

TJKoury:
Unfortunately all of those are closed source.  We can close this for now, if I can reproduce it with an example I’ll reopen it.

PR #11310, opened 2023-05-25
Enforce memory limit for `Cesium3DTilesetCache`

### LINKED ISSUE #6226: Force tileset to never exceed maximum memory usage
For a `Cesium3DTileset`, the `maximumMemoryUsage` property controls the size of the tileset's cache but does not actually limit how many tiles are loaded. As the doc states:

> If tiles sized more than maximumMemoryUsage are needed to meet the desired screen space error, determined by Cesium3DTileset#maximumScreenSpaceError, for the current view, then the memory usage of the tiles loaded will exceed maximumMemoryUsage. For example, if the maximum is 256 MB, but 300 MB of tiles are needed to meet the screen space error, then 300 MB of tiles may be loaded. When these tiles go out of view, they will be unloaded.

The name is a bit misleading and should really be called `cacheMemoryUsage`.

Really `maximumMemoryUsage` should force the tileset to never consume more than that amount of memory. This could be done by sorting all tiles by screen space error and only loading those with the greatest error until the memory limit is hit. We could also prioritize requests based on SSE as well so we don't end up requesting tiles that won't fit. Unfortunately we don't know a tile size until it is downloaded but we can save that value for later use if the tile is unloaded.

We'll also have to deal with the issue of request thrashing when the maximum memory usage is nearly reached. It may help to reserve some of the memory as a cache.

Going beyond, we should have a global `maximumMemoryUsage` that is shared by all tilesets loaded in the scene.

jjhembd:
Resolves #6226.

This PR deprecates the  `Cesium3DTileset` constructor option `maximumMemoryUsage`, and replaces it with two new options:

1. `cacheBytes`: The size (in bytes) to which the tile cache will be trimmed, if the cache contains tiles not needed for the current view.
2. `maximumCacheOverflowBytes`: The maximum _additional_ memory (in bytes) to allow for cache overflow, if more than `cacheBytes` are needed for the current view.

If the tileset's `totalMemoryUsageInBytes` exceeds `cacheBytes + maximumCacheOverflowBytes`, processing of new tiles will be interrupted, and the target screen-space error used to drive level-of-detail refinement will be adjusted higher. When the camera moves to an area where `totalMemoryUsageInBytes` drops below `cacheBytes`, the target screen-space error will be adjusted back down toward the user-provided value in `maximumScreenSpaceError`.

The adjusted screen-space error is exposed as a new read-only property `memoryAdjustedScreenSpaceError` 

Here is a [hosted Sandcastle](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/tileset-cache/Apps/Sandcastle/index.html#c=zVftbxo3GP9XPDRVMAUfCSFpCGQLCdWihTUqabYPSK25M+DGZ1PbR0Kj/O977PO9EUq6fZqQOM5+nt/z/kIohTZoxegDVaiPBH1AF1SzJMZ37qw+qYXu/UIKQ5igalLbQ08TgRCLyZyq9Y2SKxZR1UUzwjXds1dTouk1WVN1w8L76hURwGeYFOVDw2LKAbw4e26cTsREhE69J6RDKugeCklMFYEnl+E9egaFU81PC0ojP5CIEaHdtbdlRMyihJfC3EgNFM4UTkwXHbbw8eHRyfH+285Jq+304mLeRc3jNj457hwcHRwedU7cOeV05Y1ot45x+607XTITLoB+v+NeFxQUsQAta48THwRoTDP5aCk1syCgljvAmhrr9brTKaLaMOGleDsuiDLwi4g2nikZX9K5olTXLTkqrMKg9t6LM2JenOVm2IuGu5aKUQh0KvUpZcgNyX1bLzD8ZcODex9so3RXGZ2SnINncMu9PxcRBxddMk2mnCKzoC6TIDyQSvbKpQGeczmlWC/kA8TPJUwptlySaAwGJDa4kQyTGOzBc2qGnNqfg/VVBEldkE1qjTx9llRYYz7Qrwl4fzfCBm0ZxjBONdQFVI4Gkp0wG7RlmJCECzpm3+hOgJyqzKohtKGh0a1F38leoSxDxDSWav1RQ5HvBCjRldlJ9CXRgDseD3eyl+hSdg4Fog0k+i00Bf/O9LV0aVYJOKTKeRSh9iVKzXSupAZPRFYu7tG+vPUXtmY+Ku7qZVJbGLPU3SCwbHgu5ZxTsmQahzIOVvtBO3J4gZLS4C9ail/v6bp/fvWNjNeD4fAmvgvF36uj4U3SUcO/RneSnY/Cd38MFo9fbkJoklaILyCbqheKRszo92IcQs1CcRmVUF8MQRDRaTIfA9lAJi6t7iQHh1WoXKAHa2gAXdTZP0C/oP3WwaF/eJqYPLIYmrUlfb+iasYBMmU56BxtYXmeiMZEYCg1Uff+Q/2zVPG02paKxdCmVlRjEkUZkY0Uyj1uqwlyGLqRdlTDFQT4mkFcYWDUZ4kIbUNBvlGJJJ5S9X52U62gvert7UZh2BaVOXSj9jATIOf329G1HWHb0U9Tzo1y28q5Idlzshmqfwcb9ft91EJv3nwPIiXI1Yf2R02ihEd+LgT8lGd6iTivBtBxCTGVKiYipFjIh3rjNKMq14jNmvyiaHUVayc1T48xntTKqjxXg8sEhJ9wZ5JlodGuEOd6F+0YZHjt6/UX+qNmYV8DBZCZLZgKDWzkO/ZIo3o7s/A7Zny+StVL1XVUNEJMoJ+fSrKfoSOCQpH+fLrFQsL/t9b9CQvZv7EMbe2VucVuccJS3MJe9mN2ZtOlolbmuE/uGn/igIA5FXO3ZrmcLY+V7bxVkgp3aYyOBtazGZORhvBRMXGuhOtuzrPQ2NJH4d79zL2lIfVSmdHAU5Vm0VaVU5TzjMo18vGShHSolFS51ING4XL4qu3VetqsOT2zp7+xeCmVQQnMIYwDQ+MlrGYwZ6YJrMoGh1qn7F6aTQqd9z2/MXYRmWqYECar8ikJ7+fKzo5mKLmEjVvNp/UWsp8A1qyjzBEzWOObGuLZha1Kzd3q7MseBl8iDPwN8MIipkG1dRdSzu7mzanNHQ/zwCIDa95hmEXM0EfTJJzNQTfF5guTOR6EMNF0R7As5/ROjxmJGQcBCWvGUkhtXVnRx270P6zM0avKvCq0F2SB6kVshVjU3/L3B4qIaA03s4TzdPE66wVAX2HzJWqHMOhtSRb7Z0XH7QXw+pLLSMmnRG0iZhJLKQEkVuleSmWt64EpohCebbeAZC9S6iAjL/H5YQYzKZ1m3TLSiy0318VnCwholSQgv4st89lXgXux7b4Ct1XhVAI0Qtd9KvilZfhVRce+/6TlXYHZWIr/i5Jpk0KJbTkV7Mq+XELWpgSLRoNXkJtZs4I1zfYh5BIZUduJKvIqC/amvC1WpM9/AA) demonstrating the interaction between memory usage and adjusted screen-space error.

### DIFF
--- a/packages/engine/Source/Scene/Cesium3DTileset.js
+++ b/packages/engine/Source/Scene/Cesium3DTileset.js
@@ -229,7 +229,7 @@ function Cesium3DTileset(options) {
   if (defined(options.maximumMemoryUsage)) {
     deprecationWarning(
       "Cesium3DTileset.maximumMemoryUsage",
-      "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.106.  It will be removed in CesiumJS 1.109. Use Cesium3DTileset.cacheBytes instead."
+      "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.107.  It will be removed in CesiumJS 1.110. Use Cesium3DTileset.cacheBytes instead."
     );
     defaultCacheBytes = options.maximumMemoryUsage * 1024 * 1024;
   }
@@ -1578,14 +1578,14 @@ Object.defineProperties(Cesium3DTileset.prototype, {
     get: function () {
       deprecationWarning(
         "Cesium3DTileset.maximumMemoryUsage",
-        "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.106.  It will be removed in CesiumJS 1.109. Use Cesium3DTileset.cacheBytes instead."
+        "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.107.  It will be removed in CesiumJS 1.110. Use Cesium3DTileset.cacheBytes instead."
       );
       return this._maximumMemoryUsage;
     },
     set: function (value) {
       deprecationWarning(
         "Cesium3DTileset.maximumMemoryUsage",
-        "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.106.  It will be removed in CesiumJS 1.109. Use Cesium3DTileset.cacheBytes instead."
+        "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.107.  It will be removed in CesiumJS 1.110. Use Cesium3DTileset.cacheBytes instead."
       );
       //>>includeStart('debug', pragmas.debug);
       Check.typeOf.number.greaterThanOrEquals("value", value, 0);
@@ -1610,9 +1610,11 @@ Object.defineProperties(Cesium3DTileset.prototype, {
    * If tiles sized more than <code>cacheBytes</code> are needed to meet the
    * desired screen space error, determined by {@link Cesium3DTileset#maximumScreenSpaceError},
    * for the current view, then the memory usage of the tiles loaded will exceed
-   * <code>cacheBytes</code>.  For example, if the maximum is 256 MB, but 300 MB
-   * of tiles are needed to meet the screen space error, then 300 MB of tiles may
-   * be loaded. When these tiles go out of view, they will be unloaded.
+   * <code>cacheBytes</code> by up to <code>maximumCacheOverflowBytes</code>.
+   * For example, if <code>cacheBytes</code> is 500000, but 600000 bytes
+   * of tiles are needed to meet the screen space error, then 600000 bytes of tiles
+   * may be loaded (if <code>maximumCacheOverflowBytes</code> is at least 100000).
+   * When these tiles go out of view, they will be unloaded.
    * </p>
    *
    * @memberof Cesium3DTileset.prototype
@@ -2794,24 +2796,49 @@ function handleTileFailure(error, tileset, tile) {
 }
 
 /**
+ * @private
+ * @param {Cesium3DTileset} tileset
+ */
+function filterProcessingQueue(tileset) {
+  const tiles = tileset._processingQueue;
+
+  let removeCount = 0;
+  for (let i = 0; i < tiles.length; ++i) {
+    const tile = tiles[i];
+    if (
+      tile.isDestroyed() ||
+      tile._contentState !== Cesium3DTileContentState.PROCESSING
+    ) {
+      ++removeCount;
+      continue;
+    }
+    if (removeCount > 0) {
+      tiles[i - removeCount] = tile;
+    }
+  }
+  tiles.length -= removeCount;
+}
+
+/**
  * Process tiles in the PROCESSING state so they will eventually move to the READY state.
  * @private
  * @param {Cesium3DTileset} tileset
  * @param {Cesium3DTile} tile
  */
 function processTiles(tileset, frameState) {
-  const tiles = tileset._processingQueue.filter(
-    (tile) =>
-      !tile.isDestroyed() &&
-      tile._contentState === Cesium3DTileContentState.PROCESSING
-  );
-  tileset._processingQueue = tiles;
+  filterProcessingQueue(tileset);
+  const tiles = tileset._processingQueue;
 
   const { cacheBytes, maximumCacheOverflowBytes, statistics } = tileset;
   const cacheByteLimit = cacheBytes + maximumCacheOverflowBytes;
 
   let memoryExceeded = false;
   for (let i = 0; i < tiles.length; ++i) {
+    if (tileset.totalMemoryUsageInBytes > cacheByteLimit) {
+      memoryExceeded = true;
+      break;
+    }
+
     const tile = tiles[i];
     try {
       tile.process(tileset, frameState);
@@ -2823,11 +2850,6 @@ function processTiles(tileset, frameState) {
     } catch (error) {
       --statistics.numberOfTilesProcessing;
       handleTileFailure(error, tileset, tile);
-    }
-
-    if (tileset.totalMemoryUsageInBytes > cacheByteLimit) {
-      memoryExceeded = true;
-      break;
     }
   }
 
--- a/packages/engine/Source/Scene/Cesium3DTileset.js
+++ b/packages/engine/Source/Scene/Cesium3DTileset.js
@@ -225,18 +225,18 @@ function Cesium3DTileset(options) {
   );
   this._memoryAdjustedScreenSpaceError = this._maximumScreenSpaceError;
 
-  let defaultCachedBytes = 512 * 1024 * 1024;
+  let defaultCacheBytes = 512 * 1024 * 1024;
   if (defined(options.maximumMemoryUsage)) {
     deprecationWarning(
       "Cesium3DTileset.maximumMemoryUsage",
       "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.106.  It will be removed in CesiumJS 1.109. Use Cesium3DTileset.cacheBytes instead."
     );
-    defaultCachedBytes = options.maximumMemoryUsage * 1024 * 1024;
+    defaultCacheBytes = options.maximumMemoryUsage * 1024 * 1024;
   }
-  this._cacheBytes = defaultValue(options.cacheBytes, defaultCachedBytes);
+  this._cacheBytes = defaultValue(options.cacheBytes, defaultCacheBytes);
   this._cacheHeadroomBytes = defaultValue(
     options.cacheHeadroomBytes,
-    defaultCachedBytes
+    defaultCacheBytes
   );
 
   this._styleEngine = new Cesium3DTileStyleEngine();
--- a/packages/engine/Source/Scene/Cesium3DTileset.js
+++ b/packages/engine/Source/Scene/Cesium3DTileset.js
@@ -222,6 +222,7 @@ function Cesium3DTileset(options) {
     16
   );
   this._maximumMemoryUsage = defaultValue(options.maximumMemoryUsage, 512);
+  this._memoryAdjustedScreenSpaceError = this._maximumScreenSpaceError;
 
   this._styleEngine = new Cesium3DTileStyleEngine();
   this._styleApplied = false;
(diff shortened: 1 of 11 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/tileset-cache/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Instead of `cacheHeadroomBytes`,  perhaps `maximumCacheOverflowBytes`?

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -221,7 +223,21 @@ function Cesium3DTileset(options) {
...
+  let defaultCacheBytes = 512 * 1024 * 1024;
+  if (defined(options.maximumMemoryUsage)) {
+    deprecationWarning(
+      "Cesium3DTileset.maximumMemoryUsage",
+      "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.106.  It will be removed in CesiumJS 1.109. Use Cesium3DTileset.cacheBytes instead."
+    );
+    defaultCacheBytes = options.maximumMemoryUsage * 1024 * 1024;
+  }
+  this._cacheBytes = defaultValue(options.cacheBytes, defaultCacheBytes);
+  this._cacheHeadroomBytes = Math.max(
```
ggetz, line 237: I would use `Check.typeOf.number.greaterThan` instead of silently overriding the provided value.

lilleyse:
@jjhembd if it's not too much trouble would it be possible to move the cleanup changes into a separate PR? I have some minor comments there but don't want to distract from the main review.

jjhembd:
> would it be possible to move the cleanup changes into a separate PR?

@lilleyse I moved all the cleanup commits into the `texture-cleanup` branch. The changes here should now be only the memory limit changes.

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -68,7 +68,9 @@ import Cesium3DTilesetSkipTraversal from "./Cesium3DTilesetSkipTraversal.js";
  * @property {Axis} [modelForwardAxis=Axis.X] Which axis is considered forward when loading models for tile contents.
  * @property {ShadowMode} [shadows=ShadowMode.ENABLED] Determines whether the tileset casts or receives shadows from light sources.
  * @property {number} [maximumScreenSpaceError=16] The maximum screen space error used to drive level of detail refinement.
- * @property {number} [maximumMemoryUsage=512] The maximum amount of memory in MB that can be used by the tileset.
+ * @property {number} [maximumMemoryUsage=512] The maximum amount of memory in MB that can be used by the tileset. Deprecated.
+ * @property {number} [cacheBytes=536870912] The size (in bytes) to which the tile cache will be trimmed, if the cache contains tiles not needed for the current view.
+ * @property {number} [maximumCacheOverflowBytes=536870912] The maximum additional memory (in bytes) to allow for cache headroom, if more than {@link Cesium3DTileset#cacheBytes} are needed for the current view. Must be at least 10 * 1024 * 1024.
```
lilleyse, line 73: Make sure to include instructions in `CHANGES.md` for replicating the previous behavior.

I think that would be setting `maximumCacheOverflowBytes` to `Number.MAX_VALUE`?

lilleyse, line 73: > Must be at least 10 * 1024 * 1024

This seems a bit arbitrary. Why can't it be 0?

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -1545,17 +1571,123 @@ Object.defineProperties(Cesium3DTileset.prototype, {
...
+   * <p>
+   * If decreasing this value results in unloading tiles, the tiles are unloaded the next frame.
+   * </p>
+   * <p>
+   * If tiles sized more than <code>cacheBytes</code> are needed to meet the
+   * desired screen space error, determined by {@link Cesium3DTileset#maximumScreenSpaceError},
+   * for the current view, then the memory usage of the tiles loaded will exceed
+   * <code>cacheBytes</code>.  For example, if the maximum is 256 MB, but 300 MB
+   * of tiles are needed to meet the screen space error, then 300 MB of tiles may
+   * be loaded. When these tiles go out of view, they will be unloaded.
```
lilleyse, line 1615: This description is a bit incomplete without also mentioning `maximumCacheOverflowBytes`

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -1545,17 +1571,123 @@ Object.defineProperties(Cesium3DTileset.prototype, {
    *
    * @exception {DeveloperError} <code>maximumMemoryUsage</code> must be greater than or equal to zero.
    * @see Cesium3DTileset#totalMemoryUsageInBytes
+   *
+   * @deprecated
    */
   maximumMemoryUsage: {
```
lilleyse, line 1577: Several tests should be updated to use `cacheBytes`.

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -2661,43 +2793,26 @@ function handleTileFailure(error, tileset, tile) {
...
 function processTiles(tileset, frameState) {
-  filterProcessingQueue(tileset);
-  const tiles = tileset._processingQueue;
-  const statistics = tileset._statistics;
-  let tile;
+  const tiles = tileset._processingQueue.filter(
+    (tile) =>
+      !tile.isDestroyed() &&
+      tile._contentState === Cesium3DTileContentState.PROCESSING
+  );
```
lilleyse, line 2807: While more concise, this will allocate an array every frame (which is what `filterProcessingQueue` avoids)

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -2709,7 +2824,34 @@ function processTiles(tileset, frameState) {
...
+  }
+}
+
+function increaseScreenSpaceError(tileset) {
+  tileset._memoryAdjustedScreenSpaceError *= 1.02;
+  const tiles = tileset._processingQueue;
+  for (let i = 0; i < tiles.length; ++i) {
+    tiles[i].updatePriority();
   }
+  tiles.sort(sortTilesByPriority);
```
lilleyse, line 2847: I wonder if we should always be updating priority and sorting, not just when the SSE changes. Because it doesn't seem strictly necessary to just do it here.

Also is there a reason it's updated here and not in `decreaseScreenSpaceError`?

packages/engine/Source/Scene/Cesium3DTileset.js
lilleyse, line 1: Reminder to add some unit tests

packages/engine/Source/Scene/Cesium3DTileset.js
jjhembd: We currently allow `maximumMemoryUsage == 0`, to force the cache to be trimmed to the minimum size possible. This is used several times in the specs, so I also allowed `cacheBytes == 0` for consistency.

If both `cacheBytes == 0` and `maximumCacheOverflowBytes == 0`, then no tiles will be processed and `memoryAdjustedScreenSpaceError` will grow to infinity. This feels buggy, as Matt noticed already, even with a small non-zero value.

I chose 10 MB as a number that should guarantee we render at least one tile.

packages/engine/Source/Scene/Cesium3DTileset.js
jjhembd: I restored `filterProcessingQueue` for now.

We are also potentially re-allocating by using `tileset._processingQueue.push()` in `requestContent`. If this is really an issue, perhaps a separate PR could convert `._processingQueue` to a `ManagedArray`, and move the custom filter function to a method of `ManagedArray`.

packages/engine/Source/Scene/Cesium3DTileset.js
jjhembd: As discussed with @ggetz, I changed this to allow `maximumCacheOverflowBytes == 0`, and simply added a `oneTimeWarning` inside `increaseScreenSpaceError`, to alert the developer that the values of `cacheBytes` and `maximumCacheOverflowBytes` might be too small.

jjhembd:
Thanks @lilleyse! I think I've addressed all your feedback.

@ggetz can you give this a final look?

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -221,7 +223,30 @@ function Cesium3DTileset(options) {
...
+
+  let defaultCacheBytes = 512 * 1024 * 1024;
+  if (defined(options.maximumMemoryUsage)) {
+    deprecationWarning(
+      "Cesium3DTileset.maximumMemoryUsage",
+      "Cesium3DTileset.maximumMemoryUsage was deprecated in CesiumJS 1.107.  It will be removed in CesiumJS 1.110. Use Cesium3DTileset.cacheBytes instead."
+    );
+    defaultCacheBytes = options.maximumMemoryUsage * 1024 * 1024;
+  }
+  this._cacheBytes = defaultValue(options.cacheBytes, defaultCacheBytes);
```
ggetz, line 236: We should check to ensure this value is greater than or equal to 0, right? As we do in the `cacheBytes` setter.

packages/engine/Specs/Scene/Cesium3DTilesetSpec.js
```diff
@@ -3471,11 +3471,11 @@ describe(
       });
     });
 
-    it("Unload some cached tiles not required to meet SSE using maximumMemoryUsage", function () {
+    it("Unload some cached tiles not required to meet SSE using cacheBytes", function () {
```
ggetz, line 3474: Since it was deprecated, can we leave the old tests specifically for `maximumMemoryUsage`, and add new ones to cover `cacheBytes`?

ggetz (commented):
Thanks @jjhembd! This is working well! Just a few comments.

packages/engine/Source/Scene/Cesium3DTileset.js
jjhembd: Yes, I added this check just below. It was in the last commit which was affected by the GitHub outage, but it should have come through now.

jjhembd:
Thanks @ggetz! I think I addressed all your feedback. Let me know if there's anything else.

ggetz:
All looks good to me! Thanks @jjhembd!

Issue #11286, opened 2023-05-16
V1.105.1 After set viewer.trackentity，camera cannot zoom 


easydu2002:
Sandcastle example 1: https://sandcastle.cesium.com/?src=3D%20Models.html
Sandcastle example 2: https://sandcastle.cesium.com/#c=bVFda8IwFP0rlz5V6NJWO3W1yobubbCB4FNhxPbqwvIhSerHhv99aaowNhMI955zz8nNTaWksbBneEANU5B4gDka1giy8lhYBpXP50payiTqMuhNStntyqtRWmZPTt3ZEJ8zNITWdfhdSoCdMg5RMr+az6m2LqJyQDZaiQVuNaIJ79L+gCSjLBumDxFkGUnuk8EoGUYwTpJe1FoJVSPPwdsCNJrlUAaExEsqdhwX1NLYl5i4u+qJ6Uv07kKy5esyiDqxYJKJRryxI/Il+8Ic0v74ytFjyy0ryh3eT9zyzNmd53YAl7daTatPrJ+vI+hmMQmioDD2xHHW2QE8MrFT2rqOeejatejapW4G8bpxBpZUxrS2bWkR/5YWNdsDq6c3PgIqTo1xzKbh/gVlMCtiV/9PyhWtmdy+7lFzemrLPtLZSwcSQorYpbeVVim+pvqP8w8

Operating System:  Window11 Chrome/113.0.0.0

PR #11268, opened 2023-05-03
Adjust camera controls for when the globe is off, and when panning in 3D

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/7171

Picking for all camera controls had some strange conditions:
 - `scene.pickPosition` was never used if the globe was undefined (so the ellipsoid was picked instead), which is not a strict requirement. Now, we always go with the `scene.pickPosition` result if defined and the globe isn't, then fallback to the ellipsoid as needed.
 - The closer of either the globe or the scene position was picked under all scenarios. This is now only the case when underground or when globe translucency is enabled.
 - Since we're not always picking the globe, I renamed the `pickGlobe` function to `pickPosition`

Panning formerly did not pick at all and instead used only the ellipsoid. 3D panning now picks a position based off the initial mouse position and moves in worldspace based on the mouse translation. This allows for much better control when panning through 3D tiles such as a cityscape, or other more vertical surfaces.

I did my best to test a variety of scenarios including combinations of the following, but let me know if there are any blind spots:
- panning, zooming, & orbiting
- "throwing" the globe or scene when panning with inertia
- default globe, globe with terrain
- far away, close to the ellipsoid or terrain, 3/4 views, straight down, steep horizon views, lookup up at the sky
- translucent globe, underground
- 3D tiles on the default globe, on a globe with terrain, or with `globe: false`
- 3D tiles as terrain (S2 globe example)
- models and entities on the default globe, on a globe with terrain, or with `globe: false`
- perspective and orthographic frustums
- different aspect ratios and resolutions

PR #11319, opened 2023-05-30
Fix for zoom on tracked entity

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11286

In https://github.com/CesiumGS/cesium/pull/11268, we tweaked the camera controls when the globe is off.

However, the internals of screenSpaceCameraController override the internal value of `._globe` when a non-identity camera transform is used, such as when tracking an entity. In most cases, this didn't matter. However in non-3D zoom this value did matter, and was now unexpectedly picking a position under the mouse to zoom to, which led to unexpected behavior-- no zooming. I added some additional unit tests to catch this edge case in case it comes up again in the future.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 
 ##### Fixes :wrench:
 
+- Fixed tracked entity camera controls. [#11286](https://github.com/CesiumGS/cesium/issues/11286)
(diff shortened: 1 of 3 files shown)


jjhembd (approved):
Looks good! 
The code change makes sense. (I hadn't noticed that the `.update` method sets the globe to undefined if a camera transform is set). 
If I revert the change, the new spec catches the bug.
Thanks @ggetz!

Issue #11324, opened 2023-06-01
Update tween.js dependency


jjhembd:
The latest [major version of tween.js](https://github.com/tweenjs/tween.js/releases/tag/v20.0.0) results in build warnings:

```bash
  Build/Cesium/Cesium.js  3.7mb ⚠️
⚡ Done in 1556ms
▲ [WARNING] Import "Easing" will always be undefined because the file "node_modules/@tweenjs/tween.js/dist/tween.cjs.js" has no exports [import-is-undefined]
    packages/engine/Source/Core/EasingFunction.js:1:9:
      1 │ import { Easing } from "@tweenjs/tween.js";
        ╵          ~~~~~~
▲ [WARNING] Import "Tween" will always be undefined because the file "node_modules/@tweenjs/tween.js/dist/tween.cjs.js" has no exports [import-is-undefined]
    packages/engine/Source/Scene/TweenCollection.js:8:9:
      8 │ import { Tween as TweenJS } from "@tweenjs/tween.js";
        ╵          ~~~~~
```

These warnings will need to be resolved before updating the dependency.

ggetz:
Based on [the docs](https://www.npmjs.com/package/@tweenjs/tween.js), it looks like we'd want to make sure the import points to `tween.es.js`.

PR #11325, opened 2023-06-01
dependency updates

jjhembd:
Updates a few outdated npm dependencies.

tween.js is not updated yet, pending resolution of #11324.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/npm-updates/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

package.json
```diff
@@ -55,19 +55,19 @@
     "@cesium/widgets": "2.3.0"
   },
   "devDependencies": {
-    "@aws-sdk/client-s3": "^3.276.0",
-    "@aws-sdk/lib-storage": "^3.276.0",
-    "@rollup/plugin-commonjs": "^24.0.1",
-    "@rollup/plugin-node-resolve": "^15.0.1",
```
ggetz, line 61: Since the old version is still compatible, we usually don't bump the patch versions. The exception would be if the old version is no longer working for whatever reason.

ggetz:
Thanks @jjhembd! I've tested the updates, and everything is looking good.

Just that one comment, and this should be good to go.

package.json
jjhembd: I just reverted the changes that only affected patch versions. All changes should now be minor or major version updates.

ggetz:
Looks great, thanks @jjhembd!

Issue #11335, opened 2023-06-06
Offline and Cesium Ion Logo


Kai-W:
I have an Cesium application that should not connect to Ion services. I followed the steps in
https://github.com/CesiumGS/cesium/tree/main/Documentation/OfflineGuide
and still getting a Cesium Ion Logo. This behavior is also present in the offline sandcastle example.
Either something is missing in the docs/example to make it fully offline without Ion or the Cesium Ion logo is not displayed correctly
Additionally the link to the docs in the offline Sandcastle links to old migrated 

Sandcastle example:
https://sandcastle.cesium.com/index.html?src=Offline.html

Browser:
Chrome 113

Operating System:
Windows 11

ggetz:
Thanks @Kai-W. You can [remove the Cesium ion logo](https://community.cesium.com/t/cesium-ion-logo-removal/8979) if you are not using any Cesium ion services. Though we ask you still credit CesiumJS in your application, it is not strictly required.

I've updated the link in https://github.com/CesiumGS/cesium/pull/11342.

Kai-W:
I don't like the fact that i have to remove credits like this. Why isn't the default that cesiumjs is the default logo/credit and if an ion service ist accessed the Ion logo gets swapped in. 
Besides in the referenced community post the complete credit container gets removed, so attributions like OSM also don't show. Just removing the logo should be the way to go:
`.cesium-credit-logoContainer {
    display: none !important;
}`

Kai-W:
`Cesium.CreditDisplay.cesiumCredit = null` looks like a cleaner solution to remove the logo.

PR #11342, opened 2023-06-07
Update link to offline guide

ggetz:
Updates link to offline guide as noted in https://github.com/CesiumGS/cesium/issues/11335.


jjhembd (approved):
Thanks @ggetz!

Issue #11068, opened 2023-02-03
Automated Testing


ggetz:
As part of our [release process](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/ReleaseGuide#release-testing-and-packaging), and often as testing while reviewing wide-reaching PRs, we manually run through every Sandcastle example and spot test any differences.

An automated end-to-end testing solution with screenshot comparisons like [Playwright](https://playwright.dev/) would help speed up these processes considerably, and will also likely be less error-prone about spotting any differences.

PR #11348, opened 2023-06-12
Add Automated Screenshot (E2E) Tests with Playwright

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11068

This PR adds automated "end-to-end" testing with [Playwright](https://playwright.dev/). Playwright allows us to set up automated screenshot testing runnable in NodeJS, and is used in some of our other internal projects.

### Running Tests
Tests are runnable with the following commands:
```
npm run test-e2e
npm run test-e2e-report
```

For running the tests against release code:
```
npm run test-e2e-release
npm run test-e2e-report
```

See `Documentation/TestingGuide/README.md` for a full descriptions of the added commands.

 ### Tests
The majority of tests added in this PR are tests for each sandcastle, automated in `sandcastle.spec.js`. It pulls all examples and runs them in standalone mode (no surrounding Sandcastle UI).

I did tweak a few sandcastle so that the initial view resolves faster. This benefits the automated testing, but I think it also helps users get to the important parts of the example faster.

The rest are a few illustrative examples, testing certain features by evaluating CesiumJS code more directly. There are a few "dirt simple" examples like the Viewer, but also a few that were a bit more difficult to test through a raw sandcastle example snapshot, like picking and model animations.

### Screenshots
Screenshots are saved to disk and committed to the repo. Since there are many, I set the viewport size fairly small.

If screenshots ever need to be updated, such as when a new sandcastle is added or behavior intentionally changes, `test-e2e-update` can be used to run the relevant E2E tests and generate any new screenshots.

### Config
The setup required a fairly basic configuration, specified in `Specs/e2e/playwright.config.json`.

The biggest notable adjustment to get playwright working on CesiumJS examples is time. Time is often taken into account for things like the timeline rendering, animations, positions of the skybox relative to the rotation of the earth-- All critical to ensuring a consistant render for screenshot testing. To account for this, we use [Sinon](https://sinonjs.org/) to mock system time, ensuring consistency for all end to end tests. It allows us to set a specific datetime to use for all tests without modifying any test code, and advance system time as needed for specific tests. This is configured in Specs/e2e/test.js. This is the [current recommended workaround until playwright directly supports time emulation](https://github.com/microsoft/playwright/issues/6347#issuecomment-965887758).

For sandcastle tests especially, we use super lenient timeouts here, and rely on network idle event (which is [discouraged in the playwright docs](https://playwright.dev/docs/api/class-frame#frame-wait-for-load-state)) since a comparable screenshot needs to ensure all the imagery is loaded in. If we phase away from Sandcastle examples, and move to more test-oriented examples and more offline data, it'll be possible to better optimize these so that we don't rely on these longer timeouts. Hence there's a difference in the default config vs the sandcastle tests in terms of timeouts.

The visual threshold though are still fairly conservative, since once the data is actually loaded in, there should be a fairly low tolerance for differences in rendering given the set parameters.

Given the long timeouts and network loading, these do take a few minutes to execute. I would recommend only running these locally or in the main branch for CI to avoid the overhead in regular CI runs.

TODO:
- [x] Cleanup commit history
- [x] Update release guide to include e2e step before release
- [x] ~Running in CI~
- [ ] Coverage task?


Documentation/Contributors/TestingGuide/README.md
```diff
@@ -239,6 +240,58 @@ It is possible to have 100% code coverage with two tests: one test where `a` and
...
 
+### Run End to End Tests
+
+End to end (E2E) testing is a type of testing that tests the entire stack from the user's perspective. This is different to unit testing, which validates a small isolated piece of functionality at the class or function level. In CesiumJS, this testing consists mainly of screenshot comparisons.
+
+[Playwright](https://playwright.dev/) is used to conduct end to end testing. The Playwright tests run in Node and drive instances of the browsers. It is recommended that your read through the [Playwright documentation](https://playwright.dev/docs/intro) to get up to speed before writing or reviewing tests.
+
+Since CesiumJS often takes the current time into consideration for things like lighting, animation, and the position of the skybox corresponding to the earth's rotation, we use [Sinon](https://sinonjs.org/) to mock system time, ensuring consistency for all end to end tests.
+
+#### E2E Test Tasks
```
jjhembd, line 251: The acronym (E2E) is useful for the script names, and perhaps the body of the text. But maybe spelling it out is still better for section headers.

jjhembd (commented):
Thanks @ggetz, it's exciting to think of automating part of our release process!

I was able to get the tests running on my machine. A few observations:

## 1. Setup
My first run got this error:
image

I had to manually install the dependencies using `apt install`. The `playwright` command complained:
```bash
/usr/bin/env: ‘node’: No such file or directory
```

If these setup steps cannot be automated as part of `npm install`, then a note in the TestingGuide would be helpful.

## 2. Screenshot filenames
The screenshots have names like this: `loads-animated-model-1-chromium-darwin.png`. When I run the tests on my machine, they are re-generated with a `-linux.png` postfix, and the comparison fails.

I can get the tests to work by simply running twice: once to generate the baseline screenshots, then once (after switching to the new commit/branch/tag) to compare. But it would be helpful in some cases to compare across different operating systems, which we could do if we drop the OS postfix.

## 3. Screenshot size
The screenshots are currently 1280x720 pixels. All together they add 200MB to the repo size. 

Would 640x360 pixels be big enough?

## 4. Sandcastle starting views
I noticed you already changed some of the Sandcastles to have a more meaningful starting view. A few more could use similar tweaks. These observations are based on the screenshots:

- Batch Table Hierarchy: start with one of the styles selected
- 3D Tiles Point Cloud Styling: start with a style selected
- ArcticDEM: start with the camera pointing at Denali
- Cesium World Terrain: zoom to show terrain
- Clamp to 3D Model: clamped point was not captured
- Custom Shaders Models: start camera at model position (flight didn't complete for screenshot)
- Custom Shaders Property Textures: The Sandcastle change makes sense, but somehow the screenshot is still not showing a custom shader result
- Fog Post Process: screenshot shows "Loading..."
- GPX: zoom to track location
- Globe Translucency: flight didn't complete. Maybe just `setView` directly to the target position?
- I3S examples could start zoomed in
- Imagery Adjustment: start from adjusted values? (the current starting values show no change in the imagery)
- MSAA: start with it on
- PAMAP terrain: start zoomed in
- Particle System...: Neither the Fireworks nor the Tails were captured in the screenshot. Perhaps we need a longer delay before capturing?
- Polylines on 3D Tiles: the line was not captured
- Sample Height from 3D Tiles: the sampling was not captured
- Terrain: adjust system time? It looks dark
- Washington DC 2017: zoom in

There are also a few others which are harder to evaluate from a screenshot, though I'm not sure they can be addressed in this PR:
- Articulations
- Custom Geocoder
- Drawing on Terrain
- Heading Pitch Roll
- KML Tours
- Star Burst

ggetz:
1. The additional install step is set up to run on postinstall. 

> cesium@1.106.1 postinstall
> playwright install

It works for me from a clean install. `apt-get` isn't available for me on a non-linux system. @jjhembd Does either `playwright install` or `playwright install-deps` work for you at all, or only `apt-get`?

ggetz:
2. 

> Screenshots differ between browsers and platforms due to different rendering, fonts and more, so you will need different snapshots for them. If you use multiple projects in your [configuration file](https://playwright.dev/docs/test-configuration), project name will be used instead of chromium.

Is is possible to drop the suffix, but I'm concerned we'll still see minor differences:

I think the main difference we will end up seeing across systems is differences in font. We could hide all UI elements, but that means we are getting further away from what things actually look like.

3.

I set the default viewport smaller, but this was actually being [overridden by the device configuration](https://github.com/microsoft/playwright/issues/13815). Updated so we're using 960x540. Any smaller, and there starts to be issues with the UI cluttering the rendered scene.

This does bring up the question of storage in general. As more tests or sandcastles get added and the screenshots get updated, plus maintaining a version for each system, this will quickly bloat the size of the repo. Normally, the solution here would be to use git-lfs, but doing so in a popular public repo can affect the bandwidth quota for the entire Cesium organization. 

Given that, the fact we plan to use this mostly for release testing, and the answer to 2, I think the simplest thing to do for the time being is to skip committing the screenshot to the repo. This does result in not being able to run the tests as easily in CI. However, the tests take ~7 minutes to run on my machine, are network dependent, and somewhat system dependent since we're not running in docker. This also requires developers to run the tests once locally against main as a baseline to generate local versions of the screenshots.

If we do want to enable CI, at that point I think it may make sense to have CI run the test in the proper environment and upload and compare results to files stored externally, such as in a S3 bucket. Plus probably only run it in the `main` branch due to the length of time it takes to run.

To sum up, given these test are still useful to accelerate local testing such as during releases or when testing wide reaching PRs like moving WebGL2 by default, I think we keep things simple and skip committing the screenshot and CI. I'll update the guides to describe the local workflow. If we find the test useful over time and want to add them to the CI workflow, we'll want write a custom script which generates any missing screenshot in the specific environment and handles storage separately.

ggetz:
Thanks @jjhembd! Besides the particle system examples, I added improvements to the Sandcastle examples themselves, or fixed bugs in the standalone versions.

For particles systems (and other animations) we now tick time forward a few times to ensure they get a few frames of execution before capturing the screenshot.

Specs/e2e/sandcastle.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    test.setTimeout(100000);
+
+    await page.goto(example);
+
+    await page.waitForLoadState("networkidle");
+
+    await page.evaluate(() => window.__clock.tick(1000));
+    await page.evaluate(() => window.__clock.tick(1000));
+    await page.evaluate(() => window.__clock.tick(1000));
+    await page.evaluate(() => window.__clock.tick(1000));
```
jjhembd, line 17: These `page.evaluate` calls gave me a confusing error in one test run:

```bash
  1) [chromium] › sandcastle.spec.js:7:3 › Apps/Sandcastle/gallery/Particle System Weather.html renders

    Error: page.evaluate: TypeError: Cannot read properties of undefined (reading 'tick')

        at eval (eval at evaluate (:197:30), <anonymous>:1:22)
        at UtilityScript.evaluate (<anonymous>:199:17)
        at UtilityScript.<anonymous> (<anonymous>:1:44)
        at eval (/home/jhembd/CesiumGS/cesium/eval at evaluate (:197:30), <anonymous>:1:22)
        at UtilityScript.evaluate (/home/jhembd/CesiumGS/cesium/<anonymous>:199:17)
        at UtilityScript.<anonymous> (/home/jhembd/CesiumGS/cesium/<anonymous>:1:44)
        at /home/jhembd/CesiumGS/cesium/Specs/e2e/sandcastle.spec.js:14:16
```

I could avoid it by using optional chaining, to handle the case `window.__clock === undefined`. 

```javascript
await page.evaluate(() => window.__clock?.tick(1000));
```

This gave me a more meaningful error message.

ggetz: This means that the mock clock is not getting installed in time, which is a bigger issue... I'll take a look.

jjhembd:
Thanks @ggetz! With those updates, the tests are a lot more stable and meaningful. I'm still seeing a couple issues with a few tests.

## Test stability

My new run of `npm test-e2e-update` had 2 failures:

```bash
  2 failed
    [chromium] › sandcastle.spec.js:7:3 › Apps/Sandcastle/gallery/Polylines on 3D Tiles.html renders
    [chromium] › sandcastle.spec.js:7:3 › Apps/Sandcastle/gallery/Video.html renders ───────────────
  184 passed (33.6m)
```

Both failures were due to timeouts on this line:

```bash
    > 19 |     await expect(page).toHaveScreenshot({
         |                        ^
      20 |       timeout: 20000,
      21 |     });
```

These ran correctly on a second try. I re-ran only those tests via `npm test-e2e-update -- -g "Polylines on 3D Tiles"` (etc).

I then ran `npm test-e2e`. Only one test failed:

```bash
 1) [chromium] › sandcastle.spec.js:7:3 › Apps/Sandcastle/gallery/3D Tiles Next CDB Yemen.html renders

    Test timeout of 100000ms exceeded.

    Error: page.waitForLoadState: Navigation failed because page was closed!

      10 |     await page.goto(example);
      11 |
    > 12 |     await page.waitForLoadState("networkidle");
         |                ^
      13 |
      14 |     await page.evaluate(() => window.__clock.tick(1000));
      15 |     await page.evaluate(() => window.__clock.tick(1000));

        at /home/jhembd/CesiumGS/cesium/Specs/e2e/sandcastle.spec.js:12:16
```

I'm still seeing 4 workers running (there were 6 before the updates). Could this be introducing some timeout problems?

```bash
$ npm run test-e2e

> cesium@1.106.1 test-e2e
> playwright test -c Specs/e2e/playwright.config.js --project=chromium


Running 186 tests using 4 workers
```

## Test validity

The screenshots generally look more meaningful, thanks to your Sandcastle updates! A few of them show the important parts of the display, but the background globe is not loading very well. For example:

Multi part CZML:
image

Globe Translucency:
image

CZML Point:
image

Callback Property:
image

ggetz:
@jjhembd I reduced the number of workers to 1. Any luck with better consistency?

jjhembd:
Thanks @ggetz, the tests are now running sequentially, one at a time. I'm getting an error with one Sandcastle when I run `npm run test-e2e-update` from a clean install:

```bash
  1) [chromium] › picking.spec.js:3:1 › Shows cartographic position on mouse over ──────────────────

    Error: page.evaluate: TypeError: Cannot read properties of undefined (reading 'tickAsync')

        at eval (eval at evaluate (:197:30), <anonymous>:1:22)
        at UtilityScript.evaluate (<anonymous>:199:17)
        at UtilityScript.<anonymous> (<anonymous>:1:44)
        at eval (/home/jhembd/CesiumGS/cesium/eval at evaluate (:197:30), <anonymous>:1:22)
        at UtilityScript.evaluate (/home/jhembd/CesiumGS/cesium/<anonymous>:199:17)
        at UtilityScript.<anonymous> (/home/jhembd/CesiumGS/cesium/<anonymous>:1:44)
        at CesiumPage.tick (/home/jhembd/CesiumGS/cesium/Specs/e2e/CesiumPage.js:25:21)
        at /home/jhembd/CesiumGS/cesium/Specs/e2e/picking.spec.js:51:20

  Slow test file: [chromium] › sandcastle.spec.js (32.5m)
  Consider splitting slow test files to speed up parallel execution
  1 failed
    [chromium] › picking.spec.js:3:1 › Shows cartographic position on mouse over ───────────────────
  185 passed (32.8m)
```

I see that the problematic `.tickAsync` reference is from `CesiumPage.js`, but I didn't trace back how this causes the failure.

Also, there are a few snapshots that still look like the background globe isn't fully loaded. For example, here's the snapshot for the "Time Dynamic Wheels" Sandcastle:

image

I'm not sure how much of an issue this is, since the key parts of the demo (which are loaded from local files) rendered correctly.

ggetz:
// Also, there are a few snapshots that still look like the background globe isn't fully loaded. For example, here's the snapshot for the "Time Dynamic Wheels" Sandcastle:

@jjhembd Are you running `npm run test-e2e-update` with a clean screenshot directory, or did you have the previous screenshots captured before you re-ran the command?

ggetz:
Ah, I think I've found the problem with `tick` and `tickAsync`. This was the mock time-based functions that _should_ have been installed for each test. Turns out [`beforeEach` shouldn't be used in a helper file, as the behavior is undefined](https://github.com/microsoft/playwright/issues/18944#issuecomment-1322691705). I've updated the tests so that these are installed via default test fixtures instead.

jjhembd (approved):
Thanks again @ggetz! The tests now run through without errors. I'm looking forward to using this on the next release!

PR #11358, opened 2023-06-16
Fixed Insufficient buffer size thrown by rendering 3dtiles.

IKangXu:
Before repair
image

After repair
image

This is the test data. (Data needs to be published independently.)
[test_data.zip](https://github.com/CesiumGS/cesium/files/11767969/test_data.zip)

This is the test code.
```js

try {
  const tileset = await Cesium.Cesium3DTileset.fromUrl('http://127.0.0.1:9092/63b0f42c-c5ef-4874-bbdc-21e9d6e3ee56/tileset.json');
  viewer.zoomTo(tileset);
} catch (error) {
  console.log(`Error loading tileset: ${error}`);
} 
```

### DIFF
--- a/packages/engine/Source/Renderer/Context.js
+++ b/packages/engine/Source/Renderer/Context.js
@@ -1340,7 +1340,14 @@ function continueDraw(context, drawCommand, shaderProgram, uniformMap) {
 
   if (defined(indexBuffer)) {
     offset = offset * indexBuffer.bytesPerIndex; // offset in vertices to offset in bytes
-    count = defaultValue(count, indexBuffer.numberOfIndices);
+    if (count) {
+      count =
+        count < indexBuffer.numberOfIndices
+          ? count
+          : indexBuffer.numberOfIndices;
+    } else {
+      count = indexBuffer.numberOfIndices;
+    }
     if (instanceCount === 0) {
       context._gl.drawElements(
         primitiveType,
@@ -1358,7 +1365,11 @@ function continueDraw(context, drawCommand, shaderProgram, uniformMap) {
       );
     }
   } else {
-    count = defaultValue(count, va.numberOfVertices);
+    if (count) {
+      count = count < va.numberOfIndices ? count : va.numberOfIndices;
+    } else {
+      count = va.numberOfIndices;
+    }
     if (instanceCount === 0) {
       context._gl.drawArrays(primitiveType, offset, count);
     } else {


cesium-concierge:
Thanks for the pull request @IKangXu!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the PR @IKangXu! 

@jjhembd Could you please review?

packages/engine/Source/Renderer/Context.js
```diff
@@ -1340,7 +1340,14 @@ function continueDraw(context, drawCommand, shaderProgram, uniformMap) {
 
   if (defined(indexBuffer)) {
     offset = offset * indexBuffer.bytesPerIndex; // offset in vertices to offset in bytes
-    count = defaultValue(count, indexBuffer.numberOfIndices);
+    if (count) {
```
jjhembd, line 1343: Can we use the `defined` method here, following the [Coding Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#basic-code-construction)? The `defined` function (like `defaultValue`) checks for both `null` and `undefined`.

jjhembd, line 1343: Can we use the `defined` method here, to conform to the [Coding Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#basic-code-construction)? The `defined` function (like `defaultValue`) checks for both `null` and `undefined`.

packages/engine/Source/Renderer/Context.js
```diff
@@ -1340,7 +1340,14 @@ function continueDraw(context, drawCommand, shaderProgram, uniformMap) {
...
+      count =
```
jjhembd, line 1344: `Math.min` might be more readable than a ternary expression here:

```javascript
count = Math.min(count, indexBuffer.numberOfIndices);
```

packages/engine/Source/Renderer/Context.js
```diff
@@ -1358,7 +1365,11 @@ function continueDraw(context, drawCommand, shaderProgram, uniformMap) {
       );
     }
   } else {
-    count = defaultValue(count, va.numberOfVertices);
+    if (count) {
+      count = count < va.numberOfIndices ? count : va.numberOfIndices;
```
jjhembd, line 1369: I think we should be comparing to `va.numberOfVertices` in this case.

jjhembd (changes requested):
Thanks for the PR @IKangXu! It does fix the bug on my machine.

I have a couple comments on the code. Once those are addressed, this should be ready to merge!

packages/engine/Source/Renderer/Context.js
IKangXu: Fixed.

packages/engine/Source/Renderer/Context.js
IKangXu: Fixed.

packages/engine/Source/Renderer/Context.js
IKangXu: Fixed.

IKangXu:
Hi @jjhembd, I have fixed and resubmitted it based on your suggestions.

engineerhe:
well done @IKangXu

Issue #11355, opened 2023-06-15
Tile debug labels in 3D Tiles Inspector not appearing


ptrgags:
Today I was inspecting a tileset in Sandcastle and noticed that the various tile debug labels (geometric error, URL, memory usage, etc.) are not showing up 

![image](https://github.com/CesiumGS/tilers/assets/8422414/734a96d1-27dc-476d-ac27-13e399d7ae78)

To reproduce:

1. Open a Sandcastle that uses the 3D Tiles Inspector (e.g. https://sandcastle.cesium.com/?src=3D%20Tiles%20Inspector.html&label=3D%20Tiles)
2. Turn on `Tile Debug Labels > Geometric error` (or one of the other label types)
3. Even after waiting a while, the labels do not appear.

The latest Sandcastle seems to be using Cesium 1.106. I can reproduce in both Chrome (114.0.5735.106) and Firefox (114.0.1) on Linux (Ubuntu 22.04.2)

PR #11293, opened 2023-05-18
Fixed label background rendering.

IKangXu:
If the initial label background color is set to false and then dynamically modified to true, rendering problems will occur.

Before repair
image

After repair
image

PR #11359, opened 2023-06-16
Fix 3D Tiles debug label rendering

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11355.

The issue was introduced in https://github.com/CesiumGS/cesium/pull/11293, but this should ensure both cases are now working. 

The debug labels are destroyed at the beginning of every frame-- Which doesn't leave any time for the promise in `Billboard.js` to resolve. This can be worked around by skipping the promise chain under specific conditions.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Thanks @ggetz! This resolves the issue on my machine.

Issue #11357, opened 2023-06-16
Rendering Stop when CZML data timeline is forwarded fast.


pyshx:
Rendering stops when a CZML data's timeline is being forwarded. We have observed this issue from v1.104 onwards. 

<img width="1509" alt="Screenshot 2023-06-15 at 12 07 06 PM" src="https://github.com/CesiumGS/cesium/assets/42397980/0df1063f-09b6-4287-8e64-9b5c170508e6">

Sandcastle example: https://sandcastle.cesium.com/index.html#c=bVLNjtowEH4VK6cgNfamUKDdLGpFpV6oemDVQxUJGXtSLBwb2RMQVH33Ok6gZHdzcma+P8+YMbJ0wBEI7oAswaumJkcFJ3ClOXLXn8kTMXDq+/RnrKVlIuL/0hrkygRGMnosTWnYTZQTcan1V458bRsnoNMc1obay0EvvQmuLJddxl/fV0QGBKmcrYNDpXTQHWpSHeAh4A7x4D8xxr0H9FTUnh50CMYb6gC4wx1Vtu8yUbGZmM6kzMRk/CGbwGSczUU+zaZyMoePsprOZ5yhqsGDU+A3+Th/eNig3Z/t+3G2bzbc8T0/8c32kE/mtI0URkJDapNWjRGorElH5E9pCAlX+iK7G70YB9pYvS6B9CcqbxBPuZTpkNYNKup+A4wKYFDhmQirNUTvbmKvLaNJ2EskhIuFjbwY57Xz32Tdmwheg+Nt6DbmzdcB8QIM3OW/WFs/2/Qq1Qb+O3pM3iWFx7OGRQttv8+qPliHpHE6pZQh1HFlnm0bsQekwkduCy3YPbWQ6kiUfHrjXRKhw5JDp2q0XqsLlMmiYAH/ito+HGV+/ziC0/zcwnb5YtUVKaUFC79vM9FaveXuTrlD/AM

Browser: Chrome Version v114.0.5735.133 / Firefox Version v114.0.1

Operating System: MacOS Ventura v13.4 / Windows 11 OS Build 22621.1848

ggetz:
Thanks @pyshx! I added a fix in https://github.com/CesiumGS/cesium/pull/11360. It should go out with the next release on July 3.

pyshx:
@ggetz Thanks for quick response and fix.

PR #11360, opened 2023-06-16
Fix crash with hidden 3D Tileset in CZML

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11357

There was a race condition when a 3D tileset embedded in a CZML data source was loaded and the visibility was set to false.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 ##### Fixes :wrench:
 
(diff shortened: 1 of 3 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Thanks @ggetz! The change makes sense, and fixes the problematic Sandcastle on my machine.

PR #11382, opened 2023-06-30
Fix crash due to terrain loading race condition

ggetz:
Follow up from https://github.com/CesiumGS/cesium/pull/11296

Technically after that change, due to [this line](https://github.com/CesiumGS/cesium/pull/11296/files#diff-39e084694ee46ce76a00b042bb4e510d79af3f07440a23a2248952da9757b182R941), it was possible to return a promise which resolves to undefined if a `requestTileGeometry` request gets throttled by the requestScheduler.

TODO:
- [ ] Unit test?


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/terrain-crash-fix/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Thanks @ggetz!

PR #11406, opened 2023-07-07
Vignesh patch

neshvig10:


cesium-concierge:
Thanks for the pull request @neshvig10!

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

CHANGES.md
```diff
@@ -4510,3 +4510,5 @@ _This releases fixes 2D and other issues with Chrome 29.0.1547.57 ([#1002](https
 ### a0 - 2011-06-17
 
 - Initial Release.
+- 
+-fixed a typo in line 88 of PolylineCommon.glsl
```
jjhembd, line 4514: This change should be listed at the top of the file, under "Fixes" in the 1.108 release. 
Also, I don't think this is really a typo. The old code was correct, but your code will be faster. Perhaps you could say "optimized angle calculations in Polyline shader"

Source/Shaders/PolylineCommon.glsl
```diff
@@ -85,7 +85,7 @@ vec4 getPolylineWindowCoordinatesEC(vec4 positionEC, vec4 prevEC, vec4 nextEC, f
     else {
         lineDir = normalize(nextWindow.xy - positionWindow.xy);
     }
-    angle = atan(lineDir.x, lineDir.y) - 1.570796327; // precomputed atan(1,0)
+    angle = czm_fastApproximateAtan(lineDir.x, lineDir.y) - 1.570796327; // precomputed czm_fastApproximateAtan(1,0)
```
jjhembd, line 88: If we replace the constant `1.570796327` with `czm_piOverTwo`, the code will be more self-explanatory, so the comment won't be necessary.

jjhembd (changes requested):
Thanks @neshvig10, this should be a good optimization! I just have 2 suggestions.

PR #11407, opened 2023-07-07
Fix OpenStreetMapImageryProvider usage in comments, change default url

fredj:
See https://github.com/openstreetmap/operations/issues/737 for the url change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Change Log
 
+### 1.108 - 2023-08-01
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Fixed `OpenStreetMapImageryProvider` usage in comments, change default url and add `tile.openstreetmap.org` to `RequestScheduler.requestsByServer`. [#11407](https://github.com/CesiumGS/cesium/pull/11407)
+
+#### @cesium/widgets
+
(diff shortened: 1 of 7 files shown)


cesium-concierge:
Thanks for the pull request @fredj!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/osm_usage_url/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Awesome, thanks @fredj! Please update `CHANGES.md` with a summary of the change.

> tile.openstreetmap.org supports [HTTP/2](https://en.wikipedia.org/wiki/HTTP/2) and [HTTP/3](https://en.wikipedia.org/wiki/HTTP/3) which no longer require the old (a|b|c) aliases to increase browser connection concurrency. Using a single URL improves performance and ability to cache.

Given this, I would also suggest adding `tile.openstreetmap.org` to `RequestScheduler.requestsByServer`. This will allow Cesium apps to take advantage of increased concurrency with HTTP2+ requests by default, resulting in better performance.

fredj:
@ggetz done

ggetz:
Awesome, thanks @fredj!

PR #11417, opened 2023-07-13
Move playwright install from postinstall to prepare

ggetz:
I had added the command `playwright install --with-deps` to `postinstall` which was incorrect-- This causes the release version to run the command despite `playwright` not being installed.

`prepare` is the correct script to use, as it is only run in dev, not in production. 

This can be tested using the following commands:
```
npm run pack
cp cesium-1.107.1.tgz ../<another external directory>
cd <another external directory>
npm install cesium-1.107.1.tgz
```

We're looking into adding a step to CI to catch any errors like this in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "yargs": "^17.0.1"
   },
   "scripts": {
-    "prepare": "gulp prepare && husky install",
+    "prepare": "gulp prepare && husky install && playwright install --with-deps",
     "start": "node server.js",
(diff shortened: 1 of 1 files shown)


kittenswolf:
Looked up this repository because this issue is causing a local project of mine to break (can't run `npm update` because I don't have playwright). Thanks for fixing it 👍

ggetz:
Sure! Latest version has been published to npm.

Issue #11174, opened 2023-03-20
Possible error in computation of geometric error for implicit tilesets


javagl:
Sparked by [a forum question](https://community.cesium.com/t/implicit-tiling-override-tile-geometric-error-semantic/22958/1), I created an implicit tileset that uses the `TILE_GEOMETRIC_ERROR` semantic to override the geometric error of certain tiles. The tileset is attached below.

Its structure is similar to the [`SparseImplicitQuadtree`](https://github.com/CesiumGS/3d-tiles-samples/tree/902ea3dca1821a9ef9d23d141f800c68627c452b/1.1/SparseImplicitQuadtree) sample, but with metadata in the subtree files that overrides the geometric error of the tiles. 

The geometric error is set to be **256** for all tiles up to level 3, and to **1** for all tiles in deeper levels.

Rendering the tileset in a debug sandcastle (shown below) gives this result:

![Cesium GeometricErrors](https://user-images.githubusercontent.com/5597569/226480087-87f3bab3-f813-4a60-9230-28f745904eb7.png)

Yeah. There's that 64. And there is no tile in that tileset that should have this geometric error.

A **really** quick search makes me think that this may be related to [this part in `Implicit3DTileContent.makePlaceholderChildSubtree`](https://github.com/CesiumGS/cesium/blob/9295450e64c3077d96ad579012068ea05f97842c/packages/engine/Source/Scene/Implicit3DTileContent.js#L1060), but that's only a guess for now.

---

Tileset:

[Cesium Implicit Tileset Forum 22958 quadMeta.zip](https://github.com/CesiumGS/cesium/files/11023541/Cesium.Implicit.Tileset.Forum.22958.quadMeta.zip)

Sandcastle:
```
const viewer = new Cesium.Viewer("cesiumContainer");

// Create the tileset, and set its model matrix to move it
// to a certain position on the globe
const tileset = viewer.scene.primitives.add(
  new Cesium.Cesium3DTileset({
    url: "http://localhost:8003/tileset.json",
    debugShowBoundingVolume: true,
    debugShowGeometricError: true,
    maximumScreenSpaceError: 256,
  })
);
tileset.modelMatrix = Cesium.Transforms.eastNorthUpToFixedFrame(
  Cesium.Cartesian3.fromDegrees(4.8933281, 52.3731013, 0.0)
);
const scale = 1.0;
tileset.modelMatrix = Cesium.Matrix4.multiplyByUniformScale(
  tileset.modelMatrix,
[...]

PR #11365, opened 2023-06-22
Allow using tile metadata to store a tighter bounding box for implicit tiling

ptrgags:
This PR + bit of metadata in the tileset allows `viewer.zoomTo()` to zoom to a tileset correctly when the root bounding box is loose.

My team's been working with implicit tiling octrees lately. For datasets that are wide but not very tall, we want to prevent super thin bounding volumes (which can be bad for culling), so we make the root bounding volume a cube instead, like this:

image

The problem is `viewer.zoomTo()` zooms to the bounding volume's bounding sphere (which is centered on the _center_ of the tile, which for the type of tileset above puts the camera way too high. Compare these screenshots/Sandcastles:

[...]

PR #11453, opened 2023-08-04
Fix TILE_BOUNDING_VOLUME handling for implicit tiling

ptrgags:
This fixes a problem I introduced in #11365.

When the `TILE_BOUNDING_VOLUME` semantic is added to an implicit root tile (to store a tight bounding volume), this value was being propagated to every transcoded tile. This causes a couple problems:

- Most critically, it results in all the tiles in a single LOD loading at the same time! this is because all the bounding volumes are the same, so the distance to camera in the screen-space error calculation is exactly the same for every tile at the same level (because they have the same geometric error based on how implicit tiling works)
- `debugShowBoundingVolume` and tile labels always show the root bounding volume in the inspector.

The fix: simply ignore explicit tile metadata before transcoding tiles. I think this makes sense in general since tile metadata for implicit tiling is supposed to come from the subtree file. @lilleyse let me know if this is a reasonable way to handle this from a 3D Tiles spec perspective (and do we need to make any clarifications in the spec?)

On Monday I want to do some more tests on more datasets and put together a sandcastle to help with the review, but otherwise the code is ready, I also included a unit test.

### DIFF
--- a/packages/engine/Source/Scene/Implicit3DTileContent.js
+++ b/packages/engine/Source/Scene/Implicit3DTileContent.js
@@ -536,8 +536,21 @@ function deriveChildTile(
   // user specified such as extras or extensions.
   const deep = true;
   const rootHeader = clone(implicitTileset.tileHeader, deep);
+  // The bounding volume was computed above since it may come from metadata
+  // in the subtree file.
   delete rootHeader.boundingVolume;
+  // Copying the transform to all the transcoded tiles would cause the transform
+  // to be applied multiple times. Removing it from the header avoids this issue.
   delete rootHeader.transform;
+  // The implicit tiling spec does not specify what should happen if explicit
+  // tile metadata is added to the placeholder tile. Since implicit tile
+  // metadata comes from the subtree file, ignore the explicit version.
+  //
+  // Also, when a property with the semantic TILE_BOUNDING_VOLUME is added to
+  // the placeholder tile to set a tight bounding volume (See Cesium3DTile.js)
+  // propagating it to transcoded tiles causes transcoded tiles to use the
+  // wrong bounding volume, this can lead to loading far too many tiles.
+  delete rootHeader.metadata;
   const combinedTileJson = combine(tileJson, rootHeader, deep);
 
   const childTile = makeTile(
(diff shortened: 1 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-tight-bounding-box/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
The approach here makes sense to me. Tile metadata on the root tile shouldn't be inherited by children.

javagl:
This talks only about the bounding volume, but ... could it be that this also fixes https://github.com/CesiumGS/cesium/issues/11174 ? (That issue was also about ~"something 'incorrectly' inherited" and ~"placeholder tiles" - at least, as far as I remember from trying to zoom into the relevant code paths...). In doubt, I'll try that out independently after this is merged...

ggetz:
Thanks @ptrgags! What do you think about @javagl's question?

@jjhembd Would you be able to give this a review?

ptrgags:
@javagl @ggetz in regards to #11174 I looked at the data in that issue, there's tile metadata in the subtree file, but no explicit tile metadata on the tile. Since this PR only removes explicit tile metadata, sadly I don't think this will address the GE issue. That said, I'll be testing some datasets with this PR today, I can try that dataset too.

ptrgags:
I can't share all the data but I can show some performance numbers I measured with the datasets anonymized. All around improvement!

Loading time:

| Dataset | main | this branch |
|---|---|---|
| Dataset A | N/A (overflowing tile cache) | 2.11 sec |
| Dataset B | 3.12 sec | 0.75 sec |
| Dataset C | 3.94 sec |  1.03 sec |

Tiles Loaded

| Dataset | main | this branch |
|---|---|---|
| Dataset A | 103/4845 | N/A (cache issue) | 240 / 3191 |
| Dataset B | 94/1500 | 289/1201 | 56 / 1201 |
| Dataset C | 113/374 | 345 / 346 | 49 / 346 |

And I confirmed that the debug bounding volumes and tile labels look correct now.

PR #11455, opened 2023-08-05
Make TS definition of `defined` use type predicate

### LINKED ISSUE #11419: Add type predicate typing for defined
Implement typing for `defined` as a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

Right now Typescript sees `defined` as a normal function and doesn't realize that a variable will not be undefined or null if inside of a `if(defined(x)) { ... }` block.  See [Typescript Playground example.](https://www.typescriptlang.org/play?#code/FAehAIEkFsAcBsCm1EDsAuBDdBLA9quHgGbgAmixOqiZ4ArgM7UDm4m46AnrIuLACdaOAMbZEALnAALdOliMJYAO6qAdN16MRAnLHTxMqFmrwCWIMnhGMQ0o2QBGePAGsQAJhCpMAgXmVWNVloeABiJlYAWk1EKMFhMXRERmBY8AA5PHQM+nh4LPQAVVQKKhoyAB4AFQA+cABecGrwRAAPZNLGBlLKalpwAB9wVDz4cAB+EcQAN0QBcClqgG5gYnpUEVwCcj6KgAV-WDxGWhgEZDQsbdQa2oAKZMZ0JaGesv66YdH8gEopJ7ocA4bqFXL5QolD4VO7gADewHA4CE6HoAkIP3GAEIGk1AeAAGQE957AY4vEpdCrAC+wFAEAAypgUOxutCBphnHNwI56EDUNkGMxjOxODw+AkyKJxGpwAAhTDMMT5LjgZT2IEg4GEADCKRw9GgnDwZEwXDWGy2+EI7LIOrRQgw5yQKAw2Gtj0pUiMXF+8MRyMQqPRIzG4HJnEphOJG1t4dxkeeNLpSCBgIAaph4PRJKHoI55oNY6SyINMasLZsbuQ8ClCgBBEQiRD6KEl+5cKSjfPzP0I2nAHCke62w54Y6nMjOy5um6e56Z7OIX59gNgZrSLXKMyubrC5u7coDLUUESGIR0RWi9KS6XJcDEMwjHaY9ilZ9A4tHsiy5ri7S6Po4ABlYdbZI2zatr037zugi45r8yZ0kO4AjiW9p+Fc06utcHoZlmCGrki67VP+Oh6JqjCoAA5ECIgEDM1DNnQ6AaomcEEXwWqYFs9BZvAqrQpeaabt0xzUOgIG1owDZNi2xTQZ8sHwcuwDUkAA)

mramato:
This is a simple fix for #11419 and overrides the TypeScript definition of `defined` to include a better but manually maintained version using Type predicates.

It also adds a test to ensure that if the code that creates the custom definition ever breaks, the test will break.

There are probably other places Cesium could improve it's type definitions, and the same strategy could be used to replace others as well. Longer term, it would be nice to have a custom JSDoc tag to allow for overriding the definition inline with the JSDoc that generates it.

Fixes #11419

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.109 - 2023-09-01
+
+#### @cesium/engine
+
+##### Additions :tada:
+
+- The TypeScript definition of `defined` now uses type predicates to allow TypeScript to use the result during compliation.
+
 ### 1.108 - 2023-08-01
 
 #### Major Announcements :loudspeaker:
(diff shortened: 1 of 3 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

mramato:
@angrycat9000 Since you were asking about this on Friday. Also could be a good reference PR if there are other places you notice we could improve definitions in the future.

javagl:
@mramato _Some_ aspects of the type inference are hard to explain there. But if the definition should be changed, just drop me a note.

mramato:
@javagl I actually didn't expect my PR to be merged so quickly, I'm all for improving the definition further and your example above is an issue we should probably fix. Can you just open a new PR with your suggestion?

javagl:
@mramato Done at https://github.com/CesiumGS/cesium/pull/11462

Issue #11459, opened 2023-08-07
Terrain skirts are causing artifacts in terrain shadows


calogeromauceri:
When terrain shadows are enabled, the skirts contribute in casting shadows causing shadows artifacts.

![Screenshot 2023-08-07 16 56 34](https://github.com/CesiumGS/cesium/assets/65298918/6e1f1f03-a761-4c2d-8e31-3516a2b3f76f)

This is better visible when the light source is tangent to surface (e.g. it is near the horizon).

I'm attaching a simple sandcastle example for demonstrating the current behavior: advance the time through the timeline for seeing the shadows artifacts caused by the skirts. Disabling the terrain skirts gets rid of some of those artifacts, but that of course has some other bad side effects, e.g. worse tiles edges connection

![sandcastle](https://github.com/CesiumGS/cesium/assets/65298918/4caf153a-d4cd-49ff-8312-096786947801)

Sandcastle example:
[...]

PR #11460, opened 2023-08-07
Enhance tile skirts light occlusion

calogeromauceri:
I'm proposing the following solution for this issue [#11459](https://github.com/CesiumGS/cesium/issues/11459)
The tile skirst are causing artifacts on shadows when the light is tanget to the tile, i.e. when it is near the horizon.

The tangent light should be occluded, but it leaks from below the terrain tile. The skirt is casting a thin shadow that is causing artifacts on the final result.

The solution I'm proposing is to increase the skirt size an to add a plane to the bottom of the tile, in that way the light cannot pass through the side/bottom of it and the terrain is better shadowed.

Here is a sandcastle example with the default behavior (advance the time till the artifacts are visible)
[sandcastle](https://sandcastle.cesium.com/#c=lVTJbtswFPwVwicbUGjKslYvbTYUKZL2YKMBCl0YiraJUKRBUnEW5N9LbV4SJ20FH0Ry3nDmafyIFNqAB0Y3VIEJEHQDzqlmRQ5/VXvdtEOq9bkUBjNBVdpxwEsqAGBiIc/kYwIWmGvqlFuackoMk+JKZIxgI9XB6QpncqMTYFRR7xiqlCWdtQfN1fX6RmYUXv44Pbu+vGjqZcGzU8FybOh7lm35vF7DhZL5rVQ8aza6PQt/7Y1SkQpS+a4V3eC1tV73AG63LGz7DnP8yPIiv2DaYEGohQ+QfSCqyFIxwyIjWBtOIc6yuVwuOT0rjJHCNrCxVzbu4BLtgEUhqn6BbmmErCi5p1kqenWHD9H20gYw2tr4/GZLkXYa82CnYo/68AOAyWTyl2+wE9xoaaR+yLgzBarny8cXtJBjObi4mlWYUQl6tR36V/uze6bMG9eaUEHhkss7apsrNzWmQuxsTabg5VP8ztuenqaCcEnuoQ2LMnOWl3lpLH0vOMPiwga4iueVllGAXBuSARp4JyiyvzmKE4SSMl0I/U47JfUbWrn+f1bX+5yVFEpR0co97mNXY+yK23EAn6XM57J7HO+Ao8LbZvX7YEbr6QM2zKzAiuKMiaUD1syQFbDfFijJ+U4pzqnCUFNTTqdulbyMasMELkOZ7M+vc6vBvmHhddtgeWgQhyGMgzCK42gYBA7wAy9wkQu9GA2Q64cOsJDY93zo+rEX+0HgOW15Q3zJOVtryTJ4+20WDevTXoWSitkWNlqa/0XjKQEu9P1hFKEYRR6KEApb5spsAk4Q9MPAdYchclGAhoEXN4CyBwkIoe+iyPUD14viyEUePYkqwGs91jpOZ6zNE6fTVvBXlq+lMqBQvAth39B8zW1GdP+usLE1kGjdG9XgcX+/dJyxB8CyyZHhDwjHWtuTRcH5jD3TtDMd9y3+XSmXlfGfD1Rx/FTCVu70ut6EEI77dnm80kjJ77B6w/wH)

sandcastle

and this is the result with the proposed solution

sandcastle_desired

Here is an example of the tile with the longer skirts

Screenshot 2023-08-07 17 47 21
Screenshot 2023-08-07 17 47 27

Let me know if the proposed fix makes sense or if there is any better solution

### DIFF
--- a/packages/engine/Source/Core/QuantizedMeshTerrainData.js
+++ b/packages/engine/Source/Core/QuantizedMeshTerrainData.js
@@ -522,6 +522,7 @@ QuantizedMeshTerrainData.prototype.upsample = function (
     ? this._northSkirtHeight
     : shortestSkirt * 0.5;
   const credits = this._credits;
+  const extendedSkirts = this._extendedSkirts;
 
   return Promise.resolve(upsamplePromise).then(function (result) {
     const quantizedVertices = new Uint16Array(result.vertices);
@@ -556,7 +557,7 @@ QuantizedMeshTerrainData.prototype.upsample = function (
       childTileMask: 0,
       credits: credits,
       createdByUpsampling: true,
-      extendedSkirts: this._extendedSkirts,
+      extendedSkirts: extendedSkirts,
     });
   });
 };


cesium-concierge:
Thanks for the pull request @calogeromauceri!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/skirts/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

lilleyse:
I wonder if this can also be solved by disabling back face culling for terrain shadow map commands, i.e. set [this line](https://github.com/CesiumGS/cesium/blob/9b0c4a1c02243b7806eadab92121b10b4e467fee/packages/engine/Source/Scene/ShadowMap.js#L308) to false.

calogeromauceri:
thanks @lilleyse for looking into this. I tried the solution you suggested, but it seems to me disabling it does not work well in some circumstances.
For example in the sandcastle example I posted above, I see light leaking where there should not be

Screenshot 2023-08-17 14 45 44

and here is another view where the skirts artifacts are still visible

Screenshot 2023-08-17 14 56 10

ezgif com-video-to-gif


[sandcastle](https://sandcastle.cesium.com/#c=lVVdb9owFP0rFk8gpcYkIQEK3dpSTZ3a7QG0SlNe3MSAVcdGtlP6of733TiBlJZ2W0Aots8995zra5MqaSy652zDNJogyTbonBle5PiXm2snrdSNz5W0lEumk5aHnhOJEJcLdaYeRmhBhWFeOWWYYKnlSl7KjKfUKr23uqKZ2pgRsrqoZizTGkhn24U6dTW+VhnDFz9Oz64upnW8KkR2KnlOLXvPsgufV2O80Cq/UVpk9US7A/CXznEiE5k635Wia7oG61UN8G4KYLt3nNMHnhf5lBtLZcoA7hN4MAFUt9vw4AXNuFxeSHorWAYw595lTOSMyiylxgqGaZbN1XIp2FlhrZJQ5boGZXX3lBgPLQrpiorapdt0xdI7liWyU23DPhpS1oDjndfPMwNF0qorhBoVr6j3dwlNJpO/bFQjuNZSS/2QsTGF3PPl4wRbyKFmmV7OHOa4BL1Ahf7V/uyOa/vGtUmZZHgp1C2D4qpNhXGIxtbkBD1/im+8vdJTR6RCpXcYOkrbOc/LpqotfS8Ep3IKXe56+NKoQUR60CQ+8YMjMoDvnAxHhIzKFiTkd9Iqqd/QqvX/s/aCz1nTQmsmt3IP+2hiLIwE3Bn4Sal8rtqH8R46KHxbLDheM1ZdUWjD7QqtmDtjHlpzm64Q7C3SSohGKc2ZptgwW15hbdd5GTOWS1o25ej1JXcOGuCNyqC9bayA+FEU4jgm/Sjsx37U91A/CiI/IrisVNQLg9hD/jAehkMfR/2w78fxYOBtCWrqCyH42iie4Ztvs0FYrXYcSmkORazV1CejdjVCPRwFcRQEUQC5yKDf2zI7uyN01AMZIXxIqTSAn3q9LMIIua0jZEiifjAICAn9aBgHocO8NPdfy2uNjX0U7GQr+yvP10pbVGjRxrhrWb4W0Cume1tA+1qcGtM5rsDj7uvQccbvEc8mB/4pUCqoMbCyKISY8SeWtE7GXcC/CxXK2f95z7SgjyVs1Tu5qiYxxuMuDA9HWqXELdVvmP8A)


Using the longer skirts approach I suggested, seems to produce better shadows

ezgif com-video-to-gif (2)

cesium-concierge:
Thanks again for your contribution @calogeromauceri! 

malaretv:
@lilleyse @ggetz are there any blockers to merging this in? Or any alternate approaches worth trying? Would love to collaborate on them if so

ggetz:
@lilleyse Do you have any other suggestions on the approach here? Or is this the most straightforward given the backface culling idea doesn't seem to be working in all cases?

lilleyse:
Is it possible to enable long skirts and bottom plane only if terrain shadow casting is enabled? That way this change doesn't affect performance in the general case.

calogeromauceri:
@lilleyse @ggetz thanks for looking into this. I'll update the code as you suggested

calogeromauceri:
@lilleyse I've had the chance to delve into the implementation of the changes you suggested regarding enabling long skirts and a bottom plane exclusively when terrain shadow casting is enabled. However, since the tile mesh, including skirt vertices, is generated at load time, dynamically adjusting them based on shadow casting isn't feasible.

As an alternative, we could add a flag to the terrain provider upon creation that enables these features. Do you think this solution fits our needs or do you think there are better alternatives?

lilleyse:
> As an alternative, we could add a flag to the terrain provider upon creation that enables these features. Do you think this solution fits our needs or do you think there are better alternatives?

That would be ok, I think.

calogeromauceri:
@lilleyse @ggetz I added a flag called `extendedSkirts` to CesiumTerrainProvider for enabling extended skirts. The flag is false by default. If it is set to true then the extended skirts are used: longer skirts and bottom tile plane for better light occlusion when the terrain shadows are enabled.

Some tests are failing, in parts that are not related to my changes

calogeromauceri:
@lilleyse @ggetz I finally found some time for checking and fix the tests failing.
I would appreciate your feedback on these changes

jjhembd:
@mzschwartz5 what are your thoughts on this PR? It looks like a significant improvement to shadow quality, but the merge is complicated after https://github.com/CesiumGS/cesium/pull/9961.

calogeromauceri:
I there is interest I can update the PR to work on the latest code, just let me know

mzschwartz5:
Honestly, here are my thoughts:

1. I _don't_ think the merge would be terribly complicated.
2. However, I share the hesitancy of past reviewers. This is a complex bandaid, and I feel it's likely to cause other rendering artifacts in different workflows down the road.
3. Shadow issues appear to be much broader than what's identified here. I don't know if this is a result of further regressions since this PR was opened, but playing around with the provided sandcastle, there's a ton of flickering / popping and other weird things going on with shadows (why are they so highly view dependent?).

I do think this issue deserves a more root-cause-oriented treatment. Skirts shouldn't be casting shadows at all; so lengthening a skirt certainly shouldn't be the solution to missing shadows.

PR #11462, opened 2023-08-09
Generate stricter type predicate for 'defined'

javagl:
A follow-up for https://github.com/CesiumGS/cesium/pull/11455 , with some details and an example described in https://github.com/CesiumGS/cesium/pull/11455#issuecomment-1668310430 : The previous type predicate that was added to `defined` did not work when the type already was `undefined`, because the type predicate would only result in `undefined` again. When using `NonNullable<Type>` in the type predicate, then `undefined` will be constrained to `never`, causing the compile-time error check that one might expect in this case.

### DIFF
--- a/Specs/TypeScript/index.ts
+++ b/Specs/TypeScript/index.ts
@@ -394,3 +394,10 @@ let pos: Cartesian3 | undefined | null;
 if (defined(pos)) {
   pos.clone();
 }
+function consumeDefined(pos: Cartesian3) {
+  pos.clone();
+}
+pos = undefined;
+if (defined(pos)) {
+  consumeDefined(pos);
+}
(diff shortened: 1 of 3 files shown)


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @javagl!

@mramato or @angrycat9000 Would you mind taking a pass on this?

Specs/TypeScript/index.ts
```diff
@@ -394,3 +394,10 @@ let pos: Cartesian3 | undefined | null;
 if (defined(pos)) {
   pos.clone();
 }
+function consumeDefined(pos: Cartesian3) {
+  pos.clone();
+}
+pos = undefined;
+if (defined(pos)) {
+  consumeDefined(pos);
+}
```
angrycat9000, line 403: Why are these checks needed?  They didn't fail on main and seem to be doing the same thing the lines above, but with an extra function.

angrycat9000 (commented):
Thanks for  updating this to catch the edge condition.  Had one comment about the test code.

Specs/TypeScript/index.ts
javagl: This check is intended to point out the difference in the type predicate. 

With the original 
`export function defined<Type>(value: Type | undefined | null): value is Type` 
in `Cesium.d.ts`, the call to `consumeDefined` there would complain and say

> Argument of type 'undefined' is not assignable to parameter of type 'Cartesian3'.

because the type of `pos` would be inferred to `undefined` there, and not prevent the code from entering the `if`-block. With the new one, the type is inferred to `never`. This is assignable to `Cartesian3` (but never has a value). 

Maybe whether or not this should cause a compile-time error is a matter of subjective perference?

javagl: To unfold that 'subjective' part from the last comment a bit:

One could make a case that the compiler _should_ complain about this pattern: When the compiler can determine that the value **is** definitely `undefined`, then the `if (defined...` check does not make sense. But ... for the case or the `pos.clone()`, this already applied to the state from the previous change. The new one just extends this behavior to functon calls, so to speak.... 

(BTW: Both PRs may cause a compilation to fail that _did_ compile _before_ this change - but maybe that's obvious)

PR #11489, opened 2023-08-29
Large Polygons Part 1: Stereographic coordinates, polygon bounding rectangles

ggetz:
Fixes #8033

This is PR 1 in a series which will fix issues related to large polygon rendering. It:
* Adds `Stereographic` class. A stereographic coordinate system is a 2D coordinate system in which cartesian points are projected from the south pole to a plane tangent to the north pole, or vice versa. This 2D system is useful for polygons as it preserves angles between positions. Here we use it to compute winding order and to determine whether a pole lays inside or outside the polygon. Later we will use it for operations like splitting polygons.
* Replaces the existing rectangle calculation for polygons. It did not handle going over the poles, which could result in visual artifacts.
* Deprecate the old public function as it was not accurate.

TODO:
- [x] Fix existing unit tests
- [x] Fix `build-ts`
- [ ] Once merged, open issue for `PolygonGeometry.computeRectangle` has been deprecated. It will be removed in 1.112. Use `PolygonGeometry.computeRectangleFromPositions` instead.

Opening as a draft PR, as these should likely be merged for the 1.110 release as reflected in `CHANGES.md`.

PR #11497, opened 2023-08-31
Large Polygons Part 2: Triangulate using stereographic coordinates

### LINKED ISSUE #4871: Large polygon rendering/fill issue
Large polygons spanning >90degrees across the longitude do not get rendered/filled properly. 

Here's an example.  

http://cesiumjs.org/Cesium/Apps/Sandcastle/?src=Hello%20World.html&label=Showcases&gist=fd0acbcf4d5fa84a9d988920bdb870f2

```
var viewer = new Cesium.Viewer('cesiumContainer');

var redPolygon = viewer.entities.add({
    name : 'Red polygon on surface',
    polygon : {
        hierarchy : Cesium.Cartesian3.fromDegreesArray([-179.0, 80.0,
                                                        -179.0, 0.0,
                                                        -90.0, 0.0,
                                                        0.0, 0.0,
                                                        90.0, 0.0,
                                                        179.0, 0.0,
                                                        179.0, 80.0,
                                                        90.0, 80.0,
                                                        0.0, 80.0,
                                                        -90.0, 80.0,
                                                        -179.0, 80.0]),
        material : Cesium.Color.RED
    }
});
```
Note the rendering artifact at the north pole and also note that only the region east of the meridian is filled. It almost seems like it's rendering the opposite region west of the meridian.

ggetz:
Fixes #4871

Building on https://github.com/CesiumGS/cesium/pull/11489, this addresses triangulation issues with large polygons.

We use a library called `earcut` to triangulate polygon, or break them down into triangles, so that they can be rendered in 3D. Earcut takes polygon positions in 2D as input. In order to get these positions previously, we would always project to a plane tangent to the ellipsoid. This worked great for polygons that did not span a large extent around the globe, as the projection produced little distortion. However for larger polygons that span 90 degrees or more, the projection can create distortions. For extents of more than 180 degrees, it would straight up incorrect as positions "wrapped back" on themselves in 2D.

This PR uses stereographic coordinates as input for earcut for polygons with extents of larger than 90 degrees. The stereographic projection preserve the earths curvature as well as the angles between the positions, ensuring that triangulation are valid for these large extents, even if they wrap around the globe.

The catch with stereographic projections is that that are only valid for a little greater than one hemisphere. To handle this, we split polygons of 90 degree or larger extent that cross the equator, making sure to handle the correct arc type for each edge. 

TODO:
- [x] Merge https://github.com/CesiumGS/cesium/pull/11489
- [ ] Performance stats
- [x] Fix existing unit tests


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
@GatorScott Would you have any additional test data besides what was provided in https://github.com/CesiumGS/cesium/issues/4871, especially something that may be performance sensitive? 

This branch is also available for your or any community member's testing at [this link](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/polygon-triangulation/Apps/Sandcastle/index.html).

ggetz:
Gentle bump for @GatorScott or anyone else who may have data or want to test themselves.

packages/engine/Source/Core/PolygonGeometry.js
```diff
@@ -1377,13 +1493,12 @@ PolygonGeometry.createGeometry = function (polygonGeometry) {
     return;
   }
```
jjhembd, line 1495: This looks like an existing issue, but the above call to `PolygonGeometryLibrary.polygonsFromHierarchy` (line 1368) is missing the `ellipsoid` argument. I'm not sure why this hasn't been a problem so far.

packages/engine/Source/Core/PolygonGeometryLibrary.js
```diff
@@ -466,12 +469,294 @@ PolygonGeometryLibrary.polygonOutlinesFromHierarchy = function (
...
+    const edgesOnPlane = computeEdgesOnPlane(positions, ellipsoid, arcType);
+    // If nothing intersected (no point were added), or there is only a single point on the plane, use the original polygon
+    if (positions.length === outerRing.length || edgesOnPlane.length <= 1) {
+      result[currentPolygon] = positions;
+      ++currentPolygon;
+      continue;
+    }
+
+    // Step 2: Sort the edges along the split line by the distance between their starting points and the starting point of the split line.
+    edgesOnPlane.sort((a, b) => {
```
jjhembd, line 733: Why do we need to sort here? (I don't have access to Vaněček's paper). In `wirePolygon`, we already do a `.findIndex()` to look up the appropriate edge.

packages/engine/Source/Core/PolygonGeometryLibrary.js
```diff
@@ -466,12 +469,294 @@ PolygonGeometryLibrary.polygonOutlinesFromHierarchy = function (
...
+    const swap = minLongitude;
+    minLongitude = maxLongitude;
+    maxLongitude = swap;
+  }
+
+  if (
+    intersection.longitude < minLongitude ||
+    intersection.longitude > maxLongitude
+  ) {
+    return;
```
jjhembd, line 506: I'm trying to understand why the intersection might be outside the longitude range of the line. Does this only happen if the line spanned more than 180 degrees of longitude? Does `EllipsoidRhumbLine.prototype.findIntersectionWithLatitude` already handle the IDL correctly?

packages/engine/Specs/Core/PolygonGeometryLibrarySpec.js
```diff
@@ -0,0 +1,698 @@
+import {
+  ArcType,
+  Cartesian3,
+  Ellipsoid,
+  Math as CesiumMath,
+  PolygonGeometryLibrary,
+} from "../../index.js";
+
+describe("Core/PolygonGeometryLibrary", function () {
```
jjhembd, line 9: What happens if a user supplies four points to describe a rectangle spanning >180 degrees of longitude? Do we automatically assume min/max are flipped, and treat it as spanning <180 degrees?

jjhembd (commented):
Thanks @ggetz, this is a big improvement!

I tested some large polygons in the [Drawing on Terrain Sandcastle](https://sandcastle.cesium.com/?src=Drawing%20on%20Terrain.html). Here is the (problematic) result for a triangle around the North Pole in the main branch:

image

The same test in this PR looks much better:

image

I have a few questions about the code, but this should be pretty much ready to go! :tada:

packages/engine/Source/Core/PolygonGeometryLibrary.js
ggetz: We sort the edges based on their signed distance along the splitting plane. In practice for us, this ends up being counter-clockwise along the equator. 

Having the edges in order ensures that we have the appropriate next edge spatially.

So once we find the index, `i`,  with `findIndex`, we now know that the spatially adjacent edge is at index `i+1`.

packages/engine/Source/Core/PolygonGeometryLibrary.js
ggetz: So how we currently handle intersections with `EllipspoidRhumbLines`, we use the entire rhumb line circle or spiral. It is not bound by the start and end points used to created the rhumb line. This check ensures the intersection returned _is_ bound by those end points.

packages/engine/Specs/Core/PolygonGeometryLibrarySpec.js
ggetz: We assume the shortest path. That would be assuming the min/max are flipped. This is consistant with the rest of the Cesium API. 

You can get around this behavior by supplying additional polygon positions that do not span >180 degrees.

packages/engine/Source/Core/PolygonGeometry.js
ggetz: Good catch. Likely, it doesn't come up that ellipsoid is not WGS84 in most cases, but it's best practice to pass it through.

ggetz:
Thanks @jjhembd! I appreciate the review and clarifying questions! I've responded to all of your feedback above.

jjhembd (approved):
Looks good, thanks @ggetz! Merging now.

PR #11517, opened 2023-09-12
Fix invalid metadata type in `buildings-metadata.gltf`

j9liu:
One of the test models had an incorrectly typed class property, which I caught while loading it in Cesium for Unreal. It doesn't seem like this was intentional, so I fixed it to conform to the 3D Tiles 1.1 spec.


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-spec-data/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks @j9liu!

Issue #11505, opened 2023-09-05
Uncaught TypeError: Failed to execute 'importScripts' on 'WorkerGlobalScope': Module scripts don't support importScripts()


aerialist7:
Hi Team,

I faced with `importScripts` problem after upgrading from version 1.108 to version 1.109.
I use a CDN version of Cesium.

I prepared a simple example to reproduce the problem.
Just check the browser console.

Sandcastle example: 
 * not reproduced on Sandcastle but reproduced in the simplest example – https://aerialist7.github.io/cesium-sample/
 * source code here – https://github.com/aerialist7/cesium-sample

Browser: 
 * Chrome 116.0.5845.140 (Official Build) (arm64)

Operating System: 
 * macOS 13.5.1 (22G90)

aerialist7:
@ggetz, hi! Could you please have a look?
It looks like it's related to this change #11400

ggetz:
Thanks for reporting this @aerialist7. We're taking a look now. This appears to be an issue with cross-origin workers specifically - In this case between https://aerialist7.github.io/ and the CDN.

ggetz:
@aerialist7 I have a branch,  `worker-inline`, which should resolve issues with the cross-origin workers. 

Could you try `http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/worker-inline/Build` in place of `https://cesium.com/downloads/cesiumjs/releases/1.109/Build/`?

aerialist7:
@ggetz, there is a problem with HTTP
![image](https://github.com/CesiumGS/cesium/assets/32072831/14130711-4a33-4d25-a5b3-7b9570a31680)

Do you have an ability to deploy fixed version somewhere under HTTPS?

protectione055:
> @aerialist7 I have a branch, `worker-inline`, which should resolve issues with the cross-origin workers.
> 
> Could you try `http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/worker-inline/Build` in place of `https://cesium.com/downloads/cesiumjs/releases/1.109/Build/`?

I encountered the same issue when running [Quickstart](https://cesium.com/learn/cesiumjs-learn/cesiumjs-interactive-building/#step-2-add-cesium-osm-buildings-and-cesium-world-terrain), and the browser console displayed these errors:
```plaintext
Uncaught TypeError: Failed to execute 'importScripts' on 'WorkerGlobalScope': Module scripts don't support importScripts().
[...]

Issue #10619, opened 2022-08-02
Inline Static Assets in Builds


ggetz:
The current CesiumJS distribution consists of a single, bundled JS entry point, `Cesium.js`, which inlines all library JS code. It also ships with several other static files, including standalone `Worker` JS files, third party `wasm` files, JSON data, images/textures, and CSS files. Depending on what features of the library is being used, some or all of these static files are required to be loaded at runtime.

While we've operated with these static files as part of the distribution since the beginning, it can require extra infrastructure on the part of developers depending on CesiumJS. We'd like to re-evaluate our strategy and look into providing a single-file CesiumJS distribution

As a first step, we could allow an optional `inline` flag as part of the build script. esbuild [has a `loader` option](https://esbuild.github.io/api/#loader) which allows other file types besides JS to be inlined most likely as data urls.

After ensuring everything works, we should put some thought into potentially shipping by default as part of the release zip.

ggetz:
Tangential to this discussion, for those that are using CesiumJS from `Source`, we'll also want to ensure [the proper files are exported via `package.json`](https://github.com/CesiumGS/cesium/issues/9212).

ggetz:
@TJKoury You have some experience with doing this, via rollup I believe, in https://github.com/DigitalArsenal/c137.js. Are there any gotchas that you may have experienced in the process?

TJKoury:
Plenty gotchas!  I'll start with the big one (for me): WebWorkers.

My approach was to import [all the webworkers](https://github.com/DigitalArsenal/c137.js/blob/main/src/workerPrep.ts) into a single file.  The packager can walk the dependency tree and pull everything into a single worker that can be called whenever you need a worker.

That is then injected as a string [here](https://github.com/DigitalArsenal/c137.js/blob/main/src/rollup-plugins/rollup-plugin-prep-source.ts#L29) into TaskProcessor.js.  Since this will now be part of the official build you can probably come up with a more elegant solution (import it using the esbuild loader option per the above).

TJKoury:
@ggetz I’m considering applying for a grant through the new program so I can dedicate some time towards adapting my old approach to the current build system.  If this will be duplicating effort, let me know.

ggetz:
@TJKoury Awesome! Please go ahead!

TJKoury:
@ggetz I'm assuming since it's now in the backlog that this is no longer a priority?

ggetz:
@TJKoury this is still on our shortlist. 

[...]

PR #11519, opened 2023-09-14
Inline workers into Cesium.js

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11505

Part of https://github.com/CesiumGS/cesium/issues/10619

This inlines all the worker files into `Cesium.js` rather than having `Cesium.js` request each additional worker file. This is appealing because:

* When using `Cesium.js` as a CDN as it reduces the number of additional files requested in turn
* It simplifies deployment for those not using tooling, as the built `Workers` directory does not need to be served, and because cached versions of Cesium.js and the workers will always be the same.
* Theoretically, since network bandwidth will not be used requesting these additional files, time to see content on screen can be reduced.

The CesiumJS build will still ship individual worker modules for those using the ESM version of the library.

| | **Before** | **After** |
| --- | --- | --- | 
| `CesiumUnminified/Cesium.js` | 9.6MB | 12.8MB |
| `CesiumUnminified/Workers` | 2.2MB |  |
| `Cesium/Cesium.js` | 3.8MB | 4.9 MB |
| `Cesium/Workers` | 0.9MB |  |
| `Cesium-1.*.zip` | 68.1MB | 70.2MB |

To test:
* run `npm run make-zip`
* ensure `Build/Cesium/Cesium.js` and `Build/CesiumUnminified/Cesium.js` work
* run `npm start`
* ensure each link on `localhost:8080` still works as expected

TODO:
- [x] Fix specs
- [x] Verify and document built file sizes


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/worker-inline/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd:
Thanks @ggetz. 
I'm getting HTTP 404 errors for http://localhost:8080/Build/Apps/Sandcastle/index.html and http://localhost:8080/Build/Apps/CesiumViewer/index.html.

ggetz:
Thanks @jjhembd! The issue also appears to be happening in `main` and is unrelated to the above changes.

Running `build-apps` requires the `Build/Cesium` directory to exist. You can build it by running `make-zip` before `build-apps`. I think this is confusing and I'll document the issue, but I don't think it should hold up this PR.

jjhembd (approved):
Thanks @ggetz! It's working for me now, on both Firefox and Chrome.

selimsezr:
Hi everyone,  I am currently encountering this problem while running. How can I solve?
image

ggetz:
Hi @selimsezr, do you have any info about your build process? Perhaps a webpack config file?

rodrigonzalz:
Hi, I am getting same error as @selimsezr 
@ggetz In my case I use Cesium with NextJS, have the webpack setup as described on [cesium-webpacl-example](https://github.com/CesiumGS/cesium-webpack-example ) and it has been working fine until now.
After finding this change, I have removed Workers from the CopyWebpackPlugin and everything is working now.

selimsezr:
Hello, 

I also asked the link [here](https://community.cesium.com/t/cesium-ion-1-110-webpack-problem/27351) and found my solution myself and shared it. I hope it helps everyone

ashley-mort:
In Cesium 1.115, is there a way to import (as ESM) such that it doesn't load the ~50 worker/chunk resources individually upon first polygon draw? I am doing:

import * as Cesium from './Cesium/1.115/Cesium/index.js'; //this is from the downloaded Build directory

But it's still loading a ton of resources when the first geometry is drawn. Is there a way to avoid this yet?

Issue #9815, opened 2021-09-17
PrimitiveCollection primitiveAdded / primitiveRemoved events


chris-cooper:
We've hit a couple of situations where it's useful to know when primitives have been added or removed from the scene.  In our fork we've added some events for this similar to what is seen in `DataSourceCollection` etc.  Is this something already on the roadmap?  Is there already another way we should be doing this?  Do you want us to contribute?

```
  /**
   * An event that is raised when a primitive is added to the collection.
   * Event handlers are passed the primitive that was added.
   * @memberof PrimitiveCollection.prototype
   * @type {Event}
   * @readonly
   */
  primitiveAdded: {
    get: function () {
      return this._primitiveAdded;
    },
  },

  /**
   * An event that is raised when a primitive is removed from the collection.
   * Event handlers are passed the primitive that was removed.
   * @memberof PrimitiveCollection.prototype
   * @type {Event}
   * @readonly
   */
  primitiveRemoved: {
    get: function () {
      return this._primitiveRemoved;
    },
  },
```
etc...

PR #11531, opened 2023-09-21
Primitive collection events

chris-cooper:
This adds `primitiveAdded` and `primitiveRemoved` events to `PrimitiveCollection` as discussed in #9815 .  The implementation is similar to [`DataSourceCollection`](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/DataSources/DataSourceCollection.js#L33-L57).

This change has been in our fork for some time.

### DIFF
--- a/packages/engine/Specs/Scene/PrimitiveCollectionSpec.js
+++ b/packages/engine/Specs/Scene/PrimitiveCollectionSpec.js
@@ -651,13 +651,14 @@ describe(
     });
 
     it("fires an event on primitive added", function () {
+      const p = createLabels();
       const primitiveAdded = primitives.primitiveAdded;
 
       const onPrimitiveAddedSpy = jasmine.createSpy("event");
       primitiveAdded.addEventListener(onPrimitiveAddedSpy);
       expect(onPrimitiveAddedSpy).not.toHaveBeenCalled();
-      primitives.add(createLabels());
-      expect(onPrimitiveAddedSpy).toHaveBeenCalled();
+      primitives.add(p);
+      expect(onPrimitiveAddedSpy).toHaveBeenCalledOnceWith(p);
     });
 
     it("fires an event on primitive removed", function () {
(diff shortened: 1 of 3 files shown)


cesium-concierge:
Thanks for the pull request @chris-cooper!
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/primitive-collection-events/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

packages/engine/Specs/Scene/PrimitiveCollectionSpec.js
```diff
@@ -649,6 +649,42 @@ describe(
...
     });
+
+    it("fires an event on primitive added", function () {
+      const primitiveAdded = primitives.primitiveAdded;
+
+      const onPrimitiveAddedSpy = jasmine.createSpy("event");
+      primitiveAdded.addEventListener(onPrimitiveAddedSpy);
+      expect(onPrimitiveAddedSpy).not.toHaveBeenCalled();
+      primitives.add(createLabels());
+      expect(onPrimitiveAddedSpy).toHaveBeenCalled();
```
ggetz, line 660: I would suggest using `toHaveBeenCalledWith` here and below to ensure the primitives were indeed passed to the callback.

ggetz (commented):
Thanks @chris-cooper! I think this is a sensible and low-impact addition to the Primitive API.

I just have one suggestion to make the unit tests a bit more thorough. Otherwise, the code is looking good!

packages/engine/Specs/Scene/PrimitiveCollectionSpec.js
```diff
@@ -649,6 +649,51 @@ describe(
...
+      primitiveRemoved.addEventListener(onPrimitiveRemovedSpy);
+      expect(onPrimitiveRemovedSpy).not.toHaveBeenCalled();
+      primitives.add(p);
+      expect(onPrimitiveRemovedSpy).not.toHaveBeenCalled();
+      primitives.remove(p);
+      expect(onPrimitiveRemovedSpy).toHaveBeenCalledOnceWith(p);
+    });
+
+    it("fires events on primitive remove all", function () {
+      const collection = new PrimitiveCollection({ destroyPrimitives: false });
```
chris-cooper, line 678: I'm manually destroying as the `toHaveBeenCalledWith()` test was triggering a 'this primitive has been destroyed' error.

chris-cooper:
Good idea thanks @ggetz .  It should be ready for another look now.

packages/engine/Specs/Scene/PrimitiveCollectionSpec.js
ggetz: Should this be noted in the `primitiveRemoved` docs? This seems something likely for devs to run into during development as well.

ggetz:
Thanks for the update @chris-cooper! I have one more comment based on your explanation above.

chris-cooper:
@ggetz I've added a note to the `primitiveRemoved` docs.  I've also conformed the ordering of operations between `remove()` and `removeAll()`.  It can lead to some hard to track down issues if `destroy()` is called immediately after the event handler as I found when debugging the spec.  Now in both cases `destroy()` is called before.  If the user wants to do something with the primitive when it's removed prior to being destroyed then they can handle the destroy themselves.

ggetz:
Great, thanks @chris-cooper!

Issue #11266, opened 2023-05-03
Atmosphere not rendered if globe is not defined


jjhembd:
Some tilesets (e.g. global terrain or imagery tilesets) are best rendered without a globe defined. For these datasets, the `Viewer` can be constructed without a globe:

```javascript
const viewer = new Cesium.Viewer("cesiumContainer", {
  globe: false,
});
```

However, turning off the globe also disables the atmosphere. See the [S2 Globe Sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20Next%20S2%20Globe.html). Users should be able to render an atmosphere even without the default globe.

ggetz:
See @mramato's comment about hiding the globe after startup. In that case, the atmosphere is not actually turned off.

PR #11547, opened 2023-09-26
Fix atmosphere rendering when globe is hidden

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11266

There was a bug that prevented the sky atmosphere to be shown when the globe was initially disabled, whether through setting `globe: false` in the viewer options, or by hiding the globe with `globe.show = false`. This was a quick fix in `Scene.js`.

However, it does make sense that the atmosphere would be hidden if the globe is initially set to `false`. Not every use case will be with Google's photogrammetric 3D Tiles or other 3D Tiles globes. So I did add some logic that if `globe: false` is specified in the Viewer or CesiumWidget options that the atmosphere is hidden. However it can be show by either setting `scene.skyAtmosphere.show = true` or by passing a custom `SkyAtmosphere` during initialization.

Despite the bug, this is technically breaking behavior so I could see an argument for listing a note in `CHANGES.md` under the `Breaking changes` section.


cesium-concierge:

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Thanks @ggetz. I think the change makes sense, and it works on my end. In both the [3D Tiles Next S2 Globe](https://sandcastle.cesium.com/?src=3D%20Tiles%20Next%20S2%20Globe.html) and [Google Photorealistic 3D Tiles](https://sandcastle.cesium.com/?src=Google%20Photorealistic%203D%20Tiles.html&label=All) sandcastles, a sky atmosphere is rendered if I do either of the following:
- Set `scene.globe.show = false` AFTER initializing a default `Viewer`, OR
- Initialize a `Viewer` with `globe: false`, then set `scene.skyAtmosphere.show = true`

This feels more like a bug fix to me, rather than a breaking change. The change to existing behavior would be if a user:
1. Initialized a viewer
2. Turned off the globe later.

With this PR, the atmosphere will stay on instead of disappearing. But I don't think anyone would have expected the atmosphere to turn off without a deliberate change to `scene.skyAtmosphere.show`.

ggetz:
Thanks @jjhembd! Thanks for the review. I think we're agreement here then. OK to merge once I merge in `main` then?

Issue #7398, opened 2018-12-10
Matrix4.getRotation returns rotation with scale


Vineg:
It seems to be misleading, since there is at least one place in codebase where Matrix4.getRotation is used as rotation-only. For example it leads to dim lighting of scaled point clouds, since normals are scaled as well. I think this place requires carefull review of existing usages and maybe change the name or normalize result by scale.

```javascript
var viewer = new Cesium.Viewer('cesiumContainer');
viewer.clock.currentTime = new Cesium.JulianDate(2457522.154792);

var tileset = new Cesium.Cesium3DTileset({
    url: '../../SampleData/Cesium3DTiles/PointCloud/PointCloudNormals/tileset.json'
});
viewer.scene.primitives.add(tileset);
viewer.zoomTo(tileset, new Cesium.HeadingPitchRange(0.0, -1.0, 50.0));
var style = {};
style.pointSize = Cesium.defaultValue(style.pointSize, 5.0);
tileset.style = new Cesium.Cesium3DTileStyle(style);
Cesium.Matrix4.setScale(tileset.modelMatrix, new Cesium.Cartesian3(3,3,3), tileset.modelMatrix);
```
![image](https://user-images.githubusercontent.com/1757455/49725087-f2d4a800-fc7b-11e8-88b7-dcbba4a73fe7.png)

Vineg:
this is what expected
![image](https://user-images.githubusercontent.com/1757455/49725990-054fe100-fc7e-11e8-9625-694f9f719c30.png)

hpinkos:
Thanks for opening up the issue @Vineg!  This seems like a clear cut bug.

@lilleyse can you look into this?

lilleyse:
I've been confused by this in the past too. It seems like `Matrix4.getRotation` has always just returned the upper left 3x3 matrix, which may include scale.

Should `Matrix4.getRotation` be deprecated in favor of `Matrix4.getMatrix3` or `Matrix3.fromMatrix4`?

Vineg:
`Matrix4.getMatrix3` seems nice

PR #11553, opened 2023-10-03
Fix scaling of czm_normal

jjhembd:
Normals should be transformed by the inverse transpose of the model-view matrix. See [lighthouse3d.com](https://www.lighthouse3d.com/tutorials/glsl-12-tutorial/the-normal-matrix/) and Jason McKesson's [Learning Modern 3D Graphics Programming](https://paroj.github.io/gltut/Illumination/Tut09%20Normal%20Transformation.html).

The automatic uniform `czm_normal`, as computed in `UniformState`, is currently derived from the transpose of the _rotation component_ of the inverse model-view matrix. This approach is only valid when the model-view matrix has uniform scale.

I think `UniformState` was actually correct before #8182. In that PR, we renamed `Matrix4.getRotation` as`Matrix4.getMatrix3` to [better describe what it was actually doing](https://github.com/CesiumGS/cesium/issues/7398#issuecomment-445862147). But then in `UniformState`, we mistakenly [replaced `Matrix4.getRotation` with a `Matrix4.getMatrix3` _followed by_ `Matrix3.getRotation`](https://github.com/CesiumGS/cesium/pull/8182/files#diff-7c7edcf13d6add47de5007ba19ffd488b4ae5320268c6369a75aaee51045b833R1317-R1318).

This PR removes the spurious `Matrix3.getRotation` call.

One caution: ever since #8182, `czm_normal` has been unitary. This PR will leave a scale component in the matrix, so normals will need to be re-normalized after the transform.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.111 - 2023-11-01
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Fixed `czm_normal`, `czm_normal3D`, `czm_inverseNormal`, and `czm_inverseNormal3D` for cases where the model matrix has non-uniform scale. [#11553](https://github.com/CesiumGS/cesium/pull/11553)
+
 ### 1.110 - 2023-10-02
 
 #### @cesium/engine
--- a/packages/engine/Source/Renderer/UniformState.js
+++ b/packages/engine/Source/Renderer/UniformState.js
@@ -1513,7 +1513,6 @@ function cleanNormal(uniformState) {
 
     const m = uniformState._normal;
     Matrix4.getMatrix3(uniformState.inverseModelView, m);
-    Matrix3.getRotation(m, m);
     Matrix3.transpose(m, m);
   }
 }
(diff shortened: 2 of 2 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-normal-scaling/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd:
I pushed a similar fix for `czm_normal3D`.

I'm not sure yet what is the correct approach for `czm_inverseNormal` and `czm_inverseNormal3D`.

lilleyse:
Excellent catch @jjhembd.

> I'm not sure yet what is the correct approach for `czm_inverseNormal` and `czm_inverseNormal3D`.

Shouldn't the `getRotation` calls be removed there too? Or is it more involved than that?

jjhembd:
@lilleyse I think `czm_inverseNormal`, to fit the definition, should be the transpose of the model-view matrix. This would make it the true inverse of `czm_normal`, which is the transpose of the _inverse_ model-view matrix.

We were computing it as the rotation component of the inverse model-view matrix, without any transpose. This was actually valid after #8182 for model-view matrices with uniform scale, since the rotation component is orthogonal, and for orthogonal matrices the transpose and the inverse are equal.

To honor non-uniform scale throughout, I changed `czm_inverseNormal` to be the transpose of `czm_modelView`, and `czm_inverseNormal3D` to be the transpose of `czm_modelView3D`.

lilleyse:
I ran the equations on godbolt just to confirm: https://godbolt.org/z/ar7z8dYa5. It all checks out. Thanks!

Issue #11563, opened 2023-10-10
Slope/Aspect Materials not working with globe.lightingEnabled=false


jfur1:
The expected behavior should color the globe material regardless lightingEnabled. The actual behavior is that the slope & aspect materials "disappear" when the camera gets too close to the terrain when lightingEnabled=false. For example, the elevation material seems to work regardless of if lightingEnabled is true or false. However, when lightingEnabled=false, the globe materials for slope and aspect disappear if the camera zooms in too close to the terrain. The example I am referring to is the Cesium demo for globe materials. The expected behavior is for the slope & aspect materials to keep their expected shading regardless of camera position, when the globe.lightingEnabled=false.

Sandcastle example: https://sandcastle.cesium.com/?src=Globe%20Materials.html 
- NOTE: Set globe.lightingEnabled=false and select slope or aspect material, then zoom in close to the terrain

Browser: Chrome Version 116.0.5845.188

Operating System: Windows 11 

Related Issues:
- https://community.cesium.com/t/globe-materials-not-working-w-globe-enablelighting-false/6547
- https://github.com/CesiumGS/cesium/issues/1640
- https://groups.google.com/g/cesium-dev/c/yOUjLvQ4xRY

ggetz:
Hi there,

Thanks for the report. 

I don't believe this is related to #1640. When I remove the white color stop from the slope gradient so that the last color is yellow, it instead renders yellow.

I believe this is a quick bug fix and will open a PR to fix it shortly.

PR #11567, opened 2023-10-16
Fix slope material when lighting is false

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11563

Unit testing this fix is proving difficult, so is omitted for the sake of time. Overall, globe materials appear to be fairly under-tested and any quick test I could write here would be very coupled to the shader implantation.


jjhembd (commented):
Thanks @ggetz. I'm still seeing the bug behavior on my machine: if I set `globe.enableLighting = false;`, and select the "Slope" material, it disappears when I zoom in.

ggetz:
@jjhembd I'm seeing the same issue. I'll take another look at this after today's release.

ggetz:
Superseded by https://github.com/CesiumGS/cesium/pull/11714

PR #11577, opened 2023-10-20
Fixes an issue with polygon hole rendering

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11583

This fixes an issue with polygon hole rendering that was introduced in https://github.com/CesiumGS/cesium/pull/11497/files. Holes were being projected to a slightly different tangent plane than the outer ring, which could result in slightly different coordinate systems between 2D outer positions and holes when triangulating holes with earcut.

You can see the error in the [CZML Polygon demo](https://sandcastle.cesium.com/?src=CZML%20Polygon.html). 

Also [reported on the forum with an example](https://community.cesium.com/t/cesium-v1-110-introduced-problems-with-polygons-containing-holes/27502).


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/fix-polygon-holes/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

jjhembd (approved):
Thanks @ggetz! The change makes sense, and it fixes the problematic Sandcastles on my machine.

Issue #11288, opened 2023-05-16
Keep camera from going under 3D tiles


ggetz:
Right now, the camera controller has special logic keep the camera from going under terrain, and uses a slightly different control scheme for underground camera movement. This assumes the globe is being used in the scene.

With 3D Tiles as terrain use cases, such as with [Google 3D Photorealistic Tiles](https://sandcastle.cesium.com/?src=Google%20Photorealistic%203D%20Tiles.html) and [One World Terrain](https://sandcastle.cesium.com/?src=3D%20Tiles%20Next%20S2%20Globe.html), this assumption is no longer true. 

Users want default camera controls to work well in these cases too.

Feature requested in https://community.cesium.com/t/cesium3dtileset-and-camera-collision/24007

Issue #7044, opened 2018-09-17
Add clamping functionality directly in CZML and Entity API


lilleyse:
The Clamp to 3D Tiles demo in https://github.com/AnalyticalGraphicsInc/cesium/pull/6934 manually clamps the entity's position to a tileset based on a `scene.clampToHeight` call. It would be convenient to offer this functionality from CZML and the entity API.

mramato:
I imagine the most straightforward way to do this would be to extend the HeighReference enum to cover 3D Tiles (probably always terrain and 3d tiles at the same time)

ggetz:
Also requested in https://github.com/CesiumGS/cesium/issues/11305.

ryanhlewis:
Hi from https://github.com/CesiumGS/cesium/issues/11305.

I ended up using a Raycast for everything, which looks like:

```    
    // Get position from lat/lon
    const position = Cesium.Cartesian3.fromDegrees(lng, lat, 10000); // Assume initial altitude of 10,000m

    // Cast a ray from position to the center of the Earth
    const ray = new Cesium.Ray(position, Cesium.Cartesian3.negate(Cesium.Cartesian3.fromDegrees(lng, lat), new Cesium.Cartesian3()));
    
    // Intersect the ray with the tileset
    const intersection = viewer.scene.pickFromRay(ray, [tileset]);
```

However, yes, I would love using a direct implementation, and not a hacky workaround.

ggetz:
From Sean in https://github.com/CesiumGS/cesium/issues/8606:

> I'm a little worried that scene.clampToGround won't scale super well. Terrain keeps triangle data in CPU memory, wonder if we need to do the same for 3D Tiles at some point. And then optimize ray/triangle intersection tests.

ddomit:
> > We need to get the altitude of a point from 3dTiles
> 
> @ddomit Are you looking to clamp a point to the surface of 3D Tiles, or sample the height directly?

I am looking to do both actually :)

ddomit:
> > We need to get the altitude of a point from 3dTiles
> 
> @ddomit Are you looking to clamp a point to the surface of 3D Tiles, or sample the height directly?

Any new features to implement this?

PR #11581, opened 2023-10-23
3D Tileset `getHeight`, and keeping camera above 3D Tiles

ggetz:
Fixes https://github.com/CesiumGS/cesium/issues/11288, and sets the groundwork for https://github.com/CesiumGS/cesium/issues/7044.

This PR includes 3D Tiles in camera "collision", which previously applied only to terrain, to 3D Tiles as well. It does not seek to refine or significantly alter how the camera collision itself works. Rather, this extends what is already in place for terrain to work with 3D Tiles.

It implements a `getHeight` function for 3D tilesets which copies tile data to the CPU for testing intersections, similar to the approach we take for terrain, intended for internal use such as the camera collision, and in a follow up PR clamping to 3D Tiles.

I also exposed a private `pick` function on 3D tilesets and on model primitives. This is an implementation detail of `getHeight`, but is useful for testing and internal development purposes. An example of how it works for tilesets is available in the development 3D Tiles Picking sandcastle example.

By default, since viewers default to using WebGL 2, the picking implementation will require WebGL 2 to copy buffer data at runtime.  To be compatible with WebGL 1, we take an approach similar to what we do for 2D projections of models and 3D Tiles where we allow a tileset option, `enablePick`, that keeps a copy of the loaded data on the CPU in a typed array. This is not enabled by default because it's an additional additional memory overhead.

The behavior is enabled by default for all tilesets, but the `disableCameraCollision` options for a tileset can be used to turn it off. The biggest reason for this would be a case where the user is navigating inside of a 3D tileset, as illustrated in the "3D Tiles Interior" sandcastle example.

### Testing Plan

* Try moving camera on & under terrain  to make sure terrain camera control haven't regressed
* Try moving camera on & under 3D Tileset to make sure the camera controls are consistant with terrain. Sandcastles to check in particular:
   * "Google Photorealistic 3D Tiles"
   * "3D Tiles Next S2 Globe"
* Ensure you can turn off the behavior for indoor tilesets, see the "3D Tiles Interior" Sandcastle
* Confirm nothing crashes when switching between 2D/3D/CV
* Confirm nothing crashes [when using a WebGL 1 context](http://localhost:8080/Apps/Sandcastle/index.html#c=bVFNa8JAEP0rQ+ghgmwQbzZKQUsvBQst7SWHrpsxLp3sprsTrRX/ezdZCyn1Epg372PeRlnjGfYaD+hgDgYPsESv21q89lhaJKqfl9aw1AZdkYzhVBiAiuwGZ7CV5HHcASpQ8IvXDevgOossAIefLXp+w01FkxmwayP9HL7n0W1hCkPIcLCOyhdN6JEDyO4YDYZ4uFAepObfG5VDyfhgbUX4tLNsw0zas1bT1UWSdglwaSi8QoOicbrWrPfohSzLdJjQsc+gJKsdpOicdaN4RijnLaEgW6Xv990CyMpSmwr+JsN0Bb0ZcLQU8RluTr3d+b2PSMZJ7vlIuIhbgDtdN9YxtI5SITLGuqFQzmebVn0EF+V9rAKQZ0NpXuo96HJ+5U+BIul92Gxbomf9jUWyyLPA/ye9dFnv0ZE8drTdZPEYQSFEnoXxupKtpY10A+fI+AE)

TODO:
-  [x] Debug sporadic intersection failures
- [x] Confirm 2D/Columbus view
- [x] Instanced models
- [x] Unit tests
- [X] Update CHANGES.md


jjhembd (commented):
Hi @ggetz, I just have 2 preliminary comments for now--one minor, one about the API.

Apps/Sandcastle/gallery/Google Photorealistic 3D Tiles.html
```diff
@@ -47,6 +47,7 @@
         try {
           const tileset = await Cesium.createGooglePhotorealistic3DTileset();
           viewer.scene.primitives.add(tileset);
+          viewer.scene.enableCollisionDetectionForTileset(tileset);
```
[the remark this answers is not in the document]
ggetz: Great suggestion. Making this a flag simplifies the API a lot. I'll update the PR description to reflect the changes.

ggetz:
Thanks @jjhembd! I've added a bullet to incorporate exaggeration once https://github.com/CesiumGS/cesium/pull/11655 is merged. Other than than, this is update and ready to go after that change.

ggetz:
@jjhembd In the interest of moving things along, I've omitted incorporating 3D Tiles Exaggeration in this PR. Instead, you should ensure that the `pick` function factors in the exaggeration values you've added in your PR.

@jjspace Could you please take a look and review this PR? It's already been reviewed for the overall approach, but still needs a final pass on the code and on testing.

ggetz:
@jjhembd I've added support for vertical exaggeration to this feature. Could you please re-review?

ggetz:
Hi @jjhembd, just reminder to review.

CHANGES.md
```diff
@@ -4,6 +4,17 @@
 
 #### @cesium/engine
 
+##### Breaking Changes :mega:
+
+- By default, the screen space camera controller will no longer go inside or under instaces of `Cesium3DTileset`. [#11581](https://github.com/CesiumGS/cesium/pull/11581)
+  - This behavior can be disabled by setting `Cesium3DTileset.disableCameraCollision` to true.
```
jjhembd, line 10: The option is currently `disableCollision`.
`disableCameraCollision` might be a better name. Or even `disableCameraCollisionCheck` ?

CHANGES.md
```diff
@@ -4,6 +4,17 @@
...
+  - This feature is enabled by default only for WebGL 2 and above, but can be enabled for WebGL 1 by setting the `enablePick` option to true when creating the `Cesium3DTileset`.
+
+##### Additions :tada:
+
+- Added `Cesium3DTileset.getHeight` to sample height values of the loaded tiles. If using WebGL 1, the `enablePick` option must be set to true to use this function. [#11581](https://github.com/CesiumGS/cesium/pull/11581)
+- Added `Cesium3DTileset.disableCameraCollision` to allow the camera from to go inside or below a 3D tileset, for instance, to be used with 3D Tiles interiors. [#11581](https://github.com/CesiumGS/cesium/pull/11581)
```
jjhembd, line 16: Check option naming, as in previous comment

jjhembd (commented):
Thanks @ggetz, I went through all the code and left a few minor notes.
I still need to run the tests locally and try out the Sandcastle.

packages/engine/Source/Scene/createGooglePhotorealistic3DTileset.js
```diff
@@ -46,6 +46,10 @@ async function createGooglePhotorealistic3DTileset(key, options) {
     options.maximumCacheOverflowBytes,
     1024 * 1024 * 1024
   );
+  options.enableCameraCollision = defaultValue(
```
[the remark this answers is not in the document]
ggetz: An oversight on my part. This should be removed now since it is default behavior for tilesets.

CHANGES.md
ggetz: While I agree `disableCameraCollision` (I used that first before changing it, which is why this is leftover here) is a clearer name for this particular function, I used `disableCollision` as this is also loads the appropriate data needed for clamping entities to 3D Tiles in https://github.com/CesiumGS/cesium/pull/11604.

jjhembd:
As a follow-up on the remaining items from my review:
- Specs run successfully on my machine (including the new `pickModelSpec`)
- New picking Sandcastle works (mostly) as expected

I noticed one strange behavior when picking on the "Bentley BIM" model. Some parts of the tileset are picked by the `Scene.pickPosition` call, but _not_ by the `tileset.pick` call. See the below screenshot, where I enlarged the red circles from the first call to make the difference more clear. (Here is the [modified local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=zVZtb9s2EP4rhD/JgEDHaTZgrhOssZ1NmJMYsdagqIeCps42N4oUSMqJOvi/76gXW03cbivQLYAhk/f66DnyTlwr68hWwAMYck4UPJARWJGn9G0pCxYdXu5HWjkmFJhFJyR/LhQhTqQgUTIgKyYthF7GlEiZE1q1hUtmYcoKMDPB/wDTVq2lXrYC7LqvF4qXkCwHBYiogkbLLSoXSoLD1BIsuL2xznxOi+bvfdQSHgKERzcgi85PWq8lkNmrcYzgKx1agwSOemYLxUnQJecXjSOpstPMiFQ4sQVLDaR6C0Gd2MOsDJ0pDl6kAYZA2AMTruGSG2AOKhyzjXYa91JYJzhiqlyCQ8wj6VmSPM+9I5w5viEBGKNNt43D06IlUKnXtfbgVS12JRHV8wlf1+yRGXJ7H5P7qyvSp6f/E2vV354hujI6jbR6Y3ETJcH3P/S/65+E7UCEpOxRpHk6R8ZBzTPGYeJff0DOwoPZ7uVwfQnKSSjIZXRNrnUC8mVy3T89Ozk7Of33vJHmAn/UOo31f0dshH2BKQ7JCyL0VyOD9mFddCjt4W/O0kzCmDnW+8Stt3+Jw+peuM2lZyxmSwm9Ojv93Wq16ByCv9xK/VZ28TlTCWcWz75HM4YVy6WLtZZLZq5B5UHd0ruldTMSjAcwYsYhS0x9Oq/24lfBYYpsMI18OtnazWGL9+/nyiioeOJMbVmZt3am/gqoLHdvuIcUrHJVLkjgz0eKAWouqpSZH3LJTFvhje7A4oth/roIqGxUe3ea1ZKKNbEiQY00gRUO2CQ4FrO7r0BdN4yE6rq8++I0sQdHgYUHM6HwZrRqmolHkHPxEcfz6UnYrrX0/bRh3e/o3WTcskiE9WdzDJnbxGDdWFRnd0Bu8nSJSGe38yiO3k4+RDdX0U0Uv9s77+pV3aF3vvgNsYYVex45S8EwugbnvynuWPE5Mts1Qe/munhBgAHDOt7KYMC5wyEdPjtm1RH8Ql2+shL/jPv+33H/bjKd3t5/K/p34eduTVxkQKeTq/jDaBqNfum+Jp2wM7SukHDRxPtRpJk2juTY+LDPOcA+hxzb3jJHBhzl1jadYthruw4TsSUiOT/y9Um4ZNaiZpXLkqNF52LYQ/tnrlKzRKj17RaMZIU32/QvppWQUjrs4fa4p6sa0ZPIfwE))

image

After I move the camera slightly, both pick calls work well on the problematic feature. However, after moving the camera, some of the reported properties are missing--I no longer see the "assembly" and "category" properties in the pop-up.

ggetz:
> I noticed one strange behavior when picking on the "Bentley BIM" model. Some parts of the tileset are picked by the Scene.pickPosition call, but not by the tileset.pick call. See the below screenshot, where I enlarged the red circles from the first call to make the difference more clear.

Interesting. I can reproduce at a very specific camera angle and zoom level (that can be difficult to replicate). From what I can tell, these are instanced parts of the model. However, the instanced node are quite similar to what is being tested in the instanced example. 

I also cannot find another tileset with similar issues. 

Finally, it is not returning an incorrect position, but rather `undefined`, meaning it should not interfere with camera collisions.

> However, after moving the camera, some of the reported properties are missing--I no longer see the "assembly" and "category" properties in the pop-up.

I don't think this is altered behavior in this PR. I can replicate in the last official release. 

I'm thinking this may be unique to the Bentley powerplant tileset. @jjhembd Let me know if you agree and if this is is a good place.

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -110,7 +110,8 @@ import Ray from "../Core/Ray.js";
...
  * @property {string|number} [instanceFeatureIdLabel="instanceFeatureId_0"] Label of the instance feature ID set used for picking and styling. If instanceFeatureIdLabel is set to an integer N, it is converted to the string "instanceFeatureId_N" automatically. If both per-primitive and per-instance feature IDs are present, the instance feature IDs take priority.
  * @property {boolean} [showCreditsOnScreen=false] Whether to display the credits of this tileset on screen.
- * @property {boolean} [disableCollision=false] When {@link ScreenSpaceCameraController#enableCollisionDetection} is true, allows the camera to go in or below the tileset surface.
+ * @property {SplitDirection} [splitDirection=SplitDirection.NONE] The {@link SplitDirection} split to apply to this tileset.
+ * @property {boolean} [disableCollision=false] Whether to turn off collisions for camera collisions or picking. While this is `false` the camera will be allowed to go in or below the tileset surface if {@link ScreenSpaceCameraController#enableCollisionDetection} is true.
```
jjhembd, line 114: I think this should read "While this is `true` the camera will be allowed..." -- unless I'm not understanding the option?

jjhembd:
> I'm thinking this may be unique to the Bentley powerplant tileset. @jjhembd Let me know if you agree and if this is is a good place.

Yes, I agree those minor quirks don't look like they are related to this PR. I wonder if there might be an existing issue in `camera.getPickRay` that affects close zooms.

I have one question about a doc, and there are a couple failing specs. But other than that, this should be ready to go!

ggetz:
Thanks @jjhembd! This should be ready.

jjhembd (approved):
Looks good, thanks @ggetz !

Issue #11564, opened 2023-10-10
Implicit tile not refining correctly 


ptrgags:
My team was working with a 3D Tiles 1.1 tileset with implicit tiling. Nothing particularly unusual about this one, it's an octree with a box bounding volume. The GE value at the root tile seems reasonable for most of the tileset except this one.

Unfortunately I can't share the data for this one, only a couple select screenshots to show the issue:

When I zoom in on this building, it's still blurry and the tile is huge. However, if you look to the right side, the neighbor tile (same size as this one) has refined once or twice.
![image](https://github.com/CesiumGS/cesium/assets/8422414/7f1332d5-ffc5-4f57-8475-f0200fe9930f)

Furthermore, if I fly to another side of the building and zoom in very close, it does finally refine, but zooming in further it goes back to blurry again:

| Blurry | Better! | Blurry again?? |
|---|---|---|
|![image](https://github.com/CesiumGS/cesium/assets/8422414/11845fc3-4024-4e19-b13f-9ad0af3cc476)|![image](https://github.com/CesiumGS/cesium/assets/8422414/76a8169f-e709-4f40-b329-73d2c0e6c347)|![image](https://github.com/CesiumGS/cesium/assets/8422414/796eec79-f0be-46ea-83df-7a788ad486b2)|

Also note that this does not happen in Cesium for Unreal, the tile refines just fine
![image](https://github.com/CesiumGS/cesium/assets/8422414/f4a50a9d-57bd-432f-aaf0-37fa2280bf99)

This effect seems to be sensitive to camera angle, the problem shows up more/less depending on which direction the camera is facing. 

As for things to look into:

- In the debugger, check the SSE computation to see if anything is unusual. To filter for the tile, I would use the implicit tile ID (TODO: determine which tile it is. I think it's one of the level 1 tiles).
- see if this is related to https://github.com/CesiumGS/cesium/issues/11174

ptrgags:
I'm starting to look into this. I started by looking at the tiles that are loaded.

![image](https://github.com/CesiumGS/cesium/assets/8422414/01ef3d99-8a8e-42a7-85f5-a909720134d3)

Observations:
- The first tile with the problem is `(level; x, y, z) = (1; 0, 0, 0)` which fortunately is easy to remember
[...]

PR #11611, opened 2023-11-06
Fix empty traversal for mixed empty and non-empty tiles

### LINKED ISSUE #9356: Refinement of mixed empty and non-empty tiles
Hello,

I'm building a level-of-detail structure of triangle meshes (Cesium 1.77) where some nodes in the tree can be empty (i.e. no "content" field, only geometric error, bounding volume and children are set).

Let's assume my tree has 7 levels, where level 7 is the coarsest and level 0 the most detailed representation. Each non-empty tile is stored as a single b3dm file. My question is regarding the refinement of a node on level 7 that has one empty and one non-empty child:
 
```
level 7:         non-empty          GE: 200
                 /      \
level 6:     empty     non-empty    GE: 100
level 5:     empty        .         GE:  50
level 4:     empty        .         GE:  25
level 3:     empty        .         GE:  12
level 2:   non-empty      .         GE:   6
level 1:   non-empty      .         GE:   3
level 0:   non-empty   non-empty    GE:   1
```

(The geometric error "GE" decreases when going down the tree and is the same for all nodes on a given level. The refinement strategy is "REPLACE".)

When level 7 is to be refined, I see that the non-empty child on level 6 gets fetched (its b3dm file) by the browser, while the empty child is not. For this reason, the level 7 tile is never refined.

Adding `"content" : ""` to the empty level 6 node and all following empty nodes, causes level 7 to be refined. However, the non-empty level 2 tile is shown instead of the empty level 6 tile essentially showing two very different geometry resolutions at the same time.
 
Instead, I would expect that the level 7 tile is replaced by its two immediate children on level 6 (one empty and one non-empty) and only when I keep zooming in, the non-empty level 2 tile would eventually be displayed.

What is the intended behavior in this case? Regarding the "content" field, I think there could be a bug because it is not required as per specification but leaving it out causes level 7 to never be refined.

For reference, I've found this report from a while ago which seems to describe a similar setup: https://github.com/CesiumGS/cesium/issues/7084#issuecomment-426891683

Thankful for any help or pointers!

ptrgags:
Fixes #9356

This PR fixes a bug in empty tile traversal code where sometimes a parent tile wouldn't refine if one of its children is a long chain of empty tiles followed by a tile with content. The old logic would assume the parent shouldn't refine because there was a renderable descendant, and refining would create a hole. However, some cases (like in #9356 and #11564) actually require the hole to render properly.

I went through and redid the logic of that function, and was able to simplify and document it so it's easier to follow.

The biggest difference is that now only tiles that would be rendered for the current camera view are examined, before it would traverse deeper in the tree if there are a lot of empty tiles. 

@jjhembd could you review?

CC @lilleyse

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Change Log
 
+### 1.112 - 2023-12-01
+
+##### Fixes :wrench:
+
+- Fixed a bug where 3D Tiles replacement refinement would sometimes not refine correctly for a tile with both empty and non-empty children.[#11611](https://github.com/CesiumGS/cesium/pull/11611)
+
 ### 1.111 - 2023-11-01
 
 #### @cesium/engine
--- a/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js
+++ b/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js
@@ -239,6 +239,24 @@ function executeTraversal(root, frameState) {
   }
 }
 
+/**
+ * Depth-first search starting at an empty tile to hint to help determine if
+ * this tile's parent should refine when using REPLACE refinement.
+ *
+ * The only reason why refinement should not happen is if there is a descendant
+ * tile that would be visible for the current camera view but has not finished
+ * loading.
+ *
+ * The traversal progresses from the root until the first tiles that go below
+ * the screen space error threshold. This is important to prevent long chains
+ * of empty tiles with renderable content beneath from holding up the
+ * refinement.
+ *
+ * @param {Cesium3DTile} root An empty tile that marks the root of the traversal
+ * @param {FrameState} frameState The frame state
+ * @returns {boolean} false if the parent tile should not refine yet (due to tiles still loading), true otherwise.
+ * @private
+ */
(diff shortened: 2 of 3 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/9356-fix-empty-traversal/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ptrgags:
I'm going through the Sandcastles in the gallery for further testing. I'm noticing some issues:

1. for Google Photorealistic 3D Tiles, when I move the camera, the view gets blurry when waiting for tiles... so something's not quite right
2. For 3D Tiles Next Photogrammetry Classification, it takes quite a bit longer to load and hits the memory limit, that doesn't happen on `main`

ptrgags:
I know the approach in the PR as it stands now is not the correct approach, I get tiles popping in and out with Google Photorealistic 3D Tiles.

It might be worth re-trying @lilleyse's original idea of just returning `true`. I think there might be some subtleties there indicated by unit tests when I tried it the first time. 

For now, I'm closing this PR.

PR #11648, opened 2023-11-22
Fix race condition in OctahedralProjectedCubeMapSpec.js

mramato:
`raises error event when environment map fails to load.` assumes that the errorEvent listener will always be fired before `projection.update` returns true, but inspecting the code shows that is clearly not the case. It's possible for an error to be thrown, the update to return true, another tick to run and then the error event is raised. This lead to a Jasmine `afterAll` failure pointing to line 222 of this file on my machine when running the tests.

Wrapping the event in a promise and waiting for both `projection.update` to return true and the event to be handled is the way to ensure no race conditions in the spec itself.

### DIFF
--- a/packages/engine/Specs/Scene/OctahedralProjectedCubeMapSpec.js
+++ b/packages/engine/Specs/Scene/OctahedralProjectedCubeMapSpec.js
@@ -208,7 +208,7 @@ describe(
         });
     });
 
-    it("raises error event when environment map fails to load.", function () {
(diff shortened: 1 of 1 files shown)


cesium-concierge:
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/octa-what-now/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

ggetz:
Thanks for the fix @mramato!

PR #11662, opened 2023-12-01
Dependency updates

jjhembd:
Most of these are minor version changes.
One bigger change was `gulp-zip`, which is now ESM only. I verified that the `make-zip` script still runs through succesfully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,8 +51,7 @@
   ],
   "dependencies": {
     "@cesium/engine": "^6.1.0",
-    "@cesium/widgets": "^4.3.0",
-    "meshoptimizer": "^0.20.0"
+    "@cesium/widgets": "^4.3.0"
   },
   "devDependencies": {
     "@playwright/test": "^1.40.1",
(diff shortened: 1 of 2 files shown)


cesium-concierge:
    ```
    ```
* :grey_question: [CHANGES.md](https://github.com/CesiumGS/cesium/blob/dependency-updates/CHANGES.md) was not updated.

- [ ] Cesium Viewer works.
- [ ] Works in 2D/CV.

package.json
```diff
@@ -51,29 +51,30 @@
   ],
   "dependencies": {
     "@cesium/engine": "^6.1.0",
-    "@cesium/widgets": "^4.3.0"
+    "@cesium/widgets": "^4.3.0",
+    "meshoptimizer": "^0.20.0"
```
ggetz, line 55: I don't think this needs to be in the top-level `package.json`, only in `packages/engine/package.json`.

jjhembd: Thanks for catching that. This should be good to go now

PR #11663, opened 2023-12-01
Our package only has major.minor version naming.

mramato:


jjhembd (commented):
We do occasionally have a patch release. For example, next week we might have 1.112.1. Wouldn't this `cut` script republish the '.0' zip?

mramato:
@jjhembd Great point, when I tested this I think it was against a `.1` version so the zip worked locally. I used the `sed` suggestion you made offline, which looks cleaner and would only ever be a problem if we got to `.10`, which means some other crazy stuff happened :smile:

jjhembd:
Thanks, your comment made me worry about `1.200`... so I escaped the `.` in the sed pattern to be safe

PR #11664, opened 2023-12-02
Fix cesium.com branch for real I swear

mramato:
We don't actually build the zip in cesium.com deploy, we fetch it from GH releases. That's the root cause of why unzipping was failing.


jjhembd (approved):
I verified the curl commands on my machine. Also this seemed to work on our last deployment, so I think it's good to go.
Thanks again @mramato!

PR #11669, opened 2023-12-04
Correct typing for ImageryProvider constructor

### LINKED ISSUE #11625: Typing for ImageryLayer contructor is incorrect
JSDoc indicates both the `provider` and  `options` arguments are required. 

https://github.com/CesiumGS/cesium/blob/003922bb52665f09c7d063a57b6ee585ec58f3c3/packages/engine/Source/Scene/ImageryLayer.js#L135-L136

However they are both defaulted to the empty object in the constructor.  Though `imageryProvider` has already been assigned to `this._imageryProvider` at this points so not sure if that is a bug or this should truly be an required argument.

https://github.com/CesiumGS/cesium/blob/003922bb52665f09c7d063a57b6ee585ec58f3c3/packages/engine/Source/Scene/ImageryLayer.js#L165-L166

This causes errors with Typescript when the options argument is not provided

image

angrycat9000:
Mark the `ImageryProvider` constructor parameters as optional.  Fixes #11625 

## Test plan

1. Open up `packages/engine/index.d.ts` on main.  Find `ImageryProvider` and verify that both parameters are required for the constructor.
2. Open this branch and run `npm run build-ts`.  Find `ImageryProvider` and verify that both are now optional.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.113
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Corrected JSDoc and Typescript definitions that marked optional arguments as required in `ImageryProvider` constructor [#11625](https://github.com/CesiumGS/cesium/issues/11625)
+
 ### 1.112 - 2023-12-01
 
 #### @cesium/engine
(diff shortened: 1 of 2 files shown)


angrycat9000:
Found this one again in my notes and realized I never opened up the PR for it

ggetz:
Thanks @angrycat9000!

PR #9634, opened 2021-06-23
Added I3S data source support in Cesium

Tamrat-B:
Esri Contribution: Adds support for I3S 3D Object and IntegratedMesh Layers in Cesium.
Co-authored-by: Alexandre Jean-Claude <ajeanclaude@spiria.com>
Co-authored-by: Elizabeth Rudkin <elizabeth.rudkin@presagis.com>
Co-authored-by: Anthony Mirabeau <anthony.mirabeau@presagis.com>
Co-authored-by: Tamrat Belayneh <tbelayneh@esri.com>

PR #11678, opened 2023-12-06
Add support for Building Scene Layer (BSL) OGC I3S standard

Tamrat-B:
Esri Contribution: Adds support for OGC I3S Building Scene Layer (BSL) in CesiumJS.
Co-authored-by: Anton Smirnov [anton.smirnov@actionengine.com](mailto:anton.smirnov@actionengine.com)
Co-authored-by: Maxim Kuznetsov [maxim.kuznetsov@actionengine.com](maxim.kuznetsov@actionengine.com)
Co-authored-by: Tamrat Belayneh [tbelayneh@esri.com](mailto:tbelayneh@esri.com)

This PR adds a new functionality to support an [OGC Indexed 3D Scene Layer (I3S) 1.3](https://www.ogc.org/standard/i3s/) : Building Scene Layer (BSL) type. The PR also adds a new I3S Building Scene Layer – Cesium Sandcastle application showcasing BSL consumption of a BIM data in CesiumJS. The PR is largely based on extending the original contribution to PR [#9634](https://github.com/CesiumGS/cesium/pull/9634) supporting I3S 3D Object and IntegratedMesh Scene Layers.


Tamrat-B:
PR Description

This PR adds a new functionality to support an [OGC Indexed 3D Scene Layer (I3S) 1.3](https://www.ogc.org/standard/i3s/) : Building Scene Layer (BSL) type. The I3S Building Scene Layer – Cesium Sandcastle application showcases BSL consumption of BIM data for [Turanga Library in Christchurch](https://www.arcgis.com/home/item.html?id=543648a92446497db8a92c06ce1ad0b1) (NZ) in CesiumJS. The PR is largely based on extending the original contribution to https://github.com/CesiumGS/cesium/pull/9634. 

 
image

Find here a [running](https://feature-bsl-new.d1pv3t0ca2deuy.amplifyapp.com/Apps/Sandcastle/index.html?src=I3S%20Building%20Scene%20Layer.html) version of I3S Building Scene Layer - Cesium SandCastle show case.

Below are overviews detailing changes brought about by this PR on both UI and Engine level.

### UI Changes Overview

- Application creates I3S data provider and adds it as a primitive data type to Cesium Viewer
- I3S BSL Explorer UI is implemented as Cesium widget and is located @ /packages/widgets/Source/I3SBSLExplorer
- I3S BSL Explorer widget implementation follows MVVM pattern and uses knockout js:
- Model is created as I3S data provider in the application
- View is represented by js file I3SBSLExplorer.js, defines html elements and data binding between the elements and ViewModel
- ViewModel is represented by I3SBSLExplorerViewModel.js, creates ViewModel based on I3S data provider and handles the following user events:
- selecting top layer for exploring
- checking/unchecking layers and sublayers visibility
- expanding/collapsing BSL tree view
- To display I3S Explorer widget the showcase application passes two arguments to Cesium.I3SBSLExplorer – id of the container html element where the widget is planned to be displayed and I3S data provider

#### List of files added/modified:

Apps/Sandcastle/gallery/
&nbsp; &nbsp; I3S Building Scene Layer.html 
&nbsp; &nbsp; I3S Building Scene Layer.jpg 

packages/widgets/Source/
&nbsp; &nbsp; I3SBSLExplorer/ 
&nbsp; &nbsp; I3SBSLExplorer.css
&nbsp; &nbsp; I3SBSLExplorer.js
&nbsp; &nbsp; I3SBSLExplorerViewModel.js
&nbsp; &nbsp; widgets.css – added import url of the I3SBSLExplorer widget styles (I3SBSLExplorer.css) 

### Engine Changes Overview
#### High-level structure of changes:

Scene layer initialization became dependent on the layer type ("Building", "3DObject" and "IntegratedMesh" layer types are supported). [Common 3D Scene Layer](https://github.com/Esri/i3s-spec/blob/master/docs/1.8/3DSceneLayer.cmn.md) initialization is not changed.I3S Layer class still represents a single user-visible layer: 3D Object or Integrated Mesh.

When a Building Scene Layer is initialized, it accomplishes the following:
* It processes the [Building Scene Layer resource] (https://github.com/Esri/i3s-spec/blob/master/docs/1.8/layer.bld.md) of an I3S dataset, by creating the hierarchy of sublayers and loading statistics summary.
* Each I3S Sublayer represents either a group of sublayers or a user-visible layer (I3S Layer), a 3D Scene Layer resource is requested for each I3S Layer.
* I3S Data Provider stores the collection of all I3S layers in the scene (to control the life cycle of tilesets) as well as the collection of I3S sublayers (to manage sublayers and layers visibility).

#### Transcoding changes:

* If requested, when creating an I3SDataProvider the user has now a new option _enableFeatureFiltering_ to enable displaying geometries by I3S features against the default displaying by I3S nodes. By default, _enableFeatureFiltering_ is disabled for Common 3D Scene Layer, but enabled for Building Scene Layer. When the enableFeatureFiltering option is enabled, we add EXT_mesh_features and EXT_structural_metadata extensions into a glTF document to force CesiumJS to create Cesium 3D Tile Feature for each feature in a node.
* A new option, _adjustMaterialAlphaMode_ is added to enable features transparency by adjusting the alpha mode of their material definition. By default, it is disabled for Common 3D Scene Layer, but enabled for Building Scene Layer. When the adjustMaterialAlphaMode option is enabled, we separate each geometry into two mesh primitives depending on whether the vertex color has transparency or not. Then the alpha mode in the material definition is updated for each mesh primitive based on its transparency.
* A new option, _applySymbology_ is added to enable displaying outlines if they are defined by the I3S layers' symbology. By default, it is disabled for Common 3D Scene Layer, but enabled for Building Scene Layer. When _applySymbology_ option is enabled, we add CESIUM_primitive_outline extension into a glTF document to display outlines for 3D objects in the layers.
* A new option, _calculateNormals_, is added to enable normals calculation if they are missing. By default, it is disabled for Common 3D Scene Layer with textures (if textureSetDefinitions details are present), but enabled for Common 3D Scene Layer without textures and Building Scene Layer.When the calculateNormals option is enabled, we generate the normals buffer with the flat normals per face for each geometry without the vertex normal attribute. Flat normals are calculated based on the vertex position attribute. The vertex normal attribute may be stripped out from the geometry buffer to reduce the size of the buffer. This is allowed by glTF specification where it states: "When normals are not specified, client implementations MUST calculate flat normals". However, the current CesiumJS implementation doesn’t [automatically calculate flat normals for glTF 2.0](https://github.com/CesiumGS/cesium/issues/6506) yet , which leads to unlit visualization for 3D objects. 


#### About color saturation:
* [I3S material colors properties](https://github.com/Esri/i3s-spec/blob/master/docs/1.8/materialDefinitions.cmn.md) (baseColorFactor, emissiveFactor etc.) are assumed to be in the same color space as the textures, most commonly sRGB while in glTF they are interpreted as linear. The color vertex attribute is also assumed to be [Standard RGB](https://github.com/Esri/i3s-spec/blob/master/docs/1.8/geometryColor.cmn.md) (sRGB space).We convert colors from sRGB color space to linear color space during transcoding I3S to glTF.

#### About feature attribute driven filter:

* When the _enableFeatureFiltering_ option is enabled (enabled by default for BSL) and any I3S node geometry contains features, there is now a new functionality to apply a filter driven by feature attribute values. A filter may contain one or multiple attributes and each attribute may specify one or multiple values. When a filter is applied, a 3D feature object will be hidden if its value for at least one attribute in the filter doesn't match to any attribute value of the filter. A filter can be applied either to specific I3S Layer or to all I3S layers in the scene. 
* When a filter is applied, it accomplishes the following: I3S Layer stores the copy of the filter attributes and requests the root node and all children of the root node to apply the filter. If any node is not visible or even not loaded yet, then the filter is applied once the content model is created and becomes ready. We then load an I3S Field for each attribute in the filter and set the visibility for each Cesium 3D Tile Feature based on the filter and the feature attribute values. We hide the whole model for I3S Node content while filtering is in progress to apply the filter for all features at once.

#### About transparency:
* The color vertex attribute of the geometry buffer may contain transparent colors for any vertex. The interpretation of these alpha values depends on the alphaMode property of the material definition. The alphaMode value should be set to "BLEND" to render transparent objects. On the other hand, the blending mode for the fully opaque colors (100% alpha) also creates the effect of translucency. Processing the transparency in I3S layers is optional and can be configured by the _adjustMaterialAlphaMode_ option (for BSL it is enabled by default). When the _adjustMaterialAlphaMode_ option is enabled, the blending alpha mode is set for each geometry with the transparency in the color attribute.

#### About symbology:
* Each I3S layer may define symbology details for rendering. Generally, it may specify the existence and the color for outlines (edges) and the material color. Symbology may be defined for all data in the layer or for particular 3D features, based on attribute values (unique or ranges).
* When the applySymbology option is enabled, we parse data for the symbology during layers loading. As the result of parsing, we may have a list with specific symbology per feature and the default symbology for all other features. The outlines color from the default symbology is set as the default value for outlineColor option for 3D Tileset. Then we use additional processing during transcoding I3S to glTF.

#### List of files added/modified :
packages/engine/Source/
&nbsp; &nbsp; Core/
&nbsp; &nbsp; &nbsp; srgbToLinear.js
&nbsp; &nbsp; &nbsp; - Added support for conversion from sRGB color space profile to linear color space profile
&nbsp; &nbsp; &nbsp; Purpose: Helper function to fix color saturation
&nbsp; &nbsp; Scene/
&nbsp; &nbsp; &nbsp; I3SDataProvider.js
&nbsp; &nbsp; &nbsp; - Added support for BSL structure per the OGC I3S standard
&nbsp; &nbsp; &nbsp; Purpose: To display 3D BIM models and access the BSL sublayers hierarchy
&nbsp; &nbsp; &nbsp; - Added support for BSL statistics
&nbsp; &nbsp; &nbsp; Purpose: To provide the field names and values for BSL which may be used for the attribute driven filter UI
&nbsp; &nbsp; &nbsp; - Added support for the generic feature attribute driven filter
&nbsp; &nbsp; &nbsp; Purpose: To filter the drawn elements of all scene layers
&nbsp; &nbsp; &nbsp; - Added new initialization option: enableFeatureFiltering
&nbsp; &nbsp; &nbsp; Purpose: To provide an ability for common 3D Scene Layer to use the attribute driven filter
&nbsp; &nbsp; &nbsp; - Added new initialization option: adjustMaterialAlphaMode
&nbsp; &nbsp; &nbsp; Purpose: To set the 3D objects transparency based on the vertices color data
&nbsp; &nbsp; &nbsp; - Added new initialization option: applySymbology
&nbsp; &nbsp; &nbsp; Purpose: To define and show outlines for 3D Scene Layer based on symbology details in I3S Layer data
&nbsp; &nbsp; &nbsp; - Added new initialization option: calculateNormals
&nbsp; &nbsp; &nbsp; Purpose: To generate the buffer with flat normals per face if normals are stripped out from I3S geometry.
&nbsp; &nbsp; &nbsp; - Added Support for BSL structure, BSL statistics, generic feature attribute driven filter.
&nbsp; &nbsp; &nbsp; - Added new initialization option: enableFeatureFiltering, adjustMaterialAlphaMode, applySymbology and calculateNormals

&nbsp; &nbsp; I3SField.js
&nbsp; &nbsp; &nbsp; - Added support for binary attribute data with values stored in objectIds property
&nbsp; &nbsp; &nbsp; Purpose: Some attribute data for Turanga_Library BSL stores values in objectIds property (not in attributeValues) which caused runtime errors
&nbsp; &nbsp; &nbsp; - Added validation for actual and expected binary buffer size
&nbsp; &nbsp; &nbsp; Purpose: Some attribute data for Turanga_Library BSL looks corrupted, the size of the binary buffer doesn't correspond to attributeStorageInfo which caused runtime errors
&nbsp; &nbsp; &nbsp; - Added support for binary attribute data with values stored in objectIds property
&nbsp; &nbsp; &nbsp; - Added validation for actual and expected binary buffer size


&nbsp; &nbsp; I3SGeometry.js
&nbsp; &nbsp; &nbsp; - Added support for extensions and extensionsUsed properties of the glTF root object
&nbsp; &nbsp; &nbsp; Purpose: To add information about features into a glTF asset
&nbsp; &nbsp; &nbsp; - Added support for 3D objects transparency
&nbsp; &nbsp; &nbsp; - Fixed material color saturation by converting sRGB colors in I3S data to linear color space profile
I3SLayer.js
&nbsp; &nbsp; &nbsp; - Added support for BSL hierarchy of layers: Scene -> Sublayer -> ... -> Layer versus common 3D SL hierarchy: Scene -> Layer
&nbsp; &nbsp; &nbsp; Purpose: To adjust the URL to the 3DSceneLayerResource and manage the layers visibility
&nbsp; &nbsp; &nbsp; - Added support for the generic feature attribute driven filter
&nbsp; &nbsp; &nbsp; Purpose: To filter the drawn elements of a layer
&nbsp; &nbsp; &nbsp; - Added support for I3S symbology
&nbsp; &nbsp; &nbsp; Purpose: To display the outlines for 3D objects
&nbsp; &nbsp; &nbsp; - Added support for BSL hierarchy of layers
&nbsp; &nbsp; &nbsp; - Added support for the generic feature attribute driven filter
&nbsp; &nbsp; &nbsp; - Added support for I3S symbology

&nbsp; &nbsp; I3SNode.js
&nbsp; &nbsp; &nbsp; - Added support for the generic feature attribute driven filter
&nbsp; &nbsp; &nbsp; Purpose: To filter the drawn features of a node
&nbsp; &nbsp; &nbsp; - Added support for I3S symbology
&nbsp; &nbsp; &nbsp; Purpose: To display the outlines for 3D objects
&nbsp; &nbsp; &nbsp; - Added support for the generic feature attribute driven filter
&nbsp; &nbsp; &nbsp; - Added support for I3S symbology

&nbsp; &nbsp; I3SStatistics.js
&nbsp; &nbsp; &nbsp; - Added support for loading and parsing BSL statistics
&nbsp; &nbsp; &nbsp; Purpose: To provide the field names and most frequent values for BSL
&nbsp; &nbsp; I3SSublayer.js
&nbsp; &nbsp; &nbsp; - Added support for BSL structure per the OGC I3S standard
&nbsp; &nbsp; &nbsp; Purpose: To provide the BSL sublayers hierarchy and manage the layers visibility

&nbsp; &nbsp; I3SSymbology.js
&nbsp; &nbsp; &nbsp; - Added support for symbolization defined in I3S Layer data
&nbsp; &nbsp; &nbsp; Purpose: To provide the outlines and material details for the Layer

&nbsp; &nbsp; Workers/
&nbsp; &nbsp; &nbsp; decodeI3S.js
&nbsp; &nbsp; &nbsp; - Added support for EXT_mesh_features and EXT_structural_metadata glTF extensions with the feature index data
&nbsp; &nbsp; &nbsp; Purpose: To add information about features into a glTF asset
&nbsp; &nbsp; &nbsp; - Added support for 3D objects transparency
&nbsp; &nbsp; &nbsp; - Fixed vertex color saturation by converting sRGB colors in I3S geometry to linear color space profile
&nbsp; &nbsp; &nbsp; - Added support  for CESIUM_primitive_outline glTF extensions with the generated outlines indices
&nbsp; &nbsp; &nbsp;  Purpose: To display the outlines for 3D objects
&nbsp; &nbsp; &nbsp; - Added support for flat normals calculation if vertex normals are stripped out from I3S geometry
&nbsp; &nbsp; &nbsp;  Purpose: to enable lighting effects for I3S datasets without normals

pjcozzi:
@Tamrat-B 👏 🙏 

Really appreciate the work by you and the team.

We'll chat with @lilleyse and @ggetz to find some cycles for a code review.

ggetz:
Thanks @Tamrat-B! Exciting! We'll take a review pass of this before the end of the week.

Tamrat-B:
Thanks @ggetz. Looking forward to it.

packages/engine/Source/Scene/I3SDataProvider.js
```diff
@@ -631,18 +803,61 @@ I3SDataProvider.prototype._computeExtent = function () {
...
+  }
+  return attributes;
+};
+
+/**
+ * Returns the collection of values for the attribute with the given name
+ * @param {string} name The attribute name
+ * @returns {string[]} The collection of attribute values
+ */
+I3SDataProvider.prototype.getAttributeValues = function (name) {
```
ggetz, line 839: Since `name` is required, it should be validated with the `Check` function.

packages/engine/Source/Scene/I3SField.js
```diff
@@ -92,40 +100,50 @@ function getNumericTypeSize(type) {
...
+    return 1;
+  }
+  return getNumericTypeSize(type);
+}
+
+async function load(field) {
+  const data = await field._dataProvider._loadBinary(field._resource);
+
+  // Check if we have a 404
+  const dataView = new DataView(data);
```
ggetz, line 114: If the request 404's the `loadBinary` function should throw an exception instead of returning the binary.

Then, you should be able to use a `try/catch` block here instead of attempting to parse the results directly.

packages/engine/Source/Scene/I3SField.js
```diff
@@ -252,4 +272,109 @@ I3SField.prototype._parseBody = function (dataView, offset) {
...
+  for (
+    let itemIndex = 0;
+    itemIndex < this._storageInfo.header.length;
+    itemIndex++
+  ) {
+    const item = this._storageInfo.header[itemIndex];
+    headerSize += getValueTypeSize(item.valueType);
+  }
+  if (dataView.byteLength < headerSize) {
+    console.error("Invalid attribute buffer size");
```
ggetz, line 306: This case should throw an exception rather than logging to the console.

ggetz: And instead of returning `true` or `false`, when this function is invoked, tit should be surrounded by a `try/catch` block.

packages/engine/Source/Scene/I3SField.js
```diff
@@ -252,4 +272,109 @@ I3SField.prototype._parseBody = function (dataView, offset) {
...
+  }
+  return true;
+};
+
+/**
+ * @private
+ */
+I3SField.prototype._validateBody = function (dataView, offset) {
+  if (!defined(this._header.count)) {
+    console.error("Invalid attribute buffer");
```
ggetz, line 320: Same comments as above.

packages/engine/Source/Scene/I3SField.js
```diff
@@ -252,4 +272,109 @@ I3SField.prototype._parseBody = function (dataView, offset) {
...
+        }
+        const valueSize = getNumericTypeSize(desc.valueType);
+        offset += valueSize * this._header.count;
+      } else {
+        if (!defined(attributeByteCountsOffset)) {
+          console.error("Invalid attribute buffer");
+          console.error(
+            ` (field: ${this.name}, attributeByteCounts is missing)`
+          );
+          return false;
```
ggetz, line 348: Same comments as above

packages/engine/Source/Scene/I3SField.js
```diff
@@ -252,4 +272,109 @@ I3SField.prototype._parseBody = function (dataView, offset) {
...
+            attributeByteCountsOffset
+          );
+          offset += parsedValue.value;
+          attributeByteCountsOffset = parsedValue.offset;
+        }
+      }
+    } else {
+      console.error("Invalid attribute buffer");
+      console.error(` (field: ${this.name}, ${item} is missing)`);
+      return false;
```
ggetz, line 367: Same comment as above

packages/engine/Source/Scene/I3SField.js
```diff
@@ -252,4 +272,109 @@ I3SField.prototype._parseBody = function (dataView, offset) {
...
+    }
+  }
+  if (dataView.byteLength < offset) {
+    console.error("Invalid attribute buffer size");
+    console.error(
+      ` (field: ${this.name}, expected: ${offset}, actual: ${dataView.byteLength})`
+    );
+    return false;
```
ggetz, line 375: Same comments as above.

packages/engine/Source/Scene/I3SGeometry.js
```diff
@@ -267,6 +269,18 @@ I3SGeometry.prototype.getClosestPointIndexOnTriangle = function (px, py, pz) {
   };
 };
 
+function convertColorFactor(factor) {
+  const convertedFactor = [];
+  for (let i = 0; i < factor.length; i++) {
```
ggetz, line 274: Based on our coding guide, this should be.
 
```suggestion
  const length = factor.length;
  for (let i = 0; i < length; i++) {
```

packages/engine/Source/Scene/I3SGeometry.js
```diff
@@ -382,6 +417,34 @@ I3SGeometry.prototype._generateGltf = function (
     ];
   }
 
+  const gltfMaterials = [];
+  for (let meshIndex = 0; meshIndex < meshes.length; meshIndex++) {
```
ggetz, line 421: See comment above.

packages/engine/Source/Scene/I3SGeometry.js
```diff
@@ -382,6 +417,34 @@ I3SGeometry.prototype._generateGltf = function (
...
+    const primitives = meshes[meshIndex].primitives;
+    for (
+      let primitiveIndex = 0;
+      primitiveIndex < primitives.length;
```
ggetz, line 425: See comment above.

packages/engine/Source/Scene/I3SGeometry.js
```diff
@@ -382,6 +417,34 @@ I3SGeometry.prototype._generateGltf = function (
...
+      primitiveIndex++
+    ) {
+      const primitive = primitives[primitiveIndex];
+      if (defined(primitive.material)) {
+        // Create as many copies of the material as specified in the mesh primitives
```
ggetz, line 430: Why are these copies needed?

packages/engine/Source/Scene/I3SNode.js
```diff
@@ -225,23 +242,46 @@ I3SNode.prototype.loadFields = function () {
...
 
   return Promise.all(promises);
 };
 
+/**
+ * Loads the node field.
+ * @param {string} name The field name
+ * @returns {Promise<void>} A promise that is resolved when the I3S Node field is loaded
+ */
+I3SNode.prototype.loadField = function (name) {
```
ggetz, line 266: `name` should be validated with a `Check` function.

packages/engine/Source/Scene/I3SSublayer.js
```diff
@@ -0,0 +1,177 @@
...
+    },
+  },
+
+  /**
+   * Determines if the sublayer will be shown.
+   * @memberof I3SSublayer.prototype
+   * @type {boolean}
+   * @readonly
+   */
+  show: {
```
ggetz, line 116: What's the difference between `visibility` and `show`?

packages/engine/Specs/Scene/I3SDataProviderSpec.js
```diff
@@ -168,6 +307,249 @@ describe("Scene/I3SDataProvider", function () {
...
+          ...mockLayerData,
+          href: `layers/${id}/`,
+          id: parseInt(id),
+        };
+      } else {
+        mockedData = mockBSLProviderData;
+      }
+      return Promise.resolve(mockedData);
+    });
+    spyOn(Cesium3DTileset, "fromUrl").and.callFake(async () => {
```
ggetz, line 326: Throughout these units tests, I'm concerned this is mocking too much functionality for the test to be helpful.

packages/engine/Source/Scene/I3SDataProvider.js
```diff
@@ -70,6 +75,10 @@ import Rectangle from "../Core/Rectangle.js";
...
  * @property {boolean} [traceFetches=false] Debug option. When true, log a message whenever an I3S tile is fetched.
  * @property {Cesium3DTileset.ConstructorOptions} [cesium3dTilesetOptions] Object containing options to pass to an internally created {@link Cesium3DTileset}. See {@link Cesium3DTileset} for list of valid properties. All options can be used with the exception of <code>url</code> and <code>show</code> which are overridden by values from I3SDataProvider.
+ * @property {boolean} [enableFeatureFiltering=false] Determines if the features will be shown.
+ * @property {boolean} [adjustMaterialAlphaMode=false] The option to adjust the alpha mode of the material based on the transparency of the vertex color. When <code>true</code>, the alpha mode of the material (if not defined) will be set to BLEND for geometry with any transparency in the color vertex attribute.
```
ggetz, line 79: Is there a reason this is an option rather than the default behavior?

packages/widgets/Source/I3SBSLExplorer/I3SBSLExplorer.js
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @param {string} containerId The DOM element ID that will contain the widget.
+ * @param {I3SDataProvider} i3sProvider I3S Data provider instance.
+ *
+ * @demo {@link https://sandcastle.cesium.com/index.html?src=I3S%20Building%20Scene%20Layer.html|I3S Building Scene Layer}
+ */
+function I3SBSLExplorer(containerId, i3sProvider) {
+  const container = document.getElementById(containerId);
+
+  if (!defined(container)) {
```
ggetz, line 19: Use `Check` instead of throwing the developer error directly.

packages/engine/Source/Scene/I3SDataProvider.js
```diff
@@ -70,6 +75,10 @@ import Rectangle from "../Core/Rectangle.js";
...
+ * @property {boolean} [applySymbology=false] Determines if the I3S symbology will be parsed and applied for the layers.
```
ggetz, line 80: I'm curious why this is an option. I understand it parses and applies styling like outlines. Is there a reaon this would be defined, but the user would not want to use it?

ggetz (commented):
@Tamrat-B I've taken an initial pass on the code and have a few questions, as well as some code suggestions based on our Coding Guide.

I noticed the demo is a bit slow. Is there anything we can do to address this? @lilleyse for advice.

Before this is merged, there a few things that should be addressed:

- Update CHANGES.md with a summary of the fix
- Any new contributors should be added to `CONTRIBUTORS.md`

Apps/Sandcastle/gallery/I3S Building Scene Layer.html
```diff
@@ -0,0 +1,232 @@
...
+        );
+        const tours = {
+          "Turanga Library": resource.url,
+        };
+
+        try {
+          // Create i3s and Cesium3DTileset options to pass optional parameters useful for debugging and visualizing
+          const cesium3dTilesetOptions = {
+            skipLevelOfDetail: false,
+            debugShowBoundingVolume: false,
```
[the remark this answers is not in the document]
Tamrat-B: Removed false parameters. Also removed Cesium3DTilesetOptios.

Apps/Sandcastle/gallery/I3S Building Scene Layer.html
```diff
@@ -0,0 +1,232 @@
...
+            // maximumScreenSpaceError: 1, //setting to lower values will improve (quality-wise) the level of detail refinement by forcing higher lod tile descendants to 'show up sooner'.
+            // outlineColor: Cesium.Color.BLUE, // To replace/override the outline color defined in I3S symbology
```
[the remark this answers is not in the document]
Tamrat-B: Done. Moved to doc.

packages/engine/Source/Core/srgbToLinear.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param {number} value The color value in sRGB color space.
+ * @returns {number} Returns the color value in linear color space.
+ *
+ * @example
+ * const srgbColor = [0.5, 0.5, 0.5];
+ * const linearColor = srgbColor.map(function (c) {
+ *     return Cesium.srgbToLinear(c);
+ * });
+ */
+function srgbToLinear(value) {
```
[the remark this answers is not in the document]
Tamrat-B: Done.

packages/engine/Source/Core/srgbToLinear.js
```diff
@@ -0,0 +1,26 @@
...
+  if (value <= 0.04045) {
+    // eslint-disable-next-line no-loss-of-precision
+    return value * 0.07739938080495356037151702786378;
```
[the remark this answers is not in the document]
Tamrat-B: We got the implementation for srgbToLinear conversion from existing Cesium engine code. But that function is private and not available for other classes. We just created new function with the same implementation in the Core package. We think a lower precision won’t make any visual difference but tried to follow the Cesium style. Since we just reused this implementation, we are ready to reduce the precision or change the implementation based on Cesium’s suggestions.

Apps/Sandcastle/gallery/I3S Building Scene Layer.html
```diff
@@ -0,0 +1,232 @@
...
+          msaaSamples: 4, // enables Multisample anti-aliasing for improved visual quality
+        });
+
+        // More datasets to tour can be added here...
+        // The url passed to I3SDataProvider supports loading a single Indexed 3D Scene (I3S) layer (.<host>/SceneServer/layers/<id>) or a collection of scene layers (.<host>/SceneServer) from a SceneServer.
+        const resource = new Cesium.Resource(
+          "https://tiles.arcgis.com/tiles/cFEFS0EWrhfDeVw9/arcgis/rest/services/Turanga_Library/SceneServer"
+        );
+        // Set a query parameter for json to avoid failure on html pages
+        resource.setQueryParameters(
```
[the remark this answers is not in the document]
Tamrat-B: Done.

Apps/Sandcastle/gallery/I3S Building Scene Layer.html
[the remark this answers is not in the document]
Tamrat-B: Done. Removed these descriptions from sample and moved to doc.

packages/engine/Source/Scene/I3SSublayer.js
Tamrat-B: Visibility is a state of the sublayer checkbox, while show returns whether the sublayer is visible or not depending on its visibility and visibility of all its parents.
Building scene layers consist of a hierarchy of sublayers (disciplines and categories). Each sublayer could be shown\hidden by checkbox. Visibility property contains the state of this checkbox.
At the same time, a sublayer parent’s visibility may be false while the sublayer’s visibility is true. In this case the sublayer should be hidden as well and show property will return false. But the checkbox state (visibility property) will remain true.

packages/engine/Specs/Scene/I3SDataProviderSpec.js
Tamrat-B: I3S functionality highly depends on external resources. To remove this dependency and make tests as simple as possible we mocked up the data and returns. This allows to split functionality to small parts and test each part separately. We tried to follow the style of existing I3S tests.

packages/widgets/Source/I3SBSLExplorer/I3SBSLExplorer.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SDataProvider.js
Tamrat-B: This functionality does additional calculations to generate outlines indices by vertex positions. Disabling this option could save the decoding time when the performance is critical.
applySymbology option applies outlines styling if it is defined in I3S data. We use CESIUM_primitive_outline extension for glTF to draw outlines. This extension requires indices for all outlines lines. I3S data doesn’t contain this information. We generate these indices during I3S decoding, and edges detection by vertex positions takes additional time. Also, CESIUM_primitive_outline extension doesn’t support different outlines colors for features of the same Cesium3DTileset, though it may be required by I3S symbology.
We didn’t find another way in Cesium engine for outlines support, so we keep this as option to allow users disable it due to possible performance or visualization issues.

packages/engine/Source/Scene/I3SDataProvider.js
```diff
@@ -335,9 +415,104 @@ I3SDataProvider.prototype.updateForPass = function (frameState, passState) {
...
+    }
+
+    const newLayer = new I3SLayer(provider, data, provider);
+    provider._layers.push(newLayer);
+    if (defined(newLayer._extent)) {
+      provider._layersExtent.push(newLayer._extent);
+    }
+  } else {
+    // Filter other scene layer types out
+    console.log(
```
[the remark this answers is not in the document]
Tamrat-B: I3S BSL supports a mix of layer types: 3D Object and Point. Point layers are not implemented yet. Throwing an error here will cause entire Building Scene Layer to fail, though the main content should be enough to work with a building model. Support for not implemented (or even new) layer types could be added later, but it should not affect the visualization of the implemented layer types. We think the console is a good place to check by user if the scene view doesn’t meet expectations.

packages/engine/Source/Scene/I3SDataProvider.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SField.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SField.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SField.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SField.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SGeometry.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SGeometry.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SGeometry.js
Tamrat-B: Done.

packages/engine/Source/Scene/I3SGeometry.js
Tamrat-B: We need to adjust alphaMode parameter for each unique material that is referenced by mesh primitives, but all other material parameters should be the same for all materials.
This code is required for transparency support. The alpha channel in the vertex color attribute is applied by shaders only if BLEND mode is set for the material. However, BLEND mode affects opaque colors and makes them transparent as well. If any geometry buffer contains both opaque and transparent colors, we need to separate these vertices from each other in different mesh primitives. For each primitive we need to use separate material definition with BLEND mode for transparent colors and OPAQUE for opaque colors. We create copies of the common material definition to keep other material attributes the same.

Tamrat-B:
@ggetz please have a look for the requested changes based on your reviews.

ggetz:
Thanks for the updates @Tamrat-B! We'll take another pass soon.

@lilleyse Let us know if you have any thoughts or suggestions on performance.

Tamrat-B:
@ggetz As to the question regarding performance, we believe there are two main things that affect performance: applySymbology option affecting I3S layer initialization and using a detailed terrain provider (such as ArcGIS elevation provider) seems to affect both initialization and navigation. With disabled terrain and outlines the performance looks quite good.

Unfortunately, if we were to use the default CesiumTerrainProvider provider there is a severe mismatch with the Turanga model and half of it will be submerged underground and is not a good option (though performance with quantized mesh looks good).

As per the open issue: https://github.com/CesiumGS/cesium/issues/8481, it looks currently in CesiumJS the one or two globe picks per frame seems to slow down Cesium for height map based detailed terrain providers like ArcGIS. There is also some community discussions around same topic [here](https://community.cesium.com/t/why-arcgis-tile-terrain-make-my-globe-is-super-slow/8899/6). Note: as suggested in the work around of that issue setting Cesium.ScreenSpaceCameraController.enableCollisionDetection to false seems to help, which we can set in the sample.

packages/engine/Source/Core/srgbToLinear.js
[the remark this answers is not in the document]
ggetz: Got it. The existing implementation is in gltf-pipeline, so I wouldn't recommend refactoring out the code into a shared function.

Ok to leave this as-is.

packages/engine/Source/Scene/I3SDataProvider.js
ggetz: Is there any way to detect a fully opaque i3S data sources before decoding?

If not, let's be explicit in the documentation that there is a performance hit, but it is only needed in the case where transparency support is needed.

packages/engine/Source/Scene/I3SDataProvider.js
ggetz: OK. Let's add a bit more information here that this will generate outlines if present, which could be a performance hit.

packages/engine/Source/Scene/I3SField.js
```diff
@@ -92,40 +101,37 @@ function getNumericTypeSize(type) {
...
-        offset = Math.ceil(offset / valueSize) * valueSize;
-      }
+  if (defined(this._loadPromise)) {
+    return this._loadPromise;
+  }
 
-      that._parseBody(dataView, offset);
-    }
+  this._loadPromise = load(this).catch(function (error) {
+    console.error(error);
```
ggetz, line 132: Ideally, the error should be thrown so that the developer can handle it in their app accordingly. 

Can we remove the catch and console.log?

packages/engine/Source/Scene/I3SSublayer.js
```diff
@@ -0,0 +1,177 @@
...
+    get: function () {
+      return this._visibility;
+    },
+    set: function (value) {
+      //>>includeStart('debug', pragmas.debug);
+      Check.defined("value", value);
+      //>>includeEnd('debug');
+
+      this._visibility = value;
+      for (let i = 0; i < this._i3sLayers.length; i++) {
```
ggetz, line 104: Since this is called for every sublayer recursively, I think it makes sense to check if the value of `this._visibility` has actually changed, and returning early if it has not.

Tamrat-B: done.

packages/engine/Source/Scene/I3SField.js
Tamrat-B: For this one, we recommend keeping the catch. This logic handles any errors with availability or binaries for attributes. When such a condition is present (the issue is related to source I3S data not having complete information and/or unexpected buffer sizes) app developers can do nothing to fix it. Catching these errors allows working with valid attributes, considering invalid attributes as missing. 
See below screenshot; the door asset would still be able to show the attributes associated with it with the error catching and logs the issue; without error catching, this asset wouldn’t be able to display the attribute pop up at all.
image

Tamrat-B:
@ggetz updated based on latest review. Thanks.

ggetz:
Thanks @Tamrat-B! I was able to sync up with @lilleyse offline. He agrees that for the scope of this PR, the performance is acceptable.

I have a few comments after testing.

1. Would it be possible to start in "Overview" mode? I think it's a bit confusing that nothing appears on screen to start.
image

2. I also noticed in the example that the following is logged to the console for the Sandcastle example. I understand the first message, but would it be possible to pass the geoid service providers like is done in the other I3S examples?
```
Point layer LocationPoints is skipped as not supported.
No Geoid Terrain service provided - no geoid conversion will be performed.
```

3. Given that we're already showing the metadata property values in the info box, can we skip the additional `console.log(`${fieldName}: ${fields[fieldName]}`);` when a feature is selected?

CHANGES.md
```diff
@@ -24,6 +24,8 @@
 - Fog rendering now applies to glTF models and 3D Tiles. This can be configured using `scene.fog` and `scene.atmosphere`. [#11744](https://github.com/CesiumGS/cesium/pull/11744)
 - Added `scene.atmosphere` to store common atmosphere lighting parameters. [#11744](https://github.com/CesiumGS/cesium/pull/11744) and [#11681](https://github.com/CesiumGS/cesium/issues/11681)
 - Added `createWorldBathymetryAsync` helper function to make it easier to load Bathymetry terrain. [#11790](https://github.com/CesiumGS/cesium/issues/11790)
+- The `Cesium3DTileset.dynamicScreenSpaceError` optimization is now enabled by default, as this improves performance for street-level horizon views. Furthermore, the default settings of this feature were tuned for improved performance. `Cesium3DTileset.dynamicScreenSpaceErrorDensity` was changed from 0.00278 to 0.0002. `Cesium3DTileset.dynamicScreenSpaceErrorFactor` was changed from 4 to 24. [#11718](https://github.com/CesiumGS/cesium/pull/11718)
+- Added support for I3S Building Scene Layer. [#11678](https://github.com/CesiumGS/cesium/pull/11678)
```
ggetz, line 28: Please merge in `main` and move this bullet to the March 1 release.

Tamrat-B:
@ggetz we've resolved all the issues you brought up in the latest reviews (1-3 above). Thanks!

ggetz:
Thanks @Tamrat-B! 

I did tweak the naming of "I3SBSL*" to "I3SBsl*" to be consistant with out [naming conventions](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#naming), and added an additional bullet in `CHANGES.md` for the updates to the widget package.

Assuming CI passes, this should be good to merge.

ggetz:
🎉 Thanks again @Tamrat-B and team!

Tamrat-B:
Thanks much @ggetz @pjcozzi @lilleyse. Looking forward for users using this workflow!

littlekid-zxt:
Hello, I found information for a console while running the sandcastle example I3S Building Scene Layer.



After further analysis, it was found that the problem was I3SDataProvider.fromUrl, and there was indeed a definition of LocationPoints in the Turanga Library.

a4714457082d7b7ee83a0f2da94acdf

At the same time, I also found that the example I3S Building Scene Layer is an upgrade to **I3S 1.3**, but in the introduction of I3SDataProvider.fromUrl it says "Currently supported I3S versions are 1.6 and 1.7/1.8 (**OGC I3S 1.2**)".

72bc69faf9513d8297ea89d72336c88

Is this problem normal, or is there another way to fix it? However, there is no negative impact on the operation of the sample at this time.

Issue #11665, opened 2023-12-02
Wrong conversion of quaternion to axis


javagl:
The `Quaternion.computeAxis` function seems to be questionable in one case, and wrong in another case. When executing the following
```
const q0 = new Cesium.Quaternion(0,0,0,1);
console.log(Cesium.Quaternion.computeAngle(q0));
console.log(Cesium.Quaternion.computeAxis(q0, new Cesium.Cartesian3()));

const q1 = new Cesium.Quaternion(0,0,0,-1);
console.log(Cesium.Quaternion.computeAngle(q1));
console.log(Cesium.Quaternion.computeAxis(q1, new Cesium.Cartesian3()));
```
then the output will be
```
0
(0, 0, 0)
6.283185307179586
(NaN, NaN, NaN)
```
One _could_ make a case for the first part: The unit quaternion `0,0,0,1` describes "no rotation" - so what is the axis of "no rotation"? It's arbitrary. But I'd be **strongly** in favor to return a "valid" (unit-length!!!) axis there as well. So it should probably be `(1,0,0)`.

(I wonder how many "normalized result is not a number" errors have been caused by this under the hood...)

But the second output cannot be justified: These `NaN`s should not be there. One could argue whether it is a rotation about 0° or about 360°, of course, but `computeAngle` returns 360°, so the rotation axis could just be `(1,0,0)` as well.

Both cases boil down to [this part of the code](https://github.com/CesiumGS/cesium/blob/b9db20e9d56c3b01f56304a51e5440e2b16a5159/packages/engine/Source/Core/Quaternion.js#L680), which currently says
```
  const w = quaternion.w;
  if (Math.abs(w - 1.0) < CesiumMath.EPSILON6) {
    result.x = result.y = result.z = 0;
    return result;
  }
```

If everybody agrees, I could create a PR to change that to
```
  const w = quaternion.w;
  if (Math.abs(w - 1.0) < CesiumMath.EPSILON6 || Math.abs(w + 1.0) < CesiumMath.EPSILON6) {
    result.x = 1;
    result.y = result.z = 0;
    return result;
  }
```
(And update/add the unit tests, which probably do not cover that right now)

Then, the output will be
```
0
(1, 0, 0)
6.283185307179586
(1, 0, 0)
```

---

The following is ... somewhat unrelated, but I noticed this while debugging a bit: 

EDIT: Moved to https://github.com/CesiumGS/cesium/issues/11685

javagl:
[...]

PR #11684, opened 2023-12-11
Fix `Quaternion.computeAxis`

javagl:
Addresses https://github.com/CesiumGS/cesium/issues/11665 : 

Use the x-axis as the result for the unit quaternion and for the quaternion that describes as rotation about 360 degrees.


javagl:
(Not sure which parts of the build failures are related to this PR itself, but ... I assume ... none of them...)

ggetz:
> (Not sure which parts of the build failures are related to this PR itself, but ... I assume ... none of them...)

Huh. @javagl Would you mind merging in `main` just to confirm there's no discrepancies?

javagl:
Not sure what happened here - I'm pretty sure that I branched off from `main`, but now merging in main caused something to wreak havoc up to a point where the pre-commit hooks couldn't be run.  Some juggling in the style of xkcd 1597 may have fixed this, otherwise, this will have to be closed and fixed at a later point in time.

ggetz:
@javagl The CI failure is fixed in https://github.com/CesiumGS/cesium/pull/11689#event-11227052354. I'm going to update the branch for you.

ggetz:
Thanks @javagl!

Issue #6167, opened 2018-01-27
Pull request template


pjcozzi:
You can now make pull request templates as well as different templates for issues, e.g., bugs vs. features.

https://github.com/blog/2495-multiple-issue-and-pull-request-templates

For this repo and other repos in our organization.

hpinkos:
We were talking about this a little bit today.  It might be helpful to have a small checklist like 
- [ ] Test coverage
- [ ] Runs in Firefox
- [ ] Runs in IE
- [ ] Runs in Edge

We've had some problems with big features we've merged lately breaking cross-browser compatibility and this would help us catch that before it gets merged instead of when we're putting together the release

ggetz:
Maybe as part of this (or the concierge update that covers the checklist), prompt for a maintainer to be @ mentioned or assigned as a reviewer.

I see a lot of PRs sit there since no one is directly "responsible" for reviewing.

PR #11688, opened 2023-12-12
Added PR Template

ggetz:
# Description

We've gotten rid of the Cesium Concierge, who will no longer post [handy comments like this](https://github.com/CesiumGS/cesium/pull/11581#issuecomment-1775863898).

While CLA checking is still in the works as part of a separate effort, this PR template aims to replace the functionality of that comment.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/6167

## Testing plan

I've used the template for this PR. 
It will show up for PRs once merged into `main`.

# Author checklist

- [ ] ~I have updated `CHANGES.md` with a short summary of my change~ N/A
- [ ] ~I have added or updated unit tests to ensure consistant code coverage~ N/A
- [ ] ~I have update the inline documentation, and included code examples where relevant~ N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,35 @@
+<!--
+Thanks for the Pull Request!
+
+Please review [Contribution Guide](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md) before opening your first Pull Request.
+
+To ensure your Pull Request is reviewed and accepted quickly, please refer to our [Pull Request Guidelines](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#pull-request-guidelines).
+
+--->
+
+# Description
+
+<!--- Describe your changes in detail -->
+
+<!--- Consider: Why is this change required? What problem does it solve? -->
+
+<!--- Include screenshots if appropriate --->
+
+## Issue number and link
+
+<!--- If it fixes an open issue, link to the issue here -->
+
+<!--- Consider: If suggesting a new feature or change, discuss it in an issue first. -->
+
+## Testing plan
+
+<!--- Describe in detail how you tested your changes -->
+
+# Author checklist
+
+- [ ] I have submitted a Contributor License Agreement
+- [ ] I have added my name to `CONTRIBUTORS.md`
+- [ ] I have updated `CHANGES.md` with a short summary of my change
+- [ ] I have added or updated unit tests to ensure consistant code coverage
+- [ ] I have update the inline documentation, and included code examples where relevant
+- [ ] I have performed a self-review of my code
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,21 +9,21 @@ To ensure your Pull Request is reviewed and accepted quickly, please refer to ou
 
 # Description
 
-<!--- Describe your changes in detail -->
+<!-- Describe your changes in detail -->
 
-<!--- Consider: Why is this change required? What problem does it solve? -->
+<!-- Consider: Why is this change required? What problem does it solve? -->
 
-<!--- Include screenshots if appropriate --->
+<!-- Include screenshots if appropriate -->
 
 ## Issue number and link
 
-<!--- If it fixes an open issue, link to the issue here -->
+<!-- If it fixes an open issue, link to the issue here -->
 
-<!--- Consider: If suggesting a new feature or change, discuss it in an issue first. -->
+<!-- Consider: If suggesting a new feature or change, discuss it in an issue first. -->
 
 ## Testing plan
 
-<!--- Describe in detail how you tested your changes -->
+<!-- Describe in detail how you tested your changes. If this fixes a bug, list the steps to reproduce the original issue. -->
 
 # Author checklist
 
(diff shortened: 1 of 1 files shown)


ggetz:
The CI failure is due to the latest update of our eslint rules. I'll open a separate PR to resolve it.

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+
+--->
+
+# Description
+
+<!--- Describe your changes in detail -->
+
+<!--- Consider: Why is this change required? What problem does it solve? -->
+
+<!--- Include screenshots if appropriate --->
```
jjspace, line 16: I think it would be better to condense these to a single comment that pr authors can write in before or after. Unless the intention is that authors specifically answer each of these separately then it works as is.

```suggestion
<!--
  Describe your changes in detail
  Consider: Why is this change required? What problem does it solve?
  Include screenshots if appropriate
-->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Issue number and link
+
+<!--- If it fixes an open issue, link to the issue here -->
+
+<!--- Consider: If suggesting a new feature or change, discuss it in an issue first. -->
```
jjspace, line 22: Same here

```suggestion
<!-- 
  If it fixes an open issue, link to the issue here
  Consider: If suggesting a new feature or change, discuss it in an issue first.
-->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Testing plan
+
+<!--- Describe in detail how you tested your changes -->
```
jjspace, line 26:
```suggestion
<!--
  Describe in detail how you tested your changes and how reviewers should verify those changes
  If this fixes a bug list the steps to reproduce the original issue
-->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
+<!--
+Thanks for the Pull Request!
+
+Please review [Contribution Guide](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md) before opening your first Pull Request.
+
+To ensure your Pull Request is reviewed and accepted quickly, please refer to our [Pull Request Guidelines](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#pull-request-guidelines).
+
+--->
```
jjspace, line 8: nitpick
```suggestion
-->
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,35 @@
...
+<!--- If it fixes an open issue, link to the issue here -->
+
+<!--- Consider: If suggesting a new feature or change, discuss it in an issue first. -->
+
+## Testing plan
+
+<!--- Describe in detail how you tested your changes -->
+
+# Author checklist
+
```
jjspace, line 29: Maybe an extra comment should be included here for how to handle items that don't apply to a given PR? Should they be crossed out? deleted? checked but include `N/A`?
Just a one line suggestion to avoid confusion

jjspace (commented):
Looks good, couple small comments.

It seems like the main two things losing Concierge lost was checking for a lack of changes in specific files. I feel like there should be solutions to do that same check in GHA if we wanted to instead of the checklist that could quickly start to get ignored? Just a thought

.github/pull_request_template.md
ggetz: I'd rather have people answer in different ways than not answer at all– And I think checkboxes are about as low effort as they can get.

If that's too confusing though, I think we'll be able to tell from the submitted PR descriptions and we can alter it later. What do you think?

.github/pull_request_template.md
ggetz: Though I will fix the inconsistent number of `-` character on the single comments if you agree with keeping them separate.

ggetz:
> It seems like the main two things losing Concierge lost was checking for a lack of changes in specific files. I feel like there should be solutions to do that same check in GHA if we wanted to instead of the checklist that could quickly start to get ignored? Just a thought 

Good point. We might be able to do so with GitHub actions instead. I'll take a quick look. Either way, that can be a follow-up to this PR for the time being.

.github/pull_request_template.md
jjspace: We can leave it out for now if you want. I was just trying to get ahead of someone asking "What do I put if one doesn't apply? do I check it or not?"

.github/pull_request_template.md
jjspace: I think that it's likely there would be cases of myself or others filling out the template and doing parts 2 and 3 under the first then having to extract that information to the other parts if that's the intention. I was thinking along the lines of it being better to include the whole "prompt" and let the author decide how to structure it all together.

I don't have a strong opinion either way though. If you wanna stick with separate comments that's fine, just update the `-` and I can merge

ggetz:
Thanks @jjspace! Updated.

jjspace (approved):
Thanks @ggetz LGTM!

jjspace:
@ggetz One downside to doing the confirmations with checkboxes is that GH considers them tasks and will include the completion count in a handful of places around the site. Might still be handy but it does add a bit of distraction/confusion especially if there is an "actual" tasklist in a PR

PR #11689, opened 2023-12-12
Fix global require usage

ggetz:
This was failing since our update to the eslint rules. Resolve with a ternary operator as suggest in [the `n/global-require` docs](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/global-require.md)

It was only be picked up during CI since the local commit hook is only ran on changed files.

### DIFF
--- a/index.cjs
+++ b/index.cjs
@@ -2,15 +2,11 @@
 "use strict";
 
 const path = require("path");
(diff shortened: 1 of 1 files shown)


jjspace (approved):
Apologies I should've checked this after my change in https://github.com/CesiumGS/eslint-config-cesium/pull/5

LGTM!

PR #11691, opened 2023-12-12
Run dev workflow for external contributor PRs

ggetz:
# Description

I noticed that https://github.com/CesiumGS/cesium/pull/11678 and other external PRs do not have any GitHub Action workflows run. Nor do I see the approval workflow as described in [Approving workflow runs from public forks](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).

While its not very clear from the GitHub actions documentation, it appears that the external contributor workflows are expected to be triggered from PRs rather than pushes.

This PR also cleans up some missed references when transitioning from Node 16 and 18 to Node 18 and 20 to reflect LTS status.

## Issue number and link

N/A

## Testing plan

This should:
1) Trigger the dev workflow for this PR once opened
2) Once merged into main, newly opened external PRs and their updates should trigger the dev workflow.

# Author checklist

- [ ] ~I have updated `CHANGES.md` with a short summary of my change~ N/A
- [ ] ~I have added or updated unit tests to ensure consistant code coverage~ N/A
- [ ] ~I have update the inline documentation, and included code examples where relevant~ N/A

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -4,6 +4,7 @@ on:
     branches-ignore: 
       - 'cesium.com'
       - production
+  pull_request:
 concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -1,12 +1,11 @@
 name: dev
 on: 
   push:
-    branches-ignore: 
-      - 'cesium.com'
-      - production
+    branches: 
+      - main
   pull_request:
 concurrency:
-  group: ${{ github.ref }}
+  group: dev-${{ github.ref }}
   cancel-in-progress: true
 jobs:
   lint:
@@ -34,10 +33,10 @@ jobs:
       BRANCH: ${{ github.ref_name }}
     steps:
       - uses: actions/checkout@v3
-      - name: install node 18
+      - name: install node 20
         uses: actions/setup-node@v3
         with:
-          node-version: '18'
+          node-version: '20'
       - name: npm install
         run: npm install
       - name: build
@@ -51,10 +50,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: install node 18
+      - name: install node 20
         uses: actions/setup-node@v3
         with:
-          node-version: '18'
+          node-version: '20'
       - name: npm install
         run: npm install
       - name: release build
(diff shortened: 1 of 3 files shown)


ggetz:
After speaking offline with @jjspace, we agreed on avoiding the double runs on PR updates where possible.

To avoid the double runs, we're going to run the deployment only for pushes (as these cannot be run by external contributors anyway), expect for those to the `main` or production branches. The the rest of the dev workflow will run on the `pull_request` trigger.

Ex.
````
on:
  push:
    branches:
      - main
      - production
      - "cesium.com"
  pull_request:
 ````

**Comment by ggetz** [comment]
@jjspace Updated!

**Comment by ggetz** [comment]
@jjspace Please review when you get the chance.

File: .github/workflows/main.yml [hunk 1]
```diff
@@ -1,7 +1,7 @@
 name: main
 on: 
   workflow_run:
-    workflows: [dev]
+    workflows: [dev, prod]
```
jjspace, line 4: Is this supposed to use the new workflow's name?
```suggestion
    workflows: [dev, deploy]
```

.github/workflows/dev.yml
```diff
@@ -33,10 +33,10 @@ jobs:
       BRANCH: ${{ github.ref_name }}
     steps:
       - uses: actions/checkout@v3
-      - name: install node 18
+      - name: install node 20
```
jjspace, line 36: sanity check, did we mean to update to v20 in this PR alongside the other changes?

.github/workflows/dev.yml
```diff
@@ -1,11 +1,11 @@
 name: dev
 on: 
   push:
-    branches-ignore: 
-      - 'cesium.com'
-      - production
+    branches: 
+      - main
+  pull_request:
```
jjspace, line 6: I don't know enough about GHA but I just want to confirm this is actually what we want to do. There's a warning in the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) saying that:
> workflows that are triggered by the `pull_request_target` event, the `GITHUB_TOKEN` is granted read/write repository permission unless the `permissions` key is specified and the workflow can access secrets, even when it is triggered from a fork.

It also says these are run in the context of the base branch. Does that mean it's trying to access the secrets from the PR repo's fork? or giving access to our secrets from our repo?

jjspace (commented):
Just had a couple questions

.github/workflows/dev.yml
ggetz: Apologies, I didn't mention this in the PR description. I will add it.

This cleaned up some missed references [when transitioning from Node 16 and 18 to Node 18 and 20 to reflect LTS status](https://github.com/CesiumGS/cesium/pull/11598).

.github/workflows/main.yml
ggetz: `main` reports a failure on the main or production branches to slack since we'll want to resolve any issue here ASAP. 

I omitted `deploy`, but it could be included here as well.

.github/workflows/dev.yml
ggetz: That is true for the `pull_request_target` event.  This is the `pull_request` event, which only grants those permissions to _internal_ contributors, not external contributors.

For external contributors, the docs say the following:

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The GITHUB_TOKEN has read-only permissions in pull requests from forked repositories. For more information, see "[Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)."

ggetz:
@jjspace I answered your questions above. Is this OK to merge?

jjspace:
@ggetz Yes, I was looking in the wrong part of the docs. LGTM :pray:

Issue #11592, opened 2023-10-26
Make collections iterable


angrycat9000:
Right now to iterate over a collection I have do a for loop with the index and then call  `get` to get the object.
```js
for(let i = 0; i < viewer.imageryLayers.length; i++ ) {
  const layer = viewer.imageryLayers.get(i);
  // do something with layer
}
```

It would be nicer if the collections provided iterators so I could do:
```js
for(const layer of viewer.imageryLayers) {
   // do something with layer
}
```

javagl:
I like the concepts of Iterators and Generators (also "Streams" in other languages), and think that this could be a low-hanging fruit to increase usability, readability, and maintainability. Particularly in cases where people felt the urge to pull out the collection and its length...
```
const theLayers = viewer.imageryLayers;
const theLength = thelayers.length;
for(let i = 0; i < theLength; i++ ) {
  const layer = theLayers.get(i);
  // do something with layer
}
```
because of ~"performance". 

(An aside: One could be tempted to quote Knuth here and talk about "premature optimization". And ... if someone claims that pulling the `length` into a constant increases the performance, then I'd like to see the benchmark. *But*... due to quirks of the language, this _may_ actually make a difference in _some_ cases - at least theoretically: Nobody knows whether something like `collection.length` only returns a value of a propery, or whether it calls a 'getter' that internally iterates the whole collection or whatnot...)

These point is: There are many _possible_ considerations for "how to write 'good' code" (and we aren't even talking about `i++` vs. `++i` yet). All these questions could become obsolete, when the answer is: "'Idiomatic code' is 'good code' - just use a standard `for...of` loop".

Still, addressing this in all depth _does_ raise further questions. 

One is whether the coding guide at https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md should be updated accordingly (it uses the `const length = ...` pattern a lot...). 

Another one is whether we should 
1. try to find some sensible "benchmark" run that is representative for "CesiumJS as a whole"...
2. change **ALL** iterations in the whole CesiumJS codebase from the `i < length` pattern to the `for...of` pattern
3. do the same benchmark run again, and see whether there is a difference

---

A draft PR with a few, first iterators for the `...Collection` classes is at https://github.com/CesiumGS/cesium/pull/11697

PR #11697, opened 2023-12-15
Add iterator to all collections

javagl:
A draft for how https://github.com/CesiumGS/cesium/issues/11592 could be addressed: It just adds the `[Symbol.iterator]` function to all `...Collection.js` types, with the same pattern in each case, e.g. 
```
/**
 * Returns an iterator over the elements of this collection.
 *
 * @returns {Iterator<ImageryLayer>} The iterator
 */
ImageryLayerCollection.prototype[Symbol.iterator] = function* () {
  yield* this._layers;
};
```
(with `OrderedGroundPrimitiveCollection` and `CompositeEntityCollection` being the only exceptions here, because they are "nested" colletions)

---

For example, the simplified iteration would look like this:
```javascript
const labels = new Cesium.LabelCollection();
labels.add({
  position : new Cesium.Cartesian3(1.0, 2.0, 3.0),
  text : 'A label'
});
labels.add({
  position : new Cesium.Cartesian3(4.0, 5.0, 6.0),
  text : 'Another label'
});

// Old:
for (let i=0; i<labels.length; i++) {
  const label = labels.get(i);
  console.log(label.text);
}

// New
for (const label of labels) {
  console.log(label.text);
}
```

---

### DIFF
--- a/Specs/TypeScript/index.ts
+++ b/Specs/TypeScript/index.ts
@@ -403,17 +403,16 @@ if (defined(pos)) {
   consumeDefined(pos);
 }
 
-
 // Verify that a collection (like LabelCollection) is iterable
 // and can be used in a for-of loop
 const labels = new LabelCollection();
 labels.add({
-  position : new Cartesian3(1.0, 2.0, 3.0),
-  text : 'A label'
+  position: new Cartesian3(1.0, 2.0, 3.0),
+  text: "A label",
 });
 labels.add({
-  position : new Cartesian3(4.0, 5.0, 6.0),
-  text : 'Another label'
+  position: new Cartesian3(4.0, 5.0, 6.0),
+  text: "Another label",
 });
 
 for (const label of labels) {
--- a/Specs/TypeScript/index.ts
+++ b/Specs/TypeScript/index.ts
@@ -50,6 +50,7 @@ import {
   ImageryProvider,
   IonImageryProvider,
   KmlDataSource,
+  LabelCollection,
   MapboxImageryProvider,
   MapboxStyleImageryProvider,
   MaterialProperty,
@@ -400,4 +401,21 @@ function consumeDefined(pos: Cartesian3) {
 pos = undefined;
 if (defined(pos)) {
   consumeDefined(pos);
+}
+
+
+// Verify that a collection (like LabelCollection) is iterable
+// and can be used in a for-of loop
+const labels = new LabelCollection();
+labels.add({
+  position : new Cartesian3(1.0, 2.0, 3.0),
+  text : 'A label'
+});
+labels.add({
+  position : new Cartesian3(4.0, 5.0, 6.0),
+  text : 'Another label'
+});
+
+for (const label of labels) {
+  console.log(label.text);
 }
--- a/Specs/TypeScript/tsconfig.json
+++ b/Specs/TypeScript/tsconfig.json
@@ -1,5 +1,10 @@
 {
   "compilerOptions": {
+    "lib": [
+      "ES2015",
+      "DOM"
+    ],
+    "downlevelIteration": true,
     "noEmit": true,
     "types": [],
     "strict": true,
--- a/Tools/jsdoc/tsconfig.json
+++ b/Tools/jsdoc/tsconfig.json
@@ -1,5 +1,9 @@
 {
   "compilerOptions": {
+      "lib": [
+        "ES2015",
+        "DOM"
+      ],
       "noEmit": true,
       "types": []
   },
--- a/packages/engine/Source/Core/TimeIntervalCollection.js
+++ b/packages/engine/Source/Core/TimeIntervalCollection.js
@@ -16,6 +16,7 @@ function compareIntervalStartTimes(left, right) {
 /**
  * A non-overlapping collection of {@link TimeInterval} instances sorted by start time.
  * @alias TimeIntervalCollection
+ * @implements {Iterable<TimeInterval>}
  * @constructor
  *
  * @param {TimeInterval[]} [intervals] An array of intervals to add to the collection.
--- a/packages/engine/Source/Core/TimeIntervalCollection.js
+++ b/packages/engine/Source/Core/TimeIntervalCollection.js
@@ -1133,4 +1133,14 @@ TimeIntervalCollection.fromIso8601DurationArray = function (options, result) {
     result
   );
 };
+
+/**
+ * Returns an iterator over the elements of this collection.
+ *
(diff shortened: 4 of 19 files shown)


ggetz:
Hi @javagl. I think using iterators is a fine suggestion. I'm curious what would be left here to make this PR ready for proper review.

javagl:
What's missing here is mainly tests. I'll try to allocate some time to wrap this up.

javagl:
In order to convince JSDoc that the results are "valid" (nowadays), I had to do two things:

1. Add `@implements {Iterable<TimeInterval>}` to the "class" declarations
2. Add the proper hints and `lib` entries to the `tsconfig.json` files

The latter refers to  
```
    "lib": [
      "ES2015",
      "DOM"
    ],
    "downlevelIteration": true,
```
that had to be added - see https://github.com/CesiumGS/cesium/pull/11697/commits/776be36a40af7be4af3952c70a1c634aab3acf63#diff-7f5d1dee481aa2b1db50d06034ef162c80c4cfaa5541e119b061a26798276e9cR3 

Locally, running a sandcastle with a `for-of` loop (like in the first post here) now works, which originally resulted in the expected
`TypeError: labels is not iterable (on line 20)`
message.

# But ...  

... I'm not sure whether adding that `lib` information may have undesired implications (like CesiumJS no longer running in Mosaic 2.1.1 or so...). 

If the change itself is OK, I'll add some tests. (The most important ones being the `OrderedGroundPrimitiveCollection` and `CompositeEntityCollection`, because they are "nested" and the first one seems to claim an ordering that will have to be verified as part of the spec).

javagl:
There are probably many things "wrong" (or at least "questionable") with this benchmark...
```

function createLabels(n) {
  const labels = new Cesium.LabelCollection();
  for (let i=0; i<n; i++) {
    labels.add({
      position : new Cesium.Cartesian3(n, 0, 0),
      text : 'Label' + n
    });
  }
  return labels;
}

const iterateFor = (labels) => {
  let blackHole = 0;
  for (let i=0; i<labels.length; i++) {
    const label = labels.get(i);
    blackHole += label.text.length;
  }
  return blackHole;
};

const iterateForOf = (labels) => {
  let blackHole = 0;
  for (const label of labels) {
    blackHole += label.text.length;
  }
  return blackHole;
};

function runTimed(name, runs, f) {
  const startMs = performance.now();
  let result = 0;
  for (let r=0; r<runs; r++) {
    result += f();
  }
  const endMs = performance.now();
  const differenceMs = (endMs - startMs);
  console.log(`${name} took ${differenceMs.toFixed(3)} ms, result ${result}`);
}

const runs = 100;
for (let n=10000; n<100000; n+=10000) {
  const labels = createLabels(n);
  runTimed(`n=${n}, for   `, runs, () => iterateFor(labels));
  runTimed(`n=${n}, for-of`, runs, () => iterateForOf(labels));
}
```
but if it is not completely wrong, then the results are ... disappointing:
```
n=10000, for    took 5.400 ms, result 10000000
n=10000, for-of took 31.800 ms, result 10000000
n=20000, for    took 10.800 ms, result 20000000
n=20000, for-of took 51.900 ms, result 20000000
n=30000, for    took 14.600 ms, result 30000000
n=30000, for-of took 74.500 ms, result 30000000
n=40000, for    took 24.800 ms, result 40000000
n=40000, for-of took 119.900 ms, result 40000000
n=50000, for    took 63.900 ms, result 50000000
n=50000, for-of took 200.700 ms, result 50000000
n=60000, for    took 123.100 ms, result 60000000
n=60000, for-of took 216.000 ms, result 60000000
n=70000, for    took 135.300 ms, result 70000000
n=70000, for-of took 245.000 ms, result 70000000
n=80000, for    took 210.000 ms, result 80000000
n=80000, for-of took 314.200 ms, result 80000000
n=90000, for    took 269.000 ms, result 90000000
n=90000, for-of took 398.200 ms, result 90000000
```

It might be worthwhile to create a dedicated performance test (outside of Cesium and a Sandcastle), but ... there is the truth, and there are benchmarks. It's difficult.

ggetz:
Thanks for the performance benchmarks @javagl!

While I definitely like the ease-of-use of the iterators, I'm worried about encouraging users to make us of an API that is significantly slower that the classic `for` loop. What are your thoughts?

javagl:
@ggetz 

This may indeed raise some concerns. 

(I'm assuming that the performance test is not "totally flawed", and that the values are in the right ballpark, at least _relative_ to each other), 

The test was run on Chrome. 
On Edge, the results are similar (I think they are using the same engine anyhow). 
The same test on FireFox gives this:
```
n=10000, for    took 3.000 ms, result 10000000
n=10000, for-of took 106.000 ms, result 10000000
n=20000, for    took 5.000 ms, result 20000000
n=20000, for-of took 212.000 ms, result 20000000
n=30000, for    took 9.000 ms, result 30000000
n=30000, for-of took 320.000 ms, result 30000000
n=40000, for    took 11.000 ms, result 40000000
n=40000, for-of took 425.000 ms, result 40000000
n=50000, for    took 14.000 ms, result 50000000
n=50000, for-of took 594.000 ms, result 50000000
n=60000, for    took 19.000 ms, result 60000000
n=60000, for-of took 692.000 ms, result 60000000
n=70000, for    took 40.000 ms, result 70000000
n=70000, for-of took 1082.000 ms, result 70000000
n=80000, for    took 75.000 ms, result 80000000
n=80000, for-of took 1243.000 ms, result 80000000
n=90000, for    took 97.000 ms, result 90000000
n=90000, for-of took 1494.000 ms, result 90000000
```
Yeah. _One and a half seconds_ for iterating over 90000 elements. Well. 

(I can imagine that it's difficult to create a VM/JIT that is smart enough to detect that these things are plain, dumb for-loops. The crux is very likely in the underlying "language", and the browser people cannot spin gold from straw...)

On the one hand, one could make a case that the `for-of` loop is convenient and idomatic, and for "iterating over 10 labels" the performance difference does not matter. But when people also use it for iterating over 100000 labels, and an operation that took 0.1s suddently takes _fifteen times longer_, that's a no-go for peformance-critical stuff like CesiumJS. 

Maybe @angrycat9000 (who opened the issue that is addressed with this PR) wants to share some thoughts...?

ggetz:
Unless @angrycat9000 chimes in with a compelling use case, I think I'm going to go ahead and close this PR given the performance implications.

@javagl Is that OK for the time being?

angrycat9000:
> Unless @angrycat9000 chimes in with a compelling use case, I think I'm going to go ahead and close this PR given the performance implications.

Nope.  This was just for developer convience.

Issue #11685, opened 2023-12-11
Inconsistent type assumption for scratch variable


javagl:
<sup>(I originally mentioned this in https://github.com/CesiumGS/cesium/issues/11665, but should probably be an own issue)</sup>

The [scratchVariable in ModelAnimationChannel](https://github.com/CesiumGS/cesium/blob/b9db20e9d56c3b01f56304a51e5440e2b16a5159/packages/engine/Source/Scene/Model/ModelAnimationChannel.js#L218) is initialized depending on the "path" of the channel: For `"translation"` animations, it is a `Cartesian3`, and for `"rotation"` animations, it is a `Quaternion`. But ... it is a global variable 😬 When initializing multiple channels, then there are cases where this should be a `Quaternion`, but it actually is a `Cartesian3`, meaning that some of the functions are just shoving some `.w` property into that poor `Cartesian3`. It happens to work (hooray to untyped languages), but ... feels wrong. 

Would the preferred way of fixing this be to convert that scratchVariable an instance variable in the `ModelAnimationChannel`?

ggetz:
I think the preferred solution would be to have two scratch variables, one for each type.

PR #11700, opened 2023-12-16
Use proper type for scratch variables

javagl:
Fixes https://github.com/CesiumGS/cesium/issues/11685 

It is not really "required", and does not solve "problem" (except for that itchy feeling in the head of people who usually use _typed_ languages): The `scratchVariable` that was used either as a `Cartesian3` or as a `Quaternion` has now been split into a `scratchCartesian3` and `scratchQuaternion` that are used depending on the animated `path` (i.e. the cartesian for `"translation"` and `"scale"`, and the quaternion for the `"rotation"`).

Should this be mentioned in `CHANGES.md`? It seems like something that is "too internal" to be mentioned...

I considered to add a unit test - basically consisting of
```
    // Create rotation and translation (leaving the scratch variable to be a Cartesian3)
    const runtimeChannelRotation = new ModelAnimationChannel({.,.});
    const runtimeChannelTranslation = new ModelAnimationChannel({...});
...
    // Animate rotation (using the Cartesian3 scratch variable...)
    runtimeChannelRotation.animate(time);
```
But none of that is "visible" at the outside: The `runtimeNode["rotation"]` is still a `Quaternion`, because the [setter in `ModelRuntimeNode`](https://github.com/CesiumGS/cesium/blob/b57f9fb52e3a6444086922ef0fb4394c0cc2eb51/packages/engine/Source/Scene/Model/ModelRuntimeNode.js#L370) clones the input `value` - which **is** a `Cartesian3`, but it **does** have that `.w` component, so ... it happens to work 😬 If anyone has an idea for how to "test" this, I'd add the tests accordingly...

### DIFF
--- a/packages/engine/Source/Scene/Model/ModelAnimationChannel.js
+++ b/packages/engine/Source/Scene/Model/ModelAnimationChannel.js
@@ -215,8 +215,8 @@ function createSplines(times, points, interpolation, path, count) {
   return splines;
 }
 
-let scratchCartesian3;
-let scratchQuaternion;
+const scratchCartesian3 = new Cartesian3();
+const scratchQuaternion = new Quaternion();
 
 function initialize(runtimeChannel) {
   const channel = runtimeChannel._channel;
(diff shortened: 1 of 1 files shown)


packages/engine/Source/Scene/Model/ModelAnimationChannel.js
```diff
@@ -240,10 +241,10 @@ function initialize(runtimeChannel) {
   switch (path) {
     case AnimatedPropertyType.TRANSLATION:
     case AnimatedPropertyType.SCALE:
-      scratchVariable = new Cartesian3();
+      scratchCartesian3 = new Cartesian3();
```
ggetz, line 244: @javagl Correct me if I'm wrong, but is this switch statement needed anymore? We can just initialize the scratch variables at the module scope, as we do for most cases.

```
const scratchCartesian3 = new Cartesian3();
const scratchQuaternion = new Quaternion();
```

javagl: Sorry, I misinterpreted the comment at https://github.com/CesiumGS/cesium/issues/11685#issuecomment-1852214923 

When it's OK that both are initialized (even if one of them may remain unused), then the switch statement can indeed be removed.

ggetz:
Great, thanks @javagl!

Issue #11708, opened 2023-12-19
`KHR_texture_transform` is not applied to property textures


javagl:
The texture coordinate transformation that is defined with the [`KHR_texture_transform`](https://github.com/KhronosGroup/gltf/tree/main/extensions/2.0/Khronos/KHR_texture_transform) extension is applied to "visible" textures (like the base color texture). But is is apparently not applied to [Property Textures in `EXT_structural_metadata`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata#property-textures).

The following is an archive with everything that is necessary to reproduce the issue:

[propertyTextureTransformTest-2023-12-19.zip](https://github.com/CesiumGS/cesium/files/13718858/propertyTextureTransformTest-2023-12-19.zip)

It contains

- A glTF asset that only contains a unit square. It uses the **same** texture for the base color and for a Property Texture: The texture just contains 8x8 pixels with increasing 'red' component values (so the base color will be black ... red, and the property values in the property texture will be in [0 ... 64)*3. It uses the _same_ texture transform for both usages of the texture, namely with an offset of [0.25, 0.25], and a scale of [0.5, 0.5]. 
- The JSON part of that GLB (for reference and convenience)
- A tileset JSON that refers to that GLB
- A sandcastle for testing

Switching between the "Default Shading" and the "Property Values" (which just renders the property texture values as a 'brighness' value with a custom shader) gives this result:

![Cesium Property Texture Transform Issue](https://github.com/CesiumGS/cesium/assets/5597569/d8fd4226-f973-4735-b9ab-e498fd56a826)

Showing that the texture transform is only applied in one case, but not the other.

**NOTE**: As part of fixing this, it should also be reviewed whether the texture transform is applied to [Feature ID Textures in `EXT_mesh_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features#feature-id-by-texture-coordinates). I have not tested this yet, but assume that these will also not be affected by the texture transform (in doubt, I can create a similar test example if necessary).

ptrgags:
@javagl indeed, `KHR_texture_transform` was only implemented for the material textures. It was not implemented for property textures (until your PR) nor for feature ID textures.

This is the first time I've seen these extensions used together, I think it's a fine addition if it's useful.

javagl:
[...]

PR #11709, opened 2023-12-19
Apply texture transforms to property textures

javagl:
A **DRAFT** addressing https://github.com/CesiumGS/cesium/issues/11708 

With this change, it appears to transform the texture properly:

Cesium Property Texture Transform Issue Fix Draft

This is a first shot (I haven't done anything in the CesiumJS+Metadata+Shader area before), and not yet tested extensively. Maybe someone wants to have a look whether this might go in the right direction. (If it does, the same will probably have to be done for feature ID textures - to be confirmed)

---


packages/engine/Source/Scene/Model/MetadataPipelineStage.js
```diff
@@ -450,9 +457,30 @@ function addPropertyTexturePropertyMetadata(renderResources, propertyInfo) {
...
+      ShaderDestination.FRAGMENT
+    );
+    uniformMap[transformUniformName] = function () {
+      return transform;
+    };
+
+    // Update the expression for the texture coordinates
+    // with one that transforms the texture coordinates
+    // with the transform matrix first
+    texCoordVariableExpression = `vec2(${transformUniformName} * vec3(${texCoordVariable}, 1.0))`;
```
ptrgags, line 481: @javagl to answer your question, I think this inlining is a fine way to implement this extension for property textures (and feature ID textures will be similar).

for the `MaterialPipelineStage`, the use of `#define`s was because the possible textures (base color, normal, etc.) are all known in advance according to the glTF spec.

Meanwhile, metadata is much more variable in the sense that the quantity and names of uniforms depends on the number of properties. This is why this pipeline stage uses a code generation approach.

packages/engine/Source/Scene/Model/MetadataPipelineStage.js
ptrgags, line 1: Some ideas for unit testing:

- Definitely refer to `MetadataPipelineStageSpec` and `MaterialPipelineStageSpec` for examples of testing KHR_texture_transform and property textures
- From #11708, it looks like you have a small glTF example that exercises the feature. Maybe consider making a render test in `ModelSpec` that would render e.g. green if the transform is applied correctly but red if the transform was omitted?

ptrgags (commented):
@javagl I think this approach looks good! I left a couple comments explaining why MetadataPipelineStage works a bit differently than MaterialPipelineStage, and some ideas for unit testing.

javagl:
The `*PipelineSpecs` seem to mainly focus on whether the "ShaderBuilder" generates the right text. I could probably add a few checks there. Some sort of "functional" test would be nice, though.

I tried this...

> Maybe consider making a render test in ModelSpec that would render e.g. green if the transform is applied correctly but red if the transform was omitted?

... but... what to check there? I added test to render the model (and it took me a few hours and some swearing before I resorted to just restoring the default camera orientation after the "fly-to"), and it renders the "invalid" areas (which should not be visible after the transform) in 'red' with a custom shader. But it is not clear what that "RGB value" should be that is checked there. (And... using/setting `debugCanvasWidth` will break this check anyhow, because it will always look at the "upper left pixel" of the rendered image....). I'll try to find a sensible solution there. Something like `expect(allRenderedPixels.toNotBe(red))` could do it, but does not seem to be used anywhere until now.

javagl:
I found it a bit difficult to sensibly structure the test in the `ModelSpec.js`.The fact that there is no "convenience function" like
`expect(magicallyFetchTheTextureValueAt(x,y)).toEqual(expected);`
makes this feel a bit brittle. 

What it is doing now:

- There is that test model with the unit square (as shown in the issue): 
  Cesium Property Texture Transform Issue
- It creates a shader that returns "RED" for "low" and "high" values (i.e. for the values that should be "moved out" of the unit square area by the texture transform
- The rendered square _without_ the bugfix will look like this:
  Cesium Property Texture Transform Rendering
- It is moving the camera at a very specific location in the upper left (at (0.1, 0.1), with a distance of 0.15), meaning that the result of that 1x1-pixel rendering should be "completely red"

It "works" insfofar that this test...
- fails without the fix
- passes with the fix

which might be "enough". But maybe someone has ideas for improving that. The relevant spec starts at https://github.com/CesiumGS/cesium/pull/11709/files#diff-7118f1e2b4539329dcd5314bc1898c6007565fe4e096211621aef6be3d16543dR722

---

Running the tests locally results in a bunch of failed tests, e.g.

> Scene/Scene > pickPosition returns undefined when useDepthPicking is false
> DeveloperError: normalized result is not a number

or 

> Scene/Pick > sampleHeight > samples height from the globe
> Expected 6343428.618064475 to be greater than 6356752.314245179.

(?!?!)

and a few timeouts. But this is unrelated to this change, and has to be tracked or addressed separately.

javagl:
@ptrgags I tried to add some tests here

- The [test in `MetadataPipelineSpec`](https://github.com/CesiumGS/cesium/pull/11709/files#diff-48b8b2d33ff53d6f9231ca677a63b169e9708874f14cb69b04e5eacc12af2709R345) is largely "copy-and-paste" from the other tests - the most specific part is that it checks that the `uniform mat3 u_propertyTexture_0Transform;` is present
- In line with your suggestion of
  > Maybe consider making a render test in ModelSpec that would render e.g. green if the transform is applied correctly but red if the transform was omitted?

  I tried to add such a [test in `ModelSpec`](https://github.com/CesiumGS/cesium/pull/11709/files#diff-7118f1e2b4539329dcd5314bc1898c6007565fe4e096211621aef6be3d16543dR722) that zooms into a region of the quad that is 'red' when the transform is _not_ applied, and then checks that this 1x1 pixel rendering is "not red". It feels a bit brittle, and does not check whether the access "returns the right property values" or so, but ... it 'fails' _without_ the fix, and 'passes' _with_ the fix, so this may be as good as it gets...

packages/engine/Specs/Scene/Model/ModelSpec.js
```diff
@@ -715,6 +719,73 @@ describe(
       });
     });
 
+    it("transforms property textures with KHR_texture_transform", function () {
```
ptrgags, line 722: For new tests, we now prefer to make the spec function `async` and then use `await` rather than promise chains. 

You don't need to update everything in this file (there's a lot of cleanup throughout the codebase), just the specs you added.

Specs/Data/Models/glTF-2.0/PropertyTextureWithTextureTransform/glTF/PropertyTextureWithTextureTransform.gltf
```diff
@@ -0,0 +1,197 @@
...
+                  ]
+                }
+              }
+            }
+          }
+        }
+      ]
+    }
+  }
+}
```
ptrgags, line 197: missing newline at the end of file

packages/engine/Specs/Scene/Model/MetadataPipelineStageSpec.js
```diff
@@ -340,6 +342,65 @@ describe(
       });
     });
 
+    it("Adds property texture transform to the shader", function () {
```
ptrgags, line 345: same here, use `async/await`

CHANGES.md
ptrgags, line 1: Make sure to sync with `main` and fix the merge conflict.

ptrgags (commented):
@javagl just a couple comments.

I'm going to try out the Sandcastle you posted in the original issue next. Do you have any other Sandcastles you've used while testing or just that one?

ptrgags:
I tried your [Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=lVVRb9s2EP4rV6MPEqBQTtK0TZoEQxMEG7C0RePkYdOw0tLJ4kaRAkk5cQL/9x1FSbbTDOseDJMn3vG+77475lpZB0uB92jgDBTewwVa0dbsrrNF2STv9hdaOS4UmmwSf8hUptIULgxyh+Aq+gmJFh0I1W1DwEzlXfjh41lvZzZHhawxohZOLNEyXhRRpgD4PRduyCD8HV7Ogjsrja5vjewOAmSTyrnmJE2lzrmstHUn76fTw7S/jP1ltcomCTyF4wXO28VNpe8/6lYVQi3utGxrPAFnWkz8mXWcqRHatV7uAnMaOORoPAnQaEuJawU6wF1IPfdByJGrAiwlhEBA2makwHBlS21qIqGHNxtMliG37pM2rrptZvpKPGBxZXiNHdKBDG4crbg67Hi4xIVBtNHeuyO2f3TwZvo+gcNjdvzm7fG7owQOpj2YcHtI6Az2j9h0NNa6QHnNnREPm6TC/g2rW+lEI1cfV7dK+CRvfIguoRFKx1oXulttaaePEvVJDCXZvXFrN7D+m9a1Z3qL+ATuhauIe1tzKUGXpa+G9We465yIZ2GhbKVckb6smEscIPand3T9M3Jf/i/C5dVXrhY7LFM6FXP6Kx0hkNHewQE7ipMfOvB2yqYBbi/yRwIz09GIIyQTFFa2Ku8ElHctdFPxgnqtMbohha0+Ue0TwtiaHK+FGpf8IQ5y7qvaee2iu2it03Ufr5d+afiiRuWCdYYP7gS+hU8ASy2K8cQ1iTu66je/qKZ1UNruP6HO1rTNH+s/h8KhEVxC3S/iIeLTsKCbpeY0Xbhsvfq6XdQHZDU6XnDH2eunbdxrT9Cuv/FVGv1fP41srGPYe24V6oUIcyMWlVNoLYWJQj4vesaQhuu2QgwAWSHKsrU+kyXmh9Em6ObCdVh86wdKZzfoWqP6apFl/a8C+Nx4EynGUUvBD6uhv6Dn3XX1DTGChcSCEnOyjrdG8aZMooToVS+fAkua8cWOEuN4u6RDL+dbQiNGaKYG1w0VgJLI+m/X/9sCz8leB7IHZrd7IxDqi/571zbfc51NLrHkNOzAm2ku+CdjBJMANXUC+2waevzFAF/6lOHO68r6ANkEH3jdSBy+eeMmFNDUonFPA008YvEqU390Q+GG3o6cXgKJ/j2caS3n3FyjaqMdNPGHSTI5tW4l8Xyg4idRN/R+QEvPI2OpQ7qcMrXpvM3/9pRvRHqabrueFmIJojh74ZmHXHJr6YsfrTeUaTY5P03p/Heu1Eeeus9LNJKv/LFq//zXYGSMnaa0fdnTBZDPIv8D) with the data from [the zip file](https://github.com/CesiumGS/cesium/files/13718858/propertyTextureTransformTest-2023-12-19.zip) from the description in #11708, and confirmed that it renders the bottom right quadrant of the texture:

2024-01-12_TransformFix

I also did a pass through all the metadata-related Sandcastles in the gallery, everything there is still working fine.

So this PR looks close, once you address my feedback above we can merge this.

packages/engine/Specs/Scene/Model/ModelSpec.js
javagl: I'm usually leaning towards some form of "consistency" here (and ... once such a preference is established, would just update everything in one cleanup pass, to prevent others from _also_ following the pattern that they see in the file and of which they don't know that it is not following the latest guidelines). But for now, I only I updated this and the other test.

packages/engine/Specs/Scene/Model/MetadataPipelineStageSpec.js
javagl: This is an "auto-generated" file (i.e. written by VSCode glTF plugin in this exact form), so one could make a case to _not_ apply hand-edits. But it will eventually be maintained manually (and human-readable), so that's probably not a huge deal here.

Specs/Data/Models/glTF-2.0/PropertyTextureWithTextureTransform/glTF/PropertyTextureWithTextureTransform.gltf
javagl: This is an "auto-generated" file (i.e. written by VSCode glTF plugin in this exact form), so one could make a case to not apply hand-edits. But it will eventually be maintained manually (and human-readable), so that's probably not a huge deal here.

javagl:
> I'm going to try out the Sandcastle you posted in the original issue next. Do you have any other Sandcastles you've used while testing or just that one?

I only tested this specific model with the one sandcastle. The intention here was to "zoom in" to what is wrong. (This is dangerous insofar that corner cases or "unusual" cases could be overlooked, but I hope that the scope of this PR is narrow enough for that not to happen...)

javagl:
I think this _could_ be merged. One thing that I'm not sure about: 

In the follow-up PR for fixing the same issue for feature ID textures, it was necessary to pass `incrementallyLoadTextures: false` to one of the spec functions, at https://github.com/CesiumGS/cesium/pull/11766/files#diff-7118f1e2b4539329dcd5314bc1898c6007565fe4e096211621aef6be3d16543dR825 

Here this does not seem to be necessary: https://github.com/CesiumGS/cesium/pull/11709/files#diff-7118f1e2b4539329dcd5314bc1898c6007565fe4e096211621aef6be3d16543dR753

I wonder whether there is a reason for that, or whether this is one of these cases where a "spurious test failure" is about to happen...

Should I just add the `incrementallyLoadTextures: false` here as well, "to be safe(r)"?

packages/engine/Specs/Scene/Model/ModelSpec.js
```diff
@@ -715,6 +719,71 @@ describe(
...
+            } else {
+              material.diffuse = vec3(0.0, 0.0, 0.0);
+            }
+          }
+          `,
+      });
+
+      const gltf = await resource.fetchJson();
+      await loadAndZoomToModelAsync(
+        {
```
ptrgags, line 754: In light of our slack discussion and your comment on the PR, I think you should add `incrementallyLoadTextures: false` in the options here. I would also add a short comment explaining it's because the test relies on the texture values so we need to ensure the texture is loaded.

javagl: Done via https://github.com/CesiumGS/cesium/pull/11709/commits/e6026af4ee023bf46db38b824b7af24bea80bed3

ptrgags:
This looks good now, thanks @javagl for the updates!

Issue #4196, opened 2016-08-12
Make Models be affected by Fog


laurensdijkstra:
glTF Models are currently not affected by Fog. This creates a highly unrealistic image. It would be desirable to have Models be affected by Fog just like the Terrain Primitives are.

ptrgags:
This feature has two parts: fog culling (tiles in the distance for horizon views should use lower-resolution tiles) and fog rendering (Blend the model color with the atmosphere color like terrain does.

Fog culling is already implemented in CesiumJS, but not enabled by default. It goes by the term `dynamicScreenSpaceError`. I find this a bit confusing, and [the documentation](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html?classFilter=Cesium3dtileset#dynamicScreenSpaceError) could be written in a clearer manner.

This is an unfamiliar part of the 3D Tiles code for me, so I explored the code a bit yesterday and now can explain the parameters better:

| Parameter | Effective range of values | Description | 
| --------- | ----------- | --- |
| `dynamicScreenSpaceError` | `{true, false}` | The switch that turns fog culling on/off. |
| `dynamicScreenSpaceErrorFactor` | `>=0.0 px`| Fog culling works by reducing the computed screen space error (SSE) in pixels for tiles far away from the camera for horizon views. This factor is the maximum SSE adjustment. For example, if this is the default of 4.0, then tiles far away from the camera will be `SSE = tile.SSE - 4 px`. Increasing this value is the easiest way to increase the intensity of the fog culling's effects. Setting this to 0.0 has the same effect as turning off fog culling. Furthermore, this should be restricted to positive values, as negative values will cull tiles _near the camera_ |
| `dynamicScreenSpaceErrorDensity` | `>=0.0` | The adjusted SSE falls off from the original value like a bell curve as distance from the camera increases, to simulate fog being more intense in the distance. Increasing this density parameter makes the bell curve have a sharper peak, which simulates thicker fog. Setting this value to 0.0 Has the same effect as turning off fog culling. |
[...]

Issue #11716, opened 2023-12-21
Visualize fog for 3D Tiles/Models


ptrgags:
Splitting up #4196 into smaller, more focused tasks.

While #11715 focuses on the performance aspects of fog, this issue focuses on the visual aspects of fog. 

The main objective is to add a `FogPipelineStage` to the `Model` pipeline that applies fog to models far away from the camera in horizon views. I had started prototyping such an architecture in the branch .

Some details, like using `czm_fog()` are straightforward. However, the main difficulty comes from computing the atmosphere color. Most of the details here are documented in other issues, so see:
 - [`4196-3d-tiles-fog`](https://github.com/CesiumGS/cesium/compare/main...4196-3d-tiles-fog) - a branch where I partially implemented this feature
 - [this checklist of remaining details from #4196](https://github.com/CesiumGS/cesium/issues/4196#issuecomment-1861521328)
 - [this issue](https://github.com/CesiumGS/cesium/issues/11681#issuecomment-1865080064) for a proposed improvement in the atmosphere architecture
 - See `GlobeVS.glsl/GlobeFS.glsl` for the existing fog shader code for terrain.
 
 Note that fog rendering is enabled separately from ground atmosphere, I'll open a different issue for the latter.

ptrgags:
Yesterday I was talking to @jjhembd about the difficulties with projecting a vertex to the ellipsoid surface, and he had some good insights and resources to look at. Recording them here for the future:

### Approach 1: Use a local azimuthal radius of curvature

* Formula can be found in [this PDF](https://www.oc.nps.edu/oc2902w/geodesy/radiigeo.pdf) page 4.
* It's used in [VerticalExaggerationStageVS.glsl](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Shaders/Model/VerticalExaggerationStageVS.glsl)
* This makes use of local curvature at the camera position, using the center of the tile may be more accurate. This might be important since fog applies to tiles far in the distance
* Background concepts about the azimuthal radii of curvature:
    * There's more context on [this Wikipedia page](https://en.wikipedia.org/wiki/Earth_radius#Azimuthal)
    * The "prime vertical radius of curvature" is a fancy name for the principal curvature in the east-west direction. [diagram](https://www.researchgate.net/figure/Geodetic-latitude-ph-the-vector-normal-to-the-surface-of-the-ellipsoid-and-the-radius-of_fig7_289101489)
[...]

PR #11718, opened 2023-12-21
Turn on `dynamicScreenSpaceError` by default

### LINKED ISSUE #6143: `dynamicScreenSpaceErrorDensity ` doesn't get set in the 3D Tiles Inspector
See https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Widgets/Cesium3DTilesInspector/Cesium3DTilesInspectorViewModel.js#L646. I don't think this gets hit when using the corresponding slider in the 3D Tiles Inspector.

### LINKED ISSUE #11677: Cesium3DTileset constructor ignores some dynamicScreenSpaceError options
While setting up performance tests for `dynamicScreenSpaceError` (needed for #4196), I discovered that the parameters I passed into the `Cesium3DTileset` constructor options were doing nothing.

On inspecting the code, the options like `dynamicScreenSpaceErrorFactor` are ignored and just hard-coded as 4. 

The fix is to just make sure the option gets propagated with `defaultValue(option, default)` as usual.

Note that this is distinct from #6143 which is about one of these parameters not propagated correctly in the 3D Tiles Inspector.

### LINKED ISSUE #11715: Enable `tileset.dynamicScreenSpaceError` by default for improved performance
Splitting up https://github.com/CesiumGS/cesium/issues/4196 into smaller, more focused tasks.

From my initial performance tests (see https://github.com/CesiumGS/cesium/issues/4196#issuecomment-1843560992), we decided it makes sense to proceed with turning on dynamic SSE for tilesets, as it's a significant performance improvement (>= 29% improvement in time to first render depending on settings). 

This task involves:

- Do more performance testing to fine-tune the parameters. E.g. try a few values between 
- Update default settings from testing results
- Set dynamicScreenSpaceError to `true` for tilesets by default, and document this in the changelog
- Improve documentation of some of the parameters, see https://github.com/CesiumGS/cesium/issues/4196#issuecomment-1843100677 for ideas

ptrgags:
# Description

This PR:
- Sets `tileset.dynamicScreenSpaceError` to `true`, enabling this optimization for lowering the resolution of tiles far away in horizon views
- Adds missing code in the `Cesium3DTilesetConstructor` to assign initial values based on constructor options
- Fixes the 3D Tiles Inspector SSE Density slider (which wasn't updating the tileset), as this is helpful for testing.
- Improves the documentation for `dynamicScreenSpaceError*` parameters

Example of it in action: [local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=dVTbThsxEP2VUdSHRQre3ewt2wZUBKhFAjVqUZ/ygOMdEguvHdlOQor4985egHB7WWXOzJwzN0cY7TxsJG7RwhFo3MIpOrmu2d8WC2YD0dqnRnsuNdrZYAgPMw3gZY2KkK9wy5XDYYNxLWvupdH7oBOo8cpUOJXiDu2+a84dXvId2veuMITrJcJCmTlCZdCBNp4KxAq8gQaTbqUot3oKd1ILBE9J06XxxiJX0nkpIDmDa6mIgQLUuqIYtJaaafJa/hfdx4NvM8KJ7VzzuUKwqCu0Ui9aYne3m+luWqxtixFy4mvjVku0ZC3Nlsbo7Rpbnj4U7z3RBP1kO7AzkrO2sgvtViio5it5L/VzDSdV9VkvM+3trluEaHfoGxQ9qfMtl/5pjYJSPf4wZqHwNVUvjT5o9ABetbWyspZebtAxXlVBT94EPoLgXiwhoBkae/BSglHIlFkEN+eNA5ThVTO2z3bRU7Im/ctDS/Z40wr0zU+N1L6duuA1Wg68s7peVgrv36yiC2NE2txu0BZWIanq/iT3rvuUW0+/uE6CJgzgcJSXSVEWLMviKI/Sohj2jnRUFnGSsSwuojItRlnvSMZZWkQRK+ibxmmSpw1+0HqNlUjv5Z3sT2yHMpU0wd9GqV48ZXmWRaRbjrI4LstnjcOIjcZ5Mi5T+sbJeJzkvSdmSZbHBVU6yooiiYsED4unAuiKYTAcTJzfKTzuEgC+y3plrIe1VQFjocea3g9NIZyv6fF5JpzrLgFgEu6nTiq5AVkdffBfAEJx58hzu1bqj/yHs8HxJKT4d6n9PfzaoKVX24Qt4+PLDmSMTUIyP870xqg5t2+Y/wM) -- open up the `Update` section and adjust the SSE Density/Factor sliders:

image

## Issue number and link

Fixes #6143 
Fixes #11715 
Fixes #11677

### Choosing better defaults

Tomorrow I plan to focus on selecting better default values for these parameters.

I'm going to start with the performance testing page from https://github.com/CesiumGS/cesium/issues/4196#issuecomment-1843560992

(here modified to use the copy of CesiumJS from `npm build release` and the local Cesium Dev server) - [performance-local.html.txt](https://github.com/CesiumGS/cesium/files/13746276/performance-local.html.txt).

This involves:

1. Make more versions the perf testing page as needed to get a variety of datasets and camera views. I want to also include One World Terrain and Melbourne.
2. I want to do a visual inspection using the 3D Tiles Inspector. How large can I increase the SSE factor/density without the quality loss being obvious? (Note: for Google P3DT, Google Earth serves as a visual benchmark) 
3. Once I narrow down the range of values, performance test a few more cases to verify that the changes give a performance boost everywhere.

# Author checklist

- [ ] Experiment with different SSE Factor and SSE Density settings to get the best performance without a noticeable reduction in quality (See Testing Plan above)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 #### @cesium/engine
 
+##### Additions :tada:
+
+- The `Cesium3DTileset.dynamicScreenSpaceError` optimization is now enabled by default, as this improves performance for street-level horizon views. TODO: PR link
+
 ##### Fixes :wrench:
 
 - Changes the default `RequestScheduler.maximumRequestsPerServer` from 6 to 18. This should improve performance on HTTP/2 servers and above [#11627](https://github.com/CesiumGS/cesium/issues/11627)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 - Changes the default `RequestScheduler.maximumRequestsPerServer` from 6 to 18. This should improve performance on HTTP/2 servers and above [#11627](https://github.com/CesiumGS/cesium/issues/11627)
 - Corrected JSDoc and Typescript definitions that marked optional arguments as required in `ImageryProvider` constructor [#11625](https://github.com/CesiumGS/cesium/issues/11625)
 - The `Quaternion.computeAxis` function created an axis that was `(0,0,0)` for the unit quaternion, and an axis that was `(NaN,NaN,NaN)` for the quaternion `(0,0,0,-1)` (which describes a rotation about 360 degrees). Now, it returns the x-axis `(1,0,0)` in both of these cases.
+- Fixed a bug where the 3D Tiles Inspector's `dynamicScreenSpaceErrorDensity` slider did not update the tileset [#6143](https://github.com/CesiumGS/cesium/issues/6143)
 
 ### 1.112 - 2023-12-01
 
--- a/packages/engine/Source/Scene/Cesium3DTileset.js
+++ b/packages/engine/Source/Scene/Cesium3DTileset.js
@@ -17,6 +17,7 @@ import IonResource from "../Core/IonResource.js";
 import JulianDate from "../Core/JulianDate.js";
 import ManagedArray from "../Core/ManagedArray.js";
 import CesiumMath from "../Core/Math.js";
+import Matrix3 from "../Core/Matrix3.js";
 import Matrix4 from "../Core/Matrix4.js";
 import Resource from "../Core/Resource.js";
 import RuntimeError from "../Core/RuntimeError.js";
@@ -437,7 +438,10 @@ function Cesium3DTileset(options) {
    * @type {number}
    * @default 0.00278
    */
-  this.dynamicScreenSpaceErrorDensity = 0.00278;
+  this.dynamicScreenSpaceErrorDensity = defaultValue(
+    options.dynamicScreenSpaceErrorDensity,
+    0.00278
+  );
 
   /**
    * The maximum screen space error (SSE) reduction when {@link Cesium3DTileset#dynamicScreenSpaceError} is used. The
@@ -457,7 +461,10 @@ function Cesium3DTileset(options) {
    * @type {number}
    * @default 4.0
    */
-  this.dynamicScreenSpaceErrorFactor = 4.0;
+  this.dynamicScreenSpaceErrorFactor = defaultValue(
+    options.dynamicScreenSpaceErrorFactor,
+    4.0
+  );
 
   /**
    * A ratio of the tileset's height above which the effects of {@link Cesium3DTileset#dynamicScreenSpaceError} begin
@@ -470,7 +477,10 @@ function Cesium3DTileset(options) {
    * @type {number}
    * @default 0.25
    */
-  this.dynamicScreenSpaceErrorHeightFalloff = 0.25;
+  this.dynamicScreenSpaceErrorHeightFalloff = defaultValue(
+    options.dynamicScreenSpaceErrorHeightFalloff,
+    0.25
+  );
 
   // Updated based on the camera position and direction
   this._dynamicScreenSpaceErrorComputedDensity = 0.0;
@@ -2282,6 +2292,7 @@ const scratchMatrix = new Matrix4();
 const scratchCenter = new Cartesian3();
 const scratchPosition = new Cartesian3();
 const scratchDirection = new Cartesian3();
+const scratchHalfHeight = new Cartesian3();
 
 /**
  * @private
@@ -2346,10 +2357,15 @@ function updateDynamicScreenSpaceError(tileset, frameState) {
       direction = Cartesian3.normalize(direction, direction);
       height = positionLocal.z;
       if (tileBoundingVolume instanceof TileOrientedBoundingBox) {
-        // Assuming z-up, the last component stores the half-height of the box
-        const boxHeight = root._header.boundingVolume.box[11];
-        minimumHeight = centerLocal.z - boxHeight;
-        maximumHeight = centerLocal.z + boxHeight;
+        // Assuming z-up, the last column
+        const halfHeightVector = Matrix3.getColumn(
+          boundingVolume.halfAxes,
+          2,
+          scratchHalfHeight
+        );
+        const halfHeight = Cartesian3.magnitude(halfHeightVector);
+        minimumHeight = centerLocal.z - halfHeight;
+        maximumHeight = centerLocal.z + halfHeight;
       } else if (tileBoundingVolume instanceof TileBoundingSphere) {
         const radius = boundingVolume.radius;
         minimumHeight = centerLocal.z - radius;
--- a/packages/engine/Source/Scene/Cesium3DTileset.js
+++ b/packages/engine/Source/Scene/Cesium3DTileset.js
@@ -75,10 +75,10 @@ import Cesium3DTilesetSkipTraversal from "./Cesium3DTilesetSkipTraversal.js";
  * @property {boolean} [preloadWhenHidden=false] Preload tiles when <code>tileset.show</code> is <code>false</code>. Loads tiles as if the tileset is visible but does not render them.
  * @property {boolean} [preloadFlightDestinations=true] Optimization option. Preload tiles at the camera's flight destination while the camera is in flight.
  * @property {boolean} [preferLeaves=false] Optimization option. Prefer loading of leaves first.
- * @property {boolean} [dynamicScreenSpaceError=false] Optimization option. Reduce the screen space error for tiles that are further away from the camera.
- * @property {number} [dynamicScreenSpaceErrorDensity=0.00278] Density used to adjust the dynamic screen space error, similar to fog density.
- * @property {number} [dynamicScreenSpaceErrorFactor=4.0] A factor used to increase the computed dynamic screen space error.
- * @property {number} [dynamicScreenSpaceErrorHeightFalloff=0.25] A ratio of the tileset's height at which the density starts to falloff.
+ * @property {boolean} [dynamicScreenSpaceError=true] Optimization option. Reduce the screen space error for tiles that are further away from the camera so that lower resolution tiles are used.
+ * @property {number} [dynamicScreenSpaceErrorDensity=0.00278] Similar to fog density, this value determines how far away from the camera that screen space error falls off. Larger values will cause tiles closer to the camera to be affected by dynamicScreenSpaceError.
+ * @property {number} [dynamicScreenSpaceErrorFactor=4.0] The maximum screen space error adjustment in px to apply to a tile for tiles far away on the horizon. Increase this value to make the SSE adjustment stronger, which in turn will select lower LOD tiles near the horizon.
+ * @property {number} [dynamicScreenSpaceErrorHeightFalloff=0.25] A ratio of the tileset's height above which the density decreases for dynamic screen space error. This way, dynamicScreenSpaceError has the strongest effect when the camera is close to street level.
  * @property {number} [progressiveResolutionHeightFraction=0.3] Optimization option. If between (0.0, 0.5], tiles at or above the screen space error for the reduced screen resolution of <code>progressiveResolutionHeightFraction*screenHeight</code> will be prioritized first. This can help get a quick layer of tiles down while full resolution tiles continue to load.
  * @property {boolean} [foveatedScreenSpaceError=true] Optimization option. Prioritize loading tiles in the center of the screen by temporarily raising the screen space error for tiles around the edge of the screen. Screen space error returns to normal once all the tiles in the center of the screen as determined by the {@link Cesium3DTileset#foveatedConeSize} are loaded.
  * @property {number} [foveatedConeSize=0.1] Optimization option. Used when {@link Cesium3DTileset#foveatedScreenSpaceError} is true to control the cone size that determines which tiles are deferred. Tiles that are inside this cone are loaded immediately. Tiles outside the cone are potentially deferred based on how far outside the cone they are and their screen space error. This is controlled by {@link Cesium3DTileset#foveatedInterpolationCallback} and {@link Cesium3DTileset#foveatedMinimumScreenSpaceErrorRelaxation}. Setting this to 0.0 means the cone will be the line formed by the camera position and its view direction. Setting this to 1.0 means the cone encompasses the entire field of view of the camera, disabling the effect.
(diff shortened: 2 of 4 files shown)


packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -75,10 +76,10 @@ import Cesium3DTilesetSkipTraversal from "./Cesium3DTilesetSkipTraversal.js";
...
- * @property {number} [dynamicScreenSpaceErrorDensity=0.00278] Density used to adjust the dynamic screen space error, similar to fog density.
- * @property {number} [dynamicScreenSpaceErrorFactor=4.0] A factor used to increase the computed dynamic screen space error.
- * @property {number} [dynamicScreenSpaceErrorHeightFalloff=0.25] A ratio of the tileset's height at which the density starts to falloff.
+ * @property {boolean} [dynamicScreenSpaceError=true] Optimization option. Reduce the screen space error for tiles that are further away from the camera so that lower resolution tiles are used.
+ * @property {number} [dynamicScreenSpaceErrorDensity=0.00278] Similar to fog density, this value determines how far away from the camera that screen space error falls off. Larger values will cause tiles closer to the camera to be affected by dynamicScreenSpaceError.
+ * @property {number} [dynamicScreenSpaceErrorFactor=4.0] The maximum screen space error adjustment in px to apply to a tile for tiles far away on the horizon. Increase this value to make the SSE adjustment stronger, which in turn will select lower LOD tiles near the horizon.
```
ptrgags, line 81: Once I determine good defaults tomorrow, make sure to update every place where they are listed in this file, which includes documentation comments and examples.

ptrgags (commented):
Self review, had a few comments about documentation

ptrgags:
In setting up Maxar's One World Terrain base globe, I'm realizing that dynamic screen space error doesn't really impact this dataset much. Even the smallest tiles are relatively large, so to get a good horizon view you have to position the camera very far above the dataset (where the dynamicSSE has essentially no effect). So I'll try other datasets.

image

ptrgags:
Today I've been experimenting with the `dynamicScreenSpaceErrorFactor` and `dynamicScreenSpaceErrorDensity`.

First I kept the density the same and adjusted the SSE Factor. 8 px seemed to be reasonable for most views I tried for the default density.

However, then I looked at the density value and realized that right now the fog is very close to the camera! And compared with `scene.fog.density`, it is a whole order of magnitude larger (closer to camera)

One conceptual thing I learned today. If you compare our fog function with a normal distribution, equate the exponents and solve for the standard deviation (i.e. distance to inflection point, you get: $$\sigma = \frac{1}{2\rho}$$ (where sigma is the stddev and rho is the density). This is handy for estimating how far away the fog's inflection point is from the camera.

So the existing dynamicSSEDensity default of 0.00278 gives a fog distance of around 180m, that's pretty close to the camera. 

Meanwhile `scene.fog.density` of 2e-4 = 0.0002 gives a fog distance of 2.5 km. This seems more in the ballpark of a horizon view in reality (though it depends on elevation)

So I think the thing to do will be set the density default to match terrain (to avoid unnecessary loss in visual quality), and then adjust the SSE factor (maybe it can be made a bit higher than 8 to compensate performance wise?)

CHANGES.md
```diff
@@ -7,12 +7,14 @@
 ##### Additions :tada:
 
 - Vertical exaggeration can now be applied to a `Cesium3DTileset`. Exaggeration of `Terrain` and `Cesium3DTileset` can be controlled simultaneously via the new `Scene` properties `Scene.verticalExaggeration` and `Scene.verticalExaggerationRelativeHeight`. [#11655](https://github.com/CesiumGS/cesium/pull/11655)
+- The `Cesium3DTileset.dynamicScreenSpaceError` optimization is now enabled by default, as this improves performance for street-level horizon views. [#11718](https://github.com/CesiumGS/cesium/pull/11718)
```
ggetz, line 10: 1. This will be a breaking change, so it should be documented in the breaking change section.
2. Let's move this to next month's release given today's date.

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -75,10 +76,10 @@ import Cesium3DTilesetSkipTraversal from "./Cesium3DTilesetSkipTraversal.js";
  * @property {boolean} [preloadWhenHidden=false] Preload tiles when <code>tileset.show</code> is <code>false</code>. Loads tiles as if the tileset is visible but does not render them.
  * @property {boolean} [preloadFlightDestinations=true] Optimization option. Preload tiles at the camera's flight destination while the camera is in flight.
  * @property {boolean} [preferLeaves=false] Optimization option. Prefer loading of leaves first.
- * @property {boolean} [dynamicScreenSpaceError=false] Optimization option. Reduce the screen space error for tiles that are further away from the camera.
- * @property {number} [dynamicScreenSpaceErrorDensity=0.00278] Density used to adjust the dynamic screen space error, similar to fog density.
- * @property {number} [dynamicScreenSpaceErrorFactor=4.0] A factor used to increase the computed dynamic screen space error.
- * @property {number} [dynamicScreenSpaceErrorHeightFalloff=0.25] A ratio of the tileset's height at which the density starts to falloff.
+ * @property {boolean} [dynamicScreenSpaceError=true] Optimization option. Reduce the screen space error for tiles that are further away from the camera so that lower resolution tiles are used.
```
ggetz, line 79: This is knit-picky, but the user will likely care a about the effect (less data downloaded, faster stable view) than the mechanism (reducing SSE). Can we re-order the statements to reflect the user's priorities better?

Same comment for the related properties below.

ggetz: For reference, https://docs.unrealengine.com/4.27/en-US/BuildingWorlds/FogEffects/HeightFog/ does a good job of explaining some factors.

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -361,18 +362,18 @@ function Cesium3DTileset(options) {
...
-   * The algorithm is biased towards "street views" where the camera is close to the ground plane of the tileset and looking
-   * at the horizon. In addition results are more accurate for tightly fitting bounding volumes like box and region.
+   * Optimization option. Whether the tileset should adjust the screen space error for tiles on the horizon. Tiles that
+   * are far away will be adjusted to render at lower detail than tiles closer at the camera. This improves performance
+   * by loading and rendering fewer tiles, but may result in a slight drop in visual quality in the distance. This
+   * optimization is strongest when the camera is close to the ground plane of the tileset and looking at the horizon.
+   * Furthermore, the results are more accurate for tightly fitting bounding volumes like box and region.
    *
    * @type {boolean}
    * @default false
```
ggetz, line 372:
```suggestion
   * @default true
```

packages/engine/Specs/Scene/PickingSpec.js
```diff
@@ -134,6 +134,10 @@ describe(
     function createTileset(url) {
       const options = {
         maximumScreenSpaceError: 0,
+        // The camera is zoomed pretty far out for these tests, so
```
ggetz, line 137: Hmm, I wouldn't expect this to be needed. Why is this needed for 3D Tiles but not terrain?

ptrgags:
This morning I set `dynamicScreenSpaceErrorDensity` to `2e-4` like terrain, and then explored the SSE Factor setting for this configuration. 

Considering that the "fog" is now a lot further into the distance, you can dial up the SSE Factor more without huge changes in visual appearance. For P3DT and Melbourne, 24 px seemed to be the highest I could go before it became noticeable, at least to my eyes (visual quality is subjective). The difference was most noticeable for the Tokyo aerial view, so here are some screenshots for comparison:

| No dynamic SSE | density 2e-4, SSE factor 16 | density 2e-4, SSE factor 24 |
| ---|---|--- |
| image | image | image |

I also ran the performance tests again:

Configuration | Average time to first load (s) | % difference from no fog
-- | -- | --
No Fog | 7.989 | N/A
Fog with old defaults (density 0.00276, SSE Factor 4) | 5.678 | -28.93%
Fog with density 2e-4, SSE Factor 16 | 5.216 | -34.71%
Fog with density 2e-4, SSE Factor 24 | 4.888 | -38.82%

Raw data: 
[Performance Stats CSV](https://github.com/CesiumGS/cesium/files/13811540/Dynamic.SSE.Perf.Testing.-.P3DT.-.New.Settings.csv)

Snapshot of testing HTML files: (these require a local release build of CesiumJS)

[performance-melbourne.html.txt](https://github.com/CesiumGS/cesium/files/13811554/performance-melbourne.html.txt)
[performance-p3dt.html.txt](https://github.com/CesiumGS/cesium/files/13811555/performance-p3dt.html.txt)

So even though I had to reduce the density parameter, it's still better performance than the old defaults!

@ggetz what do you think, is 24 a reasonable default, or would it be better to pick a more conservative value (say 16 or less) for now until #11716 is implemented?

One other detail to put things in perspective: I notice that for terrain, the SSE Factor is equal to the maxSSE for terrain (2, compared with 3D Tiles which uses 16 as the default). This would mean 16 is roughly analogous. 24 is a bit more intense than that in comparison

CHANGES.md
ptrgags: I moved the entry. I also added a couple entries for the changed defaults (although one value is still pending on our discussion). If that's too much detail, I could combine the entries.

ptrgags:
I talked with @ggetz, a summary:

- I should check my unit tests for race conditions, for some reason specs are failing on the command-line (both in CI and locally)
- the SSE Factor value of 24 seems good
- but to be extra sure, get a comparison screenshot between 3D Tiles P3DT and Google Earth to make sure it looks reasonable
- Add unit tests for checking the default parameter values
- For the changelog:
    - condense the entries about the defaults into one entry.
    - The changes about the inspector go in the widgets section, not the engine section

ptrgags:
@ggetz Here's a visual comparison of CesiumJS and Google Earth. Again I used Tokyo as that had the biggest visual difference compared 

CesiumJS (density 0.0002, SSE Factor 24)
image

Google Earth:

image

I find it hard to spot specific differences, as they're small and lost in the high-frequency details of the image. If anything, Google Earth looks a bit lower resolution in the distance.

ptrgags:
Strange, the picking test timing seems to be sensitive to the dynamic SSE feature, but I'm not sure why.

I'm looking at the test "Scene/Pick > pickFromRay > picks a tileset" (though there are other similar tests also affected).

The main difference I notice is the timing of loading the tile and content. For some reason `tileset.tilesLoaded` is `true` before the content is ready, and this leads to doing the picking call too early for the test to succeed. I'm not sure why this only seems to happen when `dynamicScreenSpaceError` is on.

| Dynamic SSE off (test works) | Dynamic SSE on (test fails) |
|---|---|
| image | image | 

Some other observations:

- When the picking happens, the camera is _above_ the tileset, so dynamic SEE is computed as 0 (not a horizon shot)
- however, the root tile has a transform that puts it somewhere in the US. This transform is replaced with a new ENU transform that puts it at cartographic(0, 0), but for the first few frames technically dynamic SSE is applied and acts like an almost-horizon view. So that tile gets culled until the new matrix is applied. I wonder if this is related.
- the picking pass itself isn't affected by dynamic SSE, as picking uses an ortho frustum. the problem is that picking happens too early.

[Sandcastle to show the test setup](http://localhost:8080/Apps/Sandcastle/index.html#c=bVNdbxoxEPwrFk8gUV+iti+ERC2QSJHaVAokVSVeHHvh3Prs03oPSKv8967PfIVESD5sz8zNzu7p4COJlYU1oLgUHtZiDNE2lXxsz7rzjm734+BJWQ847/Qu5j7/dMt2CpdwD5qUXzpgla3C/kguMFQTWCJA7H44l2d9kdfdkhWzGoI51srWJHiyZCFKZUz339wL4VUFAzHv3IMRhMpH12hGMX9Lnnf6CbffD0TLE0KHgMZ6RRAHJ+b7GVHxHVrlBrtSxsEFlPfXE7m2VH51dam6Z/Jzr8W/8PpyXIJma6gO5iMbA5lPGZb/yAiUMs7VGIiULNngTz2diJN1wNRT9RptxQmtthGptbK0d98+Pk5mmdq24wEd97aY1qBjMVGkileoYqRIl2B2z59c9iylvAhYjcKm2NqQv2PwnHTOtigMPDXLaRnWo9B4Dnn5GFyTGkXY5HQrtbFVU001T4Of1krDNWLAgThrr80zN9bqt9dJIEXxOmiksERVl1a/N3dpIHiGX+eZBA50pih/4I6PFCWF8Q7RPX5VUtjVjyGQpF0yB6HvitBuPsnbyfXd7Hb264hSBQMu3x/w+3CjBBXpLiCVD/Us3NgNmBvkkenu7fYuOv3OMNKzg6s8sEJ8sVXNHNFwX6UsCKrapQkvnhr9h1+qY0y2E3RYHFOHxq6ENZfvfOhCOxUj3ywa56b2L39SV8OC8W+oLqjU7B8rQKeeE6w8v/qWD6WUw4K37zMpBPek8ET5Pw)

packages/engine/Specs/Scene/PickingSpec.js
ptrgags: the problem seemed to be a race condition when waiting for the tileset to load in this particular test setup. I opened https://github.com/CesiumGS/cesium/issues/11732 for this since it only seems to affect the unit tests, not the rendering itself.

ptrgags:
@ggetz this is ready for review again!

packages/widgets/Source/Cesium3DTilesInspector/Cesium3DTilesInspectorViewModel.js
```diff
@@ -722,25 +722,29 @@ function Cesium3DTilesInspectorViewModel(scene, performanceContainer) {
...
+   * @default 2.0e-4
    */
   this.dynamicScreenSpaceErrorDensitySliderValue = undefined;
   knockout.defineProperty(this, "dynamicScreenSpaceErrorDensitySliderValue", {
     get: function () {
       return Math.pow(dynamicScreenSpaceErrorDensity(), 1 / 6);
     },
     set: function (value) {
-      dynamicScreenSpaceErrorDensity(Math.pow(value, 6));
+      const scaledValue = Math.pow(value, 6);
```
ggetz, line 743: Should this also have a spec to ensure this doesn't break in the future?

ggetz (commented):
Thanks @ptrgags! This is close!

ptrgags:
@ggetz I addressed your feedback!

ggetz:
Awesome, thanks @ptrgags!

PR #3675, opened 2016-03-07
fix #3674

anne-gropler:
camera.flyTo() -> createHeightfunction.
if optionAltitude was given, maxHeight was undefined and therefore no appropriate height function was calculated.

PR #11741, opened 2024-01-05
Fix postUpdate spec

anne-gropler:
# Description

I noticed that there are 2 Scene specs that do the same, but have different descriptions. I guess this is just a copy paste error, that's why `postUpdate` was missing. I updated that test, and fixed a typo in the other test.

## Issue number and link

Let me know if you need an issue for that, I didn't create one because it's neither a feature nor a bug, just fixing a test that I presume not to be doing what it should do.

## Testing plan

I ran the tests using `npm run test` and it works.

# Author checklist

- [x] I have submitted a Contributor License Agreement **yes, see https://github.com/CesiumGS/cesium/pull/3675#issuecomment-193306141**
- ~~[ ] I have updated `CHANGES.md` with a short summary of my change~~ **no, because it did not change the Cesium API**
- ~~[ ] I have update the inline documentation, and included code examples where relevant~~


ggetz:
Great, thanks for catching this @anne-gropler! The fix works perfectly.

Issue #11681, opened 2023-12-07
Proposal: Consolidate SkyAtmosphere and Globe atmosphere settings?


ptrgags:
As I'm looking into #4196, one thing I notice is that the following settings exist on both `viewer.scene.skyAtmosphere` (for the sky) and `viewer.scene.globe` (for ground and fog for terrain):

* `atmosphereBrightnessShift` / `brightnessShift`
* `atmosphereHueShift` / `hueShift`
* `atmosphereLightIntensity`
* `atmosphereMieAnisotropy`
* `atmosphereMieCoefficient`
* `atmosphereMieScaleHeight`
* `atmosphereRayleighCoefficient`
* `atmosphereRayleighScaleHeight`
* `atmosphereSaturationShift` / `saturationShift`

Ultimately these impact the shader uniforms in `AtmosphereCommon.glsl`:

```
uniform vec3 u_radiiAndDynamicAtmosphereColor;
uniform float u_atmosphereLightIntensity;
uniform float u_atmosphereRayleighScaleHeight;
uniform float u_atmosphereMieScaleHeight;
uniform float u_atmosphereMieAnisotropy;
uniform vec3 u_atmosphereRayleighCoefficient;
uniform vec3 u_atmosphereMieCoefficient;
```

Since the ground and sky atmosphere both represent the Earth's atmosphere, conceptually they should always use the same settings.

There's also two other benefits of combining them:

1. This streamlines the interface. Instead of changing settings in two places (skyAtmosphere and globe), you only need to change one (I'm thinking `Globe`, but open to suggestions).
2. It also would streamline things deeper in the implementation details. In #4196 I'm already considering changing the Globe's implementation to use `AutomaticUniforms`, this way it's easier to implement fog for 3D Tiles.

I'm curious if the community would find this to be a reasonable breaking change, or if there are use cases where configuring `SkyAtmosphere` and `GroundAtmosphere` differently is helpful. I'll ask about this on the forum as well.

lilleyse:
> I'm thinking `Globe`, but open to suggestions

I think the atmosphere options should be in its own object under `Scene` in case you don't have a globe

ptrgags:
@lilleyse That's a good point, I like that for 2 more reasons:

1. It puts the parameters in their own dedicated struct to make 
2. The implementation would have more symmetry with how `scene.fog` is implemented

ptrgags:
[...]

Issue #11717, opened 2023-12-21
Visualize ground atmosphere for 3D TIles/Models


ptrgags:
Splitting up #4196 into smaller tasks

This task is closely related to #11716, but it focuses on ground atmosphere rather than fog.

In `GlobeFS.glsl`, the globe shader handles both fog and ground atmosphere. In the spirit of `Model`'s pipeline stages, I envision this being implemented as a `GroundAtmospherePipelineStage` that's separate from `FogPipelineStage`

Also, see https://github.com/CesiumGS/cesium/issues/11681#issuecomment-1865080064 for the proposed `scene.groundAtmosphere` object that will make it easier to propagate settings via the FrameState and/or automatic uniforms.

One detail to note: Some details like computing the atmosphere color would be expensive to duplicate if both fog and ground atmosphere are needed for the same model, so it's best to avoid that.

ptrgags:
One note from #11744: right now the `FogStageFS` (and eventually `FogStageVS`) calls the atmosphere functions. To avoid computing the color twice, it may make sense to move some calls up a level and pass the colors into the scattering function.

Another question is should ground atmosphere be a separate pipeline stage at all? it mainly adds a bit of shader code and at most a couple more uniforms. 

Maybe renaming the pipeline stage to `AtmosphereEffectsPipelineStage` would be sufficient here.

ptrgags:
After exploring the globe shader a bit, ground atmosphere and daynight shading are starting to make more sense. The fading logic is a bit complex, so here's a quick summary from `GlobeFS`

* Start with the raw color of the globe from imagery, globe materials, etc. (or in the case of 3D Tiles, it'll just be the computed color at that point in the shader)
* Compute a Lambert diffuse term for the globe. Note that while this dims the color of the globe, it's only medium-dark, night-time shading happens later.
* Depending on the camera height in relation to `lightFade(Out|In)Distance`, blend between unlit and diffuse lighting.
* The atmosphere scattering function computes:
  * a ground atmosphere color. Note that this is bright near the sun, and near black on the night-time side of the globe.
  * a transmittance value. This represents how the circumference of the earth looks brighter since light has to travel through more atmosphere there. 
* Use these atmosphere values to brighten up the globe around the edges (`color = diffuse + groundAtmosphere * transmittance`)
[...]

PR #11744, opened 2024-01-05
11716 visualize fog

### LINKED ISSUE #11716: Visualize fog for 3D Tiles/Models

ptrgags:
# Description

While #11718 enabled and tuned the performance aspects of fog for 3D Tiles, this PR is about rendering the fog for 3D Tiles. This includes atmosphere color and dynamic lighting to make more realistic looking fog that accounts for [Rayleigh and Mie scattering](https://www.scratchapixel.com/lessons/procedural-generation-virtual-worlds/simulating-sky/simulating-colors-of-the-sky.html). The goal is to make world-scale 3D Tiles tilesets like Google Photorealistic 3D Tiles look more realistic like we do for terrain.

image
image

[Local Sandcastle Example](http://localhost:8080/Apps/Sandcastle/index.html#c=lVRbT9swFP4rR2gPQSpO0yRNOwoag90ktqHR7akPc+NDa+HYle3Cuon/vmMnpTCGpr2kPZfvfOfq2mjn4UbiLVo4Ao23cIpOrhv2LeqS2V4d5VOjPZca7WyvB79mGiBNvWxQke4lXHHlsNdquZYN99Loh2pXo8aPRuCFrK/RPjTNucNzvkH71JSmMF0iLJSZIwiDDrTxlCQK8AaCTrqVIqzYujupawRPoIul8cYiV9J5WUN+BlOpKAI5qLUgH7SWCgq4GH/He7d/OCM9RXuj+VwhWNQCrdSLGNhdb2a6jn2LVVHb2v6xKBI2/jLyO/GNcaslWpKW5pY8vV3vPPjOLDaaN7I+l4ulD0RH2zGctYZdpK3LdLNCdvn10/mHd++nMeGY8okQz5U+095u2tm16fugRU9k/JZLv6WsCerxnTELhY9D5WfTFpKEFsGjutnKykZ6eYOOcSGSLnhwvIOa+3oJCbXc2P1dCkYhU2aRfH8TDKAMF6H450bXhWQB/uJXDHb3PRJ0xV8YqX0cUs0btBx4K7W1rBT+2La+tTOKFvY8iRkJJDrdre6DSzjl1tM/rvMkuAEcDIbjvBpXrCyz/rBfVFWvMxSDcZXlJSuzqj8uqkHZGfJRWVT9PqvoW2RFPiyCfj9ajZVIt/WE9j3GblxIat0Xo1RHXrBhWfaJdzwos2w8vuc46LPBaJiPxgV9s3w0yoedJWN5OcwqynRQVlWeVTkeVNsEaNshtO+Sa1Fz52kkNL6pWVDHXq+9N5regLdm0R2DCPcftrgHSb1EulixD0fH7Uwf7cOVWTBsMbRhne/h/Xn9k+9LvLqA/y9Kew97yrrX25s4v1F43DYG4JVsVsZ6WFuVMJZ6bOg9oWmn8zUBPauda1cdYJI+hE6EvAEpjv7yPkKtuHNkuVordSl/Uv7Hk5T8n0C7hf98g5ZeseC2zI7PWyVjbJKS+HekN0bNuf0j8m8)

To do this, it involves quite a few moving parts, as `Model` does not have access to the Scene, and the model isn't always added to the scene anyway.

* Created an `Atmosphere` class (`scene.atmosphere`) that will hold the atmosphere related settings (see #11681 for the longer-term plan which involves breaking changes to the API)
* The atmosphere updates the `FrameState` each frame (this is modeled after what `Fog` does)
* The `UniformState` adds some new uniforms (`czm_atmosphere*`) whose values come from the FrameState.
* A new Model pipeline stage, `FogPipelineStage` adds the shader code for enabling fog
All this together, 

## Issue number and link

Fixes #11716 

## Testing plan

For testing, I've been adjusting 

Here's a roundup of the Sandcastles I used for testing:

* [Comparison with the Globe](http://localhost:8080/Apps/Sandcastle/index.html#c=rVbrb9s2EP9XuGCAZcClSUqiJDfJliVp2iFrg8XdgMEfSkuMTYQWDZJK6hb533d6+JFHiwRIPjjive93xzvmpnQe3Sh5Ky06QKW8RcfSqWqB/2lowWQvb87HpvRCldJO9gbo+6REaDj0aiE10EboSmgnBy1VlGohvDLlLtnlspR/mUJeqPxa2l3WVDh5LlbSPmZ5aS04Ha1jGrdnfGXN4l9jddERgj6I3/XfTspJORyio6JAF3PjjZVCK+dVjsITNFZaukmZNxn7+iA9pCxuhfJrBzloeHlmzEzL+xbCk3GrEtRuWsBwkxVeWrVQXt1Ih0VRBJ3pTTSX4KZaIlWCjNAtEg8suOvVkV8Yt5xLC6e5uYXAGhge+royMyxLMdWy+JGI2FoqVqVYqPxczeZelTPQ6PI8aRlbp2uR8Wop8cdPH08fWp1pM30UWpNf3SgAt9J6hYRDfi6Rq0qkHJqZ2mlhbssHxnKxkFZgQKlWDpp2KiTAXHaNs9OHx8J6+BJlGFAWE8ZjnISUc8KTcIDeREkYUkYw4WlMKQmzAYpIEqc8wSljHCiU95tmMlZJ6OFHDt5LUUCcF8rn87+N1gHDJA6TKE6jhEUR5QzcEEyzmKYpi+KIADHkAwRO4Y9mISOchlmWRjGLw00rdhnn2uTXOK+sBe9juDHbKvxZacjrBDqu6egPzqScULhyjLDoDaFvSDambMT4iBDMWRLRhFLGspCmSfzfZK9/rwQ18HO4elqsoAxFZWvwa2IhVq+DfxqmWRhjyJ5RTpMMgImjhBLCcRKHUZKQLBmgkCYAO8cAPONpxtiL4Y8YJySOMsIyTqK4hR9cUwAgTBmJk5RGA8QxSxkAEkU8pBmJ6StDT9gojEYhuAmjlKUp5EdIGnLON9jfCItyAf8cmF57bc7Abj9qlI+8t2paeRn0vJiqspBfewPUI73+WxiY6KLy6MrklUOmbCrWam4smDLXMBvri1eVeY1g0G9ncCeJG+1mMt01ca3lgL/0lZXHTcFhEtV33AVtpJ2NdiC2LbGbRX1++1DgGEY4CH15jXb69XunuDRO1YL4690APaaunqR+u3thX21MzFvOPau11C7BgsZd109fNigYLbE2s2CLRQN5jXgBFVhAIPUKOL2Bj3NYGwCODXrXclXPQKj4piyBXBdQXcEBgwg6ODhAvby3ZqCf166JCd3Vv0/Z8bt2diL/0c3Y2uvW6KUoi1w4D3qQ0djMYCP+UXkPvTfZu6zXwFm9EerXQLuvUZDPJSzwoo8ODlvfP14gnWjjtNuXa9YvW95zQnlnZui0XYjPj+X+Fn2Zw8vrFdruzZf7fnLXvyyEz06iboOj9eJ+fgA/fR90qui3570ULj9/PP9w9n6MRi96WTzZGm1BHsfSgrI32Nt3fqXlYdfV6He1WBoLbyurA4yHXi6WGua5G04rUPM4d67rarQ/3FXdL9QNUsXBE09blGvhHHCuKq0v1Tfo78P9Icg/UtWmGSKfbqSFhVuLzenheUvEGO8P4fi0pjdGT4V9YPl/)
* [Cycle through dynamic lighting options](http://localhost:8080/Apps/Sandcastle/index.html#c=lVVtb9s4DP4rRHAHOFgq27EdJ9ekWK9N7wZ0XbG2+zIPm2KziTBZCmQlXTb0v4+ynTbpCw73xZFIPnwo6qGSa1VZWAu8QwMTUHgHJ1iJVck+1TYv6+T1/kQry4VCk3V68CtTAL5vRYmSbH/BLZcV9horV6LkVmi1a65yVPheF3gp8u9odl0zXuE536B57vJ9uF4gzKWeIRQaK1DaUpFYgNXgbKJaSsIW2/BKqBzBEuhyoa02yKWorMghOoVrISkDBchVQTFoDB3I4er8j7z33cOM7JRtqvhMIhhUBRqh5nXi6vsmU03HWH0sRpZjW+pquUBDu4W+o1Zas8JtnuOieK2eTFmzaRqa13dhnRUtZeB3XNjtdeQEtfiP1nOJ+6mi0+sG4rm6AfZKWxpRCivWWDFeFF6b3AXeQ85tvgCP+qBN97EELZFJPfe+TZ0DpOaFO/pr/WxTMgf/41ed7P5bTdAe/lILZevO5bxEw4E3u+YsS4k/nrSzCWOU1GnQqwsrkFhVK6sdlZ5wY2nFVeS5MICD/mAUpaOUJUkYDII4TXutI+6P0jBKWBKmwShO+0nriIZJnAYBS+kbh3E0iJ29W3u1EUi6f0b7L9ZNuRTUwY9aypY8ZoMkCYh31E/CcDR64DgIWH84iIajmL5hNBxGg9YTsigZhClV2k/SNArTCA/SbQGkRHBdbJQhxXxh94f0VBjMXXVcnjtv26yt9bVWhT0Iuq3Q91q/pah/a/Vm6oqrIueVJVmQhK71nG7t75W1WtHjcKbn7ZQU7mFwou+Bly+QRrnowuSo0dUeya2eM2wwRNXGHj7M3X/yfazH0eH/F6V5gL3EmilJQ0dxU7UqKSAgKynwnaJ3Ys2l5z1mbm5jenHz/uun4/Ob6RWFf846H87OXD1Z5+pkejFtlzcXWefL4dPR2sF+bim/NMP7rHL++LAUG8VLkdcX7SZysi23RrbrN292d/DnBCJ3yB6EQRB0Dzu9zriyG4lHjfwA3opyqY2FlZEeY77Fkl5UEoo/W1GHLMurqikNYOzvQseFWIMoJi/8Q0AueVWR53Yl5ZX4SRd1NPYp/hm0fV0+rNHQO+7CFuHReWNkjI192r6MtFrLGTdPMv8G)
* [Shading models in the sky](http://localhost:8080/Apps/Sandcastle/index.html#c=tVdtb9s2EP4rnD8pg0fJTtJunhssibcmWNJ1tVMMmIeClk4WEYoUSCqJV+S/90hJtqx4H4zMRgCTx+Pd89yLeYmVNJY8cHgETd4RCY/kEgwvc/rZy4J5L/b7SyUt4xL0vNcnX+eSEC5TdaGeRiRlwkDfiQwIiC1X8lomPGZW6a3TjCXq0YyI1WUjUaVIziXPmYW1/Pno57mcywoUNTFIoKla0gSk4XaFMAc0IiEJhvj1PRlEUeRubOkzmytTZKCBJivJch7f8GVmuVzi9ZrhpDo4X6s2KrNVAXR69+Hm+v3VrGt5KdQCKEi2ENCy6bB72GkpfQhIrAFZ3aoERFBq0ScZOPWjKnq1TZCWWw6GasjVA5wLEVTsCYl9agqFnJ25NexLpi2umDymqVb5BJYawATuCiE/DIbHNHp7cvJm8FO/Ep2c0Og0On4bvakFFQ63dq4aRxmwZCs8t8xm1KpPKGbSBIPj07Z6wW2coXLUkmklREeUFZ2yuqrcfHTXP6F+UPvtVxb73kjbkdIcg8S2YzDTCClVOjc06xj8s8Soa4nqdUiaCDbsC91Q3zjxaXCV1U0LS5Lga3URiwVr1GVy2+6o66GFeNTe1Me5q4gRqa0StMjbZlGBY0OU+Uf+BGLK/0Wng+GPm1P25E6nMRN4Mozcpz589t/PVfTCsOZiNYvvIfm1oVhxRZ1nF4E6yIUDaPD0b3e3xmbhyY7IvHfOdaxZarH1a4ay6nTs7qbag6MNo3blNzKCdigN8W/K8kLAhFkW+lCYsEoqeqlXX3CJfbZY+3OfU+RJo2ZfcVxT7u9APUUMQnALh4U90UpCe/0C+cDnaD/w77UqZUI+Q8ZjcTAGlZfayfbuBYt90F8pSzCH5IIJoZQ8bAJqJ9u7nUnYKwO3XNyTmS7j+8PCd368m+7+VRmY3nMpISGXGcP2t/7ZPiQJ1sT/Cy5fhfxOYs8SHCwOhRhNex/rBcK1aadY9iqVCYZYkUuVFxqMwah7MIfC771tnO0omw6bfaj8Pvtr2GayZwfzlATf1W/0b0ir1DABW82E1JRFobQ1F8xwE7SHqqOWCUIeucQ5keL7pm0rIi4ms4wbstA4ReK4migwROJPTW2YeMPkTvIH0IYJfNvXPBw/BGPmvY3BJhwYiv/lF8jF7qXk9b9E1/hCMxm7bBy0K9Z+tjY7uuP0v8H/48eqKZNJzIzFZwQHqJlSYsH0LcgyqAcNd7HX742NXQk4a8z9wnOfRxyGAgRpAUEiCRMusK7B0tiYxuM4bF8dJ/yB8OTdjn9WSCyYMXiSlsJPU/Pe2ThE/RdXhfKD5B9YPIKtnFo2OLuphJTScYjb3TdtRbBj+Rs)
* [3D Tiles and globe in fog](http://localhost:8080/Apps/Sandcastle/index.html#c=vRoNU9tG9q/sOdeJ3BphIMmlJnBHwEm4g8DFTnMzcSddS2tbg6TVaVcQJ8N/v/f2Q1rJMmDaHp3G1u77/tr3Vg54KiS5jtgNy8kBSdkNOWYiKhL/F7XmTTqBej7mqaRRyvJJp0e+T1JC5owHPGT5gMxoLFgP10TAUnYOq5dRcFXfWvCEvS6k5Km7mtLraE5lxNN3LM5W96dUsDO6ZHmT4G13f5JOUi25r/j685hPmc9SOo3ZWTRfyCidg1IyL9h+A3TG537IUhHJJUDs+H2yTbxd+PiR7PT7/W4TnsqEi2zBcuaHy5QmUeAwMBY70RtHJagFGS8z5o+Oh++HX85O374bK8m3t8mREEySM86vioxIlFqQgw3/DKkzNpNkQeMZ4TMiFww8kTOWDtTmOALKTAqS5TwsAhaS6VIBacEJWJ/snRD0W6yA80kaqMCIgayS8jQE0bTfJ52jt6fk3b8nnQF51n/54oVy1DmLp7zIUzYgL37ee9lHF1k9P6Adfpd0HxiN0VU14XIke5d0u3u7z3f6K/Lp5b+VEmpqLI6jTPAovMz5dRQ202Fot8cszyERLJTXLb2p+JIY3CnIlEkJJG4iubB6fOJ5HBKD3iPTQlZCaSxFxmAWAmMLzVAKRqTG9a3IRRZSyQzFN0UatBnB03HcJQeHepOQWmSD+QwFz6hqHv1ZzhMltN3voq6E3DYNup6FrBsL5Fsxc0myDJezCHTDQNGBATUigcSY8ZxA1FCVczzDkiEITUMSpSH7iotRKjlaTBGJa1lFp/yalXYzhN8rugfkc2WuHpiuVGzS+dWKZGIUcpWGKucf8WdIHaMOjMgbDgEfJZGMrhkJeByzQOnUIxziAbVlNFi0ZI2vJVpEgtzQJbkBbJqSIGY0V6irGIRKQkHrpDIBJvZxyRSMUIuJUjDh0zD0VKGuUuHS7lYEvO4kxeBwEvNPoK6jA1ygtNEeAZe3K90jOZPgRptGBmGSUrGETJlBuijhYqBn3Os5gdEjIoOScxLlWoaujmytYCRUvT1owGBYWDVGtR3/bPhGl31LglZ1y1D7e73aDur1bb9CDVzL1pAdmw+aXmjjDvhWjs+O7r8qZtGMeH8x2oRsBmd/6BnobtfmeSWLn7MEkuwojj1dJohxgE5wl7uAvCcHGmYjC4Kakw4qOumAgpOOUnHSKVVbK0st71HrGxpJy0h/7J2YKFB175SnxvKlzoqQIeE3xW7oYSSCcP20gJC/htKnQ1aFm9guUvXZMyVch62kUlUqOBjeXn4kCeiQLw0ZLHQKP+BFKn2dBZjREAAQ/GkYw7kpWEZzIBcvVf3QJ4klEARFToOlX5lD8x4h22Max1MaXIEiKgncUu6AXdKUxR66r2ctYY6EmnHwEwXE9B5es1RiRQcz5F4Ly+4K6kdlnAcjN1yPJcWRDXd1HFoesKiisVEHsERtWgeaAfXQWqJUNjXxG+fJmLsi3yOdcEkbcWxddAqdgDRLiliCy3gh4qVl2Srf+pSrZN0A8QN2uetUmbI0WCQ0v9rU2n9C2SD63GwEd7dWKyEvc3keQd8iGKyokp2xHDIsoWnA/JTf3FllNggKJ1jxRIQWKQGmnpczwePrWlbahIlSOCxprBhgCLCWxPFcRCsmS8MHKVVhwHQg6ciC2z1CvCaprVWbdWG+wsHK75dEdTqfKeskzFQVl0fF3+hfSnRrqs5DQ2xNyrznErrXN6pUGgToFnrKY1UaYXtRNlBEqBZTgQCsIQSNB0JZz4ecifSpMhhduuC2Q+HwT64qsQ6Ru1PivuR8PI16noIiH08f1dq29bm2ZQ5znoX8JiVbm/9ZUmCyWZQ7uRWlyo4l7RsINUILyROYDwI4HJZgDzgP8VCAjJikMSDZx3FlGLwdcIaBz33d8ZeBpAYOdqGmDS+tgsckqc1F9lUO1JjS0wuQLorXoApJr1slYLsgiF9GfK3eK8Z+QGWw8DAVecx8GKx4bmtGlU7OMPgZ0X61s5kD2Sx5tjZW+QYV5LUNJ0iSKGAtQKtkTF29h04DSk+TtzYIW00vvNpBYKe/uvP8hGaei6U5GFcZHMtmBC1TQIUES4KQY86hncjPWVp4Db44EUq9PVJu05LrUY4nWSGZrRfgxcdG+JuyXSsbNeiwcMSbqisx0EI1jPUDHmvOdIkfiohTmrQ0a8oSnc1Aj6oQrbGGvozz9H2CVVXRhZkLmcC4rMpPvfZWgb5ahVsC371SWLG0M/YdkZRLxRrlLgRYSi6odLTN2X8LqGzQQlujZfAknJBqCUd13GuBQx4UCRyb/pzJYczw6+sl9P+//fU7Qt02UH/rwtkLp+u78fmZPQ0nnUtkSZ62mespfkkYFUXuOEm1Ibb8qmwiGaaTeEy9tBFVJVFbd/MwdcemicfDua4rqIl8dP/0u6wGlCr9S6FbGwPpiGP7g7sV0S5pVQe36oK0kPclB28GkcDav9ddI+bauajWuSq/HlTTIzxDcxYFYv+hrjDNXVvUKep+WiRTll/MNJxCGXNJ44dzUOD3M6iRrnScZ8W5mlhfLyUTddQ54wmTeuuMpXO5ID+ZLTw7ISFEtbXfQvP8NVivwQEbyt1n5mP/4ZHwtiLaEggOyzUBAHma0K8wWyVkNBqSUaxuNh+Vp0H59kXfvZvrW0N+pO6xRhkN2BBP+wHZeeFXF+umpbtKeXDFCxid4dS48kpiXedu25RVoL/WRLb06uOtSZtmWbx8HaXYxIqKR89S7rZJhKogn4spVOtrvIh1MSedNWqCBApTFFMR5NEU5p9rGhe16ccYDpffQCnFYSujudAPBtz0F3iNVb8b82Mdgoek321ORQg5Lge4Bh7o4vWr7saB9dfoglefpZC24ankalzR3SWYAq0ka2I2RHOhN5bttjx01bggHxPdqy089DL4uhDvq9RtFza9VNhOfu9EDwyqp6JwQDotiRUG2wyEzbiIVAW29706A9W7FvWMUpdQNgcM1F0poEF0BtTfkSDFS8tWlTYN6/PZDEysrmC37RrKn1rSBuJTFKrKZqXRV3aNd0yuh3CAf6eBPE1XewUHGXW3GagXBwf6hWh9asF9D/9BAUwkqdvcCrGMr3V3tNBwUgS8UOIDH0vPh2nfmsL/6hTrFiMR4q3a6acG7S7ZNmfF/eYrgYRcQmTE+hYeKvsOXi2QH+tC3P7wm3vT1urLOoat8cY9+HLsNIVO7ki/h1hpcmueMO+Zb3vr/KneA+Pc/uXk4tN79OefyOfy9P3xuy+j8dGHsY6cNbyQ9N2Uzi8+joZfzi9+Gd4h8v1ktECWzCN1t9H+ECN/vHy0iR/IRqs0fH+iNer0Oq9UYB7aOvyPKMl4LkmRx57vb0uWZBD6TGxPi+AKSnMghDuZPzFF6nt1uWbL2IDQKQz+MFTsV5sY/wPyvP+DsyZ5NiD97KuzhDfl85wXabgV8BgbiSfhHv7nwNxghgGtGuKC4TkywJs7l8W3LfXOdUB+hr9y/XZFj8GCX9e1gdFWIH92swU9X/SNtWOrK377g5MHW0NpvvP8h1WtavI7nHw1arkMVk2Vz6fUe7bbI/b/vv+y67DIYKqErmhAntVtznOwwFZOw6gQzc0Z6LaF6oNo7o4rmrCDRbv+toauccvL9TTHERyqLlFHmt1+uzRPVKV1cbBpGKgAdD0AzfwWjaN52thySanepIWWWl9DrL7nqjStz5e1YDOxPpsx1u+72PjxattN1VdhdE2i8KDl504wRFMhYGdWxPEIzDTplPldodne4fDVNqy1AdSC2qHhApmXFJZjFQMqUGtITTTl1iauWayhkarZwt89ef+8HL4ldgzr1hjUVFllqWbfFeqatB5DB+SVANFdKctRFhChPwQ3AMCh7QFKRi1oauasYTVwpnljBV+j6mluRRBnCqwLAs/3UG1QatxwOC6YNncOHyg3qqtf6OF9BDRR3VZD2guLDWzSdOnKc+lgc027cSya0eMxwbhDvH+N/7O7eTAqnhtGYzUibRaOFd4fFo+K5B8TkIrU/z8iS6P84SHpPtW+V5VIv1u7gFYjpktUabFzaH6o5fv+q214bMcsrzxK+zRDewXW3kyvLfMwbI9Gw1oYntvbIv1jLNVDEn3VUSkdYW9KJHSVaE6aztFtSZTCE0w2+J1+he8vnqkHIVmmdnbgIaSSbk3h7IcVNdEP1t0g9fTE/1HdXA7IU8X0af0sqgmCmYjs4NSDp+cbMXPNuMaJ/wM) -- note that the colors look seamless at the tileset borders when in fog.
* [Cycle color shifts](http://localhost:8080/Apps/Sandcastle/index.html#c=nVZbb9s2FP4rB8Ee5MWhJcuy4uWCdWnRFWjRoAn25IfS0rHFlSYNknLmFfnvO6TkyI6dINiDLzznO9+5Uyq0sg7WAh/QwBUofIAbtKJesr+CLJqeFOF8o5XjQqGZnvTh51QBOLFESZLfYM6lxb6XcSWW3AmtdoW2QIVfdIm3oviBZlc14xY/8w2aQ9VgAPcVwkLqGUKp0YLSjgLEEpwGLxN2Jcm23MKtUAWCI6PbSjttkEthnSggfQ/3QhIDAWRdEgaNoWS8XeDv/D72LqYkJ7YPis8kgkFVohFqEYjtj81UNdViIS1Gknduqe2qQkOnSj9QGZ2pccvzrixfimeqnNk0xSxCH5yXoiMG/sCF27aiIFOHH7VeSNynSt/fNyaRjxtgL7SVEUvhxBot42UZteQe+AgFd0UFEdVBm14XgpbIpF5E3z94BUjNS5/6S/VsKZk3/+VnIHv8Hhy0yd9qoVyoXMGXaDjw5tTkspL4z7NyNjBGpH7+ohBYieRVtWO1M6E33Dj6x1UaeRjA2XA8SfNJzrIsicfxKM/7rWI0nORJmrEsyePJKB9mrSI9z0Z5HLOcvkfJKB2PvLwXtNoIpJk/cPsnhqLcCqrgNy1l63zExlkWk9/JMEuSyeTJx1nMhufj9Hwyou8kPT9Px60mYWk2TnKKdJjleZrkKZ7l2wBoEsFXsZmMueQLS3MRClLVe0tnuavNwdbNjFhUTqG1O9MdhvKOq7Lg1lGvaS7u9YJa8UftnFa07TebgqaePPg9b+wgKiqk9Sx7cHXdRBDCYYSikFrlxdPyvIG/C/kNbjrw//TWleIN3jrwMW9TJWk/C10rFy7MmKQ0rJ/8cc1lFHV0TeP40+VA6L1R7zQXHXzvOnlusacMRv4j5hA99aPd5XAf+vZwSWtbbkCvaK1oW6AQpqglN3LTh79rH2AZVlxYmNNlzhrrLjZPeleJuYNTypbFw4sGsX/vHQU9AlKltwEdowzg1wk7yOM24bawihpN6iGL4Vf4wl3Fbj/RP4In9NO2aKe0D3zt8QFJj4ooEPSeAYa7iCHx7KD2i91NZe9Iip12m4ZnP5rqK9AXK3hoE7+R+1ktu3y6uT+WT6fdDfL4MLyGfTGjQ6P4jeQH0xEaf3rql7YPSRz3Lk76J5fWbSReN5QAv4vlShsHtZERYwOHS3qXoP0YzGpaeMcKKkPr/3Kwa3pZijWI8urIexEUkltLmnkt5Z34l67Q68sB4Q9M2+fq1zUaeoPxsCq5/twIGWOXAzoet3Rayxk3z5j/Aw)

# Author checklist

- [x] Loose ends
    - [x] Try the other method of computing the ellipsoid position mentioned in #11716 and compare visual quality and performance. The first method does a reasonable job
    - [x] Find a way to mark a tile as being "in fog" -- the fog density calculation for dynamic SSE will be relevant here, and only apply the shader to tiles in fog
    - [x] Allow selecting between vertex/fragment shader for doing the atmosphere calculations.
    - [x] Add a render test that puts a tileset on the horizon, adjusts the sun position, and verifies that the color changes. This may require turning off dynamic SSE so the tileset doesn't get culled.
    - [x] Go through `GlobeVS` and `GlobeFS` and see what other details I might have overlooked, as those shaders are rather involved.


packages/engine/Source/Shaders/Model/FogStageFS.glsl
```diff
@@ -0,0 +1,94 @@
+vec3 computeEllipsoidPosition(vec3 positionMC) {
```
ptrgags, line 1: Most of this code is identical to code in vertical exaggeration. Consider pulling out a builtin function for this.

The one tricky thing is vertical exaggeration only needs the height, but fog needs the position on the ellipsoid in world coordinates. So maybe split it into smaller functions.

packages/engine/Source/Shaders/Model/FogStageFS.glsl
```diff
@@ -0,0 +1,94 @@
...
+    //rayleighColor = vec3(1.0, 0.0, 0.0);
+    //mieColor = vec3(0.0, 1.0, 0.0);
+
+    vec4 groundAtmosphereColor = czm_computeAtmosphereColor(positionWC, lightDirection, rayleighColor, mieColor, opacity);
+    vec3 fogColor = groundAtmosphereColor.rgb;
+
+    // Darken the fog
+
+    // If there is lighting, apply that to the fog.
+//#if defined(DYNAMIC_ATMOSPHERE_LIGHTING) && (defined(ENABLE_VERTEX_LIGHTING) || defined(ENABLE_DAYNIGHT_SHADING))
```
ptrgags, line 59: I need to re-enable this. Compare the colors with the globe version and see if it looks comparable the same time of day.

packages/engine/Source/Shaders/Model/FogStageFS.glsl
```diff
@@ -0,0 +1,94 @@
...
+    czm_computeGroundAtmosphereScattering(
+        positionWC,
+        lightDirection,
+        rayleighColor,
+        mieColor,
+        opacity
+    );
+
+    //rayleighColor = vec3(1.0, 0.0, 0.0);
+    //mieColor = vec3(0.0, 1.0, 0.0);
```
ptrgags, line 51: Clean up the debugging code in this file

packages/engine/Source/Shaders/Model/FogStageFS.glsl
```diff
@@ -0,0 +1,94 @@
...
+  return (czm_inverseView * vec4(ellipsoidPositionEC, 1.0)).xyz;
+
+}
+
+vec3 computeFogColor(vec3 positionMC) {
+    vec3 rayleighColor = vec3(0.0, 0.0, 1.0);
+    vec3 mieColor;
+    float opacity;
+
+    vec3 positionWC = computeEllipsoidPosition(positionMC);
```
ptrgags, line 38: This should really be `ellipsoidPositionWC`, it's the position of the vertex projected onto the ellipsoid surface.

ptrgags, line 38: Also this should be handled differently if the scattering is computed in the vertex shader

packages/engine/Source/Shaders/Model/FogStageFS.glsl
```diff
@@ -0,0 +1,94 @@
...
+    //fogColor *= darken;
+//#endif
+
+    // Tonemap if HDR rendering is disabled
+#ifndef HDR
+    fogColor.rgb = czm_acesTonemapping(fogColor.rgb);
+    fogColor.rgb = czm_inverseGamma(fogColor.rgb);
+#endif
+
+    // TODO: fogColor.a is only used for ground atmosphere... is that needed?
```
ptrgags, line 71: Make a note of this in #11717

packages/engine/Source/Shaders/GlobeFS.glsl
```diff
@@ -507,22 +507,24 @@ void main()
...
                 finalAtmosphereColor.rgb = vec3(1.0) - exp(-fExposure * finalAtmosphereColor.rgb);
             #else
                 finalAtmosphereColor.rgb = czm_saturation(finalAtmosphereColor.rgb, 1.6);
             #endif
-            
+
             finalColor.rgb = mix(finalColor.rgb, finalAtmosphereColor.rgb, fade);
         #endif
+
+        //finalColor.rgb = computeEllipsoidPosition() / 1e7;
```
ptrgags, line 527: Remove debugging code

packages/engine/Source/Scene/Model/FogPipelineStage.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @namespace FogColorPipelineStage
+ *
+ * @private
+ */
+const FogPipelineStage = {
+  name: "FogColorPipelineStage", // Helps with debugging
+};
+
+FogPipelineStage.process = function (renderResources, model, frameState) {
```
ptrgags, line 15: This pipeline stage also needs to handle:

1. once I have a way to mark a tile as in fog or not, check for that and set a uniform accordingly. (A uniform is needed so we don't recompile shaders every frame)
2. Once I determine when to use vertex vs fragment shader, the appropriate defines need to be set here.

packages/engine/Source/Scene/Model/FogPipelineStage.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * The fog color pipeline stage is responsible for applying fog to tiles in the distance in horizon views.
+ *
+ * @namespace FogColorPipelineStage
+ *
+ * @private
+ */
+const FogPipelineStage = {
+  name: "FogColorPipelineStage", // Helps with debugging
```
ptrgags, line 12:
```suggestion
  name: "FogPipelineStage", // Helps with debugging
```

packages/engine/Source/Shaders/Builtin/Functions/getDynamicAtmosphereLightDirection.glsl
```diff
@@ -0,0 +1,23 @@
...
+ * Select which direction vector to use for dynamic atmosphere lighting based on the czm_atmosphereDynamicLighting enum.
+ *
+ * @name czm_getDynamicAtmosphereLightDirection
+ * @glslfunction
+ * @see DynamicAtmosphereLightingType.js
+ *
+ * @param {vec3} positionWC the position of the vertex/fragment in world coordinates. This is normalized and returned when dynamic lighting is turned off.
+ * @return {vec3} The normalized light direction vector. Depending on the enum value, it is either positionWC, czm_lightDirectionWC or czm_sunDirectionWC
+ */
+vec3 czm_getDynamicAtmosphereLightDirection(vec3 positionWC) {
```
ptrgags, line 11: Add a note in #11681 that this can be used in the SkyAtmosphereShader (and perhaps the globe shader if we are okay with switching to a uniform rather than a define.

packages/engine/Source/Shaders/Builtin/Functions/computeScattering.glsl
```diff
@@ -0,0 +1,161 @@
...
+    // Compute the scattering amount.
+    rayleighColor = czm_atmosphereRayleighCoefficient * rayleighAccumulation;
+    mieColor = czm_atmosphereMieCoefficient * mieAccumulation;
+
+    // Compute the transmittance i.e. how much light is passing through the atmosphere.
+    opacity = length(exp(-((czm_atmosphereMieCoefficient * opticalDepth.y) + (czm_atmosphereRayleighCoefficient * opticalDepth.x))));
+
+    //rayleighColor = vec3(atmosphereInnerRadius / 1.0e7, lastVals.x / 1.0e7, 0.0); //lastVals;
+    //vec3(float(PRIMARY_STEPS) / 16.0, float(LIGHT_STEPS) / 4.0, 0.0);//mieAccumulation; //rayleighAccumulation;
+    //rayleighColor = w_stop_gt_lprl /*w_inside_atmosphere*/ * vec3(1.0, 1.0, 0.0); //w_inside_atmosphere, 0.0);
```
ptrgags, line 160: Remove debugging code

packages/engine/Source/Shaders/Builtin/Functions/computeGroundAtmosphereScattering.glsl
```diff
@@ -0,0 +1,30 @@
...
+ * @name czm_computeGroundAtmosphereScattering
+ * @glslfunction
+ *
+ * @param {vec3} positionWC The position of the fragment in world coordinates.
+ * @param {vec3} lightDirection The direction of the light to calculate the scattering from.
+ * @param {vec3} rayleighColor The variable the Rayleigh scattering will be written to.
+ * @param {vec3} mieColor The variable the Mie scattering will be written to.
+ * @param {float} opacity The variable the transmittance will be written to.
+ */
+void czm_computeGroundAtmosphereScattering(vec3 positionWC, vec3 lightDirection, out vec3 rayleighColor, out vec3 mieColor, out float opacity) {
```
ptrgags, line 14: Add a note in #11681 that this function can eventually be used for the globe's ground atmosphere after after the API change.

packages/engine/Source/Shaders/Builtin/Functions/computeEllipsoidPosition.glsl
ptrgags, line 1: Not sure yet if this file will stay, I need to try the two different implementations from the issue and compare.

packages/engine/Source/Shaders/Builtin/Functions/computeAtmosphereColor.glsl
```diff
@@ -0,0 +1,44 @@
...
+ * @name czm_computeAtmosphereColor
+ * @glslFunction
+ *
+ * @param {vec3} positionWC Position of the fragment in world coords (low precision)
+ * @param {vec3} lightDirection Light direction from the sun or other light source.
+ * @param {vec3} rayleighColor The Rayleigh scattering color computed by a scattering function
+ * @param {vec3} mieColor The Mie scattering color computed by a scattering function
+ * @param {float} opacity The opacity computed by a scattering function.
+ */
+vec4 czm_computeAtmosphereColor(
```
ptrgags, line 15: Add a note to #11681 that this can be used in other shaders once the API change happens.

ptrgags (commented):
Notes on things I still need to do

ggetz:
Thanks @ptrgags! Can/Should there be an API option to alter the fog falloff visually in order to bring it closer to the camera?

ptrgags:
@ggetz there already is, `fog.density` controls that

ptrgags:
Today I've been comparing the different approaches to computing the ellipsoid position in the shader to see the visual differences and performance differences.

Observations:

* The curvature method and the iterative method (1 iter) don't produce a perceptible difference: (and yes I've double checked that the button does change which shader function is called)
2024-01-08_IterativeNoChange, so it really comes down to performance difference.
* For future reference, there is a difference between the two computations given the approximation methods used. The GIF below is visualizing `abs(curvature - iterative1) / 1000`. The difference is bigger at larger camera distances
2024-01-08_CurvatureVsIterative
* Also, following @jjhembd's suggestion, I tried comparing 1 iteration vs 3 iterations for the iterative method. the difference is small, less than 100m and this happens mostly at the equator, with a smaller difference happening at the poles.
image
image
* In light of the above, if we use the iterative method, one iteration seems sufficient to give realistic fog.

I tried the FPS monitor but it doesn't really have the granularity I need to distinguish the two methods (both hover at 59-60 FPS) . I'll see if there's another way to measure it. If there isn't a large perf difference, I'd rather use the iterative method since it better matches the ellipsoid model of the earth.

ptrgags:
Since it's hard to get a precise number for the shader execution time, I'm trying the approach of running the code many times in a loop to amplify differences so it's noticeable in the FPS monitor.

Also note that I have the fog density turned up and fog still is being applied to tiles, so this is giving a measure of worst-case performance.

In the views below I made sure to include close-up, zoomed out, near-equator and near-poles views.

| View | Curvature method | 1-iteration method |
|---|---|---|
| image | 14 FPS | 20 FPS |
| image | 25 FPS | 32 FPS |
| image | 23 FPS | 31 FPS |
| image | 13 FPS | 17 FPS |

So it looks like the 1-iteration method is a clear winner.

ptrgags:
I made a [local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=rVbrb9s2EP9XuGCAZcClSUqiJDfJ1iVN2iJbg8XdgMEfSkuMTYQWDZJy6hb533d6+JFHiwRoPjjive93xzvmpnQerZS8lRYdoVLeohPpVLXA/zS0YHKQN+cTU3qhSmknBwP0bVIiNBx6tZAaaCN0LbSTg5YqSrUQXplyn+xyWco/TSEvVX4j7T5rKpy8EGtpH7O8tBacjjYxjdszvrZm8a+xuugIQR/E7/qvJ+WkHA7Rm6JAl3PjjZVCK+dVjsJTNFZaukmZNxn7+iA9pCxuhfIbBzloeHluzEzL+xbC03GrEtRuWsBwkxVeWrVQXq2kw6Iogs70NporcFMtkSpBRugWiQcW3M36jV8Yt5xLC6e5uYXAGhge+ro2MyxLMdWy+J6I2Fkq1qVYqPxCzeZelTPQ6PI8bRk7pxuR8Xop8cezs4dGZ9pMH0XWpFf3CaCttF4j4ZCfS+SqEimHZqb2WZjb8oGxXCykFRhAqpWDppsKCSiXXd/steGJsB6+RBkGlMWE8RgnIeWc8CQcoFdREoaUEUx4GlNKwmyAIpLEKU9wyhgHCuX9ppeMVRJa+JGDd1IUEOel8vn8b6N1wDCJwySK0yhhUUQ5AzcE0yymacqiOCJADPkAgVP4o1nICKdhlqVRzOJw24ldxrk2+Q3OK2vB+xguzK4IHyoNeZ1CwzUN/d6ZlBMKN44RFr0i9BXJxpSNGB8RgjlLIppQylgW0jSJ/5sc9O+VoAZ+DjdPizWUoahsDX5NLMT65+CfhmkWxhiyZ5TTJANg4iihhHCcxGGUJCRLBiikCcDOMQDPeJox9mL4I8YJiaOMsIyTKG7hB9cUAAhTRuIkpdEAccxSBoBEEQ9pRmL6k6EnbBRGoxDchFHK0hTyIyQNOedb7FfColzAPwemN16bM7DbjxrlN95bNa28DHpeTFVZyC+9AeqRXv81zEt0WXl0bfLKIVM2FWs1txZMmWsYjfXFq8q8RjDotyO4k8SNdjOY7pq4NnLAX/rKypOm4DCI6ivugjbSzkY7D9uW2M+iPr9+KHACExyEPv+Mdvr1W6e4NE7VgvjL3QA9pq6fpH69e2FfbU3MW849q7XUPsGCxl3XT5+3KBgtsTazYIdFA3mNeAEVWEAg9QZ4u4KPC9gaAI4NejdyXc9AqPi2LIHcFFBdwwGDCDo6OkK9vLdhoB/XrokJ3dW/T9nx+3b2Iv/ezdjZ67bolSiLXDgPepDR2MxgIf5ReQ+9Nzm4qtfAeb0R6sdAu65RkM8l7O+ij46OW9/fXyCdaOO0W5cb1i873nNCOTMz9Lbdh8+P5f4SfZnDq5s12q3Nl/t+ctW/LIRPTqJugaPN3n5+AD98HnSq6LfnPRSuPv118f783RiNXvKweLIz2no8DqXF5GBwcOj8WsvjrqnR72qxNBZeVlYHGA+9XCw1jHM3nFag5nHuXNfU6HC4r3pYqBVSxdETD1uUa+EccK4rra/UV2jv48MhyD9S1aaZIR9X0sK+rcXm9PiiJWKMD4dwfFrTG6Onwj6w/D8) to give a visual comparison of P3DT with terrain.

| Case | Globe | P3DT | Notes |
|---|---|---|---|
| No fog/atmosphere | image |image | |
| + fog enabled | image | image | Even though both implementations use `scene.fog.density` and the same math, for some reason the fog looks closer to the camera for terrain :thinking: |
| + sky atmosphere | image | image ||
| + dynamic lighting | image | image | Color is slightly different. See notes below. |
| + adjust timeline | image | image | Sunset colors are noticeably different. See notes below. |

Notes:

* The terrain geometry is a bit different between the two datasets and/or the culling methods used.
* There's still some code that I need to add to the shader that may impact color. Most notably, There's a block of code in the globe shader that darkens the color based on the view and light direction. This might impact why the sunset colors are so different. I'll work on adding this today.

ptrgags:
Next, a side-by-side with Google Earth. To give a fairer comparison, I adjusted the following settings:

* `globe.atmosphereLightIntensity`/`atmosphere.lightIntensity` I turned up to 25 from the default of 10 since our fog tends to look darker and hazier by default
* Google Earth tends to put the fog closer to the camera, so I set `fog.density` to `1 / (2 * 1500m)`
* Google Earth doesn't use dynamic lighting like we do, so I turned that off

[Sandcastle with updated settings](http://localhost:8080/Apps/Sandcastle/index.html#c=rVZZbxs3EP4rrFFAq0KhSO6t2G5TO05SuI1RKy1Q6CHULi0RpkiB5NpRAv/3zh46bMuBDcQP1nI45zfDmSmMdh7dSHErLDpCWtyiE+FktcD/NLRgclA05xOjPZda2MnBAH2baISGQy8XQgFthK64cmLQUrmWC+6l0btkVwgt/jSluJDFtbC7V1PuxDlfCfv4ygtrweho7dO4PeMraxb/GqvKjhD0gf2u/3qiJ3o4RG/KEl3MjTdWcCWdlwUKT9FYKuEmumgi9vVBeAiZ33Lp1wYKkPDinTEzJe5rCE/HrUhQm2kBw01UeGnlQnp5IxzmZRl0qjfeXIKZaomkBh6uWiQeaHDXqzd+YdxyLiyc5uYWHGtgeGjrysyw0HyqRPkUC99qKleaL2RxLmdzL/UMJLo4T9uLrdE1y3i1FPjj2dlDpTNlpo882+NbKbSTfgVMFBM0RAGDn18QjQnBpL9fK7/vxQftN0pYjMl34lNP8De41wUMZSCVWiHukJ8L5CqNpEMzU4NRmtuHERR8ISzHkL1aOGjKvBSQft0V9M77OOHWwxfXYUBZTFgS4zSkSUKSNBygV1EahpRB0EkWU0rCfIAiksZZkuKMsQQoNOk3RW6sFPC2Hhl4L3gJfl5IX8z/NkrVUMZhGsVZlLIoogkDMwTTPKZZxqI4IkAMkwECo/BH85CRhIZ5nkUxi8PNE+kiLpQprnFRWQvWx/CSt9XxR6UgrlN4Cc1L++BMlhAKrYARFr0i9BXJx5SNWDKCrCYsjWhKKWN5SLM0/m9y0Jb+bhpq8OfQFhRfQSrKytYJqIklX/2YHGRhlocxBgQYTWiaAzhxlFJCEpzGYZSmJE8HKKQpQJ9gAJ8lWc7Yi1MQsYSQOMoJyxMSxW0KwDQFEMKMkTjNaDRACWYZA1CiKAlpTmL6g+EnbBRGoxDMhFHGsgziIyQLkyTZ4H/DLSo4/DhQvbbanOG6/ahRfuO9ldPKi6Dn+VTqUnzpDVCP9PqvoZmji8qjK1NUDhndZKyV3GgwulDQt+uuUOmiRjDot/Oh48SNdNM17xq/1nxwv/SVFSdNwqFL1v3HBa2nnY62WbclsRtFfX79kOEExgswff4R5fTzt05waaC1ACP+cjdAj6mrvdSvdy+sq42KeXtzT2vNtUuwIHHX1dPnDQpGQT80s2CLRQN5jXgJGViAI/V4ensDH+cw0gAcG/Suxarug5DxTVoCsU6gvIIDBhZ0dHSEekVvfYG+n7vGJ3RX/9+nx+/q2fH8qZex1deN+Euuy4I7D3IQ0djMYFr/XnkPtTc5uKxn1Lt6sNSbSrtLoKCYC1guyj46Om5tPz3dOtbGaDfL11c/be+e48qZmaG37bB+vi/3J/zLDF5er9B2pr/c9t495GUufHICddsFWi8Vz3fgu7tLJ4p+fd4Wc/npr/MP796P0eglW8/eymjz8diVFpODwcGh8ysljruiRr/JxdJYWPusCjAeerFYKmjnbjitQMzjwrmuqNHhcFf0sJQ3SJZHe7ZuVCjuHNxcVUpdyq9Q3seHQ+B/JKpM00M+3ggL87Zmm9Pj85aIMT4cwnG/pDdGTbl9oPl/)

| Globe | P3DT | Google Earth |
| --- | --- | --- |
| image | image | image |

The thing that stands out to me is Google Earth's atmosphere effects have more pinks and purples when ours only has blue and white (with dynamic lighting off). Even our sunsets are more orange than the pinks and purples.

ptrgags:
I added the code to darken the lighting, now the colors match better between 3D Tiles and the globe:
2024-01-09_DarkenLighting

Perhaps due to different geometry and different methods for computing the ellipsoid position, 3D Tiles fog tends to go dark at slightly higher sun angles than terrain does: (but if I wasn't looking for such subtle differences I might not have noticed this.
2024-01-09_DifferenceInSunsets

packages/engine/Source/Shaders/Model/FogStageFS.glsl
ptrgags: I ended up not using this version of the function, so moot point.

packages/engine/Source/Scene/Model/ModelRuntimePrimitive.js
```diff
@@ -303,6 +305,10 @@ ModelRuntimePrimitive.prototype.configurePipeline = function (frameState) {
 
   pipelineStages.push(AlphaPipelineStage);
 
+  if (fogRenderable) {
+    pipelineStages.push(FogPipelineStage);
+  }
```
ptrgags, line 310: I just noticed something: this pipeline stage only needs to be called once per model, not once per primitive, so move this to ModelSceneGraph.

ptrgags:
I looked through the old Globe code closer and learned a few things I didn't realize before:

* The check of whether a tile is "in fog" has a different purpose than I thought. Since the threshold is very small (EPSILON3), it's to turn off fog rendering when the camera is in space. I did try setting the threshold higher so it also filters out tiles close to the camera, but this produces a noticeable color seam between tiles.
* The fragment vs vertex check is actually only relevant for ground atmosphere and even then, only when the camera is in space. Fog is actually supposed to use the vertex shader always. The existing `GlobeFS` makes this a bit confusing, so I'll add some comments to clarify (and notes in #11717 as needed.). So I'll reorganize code around to match this.
* The ellipsoid position calculation is actually for the ground atmosphere, not fog. For fog you just use the fragment position in world coords :sweat_smile:. It wasn't a waste though, the function will still be needed for the ground atmosphere PR

ptrgags:
While moving code to the vertex shader like the globe does, I realized I was computing the camera distance incorrectly in the shader. Fixing that makes the fog distance look more consistent:

2024-01-09_FixedLighting

What's left?

- [ ] fix existing unit tests
- [ ] Add more unit tests
- [ ] Check what happens if a model is far away but above the camera (e.g. a satellite)
- [ ] Consider adding an option to opt-out of atmosphere lighting for an individual tilesets/models, or is it better to just turn it off for the scene entirely.

packages/engine/Source/Shaders/SkyAtmosphereFS.glsl
```diff
@@ -42,14 +43,7 @@ void main (void)
     #endif
 
     #ifdef COLOR_CORRECT
-        // Convert rgb color to hsb
-        vec3 hsb = czm_RGBToHSB(color.rgb);
-        // Perform hsb shift
-        hsb.x += u_hsbShift.x; // hue
-        hsb.y = clamp(hsb.y + u_hsbShift.y, 0.0, 1.0); // saturation
-        hsb.z = hsb.z > czm_epsilon7 ? hsb.z + u_hsbShift.z : 0.0; // brightness
```
ptrgags, line 50: @ggetz or @lilleyse is there any reason the old code doesn't allow brightness shifts less than epsilon 7? Most notably, why aren't negative values allowed?

I'm assuming not, so when making `czm_applyHSBShift()` I implemented this as

`hsb.z = clamp(hsb.z + u_hsbShift.z, 0.0, 1.0);` similar to saturation, that works fine even for negative values.

packages/engine/Source/Shaders/Builtin/Functions/approximateTanh.glsl
```diff
@@ -0,0 +1,10 @@
+/**
+ * Compute a rational approximation to tanh(x)
+ *
+ * @param {float} x A real number input
+ * @returns {float} An approximation for tanh(x)
+*/
+float czm_approximateTanh(float x) {
+    float x2 = x * x;
+    return max(-1.0, min(+1.0, x * (27.0 + x2) / (27.0 + 9.0 * x2)));
```
ptrgags, line 9:
```suggestion
    return max(-1.0, min(1.0, x * (27.0 + x2) / (27.0 + 9.0 * x2)));
```

packages/engine/Source/Shaders/Model/FogStageFS.glsl
ptrgags, line 1: Similarly, these shaders should be renamed to `AtmosphereStage(FS|VS).glsl`

packages/engine/Source/Shaders/Builtin/Functions/computeScattering.glsl
```diff
@@ -0,0 +1,152 @@
...
+            lightOpticalDepth += exp(-lightHeight / heightScale) * lightStepLength;
+
+            // Increment distance on light ray.
+            lightPositionLength += lightStepLength;
+        }
+
+        // Compute attenuation via the primary ray and the light ray.
+        vec3 attenuation = exp(-((czm_atmosphereMieCoefficient * (opticalDepth.y + lightOpticalDepth.y)) + (czm_atmosphereRayleighCoefficient * (opticalDepth.x + lightOpticalDepth.x))));
+
+        //lastAttenuation = vec3(rayStepLength, lightStepLength);
```
ptrgags, line 136: I see dead code.

packages/engine/Source/Shaders/Model/ModelVS.glsl
```diff
@@ -104,7 +104,11 @@ void main()
 
     // Compute the final position in each coordinate system needed.
     // This returns the value that will be assigned to gl_Position.
-    vec4 positionClip = geometryStage(attributes, modelView, normal);    
+    vec4 positionClip = geometryStage(attributes, modelView, normal);
+
+    #ifdef HAS_FOG
```
ptrgags, line 109: I would add a note that this must go after the geometry stage as it uses v_positionWC

packages/engine/Source/Scene/Model/Model.js
ptrgags, line 1: Make sure to add tests for rendering a model

packages/engine/Source/Scene/DynamicAtmosphereLightingType.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Get the lighting enum from the older globe flags
+ *
+ * @param {Globe} globe The globe
+ * @return {DynamicAtmosphereLightingType} The corresponding enum value
+ *
+ * @private
+ */
+DynamicAtmosphereLightingType.fromGlobeFlags = function (globe) {
```
ptrgags, line 42: Add unit tests for these

ptrgags (commented):
Some notes about what tests to add and other cleanup tasks

ptrgags:
with [this local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=tVdtb9s2EP4rnD8pg0fJTtJunhssibcmWNJ1tVMMmIeClk4WEYoUSCqJV+S/90hJtqx4H4zMRgCTx+Pd89yLeYmVNJY8cHgETd4RCY/kEgwvc/rZy4J5L/b7SyUt4xL0vNcnX+eSEC5TdaGeRiRlwkDfiQwIiC1X8lomPGZW6a3TjCXq0YyI1WUjUaVIziXPmYW1/Pno57mcywoUNTFIoKla0gSk4XaFMAc0IiEJhvj1PRlEUeRubOkzmytTZKCBJivJch7f8GVmuVzi9ZrhpDo4X6s2KrNVAXR69+Hm+v3VrGt5KdQCKEi2ENCy6bB72GkpfQhIrAFZ3aoERFBq0ScZOPWjKnq1TZCWWw6GasjVA5wLEVTsCYl9agqFnJ25NexLpi2umDymqVb5BJYawATuCiE/DIbHNHp7cvJm8FO/Ep2c0Og0On4bvakFFQ63dq4aRxmwZCs8t8xm1KpPKGbSBIPj07Z6wW2coXLUkmklREeUFZ2yuqrcfHTXP6F+UPvtVxb73kjbkdIcg8S2YzDTCClVOjc06xj8s8Soa4nqdUiaCDbsC91Q3zjxaXCV1U0LS5Lga3URiwVr1GVy2+6o66GFeNTe1Me5q4gRqa0StMjbZlGBY0OU+Uf+BGLK/0Wng+GPm1P25E6nMRN4Mozcpz589t/PVfTCsOZiNYvvIfm1oVhxRZ1nF4E6yIUDaPD0b3e3xmbhyY7IvHfOdaxZarH1a4ay6nTs7qbag6MNo3blNzKCdigN8W/K8kLAhFkW+lCYsEoqeqlXX3CJfbZY+3OfU+RJo2ZfcVxT7u9APUUMQnALh4U90UpCe/0C+cDnaD/w77UqZUI+Q8ZjcTAGlZfayfbuBYt90F8pSzCH5IIJoZQ8bAJqJ9u7nUnYKwO3XNyTmS7j+8PCd368m+7+VRmY3nMpISGXGcP2t/7ZPiQJ1sT/Cy5fhfxOYs8SHCwOhRhNex/rBcK1aadY9iqVCYZYkUuVFxqMwah7MIfC771tnO0omw6bfaj8Pvtr2GayZwfzlATf1W/0b0ir1DABW82E1JRFobQ1F8xwE7SHqqOWCUIeucQ5keL7pm0rIi4ms4wbstA4ReK4migwROJPTW2YeMPkTvIH0IYJfNvXPBw/BGPmvY3BJhwYiv/lF8jF7qXk9b9E1/hCMxm7bBy0K9Z+tjY7uuP0v8H/48eqKZNJzIzFZwQHqJlSYsH0LcgyqAcNd7HX742NXQk4a8z9wnOfRxyGAgRpAUEiCRMusK7B0tiYxuM4bF8dJ/yB8OTdjn9WSCyYMXiSlsJPU/Pe2ThE/RdXhfKD5B9YPIKtnFo2OLuphJTScYjb3TdtRbBj+Rs) I confirmed that fog does apply to models far above the camera (see the "Satellite" option in the dropdown)

image

ptrgags:
@ggetz this PR is now ready for review!

ggetz:
Thanks @ptrgags! Would you mind updating the Testing Plan section in the PR description when you get the chance?

packages/engine/Source/Scene/Scene.js
```diff
@@ -512,6 +514,8 @@ function Scene(options) {
    */
   this.cameraEventWaitTime = 500.0;
 
+  this.atmosphere = new Atmosphere();
```
ggetz, line 517: I believe this warrants:
* Inline documentation (Please include a code example if applicable)
* Updates to `CHANGES.md`

packages/engine/Specs/Scene/SkyAtmosphereSpec.js
```diff
@@ -52,9 +53,9 @@ describe(
       s.destroy();
     });
 
-    it("draws sky with setDynamicAtmosphereColor set to true", function () {
+    it("draws sky with dynamic lighting (scene light source)", function () {
       const s = new SkyAtmosphere();
-      s.setDynamicAtmosphereColor(true, false);
```
ggetz, line 57: So are SkyAtmosphere and related classes now deprecated?

ptrgags: no, there are no deprecations in this PR! this was a private method called by Scene to pass globe settings to the `SkyAtmosphere`. I just changed it to take the new `DynamicAtmosphereLightingType` enum for consistency. The deprecation would happen in #11681, but not this PR.

in #11681, `SkyAtmosphere` and `Globe` may have some parameters moved to `Atmosphere`, but they still will exist for rendering the sky and terrain.

packages/engine/Source/Scene/Scene.js
ptrgags: I did mention `scene.atmosphere` in the changelog entry, is this big enough to warrant a separate changelog entry?

packages/engine/Specs/Scene/Model/AtmospherePipelineStageSpec.js
```diff
@@ -0,0 +1,113 @@
...
+import ShaderBuilderTester from "../../../../../Specs/ShaderBuilderTester.js";
+
+describe("Scene/Model/AtmospherePipelineStage", function () {
+  const mockModel = {
+    boundingSphere: {
+      center: Cartesian3.fromDegrees(0, 0, 0),
+    },
+  };
+
+  function mockFrameState() {
```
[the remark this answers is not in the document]
ptrgags: We might want to discuss this a bit. This sort of mocking is a common pattern throughout the pipeline stage specs :sweat_smile:.

I agree that it's mocking too much, I've been bit plenty of times before by refactoring while working on `Model` a couple years ago.

On the other hand, the intent with the pipeline stages is they are meant to be static functions that should (in theory) be easy to test in isolation from the rest of the code. However, all three inputs (renderResources, model, frameState) are non-trivial objects to construct (especially model which requires async loading), so it very quickly goes from unit to integration test.

ptrgags:
@ggetz I rounded up all the Sandcastles from the issue and this PR and put them in the description (+ some new ones!). 

I'll look into the test failure, I must have formatted my JSDoc incorrectly or referenced something incorrectly.

ptrgags:
@ggetz I've addressed your feedback, this is ready for review again.

packages/engine/Source/Scene/Atmosphere.js
```diff
@@ -0,0 +1,151 @@
...
+import DynamicAtmosphereLightingType from "./DynamicAtmosphereLightingType.js";
+
+/**
+ * Common atmosphere settings used by sky atmosphere, ground atmosphere, and fog.
+ *
+ * <p>
+ * This class is not to be confused with {@link SkyAtmosphere}, which is responsible for rendering the sky.
+ * </p>
+ * <p>
+ * Currently, these settings only apply to 3D Tiles and models, but will eventually affect the sky atmosphere and globe. See {@link https://github.com/CesiumGS/cesium/issues/11681|issue #11681}.
```
ggetz, line 11: Let's remove this line, just in case the API changes take longer than expected.

packages/engine/Source/Scene/Atmosphere.js
```diff
@@ -0,0 +1,151 @@
+import Cartesian3 from "../Core/Cartesian3.js";
+import DynamicAtmosphereLightingType from "./DynamicAtmosphereLightingType.js";
+
+/**
+ * Common atmosphere settings used by sky atmosphere, ground atmosphere, and fog.
```
ggetz, line 5:
```suggestion
 * Common atmosphere settings used by 3D Tiles and models for rendering sky atmosphere, ground atmosphere, and fog.
```

packages/engine/Source/Scene/DynamicAtmosphereLightingType.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+const DynamicAtmosphereLightingType = {
+  /**
+   * Do not use dynamic atmosphere lighting. Atmosphere lighting effects will
+   * be lit from directly above rather than using the scene's light source.
+   *
+   * @type {number}
+   * @constant
+   */
+  OFF: 0,
```
ggetz, line 16: This is a bit picky, but other enums opt for `NONE` for the negative case, ie. `HeightReference.NONE`.

packages/engine/Source/Scene/Scene.js
```diff
@@ -512,6 +514,17 @@ function Scene(options) {
    */
   this.cameraEventWaitTime = 500.0;
 
+  /**
+   * Settings for atmosphere lighting effects. This is not to be confused with
+   * {@link Scene#skyAtmosphere} which is responsible for rendering the sky.
+   * <p>
+   * Currently these settings only apply to 3D Tiles and models. In the future this will apply to the globe as well, see {@link https://github.com/CesiumGS/cesium/issues/11681|issue #11681}.
```
ggetz, line 521: Same comment as above, let's omit this line.

packages/engine/Source/Scene/Scene.js
```diff
@@ -512,6 +514,17 @@ function Scene(options) {
    */
   this.cameraEventWaitTime = 500.0;
 
+  /**
+   * Settings for atmosphere lighting effects. This is not to be confused with
```
ggetz, line 518:
```suggestion
   * Settings for atmosphere lighting effects affecting 3D Tiles and model rendering. This is not to be confused with
```

packages/engine/Source/Shaders/SkyAtmosphereFS.glsl
ggetz: None that I know of.

ggetz:
@jjhembd Would you mind taking a pass on the shaders in this PR?

ptrgags:
@ggetz @jjhembd Updated the PR based on @ggetz's feedback.

packages/engine/Source/Shaders/Builtin/Functions/computeAtmosphereColor.glsl
jjhembd: Can we eliminate the code duplication even before the API change?
I know at one point you mentioned the Globe using defines rather than a uniform. That seems like a change we could make internally without affecting the API.

jjhembd (commented):
Thanks @ptrgags, this looks like a very nice cleanup of the atmosphere shaders.
My main question is: can we drop some of the old versions sooner, even before the API change suggested in https://github.com/CesiumGS/cesium/issues/11681? Some of the smaller functions can be trivially consolidated (see comments below). It would be nice to do a bigger consolidation now if we can.

packages/engine/Source/Shaders/Builtin/Functions/applyHSBShift.glsl
```diff
@@ -0,0 +1,21 @@
...
+ */
+vec3 czm_applyHSBShift(vec3 rgb, vec3 hsbShift) {
+    // Convert rgb color to hsb
+    vec3 hsb = czm_RGBToHSB(rgb);
+
+    // Perform hsb shift
+    // Hue cycles around so no clamp is needed.
+    hsb.x += hsbShift.x; // hue
+    hsb.y = clamp(hsb.y + hsbShift.y, 0.0, 1.0); // saturation
+    hsb.z = clamp(hsb.z + hsbShift.z, 0.0, 1.0); // brightness
```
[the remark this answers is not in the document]
ptrgags: ohh that's what that ternary operator was for in the original function. I'll update the new function.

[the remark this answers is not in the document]
ptrgags: I added the old code, but with a parameter `bool ignoreBlackPixels` so if some future shader wanted a HSB shift but did not want that behavior, it can be turned off.

packages/engine/Source/Shaders/Builtin/Functions/computeAtmosphereColor.glsl
ptrgags: I certainly would like that but I need to examine how big of a change this would be. The problem is one version of these functions use uniforms set by a specific primitive (`u_atmosphereWhatever`) and others use the new automatic uniforms (`czm_atmosphereWhatever`). You'd need to tie the two together. I have an idea about how to do that, but I have to see if it's too big for this PR.

ptrgags: https://github.com/CesiumGS/cesium/issues/11681#issuecomment-1899006593 -- here are more ideas of how to make this work during the deprecation period (in short, have the globe/sky hold a reference to the atmosphere whenever attached to a scene, and the old parameters will just update the new common one). 

While certainly this could be done before the API deprecation, I think it's beyond the scope of this PR since it would add a decent chunk of code here. @ggetz would you agree?

packages/engine/Source/Shaders/Builtin/Functions/computeScattering.glsl
```diff
@@ -0,0 +1,150 @@
...
+ * @glslfunction
+ *
+ * @param {czm_ray} primaryRay The ray from the camera to the position.
+ * @param {float} primaryRayLength The length of the primary ray.
+ * @param {vec3} lightDirection The direction of the light to calculate the scattering from.
+ * @param {vec3} rayleighColor The variable the Rayleigh scattering will be written to.
+ * @param {vec3} mieColor The variable the Mie scattering will be written to.
+ * @param {float} opacity The variable the transmittance will be written to.
+ */
+void czm_computeScattering(
```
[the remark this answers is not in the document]
ptrgags: I'd like to, but the `czm_` version uses automatic uniforms but the `computeScattering` one uses uniforms set by the particular sky/globe primitive, so it's not trivial to switch over. See my other comment for thoughts on how to do that.

packages/engine/Source/Shaders/Builtin/Functions/computeAtmosphereColor.glsl
ggetz: Let's try to keep the scope of this PR small if possible.

ptrgags:
@ggetz @jjhembd I addressed all the remaining feedback

ggetz:
@jjhembd Please merge once you are happy!

jjhembd (approved):
Looks good, thanks @ptrgags!

PR #11746, opened 2024-01-07
Low resolution model could be loaded when the camera is far from the model

Leosunshine:
# Description
   The rendering could be low efficient when lots of models at high resolution are loaded and rendered. A property named as "uriLowRes" and "uriCriticalDistance" are added into the modelGraphics and some code added into modelVisualizer, so that the gltf model whose uri is "uriLowRes" will be rendered when the distance between the camera and the model farther than the uriCriticalDistance(whose unit is meter).

 The implement of this feature is added a distance calculation (ModelVisualizer.prototype.postUpdate) after the update of the ModelVisualizer. The postUpdate operation is also after the EntityView.update so that the camera.positionWC is Synchronize with the entity.position.

With this improvement, the fps of the frame will be improved when lots of 3D models at high resolutions in the scenes. I also wrote an example in the sandcastle(development/UriLowRes.html), but I don't know how to update the gallery_index.js autometically.

The CesiumJS is an awesome project and I have learned a lot about webgl rendering thanks to the fantastic open-source spirits of the contributers. This is my first time to submit an pull request, and if there any improper action or low quality code, please contact me. thank you.

with the new feature
捕获

without the new feature
捕2

## Testing plan

# Author checklist

- [-] I have added my name to `CONTRIBUTORS.md`


Leosunshine:
Only the model near the camera will be rendered as a high resolution model as the follow pic shows.

Leosunshine:
Only the model near the camera is rendered as a high resolution model if the uriCriticalDistance is defined. as the follow pic says.
捕3

ggetz:
Hi @Leosunshine, before we can review any code, you'll need to submit a Contributor License Agreement (CLA). See [`CONTRIBUTING.md`](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md) for more info.

Also before we review the code, could you please explain how this feature would be different from using [the distance display condition property](https://sandcastle.cesium.com/?src=Distance%20Display%20Conditions.html)? It sounds like what you're describing here is already possible with that feature.

Leosunshine:
> Hi @Leosunshine, before we can review any code, you'll need to submit a Contributor License Agreement (CLA). See [`CONTRIBUTING.md`](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md) for more info.
> 
> Also before we review the code, could you please explain how this feature would be different from using [the distance display condition property](https://sandcastle.cesium.com/?src=Distance%20Display%20Conditions.html)? It sounds like what you're describing here is already possible with that feature.

Thank you for your reply.  I've signed the CLA as the CONTRIBUTING.md guide.
Using the distance display condition property is to determined whether the model (or billboard, label, or other elements of an Entity) visible when the camera is near or far from the model. 

The new feature in this pull request is just for the gltf model, and different 3D models will be rendered when the camera in different positions. Two models at different resolutions will be binded to an Entity, not one. The entity could be display as a 3D model when the camera is far enough from it, and it could be distinguished when the minimumPixelSize and maximumScale is defined.

For Example, if there are many satellite in the scene, if the uriLowRes and the uriCriticalDistance is defined, the users will see lots of satellites 3D models at low resolution and A 3D model at high resolution, which belone to the trackedEntity. Even at different resolutions, they are all satellite 3D models.

捕获
in this example, the ball near the camera is a 3D model, and the ball far from the camera is also a 3D model, but at low resolution.

onsummer:
This feature like you implement a smaller "3d-tile-set" that only has 2 level of tiles

Leosunshine:
> This feature like you implement a smaller "3d-tile-set" that only has 2 level of tiles

Yes, you're right. This feature is to implement a similar effect that applied to the normal entity in the scene. For normal Entity, its position could be time-variant and the distance between the entity and the camera is dynamic, too.

ggetz:
Thanks for the explanation @Leosunshine! However, I don't think this feature provides functionality that is not already possible using entities with a [DistanceDisplayCondition](https://sandcastle.cesium.com/?src=Distance%20Display%20Conditions.html).

Given that, I'm going to close this PR. But thank you for the suggestion!

Issue #11705, opened 2023-12-18
writeTextToCanvas() background longer than rendered text


AnonymousSausage77:
I am trying to draw text to a label and rotate it. I also want to give the rotated text a background. I found this function:
Cesium.writeTextToCanvas() which does this, except it always appends whitespace to the beginning of the word leaving an ugly 'padding'. In the sandcastle, I have illustrated this behaviour by creating two examples, one with a single character '3' and a longer word, both with this whitespace appended.

![image](https://github.com/CesiumGS/cesium/assets/579118/00d9b09f-5bf3-4a64-871f-0620f573c1aa)
![image](https://github.com/CesiumGS/cesium/assets/579118/3c493918-0d8e-4b9a-9efb-9b804ed7a66e)

What I would expect to happen: I expect that the background (in the sandcastle, rendered as blue) would be the same length as the word that you enter and not any longer/shorter.

Sandcastle example:
[...]

Issue #10649, opened 2022-08-07
writeTextToCanvas treats "-" hyphen/minus, or "_" underscore, characters as space in some cases.


maxizrinUX:
writeTextToCanvas doesn't show anything for only hyphens or underscores, and also cuts off sometimes, but that seems to be another issue.
(Running Cesium JS in Electron, there's no cutoff, but all hyphen or underscore text still does not render.)

Text with mixed hyphens and underscores renders only hyphens.

Checking the return value, it seems canvas height is set to 0, when the string contains only hyphens or only underscores.

An HTML canvas works as a workaround.

Sandcastle example:
https://sandcastle.cesium.com/#c=vZVRa9swEMe/yuGXOmDJabvBaN1sLHsc7GFhT4Yg2+dETNYF6ezMG/3uk2NStq5rUkh6tsF30v+O31mSO+Wg07hFB3dgcQtz9Lpt5LddLM6jcufPybLSFl0eTW5zm9tRI9GyZo1eqqqKf+UWYEM+RMje7DPNlePwpuy1rB01n3DlEH08TWA6SQZFoY0pSLnqBnYZANIUFmvtwa9p6xMoWobgsdNNgxWQNT0MqYDXCAZrfj/KdKNW+FB36zTjAn/wgubKdsoHGPi/MW2EeGY8gCdjGV8qE8pcDd59iN2friNiKqfBrg81piL0YIl3HYKSnMOSTS+P7YMQFyLYwLRj+4sreQRWkvUMGNbHCRHfnh0RlsulgPG6eM2v9+7caIFL7J8DYKfDunwZlWJQxrwAKdjrsVydFUaI/eZ6dsmdlOjNIaJwl+Ecpbo+muMjMYdTdogeD1ObfkExTm6jJMo89wZnoxDgg2425BhaZ2IpU8ZmY1TASou2/I4sS++HZMPULP1TmlW6A13dPfE3gtIo78NI3RrzVf/EPJplaZj/j9SQqrRdfenQGdUP09aXs89jUEqZpcF9WslEplDuUebf

Browser:
Vivaldi 5.3.2679.70 (Stable channel) (64-bit) 

Operating System:
Windows 10 Pro 21H2 64-bit

ggetz:
[...]

Issue #9945, opened 2021-11-30
Writing some special character in label will cause Cesium broken


ZenDay:
Sandcastle example: https://sandcastle.cesium.com/#c=bZCxboMwEIZf5cQSkJCdNBshqBIdK3Vo1cmLgSO1amxkH6S0yrvXQCNVbTbf5/87/bpROhgVntHBEQyeoUSvho69LiwWUb3MpTUklUEnouQgjDC1NZ4Ag7TaDA0pUuiZbJr4SxiA3vpArMmuO0vpKLyk2bPW2e4BTw7Rx9sUtkk6G1pWqDNYbADCD8pgI0S1u9tvVsb5FV/ZJZiXudNPj1ZPLzbG5BClUe5p0lisKsC96nrrCAanY8Y4YddrGRrxaqjfkVjt/bxojub8t5o3agTVHG9cA2otvQ8/7aD1s/pEERU5D/l/qrayUeb0NKLTcppjb7vicYWMsZyH8bZJ1upKuj+bvwE

Browser: Chrome 96.0.4664.55

Operating System: macOS 11.6

Using some special character like '\b' will cause Cesium Broken.

I found that the problem may be at file `/Source/Core/writeTextToCanvas.js` L46
```js
    var ascent, descent;
    // Find the number of rows (from the top) until the first non-white pixel
    for (i = 0; i < length; ++i) {
      if (pixelData[i] !== 255) {
        ascent = (i / width4) | 0;
        break;
      }
    }

    // Find the number of rows (from the bottom) until the first non-white pixel
    for (i = length - 1; i >= 0; --i) {
      if (pixelData[i] !== 255) {
        descent = (i / width4) | 0;
        break;
      }
    }

    var minx = -1;
    // For each column, for each row, check for first non-white pixel
    for (i = 0; i < width && minx === -1; ++i) {
      for (j = 0; j < height; ++j) {
        var pixelIndex = i * 4 + j * width4;
        if (
          pixelData[pixelIndex] !== 255 ||
          pixelData[pixelIndex + 1] !== 255 ||
          pixelData[pixelIndex + 2] !== 255 ||
          pixelData[pixelIndex + 3] !== 255
        ) {
          minx = i;
          break;
        }
      }
    }

    return {
      width: metrics.width,
      height: descent - ascent,
      ascent: baseline - ascent,
      descent: descent - baseline,
      minx: minx - padding / 2,
    };
```
[...]

Issue #9767, opened 2021-08-30
Investigate `writeTextToCanvas` and `measureText` function


ebogo1:
In https://github.com/CesiumGS/cesium/pull/9765 we settled on a manual pixel-counting approach instead of using the stock browser function `CanvasRenderingContext2D.measureText()`. This is because browser results were inconsistent, and at the time of the PR, some of the TextMetrics properties were unsupported by Firefox for Android.

If `CanvasRenderingContext2D.measureText()`, it would be cleaner to use it instead of drawing the text to a canvas and counting each pixel. Instead we could replace `ascent`, `descent`, and `minx` by the `TextMetrics` properties `actualBoundingBoxAscent`, `actualBoundingBoxDescent`, and `actualBoundingBoxLeft`. 

Until then, maybe it's worth taking another look at why `writeTextToCanvas` is needed..

javagl:
Related to https://github.com/CesiumGS/cesium/issues/10649#issuecomment-1222595171 : It might be easy to replace this function with the built-in functions now.

ggetz:
Additional issues reported in https://github.com/CesiumGS/cesium/issues/11705.

donmccurdy:
Fix in progress:

- #13081

PR #11747, opened 2024-01-07
Remove custom measure text

javagl:
# Description

The `writeTextToCanvas` function writes text into a canvas, basically to create "an image of the text". This is used for labels and Billboards. The _size_ of this canvas is computed with a custom function that was supposed to serve a smilar purpose as [`measureText`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText). This custom function has some issues. The reason for using the custom implementation was probably that certain properties of the `TextMetrics` that are required by `measureText` had not been supported on all browsers. As mentioned in [a related comment](https://github.com/CesiumGS/cesium/issues/10649#issuecomment-1222595171), the [`TextMetrics` browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics#browser_compatibility) summary suggests that all required properties are now widely available. So the custom implementation can be replaced by the built-in `measureText` function.

## Issue number and link

Fixed issues:

- https://github.com/CesiumGS/cesium/issues/10649 
- https://github.com/CesiumGS/cesium/issues/9767
- https://github.com/CesiumGS/cesium/issues/11705

Related issues: 

- https://github.com/CesiumGS/cesium/issues/378 appears to be obsolete
- https://github.com/CesiumGS/cesium/issues/2521 may need clarification about the expected behavior
- https://github.com/CesiumGS/cesium/issues/9945 may also need clarification (What should Cesium display for a "backspace"?)

Caused issues: 

- It will certainly cause _some_ issues. We'll see...

## Testing plan

Until now, I only tested it **visually**. The following is a sandcastle that is based on https://github.com/CesiumGS/cesium/issues/10649 , with some additional cases: 

https://sandcastle.cesium.com/index.html#c=zZbbattAEIZfZdBNZLDWpx5dx7R1Cr1oKaSmvRGYlTRWlqx3ze7Krhv8DoWQi0IP1+0L6XU6smK3pCkB2S4eyUYzO/9ovrWEZ8YNzATO0cAxKJzDAK3IJuzdKuaHXrzyB1o5LhSa0Ks9CVWpYKNy8b1IUnTkGUyE26Qy6xYSWSLsVPIFlQ89pRWGHhXYlEDlhBNoGU8S/yJUAFNtKaJVd93KgBtHV1x12NjoyQmmBtH6zTo0a/VCEQkpI81N0oVVBQAx4SluCsyNcDjED26oB1zNuPXLLKCO4IY5PQ0C+LeFXn0tvoCIx+ep0ZlKBlpq87vjwmOnL05gWSbXrkU25pL6aq/cJX0vazvbjKDJmmSdynviBcFRQEaIaw3cjXhNtrwLMdbKOkB6DHYI+2ALWBiNRgGUx9F+mHdI+rg6KWEG688+OHdH2WpvQ0l24Hj3t3k39/dq7pDwUXXC59o5PYEieuiY7VZ1zFORnrnA6UDi2HXhNbeO/nv9/FN+mV/l32tdeIlSash/5l/yb/nn/BJ867KE2qa1/AdFr/KvB79D97Z4EFDqObgzYYFODnbKYwSt5OLgqR9Wp4ZDh+tUn7a8Z5Ge4f//ScdyMdQ+FjGv7vVW82h/fbOnYjLVxkFmpM9Yw+GExlTagUaUxec0zsbWFsIitdf4U9pLxAxEcnzLbAyx5NbSyjiT8q34SMNuv9eg/L+kUvNEqPTNDA0Nx0XaWav/qgwyxnoNcm9XOq1lxM2Nyr8A

The comparison of the current state and the state after this PR is shown here:

Cesium MeasureText Test

It shows that the missing underlines and the missing (?) leading spaces are now included.

The following is a comparison of the current state and the new state for the example from https://github.com/CesiumGS/cesium/issues/11705 :  

Cesium MeasureText Comparison 11705

It shows that the background is no longer longer than the rendered text. 

The following is a comparison of the output of [the "Labels" Sandcastle](https://sandcastle.cesium.com/index.html?src=Labels.html) for the old and the new state:

Cesium MeasureText Comparison

It shows that there are _small_ differences in the positioning, but ... it's nearly impossible to say what is "right" or "wrong" here. The only thing that bugs me a bit is that in some cases, it looks like the text is not properly aligned to the baseline:

Cesium MeasureText Alignment

After looking at the approach of the `LabelCollection` and its "Glyph-based" rendering, I wonder whether this can be avoided in all cases. (Subjectively, in other cases, the text appears _less_ jagged with the new state, so...)

Beyond that, I'm reasonably sure that some of the automated tests **will** fail with the new state, e.g. some [not-really-but-somehow pixel-exact check in `writeTextToCanvasSpec`](https://github.com/CesiumGS/cesium/blob/31293c147e9e14a32ba3d6e04662198e5874fed4/packages/engine/Specs/Core/writeTextToCanvasSpec.js#L27) or so. But this will have to be investigated. Some tests can probably be fixed pragmatically by adjusting them to "the new truth"...

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/writeTextToCanvas.js
+++ b/packages/engine/Source/Core/writeTextToCanvas.js
@@ -3,98 +3,6 @@ import defaultValue from "./defaultValue.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 
-function measureText(context2D, textString, font, stroke, fill) {
-  const metrics = context2D.measureText(textString);
-  const isSpace = !/\S/.test(textString);
-
-  if (!isSpace) {
-    const fontSize = document.defaultView
-      .getComputedStyle(context2D.canvas)
-      .getPropertyValue("font-size")
-      .replace("px", "");
-    const canvas = document.createElement("canvas");
-    const padding = 100;
-    const width = (metrics.width + padding) | 0;
-    const height = 3 * fontSize;
-    const baseline = height / 2;
-    canvas.width = width;
-    canvas.height = height;
-
-    const ctx = canvas.getContext("2d");
-    ctx.font = font;
-    ctx.fillStyle = "white";
-    ctx.fillRect(0, 0, canvas.width + 1, canvas.height + 1);
-
-    if (stroke) {
-      ctx.strokeStyle = "black";
-      ctx.lineWidth = context2D.lineWidth;
-      ctx.strokeText(textString, padding / 2, baseline);
-    }
-
-    if (fill) {
-      ctx.fillStyle = "black";
-      ctx.fillText(textString, padding / 2, baseline);
-    }
-
-    // Context image data has width * height * 4 elements, because
-    // each pixel's R, G, B and A are consecutive values in the array.
-    const pixelData = ctx.getImageData(0, 0, width, height).data;
-    const length = pixelData.length;
-    const width4 = width * 4;
-    let i, j;
-
-    let ascent, descent;
-    // Find the number of rows (from the top) until the first non-white pixel
-    for (i = 0; i < length; ++i) {
-      if (pixelData[i] !== 255) {
-        ascent = (i / width4) | 0;
-        break;
-      }
-    }
-
-    // Find the number of rows (from the bottom) until the first non-white pixel
-    for (i = length - 1; i >= 0; --i) {
-      if (pixelData[i] !== 255) {
-        descent = (i / width4) | 0;
-        break;
-      }
-    }
-
-    let minx = -1;
-    // For each column, for each row, check for first non-white pixel
-    for (i = 0; i < width && minx === -1; ++i) {
-      for (j = 0; j < height; ++j) {
-        const pixelIndex = i * 4 + j * width4;
-        if (
-          pixelData[pixelIndex] !== 255 ||
-          pixelData[pixelIndex + 1] !== 255 ||
-          pixelData[pixelIndex + 2] !== 255 ||
-          pixelData[pixelIndex + 3] !== 255
-        ) {
-          minx = i;
-          break;
-        }
-      }
-    }
-
-    return {
-      width: metrics.width,
-      height: descent - ascent,
-      ascent: baseline - ascent,
-      descent: descent - baseline,
-      minx: minx - padding / 2,
-    };
-  }
-
-  return {
-    width: metrics.width,
-    height: 0,
-    ascent: 0,
-    descent: 0,
-    minx: 0,
-  };
-}
-
 let imageSmoothingEnabledName;
 
 /**
@@ -168,7 +76,15 @@ function writeTextToCanvas(text, options) {
   canvas.style.visibility = "hidden";
   document.body.appendChild(canvas);
 
-  const dimensions = measureText(context2D, text, font, stroke, fill);
+  const metrics = context2D.measureText(text);
+  const dimensions = {
+    width: metrics.width,
+    height: metrics.actualBoundingBoxDescent + metrics.actualBoundingBoxAscent,
+    ascent: metrics.actualBoundingBoxAscent,
+    descent: metrics.actualBoundingBoxDescent,
+    minx: metrics.actualBoundingBoxLeft,
+  };
+
   // Set canvas.dimensions to be accessed in LabelCollection
   canvas.dimensions = dimensions;
 
--- a/packages/engine/Source/Scene/LabelCollection.js
+++ b/packages/engine/Source/Scene/LabelCollection.js
@@ -23,6 +23,20 @@ import GraphemeSplitter from "grapheme-splitter";
 // not have a billboard, depending on whether the texture info has an index into
 // the the label collection's texture atlas.  Invisible characters have no texture, and
 // no billboard.  However, it always has a valid dimensions object.
+//
+// A valid 'dimensions' object is an object that has the properties that are
+// derived from a https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics
+// object (in pixels)
+// - width: The width of the glyph, as given by TextMetrics.width
+// - ascent: The the distance from the 'textBaseline' attribute to
+//   the top of the bounding rectangle of the glyph, as given by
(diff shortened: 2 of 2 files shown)


ggetz:
Thanks @javagl! Great to removal of redundant code.

>  it looks like the text is not properly aligned to the baseline

I believe the baseline issue is unrelated to this change. Rather, this is caused by quantization error as described in https://github.com/CesiumGS/cesium/issues/8474.

Is this PR ready for official review?

javagl:
One of the reasons why this is still a "draft" is that there's still a
>TODO See if baseline jagging can be avoided
in my local TODO list. I'll read the linked issue(s) and see whether the change here did really make it "worse", or whether it is now (equally 'bad' but) only 'slightly different'. A single pixel can be a _lot_ for text...

javagl:
Based on a quick comparison with the sandcastle in the linked issue, the jagging seems to have become worse for small fonts:

Cesium Jagged Baseline

It is not "terrible", and only happens for fonts that are ... well, nearly too small to read them anyhow, but ... I'd still like to investigate that further (with the goal to either 1. not make it worse or 2. (preferred) make it _better_), but I'll have to zoom into the exact place where the glyphs are generated, look at the exact differences between the old and the new function, and try to figure out where this difference comes from.

javagl:
I did zoom a little bit into that, but don't have a perfect solution (yet?). The following is rather a few "notes" to keep track of the considerations here.

The most critical part of the "metrics" that are computed for the text are the `height` and the `descent`. The original `measureText` function relied on "counting filled pixels", which is brittle in some way. The updated `measureText` functionality derived the required information from the `TextMetrics`, using `TextMetrics.actualBoundingBoxDescent` and similar properties, which solved the main part of https://github.com/CesiumGS/cesium/issues/10649 

But **both** approaches cause "jagged" lines in some cases. 

The `height/ascend` properties are used for "assembling" the actual label text, out of individual glyphs. The positioning of the glyphs is computed by mushing together the _height_ of each glyph and a value that was computed from _all_ glyphs, (Details are omitted here. Reverse-engineering the [details of `LabelCollection#repositionAllGlyphs`](https://github.com/CesiumGS/cesium/blob/dcb935026be981fc968651ce045eb900fff758b1/packages/engine/Source/Scene/LabelCollection.js#L436) could be a challenge...)

For a combination of letters like `x`, and `d` (ascender), and `p` (descender), there is no common "height" or "y-value" that can reliably be computed from the bounds of _individual_ letters.

This can be solved, to some extent, by not using the `actualBoundingBoxDescent`, but the **`font`**`BoundingBoxDescent`. This is a "global" property for the font that is used for rendering. And this actually improves things _a lot_. Here is an example of the _current_ state, and the state that uses the `font...` family of properties:

Cesium Jagged Baseline No More

The current one looks jagged. The new one looks nice and clean.

But... as the name suggests, the `fontBoundingBoxDescent` property does not take into account the actual bounds. This means that some of the ~"more unusual" characters, like ꧁Javanese rerenggans꧂ or combined unicode characters like À̴̴̴̴̴̖̖̖̖̖̀̀̀̀  are not handled properly. It can be seen in this sandcastle (linked below) that all other cases work well (and there is no jagging for smaller fonts), but these "unusual" characters are not displayed properly, because the `actual....` height is larger than the `font...` height:

Cesium Font Bounds

Rough ideas for addressing this:

- One could slam in some `height = max(font..., actual....)` and hope that this works. But this is unlikely, because it will re-introduce the problem of glyphs having different heights, which caused the "jagged" look
- One could try to compute the _actual_ bounds of the _whole_ text, and use this as the size for the _final_ label. But this would require a (complete understanding and) probably large rewrite/restructuring of the whole "glyph-based" rendering approach

Maybe there's an intermediate solution. My gut feeling is that this [`maxGlyphY`](https://github.com/CesiumGS/cesium/blob/dcb935026be981fc968651ce045eb900fff758b1/packages/engine/Source/Scene/LabelCollection.js#L378) could be the crux here. But that has to be investigated further.


---

The test sandcastle:

http://localhost:8080/Apps/Sandcastle/#c=1ZfNattAEMdfZdAlMliyHadfrhOaOoVSWgppaC8Cs5LG8pLVrtldyXaDoe0TFEIKhX4cQ3vuY+Sud8hTdGzHaUkTCnZS1LUlPLPzH89vJdk7OdOQcxyihk2QOIQOGp6l/suZzw2caGZ3lLSMS9SBU7kfyLnC784nX/E4QUuWxpjb81Df2LFAP+ZmINiY0geOVBIDhxKcp0BpueVofBbH7kEgAQbKkEfJ1qKUDtOWPjHZ9HtapTuYaETj1qtQr1SnipALESqm4xbMMgDwlCV4nmCoucU9HNk91WEyZ8adRwFVBBeGVQPPg6tH4FQX4gMIWbSfaJXJuKOE0r8qnlr+7qMdmMyDK2ciEzFBda3PzAmdJ5VrWwyv7tfrzaVXxPG8NY8GAS408HfAM67JPwO8vQIgdLtdD+avtZJz3luekyC9xXETlNfF2FhfhZFGqeFurfIklv9BbNxdnu+hslalMPWWG3K9sTzkLk/61rPKE9izLXjGjKX/WLd4XxwWR8XXSgseoxAKiu/Fp+JL8bE4BNfYLKaiaa74Rt6j4nPJ12djhZsAhRqC7XMD9GZgBixCUFKMS858Z3lmKDdac/n9lLMdqhz/w8vZXGHHdHr89gnLmUSDoFGjTBImzenxu5Ijr7CHgu2TNyc/Tj5cdQYJOGLpQCD0lIZM8kjFCJFKQ2oJYoj6TLOIfglNuddoY4W9yehG0XpivKfcC5zTCKfqtGcd19biqx/wdKC0hUwL1/drFum6MIKuhVm0Tw1bZGbCaWi79ru0HfMceLx5SfcHkWDG0EwvE+IFf03t3Fa7RvF/SIViMZfJ8xw1tX/TsH5j6+nc6ft+u0bm5UqrlAiZvpD5Jw

PR #11754, opened 2024-01-10
Keep voxel data in memory for picking

jjhembd:
# Description

This PR is an incremental step towards Voxel Picking. Key changes include:
- Voxel metadata is retained in CPU memory, even after it is loaded to a texture
- `VoxelTraversal` has a new helper function to look up a voxel tile in memory

Some additional changes along the way:
- Updated voxel traversal and picking-related code for clarity
- Fixed some broken specs in `VoxelTraversalSpec`

## Testing plan

See updated `VoxelTraversalSpec`. More complete testing will be performed in the target staging branch.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/SpatialNode.js
+++ b/packages/engine/Source/Scene/SpatialNode.js
@@ -353,7 +353,6 @@ SpatialNode.prototype.addKeyframeNodeToMegatextures = function (
   for (let i = 0; i < megatextures.length; i++) {
     const megatexture = megatextures[i];
     keyframeNode.megatextureIndex = megatexture.add(keyframeNode.metadatas[i]);
-    keyframeNode.metadatas[i] = undefined; // data is in megatexture so no need to hold onto it
   }
 
   keyframeNode.state = KeyframeNode.LoadState.LOADED;
--- a/packages/engine/Source/Scene/SpatialNode.js
+++ b/packages/engine/Source/Scene/SpatialNode.js
@@ -23,7 +23,6 @@ import OrientedBoundingBox from "../Core/OrientedBoundingBox.js";
  */
 function SpatialNode(level, x, y, z, parent, shape, voxelDimensions) {
   /**
-   * @ignore
    * @type {SpatialNode[]}
    */
   this.children = undefined;
@@ -35,29 +34,24 @@ function SpatialNode(level, x, y, z, parent, shape, voxelDimensions) {
   this.z = z;
 
   /**
-   * @ignore
    * @type {Cartesian3}
    */
   this.dimensions = Cartesian3.clone(voxelDimensions);
   /**
-   * @ignore
    * @type {KeyframeNode[]}
    */
   this.keyframeNodes = [];
   /**
-   * @ignore
    * @type {KeyframeNode[]}
    */
   this.renderableKeyframeNodes = [];
 
   this.renderableKeyframeNodeLerp = 0.0;
   /**
-   * @ignore
    * @type {KeyframeNode}
    */
   this.renderableKeyframeNodePrevious = undefined;
   /**
-   * @ignore
    * @type {KeyframeNode}
    */
   this.renderableKeyframeNodeNext = undefined;
--- a/packages/engine/Source/Scene/SpatialNode.js
+++ b/packages/engine/Source/Scene/SpatialNode.js
@@ -36,6 +36,11 @@ function SpatialNode(level, x, y, z, parent, shape, voxelDimensions) {
 
   /**
    * @ignore
+   * @type {Cartesian3}
+   */
+  this.dimensions = Cartesian3.clone(voxelDimensions);
+  /**
+   * @ignore
    * @type {KeyframeNode[]}
    */
   this.keyframeNodes = [];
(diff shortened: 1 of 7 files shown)


packages/engine/Source/Scene/SpatialNode.js
```diff
@@ -34,6 +34,11 @@ function SpatialNode(level, x, y, z, parent, shape, voxelDimensions) {
   this.y = y;
   this.z = z;
 
+  /**
+   * @ignore
```
ggetz, line 38: I see we already use `ignore` in the class already, but why is this used over the `@private` tag which is more typical in the rest of the codebase?

packages/engine/Source/Scene/VoxelEllipsoidShape.js
```diff
@@ -939,7 +943,11 @@ VoxelEllipsoidShape.DefaultMinBounds = Object.freeze(
  * @readonly
  */
 VoxelEllipsoidShape.DefaultMaxBounds = Object.freeze(
-  new Cartesian3(+CesiumMath.PI, +CesiumMath.PI_OVER_TWO, +Number.MAX_VALUE)
+  new Cartesian3(
+    +CesiumMath.PI,
```
ggetz, line 947: Omit the `+`, this is atypical style for number formatting.

packages/engine/Source/Scene/VoxelTraversal.js
```diff
@@ -492,42 +498,23 @@ function loadAndUnload(that, frameState) {
...
   const primitive = that._primitive;
   const shape = primitive._shape;
-  const { dimensions } = primitive;
   const targetScreenSpaceError = primitive.screenSpaceError;
   const priorityQueue = that._priorityQueue;
-  const keyframeLocation = that._keyframeLocation;
   const keyframeCount = that._keyframeCount;
-  const rootNode = that.rootNode;
+
+  // Some values are constant because of incomplete support for time-dynamic data
```
ggetz, line 505: I'm a bit confused by this comment. Which values, or types of values, can be time dynamic?

packages/engine/Specs/Scene/VoxelTraversalSpec.js
```diff
@@ -145,32 +146,26 @@ describe(
...
+      await pollToPromise(function () {
+        scene.renderForSpecs();
+        traversal.update(
+          scene.frameState,
+          keyFrameLocation,
+          recomputeBoundingVolumes,
+          pauseUpdate
+        );
+        return traversal.megatextures[0].occupiedCount > 0;
+      }).then(function () {
```
ggetz, line 162: The `then` is not needed if you are using await.

````
await pollToPromise(...);
scene.renderForSpecs();

const megatxture = ...;
expect(...);
```

packages/engine/Specs/Scene/VoxelTraversalSpec.js
```diff
@@ -145,32 +146,26 @@ describe(
...
-      traversal.update(
-        scene.frameState,
-        keyFrameLocation,
-        recomputeBoundingVolumes,
-        pauseUpdate
-      );
-      tilesInMegatextureCount = traversal.megatextures[0].occupiedCount;
-      const tileNotInQueueWhenLookingAway = tilesInMegatextureCount === 0;
-      expect(tileNotInQueueWhenLookingAway).toBe(true);
+      expect(megatexture.occupiedCount).toBe(1);
```
ggetz, line 168: Is there a test we can add to ensure the metadata values are correct?

ggetz (commented):
Thanks @jjhembd! I didn't do any manual testing of this PR, since as you stated, it will be more valuable to test in the upstream staging branch. Instead, it would be valuable if we can make sure any new functionality is unit tested.

packages/engine/Source/Scene/VoxelTraversal.js
jjhembd: On second look, my comment was incorrect. I noticed these values were always being assigned the same value in `addToQueueRecursive`, so I pulled them out as constants. But this is actually unrelated to support for time dynamic data.

As for what can be time dynamic, here is a brief explanation:
`keyframes` represent (integer) points along a time series where we have voxel samples in a time-varying dataset. `that._keyframeLocation` is the current time we want to render, which could be non-integer, requiring us to interpolate between samples in the time direction.

However, we currently don't allow any more than one `keyframe` (see [Cesium3DTilesVoxelProvider.js#L427C1-L430C4](https://github.com/CesiumGS/cesium/blob/50468896ca7d6071e0f8fb359598c03879fbf0a2/packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js#L427C1-L430C4)), so time-varying data cannot be loaded and rendered.

lilleyse: There could be other providers that allow for multiple keyframes though. Not sure if that's relevant for this PR, but just something to keep in mind.

jjhembd:
@ggetz Thanks for the feedback! I think I've addressed everything

ggetz:
Thanks @jjhembd! This is looking good to me.

PR #5138, opened 2017-03-22
LoadWithXhr for node

jimmyangel:
Added loadWithHttpRequest function which is invoked when LoadWithXhr is called from a node.js application. It uses node http(s).request instead of XMLHttpRequest which is only available in the browser.

I tested using sampleTerrain using a node.js application, which is the main reason I made this change. I do not have an extensive test suite for this functionality under node.js, since sampleTerrain is the only use case I have identified for running Cesium functions in node.js.

Jasmine test result for Core/loadWithXhr: 7881 specs, 0 failures, 6 pending specs.

All responseTypes except for 'json' will return the result buffer. Also, a compressed response converts the result buffer to ArrayBuffer.

PR #11773, opened 2024-01-18
Swap node internal packages out for `fetch` in `Resource`

jjspace:
# Description

Working on updating [`cesium-webpack-example`](https://github.com/CesiumGS/cesium-webpack-example) and creating [`cesium-vite-example`](https://github.com/CesiumGS/cesium-vite-example) I realized there are 2 main sticking points to using Cesium in various JS build environments.

1. Copying the required static files for workers (https://github.com/CesiumGS/cesium/issues/10619)
2. Marking select node modules as external, `http`, `https`, `url` and `zlib`, which are only used in `Resource.js` when on node

This PR aims to remove that second point by swapping out the "node only" logic for `fetch()` which has been enabled by default since Node v18. 

As far as I can tell I was able to mimic the existing behavior except for one issue with an error that's thrown. There's currently no Specs set up for running in a Node environment so running the smoke screen tests `Specs/test.cjs` and `Specs/test.mjs` was the only way to verify and that works correctly.

I'd really appreciate more eyes on this as I know `Resource` is a fairly core class to cesium and I don't want to negatively impact anything.

## Issue number and link

No issue

## Testing plan

- run `node Specs/test.cjs`
- run `node Specs/test.mjs`

# Author checklist

### DIFF
--- a/Specs/test.cjs
+++ b/Specs/test.cjs
@@ -1,8 +1,6 @@
 /*eslint-env node*/
 "use strict";
 
-// NodeJS smoke screen test
-
 const assert = require("node:assert");
 const {
   Cartographic,
@@ -10,16 +8,23 @@ const {
   sampleTerrain,
 } = require("cesium");
 
+// NodeJS smoke screen test
+
 async function test() {
+  console.log("start");
   const provider = await createWorldTerrainAsync();
+  console.log("loaded terrain");
   const results = await sampleTerrain(provider, 11, [
     Cartographic.fromDegrees(86.925145, 27.988257),
     Cartographic.fromDegrees(87.0, 28.0),
   ]);
+  console.log("sampled terrain");
   assert(results[0].height > 5000);
   assert(results[0].height < 10000);
   assert(results[1].height > 5000);
   assert(results[1].height < 10000);
+
+  console.log("all assertions passed");
 }
 
-test();
+test().finally(() => console.log("done"));
--- a/Specs/test.cjs
+++ b/Specs/test.cjs
@@ -11,20 +11,15 @@ const {
 // NodeJS smoke screen test
 
 async function test() {
-  console.log("start");
   const provider = await createWorldTerrainAsync();
-  console.log("loaded terrain");
   const results = await sampleTerrain(provider, 11, [
     Cartographic.fromDegrees(86.925145, 27.988257),
     Cartographic.fromDegrees(87.0, 28.0),
   ]);
-  console.log("sampled terrain");
   assert(results[0].height > 5000);
   assert(results[0].height < 10000);
   assert(results[1].height > 5000);
   assert(results[1].height < 10000);
-
-  console.log("all assertions passed");
 }
 
-test().finally(() => console.log("done"));
+test();
--- a/packages/engine/Source/Core/Resource.js
+++ b/packages/engine/Source/Core/Resource.js
@@ -2045,17 +2045,6 @@ Resource.createImageBitmapFromBlob = function (blob, options) {
   });
 };
 
-function decodeResponse(loadWithHttpResponse, responseType) {
-  switch (responseType) {
-    case "text":
-      return loadWithHttpResponse.toString("utf8");
-    case "json":
-      return JSON.parse(loadWithHttpResponse.toString("utf8"));
-    default:
-      return new Uint8Array(loadWithHttpResponse).buffer;
-  }
-}
-
 function loadWithHttpRequest(
   url,
   responseType,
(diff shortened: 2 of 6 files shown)


packages/engine/Source/Core/Resource.js
```diff
@@ -2045,17 +2045,6 @@ Resource.createImageBitmapFromBlob = function (blob, options) {
...
-    case "text":
-      return loadWithHttpResponse.toString("utf8");
-    case "json":
-      return JSON.parse(loadWithHttpResponse.toString("utf8"));
-    default:
-      return new Uint8Array(loadWithHttpResponse).buffer;
-  }
-}
-
 function loadWithHttpRequest(
```
jjhembd, line 2048: The `data` and `overrideMimeType` arguments are not used

packages/engine/Source/Core/Resource.js
```diff
@@ -2066,64 +2055,37 @@ function loadWithHttpRequest(
...
+    headers,
+  })
+    .then(async (response) => {
+      if (!response.ok) {
+        const responseHeaders = {};
+        response.headers.forEach((value, key) => {
+          responseHeaders[key] = value;
+        });
+        deferred.reject(
+          // TODO: there is not directly equivalent to http.IncomingMessage, is the full obj ok as the second arg here?
```
jjhembd, line 2069: I don't see a problem in `RequestErrorEvent` -- the second argument is simply reported

packages/engine/Source/Core/Resource.js
```diff
@@ -2241,7 +2203,7 @@ Resource._Implementations.loadWithXhr = function (
     } else if (
       (browserResponseType === "" || browserResponseType === "document") &&
       defined(xhr.responseXML) &&
-      xhr.responseXML.hasChildNodes()
+      xhr.responseXML?.hasChildNodes()
```
jjhembd, line 2206: With optional chaining here, the previous `defined` check is redundant.

jjhembd (commented):
Thanks @jjspace! This looks like a nice cleanup.
I just have a couple minor comments in the code. Also, can we now update build.js and gulpfile.js to not declare `zlib` and `url` as external?

ggetz (commented):
Thanks @jjspace! As for a testing plan, based on [the original PR where NodeJS compatibility was added](https://github.com/CesiumGS/cesium/pull/5138):

> I tested using sampleTerrain using a node.js application, which is the main reason I made this change. 

So the `test.*js` test cases are likely a good candidate for whether or not this is working as needed.

Potentially, another node script which load a 3D tileset and checks some basic properties may also be a good use case.

packages/engine/Source/Core/Resource.js
jjspace: That is intentional. They were not used before and I aimed to preserve the existing behavior as close as possible.

I don't know exactly why they weren't used in the first place or why this only transforms text and json but the goal of this PR was not to re-write the capability of this function; just swap out the components it uses to make building easier.

packages/engine/Source/Core/Resource.js
jjspace: If I drop the `defined` check that's there it would be relying on the last line to be falsey when it's `undefined`. Normally that would mean we should wrap it in a `defined` check for consistency but that doesn't equate to the same check. Not sure what the pattern/style normally is for CesiumJS, thoughts?

cc @ggetz

Specs/test.cjs
```diff
@@ -1,25 +1,30 @@
...
     Cartographic.fromDegrees(86.925145, 27.988257),
     Cartographic.fromDegrees(87.0, 28.0),
   ]);
+  console.log("sampled terrain");
   assert(results[0].height > 5000);
   assert(results[0].height < 10000);
   assert(results[1].height > 5000);
   assert(results[1].height < 10000);
+
+  console.log("all assertions passed");
```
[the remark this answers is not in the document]
jjspace: Do you have a specific reason? I feel like there should be an easy way to specifically see when it's successful. When there is no output it's not easy to tell when there was an invisible crash/failure vs a success. I'm open to reducing the number of them or changing what they say but I think it should output _something_.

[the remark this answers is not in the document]
ggetz: At the very least, it will clutter the CI logs, which we're already pretty bad about. It will fail on error if the code does not execute as excpected.

The fact that `test.cjs` it's not running on your system is a problem that we should look into, but that would a separate issue from this one.

packages/engine/Source/Core/Resource.js
ggetz: If something is changed for the downstream user, it should be noted in `CHANGES.md`.

jjspace:
@ggetz @jjhembd updated with pr comments. I wasn't sure if this counts as a breaking change or not but put it under that category anyway because it does affect build tools.

ggetz:
Thanks @jjspace! I'll take a final pass on this shortly.

> I wasn't sure if this counts as a breaking change or not but put it under that category anyway because it does affect build tools

I agree this counts as a breaking change and should be listed as such.

ggetz:
Looks good @jjspace. I have tested the packed npm module locally as well. Everything is working as expected!

Issue #11749, opened 2024-01-09
Ongoing documentation fixes


javagl:
There occasionally are "small" fixes for the generated docmentation. These range from "fixing typos" to "fixing broken links" or "fixing type annotations". These usually do not involve rigorous tests, because this _only_ refers to fixes that do _not_ affect the functionality. A corner case is that of TSDoc fixes that may affect the generated typings. These may warrant some scrutiny, but still are not covered by real specs, but maybe only by a few lines in the [`Specs/TypeScript/index.ts` file](https://github.com/CesiumGS/cesium/blob/6e8b5d8fba959dd4df92c1aa4c69c6a62900aa0d/Specs/TypeScript/index.ts).

I'd suggest listing these issues here. Everybody can add comments with things that have to be fixed, and we can try to keep track of the things that have been addressed using GitHub `[ ] Task checkboxes`.

- [x] The `Cesium3DTileset.from...` functions claim to throw a `DeveloperError` when the tileset version is not `0.0` or `1.0`. (E.g. [`fromUrl`](https://github.com/CesiumGS/cesium/blob/6e8b5d8fba959dd4df92c1aa4c69c6a62900aa0d/packages/engine/Source/Scene/Cesium3DTileset.js#L1933) and `fromIonAssetId`). 
  **Suggested fix:** This should include `1.1` as a valid version number. 
- [x] These functions are actually not throwing a `DeveloperError`, but a `RuntimeError`, as of [this check](https://github.com/CesiumGS/cesium/blob/6e8b5d8fba959dd4df92c1aa4c69c6a62900aa0d/packages/engine/Source/Scene/Cesium3DTileset.js#L2122). 
  **Suggested fix:** Considering that the developer does not have influence on whether ~"the data is valid", it should probably be a `RuntimeError` (and documented accordingly)
- [x] The `...GeometryUpdater#isDynamic` property documentation contains invalid links. (For example, the [`PolylineGeometryUpdater` documentation](https://cesium.com/learn/cesiumjs/ref-doc/PolylineGeometryUpdater.html#isDynamic), generated from [`PolylineGeometryUpdater`](https://github.com/CesiumGS/cesium/blob/9f43e81832fd90a4d3e6b67f6b168b6131bf7ce6/packages/engine/Source/DataSources/PolylineGeometryUpdater.js#L259), but also other updater classes). 
  **Suggested fix:** It should just omit the part that says
  ```
[...]

PR #11776, opened 2024-01-20
Documentation fixes 2024-01-20

javagl:
Addresses https://github.com/CesiumGS/cesium/issues/11749#issue-2072624863 (not the issue as a whole, but only the points mentioned in this comment)

The commits basically correspond to the points that are listed in the comment, each implementing the 'Suggested fix'.

For the last point (entity constructor options) I also documented `viewFrom` as `{Property | Cartesian3}`. (It could probably be a `PositionProperty`, but ... maybe that would be confusing..?)

Note: The `EntityCollection#add` function originally threw a `RuntimeError` for duplicate IDs, and now throws a `DeveloperError` (as of  https://github.com/CesiumGS/cesium/commit/80d09540657df549ec70bdd0803167f827ce8aeb ). This could be considered as a "breaking change", or.... as a bugfix: The documentation already said that it throws a `DeveloperError`, so this just aligns the implementation with the existing documentation...

Beyond that, I'm not sure whether anything here should be mentioned in the `CHANGES.md`...

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,8 @@
 - Fixed `Entity` documentation for `orientation` property. [#11762](https://github.com/CesiumGS/cesium/pull/11762)
 - Fixed an issue where `DataSource` objects incorrectly shared a single `PolylineCollection` in the `PolylineGeometryUpdater`. Updated `PolylineGeometryUpdater` to create a distinct `PolylineCollection` instance per `DataSource`. This resolves the crashes reported under [#7758](https://github.com/CesiumGS/cesium/issues/7758) and [#9154](https://github.com/CesiumGS/cesium/issues/9154).
 - Fixed a bug where transforms that had been defined with the `KHR_texture_transform` extension had not been applied to Feature ID Textures in `EXT_mesh_features`. [#11731](https://github.com/CesiumGS/cesium/issues/11731)
+- The `EntityCollection#add` method was documented to throw a `DeveloperError` for duplicate IDs, but did throw a `RuntimeError` in this case. This is now changed to throw a `DeveloperError`. [#11776](https://github.com/CesiumGS/cesium/pull/11776)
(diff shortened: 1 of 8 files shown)


ggetz:
@javagl Are all your changes in?

javagl:
@ggetz I added the change notes in https://github.com/CesiumGS/cesium/pull/11776/commits/328b12df88a5f3c3f56f4bbdaeba27ccc3e3fe75 , explicitly pointing out the `RuntimeError`-to-`DeveloperError` fix, but summarizing the others, mentioning that they may resolve potential issues with the TypeScript typings. 

An aside: Is there a "good" reason to _not_ track `Cesium.d.ts` in Git? (Besides its size ... 2MB...). It could be good to have another way of ensuring that there are not accidental changes, except for locally diff'ing an old and a new version...

Cesium TS diff

ggetz:
Thanks @javagl!

> Is there a "good" reason to not track Cesium.d.ts in Git? (Besides its size ... 2MB...). It could be good to have another way of ensuring that there are not accidental changes, except for locally diff'ing an old and a new version...

As a generality, we usually don't commit generated files. I may need help imagining the use cases for tracking the changes directly. We do review the source documentation changes.

ggetz:
Changes look good @javagl. I'll merge as soon as CI passes.

javagl:
> I may need help imagining the use cases for tracking the changes directly.

I thought that it might be tricky to foresee the implication that a change in the _JSDoc documentation_ has on the actual _bindings_ that are generated. 

Or to put it that way: The goal would be to avoid the case where a 'trivial, innocent-looking, typo-like JSDoc fix' causes a _breaking change_ for TypeScript users. 

(Right now, generating the TS bindings causes the message _"...this is likely due to invalid JSDoc"_ being printed 2716 times, and I think that nobody would notice if this suddenly turned to 2717....)

But maybe that's unlikely enough to not be an issue.

PR #11783, opened 2024-01-23
Automate CLA checking

siddheshranade:
# Description

Whenever a contributor opens a new PR on this repo, we want to check that they've signed an individual/corporate CLA, and automate writing a comment on the PR accordingly. 

[Cesium concierge](https://github.com/CesiumGS/cesium-concierge) performed this automation before, but we stopped using it and hence need to write a new GitHub Action on this repo to continue checking for CLAs.

## Issue number and link

[Issue in engineering repo](https://github.com/CesiumGS/engineering/issues/235).

## Testing plan

1. Open any new pull request on this repo.
2. This should trigger a new workflow called "CLA Checking", which should run successfully.
3. The pull request should now have a new comment written by the github-actions bot, either asking the author to sign a CLA, or confirming that they have done so.
4. If no new comment appears on a new PR the workflow has failed. If a comment appears with an error the workflow has failed.

# Author checklist


siddheshranade:
@ggetz 

This code is currently stable and configured to write comments on [this PR](https://github.com/CesiumGS/cesium/pull/11781) on any `git push`. (See [example comment](https://github.com/CesiumGS/cesium/pull/11781#issuecomment-1906998012).)

Need your help with the following to proceed forward:

A. Google Sheets

1. We need to get Ids for both the Individual and Corporate CLA Google Sheets
2. Add them to this repo's Secrets
3. Remove the hardcoded Ids from [this line](https://github.com/CesiumGS/cesium/pull/11783/files#diff-9841fbe48fe3ec651d27e14ecdd439e827478d6a3eeb9de48e32144bfefb68f8R15) and retrieve from Secrets

B. Google account configuration

1. Need to get a JSON file for a Google Service account credentials ([docs](https://github.com/googleapis/google-api-nodejs-client/tree/main?tab=readme-ov-file#service-account-credentials)) and store it in Secrets again
2. Need to retrive/parse JSON from Secrets when authorizing the Google API [here](https://github.com/CesiumGS/cesium/pull/11783/files#diff-9841fbe48fe3ec651d27e14ecdd439e827478d6a3eeb9de48e32144bfefb68f8R62)
3. Need to remove TempGoogleJSON of my personal Service Account

C. Contributors 

1. Is [this the correct link](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTORS.md) to the contributors list to add to the comment?

ggetz:
Thanks @siddheshranade!

> This code is currently stable and configured to write comments on https://github.com/CesiumGS/cesium/pull/11781

I'm curious? Why the other PR instead of this one? And I assume we want to change the trigger to an opened PR instead of `push`, correct?

> Need your help with the following to proceed forward

Sure, I'll sync up with you offline to get the proper values and I'll configure them in the repo settings.

siddheshranade:
@ggetz 

> Why the other PR instead of this one?

Just wanted to keep things on this PR clean, and make it easy to test as many times as we want. Will remove the hardcoded PR Id eventually so the comment appears on whatever PR triggers the action.

> I assume we want to change the trigger to an opened PR instead of push, correct?

Yes, correct, trigger on `push` is only temporary. If you see the code, I've added a bunch of `TODO`s for all temporary logic that we want to change before testing and merging this.

siddheshranade:
@ggetz 

After our discussion yesterday we decided to open up a ticket to get a Google Service account for Cesium, and I was to look into how to store the Google API credentials without exposing them.

I tried using a different authorization method and use a simple API key at first (docs [here](https://github.com/googleapis/google-api-nodejs-client/tree/main?tab=readme-ov-file#using-api-keys) and [here](https://cloud.google.com/docs/authentication/api-keys)), which doesn't require a Service Account, and which can be stored in the repo's Secrets. The Google API does work using just this key, but the catch is the Sheet we read from needs to be public and shared with an "anyone with the link can view" permission. Since our CLA spreadsheets are within Cesium's Google Drive I don't think we should (or even can) make them public, so this isn't useful.

I did look at the `Concierge` repo again and saw that we were using a Google Service Account there. The `.gitignore` file in that repo lists a `GoogleConfig.json` . The Google API is set up to read from a JSON using the `keyFile` param:
```
googleapis.google.auth.getClient({
        keyFile: googleConfigFilePath,
        scopes: ['https://www.googleapis.com/auth/spreadsheets']
    })
```

So this does seem like the right way to go. But I'm not able to understand how/where `GoogleConfig.json` was stored and read from in that case if it's not in the repo. I've looked up the googleapis docs, stackoverflow and chatGpt, but haven't found a valid solution/example on how to implement this.

I'll continue looking, but please meanwhile let me know if you have any suggestions, or if anyone else on the team might know how Concierge was set up to do this so we may replicate the same.

siddheshranade:
@ggetz

Ignore my previous (now deleted) comment. We can use the Google Service Account for the API by adding its JSON file as a GitHub secret, and then writing the JSON to a temporary file that's accessed at runtime. (This is how Concierge did it as well.)

1. PR is ready, please review when you can.

Once we have the Service Account from Chris (from [this issue](https://github.com/CesiumGS/requests/issues/875)):

2. Can you please share both Google Sheets to the Service Account email? Otherwise the API wont' be able to read from the sheets.

3. The workflow also won't currently work - can you please add the following Repository Secrets to this repo:

| Secret Name | Value
| -- | -- |
| `GOOGLE_KEYS` | The entire JSON from the "API Keys" tab of the Google Service Account, copy-pasted as is into this secret. |
| `INDIVIDUAL_CLA_SHEET_ID` | The Id of the Google Sheet which lists the individual CLAs. |
| `CORPORATE_CLA_SHEET_ID` | The Id of the Google Sheet which lists the corporate CLAs. |

Thanks!

.github/workflows/cla.yml
```diff
@@ -0,0 +1,26 @@
+name: CLA Checking
+on:
+  pull_request:
```
ggetz, line 3: We just want the `open` action to trigger this, right? [By default, `pull_request` includes other triggers as well](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request).

ggetz (commented):
This is looking great @siddheshranade! I'll add the secrets to Github as soon as I get the info.

.github/workflows/cla.yml
siddheshranade: Ah, my bad! I think I didn't catch this because it works similar to how we want by default:

> For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

But good to have this only when a new PR created. Fixing in next commit.

cesiumgs-admin:
@siddheshranade I was able to create a new service account, share the spreadsheets with that account, and set the repository secrets. Please confirm and let me know if we're good to go.

siddheshranade:
@ggetz 

This StackOverflow answer should help.

siddheshranade:
@ggetz 

Since you have the Service Account access can you resolve the error `github-actions` is giving above? It says:
> Error: Google Sheets API has not been used in project 698491282566 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/sheets.googleapis.com/overview?project&#x3D;698491282566 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.


[This SO answer](https://stackoverflow.com/questions/68721853/how-to-fix-google-sheets-api-has-not-been-used-in-project) should help if required.

ggetz:
@siddheshranade This should now be enabled!

siddheshranade:
@ggetz 

Thanks for your patience - everything works as expected now. See: latest [test comment on a new PR](https://github.com/CesiumGS/cesium/pull/11854#issuecomment-1961940249).

Feel free to merge as you wish.

ggetz:
Awesome! Thanks for adding this @siddheshranade!

PR #11604, opened 2023-11-03
3d tiles height reference

ggetz:
To be merged after https://github.com/CesiumGS/cesium/pull/11581

Fixes https://github.com/CesiumGS/cesium/issues/7044
Fixes #9680
CC https://github.com/CesiumGS/cesium/issues/11544

Building on this previous PR, this allows more user-friendly APIs for clamping to 3D Tiles as well as or instead of terrain by extending the `HeightReference` types:
 - `CLAMP_TO_GROUND`/`RELATIVE_TO_GROUND` now clamps to or is relative to terrain or 3D Tiles, whichever is higher
 - Added `CLAMP_TO_TERRAIN`/`RELATIVE_TO_TERRAIN` to clamp to or position relative to terrain only
 - Added `CLAMP_TO_3D_TILE`/`RELATIVE_TO_3D_TILE` to clamp to or position relative to terrain only
 
 This was to align to the possible `ClassificationType` values, `TERRAIN`, `CESIUM_3D_TILE`, or `BOTH`.

I've replaced the sandcastle for the (now outdated) `scene.clampToHeight` method of clamping, which is now [Clamp Model to Ground](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/3d-tiles-height-reference/Apps/Sandcastle/index.html?src=Clamp%20Model%20to%20Ground.html). I've also extended this and the [Clamp Entities to Ground](http://cesium-dev.s3-website-us-east-1.amazonaws.com/cesium/3d-tiles-height-reference/Apps/Sandcastle/index.html?src=Clamp%20Entities%20to%20Ground.html) sandcastle a bit to showcase available options.

Testing Plan:
- Make sure all options are working on the new sandcastles, "Clamp Model to Ground" and "Clamp Entities to Ground"
[...]

PR #11792, opened 2024-01-29
Fix #11787 – Improve scene height picking performance

### LINKED ISSUE #11787: 3D Tiles BIM sandcastle low FPS when zoomed in
### What happened?

When viewing the BIM sandcastle locally I'm getting 1-2 fps when I zoom in closer to the model

Interestingly I noticed it does settle back to 40-50 fps after a second or two but falls back down if I try to move the camera. It may be due to loading some part of the tile? or maybe something with the new shaders and rendering new parts of high detail models?

I tested this on the [deployed version of Sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20BIM.html) and I can't reproduce the slow framerate so I suspect it's something to do with the changes in the past month

### Reproduction steps

1. Load the [3D Tiles BIM](http://localhost:8080/Apps/Sandcastle/index.html?src=3D%20Tiles%20BIM.html) sandcastle locally
2. Add the debug fps viewer `
viewer.scene.debugShowFramesPerSecond = true;`
3. Zoom in close to the model

### Sandcastle example

_No response_

### Environment

Browser: Chrome
CesiumJS Version: current `main`
Operating System: Linux

ggetz:
# Description

* Pulls in some performance fixes from https://github.com/CesiumGS/cesium/pull/11604 (whichever gets merged first is fine)
* Don't pick the ground height each frame: Only when the camera moves or when additional data is loaded.
* Configure the powerplant tileset to skip camera collision as it is an "indoor tileset"

## Issue number and link

Fixes #11787

## Testing plan

Add the following line to any Sandcastle example to check the framerate:

```
```

- [ ] Open and move around updated sandcastle examples, verifying thee frame rate is not impacted and that you can navigate inside the powerplant tileset
- [ ] Verify other 3D tiles examples, such the Google examples, and confirm frame rate isn't significantly impacted

# Author checklist

- [ ]~I have updated `CHANGES.md` with a short summary of my change~ Not needed, the original feature has not yet been shipped


packages/engine/Source/Scene/Scene.js
```diff
@@ -3683,7 +3697,10 @@ Scene.prototype.initializeFrame = function () {
 
   this._tweens.update();
 
-  this._globeHeight = getGlobeHeight(this);
+  if (this._globeHeightDirty) {
```
jjspace, line 3700: sanity check. are camera move and tile loading the only things that should make this dirty? are there any other instances of models loading or moving or anything that should affect this?

jjspace (commented):
Updates look good and most of the Sandcastles run much smoother.

I did notice pretty bad framerate in the `3D Tiles Clipping Planes` and `MSAA` Sandcastles still.

Also had one small question.

packages/engine/Source/Scene/Scene.js
ggetz: I don't believe so– This is for the case where camera is being kept above terrain and 3D Tiles. In that case, either the camera moves to be above a new tile, or more data loads underneath the camera.

ggetz:
Thanks @jjspace, this should be ready to go.

PR #11810, opened 2024-02-02
Change question issue template to contact link

jjspace:
# Description

I just discovered that you actually _can_ just have a link to an external site instead of a weird mostly empty template that just includes a checkbox. It's hidden in the [settings for templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser) as a whole, not in the specific template documentation.

Won't be able to see it until this is merged but this is what it looks like:

This should hopefully further reduce the number of issues we have opened that really belong as questions on the forums


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+- name: Ask a question
+  url: https://community.cesium.com/c/cesiumjs/5
+  about: Please use the community forum for general questions about using CesiumJS. We actively monitor it and love seeing what people are working on
```
ggetz, line 5: My 
```suggestion
  about: CesiumJS is a big project, so we use GitHub for feature requests and bug tracking exclusively. Please use the community forum for general questions. The Cesium team and community actively monitor it and love seeing what people are working on!
```

Just a suggestion. I pulled some of this from my saved reply text.

jjspace: @ggetz I had trimmed it down a little bit to try and avoid having too big a chunk of text. Can definitely change to this if you want. I do agree it provides a bit more info. Up to you?

ggetz: I personally think the extra context helps. But I'm not going to bikeshed.

jjspace:
@ggetz Should be good to go now

ggetz:
Thanks @jjspace!

Issue #11814, opened 2024-02-06
Possible performance issues for model picking


javagl:
While having a _short_ look at the reasons for https://github.com/CesiumGS/cesium/issues/11811 , I think that _most_ of the time is spent in the model picking functionality. While some of this may be necessary for the new functionality and behavior of the Screen Space Camera Controller, it led me into a the rabbit hole that is the `pickModel.js` function.

There are, from a quick glance, at least two things that _might_ be costly in terms of performance.

**1.** Re-creating the indices

When a given primitive does not have its indices stored in a `typedArray`, then this array is array is re-created and re-filled (from the buffer) here: https://github.com/CesiumGS/cesium/blob/c6ce5aa89afbcd07dd7acc7144c4df97aff4ab6f/packages/engine/Source/Scene/Model/pickModel.js#L208 

There are some unknowns for me. Some of that is related to WebGL1 vs. 2, and whether outlines or wireframes are enabled. Eventually, it _might_ boil down to a trade-off between "memory requirements" and "performance". But this part may have to reviewed with some scrutiny.

**2.** Transforming the vertices, again and again...

The part that is the most expensive one is the one where vertices are fetched for each triangle, and the triangle is tested for intersections with the ray, at https://github.com/CesiumGS/cesium/blob/c6ce5aa89afbcd07dd7acc7144c4df97aff4ab6f/packages/engine/Source/Scene/Model/pickModel.js#L274 

Focussing on the approach with some pseudocode:
```javascript
for (let i = 0; i < indicesLength; i += 3) {
  const i0 = indices[i];
  const i1 = indices[i + 1];
  const i2 = indices[i + 2];

  for (const instanceTransform of transforms) {
    const v0 = getVertexPosition(..., instanceTransform, ...);
    const v1 = getVertexPosition(..., instanceTransform, ...);
    const v2 = getVertexPosition(..., instanceTransform, ...);
    
    const t = IntersectionTests.rayTriangleParametric(ray, v0, v1, v2);
    if (defined(t)) {
      if (t < minT && t >= 0.0) {
        minT = t;
      }
    }
  }
}
``` 
Verbally:

- For each triangle:
  - Fetch the vertices
  - Transform the vertices with the `instanceTransform` (done in `getVertexPosition`)
  - Check the transformed triangle for intersections with the ray

I **think** that it should be possible to write this in a different form:
```javascript
// Store rays that are transformed with the inverse of the instance transform
const transformedRays = [];
[...]

Issue #11811, opened 2024-02-05
Performance regression between 1.113 and 1.114


javagl:
There seems to be a performance regression in 1.114 compared to the previous version.

What I tested with 1.113 and 1.114:

- Run `npm run start` locally
- Open the "OSM Buildings" Sandcastle on localhost
- Add `viewer.scene.debugShowFramesPerSecond = true;` to show the FPS
- Zoom in and out a bit with the mouse wheel.

With `1.113`, the FPS remains basically constant at 59.
With `1.114`, the FPS drops to 4 ... 10 while zooming. (When not zooming, it goes back to 59)

This is apparently related to the new Screen Space Controller behavior that was implemented with https://github.com/CesiumGS/cesium/pull/11581 . When creating the OSM buildings tileset with
```
const osmBuildingsTileset = await Cesium.createOsmBuildingsAsync({
  disableCollision: true
});
```
then the FPS remains 59 even with version 1.114.

(I have not yet extensively tested it with other tilesets. Things like the Google P3DT do not seem to be affected so noticably, although a _dedicated_ analysis (whether ~"the average FPS dropped by 5%" or so) could be difficult...)

jjspace:
Possibly related to https://github.com/CesiumGS/cesium/issues/11787

ggetz:
I'm looking at this now. Basically, we were using bounding spheres rather than the bounding volumes to pre-optimize picking on tilesets. I'm working on the ray volume intersection logic and will have a PR open soon.

PR #11817, opened 2024-02-07
3D Tiles pick improvement: Check intersections with tile bounding volumes

ggetz:
# Description

This adjust the tileset pick code such that we are testing intersections with the tile content bounding volumes directly rather than the bounding spheres. 

The bounding sphere could be quite large, especially for tileset bounding region volumes, leading to need to check intersections more models directly, leading to poorer performance particularly in tilesets like OSM.

Along with @javagl's suggestions in https://github.com/CesiumGS/cesium/issues/11814 (which can go in separately), this should remedy the drop in performance seen for OSM.

## Issue number and link

At least partial addresses https://github.com/CesiumGS/cesium/issues/11811.

## Testing plan

- [ ] Test OSM performance in Sandcastle by adding:
```
```
- [ ] Spot check other 3D tilesets (like Google P3DT) to wnsure performance was not negatively affected
- [ ] Test to make sure camera collisions still work with the surface of 3D Tiles

# Author checklist

- [ ] I need to address the remaining TODO in S2 tileset bounding volumes

### DIFF
--- a/packages/engine/Source/Core/OrientedBoundingBox.js
+++ b/packages/engine/Source/Core/OrientedBoundingBox.js
@@ -14,6 +14,7 @@ import CesiumMath from "./Math.js";
 import Matrix3 from "./Matrix3.js";
 import Matrix4 from "./Matrix4.js";
 import Plane from "./Plane.js";
+import Ray from "./Ray.js";
 import Rectangle from "./Rectangle.js";
 
 /**
@@ -670,13 +671,8 @@ OrientedBoundingBox.clone = function (box, result) {
  */
 OrientedBoundingBox.intersectPlane = function (box, plane) {
   //>>includeStart('debug', pragmas.debug);
-  if (!defined(box)) {
-    throw new DeveloperError("box is required.");
-  }
-
-  if (!defined(plane)) {
-    throw new DeveloperError("plane is required.");
-  }
+  Check.defined("box", box);
+  Check.defined("plane", plane);
   //>>includeEnd('debug');
 
   const center = box.center;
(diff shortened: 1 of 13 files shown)


ggetz:
@jjspace Can you take a review pass on this?

packages/engine/Source/Core/OrientedBoundingBox.js
```diff
@@ -714,6 +710,168 @@ OrientedBoundingBox.intersectPlane = function (box, plane) {
...
+  Check.defined("ray", ray);
+  //>>includeEnd('debug');
+
+  const transformedRay = scratchRay;
+  transformedRay.origin = Cartesian3.subtract(
+    ray.origin,
+    box.center,
+    transformedRay.origin
+  );
+  transformedRay.direction = Cartesian3.clone(ray.direction);
```
javagl, line 737: transformedRay.direction = Cartesian3.clone(ray.direction, transformedRay.direction);
to avoid allocation?

BTW: I think the X/Y/Z cases below could go into some

  // Implementation based on https://education.siggraph.org/static/HyperGraph/raytrace/rtinter3.htm
  if (!updateRange(Cartesian3.UNIT_X, origin.x)) {
    return;
  }
  if (!updateRange(Cartesian3.UNIT_Y, origin.y)) {
    return;
  }
  if (!updateRange(Cartesian3.UNIT_Z, origin.z)) {
    return;
  }

but I probably overlooked something here.

ggetz: Since it's a fairly short repetition, I left it unfolded for better readability.

packages/engine/Source/Core/Ray.js
```diff
@@ -77,4 +78,33 @@ Ray.getPoint = function (ray, t, result) {
...
+  Check.typeOf.object("ray", ray);
+  Check.typeOf.object("transform", transform);
+  //>>includeEnd('debug');
+
+  if (!defined(result)) {
+    result = new Ray();
+  }
+
+  result.origin = Matrix4.multiplyByPoint(transform, ray.origin, result.origin);
+  result.direction = Matrix4.multiplyByPoint(
```
javagl, line 101: Shouldn't that be `multiplyByPointAsVector`?

ggetz:
Closing this for the time-being since the performance improvement isn't as significant as I'd hoped.

Issue #11821, opened 2024-02-09
b3dm tileset created with gltf1.0 KHR_materials_common CONSTANT technique does not show textures


deng0:
### What happened?

We have a bunch of old tilesets that show up white, when viewed in newer version of cesium:
![image](https://github.com/CesiumGS/cesium/assets/4224837/549a35f5-b78b-47a5-833b-75137d307d9a)

In old versions they were displayed correctly.

I've looked into the cesium code and found the problem:

As these b3dm/gltf1 models are not supposed to be realtime lighted, they have no normals baked in. Instead they use the CONSTANT technique of the KHR_materials_common extension:
https://github.com/KhronosGroup/glTF/blob/main/extensions/1.0/Khronos/KHR_materials_common/README.md

When the new cesium version converts these materials to gltf2 it uses the KHR_materials_unlit extension:
https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_unlit/README.md

This makes sense, but the problem is, that the KHR_materials_unlit extension only uses the baseColor of the material, which is only set from the old diffuse value. On the other hand the CONSTANT technique only uses emission and ambient values 
(at least that's how I understand the KHR_materials_common specification).

In any case the models we created, do not specify diffuse but instead use emission for the texture.

I would propose the following fix:
if CONSTANT technique was used and diffuse is not defined, use emission (or ambient) value for diffuse, which will then be used for baseColor.

I've forked gltf-pipeline and committed a suggested fix:
https://github.com/deng0/gltf-pipeline/commit/58b138d2080a3b1c9c2f99b6b03664038c06fe70

It would be great, if this or a similar fix could be added to cesium, so that our old tilesets still work with new version of cesium.

### Reproduction steps

The provided sandcastle example shows the problem with a small test tileset.

This test tileset can also be downloaded here: https://api.cesium.com/v1/assets/2455188/archives/59489/download
or here: [tileset_gltf1_constant_technique.zip](https://github.com/CesiumGS/cesium/files/14223702/tileset_gltf1_constant_technique.zip)

### Sandcastle example

[...]

PR #11825, opened 2024-02-12
Update gltf-pipeline files

javagl:
Contains:
* https://github.com/CesiumGS/gltf-pipeline/pull/655

This required some manual formatting fixes 😕 maybe this could be aligned somehow...

### DIFF
--- a/packages/engine/Source/Scene/GltfPipeline/updateVersion.js
+++ b/packages/engine/Source/Scene/GltfPipeline/updateVersion.js
@@ -1079,6 +1079,26 @@ function convertTechniquesToPbr(gltf, options) {
   removeExtension(gltf, "KHR_blend");
 }
 
+function assignAsBaseColor(material, baseColor) {
+  if (defined(baseColor)) {
+    if (isVec4(baseColor)) {
+      material.pbrMetallicRoughness.baseColorFactor = srgbToLinear(baseColor);
+    } else if (isTexture(baseColor)) {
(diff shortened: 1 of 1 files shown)


ggetz:
> This required some manual formatting fixes

@javagl Was this to ensure prettier was happy?

ggetz:
This fixes, https://github.com/CesiumGS/cesium/issues/11821, correct?

Can we add that to `CHANGES.md`?

javagl:
> Was this to ensure prettier was happy?

Yes. Apparently, `gltf-pipeline` and CesiumJS disagree here when it comes to some details:

Cesium Formatting difference

And this was also not auto-fixed by prettier (or I mistyped something - didn't investigate it further, but could do that, if desired). 

> This fixes, https://github.com/CesiumGS/cesium/issues/11821, correct?

It was supposed to, but I have to admit that I did not (yet) check it in the _specific_ sandcastle (only in local tests).

> Can we add that to CHANGES.md?

Done. The conditions are **very** specific (I could have just written "Fixed a bug with legacy B3DMs"...). But there certainly is room for debate about the "right" behavior. Some parts of the `KHR_materials_common` spec are not as 'strict' as the newer glTF specs (e.g. it does not define color spaces and such). But hopefully, the new behavior makes sense for the (rare) case that this happens in the real world...

ggetz:
Thanks @javagl! We'll look into making sure prettier version are matching to avoid the formatting issue in the future.

Issue #11816, opened 2024-02-07
Model picking does not find right intersections


javagl:
(This is a spin-off from https://github.com/CesiumGS/cesium/issues/11814#issuecomment-1932619298 , where the info that is posted here has been removed)

This model is a (low-resolution) sphere that can be put into the `/Apps/SampleData/models/CesiumAir/` for the test:

[sphere_V60_radius_5.zip](https://github.com/CesiumGS/cesium/files/14201136/sphere_V60_radius_5.zip)

Then one can run this sandcastle locally:

```javascript
const viewer = new Cesium.Viewer("cesiumContainer", {
  globe: false
});

const urlA = "../../SampleData/models/CesiumAir/Cesium_Air.glb";
const urlB = "../../SampleData/models/CesiumAir/sphere_V60_radius_5.glb";
const scene = viewer.scene;

scene.camera.position = new Cesium.Cartesian3(0.0, 0.0, 75.0);
scene.camera.direction = Cesium.Cartesian3.negate(
  Cesium.Cartesian3.UNIT_Z,
  new Cesium.Cartesian3()
);
scene.camera.up = Cesium.Cartesian3.clone(Cesium.Cartesian3.UNIT_Y);
scene.camera.frustum.near = 0.01;
scene.camera.frustum.far = 500.0;

const modelA = await Cesium.Model.fromGltfAsync({
  url: urlA
});
viewer.scene.primitives.add(modelA);  

const modelB = await Cesium.Model.fromGltfAsync({
  url: urlB
});
viewer.scene.primitives.add(modelB);  

const handler = new Cesium.ScreenSpaceEventHandler(scene.canvas);
handler.setInputAction(function (movement) {
  const ray = scene.camera.getPickRay(movement.position);
  const pickedA = modelA.pick(ray, scene.frameState, 1.0, 0.0, new Cesium.Cartesian3());
  const pickedB = modelB.pick(ray, scene.frameState, 1.0, 0.0, new Cesium.Cartesian3());
  console.log("pickedA: "+pickedA);
  console.log("pickedB: "+pickedB);
  
  if (Cesium.defined(pickedA)) {
    viewer.entities.add({
      position: pickedA,
      point: {
        pixelSize: 8,
        color: Cesium.Color.YELLOW,
        disableDepthTestDistance: Number.POSITIVE_INFINITY,
      },
    });
  }
  if (Cesium.defined(pickedB)) {
    viewer.entities.add({
      position: pickedB,
      point: {
[...]

Issue #11812, opened 2024-02-05
Screen Space Camera Controller glitches in version 1.114


javagl:
There are some cases where new behavior of the screen space camera controller that was [implemented in 1.114](https://github.com/CesiumGS/cesium/blob/main/CHANGES.md#1114---2024-02-01) causes problems with the navigiation.

I noticed this with one particular data set. It's not clear whether this data set can be shared. I'll share a link internally, and try to update this issue when I find a case where it happens for other data sets. (Note: It might also be that something is (vaguely:) "wrong" with the data itself. But I couldn't imagine what this could be, and in any case, it worked as expected with 1.113).

When serving this data set on localhost and loading it in a Sancastle (shown below), then it will look at a certain set of buildings, and it should be possible to zoom to the buildings by right-dragging the mouse down or using the mouse wheel.

With 1.113, this works as expected
With 1.114, it seems to prevent the zooming and afterwards behave erratically when trying to rotate and zoom...

![Cesium 114 glitch](https://github.com/CesiumGS/cesium/assets/5597569/af790b4d-301e-4d26-a681-23f8108b1270)

Disabling the collision checks with `disableCollision: true` resolves this.

---

```
const viewer = new Cesium.Viewer("cesiumContainer");

const tileset = viewer.scene.primitives.add(
  await Cesium.Cesium3DTileset.fromUrl(
    "http://localhost:8003/tileset.json", {
    debugShowBoundingVolume: true,
  })
);

viewer.scene.camera.setView({
  destination: new Cesium.Cartesian3(1333182.0955672108, -4658338.042458538, 4137747.1821384085),
  orientation: new Cesium.HeadingPitchRoll(0.5789641954988802, -0.7513114431912946, 6.283165796771483),
});
```

javagl:
I'm currently playing with ~"automatic generation of artificial tilesets" (some spare-time stuff that derailed a bit), and I noticed a case where some of the glitches can be reproduced in a _relatively_ isolated and _reproducible_ (!) fashion. This test case is here:

[Cesium 114 interaction tiles-test-2024-02-11.zip](https://github.com/CesiumGS/cesium/files/14230778/Cesium.114.interaction.tiles-test-2024-02-11.zip)

The archive contains a tileset that is essentially a 4-level quadree of a 100mx100m plane, and a sandcastle for showing it. The interaction is:

- Zooming in by right-dragging the mouse down
- Rotating by middle-dragging the mouse up

[...]

PR #11830, opened 2024-02-14
Refactoring for `pickModel.js`

javagl:
There are some issues that are related to model picking, most of them boiling down to the `pickModel.js` functions. These became more apparent with the changed behavior of the screen space camera controller that was introduced in Cesium 1.114. This change caused the `pickModel.js` functions to be called _many_ times, and the behavior was crucially dependent on the correctness of the results. Some of these issues are 

- https://github.com/CesiumGS/cesium/issues/11811
- https://github.com/CesiumGS/cesium/issues/11812
- https://github.com/CesiumGS/cesium/issues/11814
- https://github.com/CesiumGS/cesium/issues/11816

This PR attempts to refactor the `pickModel.js` functions to either resolve/alleviate these issues, isolate their causes, and/or make fixing them a bit easier. 

**NOTE:** The current state is an _intermediate_ state, containing some debug output and compile-time flags. There are a few places marked with `TODO` where the correctness of the refactored state has to be verified, or where specific reasons for some of the issues have been identified but not fixed yet.

(BTW: The build failure is only due to TSDoc issues - this can trivially be resolved. The unit tests are still passing, apparently)

### Changes until now

The change started with breaking the original `pickModel` function into smaller pieces. These should not include any changes in the behavior or approach. The was only intended to have functions that may be easier to test and modify.

The main actual _changes_  for now try to address the performance issues. The old behavior for this part can be restored via the `DEBUG_USE_ORIGINAL` compile-time flag.

**Old approach**

The original approach for finding intersections is now carved out into a function `computeClosestOriginal`. This approach was as follows:

- For each instance transform `t`
  - For each triangle (v0,v1,v2)
    - Transform the triangle vertices with `t`  (done on-the fly in `getVertexPosition`)
    - Test the transformed triangle for intersection with the `ray`

This causes _all_ vertices to be transformed _many_ times (once for each instancing transform).

**New approach**

The new approach is in the `computeClosestWithTransformedRays` function. It works as follows

- For each instance transform `t`
  - compute a `transformedRay` that was computed as `t^-1 * ray`
- For each triangle (v0,v1,v2)
  - Test the triangle for intersections with each `transformedRay`

Now, the vertices do not have to be transformed any more.

---

A comparision of a very crude "benchmak" (details below), with ballpark estimates for picking times:

- `TOTAL took 16198.300000000745ms for 81 hits` with current `main`
- `TOTAL took 16834.5ms for 81 hits` with the refactored state, **without** the new approach
- `TOTAL took 10083.099999999627ms for 81 hits` **with** the new approach

---

### Current state

The original `pickModel` function is wrapped into one that prints timing- and debug information. (Before merging, the lines that contain `debugInfo` may simply be removed, but _maybe_ something like that could be useful later, maybe via some compile-time flag?). The debug information just collects info about how many nodes/primitives/indices have been considered during the picking. This may be an output like
```
{
  "numNodes": 10,
  "nodeInfos": [
    {
      "numTransforms": 1,
      "numPrimitives": 1,
      "numCheckedPrimitives": 1,
      "primitiveInfos": [
        {
          "numIndices": 372006
        }
      ]
    },
    ...
  ],
  "totalNumIndices": 3720060
}
took 125.5ms
``` 
The idea is that this could give valuable hints about **why** a certain pick operation took as long as it took.

Here is an archive that contains a Sandcastle for picking tests, and a few test models:

[Cesium 114 picking issues PR 2024-02-14.zip](https://github.com/CesiumGS/cesium/files/14284008/Cesium.114.picking.issues.PR.2024-02-14.zip)

The archive also contains a `README.md` explaining some details about the test models. 

The sandcastle can load one of these models, and allows "debug picking". It also contains some sort of crude "benchmark". For example, when loading the `plane-xz-p-n-250x250-times-10.glb` test model and then **RIGHT** clicking once, it will send many picking rays through the model and print the accumulated timing information. 

### Further steps

The main change until now was to avoid transforming the vertices, and instead, transforming the picking ray. The function `getVertexPositionRaw` is obtaining the vertices in their _un_-transformed form. **But** this function is still doing dequantization and other things. Considering that each vertex is contained in (on average) six triangles, this will still do a lot of work _multiple times_. It may be worthwhile to consider doing this only _once_, and storing the "fully decoded" vertices, before going through the indices/triangles and doing the intersection tests. 

I have not yet created test models that contain quantization. One reason for that is that I'd do this with glTF-Transform, and this always uses interleaved buffers, which don't work at all right now - see below. As soon as this issue is resolved, further test models with quantization can be created, to perform the appropriate benchmark runs and comparisons.

## Testing plan

As mentioned above, the current state is a **DRAFT**. Aspects that require review and further tests are marked with `TODO` in the current code. This includes...

- Models with _interleaved accessors_ do not work. The reasons are pointed out in https://github.com/CesiumGS/cesium/issues/11816#issuecomment-1939803607 , and the archive contains two test models for that: One "unit square" with standard accessor layout, and once with _interleaved_ accessor layout
- The new approach for the intersection tests will transform the _ray_, and not the _vertices_. One place in the code is marked with a `TODO` to point out that the "vertical exaggeration" may have to be taken into account here (I haven't yet thought about the details _how_ this could be done)
- Models that use `EXT_mesh_gpu_instancing` cannot be picked. The point for addressing this is in the part that is now carved out into a function called `computeInstancesTransforms`. 

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Model/pickModel.js
+++ b/packages/engine/Source/Scene/Model/pickModel.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
 import AttributeCompression from "../../Core/AttributeCompression.js";
 import BoundingSphere from "../../Core/BoundingSphere.js";
 import Cartesian3 from "../../Core/Cartesian3.js";
@@ -152,7 +153,7 @@ function computeInstancesTransforms(
   runtimeNodeMatrices,
   usesWebgl2
 ) {
-  const node = runtimeNode;
+  const node = runtimeNode.node;
   const instances = node.instances;
   const computedModelMatrix = runtimeNodeMatrices.computedModelMatrix;
   if (!defined(instances)) {
@@ -283,6 +284,7 @@ function obtainPickedGeometry(primitive, usesWebgl2) {
   if (!defined(vertices)) {
     const verticesBuffer = positionAttribute.buffer;
 
+    // XXX TODO This does not work for interleaved data!!!
     if (defined(verticesBuffer) && usesWebgl2) {
       const elementCount = vertexCount * numComponents;
       vertices = ComponentDatatype.createTypedArray(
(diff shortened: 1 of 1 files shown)


javagl:
I mentioned

> The new approach for the intersection tests will transform the ray, and not the vertices. One place in the code is marked with a TODO to point out that the "vertical exaggeration" may have to be taken into account here (I haven't yet thought about the details how this could be done)

and I've accidentally been thinking about this. A relatively straightforward approach could be to apply ... the inverse of the exaggeration (a ... belittling... or floccinaucinihilipilification ... ?) to the ray before transforming it with the actual inverse transform. The difficult point here is that the exaggeration is non-linear (it is linear _along the normal_ (at least, I think that there is no such thing like a 'logarithmic exaggeration', even though that could be interesting as well), but it is not linear on the scale of the globe). As a last resort, there would always be the 'safe fallback' like `if (verticalExaggeration !== 1.0) useTheOldApproach();`...

ggetz:
@javagl I think the proposed approach could be used with vertical exaggeration if both the ray and the position are transformed into the vertically exaggerated ellipsoid space. However, this will still end up applying a transform to each vertex, which I assume would nullify the optimization.

As you said above, we could do this only when `verticalExaggeration !== 1.0`. While still slower in the vertical exaggeration case, transforming both the ray and position to scaled ellipsoid space will avoid too much branch code.

ggetz:
Actually, upon further review, that is only possible for geocentric exaggeration, not geodetic exaggeration as vertical exaggeration works elsewhere.

ggetz:
@javagl I'm going to go ahead and close this PR for now. Some of these improvements were integrated with https://github.com/CesiumGS/cesium/pull/11842, and we'll want to make sure we have the proper time to go through the rest.

Thanks!

Issue #10576, opened 2022-07-20
Viewer.destroy not checking if ScreenSpaceEventHandler is defined


TJPovey:
Sandcastle example:
https://sandcastle.cesium.com/#c=hVA9a8MwEP0rhztYhiDTuY4ppIUOgQ6BTl4U6ZqKyJIrnRzc0v9eOUogoYFqu/dx7+mks4Fg1HhAD0uweIAVBh17/nbEWFfI47xyloS26LuieuhsZ7OHB+kR7WYQEp9HtPQirDIJVxjIu4llsXIy9onlnxH9tEGDkpxn5d1JlsPKiguljmvWOhCmNFZKo+W+XACrYNnCd2cBZCrtDHLjduyfGjo85QRUrJq7wOmzVwV/cstiUTSBJoPtrJvfo+4H5wmiN4zzmrAfjCAM9TbKPRKXIeSlAE19aW2UHkGr5Y3zgTQihMS8R2M2+gu7om3qpP9jNU4obXevI3ojpln2cd+uM8g5b+o03naSc2YrUtiZ2UYiZzN5dfMkOV0IMtDUWXt2Xuz/BQ

If the viewers ScreenSpaceEventHandler has been destroyed prior to the viewer being destroyed, unminifed cesiumjs is throwing a developer error:
```The object is destroyed...```
The viewers destroy method is not checking if the handler has been destroyed prior to using it, like it is for other fields:
https://github.com/CesiumGS/cesium/blob/51f3c656e638dd2ead7a2e297f8a391bcb75b915/Source/Widgets/Viewer/Viewer.js#L1681

Happy to create a PR or remove the issue if it is desired behaviour. Although it does look like the viewer is only removing two input actions and not actually destroying the handler:
```
  this.screenSpaceEventHandler.removeInputAction(
    ScreenSpaceEventType.LEFT_CLICK
  );
  this.screenSpaceEventHandler.removeInputAction(
    ScreenSpaceEventType.LEFT_DOUBLE_CLICK
  );
```

ggetz:
@TJPovey  Thanks for the report! Though I think there was an underlying assumption that the screenSpaceEventHandler wouldn't be destroyed separately, I don't see why a simple verification in the destroy method would hurt.

nayangoyal:
Hi, @ggetz can you assign this issue to me?

ggetz:
[...]

PR #11840, opened 2024-02-19
Destroy event handler

huffmantayler:
Updated Viewers destroy method to check if screenSpaceEventHandler is defined before removing.

Fixes https://github.com/CesiumGS/cesium/issues/10576

### DIFF
--- a/packages/widgets/Source/Viewer/Viewer.js
+++ b/packages/widgets/Source/Viewer/Viewer.js
@@ -1718,15 +1718,17 @@ Viewer.prototype.isDestroyed = function () {
  */
 Viewer.prototype.destroy = function () {
   let i;
-  if(defined(this.screenSpaceEventHandler)){
+  if (
+    defined(this.screenSpaceEventHandler) &&
+    !this.screenSpaceEventHandler.isDestroyed()
+  ) {
     this.screenSpaceEventHandler.removeInputAction(
-    ScreenSpaceEventType.LEFT_CLICK
+      ScreenSpaceEventType.LEFT_CLICK
     );
     this.screenSpaceEventHandler.removeInputAction(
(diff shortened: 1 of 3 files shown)


packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -1718,13 +1718,15 @@ Viewer.prototype.isDestroyed = function () {
  */
 Viewer.prototype.destroy = function () {
   let i;
-
-  this.screenSpaceEventHandler.removeInputAction(
+  if(defined(this.screenSpaceEventHandler)){
```
ggetz, line 1721: A more thorough fix would also be to check if the screenspace event handler is defined, but destroyed:

```suggestion
  if(defined(this.screenSpaceEventHandler) && !this.screenSpaceEventHandler.isDestroyed()){
```

ggetz (changes requested):
Thanks for the PR @huffmantayler! I can confirm we have a CLA on file for you.

I have one suggestion to make your fix more thorough. 

Besides that, I noticed CI is failing due to the code formatting check. To fix this, you can locally run `npm run prettier` which should format the code.

packages/widgets/Source/Viewer/Viewer.js
huffmantayler: Thanks for the suggestion @ggetz! I've updated and run `npm run prettier` as well.

ggetz:
Fantastic, thanks @huffmantayler! I've gone ahead and tested this with the Sandcastle example in https://github.com/CesiumGS/cesium/issues/10576 and this works great.

Finally, please add a short note to `CHANGES.md` with a summary of the fix.

ggetz (approved):

PR #11842, opened 2024-02-19
Fix issue with pickModel for interleaved buffers

ggetz:
# Description

Fixes an issue where interleaved buffers where not being properly accounted for in `pickModel`. This manifested in strange camera controller behavior for some tilesets, as incorrect positions were being returned from the pick function.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/11816
Fixes https://github.com/CesiumGS/cesium/issues/11812

## Testing plan

1. Ensure the Sandcastle example from https://github.com/CesiumGS/cesium/issues/11816 properly pick the sphere model at all positions
2. Ensure the tileset in https://github.com/CesiumGS/cesium/issues/11812 can be zoomed in to, and that the camera cannot go beneath the tileset.

# Author checklist

- [ ] ~I have update the inline documentation, and included code examples where relevant~


ggetz:
@javagl Would you mind taking a look at this since you opened both of the original issues?

javagl:
I tried out the picking sandcastle from https://github.com/CesiumGS/cesium/issues/11816 , using the "sphere" that was attached there, as well as the "unitSquareXY.gltf" and "unitSquareXZ_INTERLEAVED.glb" that are part of the package that I attached in https://github.com/CesiumGS/cesium/pull/11830 , and it did find all intersections as expected.

I tried out the "New York" data set from https://github.com/CesiumGS/cesium/issues/11812 , and from a _quick_ test, the interaction seems to work as expected: It is possible to navigate and zoom (I have not maliciously tried to "break" it, but it seemed to work fine). 

(An aside: The data set there was created from a "legacy" data set using the `3d-tiles-tools` `upgrade` command. This is using glTF-Transform to upgrade B3DM to GLB. And glTF-Transform is _always_ writing out interleaved buffers. That's why this issue became only apparent with tilesets that use glTF-Transform-generated GLBs in one way or another...)

For the test data set of the "flat square" at https://github.com/CesiumGS/cesium/issues/11812#issuecomment-1937350303 it ... _mostly_ works as expected: While quickly and randomly rotating the camera, I _sometimes_ managed to move the camera (eye point) below the plane, but ... _really_ rarely, and not in a way that I could systematically reproduce, so there is nothing actionable that can be derived from that.

I had a short look at the code changes, but am not sure whether this was requested implicitly or explicitly. 

---

Aside: 

One could consider hiding ...  a few things, so that this "client code" (that just wants to access a plain indexed triangle geometry) does not have to care about quantization and byte stride. This would basically mean that the `getVertexPosition` (that now receives a _lot_ of arguments) could be replaced by some
```
class PrimitiveDataAccessor {
    constructor(primitive) {
        // Fetch quanization, normalization, byte stride info,
        // and whatever else you need...
    }

    nowReallyJustGiveMeThatFinalVertexAt(index) {
        return ... // what getVertexPosition is doing
    }
}
```
so that the client code could just do
```
const p = new PrimitiveDataAccessor(primitive);
for (let i = 0; i < indicesLength; i += 3) {
  const i0 = indices[i];
  ...
  const v0 = p.nowReallyJustGiveMeThatFinalVertexAt(i);
  ...
}
```
But that would rather be a "refactoring" than a "bugfix".

---

One thing that caught my attention was the `getBufferData` call at https://github.com/CesiumGS/cesium/pull/11842/files#diff-c25ab54b614865d4ba17f0fbc27717f5cbaa07931f69deb0b83b7d0daa1d3505R263 and the offset that is `0` in the interleaved case. I have **not** yet thought that through or tried it out, but I guess in order to _really_ check whether this is correct, it would be necessary to ...

- create a GLB where the interleaved data does **not** start with the `POSITION` (e.g. data where the `TEXCOORDS_0` appear at offset 0, and the `POSITION` appear at offset `sizeof(vec2<float>)`
- run this in a configuration where this code path is actually taken (depending on the presence of the `vertices` typedArray and such)

---

If a more thorough code review and more rigurous tests for such possible (corner?) cases was intended, drop me a note.

ggetz:
Thanks @javagl! 

Since `getVertexPosition` is a module-scoped function, I'll leave it as-is for the scope of this PR, but we'll keep your suggestion in mind.

> create a GLB where the interleaved data does not start with the POSITION (e.g. data where the TEXCOORDS_0 appear at offset 0, and the POSITION appear at offset sizeof(vec2<float>))
> run this in a configuration where this code path is actually taken (depending on the presence of the vertices typedArray and such)

This is actually the case for the unit test included here. `BoxInterLeaved` has an interleaved buffer where the normals are encoded at a `byteOffset` of 0, and the positions at 12.

ggetz:
@jjhembd Could you please give this a code review?

javagl:
Indeed, the test `returns position of intersection with interleaved model` goes this code path (and others as well). 

I think that isolating some of that functionality, _independent of the picking_ could be good, to allow tests like
`getVertexPosition returns the right position for * input`
where `*` stands for `interleaved/compressed/quantized/normalized...`. But this (as well as things like being able to load a GLB file as a `Model` _without_ having a rendering context) might have to be carvet out as dedicated refactorings, and ... I am aware of the trade-offs here.

jjhembd (approved):
Thanks @ggetz ! The code changes make sense, and both the testing Sandcastles are fixed with this PR.
Also the new spec fails before this PR, and succeeds after it.

Issue #9365, opened 2021-02-03
Camera.twistLeft and Camera.twistRight don't trigger Camera.changed


easternmotors:
I have been using `Camera.changed` when the position/orientation of the camera changes but it seems that both `Camera.twistLeft(amount)` and `Camera.twistRight(amount)` don't trigger an event (though it seems like it may rarely?).

The twist command doesn't change the position of the camera, only the orientation (from what I understand). Even though `Camera.changed` doesn't specifically mention if it only triggers on camera position change or both camera position and orientation change, `Shift + Click + Drag` (to my knowledge) only changes the camera orientation yet still triggers `Camera.changed`.

`Camera.twistLeft(amount)` and `Camera.twistRight(amount)` don't appear in the CHANGELOG but `Camera.changed` has been around for a while (Released as part of 1.23) so my thought was maybe it was introduced after the `Camera.changed` work was done?

Just wanted to verify this was a bug and not intended functionality.

Sandcastle example: https://sandcastle.cesium.com/#c=bVNBbtswEPzKQhcrQErJ11Y2mjoJUMBAgbjpSRdG3MiEKa5KUhLcon8vKSWOLVo3zszODJZizw30Egc0sAKNA2zQyq5hv0YsLZNqPG9IOy41mjK5+VLqUmcZ7NA5qWto0VTo6Ro3e65rFGAJNEFDPTaeALc3aPekhAVuEAz+7qRBMZrc6SNUvEHDP/Re2ynh4c4icFBUl3rqyCYpiyNXkJ/1gq6FzeQ6CWArrcPQf+ZUTfOMC/HQe8t3XZrewGoNf0sNUJG2pJD5IuniwlYswjb+nVayMcgdzqLvKifJ84PUggZm0X3XDk3P1XnIZS03+B5Pst67NGf5cky5hWWen6KeHnbP25+7z+NBER3uadDwCQYyB3sCt/jqInD0jdDn9gIKt/GNV4eBGxERUVYAH8lcFUcdAhh3COisgyHn1xmFTXBkO8Gx8YTPrMcFv1kIQqsXbmQ/yHejcza5TQrrjgrX4crC91U2LRn/wxmVMpY5bFrl02z20lUHdKyyNlxZkBbZ+WghZA9SrK68MKgUt9Yzr51SO/kHy2RdZF4fjSriwj/BHz0axY9Btl+utxPIGCsyf7w+6YjUCzcz5/8

Browser: Chrome 88, Firefox 82

[...]

PR #11844, opened 2024-02-21
updating Camera.changed to account for changes in roll

malaretv:
# Description

I attempted to fix the remaining issues with #9365, which I think was closed prematurely.

Current behavior: For cases where camera is not pointed straight down to surface twistLeft/Right do not cause a Camera changed event to fire.

Attempted fix: Add a similar check performed for `heading` to be also performed for `roll`. This seems to properly cause the change event to be triggered. I do not know if this is the most optimal fix, just copying what was done by others to handle this case too.

## Issue number and link

#9365 Can open a new issue if appropriate

## Testing plan

I added a test case that listens for the Camera changed event for twistLeft after first telling the camera to lookUp by 45deg. This test failed before my changes, and passed after my changes.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Camera.js
+++ b/packages/engine/Source/Scene/Camera.js
@@ -428,7 +428,6 @@ Camera.prototype._updateCameraChanged = function () {
     camera._changed.raiseEvent(
       Math.max(rollChangedPercentage, headingChangedPercentage)
     );
-    return;
   }
   if (camera._mode === SceneMode.SCENE2D) {
     if (!defined(camera._changedFrustum)) {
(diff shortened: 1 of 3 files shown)


ggetz:
Thanks @malaretv! I can confirm we have a CLA on file for you.

packages/engine/Source/Scene/Camera.js
```diff
@@ -380,25 +380,56 @@ Camera.prototype._updateCameraChanged = function () {
...
+    camera._changedRoll = currentRoll;
+  }
+  if (
+    rollChangedPercentage > percentageChanged ||
+    headingChangedPercentage > percentageChanged
+  ) {
+    camera._changed.raiseEvent(
+      Math.max(rollChangedPercentage, headingChangedPercentage)
+    );
+    return;
```
ggetz, line 431: I noticed there is no early return in the heading check previously. I believe this is because the function needs to clone the frustum before exiting, as is done below.

ggetz (commented):
Thanks for your PR! I just have one comment.

ggetz:
Looks good! Thanks @malaretv!

PR #11831, opened 2024-02-14
docs: Improve the types of Billboard&BillboardCollection.add and Label&LabelCollection.add

s3xysteak:
# Description

It improves the types of `Billboard`&`BillboardCollection.add` and `Label`&`LabelCollection.add`.

Fix a doc bug that they both do not contain `disableDepthTestDistance` in option, which should be contained.
Add types about the constructOptions of Label and Billboard. They are copy from `BillboardGraphics` / `LabelGraphics`. 

Future improvement:
I think it may be need a little discussion that The [BillboardGraphics](https://github.com/CesiumGS/cesium/blob/1.114/packages/engine/Source/DataSources/BillboardGraphics.js#L50) and [LabelGraphics](https://github.com/CesiumGS/cesium/blob/1.114/packages/engine/Source/DataSources/LabelGraphics.js#L8) already have similar type declarations. Should I
use the type declarations here directly in Billboard and Label,
or declare them in Billboard and Label and then reference them elsewhere,
or simply copy the type declarations and use them in both Label and LabelCollection?

I believe both approaches would not have any impact on the usage of the code (including TypeScript users). However, a less optimal solution might cause confusion during future maintenance.

## Issue number and link

see #11813.

## Testing plan

The PR is only about docs. Run `build-docs` and `build-ts`, and check the effect.
Before: BillboardCollection.add and LabelCollection.add have no intellisense.
After:  They should have correct intellisense. 

Run `npm run build-docs` to confirm the document should be correct to show the information.

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [x] I have update the inline documentation, and included code examples where relevant
- [x] I have performed a self-review of my code

PR #11864, opened 2024-02-27
docs: Improve the types of Billboard&BillboardCollection.add and Label&LabelCollection.add

### LINKED ISSUE #11813: [DOCS] billboardCollection.add and labelCollection.add have doc issues.
### Feature

`billboardCollection.add` and `labelCollection.add` have a same issue: the option do not contain `disableDepthTestDistance`, which should be contained.
In [BillboardCollection source code](https://github.com/CesiumGS/cesium/blob/1.114/packages/engine/Source/Scene/BillboardCollection.js#L475) you can see:
```js
BillboardCollection.prototype.add = function (options) {
  const billboard = new Billboard(options, this);
  // ...
}
```
but class `Billboard` do accept a `disableDepthTestDistance` option, you can see [class Billboard source code](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/Billboard.js#L66):
```js
// ...
// It checked if disableDepthTestDistance is defined
  if (
    defined(options.disableDepthTestDistance) &&
    options.disableDepthTestDistance < 0.0
  ) {
    throw new DeveloperError(
      "disableDepthTestDistance must be greater than or equal to 0.0."
    );
  }
//...
```
It also happened in labelCollection.

I think the problem actually caused by the type is not declared clearly. So I suggest that we can provide a param type for class `Billboard` and import the type from it to other function, like: 
```js
// in Billboard.js
/**
 * @param {{
 *   image: string | Property
 *   disableDepthTestDistance: number | Property
 *   // ...
 * }} options
 * @param {Instancetype<import('...../BillboardCollection.js').default>} billboardCollection
 */
function Billboard(){
  //...
}

// in BillboardCollection.js and any other used Billboard
/**
 * @param {Parameters<typeof import(...../Billboard.js).default>[0]} options
 */
BillboardCollection.prototype.add = function (options) {
  const billboard = new Billboard(options, this);
 // ...
}
``` 
If possible, I would be happy to provide a PR to address this issue. The PR would focus solely on modifying the comments and would not involve any code changes.

s3xysteak:
# Description

Reopen [a closed PR](https://github.com/CesiumGS/cesium/pull/11831) because the PR missed some properties.
I think I have signed the CLA this time, sorry for my previous carelessness :pray: .

It improves the types of Billboard&BillboardCollection.add and Label&LabelCollection.add.

Fix a doc bug that they both do not contain disableDepthTestDistance in option, which should be contained.
Add types about the constructOptions of Label and Billboard. They are copy from BillboardGraphics / LabelGraphics.

Future improvement:
I think it may be need a little discussion that The [BillboardGraphics](https://github.com/CesiumGS/cesium/blob/1.114/packages/engine/Source/DataSources/BillboardGraphics.js#L50) and [LabelGraphics](https://github.com/CesiumGS/cesium/blob/1.114/packages/engine/Source/DataSources/LabelGraphics.js#L8) already have similar type declarations. Should I
use the type declarations here directly in Billboard and Label,
or declare them in Billboard and Label and then reference them elsewhere,
or simply copy the type declarations and use them in both Label and LabelCollection?

I believe both approaches would not have any impact on the usage of the code (including TypeScript users). However, a less optimal solution might cause confusion during future maintenance.

## Issue number and link

close #11813.

## Testing plan

The PR is only about docs. Run build-docs and build-ts, and check the effect.
Before: BillboardCollection.add and LabelCollection.add have no intellisense.
After: They should have correct intellisense.

Run npm run build-docs to confirm the document should be correct to show the information.

# Author checklist


s3xysteak:
I checked the result of `build-docs` and `build-ts` again. There should not be any more obvious mistakes then.

packages/engine/Source/Scene/Billboard.js
```diff
@@ -22,6 +22,35 @@ import SceneMode from "./SceneMode.js";
...
 import VerticalOrigin from "./VerticalOrigin.js";
 
+/**
+ * @typedef {object} Billboard.ConstructorOptions
+ *
+ * Initialization options for the first param of Billboard constructor
+ *
+ * @property {Cartesian3} position The position of the billboard.
+ * @property {string} [id] A user-defined object to return when the billboard is picked with {@link Scene#pick}.
+ * @property {boolean} [show=true] A boolean Property specifying the visibility of the billboard.
```
ggetz, line 32:
```suggestion
 * @property {boolean} [show=true] Determines if this billboard will be shown. 
```

ggetz: For this and for all the following properties, please remove the mention of "Property".

ggetz (changes requested):
Thanks @syzdev! I have some suggestion about the description wording.

ggetz:
Great, thanks @s3xysteak! Congrats on your first PR! 🎉

PR #11866, opened 2024-02-29
GitHub issue links in WebStorm Git log

turansky:
## Overview

GitHub issue links in WebStorm Git log

---

### Definition of Done

Clickable links in WebStorm Git log 

Before - `#2236` 
After -  #2236 

Screenshot 2020-04-05 at 16 26 37

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="IssueNavigationConfiguration">
+    <option name="links">
+      <list>
+        <IssueNavigationLink>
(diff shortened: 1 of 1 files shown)


ggetz:
Hi @turansky! Most of our internal developers have moved from WebStorm to VSCode. For those still using WebStorm, does this make sense to be an individual developer setting or committed to the project?

turansky:
> For those still using WebStorm, does this make sense to be an individual developer setting or committed to the project?

Commited settings are preferred.
1. Otherwise I will modify `vcs.xml`, which already exists in repo.
2. Usually I fork Cesium from scratch and lose all IDE settings.

ggetz:
Thanks @turansky!

PR #11076, opened 2023-02-07
Voxel lighting for voxels with BOX shape

jjhembd:
This PR enables voxel lighting calculations by adding a `surfaceNormal` field to the `fsInput.voxel` struct.

The normals come from three places:
1. Shape intersection tests
2. Clipping plane intersection tests
3. Individual voxel intersection tests (as described in #11050)

For the shape and individual voxel intersections, this PR only computes normals in BOX shaped voxels. For CYLINDER and ELLIPSOID shape types, the `fsInput.voxel.surfaceNormal` struct field will still be present, but the values are currently not meaningful.

In the shader code, the data structure for intersections has changed as follows:

|   | Old code | This PR |
| --- | --- | --- |
| Ray-surface intersections | `vec2`, with `.x` being the ray distance, `.y` encoding the intersection type | `vec4`, with `.xyz` storing the surface normal at the intersection point, and `.w` the ray distance. Intersection type is encoded as the scaling of the normal |
| Ray-shape intersections | `vec2`, with `.x` being the ray distance to the shape entry, `.y` the ray distance to the shape exit | A new `RayShapeIntersection` struct, with `.entry` and `.exit` fields storing the relevant ray-surface intersections as `vec4`s |

PR #11050, opened 2023-01-25
Optimum step size for box-shaped voxels

jjhembd:
This PR implements an optimum step size for voxel raymarching, based on ray-shape intersections at each step. Partially addresses #11020, for box-shaped voxels only.

Raymarching with a constant step size will sometimes miss the edge of a voxel, leading to a fuzzy or heavily aliased appearance at voxel boundaries. Smaller constant steps can improve the image, but are less performant.

This PR computes a ray-shape intersection at each step, and advances the ray to the boundary of the next voxel. The resulting image is sharper, while still reasonably performant. Here are some performance numbers from my machine, with a large dataset rendered full-screen:

| Branch | Step size | Frame rate |
| --- | --- | --- |
| `main` | 1.0 | 25 FPS |
| `main` | 0.37 | 16 FPS |
| `voxel-resolution` | (optimal) | 22 FPS |

PR #11875, opened 2024-03-07
Optimum step size and lighting for cylinder- and ellipsoid-shaped voxels

jjhembd:
# Description

This PR implements an optimum step size calculation for cylinder- and ellipsoid-shaped voxels. This generalizes the work done in #11050 for box-shaped voxels, using a new approach based on the partial derivatives of the coordinate transform from world space to shape space.

The partial derivative approach also enables a simpler method to derive the surface normal of the current voxel. This extends the functionality of #11076 to all shape types.

## Background

Ray marching is performed in a Cartesian space, which has been scaled to a "UV" space where the range [0, 1] spans the bounds of the shape. At each step of the marching, the voxel shaders convert the Cartesian "UV" ray position into the UV coordinates of the shape, whether BOX (`convertUvToBox.glsl`), CYLINDER (`convertUvToCylinder.glsl`), or ELLIPSOID (`convertUvToEllipsoid`).

We then need to estimate two things:

1. The surface normal of the current voxel cell (at the point where the ray first entered the cell). This is needed to apply lighting effects in the rendering.
2. The optimal distance to step along the ray

The step length could simply be some fraction of the average voxel size. (This is the behavior for CYLINDER and ELLIPSOID prior to this PR.) However, we need to balance two factors:
- For sharp rendering of voxel edges, the step needs to be small enough that the ray position does not skip over any voxels. If the ray is crossing an edge or corner of a voxel, the required step could be arbitrarily small.
- For good performance, the step needs to be reasonably large.

## How surface normals and optimum step size are computed

Prior to this PR, both normals and optimum step size for BOX voxels were computed by constructing the bounding box of the current cell, and performing a new ray-shape intersection with that bounding box. This intersection yielded both the normal of the incident face, and the distance to the far side of the voxel. Using the distance to the far side as the next step length ensured that the step would be as long as possible, without skipping across any voxels.

Intersections with individual voxel cells suffer from two problems:

1. The bounding volume for CYLINDER and ELLIPSOID cells is complicated and not convex. A ray can have multiple intersections with the same cell.
3. Intersection tests suffer from precision problems as the cells become smaller.

This PR uses the partial derivatives of the coordinate transform to get a linear estimate of the surface normal and step length. 

The derivatives of the coordinate transform describe how fast the shape coordinate is changing for a small change in Cartesian position.  As an example, consider the [Jacobian matrix](https://en.wikipedia.org/wiki/Jacobian_matrix_and_determinant) of the partial derivatives of the transform from Cartesian coordinates $`(x, y, z)`$ to ellipsoidal coordinates $`(\phi, \theta, h)`$:

```math
J = \begin{bmatrix}
  \frac{\partial \phi}{\partial x} & \frac{\partial \phi}{\partial y} & \frac{\partial \phi}{\partial z} & \\
  \frac{\partial \theta}{\partial x} & \frac{\partial \theta}{\partial y} & \frac{\partial \theta}{\partial z} & \\
  \frac{\partial h}{\partial x} & \frac{\partial h}{\partial y} & \frac{\partial h}{\partial z}
\end{bmatrix}
```

This matrix is evaluated at the current Cartesian ray position, and then multiplied with the ray direction to obtain the gradient (the rate of change) of the ellipsoidal coordinates along the ray:

```math
\nabla (\phi, \theta, h) = J \vec{d}
```

Once we know how fast each ellipsoidal coordinate is changing for a small step along the ray, it is relatively simple to estimate the step length needed to reach the next cell boundary.

The surface normal at the point of entry into the voxel is aligned with the coordinate vector along which the ray last crossed a cell boundary. Conveniently, the rows of the Jacobian matrix are scaled versions of the local ellipsoidal coordinate vectors in Cartesian space. We can therefore simply normalize the appropriate row, and use the sign of the gradient, to obtain the surface normal at the point of intersection. For example, if the most-recently-crossed cell boundary (along the ray direction) is a boundary in longitude ($`\phi`$), the relevant surface normal is $`-sign(J \vec{d}) \frac{J_{123}}{|| J_{123} ||}`$

## Key code changes
- Removed the `computeApproximateStepSize` method (and associated specs) from `VoxelShape`, since all step size calculations are now done in the shader.
- Simplified `IntersectBox.glsl` to use a simpler logic, more similar to the other shapes. Note that the `intersectBox` function is no longer used for individual voxel intersections. It is only used for the initial shape intersection to find the shape bounds.
- Moved some shared utility functions into `VoxelUtils.glsl`.\
- Rewrote the `getStepSize` function in `VoxelFS.glsl` to use the same Jacobian-based method for all shape types.
- Reworked `convertUvToBox.glsl`, `convertUvToCylinder.glsl`, and `convertUvToEllipsoid.glsl` to return the Jacobian matrix of partial derivatives, along with the converted shape space coordinate. 

For the `convertUvTo???.glsl` changes: the old `convertUvToShapeUvSpace` methods converted from Cartesian space, to shape space, to a UV shape space (spanning [0, 1]), all in one method. This has now been broken into two steps: 
1. `convertUvToShapeSpaceDerivative`, which transforms the coordinate to shape space and computes the partial derivatives of the transformation.
2. `convertShapeToShapeUvSpace`, which converts from the native shape space to a UV space spanning [0, 1].

## Testing plan

Load [this local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=vVhtb9s2EP4rgj8prSunSQMMeSmWuulmIG6DOuvaxsVGS2ebG0V5JOVaGfzfdyRFmZTcNR22BUVlHu85Hu+VZFpwqaI1hc8goouIw+doCJKWefLO0OJpLzXjYcEVoRzEtNeP/pzyKJoRCdekAnHqIKOcLEBUhpjMRZHfiGJNMxCXsuJprEGR472lDMZkNQGxpinUSMdvwD8JVkMa0KykLBsXWcnAzPYupQQlB7ewUaUAOXhN8EvYFRFqORpNewdWgPkc9AOtb2j6u9Z9TpgEM7WAIi2ykKZoDgy37dMIpzlRtOA74vbgbMqn3NoxQW2AZ3GttCW+KzbARlyuIFWFGNMN5RaTGg/8GckUOPSjlOQgSLRFZ1ggMpmpJINZuZgsi8+vBPLIGxATQHCGnEqU4AmTqSAqXQ4LVrScakixXXhe8lTvIkKrp5Bpu41Lpqj2jNHWeSOWS7KCA+t1taQyMQSUbL5nDTmn/EVR8kzi1ALU2A1rAR4j2QSMbthhzND8XKKSsrUPdDD+Ivw4/q4f4b9nHohr8yD/Hcau3u+092k3qaqVnawljUGRjChyi/Tk3dXwmcebFvmq4MDV7RdAQ58heXX95vL2+MgT8AuDNbAh7k0h+hgntoHl9xjJWpnOo7g28sVFk5HaKRNNNctdXV+PbiZvRi9rTBRZ7/te6JgrSRkq3M6rluQ9atV5dGa/zRLJBhc5TA479ArpT8ZELZObUTSITvaw3GuWo0Pz10wLwATmOy5DNkZr5pAMc0zJbK85W6H0L5jTi9V/bs6WWi1zulljzp3VjnZWaziqgONkD4e265fM6pgeYtavlYVkJQpV6HxKBPxRglQvMSVw8cYfxi7FSv+WU17b1dU7hfKudXr0zc/39vPBfj6aAlhDz6ye2okNKHp+0UmxxnF7dtTs1i4f1JVWpTnbsenNOQZTOO4OP3nTgvCsyCcAWIPtylb3R558dFgw3ESP7Ub3k9974k1ZQhpqYAiiTNVbs6SmamPHOwl9o2zf0+ngLPAvui2nUvtKFmwN8Z0T/0kzbsOW8I/Ws+Z3JVLHKDZnK+B1mc90v4IsDhV0e13ryHKFMrDL3xnz3pOQLgnnOxn7yjsmYlOxDV+s93DQMrnfNl+xgqjjo0shSBV7Oj4KlnOWniMuZqAinYKHZ/g5D7WN7h8/bqK04a4sdxVyV0jxuD3+jeXfhPwbpAT8bk8MxArEewRtsGLEgXGfRE9dEWoDPiCgagGqvwN8RMB9C3DvADvIYGBBS2RfloBx7xR8hJ3kBAFufT0+buY/mvFRZ/Gl52wdcxxPX37QxV19ddI/TQ5bS3VFM9PcTjr0tI4Q/1Bljqw/ShYv+xEmB1Y1XKEfHMVCO1hRFGvUxlUP+3f/1fqxh7TpKGnz7OjbrUPYakksj+s1c1bgsdFJfKyNEm6myZs7u6EwPzTik6li2lCiLsgPRD7dIRcCgH8L9miHnbESvgV6rKGeKRrs1v7YdjtoI9N10DpaSqmw4C1J1r5iDb2Z2OTtXJAF+lVZmr7WnEa/rguaNTNjvITFr+rBiK9KFc2l+fYxmAocpvf5LzneYxi6DgQlDNu+/VEfO5oKsYb02Fbe14XIiQ53bn7Qe4i1GDtC49RLJIY5kaWYkxQsaBdAc10sMTDn81LqpoWnjfhQJ0FWqNhbpm9UZHSxVC+pANNxroYHbUGGgfKFTUIbdSbwjXwv/AYDC1gIoi9Doa6aSNFUSRUcqL6CCnaY6Iw8OdyD1UE1SYlp0bagNHvWYg86AIz9L/A/aQOc05KdQbW/4mbN/k7ao2gP9SHCwkX3cDmz5HUfTeocXsxwUeehrgCTOF+Gkxqx/XV3d65P23XgCroJKheOn5kSa7YWewcN7zCu568Y6MyQ4WvDFWN0JTGPkp9/mHz3TF9AaV7mbzE2Stn/T1jNk8OUt+7aqQC00Y2gOVUUT2H+LaV+PbC3/ZVjkVgw82INl4zFruhaW63qg3hdVB54jW8fuhpdmheHrgYky2p7+g9EATpuqopT67T51XdTfik8DUYNixcAp/6gZtjapxyziVB9bG0Ez7ZqQvIVs2WjfhbpcP5GlTJmM883tU3rva+DdxrzeDPWWiRdYwUEsw6ebczzDTbM6rYwVy3UZLJagoC4pcQsmK3f1AYDbE3oQvu6pAuDpW5bh/nO2qbfTLBTpUQqBtpjt0XBZkSMgZfxncbWHlKmq0x7TfRiGdrFTqTjBs/t015tcYwUYGiMU+9e550y2wH9tcu1y3wj3f7f0mtYMd2Txf+g1vDD9ej1y6u3D9DqRbH5HxR68eZ9Vxe8oPX6vXOpKgbPNfF7ipcYoaJSsDhJBgow4lG4HMzK9HdQSSqllnI+cJDzjK4jml3secyNUkakxJl5ydgEG/+09/x8gPwBDJuXjtU3axCMVJpl+fT5tSUmSXI+wGEXpWwEehL/Ag) and verify that all 3 shape types are rendered with sharp edges and lighting.

# Author checklist


packages/engine/Source/Scene/VoxelEllipsoidShape.js
```diff
@@ -420,9 +421,14 @@ VoxelEllipsoidShape.prototype.update = function (
     shapeMaxExtent,
     shaderUniforms.ellipsoidRadiiUv
   );
-  const axisRatio =
-    Cartesian3.minimumComponent(shapeOuterExtent) / shapeMaxExtent;
+  const { x: radiiUvX, z: radiiUvZ } = shaderUniforms.ellipsoidRadiiUv;
+  const axisRatio = radiiUvZ / radiiUvX;
   shaderUniforms.eccentricitySquared = 1.0 - axisRatio * axisRatio;
+  shaderUniforms.evoluteScale = Cartesian2.fromElements(
```
jjhembd, line 427: This `evoluteScale` used to be computed in single precision on the shader, in the `nearestPointOnRadius` function in `convertUvToEllipsoid.glsl`. It is the difference of two very similar values, so it is safer to compute it in double precision on the CPU to avoid subtractive cancellation.

packages/engine/Source/Shaders/Voxels/VoxelFS.glsl
```diff
@@ -23,30 +23,41 @@ uniform float u_stepSize;
...
+
+    float lastEntry = minComponent(distanceFromEntry);
+    bvec3 isLastEntry = equal(distanceFromEntry, vec3(lastEntry));
+    vec3 normalShapeSpace = -1.0 * vec3(isLastEntry) * directions;
+    vec3 normal = normalize(jacobianT * normalShapeSpace);
+    vec4 voxelEntry = vec4(normal, currentT - lastEntry);
+    vec4 entry = intersectionMax(shapeIntersection.entry, voxelEntry);
+
+    float firstExit = minComponent(distanceToExit);
+    float stepSize = clamp(firstExit, fixedStep * 0.02, fixedStep);
```
ggetz, line 58: Where does the `0.02` come from? Should this be configurable, or at least declared as a constant?

packages/engine/Source/Shaders/Voxels/VoxelUtils.glsl
```diff
@@ -0,0 +1,27 @@
+struct Ray {
+    vec3 pos;
+    vec3 dir;
+    vec3 rawDir;
+};
+
+#if defined(JITTER)
+float hash(vec2 p)
+{
+    vec3 p3 = fract(vec3(p.xyx) * 50.0); // magic number = hashscale
```
ggetz, line 10: I know this was existing code that was just moved to this file in this PR, but would it be possible to better explain what these magic numbers are in a comment?

packages/engine/Source/Shaders/Voxels/convertUvToCylinder.glsl
```diff
@@ -23,20 +23,37 @@
...
-    vec3 positionLocal = positionUv * 2.0 - 1.0; // [-1,+1]
+PointJacobianT convertUvToShapeSpaceDerivative(in vec3 positionUv) {
+    // Convert from UV space [0, 1] to local space [-1, 1]
+    vec3 position = positionUv * 2.0 - 1.0;
 
-    float radius = length(positionLocal.xy); // [0,1]
+    float radius = length(position.xy); // [0, 1]
+    vec3 radial = normalize(vec3(position.xy, 0.0));
+
+    // TODO: why??? Why not the local z in [-1, 1] ??
```
ggetz, line 33: Are you able to address this TODO?

packages/engine/Specs/Scene/VoxelBoxShapeSpec.js
```diff
@@ -829,43 +829,4 @@ describe("Scene/VoxelBoxShape", function () {
       );
     }).toThrowDeveloperError();
   });
-
```
ggetz, line 832: Is there a relevant replacement spec for the Jacobian approach? (It's OK if there's not.)

jjhembd: No, all step size math has been moved to the shader.

packages/engine/Source/Shaders/Voxels/VoxelUtils.glsl
jjhembd: I haven't been able to trace where this hash function came from. It is similar to one of the functions in a [popular shadertoy example](https://www.shadertoy.com/view/4djSRW), but most uses that I have seen use a different value for the hashscale. See for example the `hash12` function in [this gist](https://gist.github.com/lhog/cfc256a0486d35e7dd90bc4970e1ebc0).
I added a link to the shadertoy in a comment.

ggetz:
@jjhembd I'm seeing some artifacts in the center of the screen when zoomed out for both box and cyllinder types, but not ellipsoid:

Box:
image

Cyllinder:
image

They seem to flicker, even when the camera is not moving. Maybe some kind of z-fighting?

jjhembd:
> @jjhembd I'm seeing some artifacts in the center of the screen when zoomed out for both box and cyllinder types, but not ellipsoid:

This should be fixed. Let me know if it works on your end now!

ggetz:
Thanks @jjhembd, the cylinder is looking good.

However, I still seem to see some artifacts with the box shape. They persist despite adjust the step size in the voxel inspector, but do change appearance when the value is changed.

image

image

PR #11880, opened 2024-03-14
fix: Fix Documentation Errors

Steven4857:
This should be a doc error


ggetz:
Thanks @Steven4857! Given this is a small documentation/typo fix, we can merge without a signed CLA.

PR #11859, opened 2024-02-26
Add extensibility to various classes

jjspace:
# Description

Add new functions to register various updaters and visualizers to support custom entities and data constructs.

## Issue number and link

No issue

## Testing plan

- Run the specs and make sure they all still pass
- Try the sandcastles related to entities, primitives and CZML and make sure they all still behave as intended.

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [x] I have added or updated unit tests to ensure consistent code coverage
- [x] I have update the inline documentation, and included code examples where relevant
- [x] I have performed a self-review of my code

PR #11901, opened 2024-03-26
Add TS types for `Check`, remove circular dependency

jjspace:
# Description

Adds types for the `Check` group of functions with type assertions to help in TS based projects.
I chose to add this as a `Check.d.ts` file right next to `Check.js` to keep them closer together. `tsd-jsdoc` doesn't seem to have a way to import this types file though so it still needs a custom replace in our generate TS build step.

Also removed a circular dependency I accidentally added in #11859 

## Issue number and link

No issue

## Testing plan

- Run `npm run build-ts` and make sure it succeeds
- Check `Source/Ceisum.d.ts` and `packages/engine/index.d.ts` that they have the correct definition for `Check`
- Run `npm run build-docs` and make sure it succeeds
- Check that the Documentation now has `Check` on the globals page.

For the circular dependency you can use [`dpdm`](https://github.com/acrazing/dpdm) to check in `main` and this branch with the command `npx dpdm packages/engine/Source/DataSources/CzmlDataSource.js`

# Author checklist


gulpfile.js
```diff
@@ -1188,7 +1188,16 @@ function generateTypeScriptDefinitions(
...
-    .replace(/^(\s*)(export )?const enum (\S+) {(\s*)$/gm, "$1$2enum $3 {$4");
+    .replace(/^(\s*)(export )?const enum (\S+) {(\s*)$/gm, "$1$2enum $3 {$4")
+    // Replace JSDoc generation version of defined with an improved version using TS type predicates
+    .replace(
+      /defined\(value: any\): boolean/gm,
+      "defined<Type>(value: Type): value is NonNullable<Type>"
+    )
+    .replace(
+      /\/\*\*[\*\s\w]*?\*\/\nexport const Check: any;/m,
+      `\n${readFileSync("./packages/engine/Source/Core/Check.d.ts").toString()}`
```
ggetz, line 1199: Can this be made generic to and `**/Source/**/*.d.ts` file? That way any future additions don't need to tweak build code.

packages/engine/Source/Core/Check.d.ts
```diff
@@ -0,0 +1,127 @@
...
+ * or meet specified conditions
+ */
+export const Check: {
+  /**
+   * Throws if test is not defined
+   *
+   * @param {string} name The name of the variable being tested
+   * @param {*} test The value that is to be checked
+   * @exception {DeveloperError} test must be defined
+   */
```
ggetz, line 12: Can you explain why the inline docs need to be duplicated?

ggetz:
> I have updated CHANGES.md with a short summary of my change

Technically speaking, this should be added as an addition in `CHANGES.md`.

gulpfile.js
jjspace: I would love to do that but I couldn't find or think of a nice clean way to do that. As it is it needs to be specific to replace the definition in the types file for the corresponding code. Otherwise we'd end up with a duplicated definition which breaks TS. I'd love if `tsd-jsdoc` let you pass in a set of types files to use in places like this.

packages/engine/Source/Core/Check.d.ts
jjspace: Without them the intellisense that VSCode picks up has no description of any of the functions. The way the build is adding this it's taking the file wholesale. I was not able to find a way to get the `tsd-jsdoc` to output anything other than `const Check: any` so it also didn't include any of the descriptions other than the top one for `Check`. It's annoying and I'd love to remove the duplication but I couldn't figure out a way to do that and I didn't want to spend a ton more time on digging into it.

jjhembd (approved):
Thanks @jjspace, this looks like a reasonable workaround for now.
`ts-jsdoc` is a little messy, as we have noticed in https://github.com/CesiumGS/cesium/issues/10455, and the Check object itself could use some refactoring. But until we get to those bigger projects, this works to generate the needed TS types.

PR #11906, opened 2024-03-27
Use gpu for e2e tests

jjspace:
# Description

I was having some issues running our end to end tests where it spiked my CPU to 100% and took 20-60+ seconds _per test_ which would mean they take over an hour to complete. After this change it brings most tests down to ~2-5 seconds and only a very small number taking longer than 10 seconds which makes the whole suite take closer to 15 minutes now.

The changes I made were based on this article along with a few others I found: https://www.createit.com/blog/headless-chrome-testing-webgl-using-playwright/

## Issue number and link

No issue

## Testing plan

- Run `npm run build` to make sure it's built
- Run `npm run test-e2e-update` and `npm run test-e2e`
    - make sure they complete in a reasonable timeframe
    - Check the screenshots in `Specs/e2e/webgl-check` and make sure they show what you'd expect. For example the webgl report screenshots should not have a "Major performance caveat" as mine did before adding the chrome argument.

# Author checklist


Specs/e2e/webgl-check.spec.js
```diff
@@ -0,0 +1,58 @@
...
+const webGlReport1 = "https://webglreport.com/?v=1";
+const webGlReport2 = "https://webglreport.com/?v=2";
+const aquarium = "https://webglsamples.org/aquarium/aquarium.html";
+
+test.beforeEach(async ({ page }, testInfo) => {
+  testInfo.setTimeout(testInfo.timeout + 160000);
+});
+
+/**
+ * This is used to check how WebGL is running in the testing environment to spot things like
```
ggetz, line 20: Is this a test we should commit, or one we should use for verification in this PR and then delete? I'm leaning towards the later since this PR should remove the need to have this test.

ggetz:
The updates are working for me on MacOS. Everything is running the same as it was in main. So clear improvement for Linux.

Specs/e2e/webgl-check.spec.js
jjspace: I was a bit back and forth on it too but I thought it was more valuable to have in case we need to check in the future. I don't think it really does any harm being there. It definitely would've made my current investigation easier and now there'd be an easy thing to point to for anyone else having issues with the e2e tests?

I don't feel _super_ strongly about it and can take it out if you want

jjspace: After discussing with @jjhembd I decided to move this into the testing guide itself with a new section for e2e test performance instead of always keeping them around in an actual file.

Issue #11877, opened 2024-03-11
Camera is rotating without user interaction


javagl:
### What happened?

When loading a model (e.g. a GLB) at the origin and looking at it, then the camera moves/rotates without any user interaction.

It does not happen when `globe:false` is set.

It does happen when manually setting the camera with `camera.lookAt` (unless the target is (0,0,0)), and it also happens when setting `viewer.trackedEntity = entity;`. 

The movement seems to be a bit ... unsteady and "random". In fact, with the sandcastle below, the camera is slowly rotating, and after ~30 seconds, it screws up the orientation in a way so that the model is no longer visible... 

### Reproduction steps

Do nothing 😁 

(After opening the sandcastle below...)

### Sandcastle example

https://sandcastle.cesium.com/index.html#c=fVJRa8IwEP4roU8tlEQZe6mdTNzeBmMIewqMtD1dME0kSRU3/O+7JtbplL2Uu8t93333XWujnSdbCTuw5IFo2JE5ONm19D3UUp7UIZ8b7YXUYHmSk2+uCWFspUwFBVkK5YDrQzbhmus6MHZWIR1PKGUL0W4UPAkvWGsaUI7FCTNpj9EHhnSlKp5MBvzGOOml0Zea5sJ6jIS+S0d0lJPhk51wxkpAodfQt054sBrrl9CcjCN+YEC49HsER1NoyCU4KpomDYtr0eLWuGHeZ4PS4hSF8pmQ4jwJj8GHItpIkEn+0h3yk5NHATWOs4IqY9YznwYBtx3BfXCbLCf/9tz3fVmcwNhxhreiXkPzPOweTcCeJE9K5/cKplErIY+y3RgbLpzidT3gddFax6oOKTytnevJ+9aSnUPLRm6JbB5u/FGkVsI5fFl2Si3kF/BkWjLsv4IqIxqpV69bsErs+7bP8fQlFimlJcP0NtIboyph/zD/AA

### Environment

Browser: Firefox/Chrome
CesiumJS Version: 1.115

PR #11907, opened 2024-03-28
Solve the issue of #11877

HapDragon:
# solve the issue #11877

###11877

## Testing plan

when camera lookAt something, using transform, which will lead to recompute camera position, add a restrict in the terrain adjustment process by comparing the magnitude of the transform's origin to a height level, to filter the position which is far away from the globe surface. 

# Author checklist


ggetz:
Thanks @HapDragon! We'll review this shortly!

HapDragon:
Hi, @ggetz , like the graph below:
image

Origin P will turn to P1 after terrain adjustment, actually I am not sure whether there should add some property to specify the threshold. Or use some angle threshold to determine whether to ignore the height change, I mean if the difference between origin P centric angle and new P1 centric angle bigger than some threshold, we will ignore the height change.

Thank you for reviewing.

packages/engine/Source/Scene/ScreenSpaceCameraController.js
```diff
@@ -2888,31 +2889,38 @@ function adjustHeightForTerrain(controller, cameraChanged) {
...
-        if (mode === SceneMode.SCENE3D) {
-          ellipsoid.cartographicToCartesian(cartographic, camera.position);
-        } else {
-          projection.project(cartographic, camera.position);
+    let transformOriginLen = Infinity;
+    if (defined(transform)) {
+      Matrix4.getTranslation(transform, transformOrigin);
+      transformOriginLen = Cartesian3.magnitude(transformOrigin);
+    }
+    if (transformOriginLen > controller._minimumCollisionTerrainHeight) {
```
ggetz, line 2897: I'm not sure about the use of `controller._minimumCollisionTerrainHeight` here. That value was design to be the height above the ellipsoid surface, not the absolute value. 

Additionally, that put us in this strange "middle ground" where the camera still re-orients strangely, like in this example below:

```js
  //globe: false
});

const url = "../../SampleData/models/CesiumAir/Cesium_Air.glb";
const position = new Cesium.Cartesian3(0.0, 100000.0, 0.0);
const orientation = new Cesium.Quaternion(0.0, 0.0, 0.0, 1.0);

  name: url,
  position: position,
  orientation: orientation,
  model: {
    uri: url,
  },
});

viewer.camera.lookAt(
  new Cesium.Cartesian3(0.0, 100000.0, 1), 
  new Cesium.Cartesian3(0, 50, 0));
```

How would you feel about instead using a percentage of the ellipsoid radius? That will ensure we scale to other ellipsoids if needed without having to manually set a value.

ggetz:
Thanks for the suggestion @HapDragon! I have one comment before we proceed.

packages/engine/Source/Scene/ScreenSpaceCameraController.js
HapDragon: Hello ，
At line 2897, the comparision: controller._minimumCollisionTerrainHeight, it seemed should not be such a value. What I thought is to put a threshold to specify the maximum centric length of cartesian3, then let any point in a centric ball with the threshold radius will not do the height adjustment (if there is a transform upon the camera).

Thanks for your advice, I am a little confused about using a percentage of the ellipsoid radius.  Does it mean using percentage of ellipsoid radius as the height threshold test? 

I modify the adjustHeightForTerrain function again: it will avoid the problem in your example above and issue of #11877 as well：

```
function adjustHeightForTerrain(controller, cameraChanged) {
  controller._adjustedHeightForTerrain = true;

  const scene = controller._scene;
  const mode = scene.mode;
  const globe = scene.globe;

  if (mode === SceneMode.SCENE2D || mode === SceneMode.MORPHING) {
    return;
  }

  const camera = scene.camera;
  const ellipsoid = defaultValue(globe?.ellipsoid, Ellipsoid.WGS84);
  const projection = scene.mapProjection;


  let transform;
  let mag;
  if (!Matrix4.equals(camera.transform, Matrix4.IDENTITY)) {
    transform = Matrix4.clone(camera.transform, scratchAdjustHeightTransform);
    mag = Cartesian3.magnitude(camera.position);
    camera._setTransform(Matrix4.IDENTITY);
  }

  const cartographic = scratchAdjustHeightCartographic;
  if (mode === SceneMode.SCENE3D) {
    ellipsoid.cartesianToCartographic(camera.position, cartographic);
  } else {
    projection.unproject(camera.position, cartographic);
  }

  let heightUpdated = false; 
  if (cartographic.height < controller._minimumCollisionTerrainHeight) {  
    if((defined(transform)&&cartographic.height>-ellipsoid.maximumRadius*0.3)||!defined(transform)){
      const globeHeight = controller._scene.globeHeight;
      if (defined(globeHeight)) {
        const height = globeHeight + controller.minimumZoomDistance;
        const difference = globeHeight - controller._lastGlobeHeight;
        const percentDifference = difference / controller._lastGlobeHeight;
  
        // Unless the camera has been moved by user input, to avoid big jumps during tile loads
        // only make height updates when the globe height has been fairly stable across several frames
        if (
          cartographic.height < height &&
          (cameraChanged || Math.abs(percentDifference) <= 0.1)
        ) {
          cartographic.height = height;
          if (mode === SceneMode.SCENE3D) {
            ellipsoid.cartographicToCartesian(cartographic, camera.position);
          } else {
            projection.project(cartographic, camera.position);
          }
          heightUpdated = true;
        }
  
        if (cameraChanged || Math.abs(percentDifference) <= 0.1) {
          controller._lastGlobeHeight = globeHeight;
        } else {
          controller._lastGlobeHeight += difference * 0.1;
        }
      }
   }  
  }

  if (defined(transform)) {
    camera._setTransform(transform);
    if (heightUpdated) {
      Cartesian3.normalize(camera.position, camera.position);
      Cartesian3.negate(camera.position, camera.direction);
      Cartesian3.multiplyByScalar(
        camera.position,
        Math.max(mag, controller.minimumZoomDistance),
        camera.position
      );
      Cartesian3.normalize(camera.direction, camera.direction);
      Cartesian3.cross(camera.direction, camera.up, camera.right);
      Cartesian3.cross(camera.right, camera.direction, camera.up);
    }
  }
}

```

It use this code to filter point in deep underground areas:
`
   (defined(transform)&&cartographic.height>-ellipsoid.maximumRadius*0.3)
`

HapDragon:
Thanks for the suggestion@ggetz , I have commented on the code several months ago, it seemed could not been seen. that is to use:  `  if((defined(transform)&&cartographic.height>-ellipsoid.maximumRadius*0.3)||!defined(transform))` to filter point in deep underground areas.

PR #11929, opened 2024-04-12
Fix Label position height may not be correctly updated when its HeightReference is relative.

plainheart:
# Description

Hi,

I noticed Label position height doesn't work anymore when its `HeightReference` is relative since the commit 8e271f10e549854e9c1aec27098dec0bc3160094 ([diff](https://github.com/CesiumGS/cesium/commit/8e271f10e549854e9c1aec27098dec0bc3160094#diff-3bcb704a66884dfe84fe3b2c3590a4cbbf129d03717790337d0f74e391dd356eR1108)).

This bug might be caused by a premature assignment to `owner._clampedPosition` [1].  As the `_clampedPosition` setter of `Label` clones the inputted value, the following modification for the `owner_clampedPosition` won't be applied to the Label glyph billboard [2], which causes the user-specified height not to work as expected. So I moved the assignment to the last line to fix this issue.

[1] https://github.com/CesiumGS/cesium/blob/1.116/packages/engine/Source/Scene/Billboard.js#L1141-L1145
[2] https://github.com/CesiumGS/cesium/tree/1.116/packages/engine/Source/Scene/Label.js#L1155-L1172 

| Stage | Before | After |
| :----: | :----: | :----: |
| Initial |  |   |
| After Update | Unexpectedly clamped to ground |  |

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.117
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Fixed a bug that Label position height may not be correctly updated when its HeightReference is relative. [#11929](https://github.com/CesiumGS/cesium/pull/11929)
+
 ### 1.116 - 2024-04-01
 
 #### @cesium/engine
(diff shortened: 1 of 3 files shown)


ggetz:
Thanks for the fix @plainheart! We'll take a look at this shortly.

ggetz:
Thanks @plainheart for the fix, and for the testing Sandcastle example making this PR easy to review! I can confirm this addresses the behavior.

Issue #11488, opened 2023-08-29
Intial orientation of S2 globe is too sensitive to changes in the boundingVolume value?


pmaxarr:
Recently we had a problem that two versions of our internal S2 globe was displayed with different initial orientation by Cesium. After investigating we realized that the boundingVolume of the two version had a small decimal difference (7th decimal of PI).

Cesium seems to be sensitive to the exact value of the boundingVolume when opening our internal globe and selecting what S2 face to display as initial orientation. 
Cesium is zooming to different S2 faces when opening our internal globe depending on the root boundingVolume value defined in the tileset.json file . 
If the root boundingVolume is as follows then Africa (face 0) shown:
"root": {
    "boundingVolume": {
      "region": [
        -3.141592653589793,
        -1.5707963267948966,
        3.141592653589793,
        1.5707963267948966,
        -10881.995181892067,
        8805.044281111099
      ]
    },
here  -3.141592653589793 is identical to the definition of PI.

If the root boundingVolume is as follows then Antarctica (face 5) is shown:
"boundingVolume": {
      "region": [
        -3.141592618820418,
        -1.5707962986969973,
        3.14159264862274,
        1.5707962837028416,
        -10881.995181892067,
        8805.044281111099
      ]
    },

The difference of the bounding volume is:  
>>> -3.14159265358979 - -3.141592618820418
-3.47693718261155e-08

Is this as expected? Is it required that the boundingVolume must be identical to PI to orient the globe to display Africa (face 0) as its initial orientation?

We were suprised that our small difference in boundingVolume resulted in the dispaly of different initial S2 faces (Africa/Antarctica).

ggetz:
Hi @pmaxarr, thanks for the repot. Would you mind including a brief code snippet or sandcastle example that shows how you are constructing your scene?

By default, the home view is oriented relative your timezone. 

If you are zooming to a bounding sphere, I would not expect such a small difference in the bounding region to cause a difference in orientation.

javagl:
Attached is a file with two tilesets, `tilesetA.json` and `tilesetB.json`, which each just define the bounding volume as described in the first comment. It also contains a sandcastle for loading them and switching between them with a dropdown menu.

[S2OrientationIssue.zip](https://github.com/CesiumGS/cesium/files/12540143/S2OrientationIssue.zip)

![Cesium S2 Orientation](https://github.com/CesiumGS/cesium/assets/5597569/a83eae39-42bd-4631-a52d-b7eb62f62693)

[...]

PR #11937, opened 2024-04-16
Draft for handling epsilon-dependence of orientation

javagl:
### Description

This changes the `camera.lookAt` implementation so that the orientation does no longer depends (so strongly) on certain "epsilon" values of the target position. Some additonal details are laid out in https://github.com/CesiumGS/cesium/issues/11488 .

The fact that the title _and_ the name of the branch contain the word `draft` should be a clear indicator:

**This is not supposed to be merged!**

This should just serve as a basis for figuring out how to solve the issue _sensibly_.

As pointed out in the comments to the issue: There is no one-fits-all solution. The whole `zoomTo`-functionality just _assumes_ that everything happens on the surface of the WGS84 ellipsoid. When a tileset has a distance of (0,0,100) or (5000,0,0) from the center of the ellipsoid, then trying to compute the ENU orientation from the point _on the surface_ that is _closest_ to this point hardly makes sense. Tilesets that are not "geolocated" (but have their center at the origin) will usually not be handled gracefully by that function. But basically every Sandcastle in the world will use the `zoomTo`/`flyTo` functionality. So every change here **will** inevitably be a "breaking change" in terms of the behavior. 

That said .... what is currently implemented here will do the following:

- It will compute the distance of the `target` (i.e. the center of the bounding sphere to zoom to) to the origin.
- It will compute the distance of the eye point to the `target` (based on the `offset` that is passed in)
- It will compute a "weight" from that. When the eye is "far away" from the target, then the target position will less affect the orientiation of the camera.

In terms of "testing", I focussed on the case of a tileset with a bounding sphere that covers the whole earth, but is centered at different positions. These are created (on the fly) in the sandcastle that is given below, with the 
``` 
  createSampleOption([0, 0, 0]),
  createSampleOption([1e-15, 1e-15, 1e-15]),
  ...
  createSampleOption([0, 100, 0]),
  ...
  createSampleOption([0, 0, 14000000]),
```
lines. 

Some examples of the old behavior, showing that the orientation strongly depends on that magic "epsilon":

Cesium S2 Orientation OLD

Some examples of the new behavior:

Cesium S2 Orientation NEW

To emphasize this again: **This is not a 'proposed solution'**. The approach feels pretty arbitrary, and I could come up with a million cases where this will lead to "unexpected results". (And the functionality of `flyTo` is not covered here at all...)

The goal is mainly to encourage people to chime in, saying what could be "The Right Orientation" for each of the `createSampleOption` cases in this sandcastle:

```javascript
  //globe: false
});

let currentTileset;

function createTilesetObject(center) {
  const tilesetObject = {
    asset: {
      version: "1.1",
    },
    geometricError: 1e10,
    root: {
      refine: "ADD",
      boundingVolume: {
        sphere: [
          center[0],
          center[1],
          center[2],
          12000000,
        ],
      },
      geometricError: 1.0,
    },
  };
  return tilesetObject;
}

function createTilesetUrl(center) {
  const tilesetObject = createTilesetObject(center);
  const url =
    "data:application/json;base64," + btoa(JSON.stringify(tilesetObject));
  return url;
}

function zoomToTilesetFixed(targetTileset) {
  const boundingSphere = targetTileset.boundingSphere;
  const transform = Cesium.Matrix4.fromTranslation(boundingSphere.center);
  const offset = new Cesium.Cartesian3(boundingSphere.radius, 0, 0);
  viewer.camera.lookAtTransform(transform, offset);
}

async function createTileset(center) {
  if (Cesium.defined(currentTileset)) {
    viewer.scene.primitives.remove(currentTileset);
    currentTileset = undefined;
  }
  const url = createTilesetUrl(center);
  currentTileset = viewer.scene.primitives.add(
    await Cesium.Cesium3DTileset.fromUrl(url, {
    })
  );
  viewer.zoomTo(currentTileset);
  //zoomToTilesetFixed(currentTileset);
}

function createSampleOption(center) {
  return {
    text: center.toString(),
    onselect: async function () {
      try {
        await createTileset(center);
      } catch (e) {
        console.log(e);
      }
    },
  };
}

const sampleOptions = [
  createSampleOption([0, 0, 0]),
  createSampleOption([1e-15, 1e-15, 1e-15]),
  createSampleOption([1e-12, 1e-15, 1e-15]),
  createSampleOption([1e-15, 1e-12, 1e-15]),
  createSampleOption([1e-10, 1e-10, 1e-10]),
  createSampleOption([1e-5, 1e-5, 1e-5]),
  createSampleOption([0, 0, 1]),
  createSampleOption([0, 1, 0]),
  createSampleOption([1, 0, 0]),
  createSampleOption([100, 0, 0]),
  createSampleOption([0, 100, 0]),
  createSampleOption([0, 0, 100]),
  createSampleOption([10000, 0, 0]),
  createSampleOption([0, 10000, 0]),
  createSampleOption([0, 0, 10000]),
  createSampleOption([1000000, 0, 0]),
  createSampleOption([0, 1000000, 0]),
  createSampleOption([0, 0, 1000000]),
  createSampleOption([10000000, 0, 0]),
  createSampleOption([0, 10000000, 0]),
  createSampleOption([0, 0, 10000000]),
  createSampleOption([14000000, 0, 0]),
  createSampleOption([0, 14000000, 0]),
  createSampleOption([0, 0, 14000000]),
];
Sandcastle.addToolbarMenu(sampleOptions);
```

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Camera.js
+++ b/packages/engine/Source/Scene/Camera.js
@@ -2302,16 +2302,30 @@ Camera.prototype.getMagnitude = function () {
 const scratchLookAtMatrix4 = new Matrix4();
 
 /**
- * Sets the camera position and orientation using a target and offset. The target must be given in
- * world coordinates. The offset can be either a cartesian or heading/pitch/range in the local east-north-up reference frame centered at the target.
- * If the offset is a cartesian, then it is an offset from the center of the reference frame defined by the transformation matrix. If the offset
- * is heading/pitch/range, then the heading and the pitch angles are defined in the reference frame defined by the transformation matrix.
- * The heading is the angle from y axis and increasing towards the x axis. Pitch is the rotation from the xy-plane. Positive pitch
- * angles are below the plane. Negative pitch angles are above the plane. The range is the distance from the center.
+ * Sets the camera position and orientation using a target and offset.
  *
- * In 2D, there must be a top down view. The camera will be placed above the target looking down. The height above the
- * target will be the magnitude of the offset. The heading will be determined from the offset. If the heading cannot be
- * determined from the offset, the heading will be north.
+ * The target must be given in world coordinates. It will be used for
+ * computing the transformation matrix that describes the local
+ * east-north-up reference frame for the given target, using
+ * {@link Transforms.eastNorthUpToFixedFrame}.
+ *
+ * The offset can be either a cartesian or heading/pitch/range in the local
+ * east-north-up reference frame centered at the target.
+ *
+ * If the offset is a cartesian, then it is an offset from the center of
+ * the reference frame defined by the transformation matrix.
+ *
+ * If the offset is heading/pitch/range, then the heading and the pitch
+ * angles are defined in the reference frame defined by the transformation
+ * matrix. The heading is the angle from y axis and increasing towards
+ * the x axis. Pitch is the rotation from the xy-plane. Positive pitch
+ * angles are below the plane. Negative pitch angles are above the plane.
+ * The range is the distance from the center.
+ *
+ * In 2D, there must be a top down view. The camera will be placed above
+ * the target looking down. The height above the target will be the magnitude
+ * of the offset. The heading will be determined from the offset. If the
+ * heading cannot be determined from the offset, the heading will be north.
  *
  * @param {Cartesian3} target The target position in world coordinates.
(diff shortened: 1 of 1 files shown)


javagl:
The approach that is shown here is a draft. Most of the value of this PR might be the sandcastle with the test cases. But it is still (too) far away from a robust, generic, agreed-upon, universal _solution_ to the problem to warrant keeping it open any longer. (Or to put it that way: Whatever we're going to do: It will be vastly different from what is currently done in this PR).

Issue #11976, opened 2024-05-10
Add support for glTF KHR_materials_anisotropy extension


jjhembd:
### Feature

The [KHR_materials_anisotropy extension](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md) extends the metallic-roughness PBR model to account for anisotropic reflectivity. A surface with anisotropic reflectivity (such as a brushed metal) will reflect differently at different orientation angles in the plane of the surface (even for the same incident and reflected angles with the normal).

CesiumJS currently does not recognize these parameters in a glTF. Here is how the anisotropy test model is currently rendered. Note the lack of variation within columns:
![image](https://github.com/CesiumGS/cesium/assets/41167620/202e5a7c-4a06-4c88-b542-dfe59797d7ea)

Compare to the result from the reference implementation with the extension enabled:
![image](https://github.com/CesiumGS/cesium/assets/41167620/c3037112-bde3-4caa-a79e-f9fdba97aa85)

PR #11988, opened 2024-05-15
Implement glTF extension KHR_materials_anisotropy

jjhembd:
Merge https://github.com/CesiumGS/cesium/pull/11970 first!

# Description

This PR implements the glTF extension [KHR_materials_anisotropy](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_anisotropy/README.md) in physically-based rendering of models.

### Parsing the extension at load time

When the glTF is loaded, `GltfLoader` reads factors and loads textures from the extension. See the new `loadAnisotropy` method, which is very similar to the `loadSpecular` method from https://github.com/CesiumGS/cesium/pull/11970. The factors and textures are then attached to the material. See the updates to the `Material` class in `ModelComponents`.

### Processing the extension properties in the pipeline

In `MaterialPipelineStage`, if `material.anisotropy` is defined, its properties are processed by the new method `processAnisotropyUniforms`, which is very similar to `processSpecularUniforms`. The sine and cosine of the `anisotropyRotation` factor are pre-computed, and combined with `anisotropyStrength` into one `vec3` uniform.

The anisotropy extension, if used, will extend the metallic roughness model. As a result, the check for the extension usually takes place inside the same code branch that handles metallic-roughness processing.

### Using the extension in the shader

The orientation of the anisotropic effect is defined relative to the "tangent" vector on the surface. The tangent vector is aligned in the direction of increasing `u` coordinate in the normal map. These tangent vectors are either supplied as a vertex attribute, or computed via derivatives of the normal map texture coordinates.

We already had code using and/or computing tangent vectors, but discarded the information early once the normals were prepared. To preserve the tangent information for anisotropic calculations, a new `NormalInfo` struct is defined, and returned from the new `getNormalInfo` function. Along the way, some of the tangent calculations were broken down into smaller functions to enable re-use for both the isotropic and anisotropic cases. The `setAnisotropy` function then stores the relevant anisotropic information on the material. See also the corresponding changes to the `czm_modelMaterial` and `czm_pbrParameters` structs.

The actual anisotropic lighting calculations can be found in `pbrLighting` and `ImageBasedLightingStageFS`. They only modify the specular reflection component of the final color.

Note that the implementation in `ImageBasedLightingStageFS` is complete only for the case of supplied environment maps in `textureIBL`. The `proceduralIBL` function will require a more significant rewrite as part of a wider IBL update, so the anisotropy calculation can be incorporated at that stage.

## Other changes

- Minor cleanup in `GeometryPipelineStage` (done while tracing back normal calculations)
- Added more docstrings in `MaterialPipelineStage`
- Fixed normalization bug in `ImageBasedLightingStageFS`. See #11994

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/11976. Resolves https://github.com/CesiumGS/cesium/issues/11994.

## Testing plan

Load the development Sandcastle [glTF PBR Anisotropy](http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FglTF%20PBR%20Anisotropy.html) and toggle through the 4 testing models, comparing to the rendering in the [reference implementation](https://github.khronos.org/glTF-Sample-Viewer-Release/?model=https://raw.GithubUserContent.com/KhronosGroup/glTF-Sample-Assets/main/./Models/AnisotropyStrengthTest/glTF-Binary/AnisotropyStrengthTest.glb).

# Author checklist

Other TODO:

- [x] Load test dataset for the Sandcastle from a server
- [x] Report normalization bug in IBL 
- [x] Describe code changes
- [x] Add image comparisons
- [x] Explain non-implementation in procedural IBL
- [x] Clean up: use TBN matrix?


jjhembd:
## Image comparisons

All images captured from the [glTF PBR Anisotropy Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FglTF%20PBR%20Anisotropy.html)

### "Barn lamp" test dataset
Before this PR:
image

With extension enabled. Notice the stretching of the specular reflection perpendicular to the brush marks in the metal.
image

### Anisotropy strength test
Extension disabled:
image

Extension enabled:
image

### Anisotropy disc test
The discs in each square are not visible without the extension.
image

jjhembd:
@ggetz this is ready for review. But it is written on top of https://github.com/CesiumGS/cesium/pull/11970 so we will want to merge that one first.

ggetz:
I'm seeing this blown-out areas on the barn lamp sample model when using non-procedural IBL:

image

It appears to be related to the image we're using for IBL, but is it expected to be this extreme? Or is this something that will be fixed with the model IBL follow up?

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -1593,17 +1616,20 @@ function loadMaterial(loader, gltfMaterial, frameState) {
       pbrSpecularGlossiness,
       frameState
     );
-  } else {
+  } else if (material.unlit === false) {
```
ggetz, line 1619: Can we simplify to?


```suggestion
  } else if (!material.unlit) {
```

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -1593,17 +1616,20 @@ function loadMaterial(loader, gltfMaterial, frameState) {
...
     if (defined(pbrMetallicRoughness)) {
       material.metallicRoughness = loadMetallicRoughness(
         loader,
         pbrMetallicRoughness,
         frameState
       );
     }
-    if (defined(pbrSpecular)) {
+    if (defined(pbrSpecular && material.unlit === false)) {
```
ggetz, line 1627:
```suggestion
    if (defined(pbrSpecular) && !material.unlit) {
```

I assume the second parenthesis for `defined` is in the wrong position. Secondly, can we simply like in my comment above?

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -102,15 +101,27 @@ MaterialPipelineStage.process = function (
...
+        material.specular,
+        uniformMap,
+        shaderBuilder,
+        defaultTexture,
+        disableTextures
+      );
+    }
+    if (
+      defined(material.anisotropy) &&
+      ModelUtility.supportedExtensions.KHR_materials_anisotropy === true
```
ggetz, line 113:
```suggestion
      ModelUtility.supportedExtensions.KHR_materials_anisotropy
```

?

packages/engine/Source/Scene/GltfLoader.js
jjhembd: Thanks, good catch

packages/engine/Specs/Scene/Model/ModelSpec.js
```diff
@@ -721,6 +723,19 @@ describe(
       verifyRender(model, true);
     });
 
+    it("renders model with the KHR_materials_anisotropy extension", async function () {
```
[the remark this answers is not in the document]
jjhembd: This was an error in my test dataset. For this extension, either tangents or a normal map are required.

jjhembd:
Thanks for the feedback @ggetz! I think I addressed everything.
I also see the rendering artifacts in the "barn lamp" sample model. It almost looks like our environment map is not normalized correctly. The artifacts appear even before this PR, so I think it is something we should look at as part of the IBL work.

packages/engine/Specs/Scene/Model/ModelSpec.js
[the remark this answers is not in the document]
ggetz: Got it. Can we catch this early and provide a better error message?

ggetz:
@jjhembd All TODO's have been completed and this is ready to merge, correct?

jjhembd:
> @jjhembd All TODO's have been completed and this is ready to merge, correct?

@ggetz Yes, the only comment I haven't addressed is your question about whether we can catch a problem with a glTF that is not according to spec. I looked at this, but it is non-trivial.

We often ask users to run problematic models through the glTF-Validator. Would that be appropriate for this case, or do we want CesiumJS to identify model problems?

ggetz:
>Your question about whether we can catch a problem with a glTF that is not according to spec. I looked at this, but it is non-trivial.

OK to skip this if it blows up the scope of this PR. Better error messages for models that do not conform to spec is something we've mentioned before, but I understand there's a trade off in terms of effort and performance.

Issue #11694, opened 2023-12-13
`SampleTerrain` and `SampleTerrainMostDetailed` regression on thrown errors


ggetz:
Historically, `SampleTerrain` and `SampleTerrainMostDetailed` would reject if a 404 error was thrown. At some point in time (I assume https://github.com/CesiumGS/cesium/pull/10149), this behavior broke and a cartographic with undefined height is returned instead.

[Sandcastle](https://sandcastle.cesium.com/#c=bVFhb9owEP0rVr40SMgEGKyogFZSjQZBGDSUQbMPxjZgMHZqO6Rh2n+fIZvUddgfrLt77927s081Sw8wkAISukYpN/cYU60juacCdEDs0HywXfUxG7NBMDsF1ZAFOhDTBvaDZrBPvj/7gxa0oFfS31tQUF1G4Xa521TD3Pu0OI2yYTR5W+xGZhxNGqFfZeF8yhdR77A8LHfhacSG/iBZWrERyxiaf/WCnXwLd7N8/LDJwtOktp7Ah0FqwkbTm07nt8faZ9V8rNez5LbpRS2yiFq095j0X+urIfNmsXMXi1hgKbQBR0YzquwYgmbAL0Z9vuTc2MGX2JfCICaoip2SZRYMaKhSNvtNySMjFwWUIWb+ahRP9C8IrpU82EXea01NQNxaveU1ao3Sxc8TEgQjbTiFiJBISr5CqpcaI4V705c3ZYB0LjBwS53uz1gAUAygqLZf8rG9RoeE0z/tR1KbB2pn4JS41+2Xwcu7BfhIGblRKNky7Hple0s/ziaLntI65HLjFp1fvEvpV+nOKTttbXJOu2fk+Xxhh0QqA1LFXQgrhlpTyFBdWaV4Tw3EWheyALQr76ltwo6Akc6VLwCYI61tZZ1y/sRONHa67YrF/0flEhEmNuMjVRzlZ9i22h0WSQhhu2LD60xTLP+D8m8)

To throw a bit of a wrench into the gears, https://github.com/CesiumGS/cesium/pull/11530 has since been added, unknowingly adding an opt-in flag to replicate the previous behavior.

We will likely remove or change the default for that flag and restore the older behavior where an error would cause the promise returned by the function to reject.

PR #11998, opened 2024-05-22
Make sampleTerrain behaviour consistent with the rejectOnTileFail flag

jesseyay:
# Description

Passes through the `rejectOnTileFail` flag so it will be respected if any tile requests fail, not just if the first request fails. Note that the linked issue discusses changing the default of this flag to `true` which I tried, but it ends up failing the test [fills in what it can when given a mix of positions with and without valid tiles](https://github.com/CesiumGS/cesium/blob/11dd728dfee6c10657f8e1197776fc5a9237ef85/packages/engine/Specs/Core/sampleTerrainSpec.js#L107)

## Issue number and link

Addresses/Related to: https://github.com/CesiumGS/cesium/issues/11694

## Testing plan

# Author checklist


chris-cooper:
@ggetz Would it be possible for you or someone on the team to have a look at this as we notice some tile errors are still slipping through unnoticed.

ggetz:
Thanks @jesseyay! We'll review shortly with the intention of including this in the next CesiumJS release.

chris-cooper:
> Do you or @chris-cooper have any thoughts about reversing the default of rejectOnTileFail to true given https://github.com/CesiumGS/cesium/issues/11694? That would restore behavior that broke in 1.92.

That makes sense if it was the old behaviour.  Do you want to remove the flag as part of this PR @jesseyay ?

jesseyay:
Hey @ggetz thanks for taking a look!

I tried changing the default in this PR for `rejectOnTileFail` to `true` but there is [this test](https://github.com/CesiumGS/cesium/blame/11dd728dfee6c10657f8e1197776fc5a9237ef85/packages/engine/Specs/Core/sampleTerrainSpec.js#L107) which looks like it is 10 years old that will fail if we change it 🤔 

I did a bit of further digging and I suppose the behaviour with a 404 response depends both on the `rejectOnTileFail` flag and on the `terrainProvider` that gets passed in, so maybe the regression https://github.com/CesiumGS/cesium/issues/11694 originates in a terrain provider?

ggetz:
Got it, thanks for looking into that @jesseyay! Sounds like we'll need to look a bit closer to find the source of the regression. No need to hold up this PR for that.

Issue #11231, opened 2023-04-13
Intermittent canceled request causing test failure in `afterAll` in CI 


ggetz:
We're seeing an intermittent failure in CI failing due to an error thrown in `afterAll` with an `undefined` error.

I'm having trouble reproducing locally, but in [`test-CI`](https://github.com/CesiumGS/cesium/tree/test-CI) I threw in a few debug statements, and the failure appears to be a canceled request.

Perhaps related to https://github.com/CesiumGS/cesium/issues/3888, but I expect this likely a bad spec.

ggetz:
Just ran into this locally-- It appears to happen in `Scene/ArcGisMapServerImageryProvider`

jjhembd:
I just saw this in CI:

```
[16:26:11] Finished 'buildRelease' after 48 s
[16:26:13] Using gulpfile ~/build/CesiumGS/cesium/gulpfile.js
[16:26:13] Starting 'test'...
Building specs...
  Build/CesiumUnminified/index.js       8.6mb ⚠️
  Build/CesiumUnminified/index.js.map  16.8mb
⚡ Done in 1408ms
  Build/CesiumUnminified/Cesium.js       9.1mb ⚠️
  Build/CesiumUnminified/Cesium.js.map  16.9mb
⚡ Done in 1273ms
  Build/CesiumUnminified/index.cjs       8.7mb ⚠️
  Build/CesiumUnminified/index.cjs.map  16.9mb
⚡ Done in 1263ms
Chrome Headless 114.0.5735.198 (Linux x86_64) ERROR
  An error was thrown in afterAll
  Unhandled promise rejection: undefined thrown
Chrome Headless 114.0.5735.198 (Linux x86_64) ERROR
  An error was thrown in afterAll
  Unhandled promise rejection: undefined thrown
0.417 secs: Core/CesiumTerrainProvider requestTileGeometry Uses query parameter extensions for ion resource
0.305 secs: Core/CesiumTerrainProvider requestTileGeometry uses multiple urls specified in layer.json
0.282 secs: Core/PolygonGeometry create geometry creates with lines with different number of subdivisions for geodesic and rhumb
0.163 secs: Core/QuantizedMeshTerrainData createMesh requires 32bit indices for large meshes
0.154 secs: Core/CesiumTerrainProvider fromIonAssetId throws without assetId
0.135 secs: Core/GoogleEarthEnterpriseTerrainData upsample works for all four children of a simple quad
[...]

Issue #4245, opened 2016-08-29
Easy API to select a non-earth ellipsoid


hpinkos:
Users on the forum want to use non-earth ellipsoids but it is unclear how they have to set it.

emackey:
Oh, if only choosing an ellipsoid were so easy.  Unfortunately not so!  You must tell the `imageryProvider` and the `mapProjection` about your choice of ellipsoid.  The `CesiumWidget.globe` will automatically grab it from `scene.mapProjection` if it needs it.

In this demo, the hover text is Cartesian3 in meters, so you can see it's the Moon's radius and not the Earth.

```
// First, select your favorite ellipsoid.
var ellipsoid = Cesium.Ellipsoid.MOON;

var imagery = new Cesium.SingleTileImageryProvider({
    // Your imagery provider needs the ellipsoid.  Specify it here.
    ellipsoid: ellipsoid,
    url: Cesium.buildModuleUrl('Assets/Textures/moonSmall.jpg')
});

var viewer = new Cesium.Viewer('cesiumContainer', {
    // The CesiumWidget and the Globe get Ellipsoid from Scene.mapProjection.
    mapProjection: new Cesium.GeographicProjection(ellipsoid),
    // Don't forget your imagery provider that uses the ellipsoid.
    imageryProvider: imagery,
    // Set some other options too.
    selectionIndicator : false,
    infoBox : false,
    baseLayerPicker: false,
    geocoder: false,
    skyAtmosphere: false
});

var scene = viewer.scene;
var handler;

var entity = viewer.entities.add({
    label : {
        show : false
    }
});

handler = new Cesium.ScreenSpaceEventHandler(scene.canvas);
handler.setInputAction(function(movement) {
    var cartesian = viewer.camera.pickEllipsoid(movement.endPosition, scene.globe.ellipsoid);
    if (cartesian) {
        var cartographic = Cesium.Cartographic.fromCartesian(cartesian);
        var longitudeString = Cesium.Math.toDegrees(cartographic.longitude).toFixed(2);
        var latitudeString = Cesium.Math.toDegrees(cartographic.latitude).toFixed(2);

        entity.position = cartesian;
        entity.label.show = true;
        entity.label.text = '(' + cartesian.x.toFixed(0) + ', ' + cartesian.y.toFixed(0) +
            ', ' + cartesian.z.toFixed(0) + ')';
    } else {
        entity.label.show = false;
    }
[...]

PR #12000, opened 2024-05-24
Ease of use for Non-WGS84 Ellipsoids

ggetz:
# Description

This PR is designed to make the API for setting an alternative ellipsoid throughout the API more streamlined. There are some additional changes needed for 3D Tiles, but that is coming in a follow-up PR.

The main addition is `Ellipsoid.default`, which is now used throughout the API as the default ellipsoid rather than `Ellipsoid.WGS84` directly. That means instead of having to do `Cartographic.fromCartesian(0, 1, 3, myEllipsoid)` everywhere, you can now set `Ellipsoid.default = myEllipsoid` once.

There were a few places which assumed WGS84 even before instantiation, so setting `Ellipsoid.default` modifies a few other classes. This is counterintuitive, but done to avoid circular dependencies (ie. `Ellipsoid` needs to depend on `Cartesian3`, so `Cartesian3` can't depend on `Ellipsoid`).

Furthermore, we make a few Earth-centric assumptions throughout the API.We now disable them when the ellipsoid is set to something other than `Ellipsoid.WGS84` or use a default proportionate to the selected ellipsoid. This includes:

* atmosphere (ground and sky)
* fog
* lighting and night imagery; fade in and fade out distances
* default camera position
* the star skybox
* the moon

In particular, I had to make some configuration changes to the base layer picker to 1) Ensure the WGS84 ellipsoid was always used when relevant, and 2) not override when you set the viewer's ellipsoid to something other than the WGS84 ellipsoid

Last, but not least, there were some instances of naming things with `WGS84` when we just mean any arbitrary ellipsoid. These have been changed, and where needed functions have been renamed and the old versions deprecated.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/4245
Fixes https://github.com/CesiumGS/cesium/issues/3543
Fixes https://github.com/CesiumGS/cesium/issues/11231

## Testing plan

[Testing Sandcastle](https://ci-builds.cesium.com/cesium/ellipsoids/Apps/Sandcastle/index.html#c=lVZtc9o4EP4rGmZuBmaosGxDIJf2jgBloATCS9qS4+ZG2ALU2JJriRC3k/9+K4N5SZhe7ottrXaf3X12V3KDKb4OcUcK7LMFXQd6zOJHFqP3aJZbaR2py1KJRvyd0nTJxRJ7WwNPhrPc7zNRKqF2TIVGjVTeHSPqgYpCWqJErmPEpUBUKabVTDReOaunyhP5wAR6PxMInLKku5q3PT7g3c7djw7p847qiFHZa3QqnYfo6+dGt4ZB6bvffjBKq/vmXdKfePZ0TFY34TDpTYZ2v32np2HL7icWuQ+H5ZtwFPbbrWTQHvJeoxvdA9jNZAjA/W9z5zrwkk7lCzj2nI7ZF9Ok6xu96dchH3xrkUFz+TRo1p/6zesa/u586n6mk4mmN8q1XLq4n0p7NBzPp8ugZdW+LuLw4vrh/rZSc1OO9om3goBHSnI/Sx9YfrV1Mxj0wciTQmkUSikGkQYSFej+NAxpFseUi9tYPnKfxZeIbijP+Mfb1+RUBy9iGQLrdVOHjp83MAgR26mar0LRPOdUsR5NDGBWppAuWZykwhQhg6urRHg7kENJd+onPjOHxLaqhczXc0rJNr9HzjZpswm2ybA+p7L8LLdttYYUGnJh8SxX3DKAMT7iJY1+rdh1LDeKxSMGrRky4TN/xJQM1kbpEi1ooFiqqnnIAsA7llHBQ/pScc/ILfceDC+HrSWTnvSPZc8FyGqbDVYeEwyHitIxDaOAmdK5adIwLbdS8W09PRgPzVSC5AJlk7bZbMCcM+Exn8fM02bQSjtJicaaewG856oUcV4aWxapEcspk4pjkTKp/vHI6W9Ok6/W84zhSHKh1WDREdA4TJmW+8uk8HNbP82e9CVMXT2SQSARIcDydicAQvTaB54sXLlwSSWTS7HcbdgOdi8cUk53nou/gHXPwL5zcMV1ycVrXMe1cdkmzhuAy2eA7QomjuOSM8C4Yld/gdpbC/mwkj46R4NTxZDruXjtMq46tmO/Adg+F++x+Wm8Fq7Z9hHBf6dttJAxyu/La/rnZZUL2wB2DckEuOJMYer7+V1k0a4P9+PeoKYbORVOOrpNtowZU7sxN/oc7yMrnkh3iWxlhX16cxZcZjxkTBh987UHWMB0G4KIG2kkI7gGFBVqTxJCcq3NtDZkIA9HU7rCzfro07hXn7Tao9b0pcEX7usVcLuXK50EbI/QM+GNjQh/7PR6/9T7zX8Gd5Nep986JMefWDBYLOAMuzw+oPZM2XmriN7ZduHgxKMBu06aHK5LmNgTsz6j8Ucaj0GFxnmCy8wuIoIt8yiziyLMWfmApOFWVcEajhIveRNgZoiQncId1sbHYeVi62TXwla2yLw/F7Me4aY6P08IGfMfEAY5QHpnSmM66DoBlvK26xSR7UKqtVrhLVU9MiU1MCVVIIiQwn8X+H9x7+y439IBLJDT7NMD/TlXzF2ljfMh8/InDyMZa7SOgzzGJc3ghKcQbWm+hlsCzmuljKVRvSodm175/BFx//2Zew15Afwiwc5iHaT8znIfrkqg/8o0kNSHv7AB/KEFNDFqK/KhtxXCtXhVguV5Sy1lMKfxC+R/AQ)

Additionally to test this, I iterated through many Sandcastle examples, setting `Cesium.Ellipsoid.default = Cesium.Ellipsoid.MOON;` or another arbitrary ellipsoid as the first line. 

`Cesium.Ellipsoid.default = Cesium.Ellipsoid.UNIT_SPHERE;` is also a possibility, and while nothing should crash, there is some weirdness purely due to the size disparity.

# Author checklist

- [x] I have updated `CHANGES.md` with a short summary of my changes


ggetz:

Additionally, @angrycat9000 can you take a look from a usability standpoint?

packages/engine/Source/Core/ArcGISTiledElevationTerrainProvider.js
```diff
@@ -446,7 +446,7 @@ ArcGISTiledElevationTerrainProvider.prototype.requestTileGeometry = function (
         encoding: that._encoding,
       });
     })
-    .catch(function (error) {
+    .catch(async function (error) {
```
ggetz, line 449: Sorry, this is technically an unrelated change, but I was able to track down what I believe was causing https://github.com/CesiumGS/cesium/issues/11231 while troubleshooting unit tests.

packages/engine/Source/Core/ArcGISTiledElevationTerrainProvider.js
```diff
@@ -455,10 +455,14 @@ ArcGISTiledElevationTerrainProvider.prototype.requestTileGeometry = function (
...
         // Don't reject the promise till the request is actually cancelled
         // Otherwise it will think the request failed, but it didn't.
-        return request.deferred.promise.finally(function () {
-          request.state = RequestState.CANCELLED;
-          return Promise.reject(error);
-        });
+        try {
+          await request.deferred?.promise;
+        } catch {
+          // Eat this error
```
angrycat9000, line 461: What is the reason not bubbling this error up or at least sending it to console.error?

Silently suppressing the error can make it harder to debug when something does go wrong

packages/engine/Source/Core/CircleGeometry.js
```diff
@@ -90,7 +90,7 @@ const scratchEllipseGeometry = new EllipseGeometry({
 const scratchOptions = {
   center: new Cartesian3(),
   radius: undefined,
-  ellipsoid: Ellipsoid.clone(Ellipsoid.UNIT_SPHERE),
+  ellipsoid: Ellipsoid.clone(Ellipsoid.default),
```
angrycat9000, line 93: Was this intentional or something that just didn't cleaned up?  This would change the behavior for  any callers not passing in a ellipsoid.

packages/engine/Source/Core/Occluder.js
```diff
@@ -519,7 +519,7 @@ Occluder.computeOccludeePointFromRectangle = function (rectangle, ellipsoid) {
   );
   const bs = BoundingSphere.fromPoints(positions);
 
-  // TODO: get correct ellipsoid center
+  // Assumes the ellipsoid is centered at the origin
```
angrycat9000, line 522: Ellipsoid doesn't define an origin so isn't it always centered on the origin?

packages/engine/Source/Core/Cartesian3.js
```diff
@@ -892,7 +892,9 @@ Cartesian3.fromDegrees = function (
 
 let scratchN = new Cartesian3();
 let scratchK = new Cartesian3();
-const wgs84RadiiSquared = new Cartesian3(
+
+// To prevent a circular dependency, this value is overridden by Ellipsoid when Ellipsoid.default is set
+Cartesian3._ellipsoidRadiiSquared = new Cartesian3(
```
angrycat9000, line 897: Would it make more sense to have break the dependency by having a `DefaultEllipsoidValues` file that both `Ellipsoid` and `Cartesian` depend on?

Putting an ellipsoid property on the Cartesian class made these classes feel to closely coupled

angrycat9000 (commented):
Hi Gabby, thanks for working on this. I wasn't able to run the sandcastle.

> TypeError: Simon1994PlanetaryPositions_default.computeSunPositionInMoonInertialFrame is not a function
>     at setSunAndMoonDirections (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:33500:54)
>     at UniformState.update (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:33578:3)
>     at render (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:235243:6)
>     at tryAndCatchError (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:235283:5)
>     at Scene4.render (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:235338:5)
>     at CesiumWidget.render (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:238457:17)
>     at render2 (https://ci-builds.cesium.com/cesium/ellipsoids/Build/CesiumUnminified/index.js:237868:18)

ggetz:
> Hi Gabby, thanks for working on this. I wasn't able to run the sandcastle.

@angrycat9000 Apologies, I forgot to push up my last commit. Everything should be good to go now.

packages/engine/Source/Core/Cartesian3.js
ggetz: We would still run into a similar problem because, to avoid the workaround, that new file would need to depend on `Ellipsoid`.

ggetz: We could omit the values, but the circular dependency is still functionally there. 

The only way to remove it entirely would be to break the public API by moving `fromDegrees` or `fromRadians` out of that class. Theoretically better archectecture, but breaking the public API which I'm erring on the side of avoiding since those function have been in place for years.

angrycat9000:
Hi @ggetz Thanks for that tweaked example.  I'm running into some issue with the camera when switching back and forth.

See this [example](https://ci-builds.cesium.com/cesium/ellipsoids/Apps/Sandcastle/index.html#c=tVXvb+I4EP1XfHxKJM4ByvbahVZHKceF5UchabcgpJNJBvDi2Fnbgc2u+r+vk8AttNxJd9KKD8HjN/PmvXicNiiaRNgVHIewJAnTHsgtSHSD5qW11rF67zgkpr8qTVaUr3BQJAQimpcac+44qCsJ16idx3seIoGBKKQFSkUiERUcEaVAqzlvvyFr5WBfbICjmzlHhhTS3nrRDeiI9tzHr251SF3l8sm7oO1eupv4+andu8YG9DnsbjLQenb/mA79oDb1qutBNE77/rg27D7qadSpDdNKdRaN3w2iSTTsdtJRd0z77V48M8UG/tgUHn5aXNyxIHUvPxri4MLN9vk07YUZbvo8pqNPnerofvVldN/6Mry/u8afLz70nojvazJQ9UqdLGdTUZuMvcV0xTqV6+eljH6728weLq/ruUfZLxBcabSlsMu95bDbO4af8pg1LxXOtgXXhHKQ81IZfcssSRTcSbFTICdgbI+AhxBOQAmWaOPue7QkTEE5g2oaATPJxzHCaUReAx1nQRT0SQrygQYbkMebKxCBCI9jL7aRUTSPVQAccKQI8UgUM1BGTn2vkoFGVA2Eeec3RXLjIH1NeMhea/cCCcC9mATQ2QLXfxYg64QqIHxLVNbBvgY2p8nlcWJOT6bLIirlAVomPF8iyy58+7uRX4p/jSx4UtkYswKZ5jYoLCESW7D+BbECbVXscqHMzjUbmqVV1N/TooO4DmM0VoKGh8NuWnmzNRiNho0ibU+sQUpzAB6k2NIwN4zsCD0MGC4e/ikIL6WIzFi1skFzQ6tau7iyyydlz+khYWgdZvJoI692KN3KzLV+TO4eeML8g7ZyZdt24wyx2qQtHQkVr0Ga1Vrsjk5Ihn9BYBb/x8KPXe+q/hM8PKvj5xpY+y/uaZlADn7JHHwp/8NM+WkMuN/5w/+r3XfbH+wGKpVLTaVTBrcFFUK/0ygWUqNEMgtjR4MZbKJBOYvEXA4aB0od+mo6x6nNkG4RDW/O3F4oYObeNzvLhDGPfoV56bbpGPybVCZIaD4tI/PZYSTNYOvqbb8IYoybjlmez9RCsAWRryp/Bw).  Click to change to moon.  Then try to zoom in.  Notice it does not let you zoom in very far.  Are there other things I would need to reset to change between moon and earth?

angrycat9000:
At this point I'm running into enough issues with stuff that was created before changing the default ellipsoid, I'm going to try just throwing out the whole viewer and recreating it if I need to change between Moon/Earth

ggetz:
@angrycat9000 Gotcha. In the meantime I'll look into seeing if we have better ways of transitioning between the two.

angrycat9000:
Is there a reason why the star skybox was removed?    ion is using a grey background so this is more noticeable than in the sandcastle demo with the black background.  Would prefer to have this left in place for the moon.

image

ggetz:
@angrycat9000 This was turned off since the star positions are not accurate when using the moon as the central position.

angrycat9000:
> @angrycat9000 This was turned off by default since the star positions are not accurate when using the moon as the central body. It can be re-enabled.

That makes sense.  I can just update the background to black for the moon for now.

Not that it has to be in this PR, but curious about what would be the level of effort to get an accurate skybox suitable for use with lunar data? Is there any issue to track that?

It might also make sense to provide a method like `Skybox.getEarthSkybox()` to get the default skybox box that the viewer uses.  That would enable consumers to turn the skybox on for the moon if they don't care about the accuracy of the star positions.

malaretv:
First off this is awesome! Very excited to have this land 😄 

> @angrycat9000 Gotcha. In the meantime I'll look into seeing if we have better ways of transitioning between the two.

Wayyyy out-of-scope for this issue, but just want to flag that transitioning between different planets will also need some way to dynamically load in different approximateTerrainHeights, once terrain and ground primitives are involved. We work with quite a few planetary bodies and I have yet found a way to seamlessly support viewing them all within one packaged application.

packages/engine/Source/Scene/Globe.js
```diff
@@ -191,9 +192,9 @@ function Globe(ellipsoid) {
    * Enable the ground atmosphere, which is drawn over the globe when viewed from a distance between <code>lightingFadeInDistance</code> and <code>lightingFadeOutDistance</code>.
    *
    * @type {boolean}
-   * @default true
+   * @default true when using an earth ellipsoid, false otherwise
```
jjspace, line 195:
```suggestion
   * @default true when using the WGS84 ellipsoid, false otherwise
```

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2625,6 +2625,7 @@ function parse(loader, gltf, supportedImageFormats, frameState) {
 
   if (defined(cesiumRtcExtension)) {
     // CESIUM_RTC is almost always WGS84 coordinates so no axis conversion needed
+    // TODO: Ellipsoids
```
jjspace, line 2628: TODO?

packages/engine/Source/Scene/Scene.js
```diff
@@ -533,6 +538,16 @@ function Scene(options) {
    * @type {Fog}
    */
   this.fog = new Fog();
+  this.fog.enabled = Ellipsoid.WGS84.equals(this._ellipsoid);
```
jjspace, line 541: Should the docs for this indicate it's only active for WGS84?

packages/engine/Source/Scene/ScreenSpaceCameraController.js
```diff
@@ -216,33 +216,46 @@ function ScreenSpaceCameraController(scene) {
     eventType: CameraEventType.LEFT_DRAG,
     modifier: KeyboardEventModifier.SHIFT,
   };
+
+  const ellipsoid = defaultValue(scene.ellipsoid, Ellipsoid.default);
+
   /**
    * The minimum height the camera must be before picking the terrain or scene content instead of the ellipsoid.
    * @type {number}
    * @default 150000.0
```
jjspace, line 225: All the docs for these properties indicate a default that's not technically true if the ellipsoid has been changed. Maybe they should be updated?

packages/engine/Source/Widget/CesiumWidget.js
```diff
@@ -123,16 +123,17 @@ function configureCameraFrustum(widget) {
  * @param {Element|string} container The DOM element or ID that will contain the widget.
  * @param {object} [options] Object with the following properties:
  * @param {Clock} [options.clock=new Clock()] The clock to use to control current time.
+ * @param {Ellipsoid} [options.ellipsoid=Ellipsoid.default] The default ellipsoid.
  * @param {ImageryLayer|false} [options.baseLayer=ImageryLayer.fromWorldImagery()] The bottommost imagery layer applied to the globe. If set to <code>false</code>, no imagery provider will be added.
- * @param {TerrainProvider} [options.terrainProvider=new EllipsoidTerrainProvider] The terrain provider.
+ * @param {TerrainProvider} [options.terrainProvider=new EllipsoidTerrainProvider(options.ellipsoid)] The terrain provider.
  * @param {Terrain} [options.terrain] A terrain object which handles asynchronous terrain provider. Can only specify if options.terrainProvider is undefined.
  * @param {SkyBox| false} [options.skyBox] The skybox used to render the stars.  When <code>undefined</code>, the default stars are used. If set to <code>false</code>, no skyBox, Sun, or Moon will be added.
```
jjspace, line 130: Should this include some sort of comment about the skybox and moon being turned off if the ellipsoid is changed?

jjspace (commented):
Other than a few small comments, some of which we've already discussed in person, I think this is looking good from a code perspective. from basic testing it seems ok too but i'm not confident I know enough to fully evaluate that but @angrycat9000 seems to have that covered

angrycat9000:
Might also be a nice change to set the globe base color to a grey instead of blue when using the moon ellipsoid.

packages/engine/Source/Core/Cartesian3.js
```diff
@@ -904,7 +906,7 @@ const wgs84RadiiSquared = new Cartesian3(
  * @param {number} longitude The longitude, in radians
  * @param {number} latitude The latitude, in radians
  * @param {number} [height=0.0] The height, in meters, above the ellipsoid.
- * @param {Ellipsoid} [ellipsoid=Ellipsoid.WGS84] The ellipsoid on which the position lies.
+ * @param {Ellipsoid} [ellipsoid=Ellipsoid.default] The ellipsoid on which the position lies.
```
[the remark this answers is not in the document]
ggetz: Discussed offline: We can't use `Ellipsoid` directly from within this class. We can externalize the logic to a different file, but the fundamental circular dependency remains.

packages/engine/Source/Core/CircleOutlineGeometry.js
```diff
@@ -14,7 +14,7 @@ import Ellipsoid from "./Ellipsoid.js";
  * @param {object} options Object with the following properties:
  * @param {Cartesian3} options.center The circle's center point in the fixed frame.
  * @param {number} options.radius The radius in meters.
- * @param {Ellipsoid} [options.ellipsoid=Ellipsoid.WGS84] The ellipsoid the circle will be on.
+ * @param {Ellipsoid} [options.ellipsoid=Ellipsoid.default] The ellipsoid the circle will be on.
```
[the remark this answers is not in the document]
ggetz: Discussed offline: This is outside of the scope of this PR, but can be addressed systematically after https://github.com/CesiumGS/cesium/pull/12019 is merged and we enable the `jsdoc/no-defaults` rule.

ggetz:
> Not that it has to be in this PR, but curious about what would be the level of effort to get an accurate skybox suitable for use with lunar data? Is there any issue to track that?

Not yet. I'll open an issue to track follow up changes. I'm not actually sure what all went into generating the default starmap.

packages/engine/Source/Scene/GltfLoader.js
ggetz: Double checked, no changes needed here.

ggetz:
> Might also be a nice change to set the globe base color to a grey instead of blue when using the moon ellipsoid.

For now, I think it's best to leave this up to the app developer. We'll keep the blue default in the API, as I don't think realism was the driving factor for the color choice. 😆

ggetz:
@jjspace This should be ready for another look.

jjspace (approved):
This looks good to me now, most of my comments were minor code and docs anyway. @angrycat9000 is there anything else you want to check or add from your testing?

angrycat9000:
> This looks good to me now, most of my comments were minor code and docs anyway. @angrycat9000 is there anything else you want to check or add from your testing?

This good from my end

ggetz:
Thanks all!

Issue #10815, opened 2022-09-24
Evaluate eslint-plugin-jsdoc


sanjeetsuhag:
To help enforce consistency in how we use [JSDoc](https://jsdoc.app/) in the CesiumJS codebase, we should evaluate the [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc). It seems like it is actively maintained and fairly popular. If we like it enough, we should also consider it for [eslint-config-cesium](https://github.com/CesiumGS/eslint-config-cesium).

jjhembd:
After the work by @onsummer in #11080, this plugin could be a good way to ensure consistency going forward.

The default rules might require too much change at once. We could phase the implementation as follows:

## First PR
1. Add the dependency
2. Update `.eslintrc.json` and `.eslintrc.json`
3. Enable the [`check-types` rule](https://github.com/gajus/eslint-plugin-jsdoc#user-content-eslint-plugin-jsdoc-rules-check-types) _only_, setting it to "error"
4. Confirm that the results match what we implemented in #11080

## Later PRs
Test more of the recommended rules, and enable the ones that give the best return on effort. We could prioritize as follows:
1. Rules that don't require much change to our docs. These will enforce existing doc style going forward
2. Rules that flag behavior that can be auto-corrected with the `--fix` option. For example, [`check-tag-names`](https://github.com/gajus/eslint-plugin-jsdoc#check-tag-names) to replace `@return` with `@returns`
3. Rules that flag serious errors like invalid tag names

PR #12019, opened 2024-06-04
Enable eslint-jsdoc

ggetz:
# Description

This PR enables [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc). I figured we'd go ahead and enable the defaults, and we'll turn off rules for now that will require a lot of manual fixes. We can systematically enable rules and fix them one-by-one where time allows or need arises. For example, type checking may be particularly helpful.

* Uses the recommended ruleset. In the future, we may want to turn on TS mode, but for now we'll default to the recommend for JS-only.
* Disable rules that are against our coding consistencies, at least for the time being and added comments why
* Disabled quite a few rules which require manual fixes, like filling in types and descriptions, until we can manually fix.
* Autofix with `npm run eslint -- --fix`

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/10815

## Testing plan

1. Run `eslint` and ensure there are no failures
2. Run `build-docs` and ensure the built documentation is not broken
3. Run `build-ts` and ensure there are no errors (there will still be warnings)

# Author checklist


packages/engine/Source/Core/AxisAlignedBoundingBox.js
```diff
@@ -47,12 +45,10 @@ function AxisAlignedBoundingBox(minimum, maximum, center) {
 
 /**
  * Creates an instance of an AxisAlignedBoundingBox from its corners.
- *
```
jjspace, line 50: I'm personally a big fan of this line break creating visual separation between a description and the tags. It looks like it's easy to enforce this with the plugin too (and auto-fixable) with the [`tag-lines` rule](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/tag-lines.md#readme).

```js
"jsdoc/tag-lines": [
  "error",
  "any",
  {
    startLines: 1,
  },
],
```

Optionally it may be nice to set it or [`sort-tags`](https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/sort-tags.md#readme) up for separation between specific blocks of tags but that's a bigger change and may need some manual intervention so I'm fine not inlcuding that for now.

jjspace:
If I could summarize the changes I'm seeing

1. rename properties that are synonyms to the "true" one, this is good imo :+1: 
    - constructor to class
    - return to returns
    - exception to throws
2. remove empty lines as outlined in my comment above :neutral_face:  
    - I agree with some of these that were excessive (double lines, ones between some tags)
    - I disagree with many of them that created valuable visual separation between chunks of tags and between the description and all tags. again, mentioned above
3. adding `@params` for those that didn't exist before, again see above :shrug: 

I've reviewed about 100 of the 800+ files and these seem to be the only changes. I'm going to stop here for now in case we change how the rules above are handled so I don't waste time reviewing files that may not actually have any changes.

ggetz:
This need a merge. I'm closing until I get the chance.

Issue #12041, opened 2024-06-19
Regression for rendering between 1.117 and 1.118


javagl:
A user provided a glTF model at https://community.cesium.com/t/measurment-tools-stopped-working/32962/12 
When uploading this to Cesium ion and tiling it as 3D tiles, this is the result:
[Forum 32962 4x panel stojak as 3D Tiles.zip](https://github.com/user-attachments/files/15901780/Forum.32962.4x.panel.stojak.as.3D.Tiles.zip)

Here is a comparison of the model rendered in 1.117 and 1.118:

![Cesium Forum 32962 compare 117 to 118](https://github.com/CesiumGS/cesium/assets/5597569/08c165d5-0cb5-446b-95df-674d51324c67)

The actual model is stored as a composite tile. For convenience/debugging/analyzing that, I extracted the GLB parts from that, including the JSON parts of the GLBs, attached here:

[Forum 32962 composite elements.zip](https://github.com/user-attachments/files/15901924/Forum.32962.composite.elements.zip)

The GLBs are using (Draco and) the rendering-related extensions
```
    "KHR_materials_specular",
    "KHR_materials_ior",
```
which might help to narrow down where the error was introduced. I didn't do a bisection or analysis of https://github.com/CesiumGS/cesium/compare/1.117...1.118 yet, but maybe someone already has an idea based on the description here.

jjhembd:
The model has the following value set in the `KHR_materials_specular` extension:
```json
        "KHR_materials_specular": {
          "specularColorFactor": [
            2,
            2,
            2
          ]
        },
```

Before version 1.118, CesiumJS did not support this extension at all. 

If I remove the `specularColorFactor` from the glTF and load in 1.118, the model renders as before:
![image](https://github.com/CesiumGS/cesium/assets/41167620/cda74ec3-8dde-489b-a812-77b4afa44d94)

It is not immediately clear to me what the impact of the `specularColorFactor` _should_ be in this case. But we could potentially have an error somewhere in https://github.com/CesiumGS/cesium/pull/11970

javagl:
The [`KHR_materials_specular`](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_specular/README.md) extension spec is... quite elaborate and math/shader-heavy, so I can not quickly say anything profound. A few things that caught my attention:

[...]

PR #12043, opened 2024-06-20
Fix base color and reflectivity calculations for PBR materials

jjhembd:
# Description

Our calculation of the base diffuse color for a material was scaled by the transmission coefficient `(1.0 - f0)` (where `f0` is the specular reflection coefficient at normal incidence):
```glsl
material.diffuse = material.diffuse * (1.0 - f0) * (1.0 - metalness);
```
This intuitively seems to make sense at normal incidence. The incident light that actually interacts with the microfacets of the surface (in the diffuse model) will be reduced by the amount of light already reflected specularly.

However, this simple 1D scaling is an oversimplification of something that is already accounted for in the BRDF, so including the scaling results in PBR materials being rendered slightly too dim.

This PR instead computes the base diffuse color using a formula more similar to the [glTF Sample Viewer](https://github.com/KhronosGroup/glTF-Sample-Viewer):
```glsl
material.diffuse = mix(material.baseColor.rgb, vec3(0), metalness);
```

Note the new `baseColor` property on the `czm_modelMaterial` struct. This helps avoid confusion from overwriting the `diffuse` property too many times, and fixes https://github.com/CesiumGS/cesium/issues/12041, where the `KHR_materials_specular` extension was using the `diffuse` property as if it were the raw base color.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/12041

## Testing plan

# Author checklist

- ~[ ] I have added or updated unit tests to ensure consistent code coverage~
- ~[ ] I have update the inline documentation, and included code examples where relevant~


packages/engine/Source/Shaders/Model/MaterialStageFS.glsl
```diff
@@ -296,12 +296,12 @@ float setMetallicRoughness(inout czm_modelMaterial material)
 
     // dielectrics use f0 = 0.04, metals use albedo as f0
     const vec3 REFLECTANCE_DIELECTRIC = vec3(0.04);
-    vec3 f0 = mix(REFLECTANCE_DIELECTRIC, material.diffuse, metalness);
+    vec3 f0 = mix(REFLECTANCE_DIELECTRIC, material.baseColor.rgb, metalness);
 
     material.specular = f0;
 
     // diffuse only applies to dielectrics.
-    material.diffuse = material.diffuse * (1.0 - f0) * (1.0 - metalness);
```
jjhembd, line 304: I have not been able to trace why we multiplied by `(1.0 - f0)` here.

CHANGES.md
```diff
@@ -14,6 +14,7 @@
 
 ##### Fixes :wrench:
 
+- Fixed diffuse color calculation for PBR materials. Many models will now appear slightly brighter. This also fixes an [issue affecting the KHR_materials_specular extension](https://github.com/CesiumGS/cesium/issues/12041). [#12043](https://github.com/CesiumGS/cesium/pull/12043)
```
ggetz, line 17: Even though they were fixed in one PR, I would suggest listing this as two different items. The former (`Fixed diffuse color calculation for PBR materials`) can link to the PR. The later (`fixes an issue affecting the KHR_materials_specular extension`) can link to the issue.

packages/engine/Source/Shaders/Model/MaterialStageFS.glsl
```diff
@@ -296,12 +296,12 @@ float setMetallicRoughness(inout czm_modelMaterial material)
...
+    material.diffuse = mix(material.baseColor.rgb, vec3(0), metalness);
```
ggetz, line 304: Shouldn't this be?

```suggestion
    material.diffuse = mix(material.baseColor.rgb, vec3(0.0), metalness);
```

jjhembd:
Thanks @ggetz! I addressed both of those comments.

ggetz:
All looks good! Thanks @jjhembd!

jjhembd:
For the record: the extra `(1 - f0)` factor has appeared in other implementations as well. See https://github.com/KhronosGroup/glTF/issues/1945 and https://github.com/KhronosGroup/glTF/pull/2022 for discussion about how it was caught and corrected in the glTF spec implementation notes.

Issue #11912, opened 2024-03-29
Improve spacing around logo and attribution elements


LisaBosCesium:
Credits (logos) and the attribution/attribution link that appear in the lower left of the screen are a bit crowded and visually misaligned. Please adjust this to look cleaner. 

For example:
![image](https://github.com/CesiumGS/cesium/assets/66366602/8ebb34ce-0857-4aa1-bf8e-aa15b8598aea)
![image](https://github.com/CesiumGS/cesium/assets/66366602/5957f1c1-5998-4e1f-8791-adb38389952b)

LisaBosCesium:
Note the layout looks a little different but also not great in Cesium Stories. 
![image](https://github.com/CesiumGS/cesium/assets/66366602/2a6f0859-678a-449b-81bb-040a67817aef)

ggetz:
@jjspace Would you be able to fit this in sometime this month?

jjspace:
Just posting some thoughts I shared offline. One of the core issues here is that Google's logo (and potentially any other logo any attribution has) has a set defined height that may or may not nicely line up with the ion logo image and by extension, any text in that same block (like the "data attribution" link). We can add specific changes for google specifically (may need a change on ion's side) but the problem is universal for every image added and it's hard to align things.

![Slice 1 (2)](https://github.com/CesiumGS/cesium/assets/8007967/c3613ae6-4264-4055-b218-0db04d069f24)
![Slice 1 (3)](https://github.com/CesiumGS/cesium/assets/8007967/c4523191-5e6d-4de7-9648-96c253c25ab6)

(stretching height is not actually as easy as it sounds given how CSS treats width and height, hard to "stretch to biggest", easier but slightly restricting to "set all to same height")

I would _like_ to come up with a universal solution the "just works" for any logo but I think that may end up being too hard. I think it would be a lot easier to come up with guidelines to either (1) modify logo attribution images to be a set height and/or (2) define how image attributions are rendered and require changes to the logo/html on ion's side to always have them render predictably.

Something like adding vertical alignment to the attribution's html styling can achieve fairly nice results in the existing element's architecture:

![2024-06-24_16-35](https://github.com/CesiumGS/cesium/assets/8007967/75c58fb1-28d2-4f61-9cce-6dd66c8e45f1)
![2024-06-24_16-00](https://github.com/CesiumGS/cesium/assets/8007967/bfd07afe-12a2-40a2-b2de-66f395194e69)


[...]

PR #12047, opened 2024-06-25
Adjust credits spacing and alignment

jjspace:
# Description

- Adjust attribution styles slightly to provide better alignment
    - All elements should lean into the `inline` nature of how they're displayed. If you have an image the recommendation now is to make sure any text in the log aligns how you want with other text using `vertical-align` or transparent padding in the image itself.
- Change `Viewer` behavior slightly to avoid long credits clipping behind the fullscreen button

## Issue number and link

Fixes: https://github.com/CesiumGS/cesium/issues/11912

## Testing plan

- Open any sandcastle and check the credits
- Open the google 3d tiles sandcastles and see how they align (will require a change in ion)
- Test with both the default ion token and your own to show with and without the long "default token" warning

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/CreditDisplay.js
+++ b/packages/engine/Source/Scene/CreditDisplay.js
@@ -163,112 +163,100 @@ function styleLightboxContainer(that) {
   }
 }
 
-function addStyle(selector, styles) {
-  let style = `${selector} {`;
-  for (const attribute in styles) {
-    if (styles.hasOwnProperty(attribute)) {
-      style += `${attribute}: ${styles[attribute]}; `;
-    }
-  }
-  style += " }\n";
-  return style;
+function appendCss(container) {
+  const style = /*css*/ `
+.cesium-credit-lightbox-overlay {
+  display: none;
+  z-index: 1;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(80, 80, 80, 0.8);
 }
 
-function appendCss(container) {
-  let style = "";
-  style += addStyle(".cesium-credit-lightbox-overlay", {
-    display: "none",
-    "z-index": "1", //must be at least 1 to draw over top other Cesium widgets
-    position: "absolute",
-    top: "0",
-    left: "0",
-    width: "100%",
-    height: "100%",
-    "background-color": "rgba(80, 80, 80, 0.8)",
-  });
+.cesium-credit-lightbox {
+  background-color: #303336;
+  color: ${textColor};
+  position: relative;
+  min-height: ${lightboxHeight}px;
+  margin: auto;
+}
+.cesium-credit-lightbox > ul > li a,
+.cesium-credit-lightbox > ul > li a:visited {
+  color: #${textColor};
+}
+.cesium-credit-lightbox > ul > li a:hover {
+  color: ${highlightColor};
+}
+.cesium-credit-lightbox.cesium-credit-lightbox-expanded {
+  border: 1px solid #444;
+  border-radius: 5px;
+  max-width: 370px;
+}
+.cesium-credit-lightbox.cesium-credit-lightbox-mobile {
+  height: 100%;
+  width: 100%;
+}
+.cesium-credit-lightbox-title {
+  padding: 20px 20px 0 20px;
+}
+.cesium-credit-lightbox-close {
+  font-size: 18pt;
+  cursor: pointer;
+  position: absolute;
+  top: 0;
+  right: 6px;
+  color: ${textColor};
+}
+.cesium-credit-lightbox-close:hover {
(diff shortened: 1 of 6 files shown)


ggetz:
@angrycat9000 Would you be able to take a quick pass on this from a CSS/style perspective?

packages/engine/Source/Scene/CreditDisplay.js
```diff
@@ -163,112 +163,100 @@ function styleLightboxContainer(that) {
...
-    "font-size": "18pt",
-    cursor: "pointer",
-    position: "absolute",
-    top: "0",
-    right: "6px",
-    color: textColor,
-  });
+  const style = /*css*/ `
+.cesium-credit-lightbox-overlay {
+  display: none;
```
jjspace, line 169: Almost all of these styles are exactly the same, Github's diff just makes it hard to read (try VSCode's if you want). I just changed to one large template string as it's functionally equivalent but _much_ easier to read and work with from a developer perspective.

angrycat9000:
> @angrycat9000 Would you be able to take a quick pass on this from a CSS/style perspective?

Yes.  I'll review first thing tomorrow.

jjspace:
> Is there a reason why the credits need to be structured like they currently are?

I agree with all your comments, I think this area could definitely use some improvement and modernization. However what's there works and I wanted to just make the smaller, more minimal change, to fix the spacing issues instead of spending the time on a full rework of these classes an elements right now.

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -1639,8 +1639,9 @@ Viewer.prototype.resize = function () {
   const timeline = this._timeline;
   let animationContainer;
   let animationWidth = 0;
-  let creditLeft = 0;
-  let creditBottom = 0;
+  let creditLeft = 5;
+  let creditBottom = 3;
```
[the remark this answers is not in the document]
jjspace: 3px is the spacing the credits are given above the timeline if the timeline is visible. This just makes sure it's that spacing from the bottom of the screen when the timeline is not there. I tried 5 too but it felt it pushed it too far up and I felt having smaller vertical spacing than horizontal spacing (in a viewer that's mostly in horizontal orientations) was totally fine and I liked it better.

packages/engine/Source/Widget/CesiumWidget.css
```diff
@@ -28,11 +28,6 @@
   padding-right: 5px;
 }
 
-.cesium-widget-credits a,
-.cesium-widget-credits a:visited {
-  color: #fff;
-}
```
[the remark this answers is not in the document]
jjspace: This rule was overwriting the hover styling that's in the `CreditsDisplay` file. I thought I had moved this into there to have it all centralized but guess that got missed somewhere, I'll take a look.

jjspace:
@angrycat9000 Thanks for the comments, I think this should be good now

ggetz:
@jjspace Can you include an update to `CHANGES.md`? I see it's checked off in the PR description, but I don't see it in the change list.

packages/engine/Source/Scene/CreditDisplay.js
```diff
@@ -351,12 +341,12 @@ function CreditDisplay(container, delimiter, viewport) {
   lightboxCredits.appendChild(creditList);
 
   const cesiumCreditContainer = document.createElement("div");
-  cesiumCreditContainer.className = "cesium-credit-logoContainer";
+  cesiumCreditContainer.className = "cesium-credit-cesiumCreditContainer";
```
ggetz, line 344: We may want to avoid changing the class names, as it would be a breaking change for anyone who may have applied custom styles. If there is a good reason to change the class names, it should be listed under "Breaking Changes".

ggetz:
Thanks @jjspace! Just one comment and this should be good to merge.

packages/engine/Source/Scene/CreditDisplay.js
jjspace: I changed this for clarity because I was _really_ confused when I first started looking at this and wondering why the Google "logo" was not in the logo container. I wasn't thinking of external devs trying to style based on this.

In general I think I would argue that this is an "internal feature" and should _not_ be used or relied on for styling in other applications. We allow you to specify the container this component is added to but I would assume we are the only ones that modify the rendering of the internal parts like this. Similar to how we treat private functions in the API that we say can break at any time without notice.

I can change it back for now but I think it might be worth having a larger discussion about this and the components we provide through the Widgets and what people can expect to have "access" to outside of this PR.

ggetz: `cesium-credit-textContainer` changes too. Can you restore that as well?

ggetz:
Looks great! Thanks @jjspace!

PR #12060, opened 2024-07-01
Fix base color calculation for 3D Tile styles

jjhembd:
# Description

After https://github.com/CesiumGS/cesium/pull/12043, we have separate properties on `czm_modelMaterial` for the material `baseColor` and the computed `diffuse` lighting result.

Since CPU styling is applied _before_ lighting calculations, it should now modify the `baseColor` rather `diffuse`. This explains a few spec failures when running locally.

## Testing plan

Run Cesium3DTilesetSpec locally

# Author checklist


ggetz:
Awesome, I can confirm this fixes the specs. Thanks @jjhembd!

Issue #6277, opened 2018-02-26
Mouse move handler does not fire when modifier keys are held


ggetz:
...even when the handler does not specify any modifier keys.

Brought up on the forum: https://groups.google.com/forum/#!topic/cesium-dev/86_Z9FTRyZQ

> I am using Cesium 1.42. When I define a screen space event listener for the MOUSE_MOVE event the event does not fire if I hold a predefined keyboard key...

> For example, if you run the below Sandcastle and you press and hold the Shift, Ctrl or Alt key the Cesium MOUSE_MOVE event will not fire but the document mousemove event will...

> https://cesiumjs.org/Cesium/Build/Apps/Sandcastle/?src=Hello%20World.html&label=Showcases&gist=b84ea232989fac2f06518c745b59d715

hpinkos:
@ggetz this is by design (although it's probably not the best choice we ever made)

I would recommend doing something like this as a workaround:

```js
function onMouseMove(movement, modifier) {
...
}

setInputAction(function(movement) {
    onMouseMove(movement);
}, Cesium.ScreenSpaceEventType.MOUSE_MOVE); 

setInputAction(function() {
     onMouseMove(movement, Cesium.KeyboardEventModifier.SHIFT);
}, Cesium.ScreenSpaceEventType.MOUSE_MOVE, Cesium.KeyboardEventModifier.SHIFT);

setInputAction(function() {
     onMouseMove(movement, Cesium.KeyboardEventModifier.CTRL);
}, Cesium.ScreenSpaceEventType.MOUSE_MOVE, Cesium.KeyboardEventModifier.CTRL);

setInputAction(function() {
     onMouseMove(movement, Cesium.KeyboardEventModifier.ALT);
}, Cesium.ScreenSpaceEventType.MOUSE_MOVE, Cesium.KeyboardEventModifier.ALT);
```

cesium-concierge:
Congratulations on closing the issue! I found these Cesium forum links in the comments above:

https://groups.google.com/forum/#!topic/cesium-dev/86_Z9FTRyZQ

If this issue affects any of these threads, please post a comment like the following:

> The issue at https://github.com/AnalyticalGraphicsInc/cesium/issues/6277 has just been closed and may resolve your issue. Look for the change in the next stable release of Cesium or get it now in the master branch on GitHub https://github.com/AnalyticalGraphicsInc/cesium.
---

[...]

PR #12073, opened 2024-07-09
fix: Movement handler stuck when clicking certain key combinations

### LINKED ISSUE #11903: Movement handler stuck when clicking certain key combinations
### What happened?

If you do the reproduction steps, you get stuck with dragging mode even when you do not hold LEFT_CLICK.

https://github.com/CesiumGS/cesium/assets/27568710/9809b620-4960-4b9b-b8ba-47f8bd3b28cd

### Reproduction steps

1. Press and hold LEFT_CLICK Down
2. Press and hold ALT Down
3. LEFT_CLICK Up
4. ALT Up

### Sandcastle example

_No response_

### Environment

Browser: Google Chrome Version 123.0.6312.59 (Official Build) (64-bit)
CesiumJS Version: 1.115
Operating System: Windows 10

EMapGIS:
# Description

fix #11903
When the mouse is pressed, the modifiers will be recorded. Modifying the modifiers during this process is invalid.

## Issue number and link

[#11903](https://github.com/CesiumGS/cesium/issues/11903)

## Testing plan

# Author checklist


ggetz:
Hi @EMapGIS, I don't see a signed CLA on file for you. We'll need one to review this PR. Could you please submit a [Contributor License Agreement](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) (CLA)?

EMapGIS:
> Hi @EMapGIS, I don't see a signed CLA on file for you. We'll need one to review this PR. Could you please submit a [Contributor License Agreement](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) (CLA)?

Sorry, I submitted a version before, but I submitted the wrong Github account. I have now resubmitted a version. Please check.

ggetz:
> I have now resubmitted a version. Please check.

@EMapGIS Perfect, thank you! I can confirm we received it.

ggetz:
@jjspace Could you please review this PR?

jjspace:
Hey @EMapGIS, thanks for the PR!

This solution does seem to fix the specific reproduction steps in the original issue however I don't think it's the right approach for a generalized answer. Changing the `ScreenSpaceEventHandler` to rely on a global variable creates an interdependence for mouse events that may not be obvious or intuitive to users and completely prevents listening to changes in the modifiers being held down throughout a drag if that's what the user _wanted_.

I believe the better approach would be to modify `ScreenSpaceCameraController` to better handle the cases where modifiers change and canceling ongoing events (like the initial drag event). I haven't dug into the the code to know the specific changes needed but I think it should be the responsibility of the default camera controller (in that class) to handle changes in modifiers similar to what's mentioned in [this comment](https://github.com/CesiumGS/cesium/issues/6277#issuecomment-368673808)

EMapGIS:
> Hey @EMapGIS, thanks for the PR!
> 
> This solution does seem to fix the specific reproduction steps in the original issue however I don't think it's the right approach for a generalized answer. Changing the `ScreenSpaceEventHandler` to rely on a global variable creates an interdependence for mouse events that may not be obvious or intuitive to users and completely prevents listening to changes in the modifiers being held down throughout a drag if that's what the user _wanted_.
> 
> I believe the better approach would be to modify `ScreenSpaceCameraController` to better handle the cases where modifiers change and canceling ongoing events (like the initial drag event). I haven't dug into the the code to know the specific changes needed but I think it should be the responsibility of the default camera controller (in that class) to handle changes in modifiers similar to what's mentioned in [this comment](https://github.com/CesiumGS/cesium/issues/6277#issuecomment-368673808)
> 
> (also please make sure CI is passing, seems there are some linting issues)

Thank you very much for your reply. I have made some changes to my code and optimized the following aspects:
1. _lastMousePassModifier was modified to be a private property of ScreenSpaceEventHandler to prevent conflicts among multiple ScreenSpaceEventHandler.
2. Added "isMousePassModifierLock" as a static property of ScreenSpaceEventHandler. When its value is true, the modifier will be locked after entering the event. By default, it is false, still maintaining the original way of Cesium. 

I have thought about what you said above about completely solving this problem. It is no small project. If there is no need to be compatible with the previous one, it would be fine. But if it needs to be compatible with the previous one, it will be very complicated. We will also sort this out to see if there is a better solution, and then we will communicate again.

jjspace:
@EMapGIS We do not need to preserve the previous behavior, that was clearly a bug. But I believe the proper solution is probably a change to the default `ScreenSpaceCameraController` not the `ScreenSpaceEventHandler`. I do not think it should lock what modifiers were held down. It should handle changing modifiers during a drag "interrupting" ongoing events better.

EMapGIS:
> @EMapGIS We do not need to preserve the previous behavior, that was clearly a bug. But I believe the proper solution is probably a change to the default `ScreenSpaceCameraController` not the `ScreenSpaceEventHandler`. I do not think it should lock what modifiers were held down. It should handle changing modifiers during a drag "interrupting" ongoing events better.

@jjspace I restored the `ScreenSpaceEventHandler`. And according to your suggestion, I modified the `CameraEventAggregator`. Currently, it is possible to modify the modifiers during the movement process, stop the previous event and execute the new event. Thank you for the ideas you provided. You can take a look at the effect of the newly modified content. 


https://github.com/user-attachments/assets/9ee9e90e-0c33-4d0c-9af4-9cc907b44efc

jjspace:
Thanks for the changes @EMapGIS I think this is much closer to what we want. I definitely like the new interaction now that it always respects the behavior of the _currently_ held modifiers.

I did have one concern when testing this. It seems the `cancel` function is getting called a _ton_ while dragging the mouse which seems unintentional/undesired. I think it's through the refresh function being aggressive with canceling actions or something but I can't quite tell. Would you be able to take a look at that?


https://github.com/user-attachments/assets/9ad5ad52-cede-4085-86c7-17e299aac0fd

EMapGIS:
@jjspace Thank you for your reply. We have located the problem and fixed it. Currently, when moving the mouse, the cancellation event will not be triggered frequently.

packages/engine/Source/Scene/CameraEventAggregator.js
```diff
@@ -219,20 +218,73 @@ function listenMouseButtonDownUp(aggregator, modifier, type) {
...
   aggregator._eventHandler.setInputAction(
     function () {
-      aggregator._buttonsDown = Math.max(aggregator._buttonsDown - 1, 0);
-      isDown[key] = false;
-      releaseTime[key] = new Date();
+      cancelMouseDownAction(getKey(type, undefined), aggregator);
+      for (const modifierName in KeyboardEventModifier) {
+        if (KeyboardEventModifier.hasOwnProperty(modifierName)) {
+          const modifier = KeyboardEventModifier[modifierName];
+          if (defined(modifier)) {
```
jjspace, line 225: Then, by extension, this should never be undefined and I believe this check is not needed either.

Looking at it further if you want to loop over the keys and values of this enum you could use [`Object.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) like this:
```js
for (const [modifierName, modifier] of Object.entries(KeyboardEventModifier)) { ... }
```

But at that point it seems you actually just want all the values which you can get with [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) and end up with code like:

```js
for (const modifier of Object.values(KeyboardEventModifier)) {
  const cancelKey = getKey(type, modifier);
  cancelMouseDownAction(cancelKey, aggregator);
}
```

packages/engine/Source/Scene/CameraEventAggregator.js
```diff
@@ -219,20 +218,73 @@ function listenMouseButtonDownUp(aggregator, modifier, type) {
...
+
 function cloneMouseMovement(mouseMovement, result) {
   Cartesian2.clone(mouseMovement.startPosition, result.startPosition);
   Cartesian2.clone(mouseMovement.endPosition, result.endPosition);
 }
 
+function refreshMouseDownStatus(type, modifier, aggregator) {
+  // first: Judge if the mouse is pressed
+  const isDown = aggregator._isDown;
+  let downFlag = false;
```
jjspace, line 255: Can we rename this to something slightly more descriptive? Like `anyButtonIsDown`

packages/engine/Source/Scene/CameraEventAggregator.js
```diff
@@ -219,20 +218,73 @@ function listenMouseButtonDownUp(aggregator, modifier, type) {
...
+  Object.keys(isDown).map((key) => {
+    if (
+      key.toString().startsWith(type) &&
+      isDown[key] &&
+      key.toString() !== getKey(type, modifier).toString()
+    ) {
+      downFlag = true;
+      cancelMouseDownAction(key, aggregator);
+    }
+  });
```
jjspace, line 265: 1. It's not good to use `map()` when not actually _mapping_ something. `forEach` is designed for looping over each item of a list.
2. `Object.keys` always returns an array of strings, no need to `toString` the keys. To avoid having to `toString` the `getKey` value I think we should modify that function itself to always return a `string` instead of sometimes a `string` and sometimes a `number` like it does now. That would be possible by just wrapping the first assignment in a template string. Please test to make sure that change doesn't break other parts of this but I don't think it should.
3. Since you want the key and the value of the `isDown` object it may be better to loop over the `Object.entries` here as well.
4. We want the value of `getKey(type, modifier)` here and down lower in the function. We also don't need to re-calculate it for every key in `isDown`. it would be better to extract it outside the loop to a `currentKey` variable.

jjspace (commented):
Thanks @EMapGIS this is looking good! I think the behavior is what we want I just had some smaller code cleanup comments

EMapGIS:
> Thanks @EMapGIS this is looking good! I think the behavior is what we want I just had some smaller code cleanup comments then we can get this in

Thanks @jjspace . I have dealt with the problem in the code.

jjspace (approved):
Thank you for the updates @EMapGIS! I made one small variable rename change but otherwise this looks good to me now

Issue #12065, opened 2024-07-03
Add screenshot testing for subtle lighting changes


jjhembd:
### Feature

Changes to the rendering pipeline can result in subtle changes to lighting, especially in models using physically-based rendering. See https://github.com/CesiumGS/cesium/issues/10346, where manual testing revealed subtle rendering differences resulting from a refactor of the pipeline.

Catching these changes can be difficult because we need to manually inspect and compare the rendering of a wide range of models. This process would be faster and more robust if we automate it via [end-to-end screenshot testing](https://github.com/CesiumGS/cesium/pull/11348).

Some ideas for these tests:
1. Render test models without any terrain or imagery in the background. This will make the tests less sensitive to network delays.
2. Use a very low tolerance for screenshot comparisons. Lighting effects can be subtle.

Issue #12028, opened 2024-06-06
Verify BRDF for image-based lighting


jjhembd:
Cesium uses the squared roughness to select the LOD of the specular environment map, which appears to be incorrect. The [glTF reference implementation](https://github.khronos.org/glTF-Sample-Viewer-Release/) uses "perceptual roughness" for environment map LOD selection.

Here are some rough comparisons in Cesium, compared to the reference implementation. Note: for these comparisons, all are using the ["Helipad Golden Hour" environment map](https://github.com/KhronosGroup/glTF-Sample-Environments/tree/main/helipad) for the specular component. But Cesium is not using the corresponding coefficients for the diffuse component.

Clearcoat Car Paint model, using squared roughness (current behavior):
![image](https://github.com/CesiumGS/cesium/assets/41167620/becdd8f0-8e2d-4f67-b72d-e9b44b88074a)

Clearcoat Car Paint model, using perceptual roughness (test):
![image](https://github.com/CesiumGS/cesium/assets/41167620/1c243a99-e8e9-4723-8453-a3ce95cd70e0)

Clearcoat Car Paint model, reference implementation:
![image](https://github.com/CesiumGS/cesium/assets/41167620/d5bbe9c4-6aae-4f7a-994a-abc23bc5fc4d)

At first glance, the squared roughness result in Cesium appears to be noisier. Also the environment map appears to be mirror-flipped relative to the reference?

Barn Lamp model, using squared roughness (this affects both LOD selection and anisotropy calculations):
![image](https://github.com/CesiumGS/cesium/assets/41167620/a1bdf713-f80f-40c3-afc2-23dac5503e48)

Barn Lamp model, using perceptual roughness:
![image](https://github.com/CesiumGS/cesium/assets/41167620/b67e44ee-6841-4d8c-9286-785ab8854264)

Barn Lamp model, reference implementation:
![image](https://github.com/CesiumGS/cesium/assets/41167620/1ce08aaa-ea41-4d74-94de-c59709cbd872)

The squared roughness result does not show as much smearing along the brush strokes in the metal, indicating that perceptual roughness should be used for the anisotropy calculations.

We will need to verify the correctness before switching. Cesium's approach has some fundamental differences with the [glTF reference implementation](https://github.khronos.org/glTF-Sample-Viewer-Release/) which can make comparisons challenging:

[...]

PR #12082, opened 2024-07-12
Correct diffuse component for image-based lighting

jjhembd:
# Description

In our image-based lighting, the diffuse component is constructed from spherical harmonic (SH) coefficients supplied by the user. These coefficients represent the irradiance coming from the environment map as a function of the surface normal. See for example [R. Ramamoorthi and P. Hanrahan, SIGGRAPH 2001](https://cseweb.ucsd.edu/~ravir/papers/envmap/envmap.pdf) (at the top of page 2):

> The irradiance _E_ is then a function of the surface normal **n** only

However, our code was using the reflection vector (the same one used for the specular component) in the calculation, rather than the surface normal. This PR corrects the diffuse calculation to use the surface normal.

During testing, I noticed that the SH coefficients used in our testing Sandcastles were incorrect. These coefficients generated a diffuse lighting that did not align with the specular component. The coefficients appear to have been generated using [google filament's cmgen](https://github.com/google/filament), but without the needed `--no-mirror` option. 

This PR updates the coefficients in the Sandcastles with corrected `cmgen` results. I also edited the inline comments to specify the use of the `--no-mirror` option.

## SH coefficient results
Here are some snapshots of the [Clearcoat Wicker test model](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/ClearcoatWicker/README.md), with the clearcoat component turned off for simplicity. The model is viewed from above (camera pointing directly down), using the [Kiara 6 Afternoon environment map](https://polyhaven.com/a/kiara_6_afternoon). Note: these images were captured _before_ correcting the lookup from the reflection vector to the normal vector.

Diffuse + specular components, old coefficients:
image

Diffuse component only, old coefficients:
image

Specular component only:
image

Diffuse component only, new coefficients:
image

Diffuse and specular components, new coefficients:
image

Notice the better alignment between the diffuse and specular components, using the new coefficients.

## Comparison of using reflection vs normal vectors for diffuse lighting
These captures use the new (corrected) coefficients, with the clearcoat turned back on.

Using reflection vector for diffuse calculation:
image

Using normal vector for diffuse calculation:
image

Note the more even lighting when using the normal vector for diffuse calculations.

## Issue number and link

Addresses part of https://github.com/CesiumGS/cesium/issues/12028.

## Testing plan

Load the "Image-based Lighting" and "glTF PBR Extensions" Sandcastles, and compare renderings with the main branch. Note that the Sandcastle code itself has changed, and some models may need to be rotated toward the brighter side. I did not update the Sandcastle orientations yet, because we will need a follow-up PR to address IBL orientation (see https://github.com/CesiumGS/cesium/issues/12028).

# Author checklist

- ~[ ] I have added or updated unit tests to ensure consistent code coverage~ See https://github.com/CesiumGS/cesium/issues/12065

### DIFF
--- a/Apps/Sandcastle/gallery/Image-Based Lighting.html
+++ b/Apps/Sandcastle/gallery/Image-Based Lighting.html
@@ -65,63 +65,63 @@
           );
         }
 
-        const environmentMapURL =
-          "https://cesium.com/public/SandcastleSampleData/kiara_6_afternoon_2k_ibl.ktx2";
         const modelURL = "../../SampleData/models/Pawns/Pawns.glb";
 
         // This environment map was processed using Khronos's glTF IBL Sampler. To process your own:
         // 1 - Download and build the Khronos glTF IBL Sampler (https://github.com/KhronosGroup/glTF-IBL-Sampler).
         // 2 - Run `cli -inputPath /path/to/image.hdr -outCubeMap /path/to/output.ktx2`. Run `cli -h` for all options.
+        const environmentMapURL =
+          "https://cesium.com/public/SandcastleSampleData/kiara_6_afternoon_2k_ibl.ktx2";
 
         // To generate the spherical harmonic coefficients below, use Google's Filament project:
         // 1 - Download the Filament release (https://github.com/google/filament/releases).
-        // 2 - Run `cmgen --type=ktx --deploy=/path/to/output /path/to/image.hdr`. Other formats are also supported. Run `cmgen --help` for all options.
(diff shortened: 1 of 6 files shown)


ggetz:
Awesome, looks much better @jjhembd! I can particularly see the benefit in the "Imaged-Based Lighting" example on the diffuse pieces: 

| main | branch |
| --- | --- |
| image | image |

jinoumoming:
There seems to be something wrong with this change
image

ggetz:
Hi @jinoumoming, that artifact was fixed in https://github.com/CesiumGS/cesium/pull/12116. It will be out with the next release. Thanks!

PR #12094, opened 2024-07-25
Add check for valid imagery provider structure for Bing Imagery

### LINKED ISSUE #12088: BingMapsImageryProvider sometimes errors
### What happened?

Our error logs show we occasionally get a type error at this line: https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/BingMapsImageryProvider.js#L101
```
Cannot read properties of undefined (reading '1')
```

I was lucky enough to catch it happen once, and could see the reason is that https://dev.virtualearth.net/REST/v1/Imagery/Metadata did not return `bbox` in the `coverageAreas`, but instead returned `jsonBoundingBox`.

image

### Reproduction steps

```
const bingRoadImageryPromise = BingMapsImageryProvider.fromUrl(
    'https://dev.virtualearth.net',
    {
        key: BING_API_KEY,
        mapStyle: BingMapsStyle.ROAD_ON_DEMAND,
    }
)
```
It also happens with other map styles.
It happens very rarely.

### Sandcastle example

_No response_

### Environment

Browser:
CesiumJS Version:
Operating System:

jjspace:
# Description

There seems to be an issue with the Bing Imagery API where it sometimes returns the bounding box as `jsonBoundingBox` instead of `bbox` as we expect. This PR adds code to skip imagery provider attributions that don't have the expected format to prevent a crash/error.

## Issue number and link

Fixes #12088

## Testing plan

1. Load [this sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVBha8IwEP0roV90ICkbTIerMl0HbjjdKh1+KIy0vbbBNK1J2tIN//vSdkNx3pfj7t67e/eCjEuFfMpjJyPhc0piEPWbyFIqAU0QqQhV6BEkLVI816hXkssjqqQhCBxpuCtY3+NIRy9RKpdj0wyhxCUVqiAMiFAJ5qB6gw703aUmdlCPkWfM0m3oh3cuuC+KOBs+qj+qehu/s/nc3TO69Bd5zG+C0dCOikU120aOiDe3+ypauEMnSe88Y3DcmZJ8o2oG43PpbRc765n9uV592k+vs5X9yzt4/Ore4x4PWktKChUIbQGH6m/LR9vre0bQ1o8ZV4RyEPp295FPJCxJDWJ8Svv1qx30L1t9pUUcuvPGwLBkI3P6988DTfNMKFRojzE2FaQ5Iwqk6RfBDhQOpGyoDdQyT6lWSEtEw8kFwShgREo9iQrGNvQLPGNqmRr/j8q0WK15XYJgpG5gyfV02TUxxpapy8tMlWXMJ+Js8w8)
2. Click `Run` a number of times and make sure there are no errors in the console.
3. May be helpful to add something like `console.log("metadata success", resource.imageryProviders, validProviders);` to the `metadataSuccess` function to check when the filter is actually removing results.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/BingMapsImageryProvider.js
+++ b/packages/engine/Source/Scene/BingMapsImageryProvider.js
@@ -120,11 +120,13 @@ function metadataSuccess(data, imageryProviderBuilder) {
   imageryProviderBuilder.maximumLevel = resource.zoomMax - 1;
   imageryProviderBuilder.imageUrlSubdomains = resource.imageUrlSubdomains;
   imageryProviderBuilder.imageUrlTemplate = resource.imageUrl;
-  const validProviders = resource.imageryProviders.filter((provider) =>
-    // prevent issues with the imagery API from crashing the viewer when the expected properties are not there
-    // See https://github.com/CesiumGS/cesium/issues/12088
-    provider.coverageAreas?.some((area) => defined(area.bbox))
-  );
+  const validProviders = !defined(resource.imageryProviders)
+    ? resource.imageryProviders
+    : resource.imageryProviders.filter((provider) =>
+        // prevent issues with the imagery API from crashing the viewer when the expected properties are not there
+        // See https://github.com/CesiumGS/cesium/issues/12088
+        provider.coverageAreas?.some((area) => defined(area.bbox))
+      );
   imageryProviderBuilder.attributionList = validProviders;
 }
 
--- a/packages/engine/Source/Scene/BingMapsImageryProvider.js
+++ b/packages/engine/Source/Scene/BingMapsImageryProvider.js
@@ -120,13 +120,15 @@ function metadataSuccess(data, imageryProviderBuilder) {
   imageryProviderBuilder.maximumLevel = resource.zoomMax - 1;
   imageryProviderBuilder.imageUrlSubdomains = resource.imageUrlSubdomains;
   imageryProviderBuilder.imageUrlTemplate = resource.imageUrl;
-  const validProviders = !defined(resource.imageryProviders)
-    ? resource.imageryProviders
-    : resource.imageryProviders.filter((provider) =>
-        // prevent issues with the imagery API from crashing the viewer when the expected properties are not there
-        // See https://github.com/CesiumGS/cesium/issues/12088
-        provider.coverageAreas?.some((area) => defined(area.bbox))
-      );
+
+  let validProviders = resource.imageryProviders;
+  if (defined(resource.imageryProviders)) {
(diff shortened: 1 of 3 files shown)


ggetz:
@jjspace I'm seeing some test failures in CI: `resource.imageryProviders is undefined`.

ggetz:
Otherwise, the updates look good to me, thanks! Let's just ensure these tests are passing and we should be good to go.

jjspace:
@ggetz Just pushed up a fix for the tests. Some were real errors where I wasn't handling undefined. The one in the actual `BingMapsImageryProviderSpec` for checking the cache is a bit weird though. Trying to follow it through the way that class is doing "caching" feels a bit wrong and the issue is we're doing transformation to the result _after_ retrieving from the cache so checking that objects are exactly the same is odd for the test. I made the change that works and preserves the test but overall I think that workflow could use adjustments beyond the scope of this PR.

packages/engine/Source/Scene/BingMapsImageryProvider.js
```diff
@@ -120,7 +120,14 @@ function metadataSuccess(data, imageryProviderBuilder) {
...
   imageryProviderBuilder.imageUrlTemplate = resource.imageUrl;
-  imageryProviderBuilder.attributionList = resource.imageryProviders;
+  const validProviders = !defined(resource.imageryProviders)
+    ? resource.imageryProviders
+    : resource.imageryProviders.filter((provider) =>
+        // prevent issues with the imagery API from crashing the viewer when the expected properties are not there
+        // See https://github.com/CesiumGS/cesium/issues/12088
+        provider.coverageAreas?.some((area) => defined(area.bbox))
+      );
+  imageryProviderBuilder.attributionList = validProviders;
```
ggetz, line 130: Not trying to bikeshed, but I think this line is getting a bit clunky in terms of readability. Can we simplify? ie

```suggestion
  let validProviders;
  if (defined(resource.imageryProviders)) {
    // prevent issues with the imagery API from crashing the viewer when the expected properties are not there
    // See https://github.com/CesiumGS/cesium/issues/12088
    validProviders = resource.imageryProviders.filter((provider) =>
        provider.coverageAreas?.some((area) => defined(area.bbox))
      );
  }
  imageryProviderBuilder.attributionList = validProviders;
```

jjspace: I don't know if I find it personally more or less readable than what I had but no need to go back and forth. Updated

ggetz:
Thanks @jjspace!

PR #12099, opened 2024-07-30
Make time parameter optional for Property

### LINKED ISSUE #12009: `time` param of CallbackProperty.Callback should be optional
### Feature

According to the [current signature](https://cesium.com/learn/cesiumjs/ref-doc/CallbackProperty.html#.Callback), the callback has 2 params:
1. `time`, of type `JulianDate`
2. `result`, an optional object

This implies that every CallbackProperty is time dependent. But this is not necessarily the case, as shown in [this Sandcastle](https://sandcastle.cesium.com/#c=rVLBTgIxEP2VcS/bjTgbRCDAQkwwnkw0MfG0l9IdsKHbkraLouHf7e4KMYon7WHSmb735nVaYbTzsJX0QhamoOkF5uRkVeJTU2N5JJp8brTnUpPNo2SS61yLhqn4glQgtgpI2ksvySEvCvaea4CNcaFi9PigO+fWhx3XPVxaU97QyhI5djHsY3dw1R0Mhh3ojXDUv6xX0qlFmi5jaAQBPL368Verc67Ugov1gzUbsn7HWhwAS2A6O9G4kM5zLegIhL+4a9fnBAQvyXI8XPsISNCbW/lKBbtM4BxiKOPj2ZIrR59ZK7kPcd/OOdf/NNtg/Or3ccZvxpQg9Vn83UD90EYRKrNiDRPbWPNwRf6Jq4qYrpRKkknUiTLnd4pmh8tdy3JjrIfKKoaYeio3igeX6aISa/IonKv71NAs/UrNCrkFWUxP/EAQijsXTpah66N8ozyaZWnA/6AqwwupV/dbsorvathzd3bXFhExS0N6mumNCZ/KflP+AA).

In my local app, I have the use case that sometimes, I need to get the current value of such a property elsewhere. I can do this using `getValue(time)`, although I don't need `time`, so I just pass `undefined` instead. However, I am using Typescript strict type checking, so the ts compiler does not allow me to pass `null` or `undefined` or anything other than a JulianDate. My current workaround is to write `getValue(undefined!)` , which is syntactically correct, but looks confusing at best, but mostly it looks wrong.

Would it be possible to make the `time` parameter optional, just as the `result` parameter? Or would that break something?

anne-gropler:
# Description

The callback in a `CallbackProperty` could do anything, even ignoring the `time` parameter.

Users in typescript strict mode always have to pass a time parameter, even if their callback doesn't use it, and are forced to write something like `.getValue(undefined!)`. 

This pull requests makes the `time` paramter optional for all `Property`s, so that users can write `.getValue()` without problems. As suggested in the linked issue comments, a default for `time`  (i.e. `JulianDate.now()`) is used for compatibility. This default was added for all properties, because the abstract `Property` could otherwise not make the `time` parameter optional.

## Issue number and link

Closes #12009 

## Testing plan

Added tests that check that `JulianDate.now()` is called when `time` is `undefined`.

Built Cesium and linked it in another typescript app (strict mode) and checked that the `time` parameter is indeed optional now.

# Author checklist


packages/engine/Source/DataSources/CallbackProperty.js
```diff
@@ -49,11 +50,14 @@ Object.defineProperties(CallbackProperty.prototype, {
...
  * Gets the value of the property.
  *
- * @param {JulianDate} time The time for which to retrieve the value.
+ * @param {JulianDate} [time=now] The time for which to retrieve the value. If omitted, now is used.
  * @param {object} [result] The object to store the value into, if omitted, a new instance is created and returned.
  * @returns {object} The modified result parameter or a new instance if the result parameter was not supplied or is unsupported.
  */
 CallbackProperty.prototype.getValue = function (time, result) {
+  if (!defined(time)) {
+    time = JulianDate.now();
```
ggetz, line 59: This will allocate a new object each time `getValue` is called without a `time` parameter. To avoid a performance hit, we should use a [scratch variable](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#result-parameters-and-scratch-variables).

packages/engine/Source/DataSources/CallbackProperty.js
```diff
@@ -49,11 +50,14 @@ Object.defineProperties(CallbackProperty.prototype, {
 /**
  * Gets the value of the property.
  *
- * @param {JulianDate} time The time for which to retrieve the value.
+ * @param {JulianDate} [time=now] The time for which to retrieve the value. If omitted, now is used.
```
ggetz, line 53: I would suggest:

```suggestion
 * @param {JulianDate} [time=JulianDate.now()] The time for which to retrieve the value. If omitted, the current system time is used.
```

just to be a bit more precise.

ggetz (changes requested):
Thanks for the PR @anne-gropler! I have a two comments which apply to each of the files changed.

packages/engine/Source/DataSources/CallbackProperty.js
anne-gropler: done!

packages/engine/Source/DataSources/CallbackProperty.js
ggetz: Close! The `JulianDate.now` function is designed to take a `result` parameter. Instead of creating a new JulianDate object, it'll be copied into the existing scratch variable:

```js
const timeScratch = new JulianDate();

/**
 * Gets the value of the property.
 *
 * @param {JulianDate} [time=JulianDate.now()] The time for which to retrieve the value. If omitted, the current system time is used.
 * @param {object} [result] The object to store the value into, if omitted, a new instance is created and returned.
 * @returns {object} The modified result parameter or a new instance if the result parameter was not supplied or is unsupported.
 */
CallbackProperty.prototype.getValue = function (time, result) {
    if (!defined(time)) {
    time = JulianDate.now(timeScratch);
  }

  return this._callback(time);
};
```

Does that make sense?

anne-gropler: Ohh, right! This does make more sense! 

I adapted the code. Thanks for your patience!

CHANGES.md
```diff
@@ -9,6 +9,7 @@
 - Added `Transforms.computeIcrfToMoonFixedMatrix` and `Transforms.computeMoonFixedToIcrfMatrix` to compute the transformations between the Moon's fixed frame and ICRF at a given time.
 - Added `Transforms.computeIcrfToCentralBodyFixedMatrix` to specific the default ICRF to fixed frame transformation to use internally, including for lighting calculations.
 - Added SplitDirection property for display PointPrimitive and Billboard relative to the `Scene.splitPosition`. [#11982](https://github.com/CesiumGS/cesium/pull/11982)
+- Made the `time` parameter optional for `Property`, using `JulianDate.now()` as default. [#12099](https://github.com/CesiumGS/cesium/pull/12099)
```
ggetz, line 12: One last thing: Please move this under a new section.

````md
### 1.121 - 2024-09-01
````

ggetz (changes requested):
Awesome, thanks @anne-gropler!

CHANGES.md
anne-gropler: new month, new section :) done!

ggetz (approved):

Issue #11176, opened 2023-03-21
i3dm renders incorrectly when model has non-identity matrix


bbbbx:
When the transformation matrix of the glTF node of i3dm is not the identity matrix,
the calculation of the bounding sphere of the model(scene graph/primitive/draw command) is wrong,
causing the far and near planes of the viewing cone to be smaller than expected,
and finally the pixels outside the bounding sphere are discarded in the fragment shader.

Screenshot:

https://user-images.githubusercontent.com/22176164/226557527-dbdad565-b8fc-4ee7-8f43-d4ead14cfa95.mp4

Sandcastle example: [url](https://sandcastle.cesium.com/#c=bVLRbpswFP0Vlz2USsRkXR8iSqNtnTRNmtSHRn3ixdg3cFdjI/uSLJv677MDWRMlRgh87zn3HB+Q1nhiG4QtOPbADGzZI3gcOv6yr6VVIvf7R2tIoAFXJRn7WxnGfCuU3fqCkRsgq8zbzX1lKjPO4l6CAd5oWwNX0FO7Ak9fmjDC0wqcCy9BL1L3LLn3QajBA50aGR+fvq3GZroXH5wuWJW0RH2R59pKoVvrqVjMF/P8R5gljAT15BCCbUJr8mk2/+WtCWeY/J647R12SLgBz4VS6cQ4gq31bmWP6/F6FkZJ4UlDZK1s02j4OhBZk15/j+dnL+ixRo20u87GtNh6MDLaSmUL8hXUzZjphfB8a7chkKurCXn/P+nKJFlSetppWEZyXJ+x662jmE/KeU7Q9VoQ+LweApm49D6SR/AHslbXwo3a46qFfG2cHYwqmGtqkd7dZuxwz/ninc1YH86LpinYXf/7qFxbp8DNnFA4+NPm25k0mn6gYwMbcIThc86ExsYUrEOlNJyrzsj2Bbs9UT60ahvi7067k3SZHwdWKtwwVA8X/nImtfA+dNaD1s/4B6pkWeYBf0bVVkTVp+Bci12EtR+XP8ci57zMw/YycwohUA7m3lH/AA)

Browser: Google Chrome  111.0.5563.65

Operating System: Windows 10

bbbbx:
data: [InstancedOrientation.zip](https://github.com/CesiumGS/cesium/files/11026476/InstancedOrientation.zip)

bbbbx:
Model bounding sphere:

[...]

Issue #10846, opened 2022-10-08
Will the ModelInstanceCollection class be restored?


zhwy:
I've noticed the ModelInstanceCollection has been removed since 1.96. It used to be placed in the Development category in the Cesium sandcastle.
![image](https://user-images.githubusercontent.com/16789591/194685821-acc8936b-49cf-480e-a70b-5ada468d3959.png)
ModelInstanceCollection is really useful and efficient to display a large amount of same models. Will it be restored in a subsequent version？

ggetz:
Hi @zhwy, `ModelInstanceCollection` was marked as a private class, so it was not an official part of the API and was subject to change.

@j9liu Is there anyway for a user to hook into model instancing without 3D Tiles?

j9liu:
Hi @zhwy,

Models can be instanced on the GPU if they contain the [`EXT_mesh_gpu_instancing` extension](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Vendor/EXT_mesh_gpu_instancing/README.md), but as of now there's no class to instance models without the extension, or instance them on the CPU.

You can still manually create a collection of `Model`s with different model matrixes -- is there a reason that you needed `ModelInstanceCollection` in particular?

zhwy:
Hi @j9liu ,

I use `ModelInstanceCollection` to generate large batches of models dynamiclly. One application is that I need to display many vehicle models based on GPS data and update each instance's model matrix regularly.

j9liu:
Hi @zhwy,

Thanks for the response. I'm not sure how much effort this would be to implement since `ModelInstanceCollection` may not be compatible with the new `Model` architecture. For now I've marked this as an enhancement to consider adding this in the future.

kikitte:
The same usage like what zhwy said, displaying large numbers of the same model. It would be nice to see that.

abearxiong:
has other way ？

ggetz:
@ptrgags Would you be able to advise on how something like ModelInstanceCollection could be implemented with the new model architecture?

yjwbill:
Using ModelInstanceCollection to creating gltf instances is efficient and it can raplace the modelMatrix of each instance.Creating hundreds of moving cars is an easy thing by using  ModelInstanceCollection ,but now i cannot do it.
Looking forward to seeing this feature again.

ptrgags:
[...]

PR #12105, opened 2024-08-01
Draft for fixing disappearing I3DM models

javagl:
# Description

Under certain conditions, parts of models in I3DM could disappear in certain view configurations. This has been reported in several forum threads. Most recently, in https://community.cesium.com/t/disappearing-i3dms-on-map-navigation-again/34035 . It is currently tracked in https://github.com/CesiumGS/cesium/issues/11176.

A short ping @bertt : Maybe you want to try this out. It's not yet tested extensively, but might fix the issue. 

---

#### For the reviewers:

Some details from investigating the _reason_ for the error are summarized in https://github.com/CesiumGS/cesium/issues/11176#issuecomment-2254227941 

The attempt of a **tl;dr** summary:

- The `boundingVolume` that is associated with a `DrawCommand` determines whether that command is executed or not (depending on the view configuration)
- For a glTF model, the bounding volume of the draw command was computed from the [`PrimitiveRenderResources.positionMin`/`positionMax`](https://github.com/CesiumGS/cesium/blob/3cdc3e1872767d607d758a1e4c5660abbadb3753/packages/engine/Source/Scene/Model/PrimitiveRenderResources.js#L235) properties. These `positionMin/positionMax` properties already include the _instancing_ translation
- The bounding sphere that was computed from these values was [transformed with the node matrix](https://github.com/CesiumGS/cesium/blob/3cdc3e1872767d607d758a1e4c5660abbadb3753/packages/engine/Source/Scene/Model/buildDrawCommand.js#L74)
- This means that the contribution of the instancing translation was affected by the node transform

For nodes with different transforms, this _had_ to be inconsistent. 

This PR is currently a **DRAFT** state, also because it contains some debug output. Some of this output - from the example that was posted at the bottom of https://github.com/CesiumGS/cesium/issues/11176#issuecomment-2260871932 - is shown here to convey the idea. It is an I3DM with 4 instances of a GLB that contains 4 primitives with different sizes that are attached to nodes with different scaling factors.

Before the fix, the output is this:
```
buildDrawCommand
  positionMin Cartesian3 {x: -5, y: -5, z: 0}
  positionMax Cartesian3 {x: 6, y: 6, z: 1}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  1
  instances transformInWorldSpace  true
  radius 7.794228634059948
buildDrawCommand
  positionMin Cartesian3 {x: -5, y: -5, z: 0}
  positionMax Cartesian3 {x: 5.1, y: 5.1, z: 0.1}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  10
  instances transformInWorldSpace  true
  radius 71.41953514270448
buildDrawCommand
  positionMin Cartesian3 {x: -5, y: -5, z: 0}
  positionMax Cartesian3 {x: 15, y: 15, z: 10}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  0.1
  instances transformInWorldSpace  true
  radius 1.5
buildDrawCommand
  positionMin Cartesian3 {x: -5, y: -5, z: 0}
  positionMax Cartesian3 {x: 6, y: 6, z: 1}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  1
  instances transformInWorldSpace  true
  radius 7.794228634059948
```

The `instancingTranslationMin/Max` are always the same. They are determined directly from the instancing information.
The computed `positionMin/Max` show differences that depend on the `node scale ...`
The most obvious (and relevant) effect is on the bounding shere radius: 

- When the node scale is 1, then the radius is 7.79
- When the node scale is 10, then the radius is 71.4
- When the node scale is 0.1, then the radius is 1.5 (which is already smaller than the instancing translations...)

Shown in this screenshot:

screenshotBefore

----

I considered different approaches for fixing this. One difficulty is...

- the `positionMin/Max` do not say whether they are in world space or object space
- some places already assume that they are in world space (and have to be transformed with the node transform to obtain their "true value")
- the `instancingTranslationMin/Max` does not say whether it is in world space. But there is a subtle hint via [`ModelComponents/Instances.transformInWorldSpace`](https://github.com/CesiumGS/cesium/blob/49da5c832f6caf8a235440ab9020c1f664bcfa85/packages/engine/Source/Scene/ModelComponents.js#L659) which says that it can be both...

The suggested fix is [summarized in this comment](https://github.com/CesiumGS/cesium/commit/d381c1dd309fe9331d5b47e1bcf10b5f7a43c267#diff-eb3a015c9afe04119db54bff6fa810de3498c2b8274ead37cc26e73821d0ae95R236). 

It's a bit... pragmatic, and having to compute the inverse of the node transform is ... not so nice. But other approaches had a larger potential for messing up other uses of the `positionMin/Max` (for which I now added a few lines of comments, by the way).

With this fix, the log output is as follows:
```
buildDrawCommand
  positionMin Cartesian3 {x: -5, y: -5, z: 0}
  positionMax Cartesian3 {x: 6, y: 6, z: 1}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  1
  instances transformInWorldSpace  true
  radius 7.794228634059948
buildDrawCommand
  positionMin Cartesian3 {x: -0.7, y: -0.5, z: 0}
  positionMax Cartesian3 {x: 0.4, y: 0.6, z: 0.1}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  10
  instances transformInWorldSpace  true
  radius 7.794228634059948
buildDrawCommand
  positionMin Cartesian3 {x: -50, y: -70, z: 0}
  positionMax Cartesian3 {x: 60, y: 40, z: 10}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  0.1
  instances transformInWorldSpace  true
  radius 7.794228634059948
buildDrawCommand
  positionMin Cartesian3 {x: -7, y: -7, z: 0}
  positionMax Cartesian3 {x: 4, y: 4, z: 1}
  instancingTranslationMin  Cartesian3 {x: -5, y: -5, z: 0}
  instancingTranslationMax  Cartesian3 {x: 5, y: 5, z: 0}
  node scale  1
  instances transformInWorldSpace  true
  radius 7.794228634059948
```

One can see that the `positionMin/Max` now all have to be transformed with the node transform, but then yield the actual, correct result - and it computes the same bounding sphere radius for all cases. The result is shown here:

screenshotAfter

(Yes, I see that the bounding sphere does not really include the whole model - where the mouse cursor is. Should I fix this? Is this my fault? We can talk about that...)

## Issue number and link

Addresses https://github.com/CesiumGS/cesium/issues/11176

## Testing plan

What remains to be investigated is the effect of other instancing transforms. Right now, the transforms only contain _translations_. I would _expect unexpected_ effects when the instancing transforms contain additional rotations ot scaling. (But... maybe not, if the `instanceTranslationMin/Max` are already computed properly and taking this into account 🤞 )

For testing: Locally load the tileset (+Sandcastle) that are attached at the bottom of https://github.com/CesiumGS/cesium/issues/11176#issuecomment-2260871932 , and check whether there is still a view configuration where parts of the model disappear. Maybe also try out some of the examples that had been posted in https://community.cesium.com/t/disappearing-i3dms-on-map-navigation-again/34035 (that's also still on my TODO list)

# Author checklist

(This is a **DRAFT** state with debug/log output and TODO comments - only for an early review)

### DIFF
--- a/packages/engine/Source/Scene/Model/PrimitiveRenderResources.js
+++ b/packages/engine/Source/Scene/Model/PrimitiveRenderResources.js
@@ -5,6 +5,7 @@ import clone from "../../Core/clone.js";
 import defined from "../../Core/defined.js";
 import ModelUtility from "./ModelUtility.js";
 import ModelLightingOptions from "./ModelLightingOptions.js";
+import Matrix4 from "../../Core/Matrix4.js";
 
 /**
  * Each node may have many mesh primitives. Most model pipeline stages operate
@@ -232,14 +233,66 @@ function PrimitiveRenderResources(nodeRenderResources, runtimePrimitive) {
    */
   this.primitiveType = primitive.primitiveType;
 
+  // For computing the `positionMin` and `positionMax` properties,
+  // the instancing translation (if present) has to be taken into
+  // account.
+  // The `positionMin` and `positionMax` are supposed to NOT take
+  // the node transform into account.
+  // So when the instancing translations are supposed to be applied
+  // in world space (as indicated by instances.transformInWorldSpace),
+  // then the instancing translations have to be multiplied with the
+  // inverse of the node transform, to eventually yield the true
+  // minimum/maximum after applying the node transform.
+
+  let instancingTranslationMin = this.runtimeNode.instancingTranslationMin;
+  let instancingTranslationMax = this.runtimeNode.instancingTranslationMax;
+
+  //*/
+  const instances = this.runtimeNode.node.instances;
+  if (defined(instances)) {
+    // TODO Is this check necessary?
+    if (
+      !defined(instancingTranslationMin) ||
+      !defined(instancingTranslationMax)
+    ) {
+      console.log("Something wrong, they should be defined... I guess...");
+    }
+    if (instances.transformInWorldSpace) {
+      const nodeTransform = this.runtimeNode.computedTransform;
+      // TODO Use scratches here
+      const inverseNodeTransform = Matrix4.inverse(
+        nodeTransform,
+        new Matrix4()
+      );
+      instancingTranslationMin = Matrix4.multiplyByPoint(
+        inverseNodeTransform,
+        instancingTranslationMin,
+        new Cartesian3()
+      );
+      instancingTranslationMax = Matrix4.multiplyByPoint(
+        inverseNodeTransform,
+        instancingTranslationMax,
+        new Cartesian3()
+      );
+    }
+  }
+  //*/
   const positionMinMax = ModelUtility.getPositionMinMax(
     primitive,
-    this.runtimeNode.instancingTranslationMin,
-    this.runtimeNode.instancingTranslationMax
+    instancingTranslationMin,
+    instancingTranslationMax
   );
 
   /**
    * The minimum position value for this primitive.
+   *
+   * This does not take into account the transform of the node that this
+   * primitive is attached to.
+   *
+   * But it does take into account any possible instancing transforms,
+   * multiplied with the inverse of the node transform. So transforming
+   * this position value with the node transform will yield the actual
+   * minimum position, including the instancing transforms.
    *
    * @type {Cartesian3}
    * @readonly
@@ -250,6 +303,14 @@ function PrimitiveRenderResources(nodeRenderResources, runtimePrimitive) {
 
   /**
    * The maximum position value for this primitive.
+   *
+   * This does not take into account the transform of the node that this
+   * primitive is attached to.
+   *
+   * But it does take into account any possible instancing transforms,
+   * multiplied with the inverse of the node transform. So transforming
+   * this position value with the node transform will yield the actual
+   * maximum position, including the instancing transforms.
    *
    * @type {Cartesian3}
    * @readonly
--- a/packages/engine/Source/Scene/Model/buildDrawCommand.js
+++ b/packages/engine/Source/Scene/Model/buildDrawCommand.js
@@ -73,8 +73,7 @@ function buildDrawCommand(primitiveRenderResources, frameState) {
 
     boundingSphere = BoundingSphere.transform(
       primitiveRenderResources.boundingSphere,
-      modelMatrix,
-      primitiveRenderResources.boundingSphere
+      modelMatrix
     );
   }
 
--- a/packages/engine/Source/Scene/Model/buildDrawCommand.js
+++ b/packages/engine/Source/Scene/Model/buildDrawCommand.js
@@ -78,6 +78,27 @@ function buildDrawCommand(primitiveRenderResources, frameState) {
     );
   }
 
+  console.log("buildDrawCommand");
+  console.log("  positionMin", primitiveRenderResources.positionMin);
+  console.log("  positionMax", primitiveRenderResources.positionMax);
+  console.log(
+    "  instancingTranslationMin ",
+    primitiveRenderResources.runtimeNode.instancingTranslationMin
+  );
+  console.log(
+    "  instancingTranslationMax ",
+    primitiveRenderResources.runtimeNode.instancingTranslationMax
+  );
+  console.log(
+    "  node scale ",
+    primitiveRenderResources.runtimeNode.computedTransform[0]
+  );
+  console.log(
+    "  instances transformInWorldSpace ",
+    primitiveRenderResources.runtimeNode.node?.instances?.transformInWorldSpace
+  );
+  console.log("  radius", primitiveRenderResources.boundingSphere.radius);
+
   // Initialize render state with default values
   let renderState = clone(
     RenderState.fromCache(primitiveRenderResources.renderStateOptions),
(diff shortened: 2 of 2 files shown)


bertt:
Quick n = 1 test with the windmills with rotor animation:

1] Issue windmill disappear on map navigation (without this PR): 

Demo: https://bertt.github.io/cesium_issues/i3dm_disappearing2/issue/

2] Testing this PR

Demo: https://bertt.github.io/cesium_issues/i3dm_disappearing2/testing_pr_12105/

Result: 

- Windmills don't disappear  👍 

- Animation rotor still works 👍

bertt:
Anther testcase I see still an instance disappear (the western one)

https://bertt.github.io/cesium_issues/i3dm_disappearing2/testing_pr_12105_kast/

kast

GLB used is quite simple: https://bertt.github.io/cesium_issues/i3dm_disappearing2/testing_pr_12105_kast/kast.glb

javagl:
Thanks for the quick test and feedback, I really appreciate that 👍 

From _very_ quickly skimming/debugstepping over the case with the "blocks" that _doesn't_ work: Yes, in this case, the bounding sphere is still wrong. It doesn't include the second block:

StillWrong

I think that this might be related to some of the points that I already mentioned

- The I3DM for my simple test only uses `POSITION`
- The I3DM for the windmills additionally normals and `SCALE`, but the scales are all (1,1,1) (!)
- The I3DM for the blocks also uses normals and `SCALE`, with the scales being (2,2,2) (!)
- The I3DM for the blocks contains an `RTC_CENTER` 

These are just observations for now, and I haven't yet identified the _reason_.

---

Admittedly, just transforming the `instanceTranslationMin/Max` with the inverse node transform was a shot from the hip. I already had some doubts (and these doubts become stronger the more I think this through), and they refer to the idea of computing the bounding sphere from `instanceTranslationMin/Max` in general.

I'll probably need pen+paper to "mathematically confirm" this, but intuitively:

- Imagine an I3DM with two instances, and `POSITION` and rotation attributes. 
- The positions are both (0,0,0). So the `instanceTranslationMinMax` will be (0,0,0) as well. They will not affect the size of the bounding sphere at all
- Imagine the rotations being 
  - 0° around the x-axis
  - 90° around the x-axis
- Imagine the GLB containing a primitive that has a translation of 100 along the y-axis
- The rotations **will** affect the bounding sheres of the primitive instance
- When _only_ looking at the `instanceTranslationMin/Max`, this rotation is never taken into account anywhere...

Well ... I could go through this with pen+paper, or just create a test model - and I'm probably lazy and do the latter. 
I know that the formal approach could already be a step towards a solution, but my gut feeling is that the solution will have to involve storing the actual transform _matrices_ of the instances ... somewhere, somehow. 
(Just like for a [ModelInstanceCollection](https://github.com/CesiumGS/cesium/issues/10846) 🤔 )

bertt:
Another thing to consider: when the used glTF has larger textures. 

For example see https://bertt.github.io/cesium_issues/i3dm_disappearing2/issue_textured_trees/ , the used model https://bertt.github.io/cesium_issues/i3dm_disappearing2/issue_textured_trees/tree_textured.glb has 2 textures to visualize the trees 

textured_trees

bertt:
Btw I'm also wondering what's the purpose of this instancingTranslationMin, instancingTranslationMax compution, as the visibility of the i3dm tile is already defined in tileset.json...

javagl:
@bertt I don't understand in how far _textures_ come into play here. I'd think that the issue is independent of the material of the objects...?

> Btw I'm also wondering what's the purpose of this instancingTranslationMin, instancingTranslationMax compution, as the visibility of the i3dm tile is already defined in tileset.json...

I also only started zooming into this. So this attempt to summarize a few things should be taken with a grain of salt:

- The code is examining each mesh primitive that is attached to a node in the glTF
- Each node may have instancing information. This instancing information can come...
  - either from an `EXT_mesh_gpu_instancing` in that node 
  - or be instancing information that was created from the I3DM data, and basically "attached to all nodes of the glTF"

  The idea here apparently is that the same mechanism is used for _both_ the `EXT_mesh_gpu_instancing` _and_ the I3DM instancing. Which leads to the issue mentioned in the forum thread, namely that `EXT_mesh_gpu_instancing` cannot work properly in I3DMs. It's not yet clear how this could (or should) be addressed...

- The code that uses the `instancingTranslationMin/Max` seems to be an attempt to optimize the number of draw calls. Imagine one glTF with a primitive that is attached to a node with a translation of (1000, 0, 0), and that uses instancing. When looking at the origin, then the instanced primitive will not have to be rendered.... _unless_ one of the instance translations is (**-1000**, 0, 0), because then, this instance would have to appear at the origin (roughly like that...)

The computation of the bounding volumes, and how they affect the execution of the draw commands, will have to be revised. In the context of the investigations here, I noticed that the bounding volumes of the models are not computed properly to begin with (regadless of the instancing). This is now tracked in https://github.com/CesiumGS/cesium/issues/12108 . So my "quick shot" in this PR will not be sufficient, but my gut feeling is that when ~"the bounding volume computation" is addressed (properly and holistically), the resulting state _could_ also already include a fix for the I3DM issue...

Issue #12108, opened 2024-08-05
Wrong bounding volume computations for model


javagl:
[This code in `ModelSceneGraph.js`](https://github.com/CesiumGS/cesium/blob/7b93161da1cc03bdc796b204e7aa51fb7acebf04/packages/engine/Source/Scene/Model/ModelSceneGraph.js#L541) is trying to compute the model bounding volume (`modelPositionMin/Max`) from the primitive (`primitivePositionMin/Max`). This is done by transforming the min/max of the primitive with the node transforms, and computing the component-wise minimum and maximum from the transformed minimums/maximums.

This will yield wrong results in most cases. 

A 2D illustration:

![Cesium BoundingVolumes 0001](https://github.com/user-attachments/assets/36ded33a-bc97-4cdf-aff4-709fef29a99f)

- Assume that there is a primitive with a min/max of (0,-1)-(6,1).
- It is once attached to a node with an identity transform (right)
- And once attached to a node with a rotation about 180° (left)
- The maximum point will become the "leftmost" one due to the rotation. But it will not contribute to the global minimum (because it was the maximum, originally).

---

The following is an archive that contains two glTF assets, and a Sandcastle for testing this:

[Cesium Model Bounding Volume Issue 0001.zip](https://github.com/user-attachments/files/16500077/Cesium.Model.Bounding.Volume.Issue.0001.zip)

The glTF assets are visually equal, and similar to the 2D case above:
- They contain one primitive, which is a box that ranges from (0.0, -0.5, -0.5) to (10.0, -0.5, -0.5) 
- The primitive is attached to node with the identity transform
- In the `translated` asset, it is also attached to a node with a translation of -10 along x
- In the `rotated` asset, it is also attached to a node with a rotation of 180° around the z-axis

So both assets look like this:

![Cesium Model Bounding Volume](https://github.com/user-attachments/assets/05cf0290-5966-452e-9282-876697f1fe22)

And gltf.report prints the proper bounding box for both of them.

The sandcastle loads both models, and prints their bounding sphere information:

![Cesium Model Bounding Volume Sandcastle](https://github.com/user-attachments/assets/59dc275c-bf3f-4aa0-9216-6cf482044381)

The bounding sphere is wrong in the case of the 'rotated' second primitive.

[...]

Issue #5684, opened 2017-07-27
Bounding sphere debug primitive cut in half at north pole


hpinkos:
![image](https://user-images.githubusercontent.com/3451886/28681119-69a55e3e-72c6-11e7-8b82-99635add500e.png)

``` javascript
var viewer = new Cesium.Viewer('cesiumContainer');
viewer.entities.add({
  position : Cesium.Cartesian3.fromDegrees(0, 90),
  model : {
    uri: '../../SampleData/models/CesiumAir/Cesium_Air.glb'
  },
    scale: 20
});
viewer.zoomTo(viewer.entities);

viewer.extend(Cesium.viewerCesiumInspectorMixin);
```

PhilippeSymbol:
Same issue here
<img width="926" height="915" alt="Image" src="https://github.com/user-attachments/assets/4775b23b-4527-44d0-b3e9-492be1bc0153" />

PR #12112, opened 2024-08-07
Fix computation of model bounding volume

javagl:
Addresses https://github.com/CesiumGS/cesium/issues/12108 

Only an early draft for now. Further details will be added here soon.

# Description

...

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12108 

## Testing plan

...

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Model/ModelSceneGraph.js
+++ b/packages/engine/Source/Scene/Model/ModelSceneGraph.js
@@ -26,6 +26,8 @@ import ModelSplitterPipelineStage from "./ModelSplitterPipelineStage.js";
 import ModelType from "./ModelType.js";
 import NodeRenderResources from "./NodeRenderResources.js";
 import PrimitiveRenderResources from "./PrimitiveRenderResources.js";
+import OrientedBoundingBox from "../../Core/OrientedBoundingBox.js";
+import Matrix3 from "../../Core/Matrix3.js";
 
 /**
  * An in memory representation of the scene graph for a {@link Model}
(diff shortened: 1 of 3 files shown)


javagl:
I'll try to post intermediate updates, even though they are **HIGHLY** preliminary. I hope that they help to illustrate _why_ all this takes far longer than it should. 

Here is the output of a debugging Sandcastle. This is similar to the one that I [posted in the issue](https://github.com/CesiumGS/cesium/issues/12108#issuecomment-2273476766). But it shows...

- the bounding volume of the first primitive in gray
- the bounding volume of the second primitive in red
- the bounding volume of the whole model in yellw. 

For the current `main` state:

Cesium Bounding Spheres 0001 old

<sub>("Hey, there is no 'gray' one! And/or the one for the model is completely wrong! And/or one of the bounding volumes seems to be missing!" - "Yeah. That's the point...")</sub>

For the current (highly preliminary) state of this PR:

Cesium Bounding Spheres 0002 newer

It looks "less wrong". In fact, it _looks_ completely correct. Yay! 🥳 

**But...** ... some ....   things are ... happening at different places. What is currently shown in that sandcastle are the _actual_ bounding spheres of the model and the primitives. Let's look at the bounding spheres that are stored in the draw commands (dense red grid): 

Cesium Bounding Spheres 0003 newer with debug

I could say that "the bounding spheres are wrong", but ... in many cases, they are not even _spheres_ 😬 Tracking down _which_ bounding spheres this are, _where_ they come from, _where_ they are modified, and _where_ they are becoming "wrong" is a challenge. It's easy to point to "some place where this might be happening". But not where it actually **is** happening. I'll keep digging.

javagl:
A small update, starting with some comments that I wrote locally in one of the files. I'm trying to summarize this as a form of "rubber duck debugging":

---

The `PrimitiveRenderResources` contains `positionMin/Max` and a bounding sphere. It does not say what these properties ARE - and specifically, which transforms they include or not.

The `ModelRuntimePrimitive` stores a clone of the bounding sphere of the `PrimitiveRenderResources`. The bunding sphere is documented as
> The bounding sphere of this primitive in object-space. 

It is not clear what the "object space" is. Is it the space of the MODEL, or the space of the PRIMITIVE? (I.e. does it include the transform of the node that the primitive is attached to or not?)

The `ModelDrawCommand` contains a bounding sphere that is computed as 
`drawCommand.boundingSphere = runtimePrimitive.boundingSphere * drawCommand.modelMatrix`
This will actually be updated in the setter of the `ModelDrawCommand.modelMatrix`, which is called as part of `ModelMatrixUpdateStage.updateRuntimeNode`. The `ModelDrawCommand.modelMatrix` at this point includes the transform
of the node that the primitive is attached to (so it is not really a `modelMatrix`, but rather a `primitiveMatrix`)

---

The bunding volumes that are computed right now include

- the instancing transforms (if present)
- the `nodeRenderResources.model.sceneGraph.components.transform`
- the `nodeRenderResources.model.sceneGraph._axisCorrectionMatrix`
- the `runtimeNode.computedTransform`

and are "correct" in terms of the visual output. But but when they are part of the `drawCommand` and will later be multiplied with the `modelMatrix`, which **also** includes the `runtimeNode.computedTransform`, then the result will be wrong. Still, they _have_ to be transformed with **a** `modelMatrix`. But one that does _not_ include the node transform. Whether or not it has to include any other elements of this transform (axis correction, `components.transform`) remains to be investigated. 

I'm tempted to either

- remove the 'node transform' part from the `ModelDrawCommand.modelMatrix` (but this will likely break a thousand other places)
- rename the `ModelDrawCommand` to `PrimitiveDrawCommand` (because that's what it **is**), and pull through the refactoring steps that this entails (but I cannot foresee what this would mean, and it may introduce other bugs...)

---

**tl;dr** The questions of 'Which bounding sphere includes which transform?' and 'Which matrix contains which transform?' are hard to answer.

javagl:
These "broken bounding volume visualizations" are actually tracked in https://github.com/CesiumGS/cesium/issues/5684 (from 2017...)

javagl:
People who encounter disappearing I3DM instances will keep being forwarded to https://github.com/CesiumGS/cesium/issues/12108. The "simple" fix that was proposed in the issue could cover about 99% of the cases, but for a "perfect" fix, there are simply too many unknowns, quirks, special cases, and interferences with whatever is or isn't happening in the context of https://github.com/CesiumGS/cesium/issues/10846 . I'll probably reopen this (or a new PR) when the model instance collection was restored.

PR #12120, opened 2024-08-12
Add missing CallbackProperty type that is supposed to be here

Levi-Montgomery:
# Description

I added a missing type, you are supposed to be able to update entity position with a callback property.

Previously you would get a type error when trying to update the entity position with a callback (a supported feature). This could be worked around by ignoring the type error but this is a proper fix.

## Testing plan

I see type error go away

# Author checklist

- [X ] I have submitted a Contributor License Agreement
- [ X] I have added my name to `CONTRIBUTORS.md`
- [ X] I have updated `CHANGES.md` with a short summary of my change
- [ X] I have added or updated unit tests to ensure consistent code coverage
- [ X] I have updated the inline documentation, and included code examples where relevant
- [ X] I have performed a self-review of my code


ggetz:
Thanks for the fix @Levi-Montgomery!

PR #12148, opened 2024-08-23
Fix package.json exports order

jjspace:
# Description

Apparently the order of keys in the exports object matters: 

> Within the ["exports"](https://nodejs.org/api/packages.html#exports) object, key order is significant. During condition matching, earlier entries have higher priority and take precedence over later entries. The general rule is that conditions should be from most specific to least specific in object order.

It's very bizarre to me that key order matters in an _object_ not an array but apparently it does. `esbuild` now seems to warn about it which is why I noticed: 

This PR just fixes that order and the warnings are gone. I don't think this changes our build and if I understand it correctly I don't think it should change downstream users but would appreciate someone double checking me on that.

## Issue number and link

No issue

## Testing plan

1. `git clean -dxf`
2. `npm install`
3. `npm run build`
4. Make sure there are no warnings

# Author checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,9 +38,9 @@
     "./Build/*": "./Build/*",
     "./Build/*.js": null,
     ".": {
+      "types": "./Source/Cesium.d.ts",
       "require": "./index.cjs",
-      "import": "./Source/Cesium.js",
-      "types": "./Source/Cesium.d.ts"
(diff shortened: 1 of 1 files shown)


ggetz:
Thanks for the fix @jjspace!

Issue #11507, opened 2023-09-06
JulianDate.toIso8601 builds a non compliant string with very small milliseconds


jfayot:
```js
console.log(Cesium.JulianDate.toIso8601(Cesium.JulianDate.fromIso8601("2011-03-25T16:00:00.00000001Z"))
```
outputs:
```sh
2011-03-25T16:00:00.10000803740695119e-7Z
```
Which is not ISO8601 compliant !

In JulianDate.js, line 787, the comment says:
```js
//Forces milliseconds into a number with at least 3 digits to whatever the default toString() precision is.
```
but toString will automatically convert the milliseconds into scientific notation when lower than 1e-6.
number.toFixed is not appropriate neither, something more sophisticated needed here...

ggetz:
Thanks for the report @jfayot! 

We'd be happy to review a fix if you or anyone else has the bandwidth to contribute!

Alforoan:
can I take this please?

jfayot:
> can I take this please?

Sure !

jjhembd:
[...]

PR #12155, opened 2024-08-26
Build an ISO compliant string - 11507

Alforoan:
# Description

The function now properly converts the milliseconds into a fractional second with the specified precision, avoiding scientific notation.

## Issue

Fixes https://github.com/CesiumGS/cesium/issues/11507

## Testing plan

I ran the file and made sure it could run without bugs.

# Author checklist


ggetz:
Thanks @Alforoan! I can confirm we have a CLA on file from you now.

ggetz:
@jjspace Can you please review?

packages/engine/Source/Core/JulianDate.js
```diff
@@ -783,40 +783,16 @@ JulianDate.toIso8601 = function (julianDate, precision) {
...
-    .replace(".", "")
-    .slice(0, precision);
+  if (millisecond !== 0 || defined(precision)) {
+    if (!defined(precision)) {
+      precision = 7; // Default precision if not provided
+    }
+    // Forces milliseconds into a number with the specified precision, without scientific notation
+    const fractionalSeconds = (second + millisecond * 0.001).toFixed(precision);
+    const parts = fractionalSeconds.split('.');
+    millisecondStr = parts.length > 1 ? parts[1] : '';
```
jjspace, line 793: Can you expand on why you're adding the seconds here then splitting on the `.`? I think there may potentially be other, better, ways to get the decimal part of a number if that's the desire.

Also the original issue called out that `toFixed` does not necessarily eliminate the issue of getting scientific notation when converting a number to string if the precision is right (see [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision)). Does this addition avoid that somehow? or is that not accounted for?

jjspace (changes requested):
Thanks @Alforoan, it's a good start I just had some comments.

Also can you please add or update the tests for this class to verify the new changes work as expected (and so we know they don't break in the future)? A test that specifically uses the value shown to cause the error in the [original issue](https://github.com/CesiumGS/cesium/issues/11507) would be good

ggetz:
Hi @Alforoan! Is the plan to still move forward with this PR? Do you need any additional help to get this over the finish line?

packages/engine/Source/Core/JulianDate.js
Alforoan: I am splitting on . to separate the fractional part (milliseconds) from the whole seconds.
I am not getting scientific notations for very small numbers.

jjspace: My initial thought is that splitting on `.` is a bad way to manipulate numbers but looking into it a bit this seems like a fine way to do it. However, I'm still unclear about why you're adding the seconds to it first if you just strip them off.
It's possible there's a more simple solution that _does_ include combining the seconds and milliseconds that looks something like this `(n * 0.001 + 45).toFixed(precision).padStart(2 + precision + 1, '0')`. Just pay special attention to the padding for the seconds.

Also it looks like we have constants for some of these "magic numbers", please swap out `0.001` with `TimeConstants.SECONDS_PER_MILLISECOND` as we use elsewhere in this code.

packages/engine/Source/Core/JulianDate.js
```diff
@@ -783,40 +783,16 @@ JulianDate.toIso8601 = function (julianDate, precision) {
...
-
-  //Forces milliseconds into a number with at least 3 digits to whatever the specified precision is.
-  millisecondStr = (millisecond * 0.01)
-    .toFixed(precision)
-    .replace(".", "")
-    .slice(0, precision);
+  if (millisecond !== 0 || defined(precision)) {
+    if (!defined(precision)) {
+      precision = 7; // Default precision if not provided
+    }
```
jjspace, line 789: The previous version of this function did _not_ use a default precision. Any revisions should also not use a default precision to avoid breaking existing applications.

jjspace (changes requested):
1. Please check the failing tests and make sure any changes you've made allow them to still pass
2. Please add new tests in `JulianDateSpec.js` specifically for these changes and to ensure milliseconds are preserved as outlined in the original issue
    - You can check our [Testing Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/TestingGuide) for more info on how to add and run these
 3. It looks like you skipped running prettier somehow. please run `npm run prettier` and also make sure the pre-commit git hook is set up correctly.

CONTRIBUTORS.md
```diff
@@ -408,7 +408,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Vladislav Yunev](https://github.com/YunVlad)
 - [Levi Montgomery](https://github.com/Levi-Montgomery)
 - [Brandon Berisford](https://github.com/BeyondBelief96)
+- [Jay Yong](https://github.com/Alforoan)
 - [Lawrence Owen](https://github.com/ljowen)
-- [Adam Wirth](https://github.com/adamwirth)
```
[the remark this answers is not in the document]
Alforoan: That was a mistake; thanks for catching that and I added him back.

CHANGES.md
```diff
@@ -70,6 +70,7 @@ This is an npm-only release to extra source maps included in 1.121
 - Fixed environment map transform for image-based lighting. [#12091](https://github.com/CesiumGS/cesium/pull/12091)
 - Updated geometric self-shadowing function to improve direct lighting on models using physically-based rendering. [#12063](https://github.com/CesiumGS/cesium/pull/12063)
 - Prevent Bing Imagery API format issues from throwing errors [#12094](https://github.com/CesiumGS/cesium/pull/12094)
+- Fix JulianDate to always generate valid ISO strings for fractional milliseconds [#11507](https://github.com/CesiumGS/cesium/issues/11507)
```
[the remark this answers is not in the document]
Alforoan: would that be under ### 1.122 - 2024-10-01?

packages/engine/Source/Core/JulianDate.js
Alforoan: Removed.

CHANGES.md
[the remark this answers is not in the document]
jjspace: yes, that will be the next release

ggetz:
@Alforoan Is this ready for another look?

ggetz:
I'm closing this issue due to inactivity. If you believe this is still an issue, please feel free to re-open. Thanks!

Issue #10486, opened 2022-06-27
ES6 Modernization: when to update old usage patterns?


ptrgags:
As CesiumJS relaxes ES5-only syntax (see https://github.com/CesiumGS/cesium/pull/10396) and eventually allows more modern features, the question is what to do about the existing code, when does it make sense to update old patterns to use newer features?

To take a concrete example, consider the `AssociativeArray` class. This is a class that's designed to store key-value pairs and allow iteration. This sounds very much like an [ES6 `Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) in terms of what operations it supports. Though this leads to questions like:

* Should we remove `AssociativeArray` and replace it with `Map`? (after considering overall performance impact (not microbenchmarks) and how much it impacts the public API).
* If we need a backwards-compatible interface, should the `AssociativeArray` just store a `Map` and delegate functions?

Another example: There are spots in the code where a dictionary is used like a set (need a collection of keys with `O(1)` lookup time):

* In `CustomShader`, I used a dictionary with dummy values for storing shader variables (e.g. `{key1: true, value: true}`), though a `Set` would have had a clearer intent that I just need it for fast key lookups. (This might be a bad example, as set union/intersection would be helpful as well for comparing sets of shader variables more like a Python `set`)
* In `ModelOutlineLoader`, the edges are stored in a strange data structure: an array used like a dictionary except index 0 stored the original vertex count... The `ModelExperimental` replacement code for this feature will use the dictionary approach (which can be easily swapped out for a `Set`) 
* Where else? We'd have to search through the code to see what other parts of the code fake a set data structure using an `AssociativeArray`, a dictionary, or something else.

jjhembd:
Here's another pattern that appears frequently in our code:
```javascript
for (const key in myObject) {
  if (myObject.hasOwnProperty(key)) {
    const value = myObject[key];
    // do something with value
  }
}
```
This pattern loops unnecessarily over inherited properties.

We could now use the ES8 method [Object.values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values), which only returns 'own' properties), and skip the call to `.hasOwnProperty()`:
```javascript
for (const value of Object.values(myObject)) {
  // do something with value
}
```
[...]

PR #12188, opened 2024-09-11
Improve documentation and readability in Scene and depth-related code

jjhembd:
# Description

This PR is a first step towards improved maintainability in some of the core rendering code, especially code that touches depth buffers. The focus is on classes like `Scene`, `View`, `GlobeDepth`, `Framebuffer`, and `DerivedCommand`, with minor documentation updates in connected classes.

Main changes include:
- Added documentation for private functions
- Simplified function signatures (to be more documentable)
- Broke `executeCommands` in `Scene` into smaller functions, borrowing ideas (and some code!) from https://github.com/CesiumGS/cesium/pull/12058
- Cleaned code to better follow the Coding Guide, especially this point:

> Declare variables where they are first used.

## Testing plan

- Run all specs locally
- Spot check Sandcastles

# Author checklist


packages/engine/Source/Scene/Scene.js
```diff
@@ -2410,31 +2395,168 @@ function executeCommands(scene, passState) {
...
+    invertClassification = scene._invertClassification;
+  }
+
+  const executeTranslucentCommands = obtainTranslucentCommandExecutionFunction(
+    scene
+  );
+
+  context.uniformState.updatePass(Pass.TRANSLUCENT);
+  const commands = frustumCommands.commands[Pass.TRANSLUCENT];
+  commands.length = frustumCommands.indices[Pass.TRANSLUCENT];
```
javagl, line 2437: I know that this was **not changed** in this PR (i.e. it has already been like that before the changes here). But I recently asked about that line, and for me, this _strongly_ looks like an ancient copy-and-paste error: Usually, the `length` is fetched from this `indices` array. But here, the length is **assigned to** the array of commands. Are we sure that this is correct and makes sense?

javagl (commented):
First, a big 👍: I'm strongly convinced that there should (have been) more cleanup passes like this. Otherwise, the effect is commonly referred to as https://en.wikipedia.org/wiki/Lava_flow_(programming) . People are afraid of _touching_ code (that it hard to read and fully understand), out of the fear of breaking things. And whatever they are going to do instead: It can not improve the situation. It can only make it worse. (Now iterate that...)

In terms of the review, a general disclaimer: When ~"code that needs to be cleaned up" eventually **is** cleaned up, it can be really hard to look at the change alone and be 100% sure that it does not introduce changes in the behavior (including bugs). But multiple reviews and tests should make "serious" bugs less likely.

---

I did look over the changes via the commits, the "Files changed" as a whole, and for some  of them, at the new state only (locally in VSCode).

- Some commits seem to only address the "Declare variables where they are first used." guideline. That's good. These nearly-global variables also bugged me. The `let` always raises the questions: Will it change? And where? And how? We might even consider allocating some time/cycles to _just do this_, mechanically, as part of maintenance (orchestrated properly, to avoid merge conflicts)
- Other commits did further structural cleanups. Some of them may be subjective. 
  - I'm usually not using the `const { destructuring, assignment } = syntax;`, but ... it's concise and clean, so hard to argue against. 
  - I also prefer "early returns" when they are used as "watchdogs" at the beginning of a function, and decrease the indentation level, and have seen that this was done in a few places
  - Pulling out `repetitive.parts.of.property.accesses` seems uncontoversial
  - Similarly, pulling out common blocks  into functions that have _names_ and _documentation_
- I've been staring at https://github.com/CesiumGS/cesium/pull/12188/commits/ad274f8a0f3b0f35d74ace17b8ee19ec56ae68e2 for quite a while. And I hesitate to ask 😆...

There are some minor inlined comments, but nothing critical, I guess.

---

I also compiled and started that state locally, and tried out _some_ of the Sandcastles. (I did not try _all_ of them, and all possible interactions, but ... focussed on some of the "rendering-heavy" ones with sophisticated features, like Depth-Of-Field and whatnot). I did not notice any issues.

---

(I also ran the tests locally, and some of them did fail, **but** I'm currently having some network connection issues, so that doesn't mean anything for this PR...)

----

Unrelated to this PR, but maybe as sorts of follow-ups:

- The Coding Guide does not say anyhing about "destructuring assignment". Is this preferred or discouraged, or does it not matter?
- One of the more obscure abominations of what is considered to be "JavaScript code" is that `for (const key in obj) {  if (!obj.hasOwnProperty(key)) { ... } }`. Should the Coding Guide recommend `for (const key of Object.keys(obj))` instead? (I'd say: YES!)
- Some commits removed unused  parameters. Maybe this should be checked via linting? (VSCode points it out, at least...)

packages/engine/Source/Scene/PostProcessStage.js
```diff
@@ -93,9 +93,11 @@ import PostProcessStageSampleMode from "./PostProcessStageSampleMode.js";
  */
 function PostProcessStage(options) {
   options = defaultValue(options, defaultValue.EMPTY_OBJECT);
-  const fragmentShader = options.fragmentShader;
-  const textureScale = defaultValue(options.textureScale, 1.0);
-  const pixelFormat = defaultValue(options.pixelFormat, PixelFormat.RGBA);
+  const {
```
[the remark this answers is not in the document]
jjhembd: Good point. We do have to consider `null` when working with DOM elements, since [getElementById](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById) etc. may return `null` if nothing is found.

In this case, I think the destructuring with defaults is OK, because:
- The user would have to explicitly set `null` as a value. (Leaving it blank would result in the default being set).
- We immediately check the type of these options, so the user would get a meaningful error message if they set a value to 'null'.

packages/engine/Source/Scene/Scene.js
jjhembd: > Are we sure that this is correct and makes sense?
I'm not sure about either of those, but I would say that it is currently necessary.

The structure of `FrustumCommands`, as modified by the `insertIntoBin` function in `View`, does not (obviously) imply that we can depend on the `commands` array to have a useful `.length` property. However, the `executeTranslucentCommands` function eventually may pass the `commands` array to `mergeSort`, which _does_ assume that the array should be sorted along its `length`. Alternatively, it will be sent to `OIT.prototype.executeCommands`, which also uses the length.

I didn't want to change `Core/mergeSort` in this PR, so for now we do need to set the `.length` property. `performVoxelsPass` has the same issue.

packages/engine/Source/Scene/Scene.js
```diff
@@ -1787,69 +1792,66 @@ function updateDerivedCommands(scene, command, shadowsDirty) {
...
-  const useLogDepth = frameState.useLogDepth;
   const useHdr = this._hdr;
-  const derivedCommands = command.derivedCommands;
   const hasLogDepthDerivedCommands = defined(derivedCommands.logDepth);
   const hasHdrCommands = defined(derivedCommands.hdr);
   const hasDerivedCommands = defined(derivedCommands.originalCommand);
   const needsLogDepthDerivedCommands =
     useLogDepth && !hasLogDepthDerivedCommands;
   const needsHdrCommands = useHdr && !hasHdrCommands;
+  // TODO: if (!useLogDepth && !useHdr), why do we need derived commands?
```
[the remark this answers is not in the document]
jjhembd: I took another look, but could not confirm that the line in question is a typo. If shadow commands are needed, we do need to set the flag, even with no log depth or HDR.
I suspect the current logic may occasionally result in unnecessary calls to `updateDerivedCommands`, so I converted the TODO to a PERFORMANCE_IDEA.

jjhembd:
@javagl I think I addressed all your feedback. I also ran the end-to-end tests, comparing all the Sandcastle screenshots against the current main branch. No problems were detected.

jjhembd:
@javagl for your more general comments,
- The test failure fixed by https://github.com/CesiumGS/cesium/commit/ad274f8a0f3b0f35d74ace17b8ee19ec56ae68e2 stumped me too! But that commit is reverting to the code as it was before this PR.
- Destructuring assignment is not mentioned in the Coding Guide: good point! I opened https://github.com/CesiumGS/cesium/issues/12196 to discuss further.
- The pattern `for (const key in obj) {  if (!obj.hasOwnProperty(key)) { ... } }` was discussed [here](https://github.com/CesiumGS/cesium/issues/10486#issuecomment-1175567773). The hesitation to change it arises from the knowledge that `for...of` is known to be slower than `for (let i = 0; i < length; i++)`. I haven't yet seen a benchmark comparing `for...of` vs. `for...in`, especially when that `Object.hasOwnProperty()` call is factored in. Perhaps we need to generate these benchmarks ourselves. Or perhaps the current pattern is an artifact of some prehistoric premature optimization that we should abandon. But I think that is out of scope for this PR.
- Unused parameters _could_ be checked via `eslint`, but I think it might require too many exceptions. For example, we force all the `*PipelineStage`s to have the same signature for their `process` method, so we can loop over them without needing to know what kind of stages are in the pipeline at every given moment. But this necessarily means that some pipeline stages will not use all the arguments--see `AlphaPipelineStage`.

javagl:
> The test failure fixed by https://github.com/CesiumGS/cesium/commit/ad274f8a0f3b0f35d74ace17b8ee19ec56ae68e2 stumped me too!

It may be a detail, but ... you know, I'm a strickler. To be clear: Unless I'm overlooking something obvious, the code change in this commit should not have an effect. The behavior should be the same in both cases. Everybody could (and should be able to!) throw in such a change into a PR, as part of a cleanup, and it should _not_ cause randomly and inexplicably cause the build to break.

Do you happen to know (whether the test failure was _really_ related to that, and not just a random fluke, and) what _exactly_ the failure was? (It may not be worth spending additional time for that, but it caused quite a 🤨 for me...)

---

> Unused parameters could be checked via eslint, but I think it might require too many exceptions. 

I guess that this is somewhat related to the concept of an "interface" that does not exist on the same level as it does in a proper(ly) typed language. There probably _are_ many places where parameters are passed in, just to match an "interface". There probably are solutions for that. (Maybe something like letting the parameter name start with an `_underscore` so that eslint does not complain about it being unused). 
Regardless of that, one could consider _enabling_ this rule, once, locally, and cleaning up all unused parameters that _can_ be cleaned up, and then _disabling_ it again, for the CI. Not really important, though...


---

Beyond the scrope of (and somewhat unrealated to) this PR, but a few short notes:

> The pattern ... was discussed https://github.com/CesiumGS/cesium/issues/10486#issuecomment-1175567773.

The pattern of
`for (const key in myObject) ... hasOwnProperty .. const value = myObject[key];`
has the obscure part of the `hasOwnProperty` call, and the _additional_ overhead of another "map lookup" via `myObject[key]`. 

In terms of alternatives (or rather: possible improvements), there are many things to consider. Just quick bullet points:

- `Object.keys(...)` creates an array. This may be prohibitively expensive and generate a lot of garbage. Who knows...
- The `for (key in object)` pattern _might_ internally also create an array. Or it boils down to some `Iterable` construct. Who knows...
- When only the _values_ are required, then `Object.values(....)` may be beneficial. It avoids the lookup (but still creates an array)


One important question are _modifications_. I just did a quick test. The behavior is not tooo surprising, but ... subtle.

The `runModify....` functions iterate over keys/values in different ways, and they all add/modify a property.

- `runModifyA` iterates over the properties "the old way", and adds a property. Will this be picked up in the iteration?
  - No. That's OK. But it suggests that the set of keys _also_ is "stored" somewhere, like an array...
- `runModifyA` and `runModifyB` iterate over the properties, and change the value of a property. Will the modified value be visible during the iteration? 
  - Yes, of course
- `runModifyC` iterates over the `Object.values`. Will changes in the values be picked up during the iteration?
  - No, of course not! The `Object.values` creates a "snapshot" of the array when the iteration starts

Now, breaking that down to machine code and examining the performance- and memory implications for objects with 10, 1000, 100000 properties is a different story, but I was curious about the basics. You never know...

```javascript
function Example() {
  this.propertyA = "A";
  this.propertyB = true;
  this.propertyC = 123;
}

function runModifyA(example) {
  for (const key in example) {
    if (example.hasOwnProperty(key)) {
      const value = example[key];
      console.log(value);
      if (value === true) {
        example.propertyC = "CHANGED_VALUE";
        example.propertyD = "ADDED_VALUE";
      }
    }
  }
}

function runModifyB(example) {
  for (const key of Object.keys(example)) {
    const value = example[key];
    console.log(value);
    if (value === true) {
      example.propertyC = "CHANGED_VALUE";
      example.propertyD = "ADDED_VALUE";
    }
  }
}

function runModifyC(example) {
  for (const value of Object.values(example)) {
    console.log(value);
    if (value === true) {
      example.propertyC = "CHANGED_VALUE";
      example.propertyD = "ADDED_VALUE";
    }
  }
}

function print(example) {
  for (const value of Object.values(example)) {
    console.log(value);
  }
}

const exampleA = new Example();
const exampleB = new Example();
const exampleC = new Example();

console.log("runModifyA");
runModifyA(exampleA);
console.log("\n");

console.log("result:");
print(exampleA);
console.log("\n");
console.log("\n");

console.log("runModifyB");
runModifyB(exampleB);
console.log("\n");

console.log("result:");
print(exampleB);
console.log("\n");
console.log("\n");

console.log("runModifyC");
runModifyC(exampleC);
console.log("\n");

console.log("result:");
print(exampleC);
console.log("\n");
console.log("\n");
```

jjhembd:
@javagl 

> Do you happen to know (whether the test failure was really related to that, and not just a random fluke, and) what exactly the failure was?

Here is the spec that breaks when that last `frustumSplits` assignment is pulled out of the loop:
```bash
1) disables frustum culling
     Scene/Model/Model frustum culling
     TypeError: Cannot read properties of undefined (reading 'far')
    at View.createPotentiallyVisibleSet (packages/engine/Source/Scene/View.js:440:69 <- Build/CesiumUnminified/Cesium.js:213675:72)
```

For completeness, here is the change I make to get that error. In `View`, starting at line 434,
```javascript
  for (let j = 0; j < numFrustums; ++j) {
    frustumSplits[j] = frustumCommandsList[j].near;
    //if (j === numFrustums - 1) {
    //  frustumSplits[j + 1] = frustumCommandsList[j].far;
    //}
  }
  frustumSplits[numFrustums] = frustumCommandsList[numFrustums - 1].far;
```

The failure (and the fix) have been 100% repeatable for me over many runs, so I don't think it's a random fluke.

Your `runModify` tests might give a clue? Perhaps some hidden state of `frustumCommandsList` is stored somewhere for the loop, and changes after modifying previous elements? But I haven't been able to track it down.

javagl:
The reason for why that commit made sense was (of course, embarrassingly obvious in hindsight): For the case of `numFrustums === 0`, it accessed the `frustumCommandList[-1]`. 
(With the version that works, the `frustumSplits` are a 1-element array that only contains `undefined`, which ... may not make sense either, but apparently, these `frustumSplits` are only used for the `DebugCameraPrimitive`, so this may usually not cause an issue...)

jjhembd:
Thanks, that makes sense @javagl! Basically if `numFrustums == 0`, we don't even execute the loop so we avoid the problem.

@ggetz do you want to take a second pass at this? I think it's ready to go.

packages/engine/Source/Scene/Scene.js
```diff
@@ -1787,69 +1792,66 @@ function updateDerivedCommands(scene, command, shadowsDirty) {
...
-  const useLogDepth = frameState.useLogDepth;
   const useHdr = this._hdr;
-  const derivedCommands = command.derivedCommands;
   const hasLogDepthDerivedCommands = defined(derivedCommands.logDepth);
   const hasHdrCommands = defined(derivedCommands.hdr);
   const hasDerivedCommands = defined(derivedCommands.originalCommand);
   const needsLogDepthDerivedCommands =
     useLogDepth && !hasLogDepthDerivedCommands;
   const needsHdrCommands = useHdr && !hasHdrCommands;
+  // PERFORMANCE_IDEA: if (!useLogDepth && !useHdr), we may be able to save a call to updateDerivedCommands.
```
ggetz, line 1820: This may be up for debate, but I'm generally not a fan of leaving TODO comments like this in the code. It's too easy to forget about them and then they never get implemented. Do you think this will indeed be useful for a future developer?

packages/engine/Source/Scene/Scene.js
```diff
@@ -1871,31 +1873,36 @@ const requestRenderModeDeferCheckPassState = new Cesium3DTilePassState({
 
 const scratchOccluderBoundingSphere = new BoundingSphere();
 let scratchOccluder;
-
+/**
+ * Get the central body occluder for the scene.
+ *
+ * TODO: The occluder is the top-level globe. When we add
```
ggetz, line 1879: Similar comment about TODO comments in the code. Would this be more helpful if it lived in an issue instead?

packages/engine/Source/Scene/Scene.js
```diff
@@ -1998,18 +2005,31 @@ Scene.prototype.updateFrameState = function () {
 };
 
 /**
+ * Check whether a draw command will render anything visible in the current Scene,
+ * based on its bounding volume.
+ *
+ * @param {DrawCommand} [command] The draw command
+ * @param {CullingVolume} cullingVolume The culling volume of the current Scene.
```
ggetz, line 2012: The order of "optional, required, optional" parameters is non-ideal. Again, not strictly required for this PR, but it would be best if we ordered the signature with 'required, optional, optional".

ggetz (commented):
A few comments:

packages/engine/Source/Scene/DerivedCommand.js
```diff
@@ -12,89 +12,77 @@ const fragDepthRegex = /\bgl_FragDepth\b/;
...
-        vertexShaderSource: shaderProgram.vertexShaderSource,
-        fragmentShaderSource: fs,
-        attributeLocations: attributeLocations,
-      }
-    );
+  const usesLogDepth = fs.defines.indexOf("LOG_DEPTH") >= 0;
+
+  if (!writesDepthOrDiscards && !usesLogDepth) {
+    const source =
+      "void main() \n" + "{ \n" + "    out_FragColor = vec4(1.0); \n" + "} \n";
```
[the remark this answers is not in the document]
jjhembd: Yes, some of this was a little messy. I went ahead and updated to template strings in this file.

packages/engine/Source/Scene/Scene.js
jjhembd: In this case, I couldn't find any discussion about "support for multiple central bodies," so I don't think an issue would be appropriate.
I changed the comment to simply describe the assumptions made by this method, so that if someone _does_ want to discuss multiple bodies, they will understand what to change.

jjhembd:
@ggetz thanks for the review! I think I addressed all your comments.

ggetz:
Great, thanks @jjhembd! To merge in main, you'll want to follow @jjspace's [instructions](https://github.com/CesiumGS/cesium/pull/12206#issuecomment-2377930855) to make things less of a headache.

jjhembd:
@ggetz merge is done!

ggetz (approved):

ggetz:
Looks good! Thanks @jjhembd!

PR #12193, opened 2024-09-13
Updates to typedefs and documentation defaults 

lukemckinstry:
# Description

Type definition and documentation fix for the problem identified & described here https://github.com/CesiumGS/cesium/issues/11749#issuecomment-1950184016
which identifies an inconsistency in the documentation where we list default: undefined but our typedef requires a specific type. 

Made fixes using the following logic
 - if the argument is required in pragmas on the constructor
   - make sure it is not marked `optional` in the constructor args
   - make sure the `@default` is not listed as undefined (either change to default used in constructor or remove the line entirely)
 - if the argument is optional
   - change the typedef to allow undefined eg `... | undefined`

This PR covers fixes for `Source/Core`, other areas of the codebase will be covered in future PRs 

## Issue number and link

https://github.com/CesiumGS/cesium/issues/11749#issuecomment-1950184016

## Testing plan

 - build the documentation `npm run build-docs` and note changes to `default` values for targeted properties 
 - when using cesium with typescript, note you can set targeted properties to undefined

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 ##### Fixes :wrench:
 
 - Use first geometryBuffer if no best match found in I3SNode [#12132](https://github.com/CesiumGS/cesium/pull/12132)
+- Update type definitions to allow undefined for optional parameters [#12193](https://github.com/CesiumGS/cesium/pull/12193)
 
 ### 1.121.1 - 2024-09-04
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -167,6 +167,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
   - [jjspace](https://github.com/jjspace)
   - [Siddhesh Ranade](https://github.com/siddheshranade)
   - [Adam Morris](https://github.com/weegeekps)
+  - [Luke McKinstry](https://github.com/lukemckinstry)
 - [Northrop Grumman](http://www.northropgrumman.com)
   - [Joseph Stein](https://github.com/nahgrin)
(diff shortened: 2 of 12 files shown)


ggetz:
Thanks @lukemckinstry! I can confirm we now have you on file in the Cesium CLA.

ggetz:
> So for attributes with default values they cannot later be reassigned to undefined, only attributes without defaults can.
Does this make sense? Would it be useful to design the types so the developer could later reassign those attributes to undefined as well?

Let me confirm my understanding– You're suggesting that we should allow the properties with default value to be set to `undefined`, correct?

lukemckinstry:
> Let me confirm my understanding– You're suggesting that we should allow the properties with default value to be set to `undefined`, correct?

as discussed, this additional change is not necessary as the primary purpose of this PR is to update type definitions for optional parameters to include undefined.

lukemckinstry:
@ggetz updated CHANGES.md so this is ready for review (pending whether the CI occurs again)

ggetz:
Awesome, thanks @lukemckinstry!

javagl:
Just stumbled over this: I assume that the first checkbox from https://github.com/CesiumGS/cesium/issues/11749#issuecomment-1950184016 can now be ticked. (At least, I did that, updating that comment to point to this PR)

lukemckinstry:
> Just stumbled over this: I assume that the first checkbox from [#11749 (comment)](https://github.com/CesiumGS/cesium/issues/11749#issuecomment-1950184016) can now be ticked. (At least, I did that, updating that comment to point to this PR)

Hey @javagl, this didn't address all instances of the problem, only most of the ones that occurred in `Source/Core`, so I think the comment should be left unchecked until we address all cases.

Issue #11657, opened 2023-11-29
Shaking occurs when the camera is near an object.


syzdev:
When the camera approaches the lines, it is evident that the lines are shaking.

1. [CZML Path](https://sandcastle.cesium.com/?src=CZML%20Path.html): Zoom the camera to the nearest position.
![GIF2](https://github.com/CesiumGS/cesium/assets/25605081/c7f20b41-2fdc-4a21-8c83-d1dd09eb408a)

2. [Clamp to 3D Tiles](https://sandcastle.cesium.com/?src=Clamp%20to%203D%20Tiles.html): Move the camera in a very close position.
![GIF](https://github.com/CesiumGS/cesium/assets/25605081/1fb6d667-7f67-4259-b1ae-8f2260c02b57)

Sandcastle example: [CZML Path](https://sandcastle.cesium.com/?src=CZML%20Path.html)、[Clamp to 3D Tiles](https://sandcastle.cesium.com/?src=Clamp%20to%203D%20Tiles.html)

Browser: The latest version of Chrome

Operating System: Windows 11

syzdev:
CC [#4049](https://github.com/CesiumGS/cesium/issues/4049)

syzdev:
Hi @ggetz , I'm not sure if this is the same issue as [#9417](https://github.com/CesiumGS/cesium/issues/9417) .The difference is that I'm not using an Apple device. Has this been discussed in the past?

ggetz:
Hi @syzdev, thanks for the report! This is similar to #9417 as the shaking is likely coming from precision issues. But since you are replicating on Windows, the source of the issue is likely different.

I'm able to reproduce a slight jitter when zoomed in very close.

Would you be able to let us know your use case for zooming in very close?

syzdev:
@ggetz 

My use case is very similar to [Multi-part CZML](https://sandcastle.cesium.com/?src=Multi-part%20CZML.html), where I need to use `viewer.trackedEntity` to track objects along a path. In the official example, the model is often enlarged, but in reality, the model is very small, such as for a person or a car. Therefore, many times it is necessary to zoom the camera to the nearest position, which leads to shaking problems.

[...]

PR #12194, opened 2024-09-15
Added Entity.trackingReferenceFrame

### LINKED ISSUE #12293: Camera orientation problem with sampledPositionProperty and trackedEntity
### What happened?

When we have an Entity with a SampledPositionProperty set as trackedEntity, and change the position by a large distance (ie. > 0.01 degree lon or lat) the orientation of the camera changes during the repositioning.

Reported in the forums https://community.cesium.com/t/camera-position-problem-with-sampledpositionproperty-and-trackedentity/36360

From my own reproduction 
[Screencast from 11-07-2024 12:45:46 PM.webm](https://github.com/user-attachments/assets/efdfe365-4d85-4b04-af90-f2207b8244bb)

### Reproduction steps

1.
2.
3.
...

### Sandcastle example

https://sandcastle.cesium.com/index.html#c=pVdbU+M2FP4rmrysMzVKAuxuGy5TCuwsLcxSYPuUF2GfEBVZykhyQrqT/96jix3bCVt2CsOMfSR95zvfuchkShpLFhyWoMkJkbAk52B4WdC/vC2Z9DL/fq6kZVyCnvRS8m0iCeFyqn5TL2MyZcJA6kwGBGSWK3klc54xq3RrdcZytTRjYnVZWVQp8jPJC2ahtq/7RxMZfjNPD1l9EvxpZg1S9L4nva/3l3eT3ji8EiKYHZPDQzp8Pzz4OPyQRquSY7I32j+gw4+Hhx9Gv0Q7E7h7tD+Mr2aej0kpc5higHk0Wv28bZwrw11840qlc6YtPjF5QKdaFRfwpAFM0vLZ5OW89j3YGgP1cU5L6TUjMyZzAX/A6kItZQILkLZfxReE4DLTUKAddRjS4eiIDAbkLP+7xDVmUCfIIY+hgyXfnCypUyF1IZN1yHDUkjoJPQOvwZLbbEaCW/oMq9q1d88MkHfLd+ONKepIfjoho6EncuXYuY1o5rbMgTyucA1F5Na0Dz7ixuejLr7Zhb9X41/A/8Q/01otv867XlAkF0Utrnd2oxZApNJ29mZkl7Zd2Hs7sY0qfwD7GqZ2CxtrZjf2Eox9M/Sdq4Zd2Ls1wQz8Bza2DCuxw9q7NNhSS49z+cItmSpNlJ3h2MFii+lbV9UYyr3qNizb77ebr3Bf61ga/UilU+nV6HA/lNLOalotOZT6GWGrZwSunyte0eDa2DNfcokDjrI8v3RddM2NBZyYyTuMMHfFkbZ7PMw516ncfOLauLZ2M7A9FlzRW7hROYik1AIxwPGO/RlGN0V3SAkMxXRhls6ESPqxt39QyxDTjqG5NV0DD/8SNA+eDCvmAvLbjcPGrXLfXrzVag7arpIA0DlKsUaWTOeXL1azucLUoPFhNYdNEFtL9POX64tW5DNgOZdPmzM3zM6oVXdoZtIko4P3TfpzPwdxvjZsWgnRMc3mnfvyc3Bz647f4f4k+k0DYupBmo6U5pg11k7Kg0ZKGHZh6KwD+GeJZaAlbk/qAnQ02nn2hbBCwG5hYFUmsQEkK/C2dbXUvdU6CYjrDabj5ktcLlxpjjfdVWo+xhua0kHI9gWzbOA3mUEI80IrCc1n+iQe8cOiQig4fhKUxS1/AXHP/wF3Wf+8WWUvbvU+YwJX9of4U3ViuFmDylEArI/sGfLLSpcgkN/h/sJY2ljXrc+O+lSzCSc99w3kLvJ2o6LCYaRU8iWBQrpBik0b8Kf1N832pVztwZVGI8UTtEpYnHR8SpI4Qdq3duhHi42+qbDfS4FljzkBKtUy6TcGd1QsEyp7pv7YAy9cs/lnZ5fw+oGs1BplfOVIl9XUsY17t5mhlPeAG3OTvOYjJcO02X+b00m/ybHWvlbNo/vKTGoWaVPpNt/NbPafshF6TbCcYVvttgqvkT/qHkPvbxGjFf/ou/G/UYHo+JX417EhDNgridNnwUSyVebIxLUg2a52j9NLe8fGrgScVpR+5cUcv6zcBEpwSlhAJsjbDB5LbFWsGmOqDB4PmkePc74gPD/Z8U8JyQQzBlempfAzY9I7PR7g/q2jQvm5+mUBWrCV2zYbnV4HI34YHA/wdfdJq5R4ZLqD/C8

### Environment

Browser:
CesiumJS Version:
Operating System:

jfayot:
# Description

Added `Entity.trackingReferenceFrame` to allow tracking entities in their own reference frame.

## Issue number and link:
Fixes #12293

## Testing plan

Updated:
* EntitySpec.js
* EntityViewerSpec.js
to cover this new property

Updated Sandcastles CallbackPositionProperty and Interpolation for demo

Tested in the 3 different scene modes.
Tested non regression on fast satellites and slow moving objects.

# Author checklist


ggetz:
Thanks for the PR @jfayot! Would you mind expanding a bit on the motivation behind this change? Is it mostly to address https://github.com/CesiumGS/cesium/issues/11657?

jfayot:
Not really, in fact. I wanted to relate this PR to something similar, but the real motivation is to achieve the same effect as what you can see in Callback Position Property sandcastle in a one-liner and still be able to move the camera around the tracked entity.
May be the best would be to try the updated Interpolation sandcastle to give you an idea...

ggetz:
@jfayot Thank you, got it. I do think your Sandcastle example is useful behavior!

My concern here is I think there is some mismatch about how CesiumJS defines an "inertial reference frame". It appears that you are using it to refer to the _entity's_ inertial reference frame (as in [this example](https://en.wikipedia.org/wiki/Inertial_frame_of_reference#Simple_example)). In the rest of the CesiumJS API, I believe it is used to refer to _Earth's_ inertial reference frame, or [ECI](https://en.wikipedia.org/wiki/Earth-centered_inertial). What are your thoughts?

jfayot:
You're perfectly right: the `ReferenceFrame.INERTIAL` refers to the International Celestial Reference Frame, not to the entity's inertial reference frame.
Would it be ok if I rename the `ReferenceFrame` enum like this ?
* EARTH_FIXED
* CELESTIAL
* BODY (or ENTITY)

jfayot:
Or may be just create a new enum:
```javascript
/**
 * Constants for identifying well-known tracking reference frames.
 *
 * @enum {number}
 */
const TrackingReferenceFrame = {
  /**
   * The east-north-up entity's frame.
   *
   * @type {number}
   * @constant
   */
  EAST_NORTH_UP: 0,

  /**
   * The entity's inertial frame.
   *
   * @type {number}
   * @constant
   */
  INERTIAL: 1,
};
```

ggetz:
> Or may be just create a new enum:

I do think it's best to separate this behavior from the position's `ReferenceFrame`. :+1:

@jfayot My only remaining concern is that we do have some built-in behavior where `EntityView.js` "autodetects" satellites and and other orbital entities based on how fast they are moving and changes the camera view to Earth's inertial reference frame. I'm concerned that specifying `EAST_NORTH_UP` or `INERTIAL` will be confusing if that autodetect kicks in. What are your thoughts here?

We'll also need to test to ensure everything is working with the satellite case before this PR is merged.

jfayot:
@ggetz I have added a new sandcastle demo to illustrate the different tracking reference frames: Entity tracking.

jfayot:
Hi @ggetz ! Any update on this PR?

ggetz:
Thank you @jfayot! We'll get back to this in the next few days. Thanks for your patience, we appreciate your contributions!

packages/engine/Source/Core/TrackingReferenceFrame.js
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * The entity's inertial reference frame. If entity has no defined orientation
+   * property, a {@link VelocityOrientationProperty} is used instead, thus
+   * falling back to TrackingReferenceFrame.VELOCITY.
+   * When selected, the auto-detect algorithm is overridden.
+   *
+   * @type {number}
+   * @constant
+   */
+  INERTIAL: 1,
```
ggetz, line 28: Would it make sense to fall back to autodetect in the case where an entity has no defined orientation?

packages/engine/Source/Core/TrackingReferenceFrame.js
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * The entity's inertial reference frame with orientation fixed to its
+   * {@link VelocityOrientationProperty}, ignoring its own orientation.
+   * When selected, the auto-detect algorithm is overridden.
+   *
+   * @type {number}
+   * @constant
+   */
+  VELOCITY: 2,
```
ggetz, line 38: What happens when an entity's velocity is 0?

packages/engine/Specs/DataSources/EntityViewSpec.js
```diff
@@ -125,6 +156,13 @@ describe(
       const view = new EntityView(entity, scene);
       view.update(JulianDate.now());
     });
+
+    it("update returns with entity.trackingReferenceFrame set to INERTIAL and without entity.orientation property.", function () {
+      const entity = new Entity();
+      entity.trackingReferenceFrame = TrackingReferenceFrame.INERTIAL;
+      const view = new EntityView(entity, scene);
+      view.update(JulianDate.now());
```
ggetz, line 164: Should there be an `expect` assertion in this test?

packages/engine/Source/Core/TrackingReferenceFrame.js
```diff
@@ -0,0 +1,40 @@
...
+   * Auto-detect algorithm. The reference frame used to track the Entity will
+   * be automatically selected based on its trajectory: near-surface slow moving
+   * objects will be tracked in the entity's local east-north-up reference
+   * frame, while faster objects like satellites will use VVLH (Vehicle Velocity,
+   * Local Horizontal).
+   *
+   * @type {number}
+   * @constant
+   */
+  AUTODETECT: 0,
```
ggetz, line 17: Many of our enums use `NONE` for an undefined or default state. Does that make sense here>

ggetz (changes requested):
Awesome, I think this is a good approach @jfayot.

Apologies for the delay in getting back to this PR! I hope you still have interest in getting this change merged.

packages/engine/Source/Core/TrackingReferenceFrame.js
jfayot: Most of the enum with a NONE value I've seen are reflecting something meaningful:
- `ArcType.NONE`
- `ExtrapolationType.NONE`
- `Visibility.NONE`
- `HeightReference.NONE`
- ...
I'm not sure a `TrackingReferenceFrame.NONE` value would be of any interest. This would represent an invalid value that should be tested against, and add more noise to the code IMO.
Unless you would prefer me to replace the `TrackingReferenceFrame.AUTODETECT` by `TrackingReferenceFrame.NONE`?

packages/engine/Source/Core/TrackingReferenceFrame.js
jfayot: I agree on that point, as we already have `TrackingReferenceFrame.VELOCITY`. I'll change that!

packages/engine/Specs/DataSources/EntityViewSpec.js
jfayot: In fact this test is no more needed as the case descibed here has the AUTODETECT fallback...

jfayot:
Hi @ggetz! Thanks for your comments. This is ready for another review.

ggetz (approved):

ggetz:
Looking good @jfayot! I pushed some minor formatting tweaks to the documentation, but otherwise this is great!

Thanks!

PR #12202, opened 2024-09-18
Moved Viewer functionality to CesiumWidget class

### LINKED ISSUE #11967: Move `Viewer` functionality to `CesiumWidget` class
### Feature

There is functionality that's currently in the `@cesium/widgets` `Viewer` class that is not in the `@cesium/engine` `CesiumWidget` class that is not directly related to the widgets we provide in `@cesium/widgets` but _is_ valuable and useful to developers from an API perspective. Recently there was discussion of users wanting to use Cesium for the 3d view but roll their own UI and widgets. Installing only `@cesium/engine` and using the `CesiumWidget` is a great way to do that and reduce dependencies but there's more convenience we could offer like the `Viewer` does.

Things like:
- Easy access to an `entityCollection` that's already set up to render in the viewer
- a data source display already synced with the `clock` to update every tick (like needed in this [forum question](https://community.cesium.com/t/using-cesiumwidget-to-display-a-geojson/30349/2))
- convenience functions like `flyTo`, `zoomTo` and `trackedEntity`

These make it harder and more confusing to developers that `CesiumWidget` is not a drop in replacement for `Viewer` when they don't want the extra widgets like the timeline or geolocator. I think we should move all the code in `Viewer` that doesn't interact with `@cesium/widgets` classes/components into the `CesiumWidget` so `Viewer` is more of a direct `extends` type class. 

In other words I would expect:

```js
const viewer = new CesiumWidget('container', { ...options });
```
to function the same as 
```js
const viewer = new Viewer('container', {
  ...options,
  baseLayerPicker: false,
  selectionIndicator: false,
  timeline: false,
  geocoder: false,
  // etc...
});
```

jfayot:
# Description

This is a proposal to move Viewer functionalities not related to widgets to CesiumWidget class as proposed in #11967.

## Issue number and link

Fixes #11967 

## Testing plan

* Updated ViewerSpec.js
* Updated CesiumWidgetSpec.js
* Updated Cesium Widget sandcastle

I kept duplicate tests in ViewerSpec, may be some can be removed as they are implemented in CesiumWidgetSpec now?

# Author checklist


ggetz:
@jjspace Could you please take a pass on this PR and see if this is a direction we'd like to go API-wise? Please be mindful of any breaking changes, and deprecate or preserve functionality where we can.

jfayot:
@jjspace any update on this one? 👀

packages/engine/Source/Widget/CesiumWidget.js
```diff
@@ -387,6 +442,94 @@ function CesiumWidget(container, options) {
...
+        }
+      } else if (defined(that.trackedEntity)) {
+        that.trackedEntity = undefined;
+      }
+    }
+
+    this._screenSpaceEventHandler.setInputAction(
+      pickAndTrackObject,
+      ScreenSpaceEventType.LEFT_DOUBLE_CLICK
+    );
```
jjspace, line 532: I think this click handler should be moved back to the `Viewer`. Moving the _functionality_ to track an entity into the `CesiumWidget` is good so you can do `CesiumWidget.trackedEntity = ...` but setting up the click handler interaction is a `Viewer` oriented feature. Basically I would expect the handler in `Viewer` to be able to do `this._cesiumWidget.trackedEntity = pickEntity()` (with the extra checks it previously had).

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -954,10 +887,6 @@ Either specify options.terrainProvider instead or set options.baseLayerPicker to
     pickAndSelectObject,
     ScreenSpaceEventType.LEFT_CLICK
   );
-  cesiumWidget.screenSpaceEventHandler.setInputAction(
-    pickAndTrackObject,
-    ScreenSpaceEventType.LEFT_DOUBLE_CLICK
-  );
```
jjspace, line 960: Given my other comment this should be brought back.

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -1740,26 +1633,13 @@ Viewer.prototype.isDestroyed = function () {
...
-
   // Unsubscribe from data sources
   const dataSources = this.dataSources;
   const dataSourceLength = dataSources.length;
-  for (i = 0; i < dataSourceLength; i++) {
+  for (let i = 0; i < dataSourceLength; i++) {
     this._dataSourceRemoved(dataSources, dataSources.get(i));
   }
-  this._dataSourceRemoved(undefined, this._dataSourceDisplay.defaultDataSource);
+  this._dataSourceRemoved(undefined, this.dataSourceDisplay.defaultDataSource);
```
jjspace, line 1642: This logic now lives in `CesiumWidget.destroy()` too, is it still needed here?

packages/engine/Source/Widget/CesiumWidget.js
```diff
@@ -661,6 +841,104 @@ Object.defineProperties(CesiumWidget.prototype, {
...
+   * @readonly
+   */
+  trackedEntityChanged: {
+    get: function () {
+      return this._trackedEntityChanged;
+    },
+  },
+
+  /**
+   * Gets or sets the data source to track with the viewer's clock.
```
jjspace, line 927:
```suggestion
   * Gets or sets the data source to track with the widget's clock.
```

I'll comment on a couple other instances of this but please do a pass yourself to make sure code copied over refers to a widget not a viewer in this file

packages/engine/Source/Widget/CesiumWidget.js
```diff
@@ -133,10 +170,13 @@ function configureCameraFrustum(widget) {
...
+ * @param {boolean} [options.automaticallyTrackDataSourceClocks=true] If true, this widget will automatically track the clock settings of newly added DataSources, updating if the DataSource's clock changes.  Set this to false if you want to configure the clock independently.
  * @param {ContextOptions} [options.contextOptions] Context and WebGL creation properties passed to {@link Scene}.
  * @param {Element|string} [options.creditContainer] The DOM element or ID that will contain the {@link CreditDisplay}.  If not specified, the credits are added
  *        to the bottom of the widget itself.
  * @param {Element|string} [options.creditViewport] The DOM element or ID that will contain the credit pop up created by the {@link CreditDisplay}.  If not specified, it will appear over the widget itself.
+ * @param {DataSourceCollection} [options.dataSources=new DataSourceCollection()] The collection of data sources visualized by the widget.  If this parameter is provided,
+ *                               the instance is assumed to be owned by the caller and will not be destroyed when the viewer is destroyed.
```
jjspace, line 179:
```suggestion
 *                               the instance is assumed to be owned by the caller and will not be destroyed when the widget is destroyed.
```

packages/engine/Source/Widget/CesiumWidget.js
```diff
@@ -829,4 +1136,512 @@ CesiumWidget.prototype.render = function () {
...
+
+  that.scene.requestRender();
+  return zoomPromise;
+}
+
+function clearZoom(viewer) {
+  viewer._zoomPromise = undefined;
+  viewer._zoomTarget = undefined;
+  viewer._zoomOptions = undefined;
+}
```
jjspace, line 1438: For this and a handful of functions below that were moved over the argument names should be swapped out so they don't refer to a `viewer` anymore
```suggestion
function clearZoom(widget) {
  widget._zoomPromise = undefined;
  widget._zoomTarget = undefined;
  widget._zoomOptions = undefined;
}
```

jjspace (commented):
Thanks @jfayot, this is looking good. sorry for the delay, github doesn't make it very easy to review diffs for things that move between files... I had a few initial comments

> I kept duplicate tests in ViewerSpec, may be some can be removed as they are implemented in CesiumWidgetSpec now?

Generally we don't write specs that test "dependencies" of an object. I would say any tests that are for viewer functions that are now just passed through to widget functions should be removed.

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -1740,26 +1633,13 @@ Viewer.prototype.isDestroyed = function () {
...
  */
 Viewer.prototype.destroy = function () {
-  let i;
-  if (
-    defined(this.screenSpaceEventHandler) &&
-    !this.screenSpaceEventHandler.isDestroyed()
-  ) {
-    this.screenSpaceEventHandler.removeInputAction(
-      ScreenSpaceEventType.LEFT_CLICK
-    );
```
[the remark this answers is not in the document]
jfayot: Was initially moved in the widget's destroy (even though the event subscription was still done in the viewer), but as mouse event handlers are being brought back to the viewer, then yes, the unsubscription should be brought back too...

packages/engine/Source/Widget/CesiumWidget.js
jfayot: I agree that it was quite strange to have the click handled in the viewer and the double click handled in the widget, but the intent was to have the tracking feature activated with double click available within the widget without further user coding.
However this is still possible through the API without too much effort, so yes, let's bring it back!

packages/widgets/Source/Viewer/Viewer.js
jfayot: Sure

packages/engine/Source/Widget/CesiumWidget.js
jfayot: Copy paste hell!
FWIW, there were some viewer's references prior to my PR. I fixed them also...

packages/engine/Source/Widget/CesiumWidget.js
jfayot: Fixed

packages/engine/Source/Widget/CesiumWidget.js
jfayot: Fixed

packages/widgets/Source/Viewer/Viewer.js
jfayot: Forgot this one

packages/engine/Source/Widget/CesiumWidget.js
jfayot: @jjspace Thanks for the review. I agree, this is not very easy to track the changes when things move around between files, whatever the tool...
Following your suggestion, I've removed the viewer's tests that are now done in the widget's spec. Unsurprisingly, the coverage decreases for the former, while it increases for the latter...
Tell me if you see anything else!

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -1887,24 +1754,11 @@ Viewer.prototype._dataSourceRemoved = function (
 Viewer.prototype._onTick = function (clock) {
   const time = clock.currentTime;
 
-  const isUpdated = this._dataSourceDisplay.update(time);
-  if (this._allowDataSourcesToSuspendAnimation) {
+  const isUpdated = this._cesiumWidget.dataSourceDisplay.ready;
```
jfayot, line 1757: I don't know if you noticed this change, but I'm not quite confident with it, and I have no idea how to do it differently. What do you think?

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -493,13 +473,16 @@ Either specify options.terrainProvider instead or set options.baseLayerPicker to
         ? false
         : undefined,
     clock: clock,
+    shouldAnimate: options.shouldAnimate,
```
jjspace, line 476: Can you please be sure to add these new `options` to the JSDoc for this class? The descriptions can match the ones from the `CesiumWidget` they're being passed to.

packages/engine/Source/Widget/CesiumWidget.js
```diff
@@ -387,6 +430,72 @@ function CesiumWidget(container, options) {
...
+      scene: scene,
+      dataSourceCollection: dataSourceCollection,
+    });
+
+    const eventHelper = new EventHelper();
+    this._dataSourceChangedListeners = {};
+    this._automaticallyTrackDataSourceClocks = defaultValue(
+      options.automaticallyTrackDataSourceClocks,
+      true
+    );
```
jjspace, line 451: We are phasing out `defaultValue`, see https://github.com/CesiumGS/cesium/pull/12207. I know this was just copied over but to avoid future merge conflicts can you replace usages like this in "new" code with `??`?
```suggestion
    this._automaticallyTrackDataSourceClocks = 
      options.automaticallyTrackDataSourceClocks ?? true;
```

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -872,25 +836,12 @@ Either specify options.terrainProvider instead or set options.baseLayerPicker to
...
-  this._zoomPromise = undefined;
-  this._zoomOptions = undefined;
   this._selectedEntityChanged = new Event();
-  this._trackedEntityChanged = new Event();
 
-  knockout.track(this, [
-    "_trackedEntity",
-    "_selectedEntity",
-    "_clockTrackedDataSource",
-  ]);
```
jjspace, line 893: Just commenting on this. I'm not completely certain why this was here previously but I don't see anything to replace it and I just want to be sure it's not setting something up that's used elsewhere that we're missing. Maybe something with the selection details widget?

packages/widgets/Source/Viewer/Viewer.js
jjspace: I do remember thinking something about this seemed wrong but I forgot to loop back to it. I think it might be the case that this can be skipped because the `onTick` for the `CesiumWidget` checks the same thing and stops the same clock object if the data source display doesn't update? I also don't 100% understand what these functions are doing though so I might be wrong. Maybe we should even be initializing the `clock` differently in the `Viewer` now?

@ggetz would you be able to take a look specifically at this chunk of code/functionality?

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -138,26 +125,23 @@ function pickEntity(viewer, e) {
 
 const scratchStopTime = new JulianDate();
 
-function trackDataSourceClock(timeline, clock, dataSource) {
+function trackDataSourceClock(timeline, dataSource) {
```
jjspace, line 128: Optional change: It might be nice to rename this function now to something like `linkTimelineToDataSourceClock`? it's a little more verbose but describes better the intent?

packages/engine/Source/Widget/CesiumWidget.js
jjspace: > However this is still possible through the API without too much effort, so yes, let's bring it back!

Exactly, I think it's good to expand the widget API with this functionality, making it easier for users to add _if they want_ but still only do it for them in the viewer

jjspace (commented):
Thanks for the changes @jfayot, I think this is getting pretty close. Just had a few more small comments.

I'm going to try and do a bit more testing tomorrow. For my sanity given that the `Viewer` and `CesiumWidget` are basically the "core classes" of CesiumJS and used by every project I'd feel more comfortable with a second review pass, maybe @ggetz or @lukemckinstry can take a look through?

jfayot:
Hi @jjspace ! I'll be off for one week starting from now with no connectivity at all...
I'll catch up on oct 7th
See you soon

lukemckinstry:
Did some high level testing, on sandcastles like this https://sandcastle.cesium.com/?src=GeoJSON%20simplestyle.html when using this branch should we be able to replace `Cesium.Viewer` with `Cesium.CesiumWidget`?

jjspace:
> when using this branch should we be able to replace `Cesium.Viewer` with `Cesium.CesiumWidget`

The basic answer is yes. After this PR the API for `CesiumWidget` and `Viewer` should be the same so they can be exchanged except for where code interacts with the actual widgets that are added by the `Viewer` like the timeline and animation controls.

Swapping it in that one you linked works fine for me, were you seeing an issue? I had tried on a handful of sandcastles already but if you see specific issues please share them so we can take a look.

lukemckinstry:
> Swapping it in that one you linked works fine for me, were you seeing an issue? I had tried on a handful of sandcastles already but if you see specific issues please share them so we can take a look.

I did see an issue in that sandcastle, but it is working fine now :+1:, so perhaps I set it up wrong

packages/widgets/Source/Viewer/Viewer.js
jfayot: `shouldAnimate` was already there in the `Viewer`'s ctor options before this PR. I added it to `CesiumWidget` ctor options to be passed to the `_clock`.
Did you mean something else?

packages/widgets/Source/Viewer/Viewer.js
jfayot: ok

packages/widgets/Source/Viewer/Viewer.js
jfayot: From what I understand, this block's intent is to temporarily stop the animation if something is still buffering into the dataSource. Since `ClockViewModel` only pass through the `Clock`'s `canAnimate` property, and the clock is already updated in `CesiumWidget`, I first thought that this was no more needed into the Viewer.
To be sure, I did try to duplicate the test "suspends animation by dataSources if allowed" from `CesiumWidget` spec to `Viewer` spec (replacing widget.clock by viewer.clockViewModel) and removed the mentioned code block: the test failed!
I guess this is because the `ClockViewModel` properties are not synchronized with `Clock`'s properties unless `ClockViewModel.synchronize()` is called. Therefore, in the case where `dataSource.update()` would return false, you would end up with `clock.canAnimate` set to false, and `clockViewModel.canAnimate` still equal to true.
May be the clockViewModel is being synchronized later but I can't be certain...

packages/engine/Source/Widget/CesiumWidget.js
jfayot: Sure!

jfayot:
Hi @jjspace, I'm back :grin:
I've addressed most of your comments I think, but there are still some pending points...
Tell me what should be the next step!

packages/widgets/Source/Viewer/Viewer.js
jjspace: @ggetz do you know if this was used in any sandcastles or older widgets at some point?

packages/widgets/Source/Viewer/Viewer.js
jjspace: If they're all already there that's fine then. I was just looking at the git diff and didn't see them so probably missed it.

packages/widgets/Source/Viewer/Viewer.js
ggetz: Not to my knowledge. Looks like this are just relics. The subscriptions themselves appear to have been removed in https://github.com/CesiumGS/cesium/commit/dfb452824b2e1b7db0e93078575aefd428c9b630#diff-f348e35101f15dcc67c86e7ec8060650fd7d17d0f3f2eeed58ffc7455520a5bbL738.

packages/widgets/Source/Viewer/Viewer.js
ggetz: Yes, I think this is a sensible change. Just as @jfayot stated, we need to update the `clockViewModel`, which `CesiumWidget` does not have a reference to.

ggetz:
@jjspace I follow up on the questions you had! 

@jfayot and @jjspace A quick reminder to make sure `CHANGES.md` is updated to reflect any new properties or functionality available on the `CesiumWidget` class. Thanks!

jjspace:
Awesome, thanks @ggetz!

@jfayot Can you please merge in main to make sure there are no more conflicts? Then update the `CHANGES.md` as Gabby mentioned. I think most of the changes can probably go under a main bullet point of something like "move Viewer functionality to CesiumWidget to increase usability".

Overall with our previous reviews I think this is really close. I'll take another look once the branch is updated!

jjspace (approved):
Sorry for the delay on this @jfayot. All looks good to me now, thank you for the changes!

Issue #12187, opened 2024-09-10
Scrolling to zoom in/out is way too fast on high refresh rate screens


tigerchase96:
### What happened?

If I use my scroll wheel to zoom in/out in Cesium, it scrolls extremely fast. It's much too sensitive to be usable. This behavior goes away if I intentionally lower my monitor's refresh rate to 60 Hz. Then scrolling behavior returns to normal. 

My monitor's native refresh rate is 240 Hz. I've also tested setting my monitor to 120 Hz, and the problem still persists, but to a lesser extent. 

I imagine there must be some logic in the source code that is dependent on the refresh rate of the webpage? I've only just started following the quickstart guide so I probably can't contribute to fixing the bug. 

But I did discover a workaround in the meantime: If I decrease the `ScreenSpaceCameraController`'s zoom factor, zoom behavior is seemingly identical to the 60 Hz example. Unfortunately, this fix means I have to know the refresh rate of the user's screen.

```ts
const refreshRate = 240;
viewer.scene.screenSpaceCameraController.zoomFactor *= 60 / refreshRate;
```

### Running at 60 Hz: (expected behavior)

https://github.com/user-attachments/assets/13badc67-557e-4c83-96be-077031d57398

### Running at 240 Hz

In this example I am only moving my scroll wheel a few clicks in each direction, and yet I reach ground level and infinite distance way too quickly.

https://github.com/user-attachments/assets/1bbb4f5a-3794-49b0-9122-b6e7188e07cd

### Reproduction steps

1. Use a monitor that supports refresh rates above/below 60 Hz
2. Set the monitor to 60 Hz
3. Run any instance of Cesium, e.g. https://cesium.com/platform/cesiumjs/
4. Scroll in and out and observe how many clicks it takes to reach ground level
5. Set the monitor to another value like 120 Hz or 24 Hz
6. Go back to Cesium and scroll in and out again
7. Observe that the scroll sensitivity is significantly different than before

### Environment

Browser: I've tested Chrome 128 and Firefox 130 and it occurs on both
CesiumJS Version: the latest version on the website, and v1.121.1 locally
Operating System: Windows 11

s3xysteak:
[...]

PR #12211, opened 2024-09-23
Base zoom speed on browser refresh rate

lukemckinstry:
# Description

Bases zoom speed on the render rate of the user's browser.
Added fields in FrameRateMonitor to track the timestamps of the n most recent renders.
Then when calculating the zoom distance on each frame, adjust it based on the fps zoom speed to provide consistent experience for browsers with different speeds.
If the user wants faster or slower zoom, they can still adjust zoomFactor through the public api as before with the same effects.

## Issue number and link
https://github.com/CesiumGS/cesium/issues/12187

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Scene.js
+++ b/packages/engine/Source/Scene/Scene.js
@@ -79,8 +79,6 @@ import VoxelCell from "./VoxelCell.js";
 import VoxelPrimitive from "./VoxelPrimitive.js";
 import getMetadataClassProperty from "./getMetadataClassProperty.js";
 import PickedMetadataInfo from "./PickedMetadataInfo.js";
-import getTimestamp from "../Core/getTimestamp.js";
-import FpsTracker from "../Core/FpsTracker.js";
 
 const requestRenderAfterFrame = function (scene) {
   return function () {
--- a/packages/engine/Source/Scene/Scene.js
+++ b/packages/engine/Source/Scene/Scene.js
@@ -79,6 +79,8 @@ import VoxelCell from "./VoxelCell.js";
 import VoxelPrimitive from "./VoxelPrimitive.js";
 import getMetadataClassProperty from "./getMetadataClassProperty.js";
 import PickedMetadataInfo from "./PickedMetadataInfo.js";
+import getTimestamp from "../Core/getTimestamp.js";
+import FpsTracker from "../Core/FpsTracker.js";
 
 const requestRenderAfterFrame = function (scene) {
   return function () {
--- a/packages/engine/Source/Scene/ScreenSpaceCameraController.js
+++ b/packages/engine/Source/Scene/ScreenSpaceCameraController.js
@@ -578,7 +578,7 @@ function handleZoom(
   const minDistance = distanceMeasure - minHeight;
 
   let fpsMultiplier = 1;
-  const fps = object.frameRateMonitor.lastFramesPerSecond;
+  const fps = object.frameRateMonitor.averageFramesPerSecond;
   if (fps) {
     fpsMultiplier = 30 / fps;
   }
--- a/packages/engine/Source/Scene/ScreenSpaceCameraController.js
+++ b/packages/engine/Source/Scene/ScreenSpaceCameraController.js
@@ -579,6 +579,7 @@ function handleZoom(
 
   let fpsMultiplier = 1;
   const fps = object.frameRateMonitor.averageFramesPerSecond;
+  // we set the ideal browser refresh rate to 30hz
   if (fps) {
     fpsMultiplier = 30 / fps;
   }
--- a/packages/engine/Source/Scene/ScreenSpaceCameraController.js
+++ b/packages/engine/Source/Scene/ScreenSpaceCameraController.js
@@ -577,11 +577,11 @@ function handleZoom(
 
   const minDistance = distanceMeasure - minHeight;
 
-  let fpsMultiplier = 1;
+  let fpsMultiplier = 1.0;
   const fps = object.frameRateMonitor.averageFramesPerSecond;
-  // we set the ideal browser refresh rate to 30hz
-  if (fps) {
-    fpsMultiplier = 30 / fps;
+  // we set the ideal browser refresh rate to 60hz
+  if (defined(fps) && fps > 0) {
+    fpsMultiplier = 60.0 / fps;
   }
 
   let zoomRate = zoomFactor * minDistance * fpsMultiplier;
--- a/packages/engine/Source/Scene/ScreenSpaceCameraController.js
+++ b/packages/engine/Source/Scene/ScreenSpaceCameraController.js
@@ -25,6 +25,7 @@ import MapMode2D from "./MapMode2D.js";
 import SceneMode from "./SceneMode.js";
 import SceneTransforms from "./SceneTransforms.js";
 import TweenCollection from "./TweenCollection.js";
+import FrameRateMonitor from "./FrameRateMonitor.js";
 
 /**
  * Modifies the camera position and orientation based on mouse input to a canvas.
@@ -144,6 +145,8 @@ function ScreenSpaceCameraController(scene) {
    * @default 5.0
    */
   this.zoomFactor = 5.0;
+
+  this.frameRateMonitor = FrameRateMonitor.fromScene(scene);
 
   /**
    * The input that allows the user to pan around the map. This only applies in 2D and Columbus view modes.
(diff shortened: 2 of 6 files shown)


lukemckinstry:
posted this for initial review/feedback/discussion @ggetz. I still need to add unit tests.

javagl:
This looks somewhat similar to what I dumped into [a forum thread](https://community.cesium.com/t/fps-drop-while-rendering-sparse-point-cloud-data-as-3d-tiles/28657/8) a while ago. The goal _there_ was also to have an "average frame rate over the last `n` frames". 

In that forum thread, I put this into some `FpsTracker` class. The reasons may be obvious. The functionality is _encapsulated_ in a class where the name says what the class does (`/* An FpsTracker tracks the FPS */` 😁  ). It offers a `getAverageFps` function _directly_. It avoids that `const timeStamps = object._scene.frameState?.timeStamps;` (aka [Train Wreck](https://wiki.c2.com/?TrainWreck)). And it avoids the manual computation (with magic constants like `10` and `30`) inside the `handleZoom` function.

The purpose of that `FpsTracker` there was to plug it into some `MsseUpdater`, which was supposed the update the maximum screen space error based on the average FPS. (Fleshing this out into a _real feature_ is still on my TODO list - I think that this could make a whole lot of sense...). 

**tl;dr** : If there was "something" where a function like `getAverageFps` was offered, then this could probably be re-used in multiple places...

lukemckinstry:
I like your idea to create a `FpsTracker` class to make this more reusable. I will work on this change and update this PR when it is ready.

ggetz:
I like the idea of isolating the average FPS logic in one place. The update itself should probably still happen in `Scene.js` though.

packages/engine/Source/Scene/Scene.js
```diff
@@ -162,6 +163,7 @@ function Scene(options) {
   this._frameState.scene3DOnly = defaultValue(options.scene3DOnly, false);
   this._removeCreditContainer = !hasCreditContainer;
   this._creditContainer = creditContainer;
+  this._fpsTracker = new FpsTracker();
```
ggetz, line 166: Since we are accessing this from another class, can we expose this as `fpsTracker` with a getter? It can be marked as `@private` so it doesn't show up in the developer API documentation.

lukemckinstry: updated with this change :+1:

packages/engine/Source/Core/FpsTracker.js
```diff
@@ -0,0 +1,40 @@
...
+ * @private
+ */
+function FpsTracker(options) {
+  this._timeStamps = [];
+}
+
+/**
+ * Calculates average fps based on timeStamps of recently rendered frames
+ */
+FpsTracker.prototype.getAverageFps = function () {
```
javagl, line 14: The hard-coded `10` (and the corresponding `9`s below) make my eye twitch 🙁 

But beyond that: This does not really compute the "average FPS". 

Consider time stamps like
`1 2 3 4 5 6 7 8 9 10`
It will compute `(10 - 1) / 9 = 1.0`. 
This means that the ~"time per frame" is `1.0`. (This isn't really the "elapsed time", but ... well...)

For
`1 2 3 4 5 6 7 8 9 19`
it will compute `(19 - 1) / 9 = 2.0`.
So even though the "time per frame" will constantly be `1.0` (except for the outlier of the last frame), the "time per frame" will be computed as `2.0`. The whole idea behind storing multiple time stamps is _exactly_ to "smooth (average) out" single outliers like that last `19`. 

One could mention that, in the current approach, it wouldn't even be necessary to store the 10 time stamps. Storing **2** (and a 'counter') would be enough. For a proper "sliding window", one could examine the performance of `unsihift`/`pop` vs. `push`/`shift`. Or one avoid these considerations altogether and use a fixed-sized array and a 'cursor' (aka "circular buffer").

packages/engine/Source/Core/FpsTracker.js
```diff
@@ -0,0 +1,40 @@
...
+FpsTracker.prototype.updateFrameTimeStamp = function () {
+  const time = getTimestamp();
+  this._timeStamps.unshift(time);
+  if (this._timeStamps.length > 10) {
+    this._timeStamps.pop();
+  }
+};
+
+/**
+ * Destroys the WebGL resources held by this object.
```
javagl, line 35: This does not have WebGL resources

lukemckinstry: thank you, since we will now use the FrameRateMonitor class this is no longer relevant

packages/engine/Source/Core/FpsTracker.js
javagl: This should be obsolete with the `FrameRateMonitor`. 

(While some aspects of this class are hard to understand at a glimpse (like that "visibility" thing....), it seems to offer the required functionality, so _not_ duplicating that (even in a simpler, more self-contained form) seems reasonable)

lukemckinstry:
We realized the existing FrameRateMonitor class provides close to the functionality this feature needs. We decided to make some small additions to it rather than create a new class (side notes: All the work up to this commit 331650c67d80b9011ec15a077fa7c8ea6e8b6f6d got re-written, so it may make sense to squash the early commits )

There may be questions as to why we dont get FPS from `FrameRateMonitor.lastFramesPerSecond`, this should be made clear by the doc updates included in this PR.

One open comment that I have not addressed: Including the target/idea browser refresh rate as constant value (30 as in 30hz) in the code. I am unsure if we should A) expose setting this in the API (but we already have zoomFactor in the public API) or B) set it in a more clear place, like a constants or system settings file.

packages/engine/Source/Scene/ScreenSpaceCameraController.js
```diff
@@ -573,7 +576,16 @@ function handleZoom(
   const maxHeight = object.maximumZoomDistance;
 
   const minDistance = distanceMeasure - minHeight;
-  let zoomRate = zoomFactor * minDistance;
+
+  let fpsMultiplier = 1;
+  const fps = object.frameRateMonitor.averageFramesPerSecond;
+  // we set the ideal browser refresh rate to 30hz
+  if (fps) {
+    fpsMultiplier = 30 / fps;
```
lukemckinstry, line 584: This is one open comment that I have not addressed: Including the target/ideal browser refresh rate as constant value (30 as in 30hz) in the code here. I am unsure if we should A) expose setting this in the API (but we already have zoomFactor in the public API) or B) set it in a more clear place, like a constants or system settings file.
I prefer option B

ggetz (commented):
Looking good @lukemckinstry! I have a few comments on code style, and a few thoughts on behavior.

ggetz:
Please excuse the comments on outdated code, I'm not sure how they snuck in there.

packages/engine/Source/Scene/ScreenSpaceCameraController.js
lukemckinstry: I added a check to make sure `fps > 0`, which should ensure we avoid division by 0

packages/engine/Source/Scene/ScreenSpaceCameraController.js
```diff
@@ -573,7 +576,16 @@ function handleZoom(
...
+  }
+
+  let zoomRate = zoomFactor * minDistance * fpsMultiplier;
```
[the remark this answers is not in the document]
ggetz: I like this solution. I wonder if we can set the min to something like `0.5` to still account for time when the FPS does dip a bit.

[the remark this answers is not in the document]
javagl: I might be lacking some context, but would have thought that taking the _average_ FPS was supposed to smooth out any hiccups in the FPS.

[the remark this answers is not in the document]
lukemckinstry: Happy to fill in context @javagl 
The situation we are seeing when we test the branch on scenes with heavy data loading operations (like this sandcastle using the NY Buildings data https://sandcastle.cesium.com/?src=3D%20Tiles%20Feature%20Picking.html) is these heavy operations can make the render rate can slow down for a while (several seconds) or even indefinitely if the user continues operations that forces data reloading.

Low FPS leads to a high fpsMulitplier. 

In the equation `let zoomRate = zoomFactor * minDistance * fpsMultiplier` minDistance is supposed to be small as we get close to the ellipsoid surface and large when far out in space to create intuitive UX. But high fpsMultiplier throws this off, yielding a high zoomRate when we are already zoomed in close, which explains the jerky/jumpy behavior we were referring to.

Back to the original complaint/issue. The problem is actually not when FPS is low, it is only when FPS is too fast. So the idea with `fpsMultiplier = Math.min(fpsMultiplier, 1.0)` is to only adjust the zoomRate downward towards what we expect from a 60 FPS screen when the screen has a much higher refresh rate (eg. 120hz).

ggetz:
@lukemckinstry I'm going to close this to keep things tidy. If you or anyone else wants to update and finish this out, please feel free to re-open!

Issue #12219, opened 2024-09-27
Texture WebGL: OUT_OF_MEMORY


cnchore:
### What happened?

When using the videoElement material, the console reported an error and couldn't draw properly。

error info:
WebGL: OUT_OF_MEMORY: texSubImage2D: bad image data

### Reproduction steps

1. Visit the website: https://sandcastle.cesium.com/?src=Video.html
2. Opening the development console
3.
...

### Sandcastle example

https://sandcastle.cesium.com/?src=Video.html

### Environment

Browser: Chrome 129.0.6668.70
CesiumJS Version:1.121
Operating System: macos

cnchore:
![image](https://github.com/user-attachments/assets/ab32467c-4191-4e99-9660-0779229a13e3)

ggetz:
@jjspace Could you please check if this is a regression from https://github.com/CesiumGS/cesium/pull/12158? If not, would you mind bisecting to see where this broke?

jjspace:
git bisect points to commit https://github.com/CesiumGS/cesium/commit/e00c68ad9d24e0373a4bbf928a5f5a441e22d7bd which was part of https://github.com/CesiumGS/cesium/pull/12100 from @jjhembd 

```sh
# "bad" commit
git checkout e00c68ad9d24e0373a4bbf928a5f5a441e22d7bd
# "good commit right before that one
git checkout e00c68ad9d24e0373a4bbf928a5f5a441e22d7bd~1
```

PR #12224, opened 2024-09-30
Fix issue with copying to a texture from a HTMLVideoElement

ggetz:
# Description

A `HTMLVideoElement` has additional `videoHeight` and `videoWidth` properties. These are the [intrinsic dimensions](
https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/videoHeight#about_intrinsic_width_and_height). If we rely on the `height` and `width` values, we can get sub optimal values, and – as in the case of the reported bug – sometimes `0` values, which can break the WebGL operation entirely.

While I was in that code, I also opted to use [`naturalHeight`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/naturalHeight) and `naturalWidth` for `HTMLImageElement`, as those the intrinsic, not scaled, dimensions.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12219

## Testing plan

1. Visit the `Video.html` Sandcastle example
2. Ensure the video is running on the sphere
3. Open the development console and ensure no WebGL errors are reported

# Author checklist

### DIFF
--- a/packages/engine/Source/Renderer/Texture.js
+++ b/packages/engine/Source/Renderer/Texture.js
@@ -815,14 +815,12 @@ Texture.prototype.copyFrom = function (options) {
   const arrayBufferView = source.arrayBufferView;
 
   // Make sure we are using the element's intrinsic width and height where available
-  if (defined(source.naturalWidth) && defined(source.naturalHeight)) {
-    width = source.naturalWidth;
-    height = source.naturalHeight;
-  }
-
   if (defined(source.videoWidth) && defined(source.videoHeight)) {
     width = source.videoWidth;
     height = source.videoHeight;
+  } else if (defined(source.naturalWidth) && defined(source.naturalHeight)) {
+    width = source.naturalWidth;
+    height = source.naturalHeight;
   }
 
   const textureWidth = this._width;
--- a/packages/engine/Source/Renderer/Texture.js
+++ b/packages/engine/Source/Renderer/Texture.js
@@ -60,11 +60,12 @@ function Texture(options) {
 
   let { width, height } = options;
   if (defined(source)) {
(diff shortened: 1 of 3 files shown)


ggetz:
@jjspace Could you please review this fix for tomorrow's release?

packages/engine/Source/Renderer/Texture.js
```diff
@@ -810,7 +811,19 @@ Texture.prototype.copyFrom = function (options) {
...
+  // Make sure we are using the element's intrinsic width and height where available
+  if (defined(source.naturalWidth) && defined(source.naturalHeight)) {
+    width = source.naturalWidth;
+    height = source.naturalHeight;
+  }
+
+  if (defined(source.videoWidth) && defined(source.videoHeight)) {
+    width = source.videoWidth;
+    height = source.videoHeight;
+  }
```
jjspace, line 826: Optional but I think this could read a bit better if you switch the order and use an `else if`. Then it more closely resembles the order of the other change? Then it's also clear that we're only doing one assignment to overwrite the initial value. 
```suggestion
  // Make sure we are using the element's intrinsic width and height where available
  if (defined(source.videoWidth) && defined(source.videoHeight)) {
    width = source.videoWidth;
    height = source.videoHeight;
  } else if (defined(source.naturalWidth) && defined(source.naturalHeight)) {
    width = source.naturalWidth;
    height = source.naturalHeight;
  }
```

jjspace (commented):
Can confirm the sandcastle is now working again. Happy to see it was a small fix. Just had one small comment

ggetz:
Thanks @jjspace! Updated!

jjspace (approved):
LGTM!

Issue #11326, opened 2023-06-01
defaultValue.EMPTY_OBJECT missing from documentation and Typescript


angrycat9000:
`defaultValue.EMPTY_OBJECT` has JSDoc and is not explicitly marked as private.  However it does not show up in either the [generated documentation](https://cesium.com/learn/cesiumjs/ref-doc/global.html?classFilter=defaultValue#defaultValue) or the type definitions.

https://github.com/CesiumGS/cesium/blob/b2d169d75e8c9e5cf6cda018f802cc18a860ebc1/packages/engine/Source/Core/defaultValue.js#L21-L27

**Screenshot of doc for defaultValue**
![image](https://github.com/CesiumGS/cesium/assets/48362463/1e665a02-ff61-4b23-84e9-573c566c29f0)

**Generated type definition for defaultValue**
```js
export function defaultValue(a: any, b: any): any;
```

anpaulan:
Hi @ggetz could I take on this issue?

ggetz:
@anpaulan Sure! I think the strategy here is to [deprecate](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#deprecation-and-breaking-changes) `defaultValue.EMPTY_OBJECT` and create a new standalone constant.

ggetz:
Thanks for the input @angrycat9000! While call signatures would work for TypeScript definitions, [there's no defined way to do so with JSDoc](https://github.com/jsdoc/jsdoc/issues/1753).

> In that case we could just edit the generated typescript definition to use. That is what we ended up doing for https://github.com/CesiumGS/cesium/pull/11455.

IIUC that would fix the type definition, but is not a solution for the documentation. Am I missing something?

angrycat9000:
> IIUC that would fix the type definition, but is not a solution for the documentation. Am I missing something?

Ah I forgot that this was not in the documenation at all (as opposed to just not typed correctly).     

I think the current API design is good for usage, because it associates the `EMPTY_OBJECT` closely with the intended usage as a default value. But not sure how to fix the documentation part besides creating an contant that isn't attached to a function.

jjspace:
This was addressed by https://github.com/CesiumGS/cesium/pull/12507

[...]

PR #12207, opened 2024-09-20
Replace `defaultValue` with `??`

dave-b-b:
# Description

This is a partial pull request to verify that I am making the right kinds of changes.

## Issue number and link

This is a partial fix. I don't want to link here and have this automatically close the ticket. But the ticket is #11674.

## Testing plan

I ran `npm run test`. It failed some of the tests, but none of the tests related to the changes that I made. (It seemed like they were all timeout issues)

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [ ] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [x] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [x] I have performed a self-review of my code

Issue #11674, opened 2023-12-05
Replace `defaultValue` with `??`


jjspace:
The builtin JS [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) `??` behaves exactly the same as the `defaultValue` function

Transition options

1. All at once - replace all instances of it's use in one big PR
    - There _is_ a way to restrict usage of the function in eslint to help everyone transition
2. Stop using the function and prefer using `??` for all _new_ code 
3. Continue using `defaultValue` to match existing code
    - If this, do we specifically recommend against/restrict the use of `??`?

Considerations

- This is part of the public API ([`defaultValue`](https://cesium.com/learn/cesiumjs/ref-doc/global.html?classFilter=default#defaultValue)), how many people use it downstream?
- What should we do with `defaultValue.EMPTY_OBJECT`? (Related https://github.com/CesiumGS/cesium/issues/11326)
    - Is there a performance implication with replacing it with `{}` everywhere?
    - If we keep it as a constant where should it be defined?

ggetz:
Overall, the more we can automate things with eslint rules, the better. If we do chose to replace it all at once, we should have any eslint configurations in one commit, and the actual replacement in its own commit to make merging as easy as possible.

> This is part of the public API ([defaultValue](https://cesium.com/learn/cesiumjs/ref-doc/global.html?classFilter=default#defaultValue)), how many people use it downstream?

I agree we'd want to take stock of the usage before removing it. If we do chose to remove it, we'd go through a [deprecation period](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#deprecation-and-breaking-changes) where we transition users off.

> Is there a performance implication with replacing it with {} everywhere

Yes. This creates a new object every time, whereas `defaultValue.EMPTY_OBJECT` uses one, frozen object. There is a similar explanation in [the `defaultValue` section of the Coding Guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#default-parameter-values).

dave-b-b:
@ggetz Not sure whether this is still needed, but I can do this. I'm going to drop the other one I volunteered for.

dave-b-b:
<img width="785" alt="image" src="https://github.com/user-attachments/assets/4bc86fc7-70ed-4abe-8172-84c3550eaf79">

[...]

PR #12238, opened 2024-10-07
Deprecated defaultValue.EMPTY_OBJECT and created EmptyObject standalone constant

anpaulan:
# Description

- Fixed missing JSDOC for `EMPTY_OBJECT`
- Deprecated `EMPTY_OBJECT` property in `defaultValue` 
- Created standalone `EmptyObject` constant (via `EmptyObject.js`) per recommendation

Collaborated with @hotpocket 

## Issue number and link

Fixes Issue #11326 

## Testing plan

Tried to run this test in defaultValueSpec.js, but console.warn was not called: test did not run as expected.

```javascript
it("Calls console.warn", function () {
    spyOn(console, 'warn');
    defaultValue.EMPTY_OBJECT
    expect(console.warn).toHaveBeenCalled();
  });
```
# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Change Log
 
+##### Deprecated :hourglass_flowing_sand:
+
+- `defaultValue.EMPTY_OBJECT` has been deprecated and will be removed in 1.125. Use `EmptyObject` instead.
+
 ### 1.122 - 2024-10-01
 
 #### @cesium/engine
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -207,6 +207,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 
 ## [Individual CLA](Documentation/Contributors/CLAs/individual-contributor-license-agreement-v1.0.pdf)
 
+- [Paul An](https://github.com/anpaulan)
 - [Brandon Landry](https://github.com/hotpocket)
 - [Victor Berchet](https://github.com/vicb)
 - [Caleb Morse](https://github.com/cmorse)
--- a/packages/engine/Source/Core/EmptyObject.js
+++ b/packages/engine/Source/Core/EmptyObject.js
@@ -1,0 +1,8 @@
+/**
+ * A global immutable empty object for performance reasons
+ *
+ * @type {object}
+ */
+const EmptyObject = {};
+
+export default Object.freeze(EmptyObject);
--- a/packages/engine/Source/Core/defaultValue.js
+++ b/packages/engine/Source/Core/defaultValue.js
@@ -1,19 +1,16 @@
 /**
  * Returns the first parameter if not undefined, otherwise the second parameter.
  * Useful for setting a default value for a parameter.
- * @namespace defaultValue
- * @global
  * @function
- * 
+ *
  * @param {*} a
  * @param {*} b
  * @returns {*} Returns the first parameter if not undefined, otherwise the second parameter.
  *
  * @example
  * param = Cesium.defaultValue(param, 'default');
- * 
- * @property {object} EMPTY_OBJECT - A frozen empty object that can be used as the default value 
- * for options passed as an object literal.
+ *
+ * @property {object} EMPTY_OBJECT - A frozen empty object that can be used as the default value for options passed as an object literal.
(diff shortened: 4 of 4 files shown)


jjspace:
Thanks for the pr @anpaulan. However I don't think we want to make this change currently. I didn't remember there were 2 issues open related to this.

PR https://github.com/CesiumGS/cesium/pull/12207 for issue https://github.com/CesiumGS/cesium/issues/11674 is removing the `defaultValue` function and should be changing how we handle the `EMPTY_OBJECT` already. I don't think we want both of these changes. In that PR/issue we had settled on leaving it in the namespace of `defaultValue`. I believe  removing the function should also inherently fix the typescript issue too but we can be sure to check that in that PR.

Maybe you could work with @dave-b-b to combine the changes?

CC @ggetz

ggetz:
Good catch on the conflict here @jjspace.

To sum up the result of the discussion around #11674, it sounds like we'd want to:

* Deprecate `defaultValue` (the function)
* Keep `defaultValue` the namespace

Correct?

If that's the case, we should be able to do something like the following. If it becomes a namespace, `defaultValue` should be renamed `DefaultValue`.

```
/**
 * Utilities helpful for setting a default value for a parameter.
 *
 * @namespace DefaultValue
 */
const DefaultValue = {};

/**
 * A frozen empty object that can be used as the default value for options passed as
 * an object literal.
 * @type {object}
 * @memberof DefaultValue
 */
DefaultValue.EMPTY_OBJECT = Object.freeze({});

export default DefaultValue;
```

In that case, `defaultValue` and `defaultValue.EMPTY_OBJECT` will still need to be deprecated.

anpaulan:
> Good catch on the conflict here @jjspace.
> 
> To sum up the result of the discussion around #11674, it sounds like we'd want to:
> 
> * Deprecate `defaultValue` (the function)
> * Keep `defaultValue` the namespace
> 
> Correct?
> 
> If that's the case, we should be able to do something like the following and still resolve the typescript error. If it becomes a namespace, `defaultValue` should be renamed `DefaultValue`.
> 
> ```
> /**
>  * Utilities helpful for setting a default value for a parameter.
>  *
>  * @namespace DefaultValue
>  */
> const DefaultValue = {};
> 
> /**
>  * A frozen empty object that can be used as the default value for options passed as
>  * an object literal.
>  * @type {object}
>  * @memberof DefaultValue
>  */
> DefaultValue.EMPTY_OBJECT = Object.freeze({});
> 
> export default DefaultValue;
> ```
> 
> In that case, `defaultValue` and `defaultValue.EMPTY_OBJECT` will still need to be deprecated.

Hi! Thank you for sending this over. 

> Good catch on the conflict here @jjspace.
> 
> To sum up the result of the discussion around #11674, it sounds like we'd want to:
> 
> * Deprecate `defaultValue` (the function)
> * Keep `defaultValue` the namespace
> 
> Correct?
> 
> If that's the case, we should be able to do something like the following and still resolve the typescript error. If it becomes a namespace, `defaultValue` should be renamed `DefaultValue`.
> 
> ```
> /**
>  * Utilities helpful for setting a default value for a parameter.
>  *
>  * @namespace DefaultValue
>  */
> const DefaultValue = {};
> 
> /**
>  * A frozen empty object that can be used as the default value for options passed as
>  * an object literal.
>  * @type {object}
>  * @memberof DefaultValue
>  */
> DefaultValue.EMPTY_OBJECT = Object.freeze({});
> 
> export default DefaultValue;
> ```
> 
> In that case, `defaultValue` and `defaultValue.EMPTY_OBJECT` will still need to be deprecated.

Thank you for the guidance! I was going through this PR as well as PR #12207 and I think you are right. I can go ahead and deprecate defaultValue and update the name space. But when I do so, I would need to work with @dave-b-b to figure out what would be used in place of defaultValue? My understanding is when I deprecate defaultValue, we should then redirect folks to the replacement function.

ggetz:
> My understanding is when I deprecate defaultValue, we should then redirect folks to the replacement function.

Yes, this correct. 

I think it's be worth asking @dave-b-b what the current timeline on https://github.com/CesiumGS/cesium/pull/12207 is, and coordinate changes as needed. I'll add a message to that PR.

jjspace:
Closing in favor of https://github.com/CesiumGS/cesium/pull/12507

Issue #12126, opened 2024-08-15
Investigate visual quality defaults and new API for settings


jjspace:
Recently we've been making improvements to visual quality within CesiumJS. As part of this process we want to re-evaluate the default settings across the board to hopefully provide a better experience out of the box. We also want to explore a new API that makes it _easier for developers_ to modify the various visual quality settings in one place. We're also looking into the potential to have various of presets with a simple "High, medium, low" type setting like many games have.

Settings to look at
  - Resolution
  - Antialiasing
  - MSAA
  - FXAA
  - Lighting & atmosphere
  - HDR
  - Dynamic lighting
  - Shadows
  - Fog
  - Post Processing
  - Ambient occlusion
  - Bloom
  - Screenspace error

Example use cases to test/target and try and optimize for
-  AEC models, higher end _static_ visual quality is more important, not as oriented at intractability
- Google 3d Tiles and photogrammetry examples where intractability matters more, still want good visuals though
- Smaller scale site focused applications?
- Mobile oriented settings? Phones are a different focus and may want different defaults, specifically things like resolution

yoyomule:
Looking forward to it, Cesium needs to improve the model effect and appropriate performance balance.

jjspace:
Glad to hear that @yoyomule! We're definitely looking for input from the community through this process. If you have any defaults you use in your applications feel free to share them and your use case so we can keep it in mind.

jjspace:
[...]

Issue #8454, opened 2019-12-11
Fog culling too aggressively


lilleyse:
[Sandcastle](https://sandcastle.cesium.com/index.html#c=xZVfb9s2EMC/ysEvdgCFIimJklw3WJAM2MOKFWmwPUx9YKizzYUmDYqylw35Lvss+2SjLKdL0qFouw3hE3n/ePe7A5mm8J1erU89ds70QTsL0qugFQT0XmoLS+82ENYI56P88ts3sPXuF1QBZldoUHYI+UkCQRtsQdoW1q4LcXtzBxfY6X4DMSxpbJrCOoRtN0/T/X5PtitF+o0l2PZpK4NMx4tb3KSN3UkPO4179PAaLO6PkciPB9lsqg7HC2dDzBH9NIHfGwtxHdN+691Ot+jnD47Kowz4k/OmvR5NZieNvT951dhDZudtC+dGdrfSgnFKDii6xr6L9SjZBYNEtu0lLmVvwrVz5kb6N2j72c8f7v01zGF6iVYaPU1GYQwRAamoWPZWDSFnJw+JDmuskHQKLRIlN+glWZq7azd7ZDSsFrug7SGpDxVdSB/iTtqMDD26xJVH7GanjGekKBP484+8JgUfNgWhGFv0NKbzGiO+Meaz+4a1Rtlqu5oDJ0Vdl7xKPrbZ6qDWczilhP+D1jtj5kAJfaq6//t44H+U3T/q4YHluVVr5+VqGD2UL8yUE1YkkJdEiAR4Vfx/NCNMxmlWivoLkD4zbXt/TIXRz6T9FuVt9xKQryIYaSNkTkTNq0qIilPOeMWyBBihJc/qnBcFL7OqqBPISkq/mj4jIqMV57zOq6qqc5F/cqpFXmZ5RZkoKKeszv77Ib/SO/Q32L40+awuBM/yuhQ0UhYH8vGQ5xkr87ymWTHMPS2/Hr0gVFRlnTNaiaIuWP1J8jELljPBytiksqiY+Pfo3ycwDePTPR3kk2Sy6MKdwbPR5hu92TofoPdmRkgacLM18c/o0pte3WIgqusGt0X64LRo9Q50+7qZPPuOmgmo+Jd0UbPsjXmnf8NmcrZIo/0TN+MObH6II2Dk3WCyZmffj0JCyCKNx4+9jjU8ivgX) - click Anchorage Area

[...]

PR #3154, opened 2015-10-30
Fog

bagnell:
- [x] Tweak fade distance and density
- [x] Change tile SSE when in fog
- [x] Use preprocessor instead of conditional when enabled.
- [x] Swap shader when a tile is or is not in fog
- [x] Performance
- [x] Tests
- [x] Reference doc including advice on the performance / visual quality trade-off
- [x] Update CHANGES.md

PR #12248, opened 2024-10-10
expose new scalar to change the visual density of fog

jjspace:
# Description

- Add a new property to adjust the _visual_ density of fog relative to the "actual" density which also affects culling
- change the default visual scalar from `0.15` to `0.4` which should make the densest fog possible right around where tiles are culled

| Old default `0.15` | New default `0.4` |
| --- | --- |
|download (41) | download (42) |
| download (43) | download (44) |

<details><summary>Density function details</summary>
<p>

These graphs show the density function values. The y value between 0 and 1 is used to determine the percentage for mixing the original color and the fog color. `mix(originalColor, fogColor, percent)`
ie. closer to 1 means more fog color

Any tile with a value greater than 1 is culled and thus not rendered
Any tile or model with a value less than `0.001` does not get any fog color at all to aid in performance, the fog would be negligible.

The red line is the "true" fog function that is calculated without any scalar. This is the one used for culling
The purple line is the fog function with the modifier included. The purple area is where fog is actually being displayed

This is the before value with a modifier of `0.15`

This is the same functions with a modifier of `0.4`. As you can see the fog (purple) much closer matches reaching 1 around the same time the red fog function does. This means it's visually most dense at around the same distance tiles are culled.

The density is cranked up way higher than normal to make seeing the curves clearer
[Desmos link](https://www.desmos.com/calculator/ujgnokdrhq) if you want to play with the values yourself
(the `v` value is there to exaggerate the graphs visually so you can still see the curves even if you stretch the density out really far)

</p>
</details> 

## Issue number and link

Part of https://github.com/CesiumGS/cesium/issues/12126

## Testing plan

- Open the updated Fog sandcastle and play around with the visual density scalar value
    - Modify the actual density with different scalar values to see how they affect each other.

# Author checklist


ggetz:
> This is the same functions with a modifier of 0.4. As you can see the fog (purple) much closer matches reaching 1 around the same time the red fog function does. This means it's visually most dense at around the same distance tiles are culled.

Good explanation! So for the default, we wouldn't want to go any less that `0.4` from the sound of it.

I was playing around with GE as a point of comparison, and their fog appears much denser, even than visual density equivalent of `1.0` in our system (which I think is the maximum, based on the function visualizations). What are your thoughts?

image

packages/engine/Source/Core/Math.js
```diff
@@ -1048,6 +1048,8 @@ CesiumMath.log2 = defaultValue(Math.log2, function log2(number) {
 });
 
 /**
+ * Calculate the fog impact at a given distance. useful for culling.
```
ggetz, line 1051:
```suggestion
 * Calculate the fog impact at a given distance. Useful for culling.
```

packages/engine/Source/Renderer/AutomaticUniforms.js
```diff
@@ -1577,6 +1577,20 @@ const AutomaticUniforms = {
...
+   * An automatic GLSL uniform scalar used to mix a color with the fog color based on the distance to the camera.
+   *
+   * @see czm_fog
+   */
+  czm_fogVisualDensityScalar: new AutomaticUniform({
+    size: 1,
+    datatype: WebGLConstants.FLOAT,
+    getValue: function (uniformState) {
+      return uniformState.fogVisualDensityScalar;
+      // return 4.0;
```
ggetz, line 1590:
```suggestion
```

packages/engine/Source/Scene/Fog.js
```diff
@@ -34,6 +34,11 @@ function Fog() {
    * @default 2.0e-4
    */
   this.density = 2.0e-4;
+  /**
+   * A scalar that impacts the visual density of fog. This value does _not_ impact the culling of terrain.
```
ggetz, line 38:
```suggestion
   * A scalar that impacts the visual density of fog. This value does not impact the culling of terrain.
```

It's not markdown 😄

packages/engine/Source/Scene/Fog.js
```diff
@@ -34,6 +34,11 @@ function Fog() {
...
+   * Use in combination with the `density` to make fog appear more or less dense.
```
ggetz, line 39: `@type`?
`@default`?

Also, I know there are no code examples in this file, but I think it would benefit users.

I'm thinking some super simple examples fro the common properties, i.e.

```js
// Disable fog in the scene
viewer.scene.fog.enabled = false;
```

```js
// Increase fog appearance effect
viewer.scene.fog.visualDensityScalar = 1.0;
```

ggetz (changes requested):
Thanks @jjspace. A few comments on the code. 

Please also add a unit test and update CHANGES.md.

packages/engine/Source/Scene/Fog.js
jjspace: I wish it was :shrug:

jjspace:
> I was playing around with GE as a point of comparison, and their fog appears much denser, even than visual density equivalent of 1.0 in our system (which I think is the maximum, based on the function visualizations). What are your thoughts?

I actually think the _density_ of our fog is about the same. But ours is _much_ darker. Maybe this has to do with some of the atmosphere values?

This is approximately the same location in CesiumJS (with 0.4) and Google Earth. (I just noticed GE seems to put a touch of a fish eye effect on the fov?)



This is that same view with the `visualDensityScalar` set to `1.0`. I think this looks _much_ more dense than GE's so i'm surprised to hear you say the opposite. does this match with what you saw?

ggetz:
> I did just notice our fog seems inconsistently applied as we zoom in and out and very close to the ground it's not applied at all which I think might be the cause of your previous comment about it not being as dense as GE. They seem to apply fog consistently across all zoom levels. Not sure the cause for this.

Yes, I was looking at a horizon few similar to this one. This is what prompted my question. I agree that from your more "aerial" screenshots, the density does read correctly at a value of `0.4`.

My guess for what is causing this may be the atmosphere scattering code returning transparent colors for fog when close to the ground. Perhaps try stubbing out the atmosphere color function with one that returns a solid color for the sake of debugging. Once we identify the cause, we can figure out an implementation plan, and if it's scope.

If that's not the case, my shot in the dark is that this may be related to https://github.com/CesiumGS/cesium/issues/11922. The horizon angle is what's jumping out to me as similar.

> But ours is much darker. Maybe this has to do with some of the atmosphere values?

I agree. I assume that will be out of scope for this PR. I believe the reasons for this are documented in https://github.com/CesiumGS/cesium/issues/11717#issuecomment-1889894826.

CHANGES.md
```diff
@@ -7,6 +7,7 @@
 ##### Additions :tada:
 
 - Added `ScreenSpaceCameraController.maximumTiltAngle` to limit how much the camera can tilt. [#12169](https://github.com/CesiumGS/cesium/pull/12169)
+- Exposed `Fog.visualDensityScalar` to allow modifying the visual density of fog without affecting the culling aspects. This defaults to `0.4` but previously had an internal default value of `0.15`. Set `viewer.scene.fog.visualDensityScalar = 0.15` to get the previous behavior. [#12248](https://github.com/CesiumGS/cesium/pull/12248)
```
ggetz, line 10: > This defaults to `0.4` but previously had an internal default value of `0.15`. Set `viewer.scene.fog.visualDensityScalar = 0.15` to get the previous behavior.

I think this warrants a mention under "Breaking Changes" as well.

packages/engine/Source/Scene/Fog.js
```diff
@@ -7,6 +7,9 @@ import SceneMode from "./SceneMode.js";
  * Blends the atmosphere to geometry far from the camera for horizon views. Allows for additional
  * performance improvements by rendering less geometry and dispatching less terrain requests.
  *
+ * View the {@link http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FFog.html|Fog Sandcastle}
```
ggetz, line 10: In other places, we appear to use `@demo` tag. For example

```
 @demo {@link https://sandcastle.cesium.com/index.html?src=Box.html|Cesium Sandcastle Box Demo}
```

Also, like that example, please make sure to link to the deployed Sandcastle link, since user may be viewing docs online.

packages/engine/Source/Scene/Fog.js
```diff
@@ -15,13 +18,20 @@ function Fog() {
...
+   * viewer.scene.fog.enabled = false;
    */
   this.enabled = true;
   /**
    * <code>true</code> if fog is renderable in shaders, <code>false</code> otherwise.
    * This allows to benefits from optimized tile loading strategy based on fog density without the actual visual rendering.
    * @type {boolean}
    * @default true
+   * @example
+   * // Use fog culling but do't render it
```
ggetz, line 32:
```suggestion
   * // Use fog culling but don't render it
```

jjspace:
@ggetz I fixed up the couple docs issues and moved the fog sandcastle as discussed offline.

That said I'm still confused what's causing the fog to be basically non-existant close to the ground and have such a massive spike in density at a certain height. I tried swapping the fog color out for a static magenta `1, 0, 1` instead of the atmosphere color and still observe the same issues. I'm thinking it might have something to do with the depth/distance calculations? I made a [slightly modified sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=vVh7b9s4Ev8qhO+AdQCb1ot6OG5w2zZFC7RN0aS7wNXFLS3RsRCJ9IpUErfId7+hXpZtuXaT3P0RR3zM6zfDmSFDwaVCtzG7Yxl6gTi7Q6+YjPMU/1HM9ae9sBi/ElzRmLNs2hugH1OOkGJZBjPjmuCqHON5JtI/RZZE1UT/ZDDlDyenUz7lo9HVghXiPoiIJUhlNLyRSMGkVFQxJOZI5BlKYx4julwmcUhVLDie8rDRtCR9UWrBOJ0lLBoDq5wN9EzEuIzVaoyMYngby5wmr8vJy5AmNKuXpGT1pxaY5unLLL5eKM6kLBceTrXOCIy/ZZkq9FxrkLJ0xjKJYq4EuuEivBG5QmImWXarlZKgdYVNvYoLg/sNjxoV9DLm0RZ7YKonXl98QCxhKeNKanj03Jd38I8qBNYkaC4yFKsGISVEMqPamZEIc02Gr5k6Lzm8XL2LwKXVnmlPy99WUcO+0vrE/FquVR3UnEudt6g0hlrORWN9m3Laq9wEEoudMp/JMItnrD/Peag9jPoQe3/QJGcnpWNRFZVYhowzPBfXuGJSxmmx91TvfHiURlWYPE2jisnzaNQRqU/TroPh82i6c1qepucOu04tt+P0J/rBwQaNjtFmRxcgYIxfLmnIzrNMZG9oqMQWbBVojchWYO6L2dP29nXU7Auoje3dbjzG2xtsAJQusm6DNyi7/HPYiQVGl5RHIZUqYZhG0VWZQV7mSgne1/hPe29FFn8HzyyADtFExSqPwHtFVl67bTt4QpqyjGLJlK5T/WpRJ3+pYl7UjHG7mr2imYIvyu1+vRWhoeW4nmcbmHiO4Zue67iD1qpjBwYxPGybxLccYjqtRVhyTUKwadq26Qa236ydNF9gGKTcSpcfa9oFozq3jpGDncCyTJNYpk9c1yItActYhYsxGhrYcn3PtAPH9tzAtY21Fg/VVxGO9XDa+y5EWiIsCxhPyhoWsTnNE1WXlQ24UQln6fM96B6N7H5U9yLajWaJZBeKBxE8gN7DuiM5PkATcff/jk/PdgzDxAExTZe4PiEb4WmZjhEAzIELtvl2OzwdDxAOfGxaxPdd2/QeE54Eu7ZjEc8GXEngBF5neBpmQByfGMTzie2bhm09JkCPccQlp0u5EGo/+GUHFNat6k97oJ22tlSyZqLS5Vugb3hhqVag2oLp9La988+OnXdxpBaFZXorHMAMBH5nRbtW5tzClppRhxRgOu25hrG8n/ZOuzcWQop9/sa+KgpLkf2T7VkeQWd/0lYuXEDXztAsEbofF0vGIQRQRBVFeZZIFMUZC1WyGiBwDqIcxSm9ZjqZ0CKM76BfhPMRAyAQPg3bhVJLOR6NoLmHog099BxOEQ5FOvo71ycBABw5xPN8zzJGJY+hFl78yCEdzhLKb4YlXsOYD0s92/ADFQBQ0mJN1pgLc7jx/10WK9b/axKn10hm4Ytp758/mrPKb6nESrwGc798fg/BUVg3WvJrCIuHaQ+Nzv6quYJVUollkURnmbiDDgTpKw9cQFYaM4AkETQCl3ChFrFmoX3OwPi1WprDpgPAVay5iGmvHYgMHZ8HgkIH5i/Gw6FDWh2wIrm9F+XdzKyvYr9QISDHe4GHCTEN13A8rykQVgDZmmBiekbgeE0+t33ieIaBPfh1TMd2nX0VoiX3bZnIPulU9VkkSSXdwS4hBggOLMirQdAI0dUCkqUfODpl2r5v14XLxDZxoZIZOh16kE5tNvRaGkR5VokveTEqQfCbKkE1N+TzjWn8/t3H898//+fjxcfz6qrZXTy24N7vCesxnvB1gca+EdiW77iOva7Vhq0hN8B0sNu1165wPd/BPpRqywzM5/TEHkcEvkG2PWHbbkCI/T9yRI0vdL9U9+n9VJ6gF2eFTZ/gsMeS9ftZMQVeuopTBjeSfjZAsE/7h8oVD9cVap7oU/uJqkU/qXwl4T4NZL+rcxouNi8jW+5vKL4a38ozqktIv1SwXtQvA83GPc3IPFldiYZdnQToHY0rQ/stjcpkAGgcrM29f0OqQBe5QlcQdvp5aKcyt+z/uhXNg+0Y/jZApmEYJ/U96wjZ7/hjRFvbos1t0d2nzH5UvjMC18Wu61muZbTTHRw7AmfM9x3LMd31GSPE97BDdPfqQAv3fIfM6z5kgWsEm2eMBKZv2ZYNyuw7ZMYTD9nPffuB3kOoQ7leim7fdoX3lq8aX/YGvUlRHs/q7vRfcboUmdL9TR/jkWLpMqHguNEsD2+YwqGU9RlB6B/1s1qrUZ5BT3OdiZxHY5Rdz2jfgZCq/wzsr6mhXwb7yhvL8r41PRMZlN5hBp7L5ebiQyO6VYrb4qNYgsKrMYQ4a/PUw2HZrxXuCEWSp7y1gcKRgGYKDJYl8RDas0ztSo75Mldf1WrJoF3KKL+Ge8+3tgpFpzFGLksPEPNcP5V2U5Nd6smo7apJFN+iOHrR0bOjMKFSwso8T5JL3TX1ziYj2L9DqnsycMAFtGEAmt62MM/el5MY48kIht2UzWNpEzkTpd90ztaWTNRMRKvWhJ7KNsZ6JjqrXoMmI/jeXZ0UmKEKs3DBwpuZgKa+6MOHM2hx62n94N08qZ7tsIOJ7AhtqsemPdpszsDchnZVOGyodqvfx8b1A/wAFeMvS9gCs78V5L8BSRpz2AydBdxt9ZDeV0MLRnB/WLZWD2hRx9Uv8fyJzjsCHwtt+RhXs4ULn36OexLQjYVr+zZt22NZx7PgPs/8MthPV+W5ANc9EQgq9B1C0QAVZ63X6bM3UEdgDq0nnzvqd94/D8V/g525jd2xfjhGi0fIe2zM6wfmmMNVncLHvHhGPirRKXavtEqQvmFEug0rnvQPZjoYbmRiGG+k6nVW37jfbmb+1vd/AQ) to easily reproduce the zoom.

https://github.com/user-attachments/assets/8a65c731-5a20-4773-89f7-04e54770f58f

ggetz:
> That said I'm still confused what's causing the fog to be basically non-existant close to the ground and have such a massive spike in density at a certain height. I tried swapping the fog color out for a static magenta 1, 0, 1 instead of the atmosphere color and still observe the same issues. I'm thinking it might have something to do with the depth/distance calculations?

Hmm, I see it as well. I wonder if we need to scale based on the camera height. For instance, here's an exponential decay function based on camera height which seems to be producing decent results:

```
const float maxHeight = 10000.0;
const float minHeight = 500.0;
float height = (czm_eyeHeight - minHeight) / (maxHeight - minHeight);
float heightModifier = 1.0 + exp(-decay * height);

const float modifier = 0.4;
finalColor = vec4(czm_fog(v_distance * heightModifier, finalColor.rgb, fogColor.rgb, modifier), finalColor.a);
```

jjspace:
@ggetz thank you for the pair programming on this. recapping a little here.

- Ever since fog was [initially implemented](https://github.com/CesiumGS/cesium/pull/3154) it's used an internal table of heights and densities at those heights that seems to be somewhat arbitrary. These were not a fully smooth curve and were the cause of the weird "spikes" in density and practically no fog near the ground that we were observing.
- Those tables have been replaced with a smooth power function that somewhat closely matches the previous values.
- The "magic numbers" in that new function have been exposed as a set of new properties on the `Fog` class to allow customization. We should have good defaults so they don't normally have to be altered but the option is there.
- With this latest change I also updated the `visualDensityScalar` up to `1` from `0.4` which seems to have a nicer effect. We may even want to default to slightly higher.

I think the function and implementation is good now. The sandcastle has been updated with the new properties. Please play with them a little and see if you think there's a better set of default values, changing the numbers is easy.

packages/engine/Source/Scene/Fog.js
```diff
@@ -32,8 +40,41 @@ function Fog() {
...
+   * // Double the default fog density
+   * viewer.scene.fog.density = 0.0004;
    */
   this.density = 2.0e-4;
+  /**
+   * A scalar used in the function to adjust density based on the height of the camera above the terrain.
+   * @type {number}
+   * @default 0.001
+   */
+  this.fogScalar = 0.001;
```
ggetz, line 53:
```suggestion
  this.heightScalar = 0.001;
```

Fog should be implicit since this is the `Fog` class... but height may not be.

packages/engine/Source/Scene/Fog.js
```diff
@@ -32,8 +40,41 @@ function Fog() {
...
+  /**
+   * Exponent factor used in the function to adjust density based on the height of the camera. Higher values will lead
+   * to more dense fog at lower heights and a slower falloff as you get higher.
+   * Positive values only, if the value is negative it will be clamped to 0;
```
ggetz, line 57: Could we add a `Check` for this value via a setter?

ggetz, line 57:
```suggestion
   * Exponent factor used in the function to adjust how density changes based on the height of the camera above the ellipsoid. Smaller values produce a more gradual transition as camera height increases.
   * Value must be greater than 0.
```

Totally borrowing from [Unreal's doc](https://dev.epicgames.com/documentation/en-us/unreal-engine/exponential-height-fog-in-unreal-engine) here.

packages/engine/Source/Scene/Fog.js
```diff
@@ -145,18 +122,13 @@ Fog.prototype.update = function (frameState) {
...
-
-  // Again, scale value to be in the range of densityTable (prevents divide by zero) and change to new range.
-  const startDensity = this.density * 1.0e6;
-  const endDensity = (startDensity / tableStartDensity) * tableEndDensity;
-  density = density * (startDensity - endDensity) * 1.0e-6;
+  let density =
+    this.density *
+    this.fogScalar *
+    Math.pow(
+      Math.max(height / this.maxHeight, 1e-6),
```
ggetz, line 129:
```suggestion
      Math.max(height / this.maxHeight, CesiumMath.EPSILON6),
```

ggetz: What happens when `this.maxHeight` is `0`?

packages/engine/Source/Scene/Fog.js
```diff
@@ -145,18 +122,13 @@ Fog.prototype.update = function (frameState) {
...
+      -Math.max(this.heightFalloff, 0),
```
ggetz, line 130:
```suggestion
      -Math.max(this.heightFalloff, 0.0),
```

ggetz (changes requested):
Thanks @jjspace! I'm glad this all cleaned up nicely!

Do you think we can say this addresses https://github.com/CesiumGS/cesium/issues/8454?

In addition to updating tests, please make sure new or properties are reflected in `CHANGES.md`.

I'm testing out behavior now.

packages/engine/Source/Scene/Fog.js
jjspace: > What happens when `this.maxHeight` is 0?

It functionally turns off fog unless you're underground because of the check at the beginning of this function for if you're in space. If you're under height 0 the max value kicks in here and the divide by 0 doesn't seem to be a problem

ggetz: Conceptually I agree! In practice though, I think there is a case where, if `maxHeight` is `0` and `height` happens to be `0`, we can get a value of `NaN`. Can we prevent that?

ggetz:
@jjspace I did some quick network testing:

* On average, the current settings are using around 1.5MB network for horizon views at various heights
* To get the same network usage as in `main`, I used the following settings:
````
viewer.scene.fog.density = 0.001;
viewer.scene.fog.heightFalloff = 0.65;
````
* What do you think of splitting the difference? This is, on average, using `0.4MB` more network, and I'm still seeing increased resolution at the mid-range and closer to the camera.
````
viewer.scene.fog.density = 0.0006;
viewer.scene.fog.heightFalloff = 0.59;
````

**Comment by jjspace** [comment]
@ggetz I think I'm fine with those values, especially if it helps with the network cost.

I was also thinking of switching the `visualDensityScalar` back down to `0.5` or `0.6` instead of `1` as I think 1 gets a bit too aggressivly dense at higher altitudes. Thoughts?

I'm still working through the tests, mostly value checking which these numbers but that should just be a simple value update.

**Comment by ggetz** [comment]
> I was also thinking of switching the visualDensityScalar back down to 0.5 or 0.6 instead of 1 as I think 1 gets a bit too aggressively dense at higher altitudes. Thoughts?

I did some digging into [why fog is so dark](https://github.com/CesiumGS/cesium/issues/12262#issuecomment-2438748126), and I think that the opacity of the ground atmosphere color is not being correctly handled. If we go with something similar to the fix I have there, I think `1` works, as the fog closer to the camera will actually be pretty thin and therefor the color more transparent.  However, I doubt we could get that fix in before the next release. 

If possible, I'd like to avoid breaking changes for this default. Maybe a weird idea, but what do you think of marking this value as `experimental` for the first release in case we want to adjust it after the fact?

**Comment by jjspace** [comment]
@ggetz I've added tests and marked the new scalar as experimental. I think that's fine for now but I also thought that if we want to update this in the future we could switch it back to `0.15` like it was before this PR to minimize impact until we get your other fix in?

**Comment by ggetz** [comment]
> I think that's fine for now but I also thought that if we want to update this in the future we could switch it back to 0.15 like it was before this PR to minimize impact until we get your other fix in?

Sure, please change it to whatever the equivalent was for the previous behavior.

File: packages/engine/Source/Scene/Fog.js [hunk 12]
```diff
@@ -31,9 +41,36 @@ function Fog() {
    * 1000.0m above the ellipsoid, increasing the value to 3.0e-3 will cause many tiles close to the viewer be culled.
    * Decreasing the value will push the fog further from the viewer, but decrease performance as more of the terrain is rendered.
    * @type {number}
-   * @default 2.0e-4
+   * @default 0.0006
+   * @example
+   * // Double the default fog density
+   * viewer.scene.fog.density = 0.0004;
```
ggetz, line 47: Am I missing something obvious, or should this be `0.0012` given the comment?

packages/engine/Source/Scene/Fog.js
```diff
@@ -31,9 +41,36 @@ function Fog() {
...
+   */
+  this.density = 0.0006;
```
ggetz, line 49: Make sure you mention the adjusted default value in `CHANGES.md`.

ggetz (changes requested):
Thanks @jjspace! This is super close!

> I've added tests and marked the new scalar as experimental. I think that's fine for now but I also thought that if we want to update this in the future we could switch it back to 0.15 like it was before this PR to minimize impact until we get your other fix in?

I might have misunderstood you, but comparing main to this branch, I actually think a a value of `0.5` approximates the existing behavior better than `0.15` does. I think this is due to the changes made to the `density` itself.

packages/engine/Specs/Scene/FogSpec.js
```diff
@@ -0,0 +1,170 @@
...
+  describe("update", () => {
+    /** @type {Scene} */
+    let scene;
+    /** @type {Fog} */
+    let fog;
+
+    beforeEach(() => {
+      fog = new Fog();
+      scene = createScene({
+        canvas: createCanvas(10, 10),
```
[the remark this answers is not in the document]
jjspace: Copied this from another test to recreate the `frameState` for every test. didn't realize you don't need to pass in the canvas

packages/engine/Source/Scene/Fog.js
jjspace: oversight from changing the default

jjspace:
@ggetz I believe this should now have all your comments addressed

ggetz (approved):

ggetz:
Looks good! Thanks @jjspace!

Issue #11958, opened 2024-04-25
`Unhandled promise rejection: undefined thrown` in CI


ggetz:
I'm seeing some intermittent CI failures in the release test task (but not coverage) which appears to be due to a race condition in the `Cesium3DTileset` specs:

```
  Scene/Cesium3DTileset
    ✗ verify statistics
	Unhandled promise rejection: undefined thrown
```

jjspace:
I've noticed this one happening again more recently

https://github.com/CesiumGS/cesium/pull/12007#issuecomment-2140853488
https://github.com/CesiumGS/cesium/actions/runs/9354720455/job/25748200946
https://github.com/CesiumGS/cesium/actions/runs/9352809979/job/25741821665

jjspace:
These seem to be much more frequent than "intermittent" lately. I feel like I'm re-running a job almost daily because of this one "verify statics" test failure. any chance we can try and track this one down soon or exclude it completely until it can be fixed so we can rely on the little red `X` from CI actually being an error more?
https://github.com/CesiumGS/cesium/actions/runs/10684175472/job/29613965055

ggetz:
Let's exclude it for the time being, but leave this issue open to track the problem.

jjspace:
Been seeing more of the other "statistics" tests failing apart from the one I already excluded

https://github.com/CesiumGS/cesium/actions/runs/10903427776/job/30257483763
```
  Scene/Cesium3DTileset
    ✗ verify batched features statistics
	Unhandled promise rejection: undefined thrown
```

lukemckinstry:
This continues to pop up https://github.com/CesiumGS/cesium/actions/runs/11126864591/job/30917806637 this is from CI after merging to main

```
1) verify batched features statistics
     Scene/Cesium3DTileset
     Unhandled promise rejection: undefined thrown
```

ggetz:
I'll add that these have started to bubble up in pretty much 50% or more of CI runs, making the process pretty clunky.

ggetz:
Another potential clue: I _think_ I am only able to reproduce locally when using the `--webgl-stub` option.

jjspace:
I was also able to trigger the statistics `undefined thrown` error with `--webgl-stub`

javagl:
> I was also able to trigger the statistics undefined thrown error with --webgl-stub

[...]

PR #12250, opened 2024-10-12
Debugging CI failures

javagl:
This mainly refers to https://github.com/CesiumGS/cesium/issues/11958 (and the tests that already have been excluded due to this error). 

Unfortunately, these errors **only** seem to happen in CI (at least, I could not yet reproduce them locally). But they nearly **always** happen there (and I had to re-run CI several times for the last PRs). 

I don't know a sensible way of debugging something like that. If someone knows a _deterministic_ way: Let me know. 
And I cannot imagine why someone should throw `undefined` anywhere to begin with. If someone knows a reason (and a way how to even figure out where it was thrown): Let me know. 

Until then, let's see whether I can make that annoying, frustrating green checkmark that I already see below turn into an ❌ ...

### DIFF
--- a/Specs/pollToPromise.js
+++ b/Specs/pollToPromise.js
@@ -15,6 +15,8 @@ function pollToPromise(f, options) {
       try {
         result = f();
       } catch (e) {
+        console.error(`Error '${e}' thrown in polled function ${f}`);
+        console.trace();
         reject(e);
         return;
       }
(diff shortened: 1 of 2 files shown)


javagl:
Hm. Great. Apparently, it's not even possible to simply dump out some `console` logs during CI. 
If anyone has an idea about how to address that: Let me know...

ggetz:
@javagl You can run the tests in CI without an open PR– You only need to push up the branch. I might recommend closing this PR until there's a proposed solution.

In the meantime, running tests hides output by default (our tests and source code can actually be quite noisy). This can be disabled with the `--verbose` option. To run that in CI, in context you would edit `dev.yml` to make the `release tests (chrome)` task run:

```
npm run test -- --browsers ChromeHeadless --failTaskOnError --webgl-stub --release --suppressPassed --verbose
```

Issue #12182, opened 2024-09-04
Sourcemap of Cesium.js (iife) points to missing source files (v1.121.0)


s3xysteak:
### What happened?

#11993 was fixed in `v1.118` but it comes back in `v1.121.0`.

> Sourcemap for "/foo/bar/node_modules/.pnpm/cesium@1.121.0/node_modules/cesium/Build/CesiumUnminified/Cesium.js" points to missing source files

### Reproduction steps

1. pnpm i
2. pnpm dev

The warning will be throwed by vite server.

### Sandcastle example

none

### Environment

Browser: Edge
CesiumJS Version: v1.121.0
Operating System: windows 10

jjspace:
Thanks for catching this @s3xysteak
We should not normally be publishing source maps. I believe this slipped in after I ran an extra `build` command while troubleshooting some issues publishing `1.121`. This should be addressed in `1.121.1` which was just published. I'll be opening a separate issue to start taking a look at some better release automation/checks so this sort of thing doesn't happen again shortly.
Please let me know if the new version doesn't fix this

s3xysteak:
Thank you for such a quick response! In my application, it has been fixed in `v1.121.1` so I gonna close the issue.

s3xysteak:
@jjspace It came back again in `v1.122.0` 😇

jjspace:
Thanks for bringing it up again @s3xysteak. Not really sure what the cause is, our release guide _should_ be preventing this but maybe something's slipping through. We're not planning to do a minor release this time but this should be resolved on future releases.

@ggetz in addition to potentially reviewing the steps in the release guide I had some other thoughts for more automated prevention

The `make-zip` build process does not generate the sourcemaps at all but `build` does.
I see a couple options to fix/prevent this but not sure what's the best process wise?

1. add `/Build/**/*.js.map` to the `.npmignore` so it never gets published
    - I think this would work but could also end up being a "band-aid" that bites us later if `build` and `make-zip` generate different JS files
2. Update the release guide to say run `npm run make-zip` again right before the publish steps
3. add a `prepublishOnly` script to the `package.json` to run `gulp makeZip` before every publish to guarantee it's the files we want
    - I think this is the best case? it makes the publish command a bit slower to wait for it to generate the docs and stuff but is probably the safest and most foolproof

Thoughts on some or all of the above changes?

CC @lukemckinstry since you're doing the next release

ggetz:
[...]

PR #12263, opened 2024-10-28
Don't rebuild during the release process

jjspace:
# Description

The past couple releases we've had some extra `.map.js` files bleed through into the published version. These cause warnings for users downstream. See https://github.com/CesiumGS/cesium/issues/12182

It seems these are generated when we run `npm run test` even with the `--release` option. I also noticed these would be generated any time we run `npm start` without `--production` as the dev server does a build when it starts up.

This PR adds a check for the release flag to skip building for `npm run test`. It also updates the release guide to specify the dev server should only be started with the `--production` flag.

One extra precaution we could add is a `prepublishOnly` [lifecycle script](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts) to our `package.json` that calls `npm run make-zip` right before publish. Would make the publish command slower but guarantee we always build the version of files we want published.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12182

## Testing plan

- Run `git clean -dxf` to remove all extra files
- Run `npm run make-zip` like you would during the release
- Run `npm run test -- --failTaskOnError --release` like in the [release guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/ReleaseGuide/README.md)
- Make sure there are no messages about building and no `.map.js` files generated

# Author checklist

### DIFF
--- a/Documentation/Contributors/ReleaseGuide/README.md
+++ b/Documentation/Contributors/ReleaseGuide/README.md
@@ -41,7 +41,7 @@ There is no release manager; instead, our community shares the responsibility. A
 13. Run `npm install`.
 14. Make sure `ThirdParty.json` is up to date by running `npm run build-third-party`. If there are any changes, verify and commit them.
 15. Create the release zip `npm run make-zip`.
-16. Run tests against the release `npm run test -- --failTaskOnError --release`. Test **in all browsers** with the `--browsers` flag (i.e. `--browsers Firefox,Chrome`). Alternatively, test with the browser Spec Runner by starting a local server (`npm start`) and browsing to http://localhost:8080/Specs/SpecRunner.html?built=true&release=true.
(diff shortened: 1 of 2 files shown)


ggetz:
Awesome, thanks @jjspace!

> One extra precaution we could add is a prepublishOnly [lifecycle script](https://docs.npmjs.com/cli/v10/using-npm/scripts#life-cycle-scripts) to our package.json that calls npm run make-zip right before publish. Would make the publish command slower but guarantee we always build the version of files we want published.

We can follow up about this in a separate thread when appropriate. But the only concern I have with this is that it means the versions of each of the files we just tested will not be the ones that get published. Perhaps a simpler measure would be to add the maps to .npmignore?

ggetz (approved):

jjspace:
> Perhaps a simpler measure would be to add the maps to .npmignore?

The problem with this when coupled with our current build process is that the flow when maps are generated is slightly different than the flow when they are not. If we just ignored the map files it could hide the fact that the files were rebuilt differently than we may want them to be.

PR #12279, opened 2024-11-01
Closes #12278 - added filtering

### LINKED ISSUE #12278: control characters in labels causes the render to stop
### What happened?

if you try to render a label with a control character like /u200f, the engine will stop after an execption

### Reproduction steps

i will soon open a PR to fix this by filtering all problematic characters

### Sandcastle example

_No response_

### Environment

Browser:
CesiumJS Version:
Operating System:

MNPCMW6444:
# Description

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12278
issue #12278 

## Testing plan

# Author checklist


ggetz:
Thanks for the PR @MNPCMW6444!

I can confirm we have a CLA on file for you.

packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -118,8 +118,74 @@ function addGlyphToTextureAtlas(textureAtlas, id, canvas, glyphTextureInfo) {
 
 const splitter = new GraphemeSplitter();
 
+// Filter to remove unsupported control characters like RLM (\u200f)
```
ggetz, line 121: Can you explain a bit about how this list was determined? Would it make sense to test for [a unicode range](https://stackoverflow.com/a/51602415) instead of listing each character individually?

packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -118,8 +118,74 @@ function addGlyphToTextureAtlas(textureAtlas, id, canvas, glyphTextureInfo) {
...
+  console.log(text, text.replace(problematicCharactersRegex, ""));
+  return text.replace(problematicCharactersRegex, "");
+}
+
 function rebindAllGlyphs(labelCollection, label) {
-  const text = label._renderedText;
+  let text = label._renderedText;
+
+  // Filter out unsupported control characters
+  text = filterUnsupportedCharacters(text);
```
ggetz, line 187: Regex operations can be performance intensive. Is there a way we can ensure the operation runs only once for each label text string, as apposed to every time the glyphs need to be rebound?

packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -118,8 +118,74 @@ function addGlyphToTextureAtlas(textureAtlas, id, canvas, glyphTextureInfo) {
...
+    "slash u0017",
+    "slash u001D",
+  ];
+
+  const regexPattern = problematicCharactersByTesting
+    .map((char) => char.replace("slash u", "\\u")) // Replace "slash u" with "\u"
+    .join("|"); // Join with | to match any character in the list
+
+  const problematicCharactersRegex = new RegExp(regexPattern, "g");
+  console.log(text, text.replace(problematicCharactersRegex, ""));
```
ggetz, line 179:
```suggestion
```

ggetz (changes requested):
Could you please add a unit test to verify this fix?

MNPCMW6444:
@ggetz Thanks for your CR!
For the purpose of determing what characters are problematic (turns out not all control chars) and TESTING after the patch  I added a full test to sandcastle. but I didn't commit it to PR because I don't think we want it in the sandcastle.
you can see it here - https://github.com/MNPCMW6444/cesium/tree/control-characters-in-labels-causes-the-render-to-stop-%2312278-with-full-test
tell me if you still think I need a unit test or any other comment.

packages/engine/Source/Scene/LabelCollection.js
MNPCMW6444: I can implement this with a simple .replace(char, '') instead of a RegExp, will it be acceptable?

packages/engine/Source/Scene/LabelCollection.js
MNPCMW6444: see my comment please

MNPCMW6444:
> Could you please add a unit test to verify this fix?

@ggetz Thanks for your CR!
For the purpose of determing what characters are problematic (turns out not all control chars) and TESTING after the patch I added a full test to sandcastle. but I didn't commit it to PR because I don't think we want it in the sandcastle.
you can see it here - https://github.com/MNPCMW6444/cesium/tree/control-characters-in-labels-causes-the-render-to-stop-%2312278-with-full-test
tell me if you still think I need a unit test or any other comment.

packages/engine/Source/Scene/LabelCollection.js
ggetz: I think using a regex is not a bad idea, we should just make sure it only has to happen one time.

Consider moving the `filterUnsupportedCharacters` to the `set` function of `text` in `Label.js`. We already have a bit of logic there to filter and modify the string before assigning it to `_renderedText`.

ggetz:
Hi @MNPCMW6444,

> For the purpose of determing what characters are problematic (turns out not all control chars) and TESTING after the patch I added a full test to sandcastle. but I didn't commit it to PR because I don't think we want it in the sandcastle.

We want to make sure there is some kind of repeatable test in place so we can 1) ensure the code is correct, and 2) keep this bug from happening again if the code ever changes. Most of the time, we use unit tests to validate functionality like this (rather than Sandcastle, as you mentioned) as we can run it as part of our [CI process](https://cesium.com/blog/2024/08/12/ci-for-cesiumjs-github-actions-workflow/).

See the [Testing Guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/TestingGuide/README.md) for more information, but for this particular case, I think we want to:

1. **Architect the code so it can be easily tested**: In this case I would move `filterUnsupportedCharacters` to a static function that can be accessed outside of the `LabelCollection.js` file. If we [move it to `Label.js` like I suggested above](https://github.com/CesiumGS/cesium/pull/12279#discussion_r1843915523), then this would look like:
```
/**
 * Removes control characters, which will cause an error when rendering a glyph.
 * @private
 * @param {string} text The original label text
 * @returns {string} The renderable filtered text
 */
Label.filterUnsupportedCharacters = function (text) {
  ...
}
```
2. **Create unit tests for `filterUnsupportedCharacters`**: Add a new set of unit tests to `LabelCollectionSpec.js`. The sandcastle code you linked to included some great test cases you should be able to re-use! For example:
```
it("filterUnsupportedCharacters removes emoji characters from text", function () {
  const text = "a😀b";
  const expectedText = "ab";
  expect(Label.filterUnsupportedCharacters(text)).toEqual(expectedText);
});
```
3. **Add a label unit test**: Finally, we should add one test to `LabelCollectionSpec.js` to ensure the filtering is happening. Since the tests from step 2 should validate the details of the `filterUnsupportedCharacters` function, this test can be relatively simple, i.e.:
```
it("filters unsupported characters from label text", function () {
  const text = "my\u200btest\u001Dstring😀";
  const expectedText = "myteststring";
  const label = labels.add({
    text: text,
  });

  expect(label.text).toEqual(text);
  expect(label._renderedText).toEqual(expectedText);
});
```

What do you think? Could you give it a try?

packages/engine/Source/Scene/LabelCollection.js
ggetz: Any thoughts on using the unicode ranges to make this more concise? I think the following regex matches of these values:

`/[\u0000-\u001F\u202a-\u206f\u200b-\u200f]/`

MNPCMW6444: got it

packages/engine/Source/Scene/LabelCollection.js
MNPCMW6444: got it

MNPCMW6444:
so i will make the regex simpler with a unicode range and apply it in Label.js, anc create a unit test
one thing im not sure, @ggetz if its a private static method ```Label.filterUnsupportedCharacters = function (text) {``` can the ```set``` of ```text``` access it? how?

ggetz:
> if its a private static method Label.filterUnsupportedCharacters = function (text) { can the set of text access it? how?

@MNPCMW6444 Good question! The `private` tag here affects the generated documentation and generated typescript definitions for the function. It will exclude `Label.filterUnsupportedCharacters` from the API reference doc and the library typescript definitions, with the intention that `Label.filterUnsupportedCharacters` will bot be used directly by end users of the library. 

It can, however, still be used within the codebase itself, such as inside of the `text` `set` function. It can also be used in the unit tests to verify the functionality.

MNPCMW6444:
ok, i pushed the new solution @ggetz 
thanks

ggetz:
Thanks for the updates @MNPCMW6444! This is looking good!

I pushed two small tweaks:
1. Incorporate a previously existing filtering rule into `filterUnsupportedCharacters`
2. Ensure `\t`, `\n`, and `\r` characters were being honored.

MNPCMW6444:
whats the next step @ggetz

MNPCMW6444:
whats the next step @ggetz

ggetz (approved):

Issue #12282, opened 2024-11-01
CZML model disappears on zoom


lukemckinstry:
### What happened?

The CZML model in this sandcastle example disappears when the user zooms. https://sandcastle.cesium.com/?src=CZML%20Model%20-%20Node%20Transformations.html

### Reproduction steps

A. Load sandcastle, zoom immediately, model disappears
B. Load sandcastle, pan and/or rotate, then zoom, model zooms as expected

### Environment

Browser:
CesiumJS Version: Happening since at least 1.122
Operating System:

lukemckinstry:
may be related to https://github.com/CesiumGS/cesium/issues/12231

PR #12284, opened 2024-11-01
Update Viewer timeline when data sources change

jjspace:
# Description

In https://github.com/CesiumGS/cesium/pull/12202 we moved most of the tracking logic into the `CesiumWidget` class. However this skipped the update for the timeline to zoom it to the correct time period when data sources are added, like in this [sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=CZML%20Model%20-%20Node%20Transformations.html)

There may be a cleaner solution here but I think this closely matches the previous behavior. Longer term I would've expected the timeline to reflect the values of the `clock` closer including the start/end times like the Animation widget does.

(We also noticed the model disappears but that's a separate issue https://github.com/CesiumGS/cesium/issues/12282)

## Issue number and link

No issue

## Testing plan

- Open the [CZML Model - Node Transformation](http://localhost:8080/Apps/Sandcastle/index.html?src=CZML%20Model%20-%20Node%20Transformations.html) sandcastle
- Make sure the timeline is zoomed to the correct time frame for the model's availability
- Make sure the timeline head is still looping the full timeline correcly.

# Author checklist

### DIFF
--- a/Apps/Sandcastle/gallery/CZML Model - Node Transformations.html
+++ b/Apps/Sandcastle/gallery/CZML Model - Node Transformations.html
@@ -36,8 +36,8 @@
             name: "CZML Model",
             version: "1.0",
             clock: {
-              interval: "2015-01-01T00:00:00Z/2015-01-01T00:00:20Z",
-              currentTime: "2015-01-01T00:00:00Z",
+              interval: "2015-01-01T12:00:00Z/2015-01-01T12:00:20Z",
+              currentTime: "2015-01-01T12:00:00Z",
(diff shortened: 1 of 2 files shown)


lukemckinstry (approved):
Makes sense, looks like a good fix :+1:

jjspace:
CC @jfayot No action needed but thought you might be interested in this follow up. Something I missed while reviewing your PR. Odd little side effect of the weird way these classes interact with each other and the order event handlers are triggered. If you _do_ think of a better solution than this please let us know.

PR #12289, opened 2024-11-05
iModel integration

jjspace:
# Description

Beginning steps for the iTwin/iModel integration utilizing the [`mesh-export` API](https://developer.bentley.com/apis/mesh-export/overview/)

Opening as a draft for discussion, more details later.

## Issue number and link

no issue number

## Testing plan

- Running the test sandcastle
    - Go to the [`start export` route](https://developer.bentley.com/apis/mesh-export/operations/start-export/) and click "Try it now"
    - Set the auth to Bentley and copy that token into the new sandcastle `iTwin Demo.html`. These keys only last 1 hour so you may have to refresh it
    - Load the [new sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=iTwin%20Demo.html) and make sure it loads. note this may take up to a min of 20ish seconds, I usually see around 7 sec. check the console for actual errors

# Author checklist

### DIFF
--- a/Apps/Sandcastle/gallery/iTwin Demo.html
+++ b/Apps/Sandcastle/gallery/iTwin Demo.html
@@ -1,0 +1,295 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"
+    />
+    <meta
+      name="description"
+      content="Use Viewer to start building new applications or easily embed Cesium into existing applications."
+    />
+    <meta name="cesium-sandcastle-labels" content="Beginner, Showcases" />
+    <title>Cesium Demo</title>
+    <script type="text/javascript" src="../Sandcastle-header.js"></script>
+    <script type="module" src="../load-cesium-es6.js"></script>
+  </head>
+  <body class="sandcastle-loading" data-sandcastle-bucket="bucket-requirejs.html">
+    <style>
+      @import url(../templates/bucket.css);
+    </style>
+    <div id="cesiumContainer" class="fullSize"></div>
+    <div id="loadingOverlay"><h1>Loading...</h1></div>
+    <div id="toolbar">
+      <div id="checkbox"></div>
+      <output id="status">Initializing</output>
+    </div>
+    <script id="cesium_sandcastle_script">
+      window.startup = async function (Cesium) {
+        "use strict";
+        //Sandcastle_Begin
+        // must be created for the Start export route if you want to create new exports
+        // Needs to have the mesh-export:modify scope not just mesh-export:read
+        const accessToken =
+          "Bearer eyJhbGciOiJSUzI1NiIsImtpZCI6IkJlbnRsZXlJTVNfMjAyNCIsInBpLmF0bSI6ImE4bWUifQ.eyJzY29wZSI6WyJtZXNoLWV4cG9ydDpyZWFkIiwibWVzaC1leHBvcnQ6bW9kaWZ5Il0sImNsaWVudF9pZCI6Iml0d2luLWRldmVsb3Blci1jb25zb2xlIiwiYXVkIjpbImh0dHBzOi8vaW1zLmJlbnRsZXkuY29tL2FzL3Rva2VuLm9hdXRoMiIsImh0dHBzOi8vaW1zb2lkYy5iZW50bGV5LmNvbS9hcy90b2tlbi5vYXV0aDIiLCJodHRwczovL2ltc29pZGMuYmVudGxleS5jb20vcmVzb3VyY2VzIiwiYmVudGxleS1hcGktbWFuYWdlbWVudCJdLCJzdWIiOiJjMWM1MzRhNy0zZDk2LTQ2MzMtYjY5ZC1jMGEzNzE5OWQwZGUiLCJyb2xlIjoiQkVOVExFWV9FTVBMT1lFRSIsIm9yZyI6ImZhYjk3NzRiLWIzMzgtNGNjMi1hNmM5LTQ1OGJkZjdmOTY2YSIsInN1YmplY3QiOiJjMWM1MzRhNy0zZDk2LTQ2MzMtYjY5ZC1jMGEzNzE5OWQwZGUiLCJpc3MiOiJodHRwczovL2ltcy5iZW50bGV5LmNvbSIsImVudGl0bGVtZW50IjpbIkJFTlRMRVlfTEVBUk4iLCJJTlRFUk5BTCIsIlNFTEVDVF8yMDA2IiwiQkVOIiwiQkROIl0sInByZWZlcnJlZF91c2VybmFtZSI6Ikpvc2guUm91emVyQGJlbnRsZXkuY29tIiwiZ2l2ZW5fbmFtZSI6Ikpvc2giLCJzaWQiOiJHMGZTamlOXzBMRjE1Vy1IYnRTaWtoeDNuSXMuU1UxVExVSmxiblJzWlhrdFZWTS5LUDVRLjZCakNoYXptMldBMjBIdXVWMUxwNFB6RVAiLCJuYmYiOjE3MzA3NDc3NjUsInVsdGltYXRlX3NpdGUiOiIxMDAxMzg5MTE3IiwidXNhZ2VfY291bnRyeV9pc28iOiJVUyIsImF1dGhfdGltZSI6MTczMDc0ODA2NSwibmFtZSI6Ikpvc2guUm91emVyQGJlbnRsZXkuY29tIiwib3JnX25hbWUiOiJCZW50bGV5IFN5c3RlbXMgSW5jIiwiZmFtaWx5X25hbWUiOiJSb3V6ZXIiLCJlbWFpbCI6Ikpvc2guUm91emVyQGJlbnRsZXkuY29tIiwiZXhwIjoxNzMwNzUxNjY2fQ.0yEQZAyKKAdwNDiwHSD6f_Qzq0M8cbHJcMfT6JidBldw9qiyU4jx6ZdqILddrL-seWCkf9sRtWuoHm7Fw-j_wtaLASaOpHMMwC7IVdh25pbRB-D3mN8_rmQiDbXUadJ1MwH8-pNCubrER1lZLYEPrQ4zJcRtAblbJGNjFdoOi3FXB-y3JLleH4qYykLceDkbW3l2lZRfdIW2pytCIuZs7XZ9Hr6F_cLsYIrs9iRfBFVxxHxxfwgRLQoRuPmDuzKl9-ylLZUFN6CQZUfv7vzL9feoXJXGdZeGgqlyFuMOCiVbYU_elx7P7fFm8G13HvTti5hz98mV1r2bXYvfNpuklg";
+        Cesium.ITwin.defaultAccessToken = accessToken;
+        // this is the iModel in the "Hello iTwinCesium" iTwin that we should all have access to
+        // https://developer.bentley.com/my-itwins/b4a30036-0456-49ea-a439-3fcd9365e24e/home/
+        const imodelId = "2852c3d7-00c3-4b5d-a0ce-82bbde4f061e";
+
+        // const knownExportId = undefined;
+        // const knownExportId = "ab9953b2-bc8e-48ac-a5b0-5d43d68593e8";
+
+        // async function startExport(iModelId, changesetId, accessToken) {
+        //   console.log("Start Export");
+
+        //   const requestOptions = {
+        //     method: "POST",
+        //     headers: {
+        //       Authorization: accessToken,
+        //       Accept: "application/vnd.bentley.itwin-platform.v1+json",
+        //       "Content-Type": "application/json",
+        //     },
+        //     body: JSON.stringify({
+        //       iModelId,
+        //       changesetId,
+        //       exportType: "CESIUM",
+        //     }),
+        //   };
+
+        //   // initiate mesh export
+        //   const response = await fetch(
+        //     `https://api.bentley.com/mesh-export/`,
+        //     requestOptions,
+        //   );
+        //   if (!response.ok) {
+        //     if (response.status === 401) {
+        //       console.error("Unauthorized, bad token, wrong scopes or headers bad");
+        //     } else if (response.status === 403) {
+        //       console.error("Not allowed, forbidden");
+        //     } else if (response.status === 422) {
+        //       console.error("Unprocessable: Cannot create export job");
+        //     } else if (response.status === 429) {
+        //       console.log("Too many requests");
+        //     } else {
+        //       console.error("Bad request, unknown error", response);
+        //     }
+        //     return undefined;
+        //   }
+        //   const result = JSON.parse(JSON.stringify(await response.json()));
+        //   return result?.export?.id;
+        // }
+
+        // async function getExport(exportId, accessToken) {
+        //   const headers = {
+        //     Authorization: accessToken,
+        //     Accept: "application/vnd.bentley.itwin-platform.v1+json",
+        //   };
+
+        //   // obtain export for specified export id
+        //   const url = `https://api.bentley.com/mesh-export/${exportId}`;
+        //   try {
+        //     // TODO: this request is _really_ slow, like 7 whole second alone for me
+        //     const response = await fetch(url, { headers });
+        //     if (!response.ok) {
+        //       if (response.status === 401) {
+        //         console.error("Unauthorized, bad token, wrong scopes or headers bad");
+        //       } else if (response.status === 404) {
+        //         console.error("Requested export is not available", exportId);
+        //       } else if (response.status === 429) {
+        //         console.error("Too many requests");
+        //       } else {
+        //         console.log("Unknown request failure", response);
+        //       }
+        //       return undefined;
+        //     }
+        //     const result = JSON.parse(JSON.stringify(await response.json()));
+        //     return result;
+        //   } catch (err) {
+        //     return undefined;
+        //   }
+        // }
+
+        const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+
+        // Grabbed mapping from the iTwin Viewer
+        const classes = {
+          2199023255632: "Building Roof",
+          2199023255694: "Building Wall",
+          2199023255696: "Building Window",
+        };
+
+        let selectedFeature;
+        let picking = false;
+
+        Sandcastle.addToggleButton(
+          "Per-feature selection",
+          false,
+          function (checked) {
+            picking = checked;
+            if (!picking) {
+              unselectFeature(selectedFeature);
+            }
+          },
+          "checkbox",
+        );
+
+        // Set up viewer
+        const viewer = new Cesium.Viewer("cesiumContainer", {
+          terrain: Cesium.Terrain.fromWorldTerrain(),
+        });
+        const scene = viewer.scene;
+        scene.globe.show = true;
+        scene.debugShowFramesPerSecond = true;
+
+        // HTML overlay for showing feature name on mouseover
+        const nameOverlay = document.createElement("div");
+        viewer.container.appendChild(nameOverlay);
+        nameOverlay.className = "backdrop";
+        nameOverlay.style.display = "none";
+        nameOverlay.style.position = "absolute";
+        nameOverlay.style.bottom = "0";
+        nameOverlay.style.left = "0";
+        nameOverlay.style["pointer-events"] = "none";
+        nameOverlay.style.padding = "4px";
+        nameOverlay.style.backgroundColor = "black";
+        nameOverlay.style.whiteSpace = "pre-line";
+        nameOverlay.style.fontSize = "12px";
+
+        function selectFeature(feature, movement) {
+          feature.color = Cesium.Color.clone(
+            Cesium.Color.fromCssColorString("#eeff41"),
+            feature.color,
+          );
+          selectedFeature = feature;
+
+          nameOverlay.style.display = "block";
+          nameOverlay.style.bottom = `${
+            viewer.canvas.clientHeight - movement.endPosition.y
+          }px`;
+          nameOverlay.style.left = `${movement.endPosition.x}px`;
+          const element = feature.getProperty("element");
+          const subcategory = feature.getProperty("subcategory");
+          const message = `
+      Element ID: ${element}
+      Subcategory: ${classes[subcategory] ?? subcategory}
+      Feature ID: ${feature.featureId}`;
+          nameOverlay.textContent = message;
+        }
+
+        function unselectFeature(feature) {
+          if (!Cesium.defined(feature)) {
+            return;
+          }
+
+          feature.color = Cesium.Color.clone(Cesium.Color.WHITE, feature.color);
+          selectedFeature = undefined;
+          nameOverlay.style.display = "none";
+        }
+
+        const statusOutput = document.querySelector("#status");
+        async function init() {
+          // TODO: just for testing
+          // await Cesium.createIModel3DTileset.deleteExport(
+          //   "3a627319-cf6e-4535-9499-da5320a69791",
+          // );
+          // console.log(await Cesium.createIModel3DTileset.checkForCesiumExport(imodelId));
+          // console.log(await Cesium.createIModel3DTileset.getExports(imodelId));
+
+          const changesetId = "";
+          const cesiumExport = await Cesium.createIModel3DTileset.checkForCesiumExport(
+            imodelId,
+            changesetId,
+          );
+          let exportId = cesiumExport?.id;
+          if (!Cesium.defined(cesiumExport)) {
+            exportId = await Cesium.createIModel3DTileset.createExportForModelId(
+              imodelId,
+              changesetId,
+              accessToken,
+            );
+          }
+
+          statusOutput.innerText = "Starting export";
+          // const exportId =
+          //   knownExportId ??
+          //   (await Cesium.createIModel3DTileset.createExportForModelId(
+          //     imodelId,
+          //     "",
+          //     accessToken,
+          //   ));
+
+          if (!exportId) {
+            console.error("No export id returned");
+            return;
+          }
+          console.log("Using export id", exportId);
+
+          const start = Date.now();
+          // let result = await getExport(exportId, accessToken);
+          // let status = result.export.status;
+          // while (status !== "Complete") {
+          //   await delay(5000);
+          //   result = await getExport(exportId, accessToken);
+          //   status = result.export.status;
+          //   console.log(`Export is ${status}`);
+
+          //   if (Date.now() - start > 300000) {
+          //     throw new Error("Export did not complete in time.");
+          //   }
+          // }
+
+          // if (result.export.request.exportType !== "CESIUM") {
+          //   console.error("Wrong export type", result.export.request.exportType);
+          //   throw new Error(`Wrong export type ${result.export.request.exportType}`);
+          // }
+
+          // // This link is only valid 1 hour
+          // let tilesetUrl = result.export._links.mesh.href;
+          // const splitStr = tilesetUrl.split("?");
+          // // is there a cleaner way to do this?
+          // tilesetUrl = `${splitStr[0]}/tileset.json?${splitStr[1]}`;
+
+          // const tileset = await Cesium.Cesium3DTileset.fromUrl(tilesetUrl);
+          statusOutput.innerText = "Creating Tileset";
+          const tileset = await Cesium.createIModel3DTileset(exportId);
+          scene.primitives.add(tileset);
+          tileset.colorBlendMode = Cesium.Cesium3DTileColorBlendMode.REPLACE;
+
+          statusOutput.innerText = "Loaded";
+
+          viewer.zoomTo(tileset);
+          console.log(`Finished in ${((Date.now() - start) / 1000).toString()} seconds`);
+
+          const handler = new Cesium.ScreenSpaceEventHandler(scene.canvas);
+          handler.setInputAction(function (movement) {
+            if (!picking) {
+              return;
+            }
+            unselectFeature(selectedFeature);
+
+            const feature = scene.pick(movement.endPosition);
+
+            if (feature instanceof Cesium.Cesium3DTileFeature) {
+              selectFeature(feature, movement);
+            }
+          }, Cesium.ScreenSpaceEventType.MOUSE_MOVE);
+        }
+
+        init();
+
+        //Sandcastle_End
+        Sandcastle.finishedLoading();
+      };
+      if (typeof Cesium !== "undefined") {
+        window.startupCalled = true;
+        window.startup(Cesium).catch((error) => {
+          "use strict";
+          console.error(error);
+        });
+      }
+    </script>
+  </body>
+</html>
--- a/packages/engine/Source/Core/ITwin.js
+++ b/packages/engine/Source/Core/ITwin.js
@@ -1,0 +1,66 @@
+import Resource from "./Resource.js";
+
+/**
+ * Default settings for accessing the iTwin platform.
+ *
+ * Keys can be created using the iModels share routes {@link https://developer.bentley.com/apis/imodels-v2/operations/create-imodel-share/}
+ *
+ * An ion access token is only required if you are using any ion related APIs.
+ * A default access token is provided for evaluation purposes only.
+ * Sign up for a free ion account and get your own access token at {@link https://cesium.com}
+ *
+ * @see IonResource
+ * @see IonImageryProvider
+ * @see IonGeocoderService
+ * @see createWorldImagery
+ * @see createWorldTerrain
+ * @namespace ITwin
+ */
+const ITwin = {};
+
+/**
+ * Gets or sets the default iTwin access token.
+ *
+ * TODO: I'm not sure we can even do this kind of access token. Each route seems to need it's own scopes
+ * and we may not be able to guarantee this "top level token" has them all
+ * So far we use
+ * `mesh-export:read` for loading meshes GET /mesh-export(s)
+ * `mesh-export:modify` if we want to include a function to create an export
+ * `itwin-platform` if we want to use the iModel shares ourselves  GET /imodels/{id}/shares
+ *
+ *
+ * @type {string|undefined}
+ */
+ITwin.defaultAccessToken = undefined;
+
+/**
+ * Gets or sets the default Google Map Tiles API endpoint.
+ *
+ * @type {string|Resource}
+ * @default https://api.bentley.com
+ */
+ITwin.apiEndpoint = new Resource({
+  url: "https://api.bentley.com",
+});
+
+// TODO: this should only be needed if we have a way to generate really long term access tokens
+// to sample data that is accessible to everyone
+// ITwin.getDefaultTokenCredit = function (providedKey) {
+//   if (providedKey !== defaultAccessToken) {
+//     return undefined;
+//   }
+
+//   if (!defined(defaultTokenCredit)) {
+//     const defaultTokenMessage =
+//       '<b> \
+//             This application is using Cesium\'s default ion access token. Please assign <i>Cesium.Ion.defaultAccessToken</i> \
+//             with an access token from your ion account before making any Cesium API calls. \
+//             You can sign up for a free ion account at <a href="https://cesium.com">https://cesium.com</a>.</b>';
+
+//     defaultTokenCredit = new Credit(defaultTokenMessage, true);
+//   }
+
+//   return defaultTokenCredit;
+// };
+
+export default ITwin;
--- a/packages/engine/Source/Scene/createIModel3DTileset.js
+++ b/packages/engine/Source/Scene/createIModel3DTileset.js
@@ -1,0 +1,371 @@
+import Cesium3DTileset from "./Cesium3DTileset.js";
+import defined from "../Core/defined.js";
+import Resource from "../Core/Resource.js";
+import ITwin from "../Core/ITwin.js";
+import DeveloperError from "../Core/DeveloperError.js";
+
+function delay(ms) {
+  return new Promise((res) => setTimeout(res, ms));
+}
+
+/**
+ * @enum {string}
+ */
+const ExportStatus = Object.freeze({
+  NotStarted: "NotStarted",
+  InProgress: "InProgress",
+  Complete: "Complete",
+  Invalid: "Invalid",
+});
+
+/**
+ * Type of an export currently, only GLTF and 3DFT are documented
+ * The CESIUM option is what we were told to use with Sandcastle
+ * I've also seen the IMODEL one but don't know where it's from
+ * @enum {string}
+ */
+const ExportType = Object.freeze({
+  "3DFT": "3DFT",
+  GLFT: "GLTF",
+  IMODEL: "IMODEL",
+  CESIUM: "CESIUM",
+});
+
+/**
+ * @typedef {Object} GeometryOptions
+ * @property {boolean} includeLines
+ * @property {number} chordTol
+ * @property {number} angleTol
+ * @property {number} decimationTol
+ * @property {number} maxEdgeLength
+ * @property {number} minBRepFeatureSize
+ * @property {number} minLineStyleComponentSize
+ */
+
+/**
+ * @typedef {Object} ViewDefinitionFilter
+ * @property {string[]} models Array of included model IDs.
+ * @property {string[]} categories Array of included category IDs.
+ * @property {string[]} neverDrawn Array of element IDs to filter out.
+ */
+
+/**
+ * @typedef {Object} StartExport
+ * @property {string} iModelId
+ * @property {string} changesetId
+ * @property {ExportType} exportType Type of mesh to create. Currently, only GLTF and 3DFT are supported and undocumented CESIUM option
+ * @property {GeometryOptions} geometryOptions
+ * @property {ViewDefinitionFilter} viewDefinitionFilter
+ */
+
+/**
+ * @typedef {Object} Link
+ * @property {string} href
+ */
+
+/**
+ * @typedef {Object} Export
+ * @property {string} id
+ * @property {string} displayName
+ * @property {ExportStatus} status
+ * @property {StartExport} request
+ * @property {{mesh: Link}} _links
+ */
+
+/**
+ * @typedef {Object} ExportResponse
+ * @property {Export} export
+ */
+
+/**
+ * Creates a {@link Cesium3DTileset} instance for the Google Photorealistic 3D Tiles tileset.
+ *
+ * @function
+ *
+ * @param {string} exportId
+ * @param {Cesium3DTileset.ConstructorOptions} [options] An object describing initialization options.
+ * @returns {Promise<Cesium3DTileset>}
+ *
+ * @see ITwin
+ *
+ * @example
+ * // Use your own iTwin API key for mesh export
+ * Cesium.ITwin.defaultApiKey = "your-api-key";
+ *
+ * const viewer = new Cesium.Viewer("cesiumContainer");
+ *
+ * try {
+ *   const tileset = await Cesium.createIModel3DTileset();
+ *   viewer.scene.primitives.add(tileset));
+ * } catch (error) {
+ *   console.log(`Error creating tileset: ${error}`);
+ * }
+ */
+async function createIModel3DTileset(exportId, options) {
+  if (!defined(ITwin.defaultAccessToken)) {
+    throw new DeveloperError("Must set ITwin.defaultAccessToken first");
+  }
+
+  options = options ?? {};
+
+  const timeoutAfter = 300000;
+  const start = Date.now();
+  let result = await createIModel3DTileset.getExport(exportId);
+  let status = result.export.status;
+
+  if (result.export.request.exportType !== ExportType.CESIUM) {
+    // This is an undocumented value but I think it's the only one we want to load
+    // TODO: should we even be checking this?
+    throw new Error(`Wrong export type ${result.export.request.exportType}`);
+  }
+
+  // wait until the export is complete
+  while (status !== ExportStatus.Complete) {
+    await delay(5000);
+    result = await createIModel3DTileset.getExport(exportId);
+    status = result.export.status;
+    console.log(`Export is ${status}`);
+
+    if (Date.now() - start > timeoutAfter) {
+      throw new Error("Export did not complete in time.");
+    }
+  }
+
+  // This link is only valid 1 hour
+  let tilesetUrl = result.export._links.mesh.href;
+  const splitStr = tilesetUrl.split("?");
+  // is there a cleaner way to do this?
+  tilesetUrl = `${splitStr[0]}/tileset.json?${splitStr[1]}`;
+
+  const resource = new Resource({
+    url: tilesetUrl,
+  });
+
+  return Cesium3DTileset.fromUrl(resource, options);
+}
+
+/**
+ * @param {string} exportId
+ */
+createIModel3DTileset.getExport = async function (exportId) {
+  const headers = {
+    Authorization: ITwin.defaultAccessToken,
+    Accept: "application/vnd.bentley.itwin-platform.v1+json",
+  };
+
+  // obtain export for specified export id
+  const url = `${ITwin.apiEndpoint}mesh-export/${exportId}`;
+
+  // TODO: this request is _really_ slow, like 7 whole second alone for me
+  // Arun said this was kinda normal but to keep track of the `x-correlation-id` of any that take EXTRA long
+  const response = await fetch(url, { headers });
+  if (!response.ok) {
+    const result = await response.json();
+    if (response.status === 401) {
+      throw new Error(
+        `Unauthorized, bad token, wrong scopes or headers bad. ${result.error.details[0].code}`,
+      );
+    } else if (response.status === 404) {
+      throw new Error(`Requested export is not available ${exportId}`);
+    } else if (response.status === 429) {
+      throw new Error("Too many requests");
+    }
+    throw new Error(`Unknown request failure ${response.status}`);
+  }
+
+  /** @type {ExportResponse} */
+  const result = await response.json();
+  return result;
+};
+
+/**
+ * Get the list of exports for the given iModel + changeset
+ *
+ * @param {string} iModelId
+ * @param {string} changesetId
+ */
+createIModel3DTileset.getExports = async function (iModelId, changesetId) {
+  if (!defined(ITwin.defaultAccessToken)) {
+    throw new DeveloperError("Must set ITwin.defaultAccessToken first");
+  }
+  const headers = {
+    Authorization: ITwin.defaultAccessToken,
+    Accept: "application/vnd.bentley.itwin-platform.v1+json",
+    Prefer: "return=representation", // or return=minimal (the default)
+  };
+
+  // obtain export for specified export id
+  let url = `${ITwin.apiEndpoint}mesh-export/?iModelId=${iModelId}`;
+  if (defined(changesetId) && changesetId !== "") {
+    url += `&changesetId=${changesetId}`;
+  }
+
+  const response = await fetch(url, { headers });
+  if (!response.ok) {
+    const result = await response.json();
+    if (response.status === 401) {
+      throw new Error(
+        `Unauthorized, bad token, wrong scopes or headers bad. ${result.error.details[0].code}`,
+      );
+    } else if (response.status === 422) {
+      throw new Error(
+        `Unprocessable Entity:${result.error.code} ${result.error.message}`,
+      );
+    } else if (response.status === 429) {
+      throw new Error("Too many requests");
+    }
+    throw new Error(`Unknown request failure ${response.status}`);
+  }
+
+  /** @type {{exports: Export[]}} */
+  const result = await response.json();
+  return result;
+};
+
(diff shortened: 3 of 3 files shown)


jjspace:
@ggetz just pushed an update to restructure the api a bit. I moved the itwin API functions under the `ITwin` namespace. I also split the tileset functions into 2 functions. One loads a tileset from only the export id and the other takes a model id + changeset combo and tries to find and load a `CESIUM` export if possible.

Looking at the API more it seems exports can be created for the same "model id + changeset" with different export parameters which _could_ mean multiple CESIUM type exports for the same combo. I think this is probably less likely and when it's needed then users can use the load function with the export id directly and manage their own handling of the list.

pjcozzi:
@aruniverse would you or another iTwin colleague be able to take a look at this?

aruniverse:
> @aruniverse would you or another iTwin colleague be able to take a look at this?

Thanks for sharing, this is exciting to see! I'll take a pass over it. 

fyi @danieliborra

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,211 @@
...
+    <div id="toolbar">
+      <div id="checkbox"></div>
+      <output id="status">Initializing</output>
+    </div>
+    <script id="cesium_sandcastle_script">
+      window.startup = async function (Cesium) {
+        "use strict";
+        //Sandcastle_Begin
+        // must be created for the Start export route if you want to create new exports
+        // Needs to have the mesh-export:modify scope not just mesh-export:read
```
aruniverse, line 34: We recently consolidated all scopes into a single one, "itwin-platform". That should be the recommended scope to use. Please see https://developer.bentley.com/itwin-platform-scope-introduction/
The "try it out" feature in the developer portal is just still using granular scopes.

```suggestion
        // Needs to have the `itwin-platform` scope
```

fyi @dsmailys

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+
+/**
+ * Type of an export currently, only GLTF and 3DFT are documented
+ * The CESIUM option is what we were told to use with Sandcastle
+ * I've also seen the IMODEL one but don't know where it's from
+ * @enum {string}
+ */
+const ExportType = Object.freeze({
+  "3DFT": "3DFT",
+  GLFT: "GLTF",
```
aruniverse, line 24: nit

```suggestion
  GLTF: "GLTF",
```

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+
+/**
+ * @typedef {Object} ExportResponse
+ * @property {Export} export
+ */
+
+/**
+ * Default settings for accessing the iTwin platform.
+ *
+ * Keys can be created using the iModels share routes {@link https://developer.bentley.com/apis/imodels-v2/operations/create-imodel-share/}
```
aruniverse, line 78: Please note, this API has the following note:
> Important : Share operations are in closed preview mode currently. Hence only selected applications can utilize the Share API.

fyi @robertas-kerpys, we will need to work with Cesium team here

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+ * @see createWorldImagery
+ * @see createWorldTerrain
+ * @namespace ITwin
+ */
+const ITwin = {};
+
+/**
+ * Gets or sets the default iTwin access token.
+ *
+ * TODO: I'm not sure we can even do this kind of access token. Each route seems to need it's own scopes
```
aruniverse, line 96: this should be ok now that we've consolidated all scopes to "itwin-platform"

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+ * @param {string} exportId
+ */
+ITwin.getExport = async function (exportId) {
+  if (!defined(ITwin.defaultAccessToken)) {
+    throw new DeveloperError("Must set ITwin.defaultAccessToken first");
+  }
+
+  const headers = {
+    Authorization: ITwin.defaultAccessToken,
+    Accept: "application/vnd.bentley.itwin-platform.v1+json",
```
aruniverse, line 128: This is fine for this specific use case where youre using just a single API (mesh-export), but might prove to be an issue when you are mixing APIs(iTwins, iModels, etc)
We use request headers to denote which verison of an API to use. So far, most APIs are still on v1, but there are some for example the iModels APIs which are on v2, and the the `accept` header would be need to be `application/vnd.bentley.itwin-platform.v2+json`

Please see [API Version Policy](https://developer.bentley.com/apis/overview/lifecycle/#api-versions)

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+  };
+
+  // obtain export for specified export id
+  const url = `${ITwin.apiEndpoint}mesh-export/${exportId}`;
+
+  // TODO: this request is _really_ slow, like 7 whole second alone for me
```
aruniverse, line 134: fyi @danieliborra @arvindsvenkat

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+    body: JSON.stringify({
+      iModelId,
+      changesetId,
+      exportType: "CESIUM",
+    }),
+  };
+
+  // initiate mesh export
+  const response = await fetch(
+    `https://api.bentley.com/mesh-export/`,
```
aruniverse, line 230: nit: 

```suggestion
    `${ITwin.apiEndpoint}mesh-export/`
```

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+  /** @type {ExportResponse} */
+  const result = await response.json();
+  return result;
+};
+
+/**
+ * Get the list of exports for the given iModel + changeset
+ *
+ * @param {string} iModelId
+ * @param {string} changesetId
```
aruniverse, line 160: this could be undefined

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+  /** @type {{exports: Export[]}} */
+  const result = await response.json();
+  return result;
+};
+
+/**
+ * Start the export process for the given iModel + changeset.
+ *
+ * @param {string} iModelId
+ * @param {string} changesetId
```
aruniverse, line 205: could also be undefined here as well

packages/engine/Source/Scene/createIModel3DTileset.js
```diff
@@ -0,0 +1,127 @@
...
+    console.log(`Export is ${status}`);
+
+    if (Date.now() - start > timeoutAfter) {
+      throw new RuntimeError("Export did not complete in time.");
+    }
+  }
+
+  // This link is only valid 1 hour
+  let tilesetUrl = exportObj._links.mesh.href;
+  const splitStr = tilesetUrl.split("?");
```
aruniverse, line 41: Hmm, im not sure if you need to do all this wth splitting the link url. 

This is how I do it in my test app

```ts
const baseUrl = new URL(meshExportUrl.export._links.mesh.href);
baseUrl.pathname = baseUrl.pathname + "/tileset.json";
tilesetUrl = baseUrl.toString()
```

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,211 @@
...
...hNy0zZDk2LTQ2MzMtYjY5ZC1jMGEzNzE5OWQwZGUiLCJyb2xlIjoiQkVOVExFWV9FTVBMT1lFRSIsIm9yZyI6ImZhYjk3NzRiLWIzMzgtNGNjMi1hNmM5LTQ1OGJkZjdmOTY2YSIsInN1YmplY3QiOiJjMWM1MzRhNy0zZDk2LTQ2MzMtYjY5ZC1jMGEzNzE5OWQwZGUiLCJpc3MiOiJodHRwczovL2ltcy5iZW50bGV5LmNvbSIsImVudGl0bGVtZW50IjpbIkJFTlRMRVlfTEVBUk4iLCJJTlRFUk5BTCIsIlNFTEVDVF8yMDA2IiwiQkVOIiwiQkROIl0sInByZWZlcnJlZF91c2VybmFtZSI6Ikpvc2guUm91emVyQGJlbnRsZXkuY29tIiwiZ2l2ZW5fbmFtZSI6Ikpvc2giLCJzaWQiOiJHMGZTamlOXzBMRjE1Vy1IYnRTaWtoeDNuSXMuU1UxVExVSmxiblJzWlhrdFZWTS5LbWlWLlF2UExVcGJyc2R3engwWGxPbkF3eTFTT0IiLCJuYmYiOjE3MzA4MzgyNDIsInVsdGltYXRlX3NpdGUiOiIxMDAxMzg5MTE3IiwidXNhZ2VfY291bnRyeV9pc28iOiJVUyIsImF1dGhfdGltZSI6MTczMDgzODU0MiwibmFtZSI6Ikpvc2guUm91emVyQGJlbnRsZXkuY29tIiwib3JnX25hbWUiOiJCZW50bGV5IFN5c3RlbXMgSW5jIiwiZmFtaWx5X25hbWUiOiJSb3V6ZXIiLCJlbWFpbCI6Ikpvc2guUm91emVyQGJlbnRsZXkuY29tIiwiZXhwIjoxNzMwODQyMTQyfQ.N_WrgjL2bqxdNLEM5nHh4Fg-FzeA-qxxpryaoaMKz8onnpgzmp-X8dyQ1TyMRqKQY99iLypE9bU45DwRJs7vBgNQ73d53SkC9TKGYn8AAOTJz8c_oEHgkoTEDaaLsMPCw88tqZ34pY0e0oIHIofVDTCvwlzwaJPADfkIxz-8GzhIt2WrXR7f6LWBFSlWNrtNhIr9Tz7UNaLOh97_3fS9KVU1I084CpBga9cj_mjGBeki7mIEQvpqMj8x2bJPae_c6WrPEjKLayrOzq4q0X0Kvle0ZFAm-Se9MFCusTFS51bYrI3IsjagGeIP2U06zzcMJ22NylomE60hz6GK_4uB3g";
```
aruniverse, line 36: Would recommend not to commit an access token in text like this, but these are short lived (1 hr) so its probably expired now

aruniverse (commented):
Did a quick first pass, Josh this looks pretty good so far. I will look at trying to run this locally tomorrow. 

Added some minor nit comments and notes. I've tagged some of the other leads from iTwin Platform working on the services you are using here for visibility.

pjcozzi:
>>@aruniverse would you or another iTwin colleague be able to take a look at this?
>
>Thanks for sharing, this is exciting to see! I'll take a pass over it.

Appreciate the fast turn, thanks @aruniverse.

packages/engine/Source/Core/ITwin.js
jjspace: I was going off the sample code provided to us and the [documentation](https://developer.bentley.com/apis/mesh-export/operations/get-export/#request-headers) which states for every route in the mesh export API that:

> Setting to `application/vnd.bentley.itwin-platform.v1+json` is recommended

If the docs are wrong then they should be updated. On top of that I'm currently setting this per request so different routes could definitely use different values if needed for different versions as you say. As a consumer of the API I would fully expect all the routes on the same API to still work with each other if versioning is done this way.

packages/engine/Source/Core/ITwin.js
jjspace: I was occasionally seeing this jump to ~20-23 seconds. It seemed very repeatable that I got one of those two timings for the [Get export](https://developer.bentley.com/apis/mesh-export/operations/get-export/) route specifically. Interestingly the [Get exports](https://developer.bentley.com/apis/mesh-export/operations/get-exports/) route for the full list is really fast and almost always under 1 second, usually under 200-300ms too even with `return=representation`.

packages/engine/Source/Core/ITwin.js
jjspace: I don't think we plan to use the model share api here at all, I don't think it gives us what we need. This was an early comment when I thought we would be.

We definitely will need a good strategy for auth in general though. it's one of the biggest open questions we have right now. I think that discussion should live outside this specific thread.

jjspace:
> Added some minor nit comments and notes.

@aruniverse thank you for the thorough comments. This is still a Draft PR and very much a work in progress and nothing is final. I agree with most of your comments but it was not ready for such scrutiny :sweat_smile: I opened the PR to work through the general process and API structure that we'd need. A lot of the code is still very much thrown together to prove out the concept

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+ * The CESIUM option is what we were told to use with Sandcastle
+ * I've also seen the IMODEL one but don't know where it's from
+ * @enum {string}
+ */
+const ExportType = Object.freeze({
+  "3DFT": "3DFT",
+  GLFT: "GLTF",
+  IMODEL: "IMODEL",
+  CESIUM: "CESIUM",
+});
```
danieliborra: We're currently working on the `mesh-export` API, and we changed this quite a bit. I updated the documentation on Monday, but it will take some time to be deployed.
3DFT and GLTF are now deprecated, IMODEL is for internal use only (until we deprecate it) and now you can use "3DTILES", which is what I would recommend to use.

packages/engine/Source/Core/ITwin.js
danieliborra: For demo purposes, we're working right now on adding shared keys to the mesh export service. I expect to have it deployed in the next 4 weeks, if all goes as planned.

packages/engine/Source/Core/ITwin.js
danieliborra: That is because `get-export` is not very used and so we haven't optimized it yet. All apps use `get-exports`, and we have kept `get-export` to avoid breaking changes.
This quarter we will clean-up the API.

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  const headers = {
+    Authorization: ITwin.defaultAccessToken,
+    Accept: "application/vnd.bentley.itwin-platform.v1+json",
+    Prefer: "return=representation", // or return=minimal (the default)
+  };
+
+  // obtain export for specified export id
+  let url = `${ITwin.apiEndpoint}mesh-export/?iModelId=${iModelId}`;
```
danieliborra, line 164: I'd recommend to use `&exportType=3DTILES` to filter out other export types, and change the exportType enum above to 3DTILES too.
Also, you can do `&$top=5` to get the latest 5 exports, for performance reasons.

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+  return result;
+};
+
+/**
+ * Start the export process for the given iModel + changeset.
+ *
+ * @param {string} iModelId
+ * @param {string} changesetId
+ */
+ITwin.createExportForModelId = async function (iModelId, changesetId) {
```
danieliborra, line 197: Today, we're auto-generating tiles for all iModels but for the exportType=IMODEL.
We plan to also auto-generate "3DTILES". Development is on-going, we expect to have it end of this quarter.
We also plan to deprecate this endpoint when the auto-generation is in place.

packages/engine/Source/Core/ITwin.js
```diff
@@ -0,0 +1,306 @@
...
+
+  /** @type {ExportResponse} */
+  const result = await response.json();
+  return result.export.id;
+};
+
+/**
+ * Delete the specified export
+ *
+ * TODO: I'm not sure if we want this or not. Might belong better as an APP level function
```
danieliborra, line 257: You won't need it, but it is a good exercise to test the API

packages/engine/Source/Scene/createIModel3DTileset.js
```diff
@@ -0,0 +1,127 @@
...
+ * multiple exports per "iModel id + changeset id" you should switch to using {@link createIModel3DTileset}
+ * with the export id directly
+ *
+ * @example
+ * // Try to load the corresponding tileset export or create it if it doesn't exist
+ * let tileset = await Cesium.createIModel3DTileset.fromModelIdimodelId, changesetId);
+ * if (!Cesium.defined(tileset)) {
+ *   const exportId = await Cesium.ITwin.createExportForModelId(imodelId, changesetId, accessToken);
+ *   tileset = await Cesium.createIModel3DTileset(exportId);
+ * }
```
danieliborra, line 106: We're doing it right not for demo purposes, but that will go away in the future.

danieliborra (commented):
It looks promising.
We're now modifying the API, so we will keep you updated of any change in the API.

packages/engine/Source/Core/ITwin.js
aruniverse: Yes this will work, the entire mesh-export service apis are still on v1 so the accept header here is right.

My apologies, I jumped the gun when I had initially commented this thinking you were going to use multiple api routes in a single function.

packages/engine/Source/Core/ITwin.js
jjspace: @danieliborra I was hoping for a way to filter based on type but none are [documented](https://developer.bentley.com/apis/mesh-export/operations/get-exports/) so I didn't think it was possible. Are there other properties filterable like this?

Apps/Sandcastle/gallery/iTwin Demo.html
dsmailys: Hey, thanks for noticing and reporting this. I've let the leadDeveloper of mesh-export know about this inconsistency and will hopefully get resolved. itwin-platform scope will still work/be accepted for mesh-export.

dsmailys: > The "try it out" feature in the developer portal is just still using granular scopes.

Reported this back to the developer portal team. Thanks for bringing this up.

packages/engine/Source/Core/ITwin.js
danieliborra: You are right. Since the API is in beta/tech. preview but products/users are still using it, we have added some experimental parameters that are not documented because we planed to remove them in the future (this quarter). We didn't want users to implement them and then introduce breaking changes.
We're currently updating the API documentation and cleaning some of those parameter, but it will still take a few days to be deployed in PROD.

packages/engine/Source/Core/ITwin.js
ggetz: Hi @danieliborra! We're targeting these changes for the end of November of so that we can (ideally) ship them with the December 1 CesiumJS release. Given that, what is the recommended approach?

packages/engine/Source/Core/ITwin.js
ggetz: @danieliborra In the meantime, would using `get-exports` be preferred then?

packages/engine/Source/Core/ITwin.js
ggetz: @danieliborra We plan on marking all of these new APIs as "experimental" so we have some more flexibility with breaking changes. So we can remove this function when recommended.

packages/engine/Source/Core/ITwin.js
ggetz: We'll likely want to pair the API here down to the minimum necessary for visualization– Managing exports and other function is something that would not be unique to CesiumJS, so [to reduce the maintenance burden](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#design) we can omit it.

packages/engine/Source/Scene/createIModel3DTileset.js
ggetz: @danieliborra Is this referring to the `CESIUM` export type? Or something else?

danieliborra: Generating exports manually will go away, as we're implementing right now the auto-exporting for all formats, including CESIUM.

jjspace:
Given that the iTwin platform currently has no way to create long running API keys I spent some time looking into the oauth workflows. The main goal right now is to be able to create one or more sandcastles that can demonstrate using iModels in CesiumJS _without_ requiring a login from the users. @ggetz had the idea to use the ion servers for oauth and return a token that can be used in Sandcastle.

My latest commit includes a demo server in `itwin-oauth-demo` to act in place of ion to demonstrate using the iTwin [Service App](https://developer.bentley.com/apis/overview/authorization/service-app/) and iTwin [Web App](https://developer.bentley.com/apis/overview/authorization/web-app/) auth processes. Run with `cd itwin-oauth-demo && node server.js`. Ask me for the config if you wanna use my app keys.
The Web App process requires the user to log in but then gives access to all of that user's iTwins/iModels. The Service App process does not require a login but the "service app user" needs to be added to a specific iTwin and given permissions to do things with that iTwin/iModels. The [Projects page](https://connect.bentley.com/SelectProject/Index) can be used to view iTwin projects and manage users/roles.

The only way I found to add the app user was through the [Access Control](https://developer.bentley.com/apis/access-control/operations/add-iTwin-members/) API which needs you to create a role first for the role id. I found the sandcastle was able to load data using a service app token with these permissions on it's role:

The roles don't perfectly match with the previous API scopes and there's none that describe the model exports so I'm not 100% sure which role permissions we'll need but this was just to prove out the concept and it is working.

The demo sandcastle has been updated to use the new local auth server. If you want to try the Web App flow that requires a login just uncomment/comment the code near the top and reload the sandcastle then log in. You will need access to the iModel you're trying to load.

packages/engine/Source/Scene/createIModel3DTileset.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {Export} exportObj
+ * @param {Cesium3DTileset.ConstructorOptions} [options]
+ */
+async function loadExport(exportObj, options) {
+  let status = exportObj.status;
+
+  if (exportObj.request.exportType !== ExportType["3DTILES"]) {
+    // This is an undocumented value but I think it's the only one we want to load
+    // TODO: should we even be checking this?
+    throw new Error(`Wrong export type ${exportObj.request.exportType}`);
```
ggetz, line 22:
```suggestion
    throw new RuntimeError(`Wrong export type ${exportObj.request.exportType}`);
```

packages/engine/Source/Scene/createIModel3DTileset.js
```diff
@@ -0,0 +1,115 @@
...
+  tilesetUrl = `${splitStr[0]}/tileset.json?${splitStr[1]}`;
+
+  const resource = new Resource({
+    url: tilesetUrl,
+  });
+
+  return Cesium3DTileset.fromUrl(resource, options);
+}
+
+const createIModel3DTileset = {};
```
ggetz, line 52: If we're going this route, make sure this naming is updated to reflect that it is a namespace and not a function:
 * Namespaces are capitalized
 * Namespaces usually don't have verbs like `create`

ggetz (changes requested):
Thanks @jjspace!

I left a few comments for the things that jumped out to me.

However, I understand that some of this is in flux. I believe at this point we have enough info to settle on the public API. Please do a pass on that and the docs, then ping me when ready for another a review pass.

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,188 @@
...
+    <div id="toolbar">
+      <div id="checkbox"></div>
+      <output id="status">Initializing</output>
+    </div>
+    <script id="cesium_sandcastle_script">
+      window.startup = async function (Cesium) {
+        "use strict";
+        //Sandcastle_Begin
+        const serviceResponse = await fetch("http://localhost:3000/service");
+        const { token } = await serviceResponse.json();
```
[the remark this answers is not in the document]
jjspace: I could add a function like that but I don't think users should ever be running this themselves. They should do their own oauth and just set the token. This request is just for the sandcastle examples and will only provide access to the sample itwins that we add to the service account.

jjspace:
@ggetz thanks for the comments, this should be updated in response to all of them. I'm still looking into the reality data functionality and hope to have some functions for that added early next week.

ggetz:
> I'm still looking into the reality data functionality and hope to have some functions for that added early next week.

Sounds good! Please include those in a separate PR if this is not merged first.

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Check the exports for the given iModel + changeset combination for any that
+ * have the desired CESIUM type and returns the first one that matches as a new tileset.
+ *
+ * If there is not a CESIUM export you can create it using {@link ITwinPlatform.createExportForModelId}
+ *
+ * This function assumes one export per type per "iModel id + changeset id". If you need to create
+ * multiple exports per "iModel id + changeset id" you should switch to using {@link ITwinData}
+ * with the export id directly
```
danieliborra, line 88: I would not include the functionality to specify the `changeset Id`, as it makes things more complicated while 99% of users will just get the latest version, and the other 1% will mostly be specific products that implements version comparison.

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @example
+ * TODO: example after API finalized
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @param {string} iModelId
+ * @param {string} changesetId
```
danieliborra, line 96: I'd remove changesetId

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,186 @@
...
+            // TODO: this is temporary, we should not have to call the Start Export route ever after
+            // auto generation is set up
+            statusOutput.innerText = "Starting export";
+            const exportId = await Cesium.ITwinPlatform.createExportForModelId(
+              imodelId,
+              changesetId,
+            );
+            statusOutput.innerText = "Creating Tileset from export";
+            tileset = await Cesium.ITwinData.createTilesetFromExportId(exportId);
+          }
```
danieliborra, line 140: I'd remove the logic to create an export. It is something we don't want to expose in our side, especially when in a few days (this week) we activate the auto-triggering for 3d tiles.
We had problems exposing this functionality in the past, so we can't expose it in this example.

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,186 @@
...
+        //Sandcastle_Begin
+        const serviceResponse = await fetch("http://localhost:3000/service");
+        const { token } = await serviceResponse.json();
+
+        Cesium.ITwinPlatform.defaultAccessToken = token;
+        // this is the iModel in the "Hello iTwinCesium" iTwin that we should all have access to
+        // https://developer.bentley.com/my-itwins/b4a30036-0456-49ea-a439-3fcd9365e24e/home/
+        // const imodelId = "2852c3d7-00c3-4b5d-a0ce-82bbde4f061e";
+        const imodelId = "88673c1d-12b8-48f1-8beb-5000d0edbd0b";
+        const changesetId = "";
```
danieliborra, line 41: I'd recommend to remove it, it generates noise around a concept that should be transparent to users.

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,186 @@
...
+
+        const statusOutput = document.querySelector("#status");
+        async function init() {
+          statusOutput.innerText = "Starting export";
+
+          const start = Date.now();
+
+          statusOutput.innerText = "Creating Tileset";
+
+          let tileset = await Cesium.ITwinData.createTilesetFromModelId(
```
danieliborra, line 126: What about `get` instead of `create`?
I guess you have a standard in the API for this.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @param {string} exportId
+ *
+ * @throws {RuntimeError} Unauthorized, bad token, wrong scopes or headers bad.
+ * @throws {RuntimeError} Requested export is not available
+ * @throws {RuntimeError} Too many requests
+ * @throws {RuntimeError} Unknown request failure
+ */
+ITwinPlatform.getExport = async function (exportId) {
```
danieliborra, line 123: Can be removed

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+  return result;
+};
+
+/**
+ * Get the list of exports for the given iModel + changeset
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @param {string} iModelId
+ * @param {string} [changesetId]
```
danieliborra, line 167: ChangesetId not necessary

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * @throws {RuntimeError} Unauthorized, bad token, wrong scopes or headers bad.
+ * @throws {RuntimeError} Not allowed, forbidden
+ * @throws {RuntimeError} Unprocessable: Cannot create export job
+ * @throws {RuntimeError} Too many requests
+ * @throws {RuntimeError} Unknown request failure
+ */
+ITwinPlatform.createExportForModelId = async function (iModelId, changesetId) {
```
danieliborra, line 250: This function can be removed

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,118 @@
...
+) {
+  const { exports } = await ITwinPlatform.getExports(iModelId, changesetId);
+  const cesiumExport = exports.find(
+    (exportObj) =>
+      exportObj.request?.exportType === ITwinPlatform.ExportType["3DTILES"],
+  );
+  if (!defined(cesiumExport)) {
+    return;
+  }
+  return loadExport(cesiumExport, options);
```
danieliborra, line 115: You can simplify this a lot just by querying getExports until you find the right export. Use `return=representation`

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,118 @@
...
+ * @param {Cesium3DTileset.ConstructorOptions} [options] An object describing initialization options.
+ * @returns {Promise<Cesium3DTileset>}
+ *
+ * @throws {RuntimeError} Wrong export type
+ * @throws {RuntimeError} Export did not complete in time.
+ *
+ * @example
+ * TODO: example after API finalized
+ */
+ITwinData.createTilesetFromExportId = async function (exportId, options) {
```
danieliborra, line 72: I'd remove this function too. We don't want to expose `getExport` either

Apps/Sandcastle/gallery/iTwin Demo.html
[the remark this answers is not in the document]
ggetz: I ask as it seams a bit strange to me that we would include a direct call to `fetch` in a Sandcastle example. 

It doesn't seem too foreign to me that we would have a function that abstracts away the relevant Cesium ion URL to get the access token under the hood that we plan on using for authentication. For example, this is similar to what `IonResource.fromAssetId ` is doing. In the future, we could potentially allow this function to reach out to an iTwin endpoint if available, or a user-provided endpoint without having to change the flow.

But let me know what you think.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+ * @param {string} iModelId
+ * @param {string} [changesetId]
+ *
+ * @throws {RuntimeError} Unauthorized, bad token, wrong scopes or headers bad.
+ * @throws {RuntimeError} Not allowed, forbidden
+ * @throws {RuntimeError} Unprocessable Entity
+ * @throws {RuntimeError} Too many requests
+ * @throws {RuntimeError} Unknown request failure
+ */
+ITwinPlatform.getExports = async function (iModelId, changesetId) {
```
danieliborra, line 175: You should combine `loadExports` and `getExports`, by using:

image

That will give you the url of the exports.
`get-export` will probably be deprecated too. It has performance problems because we will drop it soon, so I'd just use `get-exports`with `return=representation`

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+ * @property {number} maxEdgeLength
+ * @property {number} minBRepFeatureSize
+ * @property {number} minLineStyleComponentSize
+ */
+
+/**
+ * @typedef {Object} ViewDefinitionFilter
+ * @property {string[]} models Array of included model IDs.
+ * @property {string[]} categories Array of included category IDs.
+ * @property {string[]} neverDrawn Array of element IDs to filter out.
```
danieliborra, line 22: You can remove these options.
We will remove it from the documentation in the next days, so using it may break your sample

danieliborra (changes requested):
My apologies for not keeping the documentation up-to-date, which made you develop all the endpoints, when a few would have been enough.
On the other side, I think it has also helped us to understand how to improve the documentation.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+  const result = await response.json();
+  return result;
+};
+
+/**
+ * Get the list of exports for the given iModel + changeset
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @param {string} iModelId
```
ggetz, line 166: Here and throughout, I think we are missing parameter descriptions.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+    throw new RuntimeError(`Unknown request failure ${response.status}`);
+  }
+
+  /** @type {ExportResponse} */
+  const result = await response.json();
+  return result;
+};
+
+/**
+ * Get the list of exports for the given iModel + changeset
```
ggetz, line 162: Is this the intended final description? Should it be reviewed?

.prettierignore
```diff
@@ -11,6 +11,7 @@
 !packages/**/
 !Specs/**/
 !Tools/**/
+!itwin-oauth-demo/**/
```
ggetz, line 14: To confirm my understanding: The plan is to remove these files before merging, correct? It's just for demo purposes until the ion route is in place?

ggetz (commented):
Thanks @jjspace! Once you take a pass on these comments, could you please update the PR description with the current status? I'm a bit lost on what you're considering ready vs what is temporary.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,313 @@
...
+  if (defined(changesetId) && changesetId !== "") {
+    url.searchParams.set("changesetId", changesetId);
+  }
+  url.searchParams.set("exportType", ITwinPlatform.ExportType["3DTILES"]);
+  url.searchParams.set("$top", "1");
+  url.searchParams.set("client", "CesiumJS");
+  /* global CESIUM_VERSION */
+  if (typeof CESIUM_VERSION !== "undefined") {
+    url.searchParams.set("clientVersion", CESIUM_VERSION);
+  }
```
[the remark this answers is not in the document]
danieliborra: Only for `get-exports`.

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,186 @@
...
+            imodelId,
+            changesetId,
+          );
+          if (!Cesium.defined(tileset)) {
+            // TODO: this is temporary, we should not have to call the Start Export route ever after
```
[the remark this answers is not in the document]
danieliborra: Yes

itwin-oauth-demo/index.html
```diff
@@ -0,0 +1,34 @@
+<!doctype html>
```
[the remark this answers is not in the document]
jjspace: @ggetz I was thinking it might be good to keep something like this around and maybe actually solidify the popup login mechanic for development sandcastles. If we want to experiment with other iTwins it's probably quickest to add our user accounts to them instead of a service account meaning we need to do user oauth or pull the token directly from the docs or the auth API request.

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -119,17 +119,32 @@
...
+          } catch (error) {
+            if (!error.message.includes("not completed")) {
+              throw error;
+            }
+            statusOutput.innerText = "Tileset not completed, retrying...";
+
+            const timeoutAfter = 300000;
+            // wait until the export is complete
+            while (Date.now() - start > timeoutAfter) {
+              await delay(5000);
```
danieliborra, line 137:
```suggestion
              await delay(15000);
```

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,172 @@
+import Check from "./Check.js";
+import defined from "./defined.js";
+import DeveloperError from "./DeveloperError.js";
+import Resource from "./Resource.js";
+import RuntimeError from "./RuntimeError.js";
+
+/**
+ * @typedef {Object} GeometryOptions
```
danieliborra, line 8: Deprecated

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,172 @@
...
+ * @property {number} chordTol
+ * @property {number} angleTol
+ * @property {number} decimationTol
+ * @property {number} maxEdgeLength
+ * @property {number} minBRepFeatureSize
+ * @property {number} minLineStyleComponentSize
+ */
+
+/**
+ * @typedef {Object} ViewDefinitionFilter
```
danieliborra, line 19: Deprecated

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,172 @@
...
+ * @property {string[]} models Array of included model IDs.
+ * @property {string[]} categories Array of included category IDs.
+ * @property {string[]} neverDrawn Array of element IDs to filter out.
+ */
+
+/**
+ * @typedef {Object} StartExport
```
danieliborra, line 26: Deprecated

danieliborra (commented):
It looks good to me!
(I just added a few nitpicking comments)

jjspace:
@ggetz I think this is getting close and ready for another look. I also updated the main PR description to indicate a summary of our decisions and the remaining tasks

.prettierignore
ggetz: CC https://github.com/CesiumGS/cesium/pull/12289#discussion_r1850829211

Apps/Sandcastle/gallery/iTwin Demo.html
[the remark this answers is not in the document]
ggetz: @jjspace Thoughts?

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,194 @@
...
+          statusOutput.innerText = "Starting export";
+
+          const start = Date.now();
+          const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+
+          statusOutput.innerText = "Creating Tileset";
+          let tileset;
+          try {
+            tileset = await Cesium.ITwinData.createTilesetFromModelId(imodelId);
+          } catch (error) {
```
ggetz, line 128: Do we expect this case to happen really ever for the Sandcastle sample data?

If not, I think for simplicity's sake it should be removed.

I do think it would be helpful to be moved to a tutorial or similar, as discussed in https://github.com/CesiumGS/cesium/pull/12289#discussion_r1850829211.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,155 @@
...
+    url: `${ITwinPlatform.apiEndpoint}mesh-export`,
+    headers: {
+      Authorization: `Bearer ${ITwinPlatform.defaultAccessToken}`,
+      Accept: "application/vnd.bentley.itwin-platform.v1+json",
+      Prefer: "return=representation",
+    },
+    queryParameters: {
+      iModelId: iModelId,
+      exportType: ITwinPlatform.ExportType["3DTILES"],
+      // TODO: If we're only requesting the top 1 is there a chance it's `Invalid` instead of `Complete`
```
ggetz, line 121: Address or remove this TODO. 

I think we may have discussed offline that if the state is `Invalid`, then we do the simplest thing: Either return the response as-is or throw an error. I think I'm leaning towards to former.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,155 @@
...
+ */
+
+/**
+ * @typedef {Object} GetExportsResponse
+ * @property {ExportRepresentation[]} exports The list of exports for the current page
+ * @property {{self: Link, next: Link | undefined, prev: Link | undefined}} _links Pagination links
+ */
+
+/**
+ * Get the list of exports for the specified iModel at it's most current version. This will only return exports with {@link ITwinPlatform.ExportType} of <code>3DTILES</code>.
```
ggetz, line 89: If I understand correctly, this is returning one export, not a list correct?

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,79 @@
...
+import ITwinPlatform from "../Core/ITwinPlatform.js";
+import RuntimeError from "../Core/RuntimeError.js";
+import Check from "../Core/Check.js";
+
+/**
+ * Methods for loading iTwin platform data into CesiumJS
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @see ITwinPlatform to set the API token and base api url
```
ggetz, line 13:
```suggestion
 * @see ITwinPlatform to set the API token and base API URL
```

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,79 @@
...
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @param {string} iModelId The id of the iModel to load
+ * @param {Cesium3DTileset.ConstructorOptions} [options] Object containing options to pass to an internally created {@link Cesium3DTileset}.
+ * @returns {Promise<Cesium3DTileset | undefined>} Will return <code>undefined</code> if there is no export for the given iModel id
+ *
+ * @throws {RuntimeError} Wrong export type [type]
+ * @throws {RuntimeError} Export is not completed. [id] - [status]
+ */
+ITwinData.createTilesetFromModelId = async function (iModelId, options) {
```
ggetz, line 70: Shouldn't this be:

```suggestion
ITwinData.createTilesetFromIModelId = async function (iModelId, options) {
```

? 

I know the capitalization is a but funky, but I think using the correct product name is important, just like `GooglePhotorealistic3DTiles`.

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Loads the export for the specified iModel with the export type that CesiumJS can load and returns
+ * a tileset created from that export.
+ *
+ * If the export is not finished processing this will throw an error. It is up to the caller
+ * to re-attempt loading at a later time
+ *
+ * @example
+ * const tileset = await Cesium.ITwinData.createTilesetFromModelId(imodelId);
```
ggetz, line 58:
```suggestion
 * const tileset = await Cesium.ITwinData.createTilesetFromModelId(iModelId);
```

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @param {string} iModelId The id of the iModel to load
+ * @param {Cesium3DTileset.ConstructorOptions} [options] Object containing options to pass to an internally created {@link Cesium3DTileset}.
+ * @returns {Promise<Cesium3DTileset | undefined>} Will return <code>undefined</code> if there is no export for the given iModel id
+ *
+ * @throws {RuntimeError} Wrong export type [type]
+ * @throws {RuntimeError} Export is not completed. [id] - [status]
+ */
+ITwinData.createTilesetFromModelId = async function (iModelId, options) {
+  const { exports } = await ITwinPlatform.getExports(iModelId);
```
ggetz, line 71: If I recall correctly, I think we discussed with @danieliborra that we should be caching the results of this call on the client to avoid overloading the REST API.

We do this occasionally in other parts of the API. For instance, see `requestMetadata` in `BingMapsImageryProvider.js`.

ggetz (changes requested):
Overall this is looking good @jjspace! Looking much cleaner overall.

Are we at a point where we can add unit tests and a `CHANGES.md` update?

packages/engine/Source/Core/ITwinPlatform.js
jjspace: I think this is a question for @danieliborra.

When I was exploring the API thus far (and earlier using start-export multiple times) there was a chance there would be multiple exports for the same version of an iModel but some were `Invalid` and some were `Complete`.

I don't know how it will be handled in the auto-generation.
1. If the export fails for the type we need will the export get regenerated?
2. Will the `Invalid` export still be in the list?

Depending on the answer here I can update this.

My concern is that this API request returns a list and I don't see any way to filter the API request based on it's status. If `Invalid` exports don't get deleted then there's a chance the full page of requests we get returned would all be invalid (in this case only 1 but same concern for larger pages). If that's the case even if we do return that result to the user they can't do anything about it because we removed the ability to start exports or load specific export ids.

packages/engine/Source/Core/ITwinPlatform.js
jjspace: Right now I'm only requesting a page of `$top=1` based on the assumption it should always exist with the auto-generation. That said this _is_ still returning a list matching the response from the API. It's just a list of length 1.

packages/engine/Source/Scene/ITwinData.js
jjspace: I believe caching was suggested only if the Sandcastle was going to get a massive amount of traffic?

Regardless of that part I don't think we want to cache this without some extra thought. This route can return different results for the same input. For example when:
- The export is still processing and not `status = Complete`
- When the iModel has changed and needs to have a new export for the latest changeset

packages/engine/Source/Core/ITwinPlatform.js
danieliborra: You're completely right. I'd suggest the following:
- Get the latest 5 exports => $top=5
- Get the first one with status == "Complete"

If you don't find any, wait 15 seconds and try again.

packages/engine/Source/Scene/ITwinData.js
danieliborra: From the numbers Gabby provided, we won't have problems, so you can skip the cache, it will avoid headaches.
I was only afraid in the case where we had hundreds of concurrent users, which won't be the case (get-exports returns in 1-2 seconds max, so 1000 concurrent users would mean 1M/h users, which I assume won't happen).

itwin-oauth-demo/index.html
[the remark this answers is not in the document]
jjspace: I've had to go back to using this workflow for user auth for the Reality data and it will definitely be handy to have around. That said I only really added it to this repo to make it easier to share with those reviewing this PR. I'll pull it all out into another project that I can share as needed.

packages/engine/Source/Core/ITwinPlatform.js
jjspace: Discussed with @danieliborra offline and found the piece of the puzzle I was missing.

- The auto generation of exports automatically deletes the 6th export if it exists

Basically meaning there _should_ never be more than 5 exports. If they're all invalid then there are bigger problems with that iModel. With that info I'm perfectly fine to only request the `top=5` results and have updated the code to account for this.

aruniverse (commented):
This looks great, thanks @jjspace, can't wait to get our hands on it!

jjspace:
This is very nearly there now. We're just waiting on the ion route to be updated so auth can work for everyone but the sandcastle has been updated with the new sample data.

Remaining:

- Add tests for the new API functions (first on my list on Monday)
- Update the Sandcastle auth route
- Potentially (only if it's really quick to do so) implement some camera locations inside the station since I realized it's fully modeled and would be a good AEC showcase

The sandcastle should also have reality mesh data in it but that will be implemented through a follow up PR on Monday.

danieliborra (approved):
Looks good to me!

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @see ITwinPlatform to set the API token and base API URL
+ * @namespace ITwinData
+ */
+const ITwinData = {};
+
+/**
+ * Create a {@link Cesium3DTileset} for the given iModel id using the mesh export service.
```
ggetz, line 18:
```suggestion
 * Create a {@link Cesium3DTileset} for the given iModel id using iTwin's Mesh Export API.
```

Pulling names and capitalization from https://developer.bentley.com/apis/mesh-export/.

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,74 @@
...
+ * const tileset = await Cesium.ITwinData.createTilesetFromIModelId(iModelId);
+ * if (Cesium.defined(tileset)) {
+ *   viewer.scene.primitives.add(tileset);
+ * }
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ *
+ * @param {string} iModelId The id of the iModel to load
+ * @param {Cesium3DTileset.ConstructorOptions} [options] Object containing options to pass to the internally created {@link Cesium3DTileset}.
+ * @returns {Promise<Cesium3DTileset | undefined>} Will return <code>undefined</code> if there is no completed export for the given iModel id
```
ggetz, line 34:
```suggestion
 * @returns {Promise<Cesium3DTileset | undefined>} A promise that will resolve to the created 3D tileset or <code>undefined</code> if there is no completed export for the given iModel id
```

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @see ITwinPlatform to set the API token and base API URL
+ * @namespace ITwinData
+ */
+const ITwinData = {};
+
+/**
+ * Create a {@link Cesium3DTileset} for the given iModel id using the mesh export service.
+ *
+ * If there is not a completed export available for the given iModel id this function will return <code>undefined</code>.
```
ggetz, line 20:
```suggestion
 * If there is not a completed export available for the given iModel id, the returned promise will resolve to <code>undefined</code>.
```

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -0,0 +1,74 @@
...
+ * We recommend waiting 10-20 seconds and trying to load the tileset again.
+ * If all exports are Invalid this will throw an error. In that case there's likely something wrong with the iModel itself
```
ggetz, line 22:
```suggestion
 * If all exports are Invalid this will throw an error.
```

ggetz (changes requested):
Thanks for the update here @jjspace! The tests look great.

I did a final code pass, mainly on documentation and wording.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -0,0 +1,156 @@
...
+ * @namespace ITwinPlatform
+ */
+const ITwinPlatform = {};
+
+/**
+ * Status states for a mesh-export export
+ * @enum {string}
+ */
+ITwinPlatform.ExportStatus = Object.freeze({
+  NotStarted: "NotStarted",
```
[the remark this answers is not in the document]
jjspace: I had intended these to be private but then the types generation will complain about the places they're used. I'll add descriptions but no users should really ever have to use these

Apps/Sandcastle/gallery/iTwin Demo.html
```diff
@@ -0,0 +1,322 @@
...
+          sceneModePicker: false,
+          geocoder: false,
+          homeButton: false,
+        });
+        const scene = viewer.scene;
+        scene.globe.show = true;
+        scene.debugShowFramesPerSecond = true;
+
+        // Grabbed mapping from the iTwin Viewer
+        // TODO: grab updated categories for the new models
```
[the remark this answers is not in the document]
jjspace: We decided to drop these for the first version

packages/engine/Source/Core/ITwinPlatform.js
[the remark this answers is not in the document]
jjspace: It's not ideal but I've added the values in the enum descriptions directly. I think that's an ok compromise for now.

ggetz:
🎉 This is looking great @jjspace! Two comments on the Sandcastle:

1. I think we're missing a thumbnail image:

image

2. Can we start with the "Focus Station" view? I think it's more visually interesting than the bird's eye view.

jjspace:
@ggetz Good to go I think

ggetz:
Awesome! Thanks @jjspace! And a big thanks to @danieliborra and @aruniverse for your help!

ggetz (approved):

PR #12301, opened 2024-11-12
Fix label rendering bug in WebGL1 contexts

### LINKED ISSUE #12249: An error occurred while rendering a label when contextOptions.requestWebgl1 is true
### What happened?

```
  contextOptions: {
      requestWebgl1: true,
  },
});

function addLabel() {
  viewer.entities.add({
    position: Cesium.Cartesian3.fromDegrees(-75.1641667, 39.9522222),
    label: {
      "text": "text",
    },
  });
}
addLabel();
```
the code above will produce an error:
```
TypeError: Failed to execute 'texSubImage2D' on 'WebGLRenderingContext': parameter 9 is not of type 'ArrayBufferView'.
```
here is [sandcastle example](https://sandcastle.cesium.com/#c=bVDLbsIwEPwVy6cgpY5SCogQUCV6ROJQqb344jgLterYqR9QWuXfaxOgqGUP1u7szHhsrpV1aCdgDwbNkYI9WoIVviEvRyyhmB/npVaOCQWG4hR9U4UQDwh8unXrRDApejCWgQ8P1r1CtZV5gZzxkMZdF85uMKOKqo1XPMoQq+sVq0Amg17fJyGgnHACLAn75GTcaiuipjgnXDLjQsfUkGyMbp5gawBscjcZkXz8kI/HkxQNp2Q6uo81SHsbGa+7SktxfAXFxaU7Ebs+dAzcUfUbdIZTXFp3kLA4ezyKptXGIW9kQkjmoGklC9myyvN3cIRbG20itcyupWUtdkjU8xu/jLhk1obNxkv5LL6A4kWZBf4/qdSsFmq73oGR7BBpb/li1YOEkDIL422l01pWzPxx/gE)

### Reproduction steps

_No response_

### Sandcastle example

_No response_

### Environment

Browser: Chrome 129
CesiumJS Version: 1.122
Operating System: windows11

jjhembd:
# Description

This PR fixes a texture loading bug in WebGL1 contexts.

#12100 refactored the `Texture` class for better readability and documentation. In the process, some functions made use of the longer signature for the [texImage2D](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D) and [texSubImage2D](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D) methods, for pixel data supplied as _either a `TypedArray` or an `HTMLImageElement`_. However, only WebGL2 allows the longer signature for `HTMLImageElement` (or similar) data sources. This resulted in a `TypeError` when loading some textures in a WebGL1 context.

The errors came from the `Texture.prototype.copyFrom` method. As part of this fix, the loading calls for different `copyFrom` source types were pulled out into helper functions, for better readability and consistency with the `Texture` constructor.

## Issue number and link

Fixes #12249 

## Testing plan

Load this [local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVDLbsIwEPwVy6cgpY5SCogQUCV6ROJQqb344jgLterYqR9QWuXfaxOgqGUP1u7szHhsrpV1aCdgDwbNkYI9WoIVviEvRyyhmB/npVaOCQWG4hR9U4UQDwh8unXrRDApejCWgQ8P1r1CtZV5gZzxkMZdF85uMKOKqo1XPMoQq+sVq0Amg17fJyGgnHACLAn75GTcaiuipjgnXDLjQsfUkGyMbp5gawBscjcZkXz8kI/HkxQNp2Q6uo81SHsbGa+7SktxfAXFxaU7Ebs+dAzcUfUbdIZTXFp3kLA4ezyKptXGIW9kQkjmoGklC9myyvN3cIRbG20itcyupWUtdkjU8xu/jLhk1obNxkv5LL6A4kWZBf4/qdSsFmq73oGR7BBpb/li1YOEkDIL422l01pWzPxx/gE) and verify that the label renders without errors.

# Author checklist

### DIFF
--- a/packages/engine/Source/Renderer/Texture.js
+++ b/packages/engine/Source/Renderer/Texture.js
@@ -390,7 +390,7 @@ function loadBufferSource(texture, source) {
  * Load texel data from a buffer into part of a texture
  *
  * @param {Texture} texture The texture to which texel values will be loaded.
- * @param {object} source The source for texel values to be loaded into the texture.
+ * @param {TypedArray} arrayBufferView The texel values to be loaded into the texture.
  * @param {number} xOffset The texel x coordinate of the lower left corner of the subregion of the texture to be updated.
  * @param {number} yOffset The texel y coordinate of the lower left corner of the subregion of the texture to be updated.
  * @param {number} width The width of the source data, in pixels.
@@ -400,7 +400,7 @@ function loadBufferSource(texture, source) {
  */
 function loadPartialBufferSource(
   texture,
-  source,
+  arrayBufferView,
   xOffset,
   yOffset,
   width,
@@ -420,7 +420,6 @@ function loadPartialBufferSource(
   gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
   gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, false);
 
-  let { arrayBufferView } = source;
   if (texture.flipY) {
     arrayBufferView = PixelFormat.flipY(
       arrayBufferView,
--- a/packages/engine/Source/Renderer/Texture.js
+++ b/packages/engine/Source/Renderer/Texture.js
@@ -898,7 +898,6 @@ Texture.prototype.copyFrom = function (options) {
   gl.bindTexture(target, this._texture);
 
   let { width, height } = source;
-  const { arrayBufferView } = source;
 
   // Make sure we are using the element's intrinsic width and height where available
   if (defined(source.videoWidth) && defined(source.videoHeight)) {
@@ -927,8 +926,8 @@ Texture.prototype.copyFrom = function (options) {
       height === this._height
     ) {
       // initialize the entire texture
-      if (defined(arrayBufferView)) {
-        loadBufferSource(this, arrayBufferView);
+      if (defined(source.arrayBufferView)) {
+        loadBufferSource(this, source);
       } else {
         loadImageSource(this, source);
       }
@@ -942,7 +941,7 @@ Texture.prototype.copyFrom = function (options) {
   }
 
   if (!uploaded) {
-    if (defined(arrayBufferView)) {
+    if (defined(source.arrayBufferView)) {
       loadPartialBufferSource(this, source, xOffset, yOffset, width, height);
     } else {
       loadPartialImageSource(this, source, xOffset, yOffset);
--- a/packages/engine/Source/Renderer/Texture.js
+++ b/packages/engine/Source/Renderer/Texture.js
@@ -941,7 +941,14 @@ Texture.prototype.copyFrom = function (options) {
 
   if (!uploaded) {
     if (defined(source.arrayBufferView)) {
-      loadPartialBufferSource(this, source, xOffset, yOffset, width, height);
+      loadPartialBufferSource(
+        this,
+        source.arrayBufferView,
+        xOffset,
+        yOffset,
+        width,
+        height,
+      );
     } else {
       loadPartialImageSource(this, source, xOffset, yOffset);
     }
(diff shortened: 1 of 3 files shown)


packages/engine/Source/Renderer/Texture.js
```diff
@@ -448,6 +505,35 @@ function loadImageSource(texture, source) {
...
+ */
+function loadPartialImageSource(texture, source, xOffset, yOffset) {
+  const context = texture._context;
+  const gl = context._gl;
+
+  gl.pixelStorei(gl.UNPACK_ALIGNMENT, 4);
+  gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, texture.preMultiplyAlpha);
+  gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, texture.flipY);
+
+  gl.texSubImage2D(
```
ggetz, line 526: Maybe I'm just confused by the documentation, but [are height and with not required for a WebGL 2 context](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texSubImage2D)?

ggetz (changes requested):
Thanks for the fix @jjhembd! 

Would it be possible to add a spec to make sure this regression never happens in the future?

packages/engine/Source/Renderer/Texture.js
```diff
@@ -853,94 +919,33 @@ Texture.prototype.copyFrom = function (options) {
...
     } else {
       gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
       gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, false);
-      // initialize the entire texture to zero
-      pixels = PixelFormat.createTypedArray(
-        pixelFormat,
-        pixelDatatype,
-        textureWidth,
-        textureHeight,
-      );
```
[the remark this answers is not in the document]
jjhembd: After those docs burned me the first time around, I'm now relying on the spec rather than MDN.
The WebGL2 spec only mentions _differences_ from WebGL1. The way `null` is handled is mentioned [here](https://registry.khronos.org/webgl/specs/latest/1.0/index.html#TEXIMAGE2D):

> If pixels is null, a buffer of sufficient size initialized to 0 is passed.

packages/engine/Source/Renderer/Texture.js
jjhembd: The WebGL1 signature is [still supported in WebGL2](https://registry.khronos.org/webgl/specs/latest/2.0/#3.7.6:~:text=interface%20mixin%20WebGL2RenderingContextOverloads). Basically WebGL2 is [almost 100% backwards compatible with WebGL1](https://webgl2fundamentals.org/webgl/lessons/webgl1-to-webgl2.html), so any function that does not explicitly use WebGL2 features (aside from some extensions) will still work.

jjhembd:
Thanks for the review @ggetz !

> Would it be possible to add a spec to make sure this regression never happens in the future?

The existing `Texture` specs caught the bug, when run in WebGL1. But we have been running them in WebGL2 only.

Since the `Texture` class is so widely used, I think the safest route will be to test _everything_ in both WebGL1 and WebGL2. I updated `TextureSpec` to do this, using the `createWebglVersionHelper` method from #11201.

I think I addressed all the other feedback.

ggetz:
Thanks for the updates @jjhembd! I like the plan to run with both WebGL 1 and 2.

I'm seeing some hangups and unexpected failures when running the tests locally. Are you seeing anything similar?

jjhembd:
> I'm seeing some hangups and unexpected failures when running the tests locally. Are you seeing anything similar?

Yes, sorry, that was an oversight from https://github.com/CesiumGS/cesium/pull/12301/commits/d385f34702edd1a52b128cb84b2f4a1187f72257. Should be fixed now!

ggetz:
Awesome, this is now working for me! Thanks @jjhembd!

ggetz (approved):

PR #12307, opened 2024-11-13
Fixes unhandled rejection in Skybox.update  #12306

### LINKED ISSUE #12306: CesiumWidgetSpec leaves requests pending after tests finish
### What happened?

Random unit test failure due to interaction between `RequestScheduler.clearForSpecs()` and `CesiumWidgetSpec.js`.  It appear that the `CesiumWidgetSpec.js` tests have outstanding requests pending for loading the cubemap faces after they are finished.  If theses requests are still pending when  `RequestScheduler.clearForSpecs()`  is called it can result in an  exception being thrown in `afterAll` .

### Reproduction steps

1. Add `RequestScheduler.clearForSpecs();` to `afterEach` in `packages/engine/Specs/Widget/CesiumWidgetSpec.js`
2. Focus one of the tests in that file.  I used "flyTo flies to entity when offset is defined but other options for flyTo are not" but should work for most tests.
3. Run `npm run test`

The result is the following error:

>   Widget/CesiumWidget
>     ✓ flyTo flies to VoxelPrimitive with default offset when options not defined
> Chrome 130.0.0.0 (Linux x86_64) ERROR
>   An error was thrown in afterAll
>   Unhandled promise rejection: undefined thrown
> Chrome 130.0.0.0 (Linux x86_64) ERROR
>   An error was thrown in afterAll
>   Unhandled promise rejection: undefined thrown

### Sandcastle example

_No response_

### Environment

Browser:
CesiumJS Version:
Operating System:

angrycat9000:
# Description

Adds an error handler to `loadCubeMap` in `Skybox.update` and then throw any errors during the next call.

This allows the error to be thrown in normal usage when `update` is called repeatedly.  It prevents throwing the error after the end of a test because `update` is not called after the test ends.   Previously the error would be thrown in `afterAll` when the tests were running.

## Issue number and link

Fixes  #12306

## Testing plan

1. Checkout this branch
2. Focus the "handles error when Resources fail to load" spec
3. Run `npm run tests`. 
    - Verify the tests pass
5. Run `git checkout main -- packages/engine/Source/Scene/SkyBox.js` to get the previous version
3. Run `npm run tests`. 
    - Verify the tests fails with an error in `afterAll`

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.124.0
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Handle uncaught promise rejection in `Skybox.update` that could cause tests to fail[#12307](https://github.com/CesiumGS/cesium/pull/12307)
+
 ### 1.123.1 - 2024-11-07
 
 #### @cesium/engine
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,13 +1,5 @@
 # Change Log
 
-### 1.124.0
-
-#### @cesium/engine
-
-##### Fixes :wrench:
-
-- Handle uncaught promise rejection in `Skybox.update` that could cause tests to fail[#12307](https://github.com/CesiumGS/cesium/pull/12307)
-
 ### 1.123.1 - 2024-11-07
 
 #### @cesium/engine
--- a/packages/engine/Source/Scene/SkyBox.js
+++ b/packages/engine/Source/Scene/SkyBox.js
@@ -81,6 +81,8 @@ function SkyBox(options) {
 
   this._attributeLocations = undefined;
   this._useHdr = undefined;
+  this._hasError = false;
+  this._error = false;
 }
 
 /**
@@ -109,6 +111,15 @@ SkyBox.prototype.update = function (frameState, useHdr) {
   // The sky box is only rendered during the render pass; it is not pickable, it doesn't cast shadows, etc.
   if (!passes.render) {
     return undefined;
+  }
+
+  // Throw any errors that had previously occurred asynchronously so they aren't
+  // ignored when running.  See https://github.com/CesiumGS/cesium/pull/12307
+  if (this._hasError) {
+    const error = this._error;
+    this._hasError = false;
+    this._error = undefined;
+    throw error;
   }
 
   if (this._sources !== this.sources) {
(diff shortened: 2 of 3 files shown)


packages/engine/Source/Scene/SkyBox.js
```diff
@@ -141,10 +141,15 @@ SkyBox.prototype.update = function (frameState, useHdr) {
...
-        that._cubeMap = that._cubeMap && that._cubeMap.destroy();
-        that._cubeMap = cubeMap;
-      });
+      loadCubeMap(context, this._sources).then(
+        function (cubeMap) {
+          that._cubeMap = that._cubeMap && that._cubeMap.destroy();
+          that._cubeMap = cubeMap;
+        },
+        function (error) {
+          console.error("Skybox cube map failed to load", error);
```
ggetz, line 150: Typically, we don't like to log to the console, and prefer throwing errors where possible so the user can deal with it as appropriate for their app. Though the fact that this is in the update loop complicates things.

1. Ideally, the async code would be removed from the update loop entirely, and there would be a top-level async function like `SkyBox.fromCubeMap()`  or `SkyBox.fromImages()`. But that may require more refactoring than you're willing to do at this time.
2. Next best would be to await the promise, catch the error, and throw it synchronously in the next call to `update`. That way the unit tests can use `pollToPromise` call `update` until an error is thrown or  `_cubeMap` is defined.
3. Finally, as a comprise, could we instead use a [`oneTimeWarning`](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/oneTimeWarning.js#L29)?

CHANGES.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.124.0
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Handle uncaught promise rejection in `Skybox.update` that could cause tests to fail[#12307](https://github.com/CesiumGS/cesium/pull/12307)
```
ggetz, line 9: Unless the change affects how a developer would use the CesiumJS library, such as if a new function is added or a new type of error may be thrown, there's no need to mention it in the changelog.

ggetz (changes requested):
Thanks for getting to the bottom of this @angrycat9000! A few comments about how we might fix it.

packages/engine/Source/Scene/SkyBox.js
angrycat9000: 2 is good.  That preserves the expectation that the error is getting thrown, but allows the test to avoid throwing the error (as `update` is never called  second time)

angrycat9000:
Thanks for the feedback @ggetz .  Updated and ready for another look

angrycat9000:
Test failure only occurs in release mode.  For some reason `scene.mode = SceneMode.SCENE2D` when the test was run instead of `SCENE3D`.  Gave each test its own instance of `Scene` to prevent cross test contamination.

ggetz:
>  For some reason scene.mode = SceneMode.SCENE2D when the test was run instead of SCENE3D

The calls with the `toThrowDeveloperError` expectations are not run in the release tests, so it's likely a side effect of that. The fix here works. :+1:

Thanks @angrycat9000, this all looks good!

Issue #11067, opened 2023-02-03
Depth texture is unstablein post processing


bbbbx:
Here is a screen record:

https://user-images.githubusercontent.com/22176164/216549296-44c87e87-4c25-44d7-baf2-63cd5ddf8a96.mp4

[...]

PR #12129, opened 2024-08-16
Procedural IBL Enhancement in CesiumJS

ggetz:
# Description

This pull request introduces a new system for "procedural" lighting that builds on the Image-Based Lighting (IBL) approach in CesiumJS, aimed at improving the realism and visual quality of the default lighting environment. The changes focus on techniques that create a plausible environment map dynamically based on the current lighting conditions an existing atmospheric scattering code which renders the sky.

### [Image-Based Lighting Sandcastle](https://ci-builds.cesium.com/cesium/procedural-ibl/Apps/Sandcastle/index.html?src=Image-Based%20Lighting.html)

**After**
image

**Before**
image

### [Mirror Ball](https://sandcastle.cesium.com/?src=glTF%20PBR%20Extensions.html)

**After**
image

**Before**
image

[...]

PR #12316, opened 2024-11-19
Distance-based Gaussian weighting for Ambient Occlusion

### LINKED ISSUE #10106: Ambient Occlusion doesn't work！

jjhembd:
# Description

Ambient Occlusion (AO) for geospatial data needs to produce good results at a wide range of scales, without requiring the user to adjust parameters for each zoom level. This PR redefines the weight function for AO as a normalized Gaussian window, with a variance that scales with the distance from the camera.

Before this PR, the AO effect disappears when zoomed out:
image

After this PR, the AO effect blurs wider, remaining visible (though lower intensity) when far from the camera:
image

When viewing a scene from larger distances, some flickering and noise may be visible, partly due to https://github.com/CesiumGS/cesium/issues/11067. These can be reduced by setting `Viewer.camera.frustum.near` to `1.0` or more, to improve precision in the depth buffer.

## Issue number and link

Fixes #10106.

## Testing plan

# Author checklist

#TODO

- [x] Rework uniforms to expose loop counts and Gaussian window size
- [x] Add fallback to support variable loop counts in WebGL1
- [ ] ~Add fallback with deprecation message to support old uniforms~
- [x] Test performance, pre-compute window weights if needed
- [x] Choose good defaults
- [x] Resolve artifact at frustum boundary
- [x] Remove blur method, possibly replace with bilateral filtering


jjhembd:
The code is now ready for performance testing with different sample counts. There are 2 parameters:
- `directionCount`: The number of directions in which to raytrace away from a given pixel
- `stepCount`: The number of steps to take in the ray tracing along a given pixel

See [this local Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=xRmJctrI8lcmrtRapLDQLcCO34uxkyXlKzHOZreoioU0wBghsTrA2Ot/3+4ZSYjD18bvbcoVpJ6+u6ene9SiMUvHcjsMZI/2ndRPPrgujeNOOKIBeU+6W3T+edj75LIz9rl9edtWT1k7bgdfTbfVttqjyfdvrc8NGZD+9D6NAKmtnv12pJ3eurd/XCjGyWH79rjzZX7aOUkAfnvWUv3TzqX6+/U3dnZ4CeuX7Lj1efIHMDu5vjTx+ffvX9jZNfC4dm9POiPtrPOtIe8cNwbmj/j76VnHovS613fnOyPrcNr5czZIj347nXm9a8v0T75cG92t3W7QDWq1lrDtwqUBza07DgeHdJIMD9J+n0ZgYN/xY8oJ3DCIEzJldMYXAjojGYdvHCZ1t1z+3gqDxGEBjbpbVXLXDQhJ2Jj6AGkKdlWEOQEbOwkLgzJwQEM39GhUhsWo3wlAz5k7Ki/dV0p63RHXCaZOXIXfMY0c+PVDd1QV5OQeNBaqAw0HyQJP7kdpnIARAXXQLFVWSlwReO47Af1KJ2GUwLoXuumYBok8oMmRT/HxYN72wPgCt7uFigkGQshHIeMZTJbwBSOh7SSMk680AN/IjucdTYHkmMUJLEVSPw1c9CSRKsLfK2rLLAC0XzsnxyBbsOtHIOgicRKaO+Bi4rMkRmpC5LEzkWIEkPf75DQd90AKf6/ISfiR3VBP0iqoHCG5+1FmFWITcV8vO1cgrrtiSbGrt3fI475K3t4Bm/ur3UWIeWjlmCYfkiRivTSh0nbi9Bj442a7SraV7couqdXIeZqQPjg3JuCNZEizpCg4hIHrQxZhXmc+y12WYcqcWkKx91x04VvXmSRpRFvciguaJCwYxDl1OdgtSFU0J3MBKI37Q+J4Ho2BLkv70g5qOVECT06gS2/vMkKIOENE+Ya7ZBU63wi9va/wTRNGDDJkXdCv1PFA8XOWuMOvoe8vxA3FyhJXxCoDIqDgAiAuV0X0Q5/KfjiQFuZz96H3ikRfy9ntEZ174SyA6C3Sl+bBYH14kQGFvH//nmy72/kCeSgOXBtyv0gZ3P2ym0YRSO1A/YGQZD74nPrg6UOR++G4HYd1S1ElZNDd0hRN21HqO4raUZQm//uju9UNBFfUC/HeZKzOYVeeRyGeCLCZBvSY9SInmsss/jDuYQTOXNdPYzDuIp1gxsPG4RuwAhyFSTNI4nAmOz6NEknY2N3qDFlMelE4i6HWeiGNSRAmJBY8eGI7gj8JcwEEKwSZCGVkVJmQIg4iPz2s7JiMF0PH40X8KgWTwmhMYmc8AQ20Q4HUoTfoZCDPEfp+6CRkzAJ+PKwvODf5AngpIFPqamT6IxF8WmEYeZj3NOYI05B5QMICCZ/AF1lwBS+uAdaQ2/GPCLKSM5bKelU3ss5yIOfCoGA588OMmaAnO4UJFVIjUq52GZ5xWaF3ffCQVAZWiSIrVTwwchIwWs/pWqEfRgu5e4hcIf/hOJKKhEr2X4U0BbTMPGcZpsmPj5EzyNkBoiGVRRQKQJyvdtcivXxQr6arKEpwFgxwk4qsaK6mCa8nWbTjZr4Pc3c1uR0ZLPNmE3XisPv8mF4cYmX5MdYFqZBXPs858JBGbApFbEpfkLH/ev5x2TdQuogmK+TdRlxIOESTlrPHwE181Mpks2BKo5iCw66pKJDvRPxv5lWh5FL25aTityZ+5dnSpoj9cIKVcEdTFK5bf8Y8ME7g3j6adTxHBYOc+W3lkfRbBO+nEnE1B56dUcvyl3tFZ6U6857lAY5AuIou59tB7vlpdJHQyQW7Rb8qS90jNpzYuPqwwg30s+66SZIoFd0t4mSwRcu7on8JfX3brW+6eBjOSiSryp8F/ry0zAI4kKF7AJieMegxJ14I8GkwSIYtZ4IgU0jILC7JBEgrTIOkSSxD2MAikbUZWLWqeU+VJcEogCMaUk1OIscdSYW3Ft1zEoZ+j7flD3fMGY7olVc5O5OJPz+AE5Y3CYWEas4ZaSCQRMrafWgqCFaPQhURNzz2C5g8dOKzWQA5MoEjey4hUaXoT1Y0yJoWgnqf9eA0nzo9n5Y14eQFWpz2Yhc6XCqlEw8KRWmjiM15n53oRdO0hpf3pCKdIeWciCXDMXOXB7sD8AB1gpJdeXYKQc+aDp9iUpR2mQZouJeNYQK3yH3yyy/kzQK+kv67G7bE83g+gyVnVGznfHsBy2zqWbBYbhA2EefNxCbibE2UofVNubBFpMKbBywqscRtTv76q2Tkpp0utH24gG2iKQU2O5oeFZF3mo+JKarMJucUi09xwcK0iQHCn6ItytgmBsXiU1yKOreJS7H4FJfl2riJ1TLGc7TKTqCNSuFadkhvKBfiOujdyii5fGCXBlRURdV0Q9VsWdMahq0qalbzd0wFRylVtizLMHTA4GDd0uGfLSuGquumqhimmKcyiYWtj4rcgcNHVxXFMvSGpamqpavVDK4rWqOhqHpdrSv1usFPGr5gN1TdsAzFrlt1zVSNJbERGwyTp0TWDatuNjTFNg3NUmzBWZFV0zYUU1VhzTAtw87ApmI36g2tDqgGuGFJXDpZTKALObIbhRCDxf2IUKtaAhTuqYqrnU3a4ji5+64m7vXWLx5qtWddPWBUIa4yuFk1dR18VuUhBQfYstrQINRmw6zyeBqNugzY6Ni6YeKdAAp59rWDDrGpm5Zp1BXNNm1TN6robgWk2Wq9Du91S9dsg09Iig7utK2GpdfriglOFuLuS9596S0LN1VpyJoNtti6rRam2ipmNcQdjMtMteqyqelaHXLOtl54v6LLYCFkq6FBEjaADbfTMGwdzKtDxjZsy25YVWLJhg0Z1VBsTdFNu2HSHbuyuOVMork41vGWK8QrWLzgw1Y7ZoOAjHmbCdP/lHlwivTm5AD08+mcXMyhAozjxS1VwnwK7oJsdGYOSwrn8B/9sCOWxdVIGHyI4QUaLQ0cYpmqqERZpxyxMcPTVDTdGV9eZyB3wQFEolEURqU7svyq6OoIF6Af5t7KVWqSt3ec4v5KcNmqbu3Fydyn+8jgv2zMrz/SyJdkuQZWwRQMAa31UncEGrt5jyRO+8t23pP1oL0cRFBIvR0XJ5omiQY9R8LxW/zJRjFEgSWgEbSr7jDvt/ZquQ57HpsS5r3fcNONFwRxDCv91Pex3na39vdqgL9Elhl8BoMdDPCIMlT3jwVQluW9GryuUxUtLnfDXoKdQiEvszVbxOVe6M33845yL4mKZ3zz9i9jSqCRI7yn2KsBZHl9jwWTFNJkPqFo6JC6o154AxbCweHs4J1rDqZesxhpuLklXvASPaIEHwkf0mDxBu8v0qbolQp3iH+fInAxxTHkZfxW2sYlri8x9wSGCtHDHYq7mxdZHTnBAGTj8AdvOyo+OjfwqMETnu/wCGVSXbFi6vgpjGh531olHHDJG4Am2eYitv+5Sc7Nv2dS1k2/sknZnS0pOqyf3h3Yo79wZ6wpQbDHftVdsrGJ/6dOa+fN+0/EfxF+VVnE/4HgF9PCK0f/mM8ABIaAn7BEWVjynEQu5o5XtgWvpAgfHV4lKpbxZFSK2eeVLTkshoPXM0e1njRnef76X0QHe4RX3zIPZ1o+Br6yKQcwef//tgvO+T9hADwWvRE8YxslnrHTeqifyj5YE/75O4YGIp44gejLSl/8ocQjfH+vlwsTHyhJ9gV8iW7lI39OK9QSzR9y+Bs) for a demonstration of the parameters.

> Is the code ready for review yet, or should we wait for the uniform changes to go in first?

I think it would make sense to wait until the uniform changes are done. I still have one change to implement: the step size should be automatically computed from the window size (`lengthCap`) and `stepCount`. This will take a little more thought since the step size is in pixels, while the Gaussian window is currently defined in meters.

packages/engine/Source/Shaders/PostProcessStages/AmbientOcclusionGenerate.glsl
```diff
@@ -67,16 +76,35 @@ void main(void)
...
+    {
+        if (i >= directionCount) {
+            break;
+        }
+#endif
         sampleDirection = rotateStep * sampleDirection;
 
         float localAO = 0.0;
+        float accumulatedWindowWeights = 0.0;
+        //vec2 radialStep = stepLength * sampleDirection;
```
ggetz, line 95:
```suggestion
```

packages/engine/Source/Shaders/PostProcessStages/AmbientOcclusionGenerate.glsl
```diff
@@ -51,11 +60,11 @@ void main(void)
     }
 
     vec3 normalEC = getNormalXEdge(positionEC);
+    float gaussianVariance = lengthCap * sqrt(-positionEC.z);
+    // TODO: mix of units. Steps are in pixels; variance is in meters.
```
ggetz, line 64: Does `czm_metersPerPixel` help at all here?

ggetz:
I'm also seeing some flickering due to occlusion when I look up at the sky. Is that just me?

ggetz:
@jjhembd What's the plan for blur? I noticed your sandcastle disables it entirely. 

I agree the look of the blur is not great. Would it make sense to remove it completely? Maybe we could use MSAA to remove "noise" from few samples?

ggetz:
I'm seeing good results and a stable 60 FPS with 

```
  stepCount: 32,
  directionCount: 8,
```

I think stepCount could potentially go even lower to `16`, if we're looking fewer samples overall, and still look good.

That looks to be working well at other scales as well, such as buildings:

image

No AO, for comparison:
image

jjhembd:
@ggetz thanks for the initial feedback!

I reworked the uniforms to compute the step length from the window size and step count. This value also works as the scalar to account for discrete sampling of the Gaussian, so we don't have to compute normalization factors in the inner loop.

> I'm consistently seeing a band along the top of the screen

This does look like a frustum boundary. The occlusion should be neglecting this, since it appears in the depth buffer as a huge jump--much bigger than the window length. It's possible we're not correctly handling zero values in the depth buffer. I'm looking at this now.

> What's the plan for blur?

I agree the current blur doesn't look good. We discussed using a bilateral filtering at one point, which would better preserve edges. I will take a look today to see how much work that would be.

We still need to add some kind of deprecation for the old uniform names. I updated the TODO list in the description.

jjspace:
@jjhembd I'm seeing some flickering across the entire terrain as I zoom out in the sandcastle. I'm also seeing some banding on the smokestack things but I'm not sure if those are related. The flicker is quick and goes away when the camera settles but is that expected?


https://github.com/user-attachments/assets/785b8c92-7d9b-4bbc-9870-7f51c8701b10

packages/engine/Source/Scene/PostProcessStage.js
```diff
@@ -94,9 +94,16 @@ import PostProcessStageSampleMode from "./PostProcessStageSampleMode.js";
 function PostProcessStage(options) {
   options = defaultValue(options, defaultValue.EMPTY_OBJECT);
   const {
+    name = createGuid(),
     fragmentShader,
+    uniforms,
     textureScale = 1.0,
+    forcePowerOfTwo = false,
+    sampleMode = PostProcessStageSampleMode.NEAREST,
```
jjspace, line 102: This method of setting default values is _not_ equivalent to the previous `defaultValue` calls when the value is `null`. I just want to make sure that's intended

jjspace (commented):
This is definitely a nice change to be able to see the AO from further away. I _think_ the code looks good with my somewhat limited shader knowledge. I just had one small comment about the JS code. Also the bug(?) that I pointed out in my comment above

jjhembd:
@jjspace thanks for the review!

> I'm seeing some flickering across the entire terrain as I zoom out in the sandcastle.

I think this is related to https://github.com/CesiumGS/cesium/issues/11067#issuecomment-2322908364. We will still need to investigate the depth stability further, but I _think_ we can leave that as a separate issue--let me know if that approach makes sense to you.

> I'm also seeing some banding on the smokestack things

When using the log depth buffer with a very small value for the near plane (as is default), there is some inaccuracy in the far-away depth values. I just pushed an adjustment to the Sandcastle to push the near plane out to 1m. Does this reduce the artifacts on your end?

jjspace:
> We will still need to investigate the depth stability further, but I think we can leave that as a separate issue--let me know if that approach makes sense to you.

I think I'm ok with that. I think I agree that benefit of ao at more distances outweighs the slight flickering at certain heights. Hopefully we can swing back around to that depth plane sooner than later though.

>  I just pushed an adjustment to the Sandcastle to push the near plane out to 1m. Does this reduce the artifacts on your end?

It does, the smokestacks look good. It also seems to reduce the other flickering issue pretty significantly. Would it make sense to change the overall default?

jjhembd:
@jjspace adjusting the default near plane distance might have negative impacts on some other uses, where we need to render things close to the camera.
I updated both this PR description and `CHANGES.md` to recommend a larger near plane distance when AO at large distances is desired.

I also pushed a minor edit to `ConvolveSpecularMapFS` (from https://github.com/CesiumGS/cesium/pull/12129) to fix an error showing up in WebGL1 contexts.

Thanks again for your feedback! I think this should be ready to go now.

jjspace (approved):
Thanks @jjhembd! As discussed offline I think I agree that this is fine to go with the depth buffer issues which will need to be addressed as part of a larger effort for all post processing. I'm ok to merge this now

PR #12334, opened 2024-11-25
iTwin Reality Data integration

jjspace:
# Description

This PR is building off of #12289 to add an easy function to load Reality Data meshes. Currently we are only supporting 3D Tiles meshes and pointcloud types of Reality Data to keep the initial integration simple. More can be added in the future but they will require different datasources and constructors in CesiumJS and will require a little more thought on how to design the API.

## Issue number and link

No issue

## Testing plan

- Running the test sandcastle
    - Load the [new sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=iTwin%20Demo.html) and make sure it loads.
    - Check that the reality mesh in that iTwin loads and shows in the surrounding area.

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/ITwinPlatform.js
+++ b/packages/engine/Source/Core/ITwinPlatform.js
@@ -93,8 +93,6 @@ ITwinPlatform.SupportedRealityDataTypes = [
   ITwinPlatform.RealityDataType.PNTS,
   ITwinPlatform.RealityDataType.RealityMesh3DTiles,
   ITwinPlatform.RealityDataType.Terrain3DTiles,
-  ITwinPlatform.RealityDataType.KML,
-  ITwinPlatform.RealityDataType.Unstructured,
 ];
 
 /**
--- a/packages/engine/Source/Core/ITwinPlatform.js
+++ b/packages/engine/Source/Core/ITwinPlatform.js
@@ -38,6 +38,66 @@ ITwinPlatform.ExportType = Object.freeze({
 });
 
 /**
+ * Types of Reality data
+ * https://developer.bentley.com/apis/reality-management/rm-rd-details/#types
+ * @enum {string}
+ */
+ITwinPlatform.RealityDataType = Object.freeze({
+  Cesium3DTiles: "Cesium3DTiles",
+  PNTS: "PNTS",
+  OPC: "OPC",
+  RealityMesh3DTiles: "RealityMesh3DTiles",
+  Terrain3DTiles: "Terrain3DTiles",
+  "3MX": "3MX",
+  "3SM": "3SM",
+  CCCloudProject: "CCCloudProject",
+  CCImageCollection: "CCImageCollection",
+  CCOrientations: "CCOrientations",
+  ContextCaptureInputs: "ContextCaptureInputs",
+  ContextDetector: "ContextDetector",
+  ContextScene: "ContextScene",
+  DAE: "DAE",
+  DGN: "DGN",
+  DSM: "DSM",
+  FBX: "FBX",
+  GLB: "GLB",
+  GLTF: "GLTF",
+  KML: "KML",
+  LAS: "LAS",
+  LAZ: "LAZ",
+  LOD: "LOD",
+  LodTree: "LodTree",
+  OBJ: "OBJ",
+  OMI: "OMI",
+  OMR: "OMR",
+  Orthophoto: "Orthophoto",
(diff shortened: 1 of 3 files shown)


ggetz (changes requested):
Overall API looks good @jjspace! Just a few comments.

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -37,6 +37,64 @@ ITwinPlatform.ExportType = Object.freeze({
...
+  SHP: "SHP",
+  SLPK: "SLPK",
+  SpaceEyes3D: "SpaceEyes3D",
+  STL: "STL",
+  TSM: "TSM",
+  Unstructured: "Unstructured",
+  Other: "Other",
+});
+
+ITwinPlatform.SupportedRealityDataTypes = [
```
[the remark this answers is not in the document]
jjspace: I shifted this supported list into the only function that uses it. `RealityDataType` cannot be made private if I want to set it as the type of the param in `createTilesetForRealityDataId` because of the way types get generated and skipped for `private`

jjspace:
@ggetz I've added tests now and I think addressed all your comments above. Should be ready for another look and I think it's close to done if not already there

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -71,4 +72,63 @@ ITwinData.createTilesetFromIModelId = async function (iModelId, options) {
   return Cesium3DTileset.fromUrl(resource, options);
 };
 
+/**
+ * Create a tileset for the specified reality data id. This function only works
+ * with 3D Tiles meshes and point clouds.
+ *
+ * If the <code>type</code> or <code>rootDocument</code> are not provided this function
+ * will first request the full metadata for the specified reality data to fill these values.
+ *
```
ggetz, line 81: For consistency with `createTilesetFromIModelId`.

```suggestion
 *
 * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
```

ggetz (changes requested):
This is looking good @jjspace!

I noticed that the reality mesh looks a bit blurry at the default settings. I was getting a comparable visual to other photogrammetry tilesets after setting the max SSE to `4` —

```
realityMesh.maximumScreenSpaceError = 4;
```

I'm guessing this may be because the geometric error of the tileset is balanced against the iTwin viewer, what may have different tile selection metrics. Have you seen the same for other Reality Data meshes?

What do you think of passing a default of `4` for all with the type of `RealityMesh3DTiles`?

jjspace:
@ggetz I agree the tileset in the sandcastle was too low quality by default. I check some other reality meshes I have access to and sse of 4 looked better on all of them too so I just set it as the default as you suggested.

ggetz (approved):
This looks fantastic, thanks @jjspace!

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -71,4 +72,67 @@ ITwinData.createTilesetFromIModelId = async function (iModelId, options) {
...
+ * @param {string} realityDataId The id of the reality data to load
+ * @param {ITwinPlatform.RealityDataType} [type] The type of this reality data
+ * @param {string} [rootDocument] The path of the root document for this reality data
+ * @returns {Promise<Cesium3DTileset>}
+ */
+ITwinData.createTilesetForRealityDataId = async function (
+  iTwinId,
+  realityDataId,
+  type,
+  rootDocument,
```
matmarchand, line 94: I believe type and rootDocument should be extracted from the reality data metadata to ensure that they are consistent with what actually exist in the storage container... or maybe this is our way to avoid a call to get metadata since they were already extracted and assumed consistent?

jjspace: > or maybe this is our way to avoid a call to get metadata since they were already extracted and assumed consistent?

@matmarchand You are correct. If either of these values is missing we will fetch the metadata needed to load the tileset. This will likely only the rootDocument commonly but we grab and update both from the new request anyway.

However if the application using this function already has these values we can skip that network call. This could happen if the app [loads the list of reality data](https://developer.bentley.com/apis/reality-management/operations/get-all-reality-data/) using `return=representation` instead of `return=minimal`. Or if the app already called the individual metadata route themselves. No need to waste extra time and bandwidth if the info is already accessible

PR #12341, opened 2024-11-26
Fix sandcastle contributors guide link

jjspace:
# Description

This is a very small edgecase scenario that happens when someone tries to open a sandcastle html file directly instead of using the local server. The page it was trying to direct to is on the Wiki we don't really use anymore and instead tried to create a page (if you have access) or redirects to the home page instead of the info you need.
There was one other instance of this check that _did_ have a valid link so this just updates it to match that one
https://github.com/CesiumGS/cesium/blob/fddd0f17573373275d3026eaf502f8ede03c895d/Apps/Sandcastle/index.html#L36-L37

## Issue number and link

No issue

## Testing plan

- Open the Sandcastle gallery in your file explorer
- Open any html file directly
- You should see an alert about using the local server and when clicking OK it should redirect you to the contributor docs.

# Author checklist

### DIFF
--- a/Apps/Sandcastle/Sandcastle-header.js
+++ b/Apps/Sandcastle/Sandcastle-header.js
@@ -105,7 +105,7 @@
       )
     ) {
       window.location =
-        "https://github.com/CesiumGS/cesium/wiki/Contributor%27s-Guide";
+        "https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/BuildGuide/README.md#quickstart";
     }
   }
 })();


lukemckinstry:
Hi, what do you mean by this step? 
> Open any html file directly

jjspace:
Open the file directly, like double click it and open in a browser not loading it from the localhost server

jjhembd (approved):
Here's what I see when I load one of the HTMLs from Windows File Explorer:
image

If I click on "OK" then it takes me here:
image

Looks good to me!

PR #12343, opened 2024-11-27
add param computeLength and append v_distanceCamera when needed

mickae1:
I'm working on custom shader, and I've some change needed to be done, on the core of cesium that I think will help others also.

By example, the dash material for polyline will be improved later ( the dash pattern depend of the position of the camera, which make the dash pattern move when the camera move which is not nice )

To help that, I need to know the distance of the fragment shader from the first point of the polyline. 
That why, I added an attribute in the class PolylineGeometry. By default it's desactived, and it can be activated with the option computeLength.

In my custom shader, I need also to know the distance of the camera, which is already done here : 

https://github.com/CesiumGS/cesium/blob/c403b50282030c17739ea65d6c7b9a835792f4c2/packages/engine/Source/Scene/Primitive.js#L1004

so I just prepend in the vertexshader  : 
    "out float v_distanceCamera; \n" +

when the function _appendDistanceDisplayConditionToShader  is called.

which help me to show/hide, change object property in the custom shader depending of the distance of the camera from the object.

I think this will help also other.

Michael,


ggetz:
Thanks for the suggestion @mickae1! A world-space polyline dash material (rather than our current screenspace implementation) has been a common community request, so happy to hear you are working on it.

@jjhembd Would you be able to take a look at the overall approach here?

PR #12344, opened 2024-11-27
iTwin Reality Data - add GeoJSON/KML support

jjspace:
# Description

This is a follow up to #12334 to add support for GeoJSON and KML data types from the Reality Management API.

The implementation should be good to go but the Sandcastle will need updated to the iTwin and data we actually want to use.

- [x] Updated Sandcastle data

## Issue number and link

No issue

## Testing plan

- Check the new `iTwin Feature Service` sandcastle and make sure all data is loading as expected
    - Confirm the thumbnail loads as expected
    - Confirm all the layer toggles work and that the data always looks good and visible with different sets turned on and off

# Author checklist


packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -38,54 +38,19 @@ ITwinPlatform.ExportType = Object.freeze({
 });
 
 /**
- * Types of Reality data
+ * Types of Reality data. This is a partial list of types we know we can support
```
jjhembd, line 41: Why are we dropping the unsupported types here?
In the loading functions, we already create a local `supportedRealityDataTypes` array, so it seems like this enum could still retain all valid types, including unsupported ones.

Apps/Sandcastle/gallery/iTwin Feature Service.html
```diff
@@ -0,0 +1,126 @@
...
+          "ebec69b5-0b5f-49d8-9081-e29bcd517f6b",
+        );
+        viewer.dataSources.add(points);
+        viewer.dataSources.add(lines);
+        viewer.dataSources.add(areas);
+
+        // Create tileset of the reality data mesh
+        // TODO: swap this out with a different mesh
+        const realityMeshId = "85897090-3bcc-470b-bec7-20bb639cc1b9";
+        const realityMesh = await Cesium.ITwinData.createTilesetForRealityDataId(
```
jjhembd, line 63: This tileset is really blurry for me, unless I force the `maximumScreenSpaceError` smaller.

Current `maximumScreenSpaceError: 4`:
image


With `maximumScreenSpaceError: 1`:
image

packages/engine/Source/Core/ITwinPlatform.js
jjspace: This was requested by the iTwin team. The types are potentially changing and/or not a complete list so I trimmed this down to only those types we really care about. I still wanted to keep it in the enum instead of only in the array of `supportedRealityTypes` to make it centrally located and easy to change later.

Apps/Sandcastle/gallery/iTwin Feature Service.html
jjspace: This was a problem in the other sandcastle as well. `4` seemed to be a decent value overall but maybe not that far zoomed out. This is not the final mesh going to be used for this Sandcastle so I can re-evaluate that when I update to the other data.

jjspace:
Thanks for the review @jjhembd!

> I guess we want to wait for the final data before merging?

That is correct. The code shouldn't change but I will change the iTwin and specific reality data assets the sandcastle is pointing at before this is ready to merge

jjspace:
@jjhembd I believe this should be good to go now, please take another look

jjhembd (approved):
Thanks @jjspace, the new data looks really good!

ggetz:
@jjspace I don't see any updates in `CHANGES.md` covering these updates. Could you open a PR with them for today's release?

jjspace:
@ggetz the changes added here combined the changes from this PR and #12334 into one comment under the previous version because I expected both PRs to make it into the last release. Just opened #12407 to fix this.

PR #12354, opened 2024-12-02
Add label to prs that need a CLA signed

jjspace:
# Description

The label is helpful when viewing PRs in other Github views. I've been doing this manually to a few PRs but it should be easily automatable

## Issue number and link

no issue

## Testing plan

Was a label added to this pr when it was not a conditional check? Then we're good

# Author checklist


jjspace:
@siddheshranade I recall you did the initial implementation of the CLA check, any ideas why this is not working? Do re-runs of the workflow pull from the latest version of the script on this branch or from `main`?

siddheshranade:
@jjspace I remember facing the same issue. Yes, manual re-runs of the workflow will execute the code in `main`. 

A workaround I remember using was to update `cla.yaml` to trigger the action on a `push` instead of `pull_request_target`:

```
on:
  push:
```  

This way you don't have to manually re-run the workflow, it triggers on every `git push`, and executes the code on whichever branch you're pushing to. You can reset this YAML file change once you confirm the label is being set.

Correct me in case I'm wrong about this and I'm happy to look into it more deeply.

ggetz:
Yes, workflows with [`pull_request_target` ](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target) do not run in the context of the branch for security reasons (as the granted GitHub token has write access). So that means this will not trigger in this or any PR until the code is merged into main.

TLDR; It's a pain to test for a reason.

Automating this label gets a 👍 from me.

ggetz (approved):

PR #12358, opened 2024-12-04
fixes issue #12356 performance drop due to EnvironmentMap

### LINKED ISSUE #12356: Severe performance drop since 1.123
### What happened?

Since Cesium v1.123 I'm getting a severe performance drop when moving 10-20 entities on large distances.

### Reproduction steps

Check below sandcastle. In Cesium v1.122 the FPS is 60 while here it is 6.

The strange thing is that if you change the playRate value from 64 to 1 (interval update staying unchanged to 50ms) the FPS is back to 60 with some drops from time to time.

The same thing occurs when using CesiumWidget instead of Viewer.
The same thing occurs when using PositionCallbackProperty instead of direct position update.

### Sandcastle example

https://sandcastle.cesium.com/#c=jVNtb9owEP4rJ7QPoWQOUKgm3rSOalKlTavGtE+RVhMfYM2xke3AWMV/38VJaDfQNkTil3t57p57khntPOwk7tHCFDTuYY5OFjn7Gu6itJWF89xoz6VGm7ZieEo1gNuYQolbLXPucQQrrhzGqT62x6lOdZWSuQw1MoHLYr3YmP17y3N0D2gXmBktCNHbAkPAqtCZl0ZDZpESfjQCVSRFDFvjZGloV7AWfWF1XTJD7cmIjnEhomAHkGJET1wdmujRaVcb8hJgBHUMQGHlCNIWYwn9FzzfKrzjnifBzyUVK7fS1rtvtGVrtSQ6mgy5JC6K/EH+QLWQP4mT635tPIY1UHMsm1XoQXEqvRBIJAyGZFgZC1FpkHTTHdMygT6tnY6sW3/JzGPo/PDqSR4f42Zmc2497bi+Zitr8jtcW0QXve6xYXzCi+FNt9ttl7XAcxEdwmS9pr6a3YymZTlbqcMXU7Er0HmpecXo/6AO6Ol1w68dSBCFrcO7J7GUXWvqusTPgiK3ih8+U7N0eTMIPg79vfZod1xFURums5qT4K6MXjdklrjQoXxXz1muqDmqoN+7qbq+QD/AvwcAZ7Jbo393uBe/jaPNGqlRnr9zdKr7bDzjCu98PiSj8tXp6HJYMQyDbytuTZw/KJw1anwr862xnnStIlK0R1I0UeGSZZF9R88y5xqQSfIydCLkjr6f6YUvHzLFnSPLqlBB4mlrNknI/yxUGS6kXn/aoaUZlG6b3uxDdckYmyR0vBzpjVFLbv/I/As

### Environment

Browser: Chrome 131
CesiumJS Version: v1.123 and above
Operating System: masOS Sequoia

jfayot:
# Description

This is a proposal to fix / workaround performance drop due to EnvironmentMap.

## Issue number and link

Fixes #12356 

## Testing plan

- Updated Cesium3DTilesetVisualizerSpec and ModelVizualizerSpec
- Run following [Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=jVNtb9owEP4rJ7QPoWQOUKgm3rSOalKlTavGtE+RVhMfYM2xke3AWMV/38VJaDfQNkTil3t57p57khntPOwk7tHCFDTuYY5OFjn7Gu6itJWF89xoz6VGm7ZieEo1gNuYQolbLXPucQQrrhzGqT62x6lOdZWSuQw1MoHLYr3YmP17y3N0D2gXmBktCNHbAkPAqtCZl0ZDZpESfjQCVSRFDFvjZGloV7AWfWF1XTJD7cmIjnEhomAHkGJET1wdmujRaVcb8hJgBHUMQGHlCNIWYwn9FzzfKrzjnifBzyUVK7fS1rtvtGVrtSQ6mgy5JC6K/EH+QLWQP4mT635tPIY1UHMsm1XoQXEqvRBIJAyGZFgZC1FpkHTTHdMygT6tnY6sW3/JzGPo/PDqSR4f42Zmc2497bi+Zitr8jtcW0QXve6xYXzCi+FNt9ttl7XAcxEdwmS9pr6a3YymZTlbqcMXU7Er0HmpecXo/6AO6Ol1w68dSBCFrcO7J7GUXWvqusTPgiK3ih8+U7N0eTMIPg79vfZod1xFURums5qT4K6MXjdklrjQoXxXz1muqDmqoN+7qbq+QD/AvwcAZ7Jbo393uBe/jaPNGqlRnr9zdKr7bDzjCu98PiSj8tXp6HJYMQyDbytuTZw/KJw1anwr862xnnStIlK0R1I0UeGSZZF9R88y5xqQSfIydCLkjr6f6YUvHzLFnSPLqlBB4mlrNknI/yxUGS6kXn/aoaUZlG6b3uxDdckYmyR0vBzpjVFLbv/I/As).

# Author checklist


ggetz:
Thanks for the PR @jfayot! Instead of a specific `enableEnvironmentMap` property, what do you think of allowing the constructor options, similar to [what we do for Model.js](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/Model/Model.js#L159).

Also, given that entities' positions are usually more dynamic than static models or tilesets, I'm thinking for entities specifically, we make the threshold for updating environment maps much higher or disable them entirely. That way, the environment map generation only runs once on load.

jfayot:
Thanks for the review, @ggetz !

> what do you think of allowing the constructor options, similar to [what we do for Model.js](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/Model/Model.js#L159)?

As far as you consider `DynamicEnvironmentMapManager` as a high-level interface, yes, this can be added to `Entity`'s constructor options. I didn't go this way because, in my understanding, it was rather a low-level interface associated with primitives.

> for entities specifically, we make the threshold for updating environment maps much higher

I think that finding a threshold that fits for all use cases will be difficult ! It's probably better to disable the environment map by default for entities and let the user decide the appropriate options.

jfayot:
@ggetz I've added environmentMapOptions to Entity constructor. Tell me if this is what you had in mind...

jfayot:
Hi @ggetz , hope you're doing well !
Do you think there's a chance to get this PR in 1.125 ?

ggetz:
Hi @jfayot! Apologies on the delayed responses– Many have been out for the holidays.

Work done in https://github.com/CesiumGS/cesium/pull/12370 appears to address the performance issue demonstrated in https://github.com/CesiumGS/cesium/issues/12356. Can you confirm?

However, the additional failsafe in this PR is also beneficial IMO. We'll take a pass on this and hopefully get it into today's release. Thank you!

jfayot:
Hi @ggetz ! Happy new year to you and the Cesium team !

Yes I can confirm that your PR #12370 fixes the issue: back to 60fps when moving entities more than 1000m.

ggetz:
> I think that finding a threshold that fits for all use cases will be difficult ! It's probably better to disable the environment map by default for entities and let the user decide the appropriate options.

I was thinking we don't entirely disable the environment map, but set the update timing to Infinity so that it's only generated once. If you agree, I can push the update to this branch.

jfayot:
> > I think that finding a threshold that fits for all use cases will be difficult ! It's probably better to disable the environment map by default for entities and let the user decide the appropriate options.
> 
> I was thinking we don't entirely disable the environment map, but set the update timing to Infinity so that it's only generated once. If you agree, I can push the update to this branch. 

Sure, that's fine for me!

ggetz:
@jfayot If these updates work for you, I think this should be ready to go!

jfayot:
> @jfayot If these updates work for you, I think this should be ready to go!

It works for me @ggetz ! Thanks for the update.

ggetz:
Thank you @jfayot!

Issue #12053, opened 2024-06-26
CesiumJS should not use `push.apply` for potentially large arrays


javagl:
The issue actually is about a `RangeError: Maximum call stack size exceeded`. But I'll go out on a limb and write this directly as the proposed solution here.

There are several places where the `push.apply` is used for potentially large arrays. But `push.apply` **puts the arguments on the stack!** (The same as with the `...spread` operator). That's not good. 

The following is an example to reproduce the main problem: 
```
function test() {
  const targetArray = [];
  const sourceArray = [];
  for (let i=0; i<150000; i++) {
    sourceArray.push(i);
  }
  // eslint-disable-next-line prefer-spread
  targetArray.push.apply(targetArray, sourceArray);

}

try {
  test();
} catch (error) {
  console.log(error);
}
```
Just running it with `node _CallStackTest.js` will cause the
`RangeError: Maximum call stack size exceeded`

The `push.apply` pattern should be replaced by a `for`-loop. Preferably, with a target array with a pre-allocated size (offering an additional potential performance benefit - to be benchmarked if necessary...). 

(An aside: As seen in the code snippet, there's actually an eslint rule for that - but the `...spread` operater would **not** solve the issue here!)

---

One specific example of this pattern is in [`GltfLoader.parse`](https://github.com/CesiumGS/cesium/blob/9d7b9802dc37f59cfb5738c17af998fa2454b411/packages/engine/Source/Scene/GltfLoader.js#L2686):
```javascript
// Gather promises and handle any errors
const readyPromises = [];
readyPromises.push.apply(readyPromises, loader._loaderPromises);
```

This actually **did** trigger the `RangeError` for a real-world glTF model that contained >150000 accessors. While such a model may certainly fall into the category of [Models that challenge engine's performance or runtime limits](https://github.com/KhronosGroup/glTF-Sample-Assets/issues/56) , it should not cause a crash.

Simply replacing the above snippet with
```javascript
  // Gather promises and handle any errors
  const readyPromises = [];
  const n = loader._loaderPromises.length;
  for (let i=0; i<n; i++) {
    readyPromises.push(loader._loaderPromises[i]);
  }
```
solved the error. But of course, there are other places that may have to be fixed.

---

If someone wants to try it out: Here is an **artificial** model that causes the error:

[callStackTest_250x250.zip](https://github.com/user-attachments/files/15995838/callStackTest_250x250.zip)

---

[...]

Issue #8359, opened 2019-11-05
Use ES6 "class" as module syntax


worlddai:
So I can extend your class directly using `extends`,and so on.

OmarShehata:
This is to support browsers that don't yet support ES6 syntax without requiring a transpilation step. Here's a quote [from our blog post going over the whole migration](https://cesium.com/blog/2019/10/31/cesiumjs-es6/) that I encourage you to check out:

> Even though we’ve moved to ES6, the combined Cesium.js file still supports browsers that lack ES6 support, such as IE 11. As long as your build system can target IE 11, Cesium ES6 modules will continue to work without requiring additional transpilation.

With that said, having a transpilation is something we want to support in order to use newer ES6 syntax and features. Also from the blog:

> Since Chrome, Firefox, and Safari already support some of the latest JavaScript features, we should be able to keep our minimal build during development policy in place and start to leverage newer features such as async/await/Promise/generators/etc… We would still need to run the final output through Babel as part of the build process, but Rollup already has good support for this step. 

I'm going to leave this open as a feature request to use this particular ES6 feature, since even if we do add a transpilation step it would have to be another effort to change how classes are declared.

mramato:
@OmarShehata I'm actually going to close this, the use of class will be part of a much larger discussion/decision/change so having this open as a standalone issue isn't really actionable.

thw0rted:
@mramato  is there a different (good) place to discuss this publicly?  I'm currently consuming Cesium from Typescript (c.f. #5717 #4434 etc) and I *think* I could fully automate that process if the "class-like" new-able function pattern that Cesium uses were to move to ES2015 class syntax.  I explain in the "Typings" issue, but briefly, Typescript can successfully extract type information from JSDoc for *almost* all of Cesium, except where getter/setter functions are passed to `Object.defineProperties` and decorated with `@memberof`. 

As a small test case, I converted `Viewer.js` by hand, which took about 5 minutes of manual copy/pasting.  Most of the work was converting the getter/setter syntax from `foo: { get: function(){...}, set: function(...) { ... } }` to `get foo() {...}; set foo(...) { ... };`.  I haven't looked at automating this yet but I believe I can at least make the manual process pretty straightforward.  After making this conversion, the Typescript output (`.d.ts` files) is correct, and the test suite still passes.

ggetz:
I think it's valid to have this issue open:

1. We no longer support IE 11
[...]

PR #12361, opened 2024-12-05
Replace `push.apply` with dedicated function

javagl:
Fixes https://github.com/CesiumGS/cesium/issues/12053
Makes https://github.com/CesiumGS/cesium/pull/12176 obsolete

# Description

The title of the issue is _"CesiumJS should not use `push.apply` for potentially large arrays"_ , which bears the question of what is a "large" array. Given that the answer to that question is the same as to "How long is a piece of string?", this PR just replaces **all** appearances of 
`target.push.apply(target, source);` 
with
`addAll(source, target);`
using a newly introduced `addAll` function.

This `addAll` function is a top-level function, but marked as `@private` for now.

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12053

## Testing plan

The case where this issue originally came up was that of a glTF with many accessors: Each accessor causes a 'promise' to be created, and these promises had been pushed to a target array with `push.apply`. For many accessors, this caused a `RangeError`, as described in the issue.

The following is a sandcastle that loads such a glTF asset, and can be used to verify that the issue no longer appears:

```javascript
  globe: false
});

const url = "http://localhost:8003/callStackTest_250x250.glb";

const camera = viewer.scene.camera;
camera.position = new Cesium.Cartesian3(500, 140, 140);
camera.direction = Cesium.Cartesian3.negate(
  Cesium.Cartesian3.UNIT_X, new Cesium.Cartesian3());
camera.up = Cesium.Cartesian3.clone(Cesium.Cartesian3.UNIT_Z);
camera.frustum.fov = Cesium.Math.PI_OVER_THREE;
camera.frustum.near = 0.01;
camera.frustum.far = 1000.0;
camera.lookAtTransform(Cesium.Matrix4.IDENTITY);

let totalDurationMs = 0;


async function addModel(runs, currentRun) {
  if (currentRun === runs) {
    const averageDurationMs = totalDurationMs / runs;
    console.log("runs: "+runs);
    console.log("totalDurationMs: "+totalDurationMs);
    console.log("averageDurationMs: "+averageDurationMs);
  } else {
    
    console.log("Adding, run " + currentRun + " of " + runs);
    const before = performance.now();
      url: url,
    });

    model.readyEvent.addEventListener(async () => {
      const after = performance.now();
      const durationMs = after-before;
      console.log("ready, took "+durationMs+" ms");
      totalDurationMs += durationMs;
      if (currentRun < runs - 1) {
        viewer.scene.primitives.remove(model);
      }
      await addModel(runs, currentRun + 1);
    });
  }  
}

addModel(1, 0);
//addModel(10, 0);
```

The following archive contains the main asset for the test:

[callStackTest.zip](https://github.com/user-attachments/files/18027162/callStackTest.zip)

When using the `callStackTest_250x250.glb`, then the current `main` state causes an error. With the fixed state, it works.

--- 

The archive contains a few smaller assets as well, and the sandcastle offers a **very** basic "benchmarking" functionality, to repeatedly load an asset. This is intended for checking whether this change has a noticable performance impact **on this particular code path**. This makes limited sense: There have been _many_ places that used `push.apply`. Many of them never saw anything that even remotely was a "large" array. In the specific case of the glTF models, most of the time was **not** spent in `push.apply`, but with loading the data to begin with. But as a baseline for this case that can easily be tested (beyond some "microbenchmarks", as shown in [the linked PR](https://github.com/CesiumGS/cesium/pull/12176#issuecomment-2379150741)), I ran this test with the 120x120 asset, and there doesn't seem to be a significant performance impact.

---

# Author checklist


ggetz:
Thanks @javagl!

@jjspace Can you please review?

packages/engine/Source/Core/addAll.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {Array|undefined} source The source array
+ * @param {Array} target The target array
+ *
+ * @example
+ * const target = [ 0, 1, 2 ];
+ * const source = [ 3, 4, 5 ];
+ * Cesium.addAll(source, target);
+ * // The target is now [ 0, 1, 2, 3, 4, 5 ]
+ */
+function addAll(source, target) {
```
jjspace, line 27: I don't think I like the name `addAll`. I don't feel it's descriptive enough to understand what it does without looking at the documentation or the function itself. `addAllToArray` or maybe `combineArrays` could be better? Or, given my previous [comment](https://github.com/CesiumGS/cesium/pull/12176#issuecomment-2379407987) on this topic with this essentially doing what [`Array.concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) does, maybe `concatInto` or `concatIntoArray` would be more accurate?

packages/engine/Source/Core/addAll.js
```diff
@@ -0,0 +1,41 @@
...
+  if (!defined(source)) {
+    return;
+  }
+  const s = source.length;
+  if (s === 0) {
```
jjspace, line 32: it's a nitpick but in general I think we should avoid single letter variable names outside the looping variables `i`, `j`, `k`, and sometimes `l`.

Maybe it's just personal preference but I find this a little more readable even if it is more verbose
```js
  const sourceLength = source.length;
  if (sourceLength === 0) {
    return;
  }
  const targetLength = target.length;
  target.length += sourceLength;
  for (let i = 0; i < sourceLength; i++) {
    target[targetLength + i] = source[i];
  }
```

packages/engine/Source/Core/addAll.js
```diff
@@ -0,0 +1,41 @@
...
+ * @private
+ *
+ * @param {Array|undefined} source The source array
+ * @param {Array} target The target array
+ *
+ * @example
+ * const target = [ 0, 1, 2 ];
+ * const source = [ 3, 4, 5 ];
+ * Cesium.addAll(source, target);
+ * // The target is now [ 0, 1, 2, 3, 4, 5 ]
```
jjspace, line 25: Nitpick on semantics, but this time I do think it actually matters. When I first glanced at this code I expected the order of the arguments to be reversed: `addAll(target, source)`.
This is largely because I am comparing this to similar functions of `target.push(...source)` or `target.concat(source)` etc. In all the builtin array functions the "result" is on the left and the "source" is on the right. 
Another way I think about it is that this is replacing the syntax `[...target, ...source]` to put all the target elements first and the source elements second. So the target variable should be the first argument. Even in this example you're doing `addAll([ 3, 4, 5 ], [ 0, 1, 2 ]) = [ 0, 1, 2, 3, 4, 5 ]`
(Maybe "source" and "target" are the wrong names in the first place)

I think swapping them also aligns better with the idea that the "source" could be `null`, `undefined` or `[]`. Then the first argument is always defined even if the second only sometimes is.

packages/engine/Source/Core/addAll.js
```diff
@@ -0,0 +1,41 @@
+import defined from "./defined.js";
+
+/**
+ * Adds all elements from the given source array to the given target array.
+ *
+ * If the `source` is `undefined`, then nothing will be done. Otherwise,
```
jjspace, line 6:
```suggestion
 * If the <code>source</code> is <code>undefined</code>, then nothing will be done. Otherwise,
```

I know we're not generating docs for this because it's `@private` but something I always forget about too is that JSDoc comments are not markdown (as much as I wish they were)

jjspace (changes requested):
I tested with the asset you provided and did confirm it works on this branch and not in `main`. I did have a few comments about the new function itself though.

packages/engine/Source/Core/addAll.js
javagl: I nearly espected that, and can change it accordingly. 
But regarding the variable name `1`: Here is an animated GIF of `l` and `1`, switching every second:

Cesium L

packages/engine/Source/Core/addAll.js
javagl: I also hesitated here. Leaving the source as the last parameter would allow omitting it, as in
`addAll(target);`
which is probably not ideal. 

But in general, I'm open to changing it, and embarassingly have to point out that commit https://github.com/CesiumGS/cesium/pull/12361/commits/c00243a948e5a8dd99a0fe6bdd5d91d5d60bbbb2 ...

packages/engine/Source/Core/addAll.js
javagl: No _really_ strong opinions, just things that I thought about:

- In the linked (soon to be obsolete) PR, I initially called that function `pushAll`, but I wanted to emphasize that it does exactly not `push`...
- Any name that involves `concat` could easily be confused with `Array.concat`, and (as discussed in the other PR), that has a completely different semantic: People might expect it to return a concatenation. I think it's important to make clear that it operates _in-place_, and _on_ the given target array
- `combine...` could be OK (_maybe_ people would expect it to return something there, but not necessarily)

Maybe something like `appendAll` or `append(All?To?)Array` or so ...?

packages/engine/Source/Core/addAll.js
javagl: From my all-time favorite [presentation on API design](https://www.youtube.com/watch?v=aAb7hSCtvGw):

Cesium Parameter order

packages/engine/Source/Core/addAll.js
javagl: Renamed to `sourceLength` and `targetLength`

packages/engine/Source/Core/addAll.js
javagl: Changed to `addAllToArray` - sounds reasonable and is long and clear enough for a global function.

packages/engine/Source/Core/addAll.js
javagl: Swapped the parameters. Fortunately, someone [posted the Regex for that](https://stackoverflow.com/a/71259650), which is to ...
replace `(addAllToArray)\(([^,]+), *([^,]+)\)`
with `$1($3, $2)`

ggetz:
@javagl Is this ready for review and potential merge again?

ggetz:
> A tangential connection to https://github.com/CesiumGS/cesium/issues/8359 : One of the rules/transforms of Lebap is...

I think this is a good reason to be targeted about modernization pushes we do, and to not blanket apply transforms: We need to ensure that there are no performance implications when we do so.

> When they didn't use push.apply, they aren't caught in this PR, and could still benefit in terms of readabiltiy and maybe also performance, due to the pre-allocation of the required .length

I agree at minimum, this is out of scope with the original scope of this PR and there's no need to do it here. Unless we confirm that there is a performance implication to motivate any other use case, I think we're getting into bikeshedding territory and we should hold off.

jjspace (commented):
This is looking good now. Just 2 comments:

1. Should there be a section added to our Coding Guide (or some other guide) explaining why we want to avoid `.push.apply`? maybe linking back to the discussion in this PR or the issue?
2. Optional change but should we actually restrict usage of this function through eslint? It's possible to do that through the [`no-restricted-syntax`](https://eslint.org/docs/latest/rules/no-restricted-syntax) rule with either a warning or a full error. @ggetz thoughts on this?
```js
rules: {
  "no-restricted-syntax": [
    "warn",
    {
      // Link to issue here
      selector:
        "CallExpression[callee.object.property.name=push][callee.property.name=apply]",
      message:
        "We prefer to avoid Array.push.apply() in favor of addAllToArray()",
    },
  ],
},
```

javagl:
@jjspace No strong opinion from my side, just minor thoughts:

As of [my comment above](https://github.com/CesiumGS/cesium/pull/12361#issuecomment-2588066276), one 'Lebap' modernization transform allows converting `push.apply` to `...spread` automatically. So one could make a case that the Coding Guide _could_ just have a hint like "Write modern, idiomatic code". **But**... the `...spread` operator could similarly be discouraged. So these points may indeed be worth mentioning. (I'm pretty sure that even many experienced developers don't have that 'call stack' issue on the radar...). 

(BTW: I wasn't aware of the `no-restricted-syntax` and that it could be defined so easily (well... "easy", when you know the _syntax_ for it 😁 ))

Maybe we should just collect some ideas in some "Possible updates for the Coding Guide" issue, to better keep track of that? (I mean, this could also include other "modernization" hints, like "Don't do `array.indexOf(element) !== -1`, but `array.includes(element)`" or so....)

jjspace:
> (I'm pretty sure that even many experienced developers don't have that 'call stack' issue on the radar...).

@javagl I think this is exactly why it would be worth calling out in the coding guide somewhere. I don't think we need to add a section on modernization in general. It would be more on this as a performance topic like the original issue.

> Maybe we should just collect some ideas in some "Possible updates for the Coding Guide" issue, to better keep track of that?

If you have other immediate ideas of things to add then it could be worth opening an issue but I think it's more likely to actually get in if we add it to a PR directly. I think something should be added alongside this change just focused on the pitfall of `Array.push.apply`. I'm not entirely sure what heading it should go under so maybe a new one makes sense even if this is the only topic in it for now. Maybe just a general "Performance" section

I also do think we should add the eslint rule but only as a warning. This will ensure it's automated and shows up in the actual code instead of relying on everyone remembering everything in the coding guide at all times. The codeblock above will achieve this, I think it can be added to the `files: ["packages/**/*.js"...` section. Just please make sure to leave comments there explaining why and point to the original issue.

Once those changes are in I think this should be good to merge

javagl:
@jjspace I added that linting rule as you suggested. I tried to elaborate the rule/warning text a bit:

> Avoid Array.push.apply(). Use addAllToArray() for arrays of unknown size, or the spread syntax for arrays that are known to be small 

The spread syntax _can_ be nice, modern, and convenient, and for something like  
`nodes.push([...binaryTree.children]); // Binary = At most 2 elements`
they can be fine, so I tried to give a "reasonable" guidance in that message.

For some potential further updates of the Coding Guide (including `push.apply` hints), I opened https://github.com/CesiumGS/cesium/issues/12452

Issue #10284, opened 2022-04-11
Support sparse accessors in ModelExperimental


ptrgags:
The old `Model` class didn't handle [sparse accessors](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#sparse-accessors) from the glTF spec. For better spec compliance, the new `ModelExperimental` should add this.

Not sure the exact scope of how to implement this, but some thoughts:
* Certainly `GltfLoader` (or related loaders) would need more parsing logic for sparse accessors
* If it's possible, handle this completely at load time so ModelComponents and runtime code such as the `GeometryPipelineStage` don't need to worry about whether the accessor was sparse or dense.

j9liu:
This would also be useful for animation with morph targets, since the weights are packed per keyframe. It gets even more cumbersome to pull the data when `CUBICSPLINE` interpolation is involved, since there are three values per morph weight per keyframe.

ggetz:
In https://github.com/CesiumGS/cesium/issues/11961, @mramato pointed out a glTF which fails to load, and @javagl Identified lack of support for sparse accessors as the cause for the failure.

> [house.glb.txt](https://github.com/CesiumGS/cesium/files/15138348/house.glb.txt) is a draco encoded glTF model that loads fine in all viewers I tried except for CesiumJS. It validates cleanly with no errors.

> When loaded into CesiumJS, it reports:

```
RuntimeError: Failed to load model: ../../SampleData/models/house.glb
Failed to load glTF
One of options.bufferViewId and options.draco must be defined.
```

> ... the way CesiumJS is failing here is not very useful and a clear error message or warning about lack of sparse accessors would be the least we could do. But it definitely bugs me that there is part of the glTF standard we seem to not support, especially since tools like gltf-transform's optimize function produces sparse accessors where applicable.

> @ggetz @lilleyse What is the level of effort to support them?

ggetz:
> What is the level of effort to support them?

Correct me if I'm wrong @lilleyse, but supporting sparse accessors is a matter of updating the loader parsing and shouldn't be complex to support.

lilleyse:
Yeah it seems pretty straightforward. The simplest approach would be to expand the sparse accessor at load time.

javagl:
An example that I created for https://github.com/CesiumGS/cesium/pull/12379 : 

[...]

PR #12379, opened 2024-12-18
Add support for sparse accessors

javagl:
Opening this as a draft for addressing https://github.com/CesiumGS/cesium/issues/10284 

# Description

The [cx20 `gltf-test`](https://github.com/cx20/gltf-test) repo is an invaluable resource for checking the compatibility of glTF viewers with the glTF specification. CesiumJS is among the ones with the larger number of ✅'s. But there is still that `SimpleSparseAccessor` example, which uses a pretty fundamental glTF concept - namely, sparse accessors - that shows a ❌...

This PR is a **DRAFT**, and the state is kinda messy, but it already loads that sample:

Cesium Sparse 0001

A summary:

A sparse accessor is an accessor that has a `sparse` property. The `sparse` object contains two objects, namely `indices` and `values`. These are actually "small accessors": They refer to a `bufferView` and have a `byteOffset`, one has a `componentType`. The type is taken from the _containing_ accessor.

The `loadFromBufferView` function of `GltfVertexBufferLoader` checks whether the accessor whose data is supposed to be loaded has the `sparse` property. If it does, then it calls `handleSparseAccessor`, which does the main work:

- The function calls a new function `loadDependencyBufferViewTypedArray` which loads the required buffer views for the indices and values
- These buffer views are used for creating typed arrays of the indices and values. (This is a bit quirky - see notes below...)
- The `handleSparseAccessor` finally performs the sparse substitution. It just writes the substituted values into the typed array of the _original_ accessor data. The pseudocode for that is just `accessorData[indices[i]] = values[i]`.

TODO:

The functionality for creating typed arrays from buffer views (based on "accessor information") is required elsewhere - namely, for _standard_ index- and attribute accessors After all, the `indices` and  `values` from the `sparse` object basically _are_ just "accessors". For now, I just added some helper functions like `GltfIndexBufferLoader.createIndicesTypedArrayFromBufferViewTypedArray` for that 🤪 It's not clear what could be the best place for offering these. They could be standalone functions, or maybe 'static' functions in `GltfLoader` ...?

Beyond that: Tests, tests, tests, comments, comments, comments, ....

## Issue number and link

https://github.com/CesiumGS/cesium/issues/10284 

## Testing plan

Load the `SimpleSparseAccessor` model (and other models with sparse accessors), and check that they are displayed properly.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/GltfDefaultVertexBufferLoader.js
+++ b/packages/engine/Source/Scene/GltfDefaultVertexBufferLoader.js
@@ -1,0 +1,96 @@
+import Check from "../Core/Check.js";
+import defaultValue from "../Core/defaultValue.js";
+import defined from "../Core/defined.js";
+import GltfVertexBufferLoader from "./GltfVertexBufferLoader.js";
+import ResourceLoaderState from "./ResourceLoaderState.js";
+
+/**
+ * Loads a vertex buffer from a glTF buffer view.
+ * <p>
+ * Implements the {@link ResourceLoader} interface.
+ * </p>
+ *
+ * @alias GltfDefaultVertexBufferLoader
+ * @constructor
+ * @augments GltfVertexBufferLoader
+ *
+ * @param {object} options Object with the properties that are required for {@link GltfVertexBufferLoader}, and the following properties:
+ * @param {number} [options.bufferViewId] The bufferView ID corresponding to the vertex buffer.
+ *
+ * @private
+ */
+function GltfDefaultVertexBufferLoader(options) {
+  options = defaultValue(options, defaultValue.EMPTY_OBJECT);
+  GltfVertexBufferLoader.call(this, options);
+
+  const bufferViewId = options.bufferViewId;
+
+  //>>includeStart('debug', pragmas.debug);
+  Check.typeOf.object("options.bufferViewId", bufferViewId);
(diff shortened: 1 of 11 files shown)


javagl:
After having spent a considerable amount of time trying to wrap my head around what is happening in the `GltfLoader`, a short "heads up": At https://github.com/CesiumGS/cesium/blob/47b5d295facaca795381bf607b78cc60c643fd85/packages/engine/Source/Scene/GltfLoader.js#L1944 , a possible Draco-compression is detected. This draco object is passed down the whole call chain. At many places of this call chain, the code behaves _completely_ differently, depending on whether that `draco` object is defined or not. And at several places, the behavior is plainly wrong, including the (several) places where the message

> One of options.bufferViewId and options.draco must be defined.

that is emitted (as mentioned in the issue). I'd try to break up a few things here, and along the way, would try to add some documentation. And ... that sparse accessors thing ("Hal changing a light bulb") - I'll try to keep that in mind.

javagl:
I thinkt that it is not possible to add support for sparse accessors without a _considerable_ refactoring of the glTF loader infrastructure. This means that adding support for sparse accessors is not a that "low-hanging fruit" that I expected it to be, given that they are actually a fairly trivial feature of glTF that, at its core, can be implemented with a few lines of code.

Issue #12389, opened 2024-12-23
Type error in ClippingPolygon.equals


rruhdorfer:
### What happened?

The `static ClippingPolygon.equals()` method is typed wrong. 
The parameters `left` and `right` should be typed as `ClippingPolygon` not `Plane`.

### Reproduction steps

1. Create two ClippingPolygon
2. Try comparing them using ClippingPolygon.equals(left, right)
3. TypeScript will throw an error:

Argument of type 'ClippingPolygon' is not assignable to parameter of type 'Plane'.
  Type 'ClippingPolygon' is missing the following properties from type 'Plane': normal, distance

### Environment

Browser: -
CesiumJS Version: 1.124.0
Operating System: -

javagl:
The typing error looks uncontroversial for me, and I opened a PR at https://github.com/CesiumGS/cesium/pull/12394 to fix this.

Broader questions:

1. Usually, classes that have a `ClassName.equals` function also have a `ClassName.prototype.equals` function that _delegates_ to the first one. Should this be added here as well? (I'd be in favor of "yes")
2. Usually, for these `ClassName.equals` functions, **both** parameters are _optional_. There are very few exceptions to that, and  `Plane` (which this `equals` function was copied-and-pasted from) is one of these few. For some others (like `Spherical` or `OrientedBoundingBox`), the parameters _are_ optional, but not _declared_ to be optional via the `[brackets]`. This should be made more consistent...
3. What is "equality"? 🙂 

The last one refers to the fact that one could **very** reasonably expect the following to print `true`:
```javascript
const polygonA = new Cesium.ClippingPolygon({
  positions: [
    new Cesium.Cartesian3(0, 1, 2),
    new Cesium.Cartesian3(2, 3, 4),
    new Cesium.Cartesian3(4, 5, 6),
  ]
});
const polygonB = new Cesium.ClippingPolygon({
  positions: [
    new Cesium.Cartesian3(0, 1, 2),
    new Cesium.Cartesian3(2, 3, 4),
    new Cesium.Cartesian3(4, 5, 6),
  ]
});
console.log(Cesium.ClippingPolygon.equals(polygonA, polygonB));
```
But it prints `false`, because it is comparing arrays via `===`. I'd consider to 1. fix this, and 2. add an `equalsEpsilon` function, like it is done in similar classes...

javagl:
[...]

PR #12394, opened 2024-12-25
Fix for `ClippingPolygon` `equals` funciton

javagl:
# Description

Until now, this **only** fixes the JSDoc parameter type for `ClippingPolygon#equals` from `Plane` to `ClippingPolygon`.

See additional comments in the issue.

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12389

## Testing plan

For now, there's not much to test. With `main`, running `npm run build.ts` generated
```
export class ClippingPolygon {
...
    /**
     * Compares the provided ClippingPolygons and returns
    <code>true</code> if they are equal, <code>false</code> otherwise.
     * @param left - The first polygon.
     * @param right - The second polygon.
     * @returns <code>true</code> if left and right are equal, <code>false</code> otherwise.
     */
    static equals(left: Plane, right: Plane): boolean;
```
and now it generates
```
export class ClippingPolygon {
...
    /**
     * Compares the provided ClippingPolygons and returns
    <code>true</code> if they are equal, <code>false</code> otherwise.
     * @param left - The first polygon.
     * @param right - The second polygon.
     * @returns <code>true</code> if left and right are equal, <code>false</code> otherwise.
     */
    static equals(left: ClippingPolygon, right: ClippingPolygon): boolean;

```

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/ClippingPolygon.js
+++ b/packages/engine/Source/Scene/ClippingPolygon.js
@@ -117,8 +117,8 @@ ClippingPolygon.clone = function (polygon, result) {
  * Compares the provided ClippingPolygons and returns
  * <code>true</code> if they are equal, <code>false</code> otherwise.
  *
- * @param {Plane} left The first polygon.
- * @param {Plane} right The second polygon.
+ * @param {ClippingPolygon} left The first polygon.
+ * @param {ClippingPolygon} right The second polygon.
  * @returns {boolean} <code>true</code> if left and right are equal, <code>false</code> otherwise.
  */
 ClippingPolygon.equals = function (left, right) {


ggetz:
Looks good @javagl, thanks! Is this in a draft state for any particular reason?

Before we merge, we should update `CHANGES.md` with this fix since it does affect the typescript definitions.

javagl:
> I agree that point 2 could be a quick fix, and don't mind if you include them here. 

I did a quick pass over the search results of `.equals = function` and marked the parameters as optional using `[brackets]` where appropriate. These changes are all in https://github.com/CesiumGS/cesium/pull/12394/commits/50002d99b447110866c2639080baf3730ca7a9cf 

I also quickly skimmed over the diff of the generated typescript files, and it looked OK for me...

Cesium Equals Optional

(This is not a breaking change in terms of the usage. For cases like "inheritance", we could argue...)

ggetz (approved):

ggetz:
Thanks for the cleanup @javagl! This all looks great to me.

ggetz:
Thanks @javagl!

PR #12405, opened 2025-01-02
Rename release schedule

jjspace:
# Description

Simply renames the release schedule to make it clear what it actually is. We have not had a slackbot for a long time so there's no need for it's config anymore.


ggetz:
Thanks @jjspace! The link in `Documentation/Contributors/ReleaseGuide/README.md` will need to be updated to this new location.

Also what do you think about moving this to `Documentation/Contributors` since this is only intended to be used by contributors or maintainers?

jjspace:
Good call, I relocated it to be next to the release guide itself since I think that makes the most sense.

I also reordered it `Date -> User` as that's probably more likely the order of information we'll usually be looking for (at least I do)

Should be good to go

ggetz (approved):

ggetz:
Perfect, thanks @jjspace!

ggetz:
@MonicaWamsley Just an FYI that the release schedule has been moved to https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/ReleaseGuide/ReleaseSchedule.md, in case you need to update any processes or documentation.

Issue #12403, opened 2025-01-02
Error updating 3DTiles transformation matrix.


syzdev:
### What happened?

As described in Example [3D Tiles Adjust Height](https://sandcastle.cesium.com/?src=3D%20Tiles%20Adjust%20Height.html), updating the `modelMatrix` of the model can complete the spatial position transformation, and sometimes the model needs to be restored to the original position, which will cause errors:
> TypeError: result.computeBoundingVolumes is not a function

The error is located in the following function:
https://github.com/CesiumGS/cesium/blob/47b5d295facaca795381bf607b78cc60c643fd85/packages/engine/Source/Scene/Cesium3DTile.js#L1739-L1768

After modifying the transformation matrix, `transform` and `initialTransform` are no longer equal, and `TileOrientedBoundingBox` is created. Later, when `transform` is equal to `initialTransform`, the `computeBoundingVolumes` function on the `TileOrientedBoundingBox` is called, but this function does not exist on the `TileOrientedBoundingBox`.

Related commit: https://github.com/CesiumGS/cesium/commit/55d6dfa8cb12b326676872e2163c642bfef41718

### Reproduction steps

https://sandcastle.cesium.com/?src=3D%20Tiles%20Adjust%20Height.html

1. Height -> Any number
2. Height -> 0

### Sandcastle example

https://sandcastle.cesium.com/?src=3D%20Tiles%20Adjust%20Height.html

### Environment

Browser: Latest version of Chrome
CesiumJS Version: 1.124
Operating System: Windows 11

syzdev:
@jjhembd Could you please triage this issue?

javagl:
This is a duplicate of https://github.com/CesiumGS/cesium/issues/12002 , where the triage already happened, and a fix was already suggested. @jjhembd Unless you scream 🛑 STOP, I'd create a PR with the change that you suggested in https://github.com/CesiumGS/cesium/issues/12002#issuecomment-2150805368

Issue #12002, opened 2024-05-28
Cesium3DTileset modelMatrix reset error


wonphll536:
### What happened?

I applied the transformation matrix to Cesium3DTileset and reapplied the initial value, but an error occurred.
![image](https://github.com/CesiumGS/cesium/assets/66579004/614e0693-a261-4277-a9ca-da0f1b939c90)

### Reproduction steps

1. Create Cesium3DTeilset
2. Clone the original modelMatrix
3. Apply any transformation matrix
4. Apply original modelMatrix ---> Error occurs after

I applied the above process to Sandcastle. If you wait a while after running, an error will occur.

### Sandcastle example

[...]

PR #12409, opened 2025-01-03
Allow resetting model matrix when tiles have regions

javagl:
# Description

When loading a tileset where the tiles contained bounding regions, changing the `modelMatrix` of the tileset, and then setting the `modelMatrix` back to its initial value, the update caused a

> TypeError: result.computeBoundingVolumes is not a function

A detailed description of what went wrong is in https://github.com/CesiumGS/cesium/issues/12002#issuecomment-2135572779

While creating a (pretty strict, dedicated) spec, I noticed that the [suggested fix](https://github.com/CesiumGS/cesium/issues/12002#issuecomment-2150805368) does not fully cover it. There are basically four cases to consider:

- When the transform **did** change (compared to the initial one)
  - When the `result` **is** a `TileOrientedBoundingBox`, then reuse it
  - When the result is **not** a `TileOrientedBoundingBox`, then create a new one
- When the transform did **not** change (compared to the initial one)
  - When the `result` **is** a `TileBoundingRegion`, then reuse it
  - When the result is **not** a `TileBoundingRegion`, then create a new one

(It's hard to be 100% sure that all cases can happen in practice, but easy to handle them generically)

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12002
Fixes https://github.com/CesiumGS/cesium/issues/12403

## Testing plan

The issue descriptions contain examples that caused the crash (basically the same example, though). 

Also, I added a dedicated spec for this. 

**NOTE**: The whole spec is currently wrapped into a `expect ... not.toThrow`. This is the actual _change_ that is done here. One could consider this as the "baseline", though, and remove this check, and only use the expect's for the respective bounding volume types.

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.126 - 2025-02-03
+
+#### @cesium/engine
+
+##### Fixes :wrench:
+
+- Fixed error when resetting `Cesium3DTileset.modelMatrix` to its initial value. [#12409](https://github.com/CesiumGS/cesium/pull/12409)
+
 ### 1.125 - 2025-01-02
 
 #### @cesium/engine
--- a/packages/engine/Source/Scene/Cesium3DTile.js
+++ b/packages/engine/Source/Scene/Cesium3DTile.js
@@ -1729,28 +1729,57 @@ function createBoxFromTransformedRegion(
 }
 
 /**
+ * Creates a TileBoundingVolume from the given region and transform
+ * information.
+ *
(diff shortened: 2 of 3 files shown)


jjhembd:
@javagl thanks for the fix! The code change makes sense, and the Sandcastle now works for me. I am still trying to run the specs--need to fix a configuration somewhere on my new laptop.

Can you add a note to CHANGES.md?

jjhembd (approved):
I fixed my configuration and the tests ran through successfully.

Thanks again @javagl!

Issue #12297, opened 2024-11-08
Voxel API should conform to Cesium3DTileset interface


ggetz:
Right now `Cesium3DTilesVoxelProvider` and other voxel classes are distinct from how other 3D tilesets are used in CesiumJS:

```js
const provider = await Cesium.Cesium3DTilesVoxelProvider.fromUrl(
  "..."
);
const voxelPrimitive = viewer.scene.primitives.add(
  new Cesium.VoxelPrimitive({
    provider: provider,
  })
);
```

The goal is eventually to bring voxels into 3D Tiles as an extension. As such, we want to treat it in the API like like other 3D Tilesets, and be able to operate using familiar paradigms. Furthermore, [the properties and functions associated with other 3D Tilesets](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html) should be available for voxel tilesets too.

lukemckinstry:
Question regarding the approach for reproducing the 3DTileset events for Voxels. Looking at the generated sample dataset we only have a single LOD at 0/0/0. Is this always the case now? Or is this due to the specifics on the source data and/or the specific tiler used?

```
└── Temperaturmodell-ellipsoid
    ├── subtrees
    │   └── 0
    │       └── 0
    │           └── 0
    │               └── 0.subtree
    ├── tilers-stats.json
    ├── tiles
    │   └── 0
    │       └── 0
    │           └── 0
    │               └── 0.voxel
    └── tileset.json
```

It seems like the `3DTileset` event properties either operate on a per-tile or per-tileset basis. If `VoxelPrimitive` only uses info for a single LOD, does it make sense to only implement the per-tileset events and not the per-tile events?

Many of the 3DTileset events track progress relating to individual tiles which I believe are representing a specific LOD. 
 - Per-tile events: `loadProgress, tileLoad, tileFailed, tileVisible, tileUnload` - do not seem to have an equivalent in the current voxel setup.
 - `allTilesLoaded` makes sense to keep, and fire when the `VoxelPrimitive` has loaded (it could be a better description to call it primitiveLoaded, but allTilesLoaded keeps the API naming consistent). `initialTilesLoaded` seems to be redundant with allTilesLoaded for voxels and no necessary to implement separately.

jjhembd:
[...]

PR #12430, opened 2025-01-10
Add tile loading events for Voxels

lukemckinstry:
# Description

Takes the following events from Cesium3DTileset and adds them to VoxelPrimitive
 - `allTilesLoaded`
 - `initialTilesLoaded`
 - `loadProgress`
 - `tileLoad`
 - `tileUnload`

Also makes `VoxelPrimitive` use `Cesium3DTilesetStatistics` class, but only makes use of 2 of the stats in the class so far.

## ToDo
 
 - [x] Did not carry over these event types
   - [x] tileFailed - issues described in this comment https://github.com/CesiumGS/cesium/issues/12297#issuecomment-2578516229
   - [x] tileVisible - not sure is there is an equivalent to `selectedTiles` (from `Cesium3DTileset`) in `VoxelPrimitive`, if so I believe this would be simple.
 - [x] Still need to add tests, but these look straighforward to carry over from `Cesium3dTilesetSpec` to `VoxelPrimitiveSpec`.
 - [x] move code to call events out of `VoxelTraversal.printDebugInformation`.

## Issue number and link

related to #12297 

## Testing plan

Many events currently fire inside of `VoxelTraversal.printDebugInformation`. Enable this but comment out the last line of the function which logs the debug output so you can read the output from listeners you set up to listen to the new events.

# Author checklist


ggetz:
@lukemckinstry 

> tileVisible - not sure is there is an equivalent to selectedTiles (from Cesium3DTileset) in VoxelPrimitive, if so I believe this would be simple.

There are definitely visibility checks for the tiles (nodes), it's just not in a neat array like `selectedTiles` are unfortunately. I believe all the visibility checks happen within `generateOctree`– Once the `else` condition is reached in `buildOctree` and the function no longer recurses, I think that is the highest resolution node which has been marked as renderable.

lukemckinstry:
> There are definitely visibility checks for the tiles
I see now where tileVisible would fire within the generateOctree function. 

For Cesium3DTileset tileVisible,  the `Tile` is passed to the event listener so the user can glean info and modify it before the next render. For VoxelPrimitive, the spatialNode is not part of the public API, so do we want to pass it to event listener in this same way?

ggetz:
> For Cesium3DTileset tileVisible, the Tile is passed to the event listener so the user can glean info and modify it before the next render. For VoxelPrimitive, the spatialNode is not part of the public API, so do we want to pass it to event listener in this same way?

It may be worth considering making the nodes part of the public API, and if they and `Cesium3DTile` should conform to some interface. Probably something worth noting in #12297, but a bit much to bite off in this PR. For now, I think it's fine not to pass anything.

lukemckinstry:
This is ready for review. @ggetz  @jjhembd 
ToDo item
 - [ ] I do not have a unit test yet for the `tileUnload` event. There is an existing `xit`ed out spec covering the voxel unload process. This seems like it would be the right place to test the tileUnload event.

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -450,6 +457,136 @@ function VoxelPrimitive(options) {
...
+   *     console.log('An error occurred loading tile.');
+   * });
+   */
+  this.tileFailed = new Event();
+
+  /**
+   * The event fired to indicate that a tile's content was unloaded.
+   *
+   * @type {Event}
+   * @default new Event()
```
ggetz, line 517:
```suggestion
```

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -450,6 +457,136 @@ function VoxelPrimitive(options) {
...
+   * This event is fired at the end of the frame after the scene is rendered.
+   * </p>
+   *
+   * @type {Event}
+   * @default new Event()
+   *
+   * @example
+   * voxelPrimitive.loadProgress.addEventListener(function(numberOfPendingRequests, numberOfTilesProcessing) {
+   *     if ((numberOfPendingRequests === 0) && (numberOfTilesProcessing === 0)) {
+   *         console.log('Stopped loading');
```
ggetz, line 545: Nitpick:

```suggestion
   *         console.log('Finished loading');
```

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -450,6 +457,136 @@ function VoxelPrimitive(options) {
...
+   * <p>
+   * The number of pending tile requests, <code>numberOfPendingRequests</code>, and number of tiles
+   * processing, <code>numberOfTilesProcessing</code> are passed to the event listener.
+   * </p>
+   * <p>
+   * This event is fired at the end of the frame after the scene is rendered.
+   * </p>
+   *
+   * @type {Event}
+   * @default new Event()
```
ggetz, line 540:
```suggestion
```

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -450,6 +457,136 @@ function VoxelPrimitive(options) {
...
+
+  /**
+   * The event fired to indicate that all tiles that meet the screen space error this frame are loaded. The voxel
+   * primitive is completely loaded for this view.
+   * <p>
+   * This event is fired at the end of the frame after the scene is rendered.
+   * </p>
+   *
+   * @type {Event}
+   * @default new Event()
```
ggetz, line 562:
```suggestion
```

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -450,6 +457,136 @@ function VoxelPrimitive(options) {
...
+
+  /**
+   * The event fired to indicate that all tiles that meet the screen space error this frame are loaded. This event
+   * is fired once when all tiles in the initial view are loaded.
+   * <p>
+   * This event is fired at the end of the frame after the scene is rendered.
+   * </p>
+   *
+   * @type {Event}
+   * @default new Event()
```
ggetz, line 579:
```suggestion
```

packages/engine/Source/Scene/VoxelTraversal.js
```diff
@@ -443,34 +465,33 @@ function requestData(that, keyframeNode) {
...
-  });
+  const promise = provider.requestData(requestOptions);
 
   if (defined(promise)) {
     that._simultaneousRequestCount++;
     keyframeNode.state = KeyframeNode.LoadState.RECEIVING;
     promise.then(postRequestSuccess).catch(postRequestFailure);
   } else {
     keyframeNode.state = KeyframeNode.LoadState.FAILED;
+    that._primitive.tileFailed.raiseEvent();
```
ggetz, line 494: This may be a question beyond the scope of this PR, but when a provider returns `undefined` instead of a promise, is that actually a failure state? Typically when `undefined` is returned from a function like this, it signals that the request couldn't be scheduled this frame and will be tried again next frame.

CC @jjhembd

ggetz (commented):
Looking good so far @lukemckinstry! A few comments, mostly around the inline documentation.

ggetz:
@jjhembd Would you be able to provide any tips on testing the unload event since you were recently looking into that part of the code?

ggetz:
Bump @jjhembd for two questions regarding voxel content above.

@lukemckinstry Let me know when this is ready for another pass.

packages/engine/Source/Scene/VoxelTraversal.js
jjhembd: For both `Cesium3DTilesVoxelProvider` and the procedural providers in the Sandcastles, an `undefined` return happened when the tile didn't exist. I agree that this is not really expected behavior.

In https://github.com/CesiumGS/cesium/pull/12432 I changed the providers to return a rejected Promise if the tile doesn't exist.

ggetz: I added a few comments in https://github.com/CesiumGS/cesium/pull/12432 that mentioned `undefined` may still be a valid return value that needs to be accounted for. But, yes, it should never be a failure state.

lukemckinstry: > In https://github.com/CesiumGS/cesium/pull/12432 I changed the providers to return a rejected Promise if the tile doesn't exist.

This change here as described above https://github.com/CesiumGS/cesium/blob/41dbe65ce3746a4bb80962698ee49b6af0eca6bd/packages/engine/Source/Scene/VoxelTraversal.js#L429

This simplifies the logic on whether to raise the tileLoad or tileFailed event
Something like this at the end of the `loadTileSuccess` would be all that is needed.
```
    if (keyframeNode.state === KeyframeNode.LoadState.PROCESSING) {
      that._primitive.tileLoad.raiseEvent();
    }
    that._primitive.tileFailed.raiseEvent();
```

ggetz (approved):

Issue #12368, opened 2024-12-10
Support for newer `EXT_primitive_voxels` in CesiumJS


ggetz:
`EXT_primitive_voxels` [is in the process of being revised](https://github.com/CesiumGS/glTF/pull/69) to be more consistent with 3D Tiles 1.1 and the existing glTF ecosystem.

We'll want to update the CesiumJS implementation of voxel to support this.

PR #12432, opened 2025-01-11
Support voxel tilesets containing glTF tiles using the `EXT_primitive_voxels` extension

jjhembd:
# Description

This PR updates `VoxelPrimitive` and related classes to support tilesets using the [proposed update to `EXT_primitive_voxels`](https://github.com/CesiumGS/glTF/pull/69).

Along the way, the internal APIs were cleaned up and made more consistent with `Cesium3DTileset`, to prepare for work on https://github.com/CesiumGS/cesium/issues/12297.

Key code changes include:
- The `VoxelProvider` interface always returns a `VoxelContent`, rather than an array of metadata. This affects not just `Cesium3DTilesVoxelProvider`, but procedural data as well--see the Sandcastles for an example.
- `VoxelContent` can now be constructed before the data is ready, via the asynchronous `VoxelContent.fromGltf` method, and updated updated via `VoxelContent.prototype.update`. This makes it more similar to `Cesium3DTileContent`. For synchronous construction (e.g. with procedural data), use `VoxelContent.fromMetadataArray`.
- `VoxelTraversal` was refactored to use smaller functions that are structured more like `Cesium3DTilesetTraversal`. For example, `loadAndUnload` was broken into `selectKeyframeNodes` (similar to `Cesium3DTilesetTraversal.selectTiles`) and `updateTiles`. The `updateTiles` loop is where we call `VoxelContent.update` to process glTFs.

Test datasets were also updated to the new format. Note that the tilesets in `Apps/SampleData/Cesium3DTiles/Voxel/` are a copy of `Specs/Data/Cesium3DTiles/Voxel`.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/12368.

Related: https://github.com/CesiumGS/cesium/issues/12297

## Testing plan

Load the three Voxel-related Sandcastles and verify all datasets load and functionalities are unchanged.
- http://localhost:8080/Apps/Sandcastle/index.html?src=Voxels.html
- http://localhost:8080/Apps/Sandcastle/index.html?src=Voxels%20in%203D%20Tiles.html
- http://localhost:8080/Apps/Sandcastle/index.html?src=Voxel%20Picking.html

# Author checklist


packages/engine/Source/Scene/VoxelContent.js
```diff
@@ -1,167 +1,202 @@
...
+function processAttributes(attributes, primitive) {
+  const { names, types, componentTypes } = primitive.provider;
+  const data = new Array(attributes.length);
+
+  for (let i = 0; i < attributes.length; i++) {
+    // The attributes array from GltfLoader is not in the same order as
+    // names, types, etc. from the provider.
+    // Find the appropriate glTF attribute based on its name.
+    // Note: glTF custom attribute names are prefixed with "_"
+    const name = `_${names[i]}`;
```
j9liu, line 155: Could this change to retrieve the property names from the glTF's property attribute? The attribute for a `temperature` property might be named `_TEMPERATURE`, which will mess with the `find` function below.

packages/engine/Source/Scene/Model/ModelUtility.js
```diff
@@ -350,18 +350,20 @@ ModelUtility.supportedExtensions = {
   EXT_mesh_features: true,
   EXT_mesh_gpu_instancing: true,
   EXT_meshopt_compression: true,
+  EXT_primitive_voxels: true,
```
j9liu, line 353: Oops forgot to mention, could we add `EXT_implicit_cylinder_region` and `EXT_implicit_ellipsoid_region` to this list? They're not required for our implementation but they'll be included in generated cylinder / ellipsoid voxels

https://github.com/CesiumGS/glTF/tree/ext-primitive-voxels-revisions/extensions/2.0/Vendor/EXT_implicit_cylinder_region
https://github.com/CesiumGS/glTF/tree/ext-primitive-voxels-revisions/extensions/2.0/Vendor/EXT_implicit_ellipsoid_region

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -47,54 +47,145 @@ import VoxelShapeType from "./VoxelShapeType.js";
...
+   * @readonly
+   */
   this.globalTransform = undefined;
 
-  /** @inheritdoc */
+  /**
+   * Gets the {@link VoxelShapeType}
+   *
+   * @memberof Cesium3DTilesVoxelProvider.prototype
+   * @type {VoxelShapeType}
```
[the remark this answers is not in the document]
ggetz: Whatever type we end up used here does need to be consistent with the VoxelProvider interface.

CHANGES.md
```diff
@@ -7,6 +7,17 @@
 #### Breaking Changes :mega:
 
 - Changed behavior of `DataSourceDisplay.ready` to always stay `true` once it is initially set to `true`. [#12429](https://github.com/CesiumGS/cesium/pull/12429)
+- Updated `Cesium3DTilesVoxelProvider` to load glTF tiles using the new [`EXT_primitive_voxels` extension](https://github.com/CesiumGS/glTF/pull/69). Tilesets using the previous custom JSON format are no longer supported. [#12432](https://github.com/CesiumGS/cesium/pull/12432)
```
ggetz, line 10: If you want to provide a but of motivation behind the change, consider:

```suggestion
- Updated `Cesium3DTilesVoxelProvider` to load glTF tiles using the [`EXT_primitive_voxels` extension](https://github.com/CesiumGS/glTF/pull/69) to more closely align with the rest if the 3D Tiles ecosystem. Tilesets using the previous custom JSON format are no longer supported. [#12432](https://github.com/CesiumGS/cesium/pull/12432)
```

packages/engine/Source/Scene/VoxelProvider.js
```diff
@@ -214,17 +203,17 @@ Object.defineProperties(VoxelProvider.prototype, {
 });
 
 /**
- * Requests the data for a given tile. The data is a flattened 3D array ordered by X, then Y, then Z.
- * This function should not be called before {@link VoxelProvider#ready} returns true.
- * @function
+ * Requests the data for a given tile.
+ *
+ * @private
```
ggetz, line 208: Since this is a public interface which defines a "contract", I'm curious why this method should be private. For instance, `TerrainProvider.prototype.requestTileGeometry` is not private. 

I think it makes sense to say in the docs that this function should not be called directly, but I think making it explicitly private goes against the purpose of an interface.

packages/engine/Source/Scene/VoxelProvider.js
```diff
@@ -214,17 +203,17 @@ Object.defineProperties(VoxelProvider.prototype, {
...
  *
  * @param {object} [options] Object with the following properties:
  * @param {number} [options.tileLevel=0] The tile's level.
  * @param {number} [options.tileX=0] The tile's X coordinate.
  * @param {number} [options.tileY=0] The tile's Y coordinate.
  * @param {number} [options.tileZ=0] The tile's Z coordinate.
  * @privateparam {number} [options.keyframe=0] The requested keyframe.
- * @returns {Promise<Array[]>|undefined} A promise to an array of typed arrays containing the requested voxel data or undefined if there was a problem loading the data.
+ * @returns {Promise<VoxelContent>} A promise resolving to a VoxelContent containing the data for the tile.
```
ggetz, line 216: Again, I think `undefined` is still a valid return type here if the request needs to retried next frame.

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -47,54 +47,145 @@ import VoxelShapeType from "./VoxelShapeType.js";
...
+   */
   this.maximumTileCount = undefined;
 
   this._implicitTileset = undefined;
   this._subtreeCache = new ImplicitSubtreeCache();
 }
 
 /**
- * Creates a {@link VoxelProvider} that fetches voxel data from a 3D Tiles tileset.
+ * Creates a {@link Cesium3DTilesVoxelProvider} that fetches voxel data from a 3D Tiles tileset.
```
ggetz, line 188: Could we add a short `@example` while we're in here?

ggetz (changes requested):
Thanks @jjhembd! 

Tests and examples are working well! I had a few comments around the API, but otherwise the code is looking good too.

Let me know once you get a chance to address those and @j9liu's comments.

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -402,27 +471,40 @@ async function getSubtreePromise(provider, subtreeCoord) {
...
+ *
+ * @private
+ *
+ * @param {object} [options] Object with the following properties:
+ * @param {number} [options.tileLevel=0] The tile's level.
+ * @param {number} [options.tileX=0] The tile's X coordinate.
+ * @param {number} [options.tileY=0] The tile's Y coordinate.
+ * @param {number} [options.tileZ=0] The tile's Z coordinate.
+ * @privateparam {number} [options.keyframe=0] The requested keyframe.
+ * @returns {Promise<VoxelContent>} A promise resolving to a VoxelContent containing the data for the tile.
```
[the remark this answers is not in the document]
jjhembd: Yes. We are not handling undefined promises in quite a few places, including `Cesium3DTile` and `ImplicitSubtree`. Also `preprocess3DTileContent` could be called with an `undefined` argument, and it is hard to figure out what the effect will be.

packages/engine/Source/Scene/VoxelProvider.js
jjhembd: This was based on making things more like `Cesium3DTileset`, which manages tile requests via private methods like `Cesium3DTileset.prototype.update`.
If we want to continue supporting custom providers, then I guess leaving this method public could make the documentation better for people trying to build a provider.

packages/engine/Source/Scene/VoxelProvider.js
jjhembd: OK but we currently aren't retrying voxel tile requests anywhere. This would require changes to VoxelTraversal.

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -444,23 +526,27 @@ Cesium3DTilesVoxelProvider.prototype.requestData = function (options) {
...
-    .then(function (voxelContent) {
-      return names.map(function (name) {
-        return voxelContent.metadataTable.getPropertyTypedArray(name);
-      });
-    });
+  const subtree = await getSubtree(that, subtreeCoord);
+  // NOTE: these two subtree methods are ONLY used by voxels!
+  const available = isSubtreeRoot
+    ? subtree.childSubtreeIsAvailableAtCoordinates(tileCoordinates)
+    : subtree.tileIsAvailableAtCoordinates(tileCoordinates);
```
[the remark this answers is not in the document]
jjhembd: If we pre-select the method (to explicitly call only one of them), we need to provide the `this` value via [Function.prototype.call](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call).
```javascript
  const isAvailable = isSubtreeRoot
    ? subtree.childSubtreeIsAvailableAtCoordinates
    : subtree.tileIsAvailableAtCoordinates;

  const available = isAvailable.call(subtree, tileCoordinates);
```

packages/engine/Source/Scene/VoxelContent.js
```diff
@@ -1,167 +1,224 @@
...
+    get: function () {
+      return this._ready;
+    },
+  },
+
+  /**
+   * The metadata for this voxel content.
+   * The metadata is an array of typed arrays, one for each field.
+   * The data for one field is a flattened 3D array ordered by X, then Y, then Z.
+   * TODO: use a MetadataTable?
```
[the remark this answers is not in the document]
jjhembd: I thought the array of flattened 3D arrays felt a little arbitrary and hard to verify. But `MetadataTable` is a private class, and `VoxelContent` is public. So `MetadataTable` is probably not the right solution.
I deleted the TODO.

packages/engine/Source/Scene/VoxelContent.js
```diff
@@ -1,167 +1,224 @@
...
-  const propertyTableIndex = chunks.json.voxelTable;
-  const propertyTableJson = chunks.json.propertyTables[propertyTableIndex];
+  // Ensures frames continue to render in requestRender mode while resources are processing
+  frameState.afterRender.push(() => true);
 
-  const content = new VoxelContent(resource);
+  if (!defined(loader)) {
+    this._ready = true;
+    return;
+  }
```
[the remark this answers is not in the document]
jjhembd: I think it's safer to assume the data won't be ready for rendering until the next frame. This will make the code less likely to be broken during a refactor.

By requiring at least one call to `VoxelContent.prototype.update` for every tile (even if procedural) we are effectively deferring the load to the GPU to the next frame after the data is received.

The way `updateKeyframeNodes` in `VoxelTraversal` is currently written, this particular case would be OK without the extra frame.
```javascript
    if (highPriorityKeyframeNode.state === KeyframeNode.LoadState.PROCESSING) {
      const { content } = highPriorityKeyframeNode;
      content.update(that._primitive, frameState);
      if (!content.ready) {
        continue;
      }
```

However, the reason for it being OK is pretty subtle (it requires a careful comparison of `VoxelContent` and `VoxelTraversal`), and likely to be missed when the code is modified. I think it's safer to keep the 1-frame delay for now, especially since https://github.com/CesiumGS/cesium/issues/12297 is likely to lead to additional tweaking of `VoxelTraversal`.

jjhembd:
Thanks for the feedback @j9liu and @ggetz! I think I addressed everything.

ggetz:
Thanks @jjhembd! Could you please merge in `main`?

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -47,54 +69,184 @@ import VoxelShapeType from "./VoxelShapeType.js";
...
-  /** @inheritdoc */
-  this.globalTransform = undefined;
-
-  /** @inheritdoc */
-  this.shape = undefined;
-
-  /** @inheritdoc */
-  this.minBounds = undefined;
+  const {
+    className,
```
ggetz, line 73: We should check `className` too since its required, correct?

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -47,54 +69,184 @@ import VoxelShapeType from "./VoxelShapeType.js";
...
+  Check.typeOf.object("types", types);
+  Check.typeOf.object("componentTypes", componentTypes);
+  Check.typeOf.string("shape", shape);
+  Check.typeOf.object("dimensions", dimensions);
+  //>>includeEnd('debug');
 
-  /** @inheritdoc */
-  this.maximumTileCount = undefined;
+  /**
+   * A transform from shape space to local space. If undefined, the identity matrix will be used instead.
```
ggetz, line 99:
```suggestion
   * A transform from shape space to local space.
```

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -47,54 +69,184 @@ import VoxelShapeType from "./VoxelShapeType.js";
...
+   *
+   * @memberof Cesium3DTilesVoxelProvider.prototype
+   * @type {Matrix4}
+   * @default Matrix4.IDENTITY
+   * @readonly
+   */
+  this.shapeTransform = shapeTransform;
+
+  /**
+   * A transform from local space to global space. If undefined, the identity matrix will be used instead.
```
ggetz, line 109:
```suggestion
   * A transform from local space to global space.
```

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
ggetz: The `undefined` case is handled in the constructor, but this property should never be `undefined`.

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
ggetz: Same comment as above.

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -385,6 +532,7 @@ async function getSubtreePromise(provider, subtreeCoord) {
   // Check one more time if the subtree is in the cache.
   // This could happen if there are two in-flight tile requests from the same
   // subtree and one finishes before the other.
+  // TODO: only allow one request per subtree
```
ggetz, line 535: Should this TODO be resolved here? Or should we document this in an issue?

packages/engine/Source/Scene/VoxelTraversal.js
```diff
@@ -417,61 +423,36 @@ function requestData(that, keyframeNode) {
...
-  } else {
-    keyframeNode.state = KeyframeNode.LoadState.FAILED;
-  }
+  };
+  that._simultaneousRequestCount++;
+  keyframeNode.state = KeyframeNode.LoadState.RECEIVING;
+  provider
+    .requestData(requestParameters)
+    .then(postRequestSuccess)
+    .catch(postRequestFailure);
```
ggetz, line 455: I think the (right now hypothetical) case where `provider.requestData` returns `undefined` can be accounted for by setting the state back to `UNLOADED`. But please let me know if I missed something.

```suggestion
  const promise = provider
    .requestData(requestParameters);
  if (!defined(promise)) {
    keyframeNode.state = KeyframeNode.LoadState.UNLOADED;
    return;
  }
  
  promise.then(postRequestSuccess)
    .catch(postRequestFailure);
```

ggetz (changes requested):
This is looking great @jjhembd! Just a few (hopefully) small comments, and then this should be good to go.

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
jjhembd: https://github.com/CesiumGS/cesium/issues/12457

jjhembd:
Thanks @ggetz, I think I incorporated all those comments. This should be ready for another look.

ggetz (approved):

ggetz:
Thanks @jjhembd! The code changes are looking good!

I did see an error when loading the ["Voxels"](https://ci-builds.cesium.com/cesium/voxel-gltf/Apps/Sandcastle/index.html?src=Voxels.html&label=Voxels) and ["Voxel Picking"](https://ci-builds.cesium.com/cesium/voxel-gltf/Apps/Sandcastle/index.html?src=Voxel%20Picking.html&label=Voxels) Sandcastle examples:

```
TypeError: Cannot read properties of undefined (reading 'availableLevels')

TypeError: Cannot read properties of undefined (reading 'availableLevels')
    at requestData (https://ci-builds.cesium.com/cesium/voxel-gltf/Build/CesiumUnminified/index.js:234793:54)
...
```

Could you please take a look?

jjhembd:
As discussed with @j9liu and @lilleyse, the current code is not reading the data from the glTF correctly: it is assuming a 3DTiles (z-up) ordering of the 3D buffer, whereas glTF uses a y-up ordering.

We hadn't yet noticed the error in our testing because our test datasets were also ordered incorrectly.

I'll move this PR back into draft for now, until the data ordering is fixed.

jjhembd:
@ggetz this is ready for another look.

I updated two of the Sandcastle datasets, `VoxelBox3DTiles` and `VoxelCylinder3DTiles`. These have data ordered according to the [proposed glTF extension](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_primitive_voxels), but each sample has an RGB color corresponding to its XYZ position in the coordinates of the proposed [3DTILES_content_voxels extension](https://github.com/CesiumGS/3d-tiles/tree/cylinder-changes/extensions/3DTILES_content_voxels). This should help verify that we are handling the different coordinate systems correctly.

image

ggetz (approved):
Looks great @jjhembd! Love that the sandcastles are much simpler easier to understand now.

Just as an FYI— I pushed up a commit with some minor doc fixes.

Issue #11960, opened 2024-04-26
External tilesets are disallowed inside multiple contents


javagl:
The title basically says it all: When using multiple contents that refer to external tilesets, then a message
> Error: External tilesets are disallowed inside multiple contents

is printed and the external tilesets are not rendered.

This _might_ be a leftover from the `3DDILES_multiple_contents` extension: This explicitly disallowed external tilesets in the contents (see https://github.com/CesiumGS/3d-tiles/pull/583/files ). But in 3D Tiles 1.1, there is no such restriction.

Attached a _really_ minimal test case of one tileset that only has two contents, which in turn are (external) tilesets, to reproduce the issue:

[external tilesets in multiple contents.zip](https://github.com/CesiumGS/cesium/files/15134199/external.tilesets.in.multiple.contents.zip)

javagl:
I did have a short look at this. And I pragmatically created a test case:

- a tileset with a root that as 4 `contents`
- the _first_ content is a *red* unit cube GLB
- the other contents are three external tilesets (that refer to the green, blue, and white content, respectively)

[multipleContentsExternal 2024-12-04b.zip](https://github.com/user-attachments/files/18014662/multipleContentsExternal.2024-12-04b.zip)

Then, I simply commented out the [check for external tilesets in multiple contents](https://github.com/CesiumGS/cesium/blob/382c517418bb78be96d5a1988085bf3b9bf1c467/packages/engine/Source/Scene/Multiple3DTileContent.js#L521), and it works:

![Cesium Multiple External 0001](https://github.com/user-attachments/assets/bcfd0f29-304f-4e55-b2a2-5c67e713ff20)

Yay 🎉 

But ... there is a flag `tile.hasTilesetContent`. This is a `boolean` that is set for tiles that have external tilesets as content. Now ... one could be tempted to think that the flag still can to `true` here, because after all, this tile **does** have tileset content. But it also has non-tileset content 🤔 

Searching for this flag reveals several places where this is checked. Among them are things like this...
```javascript
/**
 * @private
 * @param {Cesium3DTileset} tileset
 * @param {Cesium3DTile} tile
 * @returns {number}
 */
function getPriorityReverseScreenSpaceError(tileset, tile) {
  const parent = tile.parent;
  const useParentScreenSpaceError =
    defined(parent) &&
    (!tileset.isSkippingLevelOfDetail ||
      tile._screenSpaceError === 0.0 ||
      parent.hasTilesetContent ||
[...]

PR #12440, opened 2025-01-17
Allow external tilesets in multiple contents

javagl:
Addresses https://github.com/CesiumGS/cesium/issues/11960 

# Description

It removes the check that previously did throw a `RuntimeError` with the message

> "External tilesets are disallowed inside multiple contents"

when trying to load a tileset where an external tileset was found in the (multiple) `contents` array.

## Detailed description

The relevant part of the "trace" that leads to this point is drafted here:

- It starts at [`Cesium3DTileset.requestContent`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Cesium3DTileset.js#L2494) for a certain tile
- This calls [`Cesium3DTile.requestContent`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Cesium3DTile.js#L1130). When the tile contained the `contents` array, this delegates to [`Cesium3DTile.requestMultipleContents`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Cesium3DTile.js#L1157)
- This creates the `Multiple3DTileContent` on which it calls [`requestInnerContents`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Multiple3DTileContent.js#L364). This calls `requestInnerContent` for each content, then calls `createInnerContents` which waits for the requests, and then calls `createInnerContent` for each response.
- Finally, [`Multiple3DTileContent.createInnerContent`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Multiple3DTileContent.js#L511) did throw the error when it found that the content type was `EXTERNAL_TILESET`.

The place that originally threw up _can_ in fact handle the creation of external tileset content. By just commenting out the check, the issue was fixed, and it "worked": It just used the `Cesium3DTileContentFactory` to create the content., which then was just a `Tileset3DTileContent`.

#### Caveats....

For single-content external tilesets, an external tileset caused the `tile.hasTilesetContent = true;` to be set. And this `hasTilesetContent` is checked in various places. 

In most places, it seems to be used in the meaning that there are some children to traverse. For example, when `contentExpired` turns `true`, it causes `destroySubtree` to be called. This should be fine.

In [`getPriorityReverseScreenSpaceError`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Cesium3DTile.js#L1011), it is taken into account in a non-obvious way. I don't see a reasonable way to understand the implications of this flag at this point, but hope that having it being `true` there is fine as well.

The `Cesium3DTile#hasRenderableContent` may be the most critical one. It was implemented as
```javascript
  /**
   * Determines if the tile's content is renderable. <code>false</code> if the
   * tile has empty content or if it points to an external tileset or implicit content
   *
   * @memberof Cesium3DTile.prototype
   *
   * @type {boolean}
   * @readonly
   *
   * @private
   */
  hasRenderableContent: {
    get: function () {
      return (
        !this.hasEmptyContent &&
        !this.hasTilesetContent &&
        !this.hasImplicitContent
      );
    },
  },
```

A somewhat shallow way of putting it is: With multiple contents, the tile _may_ have renderable content _even though_ it has a tileset content!

So I tried to examine that more thoroughly. And the `hasRenderableContent` is checked in various places:
- It must be `true` for `tile.contentAvailable` to ever be `true`
- It must be `true` for `tile.hasUnloadedRenderableContent` to be `true`. The `hasUnloadedRenderableContent` is checked in various places. It has effects on loading, often in the context of expiration.
- It must be `true` for `tile.unloadContent` to do anything(!)
- It affects colors in `applyDebugSettings`
- It affects the traversal, for example, for [the refinement](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L108), but also at many other places. 

The effect of `hasRenderableContent` (and its dependent, `hasUnloadedRenderableContent`) are scattered and not obvious. They usually appear in private (aka undocumented) functions. I had to stop "zooming" into this at some point.

#### Trying to deal with the caveats...

The change here revolves around two assumptions:

- I **assume** that `hasTilesetContent` also has to be set to `true` when the external tileset is one of _multiple_ contents.
- I **assume** that `hasRenderableContent` must usually be `true`, except for the case of 1. empty content, 2. single external, 3. single implicit

So made `hasRenderableContent` an explicit variable (not just a getter that returns the `&&`-result of three others). This has to be set to `true` or `false`, consistent with its previous behavior and the assumptions. 

Previously, `hasRenderableContent` depended on `hasEmptyContent`, `hasTilesetContent`, and `hasImplicitContent`. What makes this more difficult are the comments in `hasTilesetContent` and `hasImplicitContent` that say
`This is <code>false</code> until the tile's (implicit) content is loaded.`
meaning that the return value of `hasRenderableContent` would _change_ through the loading process. What's even more quirky: When _nothing is loaded yet_, then `hasRenderableContent` will return `true` by default. Ouch.

So what is done here:

- `hasRenderableContent` is a variable that starts as `true` except for empty content
- It is only set to `false` when encountering a _single_ content that is an external tileset or implicit content

## Issue number and link

https://github.com/CesiumGS/cesium/issues/11960

## Testing plan

I tried out the test case that was [posted in the issue](https://github.com/CesiumGS/cesium/issues/11960#issuecomment-2518686177), and other tilesets that involve external tilesets. For (somewhat helpless) testing, I inserted a brutal `this._cache.trim();` in `Cesium3DTileset.postPassesUpdate`, to basically "force" the unloading/reloading in each frame, and then played with moving the tileset out of the viewport and back in again, to ensure that it is properly reloaded each time.

Beyond that, the difficulty of _really_ testing this is why I added so many "Details" above. 

In terms of "specs": There are currently two distinct specs:

- "multiple contents": "raises tileFailed for external tileset inside multiple contents"
- "3DTILES_multiple_contents: "raises tileFailed for external tileset inside multiple contents (legacy)"

Theoretically, only the second one should keep throwing. 

But in practice, nobody is ever going to use `3DTILES_multiple_contents`, any more, and if someone uses it for external tilesets (i.e. not compliant to the spec), and this happens to work in CesiumJS, that could be OK.

(Otherwise, we'd have to carry along the information of whether something was created from `3DTILES_multiple_contents` or 3D Tiles 1.1, just to throw an "unnecessary" error at the right place ...)

The specs are not updated yet. I have no idea why the tests are currently passing. 
<strike>I'll have to investigate that. Hey. Passing tests being wrong. That's new 🙂 </strike>
Update: It looks like they are failing now. That's good.

# Author checklist


ggetz:
@javagl Let me confirm my understanding about the state of this PR: This is ready for review on the approach, then you'll proceed with updating the specs accordingly. Correct?

@lilleyse Would you be able to take a pass on the approach here?

javagl:
> This is ready for review on the approach, then you'll proceed with updating the specs accordingly. 

Yes, given the depth of possible implications, I think that it could be worthwhile to review the approach itself, with scrutiny, by someone who might come up with ideas of what _might_ go wrong in areas where I couln't zoom to the "bottom".

Updating the specs _chould_ then be simple: There are four failing specs right now. The ones that expected the error to be thrown could be changed to just _not_ expect the error to be thrown. One talks about color - that could be a fluke. The one at https://github.com/CesiumGS/cesium/actions/runs/12836753145/job/35799051528?pr=12440#step:6:43 warrants another look. It is about refinement, which _might_ be a side-effect of the changes here (although not immediately obvious).

javagl:
The failing test was [setting `root.hasEmptyContent`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Specs/Scene/Cesium3DTilesetSpec.js#L1410) to `false`, to `// mock content`, as it says.

Specifically, this had the effect that [this call to `hasRenderableContent`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/packages/engine/Source/Scene/Cesium3DTilesetBaseTraversal.js#L214) returned `true`.

So this "mocking" relied on the very intricate implementation of `hasRenderableContent`. I fixed this in https://github.com/CesiumGS/cesium/pull/12440/commits/cefb360dfaa4e0b32bec5c95baeeb04a128fd068 to set the `hasRenderableContent` (which now is an explicit boolean property) to `true`.

javagl:
The spec that just said `expected color` also did set `hasEmptyContent=false`, and I fixed it the same way.

Now, the remaining failing ones are 🤞 only the ones where the expectation has to be changed from 'expecting an error' to 'expecting no error'. 

Futher specs should be considered - like one that checks that this external tileset actually _is_ traversed. But I'd address that when the general approach is confirmed.

ggetz:
Bump @lilleyse, would you have time to take a pass on this?

lilleyse:
@ggetz yes, should have a chance this week.

javagl:
The specs update here had still been pending. Although it might not make its way into the next release, I wanted to wrap this up, in doubt to that it may be merged shortly _after_ the release and undergo some sort of additional "implicit testing" before the next one.

The specs update consisted of these two steps:

The spec 
`"raises tileFailed for external tileset inside multiple contents (legacy)"`
has been removed. While this extension did disallow external tilesets in multiple contents, it's probably not worth keeping this spec: It would be necessary to keep track of _where the tileset came from_, just to throw an (now unnecessary) error...

The spec
`"raises tileFailed for external tileset inside multiple contents"`
has been replaced with
`"renders external tilesets in multiple contents"`

This spec uses the tileset that was [added in the issue](https://github.com/CesiumGS/cesium/issues/11960#issuecomment-2518686177). It renders the following (debug canvas): 

Cesium Multiple External Spec

and checks that 4 tiles are selected and visited (the root of the main tileset, and the three roots of the external tilesets).

lilleyse:
Should `hasRenderableContent` be false if all contents are external tilesets?

Otherwise, the approach makes sense. I'm less worried about the case where there's a mix of renderable content and external tilesets. It's harder to reason about but also less likely to occur in practice (I think).

javagl:
I think that this _should_ make sense, and I considered it at one point. The reasons why I did not implement this initially:

- I naively started with setting `hasRenderableContent` to _false_ at the beginning, because ... initially, the tile certainly does not have renderable content. But then, no traversal took place at all...
- The breadth of possible implication of doing this - for example, in [`unloadContent`](https://github.com/CesiumGS/cesium/blob/92113d8191cb3493f5e379e912df596b8dda9f4f/packages/engine/Source/Scene/Cesium3DTile.js#L1422) (but also all the other places mentioned above)
- The way how the "inner contents" of a `Multiple3DTileContent` are created

The latter refers to the fact that it generates some promises, waits for them, and eventually has an array of `content` objects without _explicit_ information about their _type_. This means that there are two options for implementing this:

- Add some flag like `isTilesetContent` in _the content itself_ (and afterwards, count how many of the inner contents have this flag set to `true`)
- Keep track of the number of tileset contents _in_ the `Multiple3DTileContent` itself

For now, I took the second approach: There now is a `_externalTilesetCount` that counts how many of the contents have been external tilesets, and sets `tile.hasRenderableContent` to `false` if it was all of them. 

One caveat: There might be cases where an inner content is cancelled, failed, or its loading is interrupted, and depending on where exactly that _can_ happen, it's hard to say whether this counter reflects the truth. If you think that the (more invasive, wide-ranging), but somehow "safer" approach of keeping track of an `isTilesetContent` in _each_ content would be better, I'll change it accordingly.

(I also added a spec for this case, based on the existing one, by just wrapping the red cube into another tileset)

lilleyse:
I prefer the second approach (the one you took). Just update `CHANGES.md` and I'll merge.

lilleyse (approved):

javagl:
Done. 
(I wonder whether the links in the CHANGES.md should be to the `issue` or to the `pull` - but they probably are connected enough so that it does not really matter...)

PR #12458, opened 2025-02-03
Fix availability check in VoxelTraversal for procedural tilesets

jjhembd:
# Description

This PR fixes a small bug introduced in https://github.com/CesiumGS/cesium/pull/12430 affecting procedural tilesets.

The `requestData` helper function in `VoxelTraversal` included a check of `provider._implicitTileset.availableLevels` to avoid generating spurious `tileFailed` events. However, custom procedural providers (as used in some of our sandcastle examples) do not have an `._implicitTileset`.

This PR adds a public `availableLevels` property to `VoxelProvider`, which is allowed to be undefined. `VoxelTraversal` now verifies the property is defined before checking its value.

## Testing plan

Load the "Voxels" Sandcastle and verify that all datasets render without errors.

# Author checklist

- ~[ ] I have updated `CHANGES.md` with a short summary of my change~ Fixes a bug not present in the previous release

### DIFF
--- a/packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
+++ b/packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
@@ -98,6 +98,7 @@ Object.defineProperties(Cesium3DTilesVoxelProvider.prototype, {
   /**
    * The number of levels of detail containing available tiles in the tileset.
    *
+   * @memberof VoxelPrimitive.prototype
    * @type {number|undefined}
    * @readonly
    */
--- a/packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
+++ b/packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
@@ -89,12 +89,24 @@ function Cesium3DTilesVoxelProvider(options) {
   /** @inheritdoc */
   this.maximumTileCount = undefined;
 
-  /** @inheritdoc */
-  this.availableLevels = undefined;
-
+  this._availableLevels = undefined;
   this._implicitTileset = undefined;
   this._subtreeCache = new ImplicitSubtreeCache();
 }
+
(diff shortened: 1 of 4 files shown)


packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
```diff
@@ -89,6 +89,9 @@ function Cesium3DTilesVoxelProvider(options) {
   /** @inheritdoc */
   this.maximumTileCount = undefined;
 
+  /** @inheritdoc */
+  this.availableLevels = undefined;
```
ggetz, line 93: To match the VoxelProvider interface his needs to be defined as a getter, right?

ggetz (changes requested):
Thanks for the fix @jjhembd! 

Would it be possible to add a unit test for this case to prevent a similar regression in the future? I'm surprised we don't have one that failed already!

packages/engine/Source/Scene/Cesium3DTilesVoxelProvider.js
jjhembd: Yes, same for all 14 of these properties. I was trying to do a minimal fix and leave the refactor for https://github.com/CesiumGS/cesium/pull/12432

ggetz: Fair, but let's just do this one correctly to minimize conflicts.

jjhembd:
> Would it be possible to add a unit test for this case to prevent a similar regression in the future? I'm surprised we don't have one that failed already!

@ggetz The specs that actually loaded tiles were all rendering one of the test 3D Tiles voxel datasets. I added a spec to load a minimal procedural provider. This spec fails in main, and succeeds in this branch.

lukemckinstry:
@jjhembd this fix looks good to me

ggetz (approved):

Issue #7864, opened 2019-05-22
GeoJSON crashes in Cesium


OmarShehata:
This is a regression where it used to work in 1.53. In [this Sandcastle](https://cesiumjs.org/Cesium/Build/Apps/Sandcastle/#c=bZJRT+MwDMe/itWXtWLnbEK8bGUCDcTBndhJQ8cDu4fQpGtuWVIlaasx8d0vXbqxwfXFsfP/2bXjmhqoBW+4gUtQvIEpt6Ja4+9dLO5lO3eqlaNCcdNLxgu1ULWnllw/zGePHuuQOx+wWt1QR+e6MhlHqSmLe4VzpR0RUqlytcRMr0mjjWTfqJPUXg2DR4bDwRr/er7Xh21bA7rPOqNXfLSvMtVSG/w+e/p1//ijfyzMhZSfZK0GG+GKa1kWNB7gRdL/mvtZMFeM4Ly9eQ8Nds2hK7iK80plTmgVM99asg04ITAtqBeC1wA12dOm9GcNd7ezm9v5/bQPTSGyAoT1B+pAOGioBaG6X3yYwxAvzjGky7WBuB2r8AMdjL1JoS2HXDnhBLdYU1l5I7lausILzs4S2H600rI77cYn+B/5Iv6MP+Qih7gbFeO5f1oWBxpLLTdLrZKT7F3HpxLcd31YgesQwP0Mjgq+h2NnwsohO+yKRcpYGPD48ApRP0qt20g+CdSVWJfaOKiMjBGJ4+tSUsctea2yFXeYWduCKdlDKRM1CHa5iD7t8SKCzG+f9Td5JeVcvPFFNEmJ159g7QYLtZzV3Ei6aSXFcPIzBBExJd79Sjmt5Ss1Rxn/AQ), a GeoJSON is loaded and crashes with:

```
DeveloperError: EllipsoidRhumbLine must have distinct start and end set.
```

I'm not sure if it's related to https://github.com/AnalyticalGraphicsInc/cesium/issues/7550. But you can fix this by forcing the arcType to be `Cesium.ArcType.GEODESIC` (by uncommenting line 18 in the Sandcastle).

Should this be a case where we handle this error and make it a warning, and attempt to draw the GeoJSON anyway? Or was this never a valid GeoJSON to begin with?

OmarShehata:
@shehzan10 do you have any insight here? I believe it broke with the rhumb line PR we added.

[...]

Issue #12456, opened 2025-01-29
Support OGC API - Features


jjspace:
## The Spec

This issue is to track adding full support for the [OGC API - Features](https://ogcapi.ogc.org/features/) spec as well as any discussion around the process.

The spec currently includes 5 parts with parts 4 and 5 currently under draft. The initial plan will be to support Part 1: Core with others as follow up tasks. I currently do not expect us to support Part 2 which adds CRSs which Cesium already does not do much work with.

> - [OGC API - Features - Part 1: Core corrigendum, version 1.0.1](https://docs.opengeospatial.org/is/17-069r4/17-069r4.html) standard
>     - [Latest Editor's Draft of next version of OGC API - Features - Part 1: Core](http://docs.opengeospatial.org/DRAFTS/17-069r5.html)
>     - [ISO 19168-1:2020 Geographic information - Geospatial API for features - Part 1: Core](https://www.iso.org/standard/32586.html) standard
> - [OGC API - Features - Part 2: Coordinate Reference Systems by Reference corrigendum, version 1.0.1](https://docs.opengeospatial.org/is/18-058r1/18-058r1.html) standard
>     - [Latest Editor's Draft of next version of OGC API - Features - Part 2: CRS by Reference](https://docs.opengeospatial.org/DRAFTS/18-058r2.html)
> - [OGC API - Features - Part 3: Filtering, version 1.0.0](https://docs.ogc.org/is/19-079r2/19-079r2.html)
> - [OGC API - Features - Part 4: Create, Replace, Update and Delete](https://docs.ogc.org/DRAFTS/20-002r1.html) **draft**
> - [OGC API - Features - Part 5: Schemas](https://docs.ogc.org/DRAFTS/23-058r1.html) **draft**

The spec does not mandate any specific [encoding](https://docs.ogc.org/is/17-069r4/17-069r4.html#_encodings) or format but strongly favors HTML and GeoJSON with additional suggestions for GML. We will focus the initial implementation only around GeoJSON (especially since we already support it) with other formats being optional secondary efforts in the future as requested.

**Side note:** the API does not dictate a specific way to designate what format to serve. I believe most servers will implement this with the content type in headers but another _suggestion_ [from the Spec](https://docs.ogc.org/is/17-069r4/17-069r4.html#encodings) is to use a `f=[format]` query parameter which I've observed utilized in other libraries and servers so my current assumption is this is one of the more generally accepted methods.

## Example data

[...]

PR #12460, opened 2025-02-03
Fix geojson `EllipsoidRhumbLine` error for duplicated points

jjspace:
# Description

Sometimes polygons have duplicated points in their definition. This causes the code we have to subdivide the lines for the `EllipsoidRhumbLine` arc type to fail. Even if it's not ideal the duplicated points shouldn't cause our render to crash, especially when other applications like `geojson.io` render correctly and validators show nothing wrong with the GeoJSON.

Note: I'm not 100% confident this fix is done at the right "level" of the code. I was also able to get the failing examples working by removing the error call itself. That said I think it's a valid error if you're using that class directly but in this instance for polygons we simply don't care and can short circuit the function before we even try the code that "fails".

For context I encountered this during work on #12456 with one of the [example sources](https://demo.ldproxy.net/vineyards/collections/vineyards/items/810503)

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/7864

## Testing plan

- Test these local sandcastles in `main` and this branch and check that this branch does not crash out
    - [Country outlines](http://localhost:8080/Apps/Sandcastle/index.html#c=bZJRT+MwDMe/itWXtWLnbEK8bGUCDcTBndhJQ8cDu4fQpGtuWVIlaasx8d0vXbqxwfXFsfP/2bXjmhqoBW+4gUtQvIEpt6Ja4+9dLO5lO3eqlaNCcdNLxgu1ULWnllw/zGePHuuQOx+wWt1QR+e6MhlHqSmLe4VzpR0RUqlytcRMr0mjjWTfqJPUXg2DR4bDwRr/er7Xh21bA7rPOqNXfLSvMtVSG/w+e/p1//ijfyzMhZSfZK0GG+GKa1kWNB7gRdL/mvtZMFeM4Ly9eQ8Nds2hK7iK80plTmgVM99asg04ITAtqBeC1wA12dOm9GcNd7ezm9v5/bQPTSGyAoT1B+pAOGioBaG6X3yYwxAvzjGky7WBuB2r8AMdjL1JoS2HXDnhBLdYU1l5I7lausILzs4S2H600rI77cYn+B/5Iv6MP+Qih7gbFeO5f1oWBxpLLTdLrZKT7F3HpxLcd31YgesQwP0Mjgq+h2NnwsohO+yKRcpYGPD48ApRP0qt20g+CdSVWJfaOKiMjBGJ4+tSUsctea2yFXeYWduCKdlDKRM1CHa5iD7t8SKCzG+f9Td5JeVcvPFFNEmJ159g7QYLtZzV3Ei6aSXFcPIzBBExJd79Sjmt5Ss1Rxn/AQ) from #7864 
    - [Vineyard](http://localhost:8080/Apps/Sandcastle/index.html#c=bZDRS8MwEMb/ldCnDkayIYK4tgoTBBF8mPjUlyy5uWiaK8m1s/vrTVqVMZc83d3v++7jFLpArDdwAM9K5uDA1hBM1/C3sZfXmRrrNTqSxoGvs9mqdumrUfsR0EXlZMG1JLnBzkcRl1rntWO/ho+AT5F9+CO4RRmJbE/UhlshNDTIrW49fg3cAYk+7huk10EotBYUmbjxpGsImiBulovrxdXdrkxJYrp57aaEP5GOiM0r5mk6W2XzrAg0WKhSsvTuTdOiJ9Z5m3MuomVrJUEQ2059AnEVQrJLaCFOpYU2PTO6vHAhpqwMIU52nbUbc4Q6qwoR+X/SdALj3l968FYOCdsvq+epyTkvRCwvKwnRbqU/c/4G) from sample data from #12456 (if it 404s just try reloading)

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/PolygonGeometryLibrary.js
+++ b/packages/engine/Source/Core/PolygonGeometryLibrary.js
@@ -323,19 +323,24 @@ PolygonGeometryLibrary.subdivideRhumbLine = function (
   const c1 = ellipsoid.cartesianToCartographic(p1, scratchCartographic1);
   const rhumb = new EllipsoidRhumbLine(c0, c1, ellipsoid);
 
-  if (rhumb.surfaceDistance === 0) {
-    // no need to try and subdivide a line with the same start and end
-    return [p0.x, p0.y, p0.z];
+  if (!defined(result)) {
+    result = [];
+  }
+
+  if (rhumb.surfaceDistance <= minDistance) {
+    // no need to try and subdivide a line that's already shorter than the min distance
+    // this also inherently handles duplicated points which would have 0 distance
+    result.length = 3;
+    result[0] = p0.x;
+    result[1] = p0.y;
+    result[2] = p0.z;
+    return result;
   }
 
   const n = rhumb.surfaceDistance / minDistance;
   const countDivide = Math.max(0, Math.ceil(CesiumMath.log2(n)));
   const numVertices = Math.pow(2, countDivide);
   const distanceBetweenVertices = rhumb.surfaceDistance / numVertices;
-
-  if (!defined(result)) {
-    result = [];
-  }
 
   const positions = result;
   positions.length = numVertices * 3;
--- a/packages/engine/Source/Core/PolygonGeometryLibrary.js
+++ b/packages/engine/Source/Core/PolygonGeometryLibrary.js
@@ -323,6 +323,11 @@ PolygonGeometryLibrary.subdivideRhumbLine = function (
   const c1 = ellipsoid.cartesianToCartographic(p1, scratchCartographic1);
   const rhumb = new EllipsoidRhumbLine(c0, c1, ellipsoid);
 
+  if (rhumb.surfaceDistance === 0) {
+    // no need to try and subdivide a line with the same start and end
+    return [p0.x, p0.y, p0.z];
+  }
+
   const n = rhumb.surfaceDistance / minDistance;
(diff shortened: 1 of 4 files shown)


packages/engine/Source/Core/PolygonGeometryLibrary.js
```diff
@@ -323,6 +323,11 @@ PolygonGeometryLibrary.subdivideRhumbLine = function (
   const c1 = ellipsoid.cartesianToCartographic(p1, scratchCartographic1);
   const rhumb = new EllipsoidRhumbLine(c0, c1, ellipsoid);
 
+  if (rhumb.surfaceDistance === 0) {
```
ggetz, line 326: Can we use an [`equalsEpsilon`](https://cesium.com/learn/ion-sdk/ref-doc/Math.html#.equalsEpsilon) check?

Or maybe better yet, we can return earlier as long as the surface distance is less then or equal to `minDistance`, right?

ggetz:
Nice, this is a long running bug that I'm sure many will love to see fixed. Thanks @jjspace!

I just have one comment about how we can make this fix a bit more general-purpose. Please update `CHANGES.md` and look into adding a unit test at least for this case. 

I don't think we have any tests for `subdivideRhumbLine` specifically, so some additional coverage would also be welcome. 😄

ggetz:
All looks good to me! I can confirm several of the example datasets from the original issue are now rendering without error 🎉

ggetz (approved):

ggetz:
Thanks @jjspace!

PR #12463, opened 2025-02-05
Add statistics to VoxelPrimitive

lukemckinstry:
For performance & profiling purposes, add additional statistics from Cesium3DTilesetStatistics to be tracked in VoxelPrimitive as they are in Cesium3DTileset.

 - textureByteLength
 - visited
 - numberOfTilesWithContentReady

Based on my understanding of voxels it seems that selected and numberOfTilesWithContentReady mean the same thing in the voxel context, so I just implemented numberOfTilesWithContentReady.

## Issue number and link

#12297
Follow up to PR #12430

## Testing plan
 - The `VoxelPrimitive.statistics` are not exposed in the public api. A good way to test is to set debug statements and look at a sandcastle with voxels (either generated or loaded).
 - either a) set calculateStatistics to true in the VoxelPrimitive constructor or b) set VoxelTraversal._calculateStatistics to true with a comment in the code.
 - Then in `VoxelTraversal.postPassesUpdate`, console.log statistics.visited, statistics.numberOfTilesWithContentReady, statistics.textureByteLength after line 800. The first 2 stats are identical to the existing values for `LOADED` and `TOTAL` in the debug message, which can be turned on by setting `VoxelTraversal._debugPrint = true`. texturesByteLength should be set when the Megatextures are created and should not change.

# Author checklist


lukemckinstry:
@jjhembd @ggetz you can hold off on a full review because I need to take another pass to add unit tests

lukemckinstry:
ready now for review @ggetz @jjhembd

packages/engine/Source/Scene/Cesium3DTilesetStatistics.js
```diff
@@ -170,7 +170,6 @@ Cesium3DTilesetStatistics.clone = function (statistics, result) {
   result.selected = statistics.selected;
   result.visited = statistics.visited;
   result.numberOfCommands = statistics.numberOfCommands;
-  result.selected = statistics.selected;
```
lukemckinstry, line 173: removing a duplicate line, not directly related to this PR

ggetz:
Thanks @lukemckinstry, I will take a pass soon! Could you please update the testing plan and `CHANGES.md` to help with review?

lukemckinstry:
@ggetz updated change log and test instructions, ready for review

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -1059,6 +1059,15 @@ Object.defineProperties(VoxelPrimitive.prototype, {
       return this._customShaderCompilationEvent;
     },
   },
+
+  /**
+   * @private
```
ggetz, line 1064: I know it's this way in `Cesium3DTileset`, but since it is helpful for internal developers to have type information and documentation too, lets try to document private properties with a similar rigor as public ones. 

Along the lines of:
```suggestion
   *  Loading and rendering information for requested content
   * @type {Cesium3DTilesetStatistics}
   * @readonly
   * @private
```

ggetz:
I had one comment on the inline documentation, but this is looking good to me! Over to @jjhembd for review and merge.

packages/engine/Source/Scene/VoxelTraversal.js
```diff
@@ -784,16 +795,28 @@ function postPassesUpdate(
...
+  primitive.statistics.numberOfTilesWithContentReady =
+    loadStateByCount[KeyframeNode.LoadState.LOADED];
+  primitive.statistics.visited = nodeCountTotal;
+
+  const checkEventListeners =
+    primitive.loadProgress.numberOfListeners > 0 ||
+    primitive.allTilesLoaded.numberOfListeners > 0 ||
+    primitive.initialTilesLoaded.numberOfListeners > 0;
+  if (!that._debugPrint && !checkEventListeners) {
+    return;
```
jjhembd, line 807: If there are no event listeners, can we exit before calling `traverseRecursive`? recursive traversal is not cheap.

packages/engine/Source/Scene/VoxelTraversal.js
```diff
@@ -685,6 +683,18 @@ function loadAndUnload(that, frameState) {
       keyframeNodesInMegatexture[addNodeIndex] = highPriorityKeyframeNode;
     }
   }
+
+  const dimensions = Cartesian3.clone(
+    primitive._provider.dimensions,
+    scratchDimensions,
+  );
+  primitive.statistics.textureByteLength =
+    VoxelTraversal.getApproximateTextureMemoryByteLength(
```
jjhembd, line 692: The texture memory size is set once in the `Megatexture` constructor. We might want to expose the actual size (something like `(textureDimension * channelCount * componentByteLength) ** 2`) in that class, rather than re-computing the "approximate" size.

Once the Texture is constructed, its size doesn't change. The `occupiedCount` number indicates how much of the texture is filled with useful data. But even if we haven't loaded anything yet, the texture has already been allocated and is using its full amount of memory. 

I think the `textureByteLength` is mainly used to get an idea of how much memory has been allocated.

We also need to keep track of _how many_ megatextures have been allocated. This is based on how many metadata fields the voxel dataset contains. See the `new Megatexture()` calls in the `VoxelTraversal` constructor.

packages/engine/Specs/Scene/VoxelPrimitiveSpec.js
```diff
@@ -83,6 +83,9 @@ describe(
       });
       expect(spyUpdate1.calls.count()).toEqual(1);
       expect(spyUpdate2.calls.count()).toEqual(1);
+      expect(primitive.statistics.numberOfTilesWithContentReady).toEqual(1);
```
jjhembd, line 86: Can we update the name/description of this spec to indicate that we are checking more than just "initial tiles loaded and all tiles loaded events" ?

jjhembd (changes requested):
Thanks @lukemckinstry! This is looking good overall. I have a couple minor comments, and also some input on how `textureByteLength` is computed.

packages/engine/Source/Scene/VoxelTraversal.js
lukemckinstry: The way I have it set up now, `statistics.numberOfTilesWithContentReady` and `statistics.visited` rely on the stats gathered in the traversal. Perhaps these stats could be gathered another way during the loadAndUnload function?

lukemckinstry: And unlike events where we can check is event listeners habe subscribed we do not really know if anybody is seeking these stats, unless perhaps we had a flag to turn off/on collection of this data.

lukemckinstry:
Added to commits to address comments from @jjhembd 
 https://github.com/CesiumGS/cesium/pull/12463/commits/254ae740cd4494a4091896c3d08365422e024db4 Makes texturesByteLength use the total allocated memory from the Megatextures(). It is set once when the Megatexture is created and not changed.
https://github.com/CesiumGS/cesium/pull/12463/commits/67b97e94581e24bda8d025b4fc998dbacd6fc145 adds a boolean field `calculateStatistics` to the VoxelPrimitive constructor. This controls whether to do the `postPassesUpdate` in VoxelTraversal which includes a costly extra traversal. 
 - The drawback of this approach is that the library will appear slower when measuring performance. The alternative I tried was implementing the statistics updates in the loadAndUnload function. This seemed less consistent, but I could investigate further if we want to avoid this drawback.

ggetz:
@lukemckinstry It looks like CI is failing on the lint step. Could you please take a look?

jjhembd (approved):
Thanks @lukemckinstry! I merged main and took care of the conflicts introduced by my PR https://github.com/CesiumGS/cesium/pull/12432.
I also pushed a small commit to use the exact value for `Megatexture.textureMemoryByteLength`, based on the rounding of the texture to power-of-two size.
I think this is good to go!

PR #12429, opened 2025-01-10
Change `DataSourceDisplay` ready state behavior for bounding sphere calculations

jjspace:
# Description

Slightly reverting the change in https://github.com/CesiumGS/cesium/pull/12230 and replacing PR https://github.com/CesiumGS/cesium/pull/12322 in favor of changing the `ready` value to fix both https://github.com/CesiumGS/cesium/issues/4647 and the newer issue https://github.com/CesiumGS/cesium/issues/12303.

The core of this change is to make the `DataSourceDisplay._ready` boolean to remain `true` once it's set to `true` even if the internal datasources become "un-ready" due to transformations or loading.

## Issue number and link

Alternative fix for https://github.com/CesiumGS/cesium/issues/4647
Fixes https://github.com/CesiumGS/cesium/issues/12303

## Testing plan

- Test [this sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVHda8IwEP9Xjj5VcKnDSZ2rMub2JuxB2FNA0uTswtKkJKnihv/70sbCPnwIubvc/T4u3Gjn4SDxiBaWoPEIa3SyrclbX0tpwvt8bbRnUqOlyeiBaqr3reZeGg1MiCepVGmYFekIvqgG2DItOHNeIRHIFbOY/myLCAC8Z0ftpT8F9iiD9LlER8JI2sMBNMbJjm0xyFsz60PE9JTsramfsbKILr3JZ2R2n+f5GO4mZDKdz6ejcYQoB/YFXEABZM0qXABNCMn62GURf7cxldmZA1rFTqTRFU0uMOf+PncWALpzUe1QIfcoXgY30VZoO3dmf28pVJNxUjh/UrgaxDzKujHWQ2tVGvR4rBvFgsmsbPkHesKdi6wARfZztBDyAFIsr/wVhOU7F172rVJb+Yk0WRVZ6P83qgwTUlev0XLX9n672sQiIaTIQnp90hujSmb/IH8D) which demonstrates the error when setting the `viewer.selectedEntity`
[...]

PR #3954, opened 2016-05-24
Model entity on terrain

tfili:
@mramato Can you take a look at the `EntityView` stuff?

PR #12467, opened 2025-02-07
Fixes issue #12465 (Entity Tracking broken)

### LINKED ISSUE #12465: Entity Tracking sandcastle broken in 1.126
### What happened?

The Entity Tracking sandcastle appears to be broken in 1.126. Worked in 1.125.

https://sandcastle.cesium.com/?src=Entity%20tracking.html

It doesn't appear to track either the satellite or drone in any reference frame.

If I comment out the following lines:

    satellite.viewFrom = new Cesium.Cartesian3(-300, 20, 100);
    drone.viewFrom = new Cesium.Cartesian3(-50, 0, 5);

It starts tracking.

### Reproduction steps

1. View https://sandcastle.cesium.com/?src=Entity%20tracking.html
2. Click drone
3. Camera doesn't track drone
...

### Sandcastle example

https://sandcastle.cesium.com/?src=Entity%20tracking.html

### Environment

Browser: Edge / Firefox / Chrome
CesiumJS Version: 1.126
Operating System: Windows 11

jfayot:
# Description

This is an attempt to fix broken entity tracking when viewFrom is set:
Restored `boundingSphereScratch` instead of `entityView.boundingSphere` inside `CesiumWidget._onTick`.

## Issue number and link

Fixes #12465 

## Testing plan

- Updated EntityViewSpec
- Tested [Entity tracking sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Entity%20tracking.html)
- Tested [Clamp Entities to Ground](http://localhost:8080/Apps/Sandcastle/index.html?src=Clamp%20Entities%20to%20Ground.html)
- Tested [this sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=bVHda8IwEP9Xjj5VcKnDSZ2rMub2JuxB2FNA0uTswtKkJKnihv/70sbCPnwIubvc/T4u3Gjn4SDxiBaWoPEIa3SyrclbX0tpwvt8bbRnUqOlyeiBaqr3reZeGg1MiCepVGmYFekIvqgG2DItOHNeIRHIFbOY/myLCAC8Z0ftpT8F9iiD9LlER8JI2sMBNMbJjm0xyFsz60PE9JTsramfsbKILr3JZ2R2n+f5GO4mZDKdz6ejcYQoB/YFXEABZM0qXABNCMn62GURf7cxldmZA1rFTqTRFU0uMOf+PncWALpzUe1QIfcoXgY30VZoO3dmf28pVJNxUjh/UrgaxDzKujHWQ2tVGvR4rBvFgsmsbPkHesKdi6wARfZztBDyAFIsr/wVhOU7F172rVJb+Yk0WRVZ6P83qgwTUlev0XLX9n672sQiIaTIQnp90hujSmb/IH8D) which demonstrates the error when setting the viewer.selectedEntity
- Tested [this sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=pVcLT+s2FP4rVqeJdKRpuWXTXSlo47HpSuiCVLqrjaLJJG5rzbEj22nprvjvO7YTxykZMK1CNDn2Of7Od15uKrjSaEPJlkh0ijjZoguiaJknv1lZtOil9v1CcI0pJ3LR658s+IIPh+iSLEGC9JqgQrAdg5cDBY+Kagp2UUQ5+vLr7OMxSoWQGeVYE9VPFjy1pzYbT9H9w0kt5mV+KyjXRjwejUZ+gQk+01hqkA8+tsRXPANhKMO63jr+MZS6ncdHXoaZprrMCEiPRiN33IIvhUQRIxpRkI9O4Gva4ILXw8M++rrgCD4eBOz0CA9B28EaeGEffQeMoCGKGg8H6Khv6AwMYQPa4zeGHOqBF75lyPOaFKVaR1U8L0ATnjAfJ0sp8kuykoQoAzM2lmPPhLXzXIX45yzzsUVa2FirlHASRLFaPa3SKCEcDFGiEpxl0dcak9s1qWlrAZ00j3GzvKWZXk/Q94EohwySFLNJnaUXggmZ/H51fX3zxW0D5M8+Qy8kAQ2E0fgS5SIjDFlwO/ADeN5SxkC8gQ1Aw6qVyN4/q3bltN5w0bkQOHM/eqjAWyst50tJJ+ggSYbwN8N5wcgl1nhoN6qhc+9SCk7C52TFHg9CPiinOSQBfSJsRv8Gen84btH1ZJZnKWaw9GE0itFrJCkCLofx7iJpS/XaMuX4xKBAnjSBryOUEwgPKogs4J2mJcMy4HFD+eoa7L6PzNfz5b7FcYtxeAWnXPl4sArnJrZQKy/ya/x2fp1fz6/2iavAa4nTv0jmfQrSxW4zTSQtpQQnP/GMPNl+UhH/S8lTg9oUFoTI8GXj0CLQofbk4i3eIVPA1q3Kcu3Zsjbord2GpmyviGqu9joYhyjeVksAsmE0asE/hEaDvg16DCN8pdcPVe8JLWZUkrQy97IJqfLRcKej8OAmknE4ixq1yHe5lxa5kDlmUAWRPzluQPRbCE3ai7woO/jewH5o/zC7DMMEKz0Ay3o9KAsUXX2e94F+yKbQU1jpcnH++dPdn3+02nv7qC6lVAqlQg/K4n9x0ToxbgNw+v+K72a5VER3osxLmBYF253vTHPBcv+Y8ZuY3X9JdCl5xwGmFzS50AHr1QOeOyqsLDJTXr51BVcV28Vs92+Fp5oDvqqcBVdFTfW0i/td1dJKiUqhs/ZCYw+2bfhh4npMUjRqe4b89o7IWh+Myit9Ys9c39sDXr/xH1NGuWlBotRAIVWovios6RNaMgrdUaLHchXomE/tSXsoJH78ti6He1DiDmcegqjPFTR8AqxBw9xgZsDYAdZMreomY1pPkwRkQ+SuGoIQ9sBCFEQ+tvdEx8YM8iaFBsGIydc7Idgjluel1oLDxfnOjAaUmbm96MXIp1GdO++aIM/vOWnOrZH/ckoJKWVu75k7oxf3pkrvGDmrh+FPNC+g7cE9hUVwRdEErijm/j58LMGQTlKl6lKeDkPVaUY3iGanHb8cUMqwUrCyLJm9ryx6Z9Mh7H+hygSG3wurGwgJwzuzbX10du2ESZJMh/DarakdN3uW/wE) to test the original issue of a "flickering" entity when it moves (it should not flicker)

# Author checklist

### DIFF
--- a/packages/engine/Source/DataSources/DataSourceDisplay.js
+++ b/packages/engine/Source/DataSources/DataSourceDisplay.js
@@ -362,8 +362,80 @@ DataSourceDisplay.prototype._postRender = function () {
   }
 };
 
-const getBoundingSphereArrayScratch = [];
 const getBoundingSphereBoundingSphereScratch = new BoundingSphere();
+
+/**
+ * Computes a bounding sphere state.
+ *
+ * @param {Entity} entity The entity whose bounding sphere to compute.
+ * @param {boolean} allowPartial If true, pending bounding spheres are ignored and an answer will be returned from the currently available data.
+ *                               If false, the the function will halt and return pending if any of the bounding spheres are pending.
+ * @returns {BoundingSphereState} BoundingSphereState.DONE if the bounding sphere could be returned,
+ *                       BoundingSphereState.PENDING if the bounding sphere is still being computed, or
+ *                       BoundingSphereState.FAILED if the entity has no visualization in the current scene.
+ * @private
+ */
+DataSourceDisplay.prototype.getBoundingSphereState = function (
+  entity,
+  allowPartial,
+) {
+  //>>includeStart('debug', pragmas.debug);
+  Check.defined("entity", entity);
+  Check.typeOf.bool("allowPartial", allowPartial);
+  //>>includeEnd('debug');
+
+  if (!this._ready) {
+    return BoundingSphereState.PENDING;
+  }
+
+  let i;
+  let length;
+  let dataSource = this._defaultDataSource;
+  if (!dataSource.entities.contains(entity)) {
+    dataSource = undefined;
+
+    const dataSources = this._dataSourceCollection;
+    length = dataSources.length;
+    for (i = 0; i < length; i++) {
+      const d = dataSources.get(i);
+      if (d.entities.contains(entity)) {
+        dataSource = d;
+        break;
+      }
+    }
+  }
+
+  if (!defined(dataSource)) {
+    return BoundingSphereState.FAILED;
+  }
+
+  const tmp = getBoundingSphereBoundingSphereScratch;
+
+  let count = 0;
+  let state = BoundingSphereState.DONE;
+  const visualizers = dataSource._visualizers;
+  const visualizersLength = visualizers.length;
+
+  for (i = 0; i < visualizersLength; i++) {
+    const visualizer = visualizers[i];
+    if (defined(visualizer.getBoundingSphere)) {
+      state = visualizers[i].getBoundingSphere(entity, tmp);
+      if (!allowPartial && state === BoundingSphereState.PENDING) {
+        return BoundingSphereState.PENDING;
+      } else if (state === BoundingSphereState.DONE) {
+        count++;
+      }
+    }
+  }
+
+  if (count === 0) {
+    return BoundingSphereState.FAILED;
+  }
+
+  return BoundingSphereState.DONE;
+};
+
+const getBoundingSphereArrayScratch = [];
 
 /**
  * Computes a bounding sphere which encloses the visualization produced for the specified entity.
@@ -372,8 +444,7 @@ const getBoundingSphereBoundingSphereScratch = new BoundingSphere();
  * @param {Entity} entity The entity whose bounding sphere to compute.
  * @param {boolean} allowPartial If true, pending bounding spheres are ignored and an answer will be returned from the currently available data.
  *                               If false, the the function will halt and return pending if any of the bounding spheres are pending.
- * @param {BoundingSphere} [result] The bounding sphere onto which to store the result.
- *                                  Result parameter may not be provided if one is only interested into the bounding sphere state.
+ * @param {BoundingSphere} result The bounding sphere onto which to store the result.
  * @returns {BoundingSphereState} BoundingSphereState.DONE if the result contains the bounding sphere,
  *                       BoundingSphereState.PENDING if the result is still being computed, or
  *                       BoundingSphereState.FAILED if the entity has no visualization in the current scene.
@@ -387,6 +458,7 @@ DataSourceDisplay.prototype.getBoundingSphere = function (
   //>>includeStart('debug', pragmas.debug);
   Check.defined("entity", entity);
   Check.typeOf.bool("allowPartial", allowPartial);
+  Check.defined("result", result);
   //>>includeEnd('debug');
 
   if (!this._ready) {
@@ -443,9 +515,7 @@ DataSourceDisplay.prototype.getBoundingSphere = function (
   }
 
   boundingSpheres.length = count;
-  if (defined(result)) {
-    BoundingSphere.fromBoundingSpheres(boundingSpheres, result);
-  }
+  BoundingSphere.fromBoundingSpheres(boundingSpheres, result);
   return BoundingSphereState.DONE;
 };
 
--- a/packages/engine/Source/DataSources/DataSourceDisplay.js
+++ b/packages/engine/Source/DataSources/DataSourceDisplay.js
@@ -372,7 +372,8 @@ const getBoundingSphereBoundingSphereScratch = new BoundingSphere();
  * @param {Entity} entity The entity whose bounding sphere to compute.
  * @param {boolean} allowPartial If true, pending bounding spheres are ignored and an answer will be returned from the currently available data.
  *                               If false, the the function will halt and return pending if any of the bounding spheres are pending.
- * @param {BoundingSphere} result The bounding sphere onto which to store the result.
+ * @param {BoundingSphere} [result] The bounding sphere onto which to store the result.
+ *                                  Result parameter may not be provided if one is only interested into the bounding sphere state.
  * @returns {BoundingSphereState} BoundingSphereState.DONE if the result contains the bounding sphere,
  *                       BoundingSphereState.PENDING if the result is still being computed, or
  *                       BoundingSphereState.FAILED if the entity has no visualization in the current scene.
@@ -386,7 +387,6 @@ DataSourceDisplay.prototype.getBoundingSphere = function (
   //>>includeStart('debug', pragmas.debug);
   Check.defined("entity", entity);
   Check.typeOf.bool("allowPartial", allowPartial);
-  Check.defined("result", result);
   //>>includeEnd('debug');
 
   if (!this._ready) {
@@ -443,7 +443,9 @@ DataSourceDisplay.prototype.getBoundingSphere = function (
   }
 
   boundingSpheres.length = count;
-  BoundingSphere.fromBoundingSpheres(boundingSpheres, result);
+  if (defined(result)) {
+    BoundingSphere.fromBoundingSpheres(boundingSpheres, result);
+  }
   return BoundingSphereState.DONE;
 };
 
--- a/packages/engine/Source/DataSources/DataSourceDisplay.js
+++ b/packages/engine/Source/DataSources/DataSourceDisplay.js
@@ -362,80 +362,8 @@ DataSourceDisplay.prototype._postRender = function () {
   }
 };
 
-const getBoundingSphereBoundingSphereScratch = new BoundingSphere();
-
-/**
- * Computes a bounding sphere state.
- *
- * @param {Entity} entity The entity whose bounding sphere to compute.
- * @param {boolean} allowPartial If true, pending bounding spheres are ignored and an answer will be returned from the currently available data.
- *                               If false, the the function will halt and return pending if any of the bounding spheres are pending.
- * @returns {BoundingSphereState} BoundingSphereState.DONE if the bounding sphere could be returned,
- *                       BoundingSphereState.PENDING if the bounding sphere is still being computed, or
- *                       BoundingSphereState.FAILED if the entity has no visualization in the current scene.
- * @private
- */
-DataSourceDisplay.prototype.getBoundingSphereState = function (
-  entity,
-  allowPartial,
-) {
-  //>>includeStart('debug', pragmas.debug);
-  Check.defined("entity", entity);
-  Check.typeOf.bool("allowPartial", allowPartial);
-  //>>includeEnd('debug');
-
-  if (!this._ready) {
-    return BoundingSphereState.PENDING;
-  }
-
-  let i;
-  let length;
-  let dataSource = this._defaultDataSource;
-  if (!dataSource.entities.contains(entity)) {
-    dataSource = undefined;
-
-    const dataSources = this._dataSourceCollection;
-    length = dataSources.length;
-    for (i = 0; i < length; i++) {
-      const d = dataSources.get(i);
-      if (d.entities.contains(entity)) {
-        dataSource = d;
-        break;
-      }
-    }
-  }
-
-  if (!defined(dataSource)) {
-    return BoundingSphereState.FAILED;
-  }
-
-  const tmp = getBoundingSphereBoundingSphereScratch;
-
-  let count = 0;
-  let state = BoundingSphereState.DONE;
-  const visualizers = dataSource._visualizers;
-  const visualizersLength = visualizers.length;
-
-  for (i = 0; i < visualizersLength; i++) {
-    const visualizer = visualizers[i];
-    if (defined(visualizer.getBoundingSphere)) {
-      state = visualizers[i].getBoundingSphere(entity, tmp);
-      if (!allowPartial && state === BoundingSphereState.PENDING) {
-        return BoundingSphereState.PENDING;
-      } else if (state === BoundingSphereState.DONE) {
-        count++;
-      }
-    }
-  }
-
-  if (count === 0) {
-    return BoundingSphereState.FAILED;
-  }
-
-  return BoundingSphereState.DONE;
-};
-
 const getBoundingSphereArrayScratch = [];
+const getBoundingSphereBoundingSphereScratch = new BoundingSphere();
 
 /**
  * Computes a bounding sphere which encloses the visualization produced for the specified entity.
--- a/packages/engine/Source/DataSources/EntityView.js
+++ b/packages/engine/Source/DataSources/EntityView.js
@@ -2,7 +2,6 @@ import Cartesian3 from "../Core/Cartesian3.js";
 import Check from "../Core/Check.js";
 import defaultValue from "../Core/defaultValue.js";
 import defined from "../Core/defined.js";
-import deprecationWarning from "../Core/deprecationWarning.js";
 import Ellipsoid from "../Core/Ellipsoid.js";
 import HeadingPitchRange from "../Core/HeadingPitchRange.js";
 import JulianDate from "../Core/JulianDate.js";
@@ -44,7 +43,7 @@ function updateTransform(
   ellipsoid,
 ) {
   const mode = that.scene.mode;
-  const cartesian = positionProperty.getValue(time, that._lastCartesian);
+  let cartesian = positionProperty.getValue(time, that._lastCartesian);
   if (defined(cartesian)) {
     let hasBasis = false;
     let invertVelocity = false;
@@ -217,6 +216,10 @@ function updateTransform(
       }
     }
 
+    if (defined(that.boundingSphere)) {
+      cartesian = that.boundingSphere.center;
+    }
+
     let position;
     let direction;
     let up;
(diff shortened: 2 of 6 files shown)


ggetz:
Thank you for the fix @jfayot!

packages/engine/Source/DataSources/EntityView.js
```diff
@@ -346,12 +342,6 @@ function EntityView(entity, scene, ellipsoid) {
    */
   this.ellipsoid = defaultValue(ellipsoid, Ellipsoid.default);
 
-  /**
-   * The bounding sphere of the object.
-   * @type {BoundingSphere}
-   */
-  this.boundingSphere = undefined;
```
jjspace, line 353: Can you explain why this was removed? More than it just works now? The correlation between this and the scratch value and the entities actual bounding sphere all still feels a little funky to me.

If we _do_ still want to remove this we will need to go through the [deprecation process](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#deprecation-and-breaking-changes) as this is part of the public API and people may be relying on it.

jfayot: I have no answer for you @jjspace.

jfayot: @jjspace I recognize that I have no idea why "it just works now!"

Even this bounding sphere parameter inside `EntityView.update` looks weird: why would you offer this as a parameter when there is no way to get a bounding sphere from an `Entity` in the public API?


In #3954, the bounding sphere was moved from `EntityView` constructor to `update` method, but again, why would you impose to provide a bounding sphere to `EntityView` (which is related to the high level Entity API) when you don't have the ability to get this bounding sphere from Entity's high level API?


Before #3954, the bounding sphere parameter in the constructor was documented as:
`@param {BoundingSphere} [boundingSphere] An initial bounding sphere for setting the default view.`

This "initial" aspect has completely disappeared. My guess is that it should be restored, AND that the Entity's bounding sphere should be computed from inside the `EntityView` implementation and should disappear from the public API.


Otherwise provide a public `Entity.computeBoundingSphere` method...

jjspace:
@jfayot I'm sorry this took so long to get back to. It turned out to be a bit of a "land mine" in the code for me to try (read: struggle) to wrap my head around...

I did some digging into this to try and understand it more completely. First I'll say this whole sequence of functions is a total mess imo with way to many "hard to follow" side effects. Ideally it'd be nice to do a refactor pass on all this camera tracking logic but that's way to big for this PR right now (the camera system itself needs some love too)

<sup>Sorry this turned into a wall of text, I tried to condense but it's important context that I think needs written down somewhere</sup>

The first important thing to realize is that `EntityView.update()` behaves differently the first/second time it's called vs the subsequent calls that happen every frame. `updateTransform` is the same way because of the `updateLookAt` and `saveCamera` arguments that are potentially only different the "first" time when the `objectChanged`. <sub>Side note, the way we're currently using `objectChanged` will _never_ be true after the first time because we always regenerate the `EntityView` completely instead of modifying it</sub>

The `EntityView.boundingSphere` is only set the _first_ time `EntityView.update()` is called (`objectChanged`) _and_ the entity does _not_ have a `viewFrom` property. This is what was causing the satellite tracking to fail because that has a `viewFrom` in the sandcastle so the `EntityView.boundingSpehere` is never set. This caused my previous "fix" to `CesiumWidget._onTick` (in #12429) to never run the `update` calls because `defined(entityView.boundingSphere)` was false. I added this check in the first place because `DataSourceDisplay.getBoundingSphere` requires the bounding sphere to be defined.

If `EntityView.boundingSphere` _is_ defined then it's center will be used in the `updateTranform` function instead of the other transform logic. This is why it was important to call `dataSourceDisplay.getBoundingSphere` _with the exisiting_ `entityView.boundingSphere` because that function updates that bounding sphere itself which is the same reference inside the `EntityView`. This wasn't a problem before #124929 because they were all using the same scratch variable in `CesiumWidget` (in `_onTick` and `updateTrackedEntity`) so it _was_ always referring to the same object.

Given all that, an alternative fix for this that works for all the provided examples without changing the `EntityView` could be this:

```js

const boundingSphereScratch = new BoundingSphere();

/**
 * @private
 */
CesiumWidget.prototype._onTick = function (clock) {
  const time = clock.currentTime;

  const isUpdated = this._dataSourceDisplay.update(time);
  if (this._allowDataSourcesToSuspendAnimation) {
    this._canAnimateUpdateCallback(isUpdated);
  }

  const entityView = this._entityView;
  if (defined(entityView)) {
    const trackedEntity = this._trackedEntity;
    const trackedState = this._dataSourceDisplay.getBoundingSphere(
      trackedEntity,
      false,
      entityView.boundingSphere ?? boundingSphereScratch,
    );
    if (trackedState === BoundingSphereState.DONE) {
      entityView.update(time, undefined);
    }
  }
};
```

This is a more minimal code change. However, in some ways this feels pretty bad too and requires understanding all the interactions in how we're calling `update` and it's internal logic that I outlined above.

I think removing `EntityView.boundingSphere` as you've done could be the right move. This would remove one of the "side effect-ish" behaviors in `updateTransform` relying on `that.boundingSphere`. This will (slightly) solidify the logic that **`EntityView.update` sets the _initial_ location of the camera then subsequent calls to `updateTransform` update the camera relative to how the entity has moved** regardless of the bounding sphere.
If this is the approach we take it probably also makes sense to replace the call to `getBoundingSphere` to a new function that only returns the status for `CesiumWidget._onTick` since that's all we really care about. Or maybe just make the `result` optional.
If this makes the most sense to you too please add the deprecation message to the removed property and update `getBoundingSphere` as/if needed. While it's being deprecated we might need to keep setting it even if we're not using it internally so the scratch variables still need to be updated as required.

Alternatively we could keep `EntityView.boundingSphere` and make sure it's position is _always_ used even for `viewFrom` entities but that seems like a bigger change that may not be necessary.

jfayot:
Thanks for the analysis @jjspace !

> If this is the approach we take it probably also makes sense to replace the call to getBoundingSphere to a new function that only returns the status for CesiumWidget._onTick since that's all we really care about. Or maybe just make the result optional.

I'm not sure to understand this: is it the call `this._dataSourceDisplay.getBoundingSphere` inside `CesiumWidget.prototype._onTick` that you want to change? If so, then I don't get the idea, because in my understanding, `EntityView.prototype.update` still needs an initial bounding sphere in order to make the call `camera.viewBoundingSphere` work properly...

jjspace:
> I'm not sure to understand this: is it the call this._dataSourceDisplay.getBoundingSphere inside CesiumWidget.prototype._onTick that you want to change?

@jfayot Yes that's where I meant. We really only care about the _return_ value from that function. We need it to update the passed in `boundingSphere` _only if_ we're still using that bounding sphere in the `EntityView#updateTransform` function to set the `cartesian`

> in my understanding, `EntityView.prototype.update` still needs an initial bounding sphere in order to make the call `camera.viewBoundingSphere` work properly...

This is one of the things that makes this whole sequence of code so confusing. If you actually follow the code through the way it's run the call to `camera.viewBoundingSphere` only happens when `objectChanged` (functionally only the "first" time) _and_ the entity does not have a `viewFrom` property. It's only called in one of multiple code paths and is never called after the first time. You can prove this by adding a log statement in `camera.viewBoundingSphere` and loading the sample sandcastles to check how many times it's called. Also the `boundingSphere` argument is already marked as optional so it's not required

jfayot:
> Also the boundingSphere argument is already marked as optional so it's not required

Yes you're right, the bounding sphere parameter doesn't change anything!
So if we push the reasoning ahead, the whole code block inside `if (!hasViewFrom && defined(boundingSphere))` statement can be removed...
And if so, the `saveCamera` variable is always true and can also be removed...

Am I getting it right?

jfayot:
> If this is the approach we take it probably also makes sense to replace the call to getBoundingSphere to a new function that only returns the status for CesiumWidget._onTick since that's all we really care about. Or maybe just make the result optional.

@jjspace I get your point now ! However, while testing the different code path, I'm convinced that the call to `getBoundingSphere` is useless inside `_onTick`, which could be rewritten like this:

```javascript
CesiumWidget.prototype._onTick = function (clock) {
  const time = clock.currentTime;

  const isUpdated = this._dataSourceDisplay.update(time);
  if (this._allowDataSourcesToSuspendAnimation) {
    this._canAnimateUpdateCallback(isUpdated);
  }

  const entityView = this._entityView;
  if (defined(entityView)) {
    entityView.update(time);
  }
};
```

While, off course, keeping the bounding sphere call and parameter in function `updateTrackedEntity`:

```javascript
  const bs =
    state !== BoundingSphereState.FAILED
      ? trackedEntityBoundingSphereScratch
      : undefined;
  widget._entityView = new EntityView(trackedEntity, scene, scene.ellipsoid);
  widget._entityView.update(currentTime, bs);
```

Concerning the deprecation of `EntityView.boundingSphere`, two questions:
1. in which release would you like to see it completely removed?
2. Although it is in the public API, it is not visible in EntityView's documentation. Should I just add a deprecation note in the header code which will not be visible unless people dig into the sources:

```javascript
  /**
   * The bounding sphere of the object.
   * @type {BoundingSphere}
   * @deprecated This property has been deprecated and will be removed in Cesium 1.xxx
   */
  this.boundingSphere = undefined;
```

... or should I make move this to `Object.defineProperties(EntityView` in order to make its documentation visible?

```javascript
Object.defineProperties(EntityView, {
  /**
   * The bounding sphere of the object.
   * @type {BoundingSphere}
   * @deprecated This property has been deprecated and will be removed in Cesium 1.xxx
   */
  boundingSphere: {
    get: function () {
      deprecationWarning(
        "EntityView.boundingSphere",
        "EntityView.boundingSphere has been deprecated and will be removed in Cesium 1.xxx",
      );
      return this._boundingSphere;
    },
    set: function (boundingSphere) {
      deprecationWarning(
        "EntityView.boundingSphere",
        "EntityView.boundingSphere has been deprecated and will be removed in Cesium 1.xxx",
      );
      this._boundingSphere = boundingSphere;
    },
  },
});
```

jjspace:
> I'm convinced that the call to `getBoundingSphere` is useless inside `_onTick`

@jfayot I think it might still be needed simply to make the check against the `BoundingSphereState.DONE` to make sure the entity is actually loaded and ready? if not then I agree this can probably be simplified down. Maybe even as far as the single line version `this.entityView?.update(time)`?

> Concerning the deprecation of `EntityView.boundingSphere`, two questions:

1. Let's do 3 releases. The release on monday will be 1.127 so 1.130 for this one
2. Just to note, it _is_ in the docs for [`EntityView`](https://cesium.com/learn/cesiumjs/ref-doc/EntityView.html). But the second example you posted is better, convert this to a getter/setter and display the deprecation message there. All that matters is that from the perspective of people _consuming_ this API it appears to work the same even if the implementation changed. Also with that change I would make sure to change the place it's currently being set to use the private `_boundingSphere` property so we're not throwing the warning unless people are actually interacting with it themselves.

Please also remember to update the specs to cover these use cases so we can remain confident the tracking is working as expected in all cases

jfayot:
This is strange, I cannot see the property from the built documentation:

Capture d’écran 2025-02-28 à 17 30 15

Am I missing something?

jjspace:
@jfayot is that the local docs on this branch where you already removed the property? If so then that's why it's not there! Check the [hosted version](https://cesium.com/learn/cesiumjs/ref-doc/EntityView.html) or locally in `main`. Running `npm run build-docs` will rebuild the local docs.

jfayot:
> @jfayot is that the local docs on this branch where you already removed the property? If so then that's why it's not there! Check the [hosted version](https://cesium.com/learn/cesiumjs/ref-doc/EntityView.html) or locally in `main`. Running `npm run build-docs` will rebuild the local docs.

Just a wrong jsdoc implementation on my side... sorry for that!

jfayot:
@jjspace this should be ready for another review pass (although still the same test failures)

jfayot:
@jjspace when you'll have the time, could you also answer my questions from #11674 please? Thanks in advance!

packages/engine/Source/DataSources/DataSourceDisplay.js
```diff
@@ -372,7 +372,8 @@ const getBoundingSphereBoundingSphereScratch = new BoundingSphere();
  * @param {Entity} entity The entity whose bounding sphere to compute.
  * @param {boolean} allowPartial If true, pending bounding spheres are ignored and an answer will be returned from the currently available data.
  *                               If false, the the function will halt and return pending if any of the bounding spheres are pending.
- * @param {BoundingSphere} result The bounding sphere onto which to store the result.
+ * @param {BoundingSphere} [result] The bounding sphere onto which to store the result.
+ *                                  Result parameter may not be provided if one is only interested into the bounding sphere state.
```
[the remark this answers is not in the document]
jfayot: @ggetz I'm not sure to clearly understand your concern: if no `result` is passed to `DataSourceDisplay.getBoundingSphere` then the test by the end of the function should handle this correctly?

```javascript
  ...
  boundingSpheres.length = count;
  if (defined(result)) {
    BoundingSphere.fromBoundingSpheres(boundingSpheres, result);
  }
  return BoundingSphereState.DONE;
};
```

What am I missing?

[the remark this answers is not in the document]
ggetz: Ah, I missed that. Because our library has the convention that when a result is undefined, a new object is created, I assumed this did the same.

Perhaps only to avoid assumptions like mine, we should still create a new function that matches our convention then.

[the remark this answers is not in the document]
jfayot: @ggetz I understand what you mean.

However, optimizing ONLY the returned value wouldn't be of great benefit IMO.
If you look at the function, you'll see that it loops over the vizualizers of the data source and call another getBoundingSphere on each one. 
If you would really like to optimize and save created objects and GC, then this would require you to create getBoundingSphereState methods for each vizualizer...
Then some vizualizers themselves rely on updaters to compute their bounding spheres... so again you have to create getBoundingSphereState on updaters...

I'm not sure all these efforts (and associated regression risk) are worth this PR.

However, in order to match the convention that when a result is undefined, a new object is created, I have duplicated code inside getBoundingSphere to create getBoundingSphereState as suggested.
As you can see, we're still creating objects in this function because of the vizualizers. And I was not able to make getBoundingSphere rely on getBoundingSphereState without breaking too much existing code.

CC @jjspace

ggetz (approved):

packages/engine/Source/DataSources/DataSourceDisplay.js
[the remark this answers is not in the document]
ggetz: > I'm not sure all these efforts (and associated regression risk) are worth this PR.

Thanks @jfayot, I agree. I only wanted to make sure we were being careful about new code added.

ggetz:
The code here, and all of the testing plan looks great to me.

I did notice some trouble tracking an entity which is clamped to ground / 3D Tiles on initial load in [the Clamp Entities to Ground Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Clamp%20Entities%20to%20Ground.html). It loads so that the camera is underground for me. Does anyone else see the same?

lukemckinstry:
> I did notice some trouble tracking an entity which is clamped to ground / 3D Tiles on initial load in [the Clamp Entities to Ground Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Clamp%20Entities%20to%20Ground.html). It loads so that the camera is underground for me. Does anyone else see the same?

I see identical behavior on local vs prod 

https://github.com/user-attachments/assets/79f222b1-7b18-48c0-a00d-2e232e4aba92

jfayot:
@ggetz I can see it too...

lukemckinstry:
I see it too, I was not set up correctly before

jfayot:
@jjspace I know you won't like it, but it looks like this `EntityView.boundingSphere` is still necessary : I have a fix, but I can't explain why "it just works now"...

jfayot:
@ggetz, @jjspace roughly, the fix consists in reintroducing the boundingSphere... should I go ahead (that is, removing all the deprecation parts)? Or are you following another track?

ggetz:
Given there is still some uncertainty here, let's hold off on getting this into the release today. Let's make sure we're fixing this once and for all. :+1:

jjspace:
@ggetz @jfayot I believe the issue is that `updateTransform` starts with a `cartesian` set to the value of the entity's position at a given `time`. But for stationary entities that are clamped this value never actually changes so the rest of the `updateTransform` logic always returns the same location. However the clamping logic _does_ change the location of an entity's `boundingSphere`. The previous logic which used the `EntityView.boundingSphere.center` and called `DataSourceDisplay.getBoundingSphere` to update said value would account for this to ensure we are always looking at the most up to date bounding sphere even if it was modified through other processes

If you switch to `main` and look closely the camera starts on the label then jumps around _with the label_ as the G3DT gets loaded in and the clamping logic can update it's position

jfayot:
> #12495 changed handling asynchronous updates for billboards and labels. Perhaps it "fixed" something related to timing here?

@ggetz I've tested this branch rebased on top of 1.126 (so without #12495) and I still can see the camera going underground.

I do agree with your analysis @jjspace, we can clearly see the camera jumping quickly to the new entity's bounding sphere.

packages/engine/Specs/DataSources/EntityViewSpec.js
```diff
@@ -89,43 +89,30 @@ describe(
...
     });
 
-    it("uses entity bounding sphere", function () {
-      const sampleOffset = new Cartesian3(
-        -1.3322676295501878e-15,
-        -7.348469228349534,
-        7.3484692283495345,
-      );
+    it("uses provided bounding sphere", function () {
+      const bs = new BoundingSphere(new Cartesian3(3, 4, 5), 6);
```
jjspace, line 93: Can you add a test that modifies the bounding sphere to prove that the camera will follow it on future calls to `update`?

(I didn't check all the other specs so if it already exists just lmk)

jjspace (commented):
Thanks for the changes @jfayot, sorry this couldn't make it out with the release yesterday. I can confirm the 3 previous sandcastle examples and the clamping one are all now working correctly as expected so I _think_ this is probably working how we want it now :crossed_fingers:

Just had one comment about the specs.
Also @ggetz would you be able to re-test as well just to double check?

packages/engine/Specs/DataSources/EntityViewSpec.js
jfayot: @jjspace, done! This should be ready for another review

packages/engine/Specs/DataSources/EntityViewSpec.js
```diff
@@ -89,43 +89,52 @@ describe(
...
+      const entity = new Entity();
+      entity.position = new ConstantPositionProperty(
+        Cartesian3.fromDegrees(0.0, 0.0),
       );
-      expect(view.scene.camera.position).toEqualEpsilon(sampleOffset, 1e-10);
+      const view = new EntityView(entity, scene, undefined);
+      view.update(JulianDate.now(), bs1);
+      expect(scene.camera.positionWC).toEqualEpsilon(positionWC1, 1e-10);
+
+      view.boundingSphere = bs2;
```
jjspace, line 135: This test is valuable to make sure changing `EntityView.boundingSphere` works but doesn't mimic the behavior of changing the actual bounding sphere it was set up with originally like calling `getBoundingSphere` does. Maybe another that replaces this line with `bs1.center = new Cartesian3(3, 4, 5);` would also be valuable.

jjspace (approved):
Thanks for the changes @jfayot. One more small comment and I think this can get merged.

@ggetz would you also be able to take one last runthrough of tests just to make sure we're not missing anything?

packages/engine/Specs/DataSources/EntityViewSpec.js
jfayot: Good catch @jjspace. Added new test. Should be ready for a final (hopefully :sweat_smile:) review!

ggetz:
Thanks @jfayot and @jjspace! This is looking great. I'm glad the code change boiled down to a simple solution.

I've run through manual testing for the problematic examples we've seen previously, and have done an overview of other Sandcastle examples. Everything is working great!

ggetz (approved):

Issue #12373, opened 2024-12-13
MOUSE_MOVE fires outside of canvas in Firefox


angrycat9000:
### What happened?

Inconsistent behavior of ScreenSpaceEventHandler with MOUSE_MOVE event between Firefox and Chrome.

In Chrome when the mouse moves out of the canvas and over another element the MOUSE_MOVE events stop.  In Firefox the MOUSE_MOVE events continue and can have coordinates outside the range of the canvas.

My expectation is that the MOUSE_MOVE event is only triggered while the pointer is moving over the canvas element, and is not fired when the pointer is outside of the canvas.

https://github.com/user-attachments/assets/f4259184-29e5-4175-971b-58578d1c65da

### Reproduction steps

1. Open the below sandcastle in Chrome. 
2. Open the Console tab
3. Move the mouse over the Cesium viewer and then over the red element
4. Note that the console logs stop when the pointer moves over the red element
5. Open sandcastle in Firefox
6. Open the Console tab
7. Move the mouse over the Cesium viewer and then over the red element
8. Note that the console logs continue while over the red element, and the y value is negative.

### Sandcastle example

https://sandcastle.cesium.com/#c=dVFta8IwEP4roftSQVLGvmktGyJsMHHg5qfCiMlNg+mlJJeKDP/7mlYnOBdIyHP3vFyItOiJNRr24NiEIezZFLwOFV91tbRMZIenFkloBFcmg3GJstNtBSpzLVxKB4DLWkiYNYD03JPSPoR7CQhcCmyEj04nD+6BXrAO9CRJW0y/AnYXlla2gar1GbDvElm3Yro1wI3d/LY5oHqzXkdR9I284/C/od4PNfD54mM5+5wvVrMoSIZJ7ulgoDjHPOqqto5YcCblPCOoaiMIfLYOcgfEpfd90llwZwN5reAyKWO1UErjZsQeHFTjS30t5G7jbEA1cqC6xrHv5tl5jFzphmk1KZOTcZkU8/a9rN2O0RYcxBN7HGGetZLTAy7qP19Y3OYZK+Ksi9bNiEOkbe+L177IOc+zFt5WkrVmLa6dfwA

### Environment

Browser:
- Chrome 131.0.6778.140
- Firefox 133.0.3
CesiumJS Version: 1.124
Operating System: Windows 11 Enterprise 22631.4602

eclctc:
@ggetz @javagl  - I would like to request working on this issue please for FCF

[...]

Issue #7003, opened 2018-09-04
Re-enable PointerEvents in Firefox once they fix their implementation


mramato:
In #6999 we disabled PointerEvents in Firefox completely because their implementation seems broken.  There is some discussion of this in #6372 (along with reproducible bugs)

Once Firefox improves their implementation, we should re-enable it (and confirm none of the bugs in the linked issue still exists)

thw0rted:
You might want to follow https://bugzilla.mozilla.org/show_bug.cgi?id=1463354 which I filed, oh wow, half a year ago 😞 , with no updates for 2 months now 😭

mramato:
Thanks for adding the link @thw0rted.  That's definitely the issue I keep track of.  Glad to see it looks like there's been some progress at least.

bkuster:
Hi there! I just stumbled across this, since our application was misbehaving in FF. In short: the `ScreenSpaceEventHandler` was firing events _for a hidden canvas_ (using `stylde.display = 'none'`). Something it did not do in any other browser. With some digging I came on to this: 

https://github.com/CesiumGS/cesium/blob/c5abf0d7ad4a9b7a23f8f1fb4e55b13a4a15d2af/Source/Core/FeatureDetection.js#L176-L177

Deleting line 177 has solved the problem for me. Is there any reason to keep FF out of the pointer event API anymore? Following @thw0rted s issue with the FF team, the pointer events should be working again as of FF 67?

Issue #6372, opened 2018-03-26
Firefox 59.0 on macOS ctrl+touchpad rotation not working


andrzejfigula:
Latest Firefox broke navigation control (ctrl + touchpad to rotate) for macbook.
This was only checked with touchpad, not sure if affects other pointing devices.

Environment:
- macOS (Sierra) on MacBook Pro (Retina, 15-inch, Mid 2014)
- Firefox 59.0.2 (64-bit)
- Cesium 1.43 (Cesium home page and my application)

Issue:
I'm always using touchpad to rotate (**ctrl + touch**), it was always possible on FF until last update.
(Chrome and Safari works fine). 

Tested on Cesium home page and with Cesium 1.43

Looking at FF release notes:
https://www.mozilla.org/en-US/firefox/59.0/releasenotes/

I found this:
"Added support for W3C specs for pointer events and improved platform integration with added device support for mouse, pen, and touch screen pointer input"

Just guessing this might be the reason.

hpinkos:
Thanks for reporting this @andrzejfigula!
@ggetz can you confirm this bug on your mac?

jimmyangel:
I can confirm this issue as well on Firefox 59.0.2 MacOS 10.13.4, Cesium 1.44. Control-drag not working work with a trackpad nor a mouse either.

hpinkos:
Thanks @jimmyangel.  Can someone with a mac confirm this bug?  @shehzan10 @tfili @shunter?

shrichards:
I'm also seeing this on my MacBook Pro (exact same specs as @andrzejfigula) with Cesium 1.44 & FireFox v60.0b13 beta.

With a mouse hooked up, ctrl-leftclick-drag also doesn't work.  However, ctrl-*rightclick*-drag does rotate.

hpinkos:
Thanks @shrichards, we'll try to look into this soon

DronePhil:
@hpinkos any updates here? At DroneDeploy we are starting to see customer complaints about this issue. Though we can fix it by rewriting the camera controls ourselves, we'd prefer not to.

mramato:
I believe Firefox has several bugs in the PointerEvents implementation, we've encountered this before in our own unit tests.  I think the best solution here is to disable PointerEvents completely in FireFox until those issues get fixed.

I'll push a branch with this change and if folks can confirm that it fixes the issue, we'll merge for the next release.

mramato:
[...]

PR #12473, opened 2025-02-10
Mousemove fires outside canvas

Riley-Howley:
# Description

Fixes Issue: #12373 
This PR addresses adding a bounds check on the `ScreenSpaceEventHandler` `handleMouseMove` this issue was only reproducible in Firefox. Checking to make sure the client is inside the bounds of the canvas solves this issue

When integration Cesium in existing application, if menus are overlayed on the canvas then in firefox the event is triggered even though its not in the canvas

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12373

## Testing plan

Using the given example in the Issue, after my change I tested in Firefox, and Chrome. Note: I did this on a Mac.

Automated tests were updated to have a valid `boundingClientRect` so that it passes the new check. Also added a new test to ensure the event is not fired when in the bounds of the canvas

# Author checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -423,3 +423,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Nick Crews](https://github.com/NickCrews)
 - [胡文康](https://github.com/XiaoHu1994)
 - [Parth Petkar](https://github.com/parthpetkar)
+- [Riley Howley](https://github.com/Riley-Howley)
--- a/packages/engine/Source/Core/ScreenSpaceEventHandler.js
+++ b/packages/engine/Source/Core/ScreenSpaceEventHandler.js
@@ -352,6 +352,18 @@ function handleMouseMove(screenSpaceEventHandler, event) {
     return;
   }
 
+  const canvas = screenSpaceEventHandler._element;
+  const rect = canvas.getBoundingClientRect();
+
+  if (
+    event.clientX < rect.left ||
+    event.clientX > rect.right ||
+    event.clientY < rect.top ||
+    event.clientY > rect.bottom
+  ) {
+    return;
+  }
+
   const modifier = getModifier(event);
 
   const position = getPosition(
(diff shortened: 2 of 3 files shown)


ggetz:
Thanks for the PR @Riley-Howley!

@jjspace Could you please revie?

jjspace (changes requested):
This ended up being an annoyingly deep rabbit hole to dig into. I think your solution does directly fix the original problem in the issue but I don't think it's actually the true cause of confusion.

The problem originally comes from the `registerListeners` function which checks whether the browser `supportsPointerEvents` and if it does _not_ it sets up a set of mouse move events instead of pointer events. This was turned off for Firefox completely because their early implementation of pointer events didn't work correctly, see https://github.com/CesiumGS/cesium/issues/6372. However these events behave differently which is causing the change in behavior between browsers.

When using mouse events we attach the `mousemove` and `mouseup` events to the full `document` instead of the `canvas` specifically so we can handle events when dragging outside of the canvas. I think this is (and should stay) the desired behavior because it feels really bad if I click inside the viewer and drag outside of it and have the drag canceled immediately when I leave the viewer. This is not a problem with pointer events because after a `pointerdown` event `pointermove` events are "re-targeted" into the target of the `pointerdown` (the `canvas`). This lets the drag behavior work as expected without attaching event handlers to the `document`.

1. I think we can enable firefox to use pointer events instead of mouse move events which are simpler and behave more how we'd expect https://github.com/CesiumGS/cesium/issues/7003
2. When `registerListeners` _does_ register the set of listeners for a browser that does not support pointer events it should behave completely the same as when it does. This means adding a check similar to the one you already did but being more selective for when it triggers
    - Mouse move events that are outside the `canvas` _while currently part of a drag event_ should still trigger
    - Mouse move events that are outside the `canvas` when _not_ part of a drag event should _not_ trigger the handler

Hopefully this all makes sense, happy to talk through things more if you want

jjspace:
Also it looks like linting is failing. Please run `npm run prettier` and also make sure your git hooks are set up to correctly lint all code before it's pushed up to avoid CI failing.

ggetz:
@jjspace What your read on this PR? Is this something we still want to move forward with?

jjspace:
@ggetz I think we do need to address the initial problem but I do not think this PR, in it's current form, is the correct approach. See my comment above. Given the lack of activity it may make sense to just close this for now until someone has time to swing back around to it.

ggetz:
Make sense. Thanks @Riley-Howley and @jjspace.

@Riley-Howley let us know if you have plans to get back to this PR, and we can discuss the approach to get this fixed.

PR #12206, opened 2024-09-18
Update prettier to v3

jjspace:
# Description

Updates prettier to v3.3.0 which crosses the major version 3 [breaking changes](https://prettier.io/blog/2023/07/05/3.0.0.html#breaking-changes)

This is a _very_ big pr touching nearly all our source files but all the changes are basically identical.

Key changes:

- Trailing commas are now always included (a very good change IMO), this is a majority of the changes
- greatly improved line wrapping logic for readability (see below)
- prefer lowercase `doctype` over `DOCTYPE`

```
// before 3.0 it preferred parens and brackets
const tileset = await Cesium.Cesium3DTileset.fromIonAssetId(
  1
);

// after 3.0 it actually somewhat uses context, much more readable imo
const tileset =
  await Cesium.Cesium3DTileset.fromIonAssetId(1);

```

I also added a new `.prettierrc` specifically for the Sandcastle Gallery to allow a slightly wider line with to account for the indentation that's removed when sandcastles are actually rendered. This should produce code that matches prettier's default 80 when in the browser editor.

When we decide to merge this I'll add some new git tags and post a guide on how to update other branches easier.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12172

## Testing plan

- Run `npm run prettier` and make sure there are no changed files
    - or just run `npm run prettier-check` and make sure it doesn't fail
- make sure specs still pass
- functionally everything should be the same so primarily review the files themselves.

# Author checklist

- [ ] I have submitted a Contributor License Agreement
- [ ] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [ ] I have performed a self-review of my code

PR #12481, opened 2025-02-18
Pin prettier version

jjspace:
# Description

Every version bump in `prettier` has a chance to change the desired output and thus break our linting. This PR changes it back to a specific pinned version instead of letting npm and semver take over.

Specifically a [recent change](https://github.com/prettier/prettier/pull/16907) in 3.5.0 was a ["breaking" change](https://github.com/prettier/prettier/issues/17094) for us and changed the desired format of a css file we haven't touched in a a while. It looks like @jjhembd addressed this in https://github.com/CesiumGS/cesium/commit/1c892c87bd3f6eae234379bcf13344f5d0db6e59 so merging main will work but branches may be "outdated" and break in CI.

We _should_ keep the prettier version updated so we don't have to do as many big refactors like #12206 but it should be intentional instead of randomly when people run `npm install` or when CI installs a different version than people have locally. I did check and pinning the version still has it show up in `npm outdated` so the release process should still catch it.

## Issue number and link

no issue

## Testing plan

- `rm -rf node_modules`
- `npm install`
- `npm run prettier-check` should pass

# Author checklist

### DIFF
--- a/Documentation/Contributors/ReleaseGuide/README.md
+++ b/Documentation/Contributors/ReleaseGuide/README.md
@@ -8,6 +8,7 @@ There is no release manager; instead, our community shares the responsibility. A
 
 1. Check for any outdated dependencies with `npm outdated`.
 2. If one or more dependencies are outdated, checkout a new branch and run `npm install <packagename>@latest` for each package to increment the version.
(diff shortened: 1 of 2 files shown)


ggetz (approved):

ggetz:
Thanks @jjspace!

PR #12493, opened 2025-02-24
OGC Features Provider and Features Layer

jjspace:
# Description

Opening as a draft for early review and feedback around implementation. Will fill out the description more later.

See https://github.com/CesiumGS/cesium/issues/12456

Also worth taking a look at @ggetz's [`feature-prototype`](https://github.com/CesiumGS/cesium/compare/feature-prototype?expand=1) branch for how we might do some of the lower level stuff.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/12456

## Testing plan

- There's a [new development sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=development%2FOGC%20Feature%20Provider.html) for the `OgcFeatureProvider` and `FeatureLayer` classes

# Author checklist


packages/engine/Source/DataSources/FeatureProvider.js
```diff
@@ -0,0 +1,42 @@
+import defined from "../Core/defined.js";
+import FeatureLayer from "./FeatureLayer.js";
+
+/**
+ * A class for things
+ * @constructor
+ * @param {Object} options
+ */
+function FeatureProvider(options) {
+  this.id = options.id;
```
ggetz, line 10: We've had some pain points with inheritance in the past. To avoid these, I would recommend treating `FeatureProvider` as an interface as much as possible. That is, I would avoid defining _any_ implementation details, and leave it to the implementor or static utility-like functions.

packages/engine/Source/DataSources/OgcFeatureProvider.js
```diff
@@ -0,0 +1,307 @@
...
+  this._nextLink = json.links?.find((link) => link.rel === "next")?.href;
+
+  partialLoadCallback(json);
+
+  while (
+    defined(this._nextLink) &&
+    // this._itemsLoaded < this.maxItems
+    itemsLoadedThisRequest < this.maxItems
+  ) {
+    if (localAbortController.signal.aborted) {
```
ggetz, line 237: To clarify my understanding, the abort controller as it is used in this class is not actively cancelling an in-progress requests— Only preventing more requests from happening. Is that correct?

As a follow up question, is this something important we need to consider for the `Resource` implementation? Would it fit in with the existing implementation, or would it necessitate more significant refactoring?

packages/engine/Source/DataSources/FeatureLayer.js
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * @import FeatureProvider from "./FeatureProvider";
+ * @import FrameState from "../Scene/FrameState";
+ * @import OGCFeaturesProvider from "./OgcFeatureProvider.js";
+ */
+
+/**
+ * A class for things
+ * @constructor
+ * @param {FeatureProvider | OGCFeaturesProvider} provider
```
ggetz, line 21: If `OGCFeaturesProvider` implements `FeatureProvieder`, I don't think we'd need to specify the type explicitly, no?

packages/engine/Source/DataSources/FeatureLayer.js
```diff
@@ -0,0 +1,213 @@
...
+  const { positionWC: cameraPosition } = camera;
+
+  const distance = Math.sqrt(
+    Cartesian3.distanceSquared(
+      this._previousCameraPosition ?? new Cartesian3(),
+      cameraPosition,
+    ),
+  );
+
+  if (distance > 400) {
```
ggetz, line 171: There are camera moved events on `Camera.js`. Perhaps they are sufficient?

packages/engine/Source/DataSources/FeatureLayer.js
```diff
@@ -0,0 +1,213 @@
...
+  // this._dataSource = new CustomDataSource();
+  // // this._primitives = new EntityCollection();
+  // /** @type {EntityCollection} */
+  // this._primitives = this._dataSource.entities;
+  // this._dataSource = new GeoJsonDataSource(options.id);
+  this._customDataSource = new CustomDataSource();
+  // this._primitives = this._dataSource.entities;
+  this._primitives = [];
+
+  this.features = options.features ?? new AssociativeArray();
```
ggetz, line 43: I couldn't find an issue for it, but `AssociativeArray` can probably be replaced with native `Map` at this point. It's easy an efficient to pull the values as an array through built-in methods.

packages/engine/Source/DataSources/FeatureLayer.js
```diff
@@ -0,0 +1,213 @@
...
+ * @param {FeatureProvider | OGCFeaturesProvider} provider
+ * @param {Object} options
+ */
+function FeatureLayer(provider, options) {
+  this.provider = provider;
+
+  this.id = options.id;
+  // this.name = options.name;
+  this._show = options.show ?? true;
+
```
ggetz, line 30: Since this defines a `Rectangle`, we should likely be specifying an ellipsoid at this granularity as well so it's meaningful.

packages/engine/Source/DataSources/parseGeoJson.js
```diff
@@ -0,0 +1,496 @@
...
+//       return that;
+//     });
+//   });
+// }
+
+/**
+ *
+ * @param {GeoJson} geoJson
+ * @param {ParseOptions} [options]
+ * @returns {Entity | Entity[] | undefined}
```
ggetz, line 467: Do you think it would be possible to map this to a list of `Feature` objects? And that, in turn, those could be mapped to Entities in another function to be used `GeoJsonDataSource`? 

That way, the future implementation could map the returned features to primitives.

packages/engine/Source/DataSources/FeatureLayer.js
```diff
@@ -0,0 +1,213 @@
...
+  this.rectangle = options.rectangle ?? new Rectangle();
```
ggetz, line 31: Is this rectangle intended to be the extent of the data, or a user defined restriction?

I think these should be separate properties, and the former should probably be read-only via the public API.

packages/engine/Source/DataSources/FeatureLayer.js
ggetz: Also, and perhaps more importantly, we should be smart about how we store data for lots and lots of features. There's potential for a lot of duplication across the graphical representation (whether than be Entities or Primitives) and a raw array. We likely want to release the raw features once the representations have been created.

packages/engine/Source/DataSources/OgcFeatureProvider.js
jjspace: > To clarify my understanding, the abort controller as it is used in this class is not actively cancelling an in-progress requests— Only preventing more requests from happening. Is that correct?

Correct, as currently written this just stops the loop and future requests from starting and ignores the results. I implemented this first and it basically is just acting as a nicer wrapper for toggling a boolean for previous calls to this function. I added the `cancelResource` function later to try and actually handle canceling the underlying network requests themselves and prevent the browser itself from blocking. There's probably a cleaner way to do this but "it worked" for now.

The [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) and signal is designed to actually be used with the native `fetch` (I picked up the pattern when working on a vscode extension). `fetch` has an [option](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#signal) to pass the `signal` in when it's created and it will just handle canceling the underlying request. This won't work for us because `Resource` is still built around XHR requests which do not have this functionality.

packages/engine/Source/DataSources/FeatureProvider.js
```diff
@@ -0,0 +1,42 @@
...
+  this.credit = undefined;
+
+  this._abortController = undefined;
+}
+
+Object.defineProperties(FeatureProvider.prototype, {});
+
+FeatureProvider.prototype.requestFeatures = function (bbox, time) {
```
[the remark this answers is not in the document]
jjspace: As discussed in person I believe this will be replaced with the full `camera` object so any implementations can calculate a bbox or other features from that.

packages/engine/Source/DataSources/FeatureLayer.js
jjspace: Technically no. However as previously discussed at length TS will not recognize inheritance that's defined purely through JSDoc so it doesn't understand this without both. I have been adding JSDoc comments for types to help me while I develop but may need to do a pass on all of them before this is merged to clean them up to be "correct"

packages/engine/Source/DataSources/OgcFeatureProvider.js
ggetz: Makes sense. I'd still hear any thoughts around the existing `Resource` implementation. Are we at the point where it's enough of a blocker that we should be considering a refactor?

jjspace: I somehow missed that you can `addEventListener` to an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal). I'm pretty sure this lets us augment `Resource`/`Request` to take a `signal` parameter like `fetch` can and have it cancel the requests when the signal is aborted. I've implemented this for a few of the `Resource.fetch**` calls in my latest commits and it works. Take a look and see if this makes sense to you then I can do the rest.

It may even make sense to do a dedicated PR for this update to `Resource` that I can then pull into this branch, open to ideas

packages/engine/Source/DataSources/OgcFeatureProvider.js
```diff
@@ -0,0 +1,307 @@
...
+ * @param {Object} options
+ * @param {Rectangle} [options.bbox]
+ * @param {JulianDate} [options.time]
+ * @param {function} options.partialLoadCallback
+ * @returns {Promise<undefined>}
+ */
+OgcFeatureProvider.prototype.requestFeatures = async function ({
+  bbox,
+  time,
+  partialLoadCallback,
```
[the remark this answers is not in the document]
jjspace: Switched to an event based system instead of callbacks, no longer relevant

jjspace:
@ggetz I just pushed up the logic for some basic "tiling" of an OGC feature provider to try and break up the area we request. It also tracks which tiles have already been requested to avoid excess requests for areas that are still visible or have been completely loaded already. This seems to help reduce the quantity of requests dramatically. I still have some open questions I need to try and address but take a look at the last commit and see if my current approach is making sense so far.

- Still need to figure out how to decide what level to request
- Need to check how this handles services that do _not_ return extents which will basically mean tiling the whole world
- General question? Do we have functions to get a list of "parent" or "child" tiles from a Tiling Scheme?
    - I'd like to be able to check if a parent tile is already `fullyLoaded` or if all children are `fullyLoaded` to try and avoid loading the same data twice.

I also responded to some of the comment threads above so please check those too

packages/engine/Source/DataSources/OgcFeatureProvider.js
ggetz: I also saw that your changes may help us remove the need to use `defer.js`.

That file was a compromise from the when to promise refactor, and we'd like to get rid of it ASAP. I think `Resource`, `RequestScheduler`, and `KmlDataSource` are the last bastions of using `defer.js`.

ggetz:
Since we have paused work on this, I'm going to close the PR, but leave the branch for reference.

Issue #12498, opened 2025-02-26
`getPickRay` returns different coordinate systems with perspective and orthographic frustums


andrewda:
### What happened?

With a perspective frustum, `getPickRay` returns its origin in world coordinates:

```
camera.frustum: PerspectiveFrustum
camera.position: (0, 0, 1000)
camera.positionWC: (-2506152.0309375315, -3848177.6543499078, 4412811.205275626)
ray.origin: (-2506152.0309375315, -3848177.6543499078, 4412811.205275626)
```

However, with the same code using an orthographic frustum, `getPickRay` returns the origin in camera coordinates:

```
camera.frustum: OrthographicFrustum
camera.position: (0, 0, 1000)
camera.positionWC: (-2506152.0309375315, -3848177.6543499078, 4412811.205275626)
ray.origin: (0, 0, 1000)
```

This can cause issues with zooming in 3D orthographic mode, because the `zoom3D` function assumes the ray origin will be in world coordinates to determine the distance:

https://github.com/CesiumGS/cesium/blob/8ca35977c25a028a71cbb4770853b1784076e7e5/packages/engine/Source/Scene/ScreenSpaceCameraController.js#L2360-L2364

I believe world coordinates are desired for both frustums, which is a pretty trivial fix:

```diff
--- a/packages/engine/Source/Scene/Camera.js
+++ b/packages/engine/Source/Scene/Camera.js
@@ -3000,11 +3000,11 @@ function getPickRayOrthographic(camera, windowPosition, result) {
   y *= (frustum.top - frustum.bottom) * 0.5;
 
   const origin = result.origin;
-  Cartesian3.clone(camera.position, origin);
+  Cartesian3.clone(camera.positionWC, origin);
 
-  Cartesian3.multiplyByScalar(camera.right, x, scratchDirection);
+  Cartesian3.multiplyByScalar(camera.rightWC, x, scratchDirection);
   Cartesian3.add(scratchDirection, origin, origin);
-  Cartesian3.multiplyByScalar(camera.up, y, scratchDirection);
+  Cartesian3.multiplyByScalar(camera.upWC, y, scratchDirection);
   Cartesian3.add(scratchDirection, origin, origin);
 
   Cartesian3.clone(camera.directionWC, result.direction);
```

### Reproduction steps

```js
const viewer = new Cesium.Viewer("cesiumContainer");
viewer.scene.globe.showGroundAtmosphere = false;

// Uncomment to test with orthographic frustum:
[...]

Issue #11743, opened 2024-01-05
Camera glitches in orthographic projection mode


javagl:
The "Projection" example sandcastle at https://sandcastle.cesium.com/index.html?src=Projection.html behaves strangely when zooming, both when using the mouse wheel as well as using right mouse drags: 

- When zooming _out_ from the initial view, the camera jumps far away from the model
- When zooming _in_, the camera orientation suddenly glitches out and moves the camera at some random position (slightly below the globe surface)

![Cesium Projection Camera Glitch](https://github.com/CesiumGS/cesium/assets/5597569/62863186-093e-41bd-959a-d1a84f3296a5)

<sub>Edit:</sub>

This only happens in the _orthographic_ view, which is initially enabled as of 
`viewer.projectionPicker.viewModel.switchToOrthographic();`

ggetz:
My guess is that the picking code is returning incorrect values, making the cameraController jump around.

PR #12500, opened 2025-02-28
Return pick ray in world coordinates with orthographic frustum

andrewda:
# Description

Currently, `Camera.getPickRay` returns results in different coordinate systems depending on whether a perspective or orthographic frustum is used. With a perspective frustum, `getPickRay` returns a ray in world coordinates, which is correct per the [`getPickRay` docs](https://cesium.com/learn/ion-sdk/ref-doc/Camera.html?classFilter=camera#getPickRay). With an orthographic frustum (both 2D and 3D), on the other hand, `getPickRay` returns the ray in camera coordinates, which differs from the documentation. This also results in weird zooming behavior in 3D orthographic mode with a tracked entity. 

https://github.com/user-attachments/assets/7973cb9a-c0a8-4a8b-b6c7-ea6e17084fc5

Note that this could be a breaking change if users have depended on `getPickRay` incorrectly returning results in camera coordinates.

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12498
https://github.com/CesiumGS/cesium/issues/11743

## Testing plan

Added four additional tests to cover:

1. get pick ray with lookAt in perspective 3D
2. get pick ray with lookAt in orthographic 3D
3. get pick ray with lookAt in orthographic 2D
4. get pick ray with lookAt in orthographic Columbus view

Only test 1 (perspective 3D) currently passes on `main`. This PR resolves the remaining 3 tests.

# Author checklist


ggetz (approved):

ggetz:
Thanks for the fix here @andrewda! The camera behavior is so much better.

I pushed a small fix to `CHANGES.md` with update instructions in case any users were relying on the previous behavior.

I'm surprised but happy to see that this fix was quite isolated. I tested out camera and picking behavior in both 3D orthographic and 2D scene modes. Everything appears to be working great.

PR #12507, opened 2025-03-03
Replaced defaultValue with ??

### LINKED ISSUE #11674: Replace `defaultValue` with `??`

jfayot:
# Description

- Deprecated `defaultValue` function, replaced by nullish coalescing operator.
- Deprecated `defaultValue.EMPTY_OBJECT`, replaced by `Frozen.EMPTY_OBJECT`.
- Added `Frozen` namespace with `EMPTY_ARRAY` and `EMPTY_OBJECT` frozen values and associated documentation.
- Replaced `[]` by `DefaultValues.EMPTY_ARRAY` only where appropriate.
- Updated coding guide.

## Issue number and link

Fixes #11674

## Testing plan

- All specs still pass.
- Functionally, everything should be the same, so primarily review the files themselves.

# Author checklist


packages/engine/Source/Scene/Vector3DTileContent.js
```diff
@@ -431,37 +430,35 @@ function initialize(content, arrayBuffer, byteOffset) {
...
         "POLYGON_COUNTS",
         ComponentDatatype.UNSIGNED_INT,
         1,
-      ),
+      ) ??
       featureTable.getPropertyArray(
         "POLYGON_COUNT",
         ComponentDatatype.UNSIGNED_INT,
         1,
-      ), // Workaround for old vector tilesets using the non-plural name
```
jjspace, line 444: Looks like this comment was dropped in the conversion process. Same 2 more times lower in this file. Probably want to recover these

packages/widgets/Source/BaseLayerPicker/BaseLayerPickerViewModel.js
```diff
@@ -25,17 +25,13 @@ import createCommand from "../createCommand.js";
...
-    [],
-  );
-  const terrainProviderViewModels = defaultValue(
-    options.terrainProviderViewModels,
-    [],
-  );
+  const imageryProviderViewModels =
+    options.imageryProviderViewModels ?? DefaultValues.EMPTY_ARRAY;
+  const terrainProviderViewModels =
+    options.terrainProviderViewModels ?? DefaultValues.EMPTY_ARRAY;
```
jjspace, line 34: Just want to make sure we follow these through and make sure these arrays aren't modified somewhere. if they are we can't make this change

jjspace (commented):
Thanks for the PR @jfayot! Overall the code changes all seem good to me, just a couple minor comments.

Would you also be able to post a comment with the steps/process you took to do this? I suspect that anyone that was using these functions (like some of our other internal libraries) is using them everywhere and need a large PR like this. It would greatly benefit them if you already found a process that works well

packages/engine/Source/Core/defaultValue.js
```diff
@@ -23,7 +32,17 @@ function defaultValue(a, b) {
  * an object literal.
  * @type {object}
  * @memberof defaultValue
+ * @deprecated This property has been deprecated and will be removed in Cesium 1.134. See <a href="https://github.com/CesiumGS/cesium/issues/11326">Issue 11326</a>.
+ * Use `DefaultValues.EMPTY_OBJECT` instead
  */
```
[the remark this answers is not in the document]
jfayot: That was the point of #12238. This doc header is indeed not generated, but it will not be generated neither if placed just above the get function. I've tested different things, but the only way to have it work properly is to move it to a namespace which is not possible because of defaultValue function.
While the deprecation note is not visible in the generated doc, you will still have the deprecation notice in the console if you try to use defaultValue.EMPTY_OBJECT

Documentation/Contributors/CodingGuide/README.md
```diff
@@ -438,39 +438,30 @@ const p = new Cartesian3(1.0, 2.0, 3.0);
...
 ```javascript
-this._mapProjection = defaultValue(
-  options.mapProjection,
-  new GeographicProjection(),
-);
+this._mapProjection = options.mapProjection ?? new GeographicProjection();
 ```
 
-is better written as
+- If an `options` parameter is optional, use `DefaultValues.EMPTY_OBJECT` or `DefaultValues.EMPTY_ARRAY`, e.g.,
```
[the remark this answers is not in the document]
jfayot: Done!

Specs/DomEventSimulator.js
```diff
@@ -165,53 +165,53 @@ function createWheelEvent(type, options) {
...
 
   const event = document.createEvent("UIEvent");
   event.initUIEvent(type, canBubble, cancelable, view, detail);
 
-  event.touches = defaultValue(options.touches, []);
-  event.targetTouches = defaultValue(options.targetTouches, []);
-  event.changedTouches = defaultValue(options.changedTouches, []);
+  event.touches = options.touches ?? DefaultValues.EMPTY_ARRAY;
+  event.targetTouches = options.targetTouches ?? DefaultValues.EMPTY_ARRAY;
+  event.changedTouches = options.changedTouches ?? DefaultValues.EMPTY_ARRAY;
```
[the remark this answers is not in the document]
jfayot: If I'm not wrong, those are readonly properties from [TouchEvent](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent), so it should be fine

packages/engine/Source/Scene/Vector3DTileContent.js
jfayot: Done!

packages/widgets/Source/BaseLayerPicker/BaseLayerPickerViewModel.js
jfayot: Those are local variables to `function BaseLayerPickerViewModel` that are used to create same name members by slicing: `this.imageryProviderViewModels = imageryProviderViewModels.slice(0);`

`this.imageryProviderViewModels` and `this.terrainProviderViewModels` members can still be modified even though locals were set to frozen [].

jfayot:
@jjspace, I think it is ready for another review.

I've created a [repo](https://github.com/jfayot/replaceCesiumDefaultValue) in which you will find the migration scripts. It has some limitations, but still could be useful and be adapted for specific cases. Of course it can be shared with the community.

jfayot:
Hi @jjspace ! Will you have some time to dedicate on reviewing final changes for this PR ?

packages/widgets/Source/BaseLayerPicker/BaseLayerPickerViewModel.js
jjspace: Yup, I'm following this now, thanks for checking! I think this makes sense and is fine :+1:

scripts/build.js
```diff
@@ -256,6 +256,7 @@ const workspaceSourceFiles = {
     "!packages/engine/Source/*.js",
     "!packages/engine/Source/Workers/**",
     "packages/engine/Source/Workers/createTaskProcessorWorker.js",
+    "!packages/engine/Source/Scene/GltfPipeline/defaultValue.js",
```
jjspace, line 259: My intuition says we probably don't want to do this. I didn't dig into it but I feel like this could mess up some of our builds?
I think a better solution may be to just rename the one for the pipeline to `gltfDefaultValue` so it's distinct?
I know there's more discussion elsewhere about the larger gltf pipeline but I think this can work fine for now. Even if we change it again later it's only in a handful of files so should be easy.

packages/engine/Source/Core/DefaultValues.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Utilities helpful for setting a default value for a parameter.
+ *
+ * @namespace DefaultValues
+ */
+const DefaultValues = {};
```
jjspace, line 6: @ggetz I wanted to get your input on the name of this namespace. We don't normally use plurals for namespaces/classes so it feels a little weird but it does also make sense to me.
I had [originally proposed](https://github.com/CesiumGS/cesium/issues/11674#issuecomment-2691437973) this to avoid a filename conflict during the deprecation process. That said this is used throughout a large number of files so I'd prefer to not have to update it again any time soon.
Looking at it again it might be better to break our rule of "filename should match export" only during the deprecation periof for this file and still set the namespace object itself to `DefaultValue`. Thoughts?

jjspace (commented):
Thanks for the changes and comments @jfayot. This is very close, just some last organizational comments

scripts/build.js
jfayot: Thinking about it, this will automatically induce changes in the import statements for GltfPipeline. As long as you change GltfPipeline files, you might prefer to replace defaultValue with ?? It's up to you!

packages/engine/Source/Core/DefaultValues.js
jjspace: @jfayot I discussed this for a bit with @ggetz and we've decided a better option for this new namespace is `Frozen` (or could be `FrozenValue` if you feel strongly about it). This avoids the filename conflict and doesn't break our rules on plural namespaces or the mismatch between filename and object name. Plus it's more descriptive to what the values actually _are_.

Once this change is in I think we're basically good to go. Please remember to update the changelog entry too.

scripts/build.js
jfayot: Or maybe you just meant to change the function's name from `defaultValue` to `gltfDefaultValue` while keeping the file name as `defaultValue.js` ? Tell me @jjspace

jjspace: Sorry @jfayot this was something else we were discussing today and I didn't have an answer yet.
For now (at least in the short term) we're taking control of these files in the CesiumJS repo so we actually don't have to jump through any of these hoops. Just replace all the `defaultValue` calls with `??` in those files like the rest. This also means the file you duplicated can be deleted too.

jfayot: @jjspace , and about `defaultValue.EMPTY_OBJECT` in GltfPipeline ? Do we replace them with `Frozen.EMPTY_OBJECT` ?

jjspace: @jfayot Yup! make all the changes we need for this PR in those files. They're no longer tied directly to the gltf pipeline files and we can modify them directly

jfayot: Done

packages/engine/Source/Core/DefaultValues.js
jfayot: Done

jjspace (approved):
LGTM, really happy to get this in! Thanks for all the work on this @jfayot!

PR #12525, opened 2025-03-17
GeoJsonDataSource: Allow setter for defaultCrsFunction

malaretv:
# Description

Adds ability to set the defaultCrsFunction used by GeoJsonDataSource.

My specific use-case is trying to load GeoJSON to display on a lunar ellipsoid, so need the ability to specify a custom defaultCrsFunction that calls Cartesian3.fromDegrees() with Ellipsoid.default.

## Testing plan

Tests were added to GeoJsonDataSourceSpec.js

# Author checklist


malaretv:
Failing test seems unrelated to my changes which only change `GeoJsonDataSource`
```
  Scene/Cesium3DTileset
    ✗ verify statistics
	Unhandled promise rejection: undefined thrown
```

ggetz:
Thanks @malaretv! While I do agree that there may be other potential use cases for custom GeoJSON to world coordinate mapping functions, I'm inclined to hold off until we have a concrete case– for the sake of simplicity. 

Always open to further discussion.

malaretv:
Makes sense! We maintain an internal fork, so was just offering it up here in case useful to the wider community.

Just to capture my general thoughts on this as an application developer:
Real-world GeoJSON is by no means predictable or spec-compliant.
The key advantage of being able to provide a custom defaultCrsFunction when loading GeoJSON is that it allows one to handle coordinate transformations based on additional metadata or some other application state, without modifying the original GeoJSON object.

PR #12530, opened 2025-03-19
Support iTwin share keys

jjspace:
# Description

- Add `ITwinPlatform.defaultShareKey` to support using the iTwin Share API for public access
    - This value will _override_ the `ITwinPlatform.defaultAccessToken` if it's set
    - Currently this value must be sent using `Basic` auth instead of `Bearer` so I added a helper function for generating the Auth headers
- Both iTwin sandcastles have been updated to use share keys and no longer need the ion token route

## Issue number and link

No issue

## Testing plan

- Load the local sandcastles and make sure they still work without the ion route
    - [iModel Mesh Export](http://localhost:8080/Apps/Sandcastle/index.html?src=iModel%20Mesh%20Export%20Service.html) ([ci version](https://ci-builds.cesium.com/cesium/itwin-share-key/Apps/Sandcastle/index.html?src=iModel%20Mesh%20Export%20Service.html))
    - [iTwin Feature Service](http://localhost:8080/Apps/Sandcastle/index.html?src=iTwin%20Feature%20Service.html) ([ci version](https://ci-builds.cesium.com/cesium/itwin-share-key/Apps/Sandcastle/index.html?src=iTwin%20Feature%20Service.html))

# Author checklist


packages/engine/Source/Scene/ITwinData.js
```diff
@@ -160,6 +160,7 @@ ITwinData.createDataSourceForRealityDataId = async function (
   realityDataId,
   type,
   rootDocument,
+  // TODO: is there a nice-ish way to thread through custom access tokens per asset?
```
ggetz, line 163: I think we agreed this doesn't make sense for the time being given the series of calls currently needed, right? Let's remove this comment for now.

Perhaps when there is a single API endpoint for various datatypes, we can consider adding this option.

Apps/Sandcastle/gallery/iModel Mesh Export Service.html
```diff
@@ -31,10 +31,13 @@
       window.startup = async function (Cesium) {
         "use strict";
         //Sandcastle_Begin
-        const serviceResponse = await fetch("https://api.cesium.com/itwin/token");
-        const { access_token: token } = await serviceResponse.json();
+        // Generate a share key for access to an itwin without oauth
+        // https://developer.bentley.com/apis/access-control-v2/operations/create-itwin-share/
```
ggetz, line 35: Can we add a link to this in the ref doc as well, like we did for [`GoogleMaps`](https://ci-builds.cesium.com/cesium/itwin-share-key/Build/Documentation/GoogleMaps.html?classFilter=google)?

ggetz (changes requested):
Thanks @jjspace! Just those two notes.

jjspace:
Thanks @ggetz, updated

Apps/Sandcastle/gallery/iModel Mesh Export Service.html
```diff
@@ -31,10 +31,13 @@
...
+        Cesium.ITwinPlatform.defaultShareKey =
+          "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpVHdpbklkIjoiNTM1YTI0YTMtOWIyOS00ZTIzLWJiNWQtOWNlZGI1MjRjNzQzIiwiaWQiOiJmZTliOTgyMS0wYWI5LTQ4ZjItYmMzOC01NjQ5MTg5MDQyOTEiLCJleHAiOjE3NDQ5OTA2MjZ9.7bQIX32CGE4slLDPkOQZ4rRr4BqBSjbUOFAUvcdrFXE";
 
-        Cesium.ITwinPlatform.defaultAccessToken = token;
+        // If you do use oauth for user login set:
+        // Cesium.ITwinPlatform.defaultAccessToken = 'your token'
```
shehzan10, line 40: @jjspace Do we need to have this here? Perhaps an update can point to the tutorial and developer documentation when ready. Same comment for the other sandcastle too.

jjspace: @shehzan10 We can certainly add more docs or tutorials later but I don't think there's an issue with providing this information in multiple locations. These sandcastles may end up being one of the main ways to quick reference how to interact with iTwin data in CesiumJS. I think it's worth keeping these 2 lines just for convenience/reference

shehzan10: @jjspace With the goal being to introduce this to the Cesium community via CesiumJS first, IMO its important to keep it simple. By having the comment `If you do use oauth for user login set`, the statement is vauge-ish because OAuth login has a very specific meaning. So either:
1. We remove this comment and instead point to the tutorials/documentation when ready.
2. Make the comment more accurate, yet simple to understand and allows the developer on which approach to use.

Open to other suggestions as well.

davidhjones: I agree with @jjspace here, I think its nice to show users that they're able to use either form of authentication to use the sandcastle demo -- sharing or oauth. If you'd like to add more detail to suggesting oauth, you could point them do our oauth docs: https://developer.bentley.com/apis/overview/authorization/

shehzan10: Would it be better to then have a comment like:
```
// For alternative forms of authentication you can use, visit https://developer.bentley.com/apis/overview/authorization/.
```

And combine this with line 35?

Apps/Sandcastle/gallery/iModel Mesh Export Service.html
```diff
@@ -31,10 +31,13 @@
       window.startup = async function (Cesium) {
         "use strict";
         //Sandcastle_Begin
-        const serviceResponse = await fetch("https://api.cesium.com/itwin/token");
-        const { access_token: token } = await serviceResponse.json();
+        // Generate a share key for access to an itwin without oauth
```
shehzan10, line 34: @jjspace Lets use correct casing throughout. Eg. iTwin, OAuth.

ggetz (approved):

ggetz:
Thanks @jjspace, as well as @shehzan10 and @davidhjones for your feedback!

Issue #4434, opened 2016-10-14
Let's slowly move to Typescript. (instructions inside)


thinkpadder1:
I set up Typescript pretty easily with Cesium (with minimal friction as well).

1. Install Typescript:
    ```
    npm install --save-dev typescript
    ```
2. Create a file in the root directory **tsconfig.json**:
    ```
    {
        "compilerOptions": {
            "target": "es3",
            "module": "amd"
        },
        "include": [
            "./Source/**/*.ts"
        ]
    }
    ```

3. Rename all js files in the pattern below (except the ones with "!") to .ts (taken from gulpfile.js):
    ```
    ['Source/**/*.js',
    '!Source/*.js',
    '!Source/Workers/**',
    '!Source/ThirdParty/Workers/**',
    'Source/Workers/createTaskProcessorWorker.js'];
    ```
    I just went in the directories (not the "!" ones) and ran the command:
    ```
    find . -name "*.js" -exec rename 's/\.js$/.ts/' '{}' \;
    ```
    
4. Use the Typescript watcher to automatically compile Typescript to Javascript upon changes:
    ```
    tsc -w -p .
    ```

5. The resulting Javascript files will be placed next to the ts files. This will let us keep our current build process and reap the benefits of Typescript with minimal effort. We can just use a .gitignore for the js files. gulp-typescript does exist (see link at bottom).

Remember that Typescript is a superset of Javascript. Therefore, all of our current code works completely fine. Now, we can incrementally migrate. With type-checking, compiler options like strictNullChecks, noImplicitAny, noImplicitReturns, and much more (see link below), a proper module system with import/export, we can create new features faster and find dormant bugs (with type checking). Also, less tests! We don't have many contributors... so Typescript would definitely help. And you can finally use a proper IDE with autocomplete and refactoring tools.

We can complicate the build process with gulp-typescript and more, but I want us to start with the bare minimum first so we don't scare anyone away. **Minimal friction** is the keyword here.

- [Typescript with Gulp](https://www.typescriptlang.org/docs/handbook/gulp.html)
- [Migrating from JavaScript](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html)
- [Incrementally Migrating JavaScript AMD Project to TypeScript](https://dzone.com/articles/incrementally-migrating-javascript-amd-project-to-1)
- [Compiler Options](https://www.typescriptlang.org/docs/handbook/compiler-options.html)

[...]

PR #12581, opened 2025-04-23
added @cesium/utils

jfayot:
# Description

This PR is an experimental approach to tackle different subjects:
- Continue further splitting of @cesium/engine into smaller packages
- Start Typescript migration into a iterative way
- Use Vite as a build tool to manage Browser / Node outputs

For the moment, only a few files have been move to the new package @cesium/utils:
- defer.js
- defined.js / d.ts
- Check.js / d.ts
- DeveloperError.js
- RuntimeError.js
- Fullscreen .js
- FeatureDetection.js

Corresponding Specs have been moved too, but unit test FeatureDetection / "detects WebGL2 support" had to be removed as it pulled unwanted dependency Scene.

There have been some imports reordering during the process, which makes quite a lot of modified files. This is because I temporarily used organize-imports-cli. This could be added as a lint-stage.

For the documentation generation, I'm still using jsdoc comment headers, but as it is with typescript files, I had to use the jsdoc-babel plugin. It works quite well, and allows to keep consistent doc across js and ts. But it could be interesting to evaluate other tools such as typedoc which are specifically made for typescript.

For the declaration types, I'm using the one created by tsc, and then inserting it at the end of Cesium.d.ts (see function createTypeScriptDefinitions).

If this approach is something you'd like to continue with, let me know, so that we can agree on what to put inside @cesium/utils

## Issue number and link
Addresses following issues:
- #10636
- #8359
- #4434

## Testing plan

- All tests should still pass
- ESLint should be OK
- Documentation generation OK with new @cesium/utils section
- make-zip should be OK
- build-ts should be OK
- build-release should be OK

# Author checklist


ggetz:
Hi @jfayot,  I'm happy to see the interest in this! We've been eyeing doing something similar for math and other non-rendering 3D and geospatial classes.

A few thoughts up front: 

- We currently use [`esbuild`](https://esbuild.github.io/) for our bundling and other build processes, which is [one of the bundlers vite depends on](https://vite.dev/guide/why#why-not-bundle-with-esbuild). I think we would likely stick with that approach over pulling in vite as a bundler. Though, it would be in our best interest to clean up our existing build scripts in anticipation ad adding additional entry points, like new CesiumJS packages and [Sandcastle](https://github.com/CesiumGS/cesium/pull/12574).
- The `utils` package architecture as you describe it here makes total sense to me, and I'm glad to see you removed the `Scene` dependency from `FeatureDetection` to support it. Are there any other files which also should be considered? Maybe `clone`, `Event` and `EventHelper`, `Resource` and `buildModuleUrl`, etc?
- Some of these functions should probably be deprecated over being moved. While I don't want to blow up the scope of a PR, deprecating first be better to reduce noise.
  - `defer.js` was [added as a workaround](https://github.com/CesiumGS/cesium/pull/8525), and is an anti-pattern which we'd like to discourage in favor of using `Promise` directly.
  - `FeatureDetection` has several functions which check for extensions or other WebGL features which have been widely supported for years, and that we assume support for in most areas of the codebase. For example `supportsTypedArrays`. We also explicitly do not support IE anymore, so we shouldn't need to check for it.
- I have not yet looked at your approach to generating documentation, but am interested to see your solution and take a look at typedoc. We are interested in [moving away from tsc-jsdoc](https://github.com/CesiumGS/cesium/issues/10455).

jfayot:
Thanks @ggetz for your review.
> We currently use [esbuild](https://esbuild.github.io/) for our bundling and other build processes

Sure I can try to stick back to esbuild, but I guess that the new Sandcastle will also be using vite as build system, so it's not really like we'll add it specifically for @ceium/utils. And here, vite is only used for new packages, @cesium/engine and @cesium/widget still rely on esbuild. This would be the opportunity to progressively modernize cesium's build system. Let me know if you still prefer to have @cesium/utils use esbuild.

> Are there any other files which also should be considered?

I was thinking of the following files to move to @cesium/utils:
- AssociativeArray
- DoublyLinkedList
- Event
- EventHelper
- Frozen
- Fullscreen
- GregorianDate
- Heap
- Interval
- Iso8601
- JulianDate
- KeyboardEventModifier
- LeapSecond
- ManagedArray
- Proxy
- Queue
- Request
- RequestErrorEvent
- RequestScheduler
- RequestState
- RequestType
- TimeConstants
- TimeInterval
- TimeIntervalCollection
- TimeStandard
- appendForwardSlash
- arrayRemoveDuplicates
- binarySearch
- clone
- combine
- createGuid
- deprecationWarning
- destroyObject
- formatError
- getBaseUri
- getAbsoluteUri
- getExtensionFromUri
- getFilenameFromUri
- getImageFromTypedArray
- getImagePixels
- getJsonFromTypedArray
- getMagic
- getStringFromTypedArray
- getTimestamp
- isBitSet
- isBlobUriuri
- isCrossOriginUrl
- isDataUri
- isLeapYear
- loadAndExecuteScript
- mergeSort
- objectToQuery
- oneTimeWarning
- parseResponseHeaders
- queryToObject
- srgbToLinear
- subdivideArray
- wrapFunction

I've excluded everything that uses Math or that has an indirect dependency on Math, as it would create circular dependency. Resource and buildModuleUrl are in this case.

> Some of these functions should probably be deprecated over being moved

Do you mean that those deprecations should be done prior to this PR?

ggetz:
> Sure I can try to stick back to esbuild, but I guess that the new Sandcastle will also be using vite as build system, so it's not really like we'll add it specifically for @ceium/utils. 

I think we can circle to back to that discussion after we sort out the structure of the proposed package. I don't think it's a matter of esbuild or vite being more modern, but what are the granularity of options needed. 🙂 

> Do you mean that those deprecations should be done prior to this PR?

I think that would be helpful over including the deprecations in this PR. That way they can be reviewed separately and don't need to depend on the rest of the code reorg here, which will likely have more discussion.

I'll follow up with a few more thoughts on what belong in each package, but I agree there should likely be a distinction between what goes in this proposed `utils` package and the Math-related modules. :+1:

javagl:
I have not yet looked at _all_ changes and the details of _all_ things that are proposed to be pulled out. One quick, drive-by comment is that

- Iso8601
- JulianDate
- LeapSecond
- TimeConstants
- TimeInterval
- TimeIntervalCollection
- TimeStandard
- getTimestamp (?)
- isLeapYear

_look_ like they could be a candidate for an own package, or, depending on whether there are considerations to have some _additional_ structure _within_ the packages, some `Temporal` subdirectory. 

Also, such a `http-helpers` package sounds reasonable (although it's not clear which of the listed classes/functions would go into that _exactly_, there are many that _sound_ like they could belong there).

jfayot:
Sorry for the long period of inactivity, but I guess you must be quite busy too with the upcoming Cesium developer conference :grin: !

I'll try to catch up and continue the work on this PR next week...

Issue #12310, opened 2024-11-15
New 1.123 Environment Map Rotates Together With Model


xtassin:
### What happened?

Testing the new 1.123 Environment Map implementation, I noticed that the env map is rotated along with the 3D model.

### Reproduction steps

1. Load PBR based GLTF2 model 
2. Rotate model
3. Environment map (reflection) rotate together with the model (up side down at roll=3.14)
...

### Sandcastle example

[...]

PR #12588, opened 2025-04-28
Add back ModelInstance class

lukemckinstry:
# Description

 - Adds back a ModelInstance Class for runtime instancing.
 - Refactor of ModelSceneGraph to resolve issues in how transforms were applied to models with multiple hierarchical nodes.

### ModelInstance and ModelInstanceCollection

Supply transformation matrices (Matrix4) to the Model for each instance by
 - calling `model.instance.add(transform)`
 - supplying `options.instances` to `Model.loadGltfAsync()`

In this way, the transformation matrix for each instance can be computed from a cartographic coordinate as shown below.

Example:
```

    url: "../../SampleData/models/GroundVehicle/GroundVehicle.glb",
  });
const lng = -75.1652;
const lat = 39.9526;
const height = 30.0;

let position;
let transform;
function genInstances(count) {
  for (let i = 0; i < count; i++) {
    position = Cesium.Cartesian3.fromDegrees(
      lng + i * Math.random() * 0.0001,
      lat + i * 0.0001,
      height + i,
    );
    transform = new Cesium.Transforms.eastNorthUpToFixedFrame(
      position,
    );
    model.instances.add(transform);
  }
}
viewer.scene.camera.setView({
  destination: Cesium.Cartesian3.fromDegrees(lng, lat, height),
  orientation: new Cesium.HeadingPitchRoll(
    2.205737333179613,
    -0.7255022564055849,
    6.283181225638178,
  ),
});
```
### Shader logic

Cesium already supported instancing from glTF (specifies instancing in model space) and i3DM (specifies instancing in world space). Because this feature wanted to set up instancing in world space, we followed i3DM logic, including for the shader.

i3DM
```
positionMC = instance * node * position

node includes axis correction ie.
node = (axis correction * node) 

view * model * Scene Graph top level * instance * node * position
``` 

Runtime instancing
```
Divide instance into instance_center and instance_relative_to_center_transform

instance_relative = instance_relative_to_center_transform * Scene graph top level * axis correction * node

The instance center is calculated relative to eye:

instance = instance_center +translate+ instance_relative

Then modelView is set to czm_modelViewRelativeToEye (transforms model coordinates, relative to the eye, to eye coordinates)

czm_modelViewRelativeToEye * instance

```

### ToDo:
 - [x] broken tests (mostly classes impacted by the scene graph transforms refactor)
   - [x] ModelSceneGraph
   - [x] ModelDrawCommands
 - [x] Add tests 
   - [x] ModelInstance Class
   - [x] RuntimeModelInstancingPipelineStage
   - [x] ModelInstancesUpdateStage
 - [ ] document edge case restrictions, have not yet restricted these in code
   - [ ]~When instances are supplied to Model, modelMatrix must be the identify matrix.~ (this is supported)
   - [x] Cannot supply instances to a model with glTF with `ext_mesh_gpu_instance extension`, throw developer error
   - [ ] 2D and columbus view - https://github.com/CesiumGS/cesium/pull/12816
 - [x] Clean up sandcastle example
 - [ ] demo picking in sandcastle
 - [ ] document important technical details on issue - including scene graph and bounding sphere calculation refactors

### Performance notes
 - Users should optimize the models they are instancing with a tool like gltf-transform optimize`gltf-transform optimize wind_turbine_small.glb wind_turbine_optimize.glb --texture-compress webp`or glTF Report
 - Applying model.scale has moderate performance cost, scale models offline beforehand if this impacts performance
   - not difficult to optimize
 - Apply model.minimumPixelSize has major performance cost, only use if needed
   - possible to optimize, ideally if we could track when the camera has moved

## Issue number and link

#10846

## Testing plan
 - General
   - Test turbine sandcastle example Apps/Sandcastle/gallery/3D Model Instancing Terrain.html
 - Instances transformed, added, or removed
   - Test development sandcastle example Apps/Sandcastle/gallery/3D Model Instancing.html
     - In setTimeout loop instance transforms will be changed
 - API options
   - add instances in Model.fromGltfAsync 
     - Use link A, point model url to non-instanced glTF eg. GroundVehicle
   - add instances via ModelInstanceCollection.add
     - Use link B, point model url to non-instanced glTF eg. GroundVehicle
 - Earth centered vs alternate center
   - Earth centered - turbine example
   - non Earth centered - Link A
 - Model scaling 
   - Test model.minimumPixelSize - use 3D Model Instancing Terrain sandcastle
     - instances should scale accordingly, same behavior as for single model
   - Test model.scale
     - instances should scale accordingly
 - Performance
   - 50,000 US wind turbines [link](http://localhost:8080/Apps/Sandcastle/index.html#c=hVbrbts2FH4VwhhQCXWoi2+x4wTb0qYbsK3Bkm4/4mChJdriQpEGSdnJ2rzTnmFPtkNSsuXY65AgkT5+5zsX8hwqk0IbtGZ0QxU6R4Ju0CXVrCrxbw4LZp3MvV9KYQgTVM06XfR5JhAyVClAJo3BrX/HCyXL36XieQ0EYXcmXsKzmZgJTg0qZU45vBn17IUcAM7JhjDTqP1sQaf1gZvFd/pZZIFjI1QpPkGzDsYR/N6QcsXpO2JI5HR0tGEi/8NUag7R4iWfQ8DeLopKJlhZldfsifIb9hedoF7qFl14qK4D1hkF05ViJTNsTTUmeR44dUt7QRkxWYECSE+q0OeQQR0lp5jLZfBwRRinOTIScUlynx9G33x2Bi8PTsRWw9eDwQ4QkdFbRYReSFWeeXwlNbiXAl4zt0tLps0nZSs16xTGrPQkiiyGl8wU1bzSVAHRUGFwJsuIV4+0zB6tvHqOxuP+fLgY03S0SFOSk6w3z4ZxOs5oPjidJ4tIkU1E42yRkNP5aDjKk9Ew6fWzJBumw/6AzOPRaEB7vYT0R2lU6ZO6whpnej3rQJALClUJ6iBDWxRsCioCRfUK4qfo/AI1z9jQJxOELRao3AJmSfU2+6S59QEp1+vYwK4EIdYrzkzwZjYTb/zWNfyCkpwqa+Es7+L7htvdMqHIKHClB1pyBv+mno05FUtTAPL2bdjE0SivCa/oTpgdEW6oSm6A93H+J82MO8Pv4bQzqoM6OlySVRAUXcTyp9CmfAfPXv8OoPv78LUiJ1AatCJK0ys4VCYAFxjAA55YHuGJ5WteQdmyOCJZ41++oHhn0pxEoNfdeUmUgSciei69d3SpKKQHjro21G6tv5M4OOT7w2YLa0yJNr9IZYpPq1t5BZ2aXylS0qAJYqfp+6pR9l164Kehv2zb3P/cEJFn4Ap6FuxupeRzor6vjJECZt4HOByX4FUR9KNYSDvzFpXIXBGCVssbmAWOdb4/Ozx6tqO1SujXcKu9d1rKyKUiq4Jl+7VuUFftbfFf1aR9UpopSkyBjWz2p+3Anh5mqpyGQHB1DoZ7Ou39/qqOFMuvCW2P2p6VR7cWqd+Zdhczd5TratVAu6LMDuH/jtAX2ZLCfS+7JuX8/+wtJzwe5Hba1+fkutnhhy463Jct+wefyASugzqnl3/+7qJrG6gFXcQO+hV8W8TGAIC/NnwER84a1tTYK9tfkjnVhgliY5i0+2zXt4Fvi/EgHo8SPB4MBnGSJumgvitP+oNx2h8kOBmN417SH/WblX4vHveHp2Bymo6TeNhzeOj+Shhy8Jlw4LZO2iVp09obGqcxHiZddDLCaRfFOG59LnS6nak2z5xeND3/LStXMBzsN0AAt7+hcPsTSCmaV9kjhYtP66bjp1HbdJqzNYzb8yPfMyjjRGtYWVTcfRbMOhfTCPgHpvY6hzw+rqni5NnSiuTiJw9ijKcRvB63NH7AvFL+Fw)
 - glTF formats
   - glTF w/ single node - CesiumBalloon
   - glTF w/ hierarchical node tree - GroundVehicle, CesiumMilkTruck
   - glTF w/ top level transform 
   - glTF with Ext_mesh_gpu_instancing - should throw runtime error
     -  `instances` supplied to Model.fromGltfAsync [link A](http://localhost:8080/Apps/Sandcastle/index.html#c=hVRrT9swFP0rVj+lozjvpC/QRhls0sbQeHxZp8kkbmvNsSvbKTDEf9+N3fQBTGsqx/fa59xzr6/j+1dElAXRhtNfJ3TOxFQUUmiDVozeU4WOkKD3aEI1qyt8a33etFNYeyKFIUxQNe10R1NAcmrQUmpmmBQjZxpFhJ5JVY1aYgYDEcAA3D9+ghshWEdes5uBLxjBa4zC5n1w0EVPDTOCX8u8r2lClIEZEbEX4KAH2HeAbWYwOFkN2PfRRFFiKCKIy4Jwp4wTy1gRo9gDSEMfL24cYCMcwq1DfbW7EjxTsrreor1WWHfkoJsM8bLWC2/D5NafG0ltMbiYA/9hnuIwS6NNjYAY3PEAD9IosznsJP8qcSvolM4VpdpzEoC3BySNsT2aSpaUuyT2a3jdKtSYQitcSGUWN8trecYeaHmmSEX3ctxhsxnZGTCSe8LMpliN0yo752b2QT+Kwntqz6JWfIimHYx9+F+RasnpKTHEt0zaP5EPn9c1LPcMPAeuaafneP7Dcq5kLcpbumAFp/sW8NxtaComWFVXl5Asv2J/6BBlSe/FSQ630xa1LeZw17DLz+6o3R3CuqCC4qViFZRwBW1BytKzEFdN9+xtLqDmimBNTXPnXOFKqg0TtuWG/7gBTloYpXEaRjiN4ijJwizrrzUfJnkcDaIIR2GSZnkeDdYLSZAnYZjiJE/SJAsGDtC1o1SMwj1/FfYTJSUT80tmisV3yfk6OJAHaR7ncRyH+SAL4zZ2gPMoTYMoSrMkSNN+0gbPcNSPw37YrMT9MN8Gt2X0dz9RH0XZ6XXG2jxyeuzgCL1n1RJatukHDzrBUOgEuOnav6uL39TgQuv2ao79Xei4ZCvEyqM3Pmio4ERrWJnV3HbFtHM89mH/KyiXtg7fVlRx8thsW4THX5wTYzz2wXwbaaTkd0S9YP4L)
     - ModelInstanceCollection.add called after model created [link B](http://localhost:8080/Apps/Sandcastle/index.html#c=hVRtb9MwEP4rVsWHlHXOe9J23QQbGyDBQGzwqRJ4ya2zcOzKdruNaf+di930hYKoqiTn8/Pc+e45h+EVk3XFjBXw/RRmXE5lpaSxZMnhHjQ5JhLuyRkYvmjoN7cWTHuVs8+UtIxL0NNe/2i6Rgo5Q9hhmdO4yJOj9TKzuJyO6ChPCrddgCVzZbjlSqJrFeSMaYtfTKb0VqvmDcw0gAmmkuAPuQdI1Br9NUejahAfmdX8YTffa82kuVW6MRTwiJdK27uv82t1wR+gvtCsgaCL/wcbWlY/kqc2kFtAYnbPuO2oP7aLLsG3wt6+No+yCp58jgstxmTaozTE/xVr5gLeMMtCx2PCU/XwHuvBZAX1jkFnyDTtDTxLGP6H561WC1l/gzteCdi1kOlmTdRwyZtF8xnPLK74LxiTIutcm8KNtw3nfm5LQlY6oKYCCXSueYPlWoKhrK4DB2m3PZOK2eqOBKC10n1ft7btCpMRahb8uGBcQE2sIkKx2gej5MWTAzz/cCRdB/iqIuv2odOAveYNqIUNgj45PvEh0EkCB8H+REf4mpC4fR8cuCT8Mbc0tiWOjc6CiEYDxL5EbPuFDy8H3wdypoFZIAxTr5ggtk0LRdgyNl51XJLzy68e4NW+d4aNwH2NMyee6w3XjhTX3aEdka/4Hq3fjKUjzwMUSxFFUTeNO52rUOyaUSxjO8ReqjUYy6ULPv5HZXwmcZKneZzQPEmTrIiLYrgS0GFWpskoSWgSZ3lRlslo5ciiMovjnGZllmdFNPKAvnsqzQEvjr2w74DVXM4+c1TSFyXEKjiSR3mZlmmaxuWoiNMudkTLJM+jJMmLLMrzYdYFL2gyTONh3HrSYVxugjtNh9t33rmse4PexNhHASceTsgr3szxrmgnOcDZs4Czhwow4c2i+gmWVsZ0TZqE29BJzZeE18d/uSFJJZgx6LldCDeH097JJMT9e9B2QLAOn5agBXtst93FJx/8IqV0EqL5d6RVStww/Qfzbw)
- Picking - see https://github.com/CesiumGS/cesium/pull/12744
- Map views
   - 2D
   - Columbus View
 - Review regression other sandcastles
   - 3D Models 

# Author checklist


packages/engine/Source/Scene/Model/RuntimeModelInstancingPipelineStage.js
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * @namespace RuntimeModelInstancingPipelineStage
+ * @private
+ */
+const RuntimeModelInstancingPipelineStage = {
+  name: "RuntimeModelInstancingPipelineStage", // Helps with debugging
+};
+
+/**
+ * Process a node. This modifies the following parts of the render resources:
```
ggetz, line 27: This was copied over from the previous instancing stage, and will need to be updated to reflect the actual workflow.

packages/engine/Source/Scene/Model/ModelSceneGraph.js
```diff
@@ -471,13 +546,14 @@ ModelSceneGraph.prototype.buildDrawCommands = function (frameState) {
     modelPipelineStage.process(modelRenderResources, model, frameState);
   }
 
-  const modelPositionMin = Cartesian3.fromElements(
+  // Positions are in local glTF scene coordinates
```
lukemckinstry, line 549: Is this a terminology change, is there a difference between model coordinates and local glTF scene coordinates?

lukemckinstry:
> Since the workflow was touched by some refactoring, we should confirm the following are still working for non-instances models:
> 
> * Clamp to ground: Viz, zooming to, picking
> * 2D Mode: Viz, zooming to, `projectTo2D`, picking
> * Animations

All these are resolved.

> We also need to check scaling with `minimumPixelSize` and `maximumScale` for both regular models and model instances.

Scaling is fixed for regular models, but not yet implemented for runtime instanced models.

lukemckinstry:
@ggetz 

1. Confirming two restrictions in the runtime model instancing feature
 -  When instances are supplied to Model, modelMatrix must be the identify matrix.
 - Cannot supply instances to a model with glTF with `ext_mesh_gpu_instance` extension, throw developer error

I did not add these to the code yet

2. Also confirming a detail related to making `model.instances` depend on a collection class so we have build in add and remove functions. Right now, `model.instances` is just a pass through to ModelSceneGraph._modelInstances, so is my understanding correct that the new collection should be in `ModelSceneGraph._modelInstances` and not `model.instances`?

ggetz:
> Also confirming a detail related to making model.instances depend on a collection class so we have build in add and remove functions. Right now, model.instances is just a pass through to ModelSceneGraph._modelInstances, so is my understanding correct that the new collection should be in ModelSceneGraph._modelInstances and not model.instances?

I think that would be fine. `ModelSceneGraph` is an implementation detail that should not be accessible by the user. The import things are that (a) the `.add` function and similar functionality of the collection is accessible from the public `Model` interface and (b) the needed information is passed through to `ModelSceneGraph`.

Apps/Sandcastle/gallery/3D Model Instancing.html
```diff
@@ -0,0 +1,206 @@
...
+      @import url(../templates/bucket.css);
+    </style>
+    <div id="cesiumContainer" class="fullSize"></div>
+    <div id="loadingOverlay"><h1>Loading...</h1></div>
+    <div id="toolbar"></div>
+    <script id="cesium_sandcastle_script">
+      window.startup = async function (Cesium) {
+        "use strict";
+        //Sandcastle_Begin
+        const viewer = new Cesium.Viewer("cesiumContainer");
```
ggetz, line 32: @lukemckinstry Once the public API is where you'd like it, a reminder to clean up the Sandcastle example.

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
```diff
@@ -0,0 +1,289 @@
...
+    beforeAll(function () {
+      scene = createScene();
+      scene.renderForSpecs();
+
+      scene2D = createScene();
+      scene2D.renderForSpecs();
+    });
+
+    afterAll(function () {
+      scene = createScene();
```
ggetz, line 43: Should this be `scene.destroyForSpecs()`?

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
```diff
@@ -0,0 +1,289 @@
...
+          expect(transformsTypedArray[i]).toEqualEpsilon(
+            expectedTransformsTypedArray[i],
+            CesiumMath.EPSILON10,
+          );
+        }
+      });
+    });
+
+    it("model instances update stage updates transform vertex attributes", function () {
+      return loadGltf(sampleGltfUrl).then(function (gltfLoader) {
```
ggetz, line 200: Let's prefer `async/await` syntax to promise chains. It flattens the test code and helps makes things more readable.

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
```diff
@@ -0,0 +1,289 @@
...
+        attributeIndex: 1,
+        attributes: [],
+        instancingTranslationMax: undefined,
+        instancingTranslationMin: undefined,
+        shaderBuilder: new ShaderBuilder(),
+        model: {
+          _modelResources: [],
+          _pipelineResources: [],
+          statistics: new ModelStatistics(),
+          sceneGraph: {
```
ggetz, line 95: Let's have this function only set up the mocks unrelated to the functionality we're testing. For instance, would probably expected to see something like the following line in the actual test block. 

````js
renderResources.sceneGraph.modelInstances = [sampleInstance1, sampleInstance2];
```` 

1. It helps make the test more readable for those who are working with them later. 
2. It allows the mock and the test code to change independently.

File: packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js [hunk 7]
```diff
@@ -0,0 +1,289 @@
...
+        const expectedTransformsTypedArray = new Float32Array([
+          -0.410076379776001, 0.7071067690849304, 0.576053261756897, 0,
+          -0.410076379776001, -0.7071067690849304, 0.576053261756897, 0,
+          0.8146623373031616, 0, 0.5799355506896973, 0, 0, 0, 0, 10, 10, 10,
+          -0.410076379776001, 0.7071067690849304, 0.576053261756897, 0,
+          -0.410076379776001, -0.7071067690849304, 0.576053261756897, 0,
+          0.8146623373031616, 0, 0.5799355506896973, 0, 0, 0, 0, 20, 20, 20,
+        ]);
+
+        const expectedTransformsBuffer = Buffer.createVertexBuffer({
```
ggetz, line 226: It looks like this test is jumping through a few hoops to validate the buffer contents. 

Instead of creating a new buffer, perhaps consider validating the output of `RuntimeModelInstancingPipelineStage._getTransformsTypedArray` against `expectedTransformsTypedArray`, and then checking `runtimeNode.instancingTransformsBuffer`'s `usage` and `byteLength` properties directly.

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
ggetz, line 95: Are we able to use an instance of `ModelSceneGraph` here rather than just a mock?

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
```diff
@@ -0,0 +1,289 @@
...
+      "west",
+    );
+    const instanceModelMatrix1 = new Transforms.headingPitchRollToFixedFrame(
+      samplePosition1,
+      headingPositionRoll,
+      Ellipsoid.WGS84,
+      fixedFrameTransform,
+    );
+
+    const instanceModelMatrix2 = new Transforms.headingPitchRollToFixedFrame(
```
ggetz, line 74: Do either of these test scale transforms?

packages/engine/Specs/Scene/Model/ModelInstanceSpec.js
```diff
@@ -0,0 +1,417 @@
...
+        },
+        scene,
+      );
+
+      const boundingSphere = instance.getBoundingSphere(model);
+      expect(
+        ModelInstance.prototype.getPrimitiveBoundingSphere,
+      ).toHaveBeenCalled();
+
+      console.log("boundingSphere --> ", boundingSphere);
```
ggetz, line 215: Make sure to remove all `console.log` calls.

packages/engine/Specs/Scene/Model/ModelInstanceSpec.js
```diff
@@ -0,0 +1,417 @@
...
+          instances: [instance],
+          modelMatrix: Matrix4.IDENTITY,
+        },
+        scene,
+      );
+
+      const boundingSphere = instance.getBoundingSphere(model);
+      expect(
+        ModelInstance.prototype.getPrimitiveBoundingSphere,
+      ).toHaveBeenCalled();
```
ggetz, line 161: So checking whether a specific function is called is always going to be more fragile than confirm the output is what we'd expect. In this case, instead of validating that `getPrimitiveBoundingSphere` is called, maybe a less fragile test would be to confirm the camera has zoomed to the expected position and orientation?

packages/engine/Specs/Scene/Model/ModelInstanceSpec.js
```diff
@@ -0,0 +1,417 @@
...
+      };
+
+      const primitiveBoundingSphere = instance.getPrimitiveBoundingSphere(
+        sampleModel.modelMatrix,
+        sampleModel._sceneGraph,
+        sampleRuntimeNode,
+        samplePrimitiveBoundingSphere,
+      );
+
+      expect(ModelInstance.prototype.computeModelMatrix).toHaveBeenCalled();
```
ggetz, line 308: Do we need this expect case? I don't think the bounding sphere values will be correct without it, so we can assume it was called correctly.

ggetz (commented):
@lukemckinstry Thanks for resolving the last few features! 

I looked over the unit tests so far and left a few comments. Let me know if there are any concerns around those.

lukemckinstry:
Another limitation that I should have included in the [above comment](  https://github.com/CesiumGS/cesium/pull/12588#issuecomment-2855200164) but was not top of mind was that the current runtime model instancing implementation does not support changing the size of the instances buffer ie. number of instances arbitrarily over time. We support changing the instances themselves, but not the number available. 

I believe we talked about this being OK for initial release earlier in development , but implementing the collection (`ModelInstanceCollection` with `add` and `remove` functions) this afternoon made this limitation a little more apparent. I do see how we can provide clear instructions so developers understand (initial) limitation.

ggetz:
> I believe we talked about this being OK for initial release earlier in development , but implementing the collection (ModelInstanceCollection with add and remove functions) this afternoon made this limitation a little more apparent. I do see how we can provide clear instructions so developers understand (initial) limitation.

If this is the case, I believe we should re-run the model pipeline with `model.resetDrawCommands();`. For an example, see `updateClippingPlanes` in `Model.js`. `resetDrawCommands` is called when the collection changes state.

ggetz:
> That works. But it raises some question about API design

@lukemckinstry and I discussed this offline— 

We'll still need `ModelInstanceUpdateStage` to avoid rerunning the entire pipeline when one instance is updated. We'll re-run the pipeline when an instance is added or removed, as we'll need to reallocate the buffers containing the instance data. 
 
Updates can be managed on the `ModelInstance` instances themselves, setting dirty flags as needed. The update stage should be able to handle those accordingly.

javagl:
> When instances are supplied to Model, modelMatrix must be the identify matrix.

Is it possible to summarize why this constraint exists? 
(Beyond: "Well, all these matrices... it's complicated...")

One could argue that it's possible to "emulate" any `modelMatrix`, by baking this `modelMatrix` into the `instances` matrices. But that's also why that constraint could be hard to justify.

ggetz:
> Is it possible to summarize why this constraint exists?

@javagl I think mostly for API simplicity. 

We're accounting for double precision for any instance locations on the globe. So a localized `modelMatrix` with instances relative to that shouldn't be needed for precision. Do you have another use cases in mind for providing a non-identity model matrix?

javagl:
The question was unrelated to precision. (Precision is a tricky issue here, but ... unrelated _for now_). 

I rather thought about cases where people want to create instances in a known, **local** coordinate space. Think about an airport runway where 200 lights are left and right of the runway, along a straight line, 10 meters apart. And then, users want to put these instanced models at a certain position on the globe, and use the `modelMatrix` for that.

In terms of _convenience_, the "best" API certainly depends on the use case. For example, [in this screenshot](https://github.com/CesiumGS/3d-tiles-tools/issues/84#issuecomment-2525245632), I used lat/lon/height as the input. In other cases, people might only have the _local_ transforms. There may also be cases where the instancing information is given as TRS properties. And people will have to write quite a bit of boilerplate code to 1. assemble these into Matrix4 objects, put them into the array, and maybe squeeze in the ENU-to-FF-matrix for the desired geolocation here. 

(API design is often sort of a trade-off ... about shifting the responsibility for doing things (correctly!) between the implementor and the user. I think that the API should be easy to use _correctly_, and hard to use _incorrectly_. A seemingly(!) very specific aspect here is: People _will_ create instances. They _will_ set the `modelMatrix`. They _will_ open an issue because the "model matrix does not work"...)

javagl:
I've seen that there are a bunch of conflicts. Most of them are in `ModelSceneGraph`. And they are "my fault", so to speak, caused by a [tiny, tiny part of the draping](https://github.com/CesiumGS/cesium/commit/9defb467af8c6084b0821a2794e2137873a475b8#diff-9352fa83c02dfe738c46bd358863b1fbf51c07fb042bc5fb694e053972045c19). The `buildDrawCommands` function does many different things (and it is doing some of them _wrong_ - namely, the computation of the bounding volumes). So I did split this up into...
```javascript
ModelSceneGraph.prototype.buildDrawCommands = function (frameState) {
  const modelRenderResources = this.buildRenderResources(frameState);
  this.computeBoundingVolumes(modelRenderResources);
  this.createDrawCommands(modelRenderResources, frameState);
};
```
with `computeBoundingVolumes` being explicit and saying
`NOTE: This contains some bugs. See https://github.com/CesiumGS/cesium/issues/12108`

This was only an attempt to follow the "boy scout rule". It was **not** a change that affected the _funcitonality_. It was supposed to help _isolating_ and _fixing_ that issue (and maybe even to cover it with specs). 

Now... the current state of the `ModelSceneGraph` _in this branch_ has been changed significantly. (Some of the changes may be caused by attempts to work around that bug, but who knows). The current state contains 13 (!) `TODO`s. I'm not sure how to handle that. (I'd prefer to _keep_ the split-up function and to not introduce additional flags and properties and bounding volumes, but ... whatever has to be done to get instancing working...). So in doubt, one way of resolving this would be to simply overwrite the _current_ (`main`) state of the class with the state from _this_ branch.

This is done here: https://github.com/CesiumGS/cesium/tree/model-instance-v2-tests 

Maybe I can allocate time for another dedicated cleanup pass for the `ModelSceneGraph` which replicates the cleanup from the commit linked above. We _have_ to fix that bounding volume issue...

javagl:
From a quick test, it _looks_ like the current state also works for a model with a non-identity `modelMatrix`? (I haven't checked the code, only ran a small sandcastle). One open point might be whether `model.boundingSphere` should take into account the instances. (Currently, it does not seem to do this, but might be useful for `flyToBoundingSphere(model.boundingSphere)`). Maybe that's already on the radar.

lukemckinstry:
> the current state of the ModelSceneGraph in this branch has been changed significantly. (Some of the changes may 

be caused by attempts to work around that bug, but who knows)
These changes were part of the refactor Gabby did to solve the problems we encountered  with glTF with multiple hierarchical nodes or glTF with a top-level transform. The transforms in the tree were not being applied correctly, leading to child nodes not aligning to their parent. We need to document this better. 

> ToDos

These are leftover from the refactor, we will resolve these.

> one way of resolving this would be to simply overwrite the current (main) state of the class with the state from this branch.

This looks good. The runtime model instance feature appears to be working correctly & the same as this branch

> Maybe I can allocate time for another dedicated cleanup pass for the ModelSceneGraph which replicates the cleanup from the commit linked above. We have to fix that bounding volume issue...

Does the commit fix the bounding volume issues you are referring to. Or is it just an incremental step towards helping isolate and later fix?

>  it looks like the current state also works for a model with a non-identity modelMatrix

That is good to hear. If you have an example sandcastle can you share the link? I had issues when I briefly tried testing this earlier but ran out of time to verify. 

> One open point might be whether model.boundingSphere should take into account the instances. 

Worth looking into further and discussing.

lukemckinstry:
Added a new Sandcastle `3D Model Instancing Terrain.html` in this commit https://github.com/CesiumGS/cesium/pull/12588/commits/95e92f3c4c9e7930f54021d72797138893cf506a

But the performance is slower than expected with roughly 1400 instances. On the initial sandcastle used for development `3D Model Instancing.html` performance was strong with up to 10000 instances (in a more contained area, there may be other factors to help explain too).

Dropping this note here as a reminder to investigate and perform benchmarking.

javagl:
> Does the commit fix the bounding volume issues you are referring to. Or is it just an incremental step towards helping isolate and later fix?

This commit did _not_ change the _functionality_ (i.e. also not fix the issue). 

The reason for that change was that `buildDrawCommands` does some farily unrelated things, namely

- building the render resources
- computing the bounding volumes
- creating the draw commands

(c.f. ["cohesive"](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#functions)). So I wanted to break it into `buildRenderResources`, `computeBoundingVolumes`, and `createDrawCommands` (and _document_ what each function is doing).  

The fix for the bounding volume computation would then go into ... drumroll ... `computeBoundingVolumes` 🙂 And I think that the change that is described at https://github.com/CesiumGS/cesium/issues/12634#issuecomment-2924994801 (a fairly trivial one, after all) could already solve most error cases (and be a general improvement, even if some cases might still be "not perfect"). 

Once the changes from _this_ PR are in, I'll probably ...

- ...replay the changes from the commit that caused the conflict (because I think that it makes sense to break that down into multiple functions)
- ...add the fix (or at least, improvement) for https://github.com/CesiumGS/cesium/issues/12108 (and all the linked/related issues)


---

> That is good to hear. If you have an example sandcastle can you share the link?

I've been doing some tests that had actually been more related to the bounding volume issue, but also to ... all other issues that may be related to "transforms" in one way or another, and I ran some tests with these on the instancing branch as well. These tests included the creation of different flavors of a "unit cube" with a very elaborate texture, to **see** whether something is right or wrong...

Cesium UnitCubes Instancing

[unitCube.zip](https://github.com/user-attachments/files/20657828/unitCube.zip)

However, here is a sandcastle that has a flag `useModelMatrix`:
 
- when this is `true`, then the `model.modelMatrix` will be set for the geo-placement
- when this is `false`, then the geo-placement will be "baked" into the instance transforms

The instance transforms are just a grid of  `size x size x size` instances, with the instance at (x, y, z) receives a translation/rotation/scale that is interpolated from a given range, for that grid position. (Again: To **see** whether something is right or wrong).

I changed the URL to the MilkTruck - it doesn't matter for now, and also looks neat:

Cesium Instances Grid

```javascript

  //url: "../../SampleData/models/unitCube/unitCube_T_TR.glb",
  url: "../../SampleData/models/CesiumMilkTruck/CesiumMilkTruck.glb",
});

const useModelMatrix = false;

const geoTransform = Cesium.Transforms.eastNorthUpToFixedFrame(
  Cesium.Cartesian3.fromDegrees(-75.1652, 39.9526, 20)
);

if (useModelMatrix) {
  model.modelMatrix = geoTransform;
}

const size = 5;
const minTranslationX = 0;
const maxTranslationX = 20;
const minTranslationY = 0;
const maxTranslationY = 20;
const minTranslationZ = 0;
const maxTranslationZ = 20;

const minRotationDegX = 0;
const maxRotationDegX = 90;
const minRotationDegY = 0;
const maxRotationDegY = 90;
const minRotationDegZ = 0;
const maxRotationDegZ = 90;

const minScaleX = 1;
const maxScaleX = 2;
const minScaleY = 1;
const maxScaleY = 2;
const minScaleZ = 1;
const maxScaleZ = 2;

function matrixFromAngleDegX(angleDegX) {
  const m = Cesium.Matrix3.fromRotationX(Cesium.Math.toRadians(angleDegX));
  return Cesium.Matrix4.fromRotation(m);
}
function matrixFromAngleDegY(angleDegY) {
  const m = Cesium.Matrix3.fromRotationY(Cesium.Math.toRadians(angleDegY));
  return Cesium.Matrix4.fromRotation(m);
}
function matrixFromAngleDegZ(angleDegZ) {
  const m = Cesium.Matrix3.fromRotationZ(Cesium.Math.toRadians(angleDegZ));
  return Cesium.Matrix4.fromRotation(m);
}

for (let x = 0; x < size; x++) {
  for (let y = 0; y < size; y++) {
    for (let z = 0; z < size; z++) {
      const alphaX = x / (size - 1);
      const alphaY = y / (size - 1);
      const alphaZ = z / (size - 1);

      const translationX =
        minTranslationX + alphaX * (maxTranslationX - minTranslationX);
      const translationY =
        minTranslationY + alphaY * (maxTranslationY - minTranslationY);
      const translationZ =
        minTranslationZ + alphaZ * (maxTranslationZ - minTranslationZ);
      const translation = new Cesium.Cartesian3(
        translationX,
        translationY,
        translationZ
      );
      const translationMatrix = Cesium.Matrix4.fromTranslation(translation);

      const rotationDegX =
        minRotationDegX + alphaX * (maxRotationDegX - minRotationDegX);
      const rotationDegY =
        minRotationDegY + alphaY * (maxRotationDegY - minRotationDegY);
      const rotationDegZ =
        minRotationDegZ + alphaZ * (maxRotationDegZ - minRotationDegZ);
      const rotationMatrix = Cesium.Matrix4.clone(Cesium.Matrix4.IDENTITY);
      Cesium.Matrix4.multiply(
        rotationMatrix,
        matrixFromAngleDegX(rotationDegX),
        rotationMatrix
      );
      Cesium.Matrix4.multiply(
        rotationMatrix,
        matrixFromAngleDegY(rotationDegY),
        rotationMatrix
      );
      Cesium.Matrix4.multiply(
        rotationMatrix,
        matrixFromAngleDegZ(rotationDegZ),
        rotationMatrix
      );

      const scaleX = minScaleX + alphaX * (maxScaleX - minScaleX);
      const scaleY = minScaleY + alphaY * (maxScaleY - minScaleY);
      const scaleZ = minScaleZ + alphaZ * (maxScaleZ - minScaleZ);
      const scale = new Cesium.Cartesian3(scaleX, scaleY, scaleZ);
      const scaleMatrix = Cesium.Matrix4.fromScale(scale);

      let instanceTransform = Cesium.Matrix4.clone(Cesium.Matrix4.IDENTITY);
      Cesium.Matrix4.multiply(
        instanceTransform,
        translationMatrix,
        instanceTransform
      );
      Cesium.Matrix4.multiply(
        instanceTransform,
        rotationMatrix,
        instanceTransform
      );
      Cesium.Matrix4.multiply(
        instanceTransform,
        scaleMatrix,
        instanceTransform
      );

      // When not using the modelMatrix, bake the geoTransform
      // into the instanceTransform
      if (!useModelMatrix) {
        Cesium.Matrix4.multiply(
          geoTransform,
          instanceTransform,
          instanceTransform
        );
      }
      model.instances.add(instanceTransform);
    }
  }
}

  console.log(model.boundingSphere);
  //viewer.scene.camera.flyToBoundingSphere(model.boundingSphere);
});

viewer.scene.camera.setView({
  destination: new Cesium.Cartesian3(
    1253512.5232461668,
    -4732922.214567729,
    4074115.474546098
  ),
  orientation: new Cesium.HeadingPitchRoll(
    2.205737333179613,
    -0.7255022564055849,
    6.283181225638178
  ),
});
```

Note: The **appearance** of the instances is different, depending on whether the model matrix is used. This **might** just be another flavour of https://github.com/CesiumGS/cesium/issues/12310 ...

Apps/Sandcastle/gallery/3D Model Instancing Terrain.html
```diff
@@ -0,0 +1,647 @@
...
+        //Sandcastle_Begin
+        const viewer = new Cesium.Viewer("cesiumContainer", {
+          terrain: Cesium.Terrain.fromWorldTerrain(),
+        });
+
+        let model;
+        try {
+          model = await Cesium.Model.fromGltfAsync({
+            //url: "../../SampleData/models/GroundVehicle/GroundVehicle.glb",
+            url: "../../SampleData/models/wind_turbine.glb",
```
javagl, line 39: This doesn't seem to be in the repo yet. I'm just mentioning it. It could make sense to add this last, when all other decisions are made, to not bloat the git history with possible additions/removals of large binaries. (BTW: Also make sure to check the model copyright).

rudacs:
I'm using 1.96 because of ModelInstanceCollection.

And I'd love to see it come back in the future. :)

I'd like to request an improvement over what already existed.
I can't highlight an instance, like showing that the user selected that item or multiple items.

DaviCoder:
I'm also experiencing this issue, and it's quite impactful for my workflow, especially when dealing with multiple selections. I agree with the original request; being able to highlight selected instances, similar to how it worked in earlier versions (e.g., 1.96 with `ModelInstanceCollection`), would be a significant improvement. I'd love to see this functionality return or be implemented in a new, robust way.

lukemckinstry:
> Added a new Sandcastle `3D Model Instancing Terrain.html` in this commit [95e92f3](https://github.com/CesiumGS/cesium/commit/95e92f3c4c9e7930f54021d72797138893cf506a)
> 
> But the performance is slower than expected with roughly 1400 instances. On the initial sandcastle used for development `3D Model Instancing.html` performance was strong with up to 10000 instances (in a more contained area, there may be other factors to help explain too).
> 
> Dropping this note here as a reminder to investigate and perform benchmarking.

The original wind turbin glTF I used
https://www.fab.com/listings/c6efa5d8-d94d-4da0-8747-09e0d4af7fb1 is 14MB, which is 10-20x bigger than other models we use in `SampleData`. 

So I swapped in this wind turbin https://www.fab.com/listings/3f826cd9-1caf-4a6a-9e8c-42118c1297cf that is 4MB (still bigger than most of the SampleData) and there is no observable lag for the sandcastle. 

The sandcastle demo plots 1400 instances over a wide area (state of New York). So it seems like this is acceptable performance for this feature, but I'll continue more benchmarking to understand roughly where the performance limits are.

Apps/Sandcastle/gallery/3D Model Instancing Terrain.html
lukemckinstry: Thanks I agree it makes sense to wait to add the model until the end and we are sure we want to use them.

The model is free to download here https://www.fab.com/listings/3f826cd9-1caf-4a6a-9e8c-42118c1297cf  

As detailed in this comment https://github.com/CesiumGS/cesium/pull/12588#issuecomment-3084511189 I initially used a much larger model https://www.fab.com/listings/c6efa5d8-d94d-4da0-8747-09e0d4af7fb1  but this caused performance issues. 

Both models have the [Creative Commons Attribution (CC BY 4.0)](https://creativecommons.org/licenses/by/4.0/) license, so to my understanding we are free to use the models as long as we provide attribution

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
lukemckinstry: I thought the crux of this test should be to make sure that when `ModelInstancesUpdateStage.update` runs, if `sceneGraph.modelInstances._instances` has changed, `runtimeNode.instancingTransformsBuffer` is updated accordingly. 

The output of `RuntimeModelInstancingPipelineStage._getTransformsTypedArray` (when run on the changed values of `sceneGraph.modelInstances._instances` in `ModelInstancesUpdateStage.update`) is not stored anywhere, so I thought testing the buffer made sense.

lukemckinstry:
> Now... the current state of the `ModelSceneGraph` _in this branch_ has been changed significantly. (Some 
> ...
> Maybe I can allocate time for another dedicated cleanup pass for the `ModelSceneGraph` which replicates the cleanup from the commit linked above. We _have_ to fix that bounding volume issue...

I rebased and resolved the merge conflicts between this branch and `main` in a way to preserve @javagl's refactor of ModelSceneGraph.buildDrawCommands. The refactor definitely made things more clear!

javagl:
There still seem to be merge conflicts in `ModelSceneGraph`. Back in https://github.com/CesiumGS/cesium/pull/12588#issuecomment-2945211831 , I mentioned that just "ignoring" my changes could be OK (they hadn't been "functional" changes, but only cleanups, to prepare the stage for finally tackling https://github.com/CesiumGS/cesium/issues/12108). However, I'll revisit this when the dust here has settled.

packages/engine/Specs/Scene/Model/RuntimeModelInstancingPipelineStageSpec.js
lukemckinstry: add logic to also test a scale transform

packages/engine/Specs/Scene/Model/ModelInstanceSpec.js
lukemckinstry: removed this logic to spy and check toHaveBeenCalled since it was not necessary. The current logic checks that the bounding volume is created correctly according to the instances.

packages/engine/Source/Scene/Model/ModelSceneGraph.js
```diff
@@ -183,6 +179,7 @@ Object.defineProperties(ModelSceneGraph.prototype, {
     },
   },
 
+  // TODO: Should model.modelMatrix really be stored in this class? Or should it live only at the model level?
```
lukemckinstry: After reviewing I think `computedModelMatrix`  can live at the model level. 

If we did this, it would also make sense to store glTF `components` at the model level, which I also reviewed and appears to be fine. Right now we pass `components` from the loader into the model scene graph and reference is there.

packages/engine/Source/Scene/Model/ModelSceneGraph.js
```diff
@@ -433,8 +539,8 @@ function traverseAndCreateSceneGraph(sceneGraph, node, transformToRoot) {
   // Create and store the public version of the runtime node.
   const name = node.name;
   if (defined(name)) {
-    const model = sceneGraph._model;
     const publicNode = new ModelNode(model, runtimeNode);
+    // TODO: Why does `_nodesByName` live on the model and not here?
```
lukemckinstry: no reason this cannot live on ModelSceneGraph instead of Model. 

The Model.getNode uses the property and can remain as is.

packages/engine/Source/Scene/Model/ModelInstance.js
```diff
@@ -0,0 +1,251 @@
...
+   *
+   * const boundingSphere = modelInstance.getBoundingSphere(model);
+   * viewer.camera.flyToBoundingSphere(boundingSphere);
+   */
+  getBoundingSphere(model, result) {
+    const modelMatrix = model.modelMatrix;
+    const sceneGraph = model.sceneGraph;
+    const instanceBoundingSpheres = [];
+
+    // TODO: runtime nodes should probably have a internal public accessor
```
lukemckinstry: Adding an internal public accessor for ModelSceneGraph._runtimeNodes seems fine and not problematic to add.

does this also indicate that ModelSceneGraph should be a public class?

packages/engine/Source/Scene/Model/ModelInstance.js
```diff
@@ -0,0 +1,251 @@
...
+  getPrimitiveBoundingSphere(
+    modelMatrix,
+    sceneGraph,
+    runtimeNode,
+    primitiveBoundingSphere,
+    result,
+  ) {
+    result = result ?? new BoundingSphere();
+
+    // TODO: Can we precompute this in sceneGraph.updateRuntimeNodeTransforms?
```
lukemckinstry, line 280: I think using a value runtimeNode.computedTransform precomputed in sceneGraph.updateRuntimeNodeTransformsin would be a performance improvement. I am still considering if this would make it easier or more difficult to ensure the most up to date value is used here for runtimeNode.computedTransform. Calculating from the original values as we do now is potentially expensive but it is easier to ensure computedTransform is up to date.

Apps/Sandcastle/gallery/3D Model Instancing.html
lukemckinstry: `Apps/Sandcastle/gallery/3D Model Instancing Terrain.html` is the sandcastle I think is best to release. This has a new wind turbine glb added to the SampleData folder. I believe the CC license makes using it ok https://github.com/CesiumGS/cesium/pull/12588#discussion_r2213680597

lukemckinstry:
ready for review @ggetz

lukemckinstry:
2D and Columbus support

 - 2D - My hunch is this requires setting `modelView` correctly in the shader
 - Columbus - we don't support for i3dm, so I don't think we should support for runtime model instancing since we dont have that working template to start from

lukemckinstry:
> 2D and Columbus support

This sandcastle [link](http://localhost:8080/Apps/Sandcastle/index.html#c=pVhpb9s2GP4rnLEP8urQkmzZOZxgXa4VWLMgyfqh9bAyMm0zoUSDopN6hf/7XpI6KFlZug5IZJHvfT2UFIs0U+iJ0Wcq0TFK6TM6pRlbJ/iD2fOmndisT0WqCEupnHZ6X6cpQllMU/pezOhhIXFb7ODb0/Or8/Bsmm67R9N0mnKqUAIEDislN8goMBtgkzwTpgodWpzjuRTJJVfzt9kmjT3DjdBa8kM07WDch79bkqw4PSOK9I2erH8pxTqdfaBLFnNaX+EFvwe3rZqEpSxZJ9fsC+W37G9wfzQ0JOMrynOBTXh4JVnCFHuiGSazmWdMabYtiomKl8ijUgrZtQHFkEsB1rhYeJ8vCON0hpRAXJCZDRajH78age1no0SnJjYF4OkCMrE3jnAwisKjcpso2B4c4IMoHJWZXIkMfBJpucGAl6QxvZMkzeZCJkCBH+QZImjwj+BnggL9++ZN7m6hBuh58k+JVHBH0oGpwBldSEozz+ZNu/gG1PyE3hO1xGBqJhKvC2sf+74f5OnVLls2Z9tmdsfNesOV2xmmJFNXQqrlH6s7cQGVml1IklCvcNnqszkttNoK7dgoE12rawzaJMEZVbrNbYfNaKZYSrT+Q9evKit5JoIwGkRBiKNwEA5HwWi0n8e+NxwPwoMwxGEwjEbjcXiQE4b+eBgEER6Oh9Fw5B9Yga65CskojNaO2V8pmbF0cc2gz24E57lxUO5H48F4MBgE44NRMChs+3gcRpEfhtFo6EfR/rAwPsLh/iDYDzRlsB+MK+POdBZ5O7JLKaxH9QpB4SX7MvC6OReQ3Jbr93VJ2HyD1JKWGpEqy32/Qdl6teIbiAsRo7kwlCEyVwBBBMrAyWaaQmnuWELFWnnQZMcntmfburrZBpymC7V0O71qPZBqsi+o8phtKOTGXYt5qNlucmLZZLgMrVdKFprc3FTaYi5S2iJvC6Hl7OCTdAHoZucPpJsTF+z7RzvsN9AvoK7m+hIrke97rs5uTbzw/WMzbIsDRdwfPddOXQWEe/NC7gY4WXPFoOx5C1dp/thr7pQbbW2XE78/xVqqUdfMqaursOeG1KvZKjTt2rDj4s4EHCswZz04t0aAhtOOLTTA8zqNTbLgsLhdLakEnM27lc2R94PtUgkYsCmbWFK1lmmhFKHyJEXILUW2uM/KPjdwdynJaonv9YkIk2fNOdUrzqxpx4jCSWl0dC1HjpuAUQC9Ocp+LapUAPKhkcBgDIa4rCHlnK0ywWaHqJSAMKB/2EsIa9RojnXWe+G+20OANPYeaoASCjazSn9CYM0IL59LTgUXEl/enJ9f4Wemlm/5akk8H4+6padbe7MtSrt7tbktYaMtwX+ZnXcFS30+gNQuJNeQ2YReaYZcpES5x2P/6HGyA22PDrK5noH+Oq49FvHUC62ZKkqFqWCNTYynhSVWs9RsF82K9vZOTMsYuU/sz0qx65oObxdUd4ChGCsHWV1SYaMi34kbIVSdqYkcQ69b0ZvuVdGUbupoisVuNGWR/0NIturmavl6zYC1sv8RRZXka5G1Hl4GlHh1fhmDLwxh95Wov8VGTeCbDe1WQ9tyCwJrV6jf/1d80gwuSjXD6NUZ2wFLU14FrgAed3uour6GUoXS70ErK711l9vdklXvLvpYKiYnh5vrkujIlUjwAEjwMHFefnI4eGjAQbNopYQpWSX/6aEGCwWmut6aM6sm0X5gvSRfryzoaW3SYjp/2VwLlqrakKK6jl6dtnu67cyp29guUz30lzLWGoBOYxuhrvKVEagf1G3qGjG1j8E3nd/ViY3abl+bilfO8Jvzs9pMBN2G51t3uX2p50qubb5pNrZoDq9/nJcPVTGnRL7TNX8iHEDT3rybdatHMPvvvMhXXNB88GxZilfPeRoffL+rZTq9ziRTG05PCo9+ZskKXnz1xw4P476iyQoAlWb9+3X8SBWOs7L2k74rOpmxJ8Rmxy3fayAQkmVAma+5+eQx7ZxM+sC/I6o/VcDA/f5EpX4JA7ZlcPKb3cQYT/qwbJdUQvB7Ihua/wE) provides code to check the bounding spheres and they look correct to me in SCENE2D.

One note, `SCENE3D` and `SCENE2D` use `ModelRuntimePrimitive.boundingSphere`. `ModelRuntimePrimitive.boundingSphere2D` is only used when `Model.ProjectTo2D` is true.

rudacs:
@lukemckinstry testing with these new commits, I noticed a problem.

If I have several instances, one with the correct position and another, with Matrix4.ZERO.

And when I try to pick, it ends up selecting the wrong instance.
It seems that when the instance is Matrix4.ZERO, it places it in the same position as instance 0.

```
//modelMatrix = ...
model.instances.add(modelMatrix);
model.instances.add(Matrix4.ZERO);
```

By removing it from ModelInstancesUpdateStage.js, it works again.

```
sceneGraph.modelInstances._dirty = false;
```

lukemckinstry:
> And when I try to pick, it ends up selecting the wrong instance. It seems that when the instance is Matrix4.ZERO, it places it in the same position as instance 0.

Can you explain your reasoning for supplying the zero transform in model.instances.add(Matrix4.ZERO)? The design we had in mind is that each instance would have some non-zero and non-identity transform.

rudacs:
> > And when I try to pick, it ends up selecting the wrong instance. It seems that when the instance is Matrix4.ZERO, it places it in the same position as instance 0.
> 
> Can you explain your reasoning for supplying the zero transform in model.instances.add(Matrix4.ZERO)? The design we had in mind is that each instance would have some non-zero and non-identity transform.

I've seen better performance from having multiple instances already created and simply repositioning them, rather than creating and removing them every time.

Since I have thousands, I only instantiate what appears on the screen.

rudacs:
Another problem I encountered is that if you remove and add another item, it doesn't "remove" or "add"
I imagine something in the position update

```

let model;
try {
  model = await Cesium.Model.fromGltfAsync({
    url: "../../SampleData/models/wind_turbine.glb"
  });
} catch (error) {
  console.log(`Failed to load model. ${error}`);
}

const positions = [
  42.77379, -78.28529,
  42.77549, -78.28439
];

let position;

let instanceTransform;
for (let i = 0; i < positions.length; i += 2) {
  const chunk = positions.slice(i, i + 2);
  position = Cesium.Cartesian3.fromDegrees(chunk[1], chunk[0], 0);
  instanceTransform = new Cesium.Transforms.eastNorthUpToFixedFrame(position);
  model.instances.add(instanceTransform);
}

viewer.scene.camera.setView({
  destination: new Cesium.Cartesian3(
    950971.9555012125,
    -4592451.1790314745,
    4309468.958291063,
  ),
  orientation: new Cesium.HeadingPitchRoll.fromDegrees(80.61, -7.2, 0.0),
});

function handleSelection(screenPosition) {
  const picked = viewer.scene.pick(screenPosition, 15, 15);
  
  if (!picked) {
    return;
  }
  
  const instance = model.instances.get(picked.instanceId);
  model.instances.remove(instance);

  const position = Cesium.Cartesian3.fromDegrees(-78.2828, 42.77989, 0);
  const instanceTransform = new Cesium.Transforms.eastNorthUpToFixedFrame(position);
  model.instances.add(instanceTransform);
}

const handler = new Cesium.ScreenSpaceEventHandler(viewer.camera.canvas);
handler.setInputAction((click) => handleSelection(click.position), Cesium.ScreenSpaceEventType.LEFT_CLICK);
```

javagl:
>  The design we had in mind is that each instance would have some non-zero and non-identity transform.

I don't know why there should be a requirement for "non-identity", but haven't been tracking the development here closely (recently). That may depend on whether the transforms are ~"expected to be in some 'local' space or not", and I assume that this will be pointed out clearly in the documentation.

The point about the 'zero' transform is an interesting and valid one, and likely what rudacs referred to: I think that it is not uncommon to use a 'scale of 0.0' to essentially _hide_ something. And there are several trade-offs, where the question of what is 'good' or 'bad' strongly depends on the usage pattern. Imagine that you want to start showing 0 instances, then incrementally show 1...10000 instances, then incrementally hide them again (think of something like ~"time-dependent construction sites on the globe" or so). In this case, using 'zero' to make an instance invisible is certainly preferable in terms of performance. The alternative would probably be to start with an instances buffer of length 0, and then do 10000 **re**-allocations (including copies) to expand that, and then another 10000 **re**-allocations incrementally shrink that buffer. That's one way of keeping the memory bus busy. Anticipating this difference, from an idealistic, longer-term engineering perspective, could lead to the consideration to differentiate beween `DynamicInstances` and `StaticInstances`, but ... that's too idealistic.

lukemckinstry:
Thanks for pointing out the behavior you see @rudacs with regards to picking when an instance transform is set to Matrix4.ZERO.

And thanks as well for pointing out the issue around the ModelInstanceCollection `add` or `remove` functions and for proposing a fix. We will review shortly. 

> using 'zero' to make an instance invisible is certainly preferable in terms of performance. The alternative would probably be to start with an instances buffer of length 0, and then do 10000 re-allocations (including copies) to expand that, and then another 10000 re-allocations incrementally shrink that buffer.

Thanks for making this clear @javagl. The design of the current implementation so far:
 - When ModelInstanceCollection `add` or `remove` is called, we run `resetDrawCommands` so everything in the shader is set up again. All the PipelineStages including RuntimeModelInstancingPipelineStage are run again.
 - When a single instance is edited, we rebuild the buffer `runtimeNode.instancingTransformsBuffer`. This is done in ModelInstancingUpdateStage. As pointed out by javagl rebuilding this entire buffer is a larger operation than needed.

So definitely seems possible to optimize design in the manner you point out as possible to better fit this use case of showing/hiding instances by setting scale to 0.

rudacs:
@lukemckinstry 

I was implementing a show inside the ModelInstance, passing a variable to the FS and discarding it.
This would make it more readable that I want to hide that element.

What do you think? Would it be ideal?

I also implemented it in my branch to allow changing the color of the instance.
https://github.com/CesiumGS/cesium/pull/12836

rudacs:
> Another problem I encountered is that if you remove and add another item, it doesn't "remove" or "add" I imagine something in the position update
> 
> ```
> const viewer = new Cesium.Viewer("cesiumContainer");
> 
> let model;
> try {
>   model = await Cesium.Model.fromGltfAsync({
>     url: "../../SampleData/models/wind_turbine.glb"
>   });
>   viewer.scene.primitives.add(model);
> } catch (error) {
>   console.log(`Failed to load model. ${error}`);
> }
> 
> const positions = [
>   42.77379, -78.28529,
>   42.77549, -78.28439
> ];
> 
> let position;
> 
> let instanceTransform;
> for (let i = 0; i < positions.length; i += 2) {
>   const chunk = positions.slice(i, i + 2);
>   position = Cesium.Cartesian3.fromDegrees(chunk[1], chunk[0], 0);
>   instanceTransform = new Cesium.Transforms.eastNorthUpToFixedFrame(position);
>   model.instances.add(instanceTransform);
> }
> 
> viewer.scene.camera.setView({
>   destination: new Cesium.Cartesian3(
>     950971.9555012125,
>     -4592451.1790314745,
>     4309468.958291063,
>   ),
>   orientation: new Cesium.HeadingPitchRoll.fromDegrees(80.61, -7.2, 0.0),
> });
> 
> function handleSelection(screenPosition) {
>   const picked = viewer.scene.pick(screenPosition, 15, 15);
>   
>   if (!picked) {
>     return;
>   }
>   
>   const instance = model.instances.get(picked.instanceId);
>   model.instances.remove(instance);
> 
>   const position = Cesium.Cartesian3.fromDegrees(-78.2828, 42.77989, 0);
>   const instanceTransform = Cesium.Transforms.eastNorthUpToFixedFrame(position);
>   model.instances.add(instanceTransform);
> }
> 
> const handler = new Cesium.ScreenSpaceEventHandler(viewer.camera.canvas);
> handler.setInputAction((click) => handleSelection(click.position), Cesium.ScreenSpaceEventType.LEFT_CLICK);
> ```
> 



Changing the position also has no effect.

```

let model;
try {
  model = await Cesium.Model.fromGltfAsync({
    url: "../../SampleData/models/wind_turbine.glb"
    //minimumPixelSize: 32,
  });
} catch (error) {
  console.log(`Failed to load model. ${error}`);
}

const lng = -75.1652;
const lat = 39.9526;
const positions = [
  42.77789, -78.2837
];

let position;

let instanceTransform;
for (let i = 0; i < positions.length; i += 2) {
  const chunk = positions.slice(i, i + 2);
  position = Cesium.Cartesian3.fromDegrees(chunk[1], chunk[0], 0);
  instanceTransform = new Cesium.Transforms.eastNorthUpToFixedFrame(position);
  const inst = model.instances.add(instanceTransform);
  
  inst.color = i % 3 === 0 ? Cesium.Color.GREEN : Cesium.Color.BLUE;
}

viewer.scene.camera.setView({
  destination: new Cesium.Cartesian3(
    950971.9555012125,
    -4592451.1790314745,
    4309468.958291063,
  ),
  orientation: new Cesium.HeadingPitchRoll.fromDegrees(80.61, -7.2, 0.0),
});

function handleSelection(screenPosition) {
  const instance = model.instances.get(0);

  const position = Cesium.Cartesian3.fromDegrees(-78.2828, 42.77989, 0);
  const instanceTransform = Cesium.Transforms.eastNorthUpToFixedFrame(position);
  
  instance.position = instanceTransform;
}

const handler = new Cesium.ScreenSpaceEventHandler(viewer.camera.canvas);
handler.setInputAction((click) => handleSelection(click.position), Cesium.ScreenSpaceEventType.LEFT_CLICK);
```

This partially solved it, but I realized that it should actually be adding and removing instances, which would be the same amount. So it would be like updating the position; the vertex should pass through. But if I keep it here, it doesn't work.

```
ModelInstancesUpdateStage.update...
sceneGraph.modelInstances._dirty = false;
```

I'm trying to see the solution for this.
I'm trying to find a solution for this, but so far I've had no luck.

hh-hang:
Hello, will this feature be updated and added to a new version?

darcyvdd:
@lukemckinstry is this still being progressed? Some renewed interest on the forum [here](https://community.cesium.com/t/questions-about-modelinstancecollection-and-primitivecollection-as-an-alternative/42459/8)

ggetz:
Just an update on this: I think the main blocker is the gltf loading and model pipeline refactoring that this change necessitates. I think breaking that into it's own PR and making changes incrementally may be the best path forward.

If time allows this week, I'd like to take a pass.

PR #12591, opened 2025-04-29
Update npm packages

jjspace:
# Description
`npm outdated`:
```
Package  Current  Wanted  Latest  Location              Depended by
express   4.21.2  4.21.2   5.1.0  node_modules/express  cesium
jsdoc     3.6.11  3.6.11   4.0.4  node_modules/jsdoc    cesium
lerc       2.0.0   2.0.0   4.0.4  node_modules/lerc     engine@npm:@cesium/engine@16.0.0
rbush      3.0.1   3.0.1   4.0.1  node_modules/rbush    engine@npm:@cesium/engine@16.0.0
```

- express - Updated, see below
- jsdoc - held back https://github.com/CesiumGS/cesium/issues/10455
- lerc - held back https://github.com/CesiumGS/cesium/issues/11034
- rbush - held back https://github.com/CesiumGS/cesium/issues/12203

Express [switched the `@latest` version to v5](https://expressjs.com/2025/03/31/v5-1-latest-release.html). This came with the major changes from v4 -> v5, [migration guide](https://expressjs.com/en/guide/migrating-5). Our server is fairly minimal so it didn't need that many changes. I believe I got them all.

- [`express.static.mime`](https://expressjs.com/en/guide/migrating-5#express.static.mime) no longer exists. They say to switch to [`mime-types`](https://github.com/jshttp/mime-types) but that does _not_ have the `.define()` we were using.
    - Switched to a custom middleware function instead
- wildcards in paths must now be named per the [new path matching syntax](https://expressjs.com/en/guide/migrating-5#path-syntax)
    - Switched some routes to regular expressions instead where the wildcard param was not used
    - Switched the proxy wildcard to a named param and updated handlers 

## Issue number and link

No issue, part of release prep

## Testing plan

- Run `npm start` and make sure that everything on the local dev server works as expected.
- You can use the script below to test routes with specific extensions for the correct mime type
- Check the proxy routes still work
   - http://localhost:8080/proxy/http://example.com/file?query=1
   - http://localhost:8080/proxy/?http%3A%2F%2Fexample.com%2Ffile%3Fquery%3D1

<details><summary>mime type test script</summary>
<p>

```js
async function checkContentType(route) {
  const request = await fetch(route);
  return request.headers.get("Content-Type");
}

const testUrls = {
  "http://localhost:8080/Apps/SampleData/simple.czml": "application/json",
  "http://localhost:8080/Apps/SampleData/population909500.json":
    "application/json",
[...]

PR #12597, opened 2025-04-30
Add wildcards back to routes in local server

jjspace:
# Description

As part of https://github.com/CesiumGS/cesium/pull/12591 @ggetz removed some of the wildcards from routes. On the surface this seemed fine, especially with the [path examples docs](https://expressjs.com/en/5x/api.html#path-examples) saying "This will match paths starting with...". However that's not how it actually works and the wildcards are actually necessary.

I also think the problem was masked by the fact that some of the routes that use the wildcards are also routes that exist on disk if you've run certain commands so they're served statically making it _seem_ like they're working. For example anything under `Build/Specs` exists on disk and will load as static files if the wildcard doesn't pick up that it may need to rebuild from the context cache.

This PR just adds them back. I opened a PR for secondary validation in case I'm missing something since express sometimes confuses me and their docs aren't always as exhaustive as I'd like.

## Issue number and link

No issue, follow up to https://github.com/CesiumGS/cesium/pull/12591

## Testing plan

- Check _both_ of the proxy methods still work
   - http://localhost:8080/proxy/http://example.com/file?query=1
   - http://localhost:8080/proxy/?http%3A%2F%2Fexample.com%2Ffile%3Fquery%3D1
- Recommend running a `git clean` to make sure there are no unnecessary build files to confuse
- Check some other urls that fall under the wildcard routes.
    - NOTE: you should also see a log statement from the server for each of these like `Built Specs/* in 0.3 seconds.` if the handler is set up correctly
    - http://localhost:8080/Build/CesiumUnminified/Workers/decodeI3S.js
    - http://localhost:8080/Build/Specs/karma-main.js
    - http://localhost:8080/Build/Specs/TestWorkers/createBadGeometry.js

# Author checklist

### DIFF
--- a/server.js
+++ b/server.js
@@ -187,7 +187,7 @@ async function generateDevelopmentBuild() {
     const workersCache = createRoute(
       app,
       "Workers/*",
-      "/Build/CesiumUnminified/Workers",
+      "/Build/CesiumUnminified/Workers/*file.js",
       contexts.workers,
     );
 
@@ -240,7 +240,7 @@ async function generateDevelopmentBuild() {
     const testWorkersCache = createRoute(
       app,
       "TestWorkers/*",
-      "/Build/Specs/TestWorkers",
+      "/Build/Specs/TestWorkers/*file",
       contexts.testWorkers,
     );
     chokidar
@@ -250,7 +250,7 @@ async function generateDevelopmentBuild() {
     const specsCache = createRoute(
       app,
       "Specs/*",
-      "/Build/Specs",
+      "/Build/Specs/*file",
       contexts.specs,
     );
     const specWatcher = chokidar.watch(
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -347,7 +347,7 @@ async function generateDevelopmentBuild() {
   });
 
   //eslint-disable-next-line no-unused-vars
-  app.get("/proxy{/:remote}", function (req, res, next) {
+  app.get("/proxy{/*remote}", function (req, res, next) {
```
ggetz, line 350: I don't think this will be picked up as a parameter without `:`. What am I missing?

ggetz (commented):
Ah, my @jjspace. That's frustrating that the express docs are just plain wrong.

server.js
jjspace: Using the `:` will not allow multiple nested routes. This is what breaks this version of the proxy url: http://localhost:8080/proxy/http://example.com/file?query=1

I don't know why it's not documented well in the [route params](https://expressjs.com/en/guide/routing.html) docs, maybe another issue to open on their side. Doing it this way is how they say to replace wildcard routes in the [Migrating to v5 page](https://expressjs.com/en/guide/migrating-5.html#path-syntax)

ggetz: So does the wildcard still triggers the `app.param` callback?

ggetz (approved):

Issue #12355, opened 2024-12-02
Automate Playwright testing during CI


jjspace:
One of the slowest parts of our current release process is running our end to end tests through Playwright. It's also very late in the process to catch any serious errors that could prevent or hold up doing the release. Playwright has [fairly robust docs](https://playwright.dev/docs/ci) for how to use it in CI but we currently don't so I'd like to start the discussion around why and whether we can enable it.

1. Why didn't we set it up with CI in the first place?
    - Looking at the [initial implementation PR](https://github.com/CesiumGS/cesium/pull/11348) it just wasn't priority at that time. I'd argue it maybe should be at this point.
2. Why are we not including snapshots in the repo itself? Is it a repo space issue?
    - I suspect one reason for this is the difference in rendering between systems. If that's the case I would still assume the _large_ majority of our tests _will_ look identical on all systems and we should set up a system to mark the ones that don't.
3. Why are we only running the e2e tests on releases? Is it just because they're too slow to run every pr? (CI might help reduce the need to run them locally)

This issue also closely couples with https://github.com/CesiumGS/cesium/issues/12281 because we will need to be much more precise with certain tests (like particles) so they actually _are_ perfectly reproducible every time.

jjspace:
Bumping this again after doing the most recent release. I'd love if we could even just get to a point that we can commit all the snapshots so we have a consistent comparison for everyone on the team without having to generate them from the previous release. That will require us to actually commit to generating them in a consistent manner which may be hard given different systems.

CC @ggetz

lukemckinstry:
Noticed this message in the playwrite logging when running `npm run test-e2e-update`

```
  Slow test file: [chromium] › Specs/e2e/sandcastle.spec.js (11.3m)
  Consider running tests from slow files in parallel. See: https://playwright.dev/docs/test-parallel
```

Perhaps this is an option for us to explore.

PR #12607, opened 2025-05-07
Do not install playwright in CI

jjspace:
# Description

When testing CI build support for #12574 I was running the CI builds a lot. I took a brief look into ways to speed up CI and realized we install all the playwright dependencies for every CI job even though none of them currently do anything with CI. On average the `npm install` steps seems to take `~1m 20s` in most jobs. Without doing the `playwright install` step this drops to only `~20-30s`. Compare the [`deploy` job in main](https://github.com/CesiumGS/cesium/actions/runs/14799064102/job/41553283927) to the [deploy job on this branch's commit](https://github.com/CesiumGS/cesium/actions/runs/14889959693/job/41819317670)

Once we decide to move forward with #12355 we can force it to install specifically in the CI workflows it's needed.

As an aside I did spend a [little time looking at](https://github.com/CesiumGS/cesium/tree/test-ci-cache) turning on [caching for the npm packages](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows) in CI. This seems like it would be really helpful but relies on a "cache key" that defines when to invalidate the cache. They heavily suggest using a `package-lock.json` file as that will always change with every dependency. However we don't use one so there's a decent chance it will not invalidate the cache when it actually should for version ranges in `package.json`. If we could find a way to set our own key based on the "wanted" values of `npm outdated` I think that would be our "ideal" setup. That said the node action currently [does not allow setting our own key](https://github.com/actions/setup-node/issues/1120) anyway but there's a PR for it.

On top of that the largest slowdown with the install step from Playwright is downloading the browser binaries and [they specifically say](https://playwright.dev/docs/ci#caching-browsers) that you should _not_ cache these. Even if you did rebuilding them would take as much time as re-downloading.

## Issue number and link

No issue

## Testing plan

- Check the CI runs of all jobs on this branch vs any others
- Run `npm install` locally and make sure playwright _does_ still install and download browsers locally
- Test with `CI=true npm install` to see that running it in CI does _not_ do the playwright install

# Author checklist


ggetz (changes requested):
Awesome, seems like a straightforward approach to a lot of saved time. :+1: from me. Just one comment on the script.

scripts/isCI.js
```diff
@@ -0,0 +1,21 @@
+// Logic and values based on is-ci https://github.com/watson/is-ci/blob/master/bin.js
+// which uses ci-info under the hood https://github.com/watson/ci-info/blob/master/index.js
+// This just extracts the specific parts we need without adding more dependencies
+
+const { env } = process;
+
+const isCI = !!(
+  env.CI !== "false" && // Bypass all checks if CI env is explicitly set to 'false'
```
ggetz, line 8: While I do appreciate the robust approach, why did we decide to cover all possible types of CI providers? Just remember that we'd need to maintain this script going forward, and I could see it becoming out of date quickly.

ggetz (approved):
All looks good to me! Thanks!

Issue #11446, opened 2023-08-02
Timeout failure in `Cesium3DTilesetSpec`


jjhembd:
A CI run threw the following error:

```bash
Scene/Cesium3DTileset
    ✗ Restrict tileset memory usage with maximumCacheOverflowBytes
	Error: Timeout - function did not complete within 5000ms in Build/Instrumented/SpecList.js (line 353664)
	poller@Specs/pollToPromise.js:26:11 <- Build/Instrumented/SpecList.js:353664:11
```

The test completed successfully on a re-run.

The reason for the timeout is not obvious. This spec is one of the simpler ones in the suite, and is only loading local data from `Specs/Data/`.

ggetz:
I've been seeing this often in CI. Maybe as a quick fix, we could pass through a higher `timeout` value through `Cesium3DTilesTester.waitForTilesLoaded` to `pollToPromise`.

jjhembd:
I saw this again in CI today

ggetz:
It's been a while, but this one just popped back up for me in CI:   
```
Scene/Cesium3DTileset
    ✗ Restrict tileset memory usage with maximumCacheOverflowBytes
	Error: Timeout - function did not complete within 5000ms
	    at poller (Specs/pollToPromise.js:26:11 <- Build/Specs/SpecList.js:27186:11)
```

PR #12609, opened 2025-05-08
Cleanup unused workflows: slack, concierge

ggetz:
# Description

With a reminder from https://github.com/CesiumGS/cesium/pull/12604, it's time to cleanup some unused workflows:

- Slack: We had an action which would alert us to failing builds in `main` via Slack. We've moved away from Slack internally. I don't see an immediate need for a replacement now that we use the merge queue. 
- Concierge: We haven't used this in years, so this just cleans up a few files.

## Issue number and link

N/A

## Testing plan

CI should pass. One merged, the `main` workflow should not run.

@jjspace Could you please take a pass?


ggetz:
A known test failure popped up (https://github.com/CesiumGS/cesium/issues/11446) but hopefully that was a fluke. Re-running CI.

jjspace (approved):
LGTM, cleanup is always nice

PR #12613, opened 2025-05-10
Fix #9393: Consistent _buttonDown usage in ScreenSpaceEventHandler.js

### LINKED ISSUE #9393: Inconsistent usage of _buttonDown in ScreenSpaceEventHandler
Sandcastle example: NA

Browser: MS Edge 85.0.564.68

Operating System: Windows 10

I would like to highlight something which may possibly be a potential issue.
I found out that when creating a new ScreenSpaceEventHandler. It will assign the following default state object to _buttonDown
`this._buttonDown = {
LEFT: !1,
MIDDLE: !1
RIGHT : !1
}`

Further debugging has led me to found out that methods such as handleMouseDown or handleMouseUp instead uses an integer value from 'button' to assign or get value from _buttonDown. Assuming a left doen click was done and never release, this will result _buttonDown to become as such
`{
0: true,
LEFT: false,
MIDDLE: false
RIGHT : false
}`

I believe the fix would be to just replace the key with the respective integer value.

ayushd150:
Changes made to _buttonDown in ScreenSpaceEventHandler.js

### DIFF
--- a/.github/actions/check-for-CLA/package.json
+++ b/.github/actions/check-for-CLA/package.json
@@ -3,6 +3,7 @@
     "version": "0.1.0",
     "main": "index.js",
     "dependencies": {
+        "@cesium/check-for-cla": "file:",
         "@octokit/core": "^6.1.2",
         "fs-extra": "^11.2.0",
         "googleapis": "^137.1.0",
--- a/ThirdParty/codemirror-5.52.0/package.json
+++ b/ThirdParty/codemirror-5.52.0/package.json
@@ -42,5 +42,8 @@
     "directories": {},
     "dependencies": {},
     "devDependencies": {}
+  },
+  "dependencies": {
+    "codemirror": "file:"
   }
 }
--- a/ThirdParty/dojo-release-1.10.4/dijit/package.json
+++ b/ThirdParty/dojo-release-1.10.4/dijit/package.json
@@ -1,26 +1,31 @@
 {
 	"name": "dijit",
-	"version":"1.10.4",
+	"version": "1.10.4",
 	"directories": {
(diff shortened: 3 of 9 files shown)


ayushd150:
I have signed the CLA in Google Forms

ayushd150:
Added myself in the Contributors.md file

ggetz:
Thanks @ayushd150!

First things first, could you let us know the reason for the change? It looks like this may be a fix for https://github.com/CesiumGS/cesium/issues/9393, is that correct?

If we do want to move forward with this PR–
- I can confirm we now have a CLA on file for you. 
- Could you please pull your change from https://github.com/CesiumGS/cesium/pull/12614 into this branch so we can review everything together?

jjspace (changes requested):
Thanks for the PR @ayushd150. I had a few comments about this approach, see below.

It also seems like you have a bunch of `package.json` and formatting changes on this branch. I'm assuming these just got added by mistake but please resolve/remove them to focus only on the code changes made.

Also you can include the updates to `CONTRIBUTORS.md` in this pr instead of a separate one.

jjspace:
I went ahead and pushed the changes from my comment above so we can try and get this merged. @lukemckinstry or @mzschwartz5 would you be able to take a quick look at this and merge if happy? You can check https://github.com/CesiumGS/cesium/issues/9393 for some context

mzschwartz5:
> I went ahead and pushed the changes from my comment above so we can try and get this merged. @lukemckinstry or @mzschwartz5 would you be able to take a quick look at this and merge if happy? You can check #9393 for some context

I'll take a look

Issue #12566, opened 2025-04-11
Rebuilding Sandcastle: Technical Implementation


jjspace:
This issue will track the _technical_ implementation of a new version of Sandcastle for CesiumJS. Design and features will primarily be worked out and discussed elsewhere. If you want to provide feedback on features you'd like to see please add them to our [post on the community forum](https://community.cesium.com/t/upgrading-sandcastle-we-need-your-input/39715/)

### High level plan

We plan to create a new workspace/package within our mono-repo to keep the new Sandcastle project separated but co-located with CesiumJS.

- `itwin-ui`: We expect to use the newest version of iTwin/Bentley UI library for both components and styling as much as possible. We may contribute improvements or components back to it as necessary.
- [**React**](https://react.dev/reference/react): This is a given if we want to make use of iTwinUI.
- [**Monaco**](https://microsoft.github.io/monaco-editor/): We'll want to balance robust feature set with usability for users who prefer a more minimal code editing experience. This should greatly increase the DX of the editor bringing familiarity for any already used to using VSCode
- **TypeScript**: Unless there is a reason not to. This will be isolated to it's own package so it's separate from the rest of CesiumJS
- **Build tools**: We'll likely get started with the [Vite plugin for react](https://react.dev/learn/build-a-react-app-from-scratch#vite), and if needed we can use esbuild like the CesiumJS library itself

CC @ggetz

jdfwarrior:
This will be a very welcomed update. Sandcastle has had the same look and feel for many years and could definitely use a refresh.

ggetz:
Some running TODO's which have come up in discussion like [the forum](https://community.cesium.com/t/upgrading-sandcastle-we-need-your-input/39715/) and [@jjspace's initial PR](https://github.com/CesiumGS/cesium/pull/12574).

[Moved to the top level post]

jjspace:
## Build structure

A big complication with this process is the actual build configuration. This is more complicated because locally we want to point at all the locally built files instead of copying them into the vite build statically. I wanted to share this breakdown I've been referencing a lot during my work on #12574. This is also not helped by the fact that when we deploy to prod we copy everything to the top level domain and basically all routes need to be adjusted somewhat to account for that. This will need to be kept in mind as we discuss how we want to change the import process into Sandcastles.

```
Local cesium server: npm start
    localhost:8080/Apps/Sandcastle/index.html
[...]

PR #12620, opened 2025-05-13
Add standalone page to Sandcastle v2

jjspace:
# Description

Adds the `standalone.html` functionality back to the new version of sandcastle. This builds off the code in #12619 now that I have a better understanding of how Vite works to bundle/build code and can import from different files. This further helps eliminate the `sandcastle-*.js` files.
There's one more that still has the `console` override logic but I think that will get wrapped into a module I'm considering making to encapsulate all the "cross window" logic. tbd on that.

## Issue number and link

Part of #12566 

## Testing plan

- run `npm run dev` in the package and make sure sandcastles work when clicking the new `Standalone` button
- run `npm run build` in the package then start the main server `npm start` and check that it works at that route too.

# Author checklist


packages/sandcastle/src/Helpers.ts
```diff
@@ -0,0 +1,81 @@
+import Pako from "pako";
+
+export function embedInSandcastleTemplate(code: string, addExtraLine: boolean) {
+  console.log("embedSandcastle");
```
ggetz, line 4: Remove extraneous logs?

packages/sandcastle/src/App.tsx
```diff
@@ -678,13 +609,36 @@ Sandcastle.addToolbarMenu(${variableName});`,
...
+      return;
+    }
+    const code = jsEditorRef.current.getValue();
+    const html = htmlEditorRef.current.getValue();
+
+    let baseHref = getBaseUrl();
+    const pos = baseHref.lastIndexOf("/");
+    baseHref = `${baseHref.substring(0, pos)}/gallery/`;
+
+    console.log([code, html]);
```
ggetz, line 631: Here as well.

packages/sandcastle/src/Helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+  code: string;
+  html: string;
+  baseHref?: string;
+};
+
+export function makeCompressedBase64String(data: SandcastleSaveData) {
+  // data stored in the hash as:
+  // Base64 encoded, raw DEFLATE compressed JSON array where index 0 is code, index 1 is html
+  const { code, html, baseHref } = data;
+  const encode = baseHref !== undefined ? [code, html, baseHref] : [code, html];
```
ggetz, line 32:
```suggestion
  const encode = baseHref ? [code, html, baseHref] : [code, html];
```

Since this is now typed, be don't need to worry about `null`.

packages/sandcastle/src/Helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+  jsonString = jsonString.slice(2, 2 + jsonString.length - 4);
+  const pakoString = Pako.deflate(jsonString, {
+    raw: true,
+    level: 9,
+  });
+  let base64String = btoa(
+    // TODO: not 100% sure why I have to do this conversion manually anymore but it works
+    // https://stackoverflow.com/questions/12710001/how-to-convert-uint8-array-to-base64-encoded-string
+    String.fromCharCode(...new Uint8Array(pakoString)),
+  );
```
ggetz, line 44: Looks like a `to: "string"` parameter was removed. Because pako returns a typed array., there shouldn't be a need to create a new  one.

```suggestion
  const typedArray: Uint8Array = Pako.deflate(jsonString, {
    raw: true,
    level: 9,
  });
  let base64String = btoa(
    String.fromCharCode(typedArray),
  );
```

ggetz (changes requested):
Thanks @jjspace, Standalone is working well! I have some cleanup comments.

packages/sandcastle/src/App.tsx
jjspace: I agree long term these need to be removed and may seem excessive but I've been using these logs to debug and keep track of how the app is loading things. They've been especially helpful as I shift or try different load processes with monaco and the gallery. I will do a cleanup pass at some point but think they're fine to stay for now.

ggetz:
Thanks @jjspace!

Issue #11013, opened 2023-01-04
Voxels broken when using orthographic camera


lilleyse:
![ezgif com-gif-maker](https://user-images.githubusercontent.com/915398/210621233-9be93a3c-8fe3-4adf-b82d-21f63e937f4f.gif)

jjhembd:
The calculation of the ray did not account for orthographic frustums at all.
I have a preliminary implementation for orthographic rendering in the `voxel-surfaces` branch. It seems to work except for the depth test. I need to look further at `IntersectDepth.glsl`, which may still be using assumptions based on a perspective camera.

Perspective rendering:

![Image](https://github.com/user-attachments/assets/1d98dfe3-26f6-4250-8601-21b5aec87dd2)

Orthographic camera, depth test off:

![Image](https://github.com/user-attachments/assets/ad4d6963-2cba-4876-9dc0-1f02fb3d7f03)

Orthographic camera with depth test on. (The globe should be inside the voxel)

![Image](https://github.com/user-attachments/assets/fdefbda4-35a9-4cca-a3d4-fd10739cb32e)

jjhembd:
The orthographic depth test issues were some bugs in my preliminary implementation. Here are fixed results, using a semi-transparent voxel.

Perspective:

![Image](https://github.com/user-attachments/assets/a98f0c2f-5c33-4d6e-b465-17cd9a3bebdb)

Orthographic:

![Image](https://github.com/user-attachments/assets/19fd4da9-5773-4457-b57a-6dd5c25b6972)

PR #12629, opened 2025-05-21
Fix voxel rendering with orthographic camera

jjhembd:
# Description

This PR corrects the calculation of the starting view ray in `VoxelFS.glsl` to account for orthographic cameras.

In the usual perspective camera, the view rays all originate from the camera position and emanate in a direction that depends on the screen pixel. The camera in this case is perhaps best understood as an eye, with the screen pixels corresponding to different angular directions from a focal point.

In an orthographic camera, all view rays are exactly parallel, with the different pixels corresponding to different ray origins.

Along with the new logic in VoxelFS, code changes include:
- New uniform `cameraDirectionUv` in `VoxelPrimitive`.
- Removed the (unused, undocumented) `viewDirWorld` property from the `fsInput.voxel` struct, to facilitate pulling the ray calculation out into a helper function.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/11013

## Testing plan

Load the "Voxels" Sandcastle, and use the projection picker to switch between perspective and orthographic cameras. The voxel should render correctly in both projections, with the globe partially visible through the semi-transparent voxels. The orthographic effect may be most obvious with the "Box" examples.

# Author checklist

### DIFF
--- a/Apps/Sandcastle/gallery/Voxels.html
+++ b/Apps/Sandcastle/gallery/Voxels.html
@@ -42,6 +42,8 @@
           geocoder: false,
           animation: false,
           timeline: false,
+          projectionPicker: true,
+          sceneModePicker: false,
         });
 
         viewer.extend(Cesium.viewerVoxelInspectorMixin);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 
 - Support Texture3D and add Volume Cloud SandBox Sample. [#12550](https://github.com/CesiumGS/cesium/issues/12550)
 
+#### Fixes :wrench:
+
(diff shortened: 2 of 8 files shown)


ggetz:
@jjhembd I'm seeing the following error when I run the e2e example:

image

That seems to happen for any voxel example, even before this PR. I'm unable to duplicate in Sandcastle. I'm thinking it may have to do with chrome headless.

ggetz:
Thanks @jjhembd. If everything is working on your machine with e2e, let's move forward. Everything is looking good otherwise, and I was having the same issue here as I was in `main`.

FWIW, the other tests are all working for me without tweaking the playwright config. It's specifically the voxel examples, including the existing sandcastles, which aren't compiling the shader correctly in headless when running `test-e2e`.

Issue #11820, opened 2024-02-08
Request render mode is ignored when a Billboard is added to the system


rropp5:
### What happened?

Adding a Billboard to the system triggers a render when the Scene is configured to use requestRenderMode.

This is a regression introduced by the fix for #11529 which was committed under #11535 so the fix to this issue should also resolve the original bug as well.

Billboard.js
```
Billboard.prototype._loadImage = function () {
   ...

   const scene = that._billboardCollection._scene;
   if (!defined(scene)) {
      return;
   }
   // Request a new render in request render mode
   scene.frameState.afterRender.push(() => true);

   ...
}
```

My initial thought is the solution should be implemented within the LabelCollection instead of Billboard because this is a problem specific to a Label's background.

### Reproduction steps

1. Enable requestRenderMode on the Scene
2. Set maximumRenderTimeChange to Infinity on the Scene
3. Periodically add Billboard entities to the Viewer.
4. Observe a render occurs each time a Billboard is added even when requestRender is not invoked.

[...]

PR #12495, opened 2025-02-25
Refactor `TextureAtlas` to avoid race conditions leading to billboard rendering artifacts

ggetz:
# PR Status

This PR is ready for review, but in draft state as there are a few remaining items before this should be considered for merging:

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [x] I have updated `CHANGES.md` with a short summary of my change
- [x] I have added or updated unit tests to ensure consistent code coverage
- [x] I have updated the inline documentation, and included code examples where relevant
- [x] I have fixed all remaining unit tests
- [x] Addressed flickering in clustering billboards
- [x] Addressed issue with vector tiles
- [x] Confirm WebGL 1
- [x] Ensure comparable performance
- [x] I have done a self-review on my code

# Description

The main motivation behind this PR was to **fix flickering of the globe to black that occurred during texture atlas updates, such as loading new labels or multiple billboard images,** as described in https://github.com/CesiumGS/cesium/issues/12231.

## What's causing the bug

The above issue was being caused by frequent updates to the texture atlas while the rest of the frame was being rendered. Basically, every time a new billboard image was loaded— whenever the image load promised happened resolve– there was a copy happening to the underlying `Texture` object. This includes whenever a new glyph was added to a label.

### Why it started happening

While this likely arose starting in https://github.com/CesiumGS/cesium/pull/10149 initially (due to a change in when exactly promises are resolved), I believe it was compounded by changes in some of the refactoring work from https://github.com/CesiumGS/cesium/pull/12100 and subsequent updates which changed lower-level texture copies. That being said, it was super easy for this issue to manifest because `TextureAtlas` did not have any mechanism in place to prevent texture updates from happening at unexpected times.

## What this PR aims to do

This PR does some refactoring of `TextureAtlas` and connected classes. It fixes the issue by batching together texture updates, but also hopefully makes it things a bit more atomic under-the-hood to facilitate future improvements and fixes.

### Future work

[...]

Issue #12138, opened 2024-08-20
Label Backgrounds do not appear when setting label `show` to `false` and then to `true` in specific cases.


anne-gropler:
### What happened?

I have an app with some labels that have a background. The labels can be hidden and shown again, and they can be updated (e.g., font size changed) while being hidden. When they are shown again, they show the updated value.
This basically works.

However, there are some cases where the background is not shown again after the label was hidden.

My local case is quite specific, so I separated it into 2 issues, demonstrated in a Sandcastle each. Since both issues are about the label background not coming back properly, I think the faulty behaviour might be related, so I am only writing this one issue here.

While I have a current workaround as stated below, I am not happy with it due to its performance, and I consider this behaviour of the label backgrounds to be buggy (while well understanding that this is a very specific case that is buggy).

### Sandcastle 1:

[...]

PR #12630, opened 2025-05-21
Fix requestRender condition in DataSourceDIsplay.update

### LINKED ISSUE #12543: Request Render Mode spotty with Entities
### What happened?

Originally [reported on the community forum](https://community.cesium.com/t/request-render-does-not-work-regression/39552)

> Here’s my use case: I have a geojson data source loaded (a polyline) , then I change the color or the thickness.
When I set my renderMode=false, everything happens as expected. But for performance reasons, I need to be in renderMode=true.
> 
> Changing the color is still functional, but changing the thickness is not.
> Of course, I’ve added a call to scene.requestRender().
> 
> I can display my entity.
> Then if I change the thickness, the entity is no longer displayed. I have to manually move the camera to display the path with the correct thickness.
> 
> I found [this PR #11934](https://github.com/CesiumGS/cesium/pull/11934) and this [Sandcastle](https://sandcastle.cesium.com/#c=rVVhb9owEP0rFprUVKImgYQCpdWmdqs6tWs1qn0ZlWriA6w5dmY7UFbx33dOgAKtJk2bEcLne/d8d747Uq2sIzMBczDklCiYk3Owosjot/IsGNbSUj7XyjGhwAxrdfI8VIRMpB5Bj4yZtFD3BwZ+FmDdV1AczI3mqHSmqHQZexJZkVW6e5HB+ZSpCSKu1Fgo4RZDtTw8GarKE2pTUEA5jIrJYKrnnwzLwN6BGUCqFUdHPTHC9wxShBlGx3Jxr4NnwgvDnNCqR8I64eiaUCt5FeM5Mw53TLXo2OjsAiYGwAZHUYgGrVadRElYrkNSejdUEhxxQoIFh7IziyoXqyN0jM2ZcBv+8qd1cV+py0uutPpgUbjiQdQMO83usWcmZCeQ3IhMODEDSxnnwYreA5ckZS6dkgCM0eawuh6TYrUEKvUkePzoFURqxoWarD3rkXfPpcXysWTxsaTl209Afx7cfkHXS6phzS1yGNZ6uLsppBPX+OgDZ5ALX75CpFob7nMJ1gO/+1O/NptXgl8+q7QdddtJ0jxuR0kSxfV9TBzTJEzCbqeJ32672doGPNT/mb8V0zA+7rSb3XY3bsWt6G/4u+FbfEnYSZrNbue43e7E/0gXJ/853j/ncy2Um4dtO6yQstyxcnyz6sIFwSE5PXspN0c4c2ygC5PC7uC4xIKyWl1s1EFV4PtWd9gMwnrjlzPqyzZYlSTOGZJKluX3+tLoQvFqnlStSF4zUTcFFWy7tfJ3t0EOrvEO4AcVy17j7cywlePLnQZ94a9685UbZX/h6CjnRplF/6nVa33rFhLO1ll/L7JcG0cKIwNKGw6yXPqWaoyK9AcOi9TatYs4YNhIwuZ53fRly9cx+jXCxgTTI1H+RDBgwcl8Khyc7COOUi0lyy0O4PVug1luqN02dY7B4hDokU7+tI/tN7ZD63MxI4KfvvHX4d/TWtSMCykH4hcOmrN+A/GvTFfz63YGRrKFh02js+vqkFLab6D4tqXTWo6Y2WP+DQ)
> 
> In the sandcastle, the requestRender does not work too with 1.127 (which is the version I’m using in my app)

We've also identified a similar reproduction in the [Rendering Performance Sandcastle](https://sandcastle.cesium.com/?src=Scene%20Rendering%20Performance.html). When the "mouseover picking" option is selected, the entity does not appear until the camera moves, or a new frame is rendered for an unrelated reason. 

Curiously enough, the second case is only reproducible at larger viewport sizes.

Likely, this is due to the entities becoming "ready" for rendering after the next immediate frame.

### Reproduction steps

_No response_

### Sandcastle example

_No response_

### Environment

Browser: All
CesiumJS Version: ID'd in 1.110 and 1.127
Operating System: All

Beilinson:
# Description

Attempt to fix the requestRenderMode issue which stems from #12429.

## Issue number and link

Fixes #12543

## Testing plan

Opened the sandbox and added the following code:
```js
  requestRenderMode: true,
  targetFrameRate: 30,
});

widget.scene.debugShowFramesPerSecond = true;
const entity = widget.entities.add({
  polyline: {
    positions: Cesium.Cartesian3.fromDegreesArray([-77, 35, -77.1, 35]),
    width: 5,
    material: Cesium.Color.RED,
  },
});
widget.zoomTo(entity);
setTimeout(() => {
  entity.polyline.material = Cesium.Color.WHITE;
}, 10_000);
setTimeout(() => {
  entity.polyline.material = Cesium.Color.BLUE;
}, 20_000);
```
I've noticed the bug is more likely with a large viewport and zooming out a bit on the entity.

The setTimeouts are so that the scene is definitely **NOT** requesting any render when updating the polyline material (all the tiles should already be loaded in).

With the above test the bug is easily reproducible, and is fixed with my branch.

# Author checklist

### DIFF
--- a/packages/engine/Source/DataSources/DataSourceDisplay.js
+++ b/packages/engine/Source/DataSources/DataSourceDisplay.js
@@ -322,9 +322,7 @@ DataSourceDisplay.prototype.update = function (time) {
     result = visualizers[x].update(time) && result;
   }
 
-  // Request a rendering of the scene when the data source
-  // becomes 'ready' for the first time
-  if (!this._ready && result) {
+  if (!result) {
     this._scene.requestRender();
   }
 


Beilinson:
I'm definitely not sure as to why this fix is the correct one, I expected all update functions to always return false if nothing changed and return true only if something did change, in which case I would have expected the condition to work with:

```js
if (result) {
  this._scene.requestRender();
}
```

However, it seems that result actually is `true` on basically every frame (even with no entities changed), while it actually is consistently `false` directly after updating an entity property. I probably misunderstood when update is supposed to return `false`, however as I said from my testing (and running unit tests) this PR seems to fix the issue consistently.

I don't want to revert the change from #12429 as to me that does seem like a logical change (if the dataSourceDisplay is ready it should stay ready regardless of entity updates), and I also don't think its correct to add an additional *dirty* flag, as to my understanding thats more or less what the `result` should be for the current frame.

Beilinson:
I'll point out that with this fix, requestRenderMode reverts to the previous behavior where renders are only requested after **all** datasources under the DataSourceDisplay finish asynchronous work. This can actually cause delays in rendering, if for exampling 1 datasource finishes after 2 frames, while another takes significantly more, a render will be requested only after the longer one finishes.

I could alter this behavior if desired

jasonmprop:
Thanks for working on this patch @Beilinson, much appreciated.

We also noticed this regression breaking our app (that uses request render) when updating from 1.121 to 1.128, I think it snuck in with 1.126 as noted in the `CHANGES.md`

We have confirmed that your first commit 9d2fe69 fixes our issue (we patched it on top of 1.128), we have not yet tested your commit bd389559

In terms of the second commit, is that what causes the delays in rendering you mentioned above ?

Let us know if there is something we can do to help out.

Beilinson:
Glad I could help @jasonmprop,

The second commit is a more technically correct fix, from what I can see more or less the `requestRender()` that we call manually whenever updating any property/adding/removing entities primes the next frame, but the actual update almost always happens 1 frame directly after that, which is the frame that the update method I patched calls.

Beilinson:
Theres actually an interesting alternative fix which can change how `requestRenderMode` works, by setting:

```js
if (this._prevResult !== result) {
  this._scene.requestRender();
}
```

Then a frame render will be requested twice when the async work begins and when it finishes, which will allow for things like updating an entity property, or adding an entity to a scene to be rendered automatically (without manually calling `requestRender()`), while still keeping renders to a minimum when the scene is unchanging.

While this is not consistent with the https://cesium.com/blog/2018/01/24/cesium-scene-rendering-performance/ blog post by @ggetz, I think this is worth considering:
1. These renders will only happen whenever **any** asynchronous work begins and whenever **all** async work ends. This still means if updating thousands of entities simultaneously, there should only be 2 frames rendered, which is even better than what some developers may be doing where they requestRender as part of a loop over thousands of entities
2. This could be some alternative mode in addition to requestRenderMode, or another flag used in tandem

packages/engine/Source/DataSources/BillboardVisualizer.js
```diff
@@ -226,6 +228,7 @@ BillboardVisualizer.prototype.update = function (time) {
       time,
       defaultSplitDirection,
     );
+    ready = billboard.ready && ready;
```
ggetz: I think this "undos" the fix in https://github.com/CesiumGS/cesium/pull/12429, as this function can now return 
`true` one frame, then `false` when the billboard is loading, then `true` again when it's completed.

Beilinson: This should be fine, update returns whether or not any processing is finished, while ready is stored separately on the datasourcedisplay and stays enabled after the first time

The naming is conflated internally in the update function here, I can change that

rropp5:
@ggetz I have some concerns that the changes here (to fix #12543) will introduce excessive rendering when the system creates new Billboard entities very often.

#11820 - This was the original bug report that was caused by some of the logic in the Billboard's _loadImage function.  Essentially, it _always_ requested a render when the Billboard's image loaded which would cause a render every time a new Billboard entity was created.

I believe the best tradeoff here would be to have the Billboard load its image from the TextureAtlas _synchronously_ if the image id is already in the TextureAtlas.  Otherwise, the image should be loaded asynchronously and then a render should be requested immediately after the image is resolved.  This allows users to reuse images in request render mode without introducing unnecessary render cycles while still fixing the problem with Label backgrounds failing to render after "missing" their target frame.

I had identified a workaround to #11820 that essentially implemented this behavior by passing a "redraw" parameter to the completeImageLoad function that was defined in Billboard.js:

```
Billboard.prototype._loadImage = function () {

  ...

  function completeImageLoad(index, redraw) {

    ...

    // Request a new render in request render mode
    if( redraw )
      scene.frameState.afterRender.push(() => true);
  }

  if (defined(image)) {
    imageIndexPromise = atlas.addImage(imageId, image);
  }
  if (defined(imageSubRegion)) {
    imageIndexPromise = atlas.addSubRegion(imageId, imageSubRegion);
  }

  this._imageIndexPromise = imageIndexPromise;

  if (!defined(imageIndexPromise)) {
    return;
  }

  // If the promise has already successfully resolved, we can return immediately without waiting a frame
  const index = atlas.getImageIndex(imageId);
  if (defined(index) && !defined(imageSubRegion)) {
    // Do not request a redraw because the image exists in the atlas
    completeImageLoad(index, false);
    return;
  }

  // Request a redraw the first time an image is loaded into the atlas
  imageIndexPromise.then((index) => completeImageLoad(index, true)).catch(function (error) {
    console.error(`Error loading image for billboard: ${error}`);
    that._imageIndexPromise = undefined;
  });
};
```

Would we be able to implement something similar with the new code structure?

packages/engine/Source/DataSources/DataSourceDisplay.js
```diff
@@ -322,11 +323,12 @@ DataSourceDisplay.prototype.update = function (time) {
...
 
-  // Request a rendering of the scene when the data source
-  // becomes 'ready' for the first time
-  if (!this._ready && result) {
+  // Request a rendering of the scene when all of the data sources
+  // switch from false to true
+  if (!this._prevResult && result) {
     this._scene.requestRender();
   }
+  this._prevResult = result;
```
rropp5: This code will effectively request a render each time a Billboard Entity is created because of the way the timing works with the new TextureAtlas implementation.

I believe an alternate solution is necessary since this is not consistent with the advertised behavior of requestRenderMode (read my 2nd comment on #11820 for more info).

I commented on the PR thread and suggested that it may be feasible to synchronously load the texture if the id already exists in the TextureAtlas.  Then, for images that are not in the atlas, asynchronously load the image and then request a render once it has loaded. This will effectively request a render once per distinct image and then for all subsequent uses of the image, invokers may simply request a render after creating the Billboard since the image will load synchronously.

rropp5: Another alternative to think about is the possibility of introducing a callback mechanism from the Billboard Entity API to allow invokers to pass some code to execute once the image is loaded.

To address #11529, the Label Entity's default implementation could pass a callback that requests a render.  Then, for Billboard Entities, invokers may choose to request a render (or not) when the Billboard finishes loading. This may be a good balance to preserve the advertised behavior of requestRenderMode while also addressing #11529 without introducing excessive render requests.

Beilinson: Great insight, I've been playing around with this idea locally and it seems to work well, will iron it out and commit soon

rropp5: Any luck with either of these paths?

packages/engine/Source/DataSources/BillboardVisualizer.js
Beilinson: Changed the naming, ready state on `DataSourceDisplay` is consistent with the fix in #12429

packages/engine/Source/DataSources/DataSourceDisplay.js
Beilinson: I believe the best behavior here should be consistent across entire primitive/entity api, not only for billboards/labels.

With the new commit, if you would like to disable rerenders that trigger internally when datasources complete updating (processing on web worker/awaiting textureatlas), then set:

```js
widget.dataSourceDisplay.renderOnUpdateComplete = false;
```

Beilinson:
Hey @ggetz, I've updated the PR with a fix for #12138 and the issues described in #12543 related to billboard and label rendering with `requestRenderMode=true`.

The issue stemmed from the difference between image urls/resources and client-side images (i.e, canvas).
- urls/Resources were fetched as part of the `resolveImage` function in `TextureAtlas.js`. At the end of the fetch, a callback added in `Scene.js` (line 709) to the `RequestScheduler` would request a render, which would later trigger the `textureAtlas.update` method call in `BillboardCollection.js` (line 1813)
- Client-side images (`HTMLImageElement|HTMLCanvasElement`) do not trigger this additional render, however are forced into the async pipeline as described in #12495

~My solution was to simply treat them as "auto-completing resources", and trigger `RequestScheduler.requestCompleteEvent` inside of `resolveImage` given that the image is not a resource. This solves the issues described above.~

The solution I think is actually optimal is to allow client-side images to be added **synchronously** to the image processing queue, allowing them to be processed at the end of the rendered frame without any additional renders. The entire process is still fully async as per the TextureAtlas promise refactor, as the internal `_addImage` still returns its promise which is resolved when `TextureAtlas.update` is ran.

Regarding @rropp5's comments, I added a flag `Scene.renderOnUpdateComplete` (enabled by default to keep current behavior), which is used in `DataSourceDisplay.update` and `BillboardCollection.update` to optionally prevent the renders that happen when update processing completes.

Sandboxes tested (make sure to run the test twice to prevent slow tiles from causing the additional renders)
1. Polyline Example [fix-requestrendermode](http://localhost:8080/Apps/Sandcastle/index.html#c=rVNdb9owFP0rVp6ClDmgCiFlAW2lTJvUalVh7cMyTSa+BauOza4dUFrx32cnpIKWSmxqnu7nuefeE+daGUs2gi/AkiFRsCFjMKIs6K2ADWCYBXntj7WyTCjALIieMkUIwp8SjL0BxQGvNIeEWCwhytS28zFTDSQ1OSigHOblYrrUmy/ICjDXgFPIteJuou9x5ZnKayagrLCVi+/6a1+AoYzz0M3N7ErLSjoiCaldHzCuxHUnLfUxQ+ssps7oPeriAhYIYD4jsir8+WEwiMhZPyLOoD1v/upEDZKbaZcJ6e/cgllAwWSyf5axlhqvdplr1CtAW4X7SXozuWgQt+0xns/xqHUx02Gzpc8YsDNRgC5tGHbIcNQs1eRpuyptmRwqdAKVu6/fZpOagT0Q5EC80BdsI9L/3e123yb1UhGEQq9hb5dTZgz+bcaz6kd0f0/lX2n//+qfX/6YtLDb5jc49Ti9bnOdIApSYysJI//Q/PdJFCuNlpQoQ0pjC8VKOhYmnpf5gwPNjfEYvjSN91tTLtZE8OGRZ0xyyYxxmftSyql4hCwYpbGrf9UqNeNCLb6vASWrfNmyN7psgpTSNHbu8U6rtZwzfIH8Fw) vs [main](https://sandcastle.cesium.com/#c=rVNdb9owFP0rVp6ClDmgCiFlAW2lTJvUalVh7cMyTSa+BauOza4dUFrx32cnpIKWSmxqnu7nuefeE+daGUs2gi/AkiFRsCFjMKIs6K2ADWCYBXntj7WyTCjALIieMkUIwp8SjL0BxQGvNIeEWCwhytS28zFTDSQ1OSigHOblYrrUmy/ICjDXgFPIteJuou9x5ZnKayagrLCVi+/6a1+AoYzz0M3N7ErLSjoiCaldHzCuxHUnLfUxQ+ssps7oPeriAhYIYD4jsir8+WEwiMhZPyLOoD1v/upEDZKbaZcJ6e/cgllAwWSyf5axlhqvdplr1CtAW4X7SXozuWgQt+0xns/xqHUx02Gzpc8YsDNRgC5tGHbIcNQs1eRpuyptmRwqdAKVu6/fZpOagT0Q5EC80BdsI9L/3e123yb1UhGEQq9hb5dTZgz+bcaz6kd0f0/lX2n//+qfX/6YtLDb5jc49Ti9bnOdIApSYysJI//Q/PdJFCuNlpQoQ0pjC8VKOhYmnpf5gwPNjfEYvjSN91tTLtZE8OGRZ0xyyYxxmftSyql4hCwYpbGrf9UqNeNCLb6vASWrfNmyN7psgpTSNHbu8U6rtZwzfIH8Fw) - My branch fixes the fact that the blue polyline doesn't render properly after the last requestRender. Removing that line in my branch sandbox also behaves properly by not rendering the polyline. In main it doesn't render regardless (this was the issue fixed by the `DataSourceDisplay.update` change)
2. Billboard with canvas image [fix-requestrendermode](http://localhost:8080/Apps/Sandcastle/index.html#c=rVRdb9owFP0rFnsgSMGhhbLyUbSNbtpDp1al+5KQKhPfBquOzWwHaCv++27ihGWMSXvYC8H3nnPv8b0nibWyjqwFbMCQC6JgQ6ZgRZbSL0UsmDfi4jzVyjGhwMwbrdFceQa1MSigHBZZMlvqzQfDUrA3YGYQa8WxoDMZHMIN/MjAultQHMwnzeGvuBxwrT6vOHMw1elKgquB40J7LJx4BmXfMfV4w8wj5ssqoBzmwFLGefAyV4SstMWIVsPqklNmHP5jqksfjE4vITEANmi/PqMn/X5v0A1Jd0AHnX7nrNtrhb6GUG5IXnZztcsnsRf9rHV6p4OD5jnEgrsTKejMBUGLXExIIeZQpYFUr+GtlEHO2eePzKwA7EJyet/pdLyG4y2iiEwN4PBILAU2aluB07ZPKiYLIeVCM8OJSFkCObqcJ1NrZnGKXMdZiiQaFyXeS8hPQdMDml5kyXFbJPgETcDlZoEtYk950+srtMyWbIXtcV2/qDaPfRXcLbHCSaczOkh9BJEs3T5X3QqVoGGrfdaUFIlviA9KOZuidrvWqEUicjo65HyvcZa+absuoWKVCi4N25AHHCJwYiB2TCXSX8ttaR6fuSeZm7X5qtsbnPNFc5TTFjLDZWipTR17iwWCUnpY6QlrksPflIyO2Wdv8v9jc0L8794nQ1JWJ94xw3LfIV5r3qA0KqI28i3vr3Si79HPRrInulLJvBFW9GIjQ9Lv7SN+3rXQrnju/vVFOK9ehEbYGNt88JOq9BuRrrRxJDMyQJEO8CuCbrbRIosfAc1tre9CyDiqU8dcrIngF0c+gPg2MWsx85DhnnM7NybjCPF/UKVmXKjk2s8hhy1PJlc+SCkdR3g8znRaywUzB5V/Ag) vs [main](https://sandcastle.cesium.com/#c=rVRdb9owFP0rFnsgSMGhhbLyUbSNbtpDp1al+5KQKhPfBquOzWwHaCv++27ihGWMSXvYC8H3nnPv8b0nibWyjqwFbMCQC6JgQ6ZgRZbSL0UsmDfi4jzVyjGhwMwbrdFceQa1MSigHBZZMlvqzQfDUrA3YGYQa8WxoDMZHMIN/MjAultQHMwnzeGvuBxwrT6vOHMw1elKgquB40J7LJx4BmXfMfV4w8wj5ssqoBzmwFLGefAyV4SstMWIVsPqklNmHP5jqksfjE4vITEANmi/PqMn/X5v0A1Jd0AHnX7nrNtrhb6GUG5IXnZztcsnsRf9rHV6p4OD5jnEgrsTKejMBUGLXExIIeZQpYFUr+GtlEHO2eePzKwA7EJyet/pdLyG4y2iiEwN4PBILAU2aluB07ZPKiYLIeVCM8OJSFkCObqcJ1NrZnGKXMdZiiQaFyXeS8hPQdMDml5kyXFbJPgETcDlZoEtYk950+srtMyWbIXtcV2/qDaPfRXcLbHCSaczOkh9BJEs3T5X3QqVoGGrfdaUFIlviA9KOZuidrvWqEUicjo65HyvcZa+absuoWKVCi4N25AHHCJwYiB2TCXSX8ttaR6fuSeZm7X5qtsbnPNFc5TTFjLDZWipTR17iwWCUnpY6QlrksPflIyO2Wdv8v9jc0L8794nQ1JWJ94xw3LfIV5r3qA0KqI28i3vr3Si79HPRrInulLJvBFW9GIjQ9Lv7SN+3rXQrnju/vVFOK9ehEbYGNt88JOq9BuRrrRxJDMyQJEO8CuCbrbRIosfAc1tre9CyDiqU8dcrIngF0c+gPg2MWsx85DhnnM7NybjCPF/UKVmXKjk2s8hhy1PJlc+SCkdR3g8znRaywUzB5V/Ag) - My branch fixes the fact that the blue square billboard doesn't render properly after the last requestRender (this was the issue fixed by the `TextureAtlas - resolveImage` change)
3. Label example [fix-requestrendermode](http://localhost:8080/Apps/Sandcastle/index.html#c=rZRtb9owEMe/ipVXQWImQNu1DNCAtlIlWlCh3ZtIyHEOsHDszHag7cR3n/NA2jG6Fmlvovjuf+d7+MlUCm3QmsEGFOogARs0AM2SCD9mNtd3aHYeSGEIE6B8p/LNF3kE1hQE4BCCZDFZys21IhHoMagJUClCm9CoBPblCn4moM09iBDUrQzhXV0qGImHOCQGBjKKOZg3YprVTplhLyB0n4jVmKiV9RdZQBjrA41JGLq/fIFQLLW1SNHaNTkgytg/Ipp4rmR0CQsFoN0vX09x/ezs5KJZRc0LfOGdeafNk0o1zcFJALyFsnwIGXgyLeQ7JKC+U81tczuq1FY/iQ2KpJA6JhRKtzbPHMoKhmm6SWrC1zfD4ax3dzkbPUyHN3dXhV4mhtvB/2ChWbZQo7CuQRlGCR8ptmCvDT3+YcbT0bjQx+wJ+Gg+12Bre7PmcgIN17PNNiq7Ku06+4SuFkomImxlMy9cQWkeSC7V6zDTE+4PH67whpllj8dL4nr4/DTPubXfbcpOueYXKaOpdPfWlUpslVMWgW3ddSuo083Hvb9XBZFcQ49zN40p/QcoywTbKmrMPM/La/jcFSU6/weefXxeARoUGKOUY5SCXALzIVHHM/UeVcdzdTxZ/2TrI7rury4PwpXjZb+fJeF8R4JTddrZ+Lq7Cr6zKJbKoERxF+OaAfvw2AdI14KErsBgqnV+C0Lt2tvQdsjWiIWdA28mopxobT3zhPOJ3bTvdNs1q/8rlEsSMrEY2UVw8pzKlvXuMDdijNs1ezwcaaTkAVF7mX8D) vs [main](https://sandcastle.cesium.com/#c=rZRtb9owEMe/ipVXQWImQNu1DNCAtlIlWlCh3ZtIyHEOsHDszHag7cR3n/NA2jG6Fmlvovjuf+d7+MlUCm3QmsEGFOogARs0AM2SCD9mNtd3aHYeSGEIE6B8p/LNF3kE1hQE4BCCZDFZys21IhHoMagJUClCm9CoBPblCn4moM09iBDUrQzhXV0qGImHOCQGBjKKOZg3YprVTplhLyB0n4jVmKiV9RdZQBjrA41JGLq/fIFQLLW1SNHaNTkgytg/Ipp4rmR0CQsFoN0vX09x/ezs5KJZRc0LfOGdeafNk0o1zcFJALyFsnwIGXgyLeQ7JKC+U81tczuq1FY/iQ2KpJA6JhRKtzbPHMoKhmm6SWrC1zfD4ax3dzkbPUyHN3dXhV4mhtvB/2ChWbZQo7CuQRlGCR8ptmCvDT3+YcbT0bjQx+wJ+Gg+12Bre7PmcgIN17PNNiq7Ku06+4SuFkomImxlMy9cQWkeSC7V6zDTE+4PH67whpllj8dL4nr4/DTPubXfbcpOueYXKaOpdPfWlUpslVMWgW3ddSuo083Hvb9XBZFcQ49zN40p/QcoywTbKmrMPM/La/jcFSU6/weefXxeARoUGKOUY5SCXALzIVHHM/UeVcdzdTxZ/2TrI7rury4PwpXjZb+fJeF8R4JTddrZ+Lq7Cr6zKJbKoERxF+OaAfvw2AdI14KErsBgqnV+C0Lt2tvQdsjWiIWdA28mopxobT3zhPOJ3bTvdNs1q/8rlEsSMrEY2UVw8pzKlvXuMDdijNs1ezwcaaTkAVF7mX8D) - My branch fixes the fact that the label with red background doesn't render properly after the last requestRender (only resolved glyphs render, the letter a), (this was the issue fixed by the `TextureAtlas - resolveImage` change)

Beilinson:
@ggetz An alternative solution I considered was moving the `TextureAtlas.update` method to `scene._preUpdate.addEventListener(() => textureAtlas.update(frameState.context))`, rather than as part of the `BillboardCollection.update` method.

My reasoning stems from the fact that processing the texture atlas according to your explanations in #12495:
> - Added an update function which is designed to be called once per frame which processes all resource updates at once
>    - This means we can perform all of the needed texture copies before or after the frame is being rendered

However in `requestRenderMode=true`, the textureAtlas is only called after rendering a frame, rather than on every actual frame, unlike `DataSourceDisplay.update` which runs every frame regardless of rendering.

Is this a valid solution or have I misunderstood the PR above?

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -517,6 +517,9 @@ TextureAtlas.prototype._addImage = function (index, image) {
   Check.typeOf.number.greaterThanOrEquals("index", index, 0);
   Check.defined("image", image);
   //>>includeEnd('debug');
+  if (this.isDestroyed() || !defined(image)) {
```
ggetz: It shouldn't be possible to call `_addImage` on a destroyed `TextureAtlas` due to  the behavior of `destroyObject`. 

But that is a "should," so let me know if you see evidence to the contrary.

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -623,7 +626,7 @@ TextureAtlas.prototype.update = function (context) {
   return this._processImageQueue(context);
 };
 
-async function resolveImage(image, id) {
+function resolveImage(image, id) {
```
ggetz: This function is no longer "resolving" the image if it is a promise. For clarity, I'd suggest something like `getImage` or `obtainImage` instead.

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -660,21 +663,16 @@ TextureAtlas.prototype.addImage = function (id, image) {
 
   const index = this._nextIndex++;
   this._indexById.set(id, index);
+  image = resolveImage(image, id);
+
+  if (image instanceof Promise) {
+    // Return promise chain, the image will resolve and be processed on a later frame
+    promise = image.then((image) => this._addImage(index, image));
```
ggetz: Regarding my comment above where I said it shouldn't be possible to call `_addImage`—

If `image` resolves and `this` is destroyed, `_addImage` will now throw an error.

I have a suggestion for streamlining this function accounting for that point.

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -660,21 +663,16 @@ TextureAtlas.prototype.addImage = function (id, image) {
...
+  } else {
+    // Add the image synchronously if possible, the image will be processed on the current frame
+    promise = this._addImage(index, image);
+  }
```
ggetz:
```suggestion
  const getResolvedImage = async (index, image)  => {
    // The initial part of an async function runs synchronously up to the first await
    // If the image is synchronous, skip the await, and the image will be processed on the current frame
    if (image instanceof Promise) {
       // Effectively return a promise chain. The image promise will resolve and be processed on a later frame
      image = await image;
      if (this.isDestroyed()) {
        return -1;
      }
    }

    return this._addImage(index, image);
  }

  promise = getResolvedImage(index, image);
```

packages/engine/Source/Scene/BillboardCollection.js
```diff
@@ -1810,7 +1810,11 @@ BillboardCollection.prototype.update = function (frameState) {
       return;
     }
 
-    return textureAtlas.update(frameState.context);
+    let isUpdated = textureAtlas.update(frameState.context);
+    if (defined(this._scene)) {
+      isUpdated = isUpdated && this._scene.renderOnUpdateComplete;
+    }
+    return isUpdated;
   });
```
ggetz: I'm not entirely sure about `renderOnUpdateComplete`, or at least that it is specified on the `Scene`. It's not great architecture that `BillboardCollection` needs a reference to the scene as it is. I would like to avoid depending on that architecture further if we can.

Is there a sufficient improvement with only the updates to `TextureAtlas`? It may not fix https://github.com/CesiumGS/cesium/issues/12543 altogether, but it should fix https://github.com/CesiumGS/cesium/issues/11820.

ggetz (changes requested):
Thanks for the update here @Beilinson! I think this is getting much closer. 

I have a few comments. Let me know what you think.

packages/engine/Source/Scene/BillboardCollection.js
Beilinson: Yes this was added as POC solution to #11820, it has no effect on #12543. #12543 is solved by the `requestRender` on `DataSourceDisplay` + the updates on `TextureAtlas`.

Would you like some better solution to #11820 here, or split that out into a separate PR. Either way, what would be a better architecture?

packages/engine/Source/Renderer/TextureAtlas.js
Beilinson: It was always returning a promise which was needed to await further, agree about the naming. `getImage` sounds good to me.

packages/engine/Source/Scene/BillboardCollection.js
Beilinson: `DataSourceDisplay.onUpdateComplete` was also added mostly as a mechanism to hook in for higher control with #11820, so it can be safely removed as well if this solution is unneeded.

ggetz: > Would you like some better solution to https://github.com/CesiumGS/cesium/issues/11820 here, or split that out into a separate PR.

I'm hoping the fix in this PR to texture atlas is sufficient for fixing https://github.com/CesiumGS/cesium/issues/11820 as is. 🤞 

Breaking out targeted fixes into their own PRs would help isolate testing and prevent one fix from holding up the other. That might also help us better coordinate with changes in https://github.com/CesiumGS/cesium/pull/12841 and if one PR should supersede the other. 

But I understand and appreciate that you've already put a good deal of work into this PR. So, if you can a separate PR would be appreciated, but is not required.

ggetz: > Either way, what would be a better architecture?

If it is needed, lets discuss where it should live. I'd be open to reflecting a primitive value from `Scene` onto `FrameState`.

Beilinson: I've taken it out for now. To be completely honest I am not sure that this kind of flag is 100% needed, if so then I guess its best for the conversation to continue with @rropp5 on #11820 after this PR is merged

rropp5: > > Would you like some better solution to #11820 here, or split that out into a separate PR.
> 
> I'm hoping the fix in this PR to texture atlas is sufficient for fixing #11820 as is. 🤞
> 
> **Breaking out targeted fixes into their own PRs would help isolate testing and prevent one fix from holding up the other. That might also help us better coordinate with changes in #12841 and if one PR should supersede the other.**
> 

I very much agree with what @ggetz says here from a process perspective. Breaking work down into smaller pieces is generally better.

I also agree that it's best to avoid introducing flags that change behavior in such core functionality of the library.  However, I'd rather see a holistic solution that works in all cases and, more specifically, fixes #11820 off-the-shelf.  It sounds like there may be a question as to whether #11820 actually needs to be fixed so I'll just link to [my comment](https://github.com/CesiumGS/cesium/issues/11820#issuecomment-1936015367) where I explain why it should be fixed.

Ultimately, I can accept if the path is to proceed with this issue without the requirement to satisfy #11820. But, I still feel it is a regression that must be addressed.

@ggetz, I'm curious to hear your thoughts on how easy it would be to incorporate [this asynchronous/synchronous behavior](https://github.com/CesiumGS/cesium/pull/12630#issuecomment-2977199510) into the new TextureAtlas code if the changes in this PR aren't sufficient to also fix #11820:

> I believe the best tradeoff here would be to have the Billboard load its image from the TextureAtlas synchronously if the image id is already in the TextureAtlas. Otherwise, the image should be loaded asynchronously and then a render should be requested immediately after the image is resolved. This allows users to reuse images in request render mode without introducing unnecessary render cycles while still fixing the problem with Label backgrounds failing to render after "missing" their target frame.

Beilinson: Ok, so sounds like this is ready and any further discussion will continue on #11820 if needed.

Beilinson:
Hey @ggetz, I compared against main, here is the updated status of the sandbox examples in this PR:

1. Polyline Example [fix-requestrendermode](http://localhost:8080/Apps/Sandcastle/index.html#c=rVNdb9owFP0rVp6ClDmgCiFlAW2lTJvUalVh7cMyTSa+BauOza4dUFrx32cnpIKWSmxqnu7nuefeE+daGUs2gi/AkiFRsCFjMKIs6K2ADWCYBXntj7WyTCjALIieMkUIwp8SjL0BxQGvNIeEWCwhytS28zFTDSQ1OSigHOblYrrUmy/ICjDXgFPIteJuou9x5ZnKayagrLCVi+/6a1+AoYzz0M3N7ErLSjoiCaldHzCuxHUnLfUxQ+ssps7oPeriAhYIYD4jsir8+WEwiMhZPyLOoD1v/upEDZKbaZcJ6e/cgllAwWSyf5axlhqvdplr1CtAW4X7SXozuWgQt+0xns/xqHUx02Gzpc8YsDNRgC5tGHbIcNQs1eRpuyptmRwqdAKVu6/fZpOagT0Q5EC80BdsI9L/3e123yb1UhGEQq9hb5dTZgz+bcaz6kd0f0/lX2n//+qfX/6YtLDb5jc49Ti9bnOdIApSYysJI//Q/PdJFCuNlpQoQ0pjC8VKOhYmnpf5gwPNjfEYvjSN91tTLtZE8OGRZ0xyyYxxmftSyql4hCwYpbGrf9UqNeNCLb6vASWrfNmyN7psgpTSNHbu8U6rtZwzfIH8Fw) vs [main](https://sandcastle.cesium.com/#c=rVNdb9owFP0rVp6ClDmgCiFlAW2lTJvUalVh7cMyTSa+BauOza4dUFrx32cnpIKWSmxqnu7nuefeE+daGUs2gi/AkiFRsCFjMKIs6K2ADWCYBXntj7WyTCjALIieMkUIwp8SjL0BxQGvNIeEWCwhytS28zFTDSQ1OSigHOblYrrUmy/ICjDXgFPIteJuou9x5ZnKayagrLCVi+/6a1+AoYzz0M3N7ErLSjoiCaldHzCuxHUnLfUxQ+ssps7oPeriAhYIYD4jsir8+WEwiMhZPyLOoD1v/upEDZKbaZcJ6e/cgllAwWSyf5axlhqvdplr1CtAW4X7SXozuWgQt+0xns/xqHUx02Gzpc8YsDNRgC5tGHbIcNQs1eRpuyptmRwqdAKVu6/fZpOagT0Q5EC80BdsI9L/3e123yb1UhGEQq9hb5dTZgz+bcaz6kd0f0/lX2n//+qfX/6YtLDb5jc49Ti9bnOdIApSYysJI//Q/PdJFCuNlpQoQ0pjC8VKOhYmnpf5gwPNjfEYvjSN91tTLtZE8OGRZ0xyyYxxmftSyql4hCwYpbGrf9UqNeNCLb6vASWrfNmyN7psgpTSNHbu8U6rtZwzfIH8Fw) - Both this branch and `main` work correctly (fixed by #12841)
2. Billboard with canvas image [fix-requestrendermode](http://localhost:8080/Apps/Sandcastle/index.html#c=rVRdb9owFP0rFnsgSMGhhbLyUbSNbtpDp1al+5KQKhPfBquOzWwHaCv++27ihGWMSXvYC8H3nnPv8b0nibWyjqwFbMCQC6JgQ6ZgRZbSL0UsmDfi4jzVyjGhwMwbrdFceQa1MSigHBZZMlvqzQfDUrA3YGYQa8WxoDMZHMIN/MjAultQHMwnzeGvuBxwrT6vOHMw1elKgquB40J7LJx4BmXfMfV4w8wj5ssqoBzmwFLGefAyV4SstMWIVsPqklNmHP5jqksfjE4vITEANmi/PqMn/X5v0A1Jd0AHnX7nrNtrhb6GUG5IXnZztcsnsRf9rHV6p4OD5jnEgrsTKejMBUGLXExIIeZQpYFUr+GtlEHO2eePzKwA7EJyet/pdLyG4y2iiEwN4PBILAU2aluB07ZPKiYLIeVCM8OJSFkCObqcJ1NrZnGKXMdZiiQaFyXeS8hPQdMDml5kyXFbJPgETcDlZoEtYk950+srtMyWbIXtcV2/qDaPfRXcLbHCSaczOkh9BJEs3T5X3QqVoGGrfdaUFIlviA9KOZuidrvWqEUicjo65HyvcZa+absuoWKVCi4N25AHHCJwYiB2TCXSX8ttaR6fuSeZm7X5qtsbnPNFc5TTFjLDZWipTR17iwWCUnpY6QlrksPflIyO2Wdv8v9jc0L8794nQ1JWJ94xw3LfIV5r3qA0KqI28i3vr3Si79HPRrInulLJvBFW9GIjQ9Lv7SN+3rXQrnju/vVFOK9ehEbYGNt88JOq9BuRrrRxJDMyQJEO8CuCbrbRIosfAc1tre9CyDiqU8dcrIngF0c+gPg2MWsx85DhnnM7NybjCPF/UKVmXKjk2s8hhy1PJlc+SCkdR3g8znRaywUzB5V/Ag) vs [main](https://sandcastle.cesium.com/#c=rVRdb9owFP0rFnsgSMGhhbLyUbSNbtpDp1al+5KQKhPfBquOzWwHaCv++27ihGWMSXvYC8H3nnPv8b0nibWyjqwFbMCQC6JgQ6ZgRZbSL0UsmDfi4jzVyjGhwMwbrdFceQa1MSigHBZZMlvqzQfDUrA3YGYQa8WxoDMZHMIN/MjAultQHMwnzeGvuBxwrT6vOHMw1elKgquB40J7LJx4BmXfMfV4w8wj5ssqoBzmwFLGefAyV4SstMWIVsPqklNmHP5jqksfjE4vITEANmi/PqMn/X5v0A1Jd0AHnX7nrNtrhb6GUG5IXnZztcsnsRf9rHV6p4OD5jnEgrsTKejMBUGLXExIIeZQpYFUr+GtlEHO2eePzKwA7EJyet/pdLyG4y2iiEwN4PBILAU2aluB07ZPKiYLIeVCM8OJSFkCObqcJ1NrZnGKXMdZiiQaFyXeS8hPQdMDml5kyXFbJPgETcDlZoEtYk950+srtMyWbIXtcV2/qDaPfRXcLbHCSaczOkh9BJEs3T5X3QqVoGGrfdaUFIlviA9KOZuidrvWqEUicjo65HyvcZa+absuoWKVCi4N25AHHCJwYiB2TCXSX8ttaR6fuSeZm7X5qtsbnPNFc5TTFjLDZWipTR17iwWCUnpY6QlrksPflIyO2Wdv8v9jc0L8794nQ1JWJ94xw3LfIV5r3qA0KqI28i3vr3Si79HPRrInulLJvBFW9GIjQ9Lv7SN+3rXQrnju/vVFOK9ehEbYGNt88JOq9BuRrrRxJDMyQJEO8CuCbrbRIosfAc1tre9CyDiqU8dcrIngF0c+gPg2MWsx85DhnnM7NybjCPF/UKVmXKjk2s8hhy1PJlc+SCkdR3g8znRaywUzB5V/Ag) - My branch fixes the fact that the blue square billboard doesn't render properly after the last requestRender (this was the issue fixed by the TextureAtlas - resolveImage change)
3. Label example [fix-requestrendermode](http://localhost:8080/Apps/Sandcastle/index.html#c=rZRtb9owEMe/ipVXQWImQNu1DNCAtlIlWlCh3ZtIyHEOsHDszHag7cR3n/NA2jG6Fmlvovjuf+d7+MlUCm3QmsEGFOogARs0AM2SCD9mNtd3aHYeSGEIE6B8p/LNF3kE1hQE4BCCZDFZys21IhHoMagJUClCm9CoBPblCn4moM09iBDUrQzhXV0qGImHOCQGBjKKOZg3YprVTplhLyB0n4jVmKiV9RdZQBjrA41JGLq/fIFQLLW1SNHaNTkgytg/Ipp4rmR0CQsFoN0vX09x/ezs5KJZRc0LfOGdeafNk0o1zcFJALyFsnwIGXgyLeQ7JKC+U81tczuq1FY/iQ2KpJA6JhRKtzbPHMoKhmm6SWrC1zfD4ax3dzkbPUyHN3dXhV4mhtvB/2ChWbZQo7CuQRlGCR8ptmCvDT3+YcbT0bjQx+wJ+Gg+12Bre7PmcgIN17PNNiq7Ku06+4SuFkomImxlMy9cQWkeSC7V6zDTE+4PH67whpllj8dL4nr4/DTPubXfbcpOueYXKaOpdPfWlUpslVMWgW3ddSuo083Hvb9XBZFcQ49zN40p/QcoywTbKmrMPM/La/jcFSU6/weefXxeARoUGKOUY5SCXALzIVHHM/UeVcdzdTxZ/2TrI7rury4PwpXjZb+fJeF8R4JTddrZ+Lq7Cr6zKJbKoERxF+OaAfvw2AdI14KErsBgqnV+C0Lt2tvQdsjWiIWdA28mopxobT3zhPOJ3bTvdNs1q/8rlEsSMrEY2UVw8pzKlvXuMDdijNs1ezwcaaTkAVF7mX8D) vs [main](https://sandcastle.cesium.com/#c=rZRtb9owEMe/ipVXQWImQNu1DNCAtlIlWlCh3ZtIyHEOsHDszHag7cR3n/NA2jG6Fmlvovjuf+d7+MlUCm3QmsEGFOogARs0AM2SCD9mNtd3aHYeSGEIE6B8p/LNF3kE1hQE4BCCZDFZys21IhHoMagJUClCm9CoBPblCn4moM09iBDUrQzhXV0qGImHOCQGBjKKOZg3YprVTplhLyB0n4jVmKiV9RdZQBjrA41JGLq/fIFQLLW1SNHaNTkgytg/Ipp4rmR0CQsFoN0vX09x/ezs5KJZRc0LfOGdeafNk0o1zcFJALyFsnwIGXgyLeQ7JKC+U81tczuq1FY/iQ2KpJA6JhRKtzbPHMoKhmm6SWrC1zfD4ax3dzkbPUyHN3dXhV4mhtvB/2ChWbZQo7CuQRlGCR8ptmCvDT3+YcbT0bjQx+wJ+Gg+12Bre7PmcgIN17PNNiq7Ku06+4SuFkomImxlMy9cQWkeSC7V6zDTE+4PH67whpllj8dL4nr4/DTPubXfbcpOueYXKaOpdPfWlUpslVMWgW3ddSuo083Hvb9XBZFcQ49zN40p/QcoywTbKmrMPM/La/jcFSU6/weefXxeARoUGKOUY5SCXALzIVHHM/UeVcdzdTxZ/2TrI7rury4PwpXjZb+fJeF8R4JTddrZ+Lq7Cr6zKJbKoERxF+OaAfvw2AdI14KErsBgqnV+C0Lt2tvQdsjWiIWdA28mopxobT3zhPOJ3bTvdNs1q/8rlEsSMrEY2UVw8pzKlvXuMDdijNs1ezwcaaTkAVF7mX8D) - My branch fixes the fact that the label with red background doesn't render properly after the last requestRender (only resolved glyphs render, the letter a), (this was the issue fixed by the TextureAtlas - resolveImage change)

I updated the `CHANGES.md` to reflect that this PR now primarily fixes primitives with syncronous images

Beilinson:
Bumping this @ggetz, this is still a blocking issue for us even after #12841. Could this maybe get in before 1.135? Thanks!

javagl:
Kudos for diligently updating your PRs (even though reviewing and merging them takes longer than it should). The last merge of `main` seems to have caused a test failure. It's hard to quickly zoom into the changes here and other changes from main - do you have an idea whether this might be a hint that some _other_ PR introduced a breaking change that was _not_ discovered until now?

Beilinson:
Thanks @javagl! I think its a merge mistake it should work now

Issue #12423, opened 2025-01-09
Knockout not exported from Cesium.d.ts


ijessen-mitll:
### What happened?

`knockout` is exported from `Cesium.js`, but there is no matching export within `Cesium.d.ts`.  Therefore there is no obvious type-safe way to access knockout functionality, such as subscribing to observables within view models.  

The work around is to disable typescript checking of the relevant line:

```
// @ts-ignore 
import { knockout as ko } from 'cesium';
```

TS type definitions exist for both knockout-es5 and knockout:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/knockout.es5/index.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/knockout/index.d.ts

### Environment

CesiumJS Version: 1.123.1

ggetz:
Thanks for the report @ijessen-mitll! This would need to be explicitly added to our generation process for TS definitions.

PR #12631, opened 2025-05-22
Sandcastle - Nested gallery structure

jjspace:
# Description

Major changes/focus for this PR:

- **All sandcastle JS code is now treated as a single JS `module`** that should import `Cesium` and `Sandcastle` directly
    - if the `import` statements are not in the sandcastle code they will be added before loaded into the `bucket`. AFAICT this should account for all existing sandcastles that relied on the global objects
    - This also means these objects are not in global `window` scope anymore
    - Most of the "boilerplate" that previously existed was to set up the `window.startup` function specifically to account for modules loading out of order. Switching to `type=module` eliminates that
    - The types importing for Monaco was updated so that it should support intellisense for `Cesium` and `Sandcastle` regardless of the imports
- **New structure for the gallery** which splits the code into dedicated files and relies on a new "metadata sidecar yaml" file for info about a sandcastle
    - `index.html` is the html/css code for a sandcastle, `main.js` is the JS code for a sandcastle, `sandcastle.yaml` holds the metadata and is what dictates that a subfolder is even a gallery example in the first place
    - Note that `index.html` does _not_ contain the full page structure with `head` and `body` tags. This means you **can no longer load gallery items directly** with a path like `localhost:8080/Apps/Sandcastle/gallery/3D Models.html`.
        - <sup>we feel this was not utilized much and restricted the format too much, maybe we can revisit a way to load them outside of the full SC app in the future</sup>
    - a new script called `buildGallery.js` has been created to parse through the new gallery structure and generate a top level `list.json` file. This also does a little basic validation. Where this script lives and exactly what it does is still a little up in the air

TODO:

- [ ] Clean up the gallery build script if we're satisfied with the organization structure
    - [ ] create a "conversion script" to transform all our existing gallery examples
- [ ] Fix loading from a url id
    - Currently the app does not work in a very "react style" and uses a lot of imperative code which is massively conflicting with load order. This is my next focus as a dedicated restructuring pass. after _that_ I think loading will be a lot easier to implement
- [ ] Decide on gallery location
    - I _think_ the gallery could live outside of the sandcastle project if we wanted. Given [recent comments](https://community.cesium.com/t/upgrading-sandcastle-we-need-your-input/39715/15) around the desire to be able to point at any gallery directory I built the logic to hopefully be able to point to any directory from the client. More thought still required but it's in my mind.

## Issue number and link

Part of #12566 

## Testing plan

- run `node buildGallery.js` to generate the list file
    - [ ] need to figure out how to incorporate this into the build nicer
- Load sandcastle both with `npm run dev` and from the normal server `npm start` after `npm run build`
- Make sure the gallery still loads and clicking them loads the examples
    - [ ] may want to consider better where these files live for access from both locations and when deployed
- Make sure types in monaco still show up, with and without the import statements
- Copy in the sandcastle code from another sandcastle and make sure it still works

# Author checklist


jjspace:
@ggetz I tried to split the commits a little between the new gallery files and actual code changes if that makes it easier to review

packages/sandcastle/buildGallery.js
```diff
@@ -0,0 +1,111 @@
...
+// built during build
+// May also want to move this out of the sandcastle project so we don't have to duplicate
+// files for the local build, need to think about deploying too...
+
+const galleryDirectory = "./public/gallery";
+
+const yamlFiles = globbySync([
+  `${galleryDirectory}/*/sandcastle.yaml`,
+  `!${galleryDirectory}/list.json`,
+]);
```
jjspace, line 17: Going with the `globby` approach here instead of just listing the directories lets us "anchor" the existence of a Gallery example around a `sandcastle.yaml` file.

packages/sandcastle/public/gallery/billboards/sandcastle.yaml
```diff
@@ -0,0 +1,10 @@
+id: billboards
+legacy-id: Billboards.html
+title: Billboards
+thumbnail: ./Billboards.jpg
```
jjspace, line 4: Adding the `thumbnail` property here lets it be named anything. We had discussed maybe not including this and consistently requiring a `thumbnail.png` or something the same as I am for `index.html` and `main.js`. Not sure which approach is better here.
The value is optional regardless if we check for a file or it's defined here.

packages/sandcastle/src/App.tsx
```diff
@@ -639,9 +677,19 @@ Sandcastle.addToolbarMenu(${variableName});`,
     window.focus();
   }
 
-  function loadDemo(demo: GalleryDemo) {
-    // do stuff
-    setCode(demo.js ?? defaultJsCode, demo.html ?? defaultHtmlCode);
+  const GALLERY_BASE = "/gallery";
```
jjspace, line 680: I _think_ that doing something like this would allow us to point it at literally any path as long as it conforms to the structure we'd expect. Namely that there's a `list.json` at the root and `[slug]/(index.html|main.js)` files. This could help us swap out where it points for different local/deployed builds or even potentially a different server entirely if someone wants to host their own gallery?

jjspace:
@ggetz I know there's still some scattered commented code and console logs. I'll do a cleanup pass on those before this gets merged but wanted to just validate the approach first.

packages/sandcastle/standalone.html
```diff
@@ -8,6 +8,14 @@
...
     <title>Cesium Demo</title>
+    <script type="importmap">
+      {
+        "imports": {
+          "cesium": "__CESIUM_BASE_URL__/index.js",
+          "Sandcastle": "./templates/Sandcastle.js"
+        }
+      }
+    </script>
     <script src="__CESIUM_BASE_URL__/Cesium.js" vite-ignore></script>
```
ggetz: Is this still needed after adding the importmap?

packages/sandcastle/templates/bucket.html
```diff
@@ -21,24 +29,23 @@
         }
       }
     </script>
-    <!-- <script type="module" src="./Sandcastle.js"></script> -->
     <script src="__CESIUM_BASE_URL__/Cesium.js" vite-ignore></script>
     <script type="module">
```
ggetz: Do we need to set `window.CESIUM_BASE_URL` here as well?

packages/sandcastle/templates/tsconfig.json
```diff
@@ -3,9 +3,10 @@
   "include": ["./Sandcastle.ts"],
   "compilerOptions": {
     "declaration": true,
-    "emitDeclarationOnly": true,
+    "emitDeclarationOnly": false,
+    "allowImportingTsExtensions": false,
     "noEmit": false,
     "isolatedModules": false,
```
ggetz: This is all default behavior, and will not use a lot of the other fields specified in `tsconfig.app.json`. I'd suggest not inheriting and just starting from a new config file. It'll also help make the "lib" portion of Sandcastle less coupled to the "app" portion as we discussed offline.

packages/sandcastle/templates/Sandcastle.ts
```diff
@@ -45,7 +45,7 @@ const Sandcastle = {
       let needle = `${bucket}:`; // Firefox
       let pos = stack.indexOf(needle);
       if (pos < 0) {
-        needle = " (<anonymous>:"; // Chrome
+        needle = "<anonymous>:"; // Chrome
```
ggetz, line 48: Curious, what's this for?

packages/sandcastle/buildGallery.js
```diff
@@ -0,0 +1,111 @@
+import { readFileSync, existsSync, writeFileSync } from "fs";
+import { parse } from "yaml";
+import { globbySync } from "globby";
+import { basename, dirname, join } from "path";
+import { exit } from "process";
+
+// TODO: probably need to find a way to integrate this into the vite process to make sure it's
+// built during build
+// May also want to move this out of the sandcastle project so we don't have to duplicate
+// files for the local build, need to think about deploying too...
```
ggetz: We likely want to export this function so that the regular cesium build process can use this. Could you do the following?

1. Encapsulate this in a function and export that from this file
2. Add this file to the `files` property in `package.json` so that consumers can target it

Maybe consider nesting this in a `scripts` directory so that we can keep any build or "server-side" script organized.

packages/sandcastle/buildGallery.js
```diff
@@ -0,0 +1,111 @@
...
+
+const galleryDirectory = "./public/gallery";
+
+const yamlFiles = globbySync([
+  `${galleryDirectory}/*/sandcastle.yaml`,
```
ggetz: Can we allow `yaml` or `yml`? I believe both are valid.

packages/sandcastle/buildGallery.js
```diff
@@ -0,0 +1,111 @@
...
+  `!${galleryDirectory}/list.json`,
+]);
+
+console.log("files", yamlFiles);
```
ggetz: Please clean up the extraneous `console.log` and comments.

packages/sandcastle/buildGallery.js
```diff
@@ -0,0 +1,111 @@
...
+  if (!metadata) {
+    console.error(filePath, "is empty!");
+    continue;
+  }
+
+  const { id, title, description, thumbnail } = metadata;
+
+  // Validate metadata
+
+  const slug = basename(dirname(filePath));
```
ggetz: Maybe use [`slugify`](https://www.npmjs.com/package/slugify) to ensure we're covering our bases, and using that as the `id` directly. That way, there's less room for user error.

Also using the slugified version of the path would also be preferable to make sure we don't get some of the ugly URLs we do now, like `3D%20Tiles%20Clipping%20Planes.html`.

packages/sandcastle/buildGallery.js
```diff
@@ -0,0 +1,111 @@
...
+    console.error(id, "- Missing thumbnail", thumbnail);
+    writeFile = false;
+  }
+
+  output.entries.push({
+    id: id,
+    title: title,
+    thumbnail: thumbnail,
+    description: description,
+    isNew: false,
```
ggetz: What the plan for `isNew`?

packages/sandcastle/public/gallery/3d-models/sandcastle.yaml
```diff
@@ -0,0 +1,10 @@
+id: 3d-models
+legacy-id: 3D Models.html
+title: 3D Models
+thumbnail: ./thumbnail.jpg
+description: Create 3D models using glTF.
+labels:
+  - Tutorial
+  - Showcases
+category: Working with 3D Tiles
```
ggetz: Do we use `category` yet? If not, we may want to omit it for the time being.

packages/sandcastle/public/gallery/billboards/sandcastle.yaml
ggetz: I think I'm fine either way. Letting this be an arbitrary path could allow for more flexibility in the future, like if we decide to share resources. But not a huge sticking point IMO.

packages/sandcastle/src/Helpers.ts
```diff
@@ -1,20 +1,46 @@
...
-  Sandcastle.finishedLoading();
-  };
-  if (typeof Cesium !== 'undefined' && typeof Sandcastle !== 'undefined') {
-      window.startupCalled = true;
-    window.startup(Cesium, Sandcastle).catch((error) => {
-        "use strict";
-      console.error(error);
-    });
-}
+  //   return `window.startup = async function (Cesium, Sandcastle) {
```
ggetz: Remove all this commented out code?

packages/sandcastle/src/Helpers.ts
```diff
@@ -1,20 +1,46 @@
...
+  //   })
+  //     `;
+  //   return `import * as Cesium from 'cesium';
+  //   import Sandcastle from 'Sandcastle';
+  // ${addExtraLine ? "\n" : ""}${code.replace(/import[\w\s*;']*;/, "")}
+  // Sandcastle.finishedLoading();
+
+  let imports = "";
+
+  if (!/import.*Cesium.*from.*;?$/m.test(code)) {
```
ggetz: Should the .* be `\s+` for whitespace?

packages/sandcastle/src/main.tsx
```diff
@@ -3,6 +3,13 @@ import { createRoot } from "react-dom/client";
 import "./reset.css"; // TODO: this may not be needed with itwin-ui
 import App from "./App.tsx";
 
+const host = window.location.host;
+if (host.includes("localhost")) {
+  document.title = `${host.replace("localhost:", "")} ${document.title}`;
+} else if (host.includes("ci")) {
+  document.title = `CI ${document.title}`;
```
ggetz, line 10: What's the goal here? "CI" is an implementation detail for many who just want to view a demo.

ggetz (changes requested):

packages/sandcastle/templates/Sandcastle.ts
jjspace: I didn't dig into the reason too far but something made the stack output change and it wasn't detecting the line anymore. I _think_ it's because we switched to modules. Either way dropping the `(` will work with both forms.

This whole process needs a validation pass across browsers. I was planning to do it whenever I re-implement the console in the app.

packages/sandcastle/buildGallery.js
jjspace: I think I was thinking this could highlight new sandcastles per release. I thought we did that already but I don't think we actually do and I'm not sure where I got that idea. Either way the more I think about it the more I'm not sure it's useful, at least without _fully_ automating it. Probably better to just stick to a Showcase tag and make sure to keep that updated

packages/sandcastle/src/main.tsx
jjspace: This was a bit of a quick patch mainly to help me while developing so I can tell which "environment" each tab I have open is in. This will probably get cleaned up more and include the actual title of a sandcastle for gallery items. I don't think "CI" is that bad here since it's very likely to only be running and deployed in our own CI. This wouldn't impact an actual "prod" version

jjspace:
Reminder for myself, I realized the gallery is not working in CI. I think this is just a matter of not running the build gallery script in CI and boy updating the gallery url. I think this should be a quick fix but I'll take a look tomorrow

packages/sandcastle/src/main.tsx
ggetz: I think even in cases where something is deployed to branch, there's still cases where the Sandcastle is used for demo purposed. For instance, the recent Gaussian Splats work is being shown in example Sandcastles from that branch.

packages/sandcastle/templates/bucket.html
jjspace: It still needs to be set somewhere and in a file that's processed by Vite to get the correct path based on the env it's building for. I think this is still the best way to do that. Let me know if you have a better idea.

packages/sandcastle/buildGallery.js
jjspace: Looked at this again, I do think we want it but it needs automated. We _do_ already do this with the existing gallery:

https://github.com/CesiumGS/cesium/blob/c8b5e418de01cf42757f59ce3b6cfe5ebbb2f39f/scripts/build.js#L634-L638

Probably want this same concept again later but checking it using `git` will mark them _all_ as new until we have the initial full release of the new sandcastle.

packages/sandcastle/src/Helpers.ts
jjspace: I'm gonna leave this as is unless you think of a specific reason it should be different. I don't really care what's between the "tokens" here but there should be some line combining `import`, `Cesium` and `from`. The generated code is also slightly different between `Cesium` and `Sandcastle but this way the regex to check is the same.

jjspace:
@ggetz I've updated this PR to address most of your comments and responded to others. I think there's still a bit of cleanup I have yet to do but hopefully this is closer.

The gallery has been solidified a bit and the route is now changed depending on the environment so [the CI build](https://ci-builds.cesium.com/cesium/nested-gallery/Apps/Sandcastle2/index.html) now has a working gallery again. It also no longer gets copied and duplicated into the local build, instead it points to the `packages/sandcastle/gallery` directory so local edits will always be reflected immediately.

packages/sandcastle/src/App.tsx
```diff
@@ -326,20 +327,59 @@ function App() {
...
+      if (!galleryItem) {
+        console.error("Unable to find gallery item with id:", galleryId);
+        return;
+      }
+      loadGalleryItem(galleryItem);
+    } else if (searchParams.has("id")) {
+      const galleryId = searchParams.get("id");
+      console.log("load id", galleryId);
+      // TODO: there's probably a race condition here where the list might not be loaded yet
+      // I need to switch this to the more React declarative style
```
jjspace, line 379: This part is already addressed in #12639

javagl:
The `buildSandcastles.js` causes some hiccups on Windows.

- `import.meta.url` is `file:///C:/cesium/packages/sandcastle/scripts/buildGallery.js`
- `process.argv[1]` is `C:\cesium\packages\sandcastle\scripts\buildGallery.js`

causing it to not even enter `buildGalleryList`. When "fixing" that pragmatically, https://github.com/sindresorhus/globby/issues/155 kicks in. Investigating some options...

EDIT:

I think that
```
import { pathToFileURL } from "node:url";

...

if (import.meta.url.endsWith(pathToFileURL(process.argv[1]))) {
  const defaultGalleryDirectory = "../gallery";
  const { output, hasErrors } = buildGalleryList(defaultGalleryDirectory);
  console.log("processed", output.entries.length, "sandcastles");
  if (hasErrors) {
    exit(1);
  }
}
```
might work, if that's OK on other OSes.

jjspace:
Good catch, sorry about that @javagl. That little bit of code is something I've used in another project based on this answer for [how to detect if an es module is "main"](https://stackoverflow.com/questions/64180480/how-do-i-detect-if-an-es-module-is-the-main-module) but I've only ever used it on linux.

Your suggested solution seems to still work fine so I pushed that up as a fix. Longer term if we integrate the build function with builds correctly we may not even need to keep that "is main" detection at all, but it's been helpful while building that script so I left it for now :woman_shrugging:

javagl:
Beyond the comment above, I'm not in the position to do a profound _technical_ review (regarding vite, tsx, or CI details). 

Details that may already be on the radar:

- It's necessary to run `npm run buld-app` in the `package/sandcastle` directory (that may be evident, or OBE when this is merged, and maybe the documentation is or will be updated, but I just ran the usual `npm run start` and was surprised that I could open the Sandcastle, but not the Sandcastle2)
- It's necessary to explicitly run `npm run build-ts` (because the code completion needs the TS part, apparently)

---

Some comments that may not belong here, but of which I don't know where they belong (maybe https://github.com/CesiumGS/cesium/pull/12639 ?)

- The new "Add... " buttons are _really_ useful. 
  - One thing that I'm a big fan of, and that I've always been missing in the old sandcastle, are _sliders_. Dragging and seeing the change just makes me feel so connected to my little computing machine 🤗 . It would be great if that could go on the medium-term radar. like `Sandcastle.addSlider("Slide!", min, max, value, function(currentValue) {...})`. 
  - The "Add..." buttons are inserting at the bottom of the code. Could they insert at the current cursor position? 
- I tried the basic parts of code completion, and I'm a big fan of that as well
  - Pressing key sequences like "A [CTRL+Space] . S [ DOWN, ENTER ] ( n [CTRL+Space] [ENTER] )" to write a line like `applicationWindow.showInputDialog(nameProvider)` just feels _so_ more efficient than "real typing"...
  - The help only appears when one of the dropdown options is selected. This is the same as for VSCode. I'm probably just [spoiled](https://github.com/user-attachments/assets/59cb394d-94ac-43e2-8c81-14c9de350a56)...
- The fact that the HTML panel is always visible below the code panel might be hard to get used to...
- The window elements are all fixed size. I assume/hope that they will (all) become such "draggable splitters" like in the old sandcastle

---

Some comments that _may_ belong here:

I have a bunch of Sandcastles from Forum support. They are just a bunch of `.js` files in a directory. (And sometimes HTML, when I needed a slider 😆 ). They follow the name pattern
`Sandcastle Forum <threadId> <short summary>.js`
And I wrote a short script that just squeezes them into the required gallery structure: Create directories called `short-summary`, copy the `.js` file there as `main.js`, add some dummy thumbnail and `index.html`, and a `sandcastle.yaml`. And the running dev server immediately picked them up:

Cesium SandcastleV2 0001

Yeah, the thumbnail layout is not perfect (and there is no horizontal scrollbar), but the technical structure seems to work nicely. 

It's not yet clear how exactly the old sandcastles will be "migrated" into the new structure. But given that the _target_ structure is relatively simple (with clearly separated files, and each directory being "one sandcastle"), this shouldn't be too hard.

javagl:
Two more details:

- Where will the 'description' from the YAML appear? (In a tooltip, or somewhere "permanent"?)
- An indication which sandcastle is selected could be useful (although this would then also have to indicate whether it was _modified_ after it was selected...)

jjspace:
Thanks for the feedback @javagl. As a blanket statement, the UI is still under development and definitely not finalized. This PR is focusing mainly on the functionality and specifically the new gallery structure. 

> Details that may already be on the radar

The build process is close but I am aware of a couple quirks like you pointed out. These will be cleaned up long term.

> The new "Add... " buttons are really useful.

I agree, they're things I've used in Sandcastle for a while. I'm still trying to work out the best way to integrate them, I don't think those buttons will stay as is. I'm thinking of adding snippets to the editor as one solution for now.

> The fact that the HTML panel is always visible below the code panel might be hard to get used to...

This is not staying there, we're going back to the tabs (already done in #12639). This was just a holdover being easier to implement at the hackathon.

> TI assume/hope that they will (all) become such "draggable splitters" like in the old sandcastle

That _is_ still the goal but I haven't looked into a good way to do it yet. Again the UI is still in flux.

> And I wrote a short script that just squeezes them into the required gallery structure
> ... but the technical structure seems to work nicely.
> 
> It's not yet clear how exactly the old sandcastles will be "migrated" into the new structure. 

This is exactly what I was hoping, glad it worked for you! Hopefully this new structure will be easier to maintain. I will be creating a one time migration script to convert all the existing sandcastles once the new structure is confirmed (which I think it basically is at this point).

ggetz:
@jjspace I'm taking another pass now, thank you for the updates! It'd be super helpful for me if you go back to your original comment and update description and your TODO list. It helps me understand the status. Are you still working through items in this PR, or is everything ready for final review before merge?

jjspace:
I started working on the conversion script (pretty quick and working well) but I realized we forgot to think about the development only sandcastles

**How should we handle the development sandcastles? Still a separate directory? should it be nested? or should it just be a new label? or maybe a property of the yaml?** 

I think there's 2 options I like

1. Add a new property to the `sandcastle.yaml` metadata for `developmentOnly` which is optional and defaults to false.
    - Then we can modify the list generation to ignore any where it's true for prod
3. A separate directory _alongside_ the new `gallery` directory, not nested
    - if it was nested that's adding annoying special handling to treat it differently than the other sub-directories
    - If we do this we'd have to generate a second gallery-list
    - This is potentially easier to manage from a _maintainer_ perspective as it's more clear if a sandcastle is development only or not depending where it's located. Similar to how it is now.

**Adjacent question, does it make sense to support loading multiple galleries? probably not tooo difficult but not sure if it's worth**

This is most important if we go with option 2 above so we could load just the main gallery or the main gallery _and_ the development gallery.
This could also enable us to let developers point to their own gallery _in addition to_ the default gallery. However I don't think that would really apply to anyone other than maybe @javagl right now

@ggetz thoughts?

packages/sandcastle/gallery/3d-models-coloring/main.js
```diff
@@ -0,0 +1,176 @@
...
+  colorBlendAmount: 0.5,
+  colorBlendAmountEnabled: false,
+  silhouetteColor: "Red",
+  silhouetteColors: ["Red", "Green", "Blue", "Yellow", "Gray"],
+  silhouetteAlpha: 1.0,
+  silhouetteSize: 2.0,
+};
+
+// Convert the viewModel members into knockout observables.
+Cesium.knockout.track(viewModel);
```
ggetz, line 38: Totally out of scope of this PR, but we should really make a pass at getting rid of knockout from our examples. It leads to [users depending on it](https://github.com/CesiumGS/cesium/issues/12423), and ideally we want to move away from depending on it. Could you please write up an issue to take a content pass to get rid of the knockout bindings from Sandcastle examples?

packages/sandcastle/package.json
```diff
@@ -23,8 +24,10 @@
     "@types/react-dom": "^19.0.4",
     "@vitejs/plugin-react": "^4.3.4",
     "globals": "^15.15.0",
+    "globby": "^14.1.0",
```
ggetz, line 30: If we're being pedantic, since `globby` and `yaml` are required by `buildGallery.js`, I believe that makes then direct dependencies.

packages/sandcastle/gallery/3d-models-coloring/main.js
jjspace: I _think_ most of the places it's used in sandcastles right now are to set up sliders and more complex toolbar inputs. I was thinking this could be handled by expanding the `Sandcastle` "API" with more features. It's been in the back of my mind but I'll make sure it's written up somewhere

ggetz:
> How should we handle the development sandcastles? Still a separate directory? should it be nested? or should it just be a new label? or maybe a property of the yaml?

Making it a property of the `yaml` file seems easier to manage from a content generation perspective IMO. Though I may suggest using `development` as the property for brevity. 🙂 

Ultimately though, either approach should be fine as long as there is one standard process and it is documented.

ggetz:
> Adjacent question, does it make sense to support loading multiple galleries? probably not tooo difficult but not sure if it's worth

Worth discussing, but could we park the implementation this for now and address it in a later PR?

packages/sandcastle/gallery/3d-models-coloring/main.js
ggetz: Sounds good. I think it'll be good to write up an issue just in case we don't cover all the cases. If we end up doing so, then 🎉, we can close that issue.

packages/sandcastle/templates/tsconfig.lib.json
```diff
@@ -0,0 +1,12 @@
+{
+  "include": ["./Sandcastle.ts"],
+  "compilerOptions": {
+    "tsBuildInfoFile": "../node_modules/.tmp/tsconfig.lib.tsbuildinfo",
```
ggetz: So why do we need to set [`tsBuildInfoFile`](https://www.typescriptlang.org/tsconfig/#tsBuildInfoFile)? And why is it using a path in `node_modules`?

packages/sandcastle/src/Helpers.ts
ggetz: But to some extend we _do_ care what's between those tokens. Right now, this regex has the potential match a lot of unexpected things:

`importCesiumfrom`
`import CesiumSomething from SomethingCompletelyDifferent;`
`import something from Cesium; import somethingElse from anotherThing;`
`import(Cesium); myVariablewithfrom`

which are all really different from what we're looking for, right?

In general, [using the greedy `.*` is generally recommended against](https://unix.stackexchange.com/questions/441927/how-or-why-using-is-better-than).

packages/sandcastle/scripts/buildGallery.js
```diff
@@ -0,0 +1,139 @@
...
+import { parse } from "yaml";
+import { globbySync } from "globby";
+import { basename, dirname, join } from "path";
+import { exit } from "process";
+import { fileURLToPath, pathToFileURL } from "node:url";
+
+// TODO: probably need to find a way to integrate this into the vite process to make sure it's
+// built during build
+// May also want to move this out of the sandcastle project so we don't have to duplicate
+// files for the local build, need to think about deploying too...
```
ggetz: So can we remove this TODO now that it's brought into the root `build` file?

packages/sandcastle/buildGallery.js
ggetz: @jjspace What are your thoughts on this comment?

ggetz:
I'm still attempting to test this out, by npm appears to have other plans...

image

packages/sandcastle/package.json
jjspace: The `buildGallery` script is a build/bundle step only. It's equivalent to `vite` which included itself as a dev dependency. These packages are not needed to run the production code so they don't belong in the "true" `dependencies`

packages/sandcastle/templates/tsconfig.lib.json
jjspace: Without setting it it defaults to output this file in the same directory. I set it to inside the node modules just matching the pattern in the main `tsconfig` that was generated from the vite setup. the `node_modules` is already `gitignored` so it seems like a reasonable place to put some of these extra build related temp files.

ggetz:
@jjspace Ok to handle this in a follow up PR, just let me know the plan.
 
 I noticed two things about "New" sandcastles:

- Should a new Sandcastle now be the following?

````js

````

- I noticed the autocomplete suggestions were working for loaded gallery examples, but not for "New" sandcastles.

**Comment by ggetz** [comment]
Otherwise, I tested this out by running the following commands and things appear to be running smoothly.

```
npm run build
npm run build-sandcastle
npm start
```

**Comment by jjspace** [comment]
> Should a new Sandcastle now be the following?

It's not required but I'm fine encouraging it to follow closer to how people would use in their own applications.

> I noticed the autocomplete suggestions were working for loaded gallery examples, but not for "New" sandcastles.

They _should_ work for both scenarios. Can you make sure the declaration file is generated for Sandcastle using `npx tsc -b`? It's possible the network request is failing and then not setting up the types for intellisense.
If that's not the cause can you be more specific how to reproduce?

File: packages/sandcastle/package.json (hunk 19 above)
Comment (ggetz) [reply packages/sandcastle/package.json:30]: The built app is fetching `list.json at runtime`, correct? So this script can and is run by the downstream consumer to update it. Sounds like it's not a build/bundle step to me.

File: packages/sandcastle/templates/tsconfig.lib.json (hunk 20 above)
Comment (ggetz) [reply packages/sandcastle/templates/tsconfig.lib.json]: It's generally discouraged to write or modify any content to the `node_modules` folder, as it is inherently transitory. Please use another location and add it to the `.gitignore` as necessary. 

What's the issue that arises when that file is written to the current directory? By default, this is written to next to the "out" location, so it may make sense to adjust the output location instead.

File: packages/sandcastle/buildGallery.js (hunk 11 above)
Comment (ggetz) [reply packages/sandcastle/buildGallery.js]: @jjspace Apologies, _this_ was actually the comment I meant for "what are you thoughts on this comment?"

**Comment by ggetz** [comment]
From the repo root:

```
npm clean -dxf
npm install
npm run build
npm run build-sandcastle
npm start
```

1. The first time I open Sandcastle, see the following message. (That wasn't happening before your last commit I believe). 

image

2. If I switch to a gallery example, everything works fine.

3. If I press the "New" button, the console looks cleaner, but I still don't see suggestions for `Cesium.`

image

image

Sandcastle suggestions look to be working fine though.

image

packages/sandcastle/buildGallery.js
jjspace: @ggetz I like the idea of using `slugify` as part of the conversion script from the existing gallery to the new one. However I don't think we need to use it here. I'd rather the files be as static as possible which means they should match the slug format in the first place. If it doesn't the developer should fix it and this build script will be noisy about it. 

The FE needs to know where to access the files and images. The id is used in that route `[base page]/gallery/[id]`. If we were to generate the id as a `slugify` of the directory name it may not match.

ggetz: OK, I agree with your first point.

For the latter, sure, we do need an id in the generated JSON. But do we actually need to specify an `id` property in the YAML? Or can we instead take it directly from the file subpath (the "slug") to avoid redundancy and the possibility of the slug and id not matching in the first place?

jjspace: > But do we actually need to specify an id property in the YAML?

I was really on the fence about it. I do think it's a little redundant, but I also felt that it was nicer to have the "complete" details about a sandcastle in the yaml. I could go either way, lmk if you feel strongly about either and I can make a change if needed.

packages/sandcastle/package.json
jjspace: > The built app is fetching `list.json` at runtime, correct?

The built app is fetching `gallery-list.json` as a _static file_ at runtime. It's not generated "on the fly" using the functions in `buildGallery`.

packages/sandcastle/templates/tsconfig.lib.json
jjspace: > It's generally discouraged to write or modify any content to the node_modules folder, as it is inherently transitory.

Generally I would agree with you. However this is following the example set by the default `tsconfig` set up with the [vite template](https://vite.dev/guide/). As I understand it the `tsbuildinfo` file is also pretty temporary and only there to help speed up subsequent local builds.

> What's the issue that arises when that file is written to the current directory?

Just extra noise in the file tree and needing to add it specifically to the gitignore. 

Edit:
Looks like their motivation was the same as mine. It's just a nice easy place to put these temp files that's already under the gitignore umbrella with the rest of `node_modules` https://github.com/vitejs/vite/pull/18435

If you still think they should be moved elsewhere just lmk where. I don't want to bikeshed on this

ggetz: But this is not built using vite, correct? It's built with tsc. 

What vite does internally isn't so much my concern– But making sure our project follows good practices and is maintainable is. I'm concerned this approach is obscuring details by doing things in a non-conventional way, making things more difficult to understand and maintain for others in the future. 

`Sandcastle.js` and `Sandcastle.d.ts` are already ignored correct? Perhaps the cleanest solution would be to set the output to a subdirectory, such as `templates/dist` and ignore the entire directory in the `.gitignore`.

packages/sandcastle/package.json
ggetz: I still disagree since we're generating `gallery-list` "on the fly" in our root dev server— But I admit its pedantic and won't bikeshed it.

packages/sandcastle/buildGallery.js
ggetz: I'm a bit concerned about extra noise and cycles on errors when `id` fail to match (even though it _is_ documented, thank you!). I'm leaning towards removing it.

packages/sandcastle/templates/tsconfig.lib.json
jjspace: > But this is not built using vite, correct? It's built with tsc.

It _is_ built with vite when running the local dev server to allow for quicker updates when actively developing sandcastle directly.

> What vite does internally isn't so much my concern– But making sure our project follows good practices and is maintainable is. I'm concerned this approach is obscuring details by doing things in a non-conventional way, making things more difficult to understand and maintain for others in the future.

The point I was trying to make is that the method I was using (tmp dir in node_modules) is what the vite scaffold project does by default. I extrapolated this to mean that that _is_ the conventional or expected way to handle these files and is a "good practice". At least the expected way within vite projects. 

Regardless, I've moved them all next to their configs by just removing the config path. It's easy to change later if we want

jjspace:
@ggetz Updated, please take another look. I think it should be good to go

packages/sandcastle/templates/tsconfig.lib.json
ggetz: Thanks for updating. We can circle back to this discussion and to vite practices as needed, but I think what's there now is just easier to understand at the end of the day.

ggetz (approved):

ggetz:
Thanks @jjspace, especially for your patience! 😬

PR #12639, opened 2025-05-29
Solidify new Sandcastle React structure

jjspace:
# Description

When I originally threw this together for the hackathon I tried to reuse a lot of the existing code for current sandcastle to speed up development. This meant a lot of the logic was very "imperative" oriented and was not playing nicely with React. This worked fine for a while but started to really get in the way in #12631, specifically with multiple loading methods.

This PR restructures the application to be much more "React" oriented

- State is now much more actively maintained and controlled
    - This prevents all the direct monaco editor `getValue` and `setValue` access I was doing before
- The `iframe` wrapper logic has been pulled out to it's own `Bucket`
    - "Running" a sandcastle is now controlled by changing the `props` on the `Bucket` using the `comittedCode` and `runNumber` state
- Unified into a single Monaco `Editor` instance using multiple Models. It is now a controlled component and the contents are tracked and stored in state

This code builds off of #12631 and should be merged after that

## Issue number and link

Part of #12566 

## Testing plan

- Run sandcastle using `npm run dev` from inside `packages/sandcastle`
- Load sandcastle at `http://localhost:5173/`
- Make sure that:
    - Loading from a share URL ([like this](http://localhost:5173/#c=bY7BTsMwDIZfxeSyDqFEnMmqSbtOQgLBKZcs9dpobjzFaavx9HQbQyA4/vb32X/sj5wL3IMX2KDEoYd95h6cCpfk1JNLLgVOUmCMOGGGFSScvmj9fplVN37DqfiYMDu1/DaZUBO31WIObfZF7qB0CNL5jPD2soWJ8wGbxdIl9aCslBNh7RLAOl7rDZkqrU3B/ki+oJjdEA5YdBA5f7HmptgmjhCb1T91IJAXmTf7geg1fqBTtTUz/0sj9k1M7fOImfzpjHSP9fY61FpbM8e/VmGmnc8/Ln4C)) works and has the correct code and viewer state _on first load_
    - Clicking gallery items loads the correct code, refreshes the viewer and updates the URL
        - Clicking the same item again should reload that item but _not_ create an extra entry in history. test with back/forward buttons in browser
    - Clicking `New` resets both the JS and HTML code, refreshes the viewer and updates the URL to the base route
    - Navigating to a "legacy url" like http://localhost:5173/?src=3D%20Models.html works and replaces the url with the new `id`
    - Swapping between JS and HTML works as expected
    - The `Add button` and others work but do _not_ refresh the viewer until `Run` is clicked
    - Hitting `F8` while focusing inside the code actually reloads the viewer
    - Clicking `Run` or `F8` multiple times in a row _with no code changes_ refreshes the viewer every time
        - This is specifically why I added the `runNumber` state, this is a workflow that must work

# Author checklist


packages/sandcastle/public/templates/bucket.css
```diff
@@ -15,6 +15,27 @@ body {
   width: 100%;
   height: 100%;
   overflow: hidden;
+
+  background-color: white;
+  background-image: var(--_rings), var(--_gradient);
+
+  /* TODO: these should pull from the stratakit directly somehow eventually */
```
ggetz, line 22: Can we remove the TODO's please and get them documented in a GitHub comment?

packages/sandcastle/src/App.tsx
```diff
@@ -32,403 +22,117 @@ function getBaseUrl() {
...
-  }
+const TYPES_URL = `${__PAGE_BASE_URL__}Source/Cesium.d.ts`;
+const SANDCASTLE_TYPES_URL = `templates/Sandcastle.d.ts`;
+const GALLERY_BASE = __GALLERY_BASE_URL__;
 
-  for (i = 0, len = nodes.length; i < len; ++i) {
-    bucketDoc.head.removeChild(nodes[i]);
-  }
+function App() {
+  const [jsIsActive, setJsIsActive] = useState(true);
```
ggetz: Took me a moment to figure out what this was for. Maybe `isJsTabActive`?

packages/sandcastle/src/App.tsx
```diff
@@ -32,403 +22,117 @@ function getBaseUrl() {
...
-      appendConsole("consoleError", `Error loading ${this.src}`, true);
-      appendConsole(
-        "consoleError",
-        "Make sure Cesium is built, see the Contributor's Guide for details.",
-        true,
-      );
-    }
+  const editorRef = useRef<editor.IStandaloneCodeEditor>(null);
+
+  type Action =
```
ggetz: I believe `Action` is type used by react already. Maybe `SandcastleActions` to avoid any conflicts.

packages/sandcastle/src/App.tsx
```diff
@@ -32,402 +22,116 @@ function getBaseUrl() {
...
-    if (bucketFrame.contentDocument !== bucketDoc) {
-      // A newer reload has happened, abort this.
-      return;
-    }
-    if (nodes.length > 0) {
-      while (nodes.length > 0) {
-        node = nodes.shift();
-        if (!node) {
-          continue;
+  const [codeState, dispatch] = useReducer(
```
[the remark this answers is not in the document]
ggetz: Reducers are all good in my book 👍

packages/sandcastle/src/App.tsx
```diff
@@ -32,403 +22,117 @@ function getBaseUrl() {
...
+    function reducer(state: CodeState, action: Action): CodeState {
+      switch (action.type) {
+        case "reset": {
+          return {
+            code: defaultJsCode,
+            html: defaultHtmlCode,
+            committedCode: defaultJsCode,
+            committedHtml: defaultHtmlCode,
+            runNumber: 0,
```
ggetz: I would probably define `intitialState`...

````js
const initialState: CodeState = {
    code: defaultJsCode,
    html: defaultHtmlCode,
    committedCode: defaultJsCode,
    committedHtml: defaultHtmlCode,
    runNumber: 0,
 };
 ````

and then use it here...

````js
    return { ...initialState };
````

and as the last option to `useReducer`

File: packages/sandcastle/src/App.tsx [hunk 6]
```diff
@@ -32,403 +22,117 @@ function getBaseUrl() {
...
-      //group 1 attribute name, group 2 attribute value.  Assumes double-quoted attributes.
-      let attributeMatch;
-      while ((attributeMatch = attributeRegex.exec(bodyAttributes)) !== null) {
-        const attributeName = attributeMatch[1];
-        const attributeValue = attributeMatch[2];
-        if (attributeName === "class") {
-          bucketDoc.body.className = attributeValue;
-        } else {
-          bucketDoc.body.setAttribute(attributeName, attributeValue);
+        case "commitCode": {
```
ggetz: Perhaps a bit nitpicky, but is `commit` the right verb here? Why not `run`?

packages/sandcastle/src/App.tsx
```diff
@@ -718,60 +325,97 @@ Sandcastle.addToolbarMenu(${variableName});`,
...
-  useEffect(() => {
-    if (galleryLoaded) {
-      console.log("gallery loaded, try loading from url");
-      loadFromUrl();
-    }
-  }, [galleryLoaded]);
+  useEffect(
+    function loadFromUrl() {
+      if (galleryLoaded) {
+        console.log("gallery loaded, try loading from url");
```
ggetz: Remove debug logs please.

packages/sandcastle/src/App.tsx
```diff
@@ -718,60 +325,97 @@ Sandcastle.addToolbarMenu(${variableName});`,
...
       </div>
       <div className="gallery">
         <Gallery
           demos={galleryItems}
-          loadDemo={(item) => loadGalleryItem(item)}
+          loadDemo={(item) => {
+            loadGalleryItem(item.id);
+            // TODO: I like this method for history but we need to actually react to the user
+            // hitting forward or back, right now it does nothing. Need to look into how that works
+            // Might need to involve React-Router but I'd like to try and avoid if we don't need it
```
ggetz: What the plan here?

packages/sandcastle/src/Bucket.tsx
```diff
@@ -0,0 +1,280 @@
+import { useEffect, useRef } from "react";
+import { embedInSandcastleTemplate } from "./Helpers";
+
+// TODO: I don't think this will correctly track calls between state?
+// I need to figure what this is even here for and whether we need it still...
```
ggetz: And what's the plan here? It looks like this file still has a fair bit of cleanup needed.

angrycat9000:
Did a read through of the code.  The sandcastle v2 link didn't work for me so I wasn't sure how to run it.

The DOM manipulation in `Bucket` raised a red flag for me.  Would have expected that everything is written in React and not setting innerHTML.  If we do need to do the direct manipulation then I would clearly document why it is needed.

Otherwise looks pretty standard.

Would also recommend breaking stuff down into smaller components for readability.  The App component is a bit hefty and has a lot of logic that could probably be better isolated.  Eg the editor set/teardown could be in its own Editor component.  The gallery loading could be in its own component that uses "onGallerySelected" to communicate back to the main app.

jjspace:
Thanks for the review @angrycat9000 

> The DOM manipulation in Bucket raised a red flag for me. Would have expected that everything is written in React and not setting innerHTML. If we do need to do the direct manipulation then I would clearly document why it is needed.

The only DOM manipulation we're still doing is to modify/control the contents of the iframe, ie the "bucket", to load the sandcastle code. AFAIK there's not a way to do that more directly using React. I moved all that code into the `Bucket` file specifically to have it all isolated into that single component.

packages/sandcastle/src/App.tsx
jjspace: I used "commit" because it mirrored the names of `committedCode` and `committedHtml` but I do agree "run" reads better. Updated

packages/sandcastle/src/App.tsx
jjspace: I thought this would be more complicated than it was but I've hooked up to the browser `popstate` events to listen and respond to navigation which makes it work nicely.

jjspace:
@ggetz this PR has been cleaned up a bunch and should be good for another look

ggetz (approved):

ggetz:
Looking good! Thanks @jjspace!

PR #12679, opened 2025-06-18
Sandcastle helper snippets

jjspace:
# Description

I have long wanted to include some sort of "code helpers" for adding common chunks of code to sandcastle. Currently these exist as dedicated buttons in the new Sandcastle but I'm not sure they will stay that way in the final UI. These are mostly for more "power users" so I've taken a pass at implementing these as snippets directly in monaco. These can always be expanded later.

- `scbutton` will generate the button code
- `sctoggle` will generate the checkmark toggle button code
- `scmenu` and `scmenuitem` will generate code for the dropdown menus.

All of these snippets have tab stops to make it quick an easy to input the labels, variables and your code.

Based on the monaco [code completion example](https://microsoft.github.io/monaco-editor/playground.html?source=v0.52.2#example-extending-language-services-completion-provider-example)

## Issue number and link

Part of #12566 

## Testing plan

- Run new sandcastle (`npm run dev` in the package or `npm run build-sandcastle && npm star` at top level)
- anywhere in the editor type any of the snippet labels above and hit tab/enter
- Try out a mix and make sure they work as expected

# Author checklist

### DIFF
--- a/packages/sandcastle/src/setupSandcastleSnippets.ts
+++ b/packages/sandcastle/src/setupSandcastleSnippets.ts
@@ -12,8 +12,8 @@ function createSandcastleSnippets(range: Range): languages.CompletionItem[] {
       kind: languages.CompletionItemKind.Function,
       documentation: "Create a Sandcastle button",
       insertText: `Sandcastle.addToolbarButton(\${1:"New Button"}, function () {
-    \${0:// your code here}
-  });`,
+  \${0:// your code here}
+});`,
       insertTextRules: languages.CompletionItemInsertTextRule.InsertAsSnippet,
       range: range,
     },
@@ -22,9 +22,9 @@ function createSandcastleSnippets(range: Range): languages.CompletionItem[] {
       kind: languages.CompletionItemKind.Function,
       documentation: "Create a Sandcastle toggle button",
       insertText: `let \${2:toggleValue} = \${3:true};
-  Sandcastle.addToggleButton(\${1:"Toggle"}, \${2:toggleValue}, function (checked) {
-    \${2:toggleValue} = checked;$0
-  });`,
+Sandcastle.addToggleButton(\${1:"Toggle"}, \${2:toggleValue}, function (checked) {
+  \${2:toggleValue} = checked;$0
+});`,
       insertTextRules: languages.CompletionItemInsertTextRule.InsertAsSnippet,
       range: range,
     },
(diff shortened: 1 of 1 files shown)


javagl (commented):
I had a look at this. I cannot say much about the "code level", except that specifying the return type could make sense: `void` for `setSnippets` and
`function createDependencyProposals(range: Range) : languages.CompletionItem[]`
Just for additional documentation/type safety.

I also tried it out with different configurations, and it seems that it works as expected.

And... yeah ... it's me again...
`documentation: "Create a Sancastle button",` should be
`documentation: "Create a Sandcastle button",` 😁 
(in all `documentation` fields)

jjspace:
Thanks for taking a look @javagl! I added some types to help.

> "Sancastle" should be "Sandcastle"

I have a few typing quirks I can't seem to shake, like often typing `teh` instead of `the`. This seems to be a new one that I have to keep an eye on :laughing: I've updated my [auto correct](https://marketplace.visualstudio.com/items?itemName=sygene.auto-correct) for it

packages/sandcastle/src/SandcastleEditor.tsx
```diff
@@ -112,6 +114,79 @@ function SandcastleEditor({
     );
   }
 
+  function createDependencyProposals(range: Range): languages.CompletionItem[] {
```
ggetz: Could we move the snippets to their own file?

packages/sandcastle/src/SandcastleEditor.tsx
```diff
@@ -112,6 +114,79 @@ function SandcastleEditor({
...
+    // returning a static list of proposals, not even looking at the prefix (filtering is done by the Monaco editor)
+    return [
+      {
+        label: "scbutton",
+        kind: languages.CompletionItemKind.Function,
+        documentation: "Create a Sandcastle button",
+        insertText: `Sandcastle.addToolbarButton(\${1:"New Button"}, function () {
```
ggetz: Where does the `${1:"New Button"}` syntax come from? I've never seen it before.

jjspace: This syntax is for the various "tab stops" to make filling data easier. I can look for monaco specific documentation but I mostly referred to the VSCode snippets docs which I've used for snippets a bunch in the past https://code.visualstudio.com/docs/editing/userdefinedsnippets

jjspace:
@ggetz I moved the snippet setup to it's own file quickly while cleaning up some branches. This should be good to go now I think but no rush

ggetz:
Thanks @jjspace! We may wan to take a pass on which snippets to include and so forth as we do a final polish, but looking good overall.

Issue #6108, opened 2018-01-10
GeoJsonDataSource sourceUri parameter isn't used


tfili:
The `GeoJsonDataSource.load()` has a `sourceUri` parameter that isn't used to resolve relative links as GeoJSON files don't have any. It is used to get the datasource name, which can be set on creation of the datasource. This can be removed.

chriswingler:
Hi, I submitted a pull request for this.  Hope it works!

[remove unused sourceUri param in GeoJsonDataSource.load() #6116](https://github.com/AnalyticalGraphicsInc/cesium/pull/6116)

mark011062:
I'm with FCF and would like to work on this please

VasuParashar37:
Hi @tfili is this issue still open? I want to work on this issue.

javagl:
There was some back and forth in https://github.com/CesiumGS/cesium/pull/6116 (8 years ago) about what this issue is about and how it could be resolved. 

From my perspectivev (mainly @tfili ), there are three options

- Create a PR to solve it (**iff** it's clear what should be done here)
- or (if it's not clear: ) Elaborare the issue description here, to make clear what **should** be done
- or close it

(I could open hundreds of issues about something not being used or not making sense - we have to raise the bar...)

PR #6116, opened 2018-01-15
remove unused sourceUri param in GeoJsonDataSource.load()

chriswingler:
Hi, first time contributor here.  Hope this fixes [#6108](https://github.com/AnalyticalGraphicsInc/cesium/issues/6108)

- Chris Wingler

PR #12685, opened 2025-06-22
Remove unused sourceUri parameter from GeoJsonDataSource

khalisahkhan:
# Description

Removed the `sourceUri` parameter from the method signature and internal function calls in `GeoJsonDataSource.load()`.

- Cleaned up unused imports (e.g., `getFilenameFromUri`).
- Updated inline documentation to reflect the change.

The `sourceUri` parameter was no longer used in the logic and acted as a no-op. Removing it improves code clarity and reduces potential confusion for contributors and users of the API.

## Issue number and link

[Issue #6108](https://github.com/CesiumGS/cesium/issues/6108) by removing the unused sourceUri parameter from the GeoJsonDataSource.load() method.

## Testing plan

Verified that:
- The app builds successfully using `npm start`
- `GeoJsonDataSource.load()` continues to work without the `sourceUri` parameter
- No regressions occur in runtime behavior

# Author checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -424,3 +424,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Cody Butler](https://github.com/CodyBu)
 - [Hiwen](https://github.com/Hiwen)
 - [Matt Schwartz](https://github.com/mzschwartz5)
+- [Khalisah Khan](https://github.com/khalisahkhan)
--- a/packages/engine/Source/DataSources/GeoJsonDataSource.js
+++ b/packages/engine/Source/DataSources/GeoJsonDataSource.js
@@ -7,7 +7,6 @@ import Frozen from "../Core/Frozen.js";
 import defined from "../Core/defined.js";
 import DeveloperError from "../Core/DeveloperError.js";
 import Event from "../Core/Event.js";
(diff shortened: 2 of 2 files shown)


javagl:
There had been some confusion about this in https://github.com/CesiumGS/cesium/pull/6116 - not sure if that is all resolved now.

mzschwartz5:
Hi @khalisahkhan, thanks for taking a look at this issue. Reading over [this comment](https://github.com/CesiumGS/cesium/pull/6116#issuecomment-358757156) from the first PR that attempted to fix this, it seems like there may be some unintuitive behavior in this class that we're not addressing.

Specifically, it seems like we need to support an edge case where the user passes in a data URI geoJson to load, and not an object with a name (if I'm correctly understanding the comment I linked).

From that first PR, there's a unit test added:
```JS
  it("load works with a URL", function () {
    const dataSource = new GeoJsonDataSource();
    return dataSource.load("Data/test.geojson").then(function () {
      expect(dataSource.name).toEqual("test.geojson");
    });
  });
```

This test passes with the current `GeoJsonDataSource` class, but not after the changes in this PR - removing the `sourceURI` and using `geoJson.name`. That tells me we're still missing something.

Honestly, though, I'm sort of wondering if we should just let the existing code be - after all, it works, and it's not obviously wrong in any way.

mzschwartz5:
Closing due to inactivity - please reopen if/when you return to this PR.

Issue #12550, opened 2025-04-04
Texture3D


Hiwen:
### Feature

I want to extend Cesium to support Texture3D.
To achieve partial Texture3D effects from Three.js.
Although this is somewhat similar to Voxels, I believe they are not entirely the same and have different applicable scenarios.

![Image](https://github.com/user-attachments/assets/b4b85e97-39d9-45b1-9284-1651381d50a3)

ggetz:
Thanks @Hiwen, we agree this is related but not a replacement for voxels. We discussed potentially using them for storing voxel information.

@jjhembd or @lilleyse are there any additional things on your mind which should be documented here?

ggetz:
[Clouds](https://sandcastle.cesium.com/?src=Cloud%20Parameters.html) could be an isolated entry point to introduce 3D Textures. See [this three.js cloud rendering example](https://threejs.org/examples/#webgl_volume_cloud).

syzdev:
Regarding ray marching and volume rendering, another implementation is to use a 2D "megatexture" to simulate 3D textures, mainly based on the following considerations:

1. Support WebGL 1.0 (which may not be very important in modern browsers);
2. 2D textures are easier to apply texture compression algorithms and theoretically occupy less memory.
3. 2D texture sampling and interpolation are faster (ray marching requires frequent sampling);

Therefore, the question to be discussed is whether 2D textures will have better performance in this use case.

Hiwen:
@syzdev Maybe we can do a comparative test.

Hiwen:
@ggetz I'm very sorry. In the previous PR, I placed the sandbox example of VolumeCloud in the "develop" directory. As a result, after the release of 1.130, this example cannot be viewed in the sandbox. Now, should I submit a PR to move this example to the "gallery" directory, or should the publisher make the adjustment when releasing 1.131? Is it worth submitting a new PR for this minor issue?

ggetz:
Thanks for pointing this out @Hiwen!

@jjhembd Could you please take the lead here to get the Sandcastle moved to the correct directory?

ggetz:
@jjhembd Also consider if this issue should be closed as completed.

Hiwen:
> [@jjhembd](https://github.com/jjhembd) Also consider if this issue should be closed as completed.

@ggetz I suggest not closing this issue for now, as the work on 3D Texture has just started. I think we'll need to submit some PRs again soon for improvement.

ggetz:
[...]

PR #12701, opened 2025-06-28
Add wrapR property to Sampler and Texture3D

Hiwen:
# Description
Add wrapR property to Sampler and Texture3D, to support the newly added third dimension wrap.

## Issue number and link

#12550

## Testing plan
I have improved SamplerSpec.js.

# Author checklist


packages/engine/Source/Renderer/Texture3D.js
syzdev, line 1: typo: L22, L23

syzdev:
typo: 

image

Hiwen:
> typo:
> 
> image

nice catch

packages/engine/Source/Renderer/Sampler.js
```diff
@@ -94,6 +106,7 @@ Sampler.equals = function (left, right) {
 
 Sampler.NEAREST = Object.freeze(
   new Sampler({
+    wrapR: TextureWrap.CLAMP_TO_EDGE,
```
jjhembd, line 109: I added this line. I think it's OK to always have a `wrapR` value, even if it won't be used (for 2D Textures)

packages/engine/Source/Renderer/Texture3D.js
```diff
@@ -13,21 +13,30 @@ import Sampler from "./Sampler.js";
 import TextureMagnificationFilter from "./TextureMagnificationFilter.js";
 import TextureMinificationFilter from "./TextureMinificationFilter.js";
 
+/**
+ * @typedef {object} Texture3D.Source
```
jjhembd, line 17: I pulled this out into a separate `@typedef`, so we don't have to describe nested properties in the `Texture3D.ConstructorOptions` definition.

packages/engine/Source/Renderer/Texture3D.js
```diff
@@ -13,21 +13,30 @@ import Sampler from "./Sampler.js";
...
+ * @property {number} width The width (in pixels) of the 3D texture source data.
+ * @property {number} height The height (in pixels) of the 3D texture source data.
+ * @property {number} depth The depth (in pixels) of the 3D texture source data.
+ * @property {TypedArray|DataView} arrayBufferView The source data for a 3D texture. The type of each element needs to match the pixelDatatype.
```
jjhembd, line 21: The docs for [texImage3D](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/texImage3D) and [texSubImage3D](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/texSubImage3D) allow for these additional sources: `ImageBitmap`, `ImageData`, `HTMLImageElement`, `HTMLCanvasElement`, and `HTMLVideoElement`. Can we support those here? If so, we could generalize this type definition to `Texture.Source` and use it in `Texture.js` too!

jjhembd (commented):
Hi @Hiwen, sorry for the delayed review. This looks good overall. I re-merged `main` and tweaked a few JSDoc comments. Please take a look and let me know if the changes make sense.

I have one question about some additional source data types. Would it be easy enough to add those here, or do you think that should be a separate PR?

packages/engine/Source/Renderer/Sampler.js
Hiwen: That's good!

packages/engine/Source/Renderer/Texture3D.js
Hiwen: Good!

packages/engine/Source/Renderer/Texture3D.js
Hiwen: Using ImageBitmap, ImageData, HTMLImageElement, HTMLCanvasElement, and HTMLVideoElement as the source is a really good idea. However, I haven't actually used this yet, and I'm not sure about the specific effects. I think it would be better to open a new issue to support this.

javagl:
This PR is _somehow_ related to https://github.com/CesiumGS/cesium/issues/12781 . Linking just for awareness. Further changes may not be in the scope of this PR.

Issue #12700, opened 2025-06-27
Sandcastle CSS broken on Firefox v140


mzschwartz5:
### What happened?

On the latest version of Firefox (140 as of writing), Sandcastle's CSS appears to be broken:

<img width="1425" alt="Image" src="https://github.com/user-attachments/assets/576250fc-6a02-4d63-ad5c-2f1435af19bd" />

Firefox v139 and Chrome do not exhibit this issue.

@jjspace found that changing the import URL of the `bucket.css` file from:

`@import url(../templates/bucket.css);`
to
`@import url("../templates/bucket.css");` (quotes added)

fixes the issue on Firefox v140. He also noted that the [MDN spec says quotations are optional](https://developer.mozilla.org/en-US/docs/Web/CSS/url_function), so this *could* be a bug in Firefox. However, it seems that this bug is not reproducible in a minimal example app, so something else in Sandcastle may be contributing. 

### Reproduction steps

1. Run Sandcastle in Firefox v140 (locally or via [sandcastle.cesium.com](sandcastle.cesium.com))
2. Observe the bad layout. Compare to Chrome or Firefox v139.
3. To observe "fix," add quotes to the bucket.css import URL in one of the sandcastle gallerly html files and serve locally. 

### Environment

Browser: Firefox v144
CesiumJS Version: 1.130.1
Operating System: MacOS Sequoia 15.5 M4 Pro

jjspace:
> found that changing the import URL of the bucket.css file from:... quotes added

Interestingly changing it to `@import url(..\/templates\/bucket.css);` also works. I don't know why yet though. I'm more suspicious again that it's not on Firefox's side but instead on our processing of the code value or the way we're passing it into the iframe.

Edit: Even just adding spaces around it makes it work... 😕 `@import url(../templates/bucket.css );`

I tried to create a minimal reproduction but wasn't able to thus far...


~~(It definitely doesn't help debugging that CSS imports [do not show in the network tab](https://bugzilla.mozilla.org/show_bug.cgi?id=1650507))~~

jjspace:
Ok, now I'm mega confused. It seems like it's something about the _extremely specific_ format of the style/import tag that's there by default. When changing _anything_ about the style block it works as expected (see video). But setting it back to the default form it doesn't work. It also doesn't work with the sandcastles that add extra styles like the `3D Models Coloring` one. However, again, when changing the style tag at all it starts to work again... 😞 

[...]

PR #12704, opened 2025-06-30
Add hack to avoid Firefox styling bug

jjspace:
# Description

This is, by all means, not an ideal solution. _Something_, that I am unable to determine, is causing Firefox v140 to fail to load the imported `bucket.css` stylesheet. The network request _just doesn't happen_ after constructing the stylesheet.

Adding a random space (` `) character somewhere in the style tag seems to make it work for whatever reason. I chose to automate this and add it after the `@import` rule we can expect to exist.

More details and discussion about investigation and attempts to reproduce is in #12700.

So far this does not seem to be an issue in the new version of Sandcastle (#12566) so this "hack" should only have to exist until that's done.

## Issue number and link

Hack for #12700 

## Testing plan

- Run `npm start`
- Load [Sandcastle locally](http://localhost:8080/Apps/Sandcastle/index.html) in Firefox v140 and make sure it works
- Make sure clicking "Open in new window" opens the standalone page and it still works
- Verify Chrome and other browsers still work as well

# Author checklist

### DIFF
--- a/Apps/Sandcastle/CesiumSandcastle.js
+++ b/Apps/Sandcastle/CesiumSandcastle.js
@@ -618,7 +618,10 @@ require({
 
     // Apply user HTML to bucket.
     const htmlElement = bucketDoc.createElement("div");
-    htmlElement.innerHTML = htmlEditor.getValue();
+    // Stylesheet imports are weirdly broken in Firefox 140. This is a hacky workaround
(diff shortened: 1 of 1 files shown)


ggetz (approved):

ggetz:
Thanks for the fix @jjspace! Given we want a fix for today's release and the overall intention is to replace the current Sandcastle app code, this hack will be sufficient.

Issue #12716, opened 2025-07-02
Transparent flicker when assigning new image to a material


mzschwartz5:
When assigning a new image to a material, the primitive the material belongs to flashes out of existence briefly (perhaps just one frame). 

The issue seems to trace to the fact that updating an image actually invokes a chain of events leading to the whole primitive being destroyed and replaced. This occurs in the `StaticGroundGeometryPerMaterialBatch`, which supposedly tracks the "old primitive" and only removes it when the new one is ready, but this doesn't work well.

For a deeper dive into the issue, see discussion on [this thread](https://github.com/CesiumGS/cesium/issues/1640#issuecomment-3010962436) in #1640 

[...]

Issue #1640, opened 2014-04-24
Image material flashes white


shunter:
Because the Image material loads images from strings itself internally, it always renders the default white texture until the texture is ready.

This affects CZML too.

Forum question: https://groups.google.com/forum/#!topic/cesium-dev/GfN8YlQkQio

adamish:
Any news on this, alternative workarounds? I'm hitting this issue too...

pjcozzi:
I haven't tested this, but try changing [this line](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Renderer/Context.js#L870) to
```javascript
arrayBufferView : new Uint8Array([255, 255, 255, 0])
```
This should work in cases when alpha blending is enabled (which may not be all cases).

mplussier:
This last solution works in the sense that we don't see the white texture appearing anymore. However the geometry still flickers (it disappears and then reappears) every time its texture is changed. The solution proposed by mramato would be great for my needs: 

> initially we could not show anything until the first image is loaded and then if the image is changed, we can keep the old image until the new one is loaded.

Another idea would be to have a flag to load the texture synchronously. Thanks

hpinkos:
Updated code example
``` javascript
var czml = [{
    "id" : "document",
    "name" : "Basic CZML billboard and label",
    "version" : "1.0"
}, {
    "id" : "0",
    "polygon" : {
        "material" : {
            "image" : {
[...]

Issue #12714, opened 2025-07-02
Consolidate Batches in DataSources


javagl:
The [`DataSources`](https://github.com/CesiumGS/cesium/tree/82b2c07064997e5d197fb71293de2320ad947284/packages/engine/Source/DataSources) directory contains several classes:

- `StaticGeometryColorBatch`
- `StaticGeometryPerMaterialBatch`
- `StaticGroundGeometryColorBatch`
- `StaticGroundGeometryPerMaterialBatch`
- `StaticGroundPolylinePerMaterialBatch`
- `StaticOutlineGeometryBatch`

Each of these classes has approximately 500 lines of code. 

Here is a screenshot of comparisons between `StaticGroundGeometryPerMaterialBatch` and `StaticGroundPolylinePerMaterialBatch`, `StaticGroundGeometryPerMaterialBatch` and `StaticGeometryPerMaterialBatch`, and `StaticGeometryColorBatch` and `StaticGroundGeometryColorBatch`:

![Image](https://github.com/user-attachments/assets/90c69132-d99a-4642-a417-12cff9d027cf)

Each class is sharing most of the code with each other class (with the main differences often just being the _names_ of the class and variables). It is not even clear what the overarching purpose of these classes is, and even less what the _differences_ between the classes are. I'm reasonably sure that some of the differences are just the results of certain bugfixes going into _one_ class but not into the others. And I'm reasonably sure that future bugfixes will only go into _one_ of these classes and not into the others, because nobody would dare to touch these classes, out of the fear of breaking something. 

As usual, the classes basically contain _zero_ documentation, and I could only hazard a guess how much time has already been spent by people trying to wrap their head around what the purpose of flags like `waitingOnCreate` is, which appears in half of the classes, or what the `id` of an `id` is in lines like
`let attributes = this.attributes.get(instance.id.id);`
which appears _twice_ in _each_ of these classes. 

High-level work plan:

- Identify the purpose of these classes 
- Move common elements into a base class
- Identify the differences between these classes
- Find suitable abstractions for handling the differences

Jean-Dum:
Hello! I think this update would be really great to prevent some issues to be fixed in some of those classes but not in other ones (like #12722, which is only fixed in `StaticGroundGeometryPerMaterialBatch` for now).

PR #12722, opened 2025-07-07
Defers batch destruction until next update

mzschwartz5:
# Description

Previously, materials would flash (both transparent, then white) when changing a simple property (for example, the `cellAlpha` in a `GridMaterial` or the image in an `ImageMaterial`). This doesn't look good and was especially problematic for users doing animations.

The root cause is discussed in much more detail in the issue linked below, but is summarized in [this comment](https://github.com/CesiumGS/cesium/issues/1640#issuecomment-3029064087). In one line: Material batches get destroyed prematurely, causing the recreation of materials and primitives and circumventing the logic that ensures a seamless swap out.

There are a [couple ways one might fix this](https://github.com/CesiumGS/cesium/issues/1640#issuecomment-3029064437), but I believe the most effective is to simply defer Batch destruction until `Update`, rather than doing it immediately (this means that, if a material gets re-added to a Batch in the same frame, as happens when a material property changes, the Batch sticks around and so does the Material).

## Issue number and link

Originally discovered here: #1640
Also fixes: #12716

## Testing plan

Use the following sandcastles. For image-type materials, go to Chrome devtools and disable caching (and optionally turn throttling to Slow4G). Previously, the white default texture would show between every image change. Now, it seamlessly switches from the previous image once the new one is loaded.

Also (harder to observe / independent of caching), previously between every image change, a transparent frame would show. If you toggle quickly on the canvas example below, you can see the terrain flashing through for a frame each time.

[Simple image material](https://sandcastle.cesium.com/#c=fVPLjtMwFP0V01UqtXbSdEJpOoWhIIQEYqSp2BCE3OQ2MePYke20kxl1z5Yv4Bf5BJzXTAsVO5/7OPY59zqWQhu0Y7AHhS6RgD1agWZljj83MScaxA1eSWEoE6CiwQg9RAIhncmSJ1eC5dTAHBlVwigSh2EYiUjEDa9NpeBZ3miQGVPoOSFlwSVN8J7dshwSRrFUKalRUSMSyzy3rcRkZb4hAQmAvJM8Ad+fTn38vUhJMHWLu/FpMBqEJ1dOTq6ME4ELdkc3tMKWnxSZNJJMXG9K3IC4PvFeEH9GNpKqZOwH3oUXfPMmM/eRORIcDCo1vO/11GKPdBaSV6kUb4VhprLp1k8MNWagMU0Sp/GsK5y3DiKUMVBUxVk1721fUWXsiQofb5XM30CqAPSVUrRyvrRNCI2fX+BghKYudqejM0H/JHgx+1+0J/g67A71OBWjfH68DY3yj13mWskClKmch56wcX3ezbvjObSEh6el6Gy5lzJfS+fEtGGICEG/f/38geq0RkYik0FvWN19Q0USU2041H6upeQbql6Xxkhht3Qt05QDum7rUfPeelW3pYgNsxFn2Jp+PMZnjyA8mk77Itwh3PuBG3W264nhZb/hnfRJ2GkdjAYLbSoOy96hVywvpLJbpLiDMTGQF9wSa7Ip41swONZ6GLbFC3LcukjYDrHk8sxHRDGnWtvMtuT8ht1bwcsFsfX/tNZ/jon00w4Up1VdlnnLD20QY7wgFp7vNK3PfzH/AQ)

[Canvas image](https://sandcastle.cesium.com/#c=bVTbbtswDP0VLi92AFdpsnYrctuwrA8DNmxogz3lRbVpR5gsGZKcS4v8+yjLbp22QOCA1Dnk0SHtVCvrYCdwjwYWoHAPK7SiLtnfJhdvBmkTr7RyXCg0m8FwtlEbNRrByiB3CG6vIeVqxy1a2Au3hUzkORpUDlIttcEMKu4cGmU3Kq9V6oRWkDbsVUOMG1wCDg9uCE8bBcT0ykJdUpbptC6pIgu0W4k+iqMAiLwmaNFsLzISsYDJ9ad+eoui2Lp+PrRwB8q1mAKdvynJiKNJ1pV1B5YLKe/dUaLHerFnJ3eYuvgyAfr1NSTnrd+vFu23wmH0ckb9ffrqpjqA5cpeWDQij87Ia6/Qy0xgQk3HV5ehuEFXG9W2pczJj6pv5djr7zsf0XiiBKIQwri5c58xecN4kDX2KJMobIREB7VtUb6PMzXOXgQYMomrQuKtcsIdCRD2jqGPBVrGsyx+CtdoodOwDH5W2mRCkQo77Vb0rkOx3OjyOxYG0cYXN5eMLPl47Z8Xn5u/K0oNk1CppBpGcDntb/uPkhf4qz35Y3SFxh3jtjeA8MfTzsKkSxusyJezQituSKLgahKPaS7DZ6wzNMqK+7diCjmXFtujU8Cc6HkKRra2PGpdrnX8yrYAuecqS7l1dHcyba21fODmW+2cVnG01gXBIQyCBhUPYbEMRp7N58NLNDtzPXRizzHrPGONEUTt1fnyvFmdQZNZe5VBMphbv+bLzoWvoqy0oT0xMmZs5LCspJ/p6KFO/yG93daGRQaYj/rUeSZ2ILLFO98jSCW3lk7ymt4q8YibwXI+IvwbqtScdqj4vUMj+dHDtuPlz5BkjM1HFL7PdMHhV5X/Aw)

[Grid type material, two entities go from being in 1 batch to being split into 2](https://sandcastle.cesium.com/#c=3VTdbtMwFH6VQ69SqXPSbJOg7QajDIS0aYhGIES48JLTxsyxK9tZ6dAkHoIn5Ek4cZKu3cYd4oKbJOfnOz+fvzjTyjq4FrhCA0egcAVTtKIq2QfvC9Je5u2pVo4LhSbt9cepSpVEB3YlXFZgTsg5lxZ9IPMlS+7QCC6Hu1XfGJGft6F3Ri/RuHXwPVUAmZbajLq8aW2xT6dnZxcfGXUpTuSy4MGQRf2Bz0YpvWsE5PMuSdNNdaXcaLvhlBtHX1zFwcEADvqb1KQQ2ZVCa/+UHlNdiJuGt83OYQgzXiJwu7XfHpwk8Prt+1nCIFlpoMnE0qKFlZASFkTTJW9YcpqsAg3lFahgVT+ygqsFbsrFA4+wuOSGXC3Usvu0xv8TrZ7YU+WEW8Ow2xS9XcunUSfzDkFc8DxvdltqSx6tRg/a7LO50eUrXBhEG+wNo6juehB1e7ZnNAJfB4jvUpzzr9qcfBM0+WEUecRWUKguuL8T7A5kdKcIH7ndUU27XAy/fvwEW3BD8iAlgK3V1AFBKFqHS7ne5SD+Kxw8e/rvKIjvU9COf6N1mejgc3u2g27BLxuiEq3lJTf1r7KQSK/2jtnIgTzcK2vDWi1jyMV8joaS7iScqhlXecatk1jz1ZZ+WTmnFd1rSdNiSvngAaDVnQp7A5hXKquphaDfkLR13T3pvseey2Y41nLKutnYZhiCbNDPIWKH4P+wcctQb9CbWLeWeNxQCvBClEttHFRGBoyFDsulpKo2vKyyK3Qss7Y/bpIn4TZ0kotrEPnRIxc3ZJJbS5F5JeVM3GDaO56ElP8AKjXPhVpcXKORfF2nFcPjs8bJGJuEZD6OdA3L9yr/Bg)

[Billboards should be unaffected by this change](https://sandcastle.cesium.com/#c=bVLLbtswEPwVVicbUEjJ8iuy4z7coihQIIcEOQkoaHItsaFIgaTsukX+vaRkO3abm2a4M9qdXaaVdWgnYA8G3SEFe7QGK9oaP3XcoIhYh9daOSoUmCKK0Z9CIWQr3Ur+UYmaOsiRMy3EhXoZLgpVKNb5+qcSUu9bRJVzjc0JaRupKcd78Sxq4IJibUoSUBMQYbquvZRMyRTIVy05ZNl4nOGfTVlEiyvf0ZUv4wo34hfd0AP2JqSptNNklKRjkkxJkpH0lmRzstHU8Jtsmk7S6Y90NE/OzoWS4FBr4dup6TDRxTAbIWUn/6KccAdf0MeGIWABFlPOB100jbae0So/hbmmxvkvqjK8Nbr+DKUBsIOb2QRPbmezWYzGCU6y+TwbxsHg/K+8zxr1I+fHROOes4xKz6U4ORJ7wV2Vo0lyIioQZeU6JhAvrwt6oIozap2E0Paj1nJDzafWOa38zh91WUpAXRRh4dtWsTAQGgz7hi5zencGi6ve+5zwGeOueS94Fb8/nchxstHi2GAUR0vrDhJW/SAIfRB1o43fkJEDjImDupH+8CzZtOwZHGbWDhd98ZJcSpdc7JDgd29cMmKSWutftq2UD+K3n3W1JL7+P2k4WqHK+x0YSQ+hrEpX33sSY7wkHr6tdH20/zj/BQ)

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 ### Fixes :wrench:
 
-- Fixes material flashing when changing properties [#12722](https://github.com/CesiumGS/cesium/pull/12722)
+- Fixes material flashing when changing properties [#1640](https://github.com/CesiumGS/cesium/issues/1640), [12716](https://github.com/CesiumGS/cesium/issues/12716)
 
 #### Additions :tada:
 
--- a/packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
+++ b/packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
@@ -66,6 +66,7 @@ Batch.prototype.isMaterial = function (updater) {
  * @param {JulianDate} time
  * @param {GeometryUpdater} updater
  * @param {GeometryInstance} geometryInstance
+ * @private
  */
 Batch.prototype.add = function (time, updater, geometryInstance) {
   const id = updater.id;
@@ -101,6 +102,7 @@ Batch.prototype.add = function (time, updater, geometryInstance) {
  * on the next update.
  * @param {GeometryUpdater} updater
  * @returns true if the updater was removed, false if it was not found.
+ * @private
  */
 Batch.prototype.remove = function (updater) {
   const id = updater.id;
@@ -127,6 +129,7 @@ Batch.prototype.remove = function (updater) {
  * A new primitive is created whenever an updater is added to or removed from a Batch.
  * @param {JulianDate} time
  * @returns a boolean indicating whether the Batch was updated.
+ * @private
  */
 Batch.prototype.update = function (time) {
   let isUpdated = true;
@@ -305,6 +308,7 @@ Batch.prototype.getBoundingSphere = function (updater, result) {
 
 /**
  * Removes a Batch's primitive (and oldPrimitive, if it exists).
+ * @private
  */
 Batch.prototype.destroy = function () {
   const primitive = this.primitive;
@@ -340,7 +344,7 @@ function StaticGroundGeometryPerMaterialBatch(
  *
  * @param {JulianDate} time
  * @param {GeometryUpdater} updater A GeometryUpdater that manages the visual representation of a primitive.
- * @returns {void}
+ * @private
  */
 StaticGroundGeometryPerMaterialBatch.prototype.add = function (time, updater) {
   const items = this._items;
@@ -384,6 +388,7 @@ StaticGroundGeometryPerMaterialBatch.prototype.add = function (time, updater) {
 /**
  * Removes an updater from a Batch. Defers potential deletion until the next update.
  * @param {GeometryUpdater} updater A GeometryUpdater that manages the visual representation of a primitive.
+ * @private
  */
 StaticGroundGeometryPerMaterialBatch.prototype.remove = function (updater) {
   const items = this._items;
@@ -402,6 +407,7 @@ StaticGroundGeometryPerMaterialBatch.prototype.remove = function (updater) {
  * Updates all the items (Batches) in the collection, and deletes any that are empty.
  * @param {JulianDate} time
  * @returns a boolean indicating whether any of the items (Batches) were updated.
+ * @private
  */
 StaticGroundGeometryPerMaterialBatch.prototype.update = function (time) {
   let i;
--- a/packages/engine/Specs/DataSources/StaticGroundGeometryPerMaterialBatchSpec.js
+++ b/packages/engine/Specs/DataSources/StaticGroundGeometryPerMaterialBatchSpec.js
@@ -504,7 +504,7 @@ describe("DataSources/StaticGroundGeometryPerMaterialBatch", function () {
    * because doing so caused a batch to be destroyed and recreated prematurely which, in turn, caused the associated
    * material (and primitive) to be destroyed and recreated.
    */
-  it("doesn't immediately remove an empty batch that may be reused before the next update", function () {
+  it("doesn't immediately remove an empty batch that may be reused before the next update", async function () {
     if (
       !GroundPrimitive.isSupported(scene) ||
       !GroundPrimitive.supportsMaterials(scene)
@@ -533,35 +533,32 @@ describe("DataSources/StaticGroundGeometryPerMaterialBatch", function () {
     const updater = new EllipseGeometryUpdater(entity, scene);
     batch.add(time, updater);
 
-    let materialBefore;
-    return pollToPromise(function () {
+    await pollToPromise(function () {
       scene.initializeFrame();
       const isUpdated = batch.update(time);
       scene.render(time);
       return isUpdated;
-    })
-      .then(function () {
-        // Remove and add back the updater before the next batch update
-        // Removal should defer deletion of the batch, and the updater should find the old batch still works upon re-adding
-        // Thus, the material should not be destroyed and recreated (and will have the same guid / be the same object)
-        materialBefore = scene.groundPrimitives.get(0).appearance.material;
-        batch.remove(updater);
-        batch.add(time, updater);
+    });
 
-        return pollToPromise(function () {
-          scene.initializeFrame();
-          const isUpdated = batch.update(time);
-          scene.render(time);
-          return isUpdated;
-        });
-      })
-      .then(function () {
-        const groundPrimitives = scene.groundPrimitives;
-        const materialAfter = groundPrimitives.get(0).appearance.material;
-        expect(materialBefore).toEqual(materialAfter);
-        expect(groundPrimitives.length).toEqual(1);
+    // Remove and add back the updater before the next batch update
+    // Removal should defer deletion of the batch, and the updater should find the old batch still works upon re-adding
+    // Thus, the material should not be destroyed and recreated (and will have the same guid / be the same object)
+    const materialBefore = scene.groundPrimitives.get(0).appearance.material;
+    batch.remove(updater);
+    batch.add(time, updater);
 
-        batch.removeAllPrimitives();
-      });
+    await pollToPromise(function () {
+      scene.initializeFrame();
+      const isUpdated = batch.update(time);
+      scene.render(time);
+      return isUpdated;
+    });
+
+    const groundPrimitives = scene.groundPrimitives;
+    const materialAfter = groundPrimitives.get(0).appearance.material;
+    expect(materialBefore).toEqual(materialAfter);
+    expect(groundPrimitives.length).toEqual(1);
+
+    batch.removeAllPrimitives();
   });
 });


packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
```diff
@@ -23,35 +23,25 @@ function Batch(
   usingSphericalTextureCoordinates,
   zIndex,
```
mzschwartz5, line 24: FYI: most of the changes here are just me adding comments / documentation.

Also, there are ~10 other `Static*Batch` classes that are almost identical... we should discuss if we want to apply these changes to those other files right now (#12714)

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
```diff
@@ -361,35 +381,37 @@ StaticGroundGeometryPerMaterialBatch.prototype.add = function (time, updater) {
   items.push(batch);
 };
```
mzschwartz5, line 387: This is where the real changes are. They're minimal:
1. Defer destroying a batch (`item`) until the `update` call.
2. Remove the invalidated check (which is redundant, as explained in above comment)

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
```diff
@@ -23,35 +23,25 @@ function Batch(
...
   this.geometry = new AssociativeArray();
   this.material = undefined;
   this.updatersWithAttributes = new AssociativeArray();
   this.attributes = new AssociativeArray();
-  this.invalidated = false;
-  this.removeMaterialSubscription =
-    materialProperty.definitionChanged.addEventListener(
-      Batch.prototype.onMaterialChanged,
-      this,
-    );
```
mzschwartz5, line 43: The `invalidated` flag gets set when a material property changes via the event listener here. However, it's actually completely redundant, so I've removed it (and associated logic, later in the file). 

The redundancy comes from another material property listener, which propagates the event up all the way to the `DataSourceDisplay` class, where it then propagates back down in `Update` calls to `GeometryVisualizer` -> `GeometryUpdaterSet` -> `Static*Batch`

packages/engine/Specs/DataSources/StaticGroundGeometryPerMaterialBatchSpec.js
```diff
@@ -51,11 +53,8 @@ describe("DataSources/StaticGroundGeometryPerMaterialBatch", function () {
       return;
```
mzschwartz5, line 53: Two things:
1. Most changes in this file are just changing `scene.primitives` to `scene.groundPrimitives`, as that's what this spec should really be testing after all.
2. The first test relied on that `invalidated` flag mechanism. In order to test the higher-level listener mechanism I mentioned, I had to change the test to use higher-level constructs (i.e. instead of creating a `StaticGroundGeometryPerMaterialBatch` directly, I replaced it with a `GeometryVisualizer`, which creates the batch itself, and hooks up the proper listener). This does feel like more of an integration test now, so I'm open to feedback on how to improve it.

ggetz:
Awesome, thanks @mzschwartz5! In the first example under "Testing plan," I can clearly see the improvement.

To confirm, I should be seeing no differences between `main` and this branch for the other examples. Correct?

CHANGES.md
```diff
@@ -1,5 +1,15 @@
 # Change Log
 
+## 1.132 - 2025-07-02
+
+### @cesium/engine
+
+### Fixes :wrench:
+
+- Fixes material flashing when changing properties [#12722](https://github.com/CesiumGS/cesium/pull/12722)
```
ggetz: This is totally a nit-picky comment: Typically for fixes we link to the fixed issue (rather than the PR) to help users who may be tracking it.

But, ultimately, everything is linked together in GH so no big deal either way.

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
ggetz: What would be a rough level of effort estimate to get the other classes updated similarly? If it's not too much overhead, keeping these in sync would be helpful.

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
```diff
@@ -71,6 +61,12 @@ Batch.prototype.isMaterial = function (updater) {
   return defined(material) && material.equals(updaterMaterial);
 };
 
+/**
+ * Adds an updater to the Batch, and signals for a new Primitive to be created on the next update.
+ * @param {JulianDate} time
+ * @param {GeometryUpdater} updater
+ * @param {GeometryInstance} geometryInstance
```
ggetz, line 68: When adding documentation to `private` class members (private here meaning anything not part of the public API) make sure to add the `@private` tag. Otherwise, even if the top-level class is already marked as private, it can show up in the generated reference documentation. (Since `Batch` is not exported by this file, I don't think it will show up. But good to be explicit and get in the habit.)

```suggestion
 * @param {GeometryInstance} geometryInstance
 * @private
```

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
```diff
@@ -305,23 +316,32 @@ Batch.prototype.destroy = function () {
...
   this._appearanceType = appearanceType;
 }
 
+/**
+ * Adds an geometry updater to a Batch. Tries to find a preexisting compatible Batch, or else creates a new Batch.
+ * Used by Visualizer classes to add and update (remove->add) a primitive's Updater set.
+ *
+ * @param {JulianDate} time
+ * @param {GeometryUpdater} updater A GeometryUpdater that manages the visual representation of a primitive.
+ * @returns {void}
```
ggetz: `return(s)` can be omitted if the function return type is `void`.

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
ggetz: As part of that, we should also consider of there is enough differences between these classes to warrant them being different files.

mzschwartz5:
@ggetz with the first example under Testing Plan, the improvement is super clear and easy to test with dev tools caching disabled / slow 4g.

With the second example (canvas material), there IS still a difference between this branch and main, but it's less obvious. If you zoom in on the canvas and toggle the color quite quickly, you'll see the terrain below flashing through for one frame each time (on main). On this branch, no more transparent flashing occurs.

In theory, the same is true of the third example, but because the material property being changed is the alpha, you'll see the terrain flashing through regardless. That example was mostly useful for breakpoint debugging / testing that two entities still get separated into different batches when one's material changes. (i.e. no breaking changes)

The fourth example *should* be the same on main and on this branch. Nothing I did should have affected billboards, hopefully!

packages/engine/Specs/DataSources/StaticGroundGeometryPerMaterialBatchSpec.js
```diff
@@ -501,4 +498,70 @@ describe("DataSources/StaticGroundGeometryPerMaterialBatch", function () {
...
+    batch.add(time, updater);
+
+    let materialBefore;
+    return pollToPromise(function () {
+      scene.initializeFrame();
+      const isUpdated = batch.update(time);
+      scene.render(time);
+      return isUpdated;
+    })
+      .then(function () {
```
ggetz: No need to change existing tests, but we now prefer async/await syntax instead of promise chains, i.e., 

```js
await pollToPromise(...);

const materialBefore = scene.groundPrimitives.get(0).appearance.material;
batch.remove(updater);
batch.add(time, updater);

await pollToPromise(...);

// etc...
```

ggetz (approved):

packages/engine/Specs/DataSources/StaticGroundGeometryPerMaterialBatchSpec.js
mzschwartz5: Was wondering about that, makes total sense, I can update that usage.

Also hijacking this comment to ack your others as well; I'll include changes for those in my next revision.

packages/engine/Source/DataSources/StaticGroundGeometryPerMaterialBatch.js
mzschwartz5: They probably *shouldn't* be different files; I think this is why @javagl made #12714. The differences are very minimal. If we just want to apply these changes to those files, the main effort would be testing. 

If we want to consolidate these various files/classes, that probably wouldn't be much harder, but the changes would definitely extend a bit into other classes that reference them. Probably more risk, would need to be more careful not to break anything.

javagl (commented):
I did also spend a bit of time to investigate the original issue, and had a closer look at some code paths of which I _thought_ that they may be relevant. But I focussed on a different (and probably "wrong") point in these paths: The white material that is shown there is the "default" material, and I zoomed to the point at https://github.com/CesiumGS/cesium/blob/82b2c07064997e5d197fb71293de2320ad947284/packages/engine/Source/Scene/Material.js#L865 where this was set, and then tried to ~"work backwards through the stack trace", trying to figure out why that `texture` was undefined there. 

This didn't lead to substantial insights. As laid out in the issue comments: The image change triggers a whole cascade of things, going up and then down the hierarchies, and figuring out "one place" where something is "definitely wrong" can be difficult in such cases.

---

In terms of a review, I had a look at the changes, and (with the usual disclaimer that there are many undocumented legacy classes of which the overarching purpose and lifecycle is not always clear) they look reasonable: From my understanding (a bit imprecise and summarized), the handling of this `oldPrimitive` was not correct in the current 'main'. It was _supposed_ to be kept alive until the new primitive was ready, but due to unclear state spaces and lifecycle management, it was destroyed _before_ the new one was ready. With the new state, it is kept alive until the new one is ready. (There is a certain primitive collection that actually contains _two_ primitives for a short period of time now)

---

Regarding the fix itself and the testing plan: 

I already checked the new state with the _first_ test case ("Simple Image Material"), and it does fix the issue.

I'm not sure what the intention is behind the other test cases. I still do see the primitive disappearing for a moment in these cases:

Cesium Flicker 0001

If the intention was only to show that there are no "regressions", then I can say that this is the same behavior as in the current 'main'. 

---

I had a short look at the updated/added test cases, but have not yet fully understood what they _originally_ tested to begin with. (I also did not try out whether the new test case _fails_ in 'main' and passes in this branch, and am not sure whether this is supposed to be the case, but I can try it out if this is desired).

---

The point that was brought up in the comments - namely, whether there are other `*Batch` classes that may have to be updated - is related to https://github.com/CesiumGS/cesium/issues/12714 . As I mentioned there, it is very likely that some of the subtle differences between these classes already _are_ bugfixes that only went into one (but not all) of them. For example: Some of these classes contain a flag that is called `waitingOnCreate` which, judging by its name, _sounds_ like it could be an attempt to achieve what now has been done in _this_ PR (namely: wait with the destruction of the _old_ primitive until the _new_ primitive is ready). That's a wild guess, of course. There are no comments.

Determining the differences of these classes, finding out which of them are _intended_ differences, and how these differences interrelate with the changes _here_ is likely beyond the scope of this PR.

mzschwartz5:
Huh, the canvas test isn't supposed to show a transparent flash. I feel like it was working just yesterday, but I tested again just now and I get the same behavior... will take a look into that.

javagl:
Even if that canvas case is not fixed (maybe because it uses some other `*Batch` class?), then this PR will still fix the case of the standard image material, and have no (obvious) regressions, so that does not _necessarily_ have to hold up this PR. (Others have to decide that, of course)

mzschwartz5:
I updated the sandcastle link for the Canvas example. The old link was creating a new material each time the toggle was pressed --> new Batch --> flicker. What it was *supposed* to do was just change the image on the existing material --> same Batch --> no flicker.

Should work now (on this branch)

CHANGES.md
mzschwartz5: Interesting, the last dozen or so links in CHANGES.md are to PRs - I guess we lost our way, a bit! I'll change it to the issue #

ggetz:
Changes here are looking good to me! I am able to see the fix in the second example now. Other test cases are consistent as expected.

Thanks for tracking these down @mzschwartz5!

PR #12723, opened 2025-07-07
Adds configuration for breakpoint debugging Karma tests in IDE

mzschwartz5:
# Description

Added VSCode launch config for attaching to Karma test runner to breakpoint debug tests from IDE.

Minor changes to Gulpfile to support this.

## Issue number and link

No issue

## Testing plan

1. Run a test with the debug flag (single suite for simplicity): `npm run test-suite DataSources/StaticGroundGeometryPerMaterialBatch --debug` 
2. Run the `Attach to Karma` debug launch configuration from VSCode
3. Place a breakpoint in one of the Spec's tests
4. Click `Debug` in Chrome and observe it stop on breakpoint

# Author checklist

- [] I have added my name to `CONTRIBUTORS.md`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,6 +21,13 @@
             "type": "chrome",
             "url": "http://localhost:8080",
             "webRoot": "${workspaceFolder}"
+        },
+        {
+            "request": "attach",
+            "name": "Attach to Karma",
+            "type": "chrome",
+            "port": 9333, // This is the remote debugging port specified in karma.conf.cjs
+            "webRoot": "${workspaceFolder}"
         }
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,26 @@
             "name": "Attach to Karma",
             "type": "chrome",
             "port": 9333, // This is the remote debugging port specified in karma.conf.cjs
-            "webRoot": "${workspaceFolder}"
+            "webRoot": "${workspaceFolder}",
+            "timeout": 60000
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Test Suite",
+            "cwd": "${workspaceFolder}",
+            "runtimeExecutable": "npm",
+            "runtimeArgs": ["run", "test", "--", "--includeName", "${fileDirnameBasename}${/}${fileBasenameNoExtension}", "--debug"],
+            "console": "integratedTerminal"
+        }
+    ],
+    "compounds": [
+        {
+            "name": "Launch Test Suite and Debug in VSCode",
+            "configurations": [
+                "Launch Test Suite",
+                "Attach to Karma"
+            ]
         }
     ]
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -999,7 +999,7 @@ export async function test() {
   const release = argv.release ? argv.release : false;
   const failTaskOnError = argv.failTaskOnError ? argv.failTaskOnError : false;
   const suppressPassed = argv.suppressPassed ? argv.suppressPassed : false;
-  const debug = argv.debug ? false : true;
+  const debug = argv.debug ? true : false;
   const debugCanvasWidth = argv.debugCanvasWidth;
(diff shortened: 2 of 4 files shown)


gulpfile.js
```diff
@@ -1017,7 +1017,7 @@ export async function test() {
     });
   }
 
-  let browsers = ["Chrome"];
+  let browsers = debug ? ["ChromeDebugging"] : ["Chrome"];
```
mzschwartz5, line 1022: "ChromeDebugging" is a custom launcher defined in `karma.conf.cjs` that opens up a remote port for debugging (9333). I don't think it's been used formally, until now. But if you're wondering where this comes from, that's where to go look.

.vscode/launch.json
```diff
@@ -21,6 +21,13 @@
             "type": "chrome",
             "url": "http://localhost:8080",
             "webRoot": "${workspaceFolder}"
+        },
+        {
+            "request": "attach",
+            "name": "Attach to Karma",
```
jjspace, line 27: It feels like there's a lot of steps/hoops to jump through to get tests up and running. Is there a way to have this launch the tests first so it's one button press? bonus points if there's a version that only runs the file you currently have open.

Also is there a way to open directly to the debug tab/window instead of having to click it in the browser window?

mzschwartz5: AFAIK there's no way to access the current file from a launch or task file. You *can* prompt the user for a string though (the name of the test suite). And you can make a compound launch config that, say, asks the user for a test suite, runs the npm test command with that string, and then runs the launch config to attach to karma -- all in one go. The nice thing here is that, if you run it multiple times, it autofills your input from the last time.

The downside is, it adds quite a bit to the launch config file and adds several more tasks to the drop down:
image

image

However, it does preserve the original way of attaching, for those who prefer launching tests via command line.

Thoughts?

mzschwartz5: (Will also mention - as this commit stands, it's just one more button click than debugging through the browser. The only extra step is clicking "Attach to Karma" in VSCode after running the test command)

mzschwartz5: Also, if you cancel out of the input window, it still tries to run the command with a nonsense input, which is a little annoying and there seems to be no easy way to abort in this case.

jjspace: You definitely can get the current file to replace in any of the commands in tasks or debug configs which would be better than prompting the user. https://code.visualstudio.com/docs/reference/variables-reference
It may take some more modifications to the karma setup to have it take in a filename instead of a suite name though.

If it's too much hassle then no need to overcomplicate

mzschwartz5: Huh, you're right - I've been lied to! (about accessing file names)
I can get *pretty* close with this:
```
            "runtimeArgs": ["run", "test", "--", "--includeName", "${fileDirnameBasename}${/}${fileBasenameNoExtension}", "--debug"],
```

Which would yield something like `Scene/PolylineCollectionSpec` ... but I don't think you can trim off `Spec` in the launch file (at least not easily). I *could* modify the gulp file's `test()` function so that after:

`  const includeName = argv.includeName ? argv.includeName : "";`

I trim `"Spec"` off the end, if it exists. Then this would work.

ggetz:
@mzschwartz5 Is this ready for re-review? It's helpful if to leave a new comment when it is.

mzschwartz5:
> @mzschwartz5 Is this ready for re-review? It's helpful if to leave a new comment when it is.

@ggetz yep sorry - I had pinged @jjspace offline, but didn't comment on the PR so it likely got lost.

jjspace (approved):
Seems to work well enough. Not totally sure how I feel about replacing `Specs` in the path but it seems to work and probably will for most if not all our specs. I honestly am not sure I'll really use this so if it's helpful for those that will I'm happy to merge it.

mzschwartz5:
> Not totally sure how I feel about replacing Specs in the path...

FWIW, it only replaces `Specs` if it appears at the end of the path string. So, slightly more robust. Of course, it also only does this for the test build configuration.

PR #12727, opened 2025-07-09
Update `gltf-pipeline` 4.2.0

javagl:
# Description

Updates the `GltfPipeline` directory contents with the state of [`gltf-pipeline` 4.2.0](https://www.npmjs.com/package/gltf-pipeline/v/4.2.0), as described in the [readme](https://github.com/CesiumGS/gltf-pipeline?tab=readme-ov-file#building-for-cesiumjs-integration).

Several changes seem to be caused by minor `prettier` differences. 

The most important change of that new release is that it [removes the use of `defaultValue`](https://github.com/CesiumGS/gltf-pipeline/pull/672), which is no longer going to be part of CesiumJS in one of the next releases. Most of this has already been done manually _in CesiumJS_ via https://github.com/CesiumGS/cesium/pull/12507 , but now it's coming from the actual `gltf-pipeline` source.

This can still not be merged. Apparently, there was a change in https://github.com/CesiumGS/cesium/commit/59141130b72a7d43f3e30eb56e45ec99255edad3#diff-6a3b6ddf82583b7f0637f12718da327f21acabf8e5dac921f0fb13c0b9150869 that didn't find its way back into `gltf-pipeline`. So we'll have to do one more roundtrip here...

## Issue number and link

n/a

## Testing plan

n/a (Specs should cover it...)

# Author checklist

- [ ] **n/a** I have updated the inline documentation, and included code examples where relevant

### DIFF
--- a/packages/engine/Source/Scene/GltfPipeline/addToArray.js
+++ b/packages/engine/Source/Scene/GltfPipeline/addToArray.js
@@ -1,3 +1,5 @@
+
+
 /**
  * Adds an element to an array and returns the element's index.
  *
--- a/packages/engine/Source/Scene/GltfPipeline/ForEach.js
+++ b/packages/engine/Source/Scene/GltfPipeline/ForEach.js
@@ -80,7 +80,7 @@ ForEach.accessorWithSemantic = function (gltf, semantic, handler) {
               return value;
             }
           }
-        }
+        },
       );
 
       if (defined(valueForEach)) {
@@ -102,7 +102,7 @@ ForEach.accessorWithSemantic = function (gltf, semantic, handler) {
                 return value;
               }
             }
-          }
+          },
         );
       });
     });
@@ -124,7 +124,7 @@ ForEach.accessorContainingVertexAttributeData = function (gltf, handler) {
               return value;
             }
           }
-        }
+        },
       );
 
       if (defined(valueForEach)) {
@@ -143,7 +143,7 @@ ForEach.accessorContainingVertexAttributeData = function (gltf, handler) {
                 return value;
               }
             }
-          }
+          },
         );
       });
     });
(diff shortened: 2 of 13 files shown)


javagl:
The changes that had originally been done in CesiumJS have now been done in `gltf-pipeline` as well, via https://github.com/CesiumGS/gltf-pipeline/pull/676

So this PR now contains the state of the code that comes from the (not yet released) gltf-pipeline version 4.3.0

EDIT: Given that the changes to omit `defaultValue` also had been done in CesiumJS, the only changes here are now `prettier` changes, and a few cases where `gltf-pipeline` uses `{}` instead of `Frozen.EMPTY_OBJECT`.

ggetz:

packages/engine/Source/Scene/GltfPipeline/addToArray.js
```diff
@@ -1,3 +1,5 @@
+
+
```
jjspace, line 2: I think this is an artifact from the way the `gltf-pipeline` source is "transformed" for CesiumJS ESM syntax but it's a little annoying. No change needed now but maybe we can fix this upstream so files conform fully to `prettier`

javagl: Yeah, the formatting differences (which are now essientially the _only_ differences here) also struck me as a bit odd. Apparently, [`GltfPipeline` is excluded from prettier](https://github.com/CesiumGS/cesium/blob/394f70ba1859da73107321cc792be3f3e32325bf/.prettierignore#L28), and I don't see a strong reason do do this 🤷‍♂️ but there are so many things in limbo for `gltf-pipeline` that I didn't want to broaden the scope here.

jjspace: I _assume_ the reason it's excluded from `prettier` is that these files should match 1:1 with the built files from `gltf-pipeline`. Then merges like this would only highlight the actual changes. This is one of the other reasons I think these should be in a `ThirdParty` sub-directory

Issue #1552, opened 2014-03-13
Polygons being culled incorrectly in 2D


mramato:
I have a CZML file (which I can share) that loads a bunch of Polygons at an altitude above the earth.  It works great in 3D mode but the polygons disappear when in 2D.  They do show up in Columbus View when zoomed out, but get culled when you try and zoom anywhere near them.  I've scene similar things with other geometry types, so I'm not sure if this is a polygon specific issue or a problem with Geometry on the whole.

hpinkos:
Code example:
``` javascript
var viewer = new Cesium.Viewer('cesiumContainer');

viewer.entities.add({
    polygon : {
        hierarchy : Cesium.Cartesian3.fromDegreesArray([-115.0, 37.0,
                                                        -115.0, 32.0,
                                                        -107.0, 33.0,
                                                        -102.0, 31.0,
                                                        -102.0, 35.0]),
        material : Cesium.Color.RED,
        height: 3000000
    }
});
```

mzschwartz5:
Investigating this issue. Some immediate observations:
- As alluded to above, this is no longer a problem for CV, only 2D.
- If you remove the height field on the polygon, it does show up.
- Same issue is present for RectangleGeometry, with either `height` or even `extrudedHeight`

edit - *Interestingly*, the issue does not occur for any height value. E.g. In @hpinkos's example above, if you reduce the height from 3000000 to 1000000, the polygon *does* appear - until you zoom in a bit, then it gets culled again. 

If you turn on show frustums in the cesium inspector (see sandcastle in next comment), the culling seems to coincide with specific frustums.

mzschwartz5:
[...]

PR #12730, opened 2025-07-11
Bounding sphere fix for 2D primitives

mzschwartz5:
# Description
Previously, in 2D scene mode, polygons with a non-zero height would be culled inappropriately as the user zooms into the map. The root cause is described [in this comment](https://github.com/CesiumGS/cesium/issues/1552#issuecomment-3063275029) - in summary, the bounding volume calculation erroneously set the sphere's height to 0, which messed up the process of binning the polygon into the correct frustum partitions. 

## Issue number and link

#1552 

## Testing plan
[Use this sandcastle to test before and after](https://sandcastle.cesium.com/#c=rVNNb9swDP0rgi91gEy2ExRFUjdom+QwoF2Bpdtl3kGRWVuYLBmSnNYb+t8ryx81sg3b0PFgi4/kAx8JUim0QQcGj6DQBRLwiNagWVXgzw7zT6hz11IYwgSokyn6kQiENAUBtzKFZV+w6xG8W28/bGebRDxPzhORiJYegzDMMNCYpKnvSBAqJa8zKdASdUBjOQNFFM1rNJCviTL2RcQcPyhZbCBTAPpKKVL7X95F0SkOp2h+Zr+vNP9qA83sjTThmaOZv5Vm5mii/0NjpX2djIgKYkAxwkczllwq/HG7GWXlwLLcLFF0GjbWBp7/vNlur8Nah5X+3Ub7Bmz3oes+HA2hB2eLMbhY/A5riltoGEAv/kj79c2nbZcxCA8H4a+yqbsadwL2aLohONeG3R9nXO4Bp1Ca/B60ucrs9WhzD1YfE7bIqKpJ7gf4ZECkftdMC7bOe6FLoEaqW/bExOTcm3qxNjWHVa/ykhWlVAZVivsYBwaKklt5OthX9BsYTLVuum5S42BcGqfsgFh6kXhHN554iHKitY08VJzv2HdIvFUc2PyfSrkkKRPZ3QEUJ3WTlkermxbEGMeBdX9daaTke6KOmF8A). Do note that the before-behavior is easiest to reproduce with a single polygon, and it's also sensitive to the height of the polygon (lower heights seem more resilient to the bug)

After the change, no combination of heights or different polygons should result in inappropriate culling.

(I'm wary of potential side effects of this change, but I'm unsure what domains may be affected. If anyone has thoughts about how I can better test bounding sphere changes, leave a comment).

# Author checklist


CHANGES.md
```diff
@@ -1,5 +1,15 @@
 # Change Log
 
+## 1.132 - 2025-07-02
```
ggetz: This correspond to year-moth-day, so the next release will be:


```suggestion
## 1.132 - 2025-08-01
```

ggetz:
Checking through the git blame for this line, the fix makes sense to me. Looks like the bug started showing up as of https://github.com/CesiumGS/cesium/commit/55c1cb91cebe38e42987f638a5676e0967025358, when adjusting the height for 2D would no longer be necessary. 

I verified this is also working with and without [the logarithmic depth buffer](https://cesium.com/learn/ion-sdk/ref-doc/Scene.html#logarithmicDepthBuffer), which [will affect the number of frustums used](https://cesium.com/blog/2018/05/24/logarithmic-depth/).

@mzschwartz5 Could you merge main into this branch and write a unit test to cover this case?

CHANGES.md
mzschwartz5: 🤦🏻 I knew that... I'll fix it

jjspace: merging `main` will pick this up, I added the new headings in https://github.com/CesiumGS/cesium/pull/12736

mzschwartz5: Yeah but still requires merge conflict resolution because *both* headings are present

mzschwartz5:
@ggetz Merged from main and added a unit test, back to you.

packages/engine/Specs/Scene/PrimitiveSpec.js
```diff
@@ -1367,6 +1367,42 @@ describe(
...
+
+      frameState.scene3DOnly = false;
+      frameState.mode = SceneMode.SCENE2D;
+      frameState.commandList.length = 0;
+      primitive.update(frameState);
+
+      expect(frameState.commandList.length).toEqual(1);
+      const command = frameState.commandList[0];
+      expect(command.boundingVolume).toBeDefined();
+      expect(command.boundingVolume.center.x).toBeCloseTo(height, 1);
```
ggetz, line 1404: No need to hold up this PR, but just an FYI for the future—

We have a custom matcher `.toEqualEpsilon` for fuzzy equality matches. While it's similar in principal, I think it was put in place before `toBeCloseTo` was part of the jasmine API. `toEqualEpsilon` will also work for arrays and any classes that implement `.equalsEpsilon`, like `Cartesian3`. Since it supports most cases, and uses the vocabulary from the relevant parts of the CesiumJS API, I might suggest using it over `toBeCloseTo`. 

(As an aside, it may be worth adding more color [to the testing guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/TestingGuide/README.md#comparing-floating-point-values) which mentions the array and classes.)

Issue #11141, opened 2023-03-06
Sandcastle `console.log` does not behave properly with multiple arguments


jjspace:
The `console.log` statement in Sandcastle does not behave the same way it does in a full browser/devtools when provided with multiple arguments. In Sandcastle it only logs the first argument to the builtin "console". This is a minor issue but it has a negative impact on Developer experience. It confused me when I first ran into it and realized I had to use a template string to log multiple values if I wanted them

![2023-03-06_13-47](https://user-images.githubusercontent.com/8007967/223202796-37654b34-d4c4-4894-87bb-44e201630ed6.png)

Sandcastle example: [here](https://sandcastle.cesium.com/#c=bVFRS8MwEP4rRxE6oTRMH61D8EkQfPDRCqbpbQu75Upy2abif/faoeg00KT57ru778sZAzxgAFkjWCfZEvS4Q1IwguOQmBCEISFOnLZwHCM6aQvgLEOWNhgDS45TeOljEiBeQRIruMWg8TaMhURJDNdQCibxYVVeHXFtUGvCrHRrdJuygnKbSfxAOP7buMpjlaQXza9gXsGTfhcVXD6fX03NfRi7hb7jA9zBnjP1gIdBRR5tdbzDSZP6IOYNkN8c3XBA6NTs/reUl0kKfOmAbxFw9q4iPvSYj9u3kI+X8xM3SjuF5qfADx9FVTRJXgkXbYBp3fjtwFEgR5rVtdG3HEhfNJkuuw1K7VIa643UxvxMbXq/A99f66Qw+by95SDWB4w6MUc2JY0sM9Gjf8O2WDRG+X9SiW2vM3rYYST7OtLW88X9EazrujF6/T9TmKmz8aTyJw)

Browser: Chrome 110

jjspace:
This was addressed in https://github.com/CesiumGS/cesium/pull/13014

PR #12731, opened 2025-07-11
Sandcastle using Stratakit

jjspace:
# Description

The purpose of this PR was to do a full facelift to the new version of Sandcastle and embrace the Stratakit UI components. I also focused on solidifying the _behavior_ of the various panels and parts of the app. It should now be very close to how we want it based on our [latest designs](https://community.cesium.com/t/redesigning-sandcastle-requesting-feedback/41981).

## What this PR does

- Switch from itwin-ui to stratakit packages
    - Updated necessary CSS variables we were already using
- Restructure the main `App` component to include a header and application/platform of left navigation
- Added a console mirror with draggable height
- Re-arranged the header buttons to be with the code editor
- Restructured Gallery to be a panel covering the left side of the screen
- Early `localStorage` based settings with the theme button
- Add many icons from stratakit throughout for buttons

## What this PR does NOT do

- Gallery search - The inputs are there but do nothing
- Modals, settings and the title/description metadata
- Stratakit CSS in the bucket iframe
    - I re-define a few variables in the CSS for now, this will be changed in a separate PR

## Issue number and link

Part of #12566 

## Testing plan

- Run `npm run dev` from the Sandcastle package or `npm run build-sandcastle` and `npm start` from the root
- Open the new sandcastle and play with it, this is a mostly visual change so check spacing, layout, interactions
    - I added a ["viewerless" sandcastle](http://localhost:5173/?id=viewerless) for now to help when testing to not always have to load the viewer and WebGL context 
- Check the console is working as intended
    - Create a sandcastle that uses `console.log`, `console.warn` _and_ `console.error` to make sure it displays them all correctly
    - Should auto-open if there are any warnings or errors
    - Should stay scrolled to the bottom as new messages come in
    - The [multiple arguments issue](https://github.com/CesiumGS/cesium/issues/11141) will be fixed separately
    - Clicking to collapse and expand again should remember the height
- Make sure loading the base route starts on the gallery. Make sure any route with a share string or `id` starts on the editor
- Make sure the viewer area does not flash or double render when loading or picking new gallery items
- Make sure the theme is remembered and stays the same after refreshing the page

# Author checklist


ggetz:
Any of this can (and probably should) be addressed in a future PR, but here's UI feedback from playing around:

- **Bug**: Console is not properly cleared between runs. To reproduce:

    1. Run the following example:
    
    ```js
    
    viewer.scene.postRender.addEventListener(() => {
      console.log("hello");
    });
    ```

    2. Open the console
    3. Comment out the `console.log`
    4. Hit run. About 3 logged "hello"s will show up.
- **UI suggestion**: I agree with [this user](https://community.cesium.com/t/redesigning-sandcastle-requesting-feedback/41981/3?u=gabby_getz) that the code window should probably not open by default when selecting a example from the gallery. It's a bit jarring when you are exploring the gallery.
- **Idea** (low priority): A button in the console bar that fill pause the execution, like what's available in developer tools.

    image

package.json
```diff
@@ -163,5 +163,13 @@
...
\ No newline at end of file
+  ],
+  "overrides": {
+    "allotment": {
+      "use-resize-observer": {
+        "react": "^19.0.0",
+        "react-dom": "^19.0.0"
+      }
+    }
+  }
```
ggetz, line 174: Should this all go in `packages/sandcastle/package.json`?

packages/sandcastle/gallery/viewerless/main.js
```diff
@@ -0,0 +1,8 @@
+//import * as Cesium from "cesium";
+
+//const viewer = new Cesium.Viewer("cesiumContainer");
+
+console.log("Sandcastle loaded");
```
ggetz, line 5: Is the idea to keep this around, or clean it up before merging the PR?

> I added a ["viewerless" sandcastle](http://localhost:5173/?id=viewerless) for now to help when testing to not always have to load the viewer and WebGL context

If the former, I would probably suggest leaving it entirely blank.

packages/sandcastle/public/templates/bucket.css
```diff
@@ -16,25 +16,31 @@ body {
   height: 100%;
   overflow: hidden;
 
-  background-color: white;
+  background-color: var(--stratakit-color-bg-page-base);
   background-image: var(--_rings), var(--_gradient);
 
   /* TODO: these should pull from the stratakit directly somehow eventually */
```
ggetz, line 22: Can we remove this TODO now?

packages/sandcastle/src/ConsoleMirror.tsx
```diff
@@ -0,0 +1,90 @@
...
+  logs,
+  expanded: consoleExpanded,
+  toggleExpanded,
+}: {
+  logs: ConsoleMessage[];
+  expanded: boolean;
+  toggleExpanded: () => void;
+}) {
+  const logsRef = useRef<HTMLDivElement>(document.createElement("div"));
+  // TODO: determine if we need this lib or can implement ourselves. It's a little outdated
```
ggetz, line 26: Address or add an item to the running TODO list?

packages/sandcastle/src/Gallery.tsx
```diff
@@ -1,57 +1,100 @@
...
-
-const GALLERY_BASE = __GALLERY_BASE_URL__;
+}) {
+  const knownLabels = useMemo(() => {
+    const labels = new Set<string>();
+    for (const demo of demos) {
+      demo.labels?.forEach((label) => labels.add(label));
+    }
+    return [...labels].sort((a, b) => a.localeCompare(b));
+  }, [demos]);
```
ggetz, line 76: Consider aggregating these during the `buildGallery` script.

packages/sandcastle/src/SandcastleEditor.css
```diff
@@ -0,0 +1,29 @@
...
+      padding: 4px;
+    }
+  }
+
+  section {
+    border-top: 1px solid grey;
+    flex-shrink: 1;
+    /* Monaco needs a defined height for dynamic resizing to work */
+    --tabs-height: 1.5rem;
+    height: calc(100% - var(--tabs-height)) !important;
```
ggetz: Can we avoid the `!important` somehow?

packages/sandcastle/templates/bucket.html
```diff
@@ -39,11 +39,80 @@
...
+        font-size: 9pt;
+        padding: 0;
+        margin: 0;
+        width: 100%;
+        height: 100%;
+        overflow: hidden;
+
+        background-image: var(--_rings), var(--_gradient);
+
+        /* TODO: these should pull from the stratakit directly somehow eventually */
```
ggetz, line 63: Can we remove?

packages/sandcastle/vite.config.ts
```diff
@@ -12,6 +13,7 @@ const baseConfig: UserConfig = {
   },
   define: {
     __COMMIT_SHA__: JSON.stringify(undefined),
+    __CESIUM_VERSION__: JSON.stringify(`Cesium ${rootPackageJson.version}`),
```
ggetz, line 16: A comment that may be easier to address later, but we may be able to remove some of the build complexity by grabbing this info during the `buildGallery` step.

packages/sandcastle/src/Gallery.tsx
jjspace: This was removed in the search PR in favor of the list generated by `pagefind`. I'm gonna leave as is for this PR.

packages/sandcastle/src/SandcastleEditor.css
jjspace: iirc I added this because the monaco editor adds the height on the inline style tag which has "max priority" for styles. I'll take a peek at it again but I don't think so

package.json
jjspace: Very annoyingly npm seems to ignore overrides in workspaces in monorepos. It didn't work until I added it here. I can try and dig into it more but I think this is simply a limitation we have to work around until the dependency itself gets updated (https://github.com/ZeeCoder/use-resize-observer/issues/108) or `allotment` removes the dependency in favor of a "pure react" solution (https://github.com/johnwalley/allotment/issues/833#issuecomment-2879271232)

packages/sandcastle/public/templates/bucket.css
jjspace: I specifically left the pure CSS integration of stratakit into the bucket iframe for a separate PR. This will be addressed with that work.

packages/sandcastle/gallery/viewerless/main.js
jjspace: For now yes, I've found this very helpful while developing to save a bit of computer resources. Maybe we remove it with the content pass later but I'd like to keep it for now.

I don't want to leave it _entirely_ blank as then it wouldn't be possible to verify the code is actually being run and updated in the bucket. I could remove the commented code if that makes it cleaner but I think it should do _something_.

packages/sandcastle/templates/bucket.html
jjspace: not yet, see above, will remove with the CSS integration

packages/sandcastle/gallery/viewerless/main.js
ggetz: No problem, but please add an item to https://github.com/CesiumGS/cesium/issues/12566 so we don't forget.

package.json
ggetz: That's not ideal. I was hoping to avoid any references to react from the main package. Since it's not a direct dependency, I don't want to mislead anyone who may just skim. 

Fine to leave for now since I don't think it's worth replacing allotment because of this.

packages/sandcastle/src/SandcastleEditor.css
ggetz: As long we do due diligence to check. Please add a comment if its unavoidable.

jjspace:
Thanks @ggetz this has been updated and is ready for another pass, maybe good to go now.

ggetz:
I like the "Open code" compromise:

image

ggetz:
Looks great! Thanks @jjspace!

PR #12736, opened 2025-07-14
Specify `undefined` in many `Scene` types

### LINKED ISSUE #12729: Mismatched return type for Scene.clampToHeight, sampleHeight, and clampToHeightMostDetailed
### What happened?

Return type for [Scene.clampToHeight](https://cesium.com/learn/ion-sdk/ref-doc/Scene.html?classFilter=Scene#clampToHeight) does not match description.  Description indicates that it may return `undefined` but return type is listed as `Cartesian3` and not `Cartesian3|undefined`.   This can also cause issues in Typescript since the return type does not include `undefined`

Image

Same issue with missing `undefined` for
- [clampToHeightMostDetailed](https://cesium.com/learn/ion-sdk/ref-doc/Scene.html#clampToHeightMostDetailed)
- [sampleHeight](https://cesium.com/learn/ion-sdk/ref-doc/Scene.html#sampleHeight)

### Reproduction steps

_No response_

### Sandcastle example

_No response_

### Environment

_No response_

jjspace:
# Description

As pointed out in https://github.com/CesiumGS/cesium/issues/12729 and https://github.com/CesiumGS/cesium/issues/11749#issuecomment-1950184016 there are a handful of functions and properties that are documented as possibly `undefined` but the JSDoc types do not portray that. This was a quick pass to try and catch most (maybe all) of them in the `Scene` class.

There are **no functional changes** in this PR, only documentation and typescript types.

It is _possible_ that people using CesiumJS with TS may see some new type errors downstream. If this is the case these are all instances they _should_ have been accounting for in the past but may not have realized. If anything this should force downstream code to be more stable.

Full list of updated properties/functions
- `Scene.skyBox`
- `Scene.skyAtmosphere`
- `Scene.sun`
- `Scene.moon`
- `Scene.debugCommandFilter`
- `Scene.lastRenderTime`
- `Scene.debugFrustumStatistics`
- `Scene.pick` return type
- `Scene.pickMetadataSchema`
- `Scene.pickFromray`
- `Scene.sampleHeight`
- `Scene.clampToHeight`
- `Scene.sampleHeightMostDetailed`
- `Scene.clampToHeightMostDetailed`
- `Scene.cartesianToCanvasCoordinates`

I also included some new JSDoc declarations for a handful of functions in `Picking` to help with intellisense internally. This whole class is `private` so should not appear in documentation or the exported typescript. These mostly just helped verify the return types for some of the modified functions above.

## Issue number and link

Fixes #12729

## Testing plan

- run `npm run build-docs`
- run `npm run build-ts`
- Verify the new types show correctly in the local documentation
- Verify the correct types are in `Source/Cesium.d.ts`
- Verify the correct types are in `packages/engine/index.d.ts`

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Picking.js
+++ b/packages/engine/Source/Scene/Picking.js
@@ -272,7 +272,7 @@ function computePickingDrawingBufferRectangle(
  * @param {Cartesian2} windowPosition Window coordinates to perform picking on.
  * @param {number} [width=3] Width of the pick rectangle.
  * @param {number} [height=3] Height of the pick rectangle.
- * @returns {object} Object containing the picked primitive.
+ * @returns {object | undefined} Object containing the picked primitive.
  */
 Picking.prototype.pick = function (scene, windowPosition, width, height) {
   //>>includeStart('debug', pragmas.debug);
@@ -923,6 +923,16 @@ function getTilesets(primitives, objectsToExclude, tilesets) {
   }
 }
 
+/**
+ * @private
+ * @param {Picking} picking
+ * @param {Scene} scene
+ * @param {Ray} ray
+ * @param {Object[] | undefined} objectsToExclude
+ * @param {number | undefined} width
+ * @param {Function} callback
+ * @returns {Promise<Cartesian3 | undefined>}
+ */
 function launchMostDetailedRayPick(
   picking,
   scene,
@@ -1266,6 +1276,12 @@ const scratchSurfaceNormal = new Cartesian3();
 const scratchSurfaceRay = new Ray();
 const scratchCartographic = new Cartographic();
 
+/**
+ * @private
+ * @param {Scene} scene
+ * @param {Cartographic} cartographic
+ * @returns {Ray}
+ */
 function getRayForSampleHeight(scene, cartographic) {
   const ellipsoid = scene.ellipsoid;
   const height = ApproximateTerrainHeights._defaultMaxTerrainHeight;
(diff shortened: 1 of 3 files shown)


packages/engine/Source/Scene/Picking.js
```diff
@@ -1266,6 +1276,12 @@ const scratchSurfaceNormal = new Cartesian3();
 const scratchSurfaceRay = new Ray();
 const scratchCartographic = new Cartographic();
 
+/**
+ * @private
+ * @param {Scene} scene
```
javagl, line 1281: It's a pity that this and several other functions receive **the** most complex (and most stateful) object there is - namely, the `Scene` - when all that they need is the (small and state-free(!)) `ellipsoid`. The function should always only have received the ellipsoid to begin with.

javagl (approved):
Looks good to me.

As you said: This will likely affect TypeScript users. And for them, this may, in the strictest, sense, be a "breaking" change. But it's certainly a _good_ change, because these users may have been lured into a false safety, _assuming_ that something couldn't be `undefined`, even though it _could_ be.

I had a short look at the generated `.d.ts` bindings, before and after this change, and it didn't seem to contain any surprises. Just one "outtake"/example from one of the _slightly_ more complex changes:

Cesum Undefined Changes

Yep. That's what it should be.

Just cross-linking: This resolves https://github.com/CesiumGS/cesium/issues/11749#issuecomment-1950184016 for the `Scene` class.

mzschwartz5 (approved):

Issue #12742, opened 2025-07-16
Draped imagery visibility is not updated


javagl:
### What happened?

Under certain conditions (roughly: when _the same_ `ImageryLayer` is draped over both the terrain **and** a tileset), toggling the visibility of the imagery layer with `imageryLayer.show = state;` does not cause the imagery to be updated for the tileset.

![Image](https://github.com/user-attachments/assets/c239a512-4f0d-4dd0-83c6-56f1023d3fac)

### Reproduction steps

Toggle the visibility of the tile coordinates imagery in this sandcastle

https://sandcastle.cesium.com/index.html#c=lVNda9swFP0rwk82ZLLXL8aShq3ZwwItK03oXvyiyNe2mKwbpGuHdOy/T7YcMG0z2JMl3XPuPedIlmgcsU7BASy7ZQYObAVOtQ1/Hs7iPJLDfoWGhDJg8yiZ5yY3cmCS0uCAPFUchKITOXwuv21DmZcWmzWar85v1kV8lX26uenbhMHcSTDA91Y1ilQHjouiiMfWPWxc8gJ2bbWp8XCHrSmUqZ5Rtw346WRbmMiSiNbXBYFbN6ICe7wXx9cOp5U4N2xa64Wv3jR5tNipwqN/93DWu/ciNrKGBj5P6T9h9wBWCkK7nWDiZNYT/yS5mdpSEyHB+xn9k8j+i5ObjTCFFI409NgtVpWGu5YIzeA8j8JRHg0Cz7Tizkc/AMrWSFJoWOx9yV9QJGyM5F9UH/8Inw8xhBROd4ZlGV7SJMfvIPprflQk6ydhKhjUjsUHQTUnfPIQYVz84eKCX4eAzwCurnkWAJdZloXpY5wviM0WT29uNooJ8qJZtHB01LAMFhn7opo9WmKt1THnKUGz173jdNd6d8Slc8k8gBfplLooVMdUcfvOX8WkFs75StlqvVEv/i6Wi9Tj31A1DqH86MBqcexh9cflfTjknC9Sv32fSYh6J+yrzn8B

### Sandcastle example

See above

### Environment

Browser: All
CesiumJS Version: 1.131
Operating System: All

PR #12743, opened 2025-07-16
Update draped imagery visibility

javagl:
Fixes https://github.com/CesiumGS/cesium/issues/12742

# Description

The visibility of imagery that was draped over a tileset was not updated under certain conditions when the `show` flag was toggled. This is fixed here with two small changes:

- The `ModelImagery` stores an `ImageryConfiguration` object, to track changes of the "state" of the imagery. When this configuration changes, then the draw commands of the model are reset, causing the pipelines to be re-run. Unitl now, this configuration covered the `alpha`, `brightness`, etc. properties. Now it also covers the `show` property.
- The `ImageryPipelineStage` object creates `ImageryInput` objects that are sent to the shader, and contain information about the imagery that is draped over the model. Until now, this created these `ImageryInput` objects regardless of the `show` value of the respective imagery layer. Now, when there is an `imageryLayer.show === false`, then it will _not_ create these objects for that layer

Considerations:

There are structures (like the whole `ModelPrimitiveImagery`, storing the texture and texture coordinate information for the draping) that **could**, in theory, be discarded when setting `show=false`. But then they would need to be rebuilt from scratch when setting `show=true` later. I consider this `show` flag as a pure, quick, "rendering-time" toggle. So when setting `show=false` now, then these structures are kept. Only when the imagery layer is actually _removed_ from the `ImageryLayerCollection`, the associated structures are discarded.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12742

## Testing plan

Open the sandcastle from the issue and toggle the visibility of the tile coordinate imagery. Ensure that the visibiltiy of the tile coordinate imagery (grid and numbers) on the tileset matches the state of the toggle value.

# Author checklist

### DIFF
--- a/Specs/loadTilesetWithImagery.js
+++ b/Specs/loadTilesetWithImagery.js
@@ -1,0 +1,73 @@
+import {
+  Cartesian3,
+  HeadingPitchRoll,
+  ImageryLayer,
+  TileCoordinatesImageryProvider,
+  Transforms,
+  WebMercatorTilingScheme,
+} from "@cesium/engine";
+
+import pollToPromise from "./pollToPromise";
+import Cesium3DTilesTester from "./Cesium3DTilesTester";
+
+// A currently hard-wired tileset to be loaded for imagery draping tests
+const tileset_unitSquare_fourPrimitives_plain_url =
+  "./Data/Models/glTF-2.0/unitSquare/tileset_unitSquare_fourPrimitives_plain.json";
+
+/**
+ * Wait until the root tile of the given tileset is loaded
+ *
+ * @param {Cesium3DTileset} tileset The tileset
+ * @param {Scene} scene The scene
+ */
+async function waitForRootLoaded(tileset, scene) {
+  scene.renderForSpecs();
+  const root = tileset.root;
+  await pollToPromise(() => {
+    scene.renderForSpecs();
+    return root.contentFailed || root.contentReady;
+  });
+}
+
+/**
+ * Load and return a test tileset that defines an imagery layer,
+ * waiting until the root of that tileset is loaded.
+ *
+ * This means that the resulting <code>tileset.root.content._model._modelImagery</code>
+ * (including the <code>ModelPrimitiveImagery</code> instances) will be defined and ready.
+ *
+ * @param {Scene} scene The scene
+ * @returns {Cesium3DTileset} The tileset
+ */
+async function loadTilesetWithImagery(scene) {
+  const url = tileset_unitSquare_fourPrimitives_plain_url;
+  const tileset = await Cesium3DTilesTester.loadTileset(scene, url);
+
+  // Create a non-trivial transform for the tileset
+  const transform = Transforms.eastNorthUpToFixedFrame(
+    Cartesian3.fromDegrees(-120.0, 40.0, 1.0),
+  );
+  tileset.modelMatrix = transform;
+
+  // Set a view that fully shows the tile content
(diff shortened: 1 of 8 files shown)


javagl:
The commit at https://github.com/CesiumGS/cesium/pull/12743/commits/0cd17d84193b7116e298eea0ca5c1332d982cbc3 adds some specs. Given that the change itself was not too large, there wouldn't need to be many additional specs. But given that the `ImageryPipelineStage` until now had no tests at all (due to time constraints for the original implementation), I now added a dedicated spec for that. This included pulling out a `Specs/loadTilesetWithImagery.js` function that loads an unspecified tileset with imagery layers, to be used for the `ModelImagerySpec`, and `ImageryPipelineStageSpec`, and maybe other draping-related specs that may be added.

Specs/loadTilesetWithImagery.js
ggetz, line 1: Since this is only used within `Specs/Scene/Model`, could we please move this file to that directory?

packages/engine/Specs/Scene/Model/ModelImagerySpec.js
```diff
@@ -161,8 +92,37 @@ describe("Scene/Model/ModelImagery", function () {
...
+    const content = root.content;
+    const model = content._model;
+    const modelImagery = model._modelImagery;
+    const imageryLayer = tileset.imageryLayers.get(0);
+
+    // Initially, _imageryConfigurationsModified is false (it was just updated)
+    expect(modelImagery._imageryConfigurationsModified()).toBeFalse();
+
+    // For spec: Modify imagery configuration
+    imageryLayer.show = 0.5;
```
ggetz: Shouldn't this be a boolean? I'm hesitant to leave this in the unit tests, as unit test can sometimes serve as an informal way of documenting functionality.

It seems from the comment that you're attempting to trigger a change without hiding the layer. Maybe we can set it to `false` then back to `true`?

ggetz (changes requested):
Thanks @javagl for the fix and well as making the unit tests more robust 👍 

> I consider this show flag as a pure, quick, "rendering-time" toggle. So when setting show=false now, then these structures are kept. Only when the imagery layer is actually removed from the ImageryLayerCollection, the associated structures are discarded.

I agree with this, and it's consistent with how we handle `show` in other parts of the API.

Specs/loadTilesetWithImagery.js
javagl: `ViewportPrimitive` is only used in `Specs/Scene`. But as a rule of thumb, "making the 'scope/visibility' of something as broad as necessary, but as narrow as possible" certainly makes sense, so I moved it.

packages/engine/Specs/Scene/Model/ModelImagerySpec.js
javagl: About the type: "Ooops". (That's what reviews are for...)

About what the test is testing: It is testing that a change in the `show` flag is detected as a change (and this change eventually triggers an update of the draw commands). This is independent of whether it changes to `false` or to `true` (or to 0.5 😬 ).

Setting to `true` could be tested, but that would also only test whether _any_ change of the `show` flag tiggers an update. 

Maybe I misunderstood that point.

ggetz: Perhaps I was looking for a deeper meaning where there wasn't one. 😅 The updates look good to me!

ggetz (approved):

PR #12744, opened 2025-07-16
Model instance v2

rudacs:
# Description

Some adjustments to the implementation being made to ModelInstanceCollection

### DIFF
--- a/packages/engine/Source/Scene/Model/PickingPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/PickingPipelineStage.js
@@ -38,11 +38,11 @@ PickingPipelineStage.process = function (
   const runtimeNode = renderResources.runtimeNode;
   const shaderBuilder = renderResources.shaderBuilder;
   const model = renderResources.model;
-  const instances = runtimeNode.node.instances;
+  const instances = renderResources.instanceCount;
 
   if (renderResources.hasPropertyTable) {
     processPickTexture(renderResources, primitive, instances, context);
-  } else if (defined(instances)) {
+  } else if (defined(instances) && instances > 0) {
     // For instanced meshes, a pick color vertex attribute is used.
     processInstancedPickIds(renderResources, context);
   } else {
--- a/packages/engine/Source/Scene/Model/PickingPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/PickingPipelineStage.js
@@ -179,6 +179,8 @@ function processInstancedPickIds(renderResources, context) {
     pickIdsTypedArray[i * 4 + 1] = Color.floatToByte(pickColor.green);
     pickIdsTypedArray[i * 4 + 2] = Color.floatToByte(pickColor.blue);
     pickIdsTypedArray[i * 4 + 3] = Color.floatToByte(pickColor.alpha);
+
+    model.instances._instances[i]._pickId = pickId;
   }
 
   model._pickIds = pickIds;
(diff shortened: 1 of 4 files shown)


rudacs:
this is ready

ggetz:
Thank you for the fixes here @rudacs! I can confirm we have a CLA on file for you and Geoex Tecnologias.

@lukemckinstry Could you please review and merge into your branch?

lukemckinstry:
Thanks @rudacs for this PR. Two items 

 - I needed to rebase `model-instance-v2` to fix conflicts in `main`. Looks like this impacted the git history in this branch here, so I think you will want to rebase onto the latest version of `model-instance-v2` as well. I suggest doing a interactive rebase (`git rebase -i model-instance-v2`) and dropping all but the original 6 commits in your branch.
 - To make sure I am understanding you correctly, how are you using and testing the pick logic you added? Can you provide some test instructions for that part?

rudacs:
> * To make sure I am understanding you correctly, how are you using and testing the pick logic you added? Can you provide some test instructions for that part?

:)

The idea was that when I pick I would know which instance I was clicking on.

```
const picked = viewer.scene.pick(movement.position);
const modelInstance = picked.primitive.instances.get(picked.instanceId);
```

With the pickId in the ModelInstance, I can:

```
const edge = Cesium.PostProcessStageLibrary.createEdgeDetectionStage();
edge.uniforms.color = Cesium.Color.RED;
edge.uniforms.length = 0.05;
edge.selected = [modelInstance];

viewer.scene.postProcessStages.add(Cesium.PostProcessStageLibrary.createSilhouetteStage([edge]));
```

I also exposed the transform so I could, for example, change its position.

````
modelInstance.transform = newModelMatrix;
````

I also adjusted the remove because as I don't know what the index is and the add returns the ModelInstance, I changed the remove to pass the instance.

```
const modelInstance = collection.add(...);
collection.remove(modelInstance);
```

File: packages/engine/Source/Scene/Model/PickingPipelineStage.js [hunk 1]
```diff
@@ -38,11 +38,11 @@ PickingPipelineStage.process = function (
   const runtimeNode = renderResources.runtimeNode;
   const shaderBuilder = renderResources.shaderBuilder;
   const model = renderResources.model;
-  const instances = runtimeNode.node.instances;
+  const instances = renderResources.instanceCount;
 
   if (renderResources.hasPropertyTable) {
     processPickTexture(renderResources, primitive, instances, context);
-  } else if (defined(instances)) {
+  } else if (defined(instances) && instances > 0) {
```
lukemckinstry: We actually have slightly different code paths for instancing from glTF and i3dm versus instancing from the API which we have started to call runtime instancing. 
I believe `runtimeNode.node.instances` supports the glTF andf i3dm case. And it looks like  `processPickTexture` is expecting `runtimeNode.node.instances`.
So I think adding
```
const instanceCount =  renderResources.instanceCount;
```
and then modifying the lower line to 
```
} else if (defined(instances) || (defined(instanceCount) && instanceCount > 0)) {
```

would cover both cases.

rudacs: Adjusted.
It would actually fail for another case.

lukemckinstry:
Thanks for the updates. This is looking good. The last thing I'll check when I have time is if and how we should test the picking logic you added. I should get to it within a day, but feel free to go ahead and get a jump start working on the test question if you wish.

PR #12745, opened 2025-07-17
1.131.1 patch updates

ggetz:
# Description

We just released an npm-only prerelease version for testing. This PR bring in the version updates, plus a bit of cleanup for `CHANGES.md`.

## Issue number and link

N/A

## Testing plan

N/A— Functionality unchanged.

# Author checklist


ggetz:
@jjspace Please review and merge if happy.

ggetz:
> If this is going to become more common we may want to document the "pre-release" process so it's set in stone and clear what needs to happen and where.

Sounds good! I'm preparing the docs now.

>It also might make sense to start to tag these pre-release versions as something other than latest. Right now the way this one was published will make everyone get the new version if they npm install cesium@latest

Agreed. I found the right arguments to use when publishing to avoid this in the future. I'll include it in the docs.

PR #12752, opened 2025-07-21
Gallery conversion for Sandcastle v2

jjspace:
# Description

This is inherently a big PR. We have over 250 sandcastle examples and in the new format they are each 4 files. To help facilitate review I've intentionally broken the change into 3 commits

1. Creating the conversion script itself
2. Running the conversion directly and ignoring any errors
3. Fixing all eslint issues in the sandcastles

This should hopefully make reviewing easier but it also should help with running the script again later if we need to. I know there's a chance new sandcastles get created or updated outside this effort. We will do a final content pass later to give them all a final cleanup but for now it should be mostly automated. I'm leaving the script in the repo for now but plan to remove it as well once we can make the switch.

## Issue number and link

Part of #12566 

## Testing plan

- either run `npm run dev` from the sandcastle package or `npm run build-sandcastle` and `npm start` from the repo root
- Scroll through the gallery and make sure all items show up (UI changed drastically in #12731)
- Click on a handful of gallery items and make sure they work. lmk if you find broken ones but we will do a full pass on them all later.

# Author checklist

- [ ] I have submitted a Contributor License Agreement
- [ ] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [ ] I have performed a self-review of my code

PR #12755, opened 2025-07-21
New Gallery search for sandcastle

jjspace:
# Description

Re-implement searching for sandcastles in the Gallery using [pagefind](https://pagefind.app/).

Pagefind is primarily targeted at indexing static html files. We need to build our own custom indexes so it can search in the JS code. I've looped this into the script that already builds the gallery list of metadata. I've also chosen to store the build output inside the gallery directory so the FE can be pointed at a single "gallery path" and load everything it needs for search and the list and assets. (Eventually we will need to support multiple galleries but this is not part of this PR)

When `pagefind` builds an index it also builds the JS module we should import for that index. This makes `vite` (and TS) a bit upset because it doesn't know where the files are. We need to do a dynamic import and tell vite to ignore it to actually load the `pagefind` web module. I encapsulated all this logic in the `GallerySearch` component to keep it all in one place.

## Issue number and link

Part of #12566 

## Testing plan

- run `npm run dev` in the package or `npm run build-sandcastle` and `npm start` from the project root
- Search for something in the gallery. Test multiple terms, including ones that have no results
- Test the dropdown for filtering tags. Make sure it works with a search term _and_ without
- ~~If you want to test with all the sandcastles copy the [`convertGallery.js` script](https://github.com/CesiumGS/cesium/blob/gallery-conversion/packages/sandcastle/scripts/convertGallery.js) into `packages/sandcastle/scripts` and run it. This will generate all the other sandcastles and re-running the build scripts will index them all.~~
    - ~~Once https://github.com/CesiumGS/cesium/pull/12752 is merged I can merge it into this branch and you won't have to do this manually. (I just don't want to mix branches)~~ This branch now has all sandcastles

# Author checklist


packages/sandcastle/scripts/buildGallery.js
```diff
@@ -29,6 +35,15 @@ export function buildGalleryList(galleryDirectory, includeDevelopment = true) {
     legacyIdMap: {},
   };
 
+  const { index } = await pagefind.createIndex({
+    verbose: true,
```
ggetz, line 39: Can we turn off the verbosity by default to reduce noise:

1. In the CI logs
2. For developers running this at the project root

It could make sense to pass a `--verbose` object for sandcastle-specific dev scripts.

packages/sandcastle/scripts/buildGallery.js
```diff
@@ -101,22 +120,49 @@ export function buildGalleryList(galleryDirectory, includeDevelopment = true) {
...
       isNew: false,
     });
     if (legacyId) {
       output.legacyIdMap[legacyId] = slug;
     }
+
+    const jsFile = readFileSync(`${galleryBase}/main.js`, "utf-8");
+    await index.addCustomRecord({
+      url: `?id=${slug}`,
+      content: jsFile,
```
ggetz, line 143: Can we surround this with `<code>` tags for better suggestion display?


```suggestion
      content: `<pre><code>${jsFile}</pre></code>`,
```

packages/sandcastle/scripts/buildGallery.js
```diff
@@ -126,11 +172,17 @@ export function buildGalleryList(galleryDirectory, includeDevelopment = true) {
...
 if (import.meta.url.endsWith(`${pathToFileURL(process.argv[1])}`)) {
-  const __dirname = dirname(fileURLToPath(import.meta.url));
   const defaultGalleryDirectory = join(__dirname, "../gallery");
-  const { output, hasErrors } = buildGalleryList(defaultGalleryDirectory);
-  console.log("processed", output.entries.length, "sandcastles");
-  if (hasErrors) {
-    exit(1);
-  }
+  buildGalleryList(defaultGalleryDirectory)
+    .then(({ output, hasErrors }) => {
```
ggetz: Use `async/await` and `try/catch` for consistency with the rest of our code.

packages/sandcastle/src/Gallery.css
```diff
@@ -3,10 +3,14 @@
...
 
+  &.hidden {
+    display: none;
+  }
+
   .filters {
     display: flex;
     align-items: center;
-    gap: 0.3em;
+    gap: 8px;
```
ggetz, line 13: Does stratakit have standardized spacing which we should be referencing as variables?

packages/sandcastle/src/Gallery.tsx
```diff
@@ -15,6 +16,140 @@ export type GalleryItem = {
   isNew: boolean;
 };
 
+export function GallerySearch({
```
ggetz, line 21: Can we move this component to its own file?

packages/sandcastle/src/TestApp.tsx
ggetz, line 1: Is this file included intentionally?

packages/sandcastle/src/Gallery.tsx
```diff
@@ -15,6 +16,140 @@ export type GalleryItem = {
...
+        <Button
+          onClick={() => {
+            setSearchTerm("");
+            if (inputRef.current) {
+              inputRef.current.value = "";
+            }
+          }}
+        >
+          Clear
+        </Button>
```
ggetz: Image 

Instead of the "Clear" button (which is a bit non-standard AFAIK), can we implement a more minimal cancel icon button?

i.e., something like the following (It's fine to stick with "clear" as a label IMO) .

Image

Image

ggetz: I also noticed both of these examples use the search icon as well. I would suggest we add it. I think that would help guide users looking for search capabilities.

packages/sandcastle/src/Gallery.tsx
```diff
@@ -15,6 +16,140 @@ export type GalleryItem = {
...
+          if (debounceTimeout.current) {
+            clearInterval(debounceTimeout.current);
+          }
+          const newSearchTerm = e.target.value;
+          pagefind.current?.preload(newSearchTerm);
+          debounceTimeout.current = setTimeout(() => {
+            setSearchTerm(newSearchTerm);
+          }, 300);
+        }}
+        placeholder="Filter demos..."
```
ggetz:
```suggestion
        placeholder="Search gallery"
```

1. Let's use "Search" since all users seem to be calling it that
2. Let's use "gallery" to be specific about _where_ we're searching
3. Doing a quick comparison, I don't see a lot of `...` appended to search placeholders.

packages/sandcastle/src/Gallery.tsx
```diff
@@ -15,6 +16,140 @@ export type GalleryItem = {
...
+            setCurrentTag(e.target.value);
+            setTag(e.target.value);
+          }}
+        >
+          <option value={"All"}>All</option>
+          {Object.keys(availableFilters.tags).map((label) => (
+            <option value={label} key={label}>
+              {label}
+            </option>
+          ))}
```
ggetz, line 167: I think it may end up being a bit much to iterate on for this particular PR, but I think we discussed some alternatives for how the categories/labels are presented offline. 

Could you please add "Refine gallery filter by label presentation" to the list of TODO's?

ggetz (changes requested):
Searching is so much better 🎉 This was a HUGE user request, so it's fantastic to see.

So much so, I suggest considering adding a top-level button to the project bar— similar to VS Code. If we like it, let's implement outside of this PR.

Image

jjspace:
now that https://github.com/CesiumGS/cesium/pull/12752 is merged in this will include all sandcastles and demonstrate search better

packages/sandcastle/scripts/buildGallery.js
jjspace: I'm pretty sure this setting only affects what gets logged into the `logfile` not the console.

packages/sandcastle/scripts/buildGallery.js
jjspace: I've tried out multiple options for this and none preserve the newlines like we'd want to display them directly... This tool is very focused at plain text.

AFAICT when using [`addCustomRecord`](https://pagefind.app/docs/node-api/#indexaddcustomrecord) it _always_ processes the content as text and escapes html and considers it part of the result.

There is also an [`addHTMLFile`](https://pagefind.app/docs/node-api/#indexaddhtmlfile) function that lets you pass in a string of html to be processed _as_ html but even then it pulls out the plaintext of the content and seems to strip newlines.

It _might_ be possible to dig into this a bit more but everything I've tried so far doesn't help our unique situation. I think what we have now might be the best we get without a decent bit of extra effort and maybe changes to pagefind itself.

packages/sandcastle/scripts/buildGallery.js
jjspace: I wanted to but eslint was complaining about top level await. Updated the config to consider these files the way we do node scripts.

packages/sandcastle/src/Gallery.css
jjspace: Yes, but they were only recently added to the library. I will be doing a pass on all the CSS values like these in a separate styling/stratakit update PR (This is already on the todo list)

packages/sandcastle/src/Gallery.tsx
jjspace: I had it that way at first but then it felt excessive. These components are fairly coupled in logic and data flow. I only separated them for encapsulation of logic, specifically the `pagefind` loading. Also this component is only used inside the larger `Gallery` component that this file actually exports.

packages/sandcastle/src/Gallery.tsx
jjspace: Adding the search icon was easy with stratakit, however they don't currently support an icon as a button in an input by default. I've adjusted the style of an `IconButton` and the input to work but probably something better implemented by stratakit directly.

Also for context, I put it on the left because it's not always visible and I didn't want the input to shift around if it gets added to the right of the input.

jjspace:
@ggetz I think this is ready for another look

packages/sandcastle/src/Gallery.tsx
ggetz: Hm, hearing that they're tightly coupled seems like a red flag to me. What do you think is the biggest culprit of the coupling? Is it that we're relying on reducers and not contexts to share state?

packages/sandcastle/src/Gallery.tsx
ggetz: Updates look and are working great!

ggetz:
Overall this is looking good. I do have some concerns about the coupled components, and left a follow up comment above.

When testing out the search, it doesn't appear that sandcastle titles are being included in the search. I think they should be. No worries if that's addresses here or in a follow up PR. Just let me know what you'd prefer @jjspace.

ggetz:
Seems like it'll make sense to follow up on this in subsequent PRs, so I'll go ahead and merge this.

ggetz (approved):

Issue #12758, opened 2025-07-22
Add labels to split view sandcastles (like Gaussian Splats)


jjspace:
Initially noticed for the new [Gaussian Splatting Sandcastle](https://sandcastle.cesium.com/?src=3D%20Tiles%20Gaussian%20Splatting.html) which doesn't make it clear which side is the splats and which is the 3d tiles mesh. If you're a dev or familiar with splats it may be "obvious" but not everyone will be. We should add labels to sandcastles that have a split so it's clear what each side is.

<img width="2263" height="1496" alt="Image" src="https://github.com/user-attachments/assets/191c1e0d-ea83-47fd-b489-445a0f725e81" />

We have a handful of Sandcastles that use the split feature but only one of them labels what each side is showing

<img width="2742" height="287" alt="Image" src="https://github.com/user-attachments/assets/487e7ca3-422d-480e-be2d-df8d0f58585b" />

PR #12759, opened 2025-07-22
Add labels to split view sandcastles

jjspace:
# Description

Add clarity by including labels to sandcastles that use the split view. I'd appreciate a second pass on the wording to make sure all the labels are correct (@ggetz?).

image

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12758

## Testing plan

- run `npm start`
- Open sandcastle locally
- Search for `SplitDirection` (or part of this)
- Test each sandcastle and make sure they're labeled.
    - I specifically did not do the `SplitDirection` sandcastle itself because it'd dynamic and the focus is on the splitter not the content.

# Author checklist

### DIFF
--- a/Apps/Sandcastle/gallery/3D Tiles Compare.html
+++ b/Apps/Sandcastle/gallery/3D Tiles Compare.html
@@ -61,8 +61,8 @@
     </style>
     <div id="cesiumContainer" class="fullSize">
       <div id="slider"></div>
-      <div class="split-label left">3D Tileset</div>
-      <div class="split-label right">OSM Buildings</div>
+      <div class="split-label left">3D Tiles<br />Mesh</div>
+      <div class="split-label right">3D Tiles<br />OSM buildings</div>
     </div>
     <div id="loadingOverlay"><h1>Loading...</h1></div>
     <div id="toolbar"></div>
--- a/Apps/Sandcastle/gallery/3D Tiles Compare.html
+++ b/Apps/Sandcastle/gallery/3D Tiles Compare.html
@@ -39,9 +39,30 @@
       #slider:hover {
         cursor: ew-resize;
       }
+
+      .split-label {
+        position: absolute;
+        top: 50%;
+        transform: translateY(-50%);
+        background: rgba(42, 42, 42, 0.8);
+        padding: 8px;
+        border-radius: 4px;
+        z-index: 9999;
+        font-size: large;
+        font-weight: bold;
+        text-align: center;
+      }
+      .split-label.left {
+        left: 0;
+      }
+      .split-label.right {
+        right: 0;
+      }
     </style>
     <div id="cesiumContainer" class="fullSize">
       <div id="slider"></div>
+      <div class="split-label left">3D Tileset</div>
+      <div class="split-label right">OSM Buildings</div>
     </div>
     <div id="loadingOverlay"><h1>Loading...</h1></div>
     <div id="toolbar"></div>
--- a/Apps/Sandcastle/gallery/3D Tiles Gaussian Splatting.html
+++ b/Apps/Sandcastle/gallery/3D Tiles Gaussian Splatting.html
@@ -49,6 +49,7 @@
         z-index: 9999;
         font-size: large;
         font-weight: bold;
+        text-align: center;
       }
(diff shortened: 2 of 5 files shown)


Apps/Sandcastle/gallery/3D Tiles Compare.html
```diff
@@ -39,9 +39,30 @@
       #slider:hover {
         cursor: ew-resize;
       }
+
+      .split-label {
```
mzschwartz5, line 43: Do sandcastle's have any shared CSS file where we could put this rather than copying it to each?

mzschwartz5 (commented):
Looks good to me - only the one comment about potentially sharing CSS, otherwise gtg!

ggetz:
Thanks @jjspace! I didn't do a full review, but I have some wording suggetions:

"Bing Maps unlabeled + Washington Imagery + Labels only" ->
"Bing Maps (unlabeled) + Washington DC Satellite + Bing Maps (labels only)"

"Gaussian Splats / 3D Tiles Mesh" -> "3D Tiles Gaussian splats / 3D Tiles mesh"
(They are both 3D Tiles)

"3D Tileset / OSM Buildings" -> "3D Tiles mesh / 3D Tiles OSM buildings"
(They are both 3D Tiles)

Apps/Sandcastle/gallery/3D Tiles Compare.html
jjspace: The shared styles are mostly in that `bucket.css` that's imported above. This is a broader question about how we manage styles for sandcastles in general. Will probably aim to tackle this as part of the sandcastle rebuild.
We already duplicate the styles for the `#toolbar` element like this in the sandcastles that need it so this is not unusual for current sandcastles.

One concern with putting them all into a shared stylesheet is that they become "hidden" from the user. It's not obvious they exist and not easy to discover how to use them without copying from an existing sandcastle. Plus trying to replicate a sandcastle without these styles will make it not display correctly. It's a balance that needs a bit more thought, more than this small PR :laughing:

jjspace:
Thanks @mzschwartz5 and @ggetz. I just pushed some updated wording. I think this is good to go

Apps/Sandcastle/gallery/3D Tiles Compare.html
javagl: Related: There's one thing that always bugged me about the slider, far more "user-facing" than some CSS: 

Cesium Slider Duplication

It is _related_ to the CSS, and that may make it more difficult to fix. But having to call `document.getElementById` is ... not ideal in general, IMHO, and I wondered whether this "element" could somehow be sneaked into the Cesium `Viewer`, and just be hidden by default. 

The point is: That duplicated code should (or could?) preferably be some
```
viewer.enableSplitSlider(); // ? 
viewer.installDefaultSplitHandler(); // Sets up all the default action mappings...
```

Who wants to drag that slider with the right mouse button after all? 😆

jjspace: By including it in the viewer itself it would also lock people into the way we want it to look. We could add options to style it but it's still under our control. Maybe that's a good thing, maybe not. Part of a bigger discussion. Maybe someone wants the "split viewer" functionality but no draggable line and instead a slider under it, or just a static width.
Doing it the way we have it creates some duplicated code in our sandcastle examples but makes it more flexible for everyone.

javagl: Offering more fine-grained control does not mean that the boilerplate code that are currently required couldn't be offered as a convenience function. Analogously: What if users want to zoom with the _left_ mouse button instead of the right? Yeah, you'll have to rewire a few things in the `ScreenSpaceCameraController` or so. 

People who "just want to split that" can call `justSplitThat()`. People who want to make the slider 4 pixels wide and yellow can do this, but .... not with a `makeItFourPixelsWideAndYellow()` function, FWIW 😁

ggetz:
@mzschwartz5, if you reviewed the implementation, please feel free to merge once you're happy with any changes.

mzschwartz5 (approved):

Issue #10566, opened 2022-07-19
Add async constructor for Materials with image textures


jjhembd:
A [Material](https://cesium.com/learn/cesiumjs/ref-doc/Material.html) instance currently gives no indication of whether its textures are loaded or not. This makes unit testing difficult. 

For example, the following test will fail since the image is not loaded before the scene renders:
```javascript
it("renders blue from a blue texture", function () {
  primitive.appearance.material = Material.fromType("Image", { 
    image: "./Data/Images/Blue.png"
  });
  scene.primitives.add(primitive);
  // ... Set camera view, etc...

  expect(scene).toRender([0, 0, 255, 255]);
  // Error: Expected to render [0,0,255,255], but actually rendered [255,255,255,255].
});
```

Some [current tests](https://github.com/CesiumGS/cesium/blob/dc28421ff2f52871629299d00f310ee796f63833/Specs/Scene/MaterialSpec.js#L634-L643) fix the problem by waiting for a change in the length of the `material._loadedImages` array, but this is complicated to set up.

Proposal: add a static method `Material.fromUrl` returning a Promise. The Promise would resolve to a new Material instance with all textures ready for use. The above test would become:
```javascript
it("renders blue from a blue texture", async function () {
  const material = await Material.fromUrl({
    type: "Image",
    image: "./Data/Images/Blue.png",
   });
  primitive.appearance.material = material;
  scene.primitives.add(primitive);
  // ... Set camera view, etc...

  expect(scene).toRender([0, 0, 255, 255]);
});
```

gaurav8y785675474:
hello jjhembd
i am a beginner in open source world
can you please tell me exact location of code[related to this issue] in files
so that i can work on that

ggetz:
Hello @gaurav8y785675474,

The new function should be added to `packages/engine/Source/Scene/Material.js`.

Then search the unit tests for occurrences of `material._loadedImages`. That should be replaces with the async/await pattern shown above.

wesleykebrown:
Hi, I’m a part of the Flagship program for JTC. I would like to work on this issue ticket, please! This is a great opportunity to test my abilities.

ggetz:
@wesleykebrown Sounds great!

adamwirth:
I've been working on this for a while locally and wanted to ask some questions!

Questions I have for solution paths:

1. Could the tests be modified to use `pollToPromise`?
[...]

PR #12767, opened 2025-07-25
Async material constructor

mzschwartz5:
# Description

There's currently no way to wait for a material to load its required resources, or to know when they're loaded. Currently, the Material constructor only creates functions that *will* load resources, but those functions are only called on subsequent Update calls.

One consequence of this is that our unit tests rely on polling the internal `_imagesLoaded` list to determine when its okay to make assertions. This is verbose and flaky.

For ease of testing and consistency with other APIs (like `Model.fromGltfAsync`), we'd like an analogous `Material.fromTypeAsync` factory function. Because of the current image-loading polling model, and the complexity of the Material creation logic, creating such an async method is a challenge. For more context, see [this comment and following discussion](https://github.com/CesiumGS/cesium/issues/10566#issuecomment-3111564508).

The solution this PR takes is to extract image-loading logic into callable, async functions. The Update loop calls those functions indirectly (similar to before), but now we also call those functions on Material construction and keep a list of the resulting promises. The `fromTypeAsync` constructor can then `await` all the initialization promises, and the caller of `fromTypeAsync` can `await` the result of that.

See [my comment below](https://github.com/CesiumGS/cesium/pull/12767#issuecomment-3120030138) for a commit-by-commit breakdown of this PR.

## Issue number and link

#10566

## Testing plan

[Localhost sandcastle](http://localhost:8080/Apps/Sandcastle/index.html#c=pVRtT9swEP4rp35pkIpTBmwoK2gM2ITEtImXbR8qIZNcU2uJHdlOu4L233d24qQvfCMf2vheHj/35O7i+I3PVMYxXKARdQl3XGYpN7ZAMFJUFdoELBorZA7fuEUteMFmWpX3qwqj4XXJcxzuOYQ3k/A0DhhcaKSLwM4RFgKXqCHKhOFPRMmKEuHhGmZKgxFlVYhU2BVdv+A6BJ+CxGVbDvvpbdEw9ccLJS0XEvVwBC9TCfRwKUpuhZIJzHhhcNSY3UUFRbbWqfy397Fl+I7BJc7IBxxyVLnm1VykoDG1XOZEUkjIMNeIpuHVe04Drdtg8lJeNtHR/sF4zMYjOPS/+yf+7+iYjbu7Dxl8rkWR0dUdxFdUJVq9gqWwc7i/+n3/cHt1CQvUFv86pai+hkgeIjck2gGKWmk63kn/2srTgH/x2EkACu1xTl3DNZcpdqa7uqqUtiywY+sA6+IeMfhFgoKwTkYOgdO1NNYhNoWI9rRZyHZsqCOUnXRv6zceU8MpytB1amGlag2+p6FsqVNb8d3Ob3m4yOAjMnzJifiWHF3OuVnJNIxM14AeI4EhY7H3mLjJf7xRuXpUJFTBV6yS+XCd9ftuTHpyve40O9xCSmXVJRo/SSGoIc770A0Jd6GCiEGOZLPoMC4Ua4o6RWlfmZgPGzP9Q4tSWLFAqI3bKl7zrjVp/ayxa9hWXcYG2Q5o+0OHBjBJ1yktzx45WXsPTvd95lpJVZtXqjhhcJ5lvoSekFXeYKhy9NyflSqdVbiZ88uHeSfrcgzjWRZ1R4ffBqa8RM2ZQevWVvRCa8Qt3nY99WvEcxqMBhNjVwWeNewBPtFCpCGDWhcRdZNF2o8kuomf6vQPWpYa4xJd6CReT51kYgEiO50OttbkdABpwY0hz6wuijvxjNPB2SSm+J3UQvGMPuf3pmFd2Pzg7KYxMsYmMR1fz7RKFU9cbyH/Bw)

Also tested with [the materials sandcastle](https://sandcastle.cesium.com/?src=Materials.html) and everything looks good to me. Feel free to suggest / test with more materials-type sandcastles.

# Author checklist


packages/engine/Source/Scene/Material.js
mzschwartz5, line 1: The ultimate goal is to be able to extract image loading into its own, callable function. The complex conditional statement in the original code makes this quite difficult. This commit simplifies the conditional to facilitate the end goal.

The behavior should be the same, just less convoluted logic. There IS one behavioral difference - I completely removed this check:

`!defined(material._texturePaths[uniformId])`

This statement exists for the first time this code runs - after which, this property will always be defined. Since I intend to explicitly load the images on Material construction in the next commit, this property will always be defined in this code path, and the check will no longer be necessary.

packages/engine/Source/Scene/Material.js
```diff
@@ -880,50 +880,43 @@ function createTexture2DUpdateFunction(uniformId) {
     //  they are clonable. That's why we check the url property for Resources
     //  because the instances aren't the same and we keep trying to load the same
     //  image if it fails to load.
```
mzschwartz5: test

packages/engine/Source/Scene/Material.js
```diff
@@ -327,6 +327,15 @@ function Material(options) {
```
mzschwartz5, line 327: Now that the conditional logic is simpler (last commit), this commit extracts the image loading logic into its own function, and calls it immediately after using `createTexture2DUpdateFunction` to register an updater.

The image loading logic returns a promise, which is pushed into a list of initialization promises. Nothing is done with that list in this commit, but it will be used in the future commits for the async factory constructor.

packages/engine/Specs/Scene/MaterialSpec.js
```diff
@@ -657,26 +657,21 @@ describe(
```
mzschwartz5, line 744: Had to change one unit test. It would hang forever waiting for `_loadedImages.length` to be not 0 on the second material, but because of the refactor, this never happens because both images get handled in the first render.

This is why you don't rely on internal details of a class for unit tests :)

packages/engine/Source/Scene/Material.js
```diff
@@ -973,42 +973,53 @@ function createCubeMapUpdateFunction(uniformId) {
     }
```
mzschwartz5, line 1041: Similarly to what I did for the 2D texture image loading logic, this commit extracts the cubemap image loading into a callable function which returns a promise. The `createUniforms` function now immediately calls this image loader and pushes this promise onto the initialization promises list.

And the updater function returned by `createCubeMapUpdateFunction` now also uses this image loader function, but doesn't do anything with the promise - instead, as before, it polls the status of the image loading in the update loop. The promise is just for the async factory constructor coming in the next commit.

packages/engine/Source/Scene/Material.js
mzschwartz5, line 327: Now that the conditional logic is simpler (last commit), this commit extracts the image loading logic into its own function, and calls it immediately after using `createTexture2DUpdateFunction` to register an updater.

The image loading logic returns a promise, which is pushed into a list of initialization promises. Nothing is done with that list in this commit, but it will be used in the future commits for the async factory constructor.

mzschwartz5:
## Adding some context for this PR on a commit-by-commit basis:

### Commit 1: [Refactors conditional logic of material image loading](https://github.com/CesiumGS/cesium/pull/12767/commits/84d81c3a832b278008f7049601be80b4c4bc8d11)

The ultimate goal is to be able to extract image loading into its own, callable function. The complex conditional statement in the original code makes this quite difficult. This commit simplifies the conditional to facilitate the end goal.

The behavior should be the same, just less convoluted logic. There IS one, temporary, behavioral difference - I completely removed this check:

`!defined(material._texturePaths[uniformId])`

This statement exists for the first time this code runs - after which, this property will always be defined. Since I intend to explicitly load the images on Material construction in the next commit, this property will always be defined in this code path, and the check will no longer be necessary.

### Commit 2: [Refactors material texture 2D image loading into own function.](https://github.com/CesiumGS/cesium/pull/12767/commits/2628dfe5c38b76a4038c406998f3bd5c266b1479)

Now that the conditional logic is simpler (last commit), this commit extracts the image loading logic into its own function, and calls it immediately after using `createTexture2DUpdateFunction` to register an updater.

The image loading logic returns a promise, which is pushed into a list of initialization promises. Nothing is done with that list in this commit, but it will be used in the future commits for the async factory constructor.

### Commit 3: [Refactors cubemap image loading into own function](https://github.com/CesiumGS/cesium/pull/12767/commits/76046ceb4c3227ca61e8cf411f6571862b84f01f)

Similarly to what I did for the 2D texture image loading logic, this commit extracts the cubemap image loading into a callable function which returns a promise. The `createUniforms` function now immediately calls this image loader and pushes this promise onto the initialization promises list.

And the updater function returned by `createCubeMapUpdateFunction` now also uses this image loader function, but doesn't do anything with the promise - instead, as before, it polls the status of the image loading in the update loop. The promise is just for the async factory constructor coming in the next commit.


Also: had to change one unit test. It would hang forever waiting for `_loadedImages.length` to be not 0 on the second material, but because of the refactor, this never happens because both images get handled in the first render.

This is why you don't rely on internal details of a class for unit tests :)

### Commit 4: [Adds async factory method for Material class](https://github.com/CesiumGS/cesium/pull/12767/commits/ecff7ce5de018c8697d864393baa93da286c4990)

We finally get to the main purpose of this PR - making an async factory constructor. Behaves very similarly to fromType, but waits on the initialization promises (including those from submaterials), and returns a promise itself that can be awaited.

Note: there are two (I think?) bugs in `getInitializationPromises` in this commit. Fixed in later commit.

### Commit 5 - just adds JS docs

### Commit 6: [Adds unit tests for async material constructor](https://github.com/CesiumGS/cesium/pull/12767/commits/59161b0d50cb5339f0f9288ba3108a425c2da69d)

**Note**: while writing unit tests, I learned that `fromType` / the material cache does *not* support submaterials! This commit also adds support for it by making two small adjustments:
1. Create uniforms and submaterials *after* retrieving the cached material
2. Populate the material `.fabrics` member from the `options.fabrics` passed into the constructor.

I'm a *little* nervous that this may introduce some bugs, but the unit tests and my own testing didn't reveal any.

### Commit 7 - [Material unit test update and CHANGES.md](https://github.com/CesiumGS/cesium/pull/12767/commits/781c7bac911250673efbbc1f0bf058488732a1c6)

Updates the material unit tests to use the new async constructor. Ironically, only one test can actually make use of it. The others that test images are all testing "what if we *change* an image...", which out of scope of what the async constructor does.

### Commit 8 - [Adds Material constructor options to fromType and fromTypeAsync methods](https://github.com/CesiumGS/cesium/pull/12767/commits/7b12d896bc6a78d44ac851aad3c565e7e2fe6089)

As title says, adds the options object to these methods (for min/mag filters, strict, and translucent options). Updates unit test to use these instead of setting internal fields.

packages/engine/Specs/Scene/MaterialSpec.js
```diff
@@ -630,55 +630,115 @@ describe(
...
-            });
-          });
-        });
+      renderMaterial(materialLinear, ignoreBackground, function (rgba) {
+        expect(rgba).toEqualEpsilon(purple, 1);
+      });
+      renderMaterial(materialNearest, ignoreBackground, function (rgba) {
+        expect(rgba).not.toEqualEpsilon(purple, 1);
+      });
     });
```
mzschwartz5, line 706: Part of the goal of this PR was to make the Material unit tests more concise and less reliant on internal fields. Using `fromTypeAsync` definitely makes this unit test more readable, but it still relies on the internal filter fields, because these properties are not copied over when constructing "by type", and thus have to be set manually.

packages/engine/Specs/Scene/MaterialSpec.js
```diff
@@ -630,55 +630,115 @@ describe(
       });
     });
 
-    it("creates material with custom texture filter", function () {
-      const materialLinear = new Material({
+    it("creates a material using fromTypeAsync", async function () {
+      const material = await Material.fromTypeAsync("Color");
+      renderMaterial(material);
```
javagl, line 597: Should this call (and the one in the test below) check the rendererd color using a callback? Maybe this one is _only_ about the basic creation, but the one below could probably do a check for 'blue'.

packages/engine/Specs/Scene/MaterialSpec.js
```diff
@@ -630,55 +630,115 @@ describe(
...
+      renderMaterial(materialFromTypeAsync, false, function (rgba) {
+        expect(rgba).toEqual([0, 255, 255, 255]); // Expect cyan from green + blue
+      });
+    });
+
+    it("creates material with custom texture filter", async function () {
+      const materialLinear = await Material.fromTypeAsync("DiffuseMap", {
+        image: "./Data/Images/BlueOverRed.png",
+      });
+      materialLinear._minificationFilter = TextureMinificationFilter.LINEAR;
```
javagl: I've seen the comment below. But this is accessing a private variable. While this is can, of course, be OK for a spec, it raises the question of how users could create such a material with these filter options asynchronously. 

It may be a bit of a stretch, but ... could it make sense to have some `fromOptionsAsync` that receives the _full_ options that can be passed to the material constructor?

javagl (commented):
I had a look at the resulting code state and the individual commits, and the changes look reasonable for me. I have not thoroughly and deeply checked whether the changes from pulling out the 'loader' functions really do not affect the behavior, but from what I've seen (and the descriptions), there wasn't anything that would be obvious to break in any way.

Two inlined comments about the specs. I'm not sure whether something like that `fromOptionsAsync` could make sense...

packages/engine/Specs/Scene/MaterialSpec.js
mzschwartz5: That's a good point- the one below should probably at least check the color.

packages/engine/Specs/Scene/MaterialSpec.js
mzschwartz5: > could it make sense to have some `fromOptionsAsync` that...

I don't see why not, honestly. I was mostly following the pattern set out by `Material.fromType`, which has the same shortcomings. But maybe *both* should have an options field (it wouldn't be a breaking change if we added a new optional field to `.fromType`, right?) 

I mean, ideally, the `.fromType` func should have just accepted the full `options`, including fabric, from the get-go, but this is how we'd add the behavior without breaking the interface.

mzschwartz5: The additional arg would basically be the Material options object, minus the fabric part which comes from the type (and optional uniforms) passed in. 

image

ggetz: Hm, I'm not a fan of having to call `Material.fromType(type, undefined, options)` if there are no uniforms.

But, given we're in here and doing some major cleanup, I think it makes sense to re-evaluate the existing API from first principles. For instance, 

- We can create new static constructor functions as needed, synchronous or asynchronous
- We can expose some options via getter/setters if it makes sense
- We could compare the Material APIs of [ThreeJS](https://threejs.org/docs/#api/en/constants/Materials) and [BablyonJS](https://doc.babylonjs.com/features/featuresDeepDive/materials/advanced/work/) as a reference point

(I'm taking a review pass now, and will add other thoughts here if they come up once I look more holistically at the code.)

ggetz: Overall, the biggest complication I'm seeing is that we need to pass a `context` option to create a `Texture`, meaning we have some limitations on when we can actually create a texture.

Going through some other parts of the API, I noticed `TextureUniform` seems to be handling the aggregation of options needed for creating a texture. Maybe we could make use of that and allow it to be passed as part of the existing `uniforms` object? In the case where things are synchronous, we can resolve the resource in the `update` function. In the case of asynchronous loading, we can resolve or throw before construction.

ggetz:
I pushed up a some small test cleanup. It was not strictly related to changes from this PR, but I noticed some [code re-use that was not following jasmine's recommended best practices](https://jasmine.github.io/tutorials/sharing_behaviors).

packages/engine/Specs/Scene/MaterialSpec.js
mzschwartz5: > I noticed TextureUniform seems to be handling the aggregation of options needed for creating a texture. Maybe we could make use of that and allow it to be passed as part of the existing uniforms object? 

I'm not exactly sure what the suggestion is here, but it seems strange to me to cram unrelated properties into the uniforms object. It would mismatch the top-level options schema, but more importantly, we'd then have to parse out "which of the fields in this object are uniforms and which are other options" - if I'm understanding correctly.

mzschwartz5: I just pushed two commits, here's what I've done to address the concerns in this thread and some offline comments:

1. I removed the `options` field both from `fromType` and `fromTypeAsync`. Instead, I added setters to the Material API for min and mag filters. We can consider adding setters for other options as well (some, like `strict`, only apply at construction, so it's pointless to have a setter).
2. I added tracking of errors on initialization and now rethrow an error if one exists, in the `fromTypeAsync` method.

And here's something else to consider, based on offline discussion:

> As it stands, a Material allows you to pass in options for its texture sampler (namely: min/mag filter). However, a Material can have multiple textures - but only one sampler. The sampler should really be a property of the material.
>
> After reading up on the [Material fabric json schema](https://github.com/CesiumGS/cesium/blob/master/Documentation/Schemas/Fabric/Material.schema.json), I feel that the most sensible way to support this would be to add a property on the schema, "samplers", which would map each texture name to an object of sampler settings.
>
> Then, in the Material API, if that mapping is defined in the fabric, use it. Otherwise, default to the Material-level setting.

If we add to the fabric schema in this way, it wouldn't require any changes to the signatures of `fromType` or `fromTypeAsync`. So if we can agree on that path forward, we're good to push these changes without further consideration.

packages/engine/Source/Scene/Material.js
```diff
@@ -327,12 +327,58 @@ function Material(options) {
...
+  this._initializationPromises = [];
+
+  /**
+   * An error that occurred in async operations during material initialization.
+   * Only one error is stored.
+   *
+   * @type {Error|undefined}
+   * @private
+   */
+  this._initializationError = undefined;
```
mzschwartz5, line 346: If we ever target ES2021 or higher, we could use an `AggregateError` here.

PR #12686, opened 2025-06-23
Removed request: ^2.79.0 from package.json

easymahaffey:
# Description

Removed "request" from the package.json

 node-fetch was the replacement for request in gulpfile.js

Screenshot 2025-06-22 165923

## Issue number and link
Plan to remove request lib #11095, https://github.com/CesiumGS/cesium/issues/11095

https://github.com/CesiumGS/cesium/issues/11095

## Testing plan

Rerun package.json to confirm there is no conflict. To reproduce error add "request: ^2.79.0" on line 99 of package.json.
# Author checklist

- [ ] I have submitted a Contributor License Agreement
- [x ] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [x ] I have performed a self-review of my code

Issue #11095, opened 2023-02-21
Plan to remove `request` lib


onsummer:
Hi, I am planning to remove the old dependency `request`, since it is deprecated for 2 years, and I prefer to use `node-fetch` to instead of it.

> Here was an discussion about what libs can alternative with `request`: [Alternative libraries to request #3143](https://github.com/request/request/issues/3143)

There are two things confused me:

- What type and value may `proxy` param is? Is it a URL string? 
- How to handle error

```javascript

```

Will be:

```javascript
let statusCode = 500;
fetch(remoteUrl.toString(), {
  headers: filterHeaders(req, req.headers),
  agent: proxy // I do not know whether correct or not to set proxy here
}).then((responseStream) => {
  statusCode = responseStream.status;
  res.header(filterHeaders(req, responseStream.headers));
  return responseStream.arrayBuffer(); // use `arrayBuffer()` to pipe responseBody in raw bytes format
}).then((responseBody) => {
  res.status(statusCode).send(responseBody);
}).catch((err) => {
  // How to handle error?
})
```

Thanks!

onsummer:
For code preview: [server.js#L474](https://github.com/onsummer/cesium/blob/use-node-fetch/server.js#L474)

ggetz:
Hi @onsummer! We agree node-fetch would be the best replacement here, which was used as a replacement for `request` in `gulpfile.js`.

ggetz:
We removed "request" from all of our tooling, but not the package.json for some reason. Should be a simple cleanup.

mzschwartz5:
Using `depcheck`, looks like "request" is still in use - only in `server.js` as far as I can tell (by searching for "import request").

So I don't think we can remove the dependency quite yet, though it's low urgency to move `server.js` off "request" as it's for local development.

ggetz:
Addressed in https://github.com/CesiumGS/cesium/pull/12686. Thanks @easymahaffey and @mzschwartz5!

PR #12773, opened 2025-07-29
Removes the proxy endpoint from the server.js file

mzschwartz5:
# Description

It's generally unused and it's the only code that depends on the request package which is deprecated. Now we can remove the "request" pacakge a la https://github.com/CesiumGS/cesium/pull/12686

## Issue number and link

Original issue - https://github.com/CesiumGS/cesium/issues/11095

## Testing plan

Started up the local server to make sure everything else generally still works.

# Author checklist

### DIFF
--- a/server.js
+++ b/server.js
@@ -30,14 +30,6 @@ const argv = yargs(process.argv)
       type: "boolean",
       description: "Run a public server that listens on all interfaces.",
     },
-    "upstream-proxy": {
-      description:
-        'A standard proxy server that will be used to retrieve data.  Specify a URL including port, e.g. "http://proxy:8000".',
-    },
-    "bypass-upstream-proxy-hosts": {
-      description:
-        'A comma separated list of hosts that will bypass the specified upstream_proxy, e.g. "lanhost1,lanhost2"',
-    },
     production: {
       type: "boolean",
       description: "If true, skip build step and serve existing built files.",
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -319,7 +318,6 @@ async function generateDevelopmentBuild() {
     return result;
   }
 
-  const upstreamProxy = argv["upstream-proxy"];
```
mzschwartz5, line 322: Unused variable / satisfying the linter

server.js
```diff
@@ -346,54 +344,6 @@ async function generateDevelopmentBuild() {
     next();
   });
 
-  //eslint-disable-next-line no-unused-vars
-  app.get("/proxy{/*remote}", function (req, res, next) {
```
ggetz, line 350: The `app.param("remote"...` block above is now unused (the param is the `remote` portion of the url specified here).

server.js
```diff
@@ -346,54 +344,6 @@ async function generateDevelopmentBuild() {
...
-    if (!remoteUrl) {
-      return res.status(400).send("No url specified.");
-    }
-
-    if (!remoteUrl.protocol) {
-      remoteUrl.protocol = "http:";
-    }
-
-    let proxy;
-    if (upstreamProxy && !(remoteUrl.host in bypassUpstreamProxyHosts)) {
```
ggetz, line 370: I think this is the only place where `bypassUpstreamProxyHosts` is used. Please check if it and the associated command line parameters need some cleanup.

server.js
ggetz: This will also need to be removed from the `yargs` option in the lines above.

ggetz (commented):
Thanks @mzschwartz5! Good start, just some additional related cleanup needed.

mzschwartz5:
> Thanks @mzschwartz5! Good start, just some additional related cleanup needed.

I was a bit wary about doing a lot of clean up, but it *does* look like those args are unused (and were, already), as was the `filterHeaders` function. Removed all of that now.

ggetz:
Looks good! Thanks for the cleanup @mzschwartz5!

Issue #12709, opened 2025-07-01
Allow passing tileset options to reality data tilesets


javagl:
### Feature

The function to create tilesets from reality data IDs currently [uses `Cesium3DTileset.fromUrl` to create the tileset](https://github.com/CesiumGS/cesium/blob/a1da48ac745ae509aef450a9ca5f92f3ede17b17/packages/engine/Source/Scene/ITwinData.js#L138), and passes in some default constructor options for the tileset. This means that it is not possible to pass in _custom_ options into a reality data tileset.

We should consider to allow users to pass in additional options through the `createTilesetForRealityDataId` function.

The [tileset `ConstructorOptions`](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#.ConstructorOptions) are ... overwhelming, and one could make the strong case to try and _not_ increase the API surface area here. Specifically, there may be options that ...

1. do not _have_ to be constructor options. but can be set on the resulting tileset directly 
2. **or** that do not make sense for a reality data tilesets

and it might be worthwhile to identify the options that should _really_ be available via the `createTilesetForRealityDataId` function. But for now, this issue is rather about the broader question whether it should be possible to pass in additional options to this function _at all_.

jjspace:
iirc my motivation to _not_ include a pass through object for the tileset or url constructors was that the data loading functions use optional arguments. I thought it would feel too annoying/weird to have to set multiple arguments to `undefined` just to "get to" the constructor arguments. Maybe it's worth re-evaluating and converting the existing arguments into an options object themselves so the second argument can be the optional constructor options object.

ggetz:
Would it make sense to keep the current function signature, but allow for an options object as the third(?) argument? That would mean if the third argument is defined, we check the type and parse the options accordingly.

For the options object itself, there's president to [allow all the Cesium3DTileset constructor options](https://github.com/CesiumGS/cesium/blob/a1da48ac745ae509aef450a9ca5f92f3ede17b17/packages/engine/Source/Scene/createOsmBuildingsAsync.js#L61), and to add a few context specific options. While I do agree that the sheer number of options there is overwhelming, I think for the sake of _implementation_ simplicity, it's out of scope of this particular issue to figure out a more user-friendly design.

javagl:
The current `createTilesetForRealityDataId` receives 

  - the `iTwinId` and `realityDataId` (both required)
  - the `type` and `rootDocument` (both optional)

[...]

PR #12778, opened 2025-07-30
Add support for optional changesets when loading iModels

### LINKED ISSUE #12776: Add support for iModel changesets in createTilesetFromIModelId
### Feature

iModels typically have many, many changesets during the lifecycle of the project.  Changesets can be used to represent different design options.

A typical use case is being able to switch between different design options to visualize the differences.  In order to do this, we would need support for specifying an optional changeset id when calling `createTilesetFromIModelId`.  This is an optional parameter for the [Mesh Export API](https://developer.bentley.com/apis/mesh-export/operations/get-exports/)

This would enable users to visualize different versions of their iTwin

r-veenstra:
# Description

This PR resolves #12776 by adding support for an optional `changesetId` when loading iModels from iTwins using `Cesium.ITwinData.createTilesetFromIModelId`

Changesets define different versions of an iModel, and a common use case is to be able to load different versions and compare them side by side.  The current implementation only supports loading tilesets from the latest changeset.

Note: Specifying an invalid changeset does not cause the [Mesh Export API](https://developer.bentley.com/apis/mesh-export/operations/get-exports/) to error, instead it just returns an empty list of exports.

**For discussion**
- I originally have a check for empty exports and threw and exception, however the existing unit tests indicate that if no exports are present, returning an undefined tileset is the preferred outcome, so I have removed it for now.

- What order of parameters should `Cesium.ITwinData.createTilesetFromIModelId` have?  I've currently implemented it as imodelId, tilesetOptions, changesetId so as to not break existing code.  However, iModelId, changesetId then tilesetOptions seems more natural.  Not sure on best practice here.

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12776

## Testing plan

I tested the following:

- Loading an iModel using existing methods, as per this [sandcastle](https://sandcastle.cesium.com/?src=iModel%20Mesh%20Export%20Service.html) and verified it still works
- Loading an iModel using a valid changeset, confirming a different version appeared
- Loading an iModel with an invalid changeset string, confirming an empty list of exports was returned
- Loading an iModel with a non-string changeset, confirming an exception was thrown due to a non-string being used

# Author checklist


ggetz:
Thanks @r-veenstra! I can confirm we have a CLA on file covering you. 🙂 

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -157,9 +157,12 @@ ITwinPlatform.apiEndpoint = new Resource({
  *
  * @throws {RuntimeError} If the iTwin API request is not successful
  */
-ITwinPlatform.getExports = async function (iModelId) {
+ITwinPlatform.getExports = async function (iModelId, changesetId) {
```
jjspace, line 161: Please add the new argument to the JSDoc too

packages/engine/Source/Core/ITwinPlatform.js
```diff
@@ -191,6 +194,9 @@ ITwinPlatform.getExports = async function (iModelId) {
   if (typeof CESIUM_VERSION !== "undefined") {
     resource.appendQueryParameters({ clientVersion: CESIUM_VERSION });
   }
+  if (defined(changesetId)) {
```
jjspace: I believe this should check that it's not equal to an empty string as well `""`. I recall there being problems requesting with an "empty" param (would need to re-verify)

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -35,13 +35,19 @@ const ITwinData = {};
  *
  * @param {string} iModelId The id of the iModel to load
  * @param {Cesium3DTileset.ConstructorOptions} [options] Object containing options to pass to the internally created {@link Cesium3DTileset}.
+ * @param {string} [changesetId] The id of the changeset to load, if not provided the latest changesets will be used
  * @returns {Promise<Cesium3DTileset | undefined>} A promise that will resolve to the created 3D tileset or <code>undefined</code> if there is no completed export for the given iModel id
  *
+ * @throws {RuntimeError} If no exports for the given iModel are found
```
jjspace: If the error was removed and won't be re-added please remove this as well

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -35,13 +35,19 @@ const ITwinData = {};
...
  * @throws {RuntimeError} If all exports for the given iModel are Invalid
  * @throws {RuntimeError} If the iTwin API request is not successful
  */
-ITwinData.createTilesetFromIModelId = async function (iModelId, options) {
-  const { exports } = await ITwinPlatform.getExports(iModelId);
+ITwinData.createTilesetFromIModelId = async function (
+  iModelId,
+  options,
+  changesetId,
+) {
```
jjspace: I've been thinking back and forth on this. Especially with the recent (and somewhat lengthy) discussion around handling of arguments for this API as a whole in https://github.com/CesiumGS/cesium/issues/12709

With optional arguments, there's an argument (😉) to be made that they should be ordered from most to least used. I also think it's valid to think about the "relation" between arguments. My gut says this should go `( iModelId, [changesetId], [tilesetOptions] )` as the imodel id and changeset id are the most related. However I think it's _more likely_ that people will want to use the `iModelId` and the `tilesetOptions` and only rarely the `changesetId` which means the current arrangement is fine. 

From the discussion in https://github.com/CesiumGS/cesium/issues/12709 I almost think the best solution is `( iTwinOptions, [tilesetOptions] )` where the `iTwinOptions` is `{ iModelId, [changesetId] }`. Or maybe even better a single `({ iModelId, [changesetId], [tilesetOptions] })` argument is actually best?

Future updates and "backwards compatibility" (even if this is experimental) are also worth taking into account. This order means all existing calls will "just work".

Open to some discussion on this (maybe thoughts from @javagl?). However if we need to include this for the release I'm ok with this solution and addressing _all_ the functions with https://github.com/CesiumGS/cesium/issues/12709

CC @ggetz

jjspace (commented):
First, whatever we settle on please make sure to update the `CHANGES.md` file with this change.

Second, to give some historical context. In the original PR that implemented this API, https://github.com/CesiumGS/cesium/pull/12289, we specifically decided to _not_ include support for specifying a specific changeset ([commit](https://github.com/CesiumGS/cesium/pull/12289/commits/1757f77a0878d695d80f4f4c2d24172f994adf55) where I removed it). At the time it was my understanding that a _vast_ majority of users would never need or want to request anything but the latest version of an iModel. This was based off discussion with @danieliborra.
To be clear, I'm not against including it I just want to make sure it's very intentional _why_ we're including it. If this is to support a tutorial that's also highlighting the "best practices" and intended use cases for the iTwin platform is this actually the desired workflow?

packages/engine/Source/Scene/ITwinData.js
r-veenstra: To help add context to this decision, another parameter I would want to be able to define is a custom iTwin Share Key on a per tileset basis.

A common request from existing iTwin users is to be able to visualise multiple iTwins in one application.  Currently a user can only define `Cesium.ITwinPlatform.defaultShareKey` which won't work if multiple iTwins are required.

It's also possible we would want to define a custom `Cesium.ITwinPlatform.defaultAccessToken` too if I'm using a different authentication method.  And if the Mesh Export API that this wraps grows in capability, we'd want to easily support that in the future.

So I think the structure should lend itself to having a set of `iTwinOptions` that are easily expandable.

packages/engine/Source/Core/ITwinPlatform.js
r-veenstra: Just tested this - sending an empty string as the `changesetId` doesn't cause the API request to fail.  It just returns with no export results as there is no changeset called ""

packages/engine/Source/Scene/ITwinData.js
jjspace: We discussed this offline and decided to go with the route of a _single_ options object for this API. I've updated this PR to change all the functions signatures (the core part of discussion in https://github.com/CesiumGS/cesium/issues/12709) including the new changeset id option.

jjspace:
Thanks for the updates @r-veenstra. 

As stated above we've decided to push ahead with the changes to the API signature for _all_ functions in the `ITwinData` class. I've made these changes on this branch. There _is_ a fallback that will last for 1 version to attempt to detect the old signature and use those arguments. (I'll be opening a PR to remove this shortly that should be merged next week after the release goes out.)
This addresses half of https://github.com/CesiumGS/cesium/issues/12709, the tileset options will be added in a follow up PR but will be much easier with this current change.

@ggetz would you be able to review my changes so there's a second pair of eyes on them? I think it should be good to merge if you're happy.

CC @javagl

danieliborra:
> First, whatever we settle on please make sure to update the `CHANGES.md` file with this change.
> 
> Second, to give some historical context. In the original PR that implemented this API, #12289, we specifically decided to _not_ include support for specifying a specific changeset ([commit](https://github.com/CesiumGS/cesium/pull/12289/commits/1757f77a0878d695d80f4f4c2d24172f994adf55) where I removed it). At the time it was my understanding that a _vast_ majority of users would never need or want to request anything but the latest version of an iModel. This was based off discussion with @danieliborra. To be clear, I'm not against including it I just want to make sure it's very intentional _why_ we're including it. If this is to support a tutorial that's also highlighting the "best practices" and intended use cases for the iTwin platform is this actually the desired workflow?

This is a technical problem for a few reasons:
- We are not generating tilesets for all changesets (and we won't)
- We only keep the latest 5 exports and delete the rest (e.g. one of the iModels has +150K changesets, and it is not a small one, imagine generating and storing 150K tilesets).
- You can manually generate changesets, but it generates a lot of confusion even with Bentley colleagues/products.
- ChangesetIds are available before tileset generation, so, depending from where you get the changesetId, you may not have the tileset yet.

So, my strong advice would be not to do it, especially in a tutorial.

PS. I did some calculations, and storing +150K tilesets (3GB), would cost $9,450 per month.

CHANGES.md
```diff
@@ -4,6 +4,10 @@
 
 ### @cesium/engine
 
+#### Breaking Changes :mega:
```
ggetz: This would go under "Deprecated" this release, and then "Breaking Changes" in _next_ release.

ggetz (commented):
Thanks for the updates @jjspace, everything looks good. Just one comment on the changelog.

ggetz:
Thanks for the insight @danieliborra.

We'll discuss with @r-veenstra on whether to move forward or not given this info. If we do end up moving forward, it sounds like at minimum, some additional context or disclaimers in the ref doc would be helpful for users.

jjspace:
This _should_ be good to go but [CI keeps failing](https://github.com/CesiumGS/cesium/actions/runs/16658889601/job/47151451080) on the same test, I'm not sure why any change in this PR would affect the `Cesium3DTileset` specs... That test passes on my system, @ggetz thoughts?

ggetz:
> This should be good to go but [CI keeps failing](https://github.com/CesiumGS/cesium/actions/runs/16658889601/job/47151451080) on the same test, I'm not sure why any change in this PR would affect the Cesium3DTileset specs... That test passes on my system, @ggetz thoughts?

I'm pretty much positive that's unrelated to these changes. Perhaps something became more fragile from a recent change. But test are passing in CI on the last run, and everything is passing locally for me.

danieliborra:
Just to complete my explanation, one changeset is not a file version but a list of transactions, that is why you can get so many changesets and why we don't generate a tileset per changeset.

Said that,  I checked the tutorial and it explains it very well. So, my hesitations are gone 👍

Issue #12749, opened 2025-07-18
Jagged rendering for SPZ Gaussian Splats


TASallin:
### What happened?

The rendering behavior in cesium for splats is different than in other SPZ renderers like StorySplat, it looks jagged similar to the flickering frames from issues #12733 and #12662 but without the flickering, it simply looks this way when it's 'stable'. This is on a local build of the main branch after commit 7103190fe318fe3b0df79af40f94e5e8f1443a17 ([Merge pull request](https://github.com/CesiumGS/cesium/commit/7103190fe318fe3b0df79af40f94e5e8f1443a17) https://github.com/CesiumGS/cesium/pull/12740 [from CesiumGS/gsplat-scaling](https://github.com/CesiumGS/cesium/commit/7103190fe318fe3b0df79af40f94e5e8f1443a17)) These show the same SPZ in Cesium and StorySplat respectively:

<img width="1531" height="798" alt="Image" src="https://github.com/user-attachments/assets/542a8812-aa2a-4594-bea9-87cc7a98b1ee" />

<img width="777" height="497" alt="Image" src="https://github.com/user-attachments/assets/b268d618-1ee2-4c36-99b4-c0f20097d499" />

I'm guessing it's related to the aforementioned flickering issue which I know is being worked on, but maybe some of this can help. This is a link to the spz file https://drive.google.com/file/d/1663KUJxXvQ_fzq-_64y6L5FbF9-eX4UW/view?usp=sharing 

tileset.json for transform: 
`{
  "asset": {
    "version": "1.1"
  },
  "extensions": {
    "3DTILES_content_gltf": {
      "extensionsRequired": ["KHR_spz_gaussian_splats_compression"],
      "extensionsUsed": ["KHR_spz_gaussian_splats_compression"]
    }
  },
  "extensionsUsed": ["3DTILES_content_gltf"],
  "geometricError": 65536,
  "root": {
    "boundingVolume": {
      "box": [ 0.059326172, 0.0, 0.040161133, 2047.9287, 0.0, 0.0, 0.0, 2047.9932, 0.0, 0.0, 0.0, 2047.8969 ]
    },
    "geometricError": 32768,
    "refine": "REPLACE",
    "children": [
      {
        "boundingVolume": {
          "box": [ 0.059326172, 0.0, 0.040161133, 2047.9287, 0.0, 0.0, 0.0, 2047.9932, 0.0, 0.0, 0.0, 2047.8969 ]
        },
        "content": {
[...]

PR #12790, opened 2025-08-01
Gaussian splat spherical harmonics support

### LINKED ISSUE #12756: GaussianSplatPrimitive blank render in unit tests
Currently have a disabled test in GaussianSplatPrimitiveSpec that checks the show property on the tileset. Regardless of whether it's visible the framebuffer is empty.

Data structures appear valid, draw command is generated and added to the scene command list and other tests pass, but something is preventing it from actually rendering.

keyboardspecialist:
# Description

Adds spherical harmonic support for Gaussian splats in the SPZ format. Supports degrees 1-3 and auto detects what is available. This brings full SPZ support to CesiumJS.

## Issue number and link

Fixes #12756 

## Testing plan

View the same asset tiled with 3 degrees of SH in the current main build versus the CI brand build:

[No SH Support](https://sandcastle.cesium.com/#c=jVJrT9swFP0rVj6gVGIOpbQSW0FDhUECbQdNGUGRJuO41I1rB9t9JBP/fU6cjDJt0ypF1fU959zHuZ4HLiXiGgyIoqtlMAEIY6IU0ALkYiUBFRwgpYhWMbcY6AsOEzJDK6bPKnAoUsLBCYgdkgfzp0tMxzTwp4XfHlFf+fyuiwd+z0+zh/tBcAwN6CW5TA3IL4bnuBstou0jPegMw2B+E96xKPR1FF5so7ydPi4ujkaH001UjNLh4ZTeDILs0YgNF7gzMtwyjh5u6Xhx0RmFt5thMT0ahfNjyFjvpb1YTLftg8nd12GRF8/S76biQ3QdToIkurqS3++ve1dHX1TsfIo5FlxpsKZkQ6QZhZNNvRJ4X725sYOreCC4RpQTGTv74EfMwa+fJlKazMeGGNoYzqRYfhOSJfWD29p/o722TPGYa5lbMduHpoyYnZtG0AbRxh1o/zrnoU1XysaNs9IfP3E73eND85WKoB4FKkw4gZmkS6rpmiiIksSt5S3QVqjhhRDLUOwCSojngbMsYznQcwJq64HSOSOAzoBhky1V5YU0/ZOtlkiZ9mshWJ0QtM9VVcNz7Rbq2Yys2WriWkwL7O39Iw3NWf4HpDnTSdlqqwS3GsuaxuwU7wzf3XLF/KtiNclrzF8BRhrPgWsMFrL15qRgBDLxXL8buAE7+06/qnpaoj7TZSakBivJXAg9TZYZQ5oo72mFU9MfVqrk9b2G0k/oGtDk5A/3CDAzezaZ2YqxCS1I7Jz2PYN/R2MCJZQ/j9dEMpQbSNlGf94+vbEJCGHfM2FZ9HeuFoI9Ibmj+xM
)
[SH Support](https://ci-builds.cesium.com/cesium/gsplat-sh-release/Apps/Sandcastle/index.html#c=jVJrT9swFP0rVj6gVGIOpRQJVtBQYZBAy6BpR1CkyTgudePawXYfydT/PidOBtvYtEpRda/POfdxrueBS4m4Bn2i6HIRjADCmCgFtAC5WEpABQdIKaJVzC0G+oLDhEzRkumzChyKlHBwAmKH5MHs6RLTWxr448JvD6mvfH7fxX3/0E+zh0k/OIIG9JJcpgbkF4Nz3I3m0eaR7nUGYTC7Ce9ZFPo6Ci82Ud5OH+cXB8P98Toqhulgf0xv+kH2aMQGc9wZGm4ZRw939HZ+0RmGd+tBMT4YhrMjyNjhS3s+H2/ae6P7L4MiL56l303Fh+g6HAVJdHUlv02uD68OPqvY+RhzLLjSYEXJmkgzCifreiVwUuXc2MFV3BdcI8qJjJ1d8D3m4OdPEynNy3FDDG0Mp1IsvgrJkjrhtnZfaduWKR5zLXMrZvvQlBGzc9MIWiPauAPtX+c8tM+VsnHjrPTHT9xO92jffKUiqEeBChNOYCbpgmq6IgqiJHFreQu0FWp4IcQiFG8BJcTzwFmWsRzoGQG19UDpnBFAp8CwyYaq8kKa/slGS6RM+7UQrE4I2nRV1fBcu4V6NiNrtpq4FtMCOzv/eIbmLP8D0pzpqGy1VYJbjWVNY3aKXwx/u+WK+VfFapJtzLcAI41nwDUGC9l6dVIwApl4rvMGbsDOrtOrqp42V/CJLjIhNVhK5kLoabLIGNJEeU9LnJoesVK2FAA97y21l9AVoMnJO7cJMDM7Ny/TJWMjWpDYOe15Bv8HlQmUUP58uyKSobyEzdqnNzYJIex5JnyfqYVgT0j+pvwD)

# Author checklist


javagl:
What is the SH degree of that example asset? 

In any case, I think that proper SH support can hardly be implemented (and certainly not _tested_ properly) based on the current SPZ loader, because it [does not properly decode the SH data](https://github.com/drumath2237/spz-loader/blob/b878dad378c0941b3d2155839dccbcfb4753a12d/packages/core/lib/spz-wasm/gaussianCloud.ts#L48) - _unless_ the CesiumJS code is anticipating that wrong encoding, and compensating manually, in which case we just have to hope that it will never be fixed in the SPZ loader.

keyboardspecialist:
The underlying SPZ wasm module decodes it. The raw decoded SH buffer is all we need.

Example is degree 3

javagl:
Asked for clarification at https://github.com/drumath2237/spz-loader/issues/36#issuecomment-3144752610

There is an effect that is clearly visible:

Cesium Splats Cesium

Differences to BabylonJS may be due to coordinate system differences (it's difficult)

Cesium Splats Babylon

Test data:

[Shs Example.zip](https://github.com/user-attachments/files/21549923/Shs.Example.zip)

keyboardspecialist:
Babylon makes their lives simpler by rotating the world around the model. 

The model's inverse rotation matrix is applied to the view angle which should give us the correct sampling.

One thing to keep in mind is we do apply a axis correction matrix which may contribute to what you are seeing. When I load our tiled SPZ into Babylon its sideways. That 90 degree rotation might be what you are seeing?

javagl:
I think that the different coordinate system conventions can _likely_ explain some differences. 

Until now, I didn't dive into the maths behind the spherical harmonics. Yes, it's a bunch of coefficients that are mushed together with the view direction and eventually yield a "color". But I don't thoroughly understand the 'meaning of the values'. For example, I'd really like to create such a unit cube with spherical harmonics that looks 

1. red from the right, and cyan from the left 
2. green from the top and magenta from the bottom
3. blue from the front and orange from the back

This could help to _see_ whether the SHs are taken into account correctly, and whether there are any orientation issues. (If someone knows, from the tip of the head, what the SH values would have to be for that, I'd create an example - until then, this is scheduled in the 'When I have way too much time on my hands'-section of my TODO list).

jjhembd:
I'll put in a +1 for the "unit cube" test data. Without that, we don't have any way of knowing if the SH are rendered correctly, or wrongly but canceled by an opposite error in the data generation, or just wrong.

Here's what I see for the 2 Sandcastles.
No SH (before the PR):
image

With SH (after this PR):
image

At first glance it looks like a change in the scene lighting direction/intensity.

javagl:
Creating predefined test data for this is a bit tricky, though. Deriving the right values from the shader code is close to impossible -  there's a reason why there is some magic "iterative learner" blackbox generating these coefficients in "real" applications. I considered a lazy/sneaky approach, and just created a GLB of such a (solid) cube, recorded a video of that spinning it in all directions, and uploaded that to poly.cam, but ... that simply refuses to generate splats from that. (Even if it worked, it might not even possible to derive the right values from that). 

A bit of trial and error gave the "red-cyan" effect, but only that, and only along the diagonal...

Splats Cube

... so I think that it might be _possible_, and maybe even "easy" for someone who knows what he's doing. Not me, at least, not at the moment.

ggetz:
Hi @keyboardspecialist, what's the status here? Any input on the testing data discussion above?

CHANGES.md
```diff
@@ -23,6 +23,7 @@
 - Expand the CustomShader Sample to support real-time modification of CustomShader. [#12702](https://github.com/CesiumGS/cesium/pull/12702)
 - Add wrapR property to Sampler and Texture3D, to support the newly added third dimension wrap.[#12701](https://github.com/CesiumGS/cesium/pull/12701)
 - Added the ability to load a specific changeset for iTwin Mesh Exports using `ITwinData.createTilesetFromIModelId` [#12778](https://github.com/CesiumGS/cesium/issues/12778)
+- Added spherical harmonics support for Gaussian Splats. Supports degrees 1, 2, and 3 in the SPZ format.
```
jjhembd: Where can we find a spec or description of the SPZ format? Is the format stable?
A link to the spec, perhaps in `GltfVertexBufferLoader`, could go a long way to shortening the spin-up time for future maintainers.

javagl: The thing that is closest to a spec is the README of https://github.com/nianticlabs/spz . The actual truth is in the code. Request for clarification are in https://github.com/nianticlabs/spz/issues/42 . This does not yet cover the details of SHs and their layout, but maybe what's in the README is enough... (?)

wongshek:
@jjhembd Here is a small ply data where the correctness of higher-order SH  rendering can be clearly perceived. It would be great if it could help.

[simple_board.ply.zip](https://github.com/user-attachments/files/21753459/simple_board.ply.zip)

Below is a comparison screenshot from Supersplat.

image
image

javagl:
>  For example, I'd really like to create such a unit cube with spherical harmonics that looks
> 
> red from the right, and cyan from the left
> green from the top and magenta from the bottom
> blue from the front and orange from the back
> 
> This could help to see whether the SHs are taken into account correctly, and whether there are any orientation issues. (If someone knows, from the tip of the head, what the SH values would have to be for that, I'd create an example - until then, this is scheduled in the 'When I have way too much time on my hands'-section of my TODO list).

I guess that it's hard to just "come up" with the proper SH coefficients for that. 

So I ported the GLSL code from the [GaussianSplattingViewer](https://github.com/limacv/GaussianSplattingViewer) to Java, took the view- and projection matrix and the camera positions for the top/bottom/left/right/front/back view configuration as the inputs, the desired color as the outputs, and interpreted the whole thing as an optimization problem that I fed into the [Apache BOBYQAOptimizer](https://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/org/apache/commons/math3/optim/nonlinear/scalar/noderiv/BOBYQAOptimizer.html). 
Exactly what any sane person would do 🤡 

The result as ASCII PLY, SPZ, GLB, and a matching tileset JSON and Sandcastle:

[Splat SH orientation experiment 2025-08-13.zip](https://github.com/user-attachments/files/21756270/Splat.SH.orientation.experiment.2025-08-13.zip)

This is what it looks like:

SplatShTest

- blue from the front, yellow from the back
- green from the top, magenta from the bottom
- red from the right, cyan from the left

(This is rendered with the viewer from [JSplat](https://github.com/javagl/JSplat). BabylonJS shows the same for SPZ. For PLY, there seems to be an orientation issue, or I'm writing the coefficients in the wrong order... who knows what's "right" or "wrong" here...)

I tried to check the different views in CesiumJS:

Cesium Splats SH Test

Looks about right. The bottom is missing - let's just assume it's magenta.

CHANGES.md
keyboardspecialist: I should reword that... we load them from SPZ currently, but the rendering side is not tied to SPZ.

ggetz:
@keyboardspecialist @javagl @jjhembd Are we comfortable that the test case above proves out the CesiumJS implementation? If so, can we add it as a unit test?

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -311,6 +326,172 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+/**
+ * Determine Spherical Harmonics degree and coefficient count from attributes
+ * @param {Array} attributes - The list of attributes.
+ * @returns {Object} An object containing the degree (l) and coefficient (n).
+ */
+function degreeAndCoefFromAttributes(attributes) {
+  const prefix = "_SH_DEGREE_";
+  const shAttributes = attributes.filter((attr) =>
+    attr.name.startsWith(prefix),
+  );
```
weegeekps: You need to update this to handle both `KHR_gaussian_splatting:SH_DEGREE_` and `_SH_DEGREE_`.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -311,6 +326,172 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+  const prefix = "_SH_DEGREE_";
+  const separator = "_COEF_";
+
+  const lStart = prefix.length;
+  const coefIndex = attribute.indexOf(separator, lStart);
+
+  const l = parseInt(attribute.slice(lStart, coefIndex), 10);
+  const n = parseInt(attribute.slice(coefIndex + separator.length), 10);
+
+  return { l, n };
```
weegeekps: You also need to update this to handle both `KHR_gaussian_splatting:SH_DEGREE_` and `_SH_DEGREE_`.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -311,6 +326,172 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+ */
+function packSphericalHarmonicData(tileContent) {
+  const degree = tileContent.shDegree;
+  const coefs = tileContent.shCoefficientCount;
+  const totalLength = tileContent.pointsLength * (coefs * (2 / 3)); //3 packs into 2
+  const packedData = new Uint32Array(totalLength);
+
+  const shAttributes = tileContent.splatPrimitive.attributes.filter((attr) =>
+    attr.name.startsWith("_SH_DEGREE_"),
+  );
```
weegeekps: You also need to update this to handle both `KHR_gaussian_splatting:SH_DEGREE_` and `_SH_DEGREE_`.

weegeekps (commented):
Found some places where you need to update the SH prefixes to support the new and the old extensions.

keyboardspecialist:
> @keyboardspecialist @javagl @jjhembd Are we comfortable that the test case above proves out the CesiumJS implementation? If so, can we add it as a unit test?

I think this will work. I can put it into a test.

ggetz:
Thanks @keyboardspecialist, I'm taking a review pass on the code now.

For context, my understanding that this PR now also accounts for the new standard where glTF attribute names use extension-specific namespace prefixes. And this _in additional to_ supporting the older style of prefixing attribute names with an underscore `_`. So everything here should be backwards compatible and now just account for cases where Spherical Harmonic data is already present, and when using the new attribute naming.

> One thing to keep in mind is we do apply a axis correction matrix which may contribute to what you are seeing. When I load our tiled SPZ into Babylon its sideways. That 90 degree rotation might be what you are seeing?

@keyboardspecialist I have now seen some confusion about this from a few users. I assume this axis correction is accounting for the up-axis difference from glTF to 3D Tiles. I want to make sure I'm clear on this so I'm responding to users correctly.

1. Is it indeed the glTF to 3D Tiles up-axis correction?
2. Is the axis correction "baked-in" via the tiler, or a correction we're applying at runtime? I believe we are doing the latter everywhere else and I'd want to make sure we're being consistent.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -78,6 +79,20 @@ function GaussianSplat3DTileContent(loader, tileset, tile, resource) {
...
+   * @private
+   */
+  this.shDegree = 0;
+
+  /**
+   * The number of spherical harmonic coefficients used for the Gaussian splats.
+   * @type {number}
+   * @private
+   */
+  this.shCoefficientCount = 0;
```
ggetz: Please use `sphericalHarmonics` rather than `sh` for property names.

As per the coding guide, we avoid abbreviations. But more importantly, we already use the naming `sphericalHarmonicCoefficients` in `ImageBasedLighting.js` so we should be consistent across the codebase.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -78,6 +79,20 @@ function GaussianSplat3DTileContent(loader, tileset, tile, resource) {
    * @private
    */
   this._transformed = false;
+
+  /**
+   * The degree of the spherical harmonics used for the Gaussian splats.
+   * @type {number}
+   * @private
+   */
+  this.shDegree = 0;
```
ggetz: Do we need both `degree` and `coefficient` as independent properties? It's possible to determine one based on the other, no?

ggetz:
@keyboardspecialist It looks like this PR contains some changes from https://github.com/CesiumGS/cesium/pull/12843. Could you please merge main into this branch so we're reviewing code specific to SH?

ggetz (changes requested):

javagl:
>> One thing to keep in mind is we do apply a axis correction matrix which may contribute to what you are seeing. When I load our tiled SPZ into Babylon its sideways. That 90 degree rotation might be what you are seeing?
> 
> I have now seen some confusion about this from a few users. I assume this axis correction is accounting for the up-axis difference from glTF to 3D Tiles. I want to make sure I'm clear on this so I'm responding to users correctly.

I think that the context of the first quote was _very_ specifically the different appearances of the SH colors. When something is supposed to be "red from the front and green from the top", everything depends on what "front" and "top" _are_. 

But more generally, there are two aspects of ~"coordinate systems" standing out:

1. there are cases where a whole model is rotated by 90°. This was reported at different places and in different contexts (sometimes in the forum, referring to ion output, and to some extent tracked in internal issues). This comes _on top_ of the usual differences of up-axis conventions of glTF and 3D Tiles, and one root cause is that there already are [different conventions for PLY](https://github.com/mkkellogg/GaussianSplats3D/issues/47#issuecomment-2891611215). 
2. once upon a time, there was a state where _the splats themself_ had been rendered incorrectly, depending on certain matrices in the glTF and the tileset. Specifically, it was once necessary to 1. insert an axis conversion in glTF **and** 2. **undo** this axis conversion in the tileset (!). An example is shown in https://github.com/CesiumGS/cesium/issues/12749#issuecomment-3094441435 (but I haven't checked with the latest state)

Now one could argue that this is independent of this PR. After all, _this_ PR is only about spherical harmonics. When there's an issue with the orientation, then you can just rotate the splats. Just use the right matrix at the right place. Well... it's not that easy. On top of the (always confusing and error prone) differences of up-axis-conventions, "rotating splats" also involves "rotating the spherical harmonics (~'coefficients')". Have a look at [this `rotateSH` function](https://github.com/andrewwillmott/sh-lib/blob/8821cba4acc2273ab20417388df16bd0012f0760/SHLib.cpp#L1090) to see what this _might_ entail.

keyboardspecialist:
Some of the axis issues were due to the old SPZ loader that always rotated the incoming data around X. 

Indeed, rotating spherical harmonics is prohibitively complex and expensive at runtime which is why we avoid it. When we sample we apply the model rotation to the view, so we sample like it's in its original orientation.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
keyboardspecialist: You are correct it can be derived, but it's nice to not have to and keeps things more explicit. They are used in different ways when making checks and indexing.

keyboardspecialist:
> Thanks @keyboardspecialist, I'm taking a review pass on the code now.
> 
> For context, my understanding that this PR now also accounts for the new standard where glTF attribute names use extension-specific namespace prefixes. And this _in additional to_ supporting the older style of prefixing attribute names with an underscore `_`. So everything here should be backwards compatible and now just account for cases where Spherical Harmonic data is already present, and when using the new attribute naming.
> 
> > One thing to keep in mind is we do apply a axis correction matrix which may contribute to what you are seeing. When I load our tiled SPZ into Babylon its sideways. That 90 degree rotation might be what you are seeing?
> 
> @keyboardspecialist I have now seen some confusion about this from a few users. I assume this axis correction is accounting for the up-axis difference from glTF to 3D Tiles. I want to make sure I'm clear on this so I'm responding to users correctly.
> 
> 1. Is it indeed the glTF to 3D Tiles up-axis correction?
> 2. Is the axis correction "baked-in" via the tiler, or a correction we're applying at runtime? I believe we are doing the latter everywhere else and I'd want to make sure we're being consistent.

1. Yes
2. It's applied at runtime same as the Model pipeline

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
ggetz: Could we set one and use a getter function for the other? I'm mostly concerned about these values getting out of sync since they can both be set.

ggetz: Oh, I see now they are set privately and only exposed via the getter. Should all be fine then. 👍

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+   * @experimental This feature is using part of the 3D Tiles spec that is not final and is subject to change without Cesium's standard deprecation policy.
+   */
+  packedSphericalHarmonicsData: {
+    get: function () {
+      return this._packedSphericalHarmonicsData;
+    },
+  },
 });
 
+GaussianSplat3DTileContent.tilesetHasGaussianSplattingExt = function (tileset) {
```
ggetz: I think this was missed during the merge. The function in main is called `tilesetRequiresGaussianSplattingExt`.

CHANGES.md
```diff
@@ -17,10 +17,7 @@
 #### Additions :tada:
 
 - Added support for the [EXT_mesh_primitive_restart](https://github.com/KhronosGroup/glTF/pull/2478) glTF extension. [#12764](https://github.com/CesiumGS/cesium/issues/12764)
-
```
ggetz, line 23: Why was this section removed? I think it still relevant given the deprecation warning in the code.

packages/engine/Source/Scene/VertexAttributeSemantic.js
```diff
@@ -286,7 +286,7 @@ VertexAttributeSemantic.getGlslType = function (semantic) {
  * @param {VertexAttributeSemantic} semantic The semantic.
  * @param {number} [setIndex] The set index.
  *
- * @returns {string} The variable name.
+ * @returns {string} The varia_SCALEble name.
```
ggetz:
```suggestion
 * @returns {string} The variable name.
```

Typo?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -786,6 +873,31 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
...
+              switch (tile.content.sphericalHarmonicsDegree) {
+                case 1:
+                  coefs = 9;
+                  break;
+                case 2:
+                  coefs = 24;
+                  break;
+                case 3:
+                  coefs = 45;
+              }
```
ggetz, line 896: Should we throw an error here in case the SH is not `1`, `2`, or `3`? Or is that unlikely to happen (both with our tiler or third-party ones)?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -148,6 +169,14 @@ function GaussianSplatPrimitive(options) {
    * @see {@link GaussianSplatTextureGenerator}
    */
   this.gaussianSplatTexture = undefined;
+
+  /**
+   * The texture used to store the spherical harmonics coefficients for the Gaussian splats.
+   * @type {undefined|Texture}
+   * @private
+   */
+  this.gaussianSplatSHTexture = undefined;
```
ggetz: How about:

```suggestion
  this.sphericalHarmonicsTexture = undefined;
```

To avoid the abbreviation?

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+  const l = parseInt(attribute.slice(lStart, coefIndex), 10);
+  const n = parseInt(attribute.slice(coefIndex + separator.length), 10);
+
+  return { l, n };
+}
+
+/**
+ * Packs spherical harmonic data into half-precision floats.
+ * @param {*} data - The input data to pack.
+ * @param {*} sphericalHarmonicsDegree - The spherical harmonic degree.
```
ggetz: These look mis-matched to the function parameter in the code.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+      half = (sign << 15) | (newExponent << 10) | (mantissa >>> 13);
+    }
+  }
+
+  return half;
+}
+
+/**
+ * Extracts the spherical harmonic degree and coefficient from the attribute name.
+ * @param {String} attribute - The attribute name.
```
ggetz:
```suggestion
 * @param {string} attribute - The attribute name.
```

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+ * @returns {Object} An object containing the degree (l) and coefficient (n).
```
ggetz:
```suggestion
 * @returns {object} An object containing the degree (l) and coefficient (n).
```

Primitive types should always be lowercase.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+    case 8:
+      return { l: 2, n: 24 };
+    case 15:
+      return { l: 3, n: 45 };
+  }
+}
+
+/**
+ * Converts a 32-bit floating point number to a 16-bit floating point number.
+ * @param {Float} float32 input
```
ggetz:
```suggestion
 * @param {number} float32 input
```

`Float` is not a type in JS.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+ * @param {*} sphericalHarmonicsDegree - The spherical harmonic degree.
+ * @returns {Uint32Array} - The packed data.
+ */
+function packSphericalHarmonicsData(tileContent) {
+  const degree = tileContent.sphericalHarmonicsDegree;
+  const coefs = tileContent.sphericalHarmonicsCoefficientCount;
+  const totalLength = tileContent.pointsLength * (coefs * (2 / 3)); //3 packs into 2
+  const packedData = new Uint32Array(totalLength);
+
+  const shAttributes = tileContent.splatPrimitive.attributes.filter((attr) =>
```
ggetz: I don't think the `attributes` property was properly defined in the `GaussianSplatPrimitive` class. Looks like it was getting set as a side-effect of a helper function in `ModelUtility`. Please add and document it.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -347,8 +368,215 @@ Object.defineProperties(GaussianSplat3DTileContent.prototype, {
...
+  const prefix = attribute.startsWith("KHR_gaussian_splatting:")
+    ? "KHR_gaussian_splatting:"
+    : "_";
+  return `${prefix}SH_DEGREE_`;
+}
+
+/**
+ * Determine Spherical Harmonics degree and coefficient count from attributes
+ * @param {Array} attributes - The list of attributes.
+ * @returns {Object} An object containing the degree (l) and coefficient (n).
```
ggetz:
```suggestion
 * @param {VertexAttributeSemantic[]} attributes - The list of attributes.
 * @returns {object} An object containing the degree (l) and coefficient (n).
```

ggetz (changes requested):

packages/engine/Source/Scene/GaussianSplatPrimitive.js
keyboardspecialist: I think it's ok not to check here since the data should be checked and ready for rendering. 3 is the max anyone has ever trained that I'm aware of. I could see adding a check in the loader and notifying the user. 

With our data coming from SPZ it shouldn't be an issue.

packages/engine/Source/Scene/VertexAttributeSemantic.js
keyboardspecialist: wayward paste? Weird!

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
keyboardspecialist: It's defined in the GS content class. Very ambiguous, but it's a reference to the glTF primitive. I renamed it for better clarity.

ggetz:
@keyboardspecialist Ci is failing after your last commit. What's left here? Is it a relatively small fix that can be addressed quickly?

ggetz (approved):

ggetz:
Thanks @keyboardspecialist!

Issue #12764, opened 2025-07-24
Support glTF EXT_mesh_primitive_restart in CesiumJS


markschlosseratbentley:
See https://github.com/iTwin/itwinjs-core/pull/8312 on iTwin.js. This was recently merged in and supports the proposed [EXT_mesh_primitive_restart ](https://github.com/KhronosGroup/glTF/pull/2478) glTF extension.

We could consider adding similar support in CesiumJS. Reference the above two links as this work is considered.

javagl:
Assuming that this can be closed as of https://github.com/CesiumGS/cesium/pull/12792

PR #12792, opened 2025-08-01
Support EXT_mesh_primitive_restart glTF extension

eringram:
# Description

This PR adds support for the EXT_mesh_primitive_restart glTF extension. It is based on the implementation in iTwin.js-- see `getMeshPrimitives` in [GltfReader.ts](https://github.com/iTwin/itwinjs-core/blob/17c6f833ed1753f4e969e0edaab8aaf315ab23b8/core/frontend/src/tile/GltfReader.ts#L436). 

"Primitive restart" refers to starting a new primitive of the same type as the current primitive when a specific index value is encountered (the maximum value for that index buffer type). It is useful for batching multiple primitives of the same type in a single draw call. See the README in [this PR](https://github.com/KhronosGroup/glTF/pull/2478) for more detailed explanation and a simple example.

## Issue number and link

Addresses https://github.com/CesiumGS/cesium/issues/12764

## Testing plan

- Added a unit test to GltfLoaderSpec.js that uses the MeshPrimitiveRestart.gltf test model and confirms that the new `getMeshPrimitives` function reduces the length of the primitives array from 8 to 4 (effectively combining the 4 groups of 2 primitives each).
    - The test model includes primitive restart groups for all 4 supported modes.
- Added unit tests that confirm the extension is **not** used if the glTF model violates the spec (i.e. uses a primitive in multiple groups, a primitive has an unsupported mode, etc.)
- You can also manually test by opening MeshPrimitiveRestart.gltf in Sandcastle and stepping through the code to confirm the extension is used. This is what the model looks like:

image

(the 4 groups are two green triangles, two blue triangles, two green line strips, and two blue line loops)

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/IonResource.js
+++ b/packages/engine/Source/Core/IonResource.js
@@ -199,15 +199,8 @@ IonResource.prototype._makeRequest = function (options) {
     return Resource.prototype._makeRequest.call(this, options);
   }
 
-  if (!defined(options.headers)) {
-    options.headers = {};
-  }
+  addClientHeaders(options);
   options.headers.Authorization = `Bearer ${this._ionEndpoint.accessToken}`;
-  options.headers["X-Cesium-Client"] = "CesiumJS";
-  /* global CESIUM_VERSION */
-  if (typeof CESIUM_VERSION !== "undefined") {
-    options.headers["X-Cesium-Client-Version"] = CESIUM_VERSION;
-  }
 
   return Resource.prototype._makeRequest.call(this, options);
 };
@@ -233,8 +226,21 @@ IonResource._createEndpointResource = function (assetId, options) {
     resourceOptions.queryParameters = { access_token: accessToken };
   }
 
+  addClientHeaders(resourceOptions);
+
   return server.getDerivedResource(resourceOptions);
 };
+
+function addClientHeaders(options) {
+  if (!defined(options.headers)) {
+    options.headers = {};
+  }
+  options.headers["X-Cesium-Client"] = "CesiumJS";
+  /* global CESIUM_VERSION */
+  if (typeof CESIUM_VERSION !== "undefined") {
+    options.headers["X-Cesium-Client-Version"] = CESIUM_VERSION;
+  }
+}
 
 function retryCallback(that, error) {
   const ionRoot = that._ionRoot ?? that;
--- a/packages/engine/Source/Scene/BingMapsImageryProvider.js
+++ b/packages/engine/Source/Scene/BingMapsImageryProvider.js
@@ -160,7 +160,7 @@ async function requestMetadata(
   const cacheKey = metadataResource.url;
   let promise = BingMapsImageryProvider._metadataCache[cacheKey];
   if (!defined(promise)) {
-    promise = metadataResource.fetchJsonp("jsonp");
+    promise = metadataResource.fetchJson();
     BingMapsImageryProvider._metadataCache[cacheKey] = promise;
   }
 
--- a/packages/engine/Specs/Scene/BingMapsImageryProviderSpec.js
+++ b/packages/engine/Specs/Scene/BingMapsImageryProviderSpec.js
@@ -35,10 +35,6 @@ describe("Scene/BingMapsImageryProvider", function () {
   });
 
   afterEach(function () {
-    Resource._Implementations.loadAndExecuteScript =
-      Resource._DefaultImplementations.loadAndExecuteScript;
-    Resource._Implementations.loadAndExecuteScript =
-      Resource._DefaultImplementations.loadAndExecuteScript;
     Resource._Implementations.loadWithXhr =
       Resource._DefaultImplementations.loadWithXhr;
     Resource._Implementations.createImage =
@@ -82,46 +78,31 @@ describe("Scene/BingMapsImageryProvider", function () {
     expect(BingMapsImageryProvider).toConformToInterface(ImageryProvider);
   });
 
-  function installFakeMetadataRequest(url, mapStyle, mapLayer, culture) {
-    const baseUri = new Uri(appendForwardSlash(url));
-    const expectedUri = new Uri(
-      `REST/v1/Imagery/Metadata/${mapStyle}`,
-    ).absoluteTo(baseUri);
+  function installFakeRequest(metadataOptions, imageOptions) {
+    imageOptions = imageOptions ?? {};
 
-    Resource._Implementations.loadAndExecuteScript = function (
-      url,
-      functionName,
-    ) {
-      const uri = new Uri(url);
-
-      const query = queryToObject(uri.query());
-      expect(query.jsonp).toBeDefined();
-      expect(query.incl).toEqual("ImageryProviders");
-      expect(query.key).toBeDefined();
-
-      if (defined(mapLayer)) {
-        expect(query.mapLayer).toEqual(mapLayer);
+    let metadataExpectedUri;
+    let metadataResponse;
+    if (defined(metadataOptions)) {
+      const { url, mapStyle, response } = metadataOptions;
+      const baseUri = new Uri(appendForwardSlash(url));
+      metadataExpectedUri = new Uri(
+        `REST/v1/Imagery/Metadata/${mapStyle}`,
+      ).absoluteTo(baseUri);
+      metadataResponse =
+        response ?? createFakeBingMapsMetadataResponse(mapStyle);
+      if (typeof metadataResponse === "object") {
+        metadataResponse = JSON.stringify(metadataResponse);
       }
+    }
 
-      if (defined(culture)) {
-        expect(query.culture).toEqual(culture);
-      }
-
-      uri.query("");
-      expect(uri.toString()).toStartWith(expectedUri.toString());
-
-      setTimeout(function () {
-        window[functionName](createFakeBingMapsMetadataResponse(mapStyle));
-      }, 1);
-    };
-  }
-
-  function installFakeImageRequest(expectedUrl, expectedParams, proxy) {
     Resource._Implementations.createImage = function (
       request,
       crossOrigin,
       deferred,
     ) {
+      const { expectedUrl, expectedParams } = imageOptions;
+
       const url = request.url;
       if (/^blob:/.test(url) || supportsImageBitmapOptions) {
         // If ImageBitmap is supported, we expect a loadWithXhr request to fetch it as a blob.
@@ -135,10 +116,7 @@ describe("Scene/BingMapsImageryProvider", function () {
         );
       } else {
         if (defined(expectedUrl)) {
-          let uri = new Uri(url);
-          if (proxy) {
-            uri = new Uri(decodeURIComponent(uri.query()));
-          }
+          const uri = new Uri(url);
 
           const query = queryToObject(uri.query());
           uri.query("");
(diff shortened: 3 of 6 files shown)


ggetz:
Thanks @eringram! I can confirm we have a CLA on file covering you.

markschlosseratbentley:
I read through the code. This looks great; thanks @eringram.

ggetz:
@eringram Is this PR ready for full or partial review? If so, could you please update the PR description with the current status? Thanks!

eringram:
> @eringram Is this PR ready for full or partial review? If so, could you please update the PR description with the current status? Thanks!

Sorry just updated the description! The code in GltfLoader.js is ready for review and still working on more thorough unit tests.

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2394,6 +2395,75 @@ function loadInstanceFeaturesLegacy(
...
+  if (!defined(primitiveRestartExtension)) {
+    return meshPrimitives;
+  }
+
+  // Note: per the spec, any violation of the extension's specification should cause us to fall back to mesh.primitives, if detecting the violation is feasible.
+
+  // Start with a copy of mesh.primitives. For each group, replace the first primitive in the group with a primitive representing the entire group,
+  // and set the rest of the primitives in the group to `undefined`.
+  // This allows us to identify which remaining primitives do not use primitive restart, and any errors involving a primitive appearing in more than one group.
+  const primitives = [...meshPrimitives];
```
javagl: We had some issues with `push.apply` because it puts all elements on the stack (yeah, sounds crazy, but that's the way it is), and the same applies to the `[...spread]` operator (see https://github.com/CesiumGS/cesium/issues/12053 ). I don't know if we want to be prepared for meshes with 150000 primitives...

markschlosseratbentley: @javagl We ran into a similar issue in iTwin.js in the past, so thanks for pointing this out.  Check out this PR for that similar situation: https://github.com/iTwin/itwinjs-core/pull/7605

We found the spread operator was faster than some other methods, but could cause a stack overflow of sorts if done on too many entries - so that PR picks which approach to use based on number of entries.

Maybe could consider some similar code here.

javagl: There's now an [`addAllToArray`](https://github.com/CesiumGS/cesium/blob/4e930a7990f00dde993812276559b2bf55d34afd/packages/engine/Source/Core/addAllToArray.js#L26) function to generically handle this, without having to do these manual `if ... > 50000` checks
```javascript
const primitives = [];
addAllToArray(primitives, meshPrimitives);
```
It would be nice(r) if built-in idiomatic language constructs like the spread operator wouldn't make this necessary, but ... here we are. If future Chrome/FireFox versions make the spread operator "better" than a loop, we'll update that single function and be done 🤞

markschlosseratbentley: Thanks for the `addAllToArray` suggestion, @javagl.  @eringram, once we add usage of that function to this PR, we should consider doing something similar to this parallel extension code in iTwin.js if necessary.

eringram: Thanks, just updated this to use `addAllToArray`.

eringram:
I updated the MeshPrimitiveRestart.gltf test data to include primitive groups for all the valid modes for the extension (line loop, line strip, triangle strip, triangle fan). Also added more unit tests to ensure `getMeshPrimitives` falls back to the default `mesh.primitives` if EXT_mesh_primitive_restart violates the spec. Two of those tests are failing, where the extension is used even if incorrect, so still fixing that.

~~Also, @javagl @ggetz any idea why [this test](https://github.com/CesiumGS/cesium/pull/12792/files#diff-8f11fb72d71b01b789e5660c80ad63658fa8768871b77786e95b636aa918c6c4R4279) that calls `toBeRejectedWithError` is failing to match the error message regex?~~

~~This is the output when it runs and says it fails, but the errors look identical to me:~~

Nevermind, just noticed the cases are different in the error messages...

```
3) does not load with EXT_mesh_primitive_restart if a group's indices accessor is invalid
     Scene/GltfLoader
     Expected a promise to be rejected with RuntimeError: /Failed to load glTF
Failed to load index buffer
indexDataType is required and must be a valid IndexDatatype constant./ but it was rejected with RuntimeError: Failed to load glTF
Failed to load index buffer
indexDatatype is required and must be a valid IndexDatatype constant.
Original stack:
Original stack:
Error
    at new DeveloperError (packages/engine/Source/Core/DeveloperError.js:39:11 <- Build/CesiumUnminified/Cesium.js:8893:13)
    at IndexDatatype.getSizeInBytes (packages/engine/Source/Core/IndexDatatype.js:62:9 <- Build/CesiumUnminified/Cesium.js:26912:11)
    at createIndicesTypedArray (packages/engine/Source/Scene/GltfIndexBufferLoader.js:248:35 <- Build/CesiumUnminified/Cesium.js:70949:45)
    at loadFromBufferView2 (packages/engine/Source/Scene/GltfIndexBufferLoader.js:227:37 <- Build/CesiumUnminified/Cesium.js:70930:39)
    at async Promise.all (index 7)
Handler stack:
Error
    at new RuntimeError (packages/engine/Source/Core/RuntimeError.js:38:11 <- Build/CesiumUnminified/Cesium.js:13822:13)
    at ResourceLoader.getError (Build/Ce ... ....
    at <Jasmine>
    at UserContext.<anonymous> (packages/engine/Specs/Scene/GltfLoaderSpec.js:4263:9 <- Build/Specs/SpecList.js:209938:9)
    at <Jasmine>
```

eringram:
This is ready for full review.

Anyone have an idea why the `prettier` format check is failing? It does not cause a warning for CHANGES.md when I run locally.

javagl:
(Quick, while in another call:) 

Regarding the formatting: Indeed, something is strange here: https://github.com/CesiumGS/cesium/pull/12801#issuecomment-3161155566

The test that is failing _now_ is related to some error message not matching the expected one, at https://github.com/CesiumGS/cesium/pull/12792/files#diff-8f11fb72d71b01b789e5660c80ad63658fa8768871b77786e95b636aa918c6c4R4277 . 

A bit more generally, I think that it's a bit dubious to check for the _exact_ error message, even though this is done in many places. I think that this "overspecifies" the test. (I hope there are no plans on internationalization, FWIW...). One could try to "justify" that, but only with the severe lack of differentiation between _types_ of errors: There should be far more than just `RuntimeError` and `DeveloperError`, and the latter is used in many places where the _developer_ actually has no influence on anything, and it should rather be a `DataError` or so. 

However, the test here _could_ just be changed to expect a type error, but ... of course, such an error should preferably not be thrown to the user either...

eringram:
> Regarding the formatting: Indeed, something is strange here: [#12801 (comment)](https://github.com/CesiumGS/cesium/pull/12801#issuecomment-3161155566)

Ok good to know it wasn't just me, I ended up just reverting the whitespace. Not sure how to fix the discrepancy it but it seems related to the `blanks-around-fences` markdownlint rule.

> A bit more generally, I think that it's a bit dubious to check for the _exact_ error message, even though this is done in many places. I think that this "overspecifies" the test. (I hope there are no plans on internationalization, FWIW...). One could try to "justify" that, but only with the severe lack of differentiation between _types_ of errors: There should be far more than just `RuntimeError` and `DeveloperError`, and the latter is used in many places where the _developer_ actually has no influence on anything, and it should rather be a `DataError` or so.
> 
> However, the test here _could_ just be changed to expect a type error, but ... of course, such an error should preferably not be thrown to the user either...

These considerations make sense, you're right that the lack of error types kind of leads to comparing the error messages like this. I tried to use a generic `Error` type, but now see that the test is timing out. Any idea about that? I see other tests using `await expectAsync(...).toBeRejectedWithError()` that don't appear to increase their timeout past 5000ms.

eringram:
Tbh, after further thinking, the test that is failing is probably not necessary in the first place. Its purpose is to test that an error is thrown when a mesh primitive restart group's `indices` property is invalid (type isn't scalar, component type isn't unsigned integer). However, this is not specific to EXT_mesh_primitive_restart, but applies to the glTF being valid in general. The `primitiveGroup.indices` eventually becomes a `mesh.primitive.indices` [here](https://github.com/CesiumGS/cesium/pull/12792/files#diff-c33e97311a160159cbd7a185f112801a5330b0434e80e8e83324d249eb7ccfefR2433), so as long as `mesh.primitive.indices` being valid is tested, this not necessary.

CHANGES.md
```diff
@@ -31,6 +31,7 @@
 - Expand the CustomShader Sample to support real-time modification of CustomShader. [#12702](https://github.com/CesiumGS/cesium/pull/12702)
 - Add wrapR property to Sampler and Texture3D, to support the newly added third dimension wrap.[#12701](https://github.com/CesiumGS/cesium/pull/12701)
 - Added the ability to load a specific changeset for iTwin Mesh Exports using `ITwinData.createTilesetFromIModelId` [#12778](https://github.com/CesiumGS/cesium/issues/12778)
+- Added support for the [EXT_mesh_primitive_restart](https://github.com/KhronosGroup/glTF/pull/2478) glTF extension. [#12764](https://github.com/CesiumGS/cesium/issues/12764)
```
ggetz: Once merged, this change will go into the _next_ release, which will be on Sept 2 (typically it's on the 1st of each month, but Sept 1 is Labor Day in the US).

Please move this item to the H2 section above under a new "additions" subsection.

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2394,6 +2395,76 @@ function loadInstanceFeaturesLegacy(
...
 }
 
+/**
+ * Get an array of primitives for a given mesh. If the EXT_mesh_primitive_restart extension is present, use it to combine groups of primitives.
+ * If the extension is not present or its spec is violated, return the original mesh.primitives.
+ * @param {object} mesh A mesh from the glTF meshes array
+ * @returns {object[]} An array of mesh primitives
+ * @private
+ */
+function getMeshPrimitives(mesh) {
```
ggetz: This file is already pretty brutal in terms of length and separation of concerns. What do you think of moving this function to it's own file?

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2394,6 +2395,76 @@ function loadInstanceFeaturesLegacy(
...
+  // Start with a copy of mesh.primitives. For each group, replace the first primitive in the group with a primitive representing the entire group,
+  // and set the rest of the primitives in the group to `undefined`.
+  // This allows us to identify which remaining primitives do not use primitive restart, and any errors involving a primitive appearing in more than one group.
+  const primitives = [];
+  addAllToArray(primitives, meshPrimitives);
+  for (const group of primitiveRestartExtension.primitiveGroups) {
+    // Spec: the group must not be empty and all indices must be valid array indices into mesh.primitives.
+    const firstPrimitiveIndex = group.primitives[0];
+    if (
+      undefined === firstPrimitiveIndex ||
```
ggetz: Throughout CesiumJS, we use the `defined` function rather than checking against `undefined` directly, as it'll also handle null values.

```suggestion
      !defined(firstPrimitiveIndex) ||
```

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2394,6 +2395,76 @@ function loadInstanceFeaturesLegacy(
...
+        return meshPrimitives;
+    }
+
+    for (const primitiveIndex of group.primitives) {
+      const thisPrimitive = primitives[primitiveIndex];
+
+      // Spec: all primitives must use indexed geometry and a given primitive may appear in at most one group.
+      // Spec: all primitives must have same topology.
+      if (
+        undefined === thisPrimitive?.indices ||
```
ggetz:
```suggestion
        !defined(thisPrimitive?.indices) ||
```

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2394,6 +2395,76 @@ function loadInstanceFeaturesLegacy(
...
+        return meshPrimitives;
+      }
+
+      primitives[primitiveIndex] = undefined;
+    }
+
+    primitives[firstPrimitiveIndex] = primitive;
+  }
+
+  return primitives.filter((x) => x !== undefined);
```
ggetz:
```suggestion
  return primitives.filter(defined);
```

CHANGES.md
eringram: Fixed, I somehow missed the latest release header

packages/engine/Source/Scene/GltfLoader.js
eringram: I agree, moved to a new file.

eringram:
@pmconne Not required, but checking to see if you intended to review this

pmconne (approved):

ggetz (approved):

ggetz:
Looks great! Thanks @eringram!

PR #12794, opened 2025-08-01
New sandcastle title and dirty state tracking

jjspace:
# Description

- Clean up the window title management. Now shows the name of the sandcastle or "New sandcastle" and "Cesium Sandcastle"
- Add "dirty" state tracking for the code. When dirty add a `*` to the page title
- Warn when navigating away from the page or trying to load a different gallery item if the code is "dirty"
    - "Saving" the code by clicking "Share" which currently updates the url in place is considered "clean" again because forward/back navigation will work to that URL. This behavior may change when the share functionality gets improved.

## Issue number and link

Part of #12566 

## Testing plan

- Run `npm run dev` in the sandcastle package or `npm run build-sandcastle` and `npm start` from the project root
- load new sandcastle and check the title
- Swap to a gallery demo and check the title
- Make a change to the code and make sure the title gets it's `*`
- Try navigating away when in the "dirty" state and make sure it always shows a warning
     - Reload the page
     - Click the top logo
     - Close the tab/window
     - Click a different gallery demo
     - Click the code button for a different demo

# Author checklist


packages/sandcastle/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Sandcastle Reborn</title>
+    <title>Cesium Sandcastle</title>
```
ggetz:
```suggestion
    <title>CesiumJS Sandcastle</title>
```

I think, technically speaking, this should be the branding?

packages/sandcastle/src/App.tsx
```diff
@@ -241,11 +247,46 @@ function App() {
...
+
+  const confirmLeave = useCallback(() => {
+    if (!codeState.dirty) {
+      return true;
+    }
+
+    /* eslint-disable-next-line no-alert */
+    return window.confirm(
+      "You have unsaved changes. Are you sure you want to navigate away from this demo?",
+    );
```
ggetz, line 287: This is totally up-to-par with what we had before, but what do you think of propagating the current sandcastle to local storage or similar?

The goal would to be to persist state, like what happens when you reload a GitHub tab with an in-progress comment. Or maybe a slightly different workflow, like what happens if you leave and come back to https://geojson.io/.

packages/sandcastle/src/App.tsx
```diff
@@ -241,11 +247,46 @@ function App() {
...
 
+  useEffect(() => {
+    const host = window.location.host;
+    let envString = "";
+    if (host.includes("localhost") && host !== "localhost:8080") {
+      // this helps differentiate tabs for local sandcastle development or other testing
+      envString = `${host.replace("localhost:", "")} `;
+    }
+
+    const baseTitle = "Cesium Sandcastle";
```
ggetz:
```suggestion
    const baseTitle = "CesiumJS Sandcastle";
```

or possibly even just the following due to all the other info?

```suggestion
    const baseTitle = "Sandcastle";
```

packages/sandcastle/src/App.tsx
```diff
@@ -241,11 +247,46 @@ function App() {
...
+    const dirtyIndicator = codeState.dirty ? "*" : "";
+    if (title !== "") {
+      document.title = `${envString}${title}${dirtyIndicator} | ${baseTitle}`;
+    } else {
+      document.title = `${envString}${baseTitle}`;
+    }
```
ggetz: 1. Why shouldn't we show the dirty indicator on named sandcastles?
2. Assuming we can, couldn't we simplify this to thew following?

```suggestion
      document.title = `${envString}${title}${dirtyIndicator} | ${baseTitle}`;
```

ggetz (changes requested):
Thanks @jjspace! A few questions and branding-related comments.

packages/sandcastle/index.html
jjspace: Possibly? I was just matching what the current Sandcastle uses

packages/sandcastle/src/App.tsx
jjspace: I want this to match whatever title we pick for the base page so the specific sandcastle title is just prepended. I'm open to dropping down to just "Sandcastle" but I wasn't sure about dropping the branding

packages/sandcastle/src/App.tsx
jjspace: This check makes sure that we only add the title and the ` | ` to the page title when there's actually a title to prepend in front of the bar. The page starts with an empty title `""` in state when it initially loads because we don't know what it will be yet.
If a specific gallery demo is not loaded then we replace it with "New Sandcastle" thus providing a title and it _will_ show the dirty indicator.

packages/sandcastle/src/App.tsx
jjspace: > what do you think of propagating the current sandcastle to local storage or similar?

This (or similar) has been requested often and I _do_ really want to find some solution to try and prevent lost work. However I think it should be a separate effort, this PR just tries to match the functionality of the existing sandcastle. This is already on the running task list but further down for lower priority.happy to re-evaluate if you think it's a "must have" before release.

My current thought is _not_ to make it truly persist every page load but offer a "saved state" that we can suggest the user reload like [jsfiddle does](https://jsfiddle.net/)

image

jjspace:
@ggetz no updates yet but I responded to all your comments if you wanna take another look.

packages/sandcastle/index.html
ggetz: Let's go with CesiumJS. No need to stick to an arbitrary decision made 10 years ago. 🙂

packages/sandcastle/src/App.tsx
ggetz: No problem with adjusting this as a subsequent PR. 

We should probably link or move this conversation to the issue, which I'll let you organize as you see fit. 

But for the sake of keeping it going:

> My current thought is not to make it truly persist every page load but offer a "saved state" that we can suggest the user reload like [jsfiddle does](https://jsfiddle.net/)

Cool, that sounds appropriate to me. That seems very similar to what https://geojson.io/ does, as I linked above and shown below. But let me know if I misunderstood.

image

packages/sandcastle/src/App.tsx
ggetz: I'm looking at what we do on our other products and marketing website. Unless you see an issue, let's follow suit with what https://ion.cesium.com/ does—`<Page Title> | Cesium ion`, such as `Stories | Cesium ion` or `My Assets | Cesium ion`.

So for Sandcastle, I'm thinking `Sandcastle | CesiumJS` by default, and `<Gallery Item Title> | Sandcastle | CesiumJS`, such as `3D Models | Sandcastle | CesiumJS` or `Geometry and Appearances | Sandcastle | CesiumJS`, where appropriate.

packages/sandcastle/index.html
ggetz: See my comment below for thoughts on specific titles.

packages/sandcastle/src/App.tsx
ggetz: So to clarify, it does already use a dirty indicator symbol for loaded gallery items, i.e., `3D Models* | ...`? Assuming yes, good 👍 

> The page starts with an empty title "" in state when it initially loads because we don't know what it will be yet.
> If a specific gallery demo is not loaded then we replace it with "New Sandcastle" thus providing a title and it will show the dirty indicator.

Hm, I'm not sure if "New Sandcastle" is adding much more context, especially for the limited spaced. What do you think of, assuming we are using the naming schema above, using the dirty indicator symbol on the first title segment? For example, `3D Models* | Sandcastle | CesiumJS` when a gallery item was previously loaded, and `Sandcastle* | CesiumJS` when one was not?

But if you disagree with that, consider this instead: At the point where the user begins editing, and we would need to set a dirty flag, would it make sense to use a title of "New Sandcastle" at that point rather than an empty title?

ggetz:
@jjspace Responded above. Hopefully that's enough to move forward. But, if not, let me know and we can chat further.

jjspace:
Thanks @ggetz, I like your suggestions. The title has been simplified to `Sandcastle* | CesiumJS` or `[title]* | Sandcastle | CesiumJS`. The dirty `*` will only show up if it's actually been modified. The title will only show when viewing the actual gallery demos for now to keep the title cleaner. 

This should be ready for another look, please merge if happy

Issue #7715, opened 2019-04-04
Consider updating maximumScreenSpaceError to 1.5 by default


mramato:
Spun out of #3279

Now that our terrain and imagery system has been upgraded (#7061), would it make sense to lower the `maximumScreenSpaceError` default to something like 1.5?  This would increase tile downloads slightly but would also increase visual quality and could better align with what people expect from 2D top-down views.

We should do a head-to-head comparison of various settings and see what the request/performance metrics look like.  A slightly tweaked test I used in #7061 should work well for this too: https://github.com/AnalyticalGraphicsInc/cesium/pull/7061#issuecomment-461844082

mramato:
@kring suggests 1.33 as a possible optimal value in https://github.com/AnalyticalGraphicsInc/cesium/issues/3279#issuecomment-271481280

Issue #4043, opened 2016-06-24
maximumScreenSpaceError a bad metric for 3D Tiles


mramato:
I spoke with both @lilleyse and @bagnell about this offline and Sean asked me to submit an issue.

`maximumScreenSpaceError` does not work very well as a metric for tile selection because it doesn't take into account the size of the Cesium canvas being drawn too, i.e. the same camera view will create wildly different images depending on the size of the window.  Instead of absolute pixel size, perhaps this can be better represented as a percentage of pixels of the overall canvas?

I took a quick stab at illustrating this with the below Sandcastle example.  If you resize the width of the window, you'll notice the view stays much more consistent than the current behavior.

```javascript
var viewer = new Cesium.Viewer('cesiumContainer', {
    scene3DOnly : true,
    shadows : true
});

var scene = viewer.scene;
scene.fog.enabled = false;
scene.debugShowFramesPerSecond = true;

var tileset = scene.primitives.add(new Cesium.Cesium3DTileset({
    url : 'https://cesiumjs.org/NewYork/3DTilesGml/tileset.json'
}));

Cesium.when(tileset.readyPromise)
  .then(function(tileset) {
    var boundingSphere = tileset.boundingSphere;
    viewer.camera.viewBoundingSphere(boundingSphere, new Cesium.HeadingPitchRange(0, -2.0, 0));
    viewer.camera.lookAtTransform(Cesium.Matrix4.IDENTITY);
  });


var lastWidth;
scene.preRender.addEventListener(function(){
    var maximumScreenSpacePercentage = 0.01;
    
    var width = scene.canvas.clientWidth;
    if(lastWidth !== width){
        var maximumScreenSpaceError = maximumScreenSpacePercentage * width;
        lastWidth = width;
        console.log(maximumScreenSpaceError);
        console.log(width);    
        tileset.maximumScreenSpaceError = maximumScreenSpaceError;
    }
});
```

CC @pjcozzi

mramato:
For the record, I'm not sure my approach is the best choice either, just one possible alternative.

pjcozzi:
Added this to #3241.  Whatever we do here we would also want to do for terrain/imagery.

kring:
@mramato can you explain what problem you're trying to solve?  IMO a larger screen _should_ use a different level of detail than a small screen, given the same camera position.  That's a feature, not a bug.

mramato:
[...]

PR #12801, opened 2025-08-06
Allow passing tileset options to reality data

javagl:
# Description

Allow passing a `tilesetOptions` object to `ITwinData.createTilesetForRealityDataId`. The tileset options will be forwarded to the tileset that is created internally. One caveat is that the `maximumScreenSpaceError` for reality data tilesets was set to 4 by default. This should still be the default, with the option of overriding that default with the given options.

## Issue number and link

https://github.com/CesiumGS/cesium/issues/12709

## Testing plan

Basic specs have been added for passing on the options, including the special handling of the `maximumScreenSpaceError`. 

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/ITwinData.js
+++ b/packages/engine/Source/Scene/ITwinData.js
@@ -90,6 +90,9 @@ ITwinData.createTilesetFromIModelId = async function ({
  * If the <code>type</code> or <code>rootDocument</code> are not provided this function
  * will first request the full metadata for the specified reality data to fill these values.
  *
+ * The <code>maximumScreenSpaceError</code> of the resulting tileset will default to 4,
+ * unless it is explicitly overridden with the given tileset options.
+ *
  * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
  *
  * @param {Object} options
@@ -97,6 +100,8 @@ ITwinData.createTilesetFromIModelId = async function ({
  * @param {string} options.realityDataId The id of the reality data to load
  * @param {ITwinPlatform.RealityDataType} [options.type] The type of this reality data
  * @param {string} [options.rootDocument] The path of the root document for this reality data
+ * @param {Cesium3DTileset.ConstructorOptions} [options.tilesetOptions] Object containing
+ * options to pass to the internally created {@link Cesium3DTileset}.
  * @returns {Promise<Cesium3DTileset>}
  *
  * @throws {RuntimeError} if the type of reality data is not supported by this function
@@ -106,6 +111,7 @@ ITwinData.createTilesetForRealityDataId = async function ({
   realityDataId,
   type,
   rootDocument,
+  tilesetOptions,
 }) {
   //>>includeStart('debug', pragmas.debug);
   Check.typeOf.string("iTwinId", iTwinId);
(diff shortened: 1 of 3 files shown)


javagl:
That formatting commit was a strange one. The `prettier` run changed some random empty lines in `CHANGES.md` for me locally. I initially omitted these changes, but apparently, that's what's currently expected. The point is: I wonder how the current state of the `CHANGES.md` (with that "wrong" formatting) ended up in the repo 🤷‍♂️

packages/engine/Source/Scene/ITwinData.js
```diff
@@ -144,9 +150,15 @@ ITwinData.createTilesetForRealityDataId = async function ({
     rootDocument,
   );
 
-  return Cesium3DTileset.fromUrl(tilesetAccessUrl, {
+  // The maximum screen space error was defined to default to 4 for
+  // reality data tilesets. This will only be overridden if it was
+  // specified in the tilesetOptions explicitly.
```
jjspace: I think this comment can be more descriptive other than "it was like this before this change"

We set this value because we noticed the reality data tilesets were commonly not loading enough detail when zoomed out compared to what is expected. A comment that explains _that_ context would be useful here.
The second half is also a bit redundant. I feel the spread `...` operator and behavior should be enough to explain the behavior here (but that's bikeshedding)

@ggetz I believe there was some discussion around fixing this on the data side within the itwin platform. do you know if that ever got addressed?

javagl: This PR is now assigned to you - was that intentional? 
If not, I can update the comment, and try to be more clear. 

Regardless of the spread operator (and even though I actually tried out whether the difference of
`x = { ...y, value: 123 };` and
`x = { value: 123, ...y };` 
is the one that one could expect (I've learned to not **ever** expect **anything** in JavaScript to be "sane" in **any** way 😆 ), I think it's important to make clear that somewhere, someone, decided that 4 should be the default here. If there was any _reasoning_ behind that (beyond "We tried it, and it ~'worked better'"), then I could add this as well.

Maybe all these values are "wrong", though. See https://github.com/CesiumGS/cesium/issues/7715 and, more generally https://github.com/CesiumGS/cesium/issues/4043 ...

jjspace: > This PR is now assigned to you - was that intentional?

I just thought it might be easier to track that I would be doing the review, maybe that's the wrong process. Please continue making changes

> Regardless of the spread operator...

Objects defined with duplicate keys, especially common when using the spread operator, take the value of the "later" one, or the one lower in the code. So `{ ...options, value: 'a' }` defaults to the spread values and uses the overrides for `value`. `{ value: 'a', ...options }` would default to the `value` but use the value from the `options` if it's defined. 

> I think it's important to make clear that somewhere, someone, decided that 4 should be the default here.

Regardless I do agree we should explain _why_ which is due to the source data tending to be lower "resolution" than many other tilesets at the default sse

javagl: I would have _expected_ the values to be assigned in the order in which they are listed _or_ fetched from `...spread`ed objects (including overriding of the previous ones). I still tried it out. (Yes, it's specified in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#overriding_properties , but I **still** tried it out 😆 )

I now extended the comment a bit, saying that the _usual_ default (16) was just not enough. Whether changing the default value is preferable compared to just telling iTwin to do a blanket `geometricError *= 4`? I don't know.

jjspace (approved):
LGTM, thanks @javagl!

packages/engine/Source/Scene/ITwinData.js
jjspace: > Whether changing the default value is preferable compared to just telling iTwin to do a blanket geometricError *= 4? I don't know.

This is probably part of a larger discussion but it shouldn't hold up this PR. Happy to merge it and change later if the data changes

CC @ggetz

Issue #11166, opened 2023-03-17
Crash during rendering of point cloud with batch table


runaas:
Cesium crashes in rendering of a point cloud file with batch table. If I use a .pnts file generated from the same data without point attributes (and no batch_table) it works without problems. The tileset has been testet with the 3d-tiles-validator without problems, and there are no issues during loading.

Cesium 1.96 didn't crash, but I was unable to extract point attributes, so something was not fully corrct. All versions from v 1.98 crashes (I've not tested 1.97...)

In the Unminified build of v 1.103 the crash happens here because of an undefined feature table:

BatchTexturePipelineStage.process@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:88762:26
ModelSceneGraph.prototype.buildDrawCommands@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:93454:32
buildDrawCommands@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:96083:23
Model.prototype.update@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:95960:20
Model3DTileContent.prototype.update@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:96812:9
Cesium3DTile.prototype.process@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:114825:17
processTiles@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:117608:14
Cesium3DTileset.prototype.prePassesUpdate@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:117498:15
PrimitiveCollection.prototype.prePassesUpdate@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:134434:17
prePassesUpdate@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:209512:14
tryAndCatchError@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:209585:22
Scene4.prototype.render@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:209624:19
CesiumWidget.prototype.render@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:210969:17
render2@http://norgei3d.no/Cesium/Current/Build/CesiumUnminified/index.js:210412:18
[...]

Issue #12805, opened 2025-08-07
Read batch- and feature table only once


javagl:
There are currently roughly 6 places (for B3DM, I3DM, PNTS, CMPT, GEOM, and VCTR) where the feature- and batch table are parsed. The feature and batch table are a generic concept that applies to all file format. This parsing code should exist only **once**. 

(Subtle details, like the different header length for I3DM, have to be taken into account, but that should only be an `if (i3dm) headerLength += 4;`)

An outtake, because it's _nearly_ funny again:

<img width="401" height="306" alt="Image" src="https://github.com/user-attachments/assets/f65db3ca-f53f-4a78-ad2e-d84c667d3413" />

javagl:
And whenever there's something _nearly_ funny, someone is doing the limbo, down to the point where it's no longer funny, but outright ridiculous:

![Image](https://github.com/user-attachments/assets/8434d0a5-5f89-4b38-982a-a881ff0a3587)

I found this while investigating https://github.com/CesiumGS/cesium/issues/12872 , starting with trying to figure out where the error message is coming from. Now I could just add some `if (!binaryBody) return;` at the beginning of one of these functions, to "fix the issue", but maybe, at some point we should consider stopping with the shallow patches and trying to clean things up in a way that _prevents_ issues like this.

PR #12806, opened 2025-08-07
Handle PNTS files where the batch table contains only JSON

javagl:
# Description

When a PNTS file contained a batch table that did not involve a binary part (but that stored all batch information directly in the JSON), then CesiumJS did crash with a 

> TypeError: Cannot read properties of undefined (reading 'featuresLength')

This was because it did not insert any batch table information at all into the 'parsed content' when there was no batch table binary. 

This PR changes this, to insert the batch table JSON- and binary information into the 'parsed content' if _either_ of them is defined. The binary part will be an empty `Uint8Array` in this case - from a quick check, this _seems_ to be anticipated. More complex combinations (like the case of ... some draco compression of the batch table, which seems to be a thing...) have not been checked explicitly. 

This is a fix that applies "the minimal number of modifications that are necessary to resolve the issue". A refactoring and cleanup of these code paths should happen as part of https://github.com/CesiumGS/cesium/issues/12805 

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/11166

## Testing plan

It can be tested by trying to load the tileset that was given in the issue (and for me, this worked, just for the case that I should mention that 🙂 ). I also added a new spec with spec data, based on the existing `PointCloudBatched` data. The new data, `PointCloudBatchedJsonOnly` was created from that, by omitting the binary part of the batch table. The resulting PNTS has a `batchTableBinaryByteLength=0`, which previously triggered the issue. Here is a comparison of the `PointCloudBatched` (left) and `PointCloudBatchedJsonOnly` (right): 

Cesium issue 11166 PNTS batch table JSON only

# Author checklist


javagl:
>  More complex combinations (like the case of ... some draco compression of the batch table, which seems to be a thing...) have not been checked explicitly.

Aaaand here we go... 

>  Scene/TimeDynamicPointCloud
>    ✗ renders frames using Draco compression

javagl:
What even is a "batch table"? Is it a string, or a JSON object? Can it be `undefined`? Can it be `""` the empty string? Is an empty JSON object `{}` a valid batch table? Maybe. Maybe not. An attempt to handle the different answers that people could give to these questions is part of the last commit: When the batch table is empty, then it remains `undefined`  in the `parseContent` - skipping the question about what `parsedContent` is.

Locally, the test

> Scene/Model/Model3DTileContent > pnts > Supports back face culling when there are per-point normals

is failing for me, saying "Expected 4 to be greater than 4.". Fortunately, this also happens on the current `main` state, so I don't see that as part of this PR.

ggetz:
@jjhembd Would you be able to review this PR?

jjhembd (approved):
Thanks @javagl! This looks good to me:

- The dataset in the linked issue now loads without error
- All existing PointCloud sandcastles still work as expected

PR #12819, opened 2025-08-14
Add Sandcastle settings

jjspace:
# Description

This PR:

- Implements Sandcastle settings using local storage
    - Theme (which already existed)
    - Editor font + ligature option
        - I included 3 popular code editor fonts for now, adding more is easier. Droid Sans is the "default" for monaco even though they don't bundle it
    - Default starting panel, gallery or editor
        - When loading a specific sandcastle from the URL, named or a share link, we always default to the editor
        - If _not_ loading a specific sandcastle we fall to the setting
        - The setting will default to Gallery for new users
        - This is meant as a convenience factor for more regular/power users
- Implements dialogs for sandcastle using stratakit variables for consistent style and uses this for the settings modal

I meant to also include the popovers for the sandcastle metadata and share interaction in this PR but opted to do them in a follow up to keep the PR more focused and easier to review.

## Issue number and link

part of #12566 

## Testing plan

- Run sandcastle using `npm run dev` inside the package or `npm run build-sandcastle` and `npm start` from the project root
- Play with the new settings modal
- Make sure settings are preserved after refresh
- Make sure the editor font matches
- Make sure changing the theme from the sidebar and within the modal both work and don't conflict
- Make sure the default starting panel works as expected

# Author checklist

### DIFF
--- a/packages/sandcastle/gallery/3d-models-articulations-dev/index.html
+++ b/packages/sandcastle/gallery/3d-models-articulations-dev/index.html
@@ -1,15 +1,5 @@
 <style>
   @import url(../templates/bucket.css);
-  #toolbar {
-    background: rgba(42, 42, 42, 0.8);
-    padding: 4px;
-    border-radius: 4px;
-  }
-  #toolbar input {
-    vertical-align: middle;
-    padding-top: 2px;
-    padding-bottom: 2px;
-  }
 </style>
 <div id="cesiumContainer" class="fullSize"></div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
--- a/packages/sandcastle/gallery/3d-models-coloring/index.html
+++ b/packages/sandcastle/gallery/3d-models-coloring/index.html
@@ -1,18 +1,5 @@
 <style>
   @import url(../templates/bucket.css);
-  #toolbar {
-    background: rgba(42, 42, 42, 0.8);
-    padding: 4px;
-    border-radius: 4px;
-  }
-  #toolbar input {
-    vertical-align: middle;
-    padding-top: 2px;
-    padding-bottom: 2px;
-  }
-  #toolbar .header {
-    font-weight: bold;
-  }
 </style>
 <div id="cesiumContainer" class="fullSize"></div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
@@ -20,7 +7,7 @@
   <table>
     <tbody>
       <tr>
-        <td class="header">Model Color</td>
+        <td><b>Model Color</b></td>
       </tr>
       <tr>
         <td>Mode</td>
@@ -67,7 +54,7 @@
         </td>
       </tr>
       <tr>
-        <td class="header">Model Silhouette</td>
+        <td><b>Model Silhouette</b></td>
       </tr>
       <tr>
         <td>Color</td>
--- a/packages/sandcastle/gallery/3d-models-dev/index.html
+++ b/packages/sandcastle/gallery/3d-models-dev/index.html
@@ -1,15 +1,5 @@
 <style>
   @import url(../templates/bucket.css);
-  #toolbar {
-    background: rgba(42, 42, 42, 0.8);
-    padding: 4px;
-    border-radius: 4px;
-  }
-  #toolbar input {
-    vertical-align: middle;
-    padding-top: 2px;
-    padding-bottom: 2px;
-  }
 </style>
 <div id="cesiumContainer" class="fullSize"></div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
--- a/packages/sandcastle/gallery/3d-models-node-explorer-dev/index.html
+++ b/packages/sandcastle/gallery/3d-models-node-explorer-dev/index.html
@@ -1,15 +1,5 @@
 <style>
   @import url(../templates/bucket.css);
-  #toolbar {
-    background: rgba(42, 42, 42, 0.8);
-    padding: 4px;
-    border-radius: 4px;
-  }
-  #toolbar input {
-    vertical-align: middle;
-    padding-top: 2px;
-    padding-bottom: 2px;
-  }
 </style>
 <div id="cesiumContainer" class="fullSize"></div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
--- a/packages/sandcastle/gallery/3d-tiles-adjust-height/index.html
+++ b/packages/sandcastle/gallery/3d-tiles-adjust-height/index.html
@@ -1,15 +1,5 @@
 <style>
   @import url(../templates/bucket.css);
-  #toolbar {
-    background: rgba(42, 42, 42, 0.8);
-    padding: 4px;
-    border-radius: 4px;
-  }
-  #toolbar input {
-    vertical-align: middle;
-    padding-top: 2px;
-    padding-bottom: 2px;
-  }
 </style>
 <div id="cesiumContainer" class="fullSize"></div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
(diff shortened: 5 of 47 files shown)


packages/sandcastle/src/InfoBadge.css
```diff
@@ -0,0 +1,8 @@
+.info-badge {
+  display: inline-block;
+
+  .🥝-label > & {
```
lukemckinstry, line 4: Is text replacement from trying the write `kiwi` referring to the library. This should be changed, right?

packages/sandcastle/src/SandcastleDialog.css
```diff
@@ -0,0 +1,38 @@
...
+  background-color: var(--stratakit-color-bg-page-base);
+  padding: var(--stratakit-space-x4);
+  box-shadow: var(--stratakit-shadow-dialog-base);
+
+  .sc-dialog-footer {
+    align-self: flex-end;
+    margin-top: var(--stratakit-space-x4);
+  }
+
+  @media (min-width: 500px) {
```
lukemckinstry, line 35: The modal looks good on mobile and responsive to page resizes. The main page (viewer and gallery/editor) do not seem optimized for smaller screens yet, is that coming in a future PR?

lukemckinstry (commented):
Looks good :+1: Just one actionable comment but otherwise good to go.

packages/sandcastle/src/SandcastleDialog.css
jjspace: Longer term we've discussed making sandcastle "look nice" on mobile even if it's not very functional. However this is not a focus for the first version.
These styles were based on the example from the [AriaKit Dialog](https://ariakit.org/components/dialog) sample code which already included this so I just kept it. It's especially good because this is a very generic component so when we do take a pass for mobile styles this is already ready to go.

packages/sandcastle/src/InfoBadge.css
jjspace: This is already correct since we're using the stratakit components. This overrides the position anytime the badge is used inside a field label.

packages/sandcastle/src/SandcastleDialog.css
lukemckinstry: Sounds good re: the modal

I know it is outside the scope of this PR, but for general mobile support for the new sandcastle, I am in favor of some basic support to view at least code right off the bat, rather than following up later, if that is reasonable. I tried viewing in mobile using developer tools and it was not usable.

packages/sandcastle/src/InfoBadge.css
lukemckinstry: ok thanks for explaining the backstory offline that the strata codebase uses emojis for class names

lukemckinstry (approved):
Approved :+1: 

Not going to merge per @jjspace's request since this is dependent on another pending branch

packages/sandcastle/public/fonts/dejavu-sans-mono/DejaVu Fonts License.txt
```diff
@@ -0,0 +1,97 @@
+Fonts are (c) Bitstream (see below). DejaVu changes are in public domain.
```
ggetz: I'm a bit concerned about the license here. From what I can tell, this is not a standard license so we cannot guarantee  it's compatible with Apache 2.0.

Do we have an alternative with a standard license?

ggetz:
Can we include the font license for Fira? I believe it is [OFL](https://openfontlicense.org/) which _is_ compatible with Apache 2.0.

ggetz:
Could we also add font size to the settings? I think that's important for accessibility.

packages/sandcastle/src/SettingsModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+    <SandcastleDialog open={open} onClose={() => setOpen(false)}>
+      <SandcastleDialogHeading>Settings</SandcastleDialogHeading>
+      <div className="settings-row">
+        <div>Theme</div>
+        <div className="multi-option horizontal">
+          <IconButton
+            icon={sun}
+            // @ts-expect-error tone works but is not passed through the types from Button
+            tone={settings.theme === "light" ? "accent" : "neutral"}
+            label="Light Mode"
```
ggetz:
```suggestion
            label="Light mode"
```

packages/sandcastle/src/SettingsModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+              if (settings.theme === "dark") {
+                updateSettings({ theme: "light" });
+              }
+            }}
+          />
+          <IconButton
+            icon={moon}
+            // @ts-expect-error tone works but is not passed through the types from Button
+            tone={settings.theme === "dark" ? "accent" : "neutral"}
+            label="Dark Mode"
```
ggetz:
```suggestion
            label="Dark mode"
```

packages/sandcastle/src/SettingsModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+            onClick={() => {
+              if (settings.theme === "light") {
+                updateSettings({ theme: "dark" });
+              }
+            }}
+          />
+        </div>
+      </div>
+      <div className="settings-row">
+        <div>Editor Font</div>
```
ggetz:
```suggestion
        <div>Editor font</div>
```

packages/sandcastle/src/SandcastleDialog.tsx
```diff
@@ -0,0 +1,31 @@
+import { Dialog, DialogHeading, DialogProps } from "@ariakit/react";
+import classNames from "classnames";
+import { ReactNode } from "react";
+import "./SandcastleDialog.css";
+import { Text } from "@stratakit/bricks";
+
+export function SandcastleDialogHeading({ children }: { children: ReactNode }) {
+  return (
+    <DialogHeading>
+      <Text variant="headline-sm">{children}</Text>
```
ggetz, line 17: Perhaps a bit picky, but can we add a close button to the top right of the dialogues?

Image

packages/sandcastle/src/SettingsModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  defaultChecked={settings.fontLigatures}
+                  onChange={(e) => {
+                    updateSettings({ fontLigatures: e.target.checked });
+                  }}
+                  disabled={!selectedFont.supportsLigatures}
+                />
+              }
+            />
+            <Field.Label>
+              Use Font Ligatures
```
ggetz:
```suggestion
              Font ligatures
```

packages/sandcastle/src/SettingsModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+            </Field.Label>
+          </Field.Root>
+        </div>
+      </div>
+      <div className="settings-row">
+        <div>
+          Starting panel
+          <Text variant="caption-lg" className="caption">
+            Dictates which panel to start on when not loading a specific
+            Sandcastle
```
ggetz: I looked for a comparable setting to use as a reference. I found the Chrome settings have something similar:

Image

Maybe something like the following? I'd want to eliminate jargon like "panel" or even "Sandcastle" which may be ambiguous.

```suggestion
          On startup
          <Text variant="caption-lg" className="caption">
            Select the default view mode
```

ggetz (changes requested):
Overall this looks great! I suggested a few capitalization fixes for consistency. Our communication style guide says to prefer sentence case.

jjspace:
@ggetz thanks for the feedback, this should be ready for another look.
I added a new font size setting. Currently just a number limited between 5 and 50 to be reasonable. Might want to change it to a slider or dropdown later but this seems good for now
I also replaced DejaVu with Cascadia Code and JetBrains Mono as editor fonts. Both support ligatures too which is nice for those that like them.

ggetz:
Great, thank you @jjspace!

PR #12820, opened 2025-08-14
Always switch to editor when it should

jjspace:
# Description

Fixes a bug when you have a specific sandcastle loaded and click the "Open code" button on that same sandcastle in the gallery

## Issue number and link

Part of #12566 

## Testing plan

- Run sandcastle with `npm run dev` in the package or `npm run build-sandcastle` and `npm start` at the project root
- Open the gallery and click any sandcastle
- Once it's loaded click the "Open code" button for that same sandcastle
- it _should_ switch to the code and reload the sandcastle

# Author checklist


lukemckinstry:
I notice when I click the "open code" button for a sandcastle that is already open, it reloads the scene in the viewer. This seems unnecessary. Can we skip this?

I see toggling the "gallery" and "editor" buttons does not reload the scene in the viewer, which seems better.

jjspace:
> I notice when I click the "open code" button for a sandcastle that is already open, it reloads the scene in the viewer.

This was intentional to keep the behavior consistent. If you click the card for a gallery item it will load it even if it is already loaded. I meant this to be a way to "re-run" without switching to the code panel. I think this is good.
I left this behavior for the switch to code button too to make sure it's starting "fresh" when viewing the code but I realize it may be a bit too much. I've updated the code to avoid re-running when switching to code and it's already loaded.

(The "active" detection is a little messy now relying on the url. I plan to change this a bit later based on some other open PRs but not in this one to avoid conflicts)

lukemckinstry (approved):
Good to go once CI is passing

jjspace:
@lukemckinstry I've merged in the changes to prevent leaving with pending changes. I believe this should be good to merge now?

packages/sandcastle/src/App.tsx
```diff
@@ -590,34 +590,28 @@ function App() {
...
-                  setLeftPanel("editor");
+              }
+
+              if (switchToCode) {
+                if (!isAlreadyActive) {
+                  loadGalleryItem(item.id);
                 }
+                setLeftPanel("editor");
+              } else {
+                // Load the gallery item every time it's clicked ro act as a "rerun" button
```
ggetz:
```suggestion
                // Load the gallery item every time it's clicked to act as a "rerun" button
```

But moreover, I agree with Luke that this seems like confusing behavior.

> This was intentional to keep the behavior consistent. If you click the card for a gallery item it will load it even if it is already loaded. I meant this to be a way to "re-run" without switching to the code panel. I think this is good.

Why do we need to rerun from the gallery list at all?

jjspace: Without this behavior there is (currently) no way to "re-run" or reset a sandcastle without first switching to the code view. I don't know about others but I personally find I often want to reset sandcastles if I've moved the camera or edited values and just want a quick way to get back to the "start". Maybe we need to move the `Run` button or duplicate it but for now this solution felt like a good compromise.

jjspace: @ggetz did you have thoughts around this?

jjspace:
Possibly replaced by some of the logic in https://github.com/CesiumGS/cesium/pull/12854, revisit later

ggetz:
@jjspace May I suggest we close this PR until we agree on functionality?

jjspace:
Yes, I think it's actually working how I was expecting now anyway.

PR #12821, opened 2025-08-15
Clears globe depth even when terrain depth testing on

mzschwartz5:
# Description

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/BillboardCollection.js
+++ b/packages/engine/Source/Scene/BillboardCollection.js
@@ -1366,7 +1366,7 @@ function writeCompressedAttribute3(
     isHeightReferenceClamp(billboard.heightReference) &&
     frameState.context.depthTexture;
   if (!defined(disableDepthTestDistance)) {
-    disableDepthTestDistance = clampToGround ? 5000.0 : 0.0;
+    disableDepthTestDistance = clampToGround ? 50000.0 : 0.0;
   }
 
   disableDepthTestDistance *= disableDepthTestDistance;
--- a/packages/engine/Source/Shaders/BillboardCollectionVS.glsl
+++ b/packages/engine/Source/Shaders/BillboardCollectionVS.glsl
@@ -317,18 +317,17 @@ void main()
 if (lengthSq < disableDepthTestDistance) {
     float depthsilon = 10.0;
 
-    vec2 labelTranslate = textureCoordinateBoundsOrLabelTranslate.xy;
-    vec4 pEC1 = addScreenSpaceOffset(positionEC, dimensions, scale, vec2(0.0), origin, labelTranslate, pixelOffset, alignedAxis, validAlignedAxis, rotation, sizeInMeters, rotationMatrix, mpp);
+    vec4 pEC1 = addScreenSpaceOffset(positionEC, dimensions, scale, vec2(0.0), origin, vec2(0), pixelOffset, alignedAxis, validAlignedAxis, rotation, sizeInMeters, rotationMatrix, mpp);
     float globeDepth1 = getGlobeDepth(pEC1);
 
     if (globeDepth1 != 0.0 && pEC1.z + depthsilon < globeDepth1)
     {
-        vec4 pEC2 = addScreenSpaceOffset(positionEC, dimensions, scale, vec2(0.0, 1.0), origin, labelTranslate, pixelOffset, alignedAxis, validAlignedAxis, rotation, sizeInMeters, rotationMatrix, mpp);
+        vec4 pEC2 = addScreenSpaceOffset(positionEC, dimensions, scale, vec2(0.0, 1.0), origin, vec2(0), pixelOffset, alignedAxis, validAlignedAxis, rotation, sizeInMeters, rotationMatrix, mpp);
         float globeDepth2 = getGlobeDepth(pEC2);
 
         if (globeDepth2 != 0.0 && pEC2.z + depthsilon < globeDepth2)
         {
-            vec4 pEC3 = addScreenSpaceOffset(positionEC, dimensions, scale, vec2(1.0), origin, labelTranslate, pixelOffset, alignedAxis, validAlignedAxis, rotation, sizeInMeters, rotationMatrix, mpp);
+            vec4 pEC3 = addScreenSpaceOffset(positionEC, dimensions, scale, vec2(1.0), origin, vec2(0), pixelOffset, alignedAxis, validAlignedAxis, rotation, sizeInMeters, rotationMatrix, mpp);
             float globeDepth3 = getGlobeDepth(pEC3);
             if (globeDepth3 != 0.0 && pEC3.z + depthsilon < globeDepth3)
             {
--- a/packages/engine/Source/Scene/Scene.js
+++ b/packages/engine/Source/Scene/Scene.js
@@ -3456,9 +3456,7 @@ Scene.prototype.updateEnvironment = function () {
   }
 
   const clearGlobeDepth = (environmentState.clearGlobeDepth =
-    defined(globe) &&
-    globe.show &&
-    (!globe.depthTestAgainstTerrain || this.mode === SceneMode.SCENE2D));
+    defined(globe) && globe.show);
   const useDepthPlane = (environmentState.useDepthPlane =
     clearGlobeDepth &&
     this.mode === SceneMode.SCENE3D &&


packages/engine/Source/Shaders/BillboardCollectionVS.glsl
```diff
@@ -317,18 +317,17 @@ void main()
 if (lengthSq < disableDepthTestDistance) {
```
mzschwartz5: I removed the `labelTranslate` / pass in 0 now. I honestly don't know the original reason for it - the visual results seem much cleaner to me without it. With it, there's like padding around the billboards that causes it to render way after it's occluded by terrain.

jjspace:
This definitely seems improved at the global scale. However I noticed some weird "flickering" when zooming more locally. I assume this is just due to sampling or terrain loading or I just got lucky with the billboard "point" between some rocks that only obstruct it at certain angles. Probably still better than it was but thought it was worth noting.

https://github.com/user-attachments/assets/e5e90e16-6496-4319-a4a4-f17c0c5396a5

I also wanted to confirm, is this expected to impact 3dTiles or only Terrain? I'm still seeing it clip with the google 3d tiles

mzschwartz5:
I'm inclined to think the flickering predates this PR, as my changes should only make things more permissive on a global scale. Unless the `labelTranslate` field I messed with is responsible. I can look into it. Could also just be terrain loading as you said, it's hard to tell.

Interesting that it's still clipping on 3D tiles... I expected this to work on both terrain and 3D tiles, and I thought it *did*, in my testing.

edit- nope, this fix definitely only works for terrain. Will continue investigation into whether it can be extended to 3D tiles.

packages/engine/Source/Scene/BillboardCollection.js
```diff
@@ -315,6 +314,8 @@ function BillboardCollection(options) {
   ];
 
   this._highlightColor = Color.clone(Color.WHITE); // Only used by Vector3DTilePoints
+  this._coarseDepthTestDistance = 50000.0;
+  this._threePointDepthTestDistance = 5000.0;
```
mzschwartz5: Previously, the distance at which the three-point depth testing technique occurred was set to the `disableDepthTestDistance`, or 5km by default if not set. Setting it to `disableDepthTestDistance` was a strange and buggy choice, though, because the original author misinterpreted the setting. They reversed the meaning to *only* 3-point depth test within that distance. 

Since the original intent was to make close up viewing of billboards clamped to terrain better, I think always having it be 5km (while *respecting* the `disableDepthTestDistance` value) is okay.

packages/engine/Source/Scene/BillboardCollection.js
```diff
@@ -1365,9 +1376,6 @@ function writeCompressedAttribute3(
   const clampToGround =
     isHeightReferenceClamp(billboard.heightReference) &&
     frameState.context.depthTexture;
-  if (!defined(disableDepthTestDistance)) {
-    disableDepthTestDistance = clampToGround ? 5000.0 : 0.0;
```
mzschwartz5, line 1376: This is what I mentioned above. Removing in favor of always using a 5km distance.

packages/engine/Source/Scene/BillboardCollection.js
```diff
@@ -2023,8 +2031,7 @@ BillboardCollection.prototype.update = function (frameState) {
     ) {
       this._rsOpaque = RenderState.fromCache({
         depthTest: {
-          enabled: true,
-          func: WebGLConstants.LESS,
+          enabled: false,
```
mzschwartz5, line 2042: The crux of this PR's change. Disable automatic depth testing and do it manually in the shaders so we have more control of how it happens.

Note: Given the comment below about translucency, I'm a little wary I may have introduced regressions, as I don't do a `LEQUAL` check in the shaders, but it seems to work *better* this way (see test case in PR description)

mzschwartz5: Also note- we still leave automatic *write* to the depth buffer enabled, we just don't *test* against it automatically.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -14,18 +16,33 @@ in vec4 v_outlineColor;
 in float v_outlineWidth;
 #endif
 
+in vec4 v_compressed;                               // x: eyeDepth, y: applyTranslate & enableDepthCheck, z: dimensions, w: imageSize
+const float SHIFT_LEFT1 = 2.0;
+const float SHIFT_RIGHT1 = 1.0 / 2.0;
+
```
mzschwartz5, line 22: These changes, among several others in this PR, are just moving variables outside of macro checks. These used to be only used if `#ifdef FRAGMENT_DEPTH_CHECK` was defined, but now they're used all the time

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -159,50 +167,82 @@ void main()
...
 
-    float temp2 = (temp - floor(temp)) * SHIFT_LEFT1;
-    bool enableDepthTest = temp2 != 0.0;
-    bool applyTranslate = floor(temp) != 0.0;
+if (!enableDepthCheck) return;
 
-    if (enableDepthTest) {
-        temp = v_compressed.z;
-        temp = temp * SHIFT_RIGHT12;
+float eyeDepth = v_compressed.x;
```
mzschwartz5: More moving of variables outside compiler macros... And now we check if enableDepthCheck is turned on outside of the macro. This used to just mean "is the 3-point depth check enabled," but I repurposed it to mean "is depth testing enabled at all".

If it is, we do all forms of depth testing below (3-point, and regular)

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -14,18 +16,33 @@ in vec4 v_outlineColor;
...
+{
+    float logDepthOrDepth = czm_unpackDepth(texture(czm_globeDepthTexture, st));
+    if (logDepthOrDepth == 0.0)
+    {
+        return 0.0; // not on the globe
+    }
+
+    vec4 eyeCoordinate = czm_windowToEyeCoordinates(gl_FragCoord.xy, logDepthOrDepth);
+    return eyeCoordinate.z / eyeCoordinate.w;
+}
```
mzschwartz5, line 43: This isn't a new function as much as it is breaking the old getGlobeDepth function into two pieces, so I can call this piece separately.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
mzschwartz5: More moving of variables outside compiler macros... And now we check if enableDepthCheck is turned on outside of the macro. This used to just mean "is the 3-point depth check enabled," but I repurposed it to mean "is depth testing enabled at all".

If it is, we do all forms of depth testing below (3-point, and regular)

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -14,18 +16,33 @@ in vec4 v_outlineColor;
...
-const float SHIFT_RIGHT1 = 1.0 / 2.0;
+#endif
 
+float getGlobeDepthAtCoords(vec2 st)
+{
+    float logDepthOrDepth = czm_unpackDepth(texture(czm_globeDepthTexture, st));
+    if (logDepthOrDepth == 0.0)
+    {
+        return 0.0; // not on the globe
+    }
```
mzschwartz5, line 39: This isn't a new function, really. It just breaks the old `getGlobeDepth` function into two pieces, so we can call this piece separately.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
mzschwartz5: More moving of variables outside compiler macros... And now we check if enableDepthCheck is turned on outside of the macro. This used to just mean "is the 3-point depth check enabled," but I repurposed it to mean "is depth testing enabled at all".

If it is, we do all forms of depth testing below (3-point, and regular)

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -159,50 +167,82 @@ void main()
...
+temp = temp * SHIFT_RIGHT1;
 
-    temp = temp * SHIFT_RIGHT1;
+float temp2 = (temp - floor(temp)) * SHIFT_LEFT1;
+bool enableDepthCheck = temp2 != 0.0;
 
-    float temp2 = (temp - floor(temp)) * SHIFT_LEFT1;
-    bool enableDepthTest = temp2 != 0.0;
-    bool applyTranslate = floor(temp) != 0.0;
+if (!enableDepthCheck) return;
```
mzschwartz5: More moving of variables outside compiler macros... And now we check if enableDepthCheck is turned on outside of the macro. This used to just mean "is the 3-point depth check enabled," but I repurposed it to mean "is depth testing enabled at all".

If it is, we do all forms of depth testing below (3-point, and regular)

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
mzschwartz5, line 1: I would recommend reviewing the VS before the FS.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -159,50 +167,82 @@ void main()
...
+// If we're close, we compare against the globe depth texture (which includes depth from the 3D tile pass).
+vec2 fragSt = gl_FragCoord.xy / czm_viewport.zw;
+float globeDepth = getGlobeDepthAtCoords(fragSt);
+if (globeDepth != 0.0) {
+    float distanceToEllipsoidCenter = -length(czm_viewerPositionWC); // depth is negative by convention
+    float testDistance = (eyeDepth > -u_coarseDepthTestDistance) ? globeDepth : distanceToEllipsoidCenter;
+    if (eyeDepth < testDistance) {
+        discard;
+    }
+}
```
mzschwartz5: Again, the crux of this PR. This the manual depth testing that replaces automatic depth testing. Except, now, with manual control, we can depth test against a flat plane at global scales.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
```diff
@@ -248,10 +249,6 @@ void main()
     vec4 p = czm_translateRelativeToEye(positionHigh, positionLow);
     vec4 positionEC = czm_modelViewRelativeToEye * p;
 
-#if defined(FRAGMENT_DEPTH_CHECK) || defined(VERTEX_DEPTH_CHECK)
-    float eyeDepth = positionEC.z;
-#endif
-
```
mzschwartz5, line 256: Was never really necessary - `eyeDepth` is used in one place exactly. The point was to store off the value of `positionEC.z` before it gets mutated, but that's not necessary either. Now (down below) we just use this value where it's needed, but before the important mutations occur. (...but after we potentially set eye depth to 0 if the billboard fails the 3-point test).

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
```diff
@@ -309,26 +306,37 @@ void main()
...
     float disableDepthTestDistance = compressedAttribute3.z;
+    if (disableDepthTestDistance == 0.0 && czm_minimumDisableDepthTestDistance != 0.0)
+    {
+        disableDepthTestDistance = czm_minimumDisableDepthTestDistance;
+    }
+
+    if (lengthSq < disableDepthTestDistance || disableDepthTestDistance < 0.0)
+    {
+        enableDepthCheck = 0.0;
+    }
```
mzschwartz5, line 322: I repurposed `enableDepthCheck` - it used to mean "is the three-point depth check applied?" But now it just means what it sounds like "do we depth test at all?"

To accommodate this change, I had to move some chunks upwards in the vert shader. But it's actually more appropriate in this location anyway; i.e. all the `disableDepthTestDistance` logic is together now.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
```diff
@@ -309,26 +306,37 @@ void main()
...
 #endif
 
+    v_compressed.y = enableDepthCheck;
+
 #ifdef VERTEX_DEPTH_CHECK
-if (lengthSq < disableDepthTestDistance) {
+if (lengthSq < (u_threePointDepthTestDistance * u_threePointDepthTestDistance) && (enableDepthCheck == 1.0)) {
     float depthsilon = 10.0;
 
-    vec2 labelTranslate = textureCoordinateBoundsOrLabelTranslate.xy;
```
mzschwartz5, line 322: I honestly don't know what the purpose of `labelTranslate` here was, but it had the (bad) effect of making 3-point testing not work until a label was wayyyy occluded by terrain. Separate from all other changes in this PR, removing `labelTranslate` and using `vec2(0.0)` looks way better.

mzschwartz5:
@ggetz @jjspace, ready for re-review. See PR description: I left *detailed* test cases, and specifically read the note up top about reviewing commit-by-commit. I left a bunch of comments to guide review on one commit in particular.

packages/engine/Source/Scene/BillboardCollection.js
lukemckinstry: Can you explain this part (3 point depth test and the bug around misuse of `disableDepthTestDistance`) again more thoroughly, or point to where it is already written up? I remember talking about this in-person, but I want to make sure I (and others) can understand it clearly.

mzschwartz5: Yeah, I'll try my best - it's a little hard to wrap one's head around. First of all, [this blog](https://cesium.com/blog/2018/07/30/billboards-on-terrain-improvements/) describes the qualitative behavior of billboards clamped to terrain, before this PR. tl;dr - in order to improve billboard visibility, a special 3-point test is performed. The idea is: if any of 3 key points on a billboard is visible, the whole billboard should be visible. If all are occluded, the whole billboard gets occluded.

This special 3-point test was (before this PR) applied only if the camera was within a certain distance of the billboard. The way this distance was determined is what was buggy. The logic was, "if the camera is closer than `disableDepthTestDistance`, do this 3-point test." (And if `disableDepthTestDistance` is undefined, default to 5km). But this actually defeats the point of`disableDepthTestDistance`. `disableDepthTestDistance` is supposed to _disable_ occlusion within a given distance, but this 3-point test was actually fully occluding the billboard within that distance, if all 3 key points were occluded.

With this PR, I just hardcoded the distance to 5km - but, importantly, it should also respect `disableDepthTestDistance` now. Also importantly, the 3-point test still does what it was intended to do: preserve visibility of terrain-clamped billboards when viewed close up.

ggetz: Would it make sense for these values be either 1) exposed in the API for tweaking, or 2) specified relative to the ellipsoid size?

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
ggetz: Makes sense. Would you mind adding a comment below where `positionEC.z` is used explaining that the relevant checks must be performed there before the value is mutated? I could see it potentially getting moved if another developer doesn't have that context.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
```diff
@@ -311,26 +308,37 @@ void main()
...
 
+    float enableDepthCheck = 1.0;
 #ifdef DISABLE_DEPTH_DISTANCE
     float disableDepthTestDistance = compressedAttribute3.z;
+    if (disableDepthTestDistance == 0.0 && czm_minimumDisableDepthTestDistance != 0.0)
+    {
+        disableDepthTestDistance = czm_minimumDisableDepthTestDistance;
+    }
+
+    if (lengthSq < disableDepthTestDistance || disableDepthTestDistance < 0.0)
```
ggetz: Pedantic, but should `disableDepthTestDistance` actually be called ` disableDepthTestDistanceSquared` then?

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
ggetz: I'll need to verify but I'd expect labelTranslate is the value of the [`pixelOffset`](https://cesium.com/learn/ion-sdk/ref-doc/LabelGraphics.html?classFilter=label#pixelOffset), or maybe [`eyeOffset`](https://cesium.com/learn/ion-sdk/ref-doc/LabelGraphics.html?classFilter=label#eyeOffset), label property.

If that is the case, we might need to factor that back in.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -85,10 +93,89 @@ vec4 getSDFColor(vec2 position, float outlineWidth, vec4 outlineColor, float smo
 }
 #endif
 
+#ifdef FS_THREE_POINT_DEPTH_CHECK
+void doThreePointDepthTest(float eyeDepth, bool applyTranslate) {
+
+    if (eyeDepth < -u_threePointDepthTestDistance) return;
```
ggetz, line 99: Code style is to [always use `{}` curly brackets for any `if` blocks](https://eslint.org/docs/latest/rules/curly).

ggetz: The rest of this file has a few other occurrences. Please do a pass.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
ggetz: Well, I verified that both `pixelOffset` and `eyeOffset` are working just fine in this branch with both variants of the depth testing. No idea why `labelTranslate` was bing used then. 🤷‍♀️ 

This change should be OK as-is.

ggetz: Actually, `labelTranslate` may be handling offsets from label alignments. 

For example, [this label is right-aligned](https://ci-builds.cesium.com/cesium/billboard-terrain-clip-fix/Apps/Sandcastle2/index.html#c=rVRNT+MwEP0ro2glglScDUV7KKVaaCtAKhSVwl4igeNMUmsdu7LdQkH97+s4CaXsaaW9VJk3X29mnsuUNBbWHF9QwxlIfIEhGr4qyaPHwgPmzaGSlnKJ+qAD74kEsKi1A3pt+Ly2Sa5V+UtpkTVAeNipwqOoECrFHuRUGEzk9vA0kYlkvn1OGRfcbh60cBySgJCIl7RAE7UeUvA8CU7bBMNQoguteRNvfjgZLVFT5/UwqU3fLYrA6k3F330B1OGWCzRoXTx9ody28zCN1OKlUoXAu4WyytmCG8tZdzSvU8JqBl+p7rTUvOSWr9EQmmVhU7gJ2jpeli0gdGtR+nCPhBJIhCrC53HlA6FoxmUB+22hOwLfuGVMmgoA39590e1z22u3WpTWrW+3Km/zhqE/5FIZh6jdJYdUW/dFZdcfc4SFRjThURwfk/ikA90fJO7A9/quKRciVVRnvVoWAP5yvc837dSOBfJiYWeYo0bJ8KPf1T5OhpPzm7un+fTpcjZ9uB012WvUbglUTDUvPsnucQ8mF9P5fHrjU7b+V9AURa+hZvHV9py+LlrSk8p7nARNj9yJvPLHJ0unMSrNkUFdCe+/TLBQmr9Vz+jrDFdfHGR2fXk1/7e5z+/Hk+vbcT1587yCTtA3diNwUKE/eblU2sJKi9A9MIvlUjiFmyhdsd9OTcyYKqkftSn9jK+BZ2dJ8OUvIAmACWqM8+QrIe75GybBoB+5+L20RsdTN4KgmypkEQ8mNUgI6UfO/DvLKiVSqj9V/AM) and doesn't seem to be handled properly n three-point depth testing when goin behind terrain:

Image

Image

ggetz:
In addition to the test cases provided here, I tested the following billboard/label properties against both "variants" of the depth testing behavior and against the default ellipsoid, terrain, and Google P3DT (3D Tiles):

- `pixelOffset`
- `eyeOffset`
- `horizontalOrigin`
- `verticalOrigin`
- `scale`
- `showBackground `/`backgroundPadding`

I noticed the following during my testing:

1. Is the three-point depth test expected to work for [billboard/labels clamped to 3D Tiles](https://ci-builds.cesium.com/cesium/billboard-terrain-clip-fix/Apps/Sandcastle2/index.html#c=rVNdT9swFP0rVjSJIBVnpWgPoVSDtoNJhSII7CUSuM5NYs2xK9ttKaj/fbaT0I89TdpLlHvuPT7n+l5TKbRBSwYrUOgCCVihIWi2qPCzx8Ij6sOhFIYwAeqogz5SgVAUGVDKQnFLSOoY50pWv6TiWQOExx1HKLicQYxywjWkYnN8nopUUC+fE8o4M+snxa2HNMA4YhUpQEdtBhcsT4PzlqApCLCltW/sw88kJRUoYrMexnXo1Yxa1+brQsM4aDC2kqwIM20fVAExcC1lweG+lEbamDNtGO2NkpoSOveoEZgrVjHDlqAxybKwOdVVbKwXQ0sU2ouQ6nirLTlgLovwdewSiEuSMVGgfTXUGyGv1xrFjo7Qlw9/3ObVS2xvEYSxN7W9FR+zxpWXnkttEbkd2ZAoY/+I6PmpjaBQADo86XZPcfesg3rfcLeDvtYDnDHOZ5KoLK4bQcgPKd4dX6dOlMCK0jxADgoEhU+9m30cDyeXt/cvyfTl+mH6dDdq2EtQtn3Cp4oVO/v1vAfjq2mSTG89ZeO/nMyAx401A28mtqt01ZqeuOxpGjQaud1nl++eze06EaFPNCi3Y/+lg1Iq9u5ezGEPNwcJPBn/SP6t7cvH8eTn3bhuvHlIQSfoa7PmMHDod1bNpTJooXhon5KBas7tRutotqC/7RpRrR2pH7WUfsaWiGUXaXDw2NMAUU60tpl8wfkje4c0GPQjW79HaxZ4alvgZO1Kyu5gUoMY435kw79ZRko+I2rnxD8)? My expectation is no and that's totally fine for that to out of scope for this PR. But I wanted to confirm the intention.

image


2. (Also noted in https://github.com/CesiumGS/cesium/pull/12821/files#r2388391537) Horizontal label alignments do not seem to be properly handled for three-point depth tests against terrain, e.g., [this label that's right-aligned](https://ci-builds.cesium.com/cesium/billboard-terrain-clip-fix/Apps/Sandcastle2/index.html#c=rVRNT+MwEP0ro2glglScDUV7KKVaaCtAKhSVwl4igeNMUmsdu7LdQkH97+s4CaXsaaW9VJk3X29mnsuUNBbWHF9QwxlIfIEhGr4qyaPHwgPmzaGSlnKJ+qAD74kEsKi1A3pt+Ly2Sa5V+UtpkTVAeNipwqOoECrFHuRUGEzk9vA0kYlkvn1OGRfcbh60cBySgJCIl7RAE7UeUvA8CU7bBMNQoguteRNvfjgZLVFT5/UwqU3fLYrA6k3F330B1OGWCzRoXTx9ody28zCN1OKlUoXAu4WyytmCG8tZdzSvU8JqBl+p7rTUvOSWr9EQmmVhU7gJ2jpeli0gdGtR+nCPhBJIhCrC53HlA6FoxmUB+22hOwLfuGVMmgoA39590e1z22u3WpTWrW+3Km/zhqE/5FIZh6jdJYdUW/dFZdcfc4SFRjThURwfk/ikA90fJO7A9/quKRciVVRnvVoWAP5yvc837dSOBfJiYWeYo0bJ8KPf1T5OhpPzm7un+fTpcjZ9uB012WvUbglUTDUvPsnucQ8mF9P5fHrjU7b+V9AURa+hZvHV9py+LlrSk8p7nARNj9yJvPLHJ0unMSrNkUFdCe+/TLBQmr9Vz+jrDFdfHGR2fXk1/7e5z+/Hk+vbcT1587yCTtA3diNwUKE/eblU2sJKi9A9MIvlUjiFmyhdsd9OTcyYKqkftSn9jK+BZ2dJ8OUvIAmACWqM8+QrIe75GybBoB+5+L20RsdTN4KgmypkEQ8mNUgI6UfO/DvLKiVSqj9V/AM):

Image

Image

ggetz:
@mzschwartz5 Overall, this appears to be working super well! There are a few details about the behavior that need clarification, but I'm hoping they do not require a change to the overall approach here.

Could you please update `CHANGES.md` with a summary of the fix?

ggetz:
@lilleyse Would you be able to confirm the background translucency behavior described in the PR description? I'm not quite sure what the expect behavior is by default vs with OIT.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
mzschwartz5: Ooh interesting... that's basically the same problem that was occurring *with* label translate when left-aligned.

So I'm not sure what the solution is here, but maybe I just keep it as it was so at least the behavior is the same? I might spend a little time seeing if there's a complete solution.

packages/engine/Source/Scene/BillboardCollection.js
mzschwartz5: Yeah, maybe just relative to the ellipsoid for now? If we get requests to expose these values in the API, easy enough to add later.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
mzschwartz5: I wouldn't call that pedantic... in fact, it may actually be a bug. I'd need to go see if `compressedAttribute3.z` is actually squared before being passed in. Either way, I agree that it should be updated.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
mzschwartz5: Will do. Usually prettier catches me on those but I guess not in shader code!
(My personal preference for return-only if statements is to omit braces, so these are bound to continue to leak through in the future...)

mzschwartz5:
@ggetz

> Is the three-point depth test expected to work for [billboard/labels clamped to 3D Tiles](https://ci-builds.cesium.com/cesium/billboard-terrain-clip-fix/Apps/Sandcastle2/index.html#c=rVNdT9swFP0rVjSJIBVnpWgPoVSDtoNJhSII7CUSuM5NYs2xK9ttKaj/fbaT0I89TdpLlHvuPT7n+l5TKbRBSwYrUOgCCVihIWi2qPCzx8Ij6sOhFIYwAeqogz5SgVAUGVDKQnFLSOoY50pWv6TiWQOExx1HKLicQYxywjWkYnN8nopUUC+fE8o4M+snxa2HNMA4YhUpQEdtBhcsT4PzlqApCLCltW/sw88kJRUoYrMexnXo1Yxa1+brQsM4aDC2kqwIM20fVAExcC1lweG+lEbamDNtGO2NkpoSOveoEZgrVjHDlqAxybKwOdVVbKwXQ0sU2ouQ6nirLTlgLovwdewSiEuSMVGgfTXUGyGv1xrFjo7Qlw9/3ObVS2xvEYSxN7W9FR+zxpWXnkttEbkd2ZAoY/+I6PmpjaBQADo86XZPcfesg3rfcLeDvtYDnDHOZ5KoLK4bQcgPKd4dX6dOlMCK0jxADgoEhU+9m30cDyeXt/cvyfTl+mH6dDdq2EtQtn3Cp4oVO/v1vAfjq2mSTG89ZeO/nMyAx401A28mtqt01ZqeuOxpGjQaud1nl++eze06EaFPNCi3Y/+lg1Iq9u5ezGEPNwcJPBn/SP6t7cvH8eTn3bhuvHlIQSfoa7PmMHDod1bNpTJooXhon5KBas7tRutotqC/7RpRrR2pH7WUfsaWiGUXaXDw2NMAUU60tpl8wfkje4c0GPQjW79HaxZ4alvgZO1Kyu5gUoMY435kw79ZRko+I2rnxD8)? My expectation is no and that's totally fine for that to out of scope for this PR. But I wanted to confirm the intention.

That's correct. [The code that determines whether 3-point testing occurs](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Scene/BillboardCollection.js#L1424-L1429) checks for `depthTestAgainstTerrain` - so it's terrain specific, not for 3D-tiles. I do think it could be easily enabled for 3D tiles, if we wanted to follow up with that.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
ggetz: Fair! I'd love to eventually get our shader tooling up to par with what we have for JS.

lilleyse:
> @lilleyse Would you be able to confirm the background translucency behavior described in the PR description? I'm not quite sure what the expect behavior is by default vs with OIT.

The new behavior is consistent with how OIT works. Though I'm curious what actually caused the change.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
mzschwartz5: [It is squared before being sent to the GPU](https://github.com/CesiumGS/cesium/blob/6a5a0c469a6d3da9ce491855346f588deda24d90/packages/engine/Source/Scene/BillboardCollection.js#L1388). So it's a misnomer of a variable, I'll change it.

ggetz:
> Though I'm curious what actually caused the change.

@lilleyse This PR changes how the depth test for billboards work in order to support manual but more robust conditions for billboard display, such as what is described [in this blog post](https://cesium.com/blog/2018/07/30/billboards-on-terrain-improvements/). Much of the functionality was already there, though not working as intended.

@mzschwartz5 Could you please fill in @lilleyse with any more details of the translucency updates?

mzschwartz5:
@ggetz @lilleyse I'm. not sure there's much more detail to fill in here. The gist of this PR is that it disables automatic depth testing of billboards and switches to manual depth testing in the shader itself so we can have more control over the process.

In disabling automatic depth testing, I removed [this code](https://github.com/CesiumGS/cesium/blob/5d9cbb7a1093cabce4be66e8bb9d0daa5ff534bf/packages/engine/Source/Scene/BillboardCollection.js#L2056-L2058) that sets the depth check to `LEQUAL` when `useTranslucentDepthMask` is true. I did not reimplement `LEQUAL` in the manual depth check - so everything just uses "less than" checks, effectively. But, for whatever reason, that actually seems to _fix_ an existing bug, as shown in the screen caps at the end of the PR description.

lilleyse:
@mzschwartz5 thanks for pointing that out. I'm not super familiar with the context behind [this comment](https://github.com/CesiumGS/cesium/blob/5d9cbb7a1093cabce4be66e8bb9d0daa5ff534bf/packages/engine/Source/Scene/BillboardCollection.js#L2042-L2045), but if billboard quality looks good or better with OIT now, then I don't see any concern.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
mzschwartz5: I pushed a commit to address this issue - it's not what we thought was going on:

1. In fact, we *do* want `labelTranslate` to be (0, 0) here. That field is used by `addScreenSpaceOffset` to offset individual glyphs according to where they live in a text string. In this case, we're depth testing three key points of the whole label - we don't want or need any glyph-specific offsets. So the pre-existing use of `labelTranslate` is definitely a bug.
2. The real issue is the value of `origin` being passed to `addScreenSpaceOffset` in the three-point test. The problem is that in `rebindAllGlyphs`, [glyph billboards are _always_ given a horizontal origin of `LEFT`.](https://github.com/CesiumGS/cesium/blob/4c5b2d5d5249586b8e013521d8613d5e293ac346/packages/engine/Source/Scene/LabelCollection.js#L308). This is purposeful, for glyph alignment, but it breaks the three-point test for anything other than `LEFT`. What we _ought_ to be doing in the shader, is using a different vertex attribute altogether for labels. Not `origin` (which comes from `compressedAttribute0`) but `labelHorizontalOrigin` (which comes from `compressedAttribute2`). Making this change (and keeping the change where `labelTranslate` is always (0,0)), the behavior now works well for `LEFT` and `RIGHT` -aligned labels.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -85,10 +93,89 @@ vec4 getSDFColor(vec2 position, float outlineWidth, vec4 outlineColor, float smo
...
+
+    // Automatic depth testing of billboards is disabled (@see BillboardCollection#update).
+    // Instead, we do one of two types of manual depth tests (potentially in addition to the test above), depending on the camera's distance to the billboard fragment.
+    // If we're far away, we just compare against a flat, camera-facing depth-plane at the ellipsoid's center.
+    // If we're close, we compare against the globe depth texture (which includes depth from the 3D tile pass).
+    vec2 fragSt = gl_FragCoord.xy / czm_viewport.zw;
+    float globeDepth = getGlobeDepthAtCoords(fragSt);
+    if (globeDepth == 0.0) return; // Not on globe
+    
+    float distanceToEllipsoidCenter = -length(czm_viewerPositionWC); // depth is negative by convention
```
ggetz, line 167: Sorry, didn't notice this until now— Does `length(czm_viewerPositionWC)` make sense here? I think `PositionWC` will be world coordinates, while isn't `eyeDepth` in eye coordinates?

Out of curiosity, I was able to do a quick test and get a fallback depth based on computing the depth of the intersection with the ellipsoid (see Kevin's [horizon culling](https://cesium.com/blog/2013/04/25/horizon-culling/) blog post for more info) through each vertex. I pushed up my patch to [`billboard-horizon-patch`](https://github.com/CesiumGS/cesium/compare/billboard-terrain-clip-fix...billboard-horizon-patch?expand=1).

It seems to be working pretty well for mid-level horizon views as well as at the global scale:

| This branch | Ellipsoid test |
| --- | --- |
| Image | Image |
| Image | Image |

And it seems to be working well for 3D Tiles as well, so in that branch I was able to remove the custom `depthTestDistance` callbacks in the Moon and Mars example.

Thoughts?

ggetz (commented):
Thanks @mzschwartz5! This is looking pretty good after another round of testing. I have one more comment which shouldn't need to hold up this PR if you disagree.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
mzschwartz5: So, I do think `length(czm_viewerPositionWC)` makes sense here- basically, we want to compare the distance between the camera and the globe center, to the distance between the camera and billboard. While _position_ depends on frame of reference, _distance_ is invariant under rigid transforms. `czm_viewerPositionWC` is the position of the camera in world coords, so taking its length gets us the camera's distance relative to the globe center, which is the same in the world frame as in eye space. Does that make sense?

Separately (but perhaps making the above irrelevant) I didn't even consider horizon culling, and your images of it look aesthetically better to me. I think we should take your patch. (I wonder- was horizon culling being done on billboards *before* this PR? I think not, right?)

mzschwartz5: Re: mars and moon, I was thinking about removing the depth distance check after this PR, even as I was writing the Mars sandcastle. I'm eager for that because I don't think that stop gap worked particularly well.

ggetz: > While position depends on frame of reference, distance is invariant under rigid transforms.

Thanks for the explanation! That makes sense in retrospect.

> I wonder- was horizon culling being done on billboards before this PR? I think not, right?

Not to my knowledge, no.

ggetz: >  I think we should take your patch.

:+1: I'll go ahead and merge this PR then open a new one with my patch for your review.

mzschwartz5: After more thought, I think my above comment was wrong (well, what I said about position vs. distance is true, but the conclusion is wrong). My code was just comparing two distances (camera-to-globe-center, to camera-to-billboard). But really it should have been projecting the camera-to-billboard vector _onto_ the camera-to-globe vector, and comparing the length of the projected vector to the camera-to-globe distance. 

And, for that math, the reference frames must be the same - so I think you were right to be confused. In any case, the horizon culling approach will fix that _and_ be more precise. Just wanted to clarify for posterity, and so I'm not gaslighting anyone :)

Issue #12817, opened 2025-08-14
Multi-source terrain provider feature


juunie-roh:
### Feature

![Image](https://github.com/user-attachments/assets/e4cdb8d4-1bfa-490e-8262-15e9cb3ea38f)
Mixed terrain example with `worldTerrain` and `Ellipsoid`.

> This issue was originated from [Cesium Community](https://community.cesium.com/t/utility-library-for-cesiumjs/42484).

### What is it for:

Enables terrain to be loaded from multiple sources according to configured areas(with specific zoom level and tile coordinates) as the image shown above. It also enables setting fallback terrain, an error boundary for terrain provider.

### Concept

I have implemented the feature on [@juun-roh/cesium-utils](https://www.npmjs.com/package/@juun-roh/cesium-utils), with name `HybridTerrainProvider`. It works as below:

1. The `HybridTerrainProvider` extends existing `TerrainProvider` class.
2. Configure areas to override terrain source with zoom levels, tile coordinates, and `TerrainProvider` having different terrain sources.
3. Override `requestTileGeometry` method with configured area inclusion checks.
4. Delegate the method to configured terrain provider's `requestTileGeometry`, with configured areas.
5. If returns false, proceed with default/fallback provider's.

The overall concept of this feature is to redirect the `requestTileGeometry` with different providers.

### Implementation

You can see my full implementation [here](https://github.com/juunie-roh/cesium-utils/blob/main/src/terrain/hybrid-terrain-provider.ts).

Briefly, how I redirected the method is by this:

```typescript
requestTileGeometry(
    x: number,
    y: number,
    level: number,
    request?: Request,
  ): Promise<Awaited<TerrainData>> | undefined {
    if (!this._ready) return undefined;

    // Check regions for a match
    for (const region of this._regions) {
      if (this._regionContains(region, x, y, level)) {
        return region.provider.requestTileGeometry(x, y, level, request);
      }
    }

    // Fall back to default provider
    if (this._defaultProvider.getTileDataAvailable(x, y, level)) {
      return this._defaultProvider.requestTileGeometry(x, y, level, request);
    }

    // Final fallback
    return this._fallbackProvider.requestTileGeometry(x, y, level, request);
  }
```

### Request for Guidance

I've tried to maintain coherency with Cesium's API design, but since it's having separate development environment, I'm expecting it to take some times to open a direct pull request about this feature.

Meanwhile, I'd appreciate for feedbacks on this feature:

[...]

PR #12822, opened 2025-08-18
Multi source terrain provider

juunie-roh:
# Description

This is an initial draft of migration from external library TypeScript implementation into Cesium development environment.

The naming conventions are not considered in this draft.

You can see the detailed description about this feature at the related issue, #12817.

## Issue number and link

This PR relates to #12817.

## Testing plan

You can see conducted unit tests at [external library repository](https://github.com/juunie-roh/cesium-utils/blob/main/src/terrain/hybrid-terrain-provider.test.ts).

# Author checklist


juunie-roh:
There is a limitation in the current implementation:
* Cannot override higher zoom levels within configured terrain regions

For example:
* Level 15 region configured with Provider A
* Level 16+ in the same geographic area should use Provider B
* Currently: Level 16+ continues using Provider A

What causes this is:

* Simple region designation cannot change `TerrainAvailability`.
* Once a region matches, it returns immediately without checking further options.

This could be solved by nesting HybridTerrainProvider instances, but I'm not sure if it's the right direction.

I haven't deeply investigated this limitation yet, as it may be an edge case.

ggetz:
Thanks for the PR @juunie-roh, and thanks for your patience!

Would you mind describing the use case(s) you're looking to address here? It seems like this PR address some, but not all, of the use cases we're looking into for https://github.com/CesiumGS/cesium/issues/12817. For instance, I don't believe this handles showing multiple terrain surfaces simultaneously.

However, it does appear like this could be useful for many users and would likely still be worth consideration.

juunie-roh:
Hi, @ggetz ! 👋 
You're right, that this is not for showing multiple terrains at once.
My specific use case was to **integrate terrains** from different sources **seamlessly**.
I set a global terrain fallback, then configured the specific area where needed to show terrain with different source.(higher resolution could be one of the example)
This implementation solves the problem by delegating `requestTileGeometry` to different terrain providers based on tile coordinates and zoom levels.

https://juun.vercel.app/cesium-utils/terrain
I have working demonstration here. The global fallback terrain with `EllipsoidTerrainProvider`, and configured area with Cesium's default provider at zoom level 13. (red-colored area).

ggetz:
Thanks @juunie-roh! Your demonstration looks to be working great.

We'll take a closer look at this PR and review shortly. Please let us know if there are any updates to the status of this PR in the meantime.

juunie-roh:
Hey @ggetz, it's been a while!

I found a way to solve the addressed limitation I stated above. By letting `getTileDataAvailable` to return `true` always on defined region, it no longer upsamples the latest available terrain data, delegate the request successfully to the region's provider.

However, I haven't applied it to this PR. Just let me know if you are willing to proceed with this PR.

image

> There is a limitation in the current implementation:
> 
> * Cannot override higher zoom levels within configured terrain regions
> 
> For example:
> 
> * Level 15 region configured with Provider A
> * Level 16+ in the same geographic area should use Provider B
> * Currently: Level 16+ continues using Provider A
> 
> What causes this is:
> 
> * Simple region designation cannot change `TerrainAvailability`.
> * Once a region matches, it returns immediately without checking further options.
> 
> This could be solved by nesting HybridTerrainProvider instances, but I'm not sure if it's the right direction.
> 
> I haven't deeply investigated this limitation yet, as it may be an edge case.

lukemckinstry:
Hi @juunie-roh apologies this fell off the radar.

You mentioned finding a fix for "Cannot override higher zoom levels within configured terrain regions" but not adding to this PR. Did you add that to the external package you shared earlier https://github.com/juunie-roh/cesium-utils? Are there other features you are adding, refining, or simply unsure about in HybridTerrainProvider?

If HybridTerrainProvider is still a work in progress, it may make sense more sense to continue to maintain it in your separate package for the time being and resubmit later if you wish. This way it is available to the developer community but there is less friction for you to continue development versus if it was checked into Cesium. 

If you want to add your latest updates to this branch (and resolve merge conflicts), we would be happy to review again. 

Let us know what your preference is, we can close this PR if you decide not to pursue it at this time.

(ps: I noticed the demo link for your package appears down https://juun.vercel.app/en/cesium-utils)

juunie-roh:
Hi @lukemckinstry !
Since I was uncertain about whether this PR would move forward, I decided to leave a note here.

The HybridTerrainProvider works for specific use cases:

- Showing different terrain sources based on tile coordinates (level, x, y)
- Fallback to the native EllipsoidTerrainProvider on error

My primary concern was to show different terrain sources *seamlessly*, simplifying its usage by extending existing TerrainProvider.

However, I am not certain about the others, such as resource management when handling multiple providers and potential terrain cracking issues (skirt generation) at the boundaries of different sources.

If you think this feature is valuable for the Cesium core, I would be happy to resolve the merge conflicts and refine the code based on your architectural guidance. Let me know your thoughts!

(ps: The demo link is under fix due to the stricter [octal escape sequences checks of the latest NextJS build process](https://github.com/CesiumGS/cesium/issues/13379). Thanks for pointing it out!)

lukemckinstry:
> potential terrain cracking issues (skirt generation) at the boundaries of different sources.

There is logic in the terrain provider to handle skirt generation where adjacent tiles are at different LODs and cracks appear, so you could reference that. 

> resource management when handling multiple providers 

I am less certain what complications may arise, at a high level if you follow the control flow for the single terrain provider - destroy resources when they go out of use - I think it will be ok, but keep an eye out for lower level optimizations in the current terrain code.

juunie-roh:
The [demo link](https://juun.vercel.app/ko/cesium-utils/terrain) is back online.

image

From the thick yellow lines at the edges of terrain tiles from the image, I guess the skirt generation is working, and cracks appear just because the height difference is larger than the default `skirtHeight`. (Cesium world terrain vs Ellipsoid)

```js
const skirtHeight = provider.getLevelMaximumGeometricError(level) * 5.0
```

I'm also refining other properties (credit, errorEvent, hasVertexNormals, hasWaterMask) to ensure full compatibility. I’ll push the update as soon as it's ready!

juunie-roh:
Pushed with temporary demo identical to the external demo link, at root `HybridTerrainProviderDemo.html`

Few things:

- `credit` → concatenates all credits from registered providers (if present)
- `hasVertexNormals` & `hasWaterMask` → return true if any of registered provider has one.
- `errorEvent` → added listeners to each registered providers to forward all possible errors.

lukemckinstry:
A few comments & questions, apologies for the delayed response 
- Could you format your demo as a sandcastle? See https://github.com/CesiumGS/cesium/tree/main/packages/sandcastle/gallery
 - Do you have any sample data with which to demonstrate besides CesiumWorldTerrain as the detail and Ellipsoid Terrain Provider as fallback?
 - The jsDocs you started look good, I am getting an error when they build, please try `npm run build-docs` and resolve the errors
 - In terms of user workflow, I am wondering if supplying tile coordinates is the most user friendly? In the equivalent process for 3D Tiles in CesiumJS (detailed tileset with global fallback) users will often use clipping planes to place a detailed tileset within a global tileset. So the global tileset is clipped and the smaller tileset (ie. of a building, campus, etc) is located in the clipped area https://sandcastle.cesium.com/?id=3d-tiles-clipping-planes. Clipping planes exist in Terrain too https://sandcastle.cesium.com/?id=terrain-clipping-planes, but I do not **think** there is logic to support placing a smaller terrain subset inside in the clipped area (maybe this is possible, we should confirm) as your approach aims to enable. We should at least consider making the workflow for Terrain mirror the clipping workflow for 3D Tiles.    
 - What is the logic behind having `this._defaultProvider`, `this._fallbackProvider`, and `this._regions`, as opposed to just 2 options: regions and default?

juunie-roh:
I have formatted the demo under "multiple-terrain-sources", and updated jsDocs to pass scripts `build-docs` and `build-ts`.

- The terrain source I used in production is proprietary, so I'm afraid I cannot use it for the public demonstration. Do you have a sample detailed terrain asset from Cesium ion I could use for the demo that can comparable with CesiumWorldTerrain?
- `this._fallbackProvider` was intended to be a safety net for an asnyc-ready model failure, but since the provider now only reaches ready when all sub-providers are ready, that intent no longer applies. I have simplified those into two options as you suggested.

**On the clipping-planes workflow comparison**:

I considered a `Rectangle` based API during development and withdrew it for several reasons.

1. **As far as I know**, there is no runtime mechanism to synthesize partial tile data. 
2. The `TerrainProvider`'s contract assumes complete terrain tiles as the atomic unit.
3. `TerrainProvider` doesn't expose raw elevation data during runtime.

3D Tiles, on the other hand, are possible to replace clipped area because it handles geometric information during runtime directly.

In my case, I was able to control the terrain source by lon/lat level. I used [Cesium Inspector](https://sandcastle.cesium.com/?id=cesium-inspector) — terrain section — to track down tile coordinates and its lon/lat ranges.

(p.s. The provider assumes each sub-providers to have the same `TilingScheme`)

Issue #12815, opened 2025-08-13
Cesium.Model & modelMatrix: Cesium.Matrix4.IDENTITY & 2D Mode


mickae1:
### What happened?

when switching to 2D mode or 2.5 mode, the Cesium.Model points are not updated ( as expected ) and the result is not what the user expected, I would like to do the work in the vertex shader, but i don't know how to do it.

I need to have Cesium.Matrix4.IDENTITY, because i've thousands trajectories ( lines )

 around the globe loaded in a glb file which are  much faster.

I create this issue, because cesium should provide a function or matrix to transform a world coordinate ( ecef )  to appropriate coordinate.

when i'm printing the vertex shader to the console, I can see the function geometryStage, and as expected we are not using USE_2D_POSITIONS.

we are doing v_positionEC = (modelView * vec4(positionMC, 1.0)).xyz; 

modelView is set in ModelVS.glsl with : 

mat4 modelView = czm_modelView;

not sure it's correct.

### Reproduction steps

i attached a glb file that show a big surface, but in 2D mode or 2.5D mode the result is not correct.

[test.zip](https://github.com/user-attachments/files/21750471/test.zip)

Load a GLB/glTF file where the vertex positions are already in ECEF world coordinates.

Create a Cesium.Model from this asset.

Set the modelMatrix of the model to Cesium.Matrix4.IDENTITY.

Load the application and observe the model in 3D view (it appears correct).

Switch to 2D or 2.5D Columbus View.

Observe that the model is now incorrectly rendered.

lukemckinstry:
Hi @mickae1 do you mind elaborating a bit?

[...]

PR #12825, opened 2025-08-20
add the option customGeometryStage to let the user customize the geometry stage function in the vertex shader

mickae1:
to test : 
[test.zip](https://github.com/user-attachments/files/21891878/test.zip)

[Sandcastle](http://localhost:8080/Apps/Sandcastle/#c=nVf9bts2EH8VLv1H6ly5XboWcJJiruO6BvyF2GkXwIDLSLTNjhINkkqstP13L7V32F5pR1KSKVtZvQYGIpLH+933HUOeSIUIFmp9hSOaSnSBXp2+evX6+eszpP/myTzpEEnTOOgyRjeS0yiIyBKnTAFtQu7R/rE3T5DLsoGaTbRFwizqzrL6M6TPHsoz/0zLEhqB7yi5J6KK/8HsefOT0Kw7PFGYJkTMTxroi2a9YvyWtNwrPb3jPaaeP0++WdAcL5AhSciQR2RCwz9grbf1kgVRKrCiPAGRnrtyEqHIdrrGkZb2k5YCwWb4Eq0Ij4kS2VThFfFowlOFJoKD7JJEbaUEvU0VkQiXnw0UY/USxRpP62rWpyjhIsbMt6y/2H8lSsjjDdyNJlxSLZ7VJv8D63bsOdrk5xLRBEV0uSSCJAqucxHRBAOFzKQisazwPy3vDTug+E7UYLd/5txYVMi/Q9PVNF6pLXpqNPJ2txroRfDc94Nt9lBRiy698CFexFxs1jMaE/TTBVD6jmkOxLdY+tbdEVAum30DAx/NZiP4ZxKajT1m3ZzZ2VztOH0jTJIvP8jZNVnBe8dZW6bi8qmOOzCLRGpNUJhKxWMkbYTiJGpygaTKGIE8IZEmoo7Tn9AlguyAtIq8zng4uZ51F5PxtD/rj0eLj51F53o6Gw8X0/fty+6Vj75+/U/q6exm0P0uVXs2HE8n77tXVdLuqP120F10Bv3JpD/qwZXBTW88mvoVbUccYletsTKKIPhhdPrLs1uqSuej+zUN15BMUIaI1GUNcuAzVQoMAqcSsotVeMoQM4jxuqD9WAaSCdxjI+kJSSDpXEeBoUFR9L49XYzGV8P2YOrC2Zw3UWs/6QPx7BdAOnlo94Yd/1iwWXvU645mFTSFkxVUgyPgcsr/gfe2X4cI7jkas6T9Hip4rs0Y4hD2At1hkdFkJU2UI8XRLUFRluCYgnNZBuUZOgfUvQiCocKil9ftCd0QBoFo6veOQhJ1adl8yBG8nax+pU4JolKR1JXoMnF3hcpm/YwD/ErgDUSsZw8WJCTLBloyDjGOmaIqjUixXrJ3OFRc1HaHonVULGz5+ft9YoAtY+T986evs0jXDjA7lImlgPqhl7/foA3DCdBgEUrIIb5ED76TJlYmlrNaQF/XTQMoc78GD4e4PFnlwH//tQcMCAewEAoaNmtua4ALXgUyEBfI20ZugyA7kGFrq4Z2NhZQJjYbcKtGKRkeQIFAiiy2AJIu3IHmaVWIfaTsUaTcZo8AZXVAjpnrwk4HgFfEC9AXodIohC8+Mv8gImFGKiLyqCjMGdshpyyOH8xsNE4VhD+6k/ZDt+mDBn5R28DzG0FlqKioZHDRz1Xb+HBSF+91Pbgeoy4ZSxM4ytvU06zrDGhHwyGMqJ61RD+xhjD/j7TWLqWrpgcpGUlWau3VqOCjZ1WTlGWzdOrhJVezdFG4tCBETpB8cuZfO14U8687enecE89owehqrSDgzUDdKugG7m5wPRr0Zw1N7U7WM7JVrcqOIUkhksDRslVYqaJ0a2c7lW1ICXhtb81gL3g3GLctnHEbZinQOTxynRsFe2uWH+b84tfSiOXkVr5BrEFNqZsIGoNb7rSb3ZdJsCkOZIAj+w7D9xhmnVwCa0RdsXtMLdsyS0KvsI0Am89PgqAJvymON4xcQoVsmuCTTeheKlixW3hMWXqzP8TQ2ralgnb5MuhfQkPvz25yUjcIWpVVhaDnvohaSInU9NVvfqN4+5VG2M3AewaAmjkpD80d07KV6RTQlomupmbu1YxEagoEdH8awQBBl5k5c9hr/1lUzkjA+MrbHQYOj0Dz1kI+QrqQJKYx/sxFe0vNHHDSODk3c/abIgh+o/GGC6U94YEP4LVlCrBs3qbw2FRBKGWRZedN9+p5RO9Ag4ualy8KGZYSTpYpY1MoePOTN+dNoD+4CuUDXnqrMWQRw5kmW794M7CbQRCcN2FZf1Nxzm6x2OP8Lw)


mickae1:
@lukemckinstry @ggetz   do you think this pull request can be accepted for the next cesium release ? My teams need it.

Thank you,

mzschwartz5:
Hi @mickae1! Thanks for opening a PR and contributing this feature. I have a couple questions to kick off my review:
1. Is there a GH issue or forum post that discusses this feature? If so, can you link it so we maintain the context of this feature for posterity? If not, it might be a good idea to open an issue or write a forum post to get feedback / direction.
2. One thing I'd want to discuss in such an issue is the best way to implement this feature that fits in with the Cesium's design. Right now, you're using a compiler macro to ignore Cesium's built-in geometry stage function when a flag is set. I think a more natural way to do this would be to [pass in and override with your new geometry stage here](https://github.com/CesiumGS/cesium/blob/4c186a280f669473d658bba4aa1848c374dd2aad/packages/engine/Source/Scene/Model/GeometryPipelineStage.js#L193-L194). That way, the engine is never potentially in a situation where there's *no* geometry stage function (if, say, you forget to supply one). It's guaranteed to have at least the default.

But beyond the actual implementation, I'm curious what it is you're trying to achieve with this new feature - maybe we already support what you want to do some other way! (Again, why it's often helpful to start with or link a forum post).

I'll assign myself as a reviewer so I can get notified when you reply!

packages/engine/Source/Scene/Model/Model.js
```diff
@@ -166,6 +166,7 @@ import ModelImagery from "./ModelImagery.js";
  * @privateParam {boolean} [options.showCreditsOnScreen=false] Whether to display the credits of this model on screen.
  * @privateParam {SplitDirection} [options.splitDirection=SplitDirection.NONE] The {@link SplitDirection} split to apply to this model.
  * @privateParam {boolean} [options.projectTo2D=false] Whether to accurately project the model's positions in 2D. If this is true, the model will be projected accurately to 2D, but it will use more memory to do so. If this is false, the model will use less memory and will still render in 2D / CV mode, but its positions may be inaccurate. This disables minimumPixelSize and prevents future modification to the model matrix. This also cannot be set after the model has loaded.
+ * @privateParam {boolean} [options.customGeometryStage=false] Whether to override the default geometry stage, to let the user define it's own geometry stage.
```
mzschwartz5: Small nit- `it's` -> `its`.

But, per my overall comment, I think it might make more sense to do something like pass in the geometry stage function snippet itself rather than a flag.

mzschwartz5:
Oh, also - if you haven't already, can you add yourself to `CONTRIBUTING.md` and add your feature to `CHANGES.md`?

mickae1:
> Hi @mickae1! Thanks for opening a PR and contributing this feature. I have a couple questions to kick off my review:
> 
> 1. Is there a GH issue or forum post that discusses this feature? If so, can you link it so we maintain the context of this feature for posterity? If not, it might be a good idea to open an issue or write a forum post to get feedback / direction.
> 2. One thing I'd want to discuss in such an issue is the best way to implement this feature that fits in with the Cesium's design. Right now, you're using a compiler macro to ignore Cesium's built-in geometry stage function when a flag is set. I think a more natural way to do this would be to [pass in and override with your new geometry stage here](https://github.com/CesiumGS/cesium/blob/4c186a280f669473d658bba4aa1848c374dd2aad/packages/engine/Source/Scene/Model/GeometryPipelineStage.js#L193-L194). That way, the engine is never potentially in a situation where there's _no_ geometry stage function (if, say, you forget to supply one). It's guaranteed to have at least the default.
> 
> But beyond the actual implementation, I'm curious what it is you're trying to achieve with this new feature - maybe we already support what you want to do some other way! (Again, why it's often helpful to start with or link a forum post).
> 
> I'll assign myself as a reviewer so I can get notified when you reply!

you can find here the issue : 

https://github.com/CesiumGS/cesium/issues/12815

mickae1:
for my model loaded from glb, i want to do the projection change in vertex shader when I'm in 2D/2.5D . ( i don't want any cpu computing the 2D projection )

the code works, the projection works.

you can find here the  [sandcastle ](http://localhost:8080/Apps/Sandcastle/#c=tVbhTuM4EH4VX3+lez0XxIqVWkCHWBYhwYKuXaSTKnVN4rTec+zIdqBhxbvf2I4TpwRpddIhBLFn5puZb2Zsp1JogyhRZvsXyVil0Sk6Pjo+/nTwaY7sz0qsxAXVrCrwJees1JJlOKM5qbgBXUGf0b44WQkUQ07QdIp2SLnFkKweliEre2ll47mNJXUBl0r+oKlhUvRjuKJyo0i5Zel9q5G8F34M+MToM1V9sAe3l6xGqVtfSGEIE1StRhP008a64fKRznr+7c77DifWqiBlF9wsSmUlXn1ITTRYp1TQW5nRe5b+A2u7bZccZ5UiTfYHURZppY0sgISCGlU/LED83bpE6ImmH9GmESwM2dCECVkZBKFAeppm58Yo9lgZqhFpPycQrfmICuvU0uHWR0hIVRA+9tA//b/WSyqLEmyze6mZjdGn1GkcobKR3F5AgJ0z3O3PI4t1T72n02mxPElfinUhVbldsoKi307RIR5Hsb3xfmnhnJXlFX1w0Scd/gQADsZjvKtf5n2Y/QwBx8JETbkHdtmAzVemQ3qlXNP9ANf96Fri/5fwYm8hwC68mF5FTaXEUGkb1e/xLFFl6G6xJZmdqK4Bj0KvL2U3p4kXrGlK8wnKuSQGEW6YqTIa1jn/QlIj1WDDhW60s+s+2AtNPN641yIeixN7bBHNROLVcT2o5mgjhrRqu0mDj1+GDEoApusdGFXr+Az7YKHe1a+H9ImZD3BvU00CM6AWSJkE1+GjHndlabiHMyhw/0t8N8D+hPBzZfvQ1fWuMtAD6En7j27E3ozg6f4IBhvcm+heUs4z+r1PyhgkQ7UdnqNhL0Ot19IQEeAbrQMfotI3+C1cBonn5Fp4Stz/X+Sta+N+ESBWTsXGbJOBRMbojz41baBted8axflV61DcoIjGw1Ps75IwxfEldxFJEpcFZ5utYWLjLqdZ0LuJd/G3rzfXS3cBxufDku7MrLfjVCrBcii4ngWWeknPOu5MXdLW4TdvtYQ9/OXm7ty7c2UjvAK9CKPJeRLgPS3/GfnwoCXR/XntvVY4EfResQIK8mQLHN/vuAwCjUnmH0/kmTATfHv6cgUXOzf5ua5FmgRWFLC9GmE8hd8FKUpOP8OZNXVtp6dwpxq84Y/wYPH6bv+WwIW7a1Pzy4/4+vPl1+X18u9GNS7/rLfqKVzFT4qHxezNE8TxYR8+nhB40V1Rg8yWIkHgjJC5+3Y8qcoNv5GIZVQYltdOFt1btiaeVMkp5nKTdEIcYWCLbR2+o7rWtGAF+SHV+Y5pqziajE60qTk9C4X9kxWlVMZynAC7hhbueNXTxwoeYwanWofJOZnGpicZe4IMTgfejSjlRGuQ5BXnCzjMVqOzkynovzGFIyGD0bmDyeCktmrbw7Mbv4kxPpnCctjSSMkfidpD/hc)

please tell me what I need to do to validate this pull request

mickae1:
@mzschwartz5 , the idea is that for my model loaded from glb, i want to do the projection change in vertex shader when i'm in 2D . unfortunetly the x is reversed, also the y and i don't know why i need to return : 

 return vec3(altitude * flFactor, plate_x, plate_y);
 
 instead of  return vec3(plate_x, plate_y, altitude * flFactor);

mickae1:
I've made modifications, to start a new approach, more generic to every one, please give advice .

the updated [sandcastle ](http://localhost:8080/Apps/Sandcastle/#c=dVNha9swEP0rhz95EOSOjRRSN7S0JRTWLaxlMDAMRT4n6mTJSOck7uh/nyS3qdMmJh989949vXtyhNGOALml1U9eytbBOYy/jMenJ6dnEJ5CF/oKnWxrdqOUbJyRJSux4q0iz9W4gfdwWmgYSo4gy2ALNhaHsO4wBgF72mGfzoIXEQ031jyiIGn0vocZmqXlzUqK+Y6RHrM/FFxL3KDdF/sVe2mRiFhfGU1carRFMoJ/wetSmQVO9s4PneMHjsJUzZs3c5PBKoV+7i29uGFOoMY7U+Jcir++Du1QKla2lr9sfzKMRXGNcytrSXKNHhwKseYVcIyX/S3xDZf0ar6XrqypZ4qqS9dpkcY9AVqrJlAkjGX+d8/rRuE1J57VYcRlhI7YUi18Mj0/9u84Wbmd7ORj+ZXdXt98f7h9+B2YzyGSfml/1zMkoBWC5jWCqeJ7XMy2gowFMiBL1CSrLmKDr4C6BvsUjEKmzDJ9A9lAgwXtcOAR6h+Htaz5o7GXW+kCMRkluaNO4bRfDeBC1o2xFEJJfRyEPg7uI8gWrb8mYsLFwUDNs+FoXsq13+D8wBcFQnHnPFK1St3LJyySaZ55/odRZfz/QS9/rNEq3gXa6vP0W99kjOWZLw9PkjFqwe075f8)

packages/engine/Source/Shaders/Model/GeometryStageVS.glsl
```diff
@@ -1,18 +1,100 @@
...
+	float lat = cartographic.x;
+	float altitude = cartographic.z;
+
+	float plate_x = u_ellipsoidRadii.x * lat;
+	float plate_y = u_ellipsoidRadii.x * lon;
+
+	return vec3(altitude, plate_x, plate_y);
+}
+
+vec3 cartesianToCartographic(vec3 cartesian) {
```
mzschwartz5, line 13: I'm still processing the math here, but wondering if something [here](https://github.com/CesiumGS/cesium/blob/c8ae7b0133236022ca0c3bc0203df37f94e73b9f/packages/engine/Source/Shaders/Voxels/convertUvToEllipsoid.glsl#L23-L27) might be helpful (or if we can even just use these functions directly). It looks conceptually similar.

If we can use the above, I wonder if `isSphere` is actually a fast path, given the expense of trig functions on the GPU. Might be worth doing some performance measurements there to see. Maybe it's faster to just treat everything as an ellipsoid and use that non-trig iterative method.

packages/engine/Source/Shaders/Model/GeometryStageVS.glsl
```diff
@@ -1,18 +1,100 @@
...
     v_positionMC = positionMC;
-    v_positionEC = (modelView * vec4(positionMC, 1.0)).xyz;
 
     #if defined(USE_2D_POSITIONS) || defined(USE_2D_INSTANCING)
     vec3 position2D = attributes.position2D;
     vec3 positionEC = (u_modelView2D * vec4(position2D, 1.0)).xyz;
     computedPosition = czm_projection * vec4(positionEC, 1.0);
     #else
-    computedPosition = czm_projection * vec4(v_positionEC, 1.0);
+	if(czm_morphTime != 1.){
```
mzschwartz5, line 88: Small nit- given that, the vast majority of the time, we operate in 3D mode, it would be slightly faster to invert this check so that it short-circuits more often.

packages/engine/Source/Shaders/Model/GeometryStageVS.glsl
```diff
@@ -1,18 +1,100 @@
...
+		vec3 cartographic = cartesianToCartographic(positionMC);
+		vec3 position = projectToGeographic(cartographic);
+
+		vec3 positionEC = (czm_view * vec4(position, 1.0)).xyz;
```
mzschwartz5, line 92: Do we still need to write out to `v_positionEC` in this branch of the if-statement?

packages/engine/Source/Shaders/Model/GeometryStageVS.glsl
```diff
@@ -1,18 +1,100 @@
...
-    v_positionEC = (modelView * vec4(positionMC, 1.0)).xyz;
 
     #if defined(USE_2D_POSITIONS) || defined(USE_2D_INSTANCING)
     vec3 position2D = attributes.position2D;
     vec3 positionEC = (u_modelView2D * vec4(position2D, 1.0)).xyz;
     computedPosition = czm_projection * vec4(positionEC, 1.0);
     #else
-    computedPosition = czm_projection * vec4(v_positionEC, 1.0);
+	if(czm_morphTime != 1.){
+		vec3 cartographic = cartesianToCartographic(positionMC);
```
mzschwartz5, line 89: (This probably reveals some misunderstanding I have but) - why does this operate on the model space position instead of world or eye space?

mzschwartz5:
Overall comment - I have yet to test the changes, will do soon. Just did an overview for the moment. I think this could be a really nice contribution - if it works as promised without any regressions, we should go a step further and remove the CPU implementation of the projection. This would be the default, and we could skip the added options to the geometry pipeline stage to override the implementation.

packages/engine/Source/Shaders/Model/GeometryStageVS.glsl
mzschwartz5: (Looking back at this, I see that this is within the `USE_2D_POSITIONS` macro, so is it already always 2D mode? If we remove the CPU-side projection implementation, per my overall-review comment, *then* we would remove this macro too, I think, and we'd need the `czm_morphTime` check you added.)

packages/engine/Source/Shaders/Model/GeometryStageVS.glsl
mickae1: i don't know, If it is used in other pipeline, or used in fragment shader.

mzschwartz5: The prefix `v_` typically means it's a vertex output attribute - so it would get used in the fragment shader. In this case, you can see that [here](https://github.com/CesiumGS/cesium/blob/9755e91830b8ccd8f11276b2997607a3ccb564e5/packages/engine/Source/Shaders/Model/GeometryStageFS.glsl#L4).

mickae1:
@mzschwartz5 what i'm doing in the vertex shader, is exactly what is done in the cpu : 

https://github.com/CesiumGS/cesium/blob/4da5e56230a9038f9231440a4b154313252d03e2/packages/engine/Source/Core/GeometryPipeline.js#L616

with : 

```
 const lonLat = ellipsoid.cartesianToCartographic(
      value,
      scratchProjectTo2DCartographic,
    );
    if (!defined(lonLat)) {
      throw new DeveloperError(
        `Could not project point (${value.x}, ${value.y}, ${value.z}) to 2D.`,
      );
    }
    //>>includeEnd('debug');

    const projectedLonLat = projection.project(
      lonLat,
      scratchProjectTo2DCartesian3,
    );
```


I found out why, it need to be vec3(z,x,y ) to be working :  ( and not vec3(x,y,z) ) : 

because it's what you are doing in other shader : 

https://github.com/CesiumGS/cesium/blob/4da5e56230a9038f9231440a4b154313252d03e2/packages/engine/Source/Shaders/PolylineVS.glsl#L47

but not here, why ?
https://github.com/CesiumGS/cesium/blob/4da5e56230a9038f9231440a4b154313252d03e2/packages/engine/Source/Shaders/Model/GeometryStageVS.glsl#L12

mickae1:
I found out why I had to modify the geometryStageVS to use the matrix czm_view instead of czm_modelview :  

https://github.com/CesiumGS/cesium/blob/4da5e56230a9038f9231440a4b154313252d03e2/packages/engine/Source/Scene/Model/ModelSceneGraph.js#L340

In this function we are computing a new matrix to put the model on the 2D surface, but we shouldn't when we detect there is no translation in the matrix. Matrix without translation is surely an Identity matrix and we should let the programmer do the job in the vertex shader. In my case I'm rendering polylines, custom surfaces with a glb file. 

We should allow the user to override the geometry stage.

mickae1:
i'm closing this pull request

Issue #12410, opened 2025-01-03
Disable depth test against a specific asset for Primitives


jjspace:
This primarily stems from an observation I've made of the behavior of billboards at the horizon. Ideally billboards usually should be drawn "on top" of everything else _until they are on the opposite side of the globe_.

By default in Cesium this happens "for free" out of the box as can be observed in the [Billboards sandcastle](https://sandcastle.cesium.com/?src=Billboards.html)

When it's near the horizon but on "this side"
![2025-01-03_13-35](https://github.com/user-attachments/assets/67548eb3-5867-4816-ae3e-3389ef5d8120)
When the globe is turned slightly so the billboard is on "the other side" of the globe it gets clipped off
![2025-01-03_13-35_1](https://github.com/user-attachments/assets/ba215af0-e7a6-407f-b191-eaefeebab90c)

However, when using a global 3d tiles tileset like the Google photogrammetry tiles this behavior is _not_ the same. The billboard is _always_ drawn "under" the tileset
"In front"
![2025-01-03_13-42](https://github.com/user-attachments/assets/7e651197-1f70-48a9-866f-c6b76906bbc6)
"Behind"
![2025-01-03_13-43](https://github.com/user-attachments/assets/c5f50ed1-45ce-41e4-a6ff-1aeeb6dcb1b6)

The behavior of the 3dtiles can be replicated with terrain if you set `globe.depthTestAgainstTerrain` to `true`.
![2025-01-03_13-44](https://github.com/user-attachments/assets/6f697e48-4330-4727-b204-d28a0d5def57)

I think the default behavior of terrain is normally the _expected_ result out of the box (given issues like https://github.com/CesiumGS/cesium/issues/12396) however this requires some new architecture to add a way to ignore depth tests against a specific asset or tileset like we do with the terrain.

### Workaround

The best solution right now is to control the value of `disableDetphTestDistance` on the billboard itself. Setting this to `Number.POSITIVE_INFINITY` "works" to always show a billboard on top of 3d tiles as terrain but it also allows you to see billboards on the other side of the globe.

[...]

PR #12828, opened 2025-08-22
Adds Mars ellipsoid and trackable rovers

mzschwartz5:
# Description
Sandcastle demo for Mars terrain. Showcases geological points of interest, and the ability to see a few rovers take their journeys from landing point to current point  - defined in a CZML file, data and models from NASA.

## Issue number and link
N/A

## Testing plan
Make sure to include the CesiumGS default Ion access token when testing.

1. Review the [new sandcastle example](http://localhost:8080/Apps/Sandcastle/index.html?src=Mars.html), [CI-build link here](https://ci-builds.cesium.com/cesium/mars/Apps/Sandcastle/index.html?src=Mars.html)
2. Make sure the title, description, and thumbnail are all correct.

# Author checklist

### DIFF
--- a/ThirdParty.json
+++ b/ThirdParty.json
@@ -12,7 +12,7 @@
     "license": [
       "BSD-3-Clause"
     ],
-    "version": "2.7.71",
+    "version": "2.7.73",
     "url": "https://www.npmjs.com/package/@zip.js/zip.js"
   },
   {
@@ -109,7 +109,7 @@
     "license": [
       "MIT"
     ],
-    "version": "1.0.1",
+    "version": "1.1.0",
     "url": "https://www.npmjs.com/package/ktx-parse"
   },
   {
@@ -133,7 +133,7 @@
     "license": [
       "MIT"
     ],
-    "version": "0.24.0",
+    "version": "0.25.0",
     "url": "https://www.npmjs.com/package/meshoptimizer"
   },
   {
@@ -158,7 +158,7 @@
     "license": [
       "BSD-3-Clause"
     ],
-    "version": "7.5.3",
+    "version": "7.5.4",
     "url": "https://www.npmjs.com/package/protobufjs"
   },
   {
--- a/packages/engine/package.json
+++ b/packages/engine/package.json
@@ -47,7 +47,7 @@
     "ktx-parse": "^1.0.0",
     "lerc": "^2.0.0",
     "mersenne-twister": "^1.1.0",
-    "meshoptimizer": "^0.24.0",
+    "meshoptimizer": "^0.25.0",
     "pako": "^2.0.4",
     "protobufjs": "^7.1.0",
     "rbush": "^4.0.1",


LisaBosCesium:
@mzschwartz5 please share the sandcastle via https://sandcastle.cesium.com/ so reviewers without a local JS setup can view it.

javagl:
^This happened in the meantime. More generally: Once there is that ✔️ for the build process, the sandcastle on a branch with the name `my-example-branch` is available at
`https://ci-builds.cesium.com/cesium/my-example-branch/Apps/Sandcastle/index.html`
(i.e. https://ci-builds.cesium.com/cesium/mars/Apps/Sandcastle/index.html here)

(I also _tried_ to remember adding this link in the first post when applicable)

ggetz:
> (I also tried to remember adding this link in the first post when applicable)

Potential for automation via GitHub Actions?

packages/engine/Source/Core/Ellipsoid.js
```diff
@@ -254,6 +254,8 @@ Ellipsoid.MOON = Object.freeze(
   ),
 );
 
+Ellipsoid.MARS = Object.freeze(new Ellipsoid(3396190.0, 3396190.0, 3371726.0));
```
ggetz: Just a reminder before this PR is merged: This property needs inline documentation, and the addition should be documented in `CHANGES.md`.

mzschwartz5:
> This is looking great so far!
> 
> I love that we're using models for both the rovers. Maybe consider using [minimumPixelSize](https://cesium.com/learn/ion-sdk/ref-doc/ModelGraphics.html?classFilter=model#minimumPixelSize) so that they are visible further away?

Currently using a minimum pixel size of 64 (defined in the `.czml` file) but I can increase it! I think I've had general feedback about viewing the rovers from further out / at a scale that better matches the terrain resolution.

edit: I think it's actually the `maximumScale` that's the issue here. Set it to 1000, maybe need to increase.

LisaBosCesium:
@mzschwartz5 here are my notes from our call:

- Display instructions for rovers on the screen - eg double click, play animation controls to see rover movement 
- Adjust placement of labels on rovers - either hide when close to rover or move higher so doesn't block the rover
- Adjust lighting/apearance and raise atmosphere and provide controls for the user to adjust 
- Include links to the overlay data sources (in the metadata?)
- Nice to have - filter out the metadata fields populated only for the sake of the sandcastle 
- If needed add more comments to ensure devs understand the why of the code

mzschwartz5:
@LisaBosCesium Should all be addressed now (pushing changes momentarily). On the point of adjusting lighting, here are some before and after screen shots:

Globe view (before):
Screenshot 2025-08-28 at 8 48 38 PM

Globe view (after):
Screenshot 2025-08-28 at 8 48 31 PM

Feature view (before):
Screenshot 2025-08-28 at 8 48 50 PM

Feature view (after):
Screenshot 2025-08-28 at 8 49 06 PM

LisaBosCesium:
@mzschwartz5 -

Nice! 
- The descriptions look great and I see you filtered out the info you used to position to camera, also good. 
- Adding green target when zoom to a feature or rover is a nice change. 
- Lighting looks really good.
- Thank you to @dkbraig for reviewing the feature locations and catching some issues. You might already know about some comments that follow. 

Requests:
- What controls the order of the features on the drop down? Can we alphabetize if not a meaningful order I'm missing?
- Labels and target don't appear when zoom to some features - Alba Mons, Elysium. Guessing these are the wrong location ones you already know about. 
- Also some issues with labels showing up for Perseverance and Curiosity. Maybe also a coordinates issue?
- The Schiaperelli Crater appears to not be anchored to wherever it should be. 
- Can we add the Jezero Crater? Meant to ask this before, sorry. Is one that's mentioned a lot and for which we have high res data. 
- Please add the "Data attribution" link 
- We discussed adding an instructions text box on the page. I still think would be useful. Something like:
   Use the "Fly to rover" and "Fly to landmark" drop downs to move the camera to the selected location. 
   After flying to a rover, doubleclick on the rover to zoom to a closer view of the model.
   After flying to a rover, click the play button on the lower lefthand control for an animation of the path traveled by the rover.
   Click the home icon in the upper right to return to the original view.

mzschwartz5:
@LisaBosCesium 

> What controls the order of the features on the drop down? Can we alphabetize if not a meaningful order I'm missing?

It's just the order I added them in, easy to rearrange. I can put them in alphabetical order.

> Labels and target don't appear when zoom to some features - Alba Mons, Elysium.

Spin off of the same problem, yes - I'll fix it!

> Also some issues with labels showing up for Perseverance and Curiosity. Maybe also a coordinates issue?

This is sort of intentional - we discussed having the labels for those disappear at close distances so that they wouldn't overlap the rover models. I can adjust the interval so they appear a little further out.

> Can we add the Jezero Crater...

This is the Perseverance landing site. We have a landmark there by that name, as well as the ability to fly to the Perseverance rover. Would you prefer me to rename the landmark to Jezero Crater?

> We discussed adding an instructions text box on the page...

Something beyond this? I made it so that when you fly to a rover, the nav help menu opens up - and I added two items to it for tracking and playing animation.

image

LisaBosCesium:
@mzschwartz5 - 

Re Jezero/Perseverance - Understood, but just as we label both Gale and Curiosity, we should label Jezero as well as Perseverance because some people will be looking for Jezero.

Re instructions - I'm so used to those default instructions they are literally invisible to me - didn't notice you'd added to them. OK, let's just keep this as is. 

>> Also some issues with labels showing up for Perseverance and Curiosity. Maybe also a coordinates issue?

> This is sort of intentional - we discussed having the labels for those disappear at close distances so that they wouldn't overlap the rover models. I can adjust the interval so they appear a little further out.

The behavior felt inconsistent when I was testing, but I don't think it's worth messing with more unless you've already made a change that feels like an improvement to you.

mzschwartz5:
> I'm so used to those default instructions they are literally invisible to me

😂 I totally get it

Apps/Sandcastle/gallery/Mars.jpg
jjspace, line 1: Please shrink this waaaay down to match the other thumbnails. Should be ~250px in width

packages/engine/Source/Core/Ellipsoid.js
```diff
@@ -254,6 +254,14 @@ Ellipsoid.MOON = Object.freeze(
   ),
 );
 
+/**
+ * An Ellipsoid instance initialized to a sphere with the mean radii of Mars.
```
jjspace, line 258: Should we say where we sourced our value from? For the earth WGS84 is well known and clear. Just trying to get ahead of someone asking

Apps/Sandcastle/gallery/Mars.html
```diff
@@ -0,0 +1,560 @@
...
+            "../../SampleData/MarsPointsOfInterest.geojson",
+          );
+          viewer.dataSources.add(dataSource);
+
+          const onSelectLandmark = (landmark) => {
+            reset();
+            scene.camera.flyTo(landmark);
+          };
+
+          const toolbarMenuEntries = [
```
jjspace: If we're keeping 2 lists this should be renamed to make it clear in the code it's a different list than the first time this was declared.

jjspace (commented):
Overall the code seems fine. This is definitely a very lengthy sandcastle but I understand it's one that's more oriented at a "sales" perspective so that's ok.

Had a number of comments just using the sandcastle itself. Some are definitely optional, others I feel not so but feel free to push back on if they've already been discussed. Largely just ideas/suggestions

- Marco mentioned this but at certain angles the camera "bounces" off the terrain much higher than I would expect.
- (optional) Can we make the camera/animation tips only show up the first time? I don't need to see them every time I switch rovers
- The highlight animation is cut off by the element bounds
    - [image]
- The clock keeps playing after leaving a rover, should it?
    - Also why not just start the clock for people when zooming to a rover? we're already resetting the time to the start
- Shadows don't turn off correctly when switching the rover "mid-flight"
    - Counter point, why even enable shadows at all? Can we just turn them off completely?
    - Repro:
      - Run sandcastle
      - Click Play on animation
      - Pick Curiosity
      - Pick Perserverance before getting to Curiosity
      - See shadows
- "Spin on first load" is not working. This is because you're calling `reset` in the first menu item. That handler will always be called by default when first setting it up
- The "Landing site" POIs overlap with the nearby crater names. Can we remove them? Or make them only appear much closer? Or maybe combine them like "Gale crater (Perserverance)"?
- The order of POIs may not be the "best"? I realize they're alphabetical but I was first exploring them "top to bottom" and certain collections don't make sense (optional if it's already been decided)
    - For example There's a POI between "Curiosity Landing Site" and "Gale crater". Going "top to bottom" makes the camera jump back and forth to the same spot even though it's right next to it. Same idea around the Perserverance
    - It'd be nice to follow the POIs along the Martian Journey "in order"
- (optional) It might help a lot to use [`optgroup`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/optgroup) elements to group things like "Craters" or "Rovers" or "Martian Journey stops"
- Should all the positions be in a _single_ menu list? (easier if using optgroups)
    - With Menus you can't "re-select" the current selection. if you select a rover to see the path, then go to a POI, you can't jump back to the same rover. If they were one list it'd be easier to do that.
    - The menu doesn't reset or change once leaving a POI. I don't think it needs to but it's a bit weird to zoom to a POI but still see it listing the rover you were just at
- Did we look into keeping the camera facing somewhat down when using `flyTo`? It's a little odd to just stare at stars for some flights. If not possible then it's fine
- (optional) is there _any_ way to nicely cut out sections of the rover's path? It's not very interesting to see it just shake in one spot for may Sols because it didn't move much.
    - Curiosity barely moves until Sol 300. Can we just start it further in their path?
    - Perserverance also has a very long section it's "stationary" around the 250-300 sol mark. Honestly "max" speed of 604800x
or whatever seems better than the current `100_000`

mzschwartz5:
> This is definitely a very lengthy sandcastle

Yes... definitely pulled as much as I could out into data, but it gets lengthier every time I revise upon feedback :) 
Considering that, here's a list of the items you called out that I think definitely should be addressed, and are changes or fixes rather than additions:

- "Spin on first load" is not working
- Make camera face somewhat down when using flyTo - I vaguely recall reading somewhere that there's an option for this. I'm pretty sure.
- The "Landing site" POIs overlap with the nearby crater names. - this has undergone some revision already, but I think this is OK to change. I agree it's not ideal.
- Rover changes: starting off with a higher multiplier, potentially starting off mid path at more interesting points.
- Shadow bug. (I don't think we should disable shadows entirely)

I think some of the other changes may be more trouble than they're worth / lengthen the sandcastle more (e.g. cutting out sections of rover data, first-time-only camera tips, optgroups and other menu shenanigans) or have already been discussed in other peoples' feedback (e.g. we do want the POIs in alphabetical order).

But on the whole, many of these are easy adjustments that will improve the experience.

ggetz:
@mzschwartz5 What's the state here? Is this ready for a final pass and merge?

mzschwartz5:
Yes it is- sorry, messaged @jjspace offline but didn't update the PR.

jjspace (commented):
Thanks for the updates @mzschwartz5, this is looking really good. Especially the new flyto camera that points down; really helps keep your bearings between landmarks.

Two remaining comments

- First please update the `CHANGES.md` with the new ellipsoid as @ggetz mentioned.
- Second, I'm noticing slight flickering of the labels at global scale. Not a huge deal but thought I'd throw it out there. Maybe related to the #12410 work? It also doesn't always happen, sometimes when I run the sandcastle the default camera starts more zoomed out and they don't flicker then. See the video:

https://github.com/user-attachments/assets/a7404697-6a7a-4804-b6f2-358c0bc097eb

mzschwartz5:
I'm not sure what to make of that, honestly. I can't reproduce it so maybe it's OS specific? I notice other things look different in that vid as well, like the font. 

Will add to `CHANGES.md` when I get a moment (and, reminder to self- also need to add the data attribution)

LisaBosCesium:
@mzschwartz5 Just wanted to chime in that the final changes look great and will be super helpful for users. Thank you!

CHANGES.md
```diff
@@ -21,6 +21,7 @@
 
 - Added support for the [EXT_mesh_primitive_restart](https://github.com/KhronosGroup/glTF/pull/2478) glTF extension. [#12764](https://github.com/CesiumGS/cesium/issues/12764)
 - Added spherical harmonics support for Gaussian splats. Currently supported with the SPZ compression format.
+- Added `Ellipsoid.MARS` for use with Mars terrain and imagery. [#12828](https://github.com/CesiumGS/cesium/pull/12828)
```
ggetz, line 24: Just a heads up that I tweaked this—The changelog applies to the Source code itself, not (typically) to Sandcastle. So, we usually omit calling out new examples. On the other hand, the code change `Ellipsoid.MARS` needs to be called out specifically.

ggetz:
Looks fantastic! Thanks @mzschwartz5!

ggetz (approved):

PR #12832, opened 2025-08-22
Sandcastle popovers

jjspace:
# Description

Adds a customized [Ariakit popover](https://ariakit.org/components/popover) for Sandcastle with styles to match stratakit.

- Adds a metadata popover to display the title and description of a sandcastle when clicking the title
- Adds back the Share popover that displays the share URL so it no longer just updates the URL bar

This PR is based on the work in #12819 and should be merged after that PR.

## Issue number and link

Part of #12566 

## Testing plan

- Run `npm run dev` from the sandcastle package or `npm run build-sandcastle` and `npm start` from the project root
- Open any page and click the title next to the logo and make sure it displays the sandcastle info
    - If not loading a specific gallery demo it should display "New sandcastle" and a generic message
- Click the Share button and test the input
    - It should always select the full text when it opens
    - Clicking the copy button should work
    - Make sure changes to the code are actually reflected in the generated share URL

# Author checklist


lukemckinstry:
It would be nice to have some UI/UX feedback/sugar when I click to copy the share URL. 
For example GitHub shows a green checkmark when I click to copy a branch name or commit hash. 

Basic checkmark (this is probably sufficient)
https://github.com/user-attachments/assets/9743decb-df66-40ed-a0c5-9535580eb98f

Checkmark with "copied!" message depending on where you are on the page 
https://github.com/user-attachments/assets/f6c7ef92-458e-400e-80ea-00a2757f6073

ggetz:
@jjspace Could you please merge in main when you get the chance?

packages/sandcastle/src/SandcastlePopover.css
```diff
@@ -0,0 +1,45 @@
...
+  --popover-background-color: var(--stratakit-color-bg-elevation-base);
+  /* This should match the --stratakit-shadow-dialog-base value but with the inset part
+   * removed to avoid a border over the arrow. A border color is added instead */
+  --popover-shadow:
+    0px 1px 0px 0px oklch(0% none none / 0.08),
+    0px 0px 0px 1px oklch(0% none none / 0.8),
+    0px 1px 1px -0.5px oklch(0% none none / 0.16),
+    0px 3px 3px -1.5px oklch(0% none none / 0.16),
+    0px 6px 6px -3px oklch(0% none none / 0.16);
+}
```
ggetz: I think the appropriate token to use here is `--stratakit-shadow-dialog-base`. I noticed this because the contrast was really low (at least on my monitor) and I had trouble differentiating the boundary of the popover.

I pushed one commit with some changes to enhance the contrast via some css variables which can be overridden on a case-by-case basis.

In that same commit, I implemented @lukemckinstry's suggestions while I was there.

I also added a bit more context via the existing `title` prop. Finally, I made a few tweaks to be consistent with ion's UI (which I also saw in the GitHub UI as well) including using a text area with strata's monospaced font and moving the description below the textarea.

Screenshots of the updates and references are all below. Let me know if you hate it and I can revert.

Image

Image

Image

Image

ggetz:
Thanks @jjspace, just that one comment from me. Let me know what you think. If you're OK with it, this PR is good to merge.

packages/sandcastle/src/SandcastlePopover.css
jjspace: > I think the appropriate token to use here is `--stratakit-shadow-dialog-base`

I flipped back and forth with these but felt the `dialog-base` shadow was much too aggressive and placing the popover "higher" above the page than it should be which is why I chose the `surface-md` shadow instead. I felt this still gave a separation but wasn't over the top. Maybe we can move to the `surface-lg` one but I didn't feel the `dialog-base` worked for this. It does work for actual dialogs like the settings modal.

Second I don't think these variables should be defined on the `:root`. They are not used outside of this component so they should be isolated to it. If we want to start using them elsewhere we can consider elevating them as needed.

> I also added a bit more context via the existing title prop

Given the current state of sharing I don't think this is useful. When loading the share url it will _always_ say "New Sandcastle" for the title because that information is not stored in the shared data. This could create confusion because "it said I was sharing '3D Models' but when I load the page the title doesn't match". If/When we add the ability for people to customize the title and description and include that with the "saved" (shared) sandcastle then I think it could be helpful to include here.

jjspace: I've updated the pr to adjust some of these changes per my comment above

ggetz: > I felt this still gave a separation but wasn't over the top. Maybe we can move to the surface-lg one but I didn't feel the dialog-base worked for this. It does work for actual dialogs like the settings modal.

Can we compromise on `--stratakit-shadow-dropdown-base`? That's what the `DropdownMenu` stratakit component uses, and the elevation level there is the most comparable IMO.

ggetz:
Thanks @jjspace. Just that one response from me, and then I think this is good to go.

jjspace:
@ggetz updated, hopefully good to go now! :pray:

ggetz:
Thanks!

PR #12839, opened 2025-08-26
Adds git-blame-ignore-revs file

mzschwartz5:
# Description

Added a `git-blame-ignore-revs` file. This allows blame to skip certain pervasive changes. I included two major offenders to start: the commit that ran prettier on the whole codebase, and the commit that changed var/let to const.

To use the file:
`git config blame.ignoreRevsFile .git-blame-ignore-revs`

If using gitlens in vscode, you can add this setting:
```
    "gitlens.advanced.blame.customArguments": [
        "--ignore-revs-file",
        "${workspaceFolder}/.git-blame-ignore-revs"
    ],
```

## Testing plan

# Author checklist

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -6,6 +6,7 @@
         "cesium.gltf-vscode",
         "bierner.github-markdown-preview",
         "DavidAnson.vscode-markdownlint",
-        "streetsidesoftware.code-spell-checker"
+        "streetsidesoftware.code-spell-checker",
+        "eamodio.gitlens"
     ]
 }
--- a/Documentation/Contributors/BuildGuide/README.md
+++ b/Documentation/Contributors/BuildGuide/README.md
@@ -39,6 +39,7 @@
   - Double-check your settings for name and email: `git config --get-regexp user.*`.
   - Recommended Git settings:
     - `git config --global fetch.prune true` - when fetching remote changes, remove any remote branches that no longer exist on the remote.
+    - `git config blame.ignoreRevsFile .git-blame-ignore-revs` - uses the ignore file to skip certain noisy revisions (like formatting) when running git blame. Alternatively, for VSCode users, install the GitLens extension, which will automatically use the ignore file.
 - Have [commit access](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CommittersGuide/README.md) to CesiumJS?
   - No
     - Fork [cesium](https://github.com/CesiumGS/cesium).


jjspace:
Also I'm still seeing the prettier commit in gitlens even with the setting on. I added the prettier v3 commit to test and that one _is_ getting hidden. Any ideas?

This is in the file `engine/Source/Core/ApproximateTerrainHeights.js`
image

mzschwartz5:
@jjspace 
1 + 2. Can do.
3. It doesn't break gitlen's blame if the file exists but you don't use it. Personally, I think if we're going to commit the file, we should include the setting in our workspace settings. Can't see any reason not to use it, so why make people set it up themselves.
4. It appears it is necessary if you use `git blame` from the command line, rather than gitlens. I think we could potentially add the setting to a new `.gitconfig` repo file, but then users would have to run a command to use that config locally. So it's a question of running a git blame config command once, or running a git config command once.

jjspace:
> Can't see any reason not to use it, so why make people set it up themselves.

I was looking into this a bit more and per [this comment](https://github.com/gitkraken/vscode-gitlens/issues/947#issuecomment-583703079) it seems gitlens will respect the git setting itself and no workspace setting is required. I tested it and it seems to work as expected, I say we leave it out for now and focus on the gitconfig itself.

> I think we could potentially add the setting to a new .gitconfig repo file, but then users would have to run a command to use that config locally. So it's a question of running a git blame config command once, or running a git config command once.

Regardless it should probably be documented in the developer setup guide. If we're already going to have people run a command it might make more sense to opt for a generic one. Then we could potentially add more settings to it later and everyone will already be set up for it to work. I like the include statement like this SO answer points out: https://stackoverflow.com/a/18330114/7416863
(cc @ggetz any opinions on this?)

mzschwartz5:
1. I think the reason some blames are still showing up in the above picture is because Prettier created those new lines (new whitespace, or moving brackets / parentheses to new lines). There's no blame before that for those lines, so it shows the Prettier commit.
2. I'm still leaning towards including the gitlens settings, because it *does* apply automatically, whereas the git config does not. I'd still want both settings + documentation for the latter. edit - or just forget the new `.gitconfig` and just document the option to set it yourself.

jjspace:
> There's no blame before that for those lines, so it shows the Prettier commit.

I didn't even think of that but I think you're totally right, looks good then.

Also pretty cool that github will respect this too!

image

ggetz:
> Regardless it should probably be documented in the developer setup guide. If we're already going to have people run a command it might make more sense to opt for a generic one. Then we could potentially add more settings to it later and everyone will already be set up for it to work. I like the include statement like this SO answer points out: https://stackoverflow.com/a/18330114/7416863
(cc @ggetz any opinions on this?)

No strong opinions on either option, but a :+1: for documenting in in the build/setup guide.

mzschwartz5:
@jjspace Added documentation on these changes - should be good to merge now.

Issue #12842, opened 2025-08-26
Remove support for deprecated 3D Gaussian Splatting extension


weegeekps:
In CesiumJS 1.133, the `KHR_spz_gaussian_splats_compression` extension was deprecated in favor of the latest extension specifications defined in `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2`. The old extension needs to be cleaned up for the CesiumJS 1.135 release slated for November 2025.

I have included a patch file with the necessary changes to streamline this task, but there may be additional parts from subsequent PRs that may need to be cleaned up.

## Patch file

// TODO: Once the PR in 1.133 is merged.

## Tasks

- [ ] Apply the patch file.
- [ ] Remove any remaining parts that may have been added after the patch file was created. (e.g. for Spherical Harmonics, updates, etc.)
- [ ] Ensure the unit tests still pass successfully.
- [ ] Notify Product and the Open Standards groups that support for the old extension has been removed. (@shehzan10 & @weegeekps)

PR #12843, opened 2025-08-26
Add support for latest 3DGS extensions and deprecate original extension

### LINKED ISSUE #12837: Updating to latest 3D Gaussian Extensions and deprecating initial experimental version
# Summary

We are updating Cesium JS to support the latest version of the glTF extensions for 3D Gaussian Splatting. Support for the extension version from initial release is now **deprecated and will be removed with the November 2025 release of CesiumJS**.

# Deprecation Plan

While the 3D Gaussian splatting feature is currently **experimental**, we want to give people time to retile their content with the new tiler that will be released at the same time as these changes to CesiumJS.

## September 2025 Release

Support for the version of the glTF extension, `KHR_spz_gaussian_splats_compression`, released in June 2025 is now deprecated, but users can still load 3D Tiles tilesets using this version of the extension.

The 3D Gaussian splatting tiler in ion has been updated and outputs 3D Tiles tilesets using the current version of the glTF extension specifications: `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2`.

The 3D Gaussian splatting support within this release is still experimental and there may be changes again in the future, but we are hopeful that future changes will be less impactful.

**We advise that during this time before the November 2025 release, you retile your data using the latest version of the tiler.**

## November 2025 Release

Support for the original `KHR_spz_gaussian_splats_compression` extension will be removed, and only 3D Tiles tilesets using the `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2` tilesets will be supported.

# The why

Since the initial release of the 3D Gaussian Splatting for 3D Tiles support in CesiumJS, there have been [significant updates to the glTF extensions that we need to implement](https://github.com/KhronosGroup/glTF/pull/2490). The changes include:

 - Splitting the extension from a singular extension, `KHR_spz_gaussian_splats_compression`, to two extensions: `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2`.
 - Adding [namespaces to the custom attribute semantics](https://github.com/KhronosGroup/glTF/issues/2111#issuecomment-3145203515) defined by the glTF extension.
 - Handling the [SPZ extension as a child of the base 3DGS extension](https://github.com/KhronosGroup/glTF/pull/2490#discussion_r2225728797).

weegeekps:
# Description

This PR deprecates the `KHR_spz_gaussian_splats_compression` extension and implements support for the `KHR_gaussian_splatting` and `KHR_gaussian_splatting_compression_spz_2` extensions. These are the latest extensions and much closer to a stable point than the original extension. We only support 3D Tiles that are using both extensions and compressing the data in SPZ, and it's unknown if and when we will be adding support for uncompressed 3DGS data.

Complete removal of the `KHR_spz_gaussian_splats_compression` extension will be performed in the 1.135 release. This work can be tracked in issue #12842.

I've also updated the unit tests and sandcastles.

## Issue number and link

Resolves #12837.

## Testing plan

I tested against sandcastles using both the old and new extensions and updated the unit tests.

# Author checklist


.idea/codeStyles/Project.xml
```diff
@@ -4,9 +4,6 @@
       <option name="SPACE_BEFORE_PROPERTY_COLON" value="true" />
       <option name="SPACE_BEFORE_FUNCTION_LEFT_PARENTH" value="false" />
     </JSCodeStyleSettings>
-    <XML>
-      <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
-    </XML>
```
weegeekps, line 9: Webstorm has been pestering me about this for a while, and it seems harmless to check in.

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2020,8 +2020,21 @@ function loadPrimitive(loader, gltfPrimitive, hasInstances, frameState) {
...
+  const gaussianSplattingExtension = extensions.KHR_gaussian_splatting;
+  const spzExtension = (() => {
+    // Yes, this is gross, but we expect the extension object downstream, not a boolean.
+    if (defined(gaussianSplattingExtension)) {
+      return gaussianSplattingExtension.extensions
+        .KHR_gaussian_splatting_compression_spz_2;
+    }
+
+    return undefined;
+  })();
```
weegeekps: Not entirely pleased with this, but it seemed like the safest and most readable way to handle this was using an IIFE. The only alternative I found was a rather nasty conditional chain. I'm open to ideas.

javagl: Is the "nasty chain" something like
`const spzExtension = extensions?.KHR_gaussian_splatting?.extensions?.KHR_gaussian_splatting_compression_spz_2;`?
(The last '?' is part of my question 😆 )

I tried to follow the path that this object takes, into the depths of what would be the first point in my "This should be refactored **NOW!!!**"-list, namely the _loaders_. It's late here. I've lost track. But I'm reasonably sure (and will probably confirm tomorrow) that all this goes 💥 when there is an input file where the data is _not_ stored in `bufferView` `0`. At least, that's how I interpret [this line](https://github.com/CesiumGS/cesium/blob/2ae42ea7fd244e9afd7b05c07a69a4702435def0/packages/engine/Source/Scene/GltfSpzLoader.js#L99). 

(I actually wanted to check where the `bufferView` of the extension object is finally accessed, and whether it's _really_ necessary to juggle around with that object (or its legacy object), and create specialized `...Spz...` classes, when all that this object contains is an _index_, but ... maybe it has to be like that for some reason, even iff this `bufferView` is not accessed _at all_ - not sure, shrugging it off for now...)

weegeekps: Yes, the loaders are overdue for a refactor.

The nasty chain is that but without the optional chaining operator. I didn't realize we can use the optional chaining operator in pure JS now. Looking at the browser compatibility charts, it looks like it's been supported since 2020 in all major browsers. I'm going to correct this using that operator.

I interpret that line you found the same way. That's definitely a mistake as we should be reading the buffer view value out of the extension. I opened issue #12847 to track this.

javagl: That optional chain is still not "pretty". There's always the option to turn [this code block](https://github.com/CesiumGS/cesium/blob/7a3b28a05396a0c4d091104a0f083e516ca995fc/packages/engine/Source/Scene/GltfLoader.js#L2023-L2040) into
```
const spz = fetchSpzExtensionFrom(gltfPrimitive);
if (defined(spz)) {
  needsPostProcessing = true;
  primitivePlan.needsGaussianSplats = true;
}
```
with a function like (drafted)
```javascript
/**
 * (5 lines of comment...)
 */
function fetchSpzExtensionFrom(meshPrimitive) {
  const extensions = meshPrimitive.extensions ?? {};
  const legacySpz = extensions.KHR_spz_gaussian_splats_compression;
  if (defined(legacySpz)) {
    console.log("Hey, that's gonna be deprecated soon, but... OK");
    return legacySpz;
  }
  const gaussianSplatting = extensions.KHR_gaussian_splatting;
  if (!defined(gaussianSplatting)) {
    return undefined;
  }
  const innerExtensions = gaussianSplatting.extensions ?? {}; 
  const spzSomewhatLegacy = innerExtensions.KHR_gaussian_splatting_compression_spz;
  if (defined(spzSomewhatLegacy)) {
    console.log("The _v2 is missing, but I'll allow it...");
    return spzSomewhatLegacy;
  }
  const spz = innerExtensions.KHR_gaussian_splatting_compression_spz_2;
  if (!defined(spz)) {
    console.log("Wo-ho... gaussian splatting without SPZ? No, not right now");
    return undefined;
  }
  return spz;
}
```
(I know, it would be a "style that is inconsistent to the other code", but ... ... ... ... well)

weegeekps: That's actually the kind of nasty stuff I was trying to avoid.

In retrospect though I'm wondering if it might be a bit safer approach. The downside is that it will potentially be an additional pain point when we implement support for uncompressed data, but I doubt we will support that anytime soon. Let me try it out.

weegeekps: Okay, I took what you suggested and derived something that's a bit cleaner. I like the use of a function because it means we'll always get a single SPZ extension or an undefined object. I removed all of the logging because we're broadcasting the deprecation warning elsewhere and we don't need to barrage users, and I didn't include support for the non `_2` version of the SPZ extension name because we don't support that extension here, and we were never outputting files using that setup.

javagl: I'm lacking some context in how far uncompressed data is supposed to be handled. And while it does reflect that "(nasty) chain" in more elaborate form, it could probably be shortened a bit if it omitted the log outputs, which (on the other hand) I think can be helpful (at least compared to the "fail-silent" pattern of `if (!something) return;`). 

_Some_ of this is pretty temporary, so maybe there doesn't have to go toooo much thought about how to handle the legacy extension for the next 2 months. 

(In my local experiments, I also created something that is [pretty preliminary](https://github.com/javagl/JSplat/blob/master/jsplat-io-spz-gltf/src/main/java/de/javagl/jsplat/io/spz/gltf/SpzGltfConfig.java#L30) in that regard, just to have ~"a central place where all the differences are handled"...)

weegeekps: > I'm lacking some context in how far uncompressed data is supposed to be handled.

Today, with our implementation in CesiumJS it is not handled. This is subject to change but it's probably best if we wait until after the base extension is ratified.

> Some of this is pretty temporary, so maybe there doesn't have to go toooo much thought about how to handle the legacy extension for the next 2 months.

That's kind of my stance on further changes at the moment. I don't see harm in using the optional operator here as long as it's readable and safe, and we can likely further simplify this once the deprecated extension is removed.

ggetz:
Thanks @weegeekps! Are there any dependencies between this PR and https://github.com/CesiumGS/cesium/pull/12790? Should we review or make sure we merge them in any particular order?

weegeekps:
@ggetz I'd actually recommend merging this one first and then the SH one to make sure that no updates have gotten missed in the latter.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -80,6 +81,43 @@ function GaussianSplat3DTileContent(loader, tileset, tile, resource) {
...
+  let hasLegacyGaussianSplatExtension = false;
+  if (tileset.isGltfExtensionRequired instanceof Function) {
+    hasGaussianSplatExtension =
+      tileset.isGltfExtensionRequired("KHR_gaussian_splatting") &&
+      tileset.isGltfExtensionRequired(
+        "KHR_gaussian_splatting_compression_spz_2",
+      );
+
+    // TODO: Remove support for deprecated experimental extension for CesiumJS 1.135 release.
+    //  See: https://github.com/CesiumGS/cesium/issues/12842
```
ggetz: Please remove these comments. [We track work in GitHub issues, not in `TODO` comments in the code](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#:~:text=TODO%20comments%20need%20to%20be%20removed%20or%20addressed%20before%20the%20code%20is%20merged%20into%20main.%20Used%20sparingly%2C%20PERFORMANCE_IDEA%2C%20can%20be%20handy%20later%20when%20profiling.).

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -80,6 +81,43 @@ function GaussianSplat3DTileContent(loader, tileset, tile, resource) {
   this._transformed = false;
 }
 
+/**
+ * Performs checks to ensure that the provided tileset has the Gaussian Splatting extensions.
+ *
+ * @param tileset The tileset to check for the extensions.
```
ggetz:
```suggestion
 * @param {Cesium3DTileset} tileset The tileset to check for the extensions.
```

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -80,6 +81,43 @@ function GaussianSplat3DTileContent(loader, tileset, tile, resource) {
...
+ * @returns {boolean} Returns <code>true</code> if the necessary extensions are included in the tileset.
+ * @static
+ */
+GaussianSplat3DTileContent.tilesetHasGaussianSplattingExt = function (tileset) {
```
ggetz: This may be nit-picky, but "has extension" != "requires extension", right? Should this be called `tilesetRequiresGaussianSplattingExt`?

packages/engine/Specs/Scene/GaussianSplatPrimitiveSpec.js
```diff
@@ -47,6 +47,102 @@ describe(
...
+        scene,
+        tileset.root,
+      );
+      expect(tile.content).toBeDefined();
+      expect(tileset.gaussianSplatPrimitive).toBeDefined();
+      expect(scene).notToRender([0, 0, 0, 255]);
+
+      tileset.show = false;
+      expect(scene).toRender([0, 0, 0, 255]);
+    });
```
ggetz: Since this test isn't doing anything at the moment, let's remove this `xit`'d spec from the _legacy_ describe block as it may go away sooner.

ggetz:
Tested and confirmed behavior in this branch with the new and deprecated extensions. All looking great 👍

weegeekps:
@ggetz I've addressed all of your feedback. Thank you!

ggetz:
Awesome looks great. Thanks @weegeekps!

PR #12852, opened 2025-08-28
Sandcastle gallery list—Default to Showcases label 

ggetz:
# Description

This is a quick implementation of setting the default search filter for the gallery list panel. We discussed offline setting this to "Showcases", like the old version of Sandcastle, to make the default list view more manageable. 

image

There are certainly other improvements we can make, but this should be one small step towards guiding users to useful information as opposed to giving them an avalanche of content.

The next step is to purge the metadata for the gallery examples themselves to make better use of labels—In particular, pare down the "Showcases" tag to only a handful of recent and high-value items.

## Issue number and link

N/A

## Testing plan

1. Open the gallery panel and ensure only items tagged with "Showcases" are shown
2. Play with the tag filter to ensure "All" and other filters still work as before 

# Author checklist

### DIFF
--- a/packages/sandcastle/src/Gallery.tsx
+++ b/packages/sandcastle/src/Gallery.tsx
@@ -168,14 +168,18 @@ export function GallerySearch({
             setCurrentTag(e.target.value);
             setTag(e.target.value);
           }}
-          value={defaultTag}
         >
+          <option value={defaultTag} selected>
+            {defaultTag}
+          </option>
           <option value={"All"}>All</option>
-          {Object.keys(availableFilters.tags).map((label) => (
-            <option value={label} key={label}>
-              {label}
-            </option>
-          ))}
+          {Object.keys(availableFilters.tags)
+            .filter((label) => label !== defaultTag)
+            .map((label) => (
+              <option value={label} key={label}>
+                {label}
+              </option>
+            ))}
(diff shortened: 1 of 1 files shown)


packages/sandcastle/src/Gallery.tsx
```diff
@@ -165,6 +168,7 @@ export function GallerySearch({
             setCurrentTag(e.target.value);
             setTag(e.target.value);
           }}
+          value={defaultTag}
         >
           <option value={"All"}>All</option>
           {Object.keys(availableFilters.tags).map((label) => (
```
jjspace: @ggetz Do you think it makes sense to change this slightly to put `Showcases` at the top of the options list?

Was pretty easy to filter it like this
```jsx

          <option value={"Showcases"}>Showcases</option>
          <option value={"All"}>All</option>
          {Object.keys(availableFilters.tags).map((label) => {
            if (label === "Showcases") {
              return;
            }
            return (
              <option value={label} key={label}>
                {label}
              </option>
            );
          })}
```

ggetz: Sure!

ggetz:
@jjspace Updated!

jjspace (approved):
LGTM, thanks @ggetz

PR #12854, opened 2025-08-29
Sandcastle search and filtering updates

ggetz:
# Description

image

image

image

image

- Added a `sandcastle.config.js` file for managing gallery and search-related properties which may change more frequently than the source code, and could differ across different deployments.
- `scripts/buildGallery.js` now:
  - Uses options defined in `sandcastle.config.js`.
  - Builds gallery files, pagefind output, and `list.json` **directly to the `public` folder**, rather than relying on vite to copy later. This keeps source and output files separate, preventing unintentional overwrites.
  - Via `scripts/createGalleryRecord.js`, creates an in-memory HTML file for each gallery item that pagefind can use for indexing. 
    - We adjust the default weighting schema such that `tile` >> `description` >> `code` in terms of search relevance using the `data-pagefind-weight` attribute. Hierarchical labels like "Title", "Description", and "Labels" are ignored completed.  
    - `data-pagefind-meta` lets us grab metadata properties. `data-pagefind-filter` let's us grab the metadata used for filtering. Lastly, `id` attributes are important to assign for enabling highlighting on the front end.
- Top-level styling was added for code blocks and for highlights using the `mark` semantic element.
   - Highlighting styles point to the top-level css variables `--highlight-color-*`, including background color, border color, and text color; hover and pressed variations. **Update `--highlight-color-*` variables in `src/App.css` to tweak the highlight colors.**
- `src/utils/loadFromUrl.js` is now in it's own file—It loads per the the url param as before, but now throws errors if requested content can not be found. Errors are handled and routed to in-app console to make it obvious to the user.
- Moves gallery-related components and store the `src/Gallery` directory:
  - **`src/Gallery/GalleryItemStore.ts` creates a shared context for use across gallery components.**
  - `src/Gallery/GalleryItemList.ts` and `src/Gallery/GalleryItemCard.ts` manages the filtered list of gallery items, implementing the ariakit `Composite*` abstract components. 
    - Search matches should be highlighted in titles, descriptions, and code blocks. Labels that match the current filter will be highlighted (active).
    - When there is code sample relevant to the current search term, it will also appears on the gallery card.
    - I added a "line count" display to give users a bit more context when navigating examples (akin to say video length on youtube). I'm also hoping it reminds _us_ to write more concise examples.
    - Tweaked the styling a bit to allow for more room for descriptions as well as the complete titles. These are essential for navigating the examples, and we're starting to provide more context in descriptions.
  - `src/Gallery/GalleryItemSearchInput.tsx` handles the search term query input. **The pagefind search is performant enough that, IMO, it does _not_ require the debounce, so I left it out for the sake of simplicity.** Let me know if there are strong opinions on this.
  - `src/Gallery/GalleryItemSearchFilter.tsx` handles filtering by labels with the `DropdownMenu` component so we can show the pagefind total results per-category.
     - Filters can also be updated by clicking the label tags themselves.
- Handle third-party type definitions in a standard and automated manner—`.d.ts` files moved moved to `@types` directory that is specified in the typescript config as a types root.

## Issue number and link

N/A

## Testing plan

1. Verify gallery item loading: 
  - Try different urls
  - Click gallery examples directly
  - Click the "open code" button in the top-right of cards
  - Navigate back and forward in your browser history
2. Try searching for various terms:
  - Ensure the "clear" button clears the search term
  - Ensure you can now match titles and descriptions in addition to code
3. Use the filters:
  - Make sure we default to "Showcases"
  - Use the 'labels' dropdown
  - Use the label tags in the gallery item cards

# Author checklist

- [x] Fix page title
- [x] Merge sandcastle-v2

### DIFF
--- a/packages/sandcastle/src/App.tsx
+++ b/packages/sandcastle/src/App.tsx
@@ -452,13 +452,6 @@ function App() {
         ) {
           // only push state if it's not the current url to prevent duplicated in history
           window.history.pushState({}, "", `${getBaseUrl()}?id=${id}`);
-          if (
-            !searchParams.has("id") ||
-            (searchParams.has("id") && searchParams.get("id") !== id)
-          ) {
-            // only push state if it's not the current url to prevent duplicated in history
-            window.history.pushState({}, "", `${getBaseUrl()}?id=${id}`);
-          }
         }
 
         setTitle(title);
--- a/packages/sandcastle/src/Gallery/Gallery.css
+++ b/packages/sandcastle/src/Gallery/Gallery.css
@@ -110,6 +110,11 @@
     padding: var(--stratakit-space-x2);
     gap: var(--stratakit-space-x2);
     overflow-y: auto;
+    &:focus-visible {
+      /* small margin to allow focus highlight to be visible all the way around */
+      margin: 0 3px 4px;
+      border-radius: var(--stratakit-ext-radius-xs);
+    }
   }
 
   .empty-list div {
@@ -148,7 +153,7 @@
     align-items: flex-start;
     flex: 1 0 116px;
     padding: var(--stratakit-space-x2);
-    gap: var(--stratakit-space-x1);
+    gap: var(--stratakit-space-x2);
     border-radius: var(--stratakit-ext-radius-md);
     color: inherit;
     transition: all 150ms ease-out;
@@ -207,6 +212,7 @@
       justify-content: space-between;
       flex-grow: 1;
       overflow: hidden;
+      gap: var(--stratakit-space-x1);
     }
 
     header {
@@ -215,7 +221,7 @@
       align-items: flex-start;
       border: none;
       gap: 0;
-      padding-top: 0;
+      padding-right: var(--stratakit-space-x6);
     }
 
     h2 {
@@ -237,7 +243,6 @@
       color: var(--stratakit-color-text-neutral-tertiary);
       padding: var(--stratakit-space-x1);
       margin: var(--stratakit-space-x1) 0;
-      cursor: text;
     }
 
     ul {
--- a/packages/sandcastle/src/Gallery/GalleryItemList.tsx
+++ b/packages/sandcastle/src/Gallery/GalleryItemList.tsx
@@ -28,8 +28,14 @@ export function GalleryItemList({
 }: GalleryItemListProps) {
   const composite = useCompositeContext();
   const store = useGalleryItemContext();
-  const { isSearchPending, items, searchResults, searchTerm, searchFilter } =
-    store ?? {};
+  const {
+    isSearchPending,
+    items,
+    searchResults,
+    searchTerm,
+    searchFilter,
+    galleryLoaded,
+  } = store ?? {};
 
   const totalDisplay = useMemo(
     () =>
@@ -69,16 +75,26 @@ export function GalleryItemList({
     [searchResults],
   );
   const deferredIsEmpty = useDeferredValue(isEmpty);
-  const emptyPlaceholder = useMemo(
-    () =>
-      isSearchPending || !isEmpty ? null : (
+  const emptyPlaceholder = useMemo(() => {
+    if (!galleryLoaded) {
+      return (
         <div>
-          <h3>No results</h3>
-          <p>Try adjusting your search filters</p>
+          <h3>Gallery loading...</h3>
         </div>
-      ),
-    [isEmpty, isSearchPending],
-  );
+      );
+    }
+
+    if (isSearchPending || !isEmpty) {
+      return null;
+    }
+
+    return (
+      <div>
+        <h3>No results</h3>
+        <p>Try adjusting your search filters</p>
+      </div>
+    );
+  }, [isEmpty, isSearchPending, galleryLoaded]);
 
   const list = useMemo(
     () =>
--- a/packages/sandcastle/src/Gallery/GalleryItemSearchInput.tsx
+++ b/packages/sandcastle/src/Gallery/GalleryItemSearchInput.tsx
@@ -9,7 +9,7 @@ export function GalleryItemSearchInput() {
   const inputRef = useRef<HTMLInputElement>(null);
   const value = inputRef.current?.value;
   const { setSearchTerm, items } = store ?? {};
-  const hasInput = !!value && value !== "";
+  const hasValue = !!value && value !== "";
 
   const clearSearch = useCallback(() => {
     const input = inputRef.current;
@@ -54,7 +54,7 @@ export function GalleryItemSearchInput() {
       />
       <IconButton
         className="gallery-search-input-clear-btn"
-        hidden={!hasInput}
+        hidden={!hasValue}
         icon={close}
         label="Clear"
         onClick={clearSearch}
--- a/packages/sandcastle/src/Gallery/GalleryItemStore.ts
+++ b/packages/sandcastle/src/Gallery/GalleryItemStore.ts
@@ -47,6 +47,7 @@ export function useGalleryItemStore() {
   >>(null);
 
   // Gallery items
+  const [galleryLoaded, setGalleryLoaded] = useState(false);
   const [items, setItems] = useState<GalleryItem[]>([]);
   const [legacyIds, setLegacyIds] = useState<Record<string, string>>({});
 
@@ -181,6 +182,7 @@ export function useGalleryItemStore() {
         setLegacyIds(legacyIds);
         setSearchOptions(searchOptions);
         setDefaultSearchFilter(defaultFilters);
+        setGalleryLoaded(true);
       });
     };
 
@@ -194,6 +196,7 @@ export function useGalleryItemStore() {
 
   return {
     items,
+    galleryLoaded,
 
     filters: galleryFilters,
     defaultSearchFilter,
--- a/packages/sandcastle/src/Gallery/loadFromUrl.ts
+++ b/packages/sandcastle/src/Gallery/loadFromUrl.ts
@@ -47,7 +47,7 @@ export function loadFromUrl(
     const base64String = window.location.hash.substr(3);
     const { code, html } = decodeBase64Data(base64String);
     return {
-      title: "Sandcaste",
+      title: "New Sandcastle",
       code,
       html,
     };
--- a/packages/sandcastle/src/App.css
+++ b/packages/sandcastle/src/App.css
@@ -35,7 +35,7 @@
     var(--highlight-color-border)
   );
 
-  header {
+  & > header {
     grid-area: header;
     display: flex;
     align-items: center;


jjspace:
Thanks @ggetz this is looking pretty good! Definitely glad search is much better with the title/description.

I didn't dig into the code yet but had a handful of small initial comments

Style nit-picks

- gallery card titles are just _slightly_ too wide for the code button, increase padding
- the code examples use the text cursor but clicking them activates it and text is not selectable. Switch to pointer
- The gallery cards feel ever so slightly cramped. maybe increase gap between and padding on edges?
- the image borders seem a little weird in light mode. There's a light line inside the border from something? Maybe use just the border not the outline
- The padding on the 121 for showcases seems bad. I know tt's consistent with the others to be circles but maybe make it expand?. Would happen with any 3 digit number
- The left edge of the labels is not aligned vertically with the title/desc for cards
- Should gallery card labels always be at the bottom of the card? I'm back and forth on this, fine either way I think
- In light mode the badges kinda "disappear" is that what we want? In dark mode they have more contrast. Maybe this is just stratakit itself
- Toggling the theme now creates a "2 step" transition with the main background then the cards (and a few inputs). I didn't see this before, worth investigating?
- When there are no code examples the cards feel just slightly too big. Should we reduce that?
    - Height also seems slightly dictated by the thumbnail aspect ratio. This should not be the case, we should force them to be consistent and crop/stretch as needed
- There is no border on the gallery list header anymore so when scrolling there's not a good distinction between header and items.
    - The number of results also isn't sticky. I think making it sticky with the divider line might be good
    - In dark mode this doesn't matter as much since cards have a different background but in light mode they're the same color and blend together. Maybe we need a different color for the cards?


Functionality

- Seeing a bit of lagging UI when typing searches fast, maybe we do need that debounce
- When loading a gallery example the header title still shows the highlight. This seems weird?
- Is the number in the labels filter supposed to update for the current search term?
- Not totally sure the line number adds much to the gallery but happy to leave it for now and try it out!
- The active "pressing" style applies to the whole card when clicking the label badges which seems wrong
  - Hover styles too but that one's less of a concern
- There's a bit of a lag in updating the "Results" string when switching a label
  - This also seems to apply to the style for the current label? It still shows selected even after the list updates
- The gallery cards are no longer Tab-selectable which is pretty major for a11y
  - The entire gallery list is though? not sure why but it creates a focus outline under the header that looks bad
  - Figured out they do some arrow navigation but tab doesn't work. I assume this is aria-kit shenanagins but probably need to figure that out
- The thumbnail alt text is `[object Object]`, oops!

Video of the slight 2 step lag/update when changing theme and filters:

https://github.com/user-attachments/assets/76d37105-9cf8-412c-8923-c5a12a608406

packages/sandcastle/scripts/createGalleryRecord.js
```diff
@@ -0,0 +1,44 @@
+export function createRecordHtml({ id, title, description, image, code, labels }) {
+    // Pagefind best indexes content in a hierarchical structure.
+    // Heading levels and data-pagefind-weight attributes will factor into the search relevance.
+    // Headings with id attributes are helpful for structuring the search subresults for highlighting.
+    return `
+<html lang="en">
+    <body id="${id}" data-pagefind-meta="id:${id}">
+        <h1 id="title" data-pagefind-weight="10">${title}</h1>
+
+        <img id="thumbnail" src="${image}" alt="${title} thumbnail" />
```
jjspace, line 17: Just curious, how does the image help with the page result?

ggetz:
@jjspace I did not get time to address all of the functionality and styling suggestions yet. But, I did re-work hooks to (hopefully 🤞 ) improve the update and re-render time, and I fixed the issue with the highlighted title and mis-named tab title.

I also merged main and fixed up linting issues. 

I think this is ready for another look.

jjspace:
@ggetz at a very quick initial glance it feels _much_ snappier now! I'm still noticing one or two small hitches but I think it's more on the browser/system side now than in React.

1. Turning _off_ the label filter to go back to the full list takes a bit to load the list again. I think this might actually be down to the loading of all the thumbnails. Likely outside the scope of this PR.
2. I think the search debounce should come back. When I'm searching for a term that has multiple results at each character it feels like it's lagging as I type because it's loading/filtering every step.
    - Alternatively or in addition to that it might make sense to prevent searching below some character length. Maybe require at least 3 or 2 characters to perform a search? Up to you, probably fine either way.

I'm working my way through the code now, full review hopefully soon

packages/sandcastle/scripts/createGalleryRecord.js
```diff
@@ -0,0 +1,55 @@
...
+  // Headings with id attributes are helpful for structuring the search subresults for highlighting.
+  return `
+<html lang="en">
+    <body id="${id}" data-pagefind-meta="id:${id}">
+        <h1 id="title" data-pagefind-weight="10">${title}</h1>
+
+        <img id="thumbnail" src="${image}" alt="${title} thumbnail" />
+
+        <h2 id="description" 
+            data-pagefind-ignore">Description</h2>
```
jjspace:
```suggestion
            data-pagefind-ignore>Description</h2> 
```

This is an actual bug ^^
But I'll also nit-pick that it might be good to do a format pass on this whole string. There's some trailing spaces, it uses 4 space indents instead of our normal 2 and generally doesn't totally match the prettier formatting.

packages/sandcastle/src/App.tsx
```diff
@@ -104,7 +105,13 @@ function RightSideAllotment({
       ]);
     }
     setConsoleExpanded(!consoleExpanded);
-  }
+  }, [consoleExpanded, previousConsoleHeight]);
+
+  useImperativeHandle(ref, () => {
+    return {
+      toggleExpanded: () => toggleExpanded(),
```
jjspace, line 112: I'm pretty sure this will create a closure that doesn't actually use the latest return value from the `useCallback` where this is defined. I _believe_ you have to set up another ref to store this. I recall bumping into something similar when hooking up the editor format function.

```js
const toggleRef = useRef(toggleExpanded);
useEffect(() => {
  toggleRef.current = toggleExpanded;
}, [toggleExpanded])

useImperativeHandle(ref, () => {
  return {
    toggleExpanded: () => toggleRef.current(),
  }
})

**Comment by jjspace** [comment]
While the search filtering is much better I wonder if we need to make the title weight _even_ stronger? Your go-to example of "label" doesn't return the "Labels" sandcastle as the first result. This can definitely happen as a follow up task to fine-tune those values

image

File: packages/sandcastle/src/App.tsx [hunk 4]
```diff
@@ -531,6 +437,49 @@ function App() {
     window.open(docsUrl, "_blank")?.focus();
   }
 
+  const onRunCode = useCallback(
```
jjspace, line 440: Can we rename this? I thought it was the function called when clicking "Run" for a sandcastle, not the function to load a gallery item and run it.

packages/sandcastle/src/App.tsx
```diff
@@ -531,6 +437,49 @@ function App() {
...
+        ) {
+          // only push state if it's not the current url to prevent duplicated in history
+          window.history.pushState({}, "", `${getBaseUrl()}?id=${id}`);
+          if (
+            !searchParams.has("id") ||
+            (searchParams.has("id") && searchParams.get("id") !== id)
+          ) {
+            // only push state if it's not the current url to prevent duplicated in history
+            window.history.pushState({}, "", `${getBaseUrl()}?id=${id}`);
+          }
```
jjspace:
```suggestion
```
This was accidental duplicated code in one of my previous PRs. I removed it in #12820 but it can be removed here too.

packages/sandcastle/src/App.tsx
```diff
@@ -646,7 +597,7 @@ function App() {
             setConsoleExpanded={setConsoleExpanded}
           >
             <Allotment.Pane minSize={200}>
-              {readyForViewer && (
+              {!deferredIsLoading && (
```
jjspace, line 593: Does it make sense to use this same check for rendering/disabling the editor too? When loading a specific gallery demo the default code shows up until the demo code is loaded which looks a bit weird. This was happening before this PR too so it's fine if this should be part of a follow up.

packages/sandcastle/src/Gallery/Gallery.css
```diff
@@ -0,0 +1,308 @@
...
+      }
+    }
+
+    .filter-menu-button {
+      justify-content: space-between;
+      text-transform: capitalize;
+    }
+  }
+
+  .list {
```
jjspace, line 107: This will make the "full list focus highlight" outline that we discussed actually show around the whole list which makes it _much_ more clear what that actually means.

```suggestion
  .list {
    margin: 0 3px 4px;
    border-radius: 1px;
    ...
```
(replace with stratakit vars)

packages/sandcastle/src/Gallery/GalleryItemList.tsx
```diff
@@ -0,0 +1,124 @@
...
+      {searchTermDisplay}
+      {searchFilterDisplay}
+      {totalDisplay}
+    </summary>
+  );
+
+  const isEmpty = useMemo(
+    () => !searchResults || searchResults.length === 0,
+    [searchResults],
+  );
```
jjspace, line 76: When the page initially loads the list is empty and it shows the "no items" message. However the list is still loading so that's not technically correct. Can we add a "loading" state to this check to make it more clear that we don't actually _know_ if there are no results yet?

packages/sandcastle/src/Gallery/GalleryItemList.tsx
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * @typedef {Object} GalleryItemListProps
+ * @prop {function} [onRunCode] A callback that gets called when a gallery item is opened.
+ * @prop {function} [onOpenCode] A callback that gets called when the code editor is opened.
+ */
+
+type GalleryItemListProps = {
+  onRunCode?: (item: GalleryItem) => void;
+  onOpenCode?: (item: GalleryItem) => void;
+};
```
jjspace, line 19: In TS-land there's much less needed for these sort of JSDoc comments. These descriptions can (should?) just be added "inline"

```ts

type GalleryItemListProps = {
  /** A callback that gets called when a gallery item is opened. */
  onRunCode?: (item: GalleryItem) => void;
  /** A callback that gets called when the code editor is opened. */
  onOpenCode?: (item: GalleryItem) => void;
};
```

packages/sandcastle/src/Gallery/GalleryItemList.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Component that displays a composite list of gallery items.
+ * @param {GalleryItemListProps} props
```
jjspace, line 23: Similar comment here. The type is defined in the TS itself, no need to add it to the JSDoc block I believe

packages/sandcastle/src/Gallery/GalleryItemList.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <GalleryItemCard
+            onClick={() => onRunCode(item as GalleryItem)}
+            key={item?.id}
+            index={index}
+          >
+            <IconButton
+              icon={developer}
+              label="Open and view code"
+              onClick={() => {
+                const target = item as GalleryItem;
```
jjspace, line 112: I realize it's working fine but I wouldn't expect to need to typecast here. Can we clean up the types or guarantee them earlier when loading? We should be able to "guarantee" the properties available on the type. Maybe this needs a new "GalleryItemResult" type? Also it seems `item` could be `undefined` but "`undefined` is not assignable to type `GalleryItem`". This `as` ignores the warning for that check.

packages/sandcastle/src/Gallery/GalleryItemSearch.tsx
```diff
@@ -0,0 +1,13 @@
+import GalleryItemSearchFilter from "./GalleryItemSearchFilter.tsx";
+import GalleryItemSearchInput from "./GalleryItemSearchInput.tsx";
+
+export function GalleryItemSearch() {
+  return (
+    <form role="search">
+      <GalleryItemSearchInput />
+      <GalleryItemSearchFilter />
+    </form>
```
jjspace, line 9: nit-picking but does this need to be it's own component? why not just inline this where it's used?

packages/sandcastle/src/Gallery/GalleryItemSearchInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { close, search as searchIcon } from "../icons.ts";
+
+import { useGalleryItemContext } from "./GalleryItemStore.ts";
+
+export function GalleryItemSearchInput() {
+  const store = useGalleryItemContext();
+  const inputRef = useRef<HTMLInputElement>(null);
+  const value = inputRef.current?.value;
+  const { setSearchTerm, items } = store ?? {};
+  const hasInput = !!value && value !== "";
```
jjspace:
```suggestion
  const hasValue = !!value && value !== "";
```

I first read this as "does this have the input element"

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -0,0 +1,216 @@
...
+  useMemo,
+  useRef,
+} from "react";
+import { getBaseUrl } from "../util/getBaseUrl.ts";
+import { applyHighlightToItem } from "./applyHighlight.tsx";
+import { loadFromUrl } from "./loadFromUrl.ts";
+import "../../@types/pagefind-client.d.ts";
+
+const galleryListPath = `gallery/list.json`;
+const pagefindUrl = `gallery/pagefind/pagefind.js`;
```
jjspace, line 18: I'm a little worried about this change and the related change to move the generated gallery into the `public` directory. One of the main reasons this was a `vite` defined variable at _build_ time was to allow it to not be co-located with the files. Specifically this allowed us to point to the "nested" `packages/sandcastle/gallery` directory when running locally. This meant that any changes to the gallery would be updated immediately without needing to re-build sandcastle itself. 

I think moving where the built files are might be fine as long as things like the "on-demand rebuild" still work with the top level dev server. However I do think this variable should go back to being based on the build environment instead of relying on the public folder structure.

Also as a related side-note, we've had discussions around potentially loading multiple galleries or galleries that are hosted elsewhere. I haven't finished reviewing all the code here but I do want to make sure we're still keeping that in mind here

packages/sandcastle/src/Gallery/GalleryItemSearchFilter.tsx
```diff
@@ -0,0 +1,126 @@
...
+    [
+      deferredCheck,
+      defaultSearchFilter,
+      memoizedFilters,
+      deferredOnChange,
+      setSearchFilter,
+    ],
+  );
+
+  return <>{memoizedFilters.types.map(renderFilter)}</>;
```
jjspace, line 125: It took me a little while to actually wrap my head around what's going on here and what the `types` actually _is_ here. I think I understand that this is to make the filters dynamic if we added more filter types in the future? Please correct me if I'm wrong.

First I'm not totally sure we need that? I think what's rendered should be intentional. We know we have a labels filter and want to display that. This is guaranteed by the gallery build script and metadata structure itself.
Second I feel that trying to account for the future _potential_ of multiple filter types has complicated this component a lot. Maybe we can just define the type we want as a prop so it's still "dynamic" for future expansion but simplifies the code?

Last, and most important, when loading the page this component doesn't render anything until the pagefind search first loads. This means the search input has to shift over after the dropdown renders. I don't think this looks good. I would much prefer if we could render the filter itself from the start and only populate the dropdown options once the filter values actually load.

packages/sandcastle/src/Gallery/loadFromUrl.ts
```diff
@@ -0,0 +1,103 @@
+import { type GalleryItem } from "./GalleryItemStore.ts";
+import { decodeBase64Data } from "../Helpers.ts";
+import { getBaseUrl } from "../util/getBaseUrl.ts";
+
+const loadGist = async (gist: string) => {
```
jjspace, line 5: Can you please add back some comment that this is purely for legacy support?

Edit: I now see the comment below where this is called. Maybe should be in both places but up to you.

packages/sandcastle/src/Gallery/loadFromUrl.ts
```diff
@@ -0,0 +1,103 @@
...
+  items: GalleryItem[],
+  legacyIds: Record<string, string>,
+) {
+  const searchParams = new URLSearchParams(window.location.search);
+
+  if (window.location.hash.indexOf("#c=") === 0) {
+    const base64String = window.location.hash.substr(3);
+    const { code, html } = decodeBase64Data(base64String);
+    return {
+      title: "Sandcaste",
```
jjspace:
```suggestion
      title: "Sandcastle",
```

Also setting this results in a title of "Sandcaste | Sandcastle | CesiumJS" which is very redundant. The previous behavior was all sandcastles would have the title "New sandcastle" if they weren't a specific demo. If we want to depart from that I'm open to it but the title logic should probably update accordingly to prevent the double "Sandcastle"

jjspace (commented):
I still need to review the new gallery build code and the styles but here's a bunch of comments to start with on the functionality.

Also I notice a _lot_ of `useCallback` and `useMemo`s scattered around now. Maybe this is fine but it feels like something that was specifically called out as a pitfall of other mistakes in these articles about [memoization](https://tkdodo.eu/blog/the-uphill-battle-of-memoization) and [useCallback](https://tkdodo.eu/blog/the-useless-use-callback). What's there does seem to work for now so no immediate changes needed. I do think it would probably be worth doing a more complete performance pass later and see if these are really necessary or there's other flaws in our react structure. I definitely don't claim to be an expert here yet.

packages/sandcastle/src/App.tsx
ggetz: I think I just copied the existing hook and moved it further down in the code, in which I think this is a pre-exiting issue—Unless you already fixed it and I messed up a merge.

packages/sandcastle/scripts/createGalleryRecord.js
ggetz: It's not important for how we use it now, but an image is an expected default by pagefind along with the page title. I say we leave it, as it doesn't add much code burden at all and means the indexed data is more flexible going forward.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
ggetz: I'm not sure I see the issue. The on-demand rebuild still reflects changes to the gallery. 

> I do think this variable should go back to being based on the build environment instead of relying on the public folder structure.

A public folder already exists for static assets. And, since the gallery is always copied to the public directory, there is no longer be need to configure a gallery path per-environment build in the Sandcastle application. IMO this simplifies away a moving part—It separates the concerns of managing and mapping gallery files and the Sandcastle application itself. 

> we've had discussions around potentially loading multiple galleries or galleries that are hosted elsewhere

If anything, this makes loading multiple galleries easier: They can all be copied into the public directory without having to re-map multiple gallery paths for the sandcastle build itself.

jjspace: > I'm not sure I see the issue. The on-demand rebuild still reflects changes to the gallery.

Not the "correct" gallery though.

- Run `npm run build-sandcastle` from the repo root
- Run `npm start`
- Navigate to the static/built version locally at `http://localhost:8080/Apps/Sandcastle2/index.html`
- Load any sandcastle and check what it does
- Modify that sandcastle's source
- Reload the page. The sandcastle has not been updated
- Run `npm run build-sandcastle` again and now it's updated

This used to work without that last step because when we built Sandcastle locally we pointed it into the `packages/sandcastle/gallery` directory itself. I'm fine if we want to point into `public/gallery` still but I don't think we should be building the gallery in different places or duplicating all those files for _local_ development specifically.

I was always viewing the gallery as a distinct "entity" that exists on it's own. Yes ours is contained in the same package but I don't think it _needs_ to be. Maybe you want to point it to a gallery hosted somewhere else on a different server/url. Maybe you host it locally but from a different port. The point is I don't think the client should care as long as the files follow the expected format/spec.

packages/sandcastle/src/Gallery/GalleryItemList.tsx
jjspace: I modified this slightly in my small changes commit but it can probably be improved

packages/sandcastle/src/Gallery/loadFromUrl.ts
jjspace: I set this back to "New Sandcastle" so the other assumptions in the app behave as expected. I'm leaving this comment unresolved just so we remember to discuss it later in case we do want to change the behavior.

jjspace:
As discussed in person I've made some minor updates to help get this over the finish line for the release. I wasn't able to find any major issues with the functionality actually _using_ the app. My remaining comments above are on a more technical level and should be addressed soon after but are not blockers.

I'm happy to merge and follow up on the remaining comments later in the week.

Thanks for all the improvements here @ggetz!

Issue #12725, opened 2025-07-08
ClippingPolygonCollection + clampToHeight = Uniform update failure


angrycat9000:
### What happened?

Using `Scene.clampToHeight` with a clipping polygon throws an error during updating the shader.  Occurs while dragging (or updating frequently) the clipping polygon.

<img width="754" height="736" alt="Image" src="https://github.com/user-attachments/assets/77649d1a-24bc-4574-bd28-0c121110ef36" />

### Reproduction steps

See sandcastle.

Seems to occur with frequent updates and then also doing a clampToHeight. Does not occur without the clampToHeight call.

Re-using the clipping polgyon didn't trigger the interaction but has issues with updating display: https://github.com/CesiumGS/cesium/issues/12724

### Sandcastle example

[...]

PR #12856, opened 2025-08-29
Fix picks on clipped regions

ggetz:
# Description

This PR fixes a uniform error which can arise when picking 3D Tilesets that have a region clipped by `ClippingPolygonCollection`.

As @javagl helpfully pointed out, the `ClippingPolygonCollection._extentsTexture` property is [set during the collection update](https://github.com/CesiumGS/cesium/blob/a9480d31fdd85217c385b3bfc1659ae83c8d975c/packages/engine/Source/Scene/ClippingPolygonCollection.js#L627) (the webgl context, required to create the Texture, is not available until then). The error can manifest during picking, particularly `clampToHeight `, which attempts to render sooner than expected in the process. This leads to an uniform error.

Instead, this PR simply defaults to the default empty texture in the case where the texture has not yet had the chance to initialize.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12725

## Testing plan

**You will need to open the developer tools in order to replicate the bug.**
- [`main`, with error when you drag and move the polygon](https://sandcastle.cesium.com/#c=rVZZcyI3EP4rCk9DFRFzH74qhHXKrrJjyuDdB+PaEjMClBXSlEbgYl3897TmwGNsr7NJeADU6uPr7q97pt8fSSY0GnK5ztBsi0ZKzjEaUU0VGnBORQ8NJV+vZoygO8E2VBVMb9GtnEnN0gJdkRlG45QIwcTCOBhw+hcRmZJoomSaSs4QHNE10ZShIZMpUZxRPBWpFIVGG0YfIdQpEvQRDWnB1iv8uZRZ005anodSaMIEVdNODz1NBUIAToHkqDGYVGc8V3L1RSqe1QKr25uKXfd4KqaiCoSLlAqKU7KiiuCCahPLKp1mtNBMEM0kOG6hGRKl4R8RnmXUEPJ924l8H4e+7/iB5/i9Su66gR042HfsKE4cxw7c+sIPksB3bexFfhT7vucZcbe8lIpRyO5V0AtKMijoiOl0eSs5r0MHEDSMPC+2A9tLoiCqI/xqYzcKQztyAi/2kjgO64sQu7EDWOIo8G3bjoMmdF2Vqgm5hKYChAL6cG807k2OYexiL7CjJHDdpFemZ/vY8xI/8hKI4/aqzMIQe3YQe2Hgxg+9vXkEyknsBVHg2I4x95MI4JiPE0QGe2Xuh9gByIGb+G1rFwrsOW5ie55bBw8xfEWhDZW1w9raDXASeXGYeIFnzB/wiuSWlXfR6dnrDpYMGQA3tqDRroBmnAIbIH/ySJjeW5Y/3qdJdV2aX0oxKOBwmVlO6LuOcfOCXLliKyjnhhaYZJlVuzZq9V+ccpbnpr2SbxdV2duEe3kL48dpatpT8ZQJM4T0CGm1pmXB8trNEbp/34311GrzrvvwzAFOdeNi1CJCWaYyY2tvWJE2U2SxMON+iuaEFxUIRVNFYchBaIC1Zw4YDuZ1OZ5aiI+qcUZoyWAcVbrcHkwe5zOSfoOdlFOlt/UQIGSV3W1p1jleNG6sw3y6vca2RFwdGuEKYCtG+H6fQMmlwrfnn/Aj08sBz5fEsnFQ6e8Op2cJ+40f7rAxVIOKcU5Ser6BAlxUSlZdEliYG1IYL7W12UWXIl/rQdVqixqrMs+q63Nk/dIUvtvUTVG9VuK4hGW+KkAAo0kcQL3kJoNy1ndVCOhO1ki6x/tQLR8fREs5WeU0O4xUiifygrLFUlttd89BatN3A7zBynuM8Z6OvSb4w95nQ8O9z/9l5szn9dyZz0/P3tHrrKppNN52VXF2iAJJP8oAK7qSGwpPaasy+4GqmbwG8n/AWc9N06Fd7z26T7Y5xdc3d+Pzr9c3n89/RPRnjrf2Sr1C0EtOFc9RhuUT3LwaKNMrs2TIjNPSc9EspuMPIV6d/zH5+unmy58/jbAJ8O8h1jn+E4R3o2rhjAFhSgrNqenoRC4WnP6+1hpQGiTTzm2zhd8lNLxEtdd1eZivRXmHrHRJ028vJnK/1eurpvcGUKfXOSn0ltOzhlu/sVUulUZrxS2M+5qucg7mRX+2BmMgZlE0XD3pt01PMrZBLDt946XPDHlRwM18zfmYfafTztlJH/RfmXJZvjXdwKBysjVqS+fsqhLC3jjpw/FtSy0lPGrUgee/AQ)
- [this branch, no errors when you drag and move the polygon](https://ci-builds.cesium.com/cesium/fix-12725/Apps/Sandcastle/index.html#c=rVZZcyI3EP4rCk9DFRFzH74qhHXKrrJjyuDdB+PaEjMClBXSlEbgYl3897TmwGNsr7NJeADU6uPr7q97pt8fSSY0GnK5ztBsi0ZKzjEaUU0VGnBORQ8NJV+vZoygO8E2VBVMb9GtnEnN0gJdkRlG45QIwcTCOBhw+hcRmZJoomSaSs4QHNE10ZShIZMpUZxRPBWpFIVGG0YfIdQpEvQRDWnB1iv8uZRZ005anodSaMIEVdNODz1NBUIAToHkqDGYVGc8V3L1RSqe1QKr25uKXfd4KqaiCoSLlAqKU7KiiuCCahPLKp1mtNBMEM0kOG6hGRKl4R8RnmXUEPJ924l8H4e+7/iB5/i9Su66gR042HfsKE4cxw7c+sIPksB3bexFfhT7vucZcbe8lIpRyO5V0AtKMijoiOl0eSs5r0MHEDSMPC+2A9tLoiCqI/xqYzcKQztyAi/2kjgO64sQu7EDWOIo8G3bjoMmdF2Vqgm5hKYChAL6cG807k2OYexiL7CjJHDdpFemZ/vY8xI/8hKI4/aqzMIQe3YQe2Hgxg+9vXkEyknsBVHg2I4x95MI4JiPE0QGe2Xuh9gByIGb+G1rFwrsOW5ie55bBw8xfEWhDZW1w9raDXASeXGYeIFnzB/wiuSWlXfR6dnrDpYMGQA3tqDRroBmnAIbIH/ySJjeW5Y/3qdJdV2aX0oxKOBwmVlO6LuOcfOCXLliKyjnhhaYZJlVuzZq9V+ccpbnpr2SbxdV2duEe3kL48dpatpT8ZQJM4T0CGm1pmXB8trNEbp/34311GrzrvvwzAFOdeNi1CJCWaYyY2tvWJE2U2SxMON+iuaEFxUIRVNFYchBaIC1Zw4YDuZ1OZ5aiI+qcUZoyWAcVbrcHkwe5zOSfoOdlFOlt/UQIGSV3W1p1jleNG6sw3y6vca2RFwdGuEKYCtG+H6fQMmlwrfnn/Aj08sBz5fEsnFQ6e8Op2cJ+40f7rAxVIOKcU5Ser6BAlxUSlZdEliYG1IYL7W12UWXIl/rQdVqixqrMs+q63Nk/dIUvtvUTVG9VuK4hGW+KkAAo0kcQL3kJoNy1ndVCOhO1ki6x/tQLR8fREs5WeU0O4xUiifygrLFUlttd89BatN3A7zBynuM8Z6OvSb4w95nQ8O9z/9l5szn9dyZz0/P3tHrrKppNN52VXF2iAJJP8oAK7qSGwpPaasy+4GqmbwG8n/AWc9N06Fd7z26T7Y5xdc3d+Pzr9c3n89/RPRnjrf2Sr1C0EtOFc9RhuUT3LwaKNMrs2TIjNPSc9EspuMPIV6d/zH5+unmy58/jbAJ8O8h1jn+E4R3o2rhjAFhSgrNqenoRC4WnP6+1hpQGiTTzm2zhd8lNLxEtdd1eZivRXmHrHRJ028vJnK/1eurpvcGUKfXOSn0ltOzhlu/sVUulUZrxS2M+5qucg7mRX+2BmMgZlE0XD3pt01PMrZBLDt946XPDHlRwM18zfmYfafTztlJH/RfmXJZvjXdwKBysjVqS+fsqhLC3jjpw/FtSy0lPGrUgee/AQ)

I did not add a unit test since this is an edge case that is difficult to construct a sensible test for.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Model/ModelClippingPolygonsPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/ModelClippingPolygonsPipelineStage.js
@@ -79,10 +79,14 @@ ModelClippingPolygonsPipelineStage.process = function (
 
   const uniformMap = {
     model_clippingDistance: function () {
-      return clippingPolygons.clippingTexture;
+      return (
+        clippingPolygons.clippingTexture ?? frameState.context.defaultTexture
+      );
     },
     model_clippingExtents: function () {
-      return clippingPolygons.extentsTexture;
+      return (
+        clippingPolygons.extentsTexture ?? frameState.context.defaultTexture
+      );
     },
   };
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 
 - Removes the minimum tile threshold of four for WMTS. [#4372](https://github.com/CesiumGS/cesium/issues/4372)
 - Fixed a crash that could happen when loading PNTS (point cloud) data that contained a batch table without a binary part [#11166](https://github.com/CesiumGS/cesium/issues/11166)
+- Fixed an error picking an area hidden by a `ClippingPolygon`. [#12725](https://github.com/CesiumGS/cesium/issues/12725)
 
 #### Additions :tada:
 


ggetz:
@mzschwartz5 Could you please review? It'd be great to get this fix into Tuesday's release for the ion team.

javagl:
You have seen in the issue that I did go down the rabbit hole a bit, but didn't bother to look for the "Drink me" bottle. 

Two small comments:

For me, the issue only happened when the debug console was open. One can shrug this off: "It's a browser, not a VM 🤷". But is there any convincing reason for that? 

The change here prevents the crash. Let's hit "Merge" 😉 
But ... I'd at least consider to add
```javascript
      return (
        // When [something happens] then [in some cases] and [when the browser console is open(!?)]
        // this may be called by [some function], at a point in time where this texture is not yet initialized.
        // Using the default texture is a workaound for https://github.com/CesiumGS/cesium/issues/12725
        clippingPolygons.clippingTexture ?? frameState.context.defaultTexture
      );
```
And... it's indeed only a workaround to prevent the crash. Whatever this `_extentsTexture` **is** (I don't know that), it did probably contain specific _values_, and these values have been _used_ at the consuming side, to perform some sort of computation. Now, whatever "value" that was, it is now the value for "white" for a short moment. And depending on what was _done_ with the value, it's hard to predict the effect of this. (Some unpredictable area being clipped? Maybe showing up as soms sort of "flicker"? And.... maybe, one day, we change the color for the default texture to magenta, and suddenly, the clipping does no longer work....?). 
(I'm not saying that this "patch" is the wrong way to solve it. I just want to make sure that it is the right way to solve it...)

ggetz:
> The change here prevents the crash. Let's hit "Merge" 😉

My comment might end up boiling down to this, but... 

Thanks for for flagging the nuances @javagl — you’re right that this is more of a workaround than a root-cause fix, and your notes about the fallback texture are valid. That said, since this crash is currently blocking ion, I’d suggest we move forward with this patch as an immediate safeguard. 

To your points:

1. I ran this and related clipping examples and didn’t observe any behavioral differences.
2. I suspect the “only when console is open” effect comes from devtools overhead exposing a race condition.
3. While the fallback isn’t a perfect replacement, in the cases where it triggers it should only be used for about a frame of a non-rendering pass — and using zero values should effectively result in “no clipping.”

So my recommendation is: let’s merge this as a stopgap, and if there are still concerns on the underlying issue, let's file a follow-up issue to better re-evaluate timing of texture initialization.

mzschwartz5:
Taking a review pass now - just noticed right off the bat, though, that I'm unable to replicate this in Firefox. No problem in Chrome though. Don't know if that's worth anything but I hadn't seen it mentioned anywhere yet.

packages/engine/Source/Scene/Model/ModelClippingPolygonsPipelineStage.js
```diff
@@ -79,10 +79,14 @@ ModelClippingPolygonsPipelineStage.process = function (
 
   const uniformMap = {
     model_clippingDistance: function () {
-      return clippingPolygons.clippingTexture;
+      return (
+        clippingPolygons.clippingTexture ?? frameState.context.defaultTexture
+      );
```
mzschwartz5: The issue description said part of the problem is the lack of an available WebGL context until the `Update` function in `ClippingPolygonCollection`. Clearly though, we have a valid context here, at least. 

Question - rather than returning the `defaultTexture`, could we call a function on the `clippingPolygons` object here to create the `clippingTexture` if it's undefined at this point? (This function I'm referring to doesn't exist yet, but could we just refactor the code in `Update` that creates the texture into a callable function?)

If that seems like a big refactor / we need this sooner and it works as is, I'm fine shipping it.

mzschwartz5: (Regardless, I'll hold off merging any code changes until people are back in office)

javagl:
> My comment might end up boiling down to this, but...

And if my comment boils down to adding a comment 
`// Should never happen - quick fix for https://github.com/CesiumGS/cesium/issues/12725`
then that's OK. I know that time pressure (deadlines) is one of the very few _legitimate_ justifications for not fully analyzing the reasons and finding a holistic solution for certain issues. It's not a legitimate reason for not adding a `// Quick fix for #12725` comment, though.

packages/engine/Source/Scene/Model/ModelClippingPolygonsPipelineStage.js
ggetz: We could, but it would still be an empty texture without the relevant info being copied to the texture. And I think that really needs to be called from the `ClippingPolygonCollection.update` function, which is owned by the 3D tileset and called specifically before any passes have begun. 

Functionally, this produces the same result as creating the textures here if they are undefined.

ggetz:
I went ahead and added a comment as Marco suggested. Ready for another look @mzschwartz5.

mzschwartz5 (approved):

PR #12863, opened 2025-09-02
Model:  add the options.customGeometryStageVS, to override the default VS geometry stage

mickae1:
add the options.customGeometryStageVS, to override the default VS geometry stage, to let the user define its own VS geometry stage.


mickae1:
@mzschwartz5 I've created this pull request, can you validate it ? I will create a new issue after this one is validated to discuss about projection optimization.

javagl:
I've seen some discussion in the forum and the issue https://github.com/CesiumGS/cesium/issues/12815 . There was a lot of back and forth, with https://github.com/CesiumGS/cesium/pull/12825 and the related discussion in the issue. At some point, I stopped trying to follow the random pieces of information that have been scattered in the forum, issue, and PRs.

But regarding this PR: 

- The higher-level requirement and goal was not presented clearly (neither in the issue nor in the PRs)
- The intended approach for resolving the issue was not presented in the old PR, and also not in this one. 
  - (Dumping some GLB file into a GitHub comment is not an appropriate description for anything...) 
- The `_customGeometryStageVS` that is introduced here is a `boolean`. But it is passed to `addVertexLines`, which only accepts `string|string[]`. 
- This property is intended for replacing a _whole shader function_ that has the signature
  `vec4 geometryStage(inout ProcessedAttributes attributes, mat4 modelView, mat3 normal)`
  This is a private implementation detail. It is not intended to be overridden. It is not documented anywhere.
- Don't simply delete the Pull Request template. It serves a purpose

So in the current state, that's a strong 
**NO!!!** 
for this PR.

mickae1:
@javagl, I understand, and we are here to talk 👍 

you can find here a sandcastle : 
[sandcastle](http://localhost:8080/Apps/Sandcastle/#c=jVf7b9s2EP5XuOwXefXo1M5awHlgme0EBpwHYjdYAQMuI51sbpLoUZRjecj/viMp6uEqxdoAEcnvHrz77njxRZIqAkyqzRMLeJaSS/Jp8OnT59PP58tkmYwg5VlMJ1HEt6ngAQ0gZFmkEJbAKzk+9pYJqWvrkl6P7Ik0i7azvP2M6LNDedYxvvjG160Uf4GvuEiaPtyCWEu23XD/sUR477lvFTqVOw6vIJvqns2etzzxzXokEsV4AnJ50iX/am/1v3UkXmDY8ELvvG+26yQl/JNBqp4gCUDeiQC1KJlBeR6zPY+z2J4veAyjDUvWiJomIU+4ymvIbXXhYS08dYS20B8PnZd3bodO72+m99PFZDUfPT3MZqVM6kMC1q9CZu526Hw0uZ/0xxb65kLpgmkkMZY2qNQsa+nzs1SJ+HmOiG/LJIwEUyRbtRKQnlcAEYYpaNb1T09pTd0O/AH58rh6nowWD0+r/lhbxj3vtEvw56P17kWIiPBUez8Ye50ifxJUJhPiH+JVLOR2o8NMLi/JR23grSZ3L9T/Ef2pRdSGuU2ujLA22djoj50SvMkZWYOIQcl8rtgaPJ6ITBHMOPIyheBaKclfMgUpYeVnF1OuzkiMuiLNY7MekETImEWdZVI4Y2K3FSnXbLkbYeQqFbTaP7doHno/dnsw7pSFgcpXTsNEa/ZKZ8gv2vCZVxnANNHTTofu80Nh6w2iFGrKsBkkAAEEJBSSYJKVIBu2A3LY5++b1N7tWi1SlGuxWobljPgi3mIcgsdChtjL1ppPobRu1Kos6sFUEahxnrCY+89M5jxZp14VYguss+LIZo0FA1fXC1H1Oc8WR8RUlxSfWPXFJ4sUV1kALid21zbUP/E2jZo7/x7z9YeYPZ46Xb9os43TvDz9qk+ZOr6pqdB9l2ASSj/bbuvZxQp8CN+7GA+JV5V2nYKFMa/ZXj5U3eRDpQz9tOXBD+BZg52KFrV6sSj9VrwDL5Oi6ynliWeBND8CGE4xxUoAxqP4OlTUQOb3x70+/W1sypkMm6WLKtqIYTlh2NB22VJ7Fbv+UexQNz2gdtNsa8VRxBSPG/nyWeRnaBQeyxrzGu2l6jrNLKKJCJK12pTV2SG/thLPWS6Y4eBtDMIXFyMkFezv8M32ns3nNMHiIrvU/O4S20dtA9edyWAeMmVB9qNjX3q3pI1e2Xplq5229tHCv2/nzafwpngKjdehZOsYEmX8vikW1vPwHc/v0AXJkZJx8dEhtTy6TRrwMMxSKJ9Hiu8j1Q3L0bKBZtF2w0z6B5XXTb/nGxYcD02j2olX84KQiK83Chug5lpUzhSz+i79cj+bLrp1KRcOq3EBezUso9YA2mx/D3uedytQQ6Dox0PScFM/IulGvJYeFn17kW+B3swerhd1NMalGH+IjYxJiC7rV8ZVOWyZy4VSxLeRCq/TPPFdbDKJsVieUNrDnzmLtxGMsSX0jJ60h/xRdB294Mxp8PZOt/V54HmuFfyM8yUOp2T0Zb54uFvNF9e3EydEHEsk39cGQL08o9Px5H4xXXztFsFh2ruNFEgxDE3I8BkutNSTPmysDOCtMaJHLIFHyWNk/u54FqRbd5BSFpi/GIyHXTfmOy0SYrGDGU8VSmmiGRiVwIJ8skNWaHHz4TCe1yGXV0XRWpM+i0EyGkb5QvwhsiTAZM63G5BgJxL60tgsZ/sgk8wO1Kf0tLohMYkWEdBIrL3GtVZI1hgwKQooFtpMrMewVRvXbJu38UwzqEZn9/+ke3KRqjyCK8ez33m8FVJpqnhIEgVIEjSR9l4y/29Q1E9TV8EXvbroRcB3hAeXLX/BED9iaYonYRZFc3zClidXFz3EfyeKvVoH5wHrK2K5hm0+Xs3sJqX0oofLdkmF8+8Lk0ea/wM)

with this 
[test.zip](https://github.com/user-attachments/files/22136358/test.zip)
to unzip to cesium\Apps\SampleData\models

I'm sending a glb file with points that are already in ecef coordinates, that's why the matrix identity is used. ( why ? because I can have thousands geometries all over the world )

I had no choice to override the geometryStage function because in 2D mode, the matrix that I gave is modified here :  

https://github.com/CesiumGS/cesium/blob/4da5e56230a9038f9231440a4b154313252d03e2/packages/engine/Source/Scene/Model/ModelSceneGraph.js#L340

Since I have an identity matrix, no translation is detected, which causes Cesium to compute a new matrix. However, this process failed in my case. ( I guess the case of IDENTIY matrix is not supposed to happen )

That's why in my code, I'm using czm_view. I also had to change the rotation order to ZXY instead of XYZ.: 

```
vec4 geometryStage(inout ProcessedAttributes attributes, mat4 modelView, mat3 normal)
{
    vec3 positionMC = attributes.positionMC;
    if(czm_sceneMode == czm_sceneMode3D){
        v_positionEC = (modelView * vec4(positionMC, 1.0)).xyz;
    }else{
        // needed for 2D to have zxy
        v_positionEC = (czm_view * vec4(positionMC.zxy, 1.0)).xyz;
    }
    vec4 computedPosition = czm_projection * vec4(v_positionEC, 1.0);


    setDynamicVaryings(attributes);

    return computedPosition;
}
```

**Frankly, I would prefer an option to not pass to geoemtryStage, and set myself the gl_Position**

@javagl Since this approach is generic and prevents private functions from being overridden, is it the better option? 

This will improve performance for me. I need to transform points into eye coordinates to compute width of lines, but I'm wasting time transforming them back to Model Coordinates just to follow the Cesium pipeline.

javagl:
The perfect pull request follows the pull request template, has all [✔️] checkboxes ticked, passes the CI build, implements an agreed-upon approach for solving a specific problem, and can be merged quickly.

There has been related disucssion in the forum, at https://community.cesium.com/t/show-3d-globe-from-a-glb-file-at-the-center/40421 . There has been discussion in my (private) inbox of the forum. There has been a (long) discussion in the issue at https://github.com/CesiumGS/cesium/issues/12815 , involving Luke, Janine, and Matt. There has been a long discussion in the PR at https://github.com/CesiumGS/cesium/pull/12825 (which was then just closed). Now there is **this** PR, where we could start the discussion at zero (at the risk that you're closing the PR tomorrow).

I'll **quickly** address the last comment, but what I'm writing here is also just **comments** (no need to respond to each of them - the main point is at the bottom)

> I'm sending a glb file with points that are already in ecef coordinates, that's why the matrix identity is used. ( why ? because I can have thousands geometries all over the world )

You can have thousands of geometries all over the world, even when they are _not_ in ECEF coordinates. You just have to set the `modelMatrix` accrdingly. From what I understood so far, the issue is somehow related to the fact that 1. the geometry is "large" and covers a larger part of the globe, and 2. there is some issue related to the 2D visualization. But again: It's hard or impossible to follow all the discussions when the actual issue is not clearly and succinctly described in one place. And it's even harder to reverse engineer some intention from a pull request, and say whether the approach that is implemented there is the right one.

> I had no choice to override the geometryStage function because in 2D mode, the matrix that I gave is modified here :

Yes, this code is part of something that has ... let's call it "huuuuge 'room for improvement'" 😶 I don't know what exactly is happening in this code, and where the `_computedModelMatrix2D` is used (for what), and what the difference is that depends on the translation being ZERO.

> That's why in my code, I'm using czm_view. I also had to change the rotation order to ZXY instead of XYZ.:

That's a piece of shader code. I don't know what `czm_view` is (beyond ~"some view matrix"). The fact that you have used `zxy` instead of `xyz` suggests that there is something wrong with the axis conversions, and the lack of explanation there suggests that it is not clear what _exactly_ is wrong with the axis conversions.

> Frankly, I would prefer an option to not pass to geoemtryStage, and set myself the gl_Position

Computing some 2D position from 3D, depending on the position, projection, and all the matrices that are involved there, is difficult. But whatever the goal is: Overriding an internal, undocumented shader function that uses internal, undocumented variables, with a questionable axis conversion, is not something that can become part of CesiumJS. 

> This will improve performance for me.

Before talking abot performance, we have to figure out far more fundamental aspects. 

---

To the main point: 

In the forum- and inbox threads and initial issue, you provided a GLB that, when rendered in 3D, looks like this:

Cesium Issue 12815 that projection to 2D thingy

That looks like it could be right. Now there was something about the 2D mode. And indeed, when switching to 2D, it looks like this:

Cesium Issue 12815 that projection to 2D thingy in 2D

That looks wrong. That geometry, in 2D, would probably **roughly** look like this:

Cesium Issue 12815 that projection to 2D thingy in 2D draft

# Is this what you are trying to achieve?

(And if it is: Does any of the PRs or suggested snippets achieve that?)

mickae1:
@javagl 

> From what I understood so far, the issue is somehow related to the fact that 1. the geometry is "large" and covers a larger part of the globe, and 2. there is some issue related to the 2D visualization.

That's absolutely right.

Sorry to posting at many placed, but I wanted a clear discussion here, because I in other thread I posted a lot message to find a solution. And It finally work in my case. I have forked cesium and I applied my modification to do what I want.

Now, I'm trying to discuss with you about what I've found so far: 

When cesium load a geometry, it run  the function : 

https://github.com/CesiumGS/cesium/blob/4da5e56230a9038f9231440a4b154313252d03e2/packages/engine/Source/Core/GeometryPipeline.js#L616

which from ecef points convert it back to lat,long and then apply the projection and store the result in the attribute position2D. When you look at the vertex shader of polyline, it takes the attribute position2D and then rotate the axis ( xyz to zxy ) :

https://github.com/CesiumGS/cesium/blob/6cb7dd5de2238ec17e4c82a8d10b72501784b374/packages/engine/Source/Shaders/PolylineVS.glsl#L47

I had to replace czm_modelView to czm_view because in case of 2D projection, the matrix model is modified and modification doesn't work for matrix identity.

I have successfully projected my geometries in 2D with this code, I'm sharing it to the community.
About my merge request, I prefer that you allow the developer to override the pipeline stage of cesium, this would be great for expert developer that want to do something different. But it's not my project, it's just my opinion. 

here the code: 

```
float u_earthRadius = 6366707.;
float u_offset = 200.;

const vec3 UP_VECTOR_2D = vec3(0, 0, 1);

bool isMode3D(){
    return czm_morphTime == 1.;
}

bool isNotMode3D(){
    return czm_morphTime != 1.;
}

bool isMode2D(){
    return czm_sceneMode == czm_sceneMode2D;
}

vec4 geometryStage(inout ProcessedAttributes attributes, mat4 modelView, mat3 normal)
{
    vec3 positionMC = attributes.positionMC;
    if(czm_sceneMode == czm_sceneMode3D){
        v_positionEC = (modelView * vec4(positionMC, 1.0)).xyz;
    }else{
        // needed for 2D to have zxy
        v_positionEC = (czm_view * vec4(positionMC.zxy, 1.0)).xyz;
    }
    vec4 computedPosition = czm_projection * vec4(v_positionEC, 1.0);


    setDynamicVaryings(attributes);

    return computedPosition;
}

vec3 projectToGeographic(float lat, float lon, float altitude){
    float radiusX = u_earthRadius;
    float radiusY = u_earthRadius;
    float x = radiusX * lon;
    float y = radiusY * lat;

    return vec3(x, y, altitude);
}

vec3 project(vec3 p_ecef, float altitude){
    if (isMode3D()){
        return (u_earthRadius + u_offset + altitude) * normalize(p_ecef);
    }

    vec3 normal = normalize(p_ecef);
    float lat = asin(normal.y);
    float lon = atan(normal.x, normal.z);

    // 2D/2.5D mode :
    vec3 pos = projectToGeographic( lat, lon, u_offset + altitude);

    if (isMode2D()){
        pos.z = 0.;
    }
    return pos;
}

vec3 calculatePositionMC(vec3 position)
{
    float altitude = length(position) - u_earthRadius;
    return project(position, altitude);
}

void vertexMain(VertexInput vsInput, inout czm_modelVertexOutput vsOutput) {
  vsOutput.positionMC = calculatePositionMC(vsInput.attributes.positionMC);
}
```

javagl:
The code that is involved here may be 0.001% of the CesiumJS codebase. And it is certainly one of the more math-heavy areas (and as usual, painfully undocumented). I can look at this shader code snippet, and have no idea what it does, and (important:) in how far the thing that it does differs from what CesiumJS is _currently_ doing. But it is important to _thoroughly understand_ this. I think that there currently is nobody who could review and approve or reject such a change "just so". Everybody would have to allocate several days(!) for digging through this (and related) code, to sufficiently understand the changes. (At least, that's what I would have to do).

So again, starting at the top:

It's probably not easy to achieve the shape that the example GLB should have in 2D. I'm actually pretty sure that this is not possible _at all_, because the input triangle mesh has a certain topology which can not be converted into the desired 2D shape, but maybe I'm overlooking some magic here.

Can you post a screenshot of what that GLB (that covers the north pole) looks like in 2D when your suggested changes are applied?

mickae1:
> The code that is involved here may be 0.001% of the CesiumJS codebase. And it is certainly one of the more math-heavy areas (and as usual, painfully undocumented). I can look at this shader code snippet, and have no idea what it does, and (important:) in how far the thing that it does differs from what CesiumJS is _currently_ doing. But it is important to _thoroughly understand_ this. I think that there currently is nobody who could review and approve or reject such a change "just so". Everybody would have to allocate several days(!) for digging through this (and related) code, to sufficiently understand the changes. (At least, that's what I would have to do).
> 
> So again, starting at the top:
> 
> It's probably not easy to achieve the shape that the example GLB should have in 2D. I'm actually pretty sure that this is not possible _at all_, because the input triangle mesh has a certain topology which can not be converted into the desired 2D shape, but maybe I'm overlooking some magic here.
> 
> Can you post a screenshot of what that GLB (that covers the north pole) looks like in 2D when your suggested changes are applied?

I'm little sad that you don't believe me, but you want proof, you will have proof : 


3D projection : 
image



2D projection : 
image
( and this is exactly this shape i was waiting for )

I also tried with an another shape : 
3D projection : 
image


2D projection  : 
image

And I repeat, i'm just doing exactly what cesium does in cpu, my calculation is more simple because i'm using a spherical earth. 

But @mzschwartz5 said that we could use : 
https://github.com/CesiumGS/cesium/blob/c8ae7b0133236022ca0c3bc0203df37f94e73b9f/packages/engine/Source/Shaders/Voxels/convertUvToEllipsoid.glsl#L23-L27

to calculate lat, lon, altitude from ecef points.


I'm not asking the Cesium team to do my job, just for them to make their API more flexible for developers who want to do something out of the box. 
Right now, we have to respect the pipeline, but letting us set gl_Position ourselves would be awesome.

javagl:
It looks like this model does not cover/cross the north pole. I'll probably create another test model and check the result. 

> About my merge request, I prefer that you allow the developer to override the pipeline stage of cesium, this would be great for expert developer that want to do something different. But it's not my project, it's just my opinion.

Eventually, that's not my call. Right now, there is the `CustomShader` infrastructure, but I don't think that there's any precedent for overriding a single, internal, undocumented shader function. And **iff** this option should be offered, we'd have to be very clear about the inputs and outputs (or 'signature') of that function.

mickae1:
> Eventually, that's not my call. Right now, there is the `CustomShader` infrastructure, but I don't think that there's any precedent for overriding a single, internal, undocumented shader function. And **iff** this option should be offered, we'd have to be very clear about the inputs and outputs (or 'signature') of that function.

This is exactly what i've used : 

```
const customShader = new Cesium.CustomShader({
          lightingModel: Cesium.LightingModel.UNLIT,
          fragmentShaderText: customFS,
          vertexShaderText: customVS,          
          varyings: {
            v_show: Cesium.VaryingType.FLOAT
          }
});

    url: "../../SampleData/models/test.glb",
   customGeometryStageVS: "",
    modelMatrix: Cesium.Matrix4.IDENTITY, 
    asynchronous: false,
    customShader: customShader,
  });
```

You still need to adhere to the Cesium pipeline, which requires setting vsOutput.positionMC in the vertexMain(VertexInput vsInput, inout czm_modelVertexOutput vsOutput) function. I would prefer to avoid setting vsOutput.positionMC and instead set gl_Position directly, as this would solve my problem with the matrix not being set correctly in 2D/2.5D mode.

However, this would add complexity to the Cesium codebase. Perhaps the better approach is to fix the matrix configuration issue itself, possibly by detecting when it is an identity matrix.

Issue #12867, opened 2025-09-03
Build process tries to create directory outside of Cesium directory


UniquePanda:
### What happened?

When running `npm run build` inside of the `cesium` root directory, the build process tries to create a directory two levels above this root directory.
This is very likely not intended, because the script (`packages/sandcastle/scripts/buildGallery.js`) wants to create `../../Apps` and it was probably assumed that this will refer to `packages/sandcastle/Apps`.
However, the build process seems to use the repo's root directory for the relative path.

I also tried to clone the repo into a directory structure of `a/b/cesium` and indeed, after the build process ran, there was an `Apps` directory inside of the `a` directory.

The whole error is the following:
```
Error: EACCES: permission denied, mkdir '../../Apps'
    at async mkdir (node:internal/fs/promises:857:10)
    at async buildGalleryList (file:///home/x/cesium/packages/sandcastle/scripts/buildGallery.js:266:3)
    at async createGalleryList (file:///home/x/cesium/scripts/build.js:635:3)
    at async Promise.all (index 1)
    at async buildCesium (file:///home/x/cesium/scripts/build.js:1190:3)
    at async generateDevelopmentBuild (file:///home/x/cesium/server.js:64:20)
    at async file:///home/x/cesium/server.js:89:16 {
  errno: -13,
  code: 'EACCES',
  syscall: 'mkdir',
  path: '../../Apps'
}
```

### Reproduction steps

(Simply follow the build guide) 
1. Clone the repo into a directory that is not nested in multiple directories you can change (e.g. simply into your home directory)
2. Run `npm install`
3. Run `npm run build`
4. You should get an error (`Error: EACCES: permission denied, mkdir '../../Apps'`)

### Environment

CesiumJS Version: 1.133.0
Operating System: Ubuntu 24.04.2
Node.js version: v22.14.0

jjspace:
Apologies @UniquePanda. A majority of my testing time while developing the new sandcastle was inside the package directory itself and all my projects are nested deeper than the root so I never noticed this. Simple fix with the vite config. Will have a PR for it shortly.

In the meantime if you want to fix yours change the `vite.config.app.ts` file to set the `outDir` like below:

```ts
  const __dirname = dirname(fileURLToPath(import.meta.url));
  config.build = {
    ...config.build,
    outDir: join(__dirname, "../../Apps/Sandcastle2"),
  };
```

Issue #12866, opened 2025-09-03
The build-sandcastle command fails during build


Amethyst-yu:
### What happened?

Briefly describe the issue. Example: "The `build-sandcastle` command fails during build.

<img width="500" height="1280" alt="Image" src="https://github.com/user-attachments/assets/dad90650-3c45-4036-857c-1a4514f45f48" />

### Reproduction steps

1.  Clone the repo: `git clone https://github.com/CesiumGS/cesium.git`
2.  Run `npm install`
3.  Run `npm run build-sandcastle`
4.  See error

### Environment

Browser:
CesiumJS Version: v1.333
Operating System:window10

UniquePanda:
Actually (at least for me), even when simply running `npm run build` it fails with the following error:
`Error: EACCES: permission denied, mkdir '../../Apps'`
I also cloned the repo completely new and simply followed the build guide.

Edit: I just tried moving Cesium two directories deeper. Then it works. So this problem is probably related to how node is resolving relative paths.

ruomulight:
> Actually (at least for me), even when simply running `npm run build` it fails with the following error: `Error: EACCES: permission denied, mkdir '../../Apps'` I also cloned the repo completely new and simply followed the build guide.
> 
> Edit: I just tried moving Cesium two directories deeper. Then it works. So this problem is probably related to how node is resolving relative paths.

On which system are you running this?

UniquePanda:
> On which system are you running this?

On Ubuntu linux

ruomulight:
> > On which system are you running this?
> 
> On Ubuntu linux

This is probably a Linux permission issue; it requires you to grant it write permissions.

UniquePanda:
> This is probably a Linux permission issue; it requires you to grant it write permissions.

Not really. I think the script wants to write into the repository directory but instead it tries write into a directory which is two directories above the repo directory.
So yeah, I could give it write permissions to the directory two levels higher, but this would require to give it write permissions to the root directory of my OS which is probably not intended. :D

When I moved Cesium into two subdirectories (`a/b/Cesium`) it worked fine, but the build script created a `Apps` directory in the `a` directory. I'm pretty sure this is not the desired behavior.

javagl:
Yep, that's broken. 

[...]

PR #12869, opened 2025-09-03
Build sandcastle to relative paths

jjspace:
# Description

Annoyingly it sees that running an `npm` command on a workspace doesn't change the working directory. On top of that `vite` doesn't guarantee paths like the `outDir` are relative to the config file itself. Together these points mean that running `npm run build-sandcastle` can try and create files outside the repo directory which is a big no-no.

This PR updates the output paths for all vite configs that needed it to always be relative to the config file itself. This should mean they're always put where we expect them to be in the repo itself.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12867

## Testing plan

- Run `npm run build-sandcastle` and make sure files show up in `Apps/Sandcastle2` and _not_ outside the project directory

# Author checklist

### DIFF
--- a/packages/sandcastle/vite.config.app.ts
+++ b/packages/sandcastle/vite.config.app.ts
@@ -1,7 +1,10 @@
 import { defineConfig, UserConfig } from "vite";
 import { viteStaticCopy } from "vite-plugin-static-copy";
-
+import { dirname, join } from "path";
+import { fileURLToPath } from "url";
 import baseConfig, { cesiumPathReplace } from "./vite.config.ts";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
 
 export default defineConfig(() => {
   const cesiumBaseUrl = "/Build/CesiumUnminified";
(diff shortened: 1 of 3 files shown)


javagl:
There's likely an overlap in terms of the affected code paths to what could be a fix to https://github.com/CesiumGS/cesium/issues/12866 (just for awareness, maybe it can be two PRs...)

jjspace:
@javagl they're similar but distinct issues as far as I understand so I'm gonna tackle them independently. This is purely about the vite build paths regardless of the windows v linux things

ggetz:
Are you sure this fixes https://github.com/CesiumGS/cesium/issues/12867? The error stack appears to be caused by `buildGalleryList`, and I think this was something I overlooked.

I did some cleanup in https://github.com/CesiumGS/cesium/pull/12874 that appears to fix the reported issue without modifying the vite configs.

Perhaps in this branch, the updated vite config updates are masking the underlying issue by affecting the working directory as a side effect? Even so, I think my branch fixes the error at the source. Plus it fixes the gallery file watcher, which was not working correctly.

UniquePanda:
I can confirm, that this PR (at least for me) only fixes the build process for `npm run build-sandcastle`.
Running `npm start` or `npm run build` will still result in the same errors as before.
PR #12874 from @ggetz seems to fix all of the issues. :)

jjspace:
Yes, this PR was a hasty fix towards a symptom not the real cause of the issues. I meant to close it yesterday in favor of @ggetz pr

PR #12874, opened 2025-09-04
Sandcastle gallery build path fix

ggetz:
# Description

Cleans up some rough edges that I overlooked in https://github.com/CesiumGS/cesium/pull/12854.

- Fixes build paths so that output is put in the right place for both root cesium development as well as sandcastle development.
- Fixes the `gallery` file watcher and rebuild which is part of `npm start`.
- Does a bit of cleanup to `build.js` while I was in there: I noticed that it was running some async functions as a side effect. That means they were running even when importing a totally unrelated function.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12867

## Testing plan

1. From a clean repository state, run `npm install` and `npm run build`. The gallery files should be built to `Apps/Sandcastle2`, not a parent directory as in the original bug report.
2. Run `npm start`. Update a gallery file in `packages/sandcastle/gallery` and be sure the updates are rebuilt and reflected in `http://localhost:8080/Apps/Sandcastle2/index.html`.
3. Change your working directory to `packages/sandcastle`. Ensure that running `npm run dev` build the gallery files to the `packages/sandcastle/public` directory, and that the app loads examples as expected.

# Author checklist

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,10 +1,10 @@
-import child_process from "child_process";
-import { existsSync, readFileSync, statSync } from "fs";
-import { readFile, writeFile } from "fs/promises";
-import { EOL } from "os";
-import path from "path";
-import { createRequire } from "module";
-import { finished } from "stream/promises";
+import child_process from "node:child_process";
+import { existsSync, statSync } from "node:fs";
+import { readFile, writeFile } from "node:fs/promises";
+import { EOL } from "node:os";
+import path from "node:path";
+import { finished } from "node:stream/promises";
+import { fileURLToPath } from "node:url";
 
 import esbuild from "esbuild";
 import { globby } from "globby";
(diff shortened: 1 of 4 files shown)


jjhembd:
On my WSL Ubuntu build, this PR fixes the issue from https://github.com/CesiumGS/cesium/issues/12867:

- No spurious `Apps` directory created outside the `cesium` folder
- Working "Sandcastle2" link on the `localhost` page

jjspace:
Actually @ggetz I went to get a picture of the slow gallery list build time I mentioned in person and realized there's an issue with the local server rebuild. If you spam save on a file (or even just hit it twice accidentally before it finishes)  multiple builds get kicked off at the same time and they conflict. Sometimes it seems to catch it, sometimes it crashes fully.



1. I think we should revisit and speed this up. I suspect it's mostly pagefind so maybe we can add an option to skip that if a `yaml` file wasn't changed?
2. I think we may want to add a better catch or prevent multiple runs at once or something.

scripts/build.js
```diff
@@ -606,17 +610,31 @@ const externalResolvePlugin = {
...
+  const options = config.default;
+  return {
+    ...options,
+    configPath,
+  };
+}
+
+/**
+ * Indexes Sandcastle gallery files and writes gallery files to the configured Sandcastle output directory.
+ * @param {boolean} [includeDevelopmen=true] true if gallery items flagged as development should be included.
```
jjspace:
```suggestion
 * @param {boolean} [includeDevelopment=true] true if gallery items flagged as development should be included.
```

ggetz:
> there's an issue with the local server rebuild. If you spam save on a file (or even just hit it twice accidentally before it finishes) multiple builds get kicked off at the same time and they conflict. Sometimes it seems to catch it, sometimes it crashes fully.

I'll take a look and see if we can make a small change here to prevent the crash.

I do agree that outside of this PR, we can take a pass and see if there's something we can do to speed up the overall build step.

ggetz:
@jjspace Updated! 

- Added a small timeout to prevent multiple rebuilds when saving in quick succession
- Cleaned up error handling so that if the build function does throw, such as when missing a required property in `sandcastle.yml`, it is logged but does not stop the watcher or server.

Issue #12872, opened 2025-09-03
Point cloud display regression


angrycat9000:
### What happened?

A point cloud that loaded in 1.132 fails to load in 1.133.

**1.132**
<img width="1135" height="636" alt="Image" src="https://github.com/user-attachments/assets/5600306e-c675-460b-875a-4c72f8e16769" />

**1.133**
<img width="1144" height="640" alt="Image" src="https://github.com/user-attachments/assets/7ced4dbd-42b2-41ab-af78-260403a13e0f" />

Errors from console:

> RuntimeError: Failed to load model: https://assets.ion.cesium.com/us-east-1/3626944/1/3.pnts?v=1
> Failed to load Draco pnts
> Property PointSourceID requires a batch table binary.
> Original stack:
> Original stack:
> Error
>     at new RuntimeError (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:11182:13)
>     at transcodeBinaryPropertiesAsPropertyAttributes (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:97013:15)
>     at parseBatchTable (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:96869:32)
>     at makeStructuralMetadata (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:119140:14)
>     at makeComponents (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:119188:37)
>     at https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:118836:7
> Handler stack:
> Error
>     at new RuntimeError (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:11182:13)
>     at ResourceLoader.getError (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:74524:26)
>     at https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:118843:30
> Handler stack:
> Error
>     at new RuntimeError (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:11182:13)
>     at ModelUtility.getError (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:80267:26)
>     at Model.update (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:121225:51)
>     at Model3DTileContent.update (https://sandcastle.cesium.com/CesiumUnminified/Cesium.js:122288:11)
[...]

Issue #12452, opened 2025-01-26
Minor updates for the Coding Guide


javagl:
The [current `CodingGuide`](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/Documentation/Contributors/CodingGuide/README.md) contains a few recommendations that may have to be updated.

One specific point that was discussed as [part of a recent PR](https://github.com/CesiumGS/cesium/pull/12361#issuecomment-2613463937) is whether the Coding Guide should contain a recommendation like

> Don't use `Array.push.apply()` for large arrays. Use `addAllToArray` or the `[...spread]` syntax instead
> <sup>(Maybe with details about where the spread syntax should also be avoided...)</sup>

I looked at the the Coding Guide, thinking about where something like this might be added. But there are two and a half issues with that:

1. There are some plainly outdated statements in the Coding Guide
2. Some recommendations there are obsolete, because they already are enforced with linting rules
  2½. The specific case of `push.apply` is already covered with a newly introduced linting rule

I think that _very_ roughly speaking, the Coding Guide should not repeat things that are covered by the linter. (One could argue that it could _explain_ **why** these rules exist, but that's a slippery slope, and might bloat it up to a point where nobody is reading it any more...)

Some preliminary points that I noticed and that might have to be updated accordingly, starting at the the [Basic Code Construction](https://github.com/CesiumGS/cesium/blob/dc667cd6decdaf14079038bc32773f421e585724/Documentation/Contributors/CodingGuide/README.md#basic-code-construction) section:

- "Cesium uses JavaScript's strict mode, so each module (file) contains `"use strict"`
  - That's as wrong as it can be. **No** file contains this anymore 🙂 
- "To avoid type coercion..."
  - This is covered by [eqeqeq](https://eslint.org/docs/latest/rules/eqeqeq)
- "let and const variables have block-level scope" 
  - This is covered by [no-use-before-define](https://eslint.org/docs/latest/rules/no-use-before-define)
- "Modern language features may provide handy shortcuts and cleaner syntax, ..."
  - This is a bit vague. But the `push.apply`-to-`[...spread]` (with caveats) might fit here...
- "Do not use an unnecessary else block at the end of a function..."
  - This is covered by [no-else-return](https://eslint.org/docs/latest/rules/no-else-return)

Some further (potential) updates are connected to open issues or pending PRs - just a quick summary:

- "Many Cesium functions take an options parameter..."
[...]

Issue #12196, opened 2024-09-17
Update Coding Guide with notes on destructuring assignment


jjhembd:
The current [Coding Guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md) does not discuss [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

Since https://github.com/CesiumGS/cesium/issues/10486, many ES6 patterns have naturally found their way into the code, including destructuring assignments. For example,

```javascript
const {
  option1,
  option2,
  option3 = 3.0,
} = options;
```

This syntax is often both more concise and easier to read. This raises 2 questions for the Coding Guide:
- Should we recommend it?
- How do we clarify where destructuring defaults can be used, and where it is necessary to use `defaultValue`? The destructuring default syntax will replace `undefined` with the given default, but will pass `null` through.

jjspace:
Personally I'm a fan of destructuring everywhere it makes sense. I don't think it's main benefit is default values though. I'll also mention https://github.com/CesiumGS/cesium/issues/11674 just to crosslink related issues.

ggetz:
Are there any performance implications to using destructuring assignment everwhere?

javagl:
Regarding the handling of `null`, compared to `defaultValue`: There are very few places where `null` is used at all. So that may not be critical.

Regarding the performance: There are some claims, but ... these are soooo 2018 (i.e. it's hard to measure that objectively across browsers and JITs).

Regarding the eslint:
For now, this was rather _asking_ whether this should become part of the coding guide. **Iff** it should become the recommended way, then yes: Everything that can be covered with a linter should not be mentioned in the coding guide. (Except, maybe, caveats like the `null` stuff, TBD). 

But... highly subjective: I'm not sure about enforcing that particular rule. Destructuring assignment can be convenient when you really just have an object with 10 properties and want to pull out these 10 properties as local variables. But in some cases, the syntax is _really_ obscure. For example, consider something like
```
let bar;
...
bar = object.bar;
```
Now, that innocent, bread-and-butter, self-explanatory last line would be flagged by eslint. It should actually be written as
`({ bar } = object);`
(Yeah, really... 🤪 )

jjspace:
> Now, that innocent, bread-and-butter, self-explanatory last line would be flagged by eslint. It should actually be written as
> `({ bar } = object);`
[...]

PR #12893, opened 2025-09-12
Fix pnts draco loading

javagl:
# Description

Under certain conditions, the point cloud tiler generated invalid PNTS files. (This should be fixed in the meantime). This was not noticed until recently - it was a "fail silent", meaning that the resulting data did just not contain the metadata that it "should" have contained, because of unclear and inconsistent sequences of `if (defined(something)) doSomething()` patterns. After the change from https://github.com/CesiumGS/cesium/pull/12806 , these files caused a crash of CesiumJS. Details about what exactly was invalid are described in https://github.com/CesiumGS/cesium/issues/12872#issuecomment-3263757882  

This PR changes the part that loads this data: 

After the PNTS data is loaded, it checks whether there are 'binary body reference' properties in the batch table, there is no 'batch table binary', and these properties have _not_ been resolved as draco properties. If such properties are found, then they are removed from the batch table JSON, and a `oneTimeWarning` is printed (for each of them). 

## Issue number and link

This is a **workaround** for https://github.com/CesiumGS/cesium/issues/12872 . Strictly speaking, it was not a CesiumJS issue, but an issue of invalid input data. This PR just changes the logic to handle this invalid data more gracefully, without crashing.

## Testing plan

One part of the test should be to try and load the data that was linked in the issue. 

I extracted **one** PNTS file from this data set to create a new spec case. This should be OK: The input data for that asset is said to be Apache-licensed. It is converted to PNTS by ion. I assume that the result still falls under the Apache license.

The new spec data is at https://github.com/CesiumGS/cesium/tree/ba53f9f4d4b13c106b250cd91cb40a357c659433/Specs/Data/Cesium3DTiles/PointCloud/PointCloudDracoInvalid , including a README.md explaining why there is invalid data in the specs.

(An aside: The commit at https://github.com/CesiumGS/cesium/commit/a465ab2fb0e0a65e3ac504cc5d1d11b1424eddf4 just removes a duplicated function. When initially reading the error message, it was not even clear from which of these functions the error was emitted...)

# Author checklist


packages/engine/Source/Scene/BatchTableHierarchy.js
```diff
@@ -127,7 +127,7 @@ function initialize(hierarchy, hierarchyJson, binaryBody) {
   for (i = 0; i < classesLength; ++i) {
     const classInstancesLength = classes[i].length;
     const properties = classes[i].instances;
-    const binaryProperties = getBinaryProperties(
+    const binaryProperties = Cesium3DTileBatchTable.getBinaryProperties(
```
jjhembd, line 130: Nice deduplication!

packages/engine/Source/Scene/PntsParser.js
```diff
@@ -185,9 +186,73 @@ PntsParser.parse = function (arrayBuffer, byteOffset) {
...
+  );
+
+  // Collect the names of all binary body references (identified
+  // by the property having a `byteOffset`) that have not been
+  // resolved via the parsedContent.draco.batchTableProperties
+  const invalidBinaryBodyReferenceNames = [];
+  for (const name of Object.keys(batchTableJson)) {
+    const property = batchTableJson[name];
+    const byteOffset = property.byteOffset;
+    if (defined(byteOffset)) {
```
jjhembd, line 238: In the interest of [avoiding unnecessary variables](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#:~:text=Do%20not%20create%20a%20local%20variable%20that%20is%20used%20only%20once%20unless%20it%20significantly%20improves%20readability%2C%20e.g.%2C), would this be better as follows?
```javascript
if (defined(batchTableJson[name].byteOffset)) {
```

packages/engine/Source/Scene/PntsParser.js
```diff
@@ -185,9 +186,73 @@ PntsParser.parse = function (arrayBuffer, byteOffset) {
...
+ * a one-time warning will be printed.
+ *
+ * @param {object} parsedContent The parsed content
+ */
+function removeInvalidBinaryBodyReferences(parsedContent) {
+  const batchTableJson = parsedContent.batchTableJson;
+  if (!defined(batchTableJson)) {
+    return;
+  }
+  const batchTableBinary = parsedContent.batchTableBinary;
```
jjhembd, line 223: I realize we haven't actually agreed on agreed on a style preference for destructuring, but perhaps this is one of the places where [it makes sense](https://github.com/CesiumGS/cesium/issues/12196#issuecomment-2358880039).

```javascript
const { batchTableJson, batchTableBinary } = parsedContent;
if (!defined(batchTableJson) || defined(batchTableBinary)) {
  return;
}
```

jjhembd (approved):
Thanks @javagl, this looks good to me! I left a couple minor nitpicks on the style, which you could incorporate (or not) as you see fit.
I don't think the CI failure is related to this PR--it's happening in `main` too.

packages/engine/Source/Scene/PntsParser.js
javagl: I'm aware of that rule in the coding guide. Fortunately, a "significantly improved readability" is subjective 😁 

But more seriously: 

I think that _readability_ is one thing. But _debuggability_ is equally important, There are faaar too many 'train wrecks' in the code, and when the flag in an  
`if (someObject.someArray[index].someComplexThing.someFlag)` 
does not have the right value, I think that it can help to have that `someComplexThing` easily inspectable in the local variables. (Maybe there's some magic that makes this easier that I haven't discovered yet).

In this case, there's not sooo much of a trade-off. But I'd like to point out that this "detection of a binary property" was taken (verbatim) from the de-duplicated `getBinaryProperties` function (given that there is no `isBinary()` function or such). If this is a strong change request, I'll change it. (And use "git blame" to find out who to blame 😉 )

packages/engine/Source/Scene/PntsParser.js
javagl: Similar to the above: I don't have a _strong_ opinion in this case. The style resembles existing code everywhere. This doesn't mean that we cannot deviate from the existing style. (Like...with adding JSDoc 😁 ). I can change it tomorrow (hopefully, I'll be back online somewhere in the afternoon)

javagl: (Before the possible network disconnect) :

 I was about to change this, and searched for 
`const batchTableBinary = parsedContent.batchTableBinary;`
to find that place. It brought _two_ results, and one of them is https://github.com/CesiumGS/cesium/blob/d7329e25a7b1900c695f703480ff8e538b97af1b/packages/engine/Source/Scene/Model/PntsLoader.js#L549 

There are quite a few "We should..., We should..., We should..." issues. Opened years ago, but never addressed. Often, these are about fundamental flaws (with the status quo being: "Yeah, it's broken, but too complicated and too 'baked-in' to ever be fixed"). Often these are about _cleanups_ in one way or another. The destructuring assignment is a very (purely) _syntactical_ one - compared to "Don't access private variables", "Use `async/await`", or "Use `class`". But I think that what [I said in the dicsussion about `class`](https://github.com/CesiumGS/cesium/issues/8359#issuecomment-2583787511) applies here in a similar way: It could be done incrementally, ~"for new code (or code that is heavily changed", or once, in a "big bang".

I left a comment at https://github.com/CesiumGS/cesium/issues/12196#issuecomment-3297573970 , boiling down to "Are we doing this now?" 🙂

Specs/Data/Cesium3DTiles/PointCloud/PointCloudDracoInvalid/pointCloudDracoInvalid.pnts
ggetz, line 1: I'm being sensitive to the overall size of testing data recently because I've noticed our shipped zip file (which includes the unit tests and related data) has has [noticeably increased in size since the latest release](https://github.com/CesiumGS/cesium/releases).

Is it possible to pare down the file size here while still reproducing the original issue?

packages/engine/Specs/Scene/Model/PntsLoaderSpec.js
```diff
@@ -538,6 +540,18 @@ describe(
       });
     });
 
+    it("loads PointCloudDracoInvalid without crashing", async function () {
```
ggetz, line 543: This is a total nitpick, but can we make the test name a bit more descriptive of the context?

Maybe something along the lines of the following.

```suggestion
    it("loads legacy point cloud metadata with Draco compression", async function () {
```

packages/engine/Source/Scene/PntsParser.js
ggetz: If we do want to avoid the potential train wreck, does the following make sense?

`defined(batchTableJson?.[name]?.byteOffset)`

Regardless, I'm generally in favor of minor restructuring to improve code structure ad/or maintainability. So if a `isBinary` function is warranted we can add it.

packages/engine/Source/Scene/PntsParser.js
ggetz: Destructuring is all good in my book. But, as with all syntactic sugar, let's be wary of when we apply it such that it improves code readability rather than just use the most concise code possible for the sake of it.

For instance, I could see using the destructuring in the function declaration itself being helpful for devs to understand what needs to be passed into the function:

```js
function removeInvalidBinaryBodyReferences({ batchTableJson, batchTableBinary, draco }) {
...
```

ggetz:
Also please update `CHANGES.md` with a summary of the fixed issue.

javagl:
I've skimmed over the comments. I just arrived back home, and it's 3:30 am here, so I'll skip comments about the reason for the original bug, the regression, the analysis process, and the process for the additional fix. But have to say that this requires a form of patience that I'm running out of. I can respond with further comments to the inlined comments here, but unless there are clear "Change it like **that**" ones, I'll consider to just hit merge to get this out.

packages/engine/Source/Scene/PntsParser.js
javagl: There are different options. I "chose" this one (and the _exact_ code was also used elsewhere). If something should be changed here, then criteria for deciding about whether to use local variables should be added to the coding guide. Details can be discussed in https://github.com/CesiumGS/cesium/issues/12452 or a dedicated issue. Resolving this for now.

packages/engine/Source/Scene/PntsParser.js
javagl: If there should be any rules about the use or non-use of destructuring assignment, then they can be discussed at https://github.com/CesiumGS/cesium/issues/12196, but certainly not in this PR.

packages/engine/Specs/Scene/Model/PntsLoaderSpec.js
javagl: It is not legacy data. It is **invalid** data. There is no reason to surgarcoat that. The README.md that I added here makes this _explicit_.

Specs/Data/Cesium3DTiles/PointCloud/PointCloudDracoInvalid/pointCloudDracoInvalid.pnts
javagl: It certainly is possible. Of course, I don't have to right tooling to create invalid files. But I have the right tooling to create (valid) LAS files (for each version and with each point data record format), so I created a "unit cube" consisting of 9 points, uploaded it to ion, and used the result here. If I should reduce it to 2 points, just drop me a note.

This also fixed another issue, that is **waaay** more important than the file size, namely that the `tileset.json` here was originally invalid. It did not declare the `3DTILES_draco_point_compression` extension in the `extensionsUsed`. Sure, something like this should be caught during testing and validation, but ... here we are.

javagl:
This was updated. The inlined comments are tracked in other issues. It should be ready for another review.

ggetz:
Thanks for the updates @javagl! Looks good to go. Thanks for redirecting further conversation to existing issues.

ggetz (approved):

Issue #12719, opened 2025-07-04
Performance issue with Label (num characters) and CLAMP_TO_GROUND


alarkbentley:
### What happened?

Adding entities in Cesium.js using HeightReference = CLAMP_TO_GROUND is generally very costly for performance. The problem seems worse with Labels than Billboards, and strangely enough the performance decreases with not only the number of entities but primarily the length of the text labels (number of characters).

Rendering 1000 labels with 5 characters works quite well. But if I increase the number of characters to 20 or 100 the application becomes unusable. This only an issue when HeightReference is used to clamp the entities to underlying terrain.

As the letters are not individually placed relative to ground I see no reason why the number of characters should effect performance here.

### Reproduction steps

1. Make sure "Clamp to Ground" is active
2. Enable "1000 Entities" (should work fine, 60fps)
3. Enable "100 Characters" (unusable, browser hangs)

![Image](https://github.com/user-attachments/assets/ecad75ca-7c51-4c07-b011-c2c805c32b93)

### Sandcastle example

[...]

PR #12896, opened 2025-09-12
Event performance

mzschwartz5:
# Description

The Cesium `Event` class uses an array for its internal storage of listeners. To remove a listener, it must iterate over the array to find the matching object. This is unnecessarily slow. If we used a `Map`, instead, we would have constant-time access.

This PR switches the internal data structure of `Event.js` to a `Map`. The key is the `listener`, and the `value` is a set of scopes (i.e. the `this` context to run the listener in. If 5 billboards want to listen to an `updateClamping` event, they would all register their same listener function, with different scopes). 

Technically, this change should also reduce memory usage. In the billboard example, rather than storing the same listener in an array 5 times (each with a different scope), we only need to store it once (but still need to store 5 scopes). 

## Issue number and link

Partially addresses #12719. 

## Testing plan

Testing on the [sandcastle](https://dev-sandcastle.cesium.com/#c=rVdtb+M2DP4rQj45uEyx/O5cW6xI012wXlo02Q1DUxSKrSTaObYhy+2yof99lPzSJO31eusKFLYfkXxIShSZKEsLie45e2ACHaOUPaAhK3i5wV80Zsw7kf4eZqmkPGVi3umhf+YpQpIJAcigUZhV33gpss3vmUjiGjC6vXn62P04TyNNVkQsZcBVkWL9CYvNckwlhVX6QLlERvVYMhmtwZW1lHkx6PdTtqFUbGmKV1yuywXmWZ/mvL+homA8SdhdnvEC/1lk6bzT7eoXo6tJEibRmvHVWl6zJRMsjZQvdQif9hfw8OL089Xd7PLul+vL3yZnHyt1lkouOStAj5g1Bv+QDwW5mmZZppHkWYqSjMajWsHoVplTChxElTLSAeNlJkYUYjS4ZJsuOj6pJBHiS2SA8MkTa7dZUn+CyVKkHyvgsXpUecyzgmsP2uiGVEh4o6mt9+iMrQQDpzDGihQ3ClWelKF6hxpmTOPYaLl5PEBaj8e9BmtMtEBCFywZ7DoMx+YvOUDzDp13sGA5o9Kok9ft7Ygd7NHgENiVjXlBFwk7Y7lcz1ghz3ghqVaalJsFRHB1OR3Pxl9Gd+PJ+Xgynv3Ratcp08dTR/Xhg37TwOPTqVSpULt7o0OJgYSnVIU6QDeOZ4e+HeLQsr3QCTynhxw7dD0bO04QuG7gu6GCHMe2fRw4lmkFlmXfaidiLlhUG/rJxAGsBCaxQq0ZWj0EoBmGNrFsy/TBvunaPWRix1NSxCUO0FpOZazMwQqsAaXp+PAgru9axFMKph2Yoe9Ynum7ygdtJQh83yeeb9mBF1geUWYe6/o+DNLx3NDEFhDbnm+DZw744rqYED90Lds1AxWkB3IE265jEtfyvBeD9C3HDy1iuZ4XBCQIqyC9wLdM4oEzPrjo2xr1IAs2CRzfhCVwcidOvUoC4vi27dp26HheZcgnsCNWaDo2sQkJVZw+JNVyAy8MzdA03SrO2/1KXSbbWWaoja4rbPeCwhHdMAF1qoXq87yXoZ2r86nQtDm8I3dj3vaeYeQFzLptyiETHCpQo+iplNqUotepG7En4hYhz5AnUkjx64bLvLUIr+TpFWzUNVWbiktRp8isEUYLnq7O68S3DWS0B+OL8WR0en03uZyMevslOaVpHNFCJkxdSbMsSxZUfGZpadzUnam6YoYJ3eRIZugXkZVpDJ1LrUI9swQCHiBabNMItQfAaG/WdzQIfevt3fnVfaILa8+7SZayZy69xxmVqe94cFtd7m9NIDFRY+XtyTtsj2/PCDHfzffDhK8wfpfL/H+z7aLhmgoaqV749vD3J4+3x26Z76SzfnxzXyF8lYr8h1QLVsCYdfzc8OFUI9gmu2enSaLtPVZNoe0FBVxzCt9rBjFblKvpOns4F9AWiismpgzmhBjopCirafaeCrQGd5LDwXoawdiVTnMasdE9ePGpEjKaPgOahWKslTGEMU7zUp7qMIw2niy9YEs5THj01VARbMBWmzvFLugWmPfa14rJK5C/pttWZW/ua1QBa1VXSbZgOFc0YLFXoY2wmo4yyHaSrYw9pkOr3xRsu893Jcu82fBvpXK2zRm+GJ3P7oYX4+Gvc22z0+scFXKbsJNm6PuZb/JMSFSKBKbfPgyxeUKhu/UXZfSVSRwVRePMUX9X9Sjm9zD4Hr/wowhFCS0KWFmWSTLlf8PVfnLUB/lnquoIQ5+7vGcioVsltiYnFxUIs/hRHz5f1pTV1XFg+V8) from the above issue. With 1000 labels, 100 characters each, going from clamped to unclamped hangs the browser for 25-30 seconds. With this PR, it goes to about 5-10 seconds (like I said... other issues to address as well). 

I'm open to other ideas for testing, as this is a very fundamental class. Ideally, though, the behavior hasn't changed at all - only the underlying data structure. (Which is why there's no need to update unit tests; they should not be testing internal details anyway).

# Author checklist

- ~[ ] I have added or updated unit tests to ensure consistent code coverage~

### DIFF
--- a/packages/sandcastle/sandcastle.config.js
+++ b/packages/sandcastle/sandcastle.config.js
@@ -3,7 +3,7 @@ const config = {
   sourceUrl: "https://github.com/CesiumGS/cesium/blob/main/packages/sandcastle",
   publicDir: "./public",
   gallery: {
-    files: ["gallery/**/*"],
+    files: ["gallery"],
     searchOptions: {
       excerptLength: 10,
       ranking: {
--- a/packages/sandcastle/scripts/buildGallery.js
+++ b/packages/sandcastle/scripts/buildGallery.js
@@ -15,7 +15,7 @@ import createGalleryRecord from "./createGalleryRecord.js";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const defaultRootDirectory = join(__dirname, "..");
 const defaultPublicDirectory = "./public";
-const defaultGalleryFiles = ["gallery/**/*"];
+const defaultGalleryFiles = ["gallery"];
 const defaultThumbnailPath = "images/placeholder-thumbnail.jpg";
 const requiredMetadataKeys = ["title", "description"];
 const galleryItemConfig = /sandcastle\.(yml|yaml)/;
@@ -118,7 +118,7 @@ export async function buildGalleryList(options = {}) {
   };
 
   const galleryFiles = await globby(
-    galleryFilesPattern.map((pattern) => join(rootDirectory, pattern)),
+    galleryFilesPattern.map((pattern) => join(rootDirectory, pattern, "**/*")),
   );
   const yamlFiles = galleryFiles.filter((path) =>
     basename(path).match(galleryItemConfig),
@@ -261,7 +261,7 @@ export async function buildGalleryList(options = {}) {
   // regardless of if titles match the directory names
   output.entries.sort((a, b) => a.title.localeCompare(b.title));
 
-  const outputDirectory = join(publicDirectory, "gallery");
+  const outputDirectory = join(rootDirectory, publicDirectory, "gallery");
   await rimraf(outputDirectory);
   await mkdir(outputDirectory, { recursive: true });
 
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,10 +1,10 @@
-import child_process from "child_process";
-import { existsSync, readFileSync, statSync } from "fs";
-import { readFile, writeFile } from "fs/promises";
-import { EOL } from "os";
-import path from "path";
-import { createRequire } from "module";
-import { finished } from "stream/promises";
+import child_process from "node:child_process";
+import { existsSync, statSync } from "node:fs";
+import { readFile, writeFile } from "node:fs/promises";
+import { EOL } from "node:os";
+import path from "node:path";
+import { finished } from "node:stream/promises";
+import { fileURLToPath } from "node:url";
 
 import esbuild from "esbuild";
 import { globby } from "globby";
@@ -18,21 +18,23 @@ import { mkdirp } from "mkdirp";
 // This should match the scope of the dependencies of the root level package.json.
 const scope = "cesium";
 
-const require = createRequire(import.meta.url);
-const packageJson = require("../package.json");
-let version = packageJson.version;
-if (/\.0$/.test(version)) {
-  version = version.substring(0, version.length - 2);
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const projectRoot = path.join(__dirname, "..");
+const packageJsonPath = path.join(projectRoot, "package.json");
+
+async function getVersion() {
+  const data = await readFile(packageJsonPath, "utf8");
+  const { version } = JSON.parse(data);
+  return version;
 }
 
-const copyrightHeaderTemplate = readFileSync(
-  path.join("Source", "copyrightHeader.js"),
-  "utf8",
-);
-const combinedCopyrightHeader = copyrightHeaderTemplate.replace(
-  "${version}",
-  version,
-);
+async function getCopyrightHeader() {
+  const copyrightHeaderTemplate = await readFile(
+    path.join("Source", "copyrightHeader.js"),
+    "utf8",
+  );
+  return copyrightHeaderTemplate.replace("${version}", await getVersion());
+}
 
 function escapeCharacters(token) {
   return token.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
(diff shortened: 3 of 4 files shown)


javagl:
It's a pity that such a fundamental class has such a _severe_ flaw. (Of course, I once stumbled over this class and skimmed over it, but when there is no explicit task (or issue) to _review this with the goal of making it better_, one has to shrug it off). Are there already any ballpark numbers in how far this affects https://github.com/CesiumGS/cesium/issues/12719#issuecomment-3286707443 ?

mzschwartz5:
@javagl Yeah I was pretty surprised to learn that such a core class had such a flaw, as you put it. But, after talking with Josh, I have to cut a little slack to the original authors, because ~13 years ago, it seems there wasn't full, native, cross-browser support for Maps yet.

Ballpark numbers for how it affects that labels issue (all cases below were 1k labels, 100 characters each, going from clamped to unclamped):

1. Current release version: ~27 seconds
2. With new event class: ~5.5 seconds (flame graph is largely picking now, also updating custom data on quadtree tiles)
3. With new event class and octree picking: ~3.8 seconds (flame graph is 66% updating custom data, 33% picking now)

So, even if we improve the custom data part to 0 somehow, we're still looking at a ~1.3 second hang *just* for picking, even with octree. It's obviously much much better than 27 seconds, but not as great as I had been hoping, for only 1k labels. Could be much improved though if we treat each label as a unit rather than glyphs as units, when clamping and such.

mzschwartz5:
I'm realizing that a Map of listener-to-scope will not work. The listener and scope, together, form the unique key for an `Event`. To account for this, we should use a `Map` of listeners to `Set`s of scopes.

packages/engine/Source/Core/Event.js
```diff
@@ -22,10 +22,10 @@ import defined from "./defined.js";
  * evt.removeEventListener(MyObject.prototype.myListener);
  */
 function Event() {
-  this._listeners = [];
-  this._scopes = [];
+  this._listeners = new Map(); // listener => Set of scopes
```
ggetz: Totally optional, but this could be formalized with a type definition:

```suggestion
  /**
   * @type {Map<Listener,Set<object>>}
   * @private
   */
  this._listeners = new Map();
```

packages/engine/Source/Core/Event.js
```diff
@@ -128,29 +119,24 @@ function compareNumber(a, b) {
...
-
-  for (i = 0; i < length; i++) {
-    const listener = listeners[i];
-    if (defined(listener)) {
-      listeners[i].apply(scopes[i], arguments);
+  for (const [listener, scopes] of this._listeners.entries()) {
+    for (const scope of scopes) {
+      if (defined(listener)) {
+        listener.apply(scope, arguments);
+      }
```
ggetz: If the listener is undefined, we can skip iterating over the scopes at all, right?

```suggestion
  for (const [listener, scopes] of this._listeners.entries()) {
    if (!defined(listener)) {
      continue;
    }
    for (const scope of scopes) {
      listener.apply(scope, arguments);
    }
```

CHANGES.md
```diff
@@ -9,6 +9,7 @@
 - Materials loaded from type now respect submaterials present in the referenced material type. [#10566](https://github.com/CesiumGS/cesium/issues/10566)
 - Reverts `createImageBitmap` options update to continue support for older browsers [#12846](https://github.com/CesiumGS/cesium/issues/12846)
 - Fix flickering artifact in Gaussian splat models caused by incorrect sorting results. [#12662](https://github.com/CesiumGS/cesium/issues/12662)
+- Improved performance and reduced memory usage of Event class. [#12896](https://github.com/CesiumGS/cesium/pull/12896)
```
ggetz:
```suggestion
- Improved performance and reduced memory usage of `Event` class. [#12896](https://github.com/CesiumGS/cesium/pull/12896)
```

ggetz (changes requested):
Thanks @mzschwartz5! Fantastic to get such a fundamental performance bottleneck addressed with a fairly simple code change!

ggetz:
Please also merge in `main` when you can.

packages/engine/Source/Core/Event.js
mzschwartz5: Good point!

ggetz:
After merging in `main` with the known CI test fixes, it appears there are still some unit tests failures here. Would you mind checking them out @mzschwartz5?

ggetz:
Looking good! I did some various manual testing, both for clamping specifically and other features broadly, and I wasn't able to find any side effects of the change.

Thanks @mzschwartz5!

ggetz (approved):

PR #12884, opened 2025-09-09
Preserves subpixel precision on glyph alignment

mzschwartz5:
# Description

Currently, labels with small fonts suffer from poor glyph alignment. (At least, it's most noticeable with small fonts). The source of the issue is how we quantize these alignment values and pack them into vertex attributes. In doing so, we lose subpixel precision on alignment. 

To solve the issue, I introduced a scaling factor that's applied before packing (and un-applied after unpacking) to preserve some subpixel precision, at the cost of reducing the range of representable translation values. More detail in the [comment here.](https://github.com/CesiumGS/cesium/issues/8474#issuecomment-3268642867)

## Issue number and link

#8474

## Testing plan

NOTE: this issue is much more apparent on high-DPI screens, and as such should be tested on one.

[Sandcastle linked from issue](https://dev-sandcastle.cesium.com/#c=vVRRT9swEP4rVl6SSsEhbSliTStEAfHAtEmd9pSXa3JtrV3syHYCBfHfsRMQjGlahzbuyb6777vv7iy3oFkr8AY1mzGJN2yBRjQV/975orDorgslLQiJOhxMc5nL1qEIVkgX0gq7c9Ceg6O/CzQcyjK6zyVzVivjfEqyT8/kC9DWnUCO+Fqr6hw3GtFEB8dHPJ2M08nkOGajE35yNPQ2iHuerqAjeaL1ZvHWOk/4dSsISqR6KyCMX+Jrp9vHx4f1LbtCatGKAsI+4SGXD307S5BlAcYSet3flKIV6LPGWiWjMB067KUjCmO2bmThW4kG96809UPg3Zl3JWesh72UnP4u3xRA6ADpdD85w/G75HSw98n5g6Butn8v6M1K/tl8lhUQsaVH/V9Bh3y0p6TPWLpX/zGaJnvv7UxsPkLS096COMiM3RHOPeBUVLXSljWaIs4Ti1VN4L6EZNUUP9DywhjfQ5Y8Q7JStEyUszx48yHlASsIjHGRdUO0FHeYB/Mscfk/wUhBKeTmS4uaYOdTtun8undyzrPEXX9F2X5krxgfAQ).

Before changes (current state) - 12px font:
image

After changes - also 12px font:
image

[...]

PR #12897, opened 2025-09-15
Clarify Scene creation creditContainer documentation that canvas needs a parent

markschlosseratbentley:
# Description

This is a documentation change that makes the following `Scene.js` code clearer to a caller: https://github.com/CesiumGS/cesium/blob/29458dc47ac458ca78c3fc22f948fe280a13443f/packages/engine/Source/Scene/Scene.js#L160

I have updated the documentation for the optional `creditContainer` parameter which can be used when creating a [Scene](https://cesium.com/learn/ion-sdk/ref-doc/Scene.html) object. The update to the documentation makes it clear that, if not specified, a `creditContainer` will be implicitly created which expects the `canvas` to already have parent. I adopted the "sibling" phrasing from the subsequent `creditViewport` documentation to clarify this.

I also fixed a slight typo in the docs for `creditViewport`.

This change makes it less likely that a user will create a `Scene` object, pass it a parent-less canvas, and encounter a runtime error when CesiumJS tries to access that undefined parent.

## Issue number and link

Fixes https://github.com/iTwin/itwinjs-core/issues/8494

## Testing plan

This is merely a documentation change on the CesiumJS side. I did test and verify my assumptions on the iTwin.js (caller) side via debugging and stepping through the CesiumJS code. This verified that specifying a `creditContainer` when creating a `Scene` object resolves the runtime error we encountered when passing in a parent-less canvas.

# Author checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -102,6 +102,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
   - [Paul Connelly](https://github.com/pmconne)
   - [Jason Crow](https://github.com/jason-crow)
   - [Erin Ingram](https://github.com/eringram)
+  - [Mark Schlosser](https://github.com/markschlosseratbentley)
 - [Flightradar24 AB](https://www.flightradar24.com)
   - [Aleksei Kalmykov](https://github.com/kalmykov)
 - [BIT Systems](http://www.caci.com/bit-systems)


markschlosseratbentley:
@ggetz Going through the checklist on the PR description:
1. Do documentation changes warrant CHANGES.md entries?
2. What do we do for the unit test checkbox for strictly a documentation change?

ggetz:
Good questions @markschlosseratbentley!

> Do documentation changes warrant CHANGES.md entries?

Typically, no. We only add entries for fixes or API addition or changes. The one exception is if you need to update a type, as it will change the `.d.ts` typescript definitions file we generate (but that's not relevant here).

> What do we do for the unit test checkbox for strictly a documentation change?

Unit tests should be added typically to cover new features or to account for an edge case such as when a bug is fixed. If you didn't need to change the source code, likely you don't need to add or update a unit test.

ggetz:
Thanks for the PR @markschlosseratbentley! The updates look good to me.

The tests failures in CI are unrelated to your changes here (see https://github.com/CesiumGS/cesium/pull/12884#issuecomment-3286846225). We're looking into them separately, and don't need to hold up this PR.

ggetz (approved):

Issue #12894, opened 2025-09-12
Rebuilding Sandcastle: Post release tasklist


jjspace:
We recently soft released the new version of Sandcastle. Most of the work for that was tracked in this issue: https://github.com/CesiumGS/cesium/issues/12566

That issue has gotten quite large but has a lot of good historical discussion.
I'm opening this new issue to help refocus on the remaining tasks and consolidate extra tasks like those pointed out in the feedback issue: https://github.com/CesiumGS/cesium/issues/12857
I would expect this issue to break out into others as needed.

## Running Task list

Loosely sorted by a coarse 1-3 priority 

- [ ] 1 Content pass on all sandcastles
    - Some work on this was already done in <https://github.com/CesiumGS/cesium/pull/12865>
    - We may be able to cull a bunch of them. Others may need broken apart into multiple smaller ones
        - Specifically many of the ones under the [`development`](https://github.com/CesiumGS/cesium/tree/main/Apps/Sandcastle/gallery/development) directory seem to be old versions or duplicates of others we have and don't need anymore
    - Consider actual _categories_
    - Convert "Viewerless sandcastle" into a Sandcastle API demo
- [x] 1 Try the ion sdk setup and identify issues
- [x] 1 Document and solidify build process from the repo root (related: [comment](https://github.com/CesiumGS/cesium/pull/12631#issuecomment-2964026025)) Updates in <https://github.com/CesiumGS/cesium/pull/12904,> may need more in the future
- [x] 1 Add more documentation around building sandcastle itself
    - <https://github.com/CesiumGS/cesium/pull/12904>
    - [ ] Better document how to use `buildSandcastle` ([comment](https://github.com/CesiumGS/cesium/pull/12904#discussion_r2430120195))
    - [ ] document desired thumbnail size (also update styles with aspect ratio) ([comment](https://github.com/CesiumGS/cesium/pull/12904#discussion_r2430129760))
- [x] 1 build improvements - <https://github.com/CesiumGS/cesium/pull/12904>
   - Look at what files we include in the zip download version
   - [x] 2 Verify how we're "building" cesium into the build, maybe we can point at a different path instead of copying built files.
       - Settled on relative paths and only bundling the files for the production deployment versions. See [Sandcastle build config updates by jjspace · Pull Request #12904](https://github.com/CesiumGS/cesium/pull/12904)
   - [x] 2 finalize `eslint` config for new package files. First time TS and React is in there
[...]

PR #12904, opened 2025-09-19
Sandcastle build config updates

jjspace:
# Description

This PR contains updates to the build process for Sandcastle. These were largely inspired/required by the ability to build sandcastle for the ion-sdk repo/library.

- New `buildStatic` and `createSandcastleConfig` allow building sandcastle from outside the package directory. This should reduce the pseudo-dependency that the sandcastle package has on knowing it's nested in the CesiumJS repo.
    - This function should also now make it much more clear which files/resources are "external" by just setting all the necessary paths to access them wherever it's hosted.
    - [ ] I still need to actually relocate the configs into the larger CesiumJS build processes but I wanted to confirm the approach first. I think we'll be able to remove all the vite configs except the dev one after this.
- The `engine` and `widgets` packages are now bundled into a single file to make it easier to load them in the browser. 
    - Using this and an `importmap` we're also now using the version of `cesium` that imports/re-exports from `engine`/`widgets` to avoid duplicate cesium versions in the browser
    - This also enables easier importing of the ion-sdk packages when we need to
    - The types import was also made dynamic based on build-time paths. This can probably still be improved in the future but works and I've tested it with the ion-sdk
- Development sandcastles will no longer be included in the production build. This was a small config change while I was already in there

## Issue number and link

Part of #12894 

## Testing plan

- This was mostly an internal, build system, related change so there's not much to test in the app.
- Make sure `npm run build-sandcastle` works from the project root
- Make sure things are built to the correct, expected locations

# Author checklist


ggetz:
Thanks for the refactoring @jjspace!

I'm not sure if this is covered by the TODO item in your PR description, but [the deployed app](https://ci-builds.cesium.com/cesium/sandcastle-build-updates/Apps/Sandcastle2/index.html) is not loading:

image

In broad strokes, the approach you describe makes sense to me. I'm taking a closer look at the code itself now.

packages/sandcastle/scripts/typescriptCompile.js
```diff
@@ -0,0 +1,83 @@
...
+  return configParseResult;
+}
+
+/**
+ * Compile a typescript project from it's config file
+ *
+ * @param {string} configPath Absolute path to the config file to build
+ * @returns {number} exit code from the TS build process
+ */
+export default function typescriptCompile(configPath) {
```
ggetz: I'm a bit confused why this script is needed. Can't we run `tsc` directly?

packages/sandcastle/vite.config.js
```diff
@@ -1,10 +1,20 @@
...
+  const { version } = JSON.parse(data);
+  return version;
+}
+
+/** @import {UserConfig} from 'vite' */
 
 // https://vite.dev/config/
-const baseConfig: UserConfig = {
+/** @type {UserConfig} */
+const baseConfig = {
```
ggetz, line 11: Why now JS instead of TS?

packages/sandcastle/scripts/buildStatic.js
```diff
@@ -0,0 +1,141 @@
...
+    Sandcastle: "templates/Sandcastle.d.ts",
+  };
+  for (const [key, value] of Object.entries(imports)) {
+    importMap[key] = value.path;
+    typePaths[key] = value.typesPath;
+  }
+
+  config.define = {
+    ...config.define,
+    __VITE_TYPE_IMPORT_PATHS__: JSON.stringify(typePaths),
```
ggetz, line 105: It seems like we'er bending over backwards to get the type definition locations. I'm thinking this would be a lot easier if we build out the definitions relative to the bundled files—Then, we wouldn't need to configure the path to `Source/Cesium.d.ts` and instead could get the path at runtime based on the resolved module URLs from the import maps:

```js
const cesiumModulePath = import.meta.resolve("cesium");
const tsdPath = new URL("./Cesium.d.ts", cesiumModulePath).href;
```

packages/sandcastle/src/SandcastleEditor.tsx
```diff
@@ -192,25 +189,97 @@ function SandcastleEditor({
...
+          if (typesContent.trim().startsWith("import")) {
+            const declareModuleLine = typesContent.match(
+              /declare module "([\w@\-\/]+)" {/gm,
+            )?.[0];
+            if (declareModuleLine) {
+              return `${declareModuleLine}
+              ${typesContent.replace(declareModuleLine, "")}`;
+            }
+          }
+          return typesContent;
```
ggetz: I'm not clear on all the things that are happening here and in `transformTypes` in general, but is it possible to build all this into the tsd files directly?

packages/sandcastle/vite-plugins.js
```diff
@@ -0,0 +1,62 @@
...
+  };
+};
+
+/**
+ * Specify an import map for the built html files
+ * @param {Object<string, string>} imports map of imports
+ * @param {string[]} [filenames]
+ * @returns {PluginOption}
+ */
+export const insertImportMap = (
```
ggetz, line 31: Huh, is there no standard plugin for this? If not, are we perhaps missing something that serves a similar function in vite?

packages/sandcastle/vite.config.dev.ts
```diff
@@ -1,46 +1,58 @@
...
+      typesPath: "/packages/engine/index.d.ts",
+    },
+    "@cesium/widgets": {
+      path: "/packages/widgets/Build/Unminified/index.js",
+      typesPath: "/packages/widgets/index.d.ts",
+    },
+  },
+  // IN DEV THIS DOES NOT COPY FILES it simply sets up in-memory server routes
+  // to the correct locations on disk in the parent directories
+  // This config should not be used for the actual build, only development of Sandcastle itself
```
ggetz: I think my question is more along the lines of: If this is only being used for the development server, why are we using the `viteStaticCopy` plugin at all? Is there a more standard way to configure these paths? Should they be aliased instead?

packages/sandcastle/README.md
```diff
@@ -7,7 +7,7 @@ This package is the application for Sandcastle.
 - `npm run dev`: run the development server
 - `npm run build`: alias for `npm run build-app`
 - `npm run build-app`: build to static files in `/Apps/Sandcastle2` for hosting/access from the root cesium dev server
-- `npm run build-ci`: build to static files in `/Apps/Sandcastle2` and configure paths as needed for CI deployment
+  - If the env variable `BASE_URL` is set it will be prefixed on all required paths in the application. This is useful for building to a "nested" url like we do in CI
```
ggetz: This is the app's base public path, correct? I'm not sure that's entirely clear in the wording, especially since `CESIUM_BASE_URL` is in play.

jjspace:
> I'm not sure if this is covered by the TODO item in your PR description, but [the deployed app](https://ci-builds.cesium.com/cesium/sandcastle-build-updates/Apps/Sandcastle2/index.html) is not loading:

Mismatch in my local CI build test not including the trailing `/` in the path that the CI job does. [Fixed now!](https://ci-builds.cesium.com/cesium/sandcastle-build-updates/Apps/Sandcastle2/index.html)
I'm looking through your other comments now.

packages/sandcastle/vite-plugins.js
jjspace: There seems to be a couple plugins that do this but none are popular or seem to be actively maintained. This code is loosely based on [this one](https://github.com/pakholeung37/vite-plugin-import-maps) but mostly just using the [Plugin API](https://vite.dev/guide/api-plugin). It's a small requirement and could be made to fit our use cases so I felt it was better to just fully "own" the code and implement the plugin ourselves.

packages/sandcastle/vite.config.dev.ts
jjspace: > If this is only being used for the development server, why are we using the `viteStaticCopy` plugin at all?

Largely because we already have it as a dependency. It didn't feel necessary to add another plugin or system when this one works well for our usecase. It's also how we tell people to set up a vite application in our own [example repo](https://github.com/CesiumGS/cesium-vite-example)

> Is there a more standard way to configure these paths? Should they be aliased instead?

Not that I've been able to find. This is once again a slightly "special" case for us because we want to statically host files on the dev server that are _not in the public directory_. Most vite articles say to "just add your files to `public` and it'll work" but that kinda defeats our purpose of using the local files directly. The alternative seems to be to set up a [custom plugin](https://vite.dev/guide/api-plugin.html#configureserver) with middleware/routing on the `serve` part of `vite` to "host" the expected files at the expected paths. We _could_ implement this ourselves but the [vite static copy plugin already does](https://github.com/sapphi-red/vite-plugin-static-copy/blob/main/src/serve.ts) all this heavy lifting for us. [`server.proxy`](https://vite.dev/config/server-options.html#server-proxy) does not work, it proxies to a different hostname which is not what we want. [`resolve.alias`](https://vite.dev/config/shared-options.html#resolve-alias) does not work either, it seems to only care about the built files which these are not.

packages/sandcastle/vite.config.js
jjspace: To enable easy imports outside of the sandcastle package without requiring it be compiled with typescript first.
`/scripts/build.js` -> `/packages/sandcastle/buildStatic.js` -> `/packages/sandcastle/vite.config.js`

packages/sandcastle/scripts/typescriptCompile.js
jjspace: We _could_ run it as a child process but that felt more messy and potentially unreliable when called from outside the package. I had a version that did that based off [this answer](https://stackoverflow.com/a/53624474/7416863). This way felt like the more reliable/stable method to actually use the package's dependencies.

javagl:
Also see https://github.com/CesiumGS/cesium/pull/12911

packages/sandcastle/src/SandcastleEditor.tsx
jjspace: > is it possible to build all this into the tsd files directly?

I was going to say "no" because I assumed this was just the result of the `tsd-jsdoc` output. But apparently this specific part was our own modifications _after_ the jsdoc stuff was generated.

I was also worried about modifying the types files we generate and export with the library but long story short I think this change is fine and things _seem_ to still "just work" as they currently do. I've pushed a new commit adjusting the build and removing most of these transforms

<details><summary>Short story long:</summary>
<p>

Our types files for the packages themselves were roughly generated in the format

```typescript
import { ... } from '@cesium/engine';
declare module '@cesium/widgets' {
// ...
}
```

When this file was added to the monaco `extraLibs` to enable the intellisense it wasn't making the connection to the `@cesium/engine` module. So `viewer.cesiumWidget` was not a known type. Switching the order to have the import statements _inside_ the module declaration file made it work for monaco; that's what most of these transforms were in the code.

Reading into [Typescript Modules](https://www.typescriptlang.org/docs/handbook/modules/reference.html) a bit and there are multiple ways to declare modules that all behave a little different. Normally (I believe) an entire file is treated as a single module itself. When using the `declare module` syntax this becomes an [Ambient Module](https://www.typescriptlang.org/docs/handbook/modules/reference.html#ambient-modules). However if there are any `import` or `export` statements _also_ at the "top level" of the file then the `declare module` syntax becomes a [_module augmentation_](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) even though the syntax is identical. :woman_facepalming: 

In other words, our current types exported from the packages were _module augmentations_ (though technically only the `widgets` and ion-sdk packages, engine had no imports and was only ambient). My latest change to move the imports inside the `declare` makes both packages ambient.
In some quick testing this doesn't _seem_ to make any difference to the TS server when importing `widgets` in a new npm project in VSCode. I'm definitely not an expert on this part of TS though so I could be wrong.

Also worth reading a bit into the "correct" structures for [declaration files for libraries](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html) and this page on the actual [module `.d.ts` file](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html)
</p>
</details>

javagl:
I just reopened https://github.com/CesiumGS/cesium/issues/12910 : There is another `import` at https://github.com/CesiumGS/cesium/blob/20f3edc9aabc077525ab01f72498baf908bede21/packages/sandcastle/scripts/buildGallery.js#L306 that will have to become
`const config = await import(pathToFileURL(configPath).href);`

(I could create a PR, but... let's not over-formalize the process here...)

packages/sandcastle/scripts/typescriptCompile.js
ggetz: I'm a bit concerned simply because it's so atypical and therefore poses more of a maintenance burden. [The `tsc` CLI is documented](https://www.typescriptlang.org/docs/handbook/compiler-options.html), but there doesn't seem to be much [for the Node API](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API). Plus it's either an unstable API or the docs haven't been updated to reflect its stability.

I don't mind wrapping a helper function for typescript in it's own script. But if we do so:
1. I'd lean towards running a child processes.
2. Can we move the script to the top-level `cesium` package? Nothing needs to be sandcastle specific, right?

packages/sandcastle/vite-plugins.js
ggetz: At first brush, that all sounds fine. The potential gotcha I'm worried about is that 

>  but none are popular or seem to be actively maintained

is because we're missing a simpler or more integrated way of getting to the same result. Nothing is top of mind, but I may investigate further.

My ask is just that we've done the due diligence on alternative approaches.

packages/sandcastle/scripts/buildStatic.js
```diff
@@ -0,0 +1,141 @@
...
+ * Create the Vite configuration for building Sandcastle.
+ * Set where it should build to and the base path for vite and CesiumJS files.
+ *
+ * Most importantly specify the paths the app can find the library imports.
+ *
+ * If you are copying files to the built directory ensure the source files exist BEFORE attempting to build Sandcastle
+ *
+ * @param {object} options
+ * @param {string} options.outDir Path to build files into
+ * @param {string} options.viteBase Base path for files/routes
```
ggetz: Fundamentally, the application base path is not vite-specific. 

```suggestion
 * @param {string} [options.base="/"] Base URL from which all non-absolute URIs are specified when the application is served in development or production.
```

packages/sandcastle/scripts/buildStatic.js
```diff
@@ -0,0 +1,141 @@
...
+  checkForImport(imports, "@cesium/widgets");
+  if (imports["Sandcastle"]) {
+    throw new Error(
+      "Don't specify the Sandcastle import this is taken care of internally",
+    );
+  }
+
+  /** @type {Object<string, string>} */
+  const importMap = {
+    Sandcastle: `${viteBase}/templates/Sandcastle.js`,
```
ggetz: Rather than directly concatenating, I'd suggest using `path.join` or constructing a `new URL` so that the results are properly configured when the base path contains a trailing `/`.

packages/sandcastle/standalone.html
ggetz, line 1: Why does `bucket.html` live in `templates`, but not `standalone.html`?

packages/sandcastle/vite.config.js
ggetz: consider simplifying these a bit with [`path.resolve`](https://nodejs.org/api/path.html#pathresolvepaths), which should ensure the absolute path:

```suggestion
index: resolve(__dirname, "index.html"),
bucket: resolve(__dirname, "templates/bucket.html"),
standalone: resolve(__dirname, "standalone.html"),
```

packages/sandcastle/vite.config.js
```diff
@@ -1,10 +1,20 @@
...
-import rootPackageJson from "../../package.json";
+import { dirname, join } from "path";
+import { readFileSync } from "fs";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+function getCesiumVersion() {
+  const data = readFileSync(join(__dirname, "../../package.json"), "utf-8");
+  const { version } = JSON.parse(data);
+  return version;
+}
```
ggetz: Should the value of Cesium version be passed as an option rather than autodetected here? `getCesiumVersion` is now defined in multiple places, despite serving the same purpose.

packages/sandcastle/scripts/buildStatic.js
```diff
@@ -0,0 +1,141 @@
...
+  };
+  for (const [key, value] of Object.entries(imports)) {
+    importMap[key] = value.path;
+    typePaths[key] = value.typesPath;
+  }
+
+  config.define = {
+    ...config.define,
+    __VITE_TYPE_IMPORT_PATHS__: JSON.stringify(typePaths),
+    __COMMIT_SHA__: JSON.stringify(commitSha ?? undefined),
```
ggetz, line 107: Is there a reason not to use `process.env.GITHUB_SHA` directly?

packages/sandcastle/vite.config.js
ggetz: Another option would be to set a env variable with the value of cesium version. We already do this for building the docs. 

I'm leaning towards this route as I think we could streamline several areas of our build processes and CI workflows if we use an environment variable instead. We could extract and set the env var as part of the `postinstall` step to ensure it's only run once and before all of the other `build` steps.

packages/sandcastle/scripts/buildStatic.js
```diff
@@ -0,0 +1,141 @@
...
+  config.base = viteBase;
+
+  config.build = {
+    ...config.build,
+    outDir: outDir,
+  };
+
+  const copyPlugin = viteStaticCopy({
+    targets: [
+      { src: "templates/Sandcastle.(d.ts|js)", dest: "templates" },
```
ggetz, line 76: Can the TS build output to the `public` directory instead to save a step?

packages/sandcastle/vite-plugins.js
```diff
@@ -0,0 +1,62 @@
...
+      config.define = {
+        ...config.define,
+        __CESIUM_BASE_URL__: JSON.stringify(cesiumBaseUrl),
+      };
+    },
+    transformIndexHtml(html) {
+      return html.replaceAll("__CESIUM_BASE_URL__", `${cesiumBaseUrl}`);
+    },
+  };
+};
```
ggetz, line 23: What all do we need `__CESIUM_BASE_URL__` after using import maps? Consider  also that we can dynamically get the paths relative to the Cesium ESM import at runtime based on the import map.

packages/sandcastle/scripts/buildStatic.js
```diff
@@ -0,0 +1,141 @@
...
+/** @import { UserConfig } from 'vite' */
+
+/**
+ * @typedef {Object} ImportObject
+ * @property {string} path The path to use for the import map. ie the path the app can expect to find this at
+ * @property {string} typesPath The path to use for intellisense types in monaco
+ */
+
+/**
+ * @typedef {Object<string, ImportObject>} ImportList
```
ggetz, line 18:
```suggestion
 * @typedef {Record<string, ImportObject>} ImportList
```

packages/sandcastle/vite.config.prod.ts
```diff
@@ -1,51 +1,50 @@
...
+      path: "/js/widgets/index.js",
+      typesPath: "/js/widgets/index.d.ts",
+    },
+  },
+  copyExtraFiles: [
+    { src: join(__dirname, `${cesiumSource}/ThirdParty`), dest: cesiumBaseUrl },
+    { src: join(__dirname, `${cesiumSource}/Workers`), dest: cesiumBaseUrl },
+    { src: join(__dirname, `${cesiumSource}/Assets`), dest: cesiumBaseUrl },
+    { src: join(__dirname, `${cesiumSource}/Widgets`), dest: cesiumBaseUrl },
+    { src: join(__dirname, `${cesiumSource}/*.(js|cjs)`), dest: cesiumBaseUrl },
```
ggetz: Should we copy the Cesium build artifacts to S3 as part of the GitHub Actions workflow? That would be more consistent with our other deployments.

jjspace:
> Let me know the status of this PR, as I think you mentioned there are still updates incoming?

@ggetz yes, sorry for the delay it took a bit longer getting things together than I thought it would. I've just pushed an update to extract the build configs out to the top level and eliminate all of the "pseudo-dependency" on being in the cesium repo. This includes now exporting the buildStatic and buildGallery functions for use in the cesium build scripts/gulpfile.
I'm looking through your new comments now, I think some might already be addressed by this change

packages/sandcastle/scripts/buildStatic.js
jjspace: No, but I had left it this way as a bit of an extra hint for people that _do_ understand vite. Fine with changing it though.

packages/sandcastle/standalone.html
jjspace: Largely historical reasons. I believe there were some issues with routes before. Possibly resolved now but I would like to leave it alone in this pr as it's working as is.

packages/sandcastle/vite.config.js
jjspace: I believe I took this from some reference docs or an example somewhere but don't mind changing it

packages/sandcastle/vite.config.js
jjspace: > Should the value of Cesium version be passed as an option rather than autodetected here? 

Yes, I thought the same thing, already updated.

> Another option would be to set a env variable with the value of cesium version

I think my new approach/opinion is to leave any env variable reliance out of the sandcastle scripts. This could be used in the main build/gulpfile scripts but I think that can be separate from these changes if we want to go that route

packages/sandcastle/scripts/buildStatic.js
jjspace: We may not always want to use it. For example in Prod we don't want to show it but it's currently set in the github workflow. Doing it this way makes it customizable and explicit from wherever we call this function. Using the env variable directly creates more, almost, "side effect" logic that people have to be aware of

packages/sandcastle/vite-plugins.js
jjspace: We need the cesium base url for all of the assets and files that CesiumJS loads. This is setting up the app the same way we recommend/require everyone else to do. I think we should leave it like this

packages/sandcastle/scripts/buildStatic.js
jjspace: This is JS doc not TS, `Record` doesn't exist in JS. Intellisense would probably pick it up fine in most environments and I'm already stretching the rule a little with the `@import` lines but it was intentional to use `Object` not `Record` here.

packages/sandcastle/vite.config.prod.ts
jjspace: It might be possible to do that but then it would leave the built files not working by themselves. This way it creates a single directory that can be copied to S3, just like the other build processes do. Plus this way we can test it locally by doing `PROD=true npm run build-sandcastle` and then hosting those static files with something like `npx http-server`

(FYI This file and configuration has been removed in favor of incorporating it into the gulpfile like our other build steps)

packages/sandcastle/vite-plugins.js
jjspace: > My ask is just that we've done the due diligence on alternative approaches.

AFAIK I have. This seems like the cleanest and easiest solution for what we need. If we didn't want to request external files then the import map wouldn't be needed and everything could be bundled by vite. I assume that's why this is not a more "common" process.

packages/sandcastle/scripts/typescriptCompile.js
jjspace: Switching to the tsc cli is a bit tricky too because we need to figure out where the binary actually is in the node_modules folder. I think I figured out an ok way to do this and just updated with a child process.

> Can we move the script to the top-level cesium package? Nothing needs to be sandcastle specific, right?

The function itself is not necessarily Sandcastle specific but using it _is_. It's required to be called before the vite build kicks off so it's nicer in the Sandcastle package itself. If we moved it up then that'd be another dependency sandcastle has on cesium.

packages/sandcastle/scripts/buildStatic.js
jjspace: I like the idea of this change but it would involve some larger changes with the rest of our build systems. Happy to take a look at it at some point but I think I'd prefer to leave this as is for this current PR. Maybe it's rather verbose in the config but at least it should be clear what it's doing.

jjspace:
I need to finalize how this works in the release zip file. Currently it can't "install" the sandcastle package because it's not included. We need to build it before the zip and/or sort out which files to include.

.github/workflows/prod.yml
```diff
@@ -41,14 +41,17 @@ jobs:
...
-      - name: build prod sandcastle
-        run: npm run build-prod -w packages/sandcastle -- -l warn
+      - name: build apps
+        run: npm run build-apps
       - name: deploy to cesium.com
         if: ${{ env.AWS_ACCESS_KEY_ID != '' }}
+        # Download zip from the Github release and unzip to Build/release/
+        # Publish that unzipped code to the bucket for https://cesium.com/downloads/cesiumjs/releases/[version]/... urls
+        # Publish the documentation files to the bucket for https://cesium.com/learn/cesiumjs/ref-doc/... urls
+        # Publish the simple viewer app
```
jjspace: @ggetz do you know where the `CesiumViewer` app is deployed/hosted?

packages/sandcastle/standalone.html
jjspace: I thought about this more as I was working and I'm _assuming_ that this might have been an effort to keep the urls "cleaner"? Instead of `sandcastle.cesium.com/templates/standalone.html` it only has to be `sandcastle.cesium.com/standalone.html`. Regardless the shorter url should still work in case anyone has saved it.

I did some work to simplify paths across the board by now utilizing relative paths as much as possible. I coupled this with setting the `<base>` element for the standalone so it _behaves_ like it's the `templates/bucket.html` file. This seems to simplify the build and still leave this path working as expected so I think it's a good compromise

jjspace:
@ggetz I've pushed a handful of updates with a focus on stability and simplification to support our various build methods.

One of the bigger/more important changes is that Sandcastle is now relying on _relative_ urls for it's assets and resources as much as possible for non-PROD builds. This means the same build should work locally and in CI and deployed to the releases CDN page. PROD builds still set absolute paths because they're intended to fully "bundle" all necessary assets.

Alongside that change I also had to modify the `standalone.html` file a little further to account for the relative routes which will be different compared to `bucket.html`. 

Testing different builds:
Run `git clean -dxf` between each to start as if you just cloned the repo

- Zip file
     - `npm install && npm run make-zip`
     - Extract to some temp folder
     - run `npm install && npm start` in said folder
- "cold start" or Fresh install flow
    - `npm install`
    - `npm start` - this should now also build sandcastle if it wasn't already
- PROD deployment
    - `npm install`
    - `PROD=true npm run website-release`
    - `PROD=true npm run build-ts`
    - `PROD=true npm run build-apps`
    - Inside `Build/Sandcastle2` run `npx http-server` to check that it all works as an isolated set of static files
 
I also want to share a small server script that's useful for checking the ci route and release route both work locally
<details><summary><code>ci-server.js</code></summary>
<p>

Just put this file at the root of the repo named `ci-server.js` or something and run it in place of `npm start`
Then navigate to 
- http://localhost:8080/cesium/branch/Apps/Sandcastle2/index.html
- http://localhost:8080/downloads/cesiumjs/releases/1.133/Apps/Sandcastle2/index.html

```js
import path from "path";
import yargs from "yargs";
import express from "express";

function excludeRoutes(excludedRoutes) {
  return function middleware(req, res, next) {
    if (excludedRoutes.some((route) => req.path.startsWith(route))) {
      return res.status(404).send();
    }
    next();
  };
}

const argv = yargs(process.argv)
  .options({
    port: {
      default: 8080,
      description: "Port to listen on.",
    },
  })
  .help().argv;

(function () {
  const app = express();

  app.use(
    "/downloads/cesiumjs/releases/1.133",
    express.static(path.resolve(".")),
  );
  app.use(
    "/cesium/branch",
    // the CI deploy excludes these folders, don't serve them here
    excludeRoutes([
      "/.git/",
      "/.github/",
      "/.husky/",
      "/.vscode/",
      "/Build/Coverage/",
      "/Build/CesiumDev/",
      "/Build/Specs/e2e",
      "/Documentation/",
      "/node_modules/",
      "/scripts/",
      "/Tools/",
    ]),
    express.static(path.resolve(".")),
  );

  const server = app.listen(argv.port, "localhost", () => {
    console.log(
      "Cesium mock ci server running locally.  Connect to http://localhost:%d/cesium/branch/",
      server.address().port,
    );
  });

  server.on("close", function () {
    console.log("Cesium deployment test server stopped.");
    process.exit(0);
  });

  let isFirstSig = true;
  process.on("SIGINT", function () {
    if (isFirstSig) {
      console.log("\nCesium deployment test server shutting down.");

      server.close();

      isFirstSig = false;
    } else {
      throw new Error("Cesium deployment test server force kill.");
    }
  });
})();
```

</p>
</details>

.github/workflows/prod.yml
ggetz: @jjspace I believe this is https://cesium.com/cesiumjs/cesium-viewer/

.github/workflows/sandcastle-dev.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: build types
         run: npm run build-ts
       - name: build prod sandcastle
-        run: npm run build-prod -w packages/sandcastle -- -l warn
+        run: npm run build-sandcastle
```
ggetz, line 31: Should we still keep this workflow file around? At the very least, I would think this shouldn't be run on the "cesium.com" branch.

If we do keep it around, maybe it should only be run on "main"?

gulpfile.js
```diff
@@ -287,9 +289,20 @@ export async function buildTs() {
   await createTypeScriptDefinitions();
 }
 
-export function buildApps() {
-  return Promise.all([buildCesiumViewer(), buildSandcastle()]);
-}
+export const buildNewSandcastle = gulp.series(
```
ggetz: I assume we'll want to remove the older Sandcastle build code shortly. Can we change this to `buildSandcastle` and rename the old one `buildLegacySandcastle`?

gulpfile.js
```diff
@@ -582,121 +595,129 @@ async function pruneScriptsForZip(packageJsonPath) {
...
-  const widgetsPackageJsonSrc = await pruneScriptsForZip(
-    "packages/widgets/package.json",
-  );
+export const makeZip = gulp.series(
+  release,
+  async function buildSandcastleApp() {
+    return buildNewSandcastleApp(false);
+  },
+  async function buildGallery() {
+    return buildSandcastleGallery(false);
```
ggetz: So building Sandcastle and building the gallery are two separate steps here, correct? I'm not sure I'm clear on why—What's the goal behind that?

I know we'd like to build the gallery only for local development, but could it be a part of the overall sandcastle build as well?

scripts/build.js
```diff
@@ -627,6 +675,99 @@ export async function getSandcastleConfig() {
   };
 }
 
+async function importSandcastleBuildFunctions() {
+  // Import asynchronously, for now, because this script is not included or run in the release zip;
```
ggetz: I put this in as a short-term fix since I only noticed it was a problem while prepping a release.

Ideally, I think we should be avoiding the async import. But I do understand that this is tricky given this is a ultimately a dependency of `make-zip`.

What do you think of moving functions for npm "scripts" that are excluded from the release zip to a separate file to avoid the need for async imports?

scripts/build.js
```diff
@@ -307,6 +307,54 @@ export async function createCesiumJs() {
...
+ * @param {object} options
+ * @param {string} options.outputDirectory Directory where build artifacts are output
+ * @param {string} options.entryPoint script to bundle
+ * @param {boolean} [options.minify=false] true if the output should be minified
+ * @param {boolean} [options.removePragmas=false] true if the output should have debug pragmas stripped out
+ * @param {boolean} [options.sourcemap=false] true if an external sourcemap should be generated
+ * @param {boolean} [options.incremental=false] true if build output should be cached for repeated builds
+ * @param {boolean} [options.write=true] true if build output should be written to disk. If false, the files that would have been written as in-memory buffers
+ */
+export async function bundleIndexJs(options) {
```
ggetz, line 321: I'd love to clean this up a bit to reduce some complexity and wrote duplication with some other parts of our build code...But that is a nice to have and shouldn't be strictly needed for this PR.

gulpfile.js
```diff
@@ -287,9 +289,20 @@ export async function buildTs() {
...
 
-export function buildApps() {
-  return Promise.all([buildCesiumViewer(), buildSandcastle()]);
-}
+export const buildNewSandcastle = gulp.series(
+  async function buildSandcastleApp() {
+    return buildNewSandcastleApp(isProduction);
+  },
+  async function buildGallery() {
+    return buildSandcastleGallery(!isProduction);
```
ggetz: Why do these take opposite args? It's a bit confusing, no? I _think_ my confusion is because  "Production" within the context of the gulpfile means something different than "Production" for the Sandcastle app.

packages/sandcastle/src/SandcastleEditor.tsx
```diff
@@ -210,25 +207,63 @@ function SandcastleEditor({
...
+          const typesContent = transformTypes
+            ? transformTypes(responseText)
+            : responseText;
+          monaco.languages.typescript.javascriptDefaults.addExtraLib(
+            typesContent,
+            filename,
+          );
+        } catch (error) {
+          console.error(`Unable to load types for ${filename} at ${url}`);
+          console.error(error);
```
ggetz, line 264: Should this throw instead of just logging the error? It could be easy to miss that this breaks.

```suggestion
          throw new Error(`Unable to load types for "${filename}" at "${url}"`, { cause: error });
```

packages/sandcastle/README.md
```diff
@@ -5,12 +5,23 @@ This package is the application for Sandcastle.
...
+There are 2 main conceptual ways that Sandcastle gets built which mostly revolve around how to access CesiumJS resources:
+
+1. Sandcastle points to "external" paths for CesiumJS resources
+2. Sandcastle is built to 1 static location that is co-located with all CesiumJS files. ie they're all copied into the built location
+
+The first method is useful and desired when developing the project locally and you want to refer to the actively built and updated CesiumJS files as you do other work. This is how the Sandcastle development server (`npm run dev`) and the local static version at `/Apps/Sandcastle2` are built.
+
+The second method is used when building Sandcastle to be deployed to the website or other static location. You can think of this as "bundling" all the necessary files needed for Sandcastle into 1 single directory.
+
+Regardless the method you want to use Sandcastle is always built using the exported `buildStatic`, `createSandcastleConfig` and `buildGalleryList` functions.
```
ggetz, line 23: Based on this summary, I'm not sure I would understand _how_ to use one method over the other. Maybe provide some examples to make this a bit more concrete?

packages/sandcastle/README.md
```diff
@@ -48,6 +59,10 @@ thumbnail: thumbnail.jpg
 development: false
 ```
 
+### Thumbnails
+
+Thumbnails should be any image that represents what the sandcastle does. Often this will just be the Viewer with or without any Sandcastle interaction buttons. Thumbnail files should be limited in size to help save on bandwidth. Currently most are around 225px in width.
```
ggetz, line 64: 1. I would also suggest adding compression to further reduce the memory footprint. I do this by saving as a jpg and reducing the quality until the file size is consistent with the others, but would be open to other approaches.
2. Could you please provide exact recommended dimensions? It's ok if the existing thumbnails do not comply, but let's try to be consistent going forward.
3. Could you document the steps to do this? It would be ideal to have this at the level of granularity similar to the release guide.

packages/sandcastle/src/SandcastleEditor.tsx
jjspace: I specially chose _not_ to throw an error here because it does not prevent the rest of the editor/application from working. Maybe the types won't load and intellisense won't show up but the app should operate fine without them.
This may happen commonly in dev specifically if someone hasn't recently run `npm run build-ts` or `make-zip` to generate the types files.
This is similar to how the old sandcastle did not strictly require the documentation to be built and load and just logged an error when it couldn't.

ggetz: Gotcha, make sense. Will this show only in the developer tools, or will it be reflected in the Sandcastle console? If the former, I might suggest an update to make sure it is obvious to the user.

scripts/build.js
```diff
@@ -1079,6 +1213,19 @@ export const buildEngine = async (options) => {
   // Create index.js
   await createIndexJs("engine");
 
+  await bundleIndexJs({
```
jjspace: Reminder for myself. These need hooked up in the dev `server.js` to rebuild when sandcastle refreshes if there have been changes. Right now it only does the top level files.

.github/workflows/sandcastle-dev.yml
jjspace: @ggetz I'm open to anything. It goes to the larger question of what the `dev-sandcastle.cesium.com` page should be. I think changing it to `main` could be good to act as a more stable "pre-release" version?

ggetz: > I think changing it to main could be good to act as a more stable "pre-release" version

Sounds good to me! Please update here as I think it's a one-liner. We can reconsider if we need something different longer term, but a stable "pre-release" sandcastle deployment should still be very helpful.

packages/sandcastle/src/SandcastleEditor.tsx
jjspace: Right now it will only show in the developer tools. I tried hooking it up to display in the console mirror. However, because it's cleared every time the sandcastle is run the message is quickly hidden and not actually visible. I think it could make sense to have a new category of message, something like a "system message" that persists beyond viewer refresh but that's definitely outside the scope of this PR.
I'll note it down for future improvement

gulpfile.js
jjspace: This is down to them meaning different things in each function and falling directly into the [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).
I've updated them to take options arguments to try and help make it clearer.

gulpfile.js
jjspace: "Because that's how it was" (is a bad answer).

Real answer is because sometimes we only want to do one or the other. Like in the dev server there's no reason to run the slower `buildSandcastle` for changes only in the gallery.
That said, I noticed every time I was calling `buildSandcastle` that `buildGallery` was called directly after. The inverse was not true. I moved the `buildGallery` call into the `buildSandcastle` function to make it one sequence while still allowing `buildGallery` to be called on it's own.

scripts/build.js
jjspace: This got a little tricky because these functions are also used in the dev `server.js` when not in `production`.

I do agree with wanting to avoid dynamic imports overall though.

I've pushed some changes to extract the `make-zip` and `build-apps` / `build-sandcastle` gulp tasks into separate gulp files that are _not_ included in the release zip. This lets us use the "true" imports in `scripts/buildSandcastle.js` which these new gulpfiles import. The `server.js` file was also updated to dynamically import these functions and account for if they're missing.

Everything should be verifiable by just running `npm run make-zip`, extracting and trying `npm start` or checking the extracted files.

scripts/build.js
jjspace: This function was "backported" from cesium-analytics from our rework into packages. Definitely agree it could probably be cleaned up further but I think we can do it outside this PR.

scripts/build.js
jjspace: This has been added

jjspace:
@ggetz This is ready for another look. I'm still planning to make some updates to the docs based on your last comments but I believe the _code_ should be good to go.

ggetz:
Thanks @jjspace! I've reviewed the code and I think we're in a good place. This just needs to be verified and tested and then we should be good to merge.

ggetz:
Everything seems to be working, from a functionality standpoint! 

The only thing I see is that the published build artifacts have drastically increased in size and are probably including too many things. 

- The release zip is including both old and new Sandcastle. We definitely shouldn't need to ship both, as it drastically increases the zip download size.
- `@cesium/engine` and `@cesium/widgets`, when packed with `npm pack`, include the `Unminfied` bundles. Should they?  The `engine` install for instance has over doubled in size, going from `22.7 MB` to `51 MB` unpacked.

These can be addressed in this PR, or a follow-up, but definitely need to be addressed before the next release. Just let us know your plan.

jjspace:
> @cesium/engine and @cesium/widgets, when packed with npm pack, include the Unminfied bundles. Should they? The engine install for instance has over doubled in size, going from 22.7 MB to 51 MB unpacked.

@ggetz I think it's fine to keep the bundles out of the packages for now. I've removed it from the engine files so the pack is smaller again. Widgets seemed unaffected to me. We can always add them back later if we want

> The release zip is including both old and new Sandcastle.

I was going to leave them both there for this release. The zip file itself only seems to change by 5mb to me when removing the old sandcastle? The zipped files are also the ones deployed to  the CDN-esque routes like `https://cesium.com/downloads/cesiumjs/releases/1.134/Apps/Sandcastle/index.html`. Currently that deployment doesn't have the new sandcastle at all. This release can deploy both to prove it's hooked up correctly then legacy removed from the zip in the next one?

I think this PR can be merged then and a quick follow up if we need to adjust the zip file

jjspace:
I'm going to merge this so it's in for the release today based on my comment above. We will probably revisit the zip file before the next release and can address any size concerns if they come up. The full removal of the legacy sandcastle probably needs some more discussion anyway but should be soon.

PR #12905, opened 2025-09-21
Reduces redundant clamping on labels

mzschwartz5:
# Description

Clamping things to ground is a slow operation (currently - #9961 will speed it up, but it's still relatively expensive). When a label is clamped to ground, it sets each of its glyphs (which are individual billboards) to clamp as well. 

This is slow and unnecessary - we already have [logic that copies the clamped position of the label to each of its glyphs.](https://github.com/CesiumGS/cesium/blob/83eb415e2ae339f1e09bc7de5f32f8dfbde4773a/packages/engine/Source/Scene/Label.js#L1143-L1150). It even claims to do this to reduce clamping redundancy, but unfortunately it's not sufficient. 

To address the issue, this PR adds a new boolean flag to `Billboard`: `positionFromParent` - when true, this flag indicates that the billboard gets its position from some ancestor, and thus expensive operations like clamping can be skipped. (I'm generally not a fan of adding special flags, but I considered a few other options and nothing was particularly simpler or feasible without a much larger refactoring effort. Happy to discuss alternative ideas though).

## Issue number and link

This is one piece of an overall labels performance improvement effort, stemming from #12719

## Testing plan

Test with [this sandcastle](https://dev-sandcastle.cesium.com/#c=rVdtb9s2EP4rhD/JqEeLepebBAscpzWaOkHsdRjiIKAl2uYqSwJFJfOG/PcdqRfbSZumywIEkh7e3XN35PHOUZYWEt1z9sAEOkYpe0BDVvByg79ozJh3Iv09zFJJecrEvNND/8xThCQTApBBozCrvvFSZJvfM5HENWB0e/P0sft+nkaarIhYyoCrIsX6Exab5ZhKCqv0gXKJjOqxZDJagytrKfNi0O+nbEOp2NIUr7hclwvMsz7NeX9DRcF4krC7POMF/rPI0nmn29UvRleTJEyiNeOrtbxmSyZYGilf6hA+Hi7g4cXp56u72eXdh+vL3yZn7yt1lkouOStAj5g1Bv+QDwW5mmZZppHkWYqSjMajWsHoVplTChxElTLSAeNlJkYUYjS4ZJsuOj6pJBHiS2SA8MnxjrbbrKk/wWQp0vcV8Fg9qkTmWcG1C214QyokvNHU1pt0xlaCgVcYY8WKG4UqUcpQvUUNM6ZxbLTcPB4grcfjXoM1JlogoQuWDPYdhnPzlxygeYfOO1iwnFFp1Nnr9vbEnmzS4CmwLxvzgi4SdsZyuZ6xQp7xQlKtNCk3C4jg6nI6no2/jO7Gk/PxZDz7o9WuU6bPp47q3Tv9poHH3bFUqVDbe6NDiYGEp1SFOkA3jmeHvh3i0LK90Ak8p4ccO3Q9GztOELhu4LuhghzHtn0cOJZpBZZl32onYi5YVBv6xcQBrAQmsUKtGVo9BKAZhjaxbMv0wb7p2j1kYsdTUsQlDtBaTmWszMEKrAGl6fjwIK7vWsRTCqYdmKHvWJ7pu8oHbSUIfN8nnm/ZgRdYHlFmHusCfxqk47mhiS0gtj3fBs8c8MV1MSF+6Fq2awYqSA/kCLZdxySu5XnfDNK3HD+0iOV6XhCQIKyC9ALfMokHzvjgom9r1IMs2CRwfBOWwMm9OPUqCYjj27Zr26HjeZUhn8COWKHp2MQmJFRx+pBUyw28MDRD03SrOG8PS3WZbGeZoTa6rrD9GwpHdMMEFKoWqs/zQYb27s5doWlzeE/uxrztPcPINzDrtimHTHCoQI2iXSm1KUUvUzdiO+IWIc+QHSmk+GXDZd5ahFeyewUbdU3VpuJS1Ckya4TRgqer8zrxbQcZHcD4YjwZnV7fTS4no95hSU5pGke0kAlTV9Isy5IFFZ9ZWho3dWuqrphhQjc5khn6ILIyjaF1qVWoZ5ZAwANEi20aofYAGO3N+oYOoW+9g0u/uk90YR14N8lS9syltzijMvUDD26ry/21CSQmaqy8PnlP++PrM0LMN/P9NOELjD/kMv/fbLtouKaCRqoXvj78w9Hj9bFb5hvprJ/f3BcIX6Qi/yHVghUwZx0/N/x0qhFsk92z0yTR9h6rptD2ggKuOYUfNIOYLcrVdJ09nAtoC8UVE1MGc0IMdFKU1Th7TwVagzvJ08l6GsHYlU5zGrHRPXjxsRIymj4DmoVirJUxhDFO81Ke6jCMNp4svWBLOUx49NVQEWzAVps7xS7oFpgP2teKySuQv6bbVuVg7mtUAWtVV0m2YDhXNGCxV6GNsJqOMsh2kq2MA6anVr8r2HafH0qWebPh30vlbJszfDE6n90NL8bDT3Nts9PrHBVym7ATpfwr3+SZkKgUCUy+fRhg84RCZ+svyugrkzgqdPKP+o3KUczvYdg9/sYvIRQltChgZVkmyZT/Ddf5yVEf5A/U1JGFvnZ5z0RCt0pkTU4uKhBm76M+fD7XktU1sWfxXw). Specifically, set labels to 1000, and toggle the number of characters between 5 and 100. Before this PR, there's a pretty big lag in creating / destroying the glyphs, and panning around is pretty intolerable as well. With this PR, it's noticeably faster (but still much room for improvement, which I plan to address).

Two notes:
- Depending on when you test this, there's another PR (#12896) which may or may not be released yet which significantly speeds up the destruction of glyphs. (Creation is unaffected).
- I believe there's also some caching going on which makes panning behave better over time. So it can be a little hard to discern the performance impact of this PR vs. caching and other performance improvement PRs. This PR's clearest, largest benefit is in the time it takes to create all the glyphs.

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,8 @@
 - Improved performance and reduced memory usage of `Event` class. [#12896](https://github.com/CesiumGS/cesium/pull/12896)
 - Fixes vertical misalignment of glyphs in labels with small fonts [#8474](https://github.com/CesiumGS/cesium/issues/8474)
 - Prevent runtime errors for certain forms of invalid PNTS files [#12872](https://github.com/CesiumGS/cesium/issues/12872)
-- Improved performance of clamped labels [#12905](https://github.com/CesiumGS/cesium/pull/12905)
+- Improved performance of clamped labels. [#12905](https://github.com/CesiumGS/cesium/pull/12905)
+- Fixes issue where multiple instances of a Gaussian splat tileset would transform tile positions incorrectly and render out of position. [#12795](https://github.com/CesiumGS/cesium/issues/12795)
 
 #### Additions :tada:
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -431,3 +431,4 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to contribute to Cesiu
 - [Easy Mahaffey](https://github.com/easymaahffey)
 - [Pamela Augustine](https://github.com/pamelaAugustine)
 - [宋时旺](https://github.com/BlockCnFuture)
+- [Marco Zhan](https://github.com/marcoYxz)
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -123,7 +123,7 @@ function handleBuildWarnings(result) {
 export async function build() {
   // Configure build options from command line arguments.
   const minify = argv.minify ?? false;
-  const removePragmas = argv.pragmas ?? false;
+  const removePragmas = argv.removePragmas ?? false;
   const sourcemap = argv.sourcemap ?? true;
   const node = argv.node ?? true;
 
--- a/packages/engine/Source/Scene/ClippingPolygon.js
+++ b/packages/engine/Source/Scene/ClippingPolygon.js
@@ -46,7 +46,90 @@ function ClippingPolygon(options) {
   //>>includeEnd('debug');
 
   this._ellipsoid = options.ellipsoid ?? Ellipsoid.default;
-  this._positions = [...options.positions];
+  this._positions = copyArrayCartesian3(options.positions);
+
+  /**
+   * A copy of the input positions.
+   *
+   * This is used to detect modifications of the positions in
+   * <code>coputeRectangle</code>: The rectangle only has
+   * to be re-computed when these positions have changed.
+   *
+   * @type {Cartesian3[]|undefined}
+   * @private
+   */
+  this._cachedPositions = undefined;
+
+  /**
+   * A cached version of the rectangle that is computed in
+   * <code>computeRectangle</code>.
+   *
+   * This is only re-computed when the positions have changed, as
+   * determined  by comparing the <code>_positions</code> to the
+   * <code>_cachedPositions</code>
+   *
+   * @type {Rectangle|undefined}
+   * @private
+   */
+  this._cachedRectangle = undefined;
+}
+
+/**
+ * Returns a deep copy of the given array.
+ *
+ * If the input is undefined, then <code>undefined</code> is returned.
+ *
+ * Otherwise, the result will be a copy of the given array, where
+ * each element is copied with <code>Cartesian3.clone</code>.
+ *
+ * @param {Cartesian3[]|undefined} input The input array
+ * @returns {Cartesian3[]|undefined} The copy
+ */
+function copyArrayCartesian3(input) {
+  if (!defined(input)) {
+    return undefined;
+  }
+  const n = input.length;
+  const output = Array(n);
+  for (let i = 0; i < n; i++) {
+    output[i] = Cartesian3.clone(input[i]);
+  }
+  return output;
+}
+
+/**
+ * Returns whether the given arrays are component-wise equal.
+ *
+ * When both arrays are undefined, then <code>true</code> is returned.
+ * When only one array is defined, or they are both defined but have
+ * different lengths, then <code>false</code> is returned.
+ *
+ * Otherwise, returns whether the corresponding elements of the arrays
+ * are equal, as of <code>Cartesian3.equals</code>.
+ *
+ * @param {Cartesian3[]|undefined} a The first array
+ * @param {Cartesian3[]|undefined} b The second array
+ * @returns {boolean} Whether the arrays are equal
+ */
+function equalsArrayCartesian3(a, b) {
+  if (!defined(a) && !defined(b)) {
+    return true;
+  }
+  if (defined(a) !== defined(b)) {
+    return false;
+  }
+  if (a.length !== b.length) {
+    return false;
+  }
+  const n = a.length;
+  for (let i = 0; i < n; i++) {
+    const ca = a[i];
+    const cb = b[i];
+    if (!Cartesian3.equals(ca, cb)) {
+      return false;
+    }
+  }
+  return true;
 }
 
 Object.defineProperties(ClippingPolygon.prototype, {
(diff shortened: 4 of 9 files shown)


packages/engine/Specs/Scene/LabelCollectionSpec.js
```diff
@@ -2446,7 +2446,6 @@ describe("Scene/LabelCollection", function () {
           });
 
           expect(l._clampedPosition).toBeDefined();
-          expect(l._glyphs[0].billboard._clampedPosition).toBeDefined();
```
mzschwartz5, line 2449: I removed this line because:
1. Now that `updateClamping` is skipped for glyph billboards, this value is undefined. On changes to the label's clamped position, it will propagate to the glyph. But not on construction (not that it really matters - only for the sake of this test). 
2. It wasn't a particularly robust test anyway, poking at the internal details of a label's glyph's billboard... we should be doing less of this.

javagl: 👍, `expect(some._nested._private._property).toBe(42)` is an antipattern (not only in tests)

javagl:
Maybe I could derive it by reading the surrounding code more thoroughly. But you likely have all that in the working memory right now, and maybe it's an easy question:

> when true, this flag indicates that the billboard gets its position from some ancestor, and thus expensive operations like clamping can be skipped.

I see that this flag causes an early return in `_updateClamping`. But ... where does anything actually receive a "position from a parent" now? 



---

Totally unrelated: I stumbled over that `_updateClamping` function a few months ago. The context where I stumbled over it: There is an abundant pattern of _wrong ownership directions_. If one had to describe what a `LabelCollection` is, on the level of Object-Oriented Design, then one could start rambling, for hours: How it is constructed. What it consists of. What it can do. How it is used. Nobody would ever say: "A `LabelCollection` contains a `Scene`". That's just wrong. (A `Scene` may contain a `LabelCollection`, but let's skip the question of what a `Scene` is for now). There's an implicit TODO on my list to suggest some refactorings there, but for now, here are some notes from a file on my Desktop that is called `Cesium Scene Everywhere.txt`: 

> - The `scene` is passed to the `LabelCollection` constructor
> - It is stored as a property of the `LabelCollection`, but not _used_ there - only forwarded to the `BillboardCollection` constructor
> - In the `BillboardCollection`, it is stored, as `this._scene`
>   - It is used in the constructor, to attach some event listener to the `scene.terrainProviderChanged` event
>   - It is used in `writeTextureCoordinateBoundsOrLabelTranslate` (whatever _that_ is...), but only to set some flag based on `scene.globe.depthTestAgainstTerrain`
>   - **But** that **private** `_scene` property of `BillboardCollection` is also accessed by `Billboard`!
> 
> 
> - In `Billboard`, it is used in different ways:
>   - It is accessed via `this._billboardCollection._scene.globe._surface.removeTileCustomData` (No kidding...)
>   - It is accessed in `Billboard._updateClamping`, which, let me take a guess, `/** Updates the clamping */`...
>   - This `_updateClamping` throws an error if the scene is not defined...
>   - In `_updateClamping`, it is mainly used to call `scene.updateHeight` and `scene.getHeight`. 

Regardless of that: Avoiding redundant work is good. Further optimizations could be related to reducing the work implied by `updateHeight` and `getHeight`, but these might require further restructurings first.

mzschwartz5:
@javagl 

1. [Here](https://github.com/CesiumGS/cesium/blob/a158b2f01b4064305b579fa824d510731c9c3600/packages/engine/Source/Scene/Label.js#L1132-L1156) is where the glyphs positions get set from the parent (the label). The comment there claims to avoid redundant work, but it wasn't sufficient. 
2. Yes, the patterns ownership and object-oriented flow of information are problematic in quite a few places, but are especially bad with billboards, labels, and their respective collections... but it would take a full refactor of all four (giant) classes to address. 

> Further optimizations could be related to reducing the work implied by updateHeight and getHeight, but these might require further restructurings first.

What do you mean here? There's the [octree picking PR](https://github.com/CesiumGS/cesium/pull/9961), which I intend to pull across the finish line. 

In general, when I think "further optimizations" for labels, I would want to completely restructure it so glyphs aren't each their own billboard. But that would also require significant shader changes... ugh.

javagl:
> What do you mean here? 

On a _very_ high level, there are structural similarities between the clamping performance ( https://github.com/CesiumGS/cesium/issues/12719 ) and _picking_ performance: https://github.com/CesiumGS/cesium/issues/11814. Both functionalities are elaborately wired in without proper abstractions, they use an inefficient implementation, and they are called faaar too often. (And ... maybe I'm misremembering something, but isn't clamping (under certain conditions) even _using_ some `pick` function under the hood, with a ray pointing to the center of the earth...?)

However, similar to the "caching" that was drafted in https://github.com/CesiumGS/cesium/tree/cache-model-picks to address the model picking performance, one could have further options if there was an abstraction for the clamping. The current patterrn of
```javascript
constructor(...scene...) {
   this.scene = scene;
   ..
}
...
setHeightReference(...) {
  this.heightReferece = value;
}
...
doTheClamping() {
  if (this.heightReference !== NONE) {
    if (this.scene === undefined) throw up; // Ouch...
    const height = scene.getHeight(...);
}
```
could benefit from some `HeightProvider`, i.e. something like
```javascript
...
setHeightProvider(...) {
  this.heightProvider = value;
}
...
doTheClamping() {
  if (this.heightProvider !== undefined) {
    const height = this.heightProvider.getHeight(...);
  }
}
```
It would avoid the wrong ownership direction of a `Billboard` using a `Scene`. It would avoid inconsistencies between a `HeightReference` with a missing `Scene`. And it would allow things like wrapping whatever that `HeightProvider` is into a `CachingHeightProvider`, so that it is not necessary to perform the (highly complex and costly) computations that `Scene.getHeight` is doing thousands of times just because the camera moved by one pixel.

ggetz (approved):
This works _much_ better. Thanks @mzschwartz5!

PR #12907, opened 2025-09-22
Quadtree performance

mzschwartz5:
# Description

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/QuadtreeTile.js
+++ b/packages/engine/Source/Scene/QuadtreeTile.js
@@ -108,7 +108,7 @@ function QuadtreeTile(options) {
   this._distance = 0.0;
   this._loadPriority = 0.0;
 
-  this._customData = [];
+  this._customData = new Set();
   this._addedCustomData = [];
   this._removedCustomData = [];
   this._lastSelectionResult = TileSelectionResult.NONE;
@@ -327,7 +327,7 @@ QuadtreeTile.prototype._updateCustomData = function () {
       continue;
     }
 
-    customData.push(data);
+    customData.add(data);
 
     const child = childTileAtPosition(this, data.positionCartographic);
     child._addedCustomData.push(data);
@@ -336,9 +336,8 @@ QuadtreeTile.prototype._updateCustomData = function () {
 
   for (let i = 0; i < removed.length; ++i) {
     const data = removed[i];
-    const index = customData.indexOf(data);
-    if (index !== -1) {
-      customData.splice(index, 1);
+    if (customData.has(data)) {
+      customData.delete(data);
     }
 
     const child = childTileAtPosition(this, data.positionCartographic);
@@ -539,9 +538,9 @@ Object.defineProperties(QuadtreeTile.prototype, {
   },
 
   /**
-   * An array of objects associated with this tile.
+   * A set of objects associated with this tile.
    * @memberof QuadtreeTile.prototype
-   * @type {Array}
+   * @type {Set}
    */
   customData: {
     get: function () {
--- a/packages/engine/Source/Scene/QuadtreeTile.js
+++ b/packages/engine/Source/Scene/QuadtreeTile.js
@@ -1,6 +1,7 @@
 import defined from "../Core/defined.js";
 import DeveloperError from "../Core/DeveloperError.js";
 import Rectangle from "../Core/Rectangle.js";
+import Cartographic from "../Core/Cartographic.js";
 import QuadtreeTileLoadState from "./QuadtreeTileLoadState.js";
 import TileSelectionResult from "./TileSelectionResult.js";
 
@@ -108,7 +109,8 @@ function QuadtreeTile(options) {
   this._loadPriority = 0.0;
 
   this._customData = [];
-  this._frameUpdated = undefined;
+  this._addedCustomData = [];
+  this._removedCustomData = [];
   this._lastSelectionResult = TileSelectionResult.NONE;
   this._lastSelectionResultFrame = undefined;
   this._loadedCallbacks = {};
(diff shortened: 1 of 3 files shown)


packages/engine/Source/Scene/QuadtreeTile.js
```diff
@@ -311,52 +313,66 @@ QuadtreeTile.prototype.clearPositionCache = function () {
...
+      return tile.northwestChild;
+    case 1:
+      return tile.northeastChild;
+    case 2:
+      return tile.southwestChild;
+    default:
+      return tile.southeastChild;
+  }
+}
+
```
mzschwartz5, line 379: Two comments / questions on this:
1. Is this reliable? I don't think it works if the tile crosses the IDL, so is that a bad assumption to make here?
2. Instead of a switch statement, we could define an array out of this function that contains the names of the children and index into that. _But_ that means if the variable name changes, so must the array. I wasn't sure if that was a good tradeoff for a minimal speed boost (if there's any), but input is welcome.

packages/engine/Source/Scene/QuadtreeTile.js
```diff
@@ -311,52 +313,66 @@ QuadtreeTile.prototype.clearPositionCache = function () {
...
+  const added = this._addedCustomData;
+  const removed = this._removedCustomData;
+  if (added.length === 0 && removed.length === 0) {
+    return;
+  }
+
+  const customData = this.customData;
+  for (let i = 0; i < added.length; ++i) {
+    const data = added[i];
+    if (!Rectangle.contains(this._rectangle, data.positionCartographic)) {
```
mzschwartz5: I was just reviewing this code again. This rectangle check currently only does anything useful for level-zero tiles. Level zero tiles are given _all_ custom data, and use this check to figure out which belong to them. All other levels are now _given_ the correct custom data based on the `childTileAtPosition` function I wrote.

But it does mean that most tiles are doing unnecessary work with this `Rectangle.contains` check. I may look into getting rid of it entirely, if it's simple enough (the performance improvement is already good enough either way).

mzschwartz5: (Fixed it in the next revision)

packages/engine/Source/Scene/QuadtreeTile.js
kring: @mzschwartz5 I'm responding here belatedly after you pinged me about this on Teams...

I don't think the IDL should be an issue, because I don't think our quadtree tiles can ever cross it?

But I think there's a bigger problem here.

First, consider a quadtree with a Web Mercator tiling scheme. The four child tiles aren't subdivided at the center latitude at all in that case (they're divided in the center in Web Mercator coordinates, which do not map linearly to latitude), so this check is very wrong.

Even in the Geographic projection, where this seems correct, it might still cause subtle problems. It's a floating point thing, and a bit hard to explain, so bear with me...

If we want to know the coordinates of a particular tile given its Level / X / Y quadtree coordinates, there are two ways we can do that. One is that we can repeatedly subdivide the appropriate tile until we reach the one we care about. Level zero spans -90 to 90 degrees latitude, let's say, so we can divide that level 0 tile at 0 degrees latitude, pick the appropriate level 1 tile from among the four, and then divide that to get our level 2 tiles, and so on. Mathematically, it's repeated division, right?

Or, the other way to do it is to divide the entire coordinate space by the number of tiles in the level, giving us a tile width/height, and then multiply that by the tile coordinate. One division, one multiplication.

Mathematically, they're identical. In floating point land, they're not. The repeated divisions in the first approach can lead to accumulation of rounding errors. After drilling down 20+ levels, those tiny errors get magnified to really actually matter. The "dividing line" of a tile computed with the two approaches can be a significant tile width away. We've seen this happen.

So, as a policy, in the name of both performance and precision, CesiumJS always uses the second approach to compute tile coordinates. The subtle problem is that the code above is effectively determining tile boundaries using the first approach. If there's any possibility of this operation happening repeatedly, and thus of errors accumulating, then it could lead to wrong results.

You should be able to solve both problems (the Web Mercator one, and this subtle one) by letting the tiling scheme compute the center point rather than averaging the coordinates.

javagl: (I don't have much time to zoom into this right now, but ... it raises the question for me about how reliable the `TilingScheme` `positionToTileXY` function is...)

mzschwartz5: Thanks for the input @kring. I believe I understand your explanation. I don't think the floating point math is a problem here, because the logic in `childTileAtPosition` is based off of `tile.rectangle`, [which is calculated in the floating-point safe manner you described.](https://github.com/CesiumGS/cesium/blob/b497b84088c258440f5f9c61b6dd1e34f83e5670/packages/engine/Source/Core/GeographicTilingScheme.js#L163-L191). 

But, it does sound like the real problem is my assumption that four child tiles are always split at the parent's center lat / lon. It sounds like I need to add a new method to the `TilingScheme` interface to get the correct child given a cartographic position - and I need to implement it in each tiling scheme we provide.

Question - any pointers for what that implementation would look like for WebMercator tiling? 

@javagl at a glance, I think the `TilingScheme` `positionToTileXY` function does exactly the floating-point-safe method that Kevin described.

kring: @mzschwartz5 You can compute the correct center by calling `tileXYToRectangle` on any child tile and choosing the coordinates of the appropriate corner (e.g., the southeast corner of the northwest tile). That'll do some extra unnecessary math, of course. I'm not sure offhand if that will be worth optimizing.

mzschwartz5: Ooh, that's a neat idea. Yeah that would work for any tiling scheme - and it's not any extra work, we just replace the getting the parent rectangle (and asking for its center), with getting a child's rectangle (and asking for its center-facing corner).

packages/engine/Specs/Scene/QuadtreeTileSpec.js
```diff
@@ -458,4 +458,53 @@ describe("Scene/QuadtreeTile", function () {
       expect(cachedData).toEqual(dummyPosition);
     });
   });
+
```
mzschwartz5, line 461: Added some unit tests to make sure the update function is doing what it's supposed to, and that it works regardless of tiling scheme...

That said, I think it's bad practice to have a test per tiling scheme here. If someone adds a new one, how would they know we want to test it in `QuadtreeTile` of all places? Perhaps, instead, I should have defined the `childTileAtPosition` function in each tiling scheme instead of in `QuadtreeTile`, and that way each tiling scheme tests itself. 

However, `childTileAtPosition` is currently implemented in such a way that a single implementation _should_ be correct for all tiling schemes... So it doesn't need to be implemented per tiling scheme, and perhaps testing the two existing schemes is sufficient for now and the future? Input welcome.

Issue #12910, opened 2025-09-23
Build failure on Windows for 1.133.1


javagl:
With the current `main` state (i.e [this state](https://github.com/CesiumGS/cesium/tree/a158b2f01b4064305b579fa824d510731c9c3600) as of the time of writing this), the steps of

- cloning the repo
- running `npm install`
- running `npm run build`

cause an error on Windows systems.

The error message is
```
> cesium@1.133.1 build
> gulp build

[16:39:41] Using gulpfile C:\cesium\gulpfile.js
[16:39:41] Starting 'build'...

  Build\CesiumUnminified\index.js       9.5mb
  Build\CesiumUnminified\index.js.map  18.5mb

Done in 348ms

  Build\CesiumUnminified\Cesium.js      13.1mb
  Build\CesiumUnminified\Cesium.js.map  21.7mb

Done in 457ms

  Build\CesiumUnminified\index.cjs       9.6mb
  Build\CesiumUnminified\index.cjs.map  18.5mb

Done in 332ms
[16:39:46] 'build' errored after 4.39 s
[16:39:46] Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:227:11)
    at defaultLoad (node:internal/modules/esm/load:109:3)
    at ModuleLoader.load (node:internal/modules/esm/loader:666:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:479:43)
    at #createModuleJob (node:internal/modules/esm/loader:503:36)
    at #getJobFromResolveResult (node:internal/modules/esm/loader:274:34)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:242:41)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:542:25)
```

The error message (with the "Received protocol 'c:'" part) indicates that a _file path_ is used at a place where a URL is required. This basically means that a call to `pathToFileURL` is missing. 

[...]

PR #12911, opened 2025-09-23
Use proper URLs for import statements

javagl:
This was intended to become a fix for https://github.com/CesiumGS/cesium/issues/12910 , but to avoid merge conflicts given the larger restructuring that seems to be in progress at https://github.com/CesiumGS/cesium/pull/12904, this is now a PR into that state.

This only changes two `import` calls to make sure that they don't receive a **path**, but an actual **url**. 

Ping @jjspace

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -664,7 +664,7 @@ const externalResolvePlugin = {
 export async function getSandcastleConfig() {
   const configPath = "packages/sandcastle/sandcastle.config.js";
   const configImportPath = path.join(projectRoot, configPath);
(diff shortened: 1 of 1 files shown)


ggetz:
Thanks @javagl and @jjspace! Would it be possible to also get this change into `main`? I'm not sure when exactly https://github.com/CesiumGS/cesium/pull/12904 will land, and would like to avoid any scrambling right before release.

jjspace:
Yup, definitely! I was already in the process of doing just that, PR shortly. Edit: https://github.com/CesiumGS/cesium/pull/12912

PR #12913, opened 2025-09-23
Google2D & Azure2D ImageryProvider classes

lukemckinstry:
# Description

There will be multiple ways to add Google2D data - explained in docs 
 - IonImageryProvider.fromAssetId
   - Works exactly the same as all other assets hosted in ion. ion will have 5 Google Maps assets.
- Google2DImageryProvider.fromIonAssetId
  -  similar to IonImageyProvider.fromAssetId, but user can add params to create custom maps and overlays beyond the 5 available in ion. Data still streams via ion.
- Google2DImageryProvider.fromUrl 
   - For users who opt not to stream via ion, sets up session and steams tiles directly from Google api
- Google2DImageryProvider constructor - 
  - docs have note `This object is normally not instantiated directly, use Google2DImageryProvider.fromIon or Google2DImageryProvider.fromUrl.`

### ToDo
 DO NOT MERGE until complete
 - [x] Will 2D Azure go out
   - If no. set class to private, pull sandcastle??, 
 - [x] Prod asset IDs
 - [x] BaseLayerPicker in viewer, which layers to include? (also need to make thumbnails)
 - ~~one-time check for google geocoder? https://github.com/CesiumGS/cesium/pull/12913#discussion_r2376153298~~
 - [x] logo attribution https://github.com/iTwin/curated-data/issues/634#issuecomment-3343464031
 - [x] sandcastle thumbnails - waiting for logo attribution, so that thumbnails have logos  visible 

#### Sandcastles (to publish in prod)
To view the sandcastles, add the key to your Ion Staging account. 
 - Add Google 2D Data [ci-build link](https://ci-builds.cesium.com/cesium/google-azure-2d-imagery/Apps/Sandcastle/index.html?src=Google%202D%20Tiles.html)
 - Add Azure 2D Data [ci-build link](https://ci-builds.cesium.com/cesium/google-azure-2d-imagery/Apps/Sandcastle/index.html?src=Azure%202D%20Tiles.html)
 - Add Google 2D Data with custom styles [ci-build link](https://ci-builds.cesium.com/cesium/google-azure-2d-imagery/Apps/Sandcastle/index.html?src=Google%202D%20Tiles%20with%20Roadmap%20Styles.html)
 - Imagery Assets available in ion [ci-build link](https://ci-builds.cesium.com/cesium/google-azure-2d-imagery/Apps/Sandcastle/index.html?src=Imagery%20Assets%20available%20from%20ion.html)

#### Test all ways to add Google 2D data
 - IonImageryProvider.fromAssetId() - sandcastle 1&2
 - Google2DImageryProvider.fromIonAssetId (with and params to create maps, overlays and custom styles on a roadmap)- sandcastle 3
 - Google2DImageryProvider.fromUrl
   - In sandcastles replace Google2DImageryPrivider `fromIon` with `fromUrl`
   - add Google api key 
```
Cesium.Google2DImageryProvider.fromUrl({
  key: apiKey,
  mapType: "terrain",
}),
```

 - test token refresh
```
//add to bottom of sandcastle using IonImageryProvider.fromAssetId
setTimeout(() => {
  google._imageryProvider._imageryProvider._imageryProvider._resource.setQueryParameters({key: "invalid-key"})
},3000)

//add to bottom of sandcastle using Google2DImageryProvider.fromIon
setTimeout(() => {
  google._imageryProvider._imageryProvider._resource.setQueryParameters({key: "derp"})
},3000)
``` 
 - Check attribution - satellite map has most attribution

# Author checklist


packages/engine/Source/Core/IonResource.js
```diff
@@ -269,7 +275,16 @@ function retryCallback(that, error) {
...
+        // Reset the session token for Google 2D imagery
+        if (newEndpoint.externalType === "GOOGLE_2D_MAPS") {
+          ionRoot.setQueryParameters({ session: newEndpoint.options.session });
+        }
+        if (newEndpoint.externalType === "AZURE_MAPS") {
+          ionRoot.setQueryParameters({
+            "subscription-key": newEndpoint.options["subscription-key"],
+          });
+        }
+        return ionRoot._ionEndpoint;
```
ggetz, line 297: From a code design standpoint, this couples the `IonResource` class to the specific providers. Is this something we can specify externally, such as ~in `IonImageryProvider` and~ passing in a callback like `sessionCreatedCallback`?

**Edit**: No need to route this through `IonImageryProvider`. I think it can be passed to `IonResource` directly.

packages/engine/Source/Scene/IonImageryProvider.js
```diff
@@ -308,7 +322,14 @@ IonImageryProvider.fromAssetId = async function (assetId, options) {
     const options = { ...endpoint.options };
     const url = options.url;
     delete options.url;
-    imageryProvider = await factory(url, options);
+    if (["GOOGLE_2D_MAPS", "AZURE_MAPS"].includes(endpoint.externalType)) {
```
ggetz, line 325: Could we declare `["GOOGLE_2D_MAPS", "AZURE_MAPS"]` as a constant at the top of the file?

packages/engine/Source/Scene/CreditDisplay.js
```diff
@@ -423,13 +423,18 @@ CreditDisplay.prototype.addCreditToNextFrame = function (credit) {
...
+      // Just use the default credit (which is identical) to avoid blinking
+      if (!defined(this._defaultCredit)) {
+        this._defaultCredit = Credit.clone(getDefaultCredit());
+      }
+      this._currentCesiumCredit = this._defaultCredit;
+      return;
     }
-    this._currentCesiumCredit = this._defaultCredit;
+  } catch {
+    //console.log("problem with credit")
```
ggetz: Just a reminder to clean this up. 🙂

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,476 @@
...
+
+const trailingSlashRegex = /\/$/;
+
+/**
+ * <div class="notice">
+ * This object is normally not instantiated directly, use {@link Google2DImageryProvider.fromMapType} or {@link Google2DImageryProvider.fromSessionToken}.
+ * </div>
+ *
+ *
+ * Provides 2D image tiles from Google.
```
ggetz: Let's add the inline doc disclaimer about the Google geocoder, similar to what's in `createGooglePhotorealistic3DTileset.js`.

I'd like to avoid any in-code checks and warnings if possible, but let's defer to @LisaBosCesium may recommend here to ensure we're being compliant with Google's 2D imagery and geocoding.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,476 @@
...
+      return this._imageryProvider.hasAlphaChannel;
+    },
+  },
+});
+
+/**
+ * Creates an {@link ImageryProvider} which provides 2D global tiled imagery from Google.
+ * @param {object} options Object with the following properties:
+ * @param {string} options.apiKey the Google api key
+ * @param {Google2DImageryMapType} options.mapType The map type of the Google map imagery. Valid options are {@link ArcGisBaseMapType.SATELLITE}, {@link ArcGisBaseMapType.OCEANS}, and {@link ArcGisBaseMapType.HILLSHADE}.
```
ggetz: Does it make sense to use a default here, maybe satellite?

ggetz: Also this is referencing `ArcGisBaseMapType`. A reminder to create `Google2DImageryMapType`,  or declare the types inline like [you do below](https://github.com/CesiumGS/cesium/pull/12913/files#diff-7142af57f22c9916aeeac920512bb9d8211e4a6b769201230de03902b119fa1bR223), and use it here.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,476 @@
...
+   */
+  hasAlphaChannel: {
+    get: function () {
+      return this._imageryProvider.hasAlphaChannel;
+    },
+  },
+});
+
+/**
+ * Creates an {@link ImageryProvider} which provides 2D global tiled imagery from Google.
```
ggetz: I'd recommend adding a link to the relevant Google API doc page in this comment block so users can look into more details around the relevant parameters.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,476 @@
...
+ * @returns {Promise<Google2DImageryProvider>} A promise that resolves to the created Google2DImageryProvider.
+ *
+ * @example
+ * // Google 2D imagery provider
+ * const googleTilesProvider = Cesium.Google2DImageryProvider.fromMapType({
+ *     apiKey: 'thisIsMyApiKey',
+ *     mapType: "SATELLITE",
+ *     language: "en_US",
+ *     region: "US"
+ * });
```
ggetz, line 314: I'd recommend providing one option with the "bare minimum"—only the required parameters—and one with other with a more "curated" styling example.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,468 @@
...
+      "valid values for options.mapType satellite, terrain, or roadmap",
+    );
+  }
+  //>>includeEnd('debug');
+
+  //>>includeStart('debug', pragmas.debug);
+  if (defined(options.overlayLayerType)) {
+    if (
+      !["layerRoadmap", "layerStreetview", "layerTraffic"].includes(
+        options.overlayLayerType,
```
ggetz: Prefer using `Check.defined`

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,468 @@
...
+      )
+    ) {
+      throw new DeveloperError(
+        "valid values for options.overlayLayerType are layerRoadmap, layerStreetview or layerTraffic",
+      );
+    }
+  }
+  //>>includeEnd('debug');
+
```
ggetz: This is a minor comment, but I don't think we need to explicitly check for this.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,468 @@
...
+  //>>includeStart('debug', pragmas.debug);
+  if (!defined(options.tileWidth)) {
+    throw new DeveloperError("options.tileWidth is required.");
+  }
+  //>>includeEnd('debug');
+
+  //>>includeStart('debug', pragmas.debug);
+  if (!defined(options.tileHeight)) {
```
ggetz: Would it be likely that users call this directly? Should it be marked as `private`?

packages/engine/Source/Scene/Azure2DImageryProvider.js
```diff
@@ -0,0 +1,313 @@
...
+ */
+
+/**
+ * Provides 2D image tiles from Azure.
+ *
+ * @alias Azure2DImageryProvider
+ * @constructor
+ *
+ * @param {Azure2DImageryProvider.ConstructorOptions} options Object describing initialization options
+ *
```
ggetz, line 34: Should we mark this as `private` until we are ready to "ship" Azure support?

packages/engine/Source/Scene/Google2DImageryProvider.js
lukemckinstry: I moved this logic to call Google2DImageryProvider with a session token to the constructor. I put a note at the top of the docs `his object is normally not instantiated directly, use Google2DImageryProvider.fromIon or[Google2DImageryProvider.fromUrl. We use it internally, I guess a user could too if they want.

packages/engine/Source/Scene/Google2DImageryProvider.js
lukemckinstry: added disclaimer, put note in ToDos about code check

packages/engine/Source/Scene/Google2DImageryProvider.js
lukemckinstry: got rid of ArcGisMapType, and made `satellite` the default. I am not sure what your creation you were referring to, it might have gotten taken out, let me know if/how to add it back.

packages/engine/Source/Scene/Azure2DImageryProvider.js
lukemckinstry: added this as a ToDo pending go/no-go decision

packages/engine/Source/Core/IonResource.js
lukemckinstry: it is coupling to `externalType`, the attribute on ion assets, which yes in this case is specific to the Google and Azure providers. We do however already have logic in IonResource tied to `externalType`s, so one could make the case it OK to use them here too.

packages/engine/Source/Scene/IonImageryProvider.js
lukemckinstry: we only couple GOOGLE_2D_MAPS and AZURE_MAPS together in this one instance, the other place we call out the types they have different logic

Apps/Sandcastle/gallery/Google 2D Tiles with Custom Styles.html
ggetz, line 1: Once we've settled on the example and code itself, @jjspace has a script that will convert these to the new Sandcastle format. 

Please ask @jjspace to advise on that process once your ready @lukemckinstry.

packages/engine/Source/Core/IonResource.js
ggetz: Yes, but now we're starting to get into implementation specific details. Wouldn't it make sense for that to live in the same place where we're specifying similar details—like in the imagery provider?

Consider as more external services are added: Do we want to have one huge switch statement for each imagery provider? Or should we allow the imagery provider to define this behavior against a common interface?

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+import defined from "../Core/defined.js";
+import DeveloperError from "../Core/DeveloperError.js";
+import Resource from "../Core/Resource.js";
+import IonResource from "../Core/IonResource.js";
+//import Rectangle from "../Core/Rectangle.js";
+import Check from "../Core/Check.js";
+import UrlTemplateImageryProvider from "./UrlTemplateImageryProvider.js";
+import TileAvailability from "../Core/TileAvailability.js";
+//import CesiumMath from "../Core/Math.js";
+//import Cartographic from "../Core/Cartographic.js";
```
ggetz: Please cleanup any unused imports.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+
+const trailingSlashRegex = /\/$/;
+
+/**
+ * @typedef {Object} Google2DImageryProvider.ConstructorOptions
+ *
+ * Initialization options for the Google2DImageryProvider constructor
+ *
+ * @property {object} options Object with the following properties:
+ * @property {string} options.key The key to send with tile requests. If using Cesium ion this must be the key issued by ion. If using Google this must be the Google api key.
```
ggetz: I'm not sure users will have the context for Cesium ion vs Google API.

```suggestion
 * @property {string} options.key The Google API key to send with tile requests.
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @property {string} options.session The Google session token that tracks the current state of your map and viewport.
+ * @property {(string|IonResource)} options.url The IonResource for the ion proxy endpoint is using Cesium ion, or the url for the google API tile service
```
ggetz:
```suggestion
 * @property {string|Resource|IonResource} options.url The Google 2D Maps endpoint
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @property {string} options.tileWidth The width of each tile in pixels.
+ * @property {string} options.tileHeight The height of each tile in pixels.
+ * @property {Ellipsoid} [options.ellipsoid=Ellipsoid.default] The ellipsoid.  If not specified, the default ellipsoid is used.
+ * @property {number} [options.minimumLevel=0] The minimum level-of-detail supported by the imagery provider.  Take care when specifying
+ *                 this that the number of tiles at the minimum level is small, such as four or less.  A larger number is likely
+ *                 to result in rendering problems.
+ * @property {number} [options.maximumLevel] The maximum level-of-detail supported by the imagery provider, or undefined if there is no limit.
```
ggetz:
```suggestion
 * @property {number} [options.maximumLevel=22] The maximum level-of-detail supported by the imagery provider.
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @property {Rectangle} [options.rectangle=Rectangle.MAX_VALUE] The rectangle, in radians, covered by the image.
+ * @example
+ * // Google 2D imagery provider
+ * const googleTilesProvider = Cesium.Google2DImageryProvider({
+ *     key: 'thisIsMyApiKey',
+ *     session: 'thisIsSessionToken'
+ * });
```
ggetz: Generally, make sure to provide code examples based on how we recommend using the API, i.e., instead of 

```
 * const googleTilesProvider = Cesium.Google2DImageryProvider({
 *     key: 'thisIsMyApiKey',
 *     session: 'thisIsSessionToken'
 * });
```

prefer

```
 * const googleTilesProvider = Cesium.Google2DImageryProvider.fromUrl({
 *     key: "thisIsMyApiKey",
 *     mapType: "SATELLITE"
 * });
```

However, this is the typedef for the options object, so it shouldn't need an example at all.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * // Google 2D imagery provider
+ * const googleTilesProvider = Cesium.Google2DImageryProvider.fromIon({
+ *     assetId: 1687
+ * });
+  * @example
+ * // Google 2D imagery provider
+ * const googleTilesProvider = Cesium.Google2DImageryProvider.fromUrl({
+ *     key: 'thisIsMyApiKey',
+ *     mapType: "SATELLITE"
+ * });
```
ggetz: To be consistent with Google 3D Photorealistic Tiles, I'd suggest using the same workflow with `GoogleMaps.defaultApiKey` as we use in `createGooglePhotorealistic3DTileset.js`:

```suggestion
 *  * // Use your own Google Maps API key
 * Cesium.GoogleMaps.defaultApiKey = "your-api-key";
 * 
 * const googleTilesProvider = Cesium.Google2DImageryProvider.fromUrl({
 *     mapType: "SATELLITE"
 * });
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @example
+ * // Google 2D imagery provider
+ * const googleTilesProvider = Cesium.Google2DImageryProvider.fromUrl({
+ *     apiKey: 'thisIsMyApiKey',
+ *     mapType: "satellite"
+ * });
+ * @example
+ * // Google 2D roadmap overlay with custom styles
+ * const googleTileProvider = Cesium.Google2DImageryProvider.fromUrl({
+ *     assetId: 1687,
```
ggetz: This looks like a copy/paste error from `fromIon`.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+    tileHeight: endpointOptions.tileHeight,
+    url: new IonResource(endpoint, endpointResource),
+    endpointOptions,
+  });
+};
+
+/**
+ * Creates an {@link ImageryProvider} which provides 2D global tiled imagery from {@link https://developers.google.com/maps/documentation/tile/2d-tiles-overview|Google 2D Tiles}.
+ * @param {object} options Object with the following properties:
+ * @param {string} options.assetId The assetId to call in Cesium ion. This must be an Imagery asset with externalType = "GOOGLE_2D_MAPS".
```
ggetz:
```suggestion
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @param {string} options.key The Google api key
```
ggetz: To follow up on my comment above above about `GoogleMaps.defaultApiKey`, I would make the behavior here:


```suggestion
 * @param {string} [options.key=GoogleMaps.defaultApiKey] Your API key to access Google Photorealistic 3D Tiles. See {@link https://developers.google.com/maps/documentation/javascript/get-api-key} for instructions on how to create your own key.
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @param {"satellite" | "terrain" | "roadmap"} [options.mapType="satellite"] The map type of the Google map imagery. Valid options are satellite, terrain, and roadmap. If overlayLayerType is set, mapType is ignored and a transparent overlay is returned. If overlayMapType is undefined, then a basemap of mapType is returned. layerRoadmap overlayLayerType is included in terrain and roadmap mapTypes.
+ * @param {string} [options.language='en_US'] an IETF language tag that specifies the language used to display information on the tiles
+ * @param {string} [options.region='US'] A Common Locale Data Repository region identifier (two uppercase letters) that represents the physical location of the user.
+ * @param {"layerRoadmap" | "layerStreetview" | "layerTraffic"} [options.overlayLayerType=undefined] Returns a transparent overlay map with the specified layerType. If no value is provided, a basemap of mapType is returned. Use multiple instances of Google2DImageryProvider to add multiple Google Maps overlays to a scene. layerRoadmap is included in terrain and roadmap mapTypes, so adding as overlay to terrain or roadmap has no effect.
```
ggetz:
```suggestion
 * @param {"layerRoadmap" | "layerStreetview" | "layerTraffic"} [options.overlayLayerType] Returns a transparent overlay map with the specified layerType. If no value is provided, a basemap of mapType is returned. Use multiple instances of Google2DImageryProvider to add multiple Google Maps overlays to a scene. layerRoadmap is included in terrain and roadmap mapTypes, so adding as overlay to terrain or roadmap has no effect.
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ *             featureType: "road",
+ *             elementType: "geometry",
+ *             stylers: [{ lightness: 100 }, { visibility: "simplified" }],
+ *         },
+ *     ],
+ * });
+ *
+ */
+
+Google2DImageryProvider.fromIon = async function (options) {
```
ggetz: `fromIonAssetId` is the naming convention, as per `Cesium3DTileset.js`, `CesiumTerrainProvider.js`, etc.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ *         {
+ *             featureType: "road",
+ *             elementType: "geometry",
+ *             stylers: [{ lightness: 100 }, { visibility: "simplified" }],
+ *         },
+ *     ],
+ * });
+ *
+ */
+
```
ggetz:
```suggestion
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+    get: function () {
+      return this._imageryProvider.hasAlphaChannel;
+    },
+  },
+});
+
+/**
+ * Creates an {@link ImageryProvider} which provides 2D global tiled imagery from {@link https://developers.google.com/maps/documentation/tile/2d-tiles-overview|Google 2D Tiles}, streamed using the Cesium ion REST API.
+ * @param {object} options Object with the following properties:
+ * @param {string} options.assetId The assetId to call in Cesium ion. This must be an Imagery asset with externalType = "GOOGLE_2D_MAPS".
```
ggetz:
```suggestion
 * @param {string} options.assetId The Cesium ion asset id.
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @param {"satellite" | "terrain" | "roadmap"} [options.mapType="satellite"] The map type of the Google map imagery. Valid options are satellite, terrain, and roadmap. If overlayLayerType is set, mapType is ignored and a transparent overlay is returned. If overlayMapType is undefined, then a basemap of mapType is returned. layerRoadmap overlayLayerType is included in terrain and roadmap mapTypes.
+ * @param {string} [options.language='en_US'] an IETF language tag that specifies the language used to display information on the tiles
```
ggetz:
```suggestion
 * @param {string} [options.language="en_US"] an IETF language tag that specifies the language used to display information on the tiles
```

I noticed inconsistent use of single and double quotes throughout the doc comments. Please do a pass and make sure to use double quotes everywhere.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+  options.region = options.region ?? "US";
+
+  const overlayLayerType = options.overlayLayerType;
+  //>>includeStart('debug', pragmas.debug);
+  if (defined(overlayLayerType)) {
+    Check.typeOf.string("overlayLayerType", overlayLayerType);
+  }
+  if (!defined(options.assetId)) {
+    throw new DeveloperError("options.assetId is required.");
+  }
```
ggetz:
```suggestion
Check.defined("options.assetId", options.assetId);
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+Google2DImageryProvider.fromIon = async function (options) {
+  options = options ?? {};
+  options.mapType = options.mapType ?? "satellite";
+  options.language = options.language ?? "en_US";
+  options.region = options.region ?? "US";
+
+  const overlayLayerType = options.overlayLayerType;
+  //>>includeStart('debug', pragmas.debug);
+  if (defined(overlayLayerType)) {
+    Check.typeOf.string("overlayLayerType", overlayLayerType);
```
ggetz:
```suggestion
    Check.typeOf.string("options.overlayLayerType", overlayLayerType);
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+  if (!defined(options.tileHeight)) {
+    throw new DeveloperError("options.tileHeight is required.");
+  }
+  //>>includeEnd('debug');
+
+  //>>includeStart('debug', pragmas.debug);
+  if (!defined(options.key)) {
+    throw new DeveloperError("options.key is required.");
+  }
+  //>>includeEnd('debug');
```
ggetz:
```suggestion
  //>>includeStart('debug', pragmas.debug);
  Check.defined("options.session", options.session);
  Check.defined("options.tileWidth", options.tileWidth);
  Check.defined("options.tileHeight", options.tileHeight);
  Check.defined("options.key", options.key);
  //>>includeEnd('debug');
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @param {"satellite" | "terrain" | "roadmap"} [options.mapType="satellite"] The map type of the Google map imagery. Valid options are satellite, terrain, and roadmap. If overlayLayerType is set, mapType is ignored and a transparent overlay is returned. If overlayMapType is undefined, then a basemap of mapType is returned. layerRoadmap overlayLayerType is included in terrain and roadmap mapTypes.
+ * @param {string} [options.language='en_US'] an IETF language tag that specifies the language used to display information on the tiles
+ * @param {string} [options.region='US'] A Common Locale Data Repository region identifier (two uppercase letters) that represents the physical location of the user.
+ * @param {"layerRoadmap" | "layerStreetview" | "layerTraffic"} [options.overlayLayerType=undefined] Returns a transparent overlay map with the specified layerType. If no value is provided, a basemap of mapType is returned. Use multiple instances of Google2DImageryProvider to add multiple Google Maps overlays to a scene. layerRoadmap is included in terrain and roadmap mapTypes, so adding as overlay to terrain or roadmap has no effect.
```
ggetz:
```suggestion
 * @param {"layerRoadmap" | "layerStreetview" | "layerTraffic"} [options.overlayLayerType] Returns a transparent overlay map with the specified layerType. If no value is provided, a basemap of mapType is returned. Use multiple instances of Google2DImageryProvider to add multiple Google Maps overlays to a scene. layerRoadmap is included in terrain and roadmap mapTypes, so adding as overlay to terrain or roadmap has no effect.
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+
+  this._session = options.session;
+  this._key = options.key;
+  this._tileWidth = options.tileWidth;
+  this._tileHeight = options.tileHeight;
+
+  const resource =
+    options.url instanceof IonResource
+      ? options.url
+      : Resource.createIfNeeded(options.url ?? "https://tile.googleapis.com");
```
ggetz: Can we use `GoogleMaps.mapTilesApiEndpoint` instead? Or tweak the `/v1/` postfix to be added only in `createGooglePhotorealistic3DTileset`?

packages/engine/Source/Core/IonResource.js
lukemckinstry: I definitely agree with avoiding large switch statements, and the code is starting to trend in that direction. I also agree passing to IonResource makes more sense than to IonImageryProvider.

We do call IonResource from different places - IonImageryProvider.fromAssetID and Google2DImageryProvider - so we would need logic to pass in the callback from both places.

Then in IonResource we need to differentiate based on type both 1) the retryCallback and 2)  the `options` passed to `Resource.call(this, options);` in the constructor (currently determined by looking at `externalType`). 

I like the idea but this doesn't feel like a clear cut simplification to me, at least not now.

Apps/Sandcastle/gallery/Google 2D Tiles with Custom Styles.html
lukemckinstry: I think I created the "new" format sandcastles from scratch correctly after checking with @jjspace last week about the process. But @jjspace if you don't mind please have a look and let me know if things look ok :pray:

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+ * @param {number} x The tile X coordinate.
+ * @param {number} y The tile Y coordinate.
+ * @param {number} level The tile level;
+ * @returns {Credit[]} The credits to be displayed when the tile is displayed.
+ */
+Google2DImageryProvider.prototype.getTileCredits = function (x, y, level) {
+  return undefined;
+};
+
+//const rectangleScratch = new Rectangle();
```
ggetz: Remove?

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+  this._imageryProvider = provider;
+
+  this.getViewportCredits({
+    viewportUrl: this._viewportUrl,
+    key: this._key,
+    session: this._session,
+    maximumLevel: this._maximumLevel,
+  }).then((data) => {
+    this._imageryProvider._credit = new Credit(data);
+  });
```
ggetz: 1. Let's try to avoid using internal properties, such as `_credit` when we can. Instead, lets have this class store it's own credit object, `this._credit`, which can be returned directly in the `credit` getter.

```
credit: {
    get: function () {
      return this._credit;
    },
  },
```

2. _However_, this may be more appropriate to call in `getTileCredits`, as `credit` is typically one credit for the entire provider, and `getTileCredits` is equipped to handle many asynchronous credits and de-dupe as needed.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+   * @readonly
+   */
+  tileDiscardPolicy: {
+    get: function () {
+      return this._imageryProvider.tileDiscardPolicy;
+    },
+  },
+
+  /**
+   * Gets an event that is raised when the imagery provider encounters an asynchronous error..  By subscribing
```
ggetz:
```suggestion
   * Gets an event that is raised when the imagery provider encounters an asynchronous error.  By subscribing
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+  options.region = options.region ?? "US";
+
+  const overlayLayerType = options.overlayLayerType;
+  //>>includeStart('debug', pragmas.debug);
+  if (defined(overlayLayerType)) {
+    Check.typeOf.string("overlayLayerType", overlayLayerType);
+  }
+  if (!defined(options.key)) {
+    throw new DeveloperError("options.key is required.");
+  }
```
ggetz, line 439:
```suggestion
  }
Check.defined("options.key", options.key);
```

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+  x,
+  y,
+  level,
+  longitude,
+  latitude,
+) {
+  return undefined;
+};
+
+Google2DImageryProvider.prototype.getViewportCredits = async function (
```
ggetz: Please document, but mark as `private`.

packages/engine/Source/Scene/Google2DImageryProvider.js
```diff
@@ -0,0 +1,627 @@
...
+  return queryOptions;
+}
+
+async function createGoogleImagerySession(options) {
+  const { language, region, key } = options;
+
+  const queryOptions = buildQueryOptions(options);
+
+  const response = await Resource.post({
+    url: "https://tile.googleapis.com/v1/createSession",
```
ggetz: You'll want to use the base endpoint specified in `options.url`, for which you can default to `GoogleMaps.mapTilesApiEndpoint`.

packages/engine/Source/Scene/Google2DImageryProvider.js
lukemckinstry: My thinking was ideally the credits from `getViewportCredits` would be passed downstream from `Google2DImageryProvider` to `UrlTemplateImageryProvider` in `options.credit`, but `getViewportCredits` is async and the `Google2DImageryProvider` constructor is synchronous. 

From looking at a few other ImageryProviders that also pass all their info downstream to UrlTemplateImageryProvider, it does not seem like there is a way for UrlTemplateImageryProvider to call properties like credits from an "upstream" ImageryProvider.

For getTileCredits, I think the plan is we want something that runs just once when the ImageryProvider is loaded as opposed to every frame, so I put the process in the constructor.

lukemckinstry:
@ggetz thanks for the PR feedback. I addressed all the comments from the latest review.

packages/engine/Source/Core/GoogleMaps.js
```diff
@@ -28,7 +28,7 @@ GoogleMaps.defaultApiKey = undefined;
  * @default https://tile.googleapis.com/v1/
```
ggetz:
```suggestion
 * @default https://tile.googleapis.com/
```

Apps/Sandcastle/gallery/Google 2D Tiles with Custom Styles.html
jjspace: As a quick sanity check I ran the convert script on the 4 new sandcastles and they lined up perfectly! The only thing different was the removed Beginner label which we don't want in the new sandcastle anyway.

All good here! :+1:

lukemckinstry: That's good to hear. I marked the sandcastles in this PR with the "showcase" label because I found an empty "labels" field or omitting the labels header threw an error in the build process. 

1) I am not sure "showcase" is the right label for these, does anybody have suggestions?
2) Is the requirement/expectation to always have a label intentional or a bug?

ggetz:
@lukemckinstry Pushed up a fix/refactor to how we're handling `IonResource`s and credits, as discussed. Please let me know if there are any questions.

Apps/Sandcastle/gallery/Azure 2D Tiles.html
```diff
@@ -0,0 +1,75 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"
+    />
+    <meta
+      name="description"
+      content="Use Viewer to start building new applications or easily embed Cesium into existing applications."
```
ggetz: Make sure to update the descriptions, both here and for the `sandcastle.yaml` files.

Apps/Sandcastle/gallery/Google 2D Tiles with Custom Styles.html
jjspace: @ggetz can confirm but I'd assume these can probably fall under the `Imagery` label?

It's intentional that every sandcastle requires a label so it can be categorized in the gallery. Maybe there's an argument for some sandcastles to not have labels but I don't know if that would actually be very useful.

lukemckinstry:
> @lukemckinstry Pushed up a fix/refactor to how we're handling `IonResource`s and credits, as discussed. Please let me know if there are any questions.

@ggetz 
Looks like changing the attribution list on the fly is causing problems for our credits popup. It does not resize when the length of of the  credits content changes, unless the user resizes the browser viewport. 


https://github.com/user-attachments/assets/d5c09727-46c5-403a-bb65-f3f31f8f2b69

Going back to the static list instead of the "level-specific"  list would be an easy fix for this.

ggetz:
> Looks like changing the attribution list on the fly is causing problems for our credits popup.

Ah, that's no good. However, and correct me if I'm mistaken, this issue is not unique to the Google imagery provider–which is now using the more standard method of attribution.

Non-ideal for sure, but I don't think it need to hold up this PR. @lukemckinstry Could you open a new issue for the data attribution popup, assuming there is not already one open?

Issue #9660, opened 2021-07-02
Scene#drillPick performance scales poorly


thw0rted:
[Sandcastle example](https://sandcastle.cesium.com/#c=bVTvb9owEP1XTqgfnC0zlPXD1ALaBkyb1m7VQNsXpGHig5o5dhQ7tAzlf9/lByR0/RTn7t27d8+XRNY4DzuFj5jCEAw+whidymL+s4yxRScq38fWeKEMpotOCIeFAXARGryzEq+PJbNjhM/G02/T/iQscF7FqKnyGtZCOyxjwqhYeGVNE8yDm4VZmKgUdKia/lJygz6sWuWkrxLaAiarc9X3ynzMlJakvOBb2xSYRg+KYL0begygXzxfDy+DaowTZFtBtjVk20Cg7svReOUVOi6kZHUKILFOVbPUIsYi9XQS5i1fpzae4CZFdKzHL+EVqBCqwzYIjwwrpfXKilRew4kVgCzakGvJqmSZ45Nny4uDysOLwzZfhqduVtuU/5hOQrh613Dm9an0lR7kcGFbPcla7+eWPRurvoGW9dxFJN3MEhHhdEe4z8JITSUO/ReTZP5DVEzO2OFoQh7AcFRNUdyQ1ci13bDlvYr+KLOBFfpHYoRe2KMtkHBxquRP3Nv7FCPlCsqroBj0lNw/Ty6ruao1iJUkWeercLqFPmtadPshNJzdfpvEeaqoOSbCIws4WTCn9WVtWIoFqFxJLlO6uWI0VikIITMS17TtstXnqd2zoTrzBgsriJprNBv/kEOijANlKHon/ANdmLUpe0kavKmUB3nseOlKHjZf5Pn1zfcJ8tvpp/nv8e2X8dfgphN2Bs7vNY6Oe/NexYklH7JUM867HuNEU0fXXWWk0vPIlXsy6LbLBlLtQMnhC38LiLRwjjLrTOuZ+ouLzmjQJfx/pdoKSSvyfYepFvsC9nA5uq2CnPNBl15frvTW6pWgZqcpmpyinwAlCh6Ux/I2zT8): Open, then click anywhere on the map to select between the top-left corner of the canvas and the clicked location.  The console tab will log the number of pins included (to make sure the correct point was used) and the time in milliseconds spent in the call to `drillPick`.

Browser: Chrome stable, latest

Operating System: Win10 latest

----

[...]

PR #12916, opened 2025-09-24
drill picking performance

Beilinson:
# Description

The goal of this PR is to improve performance with `scene.drillPick`, without changing the algorithm significantly.

The first commit changing the internal data structure of `_pickObjects` to `Map` was inspired by the performance boosts seen in #12896, and led to a surprising 75% decrease in time.

I tried further reducing unneeded code by removing some funny `byte-float-byte` transformations and reusing a scratch array for pixels in `Context.readPixels`, which also led to around ~10% improvements over the previous iteration.

Further, I realized that in sandcastles such as the one described in #9660, where may non-fully overlapping entities may be drawn, and the drill pick rectangle may be particularly large, instead of `PickFramebuffer.end` exiting early after one entity was found, it could continue counting entities and iterate over the entire pickPolygon. This led to the most significant performance boost, the attached sandcastle now takes ~30ms down from ~3s on my machine, since the slowest part of picking is the need to rerender for every single entity in the pick rectangle currently (i.e, 220 entities = 220 renders + 220 buffer reads + 220 iterations over bigger and bigger parts of the image)

## Issue number and link

#9660 

## Testing plan

Sandcastle example modified from #9660:
[local](http://localhost:8080/Apps/Sandcastle/index.html#c=bVTbctowEP2VHSYPcuvKQJNMhtt0CnnrJZNk2hdmim0tRFSWqCQnUMb/3vUFcFKeJK3OHp09u3ZqtPPwLPEFLYxB4wtM0ck84z+qGJt30uo8NdrHUqOdd0LYzzWAS1HjVyNwcEh5OET4w/T2221/FpY4LzNUlDmAZawcVrFYyyz20uh20OKfHJ2/Ry3Q1sTe5uVdEQzneq7TSuy+FvRTihX6sJZRkPa6iBZwk7yu6E7qz7lURM5KvqWxwBR6kATrDmkZQb9c3497QV3iEbKuIesGsj5BoHmXo/bSS3Q8FoI1VwAb42RdZyNiGltPu1h/5EtrshmuLKJjXd6DdyBDqDfrIDwwJFKpxMRWDODICkD2rcifTVKxPOLWs8XFXhbhxX5dLMLja0YZy+9vZyFc3pw4i2ZX+UoLOVza1lSyVLtHw96URU0HkdumaT1oWuLQP1J/Te4ZC2A8qTU2/je1k3d72A7gun8Two6Sr6+gGB5wRiFXZsUWdzL9LfUKEvQviBq6YZfmRMDF/kDEt9ybO4updHRil0FZ7vFy9/ZyEQxPYjIphMLXA3HsRZ+dnoj6IZw4o36bxHnKaDhmsUcW8FVtAGvDLJagajC5sNS/sjRWKwghp/Fe0vcgWu9s228G573B0gqi5gr1yj8VsJHagdQU/Rr7J2qbMZadkwYfauVBkTleuVKEcPWr2+0Gw07YGTm/Uzg5DMcnmW0MlZlbxTiPPGYbRYQuSnIS4XnqXK0QYBS1U0dCPoMU4zO/DEhV7BzdLHOlHuRfnHcmo4jw/6UqEwuagu/PaFW8K2FPvcmXOsg5H0V0PJ/pjVFJbN8w/wM)

image

[main](https://sandcastle.cesium.com/#c=bVTbctowEP2VHSYPcuvKQJNMhtt0CnnrJZNk2hdmim0tRFSWqCQnUMb/3vUFcFKeJK3OHp09u3ZqtPPwLPEFLYxB4wtM0ck84z+qGJt30uo8NdrHUqOdd0LYzzWAS1HjVyNwcEh5OET4w/T2221/FpY4LzNUlDmAZawcVrFYyyz20uh20OKfHJ2/Ry3Q1sTe5uVdEQzneq7TSuy+FvRTihX6sJZRkPa6iBZwk7yu6E7qz7lURM5KvqWxwBR6kATrDmkZQb9c3497QV3iEbKuIesGsj5BoHmXo/bSS3Q8FoI1VwAb42RdZyNiGltPu1h/5EtrshmuLKJjXd6DdyBDqDfrIDwwJFKpxMRWDODICkD2rcifTVKxPOLWs8XFXhbhxX5dLMLja0YZy+9vZyFc3pw4i2ZX+UoLOVza1lSyVLtHw96URU0HkdumaT1oWuLQP1J/Te4ZC2A8qTU2/je1k3d72A7gun8Two6Sr6+gGB5wRiFXZsUWdzL9LfUKEvQviBq6YZfmRMDF/kDEt9ybO4updHRil0FZ7vFy9/ZyEQxPYjIphMLXA3HsRZ+dnoj6IZw4o36bxHnKaDhmsUcW8FVtAGvDLJagajC5sNS/sjRWKwghp/Fe0vcgWu9s228G573B0gqi5gr1yj8VsJHagdQU/Rr7J2qbMZadkwYfauVBkTleuVKEcPWr2+0Gw07YGTm/Uzg5DMcnmW0MlZlbxTiPPGYbRYQuSnIS4XnqXK0QYBS1U0dCPoMU4zO/DEhV7BzdLHOlHuRfnHcmo4jw/6UqEwuagu/PaFW8K2FPvcmXOsg5H0V0PJ/pjVFJbN8w/wM)

image

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Picking.js
+++ b/packages/engine/Source/Scene/Picking.js
@@ -790,11 +790,11 @@ function addDrillPickedResults(
 
 /**
  * Drill pick by repeatedly calling a given `pickCallback`, each time stripping away the previously picked objects.
- * @param {number} [limit=Number.MAX_VALUE] If supplied, stop drilling after collecting this many picks
  * @param {function(number): object[]} pickCallback Pick callback to execute each iteration
+ * @param {number} [limit=Number.MAX_VALUE] If supplied, stop drilling after collecting this many picks
  * @returns {object[]} List of picked results
  */
-function drillPick(limit, pickCallback) {
+function drillPick(pickCallback, limit) {
   // PERFORMANCE_IDEA: This function calls each primitive's update for each pass. Instead
   // we could update the primitive once, and then just execute their commands for each pass,
   // and cull commands for picked primitives.  e.g., base on the command's owner.
@@ -863,7 +863,7 @@ Picking.prototype.drillPick = function (
       exclude: false,
     }));
   };
-  const objects = drillPick(limit, pickCallback);
+  const objects = drillPick(pickCallback, limit);
   return objects.map((element) => element.object);
 };
 
@@ -1120,7 +1120,7 @@ function drillPickFromRay(
     );
     return pickResult ? [pickResult] : undefined;
   };
-  return drillPick(limit, pickCallback);
+  return drillPick(pickCallback, limit);
 }
 
 function pickFromRay(
--- a/Specs/TypeScript/tsconfig.json
+++ b/Specs/TypeScript/tsconfig.json
@@ -3,7 +3,8 @@
     "noEmit": true,
     "types": [],
     "strict": true,
-    "isolatedModules": true
+    "isolatedModules": true,
+    "lib": ["ES2020", "DOM"]
   },
   "include": [
     "index.ts",
--- a/Tools/jsdoc/tsconfig.json
+++ b/Tools/jsdoc/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
-      "noEmit": true,
-      "types": []
+    "noEmit": true,
+    "types": [],
+    "lib": ["ES2020", "DOM"]
   },
   "files": [
     "../../Source/Cesium.d.ts"
--- a/packages/engine/Source/Renderer/Context.js
+++ b/packages/engine/Source/Renderer/Context.js
@@ -1580,7 +1580,7 @@ Context.prototype.getObjectByPickColor = function (pickColor) {
 
 /**
  *
- * @param {Map} pickObjects
+ * @param {Map<number, object>} pickObjects
  * @param {number} key
  * @param {Color} color
  */


packages/engine/Specs/Scene/Model/PickingPipelineStageSpec.js
```diff
@@ -150,8 +150,8 @@ describe(
           "uniform vec4 czm_pickColor;",
         ]);
 
-        const pickObject =
-          context._pickObjects[Object.keys(context._pickObjects)[0]];
+        console.log(context._pickObjects);
```
mzschwartz5: Looks like this log statement was left here accidentally

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -48,15 +48,14 @@ PickFramebuffer.prototype.begin = function (screenSpaceRectangle, viewport) {
...
- * Return the picked object rendered within a given rectangle.
+ * Return the picked objects rendered within a given rectangle.
  *
  * @param {BoundingRectangle} screenSpaceRectangle
- * @returns {object|undefined} The object rendered in the middle of the rectangle, or undefined if nothing was rendered.
+ * @param {number} [limit=1] If supplied, stop iterating after collecting this many objects.
+ * @returns {object[]} A list of rendered objects, ordered by distance to the middle of the rectangle.
  */
-PickFramebuffer.prototype.end = function (screenSpaceRectangle) {
+PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit) {
```
mzschwartz5: You can set a default value for an argument directly in the function signature, in javascript. So, in this case:
`PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1)`

rather than `limit = limit ?? 1`

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -93,22 +95,19 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle) {
...
 
-      const object = context.getObjectByPickColor(
-        colorScratchForPickFramebuffer,
-      );
+      const object = context.getObjectByPickColor(pickColor);
       if (defined(object)) {
-        return object;
+        objects.add(object);
+        if (0 >= --limit) {
+          break;
```
mzschwartz5: nit - this is a slightly weird way to phrase this. I think it would be slightly clearer to do either:

`if (limit-- <= 0) break;`
or
```JS
limit--;
if (limit <= 0) break;
```

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -123,8 +122,7 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle) {
     x += dx;
     y += dy;
   }
-
-  return undefined;
+  return [...objects];
```
mzschwartz5, line 123: The spread operator (`...`) is not super performant. I don't have an idea of how large `objects` may be, but `return Array.from(objects);` would be the better option if it can be large.

packages/engine/Source/Scene/PickFramebuffer.js
mzschwartz5, line 1: Question: your PR description says "PickFramebuffer.end could continue counting entities, rather than exiting early the moment one was found". I _thought_ what you meant was that exiting early would speed up picking, but it sounds like you're saying the opposite, actually.

Can you elaborate on why _not_ exiting early is faster? (And it may help to update the description as well)

packages/engine/Source/Scene/Picking.js
```diff
@@ -1016,7 +1036,7 @@ function getRayIntersection(
   scene.resolveFramebuffers(passState);
 
   let position;
-  const object = view.pickFramebuffer.end(drawingBufferRectangle);
+  const object = view.pickFramebuffer.end(drawingBufferRectangle, 1)[0];
```
mzschwartz5, line 1063: Since 1 is the default limit, you could omit it. If you choose to do that, you might want a comment explaining why we need to use `[0]`.

packages/engine/Source/Scene/Scene.js
```diff
@@ -4402,7 +4402,7 @@ Scene.prototype.clampLineWidth = function (width) {
  * @returns {Object | undefined} Object containing the picked primitive or <code>undefined</code> if nothing is at the location.
  */
 Scene.prototype.pick = function (windowPosition, width, height) {
-  return this._picking.pick(this, windowPosition, width, height);
+  return this._picking.pick(this, windowPosition, width, height, 1)[0];
```
mzschwartz5, line 4406: Same comment here about the default limit value and `[0]`

packages/engine/Source/Scene/Picking.js
```diff
@@ -262,23 +262,31 @@ function computePickingDrawingBufferRectangle(
...
  * <p>
  * When a feature of a 3D Tiles tileset is picked, <code>pick</code> returns a {@link Cesium3DTileFeature} object.
  * </p>
  * @param {Scene} scene
  * @param {Cartesian2} windowPosition Window coordinates to perform picking on.
  * @param {number} [width=3] Width of the pick rectangle.
  * @param {number} [height=3] Height of the pick rectangle.
- * @returns {object | undefined} Object containing the picked primitive.
+ * @param {number} [limit=1] If supplied, stop iterating after collecting this many objects.
+ * @returns {object[]} Objects containing the picked primitives.
```
mzschwartz5: I want to be careful that there's no code calling `pick` that then checks `is undefined` after. With this change, that code would now need to check `is length > 0`. How did you search for any such occurrences?

packages/engine/Source/Scene/Picking.js
```diff
@@ -262,23 +262,31 @@ function computePickingDrawingBufferRectangle(
...
  */
-Picking.prototype.pick = function (scene, windowPosition, width, height) {
+Picking.prototype.pick = function (
+  scene,
+  windowPosition,
+  width,
+  height,
+  limit,
```
mzschwartz5: Same comment here as before - default value for `limit` can be defined in the signature.

packages/engine/Source/Scene/Picking.js
```diff
@@ -328,7 +336,7 @@ Picking.prototype.pick = function (scene, windowPosition, width, height) {
   scene.updateAndExecuteCommands(passState, scratchColorZero);
   scene.resolveFramebuffers(passState);
 
-  const object = pickFramebuffer.end(drawingBufferRectangle);
+  const object = pickFramebuffer.end(drawingBufferRectangle, limit);
```
mzschwartz5: `object` should be renamed to:
1. Reflect that it's now an array, not a single object
2. Be more descriptive than just "object". 

How about `pickedObjects`?

packages/engine/Source/Scene/Picking.js
```diff
@@ -725,59 +733,64 @@ function drillPick(limit, pickCallback) {
...
-      break;
-    }
-
-    if (!defined(object) || !defined(object.primitive)) {
-      break;
-    }
+  let pickedResults = pickCallback();
+  let shouldBreak = false;
+  while (defined(pickedResults) && pickedResults.length > 0 && !shouldBreak) {
+    for (const pickedResult of pickedResults) {
```
mzschwartz5: Surely we don't need a for loop inside a while loop, right? I haven't fully processed this code, but what is the while loop doing, exactly? Can we just... remove it, and only use the for loop?

mzschwartz5: Side note- if we _do_ really need this while loop, I think I'd rather see the inner for loop as its own function (to be frank, I'd _prefer_ that the existing code had been written as _several_ smaller functions). 

Then, you would no longer need `shouldBreak` - you could just say `if !myForLoopFunction(...) break`.

packages/engine/Source/Scene/Picking.js
```diff
@@ -808,15 +821,22 @@ Picking.prototype.drillPick = function (
   height,
 ) {
   const that = this;
+  if (!defined(limit)) {
+    limit = Number.MAX_VALUE;
+  }
```
mzschwartz5: nit - you can consolidate this into the call to `that.pick`

```
that.pick(
...
limit ?? Number.MAX_VALUE
);
```

mzschwartz5 (changes requested):
Hey @Beilinson

Thanks for the contribution! I left a number of comments, but they're mostly minor. The performance improvement seems huge. I just have a few questions, but they're pretty important:

1. Paraphrasing a comment I left on `PickFramebuffer.js`, can you elaborate on why _not_ returning early from `PickFramebuffer.end` speeds _up_ picking? Seems counterintuitive to me.
2. Following from that, does this performance gain come at the cost of some other use case of picking? (like, does it speed up drill picking at the cost of slowing down regular picking?)
3. In the example sandcastle you linked under the testing section, it only picks 40 pins (compared to the 220 in your measurements). What do I change to increase it to 220?

packages/engine/Source/Scene/PickFramebuffer.js
Beilinson: right, i copied this code over from the `drillPick` method in `Picking.js`, want me to change it there as well?

Beilinson:
Hey @mzschwartz5 , thanks for the review!

> I was particularly surprised the `Map` change gave a big difference, since the plain object `{}` is also backed by a hash table. (In contrast to #12896, where we were iterating over an array. Map had a huge impact there).

`delete obj[key]` is very slow on objects compared to `map.delete(key)` (about 75% difference apparently) [source](https://www.builder.io/blog/maps). Because the picking is a depth peeling algorithm, the entire `_pickObjects` map is reconstructed every iteration, which means deleting and reinserting all keys each time, which is where much of the slowdown came from.



>     1. Paraphrasing a comment I left on `PickFramebuffer.js`, can you elaborate on why _not_ returning early from `PickFramebuffer.end` speeds _up_ picking? Seems counterintuitive to me.

The drill picking implemented by Cesium is similar to [Depth Peeling](https://en.wikipedia.org/wiki/Depth_peeling). Each iteration, entities are rendered to some offscreen buffer, that buffer is then read into memory and then we iterate over the pixels (spirally from the center, this is primarily used for picking the entities "closest" to the pointer, not really useful when picking an actual rectangle). The old algorithm would return the moment one primitive was found.

The sandcastle demonstrates a scenario where picking isn't directly of a 3x3 pixel rectangle at the pointer, rather a significant portion of the screen. This is also a use case we have in our enterprise system, where we support "drag selecting" multiple entities.

The important thing is depth peeling is critical when entities are rendered 100% directly on top of another and you are picking a very small (3x3) area. In this use case however, the entities dont have a significant amount of overlap, meaning each iteration we could find more entities rendered, but instead we exit early to re-render everything and re-read everything. 

By letting the `PickFramebuffer` continue trying to search for entities, we solve this so over a large rectangle at the given zoom in the sandcastle all entities are found in one pass, and only one re-render is needed (to make sure that there are no entities under those ones). Thats the significant performance boost.

>     2. Following from that, does this performance gain come at the cost of some other use case of picking? (like, does it speed up drill picking at the cost of slowing down regular picking?)

No, regular picking still picks a (3x3 by default) rectangle, and stops after finding the first entity (`limit=1`).  Also I didn't remove the spiral code (https://github.com/CesiumGS/cesium/blob/ce6c3e28fc7e550b32a5c67335aa4490d0e84c34/packages/engine/Source/Scene/PickFramebuffer.js#L82C3-L86C83), so that also still works the same.

>     3. In the example sandcastle you linked under the testing section, it only picks 40 pins (compared to the 220 in your measurements). What do I change to increase it to 220?

Probably has to do with screen size, you could try resizing the viewer size or alter the sandcastle so the pickRectangle is the entire canvas, but that would be even more unfair

packages/engine/Source/Scene/Picking.js
Beilinson: TBH i would've liked to use a label on the outer while loop rather than the `shouldBreak` flag, but I see your point. I''m not sure refactoring it out into a separate function would be great, this is the bread and butter of this method and it would need to get multiple parameters to refactor. thoughts?

Beilinson: Regarding the double loops:

1. `while` loops as long as there are entities (and hasn't reached the limit) and redoes the `pickCallback` 
2. `for` loop to actually iterate over all the entities and add/break once hit the limit.

packages/engine/Source/Scene/PickFramebuffer.js
Beilinson: They have the same performance (its a set so they do the same thing at the engine level), tested on all major browsers (the image is on chrome)

basic-example

I personally prefer the spread operator for converting sets to arrays, but its a style choice so it doesn't really matter.

packages/engine/Source/Scene/PickFramebuffer.js
Beilinson: Answered in the full comment below

packages/engine/Source/Scene/Picking.js
Beilinson: My thinking was I prefer to be overly **explicit** about what we are doing here, at any point `PickFramebuffer` might be refactored to a different default limit.

I'll comment it though and can remove if you think otherwise

packages/engine/Source/Scene/Picking.js
Beilinson: Theres only 2 methods that use `pick`, thats the regular `Scene.prototype.pick` in `Scene.js` and `Picking.prototype.drillPick` in `Picking.js`. The first one I always take the first object (which is either defined or undefined and is consistent with intended return type), and the second one was the one I fixed (the while loop in `drillPick` has the check for `objects.length`)

packages/engine/Source/Scene/Picking.js
Beilinson: Reason I left it out was because the rest of the code doesn't use this practice almost at all (even width and height above it get defaults internally using the same `width = width ?? 3` code). I agree that signature defaults are better (and more readable)

packages/engine/Source/Scene/Picking.js
Beilinson: This could also be refactored to use arrow functions, that could get rid of the `that = this` funny business (https://github.com/CesiumGS/cesium/issues/10486#issuecomment-1188218009 mentions this)

packages/engine/Source/Scene/PickFramebuffer.js
javagl: The spread operator puts all operands on the stack. Doing a spread with an array of 200000 elements will cause a stack overflow error. The exact limit depends on many factors, and in any case, it's unlikely that this will be hit here, but something to keep in mind. (We stumbled over this in https://github.com/CesiumGS/cesium/issues/12053 )

Beilinson:
Essentially this PR has two separate (not necessarily exclusive) improvements:

1. The ~75% speedup in drillPicking a small area (3x3) with many overlapping entities (due to the `Map` switch)
2. A more significant speedup (at best, N=# of entities, then from **O(N)** to **O(1)**), when picking a large area with many spread out non overlapping entities

To be precise, 2 speeds up from **O(N)** to **O(N2),** where N2 is the maximum # of fully overlapping entities at the given zoom and render resolution.

packages/engine/Source/Scene/PickFramebuffer.js
Beilinson: https://v8.dev/blog/spread-elements

its the same code under the hood (in v8/spidermonkey for sure, safari haven't checked yet). Sets are specifically handled.

From a style perspective, the codebase has only 1 instance of `Array.from`, and 8 for the spread operator. in fact, theres 2 instances of using `[...new Set(<>)]` in two of the imagery classes.

packages/engine/Source/Scene/PickFramebuffer.js
mzschwartz5: Sure, why not!

packages/engine/Source/Scene/Picking.js
mzschwartz5: Yeah I don't mind deviating from precedent here. We should be using signature defaults where possible. (Maybe, back when those other instances were written, signature defaults weren't widely supported in JS? Just a guess, not that it matters)

packages/engine/Source/Scene/Picking.js
mzschwartz5: I don't feel too strongly about this. On the one hand, being explicit is good. On the other hand, why have a default value if you think it's a bad idea to use it? (I guess maybe if you're passing in a variable as the `limit` value and the variable could be undefined?)

I think I've convinced myself that being explicit here is fine.

mzschwartz5:
@Beilinson 
> No, regular picking still picks a (3x3 by default) rectangle, and stops after finding the first entity (limit=1).

Got it. So the behavior of regular picking is exactly the same as before this PR, because it uses limit=1. Perfect.

I guess the natural follow up is then, does this slow down the performance of regular 3x3 drill picking, where we most likely *do* want 1 result at a time? (I need to go over the PR again, and maybe I'll figure out the answer myself, but good to document it anyway).

___

Took another pass, still interested in your response to the above question, but otherwise I think this is looking pretty good barring those few style-related discussions.

packages/engine/Source/Scene/Picking.js
mzschwartz5: If you were to refactor the inner for loop into its own function, the args would be`pickedResults`, `results`, and `pickedPrimitives` - did I miss anything? That doesn't sound too bad to me. 

Unrelated, is `defined(pickedResults)` a necessary while loop condition? Isn't `pickedResults` guaranteed to be defined, it just may be 0-length?

Beilinson: `getRayIntersection` in `Picking.js` returns `object[] | undefined`, which is used as the drillPick callback in `getRayIntersections`. Do you prefer me to change it to return `Frozen.EMPTY_ARRAY`?

Beilinson:
@mzschwartz5 
> I guess the natural follow up is then, does this slow down the performance of regular 3x3 drill picking, where we most likely _do_ want 1 result at a time? (I need to go over the PR again, and maybe I'll figure out the answer myself, but good to document it anyway).

It's the exact same code flow as previously because it returns early after finding one result `(limit <=0) break` in `PickFramebuffer.prototype.end`, so should be the same performance as before

packages/engine/Source/Scene/Picking.js
Beilinson: I refactored the for loop to an external function, it does take quite a few arguments but overall looks cleaner to me

packages/engine/Source/Scene/PickFramebuffer.js
javagl: The two places of `[...new Set]` have been written by me, to make an array of projections unique. I'm pretty sure that many things will break when there is ever more than _one_ projection involved in all that, but ... let's not talk about that. It is what it is.

More generally, the question is less about _performance_, but about _whether it will crash_. For example, this will not work
```
function test() {
  const sourceArray = [];
  for (let i=0; i<150000; i++) {
    sourceArray.push(i);
  }
  const targetArray = [];
  targetArray.push(...sourceArray);
  console.log("Done", targetArray);
}

try {
  test();
} catch (error) {
  console.log(error);
}
```

Interestingly with
`const targetArray = [...sourceArray];`
it **will** work. Apparently, things are treated differently when it is about the _construction_ of an array. It's all a huge black box. Who knows.

Beilinson: Not attached to this, can update if there's a preferred approach

javagl: It looks like `return [...set]` works even for large sets, so no reason to change it _here_, IMHO. The general advice to be careful with the `...spread` operator when the spreaded collection may be "large" still stands. I think that there are neither rules in the coding guide, nor rules on the level of any specification that could _reliably_ be applied here.

mzschwartz5:
@Beilinson 
> It's the exact same code flow as previously because it returns early after finding one result `(limit <=0) break` in `PickFramebuffer.prototype.end`, so should be the same performance as before

No, I don't think this is true. Take this example: I'm doing a 10x10 drill pick, on a region of the screen where 5 entities are overlapped, and I've set the limit to 5. On each pick pass, now, even once we've found 1 entity, `PickFramebuffer.prototype.end` will continue to search through the whole 10x10 region because it hasn't hit the limit on this pass - even though there's nothing left to find on this layer. Does that sound right to you? (I'm no expert on picking but that's my understanding of how this works)

Essentially, this PR is a tradeoff - better performance for large pick rectangles when entities are non-overlapping, at the same depth, but worse performance when they are overlapping at various depths. Is the tradeoff worth it? We might need to discuss. The primary use-case of drill picking is to pick objects that _are_ overlapping each other at various depth layers.

So *iff* this has a significant performance impact on large-rectangle drill picks containing overlapping objects, we may want to consider other options. (like maybe a new picking API specifically for this type of picking?).

mzschwartz5:
Also, won't this change _which_ entities get picked? The old method prioritizes entities closest to the mouse position, depth-first if you will. The new method prioritizes breadth-first.

This is probably more important of a question than the performance one above, as I assume that, overall, performance is probably better all-around given the other changes (map, rgb conversion)

Beilinson:
Ah sorry I misread your question. Yes you're right, this PR causes breadth-first instead of depth-first search. 

Regarding the fact that you iterate over empty pixels after already finding an entity rather than early return, I took the liberty to benchmark the iteration speed before and after my map+byte conversion improvements:

[New sandbox](http://localhost:8080/Apps/Sandcastle/index.html#c=hVTbbuM2EP2VgZEHClGp2M1DkzhBu46LDZq0wcZoXwysaXLsMEuTWpJy4hr6944uvqXeXUCAJM45hzNnhpTOhghLja/o4RosvsIAgy4W/O96jY07sv4fOBuFtujHnRTWYwsQJFp8cAovN5SnzQp/Ggz/HPZu0woX9QINMS9hJkzAek1YvRBRO7u/6PFrgSF+QqvQN8LRF1WsTK7Gdmxlney6SegfreYY0yaNknJvitgD5tPDih61/VBoQ+Ks0ps5D8xgBE2wsyt69aFXvU+vu0lT4hby0kBeWsjLDgLtvhxt1FFj4EIp1oYAchd0U2ebxED4SF/C/sxn3i1uce4RA+um0E3SDWuqjZk64dUlbJUAyLI5eZJPa+YI3yKbnKx1mZ6sX8pJut3BGef5p+FtCue/7DTL9qv2kl7k6sYosURP0lQjE96LVQLXN1B/cY+qkMiYkDIFWXjfxKSE0/o3hbMEshZs0M7jc9WA1pOZWY0ce2cQjQ+owrft78KmuXtd5UGSK/YpFxKHSyJ+FFYZ0ggY72xexN9kxWZsvfG3rPOqzapqcga5cXM2edTyi7ZzmGJ8JUU42TL4G4/u0aPUoZI6Tyoft8HV+yC86vgMAi7eLmhQZRR2bnDSeJlljY8LrSjLw6Hb9rvHdjtnvRR2W2W9RqYRCZEYrcatiMgSTo6M6BCxfZjHClQPP1ee5qWqdLtFCgUdohmdOpXCBT076oE1qMgRkmpbV0KubQBd+fQg4jM10DnPjqUCPzWZJuUi8MkR+QcUlrKsznJ9BVySZjtoLLqcN6FKLFQa35EQ6lG/ofmGShv9sZCORItYo8NRsRZxKFWmu+vtcChHqxz5/fD30efB/d3gj+Sqk3b6Ia4M3mwO3a96kTtqZ+EN4zyLuMgN7RCyaUHmRy5DaNIF6Gf71L7SS9Dq+sj1C9KIECgyK4x50v/iuHPTzwj/P6pxQtHw/0UlGrGqYM/dm/tmkXPez+j3ODM6Z6bCv1P+Dw) In the new example, I click directly on the entities, which are all 100% overlapping

Main:
Picking between 469.0,210.0 with a 11x11 rectangle
Picked 400 pins in 471ms.

Color + Map improvements:
Picking between 451.2,186.6 with a 11x11 rectangle
Picked 400 pins in 358ms.

Color+Map+Iterate over all pixels:
Picking between 452.0,203.4 with a 11x11 rectangle
Picked 400 pins in 367ms.

I benchmarked all these several times, and the numbers weren't 100% consistent obviously but this was the average first click time.

As a user of both scenarios, I feel a 10ms slowdown (which isn't felt because this pr also gives 100ms speedup) for a pick of 400 entities, while getting a **O(N)** improvement over non-overlapping entities is worth it. Iterating over 100 pixels really isn't all that slow when all the iteration does is check a map

Beilinson:
> Also, won't this change _which_ entities get picked? The old method prioritizes entities closest to the mouse position, depth-first if you will. The new method prioritizes breadth-first.
> 
> This is probably more important of a question than the performance one above, as I assume that, overall, performance is probably better all-around given the other changes (map, rgb conversion)

Agreed this is 100% a more important question. On any sized rectangle, given no limit, the result will be the same (different order, but for large picks thats an improvement). On small picks (3x3) it is most likely guranteed to be the same as before, because I still iterate centrally outwards, meaning **if** any outer entities pixels exist they are are probably of entities hidden by the central entity.

Given a large rectangle, and a small limit, and a mix of overlapping and non-overlapping entities, then the results **will be different** - drillPick will now prioritize **visible** entities over completely hidden ones.

From my point of view this is a matter of intention: as a user, we use drillPick with the default 3x3 to pick things directly under the mouse. The use of drillPick with a large rectangle is to drag select, in which case if we have a limit (which we do) we prefer to highlight the visible entities over the hidden ones.

If there is a use-case for drill picking a large rectangle and still preferring a depth-first search, I am open to adding a flag which will early return after 1 entity is found. I do think breadth-first should be the default, because of its improved performance. If needed, I could add a warning in the changelog

mzschwartz5:
That's a pretty fair assessment on both fronts. I agree about the performance changes being a non-issue.

I'm going to seek input from my teammates on the selection ordering; I think it could be OK, and may just warrant a note in the change log as you said.

Beilinson:
Found this https://github.com/CesiumGS/cesium/issues/3018 randomly looking through issues related to performance/memory leaks, realized that this PR implements the transition to Map as mentioned there, so this could close that I think

mzschwartz5:
Hey @Beilinson,

Team didn't have strong opinions - I think let's just add a small `Breaking Change` note to the `CHANGES.md` about the picking order change.

Otherwise this is good to go. I'm out tomorrow and Wednesday, but will merge on Thursday (with the `CHANGES.md` note). Thanks for the work!

Beilinson:
Great @mzschwartz5!

I've updated `CHANGES.md`, will be available thursday if needed.

CHANGES.md
```diff
@@ -4,9 +4,16 @@
...
+#### Breaking Changes :mega:
+
+- `scene.drillPick` now uses a breadth-first search strategy instead of depth-first. This may change which entities are picked when
+  using large values of `width` and `height` when providing a `limit`, prioritizing visible entities over obscured ones.
+
 #### Fixes :wrench:
 
 - Fixes an event bug following recent changes, where adding a new listener during an event callback caused an infinite loop. [#12955](https://github.com/CesiumGS/cesium/pull/12955)
+- Improved performance of `scene.drillPick`. [#12916](https://github.com/CesiumGS/cesium/pull/12916)
+- Improved performance when removing primitives. [#3018](https://github.com/CesiumGS/cesium/pull/3018)
```
mzschwartz5, line 17: Not sure how this got here - let's take that back out?

CHANGES.md
```diff
@@ -4,9 +4,16 @@
 
 ### @cesium/engine
 
+#### Breaking Changes :mega:
+
+- `scene.drillPick` now uses a breadth-first search strategy instead of depth-first. This may change which entities are picked when
+  using large values of `width` and `height` when providing a `limit`, prioritizing visible entities over obscured ones.
```
mzschwartz5: Since another commit will be needed for my below comment anyway, can I suggest changing the wording here to:

"...and providing a `limit`, prioritizing entities closer to the camera."

CHANGES.md
mzschwartz5: Looking at a diff of your changes since my last review. it actually looks like a lot has changed in `CHANGES.md`... maybe as the result of a merge? Let's fix that and then I can merge your PR into main.

Beilinson: Hey @mzschwartz5, I added this intentionally, this PR handles the primary issue mentioned in #3018 which was: 

> removeAll iterates through each primitive and removes it, which removes the pick IDs from the cache, one at a time.
>Deleting the pick ID changes the shape of the _pickObjects cache object, which I would expect to be a slow operation, but I expected this object to already be in "dictionary" mode, and expected dictionary-like performance.
>The new ES6 Map object may work better, or it might still be unoptimized and end up being worse. We'd have to try it and see.

So I unintentionally also have done the recommended suggestion at the end of that issue which was to change '_pickObjects' to a Map, which means that this PR also improves performance of individually removing primitives. 

I mentioned that in a comment above in the PR

mzschwartz5: Ah okay, my bad I missed that!

mzschwartz5:
@Beilinson Looks like CI is failing with some JS doc-related errors... :/

```
 ERROR: Unable to parse a tag's type expression for source file /home/runner/work/cesium/cesium/packages/engine/Source/Scene/Picking.js in line 791 with tag title "param" and text "{(limit: number) => object[]} pickCallback Pick callback to execute each iteration": Invalid type expression "(limit: number) => object[]": Expected "$", "'", ".", "0", "@", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "try", "typeof", "var", "void", "while", "with", "yield", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter, or [1-9] but " " found.
ERROR: Unable to parse a tag's type expression for source file /home/runner/work/cesium/cesium/packages/engine/Source/Scene/Picking.js in line 797 with tag title "param" and text "{(limit: number) => object[]} pickCallback Pick callback to execute each iteration": Invalid type expression "(limit: number) => object[]": Expected "$", "'", ".", "0", "@", "\"", "\\", "_", "break", "case", "catch", "class", "const", "continue", "debugger", "default", "delete", "do", "else", "enum", "export", "extends", "finally", "for", "function", "if", "implements", "import", "in", "instanceof", "interface", "let", "new", "package", "private", "protected", "public", "return", "static", "super", "switch", "this", "throw", "try", "typeof", "var", "void", "while", "with", "yield", Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode titlecase letter, Unicode uppercase letter, or [1-9] but " " found.
[16:34:25] 'buildTs' errored after 45 s
[16:34:25] Error: Command failed: npx jsdoc --configure packages/engine/tsd-conf.json
```

Can you look into this? You should be able to run the `npm run build-docs` command locally to verify if you've solved the issue, I believe.

Beilinson:
Fixing the first tsdoc error now (optional argument before required)

The second issue seems to be due to the TSDOC target library not being es2015, which I assume is a mistake since the gulpfile build step targets es2020.

@mzschwartz5  Correct me if I'm wrong. I could just fix this by removing the jsdoc error complaining about the use of `Map`, but I don't think this is a good idea as it will be confusing to future developers using this (even though the error is on a completely internal part of cesium)

Beilinson:
This seems to be failing to a flaky spec, I tested just now and it passes for me:

image

image

mzschwartz5:
Yeah that one fails pretty frequently... I'm just going to rerun the that test.

mzschwartz5:
All good to go, thanks for the contribution! This is a huge performance upgrade for drill picking.

Beilinson:
Awesome!

Issue #12857, opened 2025-08-29
Help us improve Sandcastle - Feedback thread


jjspace:
<img width="3837" height="2146" alt="Image" src="https://github.com/user-attachments/assets/4d8b2a3c-b003-46d6-aa43-c2fd4b13d7cd" />

We've recently launched our rebuild/redesign of Sandcastle for CesiumJS. We'd love to hear feedback, especially if anything's not working as expected. This issue will serve as the primary point for said feedback and bug reports.

If you're interested in the build process feel free to check out [this issue](https://github.com/CesiumGS/cesium/issues/12566).
If you're interested in some of the previous feedback and discussion around features check out this [forum thread](https://community.cesium.com/t/upgrading-sandcastle-we-need-your-input/39715/28) or [this one](https://community.cesium.com/t/redesigning-sandcastle-requesting-feedback/41981) for some design discussion.

mzschwartz5:
The new nav help menu buttons look a little awkward to me. Maybe I'm just used to the old sandcastle, where the Mouse and Touch buttons were row-aligned, but the empty space here looks weird to me.

<img width="283" height="418" alt="Image" src="https://github.com/user-attachments/assets/2662ed90-7582-43df-9845-d155be1a42b1" />

javagl:
Some CSS issue, probably, in https://dev-sandcastle.cesium.com/?id=materials :  

<img width="632" height="392" alt="Image" src="https://github.com/user-attachments/assets/f9625a33-e1ab-4d96-a421-43f8ab6d11e6" />

lukemckinstry:
Did we switch the main search from strict to fuzzy match?
I used to find it very useful to match exactly when searching for class or parameter. I see the benefit of fuzzy search but could there be an option for the old strict matching?

<img width="960" height="600" alt="Image" src="https://github.com/user-attachments/assets/b4b3df8b-bbc8-4237-9d3a-103ccfb25547" />

matjuric:
Click on the iModel Mesh Export Service (https://dev-sandcastle.cesium.com/?id=imodel-mesh-export-service) produces this issue for me.
My setup:
**Brave 1.82.161 (Official Build) (64-bit)](https://brave.com/latest/)
Chromium: 140.0.7339.80**

<img width="3817" height="1878" alt="Image" src="https://github.com/user-attachments/assets/e239cc87-99e3-4fe8-9038-6176bbb63e1a" />


**message:**
[...]

PR #12917, opened 2025-09-24
Attempt to fix highlighting issue in monaco

jjspace:
# Description

This is an attempt to fix the highlighting issue shown below from the feedback thread [comment](https://github.com/CesiumGS/cesium/issues/12857#issuecomment-3256443143)

Image

I can't reproduce this myself as it seems specific to Windows + Chrome. ~~Some investigation has pointed to potentially the ligature setting affecting this so this PR is to test that.~~ More investigation showed this was a problem with loading fonts after Monaco tried to measure them.

## Issue number and link

part of https://github.com/CesiumGS/cesium/issues/12857

## Testing plan

- Run `npm run dev` from the sandcastle package or `npm run build-sandcastle` and `npm start` from the project root
- Open the editor and do some text selection
- Click single tokens so the highlight shows up for them
- Confirm all highlights and cursors appear where you'd expect
- Test with multiple font and ligature enabled options in the settings

These steps reproduced the issue for me even on Linux:

- Set settings to a larger font size to make it more obvious
- Set font to Droid Sans, Ligatures OFF
- REFRESH
- Click Cesium in the editor to show the highlight in the background
- Switch to Cascadia Code font, observe offset (however small it is)

Same works for me from 
Cascadia -> Fira
Droid -> Cascadia
Jetbrains -> Cascadia

Toggling Ligatures on the selected font also usually fixed it _for me_

Example on Linux:

# Author checklist

### DIFF
--- a/packages/sandcastle/src/SandcastleEditor.tsx
+++ b/packages/sandcastle/src/SandcastleEditor.tsx
@@ -54,6 +54,15 @@ loader.config({ monaco });
 const TYPES_URL = `${__PAGE_BASE_URL__}Source/Cesium.d.ts`;
 const SANDCASTLE_TYPES_URL = `templates/Sandcastle.d.ts`;
 
+function fontLigaturesValue(fontLigaturesBool: boolean): boolean | string {
+  // Due to what seems like a bug in Monaco on some systems like Windows + Chrome
+  // the text highlighting and selection is offset from the actual values.
+  // This issue thread shows that setting the ligatures to an empty string resolves
+  // the issue. https://github.com/microsoft/monaco-editor/issues/3217#issuecomment-1511978166
+  // Testing has shown that `true` renders fine however.
+  return fontLigaturesBool ? fontLigaturesBool : "";
+}
+
 export type SandcastleEditorRef = {
   formatCode(): void;
 };
@@ -90,7 +99,7 @@ function SandcastleEditor({
   }, [fontFamily]);
   useEffect(() => {
     internalEditorRef.current?.updateOptions({
-      fontLigatures: fontLigatures,
+      fontLigatures: fontLigaturesValue(fontLigatures),
     });
   }, [fontLigatures]);
   useEffect(() => {
@@ -325,7 +334,7 @@ Sandcastle.addToolbarMenu(${variableName});`);
             fontFamily:
               availableFonts[fontFamily]?.cssValue ?? "Droid Sans Mono",
             fontSize: fontSize,
-            fontLigatures: fontLigatures,
+            fontLigatures: fontLigaturesValue(fontLigatures),
           }}
           path={activeTab === "js" ? "script.js" : "index.html"}
           language={activeTab === "js" ? "javascript" : "html"}
--- a/packages/sandcastle/src/SandcastleEditor.tsx
+++ b/packages/sandcastle/src/SandcastleEditor.tsx
@@ -196,6 +196,9 @@ function SandcastleEditor({
 
     setupSandcastleSnippets(monaco);
     setTypes(monaco);
+    // TODO: remove this!
+    // @ts-expect-error just testing
+    window.monaco = monaco;
   }
 
   async function setTypes(monaco: Monaco) {
--- a/packages/sandcastle/src/SandcastleEditor.tsx
+++ b/packages/sandcastle/src/SandcastleEditor.tsx
@@ -205,9 +205,6 @@ function SandcastleEditor({
 
     setupSandcastleSnippets(monaco);
     setTypes(monaco);
-    // TODO: remove this!
-    // @ts-expect-error just testing
-    window.monaco = monaco;
   }
 
   async function setTypes(monaco: Monaco) {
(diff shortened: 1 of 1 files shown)


javagl:
Disclaimer: It's surprisingly hard to systematically test this. The behavior seems to change randomly. Building and reloading (with cache disabled) sometimes seems to pick up modified code, but sometimes not, and sometimes changing the 'Settings' seems to have an effect, and sometimes not, and sometimes only after reloading the page (not once but) _multiple_ times. The settings appear to be stored in cookies (?). Maybe it's necessary to systematically delete them at certain points in the test procedure...?

However, I tried it out, and it didn't resolve it for me. 

My **strong** recommendation for resolving this issue is:

Remove that "Droid Sans Mono" option, remove the "Font ligatures" toggle, and call it a day. 

(If somebody cares, the response should be "Pull requests welcome!". How many substantially different forms of a "monospaced font" are there, after all ... ?)

--- 

Rant: 

I'm recommending that after zooming a bit into that (with another disclaimer: I'm not a CSS/React expert, and I proudly refuse to become one). Apparently, `fontLigatures` can be a `boolean` or a `string`. And I tried to find out what the apparent `string` values are for `fontLigatures` are, but didn't find anything. Zero. Nothing. I randomly tried inserting the `""` empty string or the string `"false"` and `"off"` at different places, but it didn't have any effect. Why did I try `"off"`? Well, the CSS seems to contain
`--vscode-editorCodeLens-fontFeatureSettings: "liga" off` 
or 
`font-feature-settings: "liga" 0, "calt" 0;`
at different places, so... why not?

My sincere condolences to everybody who has to deal with ... things ... like **this**:

NotCesium Font

---

Back on topic: It looks like the [`Settings` defines `fontLigatures` to be `boolean`](https://github.com/CesiumGS/cesium/blob/b04b59282d2926bb48fc8abb30cc54db38857315/packages/sandcastle/src/SettingsContext.ts#L42), which might have to be adjusted to inlcude `string` for this change anyhow.

jjspace:
I did some more digging into this this morning. I'm more confident now saying this is an async issue with the font's loading after Monaco tries to measure them as mentioned in this issue https://github.com/microsoft/monaco-editor/issues/392. This was linked from a few other alignment related issues too.

I've updated the code to revert the original ligature change but also delay setting the monaco font until it's loaded. The [font loading API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API) is not perfect and there might still be a better way of doing this. However, I have confirmed the issue is gone on my windows machine using the CI branch. @javagl can you please re-test this?

Other solutions might be to try and pre-load the font files but I wanted to avoid loading unnecessary fonts that people aren't using.

javagl:
I tried it out, and despite all evil efforts of switching the fonts and toggling the ligatures, I could no longer reproduce the issue.

So this _seems_ to fix it. 

(Given that this apparently was some "synchronization/race-condition" thingy, it's hard to say for sure, but it seems to be fixed).

---

I looked at the code change, and cannot say anything about that. But in the future, when someone asks "Why do IT projects often take so much longer and cost so much more than originally planned?", I can say: "**This** 👉". 
(Back in the days, I used to set a font for a component with `component.setFont(new Font("Monospaced", PLAIN, 10));` 👴 )

packages/sandcastle/src/SandcastleEditor.tsx
```diff
@@ -83,10 +83,28 @@ function SandcastleEditor({
...
+  const documentRef = useRef(document);
   useEffect(() => {
-    internalEditorRef.current?.updateOptions({
-      fontFamily: availableFonts[fontFamily]?.cssValue ?? "Droid Sans Mono",
-    });
+    const cssName = availableFonts[fontFamily]?.cssValue ?? "Droid Sans Mono";
+    const fontFace = [...documentRef.current.fonts.values()].find(
+      (font) => font.family === cssName && font.weight === "400",
+    );
+    if (fontFace?.status !== "loaded") {
```
ggetz, line 92: For simplicity, could we instead just wait on `document.fonts.ready` and then call `monaco.editor.remeasureFonts()`?

ggetz: Per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API): 

> Note that the font.load() returns a promise, so we could have handled the completion of font loading by chaining then afterwards. Using [document.fonts.ready](https://developer.mozilla.org/en-US/docs/Web/API/FontFaceSet/ready) can be better in some circumstances, as it is only called when all fonts in the document have been resolved and layout is complete.

jjspace: @ggetz Unfortunately not. That promise is a one time thing and it always returned instantly when I tried. On top of that, that promise and the other `onfontloaded` event handlers only trigger when the font is loaded using the fonts API and do _not_ seem to trigger when a font is loaded because of styling changes (like how monaco updates it). I tried other arrangements of this and none worked so I settled on this as the best option.

ggetz: Ah, bummer.

jjspace:
Auto merge won't go until it has an approve. CI passed on the last commit so I'm going to just merge

PR #12920, opened 2025-09-25
Add declusteredEvent to EntityCluster

### LINKED ISSUE #5760: Include declustered entities in cluster callback
Requested on the Forum here: https://groups.google.com/forum/#!topic/cesium-dev/hUrOxJQjjzg

They would like a way to access the entities that have been unclustered in the cluster callback function when the event is fired.

Oko-Tester:
Fixes #5760 by adding declusteredEvent that provides both clustered and declustered entities, plus new API methods for accessing clustering state. Maintains backward compatibility.

# Description

This PR adds comprehensive clustering state information to EntityCluster by introducing a new `declusteredEvent` and supporting API methods. The original `clusterEvent` only provided information about entities that were successfully clustered, but developers also needed access to entities that were processed but not clustered (declustered entities).

## Key Changes

- **New `declusteredEvent`**: Fires with complete clustering information including both clustered and declustered entities
- **New API methods**: 
  - `getLastClusteredEntities()` - Returns currently clustered entities
  - `getLastDeclusteredEntities()` - Returns currently declustered entities  
  - `getAllProcessedEntities()` - Returns all entities processed during clustering
- **Enhanced event coverage**: Events now fire when clustering is disabled or clusters become empty
- **Backward compatibility**: Original `clusterEvent` remains unchanged and functional

## Issue number and link

Fixes #5760 - Include declustered entities in cluster callback

This addresses multiple user requests from the forum and GitHub comments spanning several years.

### Manual Testing
1. Create CustomDataSource with clustering enabled
2. Add entities at various distances to create both clustered and declustered entities
3. Verify `declusteredEvent` fires with correct `clustered` and `declustered` arrays
4. Disable clustering and verify event fires with all entities as declustered
5. Test API methods return correct entity lists
6. Verify original `clusterEvent` still works as before

# Author checklist

### DIFF
--- a/packages/engine/Source/DataSources/EntityCluster.js
+++ b/packages/engine/Source/DataSources/EntityCluster.js
@@ -67,7 +67,8 @@ function EntityCluster(options) {
 
   this._clusterEvent = new Event();
   this._allDeclusteredEvent = new Event();
-  this._previouslyClusteredEntities = [];
+  this._clusterMembershipChangedEvent = new Event();
+  this._previousClusteredEntities = new Set();
 
   /**
    * Determines if entities in this collection will be shown.
@@ -142,7 +143,13 @@ function addNonClusteredItem(item, entityCluster) {
   }
 }
 
-function addCluster(position, numPoints, ids, entityCluster) {
+function addCluster(
+  position,
+  numPoints,
+  ids,
+  entityCluster,
+  currentClusteredEntities,
+) {
   const cluster = {
     billboard: entityCluster._clusterBillboardCollection.add(),
     label: entityCluster._clusterLabelCollection.add(),
@@ -159,7 +166,47 @@ function addCluster(position, numPoints, ids, entityCluster) {
     cluster.point.position =
       position;
 
+  for (let i = 0; i < ids.length; ++i) {
+    currentClusteredEntities.add(ids[i]);
+  }
+
   entityCluster._clusterEvent.raiseEvent(ids, cluster);
+}
+
+function updateClusterMembership(entityCluster, currentClusteredEntities) {
+  const previousClusteredEntities = entityCluster._previousClusteredEntities;
+
+  const enteredEntities = [];
+  currentClusteredEntities.forEach(function (entity) {
+    if (!previousClusteredEntities.has(entity)) {
+      enteredEntities.push(entity);
+    }
+  });
+
+  const exitedEntities = [];
+  previousClusteredEntities.forEach(function (entity) {
+    if (!currentClusteredEntities.has(entity)) {
+      exitedEntities.push(entity);
+    }
+  });
+
+  if (enteredEntities.length > 0 || exitedEntities.length > 0) {
+    entityCluster._clusterMembershipChangedEvent.raiseEvent(
+      enteredEntities,
+      exitedEntities,
+    );
+  }
+
+  if (
+    previousClusteredEntities.size > 0 &&
+    currentClusteredEntities.size === 0
+  ) {
+    entityCluster._allDeclusteredEvent.raiseEvent(
+      Array.from(previousClusteredEntities),
+    );
+  }
+
+  entityCluster._previousClusteredEntities = currentClusteredEntities;
 }
 
 function hasLabelIndex(entityCluster, entityId) {
@@ -233,6 +280,7 @@ function createDeclutterCallback(entityCluster) {
     const labelCollection = entityCluster._labelCollection;
     const billboardCollection = entityCluster._billboardCollection;
     const pointCollection = entityCluster._pointCollection;
+    const currentClusteredEntities = new Set();
 
     if (
       (!defined(labelCollection) &&
@@ -242,6 +290,9 @@ function createDeclutterCallback(entityCluster) {
         !entityCluster._clusterLabels &&
         !entityCluster._clusterPoints)
     ) {
+      entityCluster._previousClusters = [];
+      entityCluster._previousHeight = scene.camera.positionCartographic.height;
+      updateClusterMembership(entityCluster, currentClusteredEntities);
       return;
     }
 
@@ -388,7 +439,13 @@ function createDeclutterCallback(entityCluster) {
           }
 
           if (numPoints >= minimumClusterSize) {
-            addCluster(cluster.position, numPoints, ids, entityCluster);
+            addCluster(
+              cluster.position,
+              numPoints,
+              ids,
+              entityCluster,
+              currentClusteredEntities,
+            );
             newClusters.push(cluster);
 
             for (j = 0; j < neighborLength; ++j) {
@@ -469,7 +526,13 @@ function createDeclutterCallback(entityCluster) {
             1.0 / numPoints,
             clusterPosition,
           );
-          addCluster(position, numPoints, ids, entityCluster);
+          addCluster(
+            position,
+            numPoints,
+            ids,
+            entityCluster,
+            currentClusteredEntities,
+          );
           newClusters.push({
             position: position,
             width: totalBBox.width,
@@ -502,40 +565,9 @@ function createDeclutterCallback(entityCluster) {
       entityCluster._clusterPointCollection = undefined;
     }
 
-    const currentlyClusteredIds = [];
-
-    if (defined(clusteredLabelCollection)) {
-      for (let c = 0; c < clusteredLabelCollection.length; ++c) {
-        const clusterLabel = clusteredLabelCollection.get(c);
-        currentlyClusteredIds.push(...clusterLabel.id);
-      }
-    }
-
-    if (defined(clusteredBillboardCollection)) {
-      for (let c = 0; c < clusteredBillboardCollection.length; ++c) {
-        const clusterBillboard = clusteredBillboardCollection.get(c);
-        currentlyClusteredIds.push(...clusterBillboard.id);
-      }
-    }
-
-    if (defined(clusteredPointCollection)) {
-      for (let c = 0; c < clusteredPointCollection.length; ++c) {
-        const clusterPoint = clusteredPointCollection.get(c);
-        currentlyClusteredIds.push(...clusterPoint.id);
-      }
-    }
-
-    const hasActiveClusters = currentlyClusteredIds.length > 0;
-    const hadPreviouslyClusters =
-      entityCluster._previouslyClusteredEntities.length > 0;
-
-    if (!hasActiveClusters && hadPreviouslyClusters) {
-      entityCluster._allDeclusteredEvent.raiseEvent(
-        entityCluster._previouslyClusteredEntities,
-      );
-    }
-
-    entityCluster._previouslyClusteredEntities = currentlyClusteredIds;
+    entityCluster._previousClusters = newClusters;
+    entityCluster._previousHeight = currentHeight;
+    updateClusterMembership(entityCluster, currentClusteredEntities);
   };
 }
 
@@ -609,6 +641,16 @@ Object.defineProperties(EntityCluster.prototype, {
   allDeclusteredEvent: {
     get: function () {
       return this._allDeclusteredEvent;
+    },
+  },
+  /**
+   * Gets the event that will be raised when clustered entity membership changes.
+   * @memberof EntityCluster.prototype
+   * @type {Event<EntityCluster.clusterMembershipChangedCallback>}
+   */
+  clusterMembershipChangedEvent: {
+    get: function () {
+      return this._clusterMembershipChangedEvent;
     },
   },
   /**
@@ -910,9 +952,15 @@ function updateEnable(entityCluster) {
   entityCluster._clusterBillboardCollection = undefined;
   entityCluster._clusterPointCollection = undefined;
 
+  entityCluster._previousClusters = [];
+  entityCluster._previousHeight =
+    entityCluster._scene.camera.positionCartographic.height;
+
   disableCollectionClustering(entityCluster._labelCollection);
   disableCollectionClustering(entityCluster._billboardCollection);
   disableCollectionClustering(entityCluster._pointCollection);
+
+  updateClusterMembership(entityCluster, new Set());
 }
 
 /**
@@ -1037,7 +1085,7 @@ EntityCluster.prototype.destroy = function () {
 
   this._previousClusters = [];
   this._previousHeight = undefined;
-  this._previouslyClusteredEntities = [];
+  this._previousClusteredEntities = new Set();
 
   this._enabledDirty = false;
   this._pixelRangeDirty = false;
@@ -1065,6 +1113,18 @@ EntityCluster.prototype.destroy = function () {
  * });
  */
 /**
+ * An event listener function used when clustered entity membership changes.
+ * @callback EntityCluster.clusterMembershipChangedCallback
+ *
+ * @param {Entity[]} enteredEntities An array of entities that are now part of a cluster.
+ * @param {Entity[]} exitedEntities An array of entities that are no longer part of a cluster.
+ *
+ * @example
+ * dataSource.clustering.clusterMembershipChangedEvent.addEventListener(function(entered, exited) {
+ *     console.log('Entered clusters:', entered, 'Exited clusters:', exited);
+ * });
+ */
+/**
  * An event listener function used when all entities have been declustered.
  * @callback EntityCluster.declusterCallback
  *
@@ -1073,7 +1133,7 @@ EntityCluster.prototype.destroy = function () {
  *
  * @example
  * // Listen for decluster events
- * dataSource.clustering.declusterEvent.addEventListener(function(entities) {
+ * dataSource.clustering.allDeclusteredEvent.addEventListener(function(entities) {
  *     console.log('All clusters removed. Last entities declustered:', entities);
  * });
  */
--- a/packages/engine/Specs/DataSources/EntityClusterSpec.js
+++ b/packages/engine/Specs/DataSources/EntityClusterSpec.js
@@ -701,13 +701,12 @@ describe(
       );
     });
 
-    it("tracks previously clustered entities", function () {
+    it("has clusterMembershipChangedEvent property", function () {
       cluster = new EntityCluster();
-      cluster._initialize(scene);
-
-      expect(cluster._previouslyClusteredEntities).toBeDefined();
-      expect(Array.isArray(cluster._previouslyClusteredEntities)).toEqual(true);
-      expect(cluster._previouslyClusteredEntities.length).toEqual(0);
+      expect(cluster.clusterMembershipChangedEvent).toBeDefined();
+      expect(
+        typeof cluster.clusterMembershipChangedEvent.addEventListener,
+      ).toEqual("function");
     });
 
     it("fires allDeclusteredEvent when all clusters are removed", function () {
@@ -751,7 +750,9 @@ describe(
         expect(eventFired).toEqual(true);
         expect(receivedEntities).toBeDefined();
         expect(Array.isArray(receivedEntities)).toEqual(true);
-        expect(receivedEntities.length).toBeGreaterThan(0);
+        expect(receivedEntities.length).toEqual(2);
+        expect(receivedEntities).toContain(entity1);
+        expect(receivedEntities).toContain(entity2);
       });
     });
 
@@ -796,9 +797,19 @@ describe(
       });
     });
 
-    it("tracks previously clustered entities during clustering", function () {
+    it("fires clusterMembershipChangedEvent for entered and exited entities", function () {
       cluster = new EntityCluster();
       cluster._initialize(scene);
+
+      const membershipChanges = [];
+      cluster.clusterMembershipChangedEvent.addEventListener(
+        function (enteredEntities, exitedEntities) {
+          membershipChanges.push({
+            enteredEntities: enteredEntities,
+            exitedEntities: exitedEntities,
+          });
+        },
+      );
 
       const entity1 = new Entity();
       const point1 = cluster.getPoint(entity1);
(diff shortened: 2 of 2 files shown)


packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -869,6 +951,32 @@ function updateEnable(entityCluster) {
   disableCollectionClustering(entityCluster._labelCollection);
   disableCollectionClustering(entityCluster._billboardCollection);
   disableCollectionClustering(entityCluster._pointCollection);
+
+  if (allVisibleEntities.length > 0) {
+    const uniqueEntities = [...new Set(allVisibleEntities)];
```
javagl: Don't use ([push.apply](https://github.com/CesiumGS/cesium/issues/12053) or) the `...spread` operator for potentially large collections. This should be
`const uniqueEntities = Array.from(new Set(allVisibleEntities));`

Don't drain performance with useless computations. When there are no listeners attached to that event, then there is no reason to compute all this. Insert some
`if (entityCluster._declusteredEvent.numberOfListeners !== 0) { ... }`
to compute this only when there are listeners registered.

Oko-Tester:
this is ready

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -869,6 +922,67 @@ function updateEnable(entityCluster) {
...
+    }
+
+    if (defined(entityCluster._pointCollection)) {
+      for (let i = 0; i < entityCluster._pointCollection.length; i++) {
+        const point = entityCluster._pointCollection.get(i);
+        if (defined(point.id) && point.show) {
+          allVisibleEntities.push(point.id);
+        }
+      }
+    }
```
mzschwartz5: This looks like a prime candidate for a function, so we don't repeat basically the same thing 3x. Then you could even early-return on the `defined` check and save a level of indentation.

That said, these collections do not share an interface, so there's no guarantee they will continue to have these common methods and fields (show, id, get)

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -869,6 +922,67 @@ function updateEnable(entityCluster) {
   disableCollectionClustering(entityCluster._labelCollection);
   disableCollectionClustering(entityCluster._billboardCollection);
   disableCollectionClustering(entityCluster._pointCollection);
+
+  if (entityCluster._declusteredEvent.numberOfListeners !== 0) {
```
mzschwartz5: It feels like the level of conditional branching here may be unnecessary. I'm always an advocate of trying to reduce the levels of nesting for clarity - usually be introducing new functions and then early-returning from them.

javagl: I think that this was added in response to my comment at https://github.com/CesiumGS/cesium/pull/12920#discussion_r2378619175 . And the _behavior_ should be kept like that - it should not do that visible-stuff-computation when no listeners are present. 

We can argue about details of the style. Some people _strongly_ advocate for "One `return` statement per function!". But I think that "watchdogs" and "early returns" can often increase readability. So when there is something like
```
if (condition) {
    // 100 lines
} else {
    // 2 lines
}
```
I'm also leaning towards the pattern of
```
// Don't do anything when not necessary
if (!condition) {
    // 2 lines
    return;
}

// Now do the real work
// 100 lines
```
(No strong opinion here)

mzschwartz5: Yeah this was more of a comment on style than behavior. 

Also, will say, I _do_ have strong opinions about the early return pattern. I think by nature of being _early_, it's kind of an exception to the no-multiple-returns guideline. In my mind, the point of no-multiple-returns is to avoid missing a deeply nested return, which can drastically change the meaning of what comes after. Short functions, with early returns don't run into that issue.

javagl: "Rules are for the obedience of fools and the guidance of wise men" (D. Bader)

When there's a rule that there shouldn't be multiple returns per function, then some people will follow that rule, no matter what. However, I'm also in the camp of "having to _scroll_ to find an `else` is a bad sign".

packages/engine/Source/DataSources/EntityCluster.js
Oko-Tester: Yea ur right. I´ll extract this into a helper function.

packages/engine/Source/DataSources/EntityCluster.js
Oko-Tester: If there are no specific requests to change the logic or further customise the style, I will keep the current implementation unchanged, provided that is acceptable to you.

Oko-Tester:
@mzschwartz5 Is there any news regarding the approval?

mzschwartz5:
Hi @Oko-Tester, I haven't had a chance to review this yet. It might not happen until Thursday as I'm out tomorrow and Wednesday.

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -998,9 +1105,37 @@ EntityCluster.prototype.destroy = function () {
...
+ */
+EntityCluster.prototype.getLastDeclusteredEntities = function () {
+  return this._lastDeclusteredEntities.slice();
+};
+
+/**
+ * Returns all entities that were processed in the most recent clustering operation.
+ * @returns {Entity[]} Array of all processed entities
+ */
+EntityCluster.prototype.getAllProcessedEntities = function () {
```
mzschwartz5: This is an opinion, but I feel like the name of this API could use some work shopping - before I even suggest an alternative, though, let me ask: does "all" entities mean all entities that are in clusters after the most recent operation, or also entities that are declustered?

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -998,9 +1105,37 @@ EntityCluster.prototype.destroy = function () {
...
+
   return undefined;
 };
 
+/**
+ * Returns the last set of clustered entities from the most recent clustering operation.
+ * @returns {Entity[]} Array of entities that were clustered
+ */
+EntityCluster.prototype.getLastClusteredEntities = function () {
+  return this._lastClusteredEntities.slice();
```
mzschwartz5: Is there any reason to return a shallow copy of the internal array (via `slice()`)? Why not just make a property / standard getter for these arrays?

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -847,6 +900,58 @@ function disableCollectionClustering(collection) {
...
+  }
+  return visibleEntities;
+}
+
+function handleDeclusteredEvent(entityCluster) {
+  if (entityCluster._declusteredEvent.numberOfListeners === 0) {
+    return;
+  }
+  const allVisibleEntities = [
+    ...getVisibleEntitiesFromCollection(entityCluster._labelCollection),
```
mzschwartz5: I know there was previous discussion about the use of the spread (`...`) operator on large collections. Were these instances overlooked or intentionally left as-is because we know they're not large?

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -847,6 +900,58 @@ function disableCollectionClustering(collection) {
...
+    ...getVisibleEntitiesFromCollection(entityCluster._billboardCollection),
+    ...getVisibleEntitiesFromCollection(entityCluster._pointCollection),
+  ];
+
+  if (allVisibleEntities.length > 0) {
+    const uniqueEntities = Array.from(new Set(allVisibleEntities));
```
mzschwartz5: Why is it possible for `allVisibleEntities` to contain duplicates?

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -847,6 +900,58 @@ function disableCollectionClustering(collection) {
...
+      cluster: null,
+      allProcessed: [],
+    });
+
+    entityCluster._lastClusteredEntities = [];
+    entityCluster._lastDeclusteredEntities = [];
+    entityCluster._allProcessedEntities = [];
+  }
+}
+
```
mzschwartz5: I don't believe this need any conditional logic. Just set 
`const uniqueEntities = Array.from(new Set(allVisibleEntities));` 

and the rest handles itself whether `allVisibleEntities` is empty or not.

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -847,6 +900,58 @@ function disableCollectionClustering(collection) {
   }
 }
 
+function getVisibleEntitiesFromCollection(collection) {
+  if (!defined(collection)) {
+    return [];
```
mzschwartz5: small style nit - define 
`const visibleEntities = [];`

before this, and return it instead of a literal `[]`

packages/engine/Source/DataSources/EntityCluster.js
mzschwartz5: (And similarly rename the internal array)

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -567,6 +610,16 @@ Object.defineProperties(EntityCluster.prototype, {
       return this._clusterEvent;
     },
   },
+  /**
+   * Gets the event that will be raised when clustering is processed, including both clustered and declustered entities.
```
mzschwartz5: So does this event do everything the pre-existing `clusteredEvent` did, and more? Because the comment says in includes clustered entities. So it's not just an event for declustering but also for clustering?

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -485,6 +516,18 @@ function createDeclutterCallback(entityCluster) {
...
 
+    if (
+      entityCluster._lastClusteredEntities.length > 0 ||
+      entityCluster._lastDeclusteredEntities.length > 0
+    ) {
+      entityCluster._declusteredEvent.raiseEvent({
+        clustered: entityCluster._lastClusteredEntities.slice(),
+        declustered: entityCluster._lastDeclusteredEntities.slice(),
+        cluster: null,
+        allProcessed: entityCluster._allProcessedEntities.slice(),
```
mzschwartz5: Similar to my other comment - do we want / need to be creating a shallow copy of these arrays here, via `.slice()`?

packages/engine/Specs/DataSources/EntityClusterSpec.js
mzschwartz5, line 1: I'm not an expert on unit testing philosphy, so maybe this is a bad suggestion, but could we maybe have unit test(s) that check that clustering generally works for all three of: points, billboards, and labels (collections)?

I'm just a little wary given the discussion about using the `item.show` property as if these classes all have a shared abstract interface (which they don't). I don't _expect_ that to be an issue, but if someone were to ever change that property name on one of the classes, at least having a failing unit test here would prevent the issue.

mzschwartz5 (changes requested):
Hey @Oko-Tester,

Thanks for the PR. I have a number of mostly minor comments about code style and performance, but those might be irrelevant because I'm a little unsure of the overall approach. It's a lot of extra state that we're tracking now, which means more memory consumed, more bookkeeping to sync arrays, more room for potential edge cases. 

In the linked issue you said "clusterEvent does not fire when a cluster becomes empty, which makes it impossible to detect de-clustered entities." My intuition is that there's a better way to deal with this than to track more state and manually fire the decluster event accordingly. I haven't actually looked at the code closely to tell, but that's my instinct.

Oko-Tester:
Hey @mzschwartz5,
Thanks for your feedback. I'll rethink my current implementation and maybe find a better way. But I won't be able to deal with this MR for a few days.

As for your comment on unit tests, I am not an expert in this field either, but it seems sensible to me to use them to avoid potential future problems.

packages/engine/Source/DataSources/EntityCluster.js
javagl: With zero context, I'll throw in the baseline comment: Minimize mutablilty and make ownership clear. It _at least_ serves the purpose of avoiding the nearly-impossible-to-debug glitches that may result from a client doing this:
```
const entities = entityCluster.getLastClusteredEntities();

// (French accent:) A few moments later...
entities[42] = undefined;

// Even later
entityCluster.updateSomething(); // Crashes due to `undefined` in array...
```
It's a thin protection layer, but better than nothing. In other environments, this would/could/should return a read-only _view_ on the internal state (`Collections.unmodifiableList` et al).

mzschwartz5: Hmm... I see. I would still rather we protect against undefined in `updateSomething`. 

In a typed language, you would just have a compile-time contract here, marking this function as returning something `const` or read-only. But even there, a user can generally find a way to shoot themselves in the foot at run-time, if they really want to.

So, here too, a user _could_ do something crazy with the array, but we can't guard against everything. (But we should still guard against the basic things)

Oko-Tester:
Okay, I rethought my current implementation and realized that I had adapted the component too much for my specific needs.
It would be better to simply add a single event that returns the last declustered entities — nothing more. The rest can easily be handled manually by the user.

I’ve reset the EntityCluster component locally and will commit the new changes shortly after testing them. I also need to update the tests to match the new implementation.

I’d really appreciate your feedback on this new approach once the changes are pushed.

mzschwartz5:
@Oko-Tester When you're ready, just re-request review from me and/or leave a comment saying you're ready for another pass. Depending on when it's ready, I may need a few days before I can take another look.

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -500,8 +502,28 @@ function createDeclutterCallback(entityCluster) {
...
+    const currentlyClusteredIds = [];
+
+    if (newClusters.length > 0 && defined(clusteredLabelCollection)) {
+      for (let c = 0; c < clusteredLabelCollection.length; ++c) {
+        const clusterLabel = clusteredLabelCollection.get(c);
+        if (defined(clusterLabel) && defined(clusterLabel.id)) {
+          currentlyClusteredIds.push(...clusterLabel.id);
+        }
+      }
+    }
```
mzschwartz5: 1. Why do we need to check `newClusters.length > 0`?
2. Will it ever be the case that `clusterLabel` or `clusterLabel.id` is not defined? When?
3. Why do we need to use the spread operator (`...`) when pushing an ID? To make a copy of it (if so, why isn't a reference sufficient?)

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -66,6 +66,8 @@ function EntityCluster(options) {
   this._removeEventListener = undefined;
 
   this._clusterEvent = new Event();
+  this._declusterEvent = new Event();
```
mzschwartz5: I feel like this is a bit misleading. `clusterEvent` fires whenever a new cluster is added, right? But `declusterEvent` only fires when _all_ clusters are removed. (If I've understood this PR correctly)

It seems like `declusterEvent` should also fire whenever a cluster is removed. (I don't know if that's feasible to implement, though)

packages/engine/Source/DataSources/EntityCluster.js
```diff
@@ -1019,4 +1052,17 @@ EntityCluster.prototype.destroy = function () {
  *     cluster.label.text = entities.length.toLocaleString();
  * });
  */
+/**
+ * A event listener function used when all entities have been declustered.
```
mzschwartz5: Small typo- "A event" --> "An event"

packages/engine/Source/DataSources/EntityCluster.js
Oko-Tester: You are correct about the asymmetry. The behavior is as follows:
clusterEvent: Triggered every time addCluster is called, which happens in the following cases:

When new clusters are created
When existing clusters are reused (when the camera zooms in and they still meet the minimumClusterSize threshold)

declusterEvent: Only fires when transitioning from "has clusters" to "no clusters at all". It does NOT fire when individual clusters are removed.
However, users can still detect which clusters have been removed by tracking the entity IDs themselves. Since clusterEvent is triggered every frame with the current cluster composition, you can compare frame by frame to determine which entities were declustered.

packages/engine/Source/DataSources/EntityCluster.js
Oko-Tester: 1. Why check newClusters.length > 0?
This check is redundant. If there are no clusters, clusteredLabelCollection.length would be 0 anyway, so the loop wouldn't execute. I'll remove this check.
2. Will clusterLabel or clusterLabel.id ever be undefined?
In normal operation, no. clusterLabel.id is always set in addCluster(). I think keeping these defensive checks for safety would be good.
3. Why use the spread operator (...)?
Because clusterLabel.id is an array of entity IDs (set in addCluster()), not a single ID. The spread operator flattens it. Without it, you'd have an array of arrays instead of a flat array of IDs.

javagl: Random drive-by comment: 

> In normal operation, no. clusterLabel.id is always set in addCluster(). I think keeping these defensive checks for safety would be good.

When something should never be `undefined`, then this should be a _contract_ that people can _rely_ on. When one of these things _is_ ever `undefined`, then this is a bug, and it will show up as a crash, with a clear stack trace. With these checks, it will _not_ show up with a stack trace. But it will **still** be a bug. And this bug will only show up as an issue like "Sometimes some labels are not properly shown in some clusters at some zoom level" or so. And one can easily spend 5 or 10 hours of painful debugging to find the reason for an issue like that.

I know, there are different philosophies. But there's an _important_ difference between being "defensive"/"resilient", and what I refer to as "fail silent". 
"Defensive" would mean that there is a `defined` check, and when the thing is `undefined`, then it does **not** crash, but it _still_ logs an error to the console, saying "This is `undefined`, but it never should be". Ideally, this should not be necessary _within_ a library. It should only be necessary for things that are provided _from outside_, via the public API. But... that ship has sailed.
"Fail silent" means that something is wrong, and there's not way to figure out _what_ is wrong. That's not good.

packages/engine/Source/DataSources/EntityCluster.js
javagl: A wild guess is that this question may have to do with the _name_.

When you have methods like
`addSomething`/`removeSomething` or
`createSomething`/`destroySomething`
then you'd also expect a certain "symmetry". 
<sub>(And when you have `addSomething`, `removeSomething`, and `deleteSomething`, you'll wonder: Wait what...?)</sub>

In this case, one could expect this event to be raises when ... the opposite of that happened what causes a `clusterEvent`. Yes, there is the documentation that says that this event _"... will be raised when **all entities** have been declustered."_, but still.

@mzschwartz5 It's up to you, but I think that options could be to

- rename this event (to something like `declusteredAllEvent` or `allDeclusteredEvent` or ... whatever is appropriate here)
- and (!!!optional!!!) consider adding a `declusterEvent` that indeed is the opposite of what `clusterEvent` is doing

Oko-Tester: Yes, you’re right about the naming confusion. I think we should rename the event to lastDeclusterEvent (or allDeclusteredEvent, whatever you prefer) so it’s clear that it only fires when the last cluster is removed. But there’s no real need for a new event that tracks every declustering — users can already detect removed clusters by comparing the cluster composition from the clusterEvent. So I don’t think an additional event for every declustering is necessary.

packages/engine/Source/DataSources/EntityCluster.js
Oko-Tester: Good point, I didn’t consider that. I think it was mentioned earlier in this PR, but you’re right, better to fail visibly than silently.

packages/engine/Source/DataSources/EntityCluster.js
javagl: I'm not deeply involved here - it was just a comment while looking at my GitHub notifications. So I don't know in how far it makes sense to define an event for individual removed clusters. **But**... (again: a bit shallow)...: 

> users can already detect removed clusters by comparing the cluster composition from the clusterEvent. So I don’t think an additional event for every declustering is necessary.

Users _can_ do a lot of stuff. Users can track this and that and store this and that and compute some difference here and there. The question is whether there's a convincing use case for being informed about individual clusters. If there is, then uers should not be forced to write 200 lines of (possibly buggy) code to derive the information they need, when they could just listen for a single, predefined event. 

Maybe there is no such use case. And in doubt, this new event could be added _later_, when the demand arises. And following the mantra: "When in doubt, leave it out!", it's probably OK to omit it for now (and certainly in the context of this PR)

mzschwartz5: I generally agree with @javagl, I and I think it would be sufficient for this PR to simply rename this event. I liked the proposed "allDeclusteredEvent"

packages/engine/Source/DataSources/EntityCluster.js
mzschwartz5: Okay so sounds like we can address points 1 and 2 easily.

For point 3 (the spread operator) - are we copying the _all_ currently clustered IDs (i.e. every ID of every entity that is clustered) on every declutter callback? That seems like it could be bad for performance. It also doubles the memory we consume in tracking which entities are clustered (since we always have a copy of the `previouslyClustered...`)

Side note: does this only fire a declustered event for `clusteredLabelCollection`? But not for `clusteredBillboardCollection` or `clusteredPointCollection`?

Oko-Tester: I will finish this PR soon. At the moment, I don’t have much time to spare.
Thanks a lot for the valuable feedback, and sorry for the long waiting time!

And yes the clusterBillboardCollection and clusterPointCollection is not implemented yet.

mzschwartz5: Hey, no rush :) take your time. We always have plenty to review anyway!

Oko-Tester: Okay, I've addressed points 1 and 2. I also added the missing implementation for clusteredBillboardCollection and clusteredPointCollection.

Regarding point 3: The spread operator is still needed since clusterLabel/Billboard/Point.id is an array of entity IDs. However, @mzschwartz5 concern about performance is valid, how should this be addressed?

Oko-Tester:
If there are no more suggestions for improvement, I would say that the PR is ready.

mzschwartz5:
If I have time today I'll take another look! Otherwise early next week

mzschwartz5:
Okay, I've refreshed my memory on this PR. I also went back and took another look through the [original issue](github.com/CesiumGS/cesium/issues/5760) and linked forum posts. 

Correct me if I'm wrong - it seems like the original desire is basically for users to be able to track entity membership within clusters. `clusterEvent` is called whenever a new cluster is formed, but not when a cluster is completely dissolved (which makes it impossible to derive cluster membership) - that's what you're trying to add.

The main issue I have with the current state of the PR is that concern I voiced about performance and memory. Every time `createDeclutterCallback` runs, we're going to be iterating over all the entities in all clusters. And storing the last clusters' entities as well now.

And for a user who wants to determine which entities are now part of a cluster or no longer part of a cluster, they will presumably have to write code to _also_ store previously clustered entities and compare against the entities in `clusterEvent` callback. Which is again non-performant and consumes more memory. 

Ultimately, the reason this PR is challenging is not because of a shortcoming of your code. It's because the `EntityCluster` API is designed poorly (in my opinion). Rather than reconstructing clusters from scratch every time they're dirty, this class should be moving entities into and out of existing clusters (or forming new ones when needed). Done that way, the API could then _very easily_ expose callbacks that inform users _which_ entities / clusters changed, rather than just throwing the full state at them. However, that's obviously a very big refactor. But, I think it may be necessary... unless you know of a better way 😓 

P.S. (I'm not demanding that you do this big refactor. I'm just lamenting that I don't think we can bandaid this functionality in without paying a significant performance (or other) cost).

Oko-Tester:
You’re right about the reason for this PR, and I understand why the refactoring is necessary. Unfortunately, I don’t have the capacity to tackle this refactoring right now, and I won’t be able to implement it in the near term. I might be able to look into it in one or two months, but I can’t make any promises.

I’ve reviewed the implementation again, and as an interim solution, we could calculate which clusters are no longer relevant during the creation of the new clusters. This would allow us to return the IDs that are no longer in use to the user. That way, the user doesn’t need to store anything on their side, and we don’t need to persist any previous declusters either.

Oko-Tester:
I took another look at the whole thing and tried to simplify the approach for this PR.
I agree that a bigger refactor (persistent clusters with proper add/remove updates) would probably be the better long-term solution, but I didn’t go that far here.

What I changed instead:
- added clusterMembershipChangedEvent(enteredEntities, exitedEntities)
- collect clustered entities while building the clusters, instead of doing an extra full pass afterwards
- use a Set diff between previous and current membership to reduce duplication and memory usage
- left the existing clusterEvent behavior as-is
- kept allDeclusteredEvent for the non-empty → empty case

This isn’t the full architectural refactor yet, but it should fix the main membership/performance issues in the current implementation.

PR #12922, opened 2025-09-25
Fix background updating when label hidden #12138

Beilinson:
# Description

There is a bug currently where updating certain label properties with `labelPrimitive.showBackground=true` and `labelPrimitive.show=false` will cause the background to disappear after setting `labelPrimitive.show=true`. The first sandcastle from the linked issue demonstrates this.

### Cause

Label backgrounds are removed from the `backgroundBillboardCollection` when either `showBackground=false`, `show=false`, or `label.text.length === 0`. They are re-added properly, however their width and height is not updated because `repositionAllGlyphs()` is called ahead of that.

### Solution

When hiding the label entirely, don't remove the `backgroundBillboard`, rather set `backgroundBillboard.show = false`. In order to prevent useless billboards, **do remove** if `showBackground=false` is set explicitly.

## Issue number and link

#12138

## Testing plan

[Link to sandcastle 1](https://sandcastle.cesium.com/#c=zVVbT9swFP4rXrSHZCtuC7RAb9oo3ZiEtmmg7SUPuMlJa+HYke20MOh/n507pWhs0qRVahKf71y/c3ISCK40WlFYg0RjxGGNpqBoGuPvmcz1nSA7TwXXhHKQvuMNfc5AI0bmwC6o0mDEU8LYnAQ3BsvRyBicAdPEuO12cGdYiy/pTzDS/VwaZCnElH+ooW4TIbcN5PA4g0owEIxBoKngj7O/sLlNK9C1OedVYhWYfHEiaUw1XYHCJAzd2o/X8A5cU31nPBem2ZkWJvc+RygRilqrQRl4SqQ2T4Qf4EiK+AwWEkC5e0c93O0fdvv9oxY6OMEnvX3781rWSUbkAGUOEdJwqwfId/LgvtPKxWop1qeG4IUUKQ8HSMsUCiiizNYqZJ2FPeGP32azz4XOvDLdpfnj/NPVDK+pXr5nyZK4HXzc8xqBH4cz3Rig69f3ZS83yS0y/adzSa8znY25bppEVmwbLmuq/4bGTgsdFrSVRNXO0yQkGsKCs2cZ+5d8bbH1O64KluybIWGeUqa/ltUYIEp5Pts5FlZYNt6ujeXlY7Nt/AzPf8B0r2+YPimH4CnXRcT/ZT7t9eXzaVnf+Hw3rTbzvC0+vyQ8DIjSDCyJV2KxYHCaam1Wiu+c0xBQZqMMDSgiTIG5lU1zgyUENxAWLcrfZ2wTNd15VYDDbPzLBHai9t9opoRYrMDd7rhX6dJoOzSq1+7bcb2Zhzlm9St8Mn60cB8eatPRuLmlK9eoserfjNFevrutfJPfisKzLYetrinwme4UhjUfL1DflHXv7mZJZdZyb+i0nJHSdwwmZfbvaJwIqVEqmYtxW0OcMLNFVHueGjuNA6W8Iq9Ru2k6CukK0XC84xOJAkaUMkiUMmaz9p3JqG30n5gyQULKF19WIBm5s2rL7uQiF2KMR21z3G2phTBfXLnl+Rc)

# Author checklist

~- [ ] I have added or updated unit tests to ensure consistent code coverage~
~- [ ] I have updated the inline documentation, and included code examples where relevant~

### DIFF
--- a/packages/engine/Source/Scene/LabelCollection.js
+++ b/packages/engine/Source/Scene/LabelCollection.js
@@ -157,8 +157,12 @@ function rebindAllGlyphs(labelCollection, label) {
     labelCollection._backgroundBillboardCollection;
   if (!showBackground) {
     if (defined(backgroundBillboard)) {
-      backgroundBillboardCollection.remove(backgroundBillboard);
-      label._backgroundBillboard = backgroundBillboard = undefined;
+      if (label.show) {
+        backgroundBillboardCollection.remove(backgroundBillboard);
+        label._backgroundBillboard = backgroundBillboard = undefined;
+      } else {
+        backgroundBillboard.show = false;
+      }
     }
   } else {
     if (!defined(backgroundBillboard)) {
@@ -506,7 +510,7 @@ function repositionAllGlyphs(label) {
       glyphPixelOffset.y = 0;
     }
     glyphPixelOffset.y = glyphPixelOffset.y * scale;
-
+    backgroundBillboard.scale = scale;
     backgroundBillboard.width = totalLineWidth;
     backgroundBillboard.height = totalLineHeight;
     backgroundBillboard._setTranslate(glyphPixelOffset);


packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -506,7 +510,7 @@ function repositionAllGlyphs(label) {
       glyphPixelOffset.y = 0;
     }
     glyphPixelOffset.y = glyphPixelOffset.y * scale;
-
+    backgroundBillboard.scale = scale;
```
mzschwartz5: Based on my reading of the issue and your PR description, I don't yet understand why it's necessary to explicitly set the scale here. How did (the lack of setting scale previously) play into this issue?

packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -157,8 +157,12 @@ function rebindAllGlyphs(labelCollection, label) {
...
   if (!showBackground) {
     if (defined(backgroundBillboard)) {
-      backgroundBillboardCollection.remove(backgroundBillboard);
-      label._backgroundBillboard = backgroundBillboard = undefined;
+      if (label.show) {
+        backgroundBillboardCollection.remove(backgroundBillboard);
+        label._backgroundBillboard = backgroundBillboard = undefined;
+      } else {
+        backgroundBillboard.show = false;
+      }
```
mzschwartz5: 1. This is a little redundant with surrounding code. There's already this line in the corresponding `else` block:
`backgroundBillboard.show = label._show;`
It would be best to consolidate the conditional flow so we don't have two branches setting the same field
2. Similarly, there are too many levels of indentation at this point. I know it doesn't seem like a big deal, but every time someone comes along and adds an innocuous fix like this, we end up with another level of indentation, and it becomes incrementally harder to understand and change in the future.

I think we need to bite the bullet here and refactor the overarching `if (!showBackground)... else` block into functions. Then, in those functions, rather than using nested conditionals, you can early-return if certain conditions are false. (e.g. `if (!showBackground) return`).

mzschwartz5 (changes requested):
Good start, thanks for identifying the issue and a fix for it!

I have a couple style comments and questions before we merge.

packages/engine/Source/Scene/LabelCollection.js
Beilinson: Was about to write some funny comments, then I realized its because I still don't update any of the backgroundBillboard properties in this PR when `label.show === false`.

mzschwartz5:
Oh, also - unit tests! We're definitely going to want some coverage of this edge case you're fixing.

packages/engine/Source/Scene/LabelCollection.js
Beilinson: I agree this is too much indentation, after the fix given due to your other comment I managed to refactor this back down to just two levels of indentation like before

packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -155,20 +155,25 @@ function rebindAllGlyphs(labelCollection, label) {
...
-    if (!defined(backgroundBillboard)) {
-      backgroundBillboard = getWhitePixelBillboard(
-        backgroundBillboardCollection,
-        labelCollection,
-      );
-      label._backgroundBillboard = backgroundBillboard;
-    }
-
+  }
+  if (defined(backgroundBillboard)) {
```
mzschwartz5: I appreciate the effort to simplify but... I still find this very confusing (perhaps even more, now).

There's `_showBackground` and `showBackground` and `label.show` and `backgroundBillboard.show`, three places where we're checking `defined(backgroundBillboard)`... Not that you introduced these things.

I know this can be clearer, but I don't have the bandwidth currently to wrap my head around this and offer an alternative. I think my original comment about refactoring into simple functions with early returns would still help a lot. Then, at least, the conditionals would be named and thus more readable. If that doesn't make sense, I'll try to come back to this tomorrow.

Beilinson: Alright, had a go at separating it out into a separate function and added some comments, thoughts?

mzschwartz5: Yes I think this is much clearer, thank you. I would even do more, but it's personal preference:
1. Could move the `if (defined(backgroundBillboard))` check into `updateBackgroundBillboard` and return early on false.
2. Rather than having the comment `  // Create backgroundBillboard if needed` (which *did* help clarify things for me), you could just take that block of code and make it into a named function, like`maybeCreateBackgroundBillboard`.

mzschwartz5: If you think those are worthwhile changes, go for it, if not, let me know; I think it's good to merge either way.

Beilinson: 1 sounds good to me, 2 imo is not needed. Fixing conflicts and this is ready on my end

mzschwartz5:
I think you need to:
1. Resolve merge conflicts
2. Rerequest review so I can formally approve
3. Then I'll merge (though I'll be out for two days, so when I'm back)

mzschwartz5:
Good to go - thanks a lot for the contribution!

Issue #12925, opened 2025-09-25
Low quality Gaussian splat rendering when child tiles contain their world transform


keyboardspecialist:
### What happened?

When rendering splats we transform all children into root tile space. This works perfectly fine when the root of the tileset has a world transform that all tiles use. In cases where a tileset has an identity root transform and each child tile has its own world transform, the positions, scales and rotations are being transformed into world space on the cpu rather than the gpu. Our rendering code assumes incoming data is in model space and not world space. This causes precision issues which lead to misshapen and misoriented splats causing low fidelity, blurriness, gaps, etc.

### Reproduction steps

1.
2.
3.
...

### Environment

Browser:
CesiumJS Version:
Operating System:

PR #12926, opened 2025-09-25
Gsplat fixedframe transform

keyboardspecialist:
# Description

Fixes issue with tilesets that have an identity root transform and child tiles with their own world transform. This uses the tileset bounding sphere center point transformed using `Transforms.eastNorthUpToFixedFrame`.

Previously, because of the identity, it led to tiles transforming themselves into world space prematurely leading to low quality output due to precision loss.

## Issue number and link

#12925 

## Testing plan

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 - Improved performance and reduced memory usage of `Event` class. [#12896](https://github.com/CesiumGS/cesium/pull/12896)
 - Fixes vertical misalignment of glyphs in labels with small fonts [#8474](https://github.com/CesiumGS/cesium/issues/8474)
 - Prevent runtime errors for certain forms of invalid PNTS files [#12872](https://github.com/CesiumGS/cesium/issues/12872)
+- Fixes precision issues with Gaussian splat tilesets where tiles have a world transform and the tileset does not [#12925](https://github.com/CesiumGS/cesium/issues/12925)
 
 #### Additions :tada:
 
--- a/packages/engine/Source/Scene/GaussianSplatPrimitive.js
+++ b/packages/engine/Source/Scene/GaussianSplatPrimitive.js
@@ -32,6 +32,7 @@ import Quaternion from "../Core/Quaternion.js";
 import SplitDirection from "./SplitDirection.js";
 import destroyObject from "../Core/destroyObject.js";
 import ContextLimits from "../Renderer/ContextLimits.js";
+import Transforms from "../Core/Transforms.js";
 
 const scratchMatrix4A = new Matrix4();
 const scratchMatrix4B = new Matrix4();
@@ -427,6 +428,13 @@ GaussianSplatPrimitive.transformTile = function (tile) {
   const gltfPrimitive = tile.content.gltfPrimitive;
   const gaussianSplatPrimitive = tile.tileset.gaussianSplatPrimitive;
 
+  if (gaussianSplatPrimitive._rootTransform === undefined) {
+    gaussianSplatPrimitive._rootTransform = Transforms.eastNorthUpToFixedFrame(
+      tile.tileset.boundingSphere.center,
+    );
+  }
+  const rootTransform = gaussianSplatPrimitive._rootTransform;
+
   const computedModelMatrix = Matrix4.multiplyTransformation(
     computedTransform,
     gaussianSplatPrimitive._axisCorrectionMatrix,
@@ -441,7 +449,7 @@ GaussianSplatPrimitive.transformTile = function (tile) {
 
   const toGlobal = Matrix4.multiply(
     tile.tileset.modelMatrix,
-    Matrix4.fromArray(tile.tileset.root.transform),
+    Matrix4.fromArray(rootTransform),
     scratchMatrix4B,
   );
   const toLocal = Matrix4.inverse(toGlobal, scratchMatrix4C);
@@ -765,7 +773,7 @@ GaussianSplatPrimitive.buildGSplatDrawCommand = function (
 
   const modelMatrix = Matrix4.multiply(
     tileset.modelMatrix,
-    Matrix4.fromArray(tileset.root.transform),
+    Matrix4.fromArray(primitive._rootTransform),
     scratchMatrix4B,
   );
 


mzschwartz5:
1. Is there any sandcastle or anything I can use to verify the fix?
2. Reading through the issue, I'm a bit confused on why this set up causes splats to be transformed on the CPU rather than the GPU. Can you elaborate?
3. Did the issue also happen if the root tile has a non-identity transform and the child tiles have their own world transforms? Or does the root tile _have_ to have an identity transform.

javagl:
The issue should contain a test data set. The `_rootTransform` should probably no longer be called `_rootTransform`, because it isn't necessarily the "root transform". Although... it wasn't the root transform to begin with - it was the "computed transform", which includes the model matrix. In fact, the different configurations of where a certain transform takes place (in the data itself, in the root tile transform, in the child tile transform, in a glTF node matrix) would warrant ~5 (or maybe 5x5) test data sets. But creating test data where the transform happens in another glTF node than 'node 0' is moot, because this will [break anyhow](https://github.com/CesiumGS/cesium/blob/1dcf97348f468bd0998f0a6a6679c24f4af5d2bd/packages/engine/Source/Scene/GaussianSplat3DTileContent.js#L674).

However, here's a test data set that probably demonstrates what the issue was probably about: With the transform in the root tile, it looks fine. With the transform in the child tile, it looks wrong and wiggly

Cesium Splats PR 12926

The tilesets with a sandcastle:

[Cesium Splat Tileset Transform PR 12926.zip](https://github.com/user-attachments/files/22890053/Cesium.Splat.Tileset.Transform.PR.12926.zip)

(It still uses the "old" extension format. I've [prepared](https://github.com/javagl/JSpz/blob/afab5b2cfda402a97f7644dba9bdba44b8089335/jspz-examples/src/main/java/de/javagl/jspz/examples/SpzToTileset.java#L90) supporting the new extension format, but ... this was before the new extension format was ""final"", so it will have to be straightened out and reviewed...)

CHANGES.md
```diff
@@ -12,6 +12,7 @@
 - Improved performance and reduced memory usage of `Event` class. [#12896](https://github.com/CesiumGS/cesium/pull/12896)
 - Fixes vertical misalignment of glyphs in labels with small fonts [#8474](https://github.com/CesiumGS/cesium/issues/8474)
 - Prevent runtime errors for certain forms of invalid PNTS files [#12872](https://github.com/CesiumGS/cesium/issues/12872)
+- Fixes precision issues with Gaussian splat tilesets where tiles have a world transform and the tileset does not [#12925](https://github.com/CesiumGS/cesium/issues/12925)
```
lilleyse: @keyboardspecialist could you move this to the November section and perhaps reword a bit

```suggestion
- Fixed precision issues with Gaussian splat tilesets where the root tile does not have a world transform. [#12925](https://github.com/CesiumGS/cesium/issues/12925)
```

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -427,6 +428,13 @@ GaussianSplatPrimitive.transformTile = function (tile) {
   const gltfPrimitive = tile.content.gltfPrimitive;
   const gaussianSplatPrimitive = tile.tileset.gaussianSplatPrimitive;
 
+  if (gaussianSplatPrimitive._rootTransform === undefined) {
+    gaussianSplatPrimitive._rootTransform = Transforms.eastNorthUpToFixedFrame(
+      tile.tileset.boundingSphere.center,
+    );
+  }
+  const rootTransform = gaussianSplatPrimitive._rootTransform;
```
lilleyse, line 436: `GaussianSplatPrimitive.transformTile` and `GaussianSplatPrimitive.prototype.update` are both trying to set `this._rootTransform` but it's not clear to me when one would be called first. Is it possible to centralize this a bit more so that `_rootTransform` is only set in one place?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -765,7 +773,7 @@ GaussianSplatPrimitive.buildGSplatDrawCommand = function (
 
   const modelMatrix = Matrix4.multiply(
     tileset.modelMatrix,
-    Matrix4.fromArray(tileset.root.transform),
+    Matrix4.fromArray(primitive._rootTransform),
```
lilleyse: Does this need to be `Matrix4.fromArray` any more?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -441,7 +449,7 @@ GaussianSplatPrimitive.transformTile = function (tile) {
 
   const toGlobal = Matrix4.multiply(
     tile.tileset.modelMatrix,
-    Matrix4.fromArray(tile.tileset.root.transform),
+    Matrix4.fromArray(rootTransform),
```
lilleyse: Same comment

packages/engine/Source/Scene/GaussianSplatPrimitive.js
keyboardspecialist: transformTile should be the one to set it (and does). The one in update should be removed.

lilleyse: Sounds good. So the local reference frame will always be computed from the bounding sphere center? I think that makes sense.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
keyboardspecialist: We do not

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -684,7 +692,7 @@ GaussianSplatPrimitive.buildGSplatDrawCommand = function (
     const tileset = primitive._tileset;
     const modelMatrix = Matrix4.multiply(
       tileset.modelMatrix,
-      Matrix4.fromArray(tileset.root.transform),
+      tileset.root.transform,
```
lilleyse: Should this be `primitive._rootTransform`?

CHANGES.md
```diff
@@ -23,6 +23,7 @@
 - Improved performance of terrain Quadtree handling of custom data [#12907](https://github.com/CesiumGS/cesium/pull/12907)
 - Fixed picking of `GroundPrimitive` with multiple `PolygonGeometry` instances selecting the wrong instance. [#12978](https://github.com/CesiumGS/cesium/pull/12978)
 - Fixed a bug where the removal of draped imagery layers did not update the rendered state [#12923](https://github.com/CesiumGS/cesium/issues/12923)
+- Fixes precision issues with Gaussian splat tilesets where tiles have a world transform and the tileset does not [#12925](https://github.com/CesiumGS/cesium/issues/12925)
```
lilleyse: In case you missed it, I also had a rewording suggestion: https://github.com/CesiumGS/cesium/pull/12926#discussion_r2475287786

lilleyse:
Thanks @keyboardspecialist. Once those last comments are addressed this should be good to go. I confirmed that this fixes the tilesets that were having the precision issues.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
keyboardspecialist: Yes it should

lilleyse (approved):

Issue #12682, opened 2025-06-21
Follow-ups for Gaussian Splat SPZ support


javagl:
I mentioned that in a [comment in the Gaussian Splat SPZ Support PR](https://github.com/CesiumGS/cesium/pull/12582#issuecomment-2988830633), but it may be worth tracking this here:

---

There seem to be some assumptions about the structure of the glTF of which I'm not sure whether they will always hold: https://github.com/CesiumGS/cesium/blob/8a362a0aa2e038cde8e0a50fd83d60e20b1eb9c1/packages/engine/Source/Scene/GaussianSplat3DTileContent.js#L381 
There is not necessarily a `matrix` property in that node, The first primitive may not be a Gaussian Splat primitive (or the Gaussian Splat primitive may not be the first one). There may also be _multiple_ Gaussian Splat primitives in one glTF.

---

Modifying the tileset `modelMatrix` does not seem to have an effect on the gaussian splat primitive. (Or at lest not the "right" effect: It seems to affect the position and bounding volumes, but not the orientation).

---

Speaking of orientations: People might say that I'm obsessed with unit cubes. And I wouldn't even concur 😁 : I created a "Unit Cube Splat File". (It's rather a 10x10x10 cube, but unit-y in other ways). And when rendering this in BabylonJS, it looks like this:

![Image](https://github.com/user-attachments/assets/396f0e8c-2aab-4d1d-8117-4e8943f38c0f)

Exactly as expected. With the colors and such. 🔴 🟢 🔵 

When putting this into a glTF and a tileset and rendering it with CesiumJS, it looks like this:

![Image](https://github.com/user-attachments/assets/fc0e010a-da44-4f7d-8010-9e95836a26ca)

There may be a million reasons why something is "wrong". For example, the bounding box. This is coming from the tileset JSON, and may therefore be "my fault". The splat is a 10x10x10 cube, and one might expect that the bounding volume has to be
`[5.0, 5.0, 5.0, 5.0, 0.0, 0.0, 0.0, 5.0, 0.0, 0.0, 0.0, 5.0]`
But that seems to be wrong. The one from the screenshot takes the RUB->LUF coordinates of the SPZ-to-glTF into account, and is
`[-5.0, -5.0, 5.0, 5.0, 0.0, 0.0, 0.0, 5.0, 0.0, 0.0, 0.0, 5.0]`
[...]

PR #12927, opened 2025-09-25
Gsplat better tile selection

keyboardspecialist:
# Description

Improves tile selection change detection by checking if tiles were selected previous frame. Allows for LoD to load and render correctly.

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/GaussianSplatPrimitive.js
+++ b/packages/engine/Source/Scene/GaussianSplatPrimitive.js
@@ -838,10 +838,24 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
       return;
     }
 
-    if (
-      tileset._selectedTiles.length !== 0 &&
-      tileset._selectedTiles.length !== this.selectedTileLength
-    ) {
+    const selectionChanged = () => {
+      const currentSelection = tileset._selectedTiles;
+      if (this._previousSelectionLength !== currentSelection.length) {
+        this._previousSelectionLength = currentSelection.length;
+        return true;
+      }
+      for (let i = 0; i < currentSelection.length; i++) {
+        if (!currentSelection[i]._wasSelectedLastFrame) {
+          this._previousSelectionLength = currentSelection.length;
+          return true;
+        }
+      }
+
+      this._previousSelectionLength = currentSelection.length;
+      return false;
+    };
+
+    if (selectionChanged()) {
       this._numSplats = 0;
       this._positions = undefined;
       this._rotations = undefined;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 - Improved performance and reduced memory usage of `Event` class. [#12896](https://github.com/CesiumGS/cesium/pull/12896)
 - Fixes vertical misalignment of glyphs in labels with small fonts [#8474](https://github.com/CesiumGS/cesium/issues/8474)
 - Prevent runtime errors for certain forms of invalid PNTS files [#12872](https://github.com/CesiumGS/cesium/issues/12872)
+- Improved tile selection check for Gaussian splats. Allows proper LoD rendering.
 
 #### Additions :tada:
 


packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -838,10 +838,24 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
...
+      for (let i = 0; i < currentSelection.length; i++) {
+        if (!currentSelection[i]._wasSelectedLastFrame) {
+          this._previousSelectionLength = currentSelection.length;
+          return true;
+        }
+      }
+
+      this._previousSelectionLength = currentSelection.length;
+      return false;
+    };
```
mzschwartz5, line 856: nit - the `update` function is pretty long as-is, why not just extract this into a proper function instead of a lambda?

mzschwartz5: Other comments:
1. Every branch of this function sets `this._previousSelectionLength = currentSelection.length;`, so rather than doing it 3x we do it just once, at the top.
2. "selection" refers to a 3D tile, right? This function seems like it belongs to the tileset class, not to this class. The logic itself isn't specific to splats. It also seems like, if it were done in the tileset class, there may be a more performant way to do it that doesn't involve iterating over the whole surrent selection. 

(Note that 2 is mutually exclusive with my other comment above; we can only do one or the other)

javagl: The last `this._previousSelectionLength = currentSelection.length;` is unnecessary anyhow, because when they are not (already) equal, the first `if` has already kicked in. 

All this should refer to selected tiles that have _splat_ content - but this is not handled, [see other issues](https://github.com/CesiumGS/cesium/issues/12682#issuecomment-3085360256).

mzschwartz5: Good point- actually, by the same logic, the second `this._previousSelectionLength = currentSelection.length;` is also unnecessary.

ggetz:
@mzschwartz5 or @keyboardspecialist, is this PR mergable with some cleanup? Or is there enough work to get this over the finish line that we should close for the time being?

mzschwartz5:
@ggetz The changes I requested are simple enough that I could do them.

However, since there's no linked issue or testing steps, I'm not sure how to verify the changes or even what the problem we're targeting _is_. 

@javagl Do you know what this is trying to fix?

javagl:
Maybe @keyboardspecialist wants to chime in here, and

- Open an issue with a clear description of the problem and preferably test data
- Add a link to the issue here
- Add a description about why and how this change fixes the issue
- Fill out the "testing plan"

---

What I can "reverse engineer" here: 

- **Before** this change:
  - When 10 tiles from a tileset had been "selected" during an `update` call, and in the next frame, 10 **different** tiles are selected, it did **not** trigger a recomputation of the splat data structures, because the **number** of selected tiles did not change.
- **After** this change:
  - When 10 tiles from a tileset had been "selected" during an `update` call, and in the next frame, 10 **different** tiles are selected, then it **will** trigger a recomputation of the splat data structures, because for some of the new tiles, `_wasSelectedLastFrame` is probably supposed to be `false`


---

Details (feel free to skip) : 

Originally, that `if`-block (with 100-lines and two local lambdas) was executed when the length of the `_selectedTiles` in the tileset changed between two calls to the `update` function. 

---

Interlude:

What is `_selectedTiles`? In any case, it is a **private** variable, and should preferably not be accessed without a very strong reason. For completeness, I'll quote the [full documentation of this property](https://github.com/CesiumGS/cesium/blob/b5bbf6f13eec9879af2379952ac587ab29bbb8a9/packages/engine/Source/Scene/Cesium3DTileset.js#L220) here:
```

```
That was quick. I actually did [add some comment](https://github.com/CesiumGS/cesium/blob/41e39a48dbebc9b8f57c0c8e9c00ab538ad810dd/packages/engine/Source/Scene/Cesium3DTileset.js#L223) as part of an unrelated PR, but that's still only the result of reverse engineering....

---

The core of the change here is that the condition for executing the `if` block now involves this part:
```
for (let i = 0; i < currentSelection.length; i++) {
  if (!currentSelection[i]._wasSelectedLastFrame) {
    ...
    return true;      
  }
}
```
As mentioned in the inlined comments: There are other changes, and some unnecessary code, but maybe that will still be cleaned up.

---

Interlude:

What is `_wasSelectedLastFrame`? In any case, it is a **private** variable, and should preferably not be accessed without a very strong reason. For completeness, I'll quote the [full documentation of this property](https://github.com/CesiumGS/cesium/blob/b5bbf6f13eec9879af2379952ac587ab29bbb8a9/packages/engine/Source/Scene/Cesium3DTile.js#L504) here:
```

```
That was quick. 

It's likely a flag that indicates that the tile was in the `_selectedTiles` in the last frame, but I haven't checked that this is really the case, or where it is read or written.

---

So the main change is that the update is triggered based on whether the _actual_ selected tiles changed, and no longer based on whether the _number_ of selected tiles changed.

---


I think that there are additional pending bugs in that code. The subsequent code casually accesses
`const shData = tile.content.packedSphericalHarmonicsData;`
and I don't see a point where it makes sure that the selected tiles all contain splat data. But some of that is already mentioned in https://github.com/CesiumGS/cesium/issues/12682#issuecomment-3085360256

javagl:
With all the other changes to Gaussian splats that have happened recently, I'll go out on a limb and close this one. Feel free to reopen if necessary.

PR #12947, opened 2025-10-02
Fix Sandcastle urls with `code` search param

### LINKED ISSUE #12946: Linking to Showcases forces Hello World example
### What happened?

Ref: #12857 

In ION if I click on "Open complete code example" the link opens with the Hello World example instead of the intended showcase. Before the recent update to the "Sandcastle" this would open a demonstration of the asset.

This is what the link looks like -> https://sandcastle.cesium.com/?src=Hello%20World.html&label=Showcases&code=redacted
Note: I redacted my token there at the end of the URL

Image

### Reproduction steps

1. Go to Cesium ION
2. Choose an Asset from the 'My Assets' tab
3. Click the link for "Open complete code example" in the information tab in the right hand pane.
...

### Sandcastle example

_No response_

### Environment

Browser: 
CesiumJS Version:
Operating System:

jjspace:
# Description

The code examples that open from ion use a `code` parameter to encode the sandcastle code instead of the `#c=` like our share urls. I missed accounting for this previously so they're all broken now that sandcastle is fully live.

This PR updates the loading logic for these urls. I think we may want to update the urls directly in ion still but there's a chance people have them saved so we should leave the logic in sandcastle as well.

## Issue number and link

Fixes #12946 

## Testing plan

- log into ion and click on any asset in the list
- Copy the link for the "Open complete code example" link on the right to open sandcastle
    - Most browsers should let you right click it and pick "Copy link address" or similar. Clicking it will redirect you so the url disappears
- Replace the `sandcastle.cesium.com` with the local url for sandcastle if running locally or `https://ci-builds.cesium.com/cesium/fix-sandcastle-code-url/Apps/Sandcastle2/index.html` to load it in CI for this branch
- Ensure the code and viewer loads showing the asset you picked

# Author checklist

### DIFF
--- a/packages/sandcastle/src/Gallery/loadFromUrl.ts
+++ b/packages/sandcastle/src/Gallery/loadFromUrl.ts
@@ -43,6 +43,20 @@ export function loadFromUrl(
 ) {
   const searchParams = new URLSearchParams(window.location.search);
 
+  const codeParam = searchParams.get("code");
+  if (codeParam) {
+    // This is a legacy support type url that was used by ion.
+    // Ideally we use the #c= param as that results in shorter urls
+    // The code query parameter is a Base64 encoded JSON string with `code` and `html` properties.
+    const json = JSON.parse(window.atob(codeParam.replaceAll(" ", "+")));
+
+    return {
+      title: "New Sandcastle",
+      code: json.code,
+      html: json.html,
+    };
+  }
+
   if (window.location.hash.indexOf("#c=") === 0) {
     const base64String = window.location.hash.substr(3);
     const { code, html } = decodeBase64Data(base64String);


packages/sandcastle/src/Gallery/loadFromUrl.ts
```diff
@@ -43,6 +43,20 @@ export function loadFromUrl(
 ) {
   const searchParams = new URLSearchParams(window.location.search);
 
+  const codeParam = searchParams.get("code");
```
jjspace, line 46: The previous code for the url loading logic can be found starting here https://github.com/CesiumGS/cesium/blob/8371941064a32e56fc20306a590d19c8ccbc40b6/Apps/Sandcastle/CesiumSandcastle.js#L819

I believe this `loadFromUrl` function now covers all the conditional paths from that function but might be worth someone else double checking me.

packages/sandcastle/src/Gallery/loadFromUrl.ts
```diff
@@ -43,6 +43,20 @@ export function loadFromUrl(
...
+  if (codeParam) {
+    // This is a legacy support type url that was used by ion.
+    // Ideally we use the #c= param as that results in shorter urls
+    // The code query parameter is a Base64 encoded JSON string with `code` and `html` properties.
+    const json = JSON.parse(window.atob(codeParam.replaceAll(" ", "+")));
```
ggetz, line 51: Do you know what the purpose of the `.replaceAll(" ", "+"))` is? Should it cover all whitespace with `/\s/g` instead?

ggetz (approved):
I tested this out and it appears to be working well. I just have the one question before merging.

packages/sandcastle/src/Gallery/loadFromUrl.ts
jjspace: This was needed or the `atob` would fail or spit out a garbled json string. The old sandcastle uses a dojo function/library to parse the query param that seems to behave differently than the `URLSearchParams` parsing. I believe it was extracting the pure text whereas `URLSearchParams` utilizes [url encoding](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#percent_encoding) and treats [`+` as a space](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#preserving_plus_signs). I think this replacement is good for our use case since I think it's safe to assume these are coming from ion and should have a consistent format.

ggetz:
Great, thanks @jjspace! Please go ahead and deploy.

jjspace:
This has been merged to `cesium.com` and deployed successfully. ion links now work again

PR #12955, opened 2025-10-06
Fixes event bug with changes to listeners during callback

mzschwartz5:
# Description

Originally discovered in https://github.com/CesiumGS/cesium-analytics/issues/1077, but since that's a private repo, here's the gist from my comment on that issue:

>  The issue here occurs when a caller adds a new event listener from inside an event callback. In #12896, Event switched from an array to a map - the iteration over the map uses an iterator, rather than a length-based loop. This means we're now modifying a collection that we're iterating over - that's a no no.

This can lead to an infinite loop if an event callback removes and re-adds a listener to that event (that new listener gets iterated over immediately and adds and removes a listener, which gets iterated over immediately... and so on).

The solution is to defer addition of new listeners until after the event has fired - like how we defer removal of listeners. It wasn't previously necessary to defer additions, when we were using an array. Now it is.

I also took the liberty of refactoring the `toRemove` array into a map, which is both consistent with the rest of the class now (and simplifies the code), and also reduces memory (no need to store duplicate listeners for unique scopes).

## Issue number and link

Fixes the above issue in the cesium analytics repo.

Testing:
- Run `Event` unit tests
- Run `ClippingPlanesEditor` unit tests in cesium-analytics
- Confirm you can switch from "3D Tiles" to "Model" in the [ClippingPlanesEditor sandcastle.](http://localhost:8080/Apps/Sandcastle/index.html?src=Clipping%20Planes%20Editor.html). (Prior to this fix, it would hang or crash).
- Try your favorite sandcastle that uses events (I just searched for "event" in the gallery).

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/Event.js
+++ b/packages/engine/Source/Core/Event.js
@@ -27,8 +27,17 @@ function Event() {
    * @private
    */
   this._listeners = new Map();
-  this._toRemove = [];
-  this._insideRaiseEvent = false;
+  /**
+   * @type {Map<Listener,Set<object>>}
+   * @private
+   */
+  this._toRemove = new Map();
+  /**
+   * @type {Map<Listener,Set<object>>}
+   * @private
+   */
+  this._toAdd = new Map();
+  this._invokingListeners = false;
   this._listenerCount = 0; // Tracks number of listener + scope pairs
 }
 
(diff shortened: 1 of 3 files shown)


packages/engine/Source/Core/Event.js
```diff
@@ -27,8 +27,17 @@ function Event() {
...
+   * @type {Map<Listener,Set<object>>}
+   * @private
+   */
+  this._toRemove = new Map();
+  /**
+   * @type {Map<Listener,Set<object>>}
+   * @private
+   */
+  this._toAdd = new Map();
+  this._invokingListeners = false;
```
mzschwartz5, line 40: I renamed this because a portion of the code in `raiseEvent` now runs _after_ this flag is set back to false, so it would have been a bit of a misnomer.

packages/engine/Specs/Core/EventSpec.js
```diff
@@ -91,6 +91,37 @@ describe("Core/Event", function () {
...
+      event.addEventListener(function () {});
+    };
+
+    event.addEventListener(addEvent0);
+    event.addEventListener(addEvent1);
+    expect(event.numberOfListeners).toEqual(2);
+    event.raiseEvent();
+    expect(event.numberOfListeners).toEqual(4);
+  });
+
```
mzschwartz5, line 124: These unit tests should have been here in the first place. Their counterparts were ("can remove..."), and had these tests been in place, the issue would have been caught in the first place!

PR #12957, opened 2025-10-07
Small gulpfile updates

jjspace:
# Description

A few small tweaks/cleanup items to the gulpfile that I noticed while working on #12904 

I specifically never realized there were links added to the github status until @ggetz clicked one the other day. This hopefully makes it more clear. I also prefixed them all the same so they should now show up next to each other alphabetically unlike before.

image

## Issue number and link

no issue

## Testing plan

- Make sure `npm run website-release` works as expected
- Make sure the github status on the latest commit has the updated deploy artifacts messages

# Author checklist

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -425,17 +425,6 @@ export async function buildDocsWatch() {
   return gulp.watch(sourceFiles, buildDocs);
 }
 
-function combineForSandcastle() {
-  const outputDirectory = join("Build", "Sandcastle", "CesiumUnminified");
-  return buildCesium({
-    development: false,
-    minify: false,
-    removePragmas: false,
-    node: false,
-    outputDirectory: outputDirectory,
-  });
-}
-
 export const websiteRelease = gulp.series(
   buildEngine,
   buildWidgets,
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -712,17 +710,21 @@ export async function deploySetVersion() {
 export async function deployStatus() {
   const status = argv.status;
   const message = argv.message;
-
   const deployUrl = `${devDeployUrl}`;
   const zipUrl = `${deployUrl}Cesium-${version}.zip`;
   const npmUrl = `${deployUrl}cesium-${version}.tgz`;
-  const coverageUrl = `${devDeployUrl}Build/Coverage/index.html`;
+  const coverageUrl = `${deployUrl}Build/Coverage/index.html`;
```
ggetz, line 716: I believe this link is already broken in `main`, but it does not appear to be fixed with this change;  https://ci-builds.cesium.com/cesium/small-gulpfile-updates/Build/Coverage/index.html is still a 404 equivalent.

Was this update attempting to address the broken link, or was it targeted at something else?

jjspace: @ggetz this was simply because I noticed these 3 urls were using different variables and probably should be consistent (functionally it's exactly the same). I did not actually check the link itself.

ggetz: Got it. I just checked and it appears to be an issue with the GitHub Actions workflow, not the gulpfile. Out of scope for this PR.

Issue #12585, opened 2025-04-27
imageSubRegion not display


mivui:
https://sandcastle.cesium.com/?src=Billboards.html&label=All

Add maker billboards

ggetz:
Thanks @mivui, I can also reproduce the issue.

mivui:
cesium 1.126 is correct

plainheart:
Hi, can this issue be fixed in the next version? It is impossible to use sprites on the map for now.

mivui:
cesium@1.126.0 works properly.

plainheart:
Unfortunately, this bug persists from Cesium 1.127 to 1.132.

plainheart:
@ggetz @mzschwartz5 @javagl Hello, could you please take a look at this bug that has been persisting from version 1.127 through to the latest version 1.134? It's been blocking our progress for months, forcing us to remain on v1.126. We'd really appreciate your help resolving this. 🙏

[...]

PR #12958, opened 2025-10-08
Add billboard subregion condition

Beilinson:
# Description

As per #12585, image subregions currently do not work since 1.127. The cause was a conditionless `addImageSubregion` call on the `BillboardVisualizer` every frame prior to rendering, which resulted in a large promise queue but also made sure that `image.state` was `LOADING` for every render.

This PR adds a comparison against the previous loading subRegion, which skips adding a new subregion if not needed, letting the first call resolve properly and set the state to `LOADED`.

## Issue number and link

#12585

## Testing plan

Sandboxes from #12585

# Author checklist

### DIFF
--- a/packages/engine/Source/DataSources/BillboardVisualizer.js
+++ b/packages/engine/Source/DataSources/BillboardVisualizer.js
@@ -232,8 +232,12 @@ BillboardVisualizer.prototype.update = function (time) {
       time,
       boundingRectangleScratch,
     );
-    if (defined(subRegion)) {
+    if (
+      defined(subRegion) &&
+      !BoundingRectangle.equals(subRegion, item.subRegion)
+    ) {
       billboard.setImageSubRegion(billboard.image, subRegion);
+      item.subRegion = BoundingRectangle.clone(subRegion, item.subRegion);
     }
   }
   return true;
--- a/packages/engine/Source/Scene/Billboard.js
+++ b/packages/engine/Source/Scene/Billboard.js
@@ -189,8 +189,6 @@ function Billboard(options, billboardCollection) {
   this._batchIndex = undefined; // Used only by Vector3DTilePoints and BillboardCollection
 
   this._imageTexture = new BillboardTexture(billboardCollection);
-  this._imageWidth = undefined;
-  this._imageHeight = undefined;
 
   this._labelDimensions = undefined;
   this._labelHorizontalOrigin = undefined;


packages/engine/Source/DataSources/BillboardVisualizer.js
```diff
@@ -232,8 +232,12 @@ BillboardVisualizer.prototype.update = function (time) {
       time,
       boundingRectangleScratch,
     );
-    if (defined(subRegion)) {
+    if (
+      defined(subRegion) &&
+      !BoundingRectangle.equals(subRegion, item.subRegion)
+    ) {
       billboard.setImageSubRegion(billboard.image, subRegion);
+      item.subRegion = BoundingRectangle.clone(subRegion, item.subRegion);
```
Beilinson: this basically doubles all `BoundingRectangle` objects added to the scene. I tried making this test against the internal width/height, but those are only set after the `subRegion` promise resolves so that wont work.

packages/engine/Source/Scene/Billboard.js
```diff
@@ -189,8 +189,6 @@ function Billboard(options, billboardCollection) {
   this._batchIndex = undefined; // Used only by Vector3DTilePoints and BillboardCollection
 
   this._imageTexture = new BillboardTexture(billboardCollection);
-  this._imageWidth = undefined;
-  this._imageHeight = undefined;
```
Beilinson, line 193: Deleted because these are unused anywhere else in the code

packages/engine/Source/DataSources/BillboardVisualizer.js
mzschwartz5: I notice that a billboard's imageSubRegion is created as a [constant property](https://github.com/CesiumGS/cesium/blob/074acf176a4add690a0f1a098fb2e7f587b91d2d/packages/engine/Source/DataSources/BillboardGraphics.js#L316) (based on the lack of the `createPropertyCallback` arg passed in to [`createPropertyDescriptor`](https://github.com/CesiumGS/cesium/blob/074acf176a4add690a0f1a098fb2e7f587b91d2d/packages/engine/Source/DataSources/createPropertyDescriptor.js#L60)).

Since the property never changes, maybe the right move here is to remove this code from `update` entirely and find another home for it? (Like a constructor or some setter - something that's only called once, not every frame).

Beilinson: Interesting, I'll try to put something together tmrw

Beilinson: `createProperty` creates a setter that uses the [value directly](https://github.com/CesiumGS/cesium/blob/074acf176a4add690a0f1a098fb2e7f587b91d2d/packages/engine/Source/DataSources/createPropertyDescriptor.js#L25-L37) if it is a `Property`, so users can for example do:
```js
billboard.imageSubRegion = new CallbackProperty(...);
```
And then `createProperty` will set the internal value to actually be a `CallbackProperty`, so I dont think what you wrote can work here, correct me if I'm wrong.

mzschwartz5: Ooh I see what you're saying. So the the value of the property is constant, but the reference to a property can change, essentially.

I agree, that invalidates my idea. Follow up question, though: how does this "double all `BoundingRectangle` objects added to the scene`? Every time you clone, GC (eventually) deletes the old one.

Beilinson: Because I'm using `BoundingRectangle.clone` all subRegions exist twice: once on the `BillboardGraphics`, and once as a deep clone on the `item` in the `BillboardVisualizer`, to be used for equality comparison. The one on the `BillboardGraphics` isn't GC'ed since the reference to the graphics is never lost (until the graphics is removed of course)

I don't think this is a significant issue, just wanted to point that out for review in case there was a better solution

mzschwartz5: Going back to the original idea for a second, I do think [something like this](https://sandcastle.cesium.com/#c=rVXvb9owEP1XbnxZmDInHa3oKKCtbN86bSpTv4ypcpIjWDM2sh36S/3fd44JDW3XtVIlZOzk3rt7z/ZFLFfaOHgH3MIEraiWMDd6CbNOXq9mnaOZmqlcK+tgLfACDYxA4cUmmp3Vz6ImfqKV40KhmXW6Rw0uywgTwAyVE06gZbwoopuZAlhpS0+0GjSUE24czbjqMV/LFywNoo3e9w/Ywcd+vx/DfsrS3uFhrxt7gkxImWluigHUhABiyUsckAjGknpuk8B9fqJLfa7XaCS/YitVzjpxCzKtslMs61paEo91pQqhylPMHVelxCiNgX57aRhCFbc03nZbbumVV2X3SPsvH7CpzeGl86VNFsSFIS/YKjN14m09hERJCQcwr1TumSDqws3tNtcDws/WilLVhWebisE0Jf+PN7wkLzO2tZPterK77w9NaSiAvHl8/iH918r75v+fJ6/2lysHK6NXaNzVq6ubbDL82CRoiXvGweilYei+RJ9Fd8ZlhaQB8FJY5/fvfh2vIJQ1iV6maT8Nw1Oaftenfyd3gXOh6usdznvh7/3XNXWBE5KIyrcOQkc2J4ExKL6kce3Li0HLoi60C6Px9mbPIfJB8GY0ItN2pVHPoRPvKqOO7qJrMhhR+B3fTpA/TVoik7qkNnbclP7W3usJhOL1GcQiNLcATxKYCpUjXGArAtwC7+FJHfoXmsLovtvciAz90i2EBfSesED51O49w9CobdgL9AWn8kAbJNb7GyZ+g4PqKVdFzq0jTsr9U2uZcfMNVRU1Lc/HdeLO0LoriWOP/iTCh6YyMqKm7HC5kpy6fJJV+R90LLfWg4ZJAxkWYg2iGD3yYYFcks30Zl5JORXX1NzGw4Tid2BSc3+Ov4de70MWe+OT8JAxNkxo+RDlgpoW418) could work. E.g. subscribe to the definition change in the constructor.

But... it also means there's now an event listener listening to every property change on a billboard and filtering down to just `imageSubRegion` based on name. That would be in exchange for checking `BoundingRectangle` equality on every `Update` though, which maybe is worth it.

plainheart:
Could this fix be included in the upcoming v1.135?

packages/engine/Source/DataSources/BillboardVisualizer.js
mzschwartz5: I started wondering "why don't we just set the `subRegion` every frame and let the setter decide if it has changed" - [like this, for example](https://github.com/CesiumGS/cesium/blob/0ee5670050c7591748fcda95e6762b428750be50/packages/engine/Source/Scene/Billboard.js#L274-L288).

Then I realized, we don't have a proper setter here, because `Billboard.prototype.setImageSubRegion` takes two arguments - `id` and `subregion`. But then I followed the calls deeper, to `BillboardTexture.prototype.addImageSubRegion`, then to `TextureAtlas.prototype.addImageSubRegion`, and found this code:

https://github.com/CesiumGS/cesium/blob/0ee5670050c7591748fcda95e6762b428750be50/packages/engine/Source/Renderer/TextureAtlas.js#L701-L706

So ... it looks like there's already logic to say "hey, we're already tracking this subregion, don't re-add it". Can you figure out why that isn't working / sufficient?

mzschwartz5:
> Could this fix be included in the upcoming v1.135?

@plainheart. We can try. It's a simple PR, but I'm still not satisfied with our understanding and just asked for a little follow up on my recent findings.

The release is in 2 days, so I'll keep an eye on this.

packages/engine/Source/DataSources/BillboardVisualizer.js
Beilinson: >So ... it looks like there's already logic to say "hey, we're already tracking this subregion, don't re-add it". Can you figure out why that isn't working / sufficient?

Right, but this is happening after the promise for the index resolves, which only happens after the current frame finishes rendering, so each frame the image subregion is still set to "LOADING" and only resolves after the current frame, as I described in the PR description

This is also why this issue doesn't exist for manually managed primitives, the core issue is the billboard visualizer still treats this method as synchronous, when it is not since #12495.

Beilinson: Going back to the idea you mentioned, I think that something that I noticed lines up with what you mentioned on a larger scale;

Notice that all visualizers (I.e, the [PolylineVisualizer.prototype.update](https://github.com/CesiumGS/cesium/blob/0ee5670050c7591748fcda95e6762b428750be50/packages/engine/Source/DataSources/PolylineVisualizer.js#L198-L215)) in their `update` method only explicitly update the `changed` entities (which are managed through the whole `definitionChanged` events on entities and their graphics

However, the `BillboardVisualizer` does not do this, it updates all entities and their related primitives, ignoring the `changed, added, removed` arrays entirely.

Maybe what youre mentioning could be achieved by aligning the `BillboardVisualizer` (and possibly the `LabelVisualizer` and `PointVisualizer` similarly) to other visualizers which explicitly handle added/removed/changed graphics.

I will add a disclaimer here, I'm still pretty sure that this doesn't matter at all the moment the `BillboardGraphics` has a `CallbackProperty` assigned to the `imageSubRegion`, because that causes the `definitionChanged` event to always return true every frame IIRC. 

This is why a fix like the one in this PR is needed regardless. BTW, this is the same fix as applied by @ggetz in #12495 for the image property (I can't link the line change but see in the PR where the `texture` being stored on the `item` was added in the `BillboardVisualizer` update method.

Ok, after all this rambling I also noticed in the PR changes @12495 removed this exact same comparison that used to exist (except it was at the `Billboard` level, rather than at the `BillboardVisualizer` level). See :

image

Beilinson: TL;DR, a check like this was deleted from a previous PR and should be returned to fix this bug based on what I can tell.

I do think that its worth exploring explicitly handling `added/removed/changes` graphics similar to the other visualizers, but that is irrelevant to this PR as that still wouldn't work for `CallbackProperty` based `imageSubRegion`, although may bring possible performance improvements. I'm reminded of a blog I saw once: https://cesium.com/blog/2018/06/21/entity-api-performance/. It seems to me that the performance improvement suggested in this PR exists (more robustly) on all other visualizers except on the `Billboard/Label/Point` pipelines.

Beilinson: Actually for safety I'll add back in the check like it existed before, to account for the `image` changing while the `subRegion` stays the same (the PR doesn't handle this properly now)

packages/engine/Source/Scene/BillboardTexture.js
```diff
@@ -268,6 +268,11 @@ BillboardTexture.prototype.addImageSubRegion = async function (id, subRegion) {
     return;
   }
 
+  if (this._id !== id) {
+    // Another load was initiated and resolved resolved before this one. This operation is cancelled.
+    return;
+  }
+
```
Beilinson, line 313: This is a race condition that is handled by `loadImage` but wasn't handled by `addImageSubRegion` before

packages/engine/Source/DataSources/BillboardVisualizer.js
Beilinson: @mzschwartz5 this is ready for re-review

mzschwartz5: I need to look at this more tomorrow, but my initial reaction is that this still seems like a bandaid. We already _have_ state that tells us the subregion has been added already. We should find a way to use it.

Like (spitballing), maybe we can defer setting:
`  this._loadState = BillboardLoadState.LOADING;`

until we've called some function on `atlas` to check if the subregion already exists. Basically splitting up `TextureAtlas.prototype.addImageSubRegion` into `addImageSubRegion` and `checkSubRegionExists` or something...?

Beilinson: I think that might not be a bad idea however that still wouldn't fix the problem, the `checkSubRegionExists` method would still return a `promise` to an index which could still only resolve on the next frame, and if this check is initiated every frame it would never render properly. 

I've setup an alternative version of the code which now works how you suggested (split up syncronous check vs asyncronous addSubRegion), it seems to work as well, not sure how much I like this I think the previous implementation is clearer but its more or less the same thing just moving this check further up or down the code structure (actually this also changes some stuff with loadState like you suggested)

image 

Image from newest state showing that it still works

Beilinson: @mzschwartz5 this is again ready for re-review (will be available tmrw for any additional convo on this). Please compare the latest commit (using your suggestions) against the [older version](https://github.com/CesiumGS/cesium/pull/12958/files/5d00fca59617a53e72f92a268b638d18356c5253) (first 7 commits).

Beilinson:
[Sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html#c=zZZrc9o6EIb/yg6fzJQKm0uBFJjTAu045RIScoG6kxG2AIGRqCSHmk7++5EwFDr0tOlMmzkfBNJ639191rLlBpE0WiKXMxSQCY5C9cb3iZQDviAMauClSHw+G7/3aY+eu9cb1+lSV7rssug33FfuYnV30zivIO30OXi/ME7z7twtjgY3tBvb62Hu3aw96Nuj5rUaDtxNh9rrzm0r7mxGi+HtcK0HbTfOVyMdrDO/Lpr58K5Pe/NWvjtYbLpNP+42RxX0Je5d0KlDpx/e9R96A3w3GbTsD/1ScF2ax+zez/f6F7jvzFrU9VKvPeYxnzOp4IGSNREahJE1NBLYm63N8lL+dt3gTGHKiPBS6SOlIL7CbBoSqdUfPQbHId7yiAWUTS/3TlahkoFCPgNO2Yx05peCV04Gcr8lKGWgbGvnghlPEOS0wLF1ilzOjCcoHLtoJMXjqj5texISBVQ3wn6dzE1TVkRMuFhi5hPE+Noy3Ut65+MwHGN/8X3fGzvrheBaqmLLSkOtDl9NWbuea8Gh7x+pyQ1AJ2Cd5IKXuoY65Gw7nUSA/6zJXKMvXuxnOpomqdWgmP6GBPBofgRRkWAgtOUxAxMcSpJsCRwEHSwWRLylmoFjEUgruTKJmK8oZ/BDF/gKYAIn2xARpqiiRCLtbO2qXnFJTYCzQ5uE0jPM8mgi+LJJpoIQab0sFVGxUirpm1qwkZ0vl/PJHQUY7zOe7VuhMZd4Ss7084tQdjuX2ST+fZtP+T1/ICLEMVqxqZfahXnc/j8mXADZLLwJApBEu+LwkETCGEsSgGZWMwJ4rGMl6YDuTCrEEv1vuNczqsjVDK90Ace8O+eraHxJpttKjvae/embl89DLg5lmhVqu53Wadv+DHC5gPRjnq8g+zlAnRPQ40fWWCxdjY3MS8HU5RzK+uPkJdtkKDjPQ557ArnBtlHlOdjz27teeh72/M+392Wr+dc4K1vO4l/k3B8/P0Ucttrt3u0ppT4IUplUVao4JHVj/YcuV1woiERo6WoUWa5CrNGy48hfEIV8KY2umt1LqgF9ABrUfvCBAb5+M0p9ZRKF4RXdEC9Vr2a1/3eykGNzLveSd7RxmTn1dmJECFWzenmqUpxranEU8V8) with `CallbackProperty` for `imageSubRegion` showing that the new method works

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -697,28 +697,36 @@ TextureAtlas.prototype.addImageSubRegion = function (id, subRegion) {
     throw new RuntimeError(`image with id "${id}" not found in the atlas.`);
   }
 
-  const indexPromise = this._indexPromiseById.get(id);
   for (const [index, parentIndex] of this._subRegions.entries()) {
     if (imageIndex === parentIndex) {
       const boundingRegion = this._rectangles[index];
       if (boundingRegion.equals(subRegion)) {
         // The subregion is already being tracked
```
Beilinson, line 712: Do note that unlike my previous implementation, if there are many (thousands+) of subregions then this could get pretty slow, since this is checked by all billboards with subRegions every frame. This could be solved by creating a new `BidirectionalMap` core class (which would store an additional inverse `Map`), and using it instead of the regular map for this `_subRegions` map. I don't think this is needed for this bug fix, but is a possible future improvement if this implementation is chosen over my previous one.

packages/engine/Source/DataSources/BillboardVisualizer.js
Beilinson: I personally think the older version is better - smaller footprint, slightly more understandable process, slightly better  theoretical performance. The only real downside from my perspective is slightly worse theoretical memory usage, but its only reverting it to the same level that was before #12495

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -697,28 +697,36 @@ TextureAtlas.prototype.addImageSubRegion = function (id, subRegion) {
...
   }
+};
 
-  const index = this._nextIndex++;
+/**
+ * Add a sub-region of an existing atlas image as additional image indices.
+ * @private
+ * @param {string} id The identifier of the existing image.
+ * @param {BoundingRectangle} subRegion An {@link BoundingRectangle} defining a region of an existing image, measured in pixels from the bottom-left of the image.
+ * @returns {Promise<number>} A Promise that resolves to the image region index. -1 is returned if resouces are in the process of being destroyed.
```
mzschwartz5: This can also just return a number now, right? If the image subregion already exists, it just returns the index of the subregion directly.

packages/engine/Source/Scene/BillboardTexture.js
mzschwartz5: Can you elaborate? So this is a separate bug from what's originally being fixed in this PR?

packages/engine/Source/Scene/BillboardTexture.js
```diff
@@ -251,15 +251,33 @@ BillboardTexture.prototype.loadImage = async function (id, image) {
...
   this._hasSubregion = true;
 
+  const atlas = this._billboardCollection.textureAtlas;
+  const index = atlas.getImageSubRegion(id, subRegion);
+
+  if (index) {
+    this.setImageSubRegion(index, subRegion);
+  } else {
+    this.loadImageSubRegion(id, subRegion);
+  }
```
mzschwartz5:
```suggestion
  if (index) {
    this.setImageSubRegion(index, subRegion);
    return;
  } 
  
  this.loadImageSubRegion(id, subRegion);
```

Just a style preference

mzschwartz5 (commented):
Hey @Beilinson,

This is looking good - thanks for iterating on my feedback (I know it's more complex than the original fix but I think you've hit the root cause now rather than applying a bandaid on top).

I have a couple small comments/questions before approving, nothing big. Also- this fix may warrant some new unit test(s). It feels like something should have at least caught the fact that image subregions simply did not work at all.

packages/engine/Source/Scene/BillboardTexture.js
Beilinson: Yes, I caught this from just comparing the two functions it's unrelated to the original issue

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -671,54 +679,76 @@ TextureAtlas.prototype.addImage = function (id, image) {
       return -1;
     }
 
-    return this._addImage(index, image);
+    return this._addImage(index, image).then((index) => {
+      this._indexPromiseById.delete(id);
```
Beilinson: This change is necessary in order to realize in `addImageSubRegion` that the image has already been added and resolved.

I made sure to account for this change in the relevant places

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -671,54 +679,76 @@ TextureAtlas.prototype.addImage = function (id, image) {
...
+  imageIndex,
+) {
+  const imagePromise = this._indexPromiseById.get(id);
+  for (const [index, parentIndex] of this._subRegions.entries()) {
+    if (imageIndex === parentIndex) {
+      const boundingRegion = this._rectangles[index];
+      if (boundingRegion.equals(subRegion)) {
+        // The subregion is already being tracked
+        if (imagePromise) {
+          return imagePromise.then((resolvedImageIndex) =>
```
Beilinson, line 714: Unlike the previous commit, where the `subRegion` could often be cached and **synchronously** return an index before the image was fully loaded, here I make sure that the `subRegion` can never return a synchronous value until the underlying image is fully loaded and resolved.

Beilinson: My previous implementation didn't feel right since the `getImageSubRegion` could return a value even though the image hadn't been resolved and set to `LOADED` yet, which was complicated to track and understand in `BillboardTexture.addImageSubRegion`

packages/engine/Specs/DataSources/BillboardVisualizerSpec.js
```diff
@@ -178,9 +178,10 @@ describe(
       const bb = billboardCollection.get(0);
 
       await pollToPromise(function () {
+        entityCluster.update(scene.frameState);
         scene.renderForSpecs();
         visualizer.update(time);
-        return bb.show;
+        return bb.ready;
```
Beilinson, line 184: I was going to add a new unit test but I saw that this test which uses a `imageSubRegion` already existed and should have caught the bug earlier, it didn't because it was checking `bb.show` instead of `bb.ready`.

With this new test version and without the changes in the rest of this PR, this test fails with a timeout, since the billboard never becomes ready since every frame the visualizer updates it to `LOADING`, future-proofing this pr :)

Beilinson:
@mzschwartz5 BTW, this pr has some overlap with #12630, where I also had to do some sync/async fixes in `TextureAtlas` but for a slightly different issue

packages/engine/Source/Renderer/TextureAtlas.js
mzschwartz5: Interesting- this map was never cleaned up previously, was it? Seems like a good change.

Is there any possibility that this promise resolves and tries to delete the ID from the map before it's even set? (I'm not familiar with JS promises / the microtask queue / whatever, so maybe this is impossible). 

If that's a possibility, we could just tack on this `.then(...)` to the return statement of this function.

packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -653,12 +657,16 @@ TextureAtlas.prototype.addImage = function (id, image) {
...
 
   let promise = this._indexPromiseById.get(id);
+  let index = this._indexById.get(id);
   if (defined(promise)) {
-    // This image has already been added
+    // This image is already being added
     return promise;
+  } else if (defined(index)) {
+    // This image has already been added and resolved
+    return index;
```
mzschwartz5: Two things:
1. Style preference (a little personal): I prefer this to just be two separate if statements (rather than an else-if)
2. This function says it returns a Promise (or -1). So we should either change the jsdocs, or - my preference - wrap this returned index in a `Promise.resolve()`. To be honest, we should wrap the `-1` return value in `Promise.resolve()` as well, and the jsdocs should just say that we return a promise.

packages/engine/Specs/DataSources/BillboardVisualizerSpec.js
mzschwartz5: Oof - nice catch.

packages/engine/Source/Renderer/TextureAtlas.js
Beilinson: -1 is already in a promise its in the `resolveAndAddImage` method, the jsdoc should be updated that we return a promise resolving to -1. I'll wrap this case in `Promise.resolve` though its for the best.

packages/engine/Source/Renderer/TextureAtlas.js
Beilinson: It shouldn't be possible, all promises created on the current tick will resolve at best on the next tick, and on this tick the promise will safely be set. (not really ticks its the microtask queue but its the same answer regardless)

Beilinson: I'll move it out anyways though I think its more readable that we syntactically run `set` before `delete`

Beilinson: Ok so I spent the last couple commits trying to get this to work, I genuinely dont understand why tests kept failing no matter what I do this is either a jasmine specific issue or something even weirder, had to move the delete back inside the `resolveAndAddImage` because thats the only thing that worked for some reason.

mzschwartz5: That does make me suspicious lol.... but... I think I'll just ignore it this time. See you when the bug report comes in ;)

Issue #12811, opened 2025-08-12
Vertical exaggeration broken for Voxel datasets with ELLIPSOID shape


jjhembd:
### What happened?

With an ELLIPSOID-shaped voxel dataset, adjustments to `Scene.verticalExaggeration` result in the expected stretching of voxel data, but the height bounds of the shape do not change. This results in parts of the exaggerated data being clipped off.

https://github.com/user-attachments/assets/7cf2f98a-8b98-4029-b0b4-af03040b9380

### Reproduction steps

[...]

Issue #12938, opened 2025-10-01
Consolidate `VoxelBoundsCollection` and `ClippingPlaneCollection`


jjhembd:
`VoxelBoundsCollection` is very similar to `ClippingPlaneCollection`, with a few differences including:

- `VoxelBoundsCollection` assumes floating point textures are supported
- `ClippingPlaneCollection` stores the planes in the texture in model coordinates, and uploads a transform to be applied on the GPU. `VoxelBoundCollection` was reworked to store the planes in eye coordinates, for better precision when applied to global-scale models (see https://github.com/CesiumGS/cesium/pull/12933).

The amount of code duplication adds something of a maintenance burden. We should consider refactoring these as one class.

Issue #12813, opened 2025-08-12
Use local coordinates for voxel clipping


jjhembd:
[...]

PR #12959, opened 2025-10-08
Fix exaggeration of ellipsoid voxel clipping bounds

jjhembd:
# Description

This PR fixes exaggeration of ellipsoid voxels, by exaggerating the clipping bounds along with the shape bounds.

A `VoxelPrimitive` implements vertical exaggeration differently depending on the shape type:
- `VoxelShapeType.BOX` and `VoxelShapeType.CYLINDER` apply the exaggeration as a transform, which is multiplied with the `modelMatrix`. This approach assumes the exaggeration is along the same axis for every point in the voxel (e.g., `Z` for cylinders). This is a reasonable approximation for datasets only covering a small area of the Earth, with some assumptions (!) about the shape orientation.
- `VoxelShapeType.ELLIPSOID` is often used for datasets covering a large portion of the globe, so the exaggeration must be along the local ellipsoid normal, which varies across the dataset. This is implemented via scaling/shifting of the shape's height bounds.

The missing element for ellipsoids, highlighted in https://github.com/CesiumGS/cesium/issues/12811, was that clipping bounds were not exaggerated. Data near the clipping bound could therefore be clipped off when the shape was exaggerated. This PR corrects that by exaggerating the clipping bounds along with the shape bounds.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12811

## Testing plan

- Run `VoxelPrimitiveSpec` locally and verify all tests pass
- Load the [test local Sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html#c=1VZtb9s2EP4rhL9UBhzaQbEPc1xjeStmoGmCOe2XeVgp6WxzoUiBpBy7gf/7ji+SpTnJCgwdUMCwyOO9PPfweGSmpLFkw+ERNHlHJDySSzC8KuhnL0sWvczPL5W0jEvQi96APC0kISkz8IHtQI9rk1nBVqB3XkiXWhV3Wm14Dvrc7GSWOCNS695zATesnIPe8AyiZa3vjT9pEU0ao7TiIr9ReSXAr/bOjQFrhvewtZUGM/zI8MvENdN2PZstev1B8BC+4b/BfcezB4d+yYQBv7QClam8K2OSF8xyJdtCywsQyMZBtu+fLeRCZp7PJ2IykDAgGStAM7JHagPHqBQGFCGDzJOYWRB+VlsQM2lKyKzSN3zLpXPrndEc0mo1X6vH9xq9mjvQc8BwOfq2ugIfPmhuQFueIQ1btkJaPXrUOqWjsxqhi4dEgkC5301o6Y7Jy358+hoETjbwK/DV2r6m/ltH0/HkcS4rmXlUVZkzC22LpB8AvZrKx6pIsTibLGgbvqOMfDOm57x183P+9g513KsHqbIHVVlqNcseDmb9hl2rlEiZO1C5yqoCpKUrsNcC3PBiN8uxdKMO1ujZsWdWlmJ3wWXO5cocIgxqz85mqTRJQjyJBUG4POxq5JAvSSurNTO3jxLPWImk7BJn1I+KzQlrECDe29Tg6WSpgDYCb0ZNlZpM8xSS4x0M/O8jaQFhVhmrivma4em6VEL9o9lctpYTj2ip2cqxFWTugI/Jl43iebNyg+0oeR8nM1lWliyN/w6QC8yBZF+LPwuH+gYRas4EKeKg70I81d2lltKcL5eVAcQWPdECLMP8GGVUr9KzIwsmyjV7Xp8FGr4cmkNT9ZkGdHCnecFdmSVl7HtxN2KHCBVc1loGy7JQGzgXIgn+CImH2bWNxlvTa44dsDyPLbXd6DvWScNKDWrcjAb1Uns3x8d7G/X2setGrF2UVALDjm3nrCixk67aXaxhYNPph/RQyscZdwRnLR+hBdOl2N2rC1X5IzUv16CxqruI0s7qoK6PvKq74oiOfEb7OiUNeN3I4+C+8OdM5hkzVoDj/T4c3BuQVfJ7q/qsr+xF71oIXhpX3yfk7RVx16PBqzYo4S6DQA7GhLmblDSFlPQPVRxqod4r5IQ9Mm6bE+Y/b6+847jlL1y2BOFQOsSf3xy4wnoeduyH3kH4b4DXa9b9g6V/GSWbDEIZRKAvVX/U2AeSB8+wdKG2Pxo/CPn/YOZyh6fI5fWD0VPj/g4c/eFEvUFvYuxOwNSJf+FFqbQlFeJFgBYQIHozw7TCt6ClmTHOaDKsTSY53xCev3vmGUwywYzBlWUlxJx/hUVvOhmifsdMKOZ6yi0+RATbOZX16fRDEFJKJ0OcHls1jwOPemLdJTwN+U1sqvLdtKZjYvX0wPzE5tP2VTwZoqC7fJjhnPtr0+5KwKCayRUmQQoucXbqRmyLo59wZCyUOMQO6OTuhjtJcd9QtGGiwndw+wE2IF74yb8MxuSNj/ImJvN8bFfGLgzSGCP+S4yOt06aONGv0FO//kh43P0Hhk5OR6NRw9LPYRKJeoGl7sPye/HUjfJNTOGwKSscx3oLhenW/wY) and verify that the entire dataset is exaggerated without being truncated. Test all 3 shape types.
- In the same Sandcastle, use the "Voxel Inspector" to adjust the transform and clipping bounds, to verify that all shape bounds change as expected.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/VoxelPrimitive.js
+++ b/packages/engine/Source/Scene/VoxelPrimitive.js
@@ -173,38 +173,6 @@ function VoxelPrimitive(options) {
   this._maxBoundsOld = new Cartesian3();
 
   /**
-   * Minimum bounds with vertical exaggeration applied
-   *
-   * @type {Cartesian3}
-   * @private
-   */
-  this._exaggeratedMinBounds = new Cartesian3();
-
-  /**
-   * Used to detect if the shape is dirty.
-   *
-   * @type {Cartesian3}
-   * @private
-   */
-  this._exaggeratedMinBoundsOld = new Cartesian3();
-
-  /**
-   * Maximum bounds with vertical exaggeration applied
-   *
-   * @type {Cartesian3}
-   * @private
-   */
-  this._exaggeratedMaxBounds = new Cartesian3();
-
-  /**
-   * Used to detect if the shape is dirty.
-   *
-   * @type {Cartesian3}
-   * @private
-   */
-  this._exaggeratedMaxBoundsOld = new Cartesian3();
-
-  /**
    * This member is not known until the provider is ready.
    *
    * @type {Cartesian3}
@@ -239,38 +207,20 @@ function VoxelPrimitive(options) {
   this._maxClippingBoundsOld = new Cartesian3();
 
   /**
-   * Minimum clipping bounds with vertical exaggeration applied
+   * Vertical exaggeration applied to the voxel shape
    *
-   * @type {Cartesian3}
+   * @type {number}
    * @private
    */
-  this._exaggeratedMinClippingBounds = new Cartesian3();
+  this._verticalExaggeration = 1.0;
 
   /**
-   * Used to detect if the clipping is dirty.
-   * This member is not known until the provider is ready.
+   * The height relative to which the shape is exaggerated.
    *
-   * @type {Cartesian3}
+   * @type {number}
    * @private
    */
-  this._exaggeratedMinClippingBoundsOld = new Cartesian3();
-
-  /**
-   * Maximum clipping bounds with vertical exaggeration applied
-   *
-   * @type {Cartesian3}
-   * @private
-   */
-  this._exaggeratedMaxClippingBounds = new Cartesian3();
-
-  /**
-   * Used to detect if the clipping is dirty.
-   * This member is not known until the provider is ready.
-   *
-   * @type {Cartesian3}
-   * @private
-   */
-  this._exaggeratedMaxClippingBoundsOld = new Cartesian3();
+  this._verticalExaggerationRelativeHeight = 0.0;
 
   /**
    * Clipping planes on the primitive
@@ -305,30 +255,12 @@ function VoxelPrimitive(options) {
   this._modelMatrix = Matrix4.clone(options.modelMatrix ?? Matrix4.IDENTITY);
 
   /**
-   * Model matrix with vertical exaggeration applied. Only used for BOX shape type.
+   * Used to detect if the model matrix is dirty.
    *
    * @type {Matrix4}
    * @private
    */
-  this._exaggeratedModelMatrix = Matrix4.clone(this._modelMatrix);
-
-  /**
-   * The primitive's model matrix multiplied by the provider's model matrix.
-   * This member is not known until the provider is ready.
-   *
-   * @type {Matrix4}
-   * @private
-   */
-  this._compoundModelMatrix = new Matrix4();
-
-  /**
-   * Used to detect if the shape is dirty.
-   * This member is not known until the provider is ready.
-   *
-   * @type {Matrix4}
-   * @private
-   */
-  this._compoundModelMatrixOld = new Matrix4();
+  this._modelMatrixOld = Matrix4.clone(this._modelMatrix);
 
   /**
    * @type {CustomShader}
@@ -656,29 +588,7 @@ function initialize(primitive, provider) {
   primitive.minClippingBounds = minBounds.clone();
   primitive.maxClippingBounds = maxBounds.clone();
 
-  // Initialize the exaggerated versions of bounds and model matrix
-  primitive._exaggeratedMinBounds = Cartesian3.clone(
-    primitive._minBounds,
-    primitive._exaggeratedMinBounds,
-  );
-  primitive._exaggeratedMaxBounds = Cartesian3.clone(
-    primitive._maxBounds,
-    primitive._exaggeratedMaxBounds,
-  );
-  primitive._exaggeratedMinClippingBounds = Cartesian3.clone(
-    primitive._minClippingBounds,
-    primitive._exaggeratedMinClippingBounds,
-  );
-  primitive._exaggeratedMaxClippingBounds = Cartesian3.clone(
-    primitive._maxClippingBounds,
-    primitive._exaggeratedMaxClippingBounds,
-  );
-  primitive._exaggeratedModelMatrix = Matrix4.clone(
-    primitive._modelMatrix,
-    primitive._exaggeratedModelMatrix,
-  );
-
-  checkTransformAndBounds(primitive, provider);
+  checkTransformAndBounds(primitive);
 
   // Create the shape object, and update it so it is valid for VoxelTraversal
   const ShapeConstructor = VoxelShapeType.getShapeConstructor(shapeType);
@@ -1218,13 +1128,12 @@ VoxelPrimitive.prototype.update = function (frameState) {
     return;
   }
 
-  updateVerticalExaggeration(this, frameState);
-
   // Check if the shape is dirty before updating it. This needs to happen every
   // frame because the member variables can be modified externally via the
   // getters.
-  const shapeDirty = checkTransformAndBounds(this, provider);
-  if (shapeDirty) {
+  const shapeDirty = checkTransformAndBounds(this);
+  const exaggerationChanged = updateVerticalExaggeration(this, frameState);
+  if (shapeDirty || exaggerationChanged) {
     this._shapeVisible = updateShapeAndTransforms(this);
     if (checkShapeDefines(this)) {
       this._shaderDirty = true;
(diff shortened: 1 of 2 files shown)


packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -1409,6 +1459,12 @@ function updateVerticalExaggeration(primitive, frameState) {
       (primitive._minBounds.z - relativeHeight) * exaggeration + relativeHeight;
     primitive._exaggeratedMaxBounds.z =
       (primitive._maxBounds.z - relativeHeight) * exaggeration + relativeHeight;
+    primitive._exaggeratedMinClippingBounds.z =
+      (primitive._minClippingBounds.z - relativeHeight) * exaggeration +
+      relativeHeight;
+    primitive._exaggeratedMaxClippingBounds.z =
+      (primitive._maxClippingBounds.z - relativeHeight) * exaggeration +
+      relativeHeight;
```
mzschwartz5: I know this is following the precedent of the class, and I know that these aren't "heights", per se, but I think it would make sense to use `VerticalExaggeration.getHeight()` here instead of essentially re-implementing it.

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -238,6 +238,40 @@ function VoxelPrimitive(options) {
    */
   this._maxClippingBoundsOld = new Cartesian3();
 
+  /**
```
mzschwartz5, line 209: Is it really necessary to have 8 different pieces of state related to clipping bounds? ( (old, new) x (exaggerated, regular) )

I don't totally understand how these are consumed, but I have this feeling that they're not all necessary. 

Are the old members _just_ used for tracking dirtiness? Do we even need to store the exaggerated clipping bounds explicitly, or can we just update the regular clip bounds directly?

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -1645,8 +1711,8 @@ function updateShapeAndTransforms(primitive) {
     primitive._compoundModelMatrix,
     primitive._exaggeratedMinBounds,
     primitive._exaggeratedMaxBounds,
-    primitive.minClippingBounds,
-    primitive.maxClippingBounds,
+    primitive._exaggeratedMinClippingBounds,
+    primitive._exaggeratedMaxClippingBounds,
```
mzschwartz5: Looking at the `VoxelShape` interface, I don't see any `update` method that has 5 arguments. Looks like it's a special case for ellipsoid shaped voxels... I don't love that :D It degrades the benefits of using polymorphism.

mzschwartz5: Is VoxelPrimitive.update called every frame? If not, could we just calculate the exaggerated clipping bounds on the fly here rather than storing it as a member? (Honestly, even if it _is_ called every frame, it's extremely cheap to calculate the exaggeration)

packages/engine/Source/Scene/VoxelPrimitive.js
jjhembd: The expensive call is `VoxelShape.prototype.update`. This call is only necessary if `modelMatrix`, shape bounds, clipping bounds, or exaggeration have changed. (Unfortunately we expose setters for both the shape and clipping bounds, so we do have to check those.)

The checks happen in `checkTransformAndBounds`. If we want to reduce the amount of state, we could perhaps store the state of the exaggeration itself rather than exaggerated bounds. We would then have `verticalExaggerationOld` and `verticalExaggerationRelativeHeightOld`, rather than `exaggerated???BoundsOld` * 4. Would this make sense to you? Or is there a better way to reduce state?

I realized the current setup is state-heavy. I was sort of kicking the cleanup down the road until we resolve the API question: [Do we really need to expose all these bounds setters?](https://github.com/CesiumGS/cesium/issues/12813#issuecomment-3208154460).

packages/engine/Source/Scene/VoxelPrimitive.js
jjhembd: `VoxelPrimitive.update` is called every frame (we should probably shortcut out of it earlier). The reason for storing the bounds was to avoid calls to `VoxelShape.update` -- see my [previous comment above](https://github.com/CesiumGS/cesium/pull/12959#discussion_r2471188942).

packages/engine/Source/Scene/VoxelPrimitive.js
mzschwartz5: > (Unfortunately we expose setters for both the shape and clipping bounds, so we do have to check those.)

Not sure I understand- isn't this a good thing? In the setter, we could set some dirty flag to true.

jjhembd:
@mzschwartz5 thanks for the feedback! I think I addressed it all except for the question about how much (or which) state we should store. See my comment above for one possible resolution--let me know if there's a cleaner way to do it.

packages/engine/Source/Scene/VoxelPrimitive.js
javagl: > In the setter, we could set some dirty flag to true. 

Even if this was done [in some setter like this one](https://github.com/CesiumGS/cesium/blob/80abee31256ca4eefc028c93923f8f982f1ec089/packages/engine/Source/Scene/VoxelPrimitive.js#L1076), there is no sensible encapsulation<sup>\*</sup> for all this. The 'getter' still returns the `Cartesian3`, and users will do `primitive.maxClippingBounds.y += 42` and expect that to have the... expected ...  effect.

Nevertheless, there are indeed quite a few "bounds" here. It _looks_ like there are things that could be "grouped". The comment at https://github.com/CesiumGS/cesium/issues/12813#issuecomment-3208154460 mentioned dedicated classes for the "clipping" information. Without that, it looks like there could be different groupings:

- _Roughly_ something like `ExaggeratablePoint`, to store a `point` and `exaggeratedPoint`, used for `minRender, maxRender: ExaggeratablePoint` and `minClip, maxClip: ExaggeratablePoint`
- _Roughly_ something like an `ExaggeratableRange` with `min/max/exaggereatedMin/exaggeratedMax` (encapsulating the exaggeration functionality), used for `renderRange: ExaggeratableRange` and `clippingRange: ExaggeratableRange`
- (In both cases, these `Exaggeratable...Things...` could store (and hide!) the `old...` properties for the dirty checks where necessary)

This is just from browsing over the code in the GitHub view - I'll have to check this more closely in a dev environment to see which of these (or similar "groupings") could make sense....

---

<sup>\*</sup> This encapsulation would require a `ReadonlyCartesian3`. Maybe, in the future... (TODO: Insert "I would be so happy" meme here.

mzschwartz5: @jjhembd Yes I do think we should be storing the state of the exaggeration rather than the state of everything that depends on it. Then, in `updateVerticalExaggeration`, just compare to the old state and set a dirty flag if needed.

(To be honest, there should really be an event that one could listen to for vertical exaggeration changes... instead, every class that cares is storing the old exaggeration value... but that's a topic for another day)

mzschwartz5 (changes requested):
Per comment, only change I'd still like to see is storing the exaggeration and dirtying based on that, instead of storing old fields for everything that _depends_ on exaggeration

jjhembd:
@mzschwartz5 this is ready for another look.

I updated `VoxelPrimitive` to store exaggeration parameters rather than exaggerated bounds and transform. This moves many bounds and transform calculations into the `updateShapeAndTransforms` helper function, which is only called if the bounds, transform, or exaggeration have changed.

The PR has morphed from a bug fix into more of a refactor, so I updated the testing plan in the description accordingly.

packages/engine/Source/Scene/VoxelPrimitive.js
mzschwartz5, line 1: So, I know we're trying to get this PR into the release, and I don't want to hold it up over style comments, but here's my two cents anyway:

I was imagining something even simpler! No need to store any `old` bounds / clipping bounds. When exaggeration changes, first undo the previous scaling, then apply the new scale. At most, you need to store maybe the old and new exaggeration constants for this - but maybe you just need to store the old exaggeration, and use (but not store) the new exaggeration from `updateExaggeration`.

mzschwartz5: (Then you could also get rid of the chunk of new scratch variables you added, and the block in `updateShapeAndTransforms` that sets all those guys

jjhembd: If we don't store the bounds, then how do we know to update the shape when someone does this?

```javascript
voxelPrimitive.minClippingBounds.y = 3.0;
```

mzschwartz5: Ah... right, forgot about that. Was just focused on the vertical exaggeration. Will think about this, maybe we want to keep things as-is.

jjhembd: Or for that matter, they could be doing this:
```javascript
voxelPrimitive.maxClippingBounds.y = 1.0 + Math.sin(animationTime);
```

mzschwartz5: Man I wish javascript had more strict const syntax

mzschwartz5: I'll keep reviewing as is but, theoretically, what if we used a pattern of getters that return a clone of the internal field; but to avoid excessive memory allocation, the clone is just cloned into a scratch variable.

That way the user can do 
`voxelPrimitive.minClippingBounds.y = 3.0;`

all they want, but it won't change the actual internal field value.

packages/engine/Source/Scene/VoxelPrimitive.js
```diff
@@ -1633,20 +1487,109 @@ function updateBound(primitive, newBoundKey, oldBoundKey) {
...
+  const exaggeratedMaxClippingBounds = Cartesian3.clone(
+    primitive._maxClippingBounds,
+    scratchExaggeratedMaxClippingBounds,
+  );
+  const exaggeratedModelMatrix = Matrix4.clone(
+    primitive._modelMatrix,
+    scratchExaggeratedModelMatrix,
+  );
+
+  if (primitive.shape === VoxelShapeType.ELLIPSOID) {
```
mzschwartz5, line 1528: This seems like a place we'd want to use polymorphism; e.g. `shape.applyExaggeration` or `shape.applyBoundsExaggeration`.

packages/engine/Source/Scene/VoxelPrimitive.js
javagl: "Issue 13421: Voxel bounds are not properly updated" 😶 

---

From https://github.com/CesiumGS/cesium/pull/12959#discussion_r2474919603

> there should really be an event that one could listen to for vertical exaggeration changes... 

This, and not only for the vertical exaggeration, but for many, many other things (`modelMatrix` comes to my mind - but I won't tell you which one 😁 ). I know that the idea behind the "Model-View-Controller Pattern" has lost some appeal due to "Enterprise Software Development" excesses, but the abundant `xxxOld/xxxNew/update() { if (xxxOld!=xxxNew)... }` makes the state space unmanageable, and I'd call it an antipattern. 

(We should track it in its own "We should..."-issue...)

That said: This PR adds a lot of code for tracking these changes. And it's unfortunate that it _has_ to do this. And one could consider to try and "group some things" to not have this spread over 6-8 properties. But I don't think that this PR can do much about the _reason_ for why this is necessary.

mzschwartz5: Yep, totally agree. Scene mode is another egregious example.

(When I read your quote "there should really be an event..." I  thought to myself, "I was just thinking this!" Then I clicked on the link and realized it was my own comment 😂)

jjhembd: @javagl 

> This PR adds a lot of code for tracking these changes.

It _did_ add a lot of code, but I think after https://github.com/CesiumGS/cesium/pull/12959/commits/3f911ee8e18cb04e3c90c0e1c8728853b627022e this PR actually _reduces_ the amount of state tracking. We now only track the things that the user is allowed to change. We stopped tracking several intermediate transforms.

As for the remaining bounds & transform tracking: There is a little class hidden inside `ClippingPlane.js` called `UpdateChangedCartesian3` which could resolve the whole issue with people changing _members_ of "readonly" objects. It does so using Events.

**However**, before we spend time refactoring `VoxelPrimitive.prototype.minClippingBounds` etc etc, we really need to take a hard look at https://github.com/CesiumGS/cesium/issues/12297. Any work we do on these problematic members will probably be deleted soon (before the Voxels GTM), simply because `Cesium3DTileset` does not have any of these `min/maxBounds` or `min/maxClippingBounds` properties.

With that in mind, I think the current state is good enough to fix the bugs during the short remaining lifetime of these members.

jjhembd: As a preview of the likely new API:
- `min/maxBounds` will be fixed on construction, and never need to be checked again
- All clipping will be done via an expanded `ClippingPlane`-like class--see https://github.com/CesiumGS/cesium/issues/12938

packages/engine/Source/Scene/VoxelPrimitive.js
jjhembd: Good idea. I think it's out of scope for this PR.

Before we refactor the `VoxelShape` classes, we probably need to start by resolving the `VoxelPrimitive.shape` vs `VoxelPrimitive._shape` confusion. Despite the similar names, they are very different things.

mzschwartz5: > I think it's out of scope for this PR.

Fair enough

> Despite the similar names, they are very different things.

That's unfortunate lol

mzschwartz5 (approved):
Followed testing instructions and everything works as advertised.

See [comment](https://github.com/CesiumGS/cesium/pull/12959/files#r2482269403) for future work on refactoring, but we're considering this out of scope for now.

packages/engine/Source/Scene/VoxelPrimitive.js
javagl: That `UpdateChangedCartesian3` looks like an application of the "observable" concept... if it wasn't specific for the `clippingPlane`. (It could so easily have its own `Event`...). I think that this could help to model some "`Cartesian3` getters" more cleanly. (But of course, all that is beyond what this PR is about. This PR fixes a bug. Refactorings and other improvements should/could happen independently)

Issue #12296, opened 2024-11-08
Bring `Cesium3DTilesTerrainProvider` into `main`


ggetz:
We have a long-running branch—[`3d-tiles-terrain`](https://github.com/CesiumGS/cesium/tree/3d-tiles-terrain)— that adds a new `Cesium3DTilesTerrainProvider` class. This is used for rending global 3D Tiles via the existing terrain engine. 

This terrain provider supports only a subset of 3D tilesets (ie. 3D Tiles 1.1 with implicit tiling schemes), so it may not quite ready for prime time yet. As such we should use the `experimental` tag to give us some flexibility if things need to change.

In the interest of maintenance as well as making this early work generally available, let's get this branch updated and merged into `main`. We'll also want to ensure the testing, documentation, and API structure are up to par as a part of this.

ggetz:
@lilleyse Are there other specific requirements for 3D tilesets using this feature that should be called out in the docs?

lilleyse:
3D Tiles loaded with Cesium3DTilesTerrainProvider must follow this structure:

* Conforms to the same WGS84 double-headed quadtree tiling scheme as quantized mesh. Must have two root tiles with region bounding volumes and implicit tiling.
* Multiple contents are not allowed
* Tile metadata must include `TILE_MINIMUM_HEIGHT`, `TILE_MAXIMUM_HEIGHT`, `TILE_BOUNDING_SPHERE`, `TILE_HORIZON_OCCLUSION_POINT` semantics
* Only GLB is supported, not glTF
* Each GLB must have a single node, mesh, and primitive
* Primitive must have `POSITION` attribute
* Primitive must have `NORMAL` attribute if `requestVertexNormals` is true
* Primitive must have indices with `UNSIGNED_SHORT` or `UNSIGNED_INT` component type
* Attributes and indices must be tightly packed
* If `requestWaterMask` is true, primitive must reference an `EXT_structural_metadata` property texture with `WATERMASK` semantic.
* Primitive must have [`CESIUM_tile_edges`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/CESIUM_tile_edges) extension
* Model may be compressed with `EXT_meshopt_compression` / `KHR_mesh_quantization`

PR #12963, opened 2025-10-09
3d tiles terrain

jjhembd:
# Description

This PR adds experimental support for loading 3D Tiles as terrain, via `Cesium3DTilesTerrainProvider`.

## What 3D Tiles data can be used?
3D Tiles loaded with Cesium3DTilesTerrainProvider must follow this structure ([thanks @lilleyse for the list](https://github.com/CesiumGS/cesium/issues/12296#issuecomment-2476788817)): 

* Conforms to the same WGS84 double-headed quadtree tiling scheme as quantized mesh. Must have two root tiles with region bounding volumes and implicit tiling.
* Multiple contents are not allowed
* Tile metadata must include `TILE_MINIMUM_HEIGHT`, `TILE_MAXIMUM_HEIGHT`, `TILE_BOUNDING_SPHERE`, `TILE_HORIZON_OCCLUSION_POINT` semantics
* Only GLB is supported, not glTF
* Each GLB must have a single node, mesh, and primitive
* Primitive must have `POSITION` attribute
* Primitive must have `NORMAL` attribute if `requestVertexNormals` is true
* Primitive must have indices with `UNSIGNED_SHORT` or `UNSIGNED_INT` component type
* Attributes and indices must be tightly packed
* If `requestWaterMask` is true, primitive must reference an `EXT_structural_metadata` property texture with `WATERMASK` semantic.
* Primitive must have [`CESIUM_tile_edges`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/CESIUM_tile_edges) extension
* Model may be compressed with `EXT_meshopt_compression` / `KHR_mesh_quantization`

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/12296

## Testing plan

- Run the new specs `Cesium3DTilesTerrainProviderSpec` and `Cesium3DTilesTerrainDataSpec`
- Load the [Globe Materials - 3D Tiles Terrain](http://localhost:8080/Apps/Sandcastle2/index.html?id=globe-materials-3d-tiles-terrain) Sandcastle

## TODO before moving this out of "Draft":
- [x] Update Sandcastles to new gallery format
- [x] Open separate PR for doc fixes, `Check` calls: `EllipsoidalOccluder`, `GoogleEarthEnterpriseTerrainData`, `TerrainEncoding` (separate from other changes in file), `binarySearch`, `mergeSort`, `QuantizedMeshTerrainDataSpec`. Beware incorrect capitalizations like `{Object}`, `{Number[]}`! See https://github.com/CesiumGS/cesium/pull/12969
- [x] Verify changes to `ResourceCache`, avoid additional URI parsing.
- [x] Replace promises in specs with async/await
- [x] ~Consider dropping the duplicate `ImplicitSubtreeCache` in `Cesium3DTilesTerrainProvider.js` (replace with `ImplicitSubtreeCache.js`)~ The two versions of `ImplicitSubtreeCache` have some key differences, so I think a refactor is out of scope for this PR.
- [x] Understand changes to signature in `TerrainEncoding`

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/Cesium3DTilesTerrainData.js
+++ b/packages/engine/Source/Core/Cesium3DTilesTerrainData.js
@@ -858,7 +858,7 @@ const scratchUv2 = new Cartesian2();
 const scratchBary = new Cartesian3();
 
 /**
- * Computes the terrain height at a specified longitude and latitude.
+ * Computes the terrain height at a specified longitude and latitude. Returns 0.0 if the position is outside the mesh.
  * @private
  * @param {TerrainMesh} mesh The terrain mesh.
  * @param {Rectangle} rectangle The rectangle covered by this terrain data.
@@ -923,6 +923,7 @@ function interpolateMeshHeight(mesh, rectangle, longitude, latitude) {
   }
 
   // Position does not lie in any triangle in this mesh.
+  // This should not happen often since we start by clamping to the rectangle.
   return 0.0;
 }
 
--- a/packages/engine/Source/Core/TerrainProvider.js
+++ b/packages/engine/Source/Core/TerrainProvider.js
@@ -15,6 +15,7 @@ import CesiumMath from "./Math.js";
  * @see CesiumTerrainProvider
  * @see VRTheWorldTerrainProvider
  * @see GoogleEarthEnterpriseTerrainProvider
+ * @see ArcGISTiledElevationTerrainProvider
  * @see Cesium3DTilesTerrainProvider
  */
 function TerrainProvider() {
--- a/packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/sandcastle.yaml
+++ b/packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/sandcastle.yaml
@@ -1,6 +1,8 @@
 legacyId: Globe Materials – 3D Tiles Terrain.html
 title: Globe Materials – 3D Tiles Terrain
-description: Apply materials to the globe.
+description: Load a 3D Tiles tileset as terrain, and apply globe material properties to style the terrain.
 labels:
+  - 3D Tiles
+  - Terrain
   - Showcases
 thumbnail: thumbnail.jpg


packages/engine/Source/Scene/GltfJsonLoader.js
```diff
@@ -222,11 +222,19 @@ function loadEmbeddedBuffers(gltfJsonLoader, gltf) {
...
       const resourceCache = gltfJsonLoader._resourceCache;
-      const bufferLoader = resourceCache.getEmbeddedBufferLoader({
+      let bufferLoader = resourceCache.getEmbeddedBufferLoader({
         parentResource: gltfJsonLoader._gltfResource,
         bufferId: bufferId,
-        typedArray: source,
       });
+
+      if (!defined(bufferLoader)) {
+        bufferLoader = resourceCache.addBufferLoader({
```
jjhembd: Both `resourceCache.getEmbeddedBufferLoader` and `.addBufferLoader` call `ResourceCacheKey.getEmbeddedBufferCacheKey`, which involves expensive URI parsing. For other 3D Tilesets, we have previously found that [URI parsing can be a bottleneck](https://github.com/CesiumGS/cesium/pull/11202).

We may want to consider reverting to the old `getEmbeddedBufferLoader` syntax, which re-used the key.

packages/engine/Source/Core/HeightmapTerrainData.js
```diff
@@ -94,16 +94,12 @@ import TerrainProvider from "./TerrainProvider.js";
...
-  if (!defined(options) || !defined(options.buffer)) {
-    throw new DeveloperError("options.buffer is required.");
-  }
-  if (!defined(options.width)) {
-    throw new DeveloperError("options.width is required.");
-  }
-  if (!defined(options.height)) {
-    throw new DeveloperError("options.height is required.");
-  }
+  Check.defined("options.buffer", options.buffer);
```
jjspace: Edit: NO change requested right now, just commenting for discussion. Probably best as it's own issue

I think this is equivalent to:

```js
Check.defined("options", options);
Check.defined("options.buffer", options.buffer);
```
Or maybe even better `Check.typeOf.object("options", options)`

I started to call this out in other places but realized we actually use this pattern a lot and it's probably not worth addressing in this PR.

Specifically I think it's an issue when `options` is not marked as optional but the `??` makes it _behave_ as if it is. Using this will punt potential errors down the call chain to somewhere where it can't work with `undefined`. I think it would be better if this function itself throws errors because it can't access values of `undefined`.

packages/engine/Source/Core/HeightmapTerrainData.js
```diff
@@ -94,16 +94,12 @@ import TerrainProvider from "./TerrainProvider.js";
...
+  Check.defined("options.width", options.width);
+  Check.defined("options.height", options.height);
```
jjspace: Can we replace `Check.defined` with a more specific `typeOf` check where it makes sense while we're already updating these?

packages/engine/Source/Core/TerrainProvider.js
```diff
@@ -236,7 +236,79 @@ TerrainProvider.getRegularGridAndSkirtIndicesAndEdgeIndices = function (
...
+ * @returns {number}
+ */
+TerrainProvider.getSkirtIndexCountWithFilledCorners = function (
+  skirtVertexCount,
+) {
+  return ((skirtVertexCount - 4) * 2 + 4) * 3;
+};
+
+/**
+ * Adds skirt indices.
```
jjspace, line 305: I think there should be some sort of comment or callout that this function is without corners. Or just mention that if you want corners use the other new function.

packages/engine/Source/Core/binarySearch.js
```diff
@@ -4,7 +4,7 @@ import Check from "./Check.js";
  * Finds an item in a sorted array.
  *
  * @function
- * @param {Array} array The sorted array to search.
+ * @param {Array|Int8Array|Uint8Array|Int16Array|Uint16Array|Int32Array|Uint32Array|Float32Array|Float64Array} array The sorted array to search.
```
jjspace, line 7:
```suggestion
 * @param {Array|TypedArray} array The sorted array to search.
```

Should this be simplified to any `TypedArray`? Or are there types we specifically _don't_ support (or don't want to support)?

packages/engine/Source/Core/mergeSort.js
```diff
@@ -56,7 +56,7 @@ function sort(array, compare, userDefinedObject, start, end) {
  * A stable merge sort.
  *
  * @function mergeSort
- * @param {Array} array The array to sort.
+ * @param {Array|Int8Array|Uint8Array|Int16Array|Uint16Array|Int32Array|Uint32Array|Float32Array|Float64Array} array The array to sort.
```
jjspace, line 59: Same question here, `TypedArray`?

packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/sandcastle.yaml
```diff
@@ -0,0 +1,6 @@
+legacyId: Globe Materials – 3D Tiles Terrain.html
+title: Globe Materials – 3D Tiles Terrain
+description: Apply materials to the globe.
+labels:
+  - Showcases
```
jjspace, line 7: This should probably have at least one other label. Maybe `3D Tiles` and `Terrain`? @ggetz any opinions?

packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/sandcastle.yaml
```diff
@@ -0,0 +1,6 @@
+legacyId: Globe Materials – 3D Tiles Terrain.html
+title: Globe Materials – 3D Tiles Terrain
+description: Apply materials to the globe.
```
jjspace: Can we expand this description a bit? Something to make it more clear what this is and why it's different than the other "Globe Materials" sandcastles?

packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/index.html
```diff
@@ -0,0 +1,8 @@
+<style>
+  @import url(../templates/bucket.css);
+</style>
+<div id="cesiumContainer" class="fullSize"></div>
+<div id="loadingOverlay"><h1>Loading...</h1></div>
+<div id="toolbar">
+  <div id="zoomButtons"></div>
+</div>
```
jjspace:
```suggestion
<div id="toolbar"></div>
```

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
```diff
@@ -0,0 +1,936 @@
...
+/**
+ * Computes the terrain height at a specified longitude and latitude.
+ * @function
+ *
+ * @param {Rectangle} rectangle The rectangle covered by this terrain data.
+ * @param {number} longitude The longitude in radians.
+ * @param {number} latitude The latitude in radians.
+ * @returns {number|undefined} The terrain height at the specified position.  If the position
+ *          is outside the rectangle, this method will extrapolate the height, which is likely to be wildly
+ *          incorrect for positions far outside the rectangle.
```
jjspace: It would be good to describe when `undefined` is returned. In this case because the mesh itself is not defined. Sometimes nice to also include that in the function description itself.

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
```diff
@@ -0,0 +1,936 @@
...
+  waterMask: {
+    get: function () {
+      return this._waterMask;
+    },
+  },
+});
+
+/**
+ * Computes the terrain height at a specified longitude and latitude.
+ * @function
```
jjspace: I could be wrong but I believe the `@function` tag is not necessary. Double-check me. Same with other functions in this file below.

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
```diff
@@ -0,0 +1,936 @@
...
+  });
+}
+
+const scratchUv0 = new Cartesian2();
+const scratchUv1 = new Cartesian2();
+const scratchUv2 = new Cartesian2();
+const scratchBary = new Cartesian3();
+
+/**
+ * Computes the terrain height at a specified longitude and latitude.
```
jjspace: Might be worth mentioning when and why this function returns `0.0`

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
```diff
@@ -0,0 +1,936 @@
...
+/**
+ * @private
+ * @constructor
+ *
+ * @param {object} options Object with the following properties:
+ * @param {TerrainMesh} options.terrainMesh The terrain mesh.
+ * @param {number} options.skirtHeight The height of the skirt to add on the edges of the tile.
+ * @param {Credit[]} [options.credits] Array of credits for this tile.
+ */
+function Cesium3DTilesUpsampleTerrainData(options) {
```
jjspace, line 495: Should this class be in it's own file? no strong preference I think, just asking

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
```diff
@@ -0,0 +1,936 @@
...
+ *                 instead upsampled from the parent.  The bit values are as follows:
+ *                 <table>
+ *                  <tr><th>Bit Position</th><th>Bit Value</th><th>Child Tile</th></tr>
+ *                  <tr><td>0</td><td>1</td><td>Southwest</td></tr>
+ *                  <tr><td>1</td><td>2</td><td>Southeast</td></tr>
+ *                  <tr><td>2</td><td>4</td><td>Northwest</td></tr>
+ *                  <tr><td>3</td><td>8</td><td>Northeast</td></tr>
+ *                 </table>
+ * @param {Uint8Array} [options.waterMask] The buffer containing the water mask.
+ * @see TerrainData
```
jjspace, line 50: The `TerrainData` class has a list of `@see` classes too, should it link back to this one as well?

packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/main.js
```diff
@@ -0,0 +1,93 @@
+import * as Cesium from "cesium";
+
+const terrainProvider =
+  await Cesium.Cesium3DTilesTerrainProvider.fromIonAssetId(3923568, {
+    requestVertexNormals: true, // Needed for hillshade lighting
+    requestWaterMask: true, // Needed to distinguish land from water
+  });
```
jjspace, line 7: This sandcastle essentially has only 1 change compared to the "Globe materials - Water mask" sandcastle. That's the terrain provider. Would it be better to combine them and have a dropdown or something to toggle which terrain provider is being used? Or is there a better, more unique, way to demonstrate that this is using _3D Tiles_ as terrain?

packages/engine/Source/Core/Cesium3DTilesTerrainProvider.js
```diff
@@ -0,0 +1,1052 @@
...
+ *
+ * A {@link TerrainProvider} that accesses terrain data in a 3D Tiles format.
+ *
+ * @alias Cesium3DTilesTerrainProvider
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ * @constructor
+ *
+ * @param {Cesium3DTilesTerrainProvider.ConstructorOptions}[options] An object describing initialization options
+ *
+ * @see TerrainProvider
```
jjspace, line 53: Similar comment here, should this class be referenced from the `TerrainProvider` class?

packages/engine/Source/Core/Cesium3DTilesTerrainProvider.js
```diff
@@ -0,0 +1,1052 @@
...
+ * @param {number} _y The Y coordinate of the tile for which to request geometry.
+ * @param {number} _level The level of the tile for which to request geometry.
+ * @returns {Promise<void>|undefined} Undefined if nothing need to be loaded or a Promise that resolves when all required tiles are loaded
+ */
+Cesium3DTilesTerrainProvider.prototype.loadTileDataAvailability = function (
+  _x,
+  _y,
+  _level,
+) {
+  return undefined;
```
jjspace, line 664: Is this function supposed to be "unimplemented"?

jjspace (commented):
Thanks @jjhembd, overall this looks pretty good to me!

I mostly reviewed this from a docs perspective. I believe most of this has already been verified and sitting around for a while. I also tested the sandcastle some but probably not exhaustive. I think it could be good to have a second perspective on the actual logic

packages/engine/Source/Core/mergeSort.js
jjhembd: Nice... somehow I thought we didn't have that option, but I see we're using it elsewhere.

jjhembd: On second thought: everywhere else we were using `TypedArray` as a type was marked `@private`. Using it in a public class gives this error:

```bash
$ npm run make-zip
// ...
Source/Cesium.d.ts:18622:45 - error TS2304: Cannot find name 'TypedArray'
```

@jjspace I would love for us to have `TypedArray` defined somewhere. What is the clean way to add native types?

javagl: Some related discussion may be around https://github.com/microsoft/TypeScript/issues/15402#issuecomment-362415653 , which I have _not_ yet digested in all depth, but which might be relevant here. (It _looks_ like there is no clean way to integrate that on the "language" level, but for the docs, creating an own pseudo-type (union) might be a reasonable workaround...)

packages/sandcastle/gallery/globe-materials-3d-tiles-terrain/main.js
jjhembd: I updated the description in the .yaml file to emphasize the 3D Tiles as terrain aspect.
I agree there's still a lot of overlap. I can combine them if you prefer, though I think I would set the 3D Tiles data as the starting point, just to make sure we capture that dataset & code path during end-to-end testing or manual release checks.

packages/engine/Source/Core/Cesium3DTilesTerrainProvider.js
jjhembd: Good question... Only `CesiumTerrainProvider` implements this function.
`EllipsoidTerrainProvider`, `VRTheWorldTerrainProvider`, `GoogleEarthEnterpriseTerrainProvider`, and `ArcGISTiledElevationTerrainProvider` all return `undefined` too.

I haven't traced through how it's used, but this looks like an interface feature that isn't absolutely necessary for terrains to work.

packages/engine/Source/Core/Cesium3DTilesTerrainData.js
jjhembd: `Cesium3DTilesUpsampleTerrainData` and its "parent" `Cesium3DTilesTerrainData` share the helper functions `interpolateMeshHeight` and `upsampleMesh`.

To separate the classes while avoiding circular dependencies, I suppose we could make those helpers static methods of `Cesium3DTilesUpsampleTerrainData`. This structure makes sense in the case of `upsampleMesh`, which always returns a `Cesium3DTilesUpsampleTerrainData`, but feels a little forced in the case of `interpolateMeshHeight`.

The other issue is that both the helpers have very long function signatures, which is tolerable (maybe?) for an internal helper, but doesn't make for a very pretty (or maintainable or testable) API for a class method.

I don't love the current structure but I think a refactor might be too much for this PR.

jjhembd:
Thanks for the feedback @jjspace. I think I addressed it where I could, and left comments where the fix is not obvious or maybe out of scope. Let me know if I'm missing anything, or if there's a good way to clean up that I didn't think of.

jjspace (approved):
Thanks for the changes @jjhembd! LGTM 👍

packages/engine/Source/Core/mergeSort.js
donmccurdy: This might do the trick:

- https://github.com/CesiumGS/cesium/pull/13187

PR #11080, opened 2023-02-09
Use primitive type instead of constructor type

onsummer:
Make types more precision.

This PR change constructor type to primitive type, e.g.

```diff
/**
 * Removes a key-value pair from the collection.
 *
- * @param {String|Number} key The key to be removed.
- * @returns {Boolean} True if it was removed, false if the key was not in the collection.
+ * @param {string|number} key The key to be removed.
+ * @returns {boolean} True if it was removed, false if the key was not in the collection.
 */
AssociativeArray.prototype.remove = function (key) {
```

TODO:
- [x] Use primitive type instead of Constructor
- [x] Update CHANGES.md

PR #12969, opened 2025-10-10
Clean up docs and type checks for terrain providers

jjhembd:
# Description

This is another small step towards resolving https://github.com/CesiumGS/cesium/issues/12296. Key changes:

- Incorporates many of the JSDoc improvements and type checks from the `3d-tiles-terrain` branch
- Fixes use of constructor types, to align with https://github.com/CesiumGS/cesium/pull/11080. Using the constructor type is [discouraged](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#legacy-type-synonyms)

## Testing plan

# Author checklist

- [ ] ~I have updated `CHANGES.md` with a short summary of my change~ This PR only updates docs, mostly in private functions or in ways that do not change the meaning of the API docs

### DIFF
--- a/packages/engine/Specs/Core/TerrainEncodingSpec.js
+++ b/packages/engine/Specs/Core/TerrainEncodingSpec.js
@@ -40,7 +40,12 @@ describe("Core/TerrainEncoding", function () {
     expect(encoding.toScaledENU).not.toBeDefined();
     expect(encoding.fromScaledENU).not.toBeDefined();
     expect(encoding.matrix).not.toBeDefined();
-    expect(encoding.hasVertexNormals).not.toBeDefined();
+    expect(encoding.hasVertexNormals).toBe(false);
+    expect(encoding.hasWebMercatorT).toBe(false);
+    expect(encoding.hasGeodeticSurfaceNormals).toBe(false);
+    expect(encoding.exaggeration).toBe(1.0);
+    expect(encoding.exaggerationRelativeHeight).toBe(0.0);
+    expect(encoding.stride).toBe(6);
   });
 
   it("constructs without quantization", function () {


packages/engine/Source/Core/TerrainEncoding.js
```diff
@@ -24,10 +25,10 @@ const SHIFT_LEFT_12 = Math.pow(2.0, 12.0);
...
  *
  * @param {Cartesian3} center The center point of the vertices.
- * @param {AxisAlignedBoundingBox} axisAlignedBoundingBox The bounds of the tile in the east-north-up coordinates at the tiles center.
- * @param {number} minimumHeight The minimum height.
- * @param {number} maximumHeight The maximum height.
- * @param {Matrix4} fromENU The east-north-up to fixed frame matrix at the center of the terrain mesh.
+ * @param {AxisAlignedBoundingBox} [axisAlignedBoundingBox] The bounds of the tile in the east-north-up coordinates at the tiles center.
+ * @param {number} [minimumHeight] The minimum height.
+ * @param {number} [maximumHeight] The maximum height.
+ * @param {Matrix4} [fromENU] The east-north-up to fixed frame matrix at the center of the terrain mesh.
```
jjspace, line 31: If these are now optional then the type of `this.minimumHeight` etc that's defined lower should include `| undefined`

packages/engine/Source/Core/TerrainEncoding.js
```diff
@@ -188,6 +189,16 @@ function TerrainEncoding(
   this._calculateStrideAndOffsets();
 }
 
+/**
+ * @param {Float32Array} vertexBuffer The buffer to write to.
```
jjspace, line 197: There's a lot of jsdoc blocks that are added for functions in this file that don't have descriptions. It's great to get the typings for editor hints but should some of them also get descriptions? 

No true "action item" here but maybe just do a quick pass on all the blocks and double check if it makes sense to add more of a description than the function name itself.

packages/engine/Source/Scene/Globe.js
```diff
@@ -646,7 +646,7 @@ function makeShadersDirty(globe) {
   const fragmentSources = [AtmosphereCommon, GroundAtmosphere];
   if (
     defined(globe._material) &&
-    (!requireNormals || globe._terrainProvider.requestVertexNormals)
+    (!requireNormals || globe._terrainProvider.hasVertexNormals)
```
jjspace, line 649: Was this change intentional? I didn't track it down but this seems to be the only change that's not just documentation in this PR. Just want to do a double check that this is accurate.

packages/engine/Source/Core/EllipsoidalOccluder.js
```diff
@@ -247,7 +247,7 @@ EllipsoidalOccluder.prototype.computeHorizonCullingPointPossiblyUnderEllipsoid =
  *                     A reasonable direction to use is the direction from the center of the ellipsoid to
  *                     the center of the bounding sphere computed from the positions.  The direction need not
  *                     be normalized.
- * @param {number[]} vertices  The vertices from which to compute the horizon culling point.  The positions
+ * @param {number[]|Float32Array|Float64Array} vertices  The vertices from which to compute the horizon culling point. The positions
```
jjspace, line 250: I think there are a few limitations or differences between a "normal" array and a `TypedArray` in js. I assume these changes were already verified in the larger PR but just wanted to call it out to verify that these changes are ok to do.

Same in the `TerrainMesh.js` file.

packages/engine/Source/Scene/Globe.js
jjhembd: Yes, this appears to have been a simple bug, which wasn't exposed until we started implementing 3D Tiles as terrain. The `TerrainProvider` interface doesn't have a `requestVertexNormals` flag.
`CesiumTerrainProvider` does have a `._requestVertexNormals` flag (note the underscore) which is used in the `.hasVertexNormals` getter, so maybe that's where the mistake crept in.

packages/engine/Source/Core/EllipsoidalOccluder.js
jjhembd: Thanks, I verified that in this file we're just looping over the vertices with a simple `for` loop.
`TerrainMesh` is more complicated but yes I think it works with both types.

jjhembd:
Thanks @jjspace for the feedback! I think I addressed it all, let me know if I missed anything.

packages/engine/Source/Core/TerrainEncoding.js
```diff
@@ -24,15 +25,15 @@ const SHIFT_LEFT_12 = Math.pow(2.0, 12.0);
  * @constructor
  *
  * @param {Cartesian3} center The center point of the vertices.
- * @param {AxisAlignedBoundingBox} axisAlignedBoundingBox The bounds of the tile in the east-north-up coordinates at the tiles center.
- * @param {number} minimumHeight The minimum height.
- * @param {number} maximumHeight The maximum height.
- * @param {Matrix4} fromENU The east-north-up to fixed frame matrix at the center of the terrain mesh.
+ * @param {AxisAlignedBoundingBox|undefined} [axisAlignedBoundingBox] The bounds of the tile in the east-north-up coordinates at the tiles center.
+ * @param {number|undefined} [minimumHeight] The minimum height.
```
jjspace: This isn't quite what I meant, sorry if I was unclear. The `[]` alone should indicate that the function argument can be undefined. The problem is the actual class property definitions lower in the file. For example
```js

  /**
   * The minimum height of the tile including the skirts.
   * @type {number}
   */
  this.minimumHeight = minimumHeight;
```
This says that `minimumHeight` is always a number but that's not true if the argument is optional. This is actually a problem with more than just the arguments changed here like `toEnu` and `matrix`. It also could cascade to other places that use these properties (like the `.clone()` function) but it's clearly been working so I'm not sure it's entirely worth the effort to track them all down for this PR.

jjhembd: @jjspace thanks for that clarification, I think I fixed it now.
I checked the `.clone` method: the matrices are cloned via `Matrix4.clone`, which cleanly returns `undefined` for an `undefined` argument (although the return type doesn't indicate this). So at least there's not a bug at the moment. Maybe we should revisit the `Matrix?.clone` and `Cartesian?.clone` types at some point.

jjspace (approved):
LGTM, thanks @jjhembd

packages/engine/Source/Core/TerrainEncoding.js
jjspace: Yea I looked into that too. There's definitely lots of places where this cascades into other functions. Always good to take little steps than no steps, can always clean up more later.

Issue #11827, opened 2024-02-13
Coverage link broken in CI


ggetz:
### What happened?

Clicking on the "Details" link in the CI check for coverage...

<img width="1013" alt="image" src="https://github.com/CesiumGS/cesium/assets/4439461/e426503a-76c9-4fe1-9a0d-8b1c7ca32fc9">

...takes one to an Access Denied error page.

<img width="1050" alt="image" src="https://github.com/CesiumGS/cesium/assets/4439461/90807201-50b4-434a-bec2-b8a3752746b9">

PR #12971, opened 2025-10-10
Fix coverage link, streamline env vars in CI

ggetz:
## Description

TLDR; This PR should fix the link to coverage results in GitHub PRs, but more broadly, does some housekeeping of our deploy and status scripts.

### Mismatched environment variables

Between incremental updates and migrating between various services and tools, our build and deploy scripts have become quite sprawling. Some things also fall through the cracks, like [the broken test coverage link in the CI checks](https://github.com/CesiumGS/cesium/issues/11827), but we're also dragging along patterns that are no longer serving us. This PR does not fix everything, but attempts to consolidate and cleanup some of these CI/CD related things.

I noticed when reviewing https://github.com/CesiumGS/cesium/pull/12957 that there are several misconfigured environment variables in the `dev` workflow as opposed to the `deploy` workflow where most other build artifacts are uploaded. Mis-matches across the workflows were ultimately resulting in broken links.

  - When using the `pull_request` hook, the GitHub `ref_name` is different than the branch name. This ended up using different versions of "DEPLOYED_URL" leading to a broken link.
  - A similar issue arises when trying to use the commit SHA for operations like setting statuses. In `pull_request` triggers, the value of `COMMIT_SHA` will the _merge_ commit.

### Change summary

- _Does not_ change the production or dedicated sandcastle deployment workflows beyond removing redundant `env` properties.
- Cleaned up some (quite old) deploy scripts:
  - Moved the `setDeployStatus` script out of `gulpfile.js` as it is unrelated to the build processes—It is now a standalone node script with command line help docs.
  - Replaced with `setCommitStatus`. For flexibility, the command itself is more granular so it can be run per-artifact and per-status update. This allows check updates to run separately in different tasks or workflows as needed.
  - Removed the `deploySetVersion` script out of `gulpfile.js` in favor of the [`npm version` cli command](https://docs.npmjs.com/cli/v8/commands/npm-version) which we use for our release processes.
  - Since they are generally only the concerns of the maintainers, these script have been removed from `package.json` and the corresponding build guide.
- Moved the coverage job to it's own workflow, having it _only_ run on pull requests triggers. This simplifies the job a lot  and, to my knowledge, the check itself isn't super helpful until a PR is opened.
- Created `.env` files in which can configure common environment variables as described in the updated docs of [`ContinuousIntegration/README.md`](https://github.com/CesiumGS/cesium/blob/coverage-link/Documentation/Contributors/ContinuousIntegration/README.md#environment-variables)
  - This adds a dependency on [`dotenvx`](https://github.com/dotenvx/dotenvx) as it allows us to variable expansion and simple command substitution.
  - Defines values accounting for variations in environments:
    - In CI, the trigger can be either `push` or `pull_request`. [Since `pull_request` does not run in the context of the branch itself](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request:~:text=Note%20that%20GITHUB_SHA%20for%20this%20event%20is%20the%20last%20merge%20commit%20of%20the%20pull%20request%20merge%20branch.%20If%20you%20want%20to%20get%20the%20commit%20ID%20for%20the%20last%20commit%20to%20the%20head%20branch%20of%20the%20pull%20request%2C%20use%20github.event.pull_request.head.sha%20instead.), we need to account for the available [GitHub context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context) to grab branch names and commit SHA's.
    - Locally, fall back to git commands to allow for development testing.
  - Create a new env var, `CESIUM_VERSION`, that detects the current package version making it more convenient to access across various contexts.
 
## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/11827

## Testing plan

1. When CI runs, all expected jobs should start:
image

2. CI artifacts should then show up as pending:
image

3. CI artifacts should all transition to success and each link should work:
    a. "coverage results" should navigate to the coverage results index
    b. "cesium-*.tgz" should download the packed npm package
    c. "Cesium-*.zip" should download the release zip
    d. "index.html` should navigate to the deployed static build
    
image

# Author checklist


ggetz:
Wait to review; This needs an additional change.

ggetz:
This is ready for review and the PR description is up-to-date.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: install node 22
+        uses: actions/setup-node@v5
+        with:
+          node-version: "22"
+      - name: npm install
+        run: npm install
+      - name: set status pending
+        if: ${{ env.AWS_ACCESS_KEY_ID != '' }}
+        run: |
+          npx @dotenvx/dotenvx run -f ./.github/workflows/.env.pull_request -- \
```
jjspace, line 29: Why are we calling `dotenvx` from `npx` instead of using it directly in the script files? is it just so we can specify the extra `.env.pull_request` files?

Also as an aside for my curiosity, what is the reason you went with `dotenvx` instead of just `dotenv`? a quick search seems that the motivation would be to encrypt the env files with secrets but we aren't doing that. Is this future proofing?

.github/workflows/deploy.yml
```diff
@@ -61,7 +68,11 @@ jobs:
           --exclude "node_modules/*" \
           --exclude "scripts/*" \
           --exclude "Tools/*" \
-          --delete
-      - name: set status
+          --delete'
```
jjspace, line 69:
```suggestion
          --delete
```

 (I believe this is extra/a typo?)

.env
```diff
@@ -0,0 +1,17 @@
+# These variables are ovveridden in CI with GitHub Action variables if triggered from pull-reqest or push;
```
jjspace:
```suggestion
# These variables are overridden in CI with GitHub Action variables if triggered from pull-request or push; 
```

.github/workflows/deploy.yml
ggetz: I don't think so—This should be the closing single quote paired with line 58.

.github/workflows/coverage.yml
ggetz: I chose `dotenvx` specifically because it supports [variable expansion](https://dotenvx.com/docs/advanced/run-variable-expansion) and [command substitution](https://dotenvx.com/docs/advanced/run-command-substitution), which I don't believe "regular" `dotenv` supports.

ggetz:
> CI seems to be passing and working as expected. I assume there's not really any way to test this locally?

No, not end-to-end. You could test some commands locally, but if CI is working as expected I think that's unnecessary.

.github/workflows/coverage.yml
ggetz: > Why are we calling dotenvx from npx instead of using it directly in the script files? is it just so we can specify the extra .env.pull_request files?

Yes. The default `.env` is being loaded from the script itself, but this command is what allows us to override with the CI-specific values in `.env.pull_request`. If there's an alternative syntax you know of, please tell!

.github/workflows/deploy.yml
jjspace: Ah I see now... that's deceptively hard to see :detective: A consequence of needing to use `sh -c` I guess. I also found the extra `--` really hard to connect to the `npx` command at first but might just have to deal with it :shrug:

ggetz:
@jjspace Could you take a final pass on this PR and merge if happy? I think this addresses the rogue coverage artifacts you mentioned seeing.

scripts/setCommitStatus.js
```diff
@@ -0,0 +1,132 @@
+import yargs from "yargs";
+import { hideBin } from "yargs/helpers";
+import "@dotenvx/dotenvx/config";
```
jjspace, line 3: Should this be imported in the gulpfile and all our scripts?
Alternatively should all our `package.json` scripts be prefixed with `dotenvx run`?

.github/workflows/coverage.yml
jjspace: I think my expectation is from using older versions of dotenv. I was used to the paradigm of always _only_ including the package and configuring it in the file it's used in. It looks like this is still an option using [`dotenvx.config([path])`](https://dotenvx.com/docs/advanced/config-path). However I don't think this allows for the command line level variable substitution which it seems we need. 

I think it's just kinda confusing not knowing where and when things are defined, especially when the commands are different in CI. I think I get it now but it's a bit weird. 

For example, this command only uses `-f ./.github/workflows/.env.pull_request` because the `setCommitStatus.js` script imports `@dotenvx/dotenvx/config` to setup the "base" `.env` however the command lower in this file requires `-f ./.github/workflows/.env.pull_request -f .env` in order to include both.

Plus my other comment about not always being included could lead to devs getting confused why `process.env` vars are set up one way in some files but missing in others if they forget an import or to run the command wrapped with `dotenvx run`

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,45 @@
...
+      contents: read
+    steps:
+      - uses: actions/checkout@v5
+      - name: install node 22
+        uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+      - name: npm install
+        run: npm install
+      - name: set status pending
```
jjspace, line 26: What if these jobs fail? Do these pending statuses time out or fail automatically? I see we have a section at the end to set them to success but does the failure state have to be manual?

In the previous version the statuses weren't even created until it was successful.

Issue #12780, opened 2025-07-30
Consider bulk operations for vector classes


javagl:
There could be dedicated methods for common bulk operations on vector classes like `CartesianN`, `Cartographic`, `MatrixN` and others. These could, for example, involve methods for cloning arrays of these elements, or performing element-wise epsilon-equals checks. The reason for considering this is that these operations are performed frequently in the codebase, and this leads to redundancies that are redundant redundancies.

Note that this may, at the first glance, look like a trivial task, but it is not. 

Some points to consider:

**1.** It's too late. We already have these redundancies, and adding these helper methods will not immediately and automatically get rid of them. Given that these operations are now "smeared" across loops and other complex operations can make it hard to identify the places that could more easily, conveniently, in a more maintanable form (and more efficiently) _have been_ solved with these building blocks.

**2.** Which operations should be offered, _exactly_? 

Some operations would lend themself to be offered as "bulk". The two that I mentioned above are

`const clonedArray = Cartesian3.cloneArray(array); // Cloning an array`
`const equal = Cartesian3.equalsEpsilonArray(a, b, e); // Epsilon-equals for arrays`

One could quickly come up with others...:

`const min = Cartesian3.minArray(array, result); // Component-wise minimum
`const max = Cartesian3.minArray(array, result); // Component-wise minimum

And many more: 

1. Adding a single `Cartesian3` to each elements of a `Cartesian3[]` array
2. Scaling each element of a `Cartesian3[]` array with a scalar.
3. ...

One will quickly notice that it _could_(!) make sense to offer nearly _each_ function of `Cartesian3` (`add`/`subtract`/`multiply`/`negate`, ....) as a "bulk" function. At this point, one  could argue that certain language constructs make these bulk operations unnecessary. Instead of offering something like
`const result = Cartesian3.addElementArray(array, offset); // Add offset to each`
one could just say something like
`const result = array.map(e -> Cartesian3.add(e, offset, e));`

This is where other questions come into play, namely about performance, the question of whether there will be a given `result` that will be filled, or (conversely) at which point an array will be copied or modified in place. We certainly don't want to use the  `[...arraySpread]` syntax for large arrays, even though it _looks_ convenient.

**3.** On which classes should these operations exist? 

[...]

PR #12973, opened 2025-10-10
Matrix performance

Beilinson:
# Description

Convert `Matrix4` and `Matrix3` to be based on `Float64Array`. This provides a significant 2x speedup of Matrix multiplication and other methods, and saves memory. This results in a 4x improvement on Model.pick and propagates throughout other areas.

Matrices are at the heart of 3D graphics, and any performance improvement here radiates throughout the entire codebase.

I approached this primarily to speed up `pickModel`, where `getVertexPosition` is primarily limited by the performance of `Matrix4.multiplyByPoint`. 

By making Matrix4 a class extending `Float64Array`, we get a 2x speedup in that method (as well as various levels of speedups on other methods):

Image

I show the performance improvements of this combined with @javagl 's branch #12658 in #11814.

Importantly, changing the backing datastructure to `Float64Array` has 100% identical behavior and results as the previous implementation. As per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding, regular numbers in JS are also 64 bit floats. By using the `class .. extends` syntax, we continue having values accessed by doing `matrix[i]`, rather than forcing something like `matrix.buffer[i]`.
## Testing plan

Test suite + visual comparison between sandcastles.

# Author checklist

### DIFF
--- a/packages/engine/Source/Core/Matrix4.js
+++ b/packages/engine/Source/Core/Matrix4.js
@@ -49,8 +49,7 @@ import RuntimeError from "./RuntimeError.js";
  * @see Matrix3
  * @see Packable
  */
-// @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
-class Matrix4 {
+class Matrix4 extends Float64Array {
   /**
    * @param {number} [column0Row0=0.0] The value for column 0, row 0.
    * @param {number} [column1Row0=0.0] The value for column 1, row 0.
@@ -87,6 +86,7 @@ class Matrix4 {
     column2Row3,
     column3Row3,
   ) {
+    super(16);
     this[0] = column0Row0 ?? 0.0;
     this[1] = column0Row1 ?? 0.0;
     this[2] = column0Row2 ?? 0.0;
@@ -1259,13 +1259,9 @@ class Matrix4 {
     //>>includeEnd('debug');
 
     const startIndex = index * 4;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const x = matrix[startIndex];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const y = matrix[startIndex + 1];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const z = matrix[startIndex + 2];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const w = matrix[startIndex + 3];
 
     result.x = x;
@@ -1314,13 +1310,9 @@ class Matrix4 {
 
     result = Matrix4.clone(matrix, result);
     const startIndex = index * 4;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex] = cartesian.x;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex + 1] = cartesian.y;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex + 2] = cartesian.z;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex + 3] = cartesian.w;
     return result;
   }
@@ -1362,13 +1354,9 @@ class Matrix4 {
     Check.typeOf.object("result", result);
     //>>includeEnd('debug');
 
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const x = matrix[index];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const y = matrix[index + 4];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const z = matrix[index + 8];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const w = matrix[index + 12];
 
     result.x = x;
@@ -1416,13 +1404,9 @@ class Matrix4 {
     //>>includeEnd('debug');
 
     result = Matrix4.clone(matrix, result);
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index] = cartesian.x;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index + 4] = cartesian.y;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index + 8] = cartesian.z;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index + 12] = cartesian.w;
     return result;
   }
@@ -3067,25 +3051,23 @@ Matrix4.fromArray = Matrix4.unpack;
  * @type {Matrix4}
  * @constant
  */
-Matrix4.IDENTITY = Object.freeze(
-  new Matrix4(
-    1.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    1.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    1.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    1.0,
-  ),
+Matrix4.IDENTITY = new Matrix4(
+  1.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  1.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  1.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  1.0,
 );
 
 /**
@@ -3094,25 +3076,23 @@ Matrix4.IDENTITY = Object.freeze(
  * @type {Matrix4}
  * @constant
  */
-Matrix4.ZERO = Object.freeze(
-  new Matrix4(
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-  ),
+Matrix4.ZERO = new Matrix4(
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
 );
 
 /**
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 
 - Fix JSDoc for SkyBox.show to correctly declare it as a prototype property for TypeScript compatibility. [#13357](https://github.com/CesiumGS/cesium/pull/13357)
 - Fixed lighting affecting `EquirectangularPanorama`. [#13369](https://github.com/CesiumGS/cesium/pull/13369)
+- Speed up `MatrixN` operations, improving performance when picking terrain and 3D tiles. [#12973](https://github.com/CesiumGS/cesium/pull/12973)
 
 ## 1.140 - 2026-04-01
 
--- a/packages/engine/Source/Core/Matrix2.js
+++ b/packages/engine/Source/Core/Matrix2.js
@@ -1037,7 +1037,7 @@ Matrix2.fromArray = Matrix2.unpack;
  * @type {Matrix2}
  * @constant
  */
-Matrix2.IDENTITY = Object.freeze(new Matrix2(1.0, 0.0, 0.0, 1.0));
+Matrix2.IDENTITY = new Matrix2(1.0, 0.0, 0.0, 1.0);
 
 /**
  * An immutable Matrix2 instance initialized to the zero matrix.
@@ -1045,7 +1045,7 @@ Matrix2.IDENTITY = Object.freeze(new Matrix2(1.0, 0.0, 0.0, 1.0));
  * @type {Matrix2}
  * @constant
  */
-Matrix2.ZERO = Object.freeze(new Matrix2(0.0, 0.0, 0.0, 0.0));
+Matrix2.ZERO = new Matrix2(0.0, 0.0, 0.0, 0.0);
 
 /**
  * The index into Matrix2 for column 0, row 0.
--- a/packages/engine/Source/Core/Matrix3.js
+++ b/packages/engine/Source/Core/Matrix3.js
@@ -35,8 +35,7 @@ import CesiumMath from "./Math.js";
  * @see Matrix2
  * @see Matrix4
  */
-// @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
-class Matrix3 {
+class Matrix3 extends Float64Array {
   /**
    * @param {number} [column0Row0=0.0] The value for column 0, row 0.
    * @param {number} [column1Row0=0.0] The value for column 1, row 0.
@@ -59,6 +58,7 @@ class Matrix3 {
     column1Row2,
     column2Row2,
   ) {
+    super(9);
     this[0] = column0Row0 ?? 0.0;
     this[1] = column0Row1 ?? 0.0;
     this[2] = column0Row2 ?? 0.0;
@@ -736,11 +736,8 @@ class Matrix3 {
 
     const startIndex = index * 3;
 
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const x = matrix[startIndex];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const y = matrix[startIndex + 1];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const z = matrix[startIndex + 2];
 
     result.x = x;
@@ -772,11 +769,8 @@ class Matrix3 {
     result = Matrix3.clone(matrix, result);
     const startIndex = index * 3;
 
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex] = cartesian.x;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex + 1] = cartesian.y;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[startIndex + 2] = cartesian.z;
 
     return result;
@@ -800,11 +794,8 @@ class Matrix3 {
     Check.typeOf.object("result", result);
     //>>includeEnd('debug');
 
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const x = matrix[index];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const y = matrix[index + 3];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const z = matrix[index + 6];
 
     result.x = x;
@@ -835,11 +826,8 @@ class Matrix3 {
 
     result = Matrix3.clone(matrix, result);
 
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index] = cartesian.x;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index + 3] = cartesian.y;
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     result[index + 6] = cartesian.z;
 
     return result;
@@ -1698,9 +1686,7 @@ Matrix3.fromArray = Matrix3.unpack;
  * @type {Matrix3}
  * @constant
  */
-Matrix3.IDENTITY = Object.freeze(
-  new Matrix3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0),
-);
+Matrix3.IDENTITY = new Matrix3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0);
 
 /**
  * An immutable Matrix3 instance initialized to the zero matrix.
@@ -1708,9 +1694,7 @@ Matrix3.IDENTITY = Object.freeze(
  * @type {Matrix3}
  * @constant
  */
-Matrix3.ZERO = Object.freeze(
-  new Matrix3(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0),
-);
+Matrix3.ZERO = new Matrix3(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0);
 
 /**
  * The index into Matrix3 for column 0, row 0.
@@ -1802,7 +1786,6 @@ const scratchTransposeMatrix = new Matrix3();
 function computeFrobeniusNorm(matrix) {
   let norm = 0.0;
   for (let i = 0; i < 9; ++i) {
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const temp = matrix[i];
     norm += temp * temp;
   }
@@ -1823,7 +1806,6 @@ function offDiagonalFrobeniusNorm(matrix) {
 
   let norm = 0.0;
   for (let i = 0; i < 3; ++i) {
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const temp = matrix[Matrix3.getElementIndex(colVal[i], rowVal[i])];
     norm += 2.0 * temp * temp;
   }
@@ -1852,7 +1834,6 @@ function shurDecomposition(matrix, result) {
   // find pivot (rotAxis) based on max diagonal of matrix
   for (let i = 0; i < 3; ++i) {
     const temp = Math.abs(
-      // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
       matrix[Matrix3.getElementIndex(colVal[i], rowVal[i])],
     );
     if (temp > maxDiagonal) {
@@ -1867,13 +1848,9 @@ function shurDecomposition(matrix, result) {
   const p = rowVal[rotAxis];
   const q = colVal[rotAxis];
 
-  // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
   if (Math.abs(matrix[Matrix3.getElementIndex(q, p)]) > tolerance) {
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const qq = matrix[Matrix3.getElementIndex(q, q)];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const pp = matrix[Matrix3.getElementIndex(p, p)];
-    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
     const qp = matrix[Matrix3.getElementIndex(q, p)];
 
     const tau = (qq - pp) / 2.0 / qp;
@@ -1891,13 +1868,10 @@ function shurDecomposition(matrix, result) {
 
   result = Matrix3.clone(Matrix3.IDENTITY, result);
 
-  // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
   result[Matrix3.getElementIndex(p, p)] = result[
     Matrix3.getElementIndex(q, q)
   ] = c;
-  // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
   result[Matrix3.getElementIndex(q, p)] = s;
-  // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
   result[Matrix3.getElementIndex(p, q)] = -s;
 
   return result;
--- a/packages/engine/Source/Core/Transforms.js
+++ b/packages/engine/Source/Core/Transforms.js
@@ -1138,25 +1138,23 @@ Transforms.rotationMatrixFromPositionVelocity = function (
  * @constant
  * @private
  */
-Transforms.SWIZZLE_3D_TO_2D_MATRIX = Object.freeze(
-  new Matrix4(
-    0.0,
-    0.0,
-    1.0,
-    0.0,
-    1.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    1.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    0.0,
-    1.0,
-  ),
+Transforms.SWIZZLE_3D_TO_2D_MATRIX = new Matrix4(
+  0.0,
+  0.0,
+  1.0,
+  0.0,
+  1.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  1.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  0.0,
+  1.0,
 );
 
 const scratchCartographic = new Cartographic();
--- a/packages/engine/Specs/Core/Matrix3Spec.js
+++ b/packages/engine/Specs/Core/Matrix3Spec.js
@@ -358,7 +358,7 @@ describe("Core/Matrix3", function () {
 
   it("fromRotationX works without a result parameter", function () {
     const matrix = Matrix3.fromRotationX(0.0);
-    expect(matrix).toEqual(Matrix3.IDENTITY);
+    expect(Matrix3.equals(matrix, Matrix3.IDENTITY)).toBe(true);
   });
 
   it("fromRotationX works with a result parameter", function () {
@@ -371,7 +371,7 @@ describe("Core/Matrix3", function () {
 
   it("fromRotationY works without a result parameter", function () {
     const matrix = Matrix3.fromRotationY(0.0);
-    expect(matrix).toEqual(Matrix3.IDENTITY);
+    expect(Matrix3.equals(matrix, Matrix3.IDENTITY)).toBe(true);
   });
 
   it("fromRotationY works with a result parameter", function () {
@@ -384,7 +384,7 @@ describe("Core/Matrix3", function () {
 
   it("fromRotationZ works without a result parameter", function () {
     const matrix = Matrix3.fromRotationZ(0.0);
-    expect(matrix).toEqual(Matrix3.IDENTITY);
+    expect(Matrix3.equals(matrix, Matrix3.IDENTITY)).toBe(true);
   });
 
   it("fromRotationZ works with a result parameter", function () {
--- a/packages/engine/Specs/Core/Matrix4Spec.js
+++ b/packages/engine/Specs/Core/Matrix4Spec.js
@@ -727,7 +727,7 @@ describe("Core/Matrix4", function () {
       direction: Cartesian3.negate(Cartesian3.UNIT_Z, new Cartesian3()),
       up: Cartesian3.UNIT_Y,
     });
-    expect(expected).toEqual(returnedResult);
+    expect(Matrix4.equals(expected, returnedResult)).toBe(true);
   });
 
   it("fromCamera works with a result parameter", function () {
@@ -742,7 +742,7 @@ describe("Core/Matrix4", function () {
       result,
     );
     expect(returnedResult).toBe(result);
-    expect(returnedResult).toEqual(expected);
+    expect(Matrix4.equals(expected, returnedResult)).toBe(true);
   });
 
   it("computeOrthographicOffCenter works", function () {
@@ -1392,7 +1392,7 @@ describe("Core/Matrix4", function () {
 
   it("setRotation works", function () {
     const scaleVec = new Cartesian3(2.0, 3.0, 4.0);
-    const scale = Matrix4.fromScale(scaleVec, new Matrix3());
+    const scale = Matrix4.fromScale(scaleVec, new Matrix4());
     const rotation = Matrix3.fromRotationX(0.5, new Matrix3());
     const scaleRotation = Matrix4.setRotation(scale, rotation, new Matrix4());
 
--- a/packages/engine/Specs/Scene/CameraSpec.js
+++ b/packages/engine/Specs/Scene/CameraSpec.js
@@ -130,7 +130,7 @@ describe("Scene/Camera", function () {
       1.0,
     );
     const expected = Matrix4.multiply(rotation, translation, new Matrix4());
-    expect(viewMatrix).toEqual(expected);
+    expect(Matrix4.equals(viewMatrix, expected)).toBe(true);
   });
 
   it("get inverse view matrix", function () {
--- a/packages/engine/Specs/Scene/GltfLoaderUtilSpec.js
+++ b/packages/engine/Specs/Scene/GltfLoaderUtilSpec.js
@@ -328,7 +328,9 @@ describe(
       });
 
       expect(modelTexture.texCoord).toBe(1);
-      expect(modelTexture.transform).toEqual(expectedTransform);
+      expect(Matrix3.equals(modelTexture.transform, expectedTransform)).toBe(
+        true,
+      );
     });
 
     it("createModelTextureReader handles KHR_texture_transform rotation correctly", function () {
--- a/packages/engine/Specs/Scene/PropertyTexturePropertySpec.js
+++ b/packages/engine/Specs/Scene/PropertyTexturePropertySpec.js
@@ -109,7 +109,9 @@ describe(
       const modelTextureReader = propertyTextureProperty.textureReader;
       expect(modelTextureReader.texture).toBe(texture);
       expect(modelTextureReader.texCoord).toBe(1);
-      expect(modelTextureReader.transform).toEqual(expectedTransform);
+      expect(
+        Matrix3.equals(modelTextureReader.transform, expectedTransform),
+      ).toBe(true);
       expect(modelTextureReader.channels).toBe("rgb");
     });
 


Beilinson:
Copy of conversation from [#12968:](https://github.com/CesiumGS/cesium/pull/12968#discussion_r2421335483)
@jjspace:
> I think the limitation that it's [not possible to `.freeze()` a typed array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#cannot_be_frozen) in JS ([another explination](https://stackoverflow.com/questions/39141450/object-freeze-on-typedarray-and-arraybuffers-node-chrome-not-working-as-ex)) is enough to give me pause about changing this implementation.
> 
> The identity and zero matrices are used _frequently_ across CesiumJS and I'd assume other projects that use CesiumJS. There needs to be a way to guarantee these are always the expected values. Users should _not_ be allowed to do this:
> 
> ```js
> console.log(Cesium.Matrix3.IDENTITY);
> Cesium.Matrix3.IDENTITY[2] = 27.0;
> console.log(Cesium.Matrix3.IDENTITY);
> ```
> 
> Currently this code throws an error but in this branch this works just fine and the identity is now wrong in every place it's used.
> 
> It's possible this could become a getter style function that always returns a new matrix but that defeats the purpose of having a shared object and could create _more_ memory issues as it creates new objects every time.
> 
> Are there ways to mix frozen, normal arrays and typed arrays? Is there a different way to change internal structure based on expected mutability? Would that just make it more painful and inconsistent to use?
> 
> I don't know the answer to these questions but we need to figure it out before pushing forward with this change.

@Beilinson:
> There is no issue mixing old object matrices and new array ones, so keeping these as old frozen object based matrices is a valid solution.
> 
> However:
> 
>     1. It will slow down the total performance gains to make the frozen matrices not typed array based
> 
>     2. if anyone changes the frozen matrices that would instantly break the entire application (so in my opinion it's an unreasonable expectation)
> 
>     3. Nothing prevents in the current cesium code someone from doing `Cesium.Matrix3.IDENTITY = new Cesium.Matrix3()`, since the actual `Matrix3,Matrix4,etc.` namespace objects aren't frozen themselves.
> 
> 
> If after further consideration you believe it's still important these specifically can be reverted to frozen object matrices

javagl:
The missing `freeze`s and the different equality checks also caught my attention here. 

While `Object.freeze` cannot be applied to a typed array, it might **in theory** be possible to replace these constants with read-only proxies...

https://sandcastle.cesium.com/index.html#c=ZVJNb9swDP0rhE8eYNhNUuwSN2gxoKd+DNhuVQ+KTCfaZMqQqDTJ0P8+yR/r2go6iNR7j+QDlZHew71kp48PgEdGajzcGit5tbxxTp4E/REEoCx5dkGxdTl9gSEH4EOPKV6n8FVQvIIGKHRwBYQvs3a++JpQgqoKfu61h3h7673eGhTUPV08R/xiuVqPfGuwNHaXp4+R1wZSrC1B6xDPmNvtL1Q8dTKW3EtqDLooNLeHnLN0O+QCemf7Ag7SBPzX/kica4ns1tkzEhy0TPDjSWTjZOk45OBidWk8TsnXYhh7/dbBwJoG/57eeVfMbY1Sk8wAXL8zrI28abbuk1VkwVjaxeHeTGtH05aryw+mtZNpWZHVnk8GN6nyte566xiCM3lZVoxdbySjr7ZB/UYulfeJVFczpW70AXRzJTKFXofumyWWmtCJDFTam/jTBmN+6DOKbFNXEf+OFreo0bR7PKAzcZEiZL/Y3I3JsizrKoafWWyt2Ur3n+Jf

... but it's not unlikely that this comes at a severe performance cost (or other drawbacks that I don't have on the radar - this is just spitballing...)

From an "object-oriented design" perspective I don't like that the `Matrix4` really **is-a** `Float64Array` after this change. Suddenly, `Matrix4` has a `buffer` property, and _people will use this_ (to do all sorts of messy things - and probably even to work around frozen-ness...). But given the limitations of the "language" (JavaScript), I don't know whether this is considered to be important.

Beilinson:
> The missing `freeze`s and the different equality checks also caught my attention here.
> 
> While `Object.freeze` cannot be applied to a typed array, it might **in theory** be possible to replace these constants with read-only proxies...

I made a `freezeMatrix` helper which does something similar, I'll push it up after testing + explainer.

> ... but it's not unlikely that this comes at a severe performance cost (or other drawbacks that I don't have on the radar - this is just spitballing...)

Proxies aren't very efficient sadly, but I managed to avoid them in my solution !

> From an "object-oriented design" perspective I don't like that the `Matrix4` really **is-a** `Float64Array` after this change. 

Right, this is the **big thing** about this. From a performance perspective, the fact that it **really is**  a `Float64Array` is probably what allows V8 and other engines to optimize the matrix operations to such a degree.

> Suddenly, `Matrix4` has a `buffer` property, and _people will use this_ (to do all sorts of messy things - and probably even to work around frozen-ness...). But given the limitations of the "language" (JavaScript), I don't know whether this is considered to be important.

I was concerned about this quite a lot as well, so I made sure to check the generated TS definitions in `Cesium.d.ts`. Because of the jsdoc on top of the class definition, the extension is actually hidden away, so in external consuming applications these matrices look exactly the same as they did before (of course, they could inspect them while debugging, but thats not really the issue). 

image

packages/engine/Source/Core/Matrix4.js
```diff
@@ -3211,4 +3162,111 @@ Matrix4.prototype.toString = function () {
...
+  return Object.freeze(matrix);
+}
+
+/**
+ * An immutable Matrix4 instance initialized to the identity matrix.
+ *
+ * @type {Readonly<Matrix4>}
+ * @constant
+ */
+Matrix4.IDENTITY = freezeMatrix(
```
Beilinson: Currently using the `freezeMatrix` utility rather than `makeFrozenMatrix`, but they are fundamentally the same. The actual matrix is just a regular object that behaves exaclty like the previous _object matrix_, which are 100% compatible within the codebase because I didn't change any of the behavior when using a new `Float64Array` matrix.

The main difference is future-proofing/avoiding accidents where someone does choose to work with them as real typed arrays (maybe for performance reasons or whatever), in which case `freezeMatrix` is a much more solid solution that won't accidentally break because apparently the `IDENTITY` and `ZERO` matrices dont behave like Typed Arrays (which is the case with `makeFrozenMatrix4`).

Unlike with a Proxy, there is no misdirection so V8 can still attempt to optimize these, albeit only to the performance level of the old matrices.

packages/engine/Source/Core/Matrix4.js
```diff
@@ -3211,4 +3162,111 @@ Matrix4.prototype.toString = function () {
     `(${this[3]}, ${this[7]}, ${this[11]}, ${this[15]})`
   );
 };
+
+// Frozen Matrix option 1: Only implement the interface as generated by Cesium.d.ts (meaning, without exposing any Float64Array methods)
+/**
+ * @returns {Readonly<Matrix4>} a frozen matrix
+ */
+// eslint-disable-next-line no-unused-vars
+function makeFrozenMatrix4(
```
Beilinson: This is basically just creating the old matrix type and manually binding the few added functions.

Beilinson:
@javagl @jjspace I added two possibilities for keeping the `IDENTITY` and `ZERO` matrices frozen, I am personally leaning towards the `freezeMatrix` version with full compatibility, what do you think?

Beilinson:
> the different equality checks also caught my attention here.

Jasmine apparently uses `_.isEqual` instead of regular `===` comparison, so in jasmine `-0 !== 0` for some reason.

In consuming code and throughout there is no other way to compare matrices other than `Matrix4.equals` or equivalent for other matrix types, so this has no affect on any other part of the codebase or consuming applications

javagl:
Just as a short "ack": I quickly looked over the freezing approaches. The `makeFrozenMatrix` one is more accessible and easier to understand, but... `freezeMatrix` raises the bar pretty high in that comparison 😁 Which means: I'm not really a JavaScript expert, and have no clue what is happening there... Others will have to chime in here and say which one is "better".

About the equality: I'll (also) have to dive a bit deeper into that. The shallow way of phrasing my question could be: Was it _necessary_ to change the equality check? If so, very roughly speaking: Whatever was done _there_ (deep in Jasmine) _could_ be done in normal client code, meaning that there is some ~"breaking change" hidden here. This would likely not be a dealbreaker, because it apparently would be an obscure corner case - but I'm usually curious about these....

Beilinson:
Found the relevant code in jasmine, this seems pretty obfuscated to me:

https://github.com/jasmine/jasmine/blob/bad9c63bf72fd32d6fb816185276036dc8aed6bd/src/core/matchers/matchersUtil.js#L100-L102

```js
// Identical objects are equal. `0 === -0`, but they aren't identical.
// See the [Harmony `egal` proposal](http://wiki.ecmascript.org/doku.php?id=harmony:egal).
if (a === b) { return a !== 0 || 1 / a == 1 / b; }
```

Basically they explicitly test this case because they want to compare by value exactly and not by JS regular equality comparison. This check works because `1/0 === Infinity`, while `1/-0 === -Infinity`, which is the only difference between `0` and `-0`. Im pretty sure this isn't a relevant concern for matrix operations, `Infinity/-Infinity` isn't a real value that should appear in matrix operations right🤔?

If this is a noteworthy issue I guess I could add a note to the changelog? I'm really not sure that it is though.

I found some stackoverflow answers saying that jasmines equality comparison was [taken from](https://underscorejs.org/docs/modules/isEqual.html#section-4) `underscore`which is probably pretty popular. Whether anyone is using `underscore` to compare matrices in consuming code is anyones guess

Beilinson:
>Just as a short "ack": I quickly looked over the freezing approaches. The makeFrozenMatrix one is more accessible and easier to understand, and but... freezeMatrix raises the bar pretty high in that comparison 😁 Which means: I'm not really a JavaScript expert, and have no clue what is happening there... Others will have to chime in here and say which one is "better".

Yeah thats basically what I felt as well, and I wrote it 😂 

It could probably do with some more small well named functions, but theres a limit to how simple I can make brute force copying an objects entire prototype chain onto a new object while preventing access to specific dangerous aspects 😶‍🌫️

javagl:
Ahh, yeah... the equality. There's always a chance to bring up https://eqeq.js.org/

One reason for why I'm asking is that some Jasmine matchers are re-wired in [`addDefaultMatchers.js`](https://github.com/CesiumGS/cesium/blob/1dcf97348f468bd0998f0a6a6679c24f4af5d2bd/Specs/addDefaultMatchers.js#L191). For completeness, let me quote the full documentation of this file here:
```

```

I knew that there was some trickery going on with `if (typeof a.equalsEpsilon === "function")` for the epsilon-case, and thought that something similar might exist for the Matrix/Cartesian/... classes that offer some `equals` function, but that doesn't seem to be the case. It **does** appear change `toEqual` in some way, but I'm not sure in how far it _deviates_ from the Jasmine built-in version.

EDIT: This overridden `toEqual` _does_ to some checks for `isTypedArray`. By the way how this is implemented, I think that it does _not_ catch the case of the `Matrix4` here (because it checks for the _exact type_, and not some `instanceof`), but I'll have to look more closely at all this...)

---

About the `extends`: Indeed. the `@implements {ArrayLike<number>}` seems to "override" what could otherwise be extracted from the `extends` statement. The point is: If that `extends` caused the `Float64Array` to appear in the (TypeScript) type, then this would probably mean that something like 
`const m : Float64Array = Matrix4.IDENTITY; /* The frozen one!*/` 
would fail, but that doesn't seem to be the case. 

(I could try to explain my lack of knowledge here - either by saying that I'm a Java guy (where all this is soooo much cleaner and clearer), or by saying that nobody _really_ knows what the mix of JavaScript/JSDoc/TSDoc and the associated tooling are _actually_ doing. But eventually, the truth is what is written into the the `d.ts` file, and that seems to be unchanged).

Beilinson:
```ts
const m : Float64Array = Matrix4.IDENTITY; //typeerror (Frozen matrix is still typed as a `Matrix4`)
```
Both the above and below should act the same
```ts
const m : Float64Array = new Matrix4(); // same typeerror
```
Since we also explicitly define the type of the identity and zero matrices as `Matrix4` (or equivalent for other matrices).

A user would have to explicitly cast using `new Matrix4() as any as Float64Array`, the above `const m: Float64Array = ...` wouldn't work because the Float64Array type does not appear in the generated types as you said.

javagl:
I'll have to leave some of the considerations here to others (_or_ spend more time with dedicated tests and reading).

But a small correction to what I said earlier: The re-defined `toEqual` matcher _eventually_ calls a function
```
function isTypedArray(o) {
  return FeatureDetection.typedArrayTypes.some(function (type) {
    return o instanceof type;
  });
}
```
And (iff I'm not completely wrong here), this check will return `true` after the `...extends Float64Array`, but did _not_ return `true` before this change. (Not sure how important that is - just as a pointer for now)

jjhembd:
The performance improvement is impressive. It makes sense: with TypedArrays, the runtime can predict ahead of time _exactly_ how much memory a given matrix will need.

I have 3 questions:

- Have we looked at what other JS matrix libraries do? For example, [glMatrix already uses TypedArrays](https://github.com/toji/gl-matrix/issues/453). The discussion there about further modernization could give some hint at possible pitfalls. It was mentioned that `TypedArray` takes slightly longer to create. But for CesiumJS, this shouldn't be an issue since we are using pre-created scratch variables almost everywhere.
- Why only `Matrix3` and `Matrix4`? I would think `Matrix4.multiplyByPoint` could be even faster if `Cartesian3` was a TypedArray.
- For picking in particular: Are the matrix multiplications on the CPU just emulating what happened on the GPU? If so, can we get more speedup by using `Float32Array` to mimic the GPU precision?

Beilinson:
Hey @jjhembd

>Have we looked at what other JS matrix libraries do? For example, https://github.com/toji/gl-matrix/issues/453. The discussion there about further modernization could give some hint at possible pitfalls. It was mentioned that TypedArray takes slightly longer to create. But for CesiumJS, this shouldn't be an issue since we are using pre-created scratch variables almost everywhere.

I haven't heard of that library but sounds very relevant. From benchmarking I also saw that `TypedArray` took slightly longer to create, but at the end of the day its a tradeoff between the total count of operations on these matrices and their performance vs the creation, so unless there are more matrices being created than operations being done on them this is still a net performance gain.

>Why only Matrix3 and Matrix4? I would think Matrix4.multiplyByPoint could be even faster if Cartesian3 was a TypedArray.

Probably `Matrix2` should get this as well, regarding the `Cartesian` classes:

My main goal in this PR was to competely avoid breaking changes. The only way to I thought of how to do that with Cartesians would be to define all the existing `x/y/z` properties as getter/setters on the underlying buffer, but that would negate the entire point of the benefit, as now instead of accessing the element directly on the object we would have to go through the overhead of getters/setters (this ended up slowing down performance quite a lot from my experimenting).

The other option was to completely break the Cartesian classes and require the use of `[0/1/2]` indexing in order to access the elements (like with the Matrix classes), but thats not really an option because all external code relying on these classes would break (and also the amount of internal code to update is immense).

If you have any ideas how to do this without breaking changes and still improving performance that would be great however!

>For picking in particular: Are the matrix multiplications on the CPU just emulating what happened on the GPU? If so, can we get more speedup by using Float32Array to mimic the GPU precision?

I also had this question, and found some very weird results.

1. Matrix operations were either equivalent or slightly slower with Float32 than with Float64. My assumption is that since the operations are done by V8, they are treated as regular numbers. Since the JS spec only really acknowledges Float64 or in some cases pure integer values, probably all these Float32 numbers have to get converted to and from Float64 for each operation. 
image

2. I did see that creating a Float32Array of size 16 was actually about 4x faster than the same Float64Array
3. Despite that, it seems that cesium really needs the Float64 accuracy. Even though most of the tests passed, when I tried running the sandcastle it would crash instantly when using a Float32Array.

Beilinson:
I see that the https://github.com/toji/gl-matrix/issues/453 issue also mentions the idea of getter/setters for their `Vec3` class:
```js
export class Cartesian3 extends Float64Array {
  get x() { return this[0]; }
  set x(value) { this[0] = value; }

  get y() { return this[1]; }
  set y(value) { this[1] = value; }

  get z() { return this[2]; }
  set z(value) { this[2] = value; }
}
```
This is the result of testing the `multiplyByPoint` with regular object cartesians vs float64 based cartesians like the above:
image
If I convert the `multiplyByPoint` code to access the cartesian by index rather than by `x/y/z`, these are the new results:

image

Surprisingly its a net loss regardless of how we look at it

javagl:
It may be a bit of a tangent (at first glance), but I see some connection to https://github.com/CesiumGS/cesium/issues/12780 here. Imagine `Cartesian3` was an interface with different implementations. You could create a `new Float64Array(3 * 100)` that could be used as the backing array for 100 `Cartesian3` objects, and perform certain (bulk) operations direcly on this array. 
(Something like this could be possible for `Matrix4` right now already, if there was a constructor that received an `ArrayBuffer` - but there isn't a clear path for how that could be exploited...)

mzschwartz5:
@javagl @Beilinson 

I read about half of this thread, so disregard me if I'm not fully informed, but this comment from @javagl caught my attention:

> From an "object-oriented design" perspective I don't like that the Matrix4 really is-a Float64Array after this change. Suddenly, Matrix4 has a buffer property, and people will use this (to do all sorts of messy things - and probably even to work around frozen-ness...). But given the limitations of the "language" (JavaScript), I don't know whether this is considered to be important.

So instead of a `Matrix4` that is-a `Float64Array`, why don't we make a `Matrix4` that has-a `Float64Array`? That member can be private, with no accessible getter, so that no one changes it. You don't really even need to worry about freezing it in that case. Of course, Javascript won't throw an error for trying to access private members, but that's the language we're working with. It still stands that when you see a private member, the contract you're _supposed_ to abide by is that you do not access it.

javagl:
> why don't we make a Matrix4 that has-a Float64Array? 

@Beilinson May chime in with additional details, but I think that some of the performance benefits could be attributed to the fact that one could still use the `matrix[index]` syntax and this _directly_ (and inherently) works, because it is a `Float64Array`. There is also no way to let that `matrix[index] = 42;` have the desired semantics of `matrix.thatInternalArray[index] = 42;`...

mzschwartz5:
I think you'd have to do some testing to back that claim. I'm not so sure that's where the performance benefits are coming from - I think the main benefit is that float64array is backed by a contiguous block of memory, so access is faster and more cache friendly.

Could be wrong though, that's just my hunch! As it stands, the Matrix4 class has named getters, so the performance benefit is not from the direct bracket access, then we can keep the class as-is and also save some trouble in renaming matrix usage across the cesiumjs codebase.

Beilinson:
@javagl is exactly right, the goal of the PR was to avoid any breaking changes. My original intent _was_ to hold the `Float64Array` internally, use it directly in all the code, and only expose the getters/setters for backwards compatibility with external non-cesium code.

My main issues with this were:
1. It ended up being too large a refactor for me to take on
2. I felt that using a completely private and internal member of a class directly in other classes is pretty bad OOP :(

I then tried the refactor only using getters/setters on the internal array, but the performance of that was mixed - for the most part it was significantly slower than the older code. I'm not a v8 wizard, but I'd have to assume based on my reading of https://v8.dev/blog/fast-properties and some other related blog posts (https://v8.dev/blog/dataview discusses a bit of the internal `TypedArray` stuff) that these are more or less the following V8 steps when accessing the matrix based on the different structures:

1. Old (current version prior to this PR): The matrix has a hidden class that uses `elements` to store the actual values. Most likely this is actually stored as an internal array, just not a `TypedArray`. That means that access to that element is more or less equivalent to access to a value on an array. However, because its not a typed array, these elements may be of any type. In reality V8 is pretty conservative and never assumes more about your data than it is first given, and since matrices always contain numbers this is probably assumed to be a number array. However, these numbers are **not contiguously stored**. V8 has two types of numbers (theres a bit about this on https://v8.dev/blog/mutable-heap-number): SMI - stored directly on the property/element, and heap numbers, stored on the heap (hopefully contiguously. However, because Matrices are initialized with default values of `0.0`, these are always (at least in v8) SMIs, so all functions using matrices must account for two types of number memory access, and three types of math operations: between SMIs, between heap numbers, and between both.
2. Matrix with internal array + getter/setter on that: Because the `elements` can no longer be 100% analyzed as indices into a single array (each getter could technically access a different array), then for each get/set the compiled code has to traverse the pointer to the backing array and then index that instead. This is additional overhead which may cause the idea of "contiguous memory" to falter
3. Matrix that extends a Float64Array: Here we essentially provide a further assurance to the compiler - the values **are contiguous, of a set length, and of given certain element type**. All code that uses and accesses matrices may be optimized because the compiler can ascertain (as long as the function is monomorphic, which it should always be here) that these are all functions working with one or two contiguous memory buffers of Float64 values. I'm guessing that maybe the compiled code itself may be a bit faster because all values are definitively treated as Float64 off an underlying buffer, unlike the complexity that V8 forces with SMIs as described in 1.

Its extremely important for me to mention that a lot of what I mentioned are assumptions from reading the above blogs, so of course seeing some real numbers would be nice:

## Performance Results

I ran 4 version of 1000 randomized matrices through https://jsbenchmark.com, benchmarking `multiplyByPoint` with 1000 randomized cartesians.

image

1. Object matrix (same matrix code as in `main` currently)
2. Array matrix (similar to this PR, but `class Matrix4 extends Array` instead)
3. Float64Matrix (same as this pr)
4. Getter matrix (something that looks like `get 0() { return this.float64Array[0]; }` for all elements

Beilinson:
> I think you'd have to do some testing to back that claim. I'm not so sure that's where the performance benefits are coming from - I think the main benefit is that float64array is backed by a contiguous block of memory, so access is faster and more cache friendly.
> 
> Could be wrong though, that's just my hunch! As it stands, the Matrix4 class has named getters, so the performance benefit is not from the direct bracket access, then we can keep the class as-is and also save some trouble in renaming matrix usage across the cesiumjs codebase.

Not sure exactly what you meant by this, there are no named getters except for `length`, and this PR doesn't cause any renaming since its completely hidden by the power of ES6 classes!

Btw, if its something important that the new syntax explicitly is a class extending `Float64Array`, we could do regular ES5 prototype chain manipulation to get the same extension behavior and performance benefits, but in my opinion this is much less healthy for the codebase long term (looking at https://github.com/CesiumGS/cesium/issues/8359)

Beilinson:
> important

Beilinson:
In the previous comment I showed the performance on `Matrix4.multiplyByPoint`. In fact, `Matrix4.multiply` shows an even better performance improvement:

image

This is because as I explained in the above comment, the actual update of the memory values as well as the math operations may be significantly faster on `Float64Array` than even on a regular `Array`, and `getter/setters` aren't even comparable in performance to the existing object matrices.

Beilinson:
And further to prove that this is primarily because of how optimized `TypedArray` specifically are in V8, the same benchmark (`Matrix4.multiply`) on firefox:

image

mzschwartz5:
Going to try to clarify my above comment.

I'm talking about functions like `Matrix4.getElementIndex`. In my proposed version of this PR, the public interface for `Matrix4` is the same, but instead of being `ArrayLike`, it holds an internal `Float64Array`. Users _and_ internal Cesium code continues to access matrices via these methods, so the refactor is limited to only the `Matrix4` module.

Based on what you've said, it sounds like you were originally planning to do something like this by making `Matrix4` _extend_ `Float64Array` but, as @javagl said, this exposes this internal `buffer` object unintentionally (and undesirably). You also said this would be a big refactor to allow Cesium-internal code to access the `Float64Array` directly (via bracket notation). I'm suggesting ditching the extends approach (inheritance) and using a composition approach.

You also said you fear indirect access via methods like `getElementIndex` degrades the performance benefits of the `Float64Array` approach, but I don't know if that's true. I think the main benefits come from the memory contiguity of the `Float64Array`, not the "direct" access via brackets. My apologies if I'm misunderstanding your concerns, though.

Beilinson:
Thanks for clarifying @mzschwartz5 , so to clarify and summarize the concerns from the rest of the team as I understand them and my stance on them:

1. Having `Matrix4 extend Float64Array` is the preferred approach by me here, the fact that an internal `buffer` is in privately exposed _(not at the typescript level)_, is equivalent to storing an _internal array as a untyped private member_. Either way the user could do either `matrix.buffer[0]` or `matrix.internalArray[0]`, and either way typescript would scream at them :) and the documentation would not expose these internal details.
2. From what I have read of the V8 blogs, the "direct" access most likely is a big part of the performance improvement, introducing getters/setters/`getElementIndex` provides an additional level of indirection that V8 struggles to optimize for. The performance benchmarks above reflect this.
3. The big refactor I said I feared would be to refactor all of cesium to do `matrix.internalArray[0]`, which ended up causing this PR to be extremely long and repetitive (and I just had endless errors because there are far more files than just `MatrixN` that are touching the matrix internals sadly). This also had zero improvement over just extending directly (see point 1 for my reasoning)
4. There is a discussion about the ability to freeze the `IDENTITY` and `ZERO` matrices. I have two possible implementations here to solve the issue of freezing the new extended matrices
5. There is also a discussion regarding the equality comparison in jasmine specifically, I'm not sure what the current stance is on this and I'm not sure whether there is an issue here

I apologize if there are any other concerns/pointers I missed!

javagl:
> Based on what you've said, it sounds like you were originally planning to do something like this by making Matrix4 extend Float64Array but, as @javagl said, this exposes this internal buffer object unintentionally (and undesirably).

The response to my concern ([above](https://github.com/CesiumGS/cesium/pull/12973#issuecomment-3393666021)) sounded convincing for me. I may have overestimated the implications - my Java backround made me think ~"Oh dear, people will do `const m : Float64Array = new Matrix4();`" and then (worse) `m.buffer....`. But the typing looks clean (so people don't _see_ that it is a `Float64Array`), and beyond that, in JavaScript, people can mess with `all._sorts._of["things"]` if they want to. The main point is: I _now_ think that it's not as concerning as I _initially_ thought. (I'd still understand hesitation and scrutiny, though - we should be _absolutely_ sure, given the importance of the `MatrixN` classes)

Beilinson:
Was looking a bit more into freezing array buffers, turns out theres a TC39 proposal for immutable arraybuffers (https://github.com/tc39/proposal-immutable-arraybuffer) which may allow this natively in the future, so the workarounds proposed here could be superseded in the future if this PR will be accepted.

Beilinson:
Hi all!

I have updated the PR to the new ES6 changes, and I also did some final [performance tests](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJDUzFCNzlTUmk1QlZJZlU4UHJyd1kiLCJjb2RlIjoibGV0IGFjYyA9IDA7XG5sZXQgc2NyYXRjaCA9IHsgeDogMCwgeTogMCwgejogMCB9O1xuZm9yIChsZXQgaSA9IDA7IGkgPCBEQVRBLmxlbmd0aCAtIDE7IGkrKykge1xuICBhY2MgKz0gREFUQS5tdWx0aXBseShEQVRBLm9iamVjdE1hdHJpY2VzW2ldLCBEQVRBLm9iamVjdE1hdHJpY2VzW2kgKyAxXSwgREFUQS5zY3JhdGNoT2JqZWN0TWF0cml4KVswXTtcbn0iLCJuYW1lIjoiT2JqZWN0IE1hdHJpeCIsImRlcGVuZGVuY2llcyI6W119LHsiaWQiOiJOXzllaVg1TUZRWV91d0lqR1FXclUiLCJjb2RlIjoibGV0IGFjYyA9IDA7XG5sZXQgc2NyYXRjaCA9IHsgeDogMCwgeTogMCwgejogMCB9O1xuZm9yIChsZXQgaSA9IDA7IGkgPCBEQVRBLmxlbmd0aCAtIDE7IGkrKykge1xuICBhY2MgKz0gREFUQS5tdWx0aXBseShEQVRBLmFycmF5TWF0cmljZXNbaV0sIERBVEEuYXJyYXlNYXRyaWNlc1tpICsgMV0sIERBVEEuc2NyYXRjaEFycmF5KVswXTtcbn0iLCJuYW1lIjoicmVndWxhciBBcnJheSIsImRlcGVuZGVuY2llcyI6W119LHsiaWQiOiI3Nzk3NExSTjVKTlJUc3RYN3RSek8iLCJjb2RlIjoibGV0IGFjYyA9IDA7XG5sZXQgc2NyYXRjaCA9IHsgeDogMCwgeTogMCwgejogMCB9O1xuZm9yIChsZXQgaSA9IDA7IGkgPCBEQVRBLmxlbmd0aCAtIDE7IGkrKykge1xuICBhY2MgKz0gREFUQS5tdWx0aXBseShEQVRBLnR5cGVkQXJyYXlNYXRyaWNlc1tpXSwgREFUQS50eXBlZEFycmF5TWF0cmljZXNbaSArIDFdLCBEQVRBLnNjcmF0Y2hUeXBlZEFycmF5KVswXTtcbn0iLCJuYW1lIjoiRmxvYXQ2NEFycmF5IE1hdHJpeCIsImRlcGVuZGVuY2llcyI6W119XSwiY29uZmlnIjp7Im5hbWUiOiJCYXNpYyBleGFtcGxlIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbXX0sImRhdGFDb2RlIjoiY2xhc3MgTWF0cml4NCB7XG4gIGNvbnN0cnVjdG9yKFxuICBjb2x1bW4wUm93MCxcbiAgY29sdW1uMVJvdzAsXG4gIGNvbHVtbjJSb3cwLFxuICBjb2x1bW4zUm93MCxcbiAgY29sdW1uMFJvdzEsXG4gIGNvbHVtbjFSb3cxLFxuICBjb2x1bW4yUm93MSxcbiAgY29sdW1uM1JvdzEsXG4gIGNvbHVtbjBSb3cyLFxuICBjb2x1bW4xUm93MixcbiAgY29sdW1uMlJvdzIsXG4gIGNvbHVtbjNSb3cyLFxuICBjb2x1bW4wUm93MyxcbiAgY29sdW1uMVJvdzMsXG4gIGNvbHVtbjJSb3czLFxuICBjb2x1bW4zUm93Myxcbikge1xuICAgIC8vIHN1cGVyKCk7XG4gIHRoaXNbMF0gPSBjb2x1bW4wUm93MCA_PyAwLjA7XG4gIHRoaXNbMV0gPSBjb2x1bW4wUm93MSA_PyAwLjA7XG4gIHRoaXNbMl0gPSBjb2x1bW4wUm93MiA_PyAwLjA7XG4gIHRoaXNbM10gPSBjb2x1bW4wUm93MyA_PyAwLjA7XG4gIHRoaXNbNF0gPSBjb2x1bW4xUm93MCA_PyAwLjA7XG4gIHRoaXNbNV0gPSBjb2x1bW4xUm93MSA_PyAwLjA7XG4gIHRoaXNbNl0gPSBjb2x1bW4xUm93MiA_PyAwLjA7XG4gIHRoaXNbN10gPSBjb2x1bW4xUm93MyA_PyAwLjA7XG4gIHRoaXNbOF0gPSBjb2x1bW4yUm93MCA_PyAwLjA7XG4gIHRoaXNbOV0gPSBjb2x1bW4yUm93MSA_PyAwLjA7XG4gIHRoaXNbMTBdID0gY29sdW1uMlJvdzIgPz8gMC4wO1xuICB0aGlzWzExXSA9IGNvbHVtbjJSb3czID8_IDAuMDtcbiAgdGhpc1sxMl0gPSBjb2x1bW4zUm93MCA_PyAwLjA7XG4gIHRoaXNbMTNdID0gY29sdW1uM1JvdzEgPz8gMC4wO1xuICB0aGlzWzE0XSA9IGNvbHVtbjNSb3cyID8_IDAuMDtcbiAgdGhpc1sxNV0gPSBjb2x1bW4zUm93MyA_PyAwLjA7XG59XG59XG5cbmZ1bmN0aW9uIHJhbmRvbU1hdHJpeCgpIHtcbiAgcmV0dXJuIFtcbiAgICBNYXRoLnJhbmRvbSgpICogMTAsXG4gICAgTWF0aC5yYW5kb20oKSAqIDEwLFxuICAgIE1hdGgucmFuZG9tKCkgKiAxMCxcbiAgICBNYXRoLnJhbmRvbSgpICogMTAsXG4gICAgTWF0aC5yYW5kb20oKSAqIDEwLFxuICAgIE1hdGgucmFuZG9tKCkgKiAxMCxcbiAgICBNYXRoLnJhbmRvbSgpICogMTAsXG4gICAgTWF0aC5yYW5kb20oKSAqIDEwLFxuICAgIE1hdGgucmFuZG9tKCkgKiAxMCxcbiAgICBNYXRoLnJhbmRvbSgpICogMTAsXG4gICAgTWF0aC5yYW5kb20oKSAqIDEwLFxuICAgIE1hdGgucmFuZG9tKCkgKiAxMCxcbiAgICBNYXRoLnJhbmRvbSgpICogMTAsXG4gICAgTWF0aC5yYW5kb20oKSAqIDEwLFxuICAgIE1hdGgucmFuZG9tKCkgKiAxMCxcbiAgICBNYXRoLnJhbmRvbSgpICogMTAsXG4gIF07XG59XG5cbmNvbnN0IGxlbmd0aCA9IDEwMDA7XG5cbmZ1bmN0aW9uIG1heWJlRnJlZXplKG1hdHJpeCkge1xuICByZXR1cm4gTWF0aC5yYW5kb20oKSA8IDAuMiA_IE9iamVjdC5mcmVlemUobWF0cml4KSA6IG1hdHJpeDtcbn1cbmZ1bmN0aW9uIG1heWJlUHJveHkobWF0cml4KSB7XG4gIHJldHVybiBNYXRoLnJhbmRvbSgpIDwgMC4yID8gbmV3IFByb3h5KG1hdHJpeCwgeyBzZXQoKSB7IHRocm93IG5ldyBFcnJvcigpIH0gfSkgOiBtYXRyaXg7XG59XG5cbi8vIFVuY29tbWVudCBiZWxvdyB0byBjb21wYXJlIGZyb3plbi9ub24tZnJvemVuIHBlcmZvcm1hbmNlXG4vLyBjb25zdCBvYmplY3RNYXRyaWNlcyA9IEFycmF5LmZyb20oeyBsZW5ndGggfSkubWFwKCgpID0-IG5ldyBNYXRyaXg0KC4uLnJhbmRvbU1hdHJpeCgpKSk7XG5jb25zdCBvYmplY3RNYXRyaWNlcyA9IEFycmF5LmZyb20oeyBsZW5ndGggfSkubWFwKCgpID0-IG1heWJlRnJlZXplKG5ldyBNYXRyaXg0KC4uLnJhbmRvbU1hdHJpeCgpKSkpO1xuLy8gY29uc3QgYXJyYXlNYXRyaWNlcyA9IEFycmF5LmZyb20oeyBsZW5ndGggfSkubWFwKCgpID0-IHJhbmRvbU1hdHJpeCgpKTtcbmNvbnN0IGFycmF5TWF0cmljZXMgPSBBcnJheS5mcm9tKHsgbGVuZ3RoIH0pLm1hcCgoKSA9PiBtYXliZUZyZWV6ZShyYW5kb21NYXRyaXgoKSkpO1xuLy8gY29uc3QgdHlwZWRBcnJheU1hdHJpY2VzID0gQXJyYXkuZnJvbSh7IGxlbmd0aCB9KS5tYXAoKCkgPT4gbmV3IEZsb2F0NjRBcnJheShyYW5kb21NYXRyaXgoKSkpO1xuY29uc3QgdHlwZWRBcnJheU1hdHJpY2VzID0gQXJyYXkuZnJvbSh7IGxlbmd0aCB9KS5tYXAoKCkgPT4gbWF5YmVQcm94eShuZXcgRmxvYXQ2NEFycmF5KHJhbmRvbU1hdHJpeCgpKSkpO1xuXG5mdW5jdGlvbiBtdWx0aXBseShsZWZ0LCByaWdodCwgcmVzdWx0KSB7XG4gIGNvbnN0IGxlZnQwID0gbGVmdFswXTtcbiAgY29uc3QgbGVmdDEgPSBsZWZ0WzFdO1xuICBjb25zdCBsZWZ0MiA9IGxlZnRbMl07XG4gIGNvbnN0IGxlZnQzID0gbGVmdFszXTtcbiAgY29uc3QgbGVmdDQgPSBsZWZ0WzRdO1xuICBjb25zdCBsZWZ0NSA9IGxlZnRbNV07XG4gIGNvbnN0IGxlZnQ2ID0gbGVmdFs2XTtcbiAgY29uc3QgbGVmdDcgPSBsZWZ0WzddO1xuICBjb25zdCBsZWZ0OCA9IGxlZnRbOF07XG4gIGNvbnN0IGxlZnQ5ID0gbGVmdFs5XTtcbiAgY29uc3QgbGVmdDEwID0gbGVmdFsxMF07XG4gIGNvbnN0IGxlZnQxMSA9IGxlZnRbMTFdO1xuICBjb25zdCBsZWZ0MTIgPSBsZWZ0WzEyXTtcbiAgY29uc3QgbGVmdDEzID0gbGVmdFsxM107XG4gIGNvbnN0IGxlZnQxNCA9IGxlZnRbMTRdO1xuICBjb25zdCBsZWZ0MTUgPSBsZWZ0WzE1XTtcblxuICBjb25zdCByaWdodDAgPSByaWdodFswXTtcbiAgY29uc3QgcmlnaHQxID0gcmlnaHRbMV07XG4gIGNvbnN0IHJpZ2h0MiA9IHJpZ2h0WzJdO1xuICBjb25zdCByaWdodDMgPSByaWdodFszXTtcbiAgY29uc3QgcmlnaHQ0ID0gcmlnaHRbNF07XG4gIGNvbnN0IHJpZ2h0NSA9IHJpZ2h0WzVdO1xuICBjb25zdCByaWdodDYgPSByaWdodFs2XTtcbiAgY29uc3QgcmlnaHQ3ID0gcmlnaHRbN107XG4gIGNvbnN0IHJpZ2h0OCA9IHJpZ2h0WzhdO1xuICBjb25zdCByaWdodDkgPSByaWdodFs5XTtcbiAgY29uc3QgcmlnaHQxMCA9IHJpZ2h0WzEwXTtcbiAgY29uc3QgcmlnaHQxMSA9IHJpZ2h0WzExXTtcbiAgY29uc3QgcmlnaHQxMiA9IHJpZ2h0WzEyXTtcbiAgY29uc3QgcmlnaHQxMyA9IHJpZ2h0WzEzXTtcbiAgY29uc3QgcmlnaHQxNCA9IHJpZ2h0WzE0XTtcbiAgY29uc3QgcmlnaHQxNSA9IHJpZ2h0WzE1XTtcblxuICBjb25zdCBjb2x1bW4wUm93MCA9XG4gICAgbGVmdDAgKiByaWdodDAgKyBsZWZ0NCAqIHJpZ2h0MSArIGxlZnQ4ICogcmlnaHQyICsgbGVmdDEyICogcmlnaHQzO1xuICBjb25zdCBjb2x1bW4wUm93MSA9XG4gICAgbGVmdDEgKiByaWdodDAgKyBsZWZ0NSAqIHJpZ2h0MSArIGxlZnQ5ICogcmlnaHQyICsgbGVmdDEzICogcmlnaHQzO1xuICBjb25zdCBjb2x1bW4wUm93MiA9XG4gICAgbGVmdDIgKiByaWdodDAgKyBsZWZ0NiAqIHJpZ2h0MSArIGxlZnQxMCAqIHJpZ2h0MiArIGxlZnQxNCAqIHJpZ2h0MztcbiAgY29uc3QgY29sdW1uMFJvdzMgPVxuICAgIGxlZnQzICogcmlnaHQwICsgbGVmdDcgKiByaWdodDEgKyBsZWZ0MTEgKiByaWdodDIgKyBsZWZ0MTUgKiByaWdodDM7XG5cbiAgY29uc3QgY29sdW1uMVJvdzAgPVxuICAgIGxlZnQwICogcmlnaHQ0ICsgbGVmdDQgKiByaWdodDUgKyBsZWZ0OCAqIHJpZ2h0NiArIGxlZnQxMiAqIHJpZ2h0NztcbiAgY29uc3QgY29sdW1uMVJvdzEgPVxuICAgIGxlZnQxICogcmlnaHQ0ICsgbGVmdDUgKiByaWdodDUgKyBsZWZ0OSAqIHJpZ2h0NiArIGxlZnQxMyAqIHJpZ2h0NztcbiAgY29uc3QgY29sdW1uMVJvdzIgPVxuICAgIGxlZnQyICogcmlnaHQ0ICsgbGVmdDYgKiByaWdodDUgKyBsZWZ0MTAgKiByaWdodDYgKyBsZWZ0MTQgKiByaWdodDc7XG4gIGNvbnN0IGNvbHVtbjFSb3czID1cbiAgICBsZWZ0MyAqIHJpZ2h0NCArIGxlZnQ3ICogcmlnaHQ1ICsgbGVmdDExICogcmlnaHQ2ICsgbGVmdDE1ICogcmlnaHQ3O1xuXG4gIGNvbnN0IGNvbHVtbjJSb3cwID1cbiAgICBsZWZ0MCAqIHJpZ2h0OCArIGxlZnQ0ICogcmlnaHQ5ICsgbGVmdDggKiByaWdodDEwICsgbGVmdDEyICogcmlnaHQxMTtcbiAgY29uc3QgY29sdW1uMlJvdzEgPVxuICAgIGxlZnQxICogcmlnaHQ4ICsgbGVmdDUgKiByaWdodDkgKyBsZWZ0OSAqIHJpZ2h0MTAgKyBsZWZ0MTMgKiByaWdodDExO1xuICBjb25zdCBjb2x1bW4yUm93MiA9XG4gICAgbGVmdDIgKiByaWdodDggKyBsZWZ0NiAqIHJpZ2h0OSArIGxlZnQxMCAqIHJpZ2h0MTAgKyBsZWZ0MTQgKiByaWdodDExO1xuICBjb25zdCBjb2x1bW4yUm93MyA9XG4gICAgbGVmdDMgKiByaWdodDggKyBsZWZ0NyAqIHJpZ2h0OSArIGxlZnQxMSAqIHJpZ2h0MTAgKyBsZWZ0MTUgKiByaWdodDExO1xuXG4gIGNvbnN0IGNvbHVtbjNSb3cwID1cbiAgICBsZWZ0MCAqIHJpZ2h0MTIgKyBsZWZ0NCAqIHJpZ2h0MTMgKyBsZWZ0OCAqIHJpZ2h0MTQgKyBsZWZ0MTIgKiByaWdodDE1O1xuICBjb25zdCBjb2x1bW4zUm93MSA9XG4gICAgbGVmdDEgKiByaWdodDEyICsgbGVmdDUgKiByaWdodDEzICsgbGVmdDkgKiByaWdodDE0ICsgbGVmdDEzICogcmlnaHQxNTtcbiAgY29uc3QgY29sdW1uM1JvdzIgPVxuICAgIGxlZnQyICogcmlnaHQxMiArIGxlZnQ2ICogcmlnaHQxMyArIGxlZnQxMCAqIHJpZ2h0MTQgKyBsZWZ0MTQgKiByaWdodDE1O1xuICBjb25zdCBjb2x1bW4zUm93MyA9XG4gICAgbGVmdDMgKiByaWdodDEyICsgbGVmdDcgKiByaWdodDEzICsgbGVmdDExICogcmlnaHQxNCArIGxlZnQxNSAqIHJpZ2h0MTU7XG5cbiAgcmVzdWx0WzBdID0gY29sdW1uMFJvdzA7XG4gIHJlc3VsdFsxXSA9IGNvbHVtbjBSb3cxO1xuICByZXN1bHRbMl0gPSBjb2x1bW4wUm93MjtcbiAgcmVzdWx0WzNdID0gY29sdW1uMFJvdzM7XG4gIHJlc3VsdFs0XSA9IGNvbHVtbjFSb3cwO1xuICByZXN1bHRbNV0gPSBjb2x1bW4xUm93MTtcbiAgcmVzdWx0WzZdID0gY29sdW1uMVJvdzI7XG4gIHJlc3VsdFs3XSA9IGNvbHVtbjFSb3czO1xuICByZXN1bHRbOF0gPSBjb2x1bW4yUm93MDtcbiAgcmVzdWx0WzldID0gY29sdW1uMlJvdzE7XG4gIHJlc3VsdFsxMF0gPSBjb2x1bW4yUm93MjtcbiAgcmVzdWx0WzExXSA9IGNvbHVtbjJSb3czO1xuICByZXN1bHRbMTJdID0gY29sdW1uM1JvdzA7XG4gIHJlc3VsdFsxM10gPSBjb2x1bW4zUm93MTtcbiAgcmVzdWx0WzE0XSA9IGNvbHVtbjNSb3cyO1xuICByZXN1bHRbMTVdID0gY29sdW1uM1JvdzM7XG4gIHJldHVybiByZXN1bHQ7XG59O1xuXG5yZXR1cm4ge1xuICBvYmplY3RNYXRyaWNlcyxcbiAgYXJyYXlNYXRyaWNlcyxcbiAgdHlwZWRBcnJheU1hdHJpY2VzLFxuICBtdWx0aXBseSxcbiAgc2NyYXRjaE9iamVjdE1hdHJpeDogbmV3IE1hdHJpeDQoKSxcbiAgc2NyYXRjaEFycmF5OiBBcnJheS5mcm9tKHsgbGVuZ3RoOiAxNiB9KSxcbiAgc2NyYXRjaFR5cGVkQXJyYXk6IG5ldyBGbG9hdDY0QXJyYXkoMTYpLFxuICBsZW5ndGgsXG59In19) forall the different combinations and possibilities.

In the first test, I compared current `ES6 class/Array/Float64Array`, with the `Matrix4.multiply` method. `Array` and `Float64Array` are clearly superior as has been shown before here.

The second test is more important, I tested how the performance is affected when some percentage (20%) of the operations are performed with either one or two **Frozen** `ES6/Array` or **Proxied** `Float64Array` matrices. There is a **significant performance hit** compared to the regular test, presumably due to [monomorphic vs polymorphic v8 deoptimizations](https://erdem.pl/2019/08/v-8-function-optimization). 

So to summarize:
1. If there is still interest in some level of optimization here while keeping the immutable matrices, `Array`s are freezable and provide a nice theoretical speedup here of **~30%** over the current state on my machine. 
2. Removing immutable matrices entirely will provide a theoretical speedup of **5-10x** over the current state, which is also supported by my original tests in #11814 (comment). Proxied `Float64Array`s are undesirable with even worse performance than current state.

Here are the results:
Regular:
image
Freezing/Proxying:
image

javagl:
That's cool. I won't claim to _fully_ understand _everything_ here, but mainly to confirm my current understanding:

- The best option in terms of performance would be to let it extend `Float64Array`
- The problem: The `Float64Array` cannot be 'frozen' . People could (intentionally or accidentally) do things like `Matrix.IDENTITY[0] = -666.0` and mess up everything. 
  - This _could_, in theory, be avoided with a `Proxy`. But this causes a dismal performance. 
  - On top of that: I think that `Proxy` is generally considered to be a no-go for "real application" code, _except_ for _very_ specific use cases around testing and debugging - and ironically, can make debugging _much_ harder when it's already an integral part of the application codebase. So I'd be reeeeally reluctant to make this part of such a fundamental class like `MatrixN`...

Compared to that:

- The second-best option in terms of performance is to let it extend `Array`
- This _can_ be 'frozen'

I played a bit with the benchmark that you linked to (very helpful, BTW). I was curious about the effect of the "amount of 'frozen' operations". The reasoning, roughly: I have no idea about how many operations in a "standard CesiumJS application that is running for a while" actually **do** happen on frozen objects. Out of all matrix multiplications, how many of them are happening with standard matrices, and how many of them are happening with 'frozen' ones? Are 60% frozen, or 10%, or 0.1%? It's hard to tell. (One could probably _count_ it, during a benchmark run - and this, in fact, could be a case where a `Proxy` could come in handy...)

However, I tweaked the code a bit, to allow selecting the amount of `freezing`:

```javascript
const length = 10000;
const freezing = 0.001;

function maybeFreeze(matrix) {
  return Math.random() < freezing ? Object.freeze(matrix) : matrix;
}
function maybeProxy(matrix) {
  return Math.random() < freezing ? new Proxy(matrix, { set() { throw new Error() } }) : matrix;
}

let objectMatrices;
let arrayMatrices;
let typedArrayMatrices;

if (freezing === 0.0) {
  objectMatrices = Array.from({ length }).map(() => new Matrix4(...randomMatrix()));
  arrayMatrices = Array.from({ length }).map(() => randomMatrix());
  typedArrayMatrices = Array.from({ length }).map(() => new Float64Array(randomMatrix()));
} else {
  objectMatrices = Array.from({ length }).map(() => maybeFreeze(new Matrix4(...randomMatrix())));
  arrayMatrices = Array.from({ length }).map(() => maybeFreeze(randomMatrix()));
  typedArrayMatrices = Array.from({ length }).map(() => maybeProxy(new Float64Array(randomMatrix())));
}
```

But apparently, that number hardly matters: If it is non-zero, then regardless of whether it is 0.9 or 0.001, the performance impact is about a factor of 10 (!), just due to the _presence_ of any frozen object. (As you said, this is likely related to internal morphism (de)optimizations, but I could only speculate about the details here).

Regardless of that, the `Array`-based approach seems to be _consistently_ faster. So for me, this looks like an "easy win", unless I overlooked a strong reason against that.

Beilinson:
Yes I also saw that the percentage of frozen operations does not seem to matter. If all in favor of keeping frozen with just `extends Array` upgrade then I'll go ahead and update the PR, it is a shame to miss out on such performance improvements.

Would anyone be interested in a "PERFORMANCE" mode where certain guardrails are disabled for improved performance? It does seem that `Array` without freezing has about the same performance as `Float64Array` now. The code change would be very minor:

```js
function freeze(obj) {
  return window.CESIUM_EXPERIMENTAL_PERFORMANCE_MODE ? obj : Object.freeze(obj);
}

Matrix4.IDENTITY = freeze(new Matrix4(...));
```

This would keep the cesium code base safe, and allow end consumers targeting low-end hardware with advanced 3D tiles features to perform better. I'm not sure how many more uses for this there are in the code base however.

srcejon:
> Would anyone be interested in a "PERFORMANCE" mode where certain guardrails are disabled for improved performance? 

Do we want faster performance or defense against someone doing Matrix.IDENTITY[0] = -666.0?

There's a John McEnroe quote for that.

javagl:
@srcejon Websearching leads to _some_ quotes that might fit here. If this is about "You cannot be serious", note that the 
`Matrix.IDENTITY[0] = -666.0;`
is overly suggestive. Nobody would write that in actual code.

But problem that is indicated with that **is** real. CesiumJS is heavily using "scratch" variables. Properties usually don't have `/** Documentation */` that describes the role and _ownership_ of these properties. Matrices are passed around left and right, into deep call chains, and returned from deep call chains, without any `/** Documentation */` of whether a function _expects_ or _returns_ a _reference_ or a _copy_. So specifically: At no point in the code, you can be sure that the `result` parameter in a call like
`Cesium.Matrix4.multiply(left, right, result);`
is not actually the `Matrix4.IDENTITY` instance. And there should be _some_ guardrails that prevent overwriting that.

srcejon:
> And there should be _some_ guardrails that prevent overwriting that.

I suspect you're going to notice pretty quickly if you do!

If guardrails are free, then great. If they're costing you 50% performance, then just have them in debug builds.

javagl:
> I suspect you're going to notice pretty quickly if you do!

No doubt about that. And then, you'll just examine all calls to `Matrix4.multiply`, and.... _guess_ which one might have received that `IDENTITY` as its last parameter 😬 The point is that it's nearly impossible to figure that out _post-hoc_, and it _has_ to be detected at the place where it _happened_. And of course, this should preferably be wrapped into 'debug-pragmas'. These pragmas already _are_ used for this sort of test throughout the codebase (and Beilinson already drafted what this could look like for the matrix case)

PR #12975, opened 2025-10-10
Add ScreenSpaceCameraController.collisionHeightReference

Beilinson:
# Description

When a 3D tileset is added to the map, height clamping of the camera to the scene takes about 25% of total frame time, mostly due to Model.pick. This PR proposes a new flag on the `ScreenSpaceCameraController`, that will allows users to continue clamping entities to the 3D tileset without also sampling the camera height off the 3D tileset every frame.

This is useful when 3D tiles are used in conjuction and closely map to a regular `TerrainProvider`, so entities can use `HeightReference.CLAMP_TO_GROUND` and clamp to both terrain and 3D tiles, while the camera can choose to clamp only to the terrain using `HeightReference.CLAMP_TO_TERRAIN`.

This is important, as unlike entities which in the future we can improve their clamping performance, every frame that the camera moves it must sample the 3D tiles as well. The current API is insufficient in providing users with a solution to this problem, as they are either forced to disable collision entirely (including against terrain) on the camera, or disable clamping to 3D tilesets entirely (including for entities). 

## Example

```js

});
const scene = widget.scene;

scene.camera.setView({
  destination: new Cesium.Cartesian3(
    1216411.0748779264,
    -4736313.10747583,
    4081359.5125561724,
  ),
  orientation: new Cesium.HeadingPitchRoll(
    4.239925103568368,
    -0.4911293834802475,
    6.279849292088564,
  ),
  endTransform: Cesium.Matrix4.IDENTITY,
});

const tileset = await Cesium.Cesium3DTileset.fromIonAssetId(40866);
tileset.enableCollision = true;
scene.primitives.add(tileset);

// This billboard will clamp to the 3D tileset
widget.entities.add({
  billboard: {
    image: "../images/facility.gif",
    heightReference: Cesium.HeightReference.CLAMP_TO_GROUND,
  },
  position: new Cesium.Cartesian3(
    1216390.063324395,
    -4736314.814479433,
    4081341.9787972216,
  ),
});

const screenSpaceCameraController = widget.scene.screenSpaceCameraController;
// Previously, we could only disable collision detection entirely
screenSpaceCameraController.enableCollisionDetection = false;
// Now, we can pick and choose:
screenSpaceCameraController.collisionHeightReference = HeightReference.CLAMP_TO_TERRAIN; // The camera will only collide against terrain
```

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Scene.js
+++ b/packages/engine/Source/Scene/Scene.js
@@ -3831,12 +3831,26 @@ function callAfterRenderFunctions(scene) {
   functions.length = 0;
 }
 
+/**
+ * Calculate the height of the globe at the camera position based on the value of {@link ScreenSpaceCameraController.collisionHeightReference},
+ * or undefined if the height cannot be determined.
+ *
+ * @param {Scene} scene
+ * @returns {number|undefined}
+ */
 function getGlobeHeight(scene) {
-  if (scene.mode === SceneMode.MORPHING) {
+  if (
+    scene.mode === SceneMode.MORPHING ||
+    scene._screenSpaceCameraController.collisionHeightReference ===
+      HeightReference.NONE
+  ) {
     return;
   }
   const cartographic = scene.camera.positionCartographic;
-  return scene.getHeight(cartographic);
+  return scene.getHeight(
+    cartographic,
+    scene._screenSpaceCameraController.collisionHeightReference,
+  );
 }
 
 function getMaxPrimitiveHeight(primitive, cartographic, scene) {
--- a/packages/engine/Source/Scene/ScreenSpaceCameraController.js
+++ b/packages/engine/Source/Scene/ScreenSpaceCameraController.js
@@ -24,6 +24,7 @@ import MapMode2D from "./MapMode2D.js";
 import SceneMode from "./SceneMode.js";
 import SceneTransforms from "./SceneTransforms.js";
 import TweenCollection from "./TweenCollection.js";
+import HeightReference from "./HeightReference.js";
 
 /**
  * Modifies the camera position and orientation based on mouse input to a canvas.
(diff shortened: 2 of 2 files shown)


javagl:
This change is a follow-up from the part at https://github.com/CesiumGS/cesium/pull/12968#discussion_r2421259403 

It _looks_ like the repeated 'getHeight' calls are actually causing a performance issue that was reported internally.

During the investigation of the latter, I noticed that the callback wrapper at https://github.com/CesiumGS/cesium/blob/bb28a7450450612e2e506d87f838d9c22e268aae/packages/engine/Source/Scene/Scene.js#L3947 is _always called with `undefined` as it parameter. This means that all the `defined` checks there are useless, and it is _always_ calling `this.getHeight`. I _assume_ that the intention of these `defined` checks was to avoid re-computing the height multiple times, and that the `undefined` parameter causes the same height to be computed again and again (and this is _very_ costly).  

I have not yet analyzed this in detail. That `updateHeight` function and the web of callbacks that are created and passed around there is... \*sigh\* ... I'll need more time to reverse engineer that. I also have not yet cross-checked this with the changes in _this_ PR. But the point is: Maybe the low performance (that is addressed here) comes (mainly) from that `undefined`/callback thingy...?

Beilinson:
Hey @javagl,
I'm not only concerned here about repeated calls while the camera is static, but even the calls when the camera moves.

The current available apis do not provide support for the following setup:

1. Billboards and such clamp to 3d tiles
2. The camera only collides against terrain (to avoid expensive collision tests against 3D tiles every time the camera moves)

This Pr introduces such an api by specifying a collisionHeightReference property, which explicitly allows the developer to choose against which content the camera does collision checks.

The current api allows us only to either entirely disable collision (including against terrain), or entirely disable collision against 3D tilesets, but that includes height clamping for billboards and such.

Issue #10468, opened 2022-06-19
3DTILES_multiple_contents no supports more than 6 contents


engine5d:
3DTILES_multiple_contents no supports more than 6 contents, except change the cesium static setting: `Cesium.RequestScheduler.maximumRequestsPerServer = 8;` 

, or the tile cannot load. But this extension schema [link ](https://github.com/CesiumGS/3d-tiles/blob/main/extensions/3DTILES_multiple_contents/schema/tile.3DTILES_multiple_contents.schema.json)no contains this limitaiton.

`"3DTILES_multiple_contents": {    // 7 contents
				"contents": [
					{
						"uri": "B_0.b3dm"
					},
					{
						"uri": "I_0_0.i3dm"
					},
					{
						"uri": "I_0_1.i3dm"
					},
					{
						"uri": "I_0_2.i3dm"
					},
					{
						"uri": "I_0_3.i3dm"
					},
					{
						"uri": "I_0_4.i3dm"
					},
					{
						"uri": "I_0_5.i3dm"
					}
				]
			}`

lilleyse:
@engine5d thanks for the bug report. This is a CesiumJS limitation, not a 3D Tiles spec limitation.

@ptrgags you might be able to provide more details here. I remember this being a known limitation with the multiple contents implementation. Was this documented in the code or in a github issue?

ptrgags:
I found some comments in `Multiple3DTileContent.js`:

https://github.com/CesiumGS/cesium/blob/main/Source/Scene/Multiple3DTileContent.js#L345

Essentially due to race conditions with request scheduling, apparently it was possible to leak typed arrays somehow (I'm rusty on the details, it was over a year ago), so it tries to schedule all the requests at once with `RequestScheduler`. Since `RequestScheduler` caps the number of requests per server, if you need many contents in a single tile, yes, you'd have to set `RequestScheduler.maximumRequestsPerServer` to a higher value.

lilleyse:
Came up again in https://github.com/CesiumGS/3d-tiles/issues/723

Issue #12616, opened 2025-05-11
Deprecate defer function


jfayot:
`defer` function was added as a [workaround](https://github.com/CesiumGS/cesium/pull/8525). It should be deprecated in favor of using Promise directly.

See also PR [#12581](https://github.com/CesiumGS/cesium/pull/12581#issuecomment-2842652718)

jjspace:
Definitely something we want to do but already tracked:

Duplicate of https://github.com/CesiumGS/cesium/issues/10546
Duplicate of https://github.com/CesiumGS/cesium/issues/10547

PR #12977, opened 2025-10-11
Time dynamic 3D Tiles support

javagl:
The issue about [Time-dynamic 3D Tiles](https://github.com/CesiumGS/3d-tiles/issues/102) has recently has been addressed on the side of the specification, with [draft extension PRs](https://github.com/CesiumGS/3d-tiles/issues/102#issuecomment-3020925540). The subsequent comments in this issue already dived into details of the proposed extensions and approach. This has some overlap to some [discussion in one of the specification PRs](https://github.com/CesiumGS/3d-tiles/pull/817#issuecomment-3307051715).

I'll try to summarize some of the points that are relevant for the interplay between the specification and the implementation within CesiumJS, both in terms of the approach that will be taken, as well as questions that still have to be addressed.

---

### Representation

The original proposal suggested to use the 'multiple contents' concept of 3D Tiles to represent time-dynamic tiles. The idea was to associate a time range with each content (using metadata), and select the respective content for rendering depending on the time range. 

After some discussion, the intention is now rather to _separate_ the concepts of "multiple contents" and "time-dynamic content". There are many open questions about the concept of "time-dynamicness" (some listed below). Having a clean slate offers more degrees of freedom for what can go into the _specification_, and reduces the risk of interfering with existing aspects of the _implementations_.

Two options have been considered for this separation:

1. The time-dynamic content could be represented as an _extension object_ within the `tile.content`. 
  ```
    "content": {
      "uri": "content-2025-09-25-revision0.glb",
      "extensions": {
        "3DTILES_dynamic": {
          "dynamicContents": [
            {
              // Magic...
            },
          ]
        }
      }
    }
  ``` 
2. The time-dynamic content could be a _new content type_:
  ```
    "content": {
      "uri": "timeDynamicContent.json"
     }
  ```
  `timeDynamicContent.json`:
  ```
  {
    "dynamicContents": [
      {
        // Magic...
      },
    ]
  }
  ``` 

The current intention is option 2., to define it as a _new content type_. A summary of the reasoning:

- There may be many entries in that "time dynamic content" decription. The tileset JSON could become _huge_
- It's easier to update/extend a small `timeDynamicContent.json` with an _additional_ time stamp, than to re-process and modify the whole tileset JSON
- Important: This approach allows time-dynamic content to _transparently_ be used in _implicit_ tilesets! Many of the questions about how to represent the time stamps and their different content URIs in an implicit tileset become obsolete. Everything is handled _within_ the content. 

For explicit tilesets and the first draft implementations, the different ways of separating the data does not matter so much. The pseudocode difference is that of
`const d = tile.getExtension(...).dynamicContent;` vs.
`const d = tile.content.dynamicContent;`

There are maaaany low-level quirks omitted here. But one important point is that the content can be created either with the object that was fetched from the tileset JSON, or the one that was loaded from the content JSON file.

In any case, there will be a top-level extension object in the tileset JSON that describes the structure of the dynamic content in more detail.

---

### Features and capabilities

The efforts (and names) focus on **time**-dynamic 3D Tiles. That's the primary use-case. The seemingly philosophical question "What is 'time'?" is important in this context: 

- On the level of the _specification_, it is hard to answer. It could be a unix time stamp, or an ISO8601 string, or a string like `"übermorgen"`. Whatever the representation is, it could turn out to be wrong (or have severe drawbacks and limitations) for certain use-cases.
- On the level of the _implementation_, it is easy to answer: It's what determines which content will be displayed. Really, it's just 
  `const content = getContentFor(time);`

Based on some discussion in the linked issues/PRs, the "dynamicness" for 3D Tiles should not be _limited_ to "time" (given that we cannot even say what that is). There could, for example, be different "revisions" of the same tileset, and it should be possible to _dynamically_ change which revision is displayed. At this point, "time" just becomes "one dimension in the space of things that can be displayed".

One important distinction is whether these dimensions are _discrete_ (like 'revisions'), or _continuous_ (like some forms of 'time'). The _continuous_ dimensions would require some form of "binning/discretization", based on "intervals/ranges". There are some options about where and how this discretization could take place. 

The current intention is to _only_ have _discrete_ keys for the dynamic contents. An overly suggestive example:
```json
"dynamicContents": [
  {
    "uri": "content-time0-revision0.glb",
    "keys": {
      "exampleTimeStamp": "time0",
      "exampleRevision": "revision0"
    }
  },
  {
    "uri": "content-time1-revision1.glb",
    "keys": {
      "exampleTimeStamp": "time1",
      "exampleRevision": "revision1"
    }
  }
]
```
There is a representation of dynamic contents, with different revisions. The application will determine which of the contents should be displayed. The "exampleRevision" could be selected with a dropdown menu. The "exampleTimeStamp" could be selected by the application as well. For example, when the "time" is a value in [0,1], then a "time" of 0.2 could be clamped to `0` and cause the selection of the `"time0"` time stamp.

---

### Implementation

There are several open questions for the implementation, on various levels. Most of the subsequent comments here will likely revolve around the question of "(Oh dear), how are we going to handle _that_?". An attempt to sumarize the current state, open questions, and considerations:

- There now is a `Dynamic3DTileContent` class. This class **is-a** `Cesium3DTileContent`. Most functions in this interface are underspecified. Many functions that could make sense are missing. Nearly _all_ implementations of nearly _all_ functions are just returning "dummy" values (just `0` or `undefined`), because they simply don't make sense for what constitutes "(tile) content". And for nearly all functions, there are deeeep questions about what they should do for "dynamic content". For example: It is not clear whether the `Cesium3DTileContent.trianglesLength` implementation should return the number of triangles that are currently _loaded_, or the triangles that are currently _displayed_. Both options will be wrong in one way or another.
 - There are things like the `Cesium3DTileContentState`. For whatever reason, this `contentState` is stored in the _tile_, and not the _content_. (Corollary: It should be part of the `Cesium3DTileContent` interface!). Dynamic content can not go through the fixed sequence of states (like `LOADING`->`PROCESSING`->`READY`) that is assumed for other contents. Which state it should be in, at which point in time, has to be decided. (Maybe `EXPIRED` could be (ab)used to represent state transitions here, but that remains to be investigated). 
- There are the `update` function and the `ready` flag. Dynamic contents is either "always `ready`" or "never `ready"`. It will likely be the former, but ... some `update` will still have to be called, even when it already counts as `ready`. More generally: The `Cesium3DTile` class has about 15 properties that reflect the state of (and should be stored in) its content, including flags `hasEmptyContent`, `hasRenderableContent`, `hasMultipleContents`, `hasTilesetContent`, and `hasImplicitContent`, which should not exist to begin with, and which can neither be `true` nor `false` for dynamic content.
- The `Dynamic3DTileContent` class has to load other contents. And it has to do this _dynamically_. I tried to take inspiration from `Multiple3DTileContent`, but this has [limitations](https://github.com/CesiumGS/cesium/issues/10468) that should be fixed independently. The `Dynamic3DTileContent` class has to send out requests, wait for the responses, post-process that content, and possibly unload content that is no longer "needed" (until the user drags the time slider, and it is needed _again_). I'm currently trying to create sorts of "utility classes" for all that, but this pretty much experimental. 

---

(The pull request template is omitted for now. It will be inserted before this is converted to 'Ready For Review')


javagl:
The PR has been updated with a first draft of an implementation. This does not fully reflect the latest state of all decisions. There are several open questions. But can serve as a preview. 

### Data side

The dynamic content definition contains information like the following:
```json
"dynamicContents": [
  {
    "uri": "content-2025-09-25-revision0.glb",
    "keys": {
      "exampleTimeStamp": "2025-09-25",
      "exampleRevision": "revision0"
    }
  },
  {
    "uri": "content-2025-09-25-revision1.glb",
    "keys": {
      "exampleTimeStamp": "2025-09-25",
      "exampleRevision": "revision1"
    }
  }
]
```
Each entry indicates which content (URI) should be shown for the respective configuration of `exampleTimeStamp` and `exampleRevision`. (Right now, this is still stored as an extension object of the `content`. It will be in a standalone JSON file in the future).

### Application side

In the application, the selection of the dynamic content is governed by a structure like the following:
```
const dynamicContentsPropertyProvider = () => {
  return {
    exampleTimeStamp: "2025-09-25",
    exampleRevision: "revision1"
  };
};
tileset.setDynamicContentPropertyProvider(dynamicContentsPropertyProvider);
```
Yes, this returns a fixed dummy value here. The point is: There is that "dynamic contents property provider" that provides the properties that should be used for selecting the active content at any point in time. (See the Sandcastle in the example for details).

### Connecting the two

The main work for wiring this together is currently in a class called `Dynamic3DTileContent`. It is a `Cesium3DTileContent` implementation. Most of the questions about how the functions of that interface should be implemented are still open.

The `Dynamic3DTileContent` contains a `_dynamicContentUriLookup`. This is an N-dimensional map: The "dynamic contents property provider" provides the properties for the active content. This map is used for looking up all content URIs that should be active.

The `Dynamic3DTileContent` also contains something that is referred to as "content handles". This is supposed to wrap away the complexity that comes with `Request/Resource/RequestScheduler` and their `throttling/severKeys/priorityFunction` and all the promises that are associated with that. The goal was to have something that can be used like this:
```
if (contentHandle.failed) {
  console.log("Error!");
  return;
}
const content = contentHandle.tryGetContent();
if (!defined(content)) {
   console.log("Still waiting for content");
   return;
}
console.log("Got content: ", content);
```
without having to care about what that `tryGetContent` is actually doing. (It's complicated...)

Combining these functionalities, the `Dynamic3DTileContent` can offer two things:

- an `_allContents` getter that returns the contents that are currently in memory
- an `_activeContents` getter that returns the contents that are currently "active"

Based on that, during its `update` call, it can essentially set all contents "invisible", and only set the "active" contents visible.

The class maintains the `_loadedContentHandles` which stores the contents which are currently in memory. This is implemented as an LRU cache. Some details about how this cache is filled and how contents are unloaded are omitted here. But it's important to be able to _not_ always keep all contents in memory, and still try to not reload the content during each rendering pass or so.

---

### Example

The following is an example tileset and a sandcastle.

[PR0001 example.zip](https://github.com/user-attachments/files/22933076/PR0001.example.zip)


The tileset contains a single dynamic content. It defines 64 time stamps and 3 "revisions", meaning that it has 192 actual contents.

The follwing screencap summarizes a few aspects that can be tested with the sandcastle:

- Scrolling through the time selects one of the 64 time stamps
- The dropdown menu can be used for selecting the revision
- The elaborate debug log show that it's possible to switch through "recently used" configurations without triggering reloads
- Styling can be applied to the tileset, and survives content changes
- Picking only refers to the content that is currently active

Cesium Dynamic PR0001

javagl:
Until now, the dynamic contents information was fetched from an extension in the `content` object. Now the dynamic content is an actual new content type.

A strong reason for treating it as a new content type is that this should simplify using dynamic contents within _implicit_ tilesets. So I created an example for that as well. Here's a screencap of that example:

Cesium Dynamic PR0002

The previous test data (explicit) and the new data set (implicit) both with using that new content type are attached here, together with the Sandcastle:

[PR0002 example.zip](https://github.com/user-attachments/files/22953355/PR0002.example.zip)

(Note: I have a bunch of infrastructure for ~"creating test data like that". This was _developed_ in my spare time, off-the-clock. Tweaking it for cases like this sometimes requires a bit of fiddling and copy-and paste and hacks and manual steps, and now there's a `public class _CesiumImplicitDynamicTest` here. This _specific_ fiddling and test data generation is at least partially Cesium time. Maybe, one day, I can allocate a (large enough) chunk of spare time to clean this up and make it more usable...)

javagl:
The last commit squashes a few cleanups and minor extensions of the functionality. For example, the support of metadata for each content. I originally thought that this metadata might have to be "tracked", similar to the "Style" that is assigned to the content. But in fact, it turned out that this consisted of "not doing anything special": The elements for the "inner contents" are just treated as standard content objects, and as such, can have their usual content metadata. 

However, the test data package has been extended with a data set that assigns some group- and content metadata to the contents, and a sandcastle that allows showing the metadata.

Cesium Dynamic Metadata 0001

The data and sandcastles are here:

[PR0003 example.zip](https://github.com/user-attachments/files/23217635/PR0003.example.zip)

javagl:
The current state contains a few **drafts** for dealing with the "tileset statistics" that may have to be maintained for the dynamic content, and where it is by no means clear how they should be tracked, or even what their values should be.

The current handling of the tileset statistics is smeared all over the code. For example:

-  Changing the number of pending requests at https://github.com/CesiumGS/cesium/blob/0ee5670050c7591748fcda95e6762b428750be50/packages/engine/Source/Scene/Cesium3DTile.js#L1216
- Changing the number of pending requests at https://github.com/CesiumGS/cesium/blob/0ee5670050c7591748fcda95e6762b428750be50/packages/engine/Source/Scene/Multiple3DTileContent.js#L335
- Something similar will have to be done for the dynamic content...
- Eventually, the tileset needs this information at https://github.com/CesiumGS/cesium/blob/0ee5670050c7591748fcda95e6762b428750be50/packages/engine/Source/Scene/Cesium3DTileset.js#L3290
 
Given the lack of documentation, one can only try to reverse engineer (i.e. **guess**) what the difference is between `numberOfPendingRequests` and `numberOfActiveRequests`, or what `numberOfAttemptedRequests` is. Unfortunately, some of this actually affects the _behavior_ of the tileset. This means that iff one of these guesses is wrong, then there's a bug lurking somewhere. I tried to introduce some `RequestListener` that could be used for tracking these statistics. But of course, this can only be attached to the `RequestHandle` which is introduced here (and which _itself_ is only an attempt to hide some complexities). And eventually, whether this working or not, depends on whether this listener is informed about the right things at the right places. I think that _some_ of this may become clearer when trying to create specs that mimic e.g. the behavior of the `Multiple3DTileContents` specs, but there are some differences between "multiple contents" and "dynamic contents" where the _intended_ behavior (i.e. the expectations in these specs) still have to be fleshed out.

javagl:
A small addendum (something like a "TODO note") regarding the last update: This currently covers the case where each "inner content" does have metadata. There might also be the case where the _dynamic content itself_ has metadata. It _might_ be easy to support that, with some potential corner cases that are similar to the broader, structural questions that are still open in view of the `Cesium3DTileContent` interface implementation: Where, when, and how do clients deal with the "main" content object, or the "inner contents"?

javagl:
I've spent some (unattributed) time with creating some additional test data, and wondered what could be "realistic" usage patterns for this extension. Some usage scenarios are obvious, but these might only reflect a tiny part of what _should_ be possible.  The _exact_ structure of the resulting data sets is not clear. And I could not easily and quickly _create_ any real-world data. But one can always try to create artificial data sets that are "structurally similar" to what could be expected for real-world data.

The tests until now usually contained "few" inner contents, revisions, and time stamps. One test already consisted of 3 revisions with 64 time stamps, but still with only _one_ inner content being active at one time (!). I now created a test where  there is a varying number of active contents for each time stamp, namely 1 up to 64.

Cesium Dynamic Many 2

The structure is still pretty simple: One root with 64 children. But it already shows how quickly it all this could have become  pretty chunky if we had inlined that in the tileset JSON itself. Right now, the tileset has 16KB, and the content JSONs have 112KB, and ... this is still pretty small, and (important) does not carry any per-content metadata. (This could really bloat it, depending on the complexity of the metadata...). I'll just attach the data here, for tracking and testing, even though I'll certainly tweak it as we move on:

[dynamic 2025-10-31.zip](https://github.com/user-attachments/files/23247899/dynamic.2025-10-31.zip)


---

One could now toss around ideas and further considerations.

Right now, the structure of the data is:
"For each content, it says at which timestamp it should be displayed"
It could also be:
"For each time stamp, it says which contents should be displayed"

And this is actually the data structure that is built internally: It is a map from (timeStamp, revision) to the list of content URIs for this tuple. But I think that it would not necessarily make sense to try and mimic that in the actual JSON data, _unless_ the primary use cases would justify that (and from the use-cases that I've seen being considered until now, this is not the case).

javagl:
#### About `defer`

The current implementation state uses the [`defer` functionality](https://github.com/CesiumGS/cesium/blob/edceb42f3ac54aa3e3ebef17ab96f3d2f39f133c/packages/engine/Source/Core/defer.js#L29). I later noticed that there is an open issue suggesting to [deprecate the defer function](https://github.com/CesiumGS/cesium/issues/12616). And I was wondering why.

In the current state, the `defer` function is used in a `ResourceHandle` class that is supposed to encapsulate one request, and handle its lifecycle somewhat conveniently. Of course, this should not be necessary. It should be possible to perform a request with the snippet that is shown in the `Resource.fetchArrayBuffer` function:
```javascript
resource.fetchArrayBuffer().then(function(arrayBuffer) {
  // use the data
}).catch(function(error) {
  // an error occurred
});
```
But it isn't. I added a [comment](https://github.com/CesiumGS/cesium/blob/6a07d547817c14513155cbe12755f1b02d6903fa/packages/engine/Source/Core/Resource.js#L794) in that function to summarize some of the quirks: The function may return `undefined`, or a promise that resolves to `undefined`, or a promise that is rejected with `undefined`, or with an error, or with `RequestState.CANCELLED`, and it seems like the latter may not only happen when the request is cancelled, but also when there is some "throttling" going on. 

Long story short: The use of `defer` in that `RequestHandle` aims at creating something that is at least closer to the simplicitly that is suggested by this snippet: 

- The `ResourceHandle` [creates a `defer` object in its constructor](https://github.com/CesiumGS/cesium/blob/6a07d547817c14513155cbe12755f1b02d6903fa/packages/engine/Source/Scene/Dynamic3DTileContent.js#L537)
- The `ResourceHandle` [offers a "result promise"](https://github.com/CesiumGS/cesium/blob/6a07d547817c14513155cbe12755f1b02d6903fa/packages/engine/Source/Scene/Dynamic3DTileContent.js#L582) - i.e. the promise of that defer object. This always exists, never changes, and will always be rejected or resolved "cleanly".
- The `ResourceHandle` has an `ensureRequested` function that has to be called ~"in every frame" (for lack of a better description of the lifecycle). It calls the aforementioned `fetchArrayBuffer` function, and deals with the cases that can happen there:
  - When `fetchArrayBuffer` returns `undefined`, then nothing is done (waiting for the next `ensureRequested` call)
  - When `fetchArrayBuffer` returns a promise, then
    - the "result promise" will be resolved when that promise returns an actual array buffer
    - the "result promise" will be rejected when that promise is rejected or the request state indicates that the request was cancelled

There is a "draft of a draft" for a spec for this, that shows this pattern:
```javascript
    const requestHandle = new RequestHandle(resource);

    // Fetch the promise from the request handle
    const resultPromise = requestHandle.getResultPromise();
    resultPromise
      .then(function (arrayBuffer) {
        // use the data
        console.log("resolved with ", arrayBuffer);
      })
      .catch(function (error) {
        // an error occurred
        console.log("rejected with ", error);
      });

    // Ensure that there is a pending request
    requestHandle.ensureRequested();
```
And regardless of where, when, and how the request is resolved or rejected or cancelled, it should always map to these two `then/catch` blocks. This has to be tested for the difficult case where the "throttling" kicks in, but the hope is that it should be possible to keep this (simple) pattern even in this case.

---

#### About specs and mocking

There are some preliminary drafts for specs. Some of them are relatively simple. But as soon as anything has to actually _load a `Model`_ (as part of a tile content - even "dummy" tile content!) things become tricky. 

It's basically not possible to "just load a model". Everything that involves "loading a model" requires a `Scene`, which is a monolithic monster object that connects everything to everything (and which is likely the source of some race conditions that we have in the specs). And specifically, it requires a `FrameState`, which also means that "loading a model" requires a GL context. (I'll skip the details about `update` functions and the loader structures). An undesirable side effect is that these tests cannot really be run in CI. 

Well. It looks like it is ... "somewhat possible" to load a model without a context. There currently is a [function to create a mock `FrameState`](https://github.com/CesiumGS/cesium/blob/59b9f79c4a8ea2e023a40ade1cfaf874a1f6c299/packages/engine/Specs/Scene/Dynamic3DTileContentSpec.js#L160). The fact that "loading a model" requires the `scene.frameState.context.uniformState.view3D` matrix to be defined, and calls some `scene.frameState.brdfLutGenerator.update()` function are hard to explain, but ... it seems to work 🤷‍♂️ 

In any case, I'd like to be able to run some of the tests (that per se have zero connection to GL, and that are "just loading data"!) without a GL context. Maybe I'll eventually have to resort to the `beforeAll(... scene = createScene()...)`/`afterAll(...scene.destroyForSpecs()...)` pattern that is used elsewhere. But I'd like to avoid that, because this is essentially "testing" thousands of things, and "the thing that _should_ be tested" is always only a tiny fraction of that.

javagl:
The last commit contains some squashed changes related to spec experiments.

An attempt of a **tl;dr** (in addition to the open discussion points from the previous comment): 

There are several classes that are currently contained in the `Dynamic3DTileContent.js` file. 

(Aside: These classes should be in individual files. The current approach of "putting all sorts of things into one file" and "putting all sorts of files into one directory" is preventing proper code reuse and modularization).

Two of these classes are `RequestHandle` and `ContentHandle` (where the latter uses the former). The goal of these classes was to have a functionality that is as easy as the `Resource.fetchArrayBuffer` functionality, but actually _behaves_ like the snippet in that function suggests. To achieve that, these classes are trying to hide the complexity of the `Resource/Request/RequestScheduler` behavior when it comes to creating, returning, resolving, and rejecting promises, with the largely unknown and unspecified dimension of how things behave in terms of 'throttling' and cancelling requests.

So these classes are trying to hide complex, un(der)specified behavior. But in order to test whether they _really_ do that (and do it _properly_), the specs would require some mocking that _exactly_ replicates this complex, underspecified behavior. At this point, the specs cannot really "test the expected behavior". They can only test the mocking itself. (Could someone write a proper comment for [`Resource._makeRequest`](https://github.com/CesiumGS/cesium/blob/bcc5ea383e294c698940dbf9df76ac5ba6961a73/packages/engine/Source/Core/Resource.js#L1357)?)

Another important point about the specs is _what exactly they are testing_. Right now, many things are "connected" that should not be connected. For example, the `Multiple3DTileContentSpec` checks whether the `tileset.statistics` are updated based on resolved/rejected requests. This is due to the fact that `Multiple3DTileContent` is doing lots of manual request handling (with https://github.com/CesiumGS/cesium/issues/10468 being a consequence of that), and lots of manual statistics updates that are wired into this request handling. So on the one hand, of course, the spec _has_ to test this. On the other hand, this class should not really be responsible for these things. The `Multiple3DTileContent` class should be able to call some 
`const promise = sendOutRequestForContentOf(tileset, "content.glb");`
and deal with that promise, and not care about what the `RequestScheduler` is doing internally. 

Trying to "decouple things": The `ContentHandle` class currently allows adding some "content listener" that will be informed when the content is "loaded and ready", and when the content is "unloaded". This will, for example, be used for tracking the `tileset.statistics.geometryByteLength` depending on the loaded/unloaded content. But the specs for this class _should_ not care about the statistics. The specs for the `ContentHandle` should only verify that the content listeners are properly informed. What these listeners are _doing_ is not the business of that class in and for itself. (There is a short [inlined comment](https://github.com/CesiumGS/cesium/pull/12977/commits/48bbf6fd4ffe4fe3109f46148d2dfccf461aca8a#diff-08f99358f9d02317fca5cfda76fa9dd71e81026d69fb82f60ae5423987bda8c5R1145) about this).

javagl:
There is now a squashed commit with some cleanups and additional tests. 

The tests for the `NDMap` and `LRUCache` class are pretty straightforward. They could probably be extended, and parts of them could probably even be generalized to some `Map`-spec, but they should cover all the functions for now.

I also added some further tests for the actual _content_ handling, which (as described above) requires some obscure mocking. These are "workarounds" for now - I opened https://github.com/CesiumGS/cesium/issues/13052 for the broader discussion that may take place here.

---

@lilleyse There still was a TODO on my list, which revolved around considering to describe the structure of the (time-) dynamic data using _metadata_. For context: Right now, the tileset contains a top-level extension object like this:
```javascript
const basicDynamicExampleExtensionObject = {
  dimensions: [
    {
      name: "exampleTimeStamp",
      keySet: ["2025-09-25", "2025-09-26"],
    },
    {
      name: "exampleRevision",
      keySet: ["revision0", "revision1"],
    },
  ],
};
```
It describes the "dimensions" (`exampleTimeStamp` and `exampleRevision`, and their domains in the tileset). The actual tile content  then contains something like this...
```javascript
  dynamicContents: [
    {
      uri: "exampleContent-2025-09-25-revision0.glb",
      keys: {
        exampleTimeStamp: "2025-09-25",
        exampleRevision: "revision0",
      },
    },
...
```
to indicate the time stamp and revision for which that content should be "active".

One could make a point that the "dimensions" information could include further details - for example, the _data type_ of the dimensions (which might very well be a `number` for the `exampleRevision`), and possibly further information (e.g. the fact that the `exampleTimeStamp` uses a date pattern of `YYYY-MM-DD`). At some point, this top-level extension object would _resemble_ a metadata class, with `properties` and their type information.

I drafted this out here:

#### The class:

The properties would be limited to non-array "STRING" and numeric "SCALAR" values. One could carefully think through the cases of ENUM or BOOLEAN. On the one hand, one could be tempted to restrict numeric properties to integer types. On the other hand, a "SCALAR FLOAT32 timeInSeconds = 123.456" could actually make sense. On the third hand, this would likely NOT make sense without opening the can of worms of epsilon-comparisons...
```javascript
const dynamicContentStructureClass = {
  name: "Dynamic Content Metadata Class",
  description:
    "An example of describing the structure of the dynamic elements of 3D Tiles content as an actual 3D Tiles Metadata class",
  properties: {
    exampleTimeStamp: {
      description: "The time stamp, following the pattern YYYY-MM-DD",
      type: "STRING",
    },
    exampleRevision: {
      description:
        "The revision, stored as a number explicitly for illustration",
      type: "SCALAR",
      componentType: "UINT8",
    },
  },
};
```

#### The content with the metadata entity

The dynamic content properties are stored in an object that conforms to the class shown above. The properties can not trivially be stored as part of a content "metadata" entity, because that would interfere with the actual content metadata (also see notes about the schema below)
```javascript
const exampleContentWithMetadataEntity = {
  uri: "exampleContent-2025-09-25-revision0.glb",
  metadata: {
    // ...
  },
  dynamicContentProperties: {
    exampleTimeStamp: "2025-09-25",
    exampleRevision: 0,
  },
};
````

#### The schema...? 

There are two options:

1. the "dynamic content structure class" could be part of the actual tileset metadata.
    - It would mean that adding dynamic-ness to an existing tileset would imply adding a new class to the metadata structure       (and require a new schema ID). The values of the time-dynamic content properties would be part of the `content.metadata` entity.
    - There would be no way to know which class is the one that is carrying the time-dynamic information: Something like "semantics on CLASSES" is currently not possible. Using property semantics wouldn't work either, because each semantic can only appear once. (Workarounds like defining some "DYNAMIC_PROPERTY_<index>" semantics excluded here...)
    - Related to that: The case that there may be TWO classes that define time-dynamic content properties would have to be excluded somehow.

2. The "dynamic content structure" could be independent of the actual tileset metadata. This would not require it to be part of a schema, and limit the usefulness of "treating it like metadata" to begin with (even more than the inherent limitations of properties to "STRING" or "SCALAR number" already does...)

**Comment by javagl** [comment]
Just a short update on this: I'll try to wrap up ~"something that passes CI" now. 

To emphasize this: Creating the core functionality here was done in about 2 weeks. It's really not difficult. Load some JSON. Show and hide geometry. Fairly trivial, actually. 

The rest of the time was/is spent with trying to deal with what else is there. And this is just too much. The deeper issues have just been ignored for too long. So I'll also abandon the idea of creating sensible tests. I'll instead create _some_ (few) tests that are structurally similar to the existing tests, being well aware that this will 1. take far longer than it should, and 2. (more importantly) make it harder to bring the tests into a sensible shape later. For some of the "convenience" classes (i.e. the classes that try to hide some of the quirks from the existing classes), it will not be possible to create such tests _at all_. But it probably doesn't matter: When something doesn't work, let's track it in an issue... ... ...

**Comment by javagl** [comment]
I've started changing the spec approach back to the one that is used for other specs (like `Multiple3DTileContentSpec`). 

The current specs revolve around a test tileset that I created for a basic test with metadata (see https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3463428332 ). This tileset is **not** checked in - I'd like to avoid bloating the commit history with lots of binary data. (It's a "minimal" tileset for the tests, but maybe it will be replaced with another one, depending on whether the spec approach is deemed appropriate or not). There are currently only a few specs for the basic functionality. Specs for testing the metadata handling will be added (even though there's nothing specific for metadata in the dynamic content). Maybe there will be another test with an _implicit_ tileset (even though there's not much specific for implicit tileset in the dynamic content). As mentioned in https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3538922419 : The structures that could warrant extensive tests (`RequestHandle` and `ContentHandle`) are not exposed, and they are trying to hide quirks from existing classes, and the specs would therefore have to perfectly mimic the (unspecified!!!) behavior of the existing classes, so there's hardly a way to add reasonable specs for them right now.

**Comment by javagl** [comment]
The specs have been updated by reviving the specs for the `NDMap` and `LRUCache` helper classes, as well as adding specs for the basic metadata handling.

There are still some open questions about the "proper" implementation of the `Cesium3DTileContent` interface, and the metadata handling is one specific aspect of that: Right now, the content returns `undefined` as its metadata. In order to _access_ the metadata, it is necessary to access the "inner contents" (which are "active" at the current point in time). There isn't an _explicit_ (public) way of doing that. But the specs are accessing `_private` variables to check that they have the right values.

The actual spec data is still not commited (to avoid bloating the history), but attached as an archive here:

[Dynamic Spec Data 2026-01-10.zip](https://github.com/user-attachments/files/24545134/Dynamic.Spec.Data.2026-01-10.zip)

The archive contains the data that goes into `Specs\Data\Cesium3DTiles\Dynamic\Dynamic`. It consists of a tileset with a single content (and 9 actual, dynamic contents). It includes a README.md with a summary of the structure of the tileset, including a Sandcastle for loading this test data and changing the dynamic content properties.

**Comment by javagl** [comment]
Another set of test data (including a Sandcastle):

[builderExperimentsDynamic 2026-01-24.zip](https://github.com/user-attachments/files/24839510/builderExperimentsDynamic.2026-01-24.zip)

I wouldn't call it "real-world" test data, but ... it's more real-world than the previous test data, and may be something that could go into the 3D Tiles Samples, just to illustrate the idea:

Cesium Dynamic Houses 2026-01-24

**Comment by javagl** [comment]
Given that this has been "inactive" for a while now, I think that it could make sense to mark it for a first review pass.

The basic fuctionality is illustrated with the example data sets (e.g. from the last comment here), including the respective sandcastles. Actual spec data is not yet committed (only atttached in [this comment](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3733221195)) to not bloat the size, and assuming that details of that data may still change.

On the code level, there are a few inlined comments that are tagged with `XXX_DYNAMIC` where ~"some things are not finally decided", but nothing that should prevent a first review pass. This should **not** be on the "nitpicking/bikeshed level" (like single-word changes in JSDoc or renaming variables or such). This is really about whether the overall approach looks somewhat reasonable.

Tagging @lilleyse as a possible reviewer.

**Comment by javagl** [comment]
Moving up the granularity:

One consideration that was brought up early in the process was whether it should (ever) be possible to display two contents with different time stamps at the same time. At this point, the answer was a clear "No". But (maybe due to the generalization of the overall approach, beyond 'time stamps') it came up again now. And it could make sense to think about how _some_ of these generalizations could be pushed further and made available for clients.

## Current state - data side

The current structure of an "entry" of a dynamic content is this:
```
 dynamicContents: [
    {
      uri: "exampleContent-2025-09-25-revision0.glb",
      keys: {
        exampleTimeStamp: "2025-09-25",
        exampleRevision: "revision0",
      },
    },
    {
      uri: "exampleContent-2025-09-26-revision1.glb",
      keys: {
        exampleTimeStamp: "2025-09-26",
        exampleRevision: "revision1",
      },
    },
    ...
```

## Current state - client side

The client currently selects content URIs to be "active" based on specific properties (keys). There is a function that determines the active URIs by returning an object like this:
```
const currentDynamicContentProperties = {
  exampleTimeStamp: "2025-09-25",
  exampleRevision: "revision0",
};
```
This will cause the URI `exampleContent-2025-09-25-revision0.glb` to be "active".

## The "limitation"

Right now, there is no way to display _both_, 
`exampleContent-2025-09-25-revision0.glb` and
`exampleContent-2025-09-26-revision1.glb`
at the same time.

(Note: This does not make sense for "time stamps" or "revisions", but _could_ make sense for other keys - these are just used in the examples for now...)

## Generalizations

One of the [earliest comments in the underlying issue](https://github.com/CesiumGS/3d-tiles/issues/102#issuecomment-3307045781) already mentioned something that could be seen as a first step of a generalization here - quoting the relevant part:

> There could/should probably a form of "wildcard" mechanism, on the implementation side, or on the data side, or both(!)
> - On the implementation side, a call like `getCurrentContents(currentTime, undefined)` could return all 'revisions' for the given time.
> - On the data side, an entry with `{ time: "2024", revision: undefined }` could be returned regardless of the revision that was given in the getCurrentContents call
>   - Quirk: Maybe there should be a mechanism for omitting data-side wildcard values ...

While this _could_ be useful, it could be generalized further. Specifically: When there are three "revisions", `revision0`, `revision1`, `revision2`, then the wildcard approach would only allow to select _all_ of them, but not _two specific_ revisions. (One reason for why I "only" suggested the wildcard back then may be obvious: It's far easier to implement than the full, fine-grained control that will be considered now)

## Generalization considerations

### Data side

There is likely not too much need for generalization on the _data side_, except for the points that have already been considered:

- Wildcards (just leaving one of the keys `undefined`)
- Data type information (saying, in the "schema", that a certain key represents a `number` or `Date` (aka ISO8601 string))

Beyond that, the data side could already be interpreted as a table like this:

| URI |      exampleTimeStamp|  exampleRevision|
|:----------|:-------------:|:------:|
| exampleContent-2025-09-25-revision0.glb |  "2025-09-25" | "revision0" |
| exampleContent-2025-09-26-revision1.glb |  "2025-09-26" | "revision1" |
| ... |    ...    |   ...  |

And it's clear that this already offers **all** options for `SELECT`ing URIs. 

### Client side now

The main question is: What should be possible on the client side?
(And: How is this implemented?)

The client defines the "dynamic content properties provider" like this:
```
const currentDynamicContentProperties = {
  exampleTimeStamp: "2025-09-25",
  exampleRevision: "revision0",
};
const dynamicContentsPropertyProvider = () => {
  return currentDynamicContentProperties;
};
tileset.setDynamicContentPropertyProvider(dynamicContentsPropertyProvider);
```
This provider is queried in each frame, and determines the list of "active URIs" based on the properties of the returned object. 

#### Performance note 

The current implementation exploits the existing "limitation". When the content is loaded, then the implementation builds a mapping
`Map<ActiveKeys, string[]>`
that maps an object like
```
const currentDynamicContentProperties = {
  exampleTimeStamp: "2025-09-25",
  exampleRevision: "revision0",
};
```
to the list of URIs that are "activated" by these properties. This allows for an O(1) lookup of the active URIs. Depending on the number of contents and properties, any alternative that is considered here has to be examined in terms of its possible performance implications.

### Client side brainstorming

In the most general form, the selection of the active URIs may happen based on some sort of query language. In fact, this could be similar to the declarative styling.

`show`: `"${exampleTimeStamp} === '2025-09-25' && (${exampleRevision} === 'revision0' || ${exampleRevision} === 'revision1')"`

aka

`SELECT * FROM data WHERE exampleTimeStamp = '2025-09-25' AND (exampleRevision = 'revision0' OR exampleRevision = 'revision1')`

There are many, many, degrees of freedom for how this could be offered. It should probably not happen based on such a `contentQueryString` that is parsed and processed. (It could, in theory, but it should certainly not be the first iteration). One approach could be to expose the _conditions_ to the client. 

The client could use someting like this:
```
const currentDynamicContentPropertiesConditions = {
  exampleTimeStamp: (t) -> { return t === "2025-09-25" },
  exampleRevision: (t) -> { return r === "revision0" || r === "revision1" },
};
```
The properties would not be `string` values that are compared, but they would be `function` objects that receive the `string`, and return `true` or `false` (i.e. "conditions", or "predicates"). This could be _very_ generic. It would even allow "range queries" like
`exampleTimeStamp: (t) -> { return isBetween(t, "2012-01-01", "2026-12-25"); }`

The implementation would query this object, in each frame, and check all contents against the given conditions, and make their URI "active" when the condition evaluates to `true`. (Again: We'll have to keep an eye on performance here!)

Note that there still could (and should!) be "convenience" layers built on top of that. Simple things should be simple. Complex things should be _possible_. So the client could _still_ use the "simple" approach
```
const currentDynamicContentProperties = {
  exampleTimeStamp: "2025-09-25",
  exampleRevision: "revision0",
};
const dynamicContentsPropertyProvider = () => {
  return currentDynamicContentProperties;
};
tileset.setDynamicContentPropertyProvider(dynamicContentsPropertyProvider);
```
with that `setDynamicContentPropertyProvider` being implemented as 
```
function setDynamicContentPropertyProvider(propertiesProvider) {
  const conditionsProvider = () => {
    const properties = propertiesProvider();
    return {
      // Pseudocode: Map all "properties" to "conditions"
      // that return "true" for the exact property values
      exampleTimeStamp: (t) -> { return t === properties.exampleTimeStamp; },
      exampleRevision: (t) -> { return r === properties.exampleRevision;  },
    }
  };
  this._conditionsProvider = conditionsProvider;
}
```
Details TBD once we have a clearer idea about how far the generalizations should go.

**Comment by javagl** [comment]
Rather for "completeness": The generalization approach mentioned above was to be able to define conditions for the properties like this:
```
const currentDynamicContentPropertiesConditions = {
  exampleTimeStamp: (t) -> { return t === "2025-09-25" },
  exampleRevision: (t) -> { return r === "revision0" || r === "revision1" },
};
```
to make a URI "active" when all of them return `true`.

What is obviously **not** possible with this: To define conditions for a URI to be "active" based on _two_ property values. 

The last ("maximum") generalization would therefore essentially be a callback/predicate on _the whole key object_, like this
```
const uriCondition = (keys) {
  return keys.exampleTimeStamp === "2025-09-25" && keys.exampleRevision !== "revision2";
};
```

**Comment by lilleyse** [comment]
```js
const uriCondition = (keys) {
  return keys.exampleTimeStamp === "2025-09-25" && keys.exampleRevision !== "revision2";
};
```

I like this. Feels pretty clean and most declarative styling-like.

javagl:
I drafted a generalization as discussed above. Currently, this is in https://github.com/CesiumGS/cesium/tree/dynamic-3d-tiles-staging-generalization . 

This change does remove a considerable amount of code: The `NDMap` class that had been created here was used for the O(1) lookup of "keys" to the set of "active" URIs. Now, this list of active URIs is determined in each "update", by iterating over the contents, and checking the respective keys agains that URI conditon function.

As noted above, the change does not affect the data side. And in fact, the only change here was in the `_activeContentUris` getter and changing the `tileset.dynamicContentPropertyProvider` into a `dynamicContentUriConditon`. So from the client perspective, setting this "URI condition" is done with something like that:
```
  const dynamicContentUriCondition = (keys) => {
    if (keys.exampleTimeStamp !== "2025-09-30") {
      return false;
    }
    if (keys.exampleRevision !== "revision1") {
      return false;
    }
    return true;
  };
  tileset.dynamicContentUriCondition = dynamicContentUriCondition;
```
As a local test, I did insert some "call counter" there, and as expected, it is called many, many times. I still have some concerns that this may turn into a bottleneck when there are many, many contents and many, many conditions, because then, it will be called (many, many) x (many, many) x (many, many) times. On the one hand, that's the usual trade-off between genericity and performance. On the other hand, **iff** it really turns out to be expensive, one _could_ consider specialized paths or optimizations. 

(The latter might involve things that require that `NDMap` with its O(1) lookup again - that's why I didn't move this into this "staging" PR branch for now...)




---

<details><summary>A test sandcastle</summary>
<p>

```

//viewer.extend(Cesium.viewerCesium3DTilesInspectorMixin);

// The base URL of the test tilesets
const tilesetBaseUrl = "http://localhost:8003/";

// Stores the tileset that is currently selected
let currentTileset;

// Creates the tileset for the sample with the given name.
async function createTileset(relativeTilesetJsonName) {
  if (Cesium.defined(currentTileset)) {
    viewer.scene.primitives.remove(currentTileset);
    currentTileset = undefined;
  }
  currentTileset = await Cesium.Cesium3DTileset.fromUrl(
    `${tilesetBaseUrl}${relativeTilesetJsonName}`,
    {
    }
  );
  viewer.scene.primitives.add(currentTileset);
  setupTileset(currentTileset);
}


//============================================================================
// Adjust the timeline for the test tileset
const timeline = viewer.timeline;
const startTime = Cesium.JulianDate.fromIso8601("2025-09-24");
//const stopTime = Cesium.JulianDate.fromIso8601("2025-09-29");
const stopTime = Cesium.JulianDate.fromIso8601("2025-11-30");
timeline.zoomTo(startTime, stopTime);

viewer.clock.currentTime = Cesium.JulianDate.fromIso8601("2025-09-25");

//============================================================================


// Create one entry for the list of examples that can
// be selected in the dropdown menu.
function createSampleOption(name) {
  return {
    text: name,
    onselect: async function () {
      await createTileset(name);
    },
  };
}

// Create the list of available samples, and add them
// to the sandcastle toolbar
const sampleOptions = [
  //createSampleOption("test0006_asContent/tileset.json"),
  //createSampleOption("test0007_implicit/tileset.json"),
  //createSampleOption("test0008_metadata/tileset.json"),
  createSampleOption("test0009_manyActive/tileset.json"),
];
Sandcastle.addToolbarMenu(sampleOptions);


// The revision that is currently selected in the dropdown
let currentExampleRevision = "revision0";

const currentExampleProperties = {
  exampleTimeStamp: "2025-09-30",
  exampleRevision: "revision1"
};

// Utility function to create a time stamp string from
// a given time, matching the simple date format that
// is used in the example, "YYYY-MM-DD"
function createExampleTimeStampString(currentTime) {
  const currentDate = Cesium.JulianDate.toGregorianDate(currentTime);
  let y = String(currentDate.year);
  let m = String(currentDate.month);
  let d = String(currentDate.day);
  if (m.length < 2) {
    m = "0" + m;
  }
  if (d.length < 2) {
    d = "0" + d;
  }
  const currentExampleTimeStamp = y + "-" + m + "-" + d;
  return currentExampleTimeStamp;
}

// Create options to select the example "revision" property
function createRevisionSelectionOption(name) {
  return {
    text: name,
    onselect: function () {
      currentExampleRevision = name;
      currentExampleProperties.exampleRevision = name;
    },
  };
}
const revisionSelectionOptions = [
  createRevisionSelectionOption("revision0"),
  createRevisionSelectionOption("revision1"),
  createRevisionSelectionOption("revision2"),
];
Sandcastle.addToolbarMenu(revisionSelectionOptions);


//============================================================================
// Test for selecting different styles to assign to the tileset
function createStyleOption(name, color) {
  if (color === undefined) {
    return {
      text: name,
      onselect: () => {
        currentTileset.style = undefined;
      },
    };
  }
  return {
    text: name,
    onselect: () => {
      currentTileset.style = new Cesium.Cesium3DTileStyle({
        color: color,
      });
    },
  };
}

const styleOptions = [
  createStyleOption("(No style)", undefined),
  createStyleOption("Red", "color('red')"),
  createStyleOption("Green", "color('green')"),
  createStyleOption("Blue", "color('blue')"),
];
Sandcastle.addToolbarMenu(styleOptions);

//============================================================================
// (Dummy) test for picking the dynamic content
/*/
const handler = new Cesium.ScreenSpaceEventHandler(viewer.canvas);
handler.setInputAction(function (movement) {
  const picked = viewer.scene.pick(movement.position);
  if (!Cesium.defined(picked)) {
    return;
  }
  console.log("Picked: " + picked.detail?.model?._resource?.url);
}, Cesium.ScreenSpaceEventType.LEFT_CLICK);
//*/


//============================================================================
// Test for picking and metadata handling if the dynamic content

let showMetadata = false;
Sandcastle.addToggleButton("Show metadata", showMetadata, function (checked) {
  showMetadata = checked;
  tooltip.style.display = checked ? "block" : "none";
});



// Create an HTML element that will serve as the
// tooltip that displays the metadata information
function createTooltip() {
  const tooltip = document.createElement("div");
  viewer.container.appendChild(tooltip);
  tooltip.style.backgroundColor = "black";
  tooltip.style.position = "absolute";
  tooltip.style.left = "0";
  tooltip.style.top = "0";
  tooltip.style.padding = "14px";
  tooltip.style["pointer-events"] = "none";
  tooltip.style["block-size"] = "fit-content";
  return tooltip;
}
const tooltip = createTooltip();

// Show the given HTML content in the tooltip
// at the given screen position
function showTooltip(screenX, screenY, htmlContent) {
  tooltip.style.display = "block";
  tooltip.style.left = `${screenX}px`;
  tooltip.style.top = `${screenY}px`;
  tooltip.innerHTML = htmlContent;
}

// Create an HTML string that contains information
// about the given metadata, under the given title
function createMetadataHtml(title, metadata) {
  if (!Cesium.defined(metadata)) {
    return `(No ${title})<br>`;
  }
  const propertyKeys = metadata.getPropertyIds();
  if (!Cesium.defined(propertyKeys)) {
    return `(No properties for ${title})<br>`;
  }
  let html = `<b>${title}:</b><br>`;
  for (let i = 0; i < propertyKeys.length; i++) {
    const propertyKey = propertyKeys[i];
    const propertyValue = metadata.getProperty(propertyKey);
    html += `&nbsp;&nbsp;${propertyKey} : ${propertyValue}<br>`;
  }
  return html;
}

// Install the handler that will check the element that is
// under the mouse cursor when the mouse is moved, and
// add any metadata that it contains to the label.
const handler = new Cesium.ScreenSpaceEventHandler(viewer.scene.canvas);
handler.setInputAction(function (movement) {
  if (!showMetadata) {
    return;
  }
  let tooltipText = "";
  const picked = viewer.scene.pick(movement.endPosition);

  const tilesetMetadata = picked?.content?.tileset?.metadata;
  tooltipText += createMetadataHtml("Tileset metadata", tilesetMetadata);

  const tileMetadata = picked?.content?.tile?.metadata;
  tooltipText += createMetadataHtml("Tile metadata", tileMetadata);

  const groupMetadata = picked?.content?.group?.metadata;
  tooltipText += createMetadataHtml("Group metadata", groupMetadata);

  const contentMetadata = picked?.content?.metadata;
  tooltipText += createMetadataHtml("Content metadata", contentMetadata);

  const screenX = movement.endPosition.x;
  const screenY = movement.endPosition.y;
  showTooltip(screenX, screenY, tooltipText);

// Perform the setup for a tileset that was just loaded...
function setupTileset(tileset) {
  // For implicit test...
  tileset.maximumScreenSpaceError = 800;

  // Move the tileset to a certain position on the globe,
  // and scale it up
  const transform = Cesium.Transforms.eastNorthUpToFixedFrame(
    Cesium.Cartesian3.fromDegrees(-75.152408, 39.946975, 20)
  );
  const scale = 15.0;
  const modelMatrix = Cesium.Matrix4.multiplyByUniformScale(
    transform,
    scale,
    new Cesium.Matrix4()
  );
  tileset.modelMatrix = modelMatrix;

  // Zoom to the tileset, with a small offset so that
  // it is fully visible
  const offset = new Cesium.HeadingPitchRange(
    Cesium.Math.toRadians(-22.5),
    Cesium.Math.toRadians(-22.5),
    60.0
  );
  viewer.zoomTo(tileset, offset);

  let callCounter = 0;
  //*/
  // The function that is assigned to the tileset, and that
  // governs which of the dynamic contents are "active"
  const dynamicContentUriCondition = (key) => {
    callCounter++;
    if (callCounter % 100000 === 0) {
      //console.log("URI condition called "+callCounter+" times");
    }
    const currentTime = viewer.clock.currentTime;
    const currentExampleTimeStamp =
      createExampleTimeStampString(currentTime);
    //console.log("Check ", key);
    if (currentExampleTimeStamp !== key.exampleTimeStamp) {
      return false;
    }
    if (currentExampleRevision !== key.exampleRevision) {
      return false;
    }
    //console.log("Active ", key);
    return true;
  };
  tileset.dynamicContentUriCondition = dynamicContentUriCondition;
  //*/

  // A convenience function that just checks for
  // equality of all property values
  /*/
  tileset.setSimpleDynamicContentUriCondition(
    currentExampleProperties
  );
  //*/
}
```
</p>
</details>

javagl:
One question that came up, and that is related to the broader questions about ~"schemas and type definitions": 

What should the the valid types for the properties of the "keys"? 

Until now, the examples only used the `exampleTimeStamp` and `exampleRevision`, and these had both been of the type `string`. But it was never discussed or decided which types should be allowed there.

I think that there are at least three levels of "genericity":

- They could always be `string`
- They could be `string`, `number`, or `boolean` ("primitive types")
- They could be arbitray objects

In the current state, I think that there are no explicit requirements (have to check to be 100% sure). But this is mainly because the mapping from the keys to the URLs that is computed internally is using that `NDMap` class. And this class currently happens to convert the "key" into a string, using `JSON.stringify`. But that's an implementation detail! 
(There are no proper, sensible "Maps" in JavaScript - i.e. no concept of `hashCode` or `equals`).

The same applies to the generalization (that is currently only in that branch mentioned above). It will just "pass through" the keys. That "URI condition" that is added there could do a
```
const dynamicContentUriCondition = (keys) => {​
  return keys.firstKey.some.arbitrary.nested.property == 123;
}
```
depending on the structure that is allowed or expected there.

I think that we should be careful with what we allow there. Single properties ("primitive types") could use the `===` consistently and/or be converted to `string` for efficient lookups, and therefore allow some optimized or simplified paths. Arbitrary objects may raise questions about the concept of "equality", e.g. in how far the structure of what is contained in the JSON must match or be aligned with what is used for the lookup in the _implementation_.

lilleyse:
> - They could be `string`, `number`, or `boolean` ("primitive types")

This feels the most natural to me. But I don't really have a strong justification.

javagl:
The extension was renamed from `3DTILES_dynamic` to `3DTILES_content_conditional` now. 

(The specification draft PR at https://github.com/CesiumGS/3d-tiles/pull/834 has also been updated accordingly)

lilleyse:
@javagl I pushed a small fix in https://github.com/CesiumGS/cesium/pull/12977/commits/10cd65d57b1690aa74ef242cbd72ea9099e90070

javagl:
The last few commits have been about tweaking the call and implementation to `applyStyle`, which, contrary to what one might guess, does not really "apply the style" - at least, not unconditionally. Different behaviors for "point clouds" and "non-point clouds" have been observed, and further conditions are involved as well. Debugstepping into why a certain draw command does not cause something to appear on the screen is essentially impossible. I'll try to add create and some test data for the cases that caused issues until now.

javagl:
I just saw that the latest changes cause some spec failures. I'm **so** tempted to just revert all the changes in that area, open an issue
_"The 'applyStyle' function does not apply the style"_
and leave fixing that to whoever wants to investigate that further...

javagl:
The last commits here update this branch with the latest `main`, including the resolution of some merge conflicts, as well as some minor cleanups and follow-ups for the renaming of '(time) dynamic' to "conditional content".

This includes the addition/update of the corresponding spec data. It's a simple tileset with a grid of 3x3 (conditional) contents, where the horizontal content is selected with the `exampleTimeStamp` and the vertical content is selected with the `exampleRevision`. The specs include _some_ forms of "invalid data" for testing the error handling. More extensive tests could be added (always).

For reviews and manual tests, here is a test package:

[ConditionalContent 2026-09-13 RC0001.zip](https://github.com/user-attachments/files/32161146/ConditionalContent.2026-09-13.RC0001.zip)

( EDIT: From tests with some real-world data, there had been reports indicating that _point cloud data_ (in contrast to the mesh data in the given tests) did behave slightly differently, because there are some obscure special paths for the rendering in the case of point cloud data. If there are any issues with that, additional test data with point clouds may have to be created. )

It contains four example tilesets and a sandcastle that allows to (somewhat) conveniently select them and test some basic functionality:

- ConditionalContent_0001_basic : A basic test. Hence the name
- ConditionalContent_0002_implicit : A simple _implicit_ tileset with conditional content
- ConditionalContent_0003_metadata : A simple test where each content has associated metadata. This is basically the same as the spec data. The goal is to make sure that picking and metadata access works properly, depending on the currently "active" content.
- ConditionalContent_0004_complex : A test case where "many" contents are (up to 64) "active" at certain points in time, as a basic test for the loading behavior in these cases

A short screencap showing these tests in the sandcastle:

Cesium Conditional RC0001



---

An aside: Look at https://github.com/CesiumGS/cesium/pull/12977/changes/6d4404f150cdf48b3c100d3310397f774f93ae32 in view of https://github.com/CesiumGS/cesium/issues/13476 . **tl;dr**: Things are broken. We should fix them ASAP.

Issue #12818, opened 2025-08-14
The picking of Entities by viewer.scene.pick is disordered


SmallAngel:
### What happened?

I've found that when picking Entities near the browser edges, errors occur. Multiple entities are picked even when there's no overlap.

### Reproduction steps

1.It is necessary to load terrain data and enable terrain detection.
2.Create an instance and add more than four entities through the instance.
3.Drag the rendered entities to the edge of the browser and click on the entities.
4.Vertical perspective, try it a few times.

### Sandcastle example

[...]

Issue #12508, opened 2025-03-03
Cesium displays data from other side of Earth when user selects polygon


hcarter333:
### What happened?

This issue appeared yesterday (3/2) afternoon and may be related to two other issues, #12337 and #12388 that were closed with the most recent release.

When the user clicks on a polygon on the map shown below, occasionally, the selection indicator does not lock to the polygon and instead of displaying the 'name' and 'description' fields associated with the clicked polygon,  displays information for a polygon roughly on the other side of the globe. 

![Image](https://github.com/user-attachments/assets/1c257290-1938-4fdd-b0f9-23d8f755a77d)

The following screen video shows an occurrence of the issue. Notice the following:
1. The selection indicator is not locked to a grid square (polygon)
2. The value displayed, 6464 kHZ does not correspond to the color of the square the indicator is located in. The range for that color is 10762 to 12083 kHz.
3. Double clicking on the selection indicator changes the view from Antiqua to Melbourne.
4. The selected square near Melbourne has the same value as previously displayed. 

https://github.com/user-attachments/assets/60e835dc-2bdb-4afa-a49f-6707f22c5923

This does not happn when the user selects a spire. Everything works as expected in that case.

[...]

Issue #12145, opened 2024-08-22
Picking is selecting the wrong feature


eric-g-97477:
### What happened?

I clicked on a feature on the globe and it selected the furthest object found instead of the nearest.

I expected the nearest object to be selected.

### Reproduction steps

1. Use the provided sandcastle
2. Click somewhere in the area outlined by the red square

![image](https://github.com/user-attachments/assets/00b1b920-5e3c-436b-b35d-600c07a3bce4)

3. What shows up is:

![image](https://github.com/user-attachments/assets/67e63ce3-1716-48d0-b6e9-6ff9530697f8)

4. If I click most other places in that same green polygon, I get the correct answer which is:

![image](https://github.com/user-attachments/assets/228e7ce2-0126-461f-bed1-09c54e3f8ef2)

Rotating the globe so that IDN polygon is no longer beneath the CAN polygon results in the correct behavior because only a single polygon can be found at the click location.

What I did notice is that if I set `clampToGround: false`, then the picking works correctly.
However, it would be useful to me to be able to have picking work correctly with `clampToGround: true`.

### Sandcastle example

[...]

PR #12978, opened 2025-10-11
Fix clamped polygon picking

Beilinson:
# Description

#12818 seems to be caused by shadow volumes rendering in front of each other only during picking. The check for discarding the fragment that exists on a rendered shadow does not exist in the `PICK` path of the shader.

The [regular path](https://github.com/CesiumGS/cesium/blob/1dcf97348f468bd0998f0a6a6679c24f4af5d2bd/packages/engine/Source/Shaders/ShadowVolumeAppearanceFS.glsl#L81-L89):

```glsl
#else // PICK

#ifdef CULL_FRAGMENTS
    // When classifying translucent geometry, logDepthOrDepth == 0.0
    // indicates a region that should not be classified, possibly due to there
    // being opaque pixels there in another buffer.
    if (uv.x <= 0.0 || 1.0 <= uv.x || uv.y <= 0.0 || 1.0 <= uv.y || logDepthOrDepth == 0.0) {
        discard;
    }
```

The [PICK path](https://github.com/CesiumGS/cesium/blob/1dcf97348f468bd0998f0a6a6679c24f4af5d2bd/packages/engine/Source/Shaders/ShadowVolumeAppearanceFS.glsl#L69-L77):
```glsl
#ifdef PICK
#ifdef CULL_FRAGMENTS
    // When classifying translucent geometry, logDepthOrDepth == 0.0
    // indicates a region that should not be classified, possibly due to there
    // being opaque pixels there in another buffer.
    // Check for logDepthOrDepth != 0.0 to make sure this should be classified.
    if (0.0 <= uv.x && uv.x <= 1.0 && 0.0 <= uv.y && uv.y <= 1.0 || logDepthOrDepth != 0.0) {
        out_FragColor.a = 1.0; // 0.0 alpha leads to discard from ShaderSource.createPickFragmentShaderSource
        czm_writeDepthClamp();
    }
```

Except it looks like nothing is definitively setting `out_FragColor.a=0.0` (which needs to be set in order to get discarded by `ShaderSource.createPickFragmentShaderSource` apparently), which lead me to believe the supposed _discard_ isn't happening properly.

I commented out the `discard` statement to confirm, and the results seemed to align with my theory:

500195873-ae6b0643-a684-464f-b9bf-a0bbbe88407b

All the overlapping colors are fragments that normally get discard because they should be behind the terrain, but they aren't getting discarded in the PICK path.

Example highlighting the edge of `59`:

Image

example highlighting within the area of `59` that is colored by `58` from behind:

Image

## Solution

I copied the `discard` condition over into the picking path, and this seems to have the proper result:

Example:

Image

## Issue number and link

These are all duplicates

#12145
#12508
#12818 

## Testing plan

Adapted the sandcastle from the issue. 

[sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html#c=tVgLU+M4Ev4r2tzWVlKbtZMA8wgwdUzIQLaYhCIZpq4uV6xitxPtyJZPkmEyLP99W7b8JFBczVygwJb63V+3WvFEpDS5ZXAHkhyTCO7ICBRLQuc6XWsvW176PhKRpiwCuWx1yf0y0iAlvg9z8kX27gRShJ+F5L5daHceOofLyPxQtY08EiSRp5mICPX9MxC/KxG11yD+xP+fJO8StRF3HaOBEC81zqeazkUiPUAD6R1lOtdp2U8LAocL6tekpXJQEqdhvBBnUiSRPyRaJtA1O9Y2YiPglLqUg/a1y3dDSIgEncioYhKuPiwjy50oeC/FnQJ5BZ4IQ4h88K9ACZ6kHh+TgHIFqcrMN2vqmWQ+7i5bG61jNXTdNVPaWTO9SVYoVCKxhkg7KNTlyRcIvS8M2eXWXdGgD/jZCwJ4Tfff7NPgwF8N6JtefzB4+2bflfTOffX2le/tvaKw8g96+4O3A3i7Fxzs+wfBAKjv+Qfenk/3+303pjdrNOWm3+v1Qscat2zVDa6H/AoCkBB5YH3IEvQ4t2Y7C7YN1jchwoVoPy8wh47rkrneciDGPKL+m1AJJBZ8yxGSKjcupsjPL3F5LaJxpJlmoMixSdzzahywxM4t5Qko1BoISdocNGHoVO8Q/x3tlu9wiNZ6gxS//lrDbSpzi9w72f7N/pMiigWkbeHsQ4De+O2M0Ykzhk4nB3F93Qkp1iCjHDVYASPBhUwr8IpGvgjb9yRkEQuT8ITHGzokPeegS0L6tb6UVQEWQ4plG+4T3yecroATE4lm2FE/Rr0o5TVo694ITGewLtTCsWEgMRAbE5GGJ8UWQk5fmwzkIfk94YyarIETibt2JzPUZlsoZrRjhkvhTrFqC9vE96eS9K+/Sj6bOXJ8jBkuomxLPEo4z6NS6vRS9+qNckSlxica7bV7XWJ+86ayE0MN7VXckMci0y6Uae2WvAierrWlUzXS2v5Yis9umQ/vt3OPciofC7TGdJ/wK437Q9kF4v9fif3PxfXdZdXI7FNYtjkt0JcW9eS0gmYpYpCp1ehugeSMKfs8gepuSVOqsW2yCJsBwn1OmCduaO0uJGDlxoJFekjuK0vGaGwNw3qj+Hw+WYy7NbKYfQU+Z99gSAb1HZFo02tHO+SYhjNSKn2Za8miNU4N/+j3+68He8tWZ6ecz8zXm6wh1bY3wNYbXSCn0HReX3dGFycfL28Ws5uzq9mn6WldiM8UXXE4hVhvFqD0KZ6mNJU2TcIVBvVyNp8sJtfjm8n0w2Q6Wfyryv9QvKT9rxJJQjR8xdD+8fN9nv6HPyrJC4QJ/LLVfxNrEopIqJh6gANTSfK9/qEEIdk3M4zxmWTrygR23thwRuPpYnxVYb018PQeMV7Xlh+zBYzzXXk/O5/NF3UQ/RCcNEGyX9lRZgoo5F6YBKWDgfNhcnFxczI9vZl9WlxMplWTXDeF9SwIFGB+drW4QbuPJ+Orqg3fASI0E8OJ7bbkqSidApUfqLSteICjVpf0HfwzyJ57zl6nZr2WNFI8wUr3ti+S2a8INc99I7RXCs0R3jj3TStWwMHT4F9WZ5bDykwwSpQWIblkZgRdV2YA7E/Znt1q1/p1WkrjvGk/09uaNWeuA8Nsbu5W195T78u6Oc7XinA//rqzCF9eQBfjD4uC7WXFs5hdloq+s9ZfiNrfBkVqH+qXGjt44SzIm0PL3JMA0dyEZnyLeTjPiNo2MwrBBo5Ho1uqcmkIgElA9AYwqngpIUwRgUEhNL/F5IeyqfOUDpvkF5VeltCF7GaH1GmGjUBrmIObkyhO9EkKpHaBqHaI8vESpRtHdYwAA3+2+hORWoIpM9nsFXwIMP/SnpTlweq6EgwFEdzPxwccINcbbtKSgtqQmQliZzWQX37ZXSbFaFE7e58lfXKIX1ydTOeXJ1fYiw8L2O+05phgDWQjjqV8KB1ojEDVyHWMH8/s4xDTqXhipJVN6aWznpd9a6BqU1BDS3X6yR871UMXg1ikh3iJRD3FCFppuU9Ep6GtgIGVnMSITyhBmX3sTJgXWokyj4Z400hhNuacxUowv+bcLux1qwQ1uK65WIEDuaTGIJg/V5pncb9BkwrzDpuGV4bThvs/bkrdZVz67JghCTXXpqRnGdLGcJy28QpddpkpFh5q4H5SSP4Ni2mGTzW7xTYG5+Ps03x883F2PbaXm8cH2GGr2zpKJ453RuA/WRgLqUkiedtxXA1hzDE6yl0lGGPteCrtlUduznKE1y7C/OMd36GZb6SUwp0A75lm4F623h25SF9jM19noSUzbLOcbg3Jpv/uIlt0HOfIxdfHXFoIvqKyIvFv)

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@
 - Improved performance of `scene.drillPick`. [#12916](https://github.com/CesiumGS/cesium/pull/12916)
 - Improved performance when removing primitives. [#3018](https://github.com/CesiumGS/cesium/pull/3018)
 - Improved performance of terrain Quadtree handling of custom data [#12907](https://github.com/CesiumGS/cesium/pull/12907)
-- Fixed picking of `GroundPrimitive` with multiple geometry instances selecting the wrong id. [#12978](https://github.com/CesiumGS/cesium/pull/12978)
+- Fixed picking of `GroundPrimitive` with multiple `PolygonGeometry` instances selecting the wrong instance. [#12978](https://github.com/CesiumGS/cesium/pull/12978)
 
 ## 1.134.1 - 2025-10-10
 


packages/engine/Source/Shaders/ShadowVolumeAppearanceFS.glsl
```diff
@@ -71,7 +71,9 @@ void main(void)
     // indicates a region that should not be classified, possibly due to there
     // being opaque pixels there in another buffer.
     // Check for logDepthOrDepth != 0.0 to make sure this should be classified.
-    if (0.0 <= uv.x && uv.x <= 1.0 && 0.0 <= uv.y && uv.y <= 1.0 || logDepthOrDepth != 0.0) {
+    if (uv.x <= 0.0 || 1.0 <= uv.x || uv.y <= 0.0 || 1.0 <= uv.y || logDepthOrDepth == 0.0) {
+        out_FragColor.a = 0.0;
```
Beilinson: I'm not sure why have all this weird obfuscation by having the real discard happen in `ShaderSource.createPickFragmentShaderSource`, although I tested just using discard here as well and it didn't work :(

Beilinson:
BTW, despite this issue being new this code is **quite** old (I can replicate it the issue least from Cesium `1.80`), the `createPickFragmentShaderSource` method is from 12 years ago from #643! 

Working in this codebase is so cool! 🤯

Beilinson:
**Update:**

#12145, #12508, and the issue mentioned here #12818 are all duplicates. 

I tested the sancastles from all of them, the bug occurs as described in main and is solved with this branch!

ggetz:
Fantastic, thanks for investigating this @Beilinson!

@lilleyse Could I get a sanity check on the change here? I just want to make sure I'm not missing anything regarding shadow volumes (or potentially log depth).

packages/engine/Source/Shaders/ShadowVolumeAppearanceFS.glsl
lilleyse: I'm confused by this as well. On my side all these work:

* `out_FragColor.a = 0.0;`
* `discard;`
* do nothing in code block
  * `out_FragColor` might default to `vec4(0.0)` on my system (technically undefined behavior)

lilleyse: As a sanity check I replaced the whole `PICK` block with

```
#ifdef PICK
out_FragColor.a = 1.0;
discard;
#else // PICK
```

and it honors the discard by not picking anything

Beilinson: I'll check again if a discard works maybe I made a mistake earlier.

Beilinson: Yep I was mistaken, discarding works. I also compared the previous condition in the pick path against the regular path and realized that it was a full inverse condition with the exception of the last part which is a `||` instead of a `&&`, which I think was a mistake.

I pushed up a new commit unifying the logic, it should have all the same logic as this PR had before (so it still fixes the bug from my testing)

packages/engine/Source/Shaders/ShadowVolumeAppearanceFS.glsl
```diff
@@ -65,21 +65,6 @@ void main(void)
 #endif // SPHERICAL
 #endif // TEXTURE_COORDINATES
 
-#ifdef PICK
-#ifdef CULL_FRAGMENTS
-    // When classifying translucent geometry, logDepthOrDepth == 0.0
-    // indicates a region that should not be classified, possibly due to there
-    // being opaque pixels there in another buffer.
-    // Check for logDepthOrDepth != 0.0 to make sure this should be classified.
-    if (0.0 <= uv.x && uv.x <= 1.0 && 0.0 <= uv.y && uv.y <= 1.0 || logDepthOrDepth != 0.0) {
```
Beilinson, line 74: This condition should have been instead:

```js
if (0.0 <= uv.x && uv.x <= 1.0 && 0.0 <= uv.y && uv.y <= 1.0 && logDepthOrDepth != 0.0) {
```

Which is equivalent to:
```js
if (!(uv.x <= 0.0 || 1.0 <= uv.x || uv.y <= 0.0 || 1.0 <= uv.y || logDepthOrDepth == 0.0)) {
```

But its safer to extract the CULL_FRAGMENTS to occur prior to either the pick/regular path, so that there is a unified discard for both paths, and avoid making these inverse logic problems.

Then the pick path can safely explicitly set `out_FragColor.a = 1.0`, and call `czm_writeDepthClamp()` just based on the `CULL_FRAGMENTS` path.

packages/engine/Source/Shaders/ShadowVolumeAppearanceFS.glsl
lilleyse: The new commit looks great!

lilleyse:
Thanks @Beilinson - I tested again with your recent changes. Working nicely! Once you update CHANGES.md and add a test (if straightforward) this should be ready to go.

Beilinson:
@lilleyse added changelog and added a spec. The numbers for the rectangles/view are from trial and error, it correctly fails when I either comment out the discard/revert to the previous version of the frag shader, and should pass with the new version.

By reproducing the spec I also learned a few things:

1. It seems to only happen with `PolygonGeometry`, I was unable to recreate the bug using `RectangleGeometry`
2. This is specifically a bug when there are multiple geometry instances on the same ground primitive. Rendering 2 separate ground primitives did not have this issue on the old frag shader.

lilleyse:
Looks great, thanks again @Beilinson!

Beilinson:
Awesome, glad to help!

hcarter333:
I apologize if this just hasn't made it into the released version yet, but I'm still seeing misses on selection.

https://github.com/user-attachments/assets/f81802f7-1d81-4c9f-beb4-2c2acafa9187

 Here's an example.

If I rotate the globe so the North Pole is head on, I don't see misses.



https://github.com/user-attachments/assets/9dbdd810-85a5-46ce-b49a-59227628d2fc

javagl:
@hcarter333 Indeed, this was only merged two days ago, and is not part of the release. (It will be part of the November release). If you have an additional test case, you might want to try it out based on the current `main` state of the repository (or share it as a Sandcastle, so that others can test it on this state)

Issue #630, opened 2013-04-12
Optimize picking


pjcozzi:
The fix for #583 slows down the fragment shader for picking because it executes the full fragment shader for materials.  If we know a material doesn't have any transparent parts (and doesn't write depth in the future), we can just use a simple pass-through shader.

billwritescode:
Any thoughts on addressing this? Noticing that `pick` is chewing up CPU time: http://cesiumjs.org/releases/1.30/Apps/Sandcastle/?src=Hello%20World.html&label=Showcases&gist=09ea27d7643dceaacded4fe1f31a9471

pjcozzi:
@billwritescode the biggest win will come from not stalling the CPU and starving the GPU with WebGL's `readPixels`.  I think this is possible with WebGL 2, #797, and we'll investigate when we ramp up WebGL 2.

skaughtx0r:
10 years later, this is still has a huge performance impact. What's the status on this issue now that the WebGL Roadmap is closed?

ggetz:
@skaughtx0r We're putting together our roadmap for our next year of development starting in March, and are hoping to take advantage of the fact that WebGL2 is now the default in CesiumJS to do some optimizations like this.

tmarti:
It should be possible to convert `gl.readPixels` into async operations by wrapping the check for a webgl2 fence synchronization into an async Promise.

The result would be that, when the mouse moves, the feedback would not be synchronous with the mouse movement, but instead the feedback would come after the data is available, so a bit extra end-to-end time.

The nice thing with this, though, is that there would be no stalling of the GPU pipeline, and you could keep moving the camera while the pixels are being read 🙂.

Mention me if you want a code snippet example in this thread, it’s actually a small bit of code but of course in the consuming side it needs to be taken into account that the pixel reading operation would need to be `await`-ed .

skaughtx0r:
@tmarti could you post the code snippet you mentioned?

PR #12983, opened 2025-10-16
Added Scene.pickAsync for non GPU blocking picking

alarkbentley:
# Description

Scene.pick is internally using gl.readPixels that performs a synchronous GPU readback which is blocking the main render thread and very costly especially on low end GPUs. With WebGL2 a non GPU blocking method is available that uses PBO and Sync operations to asynchronously readback the pixels data without blocking.

Reference: https://registry.khronos.org/webgl/specs/latest/2.0/
Section: 3.7.10 Reading back pixels

This PR implements support for this and exposes it under a new API method Scene.pickAsync -> Promise

Asynchronous picking gives precedence to the rendering over returning the pick information. This introduces a slight delay compared to using the blocking pick method. Depending on what you prefer there might be value in having both options available in API.

## Example

```javascript

});
const scene = widget.scene;

handler.setInputAction(function(movement) {
    scene.pickAsync(movement.position).then((feature) => {
         feature.color = Cesium.Color.YELLOW;
    });
}, Cesium.ScreenSpaceEventType.LEFT_CLICK);
```

## Demo

Browser: Microsoft Edge Version 141.0.3537.71 (Official build) (64-bit)
GPU: ANGLE (Intel, Intel(R) Arc(TM) Pro Graphics (0x00007D55) Direct3D11 vs_5_0 ps_5_0, D3D11)

https://github.com/user-attachments/assets/56e5bf40-2a4f-4343-8a66-34b18f1ce091

### Profiling results

As seen in the synchronous picking (left), readpixels can stall for longer time than it takes to render the entire frame. In this example 46% of the total execution time is taken by readpixels.

In the asynchronous example (right) the bottleneck is the actual frame rendering which makes more sense.

edge_profile_sync_async

## Issue number and link

https://github.com/CesiumGS/cesium/issues/630

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Source/Renderer/Context.js
+++ b/packages/engine/Source/Renderer/Context.js
@@ -1447,11 +1447,12 @@ Context.prototype.endFrame = function () {
  *
  * Options defining a rectangle to read pixels from.
  *
+ * @private
  * @property {number} [x=0] The x offset of the rectangle to read from.
  * @property {number} [y=0] The y offset of the rectangle to read from.
  * @property {number} [width=this.drawingBufferWidth] The width of the rectangle to read from.
  * @property {number} [height=this.drawingBufferHeight] The height of the rectangle to read from.
- * @property {any} [framebuffer] The framebuffer to read from. If undefined, the read will be from the default framebuffer.
+ * @property {FrameBuffer|undefined} [framebuffer] The framebuffer to read from. If undefined, the read will be from the default framebuffer.
  */
 
 /**


packages/engine/Source/Renderer/Context.js
```diff
@@ -1460,35 +1462,70 @@ Context.prototype.readPixels = function (readState) {
   const width = readState.width ?? this.drawingBufferWidth;
   const height = readState.height ?? this.drawingBufferHeight;
   const framebuffer = readState.framebuffer;
+  const pbo = readState.pbo ?? false;
+
+  if (pbo && !this._webgl2) {
+    throw new DeveloperError("A WebGL 2 context is required.");
```
mzschwartz5: This error could be slightly more descriptive. E.g. "A WebGL 2 ocntext is required to read pixels using a PBO".

packages/engine/Source/Renderer/Context.js
```diff
@@ -1460,35 +1462,70 @@ Context.prototype.readPixels = function (readState) {
...
   }
 
-  const pixels = PixelFormat.createTypedArray(
-    PixelFormat.RGBA,
-    pixelDatatype,
-    width,
-    height,
-  );
+  let pixels;
+  if (pbo) {
```
mzschwartz5: Style comment - I'm a stickler for simplifying conditional logic when possible.

This function has basically become two different functions depending on whether `pbo` is passed in. Let's formalize that:

```JS

const readPixelsFunc = pbo ? readPixelsAsync() : readPixelsSync();
readPixelsFunc();
```

Then we have just a single branching point instead of multiple.

mzschwartz5: These two funcs can take `pixelFormat`, `pixelDatatype` etc. as args

packages/engine/Source/Renderer/Sync.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+function Sync(options) {
+  options = options ?? Frozen.EMPTY_OBJECT;
+
+  //>>includeStart('debug', pragmas.debug);
+  Check.defined("options.context", options.context);
+  //>>includeEnd('debug');
+
+  if (!options.context._webgl2) {
+    throw new DeveloperError("A WebGL 2 context is required.");
```
mzschwartz5: Same comment as earlier about using a more descriptive error.

packages/engine/Source/Renderer/Sync.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const context = options.context;
```
mzschwartz5: Small nit- let's move this up higher so we don't need to avoid the multiple `options.context`

packages/engine/Source/Renderer/Sync.js
```diff
@@ -0,0 +1,46 @@
+import Check from "../Core/Check.js";
+import destroyObject from "../Core/destroyObject.js";
+import DeveloperError from "../Core/DeveloperError.js";
+import Frozen from "../Core/Frozen.js";
+import WebGLConstants from "../Core/WebGLConstants.js";
+
+/**
+ * @private
+ */
```
mzschwartz5, line 18: Even though this is private, JSDoc would still be nice. From the file path, I see this is some sort of syncing related to rendering, but when is it used? Is it just for syncing `readPixels`? Is it more general - any time you want a GL fence?

packages/engine/Source/Renderer/Sync.js
```diff
@@ -0,0 +1,46 @@
...
+  const gl = context._gl;
+
+  const sync = gl.fenceSync(WebGLConstants.SYNC_GPU_COMMANDS_COMPLETE, 0);
+
+  this._gl = gl;
+  this._sync = sync;
+}
+Sync.create = function (options) {
+  return new Sync(options);
+};
```
mzschwartz5, line 41: Partially for my edification - why purpose does having a factory method serve here? I do see this pattern in other Cesium classes, but is it _good_? There's a decent amount of literature out there that would argue against the use of static methods when possible, but to be honest, I've never really resonated with or against that.

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -27,47 +31,18 @@ function PickFramebuffer(context) {
...
-  const context = this._context;
-  const pixels = context.readPixels({
-    x: screenSpaceRectangle.x,
-    y: screenSpaceRectangle.y,
-    width: width,
-    height: height,
-    framebuffer: this._fb.framebuffer,
-  });
-
+function colorScratchForObject(context, pixels, width, height, limit = 1) {
```
mzschwartz5: I'm not a huge fan of the rename of this function. 

The description: "Return the picked objects rendered within a given rectangle" and the name "colorScratchForObject" don't seem to be related to each other at all.

I also feel that "scratch" is not normally used in function names. It's typically up to the caller to decide to pass in scratch variables or to let the function `new` the variables for them, if they choose not to pass in scratch variables.

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+/**
+ * Creates a callback function that will once per frame poll the Sync state until signaled.
+ *
+ * @private
+ * @param {object} pickState An object with the following properties:
+ * @param {number} [readState.frameState] The active framestate.
+ * @param {number} [readState.frameNumber] The current frame number.
+ * @param {number} [readState.sync] The Sync object to poll.
+ * @param {number} [readState.ttl=10] Max number of frames to poll until reject.
+ * @param {function} onSignalCallback Callback to execute on Sync Signal.
```
mzschwartz5: It would be good to consolidate these params into a type definition for `pickState`

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+    const frameState = pickState.frameState;
+    const ttl = pickState.ttl ?? 10;
+    const syncStatus = sync.getStatus();
+    const signaled = syncStatus === WebGLConstants.SIGNALED;
+    const frameDelta = frameState.frameNumber - pickState.frameNumber; // how many frames passed since inital request
+    if (signaled || frameDelta > ttl) {
+      sync.destroy();
+      onSignalCallback(signaled);
+    } else {
+      frameState.afterRender.push(createAsyncPick(pickState, onSignalCallback));
```
mzschwartz5: Shouldn't the `pickState`'s ttl be decreased by one before pushing here? Or is that done somewhere else.

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+    }
+  };
+}
+
+PickFramebuffer.prototype.begin = function (screenSpaceRectangle, viewport) {
```
mzschwartz5, line 101: Was this function just moved or did it also change?

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+  this._passState.framebuffer = this._fb.framebuffer;
+
+  this._passState.viewport.width = width;
+  this._passState.viewport.height = height;
+
+  return this._passState;
+};
+
+/**
+ * Return the picked objects rendered within a given rectangle using asynchronously without staling the GPU.
```
mzschwartz5:
```suggestion
 * Return the picked objects rendered within a given rectangle using asynchronously without stalling the GPU.
```

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+ *
+ * @param {BoundingRectangle} screenSpaceRectangle
+ * @param {FrameState} frameState
+ * @param {number} [limit=1] If supplied, stop iterating after collecting this many objects.
+ * @returns {object[]} A list of rendered objects, ordered by distance to the middle of the rectangle.
```
mzschwartz5: The @returns isn't quite accurate, since this method returns a promise.

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+          context,
+          pixels,
+          width,
+          height,
+          limit,
+        );
+        if (signaled) {
+          resolve(pickedObjects);
+        } else {
+          reject("Picking Request Timeout");
```
mzschwartz5: Shouldn't the promise be rejected earlier, before reading back the buffer data and determining the picked objects, if the request timed out?

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,152 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+};
+
+/**
+ * Return the picked objects rendered within a given rectangle.
+ *
+ * @param {BoundingRectangle} screenSpaceRectangle
+ * @param {number} [limit=1] If supplied, stop iterating after collecting this many objects.
+ * @returns {object[]} A list of rendered objects, ordered by distance to the middle of the rectangle.
+ */
+PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
```
mzschwartz5, line 198: Same question as I left on the `begin` method - this is just moved, not changed, right?

packages/engine/Source/Scene/Picking.js
```diff
@@ -273,21 +273,24 @@ function computePickingDrawingBufferRectangle(
...
   limit = 1,
+  async,
 ) {
   //>>includeStart('debug', pragmas.debug);
   Check.defined("windowPosition", windowPosition);
   //>>includeEnd('debug');
 
   const { context, frameState, defaultView } = scene;
   const { viewport, pickFramebuffer } = defaultView;
+  async = async ?? false;
```
mzschwartz5: This could just be a default value in the function argument.

packages/engine/Source/Scene/Scene.js
```diff
@@ -3932,14 +3932,15 @@ function callAfterRenderFunctions(scene) {
   // Functions are queued up during primitive update and executed here in case
   // the function modifies scene state that should remain constant over the frame.
   const functions = scene._frameState.afterRender;
-  for (let i = 0; i < functions.length; ++i) {
-    const shouldRequestRender = functions[i]();
+  const functionsCpy = functions.slice(); // Snapshot before iterate allows callbacks to add functions for next frame
+  functions.length = 0;
```
mzschwartz5, line 3936: I'm not sure this is _strictly_ necessary, though it could be a good idea still.

Typically, new callbacks are added by `push`, so they would go to the end of the `afterRender` array. If you instead change:

`for (let i = 0; i < functions.length; ++i)`

to

```
const numFunctions = functions.length;
for (let i = 0; i < numFunctions; ++i)
```

then you can safely iterate over the functions, even while new ones are added to the end. (But then, rather than clearing the list, you'd have to slice off the beginning `numFunctions` of it).

However, two considerations:
1. Nothing stops a callback from inserting or otherwise mutating the `afterRender` list... so the above idea isn't totally safe. (We could make the `afterRender` list private and expose a single method to append to it).
2. This might be over-optimizing. I'm not sure what the typical size of the `afterRender` function list is, but it's probably not terribly expensive to copy it. (Though it does happen *every* frame).

packages/engine/Source/Scene/Scene.js
```diff
@@ -4517,6 +4518,27 @@ Scene.prototype.pick = function (windowPosition, width, height) {
...
+  return new Promise((resolve, reject) => {
+    this._picking
+      .pick(this, windowPosition, width, height, 1, true)
+      .then((result) => {
+        resolve(result[0]);
+      })
+      .catch(reject);
+  });
+};
+
```
mzschwartz5:
```suggestion
Scene.prototype.pickAsync = async function (windowPosition, width, height) {
  const result = await this._picking.pick(this, windowPosition, width, height, 1, true);
  return result[0];
};
```

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
```diff
@@ -133,6 +134,34 @@
...
+            debounceTime = now;
+          }
+          if (async) {
+            result[0] = await scene.pickAsync(position);
+          } else {
+            result[0] = scene.pick(position);
+          }
+          if (debounce) {
+            picking = false;
+          }
```
mzschwartz5: Perhaps debouncing is something that should be handled internally by the implementation? (As an option, perhaps?)

Or maybe the internal implementation should use a `set` of results each frame, rather than an array, so that debouncing isn't even necessary (duplicates across frames will hash to the same entry)

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
mzschwartz5, line 1: I don't love the bloat this introduces to the sandcastle. I'm honestly not even sure if such a change is really necessary to show off in a sandcastle. I'm not sure I was even able to tell the difference between (a)sync in the video of this PR's description.

packages/engine/Source/Scene/Picking.js
```diff
@@ -273,21 +273,24 @@ function computePickingDrawingBufferRectangle(
  * @param {number} [width=3] Width of the pick rectangle.
  * @param {number} [height=3] Height of the pick rectangle.
  * @param {number} [limit=1] If supplied, stop iterating after collecting this many objects.
- * @returns {object[]} List of objects containing the picked primitives.
+ * @param {boolean} [async=false] Use async non GPU blocking picking.
```
mzschwartz5: Should this param doc indicate that WebGL 2 is a requirement?

packages/engine/Source/Scene/Scene.js
```diff
@@ -4517,6 +4518,27 @@ Scene.prototype.pick = function (windowPosition, width, height) {
   return this._picking.pick(this, windowPosition, width, height, 1)[0];
 };
 
+/**
+ * Performs the same operation as Scene.pick but asynchonosly without blocking the main render thread.
```
mzschwartz5, line 4522: This JS doc should indicate that WebGL 2.0 is a prerequisite here.

packages/engine/Source/Renderer/Sync.js
alarkbentley: I saw this comment in Texture.js, why I followed this pattern.
https://github.com/CesiumGS/cesium/blob/074acf176a4add690a0f1a098fb2e7f587b91d2d/packages/engine/Source/Renderer/Texture.js#L581

_"This function is identical to using the Texture constructor except that it can be replaced with a mock/spy in tests."_

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: Not touched, just moved

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: Correct, only change is that it now has some extracted code into colorScratchForObject

packages/engine/Source/Scene/Picking.js
alarkbentley: I followed this recommendation and thought we for some reason want to avoid default function argument assignments:
https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#default-parameter-values

Did I misunderstand?

packages/engine/Source/Scene/Scene.js
alarkbentley: I understand the idea, but would slicing after the loop actually give any benefits over slicing before? (slice = copy)
I feel that current implementation is less error prone.

I thought about juggling two arrays to avoid memory allocations, but felt it would be over-optimizing.

packages/engine/Source/Renderer/Sync.js
javagl: There are some language-specific aspects here (I'm usually in the "No `public` constructors, period!"-camp, FWIW).

The mock/spy pointis one that sounds pretty convincing here. (Beyond others: Evolvability (like adding some  `createWithSomeDefaultValue` later without breaking anything), or anything that is related to asynchronicity or possible things that may be `throw`n during the construction...)

mzschwartz5: Hmm, I think it's generally considered best practice not to write your source code specifically to facilitate testing. Like, source code should be designed in a way to be testable (e.g. it should be modular), but usually it's discouraged to write methods specifically to facilitate testing.

packages/engine/Source/Scene/Picking.js
mzschwartz5: Good question, let me see what others think / whether this is outdated advice.

packages/engine/Source/Scene/Scene.js
mzschwartz5: I guess the main benefit would be that rather than copying `numFunctions`, you could just copy the callbacks that were added during iteration, of which there are likely fewer. 

But it's probably not worth the extra risks involved (which _could_ be made safe, but with more effort). Feel free to resolve this issue if you'd like, I think either approach is fine.

mzschwartz5: This change probably qualifies for its own comment in `CHANGES.md`, by the way - something along the lines of "users can now add new post-render callbacks within existing callbacks"

packages/engine/Source/Scene/Picking.js
mzschwartz5: From a teammate:

> the behavior between ?? and function foo(bar = "baz") is [different for null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#passing_undefined_vs._other_falsy_values). nearly all the time we treat null the same as undefined in cesiumjs so we can't just use the default parameter option

So no changes needed here. It's unfortunate, but that's just the way things are in Cesium.

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: Renamed to pickObjectsFromPixels

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: removed the pickstate entirely and replaced with individual params.

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: good catch

packages/engine/Source/Renderer/Context.js
alarkbentley: Not sure if I agree that this would be simpler, could you perhaps show in a commit?
If we go this route I would rather create two separate prototype functions. But they have a lot in common.

packages/engine/Source/Scene/Picking.js
alarkbentley: Implemented a fallback so if webgl1 context it will internally do synchronous picking but return as a promise.

packages/engine/Source/Scene/Scene.js
alarkbentley: Implemented a fallback so if webgl1 context it will internally do synchronous picking but return as a promise.

packages/engine/Source/Renderer/Context.js
alarkbentley: updated, please review

packages/engine/Source/Renderer/Sync.js
alarkbentley: updated, please review

packages/engine/Source/Renderer/Sync.js
alarkbentley: updated, please review

packages/engine/Source/Renderer/Sync.js
alarkbentley: updated, please review

packages/engine/Source/Scene/Picking.js
alarkbentley: updated, please review

packages/engine/Source/Scene/Scene.js
alarkbentley: updated, please review

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: updated, please review

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: updated, please review

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: updated, please review

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: updated, please review

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: updated, please review

packages/engine/Source/Renderer/Buffer.js
```diff
@@ -72,6 +72,24 @@ function Buffer(options) {
   this.vertexArrayDestroyable = true;
 }
 
+Buffer.createPixelBuffer = function (options) {
+  //>>includeStart('debug', pragmas.debug);
+  Check.defined("options.context", options.context);
+  //>>includeEnd('debug');
+
+  if (!options.context._webgl2) {
+    throw new DeveloperError("A WebGL 2 context is required.");
```
mzschwartz5: This error could still use a more descriptive message.

CHANGES.md
```diff
@@ -22,6 +22,7 @@
 - Improved performance when removing primitives. [#3018](https://github.com/CesiumGS/cesium/pull/3018)
 - Improved performance of terrain Quadtree handling of custom data [#12907](https://github.com/CesiumGS/cesium/pull/12907)
 - Fixed picking of `GroundPrimitive` with multiple `PolygonGeometry` instances selecting the wrong instance. [#12978](https://github.com/CesiumGS/cesium/pull/12978)
+- Fixed infinite recursion that would happen if user append post-render callbacks within existing callbacks
```
mzschwartz5: Don't forget to add your other contribution here :) (the main one, async pikcing)

packages/engine/Source/Scene/PickFramebuffer.js
mzschwartz5: I think I miscommunicated a bit; I was just suggesting using `@typedef` to document an object more cohesively. (You can find many examples of this in the codebase by searching for `@typedef`).

But what you did is also fine. Up to you!

mzschwartz5:
I haven't reviewed your unit tests yet, but given some of our conversations, I'm guessing we probably need more testing around:
1. TTL on the async pick (some unit test should have picked up on the fact that it was never being decremented)
2. (Maybe) Rejection of the async pick promise

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
alarkbentley: I decided to avoid debouncing internally because there are some tricky special cases that I want to leave open to the application logic.

Say you pick 2 different pixels [x0,y0] and [x1,y1] at the same frame. I think the API default should allow that (maybe you are doing multi picking?) but in many cases (like the Sandcastle demo) we want to debounce [x0,y0] as [x1,y1] is more up to date and you are only interested in one result.

Hashing based on [x, y, frame] would be possible but that is a very rare case and doesn't happen in Sandcastle demo as picking is only triggered on mouse move. Eg. I would still need to keep the debounce logic in the demo.

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
alarkbentley: Its an easy way to test the functionality and compare how sync/async performs on your machine.
Do you want me to remove?

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
mzschwartz5: I think there are still better options than debouncing here. Why don't you just let all the pick requests go through and whichever is the last to resolve in a given frame is the thing that's picked / highlighted? Sure, it's non-deterministic, but it probably looks reasonably good, no?

OR, append all the promises to an array, and register a post-render event that selects the last resolved promise in the array to use as the pick, then clear everything before and including that. This is still a little complicated though, and is counterproductive to my desire for simple sandcastle demos.

I still think that we don't really need an example of async picking at all, to be honest. It's pretty difficult to tell the difference, anyway. My vote is just to omit the sandcastle changes entirely.

packages/engine/Source/Renderer/Buffer.js
```diff
@@ -72,6 +72,26 @@ function Buffer(options) {
...
+      "A WebGL 2 context is required to create PixelBuffers.",
+    );
+  }
+
+  return new Buffer({
+    context: options.context,
+    bufferTarget: WebGLConstants.PIXEL_PACK_BUFFER,
+    typedArray: options.typedArray,
+    sizeInBytes: options.sizeInBytes,
+    usage: options.usage,
```
mzschwartz5, line 91: Can a PBO ever have a usage other than `DynamicRead`? If not, it should just be a constant here rather than part of options. 

I ask because I noticed the `readPixels` function simply passing in DynamicRead as a hard coded option value.

packages/engine/Source/Renderer/Context.js
```diff
@@ -1460,35 +1464,72 @@ Context.prototype.readPixels = function (readState) {
...
+    pixels._bind();
+    gl.readPixels(
+      x,
+      y,
+      width,
+      height,
+      pixelFormat,
+      PixelDatatype.toWebGLConstant(pixelDatatype, this),
+      0,
```
mzschwartz5: Am I reading this wrong? [Isn't the 7th argument supposed to be pixels](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/readPixels)? It looks like that was skipped and we go straight to the 8th argument (`dstOffset`)

packages/engine/Source/Scene/PickFramebuffer.js
```diff
@@ -121,6 +96,146 @@ PickFramebuffer.prototype.end = function (screenSpaceRectangle, limit = 1) {
...
+    if (signaled || ttl <= 0) {
+      sync.destroy();
+      onSignalCallback(signaled);
+    } else {
+      frameState.afterRender.push(
+        createAsyncPick(frameState, sync, onSignalCallback, ttl - 1),
+      );
+    }
+  };
+}
```
mzschwartz5: Does this seem like a utility function that should be generalized and go in the Sync class? I could imagine other use cases of `Sync` where devs would want such a polling function to check the state of their sync each frame and do some callback in it.

packages/engine/Source/Scene/Picking.js
```diff
@@ -335,7 +340,19 @@ Picking.prototype.pick = function (
...
+    pickedObjects = pickFramebuffer.endAsync(
+      drawingBufferRectangle,
+      frameState,
+      limit,
+    ); // Promise<Object[]>
+  } else {
+    pickedObjects = pickFramebuffer.end(drawingBufferRectangle, limit); // Object[]
+    if (async) {
+      pickedObjects = Promise.resolve(pickedObjects); // Promise<Object[]> fallback if not WebGL2
+    }
```
mzschwartz5: Should we log some sort of warning in this case (not an error, just a warning) that the user's wish to pick asynchronously was not respected because of a lack of webgl2 context?

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
mzschwartz5: If it didn't add a lot of code and complexity to the sandcastle, I'd be for it. I just don't want to make these sandcastles harder to understand; they're supposed to be easy to digest and learn from (something we don't always do well at...). So yeah, my vote would likely be to remove the sandcastle changes.

____

Maybe it would be short enough without the debouncing. Why don't you just let all the pick requests go through and whichever is the last to resolve in a given frame is the thing that's picked / highlighted? Sure, it's non-deterministic, but it probably looks reasonably good, no?

packages/engine/Source/Renderer/Context.js
mzschwartz5: > If we go this route I would rather create two separate prototype functions.

Honestly, I think that would be the ideal route. Two prototype funcs that reuse some module-internal readPixels function for shared logic. Visually, right now we have:

Image

And what you suggested (and I support) is the opposite:

Image

packages/engine/Source/Renderer/Buffer.js
alarkbentley: There is nothing in the WebGL API that prevents from using other usage types so I suggest we keep the option open.
For readback purposes however the recommendation is DynamicRead.

If you want we could default it to DynamicRead if not specified?

https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/bufferData#usage

packages/engine/Source/Renderer/Context.js
alarkbentley: The pbo version of readpixels does not use the pixels argument:

image

packages/engine/Source/Scene/PickFramebuffer.js
alarkbentley: Feels a bit overengineering but I can give it a try. Would need to abstract away frameState.

packages/engine/Source/Scene/Picking.js
alarkbentley: Sounds like a good idea, I will add.

packages/engine/Source/Scene/PickFramebuffer.js
mzschwartz5: I think it would be fine to keep `framestate` as part of the signature, personally. I just want to avoid having other people in the future define very similar functions in other classes. Having a function in `Sync.js` that accepts framestate so that it can push an `afterRender` function is still general, in my opinion.

alarkbentley: Would be the simplest solution but are we fine with creating a cross dependency between /Renderer/ and /Scene/ ?
Have not seen this pattern elsewhere. I would assume Renderer is a more low level API.

mzschwartz5: That's a fair criticism. I guess you're right; the way to keep it general and agnostic of /Scene would be to change the `frameState` argument to something like `placeToRepoll` or something (and simply pass in `frameState.afteRender` in most cases)

alarkbentley: Alternatively if we have some other place to put utils that are both Scene and Renderer aware. Let me know in that case.

mzschwartz5: Not that I'm aware of, but that doesn't mean it doesn't exist!

mzschwartz5: I do feel like it's fair to put this in Sync.js though, and in this more general form, you could have the repolling occur on some other schedule other than after every frame (say, maybe as part of some list that gets invoked every X seconds)

Apps/Sandcastle/gallery/3D Tiles Feature Picking.html
alarkbentley: Reverted the changes from sandcastle

packages/engine/Source/Renderer/Context.js
```diff
@@ -1451,7 +1451,7 @@ Context.prototype.endFrame = function () {
  * @property {number} [y=0] The y offset of the rectangle to read from.
  * @property {number} [width=this.drawingBufferWidth] The width of the rectangle to read from.
  * @property {number} [height=this.drawingBufferHeight] The height of the rectangle to read from.
- * @property {Framebuffer} [framebuffer] The framebuffer to read from. If undefined, the read will be from the default framebuffer.
+ * @property {any} [framebuffer] The framebuffer to read from. If undefined, the read will be from the default framebuffer.
```
javagl: `{FrameBuffer|undefined}` ?

alarkbentley:
> I think this is good to go! Thanks for the huge contribution. Async picking might also open the door for some other performance optimizations I've been thinking about lately.
> 
> Anything else to do or can I merge?

All good on my side! Glad to be able to contribute and thanks for the thoughtful review.
I also think this can be useful for other types of texture sampling related to camera movement or elevation queries. Ideally we should be avoiding all blocking readpixel calls. 

Let me know if a position opens up on the Cesium.js dev team.

mzschwartz5:
> I also think this can be useful for other types of texture sampling related to camera movement or elevation queries.

This is literally exactly what I had in mind (see [this issue](https://github.com/CesiumGS/cesium/issues/13005))

UniquePanda:
@mzschwartz5 Quick question: I guess this PR is not part of 1.135 and was only accidentally added to the changelog of that version, because it was created before the release? :)

mzschwartz5:
@UniquePanda Shoot, good catch - I missed that. I'll fix it in a sec.

packages/engine/Source/Renderer/Context.js
```diff
@@ -1467,12 +1543,14 @@ Context.prototype.readPixels = function (readState) {
...
 
   let pixelDatatype = PixelDatatype.UNSIGNED_BYTE;
+  let pixelFormat = PixelFormat.RGBA;
   if (defined(framebuffer) && framebuffer.numberOfColorAttachments > 0) {
     pixelDatatype = framebuffer.getColorTexture(0).pixelDatatype;
+    pixelFormat = framebuffer.getColorTexture(0).pixelFormat;
   }
 
   const pixels = PixelFormat.createTypedArray(
-    PixelFormat.RGBA,
```
alarkbentley, line 1475: Potential regression?

mzschwartz5: Did some testing, doesn't seem like it (which makes sense...)

Issue #12951, opened 2025-10-04
new Cesium Sandcastle


aipinsui:
### What happened?

I download cesium 1.134 source code, when i compiled, new sandcastle2 gallery some content can not display, i do not konw my step is wrong?

<img width="2559" height="1512" alt="Image" src="https://github.com/user-attachments/assets/32ae1b09-b877-40c1-b251-a2f3c8026d0d" />

### Reproduction steps

1.npm install
2.npm run build
3.npm run build-release
4.npm run build-apps
5.npm run build-sandcastle

### Environment

Browser:141.0.7390.55
CesiumJS Version:1.134
Operating System:windows 11

aipinsui:
the missing content is filter labels and gallery items 
<img width="2559" height="1512" alt="Image" src="https://github.com/user-attachments/assets/d9e7a5bb-c107-49f6-8a80-f7a8be76a9df" />

aipinsui:
I suspect releative path Apps/Sandcastle2/gallery/pagefind, some file is not build use npm run build-sandcastle command

javagl:
I also observed this error. And just a shot in the dark, trying to narrow down a few things: Is it correct that you are using this in an environment (Operating system _or_ browser) that is _not_ English? (I'm using a Chrome in German, and suspect that it is related to that...)

aipinsui:
> I also observed this error. And just a shot in the dark, trying to narrow down a few things: Is it correct that you are using this in an environment (Operating system _or_ browser) that is _not_ English? (I'm using a Chrome in German, and suspect that it is related to that...)
> 

My Operating System is windows11 Simple Chinese,browse is Chrome latest English,but i think this question is not relate with it,because official website display is right

ggetz:
@jjspace Could you please take a look and see if anything jumps out?

jjspace:
I'm really not sure what would cause this unless there are errors from running `npm run build-sandcastle` or more info is provided. I would suggest trying it on my branch for the build updates and see if it still happens there @aipinsui https://github.com/CesiumGS/cesium/pull/12904

javagl:
I _think_ that it might be related to the OS/Browser language, because I also stumbled over this, and this "pagefind" thingy seems to make some assumptions. In fact, had some trouble compiling/building/running the latest state at all. The stack trace from that "pagefind" is unreadable and obfuscated, though. I'll try again with a fresh clone, maybe it's possible to pinpoint what's wrong there.

[...]

PR #12988, opened 2025-10-19
Draft for sandcastle gallery pattern fix

javagl:
This is a **draft**, not supposed to be merged (in its current form). It is just intended for showing how to resolve https://github.com/CesiumGS/cesium/issues/12951

@jjspace This was another `/`-vs`\`-issue. All the path handling on Windows has to use `\`, and getting that sorted out with the mix of paths and URLs is already tricky enough. But then, `globby` comes along and casually says that it only accepts `/` in its patterns, so ... at this point, all the paths that have previously been carefully treated as such, with `\`'s everywhere, have to undergo some `\`-to-`/`-replacement again.

The solution here is a draft, with debug logs and ... _comments_. The debug logs could be removed. 

The globby documentation talks about `posix.join` for joining paths, but I'm not sure what's the best solution here, so I just used `replace`, because I felt like...

`¯\_(ツ)_/¯` or
`¯/_(ツ)_/¯` or
`¯\_(ツ)_\¯` or
`¯/_(ツ)_\¯`

---

An aside:

The wrong globby pattern caused it to not find any gallery definitions, saying
```
Successfully built gallery list.
Processed 0 gallery examples.
```
during the build. When afterwards opening the sandcastle, it showed an error in the Browser console:
```
Uncaught (in promise) Error: Pagefind Error: No language indexes found.
    at PagefindInstance.findIndex (pagefind.js:2:355)
    at PagefindInstance.init (pagefind.js:1:20007)
    at async Pagefind.init (pagefind.js:6:7599)
```

**This is independent of the search pattern fix!**

The point is: It should not show this error at runtime, _even if_ no gallery files have been found.
(But I don't have the slightest clue what's the best place for handling this...)

### DIFF
--- a/packages/sandcastle/scripts/buildGallery.js
+++ b/packages/sandcastle/scripts/buildGallery.js
@@ -118,7 +118,7 @@ export async function buildGalleryList(options = {}) {
   };
 
   const galleryFilesGlobbyPatterns = galleryFilesPattern.map((pattern) => {
-    // The join function will return the pah in a form that is normalized
+    // The join function will return the path in a form that is normalized
     // for the OS, meaning that it contains backslashes "\" on Windows
     const baseGlobbyPattern = join(rootDirectory, pattern, "**/*");
 


ggetz:
@jjspace Could you please review when you get the chance?

javagl:
@jjspace I'd assume that this is not really "reviewed as a PR", but maybe a change similar to the one here could go into https://github.com/CesiumGS/cesium/pull/12904 ...

packages/sandcastle/scripts/buildGallery.js
```diff
@@ -117,9 +117,31 @@ export async function buildGalleryList(options = {}) {
     return condition;
   };
 
-  const galleryFiles = await globby(
-    galleryFilesPattern.map((pattern) => join(rootDirectory, pattern, "**/*")),
-  );
+  const galleryFilesGlobbyPatterns = galleryFilesPattern.map((pattern) => {
+    // The join function will return the path in a form that is normalized
+    // for the OS, meaning that it contains backslashes "\" on Windows
+    const baseGlobbyPattern = join(rootDirectory, pattern, "**/*");
```
jjspace, line 123: @javagl I think the better solution is to just swap this for `posix.join` as suggested by the globby docs. Did you try that? any reason not to?

javagl: I tied it quickly, and I think that I just stubled over some issue with how to write the `import` properly 😊 There are a few things that I'd look into (e.g. does `posix.join("foo/bar", "baz/buz")` result in `"foo/bar\baz/buz"`, or does it also _replace_ the `/`'s from the existing parts?), but **iff** it performs some normalization (i.e. that 'replace-all'), then that's probably more idiomatic.

jjspace: for now you could just do `import { basename, dirname, join, relative, posix } from "node:path";` for it. Maybe a better way long term, just wanted to know if it worked. I believe the last time I looked into windows path issues it _did_ do the normalization inside existing paths. easy to verify but I'm not on windows to test

javagl: Regarding the import: I _think_ that I was trying to avoid importing `posix`, because we only need `join`, and was looking for something along the lines of
`import { ..., posix.join as posixJoin } from "node:path";`
or some sort of
`import { join as posixJoin } from "node:path:posix";`
or so, but... that's not really important.

---

What's more important: `posix.join` does _not_ do a normalization. For example, when using `posix.join` for the joining, the output for me is
```
Create globby pattern
rootDirectory  C:\Develop\CesiumGS\cesium\packages\sandcastle
pattern  gallery
baseGlobbyPattern  C:\Develop\CesiumGS\cesium\packages\sandcastle/gallery/**/*
globbyPattern  C:/Develop/CesiumGS/cesium/packages/sandcastle/gallery/**/*
Successfully built gallery list.
```
So the join itself does insert the `/` between `sandcastle` and `gallery`, but the prefix is still wrong. (There might be some alternative to the `replace` call, maybe some `normalize` or whatnot, but I used `replace` mainly to check that it works when ~"something like this is done" (no matter how...)).

jjspace:
I'm going to close this in favor if #13026. Thanks for the preliminary work @javagl

Issue #12186, opened 2024-09-07
Unused eslint directives are not detected


javagl:
There are a few `eslint-disable...` directives that are unused. They are detected by VSCode, but apparently, don't cause any problem in CI:

![Cesium Unused Eslint](https://github.com/user-attachments/assets/275b0e04-dd38-4d57-a4f3-ec2777cb5e76)

One way to check for them is to change the `eslint` call to include `--report-unused-disable-directives`, as in

`"eslint": "eslint \"./**/*.*js\" \"./**/*.html\" --cache --quiet --report-unused-disable-directives",`

But preferably, this should be included somewhere in https://github.com/CesiumGS/eslint-config-cesium . 

The [documentation](https://eslint.org/docs/latest/use/configure/rules#report-unused-eslint-disable-comments) mentions this:

> To report unused eslint-disable comments, use the reportUnusedDisableDirectives setting. For example:
> 
> ```// eslint.config.js
> export default [
>     {
>         linterOptions: {
>             reportUnusedDisableDirectives: "error"
>         }
>     }
> ];
> ```

So it might be trivial to add this for someone who knows the right place.

---

Running `eslint` with `--report-unused-disable-directives` currently generates **149** messages 😬 But it looks like they can **all** be fixed automatically with `--fix`. So this could easily be done as a one-shot cleanup.

(Some of them are about to be fixed as [part of another PR](https://github.com/CesiumGS/cesium/pull/12176#discussion_r1743628122), so it could make sense for this PR to be merged before doing that cleanup pass)

Tim-Quattrochi:
@javagl I can tackle this after the PR you referenced is finished. I took out the ones in `MetadataTableProperty.js` in said PR.

javagl:
This issue mainly referred to the build process and configuration. (I.e. it was not about the eslint warnings _themself_, but about the fact that they are not emitted during the build). 

I think that some setting has to be changed in https://github.com/CesiumGS/eslint-config-cesium (but am not sure where and how). 

But of course, _after_ that is enabled, the warnings that it will cause here, in CesiumJS, will have to be fixed. (That should be a largely automatic process, but still has to be done in a dedicated PR).

jjspace:
This should have been addressed by https://github.com/CesiumGS/cesium/pull/12990

PR #12990, opened 2025-10-21
Detect and remove unused eslint disable directives

javagl:
# Description

There have been a few
`// eslint-disable-line`
scattered im the code. These caused warnings in VSCode, but not in CI.

This change pragmatically enables the check for these directives by adding
`--report-unused-disable-directives`
to the eslint call in the `package.json` (first commit), and removes all unused directives that are found by that (second commit).

## Issue number and link

Addresses https://github.com/CesiumGS/cesium/issues/12186 . One could still add this as a default configuration option to https://github.com/CesiumGS/eslint-config-cesium , but it's unlikely that this is going to happen anytime soon.

## Testing plan

Wait for CI to pass, probably... ?!

Changing the call in the package JSON to
`"eslint": "eslint \"./**/*.*js\" \"./**/*.*ts*\" \"./**/*.html\" --report-unused-disable-directives",`
(i.e. omitting the `--cache --quiet` part) should print any infos to the console.

Right now, this prints
```
C:\cesium\packages\sandcastle\src\Bucket.tsx
  200:6  warning  React Hook useEffect has a missing dependency: 'activateBucketScripts'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```
which is unrelated. (Before this change, it reported all the unused directives)

# Author checklist

- [ ] **n/a** I have updated the inline documentation, and included code examples where relevant

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "coverage": "gulp coverage",
     "build-docs": "gulp buildDocs",
     "build-docs-watch": "gulp buildDocsWatch",
-    "eslint": "eslint \"./**/*.*js\" \"./**/*.*ts*\" \"./**/*.html\" --cache --quiet --report-unused-disable-directives",
+    "eslint": "eslint \"./**/*.*js\" \"./**/*.*ts*\" \"./**/*.html\" --cache --quiet",
     "make-zip": "gulp makeZip",
     "markdownlint": "markdownlint \"**/*.md\"",
     "release": "gulp release",
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -31,6 +31,9 @@ export default [
   },
   {
     ...configCesium.configs.recommended,
+    linterOptions: {
+      reportUnusedDisableDirectives: "error",
+    },
     languageOptions: {
       sourceType: "module",
     },


packages/engine/Specs/Scene/MetadataComponentTypeSpec.js
```diff
@@ -402,9 +390,9 @@ describe("Scene/MetadataComponentType", function () {
     // Second result is max / 5
     // Third result is (max + 1) / 2
     const expectedResults = [
-      BigInt(0), // eslint-disable-line
-      BigInt(3689348814741910323), // eslint-disable-line
-      BigInt(9223372036854775808), // eslint-disable-line
+      BigInt(0),
+      BigInt("3689348814741910323"),
+      BigInt("9223372036854775808"),
```
jjspace: These were converted from number to strings. Was that intentional? I believe this is the test that's failing in CI too.

package.json
```diff
@@ -121,7 +121,7 @@
     "coverage": "gulp coverage",
     "build-docs": "gulp buildDocs",
     "build-docs-watch": "gulp buildDocsWatch",
-    "eslint": "eslint \"./**/*.*js\" \"./**/*.*ts*\" \"./**/*.html\" --cache --quiet",
+    "eslint": "eslint \"./**/*.*js\" \"./**/*.*ts*\" \"./**/*.html\" --cache --quiet --report-unused-disable-directives",
```
jjspace: Please specify this in the `eslint.config.js` instead of here. Then it should apply uniformly to all editors and ways people may run eslint. You can use the [`reportUnusedDisableDirectives`](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-objects) option.

The first object after the `ignores` at the top should look like this and it should work
```js

  {
    ...configCesium.configs.recommended,
    linterOptions: {
      reportUnusedDisableDirectives: "error",
    },
    languageOptions: {
      sourceType: "module",
    },
  },
```

jjspace (changes requested):
Well this certainly sent me down quite the annoying rabbit hole of eslint config/issues... :face_exhaling: 

Overall fully in favor of this change. I honestly thought we already had it set because VSCode was already removing them for me if they were no longer needed. The eslint extension can fix all fixable errors on save (which this is). However the `--fix` flag wasn't working in the CLI unless `--quiet` is also off (which it's not for our script). _Even if you set this to be an error not a warning_. Very annoying :facepalm:. When I went to repro this it was working fine and I don't want to spend more time digging.

packages/engine/Specs/Scene/MetadataComponentTypeSpec.js
javagl: Yeah... this was the only _semantical_ change here. There actually _was_ a "linting" issue here: Without the quotes, it says

> This number literal will lose precision at runtime.

For some of the metadata handling, the answer to what the intended behavior should be for `*INT64` is far from obvious. (Somewhat related: https://github.com/CesiumGS/3d-tiles-validator/issues/251 ). 

For this case, it's not even obvious what the test should test: The right value is what's written in the string:
3689348814741910323

The value _without_ the string becomes
3689348814741910528 
at runtime. 

So... the test is essentially making sure that the "wrong" result is computed.

However, I undid this, and added a comment. (I'm not going to change the test or metadata handling itself here...)

---

Fun fact: eslint complains about
3689348814741910323
but not about
9223372036854775808

package.json
javagl: Yeah, that's what I considered in https://github.com/CesiumGS/cesium/issues/12186 . There are three options: 

1. In the package JSON ("CLI")
2. In the _local_ configuration
3. In the global/shared configuration ( https://github.com/CesiumGS/eslint-config-cesium )

I thought that 1. could be the "least intrusive", but now changed it to use option 2. (Maybe it will become 3. in the future...)

javagl:
> Well this certainly sent me down quite the annoying rabbit hole of eslint config/issues...  

I didn't want to cause unnecessary churn. These warnings annoyed me every time I saw them (and again today), and _thought_ that it could be a nice, quick cleanup and one more issue to close ...

javagl:
@jjspace This should be ready for another review.

---

BTW: I don't know what to do with the remaining
```
C:\cesium\packages\sandcastle\src\Bucket.tsx
  200:6  warning  React Hook useEffect has a missing dependency: 'activateBucketScripts'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```
Is this something that can/should be tracked somewhere...? (Looks like it's related to the sandcastles, but I have zero context here...)

package.json
jjspace: Thanks for updating. It might be good to move to 3 at some point but that's easy from 2. 1 is just too limited and only applies in that singular path of running the npm command. Importantly it would run in CI but not during the pre-commit hook and maybe confuse people.

packages/engine/Specs/Scene/MetadataComponentTypeSpec.js
jjspace: Talking about JS precision and `no-loss-of-precision` is giving me [flashbacks](https://github.com/eslint/eslint/issues/15767) :fearful: 

I don't know enough to say exactly what the test should be testing but I'm ok leaving it alone for now. CI is passing and it passes locally for me.


> The right value is what's written in the string: 3689348814741910323

If I put this into chrome devtools and a node REPL it seems to be converted to these values
```js
test = 3689348814741910323
// --> 3689348814741910500
test = 9223372036854775808
// --> 9223372036854776000
```
I don't remember how eslint does the check exactly but it's likely always going to have "far out" edge cases

jjspace:
> I didn't want to cause unnecessary churn. These warnings annoyed me every time I saw them (and again today), and thought that it could be a nice, quick cleanup and one more issue to close ...

All good, this was just me getting distracted on niche tooling issues :sweat_smile: 

> BTW: I don't know what to do with the remaining

Yea it's about part of sandcastle. I'll try and fix it separately but it should be fine for now.

Issue #9363, opened 2021-02-02
WMS GetFeatureInfo incorrect point position


zoran995:
[...]

PR #12993, opened 2025-10-22
fix: get feature info with projection

zoran995:
# Description

When using WMS in a WebMercatorProjection instead of a GeographicProjection, the user gets the wrong feature position in response. The PR fixes the issue by passing the tiling scheme projection as an argument to the callback, which is then used to unproject the result and set the correct feature position.

## Issue number and link 

Fixes https://github.com/CesiumGS/cesium/issues/9363

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/GetFeatureInfoFormat.js
+++ b/packages/engine/Source/Scene/GetFeatureInfoFormat.js
@@ -93,8 +93,8 @@ function geoJsonToFeatureInfo(json, projection) {
       if (!defined(projection) || projection instanceof GeographicProjection) {
         featureInfo.position = Cartographic.fromDegrees(x, y);
       } else {
-        const positionInMeters = new Cartesian3(x, y, 0);
-        const cartographic = projection.unproject(positionInMeters);
+        const positionProjected = new Cartesian3(x, y, 0);
+        const cartographic = projection.unproject(positionProjected);
         featureInfo.position = cartographic;
       }
     }


packages/engine/Source/Scene/GetFeatureInfoFormat.js
```diff
@@ -86,9 +87,16 @@ function geoJsonToFeatureInfo(json) {
...
-      featureInfo.position = Cartographic.fromDegrees(longitude, latitude);
+      const x = feature.geometry.coordinates[0];
+      const y = feature.geometry.coordinates[1];
+
+      if (!defined(projection) || projection instanceof GeographicProjection) {
+        featureInfo.position = Cartographic.fromDegrees(x, y);
+      } else {
+        const mercatorPosition = new Cartesian3(x, y, 0);
+        const cartesian = projection.unproject(mercatorPosition);
+        featureInfo.position = cartesian;
```
mzschwartz5: I don't think this is particularly robust. What if we introduce a new type of projection? Or what if a user has implemented their own projection based on the `MapProjection` interface?

I also don't think it's good practice to have `featureInfo.position` be instances of different types (`Cartographic` or `Cartesian3`) depending on which projection we're dealing with. Actually, I'm a little confused because, looking at the implementations of `GeographicProjection.unproject` and `WebMercatorProjection.unproject`, their jsdocs indicate they return `Cartographic`, _not_ `Cartesian3`. I think you may have misused the `unproject` method here. 


Side note- let's use scratch variables for methods like `unproject` which, otherwise, will needlessly allocate memory for the result. (See [here](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#result-parameters-and-scratch-variables) in coding guide)

mzschwartz5 (changes requested):

packages/engine/Source/Scene/GetFeatureInfoFormat.js
zoran995: Sorry, I guess I chose the wrong variable name for the `projection.unproject` result; it actually returns the Cartographic, which is expected behaviour in all cases for `featureInfo.position`. I have updated the variable name and added a scratch variable.

mzschwartz5: That's okay, but there's still the matter of making this change robust to different types of projections.

Also, even in its current state, does the logic here work for mercator projections? 
`const mercatorPosition = new Cartesian3(x, y, 0);`

you're creating a `Cartesian3` directly from `Cartographic` components (longitude and latitude), and then unprojecting to get... another `Cartographic`? 

Are the original longitude and latitude not correct? If so, I think the fix should target whatever calculates those values originally; rather than patching the values in here by reverse-projecting.

packages/engine/Source/Scene/GetFeatureInfoFormat.js
```diff
@@ -368,7 +377,7 @@ const wmsServiceExceptionReportRegex =
   /<ServiceExceptionReport([\s\S]*)<\/ServiceExceptionReport>/im;
 const titleRegex = /<title>([\s\S]*)<\/title>/im;
 
-function textToFeatureInfo(text) {
+function textToFeatureInfo(text, projection) {
```
mzschwartz5: It looks like the `projection` arg is unused, so let's remove that.

packages/engine/Source/Scene/GetFeatureInfoFormat.js
zoran995: Here is the [sandcastle](https://sandcastle.cesium.com/#c=jVdrb9s2FP0rnD+0MuJKDdItqxMHC9Kmy5Z2QZ2tH+qgYCRaZkORAknZ84b89x2Koiwp7lAgiKKr+zjkPfeRNdVkzdmGaTIjkm3IBTO8KuK/alm0GKX1+4WSlnLJ9GI0PlnIhUwScp5lhJJP7+eEFzRneksE3UJDruGzEV07iYFrHyPuiYOjObOkKsmykqnlShKrSA7RklFbaUa4XCqy1KogdsXqeJqZUknDJrU5lRmpZKnVV5Zawq3X/cTuyXumU2oVIKVQt6TRcTF6h4VuUL1pVSJ30BZTztRvRslbdelRXQFU9BWSMfl3IQkgQSqJk8QNcBMXtIzCKWZnXpEQj2W589MHc9W5om4wB8eZdwzjjFoK60a0RwEnLpm2nJmdWke4xwLoljzH2wdasEvc5E2rHT11MP4/D2+YSTUv3f19lyPvChm9WiLV3BD8UFIqLtvrmpDKsJoHqVI645JaHE0ta1GtGXsvfEmi5kIztgRzszYoUlkwq7dj8uwZGQpjuy2RrNmMLEY3zh8IHxIXUvf57wnZ3nUutDXugDrp2xSBX8rwpwS8oBomnMqjyPmekJfj1r6Xzp31jstxS/5oGCV4eQyX29C047PWeHSKtZJHi/LrUM8Rv4/3Xe/7pdIFtZFztBi5CliMJv6FlqXgQAQoSefDvmJyck+CXo+IaZY1FQH+rHnmmhJUB9VLyznTa56yoW6TukqLabjQxWhlbWmmSZKbF8bSe1Fn8IVRonJITcxtAoGBR6aTSoLGKtkUpkFPfJ8zU3jyH6d/1o8vpSdMo2W54DKfpytWsOk32s1tRycaN4Yl1ag9W4doqWc1lQZfmLRTvFRs0hKkvn6Hpr65pJR5i+GxeeZ7Egbvn58m+m7iCbGQbaOfW9ATRbYkQqkH4CXUkvPKAJLgFAXXNPcUqDWNDbNuevirz5ixriBwr9Nw/o+gKpU5rt116jcs1ww9wQP9Kf7x9WED+uhV/PPRcfNyeBy/Dn+/Oo6Pj2rCQADuEgfUTZ0VZoEYjrJ5Cv9yXtKUvV3j+n71SlGD2qRMMmCHg7oLNT7cMa5kWdlzPw2o2cp0N6Mi5lw1rcHdUR2k15UwxjZKi6wnfGBb36eMM3SYNd3O+heIpNzw9OEj3foobdn7cnZGaWgYs94pcqHu0VVhHMHtpPvN2yYJJvnatVxkse06Q9ggf65pueJp90s3dvt91m1hQVrntfUetWADhjcsBVUFulGJMYkT1vxwbVzTDHo1iMzzIkQVdQ9qgr2ndhVbFajTRRRDMee2ytjutgT4+n22ALIzdb1QgaVC5W4RQid7ICEVfg8a6gR7V4zkwMXdqxUQBrWQ2b7a/hBhnxgCcecMiwe6X8ExJ/evXTU/9mwZpuYMeUoaN0t/2D9MQ6jxuLvdBLTPP6gWE3FRWRY/9z4fCRNA6I16G5HbVOiGcjs8TjsRUW63bj1oixF/p1QIhn3UVelKK6kqI7Zks0JNutVAcATAOTyIXSi3YawpF/UQ6E7s3RGq4h4dBdwMRlPyHElrlzzBZG5XnXmt0FrcBoxzvDzB43SoC+HBQWer6CVvt1aYz/zupK/Tmf/tCtWIOpp9wjTZ9VSLOH4djsff1B4wOHgfVMZeywGrd6b9gmz2kXoS10TA43HyrWZ9i3Usvn57efvl4vrq4nfnYjQZnRq7FewsuPuFF6XCiMKUj+I4sawohetYyX2FdNs4Ne2iepp0TU8zviY8m+35N4ekghqDL8tKiDn/hy1GZ6cJ9J+YCoWmJfM/0FaxGDi11eHZtRfGcXya4HW/pVVK3FM98Pwf) @jjhembd created a few years ago in a related ticket that showcases how it is used.

When a user adds a WMS server with a web mercator tiling scheme, it will respond in mercator coordinates, not latitude/longitude. We use those to create cartesian3 and then convert them to cartographic. 
Sample of raw WMS get feature info response 
```
[
    {
        "type": "Feature",
        "id": "Unesco_point.108",
        "geometry": {
            "type": "Point",
            "coordinates": [
                1424421.9403,
                5155468.5755
            ]
        },
        "geometry_name": "the_geom",
        "properties": {
            "cod_unesco": "IT/VA_91bis",
            "sito": "Centro storico di Roma, le proprietà extraterritoriali della Santa Sede nella città e San Paolo fuori le Mura",
            "seriale": 1,
            "cod_compon": "091-002(b)",
            "componente": "Complesso della Santa Scala",
            "tipo_area": "sito"
        },
        "bbox": [
            1424421.9403,
            5155468.5755,
            1424421.9403,
            5155468.5755
        ]
    }
]
```

mzschwartz5: I see that makes sense. So the _meaning_ of `feature.geometry.coordinates` depends on the projection used? I.e. a WMS may return web mercator coords for that field, or it may return long/lat (for a geographic projection)?

If that's the case, these changes need to either happen as part of the `MapProjection` interface, or maybe as part of `WebMapServiceImageryProvider`, so that they can be more robust to different types of projections / meanings of `feature.geometry.coordinates`

zoran995: You are proposing to extend MapProjection with a function that converts from projection coordinates to cartographic or Cartesian coordinates? I am not sure I know enough about the Cesium codebase to do that, and in that case, IMHO, it will be easier to do it in the user part of the codebase than in Cesium (and leave a long-standing bug still existing in Cesium).

The story behind this issue is that Cesium expects a GetFeatureInfo request with cartographic coordinates, even when the tiles are requested in the Mercator projection and the actual request to the WMS server specifies `EPSG:3857`. It just doesn't make sense to do it that way.
For moving this to `WebMapServiceImageryProvider`, it has already been moved as much as possible, since processing is done using GetFeatureInfoFormat, which is created in WebMapServiceImageryProvider.
`WebMapServiceImageryProvider` will create an instance of `UrlTemplateImageryProvider` as a tile provider and supply a list of GetFeatureInfoFormats to it.
https://github.com/CesiumGS/cesium/blob/ceccffefe62892eedd217373771faba5ad02ecd6/packages/engine/Source/Scene/WebMapServiceImageryProvider.js#L257-L259
When a feature is picked, the `WebMapServiceImageryProvider` delegates everything to `UrlTemplateImageryProvider`
https://github.com/CesiumGS/cesium/blob/ceccffefe62892eedd217373771faba5ad02ecd6/packages/engine/Source/Scene/WebMapServiceImageryProvider.js#L291
which then sends a new request to the server and processes the response using the GetFeatureInfoFormat that is provided by `WebMapServiceImageryProvider`.

zoran995: Actually, reading the description of `WebMercatorProjection.unproject`, it is already doing that.

https://github.com/CesiumGS/cesium/blob/ceccffefe62892eedd217373771faba5ad02ecd6/packages/engine/Source/Core/WebMercatorProjection.js#L119-L128
Other custom projections should do the same to convert to a cartographic projection.

zoran995: Maybe a better name would be `positionInMeters`, rather than `mercatorPosition`, to reduce confusion, as similar naming is already used in WebMercatorTilingScheme
https://github.com/CesiumGS/cesium/blob/ceccffefe62892eedd217373771faba5ad02ecd6/packages/engine/Source/Core/WebMercatorTilingScheme.js#L40-L69

@mzschwartz5 Let me know what you think, but IMO everything is already handled that needs to be handled in MapProjections, and WebMapServiceImageryProvider already delegates everything to GetFeatureInfoFormat and having projection info in the format callback is also beneficial, as it is easier for users to write custom if they need

mzschwartz5: I'm not familiar enough with the WMS code, even after reading your summary, to say one way or another. I'm simply making the point that - whatever the code is now or ends up being - we should not have specific conditional logic for different types of projections. It's not good OOP practice / it degrades the utility of the `MapProjection` interface.

_How_ we achieve no-conditionals is something I'm hoping you can figure out. I was thinking along the lines of "maybe we can just call `unproject` in `getFeatureInfoFormat` _all_ the time, regardless of projection type, and let the unproject function handle it". That doesn't quite work for `GeographicProjection`, though... because it seems that `feature.geometry.coordinates` means something different based on the projection type.

So do we need something in the WMS imagery provider to specify what kind of data we get back? Do we need some additional function in `MapProjection` to handle this? Is this an issue with Web Map Servers / the OGC spec being inconsistent in the data type they return, or is it an issue with Cesium (or both)? 

This is what we need to answer if we want to make this change correctly. Until we can answer these, I would rather continue to solve this in user-land for now, than add more technical debt for us to maintain. I hope that makes sense.

zoran995: The WMS server should always return the feature position in the coordinate system we requested. So if we use GeographicTilingScheme, it will return coordinates in epsg:4326; if we use WebMercatorTilingScheme, it will return coordinates in epsg:3857; if some other projection is used, it will return coordinates in the requested coordinate system. 
The only issue is in Cesium, which expects the feature info position to be in latitude/longitude. That will be true when GeographicTilingScheme is used, and in that case, we only need to convert from degrees to radians. When a GeographicTilingScheme or another tiling scheme is used, we need to unproject the coordinates to obtain the cartographic coordinates. I really don't see a way we can implement this without conditions, except by adding a method that converts some coordinates to Cartographic, and IMHO that is overengineering (for GeographicProjection it would be from degrees to radians, and for WebMercator it would be unproject). We have an issue with naming the method so everyone can understand it.

packages/engine/Source/Scene/GetFeatureInfoFormat.js
```diff
@@ -86,9 +87,17 @@ function geoJsonToFeatureInfo(json) {
...
-      const latitude = feature.geometry.coordinates[1];
-      featureInfo.position = Cartographic.fromDegrees(longitude, latitude);
+      const x = feature.geometry.coordinates[0];
+      const y = feature.geometry.coordinates[1];
+
+      if (!defined(projection) || projection instanceof GeographicProjection) {
+        featureInfo.position = Cartographic.fromDegrees(x, y);
+      } else {
+        const positionInMeters = new Cartesian3(x, y, 0);
+        const cartographic = new Cartographic();
```
mzschwartz5: This isn't quite a scratch variable - it will still get reallocated on every call to this function. A scratch variable will be allocated outside of this function (at the module level), and reused.

[Docs for reference](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#result-parameters-and-scratch-variables)

packages/engine/Source/Scene/GetFeatureInfoFormat.js
mzschwartz5: > The WMS server should always return the feature position in the coordinate system we requested

So when we request in `GeographicTilingScheme`, why _does_ the WMS return latitude / longitude directly, instead of the projected coordinates:

```JS
  const x = cartographic.longitude * semimajorAxis;
  const y = cartographic.latitude * semimajorAxis;
```
(as in `GeographicProjection.prototype.project`).

That would be consistent with web mercator, where we would then have to run `unproject` on the returned results... See my confusion? It feels like the response of the WMS is inconsistent; sometimes it's in projected coordinates, sometimes unprojected.

zoran995: By definition, [EPSG:4326](https://epsg.io/4326) is a geographic CRS that uses angular units (latitude and longitude in degrees). Therefore, it does not produce projected Cartesian coordinates like those used in Cesium's GeographicProjection, which converts geographic positions into planar meters (making Cesium projection equirectangular, equidistant cylindrical, or plate carrée as specified in the docs).
A WMS server returns coordinates according to the definition of the requested CRS, so when the CRS is EPSG:4326, the coordinates are expressed in degrees.
Cesium implementation of GeographicProjection is, at least as far as I figured out, different from standards to make geometric calculations and 3D visualisation straightforward in Cartesian space.

p.s. Hopefully I didn't say something foolish, it had been quite some time since I learned the details around projections.

mzschwartz5: Interesting. Meanwhile, the [OGC spec for WMS](https://www.ogc.org/standards/wms/) specifies that for EPSG:4326, data is internally projected before being returned:

image

Am I misinterpreting the meaning of this?

packages/engine/Source/Scene/GetFeatureInfoFormat.js
```diff
@@ -1,9 +1,13 @@
+import Cartesian3 from "../Core/Cartesian3.js";
 import Cartographic from "../Core/Cartographic.js";
 import defined from "../Core/defined.js";
 import DeveloperError from "../Core/DeveloperError.js";
+import GeographicProjection from "../Core/GeographicProjection.js";
 import RuntimeError from "../Core/RuntimeError.js";
 import ImageryLayerFeatureInfo from "./ImageryLayerFeatureInfo.js";
 
+const scratchCartographic = new Cartographic();
```
mzschwartz5: Sorry- I was wrong about needing to use a scratch variable. It's actually going to cause every `featureInfo` to have the same position. I didn't look closely enough at the usage, but we _do_ need to allocate each new position as you were originally doing. Mea culpa.

packages/engine/Source/Scene/GetFeatureInfoFormat.js
zoran995: That part of the specification relates to the GetMap request, and my understanding is that it is only for rendering, as WMS needs to generate a planar image. Therefore, it internally maps to planar coordinates for more consistent coordinate representation and easier image generation, but it doesn't mean that the server will return data in that projection. It needs to be done that way so the data that is used is in the correct projection (or reprojected on the server) and the correct image is generated.
There are a few important parts of the WMS specifications for GetFeatureInfo:
- what GetFeatureInfo should support and provide from the server side
image
- basic details on the info format in which the server should respond
image
- the structure of the get feature info response
image

The specification doesn't provide many details about the response structure. Still, it states, `The server shall return a response according to the requested INFO_FORMAT if the request is valid, or issue a
service exception otherwise`, meaning the response must be in accordance with the format specification we requested. Here we are getting the response in GeoJSON format, which, if I recall correctly, according to the GeoJSON spec, needs to respect the actual CRS definition. In the case of GeographicProjection, it is EPSG:4326, which defines that coordinates are latitude and longitude in degrees, so we get a response like that. 

p.s. When you start diving into those specification, they get pretty complex

packages/engine/Source/Scene/GetFeatureInfoFormat.js
zoran995: No worries, I also missed the point that all feature info positions will share the same reference.

Updated

packages/engine/Source/Scene/GetFeatureInfoFormat.js
mzschwartz5: I'm reaching out to colleagues for advice here, but I still feel confident about the intuition that there's a way to generalize this behavior that is future-proofed towards unanticipated projection types. (Predicated on the WMS OGS spec being well-formed / able to tell us definitively what sort of data types to expect).

kring: @mzschwartz5 Zoran's code looks correct to me. I'd only suggest renaming `positionInMeters` to `positionProjected` or similar. The else block should work with any projection we're likely to add in the future (noting that it's been 12+ years and we still just have the two!). The reason it doesn't work with GeographicProjection is because GeographicProjection is super weird and uses "meters" for its projected coordinates, where pretty much everyone (WMS included) uses longitude/latitude degrees. WebMercatorProjection also uses meters, but that's consistent with common convention.

javagl: @kring This fixes a bug that is ~12 years old and has been reported 4 years ago. But to my understanding, the conversation here was _less_ about _correctness_, but more about the style. (And there certainly is a connection between correctness and style (as the latter affects how easy it is to achieve the former)). And from my experience, one surefire way of messing things up (in the long run) is to apply a quick and easy solution to a problem where the root cause is not well understood.

I had a short look at all this. (Not sooo short, actually). There are too many question marks to list them here. However, I think that one "entry point" for the considerations here _might_ be: 

The `WebMapServiceImageryProvider` uses some [defaults during construction](https://github.com/CesiumGS/cesium/blob/5ac1b0757dd348ed16478812a0d65e02c0a9d965/packages/engine/Source/Scene/WebMapServiceImageryProvider.js#L248). These defaults involve the `GeographicTilingScheme` and the `DefaultGetFeatureInfoFormats`. The latter are using this `callback` to interpret the data. And in the default instances that are used there, the callback **assumes** the geographic projection. So these defaults are _connected_ to each other. When setting a `WebMercatorTilingScheme`, then it is **necessary** to also use different `GetFeatureInfoFormat` objects - namely, ones where the callback performs the `unproject`.

Now, how to solve this "nicely", and in a way that is less error-prone? 

This is the _hard_ part. The part that goes beyond writing the code that makes the computer do what it has to do. 

I've been pondering with ideas, but there are always the two little devils over my shoulders (yeah, no angel - just two devils), with one saying "You don't know enough about the context to reliably find a good solution", and the other one saying "The PR already fixes it - just shrug it off and let 'em hit 'merge'".

I think that one approach could _roughly_ be:

The default handling of the `WebMapServiceImageryProvider` has to be changed. When a `tilingScheme` is passed in, then the default `GetFeatureInfoFormats` objects must **match** the tiling scheme and its projection. So it cannot always use the `DefaultGetFeatureInfoFormats`. It has to _create_ `GetFeatureInfoFormat` objects - either ones with a `callback` that just does `Cartographic.fromDegrees`, or ones with a callback that does the `unproject`. That's still ugly. In some way, it's only ~"another place to do the `instanceof` check" (with the `instanceof` being the actual red flag in terms of OO design). 

On a different level:

There are these `feature.geometry.coordinates` in some response. These can be in meters or degrees or projected or not ... whatever. What is the right place for a method like 
`const cartographicInRadians = MAGIC.convertTheseCoordinatesIntoWhatWeNeed(coordinates);`
?

Should that be in the `GeographicProjection` class? Probably not. Should that be in the `TilingScheme` class? Maybe. Should it be in the `ImageryProvider` class? That sounds reasonable (but still with a few degrees of freedom).

kring: No doubt there are things to improve here! And I don't want to discourage anyone from considering those improvements. But IMO it shouldn't hold up a community PR that is fixing a real issue in a reasonable way that is very unlikely (as far as I can see) to cause any new problems now or in the future.

Full disclosure: this PR is from my old team, and it's been fixed in TerriaJS for close to a decade now by simply supplying a custom GetFeatureInfo callback, as you suggested.

My two cents on some bigger improvements... First, this GetFeatureInfoFormat thing has always been a little optimistic. Pretty much every map tile service has its own way of providing feature information, so this idea that we can supply a single "default" that works across a bunch of them? It's mostly aspirational nonsense that maybe kind of works well enough some of the time. Fundamentally, each imagery provider needs to be responsible for retrieving and interpreting feature information. I don't think there's any `MAGIC.convertTheseCoordinatesIntoWhatWeNeed(coordinates)` that makes sense here.

Second, I kind of think `GeographicProjection` is broken. I originally added it for terrain and imagery a looong time ago, and I guess meters was convenient at the time. But it really doesn't make a lot of sense. We do this `instanceof` thing all over the code base, usually for just this reason:
* https://github.com/search?q=repo%3ACesiumGS%2Fcesium+%22instanceof+GeographicProjection%22&type=code
* https://github.com/search?q=repo%3ACesiumGS%2Fcesium+%22instanceof+WebMercatorProjection%22&type=code

So if I were going to change things, and it would absolutely be a breaking change, I'd probably make GeographicProjection's projected coordinates decimal degrees instead of meters. And provide separate methods for working with projected meters, if necessary. And then the special-case logic wouldn't be needed here anymore (and probably other places, too).

This all seems way outside the scope of this PR, though.

javagl: > But IMO it shouldn't hold up a community PR that is fixing a real issue in a reasonable way that is very unlikely (as far as I can see) to cause any new problems now or in the future.

I generally agree, with the details depending on the _scope_ of possible changes or refactorings. If this PR _could_ be improved with little effort (roughly, avoiding the `instanceof` and making this part projection-agnostic, maybe by creating this "magic" function at the right place and just calling it there), then this could be worthwhile. 

If this is just a symptom of a deeper conceptual issue (and what you said sounds like this is the case), then this PR could be merged, and the necessity for improving that part could be tracked in a new issue. This could link to this discussion and summarize some approaches for improvements. 

(I know, "🎶 1500 open issues on the wall, 1500 open issues - create a PR and close one down, ... 1501 open issues on the wall". The fact that "refactor something to make things _better_"-issues are basically never addressed is an unrelated issue on a different level)

---

>  Fundamentally, each imagery provider needs to be responsible for retrieving and interpreting feature information. I don't think there's any `MAGIC.convertTheseCoordinatesIntoWhatWeNeed(coordinates)` that makes sense here.

I'm a bit confused. The first part sounds like it's _nearly_ clear that this function should be part of the `ImageryProvider` class/interface. But how this is filled with life, for the _specific_ `ImageryProvider` types, is what I referred to as the "degrees of freedom", and may be the crux here. (I still think that a key aspect is the connection between the tiling scheme and the `GetFeatureInfoFormat` thingy, but there are several layers of questions and unknowns for me)

---

> We do this instanceof thing all over the code base, usually for just this reason:

I've seen this (and many related things that may be symptoms of what you meant with "GeographicProjection is broken" - including the ripple effect of [`rectangle.width` being broken](https://sandcastle.cesium.com/index.html#c=pVHJbsIwEP2VUU5BRQ7ppVtALVTqpYsEHH0xzjSx6tjInoDo19cpShOg5dKj5y3z3lha4wkcShKm0PgAYzC4hRl6VVds3s5jbqAdvggqGdm5yJUwPk6vbthoMPybMDoPp9fpecJlgLkZ3HEjQ1irkWlbxDzqQrOtyqkEZSDHwiH6W+ARXMCp5+Mej4+1g8Z+v4CAlFamWMgSKzy8xxPawol1qeSyx4nbbL1DToOw78N+kKV9FaQ22N22C9Ol+KXm9B81p72a0TDKPO00ThrhvarW1hHUTseMJYTVWgtCn6xq+YHEpPeNKEtaSZarDah8zCP5vXNmDQll0IUsUgvvA/Jea71Qn8ijSZYE/oFM2/CxpnjboNNi11DKdPK8HzLGsiQ8T1VkrV4J13P8Ag), which did cause quite some headaches for me...).

I sometimes try to _intentionally_ take an overly idealistic/theoretical stance on certain aspects of OOP. A seemingly extreme form of that: Whenever there is some `if (someCondition) doThis(); else doThat();`, one should at least _consider_ polymorphism. It could make sense to have a `SomethingDoer` interface with a `doIt()` method and two implementations, and replace that `if`-statement with a plain `somethingDoer.doIt();`. This is overly suggestive, and of course should not necessarily and inconsiderately be _applied_, but in (surprisingly) many cases, it does make sense.

Now,... `instanceof` is a _very_ special form of such an `if`-decision, and there are _very_ few cases where it does _not_ make sense to replace that with _some_ form of polymorphism. (With a few degrees of freedom here as well - some things are, at the core, [so vastly different](https://xkcd.com/977/) that it may be hard or impossible to define a _sensible_ common interface...)

kring: > The first part sounds like it's nearly clear that this function should be part of the ImageryProvider class/interface.

It is part of the interface, via the properly polymorphic `pickFeatures` method:
https://github.com/CesiumGS/cesium/blob/1.134.1/packages/engine/Source/Scene/ImageryProvider.js#L218

Then there's this whole weird, caveat ridden, `GetFeatureInfoFormat` thing that can be used to _implement_ that method.

> [rectangle.width being broken](https://sandcastle.cesium.com/index.html#c=pVHJbsIwEP2VUU5BRQ7ppVtALVTqpYsEHH0xzjSx6tjInoDo19cpShOg5dKj5y3z3lha4wkcShKm0PgAYzC4hRl6VVds3s5jbqAdvggqGdm5yJUwPk6vbthoMPybMDoPp9fpecJlgLkZ3HEjQ1irkWlbxDzqQrOtyqkEZSDHwiH6W+ARXMCp5+Mej4+1g8Z+v4CAlFamWMgSKzy8xxPawol1qeSyx4nbbL1DToOw78N+kKV9FaQ22N22C9Ol+KXm9B81p72a0TDKPO00ThrhvarW1hHUTseMJYTVWgtCn6xq+YHEpPeNKEtaSZarDah8zCP5vXNmDQll0IUsUgvvA/Jea71Qn8ijSZYE/oFM2/CxpnjboNNi11DKdPK8HzLGsiQ8T1VkrV4J13P8Ag)

Yeah the `GeographicTilingScheme`'s "native" coordinate space is longitude/latitude degrees. Different from the `GeographicProjection`'s projected coordinate system (meters) and Cesium's usual internal units (radians).

You certainly shouldn't assume a "native rectangle" would be in radians. That's kind of the whole point of this concept of native versus non-native. But the fact that the projection and the tiling scheme aren't consistent is pretty weird.

> Now,... instanceof is a very special form of such an if-decision, and there are very few cases where it does not make sense to replace that with some form of polymorphism.

Agreed. We just have the annoying situation here where `project` / `unproject` are _supposed_ to be the polymorphic methods that make sense here. Except in the case of `GeographicProjection` we can't rely on them because the implementation is known to be broken (by using unsuitable units).

javagl: The documentation of the `Rectangle` class says that it is in radians. The implementation of things like `width` silently _assumes_ that the values are in radians (and just returns wrong results otherwise). At no point in the codebase anyone can know whether a `Rectangle` is stored in the "right" or "wrong" coordinates, or whether any function accepts or returns one of these "right" or "wrong" rectangle. 

**tl;dr**: There should be a `CartesianRectangle` (and `Rectangle` should probably be called `CartographicRectangle`). 

But this, as well as any attempts to change the `GeographicProjection` or what `pickFeatures` is doing, involves an analysis and refactorings that would have to be stretched out over several months.

kring: Good point. FWIW, in cesium-native, we have a [`GlobeRectangle`](https://cesium.com/learn/cesium-native/ref-doc/classCesiumGeospatial_1_1GlobeRectangle.html) (always radians, handles anti-meridian) and a [`Rectangle`](https://cesium.com/learn/cesium-native/ref-doc/structCesiumGeometry_1_1Rectangle.html) (unspecified units, always minX <= maxX).

lukemckinstry:
This discussion seems to have stalled and I think the time is right to close this PR due to lack of a clear path forward. I reviewed the discussion here as there was potential tie in to the contributions in https://github.com/CesiumGS/cesium/pull/13196.

This PR attempts to address a genuine plain point when using WMS getFeatureInfo, but the sentiment seems to be this is not the right level of solution as it addresses only a narrow use case, at the cost of further complicating this section of the codebase with overly specific conditional logic.

> GetFeatureInfoFormat thing has always been a little optimistic. Pretty much every map tile service has its own way of providing feature information,

This to me is the key point, from looking at a bunch of WMS providers for updating out documentation and examples, there is not consistency in how services provide this information.

Given this, I think it makes sense to keep our current state where the users can supply a custom GetFeatureInfoFormat callback. But we should better document this workflow. 

I created this new issue for improving the docs https://github.com/CesiumGS/cesium/issues/13380

We can re-open this issue if there is renewed interest in the proposed code change.

kring:
> This to me is the key point

@lukemckinstry if that was the key point, and you've decided to close the issue, I think my point was not clear. WMS _does_ have a standard way of getting feature information. `GetFeatureInfoFormat` is meant to work across protocols (WMS plus others), and that is the part that is optimistic.

IMO this PR is fine and should be merged. In fact that's what I said the paragraph before the part you quoted. All the discussion between Marco and I was about "bigger issues" outside the scope of this PR.

Issue #8683, opened 2020-03-16
Globe incorrectly occludes things behind it when using smaller ellipsoid


OmarShehata:
When using a smaller Ellipsoid, the globe incorrectly occludes entities behind it. [Sandcastle example](https://sandcastle.cesium.com/index.html#c=jVPfb5swEP5XLF4CUmTSdVrWhkaLsikvnbYpWp94ueALeDU2sg1ROu1/r4GQQMqmISTju+9+fHcfFWiigfHSkAfy7v3dfDabLWIZhmfjh9v5x5vbuTPGsnLoiuMBtXNIPJA1Gl7m9Kmx+ZOkua6VtMAl6smU/I4lcY/JVCnYSvIcLJJ7sgdhcHryPR9XNlemyFDj0JVD8V2rX5hYruR9v+IGVaqhyHhyAfg9/xcheGEUZ35LZDo4giCWf4KG0hYkS8BYgVSjQeuI7UvZ5gu69lvOlIGFrSq1o0mBMf9Ua/2Si89nFxUKmD+hNHTvFvJCYO0MDa8/aeLAk6Cu3SR2A8Wu1eFM3zAgw7Ntv0uSCrXDYYJNbfLP2S/4BktRwk7gI08zy2Va064Hv+hD3NYOG61KyS4L6gEv1V3VsYVc9TO2spH2TsM2CUqkHa/mXPwNQHdgcK2EqnXZbaW+0tWPn6vxzHuVnmbArsgPYLlSshnEv0Cm/A9MBkwdvkIxUtWpceFNvcjYo8BlG/zJ6UVpS0otfKcki04+7ucx4a5MntHSxDQSiMIuKGK8Ipw9xN7Vbxh7JBFgjPPsSyG2/AVjbxmFDj8Iq4XrpPCtQi3gWEOym+Vja6SURqG7vo2ySokd6F7GVw). 

It almost seems like there's a rectangle around the globe that occludes entities behind it:

![occlusion](https://user-images.githubusercontent.com/1711126/76790120-9e7c7a00-6794-11ea-91fe-598a5ec83167.gif)

The effect disappears if you set the radius up to Earth-size (comment out the second line in the Sandcastle). 

This came up in [this forum thread](https://groups.google.com/forum/#!topic/cesium-dev/pDezVst5b_I) where the use case was visualizing different planets.

OmarShehata:
[...]

PR #12994, opened 2025-10-22
Depth test billboards against ellipsoid horizon

### LINKED ISSUE #13159: A fatal bug about depth test！
### What happened?

After 1.136，I find a bug about billboard or text.

When you move the Earth, billboards are not hidden by the Earth.
Especially when viewed from a downward angle, you will notice that the billboard is still visible on the opposite side of the Earth.

Image

already show ，but should be hidden！！

### Reproduction steps

1. open https://sandcastle.cesium.com/?id=billboards
2. rotate the earth
3.Especially in certain scenarios, it can significantly affect the display effect
...

### Sandcastle example

https://sandcastle.cesium.com/?id=billboards

### Environment

Browser: chrome 1.132
CesiumJS Version:1.136
Operating System:win 11

ggetz:
# Description

This is a follow-up to https://github.com/CesiumGS/cesium/pull/12821#discussion_r2449428957 which applies some basic horizon culling (see Kevin's [horizon culling](https://cesium.com/blog/2013/04/25/horizon-culling/) blog post for more info) to the fallback depth test for billboards and labels.

It seems to be working pretty well for mid-level horizon views as well as at the global scale:

| main | this branch |
| --- | --- |
| Image | Image |
| Image | Image |

And it seems to be working well for 3D Tiles as well, so I went ahead and removed the custom `depthTestDistance` callbacks in the Moon and Mars example.

## Issue number and link

Fixes #13159

## Testing plan

1. Test billboard and label properties, particularly the following, with a) the default ellipsoid, b) terrain, and c) Google P3DT (3D Tiles):

- `pixelOffset`
- `eyeOffset`
- `horizontalOrigin`
- `verticalOrigin`
- `scale`
- `showBackground /backgroundPadding`

2. Verify depth testing behavior of the `Moon` and `Mars` sandcastle examplws.

# Author checklist

- [ ] ~I have updated `CHANGES.md` with a short summary of my change~ I think this falls under the update for https://github.com/CesiumGS/cesium/pull/12821

### DIFF
--- a/packages/engine/Source/Scene/BillboardCollection.js
+++ b/packages/engine/Source/Scene/BillboardCollection.js
@@ -319,10 +319,10 @@ function BillboardCollection(options) {
 
   this._highlightColor = Color.clone(Color.WHITE); // Only used by Vector3DTilePoints
   this._coarseDepthTestDistance =
-    options.coarseDepthTestDistance ?? Ellipsoid.default.minimumRadius;
+    options.coarseDepthTestDistance ?? Ellipsoid.default.minimumRadius / 10.0;
   this._threePointDepthTestDistance =
     options.threePointDepthTestDistance ??
-    Ellipsoid.default.minimumRadius / 1000;
+    Ellipsoid.default.minimumRadius / 1000.0;
 
   this._uniforms = {
     u_atlas: () => {


mzschwartz5:
On the whole this works nicely, but I'm noticing some odd behavior on Mars and Moon. Sometimes the labels and points flash in and out when they should be visible:

https://github.com/user-attachments/assets/b844ea77-f456-48e2-82e5-15fa0f748921

https://github.com/user-attachments/assets/c0015a1d-26d8-4ac8-8e08-60c76664d816

I haven't noticed that issue on any WGS84 sandcastles, so I thought maybe it was related to some hardcoded uniforms. I stumbled across [this issue](https://github.com/CesiumGS/cesium/issues/8683), but it's already been solved. So I'm not sure what's happening here, but I still think this is an improvement, and if it only slightly affects other ellipsoids, that's not a terrible tradeoff.

mzschwartz5:
@ggetz This is ready for review again. I made three changes:
1. Merged in main - this brought in some other fixes I've done to depth testing, namely #13018, which involved some merge conflict resolution
2. Changed some of the new logic introduced in this PR, namely:

```GLSL
    ...
    float testDistance = useGlobeDepth ? globeDepth : distanceToEllipsoid;
    float testDelta = eyeDepth - testDistance;

    float depthsilon = useGlobeDepth ? u_threePointDepthTestDistance : u_coarseDepthTestDistance;
    if (testDelta < -depthsilon) {
        discard;
    }
```

I believe that `depthsilon` line is incorrect. The distance to test against isn't those uniforms. Those uniforms are the thresholds that tell us _when_ to use different test distance metrics (three-point test vs. globe depth vs. depth plane).

3. Adjusted the `u_coarseDepthTestDistance` value to 10x it's previous value. This should probably be user-adjustable, as you've previously suggested, but for now this at least addresses the egregious cases. The current value was producing some really bad results in some cases (see this [issue/comment](https://github.com/CesiumGS/cesium/issues/13150#issuecomment-3800771382)). Basically, when `u_coarseDepthTestDistance` is too small, the depth-plane test too lenient compared to a true globe-depth test (especially with terrain on), so you get billboard fragments popping through terrain all over. 

Oh! Also, in the process, I seem to have addressed the issues I commented on about the Mars and Moon sandcastles! I suspect those might have been due to the 2nd point above.

ggetz:
Thanks @mzschwartz5!

I didn't have time to give this a full review before today's release. But I would like to get this reviewed an merged ASAP after the release.

ggetz:
> Adjusted the u_coarseDepthTestDistance value to 10x it's previous value. This should probably be user-adjustable, as you've previously suggested, but for now this at least addresses the egregious cases. The current value was producing some really bad results in some cases (see this https://github.com/CesiumGS/cesium/issues/13150#issuecomment-3800771382). Basically, when u_coarseDepthTestDistance is too small, the depth-plane test too lenient compared to a true globe-depth test (especially with terrain on), so you get billboard fragments popping through terrain all over.

I'm very much leaning towards exposing these values to the user. In case there are any disagreements with the new behavior, it'll let users tweak as a workaround and may prevent us from having to rush out another adjustment.

packages/engine/Source/Shaders/BillboardCollectionVS.glsl
```diff
@@ -322,6 +322,18 @@ void main()
...
+    vec3 rayDirection = normalize(positionEC.xyz);
+    czm_ray ray = czm_ray(vec3(0.0), rayDirection);
+    vec3 ellipsoid_inverseRadii = czm_ellipsoidInverseRadii;
+    czm_raySegment intersection = czm_rayEllipsoidIntersectionInterval(ray, ellipsoidCenter, ellipsoid_inverseRadii);
+
+    if (!czm_isEmpty(intersection))
+    {
+        v_splitDirectionAndEllipsoidDepthEC.y = intersection.start;
+    }
+
```
mzschwartz5, line 322: @ggetz Question: we're calculating this distance per vertex here. When we use the result in the fragment shader, it's been interpolated ... does that make this an approximation?

I was just playing around and noticed a little flickering when the camera is right head-on a billboard. And I wonder if that may explain why.

edit- actually, if you zoom in then zoom back out, the flickering stops... so it's some transient issue. I don't think it's _directly_ related to this logic, just exposed by it. I have a strong hunch it's related to #13082

ggetz: I _think_, yes, technically this will interpolate from the vertex to fragment shader. Perhaps worth specifying the `flat` keyword? I assume we shouldn't be interpolating the split direction either.

But I agree, I don't think it should be interfering with the precision at the scale we're looking at.

ggetz: Taking a testing pass now, so I can add the keyword if you agree @mzschwartz5.

mzschwartz5: Keyword should be fine for clarity, but yeah I'm pretty sure this would be fixed by solving #13082. I looked into it last week for a couple hours but didn't get far, unfortunately.

ggetz:
In horizon views, I'm often seeing billboards disappear unexpectedly:
image
image

I think this is due to the limits of the approximation. I'm adding a small epsilon check to account for this.

mzschwartz5:
> In horizon views, I'm often seeing billboards disappear unexpectedly: <img alt="image" width="577" height="363" 

I'm having trouble reproducing this. I _can_ reproduce it, sometimes, but not consistently. It makes me wonder if it's actually an issue with this ellipsoid distance approximation.

ggetz:
> I'm having trouble reproducing this. I can reproduce it, sometimes, but not consistently. It makes me wonder if it's actually an issue with this ellipsoid distance approximation.

I've just push up my changes so you can verify, but accounting for cases where the eyeDepth and approximate ellipsoid depth are very close seems to resolve the issue.

I'm still finishing out my testing for this branch, so probably hold off on deeper review for now.

mzschwartz5:
Wait a second.... those points on the moon, are points! Not billboards or labels. So they're handled by the `PointPrimitiveCollection(FS|VS)` shaders. So this issue is completely unrelated to the ellipsoid distance checks. I think we can ignore it for the sake of this PR.

ggetz:
> Wait a second.... those points on the moon, are points! Not billboards or labels. So they're handled by the PointPrimitiveCollection(FS|VS) shaders. So this issue is completely unrelated to the ellipsoid distance checks. I think we can ignore it for the sake of this PR.

Points become billboards under-the-hood when clamped to ground. 🥲

ggetz:
> Points become billboards under-the-hood when clamped to ground.

As additional thought—Since `PointGraphics` do not allow specifying a `verticalOrigin` like `BillboardGraphics` or `LabelGraphics` do, it might be appropriate to put all point billboards into a separate collection and to always apply the three-point check to ensure the point does not clip when depth testing.

IMO, out of scope for this PR.

mzschwartz5:
> It might be appropriate to put all point billboards into a separate collection and to always apply the three-point check to ensure the point does not clip when depth testing

That's a good point (no pun intended). At the very least, this should be possible for users to do manually after this PR - by setting the `threePointDepthTestDistance` to 0 on the collection.

ggetz:
Feeling more confident above the logic I just pushed up, though I do need to run a full "test suite" across different cases.

My last set of changes made it so that:
- The "nudge" to make sure the regular depth test passes only happens when the billboard is in front of the globe. If it is on the edge or just behind, we just pass everything along as usual, preserving the horizon depth checks.
-`coarseDepthTestDistance` defaults to the minimum radius since you can only ever see approximately half the globe at most. Therefore, we have a reasonable view when there is a lot of variable terrain in the view.

Between the two of these changes, this should keep the horizon depth tests reasonable, while still preserving the behavior needed in cases like https://github.com/CesiumGS/cesium/issues/13150.

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -157,27 +168,56 @@ void doDepthTest(float globeDepth) {
...
+    // If the camera is close, compare against the globe depth texture that includes depth from the 3D tile pass.
+    if (useGlobeDepth && eyeDepth < globeDepth) {
         discard;
     }
 }
 
+void writeDepth(float eyeDepth, float globeDepth, float distanceToEllipsoid) {
+    // If we've made it here, the manual depth test above determined that this fragment should be visible.
+    // But the automatic depth test must still run in order to write the result to the depth buffer, and its results may
+    // disagree with our manual depth test's results. To prefer our manual results when in front of the globe, apply a small offset towards the camera.
```
mzschwartz5: This is kind of a nit, but the offset towards the camera is not always small. 

Imagine this: you're viewing the earth from a distance so you can see the full globe. A billboard is on the side of the globe - it's facing towards you (by definition), so roughly half of the billboard is embedded into the globe. But, since the billboard is still on the near side of the globe, it passes the coarse depth test and is visible. To make it visible, we have to offset some of those fragments embedded in the globe by up to a _whole ellipsoid radius_ towards the camera.

mzschwartz5: I wouldn't hold up merging on this, but just wanted to document it at least

mzschwartz5:
> coarseDepthTestDistance defaults to the minimum radius since you can only ever see approximately half the globe at most

I feel like this might be a little _too_ aggressive in many use cases. Below is an image from this [sandcastle](https://ci-builds.cesium.com/cesium/billboard-horizon-patch/Apps/Sandcastle2/index.html#c=rVNRT9swEP4rVl4IUnAWQHsIoVrpOoEUyFQyeImEHOeSWnPsynYLZeK/z3ESaMvTpL1Evu+7z3ffnUOl0AZtGDyDQpdIwDOagWbrFj84zD+iLpxJYQgToI4C9KcQCBlQygLxmJ73Ma6VbB+l4tUA+MdBId6OLwpRCOqK1YQyzsz2l+K2YuFhHLKWNKDDkcENqwvvYhRoCgJsat8lduE7SUkLiljWwbgPd6qBMPbGD7WLGWhMqsp3TlZSW0R+WJkRZeyJiDPn5js0CkD7J1F0iqPzAJ19xVGAvnTGECoZ56Ukqor7uSDkzMS7NoOeWAJrlmYBNSgQFN7rXe/jeJZOb38+5dlTPl8spjd3g3wDyjBKeKZYszP4hz0YX2V5nt06yZv7clICj4feDLyY2M78auw67djTwhtq1HbNHR+dr+zcidAnGlS3jP9jYSkVe+1e0qGJ6wMCp/Mf+b/5nt7P05u7ee98eHJe4CXabDlMOvQba1dSGbRW3LePzkC74sRuOizX9DcYTLXuREk4SpKKbRCrLgvv4CcoPEQ50doy9Zrze/YKhTdJQpu/J+OSVEw0mbXAybZLWUaTtAcxxklow88qIyUvidq58S8). It's just a billboard clamped to terrain - in the Grand Canyon (notable because the terrain is pretty noisy there).

At a moderate distance at certain angles, the billboard gets clipped by the terrain pretty badly - because it's testing against the globe depth buffer, not against the ellipsoid. Of course, a user can adjust these settings now; but is the default too lenient?

image

mzschwartz5:
Also - there seems to be an issue with the mars sandcastle parsing its geojson... but I just checked on main and it's happening there too. Seems to have been caused by something else recently - only present in 1.138.

ggetz:
Thanks for catching that @mzschwartz5! You make a good case, I've once again decreased the depth test distance. 

Your example at a few different view points: top down, semi-obstructed by terrain, and by the horizon.

image

image

image

mzschwartz5 (approved):
I think this is ready to go! :tada:

PR #13010, opened 2025-11-03
Remove support for the legacy splats extension

jjspace:
# Description

This is simply a deprecation PR for the removal started in https://github.com/CesiumGS/cesium/pull/12843/

I'm not completely familiar with the splats code but I believe I got all the places it needed changed. I left the warning in for now just in case but functionally it's not considered. I think we do this a few other places for data/tileset related things. Easy to remove if we want to or it's not helpful

I also added a note in our release guide to check these issues the week before because I didn't realize this needed to happen.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12842

## Testing plan

- I believe this can be tested with the splats sandcastles to make sure they still work.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/GaussianSplat3DTileContent.js
+++ b/packages/engine/Source/Scene/GaussianSplat3DTileContent.js
@@ -137,7 +137,7 @@ GaussianSplat3DTileContent.tilesetRequiresGaussianSplattingExt = function (
     ) {
       deprecationWarning(
         "KHR_spz_gaussian_splats_compression",
-        "Support for the original KHR_spz_gaussian_splats_compression extension has been deprecated in favor " +
+        "Support for the original KHR_spz_gaussian_splats_compression extension has been removed in favor " +
           "of the up to date KHR_gaussian_splatting and KHR_gaussian_splatting_compression_spz_2 extensions" +
           "\n\nPlease retile your tileset with the KHR_gaussian_splatting and " +
           "KHR_gaussian_splatting_compression_spz_2 extensions.",


packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -125,30 +125,27 @@ GaussianSplat3DTileContent.tilesetRequiresGaussianSplattingExt = function (
...
-        "of the up to date KHR_gaussian_splatting and KHR_gaussian_splatting_compression_spz_2 extensions and will be " +
-        "removed in CesiumJS 1.135.\n\nPlease retile your tileset with the KHR_gaussian_splatting and " +
-        "KHR_gaussian_splatting_compression_spz_2 extensions.",
-    );
+    if (
+      tileset.isGltfExtensionRequired("KHR_spz_gaussian_splats_compression")
+    ) {
+      deprecationWarning(
+        "KHR_spz_gaussian_splats_compression",
+        "Support for the original KHR_spz_gaussian_splats_compression extension has been deprecated in favor " +
```
jjhembd: I think the message should be "removed" rather than "deprecated".
`Support for the original KHR_spz_gaussian_splats_compression extension has been removed...`

Also should this be an error rather than a warning?

jjhembd:
One more comment, I think we need a "breaking change" note in `CHANGES.md`

jjspace:
> One more comment, I think we need a "breaking change" note in CHANGES.md

The CHANGES was already updated with the removal

> I have one question about the warning message when people try to use the removed extension. I assume the whole tileset will be broken? In that case should we throw an error rather than printing a warning?

I'm not entirely sure on that front. Maybe @weegeekps has a better idea?

weegeekps:
> I have one question about the warning message when people try to use the removed extension. I assume the whole tileset will be broken? In that case should we throw an error rather than printing a warning?

I don't think we need to overthink this. The message just needs to explain that the old version is deprecated and removed, and they need to retile using the latest version.

As a word of warning: it will not impact our implementation in CesiumJS, but the current base specification may have some small changes coming to the tiler in the near future.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
jjspace: @jjhembd Re throwing vs the warning I think this is ok for now. It looks like it gets caught further down the chain that it's an unsupported extension. Functionally it renders nothing in the viewer. I think throwing an error could cause it to fail/block rendering completely which I think is not desired. Happy if we want to reconsider in the future but I'm comfortable with this solution for now

jjhembd:
OK @jjspace sounds good. Thanks @weegeekps for your input. Merging now

Issue #13012, opened 2025-11-04
Label and Billboard depth test regression (1.135)


ppoulainpro:
### What happened?

Hello;
Release 1.135 has changed the billboard depth test logic from automatic to manual (Switch from automatic to manual depth testing for billboards #12821).
Unfortunately, it introduces some regression when using label or billboard within entities.

Best regards.
Pascal

Cesium 1.135:

<img width="462" height="306" alt="Image" src="https://github.com/user-attachments/assets/6398204b-eede-4b75-a372-c6b425e81e5b" />

Cesium 1.134

<img width="703" height="460" alt="Image" src="https://github.com/user-attachments/assets/0ba0fb8b-dfa2-4d35-a845-4e6de9bd4957" />

### Reproduction steps

See sandcastle in attachment.

### Sandcastle example

[...]

Issue #4108, opened 2016-07-12
Z-Ordering of entity collection


tfili:
It would be nice to be able to change the render order of entities. Right now we have to tweak the `eyeOffset` which needs to change based on camera distance.

hpinkos:
This is probably one of the most frequently requested things on the forum.
Also requested in #53

pjcozzi:
CC #3647

shunter:
cc AnalyticalGraphicsInc/czml-writer#20

hpinkos:
Related: #4878 #4695

hpinkos:
Also reported here: https://groups.google.com/forum/?hl=en#!topic/cesium-dev/iyNHsibp2eI

smills2929:
+1. I'd love to see this!

hpinkos:
Also requested here: https://groups.google.com/forum/?hl=en#!topic/cesium-dev/p_TufXEFwTA

emackey:
Thinking about a possible implementation for billboards:  https://groups.google.com/d/msg/cesium-dev/7CHDGRyJvwc/S55fUJ8XCAAJ

schitzN:
*push*

Seems like this is still an issue. We have the same problem with labels drawn underneath polylines.

hpinkos:
Hi @schitzN, this hasn't been addressed yet.  But stay tuned!  We're starting work on this in the next month

ggetz:
@hpinkos Will these changes also apply to billboard/lablel/polyline collections?

See on the forum: https://groups.google.com/forum/#!topic/cesium-dev/ywyFR5GSi_c

hpinkos:
@ggetz the changes I'm working on this round are specifically to address z-ordering in 3D, so I won't be making changes to billboards/labels because they are ordered spatially by their 3D position.  We will have support for ordering polylines on the earth surface.

erezy:
@ggetz Thanks!
@hpinkos changing the order of the primitives collections in 2D worked in the past (V 1.21)
I don't know why it stopped working.
Do you know when the issue will be resolved?

hpinkos:
Thanks @erezy, I didn't realize this was a regression.  I've opened a separate issue so we can look into fixing it: #6569

eatapl:
Will these changes also apply to rendering GeoJsonDataSource ?

hpinkos:
[...]

PR #13018, opened 2025-11-05
Fixes model-billboard depth interactions

mzschwartz5:
# Description

In #12821, I turned off automatic depth testing for billboards / labels, and implemented it manually in the shaders. This was to give us more control over how these entities clip against the globe at various distance scales. ([We've done a similar thing in the past](https://cesium.com/blog/2018/07/30/billboards-on-terrain-improvements/), this was supposed to improve / build on those techniques)

That PR introduced a regression: the outlines or transparent parts of billboards and labels will now clip through models no matter what. This is because:

1. The aforementioned manual depth test *reads* from the globe depth texture
2. The opaque render pass does not *write* to the globe depth texture, so parts of billboards rendered in the subsequent translucency pass do not adhere to the depth values for things (like models) written in the opaque pass.

Moreover, in digging into this bug, I realized that there's another bug: opaque billboards / labels also do not respect the depth of *other* opaque billboards / labels, since the opaque billboard pass only _reads_ from the globe depth, but doesn't write back to it (and you *can't* read-from and write-to the same depth texture in webgl2, in the same pass).

## Solution

1. Re-enable automatic depth testing for billboards and labels. 
2. Keep the manual depth testing that was added. This will still be the primary control over how billboards and labels clip or do not clip against the globe.
3. Fudge the depth value written out of the fragment shader so that any fragments that passed the manual depth test will pass the automatic depth test (at least, compared to the globe depth. Fragments can still fail if there are _other_ occluders, like models, in front).

This gets us the best of both worlds - more depth testing control but also we still utilize the standard depth testing pipeline and thus get proper clipping against models and other billboards, without poor clipping against the globe.

## Issue number and link

#13012

## Testing plan

Same testing plan as in #12821, but with an additional regression test per the linked issue:

[Test this sandcastle](https://sandcastle.cesium.com/#c=tVf7b9s2EP5XOP8kD678SJysXpItfnQNatdprLQb5iGgJcoiQpMGSeWxIv/7jhT1cgIsKZLAQMTj8bvvjnenE91shdToZ4QVGhFF0w2KpdigZSO0q2Xj1yWnmdIC8yjESjOS65QSq7fkoeBKoxtKbolEx4iTW4fqf7UyL8cdCa4x5UQuGy30fckRojwWQ3E3QDFmirSMSBFGQk0FP+MRDbEWsrab4EjcqgHSMs0lImXRKfDFmhTyh2aF2pbyYUpZtEvvvJB7mXqccmsahZIA2kxEhHmpZC2UELpOdDNjnbnqE66ppkT5kmzEDTllzMEg5OwKRS3ccW5xhKWGJ8z3fBPNMVlLQpRnjiD0rtvb8zuH+/sH3fetTLS/73f6nb3DzoETZDzswtjKLSUER5SvS0MzrBNfiwsQY6687l6/qr6lOkxAuVORScHYjijZ7gTsY2bm3By/AH3P2W1liC0LUjUkJIUo4XoQAgmUYiE3yk92AL+kEHbJQd3FJA9h7v5WFr6XVuxF3IOB3YvBUeR9z05yvIHsMHdZBx7smqhQHlQXbntjcmKAHCoCRFqFBQUKqZhuzukdYQv6Lxjt9n4pd/Gd2V2EmMFOrwN/bu/B/Wd4VTOgyZ0eLBvBZBFA2eRSyPrbIQ6v11KkPBqUBWJ9M7bncayIHlRur0i+ntdvvTvoN0s0fQ90nNrUEFgYif/hbDq9Ov08vppfBtOzz5PiQEwZGwkGtZljm4X/12Q6nX8rtESqGZT7NxrpZNDt7MqfALiYjOvRWIGhlcAyqkQEKn0NwSur2tZSAHGCTvOb6S01zOH0dPSphQ57TSiIMdb48mLqlb7fEKmhz7C5pGsKV563rprYH86DYD6rXdVDlugu57SE2yDRJE/FLCdB46HsQ2JrEknB7t/mpHPIXjD01VMqQ4ljXdwyqNpeCP0v70teswxDtUctG77fht8Cb7aMGCfbNlFVO3MHsN3TFTz6a7YyYepD9vmdzI3Cq9YT1MZScPIWvCxw9Tmn1u0/j9kfNv/RV5LQkL0yxQzbQddXOc3nUPwoNIKgoyFmTAj+QxydDKH/C6gzUl85tiVK1158IXiGEzPKrlEg0/D6bfkbO9bM7vqRDy+iv7imnJMIjRIMdartBPLqyTzDedyv4PElKXLJGdUIBqHXpQWAFrl4AE46tuX13LrHoUAjsdlKohTEz1p8qwyw1kpjT2SAJf+DKfAp+LNXdeWF5Uhj5P3k3g0fwK9UkjHR2aTqq3RrhmU1xIoqz70RVEg4aVYgELqlHKZXH979UldCYoISJFShlYTZFqbUSBCFOPQNB4wsMLrkFN5WCjOYewo/jH9ARtUik8cDYvEq/cQE77HkFfrKGbwXMQ/NfbxB9hfotUVZBf3HDP+xk2X5jWNmyEAItsJyRnjquXe4OdhoNY7s5HRiDv/uPphgFvSAjybAB1iq9gpyl2g/VPbQUTs/chTRG0Sj4ye+jlDIsFKwE6fMDpHLxslRG/Rrx5iww/MccoLhe6OSdE+mmdD3/aM2LB+f0pkzFcT/AA), and confirm that, on this branch, it looks the way it does in the Cesium 1.134 picture of the linked issue.

# Author checklist


packages/engine/Source/Scene/BillboardCollection.js
```diff
@@ -2059,7 +2061,10 @@ BillboardCollection.prototype.update = function (frameState) {
     ) {
       this._rsTranslucent = RenderState.fromCache({
         depthTest: {
-          enabled: false,
+          enabled: true,
+          func: useTranslucentDepthMask
+            ? WebGLConstants.LEQUAL
+            : WebGLConstants.LESS,
```
mzschwartz5, line 2067: Renabled as it was before #12821 (yes, that is `LEQUAL`)

ggetz:
@jjhembd Do you have the bandwidth to review this?

donmccurdy:
Tested on sandcastles from two recent forum questions, and both appear to be fixed by this PR:

- https://community.cesium.com/t/labels-flickering-with-pixels/44231
- https://community.cesium.com/t/billboard-is-drawn-under-polyline-in-the-same-entity/44242

packages/engine/Source/Scene/BillboardCollection.js
donmccurdy: Does the comment above this at line 2052 still look accurate? The part about "only the opaque pass gets the benefit of the depth buffer" doesn't fully make sense to me if neither pass is _writing_ to the depth buffer, but I might have misunderstood!

donmccurdy:
Looks good to me!

A couple questions just for my own learning —

1. Are Google Photo Realistic Tiles considered "models" in Cesium terminology?
2. Do you recall why depth writes are disabled for the opaque render pass? Or if there is a thread somewhere that I could read more about that? 

^neither question needs to block a merge, happy to merge this any time.

EDIT: Sorry, hit the 'close' button by accident – reopened.

mzschwartz5:
1. I'm not sure! If I figure it out I'll let you know, but I just tried to find out and was quickly inundated. But it looks like there aren't that many types of `*3DTileContent` classes, so Model is a decent bet.
2. Sorry for the confusion - depth writes are enabled for the opaque render pass, in general. What I meant, was that my last PR disabled depth writes for billboard specifically, for their opaque pass (and translucent pass). And that ended up being a bad idea!

packages/engine/Source/Scene/BillboardCollection.js
mzschwartz5: I agree that this sort of makes no sense.... but I also am afraid to remove or change it in case it's important lol 😆

javagl:
Whether Google Photorealistic 3D Tiles are considered "models"? I'm going out on a limb and say "Yes". There is a dedicated API to load glTF data as a `Model`. But under the hood, every tile content that is backed by glTF is a `Model3DTileContent` where the `Model` is the workhorse for actually rendering it. 
(Of course, there are a bazillion differences and details. On the "rendering layer", there are not too many differences between rendering a standalone `Model` or a `Model` that is part of a tile content. That may be the main point here...)

mzschwartz5:
Yeah I don't know nearly as much as @javagl here but knowing _why_ you want to know if GP3D tiles are treated as "models" might help us give a better answer.

In any case, should we merge this then?

donmccurdy:
Nothing blocking or urgent. :) Just hoping to understand what types of content are on the "model" side of model-billboard interactions. The concept of a model is very informal in some software I'm used to (roughly: "an external file with some 3D stuff in it") but it seemed like it must mean something more technically specific here, e.g. for depth buffer interactions. The pointer to Model.js and Model3DTileContent.js is helpful, thanks!

packages/engine/Source/Shaders/BillboardCollectionFS.glsl
```diff
@@ -243,6 +244,18 @@ void main()
...
+    float depthArg = v_depthFromNearPlusOne;
+
+    if (globeDepth != 0.0) { // On the globe
+        float globeDepthFromNearPlusOne = (-globeDepth - czm_currentFrustum.x) + 1.0;
+        float nudge = max(globeDepthFromNearPlusOne * 5e-6, czm_epsilon7);
+        float globeOnTop = max(1.0, globeDepthFromNearPlusOne - nudge);
+        depthArg = min(depthArg, globeOnTop);
+    }
+
+    czm_writeLogDepth(depthArg);
```
mzschwartz5, line 259: In hindsight, there's an issue with this approach: imagine two billboards that are roughly in the same place, but one is closer to the camera. Imagine both billboards are partially occluded by the terrain, but still deemed to be "visible", and thus occluded fragments get nudged in front of the terrain.

Fragments from both billboards will end up at the same depth - right in front of the terrain. This nudge does _not_ respect relative ordering of billboards.

I've been playing around with a similar solution: instead of nudging to right in front of the globe, nudge to some band between the globe and the near plane. Where exactly a fragment gets nudged to depends on its depth, so that ordering is preserved. It looks like this:

```glsl
    if (globeDepth != 0.0) { // On the globe
        float globeDepthFromNearPlusOne = (-globeDepth - czm_currentFrustum.x) + 1.0;
        float globeDepthFromNear = globeDepthFromNearPlusOne - 1.0;
        float behind = max(depthArg - globeDepthFromNearPlusOne, 0.0);
        float t = behind / (behind + globeDepthFromNear);
        float nudgedDepth = 1.0 + t * globeDepthFromNear;

        depthArg = czm_branchFreeTernary(behind > 0.0, nudgedDepth, depthArg);
    }
```

There are some issues with it:
1. It doesn't prevent z-fighting of billboards that are close together. That's a separate issue though (#4108).
2. It could potentially cause a billboard to end up on top of some other entity. Ideally, to solve that, billboards would be in their own render pass, after other regular entities, and then the `globeDepth` texture would include anything we care about occluding.

For those reasons, I haven't opened a PR with these changes, but I did want to document them somewhere, just in case. It sort of feels like I'm playing whack-a-mole. Every attempt to patch a depth testing issue creates a new issue. Given the time, billboards could really benefit from a complete redesign.

Issue #5127, opened 2017-03-20
Add asynchronous hint to entities


shunter:
This issue has been discussed many times, but I couldn't find an issue for it, so filing this for reference.

In many cases, using the entity layer to create geometry produces bad results, because all non-time varying geometry is created asynchronously, causing significant latency when updating due to the enormous overhead of the web workers.

Our current recommendation is to use `CallbackProperty` to work around this bug, (see related #4791) but this workaround has significant performance penalty - the callback is run every frame, and the updaters are built expecting the values to change every frame.  For truly time-varying properties, this is desirable, but in many cases, the value doesn't vary with simulation time, but does change at unknown or unpredictable times - through user input perhaps, or via updates from a server.  In these cases, the `definitionChanged` event approach used by `ConstantProperty` is the right approach for efficiency.

The real solution would be to provide a hint to declare that a particular graphics should always be created synchronously.  Then the natural and obvious approach of simply assigning new values would work as expected.

EKoetsjarjan:
Hi I'm new at cesium. 

I'm using the ConstantProperty  to update the value is set value is used.
how to do it using definitionChanged?

My updating works like this:

var trailwall =
            { 
                wall : {
                    positions: [],
                    material : Cesium.Color.RED.withAlpha(0.5),
                }
            }
var trail = datasource.entities.add(trailwall);    

// ... then inside a tick method

trail.wall.positions.setValue(newpositions(time))

hpinkos:
@EKoetsjarjan I'm sorry, I don't understand your question.  Please ask this question on the forum instead: https://groups.google.com/forum/?hl=en#!forum/cesium-dev

We use GitHub exclusively for tracking bugs and planning new features.  A lot more people are on the forum, and your post might help someone else who has the same question.  Members of our team also monitor the forum to make sure all questions get answered.  Thanks!

EKoetsjarjan:
ok thank you

OmarShehata:
I'd like to :+1: this, since it comes up so often, and even our own Sandcastle examples struggle with this. The drawing on terrain Sandcastle uses a CallbackProperty when drawing so it's smooth, but when switches to a static entity when done, which causes a very noticable flicker while the entity is being created asynchronously:

https://cesiumjs.org/Cesium/Build/Apps/Sandcastle/?src=Drawing%20on%20Terrain.html

Being able to say "I'd like this entity to be created synchronously if possible" would fix this.

[...]

PR #4410, opened 2016-10-06
Added allowPicking to Entity to allow per-entity picking.

dwhipps:
This relates to issue #2895.

Please note that this is an initial stab at this and probably needs more work. Also note that while jsHint and test-non-webgl pass, for some reason I can never (even on master branch) get the test-webgl tests to complete, so I suspect that several of my new tests have problems.

At this point I'd just like some comments on my implementation, before I spend a lot of time looking further into this.

Issue #2895, opened 2015-07-20
Only show Infobox for certain entities


hpinkos:
Several users in the forum have been looking for the ability to only show the InfoBox for certain types of entities.  For example, it would be useful to show the InfoBox for pins but not polylines.

One solution could involve this: #1592 (Picking Priority)

https://groups.google.com/forum/?hl=en#!topic/cesium-dev/H_-hJyY_KuA
https://groups.google.com/forum/?hl=en#!topic/cesium-dev/agtOrHYEhUo

OmarShehata:
I think this would solve the issue in this new forum thread: https://groups.google.com/d/msg/cesium-dev/B7QYbtyRPFo/vC2OT9UnBwAJ

They're using a KML data source and would like only the pins to be selected.

PR #13023, opened 2025-11-07
Add asynchronous hint to entities [#5127]

Beilinson:
# Description

This is a draft PR attempting to solve #5127. The issue has comments mentions several use cases, I'd personally like to add the use case of "hover" behavior which currently requires either using callback properties (not efficient on the scale in our app) or the primitive API directly (difficult to support compatibility with the entire Entity API).

The basic concept of this PR:

1. Add getter-only `asynchronous` property to `Entity` and `GeometryUpdater` (analogous to the getter-only `asynchronous` property on `Primitive`), which defaults to `true`
2. Similarly, add the property to all relevant `<TYPE>Batch` classes. Each batch will pass its asynchronous flag to its internal `Primitive`
3. In the `<TYPe>Batch.prototype.add` method, when checking whether to create a new `Batch`, compare the `asynchronous` flags between the existing `Batch`'s and the `updater`, similarly as is done for materials, zIndex, etc...

**Questions:**

1. All the `Dynamic<TYPE>Updater` classes always create a `Primitive` with `asynchronous: false`. I can't see the value in setting this to `true`, so most likely the flag on the `Entity` should just have a comment mentioning this. Should setting `asynchronous: true` on a graphics with a dynamic property throw a runtime/developer error?
2. The pipeline for polyline visualization is pretty different than most other geometry types, luckily `StaticGroundPolylinePerMaterialBatch` at least already supports an `asynchronous` flag it just isn't used. Other than that I think I can get it working there as well
3. Points, billboards, and labels have no such concept of an `asynchronous` creation state. Again I'm assuming that the flag having no effect here is a legitimate behavior. This leads me to the main question in terms of API design: should this be a flag on the `Entity`, or only on the relevant `<TYPE>Graphics` which even support asynchronous creation?

## Issue number and link

#5127

## Testing plan

[Initial example](http://localhost:8080/Apps/Sandcastle2/index.html#c=vVVtT9swEP4rXr+QTsVlYqWiL2hQsoFUVkQH00TQ5CbXxltiR7bTLpv473PspGkbNqRN4kuSs5/n7p7z+ULjhAuFXiMi0QgkTWM0FzxGXsM3ltfoe8xjPmdSoSWFFQg0RAxWBRrfmTWnxI84U4QyEF6jhX55DCEJEfiKcnbJAuoTxUUPzUkkoZXvUjbnZ/xHtfTY7JfhpA8MdDQbFhtTb5o3DmCWLqYhX70XJAZ5DWIKmhZovBIpbGTt84jnSRcJj3ITn41vXbyiKjyNkpA4B7hTxU2o/330FOmLOx5PPm/Rjo+bJlSRIzBFFQWJSRA4Rn7Co2zBWc8WA6GQgiDCD7Pe2jcRSn8Rdojz0p/DQgDIUyFI5txbEkL73QN80EKH5rl/VDfeWqNbGZb60Cw+QqCLUPVQuRETBYKSqLet8cY936mLwT+WR/MSUjtG3Vur7rBmlFI7/yn1iR54ca1HXSPvyCjqdOpGccbdtfFvWj/cuO7HZ8RW3e/merPq6tX1E5kxPxSc8VRu3eegh/ZyF3utZ8p0//CsgvVFXGeqx4m+8jvTwmPtNrriS0AqBJQIGutkjUWUWYp1koCoRBojkOJmUfEEl5pDwoJod7JNfX1gbJoQH9ylrsCFBTl2/viELYnMwxdkLEFdsiRVp2baOfOUmQ/kxDpsrB00bRWqMkMwmX3Tw1HHtU7zxTVcVz245pLmTvI4urhz5BTJBTDXUzZwNt00m2WZbQgoT3EThGnQ3wTZvYvyXDTa0nBxdHh9ZAUtT6LqkToMvRoOd72uE0Por9QasYj5aF8b3LwNqlGvEQh0R5Rh6kDTMBapu6b1pzP+lCWArya3U/fr1eTObfYbrcZAqiyCk5z7jtrfZSoiB+O2gjiJdLfK9izVaSvsS9MRg3ZJGQR0qe/E8InfI/IjIqXemadRNKU/wWucDNoav0WLOAkoW0x040YkyyHhm5OxXcQYD9rarLMU59GMiA2PvwE)

TBD

# Author checklist

### DIFF
--- a/packages/engine/Source/DataSources/Entity.js
+++ b/packages/engine/Source/DataSources/Entity.js
@@ -98,6 +98,7 @@ function createPropertyTypeDescriptor(name, Type) {
  * @property {PolylineVolumeGraphics | PolylineVolumeGraphics.ConstructorOptions} [polylineVolume] A polylineVolume to associate with this entity.
  * @property {RectangleGraphics | RectangleGraphics.ConstructorOptions} [rectangle] A rectangle to associate with this entity.
  * @property {WallGraphics | WallGraphics.ConstructorOptions} [wall] A wall to associate with this entity.
+ * @property {boolean} [asynchronous=true] Determines if the underlying primitives will be created asynchronously or block until ready.
  */
 
 /**
@@ -121,6 +122,7 @@ function Entity(options) {
 
   this._availability = undefined;
   this._id = id;
+  this._asynchronous = options.asynchronous ?? true;
   this._definitionChanged = new Event();
   this._name = options.name;
   this._show = options.show ?? true;
@@ -247,6 +249,16 @@ Object.defineProperties(Entity.prototype, {
   id: {
     get: function () {
       return this._id;
+    },
+  },
+  /**
+   * Gets whether the underlying primitives should be created asynchronously or block until ready.
+   * @memberof Entity.prototype
+   * @type {boolean}
+   */
+  asynchronous: {
+    get: function () {
+      return this._asynchronous;
     },
   },
   /**
--- a/packages/engine/Source/DataSources/GeometryUpdater.js
+++ b/packages/engine/Source/DataSources/GeometryUpdater.js
@@ -50,6 +50,7 @@ function GeometryUpdater(options) {
   const geometryPropertyName = options.geometryPropertyName;
 
   this._entity = entity;
+  this._asynchronous = entity.asynchronous;
   this._scene = options.scene;
   this._fillEnabled = false;
   this._isClosed = false;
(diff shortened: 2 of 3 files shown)


Beilinson:
I also think that the outline for this branch could apply perfectly to something like `allowPicking`, which is also pretty useful because the Entity api does not support this currently (#2895). I found #4410 which was closed with the following comment:

>I'm going to close this due to inactivity. Thanks for the work you put into this @dwhipps! This just isn't quite the approach we want and we haven't had a chance to revisit a different strategy. Hopefully this is something we'll get to before too long.

However it doesn't seem that an alternative has been implemented, leading me to question whether that PR should maybe be reopened, although I do think it wouldn't work correctly with batches (but the method in this PR should work without issue).

Issue #13019, opened 2025-11-05
compile error


aipinsui:
### What happened?

when i excute npm run build-apps,compile error appear,please repair this question

PS G:\cesium> npm run build-apps

> cesium@1.135.0 build-apps
> gulp -f gulpfile.apps.js buildApps

[22:28:12] Using gulpfile G:\cesium\gulpfile.apps.js
[22:28:12] Starting 'buildApps'...
[22:28:12] Starting 'buildCesiumViewer'...
[22:28:12] Starting 'buildLegacySandcastle'...
[22:28:12] Starting 'buildSandcastle'...
[22:28:13] 'buildSandcastle' errored after 819 ms
[22:28:13] TypeError: Invalid URL
    at new URL (node:internal/url:825:25)
    at fileURLToPath (node:internal/url:1605:12)
    at typescriptCompile (file:///G:/cesium/packages/sandcastle/scripts/typescriptCompile.js:13:19)
    at buildStatic (file:///G:/cesium/packages/sandcastle/scripts/buildStatic.js:136:26)
    at buildSandcastleApp (file:///G:/cesium/scripts/buildSandcastle.js:135:9)
[22:28:13] 'buildApps' errored after 821 ms

### Reproduction steps

1.npm run build-apps
2.
3.
...

### Environment

Browser:chrome 142.0.7444.60
CesiumJS Version:1.135
Operating System:windows11

javagl:
(@jjspace This is in the same area as the build issues that I mentioned in https://github.com/CesiumGS/cesium/pull/13013#pullrequestreview-3422070937 , but a diffrent flavor of that - maybe some of the differences are related to the Node version or OS...?)

@aipinsui 

You could try to replace the file https://github.com/CesiumGS/cesium/blob/edceb42f3ac54aa3e3ebef17ab96f3d2f39f133c/packages/sandcastle/scripts/typescriptCompile.js with this one
```
import { spawn } from "node:child_process";
import { join } from "node:path";
import { fileURLToPath } from "node:url";

/**
 * Compile a typescript project from it's config file using the tsc CLI
 *
 * @param {string} configPath Absolute path to the config file to build
 * @returns {number} exit code from the tsc command
 */
export default async function typescriptCompile(configPath) {
  const tsPath = import.meta.resolve("typescript");
  console.log("tsPath", tsPath);
  const fullPath = join(tsPath, "../../bin/tsc")
  console.log("fullPath", fullPath);
[...]

PR #13026, opened 2025-11-07
Fix Sandcastle paths for windows

jjspace:
# Description

- Alter the path built for the `tsc` binary to account for "url" vs "file" path discrepancies on Windows
- Convert all paths to `/` instead of `\` for `globby` which only accepts `/`
    - This is replacing https://github.com/CesiumGS/cesium/pull/12988/ which did have the nicest solution in the end I just wrapped it into this PR at the same time.
- Removed some unused commented code that I noticed was left over from one of my previous PRs

I am unable to easily test these changes myself as they only affected windows but I tried to validate the individual functions with fake paths as much as possible. @javagl you've seen these the most so I'd appreciate if you can validate that things are working for you now.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/13019
Fixes https://github.com/CesiumGS/cesium/issues/12951

## Testing plan

- Ensure the build steps complete with no errors
- run `npm start`
- Ensure that local sandcastle displays the gallery and search works as expected

# Author checklist


javagl (commented):
This caused the same issue as @Beilinson already mentioned in https://github.com/CesiumGS/cesium/issues/13019#issuecomment-3505318785 : It says

> Error: spawn C:/cesium/node_modules/typescript/bin/tsc ENOENT

in the `build-sandcastle` step.

Near the linked comment, you already said

> I wonder if the binary isn't set to be executable on Windows

And ... this is ... "true", in some way. It is not a binary. And the concept of "being executable" does not exist on Windows in the same way as it does on Linux. The file **does** exist (no comment about `ENOENT` at this point). But it is a file that only contains
```
#!/usr/bin/env node
require('../lib/tsc.js')
```
I'd expect it to be the same on Linux. But on Linux, one can just set the "executable" flag, and then, the first line tells Linux how to execute that thing (roughly like that...)

---

The linked issue already suggested using
`const ls = spawn(process.execPath, [binPath, "-p", configPath]);`

And from my understanding, this _should_ make sense: The `process.execPath` will be
`C:\Program Files\nodejs\node.exe`
as expected. The `binPath` is what is shown above. So this is explicitly using `node` to execute that file, which (I think) is the same what happens under Linux, but there, this is accomplished with that
`#!/usr/bin/env node`
line. 

One important point - referring to your comment at https://github.com/CesiumGS/cesium/issues/13019#issuecomment-3503629059 - is that this is **not** using the "globally installed `npx` (with its arbitrary version). It is really executing the exact `tsc` that is found in the `node_modules` directory.

Given the lack of alternatives (with the "Compiler API" being "version 0.5 beta" for 2 years now...), this sounds like a reasonabe solution, unless I'm overlooking something.

javagl:
And I didn't mention that: The "globby" fix also seems to work - it **does** display the gallery.

---

BTW: `npm run build-apps` still causes

> Error: ENOENT: no such file or directory, scandir 'C:\cesium\Build\Cesium\Assets'

**but** (as discussed elsewhere), the sequence of
`npm run build-release`
`npm run build-apps` 
**does** work.

---

My "broken record" mantra about documentation and such:

I think that the build infrastructure could and should(!) be simplified _significantly_. The 1500-line(!)-gulpfile could be shrunk to 1000 lines by removing redundancies. It should then be increased to 2000 lines by adding extensive documentation for _each and every function_. This 2000-line-file should then be broken down into at least 10 files with generic utility functionalities. One prominent example is the "generate third party" functionality. If this was implemented **once**, in a proper form, and in a standalone file, then the same file could be used in CesiumJS, in the [`3d-tiles-validator`](https://github.com/CesiumGS/3d-tiles-validator/blob/c1d3c4d2f013d57755d1ba2ce346a0c3d0bb97d0/generateThirdParty.js#L62), [`wetzel`](https://github.com/CesiumGS/wetzel/blob/f30c4fb4655facfcf91fca745f6188d5646c3f21/gulpfile.js#L10), the [`3d-tiles-tools`](https://github.com/CesiumGS/3d-tiles-tools/blob/24bf3b8a90074b5095f51bfaafbb5f2dbf11cf44/generateThirdParty.js#L99), in [`gltf-pipeline`](https://github.com/CesiumGS/gltf-pipeline/blob/4763d6aafbac8600787a850a093eb76fa6599849/gulpfile.js#L327), and about a dozen private repos where this functionality was duplicated.

(In fact, the `3d-tiles-tools` and `3d-tiles-validator` versions of that already _are_ (nearly...) equal, and already contain _some_ documentation (but could contain more). Similarly, there are [command documentations](https://github.com/CesiumGS/3d-tiles-tools/blob/24bf3b8a90074b5095f51bfaafbb5f2dbf11cf44/package.json#L71) and [hierarchical commands](https://github.com/CesiumGS/3d-tiles-tools/blob/main/IMPLEMENTATION.md#build-scripts) - I think that all of this could help longer-term maintainability and flexibility (e.g. so that you can execute specific steps, dedicatedly, with `npm run <category>-<substep>-<task>`-like calls). But ... let's first get that `\`-vs-`/` thing right...)

wewindy:
Checkout this branch, I add some logs and run `npm run build-sandcastle`, got:

PixPin_2025-11-08_22-22-57

wewindy:
But when I change to:

```diff
 export default async function typescriptCompile(configPath) {
   const tsPath = fileURLToPath(import.meta.resolve("typescript"));
-  const binPath = join(tsPath, "../../.bin/tsc");
+  const binPath = join(tsPath, "../../../.bin/tsc.cmd");
   return new Promise((resolve, reject) => {
-    const ls = spawn(binPath, ["-p", configPath]);
+    const ls = spawn(binPath, ["-p", configPath], { shell: true });

     ls.stdout.on("data", (data) => {
       console.log(`stdout: ${data}`);
     });

     ls.stderr.on("data", (data) => {
       console.error(`stderr: ${data}`);
     });

     ls.on("close", (code) => {
       if (code === 0) {
         resolve(code);
       } else {
         reject(code);
       }
     });
   });
 }
```

It works.

Okay, I think I know what's going on. Using Notepad, I opened `node_modules/typescript/bin/tsc`:

```
#!/usr/bin/env node
require('../lib/tsc.js')
```

and `node_modules/.bin/typescript/tsc`:

```
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*|*MINGW*|*MSYS*)
        if command -v cygpath > /dev/null 2>&1; then
            basedir=`cygpath -w "$basedir"`
        fi
    ;;
esac

if [ -x "$basedir/node" ]; then
  exec "$basedir/node"  "$basedir/../typescript/bin/tsc" "$@"
else 
  exec node  "$basedir/../typescript/bin/tsc" "$@"
fi
```

They are actually script files with shebang lines.

As we known, Windows does not support Unix shebang, so we can only call the `tsc.cmd` file with spawn function, and need to use the `shell: true` parameter.

The `bin/tsc` provided by the Node.js TypeScript package is not an executable program, but a script. In fact, `typescript` is a complete JavaScript package that uses JavaScript to translate TypeScript code.

jjspace:
> It is not a binary. And the concept of "being executable" does not exist on Windows in the same way as it does on Linux.

Yet another reason windows is incredibly annoying.

I've updated this to use the `process.execPath`. I don't think we want to change which tsc file it's pointing to. This should hopefully work for everyone now regardless of system :crossed_fingers:
It seems this has already been confirmed to work in multiple comments but should still be re-tested.

@mzschwartz5 could you please just checkout this branch and run `npm run build-sandcastle` to confirm it works on mac?
@javagl can you please review again and merge if happy? Not a hard requirement for the release but would be good to get in if we can.

javagl:
I tested it with a fresh clone on this branch, and it works as expected. I'll wait a bit for a possible test on Mac, but would otherwise merge this later today or early tomorrow.

mzschwartz5:
As far as I can see, works for me!

Issue #4525, opened 2016-10-25
Outline around label's characters and billboards in clustering demo


rutsky:
I often (but not always) see gray border around characters and billboards in [clustering demo](http://cesiumjs.org/Cesium/Apps/Sandcastle/index.html?src=Clustering.html&label=Showcases):

![2016-10-25_21 55 37_1](https://cloud.githubusercontent.com/assets/46573/19699974/751484a4-9b06-11e6-9841-43f94b3b2908.png)

![2016-10-25_22 01 01_1](https://cloud.githubusercontent.com/assets/46573/19700111/07ed2d44-9b07-11e6-93d3-f4a7fb762161.png)

Tested on Ubuntu 16.04 with Firefox 51.0a2 and Chromium 53.0.2785.143.

rutsky:
Same issue on [labels demo](http://cesiumjs.org/Cesium/Apps/Sandcastle/index.html?src=Labels.html&label=Showcases):

![2016-10-25_22 07 40_1](https://cloud.githubusercontent.com/assets/46573/19700326/e48bf1fe-9b07-11e6-9878-db0c34c83699.png)

![2016-10-25_22 07 29_1](https://cloud.githubusercontent.com/assets/46573/19700329/e8cb2f3c-9b07-11e6-9cbb-47fba9fbe884.png)

![2016-10-25_22 07 21_1](https://cloud.githubusercontent.com/assets/46573/19700331/eba2f528-9b07-11e6-81f7-568ac7dabb6f.png)

rutsky:
This is not reproduced when I use integrated graphics from Intel on the same Linux machine.

rutsky:
Might be related: #2130

rutsky:
@mramato since you've seen similar issue as you noted in [this](https://github.com/AnalyticalGraphicsInc/cesium/issues/2130#issuecomment-106612931) comment, maybe you can give some hints what is the reasons of this bug and how can it be resolved?

mramato:
@bagnell believes this may be related to texture coordinates being slightly off for billboards. The odd part is that it only happens on some systems, so we could have a case of undefined behavior somewhere or a rounding issue.  Billboard collection (particular the shaders) would be the first place I look to try and debug this.

elFerZur:
I have the same problem with billboards, of any type (gif, png, svg). Tested on Windows 10 with Firefox 49 and Intel graphics card.

[...]

PR #13037, opened 2025-11-18
chore(textureatlas): Refactor unit tests for visual output

donmccurdy:
# Description

Related to #4525, I'm planning to make a change to the texture atlas packing algorithm in a future PR, but was nervous to make changes to packing without an easier way to see (and unit test) the changes. This PR makes no changes to current atlas packing, but refactors the unit tests so that most of the tests are more visually-legible. I also added one new test case, "computes subregion texture coordinates", to cover some texcoord-related cases otherwise left uncovered in my refactoring. Many of the tests will rely on visual string comparison on output, like:

```plaintext
................................
.2222222222222222333333333301...
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.22222222222222223333333333.....
.2222222222222222...............
.2222222222222222...............
.2222222222222222...............
.2222222222222222...............
.2222222222222222...............
.2222222222222222...............
................................
```

Here, `.` represents unused space and integers represent image or subregion indices in the atlas. Importantly, this makes it easy to see where padding might be missing, or where the atlas might not be using free space efficiently.

## Issue number and link

Related to, but does not fix, #4525. Next PR will include changes to the texture packing algorithm to fix the underlying issue.

## Testing plan

n/a, refactoring unit tests.

# Author checklist

#### PR Dependency Tree

* **PR #13037** 👈
  * **PR #13038**


jjspace (commented):
Overall this seems like a much cleaner way to test these functions so I'm a fan. I don't really know the inner workings of the texture atlas so I can't say with 100% certainty that these still _fully_ test functionality. That said the tests are still passing in CI and locally for me so that's good.

That said, could you change the "empty" character from `•` to `.` or `-` to make it easier to type. The middle dot definitely looks nice but I don't think devs should have to copy/paste it a bunch if they can't easily type it. (Same reason I hate emojis in code)

donmccurdy:
Thanks @jjspace! I'm happy to add more tests if we can think of anything missing. Just hoping that we can unit test the numerical bits (like `computeTextureCoordinates`) in isolation so that the more high-level tests (reasonable packing and padding between images?) don't have to juggle numbers too much.

Good call about the spacing character, fixed! I think the period is the winner, hyphens and underscores were being ligature'd into a solid line in my editor, which made it harder to count pixels.

Issue #5154, opened 2017-03-27
Billboard clipped on right side


spatialillusions:
In Cesium 1.31

I have a large number (500+) of quite large (200x100pixel) billboards on my map, and this problem only occurs when the number of billboards is large. The billboard image are canvases that I draw with javascript within the application.

For some reason, the billboards are clipped on the right hand side:
<img width="526" alt="screen shot 2017-03-27 at 19 33 00" src="https://cloud.githubusercontent.com/assets/8155542/24369814/7d63a1de-1325-11e7-85a5-a02eaba99c26.png">

This can be compared to using the same canvas images as icons in Open Layers:
<img width="613" alt="screen shot 2017-03-27 at 19 43 17" src="https://cloud.githubusercontent.com/assets/8155542/24369932/ea0323d2-1325-11e7-9a63-6d79b056229a.png">

What is really strange, is that I can almost solve the problem if I set the imageSubRegion to be bigger than the canvas itself.

```js
var entity = {
  position: Cesium.Cartesian3.fromDegrees(feature.geometry.coordinates[0], feature.geometry.coordinates[1]), 
  billboard: {
    horizontalOrigin : Cesium.HorizontalOrigin.LEFT, 
    verticalOrigin : Cesium.VerticalOrigin.TOP,
    image: ctx,
    imageSubRegion: new Cesium.BoundingRectangle(0, 0, ctx.width+2, ctx.height+2),
    height: milsymbol.getSize().height,
    width: milsymbol.getSize().width,
    pixelOffset : new Cesium.Cartesian2(-milsymbol.getAnchor().x, -milsymbol.getAnchor().y)
  }
}
```
Now this will be displayed:
<img width="566" alt="screen shot 2017-03-27 at 19 34 51" src="https://cloud.githubusercontent.com/assets/8155542/24370038/3af79afc-1326-11e7-9b68-4e24a710ab4c.png">
*Better but not perfect*

If we zoom out a lot, we can also see that some of the missing lines from the right hand side is rendered on the left hand side of some symbols:
<img width="321" alt="screen shot 2017-03-27 at 19 52 39" src="https://cloud.githubusercontent.com/assets/8155542/24370290/32b483f4-1327-11e7-987e-6de6c4f7091c.png">
*Notice the vertical lines that shouldn't be there*

I figured out that Cesium stores the images in some kind of atlas, but I couldn't find where in the code this was done, and what could be the reason for this error. Please let me know if you need more information about this.

[...]

PR #13020, opened 2025-11-05
fix(billboards): Rasterize SVG content at billboard resolution

donmccurdy:
# Description

This PR is a draft for discussion, proposing two related changes when rendering SVG content in billboards:

1. When billboard.image is a URL for an SVG, rasterize the SVG using billboard width/height if specified.
2. Extend JSDoc types to allowing billboard.image to be an HTMLImageElement (already supported, other than types)

These changes are intended to improve visual quality for SVG content in billboards. Because SVGs are scalable, the internal viewBox and width/height are not necessarily the size the user wants to display, and if the user has added a billboard at 50px x 50px size, the billboard will appear pixelated if the (somewhat arbitrary) internal size of the SVG is different. Either of these changes independently provides a workaround for the issue. With (1) the SVG is automatically rasterized at billboard resolution, and with (2) the user can construct the HTMLImageElement themselves, resize it, and pass that to the billboard constructor.

It's not obvious to me that these changes, especially (1), are the 'right' approach ... is it OK for the billboard size to determine the resolution of the SVG in the texture atlas? ~~What if billboard size were in meters?~~ (fixed) But it seemed like a good place to start discussion.

Test code:

```javascript
viewer.entities.add({
  position: Cartesian3.fromDegrees(-75.59777, 40.03883),
  billboard: { image: "icon.svg", width: 100, height: 100 },
});

viewer.entities.add({
  position: Cartesian3.fromDegrees(-50.59777, 40.03883),
  billboard: { image: "icon.svg" }
});
```

Before:
before

After:
after

Remaining work and known issues:

- [x] Unit tests not yet added or updated
- [x] If the same SVG URL is provided for two billboards of different size, the first determines the texture atlas resolution and the second will use a cached texture.
  - See discussion below.
- [x] Changes to the billboard width/height probably will not change the resolution of the texture (untested).
  - See discussion below.

## Issue number and link

Related:

- #4235

This PR doesn't fully fix the issue, but improves quality when the billboard and SVG sizes don't match. More fixes (probably related to pixel alignment and anti-aliasing, as discussed in the issue) would be good followups in another PR.

## Testing plan

[...]

PR #13038, opened 2025-11-18
fix(textureatlas): Apply internal padding between images

### LINKED ISSUE #4525: Outline around label's characters and billboards in clustering demo

donmccurdy:
# Description

The texture atlas used for billboard rendering has 1px padding enabled by default (`borderWidthInPixels`), but the padding is only applied around the edges of the atlas — not between images. I believe this is the cause, or at least a cause, of #4525. This PR introduces internal padding, so that the existing option affects spacing between images in the atlas. Image subregions have no padding and are unaffected.

Screenshots:

| before | after |
|--------|-------|
| border_before | border_after |

## Issue number and link

- Fixes #4525
- Fixes half of #5154

## Testing plan

See unit tests and #13037, and:

http://localhost:8080/Apps/Sandcastle2/index.html?id=itwin-feature-service

With 'points' (pins) enabled and all other features hidden, this is a good example showing the top of one pin bleeding into the bottom of the pin above it in the texture atlas.

# Author checklist

#### PR Dependency Tree

* **PR #13038** 👈


packages/engine/Specs/Renderer/TextureAtlasSpec.js
```diff
@@ -832,16 +832,16 @@ describe("Scene/TextureAtlas", function () {
...
-•22222222222222223333333333•••••
-•22222222222222223333333333•••••
-•22222222222222223333333333•••••
-•22222222222222223333333333•••••
-•22222222222222223333333333•••••
-•22222222222222223333333333•••••
-•22222222222222223333333333•••••
+•2222222222222222•3333333333•0••
+•2222222222222222•3333333333••••
+•2222222222222222•3333333333•1••
```
jjspace: When I reviewed #13037 I assumed this representation is "top down" as it would be rendered. However if I understand the change here (which I might not) this seems to be adding an extra row of pixels to the "right" in this representation but the "bottom" in the actual billboard. Do I just not understand how the texture atlas works? Or should this representation be rotated?

donmccurdy: Ah, the origin is bottom-left and I'd assumed top-left. I don't think that changes anything implementation-wise. But certainly debugging will be easier if the test output uses the right representation, so I'll update the test output to match.

Will respond to the other part of this in the main comment thread.

donmccurdy:
> I see that adding padding to the atlas works but that's just filling the last row of pixels with transparency isn't it? 

@jjspace previously the atlas put an N-pixel border around the perimeter of the entire atlas, then packed billboards tightly inside of that. With this PR the primary goal is to have interior N-pixel borders between billboards as well.

Implementation-wise, there's more than one way to get there. TexturePacker.js implements [binary tree bin packing](https://jakesgordon.com/writing/bin-packing/), where each time a billboard is added, we find an empty rectangle (the large outer square below), insert the image (yellow tile below), and create two new empty rectangles in the remaining space for future use:

large square divided into an image tile surrounded by 2 empty rectangles

> Image Credit: Jake Gordon

The solution in this PR is that each time we do an insertion and create those two empty rectangles in the remaining space, we offset them from the inserted billboard by N pixels so that space isn't considered available:

split_annotated

Entirely possible there are other better ways to do this! With the visual tests I'm hoping it will be easier to see the tradeoffs if we decide to change how the padding is handled.

***

> Would it be better to say that the billboards are displaying 1 px too much from the atlas?

I think it's a combination of linear filtering in the texture sampler and perhaps some numerical precision limits. I also tested switching to 'nearest' filtering and while that definitely reduces the visual artifacts here, it doesn't solve them entirely (which this PR does, as far as I can tell). I may do a separate PR for the texture filtering, but that has other tradeoffs and wouldn't be the right fix for this specific issue probably.

In https://github.com/CesiumGS/cesium/pull/13020 there was also some discussion of mipmapping billboards, and padding between billboards would be a hard prerequisite to that.

Finally, I suspect this may have been the original intention of the TexturePacker.js `borderWidthInPixels` option? Or at least I can't think of a reason that padding on the atlas perimeter would have been useful without also including internal padding.

packages/engine/Specs/Renderer/TextureAtlasSpec.js
donmccurdy: Both PRs have been updated so that the `drawAtlas` test output matches the coordinate system in TextureAtlas.js, origin = bottom-left. Thanks @jjspace!

packages/engine/Source/Core/TexturePacker.js
```diff
@@ -130,12 +132,18 @@ TexturePacker.prototype._findNode = function (node, { width, height }) {
...
-        width: widthDifference,
-        height: nodeHeight,
-      });
+
+      // Apply padding only along the vertical "cut".
+      const widthDifferencePadded = widthDifference - borderPadding;
+
+      if (widthDifferencePadded > 0) {
+        node.childNode2 = new TextureNode({
+          x: rectangle.x + width + borderPadding,
```
jjspace, line 141: Given that border padding is a dynamic value would it be good to have a test with at least one other size than 1px? Maybe like 3 or something? just to prove it's working in a "non default" case.

donmccurdy: Good call, done! New test case: "adds custom padding with borderWidthInPixels".

donmccurdy:
Added the new test case, good call on that, thanks @jjspace!

jjspace (approved):
LGTM, thanks for the updates

PR #13042, opened 2025-11-20
fix(textureatlas): Allocate texture space more incrementally on resize

donmccurdy:
# Description

Texture coordinates are currently compressed to 12-bit precision, effectively on a 4096x4096 grid. That could be revisited (see #13050), but regardless, billboard-related code must be careful with texture coordinate precision, especially in large atlases. This PR makes two changes along those lines:

1. Atlases are now constrained to power-of-two dimensions on resize, aligning the atlas pixel grid cleanly with the 4096x4096 texture coordinate quantization grid, and ensuring we get the same visual results on WebGL1- and WebGL2-compatible devices.
2. When resizing atlases, TextureAtlas.js now considers more candidate sizes, and in most cases should not allocate space in the atlas with a [packing density](https://en.wikipedia.org/wiki/Packing_density) <<0.5.

Previously as the current candidate atlas size became larger, the iterative resize algorithm would use a larger multiplier to select the next candidate size, (2x, 4x, 8x, 16x, ...). In the particular case of #5154 this created a 2K-by-8K atlas, which required 16x more memory than required and (more visibly) exceeded the 4096px limit at which the 12-bit texcoord quantization could precisely address pixels in the atlas. 

Preview:

| before | after |
|--------|-------|
| Screenshot 2025-11-20 at 5 04 37 PM | Screenshot 2025-11-20 at 5 05 19 PM |

This improves image alignment with each billboard, but isn't a complete fix for all such issues. More changes in upcoming PRs, but this change seemed worthwhile on its own.

## Issue number and link

- Partial fix for #5154

## Testing plan

I'm looking through existing examples with one or more billboards. For each relevant example, compare computed texture atlas size before vs. after. Check that atlas size is "reasonable", i.e. not >> 2x the total area of the input images.

- [iTwin Feature Service](http://localhost:8080/Apps/Sandcastle2/index.html?id=itwin-feature-service)
  - Before: 768 x 3072
  - After: 1024 x 1024
- [Billboards](http://localhost:8080/Apps/Sandcastle2/index.html?id=billboards)
  - Before: 236 x 120
  - After: 128 x 32
- [Clustering](http://localhost:8080/Apps/Sandcastle2/index.html?id=clustering)
  - Before: 264 x 1040, 752 x 220, 768 x 192, 256 x 256
  - After: 512 x 512, 256 x 128, 256 x 256
- [Map Pins](http://localhost:8080/Apps/Sandcastle2/index.html?id=map-pins)
  - Before: 192 x 192
  - After: 128 x 128
- [GeoJSON simplestyle](http://localhost:8080/Apps/Sandcastle2/index.html?id=geojson-simplestyle)
  - Before: 768 x 3072
  - After: 1024 x 512
- [Particle System Tails](http://localhost:8080/Apps/Sandcastle2/index.html?id=particle-system-tails)
  - Before: 320 x 1280
  - After: 256 x 256
- [LocalToFixedFrame](http://localhost:8080/Apps/Sandcastle2/index.html?id=localtofixedframe)
  - Before: 264 x 912
  - After: 512 x 256
- [Moon](http://localhost:8080/Apps/Sandcastle2/index.html?id=moon)
  - Before: 848 x 232 
  - After: 256 x 256
- [Mars](http://localhost:8080/Apps/Sandcastle2/index.html?id=mars)
  - Before: 1072 x 268, 768 x 192
  - After: 512 x 512, 256 x 256
- [Particle System Fireworks](http://localhost:8080/Apps/Sandcastle2/index.html?id=particle-system-fireworks)
  - Before: 10240 x 1280
  - After: 1024 x 1024

# Author checklist

#### PR Dependency Tree

* **PR #13042** 👈
  * **PR #13050**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,10 +6,11 @@
 
 #### Fixes :wrench:
 
-- Improved scaling of SVGs in billboards [#13020](https://github.com/CesiumGS/cesium/issues/13020)
+- Improved scaling of SVGs in billboards [#13020](https://github.com/CesiumGS/cesium/pull/13020)
 - Billboards using `imageSubRegion` now render as expected. [#12585](https://github.com/CesiumGS/cesium/issues/12585)
 - Fixed depth testing bug with billboards and labels clipping through models [#13012](https://github.com/CesiumGS/cesium/issues/13012)
-- Fixed unexpected outline artifacts around billboards [#13038](https://github.com/CesiumGS/cesium/issues/13038)
+- Fixed unexpected outline artifacts around billboards [#4525](https://github.com/CesiumGS/cesium/issues/4525)
+- Fix texture coordinates in large billboard collections [#13042](https://github.com/CesiumGS/cesium/pull/13042)
 
 #### Additions :tada:
 
--- a/packages/engine/Source/Renderer/TextureAtlas.js
+++ b/packages/engine/Source/Renderer/TextureAtlas.js
@@ -364,26 +364,17 @@ TextureAtlas.prototype._resize = function (context, queueOffset = 0) {
     toPack.push(queue[i]);
   }
 
-  // At minimum, the texture will need to scale to accommodate the largest width and height
-  width = Math.max(maxWidth, width);
-  height = Math.max(maxHeight, height);
+  // At minimum, atlas must fit its largest input images. Texture coordinates are
+  // compressed to 0–1 with 12-bit precision, so use power-of-two size to align pixels.
+  width = CesiumMath.nextPowerOfTwo(Math.max(maxWidth, width));
+  height = CesiumMath.nextPowerOfTwo(Math.max(maxHeight, height));
 
-  if (!context.webgl2) {
-    width = CesiumMath.nextPowerOfTwo(width);
-    height = CesiumMath.nextPowerOfTwo(height);
-  }
-
-  // Determine by what factor the texture need to be scaled by at minimum
-  const areaDifference = areaQueued;
-  let scalingFactor = 1.0;
-  while (areaDifference / width / height >= 1.0) {
-    scalingFactor *= 2.0;
-
-    // Resize by one dimension
+  // Iteratively double the smallest dimension until atlas area is (approximately) sufficient.
+  while (areaQueued >= width * height) {
     if (width > height) {
-      height *= scalingFactor;
+      height *= 2;
     } else {
-      width *= scalingFactor;
+      width *= 2;
     }
   }
 
--- a/packages/engine/Specs/Renderer/TextureAtlasSpec.js
+++ b/packages/engine/Specs/Renderer/TextureAtlasSpec.js
@@ -704,56 +704,16 @@ describe("Scene/TextureAtlas", function () {
         expect(index2).toEqual(2);
         expect(index3).toEqual(3);
 
-        // Webgl1 textures should only be powers of 2
-        const isWebGL2 = scene.frameState.context.webgl2;
-        const textureWidth = isWebGL2 ? 20 : 32;
-        const textureHeight = isWebGL2 ? 32 : 16;
+        const textureWidth = 32;
+        const textureHeight = 16;
 
         const texture = atlas.texture;
         expect(texture.pixelFormat).toEqual(PixelFormat.RGBA);
         expect(texture.width).toEqual(textureWidth);
         expect(texture.height).toEqual(textureHeight);
 
-        if (isWebGL2) {
-          expect(drawAtlas(atlas, [index0, index1, index2, index3])).toBe(
-            `
-....................
-....................
-....................
-....................
-....................
-....................
-2222222222..........
-2222222222..........
-2222222222..........
-2222222222..........
-2222222222..........
-2222222222..........
-2222222222..........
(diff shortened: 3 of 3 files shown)


packages/engine/Source/Renderer/TextureAtlas.js
```diff
@@ -364,26 +364,17 @@ TextureAtlas.prototype._resize = function (context, queueOffset = 0) {
...
-  if (!context.webgl2) {
-    width = CesiumMath.nextPowerOfTwo(width);
-    height = CesiumMath.nextPowerOfTwo(height);
-  }
-
-  // Determine by what factor the texture need to be scaled by at minimum
-  const areaDifference = areaQueued;
-  let scalingFactor = 1.0;
-  while (areaDifference / width / height >= 1.0) {
-    scalingFactor *= 2.0;
```
donmccurdy, line 380: In practice this loop can't spend more than 28 iterations for the largest supportable atlas size, because we'll have ... ahem, other challenges ... if the texture atlas size exceeds 16K x 16K. No need to use an increasing scaling factor (which may cause us to skip smaller qualifying atlas sizes) to reduce the number of iterations.

donmccurdy:
Aside: Constraining to power-of-two dimensions is not strictly required to solve the precision/alignment issue, as the fix proposed in #13050 would address that issue independently.

I still personally feel that it's worthwhile. Fewer codepaths to think about, both manually and in unit tests, makes it easier to avoid precision bugs like those addressed by this PR. And we do at least need the power-of-two dimensions on WebGL 1 devices — so that code path can't go away at this time. See #13053 for background on why fewer code branches might be helpful for billboards. Even with power-of-two atlas dimensions, this PR still provides a consistent reduction in texture atlas sizes, reducing memory compared to earlier releases.

CHANGES.md
```diff
@@ -10,6 +10,7 @@
 - Billboards using `imageSubRegion` now render as expected. [#12585](https://github.com/CesiumGS/cesium/issues/12585)
 - Improved scaling of SVGs in billboards [#13020](https://github.com/CesiumGS/cesium/issues/13020)
 - Fixed unexpected outline artifacts around billboards [#13038](https://github.com/CesiumGS/cesium/issues/13038)
+- Fixed image alignment in large billboard collections [#13042](https://github.com/CesiumGS/cesium/issues/13042)
```
jjhembd: I think the link is wrong.
Normally we link to the issue we're fixing. In this case, since it's only a partial fix, maybe a link to the PR does make sense?

javagl: GitHub cleverly resolves any `/issues/<number>/` and `/pull/<number>` to the respective `/pull/` or `/issues/` version, just based on the number. (I.e. this link, even though it (literally) has `/issues/`, is redirected to **this** PR).

donmccurdy: Ahh I'd copy/pasted the PR number without noticing that we linked to the issue not the PR, thanks! I did the same on my last two PRs as well, now fixed in this PR. 😅

Agreed, I might prefer to link to the PR instead of the issue in this particular case, since there's another bug remaining to fix for the original issue.

donmccurdy:
Thanks @jjhembd! 

> In your testing plan, how are you checking what size textures were allocated?

My approach was to put a console.log statement here...

https://github.com/CesiumGS/cesium/blob/3d78834590b2416c7bb369c245be4b172b5cf4f7/packages/engine/Source/Renderer/TextureAtlas.js#L438

... and click through the demos in sandcastle manually. I could see an argument for logging a console warning (developer-facing only?) if the texture atlas exceeds 4K on any side, since that will (1) require more precision than current texture coordinate compression, and (2) likely fail on on some devices, particularly mobile. That might have made it easier to spot the issue in this case, but I haven't totally made up my mind about added logging to the PR -  happy to add it, and would be curious if unconditional `console.warn` or `//>>includeStart('debug', pragmas.debug);` would be preferred.

CHANGES.md
```diff
@@ -6,10 +6,11 @@
 
 #### Fixes :wrench:
 
-- Improved scaling of SVGs in billboards [#13020](https://github.com/CesiumGS/cesium/issues/13020)
+- Improved scaling of SVGs in billboards [#13020](https://github.com/CesiumGS/cesium/pull/13020)
 - Billboards using `imageSubRegion` now render as expected. [#12585](https://github.com/CesiumGS/cesium/issues/12585)
 - Fixed depth testing bug with billboards and labels clipping through models [#13012](https://github.com/CesiumGS/cesium/issues/13012)
-- Fixed unexpected outline artifacts around billboards [#13038](https://github.com/CesiumGS/cesium/issues/13038)
+- Fixed unexpected outline artifacts around billboards [#4525](https://github.com/CesiumGS/cesium/issues/4525)
+- Fix texture coordinates in large billboard collections [#13042](https://github.com/CesiumGS/cesium/pull/13042)
```
jjhembd, line 13: @donmccurdy sorry I missed this earlier: this changelog entry should have been under a new heading for the January release (1.137).
I can fix it in the next PR.

donmccurdy: Ah thanks for the catch! Just opened a PR:

- https://github.com/CesiumGS/cesium/pull/13042

PR #13050, opened 2025-11-24
fix(BillboardCollection): Fix precision loss in billboard image texcoords

### LINKED ISSUE #5154: Billboard clipped on right side

donmccurdy:
# Description

Currently billboards compute texture coordinates (in 0–1 UV space) for each image in the texture atlas:

https://github.com/CesiumGS/cesium/blob/ee2b3813b277ad0b4f92ab3c56987b8793ae1d0f/packages/engine/Source/Renderer/TextureAtlas.js#L195-L206

Texture coordinates are then quantized to 12 bits: 

https://github.com/CesiumGS/cesium/blob/ee2b3813b277ad0b4f92ab3c56987b8793ae1d0f/packages/engine/Source/Core/AttributeCompression.js#L308-L317

On testing all valid pixel offsets for all power-of-two atlas sizes...

```javascript
const src = {x: 0, y: 0}, dst = {x: 0, y: 0};
for (let exp = 1; exp <= 12; exp++) {
  const width = 2 ** exp;
  for (let x = 0; x <= width; x++) {
    src.x = x / width;

    const encoded = AttributeCompression.compressTextureCoordinates(src);
    AttributeCompression.decompressTextureCoordinates(encoded, dst);

    const dstX = Math.round(dst.x * width);
    assert(dstX === x, `❌ Expected dstX=${dstX} to be ${x}`);
  }
}
```

... we find coordinates off by 1px about 25% of the time, which I believe explains the remaining issues from #5154. While we _could_ modify the texture compression code to fix this, (1) texture coordinates outside of billboards may not correspond to integer pixel offsets at all, and (2) it's worth stepping back to consider what we're trying to do:

Ultimately each Billboard must sample from its image in the texture atlas. Images have integer pixel offsets and integer dimensions, which cannot exceed maximum texture sizes of ~16K. We need to pack pairs of these integers into 32-bit vertex attribute components. We can achieve this more simply by packing two integers into one 32-bit float directly...

```javascript
const encoded = x * (2 ** 16) + y;
```

... skipping both integer-to-float conversions and lossy quantization.

Preview:

TODO

## Issue number and link

- Fixes #5154

## Testing plan

Because coordinates and attribute packing differs in WebGL2 vs. WebGL1, and with or without instancing, we need to test multiple environments:

1. WebGL2, instancing
2. WebGL1, instancing
3. WebGL1, no instancing

WebGL2 guarantees instancing support, so we don't need to worry about that case. If #13042 is merged before this PR (preferred) then the WebGL1 vs. WebGL2 distinction is removed and we only need to worry about instanced vs. non-instanced code paths. Optionally instancing could be required (see #13053) but I've handled the non-instanced fallback here anyway.

More detailed test plan TBD. 

# Author checklist

#### PR Dependency Tree

* **PR #13042**
  * **PR #13050** 👈

### DIFF
--- a/packages/engine/Source/Renderer/TextureAtlas.js
+++ b/packages/engine/Source/Renderer/TextureAtlas.js
@@ -193,22 +193,21 @@ Object.defineProperties(TextureAtlas.prototype, {
 });
 
 /**
- * Get the texture coordinates for reading the associated image in shaders.
+ * TODO
  * @param {number} index The index of the image region.
  * @param {BoundingRectangle} [result] The object into which to store the result.
  * @return {BoundingRectangle} The modified result parameter or a new BoundingRectangle instance if one was not provided.
  * @private
  * @example
  * const index = await atlas.addImage("myImage", image);
- * const rectangle = atlas.computeTextureCoordinates(index);
+ * const rectangle = atlas.computeImageCoordinates(index);
  * BoundingRectangle.pack(rectangle, bufferView);
  */
-TextureAtlas.prototype.computeTextureCoordinates = function (index, result) {
+TextureAtlas.prototype.computeImageCoordinates = function (index, result) {
   //>>includeStart('debug', pragmas.debug);
   Check.typeOf.number.greaterThanOrEquals("index", index, 0);
   //>>includeEnd('debug');
 
-  const texture = this._texture;
   const rectangle = this._rectangles[index];
 
   if (!defined(result)) {
@@ -224,11 +223,6 @@ TextureAtlas.prototype.computeTextureCoordinates = function (index, result) {
     return result;
   }
 
-  const atlasWidth = texture.width;
-  const atlasHeight = texture.height;
-
-  const width = rectangle.width;
-  const height = rectangle.height;
   let x = rectangle.x;
   let y = rectangle.y;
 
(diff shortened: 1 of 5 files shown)


donmccurdy:
Closing this PR for now. As we're very close to the end of the deprecation period for #13053, I think it makes sense to wait and attempt this again after some of the fallback billboard code branches have been removed.

PR #13054, opened 2025-11-25
Prevent sandcastle "double loading"

jjspace:
# Description

The loading logic for sandcastle didn't always wait for the sandcastle to load before displaying the code and letting the viewer run. This created a weird "double load" when loading a specific sandcastle url where it first showed the default globe then the actual sandcastle.

I think the issue was partly due to how I wanted to lower the delay for _new_ sandcastles in the case that we don't need to actually load any code. Then I think the issue it was somewhat exaggerated by https://github.com/CesiumGS/cesium/pull/12854. This became a bigger issue when I started working on ion oauth and looking at modifying the default code before "first load" which the current structure doesn't really support.

In this PR sandcastle will now show `// Loading...` in the editor and "Loading..." in the viewer "bucket" before the code is actually loaded. I do think we can improve this more in the future to hopefully shorten load times but the way it is now feels even more broken.

## Issue number and link

No issue

## Testing plan

- Open sandcastle and make sure the default globe loads
- Change your default starting page to the editor and make sure it also loads correctly with a single viewer render
- Open a specific sandcastle and refresh. Ensure the default globe doesn't render and the code doesn't display until it's loaded
- Ensure that after the _initial_ page load switching gallery demos is still snappy and doesn't take longer to load

# Author checklist

### DIFF
--- a/packages/sandcastle/src/Gallery/loadFromUrl.ts
+++ b/packages/sandcastle/src/Gallery/loadFromUrl.ts
@@ -43,7 +43,6 @@ const fromItem = async ({ getHtmlCode, getJsCode, title }: GalleryItem) => {
  * don't point to a sandcastle like those needed for oauth
  */
 export function urlSpecifiesSandcastle() {
-  console.log("check url", window.location);
   const searchParams = new URLSearchParams(window.location.search);
   return (
     searchParams.has("id") ||


packages/sandcastle/src/Gallery/loadFromUrl.ts
```diff
@@ -37,6 +37,23 @@ const fromItem = async ({ getHtmlCode, getJsCode, title }: GalleryItem) => {
   }
 };
 
+/**
+ * Check whether the current page's url indicates a specific sandcastle.
+ * This is required in case other search params are included in the url that
+ * don't point to a sandcastle like those needed for oauth
+ */
+export function urlSpecifiesSandcastle() {
+  console.log("check url", window.location);
```
mzschwartz5: Guessing this log statement isn't supposed to get merged?

mzschwartz5 (approved):
Approved but let's remove that one log statement before merging

packages/sandcastle/src/Gallery/loadFromUrl.ts
jjspace: Yes, extra from the oauth testing branch

Issue #12823, opened 2025-08-19
Scene/Cesium3DTileset `Restrict tileset memory usage with maximumCacheOverflowBytes` faulure


lukemckinstry:
### What happened?

The `Restrict tileset memory usage with maximumCacheOverflowBytes` test has been flakey in CI recently working on the `model-instance-v2` branch.

It will pass on the second or third try when I rerun CI via the GitHub UI.

https://github.com/CesiumGS/cesium/actions/runs/17078964943/job/48427639426?pr=12588

```
TOTAL: 2 FAILED, 14793 SUCCESS


1) verify statistics
     Scene/Cesium3DTileset
     Unhandled promise rejection: undefined thrown

2) Restrict tileset memory usage with maximumCacheOverflowBytes
     Scene/Cesium3DTileset
     Error: Timeout - function did not complete within 5000ms
    at poller (Specs/pollToPromise.js:26:11 <- Build/Specs/SpecList.js:3291:11)


TOTAL: 2 FAILED, 14793 SUCCESS
[18:55:12] 'test' errored after 1.21 min
[18:55:12] Error: 1
    at formatError (/home/runner/work/cesium/cesium/node_modules/gulp-cli/lib/versioned/^5.0.0/format-error.js:21:10)
    at Gulp.<anonymous> (/home/runner/work/cesium/cesium/node_modules/gulp-cli/lib/versioned/^5.0.0/log/events.js:33:54)
    at Gulp.emit (node:events:530:35)
    at Gulp.emit (node:domain:489:12)
    at Object.error (/home/runner/work/cesium/cesium/node_modules/undertaker/lib/helpers/createExtensions.js:61:10)
    at handler (/home/runner/work/cesium/cesium/node_modules/now-and-later/lib/map.js:68:20)
    at f (/home/runner/work/cesium/cesium/node_modules/once/once.js:25:25)
    at f (/home/runner/work/cesium/cesium/node_modules/once/once.js:25:25)
    at tryCatch (/home/runner/work/cesium/cesium/node_modules/async-done/index.js:21:15)
    at done (/home/runner/work/cesium/cesium/node_modules/async-done/index.js:37:12)
Error: Process completed with exit code 1.
```

### Reproduction steps

1.
2.
3.
...

### Environment

CI environment 
Browser:
CesiumJS Version:
Operating System:

jjspace:
The statistics one is long known: https://github.com/CesiumGS/cesium/issues/11958
[...]

PR #13061, opened 2025-11-27
Increase timeout to 10s for flaky 3D tileset spec

Beilinson:
Some of the tests in `Cesium3DTilesetSpec.js` have consistently been failing for a long time already due to a default timeout of 5s, this PR provides a simple fix by increasing the default jasmine timeout to 10s for each of the tests in this suite exclusively, and then resetting it to the original default.

# Author checklist


Beilinson:
example runs from #12630 and #12974 with this error:

image

javagl:
This has been ""tracked"" in https://github.com/CesiumGS/cesium/issues/12823 and https://github.com/CesiumGS/cesium/issues/11446 (note: one is closed, the other one open), with several PRs associated with them where this _happened_ or was supposed to be fixed. We considered that it might be a race condition (e.g. in [this issue](https://github.com/CesiumGS/cesium/issues/11958#issuecomment-2515555610), but also that it might just be a test that sometimes takes 4999ms and sometimes 5001ms.

(I also saw that failure occasionally, but ... who knows what's really happening there...? ... so just shrugged it off and restarted the test....)

When you say that increasing the timeout fixes it, then ... I'm a bit surprised: What _is_ actually taking 5 seconds there? And... of course, it's nearly impossible to _verify_ that this _consistently_ fixes the issue. But even if it only causes the issue to come up less frequently, that could be worth merging.

Beilinson:
Agreed that its difficult to verify that this even is the problem, thing is I never have encountered this test failing on either my personal laptop or pc from a timeout, meanwhile it fails on at least half of the PR CI runs I've submitted, leading me to think its genuinely a slightly slow test suite that fails sometimes in CI. I did try debugging the test locally and couldn't see anything that would cause this test to fail as often as it does in CI.

#11446 I see that @ggetz also suggested increasing the timeout in the poller, so I added that as well just in case

Beilinson:
I dont think will fix all of the issues with that suite however, for example https://github.com/CesiumGS/cesium/actions/runs/19678965074/job/56367761968 this run has a different test failing with `Unhandled promise rejection: undefined thrown` (same as https://github.com/CesiumGS/cesium/issues/11958#issuecomment-2515555610 that you mentioned)

Beilinson:
I think its worth it to merge just because this issue only appears in CI, if it goes away then great if not then this can be reverted and back to the drawing board 🙄

ggetz:
Thanks @Beilinson!

 I think we've addressed the `Unhandled promise rejection: undefined thrown` in [a separate fix](https://github.com/CesiumGS/cesium/pull/13405). This PR should still helpful for CI consistency though.
 
 I confirmed the specs take the same amount of time overall to run both with and without this change, so it should only come into play in exceptional cases if and when things take longer than expected. 👍

ggetz (approved):

Issue #13062, opened 2025-11-29
EntityCluster BUG


Duan971231:
When I use the EntityCluster of dataSource, using the following code and only replacing the version number will yield different results. The aggregation of the new version will directly return a single entity in the part outside the window. When dragging the camera, you will see a part of the single entity, which disappears after the aggregation starts. The effect is more pronounced in 2D view
`
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <link rel="icon" type="image/svg+xml" href="/dhs.jpeg" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />

    <link rel="stylesheet" href="/Cesium-1.135/Build/Cesium/Widgets/widgets.css" />
    <script type="text/javascript" src="/Cesium-1.135/Build/Cesium/Cesium.js"></script>
  </head>
  <body style="margin: 0; overflow: hidden; background: #fff; width: 100%; height: 100%; position: absolute; top: 0">
    <div id="map" style="margin: 0 auto; width: 100%; height: 100%"></div>

    
    <script type="text/javascript">
      Cesium.Ion.defaultAccessToken = undefined;
      const viewer = new Cesium.Viewer('map', {});
      viewer.scene.debugShowFramesPerSecond = true;
      //viewer.scene.globe.depthTestAgainstTerrain = true;

      class pointCluster {
        constructor(opt) {
          this.options = {
            dataOrUrl: '',
            pixelRange: 15,
            enabled: true,
            colorArr: [
              {
                num: 1,
                size: 48,
                color: '#e6a23cbb',
              },
            ],
            img: '',
          };
          this.options = Object.assign(this.options, opt);
          this.viewer = opt.viewer;
          this.dataSources = null;
          this.clustericon = {};
          if (this.options.dataOrUrl && this.options.dataOrUrl != '') {
            this.loadjson();
          }
        }
        loadjson() {
          var this_ = this;
          new Cesium.GeoJsonDataSource().load(this_.options.dataOrUrl).then((geoJsonDataSource) => {
            this_.showcluster(geoJsonDataSource);
          });
        }
        showcluster(geoJsonDataSource) {
          var this_ = this;
          this.dataSources = geoJsonDataSource;
          this.viewer.dataSources.add(this.dataSources);

          var pixelRange = this.options.pixelRange;
          var minimumClusterSize = 1;
          var enabled = this.options.enabled;
          //开启聚合
[...]

PR #13064, opened 2025-11-29
EntityCluster BUG Resolve the issue of incomplete aggregation.

Duan971231:
fixes: https://github.com/CesiumGS/cesium/issues/13062

Based on this issue, I have reviewed the source code of ElementCluster.js and found that using kdBush after upgrading from v3.0.0 to v4.0.2, continue use Uint32Array will result in negative numbers being represented as 2 ^ 32. When using index. range in the code, there will be no expected results.

If not filled in here, the default will still be Float64Array

Expected adoption

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 - Fixes jitter artifacts on Intel Arc GPUs [#12879](https://github.com/CesiumGS/cesium/issues/12879)
 - Fixed label sizing for some fonts and characters [#9767](https://github.com/CesiumGS/cesium/issues/9767)
 - Fixed a type error when accessing the ellipsoid of a viewer [#13123](https://github.com/CesiumGS/cesium/pull/13123)
+- Fixed a bug where entities have not been clustered correctly [#13064](https://github.com/CesiumGS/cesium/pull/13064)
 
 ## 1.137 - 2026-01-05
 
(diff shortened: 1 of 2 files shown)


javagl:
@lukemckinstry  (You did the initial triage on the issue) : This PR fixes the issue, which also affects the currently deployed 'Clustering' sandcastle.

Here is a comparison of the current `main` state, and the state of this PR:

Cesium Clustering PR 13064

The state of this PR is the **right** one, literally.

lukemckinstry (approved):
Thanks for this PR and apologies for the slow follow up. 
I see the improvement clearly in the example linked in https://github.com/CesiumGS/cesium/issues/13062.
I actually do not see a difference between the current prod clustering sandcastle and this branch (I am unable to reproduce the left and right states above), but I also do not see any regressions and clustering appears to be working as expected.

javagl:
> I actually do not see a difference between the current prod clustering sandcastle and this branch 

A **somewhat** reliable way for reproducing the issue in https://sandcastle.cesium.com/index.html?id=clustering is to zoom in a bit (to the USA), move the westcoast _just_ out of view, and then drag right, moving the westcoast back into the view. The newly appearing/visible cities/points will not be clustered (only when dragging even further).

lukemckinstry:
Thanks @javagl  that makes it very clear so I see the improvement now.
@Duan971231 are you able to add a changelog entry for this PR before we merge?

Duan971231:
@lukemckinstry Hello, which changelog entry should I change? the CHANGELOG.md in the code hasn't been updated for quite some time.

javagl:
The file to be modified is https://github.com/CesiumGS/cesium/blob/main/CHANGES.md - it's just about adding a line similar to the one that was added at https://github.com/CesiumGS/cesium/commit/6a6627fc842be2bc1f1288dfe7e727f638d204cd , saying something like

` - Fixed a bug where entities have not been clustered correctly [#13064](https://github.com/CesiumGS/cesium/pull/13064)`

(adjust the wording as you see fit)

Duan971231:
@javagl Okay, the code has been submitted

Issue #13072, opened 2025-12-03
Billboard subregion render error


plainheart:
### What happened?

After the fix #12958, the initial billboards with subregion work well when creating the BillboardCollection. However, the billboards added later to the collection fail to render with the expected subregion and instead display the full sprite image.

<img width="799" height="533" alt="Image" src="https://github.com/user-attachments/assets/57a38eba-13e9-4bd2-bbc2-6c1cce7e39f8" />

### Reproduction steps

1. Open the Sandcastle example link
2. Watch if all the markers are rendered with the expected subregion

### Sandcastle example

[...]

Issue #13016, opened 2025-11-05
3d gauss splatting 3dtiles lod render error


cuijianzhu:
### What happened?

<img width="2521" height="1272" alt="Image" src="https://github.com/user-attachments/assets/8a390977-2a74-4e2e-882b-5210d754a1eb" />
bug1:
Access to XMLHttpRequest at 'http://ecn.t3.tiles.virtualearth.net/tiles/a132103233203120321321.jpeg?n=z&g=15420' from origin 'http://127.0.0.1:5503' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
Cesium.js:7476 An error occurred in "op": Failed to obtain image tile X: 1749805 Y: 850614 Level: 20.
Cesium.js:75   GET http://ecn.t3.tiles.virtualearth.net/tiles/a132103233203120321321.jpeg?n=z&g=15420 net::ERR_FAILED 504 (Gateway Time-out)

<img width="1280" height="770" alt="Image" src="https://github.com/user-attachments/assets/ea109b34-e80a-4380-9799-9362718f4eb6" />
bug2:
256[.WebGL-0x4ce400ebf000] GL_INVALID_OPERATION: Error: 0x00000502, in ..\..\third_party\angle\src\libANGLE\renderer\d3d\VertexDataManager.cpp, rx::VertexDataManager::reserveSpaceForAttrib:535. Internal error: 0x00000502: Vertex buffer is not big enough for the draw call.
demo - 副本 - 副本.html:1  WebGL: too many errors, no more errors will be reported to the console for this context.

bug3:
demo - 副本 - 副本.html:1  Access to XMLHttpRequest at 'http://ecn.t1.tiles.virtualearth.net/tiles/a1321032332031203311.jpeg?n=z&g=15420' from origin 'http://127.0.0.1:5503' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
Cesium.js:7476 An error occurred in "op": Failed to obtain image tile X: 437455 Y: 212652 Level: 18.
Cesium.js:75   GET http://ecn.t1.tiles.virtualearth.net/tiles/a1321032332031203311.jpeg?n=z&g=15420 net::ERR_FAILED 504 (Gateway Time-out)

<img width="2535" height="1255" alt="Image" src="https://github.com/user-attachments/assets/205f8081-baaf-41a8-8b05-c5b49560020b" />
bug4:
part of the model is not rendered.

[...]

PR #13075, opened 2025-12-03
Fix image subRegion... again... #13072

Beilinson:
# Description

In #12958, I fixed subRegions for entity billboards. I however ended up missing that my implementation didn't fix (or even caused a new bug I'm not sure) when adding billboards directly using a `BillboardCollection`.

The issue stems from a change that was fundamental to fixing subRegions for entities: because the subRegion is set every frame, and this used to **always await the image index promise**, the subRegion ended up being in the `LOADING` state every frame. The fix was to allow the subRegion to resolve synchronously , and even better return synchronously if the image had also already resolved. 

However this causes an issue when using regular billboards, since those are instantiated once and the properties are set only once. If the image has already loaded, then the subRegion will synchronously run, and the image will resolve a frame after due to the `Promise.resolve + await` combo that is removed by this PR.

An alternative fix exists:

When creating a new `Billboard`, chain the `addImageSubRegion` call to the promise returned by `loadImage`:

```js
// Billboard.js line 203
if (defined(image)) {
    this._computeImageTextureProperties(options.imageId, image);
    const imagePromise = this._imageTexture.loadImage(
      this._imageId,
      image,
      this._imageWidth,
      this._imageHeight,
    );
    if (defined(options.imageSubRegion)) {
      imagePromise.then(() => this._imageTexture.addImageSubRegion(this._imageId, options.imageSubRegion));
    }
  }
```

This solution is less safe though in my opinion, if we are treating the image as ready synchronously within the `TextureAtlas`, as well as treating the `subRegion` in that way, we should continue this trend throughout the `BillboardTexture` code as well, rather than having `Billboard` assume that in a certain edge case `addImageSubRegion` will run before `loadImage` because of promise shenanigans.

## Issue number and link

#13072

## Testing plan

Test both the entity sandcastle and the provided sandcastle in the linked issue (it works locally).

# Author checklist


Beilinson:
Visual confirmation of both working:

Main sandcastle (entities):
image

BillboardCollection.add (plainhearts sandcastle):
image

packages/engine/Source/Scene/BillboardTexture.js
```diff
@@ -199,7 +199,10 @@ BillboardTexture.prototype.loadImage = async function (
   let index;
   const atlas = this._billboardCollection.textureAtlas;
   try {
-    index = await atlas.addImage(id, image, width, height);
+    index = atlas.addImage(id, image, width, height);
+    if (index instanceof Promise) {
+      index = await index;
+    }
```
mzschwartz5, line 205: Help me understand how this solves the issue? The issue occurs when returning a resolved promise, and you're saying that `await`ing that causes a one-frame delay (even though it's already resolved, microtask queue shenanigans?) that causes this bug?

mzschwartz5: It's just kinda annoying - I don't like relying on `instanceof`. It generally indicates something has been designed incorrectly (which is already definitely the case in terms of overall billboard code architecture / texture loading, but I was hoping we improved on that last time).

Beilinson: Yes that's exactly right

Beilinson: I'll do `typeof` instead like I did with `addImageSubRegion`

Beilinson: But i definitely agree with you regarding the billboard texture architecture. IMO the best option long term is to do what I suggested in https://github.com/CesiumGS/cesium/pull/12958#discussion_r2475441803, but as I mentioned there this will also fail for CallbackProperties so we still need some way of synchronously accessing already resolved textures + subregions.

mzschwartz5: I think that's basically the same thing, in terms of coding practice (maybe not performance though). I'm going to take a look around and see if anything makes more sense. (Like maybe rather than returning a not-promise here, we should be returning a resolved promise for the subregion index...? If we're not already? Idk just thinking out loud)

Beilinson: That is what we're doing (returning an already resolved promise) but yeah I'd also like to experiment with some other solutionsj

mzschwartz5: Right now we're returning an already resolved promise, and your change undoes that. What I'm suggesting is keeping this as is, and returning an already resolved promise for the subregion index as well.

Like [this](https://github.com/CesiumGS/cesium/pull/13076/files). Seems to work, right?

mzschwartz5: _And_, if this does solve the issue (without introducing new ones), then it also has the benefit of reducing the complexity of that function and consumers of it a little (no more `typeof` in `BillboardTexture.prototype.addImageSubRegion`). And it consistently returns a promise instead of a union type.

Beilinson: This recreates the original issues with entity + imageSubRegion, which was run every frame and every frame was reset to `LOADING` and couldn't ever get resolved. BTW my original solution in #12958 avoided this by only adding a sync check at the `BillboardVisualizer` level, so it wouldn't have affected the regular billboards while still fixing it for entities.

mzschwartz5: 😮‍💨 Gosh, that's annoying. Well, I guess the change here is at least consistent with what was already being done in `addSubregion`, even if I don't like it. 

Very uncharacteristically for me, I'm happy to just call it a day and say this works and maybe some day we'll improve the overall handling here. Do you still want to change `instanceof` to `typeof` or should we just merge as-is?

Beilinson: Alternatively the sync check could happen at the `Billboard/BillboardTexture` level. Point being dont start a promise every frame that always resolves on a different frame. 

Again though I'm not sure how much either of those work with callback properties, meaning there might always need to be a syncOrAsync path here for ideal rendering. From a JS perspective I think wrapping things in async/await even if they're available synchronously kind of sucks as its a performance hit no matter how you slice it.

mzschwartz5: On the bright side, this isn't much of a hot code path. It's not something that really runs every frame.

Beilinson: I'm going to sleep on this. I’ve been reading up on the ["Releasing Zalgo"](https://blog.izs.me/2013/08/designing-apis-for-asynchrony/) anti-pattern, which warns against APIs that return either a sync value or a promise depending something.

Since the recommendation is to be consistent (always sync or always async), and considering this function is called every frame for the entity api, returning a Promise here might just be the root cause.

Idea:
- Always return synchronously.
- Return undefined if the internal promise hasn't resolved yet.
- For non-entity billboards, we can simply trigger a check on subsequent ticks until a value is returned.

then basically all the areas of the code that currently run after `await` will just be refactored to run given the value has returned, otherwise retrigger a subsequent check on `frameState.afterRender` or something along those lines. Thoughts?

javagl: No really _constructive_ comments (except for reiterating how much I appreciate the thoroughness and thoughtfulness that you're putting into that). 

But I cannot pass the chance to chime in when there's a JavaScript/async rant 🙂 What bugs me: `async` is _viral_. You can design a _beautiful_, clean, efficient, slim, versatile API. And at some point, 20 steps down the call chain, you suddenly have to call some `async` function, and this `async` will bubble up, until the `main` function, so to speak, and pollute your whole API with a breaking change. (And when Zalgo is already mentioned, the step to [this meme](https://www.reddit.com/r/ProgrammerHumor/comments/1h4v9i3/dontyouhateitwhenthathappens/) is not so large...)

For me, the ugly consequence is: When you define an interface, you have to declare **all** functions as `async`, period. Not doing so will result in a breaking change sooner or later. 
(And the following is not as sarcastic as it may look at the first glance: I once considered to ""design"" a new language, called 'AsyncScript', which is just JavaScript with each function declaration being `async` by default, and each function call being `await`ed by default. Let the compiler and runtime sort out optimizations here - this is a syntactical level that a software engineer should not be concerned with).

All this is somewhat unrelated to fixing an existing bug, or handle an existing shortcoming or inconsistency in a large codebase. The combination of "promises/callbacks" and the _state management_  (something being "LOADED" or "READY" or whatnot) requires great care and thorough documentation. I hope that PRs like this lead to a gradual improvement here.

mzschwartz5: I have to say, a (large) part of me agrees viscerally with this sentiment ("`async` is viral"). But there's a smaller part of me that always thinks "I must just not know the proper design principles for async programming." I feel that I have a good grasp on clean code design in general, but this must require an entirely separate regime of thought.

Could be wrong, that's just the nagging feeling I get! I will say, when I last did some async stuff in the `Material` class, my strategy was - rather than letting `async` bubble up and pollute the whole stack - to simply not `await` the promise once it got to a level where it didn't matter whether it was synchronous or not. Then, that function (and its ancestors) do not have to be marked async.

javagl: > to simply not await the promise once it got to a level where it didn't matter whether it was synchronous or not. 

That _may_ be applicable in some cases. Particularly in CesiumJS, where many structures are revolving around the assumption that the application is running indefinitely, and ~"some `update`" is called every frame, and the promise just sets some `ready`-flag so that this `update` function can do some `if (ready) render(); else dont();`. 

I recently had/currently have to deal with that. A "(web) request" involves a promise, basically for ~"the tile content". This promise is never `await`ed. When it's done, it's done. And then, it just changes some state somewhere, so that something else happens in the next `update`. 

But this has two drawbacks:

1. it is **highly** prone for race conditions. One [recent example here](https://github.com/CesiumGS/cesium/issues/13016#issuecomment-3612037630). And eventually, most forms of ~"something flickering" are boiling down to that. Making sure that the promise is not re-generated and executed multiple times is tricky. And making sure that the state change that is caused by the promise is consistent with other state changes that happen elsewere (maybe even by other promises!) can be _really_ tricky. (Another reason why the state space should be as small as possible)
2. It's hard or impossible to test. In order to test the effect of the promise, it *has* to be `await`ed, eventually. Things like [`pollToPromise`](https://github.com/CesiumGS/cesium/blob/0a49108a5025feb5e2f6a420ccd58919e39061f3/Specs/pollToPromise.js) are just crutches for converting an un-awaited promise (that is hidden deeply in opaque structures, and that does some state change) back into an awaitable promise - namely, by polling for the desired state change. 

I occasionally advocated for enabling something like https://typescript-eslint.io/rules/no-floating-promises/ - I'm pretty sure that there are some dangling promises in CesiumJS that could turn out the reasons for bugs. Even if it's sometimes _intended_, I think that making that explicit (and rasing awareness) by having to insert some ESLint-'ignore' could make sense. 

And after all, I'm in the camp of "There are no asynchronous operations,". Whether or not something should be executed asynchronously has to be decided by the _caller_. Yes, you'd usually not block/wait for some web request. But that should rather be handled with something like `promise = makeAsync(thatWebRequest());` than the other way around.

Beilinson: I think it's worth revisiting what we discussed here in a later PR

I'll do some final testing tonight to make sure this PR doesn't introduce a new regression, if not I think it's worth it to merge as is @mzschwartz5

plainheart: Hi there!

Hope you don't mind me reaching out again about a long-pending issue that I truly think deserves more attention. The approach of using image subregion as sprites for billboard rendering is actually a really common and essential technique in our daily development workflows. It would mean a lot to us if this bug could be fully fixed in the next version.

@mzschwartz5 I really appreciate all the hard work your team puts into polishing the product, and thank you @Beilinson so much for your dedication and efforts! Looking forward to seeing this issue resolved soon.

mzschwartz5: Hey there, sorry for dropping the ball - I have a few other PRs to review today. I'll see if I can make some time for this one as well.

mzschwartz5 (approved):

PR #13070, opened 2025-12-02
fix(labels): Scale SDF glyphs to max label size

donmccurdy:
# Description

DRAFT - for discussion.

Currently label screen-space size is derived from `label.font x label.scale`, while SDF size (used to display the font in WebGL) is fixed at 48px:

https://github.com/CesiumGS/cesium/blob/aebb58ad06733fced57b2bc4a099c2ffd85f1540/packages/engine/Source/Scene/SDFSettings.js#L6-L13

The idea explored by this PR is to use SDFSettings.FONT_SIZE as a _minimum_, scaling up the glyph size to match the maximum font size present in the label collection. Increasing label font size should then never result in deteriorating quality. Note that `.scale` does not change the size of the SDF glyph, and could still magnify a low-resolution glyph, but I think that's acceptable.

Two big tradeoffs: 

1. **Texture atlas size.** Large font sizes require larger glyphs and a larger atlas. Using the max font size (not all requested font sizes) is a compromise so that only one font size is stored in the atlas at a time. I expect that label sizes larger than 48px are rare, so most users will continue to have current SDF size (48px) unchanged. That said, I'd be open to the argument that each unique size should be stored; in a collection with many small labels (and, say, 1000+ characters in some languages) and only 1–2 very large labels, it would reduce atlas size to store each label's glyphs at exactly the requested font size.

2. **Custom [`measureText()`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText) implementation performance.** We iterate over pixels to compute glyph bounding boxes, and becomes slower as glyph sizes increase. For example, a 30-character label at fontSize=400px spends 100ms in `rebindAllGlyphs` compared to 25ms at fontSize/48px. Perhaps this could be resolved separately by https://github.com/CesiumGS/cesium/issues/9767. I think it's only a minor issue at this stage.

Preview:

| fontSize=12 | fontSize=48 | fontSize=144 |
|---|---|---|
| fontSize=12 | fontSize=48 | fontSize=144 |

## Issue number and link

- Fixes #11377

## Testing plan

TODO

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [ ] I have performed a self-review of my code

#### PR Dependency Tree

[...]

Issue #12443, opened 2025-01-20
Label backgrounds have too much padding on Linux + Firefox


anne-gropler:
### What happened?

When working on a sandcastle for another issue (#12138), I noticed that the labels look odd on a different machine. I was told by a collague that this issue exists for at least a year, that we know of.

The labels have too much padding on Linux and Firefox. Also note the weird scattered single pixels in the background:

![Image](https://github.com/user-attachments/assets/afbdebca-1533-4f8d-8167-073b7243507a)

They do look fine ( = as expected) on the same Linux but with Chrome. Well fitting padding and clear white-with-alpha background.

![Image](https://github.com/user-attachments/assets/9e117115-b6d7-40a3-a769-68964e8b88fb)

### Reproduction steps

1. Select a machine running Linux.
2. Open Cesium Viewer in Firefox
3. add Labels (as entities or as label primitives) that use a background

### Sandcastle example

[...]

PR #2549, opened 2015-03-04
Font kerning support for LabelCollection

mramato:
I never thought I would be this excited about text rendering, let alone font kerning.

The recent forum discussion on right-to-left language support and unicode characters led me to create a branch, `single-billboard-labels`, which switches Entity labels to use a single billboard per-word rather than through `LabelCollection` with a letter-based atlas.  Obviously this is not a viable solution overall, but it allowed me to compare our label rendering against "perfect" labels that would be generated one word at a time. (All of these are smaller text scaled up, I've also sized the actual image the same to avoid GitHub scaling screwed things up, this is what you would see in Cesium with label scale set to 4 on a 2em sans-serif font)

Here's a label rendered in master:
image

And here's the same phrase, rendered with a single-billboard:
image

Notice how the spacing (i.e. kerning) of letters is all wrong, and we take up much more horizontal space than we should.  If you look closely, you may also notice that the `j` looks slightly cut off on the tip of the curve at the bottom.

In order to visually see what's going on, I added a random background color to each item drawn by `writeTextToCanvas`, here's what the alphabet looks like as a single billboard:
image

And here's what it looks like in master:
image

Notice that there is a ton of empty space in between letters, but that space isn't "empty" all of the billboards are touching and some even slightly overlapping (unintentionally I believe).  We can clearly see that in master the lowercase `j` is being cut off on the left side.

After seeing this, I realized that our goal should be to match the `single billboard` result by properly positioning the letters in relation to one another (a process called kerning for those of you unfamiliar with the term).  It turns out, the `dimensions` object returned by `writeTextToCanvas` actually has everything we need to do this, it just took a while to figure out.

Here are the results, drawn using `LabelCollection` but with proper kerning.

image

image

We now match the single billboard image almost exactly (better than I ever thought possible)!

[...]

PR #13081, opened 2025-12-04
refactor(labels): Remove custom measureText implementation

### LINKED ISSUE #9767: Investigate `writeTextToCanvas` and `measureText` function

### LINKED ISSUE #10649: writeTextToCanvas treats "-" hyphen/minus, or "_" underscore, characters as space in some cases.

### LINKED ISSUE #11705: writeTextToCanvas() background longer than rendered text

donmccurdy:
# Description

Revival of @javagl's #11747. One difference from the original PR, I've aimed to replace the custom measureText implementation with as little divergence from the current results as possible. So I've intentionally not tried to fix #10649 here. Along the same lines, I'm rounding some metrics to the nearest pixel to match the previous behavior. This helps with the test case shown below.

## Issue number and link

- Fixes #9767
- Fixes #11705

## Testing plan

Primarily, I've tested the PR by programmatically comparing output metrics with the previous implementation, for sample characters in English, Spanish, French, and Chinese Simplified alphabets:

```javascript
const MAX_ERR_PX = 1;

describe("Core/measureText", () => {
  let canvas;
  let ctx;

  beforeEach(() => {
    canvas = document.createElement("canvas");
    canvas.width = canvas.height = 1;
    ctx = canvas.getContext("2d", { willReadFrequently: true });
    document.body.appendChild(canvas);
  });

  afterEach(() => {
    document.body.removeChild(canvas);
  });

  it("ascii lowercase", () => {
    ctx.font = "12px sans-serif";
    for (const char of "abcdefghijklmnopqrstuvwxyz0123456789".split("")) {
      const m1 = measureText1(ctx, char, ctx.font, false, true, 1);
      const m2 = measureText2(ctx, char, ctx.font, false, true, 2);
      expect(Math.abs(m1.width - m2.width)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.height - m2.height)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.ascent - m2.ascent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.descent - m2.descent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.minx - m2.minx)).toBeLessThanOrEqual(MAX_ERR_PX);
    }
  });

  it("ascii uppercase", () => {
    ctx.font = "12px sans-serif";
    for (const char of "ABCDEFGHIJKLMNOPQRSTUVWXYZ".split("")) {
      const m1 = measureText1(ctx, char, ctx.font, false, true, 1);
      const m2 = measureText2(ctx, char, ctx.font, false, true, 2);
      expect(Math.abs(m1.width - m2.width)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.height - m2.height)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.ascent - m2.ascent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.descent - m2.descent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.minx - m2.minx)).toBeLessThanOrEqual(MAX_ERR_PX);
    }
  });

  it("chinese simplified", () => {
    ctx.font = "12px sans-serif";
    for (const char of "世界您好".split("")) {
      const m1 = measureText1(ctx, char, ctx.font, false, true, 1);
      const m2 = measureText2(ctx, char, ctx.font, false, true, 2);
      expect(Math.abs(m1.width - m2.width)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.height - m2.height)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.ascent - m2.ascent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.descent - m2.descent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.minx - m2.minx)).toBeLessThanOrEqual(MAX_ERR_PX);
    }
  });

  it("spanish", () => {
    ctx.font = "12px sans-serif";
    for (const char of "áéíóúüñ¿¡".split("")) {
      const m1 = measureText1(ctx, char, ctx.font, false, true, 1);
      const m2 = measureText2(ctx, char, ctx.font, false, true, 2);
      expect(Math.abs(m1.width - m2.width)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.height - m2.height)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.ascent - m2.ascent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.descent - m2.descent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.minx - m2.minx)).toBeLessThanOrEqual(MAX_ERR_PX);
    }
  });

  it("french", () => {
    ctx.font = "12px sans-serif";
    for (const char of "çéâêîôûàèìòùëïü".split("")) {
      const m1 = measureText1(ctx, char, ctx.font, false, true, 1);
      const m2 = measureText2(ctx, char, ctx.font, false, true, 2);
      expect(Math.abs(m1.width - m2.width)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.height - m2.height)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.ascent - m2.ascent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.descent - m2.descent)).toBeLessThanOrEqual(MAX_ERR_PX);
      expect(Math.abs(m1.minx - m2.minx)).toBeLessThanOrEqual(MAX_ERR_PX);
    }
  });
});
```

The purpose of the test is to confirm that no metric is off by more than 1px, for any character, compared to the previous implementation. The tests pass. Notably, if the `Math.round` calls are removed from the implementation, many of these tests begin to fail.

I also did some basic screenshot comparisons. Visual results are not identical, but — per the 1px target — are very close. 

**Labels**

| before | after |
|---|---|
| labels_before | labels_after |

Difference of a few pixels. Main concern here would be if the font baseline is misaligned, but both appear fine in tests so far.

**Sandbox from #11747**

| before | after |
|---|---|
| punctuation_prod | punctuation_round |

Some visible differences. Unscientifically, I think `writeTextToCanvas` is now more tightly fitting the canvas to the text when padding=0, which seems ... not bad? Based on #10649, the previous behavior might not be ideal, but I've tried to just minimize changes and avoid a regression in this PR so far, leaving a fix for future work.

**Sandbox from #11705**

| case | image |
|---|---|
| before | padding_prod |
| after | padding_round |

Improvement. The padding in the original screenshot was a bug, believed to be caused by our custom implementation not supporting some fonts correctly.

# Author checklist

#### PR Dependency Tree

* **PR #13081** 👈
  * **PR #13070**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 #### Fixes :wrench:
 
 - Fixes jitter artifacts on Intel Arc GPUs [#12879](https://github.com/CesiumGS/cesium/issues/12879)
+- Fixed label sizing for some fonts and characters [#9767](https://github.com/CesiumGS/cesium/issues/9767)
 
 ## 1.137 - 2026-01-05
 
--- a/packages/engine/Source/Core/writeTextToCanvas.js
+++ b/packages/engine/Source/Core/writeTextToCanvas.js
@@ -3,95 +3,41 @@ import Frozen from "./Frozen.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 
-function measureText(context2D, textString, font, stroke, fill) {
+/**
+ * Computes dimensions for text, based on current canvas state.
+ *
+ * Rounds metrics, excluding width, to whole pixels. This is purely to minimize
+ * rendering differences with migration to in-browser measureText(), and may be
+ * revised in the future. See: github.com/CesiumGS/cesium/pull/13081
+ */
+function measureText(context2D, textString) {
   const metrics = context2D.measureText(textString);
   const isSpace = !/\S/.test(textString);
 
-  if (!isSpace) {
-    const fontSize = document.defaultView
-      .getComputedStyle(context2D.canvas)
-      .getPropertyValue("font-size")
-      .replace("px", "");
-    const canvas = document.createElement("canvas");
-    const padding = 100;
-    const width = (metrics.width + padding) | 0;
-    const height = 3 * fontSize;
-    const baseline = height / 2;
-    canvas.width = width;
-    canvas.height = height;
-
-    const ctx = canvas.getContext("2d");
-    ctx.font = font;
-    ctx.fillStyle = "white";
-    ctx.fillRect(0, 0, canvas.width + 1, canvas.height + 1);
-
-    if (stroke) {
-      ctx.strokeStyle = "black";
-      ctx.lineWidth = context2D.lineWidth;
-      ctx.strokeText(textString, padding / 2, baseline);
-    }
-
-    if (fill) {
-      ctx.fillStyle = "black";
-      ctx.fillText(textString, padding / 2, baseline);
-    }
-
-    // Context image data has width * height * 4 elements, because
-    // each pixel's R, G, B and A are consecutive values in the array.
-    const pixelData = ctx.getImageData(0, 0, width, height).data;
-    const length = pixelData.length;
-    const width4 = width * 4;
-    let i, j;
-
-    let ascent, descent;
-    // Find the number of rows (from the top) until the first non-white pixel
-    for (i = 0; i < length; ++i) {
-      if (pixelData[i] !== 255) {
-        ascent = (i / width4) | 0;
-        break;
-      }
-    }
-
-    // Find the number of rows (from the bottom) until the first non-white pixel
-    for (i = length - 1; i >= 0; --i) {
-      if (pixelData[i] !== 255) {
-        descent = (i / width4) | 0;
-        break;
-      }
-    }
-
-    let minx = -1;
-    // For each column, for each row, check for first non-white pixel
-    for (i = 0; i < width && minx === -1; ++i) {
-      for (j = 0; j < height; ++j) {
-        const pixelIndex = i * 4 + j * width4;
-        if (
-          pixelData[pixelIndex] !== 255 ||
-          pixelData[pixelIndex + 1] !== 255 ||
-          pixelData[pixelIndex + 2] !== 255 ||
-          pixelData[pixelIndex + 3] !== 255
-        ) {
-          minx = i;
-          break;
-        }
-      }
-    }
-
+  if (isSpace) {
     return {
       width: metrics.width,
-      height: descent - ascent,
-      ascent: baseline - ascent,
-      descent: descent - baseline,
-      minx: minx - padding / 2,
+      height: 0,
+      ascent: 0,
+      descent: 0,
+      minx: 0,
     };
   }
 
+  // Baseline alignment requires `height = ascent + descent`. Rounding (if any)
+  // must be done before summing, or glyph pairs like "ij" may be misaligned.
+  const ascent = Math.round(metrics.actualBoundingBoxAscent);
+  const descent = Math.round(metrics.actualBoundingBoxDescent);
+
+  // Characters like "_" may have height <0.5 at some sizes, don't round to zero.
+  const height = Math.max(ascent + descent, 1);
+
   return {
     width: metrics.width,
-    height: 0,
-    ascent: 0,
-    descent: 0,
-    minx: 0,
+    height,
+    ascent,
+    descent,
+    minx: -Math.round(metrics.actualBoundingBoxLeft),
   };
 }
 
@@ -164,28 +110,36 @@ function writeTextToCanvas(text, options) {
   canvas.style.visibility = "hidden";
   document.body.appendChild(canvas);
 
-  const dimensions = measureText(context2D, text, font, stroke, fill);
-  // Set canvas.dimensions to be accessed in LabelCollection
+  const dimensions = measureText(context2D, text);
+
+  // Set canvas.dimensions to be accessed in LabelCollection. LabelCollection
+  // hard-codes strokeWidth=0, so dimensions should not include stroke padding.
   canvas.dimensions = dimensions;
 
   document.body.removeChild(canvas);
   canvas.style.visibility = "";
 
+  // measureText does not account for stroke width, added here. LabelCollection
+  // calls writeTextToCanvas with hard-coded strokeWidth=0, so stroke padding
+  // matters only when calling `writeTextToCanvas` directly.
+  const isSpace = !/\S/.test(text);
+  const strokePadding = stroke && !isSpace ? Math.ceil(strokeWidth / 2) : 0;
+  const doubleStrokePadding = strokePadding * 2;
+
   // Some characters, such as the letter j, have a non-zero starting position.
   // This value is used for kerning later, but we need to take it into account
   // now in order to draw the text completely on the canvas
-  const x = -dimensions.minx;
+  const x = -dimensions.minx + strokePadding;
 
   // Expand the width to include the starting position.
-  const width = Math.ceil(dimensions.width) + x + doublePadding;
+  const width = Math.ceil(dimensions.width) + x + doublePadding + strokePadding;
 
   // While the height of the letter is correct, we need to adjust
   // where we start drawing it so that letters like j and y properly dip
   // below the line.
 
-  const height = dimensions.height + doublePadding;
-  const baseline = height - dimensions.ascent + padding;
-  const y = height - baseline + doublePadding;
+  const height = dimensions.height + doublePadding + doubleStrokePadding;
+  const y = dimensions.ascent + padding + strokePadding;
 
   canvas.width = width;
   canvas.height = height;


donmccurdy:
@javagl would you mind reviewing this one? 🙏

javagl:
First getting rid of the custom implementation (and ensuring that this does not change anything else) sounds like a plan. Hopefully, the related issues (e.g. hyphen/underscore handling) can then still be resolved based on that state. 

I'll need a short refresher for the last experiments that I did there - that back and forth of `actualBoundingBoxDescent` vs. the font bounding box (https://github.com/CesiumGS/cesium/pull/11747#issuecomment-1908597562). And out of curiosity (and because I tore quite a few hairs out for that other PR), I'll probably have a look at whether the change here also causes the ~"sliiightly more jagged baseline for the veeeery small fonts" (that was one of the things that contributed to that PR not being merged back then). But maybe that was effectively caused by the missing `round`s. I'l try to do the review tomorrow (but not later than during the weekend).

javagl:
That `Cesium MeasureText` folder had literally been lingering on my Desktop for nearly two years now ("Maybe, one day, I'll have another look at this... 🤔"). 

Now, it has a `revival` subfolder.

I started some review/test, but may have to wrap it up later.

---

First, regarding the test that you posted: I wanted to try this out, and added the following as a spec file 

[measureTextTempSpec.js](https://github.com/user-attachments/files/23962104/measureTextTempSpec.js)

(Just copying both `measureText` versions on top of what you posted). 

Nearly all the tests are failing with

> Expected 2 to be less than or equal 1.

Before investigating it further: Is this how this (temporary, local) test is supposed to be run? 
If so, a wild guess: Could that failure be due to some ~"resolution scaling" or other browser/OS/hardware(!)-specific setting?

<sub>(I said that _nearly_ all the tests are failing. The 'spanish' ones are passing. I don't expect anyone to make this make sense...)</sub>

---

I ran a test similar to the one from https://github.com/CesiumGS/cesium/pull/11747#issuecomment-1906940405 , using this sandcastle
```javascript
viewer._cesiumWidget._creditContainer.style.display = "none";

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(-75.1641667, 39.9522222),
  label: {
    text: "Philadelphia",
    font: "12px Helvetica",
    fillColor: Cesium.Color.SKYBLUE,
    outlineColor: Cesium.Color.BLACK,
    outlineWidth: 2,
    style: Cesium.LabelStyle.FILL_AND_OUTLINE,
  },
});
```
And I wanted to post a GIF showing the differences. But there are none (at least, none that are visible even when looking closely)

---


I tried to check the effect of this PR on https://github.com/CesiumGS/cesium/issues/12443 , but the issue seems to be Linux-specific to begin with (I tried it on FireFox on Windows, and couldn't see these inflated labels here. Maybe @anne-gropler wants to try out this branch with the sandcastle from that PR?)


---

I re-ran the Sandcastle from https://github.com/CesiumGS/cesium/pull/11747 (with minor adjustments) - I'll post it as code here, the relevant points are summarized below,

```javascript
var viewer = new Cesium.Viewer("cesiumContainer");
viewer._cesiumWidget._creditContainer.style.display = "none";

const e = viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, 0, 100),
  billboard: {
    image: Cesium.writeTextToCanvas(
      "                top--                           ",
      { backgroundColor: Cesium.Color.RED }
    ),
    scale: 2,
  },
});

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00003, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("--'----", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00006, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("- ___- - - - '", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00009, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("__--__--__--", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

// This crashes on `main` with
// Error loading image for billboard: DeveloperError: Expected image.height to be greater than or equal to 1, actual value was 0
// but works now due to the 
// height: Math.max(Math.round(height), 1),
/*/
viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00012, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("_____", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});
//*/

// This crashes on `main` (see above)
/*/
viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00015, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("------", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});
//*/

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00018, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("Bottom Text", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00021, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("Right-to-left: Master (אדון): Hello תלמיד (student): שלום", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00024, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("Below this is a space only", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

// This (a single space) crashes on `main`
// AND on this branch. Consider changing
// isSpace: ... height: 0 
// to
// isSpace: ... height: 1 
// to avoid that
/*/
viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00027, 100),
  billboard: {
    image: Cesium.writeTextToCanvas(" ", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});
//*/

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00030, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("Above this is a space only", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});


viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00034, 100),
  billboard: {
    image: Cesium.writeTextToCanvas("꧁Javanese rerenggans꧂", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(0, -0.00038, 100),
  billboard: {
    image: Cesium.writeTextToCanvas(" À̴̴̴̴̴̖̖̖̖̖̀̀̀̀ n example for unicode combined characters", {
      backgroundColor: Cesium.Color.RED,
    }),
    scale: 2,
  },
});

viewer.flyTo(viewer.entities, {
  duration: 0
});
```

#### Relevant points:

The cases of `"_____"` and `"------"` are currently crashing on `main` with

> Error loading image for billboard: DeveloperError: Expected image.height to be greater than or equal to 1, actual value was 0

Both have ~"height of 0", apparently. 

They are **working** with this PR, due to that 
`height: Math.max(Math.round(height), 1),`

---

The case of `" "` (a single space) is crashing on `main` **and** on this PR, due to the special handling of the `isSpace`. Changing that special handling to...
```javascript
  if (isSpace) {
    return {
      width: metrics.width,
      height: 1, // ------------------------ ... return 1 here...
      ascent: 0,
      descent: 0,
      minx: 0,
    };
  }
```
... could prevent that crash. 

One could argue about what should be rendered there. (Nothing? A line? A rectangle?...). But I think that this is something that should at least be anticipated. In doubt, this could be some user input, and the application should not have to manually check this case to prevent a crash. Using `height: 1` seems to be a reasonable way to prevent that crash internally.

---

And... the ꧁Javanese rerenggans꧂are still working. 
I think we can all agree that **this** is the most important point here! 🙂

---

Again: I'll do the "final" review a bit later. I already looked at the code, and there doesn't seem to be anything controversial. 

The main "issue" fow now is about that temporary/local `Core/measureText` spec. I'd at least like to know how to run it, and why it's not passing for me.

donmccurdy:
Thanks @javagl!

> Nearly all the tests are failing with `Expected 2 to be less than or equal 1.` Before investigating it further: Is this how this (temporary, local) test is supposed to be run? If so, a wild guess: Could that failure be due to some ~"resolution scaling" or other browser/OS/hardware(!)-specific setting?

Interesting! I'll re-run on my machine just in case I've accidentally changed something, but it may indeed be a browser/OS/hardware difference. I'd originally run the tests in Chrome, on macOS, connected to an external display with resolution such that `window.devicePixelRatio = 2`. 

I'll try some other browsers and display settings to see what happens. If failures persist, perhaps we'll want to add a few more tests to confirm these are differences of "a pixel or two"  and not something that will add up to larger problems on multi-character labels.

javagl:
It's not perfectly clear how thoroughly things should be tested, and how deeply ~"unexpected behaviors" should be investigated. Most of the following is just a deep dive into the craziness of font rendering.

---

Logging my `devicePixelRatio` prints `1`. I had a short look at whether that could be overridden for some canvas that is only created for the specs, but ... maybe it doesn't matter after all.

---

I logged some of the outputs and checked the values, to see why the tests are failing. And I found this:
```
For  f
{width: 3.333984375, height: 8, ascent: 9, descent: -1, minx: -1}
{width: 3.333984375, height: 9, ascent: 9, descent: 0, minx: 1}
```
Sooo... the difference between the `minx` values is too large. 

This was the case for `f` (might be ligatures?), and for `j` (yeah, probably ligatures), and for `t` (wait ...what?), and for `w`, `x`, and `z` (!?)

For the Chinese characters, all tested ones pass, _except_ for `好`, which also fails due to that `minx`. 

So I commented out all the `minx` checks, and then, all tests passed...
...
...
... *except* for `Q`, which prints
```
For  Q
measureTextTempSpec.js:171 {width: 9.333984375, height: 8, ascent: 9, descent: -1, minx: 0}
measureTextTempSpec.js:172 {width: 9.333984375, height: 10, ascent: 9, descent: 1, minx: 0}
```
where the descent and the height are failing.

Easy to fix:
`if (labelText.includes("Q")) throw new DeveloperError("Tried to use Q in a label");` 🤪 

However, a comparsion of `main` and this branch with a label `aaaaQaaaa` looks like this:

CesiumJS MeasureText Revived

where the better one (which doesn't cut off the bottom row) is from **this** PR, so I'd consider this as a plus.

---

So there isn't any reason to _not_ approve this, **except** for that `isSpace` handling.

Do you think that ensuring a `height: 1` for the `isSpace` case makes sense, at least, to prevent the crash?

javagl:
> confirm these are differences of "a pixel or two" and not something that will add up to larger problems on multi-character labels.

This does not only apply to "multi-character labels", but also to the "small" test cases.

The tests until now had been restricted to `"12px sans-serif"`. I added the following "test case"...
```javascript
it("all of the above", () => {
    const families = ["sans-serif", "serif", "monospace"];
    const sizes = [6, 9, 12, 16, 24, 96];
    console.log(
      `family;size;char;m1.width;m1.height;m1.ascent;m1.descent;m1.minx;m2.width;m2.height;m2.ascent;m2.descent;m2.minx;absDiff(width);absDiff(height);absDiff(ascent);absDiff(descent);absDiff(minx)`,
    );
    for (const f of families) {
      for (const s of sizes) {
        ctx.font = `${s}px ${f}`;
        for (const char of "abcdefghijklmnopqrstuvwxyz0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ世界您好áéíóúüñ¿¡çéâêîôûàèìòùëïüß".split(
          "",
        )) {
          const m1 = measureText1(ctx, char, ctx.font, false, true, 1);
          const m2 = measureText2(ctx, char, ctx.font, false, true, 2);

          const d_width = Math.abs(m1.width - m2.width);
          const d_height = Math.abs(m1.height - m2.height);
          const d_ascent = Math.abs(m1.ascent - m2.ascent);
          const d_descent = Math.abs(m1.descent - m2.descent);
          const d_minx = Math.abs(m1.minx - m2.minx);

          console.log(
            `${f};${s}px;${char};${m1.width};${m1.height};${m1.ascent};${m1.descent};${m1.minx};${m2.width};${m2.height};${m2.ascent};${m2.descent};${m2.minx};${d_width};${d_height};${d_ascent};${d_descent};${d_minx}`,
          );
          //expect(Math.abs(m1.width - m2.width)).toBeLessThanOrEqual(MAX_ERR_PX);
          //expect(Math.abs(m1.height - m2.height)).toBeLessThanOrEqual(MAX_ERR_PX);
          //expect(Math.abs(m1.ascent - m2.ascent)).toBeLessThanOrEqual(MAX_ERR_PX);
          //expect(Math.abs(m1.descent - m2.descent)).toBeLessThanOrEqual(MAX_ERR_PX);
          //expect(Math.abs(m1.minx - m2.minx)).toBeLessThanOrEqual(MAX_ERR_PX);
        }
      }
    }
  });
```
Yeah, it's not really a "test case". It just dumps some data to the console. This data is attached here:

[stats2 2025-12-06.zip](https://github.com/user-attachments/files/23982614/stats2.2025-12-06.zip)

The differences increase with the font size. That may be expected. For example, for `sans-serif`, `24px`, char `f`, the difference of the `minx` is `6`. But for `serif`, `96px`, char `您`, the difference in `height` is `58`(!). This looks like something that could have ~"undersired effects". Comparing `main` to this branch with 
```
viewer._cesiumWidget._creditContainer.style.display = "none";

viewer.entities.add({
  position: Cesium.Cartesian3.fromDegrees(-75.1641667, 39.9522222),
  label: {
    text: "您",
    font: "96px serif",
    fillColor: Cesium.Color.SKYBLUE,
    outlineColor: Cesium.Color.BLACK,
    outlineWidth: 2,
    style: Cesium.LabelStyle.FILL_AND_OUTLINE,
    showBackground: true,
    backgroundColor: Cesium.Color.RED,
  },
});
``` 
does not show a significant difference. I have not yet investigated why that is - maybe none of the callers of that `measureText` function actually _used_ the original `height` (which seems to be waaay off here). Maybe they all already twiddled with `ascent` and `descent` (maybe even already as a "workaround" to cope with the "wrong heights" - who knows...).

Another "dimension of things that may have to be tested" is the handling of `outlineWidth` and `style`. These parameters are no longer passed to the function, which _looks_ like it could be a big deal.

But... the room for tests here is nearly infinite. 

For me, the `isSpace` question is still the only one that prevents this from being merged.

donmccurdy:
> The case of " " (a single space) is crashing on main and on this PR ... Do you think that ensuring a height: 1 for the isSpace case makes sense, at least, to prevent the crash? ... For me, the isSpace question is still the only one that prevents this from being merged.

Currently writeTextToCanvas has these special cases...

```
if (!defined(text)) {
  throw new DeveloperError("text is required.");
}
//>>includeEnd('debug');
if (text === "") {
  return undefined;
}
```

... and then LabelCollection.js does some special-casing of whitespace-only labels, but not before 'drawing' the whitespace to the canvas here.

My feeling is that the current hard-coded dimensions for whitespace labels in `measureText` might be a low-level workaround for what would better be a higher-level behavior decision in `writeTextToCanvas` or LabelCollection. Or arguably `writeTextToCanvas` could be an internal function, but that's a breaking change.

I'm inclined leave the whitespace handling as it is, for now. Mostly out of fear of exchanging the current error for silently breaking layout on more legitimate labels that just happen to contain whitespace.

> Logging my devicePixelRatio prints 1. I had a short look at whether that could be overridden for some canvas that is only created for the specs, but ... maybe it doesn't matter after all.

I don't think this can be changed by JavaScript, but when creating a custom device emulation profile you can specify the devicePixelRatio:

Screenshot 2025-12-09 at 1 50 03 PM

Likely this would change the test results, but IMHO not in a way that is relevant to the behavior under test.

donmccurdy:
> The cases of "_____" and "------" are currently crashing on main ... they are working with this PR...

Does [this link](https://sandcastle.cesium.com/index.html#c=zZbfi9NAEMf/lSEvl0Kz/aWitVfUnuCDIpxFXwJlm0zT5Ta7ZXeT2jv6PwjHPQj+eNZ/KP+Ok+Za5Tw5aHvSaVoys/OdzmfZkMm5gVzgHA0cg8I5DNCKLGXvVzE/9KKVP9DKcaHQhF7taagqBRtVix9EnKAjz2As3CaVWbeQyGJhZ5IvqHzoKa0w9KjApgQqJ5xAy3gc+xehAphpSxGtuutWBtw4uuOqwyZGpyeYGETrN+vQrNVLxVhIOdbcxF1YVQAQKU9wU2BuhMMhfnRDPeAq59avsoA6ghvm9CwI4N8WevW1+ALGPDpLjM5UPNBSm98dlx47fXkCyyq5di2yEZfUV3vlLul3WdvbZgRN1iTrbL0nXhAcBWSEuNbA3YjXZMu7ECOtrAOkY7BH2Ec7wMJoNAqg+hzdD/MeSZ9sT0qYwfp76Jyt9i6cZAcP+HCX5/P+Hs89Ej7envCFdk6nUEYPHbPd2h7zVCRTFzgdSJy4Lrzh1tH71y8+FZfFVfG91oVXKKWG4mfxpfhWfC4uwbcui6ltWit+UPSq+HrwO/Rgh4OAUs/BTYUFujjYGY8QtJKLQ6fubD+UeM/HOsf/Tz2Ri6H2sYx5da+3Gtv66z97JtKZNg4yI33GGg5TmuZoBxrjLDqjqS+ythSWqb3Gn9JeLHIQ8fEtIyREkltLK5NMynfinGbCfq9B+X9JpeaxUMnbHA3NkGXatNV/XQUZY70GubcrndZyzM2Nyr8A) crash on `main` for you? It's working for me. Tested Chrome, Safari, and Firefox on macOS. Interesting that this would be different, but it sounds like it's OK after the PR in any case.

donmccurdy:
While it does contribute to passing the proposed unit test, I'm worried that `minx: Math.round(Math.abs(metrics.actualBoundingBoxLeft))` with Math.abs cannot really be correct.

The custom implementation was unable to detect minx values outside the (hard-coded) padding, and perhaps wasn't _quite_ measuring the same thing, I think. So there's a subtle change here. I am tempted to remove the `abs()` call, and negate the result to match what I think would be the intended 'minx' meaning. I'll try that and see what happens for unit tests on Chrome, Safari, and Firefox. The character "j" seems to be a useful one to test with the minx property.

- [x] TODO: Test without using `Math.abs` to compute minx

javagl:
> Mostly out of fear of exchanging the current error for silently breaking layout on more legitimate labels that just happen to contain whitespace.

It's hard to imagine _how_ it could break the layout, but not knowing all details of that glyph-based atlassing, it's hard to be sure. (I just saw that there are (surprisingly many) [whitespace characters](https://en.wikipedia.org/wiki/Whitespace_character), and one could go the extra mile an try to run ~"tests with all of them", but... I think that it would still be fair to categorize them into 'space' and 'all the others'...)

> Does `[this link]` crash on main for you?

Yes! Both Chrome and Firefox (on Windows (with a GeForce 2080 (with a device pixel ratio of 1 (on a Tuesday)))) are crashing with
```
Error loading image for billboard: DeveloperError: Expected image.height to be greater than or equal to 1, actual value was 0
Error
    at new DeveloperError (https://sandcastle.cesium.com/js/engine/index.js:7458:11)
    at Check.typeOf.number.greaterThanOrEquals (https://sandcastle.cesium.com/js/engine/index.js:7545:11)
    at TexturePacker.pack (https://sandcastle.cesium.com/js/engine/index.js:41348:31)
    at TextureAtlas._resize (https://sandcastle.cesium.com/js/engine/index.js:41687:42)
    at TextureAtlas._processImageQueue (https://sandcastle.cesium.com/js/engine/index.js:41771:14)
    at TextureAtlas.update (https://sandcastle.cesium.com/js/engine/index.js:41818:15)
    at Array.<anonymous> (https://sandcastle.cesium.com/js/engine/index.js:125377:25)
    at callAfterRenderFunctions (https://sandcastle.cesium.com/js/engine/index.js:247583:48)
    at Scene4.render (https://sandcastle.cesium.com/js/engine/index.js:247931:3)
    at CesiumWidget.render (https://sandcastle.cesium.com/js/engine/index.js:251321:17)
``` 
This is _somehow_ not unexpected, with the remaining mystery of where a non-0-value is coming from on your system. A wild guess: These lines/characters have a ~"height of 0.8", and with some device pixel ratio based scaling of 2.0 this becomes `int(1.6)=1` in one case, and `int(0.8)=0` in my case. 

> The custom implementation was unable to detect minx values outside the (hard-coded) padding

I'd have to re-read the details, but from a quick look, there's some padding of 100, which should be enough...?

javagl:
(My _actual_ review here is still pending - until now, my comments haven't really been a "review". Looking at the ✔️ , it looks like this is ready for another pass? I might not be able to do this today/tomorrow, but should be able to do it soon (in doubt, during the weekend)).

donmccurdy:
I've just pushed a few fixes, and some comments/clarifications.

***

One bit I'd missed is that LabelCollection passes a [hard-coded](https://github.com/CesiumGS/cesium/blob/8e4375eac0506d2c77ade7d354790c176973e7a8/packages/engine/Source/Scene/LabelCollection.js#L199-L206) `strokeWidth = 0` into `writeTextToCanvas`, since it handles the outlines separately when drawing SDFs. That's fine, and just means LabelCollection strokes/outlines are isolated from these changes.

But for users directly calling `writeTextToCanvas` (since it's a public API) I've added a `strokePadding` offset computed as 1/2 stroke width, which is a clear improvement compared to production here:

```javascript
viewer.entities.add({
  position: Cesium.Cartesian3.fromDegreesArrayHeights([
      138.604193,  -34.921922, 10 ])[0],
    billboard: {
      image: Cesium.writeTextToCanvas("This is my text with some padding!?.; 123", {
        font: "28pt monospace, sans-serif",
        backgroundColor: Cesium.Color.BLUE,
        stroke: true,
        strokeWidth: 50,
      }),
  },
});
```

Before:
stroke_padding_before

After:
stroke_padding_after

***

Characters `ij` were misaligned in earlier versions of this PR; that has been fixed. I created a small debug viewer to check the metrics against the rendered characters, compared for some common and/or troublesome characters here:

2025-01-06 Text Metrics

Red = production, green = PR. Horizontal yellow line is the baseline. The red/green outlines are supposed to tightly fit the text left/top/bottom edges; extra space on the right is expected as we aren't using `metrics.actualBoundingBoxRight`.

Results in Chrome seem good: visibly better alignment on the left and bottom edges, characters "i", "j", and "2" no longer clipped at the bottom of the box.

Safari and Firefox show the same improvement in the vertical fit, but have some unexpected space on the left coming from the `minx` value. In Firefox it's minimal, in Safari it's a bit more of an issue and could presumably affect kerning. The most practical way I could see to test that was with [kerning pairs](https://typeheist.co/blog/sample-text-for-typeface-character-testing/):

```javascript

        name: "label3",
        position: Cesium.Cartesian3.fromDegrees(103, 36.0001, 0),
        label: {
            style: Cesium.LabelStyle.FILL,
            text:`
Av Aw Ay Az Fa Fe Fi Fo Fu Kv Kw Ky Pa Pe Po
Ta Te Ti To Tr Ts Tu Ty Va Ve Vo Vr Vu Vy Wa We Wr Wv Wy
AC AT AVA AWA AYA AV AW AY AZ CT CYC
FA FE FO KV KW KY LO LV LY NG OD PA PA PE PO
TA TA TE TI TO TR TS TU TY UA VA VA VE VO VR VU VY
WA WO WA WE WR WV WY YS`.trim(),
            fillColor: Cesium.Color.WHITE,
            horizontalOrigin: Cesium.HorizontalOrigin.LEFT,
            verticalOrigin: Cesium.VerticalOrigin.BOTTOM
        }
});

```

... which shows no particular change compared to the previous implementation, except that in Safari only, some lines are offset to the right by 1-2px. That's the line starting point, not the spacing between characters. I believe that's minor enough to not block merging, but should be reported to Webkit as a bug since it differs from the other two browsers noticeably.

***

Finally I did some testing on the "Clustering" sandcastle. It looks mostly unchanged, except that the bottoms of some characters were previously cropped, and aren't now. Subtle but I believe it's an improvement. See bottom of the "5" here:

| before | after |
|---|---|
| chrome_pin_crop_before | chrome_pin_crop_after |

***

With the latest changes, the PR is no longer following the "MAX_ERR_PX = 1" numerical target I'd set in the original description, but I think the changes are for the better, e.g. to ensure baseline alignment of characters within a word.

@javagl ready for a review pass when you have the chance, yes, thanks!

javagl:
As mentioned in the first comments: This already _was_ an improvement for many cases, and I didn't find a case where it made things objectively worse. 

The main points that caused some churn:

- The single-pixel offset test failures revolving around `minx`
- The open questions about strokes/outlines

These seem to be addressed. 

I will not perform _all_ tests that _could_ be done here. But I tried to cover what I, subjectively, perceive to be the most common and most important points.

--- 

To be clear: I assume that this change _will_ cause regressions. People have already found "solutions" for things that didn't work as _they (subjectively) expected_, but there was no concept of "right" and "wrong" in terms of a "specification". (C.f. [xkcd "Workflow"](https://xkcd.com/1172/) and the [Amobea Model](https://wiki.apidesign.org/wiki/Amoeba_Model)). We can only try to minimize possible side-effects and regressions.

---

I'll dump the current state of my "test sandcastle" here, even though I've been twiddling with it ad-hoc, and it's not a "clean" test

https://sandcastle.cesium.com/index.html#c=1Vjbbhs3EP2VqV6yKrSri+PUURy3tqy2SVQHsJ0GBQQo3N2RljBFqiRXshIYaPMFBYIUKNDLY9o+9zP8rj/oQ76is7uSKjl2bHTlwKUuEMmZEefMzCG5Q6ZhyHGEGu6DxBE00PC4732djjntQpD2G0paxiXqdqF4ry0zDa+TTT7lYQ8t9TSG3M5FPWPHAr2Qm4FgYzLfLkglsV0gA20ZKGksIA1PjaG03HI0HgtD50VbAgyUoREl67NFNZi29IvJNa+rVX8XexrROJUS0LtaqRRLiZrPhfAV02EdUjMAvM96OLcy0tziIR7bQ9VgcsiMk0kBLRDONKsGrgsXt3ahNFN+AT4LjnpaxTJsKKH0v8tOet5+cxdOMuHiVMkETNC6amn3hL5Piik2q0DErXgVamv5gCm47i2XGvk504HL/Zy6d/IB/byT10/odDouZK9bN97duzndJV/d2ed6nC2X4TDiBgLNTIQGlIRnfeKFZzDiNkrnm1orDUKxkMtetmjo0siCQ7s4RKEGqFPZOjSPBxhYDDNpL0LeiyxVKfgIhBGzxGM2YhLIDH4bM5HMVUvAApt0hkzECCNmoJKuwI8tjJQ+MiDVCEKaI3EbIaSzmfU6fMVs5PXZsZP+SMFxsrkihSEBovxxeZXhrdZyh5fadcS1nHr6vug6BhGYr4ZYXD0u67nZ7LrIbAbMCp3dyOnsjrJW9SEZvel8Vqvm9HU/qUbXKldgN61YkzCBM/lh8mryevJbsQ5fohAKJn9Ofp78Ovlp8ooS1cYhrZ3mJn/Q6OvJLzceptt5U4LIdEQER5VLbwZmwAKk4hXj690EHPor4niB2T8W3+WNVHZ7bzcZSdfnayaDyAM6ThoeUjADovUeGUkluTlIDNXB87w5T1cy2rbqYpHqTATYUPEw2Svsymmq9knOMMH/g6HWcp66C9vJNvEB03Glzuetxbdvvn/IhkyiQdCoUfZ6TJq3b17edBZay7sxwfbpd6d/nf540TdIwGPWH4jsOBhLHqgQIVB9n+6VYcIEmg50qM1Nh+p23oPcDuzRVja+AmKUTrQUS/nz0afePajW1pbA6dKtvE68UtsYWOgrqdJaK4GhnHMNat5duMpeFUmC0Gp1RC5YHeOZwac8tFEd1ivLsC/gnD0CEOTgip4CTEFfvAsK5qOYA24J3QSFVj5Yp+kTqdHOHKklDFRsBaXqeeh9sd9s7i2LTZHamNlNHpvMtVqJAwfpk5TPH7RaHdojO4+fHLYe7DWX8Uxw9FJ3vUW7BO3GyslvvXIpxH+//H2O1Cz57t4ZHMNysnWpKM6D6eDRNzutJ83L8dxpbTcenYtn7T/h+d64Xq0wlnacOexdMT5UzpkYTCs0jDXLAlCZqhZKhc104VvJ/Ge8P1DaQqyF43lli0SOdNc1ZT8OjtB6gTGJ0mZ5prIZ8iHw8P45z+8goEwxNNONhTjgz7Fd2Nosk/yS2vRW/niIWrBxIhJVt1rZIB2oNsvUfVfLKiV8phcs/gM

(Link to the deployed one here)

In all tested cases, the result is the same _or better_.

---

Notes that are not strictly related to the PR:

> LabelCollection passes a hard-coded strokeWidth = 0 into writeTextToCanvas

There is one top-level point that is a bit nebulous for me: What is the difference between a "Billboard" and a "Label"? Sure, there are differences, but it's hard to pinpoint them, there seems to be some overlap, and it's a long, winding path from [the corresponding sandcastle](http://localhost:8080/Apps/Sandcastle2/#c=pZLbattAEIZfZdBNFDArOe2Vo5hSpZBCoNCa5EYQRtqxsmS1a3ZXctzid8/IshXnAAkUxKI5/D/zDdOhg07RmhxcgKE15ORV24ibXS4uomoX59YEVIZcEZ2eF2ZQiLuheKtkTYEjR1KFsVX4sNEkpPIrjRu2LyJjDRURG4wWZIIKirxAKeN/hQFYWc8Za2aHUXJ0gf/QfBFLZ5tLqh2Rj9MJ8DdN0/R00utKpXVp0ckZ7HwAVIM1jTZrpwIt6DEsbI6mQ89wP08awGdlEU0OWoAlc8x45rOvq0e4It1RUBVyy6GhxOqhdrY1Mrfauud5+0j8/nG579wO82353R4t7//IxfSYXWNJeuQODNlPPtBd97Vx7A+o/L1dfx+5ZhBcS/vS53BfUy71ZmHjV8j7LcvW4cCb7kXRJMp2VzPv699Us7IuQOt0LEQSqOFD4oUkZVs98MFV3veiLDlIMqk6UPLinauFSqP3XFm2Wv9Rf/kM51nC/S9k2qJUpv7VkeOT7Vvup/PrISmEyBIO36qCtbpEd+T4BA) to the final pixels on the screen. 

The following is a screenshot from a [sandcastle](https://sandcastle.cesium.com/index.html#c=vVNdawIxEPwryz1psTmtVsGeUntaEA4ttR8vBxLvVg2NiSQ5rS3+9+ZOrYp9q/gSspOZ2c3AstlcKgNXQDX4qFkyg7GSMwidKKtC5y4UbEMaUBFHVBuOO84eyXihiKTQBhYMl6igAQKXW1fylmG5na8vhaFMoAqd/IGS0xHyllVuLAgKwwxDTWgc575DATCX2iJS1HfGPlXG3qgok3SqNk4Uos5d125JqVopVau1AlSKpJgvpPKsQR0yKwCDn6Zu//E0ZZzGyOdTRkOnsHmUieF2Ql9yqfbd0oo8d9rHpHcWm2kdSrdbWJsVx19RkDYdpBB57AbBsNVrD/uvL0G318n4a3uuT2J4OHsM5UvFUD1jDP7ZY7i5VAy1/8XgFBwvkzZT+H67hYniOUJcg7M5p/av7iiJPtCQSOtU5Lk7iRezBbC48cfKQcSp1vZlnHA+YF92fZuea/lHMi5pzMSkv0DF6SqlTEvNYAMSQjzXlqcqIyUfUXXg+AM) with three labels:

Cesium Labels Outline

The `outlineWidth` of these labels is (from top to bottom): 15, 16, and 17 (!). So... something's wrong there. Probably. 

Again: Unrelated to this PR!

The "Before/After" pictures that you posted about that are in line with what I'm seeing. At least in terms of the core of the PR, namely, the text measurement. The appearance of this on my system is a bit different...

Cesium Outlines Again

But that's even more unrelated to this PR. 

---

> I created a small debug viewer 

That's interesting. I wouldn't have expected these differences between browsers, but it underlines that some pixel-level details are basically out of the control of CesiumJS to begin with.

> could presumably affect kerning. 

I don't see a relevant difference between the before- and after state. Details may depend on the environment (Browser/OS/Graphics card/Driver/Weather...). But as a fun fact: There once was some discussion around keming. You can find it by searching for "keming". That single pixel is important. (Nearly as important as for the "pom.xml" that is used for Maven). One example is (edit: fixed link) https://github.com/CesiumGS/cesium/pull/2549 , and some of the screenshots there resemble what is shown in your "debug viewer". One could do quite some archeology here...

---

> Finally I did some testing on the "Clustering" sandcastle.

I didn't see the "cut-off bottom" there - but that may also depend on the environment. More importantly: The screenshots from _your_ environment look like an improvement, and in _my_ environment, there is no difference, and that was one ""goal"" here after all.

donmccurdy:
> The outlineWidth of these labels is (from top to bottom): 15, 16, and 17 (!). So... something's wrong there. Probably.

Oof, I had seen some cases where the outline widths didn't seem ... quite right ... but that example is wild. As you said, unrelated, perhaps something later in the SDF rendering. I hadn't found an existing bug about this yet.

I resolved the merge conflict, and I think this is ready to go. I'll still plan to file a bug on Webkit with the TextMetrics / minx discrepancy, but I don't think that needs to block the PR.

donmccurdy:
Bug reported to Webkit at https://bugs.webkit.org/show_bug.cgi?id=305132, along with a minimal reproduction including the debug/visualization tool mentioned above.

Issue #12949, opened 2025-10-03
Label misalignment/regression after upgrading from 1.132.0 to 1.134.0


912807862:
### What happened?

After upgrading CesiumJS from 1.132.0 to 1.134.0, labels with small fonts or mixed characters (English + Chinese) are misaligned.
The same code works fine in 1.132.0, but in 1.134.0 the glyphs are vertically shifted or overlap incorrectly.

this is 1.134.0 result

<img width="833" height="404" alt="Image" src="https://github.com/user-attachments/assets/7663adde-7ce8-4e7e-8e6a-0f7632582f36" />

<img width="576" height="541" alt="Image" src="https://github.com/user-attachments/assets/28dc150b-799b-4008-af66-88f34749d4e7" />

this is 1.132.0 result

![Image](https://github.com/user-attachments/assets/4125f582-dd7b-4156-a0d3-521d60e9d8e6)

<img width="616" height="533" alt="Image" src="https://github.com/user-attachments/assets/d03723e8-a560-45af-beec-584100e34a13" />

javagl:
Can you provide a Sandcastle that demonstrates the issue? (It's hard for many developers to even identify and write the characters in the screenshots, and they don't seem to contain mixed text at all - in addition to that, it might be necessary to be clearer about what is a "small font" in this case).

hassomehide:
[...]

PR #13086, opened 2025-12-05
Fixes stale label state on unbind

mzschwartz5:
# Description

In #12905, I updated `Label` to not clamp its individual glyph billboards, but to instead use the parent's position. This was a performance optimization, since clamping is expensive and not strictly necessary on individual label glyphs.

This change introduced a regression, as (it turns out) when labels are deleted, the underlying glyph billboards are not deleted, but stored (in `_spareBillboards` on the `LabelCollection`) and reused when new labels are created. Certain fields are not reset before being reused (namely, in this case, `_clampedPosition`), which caused some misplaced labels in workflows that delete and recreate clamped labels.

The fix simply resets the field before storing the label's glyph billboard in `_spareBillboards`

## Issue number and link

#12949

## Testing plan

Zoom in, then out (perhaps a few times, find the right zoom level) on [this sandcastle](https://sandcastle.cesium.com/#c=dVRRb9owEP4rtzxUyYpMWfdEAxuilTqpWyeKtpdIlUkOsOrYyHZg3cp/nx3HkLQ0D/H5ct933/nOYeVGKgMfgWqYomZVCUslS8iivN5l0VUmMpFLoQ1sGe5QwQgE7ppo8qv2xSF+KoWhTKDKoh78ywSAQaWsZxgAc78nLs1vqXjROOIkE/ukla2ghj7ISuXYzTittJHl9eGrza1rI4s8nO4oC2LJkUUTWhTxce+CORrYKNwyWelbZKu1qRkabM5l/kSkmDO7WPDNFoW5Y9qgLTCOExiNfY1esNXoOazewEBLVJRspGaGSTGlysiVops1y8k65ANgS4jdCvChqwZeXrw/fuVPYXBRP3B21ko8Du7kXeT4NPJAmDhY4uvylUmOhMuVPecZForu/DmHrwbKihu24SyMxptEX+ATDGHQgI4dIPY47bnYzigs5RYnnMeBeikVxK49XApLe3FVGylcNtaofp/DZzu7RwEH3W0CahoCagJB7XLvcyvtHYLTSt0MtUIAQm8PA+6abC0qLusZv8aVQtSxldtzKZNeG83pAvkQOozuzvwxw6bALBpmkV0ttNeN8gM0wyUqFDkeBNx2/WR6N/n+83F+/zi/mc0m3350aPat3T4cvjW90SyvRqjV5Rqwb25u1ItSbZ45jp33K/P/lkrxmJC+wXJja0DdX1T5ExqSa+1AaT9A0oJtgRWjE/8SyDnV2n5ZVpw/sL/2ro/Tvo3vwLikBROr+y0qTp9dyHowvvNOQkjat9u3KCMlX1DVYvwP). Note how certain labels start to overlap.

Try the same on this sandcastle (deployed version from this PR). It should no longer occur.

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## 1.137 - 2026-01-01
+
+### @cesium/engine
+
+#### Fixes :wrench:
+
+- Fixes label positioning in workflows that delete and recreate clamped labels [#12949](https://github.com/CesiumGS/cesium/issues/12949)
+
 ## 1.136 - 2025-12-01
 
 ### @cesium/engine


packages/engine/Source/Scene/LabelCollection.js
```diff
@@ -118,6 +118,7 @@ function unbindGlyphBillboard(labelCollection, glyph) {
   const billboard = glyph.billboard;
   if (defined(billboard)) {
     billboard.show = false;
+    billboard._clampedPosition = undefined;
```
donmccurdy, line 121: Optional and non-blocking thought, especially since it predates this PR, but — I wonder if the Billboard class would benefit from having a public `.reset()`, `.remove()`, or `.unbind()` method? Seems like we're having to manage Billboard's private members here in LabelCollection, maybe that could be internal to Billboard.

mzschwartz5: I totally agree, I will admit I was feeling kind of lazy here 😅 but the thought crossed my mind.

I'm on another bug fix at the moment, but will see if I can get to that today. Otherwise, if we end up merging, I'll open an issue (could be an onboarding issue even).

javagl: If I had to summarize my thoughts about accessing private properties from other classes _in general_, then that summary would be a (somewhat confused) "Don't 🤷‍♂️".

One could make some cases about `Billboard` and `BillboardCollection` being kinds of "(C++) `friend`" classes, or certain properties being considered as "(Java) 'package-private'". But some classes seem to be [friend's friend's friend's friend's friends](https://github.com/CesiumGS/cesium/blob/0a49108a5025feb5e2f6a420ccd58919e39061f3/packages/engine/Source/Scene/Billboard.js#L1579).

(Of course, one could bring that up in _any_ PR, so ... let's consider this to be tracked, e.g. in the issue [around this comment](https://github.com/CesiumGS/cesium/issues/4434#issuecomment-2592934671), and hope that it will be addressed _explicitly_ at some point).

donmccurdy: Merged! No particular pressure on extra steps — refactoring could turn into a bigger scope than this fix, and just opening an issue would be more than enough IMO.

TypeScript has a distinction between `private` and [`@internal`](https://tsdoc.org/pages/tags/internal/) tags, similar to Java package-private. That can occasionally be useful, but doesn't exist in JSDoc to my knowledge. In this case ... the original choice of `private` seems reasonable, I guess the private concerns just escaped a bit over time. :)

PR #13088, opened 2025-12-08
Drafts for using dynamic content structures

javagl:
This is a **DRAFT** (never supposed to be merged), illustrating how some of the classes that are part of the [(time) dynamic 3D Tiles PR](https://github.com/CesiumGS/cesium/pull/12977) could be used within the `Multiple3DTileContent` class.

The changes for this experiment are summarized in a single commit:  https://github.com/CesiumGS/cesium/commit/6adf794327938ea414bbdd06f1949bded4ac3fe4 

I think that the `Multiple3DTileContent` class could be simplified _considerably_ with some of these helper classes (most importantly, the `ContentHandle`). It would allow to omit the
```
this._contentsCreated
this._requestsInFlight
this._cancelCount
this._externalTilesetCount
this._arrayFetchPromises
this._requests
this._innerContentResources
this._serverKeys
```
properties, as well as the
```
updatePendingRequests
cancelPendingRequests
requestInnerContents
canScheduleAllRequests
requestInnerContent
createInnerContents
createInnerContent (without 's')
handleInnerContentFailed
```
functions, which juggled with these properties (and the tileset statistics) in a less-than-straightforward way.

The state here is sufficient for loading the `Specs/MultipleContents` examples, both in their "1.0" and "1.1" versions (!), and displaying them properly. 

The specs themself will likely fail, because they are overconstrained and diligently check the behavior of the aforementioned functions. This is related to one of the greatest hurdles for actually _implementing_ this change: There are quite a few places that already do some sort of "special handling" for multiple contents. 

This is revolving around the `Cesium3DTile.hasMultipleContent` flag, some `instanceof` checks, the use of the `innerContents` getter of the `Cesium3DTileContent` interface, the usual state space and lifecycle issues related to the `Cesium3DTile` properties that involve `contentState`, `content.ready`, and `contentReady` (!), and several other details.

Eventually, one goal of actually applying this simplification could be to work towards a state where all these questions become _less_ relevant. A smaller state space is easier to reason about.

### DIFF
--- a/packages/engine/Source/Scene/Cesium3DTile.js
+++ b/packages/engine/Source/Scene/Cesium3DTile.js
@@ -1229,7 +1229,13 @@ function requestMultipleContents(tile) {
     tile._content = multipleContents;
   }
 
-  const promise = multipleContents.requestInnerContents();
+  // XXX_DYNAMIC_MULTIPLE Pragmatically claim that everything is done
+  // (which, in fact, does not have to concern the tile...)
+  tile._contentState = Cesium3DTileContentState.READY;
+  return Promise.resolve(tile._content);
+
+  /*
+  const promise = multipleContents.requestInnerContents(); 
 
   if (!defined(promise)) {
     // Request could not all be scheduled this frame
@@ -1261,6 +1267,7 @@ function requestMultipleContents(tile) {
       tile._contentState = Cesium3DTileContentState.FAILED;
       throw error;
     });
+  */
 }
 
 async function processArrayBuffer(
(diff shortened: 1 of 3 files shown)


javagl:
This is not supposed to be merged, but was intended to illustrate the possible approach: When the (formerly:) dynamic or (now: ) conditional content functionality is settled, then it should be easy to replace much of the `Multiple3DTileContent` implementation with this functionality, given that the conditonal content already needs the possibility to display multiple contents, and the current scope of `Multiple3DTileContent` is then only a special case.

PR #13090, opened 2025-12-09
Sandcastle Embedding Search

tomdicarlo:
# Description

This PR demonstrates an alternative searching approach for sandcastles that uses a vector embedding search instead of the existing pagefind search. It updates the process of building the gallery to create a vector embedding of each sandcastle, and then compares those embeddings against the user query.

To allow everything to run locally, we have selected a [small MIT licensed model from Huggingface](https://huggingface.co/avsolatorio/GIST-small-Embedding-v0). One consideration for this PR is concerns over model download size and hardware performance limitations. Memory usage from the model appeared in testing to not rise above what is consumed for several of the existing sandcastles, however the download size is approximately 30MB which is several times larger than the largest file download I have been able to produce from existing sandcastles.

I will open this PR as a draft for now, as modifications will likely need to be made to support a hybrid approach between Pagefind and Vector embedding search, which we can discuss further within the PR.

## Issue number and link

internal issue

## Testing plan

My changes were tested manually against a small subset of search queries that I found to be gaps in the current Cesium Sandcastle search.

I did not include any automated testing of the functionality as it doesn't seem testing is configured in the application, but I would be glad to add automated testing if that is requested before moving the PR out of draft.

# Author checklist

  - No, as noted in testing plan

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -188,7 +188,7 @@ APPENDIX: How to apply the Apache License to your work.
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-Copyright 2011-2020 CesiumJS Contributors
+Copyright 2011-2026 CesiumJS Contributors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/packages/sandcastle/LICENSE.md
+++ b/packages/sandcastle/LICENSE.md
@@ -188,7 +188,7 @@ APPENDIX: How to apply the Apache License to your work.
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-Copyright 2011-2020 CesiumJS Contributors
+Copyright 2011-2026 CesiumJS Contributors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.


packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Load the gallery list with embeddings
+   */
+  private async loadGalleryList(): Promise<void> {
+    if (this.galleryList) {
+      return;
+    }
+
+    try {
+      const response = await fetch("/gallery/list.json");
```
jjspace: Given the nature of the different ways sandcastle can be deployed, like CI, all paths should be relative as much as possible. This line is why it's not working in CI currently

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (this.model && this.tokenizer) {
+      return;
+    }
+
+    try {
+      console.log(
+        `Loading embedding model: ${this.modelId} (this may take a moment on first load)...`,
+      );
+      this.tokenizer = await AutoTokenizer.from_pretrained(this.modelId);
+      this.model = await AutoModel.from_pretrained(this.modelId);
```
jjspace: I assume this downloads the model directly from huggingface? Is there a way to include it locally? Or is that bad practice or limited by the license?

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -30,126 +27,149 @@ export type GalleryItem = {
...
+    VectorSearchResult[] | null
   >(null);
   const [isSearchPending, startSearch] = useTransition();
+  if (defaultSearchFilter !== null && searchFilter === null) {
+    setSearchFilter(defaultSearchFilter);
+  }
+
+  // Perform vector search when search term changes (with debounce)
   useEffect(() => {
-    const pagefind = getPagefind();
```
jjspace, line 67: I don't think we want to completely remove pagefind so we still support offline or isolated environments

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -30,126 +27,149 @@ export type GalleryItem = {
...
+      .map(
+        (result: {
+          legacy_id: string;
+          id: string;
+          rank: number;
+          distance: number;
+        }) => {
+          // Find the item by matching the legacy_id to the slug
+          // The legacyIds map should provide legacy_id -> slug mapping
+          const slug = legacyIds[result.legacy_id];
```
jjspace: This is not the right way to use this. The slug should already be the id. The legacy id is only there for loading historical urls for the previous version of sandcastle.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -30,126 +27,149 @@ export type GalleryItem = {
...
+
+          const item = items.find((item: { id: string; url: string }) => {
```
jjspace: I don't really understand this search function, is this just trying to make up for bad data from the embedding search?

packages/sandcastle/src/Gallery/GalleryItemStore.ts
tomdicarlo: I've updated the branch to now function as a hybrid of Pagefind and Vector search.

tomdicarlo: Also, is it typical to only allow the creator of a conversation to resolve it, or am I okay to mark conversations like this as resolved once I respond to the feedback?

packages/sandcastle/src/Gallery/GalleryItemStore.ts
tomdicarlo: Yeah the search logic needed to be simplified a bunch, just tidied everything up and going to switch to a normal PR (not-draft)

packages/sandcastle/src/Gallery/GalleryItemStore.ts
tomdicarlo: Same as above. Search logic has been refactored.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
tomdicarlo: Yes, this downloads from huggingface. Downloading it locally should be possible. I don't know of it being bad practice to do so, and should fall within license (MIT).

I would think that having it local is probably ideal, as we would be vulnerable to huggingface removing the model (though this seems very unlikely).

tomdicarlo:
> Thanks for the PR @tomdicarlo. I have a few initial comments about the code and some larger questions.
> 
> 1. We previously discussed a "hybrid" approach to mixing the pagefind results with the embedding results or displaying both. Is there a reason we're not? was it just as a POC or do you feel the new search is stronger and fast enough already?
> 2. The new search seems to completely ignore the labels filter. Is this intended to be incorporated at some point and just isn't included yet?
> 3. I think it would be really helpful to have a loading spinner or some sort of indication while the search is happening. This is something we probably should already have but I can see it being needed even more with potential LLM latency (even though it is pretty fast on my machine currently)
>    
>    * I thin this would be extra helpful for the initial loading of the model which may be slower on slower networks
> 4. Can you explain what knobs we have to tune the results?
>    
>    * One sub-point here. After some quick testing it feels like it sometimes returns _more_ results than it "should". For example searching for "Moon" does give the moon sandcastle but it also gives a bunch of other unrelated sandcastles. We don't have many in the first place but I wonder if it's actually helpful to include these results that are so separate.
> 5. we previously discussed potentially using the embeddings to generate extra metadata about sandcastles a build time or as a one time build step. I forget if that's still being considered? and if it is, is that part of the plan for this PR or a separate effort? (fine if it's separate, just clarifying)

1. Updated to include a hybrid approach.
2. Fixed the issue introduced with filter label
3. With the hybrid search, the pagefind results will load immediately even if embeddings results aren't ready yet. And even on my lower spec hardware (running on my phone), I was able to get the embeddings results imperceptibly fast. Not sure if the loading spinner is necessary with that in mind.
4. The main knob is the threshold for the embeddings is on line 126 of GalleryItemStore, which sets the distance threshold for considering a sandcastle to be related to the user query. I picked it based on testing of a few examples, but we could make that logic more statistically based if we wanted.
5. That isn't something that is included in this PR. I have some reservations about that approach, as generating tags feels like a more limited version of using the embeddings. Generating the tags using an LLM is doing all the work of embedding the sandcastles, but then removing most of the searchability benefit by only limiting them to a subset of the context determined by how we generate the tags. And generating the tags in of itself is a bit of a can of worms (i.e. how do we know how many tags to generate and how should we determine their scope). There's definitely a way that we can make it work, but I think there is a much better solution attainable via the embeddings model.

CHANGES.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 1.138 - 2026-01-27
+
+- Updated Sandcastle Gallery creation process to leverage MIT licensed Huggingface model to vectorize each sandcastle for embedding search
+- Modified Sandcastle application to use a hybrid pagefind and embedding search
```
jjspace: I wonder if we should create a changes file for sandcastle specifically. These changes don't really have anything to do with the cesium js library...

At the very least they should go under the sandcastle package header

packages/sandcastle/sandcastle.config.js
```diff
@@ -14,7 +14,7 @@ const config = {
         termSimilarity: 0.85, // Fuzzier word matching
       },
     },
-    defaultFilters: { labels: "Showcases" },
+    defaultFilters: null,
```
jjspace: I don't think we want to do this. Using the showcases label as the default filter was intentional to highlight a variety of CesiumJS features without overwhelming users with too many options. 

I think I saw this may have been in response to some forum feedback, can you expand on the reasoning/motivation here?

packages/sandcastle/scripts/buildGallery.js
```diff
@@ -274,6 +302,11 @@ export async function buildGalleryList(options = {}) {
   // regardless of if titles match the directory names
   output.entries.sort((a, b) => a.title.localeCompare(b.title));
 
+  const embeddings = await generateEmbeddings(output.entries);
+  output.entries.forEach((entry, index) => {
+    entry.embedding = embeddings[index];
+  });
```
jjspace: I have some concerns about this part of the build.

1. Can we split this out to a separate step? Would that even make sense?
    - Right now `buildGallery` takes on average `0.7s` on my system. With this change it takes `~70s` which is 100x slower. I'd really like to reduce or eliminate that number somehow. Is there a way to cache or save this out to a file? Open to ideas
3. The actual embeddings seem _very_ big and I think we should probably look into another way to deliver them to the client than the same `list.json` we're currently using
    - Right now on [CI in main](https://ci-builds.cesium.com/cesium/main/Apps/Sandcastle2/index.html) the `list.json` is `~122kB` which I feel is kinda reasonable. On [CI in this branch](https://ci-builds.cesium.com/cesium/SandcastleEmbeddingSearch/Apps/Sandcastle2/index.html) the list is `2.4MB` because every entry has an array of 383 precise floating point numbers.
    - I realize we probably need to get all of this data to the client to do the full vector search but I'd like to review ways we could reduce the data size and potentially load separately. The `list.json` is needed very early in loading iirc and I'd like to avoid introducing more delay to that request if we can avoid it.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,152 @@
...
+  private model: PreTrainedModel | null = null;
+  private modelId: string = "avsolatorio/GIST-small-Embedding-v0";
+  private onInitializedCallbacks: (() => void)[] = [];
+
+  async initialize(): Promise<void> {
+    if (this.galleryList && this.model && this.tokenizer) {
+      return;
+    }
+
+    const response = await fetch("gallery/list.json");
```
jjspace: It seems we are now fetching this twice on page load. Is there a reason for that? Given my other comment about the increased size of this request the problem seems compounded by doing the request twice.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -83,16 +105,34 @@ export function useGalleryItemStore() {
...
       startSearch(() => setSearchResults(values));
-    });
-  }, [searchTerm, searchFilter]);
+
+      if (
+        embeddingModelLoaded &&
+        searchTerm !== null &&
+        searchTerm.trim() !== ""
+      ) {
+        const vectorResults = await vectorSearch(searchTerm, 5, searchFilter);
```
jjspace, line 135: This search seems relatively expensive/slow with the vector comparisons. In a quick profile this single function took ~700ms. Can we:
1. Debounce the search so it doesn't happen many times while users are typing fast and thus unlikely to care about the results?
2. Utilize some sort of loading state and indicator so that the UI is not blocked and still performant while the search is taking place then load the results when it's done

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -102,7 +142,38 @@ export function useGalleryItemStore() {
 
       return applyHighlightToItem(item, result);
     });
-  }, [items, searchResults]);
+
+    if (vectorSearchResults && vectorSearchResults.length > 0) {
+      for (const vectorResult of vectorSearchResults!.reverse()) {
```
jjspace:
```suggestion
      for (const vectorResult of vectorSearchResults.reverse()) {
```
Minor thing but I think it's a code smell to use the `!` to force TS to be happy. It also doesn't look like it's actually needed here anymore.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -102,7 +142,38 @@ export function useGalleryItemStore() {
...
+        const exists = pagefindResults.find(
+          (res) => res && res.id === vectorResult.id,
+        );
+        // This similarity threshold is the cutoff for showing a a vector search result
```
jjspace:
```suggestion
        // This similarity threshold is the cutoff for showing a vector search result
```

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -102,7 +142,38 @@ export function useGalleryItemStore() {
...
+        // There is a tradeoff depending on user query complexity
+        // Often, shorter queries may want a slightly higher threshold (~0.75)
+        // However, this number was based on testing with more complex queries
+        const similarity_threshold = 0.727;
```
jjspace, line 172: You definitely know more about this area than I do. Is this something that could/should be dynamic based on the input? Definitely something we could tune later or in a separate PR but I was just curious

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -102,7 +142,38 @@ export function useGalleryItemStore() {
 
       return applyHighlightToItem(item, result);
     });
-  }, [items, searchResults]);
+
+    if (vectorSearchResults && vectorSearchResults.length > 0) {
+      for (const vectorResult of vectorSearchResults!.reverse()) {
+        const exists = pagefindResults.find(
+          (res) => res && res.id === vectorResult.id,
+        );
```
jjspace: Minor nit-pick: This value isn't used until after the threshold check which may skip the code where it's used. Can we move it down closer to the `if` statement it's actually used for? (And potentially save the computation when it's not needed)

packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -102,7 +142,38 @@ export function useGalleryItemStore() {
...
+        if (vectorResult.score < similarity_threshold) {
+          continue;
+        }
+
+        if (exists) {
+          pagefindResults.splice(pagefindResults.indexOf(exists), 1);
+          pagefindResults.unshift(exists);
+          continue;
+        }
+        if (!exists) {
```
jjspace:
```suggestion
        } else {
```

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,152 @@
...
+  async search(
+    query: string,
+    limit: number = 10,
+    filters?: Record<string, string | string[]> | null,
+  ): Promise<VectorSearchResult[]> {
+    if (!query || query.trim().length === 0) {
+      return [];
+    }
+    if (!this.galleryList || !this.model || !this.tokenizer) {
+      console.log("Search occured before initialization completed.");
```
jjspace: Debug log? I don't believe we need this?

(I leave these kinds of things in all the time too)

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
jjspace: Did you happen to look into this more? I'm not sure the best approach but I'd like to know the process if it's something we want to/can do

packages/sandcastle/src/Gallery/GalleryItemStore.ts
jjspace: > Also, is it typical to only allow the creator of a conversation to resolve it, or am I okay to mark conversations like this as resolved once I respond to the feedback?

@tomdicarlo We don't have a hard rule for the team. I _personally_ prefer to resolve things myself or let the first author mark resolved just so they see any responses. That said if it's small things like typos I'm also perfectly happy if you mark as resolved to just clean up the remaining tasks.

packages/sandcastle/sandcastle.config.js
tomdicarlo: I'll trust your feedback here and can revert it, but yes this decision was influenced by user feedback. One user specifically mentioned not liking the showcases being on by default, and another was confused by the inconsistent search behavior after the first search clearing the showcase filter.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
tomdicarlo: This is definitely something we could try to adjust dynamically, but it's going to be tricky to determine a "best" value. It's going to be a very contextual based on the query. But as noted in the comment, in my testing, more complex queries yielded a lower similarity value. An easy starting point could be to reduce the threshold based on the number of words in the query. We can start with that and expand in a future PR if we want?

CHANGES.md
tomdicarlo: I'll add the @cesium/sandcastle header. I think it makes sense either way. I would be more inclined to separate them if there was an effort to make sandcastle more standalone from the other cesium code in this repo, but I'm not sure if that's the case.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
jjspace: I've only tested a handful of queries at this point, I'd assume less than you have. I'm happy to leave it as is for now and avoid the extra complexity if you don't think it's worth it at this point. Definitely something that seems easy to adjust later as long as we know the "knob" exists

packages/sandcastle/scripts/buildGallery.js
tomdicarlo: 1. Can you elaborate on your concern of the time for this one? Is it mainly annoyance during local development? We could just cache the values and re-generate whenever a sandcastle is changed. We would still want to regenerate every time during actual deployment. So we'll still need to eat the longer embedding times on occasion. Also if the concern is for local testing, having the model downloaded locally could also help. I'll look into that.

tomdicarlo: 2. I don't think there's much of an opportunity to reduce embeddings size, they're pretty minimal as-is. In my testing, going with anything smaller incurred pretty heavy losses in terms of search quality. I think our best bet is load in as a separate file so it doesn't impact list.json loading.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
tomdicarlo: Made two changes relevant here. Now passing in the gallery list to embedding search so a second load doesn't need to occur. Also split embeddings into its own file so the initial gallery list load is reduced.

packages/sandcastle/scripts/buildGallery.js
tomdicarlo: I pushed changes that will check if any changes were made to the sandcastles and will not regenerate embeddings if they are the same (uses a simple hash to run this check). Will substantially reduce build gallery times when cache is used.

I also went ahead and moved the embeddings into their own file so the loading process for that file is separate from list.json.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
tomdicarlo: Looked into this, and it is possible. Unfortunately, the model size exceed GitHub's maximum file size (100MB), so I chose to remove it, especially since the time save was negligible in the local caching scenario.

jjspace: I think as long as the app gracefully handles a missing model and/or missing embeddings then that's fine. Hopefully that only happens in more isolated local/offline environments but we absolutely don't want it failing or crashing the page if that happens.

I think it could be good to make sure we're testing by blocking the model load request sometimes

jjspace:
Just some notes from our meeting

- We _do_ want to add a setting for this. Something like "Enable semantic search" or "Enable embeddings search" in the settings modal. We decided this does not need a visible toggle on the Gallery page itself for now
    - There should be a short description that also explains this is run _locally_ and user data is not sent anywhere
    - On by default
- There were considerations to avoid downloading the model until the user actually interacts with the search box. Could be nice to save bandwidth but may slow down the first search. I think we can leave this out for now unless we see it being a problem.
- May be good to check/consider the logic of removing a model from the cache should we switch to another one.
    - May also be worth storing the model id the embeddings are generated with in the `embeddings.json` file itself
- The removal of the default `Showcases` filter on first search isn't working currently. I think this was just a partial revert from your earlier changes. Hopefully an easy fix but that will avoid the "no results" confusion

packages/sandcastle/scripts/buildGallery.js
jjspace: The embeddings in their own file is great now. The hash check is even better and build times are back down to under a second for me after the first one. (The first is down to 30s too so whatever else you changed is nice). it'd be a "nice to have" if it could check _per sandcastle_ but not required for this PR imo.

tomdicarlo: Yeah I changed it so that it used the quantized (using a smaller floating point size) model during the embedding file creation process, this is what we use in the frontend as well, so shouldn't come with any accuracy impacts.

packages/sandcastle/src/Gallery/GalleryItemStore.ts
tomdicarlo: Just pushed a change that adds debouncing for calling the embedding search

tomdicarlo:
> Just some notes from our meeting
> 
> * We _do_ want to add a setting for this. Something like "Enable semantic search" or "Enable embeddings search" in the settings modal. We decided this does not need a visible toggle on the Gallery page itself for now
>   
>   * There should be a short description that also explains this is run _locally_ and user data is not sent anywhere
>   * On by default
> * There were considerations to avoid downloading the model until the user actually interacts with the search box. Could be nice to save bandwidth but may slow down the first search. I think we can leave this out for now unless we see it being a problem.
> * May be good to check/consider the logic of removing a model from the cache should we switch to another one.
>   
>   * May also be worth storing the model id the embeddings are generated with in the `embeddings.json` file itself
> * The removal of the default `Showcases` filter on first search isn't working currently. I think this was just a partial revert from your earlier changes. Hopefully an easy fix but that will avoid the "no results" confusion

Just pushed several changes addressing this comment. Added a setting to enable/disable embedding search. Fixed the showcase filter issue. Added the model to the embeddings.json file.

Notably, I didn't make any changes to the timing of the model download, this still occurs at the start (but not if the setting is disabled). Also I didn't look into tweaking the caching of the model since we explicitly specify the model in the embeddings now. Should allow us to adjust in the future if we decide to change model.

Let me know if you have any other thoughts, and if not, I think I've address all the comments you've provided.

LICENSE.md
```diff
@@ -1039,6 +1039,212 @@ http://www.jshint.com/
 >
 > THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE. JSHint was forked from the 2010-12-16 edition of JSLint.
 
+### transformers.js
```
jjspace: Is it necessary to include this in the license? I don't see any code we're using from them directly or a `transformers.js` file in the changes.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+  url: string;
+  id: string;
+  title: string;
+  thumbnail: string;
+  lineCount: number;
+  description: string;
+  labels: string[];
+}
+
+interface GalleryListItem {
```
jjspace: This is a bit of a nitpick but could we move and/or combine these `GalleryList` types with the `GalleryItem` in the `GalleryItemStore.js` file? That's where the primary loading of the list happens and probably where the types "belong". They're not specific to the Vector search right?

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+class EmbeddingSearch {
+  private galleryList: GalleryList | null = null;
+  private embeddings: EmbeddingsMap | null = null;
+  private tokenizer: PreTrainedTokenizer | null = null;
+  private model: PreTrainedModel | null = null;
+  private modelId: string = "avsolatorio/GIST-small-Embedding-v0";
+  private onInitializedCallbacks: (() => void)[] = [];
+
+  async initialize(galleryList: GalleryList): Promise<void> {
+    if (this.embeddings && this.model && this.tokenizer) {
```
jjspace: Can/Should this be combined into an `isInitialized` property? I think that might read clearer and is useful for other places like in the `onInitialized` function where you do the same check.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    this.galleryList = galleryList;
+
+    const embeddingsResponse = await fetch("gallery/embeddings.json");
+    this.embeddings = await embeddingsResponse.json();
+
+    this.tokenizer = await AutoTokenizer.from_pretrained(this.modelId);
+    this.model = await AutoModel.from_pretrained(this.modelId, {
+      dtype: "q8",
+    });
```
jjspace: 1. Should `this.modelId` be based on the value that is downloaded with the `embeddings.json`?
2. If it is do the options params of `dtype: 'q8'` need to also be dynamic based on the model that generated them?

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+  description: string;
+  labels: string[];
+}
+
+export interface GalleryList {
+  entries: GalleryListItem[];
+  legacyIds: Record<string, string>;
+}
+
+type EmbeddingsMap = Record<string, number[]>;
```
jjspace: This is technically correct right now but maybe not clear about the metadata. I think it might be good to change this structure like so:
```ts
type Embeddings = {
  id: string;
  model: string;
  embeddings: Record<string, number[]>;
};
```
Then it's more clear what is "metadata" and what is the actual set of embedding vectors that was generated. It would also allow easier addition of more metadata and avoid rare but possible conflicts between sandcastle id and metadata value.

This would need a slight change to the generation code too.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
```diff
@@ -0,0 +1,160 @@
...
+  title: string;
+  thumbnail: string;
+  lineCount: number;
+  description: string;
+  labels: string[];
+}
+
+export interface GalleryList {
+  entries: GalleryListItem[];
+  legacyIds: Record<string, string>;
```
jjspace: Unless I'm missing something it doesn't look like the `legacyIds` are actually used anywhere in the embedding search anymore? Should this whole type (and places it's used) be simplified to `GalleryItem[]` from the `GalleryItemStore`?

jjspace (commented):
Thanks for the updates @tomdicarlo, I think this is getting really close! I just had a few more comments taking a much closer look at the search logic.

LICENSE.md
tomdicarlo: We're using @huggingface/transformers in order to create the embeddings. I don't know if that requires us to update the license, I just saw that similar packages were included in the license.

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
tomdicarlo: I can do that, but I'll need to add a few non-null assertion operators, typescript doesn't recognize the individual components as non-null even if the isInitialized method returns true

tomdicarlo:
@jjspace oops, I just realized that I never pushed the changes that I alluded to this morning. Just pushed them, and they should address your most recent comments!

packages/sandcastle/src/Gallery/EmbeddingSearch.ts
jjspace: I didn't even think about this but now I really wish it was possible to have [multiple type predicates](https://github.com/microsoft/TypeScript/issues/26916). I'm fine leaving this as is to avoid more churn but something to keep in mind for the future.

LICENSE.md
ggetz: I don't think it hurts to include the license, but I might suggest creating a separate `LICENSE.md` for the Sandcastle app itself. Since this is not used for the core library, including it in the top-level `LICENSE.md` file could potentially create downstream confusion.

jjspace: I just went ahead and created a new `LICENSE.md` for sandcastle copying the top level one and moving this section over there to expedite merging. any edits needed can be made as follow ups but I don't think it has to block this PR

PR #13093, opened 2025-12-10
Create reference implementation for BENTLEY_materials_point_style

### LINKED ISSUE #12891: Create reference implementation for proposed glTF extension BENTLEY_materials_point_style
### Feature

Depends on https://github.com/CesiumGS/glTF/pull/91.

Within CesiumJS, consider creating a reference implementation of the proposed glTF extension BENTLEY_materials_point_style as it is described at the above link.

This item should be completed after the above-linked PR is more fleshed out / fully proposed.

markschlosseratbentley:
# Description

This PR adds support for the `BENTLEY_materials_point_style` glTF vendor extension. The specification PR (now merged) can be found [here](https://github.com/CesiumGS/glTF/pull/91) and the specification itself currently lives [here](https://github.com/CesiumGS/glTF/tree/vendor-extensions/extensions/2.0/Vendor/BENTLEY_materials_point_style).

This PR allows CesiumJS to process and apply the above extension when loading glTF files. This means point primitives will be able to have a `diameter` property specified and respected in CesiumJS when loaded via glTF.

To test these changes locally:
- Clone this branch.
- Build CesiumJS and Sandcastle.
- Run the dev server locally.
- Connect to the dev server using your local browser, clicking on Sandcastle.
- Navigate to the `Styled glTF Points` example.

Here is a screenshot of test file `points-r5-g8-b14-y10.gltf` being rendered in CesiumJS

thumbnail

This change further enables CAD-style workflows in CesiumJS.

## Issue number and link

Fixes #12891 

## Testing plan

This has been tested by:
- Adding a Cesium Sandcastle sample.
- Adding a unit test.

# Author checklist


Specs/Data/Models/glTF-2.0/StyledPoints/points-r5-g8-b14-y10.gltf
markschlosseratbentley, line 1: @ggetz The glTF test file had to be duplicated to be seen by both:
- Unit tests (`Specs/Data/Models/glTF-2.0/StyledPoints/points-r5-g8-b14-y10.gltf`).
- Sandcastle (`Apps/SampleData/models/StyledPoints/points-r5-g8-b14-y10.gltf`).

Is there a better way to do this without duplication?

packages/sandcastle/gallery/styled-gltf-points/index.html
lilleyse, line 1: This demo is pretty simple and could be confused with point cloud declarative styling.

Would it make sense to move this to the `Development` folder until we have a more interesting dataset to show?

lilleyse (commented):
Looks good. I left some comments in https://github.com/CesiumGS/glTF/pull/91, but those may or may not affect this PR.

packages/sandcastle/gallery/styled-gltf-points/index.html
markschlosseratbentley: That make sense to me; thank you.

I believe I have just done this. What I did:
1) I moved the sample to `packages/sandcastle/gallery/styled-gltf-points-dev/`.
2) Inside `sandcastle.yaml`, I added `development: true` and renamed the sample to `Styled glTF Points - Dev`. I also removed the `Showcases` label and added `Development`.

Locally, I tested this by setting `PROD=1` before building and hosting the gallery locally.  I did not see this example (or any dev-tagged examples) showing up.

Feel free to resolve this conversation if this looks good to you!

lilleyse: Ah, I'm not familiar with the new workflow. Maybe @ggetz can confirm.

ggetz: Tagging the example with `development: true` is sufficient. :+1:

Specs/Data/Models/glTF-2.0/StyledPoints/points-r5-g8-b14-y10.gltf
ggetz: If it's a development Sandcastle example, I think it should be fine to have the glTF data only in the `Specs` folder. It should be able to be referenced from a Sandcastle example, as in the "picking-dev" example.

ggetz:
@jjhembd Could you take a review pass on the code here, and merge once happy?

Specs/Data/Models/glTF-2.0/StyledPoints/points-r5-g8-b14-y10.gltf
markschlosseratbentley: I removed the copy of the glTF data from `SampleData` and left the one in `Specs`. I made the development Sandcastle example reference the `Specs` copy and it worked fine locally.

Resolving this comment (my own comment). Please feel free to reopen or leave another comment if anything looks amiss with this approach.

markschlosseratbentley:
@jjhembd Just checking in on if you have any bandwidth coming up to give this a look + merge once happy. Thank you so much!

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -1811,6 +1811,16 @@ function loadMaterial(loader, gltfMaterial, frameState) {
   material.alphaCutoff = gltfMaterial.alphaCutoff;
   material.doubleSided = gltfMaterial.doubleSided;
 
+  // BENTLEY_materials_point_style extension
+  const pointStyleExtension = extensions.BENTLEY_materials_point_style;
+  if (defined(pointStyleExtension) && defined(pointStyleExtension.diameter)) {
+    const diameter = pointStyleExtension.diameter;
+    // Validate that diameter is a positive integer as the extension specification requires.
+    if (diameter > 0 && Math.floor(diameter) === diameter) {
```
ggetz: Small thought, no need to hold up this PR for it— 

I understand the spec requires a integer diameter, which we should definitely strictly enforce on the authoring side. But should we be more permissive loading here on the runtime side and, say, just floor the diameter value as a best effort rather than not applying a diameter at all?

packages/engine/Source/Scene/ModelComponents.js
```diff
@@ -1623,6 +1623,16 @@ function Material() {
    * @private
    */
   this.unlit = false;
+
+  /**
+   * The point diameter in pixels for POINTS primitives. This is set by the
+   * BENTLEY_materials_point_style extension.
+   *
+   * @type {number}
```
ggetz:
```suggestion
   * @type {number|undefined}
```

Correct?

packages/sandcastle/gallery/styled-gltf-points-dev/index.html
```diff
@@ -0,0 +1,36 @@
+<!doctype html>
```
ggetz: Most of this `index.html` file is overwritten and all you need is the `body` content. So IIUC, this file can be replaced with:

```html
<style>
  @import url(../templates/bucket.css);
</style>
<div id="cesiumContainer" class="fullSize"></div>
<div id="loadingOverlay"><h1>Loading...</h1></div>
<div id="toolbar"></div>
```

packages/sandcastle/gallery/styled-gltf-points-dev/sandcastle.yaml
```diff
@@ -0,0 +1,7 @@
+legacyId: Styled glTF Points.html
```
ggetz:
```suggestion
```

`legacyId` is not needed for new Sandcastle examples.

packages/engine/Source/Scene/ModelComponents.js
markschlosseratbentley: Yes, correct, thanks.

packages/sandcastle/gallery/styled-gltf-points-dev/index.html
markschlosseratbentley: Yes, this suggestion seems right to me; thanks for pointing it out. I went ahead and applied this change and tested it locally. The sandcastle demo still built + ran fine.

packages/engine/Source/Scene/GltfLoader.js
markschlosseratbentley: Hey @ggetz, I agree with your feedback and pushed some changes to be more permissive here on the runtime side. Could you give these changes a look and resolve this conversation if they look okay to you? I had to change the check to be >= 1 to ensure flooring something like 0.3 did not result in a value of 0 getting passed along. I also altered a relevant unit test which needed changing based on this revision. Unit tests still pass and the sandcastle demo still runs fine.

packages/engine/Source/Shaders/Model/ModelFS.glsl
```diff
@@ -29,6 +29,14 @@ SelectedFeature selectedFeature;
 
 void main()
 {
+    #ifdef PRIMITIVE_TYPE_POINTS
```
javagl, line 32: This part turned the default shape of points from squared to circles (see https://sandcastle.cesium.com/?id=3d-tiles-point-cloud-shading )

Reported in https://community.cesium.com/t/point-cloud-rendering-default-point-shape-changed-from-square-to-round-since-v1-138/45226

ggetz: Thanks @javagl! CC https://github.com/CesiumGS/cesium/issues/13217, which @markschlosseratbentley is going to take a look at.

ggetz: (My fault for missing this during review)

PR #13097, opened 2025-12-15
Fix scene `sampleHeightMostDetailed` does not work when a tileset is added by primitive collection

plainheart:
# Description

Fix the bug that `sampleHeightMostDetailed`/`clampToHeightMostDetailed` can not get height from a tileset that is added to the `scene.primitives` by a primitive collection. This bug was originally reported in the [community](https://community.cesium.com/t/clamptoheightmostdetailed-get-the-elevation-problem/15056) 4 years ago.

#### Possible Cause:

https://github.com/CesiumGS/cesium/blob/8209a5964cdca3f8bfb5ce7d7f14873704cd8332/packages/engine/Source/Scene/Picking.js#L1008

Here we can know `scene.primitives.contains(tileset)` will return `false` when the tileset is added by primitive collection - That is not expected, and causes the height sample to fail.

Here is a comment that says:

> // Only update tilesets that are still contained in the scene's primitive collection and are still visible
> // Update tilesets continually until all tilesets are ready. This way tiles are never removed from the cache.

But we can know the `tilesets` are already collected from the `scene.primitives` from the following function (Line 1070): 

https://github.com/CesiumGS/cesium/blob/8209a5964cdca3f8bfb5ce7d7f14873704cd8332/packages/engine/Source/Scene/Picking.js#L1061-L1080

So I think we can remove the `scene.primitives.contains(tileset)` check to make the sample work, or consider adding a new function to check whether a tileset is in `scene.primitives` to replace this check?

#### Comparison

| Before | After |
| :----: | :----: |
| After | Before |

## Issue number and link

Fix https://community.cesium.com/t/clamptoheightmostdetailed-get-the-elevation-problem/15056

## Sandcastle

https://sandcastle.cesium.com/#c=7VVRa9swEP4rh5/cEeSV9KVdElbSwQodC7SsD/MeFPsSi8mSkWSHrOS/72xZTUIcBmNjL/OLrdN9d999d4dFWWnj4A1wC3O0oi5hZXQJaZR1pzR6l6pUZVpZB43ADRqYgsJN782+dLY4+M+1clwoNGl00SGTBJ65ywpwBQLdWi0RdO2q2qXK1Cr2btxuVQarWmVOaAXdBbykCoBvuHDA8/xJSLToOsCp+Vm4YmFEKZxocK6lxC5S570byHAY0Cfy1TGboUJWhVCWGSx1g7dS9lQBvBrOw0kOz6UXxL/Gd3101sp5r9WtpcN9Ho+vrsbXl9e+iHMpiVzchz9yXMntkw43I3iBvDa8LecG3sIu0PN02h4tKLewGMcGSfcGL2A688VCoM+EoqRcdnQfNM8xb9N/aFC5B2EdMTOx1y4+hAcVKpF9X2grOlGnryJw4/Ta8KoQWadAa6ArruIA3zNY6lrlQq0fqwINMhLDoRkFPy9A+/i6jjSzvKwkfkSxLtwnbd0d0vRJzOOvh7y+7WNkfgKZ1OsDJmm0IHfMoegiwYaGCSxRolkNbX5tz00ajfbQwzzMwweo9/rHwbLzH7tfT+fZuf69oa1OQx3v87kd+oNTP8Dh7MQPrsZAgP9r8q/W5GQ/CBy68hdWJRpFE+u2Emet9b3w/6/ayJixxCHVyUnBZFkTUccya1vcJAmQSS4aEPl04H8FmeTW0s2qlvJR/MA0mk0S8j+CSeo9NeFzg0bybetSXM4evJExNknoeIpyWsslNwcRfwI

Another sandcastle based on the [Sample Height from 3D Tiles](https://sandcastle.cesium.com/?id=sample-height-from-3d-tiles) example:

https://sandcastle.cesium.com/#c=vVbbbuM2EP0V1uiD3PXSulmXxBs0SFxsgNywa3SxqAssI9ExsTQpkHRcN/C/dyRKshyrRfvSF0kkZ+bMnLlQbF1IZdBPiGh0RTXbrNFSyTVaDLJqtRicL8RCZFJog14Y3VKFPiBBt7U0/rXacxr5KykMYYKqxWCEXhcCIUOVgp2zRmFu17iE+SIVz+sNZzhaiP3wvAHTGRUUsCworpaVL2yJnB+qJc44WRdz+ZGy55X5vCnKWGg+tMBbJnK5xYRTZcC/+Ypp9KTkVkMIuaQaCQkoVgkdWbqT2lxTiIPTHC8GpU/7ErkGJWuqCNbUlLE7FVZOtWGCGCYh0A47VwQc0oyIwCnFEPJ8Lwo9D7txmMRx6kfhyB68D+MgCrwAe3AUT5Kg3g/dxAsmKZ54/mQSebFvFYbVUypGge8T2I+U5Ew8PzKTrT5JzmvwEPtBmvoTzw0mURJESYPt4jD1PD8NkiBMXB/w65MI+3GahKmf+m6STKIuOBX5XBGhl1Kt2+zeEaPYHyG+uZ7dz2/mX5ucLoRRO5sXm14D5AKFkGCyJcy0hFWv4Hpuj6squZHiUsPiJneAjSgq7SE0HqOtVN81MmVit2RXb9okFYqtmWEvVGOS506NZj2pxNoKKI2g7YoKBIJAGiKtb0bC4rGxdAVE0qzk+hBFcXp43B492s6w1O/RfOsp6o+lR7OqUJQRSDdyoJ2kGh64lpxiLp+db7PyAHFZ1UYT5Bn68bXS2H9r65zoncjQciNsQBpag1PbGtqpLddtCeUHvoBniq7lC72EWmtItgxlTQd4x8T0dkaQutiNgsAPg3Ry3BghTrwwjNMwOOqM0MNpnMRp7IO6Lc7zHnD/34D70GWhFwReEiVHbenHPnb91HVTaIYOeujGOA1Cv3TLi1v0Dr7ciLLEA7fPKV07dQnzb+dUstZ56CjkcChABhLuObym1tQ5eveO1QlozMnl0vYRQ2NkraD3yLOWUAftN/Y7SJ0wgGE+FjUNHXFvdLLlt1sWs132Mzusz60n+yNeymlL86suE3YM9Mz17jR2DuEc8twU+InV4X8g8209l5322gRYSM3skD3BAFZbHijnrNCS5WeoVUVIQb+xv7sVXOyPUPMYjg5aawIXJyO8nazQ7VLhT7PrVmhff+2PGP6HQArJdxwu5457TWS6J7QWiKhsvito68qlXeP7h/tZK7RluVmdoUORHCLozsPahYeNKV93tcyjkgVc0zunw1tWBvwm/K+z29uHLwcGDozltDCrX6BI7v5HWJuB/eEH4XhSwvZgNJhqs+P0opT8mdnfrY3iDsZjQ0EcXNHjp032HW67TFdlPR03KtOcvSCWf+j5vSrzpTWcLDecf2Z/0sXgYjoG+SO1eto/vFDF4YoEkZV3cWs3McbTMSxPtYyU/ImojsW/AA

# Author checklist

- [x] I have added my name to `CONTRIBUTORS.md.`

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -66,9 +66,9 @@ jobs:
           unzip Cesium-$(cat package.json | jq -r '.version' | sed 's/\.0$//').zip -d Build/release/ -x "Apps"
           aws s3 sync Build/release/ s3://cesium-website/cesiumjs/releases/$(cat package.json | jq -r '.version' | sed 's/\.0$//')/ \
             --cache-control "public, max-age=1800" \
-            --exclude "Build/release/Apps/Sandcastle2/*" \
+            --exclude "Apps/Sandcastle2/*" \
             --delete
-          aws s3 sync Build/Sandcastle2/ s3://cesium-website/cesiumjs/releases/$(cat package.json | jq -r '.version' | sed 's/\.0$//')/Apps/Sandcastle2/ --cache-control "public, max-age=1800" --delete
+          aws s3 sync Apps/Sandcastle2/ s3://cesium-website/cesiumjs/releases/$(cat package.json | jq -r '.version' | sed 's/\.0$//')/Apps/Sandcastle2/ --cache-control "public, max-age=1800" --delete
 
   deploy-docs:
     runs-on: ubuntu-latest
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
 - Fixed memory leak when rendering Gaussian splat 3D tilesets. [#13229](https://github.com/CesiumGS/cesium/pull/13229/)
 - Fixes a regression with the NGA-GPM local extension and custom shaders. [#13247](https://github.com/CesiumGS/cesium/pull/13247)
 - Fixed scene `sampleHeightMostDetailed` does not work when a tileset is added by primitive collection. [#13097](https://github.com/CesiumGS/cesium/issues/13097)
+- No longer disables custom shaders for primitives with missing metadata, as long as the metadata exists on the overall class definition. [#13258](https://github.com/CesiumGS/cesium/pull/13258)
 
 ## 1.139 - 2026-03-02
 
--- a/packages/engine/Source/Scene/Model/CustomShaderPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/CustomShaderPipelineStage.js
@@ -730,7 +730,9 @@ function getAllPropertyIds(primitive, structuralMetadata) {
   function addUsedPropertyIds(propertyContainerIds, propertyContainers) {
     for (const id of propertyContainerIds) {
       const propertyContainer = propertyContainers[id];
-      Object.keys(propertyContainer.properties).forEach((propertyId) =>
+      const classProperties = propertyContainer.class.properties;
+
+      Object.keys(classProperties).forEach((propertyId) =>
         primitivePropertyIds.add(propertyId),
       );
     }
--- a/packages/engine/Specs/Scene/Model/CustomShaderPipelineStageSpec.js
+++ b/packages/engine/Specs/Scene/Model/CustomShaderPipelineStageSpec.js
@@ -66,6 +66,27 @@ describe("Scene/Model/CustomShaderPipelineStage", function () {
     ],
   };
 
+  const primitiveWithMetadataProperties = {
+    ...primitive,
+    propertyTextureIds: [0],
+    propertyAttributeIds: [],
+    featureIds: undefined,
+  };
+
+  const structuralMetadataWithTextureProperty = {
+    propertyTextures: [
+      {
+        class: {
+          properties: {
+            testProperty: {},
+          },
+        },
+      },
+    ],
+    propertyAttributes: [],
+    propertyTables: [],
+  };
+
   const emptyVertexShader =
     "void vertexMain(VertexInput vsInput, inout czm_modelVertexOutput vsOutput) {}";
   const emptyFragmentShader =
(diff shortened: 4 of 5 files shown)


plainheart:
@ggetz @mzschwartz5 Could you please have a look at this?

mzschwartz5:
Hi @plainheart I'll put this on my list to look at (might not be tomorrow but sometime this week)

edit- whoops, hit close by mistake

mzschwartz5:
It's not clear to me that this is a bug. The issue occurs when trying to add a `PrimitiveCollection` to a `PrimitiveCollection` via the `.add` method. Our docs don't indicate that that is supported. If anything, the "bug" is that `PrimitiveCollection.add` is not throwing an error when you try to pass something other than a `Primitive` to it.

Beside adding in such a check / error, if we want to explicitly support passing in entire collections like this, I would:
1. Explicitly document it in the jsdocs
2. Modify `.add` to iterate over the passed-in collection and add each primitive one-by-one. (Otherwise you have to do some recursion in case there are nested-within-nested-... collections. Easier to keep everything flat unless we have a reason not to).

I think the current "fix" has a high risk of introducing some unforeseen regression.

edit- okay, [the docs _do_ actually say you can do this](https://github.com/CesiumGS/cesium/blob/a9680b8fa610fd38340c066afb61e8e802f61fd4/packages/engine/Source/Scene/PrimitiveCollection.js#L8-L11)... but I would not say it's supported in the code. Giving some more thought to the "right" thing to do here.

edit 2- I think the right move is to modify the `contains` check on `PrimitiveCollection`. Right now it does not take into account the potentially "nested" nature of the collection. It only checks for guids on the top level. If `PrimitiveCollection` is truly meant to support nested collections, than `contains` should also support that.

plainheart:
Thanks for your review! I think modifying the `PrimitiveCollection.contains` method is unnecessary at this stage.

The original implementation of `getTilesets` is explicitly designed to recursively flatten tilesets from `scene.primitives`. This logic already accounts for tilesets nested within `PrimitiveCollection`. 

https://github.com/CesiumGS/cesium/blob/8209a5964cdca3f8bfb5ce7d7f14873704cd8332/packages/engine/Source/Scene/Picking.js#L1033-L1049

After this recursive flattening, the tilesets returned by `getTilesets` are already at the top level of the flattened collection, meaning the subsequent check `scene.primitives.contains(tileset)` is redundant and problematic:

https://github.com/CesiumGS/cesium/blob/8209a5964cdca3f8bfb5ce7d7f14873704cd8332/packages/engine/Source/Scene/Picking.js#L1005-L1010

In my PR, I’ve opted to remove this `contains` validation step to restore picking height functionality for tilesets in nested `PrimitiveCollection`.

mzschwartz5:
I think that, because `sampleHeightMostDetailed` is asynchronous, and each frame until it finishes, `updateMostDetailedRayPick` is called, it may be possible for a tileset to be removed from the scene's primitive collection between frames. That's why `scene.primitives.contains(tileset)` is checked each time. 

So removing the check entirely may open up an edge case. Alternatively, maybe this is just a performance optimization (as the comment says, _"Only update tilesets that are still contained in the scene's primitive collection"_). Maybe removing the check wouldn't create a bug, just a possible performance hit. 

If it's just a performance hit, that's probably fine. I can't imagine a large number of tilesets ever being removed from the scene's primitive collection between frames. So we mark a couple extra tilesets for update - who cares.

Can you try to confirm whether this would be a bug or just a performance difference? If it would be a bug, maybe we should be checking `tileset.isDestroyed()` instead.

Issue #13092, opened 2025-12-10
Camera jumps when globe translucency is enabled since 1.136


p-skakun:
### What happened?

I've encountered an issue with the camera jumping when globe translucency is enabled, starting from version 1.136
However, if we disable globe translucency, the camera movement works smoothly.
After investigating, I suspect that this issue might be related to recent changes in picking functionality.
The fix #12999 applied to https://github.com/CesiumGS/cesium/blob/ea5f4d9d2db6e9743ba8658852b4d63e41568685/packages/engine/Source/Scene/ScreenSpaceCameraController.js#L1145 `pickPosition` function (as issue affects not only zoom) helps mitigate the issue, but it also introduces a laggy camera movement due to throttled interactions, so I don't think it's a viable workaround.

### Reproduction steps

[...]

Issue #13083, opened 2025-12-04
Undefined terrain picks in 2D/CV mode.


mzschwartz5:
### What happened?

In 2D / CV modes, many picks are undefined when they shouldn't be. (See comments below for more info - original issue description misattributed issue as a regression from recent PR).

~This comes from the [terrain picking performance PR](https://github.com/CesiumGS/cesium/pull/9961), which now uses an incrementally constructed quadtree to accelerate picks.~

~1. It does not happen when the quadtree is limited to a single (root) node. 
2. The fact that it does not happen in 3D mode tells me the issue is likely related to the computation or use of the 2D transform [here](https://github.com/CesiumGS/cesium/blob/0a49108a5025feb5e2f6a420ccd58919e39061f3/packages/engine/Source/Core/TerrainMesh.js#L257-L313).~

edit- that... no longer seems to be true. Not sure if I'm doing something different now or then, but even if I set the maximum quadtree level to 0, the issue still occurs now.

### Reproduction steps

[...]

Issue #11096, opened 2023-02-21
New Feature but Bug: Add Lambert Conformal  Conic Projection support


zxj5470:
Add `LambertProjection.js` in `Source/Core`
main entry
```js
let mapbox = new Cesium.UrlTemplateImageryProvider({
    url: 'https://a.tiles.mapbox.com/v4/mapbox.satellite/{z}/{x}/{y}.jpg?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4M29iazA2Z2gycXA4N2pmbDZmangifQ.-g_vE53SD2WrJ6tFX7QHmA',
    maximumLevel: 17,
    defaultAlpha: 0,
});

const viewer = new Cesium.Viewer("app", {
    sceneMode: 2,
    mapProjection: new Cesium.LambertProjection({
    }),
    imageryProvider: mapbox
});
```

![image](https://user-images.githubusercontent.com/20026798/220347154-b45667c3-1920-4ef7-9ed5-f0f37c8e6ad9.png)

Maybe I need to be familiar with the implementation of projection matrix transformation in the `Scene.js` file?

Here's LambertProjection file:

```js
import Cartesian3 from "./Cartesian3.js";
import Cartographic from "./Cartographic.js";
import defaultValue from "./defaultValue.js";
import defined from "./defined.js";
import DeveloperError from "./DeveloperError.js";
import Ellipsoid from "./Ellipsoid.js";

const DEG_TO_RAD = Math.PI / 180;
const RAD_TO_DEG = 180 / Math.PI;

function clamp(n, min, max) {
  return Math.min(max, Math.max(min, n));
}

function degToRad(a) {
  return a * DEG_TO_RAD;
}

function radToDeg(a) {
  return a * RAD_TO_DEG;
}

const halfPi = Math.PI / 2;

function tany(y) {
  return Math.tan((halfPi + y) / 2);
}

function LambertProjection(options) {

  const { ellipsoid, centralMeridian, standardParallel1, standardParallel2, falseEasting, falseNorthing } = options;

  this._ellipsoid = defaultValue(ellipsoid, Ellipsoid.WGS84);
  this._semimajorAxis = this._ellipsoid.maximumRadius;
  this._oneOverSemimajorAxis = 1.0 / this._semimajorAxis;

  this._ellipsoid = defaultValue(ellipsoid, Ellipsoid.WGS84);
  this._centralMeridian = centralMeridian || 105;
[...]

PR #13098, opened 2025-12-15
2d picking fixes

mzschwartz5:
# Description

There are various issues with terrain picking in 2D/CV scene modes. Some of these have existed for a very long time, and some are recent regressions from #9961.

All of them manifest as `undefined` pick results when performing `globe.pick`. This PR addresses nearly all of the issues that manifest this way. They're all documented [here](https://github.com/CesiumGS/cesium/issues/13083), but tl;dr-

1. 2D mode has an "Infinite scroll" option. To support this, the terrain gets rendered multiple times in a given frame. Picking relies on testing ALL tiles rendered this frame, but `QuadtreePrimitive` only tracks the tiles rendered in a given render call.
2. Another, related issue- in infinite scroll mode, the map "wraps", and the camera's world position wraps with it, but the ray generated by `camera.pickRay()` does _not_ account for that wrap. 
3. Antimeridian issues: it seems like the vertex data near the antimeridian sometimes suggests is should be rendered on the other end of the map - some sort of disagreement between the position data and where its rendered. Easily fixed with some modulo math.
4. `TerrainPicker` regression 1: `globe.getHeight()` picks in 3D mode, even when the scene mode is 2D/CV. This causes the `TerrainPicker` to cache and use a stale transform for the incorrect scene mode.
5. `TerrainPicker` regression 2: vertex positions were passed to workers as 32 bit floats. This seems to have caused some precision issues in 2D mode (maybe in 3D mode too?), so I switched to `Float64Array`. 

Unfixed issues:
1. Still some undefined picks near the poles in 2D. Couldn't figure this one out in a reasonable time period, but this is a classic issue and one that doesn't historically have a big impact.
2. The unit tests for `TerrainPicker` 2D mode are not good. I don't think the mock terrain provider can even provide data in the 2D coordinate system, but it would be a lot of work to learn how to enhance it.

## Issue number and link

[Primary issue](https://github.com/CesiumGS/cesium/issues/13083)

Also came up [here](https://github.com/CesiumGS/cesium/issues/13092)

## Testing plan

I've been using [this sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html#c=dVTvb9owEP1XbnwKU+SsfASKVtFIRRqjKl37BWlynQOsOnZkO7Ss4n+fHecHYd2XSHf33vO7sy88L5S28BWogTkaXuaw1SqHzYBV0WYw2UimpLFw4PiGGq5B4luNJU9VLmrQcyUt5RL1ZhDDx0YCGIYSlyrDcUNZNxmynqc/09Ft7HE5LXxudNvilk2GPKwebx5TBzsNnZlgg+C7RZlFNTgkQ7CQpkBmlV7ydy6Hrf9SZrh15rJ7xaU1dSNrtNGZrGEaUa4LyjA9oLR3VGbC59EuZFHaG2a5klGEvjiE61loM5zwAYzmqGkMO6FeEE7ujFbXdT3poEIx6pUcopIirpl7ZbjPTSBJ4MaYMudyB2cF4AbsHiFXpUEo6mwnikLwwiie3XP22p0dTJHCJdMGEXkWtDbiEJ6bJVUPpNWsIMOzFjQ9ujNq8R1af+gDPUaNZodVAolQu8gxiNJ8x/9Ty7hG1qfaMMq6oWDJN+Lxcc9v4PAtRF/qR1Ffd9QqDIfhtgLs4j2QPTVRb4IOrtGWWnZPZxLol1SaZRfUSW+g7oLdVWEA1hbAXaAjj6GNXYa/o1jzP25brr7FXZ4poXS7GXMfkYf09gyhSiucofknwOe7RbU9F9Bnntn9GEZt5RR3xsLLGvdfVF0/heZO/lPPp52wq5zibtP7u/R4LJAsV7/W6e/l6in1MoN4MDX2KHDm1b7z8DcqtYgISSzmhaAWTfJSsle0hBnjSdOkoUwzfgCeXX/y/wEmqDGusi1FNdLNYDZNHL5HE4pmbslWB9SCHj1kfzX7EZKEkGniwn9ZVinxQvWZ4l8) to test: it picks as you move the mouse, and draws a point wherever there results are undefined.

After these changes, it should mostly draw nothing (except near the poles)!

# Author checklist

### DIFF
--- a/Documentation/Contributors/ReleaseGuide/README.md
+++ b/Documentation/Contributors/ReleaseGuide/README.md
@@ -25,12 +25,14 @@ There is no one release manager; instead, [our community shares the responsibili
      - If an update is incompatible, open a new issue tagged with the `dependencies` label.
      - If an update can be resolved, commit the changes and open a PR for review.
 
-   > [!IMPORTANT]
-   > We pin an exact version of **`prettier`**. If the version of `prettier` should be incremented, use an exact version by running `npm install prettier@latest --save-exact`.
-   >
-   > If you run `npm install prettier@latest` without the `--save-exact` argument, remove the `^` in `package.json`.
-   >
-   > It's reccomended to commit the package update and a seperate commit for any automated formatting updates to streamline PR reviews.
+> [!IMPORTANT]
+> We pin an exact version of **`prettier`**. If the version of `prettier` should be incremented, use an exact version by running `npm install prettier@latest --save-exact`.
+>
+> If you run `npm install prettier@latest` without the `--save-exact` argument, remove the `^` in `package.json`.
+>
+> It's reccomended to commit the package update and a seperate commit for any automated formatting updates to streamline PR reviews.
+
+<!-- markdownlint-disable MD029 -->
 
 2. **Check the [`priority - next release` issues and PRs](https://github.com/CesiumGS/cesium/labels/priority%20-%20next%20release).** Work with the team to ensure accountability for priority items—This can be done via the "CesiumJS" channel in Teams.
    - Ask the team if there are any items not tagged with the `priority - next release` label which should be.
@@ -40,6 +42,8 @@ There is no one release manager; instead, [our community shares the responsibili
    - Open a PR to address each items scheduled for removal in the next release, if any.
 4. Ensure you've generated valid [end to end testing snapshots](../TestingGuide/README.md) against a previous release tag with `npm run test-e2e-update`.
 5. Start thinking ahead for a good image for the release blog post in case you need to prepare any data, assets, or examples.
+
+<!-- markdownlint-enable MD029 -->
 
 ## Release testing and packaging
 


packages/engine/Source/Scene/Globe.js
```diff
@@ -723,14 +723,7 @@ Globe.prototype.pickWorldCoordinates = function (
...
 
-  const tilesToRender = this._surface._tilesToRender;
-  let length = tilesToRender.length;
-
-  let tile;
-  let i;
-
-  for (i = 0; i < length; ++i) {
-    tile = tilesToRender[i];
+  for (const tile of this._surface._tilesRenderedThisFrame) {
```
jjhembd, line 726: How many iterations are performed for this loop, and how often?

For very large loops, I believe `for..of` is still slower than `for (let i = 0; ...`. Since this could be running every frame, we do need to consider the performance.

References:
- https://dev.to/maafaishal/benchmarking-for-while-forof-and-arrayforeach-using-performancenow-1jjg
- https://waspdev.com/articles/2026-01-01/javascript-for-of-loops-are-actually-fast

In both cases, note the different conclusions depending on the number of iterations.

packages/engine/Source/Scene/QuadtreePrimitive.js
```diff
@@ -256,9 +257,9 @@ QuadtreePrimitive.prototype.forEachLoadedTile = function (tileFunction) {
  *        function is passed a reference to the tile as its only parameter.
  */
 QuadtreePrimitive.prototype.forEachRenderedTile = function (tileFunction) {
-  const tilesRendered = this._tilesToRender;
-  for (let i = 0, len = tilesRendered.length; i < len; ++i) {
-    tileFunction(tilesRendered[i]);
+  const tilesRendered = this._tilesRenderedThisFrame;
+  for (const tile of tilesRendered) {
```
jjhembd, line 261: As above, how many tiles are in this list? If it's a "small" number, then `for..of` is probably ok.

packages/engine/Source/Workers/incrementallyBuildTerrainPicker.js
```diff
@@ -11,6 +11,8 @@ const scratchTrianglePoints = [
   new Cartesian3(),
 ];
 const scratchTriangleAABB = new AxisAlignedBoundingBox();
+const scratchClampMax = new Cartesian3(0.5, 0.5, 0.5);
```
jjhembd: Minor naming nit: we usually use "scratch" to describe some pre-allocated memory that will be overwritten repeatedly (rather than re-allocating it each time). In this case, I think the clamp min/max values stay constant? So simply `const clampMaximum` would be clearer.

packages/engine/Source/Workers/incrementallyBuildTerrainPicker.js
```diff
@@ -102,7 +104,23 @@ function createAABBFromTriangle(inverseTransform, trianglePoints) {
     trianglePoints[2],
   );
 
-  return AxisAlignedBoundingBox.fromPoints(trianglePoints, scratchTriangleAABB);
+  const aabb = AxisAlignedBoundingBox.fromPoints(
+    trianglePoints,
+    scratchTriangleAABB,
+  );
+  Cartesian3.clamp(
```
jjhembd: Can we add a little context in an inline comment? Why do values outside the clamp min/max not make sense? Why do we need to clamp? how would the values go out of range?

packages/engine/Specs/Core/TerrainMeshSpec.js
```diff
@@ -0,0 +1,171 @@
...
+  Cartesian3,
+  Rectangle,
+  BoundingSphere,
+  TerrainEncoding,
+  Ray,
+  SceneMode,
+  GeographicProjection,
+} from "../../index.js";
+
+describe("Core/TerrainMeshSpec", function () {
```
jjhembd, line 12: Always good to see more specs!

packages/engine/Source/Scene/Globe.js
javagl: Benchmarks are good. But _good_ benchmarks are difficult. 

The first one was run on Node - translating that to a browser might be difficult, even though it could technically be "the same engine". The second one is _very_ recent (which is good - you don't want to distract people with some outdated 2025 stuff 👴 ). It claims equal performance for both for small arrays, but doesn't give much detail - it suggests to be run on V8, but not much more. The differences of whether it's an array of numbers or an array of objects are hard to account for, and even less the question of where some JIT complier might have kicked in (and what that JIT may have been doing - I've seen the JVM Hotspot JIT doing the craziest things).

However:

From a quick `console.log`, it looks like the number of tiles here is in the low two-digit range (5 ... 40 at most).

(I didn't check how many of the `this._surface._tilesRenderedThisFrame[i].data.renderedMesh` actually exist, and how many intersections with the `this._surface._tilesRenderedThisFrame[i].data.tileBoundingRegion.boundingSphere` are actually computed there, but my gut feeling is that several dot products and solving a quadratic equation (involving a square root) would outweigh many possible performance differences that could stem from different kinds of for-loops)

mzschwartz5: I think Marco's answer pretty much covers it. The number of tiles is usually pretty low, and likely not the performance bottleneck here.

packages/engine/Source/Scene/QuadtreePrimitive.js
mzschwartz5: Another instance where I don't think it's a hot path for performance, but it's an easy change to make, so might as well change it.

packages/engine/Source/Workers/incrementallyBuildTerrainPicker.js
mzschwartz5: Good point!

packages/engine/Source/Workers/incrementallyBuildTerrainPicker.js
mzschwartz5: I'll try my best 😅 - it's been quite a while and I only vaguely remember why I did this.

mzschwartz5: I believe it has to do with the way the inverse transform of the tile is calculated in 2D. Basically, the s[cale of the regular transform can have a 0 z-component](https://github.com/CesiumGS/cesium/blob/983ea47da0ed02d8145c8c980c0ce2c7661d1909/packages/engine/Source/Core/TerrainMesh.js#L293-L299), which is degenerate when inverted. To prevent that, if the tile is flat, we just set the z-scale to 1.  (I'm using "z" to mean into/out-of the 2D plane)

It "works" but it's not physically accurate, and as a result, when using the inverse transform, the AABBs of triangles can exceeed the z-bounds of the tile AABB. It suffices to simply clamp the triangle AABB to the bounds of the tile AABB.

mzschwartz5: But I'm no longer confident that's the exact reasoning, so I'm not sure if it should be codified in an inline-comment.

mzschwartz5: Decided to just put in this explanation... it may only be partially correct though.

packages/engine/Source/Core/TerrainPicker.js
```diff
@@ -535,6 +541,11 @@ function getVertexPosition(
     result,
   );
 
+  // Right near the antimeridian, in 2D/CV modes, the vertex positions may correspond to the other side of the world from the ray origin.
+  // Compare to the ray origin and adjust accordingly.
+  const worldWidth = CesiumMath.TWO_PI * projection.ellipsoid.maximumRadius;
```
ggetz, line 549: This is going to use a spherical approximation and not the true ellipsoid circumference. Is that adequate for the use case here, or should the true circumference be used?

packages/engine/Specs/Core/TerrainMeshSpec.js
```diff
@@ -0,0 +1,171 @@
...
+        6, // stride
+        undefined, // orientedBoundingBox
+        new TerrainEncoding(center),
+        [],
+        [],
+        [],
+        [],
+      );
+
+      // Mock out the dependency on TerrainPicker
```
ggetz, line 43: Just curious—What's the motivation on mocking this? Is the `TerrainPicker` difficult to construct or is there another source of friction? 

Generally speaking, it'd preferable to call through this function call, even if you are spying on it. (Though this doesn't need to prevent this PR from being merged—Just adding the unit test is already an improvement.)

packages/engine/Source/Core/TerrainPicker.js
mzschwartz5: Good point.... I need to think about this. It does seem to "work" in all my testing but is it correct?

Perhaps the reason it's valid (at least in the testing I've done) is because our two common projections (mercator and web geographic) are cylindrical projection, where this math is exact. 

It might not be valid for other projections.

packages/engine/Specs/Core/TerrainMeshSpec.js
```diff
@@ -0,0 +1,171 @@
...
+      expect(mesh._terrainPicker.rayIntersect).toHaveBeenCalledWith(
+        ray,
+        expected3DTransform,
+        false,
+        SceneMode.SCENE3D,
+        projection,
+      );
+
+      mesh.updateExaggeration(2.0, 1.0);
+      const expected2DTransform = mesh.getTransform(
```
ggetz, line 126: While calculating a value rather than hardcoding it can make tests more resilient, in this case the function computing the expected value is part of the [system under test](https://en.wikipedia.org/wiki/System_under_test). Depending on it may lead to cases where there's a bug somewhere in `TerrainMesh` but this unit test still passes. I'd opt to hard code it, or compute it outside of `TerrainMesh` operations.

(Again, this doesn't need to prevent this PR from being merged—Just adding the unit test is already an improvement.)

packages/engine/Source/Core/TerrainPicker.js
ggetz: Well, conveniently those are the only two 2D projections we support. 😄 So maybe good enough, and just worth a comment explaining why.

If in doubt, using a [root mean square approximation](https://www.johndcook.com/blog/2024/06/19/ellipse-perimeter-approximations/#:~:text=P%20%E2%89%88%202%CF%80(ab)%C2%BD,major%20and%20semi%2Dminor%20axes.) should be accurate enough for most real world applications since the radii are generally close in value. But that could be overkill given the above.

mzschwartz5: Yeah the issue is that `MapProjection` is an interface technically, so people could in theory extend it with their own projection implementations that may not be cylindrical.

In practice this hasn't happened in 15+ years so... (there are plenty of places that assume our two projections are the ONLY projections, for better or for worse).

Will leave a comment though.

packages/engine/Specs/Core/TerrainMeshSpec.js
mzschwartz5: I think my thought process was that these tests aren't supposed to be testing the terrain picker ray intersection logic, they're testing `TerrainMesh`.

Otherwise it's more of an integration test than a unit test. For what it's worth, it is also difficult to construct `TerrainPicker` - fully constructing things would require having real mesh data to give it. Once you start down that rabbit hole, where do you stop?

packages/engine/Source/Core/TerrainPicker.js
javagl: So 15 years was the limit: https://github.com/CesiumGS/cesium/issues/13149 
I could imagine that people would have _liked_ to use other projections, but accepted that this is simply not possible, despite `MapProjection` being an interface.

ggetz:
Testing this out, most of the picking is working great now. 

Besides the poles, there is one remaining issue I see when testing. It's only apparent in CV mode—

In the [testing Sandcastle linked above](http://localhost:8080/Apps/Sandcastle2/index.html#c=dVTvb9owEP1XbnwKU+SsfASKVtFIRRqjKl37BWlynQOsOnZkO7Ss4n+fHecHYd2XSHf33vO7sy88L5S28BWogTkaXuaw1SqHzYBV0WYw2UimpLFw4PiGGq5B4luNJU9VLmrQcyUt5RL1ZhDDx0YCGIYSlyrDcUNZNxmynqc/09Ft7HE5LXxudNvilk2GPKwebx5TBzsNnZlgg+C7RZlFNTgkQ7CQpkBmlV7ydy6Hrf9SZrh15rJ7xaU1dSNrtNGZrGEaUa4LyjA9oLR3VGbC59EuZFHaG2a5klGEvjiE61loM5zwAYzmqGkMO6FeEE7ujFbXdT3poEIx6pUcopIirpl7ZbjPTSBJ4MaYMudyB2cF4AbsHiFXpUEo6mwnikLwwiie3XP22p0dTJHCJdMGEXkWtDbiEJ6bJVUPpNWsIMOzFjQ9ujNq8R1af+gDPUaNZodVAolQu8gxiNJ8x/9Ty7hG1qfaMMq6oWDJN+Lxcc9v4PAtRF/qR1Ffd9QqDIfhtgLs4j2QPTVRb4IOrtGWWnZPZxLol1SaZRfUSW+g7oLdVWEA1hbAXaAjj6GNXYa/o1jzP25brr7FXZ4poXS7GXMfkYf09gyhSiucofknwOe7RbU9F9Bnntn9GEZt5RR3xsLLGvdfVF0/heZO/lPPp52wq5zibtP7u/R4LJAsV7/W6e/l6in1MoN4MDX2KHDm1b7z8DcqtYgISSzmhaAWTfJSsle0hBnjSdOkoUwzfgCeXX/y/wEmqDGusi1FNdLNYDZNHL5HE4pmbslWB9SCHj1kfzX7EZKEkGniwn9ZVinxQvWZ4l8),
1. Set the map mode to CV in the map mode picker.
2. Zoom out a bit so that the edges of the projected map are visible.
3. Mouse over the empty space on the left side of the map. As expected, red markers show up on the right side of the map indicating `undefined` values returned from the pick.
4. Mouse over the right side of the map (_the map itself, not the empty space_) in the same areas where where the markers appeared. Unexpectedly, more red markers show up, indicating `undefined` values returned from the pick.

The above also happens when repeating the steps above but mirroring left and right.

image

This doesn't need to hold up the PR if this can be handled as a separate fix as there are already improvements here.

ggetz:
In the interest of moving forward with the release, I'm going to merge this branch as-is and file and issue for the outstanding item in my last comment — https://github.com/CesiumGS/cesium/issues/13180

packages/engine/Source/Core/TerrainPicker.js
mickae1: I think you should remove the ability to customize the projection, if it's not fully supported.

mickae1: and there are other ticket before mine, the limit was not 15 years : 
2014: https://github.com/CesiumGS/cesium/issues/2084
2016 : https://github.com/CesiumGS/cesium/issues/3877
2023: https://github.com/CesiumGS/cesium/issues/11096

the lambert projection is the one that offer the least deformation and used a lot.

@javagl @mzschwartz5 @ggetz

PR #13101, opened 2025-12-16
refactor(core-math): Convert to ES6 classes

jdehorty:
## Summary

Convert all constructor function patterns in `@cesium/core-math` to ES6 class syntax, following the approach used by Svelte in [PR #8569](https://github.com/sveltejs/svelte/pull/8569) for their TypeScript to JSDoc migration.

### Changes

- **Cartesian2.js** - Convert to ES6 class with static/instance methods
- **Cartesian3.js** - Convert to ES6 class with static/instance methods  
- **Cartesian4.js** - Convert to ES6 class with static/instance methods
- **Matrix2.js** - Convert to ES6 class with static/instance methods
- **Matrix3.js** - Convert to ES6 class with static/instance methods
- **Matrix4.js** - Convert to ES6 class with static/instance methods
- **createPackableSpecs.js** - Update imports to use `@cesium/core-utils` and `@cesium/core-math` packages

### Conversion Pattern

**Before:**
```javascript
function Cartesian2(x, y) {
  this.x = x ?? 0.0;
  this.y = y ?? 0.0;
}

Cartesian2.clone = function (cartesian, result) { ... };
Cartesian2.prototype.toString = function () { ... };
```

**After:**
```javascript
class Cartesian2 {
  constructor(x, y) {
    this.x = x ?? 0.0;
    this.y = y ?? 0.0;
  }

  static clone(cartesian, result) { ... }
  toString() { ... }
}
```

### Notes

- JSDoc `@constructor` tags updated to `@class`
- Scratch variables remain at module level (after class) with `eslint-disable no-use-before-define`
- Static properties like `packedLength` moved outside class body for ESLint parser compatibility
- Method aliases (e.g., `fromArray = unpack`) remain outside class
- All debug pragmas preserved unchanged
- **Math.js** not converted (namespace object pattern, not a constructor)

## Test plan

- [x] Build passes (`npm run build`)
- [x] ESLint passes (pre-commit hooks)
- [ ] Full test suite passes
- [ ] TypeScript definitions generate correctly

PR #13104, opened 2025-12-17
Types: Enable incremental JSDoc type checking

donmccurdy:
# Description

**Summary**

Enables IDEs with [TypeScript language server](https://github.com/typescript-language-server/typescript-language-server) support (such as VSCode) to infer types from JSDoc and detect type errors automatically. The PR is 'minimal' in the sense that it does not require migrating code into subpackages, and does not require replacing tsd-jsdoc. Either of those larger changes might still be preferable for other reasons, certainly. Adoption of TypeScript checks could be entirely incremental with this approach; we can turn type-checking on one file at a time.

**Changes**

- Adds `packages/engine/tsdoc.json`, with type-checking disabled for all JS files by default. Individual JS files may opt-in to type checking with a `// @ts-check` comment at top of file.
- For TypeScript to resolve type-only imports (for example, Cartesian2 uses Cartesian3 in its parameter types, but does not import the file), we use special comments like `/** @import Cartesian3 from './Cartesian3.js'; */`. TypeScript understands these, but  JSDoc errors on them, so we update the JSDoc configuration to ignore `@import` tags. Huge thanks to @jjspace for this solution!
- Configures ESLint for ES6 class conversion. Scratch variables can be used by the same class they instantiate, which is fine at runtime but needs a workaround in ESLint.
- Configures CI to run `tsc` and check types, for files that we have opted in.
- Added documentation to the coding guide ([preview](https://github.com/CesiumGS/cesium/blob/refactor/minimal-tsc/Documentation/Contributors/CodingGuide/README.md#type-checking))

**Background**

Minimal subset of work from 12/16/2025 hackathon day, including work from @jjhembd, @jdehorty, @jjspace, and myself. Other tracks explored in the hackathon — including JSDoc-TS compatibility conversion with Gemini, smaller package reorganization, build system updates, documentation generation with typedoc, , etc. — are on the `core-math` branch. We'll likely want to use some of those results too, but this PR might be the minimum-viable starting point.

## Issue number and link

- #8359
- #4434 

## Testing plan

Run `npm run tsc`, and observe no errors reported. Add `this.x = 'hello world'` to the Cartesian2 constructor, run `npm run tsc` again, and see expected error output:

```
packages/engine/Source/Core/Cartesian2.js:32:5 - error TS2322: Type 'string' is not assignable to type 'number'.

32     this.x = "hello world";
       ~~~~~~

Found 1 error in packages/engine/Source/Core/Cartesian2.js:32
```

The same error should appear in your editor/IDE, without explicitly running `tsc`, if the editor supports the TypeScript Language Server.

# Author checklist

#### PR Dependency Tree

* **PR #13104** 👈
  * **PR #13125**
  * **PR #13126**

### DIFF
--- a/Apps/Sandcastle/CesiumSandcastle.js
+++ b/Apps/Sandcastle/CesiumSandcastle.js
@@ -26,6 +26,7 @@ require({
       location: "../Apps/Sandcastle/ThirdParty",
     },
   ],
+  //@ts-expect-error TS thinks this is using the native require when it's not
 }, [
   "CodeMirror/lib/codemirror",
   "dijit/Dialog",


package.json
```diff
@@ -122,6 +122,7 @@
     "build-docs": "gulp buildDocs",
     "build-docs-watch": "gulp buildDocsWatch",
     "eslint": "eslint \"./**/*.*js\" \"./**/*.*ts*\" \"./**/*.html\" --cache --quiet",
+    "tsc": "gulp tsc",
```
jjspace, line 125: Can we add this as part of our `lint-staged` config so it hopefully catches things before CI?

packages/engine/Source/Core/Cartesian2.js
```diff
@@ -1,8 +1,13 @@
+// @ts-check
+/* eslint-disable no-use-before-define */
```
jjspace: I think we should really try and avoid disabling rules for entire files like this. I understand this is because of a sort of "cyclic" definition problem here. I _think_ doing something like this at the top of the file might be cleaner. 

```js

/* eslint-disable no-use-before-define */
const distanceScratch = new Cartesian2();
const lerpScratch = new Cartesian2();
const angleBetweenScratch = new Cartesian2();
const angleBetweenScratch2 = new Cartesian2();
const mostOrthogonalAxisScratch = new Cartesian2();
/* eslint-enable no-use-before-define */
```

packages/engine/Source/Core/Cartesian2.js
```diff
@@ -1,8 +1,13 @@
...
 import Check from "./Check.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 import CesiumMath from "./Math.js";
 
+/** @import Cartesian3 from './Cartesian3.js'; */
+/** @import Cartesian4 from './Cartesian4.js'; */
```
jjspace: This syntax also works like this:

```js

/**
 * @import Cartesian3 from './Cartesian3.js';
 * @import Cartesian4 from './Cartesian4.js';
 */
```

I'm not completely sure which I prefer but I feel like we should make a suggestion in the docs.

packages/engine/Source/Core/Cartesian2.js
```diff
@@ -1,8 +1,13 @@
+// @ts-check
```
jjspace: - What do you think the plan should be for adding this? Is this something we should recommend doing as part of every PR to add it to the files you touch and spend the extra time to fix errors then?
- Do we have a plan for removing all these comments and turning checking on by default? 

Definitely open for discussion but we should try and figure that out before starting so it doesn't get out of hand or forgotten.

packages/engine/Source/Core/Cartesian2.js
```diff
@@ -15,726 +20,776 @@ import CesiumMath from "./Math.js";
  * @see Cartesian4
  * @see Packable
  */
-function Cartesian2(x, y) {
+class Cartesian2 {
```
jjspace: Is the class conversion required for turning on typechecking? I assume sometimes yes and sometimes no. Can you include the process of running `lebab` in the documentation if that's the plan?
This also can be quite distracting in a PR from the "real" changes. Is the recommendation to try and do this on a larger scale? or in chunks alongside other PRs? 
are there processes that can be documented for validating this to make sure behavior remains the same? The specs should cover most of it but sometimes they don't fully cover everything.

jjspace (commented):
Thanks for this @donmccurdy! Overall I'm _hugely_ in favor of this change. It should let us get more out of our intellisense and start catching some type errors even before switching fully to TS.

Can you please add a section to our documentation pages about this? Probably part of the Contribution guides, maybe the `CodingGuide`? I also had a few more comments to help guide what gets documented.

packages/engine/Source/Core/Cartesian2.js
jjspace: I wonder if an approach we can take is to convert a chunk of files the first week of every month right after the release? Then we can merge that and hopefully catch any mistakes while we develop the rest of the month? It could probably be added to the release rotation

packages/engine/Source/Core/Cartesian2.js
donmccurdy: Unfortunately scratch variables must be instantiated after the class body or we'll get a runtime error:

> Uncaught TypeError: Cartesian2 is not a constructor

With the scratch variables after the class body, ESLint's use-before-define errors are about _references to_ the scratch variables. We could disable the ESLint check for those references, but they're scattered across ~5 methods in the Cartesian2 class body here, and might get messier as we repeat that in other files. 😕 



Perhaps a compromise could be to disable the [no-use-before-define 'variables' option](https://eslint.org/docs/latest/rules/no-use-before-define#options) for ESLint, either globally or at the top of each file?

```javascript
/* eslint no-use-before-define: ["error", { "variables": false }] */
```

This will still catch use-before-define errors for variables in the same or lower scopes; it only relaxes the rule for references to variables in an upper scope, if I understand the option correctly.

package.json
donmccurdy: It looks like we'll have to create a new `lint-staged.config.js` file for the `packages/engine` directory, but yes!

https://github.com/lint-staged/lint-staged?tab=readme-ov-file#how-can-i-resolve-typescript-tsc-ignoring-tsconfigjson-when-lint-staged-runs-via-husky-hooks

packages/engine/Source/Core/Cartesian2.js
javagl: I might be missing something, but think that this is somehow related to https://github.com/CesiumGS/cesium/pull/13101#issuecomment-3662164513 , and maybe someone can confirm quickly: Is it right that
```typescript
class Cartesian2 {
  /** The scratch variable used in foo() and bar() */
  /* private */ static /* readonly */ someScratch = freezeAndWhatnot(new Cartesian2());
  ..
}
```
should be "The Way To Go®", **iff** there wasn't some part of the toolchain that complained about that (for reasons that are ~"not really technically profound")?

packages/engine/Source/Core/Cartesian2.js
donmccurdy: I definitely want to defer to the team's preference here! But my feeling is:

- Adding `// @ts-check` to files modified in everyday PRs is welcome but not required. Some files may be trickier than others to satisfy the TS compiler, and TS changes shouldn't obscure the PR's original changes.
- After a majority of JS files have checks enabled, a bulk PR enabling the option by defaulting and opting-out remaining files would make sense
- Batch updates at the top of the month could be a good option to make progress faster if everyone feels good about that. Otherwise we could wait and see how progress goes naturally.

Either way, it might be nice to have some easy way to check how many files are missing the `// @ts-check` header so we can check progress over time. On macOS/Linux, something like:

```bash
grep -L "@ts-check" Source/**/*.js | wc -l
```

packages/engine/Source/Core/Cartesian2.js
donmccurdy: I used...

```bash
npx lebab packages/engine/Source/Core/Cartesian2.js \
    -o packages/engine/Source/Core/Cartesian2.js --transform class
```

The ES6 class conversions aren't strictly required, but without them we'd need to silence or work around TS errors like this one:

```javascript
// ❌ Property equalsArray is private and only accessible within class Cartesian2. (ts 2341)
/**
 * @private
 */
Cartesian2.equalsArray = function (cartesian, array, offset) {
  return cartesian.x === array[offset] && cartesian.y === array[offset + 1];
};
```

Perhaps it would make sense to try to do the ES6 class conversions — with or without opting into TS checks — in bigger chunks up front, rather than incrementally? I agree this could be distracting if mixed into unrelated PRs.

packages/engine/Source/Core/Cartesian2.js
donmccurdy: My general feeling is that the code is fine as-is, and that ESLint is basically wrong. So I'd somewhat rather tell ESLint to get out of the way rather than move scratch variables into static members. 😅 But 100% happy to take whatever approach others prefer here, no strong preference on my side.

jjspace: `class` definitions are [hoisted](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) so I thought re-arranging would be fine. I was, of course, forgetting that `const` are also treated as hoisted the same as `class` so the order does still matter.

I could see them being `static` _private_ members of the class itself and the problem should go away. However I also think they're fine external as they currently are and just telling eslint that we acknowledge the "problems".

That said I'm still pretty strongly against disabling rules for entire files in comments like you have it. Even if it's more verbose I think it'd be better to have an `eslint-disable-next-line` for each use case so it's clear what and why we're disabling the rule. As it is it's really easy to miss at the top of the file and would prevent catching other, actually valid, warnings.

If you think it's going to be so common that we need to be doing it everywhere then I'd prefer it go into the actual eslint config for the whole project so it's consistently applied and expected. This change could also be made later after we have more anecdotal evidence

CHANGES.md
```diff
@@ -8,6 +8,10 @@
 
 - Fixes label positioning in workflows that delete and recreate clamped labels [#12949](https://github.com/CesiumGS/cesium/issues/12949)
 
+#### Additions :tada:
+
+- Begins incremental TypeScript-in-JSDoc type checking adoption [#13104](https://github.com/CesiumGS/cesium/pull/13104)
```
jjspace: I don't think we normally document non-functional changes in the `CHANGES.md` file. The changes here _shouldn't_ affect anyone downstream correct?

Thoughts @ggetz

packages/engine/Source/Core/Cartesian2.js
donmccurdy: I believe reconfiguring the rule to ignore higher-scope variables (not disabling the rule) is safe enough to apply to entire files. That said, the number of classes that rely on scratch instances of themselves appears to be pretty small. A comment on each scratch variable reference seems manageable too, just a bit of clutter in these files.

CHANGES.md
donmccurdy: Ok, makes sense! Conceivably downstream users could be affected by:

1. Adoption of ES6 classes
2. Changes to JSDoc types made to satisfy `tsc` checks

Hopefully (1) isn't an issue, because this isn't the first use of ES6 Classes in the library. And this particular PR didn't really need to do (2), though similar PRs or other files may. In some other work I might have listed this PR under a "Chores" section, but if the policy is not to include chores in the changelog (which sounds entirely reasonable too!) then I'd be inclined to just remove this from CHANGES.md yes.

packages/engine/Source/Core/Cartesian2.js
javagl: Not to distract people who know more about this (and seem to be converging to a solution), but ...

> rather than move scratch variables into static members. 

Is that "bad" in any way? 
<sup>(You know, I'm that Java guy - everything belongs into neat <strike>boxes</strike> classes).</sup>

I'm specifically asking because some aspects here require a deeeep and profound knowledge about what the virtual machine is actually _doing_ with all that, and a very specific point came up recently: When there are multiple workers, where and how _exactly_ are "scratch variables" duplicated? 

(These scratch variables are highly prone to race conditions in general, and I think that for any longer-term prototype->class transition, we should be careful with that - maybe it's not an issue, but it's hard to be _sure_ about that...)

donmccurdy: The (mild) downside I see to static members is some churn, and needing to explain different code styles for files with and without classes.

I don't personally see much downside to the current scratch variables either, though. Workers do not share module scopes, so that's OK. The ES6 class migration doesn't really change anything except unfortunately falling into a limitation in ESLint. But it's true that scratch variable usage needs to be scoped to a single method body, and not overwritten deeper in the call stack. This requires some care, but is nevertheless a fairly common pattern for hot JavaScript code. Has this been an issue for CesiumJS development in the past?

In my mind the "problem to be solved" is that ESLint doesn't understand an otherwise reasonable coding pattern. If there are particular concerns with module-scoped scratch variables though, perhaps we want to tackle that in a separate issue or PR?

javagl: > Workers do not share module scopes, so that's OK. The ES6 class migration doesn't really change anything ...

That's where my (albeit vague) concern is: Imagine some
```
const scratch = new Cartesian3();
function run() { scratch.x = 123; }
```
is converted into some
```
class Example {
  static scratch = new Cartesian3();
  static run() { Example.scratch.x = 123; }
}
```
Is this still "the same" in terms of scoping rules? When there are 5 workers that previously called the `run()` function, then (so I've been told) they magically receive different copies of that scratch variable. Is this still the case when they call the `Example.run()` in the future? 

> Has this been an issue for CesiumJS development in the past?

Maybe only/mainly for me ...? I occasionally stumbled over the case where an `Iterable` returned a value, which happened to be a scratch variable further down the call stack, and then some `const array = [ ... iterable ];` contained the same scratch variable many times. 

(Beyond that, I'm sure that some scratch-related race conditions are lurking in the code, but nothing that I could point my finger at (obviously - otherwise, I'd already have fixed it...))

donmccurdy: > Is this still "the same" in terms of scoping rules?

Yes, for better or worse, it's not possible to share an object reference between workers in JavaScript. The exceptions are explicit APIs for [transferrables](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects#transferring_objects_between_threads) and [atomics](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics). JavaScript's multi-threading model is much more limited than (my old recollections of) Java's.

> Maybe only/mainly for me ...? I occasionally stumbled over the case where an Iterable returned a value, which happened to be a scratch variable further down the call stack...

Ah, true, that's another case. Whether the scratch variables are at module scope or static members of the class, it is probably best not to return them from methods.

I guess the alternative (to avoid all that) would be [object pooling](https://en.wikipedia.org/wiki/Object_pool_pattern), but it's a bigger change and less clear to me what the implications would be for performance in hot functions. I haven't seen the pattern used all that often in JS code, for whatever reason.

javagl: > it is probably best not to return them from methods.

Ideally, scratch variables should never "escape" a method. More than that: Even using _the same_ scratch variable in _more than one_ method is highly error prone. It's nearly impossible to ensure (during refactorings or when adding new functionalities) that _none_ of the methods is _ever_ directly or indirectly(!) calling _any_ other method. (Corollary: Scratch-variables should be "function-scoped" - and doing that would defeat their purpose 😬 )

CHANGES.md
donmccurdy: On reflection, yeah, maybe easier just to remove this unless/until we have reason to think it impacts downstream users, and things we want to say about that.

packages/engine/Source/Core/Cartesian2.js
donmccurdy: Added a section to the coding guide - let's continue discussion there as needed!

packages/engine/Source/Core/Cartesian2.js
donmccurdy: Added to docs!

packages/engine/Source/Core/Cartesian2.js
donmccurdy: Added an example to the docs with lebab, and some language about adding type checking in unrelated PRs being ok as long as it doesn't obscure the functional changes.

packages/engine/Source/Core/Cartesian2.js
donmccurdy: I think I have this resolved with the changes in `eslint.config.js`, essentially a narrow override for variables in upper scopes. With that I've confirmed that ES6 Class conversion can be applied more-or-less automatically (e.g. with `lebab`) without running into these errors. ESLint will still catch errors like this:

```javascript
console.log(xyz);
const xyz = 2;
```

donmccurdy:
@jjspace I think all feedback on the PR has been addressed! Draft documentation in the coding guide:

https://github.com/CesiumGS/cesium/blob/refactor/minimal-tsc/Documentation/Contributors/CodingGuide/README.md#type-checking


I've moved the `Cartesian2.js` changes to a separate PR:

- #13126

I also attempted a bulk conversion to ES6 classes for `packages/engine/Source/Core/*.js`:

- #13125

That's too large to review and merge in bulk, for sure, but still it's encouraging to see tests passing after excluding a handful of files from the conversion.

Documentation/Contributors/CodingGuide/README.md
```diff
@@ -172,6 +173,34 @@ try {
 /*eslint-enable no-empty*/
 ```
 
+## Type Checking
+
+We are incrementally adopting [type checking for JavaScript files](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html). As of January 2026, type checks are enabled on a file-by-file basis, with `// @ts-check` annotations at the top of a file used to opt in. As adoption progresses, we expect this pattern will eventually be flipped to an opt-opt annotation instead. To see type system hints and errors in some editors or IDEs, you may need to configure or install TypeScript language server support. When in doubt, VSCode supports TypeScript language services [by default](https://code.visualstudio.com/docs/nodejs/working-with-javascript).
```
jjspace: Typo
```suggestion
We are incrementally adopting [type checking for JavaScript files](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html). As of January 2026, type checks are enabled on a file-by-file basis, with `// @ts-check` annotations at the top of a file used to opt in. As adoption progresses, we expect this pattern will eventually be flipped to an opt-out annotation instead. To see type system hints and errors in some editors or IDEs, you may need to configure or install TypeScript language server support. When in doubt, VSCode supports TypeScript language services [by default](https://code.visualstudio.com/docs/nodejs/working-with-javascript).
```

Documentation/Contributors/CodingGuide/README.md
```diff
@@ -172,6 +173,34 @@ try {
...
+
+**References:**
+
+For developers already familiar with TypeScript, writing JavaScript with JSDoc type annotations has some differences. For a summary of the annotation syntax, and supported features in JSDoc vs. TSDoc, see:
+
+- [Type Checking JavaScript Files \| TypeScript](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html)
+- [JSDoc Reference \| TypeScript](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html)
+- [JSDoc Cheatsheet and Type Safety Tricks](https://docs.joshuatz.com/cheatsheets/js/jsdoc/) by Joshua Tzucker
```
jjspace, line 187: I think we need to be careful about links like these. TS has sort of expanded or co-opted JSDoc for extra functionality that is _not_ part of JSDoc itself. Our doc generation still uses pure JSDoc so we are beholden to only features it supports. The `@import` tag is helpful but I feel like that might be the only exception we want to make until our doc generation is updated.

For example in the [TS JSDoc Reference](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#satisfies) it mentions the `@staisfies` tag. This is not a JSDoc tag and adding it will result in an error `ERROR: The @satisfies tag is not a known tag.`. I didn't test the others but I'm guessing there are other inconsistencies like this.

Also the very first recommendation in the Cheatsheet shows how to typeguard functions using `@returns {value is TYPE}`. This is also not recognized by JSDoc and results in an error like:
```
ERROR: Unable to parse a tag's type expression for source file /home/[user]/Programming/cesium/packages/engine/Source/Core/defined.js in line 28 with tag title "return" and text "{obj is Person}": Invalid type expression "obj is Person": Expected "|" but "i" found.
```

I bet we could make more edits to our JSDoc config/generation to adapt for more tags but I'm not sure the level of effort there. I mostly bring it up as a word of caution. Maybe all it needs is a warning that our doc gen still needs to work and to be careful. Or a comment that we need to restrict to only official JSDoc tags?  Happy to discuss

Documentation/Contributors/CodingGuide/README.md
```diff
@@ -172,6 +173,34 @@ try {
...
+4. **Prefer `unknown` or `@ts-expect-error` over `any`.** Casting types to `any` forces them to be accepted everywhere, and effectively disables all type-checking dependent on that type. This tends to propagate throughout a type system, reducing the effectiveness of type checks, and should be avoided when possible. Prefer casting to `unknown`, or using a `@ts-expect-error` annotation. For example, when defining an object with string keys, with values whose types we don't care about, prefer `Record<string, unknown>` to `Record<string, any>`.
+5. **Type-only imports.** When JSDoc annotations depend on types not otherwise imported in a source file, it will be necessary to tell TypeScript where to find them. To avoid otherwise-unused imports, use type-only imports in separate one-line JSDoc comments:
```
jjspace, line 201: The very first google search for "import type in JSDoc" points to an older syntax that still works for the TS side of things
```js
/**
 * @typedef {import('./File1.js').MyObject1} MyObject1
 */
```
However this is _not_ valid JSDoc and will break things. I think it's worth calling out here to _not_ use this syntax and instead always rely _only_ on the `@import` tag.

Documentation/Contributors/CodingGuide/README.md
```diff
@@ -172,6 +173,34 @@ try {
...
+2. **Common JavaScript/JSDoc obstacles.** JSDoc-based type checks have limited support for some coding patterns used in the CesiumJS codebase, such as classic prototype-based inheritance and `Object.defineProperties()`. In most cases these problems can be solved by refactoring with newer coding patterns, by hand or with tooling like [lebab](https://github.com/lebab/lebab). In other cases the solution may be less obvious, and a temporary skip-check annotation may be used to satisfy the type system. Example with lebab:
+   ```bash
+   npx lebab packages/engine/Source/Core/Cartesian2.js \
+     -o packages/engine/Source/Core/Cartesian2.js --transform class
+   ```
+3. **Prefer `@ts-expect-error` over `@ts-ignore`.** Sometimes a source file's types are _mostly_ consistent, but a few stubborn errors remain unsolved. These may be blocked by out-of-scope work, or may simply be unjustifiably difficult to solve in JSDoc-based types. Use `@ts-expect-error` annotations to relax the type system for specific lines. Unlike `@ts-ignore`, `@ts-expect-error` will raise errors when the line no longer contains an error, making it easier to clean up annotations later on.
```
jjspace, line 199: I like this suggestion for forward compatibility and cleanup. (a quick search led [to articles](https://dev.to/maafaishal/ts-expect-error-over-ts-ignore-in-typescript-5140) that agree too)

Can we expand this to always include a reason with the disable comment? 

It also could be good to look at enforcing this programmatically in eslint. We're already using `typescript-eslint` for Sandcastle but maybe it's easy to expand to engine files too? It looks like there's a [`ban-ts-comment` rule](https://typescript-eslint.io/rules/ban-ts-comment/) for this to ban `ts-ignore` and `require-reason` for `expect-error` comments.

ggetz:
This PR is looking great to me!

While it doesn't need to hold up merging what we have here, I wonder if there are any TS best practices recommended by the Cesium ion team or the iTwin.js team—Maybe @tfili @angrycat9000 or @markschlosseratbentley could point us to any relevant guides?

markschlosseratbentley:
@jjspace @donmccurdy @ggetz 

> Maybe @tfili @angrycat9000 or @markschlosseratbentley could point us to any relevant guides?

FWIW; from iTwin.js:

- [iTwin.js TypeScript Coding Guidelines](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/guidelines/typescript-coding-guidelines.md)
- [iTwin.js Documentation Linking Syntax](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/guidelines/documentation-link-syntax.md)
- [iTwin.js Release Tags Guidelines](https://github.com/iTwin/itwinjs-core/blob/master/docs/learning/guidelines/release-tags-guidelines.md)

Containing directory for guides [found here](https://github.com/iTwin/itwinjs-core/tree/master/docs/learning/guidelines).

cc @aruniverse @ben-polinsky

angrycat9000:
The [ion Typescript guide is here](https://github.com/CesiumGS/cesium-ion/tree/main/Documentation/TypeScript).

The one thing that we haven't documented (but probably should) is guidance on structuring types so that it forces the caller to pass the right type of arguments.  

Eg  Take a function that can take 3 types of objects as an input, each with their own properties and a discriminator field of `type`.

You could write the type like this:

```ts
{
  type: "A"|"B"|"C"
  a?: number;
  b?: number';
  c? : number;
}
```

Or tihs

```ts
{ type: "A", a: number } |
{ type: "B", b: number } |
{ type: "C", c: number }
````

The latter forces the user to pass the correct values.  However in some cases it requires a bit of extra work to convince typscript that you have gotten it correct when using it.

ben-polinsky:
@angrycat9000 What's the `requires a bit of extra work` you're referring to here? 
For your case, the second example like a much tighter contract.

angrycat9000:
> @angrycat9000 What's the `requires a bit of extra work` you're referring to here? For your case, the second example like a much tighter contract.


Typescript complains when destructuring option like `const {type, a, b,c} = options;`  Even if I don't actually use `a`, `b`, or `c` before checking `type`.  Run into this a bit in React where I'm typically destructring in the delcaration like `function Component({type, a, b,c }) {` 

If `type` is an enum and I declare `{ type: MyType.A, a:1}` it types it as `{ type: MyType, a: number}` which doesn't match `{type: MyType.A, a:number}`.  I need to remember to do `{type: MyType.A as const, a: 5}`.  

Those are ones that I commonly run into and have to work around.

javagl:
It's dangerous to try and come up with "solutions" for a specific typing problem, solely based on an artificial example that was made to illustrate a point (I.e. I might be _missing_ that exact point).

That said: I use enums sparingly. Encoding a _type_ in an enum locks down the set of available types _forever_ (i.e. until a "breaking change"). Passing a type and an object of that type to a function _might_ hint at something that ~"could be solved differently (without causing typing trouble)". Is it possible to illustrate this point with a _slightly_ less artificial example (maybe in a TS Playground)? (I'm just curious, and it might be too specific, so don't waste time if you think that it's not worth the effort).

Documentation/Contributors/CodingGuide/README.md
donmccurdy: Sounds good, added a note — I think @jjhembd had preferred the `@import` syntax for ergonomics/readibility too.

Documentation/Contributors/CodingGuide/README.md
donmccurdy: I think ban-ts-comment is an easy one, so I've gone ahead with adding that rule, agreed! I'm hesitant to add more exhaustive ESLint+TS configuration in this PR, though, as I think that could turn into a larger discussion that we're probably not ready to have without beginning to migrate a few files.

Documentation/Contributors/CodingGuide/README.md
donmccurdy: Hmm that's a footgun, yeah. :(

I've added a few words of caution in the docs about this. A few years back I was writing a lot of type-checked JSDoc without TypeScript (based on Closure Compiler instead). Hadn't realized that Closure _also_ has its own divergent syntax from JSDoc...

Agreed on sticking to official JSDoc tags for the time being. If/when we have a longer-term plan related to tsd-jsdoc we could decide whether it makes sense to start using TS-flavored JSDoc syntax too.

angrycat9000:
> It's dangerous to try and come up with "solutions" for a specific typing problem, solely based on an artificial example that was made to illustrate a point

My main point was that there is a trade off between having types enforcing constraints and developer effort.  Not to call out any specific typing issue.  If you are writing something that is consumed by others then it is probably important to take the extra time to make sure that the typing reflects the constraints instead of relying on the consuming developer knowing what needs to be passed to the API via some other source.

donmccurdy:
Because CesiumJS already has extensive JSDoc type declarations used to generate documentation — we're making existing types _stricter_ to satisfy TS checks, without porting code from JS to TS — I'm hoping that existing coding patterns don't need to change immediately as a result of this PR. Needing to switch to ES6 Classes is the known exception, and the biggest hurdle. If there's more we _want_ to change, especially if TS checks make that easier longer-term, let's make sure we have issues to track that!

To generalize @angrycat9000's point further, when there's more than one way to type something, the stricter option is usually preferable once a codebase (and its types) have reached a certain level of stability and maturity, with some balance of ergonomics and legibility. Related, I haven't enabled [`strict` mode](https://www.typescriptlang.org/tsconfig/#strict) in this PR. It's probably something we'd want to consider longer-term. But in the meantime I think it would make adoption less incremental and more painful. And there are some subjective decisions involved that would probably benefit from the experience we'll collectively get during initial adoption. For example, just personally, I prefer to relax strict null checks in unit tests.

The Ion TypeScript documentation's section on "Porting Existing Node.js JavaScript Code" might be especially relevant to what we're doing here. Helpful to see that iTwin.js has essentially the same warning about not repeating type information that TypeScript can already infer... I'll work on adapting some of these recommendations to the documentation section in this PR as well. Thank you @markschlosseratbentley and @angrycat9000!

Documentation/Contributors/CodingGuide/README.md
donmccurdy: I went one step further and enabled `noImplicitAny` by default. I don't think we want _all_ the strict mode options yet — adoption will be much harder — but this prevents us from having self-propagating chains of untyped code in otherwise-typed files, and that's probably something we'd want to see while upgrading any particular source file.

ben-polinsky:
@donmccurdy not sure if it belongs here, but perhaps one of the stickiest situations we've found as a public TS library is defensively coding against breaking changes. For instance, a simple case is when typing function arguments - an options object is often more resilient than a list of params since you can later make some properties optional without worrying about order. 

More on [breaking TS changes here](https://www.semver-ts.org/formal-spec/2-breaking-changes.html). Might be useful to review if you haven't seen it.

donmccurdy:
@ben-polinsky just curious, are you using [API Extractor](https://api-extractor.com) or something similar? Eventually CesiumJS will probably need to [remove or replace `tsd-jsdoc`](https://github.com/CesiumGS/cesium/issues/10455) for generating documentation, and API Extractor might be a contender... It both generates API documentation from types, and detects breaking changes in those types.

ben-polinsky:
Yes, using API Extractor. Example of changes for PR review: https://github.com/iTwin/itwinjs-core/pull/8888/changes#diff-fd4a69635139b37c9a38b94c5ad4da9d2e26666beda0a470419bdbf470502a53

javagl:
The API extractor can be _very_ useful to detect breaking changes. With two caveats: 

- It detects _changes_. Deciding whether this really is a _breaking_ change is left to the developer
- It does not directly help with _avoiding_ breaking changes to begin with - but it certainly raises awareness about how to "design defensively", and it can be a nice safeguard against accidental breaking changes

(Disclaimer: I've only been using it in the `3d-tiles-tools` and the `3d-tiles-validator`, and these do not have any "official" public API (everything is tagged as `@internal`), so I've certainly not been using it as rigorously as it _should_ be used...)

donmccurdy:
I've updated the coding guide to recommend not duplicating inferred types, unless the JSDoc is needed for documentation of public APIs, as mentioned in the ion and itwin.js guidelines:

https://github.com/CesiumGS/cesium/blob/refactor/minimal-tsc/Documentation/Contributors/CodingGuide/README.md#type-checking

With that, I believe all current feedback on the PR is addressed.

PR #13002, opened 2025-10-27
EXT_mesh_primitive_edge_visibility's material and line string

danielzhong:
# Description

This is a follow up PR to https://github.com/CesiumGS/cesium/pull/12859. Primarily implementing the missing functionality for the material and line string section of `EXT_mesh_primitive_edge_visibility`:

- Parse `EXT_mesh_primitive_edge_visibility` `materials` and `lineStrings` during glTF load, extracting `baseColorFactor` as the default edge color and honoring per-lineString overrides.

- Expand the edge visibility pipeline to merge the bitfield and line-string edge data, generate optional edge-color attributes, and pass them through to the edge-rendering shader.

- Update the edge fragment stage to use the override colors from `material` section

### Material:
Defines an accessor to the glTF material that supplies the edge color. This field is optional, when it is omitted the edges fall back to the mesh’s per‑vertex colors.

### LineStrings
Defines accessors containing one or more line strings. Each entry may reference its own material; if not specified, the material defined in the section above is used, or otherwise the mesh’s vertex colors. Edges listed in this section must not be duplicated in the visibility bitfield (Handling during encoding: https://github.com/iTwin/imodel-native-internal/pull/888).

For more details please check:
EXT_mesh_primitive_edge_visibility README: [Edge Visibility](https://github.com/KhronosGroup/glTF/blob/96aa219b4099574a4f98dbfdbc52ba7c755e04ab/extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md)

## Issue number and link
Previous PR: https://github.com/CesiumGS/cesium/pull/12859
backlog issues: https://github.com/iTwin/itwin-graphics-backlog/issues/742
EXT_mesh_primitive_edge_visibility README: [Edge Visibility](https://github.com/KhronosGroup/glTF/blob/96aa219b4099574a4f98dbfdbc52ba7c755e04ab/extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md)
Khronos Group PR: [Edge Visibility](https://github.com/KhronosGroup/glTF/pull/2479)

## Example Test files:
Two triangles: both have red fill color — one with a green outline and the other with a blue outline.
Snipaste_2025-10-26_22-51-36

[example.zip](https://github.com/user-attachments/files/23130355/example.zip)

```
"materials": [
  {
    "pbrMetallicRoughness": {
      "metallicFactor": 0.0
    },
    "doubleSided": true
  },
  {
    "pbrMetallicRoughness": {
[...]

PR #12859, opened 2025-08-30
Add support for GLTF extension: EXT_mesh_primitive_edge_visibility

danielzhong:
# Description

This PR implements the proposed GLTF extension `EXT_mesh_primitive_edge_visibility`, enabling edge rendering for glTF mesh primitives. It adds multi-pass rendering to avoid z-fighting, and implements decoding, creation, and rendering of both hard and silhouette edges:

* Parsing and handling mesh primitive data of visibility (edge type: HIDDEN, SILHOUETTE, HARD, REPEATED) and silhouette normal.
* Hard edge & silhouette edge rendering.
* Multi-pass rendering path (CESIUM_3D_TILE_EDGES) to draw edges before surfaces (CESIUM_3D_TILE).
* Feature ID/Depth integration to ensure edges display correctly without Z fighting and overlapping.

More details: [README](https://github.com/KhronosGroup/glTF/blob/96aa219b4099574a4f98dbfdbc52ba7c755e04ab/extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md)

**Demo**
(green edge = hard edge; red edge = silhouette edge):
image

## Issue number and link
https://github.com/CesiumGS/cesium/issues/12765

## Testing plan
TODO
image

## 🔍Review Notes
This draft PR is open for feedback, especially regarding the overall architecture and certain design decisions.
* The glTF extension’s material support is not included in this PR. I'm planning to set all edges default to black.
* Update `getWebGLStub.js` to support MRT in coverage tests

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [x] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [ ] I have performed a self-review of my code

PR #13110, opened 2025-12-30
Support for BENTLEY_materials_line_style glTF Extension

danielzhong:
# Description
1. This PR implements support for the `BENTLEY_materials_line_style` gltfmaterial extension in CesiumJS. This extension enables CAD-style line visualization with variable width and dash patterns. The proposed specification can be found [here](https://github.com/CesiumGS/glTF/pull/89).
2. This PR fixed silhouette normal decode issues: https://github.com/iTwin/itwinjs-core/issues/8879
in this commit: https://github.com/CesiumGS/cesium/pull/13110/commits/fe13f48f7d0a4984758a018bbe6f10e6e5fee914
3. This PR updates the `EXT_mesh_primitive_edge_visibility` gltf extension by switching from `gl_lines` to tessellated quads rendered as `gl_triangles`, enabling line-width support.
4. This PR supersedes the previous PR ([#12859](https://github.com/CesiumGS/cesium/pull/13002)). All changes have been moved here and mainly focus on completing the material and line string support for EXT_mesh_primitive_edge_visibility
5. UPDATE (1/29): Issue: https://github.com/iTwin/itwinjs-backlog/issues/1812. Added cumulative-distance in line patterns gltf extension (no screen‑space calc), chose frustum‑based `pixelsPerWorld` over `czm_metersPerPixel` to match iTwin implementation.

# Overview (`BENTLEY_materials_line_style`)
`width: 5`, `pattern: 61680`
Snipaste_2025-12-30_11-30-46
Snipaste_2026-01-01_03-02-11

The BENTLEY_materials_line_style extension allows glTF materials to specify:

`width`: Line thickness in screen pixels
`pattern`: A 16-bit repeating on/off dash pattern (each bit = 1 screen pixel)
This PR allows CesiumJS to process and apply the above extension when loading glTF files. This means lines and edges will be able to have customizable width and pattern properties specified and respected in CesiumJS when loaded via glTF.

Implementation Details:
Because variable line width is required (and many graphics APIs including WebGL do not support `gl_line` primitives with width > 1), this PR refactors the `EXT_mesh_primitive_edge_visibility` implementation to use quad-based rendering instead of the previous gl_line approach. Each line segment is tessellated into a quad (two triangles) that is dynamically expanded perpendicular to the edge direction based on the material's `width` property.

The 16-bit dash pattern is applied in the fragment shader by testing individual bits against the screen-space position along the line, providing pixel-perfect pattern rendering that remains stable under camera movement.

Changes:
Added support for `BENTLEY_materials_line_style` in `GltfLoader.js` and `MaterialPipelineStage.js`
Refactored edge rendering from line primitives to quad-based geometry in `EdgeVisibilityPipelineStage.js`
Implemented vertex shader expansion logic for variable-width lines in `EdgeVisibilityStageVS.glsl`
Implemented fragment shader pattern matching using bit extraction in `EdgeVisibilityStageFS.glsl`
Added `u_lineWidth` and `u_linePattern` uniforms to the edge rendering pipeline

To test these changes locally:
- Clone this branch.
- Build CesiumJS and Sandcastle.
- Run the dev server locally.
- Host your glb/gltf test files using: `http-server ./ --cors=X-Correlation-Id`
- Navigate to Sandcastle and open the gltf:
[line_style.zip](https://github.com/user-attachments/files/24386140/line_style.zip)
[line&point_style.zip](https://github.com/user-attachments/files/24399748/line.point_style.zip)

```
  infoBox: false,
  selectionIndicator: false,
  shouldAnimate: true,
});

function createModel(url, height) {
  viewer.entities.removeAll();

  const position = Cesium.Cartesian3.fromDegrees(
    -123.0744619,
    44.0503706,
    height
  );

  const heading = Cesium.Math.toRadians(100.0);
  const pitch = 0.0;
  const roll = 0.0;
  const hpr = new Cesium.HeadingPitchRoll(heading, pitch, roll);
  const orientation = Cesium.Transforms.headingPitchRollQuaternion(position, hpr);

    name: url,
    position: position,
    orientation: orientation,
    model: {
      uri: url,
      minimumPixelSize: 128,
      maximumScale: 20000,
    },
  });

}

createModel("http://172.20.20.20:8282/3-1-0.glb", 100.0);
```

- Or `npm run build-sandcastle` and in sandcastle search for the `Styled glTF Lines` example.

## Note:
This PR may include changes from other currently open PRs, this PR was built on top of the following PRs:
 `EXT_mesh_primitive_edge_visibility's material and line string`: 
(This PR supersedes the previous PR ([#12859](https://github.com/CesiumGS/cesium/pull/13002)). All changes have been moved here and mainly focus on completing the material and line string support for EXT_mesh_primitive_edge_visibility. Please refer to the links for more details.)

For line patterns & cumulative distance:
If the demo does not show density changes while zooming, it’s usually because the camera frustum width isn’t changing (e.g., a trackedEntity or fixed frustum). Once frustum.width updates, the pattern density changes as expected.

For example:
```
// ------------------------------
// 1) Viewer
// ------------------------------
  infoBox: false,
  selectionIndicator: false,
  shadows: false,
  shouldAnimate: true,
});
 
const scene = viewer.scene;
const camera = scene.camera;
const canvas = scene.canvas;
 
// ------------------------------
// 2) Orthographic camera
// ------------------------------
const ortho = new Cesium.OrthographicFrustum();
ortho.near = 0.1;
ortho.far = 1.0e9;
ortho.width = 400.0;
ortho.aspectRatio = canvas.clientWidth / canvas.clientHeight;
 
//  frustum
camera.frustum = ortho;
 
function updateOrthoAspect() {
  if (camera.frustum instanceof Cesium.OrthographicFrustum) {
    camera.frustum.aspectRatio = canvas.clientWidth / canvas.clientHeight;
  }
}
window.addEventListener("resize", updateOrthoAspect);
 
// ------------------------------
// 3) Wheel zoom => change frustum.width only
// ------------------------------
scene.screenSpaceCameraController.enableZoom = false;
 
function clamp(v, min, max) {
  return Math.max(min, Math.min(max, v));
}
 
canvas.addEventListener(
  "wheel",
  (e) => {
    if (!(camera.frustum instanceof Cesium.OrthographicFrustum)) return;
    e.preventDefault();
 
    const frustum = camera.frustum;
 
    const zoomFactor = Math.pow(1.0015, e.deltaY);
    frustum.width *= zoomFactor;
    frustum.width = clamp(frustum.width, 0.01, 1.0e9);
 
    frustum.aspectRatio = canvas.clientWidth / canvas.clientHeight;

  },
  { passive: false }
);
 
// ------------------------------
// 4) Load model (no trackedEntity)
// ------------------------------
function createModel(url, height) {
  viewer.entities.removeAll();
 
  const position = Cesium.Cartesian3.fromDegrees(-123.0744619, 44.0503706, height);
 
  const heading = Cesium.Math.toRadians(135.0);
  const pitch = 0.0;
  const roll = 0.0;
 
  const hpr = new Cesium.HeadingPitchRoll(heading, pitch, roll);
  const orientation = Cesium.Transforms.headingPitchRollQuaternion(position, hpr);
 
    name: url,
    position,
    orientation,
    model: {
      uri: url,
      minimumPixelSize: 128,
      maximumScale: 20000,
    },
  });
 
  // viewer.trackedEntity = entity;
 
  viewer.zoomTo(entity).then(() => {
    camera.frustum = ortho;
    updateOrthoAspect();
  });
 
  scene.requestRender();
}
 
createModel("[http://172.20.20.20:8081/9-1-0.glb"](http://172.20.20.20:8081/9-1-0.glb%22), 100.0);
```

## Issue number and link
https://github.com/CesiumGS/cesium/issues/12889

## Testing plan

- [Live DEMO - orthographic polyline pattern](https://ci-builds.cesium.com/cesium/DanielZ/BENTLEY_materials_line_style/Apps/Sandcastle2/index.html#c=pVbbcuI4EP0VL09m12VMLlObay0BkjETyMAwyZCiakuRG1AwkiPLkDCVf9+WZAccUvOw8xCgW92nb6elNCFl2cIPBfcjmJAsVg1KIU2HYg7cOXPGFXjpzB6uKLthnfD7Oqz3WJiGfHBIm+GncJ78uG12jnw0eoqu5mgU7o32wlV3eMG6LHjutvqH18PBfLSm6qZ1P+ux+uNo73LRbdH93l1/NXpssOtmJ7lHsO46POg9jp61PPrRZzeP7f3esL/urvv1bguDfP/6JC+ip0+dZqf3pTNpi94wbKirQT0YHlwsHg7a9bv2LPz37xD648rJWI65/qOCp8pZMliBxII4rJymLfrW6NxxhRq5KbgijIMcV6rvvFMKHNDZovhGPNkcU7IASfDcHPhWLJ3zJUm3zrX4LoSQaibK+d1o1VSSZMbopcxSlS1cm5kx9jkQXVHg17eUE6Or+wEcbWlXLFIz1B8GgR9s6UmaAFUDopiObRPzacyAqzvjUisrPwObztQmdVOpP7HJIYJBfTueZJwiMneyJCIKTD0NE9GtOj+1heOwieO+g2HYD8IpiMkvOvEG4Dhl99+pyXFe9Zf5WDEeiZVPoqi9RKtrliocnqaLxKTWMK54u3VtiGNHnVIJwL8lhELTZKlJJkUcI4uAk4cY7oXQjZuQOIXdxtGYLBJ36TkLxvGDPL+VLUFlkjtdomY+6l1jYCXGXdR4zrJq03ndjMuUvVuSQRxXVjOAGOuysgtV5+x802Y9qT/+36yqebonBRb4iQSdQ8teOjmvzTTNOmwoVQ74zmyN3bskVAlNelN9IlYusj+oH3oYJYJYkdEGvOBIsQ+2v2Xtn1uoHm5XUPecOhztgvwe0fIu/3QSkqZsCceWA5aB1V0qSKGQaw0pMh61uWLqxQXz5TkRTJFm6Rs3bG8UW2zdWTQWdO7TTEr0GuLRScmWyCkotLaQ2MWU6ag+am9JnIGr0YoWGCbk48ZXA+/MyLUI5UlvR5CET3U6uVuTSIW/CN/3I2YpVFCriH3X9PK8qqVcZ0AixqcbahSKvwpwQwQlBqgmPHWL9pRQEqbobINhxO2krToWYt5QeXHe9uX82Qb9qv0GujY3T8Oz0J6t+P0KKvlSnhJeJiKTVLeGrAhTBT6+yIP8CLkvFo00BRVG7sFBUN/bD6o7DbaT2wzcyAzMurtvS1w09/iDQegwLdsrN0DmI+2DoOoVrguB23SMhM0kO97kXRD5dTslW0qeiV6nocjZWt1t8kfPh/PRm7Hl+ottOMqb84r4esYuSClkaTlEDDjaaX6ymVDFq5ym6iWGc235D1skmBHWG7u+X1OwSGIMmtYeMjoH5dNUc4qf1gqX04gtHRadffAfhb5p0hRPJlkcfzPPx/lpDe1LbrEwFLpZgozJizaZ1c+vrdL3/dMairteSoj4gcgtxP8A)
This demo primarily illustrates the issue described in https://github.com/iTwin/itwinjs-backlog/issues/1812. It demonstrates that the new cumulative distance has been successfully added in CesiumJS to address the problem. The demo uses an orthographic frustum, since users affected by this issue are mainly concerned with a top-down (2D) view.

- [Live DEMO- Edge Visibility with line pattern](https://ci-builds.cesium.com/cesium/DanielZ/BENTLEY_materials_line_style/Apps/Sandcastle2/index.html#c=bVJrb9owFP0rVj5RqXJIod3W0Wo0MOSsCYPSB1WkyTiXYnBssE0pTP3vcwhBnTYpie7jnHOPb+z7qKeptCgEw9d5dIcoY2AMsgpt1VojriSixoA1qU5licJESZzBlK6Fbe/hI7UAia5Q6sE2mk16jPd5RO53JEg4MUQOz1lILshi+fQQRl+wA62y3sKByNn4jGzi0Q2Pef0t7gzOb0fDxXjHbL/zPEt4MB+ffc/jDmskj4PNeN7mt2G0fHZi8Y40k/n4rcjHTwPen3cbyWiwi3eDIO64Ifc/V/omW11EYZT8iKZdlYxI2/aGQX3UvMknzW7w2J2RX58JDFLva3G64mVKGoteOWxAuwNJ2BxWgx/2tVrqsX0eKmkpl6BT7+TItnqLfhcBQqWQBuO2yMBJ0Q3l1Z6LDQ4PLTzVKm8XKyZZrdkImp8ajVKxEgFpud06idIW3uccDKZZVjuMQ2ipjKsqeVnNCKm2LqKysR/RgRcNYGr1U+SeoF4/Oa2oucpAXKKjFEJrzS+P5o+490P0Xvk7GLKasgVk3cpnaXiPeUeMWjZDNdBa6ZOP21ECsFAvh06JLj7eqdcydivgukB+4/lSaesciRrGvoV8Kag7lz9Zu5kWM2McV7b8itLK+Cvi2dV/fhRiwt1l15muhbjjO0i965bv8H/RhKIZly/9V9CCbgvILLi+LYsY45bv0n9ZVikxofqD4h8)
This demo simply shows that line patterns work correctly with edges using the edge visibility glTF extension.

- [Live DEMO 3 - Edge Visibility with line pattern2](https://ci-builds.cesium.com/cesium/DanielZ/BENTLEY_materials_line_style/Apps/Sandcastle2/index.html#c=bVJrb9owFP0rVj5RqXJIodXa0Wo0MOSsCYPSB1WkyTiXYnBssE0pTP3vcwhBnTYpie7jnHOPb+z7qKeptCgEw9d5dI8oY2AMsgpt1VojriSixoA1qU5licJESZzBlK6Fbe/hI7UAia5R6sE2mk16jPd5RB52JEg4MUQOz1lILshi+fwYRpfYgVZZb+FA5Gx8Rjbx6JbHvP4edwbnd6PhYrxjtt95mSU8mI/PvudxhzWSp8FmPG/zuzBavjixeEeayXz8XuTj5wHvz7uNZDTYxbtBEHfckIefK32brS6iMEp+RNOuSkakbXvDoD5q3uaTZjd46s7Iry8EBqn3tThd8TIljUVvHDag3YEkbA6rwY/7Wi312D4PlbSUS9Cpd3JkW71Fv4sAoVJIg3FbZOCk6Ibyas/FBoeHFp5qlbeLFZOs1mzUL5uNZqlYiYC03G6dRGkL73MOBtMsqx3GIbRUxlWVvKpmhFRbF1HZ2I/owKsGMLX6KXJPUK+fnFbUXGUgrtBRCqG15ldH80fcxyH6qPwdDFlN2QKybuWzNLzHfCBGLZuhGmit9Mnn7SgBWKjXQ6dEFx/v1GsZuxVwUyC/8XyptHWORA1j30K+FNSdy5+s3UyLmTGOK1t+RWll/A3x7Po/Pwox4e6y60zXQtzzHaTeTct3+L9oQtGMy9f+G2hBtwVkFtzclUWMcct36b8sq5SYUP1J8Q8)
- [Live DEMO 4 - Edge Visibility only](https://ci-builds.cesium.com/cesium/DanielZ/BENTLEY_materials_line_style/Apps/Sandcastle2/index.html#c=bVJrb9owFP0rVj5RqXIItFXb0Wo0MOS0CYPSB1WkyTiXYnBsaptSmPrf5xCCOm1SEt3HOece39j3UU9TaVEIhq/y6B5RxsAYZBXaqJVGXElEjQFrUp3KEoWJkjiDKV0J297BR2oBEl2h1INNNJv0GO/ziDxsSZBwYogcnrKQnJHF8vkxjC6wA71lvYUDkca4Qdbx6IbHvP4Rdwand6PhYrxltt95mSU8mI8bP/K4w5rJ02A9nrf5XRgtX5xYvCUnyXz8UeTj5wHvz7vNZDTYxttBEHfckIefb/omezuLwii5jaZdlYxI2/aGQX10cpNPTrrBU3dGfp0TGKTet+J0xcuUNBa9c1iDdgeSsN6vBj/uarXUY7s8VNJSLkGn3tGBbfUG/S4ChEohDcZtkYGTomvKqz0XGxzuW3iqVd4uVkyyWvPiPGicBqViJQLScrtxEqUtvMs5GEyzrLYfh9BSGVdV8rKaEVJtXURlczeiA68awNTqx8g9Qb1+dFxRc5WBuEQHKYRWml8ezB9wn/vos/K3N2Q1ZQvIupXP0vAO84kYtWyGaqC10kdft6MEYKFe950SXXy8Y69l7EbAdYH8zvOl0tY5EjWMfQv5UlB3Ln+ycjMtZsY4rmz5FaWV8XfEs6v//CjEhLvLrjNdCXHPt5B61y3f4f+iCUUzLl/776AF3RSQWXB9VxYxxi3fpf+yrFJiQvUXxT8)
- Added unit tests
- Added a Cesium Sandcastle sample.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/ModelComponents.js
+++ b/packages/engine/Source/Scene/ModelComponents.js
@@ -1556,7 +1556,6 @@ function LineStyle() {
    *
    * @type {number|undefined}
    * @default undefined
-   * @private
    */
   this.width = undefined;
 
@@ -1566,7 +1565,6 @@ function LineStyle() {
    *
    * @type {number|undefined}
    * @default undefined
-   * @private
    */
   this.pattern = undefined;
 }


pmconne:
Your branch appears to contain all of the changes from #13093 too. That's unintentional, right?

Apps/SampleData/models/StyledLines/BENTLEY_materials_line_style.gltf
javagl, line 1: Validation error: "VALUE_NOT_IN_RANGE | Value 1 is out of range. | /bufferViews/5/byteStride"

danielzhong:
https://github.com/CesiumGS/cesium/pull/13110/commits/fe13f48f7d0a4984758a018bbe6f10e6e5fee914

Fixed silhouette normal decode issues: https://github.com/iTwin/itwinjs-core/issues/8879

Apps/SampleData/models/StyledLines/BENTLEY_materials_line_style.gltf
danielzhong: Fixed. Thanks

markschlosseratbentley:
Waiting for https://github.com/CesiumGS/cesium/pull/13093 to merge first.

markschlosseratbentley:
@lilleyse This has made a bunch of progress, so would you mind taking another pass on it?

cc @danielzhong

Specs/Data/Models/glTF-2.0/StyledLines/BENTLEY_materials_line_style.gltf
```diff
@@ -0,0 +1,339 @@
+{
```
markschlosseratbentley, line 1: Is this .gltf file a duplicate of `Apps/SampleData/models/StyledLines/BENTLEY_materials_line_style.gltf`? Could we only commit one copy of this file and reference it from a single location? This PR shows an example of doing that: https://github.com/CesiumGS/cesium/pull/13093/

packages/engine/Specs/Scene/Model/MaterialPipelineStageSpec.js
```diff
@@ -906,6 +906,109 @@ describe(
         u_testTexture: mockTexture,
       });
     });
+
```
markschlosseratbentley, line 912: Is there something wrong with `HAS_LINE_STYLE` and `HAS_LINE_WIDTH`?  I did a global search and see them only referenced in the test (spec) file.

I am getting the following test errors locally (related?):

```
    ✗ validates u_isEdgePass uniform and framebuffer attachments
        RuntimeError: Vertex shader failed to compile.  Compile log: ERROR: 0:220: 'u_lineWidth' : undeclared identifier
    ✗ validates EdgeVisibility shader code and uniforms
        RuntimeError: Vertex shader failed to compile.  Compile log: ERROR: 0:220: 'u_lineWidth' : undeclared identifier
    ✗ validates EdgeDetection shader code and texture sampling
        RuntimeError: Vertex shader failed to compile.  Compile log: ERROR: 0:220: 'u_lineWidth' : undeclared identifier
      ✗ renders point cloud with draco encoded positions, normals, colors, and batch table properties
      ✗ renders point cloud with batch table
      ✗ renders point cloud with batch table that contains only JSON data
      ✗ renders pnts with point size style
      ✗ picks based on batchId
      ✗ Supports back face culling when there are per-point normals
Chrome 144.0.0.0 (Mac OS 10.15.7): Executed 15111 of 15178 (15 FAILED) (skipped 67) (49.561 secs / 46.185 secs)
TOTAL: 15 FAILED, 15096 SUCCESS
     RuntimeError: Vertex shader failed to compile.  Compile log: ERROR: 0:220: 'u_lineWidth' : undeclared identifier
     RuntimeError: Vertex shader failed to compile.  Compile log: ERROR: 0:220: 'u_lineWidth' : undeclared identifier
     RuntimeError: Vertex shader failed to compile.  Compile log: ERROR: 0:220: 'u_lineWidth' : undeclared identifier
```

Specs/Data/Models/glTF-2.0/StyledLines/BENTLEY_materials_line_style.gltf
danielzhong: Fixed. Thanks!

packages/engine/Specs/Scene/Model/MaterialPipelineStageSpec.js
danielzhong: Fixed, Thanks!

Specs/Data/Models/glTF-2.0/StyledLines/BENTLEY_materials_line_style.gltf
markschlosseratbentley: Looks good, thanks!

markschlosseratbentley:
@jjhembd fyi

jjhembd:
@danielzhong thanks for this PR! It is a fairly big change, so it will take me some time to finish the review. I should have some feedback early next week. Let me know if you need anything sooner.

javagl:
Admittedly, I'm a bit surprised that `debugWireframe` still seems to work...
```javascript
  infoBox: false,
  selectionIndicator: false,
  shouldAnimate: true,
});

async function createModel(url, height) {
  const transform = Cesium.Transforms.eastNorthUpToFixedFrame(
    Cesium.Cartesian3.fromDegrees(-75.152408, 39.946975, height)
  );
    url: url,
    modelMatrix: transform,

    // TEST debugWireframe
    enableDebugWireframe: true,
    debugWireframe: true
  });
    viewer.scene.camera.flyToBoundingSphere(
      model.boundingSphere, {
      duration: 0
    });
  });
}

createModel("http://localhost:8003/line.point_style/7-1-0.glb", 20.0);
```
I haven't looked at the implementation here, but know that `debugWireframe` does some index juggling, and that could easily have affected the implementation of anything that is related to edge rendering (or relies on indices in any way).

A detail:
The test glTFs both have some validation errors of the form
`VALUE_NOT_IN_RANGE | Value 1 is out of range. | /bufferViews/5/byteStride`

danielzhong:
> Admittedly, I'm a bit surprised that `debugWireframe` still seems to work...
> 
> ```js
> const viewer = new Cesium.Viewer("cesiumContainer", {
>   infoBox: false,
>   selectionIndicator: false,
>   shouldAnimate: true,
> });
> 
> async function createModel(url, height) {
>   const transform = Cesium.Transforms.eastNorthUpToFixedFrame(
>     Cesium.Cartesian3.fromDegrees(-75.152408, 39.946975, height)
>   );
>   const model = await Cesium.Model.fromGltfAsync({
>     url: url,
>     modelMatrix: transform,
> 
>     // TEST debugWireframe
>     enableDebugWireframe: true,
>     debugWireframe: true
>   });
>   viewer.scene.primitives.add(model);
>   model.readyEvent.addEventListener(() => {
>     viewer.scene.camera.flyToBoundingSphere(
>       model.boundingSphere, {
>       duration: 0
>     });
>   });
> }
> 
> createModel("http://localhost:8003/line.point_style/7-1-0.glb", 20.0);
> ```
> 
> I haven't looked at the implementation here, but know that `debugWireframe` does some index juggling, and that could easily have affected the implementation of anything that is related to edge rendering (or relies on indices in any way).
> 
> A detail: The test glTFs both have some validation errors of the form `VALUE_NOT_IN_RANGE | Value 1 is out of range. | /bufferViews/5/byteStride`

@javagl Could you let me know which tool you used for validation and which files show this error? I ran them through `https://github.khronos.org/glTF-Validator/`, and they are reported as valid, I don’t see any errors at the moment.

javagl:
> Could you let me know which tool you used for validation and which files show this error?

Apologies for the false alert here: I ran the test with the files that had been attached in the first comment ([line_style.zip](https://github.com/user-attachments/files/24386140/line_style.zip) and [line&point_style.zip](https://github.com/user-attachments/files/24399748/line.point_style.zip)), and they still contained the errors. 

But I now checked the ones that are part of this PR, and they do not cause any errors 👍

packages/sandcastle/gallery/styled-gltf-lines-dev/index.html
```diff
@@ -0,0 +1,36 @@
+<!doctype html>
```
jjhembd: Hi @danielzhong, I am looking at this now and plan to give more feedback tomorrow. 

One initial comment: This Sandcastle HTML doesn't follow the pattern of the new Sandcastle files. Can we replace this with a more minimal file like the others? Most of this boilerplate should be taken care of by the sandcastle build step.

danielzhong: Fixed, thanks!

jjhembd:
Hi @danielzhong, I am looking through this in a little more depth. Can you help me with a couple big-picture questions here?

### Change to existing edge rendering examples
The PR description mentions this:

> This PR updates the EXT_mesh_primitive_edge_visibility gltf extension by switching from gl_lines to tessellated quads rendered as gl_triangles, enabling line-width support.

The testing plan doesn't mention how to verify that old examples using `EXT_mesh_primitive_edge_visibility` are still working. What examples can I look at, to confirm we didn't break anything for users of the old extension?

### Interaction between defines
Intuitively, I would expect a hierarchy something like this:
```glsl
#if (defined(EDGES_VISIBLE))
    // Process attributes & uniforms related to edge rendering
    // ...
    #if (defined(VARIABLE_EDGE_WIDTH))
        // Process attributes & uniforms related to variable edge width
    #endif
    #if (defined(EDGE_LINE_PATTERN))
        // Process attributes & uniforms related to line patterns like dashes etc
    #endif
#endif
```

But inside `ModelFS` I see this:
```glsl
 #if defined(HAS_LINE_PATTERN) && !defined(HAS_EDGE_VISIBILITY)
    const float maskLength = 16.0;
    // More line pattern processing
```
Why would we bother with a line pattern if the edges are not visible?

I assume `HAS_EDGE_VISIBILITY` is used for a completely separate edge rendering path. If that is true, perhaps the define should be renamed to something like `USE_EDGE_VISIBILITY_EXTENSION`, to make these `ifdef`s more readable. Even then, are we assuming the `EDGE_VISIBILITY` extension would override a line pattern?

danielzhong:
> Hi @danielzhong, I am looking through this in a little more depth. Can you help me with a couple big-picture questions here?
> 
> ### Change to existing edge rendering examples
> The PR description mentions this:
> 
> > This PR updates the EXT_mesh_primitive_edge_visibility gltf extension by switching from gl_lines to tessellated quads rendered as gl_triangles, enabling line-width support.
> 
> The testing plan doesn't mention how to verify that old examples using `EXT_mesh_primitive_edge_visibility` are still working. What examples can I look at, to confirm we didn't break anything for users of the old extension?
> 
> ### Interaction between defines
> Intuitively, I would expect a hierarchy something like this:
> 
> ```glsl
> #if (defined(EDGES_VISIBLE))
>     // Process attributes & uniforms related to edge rendering
>     // ...
>     #if (defined(VARIABLE_EDGE_WIDTH))
>         // Process attributes & uniforms related to variable edge width
>     #endif
>     #if (defined(EDGE_LINE_PATTERN))
>         // Process attributes & uniforms related to line patterns like dashes etc
>     #endif
> #endif
> ```
> 
> But inside `ModelFS` I see this:
> 
> ```glsl
>  #if defined(HAS_LINE_PATTERN) && !defined(HAS_EDGE_VISIBILITY)
>     const float maskLength = 16.0;
>     // More line pattern processing
> ```
> 
> Why would we bother with a line pattern if the edges are not visible?
> 
> I assume `HAS_EDGE_VISIBILITY` is used for a completely separate edge rendering path. If that is true, perhaps the define should be renamed to something like `USE_EDGE_VISIBILITY_EXTENSION`, to make these `ifdef`s more readable. Even then, are we assuming the `EDGE_VISIBILITY` extension would override a line pattern?

Hi, thanks for taking the time to review this.

1. There is an existing EdgeVisibility test file under:
`\Specs\Data\Models\glTF-2.0\EdgeVisibility\glTF-Binary`

This test file has not been changed, and the previous unit tests were not modified either. I’ve verified that it still renders correctly. In addition, I’ve included another [demo](https://ci-builds.cesium.com/cesium/DanielZ/BENTLEY_materials_line_style/Apps/Sandcastle2/index.html#c=bVJrb9owFP0rVj5RqXIItFXb0Wo0MOS0CYPSB1WkyTiXYnBsaptSmPrf5xCCOm1SEt3HOece39j3UU9TaVEIhq/y6B5RxsAYZBXaqJVGXElEjQFrUp3KEoWJkjiDKV0J297BR2oBEl2h1INNNJv0GO/ziDxsSZBwYogcnrKQnJHF8vkxjC6wA71lvYUDkca4Qdbx6IbHvP4Rdwand6PhYrxltt95mSU8mI8bP/K4w5rJ02A9nrf5XRgtX5xYvCUnyXz8UeTj5wHvz7vNZDTYxttBEHfckIefb/omezuLwii5jaZdlYxI2/aGQX10cpNPTrrBU3dGfp0TGKTet+J0xcuUNBa9c1iDdgeSsN6vBj/uarXUY7s8VNJSLkGn3tGBbfUG/S4ChEohDcZtkYGTomvKqz0XGxzuW3iqVd4uVkyyWvPiPGicBqViJQLScrtxEqUtvMs5GEyzrLYfh9BSGVdV8rKaEVJtXURlczeiA68awNTqx8g9Qb1+dFxRc5WBuEQHKYRWml8ezB9wn/vos/K3N2Q1ZQvIupXP0vAO84kYtWyGaqC10kdft6MEYKFe950SXXy8Y69l7EbAdYH8zvOl0tY5EjWMfQv5UlB3Ln+ycjMtZsY4rmz5FaWV8XfEs6v//CjEhLvLrjNdCXHPt5B61y3f4f+iCUUzLl/776AF3RSQWXB9VxYxxi3fpf+yrFJiQvUXxT8) for edge visibility in the PR description.

There is no distinction between "old" and "new" `EXT_mesh_primitive_edge_visibility glTF` files, the file format is unchanged. The same `.glb` works with both the old GL_LINES path and the new quad-based path (Required to support the line style extension, including line width and line pattern features.) The existing `EdgeVisibility.glb` test file and its specs in `GltfLoaderSpec.js` already serve as the backward compatibility test. 

2. `BENTLEY_materials_line_style` is a material extension that applies independently to any line primitive. A plain `GL_LINES` mesh with no `EXT_mesh_primitive_edge_visibility` can still have a dash pattern ([Example](https://ci-builds.cesium.com/cesium/DanielZ/BENTLEY_materials_line_style/Apps/Sandcastle2/index.html#c=pVbbcuI4EP0VL09m12VMLlObay0BkjETyMAwyZCiakuRG1AwkiPLkDCVf9+WZAccUvOw8xCgW92nb6elNCFl2cIPBfcjmJAsVg1KIU2HYg7cOXPGFXjpzB6uKLthnfD7Oqz3WJiGfHBIm+GncJ78uG12jnw0eoqu5mgU7o32wlV3eMG6LHjutvqH18PBfLSm6qZ1P+ux+uNo73LRbdH93l1/NXpssOtmJ7lHsO46POg9jp61PPrRZzeP7f3esL/urvv1bguDfP/6JC+ip0+dZqf3pTNpi94wbKirQT0YHlwsHg7a9bv2LPz37xD648rJWI65/qOCp8pZMliBxII4rJymLfrW6NxxhRq5KbgijIMcV6rvvFMKHNDZovhGPNkcU7IASfDcHPhWLJ3zJUm3zrX4LoSQaibK+d1o1VSSZMbopcxSlS1cm5kx9jkQXVHg17eUE6Or+wEcbWlXLFIz1B8GgR9s6UmaAFUDopiObRPzacyAqzvjUisrPwObztQmdVOpP7HJIYJBfTueZJwiMneyJCIKTD0NE9GtOj+1heOwieO+g2HYD8IpiMkvOvEG4Dhl99+pyXFe9Zf5WDEeiZVPoqi9RKtrliocnqaLxKTWMK54u3VtiGNHnVIJwL8lhELTZKlJJkUcI4uAk4cY7oXQjZuQOIXdxtGYLBJ36TkLxvGDPL+VLUFlkjtdomY+6l1jYCXGXdR4zrJq03ndjMuUvVuSQRxXVjOAGOuysgtV5+x802Y9qT/+36yqebonBRb4iQSdQ8teOjmvzTTNOmwoVQ74zmyN3bskVAlNelN9IlYusj+oH3oYJYJYkdEGvOBIsQ+2v2Xtn1uoHm5XUPecOhztgvwe0fIu/3QSkqZsCceWA5aB1V0qSKGQaw0pMh61uWLqxQXz5TkRTJFm6Rs3bG8UW2zdWTQWdO7TTEr0GuLRScmWyCkotLaQ2MWU6ag+am9JnIGr0YoWGCbk48ZXA+/MyLUI5UlvR5CET3U6uVuTSIW/CN/3I2YpVFCriH3X9PK8qqVcZ0AixqcbahSKvwpwQwQlBqgmPHWL9pRQEqbobINhxO2krToWYt5QeXHe9uX82Qb9qv0GujY3T8Oz0J6t+P0KKvlSnhJeJiKTVLeGrAhTBT6+yIP8CLkvFo00BRVG7sFBUN/bD6o7DbaT2wzcyAzMurtvS1w09/iDQegwLdsrN0DmI+2DoOoVrguB23SMhM0kO97kXRD5dTslW0qeiV6nocjZWt1t8kfPh/PRm7Hl+ottOMqb84r4esYuSClkaTlEDDjaaX6ymVDFq5ym6iWGc235D1skmBHWG7u+X1OwSGIMmtYeMjoH5dNUc4qf1gqX04gtHRadffAfhb5p0hRPJlkcfzPPx/lpDe1LbrEwFLpZgozJizaZ1c+vrdL3/dMairteSoj4gcgtxP8A)) - that case is handled in `ModelFS.glsl`. 

* No Edge Visibility -> pattern run in ModelFS.glsl
* Has Edge Visibility -> Run in EdgeVisibilityStageFS.glsl

The `!defined(HAS_EDGE_VISIBILITY)` guard simply prevents double-processing when both extensions are present, since edge visibility's own fragment stage handles the pattern internally.

lilleyse:
I haven't reviewed the code thoroughly, but I have been using this branch for my snapping experiments and functionality-wise it's working nicely.

packages/engine/Source/Shaders/Model/EdgeVisibilityStageVS.glsl
```diff
@@ -0,0 +1,96 @@
+#ifdef HAS_EDGE_VISIBILITY
+void edgeVisibilityStageVS() {
+    if (u_isEdgePass) {
```
jjhembd: The whole body of the function is inside an `if` block. An early return (similar to `EdgeVisibilityStageFS`) is easier to read, especially when the rest of the code is broken up with `#ifdef`s:

```glsl
if (!u_isEdgePass) {
    return;
}
// continue with rest of function
```

A bigger question is: why are we branching like this at all? (see our [Coding Guide on branching in shaders](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#branching-in-shaders)). If we are not in an edge pass, this whole shader shouldn't even be part of the draw call.  I realize this was pre-existing behavior in the old fragment shader, so refactoring the construction of draw commands may be out-of-scope for the PR.

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -294,53 +377,164 @@ function buildTriangleAdjacency(primitive) {
...
+  const hasGLBSilhouetteNormals =
+    defined(edgeVisibility) && defined(edgeVisibility.silhouetteNormals);
+  let silhouetteNormalsUint32 = null;
 
-  // Each edge needs 2 face normals (left and right side)
-  const edgeFaceNormals = new Float32Array(numEdges * 6); // 2 normals * 3 components each
+  if (hasGLBSilhouetteNormals) {
+    // GLB stores VEC3 BYTE as normalized normal vectors (signed bytes).
+    // Decode from signed bytes to normalized vectors, then re-encode to 16-bit octahedral format.
+    const normalize = (val) => 2 * ((val + 128) / 255) - 1;
```
jjhembd:
```suggestion
    const decodeSignedByte = (val) => 2 * ((val + 128) / 255) - 1;
```

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -294,53 +377,164 @@ function buildTriangleAdjacency(primitive) {
...
+      edgeVisibility.silhouetteNormals.length,
+    );
+    const scratchNormal = new Cartesian3();
+    const scratchEncoded = new Cartesian2();
+
+    for (let i = 0; i < edgeVisibility.silhouetteNormals.length; i++) {
+      const vec3 = edgeVisibility.silhouetteNormals[i];
+
+      // Denormalize from signed byte to normal vector
+      scratchNormal.x = normalize(vec3.x);
```
jjhembd:
```suggestion
      // Decode from signed byte to normal vector
      scratchNormal.x = decodeSignedByte(vec3.x);
```
I think `decodeSignedByte` would better explain what is happening here? The output is not normalized, because the next step is to normalize it.

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -294,53 +377,164 @@ function buildTriangleAdjacency(primitive) {
...
     const b = edgeIndices[i * 2 + 1];
     const edgeKey = `${a < b ? a : b},${a < b ? b : a}`;
     const triangleList = edgeMap.get(edgeKey);
 
-    // Expect at least one triangle; silently skip if not found (defensive)
-    if (!defined(triangleList) || triangleList.length === 0) {
-      continue;
+    const currentEdgeData = edgeData[i];
+    const edgeType = currentEdgeData.edgeType;
+    const mateVertexIndex = currentEdgeData.mateVertexIndex;
```
jjhembd:
```suggestion
    const { edgeType, mateVertexIndex } = edgeData[i];
```
Just FYI: we are using destructuring where appropriate. See https://github.com/CesiumGS/cesium/issues/12196#issuecomment-2356379404.

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -294,53 +377,164 @@ function buildTriangleAdjacency(primitive) {
...
+      // Each OctEncodedNormalPair is stored as one Uint32 value
+      // Uint32 contains 4 bytes: [byte0, byte1, byte2, byte3]
+      // normalA = byte0 | (byte1 << 8)  - little endian
+      // normalB = byte2 | (byte3 << 8)  - little endian
+
+      if (mateVertexIndex < silhouetteNormalsUint32.length) {
+        const uint32Value = silhouetteNormalsUint32[mateVertexIndex];
+
+        const decodedA = new Cartesian3();
+        const decodedB = new Cartesian3();
```
jjhembd: These `Cartesian3`s could be declared as scratch variables outside the `for` loop, to avoid re-allocation and the associated performance hit. See https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#result-parameters-and-scratch-variables

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -478,12 +829,18 @@ function extractVisibleEdges(primitive) {
...
  * @param {number} edgeFeatureIdLocation Shader attribute location for optional edge feature ID
+ * @param {number} edgeColorLocation Shader attribute location for optional edge color data
+ * @param {number} edgeOtherPosLocation Shader attribute location for the other endpoint position
+ * @param {number} edgeOffsetLocation Shader attribute location for edge offset (-1 or +1)
+ * @param {VertexColorInfo} [vertexColorInfo] Packed per-vertex colors (optional)
  * @param {Object} edgeVisibility Edge visibility extension object (may contain silhouetteNormals[])
  * @param {Float32Array} edgeFaceNormals Packed face normals (6 floats per edge)
+ * @param {Object} [cumDistAttribute] Cumulative distance attribute
+ * @param {number} [edgeCumDistLocation] Cumulative distance location
  * @returns {Object|undefined} Object with {vertexArray, indexBuffer, indexCount} or undefined on failure
```
jjhembd: I think the return type has changed: no `indexBuffer`,  new boolean `hasEdgeFeatureIds`

packages/engine/Source/Scene/ModelComponents.js
```diff
@@ -1633,6 +1641,27 @@ function Material() {
...
+  /**
+   * The line dash pattern for LINES primitives. This is set by the
+   * BENTLEY_materials_line_style extension. Encoded as a 16-bit unsigned integer
+   * where each bit represents a pixel (1=on, 0=off).
+   *
+   * @type {number}
+   * @default undefined
+   * @private
+   */
+  this.linePattern = undefined;
```
jjhembd: For other extensions (e.g., `KHR_materials_clearcoat`, `KHR_materials_anisotropy`) we added a single property to `Material`, which is an object pointing to all the properties needed for that extension. I think it would be better to follow that pattern here, both to keep `Material` clean and consistent, and also to avoid collisions with possible future extensions.

One option: `this.lineStyle` ? Then define a separate type `ModelComponents.lineStyle` with the `lineWidth` and `linePattern` properties.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -184,6 +185,124 @@ MaterialPipelineStage.process = function (
     };
   }
 
+  // Configure and handle line style for LINES primitives (BENTLEY_materials_line_style extension).
```
jjhembd: Other extensions like `KHR_materials_clearcoat` and `KHR_materials_anisotropy` are handled in a separate helper function. Following the same pattern here would give better consistency and readability.

See the comment in `ModelComponents.Material` about the material property naming. This whole section could read as follows:

```javascript
if (defined(material.lineStyle)) {
  processLineStyleUniforms(
    material.lineStyle,
    uniformMap,
    shaderBuilder,
    defaultTexture,
    disableTextures
  );
}
```

packages/engine/Source/Scene/Model/ModelDrawCommand.js
```diff
@@ -815,6 +820,15 @@ function deriveEdgeCommand(command, renderResources) {
   };
   edgeCommand.uniformMap = uniformMap;
 
+  // Set line width uniform for quad-based edge rendering
+  const lineWidth = defined(edgeGeometry.lineWidth)
+    ? edgeGeometry.lineWidth
+    : 1.0;
+  uniformMap.u_lineWidth = function () {
+    return lineWidth;
+  };
```
jjhembd: Sorry if this was obvious, but why do we need to redefine `u_lineWidth` here? We already have a `u_lineWidth` getter defined in `MaterialPipelineStage` (line 191) which looks better--it multiplies by the `frameState.pixelRatio`, which could potentially change between frames.

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -1840,6 +1840,25 @@ function loadMaterial(loader, gltfMaterial, frameState) {
     }
   }
 
+  // BENTLEY_materials_line_style extension
+  const lineStyleExtension = extensions.BENTLEY_materials_line_style;
+  if (defined(lineStyleExtension)) {
```
jjhembd: Extension-specific code should be moved to a helper function, following the same pattern used above for `specularGlossiness`, `metallicRoughness`, `specular`, `anisotropy`, etc.

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -2070,37 +2172,44 @@ function loadPrimitive(loader, gltfPrimitive, hasInstances, frameState) {
 
   // Edge Visibility
```
jjhembd: This section should be moved to a helper function, similar to the pattern of the rest of `loadPrimitive`.

See the [Coding Guide section on functions](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#functions):

> Statements in a function should be at a similar level of abstraction. If a code block is much lower level than the rest of the statements, it is a good candidate to move to a helper function

packages/engine/Source/Shaders/Model/ModelFS.glsl
```diff
@@ -107,6 +107,16 @@ void main()
     // When picking metadata END
     //========================================================================
 
+    #if defined(HAS_LINE_PATTERN) && !defined(HAS_EDGE_VISIBILITY)
```
jjhembd: This section is a much lower level of abstraction than the rest of the function. It would read better if moved to a `lineStyleStage`

packages/engine/Source/Shaders/Model/ModelFS.glsl
```diff
@@ -107,6 +107,16 @@ void main()
...
+    const float maskLength = 16.0;
+    float dashPosition = fract(v_lineCoord / maskLength);
+    float maskIndex = floor(dashPosition * maskLength);
+    float maskTest = floor(u_linePattern / pow(2.0, maskIndex));
+    if (mod(maskTest, 2.0) < 1.0) {
+        discard;
```
jjhembd: Using `discard` to implement a dash will create a hard edge, which is prone to aliasing. Have we considered an alpha taper instead?
A taper could give nicer results with zero performance penalty. A taper is a single (fast) floating-point operation, whereas a `discard` breaks parallelism.

packages/engine/Source/Shaders/Model/EdgeVisibilityStageVS.glsl
danielzhong: Agree! Thanks!
https://github.com/CesiumGS/cesium/pull/13110/commits/7bf57ccf6ffe9c4d80c1c59419902c97fce70df9

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
danielzhong: Fixed. Thanks!
https://github.com/CesiumGS/cesium/pull/13110/changes/e444a629bdca3cf4c3305db20b8e33d98265e0a7

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
danielzhong: Good catch! Thanks!
https://github.com/CesiumGS/cesium/pull/13110/commits/34c203a29f7d51ab0d53d27225e008ce928fd423

packages/engine/Source/Scene/ModelComponents.js
danielzhong: Agree!

https://github.com/CesiumGS/cesium/pull/13110/commits/ceaa209627b88d2722c7f7177f99824309181a18

packages/engine/Source/Scene/Model/ModelDrawCommand.js
danielzhong: I think the redefinition in `ModelDrawCommand` was masking a separate issue. `MaterialPipelineStage` already defines `u_lineWidth` correctly, but `EdgeVisibilityPipelineStage` was later overwriting it with a fallback value. That meant removing the override here changed behavior because the edge pass started picking up the fallback instead of the material driven value.

I updated `EdgeVisibilityPipelineStage` so it only provides fallback uniforms when they are not already defined. With that in place, the edge command can safely reuse the cloned `u_lineWidth` getter from the original uniform map.

https://github.com/CesiumGS/cesium/pull/13110/commits/2aa827a739374d94a65cc28aca4f73b34f1b8e3c

packages/engine/Source/Scene/GltfLoader.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13110/commits/67786aedbbf4d35caa7c8cdaa95ad9b1d485cfe4

packages/engine/Source/Scene/GltfLoader.js
danielzhong: Fixed, thanks!

https://github.com/CesiumGS/cesium/pull/13110/commits/3e6d500b8db7f138a37a57a83b46cd5603c54957

packages/engine/Source/Shaders/Model/ModelFS.glsl
danielzhong: https://github.com/CesiumGS/cesium/pull/13110/commits/6e133789d0f194125347ec8272c787d05b20a7a6

packages/engine/Source/Shaders/Model/ModelFS.glsl
danielzhong: Good idea. I used the current `discard` based implementation in this PR to stay aligned with the existing iTwin behavior and with the extension’s current binary on/off pattern semantics.

An alpha taper could be a useful follow up experiment, but it would change the visual behavior and may also require a broader look at blending/depth interactions. To keep the scope of this PR focused, I’d prefer to leave that exploration for a future PR.

PR #10644, opened 2022-08-05
Replace glTF model implementation

ptrgags:
Fixes #10346
Fixes #10416
Fixes #10379
Fixes #9303
Fixes #7062
Fixes #6299
Fixes #2387

This PR replaces CesiumJS's glTF model implementation. the old `Model` was removed, and `ModelExperimental` was renamed to `Model` as a drop-in replacement. The new architecture is a lot more decoupled, and it made it easier to implement features such as 3D Tiles Next metadata and `CustomShader`

This PR merges the staging branch `replace-model` into `main`. It includes changes from the following PRs:

* https://github.com/CesiumGS/cesium/pull/10595
* https://github.com/CesiumGS/cesium/pull/10600
* https://github.com/CesiumGS/cesium/pull/10623
* https://github.com/CesiumGS/cesium/pull/10639
* https://github.com/CesiumGS/cesium/pull/10643
* https://github.com/CesiumGS/cesium/pull/10633
* https://github.com/CesiumGS/cesium/pull/10635
* https://github.com/CesiumGS/cesium/pull/10637
* https://github.com/CesiumGS/cesium/pull/10642
* https://github.com/CesiumGS/cesium/pull/10646
* https://github.com/CesiumGS/cesium/pull/10647

To Do:

- [x] Update CHANGES.md with an announcement
- [x] Test Sandcastles again
- [x] Sweep through and see if we need to remove `@experimental` anywhere
- [x] ~Add changes from https://github.com/CesiumGS/gltf-pipeline/pull/619 so we can have (limited) support for `glTF 1.0` and `KHR_techniques_webgl`~ Will be done separately

PR #13111, opened 2026-01-02
Remove unused test utils, GltfBuilder.js

donmccurdy:
# Description

While reviewing uses of glTF Pipeline in the CesiumJS codebase, I ran across GltfBuilder.js (a utility in the `packages/engine/Specs/` directory). This file is currently unused, with its original use case (ModelOutlineLoader) removed by #10644 in 2022. My feeling is that it's easy enough to restore the file from version history if needed, and that it can't easily be tested/maintained in the meantime, and so could be removed for now. 

If the original use case, from #8776, were coming up again tomorrow - another option would be to generate these test files with https://gltf-transform.dev (disclaimer: I'm the author of glTF Transform). 

## Issue number and link

- Followup to #10644

## Testing plan

n/a, removing unused code.

# Author checklist

#### PR Dependency Tree

* **PR #13111** 👈

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,12 +1,26 @@
 # Change Log
 
-## 1.137 - 2026-01-01
+## 1.138 - 2026-02-02
+
+### @cesium/engine
+
+#### Fixes :wrench:
+
+- Fixes jitter artifacts on Intel Arc GPUs [#12879](https://github.com/CesiumGS/cesium/issues/12879)
+- Fixed label sizing for some fonts and characters [#9767](https://github.com/CesiumGS/cesium/issues/9767)
+
+## 1.137 - 2026-01-05
 
 ### @cesium/engine
 
 #### Fixes :wrench:
 
 - Fixes label positioning in workflows that delete and recreate clamped labels [#12949](https://github.com/CesiumGS/cesium/issues/12949)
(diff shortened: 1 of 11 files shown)


javagl:
I wasn't aware of this class, and have used a ... \*clears throat\*... ["pragmatic approach"](https://github.com/CesiumGS/cesium/blob/95214adcd8d14a72756422e3f5d846711e7ee22e/packages/engine/Specs/Scene/SceneSpec.js#L88) occasionally. This _could_ have been avoided by using this class, but ... it doesn't seem to support textures (and even less extensions), so it would need some ... _extensions_. 

More generally: Creating differently-flavored, small glTFs for the specs is something that we should do more extensively ([another example where this could be useful](https://github.com/CesiumGS/cesium/issues/12108#issuecomment-2273476766)). Maybe pulling in glTF-Transform as a dependency (for the specs) could streamline some of that.

donmccurdy:
Yes, this is roughly the approach I've taken when generating test data for glTF Transform itself. Example generator here:

- https://github.com/donmccurdy/glTF-Transform/blob/main/packages/test-utils/src/create-torus-primitive.ts

Generally I think it's nice to have test assets generated at test time. As opposed to static test assets, where it can be easy to end up with test assets that we no longer fully understand or can't easily recreate.

ggetz:
Thanks for the cleanup @donmccurdy!

Issue #12965, opened 2025-10-10
WebGL errors occur when rendering Gaussian Splat


BlockCnFuture:
### What happened?

When rendering Gaussian Splat, WebGL errors occasionally appear in the browser console. As a result, the Gaussian models fail to render correctly in certain frames — they either disappear or flicker unexpectedly.

<img width="1482" height="633" alt="Image" src="https://github.com/user-attachments/assets/4301ac96-05f0-4407-85fb-6f87c155ccf5" />

Gaussian Splat should render consistently across all frames without WebGL errors or visual glitches.

### Reproduction steps

1. load Gaussian Splat
2. move camera

https://github.com/user-attachments/assets/e0d4c51b-ee81-433c-9654-68ffd071e467

This issue is more obvious when the data is large, its actually caused by the following code:
https://github.com/CesiumGS/cesium/blob/8817a88aed70707b079ec9ddf0876e8cc5a51c4c/packages/engine/Source/Scene/GaussianSplatPrimitive.js#L1017

When the dataset is large, a single frame may not be enough to complete the sorting process.
In this case, the update function calls `_sorterPromise.then` multiple times, which creates multiple then chains.
These chains interfere with each other and pollute the sorting result in subsequent frames.

The simple solution is to add
`this._sorterState = GaussianSplatSortingState.SORTING;`

### Sandcastle example

https://sandcastle.cesium.com/?id=3d-tiles-gaussian-splatting

### Environment

Browser: Chrome 139.0.7258.157
CesiumJS Version: 1.134
Operating System: win 11

BlockCnFuture:
it can also occur on small datasets — you just need to rotate the camera quickly. Please see:
[https://sandcastle.cesium.com/?id=3d-tiles-gaussian-splatting-comparison](https://sandcastle.cesium.com/?id=3d-tiles-gaussian-splatting-comparison)

https://github.com/user-attachments/assets/7e219017-d6ef-4c1d-8f14-dd0173a1e6e8

After adding `this._sorterState = GaussianSplatSortingState.SORTING;`, you can observe that the error disappears, because this ensures that multiple `.then` chains are not added.

BlockCnFuture:
<img width="838" height="1107" alt="Image" src="https://github.com/user-attachments/assets/4bde8a43-5d6e-4bae-a33c-6c41cb84177a" />

https://github.com/user-attachments/assets/29ed9629-743a-4dec-ac7f-3fca437e892f

PR #13120, opened 2026-01-06
Avoid WebGL error in splat rendering

javagl:
<sup>Credits to @BlockCnFuture for suggesting this fix. If you want your name to appear somewhere (e.g. in CONTRIBUTORS.md), drop me a note.</sup>

# Description

_\<!-- Describe your changes in detail --\>_

I added the line that was suggested in an issue report.

_\<!-- Consider: Why is this change required? What problem does it solve? --\>_

Without this line, WebGL errors are caused. 

To reproduce: 

- Open the 3D Tiles Gaussian splatting sandcastle, as it is currently deployed at https://sandcastle.cesium.com/index.html?id=3d-tiles-gaussian-splatting 
- Open the developer console by hitting "F12"
- Rotating the view is often enough to see the error, but to more quickly and "reliably" provoke the error: Run the sandcastle, and move the view around _while_ the data is still loading. 
- You'll see the messages

  > [.WebGL-0x724c00115400] GL_INVALID_OPERATION: Error: 0x00000502, in ..\..\third_party\angle\src\libANGLE\renderer\d3d\VertexDataManager.cpp, rx::VertexDataManager::reserveSpaceForAttrib:535. Internal error: 0x00000502: Vertex buffer is not big enough for the draw call.

  appearing in the console, together with tiles flickering and randomly disappearing. 

(The WebGL errors are caused at https://github.com/CesiumGS/cesium/blob/a6e02269689225d85fa0a6edefcbdafcf9cd40a9/packages/engine/Source/Renderer/Context.js#L1384 , pointing to https://github.com/google/angle/blob/9ab2e3ad572a1d7e0e453e5d6e27bc82f5c12fcd/src/libANGLE/renderer/d3d/VertexDataManager.cpp#L535 , that that's just an aside...)

## Issue number and link

This addresses https://github.com/CesiumGS/cesium/issues/12965 and https://github.com/CesiumGS/cesium/issues/13016 

I do **not** consider this as a proper fix for the actual issue. The actual issue is a race condition, and the change here avoids this issue (as indicated by the title of this PR and the branch name). The _reason_ for the issue is that the "update" function is performing 38 (thirty-eight!) modifications of an undocumented state space. The race condition here manifests itself when the renderer is trying to work on a state that is assumed to be consistent, but that was partially modified by the "update" function. "Fixing" that would be a lot of work.

## Testing plan

Follow the procedure from the description, and ensure that there are no WebGL errors in the console. (Make sure to close the old browser tab and open a new one during these tests - after a certain number of errors, Chrome may stop reporting the errors, so it's better to start with a clean slate). 

Alternatively: Load the `levels-3-count-64` tileset from this archive...

[tilesets-2025-11-18.zip](https://github.com/user-attachments/files/24459546/tilesets-2025-11-18.zip)

... (Batteri... Sandcastle included!) and just _zoom in_. It should wreak havoc pretty quickly. 

# Author checklist

There are no tests. How should that be tested? I think that it even hardly makes sense to _try_ and add tests for this. 
I can update the CHANGES. But I'd phrase this as "_Avoid_ WebGL errors...", and not "Fixed [anything]... ".

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Change Log
 
+## 1.138 - 2026-02-02
+
+### @cesium/engine
+
+#### Fixes :wrench:
+
+- Fixes jitter artifacts on Intel Arc GPUs [#12879](https://github.com/CesiumGS/cesium/issues/12879)
+- Fixed label sizing for some fonts and characters [#9767](https://github.com/CesiumGS/cesium/issues/9767)
+
 ## 1.137 - 2026-01-05
 
 ### @cesium/engine
--- a/packages/engine/Source/Core/writeTextToCanvas.js
+++ b/packages/engine/Source/Core/writeTextToCanvas.js
@@ -3,95 +3,41 @@ import Frozen from "./Frozen.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 
-function measureText(context2D, textString, font, stroke, fill) {
+/**
+ * Computes dimensions for text, based on current canvas state.
+ *
+ * Rounds metrics, excluding width, to whole pixels. This is purely to minimize
+ * rendering differences with migration to in-browser measureText(), and may be
+ * revised in the future. See: github.com/CesiumGS/cesium/pull/13081
+ */
+function measureText(context2D, textString) {
   const metrics = context2D.measureText(textString);
   const isSpace = !/\S/.test(textString);
 
-  if (!isSpace) {
-    const fontSize = document.defaultView
-      .getComputedStyle(context2D.canvas)
-      .getPropertyValue("font-size")
-      .replace("px", "");
-    const canvas = document.createElement("canvas");
-    const padding = 100;
-    const width = (metrics.width + padding) | 0;
-    const height = 3 * fontSize;
-    const baseline = height / 2;
-    canvas.width = width;
-    canvas.height = height;
-
-    const ctx = canvas.getContext("2d");
-    ctx.font = font;
-    ctx.fillStyle = "white";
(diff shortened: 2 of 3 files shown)


ggetz:
@mzschwartz5 Could you give this a review, with the context that this a workaround and not a fundamental fix?

mzschwartz5:
Interestingly, I'm getting some _other_ warnings besides the one mentioned in this issue (and _not_ the one mentioned in the issue):

`WebGL warning: generateMipmap: Tex image TEXTURE_2D level 0 is incurring lazy initialization.`

```
WebGL warning: drawArraysInstanced: Drawing without vertex attrib 0 array enabled forces the browser to do expensive emulation work when running on desktop OpenGL platforms, for example on Mac. It is preferable to always draw with vertex attrib 0 array enabled, by using bindAttribLocation to bind some always-used attribute to location 0. 
```

```
Source map error: Error: URL constructor:  is not a valid URL.
Stack in the worker:resolveSourceMapURL@resource://devtools/client/shared/source-map-loader/utils/fetchSourceMap.js:56:22
getOriginalURLs@resource://devtools/client/shared/source-map-loader/source-map.js:74:24
workerHandler/</<@resource://devtools/client/shared/worker-utils.js:115:52
workerHandler/<@resource://devtools/client/shared/worker-utils.js:113:13

Resource URL: wasm:https://sandcastle.cesium.com/js/engine/index.js%20line%2067168%20%3E%20WebAssembly.instantiate
```

That's in Firefox on my Mac. On Chrome, I get pretty much no console messages at all.

javagl:
This has become obsolete, with https://github.com/CesiumGS/cesium/pull/13189 already being merged.

Issue #12892, opened 2025-09-11
Create reference implementation for proposed glTF extension EXT_textureInfo_constant_lod


markschlosseratbentley:
### Feature

Depends on https://github.com/CesiumGS/glTF/pull/92.

Within CesiumJS, consider creating a reference implementation of the proposed glTF extension EXT_textureInfo_constant_lod as it is described at the above link.

This item should be completed after the above-linked PR is more fleshed out / fully proposed.

markschlosseratbentley:
@danielzhong Can you link the PR?

markschlosseratbentley:
Just working on implementing final PR feedback.

eringram:
This PR was merged
Also followed up by some small spec improvements: https://github.com/CesiumGS/glTF/pull/99

PR #13121, opened 2026-01-07
Implement constant LOD glTF extension

danielzhong:
# Description

This PR adds support for the new `EXT_textureInfo_constant_lod` glTF extension. See the extension spec PR https://github.com/CesiumGS/glTF/pull/92 which is in the final stages of review.

Constant level-of-detail ("LOD") is a technique of texture coordinate generation which dynamically calculates texture coordinates to maintain a consistent texel-to-pixel ratio on screen, regardless of camera distance. As the camera zooms in or out, the texture coordinates are recalculated so that the texture pattern remains at approximately the same visual scale. The transition between scale levels is smoothly blended to avoid abrupt changes. The general use case of this is to create textures that look good at different zoom levels, commonly for surfaces like grass or gravel on a ground plane in large models.

See the spec README for detailed info about the formulas used.

Notes:
- While according to the spec the extension can be present on any `textureInfo` property, this implementation only supports it on `baseColorTexture` and `normalTexture` to a limited extent.
  - The normal map texture limitation is that if a material has both a base color texture and a normal texture that use the extension, the base color texture's constant LOD properties (repetitions, offset, etc.) will also be applied to the normal texture and override the normal texture's properties if they are different.
  - This is done to match the iTwin.js implementation, and the intention is to keep the textures in sync as the texture coordinate blending/recalculation is done.
- The feature is **not** currently supported for `emissiveTexture`, `occlusionTexture`, or any other `textureInfo`

Model with `EXT_textureInfo_constant_lod` grass texture very zoomed in:

Screenshot 2026-01-27 123932

Same model zoomed out. Note how the texture remains at the same visual scale:

Screenshot 2026-01-27 123939

Gif from iTwin.js to illustrate the effect:

constant_lod

## Issue number and link
https://github.com/CesiumGS/cesium/issues/12892

## Testing plan

I added a new dev Sandcastle to test this feature for two different base color textures, as well as a normal map. See the test models in `Specs/Data/Models/glTF-2.0/ConstantLod/gltf/`. I also added some unit tests to GltfLoaderSpec.js.

Test models free texture sources:
- [Checker texture](https://polycount.com/discussion/186513/free-checker-pattern-texture)
- [Grass texture](https://3djungle.net/textures/grass/1416/)
- [Normal map texture](https://www.filterforge.com/filters/586-normal.html)

# Author checklist

### DIFF
--- a/packages/engine/Specs/Scene/Model/MaterialPipelineStageSpec.js
+++ b/packages/engine/Specs/Scene/Model/MaterialPipelineStageSpec.js
@@ -1078,6 +1078,7 @@ describe(
       ShaderBuilderTester.expectHasVertexUniforms(shaderBuilder, [
         "uniform vec2 u_constantLodOffset;",
         "uniform float u_constantLodDistance;",
+        "uniform mat4 u_constantLodWorldToEnu;",
       ]);
 
       ShaderBuilderTester.expectHasFragmentUniforms(shaderBuilder, [


eringram:
@javagl would you be able to review this PR, given you might have some context from reading the spec? Or possibly @lilleyse as this is to support a glTF extension for design models. No specific deadline or urgency, but would be great to get some eyes on it this week

javagl:
I did not yet _thoroughly_ read the spec, but will have a closer look (and at least try out this PR) soon. I'm sure Sean can provide more profound feedback about things like best practices for this, and the implementation approach in general.

packages/sandcastle/gallery/constant-lod-dev/main.js
```diff
@@ -0,0 +1,69 @@
...
+      minimumPixelSize: 128,
+      maximumScale: 20000,
+    },
+  });
+
+  viewer.trackedEntity = entity;
+}
+
+createModel(
+  "../../../Specs/Data/Models/glTF-2.0/ConstantLod/gltf/ConstantLod.gltf",
```
markschlosseratbentley: Is it me or does `ConstantLod.gltf` not exist? Is this line of code a holdover from some earlier version? Should it be removed?

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -289,6 +419,7 @@ function processTexture(
   uniformName,
   defineName,
   defaultTexture,
+  renderResources,
```
markschlosseratbentley, line 573: Should `renderResources` get added to the JSDoc? Maybe also add `defaultTexture` while at it. Edit: I see @javagl also mentioned this earlier.

CHANGES.md
```diff
@@ -14,6 +14,10 @@
 - Fixed a bug where entities have not been clustered correctly [#13064](https://github.com/CesiumGS/cesium/pull/13064)
 - Fixes multiple issues causing undefined pick results in 2D/CV scene modes [#13083](https://github.com/CesiumGS/cesium/issues/13083)
 
+#### Additions :tada:
+
+- Added support for [EXT_textureInfo_constant_lod](https://github.com/CesiumGS/glTF/pull/92) glTF extension. [13121](https://github.com/CesiumGS/cesium/pull/13121)
```
markschlosseratbentley: Other `CHANGES.md` entries seem to use `#` before other PR links.

```suggestion
- Added support for [EXT_textureInfo_constant_lod](https://github.com/CesiumGS/glTF/pull/92) glTF extension. [#13121](https://github.com/CesiumGS/cesium/pull/13121)
```

packages/engine/Specs/Scene/GltfLoaderSpec.js
```diff
@@ -4178,6 +4178,72 @@ describe(
       expect(clearcoatNormalTexture.texture.width).toBe(256);
     });
```
markschlosseratbentley, line 4315: The unit tests in `GltfLoaderSpec.js` look great. How much work do you think it would be to add a unit test to `MaterialPipelineStageSpec.js`?

eringram:
@javagl 
Thanks, I fixed the WEBP file and incorrect docs you mentioned.

> (Why does the emissive texture appear to be rotated by -90 degrees? Haven't investigated that yet...)

I'm confused by this as well and took a look to see if the orientation of the emissive texture was changed when I removed the extension. Here's the model as-is vs. when I remove this extension from both textures (base color & emissive). I think this indicates a bug with the orientation of the UVs produced by the constant LOD shaders?

| Constant LOD | No constant LOD |
| ------------- | ------------- |
| image  | image  |

> I could imagine that there are some tricky considerations for the cases where this extension is supposed to be combined with `KHR_texture_transform`. If it is not allowed, then this should be mentioned in the specification. If it is allowed, I could create some test asset for that.
 
I don't in theory see a problem with them being combined... I just did a quick test where I added the following to the ConstantLod_Checker.gltf's `baseColorTexture` that already has constant LOD:
```json
"KHR_texture_transform": {
  "offset": [0.1, 0.1],
  "scale": [2.0, 2.0],
  "rotation": 0.75
}
```
and it seems like `KHR_texture_transform` is completely ignored when constant LOD is also present. This is because the UVs calculated from texture transform on [this line](https://github.com/CesiumGS/cesium/blob/946575357a9de4a55edf8b13ba9b1339189da109/packages/engine/Source/Shaders/Model/MaterialStageFS.glsl#L188) aren't used in the constant LOD calculation a few lines later. One solution I can think of is applying the KHR texture transform to the constant LOD coords `tc1` and `tc2` [here](https://github.com/CesiumGS/cesium/blob/946575357a9de4a55edf8b13ba9b1339189da109/packages/engine/Source/Shaders/Model/ConstantLodStageFS.glsl#L23-L24) before they are mixed.

packages/sandcastle/gallery/constant-lod-dev/main.js
eringram: Oops yes it's old, cleaned this up

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
eringram: Added

markschlosseratbentley: Looks good, thanks.

packages/engine/Specs/Scene/GltfLoaderSpec.js
eringram: Not too bad, added one

javagl:
> if the orientation of the emissive texture was changed when I removed the extension. 
>...
> I think this indicates a bug with the orientation of the UVs produced by the constant LOD shaders?

I probably should have said that more explicitly: I think that this is **not** related to this PR. (It may be worth examining that further, and maybe opening a dedicated issue - and it's quite odd that this seems to happen _only_ for the emissive texture (until now) - I'll take a TODO on my list to check this, for emissive, occlusion, etc., and maybe open an issue (or PR) eventually)

 > and it seems like KHR_texture_transform is completely ignored when constant LOD is also present. 

I mentioned "tricky considerations", but the first one is relatively easy: Should it be possible to combine them (in a form where they both have the intended effect)? 

If the answer to that is "No", then one could simply say in the spec: "This extension cannot be combined with `KHR_texture_transform`" (and be done).

If the answer is "Yes, it should be possble to combine them", then the _tricky_ considerations come in - namely, about the _where_ and _how_. I have not yet thought that through. The second place that you linked to looks like one that could make sense - i.e. transforming the `tc1` and `tc2` with a `mat3 textureTransformMatrix` that could be passed in there. This `mat3` would have to be computed in the same way as it is done for other texture lookups, and be the identity matrix by default, but then people will scream "Oh dear, the matrix is computed twice, and usually an identity (without an effect)" - and then you'll have to insert another set of `#ifdef`s, because of (jazz hands:) performance. 

Maybe the answer to the first question is 'No'...?

eringram:
> I probably should have said that more explicitly: I think that this is **not** related to this PR. (It may be worth examining that further, and maybe opening a dedicated issue - and it's quite odd that this seems to happen _only_ for the emissive texture (until now) - I'll take a TODO on my list to check this, for emissive, occlusion, etc., and maybe open an issue (or PR) eventually)

Based on the screenshots I included, is the fact that the base color texture is a different orientation w/ and w/o this new extension not a concern? that's what made me think it's related to this PR, but I'm not familiar with any existing problems with UVs so could be wrong. Here's a clearer comparison of what I mean:

| Constant LOD | Emissive texture, no CLOD | No emissive, no CLOD
| ------------- | ------------- | ------------- |
| image  | image  | image |

(See the orientation of the numbers and letters in the texture)
So I guess in other words, I suspect the emissive texture is correct and CLOD shader is making the base color texture incorrect, because while it has `offset` and `repetitions` props, it shouldn't rotate anything.

> Maybe the answer to the first question is 'No'...?

I feel like the answer is "yes it should be possible, but out of the scope of this issue since it's not a common/high priority use case"-- I'm hesitant to explicitly say it's not allowed in the spec, but maybe we don't have to spend the time on implementing it yet since it hasn't come up until now when discussing design models that use `EXT_textureInfo_constant_lod`. And we can document in the code and an issue that they are currently mutually exclusive in CesiumJS.

@weegeekps do you think this is fair on the topic of combining `EXT_textureInfo_constant_lod` and `KHR_texture_transform`?

packages/engine/Specs/Scene/GltfLoaderSpec.js
markschlosseratbentley: Thanks! Looks good!

javagl:
**Ahhh** 💡  ... sorry, I completely missed that. (For some reason, I was only looking at that "Emissive" string and thought "Well, it's 'wrong' in both cases?!?") 

But now I see: Yes, apparently the extension "rotates" the texture coordinates, and that should not be the case. I said "rotates" in quotes, because it might simply be a mixup of `u` and `v` (given that they are likey both in [0,1], that could easily explain it).

Quickly skimming over the code: There's that `constantWorldToEnu` and `czm_model` at https://github.com/CesiumGS/cesium/pull/13121/changes#diff-1fb9ea4da146861f0b9c81c8dc1e5ed29aba51dcc3b772f57044800c267f85faR9 that may play a role here. There are different conventions for the "up-axis" in glTF and 3D Tiles, and CesiumJS has to deal with both of them, which is an endless source of confusion. Depending on which coordinate conventions these matrices are using, this seems like a likely source of errors (but that's just a pointer - I didn't do the math yet, and I don't say that the problem _is_ there, only that I'd start _looking_ there...)

eringram:
After discussion with @markschlosseratbentley (and his discussion with @weegeekps), this is a summary of remaining issues on this PR that I am working on:
1. Fixing the "rotated" coordinates bug and
2. Investigating how feasible it is to support both `EXT_textureInfo_constant_lod` and `KHR_texture_transform` together
    1. Current behavior: if both are present, `EXT_textureInfo_constant_lod` UV calculations completely override `KHR_texture_transform` (maybe other undefined behavior as well)
    2. Since CLOD is an `EXT_` prefixed extension, we want it to be compatible with other common extensions. So if it's not feasible to support both (i.e. seems like a large time investment), it may be changed to `BENTLEY_` in the spec.

eringram:
1. I've [fixed the UV bug](https://github.com/CesiumGS/cesium/pull/13121/changes/0c57b677549add28c608ebca9c1f517973c39db0#diff-1fb9ea4da146861f0b9c81c8dc1e5ed29aba51dcc3b772f57044800c267f85faR10) by swapping the X and Y in the vertex shader (and removing a subsequent axis flip), it seems like there was a mix-up between which was expected to be East/North from `enuDir`.  New screenshots:

| Constant LOD | Emissive texture, no CLOD | No emissive, no CLOD
| ------------- | ------------- | ------------- |
| image  | image  | image |

----

2. On compatibility with `KHR_texture_transform`:
    - I added a new overload function of `constantLodTextureLookup` in the fragment shader that takes a 3rd argument, `textureTransform`. When both `EXT_textureInfo_constant_lod` and `KHR_texture_transform` are present, this overload is used and `czm_computeTextureTransform(tc1, textureTransform)` is applied to the CLOD texture coordinates.
    - Added `czm_computeTextureTransform` as a new shader built-in to replace `computeTextureTransform`, which was previously defined in MaterialStageFS.glsl. I added this because in the case of CLOD being used only for base color, the CLOD shader didn't have access to the `computeTextureTransform` definition in MaterialStageFS.glsl. So this way the function is now accessible anywhere these built-ins are, and doesn't have to be redefined in ConstantLodStageFS.glsl.
    - Added glTF models for these new test cases with emissive texture, transform, and normal map with transform. Added because they were useful to test edge cases with, but lmk if it's excessive to include all of them.

| Constant LOD alone | Texture transform alone | Both texture transform & CLOD |
| ------------- | ------------- | ------------- |
| image  | image | image |

constant_lod_transform

Also tested normal map support for both texture transform and CLOD. I added a normal map that matches the checkerboard color texture so it would be easy to see if they are aligned.

image

The difference between these images is a little hard to see, but the normal map makes the right image's letters and some square borders appear darker/more pronounced. You can tell both textures are transformed since they are aligned.

| Both texture transform & CLOD, no normals | Both texture transform & CLOD w/ normal map |
| ------------- | ------------- |
| image  | image |

eringram:
@javagl Adam gave the interaction with `KHR_texture_transform` in this implementation his approval and said the final review goes to the Cesium team. Would you be able to take another pass at the code?

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -60,6 +64,23 @@ MaterialPipelineStage.process = function (
...
+    uniformMap.u_constantLodDistance = function () {
+      const boundingSphere = model.boundingSphere;
+      const camera = frameState.camera;
+
+      if (defined(boundingSphere) && defined(camera)) {
+        const distance = Cartesian3.distance(
+          camera.positionWC,
+          boundingSphere.center,
+        );
+        return Math.max(distance, 0.1);
```
javagl, line 79: I don't fully understand the significance of the values of `0.1` and `100.0`

javagl (commented):
Sorry for the delay here. Apparently, the re-review was already requested, but slipped through.

---

Prelude: 

I think that I pointed to the formatting error at https://github.com/CesiumGS/glTF/blob/7467e5a45725f8c40932fdb7945aad57a4a8bc93/extensions/2.0/Vendor/EXT_textureInfo_constant_lod/README.md?plain=1#L109 before, but am not 100% sure about that.

---

Main review:

The way how `KHR_texture_transform` is integrated looks reasonable. It causes some redundancy in `ConstantLodStageFS.glsl`, due to the function once receiving the transform and once not, but 1. the implementation should be fixed and final (no maintenance overhead) and 2. it's a trade-off after all - trying to avoid this redundancy with `#define`s or so wouldn't be nicer.

I tested the sandcastles, and they seem to work as expected. I also messed around with some obvious corner cases (model at the origin, extreme values, etc), and it didn't break. 

I found it a bit hard to map the math to the expected visual appearance. Specifically: I wondered whether there was a combination of `repetitions` and `offset` that causes the texture to appear exactly like the one without CLOD, at a certain zoom level. I found it with a bit of trial-and-error...
```
              "repetitions": 1.6,
              "offset": [
                5.0,
                5.0
              ],
```
Which gave this:

Image

It _might_ not be "important" for the PR (assuming that what is done in the PR matches the spec - I didn't do the math here). But I wonder if there is an obvious reason for these values that I'm overlooking.

What I am **not** sure about, however, is the intended behavior for the case that the model is rotated. Here, the test sandcastle (on the CI build) with the model having a roll of 89°:

https://ci-builds.cesium.com/cesium/daniel/constant_lod/Apps/Sandcastle2/index.html#c=zVZrT9swFP0rVj+FqTiFQYGuoE2FDWl0MIr2ZZkqk9w21hw7sp0CQ/z3XTuPpt2mSRsIokqN7+P4nOsb2zzLlbbkFWGGjMDwIiMzrTISdWI/ijpvIsnLoAmTScyMFVDHLC0+LpKxksaSBYcb0OSQSLipUOkXbwtq3JGSlnEJOups+MxZIWPLlSSxBmZhrBIQQaFFl6TA56ndIPeRJBU0BWm55WCohkwt4J0QQQlDSEkhV4Z7uMOawIhpi29MvqaO/THMNYAJNvd2aX+/3z/Y6ZKdHu31d/d295pJVzBTYAmX8yXkmNmUWnWJZiZNgNkuoaHAbZxiMJpbVq2E+CPC/sEqRJqvVfG0pHDhoC8RKag4dcvZuh6+jaA0x1qx1VJcaZxspnRmaLoG+LnA4muJ4YEDIU0hu+UQGfm31cr45bjDCdaXhyVJcF9mSpbBgLgVXQUerE/RojxoDyp35jpjQCpUgoi8DYsBHBu2yC74LYgJ/4GTbm3vL73s1nknMRPo2e7hU/ke/P9DraySYjWLv0NyUisspWLMw7LfVe74GfR+dZkVNQu3doAfySgFRNDXiukk6tQipQEBMQY0jR9sLEW1P4LaRhCL0rD8TXKITXjMLAt9lAnn4ur95jbthSPHiUl7phI02lnbMK3IUOdoyLin17yX7dPUo/sXSeSG25RI7CYmsLr5s0v85KmMWf6fIj9oZsyzq/EsHn25IOPG8AW8lH48qfg8mVCCJxU2KRmdnR+/NNHYsY7Wo2v/eHo5dd5Cw9TWW/5LEd+cQU+4FflFf5lVaDap6T8W4ps/ppYXMHfUXiklrpkegyyC6kxyiZ1uZ2jsnYAjl/y2us3hkRmgAAtZLlCjCa8LLKSlsfFJw7BOGSZ8QXhy+JurG4kF7k7omRXCn7VR52gYYvxKmlD+jnG+AC3YnQtJt47OSiOldBji8NcsW4ppIf4E

It currently looks like this:

Image

And I _think_ that this is not what it is supposed to look like...?

eringram:
> What I am **not** sure about, however, is the intended behavior for the case that the model is rotated. Here, the test sandcastle (on the CI build) with the model having a roll of 89°:
> And I _think_ that this is not what it is supposed to look like...?

You're right, it definitely should not be affected by the angle of the surface normal or rotation of the model. Kind of a facepalm moment, I remember Daniel mentioning this issue when he started the implementation and it totally slipped my mind.

I just pushed a fix for this. Basically:
- The old approach was based on an ENU reference frame that projected the vertex positions onto a world-space ground plane, and so didn't take into account the rotation of the model or the surface normal
- The fix uses the model-space position of the vertex, as well as using the surface normal to calculate the tangent/bitangent which are used to project the positions onto the actual surface regardless of orientation

I updated the sandcastle with buttons for rotating the model to test that case. I also added two more glTF models for this, `ConstantLod_Checker_Angled.gltf` and `ConstantLod_Checker_Vertical.gltf`, which have the plane's geometry (not an orientation applied in Cesium once the model is loaded) angled (rotated 45 degrees around X-axis) and vertical (facing the +Z-axis) compared to the ground plane.

image

image

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
eringram: The 0.1 is a safeguard against this returning 0, since this `u_constantLodDistance` is used in calculating `v_constantLodUvCustom.z`, which in `ConstantLodStageFS.glsl` could be passed into a `log2()` call:

```glsl
    float effectiveDistance = atMaxClamp ? constantLodParams.y : 
                              (atMinClamp ? constantLodParams.x : v_constantLodUvCustom.z);
    
    float logDepth = log2(effectiveDistance);
```
So it was an attempt to prevent `log2(0)` from breaking things. So it could be any small value, even though not sure if the distance between the camera and center of the bounding sphere can ever be zero anyway. I will add a comment to explain what it does.

The 100.0 is even more arbitrary and used in that `effectiveDistance` calculation before the camera or bounding sphere are defined, I guess before the model is fully loaded in. Maybe a better fallback is something like 1.0 which is the default `minClampDistance` in the spec.

javagl: The 0.1 to prevent NaN makes sense. (It is "obvious" in some way, but I wondered - it's a "magic constant" after all, and I probably wouldn't have wondered if it was an (even more) obvious epsilon like 0.00001. Similarly, 1.0 instead of 100.0 might not have raised the question, as it is just some `// Default value while nothing is rendered yet`). However, no need to change anything. 

(I'll have a closer look at the resolution of the orientation issue below (a bit later today or during the weekend))

javagl:
Note: Anybody can say "Hit 'approve' already!" at any point in time.

Until then, I'll post my observations here.

---

I had another pass of playing with the new sandcastle. 

As it can be seen in the first screenshot from the previous comment, the texture from `ConstantLod_Checker_Vertical.gltf` is mirrored. First, I wasn't sure what could be the reason for this. But in addition to playing with the new sandcastle, I looked more closely at the relevant change here, https://github.com/CesiumGS/cesium/pull/13121/changes/adf8f67b20e841a80c8acd6c43db660253c85806 . 

**Disclaimer**: I didn't do all the math. The exact meaning of the `attributes` and the coordinate systems that they are in, and mapping this to the specification text, can be hard by just _reading_ it (without looking up many other references and code paths, and without pen+paper). 

But I think that I roughly understand what's happening there, with one important caveat being that it now takes into account the _normal_ from the model, `attributes.normalMC`. 
That probably explains it. 
Looking at the two new models in BabylonJS, with "Display Vertex Normals" enabled, it shows the normals as small white lines:

Cesium Constant LOD Normals

So for the "vertical" one, the normals are pointing backwards, which (I think...) explains the mirrored texture.

---

Also, the issue about the "rotated" texture coordinates was re-introduced with the latest change. The left side shows the `ConstantLod_Checker_Angled.gltf` **without** the extension. The right side shows it **with** the extension:

Cesium Constant LOD Regression

I assume that this will be a small fix, similar to the previous one.

---

I also noticed that the different orientations are "baked" into the geometry. One major source of ~"confusion and things that behave unexpectedly in CesiumJS" is the transform matrices of glTF. They are usually correctly _applied_ to the geometry, visually, but some ~"other things" (most importantly: bounding volumes) are just wrong. I was curious about the behavior here, and created a small test that contains a unit square, attached to nodes with different transforms:

- once in its original form
- once rotated by 90 degrees
- once scaled in y-direction by a factor of 5, 

(Attached to nodes with appropriate translations, and a root node with rotation)

It currently looks like this:

Cesium Constant LOD Transforms

I assume that the behavior for the scaled one is the intended one: It scales "whatever is there", so when the original geometry and texture are "stretched", then the result is stretched. 

(The fact that in the non-scaled ones, that "B2" square is centered, and it is not centered in the scaled one, can be attributed to the `offset` from the extension - I checked that. And as I said in the previous review, mapping the extension properties to the expected visual appearance (with the _additional_ question about ~"order of operations" in terms of glTF node transforms) is not straightforward. It looks "correct", though...)

The asset, as an embedded glTF (so that it can be dropped into the specs folder), with a reference to the `"checker_texture.png"`:

```json
{
  "extensionsUsed": [
    "EXT_textureInfo_constant_lod"
  ],
  "accessors" : [ {
    "bufferView" : 0,
    "byteOffset" : 0,
    "componentType" : 5123,
    "count" : 6,
    "type" : "SCALAR",
    "max" : [ 3 ],
    "min" : [ 0 ]
  }, {
    "bufferView" : 1,
    "byteOffset" : 0,
    "componentType" : 5126,
    "count" : 4,
    "type" : "VEC3",
    "max" : [ 1.0, 1.0, 0.0 ],
    "min" : [ 0.0, 0.0, 0.0 ]
  }, {
    "bufferView" : 1,
    "byteOffset" : 48,
    "componentType" : 5126,
    "count" : 4,
    "type" : "VEC3",
    "max" : [ 0.0, 0.0, 1.0 ],
    "min" : [ 0.0, 0.0, 1.0 ]
  }, {
    "bufferView" : 1,
    "byteOffset" : 96,
    "componentType" : 5126,
    "count" : 4,
    "type" : "VEC2",
    "max" : [ 1.0, 1.0 ],
    "min" : [ 0.0, 0.0 ]
  } ],
  "asset" : {
    "generator" : "JglTF from https://github.com/javagl/JglTF",
    "version" : "2.0"
  },
  "buffers" : [ {
    "uri" : "data:application/gltf-buffer;base64,AAABAAIAAQADAAIAAAAAAAAAAAAAAAAAAACAPwAAAAAAAAAAAAAAAAAAgD8AAAAAAACAPwAAgD8AAAAAAAAAAAAAAAAAAIA/AAAAAAAAAAAAAIA/AAAAAAAAAAAAAIA/AAAAAAAAAAAAAIA/AAAAAAAAgD8AAAAAAACAPwAAgD8AAAAAAAAAAAAAAAAAAAAAAACAPwAAAAAAAAAA",
    "byteLength" : 156
  } ],
  "bufferViews" : [ {
    "buffer" : 0,
    "byteOffset" : 0,
    "byteLength" : 12,
    "target" : 34963
  }, {
    "buffer" : 0,
    "byteOffset" : 12,
    "byteLength" : 144,
    "byteStride" : 12,
    "target" : 34962
  } ],
  "images" : [ {
    "uri" : "checker_texture.png",
    "mimeType" : "image/png"
  } ],
  "materials" : [ {
    "pbrMetallicRoughness" : {
      "baseColorTexture" : {
        "index" : 0,
        "extensions": {
          "EXT_textureInfo_constant_lod": {
            "repetitions": 3.0,
            "offset": [
              0.5,
              0.5
            ],
            "minClampDistance": 0.0,
            "maxClampDistance": 100.0
          }
        }
      },
      "metallicFactor" : 0.0,
      "roughnessFactor" : 1.0
    },
    "doubleSided" : false
  } ],
  "meshes" : [ {
    "primitives" : [ {
      "attributes" : {
        "POSITION" : 1,
        "NORMAL" : 2,
        "TEXCOORD_0" : 3
      },
      "indices" : 0,
      "material" : 0,
      "mode" : 4
    } ]
  } ],
  "nodes" : [ {
    "children" : [ 1, 2, 4 ],
    "rotation" : [ 0.0, 0.7071067811865475, 0.0, -0.7071067811865476 ]
  }, {
    "mesh" : 0
  }, {
    "children" : [ 3 ],
    "translation" : [ -2.0, 0.0, 0.0 ]
  }, {
    "mesh" : 0,
    "rotation" : [ 0.0, 0.0, 0.707, 0.707 ]
  }, {
    "children" : [ 5 ],
    "translation" : [ 2.0, 0.0, 0.0 ]
  }, {
    "mesh" : 0,
    "scale" : [ 1.0, 5.0, 1.0 ]
  } ],
  "scene" : 0,
  "scenes" : [ {
    "nodes" : [ 0 ]
  } ],
  "textures" : [ {
    "source" : 0
  } ]
}
```

---

One could do further tests (like nodes with scaling attached to another node with rotation, or non-uniform scaling in the `modelMatrix` and whatnot), and brainstorm about the desired or intended behavior. But I think that some cases are already covered now.

---

Again: Nobody has to wait for me to hit "Approve" here. There might not even be a "finish line", so to speak...

eringram:
@javagl Finally have the time to respond to your feedback here.

We are hoping to support this extension in CesiumJS by the end of Q3. With that in mind, in my opinion we have thorough enough test cases covered in this PR so far to merge it.

In the future we'll be able to test more realistic design models that use this extension, once the mesh export service is updated to include encoding for it. So I definitely also have the possibility in mind to come back and improve this implementation if we notice any issues that come up with real-world data, or interactions with other extensions like we found with `KHR_texture_transform`. 

---

Fixed the normals on the vertical model so they are consistent with the angled one:

Screenshot 2026-03-18 131437

---

Also fixed the "rotated" texture coordinates problem again in the vertex shader-- based on the chosen reference vector, the tangent from `cross(ref, n)` aligns with U and the bitangent from `cross(n, tangent)` with V, so I had to switch them. Also had to reverse the V direction.

image

---

Based on the above rotated texture coordinates fix, the textures of your node transforms test model are shifted:

Screenshot 2026-03-18 131837

I think this is correct though, as now the rotation is as expected (letters are upright in the middle model, rotated 90 degrees in the left one, etc.)

---

Additional small improvement: I resized most of the textures because this PR adds a lot of them compared to most sandcastles.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: Does it make sense to perhaps change these values to named constants which indicate their meaning?

eringram: Updated to just add some comments here. I wanted to include the detail that part of the motivation is preventing the `log2(0)`, and it felt redundant add a comment with that and then a constant that was returned in the next line

markschlosseratbentley:
I defer to someone else for final merge / second approval, but I gave this another pass (my earlier comments are all resolved) and clicked approve.

javagl:
(A (probably final) review for this is on my TODO list, but based on the description, the most critical issues (e.g. the "rotation regression") seem to be resolved, and the tests should cover the most important cases in the meantime)

javagl:
There is the `ConstantLod_Checker.gltf`. And it is displayed properly (probably). 

Here is that exact same file, just with the extension removed:
```
{
  "asset": {
    "version": "2.0",
    "generator": "EXT_textureInfo_constant_lod example"
  },
  "extensionsUsed": [
    "EXT_textureInfo_constant_lod"
  ],
  "scene": 0,
  "scenes": [
    {
      "nodes": [
        0
      ]
    }
  ],
  "nodes": [
    {
      "mesh": 0
    }
  ],
  "meshes": [
    {
      "primitives": [
        {
          "attributes": {
            "POSITION": 0,
            "NORMAL": 1,
            "TEXCOORD_0": 3
          },
          "indices": 2,
          "material": 0
        }
      ]
    }
  ],
  "materials": [
    {
      "pbrMetallicRoughness": {
        "baseColorTexture": {
          "index": 0,
          "texCoord": 0
        },
        "metallicFactor": 0.0,
        "roughnessFactor": 1.0
      }
    }
  ],
  "textures": [
    {
      "source": 0
    }
  ],
  "images": [
    {
      "uri": "checker_texture.png",
      "mimeType": "image/png"
    }
  ],
  "accessors": [
    {
      "bufferView": 0,
      "componentType": 5126,
      "count": 4,
      "type": "VEC3",
      "max": [
        5.0,
        0.0,
        5.0
      ],
      "min": [
        -5.0,
        0.0,
        -5.0
      ]
    },
    {
      "bufferView": 1,
      "componentType": 5126,
      "count": 4,
      "type": "VEC3",
      "max": [
        0.0,
        1.0,
        0.0
      ],
      "min": [
        0.0,
        1.0,
        0.0
      ]
    },
    {
      "bufferView": 2,
      "componentType": 5123,
      "count": 6,
      "type": "SCALAR",
      "max": [
        3
      ],
      "min": [
        0
      ]
    },
    {
      "bufferView": 3,
      "componentType": 5126,
      "count": 4,
      "type": "VEC2",
      "max": [
        1.0,
        1.0
      ],
      "min": [
        0.0,
        0.0
      ]
    }
  ],
  "bufferViews": [
    {
      "buffer": 0,
      "byteOffset": 0,
      "byteLength": 48,
      "target": 34962
    },
    {
      "buffer": 0,
      "byteOffset": 48,
      "byteLength": 48,
      "target": 34962
    },
    {
      "buffer": 0,
      "byteOffset": 96,
      "byteLength": 12,
      "target": 34963
    },
    {
      "buffer": 0,
      "byteOffset": 108,
      "byteLength": 32,
      "target": 34962
    }
  ],
  "buffers": [
    {
      "uri": "data:application/octet-stream;base64,AACgwAAAAAAAAKDAAACgQAAAAAAAAKDAAACgQAAAAAAAAKBAAACgwAAAAAAAAKBAAAAAAAAAgD8AAAAAAAAAAAAAgD8AAAAAAAAAAAAAgD8AAAAAAAAAAAAAgD8AAAAAAAACAAEAAAADAAIAAAAAAAAAAAAAAIA/AAAAAAAAgD8AAIA/AAAAAAAAgD8=",
      "byteLength": 140
    }
  ]
}
```
Loading the original one and this one in a Sandcastle:

Cesium LOD Again and Again

The changes have been approved, so this PR could be merged. Maybe the intention is to let people open dedicated issues when they encounter rotated textures.

eringram:
Ok, so I apologize for the confusion, but there's been a misunderstanding on my part of what the expected behavior should be. TL;DR, the behavior seen all the way back [here](https://github.com/CesiumGS/cesium/pull/13121#issuecomment-3892025868) was correct according to the spec, and the texture stretching issues you pointed out in [this comment](https://github.com/CesiumGS/cesium/pull/13121#pullrequestreview-3816283285) for rotated models are expected limitations of it.

`EXT_textureInfo_constant_lod` [computes UVs](https://github.com/CesiumGS/glTF/tree/vendor-extensions/extensions/2.0/Vendor/EXT_textureInfo_constant_lod#formulas) from the vertex `worldPosition.xy` only, and I didn't really piece together until now that this makes it an explicit top-down planar projection. So any surface without meaningful extent in both horizontal axes (e.g. walls) will produce stretched UVs by design.

The iTwin.js reference implementation does the same thing in its equivalent shader [Surface.ts](https://github.com/iTwin/itwinjs-core/blob/30229f14dd2836ace8afb5c47bae026221031544/core/frontend/src/internal/render/webgl/glsl/Surface.ts#L257), and produces the same stretched texture for vertical surfaces, see this iTwin.js screenshot of one of these test models:

image

My previous approach was correct, and the intention of the spec is that authors only apply this extension to materials on roughly horizontal surfaces (terrain, floors, roads). It's not explicit enough about that clearly though, based on how confused I became trying to implement it in a new runtime.

Does that all make sense...? I will also open a PR to clarify this in the spec document and see what Adam thinks.

(Side note, I discovered another possible issue with the spec which is that glTF is Y-up, so if the intention is for the texture to be projected on horizontal surfaces, it should probably use worldPosition.x***z***, not xy... xy only makes sense for iTwin.js and CesiumJS specifically because they both use Z-up.)

To have this effect work on any orientation of surface is IMO more desirable, but outside the scope of this PR I think.

javagl:
> _javagl commented on Feb 3:_
> > I did not yet thoroughly read the spec, ...

I probably should have 😶 Apologies is the _assumption_ about the expected visual appearance caused unnecessary churn here. I'll have another (probably final final) pass.

> Side note, I discovered another possible issue with the spec which is that glTF is Y-up,

I had quickly looked at the spec PR, and saw that Y->Z change. There is always some confusion with the different axis conventions... and that on top of the "world" coordinate system in CesiumJS actually being the ENU. But I _think_ that this spec change makes sense in the context of glTF.

javagl:
I just noticed that the "rotation" issue (mentioned in the previous review comment) apparently was not changed. I could _try_ to match that with the formulas of the spec. But I'll either 1. wait until another review is requested _explicitly_, or 2. wait until it is confirmed that the rotated texture is intentional, or 3. leave the review+merge to someone else.

javagl:
The rotation issue that I referred to in the last comment was the one that I had observed by removing the CLOD extension from the assed, as shown in [an earlier comment](https://github.com/CesiumGS/cesium/pull/13121#issuecomment-4103939121). I now checked it again (maybe didn't pull last time?) and it seems to be resolved:

Cesium CLOD resolved

Sorry for the noise

PR #13125, opened 2026-01-08
Types: Bulk conversion to ES6 classes

donmccurdy:
# Description

Draft for testing in CI, not for review. A large batch conversion to ES6 classes, related to #13104, without opting in to type checks yet. Intended to test (in CI) the limits of bulk automated conversion.

**lebab-batch.js**

```javascript
// @ts-check

import { transform } from "lebab";
import { glob } from "glob";
import { readFile, writeFile } from "node:fs/promises";
import { resolve, dirname, basename } from "node:path";
import { fileURLToPath } from "node:url";

const __dirname = dirname(fileURLToPath(import.meta.url));

const pattern = process.argv[2];

if (!pattern) {
  console.error(`Usage: ${process.argv[1]} path/to/*.js`);
  process.exit(2);
}

for await (const path of glob.sync(resolve(__dirname, pattern))) {
  console.log(`${path}...`);

  // Lowercase filenames are assumed to contain functions, not classes. If
  // properties are assigned to functions, lebab may assume they're classes.
  const pathBasename = basename(path);
  if (!pathBasename.match(/^[A-Z]/)) {
    continue;
  }

  const input = await readFile(path, "utf-8");

  const { code: output, warnings } = transform(input, ["class"]);

  if (warnings.length > 0) {
    console.warn(`Warnings encountered on ${path}, skipping.`, warnings);
    continue;
  }

  await writeFile(path, output);
}

console.log("✅ Done");
```

Manual fixes:

- `git checkout main -- packages/engine/Source/Core/*{TerrainProvider,TilingScheme,Resource}.js` need manual conversion
- configured ESLint's use-before-define error to allow scratch variables to be used within the same class they instantiate
- typo "Sterographic -> Stereographic" fixed, raised errors after conversion
- typo "new Matrix3.clone(...)" fixed, and other similar typos, grepping for `new [\w]+\.[a-z]`

## Issue number and link

- #8359 

## Testing plan

TODO

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [ ] I have performed a self-review of my code

#### PR Dependency Tree

* **PR #13104**
  * **PR #13125** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

PR #13126, opened 2026-01-08
Types: Add type checks in Cartesian2.js

donmccurdy:
# Description

Based on #13104, a proof-of-concept conversion of Cartesian2 to ES6 classes, with type checking enabled.

## Issue number and link

TODO

## Testing plan

TODO

# Author checklist

#### PR Dependency Tree

* **PR #13104**
  * **PR #13126** 👈

### DIFF
--- a/packages/engine/Source/Core/Cartesian2.js
+++ b/packages/engine/Source/Core/Cartesian2.js
@@ -1,7 +1,12 @@
+// @ts-check
+/* eslint no-use-before-define: ["error", { "variables": false }] */
 import Check from "./Check.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 import CesiumMath from "./Math.js";
+
+/** @import Cartesian3 from './Cartesian3.js'; */
+/** @import Cartesian4 from './Cartesian4.js'; */
 
 /**
  * A 2D Cartesian point.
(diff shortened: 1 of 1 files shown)


jjspace:
@donmccurdy It looks like the change to not allow `any` is catching this PR during linting. Can you take a peak when you get a chance?

donmccurdy:
@jjspace sure, looks like those will be quick fixes. And good to see that the `noImplicitAny` rule is (so far?) mostly flagging spots where useful types are missing. I am wondering about whether we want to review and merge this PR with its current scope though – it might be odd to have Cartesian2 be an ES6 class but not the other math classes. For ease of review, maybe next steps to try could be:

1. PR 1: just convert a core set of math classes to ES6 classes without other changes (run lebab)
2. PR 2: add types to the converted classes

jjspace:
> it might be odd to have Cartesian2 be an ES6 class but not the other math classes

I think this is likely to happen regardless if we continue with the piece-wise approach of only a few files in random other PRs. 

I'm not opposed to broadening this PR to include more of the base math type classes that change very infrequently if you want.

Just let me know your plans and when it's ready for a full review

javagl:
>> it might be odd to have Cartesian2 be an ES6 class but not the other math classes
> 
> I think this is likely to happen regardless if we continue with the piece-wise approach of only a few files in random other PRs.

Maybe I'm not fully up to date with the intended process. I've seen the "test PR" at https://github.com/CesiumGS/cesium/pull/13125 , but not looked through all changes. In any case, one thing that I'd consider to be important: **If** and **when** a class is changed to the `class` syntax (even if it's one of 142 classes that are bulk-processed with 'lebab'), we should carefully review it.

The specific thing that I'd have in mind is to check for what I consider to be an "anti-pattern", namely the [file-scoped functions](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#private-functions) that receive that implicit `this`. Yes, that's what the Watcom compiler did in the 90's when it was fed with C++ code. But maybe, maybe, we can make these functions (private) _instance_ functions, have a proper `this`, and omit all the `// Exposed for testing` cases that resulted from this guideline.

donmccurdy:
Hm, I don't mean to suggest that adoption of type checking must be done through piecewise changes in unrelated PRs. Just that if incremental adoption is possible, adoption will go more smoothly. Personally I might prefer to write PRs that take <1hr to review, somewhere between this PR and #13125 (both proof of concepts for illustration), but if bulk changes are needed or preferred we can make that happen, too.

One constraint is that ES6 classes and prototype-based inheritance do not mix well in the same inheritance tree. If a base class is converted, all of its subclasses must be converted at the same time.

I am very hesitant to mix hand refactoring (like removing file-scoped functions) into the same PR as automated AST-based changes with lebab, though. I worry the potential for error goes up too much. It would be safer to do that sort of change after type checks are in place, and ES6 Class adoption is a blocker in the way of that.

javagl:
> I am very hesitant to mix hand refactoring (like removing file-scoped functions) into the same PR as automated AST-based changes 

I see the point. Ideally, the automated changes should have "zero effect on the semantics" (and **if** they have, it will already be hard enough to pinpoint that). When adding manual tweaks in the same PR, the question of where a regression is coming from has another dimension, and identifying whether it's due to the automated or the manual part will be close to impossible. My baseline concern (even at the risk of sounding cynical): When the automated translation is done, and the conversion from file-scoped functions to member functions is "only a cleanup (that can happen later (when we have more time))", then it _will_ not happen.

donmccurdy:
For a change like that, if the team is generally agreed, I might suggest something like the [code "ratcheting" system](https://www.notion.com/blog/how-we-evolved-our-code-notions-ratcheting-system-using-custom-eslint-rules) used at Notion. This can be as simple as creating a new ESLint rule to check for such functions, and using [`eslint-seatbelt`](https://github.com/justjake/eslint-seatbelt/) to require that violations trend downward. Then a very gradual migration likely wouldn't require a larger prioritization effort.

javagl:
That 'ratchet' system sounds interesting. It prevents both cases that I've observed far too frequently:

- There are 43241 warnings in a codebase. After a change, there are 43242 warnings. Nobody cares. (The difference between 0 and 1 is more important for this)
- A certain lint rule _should_ be enabled, but it _is_ not enabled, because "it causes so many warnings". (Yeah, that's... well...)

The actual approach sounds pretty complicated, but when it's well-tested and the complexity is nicely hidden behind some ~"standard ESLint infrastructure" (as they suggest), then this may not matter. 

One thing to keep an eye on might be that the CesiumJS codebase is **huge**, and such a TSV/CSV file that they are creating there, with numFiles ⨉ numRules, could easily have several (maybe dozens of) megabytes. Depending on where and how this is processed, exactly, I could imagine it to slow down e.g. commit hooks. But ... maybe that's not an issue either. We're not trying to increase the "commits per second" here 🙂

donmccurdy:
Let's start with:

- https://github.com/CesiumGS/cesium/pull/13168

PR #13131, opened 2026-01-11
docs: document classes implementing GeometryFactory

### LINKED ISSUE #12827: Document classes implementing the `GeometryFactory` interface
The [GeometryFactory type](https://cesium.com/learn/cesiumjs/ref-doc/GeometryFactory.html) is defined as if it was meant to be describing an interface. But no other class is listed as implementing the interface, so a reader of the docs has no way of knowing what classes are acceptable for use in (for example) the [GeometryInstance constructor](https://cesium.com/learn/cesiumjs/ref-doc/GeometryInstance.html#GeometryInstance).

See [this older comment](https://github.com/CesiumGS/cesium/issues/8924#issuecomment-640276903) for some discussion on the original intent of the `GeometryFactory` interface.

See https://github.com/CesiumGS/cesium/issues/12826 for an example of where a better doc for `GeometryFactor` could have helped avoid confusion.

MohammadShujaullah:
Fixes #12827

 – Document classes implementing the `GeometryFactory` interface.

## Description

This PR improves the documentation of `GeometryFactory` and its usage:

- Clarifies that `GeometryFactory` is an abstract interface-like type implemented by geometry description classes with a static `createGeometry` method.
- Adds `@see` links in `GeometryFactory` to the main implementing geometry classes.
- Expands `GeometryInstance.options.geometry` docs to explain `Geometry` vs `GeometryFactory` (e.g. `RectangleGeometry`, `EllipsoidGeometry`) and how `createGeometry` is used.

 

## Testing plan

- [ ] `npm test` (or tests recommended in CONTRIBUTING)

## Author checklist

- [  ] I have added my name to `CONTRIBUTORS.md`

### DIFF
--- a/packages/engine/Source/Core/GeometryFactory.js
+++ b/packages/engine/Source/Core/GeometryFactory.js
@@ -1,21 +1,70 @@
 import DeveloperError from "../Core/DeveloperError.js";
 
 /**
- * Base class for all geometry creation utility classes that can be passed to {@link GeometryInstance}
- * for asynchronous geometry creation.
+ * Describes a geometry type that can be converted into a {@link Geometry}.
+ * <p>
+ * Implementations of this interface are the core geometry "description" classes
+ * such as {@link BoxGeometry}, {@link RectangleGeometry},
+ * {@link EllipsoidGeometry}, and other classes with a static
+ * <code>createGeometry</code> function. Instances of these classes can be
+ * passed to {@link GeometryInstance} and {@link Primitive} to have their
+ * vertices and indices created either asynchronously on a web worker or
+ * synchronously on the main thread.
+ * </p>
+ * <p>
+ * This type describes an interface and is not intended to be instantiated
+ * directly.
+ * </p>
  *
+ * @alias GeometryFactory
  * @constructor
- * @class
  * @abstract
+ *
+ * @see BoxGeometry
+ * @see BoxOutlineGeometry
+ * @see CircleGeometry
+ * @see CircleOutlineGeometry
+ * @see CoplanarPolygonGeometry
+ * @see CoplanarPolygonOutlineGeometry
+ * @see CorridorGeometry
+ * @see CorridorOutlineGeometry
+ * @see CylinderGeometry
+ * @see CylinderOutlineGeometry
+ * @see EllipseGeometry
+ * @see EllipseOutlineGeometry
+ * @see EllipsoidGeometry
+ * @see EllipsoidOutlineGeometry
+ * @see FrustumGeometry
+ * @see FrustumOutlineGeometry
+ * @see GroundPolylineGeometry
+ * @see PlaneGeometry
+ * @see PlaneOutlineGeometry
+ * @see PolygonGeometry
+ * @see PolygonOutlineGeometry
+ * @see PolylineGeometry
+ * @see PolylineVolumeGeometry
+ * @see PolylineVolumeOutlineGeometry
+ * @see RectangleGeometry
+ * @see RectangleOutlineGeometry
+ * @see SimplePolylineGeometry
+ * @see SphereGeometry
+ * @see SphereOutlineGeometry
+ * @see WallGeometry
+ * @see WallOutlineGeometry
  */
 function GeometryFactory() {
   DeveloperError.throwInstantiationError();
 }
 
 /**
- * Returns a geometry.
+ * Creates a {@link Geometry} from a geometry description.
+ * <p>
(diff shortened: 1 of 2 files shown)


MohammadShujaullah:
@cla-bot check

packages/engine/Source/Core/GeometryFactory.js
```diff
@@ -1,21 +1,70 @@
 import DeveloperError from "../Core/DeveloperError.js";
 
 /**
- * Base class for all geometry creation utility classes that can be passed to {@link GeometryInstance}
- * for asynchronous geometry creation.
+ * Describes a geometry type that can be converted into a {@link Geometry}.
+ * <p>
```
jjspace, line 5: You shouldn't need to add the `<p>` tags yourself. This should be handled by JSDoc itself AFAIK.

packages/engine/Source/Core/GeometryFactory.js
```diff
@@ -1,21 +1,70 @@
...
+ * passed to {@link GeometryInstance} and {@link Primitive} to have their
+ * vertices and indices created either asynchronously on a web worker or
+ * synchronously on the main thread.
+ * </p>
+ * <p>
+ * This type describes an interface and is not intended to be instantiated
+ * directly.
+ * </p>
  *
+ * @alias GeometryFactory
```
jjspace, line 19: I don't believe this needs an alias since it matches the function name itself

packages/engine/Source/Core/GeometryFactory.js
```diff
@@ -1,21 +1,70 @@
...
+ * @see PolylineGeometry
+ * @see PolylineVolumeGeometry
+ * @see PolylineVolumeOutlineGeometry
+ * @see RectangleGeometry
+ * @see RectangleOutlineGeometry
+ * @see SimplePolylineGeometry
+ * @see SphereGeometry
+ * @see SphereOutlineGeometry
+ * @see WallGeometry
+ * @see WallOutlineGeometry
```
jjspace, line 53: I'm kinda on the fence about the usefullness of this simply because it's such a big list. I could see it easily going out of sync with the actual classes. I don't think there's been much change in this area so maybe it doesn't matter.

@donmccurdy or @javagl this is sorta related to recent discussions we've had around classes and inheritance. Any thoughts here?

donmccurdy: Thank you @MohammadShujaullah for compiling the list! However I think I agree with @jjspace, I'd (personally) be hesitant to commit to maintaining such a list by hand.

I can't speak for every code editor, but using TypeScript language services in Zed, there are "Find All References" and "Go to Implementation" tools built into the editor. These don't work on GeometryFactory for me today (TS language service limitations with JSDoc probably), unless I either:

- (a) add `@extends GeometryFactory` annotations to the implementations, and also add type-only imports as described in #13104
- (b) convert to ES6 classes

Getting that information out of the language server, and into published docs, might require more steps — if that's the goal. But at least we'd have confidence that the list will remain up to date.

If we do merge #13104, adding the missing `@extends X` or `@implements X` annotations could be a good next step?

javagl: No strong opinion on _this particular case_.

Pro:

- Consistency: Implementing classes are already listed in other cases (.e.g https://cesium.com/learn/cesiumjs/ref-doc/Property.html )
- Stability: It will not change. Or is anyone up for creating a new implementation of `GeometryFactory`?

Con:

- Maintenance (but that's a weak point, given the 'Stability' above)
- It _may_ become obsolete with better tooling around TS/JSDoc (but that may take months or years...)

Some of the back-and-forth links are currently broken anyhow (e.g. `InterpolationAlgorithm` or `Packable` - see https://github.com/CesiumGS/cesium/issues/11749#issuecomment-2343560106  ). As long as we don't have a clean(er) concept of visibility and automatisms for generating such links, I think that adding the list here is OK, and be it only to know into which classes we'll have to insert `@extends GeometryFactory` after the corresponding tooling was set up 🤷‍♂️

jjspace: Thanks for the input @donmccurdy and @javagl. I think given the stability of this area which hasn't changed much I'm going to lean towards "any info is better than no info" for now. These can stay.

We should definitely look into improving this area in the future since things like this should be automated. This may come "free" or easier after some of the `class` and/or TS refactoring @donmccurdy mentioned but that's far outside the scope for this PR.

PR #13133, opened 2026-01-12
chore: deprecate IE detection helpers in FeatureDetection

### LINKED ISSUE #12615: Deprecate IE related functions from FeatureDetection.js
Support for IE was discontinued in [Cesium 1.84](https://github.com/CesiumGS/cesium/blob/main/CHANGES.md#breaking-changes-mega-30)

As a result, `isInternetExplorer` and other IE related functions from `FeatureDetection.js` can safely be deprecated.

MohammadShujaullah:
Fixes #12615

## Description

Deprecates the Internet Explorer detection helpers in `FeatureDetection`:

- `FeatureDetection.isInternetExplorer()` now always returns `false` and is marked `@deprecated`.
- `FeatureDetection.internetExplorerVersion()` now always returns `undefined` and is marked `@deprecated`.

This also removes now-unnecessary user agent parsing for Internet Explorer, simplifying the feature detection code path.

Cesium no longer supports Internet Explorer, so these helpers no longer need to perform real detection.

## Testing

- No behavioral change expected for supported browsers.

### DIFF
--- a/Specs/DomEventSimulator.js
+++ b/Specs/DomEventSimulator.js
@@ -185,94 +185,34 @@ function createPointerEvent(type, options) {
   options = options ?? Frozen.EMPTY_OBJECT;
   let event;
 
-  if (FeatureDetection.isInternetExplorer()) {
-    const canBubble = options.canBubble ?? true;
-    const cancelable = options.cancelable ?? true;
-    const view = options.view ?? window;
-    const detail = options.detail ?? 0;
-    const screenX = options.screenX ?? 0;
-    const screenY = options.screenY ?? 0;
-    const clientX = options.clientX ?? 0;
-    const clientY = options.clientY ?? 0;
-    const ctrlKey = options.ctrlKey ?? false;
-    const altKey = options.altKey ?? false;
-    const shiftKey = options.shiftKey ?? false;
-    const metaKey = options.metaKey ?? false;
-    const button = options.button ?? 0;
-    const relatedTarget = options.relatedTarget ?? null;
-    const offsetX = options.offsetX ?? 0;
-    const offsetY = options.offsetY ?? 0;
-    const width = options.width ?? 0;
-    const height = options.height ?? 0;
-    const pressure = options.pressure ?? 0;
-    const rotation = options.rotation ?? 0;
-    const tiltX = options.tiltX ?? 0;
-    const tiltY = options.tiltY ?? 0;
-    const pointerId = options.pointerId ?? 1;
-    const pointerType = options.pointerType ?? 0;
-    const hwTimestamp = options.hwTimestamp ?? 0;
-    const isPrimary = options.isPrimary ?? 0;
-
-    event = document.createEvent("PointerEvent");
-    event.initPointerEvent(
-      type,
-      canBubble,
-      cancelable,
-      view,
-      detail,
-      screenX,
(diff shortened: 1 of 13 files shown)


MohammadShujaullah:
@cla-bot check

packages/engine/Source/Core/FeatureDetection.js
```diff
@@ -88,34 +88,18 @@ function webkitVersion() {
   return isWebkit() && webkitVersionResult;
 }
 
-let isInternetExplorerResult;
-let internetExplorerVersionResult;
+/**
+ * @deprecated Cesium no longer supports Internet Explorer. This function always returns false.
+ */
 function isInternetExplorer() {
```
jjspace: We should be able to fully remove these functions. As stated in the issue these were only ever used internally so don't need to be deprecated

packages/engine/Source/Core/FeatureDetection.js
```diff
@@ -88,34 +88,18 @@ function webkitVersion() {
...
-        theNavigator.userAgent,
-      );
-      if (fields !== null) {
-        isInternetExplorerResult = true;
-        internetExplorerVersionResult = extractVersion(fields[1]);
-      }
-    }
-  }
-  return isInternetExplorerResult;
+  return false;
```
jjspace: Given that this will be removed we should look for every place in the code that these functions were used and remove the branching paths or checks that were there as well. We don't want that extra "dead code" to keep cluttering up the repo

jjspace:
@MohammadShujaullah is this ready for another review?

MohammadShujaullah:
yes it is ready for review sir

Specs/DomEventSimulator.js
```diff
@@ -185,94 +185,34 @@ function createPointerEvent(type, options) {
...
-      rotation: options.rotation ?? 0,
-      tiltX: options.tiltX ?? 0,
-      tiltY: options.tiltY ?? 0,
-      pointerId: options.pointerId ?? 1,
-      pointerType: options.pointerType ?? 0,
-      hwTimestamp: options.hwTimestamp ?? 0,
-      isPrimary: options.isPrimary ?? 0,
-    });
-  }
+  event = new window.PointerEvent(type, {
```
jjspace, line 188: Now that this isn't actually modified it can probably become a `const` assignment. Maybe even a direct return

javagl:
I haven't dissected the Git history issues and changes here. From quickly scrolling over the files, it _looks_ like the relevant changes here are affecting "few" files (e.g. the changes related to https://github.com/CesiumGS/cesium/pull/13131 are rather isolated).

Two thoughts for resolving this:

- Wait until https://github.com/CesiumGS/cesium/pull/13131 is merged. If these are the same commits, then Git should be able to sort that out, and not count these changes as "changes from _this_ branch" any more 🤞 
- If everything else fails: Check out a fresh clone. Copy the files from _this_ branch that should really be changed into the fresh clone. Commit these changes as a single, new commit (to go into a new PR)

Yeah, both approaches are very "pragmatic", but ... inspired by https://xkcd.com/1597/ , when there are the options to sweat and tremble while doing some bisect/rebase/cherryPick black Git magic, or just do some copy+paste to have one clean commit with close-to-zero effort, I'm usually leaning towards the latter ...

PR #13134, opened 2026-01-12
docs: fix broken GeoJSON Sandcastle links in JSDoc#13122

### LINKED ISSUE #13122: Broken Sandcastle links in JSDoc comments
### What happened?

As reported at https://community.cesium.com/t/cesium-documentation-links-broken-to-sandcastle/44774.

@jjspace we appear to be doing something to replace the IDs, that works for most Sandcastles. For example,
https://sandcastle.cesium.com/index.html?src=Geometry%20and%20Appearances.html
redirects to
https://sandcastle.cesium.com/index.html?id=geometry-and-appearances

However, for these two, the redirection doesn't work. 
https://sandcastle.cesium.com/index.html?src=GeoJSON%2520and%2520TopoJSON.html
https://sandcastle.cesium.com/index.html?src=GeoJSON%2520simplestyle.html

Both report `Could not find Sandcastle gallery item with ID...`

I haven't yet done an exhaustive check, so there could be more.

### Reproduction steps

1. Go to https://sandcastle.cesium.com/index.html?src=GeoJSON%2520simplestyle.html
2. See error message in Sandcastle console

### Sandcastle example

https://sandcastle.cesium.com/index.html?src=GeoJSON%2520simplestyle.html

### Environment

Browser: Chrome
CesiumJS Version: 1.137
Operating System: Windows

MohammadShujaullah:
## Description
Fixes #13122

Expands the fix to address all Sandcastle demo links that have the `%20` (space encoding) issue causing double-encoding in generated documentation, not just the originally identified 2 links.

Changed all instances of:
- `?src=Some%20Demo.html` → `?id=some-demo`

## Verification
Tested multiple new URLs directly in Sandcastle - all load correctly.

### DIFF
--- a/packages/engine/Source/DataSources/GeoJsonDataSource.js
+++ b/packages/engine/Source/DataSources/GeoJsonDataSource.js
@@ -578,8 +578,8 @@ function processTopology(dataSource, geoJson, geometry, crsFunction, options) {
  * @param {string} [name] The name of this data source.  If undefined, a name will be taken from
  *                        the name of the GeoJSON file.
  *
- * @demo {@link https://sandcastle.cesium.com/index.html?src=GeoJSON%20and%20TopoJSON.html|Cesium Sandcastle GeoJSON and TopoJSON Demo}
- * @demo {@link https://sandcastle.cesium.com/index.html?src=GeoJSON%20simplestyle.html|Cesium Sandcastle GeoJSON simplestyle Demo}
+ * @demo {@link https://sandcastle.cesium.com/index.html?id=geojson-and-topojson|Cesium Sandcastle GeoJSON and TopoJSON Demo}
+ * @demo {@link https://sandcastle.cesium.com/index.html?id=geojson-simplestyle|Cesium Sandcastle GeoJSON simplestyle Demo}
  *
  * @example
  * const viewer = new Cesium.Viewer('cesiumContainer');
--- a/packages/engine/Source/Widget/CesiumWidget.js
+++ b/packages/engine/Source/Widget/CesiumWidget.js
@@ -175,7 +175,7 @@ function configureCameraFrustum(widget) {
  *
  * @exception {DeveloperError} Element with id "container" does not exist in the document.
  *
- * @demo {@link https://sandcastle.cesium.com/index.html?src=Cesium%20Widget.html|Cesium Sandcastle Cesium Widget Demo}
+ * @demo {@link https://sandcastle.cesium.com/index.html?id=cesium-widget|Cesium Sandcastle Cesium Widget Demo}
  *
  * @example
  * // For each example, include a link to CesiumWidget.css stylesheet in HTML head,
--- a/packages/widgets/Source/CesiumInspector/CesiumInspector.js
+++ b/packages/widgets/Source/CesiumInspector/CesiumInspector.js
@@ -17,7 +17,7 @@ import CesiumInspectorViewModel from "./CesiumInspectorViewModel.js";
  * @param {Element|string} container The DOM element or ID that will contain the widget.
  * @param {Scene} scene The Scene instance to use.
  *
- * @demo {@link https://sandcastle.cesium.com/index.html?src=Cesium%20Inspector.html|Cesium Sandcastle Cesium Inspector Demo}
+ * @demo {@link https://sandcastle.cesium.com/index.html?id=cesium-inspector|Cesium Sandcastle Cesium Inspector Demo}
  */
 function CesiumInspector(container, scene) {
   //>>includeStart('debug', pragmas.debug);
--- a/packages/widgets/Source/I3SBuildingSceneLayerExplorer/I3SBuildingSceneLayerExplorer.js
+++ b/packages/widgets/Source/I3SBuildingSceneLayerExplorer/I3SBuildingSceneLayerExplorer.js
@@ -11,7 +11,7 @@ import I3SBuildingSceneLayerExplorerViewModel from "./I3SBuildingSceneLayerExplo
  * @param {string} containerId The DOM element ID that will contain the widget.
  * @param {I3SDataProvider} i3sProvider I3S Data provider instance.
  *
- * @demo {@link https://sandcastle.cesium.com/index.html?src=I3S%20Building%20Scene%20Layer.html|I3S Building Scene Layer}
+ * @demo {@link https://sandcastle.cesium.com/index.html?id=i3s-building-scene-layer|I3S Building Scene Layer}
  */
 function I3SBuildingSceneLayerExplorer(containerId, i3sProvider) {
   const container = document.getElementById(containerId);
--- a/packages/widgets/Source/Viewer/Viewer.js
+++ b/packages/widgets/Source/Viewer/Viewer.js
@@ -351,7 +351,7 @@ function enableVRUI(viewer, enabled) {
  * @see Timeline
  * @see viewerDragDropMixin
  *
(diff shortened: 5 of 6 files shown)


MohammadShujaullah:
@cla-bot check

javagl:
From quickly looking over the GItHub search results, it seems like this was the only case where such a "double-percent-encoding" was applied. So I think that this could be merged quickly.

There is a bit of inconsistency in whether or not the other links use `%20` or ` ` for "space", but that doesn't affect the outcome here. I did _not_ systematically check _all_ links, but the ones that I tested (both with `%20` and ` `) worked, and properly forwarded to the new, `id`-based version.  There is no _strong_ need to update _all_ existing links, but I added a note that we _could_ do this in https://github.com/CesiumGS/cesium/issues/11749#issuecomment-3738896741

jjspace (commented):
Thanks for the PR @MohammadShujaullah.

Just stating for completeness but please remove the extra changes from your other PRs so we can focus on only the doc updates.

I'm perfectly happy with this version of a fix if that's the route we want to take. I also think it'd be fine to just replace the `%20` with an actual space ` ` character in the doc links

```
 * @demo {@link https://sandcastle.cesium.com/index.html?src=GeoJSON and TopoJSON.html|Cesium Sandcastle GeoJSON and TopoJSON Demo}
 * @demo {@link https://sandcastle.cesium.com/index.html?src=GeoJSON simplestyle.html|Cesium Sandcastle GeoJSON simplestyle Demo}
 ```

As @javagl says we probably convert all the sandcastle all links to the `id=` format regardless but I'm not going to ask for that in this PR.

That said...

> it seems like this was the only case where such a "double-percent-encoding" was applied.

These are _not_ the only links that are broken. If you use VSCode or other editor that allows for regex project wide searches try searching for `(<?@link.*)%20` and you can see there's a bunch of links we have that encounter this problem. Even some that are not sandcastle links like the reference in [`IntersectionTest.rayTriangle`](https://cesium.com/learn/cesiumjs/ref-doc/IntersectionTests.html#.rayTriangle). I think it'd be great to address all of these in one PR if you're able to go through the rest.

**Comment by javagl** [comment]
> These are not the only links that are broken

Whoa, I searched for the wrong thing! 

I thought that the double-encoding was done in the _source_. But I now realized that the `.js` file _does_ contain only the `%20`. Apparently, the double-encoding happens when the documentation HTML is generated!

So this affects **all** links that contain that `%20`, and that is many of them!

Is this something that

- changed in the JSDoc/HTML generation recently? (unlikely)
- was not noticed with the _old_ sandcastles, but kicks in for the _new_ ones?

For example:

- `CesiumWidget` uses 
  `* @demo {@link https://sandcastle.cesium.com/index.html?src=Cesium%20Widget.html|Cesium Sandcastle Cesium Widget Demo}`
  with the `%20`. 
- The generated HTML contains the additional percent-encoding of that
  `https://sandcastle.cesium.com/index.html?src=Cesium%2520Widget.html`
- Clicking that leads to the sandcastle, which then complains in the console
  `Could not find Sandcastle gallery item with ID "Cesium%20Widget.html".`

Note that the message does _not_ contain the double-encoded version, but only the `%20`. 

EDIT: I just saw your message at https://github.com/CesiumGS/cesium/issues/13122#issuecomment-3746434741 😬  Ignore the following right now: 

@jjspace Could the mapping function that translates these URLs into the `id` handle this? I can imagine that this is quirky and may not be appropriate, but I'm wondering about the most robust solution here. I could schedule a pass to update the JSDocs, but all the _existing_ (old) JSDoc links would then still not work. If that could be fixed in one place, by tweaking that link-to-`id`-translation somehow, that could help to keep these links working...

**Comment by javagl** [comment]
@jjspace It looks like this PR is now supposed to update all links with `%20` in them, and based on your comment at https://github.com/CesiumGS/cesium/issues/13122#issuecomment-3746434741 , that's likely one part of a "good" solution. 

(Checking whether it's possible to tweak the JSDoc HTML generation to _not_ do that double-encoding for cases where a link does have that `%20` may be independent of that)

**Comment by MohammadShujaullah** [comment]
Hey, that's a really smart idea about fixing the generator! I think both approaches work well together - like having a band-aid for the cuts we can see right now (manual fixes) while maybe figuring out how to prevent the cuts from happening in the first place (generator fix). The manual route was nice because I could test each link actually works, but your generator idea would definitely be cooler for the long run!

**Review by jjspace — CHANGES_REQUESTED** [review]
Thanks for the updates here @MohammadShujaullah! I've verified all the updated urls work and look correct.

I'm still seeing the Geometry JSDoc updates on this PR, please remove them. If you need help with the git commands/process let us know.

I'm also still seeing a handful of other links that would fall victim to the same double encoding in our docs

**Comment by jjspace** [comment]
Regarding the doc generation itself. I think it could be good to look into so it doesn't bite us later. At a quick glance it seemed like the encode step was happening deeper into the JSDoc code/processes than we have access to but I could be wrong. Either way I don't have the time to dig into it myself but can review a PR if you feel like taking the time.

**Comment by MohammadShujaullah** [comment]
Hi, I've cleaned up the branch as requested. I've removed the GeometryFactory documentation changes from this PR so it now only contains the Sandcastle link fixes. I've also expanded the fixes to include all the additional broken links mentioned. The branch has been force-pushed with these changes. Please let me know if there's anything else that needs to be addressed.

**Comment by MohammadShujaullah** [comment]
If the issue persists, could you please provide the specific git commands/process to ensure the branch is completely clean of the unwanted changes?

**Comment by jjspace** [comment]
> If the issue persists, could you please provide the specific git commands/process to ensure the branch is completely clean of the unwanted changes?

Certainly! Right now the state of this branch looks like this from where it diverges from the `main` branch in the CesiumJS repo:

```mermaid
gitGraph
    commit id: "9e1f702"
    branch fix-sandcastle-links
    checkout fix-sandcastle-links
    commit id: "b63382d - clarify Geometry..."
    commit id: "8318825 - fix broken Sand..."
```

If you use VSCode I recommend the "Source control graph" in their [Source Control panel](https://code.visualstudio.com/docs/sourcecontrol/overview) or an extension like [Git Graph](https://marketplace.visualstudio.com/items?itemName=mhutchie.git-graph) to visualize this. Other editors will likely have similar tools. If you want to view it purely on the command line then [`git log`](https://www.atlassian.com/git/tutorials/git-log) is your friend (try `git log --graph --oneline --decorate`). 

For this PR we want to keep the commit `8318825 - fix broken Sandcastle links` but get rid of the changes in commit `b63382d - clarify GeometryFactory`. The easiest way to do this would probably be to do an interactive [`git rebase`](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) on the main branch using the command `git rebase -i main`. `Drop` the first commit to remove it from this branch and `Pick` to keep the second commit.

It also looks like somewhere along the way the previous commit with the majority of the link changes was lost. This is a potential danger of using rebases and force pushing. It's possible you still have it saved somewhere in your local git history but it might be gone for good :cry: It's good to verify changes to branches or commits locally before pushing using  `git log` or the graphs above and [`git diff`](https://www.atlassian.com/git/tutorials/saving-changes/git-diff) or testing in an entirely separate branch. If something goes wrong you can always pull it back down or restart the branch.

MohammadShujaullah:
Hi @jjspace 👋  

I’ve opened a new PR (#13147) that builds on this work and addresses the feedback / issues discussed here.  
This PR is created from a fresh branch and contains only the updated implementation.

Closing this PR in favor of #13147.  
Thanks for the review!

Issue #13129, opened 2026-01-09
Support for more component types in metadata textures


mzschwartz5:
**Current state**: 

For the structural metadata extension, texels can encode metadata via property textures. Currently, the only supported metadata property type is `UINT8`. A texture can encode a (fixed-size) array of uints (interpreted in the shader as `ivecN`), or normalized uints (interpreted as floats), but it cannot use multiple channels to represent higher-precision numbers.

**Desired state:**

Represent higher precision ints and floats (up to 32 bits - using all 4 channels of each texel), as well as bools and smaller fixed-size arrays of higher precision (e.g. ivec2 with half-precision). Strings and dynamic arrays will not be supported still. 

The underlying texture will still be RGB(A)8 format, but the data from each channel will be bit-shifted to form a higher precision format.

Some misc. notes:
- The structural metadata spec states: 
> For unsigned integer component types, values are normalized between `[0.0, 1.0]`. For signed integer component types, values are normalized between `[-1.0, 1.0]`

Currently, since we only support the `UINT8` component type, [we only consider the first case](https://github.com/CesiumGS/cesium/blob/9e1f70201baa7f1190fa5b2a28c418e6389899f4/packages/engine/Source/Scene/PropertyTextureProperty.js#L273-L286) here. We'll need to also consider normalized signed integer types going forward.

- webgl 1 does not support uint types in glsl

mzschwartz5:
The solution I've been prototyping follows the pseudocode laid out by the [Property Textures](https://github.com/CesiumGS/glTF/tree/proposal-EXT_structural_metadata/extensions/2.0/Vendor/EXT_structural_metadata#property-textures) section of the structural metadata spec:

```
vec4<uint8> rgba = texture(sampler, coordinates);
uint8 byte0 = rgba[channels[0]];
uint8 byte1 = rgba[channels[1]];
uint8 byte2 = rgba[channels[2]];
uint8 byte3 = rgba[channels[3]];
uint32 rawBits = byte0 | (byte1 << 8) | (byte2 << 16) | (byte3 << 24);
float32 value = uintBitsToFloat(rawBits);
```

i.e. Extract bytes from all used channels, join them together, and implement and use `uintBitsToType` function for various types.

In webgl 1, since there are no bitwise operators, you can just multiply by powers of two. However, the much bigger restriction is this:

> webgl 1 does not support uint types

[...]

PR #13135, opened 2026-01-13
Initial support for higher precision property texture types

mzschwartz5:
# Description

Until now, [property textures](https://github.com/CesiumGS/glTF/tree/proposal-EXT_structural_metadata/extensions/2.0/Vendor/EXT_structural_metadata#property-textures) only supported `UINT8` data. You could pack multiple properties into a single texture, using the `rgba` channels to represent up to 4 `UINT8` properties, but you couldn't do the inverse - use multiple channels to represent a single property of higher precision.

With this PR, you can now represent 32-bit types, including the full-range of unsigned integers (previously, even if a glTF specified `UINT` as a property's data type, it would be cast to `int` on the glsl side, to be compatible with WebGL1, which doesn't have a `uint` type, and thus could lose precision). 

Note: these improvements _are_ restricted to a WebGL 2 context.

Note 2: this represents a breaking change for users using a WebGL 2 context who have custom shaders that treat unsigned integer data as signed.  

TODO before merging: decide how to address the breaking change (feature flag? We need this feature to be available pretty soon for borehole visualization. If we use a feature flag, we need it to be user-settable, but the `MetadataPipelineStage` is internal...)

Or just document it as a breaking change, potentially? The classes surrounding metadata and property textures are all tagged as experimental and subject to change without the typical deprecation policy.

## Issue number and link

#13129

## Testing plan

On your local machine, copy [this gltf file](https://gist.github.com/mzschwartz5/435a412c5733a36aed546906d49f9733) into the `Apps/SampleData` folder and name it `SimplePropertyTexture`.

Also copy this [tileset.json](https://github.com/user-attachments/files/24656085/tileset.json) file to the same directory (and name it `tileset.json`)

[Start cesiumjs and open this sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html#c=lVVhb9s2EP0rV2Mf5EGlHDtGW9cNtiUIWmBph9rtl2loaPFkc6NIgaTsJIX/+06k5Nhth3aADYnne3fv7t3RsqqN9fAzcAeX6GRTQWlNBfmgCKd88DLXuS6Mdh62Endo4RVo3HXe7GOwJb3/pdGeS402HwwDMsvg0iL3CH5DX6nQoU+BawGV2SJID94AhwJtC4TaOOml0UCfFrFWZoU9gQ5ODCIV5grUyGorKwJt0TEuRJJrAL7jFLnjGB+Tq2WEs7bCD1YFR4Dbjff1LMuUKbjaGOdnz0fPR9mvde2yBa9qhVfc86zLzf52Rt+mEfo5PgAErpr1YmN2v5lGC6nXH41qKpyBtw12zvv2Mcx125c+WGUEqhvurbyjojq6S8u1K42tHEPu/FsSaPOhXppreYfi2vIKA/O+OG49vXE9CXVd4doiuuTpsyk7m44n42kKkxfsxfmz0XkKoy5/bKcpy9jNIz1fI2/5/yF9sXnP9TrmGqVHGYnvhnnznhyJaPJ0PGbTYXA4Z6MYv5PnwZhqaZKD6jFhHIyy0UUQugjjsdhwQXNUW1PTJNy/pSpTcKaxBd5IfXjld8PY9ViBC6jTCi4b503VxesUKi1fV6h9tC7xzs/gthdva6Q4eNzQECbX3eGNrhsPpQvPFKQ2dCweqk+9bmglV1B1L8M+4mEuKLMynDaHqwaJZheKVei5oKliP30+rnj/8hHYB2VClmXjWvQWi0kSQg0PjvtuiEP/99Fu0TdWd80hy/50EY1GoOLsPdCQhSVTMkxDh3BkI84Fj6gVgkOFhUdBHQj+gjgLs9NATWrYf2j5rm5NJL5XJOUPC9uR71rog1QxRrSQ7oHNDA5Zk+Fjx2UJyZNuEgSWdBWJk6EaDo/V6fewOJoZajTtcIQ+thlQkQbfh/7faf5SyH0UslfteMxjQx0l+TNswNe9zgdXWPJGeWjNtMb5IH0shtafjVI4Y6O4rd8M8EY7KRCWWBFzTlJgGyMfyGD/whzijafTGPGvsNcLutoLurYUtpfx0hi14vaG5iQ5qaItfJAO5s7fK7xo6fwi439RQzczY5mnVIr4uWzVFP+0jXYBNM96yFzILUjx6hv/PVAo7hz9UjZKLeQDsb2YZ+R/AqPVbJv0botW8fvWZXN28Xs0MsbmGR2/RvlY0lHEfwE)

By default, it should display nonsense - that's the gradient texture when the `RGBA` channels are interpreted literally as colors. If you toggle the view to "Inside Temperature", you now see a grayscale gradient - each pixel of the texture is reinterpreted as a single 32-bit float, which has been constructed / populated to appear as a gradient when used as brightness.

image

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,11 +7,15 @@
 #### Breaking Changes :mega:
 
 - Cartesian2, Cartesian3, and Cartesian4 are now [ES6 Classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes). This change should have no impact on most users, but note that using `new` on a static factory method, like `new Cartesian3.fromArray(...)`, will now throw an error. Omit `new` unless you are invoking a constructor directly, for these and all other factory methods, as more classes will be migrated to ES6 Classes soon. [#8359](https://github.com/CesiumGS/cesium/issues/8359)
+- [Custom Shaders](https://cesium.com/learn/cesiumjs/ref-doc/CustomShader.html?classFilter=customsh) that rely on metadata derived from the [EXT_structural_metadata extension](https://github.com/CesiumGS/glTF/tree/proposal-EXT_structural_metadata/extensions/2.0/Vendor/EXT_structural_metadata) no longer cast
+  unsigned integer metadata types to signed integers. Any existing custom shaders that assign UINT-type metadata to local integers (e.g. `int myMetadata = vsInput.metadata.myUintMetadata`) will no longer compile. Variable assignments must be changed to reflect the underlying signedness of the metadata type.
+  [#13135](https://github.com/CesiumGS/cesium/pull/13135)
 
 #### Additions :tada:
 
 - Added `AttributeCompression.encodeRGB8` and `decodeRGB8` for packing colors. [#13174](https://github.com/CesiumGS/cesium/pull/13174)
 - Added more depth testing options for billboards and labels with `BillboardCollection.coarseDepthTestDistance`, `BillboardCollection.threePointDepthTestDistance`, `LabelCollection.coarseDepthTestDistance`, and `LabelCollection.threePointDepthTestDistance`. [#12994](https://github.com/CesiumGS/cesium/pull/12994)
+- Added support for more metadata types via property textures in custom shaders. See this [issue](https://github.com/CesiumGS/cesium/issues/10248) for the current state of supported types. [#13135](https://github.com/CesiumGS/cesium/pull/13135)
 
 #### Fixes :wrench:
 


mzschwartz5:
Python script I used to generate the test data:

```py
import base64
import numpy as np
from PIL import Image

def float_to_rgba_bytes_array(f_arr):
    f32 = f_arr.astype(np.float32) # f_arr: float32 numpy array shape (H, W)
    f32 = f32.astype('<f4')  # explicit little-endian float32
    bytes_view = f32.view(np.uint8).reshape(f32.shape + (4,))  # shape (H, W, 4)
    # bytes_view[0] is the first byte in memory; little-endian means least significant byte first
    # PNG expects channel order R,G,B,A, so map directly:
    rgba = bytes_view.copy()
    return rgba  # dtype uint8

def create_gradient_png(path, width, height):
    # Horizontal and vertical gradients from 0.0 .. 1.0 combined into one float
    xs = np.linspace(0.0, 1.0, width, dtype=np.float32)
    ys = np.linspace(0.0, 1.0, height, dtype=np.float32)
    row = xs[np.newaxis, :]                 # shape (1, W)
    col = ys[:, np.newaxis]                 # shape (H, 1)
    grid = (row + col) / 2.0                # shape (H, W), values 0..1 blending horizontal+vertical
    rgba = float_to_rgba_bytes_array(grid)
    # Ensure alpha is not premultiplied/modified; bytes already contain the float bits
    img = Image.fromarray(rgba, mode='RGBA')
    img.save(path, format='PNG', optimize=False)
    print(f"Wrote {path}  ({width}x{height})")

    with open(path, 'rb') as f:
        b64 = base64.b64encode(f.read()).decode('ascii')
    data_uri = 'data:image/png;base64,' + b64
    print(data_uri)
    return data_uri

if __name__ == '__main__':
    create_gradient_png('float_rgba_gradient.png', width=256, height=256)

```

packages/engine/Source/Scene/MetadataComponentType.js
```diff
@@ -54,18 +54,14 @@ const MetadataComponentType = {
    */
   UINT32: "UINT32",
   /**
-   * A 64-bit signed integer. This type requires BigInt support.
-   *
-   * @see FeatureDetection.supportsBigInt
+   * A 64-bit signed integer.
```
mzschwartz5, line 58: Removed this (and some other related stuff) as bigint is [widely supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) now.

packages/engine/Source/Scene/MetadataComponentType.js
```diff
@@ -87,11 +83,89 @@ const MetadataComponentType = {
   FLOAT64: "FLOAT64",
 };
 
+export const ScalarCategories = {
+  INTEGER: "int",
+  UNSIGNED_INTEGER: "uint",
+  FLOAT: "float",
+};
+
+MetadataComponentType.typeInfo = {
```
mzschwartz5, line 93: Rather than using lots of switch statements whenever we want to know if a component type is an integer, vector compatible, etc., I refactored this class to use object maps.

packages/engine/Source/Scene/PropertyTextureProperty.js
```diff
@@ -195,64 +198,157 @@ PropertyTextureProperty.prototype.isGpuCompatible = function () {
...
+const unsignedIntegerTypesByComponentCount = [
+  undefined,
+  "uint",
+  "uvec2",
+  "uvec3",
+  "uvec4",
+];
+
+// Map from scalar component type to the GLSL function used to reinterpret from uint bits to the scalar type
+const uintBitsToScalarType = {
```
mzschwartz5, line 253: Better name suggestion welcome...

packages/engine/Source/Scene/PropertyTextureProperty.js
```diff
@@ -195,64 +198,157 @@ PropertyTextureProperty.prototype.isGpuCompatible = function () {
...
+  }
+
+  if (MetadataComponentType.isUnsignedIntegerType(componentType)) {
+    return unsignedIntegerTypesByComponentCount[componentCount];
+  }
+
+  return integerTypesByComponentCount[componentCount];
+};
+
+PropertyTextureProperty.prototype.unpackInShader = function (
```
mzschwartz5, line 282: This is the meat of this PR. It's admittedly quite complex, and I'm open to suggestions on how to simplify and improve this bit (or maybe that's just how dynamically generated code is...)

The goal of this function is to write code to sample and unpack a property value from a texture. Such code can take various, slightly different forms, depending on the data types involved, texture channels used, etc. 

The end result in the shader is meant to look something like this (pseudo-code from the [structural metadata extension spec](https://github.com/CesiumGS/glTF/tree/proposal-EXT_structural_metadata/extensions/2.0/Vendor/EXT_structural_metadata#property-textures)):

```
vec4<uint8> rgba = texture(sampler, coordinates);
uint8 byte0 = rgba[channels[0]];
uint8 byte1 = rgba[channels[1]];
uint8 byte2 = rgba[channels[2]];
uint8 byte3 = rgba[channels[3]];
uint32 rawBits = byte0 | (byte1 << 8) | (byte2 << 16) | (byte3 << 24);
float32 value = uintBitsToFloat(rawBits);
```

or this (in the case of a `VEC2` property):

```
vec4<uint8> rgba = texture(sampler, coordinates);
uint8 byte0 = rgba[channels[0]];
uint8 byte1 = rgba[channels[1]];
uint8 byte2 = rgba[channels[2]];
uint8 byte3 = rgba[channels[3]];
value[0] = byte0 | (byte1 << 8);
value[1] = byte2 | (byte3 << 8);
```

With some extra stuff for declaring variables, casting to the right type, normalizing if necessary... etc

mzschwartz5: Also, to complicate things a bit further, a property can be an array, a vector-type, or an array-of-vector-type!

In the shader, all of these simply end up as a `vecN` type (up to a vec4  - anything bigger will fail validation)

packages/engine/Source/Shaders/Builtin/Functions/unpackTexture.glsl
```diff
@@ -0,0 +1,35 @@
+/**
```
mzschwartz5, line 1: We have some builtins like this already: `unpackFloat`, `unpackUint`. They seem to be targeted at WebGL 1 whereas this builtin targets webgl2 and is type-agnostic (just unpacks to uint bits).

I'm open to another name besides `unpackTexture`, since this is technically usable for other purposes.

packages/engine/Specs/Scene/Model/MetadataPipelineStageSpec.js
```diff
@@ -272,6 +278,10 @@ describe(
     });
 
     it("Adds property textures to the shader", function () {
+      if (!context.webgl2) {
+        return;
+      }
```
mzschwartz5, line 283: Guarding these behind webgl2 check as the CI webgl stub seems to run a WebGL 1 context. (Obviously not ideal...)

packages/engine/Specs/Scene/MetadataComponentTypeSpec.js
```diff
@@ -103,20 +90,6 @@ describe("Scene/MetadataComponentType", function () {
...
-  it("getMaximum returns approximate number for UINT64 when BigInt is not supported", function () {
-    spyOn(FeatureDetection, "supportsBigInt").and.returnValue(false);
-    expect(MetadataComponentType.getMaximum(MetadataComponentType.UINT64)).toBe(
-      Math.pow(2, 64) - 1,
-    );
-  });
-
   it("getMaximum throws without type", function () {
     expect(function () {
       MetadataComponentType.getMaximum();
```
javagl, line 95: Okay... I'm not entirely sure how to deal with that.

Below, there's the test
"getMinimum throws if type is not a numeric type"
that checks whether
`MetadataComponentType.getMinimum(MetadataComponentType.STRING);`
throws a `DeveloperError`.

And that passes. But looking at the code, I wondered: **Why** does it pass? 

The [`getMinimum` function](https://github.com/CesiumGS/cesium/blob/151f939dddbb98bfbeebd7a9c4835228ec6185f3/packages/engine/Source/Scene/MetadataComponentType.js#L182) is currently doing a
`return MetadataComponentType.typeInfo[type].minimumValue;`
and given that `typeInfo` does not contain an entry for `STRING`, this should throw a `TypeError`.

But... `MetadataComponentType.STRING` does not exist, meaning that 
`MetadataComponentType.getMinimum(MetadataComponentType.STRING);`
is actually
`MetadataComponentType.getMinimum(undefined);`
and that **does** throw a `DeveloperError` only because of the
`Check.typeOf.string("type", type);`
in `getMinimum`.

--- 

Regardless of what ~"is currently done", and _yes_, even regardless of which "regressions/'breaking changes'" this causes, I think that we should try to move towards doing _the right thing_. And that is: The `getMinimum`/`getMaximum`... functions should handle "unexpected values" in a predetermined way. For example:
```
/**
 * ...
 * @throws DeveloperError When the given type is not a valid numeric type [TODO maybe list them...]
 *
 * @private
 */
MetadataComponentType.getMinimum = function (type) {
  Check.typeOf.string("type", type); //---------------- TODO maybe omit this?
  //>>includeEnd('debug');
  const typeInfo = MetadataComponentType.typeInfo[type];
  if (!defined(typeInfo)) {
    throw new DeveloperError(`Type ${type} is not a numeric type`); // Proper message, with the "type"!
  }
  return typeInfo.minimumValue;
};
```

Specifically, the behavior should be "defined" (and spec'ed) for
- `getMinimum(undefined)`
- `getMinimum("STRING")`
- `getMinimum("INVALID_TYPE_FOR_SPEC")`
- `getMinimum("UINT64")`
- `getMinimum("UINT32")`

Note: 

We can talk about the details. For example: `isIntegerType` already _does_ check whether the `typeInfo` is `undefined`, meaning that
`isIntegerType("INVALID_TYPE_FOR_SPEC")`
would return `false` (and **not** throw!). 
We could also say that 
`getMinimum("INVALID_TYPE_FOR_SPEC")`
should not throw, but return `undefined`. But I'd argue against that: It would _hide_ the root cause of errors - for example, when someone accidentally passed in `"VEC3"` to this function (mixing up `type` and `componentType`).

packages/engine/Source/Scene/PropertyTextureProperty.js
```diff
@@ -195,64 +198,157 @@ PropertyTextureProperty.prototype.isGpuCompatible = function () {
...
 
-  if (MetadataType.isVectorType(type) || type === MetadataType.SCALAR) {
-    if (componentType !== MetadataComponentType.UINT8) {
-      oneTimeWarning(
-        `Property texture property ${classProperty.id} has component type ${componentType}, but only UINT8 is supported`,
-      );
-      return false;
-    }
-    return true;
+  if (type === MetadataType.STRING) {
```
javagl, line 210: I remember 👴 that this was once all "fail silent", and there had been quite a few cases where I've been stumped by "Why is nothing happening here?". I added these `oneTimeWarning` cases in https://github.com/CesiumGS/cesium/commit/a44ac07cfffe10774f0b23adb7d5afdb44d07b01 based on what was supported back then.

Looking at the new code, I'll have to check whether/how `type==="BOOLEAN"` and `type==="ENUM"` are handled. (Maybe they are, implicitly - but I'll have to read the shader generation part more thoroughly to understand whether there's already a shader-side handling of that)

packages/engine/Source/Scene/PropertyTextureProperty.js
```diff
@@ -195,64 +198,157 @@ PropertyTextureProperty.prototype.isGpuCompatible = function () {
...
-  // are supported.
-  oneTimeWarning(
-    `Property texture property ${classProperty.id} has an unsupported type`,
-  );
-  return false;
+  // For all other properties, make sure the components fit in the sampled channels.
+  // (Note that it's possible to treat 64-bit types as two 32-bit components, but for now that's not supported)
+  const componentCount = MetadataType.getComponentCount(type);
+  const arrayLength = classProperty.isArray ? classProperty.arrayLength : 1;
+  const bytesPerComponent = MetadataComponentType.getSizeInBytes(componentType);
```
javagl, line 221: (See other comment as well): `componentType` may be `undefined` here, e.g. when `type==="BOOLEAN"`.

Apps/Sandcastle/gallery/Custom Shaders Property Textures.html
mzschwartz5, line 1: This is an example of the sort of breaking change this PR causes.

This sandcastle has a custom shader that assigns the metadata to `int` type variables. However, the metadata is actually `UINT8` type in the tileset. Before this PR, we would cast to `int` because that's all that's supported by WebGL 1. After this PR, we cast it to `uint`, the actual type, so the custom shader has to change its assignments.

packages/engine/Source/Scene/PropertyTextureProperty.js
mzschwartz5: Having tested everything together in #13163, I can say that, in fact: `ENUM` and `BOOLEAN` types do *not* work in this PR.

In theory, `ENUM` types _should_ work just fine, but I didn't realize they don't have a `componentType` (they instead have their own special `valueType`, which we convert to a `componentType`). That gets fixed in #13163.

`BOOLEAN`s, on the other hand, are complicated by the fact that the spec says they should be packed into buffers as bitstreams. The new `unpackInShader` function I wrote, which uses the new `czm_unpackTexture` function, does not account for this packing scheme. It operates on the assumption that values are packed into one or more bytes, not bits. 

It's theoretically possible to support but, to avoid scope creep, I added a check to `isGpuCompatible` in #13163 that excludes booleans too.

packages/engine/Source/Scene/PropertyTextureProperty.js
mzschwartz5: Yep - wish I had read this before debugging the issue myself!
Addressed in #13163

packages/engine/Specs/Scene/MetadataComponentTypeSpec.js
mzschwartz5: I'm with you, and I think your proposal makes sense (`DeveloperError` where appropriate, true / false or other specific handling where appropriate, and removing the debug pragma string checks).

mzschwartz5: I do think the `isIntegerType` type functions should also throw a developer error if the input arg isn't in the mapping.

Same reasoning you gave above - this is probably unexpected behavior that we don't want to hide.

javagl: I'm usually in favor of 1. strict checks and 2. avoiding to "fail silently". 

(A simple error case that is ignored can easily lead to these bugs where "sometimes (with my secret data), something is not rendered" - these can be real time sinks).

In this area, some care has to be taken, because there has been that [`EXT_feature_metadata`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_feature_metadata) extension that is still supposed to be working in CesiumJS. It is _nearly_ equal to the current one, but there are some really suble differences. E.g. `type` could still be `STRING` or `ARRAY` back then. So maybe these cases should _not_ throw when being passed to `isIntegerType`? Or maybe the check should be done earlier or dedicatedly, meaning that the _call_ to `isIntegerType` should not be done...? Some of that may have to be decided on a case-by-case-basis with some tradeoffs.

mzschwartz5: Hmm interesting. I'm not really familiar with the old extension. Aside from `type` potentially being string or array, are there any other such differences? Or, put another way, is it sufficient to omit these exceptions from `isIntegerType` (and `isUnsigned...` etc)?

mzschwartz5: Or, perhaps, should we include those exceptions anyway - after all, the API is marked as experimental and subject to changing without the standard deprecation policy (for this very reason - so we can change things to how they _should_ be without bending over backwards).

javagl: There was bit of back and forth revolving around `type` and `componentType`, and the role that they played in these extensions. And it's hard to be 100% sure about the "right" handling for helper methods like `isIntegerType`. 

The (idealistically) strict solution would be that
`isIntegerType("STRING")`
throws, **but** it might very well be that this function (or things like `getMinimum` etc) is casually called ~"somewhere", _maybe_ as part of dealing with the legacy extension, and _anticipating_ that it does not throw and simply returns `undefined` for the given types. 

So it is not clear whether it could throw for `STRING` or `ARRAY` or `BOOLEAN` or `ENUM` (at least, not without finding and analyzing all call sites of that function, and/or checking the specs (and hoping that they cover everything relevant (and that they cover everything _properly_, and not contain glitches like that `MetadataComponentType.STRING` being `undefined`))). 

But I think that "completely invalid" strings should definitely throw - and that should happen _explicitly_, not only in debug pragmas. 

The number of data sets that use `EXT_feature_metadata` is probably close to 0, but ... it has to be handled. And it's not clear which implicit assumptions are baked into the code and the legacy handling. 

One approach **might** be (with a bit of pseudocode):
```
// As needed...
const integerComponentTypes = [ "UINT8" ... "INT64" ];
const floatingPointComponentTypes = [ "FLOAT32"... ];
const legacyComponentTypes = [ "BOOLEAN", "STRING, "ENUM"];  // Verify that...

MetadataComponentType.isIntegerType = function (type) {
  if (integerComponentTypes.includes(type)) return true;
  if (floatingPointComponentTypes.includes(type)) return false;
  if (legacyComponentTypes .includes(type)) {
    console.log("Meh... 😕");
    return false;
  }
  const up = new DeveloperError("What is "+type+"?");
  throw up;
}
```

Yes, this is pseudocode. The `TypeInfo` would be nice, maybe there is still a way to pull this through. Things like the `isVectorCompatible` in this structure or the signed/unsigned distinction could be done with
`const numericComponentTypes = [ "UINT8" ... "INT64", "FLOAT32", "FLOAT64"];`
`const signedIntegerComponentTypes = [ "INT8" ... "INT64" ];`
`const unsignedIntegerComponentTypes = [ "UINT8" ... "UINT64" ];`
but maybe(!), conversely, there could be `TypeInfo` entries for the `legacyComponentTypes` from that pseudocode...?

(I had to deal with similar things in the 3D Tiles Tools, but this does not have to handle the legacy extensions, and I [partially adapted](https://github.com/CesiumGS/3d-tiles-tools/blob/24bf3b8a90074b5095f51bfaafbb5f2dbf11cf44/src/metadata/metadata/MetadataComponentTypes.ts#L139) some code from CesiumJS, without larger refactorings...)

packages/engine/Source/Shaders/Builtin/Functions/unpackTexture.glsl
lilleyse: Maybe these should be called `czm_unpackUint` and the old ones should be called `czm_unpackUintWebgl1`?

mzschwartz5: Hmm... the old functions are named with the target type. So `unpackFloat` unpacks 4 bytes "packed as a little-endian unsigned normalized vec4" _to_ a float.

The new functions unpack to a `uint`, true, but only as an intermediate - for the bits to then be reinterpreted as some other type. So the name `unpackUint` still doesn't seem right to me.

packages/engine/Source/Scene/PropertyTextureProperty.js
lilleyse: > This is the meat of this PR. It's admittedly quite complex, and I'm open to suggestions on how to simplify and improve this bit (or maybe that's just how dynamically generated code is...)

I found it pretty clear, don't really see any obvious ways to make it simpler.

packages/engine/Source/Shaders/Builtin/Functions/unpackTexture.glsl
lilleyse: True...

I was also considering the word `reinterpret`, like `reinterpretTextureSampleAsUint`, but that doesn't quite fit either.

The current wording is probably good enough.

lilleyse:
> Or just document it as a breaking change, potentially? The classes surrounding metadata and property textures are all tagged as experimental and subject to change without the typical deprecation policy.

I think this is fine.

lilleyse:
@mzschwartz5 that's all from me, looks good. We can get this merged after CHANGES.md is updated.

 I'll take a look at https://github.com/CesiumGS/cesium/pull/13163 to see the enum / bool handling.

packages/engine/Specs/Scene/MetadataComponentTypeSpec.js
mzschwartz5: I think I prefer the approach with `MetadataComponentType.typeInfo`, rather than (effectively) a number of parallel arrays.

Also, I did some semi-thorough call-site analysis of these functions (`getMinimum`, `getMaximum`, etc). They're pretty minimal, and scoped relatively to numeric types. I think that's why the checks are in pragmas - because these are private helper functions, so any invalid use of them should only occur by developers.

Personally, I think it would be OK to use the approach you initially outline: 

```JS
/**
 * ...
 * @throws DeveloperError When the given type is not a valid numeric type [TODO maybe list them...]
 *
 * @private
 */
MetadataComponentType.getMinimum = function (type) {
  Check.typeOf.string("type", type); //---------------- TODO maybe omit this?
  //>>includeEnd('debug');
  const typeInfo = MetadataComponentType.typeInfo[type];
  if (!defined(typeInfo)) {
    throw new DeveloperError(`Type ${type} is not a numeric type`); // Proper message, with the "type"!
  }
  return typeInfo.minimumValue;
};
```

But I think the `throws` _should_ actually be inside a debug pragma, for the reason I described above.

PR #13140, opened 2026-01-15
Account for broken double-encoded sandcastle links

jjspace:
# Description

Add an attempt to load legacy ids that were double encoded by decoding once more after processing the `URLSearchParams`

It looks like the legacy sandcastle accidentally "accounted" for this by using the [demo name directly in the request](https://github.com/CesiumGS/cesium/blob/8a0087956455547bec8ca5dc14646920bf03687c/Apps/Sandcastle/CesiumSandcastle.js#L800) to load it. This meant it didn't matter if the request was sent using spaces ` ` or `%20`s. This approach doesn't work when we're doing direct string comparisons in the new Sandcastle.

This is a companion PR to https://github.com/CesiumGS/cesium/pull/13134 which actually fixes all the incorrect links in the JSDoc itself

For the record I also tried changing the `buildGallery` script to just include the encoded version of legacy ids but that added about 7kb to the list file size. That felt like it'd be worse in the long run than a single JS transformation that only happens some of the time.

## Issue number and link

Related to https://github.com/CesiumGS/cesium/issues/13122

## Testing plan

- run `npm start`
- Load these urls and make sure they load valid sandcastles
    - http://localhost:8080/Apps/Sandcastle2/index.html?src=Map%2520Pins.html
    - http://localhost:8080/Apps/Sandcastle2/index.html?src=Spheres%2520and%2520Ellipsoids.html
    - Or check them in CI like here https://ci-builds.cesium.com/cesium/sandcastle-double-encoding/Apps/Sandcastle2/index.html?src=Spheres%2520and%2520Ellipsoids.html

# Author checklist

### DIFF
--- a/packages/sandcastle/src/Gallery/loadFromUrl.ts
+++ b/packages/sandcastle/src/Gallery/loadFromUrl.ts
@@ -96,7 +96,14 @@ export function loadFromUrl(
 
   const legacyId = searchParams.get("src");
   if (legacyId) {
-    const item = selectItemByLegacyId(legacyId);
+    // there was long period when our doc generation double encoded sandcastle links that were already
+    // encoded causing space to be `%2520`. When this loads here the string comparison doesn't match `%20` with ` `
+    // attempt to check a second decode only when a match is not found
+    // see https://github.com/CesiumGS/cesium/issues/13122
+    const item =
+      selectItemByLegacyId(legacyId) ??
+      selectItemByLegacyId(decodeURI(legacyId));
+
     if (!item) {
       if (items.length > 0) {
         throw new Error(


javagl (approved):
This looks like a very reasonable solution for me: 

- It is a very small, very local change (a single line after all)
- It is implemented as a pure fallback 
  - So it does not involve any "nightmarish" _checks_ for double-encoded links, which was a concern at https://github.com/CesiumGS/cesium/issues/13122#issuecomment-3746434741). It just tries the other option when the first one fails. Sure, this is very pragmatic, but I think that's OK here
- It is commented with a summary that justifies it
- Most importantly: It seems to work well. I tested it on a local build, and the CI result, and it worked smoothly in both cases.

I think that it does make sense to include this. There are probably dozens of "copies/mirrors" of previous versions of the CesiumJS docs out there. A pretty important one, but just one example, is https://help.agi.com/STKWebVisualizationLibrary/GeoJsonDataSource.html which is currently "broken", and should be fixed after this is released.

I'd summarize this with "LGTM"...

jjspace:
Thanks @javagl! I think if you're happy with it this is probably good to merge

MohammadShujaullah:
Hi @jjspace and @javagl,I noticed my PR #13134 (fixing Sandcastle links in JSDoc) was incorporated into your solution in #13140. Glad my work contributed to the final fix!Thanks for the collaboration - looking forward to more contributions to Cesium!"This is brief, acknowledges the collaboration, and shows enthusiasm for future contributions.

jjspace:
@MohammadShujaullah this was an adjacent, related, change to address the problem from a different angle. It doesn't include the changes you made to the docs themselves. I think we will still want to finish up and include your work from #13134

Issue #1567, opened 2014-03-23
Bug when removing primitve with shared materials


nirko:
Hi
When removing a polyline that share a material with another polyline i get an exception.

Code in sandcastle:
require(['Cesium'], function(Cesium) {
    "use strict";

```
var met = Cesium.Material.fromType(Cesium.Material.PolylineOutlineType, {
            outlineWidth : 5.0
        });
var viewer = new Cesium.Viewer('cesiumContainer');
var primitives = viewer.scene.primitives;
var polylines = new Cesium.PolylineCollection();
var polyline;
var widePolyline;
var coloredPolyline;

function createPrimitives(scene, ellipsoid) {


    // A simple polyline with two points.
     polyline = polylines.add({
        positions : ellipsoid.cartographicArrayToCartesianArray([
            Cesium.Cartographic.fromDegrees(-120.0, 40.0),
            Cesium.Cartographic.fromDegrees(-110.0, 30.0)
        ]),
        material :met
    });
    Sandcastle.declare(polyline);   // For highlighting on mouseover in Sandcastle.

    // Apply a polyline outline material
     widePolyline = polylines.add({
        positions : ellipsoid.cartographicArrayToCartesianArray([
            Cesium.Cartographic.fromDegrees(-105.0, 40.0),
            Cesium.Cartographic.fromDegrees(-100.0, 38.0),
            Cesium.Cartographic.fromDegrees(-105.0, 35.0),
            Cesium.Cartographic.fromDegrees(-100.0, 32.0)
        ]),
        material : met,
        width : 10.0
    });
    Sandcastle.declare(widePolyline);   // For highlighting on mouseover in Sandcastle.


    primitives.add(polylines);

}


viewer.screenSpaceEventHandler.setInputAction(function(movement) {
    var pickedPrimitive = viewer.scene.pick(movement.endPosition);
    Sandcastle.highlight(pickedPrimitive);
}, Cesium.ScreenSpaceEventType.MOUSE_MOVE);

createPrimitives(viewer.scene, viewer.scene.globe.ellipsoid);
Sandcastle.addToolbarButton('Remove', function() {

    polylines.remove(polyline);


});

Sandcastle.finishedLoading();
```

});

OmarShehata:
[...]

PR #13141, opened 2026-01-15
PolylineCollection: Optimize .update() performance

### LINKED ISSUE #4738: Polyline performance
After creating lots of polylines, it seems a lot of time is spent each render in PolylineCollection.update -> createCommandLists, even though nothing changes. 

Easy to reproduce by running this gist in sandcastle: https://gist.github.com/YousefED/092aa2e48ca8ac4518b5b5c39e1c1682

Wait for everything to load, then open Chrome inspector, and run the profiler:

image

donmccurdy:
# Description

### Summary

Performance optimizations, reducing costs of the per-frame `PolylineCollection#update` method for large collections.

### Background

PolylineCollection sorts polylines into buckets according to _material type_, subdivides each bucket into chunks of 64K indices, and then batches draw commands for each contiguous series of polylines sharing the "identical materials". Identifying identical materials involves computing a material cache key by serializing the material to JSON, where each polyline has its own unique white material by default. For each draw command, a bounding sphere is computed. These computations happen on every frame, even if polyline positions and materials haven't changed. For large collections (tested here with 125K polylines) the overhead is substantial.

Decisions about batching strategies might be applicable to other collection types; it isn't obvious to me yet if other collections do batching similarly. For now I've tried to keep this essentially the same.

### Changes

1. ~~If no material is specified during polyline creation, a default material is now shared by the entire collection, rather than creating a new default material per-polyline.~~ (reverted)
2. Rather than comparing materials by JSON-serialized keys, ~~a unique GUID is computed per-material~~ materials are compared first by `===`, and then by deeper equality to the current batch material. This gives slightly lower overhead if there are many unique materials, and much lower overhead when materials are reused.
3. Bounding spheres for each bucket/material pair are computed once and then cached. Changes to polyline data, or to scene mode, will invalidate this cache and recompute bounding spheres.

Changes (1) and (2) are opinionated and could be reverted ~~#13142 shows another way to optimize without them _if_ the caller is careful to reuse materials as much as possible.~~ My concern is that CesiumJS might be "promising too much" here; deduplicating 100K+ materials is expensive no matter how we do it — I think it might be better to push toward usage patterns where (a) materials are allocated more conservatively, (b) Material.ColorType materials with varying colors can still be batched, and/or (c) varying colors would not require distinct materials. But for now this is a quick improvement.

Change (3) means that bounding volumes are more conservative, computed once per material in each collection, not once per contiguous series of polylines using each material in each collection. As an improvement on #4738 I think this is fine, but the batching and caching plan could probably be improved further in the future.

### Results

I've adapted the code from #4738 by increasing the number of instances from 1,000 to 100,000. ~~Depending on whether we keep this PR as-is, or include #13142, it may be necessary to also reuse materials for the full benefit.~~ For best results, materials should be reused when possible.

```javascript

const viewer = new Cesium.Viewer('cesiumContainer');

const lines = viewer.scene.primitives.add(new Cesium.PolylineCollection());

const material = Cesium.Material.fromType(Cesium.Material.ColorType, {
  glowPower: 0.05,
  color: Cesium.Color.RED
});

for(let i = 0; i < 100_000; i++) {
    const positions = [-75 + i / 100, 35, -60, 40];
    lines.add({
        positions: Cesium.Cartesian3.fromDegreesArray(positions),
        material,
        width: 1.0
    });
}
```

| production | this PR |
|----|----|
| 380-400ms / frame | 4–8ms / frame |

Tested on macOS + M4. If we instead assign a unique material to each polyline, then this PR offers a small improvement over production at about 300ms/frame, but remains highly problematic. For strong performance guarantees, see https://github.com/CesiumGS/cesium/issues/13156. 
## Issue number and link

- Fixes #4738

## Testing plan

TODO

# Author checklist

#### PR Dependency Tree

* **PR #13141** 👈

### DIFF
--- a/packages/engine/Source/Scene/PolylineCollection.js
+++ b/packages/engine/Source/Scene/PolylineCollection.js
@@ -26,16 +26,15 @@ import Pass from "../Renderer/Pass.js";
 import RenderState from "../Renderer/RenderState.js";
 import ShaderProgram from "../Renderer/ShaderProgram.js";
 import ShaderSource from "../Renderer/ShaderSource.js";
-import Texture from "../Renderer/Texture.js";
 import VertexArray from "../Renderer/VertexArray.js";
 import PolylineCommon from "../Shaders/PolylineCommon.js";
 import PolylineFS from "../Shaders/PolylineFS.js";
 import PolylineVS from "../Shaders/PolylineVS.js";
 import BatchTable from "./BatchTable.js";
 import BlendingState from "./BlendingState.js";
-import Material from "./Material.js";
 import Polyline from "./Polyline.js";
 import SceneMode from "./SceneMode.js";
+import Material from "./Material.js";
 
 const SHOW_INDEX = Polyline.SHOW_INDEX;
 const WIDTH_INDEX = Polyline.WIDTH_INDEX;
@@ -160,6 +159,11 @@ function PolylineCollection(options) {
   this._propertiesChanged = new Uint32Array(NUMBER_OF_PROPERTIES);
   this._polylines = [];
   this._polylineBuckets = {};
+
+  /** @type {Material} */
+  this._defaultMaterial = Material.fromType(Material.ColorType, {
+    color: new Color(1.0, 1.0, 1.0, 1.0),
+  });
 
   // The buffer usage is determined based on the usage of the attribute over time.
   this._positionBufferUsage = {
@@ -459,6 +463,7 @@ PolylineCollection.prototype.update = function (frameState) {
     } else {
       const length = polylinesToUpdate.length;
       const polylineBuckets = this._polylineBuckets;
+      const polylineBucketsUpdated = new Set();
       for (let ii = 0; ii < length; ++ii) {
         polyline = polylinesToUpdate[ii];
         properties = polyline._propertiesChanged;
@@ -533,7 +538,17 @@ PolylineCollection.prototype.update = function (frameState) {
           }
         }
 
+        // Bucket may be undefined for empty Polyline.
+        if (bucket) {
+          polylineBucketsUpdated.add(bucket);
+        }
+
         polyline._clean();
+      }
+
+      // Invalidate bounding volume cache for affected buckets.
+      for (const bucket of polylineBucketsUpdated) {
+        bucket.invalidateBoundingVolumes();
       }
     }
     polylinesToUpdate.length = 0;
@@ -586,9 +601,6 @@ PolylineCollection.prototype.update = function (frameState) {
   }
 };
 
-const boundingSphereScratch = new BoundingSphere();
-const boundingSphereScratch2 = new BoundingSphere();
-
 function createCommandLists(
   polylineCollection,
   frameState,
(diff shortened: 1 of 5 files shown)


packages/engine/Source/Scene/PolylineCollection.js
```diff
@@ -1263,6 +1214,13 @@ function PolylineBucket(material, mode, modelMatrix) {
   this.shaderProgram = undefined;
   this.mode = mode;
   this.modelMatrix = modelMatrix;
+  /**
+   * Mapping from material IDs to per-material bounding volumes.
+   * @type {Object<string, BoundingSphere>}
+   */
+  this.boundingVolumes = {};
```
plainheart: Hi - I'm thrilled to see that the Cesium team has been gradually refining and improving the performance of some essential basic APIs!

From what we've actually worked on before (#12896, #12916), would it make sense to use a Map here?

javagl:
I'm in no way familiar enough with the code to provide really valuable feedback here. (That "buckets" thing sounds like what is called "batch" in other parts of the code...?). However, regarding that point

> If no material is specified during polyline creation, a default material is now shared by the entire collection, rather than creating a new default material per-polyline.

Maybe the answer is obvious, but .. when a single material instance is used, wouldn't that mean that something like a 
`polylineCollection.get(42).material.color = newColor;`
would affect all instances (i.e. the `_defaultMaterial`)?

packages/engine/Source/Scene/PolylineCollection.js
donmccurdy: Very interesting about the object vs. Map differences, thanks! It's not obvious to me whether we're using the lookup table in a way that would hit these object slow paths (long keys, frequent key deletion, etc.) but I'll try testing performance with Map here — if reviewers are OK with the per-material bounding volumes. If not then the mapping will need to be replaced anyway.

donmccurdy:
In the current draft of this PR, yes, it would affect all polyline instances on the collection if the instances were not initialized with a material explicitly. For a more narrowly-scoped fix, the patch in #13142 would avoid that limitation. The overhead is higher, but still a good improvement, while avoiding (or at least delaying...) a more obvious breaking change.

I should perhaps also note that destroying a polyline destroys its material unconditionally, today, so it might be too soon to recommend reusing materials across polylines... if there are no objections I'll merge #13142 into this PR.

packages/engine/Source/Scene/PolylineCollection.js
donmccurdy: Done! I don't see much performance difference in this case, but it's nice to be able to skip adding a new ._id property on materials, so I think it's a good change. Thanks!

donmccurdy:
Ok, I've reverted that part of the change and merged #13142 into this PR, keeping this a simpler/safer optimization without breaking changes. Materials remain unique per polyline by default.

javagl:
I didn't want to distract, given that 1. this a ... "complex"... area of the code, 2. this is still a 'draft', and 3. I have not thoroughly read and understood the changes. The question was largely due to the apparent change of using a _single_ material instance, and the possible side-effects. 

> I should perhaps also note that destroying a polyline destroys its material unconditionally, 

I had seen that one `destroy` call was removed, and assume that the `_defaultMaterial` should live forever, but beyond that, in some cases, some aspects of some lifecycles may not be perfectly clear. I'd _like_ to be more specific here, but for now, can only point to a (2014) issue at https://github.com/CesiumGS/cesium/issues/1567 (maybe as the opposite of a warning: when it crashes, it's not necessarily your fault 🙂 )

donmccurdy:
Closing. Review required; no other blockers.

PR #13151, opened 2026-01-21
Fix memory leak in GeoJsonDataSource (#9058)

### LINKED ISSUE #9058: can't free memory used by loading GeoJsonDataSource, even when not added to viewer
When calling GeoJsonDataSource.load multiple times the memory usage continues to rise and never goes back down, even when the datasource is never added to the viewer, no entities are created/shown and there is no way to 'unload' the datasource.

My use case is increasing polygon resolution as the user zooms in, and decreasing when not visible on the screen or zoomed out. Once the higher resolution is loaded the increased memory usage never goes down again. 

Sandcastle example:
I can't really provide a sandcastle example because this requires loading external geojson files from somewhere.

Browser:
Chrome Version 84.0.4147.89
Operating System:
Windows 10.0.18362 Build 18362

MohammadShujaullah:
Fixes #9058

##Description:
This PR addresses the memory leak issue in [GeoJsonDataSource](file:///C:/Users/mdshu/OneDrive/Desktop/Cesium_Bug-Problem_fix/cesium/packages/engine/Source/DataSources/GeoJsonDataSource.js#L29-L232) where loaded data was not being properly garbage collected even after removing the datasource and its entities, and not adding it to the viewer.

##Problem:
Calling [GeoJsonDataSource.load](file:///C:/Users/mdshu/OneDrive/Desktop/Cesium_Bug-Problem_fix/cesium/packages/engine/Source/DataSources/GeoJsonDataSource.js#L145-L230) multiple times caused continuous memory usage growth
Memory was not freed even when datasources were never added to the viewer
Particularly problematic for applications dynamically loading large GeoJSON files (e.g., high-resolution polygons on zoom)
Affected mobile performance, especially on iOS

##Solution:
Added a destroy() method to properly clean up resources:
Clears the promises array to cancel any pending operations
Disposes of the pin builder to free up image resources
Clears the entity collection
Cleans up resource credits
Uses proper destroyObject() pattern
Added a cancelLoading() method to allow manual cancellation of pending operations
Enhanced error handling in the loading process to ensure promises are cleared in both success and error scenarios
Added proper import for destroyObject

##Testing:
The changes have been tested to ensure that memory is properly released when the data source is destroyed. The fix maintains backward compatibility while resolving the memory leak issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,13 +4,20 @@
 
 ### @cesium/engine
 
-#### Fixes :wrench:
+#### Breaking Changes :mega:
 
-- Fixed precision of point cloud attributes when accessed in a custom fragment shader. [#13170](https://github.com/CesiumGS/cesium/pull/13170)
+- Cartesian2, Cartesian3, and Cartesian4 are now [ES6 Classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes). This change should have no impact on most users, but note that using `new` on a static factory method, like `new Cartesian3.fromArray(...)`, will now throw an error. Omit `new` unless you are invoking a constructor directly, for these and all other factory methods, as more classes will be migrated to ES6 Classes soon. [#8359](https://github.com/CesiumGS/cesium/issues/8359)
 
 #### Additions :tada:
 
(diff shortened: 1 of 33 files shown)


MohammadShujaullah:
Hey @jjspace , hope you're having a good week! When you get some free time, would you mind reviewing this PR? No pressure if you're swamped!

PR #13153, opened 2026-01-22
Add panoramic imagery feature support

lukemckinstry:
# Description

Adds support for viewing panoramic imagery in a scene, with `EquirectangularPanorama` and `CubeMapPanorama` classes. `GoogleStreetViewCubeMapPanoramaProvider` supports steaming Google Maps images from the Street View Static API. 

Usage example:
```
const position = Cesium.Cartesian3.fromDegrees(
  -75.1699, // longitude
  39.9522, // latitude
  100.0, // height in meters
);

const heading = Cesium.Math.toRadians(45.0); // rotation about up axis
const tilt = Cesium.Math.toRadians(-30.0); // pitch (negative looks down)
const roll = Cesium.Math.toRadians(10.0); // roll about forward axis

const hpr = new Cesium.HeadingPitchRoll(heading, tilt, roll);

const modelMatrix = Cesium.Transforms.headingPitchRollToFixedFrame(
  position,
  hpr,
  Cesium.Ellipsoid.WGS84,
  Cesium.Transforms.eastNorthUpToFixedFrame,
);

scene.primitives.add(
  new Cesium.EquirectangularPanorama({
    transform: modelMatrix,
    image: "path/to/image",
  }),
);

// Lock the camera to the center of the panorama
scene.camera.lookAt(
  modelMatrix,
  new Cesium.HeadingPitchRange(
    0, // heading
    0, // pitch
    10, // small offset to allow rotation
  ),
);

// Allow camera to rotate, tilt, and zoom
const controller = scene.screenSpaceCameraController;
controller.enableRotate = true;
controller.enableZoom = false;
controller.enableTilt = true;
controller.enableTranslate = false;

```

## Testing plan
 
1. ~~Note: where noted below, sandcastles require pasting a valid key for Google Street View Static API~~ Update: No key needed for sandcastles, stream direct from Cesium ion

~~To create a key or see your existing keys: Google Cloud Console  (select or create a valid project) --> APIs and Services --> Credentials --> Create credentials --> Api Key --> Under API Restrictions / Selected APIs, select the desired APIs listed above. Note you can make one key with both permissions.~~

2. View the following sandcastles
 - Panorama - basic examples with equirectangular and cube map panoramas [sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html?id=panorama)
 - GoogleStreetViewCubeMapPanoramaProvider  - See Google StreetView Panorama [sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html?id=google-streetview-panorama) 

# Author checklist


javagl:
Do we really (I mean _really_ **really**) want four ~30 MB (!) PNG files in the sample data? 

Unless this is some sort of "stress test" that is absolutely necessary, I'd strongly suggest to remove them (and squash the commits, so that they are not bloating the Git history by a staggering >100MB).

I don't have much context here. But a few weeks ago, I created some data for a "unit cube with details", and could imagine that something like this (maybe sliced into 6 files) could be useful here:

CubeMap 0001

Apps/SampleData/cathedral/bottom_left.png
jjhembd, line 1: I agree with @javagl that these files are large. Can we simply convert them from PNG to JPEG, and perhaps adjust the compression to get them under 1 MB each?

Or should they be moved to ion?

lukemckinstry: I plan to remove these files once we have settled on which demos and sandcastles to publish. I left them in for testing and evaluation purposes.

lukemckinstry:
Hi @jjhembd, I added the `GoogleStreetViewCubeMapPanoramaProvider` to the PR. As discussed this will almost definitely be the only Google provider we push initially, not `GoogleStreetViewEquirectangularPanoramaProvider`. I do not think we need to spend any more time & effort reviewing the provider we will not publish.

packages/sandcastle/gallery/panorama-equirectangular/main.js
```diff
@@ -0,0 +1,44 @@
...
+// Convert to Cartesian3 position
+const position = Cesium.Cartesian3.fromDegrees(longitude, latitude, height);
+
+// Create a transform matrix to place the sphere at the correct location
+const transform = Cesium.Transforms.eastNorthUpToFixedFrame(position);
+
+const image =
+  "https://upload.wikimedia.org/wikipedia/commons/0/08/Laon_Cathedral_Interior_360x180%2C_Picardy%2C_France_-_Diliff.jpg";
+
+const panorama = new Cesium.EquirectangularPanorama({
```
jjhembd: I don't see any image credit in the Sandcastle. I noticed the `EquirectangularPanorama` constructor accepts a credit in the options object--do we just need to add it here?

It looks like [the image is under a CC Attribution license](https://commons.wikimedia.org/wiki/File:Laon_Cathedral_Interior_360x180,_Picardy,_France_-_Diliff.jpg)

packages/sandcastle/gallery/panorama-equirectangular/main.js
```diff
@@ -0,0 +1,44 @@
...
+viewer.scene.camera.lookAtTransform(
+  transform,
+  new Cesium.HeadingPitchRange(
+    0, // heading
+    0, // pitch
+    10, // small offset to allow rotation
+  ),
+);
+
+// Allow camera to rotate, tilt, and zoom
```
jjhembd: I wanted to be able to zoom... but `enableZoom` is actually false. Do we need to update the comment?

When I enable zooming, it works well with small changes, but the camera easily zooms outside the sphere of the image.

Have we considered remapping zoom inputs to a narrowing of the field of view? This could give us a zoom-like UI without any problematic changes in the actual camera position.

packages/sandcastle/gallery/panorama-cubemap/main.js
```diff
@@ -0,0 +1,124 @@
...
+  viewer.scene.primitives.add(cubeMapPanorama);
+};
+
+const locations = {
+  Aukland: [-36.8509, 174.7645],
+  Switzerland: [46.414382, 10.013988],
+};
+cubeMapFromFiles();
+
+Sandcastle.addToolbarButton("Google Maps Aukland", function () {
```
jjhembd: If I try to select the Google Maps options with no API key set, the Sandcastle crashes with the following error message:

```
[object Object]

undefined
```

Forgetting to set a key, or using the wrong key, is a mistake that many users will make. Can we make the error message a little more helpful?

packages/sandcastle/gallery/panorama-cubemap/main.js
```diff
@@ -0,0 +1,124 @@
+import * as Cesium from "cesium";
+import Sandcastle from "Sandcastle";
+
+const viewer = new Cesium.Viewer("cesiumContainer");
+
+viewer.scene.terrainProvider = false;
+
+const googleStreetViewStaticApiKey = "key for Google Street View Static API";
```
jjhembd: Will we have a working default key for the public Sandcastle?

packages/sandcastle/gallery/panorama-cubemap/main.js
```diff
@@ -0,0 +1,124 @@
...
+      positiveX,
+      negativeX,
+      positiveY,
+      negativeY,
+      positiveZ,
+      negativeZ,
+    },
+    transform,
+  });
+  viewer.scene.camera.setView({
```
jjhembd: This Sandcastle reacts to pan inputs in the opposite way as the equirectangular Sandcastle. Personally I like this Sandcastle better, but I could see the value in both behaviors.

As it stands, these two Sandcastles behaving differently may be confusing to users. Some ideas to make it less confusing:
- Pick one camera behavior and make all the sandcastles behave the same. If people want a different behavior, they can look at Camera docs or ask on the forum.
- Add a camera mode selector. This will call attention to the different options for camera behavior, and let users try both ways.
- Combine the cubemap and equirectangular Sandcastles, and simply call out the different input formats in the drop-down descriptions. This option could work with or without the camera mode selector. I can see the value of separate sandcastles for end-to-end testing purposes, but it might be unnecessary noise for the public gallery.

packages/engine/Source/Scene/SkyBox.js
```diff
@@ -1,25 +1,7 @@
...
-import DrawCommand from "../Renderer/DrawCommand.js";
-import loadCubeMap from "../Renderer/loadCubeMap.js";
-import RenderState from "../Renderer/RenderState.js";
-import ShaderProgram from "../Renderer/ShaderProgram.js";
-import ShaderSource from "../Renderer/ShaderSource.js";
-import VertexArray from "../Renderer/VertexArray.js";
-import SkyBoxFS from "../Shaders/SkyBoxFS.js";
-import SkyBoxVS from "../Shaders/SkyBoxVS.js";
-import BlendingState from "./BlendingState.js";
+import CubeMapPanorama from "./CubeMapPanorama.js";
```
jjhembd, line 2: this is nice 🥇

packages/engine/Source/Scene/CubeMapPanorama.js
```diff
@@ -0,0 +1,371 @@
...
+import SkyBoxFS from "../Shaders/SkyBoxFS.js";
+import SkyBoxVS from "../Shaders/SkyBoxVS.js";
+import CubeMapPanoramaVS from "../Shaders/CubeMapPanoramaVS.js";
+import BlendingState from "./BlendingState.js";
+import SceneMode from "./SceneMode.js";
+import Pass from "../Renderer/Pass.js";
+import Credit from "../Core/Credit.js";
+
+/**
+ * A sky box around the scene to draw stars.  The sky box is defined using the True Equator Mean Equinox (TEME) axes.
```
jjhembd: Looks like we have generalized this code to implement more things than just star maps. Can we update the comments in the file accordingly? This would apply to every comment mentioning "sky box" in this file.

packages/engine/Source/Scene/EquirectangularPanorama.js
```diff
@@ -0,0 +1,218 @@
...
+ * (for example, derived from a position and heading-pitch-roll).
+ * @property {string|HTMLImageElement|HTMLCanvasElement|ImageBitmap} options.image A URL to an image resource, or a preloaded image object.
+ * @property {number} [options.radius=100000.0] The radius of the panorama in meters.
+ * @property {number} [options.repeatHorizontal=1.0] The number of times to repeat the texture horizontally.
+ * @property {number} [options.repeatVertical=1.0] The number of times to repeat the texture vertically.
+ * @property {Credit|string} [options.credit] A credit for the panorama, which is displayed on the canvas.
+ */
+
+/**
+ * Provides equirectangular imagery to be displayed on the surface of an ellipsoid.
```
jjhembd: Is there a use case for non-spherical ellipsoids here? I notice the `._radius` property is a single number, which doesn't sound like it allows for anything other than spheres.

packages/engine/Source/Scene/PanoramaProvider.js
```diff
@@ -0,0 +1,28 @@
+import DeveloperError from "../Core/DeveloperError.js";
+
+/**
+ *
+ * See the documentation for each PanoramaProvider Provider class for more information about how they return PanoramaProvider.
```
jjhembd: Stylistic nitpick: Can this comment be combined with the main comment for the dummy interface constructor? This would be more consistent with the pattern we use in the `ImageryProvider` and `TerrainProvider` interfaces.

packages/engine/Source/Scene/Panorama.js
```diff
@@ -0,0 +1,85 @@
+import DeveloperError from "../Core/DeveloperError.js";
+
+/**
+ *
+ * See the documentation for each Panorama class for more information about how they return images.
```
jjhembd: Stylistic nitpick: Can this comment be combined with the main comment for the dummy interface constructor?

packages/engine/Source/Scene/PanoramaProvider.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Provides imagery to be displayed on the surface of an ellipsoid.  This type describes an
+ * interface and is not intended to be instantiated directly.
+ *
+ * @alias PanoramaProvider
+ * @constructor
+ * @abstract
+ *
+ * @see EquirectangularPanorama
+ * @see SkyBoxPanorama
```
jjhembd:
```suggestion
 * @see CubeMapPanorama
```

packages/engine/Source/Scene/PanoramaProvider.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @demo {@link https://sandcastle.cesium.com/index.html?src=Imagery%20Layers.html|Cesium Sandcastle Imagery Layers Demo}
```
jjhembd: I think these demo links need to be updated to the Panorama sandcastles?

packages/engine/Source/Scene/PanoramaProvider.js
```diff
@@ -0,0 +1,28 @@
...
+ * @demo {@link https://sandcastle.cesium.com/index.html?src=Imagery%20Layers%20Manipulation.html|Cesium Sandcastle Imagery Manipulation Demo}
+ */
+function PanoramaProvider() {
+  DeveloperError.throwInstantiationError();
+}
+
+Object.defineProperties(PanoramaProvider.prototype, {});
+
```
jjhembd, line 22: Can we fill out the interface a little more? I realize we want the interface to stay general, and we don't necessarily know how other providers will behave yet. But I assume the `.fromUrl` and `loadPanorama` methods will be required in every provider?

packages/engine/Specs/Scene/GoogleStreetViewCubeMapPanoramaProviderSpec.js
```diff
@@ -0,0 +1,228 @@
...
+  let provider;
+
+  beforeEach(function () {
+    provider = new GoogleStreetViewCubeMapPanoramaProvider({
+      apiKey: "test-key",
+      url: "https://example.com/",
+      tileSize: 512,
+    });
+  });
+
```
jjhembd, line 20: Since this is an implementation of an interface, perhaps we should add a spec to confirm that it "conforms to the PanoramaProvider interface".

See `CesiumTerrainProviderSpec` for a comparable example.

packages/engine/Specs/Scene/CubeMapPanoramaSpec.js
```diff
@@ -0,0 +1,124 @@
...
+        render: true,
+      },
+      context: context,
+      panoramaCommandList: [],
+      creditDisplay: {
+        addCreditToNextFrame: jasmine.createSpy("addCreditToNextFrame"),
+      },
+    };
+  });
+
```
jjhembd, line 35: Can we add a spec to check if this conforms to the `Panorama` interface?

packages/engine/Specs/Scene/EquirectangularPanoramaSpec.js
```diff
@@ -0,0 +1,73 @@
+import EquirectangularPanorama from "../../Source/Scene/EquirectangularPanorama.js";
+import Credit from "../../Source/Core/Credit.js";
+import Matrix4 from "../../Source/Core/Matrix4.js";
+import Cartesian2 from "../../Source/Core/Cartesian2.js";
+
+describe("Scene/EquirectangularPanorama", function () {
```
jjhembd, line 7: Can we add a spec to check if this conforms to the `Panorama` interface?

jjhembd (changes requested):
Hi @lukemckinstry, it's exciting to see this working!

I looked through the interfaces, tests, and Sandcastles, and added some preliminary comments. I still need to look a little more closely at the draw command stuff.

packages/sandcastle/gallery/panorama-cubemap/main.js
lukemckinstry: I am thinking to set this in the sandcastle code, rather than the core classes

packages/sandcastle/gallery/panorama-cubemap/main.js
lukemckinstry: TBD

packages/sandcastle/gallery/panorama-cubemap/main.js
lukemckinstry: Fixed the equirectangular sandcastle example so they all use the same controls now. 

Still mulling how to consolidate the sandcastles and looking for better sample cube map data, but I agree with your suggestions.

lukemckinstry:
thanks @jjhembd for this round of feedback. I addressed all your suggestions except where noted above.

packages/engine/Source/Scene/PanoramaProvider.js
```diff
@@ -0,0 +1,43 @@
...
+  DeveloperError.throwInstantiationError();
+};
+
+/**
+ * Returns a panorama primitive.
+ *
+ * @param {object} options Input options to create the panorama primitive.
+ * @returns {Panorama} The panorama primitive for displaying panoramas in a scene.
+ */
+PanoramaProvider.loadPanorama = function (options) {
```
jjhembd:
```suggestion
PanoramaProvider.prototype.loadPanorama = function (options) {
```

packages/sandcastle/gallery/panorama/main.js
```diff
@@ -0,0 +1,187 @@
...
+    const camera = viewer.camera;
+    const frustum = camera.frustum;
+
+    let fov = frustum.fov;
+
+    // Wheel direction
+    const delta = movement;
+
+    if (delta > 0) {
+      fov *= 1.0 + zoomSpeed; // zoom out
```
jjhembd: Thanks for implementing this! It really makes the Sandcastle feel more complete.
On my machine these zoom behaviors seem to be switched. I expected rolling the wheel forward to zoom in, but instead it zooms out. Can you confirm on your machine?

packages/sandcastle/gallery/google-streetview-panorama/main.js
```diff
@@ -0,0 +1,424 @@
...
+    const camera = viewer.camera;
+    const frustum = camera.frustum;
+
+    let fov = frustum.fov;
+
+    // Wheel direction
+    const delta = movement;
+
+    if (delta > 0) {
+      fov *= 1.0 + zoomSpeed; // zoom out
```
jjhembd, line 152: Same question here about the zoom directions: they appear switched to me.

packages/sandcastle/gallery/google-streetview-panorama/main.js
```diff
@@ -0,0 +1,424 @@
...
+  terrainProvider: await Cesium.CesiumTerrainProvider.fromIonAssetId(1),
+});
+viewer.geocoder.viewModel.keepExpanded = true;
+
+const satelliteWithLabelsOverlay = viewer.imageryLayers.addImageryProvider(
+  await Cesium.IonImageryProvider.fromAssetId(3830183),
+);
+
+const tileset = await Cesium.createGooglePhotorealistic3DTileset({
+  // Only the Google Geocoder can be used with Google Photorealistic 3D Tiles.  Set the `geocode` property of the viewer constructor options to IonGeocodeProviderType.GOOGLE.
```
jjhembd, line 37: Maybe out of scope for this PR, but if Google P3DT has restrictions on which geocoder we can use, shouldn't that be implemented by default within `createGooglePhotorealistic3DTileset`, rather than depending on the user to set a flag?

packages/sandcastle/gallery/google-streetview-panorama/index.html
```diff
@@ -0,0 +1,10 @@
+<style>
+  @import url(../templates/bucket.css);
+</style>
+<div id="cesiumContainer" class="fullSize"></div>
+<div id="loadingOverlay"><h1>Loading...</h1></div>
+<div id="toolbar"></div>
+<div
+  id="cesiumContainer"
+  style="width: 100%; height: 100%; position: relative"
```
jjhembd: Should the style be implemented via `class="fullSize"` as in the other two sandcastles?

packages/sandcastle/gallery/google-streetview-panorama/sandcastle.yaml
```diff
@@ -0,0 +1,4 @@
+title: Google StreetView Panorama
+description: Display a panorama from Google StreetView
+labels:
+  - Imagery
```
jjhembd: Should this one have the `Showcases` label?

packages/sandcastle/gallery/google-streetview-panorama-navigation/main.js
```diff
@@ -0,0 +1,174 @@
...
+  );
+  Object.keys(panoIdMetadata.links).forEach((linkId) => {
+    const link = panoIdMetadata.links[linkId];
+    const destPoint = destinationPoint(startCartographic, link.heading, 15, 0);
+    // Add a point at the destination
+    viewer.entities.add({
+      position: destPoint,
+      point: {
+        pixelSize: 30,
+        color: Cesium.Color.RED,
```
jjhembd: This Sandcastle is fun. The alignment with Google P3DT is impressive.
One suggestion here: it's not immediately obvious what the red spots are for. Could we replace them with an arrow icon, or something else indicating that it is a navigation option?

packages/sandcastle/gallery/google-streetview-panorama/main.js
```diff
@@ -0,0 +1,424 @@
...
+  returnToMapButton = createButton("Return to map", async function () {
+    viewer.scene.terrainProvider =
+      await Cesium.CesiumTerrainProvider.fromIonAssetId(1);
+    viewer.scene.skyBox = Cesium.SkyBox.createEarthSkyBox();
+    returnToMap();
+    enablePicking();
+  });
+}
+
+function setMapViewToolBar() {
```
jjhembd, line 267: After I select "Return to map" from within a panorama view, I see a duplicated drop-down for the panorama type:

Image

packages/sandcastle/gallery/google-streetview-panorama/main.js
```diff
@@ -0,0 +1,424 @@
...
+);
+
+const tileset = await Cesium.createGooglePhotorealistic3DTileset({
+  // Only the Google Geocoder can be used with Google Photorealistic 3D Tiles.  Set the `geocode` property of the viewer constructor options to IonGeocodeProviderType.GOOGLE.
+  onlyUsingWithGoogleGeocoder: true,
+});
+tileset.show = false;
+viewer.scene.primitives.add(tileset);
+
+const provider =
```
jjhembd: There seems to be some conflict or overlap between the credits for the different datasets being used. When I'm in a panorama with Photorealistic 3DTiles turned on, I see a duplicated "Google Maps" logo. Clicking on the "Data attribution" link shows another mention of Google, along with some others.

Image

If I turn off the Photorealistic 3DTiles, the duplicate "Google Maps" logo disappears. The "Data Attribution" link also disappears.

Image

Disclaimer: I don't know what attribution is correct here, I'm just pointing out what I see.

packages/sandcastle/gallery/google-streetview-panorama-navigation/main.js
```diff
@@ -0,0 +1,174 @@
...
+  });
+  viewer.scene.primitives.add(tileset);
+} catch (error) {
+  console.log(`Error loading Photorealistic 3D Tiles tileset.
+          ${error}`);
+}
+
+const googleMapTilesApiKey = "Google Map Tiles API Key";
+
+const provider =
```
jjhembd: I also see some attribution duplication in this Sandcastle. The "Google Maps" logo is duplicated. 

Image

The "Data attribution" link includes things like Bing, Landsat, Gebco -- I'm pretty sure none of those are actually visible from within a panorama.

Image

Turning off Photorealistic Tiles fixes the duplicate logo:

Image

The "Data attribution" link still includes Bing.

packages/sandcastle/gallery/google-streetview-panorama-navigation/main.js
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Allow camera to rotate, tilt, and zoom
+const controller = viewer.scene.screenSpaceCameraController;
+controller.enableRotate = true;
+controller.enableZoom = false;
+controller.enableTilt = true;
+controller.enableTranslate = false;
+
+Sandcastle.addToolbarButton("Toggle Photorealistic Tiles", function () {
```
jjhembd:
```suggestion
Sandcastle.addToolbarButton("Toggle Photorealistic 3D Tiles", function () {
```
Or do we need "Google" in there too? I'll defer to @LisaBosCesium on proper naming & attribution for the different services.

TBC-GIS:
Would it be possible to turn off directional lighting from the sun, I think it's making the street view images darker. I think the scene is lit unevenly 

Sandcastle
image

Google Street View/Google Maps
image

jo-chemla:
I was AFK for some time, but if I remember correctly there was a mention of repositories of open-data 360 spheres/cubemaps that could be useful for testing or demo-ing this PR (in addition to commercial APIs like gmaps streetview), just wanted to give a few examples, that might be nice for a sandcastle, and remove the need to store test data within the repo - but would involve client-side fetching of geodata. 

Geo that also have a good api to retrieve spheres matching a given footprint bounds, with geolocation etc: 
 - [Mapillary](https://www.mapillary.com/app)
 - [Panoramax](https://api.panoramax.xyz)

Non geo, exteriors and interiors: 
 - [Wikimedia 360](https://commons.wikimedia.org/wiki/Category:360%C2%B0_panoramas)
 - [PolyHaven HDRIs](https://polyhaven.com/hdris)

packages/sandcastle/gallery/google-streetview-panorama/main.js
lukemckinstry: I suspect this is caused by this other issue you caught https://github.com/CesiumGS/cesium/pull/13153#discussion_r2835619052 
Please re-check now that I fixed it. This duplication was not happening for me, but Travis also noticed it.

packages/sandcastle/gallery/google-streetview-panorama-navigation/main.js
lukemckinstry: Good suggestion. We will actually hold on this sandcastle for the initial release and save it to showcase future work. I removed it from this PR and stored it in a different branch.

packages/sandcastle/gallery/google-streetview-panorama-navigation/main.js
lukemckinstry: I changed it to Google Photorealistic 3D Tiles to cover our bases

packages/sandcastle/gallery/google-streetview-panorama/main.js
lukemckinstry: For context I needed to use a bunch of different datasets for this sandcastle (lots of Google!)
 - Google Maps 2D Satellite with Labels
 - Google Maps 2D layerStreetview overlay
 - Google Photorealistric 3D Tiles
 - and the StreetView Imagery 
 
So this may explain the crowded attribution window. 

But there is also a bug with duplicate credits being shown that I will fix.

lukemckinstry: I fixed part of the duplicate credit issue by specifying not to add the credit for the skybox.

But I still need to figure out how we avoid duplicating logo credits when we have multiple Google assets in the viewer. It seems like we might downgrade duplicate logos to text entries, but I am not sure.

packages/sandcastle/gallery/google-streetview-panorama/main.js
lukemckinstry: I agree out of scope but worth looking into

lukemckinstry:
> I may have missed something in the thread, but can you confirm if the Sandcastles on sandcastle.cesium.com will have API keys provided? What about when people build and run locally? I'm still getting a non-descriptive error when I run locally without a key.

 - This was an open question but we have a decision now: The sandcastles on sandcastle.cesium.com will have access to Google Street View images via ion for the initial release using a `temporary` endpoint (ie. not a typical asset, since we have not settled on the design of a `Panorama` ion asset type yet). I will need to switch the url for GoogleStreetViewCubeMapPanoramaProvider to the ion endpoint and remove the prompt for users to add a key when the endpoint is available shortly.

lukemckinstry:
> repositories of open-data 360 spheres/cubemaps that could be useful for testing or demo-ing this PR (in addition to commercial APIs like gmaps streetview), just wanted to give a few examples, that might be nice for a sandcastle, and remove the need to store test data within the repo - but would involve client-side fetching of geodata.
> 
> Geo that also have a good api to retrieve spheres matching a given footprint bounds, with geolocation etc:
> 
> * [Mapillary](https://www.mapillary.com/app)
> * [Panoramax](https://api.panoramax.xyz)
> 
> Non geo, exteriors and interiors:
> 
> * [Wikimedia 360](https://commons.wikimedia.org/wiki/Category:360%C2%B0_panoramas)
> * [PolyHaven HDRIs](https://polyhaven.com/hdris)

Thanks @jo-chemla for these suggestions. Super helpful and much appreciated! 

I'll add that one big factor in making large sources of pano imagery (like these you shared and commercial apis) as useful as possible for testing and using in CesiumJS is info like position (of course), but also orientation (heading, pitch, roll) and field of view (so we know if images cover 360 degrees).

jo-chemla:
Yes sure! The first 2, open-data streetview-like panorama repositories (and viewers), mapillary and panoramax, do offer REST APIs to expose just that, a search endpoint which returns these camera params (image file, pose from exif or computed reconstructed geometry via sfm, etc). The endpoints can be filtered to only retrieve 360 spheres. 
 
 - Panoramax ([OpenAPI doc](https://docs.panoramax.fr/backend/api/openapi/), [swagger testing](https://panoramax.ign.fr/api/docs/swagger) and most important, [STAC compatibility](https://docs.panoramax.fr/web-viewer/05_Compatibility/)) look like this: example [Hoppscotch link](https://hoppscotch.io/?method=GET&url=https%3A%2F%2Fapi.panoramax.xyz%2Fapi%2Fsearch%3Fplace_position%3D2.3499%2C48.853%26place_distance%3D0-200%26limit%3D20) and [static link](https://api.panoramax.xyz/api/search?place_position=2.3499,48.853&place_distance=0-200&limit=20&pic_type=equirectangular) `https://api.panoramax.xyz/api/search?pic_type=equirectangular&place_position=2.3499,48.853&&place_distance=0-200&limit=20`. 
Since it is STAC-compatible, the api endpoint can be input to [stac-browser](https://radiantearth.github.io/stac-browser/) or [stac-map](https://developmentseed.org/stac-map/?href=https://api.panoramax.xyz/api/collections/bfd786db-1136-4386-9bf3-6adbebacf787)

 - Mapillary [API doc](https://www.mapillary.com/developer/api-documentation), example [link](https://graph.mapillary.com/images?is_pano=true&bbox=2.3378522,48.8416525,2.3619478,48.8643475&fields=id,geometry,computed_geometry,compass_angle,computed_compass_angle,camera_type,camera_parameters,altitude,computed_altitude,width,height,thumb_256_url&limit=10), requires an API azuth token from [here](https://www.mapillary.com/dashboard/developers) `MLY|xyz`, see this [hoppscotch link](https://hoppscotch.io/?method=GET&url=https%3A%2F%2Fgraph.mapillary.com%2Fimages%3Ffields%3Did%2Cgeometry%2Ccomputed_geometry%2Ccompass_angle%2Ccomputed_compass_angle%2Ccamera_type%2Ccamera_parameters%2Caltitude%2Ccomputed_altitude%2Cwidth%2Cheight%2Cthumb_256_url%26bbox%3D2.3478522%2C48.8516525%2C2.3519478%2C48.8543475%26limit%3D10&headers=%5B%7B%22key%22%3A%22Authorization%22%2C%22value%22%3A%22OAuth%20MLY%7C25620612364306326%7Cfa408cc55333af5ff27e1b1dde9dd07a%22%2C%22active%22%3Atrue%7D%5D) to test the REST api`https://graph.mapillary.com/images?is_pano=true&bbox=minLon,minLat,maxLon,maxLat&fields=...`

Below I'm attaching the mapping from the respective api calls response to the fields that you probably need. You can use these [jq](https://jqlang.org/) filters in hoppscotch filter to only retrievve the selection of fields/properties. 

JQ filter for Panoramax
```jq
[
  .features[]
  | {
      id,
      assets: {
        hd: .assets.hd.href,
        sm: .assets.sd.href,
        thumb: .assets.thumb.href
      },
      geometry_coords: .geometry.coordinates,
      view_azimuth: .properties["view:azimuth"],
      field_of_view: .properties["pers:interior_orientation"].field_of_view,
      geovisio_image: .properties["geovisio:image"],
      pers_roll: .properties["pers:roll"],
      pers_pitch: .properties["pers:pitch"]
    }
]
```

JQ filter for Mapillary
```jq
[
  .data[]
  | {
      id,
      assets: {
        thumb: .thumb_256_url
      },
      geometry_coords: (
        .computed_geometry.coordinates
        // .geometry.coordinates
      ),
      view_azimuth: (
        .computed_compass_angle
        // .compass_angle
      ),
      field_of_view: (
        .camera_parameters[0]?
      ),
      computed_altitude: (
        .computed_altitude
        // .altitude
      )
    }
]
```

<details><summary>JS script to extract useful params from panoramax/mapillary responses</summary>
<p>


```js


function transformPanoramaxResponse(data) {
  if (!data?.features || !Array.isArray(data.features)) {
    return [];
  }

  return data.features.map((feature) => ({
    id: feature.id,
    assets: {
      hd: feature.assets?.hd?.href ?? null,
      sm: feature.assets?.sd?.href ?? null,
      thumb: feature.assets?.thumb?.href ?? null,
    },
    geometry_coords: feature.geometry?.coordinates ?? null,
    view_azimuth: feature.properties?.["view:azimuth"] ?? null,
    field_of_view:
      feature.properties?.["pers:interior_orientation"]?.field_of_view ?? null,
    geovisio_image: feature.properties?.["geovisio:image"] ?? null,
    pers_roll: feature.properties?.["pers:roll"] ?? null,
    pers_pitch: feature.properties?.["pers:pitch"] ?? null,
  }));
}


function transformMapillaryResponse(data) {
  if (!data?.data || !Array.isArray(data.data)) {
    return [];
  }

  return data.data.map((item) => ({
    id: item.id,
    assets: {
      thumb: item.thumb_256_url ?? null,
    },
    geometry_coords:
      item.computed_geometry?.coordinates ??
      item.geometry?.coordinates ??
      null,
    view_azimuth:
      item.computed_compass_angle ??
      item.compass_angle ??
      null,
    field_of_view:
      Array.isArray(item.camera_parameters)
        ? item.camera_parameters[0] ?? null
        : null,
    computed_altitude:
      item.computed_altitude ??
      item.altitude ??
      null,
  }));
}

async function fetchPanoramax() {
  const url =
    "https://api.panoramax.xyz/api/search?place_position=2.3499,48.853&place_distance=0-200&limit=20&pic_type=equirectangular";

  const response = await fetch(url);
  if (!response.ok) {
    throw new Error("Panoramax error: " + response.status);
  }

  const data = await response.json();
  return transformPanoramaxResponse(data);
}

async function fetchMapillary() {
  const token = "MLY|xyz-token";

  const url =
    "https://graph.mapillary.com/images?fields=id,geometry,computed_geometry,compass_angle,computed_compass_angle,camera_type,camera_parameters,altitude,computed_altitude,width,height,thumb_256_url&is_pano=true&bbox=2.349,48.852,2.351,48.854&limit=50&access_token=" +
    token;

  const response = await fetch(url);
  if (!response.ok) {
    throw new Error("Mapillary error: " + response.status);
  }

  const data = await response.json();
  return transformMapillaryResponse(data);
}

```

</p>
</details> 

Edit: pretty funny what one can do with LLMs these days, not a fan of the bright UI, but you can see panoramax in green, mapillary in blue, update the query center and params, and it adds the results as a mapillary layer with pose and pitch + raw json and image. 

image

[camera-pose-explorer-panoramax-mapillary.html](https://github.com/user-attachments/files/25574970/camera-pose-explorer-panoramax-mapillary.html)

packages/sandcastle/gallery/google-streetview-panorama/main.js
lukemckinstry: duplication issue fixed via https://github.com/CesiumGS/cesium/pull/13228

jjhembd:
One last Sandcastle tweak: Some of the panoramas in the "Google StreetView Panorama" sandcastle start out looking like this:
image

Once I turn off Google Photorealistic 3D Tiles, I can see that I was simply looking at the ground.
image

Is it possible to set a more reasonable starting camera direction? Or maybe it would be enough to make the panorama view start out with Google Photorealistic 3D Tiles turned off?

lukemckinstry:
> One last Sandcastle tweak: Some of the panoramas in the "Google StreetView Panorama" sandcastle start out looking like this:

Thanks for calling this out @jjhembd . I spent a bunch of time fiddling with camera controls and was unable to get anything better and consistent (this would be a good use of the custom camera controller interface in the future)..... so I like your suggestion of turning off the GP3DT tileset when entering pano view. 

When I added that I realized that GoogleStreetViewCubeMapPanoramaProvider.loadPanorama should await loading the cube faces before calling the CubeMapProvider class to avoid a blank screen, so I made that change too.

Issue #13124, opened 2026-01-08
Support for Property Tables in Custom Shaders


mzschwartz5:
### Overview

Currently, custom shaders can access metadata stored in property attributes or property textures (see the EXT_Structural_Metadata spec). They cannot access metadata stored in property tables.

Representing property tables in shaders pose challenges because not all data types can be natively represented on the GPU - strings and variable length arrays, to name two. These are challenges faced in property textures as well, and CesiumJS support currently limits property textures to UINT8 component types (this might be to ensure support for WebGL 1?). 

Since property tables are 2D, it makes sense to represent them as textures as well. Thus, I've focused my investigation efforts so far on understanding how property textures are implemented in Cesium. I've summarized some important code paths below (they're not exhaustive, but hit the main points of how a property texture goes from glTF to on-screen):

### Loading + parsing the glTF into CPU memory

I'm not going to go into too much detail here because Cesium's glTF loading process is extremely complex, and it already handles property tables, but here's an image of the call stack at the point where a vertex attribute buffer is actually (finally) loaded into CPU memory:

<img width="672" height="1076" alt="Image" src="https://github.com/user-attachments/assets/b21e6525-f83a-492c-8005-4fbcaf860479" />

This can go even deeper depending on the buffer type (e.g. whether the buffer is embedded in the gltf or external, in a binary). But after this call, the buffer is officially how we want it on the CPU. For textures, it's a similar call stack:

<img width="1184" height="930" alt="Image" src="https://github.com/user-attachments/assets/abcedfc3-7981-4dcb-bcfc-8be50facca81" />

### Processing the model resources + uploading to GPU

The upload from CPU to GPU happens during the main update loop - specifically, from `Model.update`. Each frame, the model calls its loader to do some processing. The gltf loader, in turn, calls its component loaders (buffer loader, texture loader, **structural metadata loader**, etc.) to process. Since we can't hang the browser by spending too much time processing in a single frame, these loaders give a job to the `JobScheduler` which picks up work where it left it off each frame.

The call stack looks like this:

<img width="1158" height="1052" alt="Image" src="https://github.com/user-attachments/assets/ca4fdc66-f311-4c81-9d36-5091d1f6750f" />

(And a similar call stack for vertex attribute upload)

### Creating draw commands using those GPU handles

[...]

Issue #10480, opened 2022-06-24
3D Tiles Next Metadata Compatibility Matrix


ptrgags:
There are several different metadata capabilities in 3D Tiles Next and several possible methods
of accessing the metadata (mainly custom shaders, picking, and styling). It's admittedly confusing what is supported where, so we thought it would be helpful to make a table.

Key:
:green_circle: - currently implemented (sometimes with limitations)
:red_circle: - not yet implemented. 

| | Custom Shader | Picking | Styling |
|---|---|---|---|
|`3D Tiles 1.1` / `3DTILES_metadata`(legacy extension) ||||
| Tileset Metadata |  :red_circle: | :green_circle:<sup>1</sup> | :green_circle:<sup>1</sup>|
| Group Metadata | :red_circle: | :green_circle:<sup>1</sup> | :green_circle:<sup>1</sup>|
| Tile Metadata | :red_circle: | :green_circle:<sup>1</sup> | :green_circle:<sup>1</sup>|
| Content Metadata | :red_circle: | :green_circle:<sup>1</sup> | :green_circle:<sup>1</sup>|
|`EXT_mesh_features` / `EXT_instance_features` | | | |
| feature ID attributes | :green_circle: | :green_circle:<sup>2</sup> | :green_circle:<sup>2</sup> |
| feature ID textures | :green_circle: | :green_circle:<sup>2</sup> | :green_circle:<sup>2</sup> |
| default feature IDs | :green_circle: | :green_circle:<sup>2</sup> | :green_circle:<sup>2</sup> |
| Feature IDs without property table | :green_circle: | :red_circle:<sup>3</sup> | :red_circle:<sup>3</sup> |
| `EXT_structural_metadata` | | | |
| property textures | :green_circle:<sup>4</sup>| :red_circle:<sup>5</sup> | :red_circle:<sup>5</sup> |
| property tables | :green_circle:<sup>4</sup> | :green_circle: | :green_circle: | 
| property attributes | :green_circle: | :red_circle: | :red_circle: |

Footnotes:
1. These can be accessed through a picked object, see [`Cesium3DTileFeature.getPropertyInherited`](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileFeature.html?classFilter=cesium3dtilefea#.getPropertyInherited). This has some limitations, see https://github.com/CesiumGS/cesium/issues/10015
[...]

PR #13155, opened 2026-01-23
Property Table Textures

mzschwartz5:
**\*\*Depends on #13135\*\***

This PR is a step in adding support for representing property tables from the [Structural Metadata glTF extension](https://github.com/CesiumGS/glTF/tree/proposal-EXT_structural_metadata/extensions/2.0/Vendor/EXT_structural_metadata#property-tables) on the GPU - namely, for use in [Custom Shaders](https://github.com/CesiumGS/cesium/tree/main/Documentation/CustomShaderGuide).

Currently, property tables are only resident on the CPU. They can be used for selection and declarative styling, but not for custom shaders. The linked issue below goes into some of the challenges and work involved in using a property table on the GPU. 

This PR adds functionality to glTF loading that uploads property table properties as GPU textures. The next PR will augment metadata shader code generation to allow use of these textures.

## Issue number and link

#13124

## Testing plan

Since this is only half of the functionality, there's no good way to test. 

Testing should be mostly limited to regression testing. Here are some relevant sandcastles to smoke test (ensure parity between prod and local version):

[Custom shaders with property textures](https://sandcastle.cesium.com/?id=custom-shaders-property-textures)

[Legacy batch tables](https://sandcastle.cesium.com/?id=3d-tiles-1.1-cdb-yemen)

Test with the [EXT_mesh_features and EXT_structural_metadata 3D-tiles / glTF samples](http://localhost:8080/Apps/Sandcastle2/index.html#c=tVkJb9s4Fv4rXG+xo2wdOT2CYnJhM2kyNbaZFrVbTFEPEkqibU4lUSCp2J7A/33f4yFLspwm2N0CjR3p3edHhmeFkJr8k1BFLpjiZUamUmRk0ovNb5Pe8SSf5LHIlSZ3nC2YJKckZwtHHX4xzwJPfyFyTXnO5KS3ZzgHAzLSQjJF9JwRzVOmmIbvVBOuSFxKyXKdrohiKYs1SyZ5Cu/d87El7xC0zUimjOpSMjJ8S1IasbRPFnMezw0nt1w8T9iSiCm5vbo8H3/+dHkzfHuT3zZ0nsea37ErK2yYvEdR4PKBN+JCwquWOyZktQfhn0rkoM08m4G4nKgySrgES4VcTXKqVnlMpmUOyoAyNjKdt0GddI/cT3JC+JQELuAJm0J8k6AZoj1HSFySQhWznIWF5BlHf1QoWSbuWJvt2DI1n4K7Ze70GII1/qh8r3vaJzRPCEomHNIqCCUxk1gCpBCKG++Ei0MqIoaCtpTtMpkmSWDtowsK4l0E7Mert05AiNH/LFNHSsjtXOtCHQ0Gki7CGdfzMioVk1DDGtSGscgGVsSvo8GrZN94sq9oVsDnIAPTB7N0fDV4dl9PxHpQz+1t3yu7918ISVhUzkZzsfhFQPh4Pvsi0jJjR0TLklUMa/tlDz9s+JvxCDORsPSaasmXEBvn81jSXE2FzFTIqNK/QdPOPxdjccWXLLmSNGPOfR8jKqFIOc1fmfC8ZTPJmAr23xyGLw5fvnp52Cevfg5/fv3m4HWfHNSNMZ0uplObmlqnv2MUnfrIdTz/RPOZ13jQb2gGy+ehFp+AGEwO9l++DA/3HMnr8GCjyWX9LyGysWjVZd9ZYGeIqb1r+p1BF0lmh0fB4+9gDZFsyqTCysMa27S163WkdQI6x41v0s/DjkxM20PggREBhq4bEwIag7wbX78noCkDJmv2gqcpKJfQL9TMEMOhhUg1LyxJwlWR0pWdMH6o8RyTT7GfJnl7bljuwI0Am0Ev8pQkIi4zU/iG+tKaAxM74Xd2Sle5iP30DmlRsDy5mPM0CZwoS+l+CZVepSyMaPx9JrHcL0QqcDVMelEKD83eaFNXEwHJaKSgPTTrpEzZVBuqg87XWhQPvC1gbmBpIMWL18Wyi+jbpFcIDgNB7jMYz1pNen8YhlzkbAdDlIr4+77ifzFPPOV6340VxyMZpCv3rLYo2glpZa3abjA5agvD1I4T7qvUyTDkVNe3C8hkm5FbKxEFUr0qS/V735F/7ZO5ztILq8NVTzOUrhZdXsWOvLps3cLEtBrWxfJ2Z9oqsq9bZDyH2jOOn9ZN291bCqYkZNr0jSte1ewVjFQkynqwXE/1zZKTtRea65RttZfr83dgT2Ao+l5CbTn/rbWdPUW1ll1h3Aa/CfLs3shZ751E8ux2s2FtoRRSFLBCV/9mMANOva5wxvRH92aYqMB2Y5fqOn+3fkfBAcVApHaag6AIs4A5O4nOPNnRySA625CihABJuUFJ8HHScAGqI5/pObx4/ryyZstT4K0zfeN/HHdRfqFpybqDUvfbwxpj/XMw/x95pIpj+/PZfY1yTY7I5oERv26FwUUOZdUK8VdTMVCHIvoTNomb7jDTRWRGaAIzlZIRQpq/4646IsMpQiSubE2SOoq58hUJtZhbKqeWJaGh/wBv5IIroPmpwmY/teiqwrU2OKkBqmfJA8XqCNq10oEBbTa4cqIxa4YXek5pmscMsHUHSnPkm5KtJDxCp3tjFdUyMESVsE2xf+eAQlPTy37HQj5xClQ4YYGR9fs2EwAIMXgIXBOLYcsi8djWHB8IDB6uVzsYgU92J9HPe29SA0ONzOAbFTRml7h23lmioIGAY5rfUYU17GSEAEaGeVEa4CHyoMpzgPZnm+GNbehm6Zgt7QZ1E9v1kc1WG3LD00pUCHv/o9sjHn5Z5mmV9M76qk9yo/35accEnfR8nHqbOdpQ43YIqOkyKVwet0m/7iJdGdKHN2DNYDRj3d+VqvGqYOH1h8+jy5vrD18u/do+/R/+szAAch4DxE8RlAJU1GVx1Fp+rkBrgBKhgxJZAyfiZLbosvu8zJbmyLO174YgAsPx1u7+JqzkzZePhJctri3waApVziIavH7ZJ/7/Qfhmz1Vvt4A60DssluTFQQX2uhkyKmc8Hzv0eFhRuxHTYqqjt22/dwTruI69sG3t+8AL8GO4ZZ+G7w7tgGz/dhv4ADrFwSRXZu+aYcXNec2nUzkwRC1bxLaOOgksu0QscgJ5KkMyMq8xjChbTH3JwJwztZUKmjQvO5xie2J2Ug2XRVHVkdnOVQ8hNzALvSPoMHyrCrQVkK2qHBltHwozAHM47/brilrRdfl0qwVVHRHDYx9ARo3TR6R1DxPsbc7z9sqheTOzuZY59mSdOXZv10bfejuL9bzRO8qhn1MfUGXDBqWNZJk7H7qIV6MBx1ZEpS9PVYsOgsZvpmO3A2ftqmbwMDnXgKAjOIadw9x0KGiM1sB49sSXv49vAGiXMbDQ9CZjmsKqpIMnCbmqbucs6LxDABqjtxFVUJwQfVq/wqNepgsHUdyUm4dqRJuYOdCP5evtgpWHdwx9t1MeEQTMGXz5b0LwYxFPDYC2Ev9/7l+XuPjSzT2G2lgPR4NHeP94CedgnpqT6m4PRoueE70Q2D8K22DjOUSE0XgOvalEzClOLkPt1brfnO7aYebRno/mVLIn5/phLudh5VfTX+/TnN5hrii6veW1GdzmPmvT8Fk75U9O70UKgXxSOrc5utNXpaBWtSadfgPh9qF4V2aSqFhBJXxL3WKAaG7lOOFTcB+XYJXt2FvzSL8vBD5cIlx7jNM7yLs9ngKX72HjZJcPleXNW4nYKuoTbwXQa6vVe4Zn3g38w+vvsR3z17Cpg8aM32v/OQJPooWHfW7n26sRl4N6mgynLdXaVk7YctflxzAZMe0CbQgfWLS35211R7V72Ztn90bA+nZrGXeu4Yf/KmNEHbdod17f1qmfuJsbjfr0/TzdCmNrS3fE+cDemu98/+LHJdOhtiqcc+uAqUtAfo0LasRsFWir4NqPXK1OAjOm3THgl9UwgaOAJ+ntNS6WW5gPTev1eycGq5+h7//i9s+TpUyDMBxoBh2A1T6ISjhywpFDGX9OBp7lBA4dhCenHX+OtIME3kzLNB2ZS9yzkwHQN9gQ7ULbfIDtjCgUSOYvzt7bh2EYngzg122uyr/q3X8A)

- Try all drop down examples.
- Would also be a good idea to use SpectorJS to inspect the generated shaders. 

Note: some issues *may* exist that may have been solved in the [next PR](https://github.com/CesiumGS/cesium/pull/13163). For instance, I didn't know at the time that `componentType` is undefined for `ENUM` type properties - which instead use `valueType` to specify bytewidth. Many instances of `componentType` change to `valueType` in the next PR.

# Author checklist

### DIFF
--- a/packages/sandcastle/src/util/bucket-client.ts
+++ b/packages/sandcastle/src/util/bucket-client.ts
@@ -36,7 +36,10 @@ function loadSandcastle(code: string, html: string) {
   });
 
   const div = document.createElement("div");
-  div.innerHTML = sanitized;
+  // Stylesheet imports are weirdly broken in Firefox 140+. This is a hacky workaround
+  // confirmed still broken on v147
+  // https://github.com/CesiumGS/cesium/issues/12700
+  div.innerHTML = sanitized.replace(/@import/, "@import ");
   document.body.appendChild(div);
 
   const script = document.createElement("script");


packages/engine/Source/Scene/Model/MetadataPipelineStage.js
```diff
@@ -203,7 +203,7 @@ function getPropertyTextureInfo(propertyTexture, statistics) {
   const classStatistics = statistics?.classes[classId];
 
   const propertiesArray = Object.entries(propertyTexture.properties).filter(
-    ([id, property]) => property.isGpuCompatible(),
+    ([id, property]) => property.classProperty.isGpuCompatible(),
```
mzschwartz5, line 206: I'm aware this function call is missing an argument - it gets fixed in #13163

packages/engine/Source/Scene/MetadataClassProperty.js
mzschwartz5, line 1: The changes to this file are just me moving functions from `PropertyTextureProperty.js` to a lower level class so that they can be used by property table properties as well.

packages/engine/Source/Scene/parseStructuralMetadata.js
mzschwartz5, line 1: These changes constitute the meat of this PR. There has been [significant discussion](https://github.com/CesiumGS/cesium/issues/13124#issuecomment-3775520700) and investigation as to *where* to create the texture for the property table.

Ultimately, I decided on doing it in this parse function, and doing it synchronously. We can always revisit this if it proves to be an issue, but it's certainly the most straightforward way to do it, and it seems to work well (in my limited testing).

It does have some implications: the `PropertyTable` stores the texture (in contrast to other glTF resources, which are typically stored on the relevant glTF loaders), and thus some new methods are needed to clean up resources.

packages/engine/Specs/Scene/MetadataClassPropertySpec.js
mzschwartz5, line 1: These unit tests are really just moved from the other file, nothing super new here.

packages/engine/Source/Scene/PropertyTable.js
```diff
@@ -533,4 +548,15 @@ PropertyTable.prototype.getExactClassName = function (featureId) {
...
+/**
+ * Destroys any resources that need cleaning up in the property table.
+ *
+ * @private
+ */
+PropertyTable.prototype.destroy = function () {
+  if (defined(this._texture)) {
+    this._texture.destroy();
+  }
+};
```
lilleyse, line 560: [`destroyObject`](https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/destroyObject.js) is typically called in `destroy` functions. Example from `BatchTable.js`:

```suggestion
PropertyTable.prototype.destroy = function () {
  this._texture = this._texture && this._texture.destroy();
  return destroyObject(this);
};
```

packages/engine/Source/Scene/parseStructuralMetadata.js
```diff
@@ -99,4 +119,166 @@ function parseStructuralMetadata(options) {
...
+    wrapS: TextureWrap.CLAMP_TO_EDGE,
+    wrapT: TextureWrap.CLAMP_TO_EDGE,
+    minificationFilter: TextureMinificationFilter.NEAREST,
+    magnificationFilter: TextureMagnificationFilter.NEAREST,
+  });
+
+  return Texture.create({
+    context: context,
+    pixelFormat: PixelFormat.RGBA,
+    pixelDatatype: PixelDatatype.UNSIGNED_BYTE,
```
lilleyse, line 199: I see how `RGBA/UNSIGNED_BYTE` makes integration with https://github.com/CesiumGS/cesium/pull/13135 easier, but `RGBA/FLOAT` would allow for accessing larger elements, e.g. `VEC2/3/4` of `FLOAT`.

I don't want this to hold up the PR, so maybe just update the [compatibility matrix](https://github.com/CesiumGS/cesium/issues/10480) to list unsupported `type`/`componentType`

packages/engine/Source/Scene/StructuralMetadata.js
```diff
@@ -240,4 +240,18 @@ StructuralMetadata.prototype.getPropertyAttribute = function (
   return this._propertyAttributes[propertyAttributeId];
 };
 
+/**
+ * Destroys any resources that need cleaning up in the structural metadata.
+ *
+ * @private
+ */
+StructuralMetadata.prototype.destroy = function () {
```
lilleyse, line 249: Same comment here

packages/engine/Source/Scene/parseStructuralMetadata.js
lilleyse: This approach feels pretty clean to me.

lilleyse:
As discussed offline, this sandcastle is significantly slower on this branch due to property table texture creation. Need to investigate further.

```js

// A demo showing how to isolate elements in an architectural design by CategoryID
// Snowdon Towers sample data courtesy of Autodesk Revit.

// Set to 1 PM Philadelphia time in UTC
viewer.clock.currentTime = Cesium.JulianDate.fromIso8601(
  "2024-11-22T18:00:00Z",
);

// Set the initial camera view
viewer.camera.setView({
  destination: Cesium.Cartesian3.fromDegrees(-79.886626, 40.021649, 235.65),
  orientation: {
    heading: 0,
    pitch: Cesium.Math.toRadians(-20),
    roll: 0,
  },
});

// Add the architectural tileset
let archTileset;
try {
  archTileset = await Cesium.Cesium3DTileset.fromIonAssetId(2887123);
  viewer.scene.primitives.add(archTileset);
} catch (error) {
  console.log(`Error loading tileset: ${error}`);
}

// Add the site tileset
try {
  const tileset = await Cesium.Cesium3DTileset.fromIonAssetId(2887129);
} catch (error) {
  console.log(`Error loading tileset: ${error}`);
}
```

javagl:
> As discussed offline, this sandcastle is significantly slower on this branch due to property table texture creation. Need to investigate further.

I did not "strongly advocate for" or "recommend" loading this synchronously, but certainly strongly suggested that it will likely not make a measurable difference. So I was a bit curious. I added a small `allTimesLoaded` event and measurement in that sandcastle:

- [Here it is on this CI build](https://ci-builds.cesium.com/cesium/gltf-buffers-to-texture/Apps/Sandcastle2/index.html#c=pZRtb9owEMe/yglNatiogdBSymg1BH3RqdWqlu3FlBczyUGsOjayHRCr+O47O6Gl3cObSZGinO/p97+LRbHSxsF74BYmaEVZwMLoApJGGr6SxsdEJSrVyjqwjpPvBazQLLQpuEqRKb2JmsGn3YYxZFhosLneCLUEeoHTIKyW3CGgxAKVsyAUcHpMmguHqSsNlxRoxVLBfAsT8l1qs72ehpwPVCHTCmZ6g8aC5cVKImTccUh1aRzaLegFjEunKccj3ONaOLZveS2QwqhnhZsakH0LtmiPONHKcaHQJI0WPCUKYCn1HIew4NJionbPeA/oPE8X7m7hLheSZyhXueDgRIGe6utskqiqJEulTh9ZWhpDzDPvcLFv4HMpBVdT4mRe7GurB/1ON/Klk0bciU+Ou93jOJ51B8NOh57v1FmiXreR+4LCCZIu5QUaHlBfqgcbs+g8bRSwSB4nFHdCq+G+lQn3ClI3vdDKFJcG0UbHZ+dsMOj3434LTjqsE3f7J+ctiHunrH/abPls2ggCq7OF/AA58owGP4ROqzKshEvz52K33OXM6Xty4oqKxJ1m7We0lPuoXetQ83GWBdjX2+KERGJLlCQt/NGsMlCUM9uqnQMzSc83XLhn6vDqTevjagpajS19XGdRPBicdeOe7wHqDWI2RYVsZURBoq/RMp5l0UEJ77yjURAvRGiMNs2qDb+HWiKTehn9uPIHIHWQaU8xhHdPIWL3I2R5Q24J+wW4Wmv3H1znvsi/qNwLEclZp+JShqw31Dtm3u9qTfO/EdZRChMtSpX6XYia+2U4BD8ay3poNBYMAmB2VAlceTpAlf3lbnmbLWlMaQt8Nfpf4APpTaHH1fXUbIbtabQaI+u2Ei99+CdRXXOlkRFjbYd0h9DfZ9vzMn0kuNRaX2fU3oeMMrEGkV384Y6AVHJr6WRRSvkgfmLSuBy1yf9VWD3iL2s0km+9S969vKmMjLFRmz5/j3Jayzk3Bxl/AQ)
- [Here it is on `main`](https://sandcastle.cesium.com/index.html#c=pZRtb9owEMe/yglNatiogdBSymg1BH3RqdWqlu3FlBczyUGsOjayHRCr+O47O6Gl3cObSZGinO/p97+LRbHSxsF74BYmaEVZwMLoApJGGr6SxsdEJSrVyjqwjpPvBazQLLQpuEqRKb2JmsGn3YYxZFhosLneCLUEeoHTIKyW3CGgxAKVsyAUcHpMmguHqSsNlxRoxVLBfAsT8l1qs72ehpwPVCHTCmZ6g8aC5cVKImTccUh1aRzaLegFjEunKccj3ONaOLZveS2QwqhnhZsakH0LtmiPONHKcaHQJI0WPCUKYCn1HIew4NJionbPeA/oPE8X7m7hLheSZyhXueDgRIGe6utskqiqJEulTh9ZWhpDzDPvcLFv4HMpBVdT4mRe7GurB/1ON/Klk0bciU+Ou93jOJ51B8NOh57v1FmiXreR+4LCCZIu5QUaHlBfqgcbs+g8bRSwSB4nFHdCq+G+lQn3ClI3vdDKFJcG0UbHZ+dsMOj3434LTjqsE3f7J+ctiHunrH/abPls2ggCq7OF/AA58owGP4ROqzKshEvz52K33OXM6Xty4oqKxJ1m7We0lPuoXetQ83GWBdjX2+KERGJLlCQt/NGsMlCUM9uqnQMzSc83XLhn6vDqTevjagpajS19XGdRPBicdeOe7wHqDWI2RYVsZURBoq/RMp5l0UEJ77yjURAvRGiMNs2qDb+HWiKTehn9uPIHIHWQaU8xhHdPIWL3I2R5Q24J+wW4Wmv3H1znvsi/qNwLEclZp+JShqw31Dtm3u9qTfO/EdZRChMtSpX6XYia+2U4BD8ay3poNBYMAmB2VAlceTpAlf3lbnmbLWlMaQt8Nfpf4APpTaHH1fXUbIbtabQaI+u2Ei99+CdRXXOlkRFjbYd0h9DfZ9vzMn0kuNRaX2fU3oeMMrEGkV384Y6AVHJr6WRRSvkgfmLSuBy1yf9VWD3iL2s0km+9S969vKmMjLFRmz5/j3Jayzk3Bxl/AQ)

The times vary wildly between the runs. There is no real concept of "benchmarking" for CesiumJS 🤷‍♂️  But I don't see a significant or noticable _difference_ between the two. 

Looking at a profiling run with the new version...

Cesium PropertyTexture Performance

... it suggests that the texture creation takes 15ms, which might be _considerable_ if it fully happened in a single frame, but that's certainly not the case here. So whatever the performance degradation is, it may be necessary to narrow it down first.

lilleyse:
@javagl interesting... I was testing on https://github.com/CesiumGS/cesium/pull/13163 and assumed this PR was the problem but that seems not to be the case.

* [Here it is on `property-table-custom-shader`](https://ci-builds.cesium.com/cesium/property-table-custom-shader/Apps/Sandcastle2/index.html#c=pZRtb9owEMe/yglNatiogdBSymg1BH3RqdWqlu3FlBczyUGsOjayHRCr+O47O6Gl3cObSZGinO/p97+LRbHSxsF74BYmaEVZwMLoApJGGr6SxsdEJSrVyjqwjpPvBazQLLQpuEqRKb2JmsGn3YYxZFhosLneCLUEeoHTIKyW3CGgxAKVsyAUcHpMmguHqSsNlxRoxVLBfAsT8l1qs72ehpwPVCHTCmZ6g8aC5cVKImTccUh1aRzaLegFjEunKccj3ONaOLZveS2QwqhnhZsakH0LtmiPONHKcaHQJI0WPCUKYCn1HIew4NJionbPeA/oPE8X7m7hLheSZyhXueDgRIGe6utskqiqJEulTh9ZWhpDzDPvcLFv4HMpBVdT4mRe7GurB/1ON/Klk0bciU+Ou93jOJ51B8NOh57v1FmiXreR+4LCCZIu5QUaHlBfqgcbs+g8bRSwSB4nFHdCq+G+lQn3ClI3vdDKFJcG0UbHZ+dsMOj3434LTjqsE3f7J+ctiHunrH/abPls2ggCq7OF/AA58owGP4ROqzKshEvz52K33OXM6Xty4oqKxJ1m7We0lPuoXetQ83GWBdjX2+KERGJLlCQt/NGsMlCUM9uqnQMzSc83XLhn6vDqTevjagpajS19XGdRPBicdeOe7wHqDWI2RYVsZURBoq/RMp5l0UEJ77yjURAvRGiMNs2qDb+HWiKTehn9uPIHIHWQaU8xhHdPIWL3I2R5Q24J+wW4Wmv3H1znvsi/qNwLEclZp+JShqw31Dtm3u9qTfO/EdZRChMtSpX6XYia+2U4BD8ay3poNBYMAmB2VAlceTpAlf3lbnmbLWlMaQt8Nfpf4APpTaHH1fXUbIbtabQaI+u2Ei99+CdRXXOlkRFjbYd0h9DfZ9vzMn0kuNRaX2fU3oeMMrEGkV384Y6AVHJr6WRRSvkgfmLSuBy1yf9VWD3iL2s0km+9S969vKmMjLFRmz5/j3Jayzk3Bxl/AQ) - the slowness is much more noticeable

Strange, because the sandcastle doesn't even use custom shaders.

javagl:
Interesting indeed...

Cesium PropertyTexture Performance 0002

Quickly skimming over the base state, there now are these [property table infos](https://github.com/CesiumGS/cesium/pull/13163/changes#diff-f9c3ee8941a52b898f1d9c9113fba41d7bf7c83491cc6715bdc88e233cdcde2eR100). 

The number of properties does not seem to be excessively large. **But**... a wild guess what could very well cause the slowdown: It will create a new shader for each different _set_ of properties. (For different _numbers_ of properties, and if that is applicable: Even for every different _order_ of properties!). 

From quickly debugstepping through that, the number varies wildly, between 2 and ~50. So it might very well be that it has to compile and link dozens of shaders, with each of them being ~"considerably complex" for itself (because it's the model shader), and with additional (considerable) complexity due to all the metadata stuff. This will 💥 the `ShaderCache`, and could cause the lags and overhead downstream, stalling everything due to the compilation overhead.

mzschwartz5:
> a wild guess what could very well cause the slowdown: It will create a new shader for each different set of properties. (For different numbers of properties, and if that is applicable: Even for every different order of properties!).

This would surprise me, but I'll have to look into it more closely. I _believe_ I made it so that every property in a property table gets put in the shader, whether or not a given model actually uses it (and use the default value if it's unused)... so for there to be a lot of different shaders being compiled and linked here, it would have to mean a lot of different property tables.

Maybe I'm wrong about what I implemented though - in which case, the solution is simple (do what I thought I was already doing).

javagl:
Disclaimer: I have not investigated the resulting shader code. I only saw that the pipeline stage was called with all these different sets of property values (and I was actually in a call at that time, so didn't check how many shaders it actually did build). 

If all that should already go into a single shader, then the answer to the question why `getProgramParameter` suddenly seems to take multiple seconds per frame remains open. (I mean, it's not just "slow" in that branch - it feels like the whole browser tab is blocked).

mzschwartz5:
Looking closer, I think you're right Marco. In an example I debugged, each gltf primitive specifies a subset of properties from the table it references. The texture and shader only use those referenced properties. Which means a different texture and shader for each gltf primitive, potentially.

That's no good... each property table should be uploaded once, in its entirety. And each primitive referencing that table should have a shader that has access to all properties, even if it only uses some (the rest will just be default values).

This will require a little rethinking about how I load the tables as textures. Right now, it's happening per gltf primitive - but I think it's more of a per-tileset operation. (But I don't think there's a single place in a tileset where all property tables are defined... so... is this even possible?)

javagl:
> This will require a little rethinking about how I load the tables as textures. Right now, it's happening per gltf primitive - but I think it's more of a per-tileset operation. (But I don't think there's a single place in a tileset where all property tables are defined... so... is this even possible?)

The closest one could again be a schema that is "pulled up" into the tileset and defines some sort of holistic view on the metadata of all glTFs that may appear. Otherwise, it could happen at any point in time that a new tile with a glTF is loaded that uses a metadata schema that was never used before.

One rough idea for a workaround could be to have some place where all the properties are collected, and the shader builder mechanism only kicks in and regenerates (and assigns) a new shader when a new property is encountered. But I don't have enough context to say how feasible that sounds...

mzschwartz5:
Okay so here's what was happening: in certain tilesets, primitives in tiles only define a subset of properties that are available on the metadata class, as defined in the tileset schema. This is something that happens often, for instance, when Design Tiler is used without disabling `merge-classes`. Design Tiler automatically merges all classes together into one, _AND_ omits unused properties from the merged class on individual primitives. (That makes sense in a way, as a performance optimization, but ironically it's causing performance issues here).

Since different primitives of the same class have different property subsets, different shaders are created for each. The natural solution is to create shaders and property table textures based off the class schema - not based on the primitive's properties. If a property is defined in the schema but not referenced in the primitive, it should get uploaded with its `noData` value.

___

I implemented this locally, and it does perhaps speed up the shader compilation step (with the above offending sandcastle). But it's still very slow compared to the tip of `main`... I'm not exactly sure why :/ In the `MetadataPipelineStage`, this PR adds:

```JS
function processPropertyTableProperty(
  renderResources,
  propertyInfo,
  webgl2,
  propertyIndex,
) {
  addPropertyTablePropertyMetadata(renderResources, propertyInfo, webgl2, propertyIndex);
  addPropertyMetadataClass(renderResources.shaderBuilder, propertyInfo);
  addPropertyMetadataStatistics(renderResources.shaderBuilder, propertyInfo);
}
```

If I comment out everything here, the sandcastle performs very well (no surprise). If I uncomment just `addPropertyMetadataClass`, the performance issues come back. So merely adding the structs that contain the metadata - and not even initializing their values or sampling from textures - causes performance issues. There's still something causing shader compilation to explode. 

I've stepped through `addPropertyMetadataClass` and for each primitive of the same class, it appears to be uploading the same metadata properties to the structs in the same order. So what Marco said about different permutations of the same properties causing different shaders to compile probably isn't the case here? Could it be just the shear number of properties? This Snowdon example has ~200 metadata properties in some tables.

mzschwartz5:
If I log the number of shaders in the cache on load:

|  | Number of Shaders |
|---|---|
|Without Property Tables | ~50 |
|With Property Tables | ~60 |

So that's further evidence it's not the _number_ of shaders being compiled (because the time difference in compilation is far slower than a factor of 6/5). So then it must be the content of the shader. When I try to inspect the a compiled shader via SpectorJS, it fails on the Snowdon example... I've never had that happen before.

edit- this failure on SpectorJS may have to do more with the number of draw calls. Even with the property table stuff commented out, it still takes quite a while for SpectorJS - and the number of commands issued is huge, and each with like 60k indices to draw.

mzschwartz5:
Another interesting note is that it's only Firefox which seems to be slow at compiling the shaders...

Chrome is very fast - as fast as main now... That's telling. I suppose something in Firefox's webgl implementation differs and is suboptimal here. There might not be anything to do about that. 

Disabling `merge-classes` when tiling the Snowdon data would probably help a lot here (regardless of browser). That way each primitive would have a much smaller set of properties, rather than the full merged class of ~200 properties.

edit- I manually limited the number of properties to something reasonable (1, 10, 100, in turn). It definitely has an effect on the shader compilation time in firefox. And it's the same speed whether or not I include the shader code that initializes the metadata. As I said in an earlier comment, it's merely the presence of the metadata structs that's slowing down compilation.

packages/engine/Source/Scene/parseStructuralMetadata.js
mzschwartz5: Can you elaborate on this? I'm not sure I understand the problem. The idea is that this texture is essentially just four channels of raw data. Accessing larger elements is part of what I wanted to support here. (Unless I'm misunderstanding what you mean).

Regardless, I definitely will update the compatibility matrix.

lilleyse: Ah, what I'm saying is that

```
    pixelFormat: PixelFormat.RGBA,
    pixelDatatype: PixelDatatype.UNSIGNED_BYTE,
```

only gives you 4-bytes of raw data per element whereas

```
    pixelFormat: PixelFormat.RGBA,
    pixelDatatype: PixelDatatype.FLOAT,
```

would give you 16.

Certain `type`/`componentType` combinations like `VEC2` of `FLOAT32` won't fit in 4-bytes and so we'd need to mark those as unsupported in the compatibility matrix.

mzschwartz5: I see. I think that's how I intended this to work, and why `MetadataClassProperty.isGpuCompatible` checks that the total byte size of a single metadata element does not exceed 4 bytes. 

I think I also had some other reasons for using `UNSIGNED_BYTE` ... perhaps having to do with unpacking logic and / or sampling and / or normalization... if I remember these I'll comment them here. 

But yes, at the very least, I'll update the compatibility matrix. More support is better than none!

mzschwartz5:
@lilleyse The performance issues should be completely fixed now. But not on this PR (as a reminder, we learned the regression was actually introduced in the next PR). There were two pieces involved in the fix: 

1. In #13163 - I added `noData` rows to the property table texture where properties are defined in the schema but not used by a given primitive (which only really occurs with `merged` classes output by our tilers)
2. In #13207  - I changed the `MetadataPipelineStage` to only add metadata to the shader that are used in other stages (which, right now, is just the `CustomShader` and `PointCloudStyling` stages).

The first fix reduces the number of shader variants that need to be compiled (one per metadata class, rather than one per subset of metadata class properties). The second fix reduces the number of metadata properties declared in the shader in most cases, which reduces compile time (most noticeably in Firefox on MacOS).

mzschwartz5:
@lilleyse Ready for re-review

lilleyse (approved):
Looks good, thanks @mzschwartz5

packages/engine/Source/Scene/parseStructuralMetadata.js
mzschwartz5: Also, for posterity - if someone is implementing this idea: since we're already only supporting WebGL 2 with property table textures, you might as well use the `RGBA32UI` texture format. It's basically the 32-bit version of what we're already using (RGBA8 or UNSIGNED_BYTE). There are fewer pit falls when dealing with unsigned bytes directly compared to trying to pack things as floats.

javagl:
This was already merged, but "bumped" by a comment, and even though I have not been tracking all details recently, I'm a bit curious: Regarding the issue where it created ~"one shader for each primitive" (the discussion [starting at this comment](https://github.com/CesiumGS/cesium/pull/13155#issuecomment-3891492927)), I wondered whether there could/should be a dedicated spec for this, with pseudocode like
```
load("ThreePrimitivesWithDifferentAttributeSets.gltf");
...
assert(shaderCache._numberOfShaders !== 3); // Not three...
assert(shaderCache._numberOfShaders === 1); // ... but only one!
```
(Maybe it's not important or doesn't make sense - I'll need to digest the discussion and changes here...)

PR #13157, opened 2026-01-23
Fix: polyline material batching to prevent unnecessary re-renders

### LINKED ISSUE #13136: Processing a single polyline re-renders all polylines with the same material
### What happened?

When calling `CzmlDatasource.process(czmlData)` to update an existing entity, the corresponding entity will "blink" as it is re-rendered on the map. However, if there are multiple polylines with the same material properties on the map, processing a single one will re-render them all, causing a widespread "blinking" effect. 

### Reproduction steps

1. Create a datasource and add multiple polylines with the same material property (e.g. solid color)
2. Call `datasource.process()` on one of the polylines (no actual data change is needed)
3.  All polylines will be re-rendered on the map

The expected result is that only the polyline entity that is affected by the `process` call should be re-rendered.

### Sandcastle example

https://sandcastle.cesium.com/#c=5VZtT9swEP4rVj+FiaVJuqqitGhQBkNkgNYytBEEbmJaC8eObIcCE/995yRNk5WXsYkvo1LT+O655y7nx+nROBFSo3cIKzQgiqYxupQiRkEjzFZBYz3gAQ8FVxqFdzHbFmEaE65RH53+DDiCD426gI8KR9BYzc0cx8Q4Bj+++GiXiJhoSYnqoiPBbhnlpEReE6mo4Abs2k5hvj+rJz7C4RXRLuQ13nrupGB0l5KfTKkmyPhQyHCckAhpgTSRmPISzPCYsC4qKBG4b7QJ9k3YghOhS8Ezh+ckN+gzYddE0xBXAZSxgWBCdotu2tnKHu5/3/KPP5U4kWpT00PQLX9zsP878IRGetpFXmlX+paRMtI3DzA0Jntnz/fPNw+2zw+PR/7ewTzlffGbCEV11uvyaUMstZhInExpuE0mkpg9On3vem3bWUVex1ydsyWevOMVnjm1qtgeZe+0gRq+wG0yFYuz8vnuy7sYw25RzGqsSjAaFd2rmCHdAzaE5GSMIavXzhJVLouE9aTV+8XdLN+FdmnIJDUSu1KkHISoZUpqfXpYwx7qPy5h7xUk7L1tCbe8V5Nwy6lIuOX8txI2VdVF3HpKxK0lEe8yMfsnDbfeuIbXXk/DnaqGOy/R8Dy92d2XqNh1IKH7YQ0EbBI+qWKEJsB+JGYE6Bzbq7k0Togsfe0/E7/r/NUL/JoSyAPzByez+Y5/y2zWfFoagDBB0UQGjZX1eVyENVYilSGpxw7MNAW+YeazFqdH5dEBX0TaTODIqs5fKwaAZ5jOC7OjkkvZOIqsRfQSWyIFoJRVmasM5kmE9yyiZRBFMXdCxCOxVMMQ8yjESsMJgRJHQrAxllup1oJbF19JwYnyfy1khThVpFi65o0RNMbQoSto0MUqslZQfyNX2POV528wwUwrJ9DsHcqpmpIIqAJ+n1fXWG30shO9YfAfaT4bp5JZtt3UJE4YHALVHKeG1Q6VMmG95jykF9FreBH2H1CDEZpS4LlMGRvSO5h9N3pNwNfCzCZTPjmEeZjhWwOZuht+brRtu9eE5XKUzrtYYfwF

### Environment

Browser: Chrome
CesiumJS Version: 1.137
Operating System: Ubuntu

MohammadShujaullah:
Fixes #13136

Description--
When processing a single polyline through CZML, all polylines with the same material were re-rendering due to overly aggressive batch invalidation. This fix enhances the material change detection to be more intelligent.

Changes Made--
Enhanced Batch.prototype.onMaterialChanged to check if material values actually changed
Added material value comparison logic to prevent unnecessary batch invalidation
Specifically handles color materials by comparing actual color values
Stores last material value for comparison
Only invalidates batches when materials truly change functionally

Root Cause--
The issue occurred because:
CZML creates new material property instances even when data is identical
This triggers definitionChanged events
Previous code blindly invalidated entire batches
All polylines in the batch would re-render, causing widespread "blinking"

Solution--
Now the system:
Captures current material values when change events occur
Compares with previously stored values
Only invalidates batches when actual value changes are detected
Preserves existing batch structures when materials are functionally identical

Testing--
Build passes successfully
Logic correctly identifies when materials haven't actually changed
Prevents widespread re-rendering while maintaining proper update behavior

### DIFF
--- a/packages/engine/Source/Scene/Vector3DTileClampedPolylines.js
+++ b/packages/engine/Source/Scene/Vector3DTileClampedPolylines.js
@@ -6,7 +6,7 @@ import ComponentDatatype from "../Core/ComponentDatatype.js";
 import defined from "../Core/defined.js";
 import destroyObject from "../Core/destroyObject.js";
 import Ellipsoid from "../Core/Ellipsoid.js";
-import FeatureDetection from "../Core/FeatureDetection.js";
+
 import IndexDatatype from "../Core/IndexDatatype.js";
 import OrientedBoundingBox from "../Core/OrientedBoundingBox.js";
 import Matrix4 from "../Core/Matrix4.js";
@@ -557,10 +557,7 @@ function createShaders(primitive, context) {
   )(Vector3DTileClampedPolylinesFS);
 
   const vs = new ShaderSource({
-    defines: [
-      "VECTOR_TILE",
-      !FeatureDetection.isInternetExplorer() ? "CLIP_POLYLINE" : "",
-    ],
+    defines: ["VECTOR_TILE", "CLIP_POLYLINE"],
     sources: [PolylineCommon, vsSource],
   });
   const fs = new ShaderSource({
--- a/packages/engine/Source/Scene/Vector3DTilePolylines.js
+++ b/packages/engine/Source/Scene/Vector3DTilePolylines.js
@@ -4,7 +4,7 @@ import ComponentDatatype from "../Core/ComponentDatatype.js";
 import defined from "../Core/defined.js";
 import destroyObject from "../Core/destroyObject.js";
 import Ellipsoid from "../Core/Ellipsoid.js";
-import FeatureDetection from "../Core/FeatureDetection.js";
+
 import IndexDatatype from "../Core/IndexDatatype.js";
 import Matrix4 from "../Core/Matrix4.js";
 import Rectangle from "../Core/Rectangle.js";
@@ -456,10 +456,7 @@ function createShaders(primitive, context) {
   )(PolylineFS);
 
   const vs = new ShaderSource({
-    defines: [
-      "VECTOR_TILE",
-      !FeatureDetection.isInternetExplorer() ? "CLIP_POLYLINE" : "",
-    ],
+    defines: ["VECTOR_TILE", "CLIP_POLYLINE"],
     sources: [PolylineCommon, vsSource],
   });
   const fs = new ShaderSource({


packages/engine/Source/DataSources/StaticGeometryPerMaterialBatch.js
```diff
@@ -54,7 +55,42 @@ function Batch(
...
+    if (currentMaterial.type === this._lastMaterialValue.type) {
+      if (currentMaterial.type === "Color") {
+        const currentColor = currentMaterial.uniforms.color;
+        const lastColor = this._lastMaterialValue.uniforms.color;
+        if (Color.equals(currentColor, lastColor)) {
+          // Colors are the same, don't invalidate
+          return;
+        }
+      }
+      // For other material types, we might need more sophisticated comparison
```
donmccurdy, line 87: In a recent PR (for a different performance issue), I added `deepMaterialsEqual(a, b): boolean`—

- #13141

That might be helpful here, too? Or if it's _not_ helpful here, I'd like to improve that in the other PR. :)

mzschwartz5:
There are a handful of `Static*Batch` classes in CesiumJS. It's quite unfortunate, really - they're complex and buggy, and all very similar with small differences. We have an issue about that [here](https://github.com/CesiumGS/cesium/pull/12722).

Anyway, I'd like to avoid making more differences between these classes if possible. I'm guilty of this myself, though - #12722 fixed some flickering / flashing issues with materials due to aggressive batching. I only ever applied the fix to the static *ground* geometry variant class, but I wonder if applying the same changes to the `StaticGeometry...` variant would fix your issue.

If so, that would be preferable as it would mean keeping these similar classes in sync better.

mzschwartz5:
Going to close this due to inactivity, but feel free to reopen if and when you revisit it.

PR #13163, opened 2026-01-26
Property table custom shader

mzschwartz5:
# Description

**\*\*Note**: depends on #13155\*\*

This PR adds shader generation logic in the `MetadataPipelineStage` for binding and sampling property tables (represented as textures).

This is the last PR in a series to support property tables in custom shaders. There are several future follow-ups:
- Not all property types are supported in custom shaders, but many more useful ones are. 
- Further, this support only extends to WebGL 2. 
- Also out of scope for now - support for updating properties dynamically and having those updates be reflected in custom shaders.
- Finally, primitives with multiple feature ID sets that access the same property table are unsupported. This could possible be supported in the future (`Cesium3DTileset` constructor options support [specifying a feature ID set](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#featureIdLabel)).

Also, TODO: update documentation ([here](https://cesium.com/learn/cesiumjs/ref-doc/CustomShader.html?classFilter=customsh), and [here](https://github.com/CesiumGS/cesium/tree/main/Documentation/CustomShaderGuide)). I think that may be a follow-up PR.

## Issue number and link

#13124

## Testing plan

Testing plan for #13155 still applies - as regression testing.

Additionally, [this sandcastle](http://localhost:8080/Apps/Sandcastle2/index.html#c=lVaPU+I4FP5XMszdCF4NVKiuiMztgno4IqNUR93euKGNEGmT2qYg7PC/32vS8kPlbs/Rpnn53pf3vry8yoJQRBLtIhKjFo1ZEqDnSATIKbhq5hSOHe5wvYQ7gmOPPpPEl19dAMS2GFOOTgBOZxejwbnLeuyiczvvmFesE3f4jeW2OgedcXh/17o4wgB69c7HKWj02L55eTy/eLnqm/7VvDO7tK+nj/apfLDd6gOr7HfPu9bVecfs7l9XH+0xu2xdhI9A1p2fTnt2t5bOH+6vWe/ltHr1cv3Waw/nPds7wt/uQzId97rufTeJKpdWz5qYT72/3JvWXn8+nx2OX0eHr3e3M9+7zrJzBY8lmjA6pREkw+k00wLfKVsxV6MluCSM08gpGD8djhDlZODTlvB9FjPB21RSV8JLHT0TP6ZGihn6YkAzg8MXJbWl3gzHLuUUnhGlvB8Sl7ZIQCOS7hMBKSC27QBxKsa1BCLisSTuj4j3Po1WEksR6JWiCjzh7FlEQVxHagqGJ4/FoQ8hBJTLM+JKES0XEZKzEHLI6G61sw02fHbZ+2obOWxC/ARwFVyxMttCjfo5oZGkbzoMm77JOvqhQRPBvGy1C/oW79Rrh4cJHEusRgMxLmDqzoOnQHjU15heIjVIv5Q03zJshJ59QSTyGOgqlSwZH4Y58Ygk2IsYiMuH7Qxz/NF5JQwQLLl2PxVtzT2PCociZumhdVvoj1UERMqIDRJJY8xBTuLD8u7GbhnXIh1+GGnxrE5bMp/GNA1oo5bCiAWw1wRIiecVU08yJUwuS0EN1bat3XF62+Faf41h0vGKtVqlZlVrRi7hrxQ4Qu5afdU36jADDIg7PkvLO1FSL28D5FYCSGmtimM5UxuK6IYEITpxuArFTS3LggSopxSFAv6eC/5957ef7w9zgZon6EsFCnLHQGhHscB1joaD4r5lGZYJvyWnUNr52/hPmqqV0WzyHBwYyPqS/v0qkVn5lKgGJKa5D4/DygcqGSVUuWz72aCqHgGLVTUQJKmpNJMaoJwWqd5ZBWGl+LuGsVYl/XS5uHksG01sLkRgi2JGp5fy0/x4P/o+0/3JE26SmvGQylNfIb7NoASdwjYnSOR4O/Fd2nv+J69mdPi2HbHgTHUh6LYJ1423WEJZRb6PQzWHLUSqM6r7vCVyDN1Mpo3/I5FyW79VGHTOmjCk9VkXgg/UZmhltF9Jc10cF4xCQ51mM6X9k+n/AJLIL2JcljQAH2hJ5UHijqE43DjWEiGEY3102R2cMk+O6mkx/36clZVC+WRA/RykY5jV0cAX7jjrZwGJhozvDYSErlFHVviWEzTKeWQNj00Q804++fQi1ydxDCvP0E36bE6dQrNRBvyGG7RuDy5eDz4rEEAKGZnNS23EGDfKMP3oJYXwBwQ2UeIoe76bzn7PXQXS1Ok0dMZwGif/VrtLIgVfeiPUXj+n/MvbiEPCdUyfnW7TBKkAkYdQVqT5TNXs+qcbWCLCh6BUbt3CnAebwzaTX9kDxsFaweaaibyByaysLKrmU1tuKmtVc9Xfjf8A) tests actually using a property table in a custom shader to visualize boreholes and modulate their radii based on metadata. The example tileset used has three metadata properties accessible via the shader: `drillingDiameter`, `fromDepth` and `toDepth` (describing the start and end point of each borehole segment to which a `drillingDiameter` metadata sample applies). 

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/Model/MetadataPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/MetadataPipelineStage.js
@@ -771,11 +771,7 @@ function addPropertyMetadataStatistics(shaderBuilder, propertyInfo) {
   );
 }
 
-function processPropertyTableProperty(
-  renderResources,
-  propertyInfo,
-  webgl2,
-) {
+function processPropertyTableProperty(renderResources, propertyInfo, webgl2) {
   addPropertyTablePropertyMetadata(renderResources, propertyInfo, webgl2);
   addPropertyMetadataClass(renderResources.shaderBuilder, propertyInfo);
   addPropertyMetadataStatistics(renderResources.shaderBuilder, propertyInfo);
--- a/packages/engine/Source/Scene/parseStructuralMetadata.js
+++ b/packages/engine/Source/Scene/parseStructuralMetadata.js
@@ -265,8 +265,12 @@ function collectGpuCompatiblePropertyBufferViews(
 function createNoDataBufferView(classProperty, numFeatures) {
   // noData can be a number, an array of numbers (e.g. for vecN types), or a nested array of numbers (e.g. for arrays of vecN types).
   let noData = classProperty.noData;
-  const metadataComponentCount = MetadataType.getComponentCount(classProperty.type);
-  const metadataArrayLength = classProperty.isArray ? classProperty.arrayLength : 1;
+  const metadataComponentCount = MetadataType.getComponentCount(
+    classProperty.type,
+  );
+  const metadataArrayLength = classProperty.isArray
+    ? classProperty.arrayLength
+    : 1;
 
   // Special case: noData enum values are specified as strings, so we need to convert them to numbers here.
   if (classProperty.type === MetadataType.ENUM) {
@@ -286,12 +290,15 @@ function createNoDataBufferView(classProperty, numFeatures) {
   const bytesPerElement = classProperty.bytesPerElement();
   const buffer = new ArrayBuffer(bytesPerElement * numFeatures);
   const view = new DataView(buffer);
-  const setter = MetadataComponentType.getDataViewSetter(view, classProperty.componentType);
+  const setter = MetadataComponentType.getDataViewSetter(
+    view,
+    classProperty.componentType,
+  );
 
   for (let i = 0; i < numFeatures; i++) {
     for (let j = 0; j < metadataArrayLength; j++) {
       for (let k = 0; k < metadataComponentCount; k++) {
-        setter((bytesPerElement * i) + j + k, noData[j][k]);
+        setter(bytesPerElement * i + j + k, noData[j][k]);
       }
     }
   }
--- a/packages/engine/Source/Scene/MetadataComponentType.js
+++ b/packages/engine/Source/Scene/MetadataComponentType.js
@@ -480,17 +480,25 @@ MetadataComponentType.getDataViewSetter = function (view, componentType) {
   const setters = {
     [MetadataComponentType.UINT8]: view.setUint8.bind(view),
     [MetadataComponentType.INT8]: view.setInt8.bind(view),
-    [MetadataComponentType.UINT16]: (offset, value) => view.setUint16(offset, value, true /* little-endian */),
-    [MetadataComponentType.INT16]: (offset, value) => view.setInt16(offset, value, true),
-    [MetadataComponentType.UINT32]: (offset, value) => view.setUint32(offset, value, true),
-    [MetadataComponentType.INT32]: (offset, value) => view.setInt32(offset, value, true),
-    [MetadataComponentType.FLOAT32]: (offset, value) => view.setFloat32(offset, value, true),
-    [MetadataComponentType.FLOAT64]: (offset, value) => view.setFloat64(offset, value, true),
-    [MetadataComponentType.UINT64]: (offset, value) => view.setBigUint64(offset, BigInt(value), true),
-    [MetadataComponentType.INT64]: (offset, value) => view.setBigInt64(offset, BigInt(value), true),
+    [MetadataComponentType.UINT16]: (offset, value) =>
+      view.setUint16(offset, value, true /* little-endian */),
+    [MetadataComponentType.INT16]: (offset, value) =>
+      view.setInt16(offset, value, true),
+    [MetadataComponentType.UINT32]: (offset, value) =>
+      view.setUint32(offset, value, true),
+    [MetadataComponentType.INT32]: (offset, value) =>
+      view.setInt32(offset, value, true),
+    [MetadataComponentType.FLOAT32]: (offset, value) =>
+      view.setFloat32(offset, value, true),
+    [MetadataComponentType.FLOAT64]: (offset, value) =>
+      view.setFloat64(offset, value, true),
+    [MetadataComponentType.UINT64]: (offset, value) =>
+      view.setBigUint64(offset, BigInt(value), true),
+    [MetadataComponentType.INT64]: (offset, value) =>
+      view.setBigInt64(offset, BigInt(value), true),
   };
 
   return setters[componentType];
-}
+};
 
 export default Object.freeze(MetadataComponentType);


packages/engine/Source/Scene/Model/MetadataPipelineStage.js
```diff
@@ -223,10 +239,135 @@ function getPropertyTextureInfo(propertyTexture, statistics) {
...
+    };
+  }
+
+  return infoArray;
+}
+
+/**
+ * Map property tables to feature ID sets for a given primitive. In general, the mapping is not 1:1, but
+ * within a given primitive it is safe to treat it as such.
+ *
```
mzschwartz5, line 330: This isn't always the case... see [this sample tileset that explicitly tests multiple feature ID sets for a single primitive.](https://github.com/CesiumGS/3d-tiles-samples/tree/main/glTF/EXT_structural_metadata/MultipleFeatureIdsAndProperties). (Which you can test via the sandcastle linked in the testing steps [on the last PR](https://github.com/CesiumGS/cesium/pull/13155))

However, if a single primitive has multiple feature ID sets, it's ambiguous which one we should use to sample the property table (indexed by (featureID, propertyID)) in the shader. Thus, we just use the first feature ID set we find and log a warning.

There may be a better way to do this in the future, like using an "active" feature ID set or a user-specified one. But this is also a really niche use-case that realistically we don't need to support.

packages/engine/Source/Scene/MetadataClassProperty.js
mzschwartz5, line 1: A lot of the changes in this file are just moving code from `PropertyTextureProperty.js` to this, lower-level class, so that it can be reused by property table properties, which are now also backed by textures.

There are _some_ changes to the moved functions as well, though. May be handy to paste the functions into vscode and run a diff or something.

packages/engine/Specs/Scene/Cesium3DTilesetSpec.js
```diff
@@ -4690,49 +4690,52 @@ describe(
       );
     });
 
-    it("clipping planes cull tiles completely inside clipping region for i3dm", function () {
-      return Cesium3DTilesTester.loadTileset(
+    it("clipping planes cull tiles completely inside clipping region for i3dm", async function () {
```
mzschwartz5, line 4693: This test became flaky after these changes. Not entirely sure what broke the camel's back, so to speak, but this should be more robust now - waiting for tiles to load before asserting anything.

packages/engine/Specs/Scene/MetadataClassPropertySpec.js
mzschwartz5, line 1: Again, the changes here are mainly just moving tests to go along with the code that was moved.

packages/engine/Source/Scene/Model/MetadataPipelineStage.js
javagl: Following up from the internal discussion: There is https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#featureIdLabel , and one could make a strong case that this, which currently says

> Label of the feature ID set to use for picking and styling.

should say

> Label of the feature ID set to use for picking and styling **and property table access**

or something like that. 

**But...** having no clue about how complicated this is (I haven't read the whole changes in all related PRs), this is just a hint, and may be tracked as a possible follow-up for the future.

mzschwartz5: Yes I did see that. I would need to look into how that setting would plumb down to the metadata pipeline stage, and how it should be made available for use in a shader (uniform?).

It might be very easy. It's just that, given the scope of these changes already, I think it might be best as a follow-up. I'll add it to the follow-ups listed in the description.

lilleyse:
I confirmed that this is working for instanced and non-instanced models :+1: 

image

[Sandcastle code](http://localhost:8080/Apps/Sandcastle2/index.html#c=jVJdb9owFP0rFk+pxAyBQdeOVuuATmENqEChRZFa49yAG8emsUOAqv99DgkV3fowP8T33nPuuR9xGxRLIuxIgX0ISML1FaWg1FiGINAF8kqw7S3nvygbsJ5zt3PsPnOUI4YN2naaTri6n7R7Z9iQXvxfoSE5m/5zuHU73Vqf2Xz2fJfejG9TN3L1YLzY9Ec2N7g9mzoNt+bW3WmX3bR7q5kRG4xSRqbXVedZbvo7mvZ3D+mgQ9PgFg/ntDNcqGl31pkMBuHL79vT5iaMHof3/OcDeYBaGk7cdcNvdL95pe+eoFIojdYMUojNEAJS1M4Hnexjlleie78thSZMQOyVTkyeJ3S8Ra+eQKhSQX0pvjAjRAQF/xzVzpqNar1egM4HoGmffs2AvLBmHBRoU5mkhOlD7fyqd8Y5jINYRmbxV8o4jm8V+lkfqOgdKwoC8CpmEdNsDQoT37cK+ZyYVyjoOymjsTwmZJTCxTRRWkajJfH/3kr7CLH24yMUxGQRgdB5dAwbfY6ecsi0J5n/znDNBq3rwnHEKtEoUPu7jJiQxqW76DGSPnCXaIgZ4SgqjJOD4uvBMJW5JBqRGIjpslDCEWjiE01wFi8Gyw8LkLUnt1Ctiqsnx1LovRD2WRAkCozkGmjdsnG1jKrFJ19lft4QcEP7L5F9vv2ZyMEsjKdydr+Zad8QJZoukQVxLOOi2ezZSA6Yy0URN1omtVQutZTecrjMWD9YtJKxRknMLYwrGqIVN42pyjyhYfZ/lcryWpVDSstna8T8i09eO6KcKGWQIOF8xHbglS5bFcP/kGZ+hM/EYrCGmJNtRlnalzd5EGPcqhj33ywtJZ+T+EjxDw)

lilleyse:
I'm testing with a 3D Tiles 1.0 asset (using b3dm and batch table) and the custom shader doesn't seem to pick up the properties. This might be due to it using batch table JSON instead of batch table binary.

image

[Sandcastle code](http://localhost:8080/Apps/Sandcastle2/index.html#c=nVRbT9swFP4rVl8oUnFaeOtCtQk0bdLQHuj2FGm4yUnrzZfIPikUxH/fcZzQhDKtW1T5ci6ffb7zubk1HtlWwj04dskM3LMr8LLW/HtjG2ejvNlfWYNCGnDZ6PRdZjKTN5koFXhAShX3QmKXHKeL62V089JZ/c0pQuM8od+t0JWCa4EiGYQmbcLLImkP4D+9Nd3R8brc52CAV05qiXILnouiGLfxIbANe7RWL23fkZkONa89Wn27EcXr8q96nvFTZhgrnVhrMBhtS3jAObsLDsa2VhYv/huiafyx3Xw2VY2s9M08YdJY2uaP+oe2BagbgeCkUEy3i9OI9xQnOlNZgWwDcr0JHLc4XAOKgsjjnxpPU1JMkCUbt+EpO5/y6ekejL0cwwtZlrUHgtxCfjGe8emETdshMBTDnxkoCjoCoMmdHQLERTPdTTLzTPWNJqPU407BIljfS11Zh6wmcZAuEEgXdIZPVnX+KzTI+wCXJl1KWsgtk8XlG7pkuRLek6eslbqVj5CNFmlC8YM0YrSQZv11C06JXQjZzBZfopFznia0PcxCa9VK0CHNrdPoDrV9cCjzmu4srZlHU+pBQY4dCd2t4n3PVjVikHLnDn08W0kTTrFVgPFzJnqofrLn/+ivRYoqPTFCw8n/wGyFqmHOYkFQ9IvtKiCyojsyk3TUpChWbZPDZmWL3aDW0joQ+YbAUazBt8zGYLfY3zXFYpCHTUmhoqa9WAxjF/0qUxmey7Bu3FVAME6YNey70MpbGnKdXRzYxQPZD8weoSI7SX722tW/cUtiXjtH/whHt6FJ+1YREiWfNJUc30OB6Ob9l/v3j4qfh0HqWv+TVoideRhC4rF5z0PCkjfa1nUqNJwet6cHTbtzWv6Z3GWM7cMNJEKbTlu0DJpsVduJtdPvbw)

javagl:
@lilleyse 

The same error appears on the currently depolyed sandcastle: https://sandcastle.cesium.com/index.html#c=jVLda9swEP9XDj85EORkfevcMEgZFFb6kGxPhlWVzsmt+giS7Kwp/d8nW0qarn0oGJ/ufh9nnU9Y4wP0hHt0cAUG97BET51mv8Za2RRizJfWBE4GXVNMvjamMWJUBlLoMUQp33MKR3EKF9frBLPWWf3TqejGWBWfFdc7hdc88OoNtcqC06HKDdgfb82xdfpc5gUaZDtHmgL16BmXssz8gZhpB2v12p4DjTm6is4Hq1dbLv+//vIMKZ8bA9A6vtFoQqqt8W+4hPsBAOgtyRN+G8dUfs/Jjdl1AVo/ximQsTEVB/1bW4nqlgd0xBXofJgkv+cUYk9leYAt0mY7zDj7MI2Byzg8lpDxSklALZSZXsOXGZtNXs3g1IZJatvOY7TsUVyUczabwiy/hgkl+gugiqRPGIza+XuDdBjD/bQxL/F+xbSofXhSuBiq30jvrAvQxeWIexEw7kXs4auHTjwOP8j7wa6ujpJaUg8krz7YSxCKex+RtlNqRQdsikVdRf4bWZyoJLO569Ep/jRQtvPFj1RkjNVVTN+rgrXqgbszx38 

And I checked: This is **not** caused by the fact that the property is called `Height` (and not `height`) 🙂

lilleyse:
> The same error appears on the currently depolyed sandcastle: https://sandcastle.cesium.com/index.html#c=jVLda9swEP9XDj85EORkfevcMEgZFFb6kGxPhlWVzsmt+giS7Kwp/d8nW0qarn0oGJ/ufh9nnU9Y4wP0hHt0cAUG97BET51mv8Za2RRizJfWBE4GXVNMvjamMWJUBlLoMUQp33MKR3EKF9frBLPWWf3TqejGWBWfFdc7hdc88OoNtcqC06HKDdgfb82xdfpc5gUaZDtHmgL16BmXssz8gZhpB2v12p4DjTm6is4Hq1dbLv+//vIMKZ8bA9A6vtFoQqqt8W+4hPsBAOgtyRN+G8dUfs/Jjdl1AVo/ximQsTEVB/1bW4nqlgd0xBXofJgkv+cUYk9leYAt0mY7zDj7MI2Byzg8lpDxSklALZSZXsOXGZtNXs3g1IZJatvOY7TsUVyUczabwiy/hgkl+gugiqRPGIza+XuDdBjD/bQxL/F+xbSofXhSuBiq30jvrAvQxeWIexEw7kXs4auHTjwOP8j7wa6ujpJaUg8krz7YSxCKex+RtlNqRQdsikVdRf4bWZyoJLO569Ep/jRQtvPFj1RkjNVVTN+rgrXqgbszx38 

I think that's expected until this PR is merged.

> And I checked: This is **not** caused by the fact that the property is called `Height` (and not `height`) 🙂 

I checked both as well because I wasn't sure about the property id sanitization rules. But it seems like it should be `Height` instead of `height`. I updated the sandcastle.

> ### Property ID Sanitization
> 
> GLSL only supports alphanumeric identifiers, i.e. identifiers that do not
> start with a number. Additionally, identifiers with consecutive
> underscores (`__`), as well as identifiers with the `gl_` prefix, are
> reserved in GLSL. To circumvent these limitations, the property IDs are
> modified as follows:
> 
> 1. Replace all sequences of non-alphanumeric characters with a single `_`.
> 2. Remove the reserved `gl_` prefix if present.
> 3. If the identifier begins with a digit (`[0-9]`), prefix with an `_`
>

javagl:
> I think that's expected until this PR is merged.

I wasn't _entirely_ sure about that - yesterday, I looked at https://github.com/CesiumGS/cesium/issues/10480 and wondered whehther there is anything _explicit_ about the compatibility of "old/legacy" things (like batch tables) with "new" things (like custom shaders). Specifically: Batch tables should work for styling, but I wasn't sure about custom shaders. (I know that there are some translations between them under the hood, but don't know all details)

> But it seems like it should be Height instead of height.

Right now, in the shader compliation error message, one can see that the metadata structs only contain `_empty`, so it's not an issue of case sensitivity (but I assume that when they are _not_ empty, it will have to be `Height`).

mzschwartz5:
I should add that it's not my intention to add any new features for the legacy metadata workflows. So if they didn't work with custom shaders before, they probably won't now. I only want to make sure I'm not actively breaking any workflows we still support.

mzschwartz5:
Some serious issues I've found during testing:

It seems that different primitives that reference the same property tables may order these properties differently or only list certain subsets of these properties. This is leading to several problems: 

- It means each primitive compiles a different (but very similar) shader based on which properties is uses. This might be causing the performance issues noted [here](https://github.com/CesiumGS/cesium/pull/13155#issuecomment-3881315941)
- Each primitive is uploading a texture containing only the properties it uses. This is wasteful as it implies overlap between the textures from various primitives.
- The different shader versions don't all list all metadata properties in the referenced table, which can cause compilation issues.
- And even the properties that do get included may not be in the expected order, which means sampling from the property table texture may give values for the wrong property.

Basically all of these issues can be solved by uploading a single texture per property table rather than per primitive. The devil is in the details, but the concept is simple.

Second overarching issue:

So far, I've only tested on gltfs that have a single property table (because DesignTiler merges property tables together by default). For certain use cases, we want to keep those tables un-merged. This causes custom shader compilation issues when a single glTF model has properties with the same name between different tables. We need to disambiguate somehow. Maybe rather than a single `(v|f)sInput.metadata` struct, we need one per metadata class? (either like `(v|f)sInput.myMetadataClass.propertyX` or `(v|f)sInput.metadata.myMetadataClass.propertyX`)

packages/engine/Source/Scene/parseStructuralMetadata.js
```diff
@@ -213,37 +213,63 @@ function collectGpuCompatiblePropertyBufferViews(
...
+ *
+ * @private
+ */
+function createNoDataBufferView(classProperty, numFeatures) {
+  const noDataValue = classProperty.noData;
+  const bufferView = new Uint8Array(
+    numFeatures * classProperty.bytesPerElement(),
+  );
+  for (let i = 0; i < numFeatures; i++) {
+    bufferView.set(noDataValue, i * classProperty.bytesPerElement());
```
lilleyse: I'm not sure if this is correct. It assumes the `noData` value has been packed into a `Uint8Array` which doesn't seem to be the case.

mzschwartz5: I'm not sure I understand. Isn't that what this code is doing? Packing the `noData` value into a `Uint8Array`, that is.

lilleyse: I don't think [set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set) works like that. It expects an array or typed array that it can copy elements from.

mzschwartz5: Good point. I just fixed that. 

This code also didn't handle the fact that `noData` isn't always a single number. It can be an array or an array of arrays too (depending on the `MetadataComponentyType`). 

My new approach is more robust.

mzschwartz5: Also - the new approach handles the fact that `noData` for enum types are specified as strings rather than the enum value.

lilleyse:
I was trying this sandcastle again and it seems like there might have been a regression.

https://github.com/CesiumGS/cesium/pull/13163#issuecomment-3880985762

image

lilleyse:
Great, looks better now. And this time I confirmed that the "area" in the info box actually matches the styling result

image

PR #13166, opened 2026-01-28
Remove `baseHref` from Sandcastle code urls, add `dompurify` sanitation

jjspace:
# Description

The new structure of sandcastle means we no longer need the `baseHref` that was encoded into some "share urls", specifically the ones for the Standalone page. This used to be needed for the nested structure of gallery demos. The newer way we build sandcastle is more sensitive to changes to the `<base>` of the standalone page and if a different base is included in the URL it could fail to load.

I also took this opportunity to quickly add `DOMPurify` to slightly sanitize the HTML included in sandcastle code before adding it to both the viewer bucket and the standalone page. This _shouldn't_ affect any sandcastles but may in very niche circumstances if they were doing something weird with scripts outside the Sandcastle JS 

## Issue number and link

No issue, cleanup that was going to come in #13154 but I wanted to prioritize before this release

## Testing plan

- Run/Load sandcastle locally or in CI
- Make sure a mix of demos load, specifically test some that make heavier use of custom html
- Make sure the standalone page loads correctly for all sandcastles

# Author checklist

### DIFF
--- a/packages/sandcastle/src/Helpers.ts
+++ b/packages/sandcastle/src/Helpers.ts
@@ -29,8 +29,6 @@ export function makeCompressedBase64String(data: SandcastleSaveData) {
   // data stored in the hash as:
   // Base64 encoded, raw DEFLATE compressed JSON array where index 0 is code, index 1 is html
   const { code, html } = data;
-  // historically there was sometimes a third element in this array that contained a baseHref
-  // caution should be taken if we add a 3rd element back
   const encode = [code, html];
   let jsonString = JSON.stringify(encode);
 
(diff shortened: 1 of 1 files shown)


jjspace:
> Perhaps you could be more specific with this comment historically there was sometimes a third element in this array that contained a baseHref caution should be taken if we add a 3rd element back

@lukemckinstry I put this here only as a warning for future devs (or me) in case someone wants to try embedding a 3rd element in the url that some old urls may still be floating around that don't line up with the format they expect. What more do you think can be added to help it be more clear?

packages/sandcastle/src/Helpers.ts
```diff
@@ -69,10 +70,11 @@ export function decodeBase64Data(base64String: string): SandcastleSaveData {
   // index 0 is code, index 1 is html
   const code = json[0];
   const html = json[1];
-  const baseHref = json[2];
+  // historically there was a third element in this array for some links
```
ggetz: My thought is to treat the third arg as a deprecated parameter. It should be fairly straightforward to check if 1) there is a third arg, 2) it's a string, and 3) we can parse a URL. Then we can log a warning in the console if it makes sense.

I think that would ensure better future-proofing than this in-code comment, as it's more transparent to the consumer and it'll be harder to miss if we need to implement a new third parameter.

ggetz:
Thanks @jjspace, and @lukemckinstry for your testing and review! I just had one comment on how we should handle backwards compatibility. Once that's addressed, this is good to go IMO.

packages/sandcastle/src/Helpers.ts
jjspace: I don't mind adding this as an actual warning. I believe we only care if the value exists _at all_ not whether it's a valid path. I left it as a broad check that it's not `undefined`

Updated

jjspace:
Thanks @ggetz and @lukemckinstry I believe this should be good to go

ggetz:
Thanks for the update @jjspace!

PR #13168, opened 2026-01-29
Types: Convert Cartesian2-4.js to ES6 classes

donmccurdy:
# Description

Converts Cartesian2, Cartesian3, and Cartesian3 to ES6 classes. The changes are very easy to make semi-automatically (see steps below) but not entirely trivial to review. So I'm inclined to do a few "small" PRs like this first and build up confidence, then attempt bigger batches.

**Steps**

1. Grep for unwanted `new Cartesian+\.[a-z]`, and remove 'new' in each case.
2. Grep to confirm CartesianN has no base or child classes. If it does, they need to be converted at the same time.
3. Using the script from #13125, run `node ./lebab-batch.js "packages/engine/Source/Core/Cartesian*.js"`. If the script outputs warnings, decide whether to skip these files or to manually resolve them, considering (1) above. In this PR, no errors were encountered.

**Reviewing**

Review is much easier with GitHub's split view, and whitespace changes hidden ([`?diff=split&w=1`](https://github.com/CesiumGS/cesium/pull/13168/changes?w=1)). Lebab shouldn't change the order of class members unnecessarily.

## Issue number and link

- #8359 

## Testing plan

Unit test coverage and ESLint should give us pretty good confidence on a mechanical change like this — but if there are other specific concerns let's take a look. On a visual check the API documentation looks identical.

# Author checklist

#### PR Dependency Tree

* **PR #13168** 👈

### DIFF
--- a/packages/engine/Source/Core/Cartesian2.js
+++ b/packages/engine/Source/Core/Cartesian2.js
@@ -1,21 +1,25 @@
+// @ts-check
+
 import Check from "./Check.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 import CesiumMath from "./Math.js";
 
+/** @import Cartesian3 from "./Cartesian3.js"; */
+/** @import Cartesian4 from "./Cartesian4.js"; */
+
 /**
  * A 2D Cartesian point.
- * @alias Cartesian2
- * @constructor
- *
- * @param {number} [x=0.0] The X component.
- * @param {number} [y=0.0] The Y component.
  *
  * @see Cartesian3
  * @see Cartesian4
  * @see Packable
  */
 class Cartesian2 {
+  /**
+   * @param {number} [x=0.0] The X component.
+   * @param {number} [y=0.0] The Y component.
+   */
   constructor(x, y) {
     /**
      * The X component.
@@ -74,10 +78,10 @@ class Cartesian2 {
    * Stores the provided instance into the provided array.
    *
    * @param {Cartesian2} value The value to pack.
-   * @param {number[]} array The array to pack into.
+   * @param {number[]|Float64Array|Float32Array} array The array to pack into.
    * @param {number} [startingIndex=0] The index into the array at which to start packing the elements.
    *
-   * @returns {number[]} The array that was packed into
+   * @returns {number[]|Float64Array|Float32Array} The array that was packed into
    */
   static pack(value, array, startingIndex) {
     //>>includeStart('debug', pragmas.debug);
@@ -96,7 +100,7 @@ class Cartesian2 {
   /**
    * Retrieves an instance from a packed array.
    *
-   * @param {number[]} array The packed array.
+   * @param {number[]|Float64Array|Float32Array} array The packed array.
    * @param {number} [startingIndex=0] The starting index of the element to be unpacked.
    * @param {Cartesian2} [result] The object into which to store the result.
    * @returns {Cartesian2} The modified result parameter or a new Cartesian2 instance if one was not provided.
@@ -120,8 +124,8 @@ class Cartesian2 {
    * Flattens an array of Cartesian2s into an array of components.
    *
    * @param {Cartesian2[]} array The array of cartesians to pack.
-   * @param {number[]} [result] The array onto which to store the result. If this is a typed array, it must have array.length * 2 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 2) elements.
-   * @returns {number[]} The packed array.
+   * @param {number[]|Float64Array|Float32Array} [result] The array onto which to store the result. If this is a typed array, it must have array.length * 2 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 2) elements.
+   * @returns {number[]|Float64Array|Float32Array} The packed array.
    */
   static packArray(array, result) {
     //>>includeStart('debug', pragmas.debug);
@@ -139,7 +143,7 @@ class Cartesian2 {
       );
       //>>includeEnd('debug');
     } else if (result.length !== resultLength) {
-      result.length = resultLength;
+      /** @type {number[]} */ (result).length = resultLength;
     }
 
     for (let i = 0; i < length; ++i) {
@@ -641,6 +645,9 @@ class Cartesian2 {
   }
 
   /**
+   * @param {Cartesian2} cartesian
+   * @param {number[]} array
+   * @param {number} offset
    * @private
    */
   static equalsArray(cartesian, array, offset) {
--- a/packages/engine/Source/Core/Cartesian3.js
+++ b/packages/engine/Source/Core/Cartesian3.js
@@ -1,22 +1,27 @@
+// @ts-check
+
 import Check from "./Check.js";
 import defined from "./defined.js";
 import DeveloperError from "./DeveloperError.js";
 import CesiumMath from "./Math.js";
 
+/** @import Cartesian4 from "./Cartesian4.js"; */
+/** @import Ellipsoid from "./Ellipsoid.js"; */
+/** @import Spherical from "./Spherical.js"; */
+
 /**
  * A 3D Cartesian point.
- * @alias Cartesian3
- * @constructor
- *
- * @param {number} [x=0.0] The X component.
- * @param {number} [y=0.0] The Y component.
- * @param {number} [z=0.0] The Z component.
  *
  * @see Cartesian2
  * @see Cartesian4
  * @see Packable
  */
 class Cartesian3 {
+  /**
+   * @param {number} [x=0.0] The X component.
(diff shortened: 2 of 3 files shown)


Hiwen:
Personally, I would recommend using a script to uniformly convert all js files, just like[ converting from AMD to ES6 syntax before.](https://cesium.com/blog/2019/10/31/cesiumjs-es6/)

I used Babel to parse the expression tree of JavaScript, then tried to convert it into an ES6 class. Currently, most of the conversion work has been completed, but there are still some issues that need to be addressed when running it.

donmccurdy:
Hi @Hiwen! This PR was generated with the script from #13125. Internally my script uses [lebab](https://github.com/lebab/lebab) and [espree](https://www.npmjs.com/package/espree). So far the script seems to do well — it fails for a few files (not in this PR) but at least alerts us to the files it cannot safely convert, so we can handle them manually.

Would you be interested in opening a PR or two helping to convert some remaining source files in `packages/engine/Source/**` 😇? Other conversion methods are definitely welcome, too. Preferably such PRs would have clear diffs (see comment above) and be small enough to review in an hour.

javagl:
> Grep to confirm CartesianN has no base or child classes. If it does, they need to be converted at the same time.

This suggests a certain "rigurousness"(?) of the language. There are no structures. There are no contracts. There is no meaning. JavaScript receives its meaning by execution. So maybe, in some ways, JavaScript is the origin of the phrase "It is what it is", because "duck typing" is one aspect of all that. And now there are adventurous things like https://github.com/CesiumGS/cesium/blob/becdc1ff15c72d142446e7e2351dfd96f5969caa/packages/engine/Source/Scene/ClippingPlane.js#L126 , and I think that this could be considered to be a "child class" of `Cartesian3` - at least, it is _declared_ to be a `Cartesian3` in the `normal` getter. I won't argue. I'm just wondering.

Hiwen:
> Hi @Hiwen! This PR was generated with the script from #13125. Internally my script uses [lebab](https://github.com/lebab/lebab) and [espree](https://www.npmjs.com/package/espree). So far the script seems to do well — it fails for a few files (not in this PR) but at least alerts us to the files it cannot safely convert, so we can handle them manually.
> 
> Would you be interested in opening a PR or two helping to convert some remaining source files in `packages/engine/Source/**` 😇? Other conversion methods are definitely welcome, too. Preferably such PRs would have clear diffs (see comment above) and be small enough to review in an hour.

@donmccurdy Thank you very much for your invitation. I still need to do some research to further verify the feasibility of batch conversion. However, if batch conversion is carried out, it will be a huge PR, and it will be very difficult to complete the review within an hour!

donmccurdy:
@javagl yes, that's fair. I'm aiming to provide a rule that's clear and legible enough to help us make progress safely and stay well clear of more time-consuming issues. Duck-typing is not a problem. But mixing ES6 class syntax with prototypal inheritance can be a problem, if a child class needs to invoke the constructor of a parent class, and only the parent is using ES6 class syntax. This came up for three.js in https://github.com/mrdoob/three.js/pull/17425, with symptoms as shown in this JSFiddle: https://jsfiddle.net/w2xfb6eg/. Not a hard-and-fast rule, but will reduce the number of things we need to worry about and test, if we can avoid mixed the two in the same inheritance chain.

@Hiwen thank you! Please note #13125 — it contains a bulk conversion of most of `packages/engine/Source/Core`, with a script given, and passes unit tests. The difficulty is the code review ... we're not prepared to review and merge this as a single large PR. Small PRs on related groups of files would be best, I think we can review and merge those quickly.

packages/engine/Source/Core/Cartesian2.js
```diff
@@ -15,726 +15,775 @@ import CesiumMath from "./Math.js";
  * @see Cartesian4
  * @see Packable
  */
-function Cartesian2(x, y) {
+class Cartesian2 {
+  constructor(x, y) {
```
jjspace, line 23: I actually just noticed one more thing. I believe the JSDoc for constructors needs to move to _inside_ the class. It seems like the docs and type generation pick it up like this but the actual TS server doesn't seem happy with it.

I also don't think we need an `@alias` of the same name? and probably not `@contsructor` either since it's an "actual" constructor now?

donmccurdy:
Hm, I'm not sure what the `@alias` tag's original purpose was but I can't see a difference without it after the ES6 Class conversion. I think we probably want to move the `@param` tags but put the description on the class rather than the constructor, a small change in documentation rendering?

Left: PR, in green. Right: Production, in red.

constructor_jsdoc

I've added `// @ts-check` as well since it's "mostly" straightforward in this case. There are some discrepancies in what the JSDoc types allow and what the method bodies implement, which I've resolved avoiding runtime changes. If those JSDoc changes require more in-depth discussion then I'd probably prefer to remove `// @ts-check` and go ahead with ES6 Class conversion without it. From a documentation rendering perspective, tsd-jsdoc seems to understand them:

pack_jsdoc

Allowing Float64Array and Float32Array but not other typed arrays is entirely arbitrary, but my hunch is that this can/should be loosened to allow any typed array type in the future. I didn't look, yet, into how to represent a union of all typed array types non-verbosely in a way that both TS and tsd-jsdoc would understand.

jjspace:
> Allowing Float64Array and Float32Array but not other typed arrays is entirely arbitrary, but my hunch is that this can/should be loosened to allow any typed array type in the future. I didn't look, yet, into how to represent a union of all typed array types non-verbosely in a way that both TS and tsd-jsdoc would understand

We actually just recently ran into this limitation in another PR. It's really annoying that `TypedArray` doesn't _actually_ exist so JSDoc won't accept it. https://github.com/CesiumGS/cesium/pull/12963#discussion_r2466865727 If you can figure out a way to do it that'd be awesome for the future, for now I'm fine with leaving it as just the 2 types it seems we already use it with

Thanks for the changes @donmccurdy

donmccurdy:
Thanks @jjspace! Attempting the TypedArray alias here:

- https://github.com/CesiumGS/cesium/pull/13187

Hiwen:
> @Hiwen thank you! Please note #13125 — it contains a bulk conversion of most of `packages/engine/Source/Core`, with a script given, and passes unit tests. The difficulty is the code review ... we're not prepared to review and merge this as a single large PR. Small PRs on related groups of files would be best, I think we can review and merge those quickly.

@donmccurdy  I tested your script and I think it works really well. Sorry I didn't notice it before.
I am very happy to participate in this process.

donmccurdy:
Just to have a better home for it than my scattered comments elsewhere, here's a short guide on how we're handling the es6 class conversions:

- https://github.com/CesiumGS/cesium/issues/8359#issuecomment-3893700288

packages/engine/Source/Core/Cartesian2.js
```diff
@@ -1,740 +1,796 @@
...
- */
-Cartesian2.fromElements = function (x, y, result) {
-  if (!defined(result)) {
-    return new Cartesian2(x, y);
+  /**
+   * Duplicates a Cartesian2 instance.
+   *
+   * @param {Cartesian2} cartesian The Cartesian to duplicate.
+   * @param {Cartesian2} [result] The object onto which to store the result.
+   * @returns {Cartesian2} The modified result parameter or a new Cartesian2 instance if one was not provided. (Returns undefined if cartesian is undefined)
```
zoran995, line 62: Should type here be extended to support undefined? Better would be to define overload type, but not sure if it is possible with jsdoc 
#13265

donmccurdy: Thanks @zoran995! Longer response in:

- https://github.com/CesiumGS/cesium/issues/13265#issuecomment-4005345276

javagl:
Just a small "heads up": 

There have been several TypeScript-related changes recently. 
(That's **good** and overdue. While the JSDoc type annotations already helped a lot for keeping things clean (at least on the lowest API level), they still have only been "hints", and have been plainly _wrong_ in (too) many cases). 
But (as I already mentioned in [another comment](https://github.com/CesiumGS/cesium/issues/8359#issuecomment-3900053355)): We might consider to add one validation step into this category of PRs. Namely, to compare the `cesium.d.ts` _before_ and _after_ the change.

Specifically: **This** change was "breaking" for the TypeScript layer. The context:
```
static doSomething() : number[] {
    ....
    const result = Cartesian3.pack(something, Array<number>(3));
    return result;
}
```
Given that the return type of `pack` has changed from `number[]` to `number[] | Float64Array | Float32Array` (or `number[] | TypedArray` in the meantime), this will no longer compile. 

Of course, it makes sense (and is 'necessary' on some level) to be able to `pack` into a typed array! 
And it's (probably) not a big deal for clients (will check more closely now). 
But maybe there is a way to identify (and point out) such changes.

I'll add another comment in https://github.com/CesiumGS/cesium/pull/13282 - it seems to be related enough...

donmccurdy:
@javagl proposed change:

- https://github.com/CesiumGS/cesium/pull/13302

Issue #10699, opened 2022-08-17
Precision errors when passing integer metadata values to fragment shader


javagl:
When metadata values that use integral types are supposed to be used in a custom shader, their values may not match the values that they have in the input data.

The effect was noticed while [creating a `3d-tiles-samples` model](https://github.com/CesiumGS/3d-tiles-samples/pull/55) that contains point clouds with property attributes, where each point has a `classification` value that can either be 0 or 1. 

When trying to access this `classification` value in a custom fragment shader as
```
        int classification = int(fsInput.metadata.classification);
        if (classification == 0) {
            color = vec3(0,0.5,0);
        }
        else if (classification == 1) {
            color = vec3(0.5,0.5,0.5);
        }
```
then the `classification` value is sometimes `0` for points where it should be `1`. 

With checks like

    if (fsInput.metadata.classification == 1.0)  ...
    if (fsInput.metadata.classification >= 0.999 ...

it can be noticed that the actual value is indeed only "an epsilon away" from the expected value of 1.0. 

---

Some points that might be relevant:

- The issue should not be caused by floating-point precision per se: A value of 0.0 and 1.0 can be represented exactly
- This is about point clouds, so it can not (directly) be related to any interpolation issues: There should be no interpolation.

@lilleyse found a workaround for the issue in https://github.com/CesiumGS/3d-tiles-samples/pull/55#issuecomment-1218332158 : When using the value only in the _vertex_ shader (and not in the fragment shader), then the correct value can be accessed.

lilleyse:
I'm running into this as well. I think you're right @javagl that it's due to perspective interpolation. Apparently this occurs even for points. One fix would be to use `flat` interpolation (or equivalent in WebGL 1) for point vertex attributes.

javagl:
I think this is related to https://github.com/CesiumGS/3d-tiles/issues/764 , but I haven't investigated the options here (beyond the considerations in that issue). 

> One fix would be to use `flat` interpolation 

It ould be, but I'm not sure whether **w** or **c**. Did you already try something out there? It was also mentioned in that issue, but I'd have to check where and how to actually apply that (with all the context about https://wikis.khronos.org/opengl/Primitive#Provoking_vertex - never stumbled over that before).

Issue #9935, opened 2021-11-18
ModelExperimental: interpolate per-vertex metadata


ptrgags:
According to the [`EXT_mesh_features` spec](https://github.com/CesiumGS/glTF/blob/proposal-EXT_mesh_features/extensions/2.0/Vendor/EXT_mesh_features/README.md#vertex-attribute), 

>**Implementation note:** For a primitive with feature ID attributes, points in the interior of a triangle or line segment should be considered to belong to the feature associated with the nearest vertex. 

Conceptually, that looks like this:

<img width="324" alt="Capture" src="https://user-images.githubusercontent.com/915398/77575695-4f2bfd00-6eaa-11ea-98e7-0ecaab27fce9.PNG">

If we want to implement this here in WebGL, it would take a couple extra attributes to do this in the fragment shader: (below describes the case for triangles, lines would be similar but only 2 vertices)

* barycentric coordinates - `vec3 barycentric = (a, b, c)` or a `vec2` given that `a + b + c = 1` by definition
* uninterpolated feature IDs - `vec3 (featureIdA, featureIdB, featureIdC)` _per feature ID set_. All three vertices would get all three feature IDs. So for the diagram above, all three vertices would receive the value `vec3(2, 4, 5)`. 

This feels a bit wasteful, but in the fragment shader there's no way to get the original vertex values (at least as far as I know?).

This is not needed for per-face metadata with `gl.TRIANGLES`  or per-edge for `gl.LINES` (as you can set all feature IDs to the same value), but in any other case you'd need the more general "nearest vertex" interpolation.

---

All this said, It would be best to see how often this is needed before implementing this part of the spec.

javagl:
A small example data set for basic tests:

[9935 per-vertex feature IDs.zip](https://github.com/CesiumGS/cesium/files/15295776/9935.per-vertex.feature.IDs.zip)

The archive contains

- A tileset with a single glTF that contains the feature IDs in a form that is required for this issue
- A sandcastle for basic tests
- Some information about the feature ID structure

The latter is just a summary of the feature IDs that are assigned to the vertices of the four quads:
```
        int featureIds[] =
        {
            0, 0, 0, 1, // Lower left
            2, 2, 3, 3, // Lower right
            4, 5, 5, 5, // Upper left
            6, 7, 8, 9, // Upper right
        };
```
The custom shader of the sandcastle just assigns colors (basically a binary representation of the IDs), as in
```
        if (id == 0) {
[...]

PR #13170, opened 2026-01-29
Use flat interpolation qualifier for point attributes

lilleyse:
# Description

Fixes https://github.com/CesiumGS/cesium/issues/10699

I was noticing precision errors when accessing vertex attributes in the fragment shader. In my case I'm writing a custom fragment shader to hide or show points based on a bitmask attribute and the default interpolation was corrupting the values.

https://github.com/CesiumGS/cesium/issues/10699 and https://github.com/CesiumGS/3d-tiles-samples/pull/55#issuecomment-1218332158 go into more detail for why might be happening:

> It seems like the attributes are being interpolated even though the mode is `gl.POINTS`. My theory is that the GPU is treating each point as a quad, duplicating the attributes at each corner, and running interpolation math. I wonder if the `flat` qualifier in WebGL 2 would fix this.

Adding the `flat` qualifier seems to work.

Before|After
--|--
before|after

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/10699

## Testing plan

Sandcastle code below. This is a tweaked version of the code in [PropertyAttributesPointCloud](https://github.com/CesiumGS/3d-tiles-samples/tree/main/glTF/EXT_structural_metadata/PropertyAttributesPointCloud) that does styling in the fragment shader.

```js

// Load the tileset from a local server
const tileset = viewer.scene.primitives.add(
  await Cesium.Cesium3DTileset.fromUrl(
    "http://localhost:8070/v1/static/PropertyAttributesPointCloud/tileset.json",
    {
    }
  )
);

// Move the tileset to a certain position on the globe,
// and zoom the viewer to the tileset
tileset.modelMatrix = Cesium.Transforms.eastNorthUpToFixedFrame(
  Cesium.Cartesian3.fromDegrees(-75.152325, 39.94704, 0)
);
const offset = new Cesium.HeadingPitchRange(
  Cesium.Math.toRadians(-22.5),
  Cesium.Math.toRadians(-22.5),
  32.0
);
viewer.zoomTo(tileset, offset);

tileset.customShader = new Cesium.CustomShader({
  fragmentShaderText: `
  void fragmentMain(FragmentInput fsInput, inout czm_modelMaterial material)
  {
      int classification = int(fsInput.attributes.classification);
      
      vec3 color = vec3(1);
      if (classification == 0) {
          color = vec3(0,0.5,0);
      }
      else if (classification == 1) {
          color = vec3(0.5,0.5,0.5);
      }

      material.diffuse = color;
  }
  `,
});
```

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,18 +8,23 @@
 
 - Fixed precision of point cloud attributes when accessed in a custom fragment shader. [#13170](https://github.com/CesiumGS/cesium/pull/13170)
 
+#### Additions :tada:
+
+- Added `AttributeCompression.encodeRGB8` and `decodeRGB8` for packing colors [#13174](https://github.com/CesiumGS/cesium/pull/13174)
+
 ## 1.138 - 2026-02-02
 
 ### @cesium/engine
 
 #### Fixes :wrench:
 
-- Fixed error with `DynamicEnvironmentMapManager` when `ContextLimits.maximumCubeMapSize` is zero.
+- Fixed jitter artifacts on Intel Arc GPUs. [#12879](https://github.com/CesiumGS/cesium/issues/12879)
 - Improved voxel memory usage by reworking `Megatexture` to use `Texture3D`. [#12570](https://github.com/CesiumGS/cesium/issues/12570)
-- Fixes jitter artifacts on Intel Arc GPUs [#12879](https://github.com/CesiumGS/cesium/issues/12879)
-- Fixed label sizing for some fonts and characters [#9767](https://github.com/CesiumGS/cesium/issues/9767)
-- Fixed a type error when accessing the ellipsoid of a viewer [#13123](https://github.com/CesiumGS/cesium/pull/13123)
-- Fixed a bug where entities have not been clustered correctly [#13064](https://github.com/CesiumGS/cesium/pull/13064)
+- Fixed multiple issues causing undefined pick results in 2D/CV scene modes. [#13083](https://github.com/CesiumGS/cesium/issues/13083)
+- Fixed label sizing for some fonts and characters. [#9767](https://github.com/CesiumGS/cesium/issues/9767)
+- Fixed a type error when accessing the ellipsoid of a viewer. [#13123](https://github.com/CesiumGS/cesium/pull/13123)
+- Fixed a bug where entities have not been clustered correctly. [#13064](https://github.com/CesiumGS/cesium/pull/13064)
+- Fixed error with `DynamicEnvironmentMapManager` when `ContextLimits.maximumCubeMapSize` is zero. [#12606](https://github.com/CesiumGS/cesium/pull/12606)
 
 ## 1.137 - 2026-01-05
 
--- a/Documentation/Contributors/ReleaseGuide/README.md
+++ b/Documentation/Contributors/ReleaseGuide/README.md
@@ -1,4 +1,4 @@
-# Release Guide
+# CesiumJS Release Guide
 
 We release CesiumJS on the first work day of every month.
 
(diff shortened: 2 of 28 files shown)


packages/engine/Source/Scene/Model/GeometryPipelineStage.js
```diff
@@ -174,13 +174,22 @@ GeometryPipelineStage.process = function (
       index = renderResources.attributeIndex++;
     }
 
+    // Use flat interpolation for point primitive vertex attributes (only supported in WebGL 2)
```
lilleyse, line 177: `flat` is only supported in WebGL 2 and I think it would be too much effort to get this working in WebGL 1.

packages/engine/Specs/Scene/Model/GeometryPipelineStageSpec.js
```diff
@@ -1198,6 +1198,37 @@ describe(
...
+
+        ShaderBuilderTester.expectHasVaryings(
+          shaderBuilder,
+          [
+            "vec3 v_positionEC;",
+            "vec3 v_positionMC;",
+            "vec3 v_positionWC;",
+            "float v_featureId_0;",
+          ],
+          ["", "", "", "flat"],
```
javagl: I'm not an expert with Jasmine. As a result, if I had to implement things like `expectHasVaryings`, I'd implement them  ~"differently" - maybe "more pragmatically"(and I'd be open for feedback and considerations about what is "better" or "worse" in one way or another). 

That said: This test will also pass when passing in 
`["", "", "flat", ""],`
here. I think that this should not be the case...

lilleyse: Thanks for catching that... that's what I get for not understanding the Jasmine code. I decided to go with a simpler approach.

lilleyse:
I merged in main. @javagl would you mind doing another review?

javagl (approved):
My first inlined comment wasn't really the result of a "review" - just the result of a "What is that about?", followed by a moment of confusion about Jasmine...

---

The main issue seems to be resolved. Comparing the current `main` state with this branch:

- Currently deployed Sandcastle: https://sandcastle.cesium.com/index.html#c=jVTbbhMxEP0VK08baWunSSNESSqgpYBEoaIB8bBS66xnswavJ7K9SVvUf2e8l7YpFxFtNL7MnBmfOXaO1ge20bAFx+bMwpYdg9d1xb82a0k2yJv5MdogtQWXDYYvMptZIdgHlIqFEljQBjwEVjismGQGc2mYB7chd5s3KXqXeZeM+xws8LXTlQ56A55LpZLMMia3Uoe+itZMThZtOI8ZvjjTODKWDcoQ1v5QCCe3fKVDWS9rykspA9jAc6xEi/D2QkzUXlPEnpfVmqxwUHhRglReVHQ0sTKLU/Hm2+LSB1fnoXbSXFYQpJJBinOHa3Dh5lUITi/rAP4ctQ3HBmslusPx7x5tNkjb4n62hjEFy3p1UeL2NdZWabv6iqau4JBRGuic76IZZvae2zPcwA63AYnZnEqgUtkaPbGGltEXnVYGlxGKAqVV7BapD3G9ayzFPoLKbF9vhQrMmaQTXVNjOsoXTlpfoKs8B+nDR3Sh/LJe4Km+BnXqZAUN+32DpCMutLSTpjcnsHIAPtl7NuX70/FkPE3Z5Dl/fvBsdJCyUXfCVhJYFK0iHqnuHbWDGDrXIS8/S7vayUWVljzgZ3KhEpO98ZhPh+l/OUzGfNTm7uQXKVpg0jGRdsW09Pf05LUPWF2UUj29HMePdpKm0YWTq4o0164t4Docsqu4sUGt7nfPqHnJaTd5b9c13Rnf2JRpizTNb6vLvi3gNN2jqhsMI9q9qEh7LDfSe13oXDZamMfFpMPj8l6ofNcvnrHF6O0G8gnL0WA8ZZwk+w9OumDJ00Rz6uRDKfG3Ez1KR3yajh4w7voBGA9/Qdz/J2LEa/87qP2w54grXRT0AlBcE9/4NtmvSAZ3wxeDdDDz4cbAUVx8qas1yZvV9KJwLgLQw0BIXizr/EcUgPcx3Uz0ITOlN0yr+R9exbYbtFPUxlzoW8gGRzNB/jthBht9f6KX0cib6FLuH31oFznnM0HT36MCollK9wjxFw

- CI build of this branch: https://ci-builds.cesium.com/cesium/flat-interpolation/Apps/Sandcastle2/index.html#c=jVTbbhMxEP0VK08baWunSSNESSqgpYBEoaIB8bBS66xnswavJ7K9SVvUf2e8l7YpFxFtNL7MnBmfOXaO1ge20bAFx+bMwpYdg9d1xb82a0k2yJv5MdogtQWXDYYvMptZIdgHlIqFEljQBjwEVjismGQGc2mYB7chd5s3KXqXeZeM+xws8LXTlQ56A55LpZLMMia3Uoe+itZMThZtOI8ZvjjTODKWDcoQ1v5QCCe3fKVDWS9rykspA9jAc6xEi/D2QkzUXlPEnpfVmqxwUHhRglReVHQ0sTKLU/Hm2+LSB1fnoXbSXFYQpJJBinOHa3Dh5lUITi/rAP4ctQ3HBmslusPx7x5tNkjb4n62hjEFy3p1UeL2NdZWabv6iqau4JBRGuic76IZZvae2zPcwA63AYnZnEqgUtkaPbGGltEXnVYGlxGKAqVV7BapD3G9ayzFPoLKbF9vhQrMmaQTXVNjOsoXTlpfoKs8B+nDR3Sh/LJe4Km+BnXqZAUN+32DpCMutLSTpjcnsHIAPtl7NuX70/FkPE3Z5Dl/fvBsdJCyUXfCVhJYFK0iHqnuHbWDGDrXIS8/S7vayUWVljzgZ3KhEpO98ZhPh+l/OUzGfNTm7uQXKVpg0jGRdsW09Pf05LUPWF2UUj29HMePdpKm0YWTq4o0164t4Docsqu4sUGt7nfPqHnJaTd5b9c13Rnf2JRpizTNb6vLvi3gNN2jqhsMI9q9qEh7LDfSe13oXDZamMfFpMPj8l6ofNcvnrHF6O0G8gnL0WA8ZZwk+w9OumDJ00Rz6uRDKfG3Ez1KR3yajh4w7voBGA9/Qdz/J2LEa/87qP2w54grXRT0AlBcE9/4NtmvSAZ3wxeDdDDz4cbAUVx8qas1yZvV9KJwLgLQw0BIXizr/EcUgPcx3Uz0ITOlN0yr+R9exbYbtFPUxlzoW8gGRzNB/jthBht9f6KX0cib6FLuH31oFznnM0HT36MCollK9wjxFw

Or as a short screencap:

Image

I looked at the actual change, and it seems reasonable as well.

EDIT: I also ran all affected specs locally, and they still pass. (This may be independent of CI, for what it's worth...)


The change in the `ShaderBuilderTester` looks like it could make things clearer and more strict at the same time. It's a bit more code in all the affected specs, but should be justified.

---

Somewhat unrelated to this PR: 

I think that it could be worthwhile to check whether something like this could also be done for the non-`POINTS`-case, in the context of https://github.com/CesiumGS/3d-tiles/issues/764 https://github.com/CesiumGS/cesium/issues/9935#issuecomment-2107707398

lilleyse:
> I think that it could be worthwhile to check whether something like this could also be done for the non-POINTS-case, in the context of https://github.com/CesiumGS/3d-tiles/issues/764 https://github.com/CesiumGS/cesium/issues/9935#issuecomment-2107707398

That crossed my mind as well... technically feature ids are still being smooth interpolated but it works because (1) we assume each triangle has the same feature id at all vertices (which I think should be required by the spec) and (2) we round the value in the frag shader.

PR #13175, opened 2026-01-30
Sandcastle standalone wrapper

jjspace:
# Description

Currently the standalone page is a pure html page that directly loads and runs Sandcastle code. This PR changes it to a "wrapper application" like the main Sandcastle app utilizing an iframe and the _same_ viewer/bucket to run code with more isolation.

The structure of our React components actually made this fairly straightforward which was really nice. I did have to break out a few more hooks and components for easier import into the wrapper and the main app.

This also enables us to embed the same Console Mirror for messages/warnings. In the future we may add a header with links back to the editor or other features and this new structure makes that easier to do.

Standalone URLs will now also work with `?id=[slug]` type urls instead of always needing to convert existing demos into a share code URL. If a user modifies an existing demo (making it "dirty") then we will still generate the share code URL to contain their changes.
- [x] For now I've kept a `standalone-old.html` with the old version for comparison, this _will_ be removed before this is merged

This PR builds on the work in #13154.

## Issue number and link

No issue, part of work prepping for #13079

## Testing plan

- Test that existing standalone links continue to work
- Test that clicking standalone for existing demos loads with the `id=` url
    - Test that if you alter the demo yourself it still contains those changes

# Author checklist

### DIFF
--- a/packages/sandcastle/src/util/usePageTitle.ts
+++ b/packages/sandcastle/src/util/usePageTitle.ts
@@ -1,5 +1,9 @@
 import { useEffect, useState } from "react";
 
+/**
+ * Hook to easily update the page title based on the specific Sandcastle that's loaded.
+ * Handles adding an asterisk * when the code state is dirty
+ */
 export function usePageTitle() {
   const [title, setPageTitle] = useState("New Sandcastle");
   const [isDirty, setIsDirty] = useState(false);
--- a/packages/sandcastle/src/ViewerConsoleStack.tsx
+++ b/packages/sandcastle/src/ViewerConsoleStack.tsx
@@ -12,6 +12,14 @@ export type ViewerConsoleStackRef = {
   toggleExpanded: () => void;
 };
 
+/**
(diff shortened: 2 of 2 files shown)


jjspace:
@lukemckinstry would you be able to take a look at this one too? It's more or less a direct follow up to #13154 which you reviewed.

packages/sandcastle/src/util/usePageTitle.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from "react";
+
```
lukemckinstry, line 2: I see page title updates for a sandcastle title that has not been altered, but not for one where sandcastle code has changes (ie. when the share url is used instead of ?id=[stub]) 
Is that the intent? Not a huge deal but wanted to point it out in case it was not.

jjspace: Can you elaborate on what you mean or steps to reproduce? The title should show the `*` whenever you edit the code regardless of the url type.

PR #13176, opened 2026-01-30
Lazy load Sandcastle gallery images

jjspace:
# Description

Somehow I completely missed that `` tags have a browser native option for lazy loading and it's as simple as [`loading="lazy"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#loading). :facepalm: 

This PR:

- Sets `loading="lazy"` for the gallery cards
- Sets a concrete height/width for gallery images
    - This helps the browser know how tall gallery cards should be _before_ the image loads so the scrollbar doesn't change size as you go down the list
- Make sure that the actual images `cover` the image area in case they're not the right size
- Defines a concrete size in our readme for how big thumbnails should be. The _vast_ majority of our thumbnails are already `225x150` so I just picked that

## Issue number and link

No issue but helps with any gallery list situations, like when searching in #13090 

## Testing plan

- Load sandcastle locally or in CI
- Open DevTools to the network tab (pick only image requests if you want)
- Refresh the gallery page
    - Observe that initial load has less images than current sandcastle
- Scroll the gallery page, I suggest removing the Showcases filter to get a longer list
    - Observe that new image requests happen slowly as you scroll
- Inspect an image and set `display: none` or delete it from the page and make sure the card still looks ok. This may happen when images haven't loaded yet or fail to load

# Author checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "mkdirp": "^3.0.1",
     "node-fetch": "^3.2.10",
     "open": "^11.0.0",
-    "prettier": "3.7.4",
+    "prettier": "3.8.1",
     "prismjs": "^1.28.0",
     "rimraf": "^6.0.1",
     "tsd-jsdoc": "^2.5.0",
--- a/packages/sandcastle/package.json
+++ b/packages/sandcastle/package.json
@@ -21,7 +21,7 @@
     "dompurify": "^3.3.1",
     "monaco-editor": "^0.52.2",
     "pako": "^2.1.0",
-    "prettier": "^3.7.4",
+    "prettier": "3.8.1",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "react-stay-scrolled": "^9.0.0",


jjspace:
@ggetz This isn't necessarily `Priority next release` but a pretty minimal encapsulated change that could be nice to get in if you feel you have the bandwidth

javagl:
Filtering the request list in the Network tab by `thumbnail.jpg` allows to check this quickly.

When visiting the current sandcastle, https://sandcastle.cesium.com/index.html , it sends out >200 requests for thumbnails
When visiting the CI build of this branch, https://ci-builds.cesium.com/cesium/lazy-gallery-images/Apps/Sandcastle2/index.html , it sends out 22 requests for thumbnails (may vary)

When selecting "Throttling Preset '3G'" in the Network tab and scrolling down the list, one can see the images slowly coming in.

---

Remotely related: The largest thumbnail is currently this one: https://github.com/CesiumGS/cesium/blob/becdc1ff15c72d142446e7e2351dfd96f5969caa/packages/sandcastle/gallery/aec-clipping/thumbnail.jpg 

It is saved as a JPG with 95% quality (!), and a size of 111KB.

Here, the quality is reduced:

Cessium Thumbnail Size

A bit is lost in GIF, but the point is: One can hardly see a difference when the quality is 34, with a size of 30KB.

Fun fact: That GIF has 3.3 MB 😁

ggetz:
Nice! Thanks @jjspace!

packages/sandcastle/README.md
```diff
@@ -61,7 +61,9 @@ development: false
 
 ### Thumbnails
 
-Thumbnails should be any image that represents what the sandcastle does. Often this will just be the Viewer with or without any Sandcastle interaction buttons. Thumbnail files should be limited in size to help save on bandwidth. Currently most are around 225px in width.
+Thumbnails should be any image that represents what the sandcastle does. Often this will just be the Viewer with or without any Sandcastle interaction buttons.
+
+Thumbnail files should be limited in size to help save on bandwidth. For consistency we have chosen the dimensions `225x150px` for our images.
```
ggetz, line 66: @jjspace To follow up on @javagl's comment, I think we can add some recommendations around file format and compression.

That can be a follow-up though given this PR is already an improvement. :+1:

jjspace:
I have discovered that in chrome this lazy loading doesn't seem to apply if the window is in the background. For example if I open devtools, switch to a different window and then refresh through devtools it seems to load all thumbnails again. I'm not sure if this is something we're doing wrong with the filter/sorting or if chrome is unable to optimize for the viewport or what.

Likely rather niche but maybe worth looking into in the future.

PR #13178, opened 2026-02-02
Create partial reference implementation for BENTLEY_materials_planar_fill (behind)

### LINKED ISSUE #12890: Create reference implementation for proposed glTF extension BENTLEY_materials_planar_fill
### Feature

Depends on https://github.com/CesiumGS/glTF/pull/90.

Within CesiumJS, consider creating a reference implementation of the proposed glTF extension BENTLEY_materials_planar_fill as it is described at the above link.

This item should be completed after the above-linked PR is more fleshed out / fully proposed.

markschlosseratbentley:
# Description

This PR adds support for the [`BENTLEY_materials_planar_fill`](https://github.com/CesiumGS/glTF/tree/vendor-extensions/extensions/2.0/Vendor/BENTLEY_materials_planar_fill) glTF extension, enabling CAD-style planar polygon fill rendering with proper depth sorting and configurable fill behavior including background color masking and coplanar geometry ordering. Note: The `wireframeFill` property is currently a no-op as CesiumJS does not yet have a proper wireframe rendering mode. See [this wireframe rendering mode draft PR](https://github.com/CesiumGS/cesium/pull/13192) for a possible way forward.

image

## Issue number and link

Partially fixes https://github.com/CesiumGS/cesium/issues/12890

## Testing plan

- Various unit tests have been added to CesiumJS related to this feature.
- Otherwise, changes have been verified with the comprehensive test case added in the gltf-planar-fill-dev Sandcastle sample.
- Test these changes using [the planar fill Sandcastle sample deployed right here](https://ci-builds.cesium.com/cesium/BENTLEY_materials_planar_fill/Apps/Sandcastle2/index.html?id=gltf-planar-fill-dev).

# Author checklist


danielzhong:
Great work! I did a quick review of the glTF loading and the feature ID comparison logic. The approach of introducing a new pass and comparing feature IDs to determine what’s behind makes sense to me. I haven’t run it locally yet, but the overall idea and implementation look solid and follow a similar path to how edge visibility was handled.

However, While testing with some cases I generated from the tileset publisher encoding work (PR: https://github.com/iTwin/imodel-native-internal/pull/950), I noticed the results weren’t behaving as expected. The results still didn’t match the rendering in the iTwinJS PR (https://github.com/iTwin/itwinjs-core/pull/8906). Note: I temporarily commented out the edge visibility code locally since support for “line string” hasn’t been merged yet.

For example, the “behind” flag in following test cases do not appear to be applied correctly:
[PlanarFillBehind.zip](https://github.com/user-attachments/files/26076338/PlanarFillBehind.zip)


in iTwinJS:
image
image

But in the current branch in CesiumJS, both files appear as if the “behind” flag is not being read at all.
image

I have a question about the yellow color in this example. If both flags are enabled, my understanding is that it should either render as the black background or appear behind the other two colors. But it doesn’t seem to match either case here(it's rendering yellow and z fighting (meaning it's not behind anything)). 

There’s also a purple test case with the same flags enabled compare with the yellow color test case I mentioned. The only difference is wireframe mode (which maybe shouldn’t affect the result), but the output looks different. The purple I highlighted isn’t visible—possibly rendering as black instead?

Am I misunderstanding how this is supposed to work?
image

Additionally, I’ve verified that the tileset publisher encoding behaves as expected in iTwinJS for planar fill. It might be a good idea to also use this renderer to validate the tileset publisher encoding, so we can confirm that both the tileset publisher and this implementation are working correctly. We can also add more test cases if needed.

markschlosseratbentley:
> For example, the “behind” flag in following test cases do not appear to be applied correctly

Investigating.

> I have a question about the yellow color in this example. If both flags are enabled, my understanding is that it should either render as the black background or appear behind the other two colors. But it doesn’t seem to match either case here(it's rendering yellow and z fighting (meaning it's not behind anything)).
> 
> There’s also a purple test case with the same flags enabled compare with the yellow color test case I mentioned. The only difference is wireframe mode (which maybe shouldn’t affect the result), but the output looks different. The purple I highlighted isn’t visible—possibly rendering as black instead?
> 
> Am I misunderstanding how this is supposed to work?

Some of the colors in the test case are a fair bit similar - sorry about that.  The "yellow" color you point out is actually slightly yellow with a tint of green and this is what it corresponds to, which appears to work as expected to me:

image

Regarding the purple test case: you are correct, the deep purple test case is indeed not visible. Since it has `bgFill` set to true, it will render as the background color (black). This is expected.

> Additionally, I’ve verified that the tileset publisher encoding behaves as expected in iTwinJS for planar fill. It might be a good idea to also use this renderer to validate the tileset publisher encoding, so we can confirm that both the tileset publisher and this implementation are working correctly. We can also add more test cases if needed.

100% agree, and this is on my list but doesn't need to block PR review. I want to get a good "infrastructure" example converted through the tiler using this extension (and your tiler branch for planar fill). The one I have included with this PR is good for testing.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -170,13 +170,100 @@ MaterialPipelineStage.process = function (
   const disablePointCloudNormals =
     defined(model.pointCloudShading) && !model.pointCloudShading.normalShading;
 
+  if (defined(material.planarFill)) {
```
mzschwartz5, line 173: Could you extract this into a function for better organization? (As seems to be the precedent anyway in `MaterialPipelineStage.process`). Side benefit- that big block comment can just become the function docs

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -170,13 +170,100 @@ MaterialPipelineStage.process = function (
...
+      ShaderDestination.FRAGMENT,
+    );
+
+    if (hasBehind) {
+      // Enable shader code that samples the planar fill ID texture.
+      const hasFeatureIds =
+        defined(primitive.featureIds) && primitive.featureIds.length > 0;
+
+      if (hasFeatureIds) {
+        const featureIdMember = primitive.featureIds[0].positionalLabel;
```
mzschwartz5: Is there something in the planar fill spec that guarantees a primitive has one (and only one) feature ID?

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -170,13 +170,100 @@ MaterialPipelineStage.process = function (
...
+        shaderBuilder.addDefine(
+          "HAS_PLANAR_FILL_BEHIND",
+          undefined,
+          ShaderDestination.FRAGMENT,
+        );
+        shaderBuilder.addDefine(
+          "PLANAR_FILL_FEATURE_ID",
+          featureIdMember,
+          ShaderDestination.FRAGMENT,
+        );
```
mzschwartz5: Since this logic is needed in both branches, it can be hoisted above both.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -170,13 +170,100 @@ MaterialPipelineStage.process = function (
...
+      }
+    } else {
+      // Non-behind planar fill geometry participates in the feature-ID
+      // pre-pass so that behind fills can test same-object coplanarity.
+      const hasFeatureIds =
+        defined(primitive.featureIds) && primitive.featureIds.length > 0;
+
+      if (hasFeatureIds) {
+        // Mark this primitive for the pre-pass derived command.
+        renderResources.planarFillIdPass = true;
```
mzschwartz5: Not sure I understand why only this primitives _not_ marked "hasBehind"  get this flag set to true. Your comment above implies to me that both behind and non-behind planar-fill geometry participate in the feature-ID pre-pass. 

Or is it only non-behind geometry that populates the feature ID texture? (I suppose that could make sense... "behind" geometry only cares about / needs to check "non-behind" geometry.)

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -170,13 +170,100 @@ MaterialPipelineStage.process = function (
...
+        // Runtime uniform to distinguish the pre-pass from the main pass
+        // (same shader program serves both commands).
+        shaderBuilder.addUniform(
+          "bool",
+          "u_isPlanarFillIdPass",
+          ShaderDestination.FRAGMENT,
+        );
+        uniformMap.u_isPlanarFillIdPass = function () {
+          return false;
+        };
```
mzschwartz5: A comment here that this is overridden by the derived command could be helpful

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -170,13 +170,100 @@ MaterialPipelineStage.process = function (
...
+
   // Classification models will be rendered as unlit.
   const lightingOptions = renderResources.lightingOptions;
   if (
     material.unlit ||
     !hasNormals ||
     hasClassification ||
-    disablePointCloudNormals
+    disablePointCloudNormals ||
+    hasBackgroundFill
```
mzschwartz5, line 194: Tangent rant (not about your work): this pattern pops up all over CesiumJS and it bothers me. `If (complex conditional that checks for a bunch of seemingly unrelated things) ... `

I'm not sure I can articulate what exactly the fundamental issue is or how to address it. You have `MaterialPipelineStage` reaching out to collect various unrelated pieces of state and set a policy about how they affect lighting. It just feels like that doesn't belong here. There should be a dedicated owner for this policy, and the dependencies should be inverted: rather than the policy-owner reaching out to collect various pieces of state, the state-owners should reach out to disable lit lighting. 

That reduces a complex if statement to a single boolean flag on an object that is either true (by default) or has been set to false at least once. And the policy object never needs to know _what_ (ever-growing conditions) contribute to the decision, only what the decision is.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
```diff
@@ -194,6 +281,23 @@ MaterialPipelineStage.process = function (
...
 
+  // Configure background fill for BENTLEY_materials_planar_fill extension.
+  // When backgroundFill is true, the fill is rendered using the view's background color
+  // to create an invisible masking polygon. Lighting is forced to unlit above.
+  //
+  // NOTE: The wireframeFill property from BENTLEY_materials_planar_fill is intentionally
+  // not handled here. CesiumJS does not yet have a proper wireframe rendering mode
+  // (debugWireframe is not a true wireframe mode), so wireframeFill is loaded as a NO-OP.
+  // When a proper wireframe mode is added to CesiumJS, wireframeFill support should be
+  // implemented here to control fill visibility in that mode.
```
mzschwartz5: Maybe we should link an issue here so future readers can check if this comment can be removed, if we forget ourselves.

packages/engine/Source/Scene/Model/ModelDrawCommand.js
```diff
@@ -79,6 +79,10 @@ function ModelDrawCommand(options) {
 
   const needsEdgeCommands = defined(renderResources.edgeGeometry);
 
+  // Non-behind planar fill primitives write their feature IDs in a pre-pass
+  // so that behind fills can test same-object coplanarity.
+  const needsPlanarFillIdCommand = renderResources.planarFillIdPass === true;
```
mzschwartz5: I think this is redundant? First of all, couldn't you just do:
`const needsPlanarFillIdCommand = renderResources.planarFillIdPass;`

Second, we only use `needsPlanarFillIdCommand` in one place after this, so why not just simplify to
`this._needsPlanarFillIdCommand = renderResources.planarFillIdPass`

packages/engine/Source/Scene/Model/ModelDrawCommand.js
```diff
@@ -855,6 +897,43 @@ function deriveEdgeCommand(command, renderResources) {
...
+ *
+ * No polygon offset is applied: the pre-pass uses the natural depth of the
+ * non-behind planar fill geometry (same program as the main command minus
+ * the POLYGON_OFFSET path that would shift it).
+ *
+ * @param {DrawCommand} command The original draw command.
+ * @returns {DrawCommand} The derived command for the feature-ID pass.
+ * @private
+ */
+function derivePlanarFillIdCommand(command) {
```
mzschwartz5, line 918: How frequently does this function get called?

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -1753,6 +1754,33 @@ function loadClearcoat(loader, clearcoatInfo, frameState) {
...
+  const {
+    wireframeFill = PlanarFill.DEFAULT_WIREFRAME_FILL,
+    backgroundFill = false,
+    behind = false,
+  } = planarFillInfo;
+
+  const planarFill = new PlanarFill();
+  planarFill.wireframeFill = wireframeFill;
+  planarFill.backgroundFill = backgroundFill;
+  planarFill.behind = behind;
```
mzschwartz5: Any reason for not just giving `PlanarFill` a constructor that takes and deconstructs a `PlanarFillInfo` object?

side note- could we make a jsdoc typedef for the input object here?

packages/engine/Source/Scene/PlanarFillIdFramebuffer.js
```diff
@@ -0,0 +1,163 @@
...
+ * During this pass, non-behind planar fill geometry writes its per-fragment
+ * feature ID into a screen-sized texture. Later, primitives with
+ * <code>behind: true</code> (BENTLEY_materials_planar_fill) sample this
+ * texture to check whether the pixel they are about to overwrite belongs to
+ * the same logical object (same feature ID). Only then do they apply an
+ * additional depth offset, pushing the fill behind its own object's edges
+ * and non-behind fills — but NOT behind other objects.
+ *
+ * Layout:
+ *   Color attachment 0 – RGBA: R = encoded feature ID (float), GBA reserved.
```
mzschwartz5: What are GBA reserved for? Could we use not just use an attachment with a single channel (R) to save memory?

packages/engine/Source/Scene/PlanarFillIdFramebuffer.js
```diff
@@ -0,0 +1,163 @@
...
+  const width = viewport.width;
+  const height = viewport.height;
+
+  // Feature IDs are integers — use a float format for best range of values.
+  // Prefer FLOAT over HALF_FLOAT to avoid precision issues with models that
+  // have many features. Check colorBufferFloat/colorBufferHalfFloat which
+  // indicates we can render to float textures, not just create them.
+  // Fall back to UNSIGNED_BYTE if neither is supported.
+  let pixelDatatype;
+  if (context.colorBufferFloat) {
```
mzschwartz5: I wonder if we really have to check for support here. I think we decided recently that new features don't need to support webgl1, and [in webgl2 this extension is established in all browsers we support.](https://developer.mozilla.org/en-US/docs/Web/API/EXT_color_buffer_float?utm_source=chatgpt.com)

mzschwartz5: Also- why not use an `UNSIGNED_INT` data type? Does it have to do with the fact that the same shader doubles for both feature IDs and the fill?

packages/engine/Source/Shaders/Model/ModelFS.glsl
```diff
@@ -1,6 +1,19 @@
 
 precision highp float;
 
+// ──────────────────────────────────────────────────────────────────────
+// BENTLEY_materials_planar_fill constants
+// ──────────────────────────────────────────────────────────────────────
+// Depth pull factor (0.05% toward camera) for all planar fills.
```
mzschwartz5: ~~Did you mean to write 5% instead of 0.05% (and same for 2%)?~~

Also, I think this comment is a little misleading. I think it needs to call out that the depth pull/push occur in log space, after `czm_writeLogDepth`. This 5% pull is 5% in log space, which is a variable amount in eye space depending on the fragment's distance from the camera. (Which may be the intended effect)

packages/engine/Source/Shaders/Model/ModelFS.glsl
```diff
@@ -152,5 +185,57 @@ void main()
...
+        float myFeatureId = float(featureIds.PLANAR_FILL_FEATURE_ID);
+
+        // storedFeatureId < 0 means "no planar fill at this pixel".
+        if (storedFeatureId >= 0.0 && abs(storedFeatureId - myFeatureId) < FEATURE_ID_TOLERANCE) {
+            // Proportional push: multiply by >1 to move away from camera.
+            // Net effect: PLANAR_DEPTH_PULL * BEHIND_DEPTH_PUSH ≈ 0.9997,
+            // still in front of non-planar but behind same-feature non-behind fills.
+            gl_FragDepth *= BEHIND_DEPTH_PUSH;
+        }
+    }
```
mzschwartz5, line 248: I need to think about this more... generally speaking, I've encountered numerous issues (many of which in my work with billboards) when trying to fudge depth values. It's very prone to sneaky edge cases. 

With billboards, there were some issues I never solved and ultimately concluded a more robust solution would involve a separate render pass entirely...

mzschwartz5 (changes requested):
Hey @markschlosseratbentley, looking good to me. Most of my comments are code style nits or questions. Functionally, seems alright. I have more reading and testing to do but shipping this round of comments in the mean time.

markschlosseratbentley:
Thanks @mzschwartz5 - I'll be going through your feedback.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: Thanks for the suggestion. I created a `processPlanarFill()` function. Much more modular. A relevant portion of the big block comment was placed as the header of the function.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: The spec does not guarantee a single feature ID set. The spec intentionally leaves "belonging to the same logical object" implementation-defined (for Bentley/iTwin it means "originating from the same element").

This implementation defines it as the primitive's first feature ID set, which holds for Bentley-produced tiles; primitives with no feature IDs simply don't participate in the behind test. I've now documented this assumption explicitly in the new `processPlanarFill()` docs and at the `featureIds[0]` access.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: Hoisted.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: Non-behind fills write the ID texture in the pre-pass; behind fills only read it. I've reworded the comments to say READ/WRITE explicitly.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: Added a comment noting the default returns `false` for the main command and that `derivePlanarFillIdCommand` overrides it to return `true`.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: I agree with your rant. Since my changes follow the existing pattern, I've left it as-is. Your point here may warrant a wider repo-wide refactor. Perhaps an issue makes sense.

packages/engine/Source/Scene/Model/MaterialPipelineStage.js
markschlosseratbentley: Issue filed: https://github.com/CesiumGS/cesium/issues/13620

Now mentioned in code docs.

packages/engine/Source/Scene/Model/ModelDrawCommand.js
markschlosseratbentley: Good call. Collapsed this to a single line.

Also noticed this was an undeclared property, so I also now declared it as an initialized boolean field on `PrimitiveRenderResources` alongside `hasSilhouette` / `hasSkipLevelOfDetail`, which makes the bare assignment properly typed.

packages/engine/Source/Scene/Model/ModelDrawCommand.js
markschlosseratbentley: This should be called once per frame. It's invoked from `initialize()` in the `ModelDrawCommand `constructor, following the pattern of the translucent/silhouette/edge derived commands. Added a note to the function docs.

packages/engine/Source/Scene/GltfLoader.js
markschlosseratbentley: Yep, that's a good suggestion and cleans it up a lot. Done.

`PlanarFill` now has a constructor that takes and destructures the extension JSON and has `@param` docs.

packages/engine/Source/Scene/PlanarFillIdFramebuffer.js
markschlosseratbentley: GBA was a conceptual reservation during development. It ended up we didn't need them in this implementation. Good callout. I moved this to a single-channel R attachment. All seems to work correctly still.

packages/engine/Source/Scene/PlanarFillIdFramebuffer.js
markschlosseratbentley: Thanks for the pointer about not needing webgl1 support.

I went ahead and dropped the `HALF_FLOAT` branch. I did keep a minimal `context.colorBufferFloat ? FLOAT : UNSIGNED_BYTE` guard, because tests could possibly not advertise the extension.

Open to changing this further.

packages/engine/Source/Shaders/Model/ModelFS.glsl
markschlosseratbentley: The percentages are actually correct as written: 1 − 0.9995 = 0.0005 = 0.05%, and 1.0002 − 1 = 0.02%. But your second point about log space is fair. It's intended behavior. The comment block now spells all of this out.

packages/engine/Source/Scene/PlanarFillIdFramebuffer.js
markschlosseratbentley: RE: UNSIGNED_INT: integer output would require an integer output declaration. Due to shader program sharing, we keep the vec4 output here (float is fine range for representing IDs).

packages/engine/Source/Shaders/Model/ModelFS.glsl
markschlosseratbentley: Fair concern.

In the end, it's the same family of technique as the `u_polygonOffset` built into `czm_writeLogDepth`. Also, a similar proportional depth tolerance is employed in the edge visibility system. So there is precedent, which I was leaning on.

Also worth noting: the spec's "planar renders in front of non-planar at the same surface" semantics require some depth override regardless; a separate pass would still need the feature-ID comparison and a bias or depth-test trick inside it, so it isolates the problem more than it removes it (unless we sort things....)

That said, if some blocker artifact turns up, we could eventually move this into a dedicated pass and try to figure something out. Most of the scaffolding here could carry over.

packages/engine/Source/Scene/Scene.js
```diff
@@ -3753,6 +3833,12 @@ function updateAndRenderPrimitives(scene) {
     scene._enableEdgeVisibility = true;
   }
 
+  // If any primitive with BENTLEY_materials_planar_fill (behind) is present,
+  // enable the feature-ID pre-pass lazily.
+  if (frameState.planarFillRequested && scene._enablePlanarFillId === false) {
```
mzschwartz5: nit- is it necessary to check `scene._enablePlanarFillId === false` ? Even if it's already true, doesn't hurt to set it to true again.

mzschwartz5: Not a nit: I don't think `enablePlanarFillId` is ever reset to false, once it has been set to true - is that right? Like, if a model that needs planar fill gets destroyed or something, we should probably disable `enablePlanarFillId` (by seeing that `planarFillRequested` is not true this frame, I would think).

mzschwartz5 (approved):
Just one comment, but on the whole LGTM -- I'm officially approving, so if you decide to make changes based on that comment, feel free to merge after. Nice work! 🚢

packages/engine/Source/Scene/Scene.js
markschlosseratbentley: [Done](https://github.com/CesiumGS/cesium/pull/13178/commits/9d0825302012dc03c253e989c851dca5038b1a43). Ran into some wrinkles, tho. `planarFillRequested` was only set at draw-command build time, not per frame. So... I added a per-frame re-init in `pushDrawCommands`, and Scene now mirrors it each frame as per this comment. We also now release the ID framebuffer while unused (recreated on demand), a really good change.

Verified these changes with Playwright: remove/re-add cycles render identically, no GLerrors, memory actually reclaimed.

Also ran Sandcastle manually. Looked good.

P.S. `_enableEdgeVisibility` has the same pattern for edge rendering. @danielzhong, I'll file an issue (completely unrelated to this PR, but discovered here). Issue: https://github.com/CesiumGS/cesium/issues/13655

packages/engine/Source/Scene/Scene.js
```diff
@@ -3966,6 +3963,9 @@ function updateAndClearFramebuffers(scene, passState, clearColor) {
   const usePlanarFillIdFramebuffer = !picking && scene._enablePlanarFillId;
   if (usePlanarFillIdFramebuffer) {
     view.planarFillIdFramebuffer.update(context, view.viewport, scene._hdr);
+  } else if (!picking) {
```
mzschwartz5, line 3966: So like, in a typical workflow is this framebuffer getting recreated frequently? Or is this more like a step that's really only needed once to generate that ID texture, and then we can release it (and reacquire if we need to regenerate that texture later, like if more models are added or something).

markschlosseratbentley: Allocated once, re-rendered per frame. The ID texture is view-dependent, so its contents re-render regularly, but the fb allocation itself is stable.

I playwright-instrumented the raw GL calls to confirm objectively (using these code changes, not just mentally analyzing the code flow):

- 600 frames, camera orbiting a planar fill dataset: **0** `createFramebuffer`/`createTexture` calls
- model removed: exactly 1 `deleteFramebuffer` + texture deletes, once
- 300 idle frames after removal: 0 GL calls
- model re-added, 600 more frames: 0 after the one-time recreate

So, the fb release and regeneration only happens on the moments you described. i.e. last planar
fill primitive leaves, another one eventually shows up, not mid-workflow.

PR #13187, opened 2026-02-03
Types: Add 'TypedArray' and 'TypedArrayConstructor' type unions

donmccurdy:
# Description

Adds 'TypedArray' and 'TypedArrayConstructor' utility type unions, intended to make it easier to define APIs that accept typed arrays without exhaustively listing many array types. Because TypeScript (used for type checking) and tsd-jsdoc and JSDoc (used for building .d.ts files and docs) have slightly different ways of doing this, the PR tries to thread the needle a bit by putting two `@typedef` tags in `global.js`. This filename is chosen because the types will appear in the documentation under `/Build/Documentation/global.html` regardless of the source filename. I tried `types.js` instead and the result was the same. 

~~It's frankly not obvious to me how TypeScript finds the definition without an explicit import, but it does seem to. I had been somewhat expecting to need to duplicate the types in an ambient `global.d.ts` file.~~ Resolved, let's use type-only imports.

## Issue number and link

- https://github.com/CesiumGS/cesium/pull/12963#discussion_r2466865727
- https://github.com/CesiumGS/cesium/pull/13168#issuecomment-3843465409

## Testing plan

Tested running:

```
npm run build-ts
npm run build-docs
npm run tsc
```

Manually inspected `cesium` and `@cesium/engine` builds to confirm `global.js` leaves no runtime traces. Manually inspected compiled `Cesium.d.ts` to confirm the new types are exported, which is necessary because they're exposed in the public API. Inspected documentation to confirm that output is reasonable:

**Cartesian2.html**

method_types

**global.html**

union_types

# Author checklist

#### PR Dependency Tree

* **PR #13187** 👈

### DIFF
--- a/packages/engine/Source/Core/globalTypes.js
+++ b/packages/engine/Source/Core/globalTypes.js
@@ -9,8 +9,8 @@
  */
 
 /**
- * See `TYPE_ONLY_EXPORTS` in `build.js`. This file adds no runtime exports, and must only
- * define type definitions.
+ * This file is omitted from 'workspaceSourceFiles' in 'scripts/build.js', and
+ * must provide type-only exports exclusively, without runtime values.
  * @ignore
  */
 export default null;


packages/engine/Source/Core/globalTypes.js
mzschwartz5, line 1: Could we maybe name this `globalTypes.js` (or something like that)? Right now it's fine, but down the line I think it could become confusing, if we add some other global sort of namespace.

scripts/build.js
mzschwartz5, line 1: I'm not super confident in this suggestion, but I think a more robust way to prevent the new `global.js` file from being included in various build steps would be to explicitly exclude it from the [workspace source files](https://github.com/CesiumGS/cesium/blob/a618ec998ca11db637ca71aa0441846ee6f76734/scripts/build.js#L255-L266). i.e. add `"!packages/engine/Source/Core/global.js"`

donmccurdy: I like that idea thanks - done! Seems to work as expected. I guess it's possible that this file could get dropped from the documentation or the type definitions later, if we expand the use of the workspace source files. But in any case this seems easier to maintain than my original attempt. And I think the documentation build would error out and let us know, if that happened.

donmccurdy:
Thanks @mzschwartz5! All changes applied. ✅

PR #13009, opened 2025-11-03
Remove banner pointing to legacy sandcastle

jjspace:
# Description

In this release we're removing the banner pointing to legacy sandcastle as the new version of sandcastle should be stable and is the one we want people to use.

The legacy version still exists in the repo and the downloaded zip file should someone really needed. As per discussion in #12904 this will likely change next release.

## Issue number and link

No issue

## Testing plan

- Open sandcastle locally or [in CI](https://ci-builds.cesium.com/cesium/remove-banner/Apps/Sandcastle2/index.html)
- Verify the banner is gone on the new sandcastle

# Author checklist

- [ ] I have submitted a Contributor License Agreement
- [ ] I have added my name to `CONTRIBUTORS.md`
- [ ] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [ ] I have updated the inline documentation, and included code examples where relevant
- [ ] I have performed a self-review of my code

Issue #12877, opened 2025-09-07
Reduce code duplication


javagl:
There is a considerable amount of duplication in the codebase. Until now, this was rather a subjective, anecdotal observation. I occasionally copied out two code blocks and switched back and forth between them, to check whether they are in fact equal, or whether there is s subtle difference (similar to a [blink comparator](https://en.wikipedia.org/wiki/Blink_comparator)). 

It is _important_ (and often pretty easy) to avoid certain forms of duplication. And there's a reason why the first point of the first, most fundamental grade of the 'Clean Code Developer' principles is [Don't Repeat Yourself](https://clean-code-developer.de/en/the-straight/red-degree/#elementor-toc__heading-anchor-1). 

Ironically, this (obviously) has not been repeated often enough.

Don't repeat yourself.
Don't repeat yourself.
Don't repeat yourself.

Of course there is duplication - for example, with error checks, or dummy implementations for certain interface functions, or creating some JSON structure in different places. And we can argue about "how many lines have to be duplicated" vs. "how difficult it is to pull that code into a standalone function", on a case-by-case basis. 

But I recently stumbled over several places ([#12714](https://github.com/CesiumGS/cesium/issues/12714), [#12805](https://github.com/CesiumGS/cesium/issues/12805)) where no discussion is necessary: There are parts of the code that are duplicated in a way or form that cannot be justified. I mean, not _at all_ - not even with the usual "I didn't have enough time to make it 'better'".

I hoped that I could use the magic of ✨AI ✨ and just ask https://deepwiki.com/CesiumGS/cesium "What are the largest functions in the codebase that are duplicated?", but how should a "language model" know that...

So quickly tried out other tools:

- https://github.com/kucherenko/jscpd
- https://github.com/danielstjules/jsinspect
- https://github.com/pmd/pmd

Note: I have **not** performed a detailed analysis or comparison of these! I just ran them, with the default configuration, and looked at the results.

**jscpd** detects several duplications, and ends with a summary table that suggests that a staggering >11000 lines in CesiumJS are copy-pasted. The tool doesn't offer too much configurability, and the duplications that it finds are often pretty small. In JavaScript, I wouldn't even count "5 duplicated lines" as an _actual_ duplication.

**jsinspect** also finds many duplications, The tool also doesn't offer much configurability, but from scrolling over the results, some of them may be worth a look. Some of them are definitely "false positives", but maybe it's possible to narrow that down a bit.

[...]

PR #13188, opened 2026-02-04
Add Sandcastle Copilot AI Assistant 

jdehorty:
## Summary

Sandcastle now ships with **Cesium Copilot**, an in-app AI coding assistant that writes, edits, and debugs CesiumJS code in the same panel as the live preview. Users bring their own keys for Google Gemini, Anthropic Claude, or Google Vertex AI; the assistant edits the JavaScript and HTML buffers through a structured `apply_diff` tool, streams thinking and tool calls live, and can self-fix runtime errors without a manual round-trip.

Keys never leave the browser tab, the model has exactly one editing tool (no freeform code prose), and the diff matcher is built to absorb LLM drift instead of rejecting near-miss edits.

## Architecture

```mermaid
graph TB
    subgraph UI["Sandcastle UI"]
        EDITOR[Code Editor]
        CONSOLE[Console Mirror]
        CHAT[ChatPanel]
        DIFFP[SimpleDiffPreview]
    end

    subgraph Hooks["Copilot React Hooks"]
        TOOLCHAIN[useToolChainExecution]
        AUTOFIX[useAutoFix]
        MSGS[useChatMessages]
    end

    subgraph Core["Copilot Core"]
        FACTORY[AIClientFactory]
        REG["ToolRegistry: apply_diff"]
        DIFF[RooStyleDiffStrategy + DiffMatcher]
        PB[PromptBuilder]
    end

    subgraph Clients["AI Clients (streaming)"]
        GEM[GeminiClient]
        ANT[AnthropicClient]
        VTX[VertexAIClient + VertexAuth]
    end

    subgraph Creds["Credentials"]
        KM["ApiKeyManager (sessionStorage)"]
    end

    EDITOR --> CHAT
    CONSOLE -- "Send to chat" --> CHAT
    CHAT --> MSGS
    CHAT --> TOOLCHAIN
    CHAT --> AUTOFIX
    TOOLCHAIN --> FACTORY
    TOOLCHAIN --> REG
    REG --> DIFF
    PB --> CHAT
    FACTORY --> KM
    FACTORY --> GEM
    FACTORY --> ANT
    FACTORY --> VTX
    GEM -. HTTPS .-> A1[Gemini API]
    ANT -. HTTPS .-> A2[Anthropic API]
    VTX -. HTTPS .-> A3[Vertex AI]
    AUTOFIX -. synthetic message .-> CHAT
    TOOLCHAIN -- modifiedCode --> EDITOR
```

All copilot code lives under `packages/sandcastle/src/copilot/` and is opt-in: the panel is closed by default and only mounts a client when the user sends a message with credentials configured. CesiumJS itself is unchanged.

## Request lifecycle

```mermaid
sequenceDiagram
    actor U as User
    participant C as ChatPanel
    participant F as AIClientFactory
    participant L as AI Client
    participant T as ToolRegistry
    participant E as Editor + Preview
    participant A as useAutoFix

    U->>C: Prompt + JS/HTML/console snapshot
    C->>F: createClientForRoute(model)
    F-->>C: Gemini | Anthropic | Vertex client
    C->>L: generateWithContext(prompt, tools, history)
    loop streaming
        L-->>C: reasoning / text / tool_call
    end

    rect rgb(243, 244, 246)
        Note over C,T: Tool chain (up to 10 calls per turn)
        C->>T: executeTool(apply_diff)
        T-->>C: { modifiedCode, match }
        C->>E: onApplyCode(modifiedCode)
        C->>L: submitToolResult(...)
        L-->>C: next tool_call OR final text
    end

    E-->>A: ExecutionResult (console + diff errors)
    alt errors detected and auto-fix enabled
        A->>C: synthetic prompt (attempt n / 3)
        C->>L: re-enter loop
    else clean run
        A-->>C: idle
    end
```

The chain locks a `workingCodeRef` snapshot for the duration of a turn so chained `apply_diff` calls see their own prior edits, not the React state lagging a render behind.

## Diff application

The model edits code through a single `apply_diff` tool that takes `(file, search, replace)`. The matcher absorbs the small but constant drift LLMs introduce in whitespace, line breaks, and punctuation: it walks the file middle-out in chunks (about 20% of file size, minimum 20 lines), runs a fast 3-strategy cascade per chunk, and falls back to a full 5-strategy pass against the whole file if no chunk matches. The minimum match confidence is 0.9.

```mermaid
flowchart TB
    IN([apply_diff: file, search, replace]) --> MO[Middle-out chunk search<br/>chunks ~20% of file from center]
    MO --> CASC{Per-chunk cascade<br/>min confidence 0.9}
    CASC -- "1. EXACT" --> HIT
    CASC -- "2. WHITESPACE_NORMALIZED" --> HIT
    CASC -- "3. FUZZY (Levenshtein)" --> HIT
    CASC -- no chunk hits --> FB[Whole-file 5-strategy fallback<br/>+ LINE_TRIMMED + CONTEXT_BASED]
    FB -- match --> HIT
    FB -- miss --> NEAR[Return closest near-match preview<br/>+ confidence % + line so the model self-corrects]
    HIT[Match found] --> RECON[Reconcile indentation:<br/>shift replace block by source-vs-search delta]
    RECON --> SPLICE[Splice into source]
    SPLICE --> OUT([modifiedCode + match metadata])
```

After a successful apply, the tool also returns the *other* file's current contents so the model does not have to guess at it on the next edit.

## Key design decisions

1. **Bring-your-own-key, client-side only.** API keys and Vertex service account JSON live in `sessionStorage`, never persisted across tabs and never sent anywhere except the relevant model provider. There is no Cesium proxy and no Cesium-side storage. Trade-off: closing the tab clears the key.
2. **One tool, structured edits only.** The model has exactly one tool, `apply_diff`, and the system prompt forbids freeform code prose. Every change is auditable as a search/replace pair, the editor stays the source of truth, and rejection is just "do not apply".
3. **Drift-tolerant matcher with closest-match feedback.** Five strategies (exact, whitespace-normalized, line-trimmed, fuzzy, context-based) plus middle-out chunking. On a miss, the failed call returns the closest near-match (with confidence percent and line) so the model self-corrects on the next turn instead of resending a slightly-wrong pattern.
4. **Synthetic auto-fix with stall detection.** After each turn, console and diff errors are observed. If auto-fix is enabled and the error fingerprint has changed, a synthetic message kicks off the next attempt (up to 3). Identical fingerprints stop the loop early instead of churning to the cap.
5. **Provider-shaped tool history.** Gemini and Anthropic disagree on how tool results appear in conversation history (`functionResponse` parts vs `tool_use` / `tool_result` blocks). The chain code reconstructs each provider's exact shape per turn from a canonical `ToolCall` / `ToolResult` instead of papering over the difference with a leaky abstraction.

## Model routing

```mermaid
flowchart LR
    PICK[Model picker] --> SEL["ModelSelection { model, route }"]
    SEL --> R{route}
    R -->|direct| P{provider}
    R -->|vertex| RC{region<br/>in allowlist?}
    P -->|gemini| GK[Gemini API key<br/>sessionStorage] --> GC[GeminiClient]
    P -->|anthropic| AK[Anthropic API key<br/>sessionStorage] --> AC[AnthropicClient]
    RC -->|yes| SA[Service account JSON<br/>sessionStorage] --> VC[VertexAIClient]
    RC -->|no| ERR[Reject: region not supported]
    GC --> S[AsyncGenerator&lt;StreamChunk&gt;]
    AC --> S
    VC --> S
```

The picker hides routes the user does not have credentials for, so a user with only an Anthropic key sees only Claude models. Vertex Claude has a hard region allowlist sourced from Google Cloud model cards.

## Trust boundary

Code, HTML, console output, comments, and tool results are tagged as untrusted workspace data in the system prompt. The model is told to never follow instructions embedded in those surfaces and to act only on the system prompt and the user's chat input. This is the prompt-injection defense for users who paste a third-party Sandcastle into a session.

## New dependencies (sandcastle only)

| Package | Purpose |
|---|---|
| `fastest-levenshtein` | Distance metric inside the FUZZY matching strategy |
| `react-markdown` + `remark-gfm` | Render assistant message markdown |
| `react-virtuoso` | Virtualized chat scrollback for long conversations |

No CesiumJS-side dependencies were added. All copilot code is scoped to `packages/sandcastle/src/copilot/`.

## Test plan

**Credentials and routing**

- [ ] Configure Gemini key, Anthropic key, and Vertex service account JSON (with a region). Confirm the model picker shows only the routes you have credentials for, and that switching between Direct and Vertex for the same model both work.

**Editing**

- [ ] On an empty Sandcastle, ask for a basic Cesium scene. Confirm both the JS and HTML buffers populate and the preview runs.
- [ ] On existing code, ask for a small targeted change ("change the camera height to 5000"). Confirm only the requested line changes, no whole-file rewrite.
- [ ] Send a request that requires 2-3 sequential edits. Confirm the editor updates progressively as each `apply_diff` lands and the response ends cleanly.

**Error recovery**

- [ ] Auto-fix happy path: introduce a runtime typo with auto-fix on. Confirm a synthetic retry lands a fix within 3 attempts.
- [ ] Auto-fix stall: introduce an error the model cannot resolve (e.g., reference a CesiumJS API that does not exist). Confirm the loop stops at "stalled" without reaching the 3-attempt cap.
- [ ] Diff near-miss feedback: if an `apply_diff` fails during normal use, confirm the failure surfaces a closest-match preview with a confidence percent and the model adjusts on the next call.

**Integration and persistence**

- [ ] "Send to chat" on a console error line prefills the chat input with the error text.
- [ ] Refresh the tab. Confirm credentials clear (sessionStorage) but the previously selected model is still selected (localStorage).

## Screenshots

image

---

This feature requires users to provide their own API keys for the AI services they wish to use. No API keys are stored on Cesium servers.


jjspace:
Hey @jdehorty and @Guillar1, thanks for getting this started!

I had some general concerns before I dig too deep into the PR itself

1. Can we break this PR up a bit? 35,000 lines of code is a _lot_ to review all at once. Some ideas:
    - Testing setup for React as it's own PR (and then _used_ in this PR)
    - Specific feature sets
        - example basic chat this PR > more advanced diff handling or other parts second PR
        - or single AI model this PR > add more clients in follow up
        - or break out the conversation history tracking/restoration into another PR
2. It seems there was a bad merge somewhere along the way and some old sandcastle changes were reverted.
    - For example the banner linking to the "old" sandcastle which was [removed back in November](https://github.com/CesiumGS/cesium/pull/13009) is back
    - I also believe some performance optimizations we've made are missing as the app feels more sluggish than it used too
    - (also please check the CI errors)
3. It seems like none of the new code uses the Stratakit component library that Sandcastle was built with.
    - This leads to many visual inconsistencies between the copilot UI and the rest of the interface
    - It also means that only half of the new UI actually changes based on the page theme
    - I also see re-implementations of existing components like the `Toggle`
    - I realize the Stratakit package we use has recently been deprecated and replaced with a new MUI based system. Sandcastle has not switched over yet. We would like to but for now I think all the UI should use the existing library as much as possible
4. There are a handful of UX changes/decisions we probably should talk through.
    - One of the biggest that stuck out to me is that we now have 2 (or 3 depending how you count) _separate_ settings panels for the app with different layouts but that share _some_ of the same app settings.
5. The file organization on this branch is a little hard to follow
    - With so many new files and components I agree we should probably have better organization than the, mostly single, directory I was mostly using up until now. I'm open to discussion around the strategy here
        - (breaking up the PR itself will also probably help with this)
    - However it seems some organization was done but not completely or applied consistently. For example there are `Diff` related components at the top `src` directory but there are also `DiffAccordion` and `DiffApprovalModal` components in the `components` directory. Or components moved into `components` but the `Gallery` components didn't move?

There is also a general concern that I don't currently have access to API keys to actually test with the models that have been selected. We can continue that conversation in the other issues that will be opened.

lukemckinstry:
I finally got around to beginning to play around with this. I love the detailed feedback from the agent showing the thought process, code diff, and then explanation & summary of changes. I'll have much more to share shortly but wanted to share how impressive this is.

I'm curious to know if somebody wanted to incorporate MCP, skills, etc. if that is possible now and if not how much of a leap is it?

packages/sandcastle/src/ChatPanel.tsx
```diff
@@ -0,0 +1,1364 @@
...
+                },
+              });
+            }
+          }
+          return parts;
+        };
+
+        const isGemini = selectedModel.model.startsWith("gemini");
+        const conversationHistory = isGemini
+          ? previousMessages.map((msg) => ({
```
markschlosseratbentley: Is there a particular reason why, for Gemini (`isGemini`), empty messages are not filtered out, but they are for the other case (Anthropic)?

packages/sandcastle/src/copilot/hooks/useToolChainExecution.ts
```diff
@@ -0,0 +1,470 @@
...
+            timestamp: Date.now(),
+            isStreaming: true,
+            ...initial,
+          };
+          contMsgId = contMsg.id;
+          addMessage(contMsg);
+        };
+
+        if (!client.submitToolResult) {
+          return;
```
markschlosseratbentley, line 259: Would it make sense to call `addMessage` here so the user has some feedback on what happened?

packages/sandcastle/src/ChatPanel.tsx
Guillar1: Gemini side of things was not brought up to parity yet with Anthropic models. Working on that currently.

packages/sandcastle/src/copilot/hooks/useToolChainExecution.ts
Guillar1: Makes sense. Adding a response.

markschlosseratbentley: Looks, great, thanks!

packages/sandcastle/src/icons.ts
```diff
@@ -20,14 +21,29 @@ import dismiss from "@stratakit/icons/dismiss.svg";
...
+import lock from "@stratakit/icons/lock.svg";
+import send from "@stratakit/icons/send.svg";
+import stop from "@stratakit/icons/stop.svg";
+import chevronDown from "@stratakit/icons/chevron-down.svg";
+import chevronUp from "@stratakit/icons/chevron-up.svg";
+import error from "@stratakit/icons/error.svg";
+import warning from "@stratakit/icons/warning.svg";
+import pin from "@stratakit/icons/pin.svg";
+import cesiumLogo from "./assets/cesium-logo.svg";
+const close = dismiss;
```
jjspace: We should not be renaming icons like this. `dismiss` is already exported, just use that directly please.

jjspace:
@Guillar1 @jdehorty I'm working to condense and organize a bunch of review comments/thoughts but one thing that I think would really really help with the review process is to **stop developing on this branch**! 

Larger changes like implementing the auto-fix workflow should be introduced as separate PRs into this branch. Then we can target review of that code and test that specific workflow in it's entirety. Changes based on review comments are ok but anything larger would be better served in a separate, targeted, PR.

This is already a pretty _massive_ PR and when whole new suites of changes like that get added it makes it even harder to review. Honestly I almost want to request this be split up even more but that might be more effort than it's worth given we're already here.

CC @arobison

javagl:
>  one thing that I think would really really help with the review process 

With zero context here: The mechanism of declaring a PR as 'Draft' or 'Ready For Review' could/should convey the state nicely. Explicit "review requests" are another tool. 
(I only see a review request from 2 months ago, when this had 7 commits (compared to the 86 now), but I assume that most of these commits will be squashed - at least, judging from the commit ""messages"")

packages/sandcastle/src/assets/cesium-logo.svg
javagl, line 1: This is _not_ the Cesium logo! 
The logo can be found in the "Logo kit" that can be downloaded from https://cesium.com/press/

ggetz:
@Guillar1 @jdehorty Please let us know once this is ready for code review, thanks!

packages/sandcastle/src/ConsoleMirror.tsx
```diff
@@ -93,22 +95,23 @@ export function ConsoleMirror({
...
+              error: log.type === "error",
+              special: log.type === "special",
+            })}
+          >
+            <ConsoleIcon type={log.type} />
+            <span className="message-index">{i + 1}:</span>
+            <pre className="content">{log.message}</pre>
+            {renderLogAction && (
+              <div className="log-action-slot">{renderLogAction(log, i)}</div>
+            )}
```
jjspace, line 112: I'm not a huge fan of this pattern of passing a component into another component just to render. If there was conditionals around whether it was added in `App.tsx` or we were planning for multiple different types of actions I could maybe see it, for now it seems overkill and maybe even obfuscating to future devs. 

I think I'd prefer the whole `ConsoleMirror` taking a new prop for the `handleSendConsoleLineToChat` function and change these lines to directly render the actions.

Open to discussion, maybe no immediate action needed.

packages/sandcastle/src/ConsoleMirror.css
```diff
@@ -54,8 +55,24 @@
...
     }
     .content {
+      flex: 1;
+      min-width: 0;
       text-wrap: wrap;
     }
+    .log-action-slot {
+      margin-left: auto;
+      opacity: 0;
+      transition: opacity 0.15s ease;
```
jjspace, line 66: I like the fact that the button is hidden by default and only shows on hover. I believe this is how the one in Chrome Devtools works too. However, this creates an awkward empty gap on lines with long content. 

Image

Any ideas on how to address this? I don't think making the button take up space only when visible would be very nice because it would cause a reflow of the content which could be jarring. Maybe just allow it to overlap content? or maybe we can fill the gap with something like the timestamp that we don't really care if it get's overlapped?

Probably not strictly necessary so feel free to consider it scope creep and/or address in a separate PR

packages/sandcastle/src/ConsoleMirror.tsx
jjspace: I was about to add a comment about how this shouldn't display on the Standalone page because the copilot isn't enabled there. Was that the motivation for this pattern? if so then maybe this is fine as is

packages/sandcastle/src/util/ConsoleWrapper.ts
```diff
@@ -257,6 +257,29 @@ function setupWindowErrorHandler(iframeBridge: BridgeToApp) {
...
+  // Cesium's async paths (tile/terrain loading, tileset ready) surface as
+  // unhandled promise rejections, which the "error" event above misses.
+  window.addEventListener("unhandledrejection", (event) => {
+    const reason = event.reason;
+    const errorMsg =
+      reason instanceof Error
+        ? `${reason.name}: ${reason.message}`
+        : typeof reason === "string"
+          ? reason
+          : (() => {
```
jjspace: Generally not a fan of chained ternary operators. Even less a fan of calling anonymous functions inline. I think this is much more readable

```js
let errorMsg;
if (reason instanceof Error) {
  errorMsg = `${reason.name}: ${reason.message}`;
} else if (typeof reason === "string") {
  errorMsg = reason;
} else {
  try {
    errorMsg = JSON.stringify(reason);
  } catch {
    errorMsg = String(reason);
  }
}
```

(really happy you found that the other handler wasn't catching everything tho, hopefully this helps even more avoid the `[object Object]` logs)

packages/sandcastle/src/App.tsx
```diff
@@ -497,10 +688,37 @@ function App() {
...
+                    index={index}
+                    onSend={handleSendConsoleLineToChat}
+                  />
+                )}
               />
             </Allotment.Pane>
           </ViewerConsoleStack>
         </Allotment.Pane>
+        {chatPanelOpen && (
+          <Allotment.Pane
```
jjspace, line 694: When the copilot window is opened after the vertical separator has been moved it snaps to a set width. I feel like it should remain at the point it was and only move if it would prevent the chat panel from being visible. Is this a symptom of adding the allotment pane like this where it's not always rendered? 

Conversely, when the chat panel is _closed_ it stays at the size it was if it was moved.

https://github.com/user-attachments/assets/b49d4ee7-8207-48a4-bd9b-def2985cd0ed

packages/sandcastle/src/copilot/index.ts
jjspace, line 1: Generally I think we should avoid these sorts of "barrel files" for JS except at the top level of a package meant to be consumed by others (like the built files for the `engine` package. Do you have any thoughts on this? or can you share the motivation to use this approach?

packages/sandcastle/src/copilot/ErrorBoundary.tsx
jjspace, line 1: Minor organizational nitpick, this component feels general enough that it could apply to the whole app not just the copilot/chat components. Should it be outside the `/copilot` directory?

Also worth bringing up a discussion point. On another React project I'm working on they specifically do _not_ have an error boundary type component and instead take an approach of making sure they handle all errors as close to the error as possible. Is this component just a bandaid/stopgap that's cheating error handling? or is the inclusion of the chat models unpredictable enough that we can't possibly account for everything? Or even in that scenario should we be handling errors closer to the unpredictable parts?

packages/sandcastle/src/copilot/ErrorBoundary.tsx
```diff
@@ -0,0 +1,77 @@
...
+            padding: "16px",
+            display: "flex",
+            flexDirection: "column",
+            alignItems: "center",
+            justifyContent: "center",
+            height: "100%",
+            gap: "8px",
+          }}
+        >
+          <p style={{ margin: 0, fontWeight: 600 }}>Something went wrong</p>
```
jjspace: If we do keep this component it _might_ be worth adding a comment here about reporting issues on our forum/github if they happen repeatedly

packages/sandcastle/src/copilot/ErrorBoundary.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <p style={{ margin: 0, fontSize: "0.85em", opacity: 0.7 }}>
+            {this.state.error?.message}
+          </p>
+          <button
```
jjspace: should use the stratakit components to match the page style. If the errors are so deep that stratakit can't load then we have bigger problems that shouldn't have even made it past a PR imo

packages/sandcastle/src/assets/cesium-logo.svg
jjspace: the sandcastle package already includes the base logo in `public/images/cesium-logomark` with a png and svg version. There is an open internal issue for discussion of a new AI + cesium logo but until that's resolved we should just use the base one until then.

packages/sandcastle/src/main.tsx
```diff
@@ -2,11 +2,16 @@ import { StrictMode } from "react";
...
 import { SettingsProvider } from "./SettingsProvider.tsx";
+import { CopilotSettingsProvider, ModelProvider } from "./copilot";
 
 createRoot(document.getElementById("app-container")!).render(
   <StrictMode>
     <SettingsProvider>
-      <App />
+      <CopilotSettingsProvider>
+        <ModelProvider>
+          <App />
```
jjspace, line 12: We should consider the option to disable all of copilot _at build time_. This can be part of a separate PR/effort. It may be as simple as just not including these providers and all downstream components not rendering if the context doesn't exist. Something to think about

packages/sandcastle/src/util/bucket-client.ts
```diff
@@ -45,6 +45,31 @@ function loadSandcastle(code: string, html: string) {
...
+  // so late-synchronous errors post via ConsoleWrapper before we signal.
+  let signaled = false;
+  const signalRunComplete = () => {
+    if (signaled) {
+      return;
+    }
+    signaled = true;
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+        bridge.sendMessage({ type: "runComplete" });
```
jjspace, line 61: This feels like a code smell to me. is there not a better way to watch for a completed script? Maybe modifying the `Helpers.ts#embedInSandcastleTemplate` function to call something after all sandcastle code has run? I suspect that might run into issues if there are errors thrown though... 🤔 

What is the core motivation for detecting this? it probably wouldn't mean the CesiumJS code has finished running.

packages/sandcastle/src/App.css
```diff
@@ -76,6 +76,10 @@
   .content {
     grid-area: content;
   }
+
+  .right-panel {
+    overflow: hidden;
+  }
```
jjspace: Just curious what caused this to get added? iirc the allotment styles already do this, where there overflow issues you observed?

packages/sandcastle/src/App.tsx
```diff
@@ -59,6 +67,22 @@ import {
   useCodeState,
 } from "./util/useCodeState.ts";
 
+type PendingChatDraft = {
+  id: string;
+  text: string;
+};
+
+function formatConsoleMessageForChat(log: ConsoleMessage) {
```
jjspace: Only used in 1 place and only 7 lines long, consider inlining

packages/sandcastle/src/App.tsx
```diff
@@ -497,10 +688,37 @@ function App() {
...
+            maxSize={800}
+            preferredSize={450}
+            className="chat-panel-pane"
+          >
+            <ErrorBoundary>
+              <ChatPanel
+                onClose={() => setChatPanelOpen(false)}
+                codeContext={codeContext}
+                onApplyCode={handleApplyAiCode}
+                onClearConsole={handleClearConsole}
```
jjspace, line 705: `resetConsole` already exists, can that be used instead?

packages/sandcastle/src/App.tsx
```diff
@@ -135,6 +183,62 @@ function App() {
...
+
+  type RunCollection = {
+    resolve: (
+      errors: Array<{ message: string; type: "error" | "warn" }>,
+    ) => void;
+    collected: Array<{ message: string; type: "error" | "warn" }>;
+    timeoutId: ReturnType<typeof setTimeout>;
+  };
+
+  const runCollectionRef = useRef<RunCollection | null>(null);
```
jjspace: This name confused me for a while until I really read through the code that used it, same with the Type. Can we consider renaming to `runErrorCollection`? or maybe just `errorCollection`? or `messageCollection`?

jjspace (commented):
Thanks for all the cleanup here @Guillar1. I (and others) will probably review in batches as I get time to. For this round I focused on a more thorough review of the code for files that I'm already familiar with, ie only the ones with changes. I think this also focuses in on the "interface" between the new chat UI and the rest of the application. I'm happy to see it seems much less "coupled" than it previously was which is a good thing if we consider a version where the copilot is completely disabled. 

Note, some of these comments are simply comments/questions/discussion points and don't require work or changes in this PR (yet)

jjspace:
Please also merge in `main` and check the eslint errors. The last release updated the version for vite so builds should be a lot faster now!

packages/sandcastle/src/ConsoleMirror.tsx
Guillar1: Yup, it was so it does not render in the standalone. I left it as a render prop for modularity/control and maybe future expansion of functionality or actions.

packages/sandcastle/src/copilot/index.ts
Guillar1: At the time, I was thinking about build-time disabling, hoping it would be more convenient. Related to this: https://github.com/CesiumGS/cesium/pull/13188#discussion_r3222242203

packages/sandcastle/src/main.tsx
Guillar1: Agreed. I had this in mind when cleaning this up. Can be in another PR since I believe that a lot of the ground work is already set.

packages/sandcastle/src/copilot/ErrorBoundary.tsx
Guillar1: Question 1: The component is generic and not copilot-specific. It just ended up there because that's the only place using it today. We can move it outside and apply it to the whole app if needed in another PR.

Question 2: We do handle expected errors close to the source (API failures, parse errors have their own handling in the clients and hooks). This boundary is specifically for React render-phase crashes, which can't be caught with try/catch. The copilot is particularly prone to them because the tool chain can produce unexpected state shapes from the LLM output.

packages/sandcastle/src/util/bucket-client.ts
Guillar1: This was a tricky one. The new auto-fix loop needs to know when the script finishes so it can collect console errors before deciding to retry. The script load/error event plus the 2 animation frame delays gives `ConsoleWrapper` time to forward errors via `postMessage` before we close the collection window. Adding a callback to the template instead wouldn't work since it wouldn't fire when the script throws, which is the case we most need to detect. If your wondering about the 2s timeout, It's a fallback.

packages/sandcastle/src/App.tsx
Guillar1: They are not the same. `resetConsole` has `if (codeState.runNumber > 0)`, so it won't clear on the first run. The copilot uses `onClearConsole` in the auto-fix loop where we need an unconditional clear regardless of run number.

packages/sandcastle/src/App.css
Guillar1: Removed

packages/sandcastle/src/App.tsx
Guillar1: Good call

packages/sandcastle/src/App.tsx
Guillar1: Renamed to `runErrorCollectionRef`

packages/sandcastle/src/copilot/ErrorBoundary.tsx
Guillar1: Done.

packages/sandcastle/src/App.tsx
Guillar1: Yes, that's because the pane unmounts when closed and remounts with `preferredSize={450}` when reopened. We're not tracking the separator position via Allotment's `onChange` right now. Can always add a ref to store the positioning but the left panel does the same. It seems out of scope for this PR.

packages/sandcastle/src/util/ConsoleWrapper.ts
Guillar1: Refactored.

packages/sandcastle/src/icons.ts
Guillar1: Done

packages/sandcastle/src/ConsoleMirror.css
Guillar1: https://github.com/user-attachments/assets/a62ba9f8-6a98-4f46-9017-5ea4392f3f94

What do you think?

packages/sandcastle/src/copilot/ErrorBoundary.tsx
Guillar1: Added a small link.

packages/sandcastle/src/ConsoleMirror.tsx
jjspace: I almost wonder if this should be dynamically added _in_ the `ConsoleWrapper` only with the presence of the Copilot Context which wouldn't exist in the standalone page. :thinking: I do think it's getting in the range of bikeshedding for this PR so fine as is, something to consider going forward maybe.

packages/sandcastle/src/ConsoleMirror.css
jjspace: I actually like that a lot better, do you feel it's still as usable? I think it is, and the "margin" for it is spaced about the same as the left side so it doesn't look out of place anymore.

packages/sandcastle/src/App.tsx
jjspace: Can you add a comment to the new handler with that context? just something for future devs to not wonder the same thing

packages/sandcastle/src/ConsoleMirror.css
Guillar1: Feels usable. It's very similar to Chrome's console in DevTools. I pushed it. Give it a try.

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+                          ? "Key saved \u2022 enter new key to replace"
+                          : "sk-ant-..."
+                      }
+                      autoComplete="off"
+                      spellCheck={false}
+                    />
+                  }
+                />
+                <Field.Description>
+                  Get your key at console.anthropic.com/settings/keys
```
jjspace: Can we turn this into a clickable link?

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+                          ? "Key saved \u2022 enter new key to replace"
+                          : "AIza..."
+                      }
+                      autoComplete="off"
+                      spellCheck={false}
+                    />
+                  }
+                />
+                <Field.Description>
+                  Get your key at aistudio.google.com/app/apikey
```
jjspace: Same here

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+                  <Button
+                    variant="outline"
+                    onClick={() => {
+                      ApiKeyManager.clearApiKey();
+                      refreshModels();
+                      setApiKey("");
+                      setGeminiError(null);
+                    }}
+                  >
+                    Clear
```
jjspace: I'm not sure how I feel about these `Clear` buttons. Can you share a bit about the motivation for adding them here?

My main concern is with the conditional rendering pushing the `Save` button over. This could create an unexpected UX if the user expects to "click the button where save normally is" and instead deletes their key. Maybe the bottom action buttons should always be only `Save` and `Cancel` with the `Clear` moved up closer to/inside the actual input?

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+                        showVertexJson
+                          ? undefined
+                          : ({
+                              WebkitTextSecurity: "disc",
+                            } as React.CSSProperties)
+                      }
+                      {...({ "data-1p-ignore": "true" } as Record<
+                        string,
+                        string
+                      >)}
```
jjspace: What is this construct for?

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+                      className={
+                        showVertexJson ? undefined : "vertex-json-masked"
+                      }
+                      style={
+                        showVertexJson
+                          ? undefined
+                          : ({
+                              WebkitTextSecurity: "disc",
+                            } as React.CSSProperties)
+                      }
```
jjspace: If we're conditionally setting a class _and_ a style can the inline styles just be moved into the css?

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+              </Field.Root>
+              {vertexProjectId && (
+                <Field.Root>
+                  <Field.Label>Project ID</Field.Label>
+                  <Field.Control
+                    render={
+                      <TextBox.Input
+                        type="text"
+                        value={vertexProjectId}
+                        readOnly
```
jjspace: If this is not editable can we show it as just text without the input to avoid confusion?

packages/sandcastle/src/copilot/ChatMessage.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  const isMessageStreaming = message.isStreaming === true;
+
+  // Suppress empty code blocks until the closing fence arrives.
+  const hasIncompleteCodeFence = useMemo(() => {
+    if (!message.content) {
+      return false;
+    }
+
+    const fenceOpenings = ["```javascript", "```js", "```html"];
```
jjspace, line 42: I feel like this should include the opening fence without a language just to be safe

packages/sandcastle/src/copilot/ChatMessage.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  const handleCopyMarkdown = async () => {
+    try {
+      await navigator.clipboard.writeText(message.content);
+      setCopiedToClipboard(true);
+      clearTimeout(copiedTimeoutRef.current);
+      copiedTimeoutRef.current = setTimeout(
+        () => setCopiedToClipboard(false),
+        2000,
+      );
```
jjspace, line 79: While I'm not super familiar with how to use it I'm pretty sure this is exactly what the React builtin hook `useTransition` is for? This probably is fine as is but might be worth looking at

packages/sandcastle/src/copilot/ChatMessage.tsx
```diff
@@ -0,0 +1,325 @@
...
+    !hasVisibleContent &&
+    message.isStreaming &&
+    !shouldRenderTypingIndicator
+  ) {
+    return null;
+  }
+
+  const existingMessageBody = (
+    <div
+      className={`chat-message ${isUser ? "user-message" : "ai-message"} ${hasStructuredContent ? "structured-message" : ""}`}
```
jjspace: This could probably be better suited with the `classnames` library

packages/sandcastle/src/copilot/ChatMessage.tsx
```diff
@@ -0,0 +1,325 @@
...
+            <IconButton
+              label={copiedToClipboard ? "Copied!" : "Copy markdown"}
+              icon={copy}
+              variant="ghost"
+              onClick={handleCopyMarkdown}
+            />
+          )}
+        </div>
+
+        {message.attachments && message.attachments.length > 0 && (
```
jjspace, line 152: In my limited testing I don't specifically recall attachments on messages. Is this still supported? And if it is, is there an easy way to test it?

packages/sandcastle/src/copilot/ChatMessage.tsx
```diff
@@ -0,0 +1,325 @@
...
+                key={`tool-${message.id}-${toolCallItem.toolCall.id || index}`}
+                toolName={toolCallItem.toolCall.name}
+                input={toolCallItem.toolCall.input}
+                status={
+                  toolCallItem.result
+                    ? toolCallItem.result.status === "success"
+                      ? "success"
+                      : "error"
+                    : "pending"
+                }
```
jjspace: I feel like this sort of thing could get messy if statuses ever changed? It's also just a little hard to read. Would it be better to move this processing into the component so you could just pass `toolCallItem.result?.status` and maybe a `?? "pending"`?

packages/sandcastle/src/copilot/ChatMessage.tsx
```diff
@@ -0,0 +1,325 @@
...
+      suffix = " — Stalled";
+    } else if (status === "capped") {
+      suffix = " — Gave up";
+    }
+
+    return (
+      <div>
+        <Text
+          variant="body-sm"
+          style={{ display: "block", marginBottom: "0.5rem" }}
```
jjspace: I think we should be avoiding inline styles, can this be moved to classes?

packages/sandcastle/src/copilot/ApiKeyDialog.css
```diff
@@ -0,0 +1,22 @@
+.api-dialog-content {
+  display: flex;
+  flex-direction: column;
+  gap: var(--stratakit-space-x4);
```
jjspace: This creates an awkward looking gap between the tab bar and the tab content. I assume this was added to create separation between the alert and tabs, can we move that to a `margin-top` on the tabs themselves?

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+                        string
+                      >)}
+                    />
+                  }
+                />
+                <Field.Description>
+                  Download from GCP Console &gt; IAM &gt; Service Accounts &gt;
+                  Keys
+                </Field.Description>
+                <div className="api-dialog-actions">
```
jjspace: The placement of this button is creating some strange space below the textbox. Could we maybe consider shifting it above the input in the space to the right of the label?

Image

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,369 @@
...
+  const handleVertexClear = () => {
+    ApiKeyManager.clearVertexServiceAccount();
+    setVertexJson("");
+    setVertexRegion("global");
+    setVertexError(null);
+    refreshModels();
+  };
+
+  return (
+    <SandcastleDialog open={open} onClose={onClose} title="API Configuration">
```
jjspace: Values input into this modal do not seem to reset correctly when the modal is closed. I would expect values that are not _saved_ to not persist when the modal is re-opened.

1. Open the config modal
2. Input anything into the Anthropic API key input
3. Close the modal with Cancel or the `X`
4. Open the modal again and observe the value still exists

The same happens in reverse too

1. Enter vertex json and save
2. Open the modal and delete the input
3. Close without saving
4. Open the modal again and the input is still empty. I would expect it to be populated with the saved value

packages/sandcastle/src/copilot/PromptInput.tsx
```diff
@@ -0,0 +1,204 @@
...
+                    disabled={disabled}
+                  >
+                    Remove
+                  </Button>
+                )}
+              </div>
+            ))}
+          </div>
+        )}
+        <textarea
```
jjspace: Is there a reason this doesn't use the stratakit textbox? There's an example in the `SharePopover`.

Currently this input does not allow scrolling or resizing so longer prompts are not fully visible. The focus styling also seems to not match

packages/sandcastle/src/copilot/PromptInput.tsx
```diff
@@ -0,0 +1,204 @@
...
+          rows={1}
+          disabled={disabled}
+          aria-label={ariaLabel}
+          aria-multiline="true"
+        />
+      </div>
+      {isStreaming && onStop ? (
+        <IconButton label="Stop generating" icon={stop} onClick={onStop} />
+      ) : (
+        <IconButton
```
jjspace, line 198: I think we may want to consider making this button larger. Currently it's rather small to actually click and looks a little out of place creating a gap with the current height of the input

Image

I think setting it to the same height as the input could work. If you do you should also change the icon to the large one, check how the Appbar buttons do it for the syntax.

Image

Per my other comment, if we allow resizing of the prompt input then I think this button should have a locked size and rest aligned to the _top_ of the prompt. Open to discuss if you feel otherwise

packages/sandcastle/src/copilot/ThinkingBlock.tsx
```diff
@@ -0,0 +1,93 @@
...
+              <>
+                {statusLabel}{" "}
+                <span
+                  className="message-typing-dots thinking-status-dots"
+                  aria-hidden="true"
+                >
+                  <span className="message-typing-dot"></span>
+                  <span className="message-typing-dot"></span>
+                  <span className="message-typing-dot"></span>
+                </span>
```
jjspace: There is a typing indicator used in the `ChatMessage` component as well that follows almost identical structure. Would it be worth extracting to a new component like `TypingIndicator` that can encapsulate the styles and any needed logic?

jjspace (commented):
Another batch of comments as I'm starting to work my way through the new code/components. A lot of these are minor and nitpicky but I think worth mentioning. Feel free to push back on any if you feel they're not necessary and we can discuss.

packages/sandcastle/src/copilot/ai/clients/GeminiClient.ts
```diff
@@ -0,0 +1,617 @@
...
+            if (chunk.error) {
+              yield {
+                type: "error",
+                error: chunk.error.message,
+              };
+              return;
+            }
+
+            const parts = chunk?.candidates?.[0]?.content?.parts;
+            if (parts) {
```
javagl, line 241: I'm not sure what's the right time (and literally: place) for pointing this out. But in view of https://github.com/CesiumGS/cesium/issues/12877 , I'd suggest scanning the code in this PR with a duplication detector. When I once casually scrolled over this code, I noticed that https://github.com/CesiumGS/cesium/blob/012104c7e12b667f18bce8bf911005c2c37b9496/packages/sandcastle/src/copilot/ai/clients/GeminiClient.ts#L253 and https://github.com/CesiumGS/cesium/blob/012104c7e12b667f18bce8bf911005c2c37b9496/packages/sandcastle/src/copilot/ai/clients/GeminiClient.ts#L524 are essentially the same ~40 lines of code:

Image

This should be wrapped into a function, with appropriate (i.e. very elaborate) `/** JSdoc */` comments.

A code block like
```
const { functionCall } = part;
const thoughtSignature =
  part.thoughtSignature ??
  part.thought_signature ??
  functionCall.thoughtSignature ??
  (functionCall as { thought_signature?: string })
    .thought_signature;
```
should definitely not exist _twice_ in a codebase. And it preferably shouldn't even exist _once_. This block should also be a 
`const thoughtSignature = extractThoughtSignatureFrom(part);`
call, with that `extractThoughtSignatureFrom` being a function that carries about 5 lines of `/** JSDoc */` explaining what the structure of the _input_ and _output_ is.

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
Guillar1: It tells 1Password's browser extension to skip this field. Without it, 1Password injects its autofill icon into the textarea. I was using 1Password and I noticed it.

packages/sandcastle/src/copilot/ThinkingBlock.tsx
Guillar1: Extracted a TypingIndicator component

packages/sandcastle/src/copilot/PromptInput.tsx
Guillar1: Switched to `TextBox.Textarea`

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
Guillar1: I just removed `Clear` entirely. The keys are stored in `sessionStorage` anyways. They are very temporary. You can just paste a new one and save.

packages/sandcastle/src/copilot/ChatMessage.tsx
Guillar1: It would be nice, but I gave the typing indicator a delay so it doesn't flash on fast responses.

packages/sandcastle/src/copilot/ChatMessage.tsx
Guillar1: Yup, image attachments are supported. Paste or drop any `PNG/JPEG/WebP/GIF` into the prompt input and it shows as an inline thumbnail on the user message.

packages/sandcastle/src/copilot/ChatMessage.tsx
Guillar1: Already handled by the odd-count fallback on lines 52-53. Can't add bare ``` to the explicit list because `lastIndexOf` would match closing fences too, causing false positives on complete code blocks.

ggetz:
Apologies for the late-breaking feedback but, now that I'm taking a much overdue pass, I'm going to mention things as I see them. We don't necessarily need to deem these as "in scope" and require them before merging this PR. However, we should make sure we document items that we decide to defer to later updates.

---

## High level comments

### 1. Copilot has no context for example title or description

It appears that the chat has context for the current editor code and any console errors but **the chat has no context of the example title, description, tags, or other metadata**. I think passing these through would be really helpful, especially for generating explanations or help content.
   
#### Test case

I opened "3D Tiles Gaussian Splats / Multiple LODs" (the first showcase example) and in the copilot chat entered "Explain what I'm looking at." The result just talks about the code structure, not the relevance of the demo. For instance, it never says anything about splats.
   
 > This Sandcastle demo loads a **Cesium 3D Tileset** (Ion asset `4547222`) and lets you orbit around it interactively. Here's a breakdown:
 > 
> ...

If instead I provide the sandcastle metadata in my prompt like so,

> Explain what I'm looking at.

> The sandcastle example has the following properties:
> 
> ```yml
> title: "3D Tiles Gaussian Splats – Multiple LODs"
> description: A Gaussian Splats 3D Tiles dataset using multiple levels of detail (LOD). As you zoom in, the dataset transitions from a lower LOD with fewer, larger splats to a higher LOD with more detailed splats, balancing detail, visual fidelity, and performance.
> labels:
  >- Showcases
  >- 3D Tiles
 > - Gaussian Splats
 > - Performance
>```

I get a much more relevant response:

> This is a CesiumJS Sandcastle demo showcasing a **Gaussian Splats 3D Tiles dataset with multiple Levels of Detail (LOD)**, located in the **Seattle/Redmond, WA area** (coordinates near Microsoft's campus).

### 2. Don't reset session if panel is closed

Closing and opening the chat panel resets the session. Instead, it would be useful to persist the current session if the panel is closed/reopened in the same browser session. I think iterating with copilot, closing the chat panel to see the scene full screen, then re-opening the panel to further iterate would be a common workflow.
   
I think I understand that the current functionality from a simplicity of implementation standpoint. So it may make sense to address this when we tackle persisting other data like key configuration or user authorization.

### 3. Excessive error messages in chat

What's the intention behind showing the "Applying code changes to javascript" error message with the big red tag while copilot is iterating? I'm seeing them quite frequently when I've asked to update the JS code after a few iterations. As the user, I'm not clear if it needs my input or something has gone more fundamentally wrong.

image

### 4. Unrecoverable error state

Maybe related to the above, but after seeing the error a few times, I'm running into the following state, as reported by copilot. I cannot really recover from it without closing the session and starting a new one.

> The file is not empty so I can't use the empty-search form. Let me ask: **can you paste the current exact contents of the JavaScript editor?** The code shown in your prompt does not match what's actually in the file (the diff tool is finding completely different lines), so I need the real current code to target the edit correctly.

Let me know if you are having trouble reproducing any of the error states, and I can give you some example prompts.

Guillar1:
> #### Test case
> I opened "3D Tiles Gaussian Splats / Multiple LODs" (the first showcase example) and in the copilot chat entered "Explain what I'm looking at." The result just talks about the code structure, not the relevance of the demo. For instance, it never says anything about splats.
> 
> > This Sandcastle demo loads a **Cesium 3D Tileset** (Ion asset `4547222`) and lets you orbit around it interactively. Here's a breakdown:
> > ...
> 
> If instead I provide the sandcastle metadata in my prompt like so,
> 
> > Explain what I'm looking at.
> 
> > The sandcastle example has the following properties:
> > ```yaml
> > title: "3D Tiles Gaussian Splats – Multiple LODs"
> > description: A Gaussian Splats 3D Tiles dataset using multiple levels of detail (LOD). As you zoom in, the dataset transitions from a lower LOD with fewer, larger splats to a higher LOD with more detailed splats, balancing detail, visual fidelity, and performance.
> > labels:
> > - Showcases
> > - 3D Tiles
> > - Gaussian Splats
> > - Performance
> > ```
> 
> I get a much more relevant response:
> 
> > This is a CesiumJS Sandcastle demo showcasing a **Gaussian Splats 3D Tiles dataset with multiple Levels of Detail (LOD)**, located in the **Seattle/Redmond, WA area** (coordinates near Microsoft's campus).

The plan is later to tie in more metadata/context into the copilot. Even a screenshot of what the viewer itself is displaying. For now it only can see the code and console errors.

> 2. Don't reset session if panel is closed
> Closing and opening the chat panel resets the session. Instead, it would be useful to persist the current session if panel is closed/reopened in the same browser session. I think iterating with copilot, closing the chat panel to see the scene full screen, then re-opening the panel to further iterate would be a common workflow.
> 
> I think I understand that the current functionality from a simplicity of implementation standpoint. So it may make sense to address this when we tackle persisting other data like key configuration or user authorization.

I agree. I did investigate this pain point a while back and it feels out of scope for this first PR as it will probably involve modifying Sandcastle logic.
> ### 3. Excessive error messages in chat
> What's the intention behind showing the "Applying code changes to javascript" error message with the big red tag while copilot is iterating? I'm seeing them quite frequently when I've asked to update the JS code after a few iterations. As the user, I'm not clear if it needs my input or something has gone more fundamentally wrong.
> 
> image
> ### 4. Unrecoverable error state
> Maybe related to the above, but after seeing the error a few times, I'm running into the following state, as reported by copilot. I cannot really recover from it without closing the session and starting a new one.
> 
> > The file is not empty so I can't use the empty-search form. Let me ask: **can you paste the current exact contents of the JavaScript editor?** The code shown in your prompt does not match what's actually in the file (the diff tool is finding completely different lines), so I need the real current code to target the edit correctly.
> 
> Let me know if you are having trouble reproducing any of the error states, and I can give you some example prompts.

When facing these errors. Can you specify the model and prompt you were using?

ggetz:
> > 3. Excessive error messages in chat
> > 4. Unrecoverable error state

> When facing these errors. Can you specify the model and prompt you were using?

1. Open **Cesium Copilot** panel
2. **API Key Settings** > Set **Vertex AI** config JSON > **Save**
2. Open "Panorama" Sandcastle
3. Adjust settings
    - Model: **Claude Sonnet 4.6**
    - **Thinking**: on
    - **Auto-fix errors**: on
4. Prompt copilot 
    > Modify this example to show only the equirectangular panorama
5. Code eventually successfully updates, but there are a few errors in the chat along the way:
    image
   image
6. Continue to prompt copilot. Not every session, but in some, eventually the file gets so corrupt I get stuck in an error loop. Some prompts:
    > Add a glTF model with a highly reflective PBR material to the scene
    
    > Adjust the model's environment map to match the equirectangular panorama
    
    > Add another glTF model with a semi-reflective PBR material to the scene

ggetz:
> The plan is later to tie in more metadata/context into the copilot. Even a screenshot of what the viewer itself is displaying. For now it only can see the code and console errors.

> I agree. I did investigate this pain point a while back and it feels out of scope for this first PR as it will probably involve modifying Sandcastle logic.

Both sound reasonable. Could you point me to where these are documented? If there not already documented, I'd like to find an issue to do so.

packages/sandcastle/src/copilot/ChatMessage.tsx
jjspace: I had never tried pasting an image, only drag-and-dropping which did not work. I think that's probably fine for now to avoid more work but can you make sure we note down that as a follow up workflow to add a drag and drop overlay and functionality for this?

packages/sandcastle/src/App.css
jjspace: Noticed this might've been missed

packages/sandcastle/src/copilot/ApiKeyDialog.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.api-dialog-project-id {
+  display: flex;
+  flex-direction: column;
+  gap: var(--stratakit-space-x1);
+}
+
+.api-dialog-project-id-label {
+  font-size: var(--stratakit-font-size-14);
```
jjspace: I don't think we should change the size here. It seems to create a weird visual/text hierarchy

Image

packages/sandcastle/src/copilot/ApiKeyDialog.tsx
```diff
@@ -0,0 +1,348 @@
...
+              />
+              <Field.Description>
+                Get your key at{" "}
+                <a
+                  href="https://console.anthropic.com/settings/keys"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  console.anthropic.com/settings/keys
+                </a>
```
jjspace: There's a stratakit component for that 😅 I'm sorry we're using an old version of stratakit so there's not really any easy docs to point to for these things

```suggestion
                <Anchor
                  href="https://console.anthropic.com/settings/keys"
                  target="_blank"
                >
                  console.anthropic.com/settings/keys
                </Anchor>
```

Guillar1:
> > ### 3. Excessive error messages in chat
> > ...
> > ### 4. Unrecoverable error state
> > ...
> > 
> > Let me know if you are having trouble reproducing any of the error states, and I can give you some example prompts.
> 
> When facing these errors. Can you specify the model and prompt you were using?

Should behave better now.

packages/sandcastle/src/copilot/ChatMessage.tsx
Guillar1: https://github.com/CesiumGS/cesium/issues/13521

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+  const handleReject = useCallback(() => {
+    if (isApplying) {
+      return;
+    }
+    onReject();
+  }, [isApplying, onReject]);
+
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (isCollapsed || isApplying || isApplied) {
```
jjspace: This is a comment only looking at the code but what is this handler for? When is it active and what is the intention? 

If I'm reading this first condition correctly it seems like this should always early return? if that's the case I'd like to avoid unused code and window level listeners which could create confusion

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+  return { linesAdded, linesRemoved };
+}
+
+export function SimpleDiffPreview({
+  originalCode,
+  modifiedCode,
+  language,
+  fileName,
+  onApply,
+  onReject,
```
jjspace: It looks like these 2 handlers are `noop` functions in the only place this component is used. Are they necessary to keep around? If can we also remove any of the code that is set up to call these functions?

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+  isApplying = false,
+  isApplied = false,
```
jjspace: Similar comment here, in the only place I'm seeing this component used this value is always `true`. Can we just remove the prop and simplify any code branches where it would be false?

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+interface DiffLine {
+  type: "added" | "removed" | "unchanged" | "modified";
+  originalLineNum?: number;
+  modifiedLineNum?: number;
+  originalContent?: string;
+  modifiedContent?: string;
+  content: string;
+}
+
+function calculateSimpleDiff(original: string, modified: string): DiffLine[] {
```
jjspace, line 27: It seems that the logic of this function only works line by line which I expect commonly makes diffs with a "zebra" pattern like below.

Image

1. Is there a better way we can generate the diff to batch chunks?
2. It seems whitespace is included, should it be ignored? see the first few lines above?
3. Is there a library or something that can do this for us, it seems like a common problem that should be somewhat "solved" depending how complex we want to get.

I don't think this is the end of the world for an initial version. If this would be better suited to a follow-up PR/effort please make note of that wherever we're tracking future work.

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+        </div>
+      </div>
+
+      {isApplied && !isCollapsed && (
+        <div
+          className="simple-diff-preview-applied-banner"
+          role="status"
+          aria-live="polite"
+        >
+          <span className="applied-banner-icon">&#10003;</span>
```
jjspace: Image

I think we should generally avoid having any elements without padding, I think it just looks a tad bit weird and probably breaks some design rules. Maybe it would be enough to just remove the background circle on the icon?

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+        </div>
+      )}
+
+      {!isCollapsed && (
+        <>
+          <div className="simple-diff-preview-content">
+            <pre className="simple-diff-lines">
+              {diff.map((line, idx) => (
+                <div
+                  key={idx}
```
jjspace: Maybe it's ok here since the array is unlikely to change but I seem to recall some strong recommendation somewhere for React to avoid using the index as a key?

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+        </>
+      )}
+
+      {showSuccess && (
+        <div
+          className="simple-diff-preview-success"
+          role="alert"
+          aria-live="polite"
+        >
+          ✓ Changes applied successfully
```
jjspace: Why are there 2 different successful statuses?

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.css
```diff
@@ -0,0 +1,389 @@
...
+    font-size: 0.75rem;
+  }
+}
+
+.simple-diff-preview button:focus-visible {
+  outline: 2px solid var(--stratakit-color-border-accent-base);
+  outline-offset: 2px;
+}
+
+@media (prefers-contrast: high) {
```
jjspace: [Valid values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-contrast) are `no-preference`, `more`, `less`, or `custom`. These styles won't be used currently. Check other places in the new CSS too, I assume this was from the LLM somewhere along the way

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.css
```diff
@@ -0,0 +1,389 @@
...
+  border-radius: var(--stratakit-ext-radius-sm);
+  display: inline-flex;
+  align-items: center;
+  gap: var(--stratakit-space-x1);
+}
+
+.stat-added {
+  background: rgba(34, 197, 94, 0.15);
+  color: var(--stratakit-color-text-success-base);
+  border: 1px solid rgba(34, 197, 94, 0.3);
```
jjspace: There are a bunch of "magic colors" used in this file. Can we

1. Use [stratakit defined values](https://stratakit.bentley.com/tokens/) where possible?
2. Define variables for the new colors on this component where not possible to use stratakit values so they are centralized?

packages/sandcastle/src/copilot/settings/SettingsPanel.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { dismiss } from "../../icons";
+import { FeatureSettings } from "./FeatureSettings";
+import { AdvancedSettings } from "./AdvancedSettings";
+import "./SettingsPanel.css";
+
+interface SettingsPanelProps {
+  onClose: () => void;
+}
+
+export function SettingsPanel({ onClose }: SettingsPanelProps) {
```
jjspace, line 11: I'm still not a _huge_ fan of having a second settings panel than the main application panel. However I'm ok with leaving it as is for now and maybe moving later.

That said, the new panel as 2 tabs of different settings but each only has 1 option. Can we just combine them and simplify the UI for now? If we expand the settings down the road and need to make better categorization or use of the space we can always break them back up.

jjspace (commented):
Another batch of comments from a bit more in depth code review. I'm still not done reviewing all the files but I think I'm getting close :crossed_fingers: (github says i've marked about half of them as "viewed" which I was using to mean I'm happy with it)

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
Guillar1: https://github.com/CesiumGS/cesium/issues/13523

packages/sandcastle/src/copilot/ApiKeyDialog.css
Guillar1: It's default now.

packages/sandcastle/src/copilot/settings/SettingsPanel.tsx
Guillar1: Dropped the tabs.

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
Guillar1: Removed the circle.

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
Guillar1: No you are right. It's usually a last resort. I found a more unique key.

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
Guillar1: Killed the dead code.

packages/sandcastle/src/copilot/diff-preview/SimpleDiffPreview.tsx
Guillar1: Removed the toast and kept just the applied banner.

ggetz:
I was testing the suggested prompts, and I noticed I was getting some very "2D-ish" results. I experimented with additional prompt framing via **Custom System Prompt Addendum** and found this produced consistent results that were more "3D-ish".

> Whenever possible, showcase the unique 3D capabilities of CesiumJS. Prefer scenes that include terrain, 3D Tiles, buildings, elevation, lighting and atmospheric effects, subtle camera movement, time dynamics, clipping, measurement, styling, or other spatial interactions. For example, prefer **one** of either Google 3D Tiles or Cesium World Terrain as base layers over the default ellipsoid. Consider Cesium 3D Buildings, point clouds, glTF models, CZML, or other 3D Tilesets when they add value.
>
> Handle height, terrain, and object placement carefully. Do not assume that longitude/latitude positions with a hardcoded height will align with the visible ground or building surface. When placing markers, labels, billboards, models, or entities that should appear on the ground, use ground-clamping behavior where appropriate, such as `heightReference: Cesium.HeightReference.CLAMP_TO_GROUND` or `heightReference: Cesium.HeightReference.RELATIVE_TO_GROUND`, rather than absolute heights. When using `Cartesian3.fromDegrees`, carefully consider the third argument, height above the ellipsoid surface, based on the base layer or terrain.
> 
> Prefer oblique or perspective camera views to better communicate depth, scale, or height. The camera should be positioned to clearly show the main subject using zoom level, heading, pitch, and range. Be careful to consider the height of the subject, which may be above or below the default ellipsoid surface height. When flying to a specific position, use `camera.flyToBoundingSphere` with `offset` option. Be cautious when using bounding volume centers for camera targeting, since they may be above the meaningful feature.
> 
> Use imagery and base layers intentionally. Change imagery layers when doing so improves contrast, legibility, visual interest, or thematic fit. For example, use a clean vector-style map when symbols or overlays need to stand out, and use photorealistic or satellite imagery when realism and context are important.
> 
> Apply cartographic and information design principles. Use colors, avoid low-contrast overlays, provide titles and context, include legends for symbology and colors, label important features, and add a scale or distance reference when real-world size is not obvious. Prefer established cartographic color schemes, especially ColorBrewer-style palettes, when styling geospatial data. Use sequential, diverging, or qualitative palettes based on the structure and meaning of the data.
> 
> Design DOM UI to support the scene, not obscure it. Do not use `position: absolute` when positioning content inside of the `#toolbar` element. The focus of the page should be the CesiumJS. Overlays and other panels should take up minimal screen space. Be cautious of where the UI may overlap the scene and camera view when positioning DOM elements.

Could we consider adding this, or a portion of this, to the default prompt? Is length a concern here?

Guillar1:
> I was testing the suggested prompts, and I noticed I was getting some very "2D-ish" results. I experimented with additional prompt framing via **Custom System Prompt Addendum** and found this produced consistent results that were more "3D-ish".
> 
> > Whenever possible, showcase the unique 3D capabilities of CesiumJS. Prefer scenes that include terrain, 3D Tiles, buildings, elevation, lighting and atmospheric effects, subtle camera movement, time dynamics, clipping, measurement, styling, or other spatial interactions. For example, prefer **one** of either Google 3D Tiles or Cesium World Terrain as base layers over the default ellipsoid. Consider Cesium 3D Buildings, point clouds, glTF models, CZML, or other 3D Tilesets when they add value.
> > Handle height, terrain, and object placement carefully. Do not assume that longitude/latitude positions with a hardcoded height will align with the visible ground or building surface. When placing markers, labels, billboards, models, or entities that should appear on the ground, use ground-clamping behavior where appropriate, such as `heightReference: Cesium.HeightReference.CLAMP_TO_GROUND` or `heightReference: Cesium.HeightReference.RELATIVE_TO_GROUND`, rather than absolute heights. When using `Cartesian3.fromDegrees`, carefully consider the third argument, height above the ellipsoid surface, based on the base layer or terrain.
> > Prefer oblique or perspective camera views to better communicate depth, scale, or height. The camera should be positioned to clearly show the main subject using zoom level, heading, pitch, and range. Be careful to consider the height of the subject, which may be above or below the default ellipsoid surface height. When flying to a specific position, use `camera.flyToBoundingSphere` with `offset` option. Be cautious when using bounding volume centers for camera targeting, since they may be above the meaningful feature.
> > Use imagery and base layers intentionally. Change imagery layers when doing so improves contrast, legibility, visual interest, or thematic fit. For example, use a clean vector-style map when symbols or overlays need to stand out, and use photorealistic or satellite imagery when realism and context are important.
> > Apply cartographic and information design principles. Use colors, avoid low-contrast overlays, provide titles and context, include legends for symbology and colors, label important features, and add a scale or distance reference when real-world size is not obvious. Prefer established cartographic color schemes, especially ColorBrewer-style palettes, when styling geospatial data. Use sequential, diverging, or qualitative palettes based on the structure and meaning of the data.
> > Design DOM UI to support the scene, not obscure it. Do not use `position: absolute` when positioning content inside of the `#toolbar` element. The focus of the page should be the CesiumJS. Overlays and other panels should take up minimal screen space. Be cautious of where the UI may overlap the scene and camera view when positioning DOM elements.
> 
> Could we consider adding this, or a portion of this, to the default prompt? Is length a concern here?

I added the correctness parts to the default prompt: ground-clamping, the `Cartesian3.fromDegrees` height arg, and `flyToBoundingSphere`. Those prevent actual bugs.

I'd hold off on the aesthetic/cartographic ones though. They're more design preference than correctness and can override what the user actually asked for e.g. a plain 2D map. Feels like it belongs more in the `Custom System Prompt Addendum` you used to test them. Opt-in rather than baked in for everyone.

ggetz (approved):

ggetz:
> I'd hold off on the aesthetic/cartographic ones though. They're more design preference than correctness and can override what the user actually asked for e.g. a plain 2D map. Feels like it belongs more in the Custom System Prompt Addendum you used to test them. Opt-in rather than baked in for everyone.

Understood. 

I do think we want to encourage results that emphasis the value of CesiumJS over other geospatial libraries. But I understand that how this is prompted would more considerations, so [I've opened an issue for follow-up](https://github.com/CesiumGS/cesium-sandcastle/issues/14).

ggetz:
Thanks for all the work here @Guillar1 @jdehorty @jjspace and all! I think this is ready to go!

As the first iteration, there were several ideas for improvements pointed out along the way. Please make sure to document any others in issues so we don't lose track!

Issue #12967, opened 2025-10-10
When loading two Gaussian splash models with overlapping positions at the same time, the model will have a flashing bug


zhaobinJava:
### What happened?

Hello，
Before loading the Gaussian splash model, I converted the. splat model file to a regular. b3dm file using GisBox. Then I used the Cesium3DTile set. from URL method to load it. When the positions of two Gaussian splash models overlap, simply drag the map and the models will flash continuously.

### Reproduction steps

1.
2.
3.
...

### Environment

Browser:
CesiumJS Version: 1.133
Operating System:

javagl:
This was also reported in the forum at https://community.cesium.com/t/when-loading-two-gaussian-splash-models-with-overlapping-positions-at-the-same-time-the-model-will-have-a-flashing-bug/43629

mzschwartz5:
Posted some [explanation](https://community.cesium.com/t/when-loading-two-gaussian-splash-models-with-overlapping-positions-at-the-same-time-the-model-will-have-a-flashing-bug/43629/7?u=mzschwartz5) of the bug I received from @keyboardspecialist

zhaobinJava:
Hello, I noticed that the latest update does not include any changes related to my issue,I'm  looking forward to any updates on the issue.

javagl:
For me, it's clearly still visible in the sandcastle that was also provided in the forum:

[...]

PR #13189, opened 2026-02-05
Fix Gaussian Splat race condition problem

danielzhong:
https://github.com/user-attachments/assets/13cfd25f-aa05-497d-83ae-81100579054e

# Description
Fixes a race condition in Gaussian splat sorting / snapshot updates that caused WebGL errors, flickering, and inconsistent LOD loading. The update pipeline is now guarded against stale async results and against sorter unavailability during WASM init, which previously left the system stuck in a partial state.

# Issue
* The error `Vertex buffer is not big enough for the draw call` is non-deterministic and easier to trigger during loading + camera interaction, which is typical of a race rather than a deterministic size limit.
* The error appeared even for smaller splat counts, though more frequently with larger datasets.
* The failure window disappeared when state transitions were tightened, indicating timing/state inconsistency rather than a hard limit.

# Fixes
* Snapshot pipeline with explicit state (BUILDING → TEXTURE_PENDING → TEXTURE_READY → SORTING → READY) to avoid partial/invalid states being committed.
* Generation/token guards: sort results are only accepted if they match the current request and data generation.
* If radixSortIndexes returns undefined, we do not transition to SORTING; instead we keep the snapshot in TEXTURE_READY and retry later.
* Old textures are retired and released safely instead of being destroyed immediately, avoiding “bind deleted object” issues.
* Once a snapshot is committed, we do not mark _dirty again unless inputs actually change.

# Result:
* Eliminates the intermittent WebGL errors and flickering during LOD load/transition.
* Prevents the “stuck in SORTING” state when the sorter is not ready.
* When switching between different LOD levels, loading is still slow (but not flickering anymore),I believe this is a separate optimization issue and not directly related to the race condition.

## Issue number and link
https://github.com/CesiumGS/cesium/issues/12965
https://github.com/CesiumGS/cesium/issues/13016
https://github.com/CesiumGS/cesium/issues/12967

# Update: Added fix for flashing bug:
https://github.com/iTwin/platform-engineering/issues/368

Commit: 
https://github.com/CesiumGS/cesium/commit/554589c923b7832fc4f17c31edb75b3ca8120621

Multiple GaussianSplatPrimitive instances were reusing the same scratch matrix for `modelMatrix`, so one instance could overwrite another and cause flickering.
The fix is to use a per-primitive matrix (_drawCommandModelMatrix) for the stored command state, while keeping scratch matrices only for transient math.

## Testing plan
DEMO (Flashing Bug): https://ci-builds.cesium.com/cesium/DanielZ/Splat_FIX/Apps/Sandcastle2/index.html

Test GS LOD data locally:
Download the data:
[data](https://bentley.sharepoint.com/sites/Platform/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FPlatform%2FShared%20Documents%2FCesiumJS%2FSplats&p=true&ga=1&ovuser=067e9632%2Dea4c%2D4ed9%2D9e6d%2De294956e284b%2CDaniel%2EZhong%40bentley%2Ecom&OR=Teams%2DHL&CT=1770854271473&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI0OS8yNjAxMTUxMTExNiIsIkhhc0ZlZGVyYXRlZFVzZXIiOmZhbHNlfQ%3D%3D)

host the data and run in sandcastle:
```

const url = "http://localhost:8081/CoStarWithSH/tileset.json";

const tileset = await Cesium.Cesium3DTileset.fromUrl(url, {
  maximumScreenSpaceError: 16,
});


window.tileset = tileset;

viewer.zoomTo(tileset);
```

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/GaussianSplatPrimitive.js
+++ b/packages/engine/Source/Scene/GaussianSplatPrimitive.js
@@ -923,7 +923,10 @@ Object.defineProperties(GaussianSplatPrimitive.prototype, {
    */
   isStable: {
     get: function () {
-      return !this._dirty && !defined(this._pendingSnapshot) || (this._pendingSnapshot.state === SnapshotState.READY);
+      return (
+        (!this._dirty && !defined(this._pendingSnapshot)) ||
+        this._pendingSnapshot.state === SnapshotState.READY
+      );
     },
   },
 


javagl:
I just removed my review request, but as a cross-link for those who will review this: The race condition was also supposed to be avoided with https://github.com/CesiumGS/cesium/pull/13120 . I don't know what is done _here_ what is going beyond that, but if _this_ one is merged, then the other one can be closed.

danielzhong:
> I just removed my review request, but as a cross-link for those who will review this: The race condition was also supposed to be avoided with #13120 . I don't know what is done _here_ what is going beyond that, but if _this_ one is merged, then the other one can be closed.

Thank you for the cross-link.

[PR #13120](https://github.com/CesiumGS/cesium/pull/13120) addressed the specific WAITING → SORTING state transition, which closed one state gap. This change goes a bit further by adding safeguards against stale async completions (generation/request validation), committing snapshots only once they are fully READY, validating expected sort counts, and deferring texture retirement to avoid lifetime races.

In that sense, it covers the original race condition more comprehensively. If this PR is merged, I agree that [#13120](https://github.com/CesiumGS/cesium/pull/13120) can likely be closed as superseded.

markschlosseratbentley:
I tested this locally and saw greatly improved behavior using multiple-LOD splat datasets.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -834,27 +1112,69 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
...
+      ? tileset._gaussianSplatStableFrames
+      : DEFAULT_STABLE_FRAMES;
+    const isStable = this._selectedTilesStableFrames >= stableFramesTarget;
+    const isBootstrap =
+      !defined(this._snapshot) &&
+      !defined(this._pendingSnapshot) &&
+      !defined(this._drawCommand);
+    const maxStallFrames = defined(tileset._gaussianSplatMaxStallFrames)
+      ? tileset._gaussianSplatMaxStallFrames
+      : 30;
```
weegeekps: It would be good to add a comment here about the rationale behind why we feel the magic number 30 is a good value if `tileset._gaussianSplatMaxStallFrames` is unset, as well as explaining how changing this value can impact things. Understanding that seems important for the logic here.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -47,6 +48,223 @@ const GaussianSplatSortingState = {
   ERROR: 4,
 };
 
+const SnapshotState = {
+  BUILDING: "BUILDING",
+  TEXTURE_PENDING: "TEXTURE_PENDING",
+  TEXTURE_READY: "TEXTURE_READY",
+  SORTING: "SORTING",
+  READY: "READY",
+};
```
weegeekps, line 94: We need to add some documentation here explaining what is expected to be happening during each state.

javagl: Somewhat related to https://github.com/CesiumGS/cesium/issues/11749#issuecomment-3224469778

packages/engine/Source/Scene/GaussianSplatPrimitive.js
javagl: In addition to explaining the defaults and magic constants: Properties like `tileset._gaussianSplatMaxStallFrames` formally don't exist. Maybe the intention was to add these later, as configuration/streering properties, but the `_` indicates that they should be `private`.

weegeekps: Yeah. There seems to be a handful of these, like `tileset._gaussianSplatStableFrames` and `tileset._gaussianSplatSortMinFrameInterval` as well as a few more. I question if they should be in the `Cesium3DTileset` object to begin with? Seems better to have them in an additional class.

weegeekps (commented):
The logic here seems alright, but there's some cleanup to be done.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
javagl: I don't have strong opinions on this. (Or ... actually I do have some very strong opinions, but I don't _express_ them strongly). But I wonder whether anybody ever and in any context willingly, intentionally, and with a specific goal in mind, changed the `foveatedMinimumScreenSpaceErrorRelaxation` of a tileset from `0.0` to `0.1`. Maybe the ones that are discussed here are "more important" for steering the behavior. But ... "when in doubt, leave it out".

packages/engine/Source/Scene/GaussianSplatPrimitive.js
danielzhong: Thanks for the reminder! I’ve added comments to clarify what is expected to happen in each state and to better document the lifecycle transitions.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
danielzhong: I’ve removed those exposed _gaussianSplat* variables from tileset. They were only added temporarily for local debugging and I forgot to clean them up.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,262 @@ const scratchMatrix4B = new Matrix4();
...
+// Higher values reduce rebuild churn but keep stale snapshots visible longer.
+const DEFAULT_MAX_SNAPSHOT_STALL_FRAMES = 30;
+// Minimum delay between steady re-sort requests once the camera is moving.
+const DEFAULT_SORT_MIN_FRAME_INTERVAL = 3;
+// ~0.5 degree camera direction change threshold before triggering steady re-sort.
+const DEFAULT_SORT_MIN_ANGLE_RADIANS = 0.008726646259971648;
+// Minimum camera movement in world units before triggering steady re-sort.
+const DEFAULT_SORT_MIN_POSITION_DELTA = 1.0;
+
+function shouldStartSteadySort(primitive, frameState) {
```
weegeekps, line 169: Please add a docstring to these helper functions. I know we haven't been good about that in this module, but we shouldn't continue that pattern.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -906,33 +1255,34 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
                 case 3:
                   coefs = 45;
               }
-              this._shData = new Uint32Array(totalElements * (coefs * (2 / 3)));
+              aggregate = new Uint32Array(totalElements * (coefs * (2 / 3)));
```
weegeekps: I'm not a fan of this calling `new Uint32Array` inside of a loop. It was doing it before, but we probably ought to clean it up now since we're touching it. Given that this is called for every single tile for every frame, I can see this actually having a notable impact on the frame due to its potential size triggering garbage collection.

One strategy I've thought of for dealing with this is just making the array big enough to handle the max number of coefficients and then use a wrapper at the class level for the aggregate that contains the scratch typed array and the current size determined through `totalElements * (coefs * (2 / 3))`. You wouldn't need the wrapper; you could also just declare the length and your scratch array at the class level.

There may be alternative strategies that are better.

danielzhong: Good point. I’ve refactored it to reuse a class-level scratch Uint32Array instead of allocating a new one inside the loop.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
danielzhong: Added, thanks!

ggetz:
Thanks @danielzhong! Could you please add an item to `CHANGES.md` summarizing the update and link any CesiumJS bugs that are being fixed?

Also, are there any unit test that we could potentially add to confirm the update behavior and prevent any regressions in the future?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,358 @@ const scratchMatrix4B = new Matrix4();
...
+  if (
+    primitive._lastSteadySortFrameNumber >= 0 &&
+    framesSinceLastSort < DEFAULT_SORT_MIN_FRAME_INTERVAL
+  ) {
+    return false;
+  }
+
+  const camera = frameState.camera;
+  if (
+    !defined(camera) ||
```
ggetz: If there is no camera in the scene, is it worth sorting at all?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,358 @@ const scratchMatrix4B = new Matrix4();
...
+    activeSort.requestId === primitive._sortRequestId &&
+    activeSort.dataGeneration === primitive._splatDataGeneration
+  );
+}
+
+/**
+ * Destroys the GPU textures owned by a snapshot, if any, and clears the
+ * references so they are not used after destruction.
+ *
+ * @param {object|undefined} snapshot The snapshot whose textures should be destroyed.
```
ggetz: Instead of a generic `object`, which may be error prone, could we define the schema of the snapshot object and use that type here? We can do this with an inline `@typedef` comment [like we do here](https://github.com/CesiumGS/cesium/blob/e70be094db335f39f70841cf289a006bdf74fd98/packages/engine/Source/Core/GeocoderService.js#L5-L10).

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,358 @@ const scratchMatrix4B = new Matrix4();
...
+ * @param {FrameState} frameState The current frame state.
+ * @throws {DeveloperError} If the snapshot is not READY.
+ * @private
+ */
+function commitSnapshot(primitive, snapshot, frameState) {
+  if (!defined(snapshot.indexes) || snapshot.state !== SnapshotState.READY) {
+    throw new DeveloperError("Committing snapshot before it is READY.");
+  }
+
+  const frameNumber = defined(frameState) ? frameState.frameNumber : 0;
```
ggetz: When might `frameState` be `undefined`? 

Right now, based on the doc comment above, `frameState` is a required parameter. If we expect it to sometimes be undefined, mark it as optional. Otherwise, we can assume it's always defined, or guard it with the `Check.defined` function.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,358 @@ const scratchMatrix4B = new Matrix4();
...
+      defined(primitive.sphericalHarmonicsTexture) &&
+      primitive.sphericalHarmonicsTexture !== snapshot.sphericalHarmonicsTexture
+    ) {
+      retireTexture(
+        primitive,
+        primitive.sphericalHarmonicsTexture,
+        frameNumber,
+      );
+    }
+  }
```
ggetz, line 377: This looks like a set of unnecessary nested conditionals. I believe this can be simplified to the following:

 
```suggestion
  const currentSnapshot = primitive._snapshot;
  const splatTexture = defined(currentSnapshot) ? currentSnapshot.gaussianSplatTexture : primitive.gaussianSplatTexture;
  if (
    defined(splatTexture &&
    splatTexture !== snapshot.gaussianSplatTexture
  ) {
    retireTexture(
      primitive,
      splatTexture,
      frameNumber,
    );
  }
  
  const sphericalHarmonicsTexture = defined(currentSnapshot) ? currentSnapshot.sphericalHarmonicsTexture) : primitive.sphericalHarmonicsTexture;
  if (
    defined(sphericalHarmonicsTexture) &&
    sphericalHarmonicsTexture !==
      snapshot.sphericalHarmonicsTexture
  ) {
    retireTexture(
      primitive,
      sphericalHarmonicsTexture,
      frameNumber,
    );
  }
```

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,358 @@ const scratchMatrix4B = new Matrix4();
...
+}
+
+/**
+ * Creates a GPU texture that stores packed spherical harmonics coefficient
+ * data for all splats. The texture uses a two-channel unsigned-integer format
+ * ({@link PixelFormat.RG_INTEGER}) and nearest-neighbor sampling.
+ *
+ * @param {Context} context The WebGL context.
+ * @param {object} shData An object with {@code width}, {@code height}, and
+ *   {@code data} (an {@code ArrayBufferView}) describing the packed SH data.
```
ggetz: Consider also using `@typedef` to define this object's schema as a new type.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -66,6 +411,19 @@ function createSphericalHarmonicsTexture(context, shData) {
...
 }
 
+/**
+ * Creates a GPU texture that stores the packed Gaussian splat attributes
+ * (positions, scales, rotations, colors). The texture uses an RGBA
+ * unsigned-integer format ({@link PixelFormat.RGBA_INTEGER}) and
+ * nearest-neighbor sampling.
+ *
+ * @param {Context} context The WebGL context.
+ * @param {object} splatTextureData An object with {@code width},
```
ggetz: Consider also using `@typedef` to define this object's schema as a new type.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -532,69 +925,97 @@ GaussianSplatPrimitive.transformTile = function (tile) {
...
-    count: primitive._numSplats,
+    count: snapshot.numSplats,
   });
   if (!defined(promise)) {
-    primitive._gaussianSplatTexturePending = false;
+    snapshot.state = SnapshotState.BUILDING;
     return;
   }
   promise
     .then((splatTextureData) => {
```
ggetz: We prefer `async/await` syntax over promise chains. Would it be possible to define asynchronous code nested in this promise chain in one or more new async functions?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -941,54 +1436,154 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
...
+      if (!defined(this._pendingSortPromise)) {
+        if (pending.state === SnapshotState.SORTING) {
+          pending.state = SnapshotState.TEXTURE_READY;
         }
+        return;
       }
+
+      const pendingSort = this._pendingSort;
+      this._pendingSortPromise
+        .then((sortedData) => {
```
ggetz: We prefer `async/await` syntax over promise chains. Would it be possible to define asynchronous code nested in this promise chain in one or more new async functions?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1000,53 +1595,116 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
...
-      this._sorterState = GaussianSplatSortingState.ERROR;
-      this._sorterError = err;
-    });
-    this._sorterPromise.then((sortedData) => {
-      this._indexes = sortedData;
-      this._sorterState = GaussianSplatSortingState.SORTED;
-    });
+    const activeSort = this._activeSort;
+    this._sorterPromise
+      .then((sortedData) => {
```
ggetz: We prefer `async/await` syntax over promise chains. Would it be possible to define asynchronous code nested in this promise chain in one or more new async functions?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1000,53 +1595,116 @@ GaussianSplatPrimitive.prototype.update = function (frameState) {
...
```
ggetz: We prefer `async/await` syntax over promise chains. Would it be possible to define asynchronous code nested in this promise chain in one or more new async functions?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -39,14 +40,358 @@ const scratchMatrix4B = new Matrix4();
...
+  BUILDING: "BUILDING",
+  // Async texture generation/upload is in flight for this snapshot.
+  TEXTURE_PENDING: "TEXTURE_PENDING",
+  // Attribute textures are ready; snapshot can now request index sorting.
+  TEXTURE_READY: "TEXTURE_READY",
+  // Sort request is in flight for this snapshot generation.
+  SORTING: "SORTING",
+  // Sorted indexes were validated for this generation and may be committed.
+  READY: "READY",
+};
```
ggetz, line 94: Would it make sense to encapsulate Snapshots into their own class?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
danielzhong: Good point! Fixed:
https://github.com/CesiumGS/cesium/pull/13189/commits/2275042582773b22411392a85e39eb95b94d7c6e

packages/engine/Source/Scene/GaussianSplatPrimitive.js
danielzhong: Good question. I considered a Snapshot class, but in the current implementation snapshots are mostly transient data records owned by GaussianSplatPrimitive’s state machine, with no independent behavior. For this PR, I kept them as typed data objects and added explicit @typedef schemas to improve safety/readability without increasing scope/risk. I also aimed to stay consistent with the existing pattern used by `GaussianSplatSortingState`, which follows a similar state machine approach for managing sorting state. Given that similarity, I thought it would be better to keep the same pattern here as well and keep the design simple and consistent.

Please feel free to let me know what you think about this or if you have a different perspective and any suggestions!

packages/engine/Specs/Scene/GaussianSplatPrimitiveSpec.js
```diff
@@ -130,6 +133,78 @@ describe(
       });
     });
 
+    it("retries pending snapshot sorting when sorter is temporarily unavailable", function () {
```
ggetz, line 136: This is not urgent to fix before merging—A test is better than none—But, I'm concerned we're creating too many "fakes" to the point where this test is too fragile. It's closely coupled to the implementation so will need to be updated if the class internal change. And it may not necessarily catch when something actually goes wrong.

Maybe re-assessing the test(s) could be a follow-up action.

ggetz:
Thanks for the updates @danielzhong! I think the code is in a good place.

As a heads up, I pushed a few small fixes to the existing splat tests which I noticed when running locally.

PR #13192, opened 2026-02-05
Add `EdgeDisplayMode` enum for controlling edge rendering

markschlosseratbentley:
# Description

This PR adds a three-mode `EdgeDisplayMode` enum that provides finer control over how edges from the `EXT_mesh_primitive_edge_visibility` extension are rendered.

There is now an `edgeDisplayMode` property on both `Cesium3DTileset` and `Model`.

By adding this flag, we actually have implemented a CAD wireframe mode which is desired by [this PR](https://github.com/CesiumGS/cesium/pull/13178) for fully implementing glTF specification [BENTLEY_materials_planar_fill](https://github.com/CesiumGS/glTF/blob/vendor-extensions/extensions/2.0/Vendor/BENTLEY_materials_planar_fill/README.md).
 
_Note: This feature is using part of the glTF spec that is not final and is subject to change without Cesium's standard deprecation policy._

### `EdgeDisplayMode` values
 
 | Mode | Behavior |
 |------|----------|
 | `SURFACES_ONLY` (default) | Surfaces render normally; edges are hidden |
 | `SURFACES_AND_EDGES` | Surfaces render with edges composited on top via MRT |
 | `EDGES_ONLY` | Only edges render (CAD wireframe style); surfaces are suppressed |
 
### Changes
 
 - **New:** `EdgeDisplayMode` enum (`packages/engine/Source/Scene/EdgeDisplayMode.js`)
 - **Refactored:** `Cesium3DTileset.edgeDisplayMode` → `Model3DTileContent` → `Model.edgeDisplayMode` property plumbing
 - **Updated:** `ModelDrawCommand.pushCommands` / `pushEdgeCommands` logic for all three modes
 - **Added:** Production Sandcastle demo (`3d-tiles-edge-visibility`) using Ion asset 4665927
 - **Added:** Unit tests in `EdgeVisibilityRenderingSpec.js` and `ModelSpec.js`
 
### How it works
 
 The core logic lives in `ModelDrawCommand`:
 
 - **`EDGES_ONLY`:** `pushCommands` skips the surface command entirely (early return). `pushEdgeCommands` sets the edge command's pass to `CESIUM_3D_TILE_EDGES_DIRECT`, routing it to the main framebuffer after the opaque pass, bypassing the MRT edge framebuffer used for compositing.
 
 - **`SURFACES_ONLY`:** `pushEdgeCommands` returns early; no edge commands enter the command list.
 
 - **`SURFACES_AND_EDGES`:** Both paths run normally (existing behavior). Edges use `CESIUM_3D_TILE_EDGES` pass → MRT framebuffer → composited on top of surfaces.

### Usage
 
`edgeDisplayMode` is a property on both `Cesium3DTileset` and `Model`.

### Screenshot

image

## Issue number and link

N/A

## Testing plan

- A new production Sandcastle Sample was added: [View this feature in action at this Sandcastle Sample!](https://ci-builds.cesium.com/cesium/markschlosser/cad-wireframe-mode-gltf/Apps/Sandcastle2/index.html?id=3d-tiles-edge-visibility)
- Unit tests have been added.

# Author checklist


packages/engine/Source/Scene/Model/Model.js
```diff
@@ -1204,6 +1208,28 @@ Object.defineProperties(Model.prototype, {
...
+   * glTF extension are rendered relative to surface geometry. Primitives that
+   * do not declare the extension are unaffected by this setting.
+   *
+   * @memberof Model.prototype
+   *
+   * @type {EdgeDisplayMode}
+   *
+   * @default EdgeDisplayMode.SURFACES_ONLY
+   *
+   * @experimental This feature is using part of the glTF spec that is not yet final and is subject to change without Cesium's standard deprecation policy.
```
markschlosseratbentley, line 1228: @ggetz I could use feedback on marking this feature `@experimental` at its various endpoints. It seems appropriate because the spec (https://github.com/KhronosGroup/glTF/pull/2479) has not merged yet. However, I am not 100% sure and could use some feedback.

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -957,6 +959,19 @@ function Cesium3DTileset(options) {
     );
   }
 
+  /**
+   * Controls how edges contributed by the EXT_mesh_primitive_edge_visibility
+   * glTF extension are rendered relative to surface geometry. Tile content
+   * primitives that do not declare the extension are unaffected.
+   *
+   * @type {EdgeDisplayMode}
+   * @default EdgeDisplayMode.SURFACES_ONLY
```
markschlosseratbentley, line 969: @arobison @ggetz Take a look at the default value here - `SURFACES_ONLY`. This means that tilesets / glTF files with the `EXT_mesh_primitive_edge_visibility` extension present will _NOT_ render those edges by default anymore (which is the current behavior on CesiumJS).  I know we wanted to discuss this in some more detail. I am open to making this default whatever make the most sense. I welcome your feedback.

cc @danielzhong @weegeekps Just FYI: according to the proposed spec, [this proposed default behavior is spec-compliant](https://github.com/CesiumGS/glTF/blob/64ed5bea4ccffb2242d2cf86e212388f63b61603/extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md?plain=1#L73). Spec quote:
> Engines are not required to *always* render the edges - they may, for example, permit the user to toggle edge display on and off.

packages/engine/Source/Scene/Model/Model.js
ggetz: I think marking as experimental is safe given the spec is technically in  progress. Though, we should aim to avoid breaking changes if possible 😄

[Similar to our process for deprecations](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#deprecation-and-breaking-changes:~:text=Create%20an%20issue%20to%20remove%20the%20API%20with%20the%20appropriate%20remove%20in%20%5Bversion%5D%20label.), it may be worth creating an issue with the `remove in [version]` tag as a reminder to remove the experimental tag come the October release (as we're aiming for submitting the 3D Tiles 2.0 spec for ratification in September).

packages/engine/Source/Scene/Cesium3DTileset.js
ggetz: This makes sense to me, though I will defer to @arobison and PM if there are other factors.

Please correct me if I overlooked it, but we should probably document the existence of this property and default mode where we list support for `EXT_mesh_primitive_edge_visibility`. It could be confusing behavior to a user if they load a model with the extension, knowing it's supported, and don't see the edges.

packages/engine/Source/Shaders/Builtin/Constants/passCesium3DTileEdgesDirect.glsl
```diff
@@ -0,0 +1,9 @@
+/**
+ * The automatic GLSL constant for {@link Pass#CESIUM_3D_TILE_EDGES_DIRECT}
+ *
+ * @name czm_passCesium3DTileEdgesDirect
+ * @glslConstant
+ *
+ * @see czm_pass
+ */
+const float czm_passCesium3DTileEdgesDirect = 12.0;
```
danielzhong, line 9: Minor point on naming: “3DTiles” is plural, while most existing files use the singular “tile.” Should we standardize on either plural or singular for consistency?

danielzhong (commented):
While reviewing these files locally, I noticed there are two instances of edgeCommand.uniformMap = uniformMap; in ModelDrawCommand.js (lines 851 and 853). Could you remove the duplicate in this PR as well? Thanks!

packages/engine/Source/Scene/Cesium3DTileset.js
arobison: Makes sense to me too. No unexpected edges rendered for folks working with models that might contain edges and for folks who want to see the edge rendering mode asking them to enable it is sensible.

markschlosseratbentley: > Please correct me if I overlooked it, but we should probably document the existence of this property and default mode where we list support for `EXT_mesh_primitive_edge_visibility`. It could be confusing behavior to a user if they load a model with the extension, knowing it's supported, and don't see the edges.

@ggetz It does not seem `EXT_mesh_primitive_edge_visibility` is listed as supported in any of the expected places, probably because the spec did not fully "land" yet. Perhaps we wait for that?

packages/engine/Source/Shaders/Builtin/Constants/passCesium3DTileEdgesDirect.glsl
markschlosseratbentley: Fixed

markschlosseratbentley:
> While reviewing these files locally, I noticed there are two instances of edgeCommand.uniformMap = uniformMap; in ModelDrawCommand.js (lines 851 and 853). Could you remove the duplicate in this PR as well? Thanks!

I removed the duplicate.

packages/engine/Source/Scene/Model/Model.js
markschlosseratbentley: Issue created to remove experimental tag: https://github.com/CesiumGS/cesium/issues/13466

packages/engine/Source/Scene/Cesium3DTileset.js
markschlosseratbentley: I linked the spec PR everywhere this extension is mentioned, and I listed support for the extension in the latest commits (including caveat about this flag controlling the display). LMK if you think I missed a place.

ggetz (approved):
Thanks @markschlosseratbentley, this is looking great!

I pushed up one small change to enable edges and surfaces in the `styled-gltf-lines-dev` sandcastle example. I think it makes sense for the example to show both by default, but does not require any UI to change modes.

Image

PR #13196, opened 2026-02-08
Refactor WMTS imagery provider to use UrlTemplateImageryProvider and enable get feature info

zoran995:
# Description

- Refactor the `WebMapTileServiceImageryProvider` to delegate URL construction and tile fetching to `UrlTemplateImageryProvider`, removing duplicate code between the two. Based on my investigation, it appears that `WebMapTileServiceImageryProvider` was implemented before `UrlTemplateImageryProvider` and has not been refactored to use the latter.
- Implements WMTS GetFeatureInfo for both **KVP** and **RESTful** request encodings
- Adds a Sandcastle demo showcasing feature picking with both encoding styles against live WMTS servers (note: I am not sure about the terms and conditions for using publicly available servers in cesium Sandcastle; I found those available on the web, and we used them for testing the implementation).

## Testing plan

- Run tests and verify all `WebMapTileServiceImageryProviderSpec` tests pass
- Open Sandcastle `web-map-tile-service-picking` demo
  - KVP mode: verify feature info popup opens with JSON attributes upon clicking on tile
  - REST mode: verify feature info popup with JSON attributes upon clicking on tile
  - Toggle "Enable Feature Picking" off → verify clicks no longer query the server
- Verify existing WMTS demos (time-dynamic weather) still work without regression

# Author checklist

### DIFF
--- a/packages/sandcastle/gallery/web-map-tile-service-picking/main.js
+++ b/packages/sandcastle/gallery/web-map-tile-service-picking/main.js
@@ -5,6 +5,33 @@ const viewer = new Cesium.Viewer("cesiumContainer", {
   shouldAnimate: true,
 });
 
+function geoJsonToFeatureInfo(json) {
+  return json.features.map((feature) => {
+    const featureInfo = new Cesium.ImageryLayerFeatureInfo();
+    featureInfo.data = feature;
+    featureInfo.properties = feature.properties;
+    featureInfo.configureNameFromProperties(feature.properties);
+    featureInfo.configureDescriptionFromProperties(feature.properties);
+
+    // If this is a point feature, use the coordinates of the point.
+    if (Cesium.defined(feature.geometry) && feature.geometry.type === "Point") {
+      const [latitude, longitude] = feature.geometry.coordinates;
+      featureInfo.position = Cesium.Cartographic.fromDegrees(
+        longitude,
+        latitude,
+      );
+    }
+
+    return featureInfo;
+  });
+}
+
+const getFeatureInfoJson = new Cesium.GetFeatureInfoFormat(
+  "json",
+  "application/json",
+  geoJsonToFeatureInfo,
+);
+
 const kvpProvider = new Cesium.WebMapTileServiceImageryProvider({
   url: "https://wmts.marine.copernicus.eu/teroWmts",
   layer:
@@ -13,6 +40,7 @@ const kvpProvider = new Cesium.WebMapTileServiceImageryProvider({
   tileMatrixSetID: "EPSG:3857",
   format: "image/png",
   enablePickFeatures: true,
+  getFeatureInfoFormats: [getFeatureInfoJson],
 });
 
 const restProvider = new Cesium.WebMapTileServiceImageryProvider({


packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -209,79 +253,57 @@ function WebMapTileServiceImageryProvider(options) {
...
+    maximumLevel: options.maximumLevel,
+    subdomains: options.subdomains,
+    tileDiscardPolicy: options.tileDiscardPolicy,
+    credit: options.credit,
+    getFeatureInfoFormats:
+      options.getFeatureInfoFormats ??
+      WebMapTileServiceImageryProvider.DefaultGetFeatureInfoFormats,
+    enablePickFeatures:
+      options.enablePickFeatures ??
+      (this._useKvp || defined(options.getFeatureInfoUrl)),
```
zoran995, line 325: With REST encoding, there is no way of deriving the get feature info URL from the tile URL alone, so we default to false in that case.

ggetz:
Thanks for the PR @zoran995!

@lukemckinstry Could you please review?

javagl:
I remember some conversation about that `GetFeatureInfoFormat` topic. Out of curiosity: Is there anything that could be said about the connection or overlap between _this_ PR and https://github.com/CesiumGS/cesium/pull/12993 ? (This is mainly about things that may be relevant for whoever has to review this - if it's completely independent, there may not much to be said...)

lukemckinstry:
Thanks @zoran995 for opening this PR. This looks like a useful improvement to make this class easier to use and maintain. Many of our more recent ImageryProvider classes also use UrlTemplateImageryProvider.

1. 
> terms and conditions for using publicly available servers in cesium Sandcastle

We can review the terms of these servers on our side as well if we are considering publishing the sandcastles along with the PR. If you have links to the terms handy do you mind sharing them to save time tracking them down?

2.
When I tried to load the sandcastle, it seems the KVP imagery is not loading the tile requests are returning errors like.

```
<ExceptionReport xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows/1.1 https://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" xmlns="http://www.opengis.net/ows/1.1" version="1.1.0"><Exception exceptionCode="InvalidParameterValue" locator="layer">Layer NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-2D_PT15M-i_202411/uo not found in catalogue</Exception></ExceptionReport>
```
Do you know if availability changed for this data?

3. Let us know about the question above https://github.com/CesiumGS/cesium/pull/13196#issuecomment-3992813512

zoran995:
Hi @javagl @lukemckinstry, thanks for the review. I have rebased this to the latest main and updated the data URL in sandcastle. The data provider made some changes to the data, and the layer wasn't available anymore. I got the data from https://data.marine.copernicus.eu/product/NWSHELF_ANALYSISFORECAST_PHY_004_013/services. 
On #12993, this is independent of it. Still, as I used WMS as the base to implement this, to keep things consistent, it will have the same getFeatureInfo issue as WMS until that PR is merged (it will work for GeographicProjection but not WebMercatorProjection). Users can use the same workaround as for WMS, or, if that PR is accepted, it will also work for WMTS.

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -570,6 +626,106 @@ WebMapTileServiceImageryProvider.prototype.pickFeatures = function (
...
+/**
+ * The default parameters to include in the WMTS URL to get feature information.  The values are as follows:
+ *     service=WMTS
+ *     version=1.0.0
+ *     request=GetFeatureInfo
+ *
+ * @constant
+ * @type {object}
+ */
+WebMapTileServiceImageryProvider.GetFeatureInfoDefaultParameters =
```
lukemckinstry, line 711: Does it make sense to format [defaultParameters](https://github.com/CesiumGS/cesium/blob/a15123a641c8de008fe8dd0fa0cab3fa34124bf2/packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js#L12)  in `WebMapTileServiceImageryProvider`
 
```
const defaultParameters = Object.freeze({
  service: "WMTS",
  version: "1.0.0",
  request: "GetTile",
});
```
in the same way we do in [WebMapServiceImageryProvider](http://localhost:8080/Build/Documentation/WebMapServiceImageryProvider.html?classFilter=WebMapService#.DefaultParameters)
 so the APIs and docs match up? 

There is the important difference that WMTS supports both KVP and REST, and it looks like these defaults are only for KVP. I am wondering if we should we have defaults for REST too? It seems like up until now in the class we just expect the user to pass service, version & request in the `url` param when using REST. If we have the setTemplateValue calls use useAsDefault=true, then I think adding defaults could be backwards compatible.

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -40,6 +40,17 @@ const defaultParameters = Object.freeze({
...
+ *                          the GetFeatureInfo operation on the WMTS server and return the features included in the response.  If false,
+ *                          {@link WebMapTileServiceImageryProvider#pickFeatures} will immediately return undefined (indicating no pickable features)
+ *                          without communicating with the server.  Set this property to false if you know your WMTS server does not support
+ *                          GetFeatureInfo or if you don't want this provider's features to be pickable.
+ *                          Defaults to true for KVP encoding. For RESTful encoding, defaults to true only when
+ *                          {@link WebMapTileServiceImageryProvider.ConstructorOptions#getFeatureInfoUrl} is specified, and false otherwise.
+ * @property {GetFeatureInfoFormat[]} [getFeatureInfoFormats=WebMapTileServiceImageryProvider.DefaultGetFeatureInfoFormats] The formats
+ *                          in which to try WMTS GetFeatureInfo requests.
+ * @property {Resource|string} [getFeatureInfoUrl] The GetFeatureInfo URL of the WMTS service. If not specified, the value of <code>url</code> is used.
```
lukemckinstry: It would be nice to have as much doc consistency between `WebMapServiceImageryProvider` & `WebMapTileServiceImageryProvider` as possible. Specifically I am looking at the Constructor Options for the two, since our docs generator JSDoc doesn't alphabetize properties & params like it does method names.  
Do you mind reordering so they match (as much as they can). In WMSImageryprovider, I think the getFeatureInfoUrl param can be moved up from the end of the list to be with the other feature info related params. Any other reordering you can do would be appreciated 🙏

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -449,6 +471,26 @@ Object.defineProperties(WebMapTileServiceImageryProvider.prototype, {
       return true;
     },
   },
+
+  /**
+   * Gets or sets a value indicating whether feature picking is enabled.  If true, {@link WebMapTileServiceImageryProvider#pickFeatures} will
+   * invoke the <code>GetFeatureInfo</code> service on the WMS server and attempt to interpret the features included in the response.  If false,
```
lukemckinstry: WMTS

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -495,6 +537,18 @@ Object.defineProperties(WebMapTileServiceImageryProvider.prototype, {
       }
     },
   },
+
+  /**
+   * Gets the getFeatureInfo URL of the WMS server.
```
lukemckinstry: WMTS

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -570,6 +626,106 @@ WebMapTileServiceImageryProvider.prototype.pickFeatures = function (
...
+  Object.freeze(new GetFeatureInfoFormat("json", "application/json")),
+  Object.freeze(new GetFeatureInfoFormat("xml", "text/xml")),
+  Object.freeze(new GetFeatureInfoFormat("text", "text/html")),
+]);
+
+function applyDimensions(imageryProvider, resource, interval) {
+  const staticDimensions = imageryProvider._dimensions;
+  const dynamicIntervalData = defined(interval) ? interval.data : undefined;
+
+  if (!imageryProvider._useKvp) {
```
lukemckinstry: I was going to ask if we should update `WebMapServiceImageryProvider` to also support both KVP and REST? But it looks only KVP is fully standardized for WMS, while WMTS supports both, is that right?

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -1,13 +1,13 @@
 import combine from "../Core/combine.js";
-import Credit from "../Core/Credit.js";
 import Frozen from "../Core/Frozen.js";
 import defined from "../Core/defined.js";
 import DeveloperError from "../Core/DeveloperError.js";
 import Event from "../Core/Event.js";
 import Resource from "../Core/Resource.js";
 import WebMercatorTilingScheme from "../Core/WebMercatorTilingScheme.js";
```
lukemckinstry, line 7: There is a slight regression for WMTS REST calls. In the docs `example 2` works in prod but not the branch. [sandcastle](https://sandcastle.cesium.com/#c=bVJNbxoxEP0ro70AVWQDp5YQVIlGUSRQKkyay17M7rA49cfK9i5QlP8ee5etoMUnz/jNzBu/J1RprIcvwB3M0YlKwdYaBWmSNVGa3Kc61ZnRzkMtcI8WHkDj/owmv5pcv8PPjfZcaLRpMmgqKYXHA1elRBgTeGVPDNyO55iDRSlwC15IdNBfPbL1tpKDblYLWjWY8fXIN9wsebkOdQxtLTJ8VrxAe/xpTS3yQOaUaginshIm0Nt5X7oJpRvuUPGSaO6F0VzGe2Fqym1WCEctOk9d29DRSJRdUHjR8kjD1DgRLX1brhmNzOmIDMnwNvzE/FHiBz1FqkvurTgw9FfxOViZ/fk2N/KDvJdF767dQfJj+PGwxa0JHcjFORGU45ZX0nf5rbGK+/gg4g/R9xL/NvaXnJ5/XFQPx1+V6mCKH4Sq1AJrlBMYfTunM4u5iJ0vZJk3uX7vlQAj8IRGmkJkXAKrwp8de0Haj+iJ1kVEtKIt4oKO8Dz/V8UrBwzuk7tk2iw6ixS+i9a4QeM+IdRjcBj3QbhNlf1GTzLn4qwp7UqmuahB5A83jAqZ5M6Fl+A/ycQfTJPZlAb8VZk0PBe6eAnyB1EiZDeaLdokIWRKQ/h/lTdGbri96PgJ)

I think the problem is caused by this small difference in the tile request URLs.

This PR: https://basemap.nationalmap.gov/arcgis/rest/services/USGSShadedReliefOnly/MapServer/WMTS/tile/1.0.0/USGSShadedReliefOnly/%7BStyle%7D/default028mm/2/0/0.jpg

Prod:
https://basemap.nationalmap.gov/arcgis/rest/services/USGSShadedReliefOnly/MapServer/WMTS/tile/1.0.0/USGSShadedReliefOnly/default/default028mm/2/0/0.jpg

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -146,49 +157,82 @@ function WebMapTileServiceImageryProvider(options) {
...
+  const format = options.format ?? "image/jpeg";
 
   const bracketMatch = url.match(/{/g);
   if (
     !defined(bracketMatch) ||
     (bracketMatch.length === 1 && /{s}/.test(url))
   ) {
-    resource.setQueryParameters(defaultParameters);
+    resource.setQueryParameters(defaultParameters, true);
     this._useKvp = true;
```
lukemckinstry, line 220: I was going to ask if we should update WebMapServiceImageryProvider to also support both KVP and REST? But it looks only KVP is fully standardized for WMS, while WMTS supports both, is that right?

CHANGES.md
```diff
@@ -7,6 +7,7 @@
 #### Additions :tada:
 
 - Added experimental, performance-focused vector primitive APIs: `BufferPointCollection`, `BufferPolylineCollection`, and `BufferPolygonCollection`. [#13212](https://github.com/CesiumGS/cesium/pull/13212)
+- Added GetFeatureInfo support to `WebMapTileServiceImageryProvider`, enabling `pickFeatures()` for both KVP and RESTful WMTS services. New options: `enablePickFeatures`, `getFeatureInfoFormats`, `getFeatureInfoUrl`, and `getFeatureInfoParameters`. [#13196](https://github.com/CesiumGS/cesium/pull/13196)
```
lukemckinstry: slight wordsmith/nitpick on changelog entry formatting,
```
Added GetFeatureInfo support to `WebMapTileServiceImageryProvider`, enabling `WebMapTileServiceImageryProvider.pickFeatures` for both KVP and RESTful WMTS services. New class parameters include `enablePickFeatures`, `getFeatureInfoFormats`, `getFeatureInfoUrl`, and `getFeatureInfoParameters`. [#13196](https://github.com/CesiumGS/cesium/pull/13196)
```

packages/sandcastle/gallery/web-map-tile-service-picking/main.js
```diff
@@ -0,0 +1,121 @@
...
+  url: "https://wmts.marine.copernicus.eu/teroWmts",
+  layer:
+    "NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy-cur_anfc_1.5km-3D_PT1H-i_202511/ubar",
+  style: "cmap:balance",
+  tileMatrixSetID: "EPSG:3857",
+  format: "image/png",
+  enablePickFeatures: true,
+});
+
+const restProvider = new Cesium.WebMapTileServiceImageryProvider({
```
lukemckinstry, line 46: I think we could add this to the class docs as `Example 4`

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -570,6 +626,106 @@ WebMapTileServiceImageryProvider.prototype.pickFeatures = function (
...
+    },
+    tilecol: function (provider, x, y) {
+      return x.toString();
+    },
+    TileMatrixSet: function (provider) {
+      return imageryProvider._tilematrixset;
+    },
+    tilematrixset: function (provider) {
+      return imageryProvider._tilematrixset;
+    },
```
lukemckinstry, line 783: What is the purpose of producing lowercase and uppercase (PascalCase) tags? Can you add a line in the code to document?

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -449,6 +471,26 @@ Object.defineProperties(WebMapTileServiceImageryProvider.prototype, {
       return true;
     },
   },
+
+  /**
+   * Gets or sets a value indicating whether feature picking is enabled.  If true, {@link WebMapTileServiceImageryProvider#pickFeatures} will
+   * invoke the <code>GetFeatureInfo</code> service on the WMS server and attempt to interpret the features included in the response.  If false,
+   * {@link WebMapTileServiceImageryProvider#pickFeatures} will immediately return undefined (indicating no pickable
+   * features) without communicating with the server.  Set this property to false if you know your data
```
lukemckinstry, line 528: Update the end of this doc string to match what you have for `enablePickFeatures` in the `ConstructorOptions` with info about the defaults:

Defaults to true for KVP encoding. For RESTful encoding, defaults to true only when WebMapTileServiceImageryProvider.ConstructorOptions#getFeatureInfoUrl is specified, and false otherwise.

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -1,13 +1,13 @@
...
 import Frozen from "../Core/Frozen.js";
 import defined from "../Core/defined.js";
 import DeveloperError from "../Core/DeveloperError.js";
 import Event from "../Core/Event.js";
 import Resource from "../Core/Resource.js";
 import WebMercatorTilingScheme from "../Core/WebMercatorTilingScheme.js";
-import ImageryProvider from "./ImageryProvider.js";
 import TimeDynamicImagery from "./TimeDynamicImagery.js";
+import UrlTemplateImageryProvider from "./UrlTemplateImageryProvider.js";
+import GetFeatureInfoFormat from "./GetFeatureInfoFormat.js";
```
lukemckinstry, line 10: We should update the `GetFeatureInfoFormat` docs to reference WMTS in addition to WMS.

packages/engine/Specs/Scene/WebMapTileServiceImageryProviderSpec.js
```diff
@@ -722,4 +750,525 @@ describe("Scene/WebMapTileServiceImageryProvider", function () {
...
+
+      spyOn(Resource.prototype, "fetchText").and.callFake(function () {
+        const url = this.getUrlComponent(true);
+        const uri = new Uri(url);
+        const params = queryToObject(uri.query());
+        expect(params.infoformat).toBe("text/html");
+        return Promise.resolve("<html><title>x</title></html>");
+      });
+
+      return providerXml
```
lukemckinstry: remove meaningless true toBe true assertion since this will trigger code smell tests, replace with 

```
return providerXml
  .pickFeatures(0, 0, 0, 0.0, 0.0)
  .then(function () {
    return providerHtml.pickFeatures(0, 0, 0, 0.0, 0.0);
  });
```
I think the idea behind adding this is that Jasmine requires all specs to have assertions, but we have assertions elsewhere in this spec so true toBe true is not needed.

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
zoran995: We can add it, but it won't make a difference for the users, as WMTS has only one version, so the servers always advertise those params already hardcoded in the URL

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
zoran995: In short, WMS supports only KVP requests, while WMTS supports both. WMS is intended for dynamic maps with varying parameters. On the other hand, WMTS is serving mostly pre-generated fixed tiles, and there is a benefit of caching them client-side which is a lot easier with restful requests

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
zoran995: Thanks for catching this. It is caused by a mismatched casing of the `Style` parameter. I have added a `Style` value, along with `style`, for RESTful requests to make it backwards-compatible.

p.s. I am wondering if cesium could use case-insensitive matching param url template params? Not as part of this PR, but in the long run, with a proper potential breaking-change announcement. The values for WMTS Rest URLs are copied from servers, and casing is not strictly consistent. If someone writes `{STYLE} it will fail.

packages/sandcastle/gallery/web-map-tile-service-picking/main.js
zoran995: added

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
```diff
@@ -449,6 +520,27 @@ Object.defineProperties(WebMapTileServiceImageryProvider.prototype, {
...
+  /**
+   * Gets or sets a value indicating whether feature picking is enabled.  If true, {@link WebMapTileServiceImageryProvider#pickFeatures} will
+   * invoke the <code>GetFeatureInfo</code> service on the WMTS server and attempt to interpret the features included in the response.  If false,
+   * {@link WebMapTileServiceImageryProvider#pickFeatures} will immediately return undefined (indicating no pickable
+   * features) without communicating with the server.  Set this property to false if you know your data
+   * source does not support picking features or if you don't want this provider's features to be pickable.
+   * Defaults to true for KVP encoding. For RESTful encoding, defaults to true only when
+   * {@link WebMapTileServiceImageryProvider.ConstructorOptions#getFeatureInfoUrl} is specified, and false otherwise.
+   * @memberof WebMapTileServiceImageryProvider.prototype
+   * @type {boolean}
```
zoran995, line 533: @lukemckinstry I have removed `@default true` from here and added description for it

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
zoran995: I have reordered some of WMS and most of WMTS

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
lukemckinstry: gotcha, sounds like no need to add it then

lukemckinstry (commented):
A few new small comments. I just now followed up on your responses to my first batch of comments as well. 

I realized `example 3` in the class docs did not work without modifications so I opened this PR to update it and another one of our sandcastles using WMTS. https://github.com/CesiumGS/cesium/pull/13288 Feel free to take a look. Your PR looks good with WMTS time series data.

We may want to consolidate your sandcastle examples with another existing WMTS sandcastle .... I'll check with the team, but I promise this will not drag out the process of getting the PR merged.

packages/sandcastle/gallery/web-map-tile-service-picking/main.js
lukemckinstry: If we wanted to get extra fancy we could provide a sandcastle example of WMTS with both time series and picking ....

packages/engine/Source/Scene/WebMapTileServiceImageryProvider.js
lukemckinstry: Regression fix looks good.

For the case-insensitive matching param url template params, feel free to open an issue (or search if one already exists) and we would definitely consider it.

lukemckinstry:
It looks in the coordinates for the selection indicator, lat lon are getting swapped. They are correct in the info box, but the indicator is at the wrong location.

This is from a click near England:

Screenshot from 2026-03-11 14-11-40

**Update:**

I tracked this down .... this is either a small data mistake from the source at https://wmts.marine.copernicus.eu/teroWmts or wrong logic in our code (in a place already under scrutiny here https://github.com/CesiumGS/cesium/pull/12993) based on overlapping standards.


A picking request  (ie. a click near England like the screenshot above) like https://wmts.marine.copernicus.eu/teroWmts?service=WMTS&version=1.0.0&request=GetFeatureInfo&infoformat=application%2Fjson&i=151&j=240&format=image%2Fpng&tilematrix=5&layer=NWSHELF_ANALYSISFORECAST_PHY_004_013%2Fcmems_mod_nws_phy-cur_anfc_1.5km-3D_PT1H-i_202511%2Fubar&style=cmap%3Abalance&tilerow=10&tilecol=15&tilematrixset=EPSG%3A3857

returns the response:
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          49.32432375115863,
          -4.575757575757576
        ]
      },
      "properties": {
        "lat": 49.32432375115863,
        "lon": -4.575757575757576,
        "variableId": "ubar",
        "datasetId": "NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy-cur_anfc_1.5km-3D_PT1H-i_202511",
        "value": 0.16200000047683716,
        "units": "m s-1"
      }
    }
  ]
}
```
and this code is swapping the lon and lat 
https://github.com/CesiumGS/cesium/blob/7878464aae1d00b6610ae3cb4bbcedbc9aa379cb/packages/engine/Source/Scene/GetFeatureInfoFormat.js#L88-L92

There are overlapping standards to consider, epsg3857, OGC, GeoJSON. We can 
 - Use different data, or
 - override the callback passed to GetFeatureInfoFormat like this [sandcastle](https://sandcastle.cesium.com/#c=jVdrb9s2FP0rnD+0MuJKDdItqxMHC9Kmy5Z2QZ2tH+qgYCRaZkORAknZ84b89x2Koiwp7lAgiKKr+zjkPfeRNdVkzdmGaTIjkm3IBTO8KuK/alm0GKX1+4WSlnLJ9GI0PlnIhUwScp5lhJJP7+eEFzRneksE3UJDruGzEV07iYFrHyPuiYOjObOkKsmykqnlShKrSA7RklFbaUa4XCqy1KogdsXqeJqZUknDJrU5lRmpZKnVV5Zawq3X/cTuyXumU2oVIKVQt6TRcTF6h4VuUL1pVSJ30BZTztRvRslbdelRXQFU9BWSMfl3IQkgQSqJk8QNcBMXtIzCKWZnXpEQj2W589MHc9W5om4wB8eZdwzjjFoK60a0RwEnLpm2nJmdWke4xwLoljzH2wdasEvc5E2rHT11MP4/D2+YSTUv3f19lyPvChm9WiLV3BD8UFIqLtvrmpDKsJoHqVI645JaHE0ta1GtGXsvfEmi5kIztgRzszYoUlkwq7dj8uwZGQpjuy2RrNmMLEY3zh8IHxIXUvf57wnZ3nUutDXugDrp2xSBX8rwpwS8oBomnMqjyPmekJfj1r6Xzp31jstxS/5oGCV4eQyX29C047PWeHSKtZJHi/LrUM8Rv4/3Xe/7pdIFtZFztBi5CliMJv6FlqXgQAQoSefDvmJyck+CXo+IaZY1FQH+rHnmmhJUB9VLyznTa56yoW6TukqLabjQxWhlbWmmSZKbF8bSe1Fn8IVRonJITcxtAoGBR6aTSoLGKtkUpkFPfJ8zU3jyH6d/1o8vpSdMo2W54DKfpytWsOk32s1tRycaN4Yl1ag9W4doqWc1lQZfmLRTvFRs0hKkvn6Hpr65pJR5i+GxeeZ7Egbvn58m+m7iCbGQbaOfW9ATRbYkQqkH4CXUkvPKAJLgFAXXNPcUqDWNDbNuevirz5ixriBwr9Nw/o+gKpU5rt116jcs1ww9wQP9Kf7x9WED+uhV/PPRcfNyeBy/Dn+/Oo6Pj2rCQADuEgfUTZ0VZoEYjrJ5Cv9yXtKUvV3j+n71SlGD2qRMMmCHg7oLNT7cMa5kWdlzPw2o2cp0N6Mi5lw1rcHdUR2k15UwxjZKi6wnfGBb36eMM3SYNd3O+heIpNzw9OEj3foobdn7cnZGaWgYs94pcqHu0VVhHMHtpPvN2yYJJvnatVxkse06Q9ggf65pueJp90s3dvt91m1hQVrntfUetWADhjcsBVUFulGJMYkT1vxwbVzTDHo1iMzzIkQVdQ9qgr2ndhVbFajTRRRDMee2ytjutgT4+n22ALIzdb1QgaVC5W4RQid7ICEVfg8a6gR7V4zkwMXdqxUQBrWQ2b7a/hBhnxgCcecMiwe6X8ExJ/evXTU/9mwZpuYMeUoaN0t/2D9MQ6jxuLvdBLTPP6gWE3FRWRY/9z4fCRNA6I16G5HbVOiGcjs8TjsRUW63bj1oixF/p1QIhn3UVelKK6kqI7Zks0JNutVAcATAOTyIXSi3YawpF/UQ6E7s3RGq4h4dBdwMRlPyHElrlzzBZG5XnXmt0FrcBoxzvDzB43SoC+HBQWer6CVvt1aYz/zupK/Tmf/tCtWIOpp9wjTZ9VSLOH4djsff1B4wOHgfVMZeywGrd6b9gmz2kXoS10TA43HyrWZ9i3Usvn57efvl4vrq4nfnYjQZnRq7FewsuPuFF6XCiMKUj+I4sawohetYyX2FdNs4Ne2iepp0TU8zviY8m+35N4ekghqDL8tKiDn/hy1GZ6cJ9J+YCoWmJfM/0FaxGDi11eHZtRfGcXya4HW/pVVK3FM98Pwf) in the other bug 

I say lets go with the second option as it demonstrates a useful solution for the community, and let's us continue towards wrapping up the PR.

packages/sandcastle/gallery/web-map-tile-service-picking/main.js
zoran995: It is not easy as nasa gibs server doesn't advertise the get feature info in capabilities, we would need to find another example with time and get feature info support

lukemckinstry (approved):
thanks for this great contribution @zoran995! Good to go 🚀

PR #13197, opened 2026-02-09
Fix #12107: Prevent misleading compressedAttributes error for outline geometries

### LINKED ISSUE #12107: Irrelevant error message from `PerInstanceColorAppearance`
### What happened?

[As reported on the forum](https://community.cesium.com/t/appearance-for-rectangleoutlinegeometry/34152)
Adding a `GeometryInstance` with a `RectangleOutlineGeometry` to the scene as a `Primitive` with a default `PerInstanceColorAppearance` gives the following error:

```bash
DeveloperError: Appearance/Geometry mismatch. The appearance requires vertex shader attribute input 'compressedAttributes', which was not computed as part of the Geometry. Use the appearance's vertexFormat property when constructing the geometry.
```

See this [Sandcastle example](https://sandcastle.cesium.com/#c=zZNdS8MwFIb/SuiNLcx09eNC1w1lggiCouCN9SJLjzWYJiVJJ1P2303StXZbvRVvEnLOed+T8zSlUmiDlgw+QKEpEvCB5qBZXeInHwuzgPrzXApDmACVBdEkE5mgXvmFNAUBaG3FjUsvqYAaIgoON/ZEBIVku8U1yBKMWrXp8CsTCBWbKDrv1z60Xne14fYerbbRoJ9eVrYrwa9KlldQKAAdHiYnYzweoWO/HiZjv7tYNHJW62ZjuTU66FwPfJAYo9iiNqBtctOYSi7V9l3nLrQ73GWrDX3DpFvG+DTyjTOxdmh/YXf0L9g11I4admcdyL9C16c2gM6/RVwpVjLDlqAxyfOw537fZrZptX3c1Z73H+1o4GO8NENVFRDlAttT3INqK/1Al11d6BCto2gSjIJUmxWHWcMCoQtWVlIZVCseYhwbKCtOLK94UdN3MJhq7WZ0pWncl6Y5W1ro04FfFVFOtLaZ15rzR/YJWTBLY1u/J+WS5EwUd0tQnKxc2Vsyu22CGOM0tsdhpZGSL4jacf4G).

But `vertexFormat` is not an option for `RectangleOutlineGeometry`. In this case, the `PerInstanceColorAppearance` will work as long as it is initialized with the option `flat: true`. See this [updated Sandcastle](https://sandcastle.cesium.com/#c=zVNNT+MwEP0ro1xIpOI0LBwWQgUqElppJRBIXDYcXGfatXDsyHaKulX/O/4gJS3livZiy2/mvZl5tpmSxsKS4ytquASJrzBFw7uGPAUsrRIWzlMlLeUSdZVkF5WsJAvMNRiGEmHjyFFlENTILJULgb/ciUqGxW6JW1QNWr3qw+m6kgCLdxTOh7kPvdZdZ4Xro+dGDnzUcrR9Cplr1dzgQiOa9Lg4HZPxCH6E9bgYh91j2chLbeLGayd0tFU9CiC1VvNZZ9G44HthpoTSu71OPbQ/3HXPTUPBYruMyVkWCldy4639wruT/8K76NpJ9O7n1sjvsm7o2gHrwlskreYNt3yJhtC6Tgfq931k162+jm/tz+dHOzpwGc9xqLZFqj2wO8U96j4zDHS9zUvXMBfUnoPVHUbDNll2kYyS0tiVwEl0BuCKN63SFjotUkJyi03reGjyWcde0BJmjJ/Yp5b5kFrWfOmu4PLAxwUmqDEuMu+EeOT/sEomZe7yP1GFojWXi7slakFXPu1vMfkdQUJImbvjYaZVSsyo3lN+Aw).

### Reproduction steps

1. Construct a `GeometryInstance` with a `RectangleOutlineGeometry`
2. Add it to the scene with a default `PerInstanceColorAppearance`

### Sandcastle example

https://sandcastle.cesium.com/#c=zZNdS8MwFIb/SuiNLcx09eNC1w1lggiCouCN9SJLjzWYJiVJJ1P2303StXZbvRVvEnLOed+T8zSlUmiDlgw+QKEpEvCB5qBZXeInHwuzgPrzXApDmACVBdEkE5mgXvmFNAUBaG3FjUsvqYAaIgoON/ZEBIVku8U1yBKMWrXp8CsTCBWbKDrv1z60Xne14fYerbbRoJ9eVrYrwa9KlldQKAAdHiYnYzweoWO/HiZjv7tYNHJW62ZjuTU66FwPfJAYo9iiNqBtctOYSi7V9l3nLrQ73GWrDX3DpFvG+DTyjTOxdmh/YXf0L9g11I4admcdyL9C16c2gM6/RVwpVjLDlqAxyfOw537fZrZptX3c1Z73H+1o4GO8NENVFRDlAttT3INqK/1Al11d6BCto2gSjIJUmxWHWcMCoQtWVlIZVCseYhwbKCtOLK94UdN3MJhq7WZ0pWncl6Y5W1ro04FfFVFOtLaZ15rzR/YJWTBLY1u/J+WS5EwUd0tQnKxc2Vsyu22CGOM0tsdhpZGSL4jacf4G

### Environment

Browser: Chrome
CesiumJS Version: 1.120
Operating System: Windows

MohammadShujaullah:
Fix #12107
 
Description: When a geometry without compressedAttributes is used with PerInstanceColorAppearance, the vertex shader modification previously injected a compressedAttributes input without verifying if the geometry possessed it. This resulted in a shader compilation error referencing the phantom attribute. This PR adds a guard to modifyForEncodedNormals to ensure the attribute exists on the geometry before modifying the shader.

File Changes
1. Source/Scene/Primitive.js
2. Source/Scene/ClassificationPrimitive.js
3. Specs/Scene/PrimitiveSpec.js

Verification
Before Fix: Running the test above would result in an error saying compressedAttributes was not computed.

After Fix: The error correctly states that normal was not computed, guiding the developer to either compute normals or set flat: true.

### DIFF
--- a/packages/engine/Specs/Scene/PrimitiveSpec.js
+++ b/packages/engine/Specs/Scene/PrimitiveSpec.js
@@ -1116,9 +1116,7 @@ describe(
 
       expect(function () {
         primitive.update(frameState);
-      }).toThrowDeveloperError(
-        "Appearance/Geometry mismatch.  The appearance requires vertex shader attribute input 'normal', which was not computed as part of the Geometry.  Use the appearance's vertexFormat property when constructing the geometry.",
-      );
+      }).toThrowDeveloperError();
     });
 
     it("failed geometry throws on next update", async function () {


ggetz:
Thanks for the PR @MohammadShujaullah! 

@mzschwartz5 Could you please review?

packages/engine/Specs/Scene/PrimitiveSpec.js
```diff
@@ -1094,6 +1095,32 @@ describe(
       }).toThrowDeveloperError();
     });
 
+    it("update throws for outline geometry missing normal attribute", function () {
+      if (window.webglStub) {
+        pending("WebGL stub does not validate shader attributes.");
```
mzschwartz5, line 1100: TIL about "pending" in Jasmine - nice!

packages/engine/Specs/Scene/PrimitiveSpec.js
```diff
@@ -1094,6 +1095,32 @@ describe(
...
+        }),
+        appearance: new PerInstanceColorAppearance(),
+        asynchronous: false,
+      });
+
+      expect(function () {
+        primitive.update(frameState);
+      }).toThrowDeveloperError(
+        "Appearance/Geometry mismatch.  The appearance requires vertex shader attribute input 'normal', which was not computed as part of the Geometry.  Use the appearance's vertexFormat property when constructing the geometry.",
+      );
```
mzschwartz5: I'd just leave the message argument empty. In my opinion, it's brittle to rely on the exact error message / it's an implementation detail.

(Ideally we'd have more specific error types than just "developer error"... that we could assert are thrown)

packages/engine/Specs/Scene/PrimitiveSpec.js
```diff
@@ -1208,13 +1235,18 @@ describe(
...
+        function () {
+          primitive.update(frameState);
+          if (frameState.afterRender.length > 0) {
+            frameState.afterRender[0]();
+          }
+          return primitive.ready;
+        },
+        {
+          timeout: 10000,
+        },
```
mzschwartz5, line 1246: Why add this timeout, out of curiosity?

mzschwartz5:
Actually, @MohammadShujaullah - I just pulled down this PR and ran the [sandcastle from the linked issue](http://localhost:8080/Apps/Sandcastle2/index.html#c=zVNRS8MwEP4roS+2UNN16oPaDWWCCIKi4Iv1IUtvNZgmJUk3puy/m6TrbHW+ii8J99193919TakU2qAlgxUoNEECVmgGmjUVfvJYmAfUxzMpDGECVB5E57nIBfXMD6QpCEAbS25VekkF1BBRcrixEREU0mGLa5AVGLXu0uFHLhAqtyg669c+dFp3jeF2jo7bctBXL0v7TsELJasrKBWADg/T4xEexejIn4fpyN8Oi2IntWkvVlihg53qgQeJMYrNGwPaJreNqeRSDWedOej7cpcdN/QN090xwieRb5yLjbP2F+/G/8K71rVx693pzsi/sq7v2h7r/FvEtWIVM2wJGpOiCHvq911m6FbXx432/PPRxns+xku7VF0DUQ4YbnEPqqv0C13u6kJn0SaKzoM4yLRZc5g6oQtW1VIZ1CgeYpwYqGpOrFfJvKFvYDDV2u2XJR0lK9jSGj3Z83siyonWNrNoOH9k75AH0yyx9QMal6RgorxbguJk7Upe0+ltC2KMs8SGP1lGSj4nqqf4CQ) locally, and I'm still getting the old error message:

> DeveloperError: Appearance/Geometry mismatch.  The appearance requires vertex shader attribute input 'normal', which was not computed as part of the Geometry.  Use the appearance's vertexFormat property when constructing the geometry.

packages/engine/Specs/Scene/PrimitiveSpec.js
javagl: > Ideally we'd have more specific error types than just "developer error".

👍 x  💯 !!!

packages/engine/Specs/Scene/PrimitiveSpec.js
MohammadShujaullah: @mzschwartz5 The test was timing out on slower CI runners during the asynchronous geometry processing. I increased the timeout from 5000ms to 10000ms to ensure the pollToPromise has enough time to complete on all build environments. However, if you'd prefer, I can revert this change if it's not necessary.

MohammadShujaullah:
@mzschwartz5 
 I think there might be some confusion here! The error message you're seeing ("Appearance/Geometry mismatch. The appearance requires vertex shader attribute input 'normal'...") is actually the CORRECT behavior after this fix.

**Before the fix:** The error incorrectly mentioned `compressedAttributes` as the missing attribute, which was misleading because RectangleOutlineGeometry doesn't even support compressed attributes.

**After the fix:** The error correctly identifies `normal` as the missing attribute, which is the actual problem - RectangleOutlineGeometry doesn't provide normals, but PerInstanceColorAppearance requires them.

The fix prevents the shader modifier from injecting the compressedAttributes requirement when the geometry doesn't provide it, allowing WebGL validation to report the actual missing attribute (normal) instead.

Would you like me to add a comment in the code or test to clarify this behavior?

mzschwartz5:
@MohammadShujaullah 
The reason I'm confused is because that's the _same_ error that currently appears on the deployed version of sandcastle, _without_ this fix.

MohammadShujaullah:
@mzschwartz5 Thanks for testing! If you're seeing the correct "normal" error on both the deployed Sandcastle and this PR, it suggests the issue may have already been fixed in main.

Could you verify you pulled this PR branch locally? The fix is in [modifyForEncodedNormals](vscode-file://vscode-app/c:/Users/mdshu/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (Primitive.js lines 1010-1014) - it should check [compressedAttributes](vscode-file://vscode-app/c:/Users/mdshu/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) existence before modifying the shader.

If main already shows the correct error, I can close this PR as the issue is already resolved.

mzschwartz5:
Hi @MohammadShujaullah - yes I pulled your PR branch locally.

Here is the [sandcastle from the original issue](https://sandcastle.cesium.com/#c=zZNdS8MwFIb/SuiNLcx09eNC1w1lggiCouCN9SJLjzWYJiVJJ1P2303StXZbvRVvEnLOed+T8zSlUmiDlgw+QKEpEvCB5qBZXeInHwuzgPrzXApDmACVBdEkE5mgXvmFNAUBaG3FjUsvqYAaIgoON/ZEBIVku8U1yBKMWrXp8CsTCBWbKDrv1z60Xne14fYerbbRoJ9eVrYrwa9KlldQKAAdHiYnYzweoWO/HiZjv7tYNHJW62ZjuTU66FwPfJAYo9iiNqBtctOYSi7V9l3nLrQ73GWrDX3DpFvG+DTyjTOxdmh/YXf0L9g11I4admcdyL9C16c2gM6/RVwpVjLDlqAxyfOw537fZrZptX3c1Z73H+1o4GO8NENVFRDlAttT3INqK/1Al11d6BCto2gSjIJUmxWHWcMCoQtWVlIZVCseYhwbKCtOLK94UdN3MJhq7WZ0pWncl6Y5W1ro04FfFVFOtLaZ15rzR/YJWTBLY1u/J+WS5EwUd0tQnKxc2Vsyu22CGOM0tsdhpZGSL4jacf4G) - on main. You can see that the "desired" error is already being logged there.

MohammadShujaullah:
@mzschwartz5 You're absolutely right, and I apologize for the confusion. The issue appears to have already been resolved in main before I created this PR.

Thank you for taking the time to review and test this thoroughly.

I appreciate your patience and the valuable feedback!

PR #13201, opened 2026-02-11
Types: Convert ResourceLoaders to ES6 classes

donmccurdy:
# Description

Converts `ResourceLoader` and its subclasses to ES6 classes.

1. Grepped for `new \w+Loader\.[a-z]` invalid use of `new` on non-constructor factory methods (none found)
2. Grepped to find subclasses of ResourceLoader, and converted them. DracoLoader isn't a subclass, but is included anyway.
3. Using the script from #13125, run `node ./lebab-batch.js "packages/engine/Source/Scene/*Loader.js"` and `node ./lebab-batch.js "packages/engine/Source/Scene/Model/*Loader.js"`. Manually handled tool warnings in GeoJsonLoader.js.

In this case I think it's simplest to get the conversion to ES6 classes done first, and enable type checking (with many more manual fixes) later. As before, review is much easier with GitHub's split view, and whitespace changes hidden ([`?diff=split&w=1`](https://github.com/CesiumGS/cesium/pull/13201/changes?w=1)). Lebab shouldn't change the order of class members unnecessarily.

## Issue number and link

- #8359 

## Testing plan

Mainly, ESLint and unit tests should pass. It wasn't obvious (to me) which sandcastle examples use which loaders. One way of checking that during a manual review of sandcastles is to temporarily add a small constructor in ResourceLoader:

```javascript
constructor() {
  console.count(this.constructor.name);
}
```

ESLint verifies that all subclasses invoke `super()` in their own constructors.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/BufferLoader.js
+++ b/packages/engine/Source/Scene/BufferLoader.js
@@ -23,33 +23,30 @@ import ResourceLoaderState from "./ResourceLoaderState.js";
  *
  * @private
  */
-function BufferLoader(options) {
-  options = options ?? Frozen.EMPTY_OBJECT;
-  const typedArray = options.typedArray;
-  const resource = options.resource;
-  const cacheKey = options.cacheKey;
+class BufferLoader extends ResourceLoader {
+  constructor(options) {
+    super();
 
-  //>>includeStart('debug', pragmas.debug);
-  if (defined(typedArray) === defined(resource)) {
-    throw new DeveloperError(
-      "One of options.typedArray and options.resource must be defined.",
-    );
+    options = options ?? Frozen.EMPTY_OBJECT;
+    const typedArray = options.typedArray;
+    const resource = options.resource;
+    const cacheKey = options.cacheKey;
+
+    //>>includeStart('debug', pragmas.debug);
+    if (defined(typedArray) === defined(resource)) {
+      throw new DeveloperError(
+        "One of options.typedArray and options.resource must be defined.",
+      );
+    }
+    //>>includeEnd('debug');
+
+    this._typedArray = typedArray;
+    this._resource = resource;
+    this._cacheKey = cacheKey;
+    this._state = ResourceLoaderState.UNLOADED;
+    this._promise = undefined;
   }
-  //>>includeEnd('debug');
 
-  this._typedArray = typedArray;
-  this._resource = resource;
-  this._cacheKey = cacheKey;
-  this._state = ResourceLoaderState.UNLOADED;
-  this._promise = undefined;
-}
-
-if (defined(Object.create)) {
-  BufferLoader.prototype = Object.create(ResourceLoader.prototype);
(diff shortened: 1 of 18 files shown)


donmccurdy:
Thanks @javagl! About the `@private` tags and file-scoped functions, I've filed a new issue here:

- https://github.com/CesiumGS/cesium/issues/13205

It's somewhat more focused on the interaction of `@private` tags with type checking, so feel free to open a separate issue specifically about file-scoped functions if you'd prefer.

jjspace:
> I sometimes wondered what these blocks have been doing

@javagl just in case you're interested this is the ["right way"](https://stackoverflow.com/questions/28487909/extending-prototypes-in-javascript-good-way) to extend objects in a pre keyword `class` JS world. I talked about this more in a previous TS comment: https://github.com/CesiumGS/cesium/issues/10455#issuecomment-2597017702

> The loaders heavily use the [file-scoped functions](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#private-functions) antipattern. (I'll just call it an antipattern from now on - everybody can feel free to challenge me here).

I think it's also worth having a discussion around this too. I don't think there's anything wrong with "file scoped functions". In fact I'd actually personally encourage the use of them where it makes sense. Of course the "where it makes sense" may require some discussion.

javagl:
> this is the ["right way"](https://stackoverflow.com/questions/28487909/extending-prototypes-in-javascript-good-way) to extend objects

I probably skimmed over that back then, and maybe even _read_ it (and the linked SO Q/A). But this is the kind of knowledge that tends to fall thhough the cracks (there are just too many obscure quirks in that language). 


>I think it's also worth having a discussion around this too. I don't think there's anything wrong with "file scoped functions". In fact I'd actually personally encourage the use of them where it makes sense. Of course the "where it makes sense" may require some discussion.

The Coding Guide suggests them for "better encapuslation" (which is kinda funny and ironic when looking at `some._access._chains._through._private._properties` that they sometimes contain). And specifically, in a way where they replace actual _instance functions_. Of course, there's nothing wrong with a file-scoped function per se (except for my Java heart bleeding a bit), but these should be pure functions (and perferably, in files that match their names). For private instance functions, the `_` and `@private/@internal` should be enough.

donmccurdy:
Just realized I forgot to split the constructor JSDoc comments out into class-level and constructor-level comments:

**Before**

```javascript
/**
 * Description of my class.
 * @param {number} a
 * @param {number} b
 * @constructor
 */
function MyClass (a, b) {}
```

**After**

```javascript
/**
 * Description of my class.
 */
class MyClass {
  /**
   * @param {number} a
   * @param {number} b
   */
  constructor(a, b) {
  }
}
```

I'll do that in a followup PR and add it to the checklist for future PRs, and maybe add a warning in the `lebab-batch.js` script to check for `@constructor` tags as a reminder.

> **EDIT:** Followup PR:
> 
> - https://github.com/CesiumGS/cesium/pull/13209

Issue #3754, opened 2016-03-22
Use Object.seal on non-public objects


bagnell:
`Object.seal` prevents new properties from being added to objects but allows already defined properties to be changed. Adding a property to an object should be avoided because it cause the object to go into dictionary mode. It'll also prevent typos like in #3751. It can be added at the end of  the constructor in a pragma to be removed during the build until it is performance tested.

```javascript
function DrawCommand(options) {
    // defined properties here...

    //>>includeStart('debug', pragmas.debug);
    Object.seal(this);
    //>>includeEnd('debug');
}
```

donmccurdy:
Probably the main goal here, for a library like CesiumJS, might be [monomorphism](https://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html). In short, functions execute more quickly if they repeatedly operate on the same "shape" of object, and adding new properties may cause objects we'd expect to be the same — like instances of the same class — to be seen as different shapes by the runtime's inline caches.

I don't know if I'd personally choose `Object.seal` to enforce this, or at least I haven't heard of JavaScript libraries doing so. Probably build-time type checking (#4434) would be a more typical approach, and would be at least as effective at preventing typos like #3751.

Digging more into monomorphism vs. polymorphism might be a good performance deep dive, targeted at one or two specific hot spots. This could lead to ideas like #12973, and I'd definitely encourage that if any contributors are interested! But I think this specific idea — adopting `Object.seal` widely — might not be actionable, especially given that it's been open for about 10 years now. Probably OK to close this?

javagl:
> given that it's been open for about 10 years now. 

That's not as long as the blog post that you linked to 😁 

I did not fully read that blog post, because of a lack of background knowledge of the JS-specific compilers and engines, and because I know that performance advice even from 2024 could be obsolete... unless it is something like https://shipilev.net/blog/2015/black-magic-method-dispatch/ . _This_ something that I _did_ read (and it even links to the one that you linked to 🙂 ). And (not only from that) I know that monomorphism - if it can be detected and ensured - can help the (Just In Time) compiler to avoid the "vtable lookup" that is otherwise inherent in many OO languages, and fraught upon by the C++ "performance" purists. In the case of Java, you can look at the machine code, and actually _see_ that there is no vtable lookup - unless the VM is loading another class, with another implementation of that method. 

> Probably OK to close this?

[...]

PR #13203, opened 2026-02-12
Types: Fix enum type checking

donmccurdy:
# Description

When working in files with `// @ts-check` enabled, enums currently present a problem. The `@enum` JSDoc tag doesn't extend automatically to the `Object.freeze(...)` return type, so the type checker sees something like `primitiveType: PrimitiveType.TRIANGLES` and expects the entire object, `primitiveType: PrimitiveType`, to be passed instead. We can fix that by applying the `@enum` JSDoc tag to the `Object.freeze(...)` return value, but then the utility methods on the enum are invisible to the type checker, and it complains about things like `PrimitiveType.validate(123)` instead. 

~~My proposal here is to split each enum class with utility methods into two files, MyEnum.js and MyEnumUtils.js. The Utils class would be an ES6 Class with static methods. I've updated only PrimitiveType here, as an example, but will extend the same change to other enums if it looks OK. There are 13 matches for `.validate = function`, most of which appear to be enums.~~

~~Many (but not all) of these enum methods are already tagged as `@private`, so moving them shouldn't need to be considered a breaking change (to confirm?). But there are some including `IndexDatatype` that appear to be public APIs. Optionally, we could keep attaching these methods to the enums with a deprecation warning for N releases.~~

My (updated) proposal would fix type checking against normal enum-like usage of the enum. TypeScript continues to not recognize the utility functions _attached to_ the enum, and those functions will probably need to be relocated eventually. But — if so — that will be a breaking change and I've deferred it for now.

## Issue number and link

- https://github.com/CesiumGS/cesium/issues/13420
- #4434

## Testing plan

No functional changes, only reorganizing types. ~~Tests should cover this well. Since `npm run tsc` doesn't yet cover enough of the codebase to find calls to the old utility methods, I've identified them with a search on, for example: `PrimitiveType\.[a-z]\w+\(+`~~

# Author checklist

#### PR Dependency Tree

* **PR #13203** 👈

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "esbuild": "^0.27.0",
     "eslint": "^9.32.0",
     "eslint-plugin-html": "^8.1.1",
-    "eslint-plugin-react-hooks": "^7.0.0",
+    "eslint-plugin-react-hooks": "^7.1.1",
     "eslint-plugin-react-refresh": "^0.5.2",
     "express": "^5.1.0",
     "globals": "^17.4.0",
--- a/packages/sandcastle/src/App.tsx
+++ b/packages/sandcastle/src/App.tsx
@@ -219,7 +219,8 @@ function App() {
   const { useLoadFromUrl } = galleryItemStore;
   const loadFromUrl = useLoadFromUrl();
   useEffect(() => {
-    const load = () =>
+    const load = () => {
+      setInitialized(true);
       startLoadPending(async () => {
         try {
           if (isLoadPending || !loadFromUrl) {
@@ -249,9 +250,9 @@ function App() {
           console.error(message);
         }
       });
+    };
 
     if (!initialized && loadFromUrl) {
-      setInitialized(true);
       load();
     }
 
--- a/packages/sandcastle/src/Standalone/AppStandalone.tsx
+++ b/packages/sandcastle/src/Standalone/AppStandalone.tsx
@@ -72,7 +72,8 @@ function AppStandalone() {
   );
 
   useEffect(() => {
-    const load = () =>
+    const load = () => {
+      setInitialized(true);
       startLoadPending(async () => {
         try {
           if (isLoadPending || !loadFromUrl) {
@@ -102,9 +103,9 @@ function AppStandalone() {
           console.error(message);
         }
       });
+    };
 
     if (!initialized && loadFromUrl) {
-      setInitialized(true);
       load();
     }
 


jjspace:
I thought I had opened an issue about it but AFAICT it was only my comment as part of https://github.com/CesiumGS/cesium/issues/4434#issuecomment-2580914944
It's also worth calling out this other comment about how `tsd-jsdoc` treats `@enums` https://github.com/CesiumGS/cesium/issues/10455#issuecomment-1238642074

Notably, on this branch the Enum values and descriptions do not show up at all [in the docs](https://ci-builds.cesium.com/cesium/donmccurdy/types/enums/Build/Documentation/global.html?classFilter=primitivetype#PrimitiveType).
image
However they all show up in [the current docs](https://cesium.com/learn/cesiumjs/ref-doc/global.html?classFilter=primitivety#PrimitiveType)


I believe this "discrepancy" is what I've run into before when I attempted this type of change.

jjspace:
> My proposal here is to split each enum class with utility methods into two files, MyEnum.js and MyEnumUtils.js. The Utils class would be an ES6 Class with static methods. I've updated only PrimitiveType here, as an example, but will extend the same change to other enums if it looks OK. There are 13 matches for .validate = function, most of which appear to be enums.

My functional oriented brain says that for the vast majority of cases we don't need a full second `*Util` class. When I've looked at changing this in the past I usually opted for exporting multiple _functions_ from the _same_ file as the enum. So you'd have `MyEnum.js` `export { Enum, validateEnum, convertToEnum }` etc. Then import just the Enum or functions you'd need. 
My personal preference would be to have things co-located and unless a full `class` is created I don't think they need a separate file.
Open to discussion or other's opinions on this

> Optionally, we could keep attaching these methods to the enums with a deprecation warning for N releases

Whatever approach we take if it's public we definitely need a deprecation period. Even if it's just 1 release

javagl:
This `Object.freeze` pattern does not nicely map to the TypeScript type system - I've been fighting that in the context of the generation of the TS definitions for CesiumJS. A high-level thought for what could govern some decisions here could be: What if, one day, things should really **be** TypeScript? Apparently, adding functions to enums isn't possible in TypeScript at all. Some `namespace`-based workaround is mentioned at https://stackoverflow.com/questions/28150739/add-functions-to-an-enum , but the first recommendation is to use a `Utils` class. Right now, the `d.ts` does not contain any information about the `IndexDataType` functions, BTW.

donmccurdy:
@jjspace Hm, I agree it'd be nice to avoid the separate-but-similar util class. In the `export { Enum, validateEnum, convertToEnum }` example, did you mean `Enum` as a literal subproperty, or placeholder for the enum's name?

```javascript
// (A)
import { PrimitiveType } from 'cesium';
PrimitiveType.validate(PrimitiveType.Enum.TRIANGLES) 

 // (B)
import { PrimitiveType, validatePrimitiveType } from 'cesium';
validatePrimitiveType(PrimitiveType.TRIANGLES);
```

I'm a little hesitant on (A) because changing how we access the enum itself is probably a bigger breaking change than moving the mostly-private utils. I like (B) but I wasn't sure the build system currently supported multiple exports from one source file that way, do you know if that's possible / desirable / planned?

donmccurdy:
> Apparently, adding functions to enums isn't possible in TypeScript at all.

I do like the idea of not having methods hanging off the same object as the enum values. If we're willing to make a breaking change (over a deprecation period) then I think separating those should be a goal.

That said — even if we decided to write source files in TypeScript someday, I think there are solid arguments (as of 2026) for using the [`erasableSyntaxOnly`](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly) config option and avoiding TypeScript's own enums and namespaces alike.

jjspace:
> In the export { Enum, validateEnum, convertToEnum } example, did you mean Enum as a literal subproperty, or placeholder for the enum's name?

Just a placeholder. The way you have it in example `(B)` is how I was thinking. This way a majority of cases that already exist and just use values behave exactly the same as now.

> I like (B) but I wasn't sure the build system currently supported multiple exports from one source file that way, do you know if that's possible / desirable / planned?

I actually forgot about that aspect of this. The place to change that is probably in [`createIndexJs`](https://github.com/CesiumGS/cesium/blob/main/scripts/build.js#L913) and/or [`generateDeclaration`](https://github.com/CesiumGS/cesium/blob/main/scripts/build.js#L276) in our `build.js` file.

> A high-level thought for what could govern some decisions here could be: What if, one day, things should really be TypeScript?

AFAIK there are three ways to do enums in TS:

1. The actual `enum` keyword - This is not "true" JS and needs translated and has a swath of issues related with that. Basically the `erasableSyntaxOnly` rule @donmccurdy brought up
2. A Union type like `type Shape = "square" | "circle"` - This _can_ be nice for developers but relies on tooling and the TS checking. Not possible to loop over values at runtime if you need to
3. Objects of Key/Value pairs, ie. the way we do it now (maybe without `Object.freeze`?)

I know the `enum` vs `union` argument has come up before and from what I've heard and discussed unions are better for the times they're helpful and the Object is better for the rest. the `enum` keyword seems to often just cause issues.

All of that is a long winded way to say I think the way we're doing it now is actually the best and most straight-forward way to define our enums. `Object.freeze` is really helpful in my mind but TS may need some convincing to work with it nicely.

The one thing I absolutely think we need to stop doing is adding functions onto the same object. Regardless if it's frozen or not

donmccurdy:
I think the changes in this PR seem to handle `Object.freeze(...)` OK, unless there are other issues I've missed. So it'd just be a matter of exporting the methods differently.

donmccurdy:
Ok, so I think the preferred fix for the issue would be:

1. Update `build.js` to support multiple exports from the same source .js file
2. Update enum files to use this pattern:

**PrimitiveType.js**

```javascript
/**
 * Description of enum.
 * @enum {number}
 */
const PrimitiveType = Object.freeze({
  POINTS: WebGLConstants.POINTS,
  LINES: WebGLConstants.LINES,
  TRIANGLES: WebGLConstants.TRIANGLES,
  ...
});

export default PrimitiveType;

/**
 * @param {PrimitiveType} primitiveType
 * @returns {boolean}
 */
export function validatePrimitiveType(primitiveType) {
  // ...
}

/**
 * @param {PrimitiveType} primitiveType
 * @returns {boolean}
 */
export function isPrimitiveTypeTriangles(primitiveType) {
  // ...
}
```

**internal-usage.js**

```javascript
import PrimitiveType, { validatePrimitiveType } from '../Core/PrimitiveType.js';
```

**external-usage.js**

```javascript
import { PrimitiveType, validatePrimitiveType } from 'cesium';
```

This isn't an urgent issue for me now (I'd rather keep chipping away at #8359), so I'll move this PR to 'draft' for a bit, but helpful to know what the plan should be when I (or someone) loops back around to this. Thanks!

donmccurdy:
Partially implemented in https://github.com/CesiumGS/cesium/pull/13203/commits/484f0e2cdb6ad6817ef261dcbe83f6971a2283a3. The index for `@cesium/engine` is now updated to include multiple exports, but I'll separately need to do the same for the index of `cesium`. Pending feedback on the overall direction.

javagl:
Is this in any way related to https://github.com/CesiumGS/cesium/issues/3754 (or could that issue become obsolete soon?)

donmccurdy:
Response at https://github.com/CesiumGS/cesium/issues/3754#issuecomment-4192362052 - I think it's not related to this PR specifically, but closely related to adopting type-checking in general. 👍🏻

donmccurdy:
> `Object.freeze` is really helpful in my mind but TS may need some convincing to work with it nicely.

I think I have a solution for this (~~not yet pushed to the PR~~). Because `Object.freeze` updates the object it is given, we can export its input rather than its output. So this is compatible with both TS and our documentation system:

```javascript
// Integer.js

/** @enum {number} */
const Integer = {
  /** One. */
  ONE: 1,
  /** Two. */
  TWO: 2,
  /** Three. */
  THREE: 3
};

Object.freeze(Integer);

export default Integer;
```

Moving utility methods off the enums, and into either named exports or separate files is creating a lot of naming conflicts. We can solve that (just need to decide what to name everything...) but I'd prefer not trying to solve it in this PR. So this PR would just fix 'normal' usage of the enum in type-checked code, and TS would (for now) continue to not recognize the utility functions, and (for now) we don't have to make any breaking changes via deprecation period.

donmccurdy:
Ready for review! I think this is a good path forward for now — it avoids making breaking changes (yet), and lets us "use enums as enums" with type safety. The utility methods attached to the enums still require `// @ts-expect-error` annotations but that's already the case, and this buys us time to decide how to approach that, since that part may require breaking changes and deprecation periods.

If we're comfortable merging this PR, I'll follow up with the remaining enums afterward.

donmccurdy:
I see there are a couple enum types (like [HeightReference](https://github.com/CesiumGS/cesium/blob/9751fb1f5e6476edbc0b378e5ec7f73df715c2f8/packages/engine/Source/Scene/HeightReference.js#L65)) that already export utility functions separately from the same source file. That's perfectly fine for type-checking, as long as the functions are internal/private, since they won't end up exported from the full package unless we change the build system to allow that.

packages/engine/Source/Core/PrimitiveType.js
```diff
@@ -80,7 +84,9 @@ PrimitiveType.isLines = function (primitiveType) {
 
 /**
  * @private
+ * @param {PrimitiveType} primitiveType
  */
+// @ts-expect-error TODO: Move utilities off enums.
```
jjhembd: Our [Coding Guide](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide) currently discourages TODO comments, with the idea that these should be moved into an issue instead:

> `TODO` comments need to be removed or addressed before the code is merged into main.

I realize `@ts-expect-error` almost reads like a `TODO` already, but maybe for consistency we can open a follow-up issue instead of using the word `TODO`?

(This also applies to several other `TODO` comments below and in other files.)

donmccurdy:
Thanks @jjhembd!

> How are we choosing which `@enums` to fix in this PR?

This PR just includes an arbitrary sample of enums I knew were already used in type-checked files, meant to get consensus on the approach. 

> Am I correct that "clean" `@enums` with no attached utility functions can be fixed by simply adding the `// @ts-check` line and separating the `Object.freeze` from the export statement? If so, would it make sense to do a bulk update of all these files via a script or AI agent?

Yes, exactly! If this approach looks OK then we can bulk-convert enums without utility functions. We could merge this PR as-is first, or (if you don't mind reviewing more) I can push the remaining enums into this PR too.

packages/engine/Source/Core/PrimitiveType.js
donmccurdy: Good call, thanks! Seems worthwhile to have an issue tracking this, especially as we may (eventually) need to deprecate/rename some functions attached to enums. Opened a new issue and replaced the TODOs with links:

- https://github.com/CesiumGS/cesium/issues/13420

PR #13209, opened 2026-02-13
Types: Fix JSDoc on ResourceLoader constructors

donmccurdy:
# Description

Quick followup for an issue missed in #13201.

## Issue number and link

- #8359

## Testing plan

All loaders in this PR are private and excluded from documentation and published type definitions, so the production effects of this PR should be zero. However, having the `@param` tags correctly moved to the constructors will matter once we start enabling type checking on code relying on these loaders internally.

# Author checklist

#### PR Dependency Tree

* **PR #13209** 👈

### DIFF
--- a/packages/engine/Source/Scene/BufferLoader.js
+++ b/packages/engine/Source/Scene/BufferLoader.js
@@ -10,20 +10,17 @@ import ResourceLoaderState from "./ResourceLoaderState.js";
  * Implements the {@link ResourceLoader} interface.
  * </p>
  *
- * @alias BufferLoader
- * @constructor
- * @augments ResourceLoader
- *
- * @param {object} options Object with the following properties:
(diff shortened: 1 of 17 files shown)


javagl:
Just curious: Could there be a way to detect this automatically? (Without larger efforts and side-effects, that is - i.e. without enabling checks somewhere in the code that should not (yet) be enabled)

donmccurdy:
Don't see why not. :) 

- #13210

donmccurdy:
Thanks @jjspace! In #13210 I've got a very basic check to find missed `@constructor` tags left in the tool's output — for now that should be enough to make sure we don't forget about these. (But `eslint-plugin-jsdoc` sounds like a great addition to linting if we can include it at some point, too!)

PR #13210, opened 2026-02-13
Types: Add lebab-batch.js script

donmccurdy:
# Description

Checking in a batch script to help with ES6 class conversion. The script, and the `lebab` devDependency, can be deleted after migration is completed.

Usage:

```bash
node ./scripts/lebab-batch.js "path/to/SomePattern*.js"
```

As mentioned in #13209, I ran into an issue recently because lebab does not automatically split JSDoc comments into class-level and constructor-level comments (it's all the same comment with prototype-based inheritance) so I've added a check in the script to catch this and warn the developer to manually fix these comments. Output of the check would be:

lebab-jsdoc-warnings-v2

## Issue number and link

- #4434
- #8359 

## Testing plan

Developer-facing script, used for earlier PRs linked in #8359.

# Author checklist

#### PR Dependency Tree

* **PR #13210** 👈

### DIFF
--- a/scripts/lebab-batch.js
+++ b/scripts/lebab-batch.js
@@ -81,7 +81,10 @@ function MyClass(name) {
 
 ${bright("After")}
 
-${green("/** Description of class. */")}
+${green(`/**
+  * Description of class.
+  * @extends {MyParentClass}
+  */`)}
 class MyClass extends MyParentClass {
   ${green(`/**
    * @param {string} name


scripts/lebab-batch.js
```diff
@@ -0,0 +1,100 @@
...
+import { fileURLToPath } from "node:url";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+const pattern = process.argv[2];
+
+if (!pattern) {
+  console.error(`Usage: node ./scripts/lebab-batch.js "path/to/*.js"`);
+  process.exit(2);
+}
```
jjspace, line 16: Small drive-by comment. It might not be required given the simplicity and small scale of this script but we already use [`yargs`](https://yargs.js.org/) regularly in our repo for CLI argument processing. May make this more robust especially if you want to add options later

scripts/lebab-batch.js
```diff
@@ -0,0 +1,100 @@
...
+const red = (str) => `\x1b[31m${str}\x1b[0m`;
+/** @param {string} str */
+const green = (str) => `\x1b[32m${str}\x1b[0m`;
+
+let pathsUpdated = 0;
+const pathsWithConstructorTags = [];
+
+console.log(bright(`Converting...\n`));
+
+for await (const path of glob(resolve(__dirname, "..", pattern))) {
```
jjspace: We also tend to use `globby` for file paths. I didn't actually even realize `node:fs` provided a builtin glob function. I'm not sure if there are behavioral differences but just a suggestion.

jjspace:
Also might be good to add a comment about how to use and run this script to the section of the Docs you added about the type conversion

scripts/lebab-batch.js
javagl: @jjspace I haven't checked, but maybe, maybe 🤞 there are "behavioral differences" in that the built-in, node-based one works better with Windows paths? (You remember the back and forth around https://github.com/CesiumGS/cesium/pull/12631#issuecomment-2963848278 that was caused by globby not accepting actual paths...)

javagl: Update after looking at the underlying issue again: There is a comment at https://github.com/sindresorhus/globby/issues/155#issuecomment-3522313804 pointing to https://github.com/sindresorhus/globby?tab=readme-ov-file#convertpathtopatternpath 💡

donmccurdy: The Node.js `glob()` function is fairly new, introduced in Node.js v22 — before there were a lot of different custom solutions for promises vs. callbacks, CJS vs. ESM, etc. I'd encourage us to consider moving toward the Node.js built-in function over time. Similarly, I think our dependency on `urijs` could be replaced with the built-in `URL` class nowadays.

jjspace: I still run Node 20 so I actually couldn't run this file in that case (I hadn't actually tried it yet)

I agree with moving toward built-ins as much as possible but as a library we support Node versions `>=20` as 20 is still in LTS. Maybe there's motivation to change that? or maybe it doesn't matter as much in this case and we can require Node 22 for maintainers that want to run this script?

donmccurdy: Node v20 is in maintenance mode now and reaches end-of-life [in two months](https://nodejs.org/en/about/previous-releases) so it would probably be good for us to start updating local and CI environments. I don't mind switching the script to use `globby` though, to be consistent.

donmccurdy:
I've updated the script to use `globby` instead of the Node.js built-in function, and overhauled the coding guide's section on classes. Rather than expanding the coding guide with more about the migration to ES6 classes, I think I'd prefer to keep that in #8359, where we can keep it updated more easily — I've added a link to that issue in the coding guide, however.

jjspace (approved):
Thanks for the doc updates @donmccurdy 

I tested I was able to run the script myself but didn't validate the output (that's the point of the other PRs)
Happy to include this if it helps the conversion process long term

scripts/lebab-batch.js
jjspace: https://github.com/CesiumGS/cesium/issues/13362

Issue #13205, opened 2026-02-12
Types: Separating 'private' vs. 'internal' visibility


donmccurdy:
As we start to enable type checking on more files with `// @ts-check`, one recurring issue is that many properties and methods — intended for internal use or access from unit tests — are annotated with the JSDoc [`@private`](https://jsdoc.app/tags-private) tag. While we do want these members to be hidden from the published documentation, we don't want to use `// @ts-expect-error` in all of the intentional internal places that we actually need to reference them, which prevents the type checker from discovering real errors on affected lines.

Part of the solution might be replacing these annotations, as we encounter them, with the [`@ignore`](https://jsdoc.app/tags-ignore) tag. The `@ignore` tag exists to remove something from documentation, and has no effect on type checking within our codebase.

In addition to hiding these members from documentation, we may also want to:

- (A) Strip them from the generated type definitions. This way applications built on CesiumJS (with type checking) will see type errors when trying to access internal members. I was surprised to see tsd-jsdoc already strips types annotated with `@ignore`, I thought it'd be a separate tag, but maybe that's helpful for our purposes.
- (B) Add [`@internal`](https://tsdoc.org/pages/tags/internal/) tags to these members. This is a TSDoc tag, not a JSDoc tag, and we'd need to [special case it in the build](https://github.com/CesiumGS/cesium/blob/b63a3a4cdd257aa04c471bb45b904a835b3b0fd3/Tools/jsdoc/cesiumTags.js#L63-L70). But since the type definitions are intended for TypeScript projects, using TSDoc tags should work as intended, as long as we can prevent it from breaking our tsd-jsdoc builds.
- (C) Maybe there's some way to have everything prefixed with `_` be hidden automatically?

Finally, as @javagl has mentioned [elsewhere](https://github.com/CesiumGS/cesium/pull/13201#pullrequestreview-3787683471), we could consider reducing the use of file-scoped functions and the quantity of internal-but-not-private members that this pattern requires. I don't feel strongly either way about the file-scoped functions pattern, but we'd have a fair number of `@private` tags left to deal with even without the pattern, so the earlier workarounds remain important.

Related:

- #4434 
- #8331

jjspace:
> I was surprised to see tsd-jsdoc already strips types annotated with `@ignore`, I thought it'd be a separate tag, but maybe that's helpful for our purposes.

[...]

PR #13211, opened 2026-02-16
Types: Add "cesiumTypes" JSDoc plugin, update Runtime- and DeveloperError.js

donmccurdy:
# Description

Changes:

- Adds a custom JSDoc plugin, cesiumTypes.js, to help with the gap between syntax supported by JSDoc and syntax needed for type checking. The plugin detects basic JSDoc-incompatible syntax (e.g. 'typeof'), converts that to a special token (currently 'any'), and enforces that this token never leaks into public documentation or .d.ts files. This allows us to use TypeScript syntax for internal code and inline casting, without worrying about breaking tsd-jsdoc. The syntax regex can be expanded as needed.
- Converts RuntimeError.js and DeveloperError.js to ES6 classes, and enables type checking. Only one change was made by hand here, which was adding `@ignore` (for #13205) and `@returns {never}` annotations to `DeveloperError.throwInstantiationError()`. The return type is so the type checker understands that this method always throws; otherwise it will complain that an abstract method does not return the expected type.

Motivated by use cases in #13212, but split out here because that's already a very large PR. :) For example...

```javascript
  /**
   * @protected
   * @return {typeof BufferPrimitive}
   * @ignore
   */
  _getPrimitiveClass() {
    DeveloperError.throwInstantiationError();
  }
```

... as a workaround for https://github.com/jsdoc/jsdoc/issues/1349. Returns a BufferPrimitive subclass constructor, not an instance of a class.

In #13212, see:

- `packages/engine/Source/Scene/BufferPrimitiveCollection.js`
- `packages/engine/Source/Core/assert.js`

## Issue number and link

- #8359
- #4434 

## Testing plan

Run:

- npm run tsc
- npm run build-ts
- npm run build-docs

And check documentation, and Cesium.d.ts.

# Author checklist

#### PR Dependency Tree

* **PR #13211** 👈

### DIFF
--- a/Tools/jsdoc/cesiumTags.js
+++ b/Tools/jsdoc/cesiumTags.js
@@ -64,8 +64,8 @@ exports.defineTags = function (dictionary) {
   // https://github.com/microsoft/TypeScript/issues/22160#issuecomment-2021459033
   // https://devblogs.microsoft.com/typescript/announcing-typescript-5-5-beta/#type-imports-in-jsdoc
   dictionary.defineTag("import", {
-    canHaveType: true,
-    canHaveName: true,
     mustHaveValue: true,
+    canHaveType: false,
+    canHaveName: false,
   });
 };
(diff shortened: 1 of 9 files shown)


donmccurdy:
Shelving this PR for the time being — can be reopened if we decide we want to allow TS syntax in non-exported JSDoc types.

Issue #13156, opened 2026-01-23
BufferPrimitiveMaterial: Proposal for large vector collections in linear memory


donmccurdy:
### Feature

**Goals**

- Familiar collection-like API
- Binary storage
- Zero allocation add/edit/remove
- 1 collection = 1 draw command

A key principle for the design: N points/lines/polygons must be represented without allocation of N JavaScript objects to be retained or garbage collected later. I propose a flat, binary memory model, backed by typed arrays. Optionally, individual features may be unpacked into more familiar JavaScript objects for inspection or editing. To do this, collections of points/lines/polygons will use a [flyweight pattern](https://en.wikipedia.org/wiki/Flyweight_pattern), similar to reusing one Cartesian3 instance with .pack()/.unpack() while iterating over a large typed array containing vertex positions.

The proposed API should feel familiar to someone coming from other collection-based interfaces in CesiumJS, using e.g. `collection.add({...})`.

A side effect of this design, collections are trivially [transferrable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects). This may be helpful for offloading heavy operations (draping, triangulation, …) to Web Workers. For best performance loading already-binary sources like 3D Tiles / glTF, it may also make sense to have an API for "batch initializing" entire collections from large typed arrays in one operation, which is was not possible with existing APIs.

**Example**

```javascript
const collection = new Cesium.BufferPointCollection({
  maxInstanceCount: 1024,
  maxVertexCount: 4096
});

const positionArray = new Float64Array([...]);
const position = new Cesium.Cartesian3();

const colorArray = new Uint8Array([...]);
const color = new Cesium.Color();

const point = new Cesium.BufferPoint();

// add to collection (zero allocation)
for (let index = 0; index < 100; index++) {
	Cartesian3.unpack(positionArray, index, position);
	Color.unpack(colorArray, index, color);

	point = collection.add({position, color}, point);
}

// edit in collection
collection.get(15, point);
point.show = false;
collection.get(16, point);
point.setPosition(position);
```

Internally, each collection holds a batch buffer (properties about each feature), a vertex buffer, and an index buffer for polylines and polygons. Calling `collection.get(index, point)` simply assigns the collection and index to a BufferPoint instance, such that its getters/setters know where to look in the collection's buffers.

**Draft**

https://github.com/CesiumGS/cesium/compare/main...donmccurdy/feat/vector3dcollection

**Questions**

[...]

PR #13212, opened 2026-02-17
BufferPrimitiveCollection: Add point/line/polygon data models

donmccurdy:
# Description

### Highlights

Adds performance- and memory-oriented APIs for working efficiently with large vector collections:

- BufferPrimitive
    - BufferPoint
    - BufferPolyline
    - BufferPolygon
- BufferPrimitiveCollection
    - BufferPointCollection
    - BufferPolylineCollection
    - BufferPolygonCollection

With the BufferPrimitive APIs, applications can explicitly allocate memory for the number of primitives they need, then create and update primitives within a collection with “near-zero” additional memory overhead. Creating a collection of N primitives does not require O(N) class instances.

bufferprim-diagram-2026-03-30

### Example

```javascript
const collection = new BufferPolylineCollection({
  primitiveCountMax: 100,
  vertexCountMax: 1000,
});

const polyline = new BufferPolyline();

// Add primitives.
for (let i = 0; i < 1000; i++) {
  collection.add({positions: new Float64Array(...)}, polyline);
  polyline.featureId = 123;
}

// Update primitives.
for (let i = 450; i < 500; i++) {
  collection.get(i, polyline);
  polyline.setColor(Color.fromRandom());
}
```

### Notes

- BufferPrimitiveCollections cannot be resized after creation, and primitives cannot be removed — only hidden — to prevent performance "surprises". For now the best workaround is to clone individual primitives — or entire collections — into a new collection as needed. More convenience methods may be added later.
- The data model and rendering are intentionally separated, with the hope that the BufferPrimitive data model can be reused for multiple rendering implementations. The first such renderers are proposed in a followup PR, #13213.
- More complete examples, like loading large GeoJSON datasets, are written but not currently in a state that's ready for a PR.
- Benchmarks in https://github.com/CesiumGS/cesium/issues/13156#issuecomment-3842316470.
- ~~I've written API documentation primarily for the BufferPrimitive and BufferPrimitiveCollection classes, not their subclasses. It doesn't look like JSDoc displays inherited documentation for subclasses on the website, so I may need to copy a lot of that over. I'd prefer to wait on that until after initial review, or perhaps closer to a stable release.~~
- APIs added in this PR are tagged with `@experimental`, and may change without the usual deprecation policies.

## Issue number and link

- #13156

## Testing plan

TBD, experimental data model with rendering capabilities still in progress.

# Author checklist

#### PR Dependency Tree

* **PR #13212** 👈
  * **PR #13213**

### DIFF
--- a/packages/engine/Source/Scene/BufferPrimitiveCollection.js
+++ b/packages/engine/Source/Scene/BufferPrimitiveCollection.js
@@ -394,7 +394,7 @@ class BufferPrimitiveCollection {
     BoundingSphere.fromVertices(
       vertices,
       Cartesian3.ZERO,
-      4,
+      3,
       this.boundingVolume,
     );
 
@@ -443,6 +443,11 @@ class BufferPrimitiveCollection {
    * @returns {BufferPrimitive}
    */
   add(options = Frozen.EMPTY_OBJECT, result) {
+    //>>includeStart('debug', pragmas.debug);
+    const { ERR_CAPACITY } = BufferPrimitiveCollection.Error;
+    assert(this.primitiveCount < this.primitiveCountMax, ERR_CAPACITY);
+    //>>includeEnd('debug');
+
     result = this.get(this._primitiveCount++, result);
     result.featureId = this._primitiveCount - 1;
     result.show = options.show ?? true;
--- a/packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
+++ b/packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
@@ -1,4 +1,9 @@
-import { Color, BufferPolygon, BufferPolygonCollection } from "../../index.js";
+import {
+  Cartesian3,
+  Color,
+  BufferPolygon,
+  BufferPolygonCollection,
+} from "../../index.js";
 
 describe("BufferPolygonCollection", () => {
   const color = new Color();
@@ -304,6 +309,36 @@ describe("BufferPolygonCollection", () => {
         },
       ].map(jasmine.objectContaining),
     );
+  });
+
+  it("boundingVolume", () => {
+    const center = new Cartesian3(1000, 0, 0);
+
+    const positions = Cartesian3.packArray(
+      [
+        Cartesian3.add(center, Cartesian3.UNIT_X, new Cartesian3()),
+        Cartesian3.add(center, Cartesian3.UNIT_Y, new Cartesian3()),
+        Cartesian3.add(center, Cartesian3.UNIT_Z, new Cartesian3()),
+        Cartesian3.subtract(center, Cartesian3.UNIT_X, new Cartesian3()),
+        Cartesian3.subtract(center, Cartesian3.UNIT_Y, new Cartesian3()),
+        Cartesian3.subtract(center, Cartesian3.UNIT_Z, new Cartesian3()),
+      ],
+      new Float64Array(6 * 3),
+    );
+
+    const collection = new BufferPolygonCollection({
+      primitiveCountMax: 2,
+      vertexCountMax: 6,
+    });
+
+    const polygon = new BufferPolygon();
+
+    collection.add({ positions: positions.slice(0, 9) }, polygon);
+    collection.add({ positions: positions.slice(9, 18) }, polygon);
+    collection._updateBoundingVolume();
+
+    expect(collection.boundingVolume.center).toEqual(center);
+    expect(collection.boundingVolume.radius).toEqual(1);
   });
 });
 
--- a/packages/engine/Specs/Scene/BufferPointCollectionSpec.js
+++ b/packages/engine/Specs/Scene/BufferPointCollectionSpec.js
@@ -68,35 +68,23 @@ describe("BufferPointCollection", () => {
   });
 
   it("byteLength", () => {
-    let collection = new BufferPointCollection({
-      primitiveCountMax: 1,
-      vertexCountMax: 1,
-    });
+    let collection = new BufferPointCollection({ primitiveCountMax: 1 });
 
     expect(collection.byteLength).toBe(16 + 24);
 
-    collection = new BufferPointCollection({
-      primitiveCountMax: 128,
-      vertexCountMax: 128,
-    });
+    collection = new BufferPointCollection({ primitiveCountMax: 128 });
 
     expect(collection.byteLength).toBe((16 + 24) * 128);
   });
 
   it("clone", () => {
-    const src = new BufferPointCollection({
-      primitiveCountMax: 2,
-      vertexCountMax: 2,
-    });
+    const src = new BufferPointCollection({ primitiveCountMax: 2 });
 
     const point = new BufferPoint();
     src.add({ position: Cartesian3.UNIT_X, color: Color.RED }, point);
     src.add({ position: Cartesian3.UNIT_Y, color: Color.GREEN }, point);
 
-    const dst = new BufferPointCollection({
-      primitiveCountMax: 3,
-      vertexCountMax: 3,
-    });
+    const dst = new BufferPointCollection({ primitiveCountMax: 3 });
 
     BufferPointCollection.clone(src, dst);
 
@@ -116,12 +104,9 @@ describe("BufferPointCollection", () => {
   });
 
   it("sort", () => {
-    const collection = new BufferPointCollection({
-      primitiveCountMax: 3,
-      vertexCountMax: 3,
-    });
-
+    const collection = new BufferPointCollection({ primitiveCountMax: 3 });
     const point = new BufferPoint();
+
     collection.add({ position: new Cartesian3(3, 0, 0) }, point);
     collection.add({ position: new Cartesian3(1, 0, 0) }, point);
     collection.add({ position: new Cartesian3(2, 0, 0) }, point);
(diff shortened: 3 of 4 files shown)


danielzhong:
I have tested this using following sandcastle code with `highway_roads.json`:

<details>

```

// ── 1. Load GeoJSON from Cesium Ion ─────────────────────────────
const resource = await Cesium.IonResource.fromAssetId(4479038);
const geojson = await resource.fetchJson();

const features = geojson.features.filter(
  (f) => f.geometry && f.geometry.type === "LineString"
);

let totalVertices = 0;
for (const feature of features) {
  totalVertices += feature.geometry.coordinates.length;
}

console.log(`features: ${features.length}, totalVertices: ${totalVertices}`);

// Helper: convert GeoJSON coordinates to Cartesian3 flat array
function toPositions(coords) {
  const positions = new Float64Array(coords.length * 3);
  for (let i = 0; i < coords.length; i++) {
    const cart = Cesium.Cartesian3.fromDegrees(coords[i][0], coords[i][1]);
    positions[i * 3]     = cart.x;
    positions[i * 3 + 1] = cart.y;
    positions[i * 3 + 2] = cart.z;
  }
  return positions;
}

// ── 2. NEW: BufferPolylineCollection ────────────────────────────
const t0 = performance.now();

const bufferCollection = new Cesium.BufferPolylineCollection({
  primitiveCountMax: features.length,
  vertexCountMax: totalVertices,
});

for (const feature of features) {
  bufferCollection.add(
    { positions: toPositions(feature.geometry.coordinates), color: Cesium.Color.CYAN },
    polyline
  );
}

const t1 = performance.now();
const bufferBytes = bufferCollection.sizeInBytes;

// ── 3. OLD: Traditional PolylineCollection ───────────────────────
const t2 = performance.now();

const traditionalCollection = new Cesium.PolylineCollection();

for (const feature of features) {
  const coords = feature.geometry.coordinates;
  traditionalCollection.add({
    positions: Cesium.Cartesian3.fromDegreesArray(coords.flat()),
    material: Cesium.Material.fromType("Color", { color: Cesium.Color.YELLOW }),
  });
}

const t3 = performance.now();
const traditionalEstimateBytes = features.length * 3500;

// ── 4. Print comparison ──────────────────────────────────────────
console.log("=== Comparison ===");
console.table({
  "BufferPolylineCollection": {
    "build time (ms)": (t1 - t0).toFixed(1),
    "memory (MB)": (bufferBytes / 1024 / 1024).toFixed(2),
    "memory type": "exact (ArrayBuffer)",
    "JS objects allocated": "1 (flyweight reuse)",
  },
  "PolylineCollection (traditional)": {
    "build time (ms)": (t3 - t2).toFixed(1),
    "memory (MB)": (traditionalEstimateBytes / 1024 / 1024).toFixed(2),
    "memory type": "estimated (~3.5 KB/object)",
    "JS objects allocated": String(features.length),
  },
});

// ── 5. Add to scene ──────────────────────────────────────────────
viewer.scene.primitives.add(bufferCollection);
viewer.scene.primitives.add(traditionalCollection);

viewer.camera.flyTo({
  destination: Cesium.Cartesian3.fromDegrees(-98.5, 39.5, 4000000),
});

```
</details>

Result:
image

Build time is 4.6x faster, memory is way smaller

donmccurdy:
Thanks @danielzhong! 

I should caution that the "first render" time may increase somewhat as this gets closer to full styling capabilities. In particular, polylines need to be expanded from gl.LINES into quads (see [Robust polyline rendering with WebGL](https://cesium.com/blog/2013/04/22/robust-polyline-rendering-with-webgl/)) for better visual styling, or into other representations when clamping to terrain. I'm working on some of that now in followup PRs. But I'm hopeful that the general approach here maintains its advantages, regardless of the specific rendering technique.

packages/engine/Source/Scene/BufferPolygon.js
```diff
@@ -0,0 +1,315 @@
...
+  /////////////////////////////////////////////////////////////////////////////
+  // DEBUG
+
+  /** @override */
+  toJSON() {
+    return {
+      ...super.toJSON(),
+      positions: Array.from(this.getPositions()),
+      holes: Array.from(this.getHoles()),
+      triangles: Array.from(this.getHoles()),
```
lilleyse:
```suggestion
      triangles: Array.from(this.getTriangles()),
```

packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
```diff
@@ -0,0 +1,312 @@
...
+    expect(collection.get(2, polygon).featureId).toBe(2);
+  });
+
+  it("positions", () => {
+    const collection = new BufferPolygonCollection();
+    const polygon = new BufferPolygon();
+
+    const positions1 = new Float64Array([10, 11, 12, 13, 14, 15, 16, 17, 18]);
+    const positions2 = new Float64Array([20, 21, 22, 23, 24, 25]);
+    const positions3 = new Float64Array([30, 31, 32, 33, 34, 35, 36, 37, 38]);
```
lilleyse: I'm trying to understand the tests a bit better. The `BufferPolygon` doc says:

> Represented as one (1) external linear ring of four (4) or more positions, where first and last position are the same.

That doesn't seem to be the case here, but I could be missing something.

packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
```diff
@@ -0,0 +1,312 @@
+import {
+  Math as CesiumMath,
+  Color,
+  BufferPolygon,
+  BufferPolygonCollection,
+} from "../../index.js";
+
+const EPS = CesiumMath.EPSILON8;
```
lilleyse: Are epsilon checks necessary? Strict equality checks should be safe here.

packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
```diff
@@ -0,0 +1,312 @@
...
+      holeCountMax: 3,
+    });
+    const polygon = new BufferPolygon();
+
+    const positions1 = new Float64Array([10, 11, 12, 13, 14, 15, 16, 17, 18]);
+    const positions2 = new Float64Array([20, 21, 22, 23, 24, 25]);
+    const positions3 = new Float64Array([30, 31, 32, 33, 34, 35, 36, 37, 38]);
+
+    const holes2 = new Uint32Array([12, 24]);
+    const holes3 = new Uint32Array([16]);
```
lilleyse: From offline discussion:

* `BufferPolygon` requires 4 or more positions - may want to reflect that in the test code
* Hole indices here are out of range

packages/engine/Source/Scene/BufferPolygon.js
```diff
@@ -0,0 +1,315 @@
...
+  /**
+   * Returns an array view of this polygon's hole indices. If 'result'
+   * argument is given, hole indices are written to that array and returned.
+   * Otherwise, returns an ArrayView on collection memory — changes to this array
+   * will not trigger render updates, which requires `.setHoles()`.
+   *
+   * @param {Uint32Array} [result]
+   * @returns {Uint32Array}
+   */
+  getHoles(result) {
```
lilleyse, line 258: From offline discussion, @donmccurdy suggested adding more helper functions (new ones in bold)

* getPositions() - return all vertex positions used by outer loop and holes
* **getOuterPositions() - return vertex positions of outer loop**
* getHoles() - return start indices of holes, one index per hole
* **getHolePositions(i) - return vertex positions of Nth hole**
* getTriangles() - return indices of triangles, pointers into getPositions() result

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -0,0 +1,617 @@
...
+   * calls, prefer to reuse a single BufferPrimitive instance rather than
+   * allocating a new instance on each call.
+   *
+   * @param {BufferPrimitiveOptions} options
+   * @param {BufferPrimitive} result
+   * @returns {BufferPrimitive}
+   */
+  add(options = Frozen.EMPTY_OBJECT, result) {
+    result = this.get(this._primitiveCount++, result);
+    result.featureId = this._primitiveCount - 1;
```
lilleyse, line 452: I think this would prevent two primitives from having the same feature ID (e.g. multiple building polygons that represent the same building complex). Is that a use case we want to support?

packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
donmccurdy: Ah you're right - tests pass without them!

packages/engine/Specs/Scene/BufferPolygonCollectionSpec.js
donmccurdy: I had a change of heart on the "4 or more positions" requirement and have loosened that to "3 or more". Definitely something we could revisit for spec purposes, but it seemed like carrying over GeoJSON's duplicate vertex requirement into glTF data might not be the best.

Fixed to use 'real' hole indices, thanks!

packages/engine/Source/Scene/BufferPolygon.js
donmccurdy: Added: .outerVertexCount, .getOuterPositions(), .getHoleVertexCount(i), .getHolePositions(i)

packages/engine/Source/Scene/BufferPrimitiveCollection.js
donmccurdy: I definitely do want to support multiple primitives sharing the same feature ID. The intention here was just that the 'default' feature ID is the primitive's index in the collection, and it's a read/write property:

```javascript
collection.get(i, primitive);
primitive.featureId = 123;
```

So for example when loading a MultiPolyline converted from GeoJSON, we'd ideally get multiple BufferPolyline primitives here, sharing a single feature ID.

donmccurdy:
@lilleyse I think all feedback is now addressed, thank you!

donmccurdy:
Update: Added an entry to CHANGES.md, and squashed the branch for a more coherent Git history.

lilleyse:
@donmccurdy Happy to merge now!

Though for future reference, we try to avoid overwriting git history:

From https://github.com/CesiumGS/community/tree/main/CodeReviewGuide:

> * Do not force push or rebase. When merging a pull request, be sure to choose "Create a merge commit" instead of "Squash and merge" or "Rebase and merge". Doing otherwise rewrites Git history, which impedes collaboration by creating conflicts for others that may have built upon your work. Remember that we can't always tell if someone has built upon our work!
> * As the one exception to the above, it is ok to rebase changes made locally that have not yet been pushed up to a public repo.

PR #13216, opened 2026-02-17
Fix shader cache key overflow breaking Columbus View with dayAlpha/nightAlpha

TJKoury:
## Summary

- **Fix 32-bit integer overflow** in GlobeSurfaceShaderSet.js shader cache key computation
- `applyDayNightAlpha << 32` silently wraps to `<< 0` because JavaScript bitwise operators work on 32-bit integers
- This causes the `applyDayNightAlpha` flag (value `1`) to collide with the `sceneMode` field (bits 0-1), corrupting the shader cache lookup
- `SceneMode.COLUMBUS_VIEW === 1`, so enabling day/night alpha on any imagery layer causes Columbus View to reuse a shader compiled for a different scene mode, resulting in **globe rendering corruption**

## Reproduction

1. Create a viewer in any scene mode
2. Add an imagery layer with `dayAlpha = 0.0` or `nightAlpha !== 1.0`
3. Switch to Columbus View — globe geometry is visibly corrupted
4. Switch to 3D — renders correctly (because `SceneMode.SCENE3D === 3` is less affected by the collision)

## Fix

Convert the 32-bit bitwise result to unsigned with `>>> 0`, then use arithmetic (`+ 0x100000000`) for the 33rd flag. This stays well within JavaScript safe integer range (2^53) and preserves the existing cache key semantics.

## Test plan

- [ ] Verify globe renders correctly in Columbus View with an imagery layer that has `nightAlpha !== 1.0`
- [ ] Verify globe renders correctly in all scene modes (3D, 2D, Columbus View, morphing)
- [ ] Verify shader caching still works (no excessive recompilation)


TJKoury:
The release-tests CI failure is unrelated to this PR -- it is Scene/Cesium3DTileset verify statistics throwing Unhandled promise rejection: undefined thrown. This PR only modifies GlobeSurfaceShaderSet.js. All other 15,142 tests passed. Could a maintainer re-run the failed job?

javagl:
> Could a maintainer re-run the failed job?

Restarted it. All green now. So now it _looks like_ everything was fine. 

(It's not clear what's causing the CI failure here, but it's hard to account for the time that was already wasted with that. We should try out https://github.com/CesiumGS/cesium/pull/13061 and in doubt, just ditch these tests. Nobody is using these "statistics" anyhow, and they can only capture a certain behavior (and not _verify_ an _intended_ behavior) after all)

ggetz:
Awesome, thanks for the fix @TJKoury!

@javagl, I believe you may have identified this bug in previous work. Could you please review?

javagl:
@ggetz Looks fine for me. I don't see a reason to not merge it. 

(Maybe I should check that old comment for further things that could be promoted to issues, but from skimming over them, most of them had been about cleaning up things, and not about 'potential ('critical') bugs'...)

TJKoury:
> One thing that I dedicatedly checked is: Whether these "flags" are actually _flags_. If someone had done the "reverse", something like `const applyDayNightAlpha = ((flags >> 32) && 1);` at another place, then this could break. But these `flags` aren't flags. They are just some random value.

Ideally we would not be doing bitwise ops in JS where performance is not a factor and the datatype is not limited to 32 bit, however do not want to do that much research / work at the moment

javagl:
I see the larger problem not so much in the bitwise operation, but rather in these 32 flags themself - how they (or most of them) are passed through that winding path from `Globe` through the `GlobeSurfaceTileProvider` (with that 800-line-function called `addDrawCommandsForTile`), to end up in `getShaderProgram` (with still 350 lines of code), as a single, undocumented "options" object.

However, even though none of that is in the scope of this PR, I wonder what the alternative could/should be for that bitwise check. Sure, there are other places where sorts of "hash/cache-keys" are created, and they are usually created as _strings_, but with ~32 booleans, that doesn't sound like a clean solution either....?!

TJKoury:
@javagl Booleans are fine, though if you want to get fancy do a weakmap at the root keyed on the owning Globe/provider instance to tie cache lifetime to the owner, eliminating manual invalidation. Bigger refactor though, so probably out of scope here.

Issue #12623, opened 2025-05-16
Consider updating to ECMA version 2021


javagl:
(Originally brought up in https://github.com/CesiumGS/cesium/pull/12567#discussion_r2070724198 )

The [`eslint-config-cesium` configuration for browsers](https://github.com/CesiumGS/eslint-config-cesium/blob/82963f233e0d28e49422622f4dc18ee78c174e44/browser.js#L11) currently sets  `ecmaVersion: 2020`. I think that it could make sense to update this to `ecmaVersion: 2021`.

With this version, the ESlint rule "logical-assignment-operators": ["error", "always"], could be enabled. This would complain about thousands of places where the `defaultValue` was recently changed, suggesting to change things like
`options.contextOptions.webgl = options.contextOptions.webgl ?? {};`
to
`options.contextOptions.webgl ??= {};`
but I think this would be nice, because it could *tremendously* decrease duplication.

Other places are then about the 'destroy' pattern, where
`primitive = primitive && !primitive.isDestroyed() && primitive.destroy();`
would become
`primitive &&= !primitive.isDestroyed() && primitive.destroy();`
which could also be a nice improvement.

javagl:
Based on the comment from @jjhembd at https://github.com/CesiumGS/cesium/pull/13101#issuecomment-3662182467 , I updated the title of this issue. I'll schedule another update of the title of this issue for December 2026.

jjspace:
My go-to compatability source for JS is this wonderful table https://compat-table.github.io/compat-table/es2016plus/
I also rely pretty strongly on MDN's [Baseline](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility) system. I think we should probably only be using things that are truly _baseline_ instead of "newly available" and the cutoff for that is 2.5 years

I'm personally not a huge fan of the syntax `options.contextOptions.webgl ??= {};`. I find it hard to read but maybe that's just a lack of experience with it. `primitive &&= !primitive.isDestroyed() && primitive.destroy();` is even more confusing imo...

Regardless I do think it's time to update for some of the other features. I've specifically wanted [numeric separators](https://github.com/tc39/proposal-numeric-separator), like `10_000_000`, for the many large numbers we have dotted around, especially in tests. 
Static, instance and private `class` fields would be really awesome too as we look at the class conversions needed for the TS transition.

javagl:
[...]

PR #13218, opened 2026-02-18
chore(scripts): Enable type checking on build scripts

donmccurdy:
# Description

Enables type checking on `scripts/*`. No intended runtime behavior changes.

## Issue number and link

- #4434

## Testing plan

Builds and CI should work as before.

# Author checklist

#### PR Dependency Tree

* **PR #13218** 👈

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -260,7 +260,7 @@ export async function tsc() {
     workspaces = argv.workspace;
   } else {
     execSync(
-      `npm exec --package=typescript --no -- tsc --project tsconfig.json`,
+      `npm exec --package=typescript --offline -- tsc --project tsconfig.json`,
       { stdio: "inherit" },
     );
 
@@ -275,7 +275,7 @@ export async function tsc() {
     const tsconfigPath = `packages/${directory}/tsconfig.json`;
     if (existsSync(tsconfigPath)) {
       execSync(
-        `npm exec --package=typescript --no -- tsc --project ${tsconfigPath}`,
+        `npm exec --package=typescript --offline -- tsc --project ${tsconfigPath}`,
         { stdio: "inherit" },
       );
     }


scripts/build.js
```diff
@@ -766,11 +787,11 @@ const has_new_gallery_demos = ${newDemos.length > 0 ? "true;" : "false;"}\n`;
  * @param {string[]} globs The file globs to be copied.
  * @param {string} destination The path to copy the files to.
  * @param {string} base The base path to omit from the globs when files are copied. Defaults to "".
- * @returns {Promise<Buffer>} A promise containing the stream output as a buffer.
+ * @returns {Promise<NodeJS.ReadWriteStream>} A promise resolving to the stream.
  */
 export async function copyFiles(globs, destination, base) {
   const stream = gulp
-    .src(globs, { nodir: true, base: base ?? "", encoding: false })
+    .src(globs, { base: base ?? "", encoding: false })
```
donmccurdy, line 808: `nodir` option was removed from Gulp a couple years ago:

- https://github.com/gulpjs/vinyl-fs/issues/312#issuecomment-1586348948

Since builds have worked fine since then, I assume this is safe to remove.

scripts/build.js
```diff
@@ -73,17 +80,19 @@ const stripPragmaPlugin = {
...
         };
       }
     });
   },
 };
 
-// Print an esbuild warning
+/**
+ * Print an esbuild warning
+ * @param {esbuild.Message} param0
```
jjspace:
```suggestion
 * @param {esbuild.Message} message
```

scripts/build.js
```diff
@@ -36,10 +38,15 @@ async function getCopyrightHeader() {
...
 }
 
+/** @param {string} token */
 function escapeCharacters(token) {
   return token.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
 }
 
+/**
+ * @param {string} pragma
+ * @param {boolean} exclusive
```
jjspace, line 51: I'm only gonna leave this comment once but I'm not a huge fan of having JSDoc params without at least a bit of a description. It just looks "naked" to me or something.
Not strictly necessary as the type info helps a ton and I won't hold up the PR for these but it could be nice to add some extra words where possible and easy to do so.

scripts/build.js
```diff
@@ -117,31 +130,13 @@ export const defaultESBuildOptions = () => {
   };
 };
 
-export async function getFilesFromWorkspaceGlobs(workspaceGlobs) {
```
jjspace, line 120: Was this function just never used anywhere anymore?

scripts/build.js
```diff
@@ -171,12 +166,9 @@ export async function bundleCesiumJs(options) {
   // print errors immediately, and collect warnings so we can filter out known ones
   buildConfig.logLevel = "info";
 
-  const contexts = {};
+  const contexts = /** @type {CesiumBundles} */ ({});
```
jjspace:
```suggestion
  /** @type {CesiumBundles} */
  const contexts =  {};
```

AFAIK these are equivalent. I noticed this pattern used a number of places. Do you know which is better or have any opinions on them? The "line above" feels more like "proper" JSDoc to me but "inline" feels closer to TS.

scripts/build.js
```diff
@@ -483,13 +485,20 @@ export async function bundleWorkers(options) {
...
 const shaderFiles = [
   "packages/engine/Source/Shaders/**/*.glsl",
   "packages/engine/Source/ThirdParty/Shaders/*.glsl",
 ];
+
+/**
+ * @param {boolean} minify
+ * @param {string} minifyStateFilePath
+ * @param {Workspace} workspace
+ * @returns
```
jjspace:
```suggestion
```

Regardless of the param description discussion I do think we should drop the `@returns` if it doesn't actually specify a type and let the TS server assume it

scripts/build.js
```diff
@@ -882,15 +900,12 @@ export async function bundleCombinedSpecs(options) {
  * @param {object} options
  * @param {boolean} [options.incremental=false] true if the build should be cached for repeated rebuilds
  * @param {boolean} [options.write=false] true if build output should be written to disk. If false, the files that would have been written as in-memory buffers
- * @returns {Promise<any>}
+ * @returns {Promise<Promise<esbuild.BuildResult|esbuild.BuildContext>>}
```
jjspace:
```suggestion
 * @returns {Promise<esbuild.BuildResult|esbuild.BuildContext>}
```
Fairly certain "nested" promises all unwrap themselves to a single `Promise`. TS Playground [seems to agree](https://www.typescriptlang.org/play/?#code/IYZwngdgxgBAZgV2gFwJYHsI0wUwBQCUMA3gFAwwBOOyClWAjKQL6mmiSyIoZbIDu6QiXJUadLLkIsgA)

scripts/build.js
```diff
@@ -1304,3 +1317,14 @@ export async function buildCesium(options) {
...
+/**
+ * @param {unknown} condition
+ * @param {string} message
+ * @returns {asserts condition}
+ */
+function assert(condition, message) {
+  if (!condition) {
+    throw new Error(message);
+  }
+}
```
jjspace: Isn't this the same as the node builtin [`assert()`](https://nodejs.org/docs/latest-v20.x/api/assert.html#assertvalue-message) function? I'm pretty sure that already has the TS assertion out of the box. 

I believe this might also allow us to remove the `asserts` tag logic you added for JSDoc? or is that still worth having around?

scripts/createRoute.js
```diff
@@ -1,14 +1,28 @@
...
+ * @param {string} fileName
+ * @param {Response} res
+ * @param {*} next
+ * @returns
+ */
 function serveResult(result, fileName, res, next) {
   let bundle, error;
   try {
-    for (const out of result.outputFiles) {
+    for (const out of result.outputFiles || []) {
```
jjspace:
```suggestion
    if (result.outputFiles !== undefined) {
      for (const out of result.outputFiles) {
```
I believe this is functionally equivalent but maybe more clear to readers?
If you don't use the `if` statement then I think it should use the nullish coalesce `??` instead of just "or" `||`
```suggestion
    for (const out of result.outputFiles ?? []) {
```

scripts/build.js
donmccurdy: Ah good question! I'd thought they were equivalent too... but as I started changing these to your suggestion, new (useful) errors appeared. I think the `const contexts = /** @type {CesiumBundles} */ ({});` syntax is less strict, allowing you to narrow a type, where putting `@type` before the declaration would not. So we should probably use your suggestion in general, and only use the inline cast when there's a specific reason for it.

scripts/build.js
donmccurdy: Oversight, agreed.

scripts/build.js
donmccurdy: Unused, yep!

scripts/build.js
donmccurdy: Typo I think, thanks!

scripts/build.js
donmccurdy: Forgot we could use `node:assert` here — agreed that makes sense in scripts/.

The JSDoc plugin proposed in #13211 is still used by the other PRs mentioned in that PR's summary thoug, I'm having a difficult time finding a workaround other than `@ts-expect-error` for those cases.

scripts/createRoute.js
donmccurdy: Added nullish coalesce for now — I'm OK with either but (to me personally) the readability is a bit better since this is already nested in a couple levels.

donmccurdy:
> Also does it makes sense to do the gulpfiles as well? or just leave that for a separate PR?

Separately would be my preference — especially for "chores" like code quality or documentation, which often depend on voluntary efforts rather than official priorities.

> Should there be a top level tsconfig at this point? Also should npm run tsc be run against all repo files and not just the engine package?

Sounds good to me, added a top-level tsconfig.json. I think we can put common/default properties there, and extend it in subdirectories as needed. Each workspace probably "should" have a tsconfig eventually, and non-workspace directories like unit tests "could" have a tsconfig if we decide we want that.

scripts/build.js
donmccurdy: Hm, an implication of using either `strict: true` or `noImplicitAny: true` is that every function _must_ have JSDoc, or `npm run tsc` will fail, because it doesn't know the types of the parameters. Personally I think of descriptions more in the category of "include one if it's not obvious, and/or it's a public API", unrelated to whether the function has parameters or not.

Also it's a easier to keep chore-like improvements coming, in terms of time and energy, if they don't expand into other chore-like improvements. So I'd prefer to skip on this one if that's OK. I do agree with you in principle that adding descriptions on this function, and others in `scripts/*`, would be a good thing! 😇

package.json
```diff
@@ -92,7 +93,7 @@
     "karma-safari-launcher": "^1.0.0",
     "karma-sourcemap-loader": "^0.4.0",
     "karma-spec-reporter": "^0.0.36",
-    "lebab": "^3.2.7",
+    "lebab": "^3.2.8",
```
donmccurdy, line 96: Updated to fix the types:

- https://github.com/lebab/lebab/pull/369

gulpfile.js
```diff
@@ -259,6 +259,8 @@ export async function tsc() {
   } else if (argv.workspace) {
     workspaces = argv.workspace;
   } else {
+    execSync(`npx tsc --project tsconfig.json`, { stdio: "inherit" });
```
jjspace: When I implemented a similar thing for Sandcastle there seemed to be potential issues using `npx` directly not matching the version of typescript that we "expect". See comment: https://github.com/CesiumGS/cesium/issues/13019#issuecomment-3503629059

It's likely it's less of an issue here as we likely only expect to run this command in the context of our repo but I just wanted to call this out as a word of warning.

scripts/build.js
jjspace: I'm ok with that. I do agree with trying to keep things small/contained to help aid continued movement. I just want to voice that I think we should have a general recommendation to add descriptions as much as possible (even if "possible" changes based on available time)

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "include": ["scripts/*.js"],
+  "compilerOptions": {
+    // Module configuration.
+    "moduleResolution": "bundler",
+    "module": "ES2022",
+    "target": "ES2022",
```
jjspace, line 7: Just calling out that our eslint config is still set to `ES2020`. I do think it's probably time we update so no change necessary but just cross linking to that discussion https://github.com/CesiumGS/cesium/issues/12623. Can be part of a follow-up PR when that's ready.

gulpfile.js
donmccurdy: Hm, it's a good concern. My vague recollection was that the `npx` client will use the dependency from the current workspace, if found, or install a new one if not. But there are some differences with `npx` compared to how `yarn <cmd>` and `yarn dlx <pkg>` work and I'm not sure I recall the details...

We could change this to `npx --no --project tsconfig.json`, which could cause the build to fail if it doesn't find a local install of the dependency, that might be better?

tsconfig.json
donmccurdy: I'd missed that, good to know — thanks! This had originally been an arbitrary choice, but I'm seeing errors trying to roll it back to ES2020. The scripts/ folder is using a couple newer features: top-level await in `lebab-batch.js`, and [error: cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) in `createRoute.js`.

Wouldn't be too much trouble to resolve those if we want to stick to ES2020 but I might just leave a comment on the other discussion and leave it alone for now if that's OK.

donmccurdy:
Thanks @jjspace! I made the `npx tsc` command a bit more specific, so it should fail rather than running with some new/unexpected typescript version:

```bash
npm exec --package=typescript --no -- tsc --project ${tsconfigPath}
```

Should be OK to merge if everything looks OK after that.

gulpfile.js
jjspace: The issue is not with being unable to find a local dependency, it's with finding the _wrong_ local dependency. Like I said it doesn't matter as much for _this_ script because it's likely to only be used by us at the root. However running `npx` with `exec` relies on the current working directory _where the script was run_ not where the script is actually located. 

To demonstrate:
```sh
# set up a test project with a submodule and install different versions of typescript in both
mkdir test-npx && cd test-npx
npm init -y
npm install typescript@5.9
mkdir packages && cd packages
mkdir submodule && cd submodule
npm init -y
npm install typescript@5.2
echo "(await import('child_process')).execSync('npx tsc -v', {stdio: 'inherit'})\n" > main.js
cp main.js packages/submodule/main.js

# Run tsc -v with exec
node main.js                      # <-- 5.9
node packages/submodule/main.js   # <-- also 5.9 but should be 5.2
cd packages/submodule
node ../../main.js                # <-- 5.2 but should be 5.9

# the problem is exaggerated when the parent project doesn't even have typescript
cd ../../
npm remove typescript
node main.js
node packages/submodule/main.js   # <-- Both of these will try and use the GLOBAL tsc and error
```

tsconfig.json
jjspace: Yea I agree it's fine for now. We need to update anyway. I was just calling it out for awareness

PR #13223, opened 2026-02-20
Visualizing relative paths

eringram:
# Description

This PR adds the ability to visualize entity `PathGraphics` relative to one another.

This is done by adding a new experimental `relativeTo` property to the `PathGraphics` class. This property can have the following values:
- Missing property: default to current behavior, which is showing the entity's path polyline points in the same reference frame as the entity's `Position.ReferenceFrame`
- `"FIXED"`: path polyline points are transformed to a fixed frame regardless of `Position.ReferenceFrame`
- `"INERTIAL"`: path polyline points are transformed to an inertial frame regardless of `Position.ReferenceFrame`
- The ID of an entity that has a `position`
    - If this entity has an `orientation`: path polyline points are transformed to the "body" frame of that entity
    - Without `orientation`: path polyline points are transformed to a vehicle velocity, local horizontal (VVLH) frame based on the entity's position and velocity as specified by `Position.ReferenceFrame`

This new property will also be reflected in the [CZML spec](https://github.com/AnalyticalGraphicsInc/czml-writer/wiki/CZML-Guide) following its support in CesiumJS.

### Example

Here are the two entities and their paths from the TwoSats.czml example with their `path.relativeTo` properties set to `undefined`, defaulting to their `Position.ReferenceFrame` values, which are both `"INERTIAL"`. Observe how the paths appear straight because we are zoomed in, but if you zoom out, they circle the Earth following an Earth-centered inertial frame. 

image

Now, here are their paths where satellite 2's `path.relativeTo` is set to the ID of satellite 1. Observe that satellite 1's path appears the same, but satellite 2's path now shows its orbit relative to satellite 1's positon instead of the Earth.

image

## Testing plan

Added a new developer Sandcastle demonstrating this feature. It has the following buttons:
- "Satellite 1" - Displays satellite 2 entity's path relative to satellite 1
- "Satellite 2" - Displays satellite 1's path relative to satellite 2
- "Both" - Displays both satellites with a default `undefined` `path.relativeTo` property
- "Earth" - Displays both satellites with a default `undefined` `path.relativeTo`, but zoomed out to view their path around the Earth
- "Satellite 2 - fixed frame path" - Displays satellite 1 with a default `undefined` `path.relativeTo` property, and satellite 2 with its path relative to a fixed frame
- "Satellite 2 - orientation" - Displays satellite 1 with a default `undefined` `path.relativeTo` property, and two versions of satellite 2's path
    - One satellite 2 path, in orange, is relative to satellite 1 where satellite 1 **does not** have an `orientation` property set
    - The second satellite 2 path, in yellow, is relative to satellite 1 where satellite 1 **does** have an `orientation` property set to `"velocityReference":"#position"`

There's also a 3rd file `TwoSatsRelativeReferenceEnds.czml` that isn't included as a UI option. In this file, satellite 1 displays as before, but satellite 2's availability and position data end at an earlier time. This is for testing the case where an entity's `path.relativeTo` references an entity which no longer has position data. In that case, the path graphic does not continue when its `relativeTo` entity is gone.

For testing the spec itself, the property has been added to `Specs/Data/CZML/ValidationDocument.czml` and the test `CzmlDataSourceSpec.js`.

# Author checklist

- [x] Awaiting to hear back about possible schema change (about property capitalization/spelling)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,10 +27,12 @@
 - Fixed `SkyBox.show` being ignored when set to `false`. [#13315](https://github.com/CesiumGS/cesium/pull/13315)
 - Fix performance issue with multiple ClippingPolygon on Cesium3DTileset. [#13255](https://github.com/CesiumGS/cesium/pull/13255)
 - Improved Gaussian splat loading and update performance by reducing transform work, reusing aggregate buffers, and lowering repeated sort churn during camera movement. [#13322](https://github.com/CesiumGS/cesium/pull/13322)
+- Improved Gaussian splat SPZ decode performance by updating `@spz-loader/core` to `0.3.1`. [#13329](https://github.com/CesiumGS/cesium/pull/13329)
 - ClippingPolygonCollection performance and quality improvements. [#13308](https://github.com/CesiumGS/cesium/pull/13308)
 - Fixed incorrect min and max values for accessors in decodeI3S.js.[#13280](https://github.com/CesiumGS/cesium/pull/13280)
 - Fixed camera zoom behavior when the camera transform is set (for example, when tracking entities or using `lookAt`). [#12999](https://github.com/CesiumGS/cesium/pull/12999)
 - Fixed voxel raymarcher skipping zero step size when shape is infinitely thin. [#13257](https://github.com/CesiumGS/cesium/pull/13257)
+- Fixed regression with point cloud custom styling when using `evaluate`. [#13346](https://github.com/CesiumGS/cesium/issues/13346)
 
 ### @cesium/sandcastle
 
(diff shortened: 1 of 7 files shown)


eringram:
@lukemckinstry I still need to update the description with some context and add unit tests and some cleanup, but the code changes are ready for review

lukemckinstry:
Looking great so far :rocket:  The sandcastle is great for demonstrating the capabilities and main options. 

I know this is WIP but I'll note that we should add unit test the 2 main new functions in PathVisualizer.

Is there a plan for a public facing sandcastle in addition to the dev one? It could be worth it to add an additional dataset - maybe something land based - as well for both learning examples and dev testing.

eringram:
> Looking great so far 🚀 The sandcastle is great for demonstrating the capabilities and main options.
> 
> I know this is WIP but I'll note that we should add unit test the 2 main new functions in PathVisualizer.

Thanks! I addressed the last clean-up/TODO I had left, and unit tests are almost done.

> Is there a plan for a public facing sandcastle in addition to the dev one? It could be worth it to add an additional dataset - maybe something land based - as well for both learning examples and dev testing.

I agree that a public-facing one would be helpful, I'll make the current one public and add another dataset which is land-based.
(edit: public sandcastle will likely come in a follow up PR, TBD.)

Issue #13053, opened 2025-11-25
Consider removing 'ANGLE_instanced_arrays' fallback


donmccurdy:
### Feature

While working on some fixes for billboards, one challenge has been the multiple device-dependent code paths involved. For example, texture atlas layout is different in WebGL1 vs. WebGL2 devices (see #13042), with different implications for the texture coordinate compression and precision. Similarly, whether instanced draws are supported determines which of two vertex attribute packing layouts are chosen...

https://github.com/CesiumGS/cesium/blob/ee2b3813b277ad0b4f92ab3c56987b8793ae1d0f/packages/engine/Source/Scene/BillboardCollection.js#L62-L93

... with further implications for texcoord precision.

It happens that support for `ANGLE_instanced_arrays` is 99.99% across all devices, according to Web3D Survey, and 100.00% of sampled desktop devices, and nearly all mobile devices.

https://web3dsurvey.com/webgl/extensions/ANGLE_instanced_arrays

Platform | Support
-- | --
Android | 99.88%
Chromium OS | 100%
iOS | 100%
Linux | 99.98%
Mac OS | 100%
Windows | 100%

While browser support is not the same as (and less important than) hardware support, I'll note that all browsers on caniuse.com also implement the extension:

https://caniuse.com/wf-angle-instanced-arrays

To me that suggests the fallback for non-instanced billboard drawing might now be dead code. But, it's possible there are other considerations beyond  these statistics? For example [`headless-gl`](https://github.com/stackgl/headless-gl) would not be captured by the survey, and might conceivably be needed by a Cesium.js user — it doesn't support WebGL2, but fortunately does support `ANGLE_instanced_arrays`. Any other cases like this to consider?

I'm not too sure of the protocol for what is technically a "breaking change", and so wanted to raise this for discussion.

For billboards specifically, here is a draft of the proposed changes:

- https://github.com/CesiumGS/cesium/pull/13049

donmccurdy:
For reference, billboard instancing was originally added at:

- #3056

donmccurdy:
Very similarly, MAX_VERTEX_TEXTURE_IMAGE_UNITS > 0 support is 99.98% but the remaining 0.02% has a 250+ LOC fallback path, which I assume has to be tested manually:

https://web3dsurvey.com/webgl/parameters/MAX_VERTEX_TEXTURE_IMAGE_UNITS

I don't know the history, but it might be worth considering if there's a particular milestone at which the fallbacks could be removed.

[...]

PR #13229, opened 2026-02-26
Optimize GS RAM

danielzhong:
# Description
When navigating a scene with a Gaussian Splat tileset, the engine would continuously load new tiles without ever unloading tiles that had left the view. This resulted in ever-increasing memory usage, leading to a performance "brick wall" and browser crashes once approximately 20 million splats were "loaded" (consuming ~6GB of RAM).

I found 3 potential bugs:
## 1. `totalMemoryUsageInBytes = geometry + texture + batch`, but `geometryByteLength` is NaN in GS:

In `GaussianSplat3DTileContent.js`, The `ModelComponents.Attribute` object does NOT have a `byteLength`
property (it has `typedArray`, but not `byteLength`).

``` js
// Original code
return this.gltfPrimitive.attributes.reduce((totalLength, attribute) => {
  return totalLength + attribute.byteLength;  // attribute.byteLength = undefined
}, 0);

// 0 + undefined = NaN → NaN + undefined = NaN → Result: NaN
```

------------------------------------------------------------------------

NaN then is written into the tileset global statistics

``` js
// Cesium3DTilesetStatistics.js (line 89)
this.geometryByteLength += content.geometryByteLength;

// 0 + NaN = NaN → Any number + NaN = NaN
```

------------------------------------------------------------------------

totalMemoryUsageInBytes becomes NaN

``` js
// Cesium3DTileset.js (line 1722)
return (
  statistics.texturesByteLength +   // Normal number, e.g. 300MB
  statistics.geometryByteLength +   // NaN ← here
  statistics.batchTableByteLength   // 0
);

// Any number + NaN = NaN
```

------------------------------------------------------------------------

Cache eviction `while` condition is always false

``` js
// Cesium3DTilesetCache.js - unloadTiles()

while (
  node !== sentinel &&
  (tileset.totalMemoryUsageInBytes > tileset.cacheBytes || trimTiles)
  //          ↑ NaN              >   536870912
  //                               = false (JavaScript rule: any comparison with NaN is false)
) {
  // This loop body will never execute
}
```

Attempted solution:
Change the sum to `attribute.typedArray?.byteLength ?? 0`, This produces a real value (~10 MB per tile × N tiles).
 
However:
totalMemoryUsageInBytes =
    texturesByteLength (normal) +
    geometryByteLength (now has value)

= 300MB + 1000MB
= 1300MB > cacheBytes (512MB)
Cesium detects memory over budget → triggers increaseScreenSpaceError →
SSE changes → during zoom in, already loaded tiles are marked as "no longer needed" →
everything gets evicted → next frame reloads → infinite load/unload loop →
every zoom in/out stutters(because no cache/reload).

So I chose to return 0 because Gaussian Splats do not use GPU VBOs. There may be a better solution, but for now I think this approach may "fixes" the issue.

## 2 When a tile is evicted, it destroys the gaussianSplatPrimitive.
This logic looks incorrect to me because gaussianSplatPrimitive is shared across all tiles in the tileset. However, this function is triggered during Cesium3DTilesetCache.unloadTiles().
60e038d99480344990bc6993e3e25e661ebf666a

## 3 `texturesByteLength` Null Pointer Crash
When the primitive is not initialized yet or has already been destroyed,
`this._tileset.gaussianSplatPrimitive.gaussianSplatTexture` throws an error, so I added a null guard.

------------------------------------------------------------------------

# Result:
Before:
```
total=NaNMB  geo=NaNMB   ← NaN confirmed
→ no [GS evict] at all   ← LRU never fires
tex: 13.3 → 83.4 → 102.5 → 104.9MB  ← grows unbounded
```

After:
```
total=5.7MB  geo=0.0MB   ← valid numbers
→ [GS evict] fires repeatedly  ← LRU working correctly
per-tile release: pos+rot+sca+sh ≈ 7–20 MB of CPU memory freed
```

## Issue number and link
https://github.com/iTwin/platform-bentley-community/issues/309#issuecomment-3959193589

## Testing plan

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,8 +29,8 @@
 
 #### Fixes :wrench:
 
-- Fixed memory leak when rendering Gaussian splat 3D tilesets.
-- Fixed Gaussian splat race conditions in snapshot/sort updates by enforcing explicit snapshot states, preventing stale async results from causing flickering, WebGL draw errors, and unstable LOD transition performance. [#13016][#12965]
+- Fixed memory leak when rendering Gaussian splat 3D tilesets. [#13229](https://github.com/CesiumGS/cesium/pull/13229/)
+- Fixed Gaussian splat race conditions in snapshot/sort updates by enforcing explicit snapshot states, preventing stale async results from causing flickering, WebGL draw errors, and unstable LOD transition performance. [#13016](https://github.com/CesiumGS/cesium/pull/13016) [#12965](https://github.com/CesiumGS/cesium/pull/12965)
 - Fixed flashing when rendering multiple Gaussian splat primitives by storing draw-command model matrices per primitive (`_drawCommandModelMatrix`). [#12967]
 - Fixed depth-testing when `Billboard.disableDepthTestDistance` is `0`. [#13150]
 - Fixed billboard depth testing near horizon. [#13159]


danielzhong:
Since I don’t have the exact environment or an extreme edge-case test dataset, I can’t say with 100% confidence that this completely resolves the RAM issue. However, I ran local tests with a limited memory setup using `tileset.cacheBytes = 50 * 1024 * 1024;`, and it worked as expected. I’m open to any feedback on this PR!

javagl:
Each point that you mentioned seems to make sense for me. But there's something missing about the "big picture" on my side.


**About 1.:** I had seen the `NaN` in the "Geometry Memory" section, but considered this hardly be more than a ~"UI-level glitch". Specifically: It did not seem to be related to the rendering becoming "blocked". 

---

Interlude about the `geometryByteLength` in `Cesium3DTileContent`: I recently stumbled over that, in the context of https://github.com/CesiumGS/cesium/pull/12977, and my verdict right now is that [this does not make sense](https://github.com/CesiumGS/cesium/blob/41e39a48dbebc9b8f57c0c8e9c00ab538ad810dd/packages/engine/Source/Scene/Dynamic3DTileContent.js#L1338). It still has to be implemented, but nobody knows what the "right" way of implementing it is - namely, whether it should be the amount of geometry that is currently _loaded_, the geometry that is _rendered_, and (orthogonal to that) whether it should refer to the data that is on the _CPU_ or on the _GPU_. You may have seen that most impementations simply return `0` there anyhow. One could be tempted to consider this as "not being important". But it **is** important: At some point, this value turned from a purely informative (and possibly wrong...) value that was _shown in the UI_, into a (still possibly wrong...) value that _governs a crucial aspect of the behavior!_. Namely, when it was used for updating the MSSE based on the `cacheBytes`. Now, one can argue that the gaussian splatting content should return that `0` there. But technically, a gaussian splat data set is just a glTF data set, AKA a `Model`, which _does_ return the actual geometry size there (the difference is only ~"how things are rendered", so to speak). It's inconsistent, but there is no way out of that. If returning `0` resolves the issue, that's as good as it gets.

---

**About 2.:** 

> This logic looks incorrect to me because gaussianSplatPrimitive is shared across all tiles in the tileset.
 
That's definitely true ... on one level: Destroying a content should not destroy that `gaussianSplatPrimitive`.

On other levels (not really relevant for this PR, but maybe worth mentioning): 

- There is no `gaussianSplatPrimitive` property in `Cesium3DTileset` (and certainly not one that is _documented_ in any way). This property is just randomly stored there, _by_ the `GaussianSplat3DTileContent`. There is no lifecycle associated with that. It is never destroyed (even when the tileset is destroyed).
- The `gaussianSplatPrimitive` may (probably) not be part of the tileset. When two tilesets are loaded that both contain gaussian splats, there currently is no mechanism for establishing a sort order between the splats. The `gaussianSplatPrimitive` might have to go the the `Scene` level
- I used the _previous_ state (before the change here), and inserted some `consloe.log...` statements where the `gaussianSplatPrimitive` was created or destroyed. Apparently, it was only created _once_. I have never seen it being destroyed. Apparently, content is never unloaded. Too many details to zoom into...

---

Bottom line: I tried out this state, and it _seems_ to resolve the issue for me (even though I don't fully understand _why_...)

ggetz:
Thanks @danielzhong! If we're reasonably confident on the fix for each of the bugs mentioned in the PR description, could we add unit tests to verify?

ggetz:
Also, could you please add an update to `CHANGES.md`? It doesn't need too much detail—It should explain the change from the user's perspective, i.e. "Fixed memory leak when rendering Gaussian splat 3D tilesets".

CHANGES.md
```diff
@@ -29,15 +29,16 @@
...
-- Fixed shader cache lookup for day/night alpha in Columbus View. [#13216](https://github.com/CesiumGS/cesium/pull/13216)
-- Fixed precision of point cloud attributes when accessed in a custom fragment shader. [#13170](https://github.com/CesiumGS/cesium/pull/13170)
-- Fixed a point-rendering regression which caused points to render as circles rather than squares. Such points now will only render as circles when their width is specified via the [BENTLEY_materials_point_style](https://github.com/CesiumGS/glTF/pull/91) glTF extension (which requires that such points be circular). [#13217](https://github.com/CesiumGS/cesium/issues/13217)
-- Fixed a coordinate switching bug in `OpenCageGeocoderService`. [#13138](https://github.com/CesiumGS/cesium/issues/13138)
-- Fixed a regex expression used to find metadata variables in `CustomShader`s, and extended it to work with `metadataClass` and `metadataStatistics`. [#13231](https://github.com/CesiumGS/cesium/pull/13231).
+- Fixed memory leak when rendering Gaussian splat 3D tilesets. [#13229](https://github.com/CesiumGS/cesium/pull/13229/)
```
ggetz, line 32: @danielzhong Version 1.139  went out yesterday (2026-03-02). This change should be reported under the next version, 1.140 instead.

ggetz: Since this has been merged, I'll push a small update to this in `main`.

ggetz: Fixed in https://github.com/CesiumGS/cesium/commit/285623656e03f00c117a3313c0ab4bb7aef5fb03

danielzhong: Great, thanks!

hahahahahahahahahahh: > @danielzhong Version 1.139 went out yesterday (2026-03-02). This change should be reported under the next version, 1.140 instead.@danielzhong Version 1.139 went out yesterday (2026-03-02). This change should be reported under the next version, 1.140 instead.@danielzhong 1.139版本已于昨日（2026年3月2日）发布。此变更应在下一个版本1.140中进行报告。

Since version 40 has specific requirements for WebGL, could we have a minor version to support this feature?

ggetz: @hahahahahahahahahahh We typically don't do patch releases for single features. 

Are you referring to the "Breaking Changes" listed for [1.140](https://github.com/CesiumGS/cesium/blob/main/CHANGES.md#1140---2026-04-01)? What are your concerns there? It's very likely that, unless you have very specific system requirements, the breaking changes are unlikely to affect your users.

hahahahahahahahahahh: Billboards and labels now require device support for WebGL 2, or WebGL 1 with ANGLE_instanced_arrays and MAX_VERTEX_TEXTURE_IMAGE_UNITS > 0. https://github.com/CesiumGS/cesium/issues/13053 https://github.com/CesiumGS/cesium/pull/13253   
As some of my users may be using old computers, I'm worried that their machines won't be able to support this content.

javagl: Does anyone know, from the tip of their head, whether splats have specific WebGL support requirements?

(Specifically: I wouldn't be surprised if they also required WebGL 2, but maybe they don't. And ... another vague gut feeling: Any device that is so old that it does not support WebGL 2 (or the given WebGL 1 extension) will likely have a hard time rendering splats  _in general_, because they _can_ be quite demanding in terms of GPU power...)

danielzhong: I think the current Cesium JS Gaussian splat implementation is effectively WebGL 2 only. It relies on integer textures/samplers, texelFetch/textureSize, gl_VertexID, and UINT-based per-instance splat indices, and Cesium’s WebGL 1 shader fallback does not translate those features. So I would not expect CesiumJS Gaussian splats to run on WebGL 1 without a separate fallback implementation. Also, devices old enough to miss WebGL 2 are generally poor targets for splat rendering anyway.

javagl: _Also, devices old enough to miss WebGL 2 are generally poor targets for <strike>splat</strike> rendering anyway._

Fixed 😁

PR #13234, opened 2026-02-27
Fix lint-staged config for engine files

jjspace:
# Description

Apparently, according to the FAQ on [How to use `lint-staged` in a monorepo](https://github.com/lint-staged/lint-staged?tab=readme-ov-file#how-to-use-lint-staged-in-a-multi-package-monorepo)

> When running, lint-staged will always use the configuration closest to a staged file, so having separate configuration files makes sure tasks do not "leak" into other packages.
> ...
> You should treat all lint-staged configuration files as isolated and separated from each other. You can always use JS files to "extend" configurations

So basically the new config that we added for `tsc` on `engine` specific files was overriding our default top-level config and skipping prettier and eslint.

Easy fix, move config out of `package.json` and extend in the engine specific one.

## Issue number and link

No issue, local dev problem

## Testing plan

- Make a format change in any `engine` source file
- `git add .` and `npx lint-staged`
- Make sure lint staged ran prettier on said file
- Feel free to test other files in `widgets` to _not_ run `tsc`

# Author checklist


ggetz:
Thanks for the fix @jjspace! I can confirm the precommit hooks are working as expected, both for changes in `packages/engine/*` as well as in other directories.

PR #13235, opened 2026-03-01
Fix: Gaussian splat attribute texture exceeding GPU maximum texture size

danielzhong:
# Description
When the number of selected Gaussian splats exceeded ~16 million, the renderer
crashed with:
```
Error generating Gaussian splat texture: DeveloperError: Height must be less than or equal to the maximum texture size (16384).  Check maximumTextureSize.
Error
    at new DeveloperError (http://localhost:8080/packages/engine/Build/Unminified/index.js:7458:11)
    at new Texture (http://localhost:8080/packages/engine/Build/Unminified/index.js:36358:11)
    at createGaussianSplatTexture (http://localhost:8080/packages/engine/Build/Unminified/index.js:135475:10)
    at processGeneratedSplatTextureData (http://localhost:8080/packages/engine/Build/Unminified/index.js:135337:39) (on line -1)
```
Since the WASM worker has no WebGL context, it is unaware of the GPU's
`context.maximumTextureSize` limit (My machine is: 16384). With enough selected tiles,
the computed height silently exceeded this limit, causing a WebGL crash when
the texture was created. ()

A secondary crash existed for **spherical harmonics textures**: the SH texture
used `context.maximumTextureSize` as its fixed width, but its height was also unbounded,
overflowing at ~17.9M splats (for SH degree 3).

## Fix
After the WASM returns the packed data, now the JS layer uses `maximumTextureSize` directly as the texture width:

(optimalWidth = maximumTextureSize), with splatsPerRow = maximumTextureSize / 2 (each splat occupies two side-by-side texels). The required height is then ceil(numSplats / splatsPerRow).

This works because the WASM flat data layout is width-independent: splat i is always at flat position i * 8 regardless of texture width. The same buffer is reused as-is at the wider width without any repacking. Texture memory usage is identical to the previous fixed-2048 layout (numSplats × 2 texels either way).

Because `maximumTextureSize` varies by GPU (4096 / 8192 / 16384), the previously hardcoded shader expressions(2048) are replaced with uniforms:
before:
```
ivec2 posCoord = ivec2((texIdx & 0x3ffu) << 1, texIdx >> 10);
```

Now:
```
ivec2 posCoord = ivec2((texIdx & u_splatRowMask) << 1, texIdx >> u_splatRowShift);
```

Fallback:
**Splat texture:** If the splat count exceeds `maximumTextureSize × (maximumTextureSize / 2`) (roughly 134 million splats on a GPU with a 16384 texture size limit)  the texture height is clamped to `maximumTextureSize` and the attribute arrays (positions, rotations, scales, colors, and spherical harmonics data) are all truncated proportionally to match the reduced count. A `console.warn` is emitted. To avoid hitting the cap on the following frame, `maximumScreenSpaceError` is temporarily multiplied by `originalCount / clampedCount` during the next tile traversal, steering LOD selection toward coarser tiles.

**SH texture:** The SH texture width is fixed at `maximumTextureSize` and cannot be widened. If the required height exceeds `maximumTextureSize`, spherical harmonics are disabled for that snapshot: `sphericalHarmonicsDegree` is set to 0, the existing SH texture is released, and rendering falls back to base color. A `console.warn` is emitted.

## Issue number and link
https://github.com/iTwin/platform-bentley-community/issues/309#issuecomment-3977042573

## Testing plan

I’ve tested using the dataset referenced in the issue comment (“tileset-texture-memory 2026-02-28.zip”). The error is no longer occurring. I also manually adjusted the splat count to trigger the fallback logic, and it behaves as expected.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/ModelComponents.js
+++ b/packages/engine/Source/Scene/ModelComponents.js
@@ -1,386 +1,401 @@
+// @ts-check
+
 import AlphaMode from "./AlphaMode.js";
 import Cartesian3 from "../Core/Cartesian3.js";
 import Cartesian4 from "../Core/Cartesian4.js";
 import Matrix3 from "../Core/Matrix3.js";
 import Matrix4 from "../Core/Matrix4.js";
 
+/** @import ArticulationStageType from "../Core/ArticulationStageType.js"; */
+/** @import AttributeType from "./AttributeType.js"; */
+/** @import Axis from "./Axis.js"; */
+/** @import Cartesian2 from "../Core/Cartesian2.js"; */
+/** @import ComponentDatatype from "../Core/ComponentDatatype.js"; */
+/** @import Credit from "../Core/Credit.js"; */
+/** @import IndexDatatype from "../Core/IndexDatatype.js"; */
+/** @import InstanceAttributeSemantic from "./InstanceAttributeSemantic.js"; */
+/** @import InterpolationType from "../Core/InterpolationType.js"; */
+/** @import Matrix2 from "../Core/Matrix2.js"; */
+/** @import ModelPrimitiveImagery from "./Model/ModelPrimitiveImagery.js"; */
+/** @import PrimitiveType from "../Core/PrimitiveType.js"; */
+/** @import Quaternion from "../Core/Quaternion.js"; */
+/** @import StructuralMetadata from "./StructuralMetadata.js"; */
+/** @import Texture from "../Renderer/Texture.js"; */
+/** @import VertexAttributeSemantic from "./VertexAttributeSemantic.js"; */
+
 /**
  * Components for building models.
  *
  * @namespace ModelComponents
  *
- * @private
+ * @ignore
  */
 const ModelComponents = {};
 
 /**
  * Information about the quantized attribute.
  *
- * @alias ModelComponents.Quantization
- * @constructor
- *
- * @private
+ * @ignore
  */
-function Quantization() {
-  /**
-   * Whether the quantized attribute is oct-encoded.
-   *
-   * @type {boolean}
-   * @private
-   */
-  this.octEncoded = false;
+class Quantization {
+  constructor() {
+    /**
+     * Whether the quantized attribute is oct-encoded.
+     *
+     * @type {boolean}
+     * @ignore
+     */
+    this.octEncoded = false;
 
-  /**
-   * Whether the oct-encoded values are stored as ZXY instead of XYZ. This is true when decoding from Draco.
-   *
-   * @type {boolean}
-   * @private
-   */
-  this.octEncodedZXY = false;
+    /**
+     * Whether the oct-encoded values are stored as ZXY instead of XYZ. This is true when decoding from Draco.
+     *
+     * @type {boolean}
+     * @ignore
+     */
+    this.octEncodedZXY = false;
 
-  /**
-   * The range used to convert buffer values to normalized values [0.0, 1.0]
-   * This is typically computed as (1 << quantizationBits) - 1.
-   * For oct-encoded values this value is a single Number.
-   *
-   * @type {number|Cartesian2|Cartesian3|Cartesian4|Matrix2|Matrix3|Matrix4}
-   * @private
-   */
-  this.normalizationRange = undefined;
+    /**
+     * The range used to convert buffer values to normalized values [0.0, 1.0]
+     * This is typically computed as (1 << quantizationBits) - 1.
+     * For oct-encoded values this value is a single Number.
+     *
+     * @type {number|Cartesian2|Cartesian3|Cartesian4|Matrix2|Matrix3|Matrix4}
+     * @ignore
+     */
+    this.normalizationRange = undefined;
 
-  /**
-   * The bottom-left corner of the quantization volume. Not applicable for oct encoded attributes.
-   * The type should match the attribute type - e.g. if the attribute type
-   * is AttributeType.VEC4 the offset should be a Cartesian4.
-   *
-   * @type {number|Cartesian2|Cartesian3|Cartesian4|Matrix2|Matrix3|Matrix4}
-   * @private
-   */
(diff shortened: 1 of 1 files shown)


danielzhong:
I opened this PR to gather feedback. I’ll add unit tests and update changes.md once the direction is finalized.

lukemckinstry:
Hi @danielzhong, in starting to review, I see the initial error is gone, but I am seeing odd behavior in the rendering of the splats. I am using the dataset referenced in the issue comment `tileset-texture-memory 2026-02-28.zip` and the sample sandcastle included along with it. Any idea what is happening? Do you see similar behavior? When I started testing yesterday I did not see the tileset obscured in this same manner as I do today in the screencapture.

https://github.com/user-attachments/assets/d3d9a031-bb8e-4630-b90f-0ca2db17e323

danielzhong:
> Hi @danielzhong, in starting to review, I see the initial error is gone, but I am seeing odd behavior in the rendering of the splats. I am using the dataset referenced in the issue comment `tileset-texture-memory 2026-02-28.zip` and the sample sandcastle included along with it. Any idea what is happening? Do you see similar behavior? When I started testing yesterday I did not see the tileset obscured in this same manner as I do today in the screencapture.
> 
>  Screencast.from.2026-03-13.14-43-41.mp4

I'm seeing the same behavior as well. I assumed it was related to the test sample @javagl  created for investigating the GPU max texture size limitation. Since the official customer datasets render correctly and only this specific test dataset shows the issue, I didn't consider it a visual concern. I was mainly checking that the error was resolved. I wondered if it might be related to camera-based culling in this particular scene or view, rather than anything related to the Gaussian Splat rendering itself.

javagl:
Assuming that this refers to this ~"circular, screen-spaced disappearance" of splats in a certain area at a certain zoom level: 
I had seen this. 
I had not (yet) pointed it out. 
One reason for that: I had not yet identified the reason - not even narrowed it down reasonably.
Now I think that I _somehow_ narrowed it down. 
I'll open another issue soon.

javagl:
I opened https://github.com/CesiumGS/cesium/issues/13304 , but don't expect too much from the description: It's only narrowed down a tiny bit.

Most relevant for this PR @lukemckinstry : This is unrelated to the changes _here_. 

It only became more apparent with the test data here, because it somehow (strangely) depends on the density of the splats.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -422,65 +429,147 @@ async function processGeneratedSplatTextureData(
...
-      for (let i = 0; dataIndex < snapshot.shData.length; i += width * 2) {
-        texBuf.set(
-          snapshot.shData.subarray(dataIndex, dataIndex + floatsPerRow),
-          i,
+      const _shHeight = Math.ceil(snapshot.numSplats / splatsPerRow);
+
+      // SH texture width is already maxTex and cannot be widened further.
+      // When height would exceed the GPU limit, gracefully disable SH for this
+      // snapshot and fall back to base color rendering rather than crashing.
+      if (_shHeight > width) {
```
lukemckinstry: I think we should remove the underscore here and change to `shHeight` since the CesiumJS [coding code](https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide) has specific uses for variables prefixed with underscores which this does not align with, unless there is something I am missing that the underscore is trying to communicate.

javagl:
> The primary error & crash for this PR are no longer occurring on this branch, so that's good.
>
> However working with the test dataset, it seems it is not triggering the new logic in the PR. No matter how I position the camera to load as many splats as possible, I cannot trigger the logic to show the console warning.

Daniel to confirm, but based on the description from the first post...:

> Fallback:
> **Splat texture:** If the splat count exceeds `maximumTextureSize × (maximumTextureSize / 2`) (roughly 134 million splats on a GPU with a 16384 texture size limit)  the texture height is clamped to `maximumTextureSize` and the attribute arrays (positions, rotations, scales, colors, and spherical harmonics data) are all truncated proportionally to match the reduced count. A `console.warn` is emitted. ...

and

> I also manually adjusted the splat count to trigger the fallback logic,

it sounds like these warnings are only emitted with data sets that are larger than the theoretical maximum (134 million). The test data until now was much, much smaller, around 20 million. 

(Yes, eventually, people (with faster machines and huge RAM) will try 100 million splats - or [103 million, to be precise](https://community.cesium.com/t/how-to-publish-103m-splats-on-cesium-how-to-process-20-000-photos-for-a-3dgs-model-on-a-single-rtx-4090-after-3-months-of-r-d-i-found-the-answer-but-cant-publish-it-on-cesium/42910) - and they will hit this limit, but as long as the maximum texture size is fully used, and it does not crash _beyond_ that, everything should be fine)

danielzhong:
> The primary error & crash for this PR are no longer occurring on this branch, so that's good.
> 
> However working with the test dataset, it seems it is not triggering the new logic in the PR. No matter how I position the camera to load as many splats as possible, I cannot trigger the logic to show the console warning.
> 
> So this may meet our objective but it is unclear if it is working as intended.
> 
> Are others able to produce the console warnings - or debugger statements on those code paths - using the test dataset?

Thanks @lukemckinstryfor checking this, and thanks @javagl for clarifying this point.

The warning paths in this PR are only expected to trigger once we exceed the GPU's theoretical texture capacity, not with the current test dataset. On a machine with `maximumTextureSize = 16384`, the base splat texture cap is `16384 * (16384 / 2) = 134,217,728` splats, so a dataset around ~20M splats would not hit that fallback path or emit the console warning.

That is also why the original error is no longer visible: the texture sizing now uses `maximumTextureSize` up front, so we avoid creating an invalid texture in the first place. The `console.warn` paths are only for the overflow fallback cases beyond that limit.

This can be triggered locally by temporarily lowering the effective `maxTex` in `processGeneratedSplatTextureData`, which forces the current test dataset through the fallback path and allows the warning/fallback behavior to be verified.

image

PR #13238, opened 2026-03-02
Metadata picking regression fix

mzschwartz5:
# Description

#13135 and subsequent PRs to support property table usage in custom shaders did not account for backwards compatibility with the `pickMetadata` API, which reads back a pick framebuffer containing metadata values. 

There are three bugs addressed by this PR:
1. The `DerivedCommand` class, which sets up shader code for reading back picked metadata, rolls its own `glslType` function, which is no longer correct after #13135 enhanced type support for property textures. Instead, it should use the `glslType` function from the `MetadataClassProperty` class.
2. The `MetadataPipelineStage` now excludes metadata properties not used in other pipeline stages, but did not account for `pickMetadata`. This isn't easy to account for, since it's dynamic (can change what metadata properties are needed every frame, or multiple times a frame, depending on how the user calls `pickMetadata`). In this PR, I revert this optimization for property textures and property attributes. More details in the file changes and comments.
3. There was a bug in the code that unpacks property textures (and textures for property tables), for certain value types (namely, for fixed-sized arrays or vecN's that store one value per component).

## Issue number and link
No issue- this is emergent, found as a result of failing tests in the `SceneSpec` (`pickMetadata` describe block).

## Testing plan

- Run the `Scene/Scene` specs (with attention to the `pickMetadata` describe block).
- Download the zip file below for a sandcastle and simple tileset. The tileset is just a quad with different metadata values according to a property texture. On main right now, it has compilation errors. On this branch, you should see metadata values logged as you mouse over.
[...]

PR #13242, opened 2026-03-02
Fix channel string slice in texture unpacking

mzschwartz5:
# Description

Fixes a small edge case with unpacking property textures in shaders. Basically, if you define a property texture in a gltf that uses channels [1, 2] (that is, multiple channels not starting at index 0), then the current code will unpack the texture into a vec2 and sample it using the `.gb` channels. A vec2 doesn't have a `b` channel, only `rg`.

The new code slices the string literal `rgba` instead of slicing the channel string used by the property texture (which, in the above example, was `gb`).

## Testing plan

Same as #13238

# Author checklist


jjhembd (approved):
Thanks @mzschwartz5, I think this is good.
The constructed shader stuff often looks fragile and hard to read, but this change does look like a correction and it fixes the failing spec on my machine.

PR #13247, opened 2026-03-03
Adds nga-gpm extension texture IDs to gltf primitives

mzschwartz5:
# Description

With recent work to support property tables in custom shaders, there was a regression with the NGA GPM local extension:

1.138:
image

1.139:
image

The recent changes added a step to collect all metadata properties on a given primitive in the custom shader pipeline stage, and to compare that set against the set of properties used in the custom shader. If the shader uses properties the primitive does not have, the shader is disabled for that primitive.

For the nga-gpm local extension, property texture IDs are not being added to the primitive's collection. Thus, all custom shaders on nga primitives get disabled. This PR adds those texture IDs to primitives on load.

## Testing plan

Test with the [nga-gpm sandcastle](https://sandcastle.cesium.com/?id=3d-tiles-nga-gpm-visualization).

# Author checklist


mzschwartz5:
@javagl - can you advise on unit testing for this? Should we add something to the `GltfLoader` spec here and, if so, how would I generate appropriate testing data for the nga gpm extension?

mzschwartz5:
That works - just needs a small tweak to destroy remove primitives before clearing the resource cache.

javagl (approved):
No strong opinion about the place for the test (but I think it makes sense to have dedicated tests for extensions, especially such highly specific things like the GPM extension), so all fine from my side.

Issue #12258, opened 2024-10-18
Improve Clipping Polygon Performance


ggetz:
### Feature

https://github.com/CesiumGS/cesium/pull/11750 added an initial implementation of clipping polygons. However, some users are seeing performance hits depending on usage.

The performance hit is not directly correlated with the number of polygon positions. The geographic extent, concavity, and number of polygons also factor in. For example, [the AEC Clipping Example](https://sandcastle.cesium.com/?src=AEC%20Clipping.html) is a particularly concave polygon with many positions shows this performance hit.

javagl:
> The performance hit is not directly correlated with the number of polygon positions. 

Certain aspects of the polygon shape may have an _additional_ impact here. (For example, a polygon with 100 points that covers a pole and the IDL may have a larger impact than one that does not). But the performance hit is definitely correlated with the number of polygon points.

I created a sandcastle where you can crank up the number of polygon points from 4 to 500 (and it's a circle - i.e. a as convex as it can be). And one can clearly see the performance hit:

![Cesium Clipping Polygon Performance](https://github.com/user-attachments/assets/af5a55ae-7547-49e2-9e86-36c608500a1c)

The sandcastle:

[...]

PR #12650, opened 2025-06-03
Improve clipping polygon performance

javagl:
# Description

Cache the "rectangle" that is computed in the `ClippingPolygon` class, so that it does not have to be re-computed each frame.

As mentioned in [this issue](https://github.com/CesiumGS/cesium/issues/12258), the clipping polygons had a distressingly low performance for a higher number of points. As analyzed in that issue, this was largely caused by the `computeRectangle` function in `ClippingPolygon`, which did re-compute the rectangle each frame.

This PR caches the rectangle, and only re-computes it when the `_positions` of the polygon change. For that, it was necessary to actually also cache the _positions_, and compare the cached positions with the _actual_ positions (there is no other way to detect changes in the positions). 

## Issue number and link

Addresses https://github.com/CesiumGS/cesium/issues/12258

## Testing plan

[...]

PR #13255, opened 2026-03-04
Fix performance issue with multiple ClippingPolygon on Cesium3DTileset

alarkbentley:
# Description

Improves the performance of [`ClippingPolygonCollection.computeIntersectionWithBoundingVolume`](https://github.com/CesiumGS/cesium/blob/ee0bf1d5a9f994f5ae723915f16b7e6f32b1a0e2/packages/engine/Source/Scene/ClippingPolygonCollection.js#L743) by reducing redundant computation and enabling early exit. The referenced example with 200 clipping polygons applied to GooglePhotorealistic3DTileset improves from 1 fps to 30 fps.

### Hoist tile bounding rectangle computation out of the loop

Previously, the tile's bounding rectangle was recomputed on every iteration of the polygon loop, even though it is invariant across iterations. This moves the computation before the loop so it executes exactly once regardless of how many clipping polygons exist.

### Early return on first intersection

Previously, when a polygon was found to intersect the bounding volume, the result was stored in a variable and the loop continued checking remaining polygons unnecessarily. Now the method returns `Intersect.INTERSECTING` immediately on the first match, skipping all subsequent polygon checks.

### Pass scratch rectangle to `computeRectangle`

`polygon.computeRectangle()` was called without a scratch result parameter, causing a new `Rectangle` allocation on each call. A module-level `scratchRectanglePolygon` is now passed to avoid per-call allocations.

## Impact

These are pure performance optimizations with no behavioral change. The benefit scales with the number of clipping polygons in the collection — scenes with many clipping polygons will see reduced per-frame overhead during tile visibility determination.

## Testing

Added unit tests covering:
- `OUTSIDE` result when no polygons intersect the bounding volume
- `INSIDE` result when `inverse` is true and no polygons intersect
- Early return verified by spying on the second polygon's `computeRectangle` (never called when the first polygon already intersects)
- `INTERSECTING` result when only a later polygon in the list intersects
- Tile bounding rectangle computed exactly once for `TileBoundingSphere` (via `Rectangle.fromBoundingSphere` spy)
- Tile bounding rectangle computed exactly once for `TileOrientedBoundingBox` (via `Rectangle.fromCartesianArray` spy)

## Future Improvements

Performance could be further approved if we could ensure that the ClippingPolygons are immutable.

## Demo

Based on exising sandcastle "Clipping Regions". Creates 200 clipping polygons (performance starts degrading already at ~10 polygons) and attached onto GooglePhotorealistic3DTileset.

[Sandcastle](https://sandcastle.cesium.com/#c=tVdtb9s2EP4rRLAPcuvKeqeUuMGytMsKpEjQZCuwJkBoibaJyqJB0XHdIv99R4qUX2K5L8ACBJbIe8jnHt4dT2w250KiF4jU6JzWbDFDY8Fn6O4o1293Ryd34q5ijdkNqYqc1LKk1mo9YizVf86rWqJHRpdUoNeookuztvuPHnPs6ue8koRVVNwd9dE3BUVIshktYewYjUlZ034zSoACkYxX28N1Tiv6nhf0muWfqdieHJGaXpIVFfsmJ5TngINRQ+0dry6asWvBHxlM3a7m1L24urq4fKtBT72TtXd6Z3Cu8dLVr60AgwF6W5ERyCRoBSuxaoLklKL680pNa2MXXs7kjNfzKRXwNuVLWE6KxXqZzbXdgo4Wkxuw+lOQGa2vqbihQKXYBZVUoiUXZXFLhQBx9YQUK6vv5hxgyZIwaTXIBSWSftywOKtXVe40jhu5XdlMWZlgkWf7WdNJyUetb1p+Pf2EciLzKXIAwkWvpcaqgi9dUlIhnYdbpQtaQmCSCmlDpPkpMfWGyBBx0W/f9PzTQ0P06bkWrKQ1lfu1aOb2a3HB+aSk11MuObyXrJYsD98YiGOWQQgO/KoqV/qQGwgywQSkgf6IokVNC/BQTq3B9qIofIP0si5CN8BGrfRggvQBzQWfgygrxMd6xuSWDkWxyGEdxOcqPWokOToYyq7lzIHw3zWo+RFINZwu2qRQIWUy5Wn79OeCzZhkj8CUFIWzKWHv0OEqrrykbsknzsNbfZwlJ4U6zQ4loBToZVvG9q/ruOEUzorCwpRWSjeuhFc1aLRgpdrPRoZ97wiOnend+Gh+2lhwVUEE4c9qeHlXOH6CvdRPD2q3s8OvymeX2dHrgEr/cijeECgqlDaVgSVzXWXXdS6HYiPI1XjcSLBRy/+ievtrBmw/kGpCnWZfM/2eyKkr+QcwIlXtZLHr9fqHLF756dok8TzX088Ne1MKvwLxW74rXH+LZe9kJx4Iyks2nyv/BJ2AdyDmoynKXJISvJZrh+e8XE1UIr1Gn+43Cr4qloHn6ZExqO+oEGIw6J3AzxBV8PPy5dZ5SVR8AQP2AnzxPN+Egt3AnS/qqbMh6Llhed0YrKsLhDDTyd1eVudESHgiVajj7g2dCErrMyiHK2edLZ82E+dT8eWV78EpeBj7XpDhIMFhH4WZi2M/SxPs4yzFSXzf34/K4jROvSTwQt+gvCiLAIQ9HIWdqCRTrvtRFNnNvDD20iBJsR+knTAcpGGa4jjMWpgfxmozIHEA5uMo8NIkDnEDizIYC4F15Ht+N8kgSzDGMQ5jC8NJmuEsi6IgTDpxYQCKgDNhmFhckKYhhAnO4k4Y+JWGGIdZFLQslT5B5KU486JOXBxlYeRhJYzFJQmoAhTjQ7ggiHGaegHGLU3w149iHMOWnTjsx0nkp6CDlSX2Qz+AY4v8LAuSn8LFUQAUQazOAPvxsLx3xyWRTtBrx3rtXbVT7DbbkHw7v+rtcraTfee8LGmuss7moc3b4/ZpsyvcvAf/351MZbvlE9VFjKhcUloh7aLtiKBjUmVPjUF5g/vU3AwKuu7Y1R10y3k5IuI9rRaOKRdt2ZH0izyGHt/ex9Ckr3uHmirSx8jpodena8x3Gr/GZEur3cZXGTzZ4+x3UDK95i8z2tptP6Ft0tuM7veehL1EWaXuykdWsxErmVztU10h/lhICYcOH1Fqx7Y7gU8h3X2h8aLSYYGcfErhC6bodbQmlrIxO+kMlvYebC85qj9Tiu9SNJ8zdoEfIXkwI1yz8Q7r7cP6QdDBvCjomCxKqTOCVXDt1xtefM/rd7sAcLv5iPxlvy2Hn/J7P8j4fdQ/GtZyVdJTtc7v5lN9IUrHdQeSzuZQK2k9GC0ACITqGmDVcGAhw4I9Ila83vNNDn6TuoaZ8aIsb9hX+MY/HQ7Afgtm2tAroFiSlTKZ+qeXzaDrusMBvD5HyabwgLliPZQmvJr9GHRY16SipZlWBiNerE5twg6lOF3fIUNZnF7SsVTnlH9WjS3RbR8wgoIBDGC+RQ42oM+X+cAm0511VJNs8sX2zG0e5W3t7twFHlvq8Kz8VEIYQf4D)

### Before fix: 1 fps

before_200

### After fix: 30 fps

after_200

### Immutable Polygons: 60 fps
If we could ensure that the ClippingPolygons are immutable we could avoid the [equalsArrayCartesian3](https://github.com/CesiumGS/cesium/blob/ee0bf1d5a9f994f5ae723915f16b7e6f32b1a0e2/packages/engine/Source/Scene/ClippingPolygon.js#L224) check in ClippingPolygon.computeRectangle which is current bottleneck and achieve steady 60 fps.

immutable_200

## Issue number and link

## Testing plan

# Author checklist

### DIFF
--- a/Apps/Sandcastle/gallery/Mars.html
+++ b/Apps/Sandcastle/gallery/Mars.html
@@ -360,7 +360,7 @@
             entity.name = entity.properties.text.getValue();
             entity.description = createPickedFeatureDescription(entity);
 
-            const flyToDestination = new Cesium.Cartesian3.fromArray(
+            const flyToDestination = Cesium.Cartesian3.fromArray(
               entity.properties.destination.getValue(),
             );
             const orientationArray = entity.properties.orientation.getValue();
--- a/Apps/Sandcastle/gallery/SplitDirection.html
+++ b/Apps/Sandcastle/gallery/SplitDirection.html
@@ -63,17 +63,17 @@
           });
 
           points.add({
-            position: new Cesium.Cartesian3.fromDegrees(-75, 35),
+            position: Cesium.Cartesian3.fromDegrees(-75, 35),
             color: Cesium.Color.RED,
             splitDirection: Cesium.SplitDirection.RIGHT,
           });
           points.add({
-            position: new Cesium.Cartesian3.fromDegrees(-125, 35),
+            position: Cesium.Cartesian3.fromDegrees(-125, 35),
             color: Cesium.Color.WHITE,
             splitDirection: Cesium.SplitDirection.NONE,
           });
           points.add({
-            position: new Cesium.Cartesian3.fromDegrees(-100, 20),
+            position: Cesium.Cartesian3.fromDegrees(-100, 20),
             color: Cesium.Color.YELLOW,
             splitDirection: Cesium.SplitDirection.LEFT,
           });
@@ -93,17 +93,17 @@
           });
 
           billboards.add({
-            position: new Cesium.Cartesian3.fromDegrees(-75, 35),
+            position: Cesium.Cartesian3.fromDegrees(-75, 35),
             image: "../images/facility.gif",
             splitDirection: Cesium.SplitDirection.RIGHT,
           });
           billboards.add({
-            position: new Cesium.Cartesian3.fromDegrees(-125, 35),
+            position: Cesium.Cartesian3.fromDegrees(-125, 35),
             image: "../images/facility.gif",
             splitDirection: Cesium.SplitDirection.NONE,
           });
           billboards.add({
-            position: new Cesium.Cartesian3.fromDegrees(-100, 20),
+            position: Cesium.Cartesian3.fromDegrees(-100, 20),
             image: "../images/Cesium_Logo_overlay.png",
             splitDirection: Cesium.SplitDirection.LEFT,
           });
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,12 @@
 - No longer disables custom shaders for primitives with missing metadata, as long as the metadata exists on the overall class definition. [#13258](https://github.com/CesiumGS/cesium/pull/13258)
 - Fix performance issue with multiple ClippingPolygon on Cesium3DTileset. [#13255](https://github.com/CesiumGS/cesium/pull/13255)
 
+### @cesium/sandcastle
+
+#### Fixes :wrench:
+
+- Fixed split screen labels and Cartesian3 factory function calls, and edited descriptions for various gallery examples. [#13250](https://github.com/CesiumGS/cesium/pull/13250)
+
 ## 1.139 - 2026-03-02
 
 ### @cesium/engine
--- a/packages/sandcastle/gallery/3d-tiles-compare/index.html
+++ b/packages/sandcastle/gallery/3d-tiles-compare/index.html
@@ -14,9 +14,30 @@
   #slider:hover {
     cursor: ew-resize;
   }
+
+  .split-label {
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+    background: rgba(42, 42, 42, 0.8);
+    padding: 8px;
+    border-radius: 4px;
+    z-index: 9999;
+    font-size: large;
+    font-weight: bold;
+    text-align: center;
+  }
+  .split-label.left {
+    left: 0;
+  }
+  .split-label.right {
+    right: 0;
+  }
 </style>
 <div id="cesiumContainer" class="fullSize">
   <div id="slider"></div>
+  <div class="split-label left">3D Tiles<br />Mesh</div>
+  <div class="split-label right">3D Tiles<br />OSM buildings</div>
 </div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
 <div id="toolbar"></div>
--- a/packages/sandcastle/gallery/bing-maps-labels-only/index.html
+++ b/packages/sandcastle/gallery/bing-maps-labels-only/index.html
@@ -14,10 +14,36 @@
   #slider:hover {
     cursor: ew-resize;
   }
+
+  .split-label {
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+    background: rgba(42, 42, 42, 0.8);
+    padding: 8px;
+    border-radius: 4px;
+    z-index: 9999;
+    font-size: large;
+    font-weight: bold;
+    text-align: center;
+  }
+  .split-label.left {
+    left: 0;
+  }
+  .split-label.right {
+    right: 0;
+  }
 </style>
 
 <div id="cesiumContainer" class="fullSize">
   <div id="slider"></div>
+  <div class="split-label left">
+    Bing Maps (labeled)<br />+<br />Washington Imagery
+  </div>
+  <div class="split-label right">
+    Bing Maps (unlabeled)<br />+<br />Washington Imagery<br />+<br />Bing Maps
+    (labels only)
+  </div>
 </div>
 <div id="loadingOverlay"><h1>Loading...</h1></div>
 <div id="toolbar"></div>
--- a/packages/sandcastle/gallery/cardboard/sandcastle.yaml
+++ b/packages/sandcastle/gallery/cardboard/sandcastle.yaml
@@ -1,6 +1,7 @@
 legacyId: Cardboard.html
 title: Cardboard
-description: Configure viewer to add a button enabling look at a mobile device with cardboard.
+description: Set up a split screen rendering suitable for viewing on a mobile device with a cardboard VR headset.
 labels:
   - Camera
+  - VR
 thumbnail: thumbnail.jpg
--- a/packages/sandcastle/gallery/czml-box/sandcastle.yaml
+++ b/packages/sandcastle/gallery/czml-box/sandcastle.yaml
@@ -1,6 +1,6 @@
 legacyId: CZML Box.html
 title: CZML Box
-description: CZML Circles and Ellipses
+description: Load and render CZML Box shapes with different material and outline properties.
 labels:
   - CZML
 thumbnail: thumbnail.jpg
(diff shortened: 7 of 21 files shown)


alarkbentley:
@javagl Any input on supporting Immutable objects in Cesium? Would be nice if we could find a way to do this in a non breaking way and gain another >x2 performance. Was thinking we could introduce an optional Object.freeze on the positions array but we need to be careful as ClippingPolygon already exposes the mutable position array via a getter.

Such a shame if checking the position array has changed N^2 times per frame should be the bottleneck here.

javagl:
To weave the web again: This PR is addressing https://github.com/CesiumGS/cesium/issues/12258 , and you're probably pinging me because you had seen https://github.com/CesiumGS/cesium/pull/12650 

That PR was only a quick shot, from looking at the largest bar in [the performance trace in this comment](https://github.com/CesiumGS/cesium/issues/12258#issuecomment-2423908211) and avoiding the most expensive (unnecessary) recomputations: The `computeRectangleFromPositions` was by far the largest chunk. The PR focussed on the case of a _single_ (possibly complex) polygon. And it required "caching" some positions (which can be difficult - details below).

I remember that there have been...

- different _levels_ in the call hierarchy and
- different _paths_ in the call hierarchy 

where things have been computed redundantly, multiple times, but I didn't further investigate the (new) bottlenecks _after_ that PR was merged. 

_This_ PR here seems to focus on the case of having _many_ polygons. And it seems to avoid a recomputation that was _completely_ unnecessary (meaning that it doesn't even seem to require "caching" anything). So this _looks_ like another low-hanging fruit and a good candidate to be merged. (Pending more detailed reviews and tests).

---

> Any input on supporting Immutable objects in Cesium? 

I'm not sure whether you are asking this because you _overlooked_ that comment, or exactly because you have _seen_ that comment 🙂 but I pointed out some of the [difficulties of caching and mutability](https://github.com/CesiumGS/cesium/issues/12258#issuecomment-2758065912) in the issue.

(The question about (im)mutability could be phrased more broadly, and may even warrant its own issue - focussing on the clipping polygons here for now)

A _very_ important use case of the clipping polygons is that someone has one or many (possibly complex) shapes, maybe even loaded from some CSV or JSON, and just wants to clip stuff away. 
Another reasonable use-case is that someone wants to _create_ and _edit_ such a polygon interactively in CesiumJS. So that should also be _possible_. 

But people will hardly edit 5000 polygons with 500 points "interactively". The expectations for how "dynamic" things are (and _what_ is "updated" where and how) are vastly different, and have implications on the performance requirements (and on any approach that may involve forms of "caching")

It's indeed unfortunate when something like `equalsArrayCartesian3` becomes the bottleneck. I think that it could be avoided by establishing some immutability. (Along the lines of the usual "minimize mutability" mantra...). It would be a breaking change in terms of the _behavior_. Maybe people already created an "interactive clipping polygon editor", and that will simply no longer work afterwards. But I assume that this could be handled (with the usual combination of implementation notes, deprecation periods, and sensible error messages). The decision about that is not my call.

A word of caution, however: 

Trying to identify "redundant computations" and trying to _avoid_ them can be dangerous: Given the _huge_, undocumented state space that we're dealing with, and the fact that every client can do `these.call.chains.to.modify.something = 42` makes it hard to _reason_ about what is actually happening. 

I looked at the code change here, and it _looked_ safe. But still, I quickly checked if this sandcastle ...

https://sandcastle.cesium.com/index.html#c=lVVtb9s2EP4rhD8pmCPzVaRcN1iaDFiBFi3SbB8W5wMj0TYxWjQkKlk65L/v9ObaigOssGCRRz53zz3iHTNfVAE9WvNkSvQeFeYJXZnK1tv4z9YWLSdZO7/yRdC2MOVycvZuWSyLrEU+1NbltljfWmcqE8CFftI2DE66F7vul+NV6bcffXFZweRjHpFEYkVU57AjEVeZKUy8K+3WBvtoqljneTQK0wFmM/SX91sUPAob03If9iHnMx2s3/PM9NaU+stq1ZE8yPN3oxvEVxuyzY0u1iZaFmhY/KzDJg7+BrboooqYiPHZ9O31c0pj0W1IMI4xjBqqfWbfgeytH+cyPeJ2qK3zxbVZXxsXNFAGd5iId8NibtalMdVlWepnWL1rYt6dEwwMsZQE01TSRDL0y6GbKWJpLAVJVSKJTJVMxP30EJkKJRROKGbkJBLzlANQYsnZCJmkGAgSzvnpoJgJrGiiJKFqBJVUMaWkYOkbUMJEExSovYISySlWiWDyFJSnsINBPpxgMiZM00RKKSQTp6EyUalMU84pS0ZYRkElSIex5DSWKsUofIlUjKCQp2JSspTTNxg3ClKOlUwxH2EFTxnHshHrNDZJQCmgK15jKRVSKUylfIMyqEG4kAJCj7CSiIQTBWqclkoQRih8Vk7SlCY/jRWcAl2Qc3QYf/YY38crp0NED2uoNPlX757XvoAi6evQFAFaS99Z/m0iFlCBc7Sc3Jgc7frt8FR1udKZWU5aWv3CHLUQhDYWqrbMNs/zfbfTZYCRLljb6K4PSjQ6rNeuQyC01cGUVrsfDrzzZfzh0x+/xU82bC7dbqMjPLQUCGnsehPmiAiBe5Ovg4O2PEehrE1re4H/l06Dtkde5jnSKHN2t2s6Y2nW0BdR5h8hNsyDD9qBumGQrM+zGrrKQa+86p30kka9Ejtf2abZVv9LiQ6DjjpYZ+vSfGle9y3/UauMs+P41XEnH7EDMZ3JGl4dzyGv+X50KFTz+6aLPNNVcKY5GrfeuwddfqhDABfLyWdQDIWOCRwJtKqL1juKzrojMWa79blxcEOU9h8g+uPGgDlvZen0uOuSJ1OEh6ezwGAw7g17W2ehMDqnvb0x3Z/1OU2mk0UVnp25aMy/2u3OlwHVpYvieBbMdgelYqrZQ5393QhbVQ1oMRsgi9w+Ipu/P3H9w1HSVQUrq9q5b/Y71MfFYgb7j2DOt9fqFzhlDr4vbNmQi0+dMY7jxQymr1Gh0/zA438

still works on _this_ branch when hitting the "Move tileset" button. Yes, it works 🙂 But there may be further, subtle cases where "changing something" and then _not_ "re-computing" something may have surprising effects and cause a wrong clipping result.

alarkbentley:
@javagl Thanks for valuable input! I had seen some of these discussions but not all. I have some ideas how we could introduce immutability without breaking the API, I might ping you in a separate draft.

For now lets keep these changes separate as they should be non breaking, I published the PR and added some unit-tests. Please review when you have time.

alarkbentley:
Nice sandcastle demo, will help if we are looking into more optimizations. Are we ok with merging this PR for now or do you need something more from me?

javagl:
I'll tag @ggetz for general awareness.

Summary:

This PR implements an optimization for clipping polygons. The optimization consists of avoiding redundant computations. I can not imagine any negative side-effects. It is complete (with specs) and tested with a sandcastle that shows huge performance improvements, particularly for the case of _many_ clipping polygons.

From my perspective, it should be merged.

javagl:
>  Selecting 200/100 there will cause it to bail out. That could/should be tracked as an issue, right?

Opened https://github.com/CesiumGS/cesium/issues/13306

javagl:
And I'll just merge this now. What's the worst thing that could happen? A WebGL crash? 😁

PR #12592, opened 2025-04-30
Use physical values for voxel bounds

jjhembd:
# Description

This PR changes the way we handle the bounds of a voxel dataset. Previously, `VoxelPrimitive.prototype.minBounds` and `VoxelPrimitive.prototype.maxBounds` were defined in a "shape space", spanning nominal ranges such as [0..1], [-1.. 1], [-pi/2.. pi/2], or [-pi.. pi]. This was hard to document consistently, and inconsistent with the way bounds are defined in 3D Tiles.

After this PR, the `minBounds` and `maxBounds` are physical values, consistent with what is defined in the tileset.json.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/12570.

## Testing plan

Load the [Voxels in 3D Tiles Sandcastle](http://localhost:8080/Apps/Sandcastle/index.html?src=Voxels%20in%203D%20Tiles.html), and play with the clipping bounds. The extreme values of the sliders should represent the physical dimensions, and clipping should work normally.

# Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [x] I have updated `CHANGES.md` with a short summary of my change
- [ ] I have added or updated unit tests to ensure consistent code coverage
- [x] I have updated the inline documentation, and included code examples where relevant
- [x] I have performed a self-review of my code

PR #13257, opened 2026-03-04
Fix voxel step assignment

JulesBlm:
# Description

In `VoxelFS.glsl`, there is a guard intended to prevent the ray from stalling when the step size is zero which can happen for infinitely thin shapes:

```glsl
 if (step.w == 0.0) {
     // Shape is infinitely thin. The ray may have hit the edge of a
     // foreground voxel. Step ahead slightly to check for more voxels
     step.w == 0.00001;
 }
```
`step.w == 0.00001` is a comparison, not an assignment. So now it evaluates to a boolean which is discarded and the step size is never actually updated. So the guard does nothing and it could probably cause the ray to stall for infinitely thin shapes.

It's updated to `step.w = 0.00001;`

## Testing plan

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -52,7 +52,6 @@
 #### Fixes :wrench:
 
 - Fixed split screen labels and Cartesian3 factory function calls, and edited descriptions for various gallery examples. [#13250](https://github.com/CesiumGS/cesium/pull/13250)
->>>>>>> upstream/main
 
 ## 1.139 - 2026-03-02
 


packages/engine/Source/Shaders/Voxels/VoxelFS.glsl
```diff
@@ -213,7 +213,7 @@ void main()
         if (step.w == 0.0) {
             // Shape is infinitely thin. The ray may have hit the edge of a
             // foreground voxel. Step ahead slightly to check for more voxels
-            step.w == 0.00001;
+            step.w = 0.00001;
```
jjhembd:
```suggestion
            step.w = 0.001;
```

Since https://github.com/CesiumGS/cesium/pull/12592 the voxel step has been a physical value in meters. Maybe 1mm would be a more reasonable minimum?

JulesBlm: Sounds reasonable, I agree.

jjhembd:
Can we also move the changelog entry to the `1.140` release?
Sorry for the noise--I would have fixed it myself but I'm not able to commit to your branch.

JulesBlm:
> Can we also move the changelog entry to the `1.140` release? Sorry for the noise--I would have fixed it myself but I'm not able to commit to your branch.

Sure, I moved it

PR #13259, opened 2026-03-04
ESLint: Enable 'new-cap' rule

donmccurdy:
# Description

In the migration to ES6 classes, a recurring issue requiring manual checks has been that ES6 class factory methods cannot be called with 'new'. For example, this is invalid with ES6 classes:

```javascript
const a = new Cartesian3.fromDegrees(...);
``` 

The factory method must now be called without 'new'. Prior to adopting ES6 classes, either syntax would work, though I think using 'new' was arguably a typo even then.

To avoid having to manually search for these calls each time we convert a new class to ES6 class syntax, I'm proposing that we adopt an ESLint rule that detects the syntax, and preemptively fix its errors. Then future conversions to ES6 classes won't have to worry about this step. A couple of exceptions (non-constructor methods whose names begin with capital letters) are handled by disabling the eslint rule locally, particularly related to the Draco API.

## Issue number and link

- #8359

## Testing plan

No functional differences expected. Testing should include passing unit tests, passing ESLint tests (with newly-added restrictions), and no change in Sandcastle demos.

# Author checklist

#### PR Dependency Tree

* **PR #13259** 👈

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -66,9 +66,9 @@ jobs:
           unzip Cesium-$(cat package.json | jq -r '.version' | sed 's/\.0$//').zip -d Build/release/ -x "Apps"
           aws s3 sync Build/release/ s3://cesium-website/cesiumjs/releases/$(cat package.json | jq -r '.version' | sed 's/\.0$//')/ \
             --cache-control "public, max-age=1800" \
-            --exclude "Build/release/Apps/Sandcastle2/*" \
+            --exclude "Apps/Sandcastle2/*" \
             --delete
-          aws s3 sync Build/Sandcastle2/ s3://cesium-website/cesiumjs/releases/$(cat package.json | jq -r '.version' | sed 's/\.0$//')/Apps/Sandcastle2/ --cache-control "public, max-age=1800" --delete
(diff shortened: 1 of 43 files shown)


packages/engine/Source/DataSources/StaticGeometryColorBatch.js
```diff
@@ -147,6 +147,7 @@ Batch.prototype.update = function (time) {
             this.depthFailMaterial,
           );
         }
+        // eslint-disable-next-line new-cap
```
wewindy: Considering:

```javascript
const DepthFailAppearanceClass = this.depthFailAppearanceType;
depthFailAppearance = new DepthFailAppearanceClass({ /* ... */ });
```

javagl: That looks like a nice... ehm... I wouldn't even call it a "workaround", but rather a _way to make clearer what's happening there_.

donmccurdy: That's a great idea, thanks! Done in d476b89753c07c23028e59ad871e213a24d07928. :)

donmccurdy:
@javagl would you be up for reviewing this one when you have the chance? :)

PR #13271, opened 2026-03-05
Extract Vector glTF tile data into BufferPrimitiveCollection

danielzhong:
# Description

This PR adds support for rendering glTF-based vector tile content using the current `CESIUM_mesh_vector` spec.

The runtime decodes the glTF, extracts vector data into `BufferPrimitiveCollection` types, and renders them through the existing buffer collection renderer path.

## Supported vector primitive handling in this PR

- **POINTS → `BufferPointCollection`**
- **LINE_STRIP → `BufferPolylineCollection`**
  - Primitive restart values are used to split a single indexed line strip into multiple polylines.
- **TRIANGLES → `BufferPolygonCollection`**
  - Polygon metadata is read from `CESIUM_mesh_vector`.
  - The current implementation uses `polygonAttributeOffsets`, `polygonIndicesOffsets`, and optional hole metadata.

## High-level flow

1. `Cesium3DTileset` loads a `.gltf` / `.glb` tile.
2. In `Cesium3DTileContentFactory`, glTF content is routed to `VectorGltf3DTileContent` when `tileset.isGltfExtensionUsed("CESIUM_mesh_vector")` is `true`.
3. `VectorGltf3DTileContent.fromGltf()` calls `Model.fromGltfAsync()` to decode the asset without rendering the model directly.
4. On the first `update()`, once the decoded model is ready, `initializeVectorPrimitives()` calls `createVectorTileBuffersFromModelComponents()`.
5. `createVectorTileBuffersFromModelComponents()` converts glTF primitives into buffer collections:
   - `BufferPointCollection[]`
   - `BufferPolylineCollection[]`
   - `BufferPolygonCollection[]`
6. One source mesh primitive maps to one buffer collection. Local node transforms are preserved per collection and applied through `collection.modelMatrix` during update.
7. On each frame, `VectorGltf3DTileContent.update()` updates each collection and calls `collection.update(frameState)`.
8. Each collection uses the existing renderer path:
   - `renderBufferPointCollection`
   - `renderBufferPolylineCollection`
   - `renderBufferPolygonCollection`

## Testing plan

# Author checklist

### DIFF
--- a/packages/engine/Specs/Scene/Multiple3DTileContentSpec.js
+++ b/packages/engine/Specs/Scene/Multiple3DTileContentSpec.js
@@ -64,6 +64,19 @@ describe(
       return generateJsonBuffer(gltf).buffer;
     }
 
+    function createMockTileset() {
+      return {
+        statistics: {
+          numberOfPendingRequests: 0,
+          numberOfAttemptedRequests: 0,
+        },
+        tileFailed: new Event(),
+        isGltfExtensionUsed: function () {
+          return false;
+        },
+      };
+    }
+
     let originalRequestsPerServer;
 
     function setZoom(distance) {
@@ -177,12 +190,7 @@ describe(
     });
 
     it("requestInnerContents returns promise that resolves to content if successful", async function () {
-      const mockTileset = {
-        statistics: {
-          numberOfPendingRequests: 0,
-          numberOfAttemptedRequests: 0,
-        },
-      };
+      const mockTileset = createMockTileset();
       const tile = {};
       const content = new Multiple3DTileContent(
         mockTileset,
@@ -205,12 +213,7 @@ describe(
     });
 
     it("requestInnerContents returns undefined and updates statistics if all requests cannot be scheduled", function () {
-      const mockTileset = {
-        statistics: {
-          numberOfPendingRequests: 0,
-          numberOfAttemptedRequests: 0,
-        },
-      };
+      const mockTileset = createMockTileset();
       const tile = {};
       const content = new Multiple3DTileContent(
         mockTileset,
@@ -226,13 +229,7 @@ describe(
     });
 
     it("requestInnerContents handles inner content failures", async function () {
-      const mockTileset = {
-        statistics: {
-          numberOfPendingRequests: 0,
-          numberOfAttemptedRequests: 0,
-        },
-        tileFailed: new Event(),
-      };
+      const mockTileset = createMockTileset();
       const tile = {};
       const content = new Multiple3DTileContent(
         mockTileset,
@@ -263,12 +260,7 @@ describe(
     });
 
     it("requestInnerContents handles cancelled requests", async function () {
-      const mockTileset = {
-        statistics: {
-          numberOfPendingRequests: 0,
-          numberOfAttemptedRequests: 0,
-        },
-      };
+      const mockTileset = createMockTileset();
       const tile = {};
       const content = new Multiple3DTileContent(
         mockTileset,


packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,517 @@
...
+  for (let i = 0; i < indices.length; i++) {
+    const vertexIndex = indices[i];
+    const srcOffset = vertexIndex * 3;
+    values[i * 3] = positions[srcOffset];
+    values[i * 3 + 1] = positions[srcOffset + 1];
+    values[i * 3 + 2] = positions[srcOffset + 2];
+  }
+
+  polylineCollection.add({ positions: values }, polylineView);
+  setPrimitiveFeatureId(polylineView, featureIdSource, indices[0]);
```
donmccurdy: I'm hoping we'll be able to upload a range of the position array directly to the collection, but I guess the collection needs to allow more than float64array here? Opened:

- https://github.com/CesiumGS/cesium/pull/13277

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,517 @@
...
+  positions,
+  indices,
+) {
+  let triangleIndices = indices;
+  if (primitiveType === PrimitiveType.TRIANGLE_STRIP) {
+    triangleIndices =
+      ModelReader.convertTriangleStripToTriangleIndices(indices);
+  } else if (primitiveType === PrimitiveType.TRIANGLE_FAN) {
+    triangleIndices = ModelReader.convertTriangleFanToTriangleIndices(indices);
+  }
```
donmccurdy: I think it's OK just to error out if the primitive uses the vector extension with TRIANGLE_STRIP or TRIANGLE_FAN topology, since we won't be able to identify the original polygon connectivity in those cases anyway.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,517 @@
...
+
+  if (!defined(triangleIndices) || triangleIndices.length < 3) {
+    return;
+  }
+
+  polygonCollection.add(
+    {
+      positions: positions,
+      triangles: toUint32Array(triangleIndices),
```
donmccurdy: See above, with #13277 it should be OK to use whatever the source array type is.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,517 @@
...
+  nodeTransform,
+  pointCollection,
+  pointView,
+  polylineCollection,
+  polylineView,
+  polygonCollection,
+  polygonView,
+) {
+  const primitiveType = primitive.primitiveType;
+  const positions = readTransformedPositions(primitive, nodeTransform);
```
donmccurdy: Do we need to transform the vertex positions? I'm hoping that we can apply the node transform to `collection.modelMatrix` or something like that, but not 100% sure how this all fits with the rest of the loading process!

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,517 @@
...
+      loopIndices[indices.length] = indices[0];
+      appendPolylinePrimitive(
+        polylineCollection,
+        polylineView,
+        featureIdSource,
+        positions,
+        loopIndices,
+      );
+    }
+    return;
```
donmccurdy: Here I think it's OK if we throw an error if we see any primitive types not allowed by the vector extension.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,517 @@
...
+      rootNodes[i],
+      Matrix4.IDENTITY,
+      points,
+      pointView,
+      polylines,
+      polylineView,
+      polygons,
+      polygonView,
+    );
+  }
```
donmccurdy, line 617: Ah, important decision here! If the GLB has 50 mesh primitives, do we want 50 BufferPrimitiveCollections, or just 1 of each topology type? I'd been assuming that the tiler would combine mesh primitives as much as it safely can, so the runtime we could output one collection per mesh primitive and just apply the node transform with `collection.modelMatrix`. But there might be more to this, @lilleyse do you have a preference?

donmccurdy (commented):
@danielzhong Great progress here, I like how this is coming together! I left a few initial comments, mainly since we're writing the code and the spec in parallel just trying to make sure things align.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
lilleyse: > I'd been assuming that the tiler would combine mesh primitives as much as it safely can, so the runtime we could output one collection per mesh primitive and just apply the node transform with `collection.modelMatrix`

That was my assumption as well. One primitive maps to one buffer collection.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -0,0 +1,378 @@
...
+/**
+ * Vector glTF tile content. This path decodes glTF primitives into vector
+ * buffers, then renders with dedicated vector primitives.
+ *
+ * @alias VectorGltf3DTileContent
+ * @constructor
+ *
+ * @private
+ */
+function VectorGltf3DTileContent(tileset, tile, resource) {
```
donmccurdy: No change needed right now, but on naming — my understanding is that `VectorGltf3DTileContent` is intended to succeed/replace `Vector3DTileContent` which used the earlier .vctr format. If support for .vctr can be deprecated in the not-too-distant future (I believe that's true?) then it would make sense to rename `VectorGltf3DTileContent` -> `Vector3DTileContent` when that happens, since most of the other relevant 3D Tiles classes don't currently have "Gltf" in the name.

(/cc @lilleyse does this sound correct?)

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -0,0 +1,378 @@
...
+import createVectorTileBuffersFromModelComponents from "./Model/createVectorTileBuffersFromModelComponents.js";
+import ModelUtility from "./Model/ModelUtility.js";
+import BufferPolygon from "./BufferPolygon.js";
+
+/**
+ * Vector glTF tile content. This path decodes glTF primitives into vector
+ * buffers, then renders with dedicated vector primitives.
+ *
+ * @alias VectorGltf3DTileContent
+ * @constructor
```
donmccurdy: If possible, I think we'd prefer that new classes in CesiumJS begin using ES6 classes rather than prototype-based inheritance. We have a script to do this automatically for you, except that you'll need to remove the `@alias` and `@constructor` tags afterward:

```bash
node ./scripts/lebab-batch.js "packages/engine/Source/Scene/VectorGltf3DTileContent.js"
```

If you run into any non-trivial problems, though, OK to leave this as-is.

packages/sandcastle/gallery/3d-tiles-formats/main.js
donmccurdy, line 1: Although I put them first for debugging originally, we should probably add these vector samples to the end of the tileset list, rather than the front.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -0,0 +1,378 @@
...
+    },
+  },
+});
+
+VectorGltf3DTileContent.prototype.getExtension = function (extensionName) {
+  if (extensionName === "CESIUM_mesh_vector") {
+    return this._vectorBuffers;
+  }
+  return undefined;
+};
```
donmccurdy: Is `getExtension` being used? I can't see where it is used, and it wasn't on the Cesium3DTileContent interface, so might be unused code?

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -0,0 +1,378 @@
...
+VectorGltf3DTileContent.prototype.getFeature = function (featureId) {
+  void featureId;
+  return undefined;
+};
+
+VectorGltf3DTileContent.prototype.hasProperty = function (featureId, name) {
+  void featureId;
+  void name;
+  return false;
+};
```
donmccurdy: Nit – Are the `void *` calls are here because the arguments are otherwise unused? I don't think this pattern is used elsewhere in the package, I might prefer to either:

- (A) `throw new DeveloperError('Not implemented')`
- (B) Leave the parameters unused, if ESLint doesn't mind
- (C) Prefix the parameters with `_`

(similar for other stub methods below)

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+
+function readIndicesOrSequential(primitive, vertexCount) {
+  const indices = primitive.indices;
+  if (!defined(indices)) {
+    const result = new Uint32Array(vertexCount);
+    for (let i = 0; i < vertexCount; i++) {
+      result[i] = i;
+    }
+    return result;
+  }
```
donmccurdy: Minor — I might prefer to be less "defensive" here, as we're requiring in the spec that primitives be indexed. It would be OK to assume indices are always available, or to include just a dev-mode assertion:

```javascript
assert(primitive.indices, 'Primitive indices are required');
//>>includeEnd('debug');
```

Partly for simplicity, and partly to make mistakes (if we pass the wrong argument in) more obvious.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+function getMeshVectorExtension(primitive) {
+  const meshVector = primitive.meshVector;
+  if (!defined(meshVector)) {
+    return undefined;
+  }
+  if (meshVector.vector === false) {
+    return undefined;
+  }
+  return meshVector;
+}
```
donmccurdy, line 72: Similar to above, I think we can be less defensive, maybe assert `.vector === true`, and return it without if/else blocks.

packages/engine/Source/Scene/Cesium3DTileContentFactory.js
```diff
@@ -100,6 +102,13 @@ const Cesium3DTileContentFactory = {
       return GaussianSplat3DTileContent.fromGltf(tileset, tile, resource, glb);
     }
 
+    if (
+      defined(tileset.isGltfExtensionUsed) &&
```
donmccurdy: Is there a case where the `Cesium3DTileset#isGltfExtensionUsed` method would be missing? Wondering if we can skip checking whether the method is defined.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+    return;
+  }
+
+  if (primitiveType === PrimitiveType.LINE_STRIP) {
+    const indices = readIndices(primitive);
+    if (!defined(indices)) {
+      throw new RuntimeError(
+        "Vector polylines with LINE_STRIP topology must be indexed.",
+      );
+    }
```
donmccurdy: Nit - a dev-only check might be ideal here, keeping the build size a bit smaller because the warning messages don't minify well.

```js
assert(defined(indices), "Vector polylines...");
//>>includeEnd('debug');
```

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+    forEachLineStripSegment(indices, (segmentStart, segmentCount) => {
+      if (segmentCount >= 2) {
+        stats.polylinePrimitiveCount += 1;
+        stats.polylineVertexCount += segmentCount;
+      }
+    });
```
donmccurdy: Hm, is the segment count equivalent to the vertex count here?

I think we can avoid iteration over the vertex stream with something like...

```suggestion
stats.polylinePrimitiveCount += meshVector.count;
stats.polylineVertexCount += indices.count - meshVector.count;
```

... knowing that if there are N total polylines with M indices, then there are N primitive restart values, and the remaining M - N indices all represent a distinct vertex.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+    }
+
+    const polygonAttributeOffsets = meshVector.polygonAttributeOffsets;
+    const polygonIndicesOffsets = meshVector.polygonIndicesOffsets;
+    const polygonHoleCounts = meshVector.polygonHoleCounts;
+    if (!defined(polygonAttributeOffsets) || !defined(polygonIndicesOffsets)) {
+      throw new RuntimeError(
+        "Vector polygons require polygonAttributeOffsets and polygonIndicesOffsets.",
+      );
+    }
```
donmccurdy: See above — I think it's safe to assume we'll write separate data validation tools, so we don't need to worry too much about invalid tiles here.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+      const polygonIndicesCount =
+        i + 1 < polygonCount
+          ? polygonIndicesOffsets[i + 1] - polygonIndicesOffsets[i]
+          : indices.length - polygonIndicesOffsets[i];
+      stats.polygonVertexCount += Math.max(end - start, 0);
+      stats.polygonTriangleCount += polygonIndicesCount / 3;
+      if (polygonHoleCounts) {
+        stats.polygonHoleCount += polygonHoleCounts[i];
+      }
+    }
```
donmccurdy: Wondering if we can avoid iteration here.

I believe `indices.count / 3` will give us the triangle count. 

The spec currently requires that the position accessor lay out the vertices in a certain order, so I think we can assume that `position.count` is the vertex count.

I guess we have no way to get the total hole count, though, and we'll need to keep iterating (at least over the polygonHoleCounts array) for that? Maybe that's something we should address in the spec, but doesn't need to block the PR in the meantime.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+  vertexIndex,
+) {
+  const offset = vertexIndex * 3;
+  pointCollection.add(
+    {
+      position: Cartesian3.fromElements(
+        positions[offset],
+        positions[offset + 1],
+        positions[offset + 2],
+      ),
```
donmccurdy, line 280: Performance micro-nit - since we're potentially iterating very large point clouds here, I think it would be worth creating a scratch `new Cartesian3()` variable (can be at module level) and passing it into `Cartesian3.fromElements` to avoid allocating one for each point.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+    );
+    values = new TypedArray(indices.length * 3);
+    for (let i = 0; i < indices.length; i++) {
+      const vertexIndex = indices[i];
+      const srcOffset = vertexIndex * 3;
+      values[i * 3] = positions[srcOffset];
+      values[i * 3 + 1] = positions[srcOffset + 1];
+      values[i * 3 + 2] = positions[srcOffset + 2];
+    }
+  }
```
donmccurdy, line 315: (Just a comment, no change needed)

Hm, trying to decide if (similar to polygons) we should require that the polyline vertices be contiguous and in order. Would save us the iteration here – we could basically throw away the indices without looking at them, if we want.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+    if (!defined(indices)) {
+      throw new RuntimeError(
+        "Vector polylines with LINE_STRIP topology must be indexed.",
+      );
+    }
+
+    forEachLineStripSegment(indices, (segmentStart, segmentCount) => {
+      if (segmentCount < 2) {
+        return;
+      }
```
donmccurdy: I think this is a case where hiding bad data might make it more difficult to debug issues - if the segment count is somehow 0–1, I think it's OK to let it go, or maybe do a dev-mode check.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+          ? polygonIndicesOffsets[i + 1] - triangleIndexOffset
+          : indices.length - triangleIndexOffset;
+      let triangleIndices;
+      if (triangleIndexCount > 0) {
+        if (polygonVertexOffset === 0) {
+          triangleIndices = indices.subarray(
+            triangleIndexOffset,
+            triangleIndexOffset + triangleIndexCount,
+          );
+        } else {
```
donmccurdy: I think I probably wouldn't do a special case for the first polygon in the collection, OK to use the else block below for all polygons.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+          const IndexArray = /** @type {TypedArrayConstructor} */ (
+            indices.constructor
+          );
+          triangleIndices = new IndexArray(triangleIndexCount);
+          for (let t = 0; t < triangleIndexCount; t++) {
+            triangleIndices[t] =
+              indices[triangleIndexOffset + t] - polygonVertexOffset;
+          }
```
donmccurdy: (No change needed)

I'm wondering if we should change BufferPrimitiveCollection to store triangle indices as absolute offsets into the collection rather than relative offsets into the polygon. Then we wouldn't need to copy the data here. But not urgent, and doesn't need to block the PR I think.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,667 @@
...
+  const primitives = node.primitives;
+  for (let i = 0; i < primitives.length; i++) {
+    const primitive = primitives[i];
+    const stats = createStats();
+    gatherPrimitiveStats(primitive, stats);
+
+    let collection;
+    let pointView;
+    let polylineView;
+    let polygonView;
```
donmccurdy: Nit - the point/line/polygon views are being passed down through a few function calls for use, it might be simpler to put these as module-scoped variables and access them where needed?

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: I think segmentCount is equivalent to the vertex count for the generated BufferPolyline, since we upload one position per non-restart index.

I agree we can simplify this using `meshVector.count`, but I think the vertex count would be `indices.length - (meshVector.count - 1)` here, since the stream has N - 1 restart values for N polylines in the current encoding.

For example:
[0, 1, 2, RESTART, 2, 3, 4]
n = 2 (polylines), m = 7 (total indices), primitive restart value = 1, So vertex number is 6

But if I calculate using suggested code, it would become 7-2=5

So what do you think about:
```
stats.polylinePrimitiveCount += meshVector.count;
stats.polylineVertexCount += indices.length - (meshVector.count - 1);
```

packages/engine/Source/Scene/VectorGltf3DTileContent.js
danielzhong: Thanks! Fixed:
https://github.com/CesiumGS/cesium/pull/13271/commits/1143deb5e45ef5a49a2d7cf93945da10b0ebb871

packages/sandcastle/gallery/3d-tiles-formats/main.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/commits/97bcc8eb7cb352c0ae599cfd0f001f9171fe4100

packages/engine/Source/Scene/VectorGltf3DTileContent.js
danielzhong: Good catch! Deleted!

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Fixed! Thanks!
https://github.com/CesiumGS/cesium/pull/13271/changes/f82fa5546352bd3672939421e6e7a2b62b64acd4

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Fixed! Thanks!
https://github.com/CesiumGS/cesium/pull/13271/changes/f82fa5546352bd3672939421e6e7a2b62b64acd4

packages/engine/Source/Scene/Cesium3DTileContentFactory.js
danielzhong: We can skip that check here. `Cesium3DTileContentFactory` is operating on a Cesium3DTileset, so `isGltfExtensionUsed()` should always be available. I removed the extra `defined(...)` guard and call the method directly now.
https://github.com/CesiumGS/cesium/pull/13271/changes/f82fa5546352bd3672939421e6e7a2b62b64acd4

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: You are right!
https://github.com/CesiumGS/cesium/pull/13271/changes/f82fa5546352bd3672939421e6e7a2b62b64acd4

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Fixed! Thanks!
https://github.com/CesiumGS/cesium/pull/13271/changes/f82fa5546352bd3672939421e6e7a2b62b64acd4

packages/engine/Source/Scene/VectorGltf3DTileContent.js
danielzhong: I chose C, which seems closer to the style used elsewhere in the codebase
https://github.com/CesiumGS/cesium/pull/13271/changes/f82fa5546352bd3672939421e6e7a2b62b64acd4

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
donmccurdy: Oops, yes, `N - 1` restart values. I'd forgotten the last index wouldn't be a restart value. So your new proposed code in the comment above looks correct to me, let's just add a comment so we remember why. :)

Aside - in my mental model, `segmentCount = vertexCount - 1`, since a line strip like this...

```plaintext
o——o——o——o——o
```

...always has one fewer edges connecting vertices than it has vertices. But maybe we are using the terms a little differently, that's OK.

danielzhong: Thanks, this commit just changed it

https://github.com/CesiumGS/cesium/pull/13271/changes/088decc2a8499522cdd5950468f3a00210a80642

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Agree!
https://github.com/CesiumGS/cesium/pull/13271/commits/ad37d1fde218000b8ce34ed7f32aa4984153d1ad

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Good point!
https://github.com/CesiumGS/cesium/pull/13271/commits/94ec7280615675411f81cdeea56604a99cc9deda

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: I think I’m in favor of that direction. For `CESIUM_mesh_vector` it seems reasonable to require polylines to be laid out contiguously and in order, similar to the polygon constraints, since that keeps more of the work on the tiler side and lets the runtime stay simpler and cheaper.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/changes/7a6127356142dab8fb4cc3e6b85c6543b4cc765c

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Agree!
https://github.com/CesiumGS/cesium/pull/13271/changes/c1f9d7bcb6b1233e8b16868623fd5ec02e43b663

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/changes/c1f9d7bcb6b1233e8b16868623fd5ec02e43b663

ready to review all again!

packages/engine/Source/Scene/Multiple3DTileContent.js
```diff
@@ -591,7 +591,7 @@ function handleInnerContentFailed(multipleContents, index, error) {
   const tileset = multipleContents._tileset;
   const url = multipleContents._innerContentResources[index].url;
   const message = defined(error.message) ? error.message : error.toString();
-  if (tileset.tileFailed.numberOfListeners > 0) {
+  if (defined(tileset.tileFailed) && tileset.tileFailed.numberOfListeners > 0) {
```
donmccurdy: Do you recall why this change was needed? I'm not seeing errors if I remove it currently, maybe it can be removed?

If related to my sample tilesets ... those were just thrown together with a quick script, so mostly I'm wanting to check that we're not adding workarounds for what might have been invalid data on my part. :)

packages/engine/Source/Scene/VectorGltf3DTileContent.js
donmccurdy, line 1: Entirely optional suggestion, feel very free to skip this — For new files you might find it helpful to include a `// @ts-check` comment at the top of the file. If you're using an editor with TypeScript Language Services enabled (like VSCode or Zed) you should start to see better autocomplete, and errors where type information is missing or conflicting. Any errors visible in the editor will also be reported by `npm run tsc`.

For example if #13313 happens to merge before this PR, `npm run tsc` would start complaining about calls to `point.setColor(...)` (which has been removed).

We could certainly do this in a future PR too.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -0,0 +1,319 @@
...
+  isDestroyed() {
+    return false;
+  }
+
+  destroy() {
+    this._decodeModel = this._decodeModel && this._decodeModel.destroy();
+    this._pointCollections = undefined;
+    this._polylineCollections = undefined;
+    this._polygonCollections = undefined;
+    this._vectorBuffers = undefined;
```
donmccurdy, line 270: It would be best to destroy the collections before we dereference them:

```javascript
for (const collection of this._pointCollections) {
  collection.destroy();
}
this._pointCollections = undefined;
```

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,661 @@
...
+
+  if (points.length === 0 && polylines.length === 0 && polygons.length === 0) {
+    return undefined;
+  }
+
+  return {
+    points: points.length > 0 ? points : undefined,
+    polylines: polylines.length > 0 ? polylines : undefined,
+    polygons: polygons.length > 0 ? polygons : undefined,
+  };
```
donmccurdy, line 627: Optional - I think if we allowed empty arrays here, and made the arrays non-nullable, we'd need fewer if/defined protections elsewhere in the PR. But that sort of change is easier to do safely with the `// @ts-check` mode on, which might be out of scope for this PR, so feel free to leave this as-is too. :)


```suggestion
  return { points, polylines, polygons };
```

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,661 @@
...
+        triangleCountMax: stats.polygonTriangleCount,
+      });
+      polygons.push(collection);
+    }
+
+    if (!defined(collection)) {
+      continue;
+    }
+
+    collection._vectorLocalModelMatrix = Matrix4.clone(nodeTransform);
```
donmccurdy, line 563: Hm, I was hoping it might be possible to avoid putting extra data onto the collection, but I don't see an obvious way right now. Not sure if that should be the responsibility of the Cesium3DTiles classes or if the BufferPrimitiveCollection needs more metadata. Ok with me as-is though, no changes needed for now unless you have more thoughts on it.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,661 @@
...
+
+    for (let i = 0; i < indices.length; i++) {
+      appendPointPrimitive(
+        pointCollection,
+        scratchPointView,
+        featureIdSource,
+        positions,
+        indices[i],
+      );
+    }
```
donmccurdy: Since these could be large point clouds, maybe we can do without allocating sequential indices if they're missing? For example:


```suggestion
    const indices = readIndices(primitive);
    for (let i = 0, il = indices ? indices.length : vertexCount; i < il; i++) {
      appendPointPrimitive(
        pointCollection,
        scratchPointView,
        featureIdSource,
        positions,
        indices ? indices[i] : i,
      );
    }
```

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,661 @@
...
+
+  featureId = Math.trunc(featureId);
+  if (
+    defined(featureIdSource.nullFeatureId) &&
+    featureId === featureIdSource.nullFeatureId
+  ) {
+    return undefined;
+  }
+  return featureId;
+}
```
donmccurdy, line 163: It seems like we're re-implementing a bit from `packages/engine/Source/Scene/Model/FeatureIdPipelineStage.js` here and in `getFeatureIdSource` above. That's probably OK, I don't see an easier way to access its results, maybe it's just worth including a TODO for later.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -0,0 +1,659 @@
...
+    //>>includeStart('debug', pragmas.debug);
+    assert(
+      defined(indices),
+      "Vector points with POINTS topology must be indexed.",
+    );
+    //>>includeEnd('debug');
+
+    stats.pointPrimitiveCount += indices.length;
+    stats.pointVertexCount += indices.length;
+    return;
```
donmccurdy: Unlike polylines and polygons, points don't need to be indexed - I think we could just use `meshVector.count` unconditionally here rather than checking for indices.

(I fixed a bug in the sample tilesets, and the point cloud sample currently hits an error here)

donmccurdy (approved):
Awesome work, thanks Daniel! Just a few comments, no major concerns on my side.

I'd slightly prefer fewer defensive checks still, but that's easier to do with the `// @ts-check` suggestion which might be out of scope here, so not a big deal for now. Tests are working well locally.

Heads up that in earlier version of the sample tilesets I generated, the "extensionsUsed" property wasn't declared properly and the tilesets could go down the existing glTF rendering path rather than using the new vector code. I've fixed the tilesets: polylines and polygons are working as-is, there's a small fix needed for points I think (see comments).

Approving now since the comments are mostly optional stuff, the point primitive fix is the only critical thing.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
donmccurdy: (I went ahead and pushed a small fix for this, since we'll want it in for alpha testing soon)

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
donmccurdy: (I went ahead and pushed a small fix for this, since we'll want it in for alpha testing soon)

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
lilleyse: I agree 👍

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/commits/5fcd46f930a97bcebfeb7d5d8abd66880ba46588

packages/engine/Source/Scene/Multiple3DTileContent.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/commits/373817ac8f48411ae6acc7594b0cec4ce629e88f

packages/engine/Source/Scene/VectorGltf3DTileContent.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/changes/b3e30f22fc282248d17a29f7104f4bd1b0475a9f

packages/engine/Source/Scene/VectorGltf3DTileContent.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13271/commits/49bd15cc77f356f21c0a0190828718497fd95633

PR #13273, opened 2026-03-06
Add autoUpdate option to ClippingPolygon

alarkbentley:
# Description

Adds an `autoUpdate` option to `ClippingPolygon` that, when explicitly set to false, disables internal change-detection when computing bounding rectangles which can significantly improve rendering performance for ClippingPolygons with many vertices. For backwards compatibility `autoUpdate` will default to `true` but in future updates the default value will be set to `false`.

### `ClippingPolygon`

- Added `autoUpdate` constructor option (default `true`).
- `computeRectangle` skips the position-comparison fast path for non-autoUpdate polygons and returns the cached rectangle directly on subsequent calls.
- `computeSphericalExtents` implements caching similarly to `computeRectangle`.
- Added a deprecation warning that we plan to change the default behavior `autoUpdate` from Cesium.js v1.142.

## Demo

Based on exising sandcastle "Clipping Regions". Creates 5 clipping polygons with 1000 vertices each and attached onto GooglePhotorealistic3DTileset.

[Sandcastle](https://sandcastle.cesium.com/#c=7Vjbbhs3EP0VwujDqlFWkg1bdiwLdZ3UDeDAQuw0QOMApnZHEhGKFEiuVSXQv3d42asuKYImL60BWUvOcHh4ZnjIFZsvpDLkZ0I1uQLNsjmZKDknDweJaz0cnD+oB8G82x0VaUK14ZB7lT3B034SKbQhTwyWoMgFEbAMseM/XF+UR7+SwlAmQD0ctMkXO5QQw+bAse8FmVCuoe17KUKghklR79YJCHgjUxix5BOounFMNdzQFahtxinIBMdhb4D2Wopr3zdS8omh6X61gPj69vb65pUbtG6dl6tzM+Pi/Cpj1ywI6HTIK0HHSJMCgZGYmBIzA6I/razZOcfYuDRzqRczUNiaySWGMyorw1RjxymMs+kdev2m6Bz0CNQdIJS0OYiDIUupeHoPSiG5zmDUKue3asOxdEmZyTlIFFAD7ysel3olksgvPNAdG2/KacIgG/PlrlMux8XaHP3OvCYJNcmMRDhEqlYBjYlULmPKQZno8d7yQpZYmFQQ50gcPkumm5AEIDH56Yuzrx890PUmF4yDBrOdC2/bzsW1lFMOo5k0EtucacOSo5dhSBTCEIIJvxV85ZLsh5BQTAga4Y+BZBpSXKGZ5Q71oOToJXFhY0LuEI2N9BiK9JEslFwgKSsiJ84S9pYrRZUlGIfIhd0emhhJ9pZynGOWCPidRjbfIyiP6brYFLakwk5Z17O/UGzODHtCpDRNoyqFrX3JtVglh5jLafT4yqWTS5rabO5gAqXAhS0Q53+70o1ZuEzTfJjlyvImLfFWg8YZ43a+vDLy9o7iaJib9eG/ilqIrSAi8ZcaG6/TqHfS7572Tvdy15jhW+nLwzT42sPSnxLFGwvFllKVGQyZOJUtdS5BsVH0djLxFFS0/Hdw048Yon1LxRQiP28wv6FmFhv5Fp2o0NHZcdxttfd5PO+dli4n3W7cdc8efZDCzwj8XjaJa9dQtqoyfC+ndq+NwSwBBHF6lOsG6kpKqOuj3FZd4M8OLc81m6l7KfmYqjcgsuiDB1hsfQN/mRd4EuZVi0dZucM0cEjQHLXIxbAc8xV59C7VnbVxPFiHdb5B2zsgBUX+ZkS12bYDqoOuI/q4NRN5qTFhK+qJaTZmnJnVNtbtiF8zY6TAK4M9+8o9jBcGp1FkkonEliyJkhngOZ+2dmzgHHJwO68c6fZTxPHCf8VUgsIvmTA6QmqwZm6kwErzjxSLTmHhZrpNRDb3frXNasjCdeKMHz4Ghia4ayMrPQx7u+f4NShHY/PZs1aZEh+EoicjnYpb3c6lPcULgOQZcdsKdT2ieK1zjdFrfDps4T8PuRmCmjIEPocQidT/IIRfZLzI9Cz6wC1DGO5jLntr/6XAZEoE37oaNWgfSb6aIqqIZka+W6TYFQj2/RfHFb4vet1ud4N071il3VssT5VINds8t1UYbiRxS1mUW+N5r4v61u33e93Ds/7hSf+oTSqn1tFZ3D/unZ2e9Hv9s9P+CS4C5a172Cbz1r7SaJZEvjhPd0WLrzhbLKwYe4eosq/x/GPuZlDcdK+oMvhExZE7tF7CVAHoS9SKVVQ9akNqJ5jQ6LDVLk3V50qeChVo7Up/hfn1pjQkYRFlDsFdotOvSkO4bBcBogxv3RN8iUhbVij8vX+PUljetxUOm2z6kqpno2jLaesEVLU1qadK1w/VRiKvJLdyzaopzad/UTw1L2o1of7u8623yfy/mMuywi6s4v/YjLoZ/0/m90imy+GPzaaf8r+ZzoP2wUCbFYehtf8SftHJFI/iuGNgvkCZB90ZZ8gmItMah4pBJx8ySNkTYenFlp9uML1Ua7RMMs7v2Gd4OBgOOuhfGxbeVm6fQHG6si6z3vDGd8ZxPOhgc3OU8TdvdLeoByaUk5+PiYkcUQE8mK3DWKarYX44DYwalufZwKTDG5gYW47JJ/v+gwWLFyxEhDdmRID2YmSnMnQzzFs2nTXi2HepkIT81aoo+6RI3c5Z8LGAjs92nZaIQMjf)

* Enable clipping (undefined) -> ~20fps (will show deprecation warning in console)
* Enable clipping (autoUpdate=true) -> ~20fps
* Enable clipping (autoUpdate=false) -> ~60fps

image

## Testing

- Verifies `autoUpdate` defaults to `true` on regular construction.
- Confirms `autoUpdate` = false polygons do no re-caclulate `computeRectangle` on vertex mutation.
- Confirms `autoUpdate` = false polygons do no re-caclulate `computeSphericalExtents` on vertex mutation.
- Confirms `autoUpdate` = true polygons still perform position comparison and detect mutations.

Run the `ClippingPolygon` spec suite:
```
npm test -- --include Scene/ClippingPolygon
```

## Issue number and link

# Author checklist


alarkbentley:
@javagl This is an suggestion how we can introduce immutability in the API without breaking changes. Same pattern could be applicable for other objects than ClippingPolygon. Thoughts?

As demonstrated in https://github.com/CesiumGS/cesium/pull/13255 ensuring the immutability for ClippingPolygon has potential to double the final rendering performance. There are probably other use-cases within Cesium.js that could be optimized assuming immutability.

javagl:
I'll have a closer look at this (and the linked PR), maybe with an early pass today (or during the weekend). From the textual description here, it sounds like a reasonable approach. 

In view of the question of whether this approach can be applied more broadly: I already mentioned that the question about immutability could warrant a deeper discussion. There is the differentiation of "immutable" and "unmodifiable", aspects that exist in <strike>proper</strike> other programming languages, like `const` or `final` (or `@readonly`) with their differences that already are related to whether these terms are meant to be "deep" or "shallow", and the implementation efforts that can go along with trying to model all these things _cleanly_. 

(One _seemingly_ shallow point that I'll look at in the context of the review: The change here refers to `ClippingPolygon`. Now we also have that `ClippingPolygonCollection`. How does that relate to `immutable`? Can you mix mutable and immutable polygons in one collection? Maybe these are non-issues, but in order to be _sure_ that they are non-issues, they have to be thought through...)

Also very broadly, but still a bit more related to this PR: 

In many, many (faaar too many) places of the codebase, modifications are tracked with `dirty` flags. That's often straightforward. Naïvely applying this to the clipping polygons would be with pseudocode like
```javascript
class ClippingPolygon {
  constructor(...) {
    this.positions = ...;
    this.rectangle = ...;
    this.dirty = false;
  }

  computeRectangle() {
    if (this.dirty) {
      this.recangle = reallyComputeIt();
      this.dirty = false;
    }
    return this.rectangle;
  }

  setPositions(...) {
    ...
    this.dirty = true;
  }
}
```
And something like this _can_ make sense, in some cases. 

**But**... it can turn into a maintenance nightmare with zero chance of figuring out what "dirty" means ([point in case](https://github.com/CesiumGS/cesium/blob/a15123a641c8de008fe8dd0fa0cab3fa34124bf2/packages/engine/Source/Scene/Scene.js#L1870)). 

And it can only be modeled _cleanly_ when there is clear information hiding. Specifically, when users can _not_ `drill.holes.through.the.api.to.modify.things`. The lack of information hiding leads to things like [this](https://github.com/CesiumGS/cesium/blob/becdc1ff15c72d142446e7e2351dfd96f5969caa/packages/engine/Source/Scene/ClippingPlane.js#L136), which is _very_ related to this PR and the problem that it is trying to solve. 

The next step/level of "clean modeling" would therefore not only about dealing with the question of where to set some `dirty` flag to `true` or `false`, but also about "listeners". I'd try to avoid the acronym "MVC", because this is only about the "M" for now, but ~"something like this": A clearly defined state space, with a clear infrastructure for _modifying_ the state and _reacting_ to modifications - even if that "reaction" is only to set some `dirty=true`...

(I could probably have reviewed the PR instead of writing this 🙂 but I think it's necessary to talk about these things)

alarkbentley:
Your points makes sense and I agree that the core problem with immutability/unmodifiable is a complex one even outside of JavaScript and nothing I had as a goal to completely solve with this PR.

What I'm trying to find with this PR is a common pattern we could introduce for the user to implicitly _hint_ that the objects should be considered immutable to allow internal implementation to take some short-cuts without breaking existing API. There are no  guarantees that internal object actually cannot be modified (other than the shallow Object.freeze safeguards). So if the user implicitly sets the immutable property to true and intentionally modifies some internal parameter (`drill.holes.through.the.api = 6`) that would be entirely possible but then that would be treated as a user error. (this hint behavior needs to be clearly documented)

For your question around the `ClippingPolygonCollection`, the idea would be to do a similar pattern. Introduce an optional "immutable" parameter and if set we would freeze the internal ClippingPolygon array. (ClippingPolygonCollection.add, ClippingPolygonCollection.remove would throw) As stated above we would not guarantee and deep freeze of the internal objects that would be up the users discretion. Also I don't think we should introduce the immutable flag until we have a clear use-case that applies to ClippingPolygonCollection, as long as we have a common non-breaking pattern to follow when we do.

javagl:
> ... the core problem with immutability/unmodifiable is a complex one even outside of JavaScript and nothing I had as a goal to completely solve with this PR.

I see the point on the one hand. The mutability in this specific case causes actual performance issues. It can (probably) be solved with this specific form of that `immutable` flag. And in view of this: 
 
> What I'm trying to find with this PR is a common pattern we could introduce for the user to implicitly hint that the objects should be considered immutable to allow internal implementation to take some short-cuts without breaking existing API.

Such a flag could probably be applied in other places, with the strong and important benefit of not breaking existing API usages. 

But I'd still hesitate to see this as "THE" (or even just "A") common pattern for establishing the concept of immutability. 

The example of `ClippingPolygonCollection`->`ClippingPolygon` was only a specific example. But more generally: I think that it's hard to "scale" that approach through existing classes and structures. You'll have to decide on the meaning of that flag for each case. For example, you could have some structure that contains
`positions: Cartesian3[]`
`color: Color.RED;`
`transform: Matrix4`
And you'll have to decide which of them are affected by `immutable`. One could make a strong case for the blanket statement: "_All_ of them!". But that could be overly constraining (and unnecessary). You may not want to change the positions of some polyline, but still want to be able to render it with a different color.  

Conversely, there's the danger that this flag becomes ubiquitous in the codebase. Eventually, there will be functions with pseudocode
```
function changeSomething(value) {
   if (this.thing.immutable) {
     this.thing = new Thing(value);
   } else {
     this.thing.value = value;
  }
}
```
where forgetting that `immutable` check could cause the function to throw pretty unexpectedly.

alarkbentley:
Right, the vague definition on what actually immutable flag means for a general object is one of the risks here and why I wanted some input on this before going further. Currently that would be defined on per object basis. But there would be very powerful to have an awareness if an object is _intended_ to be immutable.

Two other possibilites that would not break API:
* Treat the "immutable" as a pure hint and keep the implementations by nature mutable (eg. no Object.freeze of internal variables, allow cloning into). However allow optimizations based on the immutable intent.
* Introduce immutable variants of the objects eg. ImmutableClippingPolygon that is implemented in a way that really inforces immutability. (Scalability?)

Any other ideas? 

@ggetz Any input from you on this topic?

javagl:
The bar for introducing this flag for the _specific_ case of `ClippingPolygon` is relatively low. There's an immediate issue and need and benefit, and even if such a flag is not considered to be the ~"universally applicable, scalable, long-term solution", it can be justified here.

(So I think that it could make sense to wrap up, review, and merge this PR, if others agree that this is OK here)

---

That said: There are many places where similar questions already did arise. Things like the `UpdateChangedCartesian3` are patches that are raising more questions than they are answering (just like me 😬 ). The concept of `dirty` flags can be OK, but requires a certain care and diligence, and otherwise can grow out of hand. One thing that is essentially the same as the `immutable` flag is in fact the [`Property.isConstant`](https://cesium.com/learn/ion-sdk/ref-doc/Property.html#isConstant) flag: When this is `true`, then all the updates can be skipped. 

> Introduce immutable variants of the objects eg. ImmutableClippingPolygon that is implemented in a way that really inforces immutability. (Scalability?)

A bit of a guess what the "_(Scalability?)_" refers to: It _can_ be a bit overwhelming (or appear to be "overengineered") when there are two flavors (classes) for each and every "thing", once in its mutable and once in its immutable form. But I think that this approach _can_ be a good baseline approach: 

1. Define that "thing" in its **immutable** form, as a class `Thing`. This is what is used everywhere, by default.
2. If and only if it is absolutely necessary to perform modifications to that "thing", then define a `MutableThing extends Thing` class that offers functions for modifying it, with _clearly_ defined semantics

This helps to realize that many things don't need to be "mutable", and forces developers to think deeply about which modifications are necessary and what their implications are. 

That's not a silver bullet either (and still leaves a few questions unanswered). Consumers and users still don't know what they are dealing with. They could have something like 
```
function set(polyline: Polyline) {
  this.polyline = polyline;
  this.boundingBox = computeBoundingBoxFor(polyline);
}
```
They _receive_ that (seemingly immutable) "Polyline". But it may in fact be a `MutablePolyline`, meaning that it may be modified (from outside!) in a form that would require a recomputation of that "bounding box". 

There are different solutions for that. In some cases, one could keep the `Thing` and `MutableThing` classes completely separate. But that may imply redundancies and a large overhead. (The duplicated functions for `const` and non-`const` in C++ are a shallow form of that). 
In other cases, there may be some "listener" concept, as in
```
  this.polyline = polyline;
  this.polyline.addListener(() => { recomputeBoundingVolume(); }
...
```
But that may not be feasible for many low-level, data-heavy, graphics-oriented entities. It's more suited for higher-level "Model" classes.
Another thing that can go a long way is _documentation_:
`/** ...Assumes immutability. If the 'thing' changes, call xyz() for an update ...*/`

Finding the "right" solution for each class is one of the core challenges. 

---

A bit of an anecdotal aside:

I've recently been playing with Gaussian splats, For lack of better knowledge and familiarity with the subject, I initially jotted down a [`Splat`](https://github.com/javagl/JSplat/blob/8d768492246888b94f4ca4a051f35fe41e83781b/jsplat/src/main/java/de/javagl/jsplat/Splat.java#L32) and [`MutableSplat`](https://github.com/javagl/JSplat/blob/8d768492246888b94f4ca4a051f35fe41e83781b/jsplat/src/main/java/de/javagl/jsplat/MutableSplat.java#L32) interface 🤷‍♂️ 

The following is a screenshot of 2 million splats:

SplatViews 0001

The actual data for that is a `List<Splat>`. On the CPU side (not the GPU/rendering side!), this list of 2 million splats occupies less than 100 _bytes(!)_ of memory. Useless, but... kinda cool, I think.

javagl:
I'll have a closer look at this soon, but maybe @ggetz still wants to chime in here.

donmccurdy:
Prior to this PR, ClippingPolygon iterates over all vertices, every frame — is that correct? Not an objection to this PR, but: unconditional O(n) operations in the render loop would ideally be avoided, whether the data is mutable or not. That might be a target to work toward over time.

An alternative might be renaming the new property from `.immutable` to something like `.autoUpdate`? I think the implementation could be essentially the same, but changing the meaning from "X will not change" to "X will not try to detect changes automatically" and giving the user some method to trigger change detection. Then this optimization is not limited to immutable data.

Finally, some interesting parallels here to the recently-added BufferPolygon API:

- https://github.com/CesiumGS/cesium/pull/13212

The data model is not immutable, but does impose a fixed-length buffer allocation. It might be too soon to consider an equivalent "BufferClippingPolygon" API, so I wouldn't block this PR... but a similar change detection method might be worth considering, it's very fast even for large N and large M.

javagl:
> Prior to this PR, ClippingPolygon iterates over all vertices, every frame — is that correct? Not an objection to this PR, but: unconditional O(n) operations in the render loop would ideally be avoided, whether the data is mutable or not. That might be a target to work toward over time.

That is correct. Before another PR, there was some _really_ costly computation (far more costly than the current 'iteration') that was done in each frame. Now, this costly computation is only done when the polygon changes. And it basically _never_ changes (except for _very_ rare and special cases). But that "change detection" (comparing the current positions to the ones that it had previously) is now the most costly thing.

(An aside: That check - `equalsArrayCartesian3` - looks like a big chunk in the performance trace. But the call hierarchy is deep, and many ~"things" are done there. I'm sure there are _other_ low-hanging fruits. But let's focus on this one for now, even if it raises the broader question about immutability)

---

> An alternative might be renaming the new property from `.immutable` to something like `.autoUpdate`? 

The current semantics/usage of this flag is essentially straightforward
``` 
function update() {
  if (this._immutable) return; // nothing to do
  else doExpensiveUpdate();
}
```
I'm not sure if I fully understood how that "autoUpdate" should change this, or what the "method to trigger change detection" should be. Or is this primarily about the expectations that users may have, based on the name, and the "contract" that may be established by that?

---

> Finally, some interesting parallels here to the recently-added BufferPolygon API:

I had a short(!) look at the code. There is some ["dirty"-handling](https://github.com/CesiumGS/cesium/blob/229830b0ba837c952afa6c8fc3d1238539f3067c/packages/engine/Source/Scene/BufferPrimitiveCollection.js#L464). If where _was_ any encapsulation in `ClippingPolygon`, then something similar could be applied. But this encapsulation (as it is given by the flyweight `BufferPrimitive` that sets the dirty flags) does not exist here. 

Another aside: I mentioned some cocerns about "dirty" flags above. And said that it can be OK, with the appropriate care and diligence. This care and diligence is _exactly_ that the state handling should be hidden. Specifically: That `dirty` flag should be **(really!) private**, to avoid others randomly querying or setting this flag as they please.

---

One solution that allows some form of decoupling can be that of a "modification counter". This is just a counter that is incremented on each change, and users can check whether it was increased since they last saw it. That is also not ideal (because the counter has to be exposed), and it could be dismissed as a "`dirty` flag on steroids". But it can make sense in cases where it's not clear _who_ should set such a `dirty` flag to `false` - and when.

Each paragraph has its 'aside': I didn't find an appearance of `dirty = false` in the changes of that linked PR, but it may already be at another place. (I didn't check all 33 appearances of that in the codebase).

donmccurdy:
> I'm not sure if I fully understood how that "autoUpdate" should change this, or what the "method to trigger change detection" should be. 

For example:

1. if `.autoUpdate === false` then we skip deep comparisons on `this._cachedPositions`, we just trust that it's current if it exists
2. if the user calls `.dirty = true` (setter) then we ignore or delete `this._cachedPositions`, and it's rebuilt on next render

> Or is this primarily about the expectations that users may have, based on the name, and the "contract" that may be established by that?

If an immutable flag were rolled out more widely than ClippingPolygon, then I think the expectations and the "contract" are somewhat difficult, yes. An "auto update" flag has its own difficulties, but we might prefer those difficulties.

> One solution that allows some form of decoupling can be that of a "modification counter". 

three.js does this:

```javascript
/**
 * Setting this property to `true` indicates the engine the texture
 * must be updated in the next render. This triggers a texture upload
 * to the GPU and ensures correct texture parameter configuration.
 *
 * @type {boolean}
 * @default false
 * @param {boolean} value
 */
set needsUpdate( value ) {

	if ( value === true ) {

		this.version ++;

	}

}
```

The user calls `.needsUpdate = true` (effectively a dirty flag) and internally a version counter is incremented. The `.needsUpdate` flag does not need to be reset, its value is always true, and the renderer just checks if the version has changed. Potentially, multiple renderers could be synced to different versions so the resource would be dirty for one and clean for another.

***


Or, to be really opinionated about work in the render loop ... we keep scanning `this._cachedPositions` for ~3 upcoming releases, logging a deprecation warning if there's any change, saying "please make a new ClippingPolygon". Then stop scanning entirely after ~3 releases. I'm not necessarily opposed to a concept of immutability — but in this case I think it's taking work that shouldn't be happening at all, and hiding it behind a code branch that we'll have to remember to maintain and test.

javagl:
> ... we keep scanning .. for ~3 upcoming releases, logging a deprecation warning if there's any change,

As mentioned in some other places: I think that the case where someone _does_ modify the positions is rare. The only scenario would be something like an interactive editor for the polygon in CesiumJS. But that's (IMHO) a very reasonable use-case (going so far that I'd say that CesiumJS _itself_ should offer that, including some `saveAsGeoJson` function). So we should not completely _dismiss_ this. 

What you said sounds like a reasonable solution for me, but it's not my call. (Of course, replacing the clipping polygon with a new one will cause some sort of "flickering" (everything causes some sort of flickering...). But I think that making the clipping polygon immutable could be a cleaner way to resolve the performance issue for the most common case)

javagl:
Just a minor cross-link to what I just stumbled over: This "good enough" solution could probably become obsolete soon: https://github.com/CesiumGS/cesium/blob/8df4d52f4d737a04c4fa006a9029f336e328bec1/packages/engine/Source/Scene/ClippingPolygonCollection.js#L539

alarkbentley:
> Just a minor cross-link to what I just stumbled over: This "good enough" solution could probably become obsolete soon:
> 
> https://github.com/CesiumGS/cesium/blob/8df4d52f4d737a04c4fa006a9029f336e328bec1/packages/engine/Source/Scene/ClippingPolygonCollection.js#L539

This sort of assumes that the ClippingPolygons are already immutable right? Modifying a vertex will not trigger an update, however adding or removing would.

So.. how do we proceed with the PR? Do we need input from more people or are we ready to merge?

donmccurdy:
Yeah, I don't know how to verify it but I suspect this per-frame loop over the vertices wasn't fully intentional. My preference might be to:

1. deprecate the change detection for 1–3 releases
2. if changes are detected during the deprecation period, log a warning about the deprecation
3. in the meantime, expose a temporary `experimentalPositionsChangeDetection: false` option to disable the costly checks before the deprecation period ends

As far as broader performance improvements, we have a few things in progress now, so I'm hesitant to commit to a longer-term "immutable" flag just yet.

javagl:
> This sort of assumes that the ClippingPolygons are already immutable right? Modifying a vertex will not trigger an update, however adding or removing would.

This is one aspect of the discussion that we could dive into, very deeply. Namely the difference between _structural_ modifications and _non-structural_ ("content-only") modifications. There's a huge difference between an operation that boils down to modifying some pixel in a 'texture' and actually _creating_ a _new_ texture, with a different size.

More specifically, about that linked line in the code: 

Yes, it is some sort of half-baked assumption about one form of mutability. This check could be replaced with a check for the `immutable` flag. But this is _exactly_ one of the instances that I brought up in the [first comment here](https://github.com/CesiumGS/cesium/pull/13273#issuecomment-4011407679): 

> The change here refers to `ClippingPolygon`. Now we also have that `ClippingPolygonCollection`. How does that relate to `immutable`? Can you mix mutable and immutable polygons in one collection?

The check is currently counting the `totalPositions`. It could check whether _all_ polygons in the collection have `immutable===true` instead. But when even just one of them is `false`, something has to happen, and it's not clear _what_ should happen. All this is just not properly thought through, causing a lot of churn now.

---

On a more constructive note: 

I think that the proposed approach of deprecating the "automatic updates"/"change detection" could be a viable one. The _vast_ majority of users will _not_ modify clipping polygons. So getting rid of the _assumption_ of mutability could bring an immediate performance benefit for all of them. 

One issue for that approach is that during that "deprecation period", it might not actually _deliver_ that benefit, because the change detection still has to be done. Maybe there's a workaround for that. Maybe even one along the lines of that `UpdateChangedCartesian3` that I'd generally discourage otherwise. 

But I'm considering some options here:

In all cases, there could be that `autoUpdate` flag to _enable_ automatic updates (or `immutable` to _disable_ them - that's just a 'flip'). The suggestion was that during the deprectation period, `autoUpdate` defaults to `true`. Later, it will default to `false`.

Now, how the deprecation could be handled:
 
- Option 1: During the deprectation period, if someone does a change, log a message, saying that this update will _not_ have an effect in the future, unless the `autoUpdate` flag is explicitly set to `true`
- Option 2: During a deprecation period, if someone creates a `ClippingPolygon` without _explicitly_ setting `autoUpdate` to either `true` or `false`, then print log a message saying:
  - this flag currently defaults to `true`, but will default to `false` in the future, and
  - this flag should usually be `false` for better performance, or to `true` for handling updates.

The point is: Option 2 would make users aware that they can just pass in that `autoUpdate:false` to immediately have the (full) performance benefit, _without_ us having to keep track of whether some `p.positions[i].y += 123` is called somewhere...

donmccurdy:
I think (2) would be my preference, with one change — the new flag (whatever it's called) is temporary and removed after the deprecation period. Afterward, there is no option to iterate all vertices on every render, we just won't do that. 

So, essentially the deprecation warning is saying: `Changes to ClippingPolygon positions will no longer be detected automatically after v1.XXX. To opt-in to this behavior sooner, and hide this warning, enable ".someOption". If this change negatively affects your application, please let us know.`

alarkbentley:
@javagl @donmccurdy I changed the `immutable` flag to `autoUpdate` and added a deprecation warning. Does this align with your preferences?

.github/pull_request_template.md
donmccurdy, line 1: Probably not intended changes to this PR template?

donmccurdy:
@alarkbentley I'm not sure whether you and @javagl agree with my suggestion above in https://github.com/CesiumGS/cesium/pull/13273#issuecomment-4091904137, but if so: We are trying to reach an end-state where ClippingPolygon no longer traverses all vertices on every render, exposes no option to do so. That being the case, I think we need to decide on one of two paths forward for `computeRectangle()` and `computeSphericalExtents()`:

- (a) both methods are deprecated now, and removed after the deprecation period in favor of readonly `.rectangle` and `.sphericalExtents` properties, computed once at initialization.
- (b) both methods remain available, but are no longer called during the .update() loop, and will unconditionally recompute the cached rectangles. If users modify the positions array, they are responsible for manually calling `.computeRectangle()` to rebuild the cached rectangles.

I'm happy with either of those. I think (a) is a bit easier to explain, but if there are larger performance implications perhaps that's more important. Does that make sense? I think now the PR is in a limbo-state where it deprecates the old behavior, but leaves it a bit unclear what users should do now.

.github/pull_request_template.md
javagl: Maybe related: 

I just had the commit hook failing after a `merge main`. One reason was that the 
`<!-- markdownlint-disable-next-line MD025 -->`
was apparently ~"not properly picked up", and it reported an error for 
`# Author checklist`
being the second "top-level headline". Why is it a top-level headline? No reason, I guess.

(The other reason was that some of the `...Shaders/...Buffer...Colllection.js` files could not be found, unless I ran a `npm run build` on the merged state - not sure what to do with that...)

donmccurdy: > The other reason was that some of the ...Shaders/...Buffer...Colllection.js files could not be found, unless I ran a npm run build on the merged state - not sure what to do with that...

I believe `npm run build` might be necessary after any commit adding new `.glsl` files. I worked around it by running a build before the CI checks types for example...

https://github.com/CesiumGS/cesium/blob/f5480bcd69bf403ec09b6b05492fb155e7e30077/.github/workflows/dev.yml#L28

... but maybe there are better options.

mzschwartz5:
FYI, the clipping polygon (and collection) classes are changing significantly as part of [this stacked PR](https://github.com/CesiumGS/cesium/pull/13660) (tip as of writing). The SDF-based implementation is being replaced with a vector-style approach.

I'm planning to implement something like what is discussed in this PR after those changes. It doesn't appear a consensus was reached here, but I'm in favor of computing the polygon rectangle once on initialization. ~~We don't even have to deprecate the methods, as they're marked private already.~~ (This is only true for `computeSphericalExtents`).

I also think we should move to positions being immutable (whether that means freezing them or returning a copy on read). If a user wants to change the polygon positions, they can remove and re-add it. This is simpler; we can revisit position mutability if a need arises. Changing positions would mean the polygon has to communicate the change back to the collection, and, with the new vector approach, many textures would have to be recreated. 

Immutable positions may be considered a breaking change, and I think I'll just document that in the change log unless there are strong objections. It seems unlikely to me that many (or any) workflows rely on being able to change clipping polygon positions individually.

javagl:
> It seems unlikely to me that many (or any) workflows rely on being able to change clipping polygon positions individually.

Does this refer to the case of _modifying_ a single vertex position, or also to _adding_ a new vertex? That differentiation between 'modification' and 'structural modification' is important. I could very well imagine the case where someone has a building tileset loaded in CesiumJS, on top of GP3DT, and then wants to manually define the clipping region with a few clicks that define the outline of the building. (FWIW, this is exactly what is currently demonstrated in https://sandcastle.cesium.com/index.html?id=clipping-regions )

mzschwartz5:
> Does this refer to the case of modifying a single vertex position, or also to adding a new vertex?

Both

> FWIW, this is exactly what is currently demonstrated in [sandcastle]

Not exactly - that sandcastle creates a primitive shape dynamically as you left-click around, but it only creates the _clipping polygon_ when you terminate the shape, on right click. So the clipping polygon itself is not dynamic.

javagl:
> So the clipping polygon itself is not dynamic.

But wouldn't it be nice...? 🤔 😁 

As mentioned earlier in the discussion here: I think that the vast majority of users will define the polygon _once_, and then put more emphasis on performance (even for many complex polygons, e.g. outlines read from GeoJSON), precision (straight lines being straight lines all the way down), versatility (holes and such), and consistency (no shadows from clipped parts). **Iff** truly interactive editing of the polygons becomes important, it will likely require different code paths in many places, and should then be planned accordingly. So I think that making them immutable (and mentioning that in the change log) is reasonable.

PR #13275, opened 2026-03-06
Update keys/tokens in CI

### LINKED ISSUE #13225: Update Sandcastle iTwin share keys
Currently we have 3 sandcastles that use data from 2 itwins. To access data from these itwins we use special [share keys](https://developer.bentley.com/apis/access-control-v2/operations/get-itwin-shares/) that last for a year.

The current keys will **expire on April 24, 2026**.
If we want to avoid an out-of-band release to fix them we need to get new keys in by the April release.

This issue is to track that and also the effort around documenting this process so when it comes up again next year we don't have to track down a lot of prior discussion (like I just did)

Todo:

- [x] Confirm access to the service app that can create share keys.
    - [x] Alternatively create a new one we control and tell the itwin team the new `clientId` that should be able to make our contracts
    - [x] Find a place to store access, possibly Bitwarden, possibly in github secrets if we go the CI route
- [x] Create a script to get new share keys and write up process for using said script
    - It's unlikely this needs to be part of the main release process
    - it _is_ possible to create [scheduled Github Workflows](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule). Maybe there's an easy way for it to generate new keys and ~~[create a PR](https://github.com/marketplace/actions/create-pull-request)~~ [create a PR](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli) at the end of every month with new keys
        - (if it is it could be nice to automate the other key generation we do for releases)
    - Either way we'll need a way to get the client secret to the script so this Pr could be helpful: https://github.com/CesiumGS/cesium/pull/12971

CC @ggetz

jjspace:
# Description

This PR creates a new script to update our ion token and itwin share keys in CI. This is meant to run the last week of the month to help prepare for the release.

The primary motivator for this change was difficulty in sharing the secrets needed for the itwin key update but it was easy enough to expand the system to the ion key at the same time. The ArcGIS token will still need to be manually updated as part of the release process since I didn't look too deeply into their API.

- Created a new "replacer" script system to specify a [js selector](https://estools.github.io/esquery/) and a new value to set the variable to
- Created helper scripts to create/load keys/tokens from ion and the itwin API.
    - The functions attempt to re-use keys if one was already created for a given release number. This is mainly important for the itwin keys since there is a hard limit of 10 keys and we don't want to create too many or delete ones still in use.
        - This also helps in case we need to re-run the CI job
    - [x] itwin keys will expire automatically but we need to decide how to handle the ion tokens. I did create the function to automatically delete the oldest token but I'm not sure the best place/time to run it or if it should still just be manual
- The CI job will create a PR using the template in the action directory and label it `priority - next release` so it should just need review. [Example here](https://github.com/CesiumGS/cesium/pull/13270)

## Issue number and link

Fixes #13225 

## Testing plan

The replacer script can be tested locally even wihout being able to generate new keys
```
cd .github/actions/update-tokens
npm install
node index.js
```
Check `Ion.js` and the two itwin sandcastles for updated fake values. Real values won't be requested unless run using `node index.js --update` but this won't work without env variables for the ion access token and itwin client secrets.

# Author checklist


.github/actions/update-tokens/replacements.js
```diff
@@ -0,0 +1,107 @@
...
+      selector: propertySelector("Cesium.ITwinPlatform.defaultShareKey"),
+      // automatible with the Share key API https://developer.bentley.com/apis/access-control-v2/operations/get-itwin-share/
+      // Need to get an auth token using the client id + secret first
+      newValue: async () => {
+        try {
+          const newToken = await getNewKeyForItwin(featureSeriviceItwin);
+          return newToken;
+        } catch (error) {
+          console.error(error);
+          // Just skip updating if there was an error
```
mzschwartz5, line 89: Where does this error log to? Should the PR template maybe include a step to check these logs in case of failure?

.github/actions/update-tokens/ionTokenUpdater.js
```diff
@@ -0,0 +1,156 @@
...
+  return knownTokens;
+}
+
+/**
+ * @param {number} olderThanDays
+ */
+// eslint-disable-next-line
+async function deleteOldTokens(olderThanDays) {
+  // TODO: figure out how we want to manage this
+  throw new Error("don't call");
```
mzschwartz5: What's the hold up? Just extra precaution since deleting is a dangerous thing?

Should we track TODO in an issue?

.github/actions/update-tokens/replacements.js
jjspace: I was thinking that it would be obvious if the PR itself shows a bad key or is missing the update to a file that is expected. Not opposed to adding a comment in the template if you think it would hlep

mzschwartz5: Do the failure and logs show up as a failing workflow step on the PR itself? If so, I think we're fine as-is / it'll be clear where to look.

.github/actions/update-tokens/ionTokenUpdater.js
jjspace: I didn't want to include the delete as part of the PR creation step to try and help ensure we always had 2 versions "live" with valid tokens. I just moved this function to it's own script and set it up to run as part of the `prod` workflow which is the last step of the release process. Hopefully this helps ensure we don't delete it too early while also removing the manual task.

.github/actions/update-tokens/replacements.js
jjspace: The logs are in CI but the job itself didn't fail. My thinking was to prevent blocking the other key updates from happening if a single one failed. I thought it might be enough to "just know" how many updates there should be in the PR and see that one or more are missing but that's a really fragile workflow.

I've updated this to actually replace the value with a "Failed to get value" message so it's clear it was _supposed_ to update but failed for some reason. This is separate from intentionally _skipping_ an update for whatever reason.

PR #13280, opened 2026-03-09
Fix min and max values in decodeI3S.js

yuz1:
# Description

This issue occurs during the conversion from **I3S to glTF**, where the **min/max values are assigned incorrectly**. This may affect features such as model rendering or collision detection. Although no issues were observed within this repository, when I extracted the glTF files from the browser cache and used them in other software, **raycasting detection failed**.

## Testing plan

# Author checklist


jjspace:
Thanks for the PR @yuz1. The change looks like a straightforward typo fix but can you please fill out a bit of info in the PR description and a testing plan to verify? Some sample data to test with would be helpful too if you have some

mzschwartz5:
I'm going to close this PR due to inactivity, but @yuz1 please feel free to reopen if you plan to address @jjspace's comments.

javagl:
I don't think that it makes much sense to offload the work to create test data to someone who proposes a small, obvious fix. If tests and test data are required, then they should have been added together with that `decodeI3S.js`. 

Nobody knows what all that code is doing, and it's not documented at all, and (obviously) not tested sufficiently. The case that was described in the first comment here (of "extracting" that data) is questionable on many levels, and certainly not a prime use case. For CesiumJS, the wrong `min/max` probably don't matter (_now_ - but maybe in the future, and then: good luck debugging where some "negative bounding volume size" is coming from). 

But the values are *clearly* wrong. In doubt, it can be seen in one of the I3S sandcastles, when setting a breakpoint at the right place:

Cesium I3S Fix

I'll reopen this for now.

And I suggest that we merge it. Is anybody opposed?

mzschwartz5:
I'm given a little pause by this line in the description:

> Although no issues were observed within this repository...

But I'm also generally in favor of merging and, if there's an issue later, reverting. (At least, for something like this that seems pretty low risk)

javagl:
> Although no issues were observed within this repository...

Again, the code is a handful, but that's what it looks like: This I3S conversion function seems to create a glTF asset at runtime, so that it can be forwarded to CesiumJS for loading. This glTF asset contains the `accessor.min/max` values for the POSITION (they are required by the specification). They had been wrong, but that was not noticed yet, because CesiumJS is currently not using these `min/max` values. If CesiumJS ever decides to use these `min/max` values (say, for computing some bounding box from them), then they have to be correct.

javagl:
Restarted the tests 3 times. We have to comment out that "tileset statistics" test. Maybe someone else would like to trigger it a few more times until it passes.

jjspace:
> Nobody knows what all that code is doing, and it's not documented at all, and (obviously) not tested sufficiently.

This is precisely why I had originally requested a test case. As stated before I think it's an obvious typo but I also tend to err on the side of caution in case the "bug" is actually helping create intended behavior somewhere else in the pipeline. https://xkcd.com/1172/

If you both also feel comfortable as well then don't let my review hold this up

javagl:
It's hard to imagine what such a test case could look like. From a quick look, the `I3SDecoderSpec` is _literally_ only testing "corner cases", i.e. _only_ cases with empty inputs. Maybe some actual decoding is tested as parts of other specs, but I haven't seen an _obvious_ place where some test for the `min/max` values could be added. There seem to be places in the `I3SNodeSpec` that access `.accessors`, but only ones that have been set up with some "mocking", apparently.

(In general, I'm all in favor of _making things better_, and that includes increasing the test coverage or adding documentation - but this is fighting against windmills. If this is causing a regression for someone, saying "No, the maximum must be smaller than the minimum!", we can see how to resolve it)

PR #13282, opened 2026-03-09
CartesianN: Overload .clone() signature for undefined return type

### LINKED ISSUE #13265: Cartesian clone type definition is wrong
### What happened?

After #13168, the generated TypeScript definitions for the static clone methods are incorrect. Methods are extended to accept undefined and return undefined in that case, but TypeScript will throw an error on this with `skipLibCheck: false`. 
```
TS2322: Type 'undefined' is not assignable to type 'Cartesian3'.
    102 |   static clone(cartesian, result) {
    103 |     if (!defined(cartesian)) {
  > 104 |       return undefined;
        |       ^^^^^^
    105 |     }
    106 |     if (!defined(result)) {
    107 |       return new Cartesian3(cartesian.x, cartesian.y, cartesian.z);
```
https://github.com/CesiumGS/cesium/blob/2428f04ac54f896b975a4fe1b90319feb62e080a/packages/engine/Source/Core/Cartesian2.js#L58-L76

### Sandcastle example

_No response_

### Environment

CesiumJS Version: 1.139.0

donmccurdy:
# Description

Fixes #13265.

Overload `CartesianN.clone` signatures, so TS knows that if the input could be undefined, the return value could be undefined. Our documentation system includes only the last signature, called the "implementation signature," which must be compatible with all overload signatures.

With this change, _and_ after enabling strict null checks (left for another PR) our own type checker can detect the errors that lead to #13265. Type checking must be enabled on the file for this to work. 

Unfortunately, the type definitions built with tsd-jsdoc do not include the overload signatures, so downstream projects will likely see this change as a breaking change in types. While the types are correct (the function has always been able to return undefined!) I can imagine it might be frustrating to need to mitigate this on each call to .clone():

```typescript
const a = Cartesian3.clone(value) as Cartesian3;
const a = Cartesian3.clone(value)!;
```

We would likely need to use TypeScript to build our type definitions to retain the overload signatures, which is a lorger change underway at #4434.

In the meantime I'm a bit worried this may be a "medicine worse than the disease" situation, but opening this PR for discussion.

## Issue number and link

- #13265

## Testing plan

Compare documentation and Cesium.d.ts before and after. Both should reflect that the input and output may be undefined. Neither will reflect the overload signature, i.e. the fact that providing non-nullable input guarantees the output will not be null.

# Author checklist

#### PR Dependency Tree

* **PR #13282** 👈
  * **PR #13290**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
 
 - Fixed memory leak when rendering Gaussian splat 3D tilesets. [#13229](https://github.com/CesiumGS/cesium/pull/13229/)
 - No longer disables custom shaders for primitives with missing metadata, as long as the metadata exists on the overall class definition. [#13258](https://github.com/CesiumGS/cesium/pull/13258)
+- Types for Cartesian2, Cartesian3, and Cartesian4 `.clone()` now reflect that they may return undefined. [#13265](https://github.com/CesiumGS/cesium/issues/13265)
 
 ## 1.139.1 - 2026-03-05
 


donmccurdy:
Ready for review. Changes in generated `Cesium.d.ts` are minimal as expected:

Screenshot 2026-03-11 at 11 24 06 AM

javagl:
Sparked by https://github.com/CesiumGS/cesium/pull/13168#issuecomment-4054866265 , I'd like to emphasize what I already mentioned in the [comment in the issue](https://github.com/CesiumGS/cesium/issues/13265#issuecomment-4007759132) : This will be a breaking change, but a "good" one, insofar that it brings us closer to the truth.

A bit more generally:

It would be nice if there was a way to "capture" the _actual_ type. This is possible with Java, as in
```java
static <T> List<T> doSomething(List<T> list) { ... }
//      ^ This type refers to the whole signature - when T goes in, then T comes out...
```
And it is also possible in TypeScript, e.g. as in the [Typescriptified version of `defined`](https://github.com/CesiumGS/3d-tiles-tools/blob/645723b0900d7f8c8e46b9cdff3b031405fc3063/src/base/base/defined.ts#L31) that is used in the 3D Tiles Tools (skipping some general musings about 'defined' here). 

But ... right now, I'm not even up to date about which part of the toolchain is actually reading and processing these types, and generating the `d.ts` eventually. I assume that it's **not** possible to squeeze this sort of type parameters into the documentation and force it to generate nicely and properly typed TS output. You mentioned "function overloading" in https://github.com/CesiumGS/cesium/issues/13265#issuecomment-4005345276 , maybe that's a way...? 

EDIT: Quick searches reveal a "template" tag: https://docs.joshuatz.com/cheatsheets/js/jsdoc/#generics-in-jsdoc - shouldn't that be close to what we need....? (Will have another look...)

donmccurdy:
I've tried three solutions so far:

- (A) Function overloading / `@overload` (this PR)
- (B) Generics / `@template`
- (C) `@returns {cartesian extends Cartesian3 ? Cartesian3 : undefined}`

All three work in TS-flavored JSDoc (passing `npm run tsc`), but (B) and (C) cause `tsd-jsdoc` to crash when converting JSDoc into .d.ts declarations. With (A), `tsd-jsdoc` merely drops the overloads in published .d.ts declarations, emitting types that are correct but less "narrow" than they should be. This a breaking change in types — as you mention — users with TS "strict mode" enabled will need to cast the result with `as Cartesian3` or similar.

We certainly need to replace `tsd-jsdoc` eventually, probably with tsc. But [previous discussions](https://github.com/CesiumGS/cesium/issues/10455#issuecomment-2573945180) had raised ES6 classes as a prerequisite for switching to tsc... so our migration has a cyclical dependency as the ES6 class conversion allows users' projects to begin detecting pre-existing errors in our types.

Finally, option (D) might be to keep the types as-is (pretending `.clone()` cannot return `undefined`), so https://github.com/CesiumGS/cesium/issues/13265 remains open during the migration, and we'll (internally) use `@ts-expect-error` to silence cases where existing types are incorrect while we roll out type checking.

No perfect options available, it seems. My preference might be:

1. Try to fast-track the ES6 class migration. Improving types along the way (A) is a nice-to-have, but keeping types as-is (D) and temporarily accepting issues like #13265 might be practical to get through the ~1300 source files in `packages/engine` within a year or so.
2. Get `packages/engine` passing tsc type checks
3. Switch from tsd-jsdoc to tsc to build published type definitions for `packages/engine`

If (2) could somehow happen before (3), or happen in parallel, that'd be ideal. Needing to satisfy tsd-jsdoc does restrict the syntax we can use to satisfy tsc. But that is manageable if not.

I'm starting to lean toward (D), I think. Don't force users to narrow types with `Cartesian3.clone(...) as Cartesian3` during the migration, if we can start publishing correct (overloaded) types in a year or so, and save them that churn.

donmccurdy:
Aside - it could be a big time-saver if we could see a diff of Cesium.d.ts in PR review during the migration, I wonder if that's something we could add to CI...

javagl:
I also played a bit with `@template`, but that didn't lead anywhere (i.e. resulted in some broken `.d.ts` output, but I didn't spend toooo much time investigating where that came from). 

I don't have a strong opinion on which approach to take, but see _some_ of these points in islolation: 

- `clone` can return `undefined`, so the current signature is wrong, and (c/sh?)ould be fixed (even though it's breaking). 
- It would be nicer and cleaner if we could throw in some `<T>` there to capture the parameter _and_ return type, but without being closer to the "fully typescripted" state, this doesn't seem to be possible (and there is no timeline for that). 
- We should do the conversion to `class` as soon as possible. Maybe we could change the title of https://github.com/CesiumGS/cesium/issues/4434 to "Let us <strike>slowly</strike> quickly move to TypeScript"

_Some_ typing updates and inconsistencies (like for `clone` or `pack` right now) might be unavoidable during that. Maybe it can be decided on a case-by-case basis. (For example: The `pack` type change was a fairly [trivial fix](https://github.com/CesiumGS/3d-tiles-tools/commit/27e6a403264b3ca58f2864186ee78693c8fca679), and I could imagine that this and similar changes are not overly disruptive and therefore acceptable in many cases).

donmccurdy:
Because the return type of .clone() has been wrong forever, and "fixing" it now would cause far more churn for users than waiting to fix it until after we resolve #10455, I think maybe we should wait. So I'll close this issue. There might be other fixes for #13265 available, depending on answers to some of the questions raised in that thread.

Proposed fix for pack/unpack:

- https://github.com/CesiumGS/cesium/pull/13302

PR #13286, opened 2026-03-09
Make dev landing page prettier

jjspace:
# Description

Our landing page for the local dev server has always been extremely simple. Meanwhile the landing page included in our released zip files and in the version archive on cesium.com looks much nicer. This PR brings some of that style back to the dev page so we don't have to only look at a "programmer art" version.

image

While I was there, and as requested by @mzschwartz5, I added colors for the page for darkmode. I didn't add a selector for it to avoid saving the value in localhost so it relies on the system/browser to specify the preference. (It probably [wouldn't be hard](https://alexandersandberg.com/articles/creating-a-website-theme-switcher-with-css-only/) to do though if we want)

As a side effect of needing a nicer logo for darkmode I added a higher resolution logo image which is also used on Sandcastle. That logo should now look more clean too!

## Testing plan

- Check the main `index.html` page locally or in CI
- Force the `prefers-color-scheme` to dark if it's not already using devtools or your browser settings

# Author checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,12 +11,20 @@
 #### Additions :tada:
 
 - Added experimental, performance-focused vector primitive APIs: `BufferPointCollection`, `BufferPolylineCollection`, and `BufferPolygonCollection`. [#13212](https://github.com/CesiumGS/cesium/pull/13212)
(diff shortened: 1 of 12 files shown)


mzschwartz5:
Aw yeah this is sweet. My eyes! 💯

Issue #13283, opened 2026-03-09
Crash for many splats with high spherical harmonics degree


javagl:
Test data (including a sandcastle): 

[TestForManySplats-2026-03-09.zip](https://github.com/user-attachments/files/25848152/TestForManySplats-2026-03-09.zip)

The archive contains two tilesets, each with a single GLB file with splats. Both splat files contain 4.9 million splats. 

The first one uses spherical harmonics degree 0. This can be loaded in CesiumJS.

The second one uses spherical harmonics degree 3. Loading this causes CesiumJS to crash:
```
Aborted()
A 3D tile failed to load: http://localhost:8003/manySplatsDeg3.glb
Error: Failed to load glTF
Failed to load vertex buffer
Failed to load SPZ
Aborted(). Build with -sASSERTIONS for more info.
Original stack:
Original stack:
Original stack:
RuntimeError: Aborted(). Build with -sASSERTIONS for more info.
    at jA (https://sandcastle.cesium.com/js/engine/index.js:81744:101)
    at x (https://sandcastle.cesium.com/js/engine/index.js:82201:34)
    at wasm://wasm/000a0302:wasm-function[458]:0x14597
    at wasm://wasm/000a0302:wasm-function[61]:0x3099
    at wasm://wasm/000a0302:wasm-function[235]:0x8544
    at wasm://wasm/000a0302:wasm-function[83]:0x416e
    at wasm://wasm/000a0302:wasm-function[307]:0xa38a
    at wasm://wasm/000a0302:wasm-function[430]:0x14012
    at wasm://wasm/000a0302:wasm-function[129]:0x4d89
    at wasm://wasm/000a0302:wasm-function[614]:0x1a2cd
Handler stack:
Error
    at new RuntimeError (https://sandcastle.cesium.com/js/engine/index.js:10424:11)
    at GltfSpzLoader.getError (https://sandcastle.cesium.com/js/engine/index.js:74245:26)
    at handleError6 (https://sandcastle.cesium.com/js/engine/index.js:82702:19)
    at GltfSpzLoader.process (https://sandcastle.cesium.com/js/engine/index.js:82643:7)
    at GltfVertexBufferLoader.process (https://sandcastle.cesium.com/js/engine/index.js:81407:39)
    at processLoaders (https://sandcastle.cesium.com/js/engine/index.js:94529:46)
    at GltfLoader._process (https://sandcastle.cesium.com/js/engine/index.js:94358:7)
[...]

PR #13287, opened 2026-03-09
Fix: unrecoverable WASM crash when loading large SPZ Gaussian splat datasets

danielzhong:
# Description
Loading SPZ Gaussian splat files with a high spherical harmonics degree (e.g. degree 3) and a large number of splats could cause the spz-loader WASM module to call `Aborted()`, producing an unrecoverable crash and an unhelpful "Aborted(). Build with -sASSERTIONS for more info." error message. This occurred because the WASM module is compiled with a signed 32-bit address space, imposing a hard 2 GB memory ceiling. 

For example, a dataset of ~4.9 million splats at SH degree 3 requires approximately 2.2 GB to decode, exceeding that limit.

Changes:

* Added a pre flight memory estimate in GltfSpzLoader.process() that runs before the WASM decoder is invoked.
* Point count is read from the `glTF POSITION` accessor's count field; SH degree is inferred from the highest  numbered `SH_DEGREE_n` attribute in the primitive. The SPZ binary itself cannot be inspected directly because it is gzip compressed.
* If the estimated decoded size exceeds 1.6 GB (a conservative threshold below the 2 GB ceiling), the loader fails gracefully with a descriptive RuntimeError rather than crashing the WASM module.

Before:
```
Aborted()
```
After:
```
Error: Failed to load glTF
Failed to load vertex buffer
Failed to load SPZ
SPZ data too large to decode: 4,913,000 splats with spherical harmonics degree 3 would require approximately 2212 MB, which exceeds the WASM memory limit. Consider using a lower spherical harmonics degree or splitting the dataset into smaller tiles.
```

## Issue number and link
https://github.com/CesiumGS/cesium/issues/13283

## Testing plan
Test data:
https://github.com/CesiumGS/cesium/issues/13283

# Author checklist


javagl:
It's good to see the reason for the issue being identified. 

I don't fully understand the 2.2 GB. With position, scale, opacity, rotation, and the 15*3 SH coefficients, that would be
`4900000 * (3 + 3 + 1 + 4 + 45) * 4 ~= 1 GB`
but maybe I'm missing some place where all this is `double` (last factor being 8 instead of 4).

However, when you say

> This occurred because the WASM module is compiled with a signed 32-bit address space

I wonder: Is it possible to ~"compile this with 64 bit address space"? Maybe something to raise as an issue in https://github.com/drumath2237/spz-loader ? 

The point is: Apparently, tools (including ion!) have already created data sets that cause this error. And I wonder whether it's possible to handle this on the client side. Otherwise, people would 1. have to wait until this is fixed, so that ion does no longer create data sets that cause this error, and 2. then have to re-tile their data...

danielzhong:
> It's good to see the reason for the issue being identified.
> 
> I don't fully understand the 2.2 GB. With position, scale, opacity, rotation, and the 15*3 SH coefficients, that would be `4900000 * (3 + 3 + 1 + 4 + 45) * 4 ~= 1 GB` but maybe I'm missing some place where all this is `double` (last factor being 8 instead of 4).
> 
> However, when you say
> 
> > This occurred because the WASM module is compiled with a signed 32-bit address space
> 
> I wonder: Is it possible to ~"compile this with 64 bit address space"? Maybe something to raise as an issue in https://github.com/drumath2237/spz-loader ?
> 
> The point is: Apparently, tools (including ion!) have already created data sets that cause this error. And I wonder whether it's possible to handle this on the client side. Otherwise, people would 1. have to wait until this is fixed, so that ion does no longer create data sets that cause this error, and 2. then have to re-tile their data...

Math:
Yes, the decoded output alone is roughly 4,900,000 × (3+3+1+4+45) × 4 ≈ 1.1 GB. The ×2 factor in the estimate is conservative during gzip decompression, the compressed stream and the decoded output coexist briefly on the WASM heap alongside zlib's working buffers, which can push peak usage well above the final output size. 

On Memory64:
I have opened an issue on spz-loader to raise this.
https://github.com/drumath2237/spz-loader/issues/75

lukemckinstry:
When running with the provided sample data in sandcastle, I see same error in both `main` and this branch:

```
Error: Failed to load glTF
Failed to load vertex buffer
Failed to load SPZ
Aborted(). Build with -sASSERTIONS for more info.
```
I do not see the descriptive message from the example in the PR description:
```
SPZ data too large to decode: 4,913,000 splats with spherical harmonics degree 3 would require approximately 2212 MB, which exceeds the WASM memory limit. Consider using a lower spherical harmonics degree or splitting the dataset into smaller tiles.
```
And since the `Aborted()` message still appears, is this the graceful failure and early exit that was intended?

lukemckinstry:
I see the enhanced error message as described.  (disregard my previous resolved comment as I was testing the wrong branch)

My only questions:
 - Are we handling incomplete/malformed glTF JSON?
 - Is it appropriate to add unit tests?

danielzhong:
> * Are we handling incomplete/malformed glTF JSON?

The preflight is best effort. If the glTF JSON does not contain the metadata needed for the estimate, we skip the estimate rather than failing on the JSON itself. So this change should not introduce a new failure mode for incomplete/malformed glTF JSON.

That said, if the required metadata is missing, we also cannot preflight the SPZ size, so in that case the loader falls back to the existing decode path.

I added a lightweight unit test around the preflight estimate / early-failure path. I think that is the right level of coverage here, since the goal of this change is to verify our guard logic before invoking the WASM decoder, rather than exercising the full GLB loading path or the downstream SPZ library itself.

PR #13290, opened 2026-03-11
Types: Convert MatrixN to ES6 classes

donmccurdy:
# Description

Converts Matrix2, Matrix3, and Matrix4 to ES6 classes, and enables type checking on these files. The difficulty in this conversion was that all three classes are tagged `@implements {ArrayLike<number>}`, but JSDoc doesn't support declaring the index signatures needed to explain that implementation to TypeScript:

- https://github.com/microsoft/TypeScript/issues/48096
- https://stackoverflow.com/a/71928083

The two workarounds I'm aware of are:

- (1) Extend (not just implement) the "Array" class
- (2) Add some (10-ish) `@ts-expect-error` annotations in each class

~~I've implemented (1) in this PR, similar to the approach in [`gl-matrix`](https://github.com/toji/gl-matrix) v4, but I'm not feeling strongly about (1) vs. (2) yet.~~ Switched to (2). This PR also depends on the overloading approach from #13282. 

Includes a modest changes in documentation to make the type checker happy, after vs. before:

eigen_decomp

## Issue number and link

- #8359

## Testing plan

- `npm run build-docs`, compare documentation
- `npm run build-ts`, compare `Cesium.d.ts`
- Unit tests

With https://github.com/CesiumGS/cesium/pull/13259, we no longer need to search for `new Matrix3.clone(...)` errors.

Diff of Cesium.d.ts:

<details>
<summary>Cesium.d.ts.diff</summary>

```diff
diff --git a/Source/Cesium.main.d.ts b/Source/Cesium.d.ts
index c6225ddedb..7a126c0ac3 100644
--- a/Source/Cesium.main.d.ts
+++ b/Source/Cesium.d.ts
@@ -10101,10 +10101,6 @@ export interface Matrix2 extends ArrayLike<number> {
  */
 export class Matrix2 implements ArrayLike<number> {
     constructor(column0Row0?: number, column1Row0?: number, column0Row1?: number, column1Row1?: number);
-    /**
-     * The number of elements used to pack the object into an array.
-     */
-    static packedLength: number;
     /**
      * Stores the provided instance into the provided array.
      * @param value - The value to pack.
@@ -10143,25 +10139,6 @@ export class Matrix2 implements ArrayLike<number> {
      * @returns The modified result parameter or a new Matrix2 instance if one was not provided. (Returns undefined if matrix is undefined)
      */
     static clone(matrix: Matrix2, result?: Matrix2): Matrix2;
-    /**
-     * Creates a Matrix2 from 4 consecutive elements in an array.
-     * @example
-     * // Create the Matrix2:
-     * // [1.0, 2.0]
-     * // [1.0, 2.0]
-     *
-     * const v = [1.0, 1.0, 2.0, 2.0];
-     * const m = Cesium.Matrix2.fromArray(v);
-     *
-     * // Create same Matrix2 with using an offset into an array
-     * const v2 = [0.0, 0.0, 1.0, 1.0, 2.0, 2.0];
-     * const m2 = Cesium.Matrix2.fromArray(v2, 2);
-     * @param array - The array whose 4 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
-     * @param [startingIndex = 0] - The offset into the array of the first element, which corresponds to first column first row position in the matrix.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Matrix2 instance if one was not provided.
-     */
-    static fromArray(array: number[], startingIndex?: number, result?: Matrix2): Matrix2;
     /**
      * Creates a Matrix2 instance from a column-major order array.
      * @param values - The column-major order array.
@@ -10249,7 +10226,7 @@ export class Matrix2 implements ArrayLike<number> {
      * @param result - The object onto which to store the result.
      * @returns The modified result parameter.
      */
-    static setColumn(matrix: Matrix2, index: number, cartesian: Cartesian2, result: Cartesian2): Matrix2;
+    static setColumn(matrix: Matrix2, index: number, cartesian: Cartesian2, result: Matrix2): Matrix2;
     /**
      * Retrieves a copy of the matrix row at the provided index as a Cartesian2 instance.
      * @param matrix - The matrix to use.
@@ -10415,6 +10392,57 @@ export class Matrix2 implements ArrayLike<number> {
      * @returns <code>true</code> if left and right are within the provided epsilon, <code>false</code> otherwise.
      */
     static equalsEpsilon(left?: Matrix2, right?: Matrix2, epsilon?: number): boolean;
+    /**
+     * Duplicates the provided Matrix2 instance.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Matrix2 instance if one was not provided.
+     */
+    clone(result?: Matrix2): Matrix2;
+    /**
+     * Compares this matrix to the provided matrix componentwise and returns
+     * <code>true</code> if they are equal, <code>false</code> otherwise.
+     * @param [right] - The right hand side matrix.
+     * @returns <code>true</code> if they are equal, <code>false</code> otherwise.
+     */
+    equals(right?: Matrix2): boolean;
+    /**
+     * Compares this matrix to the provided matrix componentwise and returns
+     * <code>true</code> if they are within the provided epsilon,
+     * <code>false</code> otherwise.
+     * @param [right] - The right hand side matrix.
+     * @param [epsilon = 0] - The epsilon to use for equality testing.
+     * @returns <code>true</code> if they are within the provided epsilon, <code>false</code> otherwise.
+     */
+    equalsEpsilon(right?: Matrix2, epsilon?: number): boolean;
+    /**
+     * Creates a string representing this Matrix with each row being
+     * on a separate line and in the format '(column0, column1)'.
+     * @returns A string representing the provided Matrix with each row being on a separate line and in the format '(column0, column1)'.
+     */
+    toString(): string;
+    /**
+     * The number of elements used to pack the object into an array.
+     */
+    static packedLength: number;
+    /**
+     * Creates a Matrix2 from 4 consecutive elements in an array.
+     * @example
+     * // Create the Matrix2:
+     * // [1.0, 2.0]
+     * // [1.0, 2.0]
+     *
+     * const v = [1.0, 1.0, 2.0, 2.0];
+     * const m = Cesium.Matrix2.fromArray(v);
+     *
+     * // Create same Matrix2 with using an offset into an array
+     * const v2 = [0.0, 0.0, 1.0, 1.0, 2.0, 2.0];
+     * const m2 = Cesium.Matrix2.fromArray(v2, 2);
+     * @param array - The array whose 4 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
+     * @param [startingIndex = 0] - The offset into the array of the first element, which corresponds to first column first row position in the matrix.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Matrix2 instance if one was not provided.
+     */
+    static fromArray(array: number[], startingIndex?: number, result?: Matrix2): Matrix2;
     /**
      * An immutable Matrix2 instance initialized to the identity matrix.
      */
@@ -10451,40 +10479,13 @@ export class Matrix2 implements ArrayLike<number> {
      * matrix[Cesium.Matrix2.COLUMN1ROW1] = 5.0; // set column 1, row 1 to 5.0
      */
     static readonly COLUMN1ROW1: number;
-    /**
-     * Gets the number of items in the collection.
-     */
-    length: number;
-    /**
-     * Duplicates the provided Matrix2 instance.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Matrix2 instance if one was not provided.
-     */
-    clone(result?: Matrix2): Matrix2;
-    /**
-     * Compares this matrix to the provided matrix componentwise and returns
-     * <code>true</code> if they are equal, <code>false</code> otherwise.
-     * @param [right] - The right hand side matrix.
-     * @returns <code>true</code> if they are equal, <code>false</code> otherwise.
-     */
-    equals(right?: Matrix2): boolean;
-    /**
-     * Compares this matrix to the provided matrix componentwise and returns
-     * <code>true</code> if they are within the provided epsilon,
-     * <code>false</code> otherwise.
-     * @param [right] - The right hand side matrix.
-     * @param [epsilon = 0] - The epsilon to use for equality testing.
-     * @returns <code>true</code> if they are within the provided epsilon, <code>false</code> otherwise.
-     */
-    equalsEpsilon(right?: Matrix2, epsilon?: number): boolean;
-    /**
-     * Creates a string representing this Matrix with each row being
-     * on a separate line and in the format '(column0, column1)'.
-     * @returns A string representing the provided Matrix with each row being on a separate line and in the format '(column0, column1)'.
-     */
-    toString(): string;
 }
 
+export type EigenDecompositionResult = {
+    unitary?: Matrix3;
+    diagonal?: Matrix3;
+};
+
 export interface Matrix3 extends ArrayLike<number> {
 }
 
@@ -10503,10 +10504,6 @@ export interface Matrix3 extends ArrayLike<number> {
  */
 export class Matrix3 implements ArrayLike<number> {
     constructor(column0Row0?: number, column1Row0?: number, column2Row0?: number, column0Row1?: number, column1Row1?: number, column2Row1?: number, column0Row2?: number, column1Row2?: number, column2Row2?: number);
-    /**
-     * The number of elements used to pack the object into an array.
-     */
-    static packedLength: number;
     /**
      * Stores the provided instance into the provided array.
      * @param value - The value to pack.
@@ -10545,26 +10542,6 @@ export class Matrix3 implements ArrayLike<number> {
      * @returns The modified result parameter or a new Matrix3 instance if one was not provided. (Returns undefined if matrix is undefined)
      */
     static clone(matrix: Matrix3, result?: Matrix3): Matrix3;
-    /**
-     * Creates a Matrix3 from 9 consecutive elements in an array.
-     * @example
-     * // Create the Matrix3:
-     * // [1.0, 2.0, 3.0]
-     * // [1.0, 2.0, 3.0]
-     * // [1.0, 2.0, 3.0]
-     *
-     * const v = [1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
-     * const m = Cesium.Matrix3.fromArray(v);
-     *
-     * // Create same Matrix3 with using an offset into an array
-     * const v2 = [0.0, 0.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
-     * const m2 = Cesium.Matrix3.fromArray(v2, 2);
-     * @param array - The array whose 9 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
-     * @param [startingIndex = 0] - The offset into the array of the first element, which corresponds to first column first row position in the matrix.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Matrix3 instance if one was not provided.
-     */
-    static fromArray(array: number[], startingIndex?: number, result?: Matrix3): Matrix3;
     /**
      * Creates a Matrix3 instance from a column-major order array.
      * @param values - The column-major order array.
@@ -10875,7 +10852,7 @@ export class Matrix3 implements ArrayLike<number> {
      * @param [result] - An object with unitary and diagonal properties which are matrices onto which to store the result.
      * @returns An object with unitary and diagonal properties which are the unitary and diagonal matrices, respectively.
      */
-    static computeEigenDecomposition(matrix: Matrix3, result?: any): any;
+    static computeEigenDecomposition(matrix: Matrix3, result?: EigenDecompositionResult): EigenDecompositionResult;
     /**
      * Computes a matrix, which contains the absolute (unsigned) values of the provided matrix's elements.
      * @param matrix - The matrix with signed elements.
@@ -10921,6 +10898,58 @@ export class Matrix3 implements ArrayLike<number> {
      * @returns <code>true</code> if left and right are within the provided epsilon, <code>false</code> otherwise.
      */
     static equalsEpsilon(left?: Matrix3, right?: Matrix3, epsilon?: number): boolean;
+    /**
+     * Duplicates the provided Matrix3 instance.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Matrix3 instance if one was not provided.
+     */
+    clone(result?: Matrix3): Matrix3;
+    /**
+     * Compares this matrix to the provided matrix componentwise and returns
+     * <code>true</code> if they are equal, <code>false</code> otherwise.
+     * @param [right] - The right hand side matrix.
+     * @returns <code>true</code> if they are equal, <code>false</code> otherwise.
+     */
+    equals(right?: Matrix3): boolean;
+    /**
+     * Compares this matrix to the provided matrix componentwise and returns
+     * <code>true</code> if they are within the provided epsilon,
+     * <code>false</code> otherwise.
+     * @param [right] - The right hand side matrix.
+     * @param [epsilon = 0] - The epsilon to use for equality testing.
+     * @returns <code>true</code> if they are within the provided epsilon, <code>false</code> otherwise.
+     */
+    equalsEpsilon(right?: Matrix3, epsilon?: number): boolean;
+    /**
+     * Creates a string representing this Matrix with each row being
+     * on a separate line and in the format '(column0, column1, column2)'.
+     * @returns A string representing the provided Matrix with each row being on a separate line and in the format '(column0, column1, column2)'.
+     */
+    toString(): string;
+    /**
+     * The number of elements used to pack the object into an array.
+     */
+    static packedLength: number;
+    /**
+     * Creates a Matrix3 from 9 consecutive elements in an array.
+     * @example
+     * // Create the Matrix3:
+     * // [1.0, 2.0, 3.0]
+     * // [1.0, 2.0, 3.0]
+     * // [1.0, 2.0, 3.0]
+     *
+     * const v = [1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
+     * const m = Cesium.Matrix3.fromArray(v);
+     *
+     * // Create same Matrix3 with using an offset into an array
+     * const v2 = [0.0, 0.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
+     * const m2 = Cesium.Matrix3.fromArray(v2, 2);
+     * @param array - The array whose 9 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
+     * @param [startingIndex = 0] - The offset into the array of the first element, which corresponds to first column first row position in the matrix.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Matrix3 instance if one was not provided.
+     */
+    static fromArray(array: number[], startingIndex?: number, result?: Matrix3): Matrix3;
     /**
      * An immutable Matrix3 instance initialized to the identity matrix.
      */
@@ -10965,40 +10994,15 @@ export class Matrix3 implements ArrayLike<number> {
      * The index into Matrix3 for column 2, row 2.
      */
     static readonly COLUMN2ROW2: number;
-    /**
-     * Gets the number of items in the collection.
-     */
-    length: number;
-    /**
-     * Duplicates the provided Matrix3 instance.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Matrix3 instance if one was not provided.
-     */
-    clone(result?: Matrix3): Matrix3;
-    /**
-     * Compares this matrix to the provided matrix componentwise and returns
-     * <code>true</code> if they are equal, <code>false</code> otherwise.
-     * @param [right] - The right hand side matrix.
-     * @returns <code>true</code> if they are equal, <code>false</code> otherwise.
-     */
-    equals(right?: Matrix3): boolean;
-    /**
-     * Compares this matrix to the provided matrix componentwise and returns
-     * <code>true</code> if they are within the provided epsilon,
-     * <code>false</code> otherwise.
-     * @param [right] - The right hand side matrix.
-     * @param [epsilon = 0] - The epsilon to use for equality testing.
-     * @returns <code>true</code> if they are within the provided epsilon, <code>false</code> otherwise.
-     */
-    equalsEpsilon(right?: Matrix3, epsilon?: number): boolean;
-    /**
-     * Creates a string representing this Matrix with each row being
-     * on a separate line and in the format '(column0, column1, column2)'.
-     * @returns A string representing the provided Matrix with each row being on a separate line and in the format '(column0, column1, column2)'.
-     */
-    toString(): string;
 }
 
+export type Viewport = {
+    x?: number;
+    y?: number;
+    width?: number;
+    height?: number;
+};
+
 export interface Matrix4 extends ArrayLike<number> {
 }
 
@@ -11024,10 +11028,6 @@ export interface Matrix4 extends ArrayLike<number> {
  */
 export class Matrix4 implements ArrayLike<number> {
     constructor(column0Row0?: number, column1Row0?: number, column2Row0?: number, column3Row0?: number, column0Row1?: number, column1Row1?: number, column2Row1?: number, column3Row1?: number, column0Row2?: number, column1Row2?: number, column2Row2?: number, column3Row2?: number, column0Row3?: number, column1Row3?: number, column2Row3?: number, column3Row3?: number);
-    /**
-     * The number of elements used to pack the object into an array.
-     */
-    static packedLength: number;
     /**
      * Stores the provided instance into the provided array.
      * @param value - The value to pack.
@@ -11066,27 +11066,6 @@ export class Matrix4 implements ArrayLike<number> {
      * @returns The modified result parameter or a new Matrix4 instance if one was not provided. (Returns undefined if matrix is undefined)
      */
     static clone(matrix: Matrix4, result?: Matrix4): Matrix4;
-    /**
-     * Creates a Matrix4 from 16 consecutive elements in an array.
-     * @example
-     * // Create the Matrix4:
-     * // [1.0, 2.0, 3.0, 4.0]
-     * // [1.0, 2.0, 3.0, 4.0]
-     * // [1.0, 2.0, 3.0, 4.0]
-     * // [1.0, 2.0, 3.0, 4.0]
-     *
-     * const v = [1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0, 4.0, 4.0, 4.0, 4.0];
-     * const m = Cesium.Matrix4.fromArray(v);
-     *
-     * // Create same Matrix4 with using an offset into an array
-     * const v2 = [0.0, 0.0, 1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0, 4.0, 4.0, 4.0, 4.0];
-     * const m2 = Cesium.Matrix4.fromArray(v2, 2);
-     * @param array - The array whose 16 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
-     * @param [startingIndex = 0] - The offset into the array of the first element, which corresponds to first column first row position in the matrix.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Matrix4 instance if one was not provided.
-     */
-    static fromArray(array: number[], startingIndex?: number, result?: Matrix4): Matrix4;
     /**
      * Computes a Matrix4 instance from a column-major order array.
      * @param values - The column-major order array.
@@ -11244,7 +11223,7 @@ export class Matrix4 implements ArrayLike<number> {
      * @param [result] - The object in which the result will be stored.
      * @returns The modified result parameter.
      */
-    static computeViewportTransformation(viewport?: any, nearDepthRange?: number, farDepthRange?: number, result?: Matrix4): Matrix4;
+    static computeViewportTransformation(viewport?: Viewport, nearDepthRange?: number, farDepthRange?: number, result?: Matrix4): Matrix4;
     /**
      * Computes a Matrix4 instance that transforms from world space to view space.
      * @param position - The position of the camera.
@@ -11750,6 +11729,59 @@ export class Matrix4 implements ArrayLike<number> {
      * @returns The modified result parameter.
      */
     static inverseTranspose(matrix: Matrix4, result: Matrix4): Matrix4;
+    /**
+     * Duplicates the provided Matrix4 instance.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Matrix4 instance if one was not provided.
+     */
+    clone(result?: Matrix4): Matrix4;
+    /**
+     * Compares this matrix to the provided matrix componentwise and returns
+     * <code>true</code> if they are equal, <code>false</code> otherwise.
+     * @param [right] - The right hand side matrix.
+     * @returns <code>true</code> if they are equal, <code>false</code> otherwise.
+     */
+    equals(right?: Matrix4): boolean;
+    /**
+     * Compares this matrix to the provided matrix componentwise and returns
+     * <code>true</code> if they are within the provided epsilon,
+     * <code>false</code> otherwise.
+     * @param [right] - The right hand side matrix.
+     * @param [epsilon = 0] - The epsilon to use for equality testing.
+     * @returns <code>true</code> if they are within the provided epsilon, <code>false</code> otherwise.
+     */
+    equalsEpsilon(right?: Matrix4, epsilon?: number): boolean;
+    /**
+     * Computes a string representing this Matrix with each row being
+     * on a separate line and in the format '(column0, column1, column2, column3)'.
+     * @returns A string representing the provided Matrix with each row being on a separate line and in the format '(column0, column1, column2, column3)'.
+     */
+    toString(): string;
+    /**
+     * The number of elements used to pack the object into an array.
+     */
+    static packedLength: number;
+    /**
+     * Creates a Matrix4 from 16 consecutive elements in an array.
+     * @example
+     * // Create the Matrix4:
+     * // [1.0, 2.0, 3.0, 4.0]
+     * // [1.0, 2.0, 3.0, 4.0]
+     * // [1.0, 2.0, 3.0, 4.0]
+     * // [1.0, 2.0, 3.0, 4.0]
+     *
+     * const v = [1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0, 4.0, 4.0, 4.0, 4.0];
+     * const m = Cesium.Matrix4.fromArray(v);
+     *
+     * // Create same Matrix4 with using an offset into an array
+     * const v2 = [0.0, 0.0, 1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0, 4.0, 4.0, 4.0, 4.0];
+     * const m2 = Cesium.Matrix4.fromArray(v2, 2);
+     * @param array - The array whose 16 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
+     * @param [startingIndex = 0] - The offset into the array of the first element, which corresponds to first column first row position in the matrix.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Matrix4 instance if one was not provided.
+     */
+    static fromArray(array: number[], startingIndex?: number, result?: Matrix4): Matrix4;
     /**
      * An immutable Matrix4 instance initialized to the identity matrix.
      */
@@ -11822,38 +11854,6 @@ export class Matrix4 implements ArrayLike<number> {
      * The index into Matrix4 for column 3, row 3.
      */
     static readonly COLUMN3ROW3: number;
-    /**
-     * Gets the number of items in the collection.
-     */
-    length: number;
-    /**
-     * Duplicates the provided Matrix4 instance.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Matrix4 instance if one was not provided.
-     */
-    clone(result?: Matrix4): Matrix4;
-    /**
-     * Compares this matrix to the provided matrix componentwise and returns
-     * <code>true</code> if they are equal, <code>false</code> otherwise.
-     * @param [right] - The right hand side matrix.
-     * @returns <code>true</code> if they are equal, <code>false</code> otherwise.
-     */
-    equals(right?: Matrix4): boolean;
-    /**
-     * Compares this matrix to the provided matrix componentwise and returns
-     * <code>true</code> if they are within the provided epsilon,
-     * <code>false</code> otherwise.
-     * @param [right] - The right hand side matrix.
-     * @param [epsilon = 0] - The epsilon to use for equality testing.
-     * @returns <code>true</code> if they are within the provided epsilon, <code>false</code> otherwise.
-     */
-    equalsEpsilon(right?: Matrix4, epsilon?: number): boolean;
-    /**
-     * Computes a string representing this Matrix with each row being
-     * on a separate line and in the format '(column0, column1, column2, column3)'.
-     * @returns A string representing the provided Matrix with each row being on a separate line and in the format '(column0, column1, column2, column3)'.
-     */
-    toString(): string;
 }
 
 /**
```

</details>

# Author checklist

#### PR Dependency Tree

* **PR #13290** 👈

### DIFF
--- a/packages/engine/Source/Core/Matrix2.js
+++ b/packages/engine/Source/Core/Matrix2.js
@@ -1001,36 +1001,100 @@ class Matrix2 {
   toString() {
     return `(${this[0]}, ${this[2]})\n` + `(${this[1]}, ${this[3]})`;
   }
+
+  /**
+   * The number of elements used to pack the object into an array.
+   * @type {number}
+   */
+  static packedLength = 4;
+
+  /**
+   * Creates a Matrix2 from 4 consecutive elements in an array.
+   *
+   * @function
+   * @param {number[]} array The array whose 4 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
+   * @param {number} [startingIndex=0] The offset into the array of the first element, which corresponds to first column first row position in the matrix.
+   * @param {Matrix2} [result] The object onto which to store the result.
+   * @returns {Matrix2} The modified result parameter or a new Matrix2 instance if one was not provided.
+   *
+   * @example
+   * // Create the Matrix2:
+   * // [1.0, 2.0]
+   * // [1.0, 2.0]
+   *
+   * const v = [1.0, 1.0, 2.0, 2.0];
+   * const m = Cesium.Matrix2.fromArray(v);
+   *
+   * // Create same Matrix2 with using an offset into an array
+   * const v2 = [0.0, 0.0, 1.0, 1.0, 2.0, 2.0];
+   * const m2 = Cesium.Matrix2.fromArray(v2, 2);
+   */
+  static fromArray = Matrix2.unpack;
+
+  /**
+   * An immutable Matrix2 instance initialized to the identity matrix.
+   *
+   * @type {Matrix2}
+   * @constant
+   */
+  static IDENTITY = Object.freeze(new Matrix2(1.0, 0.0, 0.0, 1.0));
+
+  /**
+   * An immutable Matrix2 instance initialized to the zero matrix.
+   *
+   * @type {Matrix2}
+   * @constant
+   */
+  static ZERO = Object.freeze(new Matrix2(0.0, 0.0, 0.0, 0.0));
+
+  /**
+   * The index into Matrix2 for column 0, row 0.
+   *
+   * @type {number}
+   * @constant
+   *
+   * @example
+   * const matrix = new Cesium.Matrix2();
+   * matrix[Cesium.Matrix2.COLUMN0ROW0] = 5.0; // set column 0, row 0 to 5.0
+   */
+  static COLUMN0ROW0 = 0;
+
+  /**
+   * The index into Matrix2 for column 0, row 1.
+   *
+   * @type {number}
+   * @constant
+   *
+   * @example
+   * const matrix = new Cesium.Matrix2();
+   * matrix[Cesium.Matrix2.COLUMN0ROW1] = 5.0; // set column 0, row 1 to 5.0
+   */
+  static COLUMN0ROW1 = 1;
+
+  /**
+   * The index into Matrix2 for column 1, row 0.
+   *
+   * @type {number}
+   * @constant
+   *
+   * @example
+   * const matrix = new Cesium.Matrix2();
+   * matrix[Cesium.Matrix2.COLUMN1ROW0] = 5.0; // set column 1, row 0 to 5.0
+   */
+  static COLUMN1ROW0 = 2;
+
+  /**
+   * The index into Matrix2 for column 1, row 1.
+   *
+   * @type {number}
+   * @constant
+   *
+   * @example
+   * const matrix = new Cesium.Matrix2();
+   * matrix[Cesium.Matrix2.COLUMN1ROW1] = 5.0; // set column 1, row 1 to 5.0
+   */
+  static COLUMN1ROW1 = 3;
 }
-
-/**
- * The number of elements used to pack the object into an array.
- * @type {number}
- */
-Matrix2.packedLength = 4;
-
-/**
- * Creates a Matrix2 from 4 consecutive elements in an array.
- *
- * @function
- * @param {number[]} array The array whose 4 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
- * @param {number} [startingIndex=0] The offset into the array of the first element, which corresponds to first column first row position in the matrix.
- * @param {Matrix2} [result] The object onto which to store the result.
- * @returns {Matrix2} The modified result parameter or a new Matrix2 instance if one was not provided.
- *
- * @example
- * // Create the Matrix2:
- * // [1.0, 2.0]
- * // [1.0, 2.0]
- *
- * const v = [1.0, 1.0, 2.0, 2.0];
- * const m = Cesium.Matrix2.fromArray(v);
- *
- * // Create same Matrix2 with using an offset into an array
- * const v2 = [0.0, 0.0, 1.0, 1.0, 2.0, 2.0];
- * const m2 = Cesium.Matrix2.fromArray(v2, 2);
- */
-Matrix2.fromArray = Matrix2.unpack;
 
 const scaleScratch1 = new Cartesian2();
 const scaleScratch2 = new Cartesian2();
@@ -1038,69 +1102,5 @@ const scratchColumn = new Cartesian2();
 const scaleScratch3 = new Cartesian2();
 const scaleScratch4 = new Cartesian2();
 const scaleScratch5 = new Cartesian2();
-
-/**
- * An immutable Matrix2 instance initialized to the identity matrix.
- *
- * @type {Matrix2}
- * @constant
- */
-Matrix2.IDENTITY = Object.freeze(new Matrix2(1.0, 0.0, 0.0, 1.0));
-
-/**
- * An immutable Matrix2 instance initialized to the zero matrix.
- *
- * @type {Matrix2}
- * @constant
- */
-Matrix2.ZERO = Object.freeze(new Matrix2(0.0, 0.0, 0.0, 0.0));
-
-/**
- * The index into Matrix2 for column 0, row 0.
- *
- * @type {number}
- * @constant
- *
- * @example
- * const matrix = new Cesium.Matrix2();
- * matrix[Cesium.Matrix2.COLUMN0ROW0] = 5.0; // set column 0, row 0 to 5.0
- */
-Matrix2.COLUMN0ROW0 = 0;
-
-/**
- * The index into Matrix2 for column 0, row 1.
- *
- * @type {number}
- * @constant
- *
- * @example
- * const matrix = new Cesium.Matrix2();
- * matrix[Cesium.Matrix2.COLUMN0ROW1] = 5.0; // set column 0, row 1 to 5.0
- */
-Matrix2.COLUMN0ROW1 = 1;
-
-/**
- * The index into Matrix2 for column 1, row 0.
- *
- * @type {number}
- * @constant
- *
- * @example
- * const matrix = new Cesium.Matrix2();
- * matrix[Cesium.Matrix2.COLUMN1ROW0] = 5.0; // set column 1, row 0 to 5.0
- */
-Matrix2.COLUMN1ROW0 = 2;
-
-/**
- * The index into Matrix2 for column 1, row 1.
- *
- * @type {number}
- * @constant
- *
- * @example
- * const matrix = new Cesium.Matrix2();
- * matrix[Cesium.Matrix2.COLUMN1ROW1] = 5.0; // set column 1, row 1 to 5.0
- */
-Matrix2.COLUMN1ROW1 = 3;
 
 export default Matrix2;
--- a/packages/engine/Source/Core/Matrix3.js
+++ b/packages/engine/Source/Core/Matrix3.js
@@ -1661,37 +1661,129 @@ class Matrix3 {
       `(${this[2]}, ${this[5]}, ${this[8]})`
     );
   }
+
+  /**
+   * The number of elements used to pack the object into an array.
+   * @type {number}
+   */
+  static packedLength = 9;
+
+  /**
+   * Creates a Matrix3 from 9 consecutive elements in an array.
+   *
+   * @function
+   * @param {number[]} array The array whose 9 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
+   * @param {number} [startingIndex=0] The offset into the array of the first element, which corresponds to first column first row position in the matrix.
+   * @param {Matrix3} [result] The object onto which to store the result.
+   * @returns {Matrix3} The modified result parameter or a new Matrix3 instance if one was not provided.
+   *
+   * @example
+   * // Create the Matrix3:
+   * // [1.0, 2.0, 3.0]
+   * // [1.0, 2.0, 3.0]
+   * // [1.0, 2.0, 3.0]
+   *
+   * const v = [1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
+   * const m = Cesium.Matrix3.fromArray(v);
+   *
+   * // Create same Matrix3 with using an offset into an array
+   * const v2 = [0.0, 0.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
+   * const m2 = Cesium.Matrix3.fromArray(v2, 2);
+   */
+  static fromArray = Matrix3.unpack;
+
+  /**
+   * An immutable Matrix3 instance initialized to the identity matrix.
+   *
+   * @type {Matrix3}
+   * @constant
+   */
+  static IDENTITY = Object.freeze(
+    new Matrix3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0),
+  );
+
+  /**
+   * An immutable Matrix3 instance initialized to the zero matrix.
+   *
+   * @type {Matrix3}
+   * @constant
+   */
+  static ZERO = Object.freeze(
+    new Matrix3(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0),
+  );
+
+  /**
+   * The index into Matrix3 for column 0, row 0.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN0ROW0 = 0;
+
+  /**
+   * The index into Matrix3 for column 0, row 1.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN0ROW1 = 1;
+
+  /**
+   * The index into Matrix3 for column 0, row 2.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN0ROW2 = 2;
+
+  /**
+   * The index into Matrix3 for column 1, row 0.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN1ROW0 = 3;
+
+  /**
+   * The index into Matrix3 for column 1, row 1.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN1ROW1 = 4;
+
+  /**
+   * The index into Matrix3 for column 1, row 2.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN1ROW2 = 5;
+
+  /**
+   * The index into Matrix3 for column 2, row 0.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN2ROW0 = 6;
+
+  /**
+   * The index into Matrix3 for column 2, row 1.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN2ROW1 = 7;
+
+  /**
+   * The index into Matrix3 for column 2, row 2.
+   *
+   * @type {number}
+   * @constant
+   */
+  static COLUMN2ROW2 = 8;
 }
-
-/**
- * The number of elements used to pack the object into an array.
- * @type {number}
- */
-Matrix3.packedLength = 9;
-
-/**
- * Creates a Matrix3 from 9 consecutive elements in an array.
- *
- * @function
- * @param {number[]} array The array whose 9 consecutive elements correspond to the positions of the matrix.  Assumes column-major order.
- * @param {number} [startingIndex=0] The offset into the array of the first element, which corresponds to first column first row position in the matrix.
- * @param {Matrix3} [result] The object onto which to store the result.
- * @returns {Matrix3} The modified result parameter or a new Matrix3 instance if one was not provided.
- *
- * @example
- * // Create the Matrix3:
- * // [1.0, 2.0, 3.0]
- * // [1.0, 2.0, 3.0]
- * // [1.0, 2.0, 3.0]
- *
- * const v = [1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
- * const m = Cesium.Matrix3.fromArray(v);
- *
- * // Create same Matrix3 with using an offset into an array
- * const v2 = [0.0, 0.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0];
- * const m2 = Cesium.Matrix3.fromArray(v2, 2);
- */
-Matrix3.fromArray = Matrix3.unpack;
 
 const scaleScratch1 = new Cartesian3();
 const scaleScratch2 = new Cartesian3();
@@ -1699,6 +1791,10 @@ const scratchColumn = new Cartesian3();
 const scaleScratch3 = new Cartesian3();
 const scaleScratch4 = new Cartesian3();
 const scaleScratch5 = new Cartesian3();
+
+const jMatrix = new Matrix3();
+const jMatrixTranspose = new Matrix3();
+const scratchTransposeMatrix = new Matrix3();
 
 /**
  * @param {Matrix3} matrix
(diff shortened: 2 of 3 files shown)


javagl:
Just a quick note: When this includes `extends Array`, wouldn't `extends Float64Array` be a ~"relatively small step" in terms of changes and implications, in view of https://github.com/CesiumGS/cesium/pull/12973

(I think that the answer is roughly: "This step is not as small as it may look like, because `Array` is just an interface, but `Float64Array` is a real 'thing'". But some of the concerns from that PR may still become obsolete. I'll have to spend some time to connect _all_ the dots here...)

(Tagging @Beilinson just for the connection)

jjspace:
I was about to post a comment but I noticed you just removed the extends. Still feel this is worth mentioning.

I'm a little worried about _actually_ extending `Array` for these classes. This inherently brings along all the "standard" array functions and may encourage use outside the expectations. For example what does `matrix.push(x)` mean? does that break any behavior? What if you call `matrix.sort()` which works _in place_? etc
These things are probably somewhat unlikely but with the `extends` they're allowed and would seem possible from the types perspective. May be best to avoid confusing users

donmccurdy:
Yeah, more worms in this can than I was looking for — I've reverted the change extending Array, and instead silenced the TS errors. Lack of https://github.com/microsoft/TypeScript/issues/48096 makes this somewhat awkward in JSDoc. I agree that Float64Array would be preferable to Array, if we do extend something.

javagl:
Based on the discussion in the linked PR, I think that there was the claim that ~"the `extends Float64Array` is not really 'visible' for the client". So specifically: Is there a way to let `MatrixN` inherit from `...Array` **privately**, so that users don't see that it is an array, so that they cannot call `sort()` and other problematic functions, but still use the `matrix[index]` access?

donmccurdy:
I've reworked this PR a bit to no longer depend on the `@overload` support, and to avoid changes to method return types. The PR description now includes a complete diff of `Cesium.d.ts` before vs. after.

donmccurdy:
Quick update — we discussed how to handle TypedArray in parameter and return types in #13302. I think this PR already reflects the decision we arrived at there, no changes needed.

Beilinson:
Saw #12973 mentioned here, is there any interest in continuing that PR or no decision made yet?

javagl:
I think that the reasoning behind making the matrices extend `Float...Array` sounded convincing (mainly in terms of performance). And the concerns that I initially had have largely been resolved (as of https://github.com/CesiumGS/cesium/pull/12973#issuecomment-3430060186 ). 

To put it that way: Your change involved converting the matrix classes to `class` as far as it was necessary for extending `Float64Array`. The matrix classes _will_ be converted to `class` anyhow, and in _this_ PR, this is happening more holistically. There may be some "deeper" aspects to the typing that may have to be sorted out, but maybe the step of extending `Float64Array` could eventually be done after the main `class`-conversion is settled?

donmccurdy:
Given that we already expose an array-like interface, I agree that extending Float64Array is probably a good direction. The results in https://github.com/toji/gl-matrix/issues/465 are particularly encouraging; we _should_ be minimizing the number of allocations and reusing instances aggressively, regardless.

I might prefer to do the ES6 class conversion first (I think this is ready any time) and to extend Float64Array in a subsequent release, so it's easier to isolate any regressions.

Beilinson:
Cool, Id be happy to update my PR when this merges :)

donmccurdy:
@Beilinson @javagl @jjspace when time allows, would one of you mind reviewing this? Note that the diff is most readable in "split" mode with whitespace changes hidden, and the diff of the generated `Cesium.d.ts` types is available in the PR description. Thank you!

packages/engine/Source/Core/Matrix2.js
```diff
@@ -24,949 +20,1024 @@ import DeveloperError from "./DeveloperError.js";
...
  */
-Matrix2.equalsArray = function (matrix, array, offset) {
-  return (
-    matrix[0] === array[offset] &&
-    matrix[1] === array[offset + 1] &&
-    matrix[2] === array[offset + 2] &&
-    matrix[3] === array[offset + 3]
-  );
-};
+Matrix2.packedLength = 4;
```
Beilinson: why not define these as static properties if already using static methods above?

packages/engine/Source/Core/Matrix2.js
```diff
@@ -24,949 +20,1024 @@ import DeveloperError from "./DeveloperError.js";
...
+   * @param {number[]} values The column-major order array.
+   * @param {Matrix2} [result] The object in which the result will be stored, if undefined a new instance will be created.
+   * @returns {Matrix2} The modified result parameter, or a new Matrix2 instance if one was not provided.
+   */
+  static fromColumnMajorArray(values, result) {
+    //>>includeStart('debug', pragmas.debug);
+    Check.defined("values", values);
+    //>>includeEnd('debug');
+
+    // @ts-expect-error TODO(tsd-jsdoc): Requires index signature support.
```
Beilinson, line 187: Most of these index signature support ts-errors should be fixed with my PR extending a TypedArray correct?

javagl (commented):
This is certainly an important step towards the ES6 class conversion. 

I think that it could be merged in its current form (and in doubt, could hit "Approve" at some point). But if there are no further changes, then I'd like to know how to organize possible follow-ups. (And apologies if _some_ of that has already been sorted out or is tracked in other issues/comments - if that's the case, then a link could be enough)

---

There are new exported types. 

One is `EigenDecompositionResult`. This is a very specific name and structure, used in exactly one function, and cannot change in any form, so it's probably OK to have that exported explicitly.

The other one is `Viewport`. And I was suprised that this class name did not exist otherwise. I wonder if we should ""waste"" this name for something that is only a "local typedef" for a single parameter of a single function. That type is _structurally_ the same as `BoundingRectangle`. The fact that the properties are _supposed_ to be integer values for `Viewport` cannot be captured. No strong opinions on all this, just musings...

---

# There is a breaking change 😨 !!!111

<sub>😉 No worries, it's a good one. </sub>
The fact that the [wrong `result` parameter type for `setColumn`](https://github.com/CesiumGS/cesium/blob/ac9521a291151183e100fa0986474f545325245b/packages/engine/Source/Core/Matrix2.js#L410) remained unnoticed for 14 (fourteen) years is impressive.

---

The style of defining the 'static' properties was already brought up by Beilinson in the inlined comments (and of course, refers to all static properties, and not only `packedLength`). I wonder if these could be pulled into the class, as in
```
class Matrix4 {

  /**
   * The number of elements used to pack the object into an array.
   * @type {number}
   * @readonly
   * @constant
   * @private
   */
  static packedLength = 16;
```

The choice of `@constant`, `@const {number}`, or `@readonly` (or combinations thereof) may depend on how well the current tooling works with each of these, and in how far they can simplify future (maybe automated) conversion steps towards actual TypeScript. Fow now, of course, _none_ of that is preventing a 
```
Cesium.Matrix4.packedLength = -666;
const m = new Cesium.Matrix4();
console.log("BOOM: "+m.length);
```
But that's a different story.

---

In a similar vein: There are some of these "free-floating pseudo-member functions" (that are one of my pet peeves), e.g. 
`function computeFrobeniusNorm(matrix)`
in `Matrix3`. I think that these _should_ be member functions. But given that this anti-pattern is used throughout the codebase, I don't know whether we should try to get rid of this as part of the incremental `class`-updates, or if we want to do a first, basic pass _everywhere_, and then try to address these pseudo-members in follow-ups.

Beilinson:
> The choice of `@constant`, `@const {number}`, or `@readonly` (or combinations thereof) may depend on how well the current tooling works with each of these, and in how far they can simplify future (maybe automated) conversion steps towards actual TypeScript. Fow now, of course, _none_ of that is preventing a
> 
> ```
> Cesium.Matrix4.packedLength = -666;
> const m = new Cesium.Matrix4();
> console.log("BOOM: "+m.length);
> ```
> 
```js
class Matrix4 {

  /**
   * The number of elements used to pack the object into an array.
   * @type {number}
   * @readonly
   * @constant
   * @private
   */
  static get packedLength() { return 16; }
```

This syntax works and allows static members to be readonly if intended

packages/engine/Source/Core/Matrix2.js
donmccurdy: I'm using [`lebab`](https://github.com/lebab/lebab) to automate conversions to ES6 Classes and make it deterministic, but I'm not sure why it didn't use `static` syntax for these. Will clean that up, I agree it's stylistically nicer.

packages/engine/Source/Core/Matrix2.js
donmccurdy: I believe so, yes!

donmccurdy:
Thanks @javagl and @Beilinson!

I've moved the property assignments from the MatrixN to `static` properties (by hand, as the lebab automation doesn't handle this). I haven't done any further changes to the existing code.

> I don't know whether we should try to get rid of this as part of the incremental class-updates, or if we want to do a first, basic pass everywhere, and then try to address these pseudo-members in follow-ups...

Personally I am not sure that moving all of this logic into OOP class hierarchies is worth the effort of a codebase-wide refactor, or necessarily an improvement... Pure functions scoped to a single file "can be" easier to reason about, though I realize there are some downstream challenges with private vs. internal visibility and code reuse to solve. If the team/maintainers are agreed on a direction, though, I'm happy to support that.

I would insist that it not be part of these (otherwise) fairly safe and automated ES6 conversions, though – we still have 500+ classes left to convert. In the interest of (1) converting quickly, (2) unblocking type checks to make future refactors safer, and (3) not having our ideas scattered across the hundreds of PRs needed to convert those 500+ classes, it might be best if we gather these ideas for future improvements to code style in a separate, longer-lived issue? 😅

donmccurdy:
Hm, it looks like I should have rebuilt `Cesium.d.ts` after moving the static class properties into the class body. Either tsd-jsdoc doesn't support that or I've gotten some part of the syntax wrong, because it's now generating types as if members like Matrix3.ZERO were instance properties and not static properties.

I'll open a separate PR tomorrow-ish as a quick fix on that. I hadn't been aware of this tsd-jsdoc limitation; maybe it's worth some automation to catch that in CI as well, similar to the proposal in #13205.

> **EDIT:** Proposed fix in #13376.

Issue #13296, opened 2026-03-13
Would adding OffscreenCanvas type to ImageryTypes be possible?


pakb:
https://github.com/CesiumGS/cesium/blob/7878464aae1d00b6610ae3cb4bbcedbc9aa379cb/packages/engine/Source/Scene/ImageryProvider.js#L8

I was wondering if there is any limitation to adding `OffscreenCanvas` here too.
I've been playing with a [third-party library to load GeoTIFF inside Cesium](https://github.com/hongfaqiu/TIFFImageryProvider) and they are using OffscreenCanvas to achieve that.
In a JS context (without type-checking) it works pretty well. But when using their code in a TypeScript env, it starts to raise type-checking errors because of their use of `OffscreenCanvas` as `ImageryType`.

Anyone with a good knowledge of the processing of `ImageryType` knows of a "show-stopper" reason not to add `OffscreenCanvas` here?

PR #13297, opened 2026-03-13
Add OffscreenCanvas to ImageryTypes typedef

pakb:
# Description

Adding a new type to ImageryTypes typedef

## Context

I was wondering if there is any limitation to adding OffscreenCanvas here too.
I've been playing with [a third-party library to load GeoTIFF inside Cesium](https://github.com/hongfaqiu/TIFFImageryProvider) and they are using OffscreenCanvas to achieve that. In a JS context (without type-checking) it works pretty well. But when using their code in a TypeScript env, it starts to raise type-checking errors because of their use of OffscreenCanvas as ImageryType.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/13296.

[I asked first here](https://github.com/CesiumGS/cesium/issues/13296) and then thought why not propose the change right away...

# Author checklist


lukemckinstry:
Thanks for submitting @pakb , I can take a first pass review

lukemckinstry:
I think this is ok following https://github.com/CesiumGS/cesium/pull/12558 since Texture accepts OffscreenCanvas. @jjhembd does this logic make sense to you as well?

PR #13301, opened 2026-03-13
BufferPrimitiveCollection: Picking API

donmccurdy:
# Description

Adds picking / interactivity support to BufferPointCollection, BufferPolylineCollection, and BufferPolygonCollection. `scene.pick(...)` returns a `{collection: BufferPrimitiveCollection, index: number}` result. From this the user can create or reuse a primitive; hovering over N polygons sequentially should only require a single BufferPolygon instance.

[bufferprim-picking.webm](https://github.com/user-attachments/assets/e00b9ddf-785f-4e6a-aa99-c4453f1f27dc)

## Issue number and link

n/a

## Testing plan

Unit tests added.

Picking code in a sandbox would look something like this:

```javascript
const handler = new Cesium.ScreenSpaceEventHandler(viewer.scene.canvas);

let selection = null;
const selectionOriginalColor = new Cesium.Color();

handler.setInputAction((movement) => {
  // Clear previous selection.
  if (selection) {
    const { collection, index } = selection;
    getCollectionPrimitive(collection, index).setColor(selectionOriginalColor);
    selection = null;
  }

  const picked = viewer.scene.pick(movement.endPosition);

  // Highlight current selection.
  if (picked && picked.collection instanceof BufferPrimitiveCollection) {
    const { collection, index } = picked;
    selection = { collection, index };

    const prim = getCollectionPrimitive(collection, index);
    prim.getColor(selectionOriginalColor);
    prim.setColor(Cesium.Color.CRIMSON);
  }

const point = new BufferPoint();
const line = new BufferPolyline();
const polygon = new BufferPolygon();

/**
 * Returns BufferPrimitive view onto collection[index]. Repeated
 * calls will modify and reuse the same result object.
 */
function getCollectionPrimitive(collection, index) {
  switch (collection.constructor) {
    case BufferPointCollection:
      return collection.get(index, point);
    case BufferPolylineCollection:
      return collection.get(index, line);
    case BufferPolygonCollection:
      return collection.get(index, polygon);
    default:
      throw new Error(`Unknown type picked: ${collection.constructor.name}`);
  }
}

```

# Author checklist

#### PR Dependency Tree

* **PR #13301** 👈


packages/engine/Source/Core/Color.js
```diff
@@ -446,10 +448,10 @@ Color.packedLength = 4;
  * Stores the provided instance into the provided array.
  *
  * @param {Color} value The value to pack.
- * @param {number[]} array The array to pack into.
+ * @param {number[]|TypedArray} array The array to pack into.
  * @param {number} [startingIndex=0] The index into the array at which to start packing the elements.
  *
- * @returns {number[]} The array that was packed into
+ * @returns {number[]|TypedArray} The array that was packed into
```
donmccurdy, line 454: This may need to change, discussion ongoing in https://github.com/CesiumGS/cesium/pull/13282.

packages/widgets/Source/Viewer/Viewer.js
```diff
@@ -116,6 +117,12 @@ function pickEntity(viewer, e) {
         feature: picked,
       });
     }
+
+    const collection = picked.collection;
+    if (collection instanceof BufferPrimitiveCollection) {
+      const PrimitiveClass = collection._getPrimitiveClass();
+      return collection.get(picked.index, new PrimitiveClass());
+    }
```
donmccurdy: Changes in `Viewer.js` are not really required, but it was nice to see the feature ID pop up. I don't think we'd want to do `new PrimitiveClass()` here on every hover event, but this appears to only run on click events, which doesn't seem too bad. Or happy to revert changes to `Viewer.js`, if we want to consider that more!

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -100,8 +109,15 @@ function renderBufferPointCollection(collection, frameState, renderContext) {
         continue;
       }
 
+      if (point._pickId === 0) {
+        pickIds[i] = context.createPickId({ collection, index: i });
```
donmccurdy: This allocates one PickId instance and one Color instance for every point/line/polygon in the collection. I'd been hoping to avoid that type of O(n) class instance allocation, but didn't see an obvious way.

I'll try some quick performance measurements here — my hunch is that it's fine for now, but a good potential spot for future cleanup.

lilleyse:
I _think_ this may require some tweaks to work with `scene.drillPick`.

Check out the code [here](https://github.com/CesiumGS/cesium/blob/85fb12556a499c9e8059e2ab92f4977814fc622c/packages/engine/Source/Scene/Picking.js#L841-L869). The pick object needs to be hideable in order to continue drill picking. The `Cesium3DTileFeature` approach might make sense to use here.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -476,6 +482,7 @@ class BufferPrimitiveCollection {
     result.featureId = this._primitiveCount - 1;
     result.show = options.show ?? true;
     result.setColor(options.color ?? Color.WHITE);
+    result._pickId = 0; // unset
```
danielzhong, line 502: I'm thinking a hard copy problem here:

`add()` correctly initializes `_pickId = 0`, but `clone()` does a `raw _copySubDataView` of the entire primitive buffer — which now includes the PICK_ID_U32 slot. Cloning a rendered collection will copy stale pick IDs into the result, and since the render path only allocates a fresh `PickId` when `_pickId === 0`, the clone never gets valid pick registrations. Should `clone()` also zero out `_pickId` for each primitive after the copy?

packages/engine/Source/Scene/renderBufferPointCollection.js
danielzhong: We should probably destroy the existing `pickIds[i]` before replacing it. 

The issue isn't just an object memory leak Cesium's `createPickId()` registers an entry in the context's global pick registry `(context._pickObjects[key] = payload)`. Without calling `destroy()`, that registry entry is never removed, so the { collection,  index } payload stays alive indefinitely and the pick color slot remains permanently occupied. Repeated `sort()` calls will continuously leak pick registrations.

```
    if (point._pickId === 0) {
      pickIds[i]?.destroy(); // destroy stale PickId before replacing
      pickIds[i] = context.createPickId({ collection, index: i });
      point._pickId = pickIds[i].key;
    }
```

donmccurdy:
@lilleyse I hadn't noticed `scene.drillPick`, thanks! That should be fixed now. I opted to make `picked.primitive` a getter returning a new BufferPrimitive, rather than allocating objects for every primitive up front. It would be nice to avoid the getter too, but this keeps the changes self-contained without changes to picking. We could reduce those allocations later with changes to picking, if we want.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
donmccurdy: Ah good catch! I agree, `clone()` should zero out the pick IDs in the new collection, done.

donmccurdy: Hm, I think .sort() has a similar problem. After sorting, the indices of the primitives no longer match the indices of the PickIds. I think we don't actually care about looking up the PickIds — they're not accessed except to destroy them — so perhaps PickIds could just be an unordered set or list.

packages/engine/Source/Scene/renderBufferPointCollection.js
donmccurdy: My intention is that _pickId is 0 before initialization, and immutable after that, so it would be a bug if a PickID is ever allocated for the same primitive twice. There's also the issue with .sort() above, so I'm inclined to not track PickIDs by index, but instead to just put them in an unordered list or set for cleanup purposes.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
donmccurdy: Ok, should be handled by https://github.com/CesiumGS/cesium/pull/13301/commits/880994561261f67c91d3fcd26097a48018457ae1!

packages/engine/Source/Scene/renderBufferPointCollection.js
donmccurdy: Should be handled by https://github.com/CesiumGS/cesium/pull/13301/commits/880994561261f67c91d3fcd26097a48018457ae1.

packages/widgets/Source/Viewer/Viewer.js
donmccurdy: Reverted the Viewer.js changes for now.

donmccurdy:
Unfortunately, after doing some performance measurements ... `.createPickId()` has a pretty high cost relative to the rest of the renderer implementation. For a collection of 300K points, the first call to `renderBufferPointCollection` is:

- 25–35ms without pick IDs
- 65–75ms with `context.createPickId`
- 100–120ms with `context.createPickId` + `result.primitive` getter

I was a bit worried about it, but didn't expect a 2–4x slowdown! For now maybe we put this feature behind an `allowPicking: true` constructor parameter on the collection? Possibly this is something that we could optimize in the picking implementation later on.

> **EDIT:** I've put this behind an `allowPicking` flag for now.

packages/engine/Source/Scene/renderBufferPointCollection.js
donmccurdy: Hm, the performance measurements weren't as good as I'd hoped. Left a comment with more details in the PR main thread.

PR #13302, opened 2026-03-13
Cartesian2-4: .pack/.packArray return type compatibility fix

### LINKED ISSUE #10455: Update or remove tsd-jsdoc

donmccurdy:
# Description

Following up on earlier discussions (see issue links below), I think our best option right now might be to keep the public-facing types as stable as possible, so that users don't need to do temporary migrations (adding/removing `as Cartesian3` casts) for issues that will no longer exist after we've converted to ES6 classes and resolved #10455.

This PR takes a first step on that, reverting a change to the return types of pack/unpack methods, and maintaining backward-compatibility with previous releases. I've kept the correct _parameter types_ intact, since there's no compatibility issue with those and it's one less thing we need to override.

## Issue number and link

- https://github.com/CesiumGS/cesium/pull/13168#issuecomment-4054866265
- https://github.com/CesiumGS/cesium/pull/13282#issuecomment-4056181130

## Testing plan

Diff for Cesium.d.ts, before / after:

```diff
diff --git a/Source/SrcCesium.d.ts b/Source/Cesium.d.ts
index 0b09075bb2..83ce2ca07b 100644
--- a/Source/SrcCesium.d.ts
+++ b/Source/Cesium.d.ts
@@ -1528,7 +1528,7 @@ export class Cartesian2 {
      * @param [startingIndex = 0] - The index into the array at which to start packing the elements.
      * @returns The array that was packed into
      */
-    static pack(value: Cartesian2, array: number[] | TypedArray, startingIndex?: number): number[] | TypedArray;
+    static pack(value: Cartesian2, array: number[] | TypedArray, startingIndex?: number): number[];
     /**
      * Retrieves an instance from a packed array.
      * @param array - The packed array.
@@ -1543,7 +1543,7 @@ export class Cartesian2 {
      * @param [result] - The array onto which to store the result. If this is a typed array, it must have array.length * 2 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 2) elements.
      * @returns The packed array.
      */
-    static packArray(array: Cartesian2[], result?: number[] | TypedArray): number[] | TypedArray;
+    static packArray(array: Cartesian2[], result?: number[] | TypedArray): number[];
     /**
      * Unpacks an array of cartesian components into an array of Cartesian2s.
      * @param array - The array of components to unpack.
@@ -1878,7 +1878,7 @@ export class Cartesian3 {
      * @param [startingIndex = 0] - The index into the array at which to start packing the elements.
      * @returns The array that was packed into
      */
-    static pack(value: Cartesian3, array: number[] | TypedArray, startingIndex?: number): number[] | TypedArray;
+    static pack(value: Cartesian3, array: number[] | TypedArray, startingIndex?: number): number[];
     /**
      * Retrieves an instance from a packed array.
      * @param array - The packed array.
@@ -1893,7 +1893,7 @@ export class Cartesian3 {
      * @param [result] - The array onto which to store the result. If this is a typed array, it must have array.length * 3 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 3) elements.
      * @returns The packed array.
      */
-    static packArray(array: Cartesian3[], result?: number[] | TypedArray): number[] | TypedArray;
+    static packArray(array: Cartesian3[], result?: number[] | TypedArray): number[];
     /**
      * Unpacks an array of cartesian components into an array of Cartesian3s.
      * @param array - The array of components to unpack.
@@ -2312,7 +2312,7 @@ export class Cartesian4 {
      * @param [startingIndex = 0] - The index into the array at which to start packing the elements.
      * @returns The array that was packed into
      */
-    static pack(value: Cartesian4, array: number[] | TypedArray, startingIndex?: number): number[] | TypedArray;
+    static pack(value: Cartesian4, array: number[] | TypedArray, startingIndex?: number): number[];
     /**
      * Retrieves an instance from a packed array.
      * @param array - The packed array.
@@ -2327,7 +2327,7 @@ export class Cartesian4 {
      * @param [result] - The array onto which to store the result. If this is a typed array, it must have array.length * 4 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 4) elements.
      * @returns The packed array.
      */
-    static packArray(array: Cartesian4[], result?: number[] | TypedArray): number[] | TypedArray;
+    static packArray(array: Cartesian4[], result?: number[] | TypedArray): number[];
     /**
      * Unpacks an array of cartesian components into an array of Cartesian4s.
      * @param array - The array of components to unpack.

```

# Author checklist

#### PR Dependency Tree

* **PR #13302** 👈

### DIFF
--- a/packages/engine/Source/Core/Cartesian2.js
+++ b/packages/engine/Source/Core/Cartesian2.js
@@ -79,7 +79,7 @@ class Cartesian2 {
    * Stores the provided instance into the provided array.
    *
    * @param {Cartesian2} value The value to pack.
-   * @param {number[]|TypedArray} array The array to pack into.
+   * @param {number[]} array The array to pack into.
    * @param {number} [startingIndex=0] The index into the array at which to start packing the elements.
    *
    * @returns {number[]} The array that was packed into
@@ -95,14 +95,13 @@ class Cartesian2 {
     array[startingIndex++] = value.x;
     array[startingIndex] = value.y;
 
-    // @ts-expect-error Requires https://github.com/CesiumGS/cesium/issues/10455.
     return array;
   }
 
   /**
    * Retrieves an instance from a packed array.
    *
-   * @param {number[]|TypedArray} array The packed array.
+   * @param {number[]} array The packed array.
    * @param {number} [startingIndex=0] The starting index of the element to be unpacked.
    * @param {Cartesian2} [result] The object into which to store the result.
    * @returns {Cartesian2} The modified result parameter or a new Cartesian2 instance if one was not provided.
@@ -126,7 +125,7 @@ class Cartesian2 {
    * Flattens an array of Cartesian2s into an array of components.
    *
    * @param {Cartesian2[]} array The array of cartesians to pack.
-   * @param {number[]|TypedArray} [result] The array onto which to store the result. If this is a typed array, it must have array.length * 2 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 2) elements.
+   * @param {number[]} [result] The array onto which to store the result. If this is a typed array, it must have array.length * 2 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 2) elements.
    * @returns {number[]} The packed array.
    */
   static packArray(array, result) {
@@ -138,6 +137,7 @@ class Cartesian2 {
     const resultLength = length * 2;
     if (!defined(result)) {
       result = new Array(resultLength);
+      // @ts-expect-error TODO(tsd-jsdoc): See https://github.com/CesiumGS/cesium/pull/13302.
     } else if (!Array.isArray(result) && result.length !== resultLength) {
       //>>includeStart('debug', pragmas.debug);
       throw new DeveloperError(
@@ -152,7 +152,6 @@ class Cartesian2 {
       Cartesian2.pack(array[i], result, i * 2);
     }
 
-    // @ts-expect-error Requires https://github.com/CesiumGS/cesium/issues/10455.
     return result;
   }
 
--- a/packages/engine/Source/Core/Cartesian3.js
+++ b/packages/engine/Source/Core/Cartesian3.js
@@ -117,7 +117,7 @@ class Cartesian3 {
    * Stores the provided instance into the provided array.
    *
    * @param {Cartesian3} value The value to pack.
-   * @param {number[]|TypedArray} array The array to pack into.
+   * @param {number[]} array The array to pack into.
    * @param {number} [startingIndex=0] The index into the array at which to start packing the elements.
    *
    * @returns {number[]} The array that was packed into
@@ -134,14 +134,13 @@ class Cartesian3 {
     array[startingIndex++] = value.y;
     array[startingIndex] = value.z;
 
-    // @ts-expect-error Requires https://github.com/CesiumGS/cesium/issues/10455.
     return array;
   }
 
   /**
    * Retrieves an instance from a packed array.
    *
-   * @param {number[]|TypedArray} array The packed array.
+   * @param {number[]} array The packed array.
    * @param {number} [startingIndex=0] The starting index of the element to be unpacked.
    * @param {Cartesian3} [result] The object into which to store the result.
    * @returns {Cartesian3} The modified result parameter or a new Cartesian3 instance if one was not provided.
@@ -166,7 +165,7 @@ class Cartesian3 {
    * Flattens an array of Cartesian3s into an array of components.
    *
    * @param {Cartesian3[]} array The array of cartesians to pack.
-   * @param {number[]|TypedArray} [result] The array onto which to store the result. If this is a typed array, it must have array.length * 3 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 3) elements.
+   * @param {number[]} [result] The array onto which to store the result. If this is a typed array, it must have array.length * 3 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 3) elements.
    * @returns {number[]} The packed array.
    */
   static packArray(array, result) {
@@ -178,6 +177,7 @@ class Cartesian3 {
     const resultLength = length * 3;
     if (!defined(result)) {
       result = new Array(resultLength);
+      // @ts-expect-error TODO(tsd-jsdoc): See https://github.com/CesiumGS/cesium/pull/13302.
     } else if (!Array.isArray(result) && result.length !== resultLength) {
       //>>includeStart('debug', pragmas.debug);
       throw new DeveloperError(
@@ -192,14 +192,13 @@ class Cartesian3 {
       Cartesian3.pack(array[i], result, i * 3);
     }
 
-    // @ts-expect-error Requires https://github.com/CesiumGS/cesium/issues/10455.
     return result;
   }
 
   /**
    * Unpacks an array of cartesian components into an array of Cartesian3s.
    *
-   * @param {number[]|TypedArray} array The array of components to unpack.
+   * @param {number[]} array The array of components to unpack.
    * @param {Cartesian3[]} [result] The array onto which to store the result.
    * @returns {Cartesian3[]} The unpacked array.
    */
--- a/packages/engine/Source/Core/Cartesian4.js
+++ b/packages/engine/Source/Core/Cartesian4.js
@@ -124,7 +124,7 @@ class Cartesian4 {
    * Stores the provided instance into the provided array.
    *
    * @param {Cartesian4} value The value to pack.
-   * @param {number[]|TypedArray} array The array to pack into.
+   * @param {number[]} array The array to pack into.
    * @param {number} [startingIndex=0] The index into the array at which to start packing the elements.
    *
    * @returns {number[]} The array that was packed into
@@ -142,14 +142,13 @@ class Cartesian4 {
     array[startingIndex++] = value.z;
     array[startingIndex] = value.w;
 
-    // @ts-expect-error Requires https://github.com/CesiumGS/cesium/issues/10455.
     return array;
   }
 
   /**
    * Retrieves an instance from a packed array.
    *
-   * @param {number[]|TypedArray} array The packed array.
+   * @param {number[]} array The packed array.
    * @param {number} [startingIndex=0] The starting index of the element to be unpacked.
    * @param {Cartesian4} [result] The object into which to store the result.
    * @returns {Cartesian4}  The modified result parameter or a new Cartesian4 instance if one was not provided.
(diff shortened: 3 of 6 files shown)


javagl:
This was apparently sparked by the comment at https://github.com/CesiumGS/cesium/pull/13168#issuecomment-4054866265 . At this point, I mainly wanted to _mention_ that the previous change was a _breaking_ one (and that it might be worth keeping an eye on that, maybe via the `d.ts` file, until something like 'api-extractor' can be used). 

But I think that this change for itself was OK. So before diving into details and considerations for a review here:

---

EDIT: I just saw https://github.com/CesiumGS/cesium/pull/13282#issuecomment-4057450889 - so if this part of a broader "rollback", that may be reviewed in a different light...

---

**1.**: Isn't the possibility to `pack` into a `TypedArray` a _requirement_ for the ongoing vector efforts?
(Of course, JavaScript doesn't care, but still...)

**2.**: Is the _new_ signature actually correct?
From skimmig over it, I think that the signature of
`static pack(value: Cartesian3, array: number[] | TypedArray, startingIndex?: number): number[];`
would allow something like
```
const value = new Cartesian3();
const target = new Float32Array(3);
const result = Cartesian3.pack(value, target, 0);
result.push(123); // XXX Called on a Float32Array, thinking that it is a number[] ?!
```
(I'll have to check, maybe I misunderstood something here)

donmccurdy:
The existing types are incorrect — I think it's just a matter of when we choose to fix them. We could fix the parameter types now without harm, but "fixing" the return types without access to TS-flavored JSDoc features will create churn for users — a breaking change now, and another breaking change later when we start using `@overload` or similar.

So I'm having a change of heart, and thinking we should leave the return types alone, or (more generally) try to avoid breaking changes that could be avoided just by waiting a while. For current needs in the vector codebase, we don't really need the return type anyway:

```javascript
Cartesian3.pack(value, target, 0);
```

javagl:
>  For current needs in the vector codebase, we don't really need the return type anyway:
> 
> Cartesian3.pack(value, target, 0);

That _avoids_ the error, but does not prevent it on the level of the types. With the current change here, it is possible to do
```javascript
const value = new Cartesian3();
const targetA = new Array(3);
const targetB = new Float32Array(3);
const resultA = Cartesian3.pack(value, targetA, 0);
const resultB = Cartesian3.pack(value, targetB, 0);
resultA.push(123); // Works
resultB.push(234); // Doesn't work
``` 

Or more specifically:

1. `@param {number[]} [result]` and `@returns {number[]}` is typesafe
2. `@param {number[]|TypedArray} [result]` with `@returns {number[]}` is **not** typesafe
3. `@param {number[]|TypedArray} [result]` with `@returns {number[]|TypedArray}` **would** be typesafe (even though it's a breaking change of the typing)

This does not necessarily have to prevent this change. But I'd leave the review/approval to others in this case.
I'd rather prefer options 1. (as it is) or 3. (a minor breaking change), compared to 2. (introducing a wrong typing...)

donmccurdy:
I think we're both agreed on what the final result should be, so it's just the intermediate state (while we're limited by tsd-jsdoc syntax) that's the challenge. My preference was 2 > 1 > 3, let's split the difference and do (1)? Or generalizing that... let's try to minimize changes to types unless they're trivial/uncontroversial for now. That way I think we can get ES6 classes done sooner, which is a blocker to being able to use `tsc` more broadly anyway.

donmccurdy:
@javagl @jjspace do we feel OK about taking approach (1) above...

> `@param {number[]} [result]` 
> `@returns {number[]}`

...at least as a general guideline during the migration? This does mean we'll need to use `@ts-expect-error` anywhere we pass a TypedArray in as a parameter in the meantime, but maybe that's preferable to changing the public API types to something now that we know isn't the final goal.

If so I'll update this PR accordingly. I'd like to try to land this PR (in some form) by end of week if that sounds OK.

donmccurdy:
Updated to intended state, thanks!

> That 'Approve' refers to the intended state here. (Is there a way to un-approve things? Sorry for the noise...

Personally I'm a _huge_ fan of “approval with comments.”[^1][^2] We don't get some benefits in a reviewer-presses-merge workflow, but nothing to apologize for in my opinion. :)

<details>
<summary>Updated diff of `Cesium.d.ts`</summary>

```diff
diff --git a/Source/Cesium.main.d.ts b/Source/Cesium.d.ts
index 375fc97355..42848258c5 100644
--- a/Source/Cesium.main.d.ts
+++ b/Source/Cesium.d.ts
@@ -1528,7 +1528,7 @@ export class Cartesian2 {
      * @param [startingIndex = 0] - The index into the array at which to start packing the elements.
      * @returns The array that was packed into
      */
-    static pack(value: Cartesian2, array: number[] | TypedArray, startingIndex?: number): number[] | TypedArray;
+    static pack(value: Cartesian2, array: number[], startingIndex?: number): number[];
     /**
      * Retrieves an instance from a packed array.
      * @param array - The packed array.
@@ -1536,14 +1536,14 @@ export class Cartesian2 {
      * @param [result] - The object into which to store the result.
      * @returns The modified result parameter or a new Cartesian2 instance if one was not provided.
      */
-    static unpack(array: number[] | TypedArray, startingIndex?: number, result?: Cartesian2): Cartesian2;
+    static unpack(array: number[], startingIndex?: number, result?: Cartesian2): Cartesian2;
     /**
      * Flattens an array of Cartesian2s into an array of components.
      * @param array - The array of cartesians to pack.
      * @param [result] - The array onto which to store the result. If this is a typed array, it must have array.length * 2 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 2) elements.
      * @returns The packed array.
      */
-    static packArray(array: Cartesian2[], result?: number[] | TypedArray): number[] | TypedArray;
+    static packArray(array: Cartesian2[], result?: number[]): number[];
     /**
      * Unpacks an array of cartesian components into an array of Cartesian2s.
      * @param array - The array of components to unpack.
@@ -1878,7 +1878,7 @@ export class Cartesian3 {
      * @param [startingIndex = 0] - The index into the array at which to start packing the elements.
      * @returns The array that was packed into
      */
-    static pack(value: Cartesian3, array: number[] | TypedArray, startingIndex?: number): number[] | TypedArray;
+    static pack(value: Cartesian3, array: number[], startingIndex?: number): number[];
     /**
      * Retrieves an instance from a packed array.
      * @param array - The packed array.
@@ -1886,21 +1886,21 @@ export class Cartesian3 {
      * @param [result] - The object into which to store the result.
      * @returns The modified result parameter or a new Cartesian3 instance if one was not provided.
      */
-    static unpack(array: number[] | TypedArray, startingIndex?: number, result?: Cartesian3): Cartesian3;
+    static unpack(array: number[], startingIndex?: number, result?: Cartesian3): Cartesian3;
     /**
      * Flattens an array of Cartesian3s into an array of components.
      * @param array - The array of cartesians to pack.
      * @param [result] - The array onto which to store the result. If this is a typed array, it must have array.length * 3 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 3) elements.
      * @returns The packed array.
      */
-    static packArray(array: Cartesian3[], result?: number[] | TypedArray): number[] | TypedArray;
+    static packArray(array: Cartesian3[], result?: number[]): number[];
     /**
      * Unpacks an array of cartesian components into an array of Cartesian3s.
      * @param array - The array of components to unpack.
      * @param [result] - The array onto which to store the result.
      * @returns The unpacked array.
      */
-    static unpackArray(array: number[] | TypedArray, result?: Cartesian3[]): Cartesian3[];
+    static unpackArray(array: number[], result?: Cartesian3[]): Cartesian3[];
     /**
      * Computes the value of the maximum component for the supplied Cartesian.
      * @param cartesian - The cartesian to use.
@@ -2312,7 +2312,7 @@ export class Cartesian4 {
      * @param [startingIndex = 0] - The index into the array at which to start packing the elements.
      * @returns The array that was packed into
      */
-    static pack(value: Cartesian4, array: number[] | TypedArray, startingIndex?: number): number[] | TypedArray;
+    static pack(value: Cartesian4, array: number[], startingIndex?: number): number[];
     /**
      * Retrieves an instance from a packed array.
      * @param array - The packed array.
@@ -2320,21 +2320,21 @@ export class Cartesian4 {
      * @param [result] - The object into which to store the result.
      * @returns The modified result parameter or a new Cartesian4 instance if one was not provided.
      */
-    static unpack(array: number[] | TypedArray, startingIndex?: number, result?: Cartesian4): Cartesian4;
+    static unpack(array: number[], startingIndex?: number, result?: Cartesian4): Cartesian4;
     /**
      * Flattens an array of Cartesian4s into an array of components.
      * @param array - The array of cartesians to pack.
      * @param [result] - The array onto which to store the result. If this is a typed array, it must have array.length * 4 components, else a {@link DeveloperError} will be thrown. If it is a regular array, it will be resized to have (array.length * 4) elements.
      * @returns The packed array.
      */
-    static packArray(array: Cartesian4[], result?: number[] | TypedArray): number[] | TypedArray;
+    static packArray(array: Cartesian4[], result?: number[]): number[];
     /**
      * Unpacks an array of cartesian components into an array of Cartesian4s.
      * @param array - The array of components to unpack.
      * @param [result] - The array onto which to store the result.
      * @returns The unpacked array.
      */
-    static unpackArray(array: number[] | TypedArray, result?: Cartesian4[]): Cartesian4[];
+    static unpackArray(array: number[], result?: Cartesian4[]): Cartesian4[];
     /**
      * Computes the value of the maximum component for the supplied Cartesian.
      * @param cartesian - The cartesian to use.

```

</details>

[^1]: [How to review code effectively: A GitHub staff engineer’s philosophy](https://github.blog/developer-skills/github/how-to-review-code-effectively-a-github-staff-engineers-philosophy/#to-approve-or-not)
[^2]: https://swizec.com/blog/approve-with-comment/

javagl:
I had another look at the changes and the TS diff. 

👍 

And I took two notes in my TODO list:

TODO Read https://swizec.com/blog/approve-with-comment/
TODO Read and bookmark https://swizec.com/blog/why-senior-engineers-get-nothing-done/

Issue #13306, opened 2026-03-15
Many clipping polygons cause WebGL crash


javagl:
### What happened?

I noticed this in a [review of another PR](https://github.com/CesiumGS/cesium/pull/13255#pullrequestreview-3908827404): 

When creating "many" clipping polygons with "many" points, then this will cause a WebGL crash.

### Reproduction steps

**1.**: Open this Sandcastle 

[...]

Issue #12649, opened 2025-06-03
Robust Clipping Polygon v2 (terrain and 3D tileset)


tbenazzi:
# Robust Clipping Polygon v2 (terrain and 3D tileset)

## Summary

We need the capability to topologically clip terrain and 3D tiles based on complex polygons with production-ready performance and precision. [While a clipping solution already exists in CesiumJS](https://sandcastle.cesium.com/?src=AEC%20Clipping.html), it has critical limitations that prevent production deployment.

## Current Issues

The existing clipping polygon implementation suffers from three major problems:

### 1. Precision Issues
- [Clipping accuracy errors of several meters between the input polygon and actual geometry clipping](https://github.com/CesiumGS/cesium/issues/12321)
- [Topological precision loss affecting project integration workflows](https://sandcastle.cesium.com/?src=AEC%20Clipping.html) 

![Image](https://github.com/user-attachments/assets/ddcf0ff7-3c3b-4f74-a89e-f2e61f15d7fc)
![Image](https://github.com/user-attachments/assets/edfb40ac-d989-45f4-aee2-ba7b98691252)
![Image](https://github.com/user-attachments/assets/b72d44ad-dc7e-4784-be69-247e5563b1e1)

### 2. Performance Degradation
- [Geometry is not cached, causing severe performance drops](https://github.com/CesiumGS/cesium/issues/12258)
- Clipping calculations performed every frame instead of being optimized
- [Frame rate becomes unusable with complex polygons or large datasets](https://github.com/CesiumGS/cesium/pull/11750#issuecomment-2416350404)

### 3. Shadow Rendering Problems
- [Shadow engine continues to cast shadows from hidden/clipped geometry](https://github.com/CesiumGS/cesium/issues/6261)
- Makes shadow studies impossible for integrated projects who replace a part of 3d tileset

## Business Impact

This is currently the **n°1 feature request** from our hundreds of production users. The ability to client-side integrate custom projects and cleanly clip portions of tilesets without performance loss is essential for shadow studies and project visualization.

This functionality is core not just for our use case, but for the entire **AEC (Architecture, Engineering, Construction) ecosystem** that needs to integrate future projects into existing base datasets.

## Use Cases

- **Urban Planning**: Integrate new building designs into existing city models
- **Construction**: Visualize project phases by clipping existing terrain/buildings
- **Shadow Analysis**: Perform accurate shadow studies on integrated projects
- **Environmental Impact**: Assess how new developments affect existing landscapes
[...]

PR #13308, opened 2026-03-17
ClippingPolygonCollection Performance Optimizations (GPU/CPU)

alarkbentley:
# Description

## Summary

This PR introduces a series of performance optimizations to `ClippingPolygonCollection` and the corresponding `PolygonSignedDistanceFS` compute shader, significantly reducing the cost of signed-distance-field computation for clipping polygons—especially when many polygons are present.

Before this PR creating a `ClippingPolygonCollection` with 64 polygons each with 256 vertices would crash the browser.  After this fix the same case renders within a few ms.

## Demo

1. Open [Sandcastle](https://sandcastle.cesium.com/index.html#c=tVdbb9s2FP4rhIGhcuMydpr04jjGtrTNOqRL0KTdQ5wHWqJtIhSpkZRTdfB/36EoiZSidAGGGUhMHp7rdy6kYym0QVtG76lCJ0jQe3RKNctT/LWkRYtBXO5PpTCECaoWgxH6e6EWAiHDUsqBNkUrwjUdOSoRLCWGSdEm65gK+kkm9JLFd1S1D5dE03NSUNV3uOZy2bGxpjIGVcBYeftRijNHu1Ryy+DousgoPru4ODt/Xwrthsf2y/65cHHpEU7oMl9fbeT9B0VSqi+puqKxFAmgYVROG6G4RMowTjU1cEjuCTO1+VhRYuiZlGtOLzfSSNhzpg2LX767diJRBZoUvPiimVj/yczGSZw10ViLobctTzPFUmbYlmpMkiSqXPFhrXIRW9yR8+aUqRi8kUwYHYECQ9W5FCNULYkZIUUSlusREnnq+IZ1al20WUmEYG9ujx19JRWKOADAgDo+hq+Zl4bt3l6jolZCgJOh/YCtfc7B4xPUOIj20CdiNhggigh67jaXH2F1MIR/zuWuCmK8ClhXKmKpn6DCBYmzXG+iG24RAnW3w+p8574UNbkSFW95tHsE9muqzTtiSJT1IWqqU3C3gclApU7RYvABxHNFTyXntNQIrVbzCKhNy2PlA3Ks9NQr8qose8AGQSqZUWUYtfyNNghpKtfxNKGrqVV1cXY6neCX09PPV28OFwO0azT41cp5CXpuvPrAhQfxtPwoWzelRhXTtpAX+5Rzwy4lL9YtBNwnllIlTADO1oEbh/HtbYtrF+z8uubZHbcyWieknVNb4XGuFFSUPbySuYIx6Buh9/xqQzI/L4guRIya4sizxPYkZ1kGvV+Fp6OyKdz6sTaUnGIu1zCJv5Q60D0MDhQInkCm9kIC7IBEYIY1GgOe0kDjZtjSJ7ahZw/jxpyKtdl027saTknAqGgqtzR6qOKGdTvqUTPlYAl7Ju6A9otSpPC5CFnXiiVX7DuF07LpVxzqJXLz5C9lQriHw2PUD0DA1D/R1t/a+u10qy0PO6NpXZTj76eGoXPejNieoe3L+MXrIzw5Ojgcv4EsgvnnaIwPJiMUlP3Lt/jt4au3r48sR9FwjPFk1M27t+5z173R4E76XUvh0wM1SJLApf5pV597Qz1VYm8vvw98elgS5VTu8j7KX3agxbKv/jBQmJ2AeEt4Dl8rJpIoKqnFEJ3MfWwVBjAW4XWTVCw4c0XRE2OdSs1sr0M2A5h6fax14Q2jiqh4U+A1NV+tX9EQN4q8gb4WcOAET7bOdIk6A7ZRO/VLP+CH3f7s7b32G7Fj0N9c3nQVKNjsjaEeyrXx6lGDe+x2Se2J7Yr3kSl7MEKTV8POUy6GS1CRi9XKveeCsH6j8EIAcWbizWci1rQq++q47HwjPwMTAeVjPB6OfsTw4vDIsxyN4YPHdtN64n2XMr2WUX+5jFrOdgOxGuzDmvtHRTDCpugAj0cB2fbpFPCoqLsf31il4g9Kpvb3QNR5zYR3zgn6I0+X8IuhcQeH0/b4gVw19/qlwmH1w9Q+coG2i6PKy52Q8Z3MDTaKxHfepOOuHmhS8iWxv4USGecp5MJ25ntO7fLX4iOMwEHFsxg4wa52kmW8+BWGC/ipvZVRrdtJlbdOBQckFzHhE9nAzFYoQGZD9MW9uHRPuSKyYsPgcur44ZvfRnCx1FRtyZLT0KNSRcCo86WOFVvSqCf7rQFR/huMBjNtCk7nlv4zSzOpIFOKRxjvG5pm3D7U9pc5/KiDntY2L2K2X4vMErZFLDnp+Y0J3U60fbWscs7ttbkYzGf7wN8Ss1cSgHyxpYqTwrJsJvNzR8QYz/Zh+1CqSV/p9cxYUOYOgplZyqSY13jMjJr7CTozydxVK5KrZrDN9oHcZgqH7oyJLDfl4xZMKztNILqUCdhN7Ip8g9XBeAxrbWhWke2V92IJNQT78rqaolallzT3HpyiZ6WNZ1VA/ZYN/WasCYASdkf/aqKlrBUjbNQTEbKt+B/wOWzwmTwVH2vxf0XHGngSNrBsagnWVZG5avwH)
2. Drag the "Number of points" slider to 100 (that's still working)
3. Drag the "Number of polygons" slider to 200

Before fix: It will (be horribly slow to add these polygons, and) eventually crash with a WebGL error.
After fix: Works fine

### New `quality` API parameter
- Added an optional `quality` parameter to `ClippingPolygonCollection` that controls the resolution of the signed distance texture (base size = `4096 * quality`, clamped to `[128, maxTextureSize]`).
- Allows users to trade visual fidelity for performance when high-resolution clipping isn't needed.

### Polygon packing improvements
- **Per-polygon extents in the texture:** Each polygon now packs its own individual spherical extents (south, west, latRange, lonRange) into the polygon texture alongside vertex data. This gives the shader per-polygon bounding information for early rejection.
- **Sorted polygon packing:** Polygons are packed into the texture sorted by their shared region/extents index, enabling the shader to skip and early-exit on region boundaries.
- Updated pixel budget: each polygon now occupies `3 + vertexCount` pixels (previously `1 + vertexCount`) to accommodate the per-polygon extent header.

### `getExtents` algorithm rewrite
- Replaced the previous O(n²·g) merge-by-restart algorithm with a cleaner O(n²) two-pass approach:
  1. Find the first overlapping group for each polygon.
  2. Absorb additional transitive overlaps in a second pass.
- Extracted `computePaddedExtents` into its own function for clarity and reuse.
- Spherical extents are now cached once per update via `polygonExtentsCache` instead of being recomputed multiple times during the merge loop.

### Shader optimizations (`PolygonSignedDistanceFS.glsl`)
- **Early exit for empty regions:** Fragments outside all region bounds return immediately.
- **Sorted-polygon skip/break:** Because polygons are packed in region order, the shader can `continue` past polygons belonging to earlier regions and `break` once it reaches a later region.
- **Per-polygon bounding box culling:** Before computing the signed distance for a polygon, the shader checks whether the fragment falls within the polygon's individual padded bounding box, skipping expensive edge-distance math for irrelevant polygons.
- **Halved position texture reads:** Cached the previous vertex across loop iterations so each edge reads only one new position instead of two.

| Before | After |
|----------|----------|
|image|image|

### Debug visualization
- Added a `debugShowDistanceTexture` option that renders the signed distance texture as a fullscreen overlay (red = outside, green = inside) for development and debugging.

### Sandcastle demo
- Added a new `clipping-performance-dev` Sandcastle example for profiling and visually testing clipping polygon performance, including Google 3D Tiles integration and adjustable quality/scale settings.

## Performance Testing Results

Compared delays for adding a ClippingPolygonCollection with ClippingPolygons (n) each with Vertices (m)

Performance Test Demo: [Sandcastle](https://sandcastle.cesium.com/index.html#c=vVj/c9u2Dv9XON92k1dXTtJkWxUn916TNq/v2iVrsu2HuHelJdrmhRY1kkqq9fy/D/wmUbKcZZdrcltNASAIfAgCIOmq4EKhHxCW6IRIWq7QXPAVmg5S8zUdHE7FNKdW7BLnWYqlYsRLNRQnqf9PeS4VuqXkjgh0hHJy53THvxta5LWf8FxhmhMxHYzQFz0VIUVXhAEtQXPMJBlZKgYTsKI8b5NlSnLynmfkgqY3RLSZMyzJO1wR0cdcEJ7CPKA6097y/MzSLgS/pcC6qgoSn52fn717bSath7WH1rfYLB9nZFYuLpf87o3AKyIviLgkAEEGritRknoSIwrdccGyK8qIJMowlKi85yEP5uI7TJW3LhUEK3LG+YKRiyVXHL4ZlYqmL07dlMipQYjnrPpN0nzxB1VLO+esdleb5DBwDrVXjiV4AssbsBy/5W4h6IoqektkjLMsCudafWuUYpUuUUSE4GLo3dNRwRmJGV9En15rFmIcZ2AmanuEXpwioxBiwZqEvv1idK0/uRU8pPMyT3VUIIvPCRUp4MNprmQExioi3vF8hNwQqxESsGIpRygvV1auZZ9ChSGC/9cfnfNzMDTSe0eBunMIP5NmNnw+e1ar8EowSFI0DsTafAYWH6HaQPQMvcdqGcOWRRjOovm4eAujvSH8Y03uqsCqUQFjpyLl8gEqrJNxUcpldM00QqDuo4+Gtf0RRJUid7L3wn5FpDrFCkdFH6LKccHcGiYFRyuB9PEGppeCnHDGiNEIicDL5HCYtIyeH5BTIZNGUaNKiwdi4KTgBRGKEi1fawOXEr5Ik4zME63q/Owk2Y1fJCcfLn/enw7QutbQjObWStBz3agPTNjwp2WHyTUrAgc9aU9qpr0vmaIXnFWLFgL2L+VcwCkBnLUB1xbjjx9bUuvgqxl7mfVha0f9hrT31G5WWgoBMaXZl7wUkKabo6DPwAb/cokLUrNNKmwlD02VWkZ20yGWVZ6iOpTKItMnmNGigJTgwJCRtdwcJEsZBRSNhK8EBU5hXl0XMPNZ7s8S8oqqzFdvMvJwffrNmIDuIGuGKx59+yX4WkMtyprlPVOPLctZAgw3smRnBpDdaP3JGdiUlf5ks7knMSP5Qi27ycel6SwQFGTFb0m0qeKads/71mWMJeGJTjub9F8hcNXESSi6EDS7pH8R4JqUNGcQzZHNdn8KFQXADof/lHCtXH/KXXxuL6HTr1982Mmdi8rk5+9qgQ6/rgE9VaU5Z89/Ooh3D/b2d36G5AvL/9CJQf334mX8cv/Hlz8daJGqV2Qn3vX5uuB3kY1cRIeBSDfWfcg0Fjdb3m0coPD/X/K82dVYF93Ajf4U3rMSnaPInuRhmMd6gk63Bc13o2FdG70ZaqYWdSdtlTdZR29QX1zHQKE678e3mJXwM6d56HFk+NUQHR17lKAaQMuZOQ5shAm14Va8Cy6pzloQJAGSvVZ6ZfGSEoFFuqziBVG/a8uiYVwrahboO1wWniAimpa6kzGjVomp1SfNsFU46GpVKjxjpNUYms3a9H59XxJoN/odq5oC39jncAHTel2uV6/b61PoDnGeQqx+1kU2sZxazqXVpJXwgyZXx2+rkV0wPiOm3d3MoaFEj6ddkgcIEah7ZqWNljpYo8X7F9pb9dqe8i1Vc2+Edn8cQW7Z24UR/BfvDDvXsxQaIoHP53N72wj27n/EdOUXFJr4DzhfEBd4jm1SleIfQAjDUjugenSfwPP9g0bkYAf+4h1bjA+Dq9RfnK+ueNR/hkYta4c9F019B2RNhxmUiwTtufWCRJoAPDXV5eTEwhU0EInGrd1FhCSbDIODs76/tzE2voF7s74FR50uObAY3PilXM3gnlx7Fod18nBjnitX/bNMPm/NcQ73zXCsjrwGo1daM9qyDqc+acfy8vcGcNjwBfWvrp+uw0MtnfXZcIF4k/P0hpcqVgKnN40lVtrdTjhnM6yfKTKeliuIPZ2eXzOih6+qt1A4Bk5mOrATu9pxUbDqFdQYMEw2q4y87iZeuzNhpfOZJOJWZ+Bw5nRggwuWjGU5k6mgMxJFNo9A2foSRiDYrhkO1t5YO+y8XzTPNrpYX1k7X5VKQYKeDlwf3N0U/UjTo3271sWCkVrpqbkadHK4VmlfZVCULkl6Q7LQP3+dcKx/46HuH+dUSNW9dozHyFqGZkTdEZIjk+XhVgRVh+amV8eGhpl+iXBvEFtRe0/yMnL3wuZ+C+7pi92VVRrc6QBHokthgqLQ0y2lx73F1A5YwX96rvHd2jcGgGH71rkdPVf3m6HHr63b3y/974bT/vnmsV53Xepzu4PMU3nt7k6D0WAiVcXIsWb9x72SloJFcTxWZFUwfWsfz0qIXqj0UufhfDL2UyYZvUU0O+p5DoUeAEsJnHnJmL6iTAfHkzHIt6a597PzWyIYrrTIcvf4nSXGcTwZw+fmrDqdGasnpv07tl5O1Ixn1bGHYqLEcYPhRGXHNqkjPq/bt8kYyG2h8PViQvOiVCHFPnuAHUL3FtNBm7eiObB2N8j4sybr7qHLkooUfVP0ZeL5DNIyMM1FIGk9IyBDs6kuQd8bM79v6xj3eOKN18EO2yRha+DrAIb3rwcCoboWavAhHoi5roJfBfH9r4i4Ld1Phbde7RFoX9oO46uAvNMP8sEWgKFRfgjGdU/0BAi7tR6Dr27cnjJp7D0SXdtoPgW2eqVHIPur7YWfMnK3Y7t38ABs/U39KdB1az0mcnXLYe6jD0vB3jzTvc74545drqlNkG/1H2QYDOsSDWNXu32R/xs)

Test Results compares:

* Pre PR - Baseline (State before Optimizations)
* Post PR - Quality = 1.0 which corresponds to a 4096x4096 internal signed-distance-field texture. (Same as Baseline)
* Post PR - Quality = 0.5 which corresponds to a 2048x2048 internal signed-distance-field texture.
* Post PR - Quality = 0.25 which correspond to 1024x1024 internal signed-distance-field texture.

Specs:

* Machine: ThinkPad P1 Gen 7
* GPU: NVIDIA RTX 3000 Ada generation laptop GPU

| Polygons (n) | Vertices (m) | Pre PR | Post PR (q=1.0) | Post PR (q=0.5) | Post PR (q=0.25) |
|----------|----------|------------|-------------|-------------|-------------|
| 2 | 256 | Noticeable delay | Instant | Instant | Instant |
| 2 | 1,024 | ~1s delay | Instant | Instant | Instant |
| 2 | 4,096 | ~2s delay | Instant | Instant | Instant |
| 2 | 8,192 | Freeze | Instant | Instant | Instant |
| 2 | 16,384 | Freeze | Noticable delay | Instant | Instant |
| 64 | 16 | Instant | Instant | Instant | Instant |
| 64 | 32 | Noticable delay | Instant | Instant | Instant |
| 64 | 128 | ~3s delay | Instant | Instant | Instant |
| 64 | 256 | Freeze | Instant | Instant | Instant |
| 64 | 512 | Freeze | Instant | Instant | Instant |
| 64 | 1,024 | Freeze | Noticable delay | Instant | Instant |
| 64 | 2,048 | Freeze | ~1s delay | Instant | Instant |
| 64 | 4,096 | Freeze | ~2s delay | Noticable delay | Instant |
| 64 | 8,192 | Freeze | ~3s delay | Noticable delay | Noticable delay |
| 128 | 8 | Instant | Instant | Instant | Instant |
| 512 | 8 | ~1s delay | Instant | Instant | Instant |
| 1,024 | 8 | ~3s delay | Instant | Instant | Instant |
| 2,048 | 8 | Freeze | ~1s delay | Noticable delay | Instant |
| 4,096 | 8 | Freeze | ~3s delay | Noticable delay | Instant |

## Quality Testing Results

Quality Test Demo: [Sandcastle](https://sandcastle.cesium.com/#c=nVhrb9s2F/4rhDFg8uDQsi625TrB0tTrsnlN13j7sNcvVlqibS6yqJGUU6/If9+hRF2TdEE/JBEPz43PuZFhh5QLhb5DRKIrKll2QFvBD2jdC/PVuvdqLdYJK9huSRKFRKqYllw1xXDqn5AnUqEjo/dUoHOU0HujG/+e06xS+xVPFGEJFeveAH3WoggpdqAx0GZoS2JJBwWVgAtEMZ60yTKkCf2FR/Q9C++oaG9uiKRLcqLiqc0d5SHIAdW4ds2TtwXtveBHBlurU0rx25ubt8uFERoO0WpP0S7mG4oiTiVKuILz0QgpjjSNyTQGk1EtIFkSUqRA7P2eKy4oiZlULETuG7RiMegAhjiLgIcKAWhoyYd+BSaoWCRkA5ALmoBXLNnl2uTdSW8XKOMcCAy0S3XgMt1TAas9vwf4lcjoE6HRqMWwbWD/OwO/1GmGRtjOnX/oyISZAAfUCsIDUga0n7KYkeQNURTrhLiWfDq2R1ahct1zbMc+s0dndrBy3Jltz9wA26OpMx4F8Mu2g/HI9cZ/QPS1RHFmc6Aw5uEdbhttrJrwLDmJEEEQvZ9ub96hLYCK7pnao5RLpnNGoohuWVIilwr+Fw0V2nKuUsBTaT0xbRBy7UqcSnAKBASVPBOh9oTcE6YamfPBbOUoXEpJ1XVkOb7rjtxRcaxSSUQUuX1SjfZfcg2mYcAxHMwqrVYFAppickhX/K3gWRLN8gCbfGvkDTIFiGuLEpMosup1k7k6O7hVc2CAGyAEySOJM/gDMEaWlVNPfXR+UbpkzpDjTCPDgFMen3Y86RvvSiQqWyUD5pnSNQ+28xp9xGdSud6syusPDn0Iik/7jCBn8iYxgH4RIYhCHu89h+zRYDTjEJIDFeRmu9VcrR71IyURpMp7psL9B5LsqNU55C9E7bHiH4CNJNIKfGyXJ3yO52zSYpratqmyGhQTLe3s60wpAKWqUhzywwFOhDd0Cw1k8YmGGVQcBHNxBKSX0E+g+gXERYekzhOKQwLNJ242AU03pv4B6FbcqlAetEAp/TIp9QCbAAiyoElx0W9WBo91qu6sjwu9hXTW6lKDDvsX5DNG33zOZR4+Gk1l/LIUgkavYpZarXZ3GUE71Z0RIsO3ulyhkKHFlnUbAcS7pCzaTcZibW9VSHRLt7PdLbrij/vGbBddjCetEna8ToyKZguYHaA2jqasOoa+FrXicOX5ccH/FIIde0+XSD2etY8rvtvFJrtgBt9qiQpdgyoM4mJMom2WhLqYkBXuKQzQqP8MpqVpw/aqM8D+24e6Eb/E+qOu8DV2ddYhRcQOUqLsGtp63ku/YLyVAfk9ACYVS1PdMIpeJqFn6nEdfaVn18mRCkmRVvsiPF7iEjNKn3WpUt8sytKCqUmCSr0wCneaWd+wIgQfus3qDG5UaWu6ll23nsjnT0yBPYPuI8L9CUNYftcDx+rjSqQ1RrsnbLfwq/buFY9jmh/PqnqjsSln6H8lCX1BRS1YCBufZvXnoGZ4qDv9/6uv6oJVt3VDas/uFwVUB7JDavdWc4q7BG5RMFyxEiS8syrThakCSsV5vCH6oh7xMDvAQNHwL2KqP1+foAuue4Zn3SsEu9pJmsan13A1AH9kbWVQ6q7zrCsJlm42koqjLpmm5Lpn0AGbWGYbGQq2oZala7455ZqttIT6oxE9/+bzI7AfPnbm7qMp1KiK+rZcX5ShTyqWmLdIM2GIUPBFErfy48y3Pdf2Xew4nutPHG+AHH/s+5MAO+7Edr2JPx2gMydwPcd2se3BbXjku/bYeGg85YJBJHKLaFZN94iJIqfRc27YeDL2XW86HoNaL/ADB4zZ2Js4zngcjP0RWPKn3mSAgOhOHPDI8YJR4NvToMrgLH1WP+hyPc93nMnE9kFoHPh+bmE68bxJYNug0LfHo2lO1D74LvDZztQJvKBvst4YijJhMC0vRpTA02n3g2lM1Ttt0SLj5fW7xeWHP9/dvFs0ni2tEipuNXgbn+C2o3fqAOsLRM4E8W3KFEqgKPbQquPuK/Y2FJQmtykJaX4B+7FgskpzICkLG0Yc69tEkmbqsuhCVbflyZJuFSQfFOeBH/OS6zcu+o3kbp2l7Dr9V//JWuXJC3iztLr1DZ47bP4qXi5+WP15tby++rl+uPUGvblUp5heaAXfm38ZZCK2MB4qeoCHMWTOcJPB/FE4lBqiZD4sReYROyIWnT/xvwH94JESdrZZHN+yf+i6dzEfAn9LzNygbmDSwRNcs+xHF8uCiDGeD2H5WKrqbLnXc6X70EUB01xteHS6KLv6XImLegTMVXTxa9FS5kP4bu/UK1gzHfkmBSEFIIJxoZ8X615778AS2LIfkcknIDtdMlz9U82OHb+7pZ9EZxvoyrCfP99m5RAaoHz9W976Zujb3MNv2/LDJw5R+q3oJ7irIQmhgBUY/oIt2GyqaoEFixJT+KzQhm8ThjJe/wI)

Because of better texture utilization the default texture size (4096x4096) now gives a better result. We could consider changing the default to 2048x2048 to achieve same quality and further improve performance.

| Baseline (4096) | Quality 1.0 (4096) | Quality 0.5 (2048) | Quality 0.25 (1024) |
|----------|----------|----------|----------|
|image|image|image|image|

## Issue number and link

https://github.com/CesiumGS/cesium/issues/13306

## Test updates

- Updated existing packing tests to account for the new per-polygon extent fields and shifted array indices.
- Added new test cases:
  - Identical/overlapping polygons are correctly merged into a single extent group.
  - Non-overlapping polygons produce separate extent groups with correct indices.
  - `computePaddedExtents` padding and clamping behavior.
  - Tests for the added `quality` parameter

## Breaking changes

None. The new `quality` parameter defaults to `1.0`, preserving existing behavior. The texture packing format change is internal and does not affect the public API. We might see better quality for clipping with default settings because of better texture utilization.

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/ClippingPolygonCollection.js
+++ b/packages/engine/Source/Scene/ClippingPolygonCollection.js
@@ -555,7 +555,7 @@ function packPolygonsAsFloats(clippingPolygonCollection) {
 
   /**
    * Pre-calculate all polygon spherical extents as it an expensive operation
-   * @type {readonly Rectangle[]}
+   * @type {ReadonlyArray<Rectangle>}
    * */
   const polygonExtentsCache = polygons.map((polygon) =>
     polygon.computeSphericalExtents(),


packages/engine/Source/Scene/ClippingPolygonCollection.js
```diff
@@ -362,91 +375,109 @@ ClippingPolygonCollection.prototype.remove = function (polygon) {
...
-  const polygonIndicesList = [];
+// into a single encompassing extent.
+//
+// For each polygon we scan existing groups for a first overlap (O(g)),
+// then on each subsequent overlap we absorb the group and restart the
+// inner scan. Each group can be absorbed at most once per polygon, and
+// each restart reduces the group count by one, so the absorb-loop does
+// at most O(g) restarts per polygon. Overall: O(n * g) where g ≤ n,
+// giving O(n²) worst case when all polygons overlap transitively, but
+// typically much better when groups are few and disjoint.
```
mzschwartz5: I think these docs could use a preface line indicating what the goal of this function is. Even better, could add some jsdoc saying what the inputs and outputs are. (After reviewing, I'm now reasonably sure I understand the outputs and purpose of the function, but I still don't know where `polygonExtentsCache` is coming from without further investigation).

mzschwartz5: It also wouldn't hurt to be explicit about what `g` and `n` are, and what you mean by `restart` and `absorb`.

packages/engine/Source/Scene/ClippingPolygonCollection.js
```diff
@@ -362,91 +375,109 @@ ClippingPolygonCollection.prototype.remove = function (polygon) {
...
+          defined(Rectangle.simpleIntersection(groups[g].extent, target.extent))
+        ) {
+          target.polygonIndices.push(...groups[g].polygonIndices);
+          Rectangle.union(target.extent, groups[g].extent, target.extent);
+          groups.splice(g, 1);
+          if (g < targetIdx) {
+            targetIdx--;
+          }
+          g = -1; // restart (loop increment brings it to 0)
+        }
```
mzschwartz5, line 520: At first I wasn't sure that there was a configuration that ever triggered a restart, but I get it now (if the extents were the exact bounds of the polygons, there would be no restart needed. But with bounding boxes, you do need it).

I asked AI if you can beat the algorithmic time complexity here and it came up with this (looks like same worst case performance, maybe better average case. _Probably_ not worth pursuing but I do notice some decent lag even with a small number of vertices per polygon when lots of polygons are overlapping).

Image

mzschwartz5: Side question - does this extent merging work around the antimeridian? Have you done any testing there?

packages/engine/Source/Scene/ClippingPolygonCollection.js
```diff
@@ -889,8 +998,10 @@ ClippingPolygonCollection.getClippingDistanceTextureResolution = function (
     return result;
   }
 
-  result.x = Math.min(ContextLimits.maximumTextureSize, 4096);
-  result.y = Math.min(ContextLimits.maximumTextureSize, 4096);
+  const quality = clippingPolygonCollection.quality;
+  const baseSize = Math.max(128, Math.ceil(4096 * quality));
+  result.x = Math.min(ContextLimits.maximumTextureSize, baseSize);
+  result.y = Math.min(ContextLimits.maximumTextureSize, baseSize);
```
mzschwartz5, line 1058: What happens if the resulting dimensions are non-integer? Does it still work?

packages/engine/Source/Shaders/PolygonSignedDistanceFS.glsl
```diff
@@ -52,48 +52,79 @@ void main() {
...
+        vec2 extentsRange = getPolygonPosition(lastPolygonIndex + 1);
+        vec4 polygonExtent = vec4(extentsSouthWest, extentsRange);
+        lastPolygonIndex += 2;
+
+        if (polygonExtentsIndex < regionIndex) {
+            lastPolygonIndex += positionsLength;
+            continue; // skip to next (TODO: could optimize further if we knew how many polygons to skip)
+        } else if (polygonExtentsIndex > regionIndex) {
+            break; // done (we know polygons are sorted by regionIndex)
+        }
```
mzschwartz5, line 76: I'm really curious how this actually impacts performance. Generally speaking, early returns in shaders are not as helpful as you might think, because GPU threads operate in groups and often have to just hang anyway while threads that didn't early-return finish their work.

If LOTS of threads are exiting early, causing whole thread groups to finish early, then it could help. But the only way to know for sure is to profile. I know you did performance testing for all the changes together, but what about just this part? If it doesn't make a huge impact, we might be better off without the extra complexity and memory consumption of storing individual extents.

mzschwartz5: I suspect this change you made:

> Halved position texture reads

Accounts for the biggest shader-performance savings by far. Memory reads on the GPU are very expensive; compute pales in comparison and is generally considered "free" (barring things like divergent logic flow, which can be bottlenecks).

mzschwartz5: Same comments apply to this optimization as well:

> Per-polygon bounding box culling

I haven't reviewed the shader changes super closely yet; curious to know first if these other changes actually make a big difference relative to halving the texture reads.

packages/engine/Source/Scene/ClippingPolygonCollection.js
```diff
@@ -471,15 +502,49 @@ function packPolygonsAsFloats(clippingPolygonCollection) {
   const extentsFloat32View = clippingPolygonCollection._extentsFloat32View;
   const polygons = clippingPolygonCollection._polygons;
 
-  const { extentsList, extentsIndexByPolygon } = getExtents(polygons);
+  // Pre-calculate all polygon spherical extents as it an expensive operation
+  const polygonExtentsCache = Object.freeze(
+    polygons.map((polygon) => polygon.computeSphericalExtents()),
+  );
```
mzschwartz5: I don't see `Object.freeze` used a lot in PRs, and thought there might be some anti-pattern around it, so I asked the team and got this feedback:

> If we simply "do not intend" for something to be modified, but it's not a case where modifications would be easy and disastrous to do by accident (like Color.RED) ...then I'd be inclined to put /** @type {readonly SomeType[]} */ on it (assuming a ts-checked file...) and let this be enforced at compile-time without adding runtime logic.

and

> Just remember that .freeze is not "deep" and the items in the array can still be manipulated

I think it makes sense to use type checking instead, as suggested.

packages/sandcastle/gallery/clipping-performance-dev/main.js
mzschwartz5, line 1: This is great - very helpful for development. My only nit is that the inputs for # vertices and # polygons and such have _very_ large ranges, making it pretty easy to accidentally brick your device (even with the performance updates). I'd make the maximum range of the sliders a lot smaller, still allowing users to manually input higher values to the input fields.

packages/engine/Source/Scene/ClippingPolygonCollection.js
alarkbentley: > At first I wasn't sure that there was a configuration that ever triggered a restart

Right, this case is not trivial to see and I missed it myself at the start. I have added a diagram to the comment that clarifies when this happens.

> I asked AI if you can beat the algorithmic time complexity

Cool! But I think n^2 is good enough for now as the bottleneck would most likely be on GPU side anyhow. We can return to this some day.

> Side question - does this extent merging work around the antimeridian

Good observation, I actually don't think it does (and has never done) as we are using Recangle.simpleIntersection instead of Rectangle.intersection. I could try to set up a case, guessing this is a very unlikely real world case though.

packages/engine/Source/Scene/ClippingPolygonCollection.js
alarkbentley: Added more details to this comment

packages/engine/Source/Scene/ClippingPolygonCollection.js
alarkbentley: Non-integer? Shouldn't be possible because of Math.ceil().
Do you mean non pow-2? That works.

packages/engine/Source/Shaders/PolygonSignedDistanceFS.glsl
alarkbentley: So unfortunately I don't have performance data after each individual change but I did manually test a lot during development. And if i recall correctly the early exit actually did a huge impact especially when we had polygons far apart (creating multiple regions in the atlas).

The "Halved position texture reads" did surprisingly little in comparison actually, I'm guessing the GPU optimized this on the fly.

packages/engine/Source/Scene/ClippingPolygonCollection.js
alarkbentley: Yea freeze I only added for documentational purposes, its intended to be immutable.
Changed to @type

packages/sandcastle/gallery/clipping-performance-dev/main.js
alarkbentley: I reduced the max to polygons=1000, vertices=1000

packages/engine/Source/Scene/ClippingPolygonCollection.js
mzschwartz5: Oh I just missed the Math.ceil operation - never mind.

packages/engine/Source/Scene/ClippingPolygonCollection.js
donmccurdy: Small correction to my earlier comment (I had suggested `readonly SomeType[]` above). While either syntax should be OK for type checking...

- (a) `@type {readonly Rectangle[]}`
- (b) `@type {ReadonlyArray<Rectangle>}`

... our dependency on [tsd-jsdoc](https://github.com/englercj/tsd-jsdoc) prevents us from using (a) until we've addressed either #10455 or #13211. In the meantime (a) will cause `npm run build-ts` to fail (visible on this PR) and we'll need to use (b). Sorry for the churn! :)

Also note that this file is not type checked in CI yet. It can be [opted-in](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/CodingGuide/README.md#type-checking) but that may add a lot of JSDoc noise to the PR, so no pressure at all to enable type checking right now.

packages/engine/Source/Shaders/PolygonSignedDistanceFS.glsl
mzschwartz5: Looking into this a bit more- does this compute command (`PolygonSignedDistanceFS.glsl`) basically just run once when the collection is added (or on a rebuild)? That is- it's not a per-frame cost?

If so, does it even need optimization? How slow must it have been before to make a noticeable impact when it only runs once??

mzschwartz5: I added some code locally to time the compute pass (using the WebGL2 extension `EXT_disjoint_timer_query_webgl2`), and played around with removing different optimizations from the shader. They all make a pretty big impact. With 982 polygons and 302 verts each (random numbers I used), the shader takes around ~300ms to run with this PR's optimizations. The early returns save 150-300ms, and the cached texture reads save ~500-600ms. 

Of course, this could be made more robust by making a matrix of measurements similar to yours above, but it's enough to convince me that this is all helping a lot.

Attaching a git patch file that can be applied to the PR (on top of commit `1b011a`) if we want to use it again later. (It modifies some compute engine code to add timing, and modifies the sandcastle to force the clipping collection to rebuild / force rerenders).

[compute-timing.txt](https://github.com/user-attachments/files/26249641/compute-timing.txt)

packages/engine/Source/Scene/ClippingPolygonCollection.js
mzschwartz5: Fair- we have plenty of existing problems with antimeridians. I don't think it's super unlikely to occur but we don't have to solve that here.

packages/engine/Source/Shaders/PolygonSignedDistanceFS.glsl
alarkbentley: > If so, does it even need optimization? How slow must it have been before to make a noticeable impact when it only runs once??

As you probably already figured out, yes this shader runs once on rebuild. Before these optimizations this single pass could take 10s of seconds or even crash the browser. We have 4096x4096 pixels that all needed to loop every single polygon and every single vertex so we are talking 100 billions of fragment operations.

tbenazzi:
Huge thank you to everyone involved in getting this merged — you're all absolute GOATs. The effort, the patience, and the quality of the work across contributors is genuinely impressive. This is what makes open source worth it.

That said, I'd like to reopen the broader conversation, because as we summarized in #12649, this PR addresses one of the three core topics we identified as blockers for a production-viable clipping solution in Cesium. Two remain open:

1. ✅ **Performance / geometry caching** — addressed by this PR
2. ⚠️ **Precision** — clipping accuracy errors of several meters between the input polygon and actual geometry clipping
3. ⚠️ **Shadow rendering** — shadows still cast from hidden/clipped geometry, making shadow studies unusable

Without these two, we can't realistically ship a production workflow for AEC use cases.
Currently we would love to have this kind of sharpness on our AEC design integration, without clipping the raw data before tiling it. 

Clean Clip

---

On the precision topic specifically, I'd like to raise a deeper question: is a **rasterization-based approach** fundamentally the right one here?

Cesium handles double-precision floats to maintain accuracy at very fine scales — that's one of its core strengths. But with a rasterization-based clipping approach, we're inevitably stuck with the staircase/aliasing artifact at polygon edges, regardless of how much we tune it. It feels like a mismatch between the engine's precision capabilities and what the clipping layer actually delivers visually.

I don't have deep enough low-level graphics knowledge to know if a **vector-based clipping approach** is feasible in this context — operating directly on geometry topology rather than rasterizing the clip region. If anyone has insight on whether that's a tractable path (or why it isn't), I'd genuinely love to understand the constraints.

Either way, thank you again. This community is exceptional.

PR #13313, opened 2026-03-17
BufferPrimitiveCollection: Material API

donmccurdy:
# Description

Material API for BufferPoint, BufferPolyline, and BufferPolygon collections.

As [discussed recently](https://github.com/CesiumGS/cesium/pull/13213#issuecomment-4005680384), putting style-related properties directly on BufferPrimitive objects doesn't scale well to supporting multiple renderers and rendering styles, so I'm splitting these properties off into new, packable material classes.

Why new material classes? Unlike the BufferPrimitive classes — where breaking from existing APIs was really necessary for our performance goals — I don't think there's any fundamental reason that BufferPrimitiveMaterial "needs" to be distinct from existing material types like `PolylineOutlineMaterial` and `PolylineDashMaterial`. But (1) working out how to share existing materials is non-trivial, (2) existing materials don't appear to cover points and polygons, and (3) our timeline is somewhat tight. So I think that designing a "common" material system to be shared by existing and Buffer- vector collections can be left as a future consideration.

### Features

| topology | color | outlineColor | outlineWidth | size | width |
|----------|-------|--------------|--------------|-----------|-------|
| point    | ✓     | ✓            | ✓            | ✓         |       |
| line     | ✓     |              |              |           | ✓     |
| polygon  | ✓     |              |              |           |       |

My intention would be to support color / outlineColor / outlineWidth for all three topologies "soon", but this seems like enough to start testing. Transparency and polygon offset will also be useful. Dashed outlines and arrows might be good additions later.

In a small change from earlier BufferPoint iterations, I've renamed 'pixelSize' to 'size'. This is meant to better match the 'pointSize' naming in Cesium3DTileStyle, and the `line.width` property which doesn't include a 'pixel' prefix, at the cost of no longer matching the naming convention of PointGraphics.

### Example

```javascript
import {BufferPoint, BufferPointCollection, BufferPointMaterial, Color} from 'cesium';

const material = new BufferPointMaterial({
  size: 8,
  color: Color.WHITE,
  outlineColor: Color.CYAN,
  outlineWidth: 2,
});

const collection = new BufferPointCollection({...});

// (a) apply material at initialization
collection.add({position, material}, point);

// (b) apply material as update
point.setMaterial(material);
```

Note that after calling `setMaterial(...)`, no reference to the BufferPrimitiveMaterial instance is held, and any subsequent changes to the material will have no effect until `setMaterial(...)` is called again. The collection simply updates the relevant buffers on each call.

### Preview

[material-api.webm](https://github.com/user-attachments/assets/e8555587-0682-447e-b678-0dae6bb7bf67)

## Issue number and link

- #13156 

## Testing plan

See attached unit tests.

<details>
<summary>Sandcastle example</summary>

```javascript

const collection = new Cesium.BufferPolylineCollection({
  primitiveCountMax: 1024,
  vertexCountMax: 1024,
});

const degreesArray = [-105.0, 40.0, -100.0, 38.0, -105.0, 35.0, -100.0, 32.0];

const positions = Cesium.Cartesian3.packArray(
  Cesium.Cartesian3.fromDegreesArray(degreesArray),
  new Float64Array((degreesArray.length * 3) / 2),
);

const material = new Cesium.BufferPolylineMaterial({
  color: Cesium.Color.GOLDENROD,
  width: 8,
});

collection.add({ positions, material }, polyline);

```
</details>

# Author checklist

#### PR Dependency Tree

* **PR #13313** 👈

### DIFF
--- a/Specs/Data/Cesium3DTiles/Style/style.json
+++ b/Specs/Data/Cesium3DTiles/Style/style.json
@@ -24,5 +24,6 @@
     "horizontalOrigin" : "0",
     "verticalOrigin" : "0",
     "labelHorizontalOrigin" : "0",
-    "labelVerticalOrigin" : "0"
+    "labelVerticalOrigin" : "0",
+    "lineWidth" : "10.0"
 }
--- a/packages/engine/Specs/Scene/Cesium3DTileStyleSpec.js
+++ b/packages/engine/Specs/Scene/Cesium3DTileStyleSpec.js
@@ -89,6 +89,7 @@ describe("Scene/Cesium3DTileStyle", function () {
         verticalOrigin: "0",
         labelHorizontalOrigin: "0",
         labelVerticalOrigin: "0",
+        lineWidth: "10.0",
       });
       expect(style.color).toEqual(new Expression("color('red')"));
       expect(style.show).toEqual(new Expression("${id} < 100.0"));
@@ -124,6 +125,7 @@ describe("Scene/Cesium3DTileStyle", function () {
       expect(style.verticalOrigin).toEqual(new Expression("0"));
       expect(style.labelHorizontalOrigin).toEqual(new Expression("0"));
       expect(style.labelVerticalOrigin).toEqual(new Expression("0"));
+      expect(style.lineWidth).toEqual(new Expression("10.0"));
     });
   });
 
@@ -160,6 +162,7 @@ describe("Scene/Cesium3DTileStyle", function () {
         verticalOrigin: "0",
         labelHorizontalOrigin: "0",
         labelVerticalOrigin: "0",
+        lineWidth: "10.0",
       });
       expect(style.color).toEqual(new Expression("color('red')"));
       expect(style.show).toEqual(new Expression("${id} < 100.0"));
@@ -195,6 +198,7 @@ describe("Scene/Cesium3DTileStyle", function () {
       expect(style.verticalOrigin).toEqual(new Expression("0"));
       expect(style.labelHorizontalOrigin).toEqual(new Expression("0"));
       expect(style.labelVerticalOrigin).toEqual(new Expression("0"));
+      expect(style.lineWidth).toEqual(new Expression("10.0"));
     });
   });
 


packages/engine/Source/Scene/BufferPrimitiveMaterial.js
```diff
@@ -0,0 +1,139 @@
...
+  /**
+   * @type {BufferPrimitiveMaterial}
+   * @ignore
+   */
+  static DEFAULT_MATERIAL;
+
+  /**
+   * @param {BufferPrimitiveMaterialOptions} [options]
+   */
+  constructor(options) {
```
danielzhong: I think `options.color` will throw TypeError when options is undefined. Maybe should we use:

```
constructor(options = Frozen.EMPTY_OBJECT) {
```

packages/engine/Source/Scene/BufferPolylineCollection.js
```diff
@@ -4,17 +4,17 @@ import defined from "../Core/defined.js";
...
 /** @import FrameState from "./FrameState.js" */
 
 /**
  * @typedef {object} BufferPolylineOptions
  * @property {Matrix4} [options.modelMatrix=Matrix4.IDENTITY] Transforms geometry from model to world coordinates.
  * @property {boolean} [show=true]
- * @property {Color} [color=Color.WHITE]
+ * @property {BufferPolylineMaterial} [material=BufferPolylineMaterial.DEFAULT_MATERIAL]
  * @property {TypedArray} [positions]
  * @property {number} [width=1]
```
danielzhong: `result.width = options.width ?? 1;` has already been removed, please remember update the docs accordingly.

Also, the JSDoc/examples for `BufferPointCollection`, `BufferPolylineCollection`, and `BufferPolygonCollection` still reference the old color initialization and `setColor()` workflow. Since this PR moves styling to material classes, those examples should we update to use `material / setMaterial()` instead?

packages/engine/Source/Scene/BufferPrimitiveMaterial.js
donmccurdy: Missed this one - thanks!

packages/engine/Source/Scene/BufferPolylineCollection.js
donmccurdy: Good catch – I think I've updated all the documentation now.

I'd like for the type checker to either catch errors like the unused width, or to extend the BufferPrimitiveMaterialOptions so we don't have to duplicate so much here, but I think that's currently blocked by #10455.

donmccurdy:
> It looks like BufferPrimitiveCollection.byteLength is not currently used anywhere in the render path, 3d tiles cache logic, or similar runtime systems?

That's correct, I believe `collection.byteLength` is unused other than unit tests now. I suspect it's helpful enough for testing and debugging that it's worth keeping, but could be talked out of it. :)

> **EDIT:** I'm also not familiar with the 3D tiles caching yet — if it could benefit from the `collection.byteLength` information that's definitely a factor. Note that resources linked to the render context aren't yet counted.

***

> Anyway, for this PR, should we still update BufferPrimitiveCollection.byteLength in BufferPrimitiveCollection.js to include this._materialView.byteLength?

Good call, I'd forgotten to add the new buffer to `collection.byteLength` yes. Updated in implementation and unit tests, thanks!

packages/engine/Source/Scene/Cesium3DTileStyle.js
```diff
@@ -1284,6 +1290,22 @@ Object.defineProperties(Cesium3DTileStyle.prototype, {
...
+   */
+  lineWidth: {
+    get: function () {
+      return this._lineWidth;
+    },
+    set: function (value) {
+      this._lineWidth = getExpression(this, value);
+      this._style.lineWidth = getJsonFromExpression(this._lineWidth);
+    },
+  },
```
donmccurdy, line 1307: I'm taking the liberty of adding a `lineWidth` property to Cesium3DTileStyle.js here, for alpha testing. That said... it isn't implemented in any _stable_ part of the 3D Tiles spec, including the older vector tiles format, so I've omitted it from published types and documentation with the `@ignore` JSDoc tag above.

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -139,15 +141,15 @@ function renderBufferPointCollection(collection, frameState, renderContext) {
...
+      showSizeAndColorArray[i * 3 + 1] = material.size;
+      showSizeAndColorArray[i * 3 + 2] = AttributeCompression.encodeRGB8(
+        material.color,
       );
 
-      outlineWidthAndOutlineColorArray[i * 2] = 0; // TODO: Material API.
+      outlineWidthAndOutlineColorArray[i * 2] = material.outlineWidth;
       outlineWidthAndOutlineColorArray[i * 2 + 1] =
-        AttributeCompression.encodeRGB8(Color.WHITE); // TODO: Material API.
+        AttributeCompression.encodeRGB8(material.outlineColor);
```
donmccurdy, line 152: I would prefer for the material to take care of packing itself into vertex attributes, rather than making the renderer know about that, but ... for now I think that gets into too many questions of how we'll bridge the draped / non-draped rendering code paths, and if/when we want to try to consolidate materials with existing vector data types in Cesium.js.

danielzhong:
It looks like there are some test errors that need to be fixed before merging. @donmccurdy

PR #13317, opened 2026-03-19
Embedding search tweaks

tomdicarlo:
# Description

Tweaked the embeddings search to reduce the amount of visual updates in the gallery when searching.
This includes an increase in the debounce time and now having the pagefind search wait for the completion of embedding to re-render.

# Author checklist

### DIFF
--- a/packages/sandcastle/src/Gallery/GalleryItemStore.ts
+++ b/packages/sandcastle/src/Gallery/GalleryItemStore.ts
@@ -110,46 +110,36 @@ export function useGalleryItemStore() {
           return data.filter(isFulfilled).map(({ value }) => value);
         });
 
-      if (
+      const doEmbedingSearch =
         embeddingSearchEnabled &&
         embeddingModelLoaded &&
         searchTerm !== null &&
-        searchTerm.trim() !== ""
-      ) {
-        const embeddingPromise = new Promise<void>((resolve) =>
-          setTimeout(resolve, 300),
-        ).then(async () => {
-          if (abortController.signal.aborted) {
-            return null;
-          }
-          return vectorSearch(searchTerm!, 5, searchFilter);
-        });
+        searchTerm.trim() !== "";
 
-        const [pagefindValues, vectorResults] = await Promise.all([
-          pagefindPromise,
-          embeddingPromise,
-        ]);
+      const embeddingPromise = doEmbedingSearch
+        ? new Promise<void>((resolve) => setTimeout(resolve, 300)).then(
+            async () => {
+              if (abortController.signal.aborted) {
+                return null;
+              }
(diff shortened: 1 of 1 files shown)


packages/sandcastle/src/Gallery/GalleryItemStore.ts
```diff
@@ -98,52 +98,64 @@ export function useGalleryItemStore() {
...
-      startSearch(() => setSearchResults(values));
-    };
-
-    performSearch();
-
-    const embeddingTimeout = setTimeout(async () => {
-      if (abortController.signal.aborted) {
-        return;
-      }
+      const pagefindPromise = pagefind
```
jjspace, line 101: This is an optional change but might help with code clarity? Or maybe it's more confusing without the explicit `if` check/branch
```js
const performSearch = async () => {
  const pagefindPromise = pagefind
    .search(searchTerm as string, { filters: searchFilter ?? undefined })
    .then(({ results }: { results: PagefindSearchResult[] }) =>
      Promise.allSettled(results.map((result) => result.data())),
    )
    .then((data: PromiseSettledResult<PagefindSearchFragment>[]) => {
      const isFulfilled = <T>(
        input: PromiseSettledResult<T>,
      ): input is PromiseFulfilledResult<T> => input.status === "fulfilled";
      return data.filter(isFulfilled).map(({ value }) => value);
    });

  const doEmbedingSearch =
    embeddingSearchEnabled &&
    embeddingModelLoaded &&
    searchTerm !== null &&
    searchTerm.trim() !== "";

  const embeddingPromise = doEmbedingSearch
    ? new Promise<void>((resolve) => setTimeout(resolve, 300)).then(
        async () => {
          if (abortController.signal.aborted) {
            return null;
          }
          return vectorSearch(searchTerm!, 5, searchFilter);
        },
      )
    : undefined;

  const [pagefindValues, vectorResults] = await Promise.all([
    pagefindPromise,
    embeddingPromise,
  ]);

  if (abortController.signal.aborted) {
    return;
  }

  startSearch(() => {
    setSearchResults(pagefindValues);
    setVectorSearchResults(vectorResults ?? null);
  });
};
```

tomdicarlo: Oh I didn't think about allowing embeddingPromise to be null. I like that idea! Will push that change shortly

donmccurdy:
@tomdicarlo The changelog entry doesn't look quite right, as anything merged before the end of this month goes into the 1.140 release on 2026-04-01. ~~Could you move the new entry into the existing section for the 1.140 release?~~ Thanks!

> **EDIT:** Fixing in https://github.com/CesiumGS/cesium/pull/13327, had some other changelog mistakes to resolve anyway.

PR #13318, opened 2026-03-19
Types: Use ES6 classes and type checks for ModelComponents

donmccurdy:
# Description

ES6 class conversion and type checking for ModelComponents. As these are data structures used elsewhere, it seemed like an easy lift that would allow us to do type checking more easily on files that depend on these data types. I've also replaced `@private` tags (which prevent any type-checked file from accessing properties) with `@ignore`  — which hides properties from docs and published types, as I believe was the intention.

Note that (as with other ES6 conversion PRs) the diff on GitHub is vastly easier to read in split mode, with whitespace changes hidden:

https://github.com/CesiumGS/cesium/pull/13318/changes?diff=split&w=1

## Issue number and link

- #4434
- #13205

## Testing plan

No functional changes expected. I diffed Cesium.d.ts before/after the change, and the output was identical.

# Author checklist

#### PR Dependency Tree

* **PR #13318** 👈

### DIFF
--- a/packages/engine/Source/Scene/ModelComponents.js
+++ b/packages/engine/Source/Scene/ModelComponents.js
@@ -35,9 +35,6 @@ const ModelComponents = {};
 /**
  * Information about the quantized attribute.
  *
- * @alias ModelComponents.Quantization
- * @constructor
- *
  * @ignore
  */
 class Quantization {
@@ -129,9 +126,6 @@ class Quantization {
 
 /**
  * A per-vertex or per-instance attribute.
- *
- * @alias ModelComponents.Attribute
- * @constructor
  *
  * @ignore
  */
@@ -303,9 +297,6 @@ class Attribute {
 /**
  * Indices used to select vertices for rendering.
  *
- * @alias ModelComponents.Indices
- * @constructor
- *
  * @ignore
  */
 class Indices {
@@ -347,9 +338,6 @@ class Indices {
 /**
  * Maps per-vertex or per-instance feature IDs to a property table. Feature
  * IDs are stored in an accessor.
- *
- * @alias ModelComponents.FeatureIdAttribute
- * @constructor
  *
  * @ignore
  */
@@ -414,9 +402,6 @@ class FeatureIdAttribute {
  * Defines a range of implicitly-defined feature IDs, one for each vertex or
  * instance. Such feature IDs may optionally be associated with a property table
  * storing metadata
- *
- * @alias ModelComponents.FeatureIdImplicitRange
- * @constructor
  *
  * @ignore
  */
@@ -488,9 +473,6 @@ class FeatureIdImplicitRange {
 /**
  * A texture that contains per-texel feature IDs that index into a property table.
  *
- * @alias ModelComponents.FeatureIdTexture
- * @constructor
- *
  * @ignore
  */
 class FeatureIdTexture {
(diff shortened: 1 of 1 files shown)


javagl:
The changes look straightforward. 

(I'd have to check in the other discussions how these `/** @import  ` things work, but it's clear that this is ~"something to import that type information")

---

Is there any purpose behind these `ModelComponents.` prefixes? I know, that's not _changed_ by this PR. But it always bugged me a bit that this made things unbrowsable...

Cesium TS 0001

And I now did a blatant
Replace all `@type {ModelComponents.`
with `@type {`
which allows browsing in VSCode - and I ran `build-ts`, and the resulting `d.ts` was still identical.

Any reason to not remove them?

---

Unrelated to this PR: The type declarations are "all wrong".
```
    /**
     * An index buffer containing indices.
     *
     * @type {Buffer}
     * @ignore
     */
    this.buffer = undefined;
```
The type should be `Buffer|undefined`. Could  you remind me at which stage of the "ES6/class/doc" efforts something like this would be detected by some linting step? (Yes, it would probably cause thousands or tens of thousands of warnings, but ... that's the goal, actually...)

donmccurdy:
> Is there any purpose behind these `ModelComponents.` prefixes?  Any reason to not remove them?

I wasn't sure about that either — sounds OK to me to remove them, I just wish I knew the original intention! :)

> The type should be Buffer|undefined. Could you remind me at which stage of the "ES6/class/doc" efforts something like this would be detected by some linting step?

In TypeScript terminology, this is a ["strict mode"](https://www.typescriptlang.org/tsconfig/#strict) setting. If we enable strict mode it will start caring about nullable vs. non-nullable types, otherwise it won't. I briefly tried enabling that setting, but without certain TypeScript syntax (which would be unblocked by #13211) it was too noisy to deal with. I don't have a strong opinion yet about whether we should start with strict mode on, or consider it further into the migration.

javagl:
A git blame didn't reveal anything obvious - this `ModelComponents.` has always been there. 

Maybe back then, this
`@alias ModelComponents.Attribute`
(that was removed here) was somehow picked up in a way that made something work 'better'? 

Maybe it was somehow intended for a usage in _external_ files,  e.g. `GltfLoader` with things like
`@returns {ModelComponents.MetallicRoughness}`?
(This still doesn't work: The type cannot be resolved either way. I think it would work if the model components were in individual files, but... 🤭)

Maybe it was just for disambiguation? Names like `Node`, `Attribute`, or `Primitive` could certainly need some sort of namespacing (but apparently, at least from "clean" typing perspective, that wasn't accomplished with this prefix either...)

donmccurdy:
Thanks @javagl!

Out of curiosity, I tried removing the `@alias` tag from every file in `packages/engine`, and diffing the types:

```diff
diff --git a/Source/Cesium.main.d.ts b/Source/Cesium.d.ts
index 5e3f9c1afd..9b7190f8d3 100644
--- a/Source/Cesium.main.d.ts
+++ b/Source/Cesium.d.ts
@@ -9592,7 +9592,7 @@ export class MapProjection {
 /**
  * Math functions.
  */
-export namespace Math {
+export module "Math" {
     /**
      * 0.1
      */
```

One tiny change, no idea why. :)

I haven't been able to find a difference in the docs either, but I don't have a programmatic way to diff those, so there could be other changes hiding. I have no objection to the `@alias` tag but I do wish I knew what it was for, to test changes like these.

javagl:
I don't have "deep" knowledge about the purpose and effect of `@alias` in all the relevant toolchains. So it's not clear what _exactly_ caused that tiny difference - and `Math` is always special. **Iff** the `@alias`es turn out to be unnecessary in _general_ but are (only) required for some specific aspect of `Math`, then it could be kept at that single place. But maybe that will be investigated as part of the upcoming changes and cleanups.

PR #13321, opened 2026-03-20
Types: Use types in VectorGltf3DTileContent

donmccurdy:
# Description

Followup to #13271, adding types everywhere I could. Found a few minor mistakes in documentation in the process.

## Issue number and link

n/a

## Testing plan

PR should pass 'npm run tsc'. Sandcastle vector tilesets still work:

http://localhost:8080/Apps/Sandcastle2/index.html?id=3d-tiles-formats

This is an internal class, so published documentation and .d.ts files shouldn't include it. Here's a diff of Cesium.d.ts before/after:

```diff
diff --git a/Source/Cesium.main.d.ts b/Source/Cesium.d.ts
index fbbe832797..3c8499696d 100644
--- a/Source/Cesium.main.d.ts
+++ b/Source/Cesium.d.ts
@@ -44313,6 +44313,10 @@ export class SkyBox {
         sources?: any;
         show?: boolean;
     });
+    /**
+     * Determines if the sky box will be shown.
+     */
+    show: boolean;
     /**
      * Gets or sets the the primitive object.
      */
@@ -44354,11 +44358,6 @@ export class SkyBox {
     static createEarthSkyBox(): SkyBox;
 }
 
-/**
- * Determines if the sky box will be shown.
- */
-export var show: boolean;
-
 /**
  * A ParticleEmitter that emits particles within a sphere.
  * Particles will be positioned randomly within the sphere and have initial velocities emanating from the center of the sphere.
@@ -45436,18 +45435,6 @@ export class UrlTemplateImageryProvider {
     pickFeatures(x: number, y: number, level: number, longitude: number, latitude: number): Promise<ImageryLayerFeatureInfo[]> | undefined;
 }
 
-export function sumCollectionProperty(collections: any[] | undefined, propertyName: string): number;
-
-export function forEachCollection(collections: any[] | undefined, callback: (...params: any[]) => any): void;
-
-export function updateCollectionArray(collections: any[] | undefined, vectorModelMatrix: Matrix4, frameState: any): void;
-
-export function destroyCollectionArray(collections: any[] | undefined): void;
-
-export function makeDecodeModelOptions(tileset: any, tile: any, content: any, gltf: any): any;
-
-export function initializeVectorPrimitives(content: any): void;
-
 /**
  * The vertical location of an origin relative to an object, e.g., a {@link Billboard}
  * or {@link Label}.  For example, setting the vertical origin to <code>TOP</code>
```

Notably, internal functions are no longer exported in public Cesium.d.ts types, and the disconnected `show` type is now attached to its intended parent type.

# Author checklist

#### PR Dependency Tree

* **PR #13321** 👈

### DIFF
--- a/packages/engine/Source/Scene/VectorGltf3DTileContent.js
+++ b/packages/engine/Source/Scene/VectorGltf3DTileContent.js
@@ -210,7 +210,6 @@ class VectorGltf3DTileContent {
       c instanceof BufferPolygonCollection;
 
     Color.clone(color, pointMaterial.color);
-    color.clone(pointMaterial.color);
     pointMaterial.size = style.pointSize?.evaluate(null);
     pointMaterial.outlineWidth = style.pointOutlineWidth?.evaluate(null);
     style.pointOutlineColor?.evaluate(null, pointMaterial.outlineColor);


packages/engine/Source/Scene/Cesium3DTileContent.js
```diff
@@ -255,7 +255,7 @@ Object.defineProperties(Cesium3DTileContent.prototype, {
    * not part of the public Cesium API.
    * </p>
    *
-   * @type {Cesium3DTileContentGroup|undefined}
```
donmccurdy, line 258: Presumably this was a typo.

packages/engine/Source/Scene/Cesium3DTileContent.js
```diff
@@ -325,6 +325,7 @@ Cesium3DTileContent.prototype.applyDebugSettings = function (enabled, color) {
  * </p>
  *
  * @param {Cesium3DTileStyle} style The style.
+ * @returns {void}
```
donmccurdy, line 328: Because the stub implementation always throws, we need to specify the return type or TS will assume that child classes must always throw too.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -20,77 +32,67 @@ import BufferPolygon from "./BufferPolygon.js";
...
 
-    /** @type {*} */
-    this._vectorBuffers = undefined;
-    /** @type {*} */
-    this._pointCollections = undefined;
-    /** @type {*} */
-    this._polylineCollections = undefined;
-    /** @type {*} */
-    this._polygonCollections = undefined;
+    /** @type {(BufferPointCollection|BufferPolylineCollection|BufferPolygonCollection)[]} */
```
donmccurdy: Rolling these together into one array worked out a bit simpler for types than trying to update types for multiple arrays and helper functions.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -169,54 +169,30 @@ class VectorGltf3DTileContent {
...
-    /** @param {*} polygons */
-    forEachCollection(this._vectorBuffers.polygons, function (polygons) {
-      const polygon = new BufferPolygon();
-      for (let i = 0; i < polygons.primitiveCount; i++) {
-        polygons.get(i, polygon);
-        polygon.setColor(enabled ? color : Color.WHITE);
-      }
-    });
+    color = enabled ? color : Color.WHITE;
+    this.applyStyle(new Cesium3DTileStyle({ color }));
```
donmccurdy, line 195: I know this isn't quite 'working' because applyStyle() is still a stub method, but I have a PR for that coming soon, and the .setColor() method used below is being removed, so this seemed easier for now.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -260,116 +229,54 @@ class VectorGltf3DTileContent {
...
-    this._decodeModel = this._decodeModel && this._decodeModel.destroy();
-    destroyCollectionArray(this._pointCollections);
-    destroyCollectionArray(this._polylineCollections);
-    destroyCollectionArray(this._polygonCollections);
-    this._pointCollections = undefined;
-    this._polylineCollections = undefined;
-    this._polygonCollections = undefined;
-    this._vectorBuffers = undefined;
+    this._decodeModel?.destroy();
+    this._decodeModel = undefined;
```
donmccurdy, line 298: This was required for types - TS doesn't like us assigning 'void' (the result of destroy()) to the property.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -400,15 +310,15 @@ function initializeVectorPrimitives(content) {
...
-  content._vectorBuffers = vectorBuffers;
-
   if (!defined(vectorBuffers)) {
     return;
   }
 
-  content._pointCollections = vectorBuffers.points;
-  content._polylineCollections = vectorBuffers.polylines;
-  content._polygonCollections = vectorBuffers.polygons;
+  content._collections = [
```
javagl, line 383: Just stumbling over that: Couldn't these collections be "large"? 
I think that the spread operator puts the elements on the stack, which isn't ideal ( https://github.com/CesiumGS/cesium/issues/12053 ) **but** I have some vague memory that it might _not_ do this when creating a _new_ array. Maybe something to check...

donmccurdy:
~~I broke tsd-jsdoc, moving PR to draft while I resolve that. :(~~

**UPDATE:** Fixed.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
donmccurdy: The collections could be large, but `vectorBuffers.points` is an _array of collections_, and there shouldn't be more than a small handful of collections. So I believe the array-spread is OK in this case! Definitely wouldn't want to do it for every point in a point cloud though, right. :)

javagl: Ah yeah. The old dilemma. A point is `point`. A set of points is `points`. What is a set of sets of points? (yeah, `pointCollections` is a mouthuful, but was generalized to `collections` on the left here anyhow).
I once wrote a library for "TimeSeries". That's a bummer 😶

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -122,42 +125,40 @@ class VectorGltf3DTileContent {
...
-  /** @param {*} value */
   set group(value) {
     this._group = value;
   }
 
   /**
-   * @param {*} _featureId
-   * @returns {undefined}
+   * @param {number} _featureId
+   * @returns {Cesium3DTileFeature}
```
danielzhong, line 174: Should this return `{Cesium3DTileFeature|undefined}` instead?

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -122,42 +125,40 @@ class VectorGltf3DTileContent {
     return this._resource.getUrlComponent(true);
   }
 
-  /** @returns {undefined} */
+  /** @type {Cesium3DTileBatchTable} */
```
danielzhong, line 149: Should this return `{Cesium3DTileBatchTable|undefined}` ?

packages/engine/Source/Scene/VectorGltf3DTileContent.js
javagl: c.f. the bottom part of https://github.com/CesiumGS/cesium/pull/13318#issuecomment-4093497566 and the response there: We are aware that there are many *many* type definitions that are "wrong", insofar that they are missing the `undefined` part. Hopefully, this can be fixed in one rip-the-bandaid-off-quickly-pass when the structures are prepared and the "strict mode" can be enabled.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
donmccurdy: See next comment on `getFeature`.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
donmccurdy: One aspect of it is TypeScript "strict mode", yes.

Another is that this is a stub implementation of the Cesium3DTileContent interface, in which these types are also non-nullable. I'm not able to actually type check against the interface (it needs to be an ES6 class). But if we could, TypeScript should complain that the implementation types don't match the interface types. I'm biasing to just match the interface for now, especially since we do want to come back and fill in this stub method later.

But more generally, yes, it's a good thing to include `|undefined` if that's the intended public interface. For a stub method to avoid conflicting types with an interface, or to avoid needing to write up JSDoc types that aren't the intended public API, I might reach for `throw new DeveloperError('not implemented')`.

packages/engine/Source/Scene/VectorGltf3DTileContent.js
donmccurdy: Addressed now that we've merged the Material API PR - applyStyle() is implemented.

PR #13322, opened 2026-03-20
Gaussian Splat Optimization

danielzhong:
# Description
This PR applies profile driven optimizations to Gaussian splat loading and update.

## Main changes:

* Adds a fast path and transform caching in `GaussianSplatPrimitive.transformTile` to reduce per-splat matrix/TRS work
* Preserves original glTF attribute data in `GaussianSplat3DTileContent` and keeps transformed copies separate for correct rebuild behavior
* Reuses aggregate snapshot buffers for positions, scales, rotations, and colors to reduce allocation churn and GC pressure
* Makes steady-state sorting less aggressive during camera movement to reduce repeated sort work

These changes were based on Chrome profiling, where the main hotspots were transform work, repeated large allocations, and repeated scheduling/cloning during texture generation and sorting. The goal is to reduce CPU time and GC pressure during Gaussian splat loading and runtime updates.

Before:
image

After:
image

## Testing plan
[DEMO link](https://ci-builds.cesium.com/cesium/DanielZ/Splat_FIX7/Apps/Sandcastle2/index.html#c=rVeNdto2FH4Vja09piMCO0B60pCNQNKRhaRJaHu6sXMibAEKsuVJMj9t8057hj3ZrvwDNrCedVtyiPH9073fd3WlMD8UUqMXiCjUoYpFPhpL4aNhyY3fhqVXQzkMzMcVgdJIaQL2LfTJiBDiIjhGB7bjYLtu2y+blVRM9DGqH+Fmvd6sNVLhlLLJFOT2Ua2G6y/XUuKxYNKlE9CAopbKQ6bdaSw9cBprqRScx8LM8GkrvzmjCyohwYAu0orwu1hmZTV1RKAJC6gclipZHZpKCbLjzGWQvGMDxnshuZcKrHKyajleNlkMK5cGFHt0FE3up2JxIYlP1Rsq7ynk5EEuWkZ0nWfq5IKRJFhRbdKz0jw8qjQLiGZik0sHEIdvJDiM04HyJaXKipnAQEAFpV+Jzr4mUJdT0IRkFEpOgqYLrYFfL9Mneoq1uAMpCbLwG3ayYCkzX3bLyNs4GeK+7JNSm7k85YHe8CvkiOkOlBOT/G8Qgs5JguYCDiTkMRbS38RcixSmROlr2CbTt+FAXLAl9WKKrVwu22mC84TuTdBjkEvgUqvYB6FQzBD0vlNBe+MWei114kLM2pvkrYzufEUpnLnt8FPC6RvD0Z3J08pI+tpO+MoWMGslL2Xz2NTGacoD2Cf7BqA7xLVXBeUd0RTWWBvsX70YZ5uXiIGjJ9zIB3ixKynEPOfUvMF88Nh8WEpcbkaP1NWYKMUmgRUxrPSK0/W8yOg6hklJRkrwSFMYJ+kwEaGR2064XMs4Hesd4cde4NFlPPfWY494ya4E01q4RAV7V3Ahjerb8Xi8lo6IO5tIEQWeUcnJiFi1SvyLm+WNlZAejZ1tCAsZMw/Ftk6jUck+New0tl0MsJEyjs1cKmMYohfEZ3xlNGqlNPUPIlZBByQMOT1IJBV0TyeCore9CroTI6EFbEVg6UBRycaFYPfsI90BKFKGRQ5MGFUgggzknbnAyYjyf8htbIs1XWpzFIAa/B5uTN8gFVIKMH73qdhG0F/xvrfs8tOff1TVQy5O3BjYJ3LCgjOhtTBDJAVr+/AE0OO59bdZsiCMdJZnYo71KqRxyGQDlfI6nwWxqrYlJstYbDtFOXASJvbYLmrmhEdmlXstof327qLillhHzG+LBfP01FDlOLU1jTmuUifo8fM5FHwFsxDGmdwUXkFWGbVOs3g7Q+E68kdgn086iY7+64xA6P9qi1y9MDdgO9DA60wZ96x4hUS7pUkKKtwq3OyaUjCM2M6R4HLhzrAIBgweO9BasToPatKLnikvcfUo1yS5rCj0+bM5Ie1mCgobI2sPst+0WqhW3lwm9p1OUmjjY24i+0K8gBQy6J9S3FC1irYONfQ9ygVCSZkJ2TAsNZyRAfrV/s2aah2q42pVRNpEwGI8Zi49bDYASL8qFqT6Qw8mUq/barf7s3b3rfNLt7e47vYXV4N3s+vXHzR8PvZX9vRmcOlfDa4fb7qXj9eD23r/kfv9ydnrdvzTtd812vV52F3dNm8//B44ZDnQi7NFZ9zmtr6/PB/fXR79rGfPnNGhPgt8cLmd3V702+0vWgzed+D57NB7TpdzFUnesp8bUH0B7LTu4PDtU6XIhJoiylkHaLkqsqoZp3ClBGrJggBC2f0jfhx2B4k6viX1RNBW8NLzrHqjfuQ4TsZGgcxQMh+OujlVpresdIHE9Am5BE55ZMHlWMhyPhUBM5GLifVwblRAanx7yPIzWyn2eXpII5k/Dw+lSukkniinJs6PLPnvBNCwMK5C4SHc4KiqjiJ3BlW4SoF3cFLNXE5gyCPmtfbc9KHVYXCBZhxxbs6aYen0pAr2Bbc0zZs5lZysjMnUPr1KhBjjkyq87nppIfiIyFzEvwA)

# Author checklist

### DIFF
--- a/packages/engine/Source/Scene/GaussianSplat3DTileContent.js
+++ b/packages/engine/Source/Scene/GaussianSplat3DTileContent.js
@@ -678,6 +678,10 @@ GaussianSplat3DTileContent.prototype.update = function (primitive, frameState) {
     this.worldTransform = loader.components.scene.nodes[0].matrix;
     this._ready = true;
 
+    // SPZ decode produces Float32Array attributes for these semantics, so
+    // typedArray.slice() preserves the expected runtime type for current data.
+    // If future splat encodings use quantized integer attributes here, revisit
+    // this assumption before relying on the copied array type.
     this._positions = ModelUtility.getAttributeBySemantic(
       this.gltfPrimitive,
       VertexAttributeSemantic.POSITION,


packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -671,26 +678,20 @@ GaussianSplat3DTileContent.prototype.update = function (primitive, frameState) {
...
+
+    this._rotations = ModelUtility.getAttributeBySemantic(
+      this.gltfPrimitive,
+      VertexAttributeSemantic.ROTATION,
+    ).typedArray.slice();
+
+    this._scales = ModelUtility.getAttributeBySemantic(
+      this.gltfPrimitive,
+      VertexAttributeSemantic.SCALE,
+    ).typedArray.slice();
```
weegeekps, line 698: This is fine and we should not change it. It does add a potential footgun that we need to be aware of. 

The `slice()` function on typed arrays is fast, but it doesn't change the type. With SPZ data, this isn't an issue since it's guaranteed to be 32-bit floating point. The glTF extension _does_ allow these to be normalized shorts and bytes for quantization. Irrelevant to our current implementation but if we decide to support other compression formats in the future it could become relevant.

You may want to add a comment pointing out that this is assuming that the semantic is a float32, and that is a safe assumption with data from SPZ.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1236,6 +1300,41 @@ GaussianSplatPrimitive.transformTile = function (tile) {
...
+    scratchMatrix3[2] * scratchMatrix3[8];
+  const dot12 =
+    scratchMatrix3[3] * scratchMatrix3[6] +
+    scratchMatrix3[4] * scratchMatrix3[7] +
+    scratchMatrix3[5] * scratchMatrix3[8];
+  const uniformScale = (col0Len + col1Len + col2Len) / 3;
+  const hasUniformScale =
+    Math.abs(col0Len - uniformScale) <= RIGID_TRANSFORM_EPSILON &&
+    Math.abs(col1Len - uniformScale) <= RIGID_TRANSFORM_EPSILON &&
+    Math.abs(col2Len - uniformScale) <= RIGID_TRANSFORM_EPSILON;
```
weegeekps: Using `1e-5` for the epsilon is fine for most of the places you're doing it, except for here.

3DGS are _very_ sensitive to small changes in scale. Gaussians that are showing very fine details often do need stable precision up to the float32 limit of `1e-7`.

It would be best if you just exclude trying to test or use uniform scale entirely. Anything else will introduce errors that will be perceptible with highly detailed tilesets.

danielzhong: Good point. I removed the uniform scale path entirely.

The fast path now only applies to unit-scale rigid transforms, so any transform with non-unit scale falls back to the full per-splat TRS path. That keeps the transform cache and the other optimizations, while avoiding scale approximation error for detailed splats.
https://github.com/CesiumGS/cesium/pull/13322/commits/28dd975a08ff7370c6f99aa666f3c6d8d1b3c5a5

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
danielzhong: Makes sense. Added comment.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1253,9 +1352,9 @@ GaussianSplatPrimitive.transformTile = function (tile) {
     VertexAttributeSemantic.SCALE,
   ).typedArray;
 
-  const position = new Cartesian3();
-  const rotation = new Quaternion();
-  const scale = new Cartesian3();
+  const position = scratchTransformPosition;
+  const rotation = scratchTransformRotation;
+  const scale = scratchTransformScale;
```
weegeekps, line 1357: Ah, you missed this line of dead code.

danielzhong: I think this may be a diff context mixup. The old `const scale = new Cartesian3();` allocation was removed here along with the old `position`/`rotation` temporaries.

The remaining `const scale = scratchTransformScale;` is still live and used later in the loop, so this one is not dead code.

weegeekps: Oh, you're right. I see it now.

ggetz:
Code looks good to me! Thanks @danielzhong and @weegeekps, great catch for performance!

javagl:
It's already merged now (I wanted to do another test run, but was not be able to do this - maybe by end of the week).

But I' curious: The `transformTile` call appeared prominently in some previous performance traces. And I noticed that it is behind some guard that is supposed to let it be executed only once, via https://github.com/CesiumGS/cesium/blob/d6634bea1094c22000f1ff8d89594846f9f10590/packages/engine/Source/Scene/GaussianSplat3DTileContent.js#L661 . Now, it is called from other places, namely in the `GaussianSplatPrimitive`.

- Is that `if (!this._transfomed...` guard and the call in `GaussianSplat3DTileContent` still required? (Or are the calls in the `GaussianSplatPrimitive` sufficient and idempotent?)
- (The opposite: ) Does it really have to be called multiple times? (I think it might not be necessary to call it multiple times, _unless_ the transform of a tile changes, with details depending on whether the relevant transform here involves then `modelMatrix` or not)

danielzhong:
> It's already merged now (I wanted to do another test run, but was not be able to do this - maybe by end of the week).
> 
> But I' curious: The `transformTile` call appeared prominently in some previous performance traces. And I noticed that it is behind some guard that is supposed to let it be executed only once, via
> 
> https://github.com/CesiumGS/cesium/blob/d6634bea1094c22000f1ff8d89594846f9f10590/packages/engine/Source/Scene/GaussianSplat3DTileContent.js#L661
> 
> . Now, it is called from other places, namely in the `GaussianSplatPrimitive`.
> * Is that `if (!this._transfomed...` guard and the call in `GaussianSplat3DTileContent` still required? (Or are the calls in the `GaussianSplatPrimitive` sufficient and idempotent?)
> * (The opposite: ) Does it really have to be called multiple times? (I think it might not be necessary to call it multiple times, _unless_ the transform of a tile changes, with details depending on whether the relevant transform here involves then `modelMatrix` or not)

Ah, great catch. I think you're right here.

I cleaned this up by making `transformTile()` own the `_transformed` state and removing the duplicate call from `GaussianSplat3DTileContent`.
https://github.com/CesiumGS/cesium/pull/13324

PR #13325, opened 2026-03-24
Add GPU lookup rendering for vector tiles

danielzhong:
<p align="center">
  image
  image
</p>

# Description

Note: This is still a very rough draft, just getting it to render for now. It will need cleanup, revisions, and refactoring.

Please read my notes to understand the implementation better!
[README.md](https://github.com/user-attachments/files/26226099/README.md)

This PR adds a `gpuLookup` rendering mode for glTF vector tiles using `CESIUM_mesh_vector`.

I'm currently keeping:
- route glTF vector content into `VectorGltf3DTileContent`
- decode it into `BufferPointCollection`, `BufferPolylineCollection`, and `BufferPolygonCollection`
- keep the normal collection renderer as the baseline / fallback path for better debug now. Turn on GPU lookup using:
```
const tileset = await Cesium.Cesium3DTileset.fromUrl(url, {
  heightReference: Cesium.HeightReference.CLAMP_TO_TERRAIN,
});

```

What changes is that polylines and polygons can now render through an alternate tile local GPU lookup path.

## Compare
| Area | Earlier experiment branch | This PR |
| --- | --- | --- |
| Input pipeline | Terrain / GeoJSON-oriented experiment path | `CESIUM_mesh_vector -> VectorGltf3DTileContent -> Buffer*Collection` |
| Points | No special lookup path | Still use normal `BufferPointCollection` rendering |
| Lines | Tile-local segment lookup on terrain surface UVs | Tile-local segment lookup on a generated tile surface patch |
| Polygons | Edge segments + ray casting / cutFlags | regular polygon fill |
| Carrier surface | Reused terrain tile mesh | Generated tile surface patch over the tile region |
| Lookup domain | Tile-local terrain UV | Tile-local geographic UV from tile bounding region |

### My attempt

- I first tried building GPU lookup data per decoded collection using a single local 2D plane. That worked structurally, but real datasets like `sample-us-outline` and `sample-us-states` are draped over large curved regions, so the collection-local plane assumption broke down. The result was that the lookup path immediately fell back to the normal geometry renderer. This PR moves lookup generation to tile-local surface / UV space instead.

- Polygons: I tried to carry over ray-casting / cut-flag style idea from the earlier experiment, but it rendered very slowly on our dataset for reasons, I need to spend more time to investigate. Polygon is currently using regular polygon fills, which is the approach we’ve decided to go with.

### Future work / Current problem

- Polyline: visual quality and performance is bad
- Refactor, Refine better `Buffer*Collection` pipeline or API

## Testing plan

<details>
<summary>Example Sandcastle code:</summary>

```js

const url = "http://localhost:8082/sample-us-states.tileset.json";

const tileset = await Cesium.Cesium3DTileset.fromUrl(url, {
  heightReference: Cesium.HeightReference.CLAMP_TO_TERRAIN,
});

window.tileset = tileset;
viewer.zoomTo(tileset);
```

</details>

# Author checklist


donmccurdy:
Thanks Daniel! I haven't had the chance to review yet, but a few quick comments in case they're helpful sooner:

> Turn on GPU lookup using: ... `vectorTileRenderingMethod: "gpuLookup"`

Might need @lilleyse to confirm, but I think from the user's perspective this would fall under the existing `heightReference` option on the `Cesium3DTileset` constructor. For example `HeightReference.CLAMP_TO_TERRAIN` would clamp to terrain, `HeightReference.CLAMP_TO_GROUND` would clamp to ground, and `HeightReference.NONE` would preserve the existing rendering method.

> Polygons: I tried to carry over ray-casting / cut-flag style idea from the earlier experiment...

My recollection is that we don't need to bring in the cut-flag implementation, and that we'd instead draw the full polygon. /cc @xuanhuang1

> Maybe remove the regular render path once GPU lookup is stable.

See above about `heightReference` — we'll want to able to render vector data _either_ clamped to terrain or unconstrained, so having two render paths for the same underlying data model is our goal. We can of course extend/improve the BufferPrimitiveCollection APIs as needed to better support the new GPU Lookup rendering path.

Probably a `heightReference` option on the BufferPrimitiveCollection constructor (per-collection, not per-primitive) would make sense to choose the rendering path.

xuanhuang1:
> My recollection is that we don't need to bring in the cut-flag implementation, and that we'd instead draw the full polygon.

Yup the prototype cut flags were a workaround to avoid rendering artificial elements injected on tile boundary, which should be removed in actual implementation since shape boundaries will be extended by an explicit [padding space](https://github.com/CesiumGS/3d-tiles/issues/825#issuecomment-3873359710).

This means in draping we just render the polygon normally.

donmccurdy:
One important architectural clarification - vector data from BufferPolyline- and BufferPolygonCollections will need to be "draped" onto geometry from terrain tiles or 3D Tiles. Unfortunately, the tiling scheme used by a vector tileset may be arbitrarily different from the tiling scheme used by the terrain or 3D Tiles:

overlay tile schemes

> _Hypothetical tiling scheme overlay, showing vector tiles (purple) and terrain (blue)._

Because of that, we probably want to compute GPU Lookup structures on the terrain or 3D Tiles tiling scheme, not on the vector tiles tiling scheme. We'll also want to use the geometry of the terrain or 3D Tiles, rather than projecting vector data onto a patch of the ellipsoid. The terrain may be mountains, for example, and vectors projected to the ellipsoid would then be occluded by the terrain, when we need them to render as decals on the terrain.

For terrain, ultimately we'll probably want to render points/lines/polygons as part of the GlobeSurfaceShader. Similarly to how GlobeSurfaceTile currently uses image data generated by an ImageryProvider, we may want to extend GlobeSurfaceTile to load vector data generated by a (hypothetical) VectorProvider, and render it with the GPU Lookup technique.

In preparation for relocating the implementation into GlobeSurfaceTile, I've pushed a few changes on this branch:

- merged latest `main` into branch
- removed vector aggregation from the VectorGLTF3DTileContent class, so that the new rendering code is isolated to `packages/engine/Source/Scene/renderBufferPolylineCollectionGpuLookup.js`
- removed some non-urgent changes (e.g. unit tests on .heightReference) so there's less to migrate into GlobeSurfaceTile

I haven't started moving the rendering into GlobeSurfaceTile yet, but will begin that if time allows this week. Rendered results should be unchanged from before my recent commits:


Screenshot 2026-04-02 at 3 09 53 PM

donmccurdy:
Crosslinking: 

- #13584

PR #13328, opened 2026-03-24
Add AI Acknowledgment to PR and Issue Templates

lukemckinstry:
# Description

## Testing plan

# Author checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -25,6 +25,8 @@ To ensure your Pull Request is reviewed and accepted quickly, please refer to ou
 
 <!-- Describe in detail how you tested your changes. If this fixes a bug, list the steps to reproduce the original issue. -->
 
+<!-- markdownlint-disable-next-line MD025 -->
+
 # Author checklist
 
 - [ ] I have submitted a Contributor License Agreement
@@ -33,3 +35,8 @@ To ensure your Pull Request is reviewed and accepted quickly, please refer to ou
 - [ ] I have added or updated unit tests to ensure consistent code coverage
 - [ ] I have updated the inline documentation, and included code examples where relevant
 - [ ] I have performed a self-review of my code
+
+## AI acknowledgment
+
+- [ ] I used AI to generate content in this PR
+- [ ] (If the above is checked) I have reviewed the AI-generated content before submitting
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -39,4 +39,12 @@ To ensure your Pull Request is reviewed and accepted quickly, please refer to ou
 ## AI acknowledgment
 
 - [ ] I used AI to generate content in this PR
-- [ ] (If the above is checked) I have reviewed the AI-generated content before submitting
+- [ ] If yes, I have reviewed the AI-generated content before submitting
+
+If yes, I used the following Tools(s) and/or Service(s):
+
+<!--(e.g., ChatGPT, GitHub Copilot, Claude, Gemini, etc.) -->
+
+If yes, I used the following Model(s):
+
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -33,3 +35,8 @@ To ensure your Pull Request is reviewed and accepted quickly, please refer to ou
 - [ ] I have added or updated unit tests to ensure consistent code coverage
 - [ ] I have updated the inline documentation, and included code examples where relevant
 - [ ] I have performed a self-review of my code
+
+## AI acknowledgment
+
+- [ ] I used AI to generate content in this PR
+- [ ] (If the above is checked) I have reviewed the AI-generated content before submitting
```
ggetz: Can we also add a field for the contributor to specify _which_ model was used? It'll be useful to start keeping track of this because, if we ever introduce automated PR reviews, we'll want to ensure we use a _different_ model to review than the the one which modified the code. That way, we'll avoid reenforcing mistakes.

ggetz (commented):
Awesome, thanks @lukemckinstry! Is there a reason this PR is in draft?

.github/pull_request_template.md
lukemckinstry: added fields for 1. service/tool and 2. model

ggetz (approved):
Updates look good to me! Thank you @lukemckinstry!

.github/pull_request_template.md
```diff
@@ -25,6 +25,8 @@ To ensure your Pull Request is reviewed and accepted quickly, please refer to ou
 
 <!-- Describe in detail how you tested your changes. If this fixes a bug, list the steps to reproduce the original issue. -->
 
+<!-- markdownlint-disable-next-line MD025 -->
```
javagl, line 28: I just stumbled over that - it looks like this was ~"not properly picked up" for me after a `merge main`, and caused an error when running the commit hook.

Is there a (very) strong reason to use this directive, and  _not_ just add that second `#` in front of the `# Author checklist`?

lukemckinstry: Thanks for pointing this out, seems good to follow normal markdown hierarchy so we don't have these pesky markdown lint issues

lukemckinstry: short PR with fix is here https://github.com/CesiumGS/cesium/pull/13368

donmccurdy:
After reading several PRs using the new template recently, I've been wondering if perhaps the AI Acknowledgement should be phrased in a way that requires some response from the contributor either way? For example:

## AI acknowledgment

- [ ] Yes, I used generative AI to create content in this PR, and...
  - [ ] I have reviewed the AI-generated content before submitting
  - [ ] Tools or Services: 
  - [ ] Models: 
- [ ] No, generative AI was not used to create content in this PR

markschlosseratbentley:
I agree with @donmccurdy. Right now I think it's easy to just scroll past and leave the section empty since it does not necessitate a negative response. With @donmccurdy's proposal, a submitter will _have_ to stop, think, consider and purposely check at least one box.

cc @ggetz @lukemckinstry

lukemckinstry:
I like this suggestion. Definitely an improvement in my opinion.
This still leaves open the situation where people are using AI in a supportive manner for research and filling in small parts of the code - rather than entirely or mostly generating code & content with a coding agent. Is there a threshold below which we would not want to burden people with the expectation they report their minimal AI usage? And do we need to wordsmith further or add a third option to make this clear (at the risk of cluttering up the form and making things less clear).

I think the initial intent was to make sure authors took ownership and did a human review of their code before submitting to us and help inform our review by making us aware whether or not the code was entirely or partially generated by an agent.

donmccurdy:
PR:

- https://github.com/CesiumGS/cesium/pull/13488

PR #13330, opened 2026-03-25
Extract Geometry from 3D Models

alarkbentley:
# Add `getGeometry` API for extracting vertex data from Models

## Summary

Adds a new public `Model.prototype.getGeometry()` method that extracts per-primitive vertex attributes (positions, normals, colors, feature IDs, etc.) from a loaded Model. This is backed by a new `ModelGeometryExtractor` utility, a generalized `GeometryResult` class, and significant refactoring of `ModelReader` to consolidate scattered vertex-reading logic into a single, reusable module.

Despite the fact that the current implementation is synchronous the introduced API endpoint is exposed with an async handle to enable future non-breaking improvements such as WebWorkers or async GPU readback.

## Motivation

There was no public API to access the underlying vertex information of 3D Models in Cesium. Use cases such as geometric analysis, custom visualizations, and spatial queries on picked features required workarounds or access to private internals.

Key driver for this was the ongoing PoC to extract 2D footprints based on model geometry in runtime: https://github.com/CesiumGS/cesium/pull/13244

## Usage example

```js
const feature = scene.pick(movement.endPosition);
const model = feature.primitive;
const results = await model.getGeometry({
  attributes: ["POSITION", "COLOR_0", "_FEATURE_ID"],
  extractIndices: true,
});
for (const entry of results) {
  console.log(`Positions: ${entry.getPositions()?.length}`);  // Cartesian3[]
  console.log(`Colors: ${entry.getColors()?.length}`);         // Color[]
  console.log(`Feature IDs: ${entry.getFeatureIds()?.length}`); // number[]
  console.log(`Indices: ${entry.indices?.length}`);
  console.log(`Primitive type: ${entry.primitiveType}`);
  console.log(`Vertices: ${entry.count}, Instances: ${entry.instances}`);
}
```

## Screenshot

image

| Composite | InstancedOrientation | PointCloud | ImplicitRange ids |
|----------|----------|----------|----------|
|composite|instanced_orientation|point_cloud|instanced_implicit_range|

### New API

- **`Model.prototype.getGeometry(options)`** — Public async method on the `Model` class. Returns a `Promise<GeometryResult[]>` with one entry per primitive.
  - `options.attributes` — Array of semantic strings to extract (e.g. `"POSITION"`, `"NORMAL"`, `"COLOR_0"`, `"_FEATURE_ID"`, `"TEXCOORD_1"`). (optional, if not set will extract all)
  - `options.extractIndices` — Whether to extract vertex indices (optional, default `false`).
- **`Cesium.GeometryResult`** — Generalized result class holding extracted per-primitive data. 
  - Properties: `attributeNames`, `attributeValues` (Map), `attributeTypes` (Map), `indices`, `primitiveType`, `count`, `instances`. 
  - Convenience methods: `getPositions()`, `getNormals()`, `getColors()`, `getFeatureIds()`, `getAttributeValues(name)`, `getAttributeType(name)`, `filter(predicate)`, `getGeometryResultByFeatureId(geometryList, featureId)`.

### New modules

- **`ModelGeometryExtractor`** — Extracts requested vertex attributes from a loaded `Model`. Performs a single scene-graph traversal with per-instance transform support, and reuses `ModelReader` for typed-array access and GPU readback.

### Bug fixes

- **`ModelReader`** — 
  - **`octDecode` bug fix** — Fixed incorrect argument order in `AttributeCompression.octDecodeInRange` call (was passing a `Cartesian3`, now correctly passes `c.x, c.y`) and fixed swapped arguments in the subsequent `Cartesian3.pack` call. (previously in [pickModel](https://github.com/CesiumGS/cesium/blob/0da68a8f1055f534f43d7d74a7050e193246fbc7/packages/engine/Source/Scene/Model/pickModel.js#L386)) [Link](https://github.com/alarkbentley/cesium/blob/6e3f61e86b83199ccebabcc0af0b2100dbdbe7c7/packages/engine/Source/Scene/Model/ModelReader.js#L342)
  - **`getInstanceTransforms` bug fix** — Fixed incorrect matrix multiplication order for meshes with EXT_mesh_gpu_instancing. (previously in [pickModel](https://github.com/CesiumGS/cesium/blob/0da68a8f1055f534f43d7d74a7050e193246fbc7/packages/engine/Source/Scene/Model/pickModel.js#L162)) [Link](https://github.com/alarkbentley/cesium/blob/6e3f61e86b83199ccebabcc0af0b2100dbdbe7c7/packages/engine/Source/Scene/Model/ModelReader.js#L1055)

### Refactoring

- **`ModelReader`** — 
  - **`readAttributeAsTypedArray`** now supports in-memory `typedArray` fast path — returns the attribute's existing typed array directly when available, skipping GPU readback entirely. Previously only handled GPU buffer reads.
  - [new] **`readImplicitRangeAsTypedArray`** Reads feature IDs from an implicit range feature ID set into a typed array.
  - [new] **`forEachPrimitive(model, options, callback)`** — New static method that traverses all runtime nodes and primitives in a model's scene graph. Accepts an optional `options` object with `mapProjection` (for 2D projection) and `perInstanceFeatureIds` (boolean, default `false`, to control whether per-instance feature IDs are fetched). Computes per-node transforms, then invokes the callback with `(runtimePrimitive, primitive, instances, computedModelMatrix)`.
- **`pickModel`** — Significantly simplified (~295 lines removed, ~86 added) by delegating node traversal and vertex unpacking to `ModelReader` instead of inlining its own scene-graph walk and vertex-reading logic.

### Sandcastle demo

- Added `3d-tiles-geometry-extraction-dev` demo showing how to pick a feature, extract its geometry, and visualize extracted vertex positions as point entities.

### Tests

- **`GeometryResultSpec`** — 31 tests covering initialization defaults, all convenience getters (`getPositions`, `getNormals`, `getColors`, `getFeatureIds`), fallback key resolution (e.g. `COLOR` vs `COLOR_0`, `_FEATURE_ID` vs `_FEATURE_ID_0`), generic `getAttributeValues`/`getAttributeType`, property storage, `filter()` (matching vertices, preserving attribute types/names, undefined when none match, all-pass, multi-instance, immutability of source), and `getGeometryResultByFeatureId` (match by ID, no match, empty list, first-match from multiple geometries, undefined when no feature IDs).
- **`ModelGeometryExtractorSpec`** — 36 tests covering validation (no options, no model, not ready), position extraction (identity transform, node transform, no-indices), index extraction (`extractIndices` true/false/default), attribute selection (`attributes` array filtering, empty `attributes`), feature ID attributes (extraction, not-requested, match by label, match by positionalLabel, no feature IDs when absent), primitiveType propagation (default TRIANGLES, non-default LINES), multi-set color extraction (`COLOR_0`/`COLOR_1`), normal extraction, instancing (position duplication, color duplication, instance featureId fallback, instance featureId precedence over per-vertex), `FeatureIdImplicitRange` (with repeat, offset+repeat, no repeat, by label, duplication per instance, instance-level implicit range), `instanceFeatureIdLabel` (default label, matching label, non-matching label, custom label on FeatureIdAttribute), and default attributes behavior (all attributes extracted when undefined, instance feature IDs extracted when undefined).
- **`ModelReaderSpec`** (`forEachPrimitive` block) — 21 tests covering scene-graph traversal (undefined sceneGraph, no runtime nodes, skips empty nodes, multi-node iteration), callback contract (once per primitive, passes runtimePrimitive/primitive/instances/computedModelMatrix), transform computation (non-identity node+model matrices, MAT4 typedArray, MAT4 buffer readback, VEC3 translation attributes, VEC3 with translation+rotation+scale, interleaved buffer with TRS), world-space instancing (`transformInWorldSpace: true`), 2D projection (`mapProjection` provided/undefined), and instance feature IDs (FeatureIdAttribute with opt-in `instanceFeatureIdLabel`, not populated without opt-in, undefined when no attribute, undefined for non-instanced nodes, with VEC3 translations, FeatureIdImplicitRange with repeat, FeatureIdImplicitRange without repeat).
- **`ModelSpec`** (`getGeometry` block) — 4 tests covering default behavior (returns array), positions extraction by default, options passthrough (`attributes: []`), and undefined options handling.

## Testing plan

- All new and updated specs pass.
- Verified with the included Sandcastle demo (`3d-tiles-geometry-extraction-dev`) against a batched 3D Tiles tileset.
- Verified that the demo `3d-tiles-picking-dev` that is utilizing the refactored `pickModel` still is working.

# Author checklist

## AI acknowledgment

GitHub Copilot

Claude 4.6 Opus


mzschwartz5:
I know this is just a draft PR but - before you spend too much (more) time on this - is there an github issue this stems from? I haven't evaluated the usefulness of this feature, but it's good to discuss new features before so we don't waste your time if it's not a good fit for Cesium.

javagl:
Until alarkbentley chimes in: This _looks_  like it is supposed to become sort of a building block of https://github.com/CesiumGS/cesium/pull/13244 

In general, the functionality to "access geometry data" is required at many, many places. And it is _implemented_ at many,  many places, which is not ideal. Some places that come to my mind are extracting the vertex positions for picking, handling "outline rendering", or extracting the vertex positions for draping. 
(In the context of the latter, I already started a `ModelReader`, but... when something is not implemented thoroughly and properly, and from the ground up (!), then everybody will implement solutions for "the part that is currently needed", duplicating work and causing churn in the long run. It is what it is)

However, the specific context here appears to aim at the possiblity to extract "something like a clipping polygon" from a tileset. To be confirmed.

alarkbentley:
@mzschwartz5 As @javagl mention this is a building block for #13244. I decided to split this up into two separate PRs to keep the complexity down. Also this capability itself can have some value outside of generating footprints in the case the user want's to do calculations based on the tile geometry.

If we decide to not expose this in the public API in the end that is fine, we could keep it as an internal utility. But I do think it could provide end user value.

I will try to complete this by the end of week and add a proper PR description for you to review.

javagl:
> Also this capability itself can have some value outside of generating footprints in the case the user want's to do calculations based on the tile geometry.
> 
> But I do think it could provide end user value.

**Yes!!!** 

I already mentioned https://github.com/CesiumGS/cesium/issues/13052 , outlining some of the use cases that are currently prevented by the bufffer-vs-typedArray handling, and which boil down to that pretty fundamental task of "accessing geometry data".

Given that this is a draft, and it's not yet clear where _exactly_ this is going, it may be too early (and maybe not even my "job") to provide first feedback here. But I've already been pondering with ~"some of this", so maybe it's OK: 

I had a short look at `ModelMeshUtility`. 

There are `readPositionData`, `readColorData`, and `readFeatureIdData`. They have quite some overlap. All of them are doing the same twiddling of `byteStide/byteOffset/IsInterleaved`. And I wonder whether there will be `readNormalData`, `readTexCoordData`, `readJointsData`, `readWeightsData`, `readScaleData`, and `readRotationData` in the future (all doing essentially the same thing). 

There is `decodeAndTransformPosition`. And I tend to apply some scrutiny when I see an  `and` in a function name. The `instanceTransform` is just passed on for that `Matrix4.multiplyByPoint` at the end. It could make sense to split this up. There will certainly be a place and time when someone wants to _only_ dequantize (and _not_ transform). Not the positions, though - at least, not in this form: It does not make much sense to apply "oct-encoding" to positions. This is only applied to normals. Right now, there is no `readNormalData` function, though. 

However... even if it's once more a shameless self-promotion: **Iff** anything that is done in the `ModelMeshUtility` can **not** be done with the [`ModelReader`](https://github.com/alarkbentley/cesium/blob/7150cedf88eb549b313882b327a28941d5a8e957/packages/engine/Source/Scene/Model/ModelReader.js), then I'd (be curious what that is, and) extend the `ModelReader` accordingly. Yes, it's certainly not complete and not perfect, because when I had to implement that, as part of another feature, I did (of course) not have enough time to iron out the kinks that result from 10 years of _simply not having such a class_. But I _tried_ to keep it generic, with the goal of having reusable building blocks. We already have sooo much redundancy, and redundancy is redundant. Maybe we can avoid that.

(One thing that the `ModelReader` does not offer is some of the instance-transforms handling and `computeNodeTransforms`. This was apparently extracted from `pickModel`. And it should be a dedicated function. But when it _is_ a dedicated function, maybe it could also be used (i.e. just _called_) in `pickModel`?)

alarkbentley:
@javagl I'm currently working on merging the capabilities of ModelMeshUtility and ModelReader (see my commit a few minutes ago). My initial mistake was to base my implementation on the pickModel (which I assume was written before we had the ModelReader). 
Currently working my way backwards with the goal of reusing as much of existing ModelReader as possible and potentially adding to it if there are missing pieces.

Would be nice to port pickModel to use the ModelReader... but I'm not sure I want to open that bucket right now.

Current state is very much a draft still, but I appreciate the early feedback.

javagl:
> I'm currently working on merging the capabilities of ModelMeshUtility and ModelReader (see my commit a few minutes ago). My initial mistake was to base my implementation on the pickModel (which I assume was written before we had the ModelReader).

It's not necessary to call this a "mistake". I certainly took ... inspiration...  from `pickModel` as well at some places. 

(If there is a "mistake", then it is that `pickModel` is a 320-line-function that calls a 60-line-function that has 10 parameters and _zero_ documentation, and overall is doing ~10 things at once. I could summarize that. It would sound even more dismissive. But to emphasize that: I'm not blaming whoever wrote `pickModel` originally. It may also have been gathered by copy-pasting a bunch of snippets, under time pressure, to get a feature done. Good software needs _time_, that's all I'm saying)

> Currently working my way backwards with the goal of reusing as much of existing ModelReader as possible and potentially adding to it if there are missing pieces.

I certainly don't want to "urge" you to do that in any way (if it causes too much effort or difficulties). But the goal of the `ModelReader` was to make something like the geometry extraction _easier_. And if it does not achieve that goal, I'd like to know how it could be improved. (I'm aware of _some_ points that are 'not ideal', but may not have all usage scenarios on the radar). I hope that this generic `readAttributeAsTypedArray` covers most needs. In the `ModelMeshUtility`, eventually, you _would_ need to add `readNormalData` and `readTexCoordData`, then already having to handle `VEC2`, `VEC3`, `VEC4` with byte/short normalizations and different quantization methods and whatnot. 

> Would be nice to port pickModel to use the ModelReader... but I'm not sure I want to open that bucket right now.

One **important** point for "aligning" things and avoiding redundancies: I've seen that some of your implementations checked for 'WebGL2'. This was related to reading data from the GPU to the CPU, and that whole buffer-vs-typedArray topic. As far as I had seen in earlier PRs, you had some ~ `allowGeometryExtraction` flag that kept the data on the CPU. So I think that this WebGL2-check is not strictly necessary. (If this was intended to be future-proofing for cases where the data is on the GPU, we'd have to talk about details). 

---

Even if this check for "WebGL2" is necessary:

The methods received the `FrameState`. Have a look at the [`FrameState` ""class""](https://github.com/CesiumGS/cesium/blob/c9bf4df7dfa3fa9c293bab39218307d0882a38be/packages/engine/Source/Scene/FrameState.js). It has the following properties:
```
  this.context = context;
  this.commandList = [];
  this.panoramaCommandList = [];
  this.shadowMaps = [];
  this.brdfLutGenerator = undefined;
  this.environmentMap = undefined;
  this.sphericalHarmonicCoefficients = undefined;
  this.specularEnvironmentMaps = undefined;
  this.specularEnvironmentMapsMaximumLOD = undefined;
  this.mode = SceneMode.SCENE3D;
  this.morphTime = SceneMode.getMorphTime(SceneMode.SCENE3D);
  this.frameNumber = 0;
  this.newFrame = false;
  this.time = undefined;
  this.jobScheduler = jobScheduler;
  this.mapProjection = undefined;
  this.camera = undefined;
  this.cameraUnderground = false;
  this.globeTranslucencyState = undefined;
  this.cullingVolume = undefined;
  this.occluder = undefined;
  this.maximumScreenSpaceError = undefined;
  this.pixelRatio = 1.0;
  this.passes = {
  this.creditDisplay = creditDisplay;
  this.afterRender = [];
  this.scene3DOnly = false;
  this.fog = {
  this.atmosphere = undefined;
  this.verticalExaggeration = 1.0;
  this.verticalExaggerationRelativeHeight = 0.0;
  this.shadowState = {
  this.splitPosition = 0.0;
  this.frustumSplits = [];
  this.backgroundColor = undefined;
  this.light = undefined;
  this.minimumDisableDepthTestDistance = undefined;
  this.invertClassification = false;
  this.invertClassificationColor = undefined;
  this.useLogDepth = false;
  this.tilesetPassState = undefined;
  this.minimumTerrainHeight = 0.0;
  this.pickingMetadata = false;
  this.pickedMetadataInfo = undefined;
  this.edgeVisibilityRequested = false;
```

I think that the _name_ of a class should reflect the purpose and usage of that class. So `FrameState` should be renamed to `RandomStuffThatIsNeededSomewhere`. These are 45(!) completely unrelated properties, some of them pretty complex data structures. And they include highly problematic ones, like the `Context`. The [`Context` ""class""](https://github.com/CesiumGS/cesium/blob/488c5213f1b7040b860d9ab2db2699a8ad0607c1/packages/engine/Source/Renderer/Context.js) has the following properties:
```
  this._canvas = canvas;
  this._originalGLContext = glContext;
  this._gl = glContext;
  this._webgl2 = webgl2;
  this._id = createGuid();
  this.validateFramebuffer = false;
  this.validateShaderProgram = false;
  this.logShaderCompilation = false;
  this._throwOnWebGLError = false;
  this._shaderCache = new ShaderCache(this);
  this._textureCache = new TextureCache();
  this._stencilBits = gl.getParameter(gl.STENCIL_BITS);
  this._antialias = gl.getContextAttributes().antialias;
  this._standardDerivatives = !!getExtension(gl, ["OES_standard_derivatives"]);
  this._blendMinmax = !!getExtension(gl, ["EXT_blend_minmax"]);
  this._elementIndexUint = !!getExtension(gl, ["OES_element_index_uint"]);
  this._depthTexture = !!getExtension(gl, [
  this._fragDepth = !!getExtension(gl, ["EXT_frag_depth"]);
  this._debugShaders = getExtension(gl, ["WEBGL_debug_shaders"]);
  this._textureFloat = !!getExtension(gl, ["OES_texture_float"]);
  this._textureHalfFloat = !!getExtension(gl, ["OES_texture_half_float"]);
  this._textureFloatLinear = !!getExtension(gl, ["OES_texture_float_linear"]);
  this._textureHalfFloatLinear = !!getExtension(gl, [
  this._supportsTextureLod = !!getExtension(gl, ["EXT_shader_texture_lod"]);
  this._colorBufferFloat = !!getExtension(gl, [
  this._floatBlend = !!getExtension(gl, ["EXT_float_blend"]);
  this._colorBufferHalfFloat = !!getExtension(gl, [
  this._s3tc = !!getExtension(gl, [
  this._pvrtc = !!getExtension(gl, [
  this._astc = !!getExtension(gl, ["WEBGL_compressed_texture_astc"]);
  this._etc = !!getExtension(gl, ["WEBG_compressed_texture_etc"]);
  this._etc1 = !!getExtension(gl, ["WEBGL_compressed_texture_etc1"]);
  this._bc7 = !!getExtension(gl, ["EXT_texture_compression_bptc"]);
  this._textureFilterAnisotropic = textureFilterAnisotropic;
  this.glCreateVertexArray = glCreateVertexArray;
  this.glBindVertexArray = glBindVertexArray;
  this.glDeleteVertexArray = glDeleteVertexArray;
  this.glDrawElementsInstanced = glDrawElementsInstanced;
  this.glDrawArraysInstanced = glDrawArraysInstanced;
  this.glVertexAttribDivisor = glVertexAttribDivisor;
  this.glDrawBuffers = glDrawBuffers;
  this._vertexArrayObject = !!vertexArrayObject;
  this._instancedArrays = !!instancedArrays;
  this._drawBuffers = !!drawBuffers;
  this._clearColor = new Color(0.0, 0.0, 0.0, 0.0);
  this._clearDepth = 1.0;
  this._clearStencil = 0;
  this._defaultPassState = ps;
  this._defaultRenderState = rs;
  this._defaultTexture = undefined;
  this._defaultEmissiveTexture = undefined;
  this._defaultNormalTexture = undefined;
  this._defaultCubeMap = undefined;
  this._us = us;
  this._currentRenderState = rs;
  this._currentPassState = ps;
  this._currentFramebuffer = undefined;
  this._maxFrameTextureUnitIndex = 0;
  this._vertexAttribDivisors = [];
  this._previousDrawInstanced = false;
  this._pickObjects = new Map();
  this._nextPickColor = new Uint32Array(1);
  this.options = {
  this.cache = {};
```
Yup, that's 64 unrelated, complex, largely undocumented properties, and only one of them is the `gl` (i.e. the _actual_ context) that cannot be created without the `Canvas`. Nothing of this can sensibly be documented, used, maintained, tested, and certainly not be mocked for tests in CI. (I recently [tried it](https://github.com/CesiumGS/cesium/pull/12977/changes/ba7f4b4744ea9fe5505d7bda162ce400c9eb75ac#diff-08f99358f9d02317fca5cfda76fa9dd71e81026d69fb82f60ae5423987bda8c5L583), but had to give up...)

The point is: The methods did not need a `FrameState`. They only need a single, simple, `boolean` parameter - namely, `webgl2`. 

(And as I mentioned above: I hope that they don't even need that. If they need "something like that", we could try to find a better solution...)

donmccurdy:
I'm not familiar enough to have an opinion on most of the above, but — especially in a performance-oriented API, it might be ideal to return `Float64Array` rather than `Cartesian[]` for positions, and perhaps `Uint8Array` for colors. Then the caller can avoid those allocations if they aren't needed.

javagl:
Most of the `ModelReader` already operates on plain, flat typed arrays. (The main method is even called `readAttributeAsTypedArray`, because there are just too many types and shapes that "attribute data" could have). 

And again: It certainly has to be improved and extended, to be **that** universal tool that I would have liked it to be some years ago, and imagined it to be in a (hopefully not so distant) future.

About avoiding allocations: Some functions already accept an optional `result` parameter, but not all of them. In many cases, the caller doesn't have a clue what the result type is. And if there's a call like `readIndices(intIndices, byteTypedArray)` then there has to be some error handling or casting or whatnot. And if the caller feels the urge to do some `if (components.indices.componentType === ...) ... else` cascade to create the "right" array at each call site, that would defeat much of the purpose. 

> Float64Array rather than Cartesian[] for positions, and perhaps Uint8Array 

Right now, there's a bit of a mix of types and requirements. At some places, "things" (tuples) are stored or expected as `CartesianX` objects or `CartesianX[]` arrays. At other places, there already are flat `number[]` arrays (for single tuples _or_ arrays of tuples). 

The different `CartesianX` types may require some [special handling](https://github.com/alarkbentley/cesium/blob/2c28144fb9764f40c89db5a2e98ee7753e810987/packages/engine/Source/Scene/Model/ModelReader.js#L244), but that's not suuuch a big deal (there are (and always will be) only three of them). 

There already are conversion functions, back and forth, with the respective `(un)packArray` functions. Further generalizations _could_ be added, but may be adding some not-so-trivial abstractions, possible API complexity, or other allocation overheads. 

One thing - of which I'm aware that it can be 'dangerous'! - is ... what you might call a "flyweight pattern". For example, when something just needs a bunch of `Cartesian3` objects to iterate over (once), and there only is a `Float32Array`, then one can do things like [`createIterableCartesian3FromTypedArray`](https://github.com/alarkbentley/cesium/blob/2c28144fb9764f40c89db5a2e98ee7753e810987/packages/engine/Source/Scene/Model/ModelImageryMapping.js#L227), with zero allocations. But when a client _does_ store such an object, bad things can happen. There's no `ReadOnlyCartesian3`, unfortunately...

All these things are design considerations that one could think about, deeply and thoroughly (even if it started on a 'green field'). Retrofitting generalizations or abstractions to existing structures can be even more challenging.

packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -146,12 +146,29 @@ class ModelReader {
...
-    if (!defined(byteStride) || byteStride === defaultByteStride) {
+
+    const isDefaultStride =
+      !defined(byteStride) || byteStride === defaultByteStride;
+    const hasTypedArray = defined(attribute.typedArray);
+    const hasZeroOffset = !defined(byteOffset) || byteOffset === 0;
+
+    // Non-interleaved, in-memory, no offset — return as-is
+    if (isDefaultStride && hasTypedArray && hasZeroOffset) {
+      return attribute.typedArray;
```
javagl, line 161: I (think that) I see the point (from quickly skimming over the changes), namely to _not_ allocate a new typed array when the `attribute.typedArray` exactly reflects the desired data. 

I'll just add this comment as sort of a "note", to  verify - either during development, or during a review - what exactly we expect from this function.

Should it return the array, or always a copy?

Consider this:
```
// Fetch the attribute as array
const array0 = ModelReader.readAttributeAsTypedArray(atribute);

// Modify the array
array0[index] = 123.0;

// Fetch the attribute as array AGAIN
const array1 = ModelReader.readAttributeAsTypedArray(atribute);

// What does this print?
console.log(array1[index]);
```
The point is: The caller has no way to know whether the returned array **IS** the attribute array, or a copy. The output of the last lines will depend on whether the attribute _happened to have_ a `byteOffset:0`, and that's hardly something that the caller should have to know.

This specific change here is in the
`readAttributeAsRawCompactTypedArray`
`                ^ RAW               `
function for the "raw" typed array, and this function could be `private` (except for the specs). So maybe the case that it returns the actual typed array of the attribute can be dealt with elsewhere.

Otherwise, I think that the function should _always_ return a _new_ array, for consistency, _unless_ we find another way to clarify the behavior.

(Or at least add a warning:
` * The result may be the exact instance OR a copy - don't modify it!`
or so...)

javagl: (BTW: I have not yet done a full review of this PR - this was just something from a recent change that I noticed)

alarkbentley: I agree with you concern but just FYI this is also how readIndicesAsTypedArray already works. (and how it worked in the old pickModel implementation)
So if we decide to change this we should probably do it for indices also.

```
const existingIndices = primitiveIndices.typedArray;
if (defined(existingIndices)) {
   return existingIndices;
}
```

javagl: Yes. Until now, the `ModelReader` was a first shot at what I thought _could_ be useful, and it currently is used only in a very narrow context. For a broader usage, we'll have to _specify_ the behavior of these functions more rigorously, to avoid surprises. 

That said: The 'indices' already undergo a very special treatment. In contrast to the other attributes, they cannot be interleaved, they don't have a "byte offset" and such - they already _do_ contain "preprocessed" data, so to speak. This is happening in [`GltfLoader.loadIndices`](https://github.com/alarkbentley/cesium/blob/2f65307d05a76256f5b5f064b2775337c39cb2db/packages/engine/Source/Scene/GltfLoader.js#L1469), but with many other places affecting the final result, depending on the presence of draco compression or whether "outline rendering" is enabled. 

(And a bit more broadly: Some of this is only a side-effect of https://github.com/CesiumGS/cesium/issues/13052 . If we had a place in the code and a point in time where the indices and attributes are part of some 'loaded model', and available in an easily accessible form on the CPU, then we wouldn't need the `ModelReader` at all...)

packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -320,8 +336,8 @@ class ModelReader {
     const c = new Cartesian3();
     for (let i = 0; i < elementCount; i++) {
       Cartesian3.unpack(quantizedTypedArray, i * 3, c);
-      AttributeCompression.octDecodeInRange(c, normalizationRange, c);
-      Cartesian3.pack(dequantizedTypedArray, c, i * 3);
+      AttributeCompression.octDecodeInRange(c.x, c.y, normalizationRange, c);
```
javagl, line 343: Added a small comment at https://github.com/alarkbentley/cesium/commit/eaaedc38f2cbe25c54cafbcb26c06a2c261fe1ad#commitcomment-181748500 , but to be more visible: I'll have to allocate a bit more time for a thorough review.

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,259 @@
...
+    },
+  );
+
+  return result;
+};
+
+/**
+ * Groups unique vertex indices by feature ID from indices or vertex count.
+ * This is shared across all attribute extraction to avoid duplicate work.
+ * @private
```
javagl, line 104: _Note_: Some _specific_ aspects of this may become obsolete, depending on whether or how the API is changed for non-featureId-cases.

---

This should include proper documentation of the cases that the input parameters are `undefined` (as JDoc text, or part of `@param` JSDocs).
```
...
If the indices are undefined, then nothing will be done, so you wouldn't have to call this method in the first place.
If the featureIdData is undefined, then the map will map the value `0` to the set of all indices, in a very inefficient way.

@param {TypedArray|undefined} indices The indices
@param {TypedArray|undefined} featureIdData The feature ID data
```

The "very inefficient" part refers to the fact that it is checking for `defined(featureIdData)` for each index. This does not change. There could be some
```
if (!defined(featureIdData) {
  map.set(0, [...indices]);
  return map;
}
```
watchdog at the beginning. (Again: depending on how relevant 'featureIdData' will be in the future!)

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+  //>>includeEnd('debug');
+
+  const model = options.model;
+  const featureIdLabel = options.featureIdLabel ?? "featureId_0";
+  const extractPositions = options.extractPositions ?? true;
+  const extractColors = options.extractColors ?? false;
+  const result = new Map();
+
+  if (!model._ready) {
```
javagl: This is "fail silent". Users will receive an empty result and have no chance of figuring out why. It should at least be **VERY** boldly documented, or print a warning. Things like `Model.boundingSphere` even go so far to throw
```
      if (!this._ready) {
        throw new DeveloperError(
          "The model is not loaded. Use Model.readyEvent or wait for Model.ready to be true.",
        );
      }
```
But I would hesitate to recommend that. It's quirky, and the concept of "ready" as a whole doesn't make sense. (Some more RAII-like approach would be nice, but ... here we are, related to https://github.com/CesiumGS/cesium/issues/13052 ...)

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,259 @@
...
+        entry.colors = [];
+      }
+      result.set(featureId, entry);
+    }
+
+    for (const vertexIndex of vertexIndicesSet) {
+      // Positions need per-instance-transform duplication
+      if (defined(posData)) {
+        for (let t = 0; t < instanceTransforms.length; t++) {
+          readPosition(posData, vertexIndex, numPosComponents, scratchPosition);
```
javagl: I think that this `readPosition` can be pulled out of the `for` loop, **if** the `multiplyByPoint` below is changed to receive a dedicated result, as in 
```
... originalScratchPosition, transformedScratchPosition
```
so that the original position only has to be read _once_ from the position data (and the same position is then transformed with all transforms)

packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -805,6 +821,232 @@ class ModelReader {
...
+ *
+ * @param {object} runtimeNode The runtime node.
+ * @param {ModelSceneGraph} sceneGraph The model scene graph.
+ * @param {Model} model The model.
+ * @param {object} result An object with scratch matrices: { nodeComputedTransform: Matrix4, modelMatrix: Matrix4, computedModelMatrix: Matrix4 }.
+ * @returns {object} The result parameter, populated with the computed transforms.
+ *
+ * @private
+ */
+function computeNodeTransforms(runtimeNode, sceneGraph, model, result) {
```
javagl: Just a cross-link: This is basically the same of what I did in https://github.com/CesiumGS/cesium/pull/11830/changes#diff-c25ab54b614865d4ba17f0fbc27717f5cbaa07931f69deb0b83b7d0daa1d3505R235 

(Unfortunately, that PR had to be abandoned, and the "revival" at https://github.com/CesiumGS/cesium/pull/12658 probably counts as "abandoned" as well in the meantime. The model picking is still highly inefficient. Maybe we can "revive the revival" after _this_ PR is merged, given that it now _does_ contain building blocks like this `computeNodeTransforms`, which should have existed right from the beginning...)

So a 👍  for that.

packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -805,6 +821,232 @@ class ModelReader {
...
+  computedModelMatrix,
+  nodeComputedTransform,
+  modelMatrix,
+) {
+  const transforms = [];
+  const node = runtimeNode.node;
+  const instances = node.instances;
+
+  if (defined(instances)) {
+    const transformsCount = instances.attributes[0].count;
```
javagl: The `ModelReader` was only a stub when I created it, and will require further iterations to be "complete". Specifically: It currently still carries a comment
`NOTE: All this does not properly handle MATn types.`

And I think that these 'transforms' that are extracted here could/should eventually be handled with the `readAttributeAsTypedArray` as well, **when** (not 'if', but 'when') it supports `MAT4` attributes.

packages/engine/Source/Scene/Model/pickModel.js
```diff
@@ -189,144 +98,87 @@ export default function pickModel(
...
         const i2 = indices[i + 2];
 
         for (const instanceTransform of transforms) {
-          const v0 = getVertexPosition(
-            vertices,
-            i0,
-            offset,
-            elementStride,
-            quantization,
+          readPosition(vertices, i0, numPosComponents, scratchV0);
```
javagl: **Note**: I have _not_ (yet) thoroughly reviewed or explicitly tested the changed in `pickModel.js`. The diff-view here on GitHub is scattered. I looked at the resulting code, ad it looks **much** simpler and cleaner than before. So it should be possible to review (only) the new state of the code with reasonable effort. 

But there are some aspects of `pickModel` that are the reason why I'd like to spend a bit more time for a review here. These aspects are related to https://github.com/CesiumGS/cesium/pull/12658 . It _looks_ like this refactoring was essentially done here as well, but I'd like to check the details. Specifically: The handling of the 'vertical exaggeration' is a bit tricky, and I'll have to _explicitly_ and _dedicatedly_ review and test this.

In the meantime and for now:

I think that the three `readPosition` calls can be pulled out of the loop, storing the results in some `scratchVX_input`, and then change the calls below to
`multiplyByPoint(..., scratchVX_input, scratchVX_output)
so that the positions only have to be read _once_, and are then transformed with all matrices.

javagl (commented):
There are some inlined comments. Some of them may become obsolete, depending on whether/how the 'High-level' comments are addressed.


### High-level:

---

> Despite the fact that the current implementation is synchronous the introduced API endpoint is exposed with an async handle to enable future non-breaking improvements such as WebWorkers or async GPU readback.

Big +1 for that. The only way to make a future-proof API is to make everything `async`. This is obscure and inconvenient, but the harsh truth for JavaScript.

---

I think that the current API is very constrained. That's not necessarily a bad thing. The API should do exactly what it is supposed to do, and not be over-generalized to accidentally offer half-baked functionalities. But in its current form, the API can hardly evolve. Adding new extractions is hardly possible without considerable changes of the structures.

**Extracted properties:**

The `GeometryResult` currently has the `positions` and `colors`. It is very reasonable to assume that people will want to extract more - mainly, the `normals`. (A very recent case: https://community.cesium.com/t/how-to-align-glb-models-to-a-gltf-tileset-surface-picking-height-and-surface-normals/45632 - someone wanted to pick something, and align a model to the picked point, based on its surface normal).

There are some options that one could consider. One could add a `normals` property to that output. And one could consider adding a `texCoords` property to that output. Eventually, it could be _very_ desirable to have the possibility to access arbitrary [`EXT_structural_metadata` property attributes](https://github.com/CesiumGS/glTF/blob/d757b9c77fe86b5a63f6514aa24f3de801effa7e/extensions/2.0/Vendor/EXT_structural_metadata/README.md#property-attributes).

At this point, one could consider a return type like this
```
class GeometryResult {
  attributeNames: string[]; // contains strings like `POSITION` and `NORMAL` or `_MY_PROPERTY`
  attributeValues: Map<string, Array>; // Contains the values - see notes below!
  attributeTypes: Map<string, ...>; // See notes below
}
```
Of course, the attribute values would be _untyped_. They could contain `Cartesian3[]` or `Color[]` or `Matrix2[]`. In the spirit of the API design mantra that "simple things should be easy, and complex things should be _possible_", one could consider _typed_ convenience functions for the common attributes:
```
class GeometryResult {
  ...
  
  getPositions() : Cartesian3[] | undefined {
    const a = this.attributeValues("POSITION");
    if (defined(a)) { return undefined; }
    return a as Cartesian3[];
  }
}
```
For other attributes, the relevant information could be stored in the `attributeTypes`, which would include the `type` (like `VEC3`) and `componentType` (like `UINT16`) in one form or the other.

The limitation of the return type is pretty much tied to the `options` and the `extractPositions` and `extractColors` flags. A pragmatic generalization would be to offer some
`options.extractedProperties = ["POSITION", "_MY_PROPERTY" ];`
to handle that. (The caller can not know beforehand which properties are there, but that's a different topic - the caller doesn't know about the presence of the `colors` either...)

**Extracted structure:**

I think that I do understand the intention behind grouping the extracted geometry by the feature ID. But I don't understand the intention behind doing that _by default_. I'll have to check some details, but think that it should be relatively easy to make this an optional, second step. _Very_ roughly (somewhat brainstorming): 

The `extractAttributesFromPrimitive` receives the primitive, the feature ID label, and the `Map` to store the grouped results. It should be fairly easy to split this up into two functions:
- `extractAttributesFromPrimitive` (which, I think, could actually return a `GeometryResult`!)
- `partitionGeometryResult`

The latter could _receive_ a `GeometryResult` and the feature ID data, and return a `Map<number, GeometryResult>` that contains the clustered/grouped geometry. This would allow differentiating between the case where the user wants the geometry of _one feature_, and the case where the user just wants the whole geometry of the picked object. The latter would be useful for the cases where the input data is not even clustered by a feature ID (e.g. point clouds without batch tables).


### Tests


I tested the new Sandcastle, and it works nicely, as intended.

---

Loading a non-batched point cloud
```
const tileset = await Cesium.Cesium3DTileset.fromUrl(
  "../../SampleData/Cesium3DTiles/PointCloud/PointCloudRGB/tileset.json",
);
```
omitting the feature ID handling
```
  //if (!(pickedFeature instanceof Cesium.Cesium3DTileFeature)) {
  //  return;
  //}

  // Highlight the selected feature
  //selectedFeature = pickedFeature;
  //selectedFeature.color = highlightColor;
```
throws
```
Error
    at new DeveloperError (http://localhost:8081/packages/engine/Build/Unminified/index.js:7460:11)
    at Check.defined (http://localhost:8081/packages/engine/Build/Unminified/index.js:7496:11)
    at AttributeCompression.dequantize (http://localhost:8081/packages/engine/Build/Unminified/index.js:29414:17)
    at _ModelReader.readAttributeAsTypedArray (http://localhost:8081/packages/engine/Build/Unminified/index.js:105275:59)
    at extractAttributesFromPrimitive (http://localhost:8081/packages/engine/Build/Unminified/index.js:125155:39)
    at http://localhost:8081/packages/engine/Build/Unminified/index.js:125084:7
    at _ModelReader.forEachPrimitive (http://localhost:8081/packages/engine/Build/Unminified/index.js:105947:9)
    at ModelGeometryExtractor.getGeometryForModel (http://localhost:8081/packages/engine/Build/Unminified/index.js:125080:23)
    at Model3DTileContent.getGeometry (http://localhost:8081/packages/engine/Build/Unminified/index.js:125352:41)
    at <anonymous>:73:37
```
This _may_ not be an issue of this PR. It _may_ be caused by the fact that the `PntsLoader` does not assign the `count` property to attributes, _except_ for the "POSITION" attribute, as of https://github.com/CesiumGS/cesium/blob/05923d23a6c42e82ba50fb40eb0a64567ac1a99b/packages/engine/Source/Scene/Model/PntsLoader.js#L525

I think that it could and should assign this property, **iff** it can be verified that this `count` is always present and correct.


---

**After** pragmatically inserting the proper `count` values:

For a point cloud, there are no indices, and that will crash
```
ModelReader.js:684 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'typedArray')
    at _ModelReader.readIndicesAsTypedArray (ModelReader.js:684:46)
    at extractAttributesFromPrimitive (ModelGeometryExtractor.js:217:33)
    at ModelGeometryExtractor.js:76:7
    at _ModelReader.forEachPrimitive (ModelReader.js:898:9)
    at ModelGeometryExtractor.getGeometryForModel (ModelGeometryExtractor.js:72:15)
    at Model3DTileContent.getGeometry (Model3DTileContent.js:233:33)
    at <anonymous>:86:37
    at cancelMouseEvent (ScreenSpaceEventHandler.js:290:9)
    at handleMouseUp (ScreenSpaceEventHandler.js:312:5)
    at handlePointerUp (ScreenSpaceEventHandler.js:863:5)
```

---

When loading composite data...
```
const tileset = await Cesium.Cesium3DTileset.fromUrl(
  "../../SampleData/Cesium3DTiles/Composite/Composite/tileset.json",
);
```
then vertices will be shown only when clicking the lower left box (not for other boxes). I haven't looked at how to resolve this. Maybe what is currently done with the "feature ID" will have to be done with some "instance ID/index" instead. This may become obsolete, depending on possible changes of the feature ID handling.


---

When loading vector data...
```
});
const geocoder = viewer.geocoder.viewModel;
geocoder.searchText = "Vienna, Austria";
geocoder.flightDuration = 0.0;
geocoder.search();

const tileset = await Cesium.Cesium3DTileset.fromIonAssetId(5737);
```
the clicking will cause
```
Uncaught (in promise) TypeError: content.getGeometry is not a function
    at <anonymous>:86:37
    at cancelMouseEvent (ScreenSpaceEventHandler.js:290:9)
    at handleMouseUp (ScreenSpaceEventHandler.js:312:5)
    at handlePointerUp (ScreenSpaceEventHandler.js:863:5)
    at HTMLCanvasElement.listener (ScreenSpaceEventHandler.js:53:5)
```
This could be avoided with some `instanceof Model3DTileContent` check at the client side, but that's probably not intended or desired.

I think whatever was supposed to be accomplished with the `DeveloperError.throwInstantiationError()` simply does not work, and that all ""implementations"" of the `Cesium3DTileContent` ""interface"" will have to explicitly implement this function (making the whole ""interface"" obsolete after all...)

mzschwartz5:
It was brought to my attention that I'm working on something somewhat similar, [here](https://github.com/CesiumGS/cesium/pull/13400/changes#diff-708d1b747bbadb8619378ae52a50e768a38e28aaa7ccb4b97131c872307e1494). 

You can read in that PR's description what the overall goal is there, but that specific link is to a general interface I'm introducing to access geometry data. The idea would then be to implement version of that interface for various geometry-containing classes (`Primitive`, `Model`, or - in your case - `Cesium3DTile`). I think what I'm building solves a general problem in one place rather than specific problems in multiple, and could maybe be reused in your work.

Note: my PR is still in draft. That interface I'm building will have more methods added to it beyond just getting/setting vertices. E.g. I could see it containing general methods for accessing generic vertex attributes.

alarkbentley:
@javagl Could you take another round at this when you get some time to spare? I did some major refactoring based on your previous comments which all made sense.

Summary of the updates:

**ModelReader**

* forEachPrimitive now support readback of VEC3 instance transforms (TRANSLATION, ROTATION, SCALE)
* Added forEachPrimitive options.perInstanceFeatureIds to toggle instance featureid extraction

**pickModel**

* Cache readPosition results in scratch
* Fixed invalid unit-test "returns position of intersection with instanced model" 
--> Previouly did not read VEC3 transforms

**ModelGeometryExtractor**

* Remove default featureid grouping (getGeometry now return a list instead of a map)
* Support Instanced Geometry (w. InstanceAttributeSemantic.FEATURE_ID)
--> Fixes Cesium3DTiles/Composite/Composite/tileset.json
* Support Non Indexed Geometry (eg. Point Cloud)
--> Fixes Cesium3DTiles/PointCloud/PointCloudRGB/tileset.json
* Opt-in for extract featureid
* Support extracting Indices (extractIndices=true)
* Replaced extractPositions, extractColors flags with a generalized attributes: ["POSITION", "COLOR", ...]

**GeometryResult**
* Generalized GeometryResult (attributeNames, attributeValues, attributeTypes w. helpers eg. getPositions)

**Cesium3DTileContent**

* Moved the getGeometry API to Cesium.Model to reduce API surface area
--> Vector data no longer relevant

composite
instanced_orientation
point_cloud

alarkbentley:
> It was brought to my attention that I'm working on something somewhat similar, [here](https://github.com/CesiumGS/cesium/pull/13400/changes#diff-708d1b747bbadb8619378ae52a50e768a38e28aaa7ccb4b97131c872307e1494).
> 
> You can read in that PR's description what the overall goal is there, but that specific link is to a general interface I'm introducing to access geometry data. The idea would then be to implement version of that interface for various geometry-containing classes (`Primitive`, `Model`, or - in your case - `Cesium3DTile`). I think what I'm building solves a general problem in one place rather than specific problems in multiple, and could maybe be reused in your work.
> 
> Note: my PR is still in draft. That interface I'm building will have more methods added to it beyond just getting/setting vertices. E.g. I could see it containing general methods for accessing generic vertex attributes.

@mzschwartz5 Thanks for bringing this to my attention. This work certainly has some potential overlap.
Could you please review my latest changes in the PR to see how we could align? I moved the API surface from Cesium3DTile into the Model. Both the ModelReader (general utilities for reading primitive data) and the ModelGeometryExtractor (which is basically a GeometryReader for Model if I understand your PR correctly) could probably be generalized for your use-case.

packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -1049,4 +1038,134 @@ function getInstanceTransforms(
...
+        )
+      : Quaternion.IDENTITY;
+
+    const scale = hasScale
+      ? new Cartesian3(
+          scaleAttribute.typedArray[i * 3],
+          scaleAttribute.typedArray[i * 3 + 1],
+          scaleAttribute.typedArray[i * 3 + 2],
+        )
+      : new Cartesian3(1, 1, 1);
```
javagl, line 1158: This could be `Cesium.Cartesian3.ONE`

packages/engine/Source/Scene/Model/ModelReader.js
javagl: This is a broader question, and applies to many existing parts of the code (i.e. it's not clear whethr something is a copy or not). While I think that this _can_ be a source of bugs and confusion, and _should_ be addressed, it's not something that can sensibly be done here. Resolving this for now.

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
javagl: I think this is obsolete for now (just going through the most recent changes)

packages/engine/Source/Scene/Model/pickModel.js
javagl: I've seen that this has been addressed (but I still have to review `pickModel`)

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,383 @@
...
+                i,
+                attr.numComponents,
+              ),
+            );
+          }
+        }
+      }
+      // Instance feature ID fallback: if feature IDs were requested but no
+      // per-vertex feature ID attribute was resolved, use the instance's ID.
+      // TODO: How to handle the case where we have both FeatureId as attribute and instance?
```
javagl, line 355: That's a good point, but could be stated more broadly (in a form that applies not only to the geometry extraction, but other places as well). 

The handling of the https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#featureIdLabel is very "global" in its nature, and not applicable for tilesets that contain models with different structures (so this may already affect things like picking and custom shaders). 

The question could be phrased like that: What happens when I3DM data refers to a glTF that contains one of the metadata extensions? How will the concept of 'instantiation' be applied to the metadata? 

What _did_ come up in the past: There could be an I3DM that refers to glTF with `EXT_mesh_gpu_instancing`. This would mean to create _instances of instances_. I think that `cesium-native` actually does handle this. I'm pretty sure that CesiumJS does _not_ handle this. (And I think that I tested it once, but would have to dig out the test model). 

**Note**: In any case, all these questions are independent of this PR! 

(But we should consider summarizing some of these questions, roughly aiming at 'converting this TODO into an issue', just to keep track of it)

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,383 @@
...
+        semantic,
+        setIndex,
+      );
+    }
+
+    if (!defined(attribute)) {
+      continue;
+    }
+
+    const typedArray = ModelReader.readAttributeAsTypedArray(attribute);
```
javagl, line 258: Only a note (some additional detail in the main review comment): 

The `readAttributeAsTypedArray` will crash because `count` for the COLOR is still `undefined` in the "PointCloudRGB" test case. 

Locally, this could be ""fixed"" with a hack like this:
```
    if (!defined(count) && defined(attribute.count)) {
      count = attribute.count;
    } else if (defined(count) && !defined(attribute.count)) {
      console.log("XXX Setting count - is this OK??? ", count);
      attribute.count = count;
    }
```
_before_ the `readAttributeAsTypedArray` call. The proper solution is still TBD.

javagl (commented):
> refactoring based on your previous comments which all made sense.

I like to be challenged. This PR was started with a certain idea (usage scenario and intended functionality) in mind. I reviewed it, trying to be neutral and objective. But I cannot rule out that there are 1. wrong presuppositions on my side or 2. things beyond what I have no the radar. If any of my comments seems to be questionable, causes concerns, or seems "unnecessary" in any way, just point that out. 
(Yes, _this_ comment may have been unnecessary - but I'd like to be sure...)

### The baseline: 

The specs all pass. I looked at them, and they look reasonable in terms of structure and coverage. 
(Reviewing them _really_ thoroughly is difficult. The variety of possible models (in terms of combinations of ~"things" that they can contain) is _huge_. Some of the _existing_ structures are a bit underspecified. And there are some things that I didn't have in my "working memory". For example, I've never dealt with that `positionalLabel`, and have to assume that the way how it is tested makes sense). 


### Previous review:

I've seen some of the previous comments being addressed, and things that go beyond that, for example:

- Cleanup and generalization for pulling out transform extraction into functions https://github.com/CesiumGS/cesium/pull/13330/changes/5a2ddef002999c4c83c639785d6181c8dd75eb5f
- Addressing the issue of `getGeometry` in `Cesium3DTileContent` with empty default implementations in  https://github.com/CesiumGS/cesium/pull/13330/changes/81e639867a1a7c386e9ffd57043aa98e6855038e
- Generalized `GeometryResult` in https://github.com/CesiumGS/cesium/pull/13330/changes/b80595106978b44cb92cee3998f21d7ef27a0cb6 (more on that below)

The comment about checking `model._ready` from https://github.com/CesiumGS/cesium/pull/13330#discussion_r3051348929 is still pending. I think that a `oneTimeWarning` could be fine, but don't have a strong opinion on that (beyond that it should not fail silently).


### Test of current state:

I did tests, similar to the snippets from the previous review, and most of the issues have been resolved. 

#### About that `count` being `undefined`

One thing that surprised me a bit: Your last screenshot showed what appears to be the `PointCloudRGB`. When I tried it out, by just changing the tileset to
```
const tileset = await Cesium.Cesium3DTileset.fromUrl(
  "../../SampleData/Cesium3DTiles/PointCloud/PointCloudRGB/tileset.json",
);
```
it still "crashed" for me (no message - error thrown in async handler function). This was still related to that `count` being `undefined` after reading the PNTS data.

I added an inlined comment in https://github.com/CesiumGS/cesium/pull/13330/changes#r3105104480  

We'll have to check whether it's OK to modify the `count`, or whether the `PntsReader` should set it immediately while reading the data. (I think that it should - to be verified!)

#### About vertical exaggeration

One thing that was still on my TODO list from the previous review was to check how all this (including the changes in `pickModel`) works with 'vertical exaggeration'. I did a _quick_ test for this, by applying non-zero vertical exaggeration to the model from the `dev` sandcastle. And it _still_ picked the right elements when clicking the exaggerated geometry. So that seems to be fine


### Further thoughts on the current state

Related to the refactoring of the `GeometryResult`: 

I think that the separation of "extracting geometry" and "grouping/filtering by feature ID" makes sense, in terms of API design and the genericity (and general applicability) of the underling extraction functionality.  So there now is this `getGeometryResultByPredicate` in the sandcastle. And I wondered whether ~"something like this" could be offered on the API level.

The reason for wondering: _Initially_, you did this "grouping by ID" as an _inherent_ part of the extraction. So I have to assume that there is a strong use-case / demand for that. 

And to be _slightly_ more specific, I quickly drafted what I think that this could _roughly_ look like:
```javascript
function getGeometryResultByPredicate(geometryList, predicate) {
  const result = new Cesium.GeometryResult();
  // Note: This assumes that all geometries have the 
  // same structure (e.g. the same type for an attribute
  // with a certain name. That should be OK, I guess...
  for (let i = 0; i < geometryList.length; i++) {
    const geometry = geometryList[i];
    for (const key of geometry.attributeNames) {
      const values = geometry.attributeValues.get(key);
      if (!Cesium.defined(values)) {
        continue;
      }
      if (!result.attributeNames.includes(key)) {
        result.attributeNames.push(key);
        const type = geometry.attributeTypes.get(key);
        result.attributeTypes.set(key, type);
      }
      let resultValues = result.attributeValues.get(key);
      if (!resultValues) {
        resultValues = [];
        result.attributeValues.set(key, resultValues);
      }
      for (let j = 0; j < geometry.count * geometry.instances; j++) {
        if (predicate(geometry, j)) {
          resultValues.push(values[j]);
        }
      }
    }
  }
  return result;
}

// Receives several GeometryResult instances, and creates
// ONE GeometryResult from everything that has the given
// feature ID. This could be a public function somewhere...
function getGeometryResultByFeatureId(geometryList, featureId) {
  const geometry = getGeometryResultByPredicate(
    geometryList,
    (geometry, index) => {
      if (Cesium.defined(featureId)) {
        const featureIds = geometry.getFeatureIds();
        if (Cesium.defined(featureIds)) {
          return featureId === featureIds[index];
        }
        return false;
      }
      return true;
    },
  );
  return geometry;
}
```

The `getGeometryResultByFeatureId` is what could be "public" in one way or another (maybe a static method on `GeometryResult` or so - TBD). 

In the sandcastle, it can be called as
`const geometry = getGeometryResultByFeatureId(geometryMap, pickedFeature.featureId);`

The `getGeometryResultByPredicate` is the core of the implementation. 

(And I think that this separation makes sense: One could easily imagine implementations of that 'predicate', _beyond_ checking a feature ID. Brainstorming: Think about some
`getGeometryResultByColor(list, Color.RED);`
or
`getGeometryResultInBoundingBox(list, boundingBox);`
and things like that. This could be really powerful, and offer some interesting options in the future...)

That function just takes the list of geometry results, and extracts a _single_ `GeometryResult`, based on the predicate. The point is that the returned thing also **is a** `GeometryResult`, with the structure that already is part of the API, and the same genericity.

As mentioned in the draft `// comments`: There are assumptions hidden in that, about the inputs having ~"the same structure". And a proper implementation would need additional guardrails and checks. 

But... do you think that ~"something like that" could make sense, to simplify the case of the ID-based extraction?

packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -1052,9 +1052,9 @@ function getInstanceTransforms(
         Matrix4.multiplyTransformation(modelMatrix, transform, transform);
       } else {
         Matrix4.multiplyTransformation(
-          transform,
           computedModelMatrix,
```
javagl, line 1055: Small addendum: I have not yet (mathematically) thought this through. From the tests, it seems to make sense. But some of the transform orders related to `Model` are tricky in general.

alarkbentley:
> I like to be challenged. This PR was started with a certain idea (usage scenario and intended functionality) in mind. I reviewed it, trying to be neutral and objective. But I cannot rule out that there are 1. wrong presuppositions on my side or 2. things beyond what I have no the radar. If any of my comments seems to be questionable, causes concerns, or seems "unnecessary" in any way, just point that out. (Yes, _this_ comment may have been unnecessary - but I'd like to be sure...)

Thanks, I got you. Your feedback so far has all made perfect sense to me but I'll definitely intervene or make modfications if I don't agree or it adds to much scope. I want to make sure these additions adds general value to Cesium and not only solves my immediate use-case. But also of course we have to limit the scope at some point.

I've done some more updates since last, I think we are ready for another review. Here is a summary:

**ModelReader**

* Added support for FeatureIdImplicitRange
  - Added method `readImplicitRangeAsTypedArray`
* Fixed the minor Cartesian3.ONE

**Model**

* If not model._ready now throws DeveloperError https://github.com/CesiumGS/cesium/pull/13330#discussion_r3051348929
  - Decided to go with the DeveloperError to align with how its handled elsewhere in Model
* If getGeometry options.attributes is undefined, we now extract all attributes by default
  - Convenient if you do not know the contents/attribute-keys beforehand

**ModelGeometryExtractor**

* Ignore attributes without defined count
  - PointClouds will no longer crash but also not support vertex color. This will be solved once we add count to the color attributes. (deemed out of scope)
* Per-instance feature ids always has precedence over per-primitive.
  - Found this comment in the [docs](https://github.com/CesiumGS/cesium/blob/9751fb1f5e6476edbc0b378e5ec7f73df715c2f8/packages/engine/Source/Scene/Model/Model.js#L171) _"If both per-primitive and per-instance feature IDs are present, the instance feature IDs take priority."_

**GeometryResult**

* Added convenience methods `GeometryResult.prototype.filter` and `GeometryResult.getGeometryResultByFeatureId` 
  - Check how its used in the sandcastle demo

javagl:
> PointClouds will no longer crash but also not support vertex color. This will be solved once we add count to the color attributes.

Yes, this _should_ not have to be done as part of this PR (I'd rather consider it as a dedicated issue. It might be a trivial fix, by just adding that `count` there. But... experience shows that everything there may have the most obscure side-effects, and I'd like to get a better idea for why it was not added for those attributes in the first place).

I'll try to do another review pass ASAP (currently scheduling things back and forth, but should be doable in the next few days)

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,492 @@
...
+ * @property {number} [setIndex] The set index, if applicable.
+ * @private
+ */
+
+/**
+ * Normalizes the caller-supplied options into a uniform array of
+ * {@link AttributeRequest} objects.
+ *
+ * @param {object} options The options passed to {@link ModelGeometryExtractor.getGeometryForModel}.
+ * @returns {AttributeRequest[]} The normalized attribute requests.
```
javagl: Nitpick: In the return type here (and the parameter type for `extractAttributesFromPrimitive`), the type is now `AttributeRequest[]|undefined`, with the meaning of 
` * If this is undefined, then all attributes will be extracted`

javagl (approved):
> - Added support for `FeatureIdImplicitRange` 

I didn't sooo actively consider this, because this was only part of an early/intermediate state of the "feature ID definitions". (Some entry point for the considerations around that is https://github.com/KhronosGroup/glTF/pull/2082#discussion_r750698295 ). In the latest state of the extension specification, these "implicit ranges" can no longer appear. But it's good to see that they are handled for backward compatibility, for the case that they appear in older assets.

---

> If not model._ready now throws DeveloperError

That's fine, in alignment with what things like `boundingSphere` are doing.

---

> If getGeometry options.attributes is undefined, we now extract all attributes by default

That makes perfect sense. Otherwise, there would have to be some querying mechanism (or lots of client-side boilerplate code) to figure out what the requests should be.

(Nitpick about the JSDoc inlined)

---

> Ignore attributes without defined count
> 
> [...] This will be solved once we add count to the color attributes. (deemed out of scope)

It probably makes sense to ignore attributes without `count` in any case. (It should never be `undefined`, but ... it shouldn't crash when it is). 

But I checked internally, and for the case of PNTS, the `count` should probably simply be set on all attributes. I opened https://github.com/CesiumGS/cesium/pull/13427 (if/when this is merged into main, it could become part of this PR, transparently, when _this_ PR is updated with the latest `main` state)

---

> Per-instance feature ids always has precedence over per-primitive

Good catch (of a pretty subtle hint...). That should make the behavior consistent with other parts of the code.

---

> Added convenience methods ...
> Check how its used in the sandcastle demo
 
I think this nicely reduces the boilerplate code for users for this case (and they 'don't have to know so much about implementation details). 

I'm tempted to try out what's possible with that, like that 
`filter = containedIn(someBoundingVolume)` 
and so, but will have to allocate some time for that.

---

I'm usually _very_ hesitant to 'Approve' larger changes, because there always are thousands of things (corner cases, rare model configurations and whatnot) that could be investigated. But this one looks pretty mature for me now.

javagl:
Minor follow up: https://github.com/CesiumGS/cesium/pull/13427 was just merged. This should not directly affect this PR, but when `main` is merged in here, things like that `PointCloud...` should now work better with the `count` values being present.

javagl:
An aside: Given the complexity of the possible inputs here, and the _huge_ variety of ~"features" that could be combined in a model (and which are difficult to cover in all depth, like "instanced metadata" and whatnot), I could imagine that it could make sense to tag the classes (that are not tagged as `@private`) with some hint
`@experimental This feature is subject to change without Cesium's standard deprecation policy.`

alarkbentley:
> > * Added support for `FeatureIdImplicitRange`
> 
> I didn't sooo actively consider this, because this was only part of an early/intermediate state of the "feature ID definitions". (Some entry point for the considerations around that is [KhronosGroup/glTF#2082 (comment)](https://github.com/KhronosGroup/glTF/pull/2082#discussion_r750698295) ). In the latest state of the extension specification, these "implicit ranges" can no longer appear. But it's good to see that they are handled for backward compatibility, for the case that they appear in older assets.

So the reason I found this was during testing with the models in this demo:
https://sandcastle.cesium.com/?id=3d-models

The "Instanced Box" in this sandcastle is using `FeatureIdImplicitRange`

alarkbentley:
@javagl Thanks for your support with this PR, I'm good to merge from my side.

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
javagl: This is resolved via https://github.com/CesiumGS/cesium/pull/13330/commits/fb87cf8a96202e15746e9aaaf680a92156150f96

packages/engine/Source/Scene/Model/ModelReader.js
javagl: I _still_ haven't verified this manually, with more complex model configuations. But I assume that is correct, based on the specs.

javagl:
Despite the huge variety of the possible input configurations and "flavors" of `Model` objects, and even though I'm pretty sure that sooner or later, there will be _some_ rare, obscure case where ~"something" doesn't work as expected (mainly because many of the possible "flavors" are not thoroughly documented or constrained to begin with), I think that this PR looks very sound and clean. I'll ask a few people to have another look, but if nobody has strong concerns, will likely merge it soon.

packages/engine/Source/Scene/Model/ModelGeometryExtractor.js
```diff
@@ -0,0 +1,494 @@
...
+ * @param {number} [setIndex] The set index, if applicable.
+ * @returns {string} The attribute key.
+ * @private
+ */
+function getAttributeKey(semantic, setIndex) {
+  if (defined(setIndex)) {
+    return `${semantic}_${setIndex}`;
+  }
+  return semantic;
+}
```
mzschwartz5, line 111: This feels like a function that should live in `VertexAttributeSemantic` (or `InstanceAttributeSemantic`) or some other shared place. Same with `parseAttributeKey`

mzschwartz5:
I would personally prefer to wait a couple weeks to merge this, if there's not immediate urgency to. I feel that there's going to be significant overlap with the mesh editing work (which has some mid-May and end-of-May deadlines, so we it won't be a terribly long wait). I'd like to see better where that overlap is going to be, but I'm almost sure that the concept of `GeometryResults` - while nicely general and great for one-off geometry reading - isn't going to be general enough for mesh editing, and will just end up being replaced by the `GeometryAccessor` class I've introduced. 

The nice thing about `GeometryAccessor` is that ownership of the CPU data still belongs with the geometry owning class. The user just reads (or writes!) the data as needed, _using_ the `GeometryAccessor`, without ever explicitly having access to typed arrays or allocating lots of `Cartesian3`s up front. Moreover, the lifetime of that data is more tightly controlled via the concept of access sessions. 

To be clear - I think what this PR introduces is very useful (for mesh editing as well, when it gets there). I just don't want to introduce a public API and then deprecate it next release, potentially. 

Unrelated, I'm a little intimidated by the scope of this PR (`+5,638 -322`). It's quite difficult to be confident in one's review for something so large. And I find myself asking, "do we really need more utility type logic + classes to support Model?"... I just feel like we're digging the `Model` hole deeper.

alarkbentley:
@mzschwartz5 That makes sense to me. I was aiming to use this for the ongoing development for generating clip polygons as mentioned under "Motivation" in the PR description.

Would an ok compromise be to break out the non API facing parts and bug fixes (eg. ModelReader, pickModel) in a separate PR that we can target to merge for 1.141? Then we postpone the ModelGeometryExtractor and GeometryResults until we are ready to align them with your work.

In the mean-time I can move the ModelGeometryExtractor externally for my PoC.

> Unrelated, I'm a little intimidated by the scope of this PR (+5,638 -322). It's quite difficult to be confident in one's review for something so large. And I find myself asking, "do we really need more utility type logic + classes to support Model?"... I just feel like we're digging the Model hole deeper.

Worth mentioning however is that +4000 of these lines are just unit-tests.

mzschwartz5:
> Would an ok compromise be to break out the non API facing parts and bug fixes (eg. ModelReader, pickModel) in a separate PR that we can target to merge for 1.141? Then we postpone the ModelGeometryExtractor and GeometryResults until we are ready to align them with your work.

> In the mean-time I can move the ModelGeometryExtractor externally for my PoC.

Yeah I think that's perfectly reasonable

alarkbentley:
@mzschwartz5 https://github.com/CesiumGS/cesium/pull/13433

Issue #2132, opened 2014-09-16
Vector Tiles


pjcozzi:
2D maps like Google, Apple, MapBox GL, and OL3 are moving to actual vectors for rendering vector data, instead of baking vector data into a raster tile.  This allows, for example, to make labels always face up even if south is up and to fade out labels so they never hit each other.

For now, this roadmap is a collection of resources on this topic.
- [Google Maps + WebGL = MapsGL](https://docs.google.com/presentation/d/1pL-OZnsWIKWcXflfJV8MPH3kKBehn0Mlx6-cdjmP3A0/edit#slide=id.g1e3a90e_1_28)
- [Vector tiles for fast custom maps](http://vimeo.com/106228141) (video)
- [Videos from FOSS4G](https://www.mapbox.com/blog/foss4g-jsgeo-videos/)
- [Mapbox Vector Tile Specification](https://github.com/mapbox/vector-tile-spec)
  - [blog post](https://www.mapbox.com/blog/vector-tiles/)
  - tools: [vector-tile-js](https://github.com/mapbox/vector-tile-js) and [mapnik-vector-tile](https://github.com/mapbox/mapnik-vector-tile)
  - Client-side: pbf.js to decode tiles (protocol buffers).
- [geojson-vt](https://github.com/mapbox/geojson-vt)
- [MMT-Vector-Tiles](https://github.com/glob3mobile/mmt-vector-tiles/)

pjcozzi:
More discussion: https://groups.google.com/forum/#!topic/cesium-dev/QnOxsA0iwa0

mramato:
Looks like ESRI is adopting Mapbox vector tiles: https://www.mapbox.com/blog/vector-tile-adoption/

pjcozzi:
Info from [FOSS4G NA 2015](http://geothought.blogspot.com/2015/03/review-of-foss4g-na-2015.html).

pjcozzi:
From Mapzen: [Look Upon Our Squares of Math in Three Dimensions](https://mapzen.com/blog/look-upon-our-squares-of-math-in-three-dimensions)

pjcozzi:
* State of the Map 2015 - Vector Rendering Panel - [video](https://www.youtube.com/watch?v=0RA5k7J5Sw0), [transcript](https://s3.amazonaws.com/assets-staging.mapzen.com/images/state-of-the-map-2015-wrapup/sotmus2015_Mapzen_transcripts.html#matt)

[...]

PR #13343, opened 2026-03-30
Vector Tiles: Picking

donmccurdy:
# Description

Changes:

- Adds Cesium3DTileVectorFeature class, representing a collection of vectors associated with a single feature
- Modifies VectorGltf3DTileContent so that when creating BufferPrimitiveCollections, the pickObjects for each primitive are overridden to use Cesium3DTileVectorFeature.
- For users creating BufferPrimitiveCollections directly, the default pickObject (collection, index, primitive) is used as before. I've moved this code up from the renderers into the BufferPrimitiveCollection class.

## Issue number and link

- #13156 
- #2132

## Testing plan

Test data:

```bash
git checkout donmccurdy/demo/bufferprim-sandcastles -- Apps/SampleData/
```

<details>
<summary>Sandcastle</summary>

```javascript

  shadows: true,
});
viewer.extend(Cesium.viewerCesium3DTilesInspectorMixin);
const inspectorViewModel = viewer.cesium3DTilesInspector.viewModel;

viewer.clock.currentTime = new Cesium.JulianDate(2457522.154792);

const scene = viewer.scene;

const tileset = await Cesium.Cesium3DTileset.fromUrl(
  //"../../SampleData/vector/sample-us-states.tileset.json",
   "../../SampleData/vector/sample-us-outline.tileset.json",
  // "../../SampleData/vector/sample-cities-spain.tileset.json",
  {enableDebugWireframe: true}
);

inspectorViewModel.tileset = tileset;

viewer.zoomTo(
  tileset,
  new Cesium.HeadingPitchRange(
    0,
    -2.0,
    Math.max(100.0 - tileset.boundingSphere.radius, 0.0),
  ),
);

  pointSize: "12",
  lineWidth: "4",
});

const handler = new Cesium.ScreenSpaceEventHandler(viewer.canvas);

handler.setInputAction(function (event) {
  const feature = scene.pick(event.endPosition);
  if (Cesium.defined(feature)) {
    feature.color = Cesium.Color.CRIMSON;
  }

handler.setInputAction(function (event) {
  const feature = scene.pick(event.position);
  if (Cesium.defined(feature)) {
    feature.show = false;
  }
}, Cesium.ScreenSpaceEventType.MIDDLE_CLICK);
```
</details>

Point and polyline samples should we working as expected. In the polygon sample, many states share the same Feature ID and are being picked as one feature accordingly. I suspect that's a bug in the data, not the loader, but we don't have a real pipeline to produce polygon tilesets yet, so I'm not sure.

# Author checklist

#### PR Dependency Tree

* **PR #13342**
  * **PR #13343** 👈

### DIFF
--- a/packages/engine/Specs/Scene/Cesium3DTileVectorFeatureSpec.js
+++ b/packages/engine/Specs/Scene/Cesium3DTileVectorFeatureSpec.js
@@ -1,0 +1,145 @@
+import {
+  Cartesian3,
+  Cesium3DTileVectorFeature,
+  Color,
+  BufferPointCollection,
+  BufferPointMaterial,
+} from "../../index.js";
+import createScene from "../../../../Specs/createScene.js";
+
+describe(
+  "Scene/Cesium3DTileVectorFeature",
+  () => {
+    let scene;
+    let content;
+
+    beforeAll(function () {
+      scene = createScene();
+    });
+
+    afterAll(function () {
+      scene.destroyForSpecs();
+    });
+
+    beforeEach(function () {
+      scene.primitives.removeAll();
+      content = { tileset: {}, _collections: [] };
+    });
+
+    it("content", () => {
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+      expect(feature.content).toBe(content);
+    });
+
+    it("tileset", () => {
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+      expect(feature.tileset).toBe(content.tileset);
+    });
+
+    it("primitive", () => {
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+      expect(feature.primitive).toBe(content.tileset);
+    });
+
+    it("featureId", () => {
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+      expect(feature.featureId).toBe(123);
+    });
+
+    it("pickIds", () => {
+      const points1 = new BufferPointCollection({ primitiveCountMax: 16 });
+      const points2 = new BufferPointCollection({ primitiveCountMax: 32 });
+
+      content._collections.push(points1, points2);
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+
+      const point1 = points1.add({ position: Cartesian3.UNIT_X });
+      point1._pickId = 101;
+      feature.addPrimitiveByCollection(0, points1.primitiveCount - 1);
+
+      points1.add({ position: Cartesian3.UNIT_Y }); // Skip.
+
+      const point2 = points1.add({ position: Cartesian3.UNIT_Z });
+      point2._pickId = 102;
+      feature.addPrimitiveByCollection(0, points1.primitiveCount - 1);
+
+      const point3 = points2.add({ position: Cartesian3.UNIT_X });
+      point3._pickId = 103;
+      feature.addPrimitiveByCollection(1, points2.primitiveCount - 1);
+
+      expect(feature.pickIds).toEqual([101, 102, 103]);
+    });
+
+    it("show", () => {
+      const points1 = new BufferPointCollection({ primitiveCountMax: 16 });
+      const points2 = new BufferPointCollection({ primitiveCountMax: 32 });
+
+      content._collections.push(points1, points2);
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+
+      const point1 = points1.add({ position: Cartesian3.UNIT_X, show: false });
+      feature.addPrimitiveByCollection(0, points1.primitiveCount - 1);
+
+      points1.add({ position: Cartesian3.UNIT_Y }); // Skip.
+
+      const point2 = points1.add({ position: Cartesian3.UNIT_Z, show: false });
+      feature.addPrimitiveByCollection(0, points1.primitiveCount - 1);
+
+      const point3 = points2.add({ position: Cartesian3.UNIT_X, show: false });
+      feature.addPrimitiveByCollection(1, points2.primitiveCount - 1);
+
+      expect(feature.show).toBe(false);
+
+      point2.show = true;
+
+      expect(feature.show).toBe(true);
+
+      feature.show = true;
+
+      expect([point1.show, point2.show, point3.show]).toEqual([
+        true,
+        true,
+        true,
+      ]);
+
+      feature.show = false;
+
+      expect([point1.show, point2.show, point3.show]).toEqual([
+        false,
+        false,
+        false,
+      ]);
+    });
+
+    it("color", () => {
+      const points1 = new BufferPointCollection({ primitiveCountMax: 16 });
+      const points2 = new BufferPointCollection({ primitiveCountMax: 32 });
+
+      content._collections.push(points1, points2);
+      const feature = new Cesium3DTileVectorFeature(content, 123);
+
+      const material = new BufferPointMaterial({ color: Color.ORANGE });
+
+      const point1 = points1.add({ position: Cartesian3.UNIT_X, material });
+      feature.addPrimitiveByCollection(0, points1.primitiveCount - 1);
+
+      points1.add({ position: Cartesian3.UNIT_Y }); // Skip.
+
+      const point2 = points1.add({ position: Cartesian3.UNIT_Z, material });
+      feature.addPrimitiveByCollection(0, points1.primitiveCount - 1);
+
+      const point3 = points2.add({ position: Cartesian3.UNIT_X, material });
+      feature.addPrimitiveByCollection(1, points2.primitiveCount - 1);
+
+      expect(feature.color).toEqual(Color.ORANGE);
+
+      feature.color = Color.CYAN;
+
+      expect(feature.color).toEqual(Color.CYAN);
+      expect(point1.getMaterial(material).color).toEqual(Color.CYAN);
+      expect(point2.getMaterial(material).color).toEqual(Color.CYAN);
+      expect(point3.getMaterial(material).color).toEqual(Color.CYAN);
+    });
+  },
+  "WebGL",
+);


packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -187,6 +215,14 @@ function appendPrimitiveToBuffers(content, primitive, collection) {
...
       featureIdAttribute.typedArray ??
       ModelReader.readAttributeAsTypedArray(featureIdAttribute);
+
+    for (let i = 0; i < featureIdArray.length; i++) {
+      const featureId = featureIdArray[i];
+      if (!features.has(featureId)) {
+        const feature = new Cesium3DTileVectorFeature(content, featureId);
+        features.set(featureId, feature);
+      }
+    }
```
donmccurdy, line 225: There are some baked-in assumptions here: 

- Feature IDs are scoped to the tile, i.e. if two features in the same tile share a Feature ID, they are the same feature
- Each vector primitive has 0 or 1 IDs

I'm not sure yet how that will evolve. We will probably want to associate features spanning multiple tiles eventually. I'm not sure whether having multiple sets of Feature IDs is needed. Hopefully these are reasonable assumptions to start?

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -296,13 +338,15 @@ function appendNodeToBuffers(content, node, parentTransform, result) {
     if (primitiveType === PrimitiveType.POINTS) {
       collection = new BufferPointCollection({
         primitiveCountMax: stats.pointPrimitiveCount,
+        allowPicking: true,
```
donmccurdy, line 341: BufferPrimitiveCollection disables picking by default, as there's non-trivial performance cost during initialization. Cesium3DTileset has an `enablePicking` property described with:

> Whether to allow collision and CPU picking with <code>pick</code> when using WebGL 1. If using WebGL 2 or above, this option will be ignored. If using WebGL 1 and this is true, the <code>pick</code> operation will work correctly, but it will use more memory to do so. If running with WebGL 1 and this is false, the model will use less memory, but <code>pick</code> will always return <code>undefined</code>. This cannot be set after the tileset has loaded.

It might be ideal if users could explicitly enable/disable picking on the tileset and that would be passed through into buffer collections, but with the WebGL1-specific behavior above I'm not sure that makes sense.

So for now, picking is just on by default in vector tilesets (which currently require WebGL2 anyway).

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -479,6 +506,46 @@ class BufferPrimitiveCollection {
...
+      this.get(i, primitive);
+
+      const pickObject = this._pickObjects[i] || {
+        collection: this,
+        index: i,
+        get primitive() {
+          // Cannot reuse primitives; scene.drillPick() appends to a list.
+          return collection.get(i, new PrimitiveClass());
+        },
+      };
```
donmccurdy, line 544: The idea here is that the user can explicitly pass in a `pickObject` when creating the primitive, and the object will be associated with the PickId instead of the default collection/index/primitive object created here.

Mainly this is intended to support vector tiles; I don't see much purpose for the custom pick object as a user-facing API.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -479,6 +506,46 @@ class BufferPrimitiveCollection {
...
+
+      const pickId = context.createPickId(pickObject);
+      primitive._pickId = pickId.key;
+      pickIds.push(pickId);
```
donmccurdy, line 548: Type checks here depend on https://github.com/CesiumGS/cesium/pull/13342.

packages/engine/Source/Scene/Cesium3DTileVectorFeature.js
```diff
@@ -0,0 +1,542 @@
...
+   * @default true
+   */
+  get show() {
+    for (const prim of this._iteratePrimitives()) {
+      if (prim.show) {
+        return true;
+      }
+    }
+    return false;
+  }
```
donmccurdy, line 113: Kind of arbitrary, but, `feature.show` returns true if at least one primitive is visible.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -121,6 +126,20 @@ class BufferPrimitiveCollection {
      */
     this._allowPicking = options.allowPicking ?? false;
 
+    /**
+     * @type {Map<Context, Destroyable[]>}
```
lilleyse:
```suggestion
     * @type {Map<Context, PickId[]>}
```

packages/engine/Source/Scene/Cesium3DTileVectorFeature.js
```diff
@@ -0,0 +1,542 @@
...
+
+  /**
+   * @type {Color}
+   * @default Color.WHITE
+   */
+  get color() {
+    for (const material of this._iterateMaterials()) {
+      return Color.clone(material.color, color);
+    }
+    return Color.clone(Color.WHITE, color);
```
lilleyse: `Cesium3DTileFeature` stores the color result in a class member rather than a scratch variable. I assume this was intentional to prevent different `color()` results from clobbering each other.

With the code here `const color0 = feature0.color; const color1 = feature1.color` would point to the same object. Maybe that was problematic for styling? I don't quite remember.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -174,10 +188,24 @@ function appendPrimitiveToBuffers(content, primitive, collection) {
...
+
   /** @type {TypedArray} */
   let featureIdArray;
+
+  /**
+   * @param {number} vertexOffset
+   * @returns {Cesium3DTileVectorFeature}
+   */
+  function getFeature(vertexOffset) {
+    const featureId = Math.trunc(featureIdArray[vertexOffset]);
```
lilleyse: Is `Math.trunc` to handle floating point imprecision? If so would `Math.round` be potentially safer?

packages/engine/Source/Scene/Cesium3DTileVectorFeature.js
```diff
@@ -0,0 +1,542 @@
...
+   * Returns whether the feature contains this property. This includes properties from this feature's
+   * class and inherited classes when using a batch table hierarchy.
+   *
+   * @see {@link https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_batch_table_hierarchy}
+   *
+   * @param {string} name The case-sensitive name of the property.
+   * @returns {boolean} Whether the feature contains this property.
+   */
+  hasProperty(name) {
+    return name === "id";
```
lilleyse: I feel like this default "id" property could be problematic, and instead there should be a `featureId` getter like in `Cesium3DTileFeature`.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
donmccurdy: Hm that's a good question. I don't know, this landed in an earlier PR and I'd assumed it was copied from somewhere else, but I don't see any other `Math.trunc()` calls in the codebase. A related question might be, is 0.5 a valid Feature ID? I don't believe `EXT_mesh_features` answers that, but has this in an implementation note...

> Because glTF accessors do not support `UNSIGNED_INT` types for 32-bit integers in vertex attributes, `FLOAT` may be used instead allowing integer feature IDs up to 2<sup>24</sup>. For smaller ranges of feature IDs, `UNSIGNED_BYTE` or `UNSIGNED_SHORT` may be used.

... which kind of leads me to believe that Feature IDs must be integers, and that if future extensions allowed uint32 or uint64 attributes then Feature IDs could use those too. In which case it *shouldn't* matter here whether we call .trunc(), .round(), or just use the value as-is.

packages/engine/Source/Scene/Cesium3DTileVectorFeature.js
donmccurdy: Yeah, tricky to compare feature colors here. That could be confusing. I am a bit worried that we're allocating a lot of JS memory for picking and related data structures, which may hit us for large datasets. But also trying to keep backward-compatibility with existing picking and Cesium3DTileFeature APIs.

I think I might be inclined to add these class members for now, and to optimize this later. Possibly trying to avoid pre-allocating a Cesium3DTileFeature for every point/line/polygon, similar to the "views" used in BufferPrimitiveCollection.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
lilleyse: I had a similar thought. Floating point can represent integers precisely so maybe just use the value as-is.

donmccurdy:
All feedback should be addressed now! This will be a small breaking change for the vector-alpha branch, if testers were relying on the "id" properties, but `feature.featureId` will give the same result.

lilleyse:
@donmccurdy did you plan on adding the `featureId` getter in this PR? There are a few occurrences of `pickObject.getProperty("id");` that may need to be updated as well.

donmccurdy:
Ahh, I'd been thinking Cesium3DTileVectorFeature extended Cesium3DTileFeature and inherited the implementation, but I see that's not the case currently. Yes thanks - I'll add a `featureId` getter and fix up the remaining `getProperty("id")` calls early next week.

donmccurdy:
Updated .getProperty -> .featureId usage, and added general unit test coverage.

PR #13363, opened 2026-04-03
Don't cut off monaco overlays

jjspace:
# Description

image

@javagl It finally annoyed me enough to track down a setting to change it.

| Before | After |
|--------|--------|
|  |  | 

Personally I would like if we could prevent the tooltip overlay from being rendered outside it's container _at all_ but after some digging this does not seem to be possible.

The alternative to using the fixed position is to remove the `overlay: hidden` on the `Allotment` panel but I don't think that's as nice of a change.

## Issue number and link

Reported here: https://github.com/CesiumGS/cesium/issues/12857#issuecomment-3505502918

## Testing plan

- Open sandcastle locally or [in CI](https://ci-builds.cesium.com/cesium/monaco-overflow/Apps/Sandcastle2/index.html)
- Hover a variable or class or something
- Check that tooltips and intellisense can render outside the editor container and are no longer truncated
- Shrink your entire window and expand the editor to max width
- Make sure overlays will not escape or get truncated by the whole window

# Author checklist

## AI acknowledgment

### DIFF
--- a/scripts/buildSandcastle.js
+++ b/scripts/buildSandcastle.js
@@ -114,17 +114,19 @@ export async function buildSandcastleApp({
       cesiumVersion: version,
       commitSha: JSON.stringify(process.env.GITHUB_SHA ?? undefined),
       imports: {
+        // The `paths` are reaching one level higher than the `typesPaths`
+        // because they're requested from the nested templates/bucket.html
         cesium: {
           path: "../../../Source/Cesium.js",
-          typesPath: "../../../Source/Cesium.d.ts",
+          typesPath: "../../Source/Cesium.d.ts",
         },
         "@cesium/engine": {
           path: "../../../packages/engine/Build/Unminified/index.js",
-          typesPath: "../../../packages/engine/index.d.ts",
+          typesPath: "../../packages/engine/index.d.ts",
         },
         "@cesium/widgets": {
           path: "../../../packages/widgets/Build/Unminified/index.js",
-          typesPath: "../../../packages/widgets/index.d.ts",
+          typesPath: "../../packages/widgets/index.d.ts",
         },
       },
     });


javagl:
In that comment I said 

> But maybe there's some CSS magic that allows fixing this "easily" (when you know exactly where and how...)

and "add `fixedOverflowWidgets: true` in line 385" here qualifies as "magic" 😁 

---

I tried it out, and it seemed to work as expected. And even though I "know" that 'longitude' comes first, it's still reaffirming to _see_ it...

Cesium Sandcastle Tooltip

---

Beyond trying it out, I cannot really "review" the change. That magic line seems to be it. And I can approve this at any time. But I'm wondering:

1. What is this `typesPath` change? 
2. What do you mean by 
    > if we could prevent the tooltip overlay from being rendered outside it's container at all 

    ? When the tooltip appears at the end of the line (like the screenshot above), I think that it _should_ be rendered outside of its container, right?

jjspace:
> That magic line seems to be it.

@javagl The "magic" if you're interested is that under the hood monaco creates an "overflow elements" container in DOM. That attribute makes it so all overflow elements use [`position: fixed`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/position#fixed) which means they don't obey the normal stacking contexts and can be displayed outside `overflow: hidden` elements.

> What is this typesPath change?

I noticed types were not loading in CI because the paths were off by one folder level. I meant to fix it a little bit ago but it was necessary to test this specific change in CI. Otherwise there was not really any overlay info to display in the first place.

> What do you mean by...

I personally kinda didn't think so. I wanted a way to restrict the space that monaco would even try to render overlays to it's own container. I was hoping it could _shift_ them back to the container instead of getting cut off for trying to render bigger than the container. Not a huge deal and I'm fine with this approach, anything is better than the truncation that we had.

javagl:
>  I wanted a way to restrict the space that monaco would even try to render overlays to it's own container. 

Ah... something like this (+line breaks in tooltip)?

Cesium Sandcastle Tooltip2

I think it's nice to let the tooltip originate at the cursor position - maybe there are other ways to do this, but I can not imagine it in a way that would be substantially "better".

About the `typesPath`... I have not checked this in all depth (builds seem to be passing - didn't inspect the logs...), but will take your word here :)

javagl (approved):
Works as expected

jjspace:
> Ah... something like this (+line breaks in tooltip)?

Yea something like that, again, not a big deal and I'm fine with it as it is.

@javagl are you comfortable with merging this?

Issue #9370, opened 2021-02-08
Setting show as true and setting the positions as CallbackProperty of a polyline which is clampToGround will cause Cesium stop rendering


ZenDay:
Sandcastle example:

https://sandcastle.cesium.com/#c=vVRNj5swEP0rVk4gUQM2BpMvVYqqXiq1UqNeSg9ecBprHRvZJhGt8t9rQlKx3Rx6ySIheDPznkdvBo7MgKPgJ27ACih+AhtuRXeA3y6xoJrVF7zRyjGhuKlm4aJSlaq1sg7shLHug3LC9Z4+6kA+YMEtZE0T/K4UAK2WvfTsObjA4Wq19UVeZA6+32Jg2sCGGeffmMLBO5QntEAIIkryMi3KIgJZkaKEUkgoIRkitIwAzpKMliksSFYWNMlQHkb/J51hiGmeIVpgNEojjGCCMSK4SOkonRcUZiUpEc0JJuFN+cf1jFqyQ7vVH43uVDMHznR8SJwrdZ46Zrl/Nm9lGXmcZfQNLLu6s5P9VgeTXQuh23MVBCFYrUd/Jklo9/rkrd0xafliSA635W4rDlx3bkq7Sxz6WLzO3iYC/87h5RezYVI+sfr5i9EtN65/eQ4AhrvOqOnoHrrvj9t4P8CrPedodDm8YI/SJEkGcJnfLJotreslX9+Y78Wh1caBzsgAwtjxQyuZbyx+6upn7mBtbXiVXsZT6rIRRyCa1Z3f0bBF1vrMrpPyq/jFq9l6Gfv6V1SpWSPUz89HbiTrh7J9uv40BiGEy9jD+0yntR+t+Uf5Dw

Browser: Chrome 88.0.4324.146

Operating System: macOS 10.14.6

Cesium Version: 1.78

Ensure that there are at least to polyline which is clampToGround. Setting show as true and setting the positions as CallbackProperty of a polyline which is clampToGround will cause Cesium stop rendering.

Reason:
In file `StaticGroundPolylinePerMaterialBatch.js`, in function `Batch.prototype.updateShows`: 
```js
var entity = updater.entity;
var instance = this.geometry.get(updater.id);

var attributes = this.attributes.get(instance.id.id);
```
[...]

PR #9436, opened 2021-03-19
fix issues 9370

Anigod:
https://github.com/CesiumGS/cesium/issues/9370

It is possible to change the position during the rendering process, so that it will not crash again after processing, and it can be rendered normally

PR #13366, opened 2026-04-05
fix: clean up showsUpdated on entity removal in ground batches (#9370)

chalabi2:
- **fix: clean up showsUpdated on entity removal in ground batches (#9370)**
- **chore: add entry to contributors list**

# Description

- Add missing `this.showsUpdated.remove(id)` in `Batch.prototype.remove` for both `StaticGroundPolylinePerMaterialBatch` and `StaticGroundGeometryPerMaterialBatch`
- When an entity's `isShowing` property changes and the entity is removed before the next render, `updateShows` iterates a stale entry, calls `this.geometry.get(updater.id)` on a removed entity, and crashes with `Cannot read properties of undefined (reading 'id')`
- Adds regression tests to both spec files that reproduce the exact crash scenario: add entity → trigger show change → remove entity → update batch

## Issue number and link

#9370

## Testing plan

- [x] Existing `StaticGroundPolylinePerMaterialBatchSpec` tests pass
- [x] Existing `StaticGroundGeometryPerMaterialBatchSpec` tests pass
- [x] New test "does not crash when removing an entity with a pending show update" passes in both specs
- [x] Manual: toggle `entity.show`, then remove entity before next frame — no crash

# Author checklist

## AI acknowledgment


chalabi2:
this is ready

chalabi2:
> this is ready

Is this PR of interest to the team?

javagl:
It should be. The change here seems to be what was already suggested in [this comment, four years ago(!)](https://github.com/CesiumGS/cesium/pull/9436#issuecomment-1020458580). That PR was closed to "keep the PR backlog tidy", which is ... something that I won't publicly comment on. The linked comment already mentioned that the necessary change was found by comparing different `...Batch...` implementations. This indicates what I already pointed out as the most fundamental root cause of all this, in https://github.com/CesiumGS/cesium/issues/12714 , where I said

> I'm reasonably sure that some of the differences are just the results of certain bugfixes going into one class but not into the others. And I'm reasonably sure that future bugfixes will only go into one of these classes and not into the others, because nobody would dare to touch these classes, out of the fear of breaking something.

---

Now, what are we going to do here? 

The code changes are trivial. But there is no point in "reviewing" them. because nobody can say whether this is the right thing to do. 

The only thing that _can_ be done: 

I tried out this branch with the sandcastle from https://github.com/CesiumGS/cesium/issues/9370. Previously, the sandcastle crashed. Now it doesn't crash. In the spirit of "Whack a thousand moles", I'd say we merge this. If someone in the future finds a scenario wheren ~"something with the 'Batches' doesn't work properly", and wants to blame me for that because I merged this PR, then I'd say 'Prove it!', lean back, and enjoy the show.

So when you update the branch by merging in the latest `main` state, and add a hint in the 'CHANGES.md', I'd hit "Approve+merge" here.

chalabi2:
Thanks for your review. Sounds like things are pretty painful here so appreciate your efforts.

javagl (approved):
As mentioned in the comment: The change looks like it aligns what is done _here_ with what is already done in other classes. Despite my lack of context of what this actually _is_, the change seems to fix the issue, and I don't (fore)see any side-effects.

PR #13372, opened 2026-04-07
Scripts: Add diff-types.sh

donmccurdy:
# Description

Adds a bash script to quickly compare `Cesium.d.ts`, the types from the /engine and /widgets packages, on the current branch vs. `main`. Usage:

```bash
./scripts/diff-types.sh
```

## Issue number and link

- #4434
- #8359 

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13372** 👈

### DIFF
--- a/scripts/diff-types.sh
+++ b/scripts/diff-types.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+# diff-types.sh
+# Compares Cesium.d.ts on current branch and on main. The script
+# should be run from a feature branch. Example:
+#
+# git checkout my-feature-branch
+# ./scripts/diff-types.sh
+
+git checkout main
+npm run build-ts
+cp Source/Cesium.d.ts Source/Cesium.main.d.ts
+git checkout -
+npm run build-ts
+git diff --no-index Source/Cesium.main.d.ts Source/Cesium.d.ts


lukemckinstry:
Having used this several times, I'm in favor of adding this!

scripts/diff-types.sh
```diff
@@ -0,0 +1,15 @@
...
+# diff-types.sh
+# Compares Cesium.d.ts on current branch and on main. The script
+# should be run from a feature branch. Example:
+#
+# git checkout my-feature-branch
+# ./scripts/diff-types.sh
+
+git checkout main
+npm run build-ts
+cp Source/Cesium.d.ts Source/Cesium.main.d.ts
```
jjspace, line 12: This should probably also check the types for each of the sub packages too

donmccurdy: I believe this file includes both /engine and /widgets, so it's intended to be a way to observe the complete diff at once, including anything that might break or conflict when joining the types of the two packages. Does that seem OK?

jjspace: It does combine them but it's sorta wrapped a little differently. Fine with this if others think it's fine but I think as a sanity check I _personally_ would want to check them all.

donmccurdy: I guess my concern is that this doubles the length of the diff — I don't think I'd want to impose that on every code review related to types and ES6. But I am hoping that we can make it "easy enough" to review these diffs on PRs related to the conversions, at least until we're pretty confident there are no more edge cases hiding that we can't detect automatically with something like #13377.

I agree that if we changed something likely to affect packaging/wrapping, e.g. the build script, that might be a good case to diff sub-package types separately... Maybe exposing an (optional) parameter for the name of the file to diff?

```bash
./scripts/diff-types.sh [ path/to/types.d.t.s ]
```

donmccurdy:
Ready for review!

scripts/diff-types.sh
jjspace: An optional parameter to include packages could be nice but up to you

jjspace:
No need to hold up this pr for my comments, especially since @lukemckinstry  said it's already helpful. 

I do wonder if there's a similarly easy way to diff the docs. I tried one time and looking at the html directly was sort of a mess. Something to think about

lukemckinstry (approved):
Good to go as this is very useful in its current state 👍 Confirmed with @jjspace offline he is good with this as-is too.

PR #12933, opened 2025-09-29
Use eye coordinates for voxel raymarching

jjhembd:
# Description

This PR reworks the voxel shaders to perform the raymarching in eye-space coordinates, rather than model-space. This avoids precision problems when zoomed in close to small samples in a large model space.

Most of the precision problem came from the sampling of the voxel. The traversal in `Octree.glsl` has been re-worked to input a two-part coordinate:
- Tile coordinate: the integer `[x, y, z, w]` indices of the tile within the tileset
- Tile UV coordinate: floating point numbers in the interval `[0, 1]` indicating the `x, y, z` positions within the tile

This new input to the traversal effectively gives us nearly double precision when sampling a large dataset.

Some parts of the code are still in a local model-centered coordinate system:

- Calculation of the "jacobian" (partial derivatives of the transform from Cartesian to cylindrical/ellipsoidal coordinates). This matrix is more easily defined in model coordinates.
- Clipping plane intersections
- Many of the initial shape intersection routines

The latter 2 items could also be converted to eye coordinates with additional work. I did not do that work in this PR, mostly because of time constraints. I also haven't observed significant precision issues with those intersections, so that work does not seem urgent at this time.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/12061

## Testing plan

[...]

PR #13375, opened 2026-04-07
Clean up voxel shape transforms

### LINKED ISSUE #12782: Misleading scale/offset defaults in voxel shapes
### What happened?

For voxels we have some scale / offset properties that are passed to the shader where X = scale and Y = offset. Typically a default scale / offset would be 1 and 0 respectively, indicating no transformation of an object. However, those values are flipped in multiple parts of the code:

https://github.com/CesiumGS/cesium/blob/553ac5760205fa6c4de4cd3a28d887fda566a565/packages/engine/Source/Scene/VoxelCylinderShape.js#L313-L314

https://github.com/CesiumGS/cesium/blob/553ac5760205fa6c4de4cd3a28d887fda566a565/packages/engine/Source/Scene/VoxelCylinderShape.js#L341-L342

https://github.com/CesiumGS/cesium/blob/553ac5760205fa6c4de4cd3a28d887fda566a565/packages/engine/Source/Scene/VoxelCylinderShape.js#L417-L418

https://github.com/CesiumGS/cesium/blob/553ac5760205fa6c4de4cd3a28d887fda566a565/packages/engine/Source/Scene/VoxelEllipsoidShape.js#L522-L523

https://github.com/CesiumGS/cesium/blob/553ac5760205fa6c4de4cd3a28d887fda566a565/packages/engine/Source/Scene/VoxelEllipsoidShape.js#L645-L646

Although this bug hasn't seemed to ever come up in a practical use case, these values should still be corrected IMO for code quality.

### Reproduction steps

N/A

### Sandcastle example

_No response_

### Environment

Browser: Chrome
CesiumJS Version: Latest
Operating System: Windows 11

jjhembd:
# Description

This PR cleans up some internal transforms for voxel shape bounds.

The `VoxelShape` classes had separate `vec2` uniforms for the bounds along each dimension, each uniform containing a scale and a translation. However, since https://github.com/CesiumGS/cesium/pull/12933, the translation was not used in the shader. This PR combines the 3 scalars into a single `vec3` uniform.

A few inline comments were added to clarify some of the bounds transform calculations.

## Issue number and link

Resolves https://github.com/CesiumGS/cesium/issues/12782.

## Testing plan

- Confirm all voxel-related specs pass locally
- Load voxel-related sandcastles and confirm rendering is unchanged. Make sure to test all shape types in each sandcastle.

## Author checklist

## AI acknowledgment

GitHub Copilot

Claude Sonnet 4


jjhembd:
@j9liu would you be able to take a look at this? I think it should make things clearer to resolve your issue #12782, but let me know if the comments or structure could be improved!

packages/engine/Source/Shaders/Voxels/convertLocalToEllipsoidUv.glsl
```diff
@@ -16,13 +16,11 @@ uniform vec3 u_ellipsoidInverseRadiiSquared;
     uniform vec3 u_ellipsoidShapeUvLongitudeMinMaxMid;
 #endif
 #if defined(ELLIPSOID_HAS_SHAPE_BOUNDS_LONGITUDE)
-    uniform vec2 u_ellipsoidLocalToShapeUvLongitude; // x = scale, y = offset
     uniform float u_ellipsoidShapeUvLongitudeRangeOrigin;
 #endif
 #if defined(ELLIPSOID_HAS_SHAPE_BOUNDS_LATITUDE)
-    uniform vec2 u_ellipsoidLocalToShapeUvLatitude; // x = scale, y = offset
 #endif
```
j9liu: Is this `#if` needed anymore?

j9liu (commented):
This looks great to me @jjhembd, definitely a lot cleaner! I caught one lingering `#if defined` that could probably go away, but otherwise looks good to me.

I haven't built this locally to try it out though -- should I be the one to do that and merge, or should someone from CesiumJS take a look?

packages/engine/Source/Shaders/Voxels/convertLocalToEllipsoidUv.glsl
jjhembd: Good catch, thanks! I went through all the `#if`s and defines and removed a few more that were unused.

lukemckinstry:
I found just one bug from basic user testing on the sandcastles: I'm not sure this is related to this PR bc it also occurs on `main` (it is possible it originated in the preceding eye coordinates PR), but in https://sandcastle.cesium.com/?id=voxels, when `Ellipsoid - Procedural Tile` or `Ellipsoid - Procedural Tileset` are selected, changing any of the scale X/Y/Z parameters under Transform in the Voxel inspector results in a runtime error `DeveloperError: Ellipsoid must be an ellipsoid of revolution (radii.x == radii.y)`.

Edit:
The above happens for increasing or decreasing scale X and scale Y from the original value. 
Scale Z works when increasing to values high than the original value, but decreasing below the original value results in the following runtime error `DeveloperError: Expected right to be typeof object, actual typeof was undefined`.

This also occurs in the same patterns in http://localhost:8080/Apps/Sandcastle2/index.html?id=voxels-in-3d-tiles for `Ellipsoid - 3D Tiles`.

If this is not related to this PR, I think we should break this off into a new issue.

Besides this issue the PR looks good to me.

jjhembd:
@lukemckinstry thanks for testing! I believe some of the ellipsoid transform errors you are seeing are due to a pre-existing limitation of the code. The `VoxelInspector` may need to be reworked to link the values for "Scale X" and "Scale Y".

The issue with the z-scalar is unexpected for me. As you noted, this issue also pre-dates this PR.

I opened https://github.com/CesiumGS/cesium/issues/13456 to discuss further.

lukemckinstry (approved):
Approving as all the issues I found are unrelated and will be handled elsewhere 👍

Issue #13336, opened 2026-03-26
Investigate device pixel ratio default behavior


arobison:
### Feature

We've noticed several instances of visual aliasing and low pixel resolution when CesiumJS is rendered on a high dpi display with UI scaling enabled. For example, my external monitor is currently configured as below:

<img width="2387" height="1823" alt="Image" src="https://github.com/user-attachments/assets/57501fe5-a8e3-47db-8845-3fdb23f07059" />

Toggling `useBrowserRecommendedResolution` appears to greatly increase visual quality (smaller logical pixels).

Let's investigate whether our defaults for resolution, anti-aliasing, etc. make sense for our current user base. 

Some ideas that have come up so far:

* Adjusting the defaults for non-mobile devices to preserve the performance-oriented defaults on smaller devices
* Clamping device pixel ratio ranges ala Three Fiber

Do we have any data on the types and breakdown of the devices that are running CesiumJS?

jjhembd:
I personally think the naming of the `useBrowserRecommendedResolution` flag is confusing. When this is true, we are setting the canvas resolution such that the pixel size equals a CSS "pixel" measurement, which can be 2-3x bigger than the device resolution.

I don't think the CSS pixel size is at all intended to be a "recommended resolution" for rendering. It is instead meant to standardize font/icon sizes etc. across different displays.

Graphical rendering should normally default to the device resolution. Our defaulting to CSS pixel size was more of a pragmatic choice for performance reasons (see https://github.com/CesiumGS/cesium/pull/8548). This choice was made years ago when devices were slower, and is due for a re-evaluation.

jjhembd:
Related: https://github.com/CesiumGS/cesium/issues/12126
As mentioned there, we may need to turn off MSAA if we are honoring device resolution on a high-res screen.

alarkbentley:
> Related: [#12126](https://github.com/CesiumGS/cesium/issues/12126) As mentioned there, we may need to turn off MSAA if we are honoring device resolution on a high-res screen.

Did some testing using MSAA and native device resolution: (tldr. it has significant performance impact)
https://github.com/CesiumGS/cesium/pull/12158#issuecomment-4295127931

So I think as @jjhembd mentions that if we decide to default to device resolution we might need to reconsider the MSAA defaults.

PR #13384, opened 2026-04-08
BufferPrimitive: Add 'alpha' support

donmccurdy:
# Description

Changes:

- Add `blendOption` parameter in BufferPrimitiveCollection constructors
- Support `material.color.alpha` in BufferPrimitiveCollection subclasses
- Support `material.outlineColor.alpha` in BufferPointCollection

Tasks:

- [x] Expose `collection.blendOption` per-collection option
- [x] Support pointMaterial.color.alpha
- [x] Support pointMaterial.outlineColor.alpha
- [x] Support polylineMaterial.color.alpha
- [x] Support polygonMaterial.color.alpha
- [x] Unit tests

## Issue number and link

- https://github.com/CesiumGS/cesium/issues/13444

## Testing plan

Unit tests added. 

Sandcastle:

```javascript

const collection = new Cesium.BufferPolylineCollection({
  primitiveCountMax: 1024,
  vertexCountMax: 1024,
});

const degreesArray = [-105.0, 40.0, -100.0, 38.0, -105.0, 35.0, -100.0, 32.0];

const positions = Cesium.Cartesian3.packArray(
  Cesium.Cartesian3.fromDegreesArray(degreesArray),
  new Float64Array((degreesArray.length * 3) / 2),
);

const material = new Cesium.BufferPolylineMaterial({color: Cesium.Color.RED, width: 10});
material.color.alpha = 0.25;

collection.add({ positions, material }, polyline);

viewer.zoomTo(collection);
```

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13384** 👈


packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -171,7 +174,7 @@ function renderBufferPointCollection(collection, frameState, renderContext) {
                 vertexBuffer: Buffer.createVertexBuffer({
                   typedArray: collection._positionView,
                   context,
-                  usage: BufferUsage.DYNAMIC_DRAW,
+                  usage: BufferUsage.STATIC_DRAW,
```
donmccurdy, line 177: Unrelated - I've set all buffer attributes to STATIC_DRAW, just to keep things consistent for now. In the future I'd like to make this an option the user can control, if we can measure a difference in performance, which I haven't managed to do (at least on my machine) yet.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -95,6 +97,15 @@ class BufferPrimitiveCollection {
     this.show = options.show ?? true;
 
     /**
+     * Collection blend option; must be OPAQUE or TRANSLUCENT.
+     * @type {BlendOption}
+     * @readonly
+     * @ignore
+     */
+    this._blendOption = options.blendOption ?? BlendOption.TRANSLUCENT;
```
donmccurdy, line 105: I chose TRANSLUCENT as the default to reduce aliasing — if aliasing were less of an issue, and/or if the default canvas resolution in CesiumJS matched the display (#13336) then OPAQUE would probably make more sense as the default.

danielzhong (commented):
Great PR! Few questions:
1. Translucent mode should submit the the pass to `Pass.TRANSLUCENT` right? I see `renderBufferPointCollection.js` and polylines, polygons still using:

```
pass: Pass.OPAQUE,
```

Is there any sample test cases for me to test it as well?

2. When `blendOption` is TRANSLUCENT, the render state doesn't set `depthMask: false`. This means transparent primitives still write to the depth buffer, I'm not sure but do you think will this affect anything?
3. `PointPrimitiveCollection` exposes `blendOption` as a plain public property, allowing users to read and update it at runtime. For now, `BufferPrimitiveCollection` only stores `_blendOption` as a private field. Not sure if we need to have getter or setter for this (maybe no need?).

donmccurdy:
> Translucent mode should submit the the pass to Pass.TRANSLUCENT right? 

Good catch, thanks! Yes, probably we should derive the pass from the blend option. Done.

> Is there any sample test cases for me to test it as well?

The sandcastle snippet above is an easy test, or any of our existing vector tilesets can be tested by using styles like this:

```javascript
  color: "color('white', 0.25)",
  pointSize: 4,
  pointOutlineWidth: 2,
  pointOutlineColor: "color('purple', 0.25)",
});
```

I'm not aware that it's possible to change the blend option on a tileset currently.

> When blendOption is TRANSLUCENT, the render state doesn't set depthMask: false. This means transparent primitives still write to the depth buffer, I'm not sure but do you think will this affect anything?

Good question. It definitely _could_ affect things, but I don't think I'd want to derive the depth write setting directly from the blend option. I think we could expose depth buffer settings on the collection later, though, if that becomes necessary.

> PointPrimitiveCollection exposes blendOption as a plain public property, allowing users to read and update it at runtime. 

I'd be OK with letting users change this setting later, but no immediate use case for it, so I'm inclined to wait and keep things simple for now.

***

/cc @lilleyse just FYI, in case any of this feels like it may be inconsistent with CesiumJS user expectations?

danielzhong (approved):

lilleyse:
>> When blendOption is TRANSLUCENT, the render state doesn't set depthMask: false. This means transparent primitives still write to the depth buffer, I'm not sure but do you think will this affect anything?
>
> Good question. It definitely could affect things, but I don't think I'd want to derive the depth write setting directly from the blend option. I think we could expose depth buffer settings on the collection later, though, if that becomes necessary.

FWIW, `PointPrimitiveCollection` derives depth write from the blend option: https://github.com/CesiumGS/cesium/blob/d81e43dd624897ad816d19c88e6fa42ad9b1c772/packages/engine/Source/Scene/PointPrimitiveCollection.js#L1003-L1033

packages/engine/Source/Scene/BufferPrimitiveCollection.js
lilleyse: `BillboardCollection` and `PointPrimitiveCollection` default to `OPAQUE_AND_TRANSLUCENT`. That might be the better option long term though it does require splitting the rendering into two commands, an opaque command and a translucent command.

donmccurdy:
Thanks @lilleyse! My only hesitation is that I'd recalled alpha blending helping to mitigate aliasing recently, for otherwise opaque geometry, related to #13336. I should probably make a reproduction to confirm that use case, but if so, then disabling depthMask when alpha blending is enabled would make it difficult to use that approach for opaque geometry.

Not feeling firmly decided on this either way. Related to the "Create a Sandcastle comparing BufferPoint/Polyline/Polygon classes to existing CesiumJS APIs..." task in #13444.

Issue #9260, opened 2020-12-01
Shadowing of partially transparent textures does not work correctly for alphaMode: MASK


runaas:
When using gltf/glb models with partially transparent textures shadowing works when using a material with alphaMode: BLEND, but not when using alphaMode: MASK

The following example is from an x-board model of a tree when using MASK:
![bilde](https://user-images.githubusercontent.com/25476272/100728168-7dfa5b80-33c7-11eb-9820-8a3de2c76aa9.png)

The same model with alphaMode BLEND:
![bilde](https://user-images.githubusercontent.com/25476272/100728499-ea755a80-33c7-11eb-82db-9a191a75c8b2.png)

lilleyse:
The fix would be to always go through the non-opaque path in the code below. The downside is that it'll render the full shader (texturing, lighting, and all). But the pick shaders do something similar for the same reason.

https://github.com/CesiumGS/cesium/blob/48385b800aa87c689e75845191a24b522f3cc930/Source/Scene/ShadowMapShader.js#L95-L106

Longer term we would have different shader variants for pick, shadow, render so that the model shader would apply the apply the base color texture for purposes of using the alpha mask but nothing else.

PR #13388, opened 2026-04-09
Fix ghost shadows from clipped geometry in shadow cast pass

### LINKED ISSUE #6261: Shadows ignore clipping planes
badshadows

Masty88:
# Description

## Problem

When geometry is clipped with `ClippingPlaneCollection` or `ClippingPolygonCollection`, the clipped portions still cast shadows.

## Cause

1. Clipping discards are skipped in the opaque shadow cast path.
2. `model_clippingPlanesMatrix` is computed using the camera view matrix and reused during shadow rendering.

## Solution

1. Execute clipping discards during shadow casting when clipping is enabled.
2. Recompute `model_clippingPlanesMatrix` per draw call using the active view matrix.

---

### Root cause 1 — clipping discards were not evaluated in the shadow cast pass

For opaque geometry, the shadow cast shader skips the original fragment shader as an optimization since only depth output is required.

Clipping logic is injected into the fragment shader. In the opaque shadow cast path, the fragment shader is skipped, so clipping discards are not evaluated and clipped fragments still write depth to the shadow map.

The fix introduces `hasClippingForShadowCast()`, which detects whether clipping planes or clipping polygons are enabled for the shader.

When clipping is enabled, the cast shader invokes `czm_shadow_cast_main()` (the original fragment shader entry point renamed during shader generation) so clipping discards are evaluated during shadow rendering.

When clipping is not enabled, the existing optimized shadow cast path is preserved.

The clipping state is also included in the derived shader cache key to prevent clipped and non-clipped shader variants from sharing the same cache entry.

**BEFORE:**

before

**AFTER:**

after

---

### Root cause 2 — `model_clippingPlanesMatrix` used the wrong view matrix

Even after evaluating clipping discards in the shadow pass, clipping was still performed using an incorrect matrix.

`model_clippingPlanesMatrix` was computed in `updateReferenceMatrices()` using `view3D` from the camera and cached on the model. During shadow rendering, `czm_view` represents the light view matrix, causing clipping planes to be evaluated in camera space rather than light space.

The fix moves the matrix computation from `Model.js` into the uniform function in `ModelClippingPlanesPipelineStage`, where it is rebuilt per draw call using the active `view3D` matrix. This ensures the clipping matrix is correct regardless of which render pass is currently executing.

This changes the computation from once per model per frame to once per draw call, but only for models with clipping enabled. Performance measurements with 8 and 64 clipped models showed a small overhead relative to overall rendering cost:

https://github.com/CesiumGS/cesium/pull/13388#issuecomment-4217352938

**BEFORE:**

before

**AFTER:**

after

---

## Issue number and link

Fixes #6261

---

## Testing plan

* Added unit tests in `ShadowMapShaderSpec.js` covering:

  * `hasClippingForShadowCast()`
  * Opaque shadow shaders
  * Opaque shadow shaders with clipping
  * Translucent shadow shaders
  * Point light shadow shaders

* Added a new Sandcastle example: **Shadows with Clipping**

  * OSM Buildings clipped with a `ClippingPolygonCollection`
  * `CesiumAir` clipped with a `ClippingPlaneCollection`
  * Clipping can be toggled to reproduce the original shadow artifact

* Full test suite passes locally (`npm run test`)

---

## Author checklist

* [x] I have submitted a Contributor License Agreement
* [x] I have added my name to `CONTRIBUTORS.md`
* [x] I have updated `CHANGES.md` with a short summary of my change
* [x] I have added or updated unit tests to ensure consistent code coverage
* [x] I have updated the inline documentation, and included code examples where relevant
* [x] I have performed a self-review of my code

## AI acknowledgment

* [x] I used AI to generate content in this PR
* [x] I have reviewed the generated code and content

**Tools:** Claude
**Model:** Claude Sonnet 4.6


Masty88:
Hi @ggetz, draft fix for #6261 — let me know if the approach makes sense or if I'm going in the wrong direction!

javagl:
(Don't worry about the test that is currently failing. That's a known issue)

Masty88:
### Performance note

To quantify the trade-off introduced by the `ClippingPlaneCollection` fix (matrix computed per draw call instead of once per model per frame), I instrumented both branches with `performance.mark` around the matrix computation and captured Chrome traces with 8 and 64 clipped models active.

On **main**, the markers wrap the existing computation in `updateReferenceMatrices` in `Model.js` (called once per model per frame):

```js
// Model.js — updateReferenceMatrices()
if (model.isClippingEnabled()) {
  performance.mark("clippingPlanesMatrix-start");
  // ... multiply + inverseTranspose ...
  performance.mark("clippingPlanesMatrix-end");
}
```

On the **fix branch**, the markers wrap the uniform function in `ModelClippingPlanesPipelineStage.js` (called once per draw call):

```js
// ModelClippingPlanesPipelineStage.js — model_clippingPlanesMatrix uniform
model_clippingPlanesMatrix: function () {
  performance.mark("clippingPlanesMatrix-start");
  // ... multiply + inverseTranspose ...
  performance.mark("clippingPlanesMatrix-end");
  return result;
},
```

| Models | Branch | Calls / frame | Cost / frame |
|--------|--------|--------------|--------------|
| 8      | main   | 9            | 0.10 ms      |
| 8      | fix    | 71           | 0.52 ms      |
| 64     | main   | 58           | 0.40 ms      |
| 64     | fix    | 396          | 2.33 ms      |

The fix introduces a ~5–6× increase in matrix computations per frame (one per draw call instead of one per model). The jump in call count includes the shadow cast pass — which is the whole point of the fix, since the matrix now has to be recomputed there too with the light's view.

One caveat on reading these numbers:

- They **include the `performance.mark` overhead itself** (two marks per call, hundreds of times per frame), which isn't present in production — so the real cost is lower than what the table shows.


**Alternatives considered:**

- *Opt-in flag* (`clippingPlanesAffectShadows`) — let the user choose the trade-off. Rejected: ghost shadows are a correctness bug, not a performance trade-off. The correct behaviour should be the default.

Masty88:
A note after digging into this :
I went back and forth for a while on whether routing through czm_shadow_cast_main() was the right call or just a workaround. Looking into it these past few days I came across [this comment from lilleyse on #9260](https://github.com/CesiumGS/cesium/issues/9260#issuecomment-736568552), from 2020, describing the same idea for alphaMode: MASK shadows — "always go through the non-opaque path" so the full shader runs and the discard fires. 

He'd already weighed the cost and was fine with it: "it'll render the full shader... but the pick shaders do something similar for the same reason." That's the same trade-off as my per-draw-call change in root cause 2, so it was reassuring to see it had been considered acceptable before.The translucent path already worked this way (discard on alpha == 0); I just gated the same pattern behind hasClipping so it also covers opaque clipped geometry.

mzschwartz5 (changes requested):
Hi @Masty88,

Thanks for the contribution. This is going to be a really useful improvement to clipping.

I have one major point of feedback: I think the performance hit in moving matrix calculations to per-draw-call is maybe too costly and -- more importantly -- avoidable. 64 shadow-casting models is perhaps a heavy workflow, but ~2.3ms per frame on matrix computations is a lot. It's 20% of the whole frame budget, roughly.

Two ideas on how this could be improved:
1. Cache the matrix values. When the camera or the lights are not moving, we shouldn't need to do any computations here. Also, for models with many primitive components - and thus many draw calls - we'll be doing a lot more redundant work. There different ways you could implement caching, I'm leaving it purposefully open-ended. 
2.  Right now, the math we're doing is `invTranspose(view3D⋅reference⋅clippingPlanes)`. This is equivalent to `invTranspose(view3D)⋅invTranspose(reference⋅clippingPlanes)`. We already store `inverseView3D` (the active view's inverse -- correct for shadow passes too) on the `UniformState` class, so calculating that first term only involves a transpose - no need for an expensive inverse. The second term is view-independent, so we can calculate it once and reuse it so long as the clipping planes do not move. All-in-all, we can completely avoid calculating matrix inverses every frame, which is the vastly more expensive piece here.

We could probably get away with just one of these two (profiling to be sure), but both could be implemented for big gains. 

___

One minor thing: love that you added a sandcastle for the PR. Very helpful for testing. I don't think it needs to be committed to the repo though. It's useful for devs, but doesn't showcase anything new for users. I'd remove it from the PR (and just link it in the description).

mzschwartz5:
Oh, also - it might be good to get some benchmarks on the performance impact in the shadow pass. Shadows are already pretty darn expensive, so it's not ideal that we're adding more there (but at least it's only for clipping and, like you said, I can't see any way around it).

Masty88:
@mzschwartz5 , thanks for the review and for the split idea, that was the key.

So I did the split you suggested. The matrix is `inverseTranspose(view3D * reference * clipping)`. With `inverseTranspose(A*B) = inverseTranspose(A) * inverseTranspose(B)` I pulled out the view part. That part is just `transpose(inverseView3D)`, and UniformState already has inverseView3D, so there is no inverse to do, only a transpose. It reads the active view so it is also right in the shadow cast pass. The other part does not depend on the view, so I moved it back to `Model.updateReferenceMatrices` and compute it once per frame, same as the ibl matrix there. So no more inverse per draw call.

Matrix cost per frame, local sandcastle with N clipped shadow casting CesiumAir models:

| | 8 models | 64 models |
|---|---|---|
| original (main) | 0.012 ms | 0.069 ms |
| first fix (per draw inverse) | 0.211 ms | 1.930 ms |
| this fix | 0.187 ms | 0.868 ms |

Per frame averages over a few hundred frames. The main numbers are near the timer floor, the big ones (1.930 vs 0.868) are solid. So about 55% off at 64 and no per draw inverse.

For the shadow pass cost you asked about: at 64 models I was vsync capped so it looked free, it is not. I pushed it to GPU bound, 128 casters and a 4096 shadow map: main 21.67 ms, this branch 27.28 ms, so +5.6 ms. That is the full fragment shader running in the cast pass for the discard, plus the per draw matrix, I can't split the two from frame time. It only really shows on heavy clipped scenes, with a few casters it is in the noise. No way around it for correct shadows.

I also took the sandcastle out of the PR, kept it locally and can link it if useful.

One thing on AI since it comes up: I spent hours with Claude trying to optimize my first fix and it could not crack it without rewriting the architecture, your suggestion was the actual fix.

On the caching idea, I can take a crack at it and measure if it helps, just say the word.

packages/engine/Source/Scene/Model/Model.js
```diff
@@ -269,6 +269,7 @@ function Model(options) {
    */
   this.referenceMatrix = undefined;
   this._iblReferenceFrameMatrix = Matrix3.clone(Matrix3.IDENTITY); // Derived from reference matrix and the current view matrix
+  this._clippingPlanesMatrix = Matrix4.clone(Matrix4.IDENTITY); // View-independent part: inverseTranspose(referenceMatrix * clippingPlanes.modelMatrix)
```
mzschwartz5: I think for a first-time viewer of this code, this comment may be confusing. "View-independent part ... of what?" (is what I hear them asking). I'd personally just do something similar to the original comment ("derived from the reference matrix and the clipping planes' own model matrix") or something

packages/engine/Source/Scene/Model/Model.js
```diff
@@ -2479,16 +2479,14 @@ function updateReferenceMatrices(model, frameState) {
   );
 
   if (model.isClippingEnabled()) {
-    let clippingPlanesMatrix = scratchClippingPlanesMatrix;
-    clippingPlanesMatrix = Matrix4.multiply(
-      context.uniformState.view3D,
+    // View-independent part: inverseTranspose(referenceMatrix * clippingPlanes.modelMatrix).
+    // The view-dependent part (transpose(inverseView3D)) is applied per draw call in
+    // ModelClippingPlanesPipelineStage, so this is correct in both the main render pass
+    // (camera) and the shadow cast pass (where inverseView3D is the light's view).
```
mzschwartz5: Similar feedback here: I think this comment would confuse a first-time viewer of this code. I'm not sure someone looking at this particular function needs to know what the `ModelClippingPlanesPipelineStage` is doing. I'd opt to omit it.

By contrast, I think the comment you added in `ModelClippingPlanesPipelineStage` is exactly what a first-time reader needs.

mzschwartz5 (commented):
Hey @Masty88,

Nice revisions; thanks for looking into the impact on the shadow pass. And pretty big gains from factoring out view-dependent piece of the matrix calculation!

I'm still put off by the performance degradation overall, though. For the case of 64 models, your table says we've added ~0.8ms of compute time in computing clipping plane matrices. You measured 396 calls / frame for 64 models. That many matrix multiplications _should_ take on the order of microseconds, not milliseconds. 

Either:
1. Something is adding a lot of overhead (something to do with the uniform getters, maybe?)
2. The measurements aren't reliable (like what you said about `performance.mark`)

I think we should figure out what's going on here. One idea: if the `performance.mark` call really IS contributing significant overhead (which I'm dubious of, but worth testing), you could measure the delta of an entire _frame_ - with and without the matrix multiplication work in the clipping matrix uniform getter. 

If I think of any other ideas, I'll drop another comment... I still like the idea of caching, but it won't help if the camera or lights are moving. Before doing anything there, it would help to know what the bottleneck is right now.

packages/engine/Source/Scene/Model/Model.js
Masty88: done, removed it — the explanation stays in ModelClippingPlanesPipelineStage like you suggested,

packages/engine/Source/Scene/Model/Model.js
Masty88: done — changed it to the plain "derived from the reference matrix and the clipping planes' model matrix".

Masty88:
Hey @mzschwartz5,

Measuring the matrix cost turned out subtler than I expected.

The naive delta (frame time with the calc vs without) does not work: turning the matrix off breaks the clipping, so the discard changes and the GPU work changes with it. That mixes the matrix CPU cost with a GPU change, and the per-call cost is below performance.now resolution anyway. I caught this before posting a bad number.

So I used the work-based approach: keep the matrix correct, repeat the computation N times inside the getter, time the whole frame, divide by N. The N lifts the cost above the timer granularity. Ref: [Shipilev (JMH author)](https://shipilev.net/blog/2014/nanotrusting-nanotime/).

Matrix work cost per frame, 64 models, fixed camera:

| repeats N | frame time |
|---|---|
| 1 | 17.43 ms (on the vsync cap, context only) |
| 100 | 51.50 ms |
| 1000 | 302.91 ms |

From the two points above the cap: (302.91 - 51.50) / 900 = 0.28 ms/frame (the N=1 pairings agree, ~0.29-0.34, slightly higher only because N=1 sits on the vsync cap). So ~0.28 ms, against the 0.868 ms my marker table showed, so the markers were inflating it about 3x.

And you were right that the math is microseconds. The 0.28 ms is the JS per-call overhead of running transpose + multiply once per primitive across both passes, exactly the redundancy your caching idea removes.

So I implemented a version of the cache locally. I cache the matrix per model and reuse it across the model's primitives within a pass. To know when to recompute I added a small counter on UniformState, `viewChangedNumber`, bumped in `setView`, so it changes once per pass. The view is the only per-draw input, so the cache stays correct, and the shadow pass bumps it too (it calls `updateCamera` with the light camera), so it recomputes with the light view and the ghost shadows do not come back.

Same method, 64 models:

| | matrix cost / frame |
|---|---|
| per primitive (before) | 0.345 ms |
| cached per model (after) | 0.060 ms |

About 6x less, in the tens of microseconds you expected.

You're right a cross-frame cache would not help while the camera or lights move. This one is different though: it dedups across a model's primitives within a pass, and the view is fixed for the whole pass, so it holds up the same moving or not. The counter bumps on every `setView`, so even my fixed-camera numbers already recompute once per pass, same as a moving camera would. So I think the 6x is just the per-primitive dedup, but tell me if I am reading your point wrong.

The leftover is the getter still called once per primitive (it just checks the counter) plus ~128 real recomputes per frame (64 models across both passes). A cross-frame cache could sit on top to skip those when the scene is static, but as you said that does not help when things move.

This is a local prototype, not pushed yet. If invalidating on the view counter looks right to you, I can push a minimal patch with it; if you would rather key it differently (say also on a model/clipping token), I will do that instead.

mzschwartz5:
Okay yeah, I was just coming to the same conclusion (that the overhead of the ~800 `performance.mark` calls really were dominating the performance measurements...)

I'm not sure I fully understand the caching solution yet, but I'd like to see it. If it doesn't add too much complexity, it sounds like it could be well worth it.

Thanks again for the work here.

Masty88:
> Okay yeah, I was just coming to the same conclusion (that the overhead of the ~800 `performance.mark` calls really were dominating the performance measurements...)
> 
> I'm not sure I fully understand the caching solution yet, but I'd like to see it. If it doesn't add too much complexity, it sounds like it could be well worth it.
> 
> Thanks again for the work here.

Yeah, easier to see now that it is pushed. It is simple memoization: a counter on `UniformState` (`viewChangedNumber`, bumped in `setView`, so it changes once per pass), and the getter recomputes only when that changes. So the matrix is computed once per model per pass instead of once per primitive: just the counter, two fields on `Model`, and the check in the getter.

The one thing it leans on is setView running once per pass, which is what makes the counter reliable. It is the simplest thing that is correct now, happy to make the invalidation more defensive if you would prefer.

mzschwartz5:
Two points of feedback:

1. A question: does this PR fix model clipping polygons or is it only for model clipping planes?
2. Caching: now that I see it, I don't love the complexity caching adds. We've had to add several cache-related fields on both Model and UniformState, and there's a tight coupling now between Model and UniformState in that Model has to compare its version number to UniformState's. 

I thought through several alternatives, but none of them are obviously better, and I shouldn't spend too much more time on this right now. I think we should revert the caching commits (keeping them in the commit history though), and leave a comment suggesting that we might want to revisit the performance of that uniform getter later.

Two ideas if we do revisit this later:

1. Event-driven cache invalidation - `UniformState` (or Scene, or something) posts events whenever the view is changed. Models can listen and dirty their clipping plane matrices. Not sure if this is really any more elegant but just throwing it out there.
2. Don't calculate this matrix product on the CPU at all - defer work to the shader. The `inverseView3D` matrix is already available in the shader as `czm_inverseView3D`. We could pass the view-independent clipping matrix as another uniform (and it would not need to be recomputed on view/pass changes). In the shader, the existing `czm_transformPlane` would now be two calls, something like:

```GLSL
// before:
return czm_transformPlane(plane, transform);

// after: (two transforms)
vec4 refPlane = czm_transformPlane(plane, transform);
return czm_transformPlane(refPlane, model_clippingPlanesInverseTransposeView);
```

More work in the shader, but if we're CPU-bound, maybe worth it. (Also, if the extra compute is used to hide latency fetching memory, it might not even affect performance at all).


___

(technically, the above code could be slightly better. `czm_transformPlane` normalizes the result. No need to do that twice in a row. Instead, we could do):

```GLSL
vec4 refPlane = transform * plane;
return czm_transformPlane(refPlane, model_clippingPlanesInverseTransposeView);
```

Masty88:
On the polygons: both are fixed. It is the same shader fix for both, it just keys on the two defines (`HAS_CLIPPING_PLANES` and `ENABLE_CLIPPING_POLYGONS`), so the cast pass runs the discard for either. The matrix work was planes-only.

Makes sense on the caching. and left a short note on the getter pointing to this discussion. The shader-deferral idea is the one I like most, might give it a try as a follow-up. Thanks for the careful review.

mzschwartz5 (approved):
Looks good to me- thanks for the contribution and iteration! If you could fix conflicts with main (seems to just be in `CHANGES.md` - this change will go under the next release), I'll merge it in (assuming CI passes).

Masty88:
Conflicts with main are fixed, the CHANGES.md entry now sits under 1.143. many thanks for the help with this pr !

mzschwartz5:
@Masty88 I think you accidentally merged in some things unintentionally on that last commit (which is causing lint to fail as well).

Masty88:
Zut, sorry, bit of a boulette on my side with that merge — should be fixed now, the diff is back to just the shadow/clipping changes.

javagl:
@mzschwartz5  I think the changes in `CHANGES.md` here should have gone into a new (1.144) section (and I'm not sure if the conflicts are properly resolved in general). In doubt, I'd take another look tomorrow.

PR #13377, opened 2026-04-08
Lint: Add 'ast-grep' linter to catch common ES6 and JSDoc issues

donmccurdy:
# Description

Adds a new lint step in CI, based on [ast-grep](https://ast-grep.github.io/), to detect known errors related JSDoc / tsd-jsdoc / TypeScript syntax. Currently the linter enforces these custom rules:

- Legacy class declarations MUST have `@memberof` tags, without which tsd-jsdoc cannot associate them with the parent class
- ES6 class members MUST NOT have `@memberof` tags, which would cause tsd-jsdoc to silently drop some members
- ES6 class declarations MUST NOT `@constructor` or `@alias` tags, which do nothing
- ES6 classes MUST NOT [static field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#static_methods_and_fields) syntax, which tsd-jsdoc parses incorrectly

In my mind this is probably temporary tooling, until we resolve #10455. But if we find it helpful to add non-JSDoc-related lint rules and want to keep it around longer I have no objection, so far the lint step appears to be very fast.

~~Currently these tests fail on `main`. I'll fix that separately in upcoming PRs and merge this (if approved) only after the tests are passing.~~ Output looks like this:

Screenshot 2026-04-07 at 10 49 50 PM

## Issue number and link

- #13205 
- #13356
- #10455

## Testing plan

The newly-added rules currently ~~fail (as expected) because their violations haven't yet been fixed on main~~ pass:

- #13378
- #13376

~~Once all errors are fixed on main the tests on this PR should begin passing, and it can then be merged. We may additionally want to include [tests for the linter rules themselves](https://ast-grep.github.io/guide/test-rule.html). I haven't done that yet, but will add some if the PR otherwise looks good to everyone.~~

I've also added a directory containing tests that verify the rules are implemented correctly, which can be run with `sg test`. When adding new tests (where no snapshots yet exist) we'd instead run `sg test --skip-snapshot-tests`. See: https://ast-grep.github.io/guide/test-rule.html 

## Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [x] I have updated `CHANGES.md` with a short summary of my change
- [x] I have added or updated unit tests to ensure consistent code coverage
- [x] I have updated the inline documentation, and included code examples where relevant
- [x] I have performed a self-review of my code

## AI acknowledgment

- [ ] I used AI to generate content in this PR
[...]

PR #13390, opened 2026-04-09
Types: Convert Cartographic.js, Ellipsoid.js, MapProjection.js to ES6 classes

donmccurdy:
# Description

ES6 class conversions and type checking:

- Cartographic.js
- Ellipsoid.js
- MapProjection.js
- WebMercatorProjection.js
- GeographicProjection.js

I think this is the first example where we're defining and implementing an interface (MapProjection) and enforcing implementation compatibility with that interface using TypeScript, so there are some new-ish pieces in the migration here.

## Issue number and link

- #4434 
- #8359

## Testing plan

- `npm run tsc` should pass
- `npm run build-ts` should pass
- `npm run sg-scan` should pass (pending #13377)
- unit tests should pass
- confirm a few relevant sandcastles work as expected

<details>
<summary>Diff on `Cesium.d.ts`</summary>

```diff
diff --git a/Source/Cesium.main.d.ts b/Source/Cesium.d.ts
index 1b6118d40b..ad9d33e123 100644
--- a/Source/Cesium.main.d.ts
+++ b/Source/Cesium.d.ts
@@ -2685,10 +2685,6 @@ export class Cartographic {
      * @returns <code>true</code> if left and right are within the provided epsilon, <code>false</code> otherwise.
      */
     static equalsEpsilon(left?: Cartographic, right?: Cartographic, epsilon?: number): boolean;
-    /**
-     * An immutable Cartographic instance initialized to (0.0, 0.0, 0.0).
-     */
-    static readonly ZERO: Cartographic;
     /**
      * Duplicates this instance.
      * @param [result] - The object onto which to store the result.
@@ -2716,6 +2712,10 @@ export class Cartographic {
      * @returns A string representing the provided cartographic in the format '(longitude, latitude, height)'.
      */
     toString(): string;
+    /**
+     * An immutable Cartographic instance initialized to (0.0, 0.0, 0.0).
+     */
+    static readonly ZERO: Cartographic;
 }
 
 /**
@@ -6040,23 +6040,6 @@ export class Ellipsoid {
      * @returns A new Ellipsoid instance.
      */
     static fromCartesian3(cartesian?: Cartesian3, result?: Ellipsoid): Ellipsoid;
-    /**
-     * An Ellipsoid instance initialized to the WGS84 standard.
-     */
-    static readonly WGS84: Ellipsoid;
-    /**
-     * An Ellipsoid instance initialized to radii of (1.0, 1.0, 1.0).
-     */
-    static readonly UNIT_SPHERE: Ellipsoid;
-    /**
-     * An Ellipsoid instance initialized to a sphere with the lunar radius.
-     */
-    static readonly MOON: Ellipsoid;
-    /**
-     * An Ellipsoid instance initialized to a sphere with the mean radii of Mars.
-     * Source: https://epsg.io/104905
-     */
-    static readonly MARS: Ellipsoid;
     /**
      * The default ellipsoid used when not otherwise specified.
      * @example
@@ -6076,10 +6059,6 @@ export class Ellipsoid {
      * @returns The cloned Ellipsoid.
      */
     clone(result?: Ellipsoid): Ellipsoid;
-    /**
-     * The number of elements used to pack the object into an array.
-     */
-    static packedLength: number;
     /**
      * Stores the provided instance into the provided array.
      * @param value - The value to pack.
@@ -6096,13 +6075,6 @@ export class Ellipsoid {
      * @returns The modified result parameter or a new Ellipsoid instance if one was not provided.
      */
     static unpack(array: number[], startingIndex?: number, result?: Ellipsoid): Ellipsoid;
-    /**
-     * Computes the unit vector directed from the center of this ellipsoid toward the provided Cartesian position.
-     * @param cartesian - The Cartesian for which to to determine the geocentric normal.
-     * @param [result] - The object onto which to store the result.
-     * @returns The modified result parameter or a new Cartesian3 instance if none was provided.
-     */
-    geocentricSurfaceNormal(cartesian: Cartesian3, result?: Cartesian3): Cartesian3;
     /**
      * Computes the normal of the plane tangent to the surface of the ellipsoid at the provided position.
      * @param cartographic - The cartographic position for which to to determine the geodetic normal.
@@ -6241,6 +6213,34 @@ export class Ellipsoid {
      * @returns The approximate area of the rectangle on the surface of this ellipsoid.
      */
     surfaceArea(rectangle: Rectangle): number;
+    /**
+     * An Ellipsoid instance initialized to the WGS84 standard.
+     */
+    static readonly WGS84: Ellipsoid;
+    /**
+     * An Ellipsoid instance initialized to radii of (1.0, 1.0, 1.0).
+     */
+    static readonly UNIT_SPHERE: Ellipsoid;
+    /**
+     * An Ellipsoid instance initialized to a sphere with the lunar radius.
+     */
+    static readonly MOON: Ellipsoid;
+    /**
+     * An Ellipsoid instance initialized to a sphere with the mean radii of Mars.
+     * Source: https://epsg.io/104905
+     */
+    static readonly MARS: Ellipsoid;
+    /**
+     * The number of elements used to pack the object into an array.
+     */
+    static packedLength: number;
+    /**
+     * Computes the unit vector directed from the center of this ellipsoid toward the provided Cartesian position.
+     * @param cartesian - The Cartesian for which to to determine the geocentric normal.
+     * @param [result] - The object onto which to store the result.
+     * @returns The modified result parameter or a new Cartesian3 instance if none was provided.
+     */
+    geocentricSurfaceNormal(cartesian: Cartesian3, result?: Cartesian3): Cartesian3;
 }
 
 /**
@@ -7034,6 +7034,9 @@ export class GeocoderService {
     geocode(query: string, type?: GeocodeType): Promise<GeocoderService.Result[]>;
 }
 
+export interface GeographicProjection extends MapProjection {
+}
+
 /**
  * A simple map projection where longitude and latitude are linearly mapped to X and Y by multiplying
  * them by the {@link Ellipsoid#maximumRadius}.  This projection
@@ -7041,7 +7044,7 @@ export class GeocoderService {
  * is also known as EPSG:4326.
  * @param [ellipsoid = Ellipsoid.default] - The ellipsoid.
  */
-export class GeographicProjection {
+export class GeographicProjection implements MapProjection {
     constructor(ellipsoid?: Ellipsoid);
     /**
      * Gets the {@link Ellipsoid}.
@@ -9562,12 +9565,11 @@ export class LinearSpline {
  * Defines how geodetic ellipsoid coordinates ({@link Cartographic}) project to a
  * flat map like Cesium's 2D and Columbus View modes.
  */
-export class MapProjection {
-    constructor();
+export interface MapProjection {
     /**
      * Gets the {@link Ellipsoid}.
      */
-    readonly ellipsoid: Ellipsoid;
+    ellipsoid: Ellipsoid;
     /**
      * Projects {@link Cartographic} coordinates, in radians, to projection-specific map coordinates, in meters.
      * @param cartographic - The coordinates to project.
@@ -18467,13 +18469,16 @@ export class WallOutlineGeometry {
     static createGeometry(wallGeometry: WallOutlineGeometry): Geometry | undefined;
 }
 
+export interface WebMercatorProjection extends MapProjection {
+}
+
 /**
  * The map projection used by Google Maps, Bing Maps, and most of ArcGIS Online, EPSG:3857.  This
  * projection use longitude and latitude expressed with the WGS84 and transforms them to Mercator using
  * the spherical (rather than ellipsoidal) equations.
  * @param [ellipsoid = Ellipsoid.WGS84] - The ellipsoid.
  */
-export class WebMercatorProjection {
+export class WebMercatorProjection implements MapProjection {
     constructor(ellipsoid?: Ellipsoid);
     /**
      * Gets the {@link Ellipsoid}.
@@ -18493,19 +18498,6 @@ export class WebMercatorProjection {
      * @returns The Mercator angle.
      */
     static geodeticLatitudeToMercatorAngle(latitude: number): number;
-    /**
-     * The maximum latitude (both North and South) supported by a Web Mercator
-     * (EPSG:3857) projection.  Technically, the Mercator projection is defined
-     * for any latitude up to (but not including) 90 degrees, but it makes sense
-     * to cut it off sooner because it grows exponentially with increasing latitude.
-     * The logic behind this particular cutoff value, which is the one used by
-     * Google Maps, Bing Maps, and Esri, is that it makes the projection
-     * square.  That is, the rectangle is equal in the X and Y directions.
-     *
-     * The constant value is computed by calling:
-     *    WebMercatorProjection.mercatorAngleToGeodeticLatitude(Math.PI)
-     */
-    static MaximumLatitude: number;
     /**
      * Converts geodetic ellipsoid coordinates, in radians, to the equivalent Web Mercator
      * X, Y, Z coordinates expressed in meters and returned in a {@link Cartesian3}.  The height
@@ -18526,6 +18518,19 @@ export class WebMercatorProjection {
      * @returns The equivalent cartographic coordinates.
      */
     unproject(cartesian: Cartesian3, result?: Cartographic): Cartographic;
+    /**
+     * The maximum latitude (both North and South) supported by a Web Mercator
+     * (EPSG:3857) projection.  Technically, the Mercator projection is defined
+     * for any latitude up to (but not including) 90 degrees, but it makes sense
+     * to cut it off sooner because it grows exponentially with increasing latitude.
+     * The logic behind this particular cutoff value, which is the one used by
+     * Google Maps, Bing Maps, and Esri, is that it makes the projection
+     * square.  That is, the rectangle is equal in the X and Y directions.
+     *
+     * The constant value is computed by calling:
+     *    WebMercatorProjection.mercatorAngleToGeodeticLatitude(Math.PI)
+     */
+    static MaximumLatitude: number;
 }
 
 /**

```

</details>

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13390** 👈

### DIFF
--- a/packages/engine/Source/Core/Ellipsoid.js
+++ b/packages/engine/Source/Core/Ellipsoid.js
@@ -1,3 +1,5 @@
+// @ts-check
+
 import Cartesian2 from "./Cartesian2.js";
 import Cartesian3 from "./Cartesian3.js";
 import Cartographic from "./Cartographic.js";
@@ -7,6 +9,24 @@ import DeveloperError from "./DeveloperError.js";
 import CesiumMath from "./Math.js";
 import scaleToGeodeticSurface from "./scaleToGeodeticSurface.js";
 
+/** @import Rectangle from "./Rectangle.js"; */
+
+/**
+ * A real valued scalar function.
+ * @callback EllipsoidRealValuedScalarFunction
+ *
+ * @param {number} x The value used to evaluate the function.
+ * @returns {number} The value of the function at x.
+ *
+ * @private
+ */
+
+/**
+ * @param {Ellipsoid} ellipsoid
+ * @param {number} x
+ * @param {number} y
+ * @param {number} z
+ */
 function initialize(ellipsoid, x, y, z) {
   x = x ?? 0.0;
   y = y ?? 0.0;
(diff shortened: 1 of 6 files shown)


packages/engine/Source/Core/WebMercatorProjection.js
```diff
@@ -1,74 +1,135 @@
...
  * @see GeographicProjection
+ *
+ * @implements MapProjection
  */
-function WebMercatorProjection(ellipsoid) {
-  this._ellipsoid = ellipsoid ?? Ellipsoid.WGS84;
-  this._semimajorAxis = this._ellipsoid.maximumRadius;
-  this._oneOverSemimajorAxis = 1.0 / this._semimajorAxis;
-}
+class WebMercatorProjection {
```
donmccurdy, line 21: One unexpected output in `Cesium.d.ts` is that we're now exporting both an _interface_ and a _class_ for WebMercatorProjection and GeographicProjection:

```diff
+export interface WebMercatorProjection extends MapProjection {
+}
```

I can't think of a reason this would be a problem, but I do wish I understood why...

packages/engine/Source/Core/Ellipsoid.js
```diff
@@ -801,7 +835,7 @@ const weights = [
  *
  * @param {number} a The lower bound for the integration.
  * @param {number} b The upper bound for the integration.
- * @param {Ellipsoid~RealValuedScalarFunction} func The function to integrate.
```
donmccurdy, line 804: Slightly opinionated change here. TypeScript doesn't recognize this tilde syntax from JSDoc...

https://jsdoc.app/about-namepaths

... so I think just renaming this as a prefixed function name is fine for now. There are other ways to express this but I'm not sure of another way that about tsd-jsdoc and TypeScript support.

Thanks to @jjspace for explaining the namepath syntax!

packages/engine/Source/Core/WebMercatorProjection.js
ggetz: Should `MapProjection` be marked as both an abstract class (`@abstract`) _and_ as an interface (`@interface`)? 

I'm thinking the former is causing the output to make `WebMercatorProjection` _extend_ `MapProjection` (treating it as a class and inheriting it's other properties, including the interface tag) rather than having `WebMercatorProjection` _implement_ `MapProjection`.

Since `MapProjection` does not implement any base functionality, I think it's safe to treat it as in interface rather than an base class to extend.

donmccurdy: Hmm I think you're right that we shouldn't do both. It does seem like conflicting information, and we could remove `@abstract`.

This doesn't appear to change tsd-jsdoc's output though, it still creates both an interface and a class for WebMercatorProjection. Reading a bit more — in their tests and github discussions — it sounds like that might be intentional:

- https://github.com/englercj/tsd-jsdoc/blob/3c6e2dbfea7d56c02e62e520fee70f02cb428f03/test/expected/class_all.d.ts#L85
- https://github.com/englercj/tsd-jsdoc/issues/53

Perhaps related to:

- https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces

donmccurdy: Pushed a change removing `@abstract`. Maybe a good extra test for #13377, checking that things may be interfaces or abstract classes but not both.

lukemckinstry (approved):
This looks good to me 👍 
In https://github.com/CesiumGS/cesium/pull/13386/ I am now able to use the MapProjection type instead of needing to list out the child class projection types.

I saw a 👍 from @ggetz but I'll tag her here instead of merging immediately in case she wants to take another look at the area where she commented.

PR #13396, opened 2026-04-13
Remove legacy sandcastle

### LINKED ISSUE #13239: Update jsdom
jsdom listed a "new API" as part of the major version change from 27 to 28. We will need to check how this impacts Sandcastle before updating the dependency on our end.

jjspace:
# Description

This is it! The big purge!
As far as I am aware the swap for our e2e tests to the new sandcastle (https://github.com/CesiumGS/cesium/pull/13381) was the last true reason to keep the legacy sandcastle around. 

This PR removes the full legacy sandcastle as well as any build functions/scripts related to it.

It's a big PR simply for the deletion of the large gallery directories. I think it's safe to skip reviewing anything under `Apps/Sandcastle` since it's all just removals.

## Issue number and link

Part of https://github.com/CesiumGS/cesium-sandcastle/issues/9
Resolves https://github.com/CesiumGS/cesium/issues/13239

## Testing plan

- `git clean -dxf --exclude="/Specs/e2e/*-snapshots/"` just to start fresh
- `npm install`
- `npm start` should run fully, build sandcastle and start up
- Test the local dev server and make sure all pages still work as expected. Check docs and the Cesium Viewer app
    - There were previously some places we used files/images from the old Sandcastle dir. I believe I've removed all these
- Check CI and test the deployment. Everything should be built correctly
- run `npm run make-zip` and extract somewhere else like we do for release
    - Make sure the local server works there too

## Author checklist

## AI acknowledgment


Specs/e2e/cesium.html
```diff
@@ -10,7 +10,65 @@
   </head>
   <body>
     <style>
-      @import url(../../Apps/Sandcastle/templates/bucket.css);
+      /* This file is automatically rebuilt by the Cesium build process. */
```
ggetz: Is this content copied from a built file already? 

If so, should it reference the output path?

If not, should this live in its own css file? It could live next to `cesium.html`.

LICENSE.md
```diff
@@ -1019,26 +1019,6 @@ https://clipboardjs.com/
 >
 > THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-### JSHint
```
ggetz, line 1022: Double check the `Example Applications` section. I think you can also remove CodeMirror, and maybe clipboardjs if we're not using it anymore.

ggetz (changes requested):
🎉 🎉 🎉  

I haven't tested anything yet, but I have looked over the changes. 

Besides the inline comments, too suggestions from me:

1. We should be able to delete both CodeMirror and jsHint from the top-level `ThirdParty` directory as well.
2. That leaves TimelineDemo as the last bastion of dojo. Since it's not currently functional anyway, can we go ahead and remove the TimelineDemo from Apps as well? Assuming we're not using them anywhere else (which I'm fairly confident of, but please double check me), we should then be able to delete the `ThirdParty/` directory entirely.

Specs/e2e/cesium.html
jjspace: This was copied from the _previously_ built file, I've removed this comment. There's a similar version built for Sandcastle but it's a little more complicated now. It could be extracted to a CSS file but it's so small I felt it was ok to leave inline here for the few e2e tests that use it.

jjspace:
> Since it's not currently functional anyway, can we go ahead and remove the TimelineDemo from Apps as well

@ggetz What's not functional about the Timeline Demo? It seems to be working for me [and in CI](https://ci-builds.cesium.com/cesium/remove-legacy-sandcastle/Apps/TimelineDemo/index.html). It's a very simple page so I think we could re-implement fairly quickly with native components and JS. I definitely agree it would be great to fully ditch dojo

ggetz:
> What's not functional about the Timeline Demo? It seems to be working for me [and in CI](https://ci-builds.cesium.com/cesium/remove-legacy-sandcastle/Apps/TimelineDemo/index.html). It's a very simple page so I think we could re-implement fairly quickly with native components and JS. I definitely agree it would be great to fully ditch dojo

Ah, I misremembered. However, I don't think it's getting the visibility that justifies us keeping dojo. OK to remove?

jjspace:
> However, I don't think it's getting the visibility that justifies us keeping dojo. OK to remove?

@ggetz yes I think so. No need to keep dojo just for this. It was pretty easy to replace with a Sandcastle demo so we can still show it off when needed. I put this in a separate PR to try and help with reviewing: #13422. 

If that was the only remaining concern I think this can get merged?

ggetz:
`+66 | -207,455`  🙌

Looking good! Thanks @jjspace!

ggetz (approved):

PR #13397, opened 2026-04-13
Adds new package 'edit'

mzschwartz5:
# Description

=== DO NOT MERGE ===
Followed #13400

Adds a new npm package, 'edit', for upcoming mesh editor capabilities. 

## Issue number and link

#13398

## Testing plan

Make sure everything builds independently and together.

## Author checklist

- [] I have updated `CHANGES.md` with a short summary of my change

## AI acknowledgment

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1055,7 +1055,7 @@ async function bundleCSS(options) {
 
 const workspaceCssFiles = {
   engine: ["packages/engine/Source/**/*.css"],
-  edit: [],
+  edit: ["packages/edit/Source/**/*.css"],
   widgets: ["packages/widgets/Source/**/*.css"],
 };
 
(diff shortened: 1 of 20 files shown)


scripts/build.js
```diff
@@ -134,6 +134,19 @@ export const defaultESBuildOptions = () => {
...
+  widgets: path.join(projectRoot, "packages/widgets/index.js"),
+};
+
+const workspaceAliases = Object.fromEntries(
+  Object.entries(workspaceEntryPoints).map(([workspace, entryPoint]) => [
+    `@${scope}/${workspace}`,
+    entryPoint,
+  ]),
+);
+
```
mzschwartz5, line 149: This is something AI suggested. Basically, I believe the idea is that instead of resolving imports through `node_modules` during bundling, it rewrites the import to resolve to the local workspace under the correct package. This is supposedly less brittle because it doesn't depend on things all being installed first.

I'm no expert here so that's why I'm pointing it out.

mzschwartz5:
Leaving a note / thought: I've been running into some issues downstream with circular dependencies (edit <--> engine packages). At its root, this is because the engine package is doing too much. There are some parts of the engine that edit should really be a peer to, or even a dependency _of_, perhaps. But because the engine package is so monolithic, edit has to depend on all or nothing. 

I will likely need to consider either:
1. Splitting up engine into more packages (better for the future, but potentially a big lift)
2. Making edit a folder in engine rather than a package on its own (worsens the issue but easier short term)

Issue #13398, opened 2026-04-13
Editable Meshes


mzschwartz5:
## Overview

We're working on some support for mesh editing workflows: meshes whose vertices, edges, and faces can be moved around by user interaction, whose topology can be changed (create new edges, verts, etc.)

The implementation will likely be based on a half-edge mesh datastructure (also called [DCEL](https://en.wikipedia.org/wiki/Doubly_connected_edge_list)). (Note that this data structure doesn't support non-manifold meshes, so it may be worth looking into radial edge data structures too (like Blender's BMesh))

## API

Overall, I expect the class hierarchy / API will look something like:

`EditableMesh`
- `Vertex[]`
- `Face[]`
- `Edge[]`
  - `HalfEdge`

With various pointers to and from various components (striking a balance between minimizing data per component and achieving low algorithmic complexity in operations requiring adjacency information).

`EditableMesh` will only represent the CPU data structure of the mesh. The rendering representation will be owned and managed separately. Ideally, `EditableMesh` would thus be compatible with different rendering representations - that is, it could wrap around a `Primitive` or a (glTF) `Model` and work equally well. I'm not sure how feasible this is yet (it would definitely require some refactoring of those APIs - but maybe minimally, just to expose access to vertex and index buffers).

Internally, a clean class structure may be:
- `EditableMesh` - lives in `packages/edit`, owns the topological representation of the mesh. 
- `GeometryAccessor` - lives in `packages/engine`, either an interface implemented by renderable objects (like `Primitive`), or an object owned by a renderable object (i.e. inheritance vs. composition). Gives indirect get/set access to GPU vertex buffers without exposing the underlying layout (to maintain decoupling).

Renderable classes that want to be editable will need to be refactored to implement or own a `GeometryAccessor`. Then an `EditableMesh` can be constructed by passing in an instance of such a renderable. 

## Usage

Using `EditableMesh` would then look something like:

```JS

const primitive = new Primitive(..., {dynamicGeometry: true});
const accessor = primitive.geometryAccessor;
const mesh = new EditableMesh(accessor);

// Use `EditableMesh` API to update mesh components directly
// Or use a mouse listener to pick and drag vertices/edges/faces (potentially offered via API in future enhancement)
```

## Picking refactoring

To determine which mesh component is selected, we'll rely on Cesium's GPU picking machinery. The main issue here is that the current renderable classes (`Primitive` and `Model`) either:

a. Don't expose a way to assign a pick ID object (`Primitive`)
[...]

PR #13400, opened 2026-04-14
Half edge mesh

mzschwartz5:
# Description

=== DO NOT MERGE ===
Preceded by #13397

Introduces a half-edge mesh data structure for a CPU-side mesh editing representation. This data structure strikes a good balance between memory size and speed (for operations requiring adjacency information).

Also introduces a new interface `GeometryAccessor`. Any class that implements a `GeometryAccessSession` subclass can expose a `GeometryAccessor` that can create access sessions scoped to certain resources (read/write vertex attributes and topology). This is intended to be a new, very cross-compatible, general way to access low-level geometry data. See the comment [here](https://github.com/CesiumGS/cesium/pull/13400#discussion_r3113059178) for a discussion of its design.

Note: this PR outlines the new APIs. The implementations will follow in the next PR. This is mostly as a design practice (quasi-TDD), and to make review easier.

## Issue number and link

#13398

## Testing plan

No testing plan - this is just a scaffold for the implementation in the next PR.

## Author checklist

### DIFF
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -25,18 +25,26 @@ class Edge {
   }
 
   /**
-   * Returns the vertices that are incident to this edge.
+   * Returns the vertices that compose this edge.
    * @returns {Vertex[]}
    */
-  lower() {
+  vertices() {
     return [this._halfEdge.vertex, this._halfEdge.next.vertex];
+  }
+
+  /**
+   * MeshComponent method to return the edges that are part of this component. For an edge, this is just itself.
+   * @returns {Edge[]}
+   */
+  edges() {
+    return [this];
   }
 
   /**
    * Returns the faces that are incident to this edge.
    * @returns {Face[]}
    */
-  upper() {
+  faces() {
     const faces = [];
     if (this._halfEdge.face) {
       faces.push(this._halfEdge.face);
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -1,6 +1,6 @@
-/** @import { Cartesian3 } from "@cesium/engine"; */
 /** @import HalfEdge from "./HalfEdge"; */
 /** @import MeshComponent from "./MeshComponent"; */
+/** @import Face from "./Face"; */
 /** @import Vertex from "./Vertex"; */
 
 /**
@@ -25,11 +25,26 @@ class Edge {
   }
 
   /**
-   * Returns the vertices that compose this component. For an edge, this is the two vertices at its endpoints.
+   * Returns the vertices that are incident to this edge.
    * @returns {Vertex[]}
    */
-  vertices() {
+  lower() {
     return [this._halfEdge.vertex, this._halfEdge.next.vertex];
+  }
+
+  /**
+   * Returns the faces that are incident to this edge.
+   * @returns {Face[]}
+   */
+  upper() {
+    const faces = [];
+    if (this._halfEdge.face) {
+      faces.push(this._halfEdge.face);
+    }
+    if (this._halfEdge.twin.face) {
+      faces.push(this._halfEdge.twin.face);
+    }
+    return faces;
   }
 }
 
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -29,7 +29,7 @@ class Edge {
    * @returns {Vertex[]}
    */
   vertices() {
-    return [this._halfEdge.vertex, this._halfEdge.twin.vertex];
+    return [this._halfEdge.vertex, this._halfEdge.next.vertex];
   }
 }
 
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -26,33 +26,38 @@ class Edge {
 
   /**
    * Returns the vertices that compose this edge.
+   * @param {Vertex[]} result Destination array.
    * @returns {Vertex[]}
    */
-  vertices() {
-    return [this._halfEdge.vertex, this._halfEdge.next.vertex];
+  vertices(result) {
+    result.push(this._halfEdge.vertex);
+    result.push(this._halfEdge.next.vertex);
+    return result;
   }
 
   /**
    * MeshComponent method to return the edges that are part of this component. For an edge, this is just itself.
+   * @param {Edge[]} result Destination array.
    * @returns {Edge[]}
    */
-  edges() {
-    return [this];
+  edges(result) {
+    result.push(this);
+    return result;
   }
 
   /**
    * Returns the faces that are incident to this edge.
+   * @param {Face[]} result Destination array.
    * @returns {Face[]}
    */
-  faces() {
-    const faces = [];
+  faces(result) {
     if (this._halfEdge.face) {
-      faces.push(this._halfEdge.face);
+      result.push(this._halfEdge.face);
     }
     if (this._halfEdge.twin.face) {
-      faces.push(this._halfEdge.twin.face);
+      result.push(this._halfEdge.twin.face);
     }
-    return faces;
+    return result;
   }
 }
 
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -22,12 +22,6 @@ class Edge {
   get halfEdge() {
     return this._halfEdge;
   }
-
-  /**
-   * Move the edge to a new position.
-   * @param {Cartesian3} newPosition
-   */
-  move(newPosition) {}
 }
 
 export default Edge;
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -1,0 +1,13 @@
+import MeshComponent from "./MeshComponent.js";
+
+/**
+ * Edge record for an EditableMesh.
+ *
+ * @alias Edge
+ * @constructor
+ *
+ * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
+ */
+class Edge extends MeshComponent {}
+
+export default Edge;
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -25,6 +25,26 @@ class Edge {
   }
 
   /**
+   * Returns the endpoint vertices of this edge. Equivalent to
+   * {@link Edge#vertices}.
+   * @param {Vertex[]} result Destination array.
+   * @returns {Vertex[]}
+   */
+  lower(result) {
+    return this.vertices(result);
+  }
+
+  /**
+   * Returns the faces incident to this edge. Equivalent to
+   * {@link Edge#faces}.
+   * @param {Face[]} result Destination array.
+   * @returns {Face[]}
+   */
+  upper(result) {
+    return this.faces(result);
+  }
+
+  /**
    * Returns the vertices that compose this edge.
    * @param {Vertex[]} result Destination array.
    * @returns {Vertex[]}
@@ -32,16 +52,6 @@ class Edge {
   vertices(result) {
     result.push(this._halfEdge.vertex);
     result.push(this._halfEdge.next.vertex);
-    return result;
-  }
-
-  /**
-   * MeshComponent method to return the edges that are part of this component. For an edge, this is just itself.
-   * @param {Edge[]} result Destination array.
-   * @returns {Edge[]}
-   */
-  edges(result) {
-    result.push(this);
     return result;
   }
 
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -1,3 +1,7 @@
+/** @import { Cartesian3 } from "@cesium/engine"; */
+/** @import HalfEdge from "./HalfEdge"; */
+/** @import MeshComponent from "./MeshComponent"; */
+
 /**
  * Edge record for an EditableMesh.
  * Unlike a HalfEdge, an Edge is purely topological - containing data specific to the edge itself, not its relationship to other components.
@@ -7,7 +11,10 @@
  * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
  */
 class Edge {
-  // Accepts a canonical half-edge from the pair of half-edges that make up this edge.
+  /**
+   * Accepts a canonical half-edge from the pair of half-edges that make up this edge.
+   * @param {HalfEdge} halfEdge
+   */
   constructor(halfEdge) {
     this._halfEdge = halfEdge;
   }
@@ -16,6 +23,10 @@ class Edge {
     return this._halfEdge;
   }
 
+  /**
+   * Move the edge to a new position.
+   * @param {Cartesian3} newPosition
+   */
   move(newPosition) {}
 }
 
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -1,6 +1,7 @@
 /** @import { Cartesian3 } from "@cesium/engine"; */
 /** @import HalfEdge from "./HalfEdge"; */
 /** @import MeshComponent from "./MeshComponent"; */
+/** @import Vertex from "./Vertex"; */
 
 /**
  * Edge record for an EditableMesh.
@@ -21,6 +22,14 @@ class Edge {
 
   get halfEdge() {
     return this._halfEdge;
+  }
+
+  /**
+   * Returns the vertices that compose this component. For an edge, this is the two vertices at its endpoints.
+   * @returns {Vertex[]}
+   */
+  vertices() {
+    return [this._halfEdge.vertex, this._halfEdge.twin.vertex];
   }
 }
 
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -1,13 +1,22 @@
-import MeshComponent from "./MeshComponent.js";
-
 /**
  * Edge record for an EditableMesh.
+ * Unlike a HalfEdge, an Edge is purely topological - containing data specific to the edge itself, not its relationship to other components.
+ * For example, an Edge may contain selection state or a crease flag. This pertains to both HalfEdges in the pair, so it makes sense to have a single Edge record for both.
  *
- * @alias Edge
- * @constructor
- *
+ * @implements MeshComponent
  * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
  */
-class Edge extends MeshComponent {}
+class Edge {
+  // Accepts a canonical half-edge from the pair of half-edges that make up this edge.
+  constructor(halfEdge) {
+    this._halfEdge = halfEdge;
+  }
+
+  get halfEdge() {
+    return this._halfEdge;
+  }
+
+  move(newPosition) {}
+}
 
 export default Edge;
--- a/packages/edit/Source/Mesh/Edge.js
+++ b/packages/edit/Source/Mesh/Edge.js
@@ -1,22 +1,24 @@
 /** @import HalfEdge from "./HalfEdge"; */
-/** @import MeshComponent from "./MeshComponent"; */
 /** @import Face from "./Face"; */
 /** @import Vertex from "./Vertex"; */
+
+import MeshComponent from "./MeshComponent";
 
 /**
  * Edge record for an EditableMesh.
  * Unlike a HalfEdge, an Edge is purely topological - containing data specific to the edge itself, not its relationship to other components.
  * For example, an Edge may contain selection state or a crease flag. This pertains to both HalfEdges in the pair, so it makes sense to have a single Edge record for both.
  *
- * @implements MeshComponent
+ * @extends MeshComponent
  * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
  */
-class Edge {
+class Edge extends MeshComponent {
   /**
    * Accepts a canonical half-edge from the pair of half-edges that make up this edge.
    * @param {HalfEdge} halfEdge
    */
   constructor(halfEdge) {
+    super();
     this._halfEdge = halfEdge;
   }
 
--- a/packages/edit/Source/Mesh/EditableMesh.js
+++ b/packages/edit/Source/Mesh/EditableMesh.js
@@ -4,11 +4,52 @@
  * EditableMesh owns the CPU-side topology with a loose coupling to the render-side geometry
  * via a mapping of topological vertex IDs to render vertex IDs.
  *
- * @alias EditableMesh
- * @constructor
- *
  * @experimental This feature is not final and is subject to change without Cesium's standard deprecation policy.
  */
-class EditableMesh {}
+class EditableMesh {
+  constructor(geometryAccessor) {
+    this._geometryAccessor = geometryAccessor;
+    this._vertices = [];
+    this._edges = [];
+    this._faces = [];
+    this._halfEdges = [];
+
+    this.#buildMesh(geometryAccessor);
+  }
+
+  get vertices() {
+    return this._vertices;
+  }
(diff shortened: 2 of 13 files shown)


packages/edit/Source/Mesh/Edge.js
```diff
@@ -0,0 +1,13 @@
+import MeshComponent from "./MeshComponent.js";
+
+/**
+ * Edge record for an EditableMesh.
+ *
+ * @alias Edge
+ * @constructor
```
donmccurdy: Minor - the `@alias` and `@constructor` tags are not needed on ES6 classes (unless the alias is something other than the class name). If we merge the custom linter rules proposed in #13377, the linter will issue warnings on these tags.

mzschwartz5: Good to know! I'll update that, thanks.

packages/engine/Source/Scene/GeometryAccessor.js
mzschwartz5, line 1: When reviewing this class, you may think "why is this so overengineered?" And maybe you're right to think that (please, let me know)! But I had many design constraints I was trying to satisfy, and JavaScript is not super feature rich... I think I still managed to keep the API simple (ish), for both implementers of the interface and consumers of it.

The design goals were:

1. Allows different geometry-owning classes to expose access to geometry via a common interface.
2. Not explicitly tied to a specific geometry representation. It could theoretically work with geometry backed by gpu buffers or cpu buffers, flat arrays or some other data structure. This also makes it easy to test.
3. Expensive operations (like getting data back from the GPU) should be done infrequently.
4. Resources clean up themselves (RAII) - leaks are not possible, and consumers don't need to worry about explicit clean up.
5. No overhead is added to geometry-owning classes when users don't need editing workflows.
6. No overhead in methods that get or set geometry - these are hot code paths, validations should be done up-front.
7. Consumers can read and write geometry in the same session.
8. Consumers should be restricted from calling functions they don't have access to. (This is difficult to strongly enforce in JS, but it should be strongly hinted at, at least).
9. Separate concerns: topology vs. attributes

packages/edit/Source/Mesh/EditableMesh.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Get a face by index.
+   * @param {number} index
+   * @returns {Face}
+   */
+  getFace(index) {
+    return getElement(this._faces, index);
+  }
```
mzschwartz5, line 82: I'm not sure yet if it's actually useful to have getters by index on these components. The way I expect editing to work, clicking will pick a mesh component, and the returned pick ID will just contain a reference to the component. So not sure when you would want to get a component by ID (unless iterating over all of them, maybe? In which case maybe I should also expose methods to get `numberOf<Component>`)

packages/edit/Source/Mesh/EditableMesh.js
mzschwartz5, line 1: Reference image for how a sample mesh would be represented as a half-edge mesh structure.

Image

Note that there are two connections not shown - the reference from a given face to an (arbitrary) half-edge on its boundary, and the reference from a given vertex to an (arbitrary) half-edge that points to it.

Image

[Credit](https://docs.google.com/presentation/d/1U_uiZ3Jbc_ltHMhWBAvBDe4o0YKAAD3RNLXI3WwGmnE/edit?slide=id.g2704aba97d_0_97#slide=id.g2704aba97d_0_97)

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
...
+  /**
+   * Updates the render vertex index for a triangle corner.
+   *
+   * @param {number} triangleIndex The triangle index.
+   * @param {number} vertexIndex The vertex index within the triangle (0, 1, or 2).
+   * @param {number} renderVertexIndex The new render vertex index.
+   */
+  setTriangleVertexIndex(triangleIndex, vertexIndex, renderVertexIndex) {
+    DeveloperError.throwInstantiationError();
+  }
```
mzschwartz5, line 288: These are kind of placeholders until I figure out what topology operations should exist on the accessor level. `getTriangleCount` is useful (unless this geometry is drawing lines or points...), but the other two functions here are pretty meaningless.

What are the basic operations the accessor should provide on the index buffer of some geometry? Swap two indices? Add, remove, insert new indices (in sets of 1, 2, or 3 depending on primitive type)? Is there a general API here or must it be specific to the primitive type?

mzschwartz5: The basic topology API  _should_ probably be just get/add/remove/insert indices. 

The real question is how whether to handle various primitive types (triangles / lines / points) - and, if so, how.

The idea of this interface was to be agnostic of the underlying geometry representation. So, in theory, the question of primitive type should maybe be punted to the render layer... but what about `EditableMesh`? How can it construct a topological mesh without knowing the primitive type?

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
+// @ts-check
+
+import DeveloperError from "../Core/DeveloperError.js";
+import oneTimeWarning from "../Core/oneTimeWarning.js";
+
+/**
+ * @typedef {object} GeometryAttributeDescriptor
+ * @property {string} semantic The attribute semantic.
+ * @property {number} [setIndex] The attribute set index.
+ */
```
donmccurdy, line 10: Note that the typedefs in this file will appear in public documentation and type definitions — if that's not intended (up to you!) they can be annotated with `@ignore`. 

If the types are used in other public type definitions, e.g. function parameters,  they would need to be public. I believe `tsd-jsdoc` should fail if the public type definitions depend on something internal/private.

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
...
+ * });
+ * ```
+ */
+class GeometryAccessor {
+  /**
+   * @param {GeometryAccessSessionConstructor} accessSessionClass The class that implements GeometryAccessSession for this accessor.
+   * @param {object} [accessSessionOptions] Options to pass to created GeometryAccessSessions.
+   */
+  constructor(accessSessionClass, accessSessionOptions) {
+    if (!(accessSessionClass.prototype instanceof GeometryAccessSession)) {
```
donmccurdy, line 95: I don't think I've seen the pattern `x.prototype instanceof X` before, would this work without including the `.prototype` property?

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
...
+   *
+   * @param {GeometryAccessScopes} scopes The requested access scopes for this session.
+   * @param {GeometryAccessSessionCallback} callback The callback to run.
+   */
+  withSession(scopes, callback) {
+    const accessSession = new this._geometrySessionClass(
+      this,
+      scopes,
+      this._accessSessionOptions,
+    );
```
donmccurdy, line 116: We have the ESLint [`new-cap`](https://eslint.org/docs/latest/rules/new-cap) rule enabled, meant to say that anything constructed with `new` should start with a capital letter. It looks like the underscore here prevents ESLint from complaining, but in the spirit of the rule maybe this should be `_GeometrySessionClass`?

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
...
+function createAccessErrorFunction(methodName) {
+  return function () {
+    throw new DeveloperError(
+      `${methodName} is not available for the provided access scopes.`,
+    );
+  };
+}
+
+export default GeometryAccessor;
+export { GeometryAccessSession };
```
donmccurdy, line 315: I think `GeometryAccessSession` will end up missing from the compiled `index.js` here — the build system doesn't currently re-export non-default exports. If you only need to use it internally that's fine though!

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
+// @ts-check
+
+import DeveloperError from "../Core/DeveloperError.js";
+import oneTimeWarning from "../Core/oneTimeWarning.js";
+
+/**
+ * @typedef {object} GeometryAttributeDescriptor
+ * @property {string} semantic The attribute semantic.
```
donmccurdy, line 8: Could include a type-only import of the semantic enum below other imports at top of file:

```javascript
/** @import VertexAttributeSemantic from "./VertexAttributeSemantic.js" */
```

And then reference it here:


```suggestion
 * @property {VertexAttributeSemantic} semantic The attribute semantic.
```

Note that this won't have much effect quite yet, still fixing up enum-based type checking in https://github.com/CesiumGS/cesium/pull/13203.

packages/engine/Source/Scene/GeometryAccessor.js
```diff
@@ -0,0 +1,315 @@
...
+
+/**
+ * @callback GeometryAccessSessionCallback
+ * @param {GeometryAccessSession} accessSession The active access session.
+ */
+
+/**
+ * @callback GeometryAttributeReader
+ * @param {number} vertexIndex The vertex index to read.
+ * @returns {*} The vertex attribute value.
```
donmccurdy, line 32: What return types would you expect here? Just curious — not suggesting that you need to specify this in the types. If bulk operations on many vertices are an expected use case, it might be important to:

- (a) allow the caller to pass a 'result' argument, or ...
- (b) provide a view (TypedArray) of many vertices at once rather than creating Cartesian3 / Color / etc. instances for each vertex

donmccurdy (commented):
Left a couple comments mainly related to JSDoc and type definitions in GeometryAccessor.js, but really nothing particularly important at this stage! The typedef usage looks solid. There are a few things we might be able to simplify later in the migration process (using TS-only features that tsd-jsdoc doesn't support) perhaps.

packages/engine/Source/Scene/GeometryAccessor.js
mzschwartz5: Good to know- I _was_ wondering about that. I'll @ignore most of them I think (there is one at least that needs to be public).

packages/engine/Source/Scene/GeometryAccessor.js
mzschwartz5: Good points/questions. A result argument might be the move here, need to think about it some more.

In general, the return type depends on the semantic requested. If I request a vertex attribute for texture coords, I'd expect back a `Cartesian2`, perhaps. If I ask for position, I'd expect back a `Cartesian3`. I'm not sure if there's a better way to type this method... if this were C++ I would express this with templates.

As for the concern over performance - I expect subclasses to implement these methods by pulling GPU data back to the CPU all at once, into some typed view, during construction of this accessor session. Then, whenever a vertex attribute value is requested, the implementation should just return the appropriate value from the typed array. (Of course, it's up to the implementation but that would be the efficient way to do it). So this method should probably accept a result argument, as you suggested.

packages/engine/Source/Scene/GeometryAccessor.js
mzschwartz5: So, I believe it's necessary. You use `x instanceof X` when you want to check if an object is an instance of a class. You use `x.prototype instanceof X` when `x` is a constructor function and you want to check if the thing it constructs is an instance of `X`.

That is:

`obj instanceof Class` --> for objects
`ClassA.prototype instanceof ClassB` → for class inheritance checks

Full transparency - this code was originally suggested to me by copilot. But this is my current understanding of it.

packages/engine/Source/Scene/GeometryAccessor.js
javagl: The pattern that was suggested in another PR was 
```
const GeometrySessionClass = this._geometrySessionClass;
const accessSession = new GeometrySessionClass(...)
```
that seems to spark joy for the linter

packages/engine/Source/Scene/GeometryAccessor.js
javagl: This, in combination with the `@ignore` comment above, is more broadly about what should be part of the exposed (and public...) API. (I just try to draw the connection - an actual review is still pending)

packages/engine/Source/Scene/GeometryAccessor.js
javagl: I _think_ that this may fall apart **iff** this is supposed to handle custom stuff like `_MY_SPECIAL_METADATA` (or even just `COLOR_666`), but I'll have to check the details here.

packages/engine/Source/Scene/GeometryAccessor.js
donmccurdy: We can also use templates in JSDoc...

```javascript
/**
 * @template T
 * @param {number} index
 * @param {T} result
 * @returns T
 */
function getValue(index, result) {
  return result;
}
```

... but I believe it's only safe for internal/private signatures due to tsd-jsdoc. May or may not be worth the trouble in this case.

packages/engine/Source/Scene/GeometryAccessor.js
donmccurdy: Ahh interesting! I missed that `accessSessionClass` was the class/constructor reference here — that makes sense.

javagl (commented):
I looked at the PR "as a whole". Given the 'Draft' state, it's not _entirely_ clear at which "level" the review should happen, but a few thoughts:

---

The separation between the CPU-side reprentation in general, and the underlying data (via the `GeometryAccessor`) certainly makes sense, insofar that it establishes a clear boundary and separation of concerns. Technically, the `EditableMesh` currently requires a `GeometryAccessor` in its constructor. But it is not (directly) stored as a property or so. It is only passed to `buildMesh`. I'd have to think through how this method will be implemented, but the point is that I _think_ that the goal was to have things separated enough so that one _could_ pull that into some
`const mesh = Magic.createFrom(geometryAccessor);`
function, so that the `EdiableMesh` does not even know that the `GeometryAccessor` exists.

---

The specific _choice_ for the CPU-side representation - that of a half-edge data structure - can be justified, technically: It is a common data structure with good trade-offs for memory and performance, and (important:) constant-time adjacency access. However, some scrutiny may be warranted in two areas:

- It can only represent _manifold_ meshes. I think that ~"some random glTF model" will not necessarily be manifold (and could imagine that some photogrammetry tiler output in fact isn't). And I wonder how ~"disruptive" it would be when we encounter such a mesh. (Related to the question about the `buildMesh` above
- It is in fact only for triangle meshes (no vector data, for example)

So this _does_ limit the applicability, but maybe that's a conscious decision: Trying to create the "eierlegende wollmilchsau"-mesh class can be a rabbit hole, and one _could_ justify limiting the scope here - roughly, by saying that _if_ that was supposed to be applied for something like vector data, one would create a dedicated, lean, clean `EditableVectorData` class for dealing with that. Given the clean separation between the mesh- and the `GeometryAccessor`, this would in fact be possible.

---

I'm not sure about the role of the `MeshComponent` interface. It currently only has this `move(newPosition)` method. And I assume that this is some sort of "placeholder" for... whatever clients should be able to do with such a mesh component. 

(If not, or if this `move(newPosition)` method should actually be "the" "final" thing there, I'd ask many questions. Otherwise, I assume that some these questions are already on some backlog, at least implicitly)

---

The "indexed access" in `EditableMesh` (where you added an inlined comment): I think that this could be a pretty big deal. First of all, if that indexed access should be offered, then yes, you'd need some `numberOf...` methods. But more importantly: This raises a bunch of questions about which editing operations should be possible. 

(You also made a connection to 'picking' there, but given the complexity of picking as a whole (in CesiumJS), I'd hesitate to couple any decisions to that...)

---

The `GeometryAccessor` class: This is where I currently see some of the more fundamental decisions. 

The `typedef`s a are a bit to chew on. (I'd have to take closer looks at the generated TypeScript - maybe in a second review pass). But the inlined `@example` snippets helped to illustrate their purposes and usage patterns.

The concept of that `withSession` looks like a clean way to handle the questions about scoping and lifetime. (It reminded me a bit of [`ResourceScope`](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.incubator.foreign/jdk/incubator/foreign/ResourceScope.html), but only tangentially).

The genericity of the 
`vertexAttributeAccessor = reader + writer` 
structure, with the 'descriptor' being passed in, and both being optional, looks like it could cover many different forms of access.

The points about typing, with the `{*}` return type and the possible optional `result` parameters have already been brought up in inlined comments. There are a few dimensions to think through. 
For example, you mentioned that one would expect a `Cartesian3` for a `POSITION`. But I wonder how that could be applied to other properties, where 1. no "established data type" exists (e.g. some custom attribute) or 2. the user would like to access it  in a different form. 

I assume that one of the first flavors of the implementations there will operate on `Buffer`/`Float32Array`. One could make a case to offer the data in the "flattest possible form". When you have a function like `get(index) : number`, you can trivially implement it based on a `number[]` array, or based on a `Cartesian3` array. When you _have_ to return a `Cartesian3` object, then this raises the question about allocations and `result` parameters. Conversely, having "the expected type" there (like `Cartesian3`) could offer some convenience.

Brainstorming (_this_ may go too far, in terms of 'overengineering'): It would in theory be possible to add some 'hint' to the descriptor, to indicate what structure that the user wants. But maybe that causes more trouble than it's worth.

---

I think that the differentiation of 1. attribute read/write and 2. topology read/write is important as well. But the topology-related functions (`getTriangleCount/get/setTriangleVertexIndex`) look surprisingly specific right now. Before musing about this or its implications (e.g. again for things like "vector data" or so), I wonder: In how far is that _intentionally_ targeting triangle meshes in particular, or whether this was primarily to illustrate the separation per se.

---

EDIT: Small addendum: I wonder if a `GeometryAccessSession` might require a GL context, and how it would get there. But maybe that's still to be sorted out.

packages/engine/Source/Scene/GeometryAccessor.js
mzschwartz5: I'm going to have to move away from design goal 4 (RAII), sadly. I thought RAII could be emulated in JS via the `withSession` design, but it makes it really difficult to use the class with long running sessions.

Someday, when Safari supports it, `Symbol.dispose` could be used to implement true RAII - until then, I'm going to offer an `openSession` method that is less safe (though still fine TBH).

javagl: I'm a bit curious what a "long running session" refers to. Does it refer to ~"time" as in "over multiple frames" or... more specifically: "From the point in time where a user _presses_ a button, until the user _releases_ it"? 

(In some way, the session-based scoping looked like it was _intended_ to "limit the lifetime", and guard the contained things ... roughly like a database transaction)

mzschwartz5: It does refer to over time / multiple frames. The `withSession` idea was nice if you need to, say, read or write lots of data all at once. Great for building up the half-edge mesh structure. But now say we want to edit the structure in a user-interactive session. For the whole duration of time that the user is making edits, we probably want to keep the vertex buffer data on the CPU. That way, whenever an edit is made, we can update that CPU buffer, then upload to the GPU. If a "session" only persists while the user is dragging a vertex, then we'll be moving data to and from the CPU/GPU quite frequently. By contrast, if the session persists the whole time editing is enabled, then we can keep that data resident on the CPU and just flush updates to the GPU as necessary.

javagl: OK, there are probably many lower-level aspects of the intended and required workflows that I don't fully have on the radar. E.g. when an "editing operation" _starts_, when it _ends_, and what happens (or has to happen) in between in terms of "updates". This also depends on the granularity (rotating a whole mesh vs. dragging a single vertex), or whether the edits affect topology or not. (On some low level, it boils down to where and when exactly data is copied between CPU<->GPU, roughly related to the quesiton about where some 'GL context' might come into play with all this). Curious to see the next updates here.

donmccurdy:
> It can only represent manifold meshes.

I hadn't read much about this before, but found this summary helpful:

https://kaba.hilvi.org/homepage/blog/halfedge/halfedge.htm

The article suggests that loose points and edges might be OK, but does advise against non-manifold meshes in a half-edge.

> The typedefs a are a bit to chew on.

Aside, and optional, it's also possible to define object types with a class annotated with `@interface`. I have no particular preference here, but maybe it's easier to read. See:

https://github.com/CesiumGS/cesium/blob/2165ff484489761431b0424e711417553ebcf992/packages/engine/Source/Core/MapProjection.js#L18-L25

javagl:
Random drive-by comment from receiving the GitHub notifications:

Even though the exact intended usage of 'highlightable' is not (yet) visible from the changes, I wonder whether this is (supposed to be) part of the "editor model", or rather a specific aspect of the UI side. I assume that in the context of the editor, these objects are _~selected_ (with the intention to let them be affected by editing operations). Whether (and how, and with which color) this selection is indicated in the UI is unrelated to that. 

This is related to whether or not "being selected" is a property of the selected object itself, or of a "higher power", and (important:) who _knows_ about an object being "selected" (Roughly: Whether there should be some `getHighlight` in addition to `setHighlight`). I've seen several software projects where the ... "longer-term evolution" of some `TeenyWeenyDataObject` consisted of adding  `isSelected`,  `isHighlighted`, `isFocused`, ... `isSelectedForMovement`, `isSelectedForDeletion`, ... `hasTheFavoriteColorOfTheUser` ), and where some `Set<ReallyTeenyWeenyDataObject> selectedObjects` could have kept the state space simpler and maintained in one place.

**But...** if the _state_ and the _process_ (!) of "being selected" already is supposed to be maintained by a higher-level structure (i.e. by the `Editor`), and the `Highlightable` is _only_ intended for passing that information to some rendered representation, then that may be fine.

mzschwartz5:
@javagl Not sure I exactly understand the question, but I can talk about the intention here:

Ultimately, the purpose here is to show some highlight when an object is selected. As you voiced, it's a good idea to separate concerns such as editing and UI. That's why I introduced this as an interface - so that the details of _how_ an object is highlighted are separated from the actual highlighting of the object. Eventually, I introduce a new widget and viewer mixin that provides a minimal UI to facilitate the editing experience. This mixin is completely optional - like all of our mixins / widgets, it's just a fast way for users to prototype, but ultimately is intended to be replaced by production application UI building on the same underlying APIs.

In this particular case, the mixin does scene picking when the user clicks, determines if the picked object implements the `Highlightable` interface, and uses that knowledge to set a highlight. For the Model class in particular, I implemented this in terms of the silhouette API - so the Model class owns the render representation, the mixin just owns the selection state.

So, overall, I think I'm mostly addressing the concerns you voiced.

javagl:
Yes, it sounds like this is really a one-direction "channel", with one important point being that the editor only accepts "something that implements this interface", and hands over the state. The editor itself doesn't really care whether an object is "highlighted" or not (i.e. it never needs some backward `isHighlighted` check). 

There are questions about granularity and such. Yes, a `Model` can be `Highlightable`, but I wonder whether this has to be broken down into smaller things...

Cesium Highlightable

Where and how an edge or vertex could support being 'Highlightable' isn't clear yet, but maybe it will become clearer with upcoming PRs.

Also... is see the point of passing in a color and intensity, but in terms of state and responsibilities, one could argue to boil that down to a `boolean`. The editor does not (or maybe should not) care whether "highlighting" means that something is shown with a red silhouette, or by showing some outline+tint (like the face in blender), or by simply drawing a dashed-line (2D) box around its on-screen-representation...

(This is *not* suggesting that this _should_ be changed! Just brainstorming about which requirements might come up)

mzschwartz5:
> Where and how an edge or vertex could support being 'Highlightable' isn't clear yet, but maybe it will become clearer with upcoming PRs.

So, the `Highlightable` interface is simply for object highlighting (equivalent to what happens in Blender if you click something in Object mode). Component-level highlighting is fully owned by the editing package. I still try to keep it decoupled from the editing experience itself, though. (Arguments could be made for the `TopologyOverlay` class I introduce later to be part of the engine package... will have to think about that)

> Also... is see the point of passing in a color and intensity, but in terms of state and responsibilities, one could argue to boil that down to a boolean

That's a fair point. I chose the current shape because it felt like a generalization of Model's silhouette interface, but we could definitely improve on it.

PR #13404, opened 2026-04-14
MVTDataProvider: Support for Mapbox Vector Tiles (MVT)

danielzhong:
# Description
This PR introduces an MVT loading pipeline that decodes MVT tile payloads, converts them to vector glTF, and renders them through `VectorGltf3DTileContent` in the 3D Tiles runtime path.

It also improves missing-tile handling by treating HTTP 404/204 MVT tile responses as empty content, allowing sparse datasets to render gracefully.

image

https://github.com/user-attachments/assets/cefa4d6c-aef7-48cd-9ff7-fa1948ca4dd3

## Overall Pipeline

In `MVTDataProvider.js`, `MVTDataProvider.fromUrlTemplate(...)` creates the provider.

**`MVTDataProvider`** (extends `UrlTemplate3DTilesDataProvider`)
- Handles configuration (`urlTemplate` / `minZoom` / `maxZoom` / `extent` / `featureIdProperty`), optionally hydrating defaults from `metadata.json`
- Builds a runtime-generated 3D Tiles tileset that maps `{z}/{x}/{y}` URL templates to tile headers
- Implements `_createCodec()` to inject an MVT-aware **runtime content codec** onto `tileset._runtimeContentCodec`. The codec carries:
  - `createContent(tileset, tile, resource, arrayBuffer)` — how to turn raw bytes into a `Cesium3DTileContent`
  - `disableSkipLevelOfDetail: true` — vector tiles do not support skip-LOD
  - `missingTilePolicy: { statusCodes: [404, 204], urlPattern }` — sparse datasets render gracefully

**`Cesium3DTile` / `Multiple3DTileContent`** (runtime, format-agnostic)
- On each tile load, `makeContent(...)` checks `tileset._runtimeContentCodec`
- If a codec is present, dispatches directly to `codec.createContent(...)` — bypassing magic-byte sniffing in `preprocess3DTileContent` and the `Cesium3DTileContentFactory` lookup
- If a request fails, `isMissingTileContentError` consults `codec.missingTilePolicy` to convert HTTP 404 / 204 responses into `Empty3DTileContent`
- Runtime never names "MVT" — adding new vector formats (MLT, GeoJSON tiles, etc.) only requires a new provider with its own codec

**MVT codec** (the closure returned by `MVTDataProvider#_createCodec`)
- Receives raw bytes for a single MVT tile (URL extension is irrelevant)
- Calls `decodeMVT(...)` to parse tile-local geometry and properties from the protobuf payload (with bounds-checked varints and truncation / over-long defenses)
- Parses `{z}/{x}/{y}` from the tile URL (suffix optional; supports query / fragment)
- Calls `buildVectorGltfFromDecodedTile(...)` to convert parsed features into vector glTF JSON using the `CESIUM_mesh_vector` extension (and optional `EXT_mesh_features` when `featureIdProperty` is set)
- Returns `Empty3DTileContent` for tiles that decoded to nothing, otherwise calls `VectorGltf3DTileContent.fromGltf(...)` to reuse the existing vector rendering / feature path

**`VectorGltf3DTileContent`**
- Converts glTF primitives into `BufferPoint` / `BufferPolyline` / `BufferPolygon` collections
- Participates in Cesium's picking / styling / update lifecycle

So:

MVT bytes → provider-injected codec → decoded vector features → vector glTF → existing Cesium vector tile content path.

Runtime is decoupled from MVT specifics: the same `_runtimeContentCodec` hook will host MLT and any future tile decoders without changes to `Cesium3DTile`, `Multiple3DTileContent`, `preprocess3DTileContent`, or `Cesium3DTileContentFactory`.

## Testing plan
Test files:
[highway_roads.zip](https://github.com/user-attachments/files/26802744/highway_roads.zip)
[highway_roads_no_clipping.zip](https://github.com/user-attachments/files/26802747/highway_roads_no_clipping.zip)
[highway_roads_uncompressed.zip](https://github.com/user-attachments/files/26802756/highway_roads_uncompressed.zip)

Sandcastle code:
```

const mvt = await Cesium.MVTDataProvider.fromUrlTemplate(
  "http://localhost:8082/highway_roads_uncompressed/{z}/{x}/{y}.pbf",
  {
    minZoom: 0,
    maxZoom: 6,
    extent: Cesium.Rectangle.fromDegrees(
      -159.780010, 19.059930, -66.979990, 70.191920
    ),
  },
);

viewer.scene.primitives.add(mvt);
viewer.camera.flyTo({
  destination: Cesium.Rectangle.fromDegrees(
    -159.780010, 19.059930, -66.979990, 70.191920
  ),
});

```

## Author checklist

## AI acknowledgment

Github Copilot

Claude Opus 4.7 for code review, GPT 5.3 codex to debug some decode logic


donmccurdy:
> Still WIP! I will update when this is ready for review!

Thanks Daniel! Removing myself from the reviewer list until then. 👍🏻 

I would just note that MVT is not part of the 3D Tiles standard and will not generally have a tileset.json file. A common structure is to have a template URL define the tileset, for example:

```
https://tiles-a.basemaps.cartocdn.com/vectortiles/carto.streets/v1/{z}/{x}/{y}.mvt
```

I suspect that this will be easier to structure similar to I3SDataProvider.js, where the user doesn't directly use 3DTiles-specific APIs. Then the user would call MVT-related APIs, which will (internally) construct 3D Tiles types like `VectorGltf3DTileContent` from MVT input, rather than creating a new `MVT3DTileContent` class, and the 3D Tiles classes would remain mostly ignorant of the MVT format. But feel free to propose a different structure if you prefer!

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
```diff
@@ -0,0 +1,704 @@
...
+  /** @type {number[]} */
+  const polygonIndicesOffsets = [];
+  let polygonCount = 0;
+
+  let featureId = 0;
+
+  for (const layer of decoded.layers) {
+    const extent = layer.extent;
+    for (const feature of layer.features) {
+      const currentFeatureId = featureId++;
```
danielzhong, line 112: Not sure this is the right way to compute featureID

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
```diff
@@ -0,0 +1,704 @@
...
+const UNSIGNED_INT = 5125;
+const ARRAY_BUFFER = 34962;
+const ELEMENT_ARRAY_BUFFER = 34963;
+const POINTS = 0;
+const LINE_STRIP = 3;
+const TRIANGLES = 4;
+
+const DEFAULT_POINT_HEIGHT = 150;
+const DEFAULT_LINE_HEIGHT = 100;
+const DEFAULT_POLYGON_HEIGHT = 50;
```
danielzhong, line 20: They’re default altitude offsets (in meters) used when converting MVT geometry into Cesium world coordinates, but I’m not sure if these values are appropriate.

donmccurdy: I'd expected a default height of `0` for all features — are you seeing any particular reasons to do otherwise?

Eventually, though, we'll need some way to manage layer order. For the traditional render path this would use [gl.polygonOffset](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/polygonOffset), and for the GPU Lookup path it would (probably?) be a more explicit ordering of features in the lookup table. MVT rendering is typically based on layers, and their order within the tile is important.

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
```diff
@@ -0,0 +1,704 @@
...
+
+/** @import { TypedArray } from "../Core/globalTypes.js"; */
+
+const FLOAT = 5126;
+const UNSIGNED_INT = 5125;
+const ARRAY_BUFFER = 34962;
+const ELEMENT_ARRAY_BUFFER = 34963;
+const POINTS = 0;
+const LINE_STRIP = 3;
+const TRIANGLES = 4;
```
donmccurdy, line 16: It's safe to use ComponentDatatype, PrimitiveType, and/or WebGLConstants to access these constants — glTF constants have the same values.

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
donmccurdy: This is fine for now, or we could omit the featureID if that doesn't cause other issues. The MVT data might have properties we could expose and allow the user to select one as the feature ID later, but I don't think we'll ultimately want to generate FeatureIDs if they're not in the tile already.

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
```diff
@@ -0,0 +1,704 @@
...
+        }
+        continue;
+      }
+
+      if (feature.type === "LineString") {
+        const lines = /** @type {VectorTilePoint[][]} */ (feature.geometry);
+        for (const line of lines) {
+          if (line.length < 2) {
+            continue;
+          }
```
donmccurdy, line 137: Minor — I wouldn't silence invalid cases unless necessary. Probably we can remove this check, and if we run into a compelling reason to silently allow lines of length 0–1 we can add a comment explaining why.

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
```diff
@@ -0,0 +1,704 @@
...
+              positions2D,
+              holeOffsets.length > 0 ? holeOffsets : undefined,
+            );
+          } catch {
+            continue;
+          }
+
+          if (!defined(triangles) || triangles.length === 0) {
+            continue;
+          }
```
donmccurdy, line 218: See above — let's only handle errors we can explain.

(similar for other error cases below)

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
```diff
@@ -0,0 +1,704 @@
...
+  function addAccessor(typedArray, options) {
+    const bufferView = addBufferView(typedArray, options.target);
+    const componentCount =
+      options.type === "SCALAR"
+        ? 1
+        : options.type === "VEC2"
+          ? 2
+          : options.type === "VEC3"
+            ? 3
+            : 4;
```
donmccurdy, line 315: Minor — can use `MetadataType.getComponentCount` here.

packages/engine/Source/Scene/MVTDataProvider.js
```diff
@@ -0,0 +1,146 @@
...
+ * @alias MVTDataProvider
+ * @constructor
+ *
+ * @param {Resource|string} urlTemplate URL template containing {z}, {x}, and {y} placeholders.
+ * @param {object} [options] Provider options.
+ * @param {number} [options.minZoom=0] Minimum requested zoom level.
+ * @param {number} [options.maxZoom=14] Maximum requested zoom level.
+ * @param {Rectangle} [options.extent] Optional geographic extent in radians to limit tile requests.
+ */
+function MVTDataProvider(urlTemplate, options) {
```
donmccurdy, line 20: Nit: As this is new code let's use ES6 classes, it can be converted automatically with:

```bash
node ./scripts/lebab-batch.js packages/engine/Source/Scene/MVTDataProvider.js
```

packages/engine/Source/Scene/MVTVectorContent.js
```diff
@@ -0,0 +1,182 @@
...
+      console.warn(
+        "MVT gzip decompress failed in DecompressionStream; falling back to pako. This MVT tile must be readable as uncompressed data after decompression.",
+      );
+    }
+  }
+
+  const { inflate } = await import("pako");
+  const out = inflate(bytes);
+  return out.buffer.slice(out.byteOffset, out.byteOffset + out.byteLength);
+}
```
donmccurdy: I haven't looked into how other implementations handle this, but I suspect that any decompression should be handled by the browser, and — if not — that indicates an HTTP server configuration problem. It wouldn't be surprised if the local dev server for sandcastle doesn't know about gzip compression in MVT tiles. Note that "pako" is a sandcastle dependency, not a CesiumJS dependency.

tl;dr — I believe it's OK for us not to explicitly handle decompression.

packages/engine/Source/Scene/VectorTileRuntime.js
```diff
@@ -0,0 +1,371 @@
...
+
+      if (result.status === "ready" && defined(result.content)) {
+        this._tileContents.set(key, result.content);
+      }
+    })
+    .catch((error) => {
+      if (defined(error) && error.statusCode === 404) {
+        this._missingTiles.add(key);
+      }
+      // Keep runtime resilient; retry transient failures on later frames.
```
donmccurdy, line 124: Is the intention here that we should retry after an HTTP 404, but not after an HTTP 204?

packages/engine/Source/Scene/VectorTileRuntime.js
donmccurdy, line 1: I like decision to separate the tile management / runtime from the MVTDataProvider here! Is the idea that this could be reused for similar formats like MLT or PMTiles?

packages/engine/Source/Scene/VectorTileRuntime.js
```diff
@@ -0,0 +1,371 @@
...
+ *
+ * @private
+ * @alias VectorTileRuntime
+ * @constructor
+ *
+ * @param {object} options Runtime options.
+ * @param {TilingScheme} [options.tilingScheme] Tiling scheme for tile coordinates.
+ * @param {number} [options.minZoom=0] Minimum requested zoom level.
+ * @param {number} [options.maxZoom=14] Maximum requested zoom level.
+ * @param {number} [options.requestRadius=2] Request radius around the center tile.
```
donmccurdy, line 50: I don't think I understand the "requestRadius" option, do you mind explaining how this is used?

packages/engine/Source/Scene/VectorTileRuntime.js
```diff
@@ -0,0 +1,371 @@
...
+
+export default VectorTileRuntime;
+
+/**
+ * Creates a URL-template tile fetcher with normalized missing-tile semantics.
+ *
+ * @param {Resource|string} urlTemplate URL template containing {z}, {x}, and {y} placeholders.
+ * @returns {UrlTemplateTileFetcher}
+ */
+export function createUrlTemplateTileFetcher(urlTemplate) {
```
donmccurdy: When running this code from Sandcastle:

```javascript
const mvt = await Cesium.MVTDataProvider.fromUrlTemplate(
  "../../SampleData/vector/highway_roads_no_clipping/{z}/{x}/{y}.pbf",
  {
    minZoom: 0,
    maxZoom: 6,
    extent: Cesium.Rectangle.fromDegrees(
      -159.780010, 19.059930, -66.979990, 70.191920
    ),
  },
);
```

I'm seeing requests to:

`http://localhost:8081/Apps/SampleData/vector/highway_roads_no_clipping/SampleData/vector/highway_roads_no_clipping/0/0/0.pbf` followed by a 404 error, I think something in the relative path handling might not be quite right?

Worked around it for now using an absolute path, `/Apps/SampleData/...`.

donmccurdy (commented):
This is looking really good! Exciting to see MVT up and running.

If I understand correctly, the organization is:

- Scene
  - MVTDataProvider
    - VectorTileRuntime
      - MVTVectorContent
        - VectorGltf3DTileContent

One way this differs from `I3SDataProvider` is that there's no `Cesium3DTileset` in this hierarchy, the MVTVectorContent holds a "tileset adapter" used as a stub tileset for the `VectorGltf3DTileContent`. 

I think that may be challenging for picking and styling later on — do you think it's feasible to have a `Cesium3DTileset` as the parent of the `VectorGltf3DTileContent` objects? 

We may or may not need the `MVTVectorContent` intermediate object in that case. It may be confusing because it does not extend or implement `Cesium3DTileContent`.

But I haven't looked to closely yet at how this would work, please let me know if you see problems with this approach.

/cc @lilleyse

danielzhong:
Should I add [highway_roads_uncompressed.zip](https://github.com/user-attachments/files/26802756/highway_roads_uncompressed.zip) into the repo as example MVT file?

packages/engine/Source/Scene/buildVectorGltfFromDecodedTile.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13404/commits/2caf761a9b374a284aeb2d98d11b3845b97fc3b6

This is a temporary logic. User can select for example `featureIdProperty: "id"` for the feature ID. But current data doesn't support feature ID yet, if none are provided, feature IDs will not be generated.

packages/engine/Source/Scene/VectorTileRuntime.js
danielzhong: Added 204, thanks

packages/engine/Source/Scene/VectorTileRuntime.js
danielzhong: Yes. but I changed the structure. Now it follows the “real 3D Tiles lifecycle”:

* MVTDataProvider generates a runtime tileset.json.
* Cesium3DTileset handles tile management, LOD, and request lifecycle.
* MVT3DTileContent is only responsible for .pbf/.mvt -> decode -> vector glTF.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -183,6 +183,9 @@ function appendPrimitiveToBuffers(
     : undefined;
 
   const featureIdComponent = primitive.featureIds?.[0];
+  const nullFeatureId = defined(featureIdComponent)
+    ? featureIdComponent.nullFeatureId
+    : 0xffffffff;
```
donmccurdy: I'm not aware that any default 'null' value exists in EXT_mesh_features — it might be best to remove the `0xFFFFFFFF` fallback, or (if it's needed) add a comment explaining why important?

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -191,14 +194,30 @@ function appendPrimitiveToBuffers(
...
+
+  /**
+   * @param {number} vertexOffset
+   * @returns {Cesium3DTileVectorFeature|undefined}
+   */
   function getFeature(vertexOffset) {
-    const featureId = featureIdArray[vertexOffset];
-    if (featureId !== featureIdComponent.nullFeatureId) {
-      return features.get(featureId);
+    if (defined(featureIdArray) && defined(featureIdComponent)) {
```
donmccurdy: I think we should consider it an error (i.e. let the code fail, rather than explicitly checking) if this function is called while featureIdArray or featureIdComponent do not exist — mainly to support legibility of the code.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -231,8 +250,10 @@ function appendPrimitiveToBuffers(
       );
 
       const pickObject = getFeature(vertexOffset);
-      pickObject.addPrimitiveByCollection(collectionIndex, i);
-      const featureId = pickObject.featureId;
+      if (defined(pickObject)) {
+        pickObject.addPrimitiveByCollection(collectionIndex, i);
+      }
```
donmccurdy: Note that because every vertex must correspond to a feature, getFeature must return a feature, and we shouldn't need to handle cases where pickObject is undefined.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -277,21 +300,23 @@ function appendPrimitiveToBuffers(
         const holeCount = polygonHoleCounts[i];
         holes = polygonHoleOffsets.slice(i, i + holeCount);
         for (let h = 0; h < holeCount; h++) {
-          holes[i] -= polygonVertexStart;
+          holes[h] -= polygonVertexStart;
```
donmccurdy: Ah good catch! Just curious, do you have a dataset with holes that helped to find this?

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
donmccurdy, line 1: Minor — I'd be inclined to name this `buildVectorGltfFromMVT` or maybe inline it at the end of MVTDataProvider? While I know we may want to generalize it to some other formats later, it might be a while before we do, and "MVT" is easier to understand than "DecodedTile" for readers.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,732 @@
...
+const DEFAULT_HEIGHT = 0;
+
+const scratchWorld = new Cartesian3();
+const scratchLocal = new Cartesian3();
+
+/**
+ * @typedef {object} VectorTilePoint
+ * @property {number} x Tile-local x (0–extent)
+ * @property {number} y Tile-local y (0–extent)
+ */
```
donmccurdy, line 29: Note that any (internal) top-level JSDoc we introduce should have `@ignore` attached, to avoid publishing it with our type definitions. I added a new script to the repository recently, meant to make it easier to see if we've accidentally introduced new types that we didn't intend to:

https://github.com/CesiumGS/cesium/blob/main/scripts/diff-types.sh

packages/engine/Source/Scene/MVTDataProvider.js
```diff
@@ -0,0 +1,180 @@
...
+          resource.getUrlComponent(true),
+        );
+        const gltf = buildVectorGltfFromDecodedTile(
+          decodedTile,
+          tileCoordinates,
+          { featureIdProperty: featureIdProperty },
+        );
+        if (!defined(gltf)) {
+          return new Empty3DTileContent(tileset, tile);
+        }
```
donmccurdy: I think it's fair to require that if a decoded tile exists, a decoded glTF must also exist — I'd let this fail rather than silencing the error, or add a comment if there's some reason we expect this case.

packages/engine/Source/Scene/UrlTemplate3DTilesDataProvider.js
```diff
@@ -0,0 +1,644 @@
...
+}
+
+/**
+ * @param {Resource} resource
+ * @param {object} options
+ * @param {number} options.minZoom
+ * @param {number} options.maxZoom
+ * @param {Rectangle} [options.extent]
+ * @param {number} options.maxTilesetNodeCount
+ * @returns {object}
```
donmccurdy, line 276: Is a "node" different from a "tile"? If not — there's a maximum number of tiles implied by the min/max zoom and the extent, so I don't think we need to support a separate option for `maxTilesetNodeCount`.

packages/engine/Source/Scene/UrlTemplate3DTilesDataProvider.js
```diff
@@ -0,0 +1,644 @@
...
+  const tileUrl = template
+    .replace(/\{z\}/gi, `${level}`)
+    .replace(/\{x\}/gi, `${x}`)
+    .replace(/\{y\}/gi, `${y}`);
+  return getAbsoluteUri(tileUrl);
+}
+
+function computeGeometricError(level) {
+  return EARTH_CIRCUMFERENCE_METERS / ((1 << level) * WEB_MERCATOR_TILE_SIZE);
+}
```
donmccurdy, line 423: What's the rational for the math here?

Minor, no change needed — perhaps ideally, we'd load MVT tiles based on screen space coverage or a general target of N < 100 tiles on screen at any given time. I assume we don't exactly know the geometric error associated with an MVT tile. But maybe that doesn't fit well with 3D Tiles current loading logic, I haven't looked too closely at that yet, it's fine to keep using geometric error in the meantime.

donmccurdy (commented):
Exciting to see this working, great progress so far! A few initial comments inline.

> Should I add highway_roads_uncompressed.zip into the repo as example MVT file?

Probably better to avoid putting a tileset into the codebase unfortunately! We can look into putting something in ion later perhaps.

***

One higher-level comment — please try to avoid overly defensive programming. Patterns like this...

```javascript
if (defined(someInternalArray)) {
  return undefined;
}
```

... will, if used too much, have the effect of swallowing the root cause of an error and making debugging more difficult, and making the code more difficult to understand, test, and change. Ideally errors are handled as early as possible, so that all later code can assume strict and valid inputs. When we do need these checks (and if it's not obvious from the context why a test is needed) it's best to include a short comment explaining why we're checking something.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: [mvt_polygon_hole_regression_z6.zip](https://github.com/user-attachments/files/27150816/mvt_polygon_hole_regression_z6.zip)


```

const mvt = await Cesium.MVTDataProvider.fromUrlTemplate(
  "http://localhost:8082/mvt_polygon_hole_regression_z6/{z}/{x}/{y}.pbf",
  {
    minZoom: 6,
    maxZoom: 6,
    extent: Cesium.Rectangle.fromDegrees(-95.625, 36.597889, -90, 40.979898),
  },
);

viewer.scene.primitives.add(mvt);
mvt.tileset.style = new Cesium.Cesium3DTileStyle({
  color: "color('red', 0.55)",
});

viewer.camera.flyTo({
  destination: Cesium.Rectangle.fromDegrees(-95.625, 36.597889, -90, 40.979898),
});

```

packages/engine/Source/Scene/UrlTemplate3DTilesDataProvider.js
danielzhong: I'm using `geometricError = 2πR / (2^z × 256)`. It's meters-per-pixel at each zoom level: earthCircumference / (2^z * tileSize). I got this formula online. Please feel free to share any additional ideas or we can optimize better in the future if needed.

yieryi:
Are there any plans to support the style of mapbox? For example, style expressions like effect,condition, and filter

donmccurdy:
Not sure yet about supporting the Mapbox style specification / syntax directly.

As the approach in this PR is to "convert" MVT tiles to Cesium 3D Tiles at runtime, the runtime representation is the same as 3D Tiles, and [existing 3D Tiles styling features](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileStyle.html) will (when this is released) work as expected, likely with some additional features specific to vector data. That includes conditions and filters — I'm not sure I understand what you mean by effects, though.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -183,42 +183,50 @@ function appendPrimitiveToBuffers(
...
-
+  if (hasFeatureIdAttribute) {
     for (let i = 0; i < featureIdArray.length; i++) {
       const featureId = featureIdArray[i];
       if (!features.has(featureId)) {
-        const feature = new Cesium3DTileVectorFeature(
-          content,
-          featureId,
-          featureIdComponent.propertyTableId,
-        );
```
donmccurdy: Possible regression, maybe from a merge conflict, here? We'll need to pass `featureIdComponent.propertyTableId` into the feature constructor at least for 3D Tiles. For MVT I'm not sure if it's necessary or not.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -183,42 +183,50 @@ function appendPrimitiveToBuffers(
...
+   * @returns {Cesium3DTileVectorFeature}
+   */
+  function getOrCreateFeature(featureId) {
+    let feature = features.get(featureId);
+    if (!defined(feature)) {
+      feature = new Cesium3DTileVectorFeature(content, featureId);
+      features.set(featureId, feature);
+    }
+    return feature;
+  }
```
donmccurdy: Just to clarify the reason for the changes around Feature IDs — is the issue that with MVT we don't have a list of Feature IDs initially, so we can't pre-allocate all the features? The fix for polygon holes makes sense to me but I'm not sure I understand the Feature ID changes.

Assuming the Feature ID changes are needed — it might be confusing having `getOrCreateFeature` as a helper function for `getFeature`, as the distinction in naming suggests that one only returns existing features and the other [upserts](https://en.wiktionary.org/wiki/upsert). I think I would suggest folding the full code into `getFeature`.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+ *
+ * @ignore
+ */
+function buildVectorGltfFromMVT(decoded, tileCoordinates, options) {
+  const tileX = tileCoordinates.tileX;
+  const tileY = tileCoordinates.tileY;
+  const tileZ = tileCoordinates.tileZ;
+  const featureIdProperty = options?.featureIdProperty;
+  const shouldUseFeatureIds =
+    typeof featureIdProperty === "string" && featureIdProperty.length > 0;
```
donmccurdy: Nit: I'd be less generous with the input here, and just check `defined(featureIdProperty)`. Empty string or non-string input can be allowed to crash rather than silently being ignored.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+
+  const origin = computeTileOriginCartesian(tileX, tileY, tileZ);
+  const nullFeatureId = 0xffffffff;
+  const primitiveRestartIndex = 0xffffffff;
```
donmccurdy: Nit: These restart indices are unique to Uint32Array and I suspect that'd be very easy for a reader to miss, do you mind adding a comment or something like:

```javascript
const MAX_INT_U32 = 0xffffffff;

...


const nullFeatureId = MAX_INT_U32;
const primitiveRestartIndex = MAX_INT_U32;
```

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+        mappedFeatureId = mapFeatureIdFromProperty(
+          feature,
+          /** @type {string} */ (featureIdProperty),
+          /** @type {Map<string, number>} */ (featureIdLookup),
+        );
+      }
+      const currentFeatureId = defined(mappedFeatureId)
+        ? mappedFeatureId
+        : nullFeatureId;
+      hasAnyFeatureId = hasAnyFeatureId || defined(mappedFeatureId);
```
donmccurdy: I haven't put too much thought into this yet... but my hunch is that even if the user doesn't specify `featureIdProperty` we should still assign some sort of Feature ID to each feature, even if those are just ascending integers 0–N. Without a Feature ID we can't properly support ~~picking or~~ access to property metadata. And by requiring that some Feature ID exists, even if it's auto-generated, we might not need as many changes in `createVectorTileBuffers...`, simplifying the code maintenance.

Does that sound reasonable and/or make sense?

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+          const polygonPositionComponents = [];
+          /** @type {number[]} */
+          const holeOffsets = [];
+          let vertexOffset = 0;
+
+          for (let ringIndex = 0; ringIndex < rings.length; ringIndex++) {
+            const ring = rings[ringIndex];
+            if (ring.length < 3) {
+              continue;
+            }
```
donmccurdy: Runtime error is probably OK here! As a user I think I'd prefer that rather than silently hiding a corrupt feature, unless there's some well-known reason these corrupt data are common. Also it's easier for us to catch bugs in our parsing if errors propagate.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+          }
+
+          const triangles = PolygonPipeline.triangulate(
+            positions2D,
+            holeOffsets.length > 0 ? holeOffsets : undefined,
+          );
+
+          if (!defined(triangles) || triangles.length === 0) {
+            continue;
+          }
```
donmccurdy: An exception to my earlier suggestion — I do agree that we should skip polygons if triangulation fails, without crashing, as triangulation may not be 100% robust to all valid inputs. I would just log a warning as well.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+    const componentCount = /** @type {*} */ (MetadataType).getComponentCount(
+      options.type,
+    );
+    const accessor = /** @type {*} */ ({
+      bufferView: bufferView,
+      byteOffset: 0,
+      componentType: options.componentType,
+      count: typedArray.length / componentCount,
+      type: options.type,
+    });
```
donmccurdy, line 165: Note to self – we should import a type definitions file for the GLTF specification at some point, now that we have type checking available, it would be helpful for cases like this. Example:

https://github.com/donmccurdy/glTF-Transform/blob/main/packages/core/src/types/gltf.ts


But way out of scope for this PR, please ignore me. :)

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
donmccurdy, line 1: Just a thought, no changes needed — it would also be possible for `buildVectorGltfFromMVT` to produce a `VectorGltf3DTileContent` directly, e.g. constructing the BufferPrimitiveCollections and attaching them directly to the tile content, rather than going encoding to glTF / 3D Tiles first and then calling `createVectorTileBuffersFromModelComponents`.

I don't have any immediate guess as to whether that's "better" or "worse", though... It sounds simpler/easier to me, but it might not be. Whichever you prefer is OK with me!

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,782 @@
...
+  }
+
+  let binary = "";
+  const chunkSize = 0x8000;
+  for (let i = 0; i < bytes.length; i += chunkSize) {
+    const chunk = bytes.subarray(i, i + chunkSize);
+    binary += String.fromCharCode.apply(null, chunk);
+  }
+  return btoa(binary);
+}
```
donmccurdy: Nit — It would be ideal if we could avoid any base64 encoding  when constructing the glTF, and pass off the binary data directly. I'm not sure how/if that's possible on the loader APIs, though, so OK to keep this as-is if it's too much trouble.

packages/engine/Source/Scene/Cesium3DTile.js
```diff
@@ -1316,6 +1329,71 @@ function requestSingleContent(tile) {
...
+    Array.isArray(policy.statusCodes) && policy.statusCodes.length > 0
+      ? policy.statusCodes
+      : [404, 204];
+  if (!statusCodes.includes(statusCode)) {
+    return false;
+  }
+
+  const url = tile._contentResource?.url;
+  return defined(url) && urlPattern.test(url);
+}
```
donmccurdy: A bit too lenient with the input here I think — could we remove any "defensive" checks here that we don't yet have a need for?

packages/engine/Source/Scene/decodeMVT.js
```diff
@@ -0,0 +1,591 @@
...
+      return v;
+    } else if (fieldNumber === 4 && wireType === 0) {
+      const value = readBigVarint(bytes, pos, end);
+      return toSafeNumberOrString(value.value);
+    } else if (fieldNumber === 5 && wireType === 0) {
+      const value = readBigVarint(bytes, pos, end);
+      return toSafeNumberOrString(value.value);
+    } else if (fieldNumber === 6 && wireType === 0) {
+      const value = readBigVarint(bytes, pos, end);
+      return toSafeNumberOrString(zigzagBigInt(value.value));
```
donmccurdy: Could we create an enum for these constants, and link to the specification at the top of the file for future readers? 

I think it should be OK to use BigInt here rather than strings.

donmccurdy (commented):
A few more comments, but getting close to MVT support, nice work here!

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
danielzhong: It's possible. But the current approach feels better structure to me — glTF serves as a natural interface boundary for the codec, so any future format (MLT, GeoJSON tiles, etc.) just needs to implement `createContent()` and produce a glTF, reusing the full rendering path downstream.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
danielzhong: Good point! I wasn't familiar with the GLB format encode, GitHub Copilot helped with the refactor.

darcyvdd:
Capturing some interest on the forum [here](https://community.cesium.com/t/trying-to-drap-large-mvt-data-on-3d-terrain/46173)

packages/engine/Source/Scene/Cesium3DTile.js
donmccurdy: Hm, there's no requirement that MVT tiles must be organized into folders like `z/x/y.mvt`, something like `tile.mvt?x=1&y=2&z=3` would be valid as well.

It's also not clear to me why the "missing tile policy" would depend on the URL — can we safely remove the `urlPattern` check?

packages/engine/Source/Scene/Cesium3DTileset.js
```diff
@@ -761,6 +761,26 @@ function Cesium3DTileset(options) {
...
+   * the standard magic-number / URL based content dispatch and delegates
+   * content construction to <code>codec.createContent(...)</code>. This
+   * keeps format-specific logic out of the runtime.
+   *
+   * Shape:
+   *   {
+   *     contentType: string,                 // diagnostic only
+   *     disableSkipLevelOfDetail?: boolean,
+   *     createContent: (tileset, tile, resource, arrayBuffer) => Promise<Cesium3DTileContent>,
+   *     missingTilePolicy?: { statusCodes?: number[], urlPattern?: RegExp }
```
donmccurdy: It might be good to include a comment, here or on `isMissingTileContentError`, explaining what the "missing tile policy" is. Perhaps something like:

> A missing tile policy specifies HTTP Status Codes to be interpreted as "no content", and rendered as empty tiles, rather than throwing errors or retrying the request. Allows tiles to be statically hosted, without generating and serving unnecessary content for empty tiles.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
donmccurdy: I don't think this comment has been resolved quite yet. It's my understanding that there must be a feature for every primitive — even if that requires generating synthetic feature IDs for the MVT tiles. 

Aside - I'm somewhat hoping that we can merge the MVT PR without changes to this file (`createVectorTileBuffersFromModelComponents.js`) other than the two typo fixes ... if the changes to this file are still required, could you help me understand where/why? Thanks!

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
donmccurdy: Ah sorry, unresolving — what I mean by runtime error is to just let the code run and crash, rather than bundling special error messages for input data validation. It's much harder to remove special cases from the code in the future than it is to add them later when we discover we need them.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,836 @@
...
+  const polygonIndicesOffsets = [];
+  /** @type {number[]} */
+  const polygonHoleCounts = [];
+  /** @type {number[]} */
+  const polygonHoleOffsets = [];
+  let polygonCount = 0;
+
+  for (const layer of decoded.layers) {
+    const extent = layer.extent;
+    for (const feature of layer.features) {
```
donmccurdy: (Optional) In the future, I think we'll want to generate one glTF "node" per source "layer" in the MVT file, with the node being named after the layer. That will allow us to do layer-specific styling, which isn't yet supported in 3D Tiles Vector Tiles but is planned.

If it's easy enough to take that approach in this PR it might be a good thing to include now, but I'm also fine with waiting and considering it in another PR.

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
```diff
@@ -0,0 +1,836 @@
...
+ * @ignore
+ */
+function computeTileOriginCartesian(tileX, tileY, tileZ) {
+  const n = 1 << tileZ;
+  const u = (tileX + 0.5) / n;
+  const v = (tileY + 0.5) / n;
+  const lon = u * 2 * Math.PI - Math.PI;
+  const lat = Math.atan(Math.sinh(Math.PI * (1 - 2 * v)));
+  return Cartesian3.fromRadians(lon, lat, 0);
+}
```
donmccurdy: Would `WebMercatorTilingScheme#tileXYToRectangle` provide the same result as `computeTileOriginCartesian` here? If so I think we can use that; if not lets add a comment explaining what this does – That's the underlying tiling scheme for MVT, in any case.

packages/engine/Source/Scene/Cesium3DTile.js
```diff
@@ -1316,6 +1329,58 @@ function requestSingleContent(tile) {
   return processArrayBuffer(tile, tileset, request, expired, promise);
 }
 
+function isMissingTileContentError(tile, error) {
```
donmccurdy: Minor - The name of this function is a bit unclear about whether it's saying that the missing tile should be interpreted as an error, or it's saying something else about the error. Because the word "missing" often implies something is wrong, where "empty" or "no content" do not... perhaps "isNoContentTile" or "isEmptyTile"?

danielzhong:
@copilot+claude-opus-4.7 help me resolve the merge conflict from main

packages/engine/Source/Scene/buildVectorGltfFromMVT.js
danielzhong: https://github.com/CesiumGS/cesium/pull/13404/commits/e9b2a1f9f2297a9fb879f50ee5c4f9a04d91442a

packages/engine/Source/Scene/Cesium3DTile.js
```diff
@@ -1316,6 +1329,60 @@ function requestSingleContent(tile) {
...
+  const policy = tileset?._runtimeContentCodec?.missingTilePolicy;
+  if (!defined(policy)) {
+    return false;
+  }
+
+  const statusCode = getErrorStatusCode(error);
+  if (!Number.isFinite(statusCode)) {
+    return false;
+  }
+  return policy.statusCodes.includes(statusCode);
```
donmccurdy: On local testing, I couldn't find cases where `error.statusCode` was not defined and numeric. I think we must try to know what the inputs to these functions are – if we don't know how to reach the code, it's effectively dead code. Let's not include dead code in new files! :) 

```suggestion
  if (!defined(error.statusCode)) {
    return false;
  }
  return policy.statusCodes.includes(error.statusCode);
```

I think that the `getErrorStatusCode` function below can then be removed. If you're sure that we _must_ handle some of the special cases that it covers, could you include comments about how you encountered those special cases?

packages/engine/Source/Scene/Multiple3DTileContent.js
```diff
@@ -604,39 +604,51 @@ async function createInnerContent(multipleContents, arrayBuffer, index) {
...
-        contentFactory(
-          tileset,
-          tile,
-          resource,
-          preprocessed.binaryPayload.buffer,
-          0,
-        ),
+        codec.createContent(tileset, tile, resource, arrayBuffer),
       );
     } else {
```
donmccurdy, line 635: On local testing, I couldn't reach this code — are changes in `packages/engine/Source/Scene/Multiple3DTileContent.js` required for MVT support? If not, would it be OK to revert this file to its state from `main`?

packages/engine/Source/Scene/Cesium3DTile.js
```diff
@@ -1328,10 +1395,36 @@ function requestSingleContent(tile) {
...
+      const subtree = tile.implicitSubtree;
+      const coordinates = tile.implicitCoordinates;
+      content.metadata = subtree.getContentMetadataView(coordinates, 0);
+    } else if (!tile.hasImplicitContent) {
+      content.metadata = findContentMetadata(tileset, contentHeader);
+    }
+    const groupMetadata = findGroupMetadata(tileset, contentHeader);
+    if (defined(groupMetadata)) {
+      content.group = new Cesium3DContentGroup({ metadata: groupMetadata });
+    }
```
donmccurdy: On local testing, it didn't look like `content.metadata` or `content.group` are ever being assigned here. If they're not required for loading MVT data, I would recommend that we remove these if/else blocks, otherwise I'm not sure how we'd know if they work or not.


```suggestion
```

packages/engine/Source/Scene/Multiple3DTileContent.js
```diff
@@ -604,39 +604,51 @@ async function createInnerContent(multipleContents, arrayBuffer, index) {
...
-      preprocessed.contentType === Cesium3DTileContentType.VECTOR;
-
+    const codec = tileset?._runtimeContentCodec;
     let content;
-    const contentFactory = Cesium3DTileContentFactory[preprocessed.contentType];
-    if (defined(preprocessed.binaryPayload)) {
+    if (defined(codec) && typeof codec.createContent === "function") {
+      if (codec.disableSkipLevelOfDetail === true) {
+        multipleContents._disableSkipLevelOfDetail = true;
+      }
```
donmccurdy: Is `codec.disableSkipLevelOfDetail` necessary? My understanding is that it disables a tileset setting, but we aren't exposing the tileset itself to the user, so the MVTDataProvider is in full control of the settings passed to the tileset. I can't see a difference in the rendered result either, so I think that these `disableSkipLevelOfDetail` overrides can be removed from the PR.

donmccurdy (approved):
LGTM for initial MVT support, thank you @danielzhong! 

I have some more comments for future improvements, but will write those up in a separate issue soon.

donmccurdy:
Suggested next steps for MVT support:

- https://github.com/CesiumGS/cesium/issues/13534
- https://github.com/CesiumGS/cesium/issues/13535
- https://github.com/CesiumGS/cesium/issues/13536

PR #13405, opened 2026-04-14
Fix `Unhandled promise rejection: undefined thrown` in CI

### LINKED ISSUE #11958: `Unhandled promise rejection: undefined thrown` in CI

ggetz:
# Description

I'm really hoping this addresses the root of the sporadic failures we've been seeing in CI. 🤞 

With some brute-force help from copilot, I tracked down where `RequestScheduler` was rejecting with with `undefined`. 

1. Since `undefined` is not very helpful in isolation, I changed it to be a slightly more descriptive `RuntimeError` which should also help by providing a stack trace if or when it is unhandled.
2. Added a fallback `catch` block to cancelled requests to ensure any rejections are handled once the reference to the original promise is lost.

## Issue number and link

Fixes https://github.com/CesiumGS/cesium/issues/11958

## Testing plan

1. CI should pass
2. In `main`, the original issue can be reproduced with `npm run test -- --webgl-stub`. In this branch, multiple runs of the same command should consistently pass.

## Author checklist

## AI acknowledgment

_No content was generated by AI, but I used GitHub Copilot to trace down the source of the rejection._

### DIFF
--- a/packages/engine/Source/Core/Iau2006XysData.js
+++ b/packages/engine/Source/Core/Iau2006XysData.js
@@ -253,7 +253,8 @@ async function requestXysChunkJson(resource, index, xysData) {
     const chunk = await resource.fetchJson();
     xysData._updateChunkData(index, chunk);
   } catch (e) {
-    // TODO: Retry?
+    // Swallow the error. Proper error handling was not initially implemented
+    // for this class, and upstream systems all have fallbacks.
   }
 }
 
--- a/packages/engine/Source/Core/Iau2006XysData.js
+++ b/packages/engine/Source/Core/Iau2006XysData.js
@@ -142,7 +142,7 @@ Iau2006XysData.prototype.preload = function (
  *                 the Terrestrial Time (TT) time standard.
  * @param {Iau2006XysSample} [result] The instance to which to copy the interpolated result.  If this parameter
  *                           is undefined, a new instance is allocated and returned.
- * @returns {Iau2006XysSample} The interpolated XYS values, or undefined if the required data for this
+ * @returns {Iau2006XysSample|undefined} The interpolated XYS values, or undefined if the required data for this
  *                             computation has not yet been downloaded.
  *
  * @see Iau2006XysData#preload
@@ -237,8 +237,28 @@ Iau2006XysData.prototype.computeXysRadians = function (
   return result;
 };
 
+Iau2006XysData.prototype._updateChunkData = function (index, { samples }) {
+  this._chunkDownloadsInProgress[index] = undefined;
+
+  const samplesPerFile = this._samplesPerXysFile;
+  const startIndex = index * samplesPerFile * 3;
+
+  for (let i = 0; i < samples.length; ++i) {
+    this._samples[startIndex + i] = samples[i];
+  }
+};
+
+async function requestXysChunkJson(resource, index, xysData) {
+  try {
+    const chunk = await resource.fetchJson();
+    xysData._updateChunkData(index, chunk);
+  } catch (e) {
+    // TODO: Retry?
+  }
+}
+
 function requestXysChunk(xysData, chunkIndex) {
-  if (xysData._chunkDownloadsInProgress[chunkIndex]) {
+  if (defined(xysData._chunkDownloadsInProgress[chunkIndex])) {
     // Chunk has already been requested.
     return xysData._chunkDownloadsInProgress[chunkIndex];
   }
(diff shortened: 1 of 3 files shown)


javagl:
I may have to update the inlined comment that I [mentioned a while ago](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3497369871). It currently says

> Also note that the returned promise may be rejected and receive 'undefined' as the error, so it's impossible to know WHY it was rejected.

It now seems to receive an error that says 'Request cancelled'. (It could say _which_ request that was, i.e. include the URL). But I think that this form of "cancelling" can also happen due to "throttling", and I'm not sure whether it's possible to differentiate that further.

ggetz:
It looks like this isn't quite a fix yet, but it does seem like we're on the right track.

```
1) verify statistics
     Scene/Cesium3DTileset
     Unhandled promise rejection: RuntimeError: Request cancelled
Error
    at new V1 (Build/Cesium/Cesium.js:75:1293)
    at mT (Build/Cesium/Cesium.js:80:5040)
    at br.clearForSpecs (Build/Cesium/Cesium.js:80:7150)
...
```

> (It could say which request that was, i.e. include the URL)

I think at least for debugging's sake, I'm going to try this. I'm not sure yet if it would make sense for production.

javagl:
Now it _does_ provide a hint...

> Unhandled promise rejection: RuntimeError: Request cancelled: "Build/Cesium/Assets/IAU2006_XYS/IAU2006_XYS_17.json"

I never stumbled over `Iau2006XysData` before. Quickly browsing over it, it looks like one of these ... 'adventurous'... areas of the code.... It looks like it is trying to load and parse (up to) 1.7 MB of JSON data from 28 files. There isn't a specific, obvious place where these loads are triggered. And there isn't an obvious possibility to ensure that the data was actually loaded (before everything is torn down and the requests are cancelled - and the fact that a cancelled request should not cause a test failure at a completely unrelated place is ... unrelated).
A wild guess: Maybe some
`await Cesium.Transforms.preloadIcrfFixed(eternity));`
in a `beforeEach` could do it?

javagl:
✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ ✔️ !?! Maybe just a glitch 😁 

Could this confirm anything here? 
```
...
} catch (e) {
    // TODO: Retry?
   console.log("It was really that one, and nothing else!", e):
}
```

jjspace:
I've been poking CI to re-run periodically while working on other stuff and it's now up to successfully passing over _10 times in a row_ :tada:. I think it's safe to say these changes do prevent the failures we were getting before (or just hide them even better). I haven't gone through the code fully but on the surface this PR seems good?

javagl:
It's hard to imagine negative side-effects of the change here. The TODO should be replaced with a proper 'Warning' (maybe `oneTimeWarning`?), but that should be it.

When trying to be _really_ thorough, one could review more of that. But the details of the affected class are... hard to grasp. It looks like it is sending out requests ...~"occasionally", based on things being `undefined`. I think the worst case would be that someone says
[`await Cesium.Transforms.preloadIcrfFixed(interval));`](https://github.com/CesiumGS/cesium/blob/0da68a8f1055f534f43d7d74a7050e193246fbc7/packages/engine/Source/Core/Transforms.js#L687)
and expects everything to be "preloaded" then, but it isn't, due to a cancelled requests (with "cancelled" possibly meaning "throttled", I think).

There is a strange connection to another PR: When I saw that commit at https://github.com/CesiumGS/cesium/pull/13188/changes/0648d54a170282995e77945ab3d3552f9fd3f69e#diff-209e88e59aaffdf671fd40687b45f17782d2c8c0aec9819ecfdecce2919ab7ebR257 , I thought "... (what the commit message says 😁 ) ...".

packages/engine/Source/Core/Iau2006XysData.js
```diff
@@ -237,8 +237,28 @@ Iau2006XysData.prototype.computeXysRadians = function (
...
+    this._samples[startIndex + i] = samples[i];
+  }
+};
+
+async function requestXysChunkJson(resource, index, xysData) {
+  try {
+    const chunk = await resource.fetchJson();
+    xysData._updateChunkData(index, chunk);
+  } catch (e) {
+    // TODO: Retry?
```
jjspace, line 256: Are there plans to address this in this PR? or is that a potential future change? It doesn't seem like the previous version had a retry so I don't think that's a change in behavior

jjspace (commented):
Ok, actually looked at the code this time and I think it all looks good. Just one question

packages/engine/Source/Core/Iau2006XysData.js
ggetz: I was investigating if it made sense to set the chunk data promise back to undefined so it could be retried next time. However, that seems to have nefarious effects which break multiple tests. 

I'll remove the TODO and add a comment to the empty catch block. That should be consistent with the existing behavior at least.

ggetz:
@jjspace Updated!

jjspace (approved):
I'm fine with this for now. CI is clearly happy. Maybe some more adjustments come in the future but I'm ok to merge and start getting CI passing in PRs again!

PR #13418, opened 2026-04-20
Add custom shader for Gaussian Splats

dbiguenet:
# Description

Currently, we cannot display a 3DT GS with a custom shader to modify the color of a splat : doing so does not have any effect, at best. 
As we are working on change detection between gaussian splats and point clouds, we need to provide a custom shader to 3DT GS and display their classification. The classification is stored in the gltf/glc (featureid_0) and is obtained through a Reality Analysis change detection job https://qa-developer.bentley.com/apis/reality-analysis/operations/jobs-create/
- 0 = no change, we display the original color
- 1 = change, we blend the original color with an arbitrary color (blue).

Expected result : 
Screenshot 2026-04-21 170512

The goal of this PR is to make 3DT GS support custom shaders, so we can display changes. It will also allow other Cesium users to display 3DT GS with custom shader for other purposes.

## Issue number and link

https://github.com/iTwin/platform-engineering/issues/370?reload=1?reload=1?reload=1%3Freload%3D1?reload=1%3Freload%3D1%3Freload%3D1%3Freload%3D1?reload=1%3Freload%3D1%3Freload%3D1%3Freload%3D1%3Freload%3D1%3Freload%3D1%3Freload%3D1%3Freload%3D1

## Testing plan

I'm not an expert of CesiumJS, however I think we should test at least those cases : 

- Display classified 3DT GS with custom shader (I will make a sandbox example)
- Display classified 3DT GS without custom shader (should display the 3DT with the original colors)
- Display unclassified  3DT GS with custom shader (trying with another custom shader. For instance, color the splats based on elevation?)
- Display unclassified 3DT GS without custom shader (should display the 3DT as usual)

If you have any other suggestion, do not hesitate to edit this section. 

## Author checklist

## AI acknowledgment


ggetz:
Thanks for the PR @dbiguenet! As we discussed offline, please update the PR description with the relevant info. When you're ready for review, please tag @danielzhong to take a first-pass review.

dbiguenet:
> Thanks for the PR @dbiguenet! As we discussed offline, please update the PR description with the relevant info. When you're ready for review, please tag @danielzhong to take a first-pass review.

Thanks @ggetz .
I have updated the description and the testing plan. However, it seems I don't have the permissions to add a reviewer.

dbiguenet:
Hi @ggetz @danielzhong the deployment checks still don't run, can you help me?

javagl:
@dbiguenet  I had asked about that internally. Apparently, the deployment checks are _not_ run for external forks (only for forks within the `cesium` repo itself). I don't think that this has to hold up any progress here (because linting and specs are running and passing), but I think that the main point that is missing here is an actual _review_.

cnovel:
Denis had to take some time off for personal reasons, I'll try to advance this in the meantime. I'll focus on providing sandbox examples. Can you point to me where I should add the example?

danielzhong:
@dbiguenet @cnovel I don’t think you’ll be able to get a deploy link for this, so no need to spend time on that, I’ll test it locally instead.

Could you please provide instructions on how to test it locally, and also include unit tests and an update to CHANGES.md?

Once that’s ready, feel free to @ me and I’ll take a look.

javagl:
I think it should be enough to add another subdirectory in `packages/sandcastle/gallery`, similar to https://github.com/CesiumGS/cesium/tree/main/packages/sandcastle/gallery/3d-tiles-gaussian-splatting 

One question is about the underlying data - namely, whether there is a gaussian splat data set that contains the classification information, and that is available on ion. 


EDIT Slight race condition here 🙂 My response was based on the assumption that there should be a dedicated sandcastle that demonstrates this feature in the future. It it's only about local testing as part of the review, then a "manual" test could be fine.

dbiguenet:
Hi @danielzhong 
Thanks for your help and your feedback. Cyril added the unit tests. He also created a sandcastle, the data can be found in the issue https://github.com/iTwin/platform-engineering/issues/370#issuecomment-4379530296. Do you need something else to test it locally?
About the CHANGES.md file, I'm not sure what to do : it seems we have to provide a version number.

danielzhong:
image
image


Great progress, I have verified the custom shader is working as expected! Please add an entry to `CHANGES.md` under the `## 1.142 section`, under `### @cesium/engine → #### Additions :tada:` (or create one). This entry should be user-facing, describing what users can now do

packages/engine/Source/Shaders/PrimitiveGaussianSplatFS.glsl
```diff
@@ -13,6 +14,39 @@ void main() {
...
+    float B = exp(A * 4.) * v_splatColor.a;
+    vec4 baseColor = vec4(v_splatColor.rgb * B, B);
+
+#if defined(HAS_CUSTOM_FRAGMENT_SHADER)
+    FragmentInput fsInput;
+    vec3 positionEC = czm_windowToEyeCoordinates(gl_FragCoord).xyz;
+    fsInput.attributes.positionEC = positionEC;
+    fsInput.attributes.positionWC = (czm_inverseView * vec4(positionEC, 1.0)).xyz;
+    fsInput.attributes.normalEC = vec3(0.0, 0.0, 1.0);
+    fsInput.attributes.color_0 = baseColor;
```
danielzhong: Minor mit (base on you preference):
color_0 is being set to baseColor, which is already pre-multiplied by `alpha (rgb * B)`. This means if a user writes `material.diffuse = fsInput.attributes.color_0.rgb` in their `fragmentMain`, they'll get a darker/washed-out result because the value hasn't been un-pre-multiplied yet.

`material.diffuse` is correctly un-pre-multiplied before `fragmentMain` is called, but color_0 is not, so the two are inconsistent. This also differs from how color_0 works in Model custom shaders, where it holds a linear color.

Maybe:
```
fsInput.attributes.color_0 = vec4(
    baseColor.a > 0.0 ? baseColor.rgb / baseColor.a : vec3(0.0),
    baseColor.a
);
```

danielzhong (commented):
It looks good to me. @ggetz Anyone suggesting doing a second round of review to check whether the code style is good enough?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1119,8 +1187,69 @@ Object.defineProperties(GaussianSplatPrimitive.prototype, {
...
+    },
+    set: function (customShader) {
+      if (this._customShader !== customShader) {
+        // Delete old custom shader entries from the uniform map
+        const uniformMap = this._uniformMap;
+        const oldCustomShader = this._customShader;
+        const oldCustomShaderUniformMap = oldCustomShader.uniformMap;
+        for (const uniformName in oldCustomShaderUniformMap) {
+          if (oldCustomShaderUniformMap.hasOwnProperty(uniformName)) {
+            // If the custom shader was set but the voxel shader was never
```
javagl: ... the <strike>voxel</strike> splat shader ...

( https://github.com/CesiumGS/cesium/blob/24c112988a734059a2ebc47786aa44d47d1c1b19/packages/engine/Source/Scene/VoxelPrimitive.js#L1044 )

packages/engine/Source/Scene/GaussianSplatRenderResources.js
```diff
@@ -3,72 +3,93 @@ import RenderState from "../Renderer/RenderState.js";
...
     RenderState.fromCache({
       depthTest: {
         enabled: true,
         func: DepthFunction.LESS_OR_EQUAL,
       },
+      cull: {
+        enabled: false,
+      },
+      depthMask: true,
+      blending: {
```
javagl, line 109: I don't know exactly what this `blending` definition is or accomplished (so this is not a _technical_ scrutiny - maybe it's necessary, maybe not). 

But in any case, the values of the properties should be the corresponding constants from `WebGLConstants`.

dbiguenet:
Conflict fixed
@danielzhong Changes file updated and I fixed basecolor
@javagl I fixed the comment and the blending values

packages/engine/Source/Shaders/PrimitiveGaussianSplatFS.glsl
dbiguenet: The fix looks good to me.
I applied the changes.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatRenderResources.js
dbiguenet: The shader defines the splat color. However, the splats being transparent, "blending" defines how to blend the splat color with the background color.
That being said, you are right about the hex values. Fixed.

javagl:
I am in many ways not deeply familiar with the splat rendering and custom shaders (and therefore, the combination of them).  So the options for the "review" here range from 1. scrolling over it, responding with 'LGTM', and hitting 'Approve', to 2. trying to do a proper, thorough _review_, based on the expectations, requirements, and a variety of test data sets.

I'll try to find a middle ground, but still have to allocate a bit of time for that. 

Until then: There is that [`Cesium3DTileset#featureIdLabel`](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#featureIdLabel). Roughly: When the data has _multiple_ feature IDs, then this label should select the one that is currently "active". It looks like the `GaussianSplat3DTileContent` has the `_featureIds` that is documented to be "...the _FEATURE_ID_0 vertex attribute". And it looks like this is always selected to be the "first" feature ID attribute. I could imagine that the `featureIdLabel` should be taken into account here. (Note: This `featureIdLabel` does not make sense in many ways, but given the _current_ role of this property, this is likely the behavior that users would expect).

danielzhong:
> I am in many ways not deeply familiar with the splat rendering and custom shaders (and therefore, the combination of them). So the options for the "review" here range from 1. scrolling over it, responding with 'LGTM', and hitting 'Approve', to 2. trying to do a proper, thorough _review_, based on the expectations, requirements, and a variety of test data sets.
> 
> I'll try to find a middle ground, but still have to allocate a bit of time for that.
> 
> Until then: There is that [`Cesium3DTileset#featureIdLabel`](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#featureIdLabel). Roughly: When the data has _multiple_ feature IDs, then this label should select the one that is currently "active". It looks like the `GaussianSplat3DTileContent` has the `_featureIds` that is documented to be "...the _FEATURE_ID_0 vertex attribute". And it looks like this is always selected to be the "first" feature ID attribute. I could imagine that the `featureIdLabel` should be taken into account here. (Note: This `featureIdLabel` does not make sense in many ways, but given the _current_ role of this property, this is likely the behavior that users would expect).

That's a great catch about feature ID. Please also fix the lint error

cnovel:
I ran into this issue yesterday, trying to shade feature_id_1 on a new dataset with labels+instance ids. I'll be looking into it today with @dbiguenet

dbiguenet:
@danielzhong @javagl 
I fixed the problem about feature id, and tested the fix with Cyril's new dataset :
[CD.zip](https://github.com/user-attachments/files/27635247/CD.zip)
You can test it with the sandcastle, just replace featureid_0 with featureid_1 in "classificationShader". You should see red, orange, yellow and green changes instead of just red.
However, I'm pretty sure my fix only works for featureid_0, 1, 2, and 3. What do you think?
@danielzhong 
I think the lint error does not come from my changes

packages/engine/Source/Scene/Vector3DTileContent.js
```diff
@@ -616,7 +687,6 @@ function initialize(content, arrayBuffer, byteOffset) {
       byteOffset,
       numberOfPoints * 3,
     );
-    byteOffset += pointsPositionByteLength;
     content._points = new Vector3DTilePoints({
       positions: pointPositions,
       batchIds: batchIds.points,
```
danielzhong, line 692: I think this is unrelated to the PR, and I think this might be the reason why the lint check failed?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -816,6 +851,57 @@ function createGaussianSplatTexture(context, splatTextureData) {
   });
 }
 
+/**
+ * Creates a GPU texture that stores per-splat feature IDs for all feature ID
+ * sets.  The texture uses RGBA unsigned-integer format; each texel holds up to
+ * 4 feature IDs (one per channel).  If more than 4 sets exist, additional
+ * texels per splat are used.
```
danielzhong, line 858: Should we create a warning for this?

javagl:
OK, I mentioned the `featureIdLabel` in a previous comment. And fortunately, one major aspect of resolving this was about the support of _multiple_ feature IDs. And this seems to be working now. I created a small test data set (maybe something for specs, or the 3D Tiles Samples repo, or ... just for testing this PR fow now): 

[Cesium PR 13418 Splats with EXT_mesh_features.zip](https://github.com/user-attachments/files/27655177/Cesium.PR.13418.Splats.with.EXT_mesh_features.zip)

It consists of four of my beloved unit cubes, each with a different feature ID, and with 4 sets of feature IDs. A sandcastle for testing this is included in the ZIP:

Cesium Splat Feature IDs

So the handling of multiple feature IDs seems to be working. 

---

Beyond that... I probably have to refresh my memory about the _role_ that the `featureIdLabel` _actually_ plays. In custom shaders, it is always possible to access the feature IDs _directly_. I think that the `featureIdLabel` is only relevant in the context of picking and styling, when the `EXT_mesh_features` extension is combined with the `EXT_structural_metadata` extension (and it might in fact be some sort of "workaround" for that case...). All this is hinted at by the "Footnote 2" of the https://github.com/CesiumGS/cesium/issues/10480 

Given that picking and styling are not (yet) a thing for splats, the `featureIdLabel` might not actually be relevant for this PR (yet...). So apologies if mentioning that `featureIdLabel` caused unnecessary confusion here.

For now, and specific for this PR: The `featureIdLabel` currently seems to set the `GaussianSplat3DTileContent` `featureIdInfo`. But this property does not seem to be used anywhere. So I _think_ that the handling of the `featureIdLabel` and that `featureIdInfo` can be removed.

(I think that _some_ handling of the `featureIdLabel` might be necessary for picking+styling. But how exactly it will have to be taken into account later ... may have to be sorted out when picking+styling are supposed to be implemented) 
 
(I have not yet done a line-by-line review of the changes, but have focussed on the tests for now. A further review pass may be added later)

---

Once more the note: The data in the "CD.zip" is _invalid_ according to the glTF validator, because of the `normalized` flag and the `UNSIGNED_INT` type used in a vertex attribute. This should be fixed before this is made part of a public demo.

packages/engine/Source/Scene/Vector3DTileContent.js
dbiguenet: You are right, it's not related to the PR : I could not commit my changes because byteoffset is set but not used, so I removed this line. Maybe this file has been modified in the main branch?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: @javagl  What do you think?

dbiguenet:
> OK, I mentioned the `featureIdLabel` in a previous comment. And fortunately, one major aspect of resolving this was about the support of _multiple_ feature IDs. And this seems to be working now. I created a small test data set (maybe something for specs, or the 3D Tiles Samples repo, or ... just for testing this PR fow now):
> 
> [Cesium PR 13418 Splats with EXT_mesh_features.zip](https://github.com/user-attachments/files/27655177/Cesium.PR.13418.Splats.with.EXT_mesh_features.zip)
> 
> It consists of four of my beloved unit cubes, each with a different feature ID, and with 4 sets of feature IDs. A sandcastle for testing this is included in the ZIP:
> 
> Cesium Splat Feature IDs
> So the handling of multiple feature IDs seems to be working.
> 
> Beyond that... I probably have to refresh my memory about the _role_ that the `featureIdLabel` _actually_ plays. In custom shaders, it is always possible to access the feature IDs _directly_. I think that the `featureIdLabel` is only relevant in the context of picking and styling, when the `EXT_mesh_features` extension is combined with the `EXT_structural_metadata` extension (and it might in fact be some sort of "workaround" for that case...). All this is hinted at by the "Footnote 2" of the #10480
> 
> Given that picking and styling are not (yet) a thing for splats, the `featureIdLabel` might not actually be relevant for this PR (yet...). So apologies if mentioning that `featureIdLabel` caused unnecessary confusion here.
> 
> For now, and specific for this PR: The `featureIdLabel` currently seems to set the `GaussianSplat3DTileContent` `featureIdInfo`. But this property does not seem to be used anywhere. So I _think_ that the handling of the `featureIdLabel` and that `featureIdInfo` can be removed.
> 
> (I think that _some_ handling of the `featureIdLabel` might be necessary for picking+styling. But how exactly it will have to be taken into account later ... may have to be sorted out when picking+styling are supposed to be implemented)
> 
> (I have not yet done a line-by-line review of the changes, but have focussed on the tests for now. A further review pass may be added later)
> 
> Once more the note: The data in the "CD.zip" is _invalid_ according to the glTF validator, because of the `normalized` flag and the `UNSIGNED_INT` type used in a vertex attribute. This should be fixed before this is made part of a public demo.

Thanks for the review.
How to validate the tileset? glb/gltf validator from Kronos does not support EXT_mesh_features and other extensions

packages/engine/Source/Scene/GaussianSplatPrimitive.js
javagl: I had seen this inlined comment, but (still) haven't read through the _details_ of the code, and the way how the case of more than 4 feature IDs is handled here.

Based on the description, I _thought_ that having more than 4 feature IDs should work, and would not cause any form of "functional limitations". The comment sounded like it should (only) have a slightly higher memory usage (and ... well, ... more data needs more memory ... 🙂 ). 

But now, I extended the test data that I already attached below, and it looks like it always returns `0` for the feature IDs >= 4:

[Cesium PR 13418 Splats with EXT_mesh_features - MANY.zip](https://github.com/user-attachments/files/27709229/Cesium.PR.13418.Splats.with.EXT_mesh_features.-.MANY.zip)


(Again, this is just from a quick test - I have not yet checked in the code what might explain this).

---

A bit more generally: 

_Usually_, I'd assume that more than 4 feature IDs would be rare. (In fact, more than 1 feature ID might already be rare). **But...:** It might be that in the context of splats and change detection, people will throw in more feature IDs more casually. 

There are some degrees of freedom for how such data could be modeled:

- There could be 1 feature ID attribute, containing IDs [1, 50]
- There could be 10 feature ID attributes, each containing IDs [1, 5] 

In terms of their "expressiveness", they could be equivalent. But the latter would need _much_ more memory. (Each attribute has 4 bytes per element after all...)

The `EXT_mesh_features` specification is not explicit about a "maximum number of attributes". But in practice, the number of vertex attributes is limitied (usually 16 on modern systems). Ironically, given that the attributes are stored into a texture here, the usual WebGL-based limits for the attributes would not even apply to them! But still, there _is_ a limit (even if it's implicit, via the maximum texture size).

So I think that a warning (probably some `oneTimeWarning`) could be warranted when the number is "very large", and more than 4 might already qualify as "very large" here. 

(I don't know if people could model splats and feature IDs with standard vertex attributes, but if they did, then the `GL_MAX_VERTEX_ATTRIBS` limit could kick in pretty quickly...)

javagl:
> How to validate the tileset? glb/gltf validator from Kronos does not support EXT_mesh_features and other extensions

For the unsupported extensions, it will create an 'INFO' message. And unfortunately, right now, it even reports 'ERROR' messages for the `KHR_gaussian_splatting: ...` attribute names. But these can be ignored for now. 

Dragging-and-dropping the `CD/BClassified/0/0.glb` into https://github.khronos.org/glTF-Validator/ will report 23 issues. Looking only at the errors (with 'severity: 0'), and ignoring the `MESH_PRIMITIVE_INVALID_ATTRIBUTE` ones about the invalid attribute names, the remaning ones are

```
            {
                "code": "ACCESSOR_NORMALIZED_INVALID",
                "message": "Only (u)byte and (u)short accessors can be normalized.",
                "severity": 0,
                "pointer": "/accessors/4/normalized"
            },
            {
                "code": "ACCESSOR_TOTAL_OFFSET_ALIGNMENT",
                "message": "Accessor's total byteOffset 18167905 isn't a multiple of componentType length 2.",
                "severity": 0,
                "pointer": "/accessors/5"
            },
            {
                "code": "MESH_PRIMITIVE_ACCESSOR_UNALIGNED",
                "message": "Vertex attribute data must be aligned to 4-byte boundaries.",
                "severity": 0,
                "pointer": "/meshes/0/primitives/0/attributes/_FEATURE_ID_0"
            },
            {
                "code": "MESH_PRIMITIVE_ATTRIBUTES_ACCESSOR_UNSIGNED_INT",
                "message": "Mesh attributes cannot use UNSIGNED_INT component type.",
                "severity": 0,
                "pointer": "/meshes/0/primitives/0/attributes/_FEATURE_ID_1"
            },
````

**Comment by cnovel** [comment]
What's our status for this PR? Do you need data for testing?

**Comment by cnovel** [comment]
Checking in again, do you need us to fix some issues on this PR?

**Comment by javagl** [comment]
I had tested it with the data attached to a [previous comment](https://github.com/CesiumGS/cesium/pull/13418#issuecomment-4434268551), and it seemed to work. 

I think that Daniel did a review on the code-level (and probably tested it, e.g. with the added sandcastle).

---

The only "issues" that are open: 

- Some preliminary test data (CD.zip) contained errors, but I haven't checked the data from the new sandcastle, so that might already be resolved
- The branch is outdated, and `main` has to be merged into this branch
- There is a CI failure, related to the 'seatbelt' file, **but** this **should** be resolved with the latest state from `main` (and if it isn't, then it is not an issue of this PR)

---

Apart from that, from my perspective, there's nothing preventing this from being merged.

**Comment by dbiguenet** [comment]
> I had tested it with the data attached to a [previous comment](https://github.com/CesiumGS/cesium/pull/13418#issuecomment-4434268551), and it seemed to work.
> 
> I think that Daniel did a review on the code-level (and probably tested it, e.g. with the added sandcastle).
> 
> The only "issues" that are open:
> 
> * Some preliminary test data (CD.zip) contained errors, but I haven't checked the data from the new sandcastle, so that might already be resolved
> * The branch is outdated, and `main` has to be merged into this branch
> * There is a CI failure, related to the 'seatbelt' file, **but** this **should** be resolved with the latest state from `main` (and if it isn't, then it is not an issue of this PR)
> 
> Apart from that, from my perspective, there's nothing preventing this from being merged.

Hi @javagl , thanks for the review
I have merged main in the feature branch
About the test data, I'm not sure, @cnovel might know

**Comment by dbiguenet** [comment]
Hi @weegeekps @javagl 
It seems that even our unclassified glb are not valid according to the gltf validator 
[unclassified.zip](https://github.com/user-attachments/files/30779457/unclassified.zip)
And this is the updated version of classified data
[classified.zip](https://github.com/user-attachments/files/30779483/classified.zip)
Do you want us to fix all the errors in the glb or only the "classified specific" ones?

@danielzhong do you know how to update/get the data used in the sandcastle? Cyril did it, and I don't remember how.

**Comment by javagl** [comment]
I do have the _strong_ preference to not produce or distribute invalid glTF files. But whether or not Cesium follows this rule is not my call. So I'd leave that to Adam or Gabby.

**Comment by weegeekps** [comment]
> Do you want us to fix all the errors in the glb or only the "classified specific" ones?

As a rule of thumb, we should be producing glTF files that are valid. There are some exceptions, such as ignoring `MESH_PRIMITIVE_INVALID_ATTRIBUTE`.

These errors need to be fixed as they have the potential to cause interoperability issues:

```json
{
    "code": "ACCESSOR_NORMALIZED_INVALID",
    "message": "Only (u)byte and (u)short accessors can be normalized.",
    "severity": 0,
    "pointer": "/accessors/4/normalized"
},
{
    "code": "ACCESSOR_TOTAL_OFFSET_ALIGNMENT",
    "message": "Accessor's total byteOffset 737833 isn't a multiple of componentType length 2.",
    "severity": 0,
    "pointer": "/accessors/5"
},
{
    "code": "MESH_PRIMITIVE_ATTRIBUTES_ACCESSOR_UNSIGNED_INT",
    "message": "Mesh attributes cannot use UNSIGNED_INT component type.",
    "severity": 0,
    "pointer": "/meshes/0/primitives/0/attributes/_FEATURE_ID_1"
}
```

Regarding the `MESH_PRIMITIVE_ATTRIBUTES_ACCESSOR_UNSIGNED_INT` error in particular, we have an implementation note [in the EXT_mesh_features extension specification](https://github.com/CesiumGS/glTF/blob/7ad354c6e7f2f73f34ec1b86ccbfbe5f693f3324/extensions/2.0/Vendor/EXT_mesh_features/README.md#feature-id-by-vertex) that points out what you should do as an alternative to using `UNSIGNED_INT` here.

The rest of the errors and warnings are due to the glTF Validator not supporting the 3DGS and metadata extensions.

RenaudKeriven:
Thanks @weegeekps. This is clear.

cnovel:
Updated data is [available](https://bentley-my.sharepoint.com/:u:/p/cyril_novel/IQDF17vvc0pDRLRexnk5N133AekphJLMaQMkaP00s25eEvA?e=D9oIPn)

dbiguenet:
Conflicts resolved

dbiguenet:
@danielzhong Sandcastle dataset updated

CHANGES.md
```diff
@@ -31,7 +32,6 @@
 - `ClippingPolygons` now use an algorithm, based on the techniques used for vector tiles, that vastly improves quality across distance scales. Warm-up cost is also modestly decreased. [#13654](https://github.com/CesiumGS/cesium/pull/13654)
 - `ClippingPolygons` now have support for specifying holes (aka islands) within each polygon. This works in inverse clipping workflows as well. [#13660](https://github.com/CesiumGS/cesium/pull/13660)
 - Added a `heightReference` option to `MVTDataProvider.fromUrl`, draping Mapbox Vector Tiles content onto terrain, 3D Tiles, or both. [#13727](https://github.com/CesiumGS/cesium/pull/13727)
-- Added a `heightReference` option to GeoJsonPrimitive constructor, draping GeoJSON content onto terrain, 3D Tiles, or both. [#13711](https://github.com/CesiumGS/cesium/pull/13711)
```
danielzhong, line 39: I think you accidently removed this line while resolving conflict?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -427,11 +430,27 @@ function commitSnapshot(primitive, snapshot, frameState) {
...
   primitive._positions = snapshot.positions;
   primitive._rotations = snapshot.rotations;
   primitive._scales = snapshot.scales;
   primitive._colors = snapshot.colors;
+  primitive._featureIds = snapshot.featureIds;
+  primitive._allFeatureIds = snapshot.allFeatureIds ?? [];
+  primitive._featureIdCount = snapshot.featureIdCount ?? 0;
+  primitive._hasFeatureIds =
+    defined(snapshot.allFeatureIds) && snapshot.allFeatureIds.length > 0;
+  primitive._featureIdTexture = snapshot.featureIdTexture;
```
danielzhong: This texture is never released on teardown: `destroySnapshotTextures()` only destroys `gaussianSplatTexture` and `sphericalHarmonicsTexture`, and `GaussianSplatPrimitive.prototype.destroy()` doesn't clear `this._featureIdTexture`. Since it's allocated at `ContextLimits.maximumTextureSize` width (usually 16384), this
leaks several MB on every tileset load/destroy cycle.

Please add it to `destroySnapshotTextures()` and null it out in `destroy()`.

packages/engine/Source/Shaders/PrimitiveGaussianSplatVS.glsl
```diff
@@ -193,4 +193,24 @@ void main() {
...
 #endif
     v_splitDirection = u_splitDirection;
+
+#if defined(HAS_FEATURE_IDS)
+    // Read feature IDs from the dedicated feature ID texture.
+    // The texture width equals maximumTextureSize (= 2 * (splatRowMask + 1)),
+    // so the row address is derived from the existing splat row uniforms.
+    // NOTE: Currently supports up to 4 feature ID sets per splat (one texel).
+    uint fidWidth = uint((u_splatRowMask + 1) << 1);
+    ivec2 fidCoord = ivec2(int(texIdx % fidWidth), int(texIdx / fidWidth));
```
danielzhong, line 203: This addressing ignores `texelsPerSplat`, but `createFeatureIdTexture` writes rows at `floor(i / width) * texelsPerSplat`. With more than 4 feature ID sets the row address no longer matches the CPU layout, so *all* IDs are read from the wrong texel, not just the ones past index 3. That matches the zeros @javagl saw in his test data.

Also, `buildCustomFragmentShaderStructs` and `addVarying` declare `featureId_0..N` for any N, but only `_0.._3` are ever written, leaving the rest undefined.

Either implement multi-texel addressing, or cap support at 4 sets with a `oneTimeWarning` and fix the JSDoc, which currently claims multi-texel works.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1571,6 +1823,61 @@ GaussianSplatPrimitive.buildGSplatDrawCommand = function (
...
+            "HAS_PROPERTY_TABLE",
+            undefined,
+            ShaderDestination.FRAGMENT,
+          );
+          shaderBuilder.addUniform(
+            "sampler2D",
+            "u_propertyTableTexture",
+            ShaderDestination.FRAGMENT,
+          );
+          uniformMap.u_propertyTableTexture = function () {
```
danielzhong: This closure captures a texture owned by the tile content's `GltfLoader`, which is destroyed in `GaussianSplat3DTileContent.destroy()`. `this._structuralMetadata` is only ever assigned once (`if (!defined(...))`) and never invalidated when tiles unload, so this can dangle.

`HAS_PROPERTY_TABLE` / `u_propertyTableTexture` are also unreferenced in `PrimitiveGaussianSplatFS.glsl` - this whole block is currently dead. I'd drop it and add it back when metadata lookups are actually implemented.

packages/engine/Source/Scene/GaussianSplatRenderResources.js
danielzhong: I think this render state is immediately discarded: `buildGSplatDrawCommand` overwrites `renderStateOptions.blending = BlendingState.PRE_MULTIPLIED_ALPHA_BLEND` a few lines later. The block has no effect and can be removed.

packages/engine/Source/Scene/GaussianSplatRenderResources.js
```diff
@@ -3,72 +3,129 @@ import RenderState from "../Renderer/RenderState.js";
...
-   *
-   * @type {boolean}
-   * @default false
-   *
-   * @private
-   */
   this.hasSkipLevelOfDetail = false;
 
+  shaderBuilder.addDefine(
+    "USE_FRUSTUM_CULLING",
```
danielzhong: This define isn't referenced by any shader in the repo. Dead code.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1070,6 +1223,27 @@ function GaussianSplatPrimitive(options) {
...
+   * @private
+   */
+  this._customShader =
+    options.customShader ?? GaussianSplatPrimitive.DefaultCustomShader;
+
+  /**
+   * @type {Event}
+   * @private
+   */
+  this._customShaderCompilationEvent = new Event();
```
danielzhong, line 1279: This event is created and exposed but never raised.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -114,6 +115,38 @@ class GaussianSplat3DTileContent {
...
+     */
+    this._allFeatureIds = [];
+
+    /**
+     * Feature ID metadata info selected by the tileset's featureIdLabel.
+     * Contains propertyTableId, featureCount, etc.
+     * @type {undefined|FeatureIdAttribute}
+     * @private
+     */
+    this._featureIdInfo = undefined;
```
danielzhong: `_featureIdInfo` / `featureIdInfo` still have no consumers. Since picking and styling aren't supported for splats yet, I'd remove this and the `featureIdLabel` selection until they are.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -543,6 +574,37 @@ async function processGeneratedSplatTextureData(
...
+    // uint32 is otherwise unused by the shader.
+    const fd = snapshot.featureIds;
+    if (defined(fd)) {
+      const splatsPerRow = optimalWidth / 2;
+      const numSplats = Math.min(snapshot.numSplats, fd.length);
+      for (let i = 0; i < numSplats; i++) {
+        const col = (i % splatsPerRow) * 2;
+        const row = Math.floor(i / splatsPerRow);
+        const base = (row * optimalWidth + col) * 4;
+        effectiveData[base + 3] = fd[i];
```
danielzhong: Nothing reads this. The VS only uses `splatPosition.xyz`, and "backward compat with position alpha readers" doesn't refer to any existing code path. This is a full pass over every splat on each snapshot for no effect, please remove.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1683,6 +1991,25 @@ GaussianSplatPrimitive.buildGSplatDrawCommand = function (
...
+  // Update the custom shader in case it has texture uniforms.
+  this._customShader.update(frameState);
+
+  if (
+    this._shaderDirty &&
+    defined(this._drawCommand) &&
+    defined(this._indexes) &&
+    defined(this.gaussianSplatTexture)
+  ) {
+    GaussianSplatPrimitive.buildGSplatDrawCommand(this, frameState);
```
danielzhong, line 2046: `buildGSplatDrawCommand` reassigns `_drawCommand` without destroying the previous `shaderProgram`, so every `customShader` change leaks one program, easy to hit with the toolbar menu in the new Sandcastle.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1683,6 +1991,25 @@ GaussianSplatPrimitive.buildGSplatDrawCommand = function (
 GaussianSplatPrimitive.prototype.update = function (frameState) {
   const tileset = this._tileset;
 
+  const targetCustomShader = defined(tileset.customShader)
+    ? tileset.customShader
+    : GaussianSplatPrimitive.DefaultCustomShader;
+  if (this._customShader !== targetCustomShader) {
+    this.customShader = tileset.customShader;
```
danielzhong, line 2034: This makes the `options.customShader` constructor argument a no-op: the first `update()` always overwrites it with `tileset.customShader`. The spec "customShader can be provided at construction" only passes because it never calls `update()`. Either drop the constructor option, or don't clobber an explicitly-set shader.

packages/engine/Source/Scene/GaussianSplatRenderResources.js
```diff
@@ -3,72 +3,129 @@ import RenderState from "../Renderer/RenderState.js";
...
-   * @readonly
-   *
-   * @private
-   */
+  // Custom shader uniforms
+  const customShader = primitive._customShader;
+  const uniformMap = defined(customShader)
+    ? combine(primitive._uniformMap, customShader.uniformMap)
+    : Object.assign({}, primitive._uniformMap);
+  primitive._uniformMap = uniformMap;
```
danielzhong: Constructing render resources shouldn't mutate the primitive that was passed in. Consider returning the combined map and letting the caller assign it.

packages/engine/Source/Scene/GltfLoader.js
```diff
@@ -1297,7 +1297,23 @@ function loadAttribute(
...
+  const isSpzDecodable =
+    defined(spz) &&
+    (gltfSemantic === "POSITION" ||
+      gltfSemantic === "KHR_gaussian_splatting:SCALE" ||
+      gltfSemantic === "_SCALE" ||
+      gltfSemantic === "KHR_gaussian_splatting:ROTATION" ||
+      gltfSemantic === "_ROTATION" ||
+      gltfSemantic === "COLOR_0" ||
+      gltfSemantic === "KHR_gaussian_splatting:OPACITY" ||
+      gltfSemantic.includes("SH_DEGREE_"));
```
danielzhong: This duplicates knowledge that lives in `GltfSpzLoader` and will drift; `includes()` is also a loose substring match. Could we export the set of SPZ-provided semantics (or an `isSpzSemantic()` helper) from `GltfSpzLoader` and use it here?

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -508,12 +545,111 @@ class GaussianSplat3DTileContent {
...
+              data = new Uint32Array(attr.typedArray);
+            }
+          }
+          // Generate implicit sequential IDs when the vertex attribute has
+          // no data (e.g. accessor had no bufferView, or SPZ didn't produce it).
+          if (!defined(data)) {
+            const count = this.pointsLength;
+            data = new Uint32Array(count);
+            for (let j = 0; j < count; j++) {
+              data[j] = j;
```
danielzhong: This branch is only reached when `fid.setIndex` is defined, i.e. the data declares a feature ID attribute but provides no values - a data error. Synthesizing 0..N-1 silently hides it. It also allocates a `Uint32Array(pointsLength)` per set, so memory scales with the number of sets.

packages/engine/Source/Scene/parseStructuralMetadata.js
```diff
@@ -236,14 +242,21 @@ function collectGpuCompatiblePropertyInfo(
     const bufferViewLength = bufferView.length;
     const bytesPerElement = classProperty.cpuBytesPerElement();
     const numBufferElements = bufferViewLength / bytesPerElement;
-    if (numBufferElements !== numFeatures) {
+    if (numBufferElements < numFeatures) {
```
danielzhong: Could you explain what data motivated this? From the validator output earlier in this -thread, the source glbs have alignment problems, and this looks like it's relaxing the loader to accept them.

packages/engine/Source/Core/clone.js
```diff
@@ -14,8 +14,8 @@ function clone(object, deep) {
 
   deep = deep ?? false;
 
-  const Constructor = object.constructor;
-  const result = new Constructor();
+  const ObjectConstructor = object.constructor;
```
danielzhong: It looks like these renames (here, `Matrix2.js`, `PixelFormat.js`, `Primitive.js`, and the blank-line move in `Vector3DTilePrimitive.js`) are unrelated to this PR.

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1119,8 +1293,69 @@ Object.defineProperties(GaussianSplatPrimitive.prototype, {
...
   },
+
+  /**
+   * Gets or sets the custom shader. If undefined, {@link GaussianSplatPrimitive.DefaultCustomShader} is set.
+   *
+   * @memberof GaussianSplatPrimitive.prototype
+   * @type {CustomShader}
+   * @see {@link https://github.com/CesiumGS/cesium/tree/main/Documentation/CustomShaderGuide|Custom Shader Guide}
+   */
+  customShader: {
```
danielzhong, line 1338: Missing `@private`.

packages/sandcastle/gallery/3d-tiles-gaussian-splatting-shading/main.js
```diff
@@ -0,0 +1,116 @@
...
+  terrain: Cesium.Terrain.fromWorldTerrain(),
+  timeline: false,
+  animation: false,
+  baseLayerPicker: false,
+  sceneModePicker: false,
+});
+
+// Height-based colorization: colors cycle through the rainbow every 10 metres.
+const heightShader = new Cesium.CustomShader({
+  lightingModel: Cesium.LightingModel.UNLIT,
```
danielzhong: `lightingModel` has no effect on the splat path, consider removing it.

CHANGES.md
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatRenderResources.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/sandcastle/gallery/3d-tiles-gaussian-splatting-shading/main.js
dbiguenet: Fixed

packages/engine/Source/Core/clone.js
dbiguenet: I had to rename these constructors, otherwise eslint checks would fail

dbiguenet: I will try to remove these changes

dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Shaders/PrimitiveGaussianSplatVS.glsl
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatRenderResources.js
javagl: > maybe it's necessary, maybe not 

Turns out it's not... **theoretically**. 

It's building a `GaussianSplatRenderResources` at https://github.com/dbiguenet/cesium/blob/9751fb1f5e6476edbc0b378e5ec7f73df715c2f8/packages/engine/Source/Scene/GaussianSplatPrimitive.js#L1475 , and this is the only place where this is done. It would be preferable to **not** have to assign some highly specific values to some properties (that IMHO preferably should be private and immutable) just to "bring this object into the desired/consistent state". 

So I'd **strongly** lean towards initializing this `GaussianSplatRenderResources` properly (i.e. **here**, in its constructor), so that the whole block at https://github.com/dbiguenet/cesium/blob/main/packages/engine/Source/Scene/GaussianSplatPrimitive.js#L1478-L1483 can be removed (unless there is a strong reason to not do this)

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -295,6 +303,10 @@ function destroySnapshotTextures(snapshot) {
     snapshot.sphericalHarmonicsTexture.destroy();
     snapshot.sphericalHarmonicsTexture = undefined;
   }
+  if (defined(snapshot._featureIdTexture)) {
+    snapshot._featureIdTexture.destroy();
+    snapshot._featureIdTexture = undefined;
+  }
```
danielzhong: Typo: should be `snapshot.featureIdTexture` (no underscore)?

dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Fixed

packages/engine/Source/Scene/GaussianSplatRenderResources.js
dbiguenet: Fixed

packages/engine/Source/Scene/GltfLoader.js
dbiguenet: Fixed

packages/engine/Source/Scene/parseStructuralMetadata.js
dbiguenet: Unfortunately, I don't remember precisely why I replaced "!==" with "<". However, I don't think this change is still relevant.
I propose to restore the strict length check

packages/engine/Source/Scene/GaussianSplatRenderResources.js
dbiguenet: Attempt to fix this in https://github.com/CesiumGS/cesium/pull/13418/changes/42fab712bf17683eb008e8ffda23d25f1edf5022, what do you think?

javagl: I (subjectively) think that this is "cleaner" now. The idea being to make sure that the constructor initializes the object, as far as reasonably possible, to be in a "consistent" state. 

(But ... dissecting the code as a whole is beyond what I can do in reasonable time. This is related to the reasons for why I have strong opinions about many things that are related to 'draw commands')

Daniel to confirm as part of the actual, final review.

dbiguenet:
Comments have been addressed, I would need another review

packages/engine/Source/Scene/GaussianSplatPrimitive.js
```diff
@@ -1118,8 +1326,70 @@ Object.defineProperties(GaussianSplatPrimitive.prototype, {
...
+  },
+
+  /**
+   * Gets an event that is raised whenever a custom shader is compiled.
+   *
+   * @memberof GaussianSplatPrimitive.prototype
+   * @type {Event}
+   * @readonly
+   */
+  customShaderCompilationEvent: {
```
danielzhong, line 1375: Please add @private to js docs

packages/engine/Source/Scene/GaussianSplatRenderResources.js
```diff
@@ -3,70 +3,114 @@ import RenderState from "../Renderer/RenderState.js";
...
-  /**
-   * Whether the model is part of a tileset that uses the skipLevelOfDetail
-   * optimization. This value indicates what draw commands are needed and
-   * is set by TilesetPipelineStage.
-   *
-   * @type {boolean}
-   * @default false
-   *
-   * @private
-   */
```
danielzhong, line 69: Is there any reason removing the JSDoc?

packages/engine/Source/Scene/GaussianSplatRenderResources.js
```diff
@@ -3,70 +3,114 @@ import RenderState from "../Renderer/RenderState.js";
...
-  /**
-   * An object storing options for creating a {@link RenderState}.
-   * The pipeline stages simply set the options, the render state is created
-   * when the {@link DrawCommand} is constructed.
-   *
-   * @type {object}
-   * @readonly
-   *
-   * @private
-   */
```
danielzhong, line 40: Any reason remove JS doc for: shaderBuilder, uniformMap, renderStateOptions, alphaOptions, hasSkipLevelOfDetail

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -114,6 +115,30 @@ class GaussianSplat3DTileContent {
      */
     this._packedSphericalHarmonicsData = undefined;
 
+    /**
+     * Per-splat feature IDs loaded from the vertex attribute selected by
+     * the tileset's featureIdLabel (defaults to _FEATURE_ID_0).
+     * @type {undefined|Uint32Array}
+     * @private
+     */
+    this._featureIds = undefined;
```
danielzhong, line 124: Nothing assigns `_featureIds` - the `featureIdLabel` selection that used to set it was removed, so it stays `undefined` forever. That makes `hasFeatureIds` in `GaussianSplatPrimitive.update()` always false, which leaves the whole singular path unreachable: the aggregation loop, `snapshot.featureIds`, its truncation block in
`processGeneratedSplatTextureData`, `primitive._featureIds`, and `_aggregateScratchBuffers.featureIds`.

Rendering goes through `_allFeatureIds` instead. Could the singular path be removed? The JSDoc here and on the `featureIds` getter still reference `featureIdLabel`, which no longer applies.

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -508,12 +537,103 @@ class GaussianSplat3DTileContent {
...
+            }
+            data = sequentialFeatureIds;
+          }
+          allFeatureIds.push(data);
+        }
+      }
+      this._allFeatureIds = allFeatureIds;
+
+      // Store structural metadata (property tables, schema, etc.)
+      if (defined(loader.components.structuralMetadata)) {
```
danielzhong, line 588: Nothing reads this - `grep structuralMetadata packages/engine/Source/Scene/GaussianSplat*.js` only finds these two assignments. The property-table wiring in `GaussianSplatPrimitive` that consumed it was removed. Can this and the `structuralMetadata` getter go too?

packages/engine/Source/Scene/GaussianSplatPrimitive.js
dbiguenet: Done

packages/engine/Source/Scene/GaussianSplatRenderResources.js
dbiguenet: No. I reverted the change

packages/engine/Source/Scene/GaussianSplatRenderResources.js
dbiguenet: No, I reverted the change

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
dbiguenet: See https://github.com/CesiumGS/cesium/pull/13418/changes/b73ca6a76a14546e189d99ace2aae95d24b91f05

packages/engine/Source/Scene/GaussianSplat3DTileContent.js
dbiguenet: See c0a195e2e9859a65c7fa59d335e7946c878bcdb3

packages/engine/Source/Scene/GaussianSplatRenderResources.js
```diff
@@ -26,7 +76,42 @@ function GaussianSplatRenderResources(primitive) {
    *
    * @private
    */
-  this.uniformMap = {};
+  const uniformMap = defined(customShader)
```
danielzhong, line 79: This block documents the `uniformMap` instance property, but it's attached to the local `const` here - the actual assignment is `this.uniformMap = uniformMap;` below.

packages/sandcastle/gallery/3d-tiles-gaussian-splatting-shading/main.js
```diff
@@ -0,0 +1,115 @@
...
+
+    material.diffuse = color;
+}
+`,
+});
+
+// Classification-based colorization: splats are tinted according to a
+// per-feature classification code stored in the first two bytes of the
+// property table texture (16-bit little-endian integer).
+// Requires a dataset that contains EXT_structural_metadata classification data.
```
danielzhong, line 42: These three lines describe reading from a property table texture, but the shader below reads `fsInput.attributes.featureId_0` - that comes from the `_FEATURE_ID_0` vertex attribute via `EXT_mesh_features`, not from a property table or `EXT_structural_metadata`. The `sandcastle.yaml` description says "stored in a property table" too. Could both be updated? Otherwise users will look for the wrong extension in their data.

danielzhong (commented):
LGTM, let's wait for @jjhembd for final approval.

danielzhong:
From Sean's comment:
I think the height shader may be suffering from float32 precision issues.

PR #13421, opened 2026-04-21
Fix: edge visibility bug

danielzhong:
# Description
Problem:
1. GLB files produced by TilesetPublisher contain degenerate triangles — triangles where two or more vertices share the same position, giving them zero area. In buildTriangleAdjacency, the cross product of the two edge vectors for such a triangle is the zero vector. Calling Cartesian3.normalize on it throws:
`Error: normalized result is not a number`

These have always existed in the data, but previously went unnoticed because face normal computation was done in the GPU shader, where normalize(vec3(0)) silently returns vec3(0).

https://github.com/CesiumGS/cesium/pull/13110 moved face normal computation to the CPU, introducing the first JS call to `Cartesian3.normalize` on per-triangle cross products. Unlike the GPU, `Cartesian3.normalize` throws a `DeveloperError` when given the zero vector, causing the tile to fail to load entirely.

This fix makes Cesium defensively skip degenerate triangles, restoring the previous tolerance for imperfect mesh data regardless of its origin.

2. Cleanup and simplify silhouette normal decoding

The previous implementation decoded GLB signed-byte normals, then immediately re-encoded them to 16-bit oct format, packed two into a Uint32, and later unpacked and decoded them back to floats before writing into the Float32Array vertex buffer. This round-trip (signed byte → oct16 → Uint32 → oct16 → float) introduced quantization loss and unnecessary complexity.

The normals are now decoded once — signed byte → normalized Cartesian3 — and stored directly in a plain JS array. Each edge lookup reads directly from this array. The final vertex buffer (Float32Array) is unchanged. The unused Cartesian2 and `AttributeCompression` imports are removed.

## Issue number and link
https://github.com/CesiumGS/tilers/issues/2156

## Author checklist

## AI acknowledgment

GitHub Copilot

Claude 4.6 Sonnect

### DIFF
--- a/packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
@@ -367,7 +367,11 @@ function buildTriangleAdjacency(primitive) {
       continue;
     }
 
-    Cartesian3.normalize(scratchCross, scratchCross);
+    Cartesian3.multiplyByScalar(
+      scratchCross,
+      1.0 / Math.sqrt(crossMagnitudeSquared),
+      scratchCross,
+    );
 
     faceNormals[base] = scratchCross.x;
     faceNormals[base + 1] = scratchCross.y;
--- a/packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
@@ -5,9 +5,7 @@ import defined from "../../Core/defined.js";
 import IndexDatatype from "../../Core/IndexDatatype.js";
 import ComponentDatatype from "../../Core/ComponentDatatype.js";
 import PrimitiveType from "../../Core/PrimitiveType.js";
-import Cartesian2 from "../../Core/Cartesian2.js";
 import Cartesian3 from "../../Core/Cartesian3.js";
-import AttributeCompression from "../../Core/AttributeCompression.js";
 import Pass from "../../Renderer/Pass.js";
 import ShaderDestination from "../../Renderer/ShaderDestination.js";
 import EdgeVisibilityStageFS from "../../Shaders/Model/EdgeVisibilityStageFS.js";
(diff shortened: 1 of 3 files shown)


packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -358,6 +356,14 @@ function buildTriangleAdjacency(primitive) {
     Cartesian3.subtract(scratchP1, scratchP0, scratchE1);
     Cartesian3.subtract(scratchP2, scratchP0, scratchE2);
     Cartesian3.cross(scratchE1, scratchE2, scratchCross);
+
+    // Skip degenerate triangles (zero-area): their cross product is the zero
+    // vector, which cannot be normalized and would produce NaN face normals.
+    const crossMagnitudeSquared = Cartesian3.magnitudeSquared(scratchCross);
```
javagl, line 362: I don't know how often this is called, and how performance-critical it is. So the following _might_ not be important, but I'll mention it anyhow: Performance purists could claim that the (squared) length is computed _twice_ here now - once for the check, and once for the `normalize` call. And these purists could suggest to change the `normalize` call to 
```
multiplyByScalar(... 1.0 / Math.sqrt(crossMagnitudeSquared) ...);
```
But my gut feeling is that this will not immediately affect _actual_ performance in any way.

danielzhong: Great point, modified!

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -5,9 +5,7 @@ import defined from "../../Core/defined.js";
 import IndexDatatype from "../../Core/IndexDatatype.js";
```
markschlosseratbentley, line 5: Consider a CHANGES.md entry.

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -5,9 +5,7 @@ import defined from "../../Core/defined.js";
...
 import ComponentDatatype from "../../Core/ComponentDatatype.js";
 import PrimitiveType from "../../Core/PrimitiveType.js";
```
markschlosseratbentley, line 7: Is a unit test possible?

markschlosseratbentley (approved):
LGTM - make sure you mark the checklist items in the PR desc.

PR #13425, opened 2026-04-21
Types: Convert Credit to ES6 Class

lukemckinstry:
# Description

types `diff` is empty!

WIP - process question, when I tried activating `// @ts-check` there were a few of what I consider legitimate type errors we will need to sort out (for example we are adding a `_creditId` identifier field to HTMLElement). I left type checking turned off rather than trying to define with jsDoc type annotations or make further changes. Is that the right move, or is it better to turn on type checking and either a) mark the trouble spots with `// ts-expect-error` or b) try to resolve the type issues with type annotations or further changes. 

## Author checklist

- [NA] I have updated `CHANGES.md` with a short summary of my change

## AI acknowledgment

### DIFF
--- a/packages/engine/Source/Core/Credit.js
+++ b/packages/engine/Source/Core/Credit.js
@@ -19,41 +19,39 @@ const creditToId = {};
  * // Create a credit with a tooltip, image and link
  * const credit = new Cesium.Credit('<a href="https://cesium.com/" target="_blank"><img src="/images/cesium_logo.png"  style="vertical-align: -7px" title="Cesium"/></a>');
  */
-function Credit(html, showOnScreen) {
-  //>>includeStart('debug', pragmas.debug);
-  Check.typeOf.string("html", html);
-  //>>includeEnd('debug');
-  let id;
-  const key = html;
+class Credit {
+  constructor(html, showOnScreen) {
+    //>>includeStart('debug', pragmas.debug);
+    Check.typeOf.string("html", html);
+    //>>includeEnd('debug');
+    let id;
+    const key = html;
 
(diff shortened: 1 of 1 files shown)


donmccurdy:
>  I left type checking turned off rather than trying to define with jsDoc type annotations or make further changes. Is that the right move...?

Personally I would probably do...

```javascript
// @ts-expect-error Consider using a data-id attribute, instead.
div._creditId = this._id;
```

...and adding JSDoc type annotations to fix the remaining errors in the file. Admittedly the above is basically a "TODO", but it's probably unavoidable that we'll have at least a bit of that as we introduce type definitions.

That said, leaving `// @ts-check` off the file is entirely OK too I think – conversion to an ES6 class is still progress, wouldn't want to block or slow that in the pursuit of other, additional progress. :)

packages/engine/Source/Core/Credit.js
```diff
@@ -1,172 +1,166 @@
+// @ts-check
+
 import DOMPurify from "dompurify";
 import Check from "./Check.js";
 import defined from "./defined.js";
 
 let nextCreditId = 0;
+/** @type {Record<string, number>} */
 const creditToId = {};
```
donmccurdy: Running `./script/diff-types.sh` again will show `creditToId` in the exported types — it needs an `@ignore` or `@private` tag. It believe that's just a tsd-jsdoc thing, since it's based on documentation it doesn't know what's exported or not. Maayyybe something we could get ast-grep to detect on CI too...

donmccurdy (approved):
One minor issue, LGTM otherwise!

PR #13431, opened 2026-04-23
Bump required node version to 22

### LINKED ISSUE #13362: Update minimum node version to `22`
Node `v20` is [reaching End of Life](https://github.com/CesiumGS/cesium/issues/13362) this month so we should bump our minimum up to Node `v22`.

This has come up a few times recently where we were [unable to update packages](https://github.com/CesiumGS/cesium/pull/13348#discussion_r3022600863) (https://github.com/CesiumGS/cesium/issues/13239) or unable to [use newer node functions](https://github.com/CesiumGS/cesium/pull/13210#discussion_r2805235506). These sort of things could be addressed at the same time.

Locally all devs need to make sure they update their node versions.

In all CI workflows we should bump to using Node 22 and/or 22 + 24 the few places we test that

jjspace:
# Description

LTS support for Node 20 [ends this month](https://nodejs.org/en/about/previous-releases). This PR bumps the minimum required version up to Node 22.0.0 and updates our CI to account for that.

## Issue number and link

Resolves #13362 

## Testing plan

- Check that CI is working as expected
- Do a smoke test locally if you were not running node >= 22 before. I'm pretty sure multiple team members were so if it broke anything I think we would've seen already.

## Author checklist

## AI acknowledgment

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -70,14 +70,31 @@ jobs:
         run: npm run test -- --browsers ChromeHeadless --failTaskOnError --webgl-stub --release --suppressPassed
       - name: cloc
         run: npm run cloc
-  node-20:
+  node-24:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - name: install node 20
+      - name: install node 24
         uses: actions/setup-node@v6
         with:
-          node-version: '20'
+          node-version: '24'
+      - name: npm install
+        run: npm install
+      - name: release build
+        run: npm run build-release
+      - name: package cesium module
+        run: npm pack &> /dev/null
+      - name: package workspace modules
+        run: npm pack --workspaces &> /dev/null
+      - uses: ./.github/actions/verify-package
+  node-22:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: install node 22
+        uses: actions/setup-node@v6
+        with:
+          node-version: '22'
       - name: npm install
         run: npm install
       - name: release build
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -70,31 +70,17 @@ jobs:
         run: npm run test -- --browsers ChromeHeadless --failTaskOnError --webgl-stub --release --suppressPassed
       - name: cloc
         run: npm run cloc
-  node-24:
+  node-smoke-test-matrix:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version: [22, 24]
     steps:
       - uses: actions/checkout@v6
-      - name: install node 24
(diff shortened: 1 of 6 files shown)


.github/workflows/dev.yml
```diff
@@ -70,14 +70,31 @@ jobs:
...
+        run: npm pack --workspaces &> /dev/null
+      - uses: ./.github/actions/verify-package
+  node-22:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: install node 22
+        uses: actions/setup-node@v6
+        with:
+          node-version: '22'
```
donmccurdy: Would we prefer to structure this as a [matrix strategy](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#adding-a-matrix-strategy-to-your-workflow-job), if there's no difference between the steps for each Node.js version? Example:

https://github.com/donmccurdy/glTF-Transform/blob/3378a5d3085a98900d382e37c6b5610b9fff3416/.github/workflows/ci.yml#L12-L26

donmccurdy (commented):
I've been using Node.js v24 locally for a while without incidents, so no issues expected in terms of the development environment!

For Node.js applications or libraries that depend on cesium, I am less sure what to expect from the update, but supporting only maintained Node.js versions seems entirely reasonable. E.g. [gltf-pipeline](https://github.com/CesiumGS/gltf-pipeline) will need to update its Node.js range the next time it upgrades CesiumJS.

.github/workflows/dev.yml
jjspace: Good call, updated!

jjspace:
> For Node.js applications or libraries that depend on cesium, I am less sure what to expect from the update

I marked this as a Breaking change in the changelog. The sub-packages will get a major version bump because of this. If people really do need older node support then they should continue to use older versions of the packages. For those that can update I expect that they do that at the same time they update CesiumJS. As you said, I don't think we should be expected to support a version that the Node.js team doesn't even support anymore.

.github/workflows/dev.yml
jjhembd, line 61: Don't we want the default workflows to run on the latest LTS? (24 in this case).

Then the smoke test is to make sure earlier LTS versions still work too.

jjspace: I honestly wasn't sure about this. I left them all on 22 to help make sure we're checking on the minimum version we support even if devs use a newer version locally. Node should be forward compatible so everything _should_ work on newer versions if people choose to use them. However it's not backwards compatible so if someone uses a newer Node feature it could break for people still using an older version.

jjhembd: OK, this is a (small) change to our approach from previous updates, but I think it makes sense!

jjhembd (approved):
Thanks @jjspace!

Issue #11683, opened 2023-12-11
EXT_mesh_features extension ignoring my class/table definitions, always looking up in propertyTable at index 0


arneschilling:
Hi, I have produced a 3D Tiles 1.1 dataset, which is shown in the following example

Sandcastle example: https://sandcastle.cesium.com/#c=jZJvb9owEMa/ipUXVZA2pwQVUY1Wy6BlYQVGG9oG5Y1JTHHi2JntQJOp333OH7Z22qa9suz73XN3z9mywEQgpsAIS5Kn0zuAwhBLCRQHBc8FIJwBJCVWMmANA13OYIS3KKfKqWGPJ5iBCxAYuJjuNpOQLMjUXZVud05c6bLbs3Dk9t0ke7wfTc+hhr5Fk0RD7qmfXifrdE1n5PTMt693N94tmcdL5T+sqV+clgsvsf14Zs8m82TmOeRmNM3WWmwWP9mzeKnF5/Gm94mGhdt/0MXDnlsxzC+mUcX6j0uyiK+6C++qmJerch7vzuEY+9x5z9wBHvt2vikO5eh8tVncdseR8/XzWRENvCVbO1+2+6vA+BCwkDOpwJ7gAxZ6TIYPrV3wvn4zAyOs7yPOFCIMi8Do6LyAKVGA7wEDoJFQhGJtpdZAB0SOpsPm6I29Jgy3gqfaZKey3Y1Muzfo9+1BpQjaLqAMMcMwEyQliuyxhCiKzFa+AZsKLV5ynnr8NVAhlgWcLKMFUDsM2o0CqQqKAdkCnY2fiawWf+wfPyuBpG6/FYL1z4DNc11V55nVCY6zaVltSGQ2TAecnPwjDPVv+w/k+PvuqlY7FdxpXAY/G2umeLOr1y7XmX9VrCd5CdgLCJEKd8DEQnDR+bVJTjGk/Kl917iGjXfGsK56WVEfSZpxoUAuqAmhpXCaUaSwtDZ5mOj+QimrvKF1TBlGZA9IdPGHrwRCqn3WkW1O6R0pcWBcDi3Nv0mjHEWEPS32WFBUaKRqY7jrXt40AQjh0NLXqujvuYpzukHile4P

[...]

PR #13432, opened 2026-04-23
Vector Tiles: Support multiple property tables per tile

donmccurdy:
# Description

A vector tile may contain any number of property tables. For example, it would be common for a tile containing a mix of point/line/polygon data to have different property tables, corresponding to different schema, for each topology. This PR implements support for multiple property tables per tile, which previously caused a runtime error.

## Issue number and link

- #11683

## Testing plan

Unit tests updated. Currently I only have internal datasets to reproduce the bug, and it's non-trivial to produce more. Requires a glTF tile containing >1 primitive using >1 batch table.

## Author checklist

    - Support for `EXT_structural_metadata` in Vector Tiles will be new in this release, no need to list pre-release bug fixes.

## AI acknowledgment

#### PR Dependency Tree

* **PR #13432** 👈


packages/engine/Source/Scene/Cesium3DTileVectorFeature.js
```diff
@@ -73,10 +73,12 @@ class Cesium3DTileVectorFeature {
   /**
    * @param {VectorGltf3DTileContent} content
    * @param {number} batchId
+   * @param {number} [batchTableId=0]
    */
-  constructor(content, batchId) {
+  constructor(content, batchId, batchTableId = 0) {
```
donmccurdy, line 79: Kept this parameter optional for API compatibility with Cesium3DTileContent, but possibly the same parameter will be needed on Cesium3DTileContent to resolve https://github.com/CesiumGS/cesium/issues/11683 ?

packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -155,19 +156,17 @@ class VectorGltf3DTileContent {
...
-      const featureTableId = model._featureTableId;
-
-      if (defined(featureTables) && defined(featureTables[featureTableId])) {
-        return featureTables[featureTableId];
-      }
-    }
+  /** @type {Cesium3DTileBatchTable[]} */
+  get batchTables() {
+    return this._model._featureTables;
+  }
```
donmccurdy, line 162: The Model class always defines `_featureTables` (default to empty array), and VectorGltf3DTileContent defines `_model` immediately, so we shouldn't need so much error handling here. Can add it back with comments/explanation if we run into runtime errors! :)

lilleyse (approved):
Looks great!

How much of a lift would it be to also fix https://github.com/CesiumGS/cesium/issues/11683, maybe in a separate PR?

donmccurdy:
Let's try it! I'd like to put some type checking in place first, to keep the Cesium3DTileContent implementations all aligned, so starting here:

- https://github.com/CesiumGS/cesium/pull/13443

donmccurdy:
Continued:

- https://github.com/CesiumGS/cesium/pull/13467

PR #13433, opened 2026-04-24
Refactor ModelReader and pickModel

alarkbentley:
## Summary

Extracts scene-graph traversal, instance transform computation, and GPU buffer readback logic from `pickModel.js` into reusable static methods on `ModelReader`, then rewrites `pickModel` to use those shared helpers. Adds comprehensive unit tests for `ModelReader`.

## Motivation

`pickModel.js` contained ~300 lines of inline logic for:
- Walking the model scene graph and computing per-node transforms
- Reading instance transforms from GPU buffers or individual Translation, Rotation, Scale attributes
- Reading vertex positions (with interleaving, quantization, and dequantization)
- Reading index buffers from the GPU

This logic is useful beyond picking (e.g. spatial queries, custom analysis) but was not reusable. This PR centralizes it in `ModelReader` so other consumers can leverage the same code paths.

### `ModelReader.js`
- **`forEachPrimitive(model, options, callback)`** — new static method that traverses every primitive in a model's scene graph. Computes node transforms (including world-space instancing and 2D map projection), builds per-instance transform matrices + optional per-instance feature IDs, and invokes a callback for each runtime primitive.
- **`readImplicitRangeAsTypedArray(featureIdSet, attributeOwner)`** — new static method that generates feature IDs from an implicit-range feature ID set (`offset + floor(i / repeat)`).
- **`readAttributeAsTypedArray`** — when an attribute already has an in-memory `typedArray`, return it directly instead of re-reading from the GPU buffer (the typed array is already tightly packed).
- Extracted private helpers: `computeNodeTransforms`, `getInstanceTransforms`, `getInstanceFeatureIds`, `buildTransformsFromTypedArray`, `buildTransformsFromAttributes`.

### `pickModel.js`
- Replaced ~300 lines of inline traversal / buffer-reading code with a single call to `ModelReader.forEachPrimitive` + `ModelReader.readAttributeAsTypedArray` / `ModelReader.readIndicesAsTypedArray`.
- Removed the `getVertexPosition` helper (quantization/dequantization is now handled inside `ModelReader.readAttributeAsTypedArray`).

### `InstancingPipelineStage.js`
- Respect when `keepTypedArray` is true, always go through the matrix code path (not just when a rotation attribute is present). This ensures `runtimeNode.transformsTypedArray` is populated for CPU-side readback even for translation+scale-only instances. Without this fix pickModelSpec will not be able to execute correctly using webgl1.

### `ModelReader.js`
- **`octDecode` bug fix** — Fixed incorrect argument order in `AttributeCompression.octDecodeInRange` call (was passing a `Cartesian3`, now correctly passes `c.x, c.y`) and fixed swapped arguments in the subsequent `Cartesian3.pack` call. (found in [pickModel](https://github.com/CesiumGS/cesium/blob/0da68a8f1055f534f43d7d74a7050e193246fbc7/packages/engine/Source/Scene/Model/pickModel.js#L386))
- **`getInstanceTransforms` bug fix** — Fixed incorrect matrix multiplication order for meshes with EXT_mesh_gpu_instancing. (found in [pickModel](https://github.com/CesiumGS/cesium/blob/0da68a8f1055f534f43d7d74a7050e193246fbc7/packages/engine/Source/Scene/Model/pickModel.js#L162))

## Tests
- **`ModelReaderSpec.js`** — new spec file with ~2100 lines of tests covering attribute reading (float, quantized, interleaved, GPU-backed), index reading, implicit range feature IDs, `forEachPrimitive` traversal (single node, instanced, multi-node, 2D projection, feature IDs), and edge cases.
- **`pickModelSpec.js`** — updated instanced picking test to be more explicit about expected geometry and camera setup; fixed one test to correctly use `sceneWithWebgl1`.

## Testing

All existing `pickModelSpec` tests pass with the refactored implementation. New `ModelReaderSpec` tests exercise the extracted helpers in isolation.

# Author checklist

## AI acknowledgment

GitHub Copilot

Claude 4.6 Opus

### DIFF
--- a/packages/engine/Source/Scene/Model/ModelReader.js
+++ b/packages/engine/Source/Scene/Model/ModelReader.js
@@ -9,8 +9,13 @@ import AttributeCompression from "../../Core/AttributeCompression.js";
 import IndexDatatype from "../../Core/IndexDatatype.js";
 import PrimitiveType from "../../Core/PrimitiveType.js";
 import Matrix4 from "../../Core/Matrix4.js";
+import Quaternion from "../../Core/Quaternion.js";
+import Transforms from "../../Core/Transforms.js";
 
 import AttributeType from "../AttributeType.js";
+import InstanceAttributeSemantic from "../InstanceAttributeSemantic.js";
+import VertexAttributeSemantic from "../VertexAttributeSemantic.js";
+import ModelUtility from "./ModelUtility.js";
 
 /**
  * A class for reading the data from a <code>ModelComponents.Attribute</code>.
@@ -144,7 +149,20 @@ class ModelReader {
     const byteStride = attribute.byteStride;
     const bytesPerComponent = ComponentDatatype.getSizeInBytes(componentType);
     const defaultByteStride = componentsPerElement * bytesPerComponent;
-    if (!defined(byteStride) || byteStride === defaultByteStride) {
+
+    const isDefaultStride =
+      !defined(byteStride) || byteStride === defaultByteStride;
+    const hasTypedArray = defined(attribute.typedArray);
+
+    // If in-memory — return as-is
+    // Important: typedArray is already tightly-packed on creation (See: ModelComponents.typedArray and GltfLoader.getPackedTypedArray)
+    //            byteOffset and byteStride should thus be ignored
+    if (hasTypedArray) {
+      return attribute.typedArray;
+    }
+
+    // Non-interleaved — copy from typedArray or read from GPU
+    if (isDefaultStride) {
       const typedArray = ComponentDatatype.createTypedArray(
         componentType,
         totalComponentCount,
@@ -161,6 +179,8 @@ class ModelReader {
     // have ONE "TypedArray[] getThemFrom(buffer)" call that
     // returns all of the (interleaved) attributes at once,
     // but this requires abstractions that we don't have.
+
+    // Read back from GPU if not available in memory
     const fullTypedArray = new Uint8Array(buffer.sizeInBytes);
     buffer.getBufferData(fullTypedArray);
 
@@ -320,8 +340,8 @@ class ModelReader {
     const c = new Cartesian3();
     for (let i = 0; i < elementCount; i++) {
       Cartesian3.unpack(quantizedTypedArray, i * 3, c);
-      AttributeCompression.octDecodeInRange(c, normalizationRange, c);
-      Cartesian3.pack(dequantizedTypedArray, c, i * 3);
+      AttributeCompression.octDecodeInRange(c.x, c.y, normalizationRange, c);
+      Cartesian3.pack(c, dequantizedTypedArray, i * 3);
     }
     return dequantizedTypedArray;
   }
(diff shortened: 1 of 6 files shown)


packages/engine/Source/Scene/Model/ModelReader.js
```diff
@@ -805,6 +852,431 @@ class ModelReader {
...
+ *
+ * @param {object} runtimeNode The runtime node.
+ * @param {ModelSceneGraph} sceneGraph The model scene graph.
+ * @param {Model} model The model.
+ * @param {object} result An object with scratch matrices: { nodeComputedTransform: Matrix4, modelMatrix: Matrix4, computedModelMatrix: Matrix4 }.
+ * @returns {object} The result parameter, populated with the computed transforms.
+ *
+ * @private
+ */
+function computeNodeTransforms(runtimeNode, sceneGraph, model, result) {
```
javagl: I think these functions should be `static` on `ModelReader`.

(These "free-floating functions" are an antipattern. Unfortunately, they are currently even part of the 'Coding Guide' of CesiumJS, but that doesn't mean that they are _not_ an antipattern (see the last point of https://github.com/CesiumGS/cesium/issues/12452). I'd need a function to compute these transforms _right now_, and I could now either refactor the `ModelReader`, or just copy and paste the whole thing...)

javagl (approved):
Some cross-link: This was largely extracted from https://github.com/CesiumGS/cesium/pull/13330 . 

---

I assume that there have not been significant actual _changes_ between that PR and this new PR. And I already did review most of the changes in the context of that other PR. This included, for example, running tests that check whether the combination of the refactored 'pickModel' and 'verticalExaggeration' still work, and they did. 

(An aside: Maybe, when there's a chance, and I "have time", I'll also check in how far the change here affects https://github.com/CesiumGS/cesium/issues/11814 . One of the main issues there - namely, that the vertices are transformed many times, unnecessarily - can _not_ be changed here without introducing some 'even more special' handling for vertical exaggeration. But my gut feeling is that even reading the raw positions only once could already have a positive effect)

---

I now also had a _slightly_ closer look at the `ModelReaderSpec`. 

Again, to provide some context: The `ModelReader` was something that I originally drafted, because 1. I needed something like that and 2. I did _not_ want to "smear" a half-baked version of that attribute-reading functionality into the relevant code paths _again_. Instead, I tried to create one central place for that. It then still was only a skeleton, and contains some "disclaimers". Of course, it was implemented in a hurry, with the usual tension between "having to get a feature done" and "idealistically trying to create something that is _useful_ (beyond a certain feature)". The specs until now had therefore only been ~"integration level" tests: I created a bunch of [unitSquare](https://github.com/CesiumGS/cesium/blob/f1830cbb60da6cf9f8401cb49a40971a8fb54d61/Specs/Data/Models/glTF-2.0/unitSquare/README.md) flavors, read all of them, and checked whether they represent the same geometry. That does (of course) not cover many of the combinations that may appear in theory.

The new specs go waaay further: They programmatically create mock data structures with many of these combinations. (Of course, I did not read them 'line my line', but from scrolling over them, it looks like they already cover a lot). I'm often a bit hesitant when it comes to certain forms of 'mocking'. I'm pretty sure that in some other specs, the specs are not really testing anything in the codebase, but instead, are mainly testing whether _something was mocked properly_. But that doesn't seem to be the case here.

---

I know that the `InstancingPipelineStage` (as part of the model loading and 'processing' infrastructure) contains some black magic and faaaar to little documentation. That single change related to `keepTypedArray` looks strange, but ... the description sounds... reasonable.

---

There could be a bit further nitpicking - for example, people could suggest to replace [some instance creations with 'scratch' variables and fill them with `Class.unpack(array i * N, scratch)`](https://github.com/CesiumGS/cesium/pull/13433/changes#diff-3f7c865e85085d8a0cf5fa39dc40546fff7ef63c1766f5e340f7ad44cfc35093R1245). But compared to the overall improvements and additional functionality here, that's a detail.

alarkbentley:
@mzschwartz5 @javagl Any chance we get get this merged before the 1.141 cut-off? (which I assume is this week)

javagl:
Some of the functionality of the `ModelReader` _might_ be offered in a similar (or even more generic) form via https://github.com/CesiumGS/cesium/pull/13400 . But regardless of that, the changes here are a clear improvement:

- Refactor that dreaded 'pickModel' 
- Offer new functions in `ModelReader` that are clear, self-contained and _useful_ building blocks (no complex class structures or APIs, just convenient _functions_ that can just be called)
- Greatly improved test coverage for the latter

Given that `ModelReader` is only internal API, slight changes (e.g. when the methods are "re-routed" to some newly introduced functionality in the future) would not be 'breaking'. So I don't see a reason to not merge it.

PR #13441, opened 2026-04-27
Fix comment spacing

Alphabket:


donmccurdy:
Hi @Alphabket — We're seeing a lot of automated PRs recently, with unclear authorship and intent, or missing PR templates. We do not have the team bandwidth to provide feedback on automated PRs, and will need to close this PR accordingly. Apologies if we've miscategorized your work; if that's the case please let us know.

PR #13443, opened 2026-04-27
Types: Convert Cesium3DTileContent to ES6 classes

donmccurdy:
# Description

A bit of preliminary cleanup before attempting a fix on #11683. Converts all Cesium3DTileContent classes to ES6 classes. 99% automated changes with lebab and `sg scan`. I manually fixed a few small JSDoc / types issues as well, but didn't turn on `// @ts-check` yet. That was turning into a bigger change because of dependencies, so left for another time.

As with other similar ES6 PRs, viewing the diff in "split" mode with whitespace changes hidden works best to simplify review.

## Issue number and link

 - #4434

## Testing plan

See unit tests, and `sg scan` results. No functional changes intended.

<details>
<summary>Diff of Cesium.d.ts</summary>

```diff
diff --git a/Source/Cesium.main.d.ts b/Source/Cesium.d.ts
index d9818ad921..b70d113da6 100644
--- a/Source/Cesium.main.d.ts
+++ b/Source/Cesium.d.ts
@@ -9571,7 +9571,7 @@ export interface MapProjection {
     /**
      * Gets the {@link Ellipsoid}.
      */
-    ellipsoid: Ellipsoid;
+    readonly ellipsoid: Ellipsoid;
     /**
      * Projects {@link Cartographic} coordinates, in radians, to projection-specific map coordinates, in meters.
      * @param cartographic - The coordinates to project.
@@ -29325,8 +29325,7 @@ export enum Cesium3DTileColorBlendMode {
  * This type describes an interface and is not intended to be instantiated directly.
  * </p>
  */
-export class Cesium3DTileContent {
-    constructor();
+export interface Cesium3DTileContent {
     /**
      * Gets the number of features in the tile.
      */
@@ -33865,14 +33864,16 @@ export class FrameRateMonitor {
     destroy(): void;
 }
 
+export interface GaussianSplat3DTileContent extends Cesium3DTileContent {
+}
+
 /**
  * Represents the contents of a glTF or glb using the {@link https://github.com/CesiumGS/glTF/tree/draft-splat-spz/extensions/2.0/Khronos/KHR_gaussian_splatting | KHR_gaussian_splatting} and {@link https://github.com/CesiumGS/glTF/tree/draft-splat-spz/extensions/2.0/Khronos/KHR_gaussian_splatting_compression_spz_2 | KHR_gaussian_splatting_compression_spz_2} extensions.
  * <p>
  * Implements the {@link Cesium3DTileContent} interface.
  * </p>
  */
-export class GaussianSplat3DTileContent {
-    constructor();
+export class GaussianSplat3DTileContent implements Cesium3DTileContent {
     /**
      * Performs checks to ensure that the provided tileset has the Gaussian Splatting extensions.
      * @param tileset - The tileset to check for the extensions.
@@ -33937,7 +33938,7 @@ export class GaussianSplat3DTileContent {
      * The resource that this content was loaded from.
      * <p>
      */
-    readonly url: Resource;
+    readonly url: string;
     /**
      * Returns whether the feature has this property.
      * @param batchId - The batchId for the feature.

```
</details>

~~... mildly unfortunate that the `readonly` annotation isn't preserved in the generated types, but I think the diff is acceptable for now.~~

UPDATE: Fixed the missing 'readonly' annotations, we need to use `@readonly` and `@constant` _both_ when describing readonly class fields: one for TypeScript, the other for tsd-jsdoc. Added an ast-grep rule to catch that in the future, so if one of the tags is added to an ES6 class field and the other is missing, the lint will fail.

tsd-jsdoc's behavior is a bit surprising to me, but produces the output we need and can be verified in static analysis, so I think that's OK. Related: https://github.com/englercj/tsd-jsdoc/issues/136

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13443** 👈


packages/engine/Source/Scene/GaussianSplat3DTileContent.js
```diff
@@ -9,321 +9,292 @@ import ModelUtility from "./Model/ModelUtility.js";
...
+  get tile() {
+    return this._tile;
+  }
+
   /**
    * The resource that this content was loaded from.
    * <p>
-   * @memberof GaussianSplat3DTileContent.prototype
-   * @type {Resource}
+   * @type {string}
```
donmccurdy, line 291: Minor typo fix - after adding `@implements Cesium3DTileContent` above, this parameter caused `npm run build-ts` to fail because the types didn't match the parent interface.

packages/engine/Specs/Scene/Cesium3DTileContentSpec.js
donmccurdy, line 1: To make Cesium3DTileContent a type-checked interface I've replaced its getters/setters with equivalents of:

```javascript
/** @interface */
class MyClass {
  /** @type {string} */
  someProperty;
}
```

Ultimately that's a win because we get type analysis to check that implementations match the interface correctly, but it does mean this test doesn't do much anymore. Removed the test.

javagl: I wasn't even aware of that spec. 

But I've recently stumbled over the "opposite", and wondered (also in another context) whether there shouldn't be some check for _all_ so-called "interfaces" to see whether they are implemented correctly. Specifically, as far as I know (and unless that has recently changed), you could just say "Implements the `Cesium3DTileContent` interface", and simply _omit_ some function (and nobody would notice it, at any point in time, unless that function happened to be called).

I wonder if there could/should be some specs to check for implementations, with pseudocode tlike
``` 

// Check if one function/property is there...
checkPresent(implementor, name) {
   expect(function () {
      return implementor[name];
    }).toDoWhateverButAtLeastNotThrow();
};

// Check if all functions/properties of an interface are there...
checkCesium3DTIleContent(implementor) {
  checkPresent(implementor, "featuresLength");
  checkPresent(implementor, "geometryByteLength");
  ...
};

// Check all implementors
checkImplementorsOfCesium3DTIleContent() {
   checkCesium3DTIleContent(Multiple3DTileContent.class);
   checkCesium3DTIleContent(Composite3DTileContent.class);
   ...
}
```

**But**... ideally, something like that will be covered by the compiler in the future 🤞

donmccurdy: If we put `@interface` on the interface and `@implements SomeInterface` on the implementation, then `npm run tsc` will catch missing implementation methods in static analysis, for files with type checking enabled (`// @ts-check`). I haven't enabled `// @ts-check` on most files in this PR as it makes the PR much larger than it already is, but that would be the natural next step. 

In this case `npm run build-ts` happened to catch some interface/implementation mismatches as well, but that's partially luck, I don't think it has the coverage/guarantees of `npm run tsc`.

javagl: I just realized that we **do** have [`toConformToInterface`](https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/Specs/addDefaultMatchers.js#L274) already. It doesn't seem to be used for `Cesium3DTileContent`, though...

donmccurdy:
@javagl not time-sensitive, but would you mind doing a review on this when time allows?

> ~~**EDIT:** Sorry, maybe better to wait on reviewing — I think we can fix the missing 'readonly' annotations in the Cesium.d.ts diff.~~

javagl:
I haven't looked at the details, but will have a close(r) look at the role of `readonly` for things like `ready` (i.e. make sure that no implementation uses an actual property for that)

donmccurdy:
Ok, I think the 'readonly' issue is resolved now! Added an ast-grep rule, with a longer explanation, to avoid that in the future.

packages/engine/Source/Scene/Cesium3DTileContent.js
```diff
@@ -38,7 +38,8 @@ class Cesium3DTileContent {
      *
      * @type {boolean}
      *
-     * @private
+     * @protected
```
javagl, line 41: Was changing this from `@private` to `@protected/ignore` necessary, or just a drive-by fix? (I'm wondering about consistency with `batchTable`, for example).

javagl (approved):
Nothing that should prevent this from being merged. The one inlined comment is rather a question. 

Asides:

The goal here is a purely technical migration, and that's what it does. I'll skip some of my thoughts about the `Cesium3DTileContent` interface (and possible restructurings of that). 

That 
`export interface GaussianSplat3DTileContent ...`
`export class GaussianSplat3DTileContent ...`
in the _same file_ (!) caught my attention. But ... I'm learning something new every day (in this case, that the TypeScript people think that something like this makes sense).

However, the class should not be public to begin with, and I assume that the missing `@private` in the previous state was an oversight. I'll count this as a "documentation fix" now, and added a comment at https://github.com/CesiumGS/cesium/issues/11749#issuecomment-4409536859

javagl:
I'm curious about the updates that will be necessary for the new content type in https://github.com/CesiumGS/cesium/pull/12977 , but don't see any specific points that may be critical (in that PR, the new type already is a `class` anyhow). There's only one way to find out - merging 🙂

packages/engine/Source/Scene/Cesium3DTileContent.js
donmccurdy: ~~Necessary, but only preemptively.~~ I believe I'd run into a type error requiring this change (as required changes like this have been very common elsewhere) but in this case I don't recall exactly.  I started fixing a few type checking errors while making this PR, before deciding that fixing all of them would clutter the PR too much. So I left the fixes that I'd already completed, and opted the files back out of type checking to silence the remaining errors.

All of the other Cesium3DTileContent classes access or re-define this property, and `Cesium3DTileContent` is an _interface_ after all ... so ultimately nothing in the interface should be "private". But now that you mention it, "protected" doesn't make so much sense for an interface either, as nothing extends the class. Perhaps just `@ignore` if we don't want something documented.

PR #13457, opened 2026-04-30
More robust playwright tests

jjspace:
# Description

While prepping for the release and running the e2e tests I got nerd sniped again by wishing they were more robust and reproducible. This PR includes a handful of changes, mostly to the sandcastle specs, that made the snapshots much more consistent on my machine. This is a slight follow up to https://github.com/CesiumGS/cesium/pull/13381

- swapped to `--use-gl=vulkan` as I believe most, if not all, of the team said this works cross platform and was annoying to continually do on linux where it was actually needed
- Updated localstorage to hide the default cesium viewer controls so snapshots actually highlight the test itself
- Added waiting for the sandcastle viewer page to be loaded before starting a spec
- added special loading logic for sandcastles we know are slow to load or have fast animations
- Fixed a handful of sandcastles themselves for errors or warnings in the console

These changes are still not _perfect_ but in my testing only a small handful of specs actually fail or need re-run with these changes. So much so that I'm planning to open another PR right after this to actually commit snapshots to the repo as a reference point. No more "generate against the previous release" during the release! :tada: 

## Issue number and link

- Improves https://github.com/CesiumGS/cesium/pull/13381 but may not fully solve it
- Helps get closer to https://github.com/CesiumGS/cesium/issues/12355

## Testing plan

- Run `npm run test-e2e -- --ui` and test running a number of specs of different types
- Test the list of sandcastles with special handling in the specs file
- Tests should consistently complete/pass and have the same snapshot (generated the first time you run if you didn't have one)

## Author checklist

## AI acknowledgment


Specs/e2e/models.spec.js
```diff
@@ -78,6 +78,12 @@ test("loads draco model", async ({ cesiumPage }) => {
 
     viewer.trackedEntity = entity;
   });
+  await cesiumPage.page.waitForLoadState("networkidle");
+  await cesiumPage.page.clock.pauseAt(new Date("2023-12-25T14:00:00"));
+
+  await cesiumPage.page.clock.runFor(1000);
+  await cesiumPage.page.clock.runFor(1000);
+  await cesiumPage.page.clock.runFor(1000);
```
donmccurdy, line 86: Is this different from `runFor(3000)`?

Specs/e2e/sandcastle.spec.js
```diff
@@ -4,29 +4,90 @@ import { basename, dirname } from "node:path";
...
+  "custom-primitive-dev",
+  "czml-path",
+  "manually-controlled-animation",
+  "material-with-custom-glsl",
+  "relative-paths",
+  "moon", // because it spins by default
+  "multi-part-czml",
+  "particle-system",
+  "time-dynamic-wheels",
+  "3d-tiles-gaussian-splatting", // not animated but needs the extra time for splat loading
```
donmccurdy, line 27: Sandcastle "3d-tiles-gaussian-splatting" is included in both knownSlow and knownAnimated, and it looks like knownAnimated takes precedence. Is that intended or should it be removed from one or the other? 

Not necessary in this PR, but maybe we'd eventually want this tracked in `sandcastle.yaml`?

jjspace:
I noticed during the release that the sandcastle specs don't ignore development only sandcastles which may not exist. This can happen during release since `npm run make-zip` will exclude them. Probably need a flag for that or maybe a check to conditionally skip dev sandcastles if it detects it can't load. something to consider

donmccurdy (commented):
After running `npm run build-sandcastle` I tried:

```bash
npm run test-e2e -- -g "3d-tiles-1.1-cdb-yemen|3d-tiles-gaussian-splatting|3d-tiles-gaussian-splats-with-lod|callback-position-property|callback-property|custom-primitive-dev|czml-path|manually-controlled-animation|material-with-custom-glsl|relative-paths|moon|multi-part-czml|particle-system|time-dynamic-wheels"
```

Result:

Image

Of the failures, about half exceeded the 100,000ms timeout. The other half had small (3–12%) pixel differences. I don't necessarily feel that any of this should block the PR if it's an improvement over current status, but wasn't sure if it was expected.

I also noticed that workers are disabled in playwright.config.js without. My results on these tests were similar (but faster) with workers turned back on — do we know of any reason that couldn't be on by default?

Specs/e2e/models.spec.js
ggetz: The `await` should allow any completed promises to resolve. I believe simply sleeping the thread does not do so.

Specs/e2e/sandcastle.spec.js
```diff
@@ -4,29 +4,90 @@ import { basename, dirname } from "node:path";
...
 
   test(`${slug} renders`, async ({ page }) => {
-    test.setTimeout(100000);
+    // the overall timeout for this test in the rare case it gets stuck waiting for network or screenshots
+    test.setTimeout(100_000);
+    page.context().setStorageState({
+      origins: [
+        {
+          origin: "http://localhost:8081",
+          localStorage: [{ name: "cesium-hasSeenNavHelp", value: "true" }],
```
ggetz, line 41: A comment here would be helpful, IMO.
```suggestion
          // Preset value so that the navigation help widget panel is never displayed
          localStorage: [{ name: "cesium-hasSeenNavHelp", value: "true" }],
```

Specs/e2e/models.spec.js
jjspace: The playwright `runFor` function is a little weird. It claims to "fast forward" any timing based functions. It _seems_ like it should work with the viewer to process data but in my testing it's still a little flaky and doesn't _actually_ advance time. Maybe it would work with a single `3000` instead of multiple, multiple were there before so I just kept it as I was tweaking things.

jjspace:
> Of the failures, about half exceeded the 100,000ms timeout

@donmccurdy As long as the system is set up correctly I wouldn't expect _any_ tests to hit the max timeout. Can you check the [troubleshooting steps from the testing guide](https://github.com/CesiumGS/cesium/blob/main/Documentation/Contributors/TestingGuide/README.md#end-to-end-test-performance) to make sure you're running them with the gpu?

Issue #13444, opened 2026-04-27
BufferPrimitiveCollection Improvements


donmccurdy:
### Feature

BufferPrimitiveCollection (recently introduced in #13156) provides performance-oriented APIs for working with large point/line/polygon datasets efficiently. This issue is intended as a short- to medium-term tracker for potential improvements to BufferPrimitiveCollection and related APIs, until they become stable enough to begin tracking each idea as a separate issue.

Ideas (and PRs) welcome! Unless otherwise indicated, these tasks are currently unassigned.

***

## General

- [ ] @danielzhong and @donmccurdy Support clamped height references  (draping on terrain and 3D tiles) for polyline and polygon collections, via GPU Lookup rendering.
    - #13325 
    - https://github.com/CesiumGS/cesium/issues/13584
- [x] @danielzhong Support loading Mapbox Vector Tiles (MVT) tilesets, by transcoding to 3D tiles at runtime.
    - [x] Geometry: #13404 
    - [x] Properties: https://github.com/CesiumGS/cesium/pull/13508
- [x] @danielzhong Support loading GeoJSON to BufferPrimitiveCollection. Result should be a set of point/line/polygon collections, not a tileset.
    - #13505
- [ ] Support rendering in 2D view. May require some considerable attention to limit duplication of code.
    - https://github.com/CesiumGS/cesium/issues/13585
- [x] @danielzhong Support a `positionNormalized` option on BufferPrimitiveCollection, allowing (u)int8 and (u)int16 attributes to be normalized. Pre-requisite to fully supporting meshopt compression on 3D Tiles. See @donmccurdy for sample data.
    - #13462
- [x] @donmccurdy Allow changing `collection.modelMatrix` after initialization
    - https://github.com/CesiumGS/cesium/pull/13448

## Styling

- [ ] BufferPolygonMaterial should support `outlineWidth` and `outlineColor` properties. In renderer, drawing outlines for _polygons_ should ideally mean reusing the _polyline_ renderer, with an additional gl.polygonOffset to prevent z-fighting.
    - #13552
    - #13587
- [ ] Support clipping vector collections to tile bounding volumes. This must be done in the fragment shader, so that wide lines are visible within tile extents, even if the underlying geometry has passed slightly outside the tile. See https://github.com/CesiumGS/3d-tiles/pull/838.
    - #13586
[...]

PR #13453, opened 2026-04-30
Types: Fix type checks against remaining enums

donmccurdy:
# Description

Continuation of #13429, with a partial fix for the remaining enums missing type checking support. These enums have utility functions attached to the enum object, and so we can't (yet) enable `// @ts-check` on their source files. We can, however, fix the `export default Object.freeze(MyEnum)` issue preventing files _depending on_ these enums from running type checks against their values.

NOTE: After this PR, type-checked files depending on these enums will need to use `// @ts-expect-error` when calling the utility functions attached to these enums, but not for passing and comparing against the enum values themselves. Steps toward the longer-term support, see #13420.t

## Issue number and link

- #13420
- #4434

## Testing plan

- npm run tsc
- npm run sg-scan
- npm run build-ts

Output of ./scripts/diff-types.sh:

\<empty>

## Author checklist

- [x] I have submitted a Contributor License Agreement
- [x] I have added my name to `CONTRIBUTORS.md`
- [ ] ~~I have updated `CHANGES.md` with a short summary of my change~~
- [ ] ~~I have added or updated unit tests to ensure consistent code coverage~~
- [x] I have updated the inline documentation, and included code examples where relevant
- [x] I have performed a self-review of my code

## AI acknowledgment

- [ ] I used AI to generate content in this PR
- [ ] If yes, I have reviewed the AI-generated content before submitting

If yes, I used the following Tools(s) and/or Service(s):

If yes, I used the following Model(s):

#### PR Dependency Tree

* **PR #13453** 👈

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

Issue #13420, opened 2026-04-20
Types: Support type-checking for enums and related functions


donmccurdy:
### Feature

A pattern currently used widely in the CesiumJS codebase...

```javascript
/**
 * @enum {number}
 */
const MyInteger = {
  ONE: 1,
  TWO: 2,
  THREE: 3,
  ...
}

/** @private */
MyInteger.validate = function(value) {
  ...
};

export default Object.freeze(MyInteger);
```

... is not fully compatible with either type checking or JSDoc documentation for three reasons:

1. TypeScript does not support enums having attached functions (type checks fail)
2. JSDoc does not support enums having attached functions (documentation and generated types are missing for the functions)
3. TypeScript does not understand that the return type of `Object.freeze(MyInteger)` is still the original enum, now readonly

As discussed at greater length in #13203, we plan to fix type checking and documentation by moving to the following pattern:

```javascript
/**
 * @enum {number}
 */
const MyInteger = {
  ONE: 1,
  TWO: 2,
  THREE: 3,
  ...
}

/**
 * @param {MyInteger} value
 * @returns {boolean}
 * @private
 */
export function validateMyInteger(value) {
  ...
}

Object.freeze(MyInteger);

export default MyInteger;
```

This pattern works if the utility functions are internal/private, as many are. Public utility functions would need to be moved to a separate file, or the build system would need to be modified to re-export them from the package.

We'll likely do this migration in a couple steps. Any enums without utility functions can be updated in bulk, with no breaking changes. Utility functions — if public — may require a deprecation period to detach from their enum objects, and some discussion/decisions in terms of naming.

Related:

- https://github.com/CesiumGS/cesium/pull/13203
- https://github.com/CesiumGS/cesium/issues/4434
- https://github.com/CesiumGS/cesium/issues/10455#issuecomment-1238642074

### AI acknowledgment

- [ ] I used AI to generate this feature request.
- [ ] (If the above is checked) I have reviewed the AI-generated content before submitting.

donmccurdy:
In #13462, @danielzhong has come up with another possible solution to the broken type checking on enum utility functions. As I understand it, the solution would involve creating a new `*.d.ts` file for each enum, and (in that file) re-declaring the enum, and also declaring a [namespace](https://www.typescriptlang.org/docs/handbook/namespaces.html) with the same name as the enum, containing function signatures for each enum utility function. Something like...

```typescript
// ComponentDatatype.d.ts

[...]

PR #13462, opened 2026-05-01
Support positionNormalized on BufferPrimitiveCollection for meshopt compression

danielzhong:
# Description
Adds `positionNormalized` and `positionDatatype` options to `BufferPrimitiveCollection` (and subclasses) so that integer position attributes (int8/uint8/int16/uint16) can remain quantized in memory rather than being dequantized upfront. When enabled, positions are de-normalized and transformed to world-space ECEF on the CPU before RTE encoding (to preserve float64 precision), and the DrawCommand uses `Matrix4.IDENTITY` to avoid a double-transform. Also updates `createVectorTileBuffersFromModelComponents` to automatically detect `positionAttribute.normalized` on glTF primitives and take the raw int16 path.

## Issue number and link
https://github.com/CesiumGS/cesium/issues/13444

## Testing plan
[highway-roads-meshopt.zip](https://github.com/user-attachments/files/27289536/highway-roads-meshopt.zip)

## Author checklist

## AI acknowledgment

Github Copilot
Claude 4.6 sonnet (encode GLB)


packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -333,16 +335,31 @@ function appendNodeToBuffers(content, node, parentTransform, result) {
...
+    const positionAttribute = ModelUtility.getAttributeBySemantic(
+      primitive,
+      VertexAttributeSemantic.POSITION,
+    );
+    const positionNormalized =
+      (positionAttribute.normalized ?? false) &&
+      !defined(positionAttribute.quantization);
+    const positionDatatype = positionNormalized
+      ? positionAttribute.componentDatatype
+      : undefined;
```
donmccurdy: I think we can use `positionAttribute.componentDatatype` unconditionally here - it's rare that a glTF file will contain float64 vertices, which is the default for BufferPrimitiveCollection, so we can save a bit of space even when the data isn't quantized by using float32 or whatever the glTF provides.

(It's also possible for meshopt quantization to be used without normalization)

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
```diff
@@ -333,16 +335,31 @@ function appendNodeToBuffers(content, node, parentTransform, result) {
 
     const stats = gatherPrimitiveStats(primitive);
 
+    const positionAttribute = ModelUtility.getAttributeBySemantic(
+      primitive,
+      VertexAttributeSemantic.POSITION,
+    );
+    const positionNormalized =
+      (positionAttribute.normalized ?? false) &&
+      !defined(positionAttribute.quantization);
```
donmccurdy: Does this work if we just use `positionAttribute.normalized` directly? I'm not entirely sure how the `attribute.quantization` type is set in CesiumJS (it isn't a glTF concept exactly), but I believe normalization shouldn't require quantization.

packages/engine/Source/Scene/BufferPointCollection.js
```diff
@@ -56,6 +57,8 @@ class BufferPointCollection extends BufferPrimitiveCollection {
   /**
    * @param {object} options
    * @param {number} [options.primitiveCountMax=BufferPrimitiveCollection.DEFAULT_CAPACITY]
+   * @param {ComponentDatatype} [options.positionDatatype=ComponentDatatype.DOUBLE]
```
donmccurdy, line 61: Oops good catch, I'd just fixed this in #13465 as well. 😅

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -471,16 +552,31 @@ class BufferPrimitiveCollection {
...
+      // De-normalize integer position values to [-1, 1] or [0, 1] before
+      // computing the bounding sphere. The model matrix will then transform
+      // from that local space to world coordinates.
+      const count = this._positionCount * 3;
+      const divisor = this._getNormalizationDivisor();
+      const rawView = this._positionView;
+      vertices = new Float32Array(count);
+      for (let i = 0; i < count; i++) {
+        vertices[i] = rawView[i] / divisor;
+      }
```
donmccurdy: I think we can use the existing `AttributeCompression.dequantize` method here — I wish we didn't need to copy the buffer, but I don't see an easy way around it for now, so that's OK!

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -277,6 +295,69 @@ class BufferPrimitiveCollection {
...
+  /**
+   * Returns the divisor used to de-normalize integer position values when
+   * {@link positionNormalized} is <code>true</code>.
+   *
+   * @returns {number}
+   * @ignore
+   */
+  _getNormalizationDivisor() {
+    // Cast through unknown→number: ComponentDatatype values are numbers, but JSDoc types
+    // ComponentDatatype as the whole enum object, causing TS switch-case errors.
```
donmccurdy: I think this has been fixed with https://github.com/CesiumGS/cesium/pull/13453. :)

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -277,6 +295,69 @@ class BufferPrimitiveCollection {
...
+  }
+
+  /**
+   * Returns the divisor used to de-normalize integer position values when
+   * {@link positionNormalized} is <code>true</code>.
+   *
+   * @returns {number}
+   * @ignore
+   */
+  _getNormalizationDivisor() {
```
donmccurdy: Very surprised this didn't exist already! I might suggest making this a helper function on the `ComponentDatatype` enum, something like:

```javascript
const floatValue = ComponentDatatype.dequantize(intValue, componentDatatype);
```

We might be able to use it elsewhere, but perhaps not in `AttributeCompression.dequantize` quite yet... we'd want to check that performance doesn't regress compared to looping over typed arrays without function calls like the method does now.

Also note that we'll need a `Math.max(v, -1)` clamp for the signed types, to match the WebGL and glTF behaviors, see:

https://github.com/donmccurdy/glTF-Transform/blob/3378a5d3085a98900d382e37c6b5610b9fff3416/packages/core/src/utils/math-utils.ts#L29-L45

packages/engine/Source/Scene/BufferPrimitiveCollection.js
```diff
@@ -277,6 +295,69 @@ class BufferPrimitiveCollection {
...
+   */
+  _denormalizePosition(cartesian) {
+    if (!this._positionNormalized) {
+      return cartesian;
+    }
+    const d = this._getNormalizationDivisor();
+    cartesian.x /= d;
+    cartesian.y /= d;
+    cartesian.z /= d;
+    return Matrix4.multiplyByPoint(this.modelMatrix, cartesian, cartesian);
```
donmccurdy: I'm not sure I understand why we'd apply the model matrix during denormalization – do we need to do these at the same time?

In the future, I'd like to update `renderBuffer*Collection` to render _without_ converting positions to float64, e.g. keeping the vertex attributes compressed/quantized on GPU. See the "Performance" section on https://github.com/CesiumGS/cesium/issues/13444. But that doesn't need to be part of this PR, it's enough of an improvement just to be keeping the data quantized on the CPU first.

donmccurdy (commented):
Thanks @danielzhong! Great to be able to keep larger datasets quantized on the CPU.

GitHub is having a bad day (https://www.githubstatus.com/), sorry if my comments are duplicated etc... it was very buggy trying to post these PR comments for some reason.

packages/engine/Source/Scene/BufferPrimitiveCollection.js
javagl: > we'd want to check that performance doesn't regress compared to looping over typed arrays without function calls 

The function call is one thing (and I really hope that this wouldn't matter, otherwise, I'd consider this as a case of bad JITting). What would be more concerning for me is the `switch (componentDataType)` that would be hidden in that function (unless this `dequantize` can be snuck into the enum value objects, so that it doesn't require that parameter to `switch` on any more). Sure, branch prediction should take care of that, but at this point, it would really be worth benchmarking.

Aside: Even if it might not be relevant _here_, any generic implementation of that `getNormalizationDivisor` function should probably cover all types, including `(UNSIGNED_)INT`, as of https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/packages/engine/Source/Core/AttributeCompression.js#L438
 , to avoid regressions.

packages/engine/Source/Scene/Model/createVectorTileBuffersFromModelComponents.js
danielzhong: Thanks! After looking at the code, `attribute.quantization` is only set for Draco compressed attributes (which decode to FLOAT and won't have normalized=true) and KHR_mesh_quantization (where accessor.normalized is typically false). So in practice these two won't occur together on a POSITION attribute. Updated!

donmccurdy:
Wasn't sure if this PR was ready for another round of review, but just two points for what I think is left:

- I don't think the `collection.modelMatrix` handling should need to be any different before/after this PR, because the matrix represents a transform inherited from a glTF node, which could be present with or without normalization. If that doesn't seem to be working as expected, let's chat!
- Your solution for type errors on `ComponentDatatype.dequantize`, by adding `ComponentDatatype.d.ts` with a namespace, is new to me! I don't know if I 100% understand how it works yet, but this could definitely be a help for #13420. There's been a bit of  back-and-forth about how to solve type checking with enums and I'd rather this PR not get blocked on discussion, so for now I would suggest:
  1. I'll leave a comment on #13420 about your solution, and we can discuss there whether that's something we want to do for enums more generally.
  2. In the meantime, I'd recommend removing `ComponentDatatype.d.ts` from this PR, and instead adding a comment like this above any type errors:
  `// @ts-expect-error https://github.com/CesiumGS/cesium/issues/13420`


Other than those two things, I believe this will be ready to merge, thanks! 🙏

danielzhong:
@donmccurdy Merged GPU normalized branch, ready to review again!

packages/engine/Source/Core/ComponentDatatype.d.ts
```diff
@@ -23,6 +23,18 @@ declare namespace ComponentDatatype {
...
+    valuesOrLength: number | ArrayLike<number>,
+  ):
+    | Int8Array
+    | Uint8Array
+    | Int16Array
+    | Uint16Array
+    | Int32Array
+    | Uint32Array
+    | Float32Array
+    | Float64Array;
```
javagl: Just curious, mainly @donmccurdy : **Iff** further functions are moved into dedicated `d.ts` files, could it make sense to pull out some sort of `globalTypes.d.ts`, equivalent to https://github.com/CesiumGS/cesium/blob/24c112988a734059a2ebc47786aa44d47d1c1b19/packages/engine/Source/Core/globalTypes.js#L3 , to not have to repeat these `|`-chains?

(I'm pretty sure that you [won't say 'No!'](https://github.com/donmccurdy/glTF-Transform/blob/3378a5d3085a98900d382e37c6b5610b9fff3416/packages/core/src/constants.ts#L81) 🙂 but maybe it will be something like "Not yet, given that we are using this only in a single place for now"...)

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -43,10 +43,24 @@ const BufferPointAttributeLocations = {
...
+ * Material attributes retain their original indices.
+ * @type {Record<string, number>}
+ * @ignore
+ */
+const BufferPointLocalSpaceAttributeLocations = {
+  position: 0,
+  pickColor: 2,
+  showSizeAndColor: 3,
+  outlineWidthAndOutlineColor: 4,
+};
```
donmccurdy: These indices don't seem to be working for me, maybe the locations need to be increments of 1? I wasn't sure at a glance, but quickly tested on [this sandcastle](https://github.com/CesiumGS/cesium/pull/13465) using Float32Array and it appears to require that change. Did not test for polylines and polygons yet.

```suggestion
const BufferPointLocalSpaceAttributeLocations = {
  position: 0,
  pickColor: 1,
  showSizeAndColor: 2,
  outlineWidthAndOutlineColor: 3,
};
```

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -71,24 +85,30 @@ const encodedCartesian = new EncodedCartesian3();
 function renderBufferPointCollection(collection, frameState, renderContext) {
   const context = frameState.context;
   renderContext = renderContext || { destroy: destroyRenderContext };
+  const useFloat64 = collection._positionDatatype === ComponentDatatype.DOUBLE;
+  const useLocalSpace = !useFloat64;
```
donmccurdy: I think there might still be a small misunderstanding here — contents of the `collection._positionView` buffer are — regardless of the component datatype or normalization — defined to be in local/model space. Normalization and data type are just an over-the-wire encoding, and so I wouldn't think of these as changing whether we're in local space or not.

So I would recommend referencing `useFloat64` everywhere that we need a conditional in this function, and we can remove any references to modelMatrix and local space from the PR.

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -71,24 +85,30 @@ const encodedCartesian = new EncodedCartesian3();
...
+          showSizeAndColor: new Float32Array(featureCountMax * 3),
+          outlineWidthAndOutlineColor: new Float32Array(featureCountMax * 2),
+        }
+      : {
+          positionHigh: new Float32Array(featureCountMax * 3),
+          positionLow: new Float32Array(featureCountMax * 3),
+          pickColor: new Uint8Array(featureCountMax * 4),
+          showSizeAndColor: new Float32Array(featureCountMax * 3),
+          outlineWidthAndOutlineColor: new Float32Array(featureCountMax * 2),
+        };
```
donmccurdy: Minor, optional — I liked this pattern that you used later for the VertexArray, maybe we can do the same here?

```suggestion
    renderContext.attributeArrays = {
      ...(useFloat64
        ? {
            positionHigh: new Float32Array(featureCountMax * 3),
            positionLow: new Float32Array(featureCountMax * 3),
          }
        : { position: collection._positionView }),
      pickColor: new Uint8Array(featureCountMax * 4),
      showSizeAndColor: new Float32Array(featureCountMax * 3),
      outlineWidthAndOutlineColor: new Float32Array(featureCountMax * 2),
    };
```

As a reader I think it's nice to see how the 'position' attribute is being handled explicitly here, too.

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -194,16 +234,38 @@ function renderBufferPointCollection(collection, frameState, renderContext) {
...
+          const attribute = /** @type {BufferPointAttribute} */ (key);
+          renderContext.vertexArray.copyAttributeFromRange(
+            BufferPointAttributeLocations[attribute],
+            renderContext.attributeArrays[attribute],
+            collection._dirtyOffset,
+            collection._dirtyCount,
+          );
+        }
+      }
     }
```
donmccurdy: I think we can simplify this using the new locations declaration. Not sure I got the bracket indentation right in GitHub's 'suggestions' box though, maybe better to do this in a code editor rather than accepting my suggestion here. 😅

```suggestion
    const locations = useFloat64
      ? BufferPointAttributeLocations
      : BufferPointLocalSpaceAttributeLocations;
    for (const key in locations) {
      if (Object.hasOwn(locations, key)) {
        const attribute = /** @type {BufferPointAttribute} */ (key);
        renderContext.vertexArray.copyAttributeFromRange(
          locations[attribute],
          renderContext.attributeArrays[attribute],
          collection._dirtyOffset,
          collection._dirtyCount,
        );
      }
    }
```

donmccurdy (commented):
Latest changes are looking excellent, thanks! Very clean and working well on my local tests after a minor fix (see comment). I'm seeing better performance for larger animated collections as well. 

Couple minor cleanup things, and then I think this is ready!

packages/engine/Source/Core/ComponentDatatype.d.ts
donmccurdy: ~~I see this file is no longer needed in the latest draft of the PR, but as a potential part of our solution for #13420 ... I believe it's all one common "types pool" from TypeScript's perspective, so `.d.ts` files can refer to types from `globalTypes.js` without issues... but I haven't tested this fully yet.~~

**EDIT:** Never mind, `npm run build-ts` did not like my idea. Something about `Import or export declaration in an ambient module declaration cannot reference module through relative module name`. Needs some investigation but mirroring basic types into a `globalTypes.d.ts` file seems reasonable!

packages/engine/Source/Scene/renderBufferPointCollection.js
danielzhong: Thanks for the clarification! I renamed the flag. I had misunderstood a few things earlier, which led to the original naming, but it’s all cleaned up now.

donmccurdy (approved):
Great work here @danielzhong! Changes look good, and are performing well on my local tests. There's much less complexity in the PR than I was worried might be necessary to support _both_ 64-bit and <=32-bit precision — that's an important feature, thanks for the iteration and review.

Issue #12989, opened 2025-10-19
Cyclomatic and cognitive code complexity


javagl:
I recently stumbled over https://github.com/geneseframework/. This is **not** a recommendation or endorsement of this library/framework. I just found it, and tried it out. It is a framework that allows certain forms of static code analysis. One of them is the `complexity` analysis. It analyzes the cyclomatic complexity ("How complex the code is"), and the cognitive complexity ("How hard it is to understand the code"). The first one may be easy to define _somewhat_ objectively. The second one can also be defined somewhat objectively, but it's a bit harder. The [section about 'Cognitive complexity'](https://github.com/geneseframework/complexity?tab=readme-ov-file#7-cognitive-complexity) contains a few details.

I think that these aspects are one _specific_ dimension on one _specific_ level of what constitutes the otherwise broad and abstract concept of "code quality": Complex code is hard to understand and maintain. Reverse engineering existing (complex) code (while implementing new features or fixing bugs) eats up time. It is usually time that cannot be attributed to a "task". There is no explicit time allocated for "reading code" - only for "implementing a feature" or "fixing a bug". But still, this complexity slows the whole development process, and one cannot even say what _exactly_ took so much time. 

I ran the default analysis on the CesiumJS 'engine' package. (This required some tweaks and fixes - a huge JavaScript codebase that has gone through 15 years of development certainly raises some challenges for anything that tries to generate something like an AST...). It generates some browsable report, eventually.

The summary diagram:

<img width="1392" height="654" alt="Image" src="https://github.com/user-attachments/assets/8f29c4df-eb28-4867-b405-e27f0b206858" />

The report offers some interesting insights and browsing functionality. For example, a table with the methods ordered by their cognitive complexity:

<img width="1114" height="694" alt="Image" src="https://github.com/user-attachments/assets/427c8c52-3bc6-47c0-8fa8-392f1d743db9" />

[...]

Issue #12163, opened 2024-08-28
About scratch variables and API consistency


javagl:
The concept of "scratch" variables and the pattern of
`result = SomeClass.doSomething(input, scratchResult);`
with the optional result parameter is important for performance-critical applications. It offers the functionality of
`SomeClass.doSomething(input, result);`
and
`result = SomeClass.doSomething(input);`
in a single function, optionally(!) leaving the aspect of memory management to the developer.
(Slightly more complex than each of the simpler ones, but the "best of both worlds")

But some of the API and corresponding documentation should be more strict and more consistent. Right now, whenever I call a function, I do not know ...

1. whether the `result` parameter is optional
2. whether the return value will be the `result` parameter

---

Point 1. can usually be looked up.

Regarding the consistency, one could notice:
`Cartesian3.clone(cartesian, result)`
`Cartesian3.negate(cartesian, result)`
For clone, the result is optional. For negate, the result is not optional. 

**Why?**

I think that the answer is: The result is optional for functions whose sole purpose is to create instances, like `clone` or `from....`. If this is the case, it could/should be pointed out in the coding guide. And one should keep an eye out for places where this rule is not followed. I think that there _are_ a few places where this is currently not the case. 

One could make a case to always make it optional, so that one can write `result = Cartesian3.negate(input);`. But I see the risk of "accidentally" creating instances, particularly by users of the API who may not be aware of the implications of this call.

One general thing that should be kept in mind:
- Changing a function to make the `result` **non**-optional would be a breaking change
- Changing a function to _make_ it optional should be safe. 

---

Point 2. is more subtle.

The Coding Guide at https://github.com/CesiumGS/cesium/tree/main/Documentation/Contributors/CodingGuide#result-parameters-and-scratch-variables says:

> Because result parameters aren't always required or returned, don't strictly rely on the result parameter you passed in to be modified. For example:
> `Cartesian3.add(v0, v1, result);`
> `Cartesian3.add(result, v2, result);`
> is better written as
> `result = Cartesian3.add(v0, v1, result);`
> `result = Cartesian3.add(result, v2, result);`

Now, this does raise one crucial question:
Under which conditions is the result parameter **not** returned?

[...]

Issue #1477, opened 2014-02-18
Cesium 'Invalid array length' error in updateFrustums


orlade:
We're having a complex issue with Cesium rendering. It only seems to occur in some applications on some machines with some versions, and we haven't yet found the common factor.

We use Cesium b25 with a custom render loop which gives the trace, but the same error occurs with the default render loop (it's just caught and logged differently). Basically, the error is the following:

```
Uncaught RangeError: Invalid array length       Scene.js:454
updateFrustums                                  Scene.js:454
createPotentiallyVisibleSet                     Scene.js:609
Scene.render                                    Scene.js:911
CesiumWidget.render                             CesiumWidget.js:539
Viewer.render                                   Viewer.js:728
(anonymous function)
```

I believe the issue doesn't occur with b19, and the cause seems to be that the `horizonDistance` is MAX so the `Occluder` test returns false and the test in `updateFrustum` falls through to `continue` instead of setting the `near` value, thus `near` ends up equal to `far` and `numFrustums` is 0, which is illegal at some point. The render loop runs happily in the background, but fails as soon as we try to actually show the Cesium widget.

I can give you some more information if it helps, but I haven't been working on this part and its growing like a weed with build systems and submodules and the like. Do you have any pointers that we might investigate?

PR #13463, opened 2026-05-02
Fix useless assignments

javagl:
# Description

There have been many places where the [`no-useless-assignment`](https://eslint.org/docs/latest/rules/no-useless-assignment) eslint rule was triggered. There is some infrastructure revolving around 'eslint-seatbelt' that just hides these cases for now. But I think that it can make (more) sense to actually clean that up.

`<!-- Consider: Why is this change required? What problem does it solve? -->`

Build hassle (mainly on Windows right now), reducing the complexity of the infrastructure (seatbelt files, dependencies...), and general code hygiene.

### Summary

The first few commit adds some auto-generated files to the ignore list.

The second commit is one example that is not necessarily a trivial, low-level assignment issue. But... the options object was unused, undocumented, and unnecessary, and likely an artifact from the 'readyPromise/async' changes of 1.107.

The next (enumerated) commits cover the trivial cases. They are usually caused by unnecessary initializations - i.e. patterns like
```
let variable = 0;
if (condition) {
  variable = 1;
} else {
  variable = 2;
}
```
where the initialization with `= 0` is simply not necessary. These changes have been done plainly mechanically. I didn't read the code beyond what was necessary in order to fix the issue. From what I've seen, these cases have usually been "embedded in (otherwise) bad style". It often looked like variables simply had been declared with a scope that was larger than necessary, and could have been converted to 'const' in a narrow scope. But we can only try to incrementally improve things here.

### Open questions

Now, why does it make sense to **not** ignore linting errors? Because they are often hints at plain bugs! Going over the remaining cases here:

---
```
C:\cesium\packages\engine\Source\Core\Transforms.js
  785:5  error  The value assigned to 'result' is not used in subsequent statements  no-useless-assignment
```
This points to https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/packages/engine/Source/Core/Transforms.js#L785  

And this looks like a clear bug for me. The call `Matrix3.fromHeadingPitchRoll` at the bottom of that function should receive the `result` parameter. Right now, the `result` parameter of that function simply doesn't do what it promises to do. If nobody disagrees, I'd fix this. (It will be a "breaking change", though...)

---
```
C:\cesium\packages\engine\Source\Scene\ArcGisMapServerImageryProvider.js
  330:5  error  The value assigned to 'credit' is not used in subsequent statements  no-useless-assignment
```
This points to https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js#L330 

This also looks like a plain bug. Whatever is supposed to happen with the `options.credit`: It is not happening. I tried it out, creating an instance and passing in the `credits`, and they simply don't appear (obviously, in view of the code).

---
```
C:\cesium\packages\engine\Source\Scene\GlobeSurfaceTileProvider.js
  2881:9  error  The value assigned to 'boundingVolume' is not used in subsequent statements  no-useless-assignment
```
This is pointing to https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/packages/engine/Source/Scene/GlobeSurfaceTileProvider.js#L2881 

One of my favorite classes, and my facorite function, `addDrawCommandsForTile`, with >800 (eight hundred) lines of code. This already came up in https://github.com/CesiumGS/cesium/issues/12989 . EDIT: It looks like this assignment could just be omitted, given that the bounding volume is also passed in as the result parameter

---
```
C:\cesium\packages\engine\Source\Scene\Model\InstancingPipelineStage.js
  297:5  error  The value assigned to 'modelMatrix' is not used in subsequent statements            no-useless-assignment
  303:5  error  The value assigned to 'nodeComputedTransform' is not used in subsequent statements  no-useless-assignment
  311:5  error  The value assigned to 'modelMatrix' is not used in subsequent statements            no-useless-assignment
  317:5  error  The value assigned to 'nodeComputedTransform' is not used in subsequent statements  no-useless-assignment
```
These are trivial on the one hand (and I'll likely fix them soon). On the other hand, there is a deeper issue lurking there, and I already pointed that out nearly two years ago, in https://github.com/CesiumGS/cesium/issues/12163 , pointing to this exact function: We **have** to be clearer (and stricter) about which functions return their `result` parameter or not. 

---
```
C:\cesium\packages\engine\Source\Scene\View.js
  216:7  error  The value assigned to 'frustumCommands' is not used in subsequent statements  no-useless-assignment
```
This points to https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/packages/engine/Source/Scene/View.js#L216

One could probably pragmatically "fix" this, but it's not even clear to be what is supposed to happen there on a _conceptual_ level. 

The reason why I think that this should **carefully** be investigated is simple: https://github.com/CesiumGS/cesium/issues/1477 , and dozens (maybe hundreds) of forum threads that reported an error in this function. There may be a bug in that function, that may be caused by (or related to) that unnecessary assignment, and it may have been there since 2014 (!), and ... well, nobody seems to care. We can fix this with `eslint-disable-next-line no-useless-assignment`. 

---
```
C:\cesium\packages\engine\Specs\Scene\ClippingPolygonSpec.js
  104:5  error  The value assigned to 'polygonB' is not used in subsequent statements  no-useless-assignment
```
This points to https://github.com/CesiumGS/cesium/blob/fbf894e589749a71178e0729e10203ee199587dd/packages/engine/Specs/Scene/ClippingPolygonSpec.js#L104 

Rather an "outtake", to lighten the mood (after the previous one): That's an easy and tricky way to make a spec pass 🙂 

## Testing plan

No explicit testing for now, except that the tests should still pass (or maybe not, that depends...)

## Author checklist

## AI acknowledgment

- I used AI to generate content in this PR
  - [ ] Yes
  - [ ] No
  - [x] NO!!!


packages/engine/Source/Core/CorridorGeometry.js
jjspace, line 89: One could probably argue that this was intended to be:

```js
    const corner = corners[i];
    const leftSide = corner.leftPositions;
    if (defined(leftSide)) {
      length = leftSide.length;
      leftCount += length;
      indicesLength += length;
    } else {
      length = corner.rightPositions.length;
      rightCount += length;
      indicesLength += length;
    }
```

packages/engine/Source/DataSources/KmlDataSource.js
```diff
@@ -1497,7 +1497,7 @@ function processPositionGraphics(
     billboard.image = undefined;
   }
 
-  let scale = 1.0;
+  let scale;
   if (defined(billboard.scale)) {
     scale = billboard.scale.getValue();
```
jjspace: This is only used in the internal scope right?
```suggestion
  if (defined(billboard.scale)) {
    const scale = billboard.scale.getValue();
```

javagl:
The first commits have already been explained. 

The enumerated ones are plainly, mechanically applying the trivial fixes. 

The most "subtle" ones have been cases of
```
array[++index] = 123;
array[++index] = 234;
```
where the last line had to be changed to
`array[(index + 1)] = 234;`
in contrast to cases of
```
array[index++] = 123;
array[index++] = 234;
```
where the last line had to be changed to
`array[index] = 234;`

---

Wrapping up the points that had been open: 

- The `result` parameter of `Transforms.computeMoonFixedToIcrfMatrix` had not been used. This is fixed now, and it is a "breaking change", and therefore pointed out in the CHANGES.md
- The cases where the error had been caused by a `result = Example.compute(..., result);` pattern have been changed to omit the assignment, as in `Example.compute(..., result);`. For broader discussions about this, feel free to reopen https://github.com/CesiumGS/cesium/issues/12163 
- The creation of the useless `credit` instance in `ArcGisMapServerImageryProvider` was removed. The documentation of the 'credit' parameter says
  > This parameter is ignored when accessing a tiled server.

  And I assume that this is the case here.
- What I originally considered to be an "outtake" is actually a bit of an issue: It is not possible to create two `ClippingPolygon` instances that are `equal` according to the current implementation. However, this is not an issue of this PR, but something that warrants a broader discussion (including the question of ["What is equality?"](https://github.com/CesiumGS/cesium/issues/12389#issuecomment-2561985063) that I brought up in 2024)

javagl:
I just removed the last useless assignment as well. It's _technically_ trivial. Whether or not the command list should involve some handling for https://github.com/CesiumGS/cesium/issues/1477 is not related to this PR.

eslint.config.js
```diff
@@ -23,6 +23,8 @@ export default [
       "index.release.html",
       "Apps/HelloWorld.html",
       "Apps/Sandcastle2/",
+      "Apps/Sandcastle/gallery/gallery-index.js",
+      "Apps/Sandcastle/jsHintOptions.js",
```
jjspace: These files were removed

packages/engine/Source/Scene/IonImageryProvider.js
```diff
@@ -26,17 +26,13 @@ import IonImageryProviderFactory from "./IonImageryProviderFactory.js";
  * @alias IonImageryProvider
  * @constructor
  *
- * @param {IonImageryProvider.ConstructorOptions} [options] Object describing initialization options
```
jjspace, line 29: Is the type for this still used somewhere else? If not can we remove it to?

eslint.config.js
jjspace: These files no longer exist with the removal of the legacy sandcastle code

jjspace (commented):
Saw this pop up in my notifications and realized I still had some pending comments I never submitted.

I probably won't have time to finalize review of this but I think all the other changes seem good with a quick glance.

packages/engine/Source/Core/CorridorGeometry.js
javagl: Yes. There are several similar places. The goal here was to fix the error.

packages/engine/Source/DataSources/KmlDataSource.js
javagl: There are several places where variables could (and should) have a smaller scope. The goal here was to fix errors.

packages/engine/Source/Scene/IonImageryProvider.js
javagl: It's used in `fromAssetId`

eslint.config.js
javagl: Removed them from the config

jjhembd (approved):
I just got annoyed with the seatbelt file myself, and started to fix the files one by one. Then I ran into a question about increment operators, searched for discussion about it, and found this PR!

Looks good to me, thanks @javagl!

jjhembd:
@javagl I just realized your `CHANGES.md` entry from https://github.com/CesiumGS/cesium/pull/13463/commits/8a6c8712a90a0e2e60e949f29132bbd304f3a722 did not make it into the final diff. Was this intentional, or did I miss something in the merge?
I'm happy to put it back, just want to make sure your understanding of the change didn't change.

javagl:
@jjhembd Yes, it looks like that change was lost when merging in 'main' at some point. We should probably find a better way to maintain the CHANGES.md. It's nearly guaranteed to cause merge conflicts, and when _creating_ a PR, it's not even clear which version the change will end up in, meaning that the entries are at least prone to ending up in the wrong version section. I tried to straighten out some of the recent results of such "race conditions" in https://github.com/CesiumGS/cesium/pull/13612 (including the change from here).

PR #13465, opened 2026-05-04
BufferPointCollection: Fix for dynamic geometry

donmccurdy:
# Description

Fixes a bug in BufferPointCollection, where calling `point.setPosition(position)` did not update the dirty flag. The bug did not affect BufferPolylineCollection or BufferPolygonCollection because they make other assignments that set the dirty flag. I've added explicit calls to `this._dirty = true` in all three collections, to make this obvious and explicit.

The unit tests didn't catch this, because the implementation _did_ update the bounding volume, causing the primitives not to render when out of view. I've updated the unit tests to verify that geometry is updated, not just the bounding volume.

## Issue number and link

n/a

## Testing plan

See attached unit test, and sandcastle script below.

```javascript
import {
  BufferPoint,
  BufferPointCollection,
  BufferPointMaterial,
  Cartesian3,
  Color,
  ComponentDatatype,
  Viewer,
} from "cesium";

const viewer = new Viewer("cesiumContainer", { shouldAnimate: true });

const PRIMITIVE_COUNT = 100_000;
const OFFSET_HEIGHT = 500_000;

const collection = new BufferPointCollection({
  primitiveCountMax: PRIMITIVE_COUNT,
  positionDatatype: ComponentDatatype.FLOAT,
});

const point = new BufferPoint();
const material = new BufferPointMaterial({ size: 2 });

// Scratch.
const position = new Cartesian3();

// Fill polyline collection.
for (let i = 0, il = collection.primitiveCountMax; i < il; i++) {
  initializePointPosition(position);
  Color.lerp(Color.RED, Color.BLUE, Math.random(), material.color);
  collection.add({ position, material }, point);
}

window.collection = collection;

// Update point positions.
viewer.clock.onTick.addEventListener(() => {
  console.time("tick");
  for (let i = 0, il = collection.primitiveCount; i < il; i++) {
    collection.get(i, point);
    point.getPosition(position);
    updatePointPosition(position);
    point.setPosition(position);
  }
  console.timeEnd("tick");
});

function initializePointPosition(result) {
  return Cartesian3.fromDegrees(
    Math.random() * 360 - 180,
    Math.random() * 180 - 90,
    Math.random() * OFFSET_HEIGHT,
    undefined,
    result,
  );
}

function updatePointPosition(result) {
  result.x += (Math.random() * 2 - 1) * 2500;
  result.y += (Math.random() * 2 - 1) * 2500;
  result.z += (Math.random() * 2 - 1) * 2500;
  return result;
}
```

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13465** 👈

### DIFF
--- a/packages/engine/Source/Scene/BufferPointCollection.js
+++ b/packages/engine/Source/Scene/BufferPointCollection.js
@@ -7,6 +7,7 @@ import Frozen from "../Core/Frozen.js";
 import renderPoints from "./renderBufferPointCollection.js";
 import BufferPointMaterial from "./BufferPointMaterial.js";
 
+/** @import ComponentDatatype from "../Core/ComponentDatatype.js"; */
 /** @import Matrix4 from "../Core/Matrix4.js"; */
 /** @import FrameState from "./FrameState.js"; */
 
@@ -55,10 +56,12 @@ import BufferPointMaterial from "./BufferPointMaterial.js";
 class BufferPointCollection extends BufferPrimitiveCollection {
   /**
    * @param {object} options
+   * @param {Matrix4} [options.modelMatrix=Matrix4.IDENTITY] Transforms geometry from model to world coordinates.
    * @param {number} [options.primitiveCountMax=BufferPrimitiveCollection.DEFAULT_CAPACITY]
    * @param {boolean} [options.show=true]
+   * @param {ComponentDatatype} [options.positionDatatype=ComponentDatatype.DOUBLE]
+   * @param {boolean} [options.allowPicking=false] When <code>true</code>, primitives are pickable with {@link Scene#pick}. When <code>false</code>, memory and initialization cost are lower.
    * @param {boolean} [options.debugShowBoundingVolume=false]
-   * @param {boolean} [options.allowPicking=false]
    */
   constructor(options = Frozen.EMPTY_OBJECT) {
     super({ ...options, vertexCountMax: options.primitiveCountMax });
--- a/packages/engine/Source/Scene/BufferPolygon.js
+++ b/packages/engine/Source/Scene/BufferPolygon.js
@@ -180,6 +180,7 @@ class BufferPolygon extends BufferPrimitive {
       positionView[(vertexOffset + i) * 3 + 2] = positions[i * 3 + 2];
     }
 
+    this._dirty = true;
     collection._makeDirtyBoundingVolume();
   }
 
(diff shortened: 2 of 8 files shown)


packages/engine/Source/Scene/BufferPointCollection.js
```diff
@@ -55,10 +56,12 @@ import BufferPointMaterial from "./BufferPointMaterial.js";
 class BufferPointCollection extends BufferPrimitiveCollection {
   /**
    * @param {object} options
+   * @param {Matrix4} [options.modelMatrix=Matrix4.IDENTITY] Transforms geometry from model to world coordinates.
    * @param {number} [options.primitiveCountMax=BufferPrimitiveCollection.DEFAULT_CAPACITY]
    * @param {boolean} [options.show=true]
+   * @param {ComponentDatatype} [options.positionDatatype=ComponentDatatype.DOUBLE]
+   * @param {boolean} [options.allowPicking=false] When <code>true</code>, primitives are pickable with {@link Scene#pick}. When <code>false</code>, memory and initialization cost are lower.
    * @param {boolean} [options.debugShowBoundingVolume=false]
```
donmccurdy, line 64: Minor, unrelated fix. `ComponentDatatype` was missing from the BufferPointCollection constructor JSDoc. Not sure how to keep this mirrored from  BufferPrimitiveCollection with TypeScript+JSDoc yet.

packages/engine/Source/Scene/BufferPolygon.js
```diff
@@ -489,6 +491,7 @@ class BufferPolygon extends BufferPrimitive {
       dstIndices[(triangleOffset + i) * 3 + 2] = indices[i * 3 + 2];
     }
 
+    this._dirty = true;
     collection._makeDirtyBoundingVolume();
   }
```
danielzhong, line 494: `this._dirty = true` in `setHoles()` and `setTriangles()` make sense to me. However, I noticed that `collection._makeDirtyBoundingVolume();` is also being called in both places. I don’t think that’s necessary—`setHoles()` and `setTriangles()` only modify index data and don’t change any vertex positions, so the bounding volume shouldn’t be affected. Right?

donmccurdy: You're right, thanks! I think these should only have been calling `._dirty = true` and not `._makeDirtyBoundingVolume()`, fixed. ✅

danielzhong (approved):

PR #13467, opened 2026-05-05
Vector Tiles: Fix missing cases for tile content with multiple property tables

donmccurdy:
# Description

Additional fixes for https://github.com/CesiumGS/cesium/issues/11683. Affects Vector Tiles only, for now, although similar changes on the Cesium3DTileContent interface will likely be wanted for other 3D Tiles content types at some point. `content.getFeature(...)` and `content.hasProperty(...)` are affected and must take a Table ID, because Feature ID is unique only to a particular property table. A fe

While adding unit tests, I also added or updated implementations of getters for `content.featuresLength`, `content.batchTableByteLength`, which weren't involved in the bug, but are relevant to supporting multiple property tables.

## Issue number and link

- #11683
- #13432

## Testing plan

Unit tests added for VectorGltf3DTileContent. Public data with multiple property tables not yet available — will share a reproduction case by DM.

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13467** 👈

### DIFF
--- a/packages/engine/Source/Scene/VectorGltf3DTileContent.js
+++ b/packages/engine/Source/Scene/VectorGltf3DTileContent.js
@@ -19,6 +19,7 @@ import createVectorTileBuffersFromModelComponents from "./Model/createVectorTile
 import defined from "../Core/defined.js";
 import destroyObject from "../Core/destroyObject.js";
 import DeveloperError from "../Core/DeveloperError.js";
+import Check from "../Core/Check.js";
 
 /** @import BufferPrimitive from "./BufferPrimitive.js"; */
 /** @import BufferPrimitiveCollection from "./BufferPrimitiveCollection.js"; */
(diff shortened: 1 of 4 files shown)


packages/engine/Source/Scene/VectorGltf3DTileContent.js
```diff
@@ -189,19 +210,25 @@ class VectorGltf3DTileContent {
...
   /**
    * @param {number} featureId
    * @param {string} name
+   * @param {number} featureTableId
    * @returns {boolean}
    */
-  hasProperty(featureId, name) {
-    const feature = this.getFeature(featureId);
+  hasProperty(featureId, name, featureTableId) {
+    const feature = this.getFeature(featureId, featureTableId);
```
danielzhong, line 231: I'm wondering should we add `featureTableId` as an optional parameter to `Cesium3DTileContent.prototype.getFeature` and `hasProperty` in this PR as well, to keep the base interface aligned with VectorGltf3DTileContent's new multi-table signatures?

packages/engine/Specs/Scene/VectorGltf3DTileContentSpec.js
```diff
@@ -0,0 +1,227 @@
...
+    content._model = { _featureTables: [] };
+  });
+
+  afterEach(() => {});
+
+  it("featuresLength", () => {
+    content._collections = [
+      createBufferPointCollection(),
+      createBufferPolylineCollection(),
+    ];
```
danielzhong, line 38: I think we can remove this part right? `featuresLength` is now implemented by summing over `_model._featureTables`, so setting `_collections` here has no effect on the result

packages/engine/Source/Scene/VectorGltf3DTileContent.js
donmccurdy: Agreed, I think we should – I'd like to wait and do that after https://github.com/CesiumGS/cesium/pull/13443 lands, though. Easier to make sure all the implementations of the interface match up if we have the ES6 classes and (maybe) type checks, and eventually we can support multiple property tables on the other 3D Tiles content types too.

packages/engine/Specs/Scene/VectorGltf3DTileContentSpec.js
donmccurdy: Ah, good question thanks! You're right that the collections are not used, but the test is intentionally checking that they're unused. I should have added a comment to that effect, and have added the following inline now:

```javascript
    // For consistency with other tile content, features are counted from the
    // batch table, not unique IDs in the geometry, so feature length should be
    // zero if no batch tables are added. We can change this in the future; the
    // test is just here to protect against accidental changes.
```

danielzhong (approved):

PR #13468, opened 2026-05-06
Fix statistics in sample voxel tilesets

j9liu:
# Description

This fixes some incorrect sample data that didn't specifiy the min/max voxel values in a correct way. An early version of the datasets may have been single floats, but the property was ultimately changed to a float `VEC4`. This requires values to be defined as arrays, which was not the case.

Discovered here: https://community.cesium.com/t/bugs-or-not-cesiumvoxelmetadata-cesium3dtilesetvoxel/45976

## Issue number and link

N/A

## Testing plan

Not sure how to test -- let me know if I'm missing something obvious?

## Author checklist

- ~[ ] I have updated `CHANGES.md` with a short summary of my change~ -- Sample data only
- [ ] I have added or updated unit tests to ensure consistent code coverage --  Is this needed? 
- ~[ ] I have updated the inline documentation, and included code examples where relevant~

## AI acknowledgment


jjhembd (approved):
Thanks @j9liu!

RE testing: Internally, we always convert the min/max to arrays (see `copyArray` in `Cesium3DTilesVoxelProvider`) and then back to scalars (if needed) in the shader struct (see `processVoxelProperties`). So this change won't affect any CesiumJS behavior.

It is still good to make our sample data conform to the spec. Thanks again for catching this!

PR #13469, opened 2026-05-06
Buffer*Collection gpu normalized

danielzhong:
# Description
Extends the `positionNormalized` API introduced in #13462 by moving position
dequantization entirely to the GPU, eliminating per-frame CPU work.

Previously, when `positionNormalized: true` was set, the CPU decoded each
int16 position to a float64 Cartesian3, applied `modelMatrix`, then
re-encoded to high/low float32 pairs before upload. This PR replaces that
path by keeping integer data in the vertex buffer and relying on WebGL's
built-in `normalize` attribute flag to decode to `[-1, 1]` on the GPU.

## Changes

- **Shaders** (`Buffer{Point,Polyline,Polygon}MaterialVS.glsl`): added
  `#ifdef QUANTIZED_POSITIONS` path using `czm_modelView * vec4(position, 1.0)`
  in place of the RTE `positionHigh`/`positionLow` path.
- **Renderers** (`renderBuffer{Point,Polyline,Polygon}Collection.js`): added
  `useQuantized` branch that reads `_positionView` (int16) directly and writes
  to typed attribute arrays; vertex buffers are created with `normalize: true`.
- **BufferPrimitiveCollection**: removed `_denormalizePosition()` and
  `_commandModelMatrix` (CPU-era artifacts); all draw commands now pass
  `collection.modelMatrix` directly.

## Issue number and link
https://github.com/CesiumGS/cesium/pull/13462
https://github.com/CesiumGS/cesium/issues/13444

## Author checklist

## AI acknowledgment

### DIFF
--- a/packages/engine/Source/Scene/renderBufferPointCollection.js
+++ b/packages/engine/Source/Scene/renderBufferPointCollection.js
@@ -281,7 +281,7 @@ function renderBufferPointCollection(collection, frameState, renderContext) {
       context,
       vertexShaderSource: new ShaderSource({
         sources: [BufferPointMaterialVS],
-        defines: useLocalSpace ? ["LOCAL_POSITION"] : [],
+        defines: useFloat64 ? ["USE_FLOAT64"] : [],
       }),
       fragmentShaderSource: new ShaderSource({
         sources: [BufferPointMaterialFS],
--- a/packages/engine/Source/Scene/renderBufferPointCollection.js
+++ b/packages/engine/Source/Scene/renderBufferPointCollection.js
@@ -18,7 +18,6 @@ import PrimitiveType from "../Core/PrimitiveType.js";
 import BufferPointMaterialVS from "../Shaders/BufferPointMaterialVS.js";
 import BufferPointMaterialFS from "../Shaders/BufferPointMaterialFS.js";
 import EncodedCartesian3 from "../Core/EncodedCartesian3.js";
-import Matrix4 from "../Core/Matrix4.js";
 import AttributeCompression from "../Core/AttributeCompression.js";
 import BufferPointMaterial from "./BufferPointMaterial.js";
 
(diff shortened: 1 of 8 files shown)


packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -72,24 +85,29 @@ const encodedCartesian = new EncodedCartesian3();
 function renderBufferPointCollection(collection, frameState, renderContext) {
   const context = frameState.context;
   renderContext = renderContext || { destroy: destroyRenderContext };
+  const useQuantized = collection._positionNormalized;
```
donmccurdy: I think we can get even better performance with this change (even for uncompressed data), if we change this to something like...

```javascript
const useFloat64 = collection._positionDatatype === ComponentDatatype.DOUBLE;
```

... and then reserve the `EncodedCartesian3` encoding only for float64 inputs. Most Vector Tiles will use float32 in the short term (until we have broader support for tiling into meshopt, at least), so we can cut the memory allocation in half for float32 inputs this way.

packages/engine/Source/Scene/renderBufferPointCollection.js
```diff
@@ -199,16 +233,38 @@ function renderBufferPointCollection(collection, frameState, renderContext) {
...
+      for (const key in BufferPointAttributeLocations) {
+        if (Object.hasOwn(BufferPointAttributeLocations, key)) {
+          const attribute = /** @type {BufferPointAttribute} */ (key);
+          renderContext.vertexArray.copyAttributeFromRange(
+            BufferPointAttributeLocations[attribute],
+            renderContext.attributeArrays[attribute],
+            collection._dirtyOffset,
+            collection._dirtyCount,
+          );
+        }
```
donmccurdy, line 267: Would it be possible to combine these two `if (useQuantized) { ... } else { ... }` cases into one, if we store something like:

```javascript
renderContext.attributeLocations = useQuantized ? X : BufferPointAttributeLocations;
```

Just thinking it will be easier to read if updates follow the same code path either way.

packages/engine/Source/Shaders/BufferPointMaterialVS.glsl
```diff
@@ -25,8 +29,13 @@ void main()
 
     ///////////////////////////////////////////////////////////////////////////
 
+#ifdef QUANTIZED_POSITIONS
+    // Normalized integer positions are hardware-decoded to [-1, 1] by the attribute pipeline.
+    vec4 positionEC = czm_modelView * vec4(position, 1.0);
+#else
```
donmccurdy: As above, I think we want the condition here to be `USE_FLOAT64` and all other cases can use a single vertex attribute for position data.

> // Normalized integer positions are hardware-decoded to [-1, 1] by the attribute pipeline.

Nit - Positions may be decoded to [0,1] or [-1,1] depending on whether the compression library chose unsigned or signed integers, or they might be quantized but not normalized... but we probably don't need to explain that here as it could be non-normalized float32 input as well, so maybe simplest just to remove the comment. Happy to chat more offline if that doesn't make sense! 😅

donmccurdy (commented):
This looks great! Haven't tested locally yet but I think this handles meshopt compression well — keeping vertex data compressed throughout the render process — and also improves other performance goals described in https://github.com/CesiumGS/cesium/issues/13444. 

Left a couple small comments but should be minor. After those changes, unless you'd prefer otherwise, I think my vote would be to merge this PR into #13462, do a final review there, and then merge the combined PR.

PR #13470, opened 2026-05-06
Edge Visibility clean up

danielzhong:
# Description
a_silhouetteNormal was uploaded to the GPU and v_silhouetteNormalView was computed in the vertex shader, but neither was ever consumed by the fragment shader. This PR removes the dead code and the associated CPU fallback.

Changes:
* Removed `a_silhouetteNormal` vertex attribute and `v_silhouetteNormalView` varying from `EdgeVisibilityStageVS.glsl`
* Removed `buildTriangleAdjacency` and the CPU-side adjacency fallback from `EdgeVisibilityPipelineStage.js`
* Simplified `generateEdgeFaceNormals` to read silhouette normals directly from the GLB accessor data

## Author checklist

## AI acknowledgment

GitHub Copilot

Cluade Sonnet 4.6

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 #### Fixes :wrench:
 
 - Fixed a bug causing `BufferPointCollection` to not update after changes to point positions. [#13465](https://github.com/CesiumGS/cesium/pull/13465)
+- Removed dead `a_silhouetteNormal` attribute and CPU fallback in `EXT_mesh_primitive_edge_visibility` pipeline stage. [#13470](https://github.com/CesiumGS/cesium/pull/13470)
 
 ## 1.141 - 2026-05-01
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,6 @@
 #### Fixes :wrench:
 
 - Fixed a bug causing `BufferPointCollection` to not update after changes to point positions. [#13465](https://github.com/CesiumGS/cesium/pull/13465)
-- Removed dead `a_silhouetteNormal` attribute and CPU fallback in `EXT_mesh_primitive_edge_visibility` pipeline stage. [#13470](https://github.com/CesiumGS/cesium/pull/13470)
 
 ## 1.141 - 2026-05-01
 
--- a/packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
+++ b/packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
@@ -16,6 +16,17 @@ import VertexAttributeSemantic from "../VertexAttributeSemantic.js";
 import AttributeType from "../AttributeType.js";
 
 /**
+ * 2-bit edge visibility values as defined by EXT_mesh_primitive_edge_visibility.
+ * @private
+ */
+const EdgeVisibilityType = Object.freeze({
+  HIDDEN: 0,
+  SILHOUETTE: 1,
+  HARD: 2,
+  REPEATED_HARD: 3,
+});
+
+/**
  * Builds derived line geometry for model edges using EXT_mesh_primitive_edge_visibility data.
  * It parses the encoded edge visibility bits, creates a separate edge-domain vertex array with
  * per-edge attributes (edge type, optional feature ID, silhouette normal, adjacent face normals),
@@ -263,7 +274,7 @@ function generateEdgeFaceNormals(edgeIndices, edgeData, edgeVisibility) {
 
   const raw = edgeVisibility.silhouetteNormals;
   if (!defined(raw)) {
-    // No silhouette edges in this primitive， the shader only checks normals for edgeType === 1.
+    // No silhouette edges in this primitive, the shader only checks normals for SILHOUETTE edges.
     return edgeFaceNormals;
   }
   const silhouetteNormalsFloat = new Float32Array(raw.length * 3);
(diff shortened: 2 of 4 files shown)


packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -128,9 +132,6 @@ EdgeVisibilityPipelineStage.process = function (
 
   // Add varying for view space position for perspective-correct silhouette detection
```
markschlosseratbentley: I realize this isn't a code comment added or modified in this PR, but I just noticed there does not actually seem to be a varying added here. Is this comment also a candidate for removal in this PR?

CHANGES.md
```diff
@@ -7,6 +7,7 @@
 #### Fixes :wrench:
 
 - Fixed a bug causing `BufferPointCollection` to not update after changes to point positions. [#13465](https://github.com/CesiumGS/cesium/pull/13465)
+- Removed dead `a_silhouetteNormal` attribute and CPU fallback in `EXT_mesh_primitive_edge_visibility` pipeline stage. [#13470](https://github.com/CesiumGS/cesium/pull/13470)
```
markschlosseratbentley: Since this is cleanup and does not introduce new functionality, does this warrant a CHANGES.md entry? Or should it be under another section? (Refactors/Cleanups or something?)

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -614,17 +425,14 @@ function extractVisibleEdges(primitive) {
         // Only process silhouette edges (type=1) as marked in GLB
         // Use computed edge index for boundary edges
         let mateVertexIndex = -1;
-        if (visibility2Bit === 1) {
+        if (visibility2Bit === EdgeVisibilityType.SILHOUETTE) {
           mateVertexIndex = silhouetteEdgeCount;
           silhouetteEdgeCount++;
         }
 
         edgeData.push({
```
markschlosseratbentley, line 434: I'm looking at EdgeVisibilityType.REPEATED_HARD... it doesn't appear to be different from HARD (2) anywhere. Am I right that the shader only branches on SILHOUETTE, and 2/3 enum values render identically? Is this intentional? Should we keep the enum value for REPEATED_HARD (keeping spec phrasing? Spec: https://github.com/CesiumGS/glTF/blob/64ed5bea4ccffb2242d2cf86e212388f63b61603/extensions/2.0/Vendor/EXT_mesh_primitive_edge_visibility/README.md?plain=1#L81), or is it worth collapsing both types to HARD at decode and dropping REPEATED_HARD from the enum here in CesiumJS?

LMK if I'm looking at any of this incorrectly.

CHANGES.md
j9liu: I think +1 for this since it's not user-facing

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
```diff
@@ -614,17 +425,14 @@ function extractVisibleEdges(primitive) {
         // Only process silhouette edges (type=1) as marked in GLB
         // Use computed edge index for boundary edges
         let mateVertexIndex = -1;
```
j9liu: One nitpick: I didn't understand what a "mate vertex" was at first, but once I traced the code fully I realized this means the index of the entry in `silhouetteNormals`. Maybe could benefit from a better name, like `silhouetteEdgeIndex` ?

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
j9liu: Perhaps there's a branch missing here -- if the edge type is `REPEATED_HARD`, we shouldn't create a new edge for it, right? And perhaps this is meant to negate the need for a seen edge hash map?

j9liu (commented):
This looks great @danielzhong! Just one minor nitpick, otherwise I also +1 Marks' comments

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
danielzhong: Great suggestion, thanks!

packages/engine/Source/Scene/Model/EdgeVisibilityPipelineStage.js
danielzhong: You're right. I've normalized REPEATED_HARD to HARD at decode time so the stored edgeType values are always 0/1/2, keeping the MRT output clean. The REPEATED_HARD entry is kept in the enum purely for documentation purposes, with a comment explaining the normalization.

danielzhong:
> Perhaps there's a branch missing here -- if the edge type is REPEATED_HARD, we shouldn't create a new edge for it, right? > And perhaps this is meant to negate the need for a seen edge hash map?

@j9liu The seenEdgeHashes set already handles this when a REPEATED_HARD edge is encountered, its key is already in the hash set (from the first occurrence, which was encoded as HARD), so it hits the continue branch and no duplicate edge is created. The hash map is the deduplication mechanism for REPEATED_HARD; no extra branch is needed. I added a comment to explain better in the code!

danielzhong:
@markschlosseratbentley @j9liu Thanks for the feedback, ready review again!

markschlosseratbentley (approved):

j9liu (approved):

PR #7502, opened 2019-01-23
Support for Proj4JS and Custom Projections in Columbus View and 2D

likangning93:
This PR adds support for Proj4js projections and custom, user-defined projections in 2D and Columbus View.

See #6986 for discussion in the previous review pass + pretty pictures.

Just since it's been a while, I'm going to take a look at the diff here first, update `CHANGES.md`, and then request review when ready.

TODO:
* [ ] fix camera issues with off-center projections like EPSG:2039 - https://epsg.io/2039
* [x] fix performance issues with repeated calls to `BoundingSphere.fromRectangleWithHeights2D`
* [ ] put a better writeup here so it's easier for fresh reviewers to look

PR #13471, opened 2026-05-06
Add support of Proj4JS & custom map projections

mickae1:
## Summary

This PR adds support for custom map projections in Columbus View and 2D scene modes. It is adapted from the unmerged [#7502](https://github.com/CesiumGS/cesium/pull/7502) by @likangning93 (open 2019–2021), brought up to date with the current codebase and made more robust against the edge cases that surfaced during the original review.

Three new public projection classes implement the `MapProjection` interface:

- **`Proj4Projection`** — proj4 Well-Known Text definitions. Covers Lambert Azimuthal Equal-Area, Mollweide, Robinson, Equal Earth, stereographic and any EPSG code supported by proj4js.
- **`CustomProjection`** — user-defined `project` / `unproject` functions. Functions must be self-contained because they are serialized via `Function.prototype.toString()` for web worker transfer.
- **`Matrix4Projection`** — 4×4 affine projection.

`MapProjection` gains an `isNormalCylindrical` property that gates the existing fast paths for `GeographicProjection` and `WebMercatorProjection`. Non-cylindrical projections fall back to grid-sampled computations for `approximateMaximumCoordinate`, `BoundingSphere.fromRectangleWithHeights2D`, and rectangle extent approximation. All projections gain `serialize()` / static `deserialize()`; `Scene` caches the serialized form on `FrameState` for worker transfer via `PrimitivePipeline`.

Five automatic GLSL uniforms expose projection parameters to custom vertex shaders:
`czm_mapProjectionType`, `czm_projectionParams`, `czm_projectionOffsets`, `czm_projectionEllipsoidParams`, `czm_projectionEllipsoidParams2`. Values are read directly from proj4's parsed representation (`projFunction.oProj.long0/lat0/x0/y0/a/b/es`) at construction time, so CPU and GPU agree on every WKT variant — alternate datums, alias parameter names, EPSG references, and WKT2 form all work identically.

A new Sandcastle (`Map Projections`) demonstrates the seven projections (Lambert Europe, Lambert USA, Geographic, Web Mercator, Mollweide, Robinson, Equal Earth) with GeoJSON country borders, polygon/polyline/extruded volume primitives and a picking overlay. A new `Documentation/CustomProjectionsGuide/` covers usage, GLSL uniforms, serialization, and known limitations.

Adds `proj4` (`^2.20.8`) as a dependency to `@cesium/engine` and credits proj4js (MIT) in `LICENSE.md`.

## Changes from the original PR #7502

The original PR was reviewed extensively but never merged, with several unresolved issues. This rework addresses the most impactful ones:

- **`Proj4Projection.project` / `unproject` are NaN-safe.** Non-finite inputs propagate as NaN components instead of throwing through proj4, matching `GeographicProjection`'s behavior. Cesium's camera / `adjustHeightForTerrain` chains regularly re-feed cartographics that may carry NaN; throwing breaks rendering.
- **Antimeridian-crossing rectangles are split.** The new grid samplers detect `east < west` and walk `[west, π] ∪ [-π, east]` so wrap-around rectangles produce correct bounds rather than sampling the complementary band.
- **Silent CPU/GPU divergence is eliminated.** Reading projection parameters from proj4's parsed AST instead of regex-matching the WKT string avoids a class of subtle bugs where the GPU shader silently used wrong coordinates because the regex didn't recognize the user's WKT variant.
- **`pickMapColumbusView` filters non-finite cartographics.** A pre-existing Cesium bug where the pick ray running near-parallel to the projection plane produced `Infinity` scalars and crashed `cartographicToCartesian`.
- **`UniformState.view2Dto3D` clamps NaN before computing the geodetic surface normal.** Similar pre-existing path, exposed by custom projections that legitimately return NaN outside their valid area.
- **`adjustHeightForTerrain` clamps `camera.position.x` to a positive minimum in `COLUMBUS_VIEW`.** Prevents tilt inputs from driving the camera below the projection plane into a degenerate state. Mirrors the issue I observed with `WebMercatorProjection` as well — it appears to be a pre-existing limitation that custom projections make more reachable.

## Tests

Adds 90 unit tests across `Proj4Projection`, `CustomProjection`, `Matrix4Projection`, `MapProjection`, `deserializeMapProjection`, `Rectangle.approximateProjectedExtents` / `CartographicExtents`, and `BoundingSphere.fromRectangleWithHeights2D` non-cylindrical path. Coverage includes happy paths, NaN/Infinity inputs, antimeridian crossings, projection-singular antipodes, ellipsoid round-trip, scratch reuse, homogeneous-coordinate (`w ≠ 1`) projection, zero-size rectangles, and serialized round-trips for every projection type.

## Known limitations (documented in `Documentation/CustomProjectionsGuide/README.md`)

These are kept honest in the new guide and would warrant follow-up PRs:

1. **Terrain tile rendering and imagery reprojection are not supported** for non-cylindrical projections. Tile alignment requires changes to the `HeightmapTessellator` / `TerrainEncoding` / worker pipeline that #7502 attempted but were not ported here. Sandcastle works around this by rendering the globe transparent.
2. **Polygon fills with default `granularity` may render empty** on non-cylindrical projections because subdivided edge midpoints in 3D ECEF are not equivalent to midpoints in projected space, so subdivided triangles can wind-reverse after projection. Workaround: `granularity: Math.PI / 8`. Documented in the guide.
3. **Off-center / off-meridian projections (e.g., EPSG:2039) crash with global geometry.** Vertices near the projection's antipode produce finite-but-very-large coordinates that overflow `BoundingSphere.fromVertices` to `Infinity`, then propagate to `View.updateFrustums`. A naive last-finite-value fallback prevents the crash but produces visual artifacts (long lines crossing the map). The proper fix is to clip polygons in `GeometryPipeline.projectTo2D` against the projection's valid area before projecting; that was also unresolved in #7502.
4. **2D scene mode wrap-around** defaults to `MapMode2D.ROTATE` (rather than `INFINITE_SCROLL`) for non-cylindrical projections, since they have no horizontal repeat.

## Test plan

- [x] All 90 new unit tests pass.
- [x] Pre-commit hooks (eslint, tsc, prettier) pass.
- [x] Sandcastle "Map Projections" renders the seven projections and supports primitive picking.
- [x] Geographic and Web Mercator paths exercise the existing fast-path code with no behavioral change.
- [ ] CesiumJS test suite passes in CI (Karma + Chrome) — please flag if anything beyond the new specs regresses.

## Why open this now

Several users have asked about #7502 over the years. This rework is currently used internally on a Lambert-93 deployment and the projection / serialization / GLSL paths are stable. The terrain tile and off-center camera limitations are real but I would rather land a solid foundation than block on a complete fix; both are tracked as follow-ups in the guide.

Adapted from CesiumGS/cesium#7502 by @likangning93 — credit to the original author for the design.


mickae1:
you can find there some screenshot : 

image
image
image

packages/engine/Source/Core/Proj4Projection.js
```diff
@@ -0,0 +1,387 @@
...
+
+    this._wgs84Bounds = defined(options.wgs84Bounds)
+      ? Rectangle.clone(options.wgs84Bounds)
+      : undefined;
+    this._projectedBounds = defined(options.projectedBounds)
+      ? Rectangle.clone(options.projectedBounds)
+      : undefined;
+
+    try {
+      this._projFunction = proj4(this._wellKnownText);
```
wewindy: This parameter `_wellKnownText` seems to be able to pass Proj4String in addition to WellKnown Text, or would it be better to call it `sourceProjection`?

mickae1: Good catch — done. Renamed `wellKnownText` to `sourceProjection` (and the internal `_wellKnownText` accordingly) across the source, specs, sandcastle, guide and `CHANGES.md`. The JSDoc and prose now also call out that any proj4js-acceptable input works (proj4 syntax, OGC WKT, registered EPSG identifier). Pushed in 14ea1e20e4.

mzschwartz5:
This looks cool. Two drive-by comments:
1. Just a note that there are very likely many places in the codebase that assume 2D projections are axis-aligned and rectangular (obviously not true generally, and code should not be assuming this... but, alas...). Things like `Globe.pick` probably won't work on projections that violate this assumption.
2. Not sure I understand the point of `CustomProjection`. `MapProjection` already _is_ a custom projection that users can implement on their own. They just make their own class that implements the interface.

packages/engine/Source/Core/Proj4Projection.js
wewindy: Good

wewindy (commented):
Good

mickae1:
## Update — removed the COLUMBUS_VIEW altitude clamp

While testing locally I noticed three pre-existing Cesium specs were now
failing under this branch:

- `Scene/ScreenSpaceCameraController > rotates in Columbus view with camera transform set`
- `Scene/ScreenSpaceCameraController > camera does not go below the terrain in CV`
- `Scene/ScreenSpaceCameraController > camera does not go below the terrain in CV with the transform set`

The cause was the COLUMBUS_VIEW altitude clamp I added in
`adjustHeightForTerrain`. Two issues with that code:

1. It clamped `camera.position.x` (local-space x), but in COLUMBUS_VIEW
   the local-space x maps to world-space y (East) via `Camera.TRANSFORM_2D`.
   The world-space altitude is actually `positionWC.x` (= local-space z).
   So the clamp was effectively freezing pan motion along the East axis
   instead of clamping altitude — which manifested as a "wall" preventing
   the user from panning across the Atlantic in Web Mercator at moderate
   zoom levels.

2. The bug it was trying to fix (camera going under the projection plane
   on extreme tilt input) is reproducible with `WebMercatorProjection` too,
   so it's a pre-existing Cesium issue rather than something specific to
   custom projections. Out of scope for this PR.

Removing the clamp restores correct pan behavior, makes those three
upstream specs pass, and keeps Web Mercator behavior consistent with
upstream. Pushed in `ef17034f9f` as a force-push amend on the same commit.

The other COLUMBUS_VIEW-related guards stay in place
(`Camera.pickMapColumbusView` NaN filter, `UniformState.view2Dto3D` NaN
clamp, `Camera.updateMembers` NaN fallback), since those address actual
crashes I observed in custom projections that are unaffected by this change.

mickae1:
@mzschwartz5
Thanks for taking a look! Both points are fair.

### On the axis-aligned / rectangular assumption

Agreed completely — there are quite a few places in the codebase that
implicitly assume the 2D projection's valid area is rectangular and
axis-aligned. The "Known Limitations" section of the new guide tries to
be honest about the most visible ones we've hit:

- **Terrain tile rendering** is unsupported for non-cylindrical projections.
  `HeightmapTessellator` / `TerrainEncoding` / the worker pipeline still
  take the rectangular-in-projection-space path. likangning93's #7502 had
  a partial fix that we did not port — left as follow-up.
- **Polygon fills** with default `granularity` show empty triangles
  (`GeometryPipeline.projectTo2D` projects edge midpoints in a way that's
  not equivalent to midpoints in projected space). Workaround:
  `granularity: Math.PI / 8`.
- **Off-center / off-meridian projections** (e.g. EPSG:2039) crash on
  global geometry. Vertices near the antipode produce finite-but-very-large
  coordinates that overflow `BoundingSphere.fromVertices` to `Infinity`.
- `Globe.pick` on the 2D ellipsoid (as you flag) — I haven't tested it
  end-to-end with a non-cylindrical projection but suspect it would have
  similar issues.

The intent of this PR is to land the infrastructure (interface +
serialization for workers + GLSL uniforms + three concrete implementations
+ 90 specs covering the projection math itself) so that follow-ups can
tackle each consumer (terrain tessellator, polygon subdivision, picking,
etc.) one at a time. Trying to do all of that in a single PR is what sank
#7502.

### On the point of `CustomProjection`

You're right that `MapProjection` is already an interface — but
`CustomProjection` solves three things that a user-defined class wouldn't:

1. **Web worker transit.** Geometry primitives are packed into web workers
   for vertex generation, and the projection has to come along for the
   ride. A user's class lives in their bundle, not in the worker's. By
   storing `project` / `unproject` as serializable function source
   (`Function.prototype.toString()`) and rehydrating with `new Function(...)`
   on the worker side, `CustomProjection` lets you ship a projection that
   actually works under primitives. That's also the reason for the
   "functions must be self-contained" constraint.
2. **Type registration.** `deserializeMapProjection` needs to know how to
   reconstruct an instance from a `SerializedMapProjection`. The
   `MapProjectionType` enum dispatches to one of the four built-ins. An
   arbitrary user class would need its own registration mechanism, which
   is more API surface for a use case that's mostly about prototyping.
3. **Ergonomics.** For one-off projections, passing
   `{ project, unproject }` is meaningfully shorter than implementing a
   full ES6 class with `isNormalCylindrical`, `serialize`, `deserialize`,
   etc.

Happy to fold this rationale into the guide if it's not clear enough from
the existing `CustomProjection` JSDoc.

mzschwartz5:
Would it make more sense then to completely replace the `MapProjection` interface with this abstract base class `Custom(Map)Projection`? (Sounds like you've thought this out pretty thoroughly, just wondering)

mickae1:
@mzschwartz5 — good question, considered this during the rework.

The blocker for collapsing the four into a single `Custom(Map)Projection`
is that the state each class needs to round-trip through
`PrimitivePipeline.packCombineGeometryParameters` →
`deserializeMapProjection` on the worker side is structurally different:

- **`Proj4Projection`** serializes to `{ sourceProjection: string,
  heightScale: number, ellipsoid: number[16], wgs84Bounds?: number[4],
  projectedBounds?: number[4] }`. The worker rehydrates by re-invoking
  `proj4(sourceProjection)` to rebuild the `oProj` AST that
  `shaderUniforms` reads from (`oProj.long0`, `lat0`, `x0`, `y0`, `a`,
  `b`, `es`). Reducing this to two serialized functions would either
  (a) ship the entire `proj4` library inline in the function source
  (~75 KB minified per primitive), or (b) lose the parsed-AST access
  that `UniformState.updateProjectionUniforms` relies on for the GLSL
  uniforms `czm_projectionParams` / `czm_projectionEllipsoidParams`.
- **`Matrix4Projection`** serializes to `{ matrix: number[16], degrees:
  boolean, ellipsoid: number[16] }`. Encoding the matrix as a closure
  inside `toString()`-able functions loses the matrix invertibility check
  that the constructor does (`Matrix4.inverse` throws on singular
  matrices, currently caught up-front; through `new Function(...)`
  rehydration it would surface as runtime NaN/Infinity on first
  `unproject`).
- **`GeographicProjection`** and **`WebMercatorProjection`** are the hot
  path for Geographic/Mercator scenes — `project` is a single `lon * R`
  / `Math.log((1 + sin(lat)) / (1 - sin(lat)))`, inlined by V8. The
  current `MapProjectionType` dispatch in `deserializeMapProjection`
  produces a class instance that V8 can monomorphize. Rehydrating via
  `new Function(GeographicProjection.prototype.project.toString())()`
  produces a fresh function object per worker per frame with no shared
  inline cache state — measurable cost in
  `GeometryPipeline.projectTo2D`, which is a tight loop over every
  vertex of every projected primitive.

There's also `isNormalCylindrical`, which is a static class shape, not a
derived behavior. `BoundingSphere.fromRectangleWithHeights2D`,
`Rectangle.approximateProjectedExtents`,
`MapProjection.approximateMaximumCoordinate`, and the SCENE2D zoom skip
in `ScreenSpaceCameraController.handleZoom` all branch on it
synchronously. With a unified `Custom(Map)Projection`, that flag would
have to be a constructor argument (effectively re-introducing the
typology that `MapProjectionType` already encodes), or worse, derived at
runtime from observed behavior of `project()` — which would push the
cost of "is this projection cylindrical?" into the bounding sphere /
extent computation it's supposed to optimize.

So the design ended up being: `MapProjection` is the structural
interface plus a couple of static helpers (`approximateMaximumCoordinate`,
`isNormalCylindrical` as a documented property); each of the four
implementations owns its serialization format because the state shapes
diverge enough that a common one would be either lossy or bloated; and
`CustomProjection` exists specifically for the "I have two stateless
JS functions, serialize them as source" case, which is genuinely the
simplest cell of the design space.

If you have a sketch where the four collapse without losing the parsed
proj4 AST on workers or the Matrix4 invertibility check, I'd want to
look at it — possible I'm missing an angle. But from where I sit this
PR is the minimal infrastructure layer; the unification you're pointing
at would need a concrete design that handles the state divergence, and I
don't see one yet.

mzschwartz5:
Sorry, I should have been clearer - I'm not suggesting collapsing the individual projection implementations into one. I'm suggesting replacing `MapProjection`, a pure interface, with `Custom(Map)Projection`, an abstract base class - where you, the author of this base class, define some functionality that helps make subclasses (even user-authored ones) function properly with web-workers and such.

That may not be the right design, I need to go back and re-read your comments more carefully. It sounds like there IS a real problem with `MapProjection` not being directly extensible to users and `CustomProjection` solves that. I'm just wondering if there's even a need for `MapProjection` at all anymore, of if `CustomProjection` can supplant it and serve as the base class for the other implementations in the library as well.

mickae1:
Ah, that clarifies it — thanks for the follow-up. You're right that there's a real
ergonomic gap: `MapProjection` is an interface (in the JSDoc sense), and users who
want to write their own can't really extend it usefully because of the worker
serialization step. `CustomProjection` papers over that by accepting two functions
and stringifying them — which works for stateless projections but is awkward for
anything with cached state (lookup tables, parsed configs, etc.).

Refactoring `MapProjection` into an abstract base class with `CustomProjection`'s
serialization machinery built in would clean up the JSDoc duplication and give
users a more idiomatic extension point. The four library implementations would
inherit from it cleanly enough — `GeographicProjection`, `WebMercatorProjection`,
`Proj4Projection`, `Matrix4Projection` each have a well-defined serialization
shape that fits as a `super.serialize()` override.

Two things still need design work in that direction, though:

1. **Worker type dispatch.** `deserializeMapProjection` currently switches on
   `MapProjectionType` (0..4) to reconstruct the right concrete class on the
   worker side. A user subclass has no slot in that enum — you'd need either
   a registry pattern (`MapProjection.register("MyProj", MyProjClass)`) or the
   user has to inject their class source into worker bundles themselves. The
   former adds public API surface; the latter is what `CustomProjection`
   already does, just via function source instead of class source. Neither is
   strictly blocked, but they're meaningfully different design decisions.

2. **API stability.** `MapProjection` is documented public surface, and there
   are user codebases doing `instanceof MapProjection` and TypeScript type
   imports against it. Renaming/removing it would be a breaking change. A
   migration could be done by keeping `MapProjection` as a type alias for the
   new abstract class, but it warrants its own design discussion.

I'd be glad to look at this as a follow-up — feels like the right cleanup but
not minimal to scope into this PR. If the Cesium team has a preference between
"registry" and "function source" for the user-extensibility story, I'd want
that pinned down before refactoring the base class.

mzschwartz5:
Hi @mickae1, with apologies if we've misunderstood or miscommunicated anything ... it appears that you've removed the required AI acknowledgement from our PR template, and are likely using an LLM in your responses to our review. As a small team we're limited on review bandwidth — as a general policy it's not fair to our team, or to the other contributors in our PR and issue backlogs — to spend review bandwidth in this way.
 
This is a large PR, which our team would presumably be maintaining if merged. To make an informed judgement as to whether that additional maintenance scope is worthwhile, we would need to have some confidence in your real need for this feature, your testing so far, and your ability to give an informed response if problems come up later. As undisclosed LLM usage significantly undermines our confidence in these areas, we are closing this PR. Please feel free to reopen this or another PR in the future, if you are willing to work within the review process on these terms. Thank you.

mickae1:
@mzschwartz5 yes I used llm ( claude code the best ) for helping me to answer because it analyzed better than i am the whole project.
also the modifications worked, could you review it? this pr is what all people are waiting for: custom projections, llm are goods to help developer. and every lines of code and every words was validated by me. the sandcastle works, it works on a real project. 

ai should be used to strengthened our work. 

i also tested the code, and it works.

it would be a waste to trash this PR, tons of people need custom projections. 

can you reopen the PR and takes your time to review it?

mickae1:
@mzschwartz5 it tooks me hours/weeks to do this PR, find the best way to do it, find a viable approach, I can't believe that this will be put in trash, there are at least 10 or more tickets for custom projection not working !!

mzschwartz5:
Hi @mickae1, I've reopened the PR - I'm sorry if I've caused any anxiety.

Please understand, though - we're in an era where it's becoming harder to tell what is legitimate and what is AI spam. We simply do not have the bandwidth to review every incoming PR (we're a small team). It's more important than ever to be transparent about where AI has been used, so we can determine whether it's been used with care; I agree that using AI can strengthen our work, but it can also weaken it if used indiscriminately. 

Going forward:

- **In this PR**: if you want to use an LLM to generate responses to reviewers' feedback, that should only be the starting point. You should still review the LLM's response, edit it, and distill it if needed. When _we_ ask for LLM review, we give it instructions to provide feedback in ways we find useful. When a _community member_ gives us LLM review, we don't have that opportunity. Please be mindful of our time.

- In future PRs: please do not delete the AI disclosure on the PR template, or we may once again close the PR out of precaution.

As for review of this particular PR - I originally only intended to leave a few small comments that came to mind, to seed discussion for anyone else who decided to give a deeper review. I don't have the time for a deeper review at the moment; hopefully we can get this some attention soon though.

javagl:
So... the issue of supporting other projections could have higher priority soon... https://en.wikipedia.org/wiki/Equal_Earth_projection#United_Nations_adoption

mickae1:
@mzschwartz5 I split the big commit into 7 logical commits to make it more readable for review:

  1. Core: extend MapProjection with introspection and serialization — abstract API (isNormalCylindrical, serialize/deserialize, approximateMaximumCoordinate), MapProjectionType enum, SerializedMapProjection
  wrapper.
  2. Core: extend geometry primitives for non-cylindrical projections — Rectangle.approximateProjectedExtents/approximateCartographicExtents, BoundingSphere.fromRectangleWithHeights2D, NaN guards in Ellipsoid
  and TerrainMesh.
  3. Core: add Proj4Projection — WKT / proj4-string based projection.
  4. Core: add Matrix4Projection — 4x4 affine projection.
  5. Core: add CustomProjection and the projection deserializer — user-defined project/unproject functions + deserializeMapProjection entry point for workers.
  6. Renderer/Scene: wire non-cylindrical projections into rendering — 5 GLSL automatic uniforms, Scene/FrameState/Camera/ScreenSpaceCameraController/PrimitivePipeline plumbing.
  7. Docs, Sandcastle demo, and changelog — Custom Projections Guide, Map Projections Sandcastle, CHANGES.md.

  Also rebased on top of the latest main (Cesium 1.146-dev). Ready for another look when you have a moment.

  Done with the help of @claude.

ggetz:
Hi all, we may want to consider the overall plan with this PR based on input from https://community.cesium.com/t/wed-love-your-input-coordinate-reference-systems-and-local-coordinates-in-cesiumjs/46968. This seems directly relevant.

I haven't taken a review pass yet, but I would want to understand the implications of depending on proj4 as part of the core library and whether we may want to consider adding a better plugin point for projection functions rather than an explicit dependency.

Issue #13291, opened 2026-03-11
Billboards throw errors when Zone.js patches Promise to ZoneAwarePromise


jhptc:
### What happened?

When using Cesium in an environment where Zone.js is loaded (for example, Angular or any app that uses Zone.js to patch window.Promise with ZoneAwarePromise), creating BillboardGraphics causes runtime errors in the browser console.

### Reproduction steps

1. Load an application using ZoneJS
2. Create a viewer
3. Add an entity with a BillboardGraphics

### Sandcastle example

https://sandcastle.cesium.com/#c=ZVJBbtswEPzKQhfJgLu04aZOVccI4PYWoIcAPQQCAppcS2woUuBSTp3Cfy8lOUHTHgRxhzPD4S6PMgDDDWiv+pZcRBVIRvpmaaiKnFUwXcxnXyrHyEElat7E2HEpRO+6pxqVb8WLd4Q/+XaBy0+4FPveaUs8wX2rsTUubeejiXfWS518Dr1T0XgHxQx+Vw5AecfeElpfF/nDZAkDmfQYAICjDHFHbPq2GJBz+t6SNyQ1yq4jp3eNsbrggVK5t3PeqacjhYCT7wPQL8PRuBrupdNKcrQEEzOl0gS1J4aGAs2BsMZy0B5T646Gnimkyzh6vgjwx4gVuRrLnXdRGkfhcoVJgSmwiYYYpdbFGAWg82yGoOWr0y7lTSvpVngIvv1KdSDi4sP6Cq8+r9frOXxc4GJ1fb2azSeLvbF272XQJVxMAUwrayohRxTjksVk/3jna//ojxSsPGHn6vxich7/57HB2TzbcDxZ2g7YrWk7HyL0wRbJLlLb2fRaOE1cPVF6PDz2fCNeJRttjmD0TZX9040qA2Ulc9o59Nbemxeqsu1GJP472TD+NJfvU8qB0iy3dxOIiBuRyv9V0Xu7l+Evxz8

### Environment

Browser: Any
CesiumJS Version: 1.139
Operating System: Windows

jhptc:
This issue appears to have been introduced in the following commit: https://github.com/CesiumGS/cesium/commit/9d1050e7ac88d51493342ed0bf15431d1b871052

[...]

Issue #7700, opened 2019-04-01
Error loading image for billboard: InvalidStateError: The source image could not be decoded.


tangtony:
Billboards loaded via a data-uri seem to have stopped working in Cesium v1.56.

Sandcastle example: 
v1.55 (Working):
https://cesiumjs.org/releases/1.55/Apps/Sandcastle/index.html#c=ZVHLbsIwEPwVKxdABTs0ICgNUSV6rNRDpZ5yceIlsXBsZJsEWvHvXROo+vBpZrwz61233JJWQgeWrImGjmzAyUND3y/acFBe6MZoz6UGOxg95jrXvYOC9tJLcJQLMfzMNcGzNw41o8nqFrXh1iPiOqFba5pnqCyAG04Wczp/WCwWYzKLaZwsl8lo3GcUUqnCcCvI6poajmx4BRg7ENzz1YUx11Z3x0aNUwSkBeuw8zqPpnSaRwQvtENWe79fMdZ1He0SamzF7uM4Dl4s6qTwdbDEMbIaZFX7G83SUtpSASmPKM1DQXm6IYtgFoDz1uwAWaF4ufsWJrfkBKUtToTQgsgjlqWhdzboRzvn+hy2Go2j1PmTgqzXn2SzN9aTg1VDSpmHZq84LpIVh3IHnpbOBRtmXU2pkC2RAtv8+bTwbMVdWMX2oNSb/IAwG8P6XzZluJC6esU9Kn4KJfU0e+lFSmnKkP53eWNUwe2PxC8

[...]

PR #13475, opened 2026-05-08
Handle promise replacements to avoid crash in BillboardTexture

javagl:
# Description

The `TextureAtlas.addImage` function can return either a `number` or a `Promise`. The `BillboardTexture` class checked whether this return value was `instanceof Promise`, and `await`ed it if necessary.

Apparently, there are ... "environments" (Zone.js, Vite, React, whatever...) that replace `Promise` with something else, causing that `instanceof` check to fail and cause crashes later. Specifically, this showed up as

> DeveloperError: Expected index to be typeof number, actual typeof was object

at a later point. 

This change here simply omits the `instanceof` check, and `await`s whatever is there. When it's already a `number`, then it stays a `number`. If it's something that is "awaitable", then it is awaited (hopefully, to become a `number`).

## Issue number and link

https://github.com/CesiumGS/cesium/issues/13291

## Testing plan

I can not sensibly test this with reasonable effort. I also won't try to set up some "mocking promise replacement polyfill substituion infrastructure" for some specs. Maybe someone can just look at the code and see that it makes sense, and maybe someone can try it out.

## Author checklist

## AI acknowledgment

- I used AI to generate content in this PR
  - [ ] Yes
  - [x] No


javagl:
Tagging @jhptc and @Beilinson due to the discussion in the issue. Maybe you want to give it a try.

jhptc:
From code inspection, I'm confident this solves my specific issue. I'll try to test later when I have some time to confirm within my full environment.

Beilinson:
If I'm not mistaken this was introduced specifically to fix either this test:
https://sandcastle.cesium.com/#c=pVZtT9tIEP4rI3QfbM6skwCnKgeoNEXXD0BOkLY6kYpu7Im9Zb22dtdAitLf3rG9ThxI0enuk70zz7zuM2OLrMi1hV3gBkZoRJnBXOcZTHei+jTd+XOqRAO65iqOuLESW8xaUuOmKsqVsXAv8AE1HIPCB+eVfaplXut3lCvLhUI93QngaaoATJqXMj6lYNziEKwuMajkFrUm5LB1NGnOrErhc65l7ASeT/Cl30nDRKiQsmjSYfWR1PWTJTKfIYuxsOkEjT1NyIWxzhcZVfFrXzLn8QXXd6iN13jnZqEimJcqsiJXsAFoamniC3pc8IKc8QcuLMzRRqlX6YF6l1pbmGEY3lshsc+iWLGMF2z/jx6LVDjoDfb7vcGb0BRaWHQP9s3kijpW+fCZTVF5nkbjw/EJ0LNWe36VZTeHS56hoSzGs28YWXaHC+O53LrQrKmBgKfUhEXdYO8JJKrEpkPo93o9WAbg3QYg4sc65lNTjEZbatWeABQFHMLXpinw2xPB4XfoL78GLaLIjaiaN4SbvYNDUl5wmzJNdKKYPtHxMIDt8i8rH1UJ1+XsCpPakavoZlXxzY8f3nPzlZI1ZflfWn/LuhPLpiGOMREhNWcGbcVez9UXE12E4k36jpQjri29cbVft+09JhrReHsHB1THmwD6eOgHqwDrns+ElLOc63iUS4kNnzbG5t1LQJtHzePhBrs3Q4QhzIWmKDyOweQZtldcKTuvAMwUUkTo9QKgpjmVaxGEMPB9h5vn+owTiR2mJgF5b27a21JN4AL5bVLddBmROiMe3NN9kJdt9p1aqiqI5Gui8jntBhjUNdAdTUSGeWk9r0vNtpb/lzhUxB/QAFSHZZVRuLtL4l14W3DNM3j69XUtt11yZRtO1WqH/LtUni+XFfuJMw2EbYjrzLc4rJvtGrSexNepzBhzIVoL300OLeyEePgfN1qhErfQnKfOSHfnIC9VLFRyRRVwlUj0tq4B9rhaD4f9fdh7pl28kKQoktRu3ynsQcQ2/YVuw7BtRZTLXK87WZ3YP2fn5+PPDnCP2oqIy7EWSeeb9mlDzN6NJ5PxhTNJcy2+V1/L50YfninY6OxycnblzOrch68U1BQwfK2s5nCFc9SoIlxH3pSz0fnpxd+3k/HtX1fjj5fvnXUhHlGO53MazY2rXBFsUO2bve5ipNHaCXaOjF1IPKmkb92vR6klMTC0mBWSfg9MOCujO7QsMqayOwpbk6NY3NPn6XjLfwZEkhtDmnkp5bX4Tv8sJ0ch4TfMqq858WxMNyX5ooKk/ZPzRkgzcBTS8aWVzXM547rj8Sc

Or the Entity API version. So this PR is basically reverting the fix

packages/engine/Source/Scene/BillboardTexture.js
```diff
@@ -200,9 +200,12 @@ BillboardTexture.prototype.loadImage = async function (
...
     index = atlas.addImage(id, image, width, height);
-    if (index instanceof Promise) {
-      index = await index;
-    }
+    // The index MIGHT be a promise here. It is not possible to sensibly check this
+    // with "instanceof Promise", because of smart frameworks replace "Promise"
+    // with ... something else. See https://github.com/CesiumGS/cesium/issues/13291
+    // Simply awaiting here will handle any promise-like objects, as well as the
+    // case where the value already is a number.
+    index = await index;
```
Beilinson:
```js
if (!(typeof index ===  'number') {
  index = await index;
}
```
a better solution, this keeps this function syncronous if possible (which was the point of the original bug fix), while not relying on problematic instanceof, since `typeof` always returns `"number"` on the index unless its a promise and zone.js (the devil) can't change this type check (unlike instanceof which fails for sad reasons).

Beilinson: I even did this check in this way in another place further down where the subRegion code is:
```js
  if (typeof indexOrPromise === "number") {
    this.setImageSubRegion(indexOrPromise, subRegion);
    return;
  }
```

No idea why I didn't keep it consistent here

javagl: OK, more broadly: I actually think that calling it `indexOrPromise` and let `index` be assigned _once_ is a cleaner solution (and could shorten that inlined comment a bit).

But I'm surprised: 

>  this keeps this function syncronous if possible

Is it really _not_ synchronous when there is this `await`? I thought that `await aNumber` should essentially be a no-op (and that it is only _truly_ asynchronous when this thing actually is 'awaitable'),

(I don't know the VM details - it might be that this `await aNumber` already causes the execution to be suspended in terms of the execution flow...)

However, I'll have another look at the sandcastle that you posted, and _likely_ change the fix to use the `typeof .. number` approach soon.

javagl:
I didn't have the whole context from https://github.com/CesiumGS/cesium/pull/13075 on the radar, but as @Beilinson correctly pointed out, the original change here indeed seemed to re-introduce the issue that this PR was fixing (or at last "an" issue): The sandcastle from the above comment inserts new billboards after a while, and the newly inserted ones had been broken on this branch:

Cesium Billboards Promise

So apparently using
`index = await index;`
is **not** a no-op when `index` already is a number. Good to know.

I now changed it to use a `typeof ... number` check. And from a local test, the sandcastle seems to work now (and of course, it still avoids the `instanceof Promise` that did cause the main issue here)

packages/engine/Source/Scene/BillboardTexture.js
Beilinson: yeah its not no-op, it forces the code to wait 1 microtask, essentially wrapping whatever follows in `Promise.resolve()` which will force a sync value into a "promise containing that value" and then awaits it.

rileyhowley-seequent:
Will this fix be in the next release?

javagl:
I just updated CHANGES.md and merged `main`. 

It's a relatively small fix, but it's difficult to test it in an _actual_ environment where the issue appeared, _unless_ you already have set up such an environment. 

Maybe someone can review it on the code level, and approve it with a "What's the worst thing that could happen?" (knowing that the answer is "A crash", so... no change 😁)

@mzschwartz5 You reviewed other TextureAtlas/Billboard related PRs, I hope it's OK to tag you here

jhptc:
I managed to get some free time today to confirm this fixes the zone related problems.

1. Upgraded my company project to the latest cesium version and tested. This continued to fail as expected.
2. Replaced the cesium version in my package.json with a file symlink to a copy of this branch. No errors and everything appears to work normally. I confirmed it is working for both the Unminified and release versions of Cesium.
3. For completeness, I also ran the cesium server with `npm start` and copy pasted the issue's sandcastle reproduction into my locally served sandcastle to confirm that it also works after the changes.

javagl:
Thanks for the feedback @jhptc , that will give additional confidence for the review (in addition to the (tiny) code change itself)

ggetz:
@rileyhowley-seequent I think the major hold up here was confidence in fix. The original issue was difficult to reproduce, and so it's also difficult to verify this fixes the issue once and for all.

It would be helpful if you're able to test this branch and confirm the new behavior works as expected.

rileyhowley-seequent:
So I pulled this branch locally and ran it within our app. We no longer got that error but I an unsure if that was a valid pass. What happened was we actually resulted in 
image
Which I was able to narrow it down to `import '@cesium/engine/Source/Widget/CesiumWidget.css';` commenting this out stopped the error but resulted in a black screen with a distant moon. Usually the error would show when the earth is loaded and then the labels get put in the scene. So I think that this error is happening before that fix is applied

ggetz:
> I was able to narrow it down to import '@cesium/engine/Source/Widget/CesiumWidget.css'; commenting this out stopped the error but resulted in a black screen with a distant moon. Usually the error would show when the earth is loaded and then the labels get put in the scene. So I think that this error is happening before that fix is applied

@rileyhowley-seequent Commenting this out may just be resulting in the error box not displaying due to lack of CSS to position it. But the error may still be thrown if you look in the developer console.

My guess is there is an image being loaded which for some reason is failing to be decoded. It'd be helpful to understand what that image is, or at least the format. For context, [this has come up before for image data loaded from data URIs](https://github.com/CesiumGS/cesium/issues/7700) and may be a similar problem.

Would you be able to give us some clues about the image or images used for the billboards?

rileyhowley-seequent:
Apologies for the delay. The affected code is the text on the label causing the error
```const entity = cesiumWidget.entities.add({
            position: cluster.center,
            id: `${cluster.id} ${cluster.count}`,
            label:
                cluster.count > 1
                    ? {
                          //   text: cluster.count.toString(),
                          font: '14px sans-serif',
                          fillColor: Color.fromCssColorString('rgba(0, 0, 0, 1.0)'),
                          style: LabelStyle.FILL_AND_OUTLINE,
                          eyeOffset: new Cartesian3(0, 0, -20),
                          verticalOrigin: VerticalOrigin.CENTER,
                          horizontalOrigin: HorizontalOrigin.CENTER,
                      }
                    : undefined,
            point:
                cluster.count > 1
                    ? {
                          pixelSize: pointSize,
                          color: Color.fromCssColorString('rgba(255, 255, 255, 1.0)'),
                          outlineColor: Color.fromCssColorString('rgba(255, 255, 255, 0.5)'),
                          outlineWidth: 10,
                      }
                    : undefined,
            billboard:
                cluster.count === 1
                    ? {
                          scale: 0.4,
                          verticalOrigin: VerticalOrigin.CENTER,
                          horizontalOrigin: HorizontalOrigin.CENTER,
                          id: cluster.count.toString(),
                      }
                    : undefined,
            show: true,
        });```

**Comment by rileyhowley-seequent** [comment]
https://github.com/user-attachments/assets/e15d3c6d-c532-49a5-ac8e-27f71af3bee2

**Comment by ggetz** [comment]
Thanks @rileyhowley-seequent! We're taking a look.

File: packages/engine/Source/Scene/BillboardTexture.js [hunk 2]
```diff
@@ -199,9 +199,11 @@ BillboardTexture.prototype.loadImage = async function (
...
   const atlas = this._billboardCollection.textureAtlas;
   try {
-    index = atlas.addImage(id, image, width, height);
-    if (index instanceof Promise) {
-      index = await index;
+    const indexOrPromise = atlas.addImage(id, image, width, height);
+    if (typeof indexOrPromise === "number") {
+      index = indexOrPromise;
+    } else {
+      index = await indexOrPromise;
```
mzschwartz5, line 206: Would this also work? Seems a bit simpler to me (no else branch. Just a style preference)

```JS
let index = atlas.addImage(id, image, width, height);
if (typeof index !== "number") { // then it's a promise
    index = await index;
}
```

mzschwartz5 (approved):
In general, I don't love the union type returned from `atlas.addImage`, but I probably already voiced that distaste in the original PR and conceded that there wasn't an immediately better solution (it's a deeper issue with JS async machinery. Perhaps though `Atlas` should have separate methods for sync/async image adding, and a method like `isImageAlreadyCached` to help callers call the right one...)

In any case, this fix is fine in my eyes. I have one style nit, but otherwise LGTM. Feel free to merge 🚢

packages/engine/Source/Scene/BillboardTexture.js
javagl: Yes, this should also work. Naming is difficult, and there may be subjective elements to this. But when I see a variable called `index`, then I'd expect this to be a `number` (and even more specifically, an integer value), and `await`ing that makes no sense. The elaborate and quirky name `indexOrPromise` is an implicit ⚠️ and acknowledging that ... "things are not ideal".

javagl:
> I probably already voiced that distaste in the original PR and conceded that there wasn't an immediately better solution 

Yeah, roughly around https://github.com/CesiumGS/cesium/pull/13075#discussion_r2586601283 

> (it's a deeper issue with JS async machinery. Perhaps though Atlas should have separate methods for sync/async image adding

Yes (for both points - i.e. it's broken, and we cannot trivially "fix" it). I also see `typeof/instanceof` as red flags. There _may_ be cases where it makes sense, But in the vast majority of cases it's hinting at a design flaw. This flaw is often "missing polymorphism", but in this case, it's not that easy.

packages/engine/Source/Scene/BillboardTexture.js
mzschwartz5: Naming aside, I mainly just prefer the simpler conditional

ggetz:
@javagl In the interest of getting the in for the release tomorrow, I can go ahead and update if you're not already on it.

javagl:
@ggetz I "started merging this" earlier today, but was hesitant due to the failing builds. It looks all fine now, so I'll just hit "merge".

PR #13477, opened 2026-05-08
BufferPrimitiveCollection: Support precomputed 'boundingVolume' constructor param

donmccurdy:
# Description

Adds a 'boundingVolume' constructor parameter to BufferPointCollection, BufferPolylineCollection, and BufferPolygonCollection. When specified, this parameter is used as a static, model-space bounding volume for the collection, and the collection will not attempt to auto-update the bounding volume when primitives are added or positions are changed, allowing greater user control of performance for large, dynamic collections. When unspecified, the collection continues to automatically compute and update the bounding volume.

## Issue number and link

- #13444

## Testing plan

Tested with large collections of animated points updated each frame. Sandcastle copied from https://github.com/CesiumGS/cesium/pull/13465, updating PRIMITIVE_COUNT to 500,000. On my machine, the sandcastle previously spent 6ms per frame updating the collection bounding volume, and that cost is fully eliminated by pre-computing and caching the bounding volume. The sandcastle remains a bit slow, but cutting 6ms here eliminates one source of the overhead.

Added unit tests to cover the new and old behaviors.

## Author checklist

## AI acknowledgment

#### PR Dependency Tree

* **PR #13477** 👈


danielzhong (commented):
Thanks!

1. In `BufferPrimitiveCollection.js`, line 449, `clone()` copies `_boundingVolume` but never updates `_boundingVolumeWC`
```
collection.boundingVolume.clone(result.boundingVolume);
// Porbably add this:
BoundingSphere.transform(result._boundingVolume, result._modelMatrix, result._boundingVolumeWC);
```

2. nit: When positions change on a static collection, `_makeDirtyBoundingVolume()` still sets `_dirtyBoundingVolume = true` each frame. I don't think there's any impact, but still worth to mentioning, should we silence it?
3. The new JSDoc says "users are responsible for updating bounding volume as needed", but if a user mutates collection.boundingVolume post construction, collection.boundingVolumeWC (used for culling) won't be updated automatically in static mode. Worth adding a note like:
`After updating, call BoundingSphere.transform(collection.boundingVolume, collection.modelMatrix, collection.boundingVolumeWC) to keep the world-space bounding volume in sync.`

Or we sync world space `BoundingSphere.transform(...)`  in static mode?

donmccurdy:
Hm, good points @danielzhong, thanks! Since this is still an experimental API, and we only actually need the world-space bounding volume for rendering, I think the simplest solution might be to consolidate the two .boundingVolume / .boundingVolumeWC properties into just one, such that .boundingVolume is now in world space. Then we don't need side effects to keep the two in sync. Updated!

> When positions change on a static collection, _makeDirtyBoundingVolume() still sets _dirtyBoundingVolume = true each frame...

I like that change – so we can just prevent setting the dirty flag when auto-updates are disabled, rather than ignoring the dirty flag. Done.

danielzhong (approved):

